PR #214, opened 2017-01-17
openshift: use new k8s.io/apimachinery GOPATHs for kube API

erikh:
This is the problem with the CI. However, the build system has some dep mgmt issues too, so I'm not sure this will pass either; but this is at least part of the fix. I'm going to try putting together a build system today for testing with dependencies, I have everything working with godep minus some mishaps. I'll report more later.

### DIFF
--- a/openshift/openshift-copies.go
+++ b/openshift/openshift-copies.go
@@ -18,9 +18,9 @@ import (
 	"github.com/ghodss/yaml"
 	"github.com/imdario/mergo"
 	"github.com/pkg/errors"
-	utilerrors "k8s.io/kubernetes/pkg/util/errors"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	utilnet "k8s.io/apimachinery/pkg/util/net"
 	"k8s.io/kubernetes/pkg/util/homedir"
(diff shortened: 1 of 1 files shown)


erikh:
This isn't going to pass until skopeo's vendored copy passes too. I'm working on it now.

erikh:
I've tried to work this out, so I removed the skopeo test. Once this is merged, vndr can be updated on skopeo, and the test for skopeo can be added back in here. Sorry if it's overly aggressive for a first-time contribution, but it's the best thing I could think of. :)

runcom:
@erikh hey, you can follow this https://github.com/projectatomic/skopeo#dependencies-management and leave the skopeo failure here. After you prepare a test PR in skopeo (following https://github.com/projectatomic/skopeo#dependencies-management) and we're green over there we might just merge this and re-vendor c/image in skopeo. If you need any help ping me on Twitter or wherever you want.

/cc @mtrmac

erikh:
This is done as mentioned on the ticket here: https://github.com/projectatomic/skopeo/pull/289

mtrmac:
:+1: based on test results in https://github.com/projectatomic/skopeo/pull/289 . Thanks!

runcom:
:+1:

PR #215, opened 2017-01-18
Create test-pristine target used to align dev and CI environments

erikh:
This is a Dockerfile + make targets to enable pristine environment testing,
which will help with isolating dependency problems. In the event the CI and
developer's workstation do not align, they can quickly recreate (theoretically
at least) the problem by running `make test-pristine`.

Hope this is useful. Thanks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,16 +3,11 @@
   sudo: required
   notifications:
     email: false
+  services:
+    - docker
   go:
     - 1.7
-  install: make deps
-  script: make .gitvalidation && make validate && make test && make test-skopeo
+  script: make test-pristine
   dist: trusty
   os:
     - linux
-  addons:
-    apt:
-      packages:
-        - btrfs-tools
-        - libdevmapper-dev
-        - libgpgme11-dev
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,27 @@
+FROM golang
+
+ENV SRCPATH /go/src/github.com/containers/image
+ENV DOCKER_VERSION 1.13.0
+ENV DOCKER_PATH docker-${DOCKER_VERSION}.tgz
+
+# needed because if the tests run as root the tests that test permissions will fail.
+RUN useradd -d /home/gouser -m -s /bin/bash gouser
+
+# NOTE: iptables and psmisc are required by docker.
+RUN apt-get update -qq && apt-get install -y sudo iptables psmisc btrfs-tools libdevmapper-dev libgpgme11-dev
+
+RUN wget -q https://get.docker.com/builds/Linux/x86_64/${DOCKER_PATH}
+RUN tar -xpf ${DOCKER_PATH} --strip-components=1 -C /usr/bin
+RUN gpasswd -a gouser root
+
+# dind is required for the skopeo tests
+COPY dind /dind
+RUN chmod +x /dind
+
+COPY . ${SRCPATH}
+WORKDIR ${SRCPATH}
+RUN make deps
+RUN chown -R gouser:gouser ${SRCPATH}
+
+CMD sh -c "cd ${SRCPATH} && bash docker-test.sh"
+ENTRYPOINT ["/dind"]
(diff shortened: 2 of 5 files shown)


erikh:
Oh I just realized this contains the fixes from PR #214 as well, since they're
required to pass.

erikh:
might need to kick this build, the hooks aren't firing, sorry for the trouble.

erikh:
Going to wait until #214 shakes out to move forward with this PR, will leave open for now.

.travis.yml
```diff
@@ -3,16 +3,11 @@
   sudo: required
   notifications:
     email: false
+  services:
+    - docker
   go:
     - 1.7
-  install: make deps
-  script: make .gitvalidation && make validate && make test && make test-skopeo
+  script: make test-pristine
```
mtrmac: Do we really need to use an extra container for Travis? I’d expect Travis to give us a fresh VM every time.

(I am really mostly worried whether this would make the tests run significantly longer. I guess the only way to tell is to try in this PR.)

erikh:
The main reason it's in travis as well is to keep the environments aligned
-- e.g., "I'm seeing X problem in CI and I want to trivially recreate it".
I realize that doesn't mean much in the context of running against travis
*itself*, but if we keep it so that it's easier to do than following the
steps in travis by hand, then we have a win for close-to-free I think.

Right now I'm struggling with the git-validation and some travis env vars
and the skopeo tests are still busted until things are fixed over there, so
I haven't validated that the skopeo tests work inside the container, and
you're right, that would change things. Maybe it's just an issue of passing
the wrong vars; I hadn't gone through the git-validation source yet. It's
also why you're seeing a lot of churn in that area.

I'll respond to the rest inline.

Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+RUN useradd -d /home/gouser -m -s /bin/bash gouser
+RUN apt-get update -qq && apt-get install -y btrfs-tools libdevmapper-dev libgpgme11-dev
+
+COPY . ${SRCPATH}
+WORKDIR ${SRCPATH}
+RUN make deps
+RUN chown -R gouser:gouser ${SRCPATH}
+
+USER gouser
+CMD make .gitvalidation validate test
```
erikh: the only reason I've been dropping test-skopeo in these changes is because the build is busted atm.

per discussion on #214 these will be reincorporated after the skopeo changes are merged, and before #214 itself is merged.

I will adjust the make target.

Makefile
```diff
@@ -53,8 +59,8 @@ EPOCH_TEST_COMMIT ?= e68e0e1110e64f906f9b482e548f17d73e02e6b1
...
 	@which git-validation > /dev/null 2>/dev/null || (echo "ERROR: git-validation not found. Consider 'make deps' target" && false)
-ifeq ($(TRAVIS),true)
-	@git-validation -q -run DCO,short-subject,dangling-whitespace
-else
 	@git-validation -q -run DCO,short-subject,dangling-whitespace -range $(EPOCH_TEST_COMMIT)..HEAD
-endif
+# ifeq ($(TRAVIS),true)
+# 	@git-validation -q -run DCO,short-subject,dangling-whitespace
+# else
+# endif
```
erikh: this is temporary while I sort out the travis issue (which I think you found above)

mtrmac:
> and the skopeo tests are still busted until things are fixed over there

FWIW that should no longer be the case, but `docker/distribution/reference` has changed its API and broken us. Looking into that now.

erikh:
This should pass now (including the skopeo tests). I also attempted to comment
the dockerfile / bootstrap script a bit.

I needed to bring in DIND because of the skopeo tests. The VFS behind it makes
things a little slower but I think that's ok for this.

erikh:
do you want me to revisit this now that we have the new vendoring strategy, or should I close it?

mtrmac:
I don’t know; do you think this is still useful? The vendoring should handle the _most frequent_ breakages, and it is simpler/faster than creating and running containers and `dind` and all that infrastructure, and there won’t be any surprises with `docker build` caching. Were you running into non-vendoring (tool-related?) breakages which needed a container to diagnose?

Simple is nice, OTOH working is more important than simple.

erikh:
I'm not as sure now. Let's table this atm. I'll close in a sec here.

erikh:
we can always reopen it later, I'll keep the branch.

PR #216, opened 2017-01-18
*: pluggable transports

### LINKED ISSUE #125: pluggable transports?
Right now if you import containers/image in docker you'll get tons of dependencies because of the openshift transport where instead docker only needs the docker transport.
There could be a way to setup transports by the caller instead of hardcoding them into https://github.com/containers/image/blob/master/transports/transports.go#L23

@mtrmac wdyt?

runcom:
Close #125 

As discussed in #125, this patch creates a new subpkg to import transports independently. projectatomic/docker would greatly benefit from this as it needs just the docker transport (right now).

Those will be the stats in projectatomic/docker + this PR:
```
Showing  76 changed files  with 1,715 additions and 6,770 deletions.
```

@mtrmac PTAL (far from being ready, just a start...)

### DIFF
--- a/signature/policy_config.go
+++ b/signature/policy_config.go
@@ -122,10 +122,8 @@ func (m *policyTransportsMap) UnmarshalJSON(data []byte) error {
 	// So, use a temporary map of pointers-to-slices and convert.
 	tmpMap := map[string]*PolicyTransportScopes{}
 	if err := paranoidUnmarshalJSONObject(data, func(key string) interface{} {
-		transport, ok := transports.KnownTransports[key]
-		if !ok {
-			return nil
-		}
+		// transport can be nil
+		transport := transports.Get(key)
 		// paranoidUnmarshalJSONObject detects key duplication for us, check just to be safe.
 		if _, ok := tmpMap[key]; ok {
 			return nil
@@ -155,7 +153,7 @@ func (m *PolicyTransportScopes) UnmarshalJSON(data []byte) error {
 }
 
 // policyTransportScopesWithTransport is a way to unmarshal a PolicyTransportScopes
-// while validating using a specific ImageTransport.
+// while validating using a specific ImageTransport if not nil.
 type policyTransportScopesWithTransport struct {
 	transport types.ImageTransport
 	dest      *PolicyTransportScopes
@@ -174,7 +172,7 @@ func (m *policyTransportScopesWithTransport) UnmarshalJSON(data []byte) error {
 		if _, ok := tmpMap[key]; ok {
 			return nil
 		}
-		if key != "" {
+		if key != "" && m.transport != nil {
 			if err := m.transport.ValidatePolicyConfigurationScope(key); err != nil {
 				return nil
 			}
--- a/signature/policy_config_test.go
+++ b/signature/policy_config_test.go
@@ -9,6 +9,8 @@ import (
 
 	"github.com/containers/image/directory"
 	"github.com/containers/image/docker"
+	// this import is needed  where we use the "atomic" transport in TestPolicyUnmarshalJSON
+	_ "github.com/containers/image/openshift"
 	"github.com/containers/image/types"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -244,6 +246,11 @@ func TestPolicyUnmarshalJSON(t *testing.T) {
 					xNewPRSignedByKeyData(SBKeyTypeSignedByGPGKeys, []byte("RHatomic"), NewPRMMatchRepository()),
 				},
 			},
+			"unknown": {
+				"registry.access.redhat.com/rhel7": []PolicyRequirement{
+					xNewPRSignedByKeyData(SBKeyTypeSignedByGPGKeys, []byte("RHatomic"), NewPRMMatchRepository()),
+				},
+			},
 		},
 	}
 	validJSON, err := json.Marshal(validPolicy)
@@ -269,9 +276,6 @@ func TestPolicyUnmarshalJSON(t *testing.T) {
 		func(v mSI) { v["transports"] = []string{} },
 		// "default" is an invalid PolicyRequirements
 		func(v mSI) { v["default"] = PolicyRequirements{} },
-		// A key in "transports" is an invalid transport name
-		func(v mSI) { x(v, "transports")["this is unknown"] = x(v, "transports")["docker"] },
-		func(v mSI) { x(v, "transports")[""] = x(v, "transports")["docker"] },
 	}
 	for _, fn := range breakFns {
 		err = tryUnmarshalModifiedPolicy(t, &p, validJSON, fn)
--- a/signature/policy_eval_test.go
+++ b/signature/policy_eval_test.go
@@ -5,8 +5,10 @@ import (
 	"os"
 	"testing"
 
+	"github.com/containers/image/docker"
 	"github.com/containers/image/docker/policyconfiguration"
 	"github.com/containers/image/docker/reference"
+	"github.com/containers/image/transports"
 	"github.com/containers/image/types"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -101,6 +103,37 @@ func (ref pcImageReferenceMock) NewImageDestination(ctx *types.SystemContext) (t
 }
 func (ref pcImageReferenceMock) DeleteImage(ctx *types.SystemContext) error {
 	panic("unexpected call to a mock function")
+}
+
+func TestPolicyContextRequirementsForImageRefNotRegisteredTransport(t *testing.T) {
+	transports.Delete("docker")
+	assert.Nil(t, transports.Get("docker"))
+
+	defer func() {
+		assert.Nil(t, transports.Get("docker"))
+		transports.Register(docker.Transport)
+		assert.NotNil(t, transports.Get("docker"))
+	}()
+
+	pr := []PolicyRequirement{
+		xNewPRSignedByKeyData(SBKeyTypeSignedByGPGKeys, []byte("RH"), NewPRMMatchRepository()),
+	}
+	policy := &Policy{
+		Default: PolicyRequirements{NewPRReject()},
+		Transports: map[string]PolicyTransportScopes{
+			"docker": {
+				"registry.access.redhat.com": pr,
+			},
+		},
+	}
+	pc, err := NewPolicyContext(policy)
+	require.NoError(t, err)
+	ref, err := reference.ParseNormalizedNamed("registry.access.redhat.com/rhel7:latest")
+	require.NoError(t, err)
+	reqs := pc.requirementsForImageRef(pcImageReferenceMock{"docker", ref})
+	assert.True(t, &(reqs[0]) == &(pr[0]))
+	assert.True(t, len(reqs) == len(pr))
+
 }
 
 func TestPolicyContextRequirementsForImageRef(t *testing.T) {
--- a/transports/alltransports/alltransports.go
+++ b/transports/alltransports/alltransports.go
@@ -1,0 +1,31 @@
+package alltransports
+
+import (
+	"strings"
+
+	// register all known transports
+	// NOTE: Make sure docs/policy.json.md is updated when adding or updating
+	// a transport.
+	_ "github.com/containers/image/directory"
+	_ "github.com/containers/image/docker"
+	_ "github.com/containers/image/docker/daemon"
+	_ "github.com/containers/image/oci/layout"
+	_ "github.com/containers/image/openshift"
+	_ "github.com/containers/image/storage"
+	"github.com/containers/image/transports"
+	"github.com/containers/image/types"
+	"github.com/pkg/errors"
+)
+
+// ParseImageName converts a URL-like image name to a types.ImageReference.
+func ParseImageName(imgName string) (types.ImageReference, error) {
+	parts := strings.SplitN(imgName, ":", 2)
+	if len(parts) != 2 {
+		return nil, errors.Errorf(`Invalid image name "%s", expected colon-separated transport:reference`, imgName)
+	}
+	transport := transports.Get(parts[0])
+	if transport == nil {
+		return nil, errors.Errorf(`Invalid image name "%s", unknown transport "%s"`, imgName, parts[0])
+	}
+	return transport.ParseReference(parts[1])
+}
--- a/transports/transports.go
+++ b/transports/transports.go
@@ -2,52 +2,61 @@ package transports
 
 import (
 	"fmt"
-	"strings"
+	"sync"
 
-	"github.com/containers/image/directory"
-	"github.com/containers/image/docker"
-	"github.com/containers/image/docker/daemon"
-	ociLayout "github.com/containers/image/oci/layout"
-	"github.com/containers/image/openshift"
-	"github.com/containers/image/storage"
 	"github.com/containers/image/types"
-	"github.com/pkg/errors"
 )
 
-// KnownTransports is a registry of known ImageTransport instances.
-var KnownTransports map[string]types.ImageTransport
+// knownTransports is a registry of known ImageTransport instances.
+type knownTransports struct {
+	transports map[string]types.ImageTransport
+	mu         sync.Mutex
+}
+
+func (kt *knownTransports) Get(k string) types.ImageTransport {
+	kt.mu.Lock()
+	t := kt.transports[k]
+	kt.mu.Unlock()
+	return t
+}
+
+func (kt *knownTransports) Remove(k string) {
+	kt.mu.Lock()
+	delete(kt.transports, k)
+	kt.mu.Unlock()
+}
+
+func (kt *knownTransports) Add(t types.ImageTransport) {
+	kt.mu.Lock()
+	defer kt.mu.Unlock()
+	name := t.Name()
+	if t := kt.transports[name]; t != nil {
+		panic(fmt.Sprintf("Duplicate image transport name %s", name))
+	}
+	kt.transports[name] = t
+}
+
+var kt *knownTransports
 
 func init() {
-	KnownTransports = make(map[string]types.ImageTransport)
-	// NOTE: Make sure docs/policy.json.md is updated when adding or updating
-	// a transport.
-	for _, t := range []types.ImageTransport{
-		directory.Transport,
-		docker.Transport,
-		daemon.Transport,
-		ociLayout.Transport,
-		openshift.Transport,
-		storage.Transport,
-	} {
-		name := t.Name()
-		if _, ok := KnownTransports[name]; ok {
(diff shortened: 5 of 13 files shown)


copy/copy.go
```diff
@@ -13,8 +13,9 @@ import (
 	"github.com/Sirupsen/logrus"
 	"github.com/containers/image/image"
 	"github.com/containers/image/manifest"
+	"github.com/containers/image/registeredtransports"
 	"github.com/containers/image/signature"
-	"github.com/containers/image/transports"
+	_ "github.com/containers/image/transports"
```
mtrmac: Why this? This would force any user of `copy.Image` to import all transports.

signature/policy_config.go
```diff
@@ -123,7 +123,7 @@ func (m *policyTransportsMap) UnmarshalJSON(data []byte) error {
 	// So, use a temporary map of pointers-to-slices and convert.
 	tmpMap := map[string]*PolicyTransportScopes{}
 	if err := paranoidUnmarshalJSONObject(data, func(key string) interface{} {
-		transport, ok := transports.KnownTransports[key]
+		transport, ok := registeredtransports.KnownTransports[key]
```
mtrmac: This will need more work; otherwise an user who imports only a subset of transports would refuse to use a `policy.json` which defines policies for other transports.

I suppose unmarshaling and remarshaling a `policy.json` should always keep it unmodified, so we can’t just drop the sections for unknown transports; instead, `policyTransportScopesWithTransport` will need to be taught to deal with a `nil` transport, and silently accept any key. It seems that `requirementsForImageRef` does the right thing, or close enough to it, even in such a situation.

It bothers me a bit that we would accept any typo in transport names, though… perhaps we could add a `registeredtransports.AllExistingTransportsRegistered bool = false`, with the “all-transports” `transports.init()` seting it to `true`, and refusing unknown values if `AllExistingTransportsRegistered`? The downside is that that would make it impossible to add policies private transports not known to `containers/image/transports` for a single-application-use.

signature/signature_test.go
```diff
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"testing"
 
+	_ "github.com/containers/image/openshift"
```
mtrmac: What is this needed for?

transports/transports.go
```diff
@@ -30,10 +26,10 @@ func init() {
 		storage.Transport,
 	} {
 		name := t.Name()
-		if _, ok := KnownTransports[name]; ok {
-			panic(fmt.Sprintf("Duplicate image transport name %s", name))
+		if _, ok := registeredtransports.KnownTransports[name]; ok {
+			continue
 		}
-		KnownTransports[name] = t
+		registeredtransports.Register(t)
```
mtrmac: These transports should have been registered merely by the fact of being imported here, shouldn’t they? So this could just verify their presence and panic if they are not already in `KnownTransports`. Or is the import and initialization order uncertain?

mtrmac:
> Or we could have `containers/image/transports` (contains `ImageName`), `containers/image/transports/registry` (contains `KnownTransports`) and `containers/image/transports/alltransports` (contains the full list of transports, and `ParseImageName`). But that feels like overengineering it.

Actually, we only need two packages. Perhaps (`/transports`, `/registeredtransports`)→(`/alltransports`, `/transports`), or (`transports/alltransports`, `/transports`)? We are breaking the API for one of the two functions either way, and `transports.ImageName` reads a bit better while `alltransports.ParseImageName` is still acceptable.

signature/signature_test.go
runcom: wrong, file this is needed in `policy_config_test.go` where we use the `atomic` transport in https://github.com/containers/image/blob/master/signature/policy_config_test.go#L242. `docker` and `dir` are already imported in that test file so we just need this or it fails with:
```
--- FAIL: TestPolicyUnmarshalJSON (0.00s)
        Error Trace:    policy_config_test.go:255
        Error:          Received unexpected error "Unknown key \"atomic\""
```
Btw, fixed

copy/copy.go
runcom: right, my bad, skopeo should actually import `transports` to register them all

signature/policy_config.go
runcom: > This will need more work; otherwise an user who imports only a subset of transports would refuse to use a policy.json which defines policies for other transports.
I suppose unmarshaling and remarshaling a policy.json should always keep it unmodified, so we can’t just drop the sections for unknown transports; instead, policyTransportScopesWithTransport will need to be taught to deal with a nil transport, and silently accept any key. It seems that requirementsForImageRef does the right thing, or close enough to it, even in such a situation.

Could you shed some light on how you want this to be done in the code?

> It bothers me a bit that we would accept any typo in transport names, though… perhaps we could add a registeredtransports.AllExistingTransportsRegistered bool = false, with the “all-transports” transports.init() seting it to true, and refusing unknown values if AllExistingTransportsRegistered? The downside is that that would make it impossible to add policies private transports not known to containers/image/transports for a single-application-use.

Not sure about this, it bothers me that ppl will need to know about this and set/unset this. It could be fine though.

transports/transports.go
runcom: no need to panic I guess, I dropped this altogether.

runcom:
> Actually, we only need two packages. Perhaps (/transports, /registeredtransports)→(/alltransports, /transports), or (transports/alltransports, /transports)? We are breaking the API for one of the two functions either way, and transports.ImageName reads a bit better while alltransports.ParseImageName is still acceptable.

done this, just have a question around signatures.

runcom:
@cyphar could you also take a look at this? We are mainly trying to decouple each transports wrt the signature pkg. Right now, if you use the signature package, you are required to _pull_ every other transport c/image has (that means, stuff like the storage transport which pulls in many files). This change aims at making easy to use the signature package with just a subset of interested transports (i.e. just docker, or oci and docker ...). Skopeo will keep pulling in all transports via the new `transports/alltransports` package but users of copy&signature pkgs can now choose which transport to use.

signature/policy_config.go
mtrmac: > > This will need more work; otherwise an user who imports only a subset of transports would refuse to use a policy.json which defines policies for other transports. …policyTransportScopesWithTransport will need to be taught to deal with a nil transport, and silently accept any key. It seems that requirementsForImageRef does the right thing, or close enough to it, even in such a situation.

> Could you shed some light on how you want this to be done in the code?

- In `policyTransportsMap.UnmarshalJSON`, if `KnownTransports[key]` does not exist, just use `nil`.
- In `policyTransportScopesWithTransport.UnmarshalJSON`, if `m.transport` is `nil` (from the change above), skip the `ValidatePolicyConfigurationScope` call (accept any scope string)
- For `requirementsForImageRef`, add a test that when a transport is not in `KnownTransports` when parsing the policy, but it is in `KnownTransports` at the time of `requirementsForImageRef`, the correct set of scopes is used (more or less re-run the existing test, except creating the policy in a different way).
- Plus as close to 100% test coverage as possible.

I hope that makes some sense.

> > It bothers me a bit that we would accept any typo in transport names, though… perhaps we could add a registeredtransports.AllExistingTransportsRegistered bool = false, with the “all-transports” transports.init() seting it to true, and refusing unknown values if AllExistingTransportsRegistered? The downside is that that would make it impossible to add policies private transports not known to containers/image/transports for a single-application-use.

> Not sure about this, it bothers me that ppl will need to know about this and set/unset this. It could be fine though. 

The idea is that this would work automatically for users who use both all of the known `AllTransports`, and any subset of them. The other side of this is that there would be nothing to manage, no extra work but also no flexibility; if `skopeo` thought it knows about all possible transport names, then it still possible to create a local tool with `mypinkponytransport`, but the system-wide `/etc/containers/policy.json` must not contain a policy for `mypinkponytransport` because `skopeo` would reject that policy file.

Yeah, not sure about this either.

signature/signature_test.go
mtrmac: Thanks. Add a comment perhaps?

transports/alltransports/alltransports.go
```diff
@@ -0,0 +1,29 @@
...
+	_ "github.com/containers/image/oci/layout"
+	_ "github.com/containers/image/openshift"
+	_ "github.com/containers/image/storage"
+	"github.com/containers/image/transports"
+	"github.com/containers/image/types"
+	"github.com/pkg/errors"
+)
+
+// ParseImageName converts a URL-like image name to a types.ImageReference.
+func ParseImageName(imgName string) (types.ImageReference, error) {
```
cyphar: IMO this should be done in `transports`. This package should just be doing dumb imports and not providing any other functionality -- though I understand why you'd want to do this (so you don't have an annoying `_` import).

transports/transports.go
```diff
@@ -2,52 +2,24 @@ package transports
...
-	"github.com/containers/image/docker/daemon"
-	ociLayout "github.com/containers/image/oci/layout"
-	"github.com/containers/image/openshift"
-	"github.com/containers/image/storage"
 	"github.com/containers/image/types"
-	"github.com/pkg/errors"
 )
 
 // KnownTransports is a registry of known ImageTransport instances.
 var KnownTransports map[string]types.ImageTransport
```
cyphar: IMO this should be hidden (with some sort of `Get` function) and have a lock around it internally. It isn't a good idea to expose a `map` globally.

transports/transports.go
```diff
@@ -2,52 +2,24 @@ package transports
...
-func ParseImageName(imgName string) (types.ImageReference, error) {
-	parts := strings.SplitN(imgName, ":", 2)
-	if len(parts) != 2 {
-		return nil, errors.Errorf(`Invalid image name "%s", expected colon-separated transport:reference`, imgName)
-	}
-	transport, ok := KnownTransports[parts[0]]
-	if !ok {
-		return nil, errors.Errorf(`Invalid image name "%s", unknown transport "%s"`, imgName, parts[0])
+// Register TODO(runcom)
+func Register(t types.ImageTransport) {
```
cyphar: As above, add a lock to this.

transports/transports.go
```diff
@@ -2,52 +2,24 @@ package transports
...
+	name := t.Name()
+	if _, ok := KnownTransports[name]; ok {
+		panic(fmt.Sprintf("Duplicate image transport name %s", name))
 	}
-	return transport.ParseReference(parts[1])
+	KnownTransports[name] = t
 }
```
cyphar: Can we add a `Deregister` function? Also it would be nice to have a simple `Get` function (as the above comment noted), so that we can hide `KnownTransports`?

transports/transports.go
```diff
@@ -2,52 +2,24 @@ package transports
...
-	"github.com/pkg/errors"
 )
 
 // KnownTransports is a registry of known ImageTransport instances.
 var KnownTransports map[string]types.ImageTransport
 
 func init() {
 	KnownTransports = make(map[string]types.ImageTransport)
-	// NOTE: Make sure docs/policy.json.md is updated when adding or updating
-	// a transport.
```
mtrmac: Please preserve this comment (in the import block of `alltransports.go` probably).

transports/alltransports/alltransports.go
mtrmac: The underlying idea is that the strings are exposed to the users, so they should _always_ mean the same thing; we should not have `skopeo` and `whateverothertool` use strings which _look_ the same but actually each accepting a different subset of transports. So, if any user of the library uses the strings in the UI and wants to use `ParseImageName`, they automatically get all transports: they can’t forget the `_` import, as you say, and they can’t even opt out — this automatically tries to do the right thing for users. I understand that this may feel annoyingly heavy-handed; is it actually problematic? I _think_ that on balance this does the right thing for users but I can of course be mistaken.

(More specialized users of `containers/image` can use `somepackage.Transport.ParseReference` to get a part of the parsing functionality, or `somepackage.NewReference` if working with the native values directly.)

transports/transports.go
mtrmac: (No strong opinion on this; conceptually this is a separate idea from this PR; why not I guess.)

transports/transports.go
mtrmac: When would it be useful to call `Deregister`? Lock or not, it is global state, and there would be a risk of one package deregistering a transport which a completely unrelated package has also imported and expects to be present.

transports/transports.go
runcom: yeah, I'll do that

transports/transports.go
cyphar: If someone wants to switch out transports -- or even for mocking? Maybe an atomic `Replace` or something?

runcom:
@mtrmac PTAL (test failure is #238)

signature/policy_config_test.go
```diff
@@ -269,9 +270,6 @@ func TestPolicyUnmarshalJSON(t *testing.T) {
 		func(v mSI) { v["transports"] = []string{} },
 		// "default" is an invalid PolicyRequirements
 		func(v mSI) { v["default"] = PolicyRequirements{} },
-		// A key in "transports" is an invalid transport name
-		func(v mSI) { x(v, "transports")["this is unknown"] = x(v, "transports")["docker"] },
-		func(v mSI) { x(v, "transports")[""] = x(v, "transports")["docker"] },
```
mtrmac: There should be a test for an unknown transport name (that it is not rejected).

signature/policy_config.go
```diff
@@ -174,7 +172,7 @@ func (m *policyTransportScopesWithTransport) UnmarshalJSON(data []byte) error {
 		if _, ok := tmpMap[key]; ok {
 			return nil
 		}
-		if key != "" {
+		if key != "" && m.transport != nil {
```
mtrmac: This needs a test.

signature/policy_config.go
```diff
@@ -122,10 +122,8 @@ func (m *policyTransportsMap) UnmarshalJSON(data []byte) error {
 	// So, use a temporary map of pointers-to-slices and convert.
 	tmpMap := map[string]*PolicyTransportScopes{}
 	if err := paranoidUnmarshalJSONObject(data, func(key string) interface{} {
-		transport, ok := transports.KnownTransports[key]
-		if !ok {
-			return nil
-		}
+		// transport can be nil
```
mtrmac: Also please update the comment at `policyTransportScopesWithTransport`: `while validating using a specific ImageTransport` **`if not nil`**

signature/policy_config_test.go
```diff
@@ -9,6 +9,7 @@ import (
 
 	"github.com/containers/image/directory"
 	"github.com/containers/image/docker"
+	_ "github.com/containers/image/openshift"
```
mtrmac: Add a comment explaining the need for this please.

mtrmac: https://github.com/containers/image/pull/216#discussion_r101483154

signature/policy_config.go
mtrmac: And another test: 

For `requirementsForImageRef`, add a test that when a transport is not in `KnownTransports` when parsing the policy, but it is in `KnownTransports` at the time of `requirementsForImageRef`, the correct set of scopes is used (more or less re-run the existing test, except creating the policy in a different way).

transports/transports.go
```diff
@@ -2,52 +2,50 @@ package transports
...
+	kt.mu.Lock()
+	name := t.Name()
+	if t := kt.transports[name]; t != nil {
+		panic(fmt.Sprintf("Duplicate image transport name %s", name))
 	}
-	transport, ok := KnownTransports[parts[0]]
-	if !ok {
-		return nil, errors.Errorf(`Invalid image name "%s", unknown transport "%s"`, imgName, parts[0])
+	kt.transports[name] = t
+	kt.mu.Unlock()
```
mtrmac: Shouldn’t this use `defer`? This way, if we `panic` above, the lock is permanently locked. Or perhaps it would be better to explicitly unlock it before that `panic`, randomly unlocking in the middle of some other code path could be risky… not that there are that many code paths here :)

transports/transports.go
```diff
@@ -2,52 +2,50 @@ package transports
...
+	"sync"
+
 	"github.com/containers/image/types"
-	"github.com/pkg/errors"
 )
 
-// KnownTransports is a registry of known ImageTransport instances.
-var KnownTransports map[string]types.ImageTransport
+// knownTransports is a registry of known ImageTransport instances.
+type knownTransports struct {
```
mtrmac: With this being a private variable of a private type, it’s really not obvious to me what defining the type and methods adds right now.

I guess it allows writing tests? Except that those tests don’t exist.

(But this does work, feel free to leave it as it is.)

signature/policy_config.go
runcom: done

signature/policy_config_test.go
runcom: done

signature/policy_config.go
runcom: isn't this https://github.com/containers/image/pull/216/files#diff-38040ba06b5a45464a80ca2fb88302c2R249 testing this code path?

mtrmac: That tests parsing the policy into a data structure, not using the data structure to look up the relevant rules to evaluate for a particular image (`requirementsForImageRef`). Right now, by inspection, it seems clear enough that `requirementsForImageRef` can correctly handle a transport which exists but is unregistered, or a transport which was not registered when parsing the policy but is registered afterwards—but the transport registry is a very non-obvious hidden state in the parsing/evaluation interaction, so it would be nice to have a test which ensures that the code keeps working in these corner cases.

runcom: added a test, hopefully done correctly (pretty late here now)

signature/policy_eval_test.go
```diff
@@ -103,6 +105,34 @@ func (ref pcImageReferenceMock) DeleteImage(ctx *types.SystemContext) error {
...
+	require.NoError(t, err)
+	ref, err := reference.ParseNormalizedNamed("registry.access.redhat.com/rhel7:latest")
+	require.NoError(t, err)
+	reqs := pc.requirementsForImageRef(pcImageReferenceMock{"docker", ref})
+	assert.True(t, &(reqs[0]) == &(pr[0]))
+	assert.True(t, len(reqs) == len(pr))
+
+	assert.Nil(t, transports.Get("docker"))
+	transports.Register(docker.Transport)
+	assert.NotNil(t, transports.Get("docker"))
```
mtrmac: Please move this cleanup into a `defer` block right after unregistering trhe transport; otherwise one of the `require`s aborting would leave `transports` in an unexpected state.

runcom:
@mtrmac fixed but your LGTM didn't work

mtrmac:
:+1: pending a successful test in updated skopeo.

runcom:
@mtrmac https://github.com/projectatomic/skopeo/pull/288 it's green over at skopeo :+1:

runcom:
LGTM

PR #159, opened 2016-11-08
[RFC Do not merge] Add copy.Options.LayerSubset.

mtrmac:
For now just to get an idea of what this would take…

FIXME: Should come with more warnings.

This is needed to allow the Python code in atomic to only download the subset of layers which is not already available locally, while still verifying signatures and validating layer digests (per discussion in
https://github.com/projectatomic/skopeo/pull/242 ).

PR #218, opened 2017-01-19
Pre-layer copy hook

erikh:
This is a proof-of-concept and still needs tests and other completeness changes
(like for things that are not docker). Please let me know if you'd like to
accept the feature so I know to continue work towards getting this fleshed out
in the other image systems, tests, etc.

erikh/box has a layer editor inside it that allows it to filter out layers as
necessary. This is made possible currently with some hand-rolled image code.
I'd love to move to containers/image but it has a very coarse copy
implementation.

This adds a simple hook that provides the source layer information to a hook in
the copier which returns bool on whether to proceed to continue to copy
(contrast with `filepath.Walk`).

Here is some example code that demonstrates how it can be used: in this case,
the `golang` image is pulled from the host docker, the first layer is clipped
off the top and pushed to a new image named `erikh/test`. The other layer
relationships have been removed and naturally the image is much smaller.

This has proven useful for building minimal images of containers in the past.

```golang
package main

import (
	"fmt"

	"github.com/containers/image/copy"
	"github.com/containers/image/docker/daemon"
	"github.com/containers/image/docker/reference"
)

func main() {
	ref, err := daemon.ParseReference("docker.io/library/golang:latest")
	if err != nil {
		panic(err)
	}

	img, err := ref.NewImage(nil)
	if err != nil {
		panic(err)
	}
	defer img.Close()

	tgtRef, _ := reference.ParseNamed("docker.io/erikh/test:latest")
	tgt, err := daemon.NewReference("", tgtRef)
	if err != nil {
		panic(err)
	}

	b, _, err := img.Manifest()
	if err != nil {
		panic(err)
	}
	fmt.Println(string(b))

	var i int
	err = copy.Image(nil, tgt, ref, &copy.Options{
		RemoveSignatures: true,
		LayerCopyHook: func(srcLayer types.BlobInfo) bool {
			i++
			return i < 2 // only the first layer
		},
	})

	if err != nil {
		panic(err)
	}
}
```

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -49,6 +49,7 @@ type imageCopier struct {
 	reportWriter      io.Writer
 	progressInterval  time.Duration
 	progress          chan types.ProgressProperties
+	layerEditor       func([]types.BlobInfo) []types.BlobInfo
 }
 
 // newDigestingReader returns an io.Reader implementation with contents of source, which will eventually return a non-EOF error
@@ -98,6 +99,7 @@ type Options struct {
 	DestinationCtx   *types.SystemContext
 	ProgressInterval time.Duration                 // time to wait between reports to signal the progress channel
 	Progress         chan types.ProgressProperties // Reported to when ProgressInterval has arrived for a single artifact+offset.
+	LayerEditor      func([]types.BlobInfo) []types.BlobInfo
 }
 
 // Image copies image from srcRef to destRef, using policyContext to validate source image admissibility.
@@ -187,6 +189,7 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 		reportWriter:      reportWriter,
 		progressInterval:  options.ProgressInterval,
 		progress:          options.Progress,
+		layerEditor:       options.LayerEditor,
 	}
 
 	if err := ic.copyLayers(); err != nil {
@@ -194,7 +197,7 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 	}
 
 	pendingImage := src
-	if !reflect.DeepEqual(manifestUpdates, types.ManifestUpdateOptions{InformationOnly: manifestUpdates.InformationOnly}) {
+	if !reflect.DeepEqual(ic.manifestUpdates, types.ManifestUpdateOptions{InformationOnly: ic.manifestUpdates.InformationOnly}) {
 		if !canModifyManifest {
 			return errors.Errorf("Internal error: copy needs an updated manifest but that was known to be forbidden")
 		}
@@ -236,11 +239,14 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 		return errors.Wrap(err, "Error writing manifest")
 	}
 
-	writeReport("Storing signatures\n")
-	if err := dest.PutSignatures(sigs); err != nil {
-		return errors.Wrap(err, "Error writing signatures")
-	}
+	/*
+		writeReport("Storing signatures\n")
+		if err := dest.PutSignatures(sigs); err != nil {
+			return errors.Wrap(err, "Error writing signatures")
+		}
+	*/
 
+	writeReport(fmt.Sprintf("Committing image %q\n", destRef.DockerReference()))
 	if err := dest.Commit(); err != nil {
 		return errors.Wrap(err, "Error committing the finished image")
 	}
@@ -250,8 +256,15 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 
 // copyLayers copies layers from src/rawSource to dest, using and updating ic.manifestUpdates if necessary and ic.canModifyManifest.
 func (ic *imageCopier) copyLayers() error {
+	origSrcInfos := ic.src.LayerInfos()
 	srcInfos := ic.src.LayerInfos()
+
+	if ic.layerEditor != nil {
+		srcInfos = ic.layerEditor(ic.src.LayerInfos())
+	}
+
 	destInfos := []types.BlobInfo{}
+
 	diffIDs := []digest.Digest{}
 	for _, srcLayer := range srcInfos {
 		var (
@@ -259,6 +272,7 @@ func (ic *imageCopier) copyLayers() error {
 			diffID   digest.Digest
 			err      error
 		)
+
 		if ic.dest.AcceptsForeignLayerURLs() && len(srcLayer.URLs) != 0 {
 			// DiffIDs are, currently, needed only when converting from schema1.
 			// In which case src.LayerInfos will not have URLs because schema1
@@ -277,11 +291,12 @@ func (ic *imageCopier) copyLayers() error {
 		destInfos = append(destInfos, destInfo)
 		diffIDs = append(diffIDs, diffID)
 	}
+
 	ic.manifestUpdates.InformationOnly.LayerInfos = destInfos
 	if ic.diffIDsAreNeeded {
 		ic.manifestUpdates.InformationOnly.LayerDiffIDs = diffIDs
 	}
-	if layerDigestsDiffer(srcInfos, destInfos) {
+	if layerDigestsDiffer(origSrcInfos, destInfos) {
 		ic.manifestUpdates.LayerInfos = destInfos
 	}
 	return nil
@@ -303,16 +318,19 @@ func layerDigestsDiffer(a, b []types.BlobInfo) bool {
 // copyConfig copies config.json, if any, from src to dest.
 func (ic *imageCopier) copyConfig(src types.Image) error {
 	srcInfo := src.ConfigInfo()
+
 	if srcInfo.Digest != "" {
 		fmt.Fprintf(ic.reportWriter, "Copying config %s\n", srcInfo.Digest)
 		configBlob, err := src.ConfigBlob()
 		if err != nil {
 			return errors.Wrapf(err, "Error reading config blob %s", srcInfo.Digest)
 		}
+
 		destInfo, err := ic.copyBlobFromStream(bytes.NewReader(configBlob), srcInfo, nil, false)
 		if err != nil {
 			return err
 		}
+
 		if destInfo.Digest != srcInfo.Digest {
 			return errors.Errorf("Internal error: copying uncompressed config blob %s changed digest to %s", srcInfo.Digest, destInfo.Digest)
 		}
--- a/docker/daemon/daemon_dest.go
+++ b/docker/daemon/daemon_dest.go
@@ -82,6 +82,8 @@ func imageLoadGoroutine(ctx context.Context, c *client.Client, reader *io.PipeRe
 		}
 	}()
 
+	// the err below will overwrite the error above; there by preventing the
+	// panic error message when the function can no longer panic.
 	resp, err := c.ImageLoad(ctx, reader, true)
 	if err != nil {
 		err = errors.Wrap(err, "Error saving image to docker engine")
@@ -180,8 +182,11 @@ func (d *daemonImageDestination) PutBlob(stream io.Reader, inputInfo types.BlobI
 	if err := d.sendFile(inputInfo.Digest.String(), inputInfo.Size, tee); err != nil {
 		return types.BlobInfo{}, err
 	}
-	d.blobs[inputInfo.Digest] = types.BlobInfo{Digest: digester.Digest(), Size: inputInfo.Size}
-	return types.BlobInfo{Digest: digester.Digest(), Size: inputInfo.Size}, nil
+
+	retval := types.BlobInfo{Digest: digester.Digest(), Size: inputInfo.Size}
+
+	d.blobs[inputInfo.Digest] = retval
+	return retval, nil
 }
 
 func (d *daemonImageDestination) HasBlob(info types.BlobInfo) (bool, int64, error) {
@@ -276,6 +281,7 @@ func (d *daemonImageDestination) sendFile(path string, expectedSize int64, strea
 	if err != nil {
 		return nil
 	}
+
 	logrus.Debugf("Sending as tar file %s", path)
 	if err := d.tar.WriteHeader(hdr); err != nil {
 		return err
@@ -312,6 +318,5 @@ func (d *daemonImageDestination) Commit() error {
 	d.committed = true // We may still fail, but we are done sending to imageLoadGoroutine.
 
 	logrus.Debugf("docker-daemon: Waiting for status")
-	err := <-d.statusChannel
-	return err
+	return <-d.statusChannel
 }
--- a/image/docker_schema2.go
+++ b/image/docker_schema2.go
@@ -98,6 +98,7 @@ func (m *manifestSchema2) ConfigBlob() ([]byte, error) {
 		if err != nil {
 			return nil, err
 		}
+
 		computedDigest := digest.FromBytes(blob)
 		if computedDigest != m.ConfigDescriptor.Digest {
 			return nil, errors.Errorf("Download config.json digest %s does not match expected %s", computedDigest, m.ConfigDescriptor.Digest)
@@ -152,9 +153,6 @@ func (m *manifestSchema2) UpdatedImageNeedsLayerDiffIDs(options types.ManifestUp
 func (m *manifestSchema2) UpdatedImage(options types.ManifestUpdateOptions) (types.Image, error) {
 	copy := *m // NOTE: This is not a deep copy, it still shares slices etc.
 	if options.LayerInfos != nil {
-		if len(copy.LayersDescriptors) != len(options.LayerInfos) {
-			return nil, errors.Errorf("Error preparing updated manifest: layer count changed from %d to %d", len(copy.LayersDescriptors), len(options.LayerInfos))
-		}
 		copy.LayersDescriptors = make([]descriptor, len(options.LayerInfos))
 		for i, info := range options.LayerInfos {
 			copy.LayersDescriptors[i].Digest = info.Digest
--- a/image/docker_schema2.go
+++ b/image/docker_schema2.go
@@ -5,6 +5,7 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"strings"
 
@@ -98,6 +99,8 @@ func (m *manifestSchema2) ConfigBlob() ([]byte, error) {
 		if err != nil {
 			return nil, err
 		}
+
+		fmt.Println(string(blob))
 
 		computedDigest := digest.FromBytes(blob)
 		if computedDigest != m.ConfigDescriptor.Digest {
--- a/image/docker_schema2_test.go
+++ b/image/docker_schema2_test.go
@@ -402,10 +402,6 @@ func TestManifestSchema2UpdatedImage(t *testing.T) {
 	})
 	require.NoError(t, err)
 	assert.Equal(t, layerInfos, res.LayerInfos())
-	_, err = original.UpdatedImage(types.ManifestUpdateOptions{
-		LayerInfos: append(layerInfos, layerInfos[0]),
-	})
-	assert.Error(t, err)
 
 	// ManifestMIMEType:
 	// Only smoke-test the valid conversions, detailed tests are below. (This also verifies that “original” is not affected.)
--- a/manifest/manifest.go
+++ b/manifest/manifest.go
@@ -54,7 +54,7 @@ func GuessMIMEType(manifest []byte) string {
 	}
 
 	switch meta.MediaType {
-	case DockerV2Schema2MediaType, DockerV2ListMediaType, imgspecv1.MediaTypeImageManifest, imgspecv1.MediaTypeImageManifestList: // A recognized type.
+	case DockerV2Schema2MediaType, DockerV2ListMediaType, imgspecv1.MediaTypeImageManifest: // A recognized type.
 		return meta.MediaType
 	}
 	// this is the only way the function can return DockerV2Schema1MediaType, and recognizing that is essential for stripping the JWS signatures = computing the correct manifest digest.
--- a/manifest/manifest_test.go
+++ b/manifest/manifest_test.go
@@ -22,7 +22,6 @@ func TestGuessMIMEType(t *testing.T) {
 		mimeType string
 	}{
 		{"ociv1.manifest.json", imgspecv1.MediaTypeImageManifest},
-		{"ociv1list.manifest.json", imgspecv1.MediaTypeImageManifestList},
 		{"v2s2.manifest.json", DockerV2Schema2MediaType},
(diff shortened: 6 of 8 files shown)


runcom:
I've not yet done a full code review but high level concept sounds really good to me. Maybe we can use the callback to also _skip_ layers we're not interested in (or the other way around). I like it. @mtrmac WDYT?

erikh:
https://erikh.github.io/box/user-guide/functions/#skip is a practical
application of this function and does exactly what you're describing.

copy/copy.go
```diff
@@ -122,10 +124,13 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
...
 	}()
 
-	// Please keep this policy check BEFORE reading any other information about the image.
-	if allowed, err := policyContext.IsRunningImageAllowed(unparsedImage); !allowed || err != nil { // Be paranoid and fail if either return value indicates so.
-		return errors.Wrap(err, "Source image rejected")
+	if policyContext != nil {
+		// Please keep this policy check BEFORE reading any other information about the image.
+		if allowed, err := policyContext.IsRunningImageAllowed(unparsedImage); !allowed || err != nil { // Be paranoid and fail if either return value indicates so.
+			return errors.Wrap(err, "Source image rejected")
+		}
```
mtrmac: I’d rather force the caller to provide a policy. It is… not _too_ difficult… to do:
```go
signature.Policy{Default: []signature.PolicyRequirement{signature.NewPRInsecureAcceptAnything()}}
```
and passing `nil` here by mistake could be disastrous.

copy/copy.go
```diff
@@ -155,7 +160,7 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 		}
 	}
 
-	canModifyManifest := len(sigs) == 0
+	canModifyManifest := len(sigs) == 0 || options.RemoveSignatures
```
mtrmac: Why is this necessary? Per
```go
var sigs [][]byte
if options != nil && options.RemoveSignatures {
    sigs = [][]byte{}
```
above, this seems redundant.

(Also, `options` can be `nil`.)

image/docker_schema2.go
```diff
@@ -150,15 +150,10 @@ func (m *manifestSchema2) UpdatedImageNeedsLayerDiffIDs(options types.ManifestUp
 // This does not change the state of the original Image object.
 func (m *manifestSchema2) UpdatedImage(options types.ManifestUpdateOptions) (types.Image, error) {
 	copy := *m // NOTE: This is not a deep copy, it still shares slices etc.
+
 	if options.LayerInfos != nil {
-		if len(copy.LayersDescriptors) != len(options.LayerInfos) {
```
mtrmac: If we do add support for this, it should land for all manifest schemas simultaneously.

image/docker_schema2.go
```diff
@@ -173,6 +168,50 @@ func (m *manifestSchema2) UpdatedImage(options types.ManifestUpdateOptions) (typ
 	return memoryImageFromManifest(&copy), nil
 }
 
+// Edits the layers in the manifest when called by replacing them with the
+// appropriate infos provided. This is metadata-only -- the actual layers still
+// have to get to the image.
+func (m *manifestSchema2) performEdits(infos []types.BlobInfo) error {
```
mtrmac: `performEdits` is too generic a name for this limited operation.

image/docker_schema2.go
```diff
@@ -173,6 +168,50 @@ func (m *manifestSchema2) UpdatedImage(options types.ManifestUpdateOptions) (typ
...
+	if err := json.Unmarshal(blob, imageConfig); err != nil {
+		return err
+	}
+
+	imageConfig.History = make([]imageHistory, infolen)
+	imageConfig.RootFS.DiffIDs = make([]digest.Digest, infolen)
+	m.LayersDescriptors = make([]descriptor, infolen)
+	for i, info := range infos {
+		imageConfig.History[i] = imageHistory{}
+		imageConfig.RootFS.DiffIDs[i] = info.Digest
```
mtrmac: This is incorrect if the layer is compressed. `DiffID` values must be the digests of the compressed layers.

image/docker_schema2.go
```diff
@@ -173,6 +168,50 @@ func (m *manifestSchema2) UpdatedImage(options types.ManifestUpdateOptions) (typ
...
+		m.LayersDescriptors[i].Digest = info.Digest
+		m.LayersDescriptors[i].Size = info.Size
+		m.LayersDescriptors[i].URLs = info.URLs
+	}
+	imageConfig.RootFS.BaseLayer = m.LayersDescriptors[0].Digest.String()
```
mtrmac: (Note to self: I didn’t review the `config.json` edits in detail here, … but this _is_ inconsistent with what `manifestSchema1.convertToManifestSchema2` does.)

copy/copy.go
```diff
@@ -92,6 +93,7 @@ type Options struct {
 	ReportWriter     io.Writer
 	SourceCtx        *types.SystemContext
 	DestinationCtx   *types.SystemContext
+	LayerCopyHook    func(types.BlobInfo) bool
```
mtrmac: This should really have a more descriptive name; `shouldLayerBeCopied` or something shorter to that effect; “hook” does not describe the function at all.

mtrmac: … and does this _really_ have to be a function? Could this be a set, as per #159 ?

Or, more generally, if we add “delete layer” here, pretty soon we will be adding an “add a layer at this index”. What would _that_ interface look like, and can we define that instead right now? (Yeah, that would be making this PR more complex. Perhaps we should just merge this, or the “layer subset” non-hook variant, right now, with a `// Warning: API likely to change`, and not block on the higher-level discussions?)

image/docker_schema2.go
```diff
@@ -173,6 +168,50 @@ func (m *manifestSchema2) UpdatedImage(options types.ManifestUpdateOptions) (typ
...
+// have to get to the image.
+func (m *manifestSchema2) performEdits(infos []types.BlobInfo) error {
+	infolen := len(infos)
+
+	if infolen < 1 {
+		return errors.New("cannot edit an image down to less than one layer")
+	}
+
+	// regurgitate the image configuration for recalculation of layers in the event the layer list has been edited.
+	imageConfig := &image{}
```
mtrmac: There is a fairly high risk that this would drop any fields of `config.json` added in the future.

image/docker_schema2.go
mtrmac: (See also the `UpdatedImageNeedsLayerDiffIDs` hack. That decision mechanism would obviously not work with the “anything can be returned” hook, and it is also too late by that time; so “should we compute DiffIDs” would need a complete rethinking.)

erikh:
The two security changes I made seemed necessary at the time but I will back them out.

erikh:
Ok. I will not close this while you decide. I realize the code is not in the best shape; it's merely meant as a discussion-starter.

For the rest of it, I'm focused on it being doable; we accomplish this in box with a layer list we manage outside of the layer list in the image (and then repackage at the end) so there's no reason we couldn't do it that way too -- I just felt a function was a safer bet given how generic they can be. No objection to doing it the other way.

mtrmac:
> Ok. I will not close this while you decide. I realize the code is not in the best shape; it's merely meant as a discussion-starter.

Yeah; this strongly depends on how you want to use it in general.

Vaguely, the two major directions I see (but I may well be missing something)

 * Expose enough of `copyLayers` that a caller who _exactly_ knows what it needs, and doesn’t care about general API availability (e.g. is restricted to schema2, only uncompressed layers, and the like) can use the copy/compress/verify-digest/future-parallel-copy functionality, and manually do the high-level image manipulation (copying/editing configs, manifests), using some but perhaps not all of the `types.Image` functions. Negative: Need to know much more about the image and the formats. Positive: almost infinite flexibility for editing the image.

* Add an `options.LayerEditor` hook, which would receive `ic.src.LayerInfos()` and return an edited value, with some layers deleted, and some added (in which case, probably also supplying an `io.Reader` for the contents of that layer).:
  ```go
  LayerEditor func([]types.BlobInfo) []struct{info types.BlobInfo; sourceIfNewLayer io.Reader}`
  ```
  Then, `copy.Image` would be on the hook for making these edits work. Negative: More work to support the various edits in containers/image (or a confusing API where some combinations work and some don’t); _only_ those kinds of edits would be possible; Positive: Much easier to use for callers who want to do a single well-targeted edit to the layers only.

Does any of this make sense? Are there other options I have missed? @runcom ?

(And would the `LayerSubset` parameter instead of a hook from #159 work for you, or is there something about the hook which is more useful?)

mtrmac:
(admittedly the `LayerEditor` name doesn’t mean anything and in general sucks.)

erikh:
Shooting from the hip; the presence of the hook during the run should be
enough to describe a capability to the appropriate image destination that
it can reject if it discovers the hook and can't process it. Requires
knowledge though, and will need to be maintained as destinations grow as a
canonical part of the API.

I don't think doing this just in schema2 is the right answer, if that's
what you were suggesting.

I need to mull over the rest. Will reply in a few hours.

mtrmac:
> I don't think doing this just in schema2 is the right answer, if that's what you were suggesting.

To be fair, my initial comment on the `docker_schema2.go` file was probably too strong. It would definitely not be the first case where some `types.Image.UpdatedImage` functionality is only available for some schemas or schema combinations, and if nobody cared to implement this for $obscure_format, well, nobody would exactly be _hurt_. But it does make the interface confusing when we can’t easily explain to users what is and is not supposed to work.

erikh:
agree entirely.

copy/copy.go
erikh: we pass the layer id so really you could do anything you want, including keeping a table of images you want to keep, which is exactly what box does.

erikh:
Hmm; maybe make the hook a part of the interface somehow, so it has to at least be considered when implementing a destination?

erikh:
Also please note that I have elided the security changes, so we can focus on the meat. :)

mtrmac:
> Hmm; maybe make the hook a part of the interface somehow, so it has to at least be considered when implementing a destination?

As far as layers go, it seems to me that `ManifestUpdateOptions.LayerInfos` along with the data in `.InformationOnly` should allow _any_ edits (additions, deletions, reorders). It would be up to `copy.Image` / `copyLayers` along with any hooks / edit parameters to supply the necessary info for the `ManifestUpdateOptions`. Am I missing anything?

erikh:
Sure; I'm talking about the requirement for the hook to be implemented (or at least, considered when implementing a destination).

Unrelated, is there an irc or slack I can join to collaborate with you folks?

erikh:
I'm sorry, I'm not being clear.

For a destination to work with copy, I think it should at least implement or at minimum, do something (perhaps by return an error) to indicate that it cannot do layer edits.

This could be done by satisfying an interface.

I hope that's clearer, sorry.

erikh:
@mtrmac I'm going to try implementing option #2 in https://github.com/containers/image/pull/218#issuecomment-273879612 tonight -- please let me know if you'd prefer I do it some other way.

I am already running with a fork with this PR in erikh/box so I could get moving, but I want to realign as soon as we can come to a conclusion of how to implement this.

erikh:
... the layer editing hook, to be precise.

mtrmac:
> @mtrmac I'm going to try implementing option #2 in #218 (comment) tonight -- please let me know if you'd prefer I do it some other way.

No, I continue to think that an interface vaguely like that would be general and flexible enough to be useful long-term.

And it is quite acceptable to implement this kind of editing only for one of the image manifest formats in the initial PR, as long as an attempt to do it with the unsupported formats is clearly rejected.

erikh:
Well ideally I'd like to set it up in a way that all (or at least, almost
all) editing capabilities are added to image formats across the library; it
wasn't my intent just to affect docker images this way. I need this for OCI
and friends too.

erikh:
I'm going to leave the new layer reader off for now. I really think this would be better done as a separate method on the src that accepts a reader for each layer.

The rest I'm working on now.

erikh:
I've taken a stab at this (which I have overwritten the previous patch with)
but the crux of the problem nwo seems to be that the config is copied verbatim
(see ConfigBlob() in manifestSchema2 and copyConfig in copy.Image) and editing
it requires knowledge of the manifest format, which is hard to reap in a
portable way here.

Looking for some suggestions on how to resolve this.

copy/copy.go
```diff
@@ -223,11 +226,14 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
...
-	writeReport("Storing signatures\n")
-	if err := dest.PutSignatures(sigs); err != nil {
-		return errors.Wrap(err, "Error writing signatures")
-	}
+	/*
+		writeReport("Storing signatures\n")
+		if err := dest.PutSignatures(sigs); err != nil {
+			return errors.Wrap(err, "Error writing signatures")
+		}
+	*/
```
mtrmac: Why?

copy/copy.go
```diff
@@ -237,15 +243,23 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 
 // copyLayers copies layers from src/rawSource to dest, using and updating ic.manifestUpdates if necessary and ic.canModifyManifest.
 func (ic *imageCopier) copyLayers() error {
+	origSrcInfos := ic.src.LayerInfos()
 	srcInfos := ic.src.LayerInfos()
+
+	if ic.layerEditor != nil {
+		srcInfos = ic.layerEditor(ic.src.LayerInfos())
```
mtrmac: Calling LayerInfos three times is a bit much… at the very least `srcInfos = ic.layerEditor(srcInfos)`

docker/daemon/daemon_dest.go
```diff
@@ -182,8 +184,11 @@ func (d *daemonImageDestination) PutBlob(stream io.Reader, inputInfo types.BlobI
 	if err := d.sendFile(inputInfo.Digest.String(), inputInfo.Size, tee); err != nil {
 		return types.BlobInfo{}, err
 	}
-	d.blobs[inputInfo.Digest] = types.BlobInfo{Digest: digester.Digest(), Size: inputInfo.Size}
-	return types.BlobInfo{Digest: digester.Digest(), Size: inputInfo.Size}, nil
+
+	retval := types.BlobInfo{Digest: digester.Digest(), Size: inputInfo.Size}
+
+	d.blobs[inputInfo.Digest] = retval
+	return retval, nil
```
mtrmac: (WRT this, comment changes, empty lines, etc…) I guess, why not, but please, in the final version, one idea per commit. It should be clear why every change was introduced, and in big commits it is often difficult whether a small change is essential for the primary task of the commit or really a typo in an unrelated cleanup which was not supposed to change behavior.

mtrmac: … OTOH, in the extreme, having 10 one-line commits is just fine with me. Obviously correct, easy to review.

mtrmac:
> but the crux of the problem nwo seems to be that the config is copied verbatim
(see ConfigBlob() in manifestSchema2 and copyConfig in copy.Image) and editing
it requires knowledge of the manifest format, which is hard to reap in a
portable way here.

Yeah.  Still doing that is really necessary.

A possible step towards making this possible could be to replace the existing `ManifestUpdateOptions.LayerInfos` with something like `LayerInfosPreservingDiffIDs`, with the existing functionality (editing the manifest only, not the config; we need to preserve this for all currently supported formats to be able compress blobs on upload), and a new `LayerInfosReplacement` (or a better name?), which requires a config update, and initially fail for all formats.  The caller would presumably be allowed to set only one of the two. (Or perhaps `ManifestUpdateOptions.LayerInfos` just needs to be richer so that it can capture the necessary information in both of these cases.)

And then, we do need to figure out what the required config updates are.  _Right now_, when the layers must come from the original `src`, the edits can only be a (possibly permuted) subset of the original layers; so we might be able to build the config updates by creating an equivalent (possibly permuted) subset of the original config entries. That seems fairly easy, or I may well be missing something essential about the config format.

The more general case, with arbitrary new layers being introduced, is noticeably harder; luckily we _do_ have the `DiffID` computation code already, so it would “only” have to be enabled if necessary, but the hook might still want to provide the `Created`/`Author`/`CreatedBy` fields (from schema2 history). Or perhaps nobody cares about this metadata?

copy/copy.go
erikh: no idea what happened here. I'll investigate.

erikh:
The problem is not editing the config, it's converting the config to the new layers; which means it has to be unmarshalled first and its properties changed and remarshalled, but to what format? At that level there's no indicator of what format to use, much less convert to.

mtrmac:
> At that level there's no indicator of what format to use, much less convert to.

I guess I don’t understand the problem; is there any ambiguity about config formats? We are already parsing / creating `config.json` files in `manifestSchema[12].convertToManifestSchema[12]`. It’s somewhat hairy but it seems perfectly doable.

Or are you perhaps saying that a schema2 manifest could refer to an OCI config, or an OCi manifest could refer to a schema1 config?

erikh:
Yes but the abstraction doesn't support that, unless I'm missing something,
by the time you're in ConfigBlob() you don't have access to the manifest
format to unmarshal/marshal it. It's a straight byte read.

erikh:
https://github.com/containers/image/blob/master/image/docker_schema2.go#L83
for reference

mtrmac:
> Yes but the abstraction doesn't support that, unless I'm missing something, by the time you're in ConfigBlob() you don't have access to the manifest format to unmarshal/marshal it.

What happens is that the copy code, after supplying all needed data, calls `.UpdatedImage`; that creates an in-memory-only variant of the original manifest (a `memoryImage` with a new manifest object); that new manifest object is initialized with an edited version of `configBlob` from the start. Follow e.g. https://github.com/containers/image/blob/master/image/docker_schema1.go#L171 → https://github.com/containers/image/blob/master/image/docker_schema1.go#L284 → https://github.com/containers/image/blob/master/image/docker_schema2.go#L56 .

(And of course the abstraction is not set in stone,  `image.genericManifest` is a private type perfectly subject to change, and even the public types in `types.go` have been modified fairly frequently in the past. The existing structure does seem to be more or less suitable for this in principle, concentrating all the true complexity in `.UpdatedImage` and its helper methods, and leaving the rest as fairly simple parsers / getters. It can very likely be improved.)

erikh:
Right, but if you read that code, it takes it right out of the original
image. Or does the getblob in the code I linked from above take from the
memory image?

erikh:
I should admit that I have tested this and the manifest is not updated.

mtrmac:
> Right, but if you read that code, it takes it right out of the original image

What path exactly? In the conversion case I have demonstrated above, https://github.com/containers/image/blob/master/image/docker_schema2.go#L84 uses the `m.configBlob != nil` path, with `m.configBlob` prepopulated with a memory-only version which does not exist in the source (obviously, when the source is not a schema2 image and there is no separate `config.json` in the original at all).

erikh:
I can check again, but I've printed that config right at that point. It's
not updated and gets processed right before uploading to docker. The layers
after editing do not save to docker because of this. This is literally the
only thing keeping it from working.

I'll amend the PR.

erikh:
I've amended the PR with the print statement and a test.go which will
exhibit the issue in the root directory of the repository. Can you validate
that I am seeing what I am seeing?

mtrmac:
That added `fmt.Println` is in the wrong place; again, in the in-memory case, `m.configBlob` is **non-`nil` from the start** and that is never reached.

If that’s not the case, I’m not immediately sure what is going on. Here’s what I have used to test that I am not going crazy:
```diff
diff --git a/vendor/github.com/containers/image/copy/copy.go b/vendor/github.com/containers/image/copy/copy.go
index d27e634..a0f3f8a 100644
--- a/vendor/github.com/containers/image/copy/copy.go
+++ b/vendor/github.com/containers/image/copy/copy.go
@@ -180,7 +180,7 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
        }
 
        pendingImage := src
-       if !reflect.DeepEqual(manifestUpdates, types.ManifestUpdateOptions{InformationOnly: manifestUpdates.InformationOnly}) {
+       if _ = reflect.DeepEqual; true { //} !reflect.DeepEqual(manifestUpdates, types.ManifestUpdateOptions{InformationOnly: manifestUpdates.InformationOnly}) {
                if !canModifyManifest {
                        return errors.Errorf("Internal error: copy needs an updated manifest but that was known to be forbidden")
                }
diff --git a/vendor/github.com/containers/image/image/docker_schema2.go b/vendor/github.com/containers/image/image/docker_schema2.go
index 76ceea7..1188cd9 100644
--- a/vendor/github.com/containers/image/image/docker_schema2.go
+++ b/vendor/github.com/containers/image/image/docker_schema2.go
@@ -151,6 +151,24 @@ func (m *manifestSchema2) UpdatedImageNeedsLayerDiffIDs(options types.ManifestUp
 // This does not change the state of the original Image object.
 func (m *manifestSchema2) UpdatedImage(options types.ManifestUpdateOptions) (types.Image, error) {
        copy := *m // NOTE: This is not a deep copy, it still shares slices etc.
+
+       if true {
+               logrus.Errorf("HERE")
+               blob, err := copy.ConfigBlob()
+               if err != nil {
+                       return nil, err
+               }
+               configJSON := bytes.Join([][]byte{[]byte("INVALIDPREFIX"), blob}, nil)
+               configDescriptor := descriptor{
+                       MediaType: "application/vnd.docker.container.image.v1+json",
+                       Size:      int64(len(configJSON)),
+                       Digest:    digest.FromBytes(configJSON),
+               }
+
+               m2 := manifestSchema2FromComponents(configDescriptor, nil, configJSON, copy.LayersDescriptors)
+               return memoryImageFromManifest(m2), nil
+       }
+
        if options.LayerInfos != nil {
                if len(copy.LayersDescriptors) != len(options.LayerInfos) {
                        return nil, errors.Errorf("Error preparing updated manifest: layer count changed from %d to %d", len(copy.LayersDescriptors), len(options.LayerInfos))
```

and `./skopeo  --policy=default-policy.json copy docker://busybox:latest dir:t`.

My best guess (but _just_ a guess) is that the `reflect.DeepEqual` test (which the above stupid hack patches out) is preventing the run of `UpdatedImage` at all.

mtrmac:
… actually, let’s step back: This PR, as far as manifest/config editing goes, _only_ updates `ManifestUpdateOptions.LayerInfos`. That of course does not edit `config.json` simply because that code doesn’t exist. `.LayerInfos` has been created to update the manifest after blobs are compressed (changing their digests but not DiffIDs). Editing the config as necessary would be a new functionality which needs to be written. It isn’t _already there_ just to enable.

erikh:
run the code! It's reached, trust me.

rhatdan:
@erikh Is this still something you want.  Could you rebase the PR and @mtrmac PTAL/

erikh:
it's ok, thanks and sorry to keep this hanging up things for so long

Issue #105, opened 2016-09-26
[performance] parallel layers upload/download in copy


runcom:
EDIT: This can be probably done in skopeo

mtrmac:
How, if skopeo only passes source and destination references to `copy.Image`?

runcom:
@mtrmac alright, then this should be implemented at copy level right

rhatdan:
@runcom @mtrmac @vrothberg I think we have this now?  Can we close this?

vrothberg:
Yup, that's done.

PR #219, opened 2017-01-19
Provide large file progress reporting hooks to copy operations

erikh:
Again, including the other commits for ease of getting things to compile; there
is an example at the end. Only the last commit is relevant in this PR.

This provides a hook and an interval timer for progressed copy operations, such
as image packaging. It only covers the docker endpoints for now and is
untested, again, as a proof of concept for approval.

Users can specify a function which is called at each interval as long as the
copy is progressing. This function is passed the current byte offset of the
copy operation. They can use this to calculate sizes, or percentages, or the
phase of the moon at the time the current byte is downloading. Resolution of
the timer is completely independent.

Here's some code that works with this: it reports byte offsets every 100ms
during both download and upload operations during this layer edit.

Thank you again for the consideration.

```go
package main

import (
	"fmt"
	"time"

	"github.com/containers/image/copy"
	"github.com/containers/image/docker/daemon"
	"github.com/containers/image/docker/reference"
)

func main() {
	ref, err := daemon.ParseReference("docker.io/library/golang:latest")
	if err != nil {
		panic(err)
	}

	copyCtx := &types.SystemContext{
		CopyHook: func(offset uint64) {
			fmt.Println(offset)
		},
		CopyHookInterval: 100 * time.Millisecond,
	}

	img, err := ref.NewImage(copyCtx)
	if err != nil {
		panic(err)
	}
	defer img.Close()

	tgtRef, _ := reference.ParseNamed("docker.io/erikh/test:latest")
	tgt, err := daemon.NewReference("", tgtRef)
	if err != nil {
		panic(err)
	}

	b, _, err := img.Manifest()
	if err != nil {
		panic(err)
	}
	fmt.Println(string(b))

	var i int
	err = copy.Image(nil, tgt, ref, &copy.Options{
		SourceCtx:        copyCtx,
		DestinationCtx:   copyCtx,
		RemoveSignatures: true,
		LayerCopyHook: func(srcLayer types.BlobInfo) bool {
			i++
			return i < 2 // only the first layer
		},
	})

	if err != nil {
		panic(err)
	}
}
```

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"io/ioutil"
 	"reflect"
+	"time"
 
 	pb "gopkg.in/cheggaaa/pb.v1"
 
@@ -46,6 +47,8 @@ type imageCopier struct {
 	diffIDsAreNeeded  bool
 	canModifyManifest bool
 	reportWriter      io.Writer
+	progressInterval  time.Duration
+	progress          chan types.ProgressProperties
 }
 
 // newDigestingReader returns an io.Reader implementation with contents of source, which will eventually return a non-EOF error
@@ -93,14 +96,22 @@ type Options struct {
 	ReportWriter     io.Writer
 	SourceCtx        *types.SystemContext
 	DestinationCtx   *types.SystemContext
+	ProgressInterval time.Duration                 // time to wait between reports to signal the progress channel
+	Progress         chan types.ProgressProperties // Reported to when ProgressInterval has arrived for a single artifact+offset.
 }
 
 // Image copies image from srcRef to destRef, using policyContext to validate source image admissibility.
 func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageReference, options *Options) error {
+	if options == nil {
+		options = &Options{}
+	}
+
 	reportWriter := ioutil.Discard
-	if options != nil && options.ReportWriter != nil {
+
+	if options.ReportWriter != nil {
 		reportWriter = options.ReportWriter
 	}
+
 	writeReport := func(f string, a ...interface{}) {
 		fmt.Fprintf(reportWriter, f, a...)
 	}
@@ -139,7 +150,7 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 	}
 
 	var sigs [][]byte
-	if options != nil && options.RemoveSignatures {
+	if options.RemoveSignatures {
 		sigs = [][]byte{}
 	} else {
 		writeReport("Getting image source signatures\n")
@@ -174,6 +185,8 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 		diffIDsAreNeeded:  src.UpdatedImageNeedsLayerDiffIDs(manifestUpdates),
 		canModifyManifest: canModifyManifest,
 		reportWriter:      reportWriter,
+		progressInterval:  options.ProgressInterval,
+		progress:          options.Progress,
 	}
 
 	if err := ic.copyLayers(); err != nil {
@@ -200,7 +213,7 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 		return err
 	}
 
-	if options != nil && options.SignBy != "" {
+	if options.SignBy != "" {
 		mech, err := signature.NewGPGSigningMechanism()
 		if err != nil {
 			return errors.Wrap(err, "Error initializing GPG")
@@ -488,6 +501,17 @@ func (ic *imageCopier) copyBlobFromStream(srcStream io.Reader, srcInfo types.Blo
 		destStream = pipeReader
 		inputInfo.Digest = ""
 		inputInfo.Size = -1
+	}
+
+	// === Report progress using the ic.progress channel, if required.
+	if ic.progress != nil && ic.progressInterval > 0 {
+		destStream = &progressReader{
+			source:   destStream,
+			channel:  ic.progress,
+			interval: ic.progressInterval,
+			artifact: srcInfo,
+			lastTime: time.Now(),
+		}
 	}
 
 	// === Finally, send the layer stream to dest.
--- a/copy/progress_reader.go
+++ b/copy/progress_reader.go
@@ -1,0 +1,28 @@
+package copy
+
+import (
+	"io"
+	"time"
+
+	"github.com/containers/image/types"
+)
+
+// progressReader is a reader that reports its progress on an interval.
+type progressReader struct {
+	source   io.Reader
+	channel  chan types.ProgressProperties
+	interval time.Duration
+	artifact types.BlobInfo
+	lastTime time.Time
+	offset   uint64
+}
+
+func (r *progressReader) Read(p []byte) (int, error) {
+	n, err := r.source.Read(p)
+	r.offset += uint64(n)
+	if time.Since(r.lastTime) > r.interval {
+		r.channel <- types.ProgressProperties{Artifact: r.artifact, Offset: r.offset}
+		r.lastTime = time.Now()
+	}
+	return n, err
+}
--- a/types/types.go
+++ b/types/types.go
@@ -294,6 +294,13 @@ type SystemContext struct {
 	DockerDisableV1Ping bool
 }
 
+// ProgressProperties is used to pass information from the copy code to a monitor which
+// can use the real-time information to produce output or react to changes.
+type ProgressProperties struct {
+	Artifact BlobInfo
+	Offset   uint64
+}
+
 var (
 	// ErrBlobNotFound can be returned by an ImageDestination's HasBlob() method
 	ErrBlobNotFound = errors.New("no such blob present")


docker/daemon/daemon_dest.go
```diff
@@ -68,25 +69,38 @@ func newImageDestination(systemCtx *types.SystemContext, ref daemonReference) (t
...
 			reader.Close()
+			ourReader.Close()
 		} else {
 			reader.CloseWithError(err)
+			ourReader.CloseWithError(err)
 		}
 	}()
 
-	resp, err := c.ImageLoad(ctx, reader, true)
+	if sysctx.CopyHook != nil && sysctx.CopyHookInterval > 0 {
```
mtrmac: `sysctx` is allowed to be `nil`

docker/daemon/daemon_src.go
```diff
@@ -73,8 +74,14 @@ func newImageSource(ctx *types.SystemContext, ref daemonReference) (types.ImageS
 		}
 	}()
 
-	if _, err := io.Copy(tarCopyFile, inputStream); err != nil {
-		return nil, err
+	if ctx == nil {
```
mtrmac: `ctx != nil && ctx.CopyHookInterval > 0 && …` ?

erikh:
Yeah this could have been an oversight on my part, I was just adding things
I needed until I could start porting. :)

I'll review this and close tonight if so.

erikh:
I've rebased this to the point tests should pass; I will leave it alone until we decide what to do. In the meantime, I'm going to port box to use containers/image so I can get a better feel for what I actually need, but I'd like to keep my progress meters :)

erikh:
I have made an attempt to pretty up the API; let me know what you think. This
supplies the copy hook at copy.Options time (for both source and destination
reporting) and in the systemcontext still (e.g., if you only want to display a
meter for a single direction, which we do in box).

mtrmac:
- This still forces every single `ImageSource` and `ImageDestination` to implement progress reporting separately; do you really need that, or do you intend to use them through `copy.Image`? If the latter, a hook in `copyBlobFromStream` should suffice.
 - Is it really necessary to have a separate progress report for the source and the destination? `copyBlobFromStream` connects the two using a series of pipes etc., which may do some compression/decompression but should generally be much faster than downloads over the network, so I’d expect progress reporting on the source and destination to report very similar percentages (though perhaps different raw values if one stream is compressed and the other uncompressed)
 - As it is, `CopyHook` has no information at all about which stream it is being called for, it just receives a successive series of completely context-less integers which have no obvious relation to each other. (This is already true when hooking both the source and destination, and it will get much worse if #105 ever happens). Hence my question to @runcom about existence of a commonly accepted interface which does provide sufficient information and context. E.g., I _guess_ (without knowing the field) that `docker/docker/pkg/progress` might be an option—it is still a series of unconnected data points, with the recipient having to sort them out, but at least they have a disambiguating `ID` for each of the parallel operations.

erikh:
> - This still forces every single `ImageSource` and `ImageDestination` to
> implement progress reporting separately; do you really need that, or do you
> intend to use them through `copy.Image`? If the latter, a hook in
> `copyBlobFromStream` should suffice.

Not exactly; if you pass the copyfunc into the options it is directly added to
each context so that you don't have to do it yourself, OR you can do the
contexts to scope it.

> - Is it really necessary to have a separate progress report for the source
> and the destination? `copyBlobFromStream` connects the two using a series of
> pipes etc., which may do some compression/decompression but should generally
> be much faster than downloads over the network, so I’d expect progress
> reporting on the source and destination to report very similar percentages
> (though perhaps different raw values if one stream is compressed and the
> other uncompressed)

This is for flexibility, it can be useful when doing it unidirectionally.
Try the latest version of the source (below) to see what I mean. The first pass at
`ref.NewImage` implements a source copy (download) hook, while the others
process the image copy.

```go

package main

import (
	"fmt"
	"time"

	"github.com/containers/image/copy"
	"github.com/containers/image/docker/daemon"
	"github.com/containers/image/docker/reference"
	"github.com/containers/image/signature"
)

func main() {
	ref, err := daemon.ParseReference("docker.io/library/golang:latest")
	if err != nil {
		panic(err)
	}

  // you can do it this way too

	img, err := ref.NewImage(&types.SystemContext{
  })
			CopyHookInterval: 100 * time.Millisecond,
			CopyHook: func(offset uint64) {
				fmt.Println(offset)
			},
	if err != nil {
		panic(err)
	}
	defer img.Close()

	tgtRef, _ := reference.ParseNamed("docker.io/erikh/test:latest")
	tgt, err := daemon.NewReference("", tgtRef)
	if err != nil {
		panic(err)
	}

	b, _, err := img.Manifest()
	if err != nil {
		panic(err)
	}
	fmt.Println(string(b))

	pc, err := signature.NewPolicyContext(&signature.Policy{
		Default: []signature.PolicyRequirement{signature.NewPRInsecureAcceptAnything()},
	})

	if err != nil {
		panic(err)
	}

	err = copy.Image(
		pc,
		tgt,
		ref,
		&copy.Options{
			CopyHookInterval: 100 * time.Millisecond,
			CopyHook: func(offset uint64) {
				fmt.Println(offset)
			},
			RemoveSignatures: true,
		})

	if err != nil {
		panic(err)
	}
}

> - As it is, `CopyHook` has no information at all about which stream it is
> being called for, it just receives a successive series of completely
> context-less integers which have no obvious relation to each other. (This is
> already true when hooking both the source and destination, and it will get
> much worse if #105 ever happens). Hence my question to @runcom about
> existence of a commonly accepted interface which does provide sufficient
> information and context. E.g., I _guess_ (without knowing the field) that
> `docker/docker/pkg/progress` might be an option—it is still a series of
> unconnected data points, with the recipient having to sort them out, but at
> least they have a disambiguating `ID` for each of the parallel operations.

Perhaps I need to comment the source; the successive series is a increasing
value based on the progress of the download. It has context, and can be
leveraged with closures to get flexibility while processing it.

I really would prefer a hook if possible. Dealing with docker's progress meters
is pretty terrible and I have a nice, colorized/tty-aware UI with clear progress
information that only triggers when necessary (e.g, for copying large files and
layers only). I really don't want to ditch that. Basically, I'd really like to
avoid using a canned progress meter. It's really not the responsibility of this
library to implement progress meters if you ask me, but I realize it's not my
say too. :)

**Comment by runcom** [comment]
> > - As it is, `CopyHook` has no information at all about which stream it is
> > being called for, it just receives a successive series of completely
> > context-less integers which have no obvious relation to each other. (This is
> > already true when hooking both the source and destination, and it will get
> > much worse if #105 ever happens). Hence my question to @runcom about
> > existence of a commonly accepted interface which does provide sufficient
> > information and context. E.g., I _guess_ (without knowing the field) that
> > `docker/docker/pkg/progress` might be an option—it is still a series of
> > unconnected data points, with the recipient having to sort them out, but at
> > least they have a disambiguating `ID` for each of the parallel operations.
> 
> Perhaps I need to comment the source; the successive series is a increasing
> value based on the progress of the download. It has context, and can be
> leveraged with closures to get flexibility while processing it.
> 
> I really would prefer a hook if possible. Dealing with docker's progress meters
> is pretty terrible and I have a nice, colorized/tty-aware UI with clear progress
> information that only triggers when necessary (e.g, for copying large files and
> layers only). I really don't want to ditch that. Basically, I'd really like to
> avoid using a canned progress meter. It's really not the responsibility of this
> library to implement progress meters if you ask me, but I realize it's not my
> say too. :)

I agree the docker progress package isn't exactly what we're looking for there (and I'm not in favor of it here!).
This copy function hook provides a nice abstraction to me though, even given @mtrmac concerns in the first comment. We could find a way to still have this closure passed down addressing @mtrmac concerns.

**Comment by erikh** [comment]
I had another idea: what if we pass a send-only channel that passes a struct,
which gives us the flexibility of modifying the payload later.

e.g., the struct could have the byte offset, but also the percentage if know it.

The channel is optional. They can do whatever the heck they want with it then. :)

**Comment by mtrmac** [comment]
> > - This still forces every single `ImageSource` and `ImageDestination` to
> > implement progress reporting separately; do you really need that, or do you
> > intend to use them through `copy.Image`? If the latter, a hook in
> > `copyBlobFromStream` should suffice.
> 
> Not exactly; if you pass the copyfunc into the options it is directly added to
> each context so that you don't have to do it yourself, OR you can do the
> contexts to scope it.

That’s not what I mean; what you say allows the _caller_ to set a single hook, but the _implementations_ are not shared. We have 6 transports, * 2 for source/destination, that’s 12 places which would need to now care about progress reporting, and 12 places which would have to be kept in sync. Whereas a progress reporting in `copyBlobFromStream` or so would be a single piece of code to maintain (which also means that it could, without 12 times the pain, be a bit more complex).

> > - Is it really necessary to have a separate progress report for the source
> > and the destination? `copyBlobFromStream` connects the two using a series of
> > pipes etc., which may do some compression/decompression but should generally
> > be much faster than downloads over the network, so I’d expect progress
> > reporting on the source and destination to report very similar percentages
> > (though perhaps different raw values if one stream is compressed and the
> > other uncompressed)
> 
> This is for flexibility, it can be useful when doing it unidirectionally.
> Try the latest version of the source (below) to see what I mean.

I can’t see that this code demonstrates anything; it doesn’t treat sources/destinations differently at all if I am not mistaken (though it also appears somehow truncated?!).

Sure, a hook in `copyBlobFromStream` would be useless for users who deal with `ImageSource`/`ImageDestination` directly, without using `copy.Image`. I’m leaning towards saying that such callers are dealing with deep enough internals that they can bother to do a tiny bit of extra work themselves. (Besides, a shared hook like your `streamcopy.WithProgress`, or more likely an `io.Reader` wrapper like `pb.ProgressBar.NewProxyReader`, is easy enough to hook up if the caller were dealing with a single stream at a time.)

> > - As it is, `CopyHook` has no information at all about which stream it is
> > being called for, it just receives a successive series of completely
> > context-less integers which have no obvious relation to each other. (This is
> > already true when hooking both the source and destination, and it will get
> > much worse if #105 ever happens).
> 
> Perhaps I need to comment the source; the successive series is a increasing
> value based on the progress of the download.

But _which_ download? With the existing code, copying a 3-layer image with layers A, B, C, the hook could very well receive
`1, 100, 1024, 0, 50, 500, 2, 200, 2048, 50, 100, 1000, …` (the values being progress of, in order, writing A, reading A, reading B, writing B, writing C, reading C, writing A, reading A, reading B, writing B, writing C, reading C, …), and the hook has no idea which values belongs to which stream.

> Basically, I'd really like to
> avoid using a canned progress meter. It's really not the responsibility of this
> library to implement progress meters if you ask me, but I realize it's not my
> say too. :)

Sure, `copy.Options` should receive an interface which implements the actual data collection, or drawing if any. So, it would be a `docker/docker/pkg/progress.Output`, not directly `docker/docker/pkg/jsomessage` or whatever draws the progress bars. (Not saying that _that_ is necessarily a good interface. Just that this might be a problem which already has a nice solution, both as a data type and as a UI drawing consumer)

**Comment by mtrmac** [comment]
> I had another idea: what if we pass a send-only channel that passes a struct,
> which gives us the flexibility of modifying the payload later.
> 
> e.g., the struct could have the byte offset, but also the percentage if know it.
> 
> The channel is optional. They can do whatever the heck they want with it then. :)

Well, a `ProgressUpdate <- chan ProgressStruct` and a `func ProgressUpdate(ProgressStruct)` are semantically substitutable, and different callers may prefer one over another. But I expect that a real progress-bar implementation would want to know more:

- Of course, progress of one of the progress-reporting operations has been updated
  - This will always include at least a raw value; it _may_ also report other stats (current / average network throughput; is the speed limited by network or local storage speed?)
- A new progress-reporting operation has been added to the work plan: (e.g. first we download a manifest; _after_ doing that and parsing it, we may decide that we need to copy N layers).
  - That operation may or may not have a duration known in advance (file size if copying it; perhaps a percentage if doing a scan of something, or compressing/decompressing as a separate step)
  - That operation may or may not have a human-readable name (specific path, “creating a temporary copy”, vs. unreadable digest values)
- One of the progress-reporting operations has successfully finished.
- One of the progress-reporting operations has failed, with a specific error message.

For all of this, a typed interface with appropriate method names would probably be a better fit than a channel (which could be used, but every consumer would then have to do a `switch incomingValue.ReportKind` or something similar using the type system).

And yes, this seems fairly complex, which is why I keep asking whether there is a nice existing interface definition we could just use without having to invent this.

**Comment by erikh** [comment]
Yeah see I don't think that's necessary at all -- nor will we have the capability of determining things like network bandwidth consumed without a lot of extra tooling.

However, one can derive that pretty trivially with the implementation above -- just divide the number by the time transpired at any reporting interval.

What I really just want -- and literally nothing else -- is something that has nothing to do with progress bars itself, but the reporting of the progress. Error states, etc can come from different sources.

I can build progress meters from that.

**Comment by mtrmac** [comment]
> Yeah see I don't think that's necessary at all

I can buy that for the extra details and errors, but not the start/stop indications.

How can anyone consume the one-value "here’s an integer” data when `copy.Image` makes many copies? How can that happen without “new copy starting” notification, when the caller does not know in advance how many copies will be made?

**Comment by erikh** [comment]
Here's the latest program and patch:

```go
package main

import (
	"fmt"
	"time"

	"github.com/containers/image/copy"
	"github.com/containers/image/docker/daemon"
	"github.com/containers/image/docker/reference"
	"github.com/containers/image/signature"
)

func main() {
	ref, err := daemon.ParseReference("docker.io/library/golang:latest")
	if err != nil {
		panic(err)
	}

	signal := make(chan types.SignalProperties)

	go func() {
		for prop := range signal {
			fmt.Println(prop.Artifact, prop.Delta)
		}
	}()

	// you can do it this way too

	img, err := ref.NewImage(&types.SystemContext{
		SignalInterval: 100 * time.Millisecond,
		Signal:         signal,
	})
	if err != nil {
		panic(err)
	}
	defer img.Close()

	close(signal)

	tgtRef, _ := reference.ParseNamed("docker.io/erikh/test:latest")
	tgt, err := daemon.NewReference("", tgtRef)
	if err != nil {
		panic(err)
	}

	if err != nil {
		panic(err)
	}

	b, _, err := img.Manifest()
	if err != nil {
		panic(err)
	}
	fmt.Println(string(b))

	pc, err := signature.NewPolicyContext(&signature.Policy{
		Default: []signature.PolicyRequirement{signature.NewPRInsecureAcceptAnything()},
	})

	if err != nil {
		panic(err)
	}

	signal = make(chan types.SignalProperties)

	go func() {
		for prop := range signal {
			fmt.Println(prop.Artifact, prop.Delta)
		}
	}()

	err = copy.Image(
		pc,
		tgt,
		ref,
		&copy.Options{
			SignalInterval:   100 * time.Millisecond,
			Signal:           signal,
			RemoveSignatures: true,
		})

	close(signal)

	if err != nil {
		panic(err)
	}
}
```

As you can see I've moved it to use a channel with a communicated struct that
current contains the offset of the copy (which is labeled delta here on
accident, will fix in a bit). The only thing I don't like about this interface
is that you have to close your own channels (in the source case, we can't avoid
it), but that's not really terrible.

mtrmac:
Thanks.

- I still think that the progress hook should only be in `copy.Image`, not in the individual transport implementations; the underlying implementation can be made available to low-level users who want to call `PutBlob`/`GetBlob` directly.
- Also, an interface would be more suitable than a channel, because it allows “starting an artifact copy” / “ending an artifact copy” notifications in a clean, type-safe way. At least the latter is very valuable.

erikh:
Hmm. I disagree on the interface but I can implement it NP. I just think a
struct is simple enough and can accept methods if absolutely necessary; I
don't see how polymorphism is a benefit here.

I'll try to get you something in the next day or two.

mtrmac:
@runcom , opinions on the channel vs. interface issue?

erikh:
Hey @mtrmac, I updated to use copy.Image and I think you will find the API
considerably more malleable this way. I also am passing a types.ImageBlob as
the artifact now instead of a string.

erikh:
The last posted example should still work.

copy/copy.go
```diff
@@ -173,6 +179,8 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 		diffIDsAreNeeded:  src.UpdatedImageNeedsLayerDiffIDs(manifestUpdates),
 		canModifyManifest: canModifyManifest,
 		reportWriter:      reportWriter,
+		signalInterval:    options.SignalInterval,
+		signal:            options.Signal,
```
mtrmac: `options` may be `nil`

copy/copy.go
```diff
@@ -489,8 +497,20 @@ func (ic *imageCopier) copyBlobFromStream(srcStream io.Reader, srcInfo types.Blo
...
-	uploadedInfo, err := ic.dest.PutBlob(destStream, inputInfo)
+	// the pipe is here to facilitate the stream copier which will act as an
+	// intermediary to signal progress information to any registered channel.
+	if ic.signal != nil && ic.signalInterval > 0 {
+		r, w := io.Pipe()
+		go streamcopy.WithProgress(w, destStream, inputInfo, ic.signalInterval, ic.signal)
+		stream = r
+	} else {
+		stream = destStream
+	}
```
mtrmac: Please use the same stage description structure as the rest of the steps, this function is pretty complex and the regularity makes it easier to understand:

- Create a separate newline-separated section for this stage in the pipeline, starting with a `// === `-prefixed comment
- Use `destStream` as both the input and output of that pipeline stage (i.e. `if ic.signal != nil … { … destStream = r }  ` with no `else` branch)

streamcopy/copy.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"bufio"
+	"io"
+	"time"
+
+	"github.com/containers/image/types"
+)
+
+const megaByte = float64(1024 * 1024)
```
mtrmac: This is not used.

streamcopy/copy.go
```diff
@@ -0,0 +1,57 @@
...
+const readerSize = 32768
+
+// WithProgress implements io.Copy with a buffered reader, then measures
+// reports to the offsetFunc with the count processed at the interval.
+// If io.EOF is not returned then the error is returned. Otherwise, it is nil.
+// WithProgress closes the writer after finishing.
```
mtrmac: The comment needs updating (e.g. there is no `offsetFunc`)

streamcopy/copy.go
```diff
@@ -0,0 +1,57 @@
...
+func WithProgress(writer io.WriteCloser, reader io.Reader, artifact types.BlobInfo, interval time.Duration, signal chan types.SignalProperties) error {
```
mtrmac: Returning a value from a goroutine is not really useful.

(It might well make sense to encapsulate a bit more, i.e. to have a public function here which accepts an `io.Reader` and returns an `io.Reader`, making the pipe and the goroutine entirely invisible to the rest of the code.)

streamcopy/copy.go
```diff
@@ -0,0 +1,57 @@
...
+	defer writer.Close()
```
mtrmac: The `writer`’s `CloseWithError` is the only real error reporting mechanism of this goroutine, so we need to take advantage of it.

See how `compressGoroutine` uses an `err` variable and `defer` to report errors from the goroutine into the `PipeWriter`.

streamcopy/copy.go
```diff
@@ -0,0 +1,57 @@
...
+	rd := bufio.NewReaderSize(reader, readerSize)
```
mtrmac: Why is the buffering useful in here? Immediately below we are creating our own buffer.

(I’m not saying it isn’t, I just can’t immediately see that it is.)

streamcopy/copy.go
```diff
@@ -0,0 +1,57 @@
...
+		rn, err := rd.Read(buf)
+		count += uint64(rn)
+
+		if err == io.EOF {
+			if rn > 0 {
+				goto write
+			} else {
+				return nil
+			}
+		}
```
mtrmac: This `goto` seems unnecessary; AFAICS this _guarantees_ that there will not be a “signal” with the final 100% `count` value. But…

streamcopy/copy.go
```diff
@@ -0,0 +1,57 @@
...
+		if err != nil {
+			return err
+		}
```
mtrmac: It is documented that `io.Reader` can return `rn >0 && err != nil` at the same time; i.e. the code should first write any outstanding data if any, and _then_ check `err`.  (See the implementation of `io.Copy`)

streamcopy/copy.go
```diff
@@ -0,0 +1,57 @@
...
+
+		if interval > 0 && signal != nil {
+			if time.Since(t) > interval {
+				signal <- types.SignalProperties{Delta: count, Artifact: artifact}
```
mtrmac: So `Delta` is not a “delta” (change) since the last update, but the total offset? It would be more natural to call the member `Offset` then.

copy/copy.go
```diff
@@ -489,8 +497,20 @@ func (ic *imageCopier) copyBlobFromStream(srcStream io.Reader, srcInfo types.Blo
...
 
+	var stream io.Reader
+
 	// === Finally, send the layer stream to dest.
-	uploadedInfo, err := ic.dest.PutBlob(destStream, inputInfo)
+	// the pipe is here to facilitate the stream copier which will act as an
+	// intermediary to signal progress information to any registered channel.
+	if ic.signal != nil && ic.signalInterval > 0 {
+		r, w := io.Pipe()
+		go streamcopy.WithProgress(w, destStream, inputInfo, ic.signalInterval, ic.signal)
```
mtrmac: Shouldn’t this use `srcInfo` instead of `inputInfo`? `inputInfo` may be `{Digest: "", Size:-1}` in the “compressing on the fly” case.

types/types.go
```diff
@@ -293,6 +293,16 @@ type SystemContext struct {
...
 	// in order to not break any existing docker's integration tests.
 	DockerDisableV1Ping bool
+
+	SignalInterval time.Duration         // time to wait between reports to the Signal
+	Signal         chan SignalProperties // Reported to when SignalInterval has arrived
+}
+
+// SignalProperties is used to pass information from the copy code to a monitor which
+// can use the real-time information to produce output or react to changes.
+type SignalProperties struct {
```
mtrmac: ”Signal” is pretty generic. `s/Signal/Progress/g`?

copy/copy.go
```diff
@@ -489,8 +497,20 @@ func (ic *imageCopier) copyBlobFromStream(srcStream io.Reader, srcInfo types.Blo
...
 	}
 
+	var stream io.Reader
+
 	// === Finally, send the layer stream to dest.
-	uploadedInfo, err := ic.dest.PutBlob(destStream, inputInfo)
+	// the pipe is here to facilitate the stream copier which will act as an
+	// intermediary to signal progress information to any registered channel.
+	if ic.signal != nil && ic.signalInterval > 0 {
+		r, w := io.Pipe()
```
mtrmac: `defer r.Close()` somewhere, otherwise we could exit with an unreferenced unclosed `PipeReader` and a `PipeWriter` still blocked on trying to write to it.

erikh:
ok I'll make these changes; some of this was just churn due to the design
adjustments fwiw.

Gimme a few hours.

erikh:
oops

streamcopy/copy.go
erikh: Are you just objecting to it becuase it uses goto?

streamcopy/copy.go
erikh: This is taken from some old code I used in my own stuff. It may not all be necessary here. I'll re-evaluate.

erikh:
Not finshed with review comments, but I did add a test if you'd like to peek.

erikh:
That test should be the exact behavior we want at this point, but it's passing fine here and not in travis. Anyone else mind testing it really quick to help me determine if I'm sane?

streamcopy/copy.go
mtrmac: No, I am not dogmatic about `goto`, though it does seem to me that a `goto`-less equivalent would be a bit shorter. The objection was really about skipping the final update with 100% progress.

Though, with the point about `io.Reader`  returning `rn >0 && err != nil`, I expect this to be restructured so much that nothing about these particular lines will remain applicable.

copy/copy.go
```diff
@@ -162,6 +168,10 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 		return err
 	}
 
+	if options == nil {
+		options = &Options{}
+	}
```
mtrmac: Fair enough, that works :) If we are doing it this way, can you move it to the top eliminate the three now redundant `options != nil` checks as well, please?

streamcopy/copy.go
erikh: Hmm. I'll review and adjust. Thanks.

copy/copy.go
```diff
@@ -489,6 +501,18 @@ func (ic *imageCopier) copyBlobFromStream(srcStream io.Reader, srcInfo types.Blo
 		inputInfo.Size = -1
 	}
 
+	// the pipe is here to facilitate the stream copier which will act as an
+	// intermediary to signal progress information to any registered channel.
+	if ic.progress != nil && ic.progressInterval > 0 {
```
mtrmac: The comment about the pipe logically seems to belong inside the `if`?

Above the `if`, please add something like
```go
// === Report progress using the ic.progress channel, if required
```
(i.e. saying _what_ this part of the pipeline does, before talking about _how_)

BTW, is it intentional that this is after the compression unlike the `bar.NewProxyReader` progress reporting which is a bit above?

streamcopy/copy.go
mtrmac: It would be more reliable to default to a `CloseWithError` and only to set `err` to `nil` if we _know_ we are successful; we can exit the function not only by `return`, but also by a `panic` anywhere in the called code.

See `compressGoroutine`; I don’t know that it is perfect but it does seem to be more paranoid :)

erikh:
```
It would be more reliable to default to a CloseWithError and only to set err to nil if we know we are successful; we can exit the function not only by return, but also by a panic anywhere in the called code.

See compressGoroutine; I don’t know that it is perfect but it does seem to be more paranoid :)
```

I really am not connecting how this extra logic is different, and it is significantly harder to read for me at least.

erikh:
The rest of the changes have been made; I'll see if I can find another host to run these tests on to suss that out.

copy/progress_test.go
```diff
@@ -0,0 +1,112 @@
...
+)
+
+// THIS TEST SHOULD BE SLOW. Read inner comments for details.
+func TestCopyProgress(t *testing.T) {
+	interval := 100 * time.Millisecond
+
+	// we shell out to avoid "client is newer than server" errors from docker
+	// which will happen when importing the client straight from github.
+	// This of course presumes docker is in the path.
+	if out, err := exec.Command("/bin/sh", "-c", "docker pull golang:latest").CombinedOutput(); err != nil {
```
mtrmac: Hum, so far we’ve been able to make the tests run as non-root, and to not affect any global state. (True, for example by not having any `docker-daemon:` tests among other things.)

Can this be preserved?

AFAICS the test could be testing `copyBlobFromStream` directly, with `ic.dest.PutBlob` a mock which just copies the input to `ioutil.Discard`. That would allow submitting a specially crafted `srcStream`, perhaps one which provides exactly one byte on each call, with an exactly 10 ms delay — in turn allowing exact equality checks on the offsets, more precise tests on the timings, and it could make the test run much faster.

copy/progress_test.go
```diff
@@ -0,0 +1,112 @@
...
+	// This validates that enough events arrived to satisfy the interval time. As
+	// a result it may require a little bit of work to get right before it will
+	// pass for all people all the time. Sorry in advance!
+	go func() {
+		var (
+			// if this is false, the time will be populated and this will be set to true.
+			// This is to ensure we're measuring the progress itself and not other
+			// operations.
+			timed bool
+			tm    time.Time
```
mtrmac: So the variables really are `firstReportedTime` and `firstReportedTimeSet`? (It’s not that overlong variable names are desirable in principle, but in this case just naming the variables precisely would 1) eliminate the need for the comment, and 2) be explicit, when so many values in here are a “time”)

[May be irrelevant if this test is reworked with a special input stream.]

copy/progress_test.go
```diff
@@ -0,0 +1,112 @@
...
+			count int
+		)
+
+		for range signal {
+			if !timed {
+				tm = time.Now().Add(interval * 2) // account for elapsed time already before the chan will tick at all
```
mtrmac: (Note to self: so `tm` is “first reported time + interval * 2”)

copy/progress_test.go
```diff
@@ -0,0 +1,112 @@
...
+				timed = true
+			}
+			count++
+		}
+
+		elapsed := time.Duration(count)*interval + interval // last one
```
mtrmac: (Note to self: `elapsed` = rough upper bound on the total time, assuming no `.Read()` blocks for more than `interval`.)

copy/progress_test.go
```diff
@@ -0,0 +1,112 @@
...
+		since := time.Since(tm)
```
mtrmac: (Note to self: `since` = time spent since the first report - `interval * 2` < actual elapsed time - `interval * 2`)

copy/progress_test.go
```diff
@@ -0,0 +1,112 @@
...
+		delta := since - elapsed
```
mtrmac: (Note to self: `delta` < (actual elapsed - `interval * 2`) - (upper bound on elapsed) < `-interval * 2`, assuming everything goes well?)

copy/progress_test.go
```diff
@@ -0,0 +1,112 @@
...
+		if delta > interval {
+			errChan <- fmt.Errorf("counts did not equal during progress signaling: (delta: %d) (since: %d) (elapsed: %d)", delta, since, elapsed)
```
mtrmac: Just from reading the code it is unclear what the test is supposed to test. AFAICS, given the “not to self” comments above this should fail if the progress reporting channel has not received enough updates over time; is that right?

streamcopy/copy.go
```diff
@@ -0,0 +1,54 @@
...
+		if err != nil {
+			writer.CloseWithError(err)
+			return
+		}
+
+		if interval > 0 && signal != nil {
+			if time.Since(t) > interval {
+				signal <- types.ProgressProperties{Offset: count, Artifact: artifact}
+				t = time.Now()
+			}
```
mtrmac: As for the cause of the test failure, _if_ I understand correctly what the test is testing for:

I think the cause is that `reader.Read` above may block for _much_ more than `interval`, but this code only sends one progress report. (This should be possible to verify by adding logging of (timestamp, artifact, offset) to the loop consuming the progress updates.)

To solve this, there could be another goroutine which converts the input `io.Reader` into a channel (of buffers?), and this top-level goroutine would `selectd` on that channel of input data and a `time.Ticker` or something like that.

But that’s much more fairly code; I’d be perfectly fine with keeping ~the current logic and just giving up on the promise that there will be an update about every `interval`.

mtrmac:
> > It would be more reliable to default to a CloseWithError and only to set err to nil if we know we are successful; we can exit the function not only by return, but also by a panic anywhere in the called code.
> > 
> > See compressGoroutine; I don’t know that it is perfect but it does seem to be more paranoid :)
>
> I really am not connecting how this extra logic is different, and it is significantly harder to read for me at least.

Consider the case when `reader.Read` panics: Then nothing closes `writer` at all, let alone closing it with an error report, and the upload hangs for an unknown reason.

And yeah, error handling and recovery with goroutines is a pain. One possible approach is to separate the goroutine-handling aspects (`defer`s to guarantee resources are freed and errors are forwarded to the parent goutine) in a separate function, and then handle the actual business logic in a separate straight-line function. `diffIDComputationGoroutine` vs. `computeDiffID` take this approach.

copy/progress_test.go
mtrmac: (or even better, instead of discarding the output, record it so that we can later verify the copy has happened correctly.)

copy/progress_test.go
mtrmac: Actually, due to the `interval * 2` fuzz, those names would not be accurate, and accurate names would be too hard. Never mind.

mtrmac:
> And yeah, error handling and recovery with goroutines is a pain

Actually, AFAICS there is no need for pipes and goroutines in here at all:
```go
struct ProgressReportingReader {
  source io.Reader
  channel ..
  interval ..
  lastTime time.Time
  total int64
}

func (r progressReportingReader) Read(p []byte) (int64, error) {
  n, err := p.source.Read(p)
  r.total += n
  if time.Since(r.lastTime) > r.interval {
    r channel <- …
    r.lastTime = time.Now()
  }
  return n, err
}
```

± some initialisation etc., seems that this should work equally well but be much simpler.

The other code uses pipes to connect `Reader`s and `Writer`s, but we don’t have to do that here.

erikh:
I'm happy to rewrite but let's make sure we're 100% clear on what you want. Anything else that needs to change?

mtrmac:
I'd be fine with either a goroutine implementation or just the `io.Reader` wrapper really, as long as it is reliable and does not discard underlying error information. I guess that the `io.Reader` wrapper would be the easier way but that’s up to you.

erikh:
I've updated to use your new pattern but the tests aren't passing. Not sure how to best test this yet, so I'm going to tinker for a while. Let me know if you'd like additional changes.

erikh:
giving up on a test unless you have a better idea :) Let me know if you'd like anything changed, but I believe this is merge-worthy at this point.

copy/copy.go
```diff
@@ -92,12 +95,19 @@ type Options struct {
...
 func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageReference, options *Options) error {
 	reportWriter := ioutil.Discard
-	if options != nil && options.ReportWriter != nil {
+
+	if options == nil {
+		options = &Options{}
+	}
+
+	if options.ReportWriter != nil {
 		reportWriter = options.ReportWriter
```
mtrmac: One last thing, please move the `reportWriter := ioutil.Discard` initialization near this `if`.

erikh:
Ok. I will make the get/putblob patch in a second one. I could probably use
it as well.

erikh:
This should be ready now.

mtrmac:
:+1: pending tests. Thanks!

erikh:
Just kicked the tests, sorry this took so long.

erikh:
tests are passing, looks like it needs another approval to get through

mtrmac:
@runcom PTAL

copy/copy.go
```diff
@@ -92,14 +95,22 @@ type Options struct {
 	ReportWriter     io.Writer
 	SourceCtx        *types.SystemContext
 	DestinationCtx   *types.SystemContext
+	ProgressInterval time.Duration                 // time to wait between reports to signal the progress channel
+	Progress         chan types.ProgressProperties // Reported to when ProgressInterval has arrived for a single artifact+offset.
```
runcom: if this is a receive only channel, could you strongly type it here?

copy/progress_reader.go
```diff
@@ -0,0 +1,28 @@
...
+	"io"
+	"time"
+
+	"github.com/containers/image/types"
+)
+
+// progressReader is a reader that reports its progress on an interval.
+type progressReader struct {
+	source   io.Reader
+	channel  chan types.ProgressProperties
```
runcom: ditto for the receive only channel I guess

copy/copy.go
erikh: This won't work because we do send to the channel eventually from our code, not the foreign code.

Unless I misunderstood you, when I implemented this, as soon as the send is hit by the compiler in `copy/progress_reader.go` it aborts.

runcom: mmm ok not blocking this PR I guess

runcom:
Could you rebase this @erikh ?

erikh:
done

runcom:
@erikh one last check before I merge, is there any skopeo PR which takes care of the test failure here?

copy/copy.go
mtrmac: AFAICS this should actually be a send-only channel. The code calling `copy.Image` needs a bidirectional channel, but `copy.Image` will only be sending to it, and the rest of the code calling `copy.Image` will only be receiving from it.

runcom: so yeah, my point was: let's strongly type the direction when we can in order to allow the go compiler (or runtime?) to be more efficient (I read this somewhere, I don't remember where, I just remember strongly typing the direction of the channel helped). Again, not blocking this PR.

mtrmac:
> @erikh one last check before I merge, is there any skopeo PR which takes care of the test failure here?

I think that test failure (`manifest/manifest.go:57: undefined: v1.MediaTypeImageManifestList`) is #240 . This should not break skopeo AFAICS.

copy/copy.go
mtrmac: ACK, strongly typing is useful enough just for the documentation value; but not blocking for me either.

runcom:
:+1:

merging then

mtrmac:
(Um, it would still have been useful to merge #240 first and see the tests pass. But, well, it’s done, not worth reverting at this point just for insiting on the format process.)

PR #207, opened 2017-01-09
Allow choosing between a gpgme and openpgp signature backend using a build tag

mtrmac:
This demonstrates an alternative to #198 , which percolates the mechanism choice through the API and does not permit file-format-dependent mechanism selection.

In here, the default mechanism is gpgme; a `containers_image_openpgp` build tag can be used to use openpgp instead.

openpgp does not currently support signing, and is based on mfojtik's implementation (adding GPG home directory support, parsing of unarmored keys, and fixing `ImportKeysFromBytes` semantics).

NOTE: The openpgp backend is not really fleshed out yet. Some of the `mechanism_test.go` tests may be better mechanism-specific, and openpgp definitely needs more tests e.g. for `optionalDir`.

PR #123, opened 2016-10-10
docker: fork docker/docker/reference pkg

runcom:
As discussed into https://github.com/containers/image/issues/99

@mtrmac PTAL, making a PR for skopeo check if everything's ok over there. I'll check with projectatomic/docker#200 as well.

Signed-off-by: Antonio Murdaca <runcom@redhat.com>

PR #221, opened 2017-01-20
Drop c/i/docker/reference (again)

mtrmac:
This is a follow-up on #220 , which minimally updated us for the changed API.

This wholesale transitions containers/image to rely on the `docker.io/library` canonicalization support now available in docker/distribution/reference, removing our c/i/docker/reference fork of docker/docker/reference.  That’s a nice simplification, but also a blind bet on how projectatomic/docker will/will not port its support for unqualified references to the new docker/distribution/reference.

There are three ways to review this PR:

 - As the resulting diff, which simply removes c/i/docker/distribution/reference and replaces all users in one go, somewhat difficult to follow for a total of +184-702 lines.
 - As a sequence of 37 commits, with a total of +864-1382 lines, but with a good explanation of every step, and tests (apart from golint) passing along the way
 - Just give up and reject it outright :)

Choose your poison.

NOTE: _This changes behavior_.  Before, `docker/reference.ParseNamed` would silently turn `name:tag@digest` into `name@digest`, dropping the tag. Now, `distribution/reference.ParseNormalizedNamed` preserves all of the information, which triggers several code paths expecting such complex values and rejecting them outright.

I don’t think that is _likely_ to hurt, and it is usually not quite trivial to see what the behavior on a tag+digest should be (especially because further consumers are or may not be ready to correctly handle this). We could, though, define and implement some semantics for such values.


runcom:
> but also a blind bet on how projectatomic/docker will/will not port its support for unqualified references to the new docker/distribution/reference.

@mtrmac what do you mean with this?

mtrmac:
> > but also a blind bet on how projectatomic/docker will/will not port its support for unqualified references to the new docker/distribution/reference.

> @mtrmac what do you mean with this?

We have had to make a copy of `docker/docker/reference` because `projectatomic/docker` was patching its version in a way which broke signatures. (#123).

AFAICT the plan with the new `docker/distribution/reference` is to eliminate `docker/docker/reference` entirely (see e.g. https://github.com/dmcgowan/docker/pull/27); if that happens, `projectatomic/docker` will have to change _something_ to keep its current behavior of searching across registries; and it is currently unknown whether that would involve changes to `docker/distribution/reference`, and if so, whether they would change the semantics / break our use cases again.

mtrmac:
@runcom Opinions on this? Rebasing this branch over time is doable, but it is rather painful.

Looking at upstream https://github.com/docker/docker/blob/master/reference/reference.go , it has now become a pretty trivial forwarding layer to docker/distribution/reference—with `docker/docker/reference.Named` conforming to the `docker/distribution/reference.Named` interface but having different normalization semantics, which seems pretty confusing; it is much better to eliminate the docker/docker/reference semantics IMO, as this PR is doing.

Separately, per the conversation in #207 , due to version drift we probably can’t avoid either vendoring or forking the docker/distribution/reference package. My proposal is to keep this PR as is, and land a single “copy docker/distribution/reference to contaienrs/image/docker/reference` commit on top. Does that sound acceptable?

runcom:
> Separately, per the conversation in #207 , due to version drift we probably can’t avoid either vendoring or forking the docker/distribution/reference package. My proposal is to keep this PR as is, and land a single “copy docker/distribution/reference to containers/image/docker/reference` commit on top. Does that sound acceptable?

works for me. The only thing I fear right now is the huge amount of work I think I need to do in `projectatomic/docker` for the multi registries story. But yeah, that sounds fine to me.

mtrmac:
> The only thing I fear right now is the huge amount of work I think I need to do in `projectatomic/docker` for the multi registries story.

I appreciate that that will be painful for `projectatomic/docker`; does that mean something should be changed in `containers/image`, though? It seems to me that `projectatomic/docker` does the search / resolution in any way it wants, and then it just submits the resolved name to `containers/image`. At worst the reference will have to go though a “format as string → parse into the containers/image-normalized format” roundtrip, which IIRC is currently happening as well.

runcom:
> I appreciate that that will be painful for projectatomic/docker; does that mean something should be changed in containers/image, though?

nope, I didn't mean this, sorry

mtrmac:
OK, updated:
 - Updated for another API change in https://github.com/docker/distribution/pull/2173
 - Dropped a no longer necessary workaround for https://github.com/docker/distribution/issues/2163
 - Added a final copy which adds a `containers/image/docker/reference` as a fork of `docker/distribution/reference`, with the `digestset` dependency removed.

runcom:
> Added a final copy which adds a containers/image/docker/reference as a fork of docker/distribution/reference, with the digestset dependency removed.

so, with this bullet in place I'm free to patch `projectatomic/docker` whatever I want wrt references and that wouldn't affect c/image signatures, correct? I'm still trying to find a way not to patch `projectatomic/docker` but I'm struggling doing that...we'll see.

mtrmac:
> so, with this bullet in place I'm free to patch `projectatomic/docker` whatever I want wrt references and that wouldn't affect `c/image` signatures, correct? 

Assuming a `docker/docker/reference.Named`→`string`→`containers/image/docker/reference.Named` roundtrip (which IIRC the current verification patch is already doing), sure.

runcom:
lgtm

mtrmac:
:+1:

PR #222, opened 2017-01-21
Integrate github.com/fzipp/gocyclo

erikh:
this is integration of a validation tool called gocyclo, which you can
find more information on [here](https://github.com/fzipp/gocyclo).

Please stop me if this is an unwelcome patch, but I personally have found it to
literally excavate large chunks of ugly code into manageable parts, and I think
it would behoove the project to incorporate it into its test/CI cycle to
encourage creating more modular code.

gocyclo counts the number of branches/loops and basically provides a safe
metric for the complexity and length of a given function.

I'll be taking the time in this PR to reorganize the offending functions:

Calculated with `gocyclo -over 15 .`

```
44 storage TestWriteRead storage/storage_test.go:230:1
32 storage (*storageImageDestination).PutBlob storage/storage_image.go:137:1
30 copy Image copy/copy.go:98:1
22 storage TestDuplicateBlob storage/storage_test.go:752:1
21 signature (*prSignedBy).UnmarshalJSON signature/policy_config.go:358:1
20 storage (*storageImageDestination).Commit storage/storage_image.go:311:1
19 signature (*prSignedBy).isSignatureAuthorAccepted signature/policy_eval_signedby.go:18:1
19 storage (storageTransport).ValidatePolicyConfigurationScope storage/storage_transport.go:220:1
18 openshift validateAuthInfo openshift/openshift-copies.go:373:1
17 docker deleteImage docker/docker_image_src.go:252:1
17 docker (*dockerClient).ping docker/docker_client.go:436:1
17 openshift (*openshiftClient).doRequest openshift/openshift.go:72:1
16 docker TestGetAuthFailsOnBadInput docker/docker_client_test.go:330:1
16 docker (*dockerClient).setupRequestAuth docker/docker_client.go:240:1
16 signature (*PolicyContext).GetSignaturesWithAcceptedAuthor signature/policy_eval.go:176:1
```

Please let me know what you think, I'll start executing immediately but I'm
happy to stop if instructed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@
   go:
     - 1.7
   install: make deps
-  script: make .gitvalidation && make validate && make test && make test-skopeo
+  script: make .gitvalidation && make validate && make test && make test-root-storage && make test-skopeo 
   dist: trusty
   os:
     - linux
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,14 @@ deps:
 	go get -u $(BUILDFLAGS) github.com/golang/lint/golint
 	go get $(BUILDFLAGS) github.com/vbatts/git-validation
 
+gocyclo:
+	go get github.com/fzipp/gocyclo
+	@out=$$(gocyclo -over 15 . | grep -v vendor); \
+	if [ -n "$$(gocyclo -over 15 . | grep -v vendor)" ]; then \
+		echo "$$out"; \
+		exit 1; \
+	fi
+
 test:
 	@go test $(BUILDFLAGS) -cover ./...
 
@@ -35,7 +43,7 @@ test-skopeo:
 		$(SUDO) make check && \
 		rm -rf $${skopeo_path}
 
-validate: lint
+validate: lint gocyclo
 	@go vet ./...
 	@test -z "$$(gofmt -s -l . | tee /dev/stderr)"
 
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,9 @@ gocyclo:
 test:
 	@go test $(BUILDFLAGS) -cover ./...
 
+test-root-storage:
+	@sudo -E `which go` test -v ./storage
+
 # This is not run as part of (make all), but Travis CI does run this.
 # Demonstarting a working version of skopeo (possibly with modified SKOPEO_REPO/SKOPEO_BRANCH, e.g.
 #    make test-skopeo SKOPEO_REPO=runcom/skopeo-1 SKOPEO_BRANCH=oci-3 SUDO=sudo
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,8 @@ deps:
 	go get $(BUILDFLAGS) github.com/vbatts/git-validation
 
 gocyclo:
-	go get github.com/fzipp/gocyclo
+	@echo Fetching/Running github.com/fzipp/gocyclo...
+	@go get github.com/fzipp/gocyclo
 	@out=$$(gocyclo -over 15 . | grep -v vendor); \
 	if [ -n "$$(gocyclo -over 15 . | grep -v vendor)" ]; then \
 		echo "$$out"; \
@@ -44,10 +45,12 @@ test-skopeo:
 		rm -rf $${skopeo_path}
 
 validate: lint gocyclo
+	@echo Running go vet...
 	@go vet ./...
 	@test -z "$$(gofmt -s -l . | tee /dev/stderr)"
 
 lint:
+	@echo Running golint...
 	@out="$$(golint ./...)"; \
 	if [ -n "$$(golint ./...)" ]; then \
 		echo "$$out"; \
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -94,8 +94,7 @@ type Options struct {
 	DestinationCtx   *types.SystemContext
 }
 
-// Image copies image from srcRef to destRef, using policyContext to validate source image admissibility.
-func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageReference, options *Options) error {
+func getReportWriter(options *Options) (io.Writer, func(string, ...interface{})) {
 	reportWriter := ioutil.Discard
 	if options != nil && options.ReportWriter != nil {
 		reportWriter = options.ReportWriter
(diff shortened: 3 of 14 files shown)


runcom:
:+1: to have this.

erikh:
Taking a break on this for a bit. Please review what I have done, if you get
time, this will be a very large patch.

runcom:
@nalind please take a look as well

erikh:
This should pass tests and be ready for review.

Note that I integrated root UID tests into travis so that the storage tests get
properly exercised (they were skipping a lot of tests before).

erikh:
Looks like there's some deps issues again.

You guys mind if I tackle a holistic deps rearch in another ticket? Seems like
there's a lot of churn here.

mtrmac:
(As a first impression, without actually reading the refactorings in detail:) Hum, I am not sure that just splitting a long sequential function into several smaller functions called sequentially makes the code substantially better; instead of a sequential flow which can be read top-to-bottom, the code adds new function references and calls, and new parameter names (and variable/parameter/return-value mapping) which all have to be followed by the reader to understand how the code works.

_If_ the long function can be split into cleanly conceptually separate parts (e.g. if the newly separate function contains non-trivial ideas, or a higher number of variables, which the the caller does not care about), sure. And I’m sure we do have such function which would benefit from a refactoring.

But making forbidding long functions by a blind metric, and splitting functions just because a metric says so, seems fairly likely to make some code worse instead of better.

runcom:
Yup, hadn't really took a look at the code here, but the main idea behing gocyclo would be to give you an _hint_ of the complexity of a function. Simply splitting it into chunks is not gonna help (if I'm reading @mtrmac's comment correctly).

erikh:
Ok. Should I close this?

mtrmac:
No, please keep it around. I haven’t had time to take a _good_ look yet, but it is very clear that at least some parts of this (`mak test-root-storage`, the `ping` refactors) are useful, either as they are or with minor modifications. Just don’t rush rewriting more functions for now.

erikh:
Sure. I will say that this should totally pass gocyclo now, fwiw.

As for breaking up the functions, I disagree that large functions are
occasionally good. I removed in many cases inline functions and very large,
IMO hard-to-read long loops into separate functions. The advantage of
keeping it around is that I think it will result in better code reviews (or
reviews that are expedited) due to smarter function usage. I'm going to
posit that anyone willing to read the source is also more than capable of
finding function references quickly, so it should not be too much
additional work. We just would need to agree on the semantics of each
change.

The changes as they stand now could use some improvement but depending on
the style of parameter passing (e.g., would you rather pass a struct or a
list, I generally will do the former for >4 params but I tried to KISS for
now) you folks prefer it may be unacceptable at this point; but we can
improve this as long as you are willing to guide me (and have adequate time
to review a large patch, heh).

Additionally the function changes are a quick port, they could totally be
reworked now or later to have a better semantic flow.

Got a lot of free time to hack on this so let me know;  I can probably turn
it around same-day.

mtrmac:
> I'm going to posit that anyone willing to read the source is also more than capable of finding function references quickly, so it should not be too much additional work.

“debugging is twice as hard as writing a program in the first place” or something :) If a refactoring causes _any_ additional work then it is counterproductive. The compiler doesn’t care how we factor things, the refactoring is supposed to help humans.

To randomly take just the first example:
```
reportWriter, writeReport := getReportWriter(options)
```
 - This shows that the function is not accurately named (and it seems difficult to find a concise and accurate name), and completely hides the relationship between the two objects.
 - This adds 5 lines (declaration, brackets, empty line, call) to isolate 7 lines of code. That’s a pretty massive overhead.
 - Meanwhile the contents of the function is really trivial; the reader of the caller does not have its job easier by hiding the contents of the function. (OTOH it would be attractive to split even a 3-line function if it were really non-trivial and it could be described with a nice identifier, so that we could have a 3-line function with a 20-line comment, completely invisible to the callers who don’t care about the internals.)
 - Now, perhaps there _is_ a way to simplify even this particular piece. It seems quite plausible that `writeReport` shouldn’t be a lambda expression, but a method, perhaps it could be a private method on `Options`. But the line count / cyclometric analysis doesn't suggest that, and doesn’t reward that. (We would have to add an `if options == nil { options = &Options{} }` into `copy.Image` to be able to call the method, which increases the complexity again.)
 - Somewhat paradoxically, wouldn’t replacing `writeReport` with inlined `fmt.Frrintf()` calls make the analysis happier? And yet that’s not really an improvement.

erikh:
Ok. The only thing at this point I can suggest is that you carry this patch
if you want to keep it. I'm not trying to be stubborn here but I feel this
is a subjective problem, not an objective one, and I can't guess what you
want.

It just feels now like we're heading into "rewrite the whole patch"
territory and while that's your choice, I don't know if that's worth
keeping this patch around for.

mtrmac:
> I'm not trying to be stubborn here but I feel this
> is a subjective problem, not an objective one, and I can't guess what you
> want.

That’s fair; we don’t have a consensus on a metric how to decide what is a good code. That’s a hard problem.

> It just feels now like we're heading into "rewrite the whole patch"
territory and while that's your choice, I don't know if that's worth
keeping this patch around for.

Basically, even if you did nothing further on this, the PR is useful _at least_ as pointing out to some areas of improvement. From my point of view it is simpler to keep it around, neither merged nor closed / branch deleted, which will preserve this useful property of the PR, than to close it and track the issues elsewhere. (It _may_ turn into one or more carried PRs, or into a set of file items. But that’s work to be done in the future.)

erikh:
Sure. Seems good.

rhatdan:
@mtrmac Can you grab the best ideas out of this and open an Issue/Trello card to work on it later and then close the PR.

rhatdan:
@mtrmac Should we close this PR?

mtrmac:
@rhatdan While this is mostly no longer _directly_ applicable (some of the code did get improved along the way, notably `copy.Image` got split a bit; a fair amount was changed enough that this has no chance of applying), there are still a few good ideas in here.

I don’t know how likely anyone will find the time to act on them, but an open PR does not hurt anything, does it?

erikh:
folks, I think I'm going to close this. it's been well over a year now, and nothing matches up any more I imagine.

If you want to crib the make file stuff, I think that's probably only the useful bits left in this patch.

PR #223, opened 2017-01-21
Fix for image spec

runcom:
https://github.com/opencontainers/image-spec/pull/411 removed the `MediaType` field in `Manifest[List]` causing builds error here. Fix that by teaching docker v2s2 to convert to OCI and remove that ugly manifest conversion in OCI image destination.

@mtrmac @cyphar PTAL

### DIFF
--- a/docker/docker_image_dest.go
+++ b/docker/docker_image_dest.go
@@ -17,6 +17,21 @@ import (
 	"github.com/pkg/errors"
 )
 
+var manifestMIMETypes = []string{
+	// TODO(runcom): we'll add OCI as part of another PR here
+	manifest.DockerV2Schema2MediaType,
+	manifest.DockerV2Schema1SignedMediaType,
+	manifest.DockerV2Schema1MediaType,
+}
+
+func supportedManifestMIMETypesMap() map[string]bool {
+	m := make(map[string]bool, len(manifestMIMETypes))
+	for _, mt := range manifestMIMETypes {
+		m[mt] = true
+	}
+	return m
+}
+
 type dockerImageDestination struct {
 	ref dockerReference
 	c   *dockerClient
@@ -47,12 +62,7 @@ func (d *dockerImageDestination) Close() {
 }
 
 func (d *dockerImageDestination) SupportedManifestMIMETypes() []string {
-	return []string{
-		// TODO(runcom): we'll add OCI as part of another PR here
-		manifest.DockerV2Schema2MediaType,
-		manifest.DockerV2Schema1SignedMediaType,
-		manifest.DockerV2Schema1MediaType,
-	}
+	return manifestMIMETypes
 }
 
 // SupportsSignatures returns an error (to be displayed to the user) if the destination certainly can't store signatures.
--- a/docker/docker_image_src.go
+++ b/docker/docker_image_src.go
@@ -39,6 +39,17 @@ func newImageSource(ctx *types.SystemContext, ref dockerReference, requestedMani
 	if requestedManifestMIMETypes == nil {
 		requestedManifestMIMETypes = manifest.DefaultRequestedManifestMIMETypes
 	}
+	supportedMIMEs := supportedManifestMIMETypesMap()
+	acceptableRequestedMIMEs := false
+	for _, mtrequested := range requestedManifestMIMETypes {
+		if supportedMIMEs[mtrequested] {
+			acceptableRequestedMIMEs = true
+			break
+		}
+	}
+	if !acceptableRequestedMIMEs {
+		requestedManifestMIMETypes = manifest.DefaultRequestedManifestMIMETypes
+	}
 	return &dockerImageSource{
 		ref: ref,
 		requestedManifestMIMETypes: requestedManifestMIMETypes,
--- a/image/docker_schema2.go
+++ b/image/docker_schema2.go
@@ -12,6 +12,7 @@ import (
 	"github.com/containers/image/manifest"
 	"github.com/containers/image/types"
 	"github.com/opencontainers/go-digest"
+	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 )
 
@@ -166,11 +167,38 @@ func (m *manifestSchema2) UpdatedImage(options types.ManifestUpdateOptions) (typ
 	case "": // No conversion, OK
 	case manifest.DockerV2Schema1SignedMediaType, manifest.DockerV2Schema1MediaType:
 		return copy.convertToManifestSchema1(options.InformationOnly.Destination)
+	case imgspecv1.MediaTypeImageManifest:
+		return copy.convertToManifestOCI1()
 	default:
 		return nil, errors.Errorf("Conversion of image manifest from %s to %s is not implemented", manifest.DockerV2Schema2MediaType, options.ManifestMIMEType)
 	}
 
 	return memoryImageFromManifest(&copy), nil
+}
+
+func (m *manifestSchema2) convertToManifestOCI1() (types.Image, error) {
+	// Create a copy of the descriptor.
+	config := m.ConfigDescriptor
+
+	// The only difference between OCI and DockerSchema2 is the mediatypes. The
+	// media type of the manifest is handled by manifestSchema2FromComponents.
+	config.MediaType = imgspecv1.MediaTypeImageConfig
+
+	layers := make([]descriptor, len(m.LayersDescriptors))
+	for idx := range layers {
+		layers[idx] = m.LayersDescriptors[idx]
+		if m.LayersDescriptors[idx].MediaType == manifest.DockerV2Schema2ForeignLayerMediaType {
+			layers[idx].MediaType = imgspecv1.MediaTypeImageLayerNonDistributable
+		} else {
+			layers[idx].MediaType = imgspecv1.MediaTypeImageLayer
+		}
+	}
+
+	// Rather than copying the ConfigBlob now, we just pass m.src to the
+	// translated manifest, since the only difference is the mediatype of
+	// descriptors there is no change to any blob stored in m.src.
+	m1 := manifestOCI1FromComponents(config, m.src, nil, layers)
+	return memoryImageFromManifest(m1), nil
 }
 
 // Based on docker/distribution/manifest/schema1/config_builder.go
--- a/image/docker_schema2.go
+++ b/image/docker_schema2.go
@@ -190,7 +190,7 @@ func (m *manifestSchema2) convertToManifestOCI1() (types.Image, error) {
 		if m.LayersDescriptors[idx].MediaType == manifest.DockerV2Schema2ForeignLayerMediaType {
 			layers[idx].MediaType = imgspecv1.MediaTypeImageLayerNonDistributable
 		} else {
-			layers[idx].MediaType = imgspecv1.MediaTypeImageLayer
+			layers[idx].MediaType = imgspecv1.MediaTypeImageLayerGzip
 		}
 	}
 
--- a/oci/layout/oci_dest.go
+++ b/oci/layout/oci_dest.go
@@ -37,7 +37,6 @@ func (d *ociImageDestination) Close() {
 func (d *ociImageDestination) SupportedManifestMIMETypes() []string {
 	return []string{
 		imgspecv1.MediaTypeImageManifest,
-		manifest.DockerV2Schema2MediaType,
 	}
 }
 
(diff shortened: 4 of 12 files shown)


runcom:
created a skopeo PR to validate this https://github.com/projectatomic/skopeo/pull/294

docker/docker_image_src.go
```diff
@@ -39,6 +40,15 @@ func newImageSource(ctx *types.SystemContext, ref dockerReference, requestedMani
 	if requestedManifestMIMETypes == nil {
 		requestedManifestMIMETypes = manifest.DefaultRequestedManifestMIMETypes
 	}
+	// if people has requested an OCI manifest here but the docker registry doesn't
+	// still understand it we should force add a request for docker v2s2. Otherwise
+	// the registry won't understand an OCI manifest request and fallback to v2s1
+	// and OCI doesn't understand it (it cannot convert from v2s1 to OCI)
+	for _, mt := range requestedManifestMIMETypes {
+		if mt == imgspecv1.MediaTypeImageManifest {
+			requestedManifestMIMETypes = append(requestedManifestMIMETypes, manifest.DockerV2Schema2MediaType)
```
mtrmac: This is ugly and suprising for an `ImageSource` to do (consider a caller who does not expect to use `image.Image`). Can we instead add the v2s1→OCI conversion? It might well be possible to do by creating a temporary v2s2 `memoryImage` (using the existing code) and then using the existing v2s2→OCI conversion.

mtrmac: (We haven’t supported a v2s1→OCI conversion before, so AFAICT this is not an essential part of this PR.)

image/docker_schema2.go
```diff
@@ -166,13 +167,40 @@ func (m *manifestSchema2) UpdatedImage(options types.ManifestUpdateOptions) (typ
...
 }
 
+func (m *manifestSchema2) convertToOCIManifest() (types.Image, error) {
+	// Create a copy of the descriptor.
+	config := m.ConfigDescriptor
+
+	// The only difference between OCI and DockerSchema2 is the mediatypes.
+	config.MediaType = imgspecv1.MediaTypeImageConfig
+
+	configBytes, err := m.ConfigBlob()
```
mtrmac: Is this really necessary to read the config here? It would perhaps be preferable to preserve the `ImageSource` and pass a `nil` `configBlob`, the way `manifestOCI1.convertToManifestSchema2` does it.

oci/layout/oci_dest.go
```diff
@@ -134,60 +133,25 @@ func (d *ociImageDestination) ReapplyBlob(info types.BlobInfo) (types.BlobInfo,
...
-		// and I don't think we should bother about this now (I don't want to have migration code here in skopeo)
-		return nil, "", errors.New("can't create an OCI manifest from Docker V2 schema 1 manifest")
-	case manifest.DockerV2Schema2MediaType:
-		if err := json.Unmarshal(m, &om); err != nil {
-			return nil, "", err
-		}
-		om.MediaType = imgspecv1.MediaTypeImageManifest
-		for i, l := range om.Layers {
-			if l.MediaType == manifest.DockerV2Schema2ForeignLayerMediaType {
-				om.Layers[i].MediaType = imgspecv1.MediaTypeImageLayerNonDistributable
```
mtrmac: This `…NonDistributable` conversion is not present in the new conversion code, is that intentional?

oci/layout/oci_dest.go
```diff
@@ -134,60 +133,25 @@ func (d *ociImageDestination) ReapplyBlob(info types.BlobInfo) (types.BlobInfo,
...
-	ociMan, mt, err := createManifest(m)
-	if err != nil {
-		return err
+	var ociMan imgspecv1.Manifest
+	if err := json.Unmarshal(m, &ociMan); err != nil {
+		return errors.Wrap(err, "error parsing manifest")
+	}
+	// TODO(runcom): not fully sure how to prevent a docker v2s2 from slipping in
+	// any idea?!?!
+	if ociMan.SchemaVersion != 2 {
```
mtrmac: I am tempted no to try at all… But if we should, I think it would be _much_ cleaner to have the caller supply a `mimeType` as a parameter than guessing by parsing contents in an unknown format in here.

image/docker_schema2.go
```diff
@@ -166,13 +167,40 @@ func (m *manifestSchema2) UpdatedImage(options types.ManifestUpdateOptions) (typ
...
+	case imgspecv1.MediaTypeImageManifest:
+		return copy.convertToOCIManifest()
 	default:
 		return nil, errors.Errorf("Conversion of image manifest from %s to %s is not implemented", manifest.DockerV2Schema2MediaType, options.ManifestMIMEType)
 	}
 
 	return memoryImageFromManifest(&copy), nil
 }
 
+func (m *manifestSchema2) convertToOCIManifest() (types.Image, error) {
```
mtrmac: Also, `docker_schema2.go` has (unlike `…schema1.go`) a reasonable test coverage, and it would be nice to preserve that (the conversions happen frequently enough for users to care but perhaps not frequently enough to prevent us from breaking them without noticing). This should be just a pretty trivial variant of `TestConvertToManifestSchema2`.

docker/docker_image_src.go
runcom: makes sense to have v2s1 -> OCI at this point, so yeah, I'll implement this.

oci/layout/oci_dest.go
runcom: I missed it somehow

oci/layout/oci_dest.go
runcom: let's skip this altogether for now and re-iterate on adding an optional mime type, but then, ppl might cheat on mime type as well no?

image/docker_schema2.go
runcom: yup, have it on my todo, just wanted to bring this PR up for general consensus on the way I'm fixing this.

docker/docker_image_src.go
mtrmac: Considering the failing tests, could we merge a version without that conversion now, and deal with the v2s1→OCI conversion as a separate issue or PR?

oci/layout/oci_dest.go
mtrmac: A lying caller will get what is coming to it :)

Right now we _do_ do this wrong, because `copy.determineManifestConversion` can decide to keep an unsupported type _anyway_ (it is treating `SupportedManifestMIMETypes` as a best-effort request), and `PutManifest` isn’t told about that. But we _would_ be sending the correct MIME type to `PutManifest` if there was a way to do that.

docker/docker_image_src.go
runcom: sure

runcom: wait, tests will fail w/o this mime type check, shall I leave this here?

oci/layout/oci_dest.go
runcom: fixed

docker/docker_image_src.go
mtrmac: What is the failure / code path?

runcom: basic skopeo copy docker://busybox oci:something, we need this to ask for at least v2s2 if we want OCI conversion (since we cannot convert from v2s1 which is the default when the registry doesn't understand the media type, see the comment, the test failure is in skopeo tests)

mtrmac: > What is the failure / code path?

Let me guess: `ociDest.SupportedManifestMIMETypes` returns OCI only, and so `copy.Image`  asks for unsupported manifest MIME type, and that gets sent to the registry.

Adding the OCI conversion is the long-term fix.

Short-term, can we:
* move the array from `dockerImageDestinationSupportedManifestMIMETypes` into a global variable (shared by src and dest)
* in `newImageSource`, ignore `requestedManifestMIMETypes` if it has no value common with the known supported MIME types (as defined by the variable from above)?

That’s still kind of a hack, but at least it is conceptually clean: `dockerImage`* knows what it supports/accepts, and does not need to know about OCI.

mtrmac: (\*shrug\*, if you want to merge it with _the existing_ hack instead, feel free. Just please make sure to file an issue to have this cleaned up in the future.)

runcom:
@mtrmac PTAL (warning, wrote it really fast just to prove it)

runcom:
No, wait, we need the conversion, otherwise, the old `createManifest` in `oci_dest.go` can't work since OCI isn't defining the MediaType field anymore...

docker/docker_image_dest.go
```diff
@@ -17,6 +17,21 @@ import (
...
 )
 
+var manifestMIMETypes = []string{
+	// TODO(runcom): we'll add OCI as part of another PR here
+	manifest.DockerV2Schema2MediaType,
+	manifest.DockerV2Schema1SignedMediaType,
+	manifest.DockerV2Schema1MediaType,
+}
+
+func supportedManifestMIMETypesMap() map[string]bool {
```
mtrmac: I think a `map[string]struct{}` is more idiomatic, but I really don’t care enough.

docker/docker_image_src.go
```diff
@@ -39,6 +39,12 @@ func newImageSource(ctx *types.SystemContext, ref dockerReference, requestedMani
 	if requestedManifestMIMETypes == nil {
 		requestedManifestMIMETypes = manifest.DefaultRequestedManifestMIMETypes
 	}
+	for _, mtrequested := range requestedManifestMIMETypes {
+		if !supportedManifestMIMETypesMap()[mtrequested] {
```
mtrmac: This is still O(N^2): the call `supportedManifestMIMETypesMap` outside the loop, save it in a variable. Then it is O(N).

(Or don’t bother and just write two nested loops. The arrays are small enough.)

docker/docker_image_src.go
```diff
@@ -39,6 +39,12 @@ func newImageSource(ctx *types.SystemContext, ref dockerReference, requestedMani
...
+			requestedManifestMIMETypes = manifest.DefaultRequestedManifestMIMETypes
+			break
```
mtrmac: This would ignore the input value if _any_ value is unrecognized; I think it should ignore input only if _all_ values are unrecognized (so that a caller can say “i prefer v2s1, then OCI, then v2s2 as a last resort”.)

docker/docker_image_dest.go
runcom: arg, that's how it was, then changed my mind, I'll put it back with `struct{}`

docker/docker_image_src.go
runcom: > This is still O(N^2): move the call supportedManifestMIMETypesMap outside the loop, save it in a variable. Then it is O(N).

silly me, ETOOLATE

docker/docker_image_dest.go
runcom: on a second though, I like comparing with just `!` - we change this any time though...

runcom:
@mtrmac PTAL

docker/docker_image_dest.go
mtrmac: Sure, whatever works.

docker/docker_image_src.go
```diff
@@ -39,6 +39,19 @@ func newImageSource(ctx *types.SystemContext, ref dockerReference, requestedMani
...
 	}
+	var (
+		mts   = supportedManifestMIMETypesMap()
+		reset = true
+	)
+	for _, mtrequested := range requestedManifestMIMETypes {
+		if !mts[mtrequested] {
+			continue
+		}
+		reset = false
```
mtrmac: `break` here, and then it’s more natural to reword as `if mts[mtrequested] { `**EDIT reset = false;**` break } `.

docker/docker_image_src.go
```diff
@@ -39,6 +39,19 @@ func newImageSource(ctx *types.SystemContext, ref dockerReference, requestedMani
 	if requestedManifestMIMETypes == nil {
 		requestedManifestMIMETypes = manifest.DefaultRequestedManifestMIMETypes
 	}
+	var (
+		mts   = supportedManifestMIMETypesMap()
+		reset = true
```
mtrmac: (and as long you are changing this) this would benefit from either clearer variable names, or a comment explicitly describing what this loop intends to do. Preferably the former, perhaps naming them `supportedMIMEs` and `acceptableRequestedMIMEs` (with the opposite semantics to `reset`, starting as `false`).

mtrmac: …also, no need for the `var()` here AFAICS, just use `:=`. Or does that have an idiomatic meaning?

runcom:
> WRT merging, do you want to include #225 in this so that we have a single PR which passes tests and fixes the build?

yeah, I'll cherry-pick #225 here

runcom:
@mtrmac I've cherry-picked #225 here and hopefully addressed your comments, PTAL

docker/docker_image_src.go
```diff
@@ -39,6 +39,20 @@ func newImageSource(ctx *types.SystemContext, ref dockerReference, requestedMani
...
+	var (
+		acceptableRequestedMIMEs = true
+		supportedMIMEs           = supportedManifestMIMETypesMap()
+	)
+	for _, mtrequested := range requestedManifestMIMETypes {
+		if supportedMIMEs[mtrequested] {
+			continue
+		}
+		acceptableRequestedMIMEs = false
+		break
```
mtrmac: This rejects `requestedManifestMIMETypes` if _any_ content is unrecognized; shouldn’t it reject them only if _none_ are recognized?

mtrmac: (i.e. when reaming `reset` to `acceptableRequestedMIMEs`, only the last `if` was flipped AFAICS.)

runcom: I believe I've fixed it now - sorry I'm in a meeting but want to get this in asap to unblock other PRs

mtrmac: I was thinking
```go
	supportedMIMEs := supportedManifestMIMETypesMap()
	acceptableRequestedMIMEs := false
	for _, mtrequested := range requestedManifestMIMETypes {
		if supportedMIMEs[mtrequested] {
			acceptableRequestedMIMEs = true
			break
		}
	}
	if !acceptableRequestedMIMEs {
		requestedManifestMIMETypes = manifest.DefaultRequestedManifestMIMETypes
	}
```

Perhaps I am being an idiot?

runcom: fixed and push forced

runcom:
another issue just popped up https://travis-ci.org/containers/image/builds/195004166#L346 - perhaps image-spec shouldn't really have a vendor directory _at all_.

mtrmac:
> another issue just popped up https://travis-ci.org/containers/image/builds/195004166#L346 - perhaps image-spec shouldn't really have a vendor directory _at all_. Commented here https://github.com/containers/image/pull/223

( https://github.com/opencontainers/image-spec/issues/527 , for the record.)

mtrmac:
:+1: pending tests

runcom:
@mtrmac will we wait till the image-spec-vendor-dir is gone before merging this?

mtrmac:
> @mtrmac will we wait till the image-spec-vendor-dir is gone before merging this?

I dunno, what other options do we have? Just keep our old checkouts on our development machines, and ignore Travis and keep merging other PRs? Ewww.

Or, give up and set up `containers/image` with our own vendoring mechanism? That would work _for us_ but I worry that it would make it even more difficult for third parties to use the library, as conflicts would inevitably happen.

Is there a chance that the vendoring will be fixed quickly?

runcom:
> Is there a chance that the vendoring will be fixed quickly?

not sure, I hope it'll be really soon

runcom:
seems like something else in OCI image-spec broke this..

runcom:
alright, pretty sure I've this sorted out now, @mtrmac PTAL

image/docker_schema2.go
```diff
@@ -166,13 +167,40 @@ func (m *manifestSchema2) UpdatedImage(options types.ManifestUpdateOptions) (typ
...
+	// media type of the manifest is handled by manifestSchema2FromComponents.
+	config.MediaType = imgspecv1.MediaTypeImageConfig
+
+	layers := make([]descriptor, len(m.LayersDescriptors))
+	for idx := range layers {
+		layers[idx] = m.LayersDescriptors[idx]
+		if m.LayersDescriptors[idx].MediaType == manifest.DockerV2Schema2ForeignLayerMediaType {
+			layers[idx].MediaType = imgspecv1.MediaTypeImageLayerNonDistributable
+		} else {
+			layers[idx].MediaType = imgspecv1.MediaTypeImageLayerGzip
```
mtrmac: This works for now (i.e. at it was perfectly valid a week ago), longer-term we should actually know whether the input is gzipped.  Filed https://github.com/containers/image/issues/229.

runcom:
alright let's merge this (and take care of this comment https://github.com/containers/image/issues/229#issuecomment-275837725 later)

runcom:
:+1:

mtrmac:
:+1:

PR #226, opened 2017-01-24
Fix reference to k8s homedir package

erikh:
This patch just replaces the k8s library with the one in
opencontainers/storage. Tests still pass and we can just fix that much easier
if we need to.

### DIFF
--- a/openshift/openshift-copies.go
+++ b/openshift/openshift-copies.go
@@ -15,12 +15,12 @@ import (
 	"reflect"
 	"time"
 
+	"github.com/containers/storage/pkg/homedir"
 	"github.com/ghodss/yaml"
 	"github.com/imdario/mergo"
 	"github.com/pkg/errors"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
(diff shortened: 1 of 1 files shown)


erikh:
Sorry I didn't see #225, if the k8s lib is still required feel free to close this.

mtrmac:
Yeah, it would be nice to use a single package; sadly, the two have a bit different semantics. #225 tries to preserve the k8s semantics, but we’ll see whether that is acceptable for #207.

So let’s keep this open for a bit. Thanks for working on this!

erikh:
If you want, I could make the changes to storage to accommodate the k8s
homedir needs so we only have one dep.

mtrmac:
At this point it seems preferable for the OpenShift transport to clearly use OpenShift semantics, and for containers/storage to clearly use container/storage semantics, whatever they are and however they differ (or not).

Unifying the dependency would be nice in principle, but honestly the packages are small enough that not having to think about how they differ from the respective upstreams and whether/why that matters is simpler. (And they _do_ matter _somehow_, otherwise everyone would just be doing `os.user.Current().HomeDir` instead of creating these packages.)

PR #227, opened 2017-01-24
Canonicalize docker references in signature.VerifyDockerManifestSignature

mtrmac:
After https://github.com/containers/image/pull/220, and especially future https://github.com/containers/image/pull/221, signing `docker/distribution/reference.Named.String()` would use the new
fully-expanded normalization (as opposed to `containers/image/docker/reference.Named.String()`, which is minimized).

For interoperability between various versions and signers, parse and normalize the expected and signed references before comparing them.

This should be equivalent to `prmMatchExact.matchesDockerReference().`

---

It is also fairly tempting to use a `reference.Named` instead of a `string` as the parameters to the public functions in `docker.go`; however, the `Named` interface does not really guarantee what kind of canonicalization is used. (_Right now_, due to the extra methods in `containers/image/docker/reference.Named`, it is impossible to substitute an upstream `docker/distribution/reference.Named` implementation, but #221 would again make that untrue.) For now I have leaned towards not changing things.


runcom:
LGTM

@mtrmac do we wait for #223 to be fully sorted out or just get this in now? I mean, feel free to bring this is

mtrmac:
:+1:

PR #228, opened 2017-01-27
ostree: new transport

giuseppe:
skopeo copy docker://busybox ostree:foo:latest@/ostree/repo

### DIFF
--- a/ostree/ostree_dest.go
+++ b/ostree/ostree_dest.go
@@ -1,0 +1,248 @@
+package ostree
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strconv"
+	"strings"
+
+	"github.com/containers/image/manifest"
+	"github.com/containers/image/types"
+	"github.com/containers/storage/pkg/archive"
+	"github.com/opencontainers/go-digest"
+	"github.com/pkg/errors"
+)
+
+type blobToImport struct {
+	Size     int64
+	Digest   digest.Digest
+	BlobPath string
+}
+
+type descriptor struct {
+	Size   int64         `json:"size"`
+	Digest digest.Digest `json:"digest"`
+}
+
+type manifestSchema struct {
+	ConfigDescriptor  descriptor   `json:"config"`
+	LayersDescriptors []descriptor `json:"layers"`
+}
+
+type ostreeImageDestination struct {
+	ref        ostreeReference
+	manifest   string
+	schema     manifestSchema
+	tmpDirPath string
+	blobs      map[string]*blobToImport
+}
+
+// newImageDestination returns an ImageDestination for writing to an existing ostree.
+func newImageDestination(ref ostreeReference, tmpDirPath string) (types.ImageDestination, error) {
+	tmpDirPath = filepath.Join(tmpDirPath, ref.branchName)
+	if err := ensureDirectoryExists(tmpDirPath); err != nil {
+		return nil, err
+	}
+	return &ostreeImageDestination{ref, "", manifestSchema{}, tmpDirPath, map[string]*blobToImport{}}, nil
+}
+
+// Reference returns the reference used to set up this destination.  Note that this should directly correspond to user's intent,
+// e.g. it should use the public hostname instead of the result of resolving CNAMEs or following redirects.
+func (d *ostreeImageDestination) Reference() types.ImageReference {
+	return d.ref
+}
+
+// Close removes resources associated with an initialized ImageDestination, if any.
+func (d *ostreeImageDestination) Close() error {
+	return os.RemoveAll(d.tmpDirPath)
+}
+
+func (d *ostreeImageDestination) SupportedManifestMIMETypes() []string {
+	return []string{
+		manifest.DockerV2Schema2MediaType,
+	}
+}
+
+// SupportsSignatures returns an error (to be displayed to the user) if the destination certainly can't store signatures.
+// Note: It is still possible for PutSignatures to fail if SupportsSignatures returns nil.
+func (d *ostreeImageDestination) SupportsSignatures() error {
+	return nil
+}
+
+// ShouldCompressLayers returns true iff it is desirable to compress layer blobs written to this destination.
+func (d *ostreeImageDestination) ShouldCompressLayers() bool {
+	return false
+}
+
+// AcceptsForeignLayerURLs returns false iff foreign layers in manifest should be actually
+// uploaded to the image destination, true otherwise.
+func (d *ostreeImageDestination) AcceptsForeignLayerURLs() bool {
+	return false
+}
+
+func (d *ostreeImageDestination) PutBlob(stream io.Reader, inputInfo types.BlobInfo) (types.BlobInfo, error) {
+	tmpDir, err := ioutil.TempDir(d.tmpDirPath, "blob")
+	if err != nil {
+		return types.BlobInfo{}, err
+	}
+
+	blobPath := filepath.Join(tmpDir, "content")
+	blobFile, err := os.Create(blobPath)
+	if err != nil {
+		return types.BlobInfo{}, err
+	}
+	defer blobFile.Close()
+
+	digester := digest.Canonical.Digester()
+	tee := io.TeeReader(stream, digester.Hash())
+
+	size, err := io.Copy(blobFile, tee)
+	if err != nil {
+		return types.BlobInfo{}, err
+	}
+	computedDigest := digester.Digest()
+	if inputInfo.Size != -1 && size != inputInfo.Size {
+		return types.BlobInfo{}, errors.Errorf("Size mismatch when copying %s, expected %d, got %d", computedDigest, inputInfo.Size, size)
+	}
+	if err := blobFile.Sync(); err != nil {
+		return types.BlobInfo{}, err
+	}
+
+	hash := computedDigest.Hex()
+	d.blobs[hash] = &blobToImport{Size: size, Digest: computedDigest, BlobPath: blobPath}
+	return types.BlobInfo{Digest: computedDigest, Size: size}, nil
+}
+
+func (d *ostreeImageDestination) importBlob(blob *blobToImport) error {
+	ostreeBranch := fmt.Sprintf("ociimage/%s", blob.Digest.Hex())
+	destinationPath := filepath.Join(d.tmpDirPath, blob.Digest.Hex(), "root")
+	if err := ensureDirectoryExists(destinationPath); err != nil {
+		return err
+	}
+	defer func() {
+		os.Remove(blob.BlobPath)
+		os.RemoveAll(destinationPath)
+	}()
+
+	err := archive.UntarPath(blob.BlobPath, destinationPath)
+	if err != nil {
+		return err
+	}
+	return exec.Command("ostree", "commit",
+		"--repo", d.ref.repo,
+		fmt.Sprintf("--add-metadata-string=docker.size=%d", blob.Size),
+		"--branch", ostreeBranch,
+		fmt.Sprintf("--tree=dir=%s", destinationPath)).Run()
+}
+
+func (d *ostreeImageDestination) importConfig(blob *blobToImport) error {
+	ostreeBranch := fmt.Sprintf("ociimage/%s", blob.Digest.Hex())
+
+	return exec.Command("ostree", "commit",
+		"--repo", d.ref.repo,
+		fmt.Sprintf("--add-metadata-string=docker.size=%d", blob.Size),
+		"--branch", ostreeBranch, filepath.Dir(blob.BlobPath)).Run()
+}
+
+func (d *ostreeImageDestination) HasBlob(info types.BlobInfo) (bool, int64, error) {
+	branch := fmt.Sprintf("ociimage/%s", info.Digest.Hex())
+	output, err := exec.Command("ostree", "show", "--repo", d.ref.repo, "--print-metadata-key=docker.size", branch).CombinedOutput()
+	if err != nil {
+		if bytes.Index(output, []byte("not found")) >= 0 || bytes.Index(output, []byte("No such")) >= 0 {
+			return false, -1, nil
+		}
+		return false, -1, err
+	}
+	size, err := strconv.ParseInt(strings.Trim(string(output), "'\n"), 10, 64)
+	if err != nil {
+		return false, -1, err
+	}
+
+	return true, size, nil
+}
+
+func (d *ostreeImageDestination) ReapplyBlob(info types.BlobInfo) (types.BlobInfo, error) {
+	return info, nil
+}
+
+func (d *ostreeImageDestination) PutManifest(manifest []byte) error {
+	d.manifest = string(manifest)
+
+	if err := json.Unmarshal(manifest, &d.schema); err != nil {
+		return err
+	}
+
+	manifestPath := filepath.Join(d.tmpDirPath, d.ref.manifestPath())
+	if err := ensureParentDirectoryExists(manifestPath); err != nil {
+		return err
+	}
+
+	return ioutil.WriteFile(manifestPath, manifest, 0644)
+}
+
+func (d *ostreeImageDestination) PutSignatures(signatures [][]byte) error {
+	path := filepath.Join(d.tmpDirPath, d.ref.signaturePath(0))
+	if err := ensureParentDirectoryExists(path); err != nil {
+		return err
+	}
+
+	for i, sig := range signatures {
+		signaturePath := filepath.Join(d.tmpDirPath, d.ref.signaturePath(i))
+		if err := ioutil.WriteFile(signaturePath, sig, 0644); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (d *ostreeImageDestination) Commit() error {
+	for _, layer := range d.schema.LayersDescriptors {
+		hash := layer.Digest.Hex()
+		blob := d.blobs[hash]
+		// if the blob is not present in d.blobs then it is already stored in OSTree,
+		// and we don't need to import it.
+		if blob == nil {
+			continue
+		}
+		err := d.importBlob(blob)
+		if err != nil {
+			return err
+		}
+	}
+
+	hash := d.schema.ConfigDescriptor.Digest.Hex()
+	blob := d.blobs[hash]
+	if blob != nil {
+		err := d.importConfig(blob)
+		if err != nil {
+			return err
+		}
+	}
+
+	manifestPath := filepath.Join(d.tmpDirPath, "manifest")
+	err := exec.Command("ostree", "commit",
+		"--repo", d.ref.repo,
+		fmt.Sprintf("--add-metadata-string=docker.manifest=%s", string(d.manifest)),
+		fmt.Sprintf("--branch=ociimage/%s", d.ref.branchName),
+		manifestPath).Run()
+	return err
+}
+
+func ensureDirectoryExists(path string) error {
+	if _, err := os.Stat(path); err != nil && os.IsNotExist(err) {
+		if err := os.MkdirAll(path, 0755); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func ensureParentDirectoryExists(path string) error {
+	return ensureDirectoryExists(filepath.Dir(path))
+}
--- a/ostree/ostree_transport.go
+++ b/ostree/ostree_transport.go
@@ -1,0 +1,235 @@
+package ostree
+
+import (
+	"bytes"
+	"fmt"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+
+	"github.com/pkg/errors"
+
+	"github.com/containers/image/directory/explicitfilepath"
+	"github.com/containers/image/docker/reference"
+	"github.com/containers/image/transports"
+	"github.com/containers/image/types"
+)
+
+const defaultOSTreeRepo = "/ostree/repo"
+
+// Transport is an ImageTransport for ostree paths.
+var Transport = ostreeTransport{}
+
+type ostreeTransport struct{}
+
+func (t ostreeTransport) Name() string {
+	return "ostree"
+}
+
+func init() {
+	transports.Register(Transport)
+}
+
+// ValidatePolicyConfigurationScope checks that scope is a valid name for a signature.PolicyTransportScopes keys
+// (i.e. a valid PolicyConfigurationIdentity() or PolicyConfigurationNamespaces() return value).
+// It is acceptable to allow an invalid value which will never be matched, it can "only" cause user confusion.
+// scope passed to this function will not be "", that value is always allowed.
+func (t ostreeTransport) ValidatePolicyConfigurationScope(scope string) error {
+	sep := strings.Index(scope, ":")
+	if sep < 0 {
+		return errors.Errorf("Invalid ostree: scope %s: Must include a repo", scope)
+	}
+	repo := scope[:sep]
+
+	if !strings.HasPrefix(repo, "/") {
+		return errors.Errorf("Invalid ostree: scope %s: repository must be an absolute path", scope)
+	}
+	cleaned := filepath.Clean(repo)
+	if cleaned != repo {
+		return errors.Errorf(`Invalid ostree: scope %s: Uses non-canonical path format, perhaps try with path %s`, scope, cleaned)
+	}
+
+	// FIXME? In the namespaces within a repo,
+	// we could be verifying the various character set and length restrictions
+	// from docker/distribution/reference.regexp.go, but other than that there
+	// are few semantically invalid strings.
+	return nil
+}
+
+// ostreeReference is an ImageReference for ostree paths.
+type ostreeReference struct {
+	image      string
+	branchName string
+	repo       string
+}
+
+func (t ostreeTransport) ParseReference(ref string) (types.ImageReference, error) {
+	var repo = ""
+	var image = ""
+	s := strings.SplitN(ref, "@/", 2)
+	if len(s) == 1 {
+		image, repo = s[0], defaultOSTreeRepo
+	} else {
+		image, repo = s[0], "/"+s[1]
+	}
+
+	return NewReference(image, repo)
+}
+
+// NewReference returns an OSTree reference for a specified repo and image.
+func NewReference(image string, repo string) (types.ImageReference, error) {
+	// image is not _really_ in a containers/image/docker/reference format;
+	// as far as the libOSTree ociimage/* namespace is concerned, it is more or
+	// less an arbitrary string with an implied tag.
+	// We use the reference.* parsers basically for the default tag name in
+	// reference.TagNameOnly, and incidentally for some character set and length
+	// restrictions.
+	var ostreeImage reference.Named
+	s := strings.SplitN(image, ":", 2)
+
+	named, err := reference.WithName(s[0])
+	if err != nil {
+		return nil, err
+	}
+
+	if len(s) == 1 {
+		ostreeImage = reference.TagNameOnly(named)
+	} else {
+		ostreeImage, err = reference.WithTag(named, s[1])
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	resolved, err := explicitfilepath.ResolvePathToFullyExplicit(repo)
+	if err != nil {
+		// With os.IsNotExist(err), the parent directory of repo is also not existent;
+		// that should ordinarily not happen, but it would be a bit weird to reject
+		// references which do not specify a repo just because the implicit defaultOSTreeRepo
+		// does not exist.
+		if os.IsNotExist(err) && repo == defaultOSTreeRepo {
+			resolved = repo
+		} else {
+			return nil, err
+		}
+	}
+	// This is necessary to prevent directory paths returned by PolicyConfigurationNamespaces
+	// from being ambiguous with values of PolicyConfigurationIdentity.
+	if strings.Contains(resolved, ":") {
+		return nil, errors.Errorf("Invalid OSTreeCI reference %s@%s: path %s contains a colon", image, repo, resolved)
+	}
+
+	return ostreeReference{
+		image:      ostreeImage.String(),
+		branchName: encodeOStreeRef(ostreeImage.String()),
+		repo:       resolved,
+	}, nil
+}
+
+func (ref ostreeReference) Transport() types.ImageTransport {
+	return Transport
+}
+
+// StringWithinTransport returns a string representation of the reference, which MUST be such that
+// reference.Transport().ParseReference(reference.StringWithinTransport()) returns an equivalent reference.
+// NOTE: The returned string is not promised to be equal to the original input to ParseReference;
+// e.g. default attribute values omitted by the user may be filled in in the return value, or vice versa.
+// WARNING: Do not use the return value in the UI to describe an image, it does not contain the Transport().Name() prefix.
+func (ref ostreeReference) StringWithinTransport() string {
+	return fmt.Sprintf("%s@%s", ref.image, ref.repo)
+}
+
+// DockerReference returns a Docker reference associated with this reference
+// (fully explicit, i.e. !reference.IsNameOnly, but reflecting user intent,
+// not e.g. after redirect or alias processing), or nil if unknown/not applicable.
+func (ref ostreeReference) DockerReference() reference.Named {
+	return nil
+}
+
+func (ref ostreeReference) PolicyConfigurationIdentity() string {
+	return fmt.Sprintf("%s:%s", ref.repo, ref.image)
+}
+
+// PolicyConfigurationNamespaces returns a list of other policy configuration namespaces to search
+// for if explicit configuration for PolicyConfigurationIdentity() is not set.  The list will be processed
+// in order, terminating on first match, and an implicit "" is always checked at the end.
+// It is STRONGLY recommended for the first element, if any, to be a prefix of PolicyConfigurationIdentity(),
+// and each following element to be a prefix of the element preceding it.
+func (ref ostreeReference) PolicyConfigurationNamespaces() []string {
+	s := strings.SplitN(ref.image, ":", 2)
+	if len(s) != 2 { // Coverage: Should never happen, NewReference above ensures ref.image has a :tag.
+		panic(fmt.Sprintf("Internal inconsistency: ref.image value %q does not have a :tag", ref.image))
+	}
+	name := s[0]
+	res := []string{}
+	for {
+		res = append(res, fmt.Sprintf("%s:%s", ref.repo, name))
+
+		lastSlash := strings.LastIndex(name, "/")
+		if lastSlash == -1 {
+			break
+		}
+		name = name[:lastSlash]
+	}
+	return res
+}
+
+// NewImage returns a types.Image for this reference, possibly specialized for this ImageTransport.
+// The caller must call .Close() on the returned Image.
+// NOTE: If any kind of signature verification should happen, build an UnparsedImage from the value returned by NewImageSource,
+// verify that UnparsedImage, and convert it into a real Image via image.FromUnparsedImage.
+func (ref ostreeReference) NewImage(ctx *types.SystemContext) (types.Image, error) {
+	return nil, errors.New("Reading ostree: images is currently not supported")
+}
+
+// NewImageSource returns a types.ImageSource for this reference,
+// asking the backend to use a manifest from requestedManifestMIMETypes if possible.
+// nil requestedManifestMIMETypes means manifest.DefaultRequestedManifestMIMETypes.
+// The caller must call .Close() on the returned ImageSource.
+func (ref ostreeReference) NewImageSource(ctx *types.SystemContext, requestedManifestMIMETypes []string) (types.ImageSource, error) {
+	return nil, errors.New("Reading ostree: images is currently not supported")
+}
+
+// NewImageDestination returns a types.ImageDestination for this reference.
+// The caller must call .Close() on the returned ImageDestination.
+func (ref ostreeReference) NewImageDestination(ctx *types.SystemContext) (types.ImageDestination, error) {
+	var tmpDir string
+	if ctx == nil || ctx.OSTreeTmpDirPath == "" {
+		tmpDir = os.TempDir()
+	} else {
+		tmpDir = ctx.OSTreeTmpDirPath
+	}
+	return newImageDestination(ref, tmpDir)
+}
+
+// DeleteImage deletes the named image from the registry, if supported.
+func (ref ostreeReference) DeleteImage(ctx *types.SystemContext) error {
+	return errors.Errorf("Deleting images not implemented for ostree: images")
+}
+
+var ostreeRefRegexp = regexp.MustCompile(`^[A-Za-z0-9.-]$`)
+
+func encodeOStreeRef(in string) string {
+	var buffer bytes.Buffer
+	for i := range in {
+		sub := in[i : i+1]
+		if ostreeRefRegexp.MatchString(sub) {
+			buffer.WriteString(sub)
+		} else {
+			buffer.WriteString(fmt.Sprintf("_%02X", sub[0]))
+		}
+
+	}
+	return buffer.String()
+}
+
+// manifestPath returns a path for the manifest within a ostree using our conventions.
+func (ref ostreeReference) manifestPath() string {
+	return filepath.Join("manifest", "manifest.json")
+}
+
+// signaturePath returns a path for a signature within a ostree using our conventions.
+func (ref ostreeReference) signaturePath(index int) string {
+	return filepath.Join("manifest", fmt.Sprintf("signature-%d", index+1))
+}
--- a/types/types.go
+++ b/types/types.go
@@ -299,6 +299,8 @@ type SystemContext struct {
 	// Note that this field is used mainly to integrate containers/image into projectatomic/docker
 	// in order to not break any existing docker's integration tests.
 	DockerDisableV1Ping bool
+	// Directory to use for OSTree temporary files
+	OSTreeTmpDirPath string
 }
 
 // ProgressProperties is used to pass information from the copy code to a monitor which
--- a/ostree/ostree_transport_test.go
+++ b/ostree/ostree_transport_test.go
@@ -1,0 +1,316 @@
+package ostree
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strings"
+	"testing"
+
+	"path/filepath"
+
+	"github.com/containers/image/types"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	sha256digestHex = "0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef"
+	sha256digest    = "@sha256:" + sha256digestHex
+)
+
+func TestTransportName(t *testing.T) {
+	assert.Equal(t, "ostree", Transport.Name())
+}
+
+// A helper to replace $TMP in a repo path with a real temporary directory
+func withTmpDir(repo string, tmpDir string) string {
+	return strings.Replace(repo, "$TMP", tmpDir, -1)
+}
+
+// A common list of repo suffixes to test for the various ImageReference methods.
+var repoSuffixes = []struct{ repoSuffix, resolvedRepo string }{
+	{"", "/ostree/repo"},
+	{"@/ostree/repo", "/ostree/repo"}, // /ostree/repo is accepted even if neither /ostree/repo nor /ostree exists, as a special case.
+	{"@$TMP/at@sign@repo", "$TMP/at@sign@repo"},
+	// Rejected as ambiguous: /repo:with:colons could either be an (/repo, with:colons) policy configuration identity, or a (/repo:with, colons) policy configuration namespace.
+	{"@$TMP/repo:with:colons", ""},
+}
+
+// A common list of cases for image name parsing and normalization
+var imageNameTestcases = []struct{ input, normalized, branchName string }{
+	{"busybox:notlatest", "busybox:notlatest", "busybox_3Anotlatest"},                                                  // Explicit tag
+	{"busybox", "busybox:latest", "busybox_3Alatest"},                                                                  // Default tag
+	{"docker.io/library/busybox:latest", "docker.io/library/busybox:latest", "docker.io_2Flibrary_2Fbusybox_3Alatest"}, // A hierarchical name
+	{"UPPERCASEISINVALID", "", ""},                                                                                     // Invalid input
+	{"busybox" + sha256digest, "", ""},                                                                                 // Digested references are not supported (parsed as invalid repository name)
+	{"busybox:invalid+tag", "", ""},                                                                                    // Invalid tag value
+	{"busybox:tag:with:colons", "", ""},                                                                                // Multiple colons - treated as a tag which contains a colon, which is invalid
+	{"", "", ""},                                                                                                       // Empty input is rejected (invalid repository.Named)
+}
+
+func TestTransportParseReference(t *testing.T) {
+	tmpDir, err := ioutil.TempDir("", "ostreeParseReference")
+	require.NoError(t, err)
+	defer os.RemoveAll(tmpDir)
+
+	for _, c := range imageNameTestcases {
+		for _, suffix := range repoSuffixes {
+			fullInput := c.input + withTmpDir(suffix.repoSuffix, tmpDir)
+			ref, err := Transport.ParseReference(fullInput)
+			if c.normalized == "" || suffix.resolvedRepo == "" {
+				assert.Error(t, err, fullInput)
+			} else {
+				require.NoError(t, err, fullInput)
+				ostreeRef, ok := ref.(ostreeReference)
+				require.True(t, ok, fullInput)
+				assert.Equal(t, c.normalized, ostreeRef.image, fullInput)
+				assert.Equal(t, c.branchName, ostreeRef.branchName, fullInput)
+				assert.Equal(t, withTmpDir(suffix.resolvedRepo, tmpDir), ostreeRef.repo, fullInput)
+			}
+		}
+	}
+}
+
+func TestTransportValidatePolicyConfigurationScope(t *testing.T) {
+	for _, scope := range []string{
+		"/etc:docker.io/library/busybox:notlatest", // This also demonstrates that two colons are interpreted as repo:name:tag.
+		"/etc:docker.io/library/busybox",
+		"/etc:docker.io/library",
+		"/etc:docker.io",
+		"/etc:repo",
+		"/this/does/not/exist:notlatest",
+	} {
+		err := Transport.ValidatePolicyConfigurationScope(scope)
+		assert.NoError(t, err, scope)
+	}
+
+	for _, scope := range []string{
+		"/colon missing as a path-reference delimiter",
+		"relative/path:busybox",
+		"/double//slashes:busybox",
+		"/has/./dot:busybox",
+		"/has/dot/../dot:busybox",
+		"/trailing/slash/:busybox",
+	} {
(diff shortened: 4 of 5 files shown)


ostree/ostree_dest.go
```diff
@@ -0,0 +1,245 @@
...
+}
+
+// AcceptsForeignLayerURLs returns false iff foreign layers in manifest should be actually
+// uploaded to the image destination, true otherwise.
+func (d *ostreeImageDestination) AcceptsForeignLayerURLs() bool {
+	return false
+}
+
+func (d *ostreeImageDestination) PutBlob(stream io.Reader, inputInfo types.BlobInfo) (types.BlobInfo, error) {
+	hash := strings.Replace(inputInfo.Digest.Hex(), "sha256:", "", 1)
```
mtrmac: Isn’t there a native method to get the hex value without the algorithm ID? And this does need a check that the hash actually uses SHA256.

mtrmac: Also note that the input digest may be `nil`. An initial implementation may just fail in that case, fully general would be to create a temporary file, and use the `computedDigest` value.

ostree/ostree_dest.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+	computedDigest := digester.Digest()
+	if inputInfo.Size != -1 && size != inputInfo.Size {
+		return types.BlobInfo{}, errors.Errorf("Size mismatch when copying %s, expected %d, got %d", computedDigest, inputInfo.Size, size)
+	}
+	if err := blobFile.Sync(); err != nil {
+		return types.BlobInfo{}, err
+	}
+
+	d.blobs[hash] = &blobToImport{Size : size, Digest : inputInfo.Digest, BlobPath : blobPath, Hash : hash}
```
mtrmac: Should this use `computedDigest` instead of `inputInfo.Digest`, which may be unknown?

(Also, this mentions a digest of the same object three times, is that truly necessary?)

ostree/ostree_dest.go
```diff
@@ -0,0 +1,245 @@
...
+	}()
+
+	err := archive.UntarPath(blob.BlobPath, destinationPath)
+	if err != nil {
+		return err
+	}
+	return exec.Command("ostree", "commit",
+		fmt.Sprintf("--add-metadata-string=docker.size=%d", blob.Size),
+		fmt.Sprintf("--branch=%s", ostreeBranch),
+		fmt.Sprintf("--tree=dir=%s", destinationPath)).Run()
```
mtrmac: (Just to be sure, is the untarred filesystem at `destinationPath` supposed to exist after this command returns, or is it supposed to be temporary?)

mtrmac: Ah, it is deleted in `.Commit`. A comment to that effect here would be nice.

ostree/ostree_dest.go
```diff
@@ -0,0 +1,245 @@
...
+func (d *ostreeImageDestination) Commit() error {
+	defer func() {
+		os.RemoveAll(encodeOStreeRef(d.ref.image))
+	}()
+
+	for _, layer := range d.schema.LayersDescriptors {
+		hash := strings.Replace(layer.Digest.Hex(), "sha256:", "", 1)
+		blob := d.blobs[hash]
+		if blob == nil {
+			continue
```
mtrmac: Silently ignoring failures seems unsafe

ostree/ostree_dest.go
```diff
@@ -0,0 +1,245 @@
...
+		}
+		err := d.importBlob(true, blob)
```
mtrmac: The two paths in `importBlob`, based on `isTar`, are so different, and each is called exactly once with a constant value of `isTar`; it might be cleaner to just have two functions (`importLayer`, `importConfig`?).

ostree/ostree_dest.go
```diff
@@ -0,0 +1,245 @@
...
+	return nil
+}
+
+func (d *ostreeImageDestination) Commit() error {
+	defer func() {
+		os.RemoveAll(encodeOStreeRef(d.ref.image))
+	}()
+
+	for _, layer := range d.schema.LayersDescriptors {
+		hash := strings.Replace(layer.Digest.Hex(), "sha256:", "", 1)
```
mtrmac: (Same comments as in `PutBlob`)

ostree/ostree_dest.go
```diff
@@ -0,0 +1,245 @@
...
+
+	return true, size, nil
+}
+
+func (d *ostreeImageDestination) ReapplyBlob(info types.BlobInfo) (types.BlobInfo, error) {
+	return info, nil
+}
+
+func (d *ostreeImageDestination) PutManifest(manifest []byte) error {
+	d.manifest = strings.Replace(string(manifest), "\n", " ", -1)
```
mtrmac: Is this necessary? For signature verification it would be _very_ useful to be able to get the original (no byte modified) manifest.

ostree/ostree_src.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+// Close removes resources associated with an initialized ImageSource, if any.
+func (s *ostreeImageSource) Close() {
+}
+
+// GetManifest returns the image's manifest along with its MIME type (which may be empty when it can't be determined but the manifest is available).
+// It may use a remote (= slow) service.
+func (s *ostreeImageSource) GetManifest() ([]byte, string, error) {
+	return nil, "", errors.Errorf(`Getting manifest not supported by "ostree:"`)
```
mtrmac: Given this, it might be simpler to just fail in `NewImageSource`.

ostree/ostree_transport.go
```diff
@@ -0,0 +1,181 @@
...
+func NewReference(path string) (types.ImageReference, error) {
+	var repo = ""
+	var image = ""
+	s := strings.SplitN(path, "@", 2)
+	if len(s) == 1 {
+		image, repo = s[0], "/ostree/repo"
+	} else {
+		image, repo = s[0], s[1]
+	}
+	return ostreeReference{ref: path, repo: repo, image: image}, nil
```
mtrmac: Conventionally we use `NewReference` for calls from Go code which knows the native values, passed as separate parameters, and `ParseReference` for calls which get a single user-specified string input and parse it into the native Go values.

mtrmac: It would be fine (and perhaps preferable) to only store the real values of the components of the reference (repo, image) and not the original string.

ostree/ostree_transport.go
```diff
@@ -0,0 +1,181 @@
...
+
+// PolicyConfigurationIdentity returns a string representation of the reference, suitable for policy lookup.
+// This MUST reflect user intent, not e.g. after processing of third-party redirects or aliases;
+// The value SHOULD be fully explicit about its semantics, with no hidden defaults, AND canonical
+// (i.e. various references with exactly the same semantics should return the same configuration identity)
+// It is fine for the return value to be equal to StringWithinTransport(), and it is desirable but
+// not required/guaranteed that it will be a valid input to Transport().ParseReference().
+// Returns "" if configuration identities for these references are not supported.
+func (ref ostreeReference) PolicyConfigurationIdentity() string {
+	return ref.ref
```
mtrmac: `PolicyConfigurationIdentity` really should be in a “canonical” form, and fully explicit (in particular dealing with the `@/ostree/repo` defaulting.

ostree/ostree_transport.go
```diff
@@ -0,0 +1,181 @@
...
+	res := []string{}
+	path := ref.ref
+	for {
+		lastSlash := strings.LastIndex(path, "/")
+		if lastSlash == -1 || lastSlash == 0 {
+			break
+		}
+		path = path[:lastSlash]
+		res = append(res, path)
+	}
```
mtrmac: This does not really make sense; for `whatever@/my/repo` this would return `whatever@/my`, `whatever@`.

Designing the hierarchy here so that it makes sense for uses is basically the major outstanding action item for `ostreeReference`/`ostreeTransport`.  (It would be possible to not have any parent namespaces, only a `PolicyConfigurationIdentity`.)

ostree/ostree_dest.go
giuseppe: I added a comment here, if blob is not in d.blobs then it means the layer was already present and we don't need to import it.

giuseppe:
I added a fixup comment ⬆️  to address all your comments. `PutBlob` handles also the case when the input digest is nil

giuseppe:
@mtrmac @runcom is there any other change to address before it could be merged?

I am not 100% happy with using the cli of OSTree, I'll probably rework this in future to use the golang bindings (the golang bindings are not yet stable AFAIK)

ostree/ostree_dest.go
```diff
@@ -0,0 +1,255 @@
...
+	"github.com/pkg/errors"
+)
+
+type blobToImport struct {
+	Size      int64
+	Digest    digest.Digest
+	BlobPath  string
+}
+
+type descriptor struct {
```
runcom: We're duplicating this too many. Worth using the official image-spec descriptor type

ostree/ostree_src.go
runcom: We could get rid of this altogether I guess (the whole ostree src)

ostree/ostree_dest.go
giuseppe: I added a new commit to export Descriptor and use it

ostree/ostree_dest.go
mtrmac: > Isn’t there a native method to get the hex value without the algorithm ID?

That’ method is `.Hex()` already, actually. So, that `strings.Replace()` should never be needed AFAICT; if it is, something is going wrong elsewhere.

image/docker_schema2.go
```diff
@@ -28,33 +28,33 @@ var gzippedEmptyLayer = []byte{
 // gzippedEmptyLayerDigest is a digest of gzippedEmptyLayer
 const gzippedEmptyLayerDigest = digest.Digest("sha256:a3ed95caeb02ffe68cdd9fd84406680ae93d633cb16422d00e8a7c22955b46d4")
 
-type descriptor struct {
+type Descriptor struct {
```
mtrmac: There are different descriptor schemas, and more importantly semantics / content expectations, in various formats (e.g. `OCI` expects different `MediaType` values). Name this perhaps `Schema2Descriptor`, please, to make it clear this is format-specific.

image/docker_schema2.go
```diff
@@ -28,33 +28,33 @@ var gzippedEmptyLayer = []byte{
...
 	MediaType string        `json:"mediaType"`
 	Size      int64         `json:"size"`
 	Digest    digest.Digest `json:"digest"`
 	URLs      []string      `json:"urls,omitempty"`
 }
 
-type manifestSchema2 struct {
+type ManifestSchema2 struct {
```
mtrmac: I am not thrilled about exporting the entire implementation of the private `genericManifest` interface. Could this type be split into public a pure JSON-marhsallable data structure with no methods, and a private type implementing `genericManifest`, probably embedding that pure data structure?

ostree/ostree_dest.go
mtrmac: (I’m sorry, this has been addressed already.)

ostree/ostree_dest.go
```diff
@@ -0,0 +1,246 @@
...
+		err := d.importConfig(blob)
+		if err != nil {
+			return err
+		}
+	}
+
+	manifestPath := filepath.Join(encodeOStreeRef(d.ref.image.String()), "manifest")
+	branch := encodeOStreeRef(d.ref.image.String())
+	err := exec.Command("ostree", "commit",
+		fmt.Sprintf("--add-metadata-string=docker.manifest=%s", string(d.manifest)),
```
mtrmac: `d.manifest` is already a string, no need to cast it here. (Or perhaps make `d.manifest` a `[]byte`, which is more natural, and keep the cast here. Either works.

ostree/ostree_dest.go
mtrmac: There isn’t any check to ensure that the layer was already present. It might also just not have been uploaded, perhaps because the caller is doing it wrong.

But, ultimately, the caller is free to upload a broken image and does not have to be told that it is incomplete, so, this works.

ostree/ostree_dest.go
```diff
@@ -0,0 +1,246 @@
...
+// AcceptsForeignLayerURLs returns false iff foreign layers in manifest should be actually
+// uploaded to the image destination, true otherwise.
+func (d *ostreeImageDestination) AcceptsForeignLayerURLs() bool {
+	return false
+}
+
+func (d *ostreeImageDestination) PutBlob(stream io.Reader, inputInfo types.BlobInfo) (types.BlobInfo, error) {
+	parentDir := encodeOStreeRef(d.ref.image.String())
+
+	err := ensureDirectoryExists(parentDir)
```
mtrmac: This creates a subdirectory in the user’s current working directory!?!

ostree/ostree_dest.go
```diff
@@ -0,0 +1,246 @@
...
+
+	blobPath := filepath.Join(tmpDir, "content")
+
+	blobFile, err := os.Create(blobPath)
+	if err != nil {
+		return types.BlobInfo{}, err
+	}
+	defer func() {
+		blobFile.Close()
+	}()
```
mtrmac: This can be written as `defer blobFile.Close()` .

ostree/ostree_dest.go
```diff
@@ -0,0 +1,246 @@
...
+
+	hash := computedDigest.Hex()
+	d.blobs[hash] = &blobToImport{Size : size, Digest : computedDigest, BlobPath : blobPath}
+	return types.BlobInfo{Digest: computedDigest, Size: size}, nil
+}
+
+func (d *ostreeImageDestination) importBlob(blob *blobToImport) (error) {
+	ostreeBranch := fmt.Sprintf("ociimage/%s", blob.Digest.Hex())
+	destinationPath := filepath.Join(encodeOStreeRef(d.ref.image.String()), blob.Digest.Hex(), "root")
+	if err := ensureDirectoryExists(destinationPath); err != nil {
```
mtrmac: This creates a subdirectory in the user’s current working directory!?!

ostree/ostree_dest.go
```diff
@@ -0,0 +1,246 @@
...
+
+func (d *ostreeImageDestination) PutManifest(manifest []byte) error {
+	d.manifest = string(manifest)
+
+	if err := json.Unmarshal(manifest, &d.schema); err != nil {
+		return err
+	}
+
+	manifestPath := d.ref.manifestPath()
+	if err := ensureParentDirectoryExists(manifestPath); err != nil {
```
mtrmac: This creates a subdirectory in the user’s current working directory!?!

ostree/ostree_dest.go
```diff
@@ -0,0 +1,246 @@
...
+		return err
+	}
+
+	return ioutil.WriteFile(manifestPath, manifest, 0644)
+}
+
+func (d *ostreeImageDestination) PutSignatures(signatures [][]byte) error {
+	if err := ensureParentDirectoryExists(d.ref.signaturePath(0)); err != nil {
```
mtrmac: This creates a subdirectory in the user’s current working directory!?!

ostree/ostree_dest.go
```diff
@@ -0,0 +1,246 @@
...
+		if err := ioutil.WriteFile(signaturePath, sig, 0644); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (d *ostreeImageDestination) Commit() error {
+	defer func() {
+		os.RemoveAll(encodeOStreeRef(d.ref.image.String()))
```
mtrmac: So, all of `encodeOStreeRef(d.ref.image.String())` is a temporary directory until commit. (Abstracting from its location), shouldn’t it actually be deleted in `ostreeImageDestination.Close`, so that the data does not lay around if the caller aborts and never calls `Commit`?

ostree/ostree_dest.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+
+	hash := d.schema.ConfigDescriptor.Digest.Hex()
+	blob := d.blobs[hash]
+	if blob != nil {
+		err := d.importConfig(blob)
+		if err != nil {
+			return err
+		}
+	}
```
mtrmac: Actually, is there a benefit to deferring the import to this point at all? `PutBlob` could store the data directly, perhaps without an on-disk temporary copy of the tarball sized in hundreds of megabytes. I suppose the difficulty would be in differentiating layer tarballs and the config?

This _does_ work fine, just a suggestion to consider whether that makes sense.

ostree/ostree_transport.go
```diff
@@ -0,0 +1,171 @@
+package ostree
```
mtrmac: (Due to the impact on signatures we generally want as close to 100% code coverage for the `ImageTransport` and `ImageReference` implementations as possible, but that will obviously wait until the formats and semantics are nailed down—and I can write that.)

ostree/ostree_transport.go
```diff
@@ -0,0 +1,171 @@
...
+	return buffer.String()
+}
+
+// manifestPath returns a path for the manifest within a ostree using our conventions.
+func (ref ostreeReference) manifestPath() string {
+	return filepath.Join(encodeOStreeRef(ref.image.String()), "manifest", "manifest.json")
+}
+
+// layerPath returns a path for a layer tarball within a ostree using our conventions.
+func (ref ostreeReference) layerPath(digest digest.Digest) string {
```
mtrmac: This is unused.

ostree/ostree_transport.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+// DeleteImage deletes the named image from the registry, if supported.
+func (ref ostreeReference) DeleteImage(ctx *types.SystemContext) error {
+	return errors.Errorf("Deleting images not implemented for ostree: images")
+}
+
+var ostreeRefRegexp = regexp.MustCompile(`^[A-Za-z0-9.-]$`)
+
+func encodeOStreeRef(in string) string {
```
mtrmac: This is always called as `encodeOStreeRef(ref.image.String())`. It seems that this should be a method on `ImageReference`, like e.g. `manifestPath` below.

ostree/ostree_dest.go
giuseppe: yes, since the content in the ostree repo cannot be modified once written, we need to store it with the correct SELinux label.  The correct SELinux label is what is assigned once checked-out to `/var/lib/containers/atomic`, we can run skopeo in a temporary directory in `/var/lib/containers/atomic` so that the files are properly labelled before being committed.

mtrmac: Users of skopeo and callers of the library will not, in general, know that they should do us this favor. (Even if they knew to, they may well want to keep the existing working directory, in order to use relative paths for other transports!)

Please add a `const` for the `/var/lib/containers/atomic` path, and explicitly create the temporary directory in this package. Or does this package perhaps not have enough information to set the SELinux context?

ostree/ostree_transport.go
```diff
@@ -0,0 +1,171 @@
...
+	if cleaned != scope {
+		return errors.Errorf(`Invalid scope %s: Uses non-canonical format, perhaps try %s`, scope, cleaned)
+	}
+	return nil
+}
+
+// ostreeReference is an ImageReference for ostree paths.
+type ostreeReference struct {
+	ref          string
+	repo         string
```
giuseppe: thanks for the catch, repo must be passed to each invocation of ostree, as --repo=$REPO

ostree/ostree_dest.go
giuseppe: we are not sure the container will be checked out to `/var/lib/containers/atomic`.  This location can be changed in atomic, and also when `--user` is specified, then the container is checked out to the user home directory.

ostree/ostree_dest.go
giuseppe: yes, I was storing the file immediately before, but then I changed to do it here as we know what must be stored as a tarball and what instead in the metadata.  We could try to detect if the file is a .tar in the PutBlob function, but I think doing it here is cleaner (at the expense of keeping them)

image/docker_schema2.go
runcom: haven't fully looked at the code, but why do we need this type to be exported _at all_? is it because ostree pkg needs it? we could move this under `manifest/` I guess

giuseppe: @runcom yes, I am exporting this type so that it could be used by OSTree, as it needs to parse the manifest to get the config and the layers

runcom: Ack, we'll, maybe we really need to export those in the manifest pkg (not as part of this PR). @mtrmac wdyt?

giuseppe: @runcom is it then fine if I drop the second patch from this PR and leave the OSTree refactoring for later?

ostree/ostree_dest.go
```diff
@@ -131,6 +131,7 @@ func (d *ostreeImageDestination) importBlob(blob *blobToImport) (error) {
 		return err
 	}
 	return exec.Command("ostree", "commit",
+		fmt.Sprintf("--repo=%s", d.ref.repo),
```
mtrmac: BTW `ostree` seems to also accept separate parameters without the `=`e.g. `[…, "--repo`, d.ref.repo, …]`. That would allow getting rid of all of these `fmt.Sprintf` calls.

ostree/ostree_dest.go
mtrmac: I know very little about ostree; is there anywhere I can read up on this issue?

What you seem to be saying is that at the time of storing an image into the ostree repo, the writer must use the “correct” SELinux types, but the writer does not know where the reader will check out the image, .e. the writer can not know what the “correct” types are?! If so, how is atomic better positioned to know this?

image/docker_schema2.go
mtrmac: Having a single schema2 data structure in the `manifest` package, used by all of `docker/daemon`, `docker_schema1.go`, `docker_schema2.go`, and the autodetection in `manifest/manifest.go`, does sound great.

mtrmac: (This _can_ be a separate PR, but if we are doing this, let’s actually do that instead of making the `image/docker_schema2.go` version public now.)

giuseppe:
atomic knows where the files will be checked out since `checkout_path` can be overriden in the `atomic.conf` file.

When we commit a directory to OSTree, OSTree stores each file by its checksum, the checksum includes also the xattrs of the file, this is why we need to store it with the correct SELinux labels.  When we extract the .tar of a layer, we need to do that to a directory where we know it will get the right SELinux labels.

We *could* use `/var/lib/containers/atomic` for the temporary directory (as this directory is used in containers-selinux) fo files to get the right labels, but I'd prefer that we are not going to hardcode this path in `containers/image` as well:  1) the atomic tool will still be the owner of that directory and its subdirectories  2) keep the freedom to change it (if necessary) from atomic and containers-selinux only.

@rhatdan what do you think of this?

Using the current working directory seemed like a good compromise to me, otherwise, should we introduce an environment variable to decide where to keep temporary files?

rhatdan:
I think we might be conflating two different concepts here. OSTRee used for the atomic host so that paths like /usr/bin/docker is labeled container_runtime_t, versus paths in a container for a system container, which probably have no label associated with them. Their is nothing in an OCI image to identify the label of a file path.  Pretty much we treat the entire content of a container as a single label
usually something that looks like system_u:object_r:container_file_t:s0:c1,c2.

In the case of system containers it will just be the default label of /var/lib/containers.

giuseppe:
@rhatdan yes, exactly it will be the label of /var/lib/containers.  The decision here to take is to leave to atomic the goal of creating the correct temporary directory and run skopeo from there, or manage the temporary directory from containers/image itself.

non root users can also pull images to their OSTree repository, not shared with the system.  So we will have to handle this case as well, as these users have no write access to /var/lib/containers.  There is already the logic in atomic to do this, I'd prefer to leave it there and run skopeo in the correct location instead of duplicating and maintaining this logic here as well.

@mtrmac is fine to keep creating the tmp files in the current working directory or is it a blocker?  Would an environment variable work fine?

rhatdan:
I think the higher level concepts should stay in atomic command and just use PWD for tmp files.

mtrmac:
> The decision here to take is to leave to atomic the goal of creating the correct temporary directory and run skopeo from there, or manage the temporary directory from containers/image itself.

To step back here, I think containers/image should ideally work stand-alone without the caller having to care about implementation details of the individual transports.

`atomic` very likely won’t remain the only caller forever: e.g. if we ever set up cri-o to use ostree, we will need this to work without running an `atomic` command, and it would be nice if cri-o didn’t have to know anything about the ostree implementation details. And it is pretty much a non-starter for a long-running daemon like cri-o to modify per-process (not even per-thread!) global state like the current working directory or environment variables.

If the only cost of this were having to parse `atomic.conf` in `ostreeImageDestination`, that seems very much worth it to me.

---

> Using the current working directory seemed like a good compromise to me,

This can fail pretty horribly in general; it does not even have to be writable! It may be on a tmpfs and cause the machine run out of memory.

> otherwise, should we introduce an environment variable to decide where to keep temporary files?

We have the `types.SystemContext` mechanism for passing transport-specific options and overrides, and that would be more suitable than per-process environment variables.

---

My preference would be (but I may _very well_ be misunderstanding the problem) to get the temporary directory path by:

1. (Optionally) first checking `types.SystemContext.OSTreeTemporaryPath`, if it is defined. (Or should that be `OSTreeSELinuxContext`, to be explicit about what we care about instead of indirectly expressing this as paths?)
2. Otherwise, if defined and available, parse `atomic.conf` for the `checkout_path` value.
3. Otherwise use a hard-coded default.

OTOH if you want to keep the current behavior, extracting directly into a temporary directory, I can live with this being an `atomic`-specific hack slightly less horrible than `skopeo layers` extracting data into the current directory, which is also an `atomic`-specific hack; at least we would get the signature support, and hopefully we can improve on this later.

mtrmac:
> OTOH if you want to keep the current behavior, extracting directly into a temporary directory, I can live with this being an `atomic`-specific hack slightly less horrible than `skopeo layers` extracting data into the current directory, which is also an `atomic`-specific hack; at least we would get the signature support, and hopefully we can improve on this later.

This would, at the very least, need a very explicit documentation about this behavior and the expectations on the caller somewhere.

giuseppe:
@mtrmac, I agree that using the CWD probably is not the best thing to do when using `containers/image` as a library.  I was looking at it only as a replacement for the current `skopeo layers`.  I am fine to change this, just one question, in atomic we will still use Skopeo as a CLI tool, not like a library.

How to specify the backend specific data (`types.SystemContext`) when using `skopeo copy`?

mtrmac:
For `skopeo copy` it would be a new command-line option, the way `SignBy` or `RemoveSignatures` are now.

(As I said above, I can live with using the current working directory, it is still an improvement; just not nearly as big as I’d like.)

giuseppe:
I added `tmpDirPath' to SystemContext and I use this directory for the temporary files

giuseppe:
@mtrmac @runcom is this fine to go?  I'll start working on the Skopeo changes as soon as it is merged

ostree/ostree_transport.go
mtrmac: Still outstanding. I’ll take a stab at this.

ostree/ostree_transport.go
```diff
@@ -0,0 +1,171 @@
...
+	return fmt.Sprintf("%s@%s", ref.image.String(), ref.repo)
+}
+
+// PolicyConfigurationNamespaces returns a list of other policy configuration namespaces to search
+// for if explicit configuration for PolicyConfigurationIdentity() is not set.  The list will be processed
+// in order, terminating on first match, and an implicit "" is always checked at the end.
+// It is STRONGLY recommended for the first element, if any, to be a prefix of PolicyConfigurationIdentity(),
+// and each following element to be a prefix of the element preceding it.
+func (ref ostreeReference) PolicyConfigurationNamespaces() []string {
+	return policyconfiguration.DockerReferenceNamespaces(ref.image)
```
mtrmac: Still outstanding — what is the conceptual hierarchy of `.repo` vs. `.image`? Is an image stored in a repo, a repo in an image, are they completely orthogonal?

ostree/ostree_transport.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// NewImage returns a types.Image for this reference, possibly specialized for this ImageTransport.
+// The caller must call .Close() on the returned Image.
+// NOTE: If any kind of signature verification should happen, build an UnparsedImage from the value returned by NewImageSource,
+// verify that UnparsedImage, and convert it into a real Image via image.FromUnparsedImage.
+func (ref ostreeReference) NewImage(ctx *types.SystemContext) (types.Image, error) {
+	return nil, nil
```
mtrmac: This needs to return a non-`nil` error.

ostree/ostree_transport.go
mtrmac: (Making `encodeOSTreeRef` a method the way `manifestPath` would be more consistent, but this does work fine.)

ostree/ostree_transport.go
```diff
@@ -0,0 +1,171 @@
...
+			buffer.WriteString(fmt.Sprintf("_%02X", sub[0]))
+		}
+
+	}
+	return buffer.String()
+}
+
+// manifestPath returns a path for the manifest within a ostree using our conventions.
+func (ref ostreeReference) manifestPath() string {
+	return filepath.Join(encodeOStreeRef(ref.image.String()), "manifest", "manifest.json")
```
mtrmac: ping?

ostree/ostree_dest.go
mtrmac: ping?

ostree/ostree_dest.go
```diff
@@ -0,0 +1,255 @@
...
+}
+
+// AcceptsForeignLayerURLs returns false iff foreign layers in manifest should be actually
+// uploaded to the image destination, true otherwise.
+func (d *ostreeImageDestination) AcceptsForeignLayerURLs() bool {
+	return false
+}
+
+func (d *ostreeImageDestination) PutBlob(stream io.Reader, inputInfo types.BlobInfo) (types.BlobInfo, error) {
+	parentDir := filepath.Join(d.tmpDirPath, encodeOStreeRef(d.ref))
```
mtrmac: If we only create files under this `parentDir`, it might be cleaner to compute that value once in the constructor, and rework `signaturePath`/`manifestPath` not to start with `encodeOStreeRef`…; that way it would be immediately obvious that the final `os.RemoveAll` really cleans up everything without having to peek into `signaturePath`/`manifestPath`.

This does work fine, it’s just a bit cumbersome to read.

ostree/ostree_transport.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// ostreeReference is an ImageReference for ostree paths.
+type ostreeReference struct {
+	ref   string
+	repo  string
+	image reference.Named
+}
+
+func withDefaultTag(ref reference.Named) reference.Named {
```
mtrmac: This is now (again) available as `reference.WithDefaultTag`.

types/types.go
```diff
@@ -292,6 +292,8 @@ type SystemContext struct {
 	// Note that this field is used mainly to integrate containers/image into projectatomic/docker
 	// in order to not break any existing docker's integration tests.
 	DockerDisableV1Ping bool
+	// Directory to use for temporary files
+	TmpDirPath string
```
mtrmac: For the public value, please name it `OSTreeTmpDirPath` or something like that; it has a specific OSTree-specific purpose (SELinux labels) and we do not use it in general.

(Having a global override of the `temporaryDirectoryForBigFiles` constant _could_ be useful, but the point of this one is the specific OSTree semantics, not a replacement for `$TMPDIR`.)

ostree/ostree_transport.go
```diff
@@ -0,0 +1,160 @@
...
+// nil requestedManifestMIMETypes means manifest.DefaultRequestedManifestMIMETypes.
+// The caller must call .Close() on the returned ImageSource.
+func (ref ostreeReference) NewImageSource(ctx *types.SystemContext, requestedManifestMIMETypes []string) (types.ImageSource, error) {
+	return nil, errors.New("Reading ostree: images is currently not supported")
+}
+
+// NewImageDestination returns a types.ImageDestination for this reference.
+// The caller must call .Close() on the returned ImageDestination.
+func (ref ostreeReference) NewImageDestination(ctx *types.SystemContext) (types.ImageDestination, error) {
+	return newImageDestination(ref, ctx.TmpDirPath), nil
```
mtrmac: This needs to somehow handle `ctx == nil || ctx.TmpDirPath == ""`.

ostree/ostree_transport.go
mtrmac: … not yet, this is blocked on the repo/image hierarchy and naming decision.

ostree/ostree_transport.go
giuseppe: An image is stored in a repository, the repository usually is `/ostree/repo` but when used by non root users, we have a separate ostree repository for each user in their home directory.  I've switched it to be repo@image now, that should be enough to indicate a specific image in a repo

ostree/ostree_transport.go
giuseppe: was it already merged?  Can't see it on master

mtrmac: My fault, I was looking at an older branch. With master as of 4f14180c47e7c5611069c060d056fab8739619e1 it is `reference.TagNameOnly`.

ostree/ostree_transport.go
mtrmac: Thanks for the explanation. Please consider also the other items above; at the very least the “prefix by repo`@`” part of 2.ii.

ostree/ostree_transport.go
giuseppe: I pushed a new version.  I changed the repo/image hierarchy separator to be ':' as I found '@' to be confusing (we already use @ for "image@repo")

ostree/ostree_dest.go
```diff
@@ -0,0 +1,252 @@
...
+		signaturePath := filepath.Join(d.tmpDirPath, d.ref.signaturePath(i))
+		if err := ioutil.WriteFile(signaturePath, sig, 0644); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (d *ostreeImageDestination) Commit() error {
+	defer os.RemoveAll(filepath.Join(d.tmpDirPath, encodeOStreeRef(d.ref)))
```
mtrmac: Shouldn’t this be just `os.RemoveAll(d.tmpDirPath)` now?

ostree/ostree_dest.go
```diff
@@ -0,0 +1,252 @@
...
+	hash := d.schema.ConfigDescriptor.Digest.Hex()
+	blob := d.blobs[hash]
+	if blob != nil {
+		err := d.importConfig(blob)
+		if err != nil {
+			return err
+		}
+	}
+
+	manifestPath := filepath.Join(d.tmpDirPath, encodeOStreeRef(d.ref), "manifest")
```
mtrmac: Doesn’t this include `encodeOStreeRef()` twice now?

ostree/ostree_transport.go
```diff
@@ -0,0 +1,177 @@
...
+			buffer.WriteString(fmt.Sprintf("_%02X", sub[0]))
+		}
+
+	}
+	return buffer.String()
+}
+
+// manifestPath returns a path for the manifest within a ostree using our conventions.
+func (ref ostreeReference) manifestPath() string {
+	return filepath.Join(encodeOStreeRef(ref), "manifest", "manifest.json")
```
mtrmac: Shouldn’t this be without the `encodeOStreeRef()` now?

ostree/ostree_transport.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+// signaturePath returns a path for a signature within a ostree using our conventions.
+func (ref ostreeReference) signaturePath(index int) string {
+	return filepath.Join(encodeOStreeRef(ref), "manifest", fmt.Sprintf("signature-%d", index+1))
```
mtrmac: Shouldn’t this be without the `encodeOStreeRef()` now?

ostree/ostree_dest.go
mtrmac: This is still outstanding AFAICS

ostree/ostree_transport.go
```diff
@@ -0,0 +1,177 @@
...
+	if len(s) == 1 {
+		imageStr, repo = s[0], "/ostree/repo"
+	} else {
+		imageStr, repo = s[0], s[1]
+	}
+
+	var imageTagged reference.Named
+	s = strings.SplitN(imageStr, ":", 2)
+
+	image, err := reference.WithName(s[0])
```
mtrmac: Shouldn’t this use `reference.ParseNormalizedNamed` like the rest? Why is this explicitly parsing the tag instead of letting `Parse*` handle it?

ostree/ostree_transport.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+// ValidatePolicyConfigurationScope checks that scope is a valid name for a signature.PolicyTransportScopes keys
+// (i.e. a valid PolicyConfigurationIdentity() or PolicyConfigurationNamespaces() return value).
+// It is acceptable to allow an invalid value which will never be matched, it can "only" cause user confusion.
+// scope passed to this function will not be "", that value is always allowed.
+func (t ostreeTransport) ValidatePolicyConfigurationScope(scope string) error {
+	if strings.Index(scope, "@") < 0 {
+		return errors.Errorf("Invalid scope %s: Must include a repo", scope)
+	}
```
mtrmac: This needs updating for the `:`-based format.

ostree/ostree_transport.go
giuseppe: ParseNormalizedNamed is also adds the registry if it is not specified.  This should not be done for system containers.  If I pass "busybox", I don't want it to be resolved to "docker.io/busybox".  Is there a way to not add the registry when it is missing?

giuseppe:
thanks for the review, I've addressed the other issues reported (the `defer func()` one was already fixed a few revisions ago, please consider the last version of the patch).  The tests are failing for a timeout on the registry, is there a way to retrigger them?

giuseppe:
@runcom, @mtrmac I have pushed -f a new version and tests are passing now

ostree/ostree_transport.go
runcom: @giuseppe if someone passes "busybox" what do you want then? what's the format you're interested in after parsing a reference? what if I pass "docker.io/busybox" or "docker.io/library/busybox" do you still want "busybox"? what about private registries? if I pass "rhel.io/busybox", do you still want "busybox"?

ostree/ostree_dest.go
```diff
@@ -0,0 +1,252 @@
...
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strconv"
+	"strings"
+
+	"encoding/json"
```
runcom: leave a new line after this and remove the blank line above this

ostree/ostree_dest.go
```diff
@@ -0,0 +1,252 @@
...
+	manifest   string
+	schema     manifestSchema
+	tmpDirPath string
+	blobs      map[string]*blobToImport
+}
+
+// newImageDestination returns an ImageDestination for writing to an existing ostree.
+func newImageDestination(ref ostreeReference, tmpDirPath string) (types.ImageDestination, error) {
+	tmpDirPath = filepath.Join(tmpDirPath, encodeOStreeRef(ref))
+	err := ensureDirectoryExists(tmpDirPath)
```
runcom: compact this with the if below:
```
if err := ensureDirectoryExists....
```

ostree/ostree_dest.go
```diff
@@ -0,0 +1,252 @@
...
+type manifestSchema struct {
+	ConfigDescriptor  descriptor   `json:"config"`
+	LayersDescriptors []descriptor `json:"layers"`
+}
+
+type ostreeImageDestination struct {
+	ref        ostreeReference
+	manifest   string
+	schema     manifestSchema
+	tmpDirPath string
```
runcom: why is this "tmp"? is this ever cleaned up somewhere?

ostree/ostree_transport.go
```diff
@@ -0,0 +1,177 @@
...
+	return "ostree"
+}
+
+// ValidatePolicyConfigurationScope checks that scope is a valid name for a signature.PolicyTransportScopes keys
+// (i.e. a valid PolicyConfigurationIdentity() or PolicyConfigurationNamespaces() return value).
+// It is acceptable to allow an invalid value which will never be matched, it can "only" cause user confusion.
+// scope passed to this function will not be "", that value is always allowed.
+func (t ostreeTransport) ValidatePolicyConfigurationScope(scope string) error {
+	if strings.Index(scope, ":") < 0 {
+		return errors.Errorf("Invalid scope %s: Must include a repo", scope)
```
runcom: `%q`?

ostree/ostree_dest.go
giuseppe: yes, in Commit() the entire directory is deleted.

ostree/ostree_transport.go
giuseppe: I copied this part from directory/directory_transport.go.  I'll change it to '%q'

runcom: nah, we should have a typed error which formats this string the way I want with %q, for now leave it the same as directory I guess

ostree/ostree_transport.go
giuseppe: no, if you pass the name with a registry, that registry is stored as well, we are only skipping the default to docker.io for system containers:

busybox:foo -> busybox:foo
docker.io/library/busybox -> docker.io/library/busybox:latest
127.0.0.1:5000/busybox:tag -> 127.0.0.1:5000/busybox:tag

giuseppe:
@runcom pushed a new version with the fixes, and the tests pass again :)

runcom:
Looks good to me, @mtrmac?

giuseppe:
any update on this?

rhatdan:
@mtrmac Can we move forward on this?

ostree/ostree_transport.go
mtrmac: (Both using `%q` and using a typed error for `ValidatePolicyConfigurationScope` results could be a nice improvement; `directory_transport.go` should by no means be considered the upper limit on code quality. But neither of these is blocking this PR from my POV.)

ostree/ostree_transport.go
mtrmac: The point is, what is the _semantics_ of hostname-less `busybox`?

That notably affects whether `docker/policyconfiguration` is relevant, and whether `DockerReference` should return the result of parsing, because these interfaces _do_ expect a `reference.Named` with the docker/distribution upstream semantics, i.e. that every reference implicitly contains a hostname, that `docker.io/[library/]` may be implicit / canonicalized, and that `.Name()` and `.String()` return canonical forms which can be compared for equality.

Per the docker/distribution upstream semantics, `busybox` and `docker.io/library/busybox:latest` are two ways to refer to the same image, and `policyConfiguration.DockerReferenceNamespaces(reference.ParseNormalizedNamed(input))` correspondingly returns the same sequence of namespaces (namely [`docker.io/library/busybox`, `docker.io/library`, `docker.io`]). If the ostree image names have a different semantics, such that `busybox` has no logical parent namespace, then the `PolicyConfiguration*` methods should be adjusted accordingly, perhaps avoiding `docker/reference` entirely and maybe doing its own parsing using… `strings.Split`, or its own regexps/rules, I guess.

Also, `.DockerReference()` is used to figure out how to sign images; if that returns a `"busybox"` reference for an ostree image, this value is used to create a signature, consumers of signatures will parse it using `reference.ParseNormalizedNamd` and consider such a signature acceptable when the consumer wants a `docker.io/library/busybox` image. If the semantics of the `busybox` value in ostree is in fact different, then `.DockerReference()` must return the right value, or perhaps `nil` if the image identity can’t be expressed using the upstream docker/distribution `refererence.Named` semantics.

ostree/ostree_transport.go
```diff
@@ -0,0 +1,177 @@
...
+	var imageTagged reference.Named
+	s = strings.SplitN(imageStr, ":", 2)
+
+	image, err := reference.WithName(s[0])
+	if err != nil {
+		return nil, err
+	}
+
+	if len(s) == 1 {
+		imageTagged = reference.TagNameOnly(image)
```
mtrmac: This, or perhaps just a check that a reference `!IsNameOnly` and returning an error if not, fits better into `NewReference` — right now it is possible for an external caller to call `NewReference` with an `IsNameOnly` reference value, violating `docker/policyconfiguration` prerequisites.

(But see also the discussion about reference semantics / normalization elsewhere.)

ostree/ostree_transport.go
```diff
@@ -0,0 +1,177 @@
...
+			return nil, err
+		}
+		imageTagged = image
+	}
+
+	return NewReference(ref, repo, imageTagged)
+}
+
+// NewReference returns an OSTree reference for a specified repo and image.
+func NewReference(reference string, repo string, image reference.Named) (types.ImageReference, error) {
```
mtrmac: It’s fine to allow an arbitrary path as a `repo`, including paths with symlinks or relative paths, but the policy namespaces must consistently refer to the same repo using the same policy namespaces regardless of how the user has chosen to refer to it; please pass `repo` through `explicitfilepath.ResolvePathToFullyExplicit` here before using it in `PolicyConfiguration*`; .

ostree/ostree_transport.go
giuseppe: yes, most of the code in ParseReference doesn't make sense there.  I have moved it to NewReference.

giuseppe:
for signing we should use the same semantics of Docker, since that is the semantic we follow for pulling an image anyway.  The extra code that caused confusion was related only to how the image is stored in OSTree itself.  This part might change in future and work as Docker, but we will need to sync with atomic so for now I've left the same logic of how "atomic pull --storage ostree" works today.  I hope the new version is cleaner as I've moved the OSTree specific bits to their own function.

giuseppe:
I am currently working on adding your tests

giuseppe:
@mtrmac I've cherry picked your commit and done some changes to it

giuseppe:
I rebased the patch on top of origin/master, but I am seeing some regressions in the Makefile, I opened a separate PR:

https://github.com/containers/image/pull/242

giuseppe:
@mitr is the last version better?

I've opened a WIP PR for Skopeo to add the other bits there:

https://github.com/projectatomic/skopeo/pull/314

giuseppe:
ping

giuseppe:
any update on this?  @runcom @rhatdan @mtrmac

rhatdan:
@runcom @mtrmac Can we get this reviewed again?

mtrmac:
(Notes to self, please ignore.

_If_ I am reading `Atomic/syscontainers.py:SystemContainers.{_encode_to_ostree_ref,_convert_to_skopeo}` correctly, `atomic` is currently treating the various Docker-equivalent `busybox` strings as equivalent for **source** pulling, and therefore signature verification (not explicitly, but because `skopeo… docker://does`), but distinct for **destination** ostree storage (except for adding a default `:latest` tag). Is there a semantic commitment or merely an UI which mixes two namespaces into a single string?

Also, `_encode_to_ostree_ref` seems to care about registries but it does not, really, in an observable way; it splits the registry part of a string, if any, but then reconstitutes it into the original form.

What/who ultimately “owns” the OStree `ociimage/` (`syscontainers.py:OSTREE_OCIIMAGE_PREFIX`) namespace and defines the rules?)

mtrmac:
> for signing we should use the same semantics of Docker, since that is the semantic we follow for pulling an image anyway. The extra code that caused confusion was related only to how the image is stored in OSTree itself. This part might change in future and work as Docker, but we will need to sync with atomic so for now I've left the same logic of how "atomic pull --storage ostree" works today. I hope the new version is cleaner as I've moved the OSTree specific bits to their own function.

I can’t quite wrap my head around this, the way I understand it. It’s not really possible for the `busybox` and `docker.io/library/busybox:latest` strings to be equivalent for purposes of signature verification, but distinct for purposes of image storage / referring to images locally. If the images are different then their signatures should not be substituable.

---

@giuseppe Reading through `syscontainers.py` and the surrounding ecosystem, my best guess at the intended interpretation is that _for ostree OCI_, the references are (a sequence of `/`-separated fragments, with a mandatory `:`tag trailer) mapped using `_`$hex$hex escape mechanism into a single libOSTree fragment (i.e. there are no further OSTree-visible `/` separators).

Within the above, the _only_ normalization / equivalence classes are formed by the implied `:latest` suffix. Is that correct?

Also, the namespace used here is _intentionally_ using this encoding and hiding it from the user, i.e. there is no desire to expose all libOSTree names (e.g. those without the `ociimages/` prefix in branch name, or those not using the `_`$hex$hex escaping).  Presumably only images using the `ociimages/`$encoded form are expected to contain the required `docker.manifest` etc. annotations, so exposing the full libOSTree namespace would not actually be desirable. Is that correct?

Within this context, while (`atomic pull `$foo) uses $foo for both pulling from registries (using docker/distribution canonicalization, with either the `docker.io` defaulting, or the `atomic` repository search semantics) and into libOSTree `ociimages/`$encoded (using no repository canonicalization, only the `:latest` default tag), my best guess is that this should be treated as an UI matter (providing a convenient(?) shortcut for the user using a single string for both services), _not_ as a claim that the two services use an equivalent namespace with equivalent rules. That’s at least _a_ possible way to solve the “is equivalent/is not equivalent, at the same time” conundrum. Is that a reasonable way to think about this, or is the intent different?

If so, I can try updating the code in that direction.

---

[Not getting at all into the details of how long names are allowed to be and what are the exact allowed characters in various components and so on; given things like `if len(branch_id) == 64` when interpreting existing branches, while not caring about that at all when creating new branches, it’s just not worth it to try to get a _precise_ spec before this PR is finalized. Let’s just do _something_ and either expanding or restricting these details can happen later as bug reports, inevitably, come in.]

giuseppe:
Yes exactly, we don't want to expose all the OSTree branches as references that are not under the 'ociimage/' namespace.  Images that are not `ociimage/` are neither OCI layers nor storing the manifest metadata which is needed to recreate the image from its layers.  We must handle only the references that are under 'ociimage/'.

Under `ociimage/` we maintain both the metadata for the images (where we just store the manifest info), and the OCI layers that are in the form `ociimage/$checksum`.  The mangling used on the image names is needed to circumvent a limitation in the branch names allowed by OSTree).

We actually store and use the full name of the image because we want to be able to use system containers also when the local Docker engine is not running, so we are not using the atomic search semantic.  We can change this in future, as long as we keep the ability of running without requiring Docker.

mtrmac:
@giuseppe Please take a look at https://github.com/mtrmac/image/tree/ostree-not-dockerlike (built on top of your work, rebased. See the commit messages of the individual commits for more details on what/why; then they can all be squashed.

Does this make sense, and correspond to what you want the semantics to be?

(Warning: Completely untested apart from tests passing.)

giuseppe:
@mtrmac thanks for the extra patches!  I've tested them with Skopeo and everything seems to work fine.  I've only added one patch to correctly register the OSTree transport.

mtrmac:
> I've only added one patch to correctly register the OSTree transport.

What the… I did notice that you have added the registration in your latest branch, and I remember quite strongly starting the work in that branch (and the presence of `ostreeBranchForImage` does suggest it’s the latest variant or something very close).  I’m afraid I have deleted my WIP branches already, so I have absolutely no idea what I have done to screw this up. Anyway, if this works for you, great.

@runcom PTAL.

runcom:
LGTM

mtrmac:
🎉 @giuseppe Please squash the commits and let’s merge this.

giuseppe:
@mtrmac sure, do you prefer a single commit or any other preference?

mtrmac:
One commit would work just fine; I don’t know whether there is any other split you may wish to preserve.

giuseppe:
I've squashed all the commits into a single one ⬆️

mtrmac:
:+1:

runcom:
needs rebase already :(

rhatdan:
Awesome work.

PR #230, opened 2017-01-30
docker: remove github.com/docker/engine-api

runcom:
Takes care of https://github.com/projectatomic/skopeo/issues/239

### DIFF
--- a/docker/daemon/daemon_dest.go
+++ b/docker/daemon/daemon_dest.go
@@ -14,7 +14,7 @@ import (
 	"github.com/containers/image/docker/reference"
 	"github.com/containers/image/manifest"
 	"github.com/containers/image/types"
-	"github.com/docker/engine-api/client"
+	"github.com/docker/docker/client"
 	"github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
 	"golang.org/x/net/context"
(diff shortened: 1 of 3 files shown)


runcom:
Corresponding skopeo PR https://github.com/projectatomic/skopeo/pull/297

runcom:
Tests green in skopeo

docker/daemon/daemon_dest.go
```diff
@@ -95,7 +95,7 @@ func (d *daemonImageDestination) Close() {
 	if !d.committed {
 		logrus.Debugf("docker-daemon: Closing tar stream to abort loading")
 		// In principle, goroutineCancel() should abort the HTTP request and stop the process from continuing.
-		// In practice, though, https://github.com/docker/engine-api/blob/master/client/transport/cancellable/cancellable.go
+		// In practice, though, https://github.com/docker/api/docker/types/blob/master/client/transport/cancellable/cancellable.go
```
mtrmac: That URL does not exist. AFAICS the link should now go to https://github.com/golang/net/blob/master/context/ctxhttp/ctxhttp_pre17.go , _and_ the comment should say that this path is only used with Go < 1.7. It seems that ≥1.7, where `net/http` has a native `context` support, can actually abort sending the request in _some_ cases, but AFAICS not in the “write request” phase (again, it would continue sending the data in the background, only return to the caller early: https://golang.org/src/net/http/transport.go `roundTrip` sends the request to `pc.writech`, and I can’t see that `CancelRequest` aborts the anything in the `writeLoop` goroutine).

mtrmac: (Really rather than doing all this research, just vaguely mentioning that some implementations may in some cases, as of January 2017 … might be good enough. The primary point of the comment is the motivation for the `CloseWithError`.)

runcom: Could you provide the actual comment you want to see here?

mtrmac: I don’t _really_ want to research the code paths in detail right now :) Let me see…
```go
// In principle, goroutineCancel() should abort the HTTP request and stop the process from continuing.
// In practice, though, various HTTP implementations used by client.Client.ImageLoad() (including 
// https://github.com/golang/net/blob/master/context/ctxhttp/ctxhttp_pre17.go and the
// net/http version with native Context support in Go 1.7) do not always actually immediately cancel
// the operation: they may process the HTTP request, or a part of it, to completion in a goroutine, and
// return early if the context is canceled without terminating the goroutine at all.
// So we need…
```
with some reasonable line wrapping.

runcom: fixed I believe

mtrmac:
:+1:

Thanks!

runcom:
:+1:

erikh:
FWIW, I just went through this on box. did you guys test against 1.12 or earlier? I don't think the new client is compatible with it.

PR #231, opened 2017-01-31
Vendor dependencies with vndr

erikh:
This adds vendored dependencies. It also adjusts the tests to avoid the vendor dir.

As mentioned in https://github.com/containers/storage/pull/24 I think this
would best be implemented eventually as vendor being a separate monorepo that
could share deps with both projects, while reducing repo churn. Maybe skopeo
too.

I started running into problems with some build tags so I imported a version of
hcsshim which has explicit build tags. You can see the appropriate ticket here:
https://github.com/Microsoft/hcsshim/pull/106

The fork is mine at erikh/hcsshim and it's a fork off master, should be good.

Please let me know if anything should change.


mtrmac:
Thanks for the PR.

Recently we’ve run into very serious issues trying to use libraries which vendor their dependencies: https://github.com/docker/distribution/issues/2130 and https://github.com/opencontainers/image-spec/issues/527 ; basically the vendored dependencies are not possible to use from callers of the library, without callers of the library using their own vendoring tool which strips the vendored dependencies… and then we might just as well not ship the vendored copies at all because we are again not in control of the versions of the dependencies.

I do agree that this January, when containers/image has been broken by dependencies changing the API about half of the time, has been very painful, but it's not clear to me that this is noticeably better.

(Also, in principle, we do want to be able to use the latest versions, or at least very recent versions, of the dependencies, most of the time; vendoring them would make it too easy to freeze very old versions without noticing. But that could, I suppose, be handled with some process change.)

@runcom ?

erikh:
This is why I'm advocating for the monorepo that's shared amongst ALL
`containers/` projects. Dependencies from other libraries will be easier to
control and prevent conflict.

mtrmac:
> This is why I'm advocating for the monorepo that's shared amongst ALL `containers/` projects. Dependencies

I don’t know, that could make sense (as a git submodule?), though it’s non-obvious that the coordination would work well. Somebody will always want to stay behind because they don’t have time now, and somebody will always want to push ahead for a new nifty feature.

But even if that monorepo / submodule did exist, the above failure cases still suggest that containers/image should not just be shipping with a `vendor` subdirectory: even if both containers/image _and_ its consumer had an _exactly identical_ `vendor` subdirectory from the monorepo, the named types would still not be interoperable! That monorepo would make sense to use probably only as an extra element in `$GOPATH` when running tests. We, kind of, get a similar behavior with `make test-skopeo` which does use the vendored versions (but, sure, we first test against `master` for the internal unit tests(.

erikh:
ok. I'm gonna close this then. You're right, this probably won't fly.

Issue #229, opened 2017-01-28
docker→OCI conversion should differentiate compressed and uncompressed layers


mtrmac:
Right now, in `manifestSchema2.convertToManifestOCI1`, we assume that all input layers are gzipped, which is not generally the case. We should somehow detect this (e.g. `copy.copyLayers` does actually know the true format at some point) and use the correct MIME type.

runcom:
I also noticed we should fix the image config when converting from docker v2s2 to OCI :cry: right now we're blindly just changing docker config media type to OCI config media type but that's wrong because docker v2s2 config has "container_config" which OCI doesn't have. Let's merge #223 and re-iterate ok?

mtrmac:
Yes, that’s why I filed this as a separate issue.

runcom:
> Right now, in manifestSchema2.convertToManifestOCI1, we assume that all input layers are gzipped,

schema2 to oci1 is always gzipped layers afaict - there's no such "no-gzipped" layers in docker v2s2

mtrmac:
> > Right now, in manifestSchema2.convertToManifestOCI1, we assume that all input layers are gzipped,

> schema2 to oci1 is always gzipped layers afaict - there's no such "no-gzipped" layers in docker v2s2

Good point, `dockerd always compresses the layers when pushing; but it does do format autodetection when pulling and supports both uncompressed layers and non-gzip compression formats.

When both us and `dockerd` send layers compressed using gzip, so just assuming gzip should be good enough.

PR #232, opened 2017-01-31
image: fix docker v2s2 -> oci image config conversion

runcom:
Fix https://github.com/containers/image/issues/229

@mtrmac PTAL

### DIFF
--- a/image/docker_schema2.go
+++ b/image/docker_schema2.go
@@ -177,12 +177,27 @@ func (m *manifestSchema2) UpdatedImage(options types.ManifestUpdateOptions) (typ
 }
 
 func (m *manifestSchema2) convertToManifestOCI1() (types.Image, error) {
-	// Create a copy of the descriptor.
-	config := m.ConfigDescriptor
+	configBlob, err := m.ConfigBlob()
+	if err != nil {
+		return nil, err
+	}
+	// docker v2s2 and OCI v1 are mostly compatible but v2s2 contains more fields
+	// than OCI v1. This unmarshal, then re-marshal makes sure we drop docker v2s2
+	// fields that aren't needed in OCI v1.
+	configOCI := &imgspecv1.Image{}
+	if err := json.Unmarshal(configBlob, configOCI); err != nil {
+		return nil, err
+	}
+	configOCIBytes, err := json.Marshal(configOCI)
+	if err != nil {
+		return nil, err
+	}
 
-	// The only difference between OCI and DockerSchema2 is the mediatypes. The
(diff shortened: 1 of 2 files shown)


image/docker_schema2.go
```diff
@@ -177,12 +177,23 @@ func (m *manifestSchema2) UpdatedImage(options types.ManifestUpdateOptions) (typ
 }
 
 func (m *manifestSchema2) convertToManifestOCI1() (types.Image, error) {
-	// Create a copy of the descriptor.
-	config := m.ConfigDescriptor
+	configBlob, err := m.ConfigBlob()
+	if err != nil {
+		return nil, err
+	}
+	configOCI := &imgspecv1.Image{}
```
mtrmac: It’s non-obvious what this does. This would benefit from a comment, saying something like “The two versions of config are mostly compatible, but v2s2 has an extra $name field; this unmarshal+re-marshal effectively drops it.”

image/docker_schema2.go
```diff
@@ -197,7 +208,7 @@ func (m *manifestSchema2) convertToManifestOCI1() (types.Image, error) {
 	// Rather than copying the ConfigBlob now, we just pass m.src to the
 	// translated manifest, since the only difference is the mediatype of
 	// descriptors there is no change to any blob stored in m.src.
```
mtrmac: This comment is no longer true. (Passing `m.src` is fine, just don’t say that we are not copying the `ConfigBlob`.)

image/docker_schema2.go
```diff
@@ -177,12 +177,23 @@ func (m *manifestSchema2) UpdatedImage(options types.ManifestUpdateOptions) (typ
...
+	if err != nil {
+		return nil, err
+	}
 
-	// The only difference between OCI and DockerSchema2 is the mediatypes. The
-	// media type of the manifest is handled by manifestSchema2FromComponents.
+	config := descriptor{}
 	config.MediaType = imgspecv1.MediaTypeImageConfig
+	config.Size = int64(len(configOCIBytes))
+	config.Digest = digest.FromBytes(configOCIBytes)
```
mtrmac: Use a single struct initializer,
```go
config := descriptor{
    MediaType: …,
    Size: …,
…
}
```
?

runcom:
@mtrmac PTAL

mtrmac:
:+1: pending tests.

runcom:
:+1:

runcom:
Wait, I forgot to add a comment on why we assume gzip, fixing it

PR #234, opened 2017-02-01
docker: cache bearer tokens

runcom:
@mtrmac PTAL, just WIP for now as I have to handle tokens refreshing and stuff like that.

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -37,23 +37,33 @@ const (
 	manifestURL   = "%s/manifests/%s"
 	blobsURL      = "%s/blobs/%s"
 	blobUploadURL = "%s/blobs/uploads/"
+
+	minimumTokenLifetimeSeconds = 60
 )
 
 // ErrV1NotSupported is returned when we're trying to talk to a
 // docker V1 registry.
 var ErrV1NotSupported = errors.New("can't talk to a V1 docker registry")
 
+type bearerToken struct {
+	Token     string    `json:"token"`
+	ExpiresIn int       `json:"expires_in"`
+	IssuedAt  time.Time `json:"issued_at"`
+}
+
 // dockerClient is configuration for dealing with a single Docker registry.
 type dockerClient struct {
-	ctx           *types.SystemContext
-	registry      string
-	username      string
-	password      string
-	scheme        string // Cache of a value returned by a successful ping() if not empty
-	client        *http.Client
-	signatureBase signatureStorageBase
-	challenges    []challenge
-	scope         authScope
+	ctx             *types.SystemContext
+	registry        string
+	username        string
+	password        string
+	scheme          string // Cache of a value returned by a successful ping() if not empty
+	client          *http.Client
+	signatureBase   signatureStorageBase
+	challenges      []challenge
+	scope           authScope
+	token           *bearerToken
+	tokenExpiration time.Time
 }
 
 type authScope struct {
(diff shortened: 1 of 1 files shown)


docker/docker_client.go
```diff
@@ -252,6 +253,11 @@ func (c *dockerClient) makeRequestToResolvedURL(method, url string, headers map[
 //
 // debugging: https://github.com/containers/image/pull/211#issuecomment-273426236 and follows up
 func (c *dockerClient) setupRequestAuth(req *http.Request) error {
+	if c.token != "" {
```
mtrmac: Move this into the `case "bearer"` I think.

runcom:
@mtrmac PTAL - followed https://github.com/docker/distribution/blob/master/docs/spec/auth/token.md and what docker does also.

runcom:
copying a multi-layers image shows a gain of around 4-5 seconds:
```sh
# w/o this patch
./skopeo copy docker://registry oci:registry2  0.31s user 0.24s system 2% cpu 21.696 total
# with this patch
./skopeo copy docker://registry oci:registry  0.27s user 0.20s system 2% cpu 17.360 total
```

docker/docker_client.go
```diff
@@ -262,26 +272,31 @@ func (c *dockerClient) setupRequestAuth(req *http.Request) error {
...
-		if !ok {
-			return errors.Errorf("missing realm in bearer auth challenge")
-		}
-		service, _ := challenge.Parameters["service"] // Will be "" if not present
-		scope := fmt.Sprintf("repository:%s:%s", c.scope.remoteName, c.scope.actions)
-		token, err := c.getBearerToken(realm, service, scope)
-		if err != nil {
-			return err
+		now := time.Now()
+		if now.After(c.tokenExpiration) || c.token == nil {
```
mtrmac: Switch the order of the checks; `c.tokenExpiration` is uninitialized if `c.token == nil` (and the `c.token == nil` check is more efficient, but that should not be the decisive criterion).

Is the `now` variable necessary? AFAICS it has a single user.

docker/docker_client.go
```diff
@@ -37,23 +37,33 @@ const (
...
 )
 
 // ErrV1NotSupported is returned when we're trying to talk to a
 // docker V1 registry.
 var ErrV1NotSupported = errors.New("can't talk to a V1 docker registry")
 
+type bearerToken struct {
+	Token     string    `json:"token"`
+	ExpiresIn int       `json:"expires_in"`
+	IssuesAt  time.Time `json:"issued_at"`
```
mtrmac: `s/IssuesAt/IssuedAt/g` ?

docker/docker_client.go
```diff
@@ -300,35 +315,33 @@ func (c *dockerClient) getBearerToken(realm, service, scope string) (string, err
...
+		return nil, err
+	}
+	var token bearerToken
+	if err := json.Unmarshal(tokenBlob, &token); err != nil {
+		return nil, err
+	}
+	if token.ExpiresIn < minimumTokenLifetimeSeconds {
+		token.ExpiresIn = minimumTokenLifetimeSeconds
+		logrus.Debugf("Increasing token expiration to: %d seconds", token.ExpiresIn)
+	}
```
mtrmac: (This works just fine. It could also be worth considering the case when `Issue[sd]At` is in the past, possibly even more than `minimumTokenLifetimeSeconds` in the past. But the current code, in that case, uses the token exactly once, which is a perfectly reasonable behavior. The alternative of extending the token to be valid at least `minimumTokenLifetimeSeconds` seems to be much worse.)

docker/docker_client.go
runcom: fixed thx

docker/docker_client.go
runcom: fixed

runcom:
@mtrmac fixed your comments.

mtrmac:
:+1: Thanks!

(Also, a nice performance improvement.)

[Approved with PullApprove](https://pullapprove.com/containers/image/pull-request/234/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=containers/image)

runcom:
:+1:

rhatdan:
Nice lets get new skopeo builds.

runcom:
@rhatdan just released https://github.com/projectatomic/skopeo/releases/tag/v0.1.18

rhatdan:
Awesome.

mrunalp:
Nice! Can we pull this into cri-o?

runcom:
@mrunalp I will (likely later today or tomorrow morning, we need to update many of the deps in CRI-O I had opened an issue iirc)

PR #235, opened 2017-02-06
Fix go-digest import

erikh:
This fixes an issue with go-digest not be imported properly in a few spots. should resolve any test failures.

### DIFF
--- a/docker/daemon/daemon_src.go
+++ b/docker/daemon/daemon_src.go
@@ -12,7 +12,7 @@ import (
 	"github.com/containers/image/manifest"
 	"github.com/containers/image/types"
 	"github.com/docker/docker/client"
-	"github.com/opencontainers/go-digest"
+	digest "github.com/opencontainers/go-digest"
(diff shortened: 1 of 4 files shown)


mtrmac:
How does the current version fail? Building and tests seem to work fine for me.

docker/daemon/daemon_src.go
```diff
@@ -12,7 +12,7 @@ import (
 	"github.com/containers/image/manifest"
 	"github.com/containers/image/types"
 	"github.com/docker/docker/client"
-	"github.com/opencontainers/go-digest"
+	digest "github.com/opencontainers/go-digest"
```
runcom: `go-digest` package should indeed contain the `digest` package and golang is able to resolve that. Tests run fine for me as well, not sure what you're doing wrong :confused:

erikh:
it wasn't there before. hm.

I'll close this and see if I can resolve it some other way.

PR #236, opened 2017-02-06
Close with error

### LINKED ISSUE #82: `Close()` should be `Close() error`
Anything against having `{Image,ImageSource,ImageDestination}.Close()` implement `io.Closer` https://golang.org/pkg/io/#Closer? (which basically is adding an error return). I think we would have this for https://github.com/containers/image/pull/63 also.

ref: https://github.com/containers/image/blob/master/types/types.go#L98

@mtrmac wdyt?

erikh:
This patch augments the interfaces' close methods with an error return which is
only actually useful in one or two spots for now. However, it can be utilized
later as things fill in.

Fixes #82

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -100,8 +100,14 @@ type Options struct {
 	Progress         chan types.ProgressProperties // Reported to when ProgressInterval has arrived for a single artifact+offset.
 }
 
-// Image copies image from srcRef to destRef, using policyContext to validate source image admissibility.
-func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageReference, options *Options) error {
+// Image copies image from srcRef to destRef, using policyContext to validate
+// source image admissibility.
+func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageReference, options *Options) (retErr error) {
+	// NOTE this function uses an output parameter for the error return value.
+	// Setting this and returning is the ideal way to return an error.
+	//
+	// the defers in this routine will wrap the error return with its own errors
+	// which can be valuable context in the middle of a multi-streamed copy.
 	if options == nil {
 		options = &Options{}
 	}
@@ -120,7 +126,12 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 	if err != nil {
 		return errors.Wrapf(err, "Error initializing destination %s", transports.ImageName(destRef))
 	}
-	defer dest.Close()
+	defer func() {
+		if err := dest.Close(); err != nil {
+			retErr = errors.Wrapf(retErr, " (dest: %v)", err)
+		}
+	}()
+
 	destSupportedManifestMIMETypes := dest.SupportedManifestMIMETypes()
 
 	rawSource, err := srcRef.NewImageSource(options.SourceCtx, destSupportedManifestMIMETypes)
@@ -130,7 +141,9 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 	unparsedImage := image.UnparsedFromSource(rawSource)
 	defer func() {
 		if unparsedImage != nil {
-			unparsedImage.Close()
+			if err := unparsedImage.Close(); err != nil {
+				retErr = errors.Wrapf(retErr, " (unparsed: %v)", err)
+			}
 		}
 	}()
 
@@ -140,10 +153,14 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 	}
 	src, err := image.FromUnparsedImage(unparsedImage)
 	if err != nil {
-		return errors.Wrapf(err, "Error initializing image from source %s", transports.ImageName(srcRef))
+		retErr = errors.Wrapf(err, "Error initializing image from source %s", transports.ImageName(srcRef))
 	}
 	unparsedImage = nil
-	defer src.Close()
+	defer func() {
+		if err := src.Close(); err != nil {
+			retErr = errors.Wrapf(retErr, " (source: %v)", err)
+		}
+	}()
 
 	if src.IsMultiImage() {
 		return errors.Errorf("can not copy %s: manifest contains multiple images", transports.ImageName(srcRef))
--- a/docker/daemon/daemon_dest.go
+++ b/docker/daemon/daemon_dest.go
@@ -91,7 +91,7 @@ func imageLoadGoroutine(ctx context.Context, c *client.Client, reader *io.PipeRe
 }
 
 // Close removes resources associated with an initialized ImageDestination, if any.
-func (d *daemonImageDestination) Close() {
+func (d *daemonImageDestination) Close() error {
 	if !d.committed {
 		logrus.Debugf("docker-daemon: Closing tar stream to abort loading")
 		// In principle, goroutineCancel() should abort the HTTP request and stop the process from continuing.
@@ -107,6 +107,8 @@ func (d *daemonImageDestination) Close() {
 		d.writer.CloseWithError(errors.New("Aborting upload, daemonImageDestination closed without a previous .Commit()"))
 	}
 	d.goroutineCancel()
+
+	return nil
 }
 
 func (d *daemonImageDestination) Reference() types.ImageReference {
--- a/types/types.go
+++ b/types/types.go
@@ -110,7 +110,7 @@ type ImageSource interface {
 	// (not as the image itself, or its underlying storage, claims).  This can be used e.g. to determine which public keys are trusted for this image.
 	Reference() ImageReference
 	// Close removes resources associated with an initialized ImageSource, if any.
-	Close()
+	Close() error
 	// GetManifest returns the image's manifest along with its MIME type (which may be empty when it can't be determined but the manifest is available).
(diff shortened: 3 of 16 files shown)


docker/daemon/daemon_dest.go
```diff
@@ -91,9 +91,11 @@ func imageLoadGoroutine(ctx context.Context, c *client.Client, reader *io.PipeRe
 }
 
 // Close removes resources associated with an initialized ImageDestination, if any.
-func (d *daemonImageDestination) Close() {
+func (d *daemonImageDestination) Close() error {
+	var err error
 	if !d.committed {
 		logrus.Debugf("docker-daemon: Closing tar stream to abort loading")
+		err = errors.New("Aborting upload, daemonImageDestination closed without a previous .Commit()")
```
mtrmac: This is not an error in `.Close()`; if the caller decides to abort without a `.Commit()`, and the upload is successfully aborted, `.Close()` should indicate success.

types/types.go
```diff
@@ -139,7 +139,7 @@ type ImageDestination interface {
 	// e.g. it should use the public hostname instead of the result of resolving CNAMEs or following redirects.
 	Reference() ImageReference
 	// Close removes resources associated with an initialized ImageDestination, if any.
-	Close()
+	Close() error
```
mtrmac: The caller in `copy.Image` should be updated to handle the error.

types/types.go
```diff
@@ -185,7 +185,7 @@ type UnparsedImage interface {
 	// (not as the image itself, or its underlying storage, claims).  This can be used e.g. to determine which public keys are trusted for this image.
 	Reference() ImageReference
 	// Close removes resources associated with an initialized UnparsedImage, if any.
-	Close()
+	Close() error
```
mtrmac: The callers in `copy.Image` (both for `UnparseImage` and `Image`) should be updated to handle the error.

And perhaps `doc.go` as well?

mtrmac:
To link the two, this fixes #82.

erikh:
Will make adjustments, and file a PR on skopeo in the next few days.

erikh:
Ok I updated this to handle close errors, I also changed the return convention (read the comments at the inner-top of copy.Image) to use output variables. All is explained in comments, but if this feels too cumbersome for the benefits we can talk about alternatives.

copy/copy.go
```diff
@@ -94,8 +94,15 @@ type Options struct {
 	DestinationCtx   *types.SystemContext
 }
 
-// Image copies image from srcRef to destRef, using policyContext to validate source image admissibility.
-func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageReference, options *Options) error {
+// Image copies image from srcRef to destRef, using policyContext to validate
+// source image admissibility.
+func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageReference, options *Options) (retErr error) {
+	// XXX this function uses an output parameter for the error return value.
+	// Setting this and returning is the ideal way to return an error.
```
mtrmac: Looking at https://golang.org/ref/spec#Return_statements , even with a named return value it would be perfectly fine to just `return somerrorvalue` AFAICS.

(Also, doesn’t `XXX` mean “this is bad”? Perhaps just `NOTE:` or something like that.)

copy/copy.go
```diff
@@ -106,9 +113,15 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
...
 	dest, err := destRef.NewImageDestination(options.DestinationCtx)
 	if err != nil {
-		return errors.Wrapf(err, "Error initializing destination %s", transports.ImageName(destRef))
+		retErr = errors.Wrapf(err, "Error initializing destination %s", transports.ImageName(destRef))
+		return
 	}
-	defer dest.Close()
+	defer func() {
+		if err := dest.Close(); err != nil {
+			retErr = errors.Wrapf(retErr, "dest: %v", err)
```
mtrmac: This should only happen if `retErr == nil`; the original value of `retErr` was the primary cause we are failing; that `dest.Close()` has failed may be completely unrelated, or it may be a side effect of the original `retErr` failure.

copy/copy.go
```diff
@@ -118,7 +131,9 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 	unparsedImage := image.UnparsedFromSource(rawSource)
 	defer func() {
 		if unparsedImage != nil {
-			unparsedImage.Close()
+			if err := unparsedImage.Close(); err != nil {
+				retErr = errors.Wrapf(retErr, "unparsed: %v", err)
```
mtrmac: Same here

copy/copy.go
```diff
@@ -131,10 +146,15 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 		return errors.Wrapf(err, "Error initializing image from source %s", transports.ImageName(srcRef))
 	}
 	unparsedImage = nil
-	defer src.Close()
+	defer func() {
+		if err := src.Close(); err != nil {
+			retErr = errors.Wrapf(retErr, "source: %v", err)
```
mtrmac: Same here

docker/daemon/daemon_dest.go
mtrmac: This is still outstanding.

copy/copy.go
erikh: Sure. I typically use it to say "WARNING" or something of the like: "pay attention" basically.

copy/copy.go
erikh: ok sure, I was just thinking that maybe it'd be nice to have all the error context in the event of a copy failure.

mtrmac: Oh, sorry, I wasn’t paying enough attention to notice that this preserves both. Let me check…

mtrmac: … something like this would be nice _if we can make that human-readable_. See how the other `errors.Wrapf` calls in here try to tell the user what is going on. I am not sure that this is workable here:
```
errors.Wrapf(retErr, "Error closing destination: %v, after primary failure", err)
```
is bordering on incomprehensible, especially if `err` were long / created by several `errors.Wrapf` calls. (We can make the underlying data type situation better by using something like `aggregateErr` from `openshift-copies.go`, but that is probably even worse for readability.)

If we can’t make that human-readable, I think preserving a non-`nil` `retErr` is more important than the failure from `.Close`.

@runcom ?

erikh: Sure, I get the concern.

What if I did something like this:

`errors.Wrapf(retErr, "(destination error %v) ", err)`

That way each indepednent error would be wrapped by parens. I could also add newlines.

I think the errors will be valuable context in the face of other, higher level errors.

mtrmac: That’s certainly better than my attempt.

The code as is (making `daemonImageSource.Close` fail using an invalid path):
```
$ sudo ./x docker-daemon:busybox:latest dir:d
FATA[0000] Error copying: source: remove makethisfail//var/tmp/docker-daemon-tar266436759: no such file or directory: Error writing blob: open d/dir-put-blob135662858: no such file or directory 
```
quite misleading about the cause/effect relationships

“primary error” wording:
```
FATA[0000] Error copying: Error closing source: remove makethisfail//var/tmp/docker-daemon-tar088311212: no such file or directory, after primary failure: Error writing blob: open d/dir-put-blob812853531: no such file or directory 
```
Still focusing on the wrong thing (closing failure)

With the parentheses:
```
FATA[0000] Error copying: (source error remove makethisfail//var/tmp/docker-daemon-tar460112488: no such file or directory): Error writing blob: open d/dir-put-blob983238823: no such file or directory 
```
Certainly better. Good enough?

Newlines are attractive (they would even allow us to build a bulleted list :) ) but risky, the caller is fairly likely not expecting them.

At this point I am really not sure, I’ll let @runcom weigh in.

copy/copy.go
```diff
@@ -106,9 +113,15 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 
 	dest, err := destRef.NewImageDestination(options.DestinationCtx)
 	if err != nil {
-		return errors.Wrapf(err, "Error initializing destination %s", transports.ImageName(destRef))
+		retErr = errors.Wrapf(err, "Error initializing destination %s", transports.ImageName(destRef))
+		return
```
cyphar: This change is not necessary. `return X` will set `retErr = x` implicitly.

erikh:
PTAL

changes:

* using parens for inner errors now
* retErr code was adjusted based on return semantics.

erikh:
https://github.com/containers/image/pull/240 is blocking this.

mtrmac:
LGTM pending tests. (Shouldn't `test-skopeo` work even after the API change? If not, the way we deal with API interdependencies is usually to

- Temporarily modify the corresponding skopeo PR to vendor from the proposed branch, show that _that_ passes tests
- Then merge the containers/image PR, ignoring the test failure
- Update the skopeo PR to vendor from the updated containers/image master again, and merge it.

mtrmac:
:+1:

erikh:
rebased

runcom:
@erikh sorry needs one last rebase before I merge this

erikh:
done

runcom:
:+1:

PR #237, opened 2017-02-15
oci: remove MIME type autodetection

cyphar:
OCI JSON blobs no longer contain a mediaType (since 1.0.0-rc4), which
breaks the auto-detection code for MIME types. However, it doesn't make
sense for us to be "auto-detecting" the MIME type of a manifest when the
OCI transport *already knows the mediaType* from the reference. So
instead just return the already-computed mediaType from the descriptor.

Unfortunately, this doesn't work for GetTargetManifest, in which case we
just default to the only valid value at the moment. In future we should
extend this logic to read `/index.json` and figure out what descriptor
referenced the manifest (though any such logic would be questionable
without running a full index on all blobs in the source image).

This is part of fixing projectatomic/skopeo#305.

### DIFF
--- a/oci/layout/oci_src.go
+++ b/oci/layout/oci_src.go
@@ -6,7 +6,6 @@ import (
 	"io/ioutil"
 	"os"
 
-	"github.com/containers/image/manifest"
 	"github.com/containers/image/types"
 	"github.com/opencontainers/go-digest"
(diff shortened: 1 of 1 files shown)


runcom:
:+1:

mtrmac:
:+1: thanks!

PR #238, opened 2017-02-18
manifest: remove ManifestList type

erikh:
This was breaking in the latest version of the opencontainers manifest library.

I removed it because the type no longer exists, so I presumed the feature no longer exists, perhaps a hasty assumption. I'll research and fix this up if necessary, but it was a very small patch.

### DIFF
--- a/manifest/manifest.go
+++ b/manifest/manifest.go
@@ -54,7 +54,7 @@ func GuessMIMEType(manifest []byte) string {
 	}
 
 	switch meta.MediaType {
-	case DockerV2Schema2MediaType, DockerV2ListMediaType, imgspecv1.MediaTypeImageManifest, imgspecv1.MediaTypeImageManifestList: // A recognized type.
+	case DockerV2Schema2MediaType, DockerV2ListMediaType, imgspecv1.MediaTypeImageManifest: // A recognized type.
 		return meta.MediaType
 	}
 	// this is the only way the function can return DockerV2Schema1MediaType, and recognizing that is essential for stripping the JWS signatures = computing the correct manifest digest.
--- a/manifest/manifest_test.go
+++ b/manifest/manifest_test.go
@@ -22,7 +22,6 @@ func TestGuessMIMEType(t *testing.T) {
 		mimeType string
 	}{
 		{"ociv1.manifest.json", imgspecv1.MediaTypeImageManifest},
-		{"ociv1list.manifest.json", imgspecv1.MediaTypeImageManifestList},
 		{"v2s2.manifest.json", DockerV2Schema2MediaType},
 		{"v2list.manifest.json", DockerV2ListMediaType},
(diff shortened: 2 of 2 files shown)


manifest/manifest.go
```diff
@@ -54,7 +54,7 @@ func GuessMIMEType(manifest []byte) string {
 	}
 
 	switch meta.MediaType {
-	case DockerV2Schema2MediaType, DockerV2ListMediaType, imgspecv1.MediaTypeImageManifest, imgspecv1.MediaTypeImageManifestList: // A recognized type.
+	case DockerV2Schema2MediaType, DockerV2ListMediaType, imgspecv1.MediaTypeImageManifest: // A recognized type.
```
runcom: `GuessMIMEType` no longer guesses mime types from OCI unfortunately. I think we need to get rid of the image manifest type as well.

manifest/manifest_test.go
```diff
@@ -22,7 +22,6 @@ func TestGuessMIMEType(t *testing.T) {
 		mimeType string
 	}{
 		{"ociv1.manifest.json", imgspecv1.MediaTypeImageManifest},
-		{"ociv1list.manifest.json", imgspecv1.MediaTypeImageManifestList},
```
runcom: you should "rm" the fixture also

erikh:
Ok thanks. Give me a few hours.

cyphar:
This patch right. It wasn't *removed*, the type was renamed in opencontainers/image-spec#546.

But as I mentioned in #193, this is breaking the build because we are trying to run our tests with `master` of a specification that is still being worked on. Which is insanity. The correct fix is to run our tests within a project that has actually vendored the specification.

Since `containers/image` still only supports OCI version `1.0.0-rc4` (because `-rc5` isn't out yet) there's no point in making half-changes when we're going to have to rewrite quite a few parts of the `oci:` transport.

erikh:
Right, I've mentioned this before as well. @mtrmac do you mind if I take
another stab at the vendoring requirements? We could vendor out-of-band
just for tests too.

cyphar:
@erikh I've started working on fixing it with projectatomic/skopeo#308 -- though it would be great if we didn't require `skopeo` in order to test anything (the current situation with the cross-vendoring is getting a bit silly).

erikh:
I agree. I'll see if I can whip up something here soon that will resolve
the issue in a way that makes everyone happy.

erikh:
I have filed #240 to resolve this issue in another way. I will close this issue now.

manifest/manifest.go
mtrmac: Yeah; when OCI has removed the `MediaType` field, we should have something like this PR whether or not we hide the build breakage by vendoring and old version.

(Or I guess we _could_ leave this in, with hard-coded strings, just to be able to deal with some existing pre-1.0 files.)

PR #240, opened 2017-02-21
.gitignore,vendor.conf,Makefile: use vndr to manage dependencies for testing.

erikh:
This PR allows us to test against known versions in a `vendor.conf` provided by
https://github.com/LK4D4/vndr, which is incorporated into the Makefile for
installation and use.

The goal of this PR is to allow us to both test against a specific set of
controllable versions and also refer to this set of versions as a canonical
"this is what we support", without actually forcing the libraries on anyone who
might want to import this package.


Makefile
```diff
@@ -15,8 +15,12 @@ deps:
 	go get -u $(BUILDFLAGS) github.com/golang/lint/golint
 	go get $(BUILDFLAGS) github.com/vbatts/git-validation
 
-test:
-	@go test $(BUILDFLAGS) -cover ./...
+vndr:
+	go get -u github.com/LK4D4/vndr
+	vndr
+
+test: vndr
```
runcom: we might want to remove `test-skopeo`? @mtrmac @cyphar wdyt?

cyphar: At the moment I don't think we have enough (read: any) integration tests for me to say we should drop the only proper integration tests we have. In the future we should drop it IMO.

erikh:
I didn't want to jump the gun on that, but would if you wanted.

erikh:
figures; gopkg.in is busted right now.

erikh:
blocked by https://github.com/projectatomic/skopeo/pull/312 -- @runcom can you help me with this one?

Makefile
mtrmac: The original goal of `test-skopeo` was to stop us from constantly breaking `test-skopeo` with API changes, without even realizing that we broke it; I think that still very much applies.

(In principle, we could instead have some kind of public API list and just fail tests every time we break the API. But at least the two of us would end up using that test failure to update skopeo, so for us it amounts to the same thing; not sure about other contributors.)

mtrmac:
This primarily seems blocked by the `vendor` mechanism breaking API compatibility, see https://github.com/docker/distribution/issues/2130 for a full explanation.

When `test-skopeo` copies containers/image into skopeo’s `vendor` subdirectory, it must not copy containers/image/vendor in there as well.

(Being dragged kicking and screaming towards containers/image vendoring anything at all, I would still like it if it were very easy to run tests against the current or latest versions, i.e. using a separate non-default directory and an explicit `GOPATH` override to run the tests. But I can live with this.)

Makefile
```diff
@@ -8,15 +8,20 @@ SUDO =
...
 	go get -u $(BUILDFLAGS) github.com/golang/lint/golint
 	go get $(BUILDFLAGS) github.com/vbatts/git-validation
 
-test:
-	@go test $(BUILDFLAGS) -cover ./...
+vndr:
+	go get -u github.com/LK4D4/vndr
+	vndr
+
+test: deps
```
mtrmac: Is this strictly necessary? Slowing down the edit-test cycle is not all that attractive, and I currently tend to use `go build ./... && make .gitvalidation test validate` instead of `make all` to avoid the `deps` step.

`make all` will run `deps` anyway.

At the very least, the `vndr` target should be able to use a timestamp so that `vndr` gets rerun only when `vendor.conf` is updated.

Makefile
```diff
@@ -36,12 +41,12 @@ test-skopeo:
...
-	@go vet ./...
-	@test -z "$$(gofmt -s -l . | tee /dev/stderr)"
+	@go vet $(PACKAGES)
+	@test -z "$$(gofmt -s -l . | grep -ve '^vendor' | tee /dev/stderr)"
 
 lint:
-	@out="$$(golint ./...)"; \
-	if [ -n "$$(golint ./...)" ]; then \
+	@out="$$(golint $(PACKAGES))"; \
+	if [ -n "$$(golint $(PACKAGES))" ]; then \
```
mtrmac: Wait, we are running golint twice here? Please replace the second reference with `$$out`, perhaps as a separate commit.

erikh:
here's a solution to the multi-vendor problem which I think also resolves your concerns about wanting to move away from the vendor directory when you need to.

If the vendor directory is populated at all in the repo, the `vndr` target will now quietly succeed without running, allowing for improved test times AND the ability to populate your own vendor directory without fear of having it wiped out.

Work for you? I also resolved the golint issues.

mtrmac:
Thanks, this does work for me.

Perhaps it would be possible to do
```make
vendor: vendor.conf
	go get -u github.com/LK4D4/vndr && vndr
```

(or perhaps there would have to be a separate timestamp file, in case `vndr` runs but decides not to change anything?) to actually take advantage of `make`, in particular for a `vendor.conf` update to cause `vndr` to be run automatically — but I am perfectly happy with the current `Makefile` which runs `vndr` once on initial checkout, and requires manual reruns.

I guess https://github.com/projectatomic/skopeo/issues/304 would end up with decisions how to manage the dependency list, no need to rehash that here.

As for the current test failure:
> When `test-skopeo` copies containers/image into skopeo’s vendor subdirectory, it must not copy `containers/image/vendor` in there as well.

runcom:
we need this in order to bump image-spec to RC5 and adapt the code for the new `ManifestIndex` stuff (I'm already working on it but waiting to have these things settled before making a PR)

Makefile
```diff
@@ -16,19 +16,21 @@ deps: vndr
...
-	 @go test $(BUILDFLAGS) -cover $(PACKAGES)
+	@go test $(BUILDFLAGS) -cover $(PACKAGES)
 
 # This is not run as part of (make all), but Travis CI does run this.
 # Demonstarting a working version of skopeo (possibly with modified SKOPEO_REPO/SKOPEO_BRANCH, e.g.
 #    make test-skopeo SKOPEO_REPO=runcom/skopeo-1 SKOPEO_BRANCH=oci-3 SUDO=sudo
 # ) is a requirement before merging; note that Travis will only test
 # the master branch of the upstream repo.
-test-skopeo:
+test-skopeo: clean
```
mtrmac: It is _extremely_ non-obvious why the `clean` here matters, and it is unintuitive that `make test-something` modifies the vendored libraries in the current directory.

At the very least this needs a detailed comment explaining why the “clean” here is essential and shouldn’t be removed; but I’d much prefer dropping this dependency, and doing `rm -rf $${vendor_path} && cp -r . $${vendor_path}`**` && rm -rf $${vendor_path}/vendor`** below.

erikh:
that's going to completely destroy vendor every time the tests are run, we
want to keep vndr out of the way between runs right?

erikh:
ah misunderstood. I'll look into it, not sure why the copy is necessary
though.

erikh:
ok. This should be passing now.

mtrmac:
:+1: for pullapprove

runcom:
:+1:

runcom:
Needs one last rebase

erikh:
it is done

mtrmac:
> If the vendor directory is populated at all in the repo, the vndr target will now quietly succeed without running, allowing for improved test times AND the ability to populate your own vendor directory without fear of having it wiped out.

(This has somehow disappeared? Anyway, being dealt with in #242; I am just mildly curious what has happened.)

erikh:
I'm confused too :)

PR #241, opened 2017-02-25
update test vendor to image-spec rc5

runcom:


runcom:
This is WIP since I need to rework the OCI transports for RC5 - just a placeholder for now

cyphar:
@runcom We need to talk about how we are going to handle references given the discussion I've had upstream especially with opencontainers/image-spec#588 -- effectively the way it should be handled is like SQL queries.

Hopefully `oci/layout` can be implemented using `umoci`'s libraries so you don't have to reimplement it.

runcom:
> Hopefully oci/layout can be implemented using umoci's libraries so you don't have to reimplement it.

mmm I don't understand why we went for using another library (umoci's) for OCI support in c/image (which is already a library)
but ok...OCI support will probably drop from c/image eventually (I guess) and skopeo can directly use umoci's library

mtrmac:
> @runcom We need to talk about how we are going to handle references given the discussion I've had upstream especially with opencontainers/image-spec#588 -- effectively the way it should be handled is like SQL queries.

(\*sigh\* that seems rather overengineered and very ambiguous (how does one do matching on `os.version`, or `variant` (is `arm8` ≥ `arm6l`?); oh well, it does define enough to be usable I guess.)

So far we’ve handled the similar docker schema2 functionality in `image/docker_list.go`, i.e. when reading the image, we resolve the list into a single manifest at the earliest opportunity. But that’s all that is supported, there is no support for manifest list destinations, and `copy.go` explicitly refuses to copy multi-images.

> but ok...OCI support will probably drop from c/image eventually (I guess) and skopeo can directly use umoci's library

I don’t particularly care about how the OCI transport is implemented (e.g. I would be perfectly fine with a set of one-liners calling `umoci`), but it does seem valuable to me to have OCI go through the `types.ImageTransport` abstraction so that signatures and the unified copy pipeline etc. can be used, and so that skopeo does not have to build a private abstraction over containers/image and umoci; that would be a bit ridiculous.

erikh:
If I may jump in I think there's a valuable discussion in making
containers/image (and perhaps parts of containers/storage) bank off of
umoci for many of its operations. I spoke to @cyphar over twitter about
making this a thing that could happen, the editing features I need are
coming to umoci and the interface in containers/image from a lib
perspective is very nice.

Perhaps getting in a position where we can consume umoci as a lib would be
the best option? I am very, very happy to do this work if desired.

erikh:
I have code to do a lot of what umoci does for docker images, and have
outstanding patches like https://github.com/docker/docker/pull/28923 that
can finish the work for docker-specific image work.

mtrmac:
> If I may jump in I think there's a valuable discussion in making containers/image (and perhaps parts of containers/storage) bank off of umoci for many of its operations.

Honestly I don’t know all that much about the wider runc / OCI / … ecosystem¹, @runcom ?

My _guess_ after a five-minute look at umoci is that c/image currently tends to deal with images as a single ~immutable unit, so converting between image formats and probably config formats makes sense as part of that, but editing layers and especially individual files inside is conceptually somewhat different (notably it really needs a semi-persistent staging area for in-progress multi-step edits). That’s not to say that the two shouldn’t be very easily interoperable, or they may even be a single repo/project—at the very least `types.ImageReference` or something like that would be a nice thing to share.

But then…
¹ I came here for signatures; source/destination abstractions, `copy.Image`, and format conversions all just grew from that and were not originally anticipated.

Really, the project will grow in whichever way contributors want to take it :)


Actually the `containers-storage:` transport is already somewhat awkward to handle using our abstraction (assuming that a layer put into an `ImageDestination` can be copied back out from an `ImageSource` unmodified, which is not true for `containers-storage`), and if I understand `umoci` correctly, having containers/image know about “local extracted images” (which are not image-at-a-time sources/destinations) might make all of this much clearer.

Perhaps the `copy` abstraction which completely hides the difference between “push”/“pull”/“copy between servers” is just untenable…

\* mitr shuts up now to stop displaying even more of his ignorance.

erikh:
I dunno. It just seems like there's a ton of overlap and that maybe a sink
for some of these core image/storage operations would be best. Perhaps I'm
too eager to unify something that doesn't need to be.

cyphar:
@mtrmac 

> (*sigh* that seems rather overengineered and very ambiguous (how does one do matching on os.version, or variant (is arm8 ≥ arm6l?); oh well, it does define enough to be usable I guess.)

If you'd like to continue the good fight within the spec, that's your decision. Stephen Day made clear that in his opinion the tagging system of Docker (and thus all of the similar systems) is _too_ simple and that a tagging system is much more useful for people.

> I don’t particularly care about how the OCI transport is implemented (e.g. I would be perfectly fine with a set of one-liners calling umoci)

`umoci` implements a set of libraries, so you wouldn't be using it as a binary. Effectively `oci/layout` would become `oci/` and it would just be a thin wrapper around `umoci`'s libraries.

> but editing layers and especially individual files inside is conceptually somewhat different (notably it really needs a semi-persistent staging area for in-progress multi-step edits).

That's the front-end of `umoci`. The libraries don't have that requirement (unless you're generating a new diff layer, which requires you to have a directory with files in it that can be used to source metadata and data). I'm not anticipating that c/image requires everything that `umoci` implements -- I only really expect that c/image needs `oci/cas` and `oci/casext`.

cyphar:
@runcom 

> mmm I don't understand why we went for using another library (umoci's) for OCI support in c/image (which is already a library)

So you don't have to reimplement the OCI CAS interface. I mean, it's up to you ultimately, but I'm a bit concerned that `skopeo` is going to end up with an incompatible reference interface to `umoci` (which, aside from being bad for the OCI community, will actually impact us at SUSE). As long as we have discussions on how the reference API should be handled, I don't mind all that much whether you re-use my work or reimplement it.

mtrmac:
> > (*sigh* that seems rather overengineered and very ambiguous (how does one do matching on os.version, or variant (is arm8 ≥ arm6l?); oh well, it does define enough to be usable I guess.)

> If you'd like to continue the good fight within the spec, that's your decision.

I'm afraid I can’t spend time on that now. It does seem that a subset of that is easy enough to use in  portable code, so I guess that’s what will end up happening… Or, \*shrug\*, I could just be wrong and this could be the right thing to do.

> > I don’t particularly care about how the OCI transport is implemented (e.g. I would be perfectly fine with a set of one-liners calling umoci)

>`umoci` implements a set of libraries, so you wouldn't be using it as a binary. Effectively `oci/layout` would become `oci/` and it would just be a thin wrapper around `umoci`'s libraries.

Sure, one-liners calling a library. Executing a binary would be more lines :)

> > but editing layers and especially individual files inside is conceptually somewhat different (notably it really needs a semi-persistent staging area for in-progress multi-step edits).

> That's the front-end of `umoci`. The libraries don't have that requirement (unless you're generating a new diff layer, which requires you to have a directory with files in it that can be used to source metadata and data). I'm not anticipating that c/image requires everything that `umoci` implements -- I only really expect that c/image needs `oci/cas` and `oci/casext`.

Ah, OK. That is indeed much closer to the `oci:` transport in this repo. I have no opinion on implementing the `oci:` transport through umoci, or not, in that case; whichever ends up being easier and more maintainable.

lucab:
@runcom barring the unfortunate refs-as-annotations situation in -rc5 and the single test failing in here, is there anything else blocking this? I did a quick roundtrip of this through skopeo and seems fine on the surface, but I may be missing some details.

lucab:
Bump on my last comment.

runcom:
@lucab, not sure, I was waiting on @cyphar I guess

runcom:
I'll fix the test meanwhile

cyphar:
I've got a branch in `umoci` for this (https://github.com/openSUSE/umoci/tree/imagespec-v1.0.0-rc5) that I haven't worked on for a couple of days (mainly because I'm working on some other related image-spec stuff related to distribution). I'm going to need to finish it ASAP in order to implement some of the stuff I'm planning in distribution, but I wanted to wait to meet with @steveooe first (which I did at DockerCon though we didn't get a chance to run through the walker implementation).

Basically I think the best way of moving forward is that skopeo implement the most "simple" tagging implementation (which can be done with the current state of my umoci branch's OCI CAS library) -- more complicated manifest trees aren't necessary for the "simple" ca se. There are some complications with "updating" a tag, but umoci's updating implementation will probably never really want to work with non-unique tags (we'll see though).

The kinda nice thing about rc5 is that I can implement references in `casext` (so references are implemented completely separately from the CAS and you can implement different reference resolution methods externally). I still need to make `casext` more `interface`-friendly though...

cyphar:
I definitely have put umoci integration on my list of things that "would be a good thing" in `containers/image` though. When I get a chance I will actually write that code.

runcom:
@cyphar sorry, I didn't get if this PR still stand on his own and could be moved forward

manifest/manifest.go
```diff
@@ -54,7 +54,7 @@ func GuessMIMEType(manifest []byte) string {
 	}
 
 	switch meta.MediaType {
-	case DockerV2Schema2MediaType, DockerV2ListMediaType, imgspecv1.MediaTypeImageManifest, imgspecv1.MediaTypeImageManifestList: // A recognized type.
+	case DockerV2Schema2MediaType, DockerV2ListMediaType: // A recognized type.
```
mtrmac: I vaguely remember there was an earlier conversation about this, but I can't remember the details: why is guessing that a manifest is OCI no longer applicable? Because a manifest can never be a stand-alone object with OCi without an index now? Or just because OCI decided not to include the type in the JSON?

runcom: No mime type in Json anymore

runcom:
ping @cyphar about https://github.com/containers/image/pull/241#issuecomment-298140971

manifest/manifest.go
cyphar: To expand on the above, it's because types are now implemented using descriptor types (so objects are no longer self-descriptive).

oci/layout/oci_transport.go
```diff
@@ -171,16 +174,54 @@ func (ref ociReference) PolicyConfigurationNamespaces() []string {
...
+		}
+		refName, ok := md.Annotations["org.opencontainers.ref.name"]
+		if !ok {
+			continue
+		}
+		if refName == ref.tag {
+			d = &md
+			break
+		}
+	}
```
cyphar: Yeah, this logic works for the basic case and was the main thing I was worried about. Once I have a nice implementation in `umoci`, `skopeo` should be able to handle more complicated cases.

lucab:
Our of partially related curiosity, what's the golang libraries plan to deal with https://github.com/opencontainers/image-spec/issues/581#issuecomment-284777108?

runcom:
Tests fixed and rebased, please take a look @mtrmac @cyphar @lucab

manifest/manifest.go
mtrmac: Either way, this is awkward.

* `dirImageSource` and `storageImageSource` depend on `GuessMIMEType` to work; otherwise `c/i/image` assumes schema1. (Not _trivially_ fixable.)
* `dockerImageDestination` depends on `GuessMIMEType` for setting `Content-Type` on upload. (This one is probably fixable by adding an explicit parameter, and using the `c/i/image` handler to get the type we are using.)

Is there another heuristic we can use? e.g. `schemaVersion` == `2` && `mediaType` missing? Or perhaps together with `config.mediaType`?

oci/layout/oci_dest.go
```diff
@@ -167,11 +171,19 @@ func (d *ociImageDestination) PutManifest(m []byte) error {
...
-	descriptorPath := d.ref.descriptorPath(d.ref.tag)
-	if err := ensureParentDirectoryExists(descriptorPath); err != nil {
-		return err
-	}
-	return ioutil.WriteFile(descriptorPath, data, 0644)
+
+	annotations := make(map[string]string)
+	annotations["org.opencontainers.ref.name"] = d.ref.tag
+	desc.Annotations = annotations
+	d.index.Manifests = append(d.index.Manifests, imgspecv1.ManifestDescriptor{
```
mtrmac: This gives a bit of an impression that the code adds more manifests to an existing index, but we actually never read the original file and blindly overwrite it if it exists.

I guess the intent is to perhaps add the “update existing index” over time? If so, this code does make enough sense; if not, and the intent is to always do a simple overwrite, the code structure could probably be a bit simpler (do nothing in `Commit`, just build a single `ImageIndex` local variable in `PutManifest`).

oci/layout/oci_transport.go
```diff
@@ -176,16 +179,54 @@ func (ref ociReference) PolicyConfigurationNamespaces() []string {
 // NOTE: If any kind of signature verification should happen, build an UnparsedImage from the value returned by NewImageSource,
 // verify that UnparsedImage, and convert it into a real Image via image.FromUnparsedImage.
 func (ref ociReference) NewImage(ctx *types.SystemContext) (types.Image, error) {
-	src := newImageSource(ref)
+	d, err := ref.getManifestDescriptor()
+	if err != nil {
+		return nil, err
+	}
+	src := newImageSource(ref, d)
```
mtrmac: This duplication between `NewImage` and `NewImageSource` rather suggests that `getManifestDescriptor` should be called within `newImageSource` instead of supplied from the outside.

(That still allows `getManifestDescriptor` to be a method on `ociReference` if you want.)

oci/layout/oci_transport.go
```diff
@@ -176,16 +179,54 @@ func (ref ociReference) PolicyConfigurationNamespaces() []string {
...
 }
 
+func (ref ociReference) getManifestDescriptor() (imgspecv1.ManifestDescriptor, error) {
+	indexJSON, err := os.Open(ref.indexPath())
+	if err != nil {
+		return imgspecv1.ManifestDescriptor{}, err
+	}
+	defer indexJSON.Close()
+	index := imgspecv1.ImageIndex{}
+	if err := json.NewDecoder(indexJSON).Decode(&index); err != nil {
```
mtrmac: This is shorter than `ReadAll` + `Unmarshal`, but doesn’t it also silently accept garbage (or a valid JSON?) after the processed object? And does that matter? (Quite possibly it doesn’t…)

oci/layout/oci_transport_test.go
```diff
@@ -239,6 +258,14 @@ func TestReferenceOCILayoutPath(t *testing.T) {
 	assert.Equal(t, tmpDir+"/oci-layout", ociRef.ociLayoutPath())
 }
 
+func TestReferenceIndexJSON(t *testing.T) {
```
mtrmac: Should this be `TestReferenceIndexPath`?

oci/layout/oci_dest.go
runcom: yup, the intent is to "update existing index"

oci/layout/oci_transport.go
runcom: shouldn't matter afaict

manifest/manifest.go
runcom: fixed I think

runcom:
@mtrmac PTAL

oci/layout/oci_dest.go
mtrmac: > yup, the intent is to "update existing index"

As in, with this PR, or at some later time? Because it is not updating an existing index now.

manifest/manifest.go
mtrmac: Thanks. Could the `if meta.SchemaVersion == 2` code be unified with `case 2` below?

mtrmac:
(skopeo needs updating as well go keep CI happy.)

manifest/manifest.go
runcom: done

oci/layout/oci_dest.go
runcom: if you call `PutManifest` many times and at the end `Commit` you'll have a multi-manifest index with this code, am I stupidly missing something?

runcom:
@mtrmac updated skopeo as well

oci/layout/oci_dest.go
mtrmac: Ah, true. I wasn’t considering that because we have no tools to do that (and the semantics of doing that + `PutSignatures` is unclear — what manifest are the signatures attached to?).

Anyway, that can be useful for someone who explicitly expects that `oci:` behavior, and it doesn’t break anything, so fair enough.

mtrmac:
LGTM pending tests / a skopeo update which passes tests.

runcom:
LGTM pending skopeo tests

PR #243, opened 2017-02-27
Makefile: fix 'make test'

giuseppe:
Read this bit that got lostwith commit 01381869c6307e5801ce5413c3400d86f75d0b07

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ SUDO =
 BUILDTAGS   = btrfs_noversion libdm_no_deferred_remove
 BUILDFLAGS := -tags "$(BUILDTAGS)"
 
-PACKAGES := $(shell go list ./... | grep -v vendor)
+PACKAGES := $(shell go list ./... | grep -v github.com/containers/image/vendor)
 
 all: deps .gitvalidation test validate
 


Makefile
```diff
@@ -23,7 +23,7 @@ clean:
 	rm -rf vendor
 
 test: deps
-	@go test $(BUILDFLAGS) -cover $(PACKAGES)
+	@go test $(BUILDFLAGS) -cover $(PACKAGES) ./...
```
mtrmac: What is this supposed to fix? What is the failure, and how does this improve on it?

Look at `PACKAGES` is computed: it is supposed to be equivalent to `./...`, except that `vendor/**` is excluded. We don’t really want to run tests in the vendored packages, so I can’t see what difference this should make.

giuseppe: not sure if it is the correct fix but `make test` doesn't run any test for me on master.  This is what I get:

```
$ make test
?   	github.com/projectatomic/skopeo/vendor/github.com/containers/image	[no test files]
```

mtrmac: That… doesn’t make _any_ sense to me. How can the list of packages tested contain _only_ a `/vendor/` package which has been explicitly filtered out?

Can you look into how `$PACKAGES` is computed, and perhaps compare that with `make --just-print test`?

giuseppe: ok so the issue was that I was running it from skopeo/vendor for quickly verifying a patch.  I've pushed another fix

mtrmac: > ok so the issue was that I was running it from skopeo/vendor for quickly verifying a patch.

How? `github.com/projectatomic/skopeo/vendor/github.com/containers/image` does not contain a copy of this `Makefile` at all.

giuseppe: I copied over this repository so that I could test directly with Skopeo.  I agree it is a weird case, and not something you probably want to support, so feel free to close this PR

mtrmac: Yeah, it _is_ a weird case, and to be clear, not at all supported. But the change is innocuous enough, I guess, so I’ll let @runcom decide.

mtrmac:
👍 @runcom deferring to your decision.

erikh:
If I may jump in, I don't think this is a good idea; `$(PACKAGES)` vs `./...` means that `./...` will always contain the subset that is in `$(PACKAGES)` (which includes `vendor/`)

runcom:
> If I may jump in, I don't think this is a good idea; $(PACKAGES) vs ./... means that ./... will always contain the subset that is in $(PACKAGES) (which includes vendor/)

could you look at the last rev of this PR? I think it's safe to pull this in now.

erikh:
yep LGTM

runcom:
LGTM

PR #245, opened 2017-03-02
Revert "manifest: remove not needed mediatype"

mtrmac:
This reverts commit 0a5f50db799793f161cfc4d8d4fb00c9aa37585e. ,
which broke tests.

Eventually we may need that commit (when updating `image-spec`?), but right now it seems unnecessary and this is the most obvious way to get tests working without having to think :)

### DIFF
--- a/manifest/manifest.go
+++ b/manifest/manifest.go
@@ -54,7 +54,7 @@ func GuessMIMEType(manifest []byte) string {
 	}
 
 	switch meta.MediaType {
-	case DockerV2Schema2MediaType, DockerV2ListMediaType: // A recognized type.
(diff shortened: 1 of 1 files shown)


runcom:
> This reverts commit 0a5f50d. ,
> which broke tests.
> 
> Eventually we may need that commit (when updating image-spec?), but right now it seems unnecessary and this is the most obvious way to get tests working without having to think :)

ops, silly me, btw, that commit wasn't totally wrong, RC4 doesn't have mediatype in the manifest anymore so that `GuessMIMEType` can't work according to my understanding for oci images. Could you instead remove the failing tests?

mtrmac:
Honestly I don't want to be thinking about this now, and #241 is doing that anyway. This is just to get tests working

(… assuming we can get past Travis; I’ve seen 2 runs fail with “unexpected EOF” in 2 different places.)

runcom:
alright, lgtm

mtrmac:
:+1:

PR #120, opened 2016-10-05
Signature file layout specification

aweiteka:
cc @mhrivnak

PR #247, opened 2017-03-04
Change sigstore layout

### LINKED ISSUE #187: Docker signature transport uses reserved URI characters
Per implementation (also doc https://github.com/containers/image/pull/120) the manifest digest reference for the "docker" transport type is HASH_FUNCTION:HASH_VALUE, using a colon (:) delimiter. The "at" symbol (@) is also used the delimit image and manifest digest (again, see https://github.com/containers/image/pull/120). Per [URI: Generic Syntax](https://tools.ietf.org/html/rfc3986#section-2.2) these are reserved delimiting characters.

**Impact**
The colon in the URI is blocking uploading of signatures to JFrog's [Artifactory repository management platform](https://www.jfrog.com/artifactory/), returning 500 error:

```
"Invalid path. ':' is not a valid name character: atomic-sigstore/aweiteka/true@sha256:f292b8573b1679a512f575d7bc2441815b7528eb114217781199e9106e742e21/"
```
Percent-encoding the path (`%3A`) does not help. Artifactory has not responded to this [related issue](https://www.jfrog.com/jira/browse/RTFACT-8695)

mtrmac:
This

* Changes the file name to use `=` instead of `:`, to avoid a Windows reserved character. Fixes #187 .
* Removes the host name from the path, per the conversation around #120 .

### DIFF
--- a/docker/lookaside.go
+++ b/docker/lookaside.go
@@ -9,12 +9,12 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/Sirupsen/logrus"
+	"github.com/containers/image/docker/reference"
+	"github.com/containers/image/types"
 	"github.com/ghodss/yaml"
 	"github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
-
-	"github.com/Sirupsen/logrus"
-	"github.com/containers/image/types"
 )
 
 // systemRegistriesDirPath is the path to registries.d, used for locating lookaside Docker signature storage.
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

mtrmac:
@aweiteka @mhrivnak (@puiterwijk) PTAL.

E.g. for [`docker.io/library/`]`busybox:latest` the path now is $root`/library/busybox@sha256=8…e/signature-1` etc.

puiterwijk:
This looks good to me.

runcom:
LGTM

PR #59, opened 2016-08-25
Image signature format

aweiteka:
Proposed OCI image signing specification

cc @philips @rhatdan @mtrmac

PR #251, opened 2017-03-11
Add documentation for the atomic signature format

mtrmac:
This documents the signature format, in more detail than #59.

### DIFF
--- a/docs/atomic-signature.md
+++ b/docs/atomic-signature.md
@@ -1,0 +1,241 @@
+% atomic-signature(5) Atomic signature format
+% Miloslav Trmač
+% March 2017
+
+# Atomic signature format
+
+This document describes the format of “atomic” container signatures,
+as implemented by the `github.com/containers/image/signature` package.
+
+Most users should be able to consume these signatures by using the `github.com/containers/image/signature` package
+(preferably through the higher-level `signature.PolicyContext` interface)
+without having to care about the details of the format described below.
+This documentation exists primarily for maintainers of the package
+and to allow independent reimplementations.
+
+## High-level overview
+
+The signature provides an end-to-end authenticated claim that a container image
+has been approved by a specific party (e.g. the creator of the image as their work,
+an automated build system as a result of an automated build,
+a company IT department approving the image for production) under a specified _identity_
+(e.g. an OS base image / specific application, with a specific version).
+
+An atomic container signature consists of a cryptographic signature which identifies
+and authenticates who signed the image, and carries as a signed payload a JSON document.
+The JSON document identifies the image being signed, claims a specific identity of the
+image and if applicable, contains other information about the image.
+
+The signatures do not modify the container image (the layers, configuration, manifest, …);
+e.g. their presence does not change the manifest digest used to identify the image in
+docker/distribution servers; rather, the signatures are associated with an immutable image.
+An image can have any number of signatures so signature distribution systems SHOULD support
+associating more than one signature with an image.
+
+## The cryptographic signature
+
+As distributed, the atomic container signature is a blob which contains a cryptographic signature
+in an industry-standard format, carrying a signed JSON payload (i.e. the blob contains both the
+JSON document and a signature of the JSON document; it is not a “detached signature” with
+independent blobs containing the JSON document and a cryptographic signature).
+
+Currently the only defined cryptographic signature format is an OpenPGP signature (RFC 4880),
+but others may be added in the future.  (The blob does not contain metadata identifying the
+cryptographic signature format. It is expected that most formats are sufficiently self-describing
+that this is not necessary and the configured expected public key provides another indication
+of the expected cryptographic signature format. Such metadata may be added in the future for
+newly added cryptographic signature formats, if necessary.)
+
+Consumers of atomic container signatures SHOULD verify the cryptographic signature
+against one or more trusted public keys
+(e.g. defined in a [policy.json signature verification policy file](policy.json.md))
+before parsing or processing the JSON payload in _any_ way,
+in particular they SHOULD stop processing the container signature
+if the cryptographic signature verification fails, without even starting to process the JSON payload.
+
+(Consumers MAY extract identification of the signing key and other metadata from the cryptographic signature,
+and the JSON payload, without verifying the signature, if the purpose is to allow managing the signature blobs,
+e.g. to list the authors and image identities of signatures associated with a single container image;
+if so, they SHOULD design the output of such processing to minimize the risk of users considering the output trusted
+or in any way usable for making policy decisions about the image.)
+
+### OpenPGP signature verification
+
+When verifying a cryptographic signature in the OpenPGP format,
+the consumer MUST verify at least the following aspects of the signature
+(like the `github.com/containers/image/signature` package does):
+
+- The blob MUST be a “Signed Message” as defined RFC 4880 section 11.3.
+  (e.g. it MUST NOT be an unsigned “Literal Message”, or any other non-signature format).
+- The signature MUST have been made by an expected key trusted for the purpose (and the specific container image).
+- The signature MUST be correctly formed and pass the cryptographic validation.
+- The signature MUST correctly authenticate the included JSON payload
+  (in particular, the parsing of the JSON payload MUST NOT start before the complete payload has been cryptographically authenticated).
+- The signature MUST NOT be expired.
+
+The consumer SHOULD have tests for its verification code which verify that signatures failing any of the above are rejected.
+
+## JSON processing and forward compatibility
+
+The payload of the cryptographic signature is a JSON document (RFC 7159).
+Consumers SHOULD parse it very strictly,
+refusing any signature which violates the expected format (e.g. missing members, incorrect member types)
+or can be interpreted ambiguously (e.g. a duplicated member in a JSON object).
+
+Any violations of the JSON format or of other requirements in this document MAY be accepted if the JSON document can be recognized
+to have been created by a known-incorrect implementation (see [`optional.creator`](#optionalcreator) below)
+and if the semantics of the invalid document, as created by such an implementation, is clear.
+
+The top-level value of the JSON document MUST be a JSON object with exactly two members, `critical` and `optional`,
+each a JSON object.
+
+The `critical` object MUST contain a `type` member identifying the document as an atomic container signature
+(as defined [below](#criticaltype))
+and signature consumers MUST reject signatures which do not have this member or in which this member does not have the expected value.
+
+To ensure forward compatibility (allowing older signature consumers to correctly
+accept or reject signatures created at a later date, with possible extensions to this format),
+consumers MUST reject the signature if the `critical` object, or _any_ of its subobjects,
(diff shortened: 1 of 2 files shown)


runcom:
ping @aweiteka @vbatts @rhatdan @rektide @imcleod

docs/atomic-signature.md
```diff
@@ -0,0 +1,233 @@
...
+in particular they SHOULD stop processing the container signature
+if the cryptographic signature verification fails, without even starting to process the JSON payload.
+
+(Consumers MAY extract identification of the signing key and other metadata from the cryptographic signature,
+and the JSON payload, without verifying the signature, if the purpose is to allow managing the signature blobs,
+e.g. to list the authors and image identities of signatures attached to a single container image;
+if so, they SHOULD design the output of such processing to minimize the risk of users considering the output trusted
+or in any way usable for making policy decisions about the image.)
+
+### OpenPGP signature verification
```
aweiteka: This is a good section. Can you speak here about the library? Does the library cover all of these verification steps? Is it viable to develop a custom verification method or is that not recommended?

docs/atomic-signature.md
```diff
@@ -0,0 +1,233 @@
...
+
+Currently the only defined cryptographic signature format is an OpenPGP signature (RFC 4880),
+but others may be added in the future.  (The blob does not contain metadata identifying the
+cryptographic signature format; it is expected that most formats are sufficiently self-describing
+that this is not necessary, and the configured expected public key provides another indication
+of the expected cryptographic signature format. Such metadata may be added in the future for
+newly added cryptographic signature formats, if necessary.)
+
+Consumers of atomic container signatures SHOULD verify the cryptographic signature
+against one or more trusted public keys
```
mtrmac: The user can set up a keyring with several keys, all equally acceptable for a particular namespace (e.g. a Fedora 30 primary key, Fedora 30 secondary key, Fedora security team universal emergency key).

docs/atomic-signature.md
```diff
@@ -0,0 +1,233 @@
...
+(e.g. an OS base image / specific application, with a specific version).
+
+An atomic container signature consists of a cryptographic signature which identifies
+and authenticates who signed the image, and carries as a signed payload a JSON document.
+The JSON document identifies the image being signed, and claims
+a specific identity of the image, and possibly other information about the image.
+
+The signatures do not modify the container image (the layers, configuration, manifest, …);
+e.g. their presence does not change the manifest digest used to identify the image in
+docker/distribution servers; rather, the signatures are an “attachment” to an immutable image.
```
mtrmac: That is indeed better, updated all three instances.

docs/atomic-signature.md
```diff
@@ -0,0 +1,233 @@
...
+An image can have any number of signatures, and image distribution systems SHOULD support
+attaching more than one signature to an image.
```
mtrmac: Updated. (Using “associating…with” per a random dictionary and some googling; if there is a semantic difference, please tell me.)

docs/atomic-signature.md
mtrmac: > Is it viable to develop a custom verification method or is that not recommended?

The current implementation is not _magic_, it is entirely possible to write another correct one (and this section, listing things to check, is also intended to help write it correctly). Actually we haven’t had an independent adversarial audit or anything similar to give us a very high level of confidence that the library one is indeed correct. :-)

Still, it is not _quite_ trivial, it is easy to forget something, and various verification APIs make it fairly easy to miss something (e.g. the OpenShift reimplementation had, it in its initial versions, missed several aspects).

And using a shared implementation should _hopefully_ increase the likelihood of it being good, many eyeballs and all that.

Finally, using the shared implementation means less work with updating as the spec expands, e.g. to support layered images—at the cost of more work with updating the callers as the library API changes.

> Can you speak here about the library? Does the library cover all of these verification steps?

Added a general note in the very first section about code just using the library instead of worrying about anything in this document, and a parenthesized note in this section that the library should take care of this. (Such a note could really be in _every_ section but this one looks a bit more scary than the others, so adding an extra note in here to reassure makes sense.)

docs/atomic-signature.md
```diff
@@ -0,0 +1,233 @@
...
+The signature provides an end-to-end authenticated claim that a container image
+has been approved by a specific party (e.g. the creator of the image as their work,
+an automated build system as a result of an automated build,
+a company IT department approving the image for production) under a specified _identity_
+(e.g. an OS base image / specific application, with a specific version).
+
+An atomic container signature consists of a cryptographic signature which identifies
+and authenticates who signed the image, and carries as a signed payload a JSON document.
+The JSON document identifies the image being signed, and claims
+a specific identity of the image, and possibly other information about the image.
```
mtrmac: This is trying to say that the format can be extended, both in the future and in independent implementations, i.e. it is not doomed to be an (image, identity) pair forever, and to somewhat motivate the backward/forward compatibility concerns.

It seems _slightly_ incorrect to just say “identifies the image… and claims a specific identity”, though the “possibly” in there does seem a rather vague.

mtrmac:
@aweiteka Updated based on your review, much appreciated!

mtrmac:
@runcom PTAL

mtrmac:
@runcom ping

runcom:
This looks good to me @mtrmac

mtrmac:
:+1:

runcom:
LGTM

docs/atomic-signature.md
TomSweeneyRedHat: Perhaps: 
"The JSON document identifies the image being signed, claims a specific identity of the image and if applicable, contains other information about the image."

docs/atomic-signature.md
```diff
@@ -0,0 +1,242 @@
...
+
+## The cryptographic signature
+
+As distributed, the atomic container signature is a blob which contains a cryptographic signature
+in an industry-standard format, carrying a signed JSON payload (i.e. the blob contains both the
+JSON document and a signature of the JSON document; it is not a “detached signature” with
+independent blobs containing the JSON document and a cryptographic signature).
+
+Currently the only defined cryptographic signature format is an OpenPGP signature (RFC 4880),
+but others may be added in the future.  (The blob does not contain metadata identifying the
```
TomSweeneyRedHat: nit, I'd drop the first semi and comma in the first two sentences here in the parens leaving:
"(The blob does not contain metadata identifying the cryptographic signature format.  It is expected that most formats are sufficiently self-describing that this is not necessary and the configured expected public key provides another indication
of the expected cryptographic signature format."

docs/atomic-signature.md
```diff
@@ -0,0 +1,242 @@
...
+  (e.g. it MUST NOT be an unsigned “Literal Message”, or any other non-signature format).
+- The signature MUST have been made by an expected key trusted for the purpose (and the specific container image).
+- The signature MUST be correctly formed and pass the cryptographic validation.
+- The signature MUST correctly authenticate the included JSON payload
+  (in particular, the parsing of the JSON payload MUST NOT start before the complete payload has been cryptographically authenticated).
+- The signature MUST NOT be expired.
+
+The consumer SHOULD have tests for its verification code which verify that signatures failing any of the above are rejected.
+
+(All of this is taken care of within the `github.com/containers/image/signature` package,
```
TomSweeneyRedHat: nit, I'd lose the parens here.

docs/atomic-signature.md
```diff
@@ -0,0 +1,242 @@
...
+
+The `critical` object MUST contain a `type` member identifying the document as an atomic container signature
+(as defined [below](#criticaltype))
+and signature consumers MUST reject signatures which do not have this member or in which this member does not have the expected value.
+
+To ensure forward compatibility (allowing older signature consumers to correctly
+accept or reject signatures created at a later date, with possible extensions to this format),
+consumers MUST reject the signature if the `critical` object, or _any_ of its subobjects,
+contain _any_ member or data value which is unrecognized, unsupported, invalid, or in any other way unexpected.
+(At minimum, this includes unrecognized members in a JSON object, or incorrect types of expected members.)
```
TomSweeneyRedHat: Nit;  Suggested:  Drop the parens and make the last sentence here read:
"Unrecognized members in a JSON object or incorrect types of expected members should also be rejected as a minimum."  

If you don't want to take that or similar, please change "(At minimum," to "(At a minimum,"

docs/atomic-signature.md
```diff
@@ -0,0 +1,242 @@
...
+### `critical.image.docker-manifest-digest`
+
+This MUST be a JSON string, in the `github.com/opencontainers/go-digest.Digest` string format.
+
+The value of this member MUST match the manifest of the signed container image, as implemented in the docker/distribution manifest addressing system.
+
+The consumer of the signature SHOULD verify the manifest digest against a fully verified signature before processing the contents of the image manifest in any other way
+(e.g. parsing the manifest further or downloading layers of the image).
+
+(Note that a single container image manifest may have several valid manifest digest values, using different algorithms.
```
TomSweeneyRedHat: Nit, I'd lose the outer parens here. ((Can you tell I'm not (a) parens fan yet)?)

docs/atomic-signature.md
mtrmac: This sentence is not _really_ a part of the format spec, it says nothing about the file contents. (Motivation for including it at all is https://github.com/containers/image/pull/251#discussion_r109444308 above.)

I guess it is more or less a footnote, but Markdown doesn’t natively support footnotes. Perhaps I should just build one manually using Unicode?

docs/atomic-signature.md
mtrmac: Dropped the parentheses and added an article to “at a minimum”.

“should _also_ be rejected” is imprecise, because “unrecognized members”/“incorrect types” are a subset of the “unrecognized, unsupported, invalid” list in the previous sentence: this sentence is an _example_ of the requirement, not an _addition_ to it. Hence the “this includes” language trying to make this relationship clear; but perhaps that attempt has failed and this needs to be even clearer? Or just drop the sentence entirely as redundant?

docs/atomic-signature.md
mtrmac: Let’s try an “implementation notes” bullet list. (But parens are awesome! (Especially nested ones (like this (or even more nested!)), allowing a complex structure to be… easily parsed by a computer, at the cost of making it difficult to understand by people (which i guess demonstrates your point)).

mtrmac:
@TomSweeneyRedHat Thanks! Updated, with the changes as a separate commit for easier review; I’ll squash them before merging if the result looks OK.

docs/atomic-signature.md
TomSweeneyRedHat: Or if you can't finagle a foot note, perhaps dropping the parens and changing this to:

Note: This signature verification is taken care of within the `github.com/containers/image/signature` package, its users do not have do do any of this manually.

and making the word 'Note' bold in that sentence?  Your call, I trust your judgement.

mtrmac: Well, *bold* is somewhat of an opposite of a footnote :)

The document already starts with a big paragraph about the library, so I replaced this parenthesized paragraph with a much shorter parenthesis in the sentence introducing the bullet list.

PR #252, opened 2017-03-13
Reject duplicate JSON members when parsing signatures

mtrmac:
Instead of relying on `json.Unmarshal` into a `map[string]interface{}`, and dealing with field presence and correct types manually, build a helper based on `paranoidUnmarshalJSONObjects` which verifies field presence automatically, relies on `json.Unmarshal` for enforcing types of unstructured values, and newly rejects JSON objects with duplicate members.

Besides the primary benefit, rejecting ambiguous signatures, this allows us to get rid of the infrastructure for safely manually parsing `map[string]interface{}`. The new `paranoidUnmarshalJSONObjectExactFields` can also be used to simplify `policy.json` parsing a bit.

Also includes one unrelated test bug fix.

See the individual commits for more details.

### DIFF
--- a/signature/json.go
+++ b/signature/json.go
@@ -122,3 +122,25 @@ func paranoidUnmarshalJSONObject(data []byte, fieldResolver func(string) interfa
 	}
 	return nil
 }
+
+// paranoidUnmarshalJSONObject unmarshals data as a JSON object, but failing on the slightest unexpected aspect
+// (including duplicated keys, unrecognized keys, and non-matching types). Each of the fields in exactFields
+// must be present exactly once, and none other fields are accepted.
+func paranoidUnmarshalJSONObjectExactFields(data []byte, exactFields map[string]interface{}) error {
+	seenKeys := map[string]struct{}{}
+	if err := paranoidUnmarshalJSONObject(data, func(key string) interface{} {
+		if valuePtr, ok := exactFields[key]; ok {
+			seenKeys[key] = struct{}{}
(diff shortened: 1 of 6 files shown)


signature/json.go
```diff
@@ -122,3 +64,25 @@ func paranoidUnmarshalJSONObject(data []byte, fieldResolver func(string) interfa
 	}
 	return nil
 }
+
+// paranoidUnmarshalJSONObject unmarshals data as a JSON object, but failing on the slightest unexpected aspect
+// (including duplicated keys, unrecognized keys, and non-matching types). Each of the fields in exactFields
+// must be present exactly once, and none other fields are accepted.
+func paranoidUnmarshalJSONObjectExactFields(data []byte, exactFields map[string]interface{}) error {
+	seenKeys := map[string]struct{}{}
```
runcom: It might be more idiomatic, but a map[string]bool is better suited here and allows to do

If seenkeys[i]... 

W/o playing around with !ok below

runcom:
LGTM just a minor nit

signature/json.go
mtrmac: I somewhat prefer the `struct{}` to make the type closer to the true type (set of `string`), see also earlier conversation in  https://github.com/containers/image/pull/223 , and perhaps earlier.

I don’t feel all that strongly about this, but if you can live with it, I’ll merge as is.

mtrmac:
👍

PR #255, opened 2017-03-25
Add support for the X-Registry-Supports-Signatures API extension

mtrmac:
This is provided by the OpenShift-integrated registry. This is equivalent to the atomic: transport (in the “openshift”) subpackage, but    it requires less code and notably does not require an OpenShift login     context to be configured.
    
See https://github.com/openshift/origin/pull/12504 and     https://github.com/openshift/openshift-docs/pull/3556 for more     information on this API extension.
    
To preserve compatibility, we always check for a configured lookaside     sigstore first; if that is set up, we use the lookaside and ignore the     registry-native signature storage.  Usually the user would not bother to    set up the lookaside, and use the native mechanism.
    
See the individual commits for more details.

NOTE: This includes tests using the recording mechanism proposed in #254. They are in separate commits, so they can be dropped if that mechanism is rejected or controversial and would block merging this.

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -18,8 +18,10 @@ import (
 	"github.com/containers/image/docker/reference"
 	"github.com/containers/image/types"
 	"github.com/containers/storage/pkg/homedir"
+	"github.com/docker/distribution/registry/client"
 	"github.com/docker/go-connections/sockets"
 	"github.com/docker/go-connections/tlsconfig"
+	"github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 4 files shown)


docker/docker_client.go
```diff
@@ -412,7 +415,13 @@ func getAuth(ctx *types.SystemContext, registry string) (string, string, error)
 	return "", "", nil
 }
 
-func (c *dockerClient) ping() error {
+// ensurePropertiesAreDetected detects various properties of the registry.
+// See the dockerClient documentation for members which are affected by this.
+func (c *dockerClient) ensurePropertiesAreDetected() error {
```
runcom: too long even if such expressive. `GatherProperties | GetRegistryProperties` or such would be better to me (and I'm not a native english speaker/writer, I just know Golang doesn't need such expressiveness according to the official doc I couldn't find a link for :smile: )

docker/docker_image_dest.go
```diff
@@ -321,6 +347,79 @@ func (c *dockerClient) deleteOneSignature(url *url.URL) (missing bool, err error
...
+				break
+			}
+		}
+		sig := extensionSignature{
+			Version: extensionSignatureSchemaVersion,
+			Name:    signatureName,
+			Type:    extensionSignatureTypeAtomic,
+			Content: newSig,
+		}
+		body, err := json.Marshal(sig)
```
runcom: why `err` isn't checked here?

docker/docker_image_dest.go
```diff
@@ -321,6 +347,79 @@ func (c *dockerClient) deleteOneSignature(url *url.URL) (missing bool, err error
...
+
+		path := fmt.Sprintf(extensionsSignaturePath, reference.Path(d.ref.ref), d.manifestDigest.String())
+		res, err := d.c.makeRequest("PUT", path, nil, bytes.NewReader(body))
+		if err != nil {
+			return err
+		}
+		defer res.Body.Close()
+		if res.StatusCode != http.StatusCreated {
+			body, err := ioutil.ReadAll(res.Body)
+			if err == nil {
```
runcom: if we're just logging here I would rather do:
```
body, _ := ioutil.ReadAll(res.Body)
logrus.Debugf("Error body %v", string(body))
```
even if it's nil, no need to check that unless you really think logging `nil` will clutter the logs. Up to your preference.

runcom:
:+1:

docker/docker_client.go
mtrmac: ACK, `detectProperties` works just as well (and the callers don’t care that we cache the data, no need to expose that in the name.)

docker/docker_image_dest.go
mtrmac: Good catch, fixed. Thanks!

docker/docker_image_dest.go
mtrmac: This is a copy&paste from `PutManifest`, and the extra check does not really hurt at all. You are right that it would be fine not to have it, but when we already do, removing it would not make the code noticeably better IMHO.

(Perhaps one day we might consolidate all the “`res.StatusCode` is unexpected” error handling into a helper for consistency, but that would be a separate PR.)

mtrmac:
@runcom Updated, and I have dropped the go-vcr tests from this PR until we somehow decide on #254 etc.; I will track those in a separate PR.

runcom:
@mtrmac LGTM

mtrmac:
:+1:

@runcom Note that there are still outstanding questions on how this interacts with the `sigstore*:` lookaside. But because `sigstore*:` takes precedence, this at least should not break anything, so merging now.

PR #256, opened 2017-03-27
Tests for the X-Registry-Supports-Signatures extension

mtrmac:
This depends on, and includes, #254 and #255 , and adds tests for the new functionality introduced in #255. This is split from #255 so that the “how to test the docker registry client” discussion is not blocking merging of the code.

See individual commit messages for a bit more detail.


rhatdan:
@mtrmac @vrothberg  What should we do with this PR? Should we rebase and move forward or should we close it?

mtrmac:
The discussion whether to do this at all should happen in #254. Until then, rebased.

mtrmac:
Not valuable enough.

PR #257, opened 2017-03-28
docker/image_dest: do not deref possibly-nil results

lucab:
This avoids dereferencing a nil result when an error is
encountered on PATCH request.

Signed-off-by: Luca Bruno <luca.bruno@coreos.com>


mtrmac:
:+1: pending tests. Thanks!

runcom:
LGTM

PR #202, opened 2017-01-07
Avoid duplication in dockerImageDestionation.{Has,Put}Blob

mtrmac:
Call `HasBlob` in `dockerImageDestination.PutBlob` so that we only have one copy of the code.

Note that the first commit of this modifies `HasBlob` to fail on unexpected HTTP response statuses, instead of returning “blob not present”; this was modified in #63 but there hasn’t been an explicit explianation why, and right now, with the code duplicated, `PutBlob` is using its own implementation which aborts on such statuses, so AFAICT the change was either unintentional or ineffective. @nalind ?

PR #258, opened 2017-03-28
docker/image_dest: ask for pull action in token scope

lucab:
After PR #202, PutBlob() now goes through an initial HasBlob()
which may require authentication in order to perform the check;
in such case a push-only scoped token will fail with:
```
service="registry.docker.io",scope="repository:<name>:pull",error="insufficient_scope"
```

This fixes the above case by changing the scope of the token
requested for uploads, allowing pull+push.

Signed-off-by: Luca Bruno <luca.bruno@coreos.com>


runcom:
LGTM

mtrmac:
:+1:

Thanks!

(Note to self: `docker/docker/distribution.v2Pusher.Push` is also using both `push` and `pull`, so this is, if anything, likely to be more compatible than the previous code.)

PR #259, opened 2017-04-01
image: expose OCI configuration struct for all images

runcom:
We can't rely on people hacking around with blobs like https://github.com/kubernetes-incubator/cri-o/blob/master/pkg/storage/runtime.go#L213-L225

Instead, return OCI configuration and fix implementations :)

@mtrmac PTAL

### DIFF
--- a/image/docker_schema1.go
+++ b/image/docker_schema1.go
@@ -10,6 +10,7 @@ import (
 	"github.com/containers/image/manifest"
 	"github.com/containers/image/types"
 	"github.com/opencontainers/go-digest"
+	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 )
 
@@ -110,6 +111,17 @@ func (m *manifestSchema1) ConfigInfo() types.BlobInfo {
 // The result is cached; it is OK to call this however often you need.
 func (m *manifestSchema1) ConfigBlob() ([]byte, error) {
 	return nil, nil
+}
+
+// OCIConfig returns the image configuration as per OCI v1 image-spec. Information about
+// layers in the resulting configuration isn't guaranteed to be returned to due how
+// old image manifests work (docker v2s1 especially).
+func (m *manifestSchema1) OCIConfig() (*imgspecv1.Image, error) {
+	v2s2, err := m.convertToManifestSchema2(nil, nil)
+	if err != nil {
+		return nil, err
+	}
+	return v2s2.OCIConfig()
 }
 
 // LayerInfos returns a list of BlobInfos of layers referenced by this image, in order (the root layer first, and then successive layered layers).
@@ -243,10 +255,10 @@ func (m *manifestSchema1) convertToManifestSchema2(uploadedLayerInfos []types.Bl
 	if len(m.History) != len(m.FSLayers) {
 		return nil, errors.Errorf("Inconsistent schema 1 manifest: %d history entries, %d fsLayers entries", len(m.History), len(m.FSLayers))
 	}
-	if len(uploadedLayerInfos) != len(m.FSLayers) {
+	if uploadedLayerInfos != nil && len(uploadedLayerInfos) != len(m.FSLayers) {
 		return nil, errors.Errorf("Internal error: uploaded %d blobs, but schema1 manifest has %d fsLayers", len(uploadedLayerInfos), len(m.FSLayers))
 	}
-	if len(layerDiffIDs) != len(m.FSLayers) {
+	if layerDiffIDs != nil && len(layerDiffIDs) != len(m.FSLayers) {
 		return nil, errors.Errorf("Internal error: collected %d DiffID values, but schema1 manifest has %d fsLayers", len(layerDiffIDs), len(m.FSLayers))
 	}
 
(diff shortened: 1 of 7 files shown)


glestaris:
Thanks, @runcom - pretty awesome!

That'll save us the hassle as well: https://github.com/cloudfoundry/grootfs/blob/704c89a72a1f6a3d0125353a0ffb39380ada67c4/fetcher/remote/docker_src.go#L232-L236.

@spikymonkey ^^

runcom:
@glestaris cool, afaict this will allow you to drop stuff like https://github.com/cloudfoundry/grootfs/blob/704c89a72a1f6a3d0125353a0ffb39380ada67c4/fetcher/remote/docker_src.go#L241-L264 also

rhatdan:
:+1:

image/docker_schema1.go
```diff
@@ -112,6 +113,17 @@ func (m *manifestSchema1) ConfigBlob() ([]byte, error) {
 	return nil, nil
 }
 
+// Config returns the image configuration as per OCI v1 image-spec
+func (m *manifestSchema1) Config() (*imgspecv1.Image, error) {
```
mtrmac: This _really_ needs to share code with the existing conversion code; see what `manifestSchema1.convertToManifestSchema2`/`configJSONFromV1Config` does. (I guess this can chain through the schema1→schema2, and schema2→OCI conversions, instead of writing a new schema1→OCI one..)

image/docker_schema1.go
```diff
@@ -112,6 +113,17 @@ func (m *manifestSchema1) ConfigBlob() ([]byte, error) {
...
+	// Best effort to have an OCI configuration from a docker v2s1 manifest.
+	// we don't compute diffIDs or other v2s2 things here.
```
mtrmac: I do agree that blob size and DiffID computation in here would be _very_ awkward, and is on balance probably best left out, OTOH it is a _really_ surprising aspect of the data output and needs to be _very prominently_ featured in the interface documentation in `types.go` (if not in the actual method name… that would be overkill I guess).

mtrmac:
(This, especially, the schema1 code, is non-trivial enough that at least smoke tests, perhaps something like `TestConvertToManifestOCI`, converting a fixture and comparing with a fixture, seem useful. That’s absolutely not a blocker, though, entirely up to you whether you think it is worth your time.)

image/docker_schema1.go
mtrmac: Yeah, computing DiffIDs would be extremely awkward.

But modifying `manifestSchema1.convertToManifestSchema2` to accept `nil` values for the `uploadedLayerInfos` and `layerDiffIDs` parameters, and to put invalid/zero/other placeholder values in the generated config instead of dereferencing the arrays if they are `nil`, seems plausible. Sure, a bit ugly, but better than having two entirely different code paths I think.

runcom:
@mtrmac sure, this was a POC to gather feedback on the direction, I'll add tests ASAP

image/docker_schema1.go
runcom: I'll document this in the API in types.go, I don't really want to go thru schema1-schema2-oci conversions. People using this API won't actually look for those layers info, rather stuff like Entrypoint and such

types/types.go
```diff
@@ -202,6 +203,8 @@ type Image interface {
 	// ConfigBlob returns the blob described by ConfigInfo, iff ConfigInfo().Digest != ""; nil otherwise.
 	// The result is cached; it is OK to call this however often you need.
 	ConfigBlob() ([]byte, error)
+	// Config returns the image configuration as per OCI v1 image-spec
+	Config() (*v1.Image, error)
```
runcom: Yup, OCIConfig sounds good to me, thanks

image/docker_schema1.go
runcom: Alright then, I'll do that (forget all the other comment I did about not doing this way, I'll follow your suggestion :))

runcom:
@mtrmac PTAL (I've added a small silly test also, if you have other tests in mind you'd like to see here, please let me know or follow up)

image/docker_schema2.go
```diff
@@ -176,18 +181,26 @@ func (m *manifestSchema2) UpdatedImage(options types.ManifestUpdateOptions) (typ
 	return memoryImageFromManifest(&copy), nil
 }
 
-func (m *manifestSchema2) convertToManifestOCI1() (types.Image, error) {
+func (m *manifestSchema2) OCIConfigFromDocker() (*imgspecv1.Image, error) {
```
mtrmac: This is a private helper, so please name it starting with a lowercase letter.

Or perhaps just move the code into the public `OCIConfig` and call `m.OCIConfig` from `convertToManifestOCI1`; one less function.

image/docker_schema1.go
mtrmac: (The documentation update is still outstanding.)

mtrmac:
:+1:, feel free to merge after dealing with the helper method name and documenting the DiffID/layer size unreliability.

image/docker_schema2.go
runcom: silly me, could you take a last look at this?

mtrmac: Perfect.

runcom:
:+1: 

I'll merge after tests pass

PR #260, opened 2017-04-11
Copy docker StrSlice implementation to reduce dependency

Callisto13:
This (and the other story listed below) would be part of an effort to remove dependency on docker/docker.

https://www.pivotaltracker.com/story/show/134910101
https://www.pivotaltracker.com/story/show/134909997

Signed-off-by: Tiago Scolari <tscolari@pivotal.io>

### DIFF
--- a/image/manifest.go
+++ b/image/manifest.go
@@ -4,8 +4,8 @@ import (
 	"time"
 
 	"github.com/containers/image/manifest"
+	"github.com/containers/image/pkg/strslice"
 	"github.com/containers/image/types"
-	"github.com/docker/docker/api/types/strslice"
 	"github.com/opencontainers/go-digest"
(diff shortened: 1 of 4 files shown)


mtrmac:
Thanks!

I am not, in principle, thrilled with copy&pasting packages around, but this one is trivial enough, and version skew is a real issue here (also encountered in https://github.com/openshift/origin/pull/13585 ).

Just one nit, could you please add an explicit acknowledgment / reference to the original where it was taken from? An one-liner `pkg/strslice/README.md` naming the repo and perhaps version would be perfect.

:+1: with that link added.

runcom:
LGTM as well after addressing Miloslav's nit :)

tscolari:
We've added a README to the package as requested!

PR #261, opened 2017-04-13
storage: improve handling of references with digests

nalind:
This PR changes the way we parse references in the "storage" transport to correctly accept references that include an ID that has a "sha256:" prefix.  It defaults to adding the fully-expanded version of a reference's string value as the image's name, and adds a check to ensure that when attempting to read an image using a reference that included both a name and an ID, that the image we located using the ID actually has the specified name.

### DIFF
--- a/storage/storage_image.go
+++ b/storage/storage_image.go
@@ -71,14 +71,9 @@ type storageImage struct {
 
 // newImageSource sets us up to read out an image, which needs to already exist.
 func newImageSource(imageRef storageReference) (*storageImageSource, error) {
-	id := imageRef.resolveID()
-	if id == "" {
-		logrus.Errorf("no image matching reference %q found", imageRef.StringWithinTransport())
-		return nil, ErrNoSuchImage
-	}
-	img, err := imageRef.transport.store.GetImage(id)
+	img, err := imageRef.resolveImage()
 	if err != nil {
-		return nil, errors.Wrapf(err, "error reading image %q", id)
+		return nil, err
 	}
 	image := &storageImageSource{
 		imageRef:       imageRef,
@@ -341,16 +336,20 @@ func (s *storageImageDestination) Commit() error {
 	}
 	logrus.Debugf("created new image ID %q", img.ID)
 	s.ID = img.ID
+	names := img.Names
 	if s.Tag != "" {
-		// We have a name to set, so move the name to this image.
-		if err := s.imageRef.transport.store.SetNames(img.ID, []string{s.Tag}); err != nil {
+		names = append(names, s.Tag)
+	}
+	// We have names to set, so move those names to this image.
+	if len(names) > 0 {
+		if err := s.imageRef.transport.store.SetNames(img.ID, names); err != nil {
 			if _, err2 := s.imageRef.transport.store.DeleteImage(img.ID, true); err2 != nil {
 				logrus.Debugf("error deleting incomplete image %q: %v", img.ID, err2)
 			}
 			logrus.Debugf("error setting names on image %q: %v", img.ID, err)
 			return err
 		}
-		logrus.Debugf("set name of image %q to %q", img.ID, s.Tag)
+		logrus.Debugf("set names of image %q to %v", img.ID, names)
 	}
 	// Save the data blobs to disk, and drop their contents from memory.
 	keys := []ddigest.Digest{}
--- a/storage/storage_image.go
+++ b/storage/storage_image.go
@@ -331,10 +331,22 @@ func (s *storageImageDestination) Commit() error {
 	}
 	img, err := s.imageRef.transport.store.CreateImage(s.ID, nil, lastLayer, "", nil)
 	if err != nil {
-		logrus.Debugf("error creating image: %q", err)
-		return err
+		if err != storage.ErrDuplicateID {
+			logrus.Debugf("error creating image: %q", err)
+			return errors.Wrapf(err, "error creating image %q", s.ID)
+		}
+		img, err = s.imageRef.transport.store.GetImage(s.ID)
+		if err != nil {
+			return errors.Wrapf(err, "error reading image %q", s.ID)
+		}
+		if img.TopLayer != lastLayer {
+			logrus.Debugf("error creating image: image with ID %q exists, but uses different layers", err)
+			return errors.Wrapf(err, "image with ID %q already exists, but uses a different top layer", s.ID)
+		}
+		logrus.Debugf("reusing image ID %q", img.ID)
+	} else {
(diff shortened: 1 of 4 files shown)


runcom:
LGTM ping @mtrmac, @nalind I assume this is needed for CRI-O

nalind:
@runcom This is currently rolled up in https://github.com/kubernetes-incubator/cri-o/pull/450, so once that passes muster, I think we should merge this.

storage/storage_image.go
```diff
@@ -341,16 +355,31 @@ func (s *storageImageDestination) Commit() error {
...
+	if s.imageRef.DockerReference() != nil {
+		names = append(names, verboseName(s.imageRef.DockerReference()))
+	}
 	if s.Tag != "" {
-		// We have a name to set, so move the name to this image.
-		if err := s.imageRef.transport.store.SetNames(img.ID, []string{s.Tag}); err != nil {
+		// Strip off any digest suffix, and if that leaves anything,
+		// use that name, too.
+		tag := s.Tag
+		if i := strings.Index(tag, "@"); i != -1 {
```
mtrmac: Similar to above: AFAICT `s.Tag` comes from `s.imageRef.reference`, i.e. `refname` in `.ParseStoreReference()`, which is set based on `name` in the same function, and `name` is set based on `refInfo[0]` which comes from `SplitN(ref, "@", 2)`, i.e. `name` in there, and by extension `s.Tag`, can never contain a `@`.

Am I missing a code path to create a `storageReference `with different properties, perhaps?

storage/storage_transport.go
```diff
@@ -100,9 +98,12 @@ func (s storageTransport) ParseStoreReference(store storage.Store, ref string) (
...
 		}
-		sum, err = digest.Parse("sha256:" + refInfo[1])
-		if err != nil {
-			return nil, err
+		sum, err = digest.Parse(refInfo[1])
+		if err != nil || sum.Validate() != nil {
+			sum, err = digest.Parse("sha256:" + refInfo[1])
+			if err != nil || sum.Validate() != nil {
+				return nil, err
+			}
```
nalind: The storage library uses hex values as identifiers, and whether the incoming string contains an algorithm name as a prefix for a hex value or not, we want to use a hex value as the identifier at that level.

storage/storage_image.go
```diff
@@ -80,6 +81,19 @@ func newImageSource(imageRef storageReference) (*storageImageSource, error) {
...
+			if name == imageRef.reference {
+				nameMatch = true
+				break
+			}
+		}
+		if !nameMatch {
+			logrus.Errorf("no image matching reference %q found", imageRef.StringWithinTransport())
+			return nil, ErrNoSuchImage
+		}
+	}
```
nalind: Yeah, probably.  I'll add it.

storage/storage_transport.go
```diff
@@ -284,8 +285,12 @@ func (s storageTransport) ValidatePolicyConfigurationScope(scope string) error {
 func verboseName(name reference.Named) string {
 	name = reference.TagNameOnly(name)
 	tag := ""
+	id := ""
 	if tagged, ok := name.(reference.NamedTagged); ok {
-		tag = tagged.Tag()
+		tag = ":" + tagged.Tag()
+	}
+	if canonical, ok := name.(reference.Canonical); ok {
+		id = "@" + canonical.Digest().Hex()
```
mtrmac: This one is also reachable through `storageTransport.GetStoreImage`, which is, possibly, callable with any kind of `types.ImageReference`.  So, if `contaiers/storage` wants a hex string, this addition seems correct.

storage/storage_image.go
```diff
@@ -341,16 +355,31 @@ func (s *storageImageDestination) Commit() error {
 	}
 	logrus.Debugf("created new image ID %q", img.ID)
 	s.ID = img.ID
+	names := []string{}
+	if s.imageRef.DockerReference() != nil {
+		names = append(names, verboseName(s.imageRef.DockerReference()))
+	}
 	if s.Tag != "" {
```
nalind: Yeah, you're right, we don't need to add Tag as a name.  That simplifies things.

storage/storage_image.go
nalind: No, you're right.  The name that was being attached with an ID suffix was coming from cri-o, so this step didn't need to be taken.

nalind:
Okay, I think the changes discussed above are in, and the patches have been been rebased.

storage/storage_reference.go
```diff
@@ -108,6 +109,23 @@ func (s storageReference) DeleteImage(ctx *types.SystemContext) error {
...
+			if name == s.reference {
+				nameMatch = true
+				break
+			}
+		}
+		if !nameMatch {
+			logrus.Errorf("no image matching reference %q found", s.StringWithinTransport())
+			return ErrNoSuchImage
+		}
+	}
```
mtrmac: Is the `resolveID`+`.GetImage`+`nameMatch` loop, which seems to be exactly the same in `newImageSource` and `DeleteImage`, likely to diverge? It might be worth sharing these 20 lines in a function.

(But it is quite fine as it is as well.)

storage/storage_image.go
```diff
@@ -336,21 +349,34 @@ func (s *storageImageDestination) Commit() error {
 	}
 	img, err := s.imageRef.transport.store.CreateImage(s.ID, nil, lastLayer, "", nil)
 	if err != nil {
-		logrus.Debugf("error creating image: %q", err)
-		return err
+		if err != storage.ErrDuplicateID {
+			logrus.Debugf("error creating image: %q", err)
+			return err
+		}
+		img, err = s.imageRef.transport.store.GetImage(s.ID)
```
mtrmac: This ignores `err`, so in principle could crash if `img` is `nil`?

AFAICT `s.ID`, via `ParseStoreReference`… `newImageDestination`, can end up empty. What does `GetImage` do with that case? Or is the argument that `ErrDuplicateID` necessarily means that `GetImage` will succeed and return an image?

storage/storage_reference.go
mtrmac: … or fold that into a function which subsumes the existing `resolveID` perhaps? Either way, non-blocking.

nalind: Ah, good catch.  Replacing `resolveID` with a `resolveImage` that folds in the additional logic makes sense.

nalind:
Okay, added that error check and refactored logic that followed `resolveID` into it, renaming it to `resolveImage`.

mtrmac:
:+1: Thanks!

@runcom Do you want to take another look? Or just merge.

runcom:
Thanks Miloslav

PR #262, opened 2017-04-24
ostree: fix usermode pull

giuseppe:
I've noticed this while working on the integration with atomic...

UntarPath would just fail when used from a non root user.

The temporary workaround is using GNU Tar.  This part should be fixed in
storage/pkg/archive.

The fix ensures every directory and file is readable/writeable by the
user.

### DIFF
--- a/ostree/ostree_dest.go
+++ b/ostree/ostree_dest.go
@@ -119,6 +119,32 @@ func (d *ostreeImageDestination) PutBlob(stream io.Reader, inputInfo types.BlobI
 	return types.BlobInfo{Digest: computedDigest, Size: size}, nil
 }
 
+func fixUsermodeFiles(dir string) error {
+	entries, err := ioutil.ReadDir(dir)
+	if err != nil {
+		return err
+	}
+
+	for _, info := range entries {
+		fullpath := filepath.Join(dir, info.Name())
+		if info.IsDir() {
+			if err := os.Chmod(dir, info.Mode()|0700); err != nil {
+				return err
+			}
+			err = fixUsermodeFiles(fullpath)
+			if err != nil {
+				return err
+			}
+		} else if info.Mode().IsRegular() {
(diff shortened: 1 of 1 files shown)


mtrmac:
…I guess this works.

Isn’t this possible to achieve the same thing with `containers/storage/pkg/archive.TarOptions`? I suppose we would still run into the inability to set SELinux xattrs…

Anyway, the tests fail on formatting. `gofmt -s -w ostree/ostree_dest.go` I think.

giuseppe:
@mtrmac the issue I've found is that when directories are created then the mode specified in the tar header is set immediately.  If the directory has no write permissions then extracting files there would fail for a non-root user.  We would need the equivalent of `tar --delay-directory-restore` or set the correct chmod for the user when creating a directory (the os.Mkdir in archive.go/createTarFile).

My proposed solution is just a quick way to get this done so that we could start integrating it into atomic without losing the currently supported features.

mtrmac:
:+1:

Thanks; if this is a temporary workaround, then is there at least an issue filed to fix this in the ideal manner?

Anyway, this is a local matter with no impact on the rest, so LGTM.

runcom:
:+1:

ostree/ostree_dest.go
```diff
@@ -119,6 +119,32 @@ func (d *ostreeImageDestination) PutBlob(stream io.Reader, inputInfo types.BlobI
...
+		return err
+	}
+
+	for _, info := range entries {
+		fullpath := filepath.Join(dir, info.Name())
+		if info.IsDir() {
+			if err := os.Chmod(dir, info.Mode()|0700); err != nil {
+				return err
+			}
+			err = fixUsermodeFiles(fullpath)
```
mtrmac: (This could probably be simpler with `filepath.Walk`, OTOH this implementation uses less memory with very large directories, so neither implementation is obviously preferable.)

PR #265, opened 2017-05-04
storage: correctly read tar header

runcom:

### DIFF
--- a/storage/storage_image.go
+++ b/storage/storage_image.go
@@ -138,9 +138,9 @@ func (s *storageImageDestination) putBlob(stream io.Reader, blobinfo types.BlobI
 		Size:   -1,
 	}
 	// Try to read an initial snippet of the blob.
-	header := make([]byte, 10240)
-	n, err := stream.Read(header)
-	if err != nil && err != io.EOF {
+	buf := [archive.HeaderSize]byte{}
+	n, err := io.ReadAtLeast(stream, buf[:], len(buf))
+	if err != nil && err != io.EOF && err != io.ErrUnexpectedEOF {
 		return errorBlobInfo, err
 	}
 	// Set up to read the whole blob (the initial snippet, plus the rest)
@@ -154,9 +154,9 @@ func (s *storageImageDestination) putBlob(stream io.Reader, blobinfo types.BlobI
 	}
 	hash := ""
 	counter := ioutils.NewWriteCounter(hasher.Hash())
-	defragmented := io.MultiReader(bytes.NewBuffer(header[:n]), stream)
+	defragmented := io.MultiReader(bytes.NewBuffer(buf[:n]), stream)
 	multi := io.TeeReader(defragmented, counter)
-	if (n > 0) && archive.IsArchive(header[:n]) {
+	if (n > 0) && archive.IsArchive(buf[:n]) {
 		// It's a filesystem layer.  If it's not the first one in the
 		// image, we assume that the most recently added layer is its
 		// parent.


runcom:
@nalind @mtrmac PTAL this fixes projectatomic/skopeo#334

mtrmac:
What is the bug and which part of this is the fix? It’s not immediately obvious to me.

(Using the `MultiReader` was nice in that after the initial header, it wasn’t necessary to copy the hundreds of megabytes of data through another buffer. That’s of course not critical, but it would be nice to preserve that—hence my question of what exactly is the bug/fix in this.)

runcom:
Afaict the initial `stream.Read` was advancing the pointer too far corrupting the reader in `IsArchive` and make it returning `io.EOF`

runcom:
`Peek` on the other end doesn't advance the pointer in the reader

mtrmac:
But `IsArchive` accepts a `[]byte`, not a stream, and `10240` >> `archive.HeaderSize`. So was this because `stream.Read()` was for some reason returning a much shorter amount of data? (That would be plausible, with `compression.DetectCompresssion` peeking only 8 bytes, and `io.MultiReader` only using one of the input readers at a time.)

In that case, perhaps the fix could be just using `io.ReadAtLeast` and ignoring `ErrUnexpectedEOF`, when reading into `header`?

runcom:
I  can try that sure, and you are correct `Read` was indeed returning less bytes, right.

runcom:
@mtrmac edited, PTAL

storage/storage_image.go
```diff
@@ -137,10 +137,9 @@ func (s *storageImageDestination) putBlob(stream io.Reader, blobinfo types.BlobI
 		Digest: "",
 		Size:   -1,
 	}
-	// Try to read an initial snippet of the blob.
-	header := make([]byte, 10240)
-	n, err := stream.Read(header)
-	if err != nil && err != io.EOF {
+	buf := [archive.HeaderSize]byte{}
+	n, err := io.ReadAtLeast(stream, buf[:], len(buf))
+	if err != nil && err != io.EOF && err != io.ErrUnexpectedEOF {
```
TomSweeneyRedHat: I'm not feeling warm and fuzzy including ErrUnexpectedEOF in this if statement.  If we get an error in the middle of reading a fixed-size block or data structure, I think we ought to error out.  My $.02.  Take it with a grain of salt that I'm still coming up to speed in golang.

storage/storage_image.go
```diff
@@ -137,10 +137,9 @@ func (s *storageImageDestination) putBlob(stream io.Reader, blobinfo types.BlobI
 		Digest: "",
 		Size:   -1,
 	}
-	// Try to read an initial snippet of the blob.
-	header := make([]byte, 10240)
-	n, err := stream.Read(header)
-	if err != nil && err != io.EOF {
+	buf := [archive.HeaderSize]byte{}
+	n, err := io.ReadAtLeast(stream, buf[:], len(buf))
```
mtrmac: Is there a difference between `buf[:]` and `buf` here? Both are slices, and `stream.Read(slice)` was fine above.

storage/storage_image.go
mtrmac: In general, `ErrUnexpectedEOF` would indeed be a problem. But in this case we don’t yet know whether there is a fixed-size block of any kind in the input: the input may be an uncompressed tar (which does have to start with a 512-byte block), a compressed file (where the initial header may be much smaller), or an uncompressed JSON (which can quite plausibly be smaller than 512 bytes).

The only reason to use `io.ReadAtLeast` here is that we do read all of `archive.HeaderSize` _if_ the input is that large, even if a single `stream.Read()` would return a smaller single chunk of data.

storage/storage_image.go
mtrmac: Stupid me, `[`size`]`type is an array, not a slice.

mtrmac:
:+1:

storage/storage_image.go
```diff
@@ -137,10 +137,9 @@ func (s *storageImageDestination) putBlob(stream io.Reader, blobinfo types.BlobI
 		Digest: "",
 		Size:   -1,
 	}
-	// Try to read an initial snippet of the blob.
```
mtrmac: I wouldn’t mind preserving this comment, especially when the comment below refers to “the initial snippet” from this comment. (Absolutely non-blocking.)

runcom:
:+1: re-added the comment as per @mtrmac request also, merging

PR #264, opened 2017-04-29
Support run-time manifest type support (e.g. schema1/schema2) autodetection

mtrmac:
This modifies `copy.Image` to try a sequence of possibly supported manifest MIME types instead of only a single preferred one; notably this allows using the `docker:` transport with schema1-only registries, although the mechanism is ultimately fully general (i.e. it can deal with OCI-only registries just as well if `SupportsManifestMIMEType` includes OCI).

See the individual commit messages for detailed description;

**NOTE** The final commits, starting with “Define a ManifestTypeRejectedError for ImageDestination.PutManifest”, implement the behavior of only trying a sequence of MIME types if the initial manifest is rejected with a recognized "this is an invalid manifest” error (as opposed to anything else, e.g. insufficient permissions or out of space). This is IMHO more sensible, but it _does_ differ from docker/docker does, which AFAICT falls back from schema2 to schema1 on _any_ error.

(WIP because I don’t have the entire set of integration tests written, in particular using non-OpenShift registries; but the primary motivation, support for the OpenShift `acceptschema2` configuration option, does work correctly. **EDIT** Also the error messages might need tweaking.)
**EDIT** Integration tests exist, error messages are acceptable.

PR #266, opened 2017-05-06
Allow using schema2 when pushing to OpenShift

mtrmac:
With #264 we can dynamically try pushing using schema2, and fall back to schema1, without relying on a static `SupportedManifestMIMETypes` list. Therefore, modify the `atomic:` transport to use the same set of of statically declared MIME types candidates as the `docker:` transport.

In practical terms, OpenShift registries configured with `acceptschema2: true` will now get schema2 images if the input is schema2 (schema1 images would be copied unmodified as schema1), while `accpetschema2:false` registries will continue to always get schema1.

One notable difference is that `SupportedManifestMIMETypes` affects what we configure the _source_ to accept: e.g. when streaming from `docker:` to s1-only `atomic:`, previously the conversion was done by the `docker:` registry server, but now the source registry gives us schema2 and we do the conversion, changing the manifest digest.  Right now that is a difference, but soon(will add link here) we will update the manifest to use the destination’s name/tag, and then the manifest digest will change in any case.

This PR depends on, and includes, #264, only the last commit is actually new. I will rebase as necessary.

@aweiteka @baude FYI

### DIFF
--- a/openshift/openshift.go
+++ b/openshift/openshift.go
@@ -338,10 +338,7 @@ func (d *openshiftImageDestination) Close() error {
 }
 
 func (d *openshiftImageDestination) SupportedManifestMIMETypes() []string {
-	return []string{
-		manifest.DockerV2Schema1SignedMediaType,
-		manifest.DockerV2Schema1MediaType,
-	}
+	return d.docker.SupportedManifestMIMETypes()
 }
 
 // SupportsSignatures returns an error (to be displayed to the user) if the destination certainly can't store signatures.


runcom:
LGTM

copy/copy.go
```diff
@@ -210,54 +209,58 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
...
+	// and at least with the OpenShift registry "acceptschema2" option, there is no way to detect the support
+	// without actually trying to upload something and getting a types.ManifestTypeRejectedError.
+	// So, try the preferred manifest MIME type. If the process succeeds, fine…
+	manifest, err := ic.copyUpdatedConfigAndManifest()
+	if err != nil {
+		logrus.Debugf("Writing manifest using preferred type %s failed: %v", preferredManifestMIMEType, err)
+		// … if it fails, _and_ the failure is because the manifest is rejected, we may have other options.
+		if _, isManifestRejected := errors.Cause(err).(types.ManifestTypeRejectedError); !isManifestRejected || len(otherManifestMIMETypeCandidates) == 0 {
+			// We don’t have other options.
+			// In principle the code below would handle this as well, but the resulting  error message is fairly ugly. Do
```
TomSweeneyRedHat: nit.  Remove 'Do' from end of this comment.

copy/copy.go
```diff
@@ -210,54 +209,58 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
...
 		}
 		sigs = append(sigs, newSig)
 	}
 
-	writeReport("Writing manifest to image destination\n")
-	if err := dest.PutManifest(manifest); err != nil {
-		return errors.Wrap(err, "Error writing manifest")
-	}
-
 	writeReport("Storing signatures\n")
```
TomSweeneyRedHat: We're Storing the signatures, but we're not writing the manifest any more?  Is that correct or have I not had enough tea this morning?

copy/copy.go
```diff
@@ -322,6 +325,45 @@ func layerDigestsDiffer(a, b []types.BlobInfo) bool {
 	return false
 }
 
+// copyUpdatedConfigAndManifest updates the image per ic.manifestUpdates, if necessary,
+// stores the resulting config and manifest ot the destination, and returns the stored manifest.
```
TomSweeneyRedHat: nit 'manifest ot' to 'manifest to'

copy/copy.go
```diff
@@ -322,6 +325,45 @@ func layerDigestsDiffer(a, b []types.BlobInfo) bool {
...
+func (ic *imageCopier) copyUpdatedConfigAndManifest() ([]byte, error) {
+	pendingImage := ic.src
+	if !reflect.DeepEqual(*ic.manifestUpdates, types.ManifestUpdateOptions{InformationOnly: ic.manifestUpdates.InformationOnly}) {
+		if !ic.canModifyManifest {
+			return nil, errors.Errorf("Internal error: copy needs an updated manifest but that was known to be forbidden")
+		}
+		if !ic.diffIDsAreNeeded && ic.src.UpdatedImageNeedsLayerDiffIDs(*ic.manifestUpdates) {
+			// We have set ic.diffIDsAreNeeded based on the preferred MIME type returned by determineManifestConversion.
+			// So, this can only happen if we are trying to upload using one of the other MIME type candidates.
+			// Because UpdatedImageNeedsLayerDiffIDs only when converting from s1 to s2, this case should only arise
```
TomSweeneyRedHat: nit  "Because UpdatedImageNeedsLayerDiffIDs only" to "Because UpdatedImageNeedsLayerDiffIDs is used only" perhaps?

copy/copy.go
```diff
@@ -322,6 +325,45 @@ func layerDigestsDiffer(a, b []types.BlobInfo) bool {
...
+			// when ic.dest.SupportedManifestMIMETypes() includes both s1 and s2, the upload using s1 failed, and we are now trying s2.
+			// Supposedly s2-only registries do not exist or are extremely rare, so failing with this error message is good enough for now.
+			// If handling such registries turned out to be necessary, we could compute ic.diffIDsAreNeeded based on the full list of manifest MIME type candidates.
```
TomSweeneyRedHat: nit 'turned' to 'turns'

copy/copy.go
TomSweeneyRedHat: NVM, I see that's it's happening further down in the code.

copy/copy.go
```diff
@@ -322,6 +325,45 @@ func layerDigestsDiffer(a, b []types.BlobInfo) bool {
...
+		}
+		pendingImage = pi
+	}
+	manifest, _, err := pendingImage.Manifest()
+	if err != nil {
+		return nil, errors.Wrap(err, "Error reading manifest")
+	}
+
+	if err := ic.copyConfig(pendingImage); err != nil {
+		return nil, err
```
TomSweeneyRedHat: We've error text for the other errors, should we have an errors.Wrap(err, "Error copying configuration") here?

docker/docker_image_dest.go
```diff
@@ -233,16 +240,31 @@ func (d *dockerImageDestination) PutManifest(m []byte) error {
...
+	errors, ok := err.(errcode.Errors)
+	if !ok || len(errors) == 0 {
+		return false
+	}
+	ec, ok := errors[0].(errcode.ErrorCoder)
+	if !ok {
+		return false
+	}
+	// ErrorCodeManifestInvalid is returned by OpenShift with acceptschema2=false.
+	// ErrorCodeTagInvalid is returned by docker/distribution (at least as of ec87e9b6971d831f0eff752ddb54fb64693e51cd)
```
TomSweeneyRedHat: nit I think you want a date rather than 'ec87e...' inside the parens?  Or perhaps reword "at least as of" to identify what ec87e... is?

TomSweeneyRedHat:
Other than nits, LGTM.

mtrmac:
@TomSweeneyRedHat Thanks for the review! I'm sorry I haven’t noticed that the changes actually come from #264 , and I have already merged that. I will prepare a separate PR to clean this up.

mtrmac:
:+1:

copy/copy.go
mtrmac: Modified to “$method is true only …”

copy/copy.go
mtrmac: `errors.Wrap` is not supposed to exactly mirror the call stack (`errors.Wrap` is automatically recording that as well, separately); it adds context for the user to see what the error was about. 

The user really doesn’t care that `copyConfig` is a separate method (with a single caller, even—created just to keep the method sizes manageable). 

`copyConfig` itself, and `copyBlobFromtream` called from it, already add context to the operation (e.g., perhaps, `Error writing blob: [timeout]`) and either explicitly say that they are about the config, or are related to generic blob I/O where the config is not special enough to deserve an explicit mention.

docker/docker_image_dest.go
mtrmac: Made more explicit using “as of _commit_ …”

mtrmac:
> @TomSweeneyRedHat Thanks for the review! I'm sorry I haven’t noticed that the changes actually come from #264 , and I have already merged that. I will prepare a separate PR to clean this up.

#267

PR #267, opened 2017-05-09
Fix typos and unclear wording in comments

mtrmac:
This fixes issues pointed out in #266.


runcom:
Wondering if test failure is legit

mtrmac:
> Wondering if test failure is legit
`Error: Timeout waiting for test registry to become available`, disappeared after a few reruns.

But we seem to be starting/stopping the registry for every test, which seems excessive. I’ll look into it.

runcom:
LGTM (I'm fine following up for that flaky)

mtrmac:
:+1:

Issue #268, opened 2017-05-10
Avoid usage of github.com/mtrmac/gpgme


hferentschik:
This library only compiles on Linux, since it uses C extensions which won't compile on OS X and Windows. This makes it for example hard/impossible to use _image_ as a library for a cross compiled CLI applications. 

Could one not use Go's golang.org/x/crypto/openpgp?

For now I removed _github.com/mtrmac/gpgme_ and commented its use out from _mechanism.go_. If there is no way to make without C extensions, one can maybe only build them for Linux, whereas for OS X and Windows it particular feature would not be available (not 100% for what it is actually used).

runcom:
you can use build tag `containers_image_openpgp` to not use `gpgme` but getting only a limited signing features, @mtrmac could you chime in here?

hferentschik:
Yes, I've seen

> Optionally, you can use the containers_image_openpgp build tag (using go build -tags …, or make … BUILDTAGS=…). This will use a Golang-only OpenPGP implementation for signature verification instead of the default cgo/gpgme-based implementation; 

I have not got this to work yet though. Also from a user point of view it would be nice, if the default would work for all OS types.

That said, there is another problem with:

    vendor/github.com/containers/storage/storage/lockfile.go:136: st.Mtim undefined (type syscall.Stat_t has no field or method Mtim)

The call to _syscall.Stat_t_ is OS specific and returns different structs depending on the target OS. I guess that's an issue for _storage_.

hferentschik:
_containers_image_openpgp_ tag aside, I still cannot compile this out of the box. _mtrmac/gpgme_ is a defined dependency which gets downloaded into the vendor directory. When I then try to build, I get:

    # github.com/containers/image/vendor/github.com/mtrmac/gpgme
    vendor/github.com/mtrmac/gpgme/data.go:4:11: fatal error: 'gpgme.h' file not found
    #include <gpgme.h>
              ^
    1 error generated.

    # github.com/containers/image/vendor/github.com/containers/storage/storage
    vendor/github.com/containers/storage/storage/lockfile.go:136: st.Mtim undefined (type syscall.Stat_t has no field or method Mtim)
    vendor/github.com/containers/storage/storage/lockfile.go:136: not enough arguments in call to time.Unix

So even if containers/image has some build tags to avoid the usage of gpgme, gpgme will cause compilation errors. 

The second error is the mentioned system specific call to syscall.Stat_t in containers/storage.

mtrmac:
[...]

PR #271, opened 2017-05-12
Fix build and usage instructions in README.md

### LINKED ISSUE #270: Wrong build instructions in README
From the README:

> For ordinary use, go build ./... is sufficient.
>  
> When developing this library, please use make to take advantage of the tests and validation.

That is not quite correct, if you check out the source, you will first need to run _make vendor_ before being able to run _go build_. 

On this note I am wondering why the actual _go build_ is not integrated into the _Makefile_, depending on vendor. In fact, imo it should be the default target.

You should be able to checkout the sources and run _make_ and everything gets bootstrapped and compiled.

mtrmac:
- Add a “command-line usage” heading and section
- Enhance the dependencies subsection
- Move the “when developing this library” sentence into a “Contributing” section
- Add another skopeo pointer to the “building” subsection
- Fix the actual building instructions
- Document the gpgme dependency and how to install it

This should help with #268, and fixes #270.

(Re: #270, this completely removes any mention of `go build ./...` without replacement; as #268 shows, `./...` is undesirable because it includes all depdendencies, and `go build $(PACKAGES)` in `Makefile` is unnecessary because `go test $(PACKAGES)` builds all of the named subpackages, even those which don’t have any tests.)

@hferentschik is this sufficient? (The containers/storage build failure will be fixed separately.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,10 @@ What this project tests against dependencies-wise is located
 
 ## Building
 
+If you want to see what the library can do, or an example of how it is called,
+consider starting with the [skopeo](https://github.com/projectatomic/skopeo) tool
+instead.
+
 For ordinary use, `go build ./...` is sufficient.
 
 
--- a/README.md
+++ b/README.md
@@ -51,13 +51,18 @@ Ensure that the dependencies documented [in vendor.conf](https://github.com/cont
 are also available
 (using those exact versions or different versions of your choosing).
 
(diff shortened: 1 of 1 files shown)


mtrmac:
@runcom PTAL

runcom:
LGTM

mtrmac:
👍

PR #273, opened 2017-05-12
WIP DO NOT MERGE: Test on macOS as well

mtrmac:
This skips (make test-skopeo) because it depends on docker, which is not
available in Travis.

Do not merge yet, I want to see what Travis does…


mtrmac:
Not going to happen, “Go builds are not available on the OSX environment.”.

PR #274, opened 2017-05-13
[WIP] Add Gluster Transport for container image.

humblec:
Issue# https://github.com/projectatomic/skopeo/issues/110
Signed-off-by: Humble Chirammal <hchiramm@redhat.com>

### DIFF
--- a/gluster/explicitfilepath/path.go
+++ b/gluster/explicitfilepath/path.go
@@ -1,0 +1,56 @@
+package explicitfilepath
+
+import (
+	"os"
+	"path/filepath"
+
+	"github.com/pkg/errors"
+)
+
+// ResolvePathToFullyExplicit returns the input path converted to an absolute, no-symlinks, cleaned up path.
+// To do so, all elements of the input path must exist; as a special case, the final component may be
+// a non-existent name (but not a symlink pointing to a non-existent name)
+// This is intended as a a helper for implementations of types.ImageReference.PolicyConfigurationIdentity etc.
+func ResolvePathToFullyExplicit(path string) (string, error) {
+	switch _, err := os.Lstat(path); {
+	case err == nil:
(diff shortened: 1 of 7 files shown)


humblec:
@runcom PTAL :)

mtrmac:
@humblec This seems to be a straightforward copy of the `directory` subpackage, with no changes except for renaming; it seems that users of this copy could just as well use the existing `directory` package, right now. So I guess the final design will be rather different; Is there anything specific you want us to comment on now?

For now the only thing that jumps out is that the `explicitfilepath` subpackage should probably not be copied, it is somewhat tricky and maintaining a single correct copy would be better: import it from `directory/explicitfilepath` the way `oci:` and `ostree:` do, or perhaps we can move it to `containers/image/pkg/explicitfilepath` (or perhaps better `containers/image/internal/explicitfilepath` to use https://golang.org/cmd/go/#hdr-Internal_Directories ?). Or do you plan to change the semantics of that?

humblec:
@mtrmac Thanks!! .. Yeah, as mentioned in the subjected issue, it would look like  the replica of `dir` trasnport. However I am adding one functionality to this PR to make sure the  mount is  `gluster` mount point which will be different from `dir` transport. Once its added, I will change the `WIP` string from the PR title.  You are correct, more access types will be followed on `gluster` tranport.  Just starting it with `fuse` mount. Regarding `directory/explicitfilepath`, I will directly import it from `directory/explicitfilepath` in next patch set.

rhatdan:
What is the goal of this patch set?  Just to copy rootfs onto a gluster share?

rhatdan:
@mtrmac Lets close this pull request.

mtrmac:
Yeah; @humblec feel free to reopen or file another one if you return to this.

PR #277, opened 2017-05-26
Support /etc/docker/certs.d

mtrmac:
This adds `types.SystemContext.DockerPerHostCertDirPath`, defaulting to `/etc/docker/certs.d`, and making it work by default.

Note that this also applies the general TLS defaults to cases when they were previously ignored, notably even when TLS verification is disabled we now restrict the TLS version. That may not be the right thing to do, but if so, we should make an explicit decision about that.

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -129,10 +129,15 @@ func newTransport() *http.Transport {
 	return tr
 }
 
-func setupCertificates(dir string, tlsc *tls.Config) error {
-	if dir == "" {
-		return nil
+// dockerCertDir returns a path to a directory to be consumed by setupCertificates() depending on ctx, or "" if no directory should be used.
+func dockerCertDir(ctx *types.SystemContext) string {
+	if ctx != nil {
+		return ctx.DockerCertPath
 	}
+	return ""
+}
+
+func setupCertificates(dir string, tlsc *tls.Config) error {
 	fs, err := ioutil.ReadDir(dir)
 	if err != nil {
 		if os.IsNotExist(err) {
@@ -203,12 +208,14 @@ func newDockerClient(ctx *types.SystemContext, ref dockerReference, write bool, 
 	}
 	tr := newTransport()
 	tr.TLSClientConfig = serverDefault()
-	if ctx != nil && (ctx.DockerCertPath != "" || ctx.DockerInsecureSkipTLSVerify) {
-		if err := setupCertificates(ctx.DockerCertPath, tr.TLSClientConfig); err != nil {
+	certDir := dockerCertDir(ctx)
+	if certDir != "" {
+		if err := setupCertificates(certDir, tr.TLSClientConfig); err != nil {
 			return nil, err
 		}
-
-		tr.TLSClientConfig.InsecureSkipVerify = ctx.DockerInsecureSkipTLSVerify
+	}
+	if ctx != nil && ctx.DockerInsecureSkipTLSVerify {
+		tr.TLSClientConfig.InsecureSkipVerify = true
 	}
 	client := &http.Client{Transport: tr}
 
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -148,6 +148,7 @@ func dockerCertDir(ctx *types.SystemContext, hostPort string) string {
 }
 
 func setupCertificates(dir string, tlsc *tls.Config) error {
+	logrus.Debugf("Looking for TLS certificates and private keys in %s", dir)
 	fs, err := ioutil.ReadDir(dir)
 	if err != nil {
 		if os.IsNotExist(err) {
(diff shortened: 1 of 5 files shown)


mtrmac:
Do not merge yet, currently untested.

docker/docker_client.go
```diff
@@ -129,12 +138,29 @@ func newTransport() *http.Transport {
...
+	if ctx != nil && ctx.DockerCertPath != "" {
+		return ctx.DockerCertPath
+	}
+	var hostCertDir string
+	if ctx != nil && ctx.DockerPerHostCertDirPath != "" {
+		hostCertDir = ctx.DockerPerHostCertDirPath
+	} else if ctx != nil && ctx.RootForImplicitAbsolutePaths != "" {
+		hostCertDir = filepath.Join(ctx.RootForImplicitAbsolutePaths, systemPerHostCertDirPath)
+	} else {
+		hostCertDir = systemPerHostCertDirPath
```
TomSweeneyRedHat: Does it make sense to add a log debug statement here?  I'm assuming this is something we don't want to happen often and it might be nice to easily debug if we need to.  Your call.

runcom:
looks fine to me overall (module tests) - we're basically replicating the behavior we already have with registries credentials where, 1) if someone pass creds via `Ctx` (skopeo flags), use them, 2) otherwise, look in `$HOME/.docker/config.json`. Here will be, 1) if someone pass cert dir via `Ctx` (skopeo flags), use it, 2) otherwise look in `/etc/docker/...`. @mtrmac is my understand right?

docker/docker_client.go
```diff
@@ -199,18 +225,20 @@ func newDockerClient(ctx *types.SystemContext, ref dockerReference, write bool,
...
 			return nil, err
 		}
-
-		tlsc.InsecureSkipVerify = ctx.DockerInsecureSkipTLSVerify
-		tr.TLSClientConfig = tlsc
 	}
-	if tr.TLSClientConfig == nil {
-		tr.TLSClientConfig = serverDefault()
+	if ctx != nil && ctx.DockerInsecureSkipTLSVerify {
+		tr.TLSClientConfig.InsecureSkipVerify = true
```
TomSweeneyRedHat: Does this get set to default when initialized?  If not, do you need an else here to do so?

mtrmac: Yes, all fields are default-initialized on Go. (Both the original and updated code depends on that if the user does not pass any configuration; now we just depend on it in the `ctx != nil` case as well.)

docker/docker_client.go
mtrmac: The quoted line (`hostCertDir = systemPerHostCertDirPath`) is actually expected to be the common case. But the debug log _should_ record which path was tried, thanks for the reminder.

mtrmac:
> Here will be, 1) if someone pass cert dir via `Ctx` (skopeo flags), use it, 2) otherwise look in `/etc/docker/`.... @mtrmac is my understand right?

Yes, except that 1) is a bit more complex than `registries.d`, more cases to handle. See `dockerCertDir`.

docker/docker_client.go
mtrmac: The path is now logged by the caller (`newDockerClient`).

mtrmac: Ultimately moved the logging into `setupCertificates`.

mtrmac:
Tested per https://bugzilla.redhat.com/show_bug.cgi?id=1455575 , ready for review.

docker/docker_client.go
```diff
@@ -48,6 +48,15 @@ const (
 	extensionSignatureTypeAtomic    = "atomic" // extensionSignature.Type
 )
 
+// systemPerHostCertDirPath is the path to certs.d, used for storing per-host TLS keys and certificates.
+// You can override this at build time with
+// -ldflags '-X github.com/containers/image/docker.systemPerHostCertDirPath=$your_path'
+var systemPerHostCertDirPath = builtinPerHostCertDirPath
```
runcom: I don't understand why are we making this configurable at compile time. This is the docker transport and as such, the default is `/etc/docker/certs.d` unless I'm missing something. Configuring this at compile time goes against what we already have for registries auths with `$HOME/.docker/config.json` - we don't let ppl override that at compile time. That's something implementation specific and I believe `/etc/docker/certs.d` is as well.

I feel I'm missing something as to why we need this compile time stuff so bear with me.

docker/docker_client.go
```diff
@@ -48,6 +48,15 @@ const (
...
+
+// builtinPerHostCertDirPath is the path to certs.d.
+// DO NOT change this, instead see systemPerHostCertDirPath above.
+const builtinPerHostCertDirPath = "/etc/docker/certs.d"
```
runcom: I can't imagine cases where people want to change this. If people have certs in another location they'll use `DockerCertPath` and I won't leave them the ability to ovveride this at build time - they'll come here with some issue and we first need to realize they built skopeo with another buildinPerHostCertDirPath and work from there. I'd rather have them explicitely use `DockerCertPath` instead.

runcom:
LGTM

mtrmac:
👍

PR #279, opened 2017-05-30
ostree: use golang bindings

giuseppe:
the main advantage, beside not requiring external processes, is that we
can commit all the layers as an unique transaction.

### DIFF
--- a/ostree/ostree_dest.go
+++ b/ostree/ostree_dest.go
@@ -11,12 +11,15 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/containers/image/manifest"
 	"github.com/containers/image/types"
 	"github.com/containers/storage/pkg/archive"
 	"github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
+
+	"github.com/ostreedev/ostree-go/pkg/otbuiltin"
 )
 
 type blobToImport struct {
@@ -158,7 +161,17 @@ func fixFiles(dir string, usermode bool) error {
 	return nil
 }
 
-func (d *ostreeImageDestination) importBlob(blob *blobToImport) error {
+func (d *ostreeImageDestination) ostreeCommit(repo *otbuiltin.Repo, branch string, root string, metadata []string) error {
+	opts := otbuiltin.NewCommitOptions()
+	opts.AddMetadataString = metadata
+	opts.Timestamp = time.Now()
+	// OCI layers have no parent OSTree commit
+	opts.Parent = "0000000000000000000000000000000000000000000000000000000000000000"
+	_, err := repo.Commit(root, branch, opts)
+	return err
+}
+
+func (d *ostreeImageDestination) importBlob(repo *otbuiltin.Repo, blob *blobToImport) error {
 	ostreeBranch := fmt.Sprintf("ociimage/%s", blob.Digest.Hex())
 	destinationPath := filepath.Join(d.tmpDirPath, blob.Digest.Hex(), "root")
 	if err := ensureDirectoryExists(destinationPath); err != nil {
@@ -186,11 +199,7 @@ func (d *ostreeImageDestination) importBlob(blob *blobToImport) error {
 			return err
 		}
 	}
-	return exec.Command("ostree", "commit",
-		"--repo", d.ref.repo,
-		fmt.Sprintf("--add-metadata-string=docker.size=%d", blob.Size),
-		"--branch", ostreeBranch,
-		fmt.Sprintf("--tree=dir=%s", destinationPath)).Run()
+	return d.ostreeCommit(repo, ostreeBranch, destinationPath, []string{fmt.Sprintf("docker.size=%d", blob.Size)})
 }
 
 func (d *ostreeImageDestination) importConfig(blob *blobToImport) error {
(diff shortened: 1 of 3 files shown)


giuseppe:
@runcom @mtrmac do you have any suggestion on how I could fix these errors?  I've updated the travis image to Zesty for having libostree-dev but I got some other failures

mtrmac:
:+1: pending tests (though note that I didn’t check whether the ostree bindings are being used correctly).

As for the build failures, https://lint.travis-ci.org/ reports errors in the section which is supposed to install packages. Perhaps we have just been very lucky so far that the packages have been installed by default?!

giuseppe:
thanks for the review.  It seems though that the travis lint has an issue and reports the wrong error: https://github.com/travis-ci/travis-yaml/issues/77

I'll investigate what is going wrong here

giuseppe:
@mtrmac it seems there is no way to get an updated Ubuntu except run the tests in a container.  I have done that but now it it complaining that the ostree go bindings are not in the skopeo `vendor.conf` file and the skopeo build fails.  Do you have any suggestion on how to address this?

mtrmac:
@giuseppe The vendoring is partially documented in the last part of https://github.com/projectatomic/skopeo#dependencies-management (it’s fine that `test-skopeo` fails _as long as it runs correctly_), if there is a skopeo PR demonstrating a corresponding update.

Though, I expect this to run into another problem – `make test-skopeo` creates a container to run its tests, so running `make test-skopeo` in a container will probably fail. (And yes we do need the `make test-skopeo` test on every containers/image PR, from bitter experience, though _how exactly_ we run those tests may of course change.)

mtrmac:
@giuseppe Perhaps one way to do this:

1. Create a separate PR which moves the test execution into a container, without changing the code/dependencies (so `make test-skopeo` should continue to work). Get it merged.
2. Then layer this PR on top (with an expected-broken `make test-skopeo`)
3. Add a skopeo PR which vendors from this PR, and updates skopeo’s dependencies to show that skopeo tests can pass in the skopeo repo. 
4. Based on 3., merge 2.
5. Update 3. to vendor from `master`, and merge it.

@runcom With the `gpgme` choice flaring up again, the recent containers/storage build difficulties, and now this making it difficult for users of only _slightly_ older distributions, it would be Really Nice™ if we can figure out a way to make all of this easy for users. I don’t know yet if or how that can be done (I only vaguely recall suggestions of a `!cgo` build condition); do you have any ideas?

ostree/ostree_dest.go
```diff
@@ -11,12 +11,15 @@ import (
...
 	"strings"
+	"time"
 
 	"github.com/containers/image/manifest"
 	"github.com/containers/image/types"
 	"github.com/containers/storage/pkg/archive"
 	"github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
+
+	"github.com/ostreedev/ostree-go/pkg/otbuiltin"
```
runcom: so this is now importing cgo code if I've understood correctly, and as @mtrmac pointed out, we also need a way to opt in to have this (as opposed as to have this cgo by default). Would it make sense to have a `ostree_dest_nocgo.go` which still uses the non-c bindings? Or well, just have build tags at least?

mtrmac: @runcom per the other scattered conversations… I guess it would be nice to have a single “I don’t want to care, give me the no-dependencies subset” switch (as long as that subset is useful to a fair amount of users), instead of a dozen flags. I haven’t yet looked into whether that is possible/easy, I guess it’s now quite urgent.

As far as the effect, I think just making the `alltransports` inclusion conditional would be quite enough?
- We really don’t want to maintain two different `ostree` implementations
- Users who cherry-pick a transport don’t have to care about anything
- Users of `alltransports` would either set up the build environment as necessary, or opt out using the TBD mechanism.

runcom: agree on everything, well, but I don't think a `!cgo` to remove the whole c deps is fully correct :/

giuseppe:
@mtrmac the Skopeo temporary branch with the version of containers/image in this PR passes the tests (https://github.com/projectatomic/skopeo/pull/351).

mtrmac:
@giuseppe Perfect. I need to leave now, we can get that merged tomorrow. In the mean time, can you rebase this branch, please?

giuseppe:
yes, sure.  Just pushed a rebased version

mtrmac:
@runcom PTAL.

runcom:
seems ok overall, even if we still need to figure out how to entirely disable cgo for windows/mac

mtrmac:
> seems ok overall, even if we still need to figure out how to entirely disable cgo for windows/mac

I’ll take a look at that this week, I hope.

giuseppe:
is this blocked by disabling cgo on windows/mac?

mtrmac:
@runcom ping

runcom:
Not blocked on that

giuseppe:
so what more is needed here?

runcom:
I think we're good to go here, @mtrmac?

mtrmac:
@runcom The immediate next step would be your +1.

runcom:
LGTM :)

mtrmac:
@giuseppe Please rebase and re-vendor https://github.com/projectatomic/skopeo/pull/351 from the just-merged master branch.

giuseppe:
@mtrmac done

PR #280, opened 2017-05-31
storage: add support for manifest list

runcom:
`storage` should support manifest list like `docker-daemon` - this is needed in CRI-O

### DIFF
--- a/storage/storage_image.go
+++ b/storage/storage_image.go
@@ -416,8 +416,15 @@ func (s *storageImageDestination) Commit() error {
 	return nil
 }
 
+var manifestMIMETypes = []string{
+	// TODO(runcom): we'll add OCI as part of another PR here
+	manifest.DockerV2Schema2MediaType,
+	manifest.DockerV2Schema1SignedMediaType,
+	manifest.DockerV2Schema1MediaType,
+}
+
 func (s *storageImageDestination) SupportedManifestMIMETypes() []string {
-	return nil
+	return manifestMIMETypes
 }
 
 // PutManifest writes manifest to the destination.


mrunalp:
:+1:

storage/storage_image.go
```diff
@@ -416,8 +416,23 @@ func (s *storageImageDestination) Commit() error {
...
 }
 
+var manifestMIMETypes = []string{
+	// TODO(runcom): we'll add OCI as part of another PR here
+	manifest.DockerV2Schema2MediaType,
+	manifest.DockerV2Schema1SignedMediaType,
+	manifest.DockerV2Schema1MediaType,
+}
+
+func supportedManifestMIMETypesMap() map[string]bool {
```
mtrmac: This is not used at all?

storage/storage_image.go
```diff
@@ -416,8 +416,23 @@ func (s *storageImageDestination) Commit() error {
 	return nil
 }
 
+var manifestMIMETypes = []string{
+	// TODO(runcom): we'll add OCI as part of another PR here
+	manifest.DockerV2Schema2MediaType,
+	manifest.DockerV2Schema1SignedMediaType,
+	manifest.DockerV2Schema1MediaType,
+}
```
mtrmac: The PR name says “support manifest list”, but this
- actually _restricts_ MIME types (returning `nil` from `SupportedManifestMIMETypes` meant “anything is acceptable”)
- does not mention manifest lists at all?!

Is the PR really doing what the title says, somehow, or is the title wrong?

storage/storage_image.go
runcom: It isn't, I'll drop it.

storage/storage_image.go
runcom: if supportedmanifestmimetypes is nil, we do accept everything as if `IsMultiImage` returns true and copy will fail. If we list some manifest types here we do get a manifest v2s2 from the original manifest list and we'll be able to copy.

mtrmac:
:+1: pending tests.

runcom:
lgtm

PR #282, opened 2017-06-02
copy: fail early when image os doesn't match host os

runcom:
We don't want to pull all layers for an image before realizing it's incompatible with the host OS. An example would be when you try to copy a windows image with skopeo:

```
$ ./skopeo copy docker://microsoft/windowsservercore docker-daemon:testwin:latest
```
The above will pull all layers (more than 3GB) before realizing it's incompatible.

We want to fail early as docker itself does today, so this patch modifies the flow to check for OS compatibility asap (relying on the OCI config):
```
$ sudo ./skopeo copy docker://microsoft/windowsservercore docker-daemon:testwin:latest
FATA[0003] image operating system "windows" cannot be used on "linux"
```

@mtrmac PTAL

needed for BZ https://bugzilla.redhat.com/show_bug.cgi?id=1418173

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"io/ioutil"
 	"reflect"
+	"runtime"
 	"strings"
 	"time"
 
@@ -157,6 +158,10 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 		}
 	}()
 
+	if err := checkImageDestinationForCurrentRuntimeOS(src, dest); err != nil {
+		return err
+	}
+
 	if src.IsMultiImage() {
 		return errors.Errorf("can not copy %s: manifest contains multiple images", transports.ImageName(srcRef))
 	}
@@ -274,6 +279,22 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 		return errors.Wrap(err, "Error committing the finished image")
 	}
 
+	return nil
+}
+
+func checkImageDestinationForCurrentRuntimeOS(src types.Image, dest types.ImageDestination) error {
+	if dest.MustMatchRuntimeOS() {
+		c, err := src.OCIConfig()
+		if err != nil {
+			return errors.Wrapf(err, "Error parsing image configuration")
+		}
+		osErr := fmt.Errorf("image operating system %q cannot be used on %q", c.OS, runtime.GOOS)
+		if runtime.GOOS == "windows" && c.OS == "linux" {
+			return osErr
+		} else if runtime.GOOS != "windows" && c.OS == "windows" {
+			return osErr
+		}
+	}
 	return nil
 }
 
--- a/types/types.go
+++ b/types/types.go
@@ -148,11 +148,11 @@ type ImageDestination interface {
 	SupportsSignatures() error
 	// ShouldCompressLayers returns true iff it is desirable to compress layer blobs written to this destination.
 	ShouldCompressLayers() bool
-
 	// AcceptsForeignLayerURLs returns false iff foreign layers in manifest should be actually
 	// uploaded to the image destination, true otherwise.
 	AcceptsForeignLayerURLs() bool
-
+	// MustMatchRuntimeOS returns true iff the destination can store only images targeted for the current runtime OS. False otherwise.
+	MustMatchRuntimeOS() bool
 	// PutBlob writes contents of stream and returns data representing the result (with all data filled in).
 	// inputInfo.Digest can be optionally provided if known; it is not mandatory for the implementation to verify it.
 	// inputInfo.Size is the expected length of stream, if known.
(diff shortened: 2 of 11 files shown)


runcom:
We need to re-vendor in skopeo after this is merged cause it's needed for atomic.

mtrmac:
I agree that failing fast in this situation is desirable, but this is overbroad. For example, `skopeo copy docker://$A docker://$B` can, and should, always work regardless of what `runtime.GOOS` is.

Some possible alternatives:
- Make “cares about architecture” (or, fully/too generally, “is a `types.Image` acceptable”?) a property of `types.ImageDestination`; that way `docker-daemon:` could declare itself to be architecture-specific, while allowing `docker:` and `dir:` to still accept anything.
  - (What about `containers-storage:`? Is it always architecture-specific like `docker-daemon`, or would it also be natural to use it for storing abitrary-arch images? E.g. with Dan’s network backend? Or is this another reason for the overbroad “is an image acceptable” callback, allowing `containers-storage:` to make a storage-backend-specific decision?)
- Punt, make the required arch, or `runtime.GOOS`-matching an opt-in option to `copy.Image` (and `skopeo copy`), and let the caller decide.

copy/copy.go
```diff
@@ -157,6 +158,17 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
...
+	c, err := src.OCIConfig()
+	if err != nil {
+		return err
+	}
+	osErr := fmt.Errorf("image operating system %q cannot be used on %q", c.OS, runtime.GOOS)
+	if runtime.GOOS == "windows" && c.OS == "linux" {
+		return osErr
+	} else if runtime.GOOS != "windows" && c.OS == "windows" {
+		return osErr
+	}
```
mtrmac: BTW please consider moving the logic out into a separate function to try and keep `copy.Image` from growing _too_ quickly, especially if it had to be much longer than this. (Absolutely non-blocking.)

runcom:
Agree on everything, I'll re-work the PR shortly

runcom:
I think the first possibility is the best here, let's not leave these decisions to users as they are pretty specific to each destination.

runcom:
@mtrmac PTAL - suggestions on naming welcome (I suck...)

types/types.go
```diff
@@ -153,6 +153,9 @@ type ImageDestination interface {
 	// uploaded to the image destination, true otherwise.
 	AcceptsForeignLayerURLs() bool
 
+	// AcceptsAnyOSSpecificImage returns true if the destination can store images for every operating system. False otherwise.
+	AcceptsAnyOSSpecificImage() bool
```
mtrmac: I don’t have _obviously_ better ideas. `AcceptsAnyOSSpecificImage` works fine with me.

`AcceptsImagesForAnyOS`? Still weird, and `Image` is kinda implied by the `ImageDestination`…

Or perhaps naming the unusual thing instead of the common case, `MustMatchRuntimeOS` or something like that?

mtrmac: … Actually explicitly saying “runtime OS” is somewhat nice, `AcceptsAny` says that everything is accepted, but when that is false, nothing says what is acceptable.  At least please say that in the comment which explains what `false` means.

types/types.go
```diff
@@ -153,6 +153,9 @@ type ImageDestination interface {
 	// uploaded to the image destination, true otherwise.
 	AcceptsForeignLayerURLs() bool
 
+	// AcceptsAnyOSSpecificImage returns true if the destination can store images for every operating system. False otherwise.
```
mtrmac: Nit, please remove the blank line above, so that blank lines separate functional groups of methods. (The line above `AcceptsForeighLayerURLs` might also be dropped.)

copy/copy.go
```diff
@@ -157,6 +158,19 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
...
+		if err != nil {
+			return err
+		}
+		osErr := fmt.Errorf("image operating system %q cannot be used on %q", c.OS, runtime.GOOS)
+		if runtime.GOOS == "windows" && c.OS == "linux" {
+			return osErr
+		} else if runtime.GOOS != "windows" && c.OS == "windows" {
+			return osErr
+		}
+	}
```
mtrmac: Yeah, please move this into a separate helper function to make `copy.Image` smaller.

copy/copy.go
```diff
@@ -157,6 +158,19 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 		}
 	}()
 
+	if !dest.AcceptsAnyOSSpecificImage() {
+		c, err := src.OCIConfig()
+		if err != nil {
+			return err
```
mtrmac: Consider 
```go
errors.Wrapf("Error parsing image configuration", err)
```
so that the caller doesn’t get a context-less `invalid character … looking for beginning of value` or whatever.

copy/copy.go
runcom: moved

runcom:
@mtrmac PTAL

runcom:
skopeo PR with test https://github.com/projectatomic/skopeo/pull/355

mtrmac:
:+1: Thanks!

runcom:
LGTM

PR #283, opened 2017-06-05
copy: copy layers concurrently

runcom:
@mtrmac @mrunalp @rhatdan PTAL (needed in CRI-O)

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -3,11 +3,13 @@ package copy
 import (
 	"bytes"
 	"compress/gzip"
+	"context"
 	"fmt"
 	"io"
 	"io/ioutil"
 	"reflect"
 	"strings"
+	"sync"
 	"time"
 
 	pb "gopkg.in/cheggaaa/pb.v1"
@@ -20,6 +22,7 @@ import (
 	"github.com/containers/image/types"
 	"github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
+	"golang.org/x/sync/errgroup"
 )
 
 type digestingReader struct {
@@ -33,7 +36,6 @@ type digestingReader struct {
 // data, that we're copying between images, and cache other information that
 // might allow us to take some shortcuts
 type imageCopier struct {
-	copiedBlobs       map[digest.Digest]digest.Digest
 	cachedDiffIDs     map[digest.Digest]digest.Digest
 	manifestUpdates   *types.ManifestUpdateOptions
 	dest              types.ImageDestination
@@ -44,6 +46,7 @@ type imageCopier struct {
 	reportWriter      io.Writer
 	progressInterval  time.Duration
 	progress          chan types.ProgressProperties
+	mu                sync.RWMutex
 }
 
 // newDigestingReader returns an io.Reader implementation with contents of source, which will eventually return a non-EOF error
@@ -196,7 +199,6 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 
 	// If src.UpdatedImageNeedsLayerDiffIDs(manifestUpdates) will be true, it needs to be true by the time we get here.
 	ic := imageCopier{
-		copiedBlobs:       make(map[digest.Digest]digest.Digest),
 		cachedDiffIDs:     make(map[digest.Digest]digest.Digest),
 		manifestUpdates:   &manifestUpdates,
 		dest:              dest,
@@ -298,9 +300,15 @@ func updateEmbeddedDockerReference(manifestUpdates *types.ManifestUpdateOptions,
 // copyLayers copies layers from src/rawSource to dest, using and updating ic.manifestUpdates if necessary and ic.canModifyManifest.
 func (ic *imageCopier) copyLayers() error {
 	srcInfos := ic.src.LayerInfos()
-	destInfos := []types.BlobInfo{}
-	diffIDs := []digest.Digest{}
-	for _, srcLayer := range srcInfos {
+	destInfos := make([]types.BlobInfo, len(srcInfos))
+	diffIDs := make([]digest.Digest, len(srcInfos))
+	// TODO: set ctx timeout somehow (should come from callers)
+	// CRI-O would call "image.Copy(ctx, ...) with ctx coming from the grpc
+	// caller for instance.
+	ctx := context.Background()
+	eg, ctx := errgroup.WithContext(ctx)
+	for i, srcLayer := range srcInfos {
+		i, srcLayer := i, srcLayer // https://golang.org/doc/faq#closures_and_goroutines
 		var (
 			destInfo types.BlobInfo
 			diffID   digest.Digest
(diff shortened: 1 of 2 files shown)


runcom:
skopeo PR projectatomic/skopeo#354

copy/copy.go
```diff
@@ -313,16 +321,22 @@ func (ic *imageCopier) copyLayers() error {
...
-			if err != nil {
-				return err
-			}
+			eg.Go(func() error {
+				destInfo, diffID, err = ic.copyLayer(srcLayer)
+				if err != nil {
+					return err
+				}
+				destInfos[i] = destInfo
+				diffIDs[i] = diffID
```
runcom: actually, `diffIDs` can contain less than `len(srcInfos)` - I'll shift it out of this loop to make sure it doesn't have holes (even if it's another loop, nothing which should impact performance though...I guess).

mtrmac:
I haven’t yet reviewed this in detail, for now two points:

- Some transports can’t “just” deal with two `PutBlob` calls in parallel. This is at least `docker-daemon:`, `docker-archive:`, and especially `containers-storage:`, which may be able to accept layers in parallel (I don’t know), but is specifically designed to require `PutBlob` calls to be made in order from bottom layer to top. So this may require a more involved rearchitecting (initiate upload / cancel upload / apply finished upload as a layer in order, or something like that). @nalind?
- Won‘t this horribly mangle the progress bars?

runcom:
> I haven’t yet reviewed this in detail, for now two points:
> 
> Some transports can’t “just” deal with two PutBlob calls in parallel. This is at least docker-daemon:, docker-archive:, and especially containers-storage:, which may be able to accept layers in parallel (I don’t know), but is specifically designed to require PutBlob calls to be made in order from bottom layer to top. So this may require a more involved rearchitecting (initiate upload / cancel upload / apply finished upload as a layer in order, or something like that). @nalind?

Indeed, @mrunalp sent out an email just yesterday about any c/storage work needed to support this (I've CC'ed you on it @mtrmac)

> Won‘t this horribly mangle the progress bars?

This is needed by CRI-O to match docker/containerd pull time. I don't plan to deal with this in skopeo soon (even if I opened the respective PR). We can leave the non-parallel copy to not break the skopeo progress bar and have an option to chose between parallel-series (CRI-O will use parallel).

mtrmac:
> This is needed by CRI-O to match docker/containerd pull time.

BTW, a very stupid question, why does parallel pull make anything faster? I’d naively expect the total time to be dominated by network and/or storage speed, where running more threads does not inherently speed up usage of either (the OS should already be queuing operations to achieve 100% utilization), but it should add overhead (more network streams competing for bandwidth = more overhead in resends andd dropped packets, more seeking / contention on hard drives).

Are we limited by the CPU time needed to decompress the layers? Are we unable to saturate Ethernet with a single CPU core, or something? Or am I completely misunderstanding the situation?

> I don't plan to deal with this in skopeo soon (even if I opened the respective PR). We can leave the non-parallel copy to not break the skopeo progress bar and have an option to chose between parallel-series (CRI-O will use parallel).

I don’t really like this; if we want to support parallel copies, let’s actually get it done fully. `copy.Image` is pretty complex and keeping two different paths in there, each suboptimal, for the user to choose their own poison, is a fair amount of technical debt and carrying that forward is pretty unappealing to me.  (I guess I could live with an implementation where the `eg.Go()` call was only somehow parametrized to only run one operation at a time, with all other code unconditionally parallel, i.e. not two separate paths.)

runcom:
> I don’t really like this; if we want to support parallel copies, let’s actually get it done fully. copy.Image is pretty complex and keeping two different paths in there, each suboptimal, for the user to choose their own poison, is a fair amount of technical debt and carrying that forward is pretty unappealing to me. (I guess I could live with an implementation where the eg.Go() call was only somehow parametrized to only run one operation at a time, with all other code unconditionally parallel, i.e. not two separate paths.)

I know, I'm thinking about 2 separate code paths just because it's the fastest alternative I have wrt the cycles I can spend on this for CRI-O - syncing the progress bar for parallel operation isn't going to be straightforward. I'll try to fix that though.

copy/copy.go
```diff
@@ -459,7 +478,9 @@ func (ic *imageCopier) copyLayer(srcInfo types.BlobInfo) (types.BlobInfo, digest
 			return types.BlobInfo{}, "", errors.Wrap(diffIDResult.err, "Error computing layer DiffID")
```
mtrmac: Above this: `GetBlob`, `srcStream.Close()` needs auditing.

copy/copy.go
```diff
@@ -459,7 +478,9 @@ func (ic *imageCopier) copyLayer(srcInfo types.BlobInfo) (types.BlobInfo, digest
...
 		}
 		logrus.Debugf("Computed DiffID %s for layer %s", diffIDResult.digest, srcInfo.Digest)
+		ic.mu.Lock()
 		ic.cachedDiffIDs[srcInfo.Digest] = diffIDResult.digest
+		ic.mu.Unlock()
```
mtrmac: Below this:
- The progress bar needs handling somehow.
- `ShouldCompressLayers()` needs auditing, or I guess preferably calling only once outside of the parallel code.
- `PutBlob` needs auditing.

copy/copy.go
```diff
@@ -44,6 +46,7 @@ type imageCopier struct {
 	reportWriter      io.Writer
 	progressInterval  time.Duration
 	progress          chan types.ProgressProperties
+	mu                sync.RWMutex
```
mtrmac: It would be nice to explicitly document which members are expected to be read-only throughout, and which members are protected by the mutex.  (Or, for now, we could just use an `x/sync/syncmap` for `cachedDiffIDs` if that’s the only protected member—but I guess that won’t last.)

copy/copy.go
runcom: nobody is going to update those at this particular index in time, errgroup pkg does have an example exactly doing this also.

rhatdan:
@jeremyeder Please could you add your thoughts?

copy/copy.go
mtrmac: Fair enough I guess.

mtrmac:
> syncing the progress bar for parallel operation isn't going to be straightforward. 

BTW I would be perfectly fine with being a bit less decorative with the progress reporting, and instead of updating several progress bars on a screen just having a single summary status line, something like
> 5/$N, 1.3 GB/$M copied
where $N and $M grow over time as more layers are opened.

mtrmac:
> the bulk of the outstanding work is in those “audit” comments

The PR as written is essentially:
```go
for i, layer := range layers {
  go func() {
    layer.SourceOpen()
    layer.DestOpen()
    io.Copy(layer.Dest, layer.Source)
    layer.SourceClose()
    layer.DestClose()
  }
}
wait()
```
i.e. _everything_ in there needs to be thread-safe. This is, in some ways, easy to understand, and it is nice in that if a source or destination cannot support parallelism, it can just block in `SourceOpen`/`DestOpen` to admit only one oepration at a time.

Still, perhaps we should consider an alternative:
```go
for i, layer := range layers {
  layer.SourceOpen()
  layer.DestOpen()
}
for i, layer := range layers {
   go io.Copy(layer.Dest, layer.Source)
}
wait()
for i, layer := range layers {
  layer.SourceClose()
  layer.DestClose()
}
```

i.e. require _only_ the `io.Reader` / `io.Writer` implementations to be thread-safe (which is trivial for all of sockets, ordinary files, temporary files), and allow the complex `*Open()/*Close()` operations to be run safely in a single thread. The downsides are:
- we would have to split the `GetBlob` /`PutBlob` into `Open`/copy/`Close` sequences, with something to carry context.
- serializing the copies, for transports that care, would have to be made more explicit than an internal mutex.
- copy.Image becomes more complex again
- we might want to make `dest.Close` run in parallel as well?

but given the amount of non-trivial code which would not have to be thread-safe, it might be well worth it. I don’t know.

rhatdan:
Are we going to come back to this, or drop the idea?

runcom:
we have to still work on a reliable way to compare download time, so far I guess nothing has changed and moreover we first need to implement locking in c/image and c/storage if I remember correctly.

mtrmac:
This PR is insufficient, but the idea is wothwhile and keeping this PR open works just as well as anything.

(The benchmarking is not strictly required, but I’d guess that benchmarking to figure out the bottlenecks and targeting only those would be much simpler than making all of c/image and c/storage multithread-safe.)

PR #285, opened 2017-06-16
Helper interface to generate a comma separate list of transports.

rhatdan:
This interface can be used by callers in their Usage statements, to
guide users to all potential transports.

### DIFF
--- a/transports/transports.go
+++ b/transports/transports.go
@@ -69,3 +69,16 @@ func Register(t types.ImageTransport) {
 func ImageName(ref types.ImageReference) string {
 	return ref.Transport().Name() + ":" + ref.StringWithinTransport()
 }
+
+// List returns a comma separate list of transports
+func ListNames() string {
+	list := ""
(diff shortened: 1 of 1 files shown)


transports/transports.go
```diff
@@ -69,3 +69,16 @@ func Register(t types.ImageTransport) {
 func ImageName(ref types.ImageReference) string {
 	return ref.Transport().Name() + ":" + ref.StringWithinTransport()
 }
+
+// List returns a comma separate list of transports
+func ListNames() string {
```
mtrmac: * While every other access is protected by the `kt.mu` mutex, this should probably be as well
* Return an array of strings, and let the caller join with `strings.Join(theList, ", ")`? That seems a bit cleaner than returning a single string and possibly forcing callers to split it and recombine.

rhatdan:
Git hub does not seem to be working correctly right now, push --force is not  updating the pull request.  I will close this and reopen with an updated pull request

PR #286, opened 2017-06-16
Interface to return list of transports names

rhatdan:
This interface can be used by callers in their Usage statements, to
guide users to all potential transports.

### DIFF
--- a/transports/transports.go
+++ b/transports/transports.go
@@ -2,6 +2,7 @@ package transports
 
 import (
 	"fmt"
+	"sort"
 	"sync"
 
 	"github.com/containers/image/types"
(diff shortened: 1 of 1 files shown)


transports/transports.go
```diff
@@ -69,3 +70,13 @@ func Register(t types.ImageTransport) {
 func ImageName(ref types.ImageReference) string {
 	return ref.Transport().Name() + ":" + ref.StringWithinTransport()
 }
+
+// List returns a comma separate list of transports
```
runcom: `s/List/ListNames/g`

transports/transports.go
```diff
@@ -69,3 +70,13 @@ func Register(t types.ImageTransport) {
...
+func ListNames() []string {
+	var names []string
+	for _, transport := range kt.transports {
```
runcom: we probably need to hold the lock when iterating here as @mtrmac pointed out in the other PR

transports/transports.go
rhatdan: Yes I just cought that and repushed, but I am not seeing updates.

runcom: weird :confused: github might have some issues

PR #288, opened 2017-06-16
RFC Do not merge: Dockerd verification support

mtrmac:
@runcom PTAL. This is an infrastructure for allowing verification of images pulled into dockerd, to allow enforcing policy at container start time or layer verification. Do not merge yet until this is proven by having a working projectatomic/docker patch.

Overall outline:
* Add `PolicyContext.IsImageAuthenticated`, distinct from `IsRunningImageAllowed` in that it is allowed to run an `insecureAcceptAnything` image, but it is not authenticated. This can be used to answer “do we know that the layers come from a known source?”; but of course layer verification of unauthenticated images (e.g. to guard against filesystem corruption) could still be done.
* Add an API to get layer DiffIDs from an image.
* Add `ImageSource.GetOriginalManifest` / `image.UnparsedOriginalFromSource`, useless for accessing the complete image, but it can be authenticated by an `PolicyContext` and used to get authenticated DiffIDs
* Add `docker/daemon/extra`, a lookaside mechanism to store signatures and the original manifest for `docker-daemon:`
* Add `docker/daemon/extra/Store.RecordImage`, intended as a single call to be added to the right place within dockerd.

Areas I am especially seeking comments about:
* The lookup design. Is associating to images by config digest (= image ID) + reference reasonable?
* The storage design. This uses a set of plain files + symlinks, which is easy to debug/inspect but somewhat costly in internal fragmentation. I was also considering sqlite, surprisingly golang does not have native/default sqlite bindings?! What is the best practice within dockerd / the rest of the infrastructure?
* Naming. I know the current naming sucks, naming the whole subpackage `docker/daemon/extra` is pretty much “I give up".

Any other comments are of course welcome as well.


mtrmac:
Example of a possible caller:
```go
		rawSource, err := ref.NewImageSource(nil, nil)
		if err != nil {
			logrus.Fatalf("Error initializing source %s: %v", transports.ImageName(ref), err)
		}
		unparsedOriginal := image.UnparsedOriginalFromSource(rawSource)
		defer /* close unparsedOriginal */

		// Please keep this policy check BEFORE reading any other information about the image.
		if authenticated, err := policyContext.IsImageAuthenticated(unparsedOriginal); !authenticated || err != nil { // Be paranoid and fail if either return value indicates so.
			logrus.Fatalf("Source image not authenticated: %v", err)
		}
		originalImg, err := image.FromUnparsedImage(unparsedOriginal)
		if err != nil {
			logrus.Fatalf("Error initializing image from source %s: %v", transports.ImageName(ref), err)
		}
		unparsedOriginal = nil
		defer /* close originalImg */

		diffIDs, err := originalImg.LayerDiffIDs()
		if err != nil {
			logrus.Fatalf("Error getting layer DiffIDs: %v", err)
		}
		if diffIDs == nil {
			logrus.Fatal("DiffIDs are not available")
		}
		// verify extracted layers against diffIDs
```

runcom:
Haven't yet gone through code, but, is this actually meant to be used with something like `atomic` and be done with it? I was wondering about something like this vendored in projectatomic/docker, but I can't see how that's possible with this (am I missing something?)

mtrmac:
This would be used:

- Immediately in `docker-daemon:`, as this PR does, i.e. (atomic pull)
- In projectatomic/docker, after https://github.com/projectatomic/docker/commit/add9c14e6ab098404d8ee0566d322676e5ae1799#diff-68f746e0175a060913947b89474cd7aeR68 suceeds and the image is succesfully pulled, by calling `extra.Store.RecordImage`. This may be a bit tricky, hence the "do not merge” until I can prototype that
- In projectatomic/docker for run enforcement. Given an image ID and a name, that should amount to ~ the code quoted above.
- In whatever wants to verify layers. The authenticated DiffIDs + layer reconstruction for `docker save` allow verifying layer tarballs, then there’s the tricky and not-quite-trivial matter of ensuring that the filesystem is exactly the same as the tarball. That may be a horrible patch to projectatomic/docker.

mtrmac:
Is there something I have overlooked which would make this impossible to vendor into projectatomic/docker?

runcom:
It seems not :smile: except the potential Docker patch you mentioned for the last point which I'd hate to carry but there's no other way around that so, sorry for the rant on the patch :P

mtrmac:
Yeah, that last part does scare me. But, in the worst case, we can decide not to do that part :) and the verification-on-run would still be useful

mtrmac:
Also the bulk of that last patch might be upstreamable.

types/types.go
```diff
@@ -226,6 +226,10 @@ type Image interface {
 	// The Digest field is guaranteed to be provided; Size may be -1.
 	// WARNING: The list may contain duplicates, and they are semantically relevant.
 	LayerInfos() []BlobInfo
+	// LayerDiffIDs returns a list of DiffIDs (= digests of the UNCOMPRESSED versions) of layers referenced by this image,
+	// in order (the root layer first, and then successive layered layers), if available, or nil if not.
+	// WARNING: The list may contain duplicates, and they are semantically relevant.
+	LayerDiffIDs() ([]digest.Digest, error)
```
runcom: what's the difference with `LayerInfos()` above? can we make this clear? e.g. maybe `LayerInfos` can contain COMPRESSED digest?

docker/daemon/extra/store.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/containers/image/docker/reference"
+	"github.com/containers/image/manifest"
+	"github.com/containers/image/types"
+	"github.com/opencontainers/go-digest"
+)
+
+// systemDockerExtraDirPath is the path to the docker extra storage, used for recording metadata allowing to authenticate docker images.
+// You can override this at build time with
+// -ldflags '-X github.com/containers/image/docker/daemon/extra.systemDockerExtraDirPath=$your_path'
+var systemDockerExtraDirPath = builtinDockerExtraDirPath
```
runcom: I still believe compile-time changes aren't good for this kind of stuff. Ideally, go codes have a way to pass a `Ctx` struct down the function calls, that `Ctx` should contain this information.

types/types.go
mtrmac: `LayerInfos` contains the digest of the downloadable blob, whether or not it is compressed. Anyone wanting to call `ImageSource.GetBlob` needs the `LayerInfos` output.

`LayerDiffIDs` contains the DiffID, which may be useless for getting the layer (i.e. the caller may need to call `LayerInfos`+`GetBlob` and uncompress, or get the data from elsewhere).

I have updated the comments a _tiny_ bit, it could probably still be more explicit. Perhaps the `LayerDiffIDs` documentation should say that “most users should call `LayerInfos` instead”?

docker/daemon/extra/store.go
mtrmac: That’s already possible as well (`SystemContext.DockerExtraDirPath`). This build-time-override has been useful for development, but I’m quite fine with dropping it.

runcom:
Looked at this briefly, I'll comment of what you need for now.

> Areas I am especially seeking comments about:
> 
> The lookup design. Is associating to images by config digest (= image ID) + reference reasonable?

Seems resonable to me. It's what docker does unless I'm terribly wrong.

> The storage design. This uses a set of plain files + symlinks, which is easy to debug/inspect but somewhat costly in internal fragmentation. I was also considering sqlite, surprisingly golang does not have native/default sqlite bindings?! What is the best practice within dockerd / the rest of the infrastructure?

So in the golang ecosystem, the best way to handle such a storage is `bolddb` - take a look at that as it's surely what we want to adopt for this.

> Naming. I know the current naming sucks, naming the whole subpackage docker/daemon/extra is pretty much “I give up".

Top level could be `docker/daemon/signature` and for the store `docker/daemon/signature/store`, seems pretty clear to me.

mtrmac:
> So in the golang ecosystem, the best way to handle such a storage is `bolddb` - 

Thanks; do you perhaps mean `boltdb/bolt`,  or `coreos/bbolt`?

runcom:
> Thanks; do you perhaps mean boltdb/bolt, or coreos/bbolt?

`github.com/boltdb/bolt`

mtrmac:
@runcom PTAL again.  Rewritten to use `boltdb`, and renamed from `c/i/docker/daemon/extra` to `…./signatures`.

https://github.com/projectatomic/skopeo/pull/462 shows passing tests.

(Still “do not merge” because the whole `OriginalManifest` vs. `UpdatedLayerInfos` design decision is unclear to me yet. More eyes on that would be welcome as well.)

rhatdan:
@runcom @mtrmac Are we moving forward on this patch?

mtrmac:
> @runcom @mtrmac Are we moving forward on this patch?

I still need to reconcile the `OriginalManifest` vs. `UpdatedLayerInfos` implementation, and probably decide on the policy format for handling this dichotomy. It’s not abandoned, at the very least.

rhatdan:
Any update on this?

mtrmac:
Still about the same as last time.

rhatdan:
@mtrmac @vrothberg  What should we do with this PR? Should we rebase and move forward or should we close it?

mtrmac:
I don’t think we care that much about teaching dockerd to record original manifests any more, but at least `PolicyContext.IsImageAuthenticated` and `types.Image.LayerDiffIDs` from this PR still seem relevant for verifying images in local c/storage.

mtrmac:
I guess we don’t care about adding run-time signature verification to `docker run` any more.  I’ll keep some of the infrastructure around for future use.

PR #292, opened 2017-06-19
make sure to check .crt suffix for certificate if key is found

### LINKED ISSUE #291: dockerClient fails if registry key present and registry cert ends with .crt
If registry key is found in the DockerCertPath e.g. 

```
/etc/secrets/
├── registry.crt
└── registry.key 
```

then `newDockerClient()` fails with error `setting up certs: missing client certificate registry.cert for key registry.key`

However if we rename registry.crt to registry.cert the error does not occur.

I can submit a quick PR to fix this issue (the problem seems to be [here](https://github.com/containers/image/blob/master/docker/docker_client.go#L190))

ilackarms:
this fixes #291

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -188,9 +188,10 @@ func setupCertificates(dir string, tlsc *tls.Config) error {
 		if strings.HasSuffix(f.Name(), ".key") {
 			keyName := f.Name()
 			certName := keyName[:len(keyName)-4] + ".cert"
+			crtName := keyName[:len(keyName)-4] + ".crt"
 			logrus.Debugf(" key: %s", fullPath)
-			if !hasFile(fs, certName) {
-				return errors.Errorf("missing client certificate %s for key %s", certName, keyName)
+			if !hasFile(fs, certName) && !hasFile(fs, crtName) {
+				return errors.Errorf("missing client certificate %s|%s for key %s", certName, crtName, keyName)
 			}
 		}
 	}


ilackarms:
@enoodle @simon3z this change is necessary (in lieu of a workaround) for https://github.com/openshift/image-inspector/pull/37

mtrmac:
This seems to be behaved as documented in https://docs.docker.com/engine/security/certificates/ , and in `types/types.go:: `*.crt` are CA roots, `*.cert` are certificates.

Why is it expected that `*.crt` should be accepted as certificates as well?

Given the pre-existing treatment of `*.crt` as CAs, treating an ordinary certificate as a CA root seems pretty risky.

Also, this patch would allow `setupCertificates` to not fail, but not to _work_: The key pair is actually being loaded in the `strings.HasSuffix(…, ".cert")` case, and this patch does not touch it.

ilackarms:
then the actual problem is that if the .key file is found before the .crt file (and no .cert file exists), the function will fail

mtrmac:
The `.crt` files (CA certificates) are quite unrelated to the `.cert`/.`key` public key/private key file pairs. Existence of `someca.crt` should AFAICT make no difference WRT existence of `keypair.cert` - whether or not they use the same file name prefix.

If the configuration contains only the private part of the public/private key pair, it seems quite reasonable to me that the function should fail—there is no way to use that private key in TLS without providing the public counterpart in a certificate trusted by the other party.

Could you describe precisely what files you have, and what is the intended contents/semantics of those files (WITHOUT revealing your actual private keys), please? The `registry.crt`/`registry.key` pair mentioned in https://github.com/containers/image/issues/291 just doesn’t make sense to me.

mtrmac:
> then the actual problem is that if the .key file is found before the .crt file (and no .cert file exists), the function will fail

This answer still does not account for anything that would make the files _work_: The only thing that reads `.key` files is the `.cert` case, so just making the function not fail does nothing to interpret the contents of the `.key` files; the `.key` file could AFAICT be simply removed and the effect would be exactly the same as applying this patch.

ilackarms:
the secrets used to create a private docker registry by openshift's `oadm` tool are named tls.crt and tls.key respectively.

only the tls.crt is needed to validate an https connection to the registry. the error I encountered was caused when using the registry secrets directory as the `DockerCertPath`. What wound up happening is that setupCertificates found the .key file first, looked for a .cert file, didn't find one and errored out. however if I deleted the .key file setupCerts did not error and pulling image succeeded

ilackarms:
this issue can also be resolved by making sure we first check for a .crt file before looking for the .cert / .key pair

mtrmac:
What “secrets”? Please be precise about their contents and semantics.

The `.key` file is the only one supposed to contain a private key, i.e. something truly secret. If you removed them and things still work, you didn’t need any “secrets”.

mtrmac:
> the secrets used to create a private docker registry by openshift's oadm tool are named tls.crt and tls.key respectively.

Do you happen to mean the files which are submitted as arguments to `oadm registry --tls-certificate=… --tls-key=…`? Thouse are secrets _of the registry_ and should never be published to clients.

ilackarms:
yes, that's what I mean. the tls certificate is necessary to validate over https

ilackarms:
the issue is simply that having a valid .crt file and **any** .key file located in the same directory can cause setupCertificates to (improperly) fail. I'm closing this PR as it does not address the issue

mtrmac:
Then
* **Urgently** change your workflow to stop distributing the `.key` file. Any client having a copy of the `.key` file can impersonate your registry. If any pre-existing copies are not under your control any more, revoke the original certificate (and, if possible, distribute this revocation to the clients so that they respect it… not sure that this is easy to do), generate a new key pair, and reconfigure your registry to use the new pair.
* After removing the `.key` file copies, things may just work, but really the right thing to do is to distribute the CA certificate used to sign the registry certificate as `*.crt`, not the exact server certificate; distributing the server certificate as the root of trust prevents the CA from reissuing certificates (e.g. after the revocation recommended above, or after the original certificate expires and must be reissued).

mtrmac:
> the issue is simply that having a valid .crt file and any .key file located in the same directory can cause setupCertificates to (improperly) fail.

The `.crt` file is completely irrelevant in this situation. The function does not fail because of the `.crt` file, it fails because it has been supplied a `.key` file without a `.cert` file, and that is by design; the two must be present together. And at least in this case it seems to have been useful to detect incorrect use of the files.

mtrmac:
> really the right thing to do is to distribute the CA certificate used to sign the registry certificate as `*.crt` not the exact server certificate;

In terms of the OpenShift-integrated CA, this is probably `openshift.local.config/master/ca.crt`.

ilackarms:
agreed, thank you @mtrmac

PR #298, opened 2017-06-29
Fix checks for storage.Errors and tests for them

nalind:
The storage library, and parts of our storage-related logic, can return errors wrapped by `github.com/pkg/errors.Wrap()` and friends, so when checking for equality to specific errors, use `errors.Cause()` to dig into them.

Tests for duplicate ID detection were broken starting in d41025d60560cf463a215fc951a6ed949523a4bf, which made it acceptable to reuse an image ID if the new image and the old image had the same top layer, so adjust the tests for that error to use different layer contents when creating the second image with the same ID as one that it had just created.


mtrmac:
:+1:

nalind:
Rebased.

TomSweeneyRedHat:
LGTM.

nalind:
Rebased.  @runcom, can you have a look?

runcom:
LGTM

PR #300, opened 2017-07-06
Expand store locations to include runRoot and opts

nalind:
Expand the syntax of the store's location to allow it to also contain the runRoot location (optional, after the root, separated by '+') and a list of driver options (optional, after the runRoot, separated by ":",
and itself broken up using ",").

### DIFF
--- a/storage/storage_test.go
+++ b/storage/storage_test.go
@@ -58,7 +58,7 @@ func TestMain(m *testing.M) {
 	os.Exit(code)
 }
 
-func newStore(t *testing.T) storage.Store {
+func newStoreWithGraphDriverOptions(t *testing.T, options []string) storage.Store {
 	wd, err := ioutil.TempDir(topwd, "test.")
 	if err != nil {
 		t.Fatal(err)
@@ -83,7 +83,7 @@ func newStore(t *testing.T) storage.Store {
 		RunRoot:            run,
 		GraphRoot:          root,
 		GraphDriverName:    "vfs",
-		GraphDriverOptions: []string{},
+		GraphDriverOptions: options,
 		UIDMap:             uidmap,
 		GIDMap:             gidmap,
 	})
@@ -92,6 +92,10 @@ func newStore(t *testing.T) storage.Store {
 	}
 	Transport.SetStore(store)
 	return store
+}
+
+func newStore(t *testing.T) storage.Store {
+	return newStoreWithGraphDriverOptions(t, []string{})
 }
 
 func TestParse(t *testing.T) {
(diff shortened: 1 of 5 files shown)


storage/storage_transport.go
```diff
@@ -238,7 +266,9 @@ func (s storageTransport) ValidatePolicyConfigurationScope(scope string) error {
 	}
 	storeSpec := scope[1:closeIndex]
 	scope = scope[closeIndex+1:]
-	storeInfo := strings.SplitN(storeSpec, "@", 2)
+	optionInfo := strings.SplitN(storeSpec, ":", 2)
+	runRootInfo := strings.SplitN(optionInfo[0], "+", 2)
+	storeInfo := strings.SplitN(runRootInfo[0], "@", 2)
```
mtrmac: From an one-minute look—if `PolicyConfiguration{Identity,Namespaces}` is not changing its format, should `ValidatePolicyConfigurationScope` change?

nalind: No, it shouldn't have had to.  This is left over from an earlier version of the patch where they did.  I'll undo this part.

nalind:
Updated, adding a way to specify UID and GID maps to be used for all stores, and using them in the unit tests to avoid permissions problems while parsing references.

storage/storage_transport.go
```diff
@@ -165,37 +200,66 @@ func (s *storageTransport) ParseReference(reference string) (types.ImageReferenc
...
+			GraphDriverOptions: options,
+			UIDMap:             s.defaultUIDMap,
+			GIDMap:             s.defaultGIDMap,
+		})
+		if err != nil {
+			return nil, err
+		}
+		store = store2
+	}
+	if store == nil {
```
mtrmac: Nit, can’t this be simply `} else {`? Or is it possible for `store2, err` above to return `nil, nil`, _and_ would we want to silently use the default if it did?

storage/storage_test.go
```diff
@@ -68,23 +68,23 @@ func newStore(t *testing.T) storage.Store {
...
 		ContainerID: 0,
 		HostID:      os.Getgid(),
 		Size:        1,
-	}}
+	}})
 	store, err := storage.GetStore(storage.StoreOptions{
 		RunRoot:            run,
 		GraphRoot:          root,
 		GraphDriverName:    "vfs",
 		GraphDriverOptions: []string{},
```
mtrmac: Non-blocking: Would it be easy, and safe, to add some values to `GraphDriverOptions`, to test the option parsing and formatting? AFAICS `c/storage/drivers/vfs/Init` ignores `options` entirely, so adding say `{"unused1", "unused2"}` would work for now—OTOH it might break anytime if `vfs` starts to parse the options.

storage/storage_transport_test.go
```diff
@@ -68,24 +68,24 @@ func TestTransportParseReference(t *testing.T) {
...
+		{"[thisisunknown@" + root + "suffix2]", "", "", ""}, // Unknown graph driver
+		{"[" + root + "suffix1]", "", root + "suffix1", ""}, // A valid root path, but no run dir
+		{"[" + driver + "@" + root + "suffix3+" + root + "suffix4]",
+			driver,
+			root + "suffix3",
+			root + "suffix4"}, // A valid root@graph+run set
+		{"[" + driver + "@" + root + "suffix3+" + root + "suffix4:options,options,options]",
+			driver,
+			root + "suffix3",
+			root + "suffix4"}, // A valid root@graph+run+options set
```
mtrmac: Tests for the working case. Nice!

mtrmac:
:+1:

runcom:
LGTM

storage/storage_transport.go
nalind: I don't think that it makes sense for `GetStore()` to ever return a `nil` for both of its returned values, so yeah, it could be changed to `} else {`.  I'll change that.

storage/storage_test.go
nalind: I think the only risk there is for consumers of both libraries that don't use the version of `storage` that we specify in the `vendor.conf` in this package.  The CI system should help catch that sort of change when we update the version listed there, so I think it's worth adding that.

nalind:
Yea, I think it's a bit much to expect every call to `skopeo copy` to include all of the settings.  The storage library recently gained a configuration file that can be used to set a shared default for everything that uses it, and I expect that to cover most uses.
This is geared more toward making it possible to use additional locations, should you find you need to use them.  One possible such case would be populating overlay-style local storage that is then shared using a networked filesystem to clients that use it as a read-only store for the overlay driver.

storage/storage_test.go
mtrmac: Consumers of libraries usually (or at least the ones I know about) don’t vendor in tests, so this should not affect them at all.

nalind:
Okay, amended to change that line to a `} else {`, and to add a test that ensures that an options list survives the round trip through `ImageReference.StringWithinTransport()` and `ImageTransport.ParseReference()`.

nalind:
Rebased.

nalind:
Any chance we can merge this, assuming it passes CI?

runcom:
yep, I'll merge after CI

runcom:
needs last rebase @nalind :(

nalind:
@runcom Okay, rebased.

PR #301, opened 2017-07-06
Implement policy configuration scopes in docker-daemon:

mtrmac:
This is necessary for verification / policy enforcement of in-daemon images. (#288)

It is awkward in that an image may have no name and be usable only by an ID; such IDs could have a policy identity but they can’t really be namespaced.

For now, implement policy configuration scopes only for named image references; ID-only references use the root scope.  This allows users to make ID-only images untrusted or to reject them, forcing users to use image names if they want the policy to approve.  This gives a fairly natural semantics, equivalent to `docker:` policies.  ID-like policy configuration scopes are forbidden.  If truly necessary, we can add support for single-ID policy IDs into the policy in the future, but that’s unlikely to be too useful—IDs change over time, so such a policy would not be likely to be persistent; and a single-use policy built for a single image can just as well use the universal `""` scope.

(Note that `docker-daemon:` and `docker:` namespaces are still separate in the policy—images from any source can be loaded into the daemon via `(docker load)` / copy to `docker-daemon:`, so there is in general no expectation that the two namespaces are equal.  Though a special-purpose tool may well want to create an in-memory policy by loading the system-wide one and grafting the `docker:`
`PolicyTransportScopes` map to `docker-daemon:` as well.)

[In many cases an image will have an unique `RepoTags`/`RepoDigests` value which can be used to get _a_ name from the ID, probably the right guess. For now we keep the `ImageReference` implementation dumb, though inspecting the image by ID and giving the reference a tagged/digested
name is at least plausible in principle.  Special-purpose tools can script this without having to wait for `daemonReference` to add this.]

### DIFF
--- a/docker/daemon/daemon_transport.go
+++ b/docker/daemon/daemon_transport.go
@@ -2,13 +2,15 @@ package daemon
 
 import (
 	"context"
-	"github.com/pkg/errors"
+	"fmt"
 
+	"github.com/containers/image/docker/policyconfiguration"
 	"github.com/containers/image/docker/reference"
 	"github.com/containers/image/image"
 	"github.com/containers/image/transports"
 	"github.com/containers/image/types"
 	"github.com/opencontainers/go-digest"
+	"github.com/pkg/errors"
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


mtrmac:
@runcom PTAL

rhatdan:
@runcom Could you take a look at this.

rhatdan:
@umohnani8 PTAL

runcom:
LGTM

TomSweeneyRedHat:
LGTM

umohnani8:
LGTM, needs rebase though.

mtrmac:
👍

PR #302, opened 2017-07-07
Add "Annotations" to OCI manifest and descriptors

nalind:
Add the "Annotations" field to our definition for the OCI manifest type, and give OCI its own descriptor type, so that UpdatedImage() doesn't lose annotations unless it's converting between formats.  Handle
conversions between OCI descriptors and traditional ones.

### DIFF
--- a/image/oci.go
+++ b/image/oci.go
@@ -12,12 +12,18 @@ import (
 	"github.com/pkg/errors"
 )
 
+type descriptorOCI1 struct {
+	descriptor
+	Annotations map[string]string `json:"annotations,omitempty"`
+}
+
 type manifestOCI1 struct {
 	src               types.ImageSource // May be nil if configBlob is not nil
 	configBlob        []byte            // If set, corresponds to contents of ConfigDescriptor.
 	SchemaVersion     int               `json:"schemaVersion"`
-	ConfigDescriptor  descriptor        `json:"config"`
-	LayersDescriptors []descriptor      `json:"layers"`
+	ConfigDescriptor  descriptorOCI1    `json:"config"`
+	LayersDescriptors []descriptorOCI1  `json:"layers"`
+	Annotations       map[string]string `json:"annotations,omitempty"`
 }
 
 func manifestOCI1FromManifest(src types.ImageSource, manifest []byte) (genericManifest, error) {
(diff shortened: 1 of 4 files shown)


image/oci.go
```diff
@@ -148,7 +154,7 @@ func (m *manifestOCI1) UpdatedImage(options types.ManifestUpdateOptions) (types.
 		if len(copy.LayersDescriptors) != len(options.LayerInfos) {
 			return nil, errors.Errorf("Error preparing updated manifest: layer count changed from %d to %d", len(copy.LayersDescriptors), len(options.LayerInfos))
 		}
-		copy.LayersDescriptors = make([]descriptor, len(options.LayerInfos))
+		copy.LayersDescriptors = make([]descriptorOCI1, len(options.LayerInfos))
 		for i, info := range options.LayerInfos {
 			copy.LayersDescriptors[i].Digest = info.Digest
 			copy.LayersDescriptors[i].Size = info.Size
```
nalind: Are we supposed to be copying any other fields here?

runcom:
@mtrmac ptal

image/oci.go
```diff
@@ -12,12 +12,18 @@ import (
 	"github.com/pkg/errors"
 )
 
+type descriptorOCI1 struct {
```
runcom: Wondering why aren't we using the upstream types

nalind:
We could alternatively slip `Annotations` into the regular `descriptor` type, which would probably make the patch smaller.  I don't have a preference.

image/oci.go
nalind: My guess is that, at the time, the spec was moving fast and breaking things?  I wouldn't object to switching to using them directly.

mtrmac: Historically probably because `image/oci.go` was created by copy&pasting `image/docker_schema2.go`, so it ended up using our `descriptor` copy.

I would be just fine with switching to upstream types as well, but it is not a blocker for this PR for me.

image/oci.go
mtrmac: Actually, yes, the definition of `LayerInfos` says that `URLs` should also be copied. But that’s unrelated to this PR, so I’m fine with both keeping this as is and adding the `URLs` copy here.

(That seems to have the effect of mostly dropping the URLs field whenever we upload a blob, even if it would make sense to keep the original URLs. But let’s worry about that in some other PR or issue.)

image/oci_test.go
```diff
@@ -29,36 +29,36 @@ func manifestOCI1FromFixture(t *testing.T, src types.ImageSource, fixture string
 }
 
 func manifestOCI1FromComponentsLikeFixture(configBlob []byte) genericManifest {
-	return manifestOCI1FromComponents(descriptor{
+	return manifestOCI1FromComponents(descriptorOCI1{descriptor: descriptor{
```
mtrmac: It would be nice to have tests for the new `annotations` fields, by adding them to `fixtures/oci1.json`, this function, and whatever else needs to know about annotations (probably `TestManifestOCI1UpdatedImage`).  Non-blocking I guess.

mtrmac:
> We could alternatively slip `Annotations` into the regular `descriptor` type, which would probably make the patch smaller.  I don't have a preference.

That would effectively add an unofficial extension to the schema2 format (though a very minor one, because the descriptor is unreachable using a public API); let’s not, please.

mtrmac:
:+1: again for `pullaprove`

runcom:
:+1:

PR #303, opened 2017-07-12
Preserve layer media types when updating images

nalind:
UpdatedImage() has been dropping the MediaType from layer descriptors when updating OCI v1 and Docker schema 2 images.  When updating them now, give the updated layer descriptors the media types of the original layers.


rhatdan:
LGTM

mtrmac:
:+1: This is clearly an improvement.

(Note it’s still not quite correct for OCI, which has separate compressed/uncompressed MIME types, and `ManifestUpdateOptions.LayerInfos` is primarily used when compressing the layers, i.e. when the MIME type should be changed; anyway, better than nothing.)

rhatdan:
@runcom PTAL

runcom:
LGTM

PR #305, opened 2017-07-13
Rework how we use containers/storage

nalind:
This is a fairly large patch, but I'm hoping the new version reads better than the old one.

HasBlob() learns to check for a blob in storage using its index-by-digest APIs, so we should be able to avoid pulling layers more often.

It reworks the storageImageDestination implementation to cache blobs in a temporary directory until Commit() is called.  This lets us do a couple of useful things at Commit() time:
* We can use the manifest's LayersInfos to guide applying layers to build up an image in storage, directly reusing layers when possible, using their contents when it isn't, or using cached blobs when there is no local copy.
* We can use the manifest's ConfigInfo to generate an ID for the local image, avoiding duplication on pulls.

When combined, the effect of these is that pulling a given image twice results in one image record in the local storage, and the second pull downloads only the manifest, configuration, and signatures.

It reworks the storageImageSource to return an updated version of the image's manifest that presents the layers in uncompressed form, so it can be used by copy.Image() now, and uses the newer DiffOptions to avoid needlessly recompressing things.  This turned up a bug in containers/storage, so we bump the version we vendor to one that includes a fix for that.


nalind:
Related currently-do-not-merge PRs in consumers:
* https://github.com/projectatomic/skopeo/pull/372
* https://github.com/kubernetes-incubator/cri-o/pull/654
* https://github.com/projectatomic/buildah/pull/187

rhatdan:
@nalind Did the tests freeze on this?

nalind:
@rhatdan Travis CI passes, but the code-review/pullapprove check is manual.
I think the changes I wanted to make are ready.  We make references that use digests usable at the cost of having to lie about the image reference returned by storageImageSource objects, but we can dependably read images now.  (The changes to make it possible to refer to images using references that include digests require some indirection to be able to retrieve layer lists from their manifests, when those manifests may no longer match those digests.)
CI is passing with updated containers/storage and copies of this branch in skopeo, buildah, and cri-o, so I think it's ready for review.

rhatdan:
@mtrmac @runcom PTAL

storage/storage_image.go
```diff
@@ -90,309 +95,654 @@ func newImageSource(imageRef storageReference) (*storageImageSource, error) {
...
+	return nil, -1, errors.Errorf(`error locating blob with blobsum %q`, blobinfo.Digest.String())
 }
 
 func (s *storageImageDestination) Commit() error {
-	// Create the image record.
+	// Find the list of layer blobs.  We have to implement enough of an ImageSource to be able to
+	// parse the manifest to get a list of which blobs are filesystem layers, leaving any cached
+	// files that aren't filesystem layers to be saved as data items.
+	if s.image == nil {
+		img, err := image.FromSource(&storageImageUnnamedDestination{named: s})
```
mtrmac: Also quite awkwkard, if the point is primarily to read the list of blobs.

storage/storage_image.go
```diff
@@ -90,309 +95,654 @@ func newImageSource(imageRef storageReference) (*storageImageSource, error) {
...
+		lastLayer = layer.ID
+		addedLayers = append([]string{lastLayer}, addedLayers...)
+	}
+	// If one of those blobs was a configuration blob, then we can try to dig out the date when the image
+	// was originally created, in case we're just copying it.  If not, no harm done.
+	var options *storage.ImageOptions
+	if inspect, err := s.image.Inspect(); err == nil {
+		logrus.Debugf("setting image creation date to %s", inspect.Created)
+		options = &storage.ImageOptions{
+			CreationDate: inspect.Created,
```
mtrmac: Ouch ouch ouch. This is the only place which requires constructing a `types.Image` (and actually makes `storageImageUnnamedDestination.GetBlob` necessary for `config.json`).

storage/storage_image.go
```diff
@@ -427,23 +777,58 @@ func (s *storageImageDestination) SupportedManifestMIMETypes() []string {
...
 }
 
-// PutManifest writes manifest to the destination.
-// FIXME? This should also receive a MIME type if known, to differentiate between schema versions.
-// If the destination is in principle available, refuses this manifest type (e.g. it does not recognize the schema),
-// but may accept a different manifest type, the returned error must be an ManifestTypeRejectedError.
+// GetManifest reads the manifest that we intend to store.  If we haven't been given one (yet?),
+// generate one.
+func (s *storageImageUnnamedDestination) GetManifest() ([]byte, string, error) {
+	if len(s.named.manifest) == 0 {
```
mtrmac: The other transports just fail if the client did not call `.PutManifest()`; are there users of containers-storage who need the manifest generated?

storage/storage_reference.go
```diff
@@ -30,6 +33,8 @@ func newReference(transport storageTransport, reference, id string, name referen
 		reference: reference,
 		id:        id,
 		name:      name,
+		tag:       tag,
+		digest:    digest,
```
mtrmac: Why aren’t the tag and digest values stored inside the `reference.Named`? That would eliminate the failure paths in `DockerReference` which now have to be silently ignored.

storage/storage_transport.go
```diff
@@ -101,60 +104,124 @@ func (s *storageTransport) DefaultGIDMap() []idtools.IDMap {
 // relative to the given store, and returns it in a reference object.
 func (s storageTransport) ParseStoreReference(store storage.Store, ref string) (*storageReference, error) {
```
mtrmac: FIXME: Didn’t really review the reference changes yet.

storage/storage_reference.go
mtrmac: (Also `reference` vs. `name`+`tag+`digest` seem redundant at a first glance, but I haven’t checked in detail.)

storage/storage_image.go
```diff
@@ -26,62 +31,62 @@ var (
...
-	Created        time.Time                   `json:"created-time,omitempty"`
-	ID             string                      `json:"id"`
-	BlobList       []types.BlobInfo            `json:"blob-list,omitempty"` // Ordered list of every blob the image has been told to handle
-	Layers         map[ddigest.Digest][]string `json:"layers,omitempty"`    // Map from digests of blobs to lists of layer IDs
-	BlobData       map[ddigest.Digest][]byte   `json:"-"`                   // Map from names of blobs that aren't layers to contents, temporary
-	Manifest       []byte                      `json:"-"`                   // Manifest contents, temporary
-	Signatures     []byte                      `json:"-"`                   // Signature contents, temporary
-	SignatureSizes []int                       `json:"signature-sizes"`     // List of sizes of each signature slice
+	directory      string                          // Temporary directory where we store blobs until Commit() time
+	counter        int32                           // A counter that we use for computing filenames to assign to blobs
```
nalind: Sure.  Changing it.

storage/storage_image.go
```diff
@@ -90,309 +95,654 @@ func newImageSource(imageRef storageReference) (*storageImageSource, error) {
...
+
+// Close cleans up any resources we tied up while reading the image.
+func (s *storageImageSource) Close() error {
+	return s.image.Close()
+}
+
+// newImageDestination sets us up to write a new image, caching blobs in a temporary directory until
+// it's time to Commit() the image
 func newImageDestination(imageRef storageReference) (*storageImageDestination, error) {
+	directory, err := ioutil.TempDir("", "storage")
```
nalind: Ah, hadn't spotted that.  Changing it.

mtrmac:
> The `ImageSource` path could be much clearer by splitting the relevant part of `UpdatedImage` into a separate manifest-only (i.e. not “consistent image”) updater

FWIW I’m working on this in https://github.com/mtrmac/image/tree/manifest-editing .

storage/storage_image.go
```diff
@@ -90,309 +95,654 @@ func newImageSource(imageRef storageReference) (*storageImageSource, error) {
...
+	var diffOptions *storage.DiffOptions
+	// We need a valid digest value.
+	err = info.Digest.Validate()
+	if err != nil {
+		return nil, -1, "", err
+	}
+	// Check if the blob corresponds to a diff that was used to initialize any layers, either
+	// before or after decompression, since we don't care.
+	uncompressedLayers, err := s.transport.store.LayersByUncompressedDigest(info.Digest)
+	compressedLayers, err := s.transport.store.LayersByCompressedDigest(info.Digest)
```
nalind: Yes, I suppose that though `storageUnnamedImageReader` is not dealing with an updated manifest, it's not going to get called to read blobs using the digests from the not-yet-updated manifest that it would return.  Reworking it to only look for layers using the index of uncompressed digests.

storage/storage_image.go
```diff
@@ -90,309 +95,654 @@ func newImageSource(imageRef storageReference) (*storageImageSource, error) {
...
+		layer = uncompressedLayers[i%len(uncompressedLayers)]
+	} else {
+		layer = compressedLayers[i%len(compressedLayers)]
+	}
+	layerID = layer.ID
+	// If we've been told to force decompression, do so.
+	if decompress {
+		noCompression := archive.Uncompressed
+		diffOptions = &storage.DiffOptions{
+			Compression: &noCompression,
```
nalind: Lately, `layerStore.Diff` only compresses a layer diff if it was originally applied in compressed form.  The files that comprise the layer are stored on disk in uncompressed form, of course, but a reader would historically ask for it using the digest from the unmodified manifest, so it was doing the best that it could by defaulting to using the same type of compression which was originally used for the layer.

A `storageImageSource` always asks for compression to be skipped, and since we're removing the ability to read layers using compressed digests from `storageUnnamedImageReader`, we don't need to bother with having it do things differently, so the `GetBlob` logic can be simplified.

storage/storage_image.go
```diff
@@ -90,309 +95,654 @@ func newImageSource(imageRef storageReference) (*storageImageSource, error) {
...
+		}
+		if layer.UncompressedSize < 0 {
```
nalind: Layers that have an `UncompressedDigest` value will have a valid `UncompressedSize`.  Layers that don't would have it set to an uninitialized 0, but we can assume it's there for anything returned by `LayersByUncompressedDigest`.  That's not an assumption we can make when walking the parent list, so I'll add those checks.

storage/storage_image.go
```diff
@@ -90,309 +95,654 @@ func newImageSource(imageRef storageReference) (*storageImageSource, error) {
...
+}
+
+// GetSignatures() parses the image's signatures blob into a slice of byte slices.
+func (s *storageUnnamedImageReader) GetSignatures() (signatures [][]byte, err error) {
+	var offset int
+	sigslice := [][]byte{}
+	signature, err := s.transport.store.ImageBigData(s.ID, "signatures")
+	if err != nil {
+		if !os.IsNotExist(errors.Cause(err)) {
+			logrus.Debugf("got error %v looking up signatures for image %q", err, s.ID)
```
nalind: A file-not-found error would be ignored, but I'll restructure it to not do that, and instead skip the attempt to read a file if we don't expect to find anything.

storage/storage_image.go
```diff
@@ -90,309 +95,654 @@ func newImageSource(imageRef storageReference) (*storageImageSource, error) {
...
+	options := types.ManifestUpdateOptions{
+		LayerInfos:      updatedBlobInfos,
+		InformationOnly: info,
+	}
+	// Return a version of the image that uses the updated manifest.
+	img, err := image.FromSource(reader)
+	if err != nil {
+		return nil, err
+	}
+	updated, err := img.UpdatedImage(options)
```
nalind: This is mainly about getting a manifest with the right digests, in the right format (i.e., one that matches the configuration blob), without hitting the check in `UnparsedImage.Manifest`.  If refactoring the manifest updating into a `Manifest` interface provides that, I'm all for it.

storage/storage_image.go
```diff
@@ -90,309 +95,654 @@ func newImageSource(imageRef storageReference) (*storageImageSource, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	// Discard a digest in the image reference, if there is one, since the updated image manifest
+	// won't match it.
+	newName := s.name
+	if s.digest != "" {
+		newName = reference.TrimNamed(newName)
+	}
+	newRef := newReference(reader.transport, verboseName(newName), reader.ID, newName, "", "")
```
nalind: That check is the main reason we have objects that contain modified image references in the second patch in the series, so if there was some way to bypass it, it would be simpler.

nalind: We have to drop the digest, if there is one, to avoid making `UnparsedImage.Manifest` return an error.  Dropping a tag, if that's what we have instead, is kind of a toss-up.  I'd rather not drop either, really.

storage/storage_image.go
```diff
@@ -90,309 +95,654 @@ func newImageSource(imageRef storageReference) (*storageImageSource, error) {
...
+		return nil, err
+	}
+	return &storageImageSource{image: image}, nil
+}
+
+// GetBlob returns either a data blob or an uncompressed layer blob.  In practice this avoids attempting
+// to recompress any layers that were originally delivered in compressed form, since we know that we
+// updated the manifest to refer to the blob using the digest of the uncompressed version.
+func (s *storageImageSource) GetBlob(info types.BlobInfo) (rc io.ReadCloser, n int64, err error) {
+	rc, n, _, err = s.image.reader.getBlobAndLayerID(info, true)
```
nalind: Changes suggested elsewhere mean that they'll probably be the same soon enough.

storage/storage_image.go
```diff
@@ -90,309 +95,654 @@ func newImageSource(imageRef storageReference) (*storageImageSource, error) {
...
 
-// putBlob stores a layer or data blob, optionally enforcing that a digest in
-// blobinfo matches the incoming data.
-func (s *storageImageDestination) putBlob(stream io.Reader, blobinfo types.BlobInfo, enforceDigestAndSize bool) (types.BlobInfo, error) {
+// PutBlob stores a layer or data blob in our temporary directory, checking that any information
+// in the blobinfo matches the incoming data.
+func (s *storageImageDestination) PutBlob(stream io.Reader, blobinfo types.BlobInfo) (types.BlobInfo, error) {
 	blobSize := blobinfo.Size
-	digest := blobinfo.Digest
+	blobDigest := blobinfo.Digest
```
nalind: Makes sense; dropping them.

nalind: Well, moving them.

storage/storage_image.go
nalind: The pathological case, I guess.  If `Commit()` is called without a manifest having been written, we would fail, but if that's what's expected for that case, it'll probably simplify things.

nalind:
Okay, I've just repushed with some of the smaller changes discussed above.  Is the `manifest-editing` branch stable enough that I can try rebasing onto it?

mtrmac:
> Is the `manifest-editing` branch stable enough that I can try rebasing onto it?

I’d call it a moderately useful proof of concept; it shows that the API basically works, but it’s missing a top-level dispatcher (`image.manifestInstanceFromBlob`-style), only the OCI implementation is finished, and most importantly it doesn’t have a solution for getting the creation date (which lives in the config, not in the manifest).

After grafting a stub for the dispatcher it could be useful for checking that the proposed approach handles everything needed for this PR, but it can’t be used to get a complete working implementation yet I’m afraid.

storage/storage_image.go
mtrmac: Perhaps something like
```go
type ImageSource interface {
   …
   // AcceptManifestDigestMismatch returns true if the references can contain manifest digests,
   // and the source is _expected_ to return manifest which don’t match them (e.g. when the
   // transport is throwing the manifests away and recreating them on demand).
   // Note that this DOES NOT and MUST NOT cause signatures with mismatching digests to be
   // accepted!
   func AcceptManifestDigestMismatch() bool
```
(someone else can probably think of a better name), and after #288 is merged, also “see also `GetOriginalManifest`”

storage/storage_image.go
mtrmac: There have been a few cases suggested to allow adding / modifying signatures in an `ImageDestination` without doing anything about the blobs/manifests… but that’s not quite relevant to this case.

In theory calling `PutBlob` is expected to also call `PutManifest`; if buildah or some other caller isn’t doing that currently, I guess I’d rather have them explicitly construct a manifest (using the discussed manifest interface I guess) than to have every destination deal with constructing a manifest from its own data, and do the otherwise often unnecessary bookkeeping to track all uploaded / reapplied blobs.

mtrmac:
@nalind 

https://github.com/mtrmac/image/tree/manifest-editing is now a reasonable proof of concept (no tests, and no way to get image creation date).

https://github.com/mtrmac/image/tree/storage-prototype shows how this could be applied to the containers/storage transport (with some horrible merges and commits that should be split into smaller pieces). Again, still can’t get image creation date, and does nothing about the reference handling (e.g. still overrides the user’s input for `storageImageSource`); and untested except that unprivileged tests run.

Does this look like a reasonable approach?

nalind:
Aaargh, missed that last comment.  Yes, I think that'll work.  Should I rebase on top of the manifest-editing branch, or do you plan to open a PR for merging it soonish?

mtrmac:
I’m not sure whether I will be able to finish the manifest-editing branch “soonish”; if you prefer to take it over and finish the missing pieces, feel free to do so.

nalind:
That's fine.  Do you have a list of what you still want to accomplish in it but haven't gotten to yet?  You mentioned a top-level manifest parsing dispatcher a la `image.manifestInstanceFromBlob` above, which seems pretty straightforward, and being able to retrieve the image creation date.  For schema 2, that probably requires more than just the manifest, so there might not be much room to improve over the version that expects an image.

I think we still need a way to update the media type so that we change the type of compression that's being indicated for a layer when we call UpdateLayerInfos().

mtrmac:
> That's fine.  Do you have a list of what you still want to accomplish in it but haven't gotten to yet?  You mentioned a top-level manifest parsing dispatcher a la `image.manifestInstanceFromBlob` above

That actually exists in the storage-prototype branch, as `manifest.FromBlob`.

> and being able to retrieve the image creation date.  For schema 2, that probably requires more than just the manifest, so there might not be much room to improve over the version that expects an image.

Yeah, this requires access to the config, which is rather out of scope for a `manifest` package. Still, that’s much less than requiring a full `ImageSource`.

I was tentatively planning something like
```go
func Inspect(configGetter func(types.BlobInfo) ([]byte, error)) (*types.ImageInspectInfo, error)
```
which is pretty much the opposite of elegant, but still much easier to call from `containers-storage:` than implementing `GetManifest`+`GetSignatures` unnecessarily.

>  think we still need a way to update the media type so that we change the type of compression that's being indicated for a layer when we call UpdateLayerInfos().

Good point; but that’s pretty independent from the rest of this work; whether it happens on a manifest editor or on `types.Image`, the code should be exactly the same.

mtrmac:
… One thing I’ve been wondering about is testing of the `manifest` subpackage. Of course all of the tests in `image` test `manifest` equivalently, so this is not at all a blocking concern. This basically only impacts measured code coverage, which does not count the `manifest` code as tested by the `image` tests; and low apparent coverage numbers might decrease motivation to keep up testing for newly added code.

Trivially we could just copy&paste the fixtures and most tests from `image` to `manifest`, at the cost of having to keep updating both in the future. I was wondering about splitting the two better (or maybe further simplifying `image` now that so much of `genericManifest` is just forwarded to the underlying `image` implementation), or at least sharing the fixtures.

Anyway, again, absolutely non-blocking for me.

nalind:
Okay, finished rebasing onto your storage-prototype branch, then
* fixed `TestTransportParseStoreReference` so that it tests the right thing
* dropped `TestCommitWithoutManifest`, since that functionality's out
* reintroduced `storageImage.Size()`, which got lost in the shuffle somewhere
* added configuration types to the `manifest` package, and replaced use of the v1 compatibility structure in the `image` package with the `manifest` version, which has the same fields
* implemented `Inspect()` for types in the `manifest` package
* used both of the above to compute image IDs for schema1 images.  Assuming the rule for OCI is "digest the config blob", as it is for schema 2, we should be good there.
* introduced an `UpdatedLayerInfos()` method to supply layerinfo for `copy.Image()`, and took rewriting completely out of `GetManifest`.  This lets us read back images that are identified by canonical digests and still manage to copy them out.  It solves the can't-copy-out problem for my cases, but it's very specific, so I'm not wedded to it.
* break retrieval of `DockerReference`s from `storageImageDestination`.  This is better than doing it in `storageImageSource`, since the information is actually preserved and can be read back later, but it's there as a workaround to prevent editing of schema1 manifests to remove embedded references while copying them.  When we weren't doing that, attempts to use a local copy of a schema1 image via a canonical reference would fail because the manifest no longer matched.  I'm not super-fond of this, either, so if there's a better way to avoid this problem, I'm all for changing it.

nalind:
Updated the schema 2 and oci manifest implementations to return MediaType information from `ConfigInfo()` and `LayerInfos()`, and to have `storageImageSource.UpdatedLayerInfos()` return a value as well.

runcom:
ping @mtrmac, I see there are just to UNFINISHED commit here, what should we do? I believe this patch is needed to fix https://github.com/kubernetes-incubator/cri-o/issues/1048

@nalind could you correct me if I'm wrong in saying that this patches along with https://github.com/kubernetes-incubator/cri-o/pull/654 should actually fix https://github.com/kubernetes-incubator/cri-o/issues/1048

nalind:
@runcom, @mtrmac, I think the patches that are marked as unfinished have their unfinished portions finished later in the series.  We add an Inspect() method for manifest objects, and use it as part of setting the creation date that the storage library tracks for an image.  We also switch to preserving the original reference in source images, though that means we need to forget them in destination images to avoid triggering editing to remove embedded references, which would otherwise change the manifest digest, which would break subsequent attempts to read the image if the reference used for locating them contained a digest.
@runcom Yes, I think this set of changes, which are bundled up in https://github.com/kubernetes-incubator/cri-o/pull/654 so that cri-o's tests can chew on them, will fix https://github.com/kubernetes-incubator/cri-o/issues/1048.

runcom:
> @runcom Yes, I think this set of changes, which are bundled up in kubernetes-incubator/cri-o#654 so that cri-o's tests can chew on them, will fix kubernetes-incubator/cri-o#1048.

@nalind, awesome, I'll test kubernetes-incubator/cri-o#654 to see if it fixes kubernetes-incubator/cri-o#1048 then.

@mtrmac are you ok merging this?

runcom:
Seems like this isn't fixing that CRI-O issue https://github.com/kubernetes-incubator/cri-o/pull/654#issuecomment-340463430

runcom:
LGTM

runcom:
@mtrmac PTAL

runcom:
@mtrmac PTAL

nalind:
Rebased to pick up the tarball transport, corrected the commit ID referenced in the "Update containers/storage" log message to match the change's contents.

storage/storage_image.go
```diff
@@ -93,309 +90,655 @@ func newImageSource(imageRef storageReference) (*storageImageSource, error) {
...
+			Type: "layers",
+		}
+		for _, fslayer := range s1.FSLayers {
+			blobSum := fslayer.BlobSum
+			diffID, ok := s.blobDiffIDs[blobSum]
+			if !ok {
+				logrus.Infof("error looking up diffID for blob %q", string(blobSum))
+				return ""
+			}
+			s2.RootFS.DiffIDs = append([]digest.Digest{diffID}, s2.RootFS.DiffIDs...)
```
mtrmac: AFAICT layers marked as `ThrowAway` in `history` are not added to `DiffIDs` in docker/docker.

storage/storage_image.go
```diff
@@ -93,309 +90,655 @@ func newImageSource(imageRef storageReference) (*storageImageSource, error) {
...
+		// For Schema2 and OCI1(?), the ID is just the hex part of the digest of the config blob.
+		logrus.Debugf("trivial image ID for configured blob")
+		configInfo := m.ConfigInfo()
+		if configInfo.Digest.Validate() == nil {
+			return configInfo.Digest.Hex()
+		}
+		return ""
+	case manifest.DockerV2Schema1MediaType, manifest.DockerV2Schema1SignedMediaType:
+		// Convert the schema 1 compat info into a schema 2 config, constructing some of the fields
+		// that aren't directly comparable using info from the manifest.
```
mtrmac: Could this share code with `image.manifestSchema1.convertToManifestSchema2`? This is complex enough that not having to maintain two copies would be nice, and the other one has at least a smoke test via `TestManifestSchema1ToOCIConfig`.

image/memory.go
```diff
@@ -71,3 +71,10 @@ func (i *memoryImage) Inspect() (*types.ImageInspectInfo, error) {
 func (i *memoryImage) IsMultiImage() bool {
 	return false
 }
+
+// UpdatedLayerInfos returns an updated set of layer blob information which may not match the manifest.
+// The Digest field is guaranteed to be provided; Size may be -1.
+// WARNING: The list may contain duplicates, and they are semantically relevant.
+func (i *memoryImage) UpdatedLayerInfos() []types.BlobInfo {
+	return i.LayerInfos()
```
mtrmac: Can’t this be just `return nil`?

types/types.go
```diff
@@ -207,6 +211,10 @@ type UnparsedImage interface {
 	Manifest() ([]byte, string, error)
 	// Signatures is like ImageSource.GetSignatures, but the result is cached; it is OK to call this however often you need.
 	Signatures(ctx context.Context) ([][]byte, error)
+	// UpdatedLayerInfos returns either nil (meaning there are no updates), or updated values for the layer blobsums that are listed in the image's manifest.
+	// The Digest field is guaranteed to be provided, Size may be -1 and MediaType may be optionally provided.
+	// WARNING: The list may contain duplicates, and they are semantically relevant.
+	UpdatedLayerInfos() []BlobInfo
```
mtrmac: FIXME

At a first glance this seems risky; it allows `Manifest()` to be validated against `Signatures()`, but the user will be encouraged to ignore the validated data and use `UpdatedLayerInfos()` instead. (The `!ic.canModifyManifest` check in `copy.Image` does prevent this from happening currently, but that’s rather non-obvious.)

#288 instead introduces `ImageSource.GetOriginalManifest` etc., forcing the caller to make a choice in advance between getting the original manifest (which may be unusable for reading data) and getting the current one (which may break signatures).  I’m not at all sure that is better, though.

storage/storage_image.go
```diff
@@ -93,309 +90,655 @@ func newImageSource(imageRef storageReference) (*storageImageSource, error) {
...
+	directory, err := ioutil.TempDir(temporaryDirectoryForBigFiles, "storage")
+	if err != nil {
+		return nil, errors.Wrapf(err, "error creating a temporary directory")
+	}
+	// Break reading of the reference we're writing, so that copy.Image() won't try to rewrite
+	// schema1 image manifests to remove embedded references, since that changes the manifest's
+	// digest, and that makes the image unusable if we subsequently try to access it using a
+	// reference that mentions the no-longer-correct digest.
+	publicRef := imageRef
+	publicRef.name = nil
```
mtrmac: `ImageDestination.Reference` is documented to record user intent, and stripping data like this is surprising; all the more when setting `.name` to `nil` does not change `StringWithinTransport`, i.e. there can be two different references with exactly equal string form but different behavior.

Could the `ImageDestination` instead get a direct ability to opt out of the `EmbeddedDockerReference` update, by adding a `NeedsEmbeddedDockerReferenceUpdate() bool` or something like that?

(FIXME? I guess this interacts with the `UpdatedLayerInfos` work which also aims to preserve the original manifest digest.)

manifest/docker_schema2.go
```diff
@@ -0,0 +1,241 @@
...
+	Interval time.Duration `json:",omitempty"` // Interval is the time to wait between checks.
+	Timeout  time.Duration `json:",omitempty"` // Timeout is the time to wait before considering the check to have hung.
+
+	// Retries is the number of consecutive failures needed to consider a container as unhealthy.
+	// Zero means inherit.
+	Retries int `json:",omitempty"`
+}
+
+// Schema2Config is a Config in docker/docker/api/types/container.
+type Schema2Config struct {
```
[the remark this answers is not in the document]
mtrmac: … and in `docker/tarfile/types.go`.

types/types.go
mtrmac: Note to self: If the same per-transport policy is used for copying out of the storage, and for run decisions within, what does that mean for the original vs. modified manifest?

image/memory.go
nalind: Hmm, yeah, that'd help ensure that we don't hit the `srcInfosUpdated = true` path in `copyLayers`.  Changing it.

manifest/docker_schema1.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+	return &types.ImageInspectInfo{
+		Tag:           m.Tag,
+		Created:       s1.Created,
+		DockerVersion: "",
+		Labels:        make(map[string]string),
+		Architecture:  "",
+		Os:            "",
+		Layers:        []string{},
+	}, nil
```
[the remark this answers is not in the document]
nalind: Whoops, no it wasn't intentional.  Fixing.

manifest/docker_schema2.go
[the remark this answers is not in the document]
nalind: Right, I wasn't 100% sure we wanted to do that in here, too.  Switching them over.

manifest/docker_schema2.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+	s2 := &Schema2Image{}
+	if err := json.Unmarshal(config, s2); err != nil {
+		return nil, err
+	}
+	return &types.ImageInspectInfo{
+		Tag:           "",
+		Created:       s2.Created,
+		DockerVersion: s2.DockerVersion,
+		Labels:        s2.Config.Labels,
```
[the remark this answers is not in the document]
nalind: Ah, right.  Adding a check for that.

manifest/oci.go
```diff
@@ -0,0 +1,108 @@
...
+		return nil, err
+	}
+	v1 := &imgspecv1.Image{}
+	if err := json.Unmarshal(config, v1); err != nil {
+		return nil, err
+	}
+	created := time.Time{}
+	if v1.Created != nil {
+		created = *v1.Created
+	}
```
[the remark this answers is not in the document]
nalind: Do you mean by changing the field to a pointer?

manifest/oci.go
```diff
@@ -0,0 +1,108 @@
...
+	return &types.ImageInspectInfo{
+		Tag:           "",
+		Created:       created,
+		DockerVersion: "",
+		Labels:        v1.Config.Labels,
```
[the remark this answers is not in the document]
nalind: Ah, right. Adding a check for that.

manifest/manifest.go
```diff
@@ -38,6 +40,33 @@ var DefaultRequestedManifestMIMETypes = []string{
...
+	// The Digest field is guaranteed to be provided; Size may be -1.
+	// WARNING: The list may contain duplicates, and they are semantically relevant.
+	LayerInfos() []types.BlobInfo
+	// UpdateLayerInfos replaces the original layers with the specified BlobInfos (size+digest+urls), in order (the root layer first, and then successive layered layers)
+	UpdateLayerInfos(layerInfos []types.BlobInfo) error
+
+	// Inspect returns various information for (skopeo inspect) parsed from the manifest,
+	// incorporating information from a configuration blob returned by configGetter, if
+	// the underlying image format is expected to include a configuration blob.
+	Inspect(configGetter func(types.BlobInfo) ([]byte, error)) (*types.ImageInspectInfo, error)
```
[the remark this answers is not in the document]
nalind: At this point we have everything we need in order to just fill in that information, so I'll change it to do that.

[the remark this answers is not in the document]
nalind: Yes, that would remove some duplication.  Changing them.

manifest/oci.go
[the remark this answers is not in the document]
nalind: Oh, wait, it's not nil in an OCI Image, so we can't compare v1.Config to nil.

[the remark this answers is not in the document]
nalind: But it can be tweaked to try a decode using the Schema2V1Image format, to read the docker_version field.  Adding that.

storage/storage_image.go
```diff
@@ -93,309 +90,655 @@ func newImageSource(imageRef storageReference) (*storageImageSource, error) {
...
-		return types.BlobInfo{}, err
+		return types.BlobInfo{}, errors.Wrapf(err, "error reapplying blob %+v", blobinfo)
 	}
-	if layerList, ok := s.Layers[blobinfo.Digest]; !ok || len(layerList) < 1 {
-		b, err := s.imageRef.transport.store.ImageBigData(s.ID, blobinfo.Digest.String())
+	blobinfo.Size = size
+	return blobinfo, nil
+}
+
+// computeID computes a recommended image ID based on information we have so far.
```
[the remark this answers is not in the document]
nalind: Adding that.

storage/storage_image.go
```diff
@@ -93,309 +90,655 @@ func newImageSource(imageRef storageReference) (*storageImageSource, error) {
...
+func (s *storageImageDestination) computeID(m manifest.Manifest) string {
+	mb, err := m.Serialize()
+	if err != nil {
+		return ""
+	}
+	switch manifest.GuessMIMEType(mb) {
```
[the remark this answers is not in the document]
nalind: Okay, I'm trying to rework this as Manifest.ID().

storage/storage_image.go
nalind: Okay, broke the Schema1 implementation of ID() into two steps: creating the config blob and digesting it, and replaced parts of image.manifestSchema1.convertToManifestSchema2() with a call to that.

storage/storage_image.go
nalind: Yup, fixing this.

storage/storage_image.go
```diff
@@ -93,309 +90,655 @@ func newImageSource(imageRef storageReference) (*storageImageSource, error) {
...
+		s1, ok := m.(*manifest.Schema1)
+		if !ok {
+			logrus.Debugf("schema type was guessed wrong?")
+			return ""
+		}
+		if len(s1.History) == 0 {
+			logrus.Debugf("image has no layers")
+			return ""
+		}
+		s2 := struct {
```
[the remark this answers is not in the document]
nalind: That type is what the `github.com/docker/distribution/manifest/schema1/config_builder.go` logic builds.  This definition keeps us from discarding parts of the `Config` structure, which can include more details than that logic knows how to encode (perhaps having been generated by a toolchain that doesn't exist any more?), so that we can incorporate its contents into the ID calculation.

storage/storage_image.go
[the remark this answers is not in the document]
nalind: Actually, making it part of the Manifest interface means it has to be implemented, so unknown should never happen.

storage/storage_image.go
nalind: Yeah, the end-goal is to be able to read the layer blobs from the image when the digests of the blobs don't match the manifest, when changing the manifest would make the image inaccessible due to the `MatchesDigest` check failing.  `UpdatedLayerInfos` tries to address the former case, which affects images with all manifest types.  This hunk tries to avoid triggering an edit for stripping out an embedded docker reference, which mainly affects Schema1 images.

types/types.go
nalind: I think we'd need some way to affect which variation `copy.Image()` asked for, but otherwise I think that would work.

nalind:
Updated, I think addressing everything except the `UpdatedLayerInfos()` and `storageImageDestination.Reference()` points.
https://github.com/projectatomic/skopeo/pull/372, https://github.com/projectatomic/buildah/pull/187, and https://github.com/kubernetes-incubator/cri-o/pull/1116 have revendored this branch, and their CI routines are chewing on it.

nalind:
Rebased.

types/types.go
runcom: > #288 instead introduces ImageSource.GetOriginalManifest etc., forcing the caller to make a choice in advance between getting the original manifest (which may be unusable for reading data) and getting the current one (which may break signatures). I’m not at all sure that is better, though.

@mtrmac is it possible to extract `GetOriginalManifest` to another new PR with just that so we can merge it and reuse it here for @nalind ? (or carry the patch here directly)

nalind:
Rebased.

mtrmac:
FWIW I have rebased the `manifest-editing` and `storage-prototype` branches on top of the merged #346.  The Git history is ugly, hopefully it’s useful to at least show what the update could look like.

runcom:
This needs another rebase

nalind:
Rebased.  The `ImageCloser` changes required a few changes to the earlier patches in the series to account for them (we had to embed different Image and Closer interfaces in storageImageCloser, instead of using an ImageCloser, at one point).

nalind:
Rebased and tweaked looking up images by canonical references.

nalind:
Corrected `GetStoreImage()` to return `github.com/containers/storage.ErrImageUnknown` instead of `github.com/containers/image/storage.ErrInvalidReference` when it fails.

nalind:
Renamed `UpdatedLayerInfos()` to `LayerInfosForCopy()`.

nalind:
Adding a change to start setting the Digest field in images that we create, using the digest of the manifest as given by manifest.Digest().

nalind:
Fixed some issues around parsing references that were possibly-truncated image IDs.

runcom:
CI is still failing, I'll merge this once it's green

runcom:
Ci is green, I'm merging, @nalind please update skopeo and CRI-O (and c/storage if needed)

giuseppe:
@nalind @runcom I've adjusted the ostree implementation to take advantage of LayerInfosForCopy.  Pretty neat and works well:

https://github.com/containers/image/pull/392

PR #307, opened 2017-07-18
docker: loop auth handlers instead of using just one

runcom:
As already done docker-side at
https://github.com/moby/moby/blob/master/vendor/github.com/docker/distribution/registry/client/auth/session.go#L98

@mtrmac PTAL this fixes https://github.com/projectatomic/skopeo/issues/375

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -308,31 +308,33 @@ func (c *dockerClient) setupRequestAuth(req *http.Request) error {
 	if len(c.challenges) == 0 {
 		return nil
 	}
-	// assume just one...
-	challenge := c.challenges[0]
-	switch challenge.Scheme {
-	case "basic":
-		req.SetBasicAuth(c.username, c.password)
-		return nil
-	case "bearer":
-		if c.token == nil || time.Now().After(c.tokenExpiration) {
-			realm, ok := challenge.Parameters["realm"]
-			if !ok {
-				return errors.Errorf("missing realm in bearer auth challenge")
+	for _, challenge := range c.challenges {
+		switch challenge.Scheme {
+		case "basic":
+			req.SetBasicAuth(c.username, c.password)
+			return nil
+		case "bearer":
+			if c.token == nil || time.Now().After(c.tokenExpiration) {
+				realm, ok := challenge.Parameters["realm"]
+				if !ok {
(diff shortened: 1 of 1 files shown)


docker/docker_client.go
```diff
@@ -308,31 +308,33 @@ func (c *dockerClient) setupRequestAuth(req *http.Request) error {
...
-			c.tokenExpiration = token.IssuedAt.Add(time.Duration(token.ExpiresIn) * time.Second)
+			req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", c.token.Token))
+			return nil
+		default:
+			logrus.Debugf("no handler for %s authentication", challenge.Scheme)
 		}
-		req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", c.token.Token))
-		return nil
 	}
-	return errors.Errorf("no handler for %s authentication", challenge.Scheme)
```
rhatdan: Shouldn't this still be a failure?  If it found no successfull authorization/Authentication?

runcom: seems like the original docker/distribution threats "no handler" as "do not auth this call" which sounds fine and actually fix this.

mtrmac: I guess it can't _hurt_ to try making the request without authentication, but it would be nice to log that only unsupported challenges (and a list?), using a non-debug level (info?) at this point.

runcom: mmm alright, what do you want in the info log then?

mtrmac: …that only unsupported challenges were sent by the server.

mtrmac: Something like
```go
logrus.Infof("None of the challenges sent by server (%s) are supported", strings.Join(schemeList, ", "))
```
with schemeList a list of the `challenge.Scheme` values collected throughout the loop, perhaps?

mtrmac:
:+1:

This works as is, though I’d be happier with logging of unsupported challenges as discussed above.

runcom:
LGTM

docker/docker_client.go
runcom: I'll defer this to a follow up since I'm updating c/image in docker and I would like to see this

mtrmac: #308

PR #309, opened 2017-07-18
Add LoadTarManifest to get the manifest from docker-archive

umohnani8:
Need it in kpod pull and kpod load. These commands need all the manifests, so took out the length checking block from loadTarManifest.

### DIFF
--- a/docker/tarfile/src.go
+++ b/docker/tarfile/src.go
@@ -145,23 +145,28 @@ func (s *Source) ensureCachedDataIsPresent() error {
 		return err
 	}
 
+	// Check to make sure length is 1
+	if len(tarManifest) != 1 {
+		return errors.Errorf("Unexpected tar manifest.json: expected 1 item, got %d", len(tarManifest))
+	}
+
 	// Read and parse config.
-	configBytes, err := s.readTarComponent(tarManifest.Config)
+	configBytes, err := s.readTarComponent(tarManifest[0].Config)
 	if err != nil {
 		return err
 	}
 	var parsedConfig image // Most fields ommitted, we only care about layer DiffIDs.
 	if err := json.Unmarshal(configBytes, &parsedConfig); err != nil {
-		return errors.Wrapf(err, "Error decoding tar config %s", tarManifest.Config)
+		return errors.Wrapf(err, "Error decoding tar config %s", tarManifest[0].Config)
 	}
 
-	knownLayers, err := s.prepareLayerData(tarManifest, &parsedConfig)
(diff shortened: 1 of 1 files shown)


umohnani8:
@rhatdan PTAL

rhatdan:
We are going to need similar functionality to this to get the tags out of the oci image spec.

docker/tarfile/src.go
```diff
@@ -180,10 +180,12 @@ func (s *Source) loadTarManifest() (*manifestItem, error) {
 	if err := json.Unmarshal(bytes, &items); err != nil {
 		return nil, errors.Wrap(err, "Error decoding tar manifest.json")
 	}
-	if len(items) != 1 {
-		return nil, errors.Errorf("Unexpected tar manifest.json: expected 1 item, got %d", len(items))
```
mtrmac: This check should remain in place when using `Source` as a `types.ImageSource` backend.

docker/tarfile/src.go
```diff
@@ -180,10 +180,12 @@ func (s *Source) loadTarManifest() (*manifestItem, error) {
...
-	}
-	return &items[0], nil
+	return items, nil
+}
+
+// LoadTarManifest loads and decodes the manifest.json
+func (s *Source) LoadTarManifest() ([]manifestItem, error) {
```
mtrmac: Is it acceptable to return a private type from a public function? AFAIK the caller won’t be able to refer to the type in parameter/return value declarations.

(If it works for you, fine; just make sure that it does.)

rhatdan: It is working in her current code.

docker/tarfile/src.go
rhatdan: Can we move it to the other function? The only one that calls this?  Then we would only have one loadmanafest function.

docker/tarfile/src.go
umohnani8: Yup it works fine. I looked it up and a private type can be accessed from an exported function, but the private type can't be created outside its own package.

mtrmac:
:+1:

docker/tarfile/src.go
umohnani8: @mtrmac I added the error check in ensureCachedDataIsPresent() after loadTarManifest() is being called.

mtrmac: That’s better, sure.

rhatdan:
@runcom PTAL

runcom:
LGTM

PR #310, opened 2017-07-19
Make docker/tarfile.manifestItem public

mtrmac:
Otherwise lint fails on Fedora 24:
> …/docker/tarfile/src.go:192:37: exported method LoadTarManifest returns unexported type []tarfile.manifestItem, which can be annoying to use

(Yes, #309 did pass tests. I don’t know what makes the difference. Anyway, I need the warning to go away to keep my workflow clean.)

This should not affect users of #309 because they were not able to name the type before anyway.


TomSweeneyRedHat:
LGTM

rhatdan:
LGTM

runcom:
LGTM

Issue #306, opened 2017-07-13
oci transport will remove existing images


cyphar:
If you try to have a multi-tag image with `skopeo` (by copying to different tags), then it will delete old tags. I also have a feeling that manifests without tags will be _really_ broken. You can verify this fairly trivially by doing:

```
% skopeo copy docker://opensuse/amd64:42.2 oci:opensuse:latest
% skopeo copy docker://opensuse/amd64:42.1 oci:opensuse:42.1
% jq . opensuse/index.json
{
  "schemaVersion": 2,
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:6842a022e2395f4d03350f5917908b4efc1aa80c302990dc74a577ebd7df0bff",
      "size": 349,
      "annotations": {
        "org.opencontainers.image.ref.name": "42.1"
      },
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      }
    }
  ]
}
```

I'll investigate this later. I found this while trying to increase umoci's testing.

cyphar:
For reference, IMO this should be handled in a similar way to umoci's `oci/casext.UpdateReference` implementation.

cyphar:
I believe this is related to projectatomic/skopeo#369.

mtrmac:
I don’t think this is https://github.com/projectatomic/skopeo/issues/369 ; it’s just that `ociDestination.Commit` ignores any pre-existing index and overwrites it with a single-manifest index for the just written manifest. That’s not intentional, contributions to fix this would be welcome.

(Also discussed in https://github.com/containers/image/issues/155 .)

rhatdan:
@mtrmac @cyphar Is this still an issue,  Just going through old issues and trying to clean them up.

mtrmac:
This was fixed in https://github.com/containers/image/pull/326 ; when a tag is specified, only that image in the index is added/replaced without affecting other tags.  (When no tag is specified, it seems that the first untagged image or image with an empty tag is added/replaced.)

(Note that this does not allow multi-arch images, with the various architectures sharing a tag, yet; see #400 for some work in that direction.)

PR #313, opened 2017-07-20
Add fixes to newImageDestination to allow kpod save to write to a file when stdout is redirected

umohnani8:
Need this fix for kpod load so that stdout can write to a file when redirected. Checks if the file exists and is empty first as the file is created then written to.

### DIFF
--- a/docker/archive/dest.go
+++ b/docker/archive/dest.go
@@ -19,15 +19,24 @@ func newImageDestination(ctx *types.SystemContext, ref archiveReference) (types.
 	if ref.destinationRef == nil {
 		return nil, errors.Errorf("docker-archive: destination reference not supplied (must be of form <path>:<reference:tag>)")
 	}
-	fh, err := os.OpenFile(ref.path, os.O_WRONLY|os.O_EXCL|os.O_CREATE, 0644)
+
+	// ref.path can be either a pipe or a regular file
+	// in the case of a pipe, we require that we can open it for write
+	// in the case of a regular file, we don't want to overwrite any pre-existing file
+	// so we check for Size() == 0 below (This is racy, but using O_EXCL would also be racy,
+	// only in a different way. Either way, it’s up to the user to not have two writers to the same path.)
+	fh, err := os.OpenFile(ref.path, os.O_WRONLY|os.O_CREATE, 0644)
 	if err != nil {
-		// FIXME: It should be possible to modify archives, but the only really
-		//        sane way of doing it is to create a copy of the image, modify
-		//        it and then do a rename(2).
-		if os.IsExist(err) {
-			err = errors.New("docker-archive doesn't support modifying existing images")
-		}
-		return nil, err
+		return nil, errors.Wrapf(err, "error opening file %q", ref.path)
+	}
+
+	fhStat, err := fh.Stat()
+	if err != nil {
+		return nil, errors.Wrapf(err, "error statting file %q", ref.path)
+	}
+
+	if fhStat.Mode().IsRegular() && fhStat.Size() != 0 {
+		return nil, errors.New("docker-archive doesn't support modifying existing images")
 	}
 
 	return &archiveImageDestination{


rhatdan:
This patch fixes the issue where a user executes
kpod save > fedora.tar
In this situation the bash command creates an empty file, which the image layer should not reject.

docker/archive/dest.go
```diff
@@ -19,17 +19,32 @@ func newImageDestination(ctx *types.SystemContext, ref archiveReference) (types.
...
+			//        it and then do a rename(2).
+			if os.IsExist(err) {
+				err = errors.New("docker-archive doesn't support modifying existing images")
+			}
+			return nil, err
+		}
+	} else {
+		fhStat, err := fh.Stat()
+		if err != nil {
+			return nil, err
```
rhatdan: Probably should wrap this err with some more useful message.
errors.Wrapf(err, "docker-archive file %v can not be stat'd", ref.path)

docker/archive/dest.go
```diff
@@ -19,17 +19,32 @@ func newImageDestination(ctx *types.SystemContext, ref archiveReference) (types.
 	if ref.destinationRef == nil {
 		return nil, errors.Errorf("docker-archive: destination reference not supplied (must be of form <path>:<reference:tag>)")
 	}
-	fh, err := os.OpenFile(ref.path, os.O_WRONLY|os.O_EXCL|os.O_CREATE, 0644)
+	// tries to open a file first to accomodate for load when using stdout
+	// as the file is created first then written to
+	// if the file doesn't exist then it creates a new file for exclusive use
+	// if the file exists, it checks to ensure the size is 0
```
mtrmac: “to accomodate for load” is difficult to understand for me as a non-native speaker.

(In general, don’t write comments describing what every line does; that should be obvious from the code, and such comments are over time more likely to become obsolete and misleading than to be helpful; instead, write comments saying what is the code trying to do, most importantly the situation/problem (“Note that this needs to account for $this_weird_case”), and, if non-obvious, the solution.)

docker/archive/dest.go
```diff
@@ -19,17 +19,32 @@ func newImageDestination(ctx *types.SystemContext, ref archiveReference) (types.
...
+		}
+	} else {
+		fhStat, err := fh.Stat()
+		if err != nil {
+			return nil, err
 		}
-		return nil, err
-	}
 
+		if fhStat.Size() > 0 {
```
mtrmac: Awkwardly, AFAICT it is not actually documented anywhere that `struct stat.st_size` has any particular value for pipes and FIFOs. I guess this could explicitly check for `(fhStat.Mode()&ModeType) == ModeNamedPipe || (fhStat.Mode().IsRegular() && fhStat.Size() == 0)` or something like that.

docker/archive/dest.go
```diff
@@ -19,17 +19,32 @@ func newImageDestination(ctx *types.SystemContext, ref archiveReference) (types.
...
+	// if the file doesn't exist then it creates a new file for exclusive use
+	// if the file exists, it checks to ensure the size is 0
+	fh, err := os.OpenFile(ref.path, os.O_WRONLY, 0644)
 	if err != nil {
-		// FIXME: It should be possible to modify archives, but the only really
-		//        sane way of doing it is to create a copy of the image, modify
-		//        it and then do a rename(2).
-		if os.IsExist(err) {
-			err = errors.New("docker-archive doesn't support modifying existing images")
+		fh, err = os.OpenFile(ref.path, os.O_WRONLY|os.O_EXCL|os.O_CREATE, 0644)
```
mtrmac: I don’t think using `O_EXCL` with the same path again is all that valuable; either `err` was an unrelated failure and this should fail again in the same way, or we already know that a file did not exist, so if it exists now and `O_EXCL` causes us to fail, we are racing against something else modifying the filesystem and failing is not necessarily the right thing to do.

Couldn’t this be simplified to a single `os.OpenFile(…O_WRONLY | O_CREATE)` call followed by the “size of the file (whether pre-existing or newly created) is 0” check? (That’s also racy against concurrent writes to the same filename, though differently, but at least it is trivially simple.)

docker/archive/dest.go
rhatdan: Yes lets remove the comment about load, and just state that the file needs to be created new and open for write, if the file prexisted it needs to be null for now.  @mtrmac @umohnani8 is going to work on the append to the tar ball soon, but we need to get this in now.

docker/archive/dest.go
rhatdan: SGTM

docker/archive/dest.go
rhatdan: I agree.

docker/archive/dest.go
mtrmac: > append to the tar ball soon

You mean to update an archive by adding new files to the end of an existing tarball? That works with the “tar as tape processed at once” model (and seems to be documented at https://www.gnu.org/software/tar/manual/html_node/multiple.html ), but it's not obviously interoperable (in particular, the reader in this package stops at the first match). Of course that particular reader can be updated, but in general?

Ultimately, what is the use case here? Why would anyone want to keep updating a set of images in a giant single file with O(N) seeking and no way to delete a layer in the middle, as opposed to using either a registry or the OCI expanded dirs+files format?

rhatdan: We are trying to match the docker save command, their use case is to put multiple versions of the same image/tags into a tar ball at the same time.

fedora:rawhide
fedora:latest
fedora:f26
fedora:f25

If we decide this is a bad idea, we can drop support for it.   Currently kpod load is able to read the images created with multiple tags, not sure if it works correctly or not.

```
docker save --help

Usage:	docker save [OPTIONS] IMAGE [IMAGE...]

Save one or more images to a tar archive (streamed to STDOUT by default)

Options:
      --help            Print usage
  -o, --output string   Write to a file, instead of STDOUT
```

rhatdan: It is curious that the documentation for docker load seems to indicate one image.
```
 docker load --help

Usage:	docker load [OPTIONS]

Load an image from a tar archive or STDIN

Options:
      --help           Print usage
  -i, --input string   Read from tar archive file, instead of STDIN
  -q, --quiet          Suppress the load output
```

umohnani8: Will fix the comment.

mtrmac: > We are trying to match the docker save command, their use case is to put multiple versions of the same image/tags into a tar ball at the same time.

Ah, OK. The containers/image design does make that awkward; there is no concept of copying a group of images at once, and appending images one by one would involve creating temporary tar file copies, potentially gigabytes in size—or, as discussed above, files with duplicate file names.

> Currently kpod load is able to read the images created with multiple tags

The containers/image archive reader currently refuses to process an archive with multiple images; it may support an image with multiple tags pointing at the same image, but only because the tag is ignored. (Though implementing the ability to read any of the images by tag should be quite possible.)

> It is curious that the documentation for docker load seems to indicate one image.

`docker load` does not allow you to _choose_ an image; it always import every image, however many there are, and it always sets all the tags set in the tarball, whatever they are. (I.e. loading a tarball from an untrusted source is not a good idea, it can redirect the `rhel:latest` link for example.)

umohnani8: One more thing - if there is an image(for example alpine) with multiple tags in docker and you do <docker save -o all.tar alpine> it saves all the tags into the archive. I guess that is why save has the cherry-picking option where you can choose which tag to save to the archive. 
This is not supported in containers/image. When a tag is not specified it defaults to the "latest" tag even if there are multiple tags of the same image.

mtrmac: Right; containers/image only works one image at a time. There is no “list tags” (other than a single very ugly hack via `docker/Image.GetRepositoryTags`), there is no “copy several images”.

(IMHO the “copy all tags of an image ~ all versions of the application” functionality is not all that useful… but I have no idea what real-world customers do. The ability to copy several differently-named images into a single archive, sharing a single copy of the base layers, is more useful OTOH. Still, because the docker archives discard manifest digests, using a centralized registry server seems more generally useful than copying tar files around using… ssh?)

umohnani8: Hmm, so I think we need to decide if we want to support saving multiple tags to the same archive like docker does or just drop that functionality completely.
Also, kpod load is able to load all the tags in an archive with multiple tags as docker saves all the repotags to the manifest.json. Not sure if containers/image actually handles that correctly but it works.

umohnani8: Right, I meant it works for multiple tags of the same image. So if we save 
alpine:2.6
alpine:2.7
into the same docker archive and load that using kpod, it will load both alpine:2.6 and alpine:2.7

The error you are getting above is because the loadTarManifest was throwing an error if it received more than one object. I changed that in containers/image to not throw an error if there is more than one so that kpod load can load all the tags in one archive.
This was the pull request for that change that got accepted and merged https://github.com/containers/image/pull/309

mtrmac: (To be precise, what matters is that the _image_ (config+layers) is the same. The _names_ may differ, both in the tag and in the repo (`busybox:latest` and `example.com/personalproject/myspecialcopy:today`), as long as they point to the same contents.)

#309 only allows _listing_ the various tags; but actually copying the images out still fails with the quoted error (see also the warning above; the code does not allow choosing which image to read).  (Just clarifying the current implementation state. The “load” path is fixable easily enough—we insist on seekable input, so it’s quite possible to choose one of the images; it’s the “save” part that is problematic.)

---

I guess short-term the easiest way to get the “save multiple” case working would be to create a temporary directory, copy all of the images in there one at a time using the `oci:` transport, and then tar it up. That bypasses the need to edit tar archives. Downsides would be:

- Extra code happening outside of containers/image, arguably working around its inflexible model
- OCI does not currently support multiple images _either_ ( #306 ), though that should be trivially fixable

mtrmac: > I guess short-term the easiest way to get the “save multiple” case working

… for the OCI format, not for the docker-compatible tar one; but, if anything, we actually need the docker-compatible more, to preserve compatibility with existing workflows.

umohnani8: Okay, so after I changed loadTarManifest (#309) I was able to get the list of tags and it pulled all the tags in. It is possible that it pulled the same image multiple times and named it differently as the tags are different - basically loading multiple images does not fail for me. 
Right, getting the docker one working is more important to match their functionality.

rhatdan: I say if an image has multiple tags we should saver the tags, but not do anything with multiple images.

docker/archive/dest.go
```diff
@@ -19,22 +19,28 @@ func newImageDestination(ctx *types.SystemContext, ref archiveReference) (types.
...
+	fhStat, err := fh.Stat()
 	if err != nil {
-		// FIXME: It should be possible to modify archives, but the only really
-		//        sane way of doing it is to create a copy of the image, modify
-		//        it and then do a rename(2).
-		if os.IsExist(err) {
-			err = errors.New("docker-archive doesn't support modifying existing images")
-		}
-		return nil, err
+		return nil, errors.Wrapf(err, "error retrieving stat of file")
```
rhatdan: What file? 
```
 +		return nil, errors.Wrapf(err, "error statting file %q", ref.path) 
```

umohnani8: Fixed.

rhatdan:
@mtrmac PTAL, I think we are in agreement, and this is needed to complete kpod save.

docker/archive/dest.go
```diff
@@ -19,22 +19,28 @@ func newImageDestination(ctx *types.SystemContext, ref archiveReference) (types.
 	if ref.destinationRef == nil {
 		return nil, errors.Errorf("docker-archive: destination reference not supplied (must be of form <path>:<reference:tag>)")
 	}
-	fh, err := os.OpenFile(ref.path, os.O_WRONLY|os.O_EXCL|os.O_CREATE, 0644)
+
+	// the file needs to be created for write only
+	// if the file prexisted it needs to be empty
```
mtrmac: This comment still says “what” but not “why”.  Perhaps something like
> ref.path may be either a regular file or a pipe. In the pipe case we must be able to open a pre-existing file.
>
> OTOH for regular files we don’t want to overwrite a pre-existing file, if there is any,
> so we check for Size() == 0 below. (This is racy, but using O_EXCL would also be racy,
> only in a different way. Either way, it’s up to the user not to have two writers to the same path.)

docker/archive/dest.go
```diff
@@ -19,22 +19,28 @@ func newImageDestination(ctx *types.SystemContext, ref archiveReference) (types.
...
-		//        sane way of doing it is to create a copy of the image, modify
-		//        it and then do a rename(2).
-		if os.IsExist(err) {
-			err = errors.New("docker-archive doesn't support modifying existing images")
-		}
-		return nil, err
+		return nil, errors.Wrapf(err, "error statting file %q", ref.path)
+	}
+
+	if fhStat.Mode()&os.ModeNamedPipe != 0 || (fhStat.Mode().IsRegular() && fhStat.Size() == 0) {
```
mtrmac: Please use `(fhStat.Mode() & os.ModeType) == os.ModeNamedPipe`.  The code above _is_ correct (unlike UNIX, `os.FileMode` uses individual bits for the various non-regular-file file types), but unexpectedly tricky because the negation of the condition is not `… == 0` (that would also include regular files). 

(Also, as a matter of personal? style, `if` (the error case) `{` return error `}` /* normal processing */ would be more idiomatic.)

umohnani8: Fixed and negated the if to return the error instead.

docker/archive/dest.go
umohnani8: Changed the comment to this.

umohnani8:
@rhatdan PTAL

docker/archive/dest.go
```diff
@@ -19,15 +19,24 @@ func newImageDestination(ctx *types.SystemContext, ref archiveReference) (types.
 	if ref.destinationRef == nil {
 		return nil, errors.Errorf("docker-archive: destination reference not supplied (must be of form <path>:<reference:tag>)")
 	}
-	fh, err := os.OpenFile(ref.path, os.O_WRONLY|os.O_EXCL|os.O_CREATE, 0644)
+
+	// ref.path can be either a pipe or a regular file
+	// in the case of a pipe, we must be able to open a pre-existing file
```
rhatdan: +	// in the case of a pipe, we must be able to open a pre-existing file 
This line is wrong.
+     //in the case of a pipe, we require that we can open it for write

umohnani8: changed.

rhatdan:
LGTM
@mtrmac PTAL

docker/archive/dest.go
```diff
@@ -19,15 +19,24 @@ func newImageDestination(ctx *types.SystemContext, ref archiveReference) (types.
...
-		//        sane way of doing it is to create a copy of the image, modify
-		//        it and then do a rename(2).
-		if os.IsExist(err) {
-			err = errors.New("docker-archive doesn't support modifying existing images")
-		}
-		return nil, err
+		return nil, errors.Wrapf(err, "error statting file %q", ref.path)
+	}
+
+	if fhStat.Mode()&os.ModeType != os.ModeNamedPipe && (fhStat.Mode().IsRegular() && fhStat.Size() != 0) {
```
mtrmac: This is not actually a mathematical negation of the original, and `fhStat.Mode().IsRegular()` always implies `fhStat.Mode()&os.ModeType != os.ModeNamedPipe`.

We could either just ignore other file types, and say
```go
if fhStat.Mode().IsRegular() && fhStat.Size() != 0 {
   /* fail */
}
```
, or something like
```go
if fhStat.Mode().IsRegular() && fhStat.Size() != 0 {
  /* fail as above */
} else if fhStat.Mode()&os.ModeType != os.ModeNamedPipe {
  return nil, fmt.Errorf("Unsupported file type with mode %x", fhstat.Mode()")
}
```

Looking at the history of this PR, all we really wanted to check for is the non-zero size, so the shorter variant should be enough.

I guess I can live with the code as is, so :+1: in the interest of speeding things up. But I’d prefer this to be cleaned up—and we are waiting for @runcom anyway.

mtrmac:
:+1:

docker/archive/dest.go
umohnani8: Okay, changed it to the if - else if method.

mtrmac: This fails tests.

umohnani8: The else if was causing it to fail. I just took it out so it is like the shorter version. Hope that is fine.

mtrmac: Sure. Thanks!

rhatdan:
@runcom PTAL

runcom:
LGTM

PR #314, opened 2017-07-27
Add oci-archive transport that creates a tar archive of an image

umohnani8:
The oci-archive transport uses oci to create a directory of the image, which is then tarred up into a file. oci-archive will be used in kpod save and kpod load for oci images.

### DIFF
--- a/oci/archive/oci_dest.go
+++ b/oci/archive/oci_dest.go
@@ -1,0 +1,135 @@
+package archive
+
+import (
+	"io"
+	"os"
+
+	"github.com/containers/image/types"
+	"github.com/containers/storage/pkg/archive"
+	"github.com/pkg/errors"
+)
+
+type ociArchiveImageDestination struct {
+	ref          ociArchiveReference
+	unpackedDest types.ImageDestination
+	tempDirRef   tempDirOCIRef
+}
+
+// newImageDestination returns an ImageDestination for writing to an existing directory.
+func newImageDestination(ctx *types.SystemContext, ref ociArchiveReference) (types.ImageDestination, error) {
+	tempDirRef, err := createOCIRef(ref.image)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error creating oci reference")
+	}
+	unpackedDest, err := tempDirRef.ociRefExtracted.NewImageDestination(ctx)
+	if err != nil {
+		if err := tempDirRef.deleteTempDir(); err != nil {
+			return nil, errors.Wrapf(err, "error deleting temp directory", tempDirRef.tempDirectory)
+		}
+		return nil, err
+	}
+	return &ociArchiveImageDestination{ref: ref,
+		unpackedDest: unpackedDest,
+		tempDirRef:   tempDirRef}, nil
+}
+
+// Reference returns the reference used to set up this destination.
+func (d *ociArchiveImageDestination) Reference() types.ImageReference {
+	return d.ref
+}
+
+// Close removes resources associated with an initialized ImageDestination, if any
+// Close deletes the temp directory of the oci-archive image
+func (d *ociArchiveImageDestination) Close() error {
+	defer d.tempDirRef.deleteTempDir()
+	return d.unpackedDest.Close()
+}
+
+func (d *ociArchiveImageDestination) SupportedManifestMIMETypes() []string {
+	return d.unpackedDest.SupportedManifestMIMETypes()
+}
+
+// SupportsSignatures returns an error (to be displayed to the user) if the destination certainly can't store signatures
+func (d *ociArchiveImageDestination) SupportsSignatures() error {
+	return d.unpackedDest.SupportsSignatures()
+}
+
+// ShouldCompressLayers returns true iff it is desirable to compress layer blobs written to this destination
+func (d *ociArchiveImageDestination) ShouldCompressLayers() bool {
+	return d.unpackedDest.ShouldCompressLayers()
+}
+
+// AcceptsForeignLayerURLs returns false iff foreign layers in manifest should be actually
+// uploaded to the image destination, true otherwise.
+func (d *ociArchiveImageDestination) AcceptsForeignLayerURLs() bool {
+	return d.unpackedDest.AcceptsForeignLayerURLs()
+}
+
+// MustMatchRuntimeOS returns true iff the destination can store only images targeted for the current runtime OS. False otherwise
+func (d *ociArchiveImageDestination) MustMatchRuntimeOS() bool {
+	return d.unpackedDest.MustMatchRuntimeOS()
+}
+
+// PutBlob writes contents of stream and returns data representing the result (with all data filled in).
+// inputInfo.Digest can be optionally provided if known; it is not mandatory for the implementation to verify it.
+// inputInfo.Size is the expected length of stream, if known.
+func (d *ociArchiveImageDestination) PutBlob(stream io.Reader, inputInfo types.BlobInfo) (types.BlobInfo, error) {
+	return d.unpackedDest.PutBlob(stream, inputInfo)
+}
+
+// HasBlob returns true iff the image destination already contains a blob with the matching digest which can be reapplied using ReapplyBlob
+func (d *ociArchiveImageDestination) HasBlob(info types.BlobInfo) (bool, int64, error) {
+	return d.unpackedDest.HasBlob(info)
+}
+
+func (d *ociArchiveImageDestination) ReapplyBlob(info types.BlobInfo) (types.BlobInfo, error) {
+	return d.unpackedDest.ReapplyBlob(info)
+}
+
+// PutManifest writes manifest to the destination
+func (d *ociArchiveImageDestination) PutManifest(m []byte) error {
+	return d.unpackedDest.PutManifest(m)
+}
+
+func (d *ociArchiveImageDestination) PutSignatures(signatures [][]byte) error {
+	return d.unpackedDest.PutSignatures(signatures)
+}
+
+// Commit marks the process of storing the image as successful and asks for the image to be persisted
+// after the directory is made, it is tarred up into a file and the directory is deleted
+func (d *ociArchiveImageDestination) Commit() error {
+	if err := d.unpackedDest.Commit(); err != nil {
+		return errors.Wrapf(err, "error storing image %q", d.ref.image)
+	}
+
+	// path of directory to tar up
+	src := d.tempDirRef.tempDirectory
+	// path to save tarred up file
+	dst := d.ref.resolvedFile
+	if err := tarDirectory(src, dst); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// tar converts the directory at src and saves it to dst
+func tarDirectory(src, dst string) error {
+	// input is a stream of bytes from the archive of the directory at path
+	input, err := archive.Tar(src, archive.Uncompressed)
+	if err != nil {
+		return errors.Wrapf(err, "error retrieving stream of bytes from %q", src)
+	}
+
+	// creates the tar file
+	outFile, err := os.Create(dst)
+	if err != nil {
+		return errors.Wrapf(err, "error creating tar file %q", dst)
+	}
+	defer outFile.Close()
+
+	// copies the contents of the directory to the tar file
+	_, err = io.Copy(outFile, input)
+
+	return err
+}
--- a/oci/archive/oci_src.go
+++ b/oci/archive/oci_src.go
@@ -1,0 +1,88 @@
+package archive
+
+import (
+	"context"
+	"io"
+
+	ocilayout "github.com/containers/image/oci/layout"
+	"github.com/containers/image/types"
+	digest "github.com/opencontainers/go-digest"
+	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/pkg/errors"
+)
+
+type ociArchiveImageSource struct {
+	ref         ociArchiveReference
+	unpackedSrc types.ImageSource
+	tempDirRef  tempDirOCIRef
+}
+
+// newImageSource returns an ImageSource for reading from an existing directory.
+// newImageSource untars the file and saves it in a temp directory
+func newImageSource(ctx *types.SystemContext, ref ociArchiveReference, requestedManifestMIMETypes []string) (types.ImageSource, error) {
+	tempDirRef, err := createUntarTempDir(ref)
+	if err != nil {
+		return nil, errors.Wrap(err, "error creating temp directory")
+	}
+
+	unpackedSrc, err := tempDirRef.ociRefExtracted.NewImageSource(ctx, requestedManifestMIMETypes)
+	if err != nil {
+		if err := tempDirRef.deleteTempDir(); err != nil {
+			return nil, errors.Wrapf(err, "error deleting temp directory", tempDirRef.tempDirectory)
+		}
+		return nil, err
+	}
+	return &ociArchiveImageSource{ref: ref,
+		unpackedSrc: unpackedSrc,
+		tempDirRef:  tempDirRef}, nil
+}
+
+// LoadManifestDescriptor loads the manifest
+func LoadManifestDescriptor(imgRef types.ImageReference) (imgspecv1.Descriptor, error) {
+	ociArchRef, ok := imgRef.(ociArchiveReference)
+	if !ok {
+		return imgspecv1.Descriptor{}, errors.Errorf("error typecasting, need type ociArchiveReference")
+	}
+	tempDirRef, err := createUntarTempDir(ociArchRef)
+	if err != nil {
+		return imgspecv1.Descriptor{}, errors.Wrap(err, "error creating temp directory")
+	}
+	defer tempDirRef.deleteTempDir()
+
+	descriptor, err := ocilayout.LoadManifestDescriptor(tempDirRef.ociRefExtracted)
+	if err != nil {
+		return imgspecv1.Descriptor{}, errors.Wrap(err, "error loading index")
+	}
+	return descriptor, nil
+}
+
+// Reference returns the reference used to set up this source.
+func (s *ociArchiveImageSource) Reference() types.ImageReference {
+	return s.ref
+}
+
+// Close removes resources associated with an initialized ImageSource, if any.
+// Close deletes the temporary directory at dst
+func (s *ociArchiveImageSource) Close() error {
+	defer s.tempDirRef.deleteTempDir()
+	return s.unpackedSrc.Close()
+}
+
+// GetManifest returns the image's manifest along with its MIME type
+// (which may be empty when it can't be determined but the manifest is available).
+func (s *ociArchiveImageSource) GetManifest() ([]byte, string, error) {
+	return s.unpackedSrc.GetManifest()
+}
+
+func (s *ociArchiveImageSource) GetTargetManifest(digest digest.Digest) ([]byte, string, error) {
+	return s.unpackedSrc.GetTargetManifest(digest)
+}
+
+// GetBlob returns a stream for the specified blob, and the blob's size.
+func (s *ociArchiveImageSource) GetBlob(info types.BlobInfo) (io.ReadCloser, int64, error) {
+	return s.unpackedSrc.GetBlob(info)
+}
+
+func (s *ociArchiveImageSource) GetSignatures(c context.Context) ([][]byte, error) {
+	return s.unpackedSrc.GetSignatures(c)
+}
--- a/oci/archive/oci_transport.go
+++ b/oci/archive/oci_transport.go
@@ -1,0 +1,226 @@
+package archive
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+
+	"github.com/containers/image/directory/explicitfilepath"
+	"github.com/containers/image/docker/reference"
+	"github.com/containers/image/image"
+	ocilayout "github.com/containers/image/oci/layout"
+	"github.com/containers/image/transports"
+	"github.com/containers/image/types"
+	"github.com/containers/storage/pkg/archive"
+	"github.com/pkg/errors"
+)
+
+func init() {
+	transports.Register(Transport)
+}
+
+// Transport is an ImageTransport for OCI archive
+// it creates an oci-archive tar file by calling into the OCI transport
+// tarring the directory created by oci and deleting the directory
+var Transport = ociArchiveTransport{}
+
+type ociArchiveTransport struct{}
+
+// ociArchiveReference is an ImageReference for OCI Archive paths
+type ociArchiveReference struct {
+	file         string
+	resolvedFile string
+	image        string
+}
+
+func (t ociArchiveTransport) Name() string {
+	return "oci-archive"
+}
+
+// ParseReference converts a string, which should not start with the ImageTransport.Name prefix
+// into an ImageReference.
+func (t ociArchiveTransport) ParseReference(reference string) (types.ImageReference, error) {
+	return ParseReference(reference)
+}
+
+// ValidatePolicyConfigurationScope checks that scope is a valid name for a signature.PolicyTransportScopes keys
+func (t ociArchiveTransport) ValidatePolicyConfigurationScope(scope string) error {
+	var file string
+	sep := strings.SplitN(scope, ":", 2)
+	file = sep[0]
+
+	if len(sep) == 2 {
+		image := sep[1]
+		if !refRegexp.MatchString(image) {
+			return errors.Errorf("Invalid image %s", image)
+		}
+	}
+
+	if !strings.HasPrefix(file, "/") {
+		return errors.Errorf("Invalid scope %s: must be an absolute path", scope)
+	}
+	// Refuse also "/", otherwise "/" and "" would have the same semantics,
+	// and "" could be unexpectedly shadowed by the "/" entry.
+	// (Note: we do allow "/:someimage", a bit ridiculous but why refuse it?)
+	if scope == "/" {
+		return errors.New(`Invalid scope "/": Use the generic default scope ""`)
+	}
+	cleaned := filepath.Clean(file)
+	if cleaned != file {
+		return errors.Errorf(`Invalid scope %s: Uses non-canonical path format, perhaps try with path %s`, scope, cleaned)
+	}
+	return nil
+}
+
+// annotation spex from https://github.com/opencontainers/image-spec/blob/master/annotations.md#pre-defined-annotation-keys
+const (
+	separator = `(?:[-._:@+]|--)`
+	alphanum  = `(?:[A-Za-z0-9]+)`
+	component = `(?:` + alphanum + `(?:` + separator + alphanum + `)*)`
+)
+
+var refRegexp = regexp.MustCompile(`^` + component + `(?:/` + component + `)*$`)
+
+// ParseReference converts a string, which should not start with the ImageTransport.Name prefix, into an OCI ImageReference.
+func ParseReference(reference string) (types.ImageReference, error) {
+	var file, image string
+	sep := strings.SplitN(reference, ":", 2)
+	file = sep[0]
+
+	if len(sep) == 2 {
+		image = sep[1]
+	}
+	return NewReference(file, image)
+}
+
+// NewReference returns an OCI reference for a file and a image.
+func NewReference(file, image string) (types.ImageReference, error) {
+	resolved, err := explicitfilepath.ResolvePathToFullyExplicit(file)
+	if err != nil {
+		return nil, err
+	}
+	// This is necessary to prevent directory paths returned by PolicyConfigurationNamespaces
+	// from being ambiguous with values of PolicyConfigurationIdentity.
+	if strings.Contains(resolved, ":") {
+		return nil, errors.Errorf("Invalid OCI reference %s:%s: path %s contains a colon", file, image, resolved)
+	}
+	if len(image) > 0 && !refRegexp.MatchString(image) {
+		return nil, errors.Errorf("Invalid image %s", image)
+	}
+	return ociArchiveReference{file: file, resolvedFile: resolved, image: image}, nil
+}
+
+func (ref ociArchiveReference) Transport() types.ImageTransport {
+	return Transport
+}
+
+// StringWithinTransport returns a string representation of the reference, which MUST be such that
+// reference.Transport().ParseReference(reference.StringWithinTransport()) returns an equivalent reference.
+func (ref ociArchiveReference) StringWithinTransport() string {
+	return fmt.Sprintf("%s:%s", ref.file, ref.image)
+}
+
+// DockerReference returns a Docker reference associated with this reference
+func (ref ociArchiveReference) DockerReference() reference.Named {
+	return nil
+}
+
+// PolicyConfigurationIdentity returns a string representation of the reference, suitable for policy lookup.
+func (ref ociArchiveReference) PolicyConfigurationIdentity() string {
+	// NOTE: ref.image is not a part of the image identity, because "$dir:$someimage" and "$dir:" may mean the
+	// same image and the two can’t be statically disambiguated.  Using at least the repository directory is
+	// less granular but hopefully still useful.
+	return fmt.Sprintf("%s", ref.resolvedFile)
+}
+
+// PolicyConfigurationNamespaces returns a list of other policy configuration namespaces to search
+// for if explicit configuration for PolicyConfigurationIdentity() is not set
+func (ref ociArchiveReference) PolicyConfigurationNamespaces() []string {
+	res := []string{}
+	path := ref.resolvedFile
+	for {
+		lastSlash := strings.LastIndex(path, "/")
+		// Note that we do not include "/"; it is redundant with the default "" global default,
+		// and rejected by ociTransport.ValidatePolicyConfigurationScope above.
+		if lastSlash == -1 || path == "/" {
+			break
+		}
+		res = append(res, path)
+		path = path[:lastSlash]
+	}
+	return res
+}
+
+// NewImage returns a types.Image for this reference, possibly specialized for this ImageTransport.
+// The caller must call .Close() on the returned Image.
+func (ref ociArchiveReference) NewImage(ctx *types.SystemContext) (types.Image, error) {
+	src, err := newImageSource(ctx, ref, nil)
+	if err != nil {
+		return nil, err
+	}
+	return image.FromSource(src)
+}
+
+// NewImageSource returns a types.ImageSource for this reference,
+// asking the backend to use a manifest from requestedManifestMIMETypes if possible.
+// The caller must call .Close() on the returned ImageSource.
+func (ref ociArchiveReference) NewImageSource(ctx *types.SystemContext, requestedManifestMIMETypes []string) (types.ImageSource, error) {
+	return newImageSource(ctx, ref, requestedManifestMIMETypes)
+}
+
+// NewImageDestination returns a types.ImageDestination for this reference.
+// The caller must call .Close() on the returned ImageDestination.
+func (ref ociArchiveReference) NewImageDestination(ctx *types.SystemContext) (types.ImageDestination, error) {
+	return newImageDestination(ctx, ref)
+}
+
+// DeleteImage deletes the named image from the registry, if supported.
+func (ref ociArchiveReference) DeleteImage(ctx *types.SystemContext) error {
+	return errors.Errorf("Deleting images not implemented for oci: images")
+}
+
+// struct to store the ociReference and temporary directory returned by createOCIRef
+type tempDirOCIRef struct {
+	tempDirectory   string
+	ociRefExtracted types.ImageReference
+}
+
+// deletes the temporary directory created
+func (t *tempDirOCIRef) deleteTempDir() error {
+	return os.RemoveAll(t.tempDirectory)
+}
+
+// createOCIRef creates the oci reference of the image
+func createOCIRef(image string) (tempDirOCIRef, error) {
+	dir, err := ioutil.TempDir("/var/tmp", "oci")
+	if err != nil {
+		return tempDirOCIRef{}, errors.Wrapf(err, "error creating temp directory")
+	}
+	ociRef, err := ocilayout.NewReference(dir, image)
+	if err != nil {
+		return tempDirOCIRef{}, err
+	}
+
+	tempDirRef := tempDirOCIRef{tempDirectory: dir, ociRefExtracted: ociRef}
+	return tempDirRef, nil
+}
+
+// creates the temporary directory and copies the tarred content to it
+func createUntarTempDir(ref ociArchiveReference) (tempDirOCIRef, error) {
+	tempDirRef, err := createOCIRef(ref.image)
+	if err != nil {
+		return tempDirOCIRef{}, errors.Wrap(err, "error creating oci reference")
+	}
+	src := ref.resolvedFile
+	dst := tempDirRef.tempDirectory
+	if err := archive.UntarPath(src, dst); err != nil {
+		if err := tempDirRef.deleteTempDir(); err != nil {
+			return tempDirOCIRef{}, errors.Wrapf(err, "error deleting temp directory %q", tempDirRef.tempDirectory)
+		}
+		return tempDirOCIRef{}, errors.Wrapf(err, "error untarring file %q", tempDirRef.tempDirectory)
+	}
+	return tempDirRef, nil
+}
--- a/oci/layout/oci_dest.go
+++ b/oci/layout/oci_dest.go
@@ -180,6 +180,10 @@ func (d *ociImageDestination) PutManifest(m []byte) error {
 		return err
 	}
 
+	if d.ref.image == "" {
+		return errors.Errorf("cannot save image with empyt image.ref.name")
+	}
+
 	annotations := make(map[string]string)
 	annotations["org.opencontainers.image.ref.name"] = d.ref.image
 	desc.Annotations = annotations
--- a/transports/alltransports/alltransports.go
+++ b/transports/alltransports/alltransports.go
@@ -10,6 +10,7 @@ import (
 	_ "github.com/containers/image/docker"
 	_ "github.com/containers/image/docker/archive"
 	_ "github.com/containers/image/docker/daemon"
+	_ "github.com/containers/image/oci/archive"
 	_ "github.com/containers/image/oci/layout"
 	_ "github.com/containers/image/openshift"
 	// The ostree transport is registered by ostree*.go
--- a/oci/archive/oci_transport_test.go
+++ b/oci/archive/oci_transport_test.go
@@ -1,0 +1,292 @@
+package archive
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/containers/image/types"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestTransportName(t *testing.T) {
+	assert.Equal(t, "oci-archive", Transport.Name())
+}
+
+func TestTransportParseReference(t *testing.T) {
+	testParseReference(t, Transport.ParseReference)
+}
+
+func TestTransportValidatePolicyConfigurationScope(t *testing.T) {
+	for _, scope := range []string{
+		"/etc",
+		"/etc:notlatest",
+		"/this/does/not/exist",
+		"/this/does/not/exist:notlatest",
+		"/:strangecornercase",
+	} {
+		err := Transport.ValidatePolicyConfigurationScope(scope)
+		assert.NoError(t, err, scope)
+	}
+
+	for _, scope := range []string{
+		"relative/path",
+		"/",
+		"/double//slashes",
+		"/has/./dot",
+		"/has/dot/../dot",
+		"/trailing/slash/",
+		"/etc:invalid'image!value@",
+		"/etc_no_image:",
+		"/etc:multiple_:separators",
+	} {
+		err := Transport.ValidatePolicyConfigurationScope(scope)
+		assert.Error(t, err, scope)
+	}
+}
+
+func TestParseReference(t *testing.T) {
+	testParseReference(t, ParseReference)
+}
+
+// testParseReference is a test shared for Transport.ParseReference and ParseReference.
+func testParseReference(t *testing.T, fn func(string) (types.ImageReference, error)) {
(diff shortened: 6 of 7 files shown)


umohnani8:
@rhatdan @mtrmac PTAL

oci/layout/oci_archive_dest.go
```diff
@@ -0,0 +1,126 @@
...
+	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/pkg/errors"
+)
+
+type ociArchiveImageDestination struct {
+	ref   ociArchiveReference
+	index imgspecv1.Index
+}
+
+var ociImgDest types.ImageDestination
```
rhatdan: Can this be added to the struct above?

oci/layout/oci_archive_dest.go
```diff
@@ -0,0 +1,126 @@
...
+	if err := ociImgDest.Commit(); err != nil {
+		return errors.Wrapf(err, "error storing image %q", ociRef.(ociReference).tag)
+	}
+
+	// path of directory to tar up
+	path := ociRef.(ociReference).resolvedDir
+	if err := tar(path); err != nil {
+		return err
+	}
+
```
rhatdan: You need to remove the ociImg when you are done.

oci/layout/oci_archive_dest.go
```diff
@@ -0,0 +1,126 @@
...
+	return nil
+}
+
+// tar converts the direstory at path to a tar file and deletes the directory
+func tar(path string) error {
+	// temporary path of the tar file
+	tempPath := path + "1"
+	defer os.Rename(tempPath, path)
+	defer os.RemoveAll(path)
```
rhatdan: THis should happen in the caller, not here.

oci/layout/oci_archive_dest.go
```diff
@@ -0,0 +1,126 @@
...
+	path := ociRef.(ociReference).resolvedDir
+	if err := tar(path); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// tar converts the direstory at path to a tar file and deletes the directory
+func tar(path string) error {
```
rhatdan: This should take an src,dest and then don't do the Rename stuff.

oci/layout/oci_archive_src.go
```diff
@@ -0,0 +1,81 @@
...
+func (s *ociArchiveImageSource) GetBlob(info types.BlobInfo) (io.ReadCloser, int64, error) {
+	return ociImgSrc.GetBlob(info)
+}
+
+func (s *ociArchiveImageSource) GetSignatures() ([][]byte, error) {
+	return ociImgSrc.GetSignatures()
+}
+
+// unTar unpacks a tar file into a directory and deletes the tar file
+func unTar(path string) error {
```
rhatdan: Should have src,dest

oci/layout/oci_archive_dest.go
umohnani8: Why not leave it in tar? It tars up the directory, removes the directory and renames the tar file to what the directory was before.

oci/layout/oci_archive_dest.go
```diff
@@ -0,0 +1,126 @@
+package layout
```
mtrmac: Please use a separate `…/oci/archive` subpackage for all of this.

oci/layout/oci_archive_transport.go
```diff
@@ -0,0 +1,100 @@
...
+
+import (
+	"github.com/containers/image/docker/reference"
+	"github.com/containers/image/image"
+	"github.com/containers/image/transports"
+	"github.com/containers/image/types"
+)
+
+func init() {
+	transports.Register(ArchiveTransport)
```
mtrmac: The `types.ImageTransport` and `types.ImageReference` types, due to their relevance for signature enforcement, should have test coverage close to 100%.  Perhaps roughly base it on `docker/archive/transport_test.go`, or I might be able to help with that later.

oci/layout/oci_archive_transport.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// ArchiveTransport is an ImageTransport for OCI archive
+// it creates an oci-archive tar file by calling into the OCI transport
+// tarring the directory created by oci and deleting the directory
+var ArchiveTransport = ociArchiveTransport{}
```
mtrmac: With the `oci/archive` package, this should be named just `Transport`.

oci/layout/oci_archive_transport.go
```diff
@@ -0,0 +1,100 @@
...
+
+type ociArchiveTransport struct{}
+
+var ociTr = ociTransport{}
```
mtrmac: The `types.ImageTransport` implementations should generally be stateless singletons (containers-storage is the only exception so far). Just call `oci/layout.Transport.$method` directly if necessary.

oci/layout/oci_archive_transport.go
```diff
@@ -0,0 +1,100 @@
...
+var ociRef types.ImageReference
```
mtrmac: This can't be a global variable, that would make it impossible to use multiple `ociArchiveReference` objects simultaneously.  Looks like this should be a member of `ociArchiveReference`.

oci/layout/oci_archive_transport.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+func (t ociArchiveTransport) Name() string {
+	return "oci-archive"
+}
+
+// ParseReference converts a string, which should not start with the ImageTransport.Name prefix
+// into an ImageReference.
+func (t ociArchiveTransport) ParseReference(reference string) (types.ImageReference, error) {
+	ociRef, _ = ociTr.ParseReference(reference)
```
mtrmac: The error return value needs to be handled.

Though, overall, it is unexpected to take the user-specified _file_ path and hand it off to a code which expects a _directory_ path; the behavior is really undefined.

This `ParseReference` should probably be doing its own parsing (though the code can look pretty similar, or even use a shared helper for parsing the reference/tag part) into a file+tag pair; then, something else (`NewImageDestination`?) should create a _temporary_ directory with a random name, and create an `oci/layout.ociReference` reference from that (directory, tag) part.

oci/layout/oci_archive_dest.go
```diff
@@ -0,0 +1,126 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// tar converts the direstory at path to a tar file and deletes the directory
+func tar(path string) error {
+	// temporary path of the tar file
+	tempPath := path + "1"
```
mtrmac: Nothing says `path+"1"` is available for this function to use. It might contain users’ irreplaceable data, for all we know.

oci/layout/oci_archive_dest.go
mtrmac: Yeah; `newImageDestinationArchive` should create a temporary directory (grep for `temporaryDirectoryForBigFiles`, not in the default path), create an `oci/layout.ociReference` pointing to that reference, then the `Commit` implementation should just tar it up into the desired destination path, and `Close` delete the temporary directory.

oci/layout/oci_archive_dest.go
mtrmac: It would be somewhat nice if the code allowed replacing an older tar _file_ in the destination location atomically with a new image. It’s by no means required _now_, but it shouldn’t be impossible.

And even disregarding atomicity, it is at least somewhat plausible that the user has used exactly the same destination path in the past, and a file already exists there, so we can’t just create a directory in the same path, at least without removing the old contents. 

Overall, sharing the path for both the file and directory, though kinda a neat trick, is just not worth the hassle AFAICT.

oci/layout/oci_archive_src.go
```diff
@@ -0,0 +1,81 @@
...
+	"os"
+
+	"github.com/containers/image/types"
+	"github.com/containers/storage/pkg/archive"
+	digest "github.com/opencontainers/go-digest"
+	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/pkg/errors"
+)
+
+var ociImgSrc types.ImageSource
```
mtrmac: This needs to be a member of `ociArchiveImageSource`, not a global variable.

oci/layout/oci_archive_dest.go
```diff
@@ -0,0 +1,126 @@
...
+
+	"github.com/containers/image/types"
+	"github.com/containers/storage/pkg/archive"
+	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/pkg/errors"
+)
+
+type ociArchiveImageDestination struct {
+	ref   ociArchiveReference
+	index imgspecv1.Index
```
mtrmac: `index` seems unused.

oci/layout/oci_archive_dest.go
rhatdan: If I call into a function with a file, I don't expect to have the file modified or changed.

rhatdan: Also allows the caller to control the naming.  I would rather have temporary names.

oci/layout/oci_archive_src.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// Close removes resources associated with an initialized ImageSource, if any.
+// Close tars the untarred directory after the image has been retrieved from source
+func (s *ociArchiveImageSource) Close() error {
+	if err := ociImgSrc.Close(); err != nil {
+		return err
+	}
+	path := ociRef.(ociReference).resolvedDir
+	if err := tar(path); err != nil {
```
mtrmac: `ociArchiveImageSource` really must not modify the original file in any way, not even with the best intentions of replacing it later:

1. In the meantime others expecting to use the original file at that path can't do so.
1. The process may crash, leaving the original replaced by an unexpected directory forever.
1. Nothing guarantees that untar+tar will reproduce the original byte-for-byte exactly. 

Similarly to the discussion about `ociArchiveImageDestination`, `newImageSourceArchive` should create a temporary directory and extract the image there.

(Alternatively, it would be also possible to implement `ociArchiveImageSource` which does not need to extract the tarball at all, by just seeking around the original tarball, like `docker/tarfile/Source.openTarComponent` and callers do.  Then `oci/archive` and `oci/layout` could share the format implementation, differing only in the backend used to access file components. But that would only work with seekable inputs; extracting the tarball and calling the `layout` implementation directly is perfectly fine.)

oci/layout/oci_archive_transport.go
mtrmac: … but let’s wait with the tests until the design, and in particular the reference parsing, is designed and nailed down.

oci/layout/oci_archive_transport.go
mtrmac: … but see the discussion in `ImageSource`/`ImageDestination` implementations; it rather seems that the `ociReference` lifetime should be tied to the lifetimes of the temporary directories, i.e. only members of `ociArchiveImageDestination`/`ociArchiveImageSource`

oci/layout/oci_archive_src.go
```diff
@@ -0,0 +1,81 @@
...
+	digest "github.com/opencontainers/go-digest"
+	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/pkg/errors"
+)
+
+var ociImgSrc types.ImageSource
+
+type ociArchiveImageSource struct {
+	ref        ociArchiveReference
+	descriptor imgspecv1.Descriptor
```
mtrmac: `descriptor` seems unused.

oci/layout/oci_archive_dest.go
rhatdan: She will have to export the interface from OCI if she does.

mtrmac: Only as long as the code keeps relying on the internals of `ociReference` etc., which should not be necessary if the rest of the code converts to use temporary directories.

umohnani8: I started out by packaging this separately, but I was having a hard time initializing ociReference. Since all the functions of oci-archive call the functions in oci, I need to be able to set ociReference and use it to call the oci functions. I was using Get and Set functions but was failing to get it work. 
Things make more sense now, so I am going to work on separating it.

mtrmac: `oci/layout.NewReference` should be all that is necessary AFAICT.

Then, perhaps, some utility subpackage to minimize the amount of copy&pasted parsing/formatting code shared between `oci/layout` and `oci/archive`, `ValidatePolicyConfigurationScope` and the like — depending on the details of how the two end up.

umohnani8:
I separated <oci-archive> to a sub package called <archive>. Got rid of the global variables and added <ociImageDestination> and <ociImageSource> to the respective Archive structs. 
I create a temp directory in </var/tmp> called <oci-image> and pass this path to <ociReference> - <ociArchiveReference> stores the path given by the user so that the tar file is stored there at the end.
This temp directory is then deleted in Close().
@rhatdan @mtrmac PTAL.

oci/layout/oci_transport.go
```diff
@@ -88,6 +88,12 @@ type ociReference struct {
 	tag         string
 }
 
+// GetOciReference returns the dir, resolvedDir, and tag of ociReference
+// so that it is accessible in oci/archive
+func GetOciReference(t types.ImageReference) (string, string, string) {
+	return t.(ociReference).dir, t.(ociReference).resolvedDir, t.(ociReference).tag
```
mtrmac: This would just panic if the caller supplied a wrong type.  But overall, after the use in `archive.ParseReference` is removed, this won’t be needed: when the caller creates a reference for the temporary directory, it can just remember the directory path.  That wastes a tiny bit of memory but the code is easier to follow and clearly can’t panic.

oci/archive/oci_transport.go
```diff
@@ -0,0 +1,117 @@
...
+
+// ParseReference converts a string, which should not start with the ImageTransport.Name prefix
+// into an ImageReference.
+func (t ociArchiveTransport) ParseReference(reference string) (types.ImageReference, error) {
+	// ociArchRef stores the dir, resolvedDir, and tag of the original path given by the user
+	ociArchRef, err := layout.Transport.ParseReference(reference)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error parsing reference %q", reference)
+	}
+	dir, resolvedDir, tag := layout.GetOciReference(ociArchRef)
```
mtrmac: This is still passing a file path to a transport which expects a directory.

Please just copy&paste most of `oci/layout/oci_transport.go` into this file, more or less with `s/dir/file/g; s/layout/archive/g` instead of bending the semantics like this.  Or perhaps create a separate helper subpackage for the shared code, which is explicitly documented to use generic paths without knowing whether they are files or directories.

oci/archive/oci_transport.go
```diff
@@ -0,0 +1,117 @@
...
+
+	// tempRef is the path where the directory of the oci image created by oci is stored
+	// this directory is deleted after a tar file is created from it
+	tempRef := "/var/tmp/oci-image:" + tag
```
mtrmac: Nothing guarantees the `"oci-image:"+tag` file name is available; use `ioutil.TempDir` to get a guaranteed-unique-and-available name.

OTOH, `ParseReference` may be used by callers who will never be calling `NewImageSource`/`NewImageDestination` (e.g. when parsing `/etc/containers/policy.json`), so this method should not be modifying the filesystem at all; so it also can’t create the layout reference.  Both will need to happen later in `newImageSource`/`newImageDestination`.

Finally, please use the `temporaryDirectoryForBigFiles` constant the way other sites do. Eventually we’d like to have that directory configurable via `types.SystemContext`; for now, just using the same constant name makes it easy to find other cases.

oci/archive/oci_transport.go
```diff
@@ -0,0 +1,117 @@
...
+	// ociRef stores the dir, resolvedDir, and tag of the temporary path
+	ociRef, err := layout.Transport.ParseReference(tempRef)
```
mtrmac: There is `layout.NewReference` with separate `dir` and `tag` arguments; it isn’t necessary to format them into a string only for `layout.Transport.ParseReference` to split them again.

oci/archive/oci_transport.go
mtrmac: For now I’d prefer to just copy that code; that will be sufficient to get this PR merged from my point of view.

We can live with two duplicates of that code for a while I think (@runcom ?); cleaning that up can be a separate PR.

runcom: agree with @mtrmac to keep duplicated code for now

umohnani8:
ignore this new push, still working on it.

oci/archive/oci_dest.go
```diff
@@ -0,0 +1,128 @@
...
+		return errors.Wrapf(err, "error tarring up directory %q", src)
+	}
+
+	return nil
+}
+
+// tar converts the direstory at src and saves it to dst
+func tar(src, dst string) error {
+	// input is a stream of bytes from the archive of the directory at path
+	input, err := archive.Tar(src, 0)
```
umohnani8: @mtrmac could you explain this a bit more? What should it be if not 0?

mtrmac: The `archive.Uncompressed` typed constant. (The constant does the value 0, but it is typed and readable without checking the source code of the `archive` package)

umohnani8:
@mtrmac PTAL. I have created a function `createOciRef` that creates the temp directory and generates the ociReference from the directory's path. The functions that need ociReference will first check to see if it exists, if not `createOciRef` is called.

oci/archive/oci_dest.go
```diff
@@ -0,0 +1,137 @@
...
+
+type ociArchiveImageDestination struct {
+	ref          ociArchiveReference
+	unpackedDest types.ImageDestination
+}
+
+// newImageDestination returns an ImageDestination for writing to an existing directory.
+func newImageDestination(ref ociArchiveReference, ctx *types.SystemContext) (types.ImageDestination, error) {
+	if ref.ociRef == nil {
+		if err := createOciRef(&ref); err != nil {
```
mtrmac: Does this work with
```go
ref := ParseReference(…)
// possibly repeated:
dest := ref.NewImageDestination()
…
dest.Close() // deletes ref.ociRef.tempDirectory ?!
// possibly repeated:
src := ref.NewImageSource() // tempDirectory unavailable
…
src.Close() // deletes ref.ociRef.tempDirectory ?!
```

the `layout.Reference`, and `tempDirectory`, AFAICS needs to be a property of the …Source / …Destination, not of the …Reference, to match the temporary directory lifetimes which terminate at *`.Close`.

oci/archive/oci_transport.go
```diff
@@ -0,0 +1,179 @@
...
+
+// ParseReference converts a string, which should not start with the ImageTransport.Name prefix
+// into an ImageReference.
+func (t ociArchiveTransport) ParseReference(reference string) (types.ImageReference, error) {
+	return ParseReference(reference)
+}
+
+// ValidatePolicyConfigurationScope checks that scope is a valid name for a signature.PolicyTransportScopes keys
+func (t ociArchiveTransport) ValidatePolicyConfigurationScope(scope string) error {
+	return ocilayout.Transport.ValidatePolicyConfigurationScope(scope)
```
mtrmac: Still mixing the semantics… either copy&paste `layout.Transport.ValidatePolicyConfigurationScope`, or extract into a shared helper which is explicitly agnostic of dir/path (i.e. rename existing `layout.Transport.ValidatePolicyConfigurationScope` to something like `oci.ValidatePathPolicyConfigurationScope`, and then call it from both `{archive,layout}.Transport.ValidatePOlicyConfigurationScope`

oci/archive/oci_transport.go
```diff
@@ -0,0 +1,179 @@
...
+
+// StringWithinTransport returns a string representation of the reference, which MUST be such that
+// reference.Transport().ParseReference(reference.StringWithinTransport()) returns an equivalent reference.
+func (ref ociArchiveReference) StringWithinTransport() string {
+	if ref.ociRef == nil {
+		if err := createOciRef(&ref); err != nil {
+			logrus.Error("error creating oci reference", err)
+		}
+	}
+	return ref.ociRef.StringWithinTransport()
```
mtrmac: This would return a path to the temporary directory, not to the user-specified archive. And like `ParseReference`, `StringWithinTransport` should not be modifying the filesystem.

 Again, copy&paste or extract into an independent helper.

oci/archive/oci_transport.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+// PolicyConfigurationIdentity returns a string representation of the reference, suitable for policy lookup.
+func (ref ociArchiveReference) PolicyConfigurationIdentity() string {
+	if ref.ociRef == nil {
+		if err := createOciRef(&ref); err != nil {
+			logrus.Error("error creating oci reference", err)
+		}
+	}
+	return ref.ociRef.PolicyConfigurationIdentity()
```
mtrmac: Should not return a path to the temporary dir — and should not be modifying the FS here.

oci/archive/oci_transport.go
```diff
@@ -0,0 +1,179 @@
...
+
+// PolicyConfigurationNamespaces returns a list of other policy configuration namespaces to search
+// for if explicit configuration for PolicyConfigurationIdentity() is not set
+func (ref ociArchiveReference) PolicyConfigurationNamespaces() []string {
+	if ref.ociRef == nil {
+		if err := createOciRef(&ref); err != nil {
+			logrus.Error("error creating oci reference", err)
+		}
+	}
+	return ref.ociRef.PolicyConfigurationNamespaces()
```
mtrmac: Should not return a path to the temporary dir — and should not be modifying the FS here.

oci/archive/oci_transport.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+// DeleteImage deletes the named image from the registry, if supported.
+func (ref ociArchiveReference) DeleteImage(ctx *types.SystemContext) error {
+	if ref.ociRef == nil {
+		if err := createOciRef(&ref); err != nil {
+			logrus.Error("error creating oci reference", err)
+		}
+	}
+	return ref.ociRef.DeleteImage(ctx)
```
mtrmac: This would delete the just-created temporary directory, not the archive file.

Luckily `layout.*.DeleteImage` just fails; this should probably as well.  Really deleting a single reference means deleting that entry in the index, and perhaps the referenced manifest and blob, it should not delete the whole archive.

umohnani8:
@mtrmac PTAL

oci/archive/oci_transport.go
```diff
@@ -0,0 +1,198 @@
...
+// Transport is an ImageTransport for OCI archive
+// it creates an oci-archive tar file by calling into the OCI transport
+// tarring the directory created by oci and deleting the directory
+var Transport = ociArchiveTransport{}
+
+type ociArchiveTransport struct{}
+
+// ociArchiveReference is an ImageReference for OCI Archive paths
+type ociArchiveReference struct {
+	file         string
```
rhatdan: Do we still need both? file and resolved file?

umohnani8: yup, ref.file is being used in `StringWithinTransport`

umohnani8:
https://github.com/containers/image/pull/318 needs to be merged in first

umohnani8:
@mtrmac PTAL.

oci/archive/oci_src.go
```diff
@@ -0,0 +1,153 @@
...
+	"github.com/pkg/errors"
+)
+
+type ociArchiveImageSource struct {
+	ref         ociArchiveReference
+	unpackedSrc types.ImageSource
+	tempDir     string
+}
+
+var tempDirRef *tempDirOciRef
```
mtrmac: This can’t be a global variable; multiple `ociArchiveImageSource` instances can exist at the same time. The necessary data can probably live inside the `ociArchiveImage$something`, or maybe just as local variables.

oci/archive/oci_dest.go
```diff
@@ -0,0 +1,188 @@
...
+
+type ociArchiveImageDestination struct {
+	ref          ociArchiveReference
+	unpackedDest types.ImageDestination
+	tempDir      string
+}
+
+// newImageDestination returns an ImageDestination for writing to an existing directory.
+func newImageDestination(ref ociArchiveReference, ctx *types.SystemContext) (types.ImageDestination, error) {
+	tempDirOciRef, err := createOciRef(ref.image)
```
mtrmac: (This creates an one-time named structure, unpacks it in the same function, and then throws it away. Perhaps `createOciRef` could just return a tuple of three values (reference, path, error) instead of bothering to define and return a structure.

OTOH structures with named members are more explicit than tuples with positional results, perhaps it is worth it. Or if `deleteTempDir` became a method on the structure.)

oci/archive/oci_dest.go
```diff
@@ -0,0 +1,188 @@
...
+
+func (d *ociArchiveImageDestination) SupportedManifestMIMETypes() []string {
+	return d.unpackedDest.SupportedManifestMIMETypes()
+}
+
+// SupportsSignatures returns an error (to be displayed to the user) if the destination certainly can't store signatures
+func (d *ociArchiveImageDestination) SupportsSignatures() error {
+	err := d.unpackedDest.SupportsSignatures()
+	if err != nil {
+		if err := deleteTempDir(); err != nil {
```
mtrmac: `SupportsSignatures` returning an error is not a critical failure, no need to clean stuff up just because `SupportsSignatures` returns an error. Besides, `Close` will do the cleanup anyway (the caller is expected to ensure that `Close` is eventually called).

oci/archive/oci_dest.go
```diff
@@ -0,0 +1,188 @@
...
+	return d.unpackedDest.MustMatchRuntimeOS()
+}
+
+// PutBlob writes contents of stream and returns data representing the result (with all data filled in).
+// inputInfo.Digest can be optionally provided if known; it is not mandatory for the implementation to verify it.
+// inputInfo.Size is the expected length of stream, if known.
+func (d *ociArchiveImageDestination) PutBlob(stream io.Reader, inputInfo types.BlobInfo) (types.BlobInfo, error) {
+	valBlobInfo, err := d.unpackedDest.PutBlob(stream, inputInfo)
+	if err != nil {
+		if err := deleteTempDir(); err != nil {
```
mtrmac: Same here, the cleanup shouldn’t be necessary. And in other places in this file.

oci/archive/oci_dest.go
```diff
@@ -0,0 +1,188 @@
...
+			return errors.Wrapf(err, "error deleting temp directory", tempDirRef.tempDirectory)
+		}
+		return errors.Wrapf(err, "error storing image %q", d.ref.image)
+	}
+
+	// path of directory to tar up
+	src := d.tempDir
+	// path to save tarred up file
+	dst := d.ref.resolvedFile
+	if err := tar(src, dst); err != nil {
```
mtrmac: (This works just fine; I can’t shake a feeling that _maybe_ `tar` could be renamed to something that makes the semantics of its arguments explicit just by reading the call, and then the `src`/`dst` assignments and the comments could be unnecessary. But I can’t think of a good alternative name for `tar`; `tarDirectoryTo` feels awkward and insufficient. Perhaps someone smarter than me can find a better name; if not, this is quite readable as is.)

oci/archive/oci_dest.go
```diff
@@ -0,0 +1,188 @@
...
+		return err
+	}
+	return nil
+}
+
+// Commit marks the process of storing the image as successful and asks for the image to be persisted
+// after the directory is made, it is tarred up into a file and the directory is deleted
+func (d *ociArchiveImageDestination) Commit() error {
+	if err := d.unpackedDest.Commit(); err != nil {
+		if err := deleteTempDir(); err != nil {
```
mtrmac: Even here the clean up should not be necessary; even after a successful `Commit` the caller is still expected to call `Close`.

oci/archive/oci_src.go
```diff
@@ -0,0 +1,153 @@
+package archive
+
+import (
+	"context"
+	"io"
+
```
mtrmac: Nit: remove this blank line please.

oci/archive/oci_src.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+	return &ociArchiveImageSource{ref: ref,
+		unpackedSrc: unpackedSrc,
+		tempDir:     tempDirRef.tempDirectory}, nil
+}
+
+// LoadManifestDescriptor loads the manifest
+func LoadManifestDescriptor(imgRef types.ImageReference) (imgspecv1.Descriptor, error) {
+	if tempDirRef == nil {
+		if err := createUntarTempDir(imgRef.(ociArchiveReference)); err != nil {
```
mtrmac: Like in #318, this cast should be checked for failure.

oci/archive/oci_src.go
```diff
@@ -0,0 +1,153 @@
...
+// GetManifest returns the image's manifest along with its MIME type
+// (which may be empty when it can't be determined but the manifest is available).
+func (s *ociArchiveImageSource) GetManifest() ([]byte, string, error) {
+	return s.unpackedSrc.GetManifest()
+}
+
+func (s *ociArchiveImageSource) GetTargetManifest(digest digest.Digest) ([]byte, string, error) {
+	valByte, valString, err := s.unpackedSrc.GetTargetManifest(digest)
+	if err != nil {
+		if err := deleteTempDir(); err != nil {
```
mtrmac: Like in the destination, cleanup should only be necessary in `Close` AFAICS. Applies troughout.

oci/archive/oci_src.go
```diff
@@ -0,0 +1,153 @@
...
+			return nil, errors.Wrapf(err, "error deleting temp directory", tempDirRef.tempDirectory)
+		}
+		return nil, err
+	}
+	return valByte, nil
+}
+
+// unTar unpacks the tar file at src and saves it at dst
+func unTar(src, dst string) error {
```
mtrmac: (Non-blocking: Having a 6-line wrapper of a 2-line call, with a single caller, seems not worth the hassle; it could just as well be inlined into the caller.)

oci/archive/oci_src.go
```diff
@@ -0,0 +1,153 @@
...
+	if err := archive.UntarPath(src, dst); err != nil {
+		return errors.Wrapf(err, "error untarring file %q", src)
+	}
+	return nil
+}
+
+// deleteTemDir deletes the temporary directory created for the oci image
+func deleteTempDir() error {
```
mtrmac: (The caller will need to track which directory to delete. This could be a simple `path` parameter, or maybe this function could be a method on `tempDirOciRef` if you want to keep that structure.)

oci/archive/oci_src.go
```diff
@@ -0,0 +1,153 @@
...
+	if tempDirRef != nil {
+		return os.RemoveAll(tempDirRef.tempDirectory)
+	}
+	return nil
+}
+
+// creates the temporary directory and copies the tarred content to it
+func createUntarTempDir(ref ociArchiveReference) error {
+	var err error
+	tempDirRef, err = createOciRef(ref.image)
```
mtrmac: Non-blocking nit: `createOciRef` is not quite explicit about what the method modifying the filesystem, and `createUntarTempDir` with two verbs is a tiny bit difficult to follow (and untar what?). Maybe something like `[create]temp[orary]OCILayout[Dir]`, `extract[ed]OCILayout[FromArchive]`? I guess that’s verging towards excessive verbosity

oci/archive/oci_transport.go
```diff
@@ -0,0 +1,199 @@
+package archive
```
mtrmac: (Not reviewing the naming and the like. at all, I assume that will mirror the outcome of #318.)

oci/archive/oci_transport.go
```diff
@@ -0,0 +1,199 @@
...
+	return errors.Errorf("Deleting images not implemented for oci: images")
+}
+
+// struct to store the ociReference and temporary directory returned by createOciRef
+type tempDirOciRef struct {
+	tempDirectory string
+	ociRef        types.ImageReference
+}
+
+// createOciRef creates the oci reference of the image
```
mtrmac: This should say layout / extracted or something like that; `ociArchiveReference` already is an “OCI” reference, that’s not quite the point.

oci/archive/oci_transport.go
```diff
@@ -0,0 +1,199 @@
...
+func createOciRef(image string) (*tempDirOciRef, error) {
+	// dir is a temporary directory used to generate the OCI image
+	dir, err := ioutil.TempDir("/var/tmp", "oci")
+	if err != nil {
+		return nil, errors.Errorf("error creating file %v", err)
```
mtrmac: Nit: s/file/directory/, and either there should be a colon before `%v`, or use `errors.Wrapf`

oci/archive/oci_transport.go
```diff
@@ -0,0 +1,199 @@
...
+
+// struct to store the ociReference and temporary directory returned by createOciRef
+type tempDirOciRef struct {
+	tempDirectory string
+	ociRef        types.ImageReference
+}
+
+// createOciRef creates the oci reference of the image
+func createOciRef(image string) (*tempDirOciRef, error) {
+	// dir is a temporary directory used to generate the OCI image
```
mtrmac: Generate… or untar and read. Perhaps it’s simpler to delete the comment than to make it exactly correct :)

transports/alltransports/alltransports.go
```diff
@@ -10,6 +10,7 @@ import (
 	_ "github.com/containers/image/docker"
 	_ "github.com/containers/image/docker/archive"
 	_ "github.com/containers/image/docker/daemon"
+	_ "github.com/containers/image/oci/archive"
```
mtrmac: Can this have a smoke test with the others?

oci/archive/oci_dest.go
mtrmac: Throughout: For no _really good_ reason, it seems that the Go convention is to use all lowercase for an acronym at the start of an identifier, but all uppercase elsewhere. So `ociArchiveReference`, but `createOCIRef`.

oci/archive/oci_src.go
umohnani8: Fixed.

oci/archive/oci_src.go
umohnani8: Okay, I got rid of the global variable and added this to the `ociArchiveReference` struct. However, I ran into an issue that I have a temporary fix for.

`LoadManifestDescriptor` needs to be able to access the image name from the index.json for the `kpod load` command. Since `LoadManifestDescriptor` is called before `newImageSource` for `kpod load`, I tried to check if `tempDirRef` has a value in the instance of `ociArchiveReference` and setting it in one of the functions. However, two instances were being created and `ociArchiveReference` wouldn't retain its value.

I can't seem to figure out why the instance of `ociArchiveReference` is not the same - could be that `LoadManifestDescriptor` is being called from `libkpod/copy.go`. I have a temporary fix in place that just creates a temp directory and deletes it in `LoadManifestDescriptor`. 

@mtrmac I know this might not be the ideal way, but just wanted your feedback and advice on this issue.

oci/archive/oci_src.go
umohnani8: Fixed.

oci/archive/oci_src.go
umohnani8: Fixed. I have cleanups in `newImageDestination` and `newImageSource` because it is possible for a failure to occur there after the temp directory is created.

oci/archive/oci_src.go
umohnani8: Fixed.

oci/archive/oci_src.go
umohnani8: Added it as a method to the tempDirOciRef struct.

oci/archive/oci_src.go
umohnani8: Not sure what to rename them to :/

oci/archive/oci_dest.go
umohnani8: this doesn't happen anymore. Fixed

umohnani8: Fixed.

oci/archive/oci_dest.go
umohnani8: Renamed to `tarDirectory` - makes more sense I guess.

transports/alltransports/alltransports.go
umohnani8: working on the tests

oci/archive/oci_dest.go
```diff
@@ -0,0 +1,134 @@
...
+	return d.unpackedDest.Close()
+}
+
+func (d *ociArchiveImageDestination) SupportedManifestMIMETypes() []string {
+	return d.unpackedDest.SupportedManifestMIMETypes()
+}
+
+// SupportsSignatures returns an error (to be displayed to the user) if the destination certainly can't store signatures
+func (d *ociArchiveImageDestination) SupportsSignatures() error {
+	return nil
```
mtrmac: This should also forward to `d.unpackedDest` (which does not return `nil`!)

oci/archive/oci_src.go
```diff
@@ -0,0 +1,89 @@
...
+		return imgspecv1.Descriptor{}, errors.Errorf("error typecasting, need type ociArchiveReference")
+	}
+	if err := createUntarTempDir(&ociArchRef); err != nil {
+		return imgspecv1.Descriptor{}, errors.Wrap(err, "error creating temp directory")
+	}
+	defer ociArchRef.tempDirRef.deleteTempDir()
+
+	descriptor, err := ocilayout.LoadManifestDescriptor(ociArchRef.tempDirRef.ociRefExtracted)
+	if err != nil {
+		if err := ociArchRef.tempDirRef.deleteTempDir(); err != nil {
```
mtrmac: There already is a `defer ….deleteTempDir()` above; and deleting the same name from a shared `*/tmp` directory twice is not, strictly speaking, safe (the name could have been reused by an unrelated process in the meantime).

If the point is to not ignore the error, that can be done in the `defer` block, e.g. see how `copy.Image` does that—but only if it is really important to preserve the cleanup error.

In this particular case it seems that the `error loading index` error is more important than the `error deleting temporary directory`, at least the `error loading index` one shouldn't be thrown away; though notifying the user that the cleanup failed might also be useful, it is less important. (See also the discussion around https://github.com/containers/image/pull/236#discussion_r100569280 if you care.)

oci/layout/oci_dest.go
```diff
@@ -177,8 +177,12 @@ func (d *ociImageDestination) PutManifest(m []byte) error {
 		return err
 	}
 
+	if d.ref.image == "" {
+		return errors.Errorf("cannot save image with empyt image.ref.name")
```
mtrmac: (ignoring the `oci/layout` parts, to be reviewed in #318 ).

oci/archive/oci_src.go
mtrmac: >  However, two instances were being created and ociArchiveReference wouldn't retain its value.

Yeah,`ociArchiveReference` is defined to be a _value_ of that `struct` type, i.e. every value of type `ociArchiveReference` is an independent copy of the value, not a reference to a shared underlying object. I.e. every method with a receiver of `(ref ociArchiveReference)` gets a fresh copy, and every parameter with type `ociArchiveReference` (notably the `ref` parameters to `newImage{Source,Destination}`) gets a copy. You are probably seeing the copies made during the `newImage{Source,Destination}` parameter passing.

In general, `ImageReference` is mostly being treated as an immutable-after-construction object, so storing state in there is a bit tricky; though at least caching information in it for speed up (without changing semantics) should be fine.  The storage backend uses references that way.

OTOH attaching the lifetime of the temporary directories to `ImageReference` is, with the current structure, problematic, because multiple `ImageSource`/`ImageDestination` objects could be created from the same `ImageReference` over time (or even concurrently!), and there is no good place to delete the temporary directory: `Image{Source,Destination}.Close` would delete it after the first user is done, possibly breaking all other users, and there is no `ImageReference.Close`.

So, yeah, extracting the archive in `LoadManifestDescriptor`, and tearing down the extracted archive immediately in that function, and then extracting it again in `NewImageSource`, is the best fit with the interface structure. Arguably that’s just evidence of a badly designed interface which can’t deal with this reasonable use case… but the cost of making the `LoadManifestDescriptor+NewImageSource` sequence performant (forcing every user of `ImageReference` to treat them as stateful objects and to call `ref.Close` eventually, and dealing with multiple users and locking and reference counts in the `ImageReference` implementations) is fairly high.

Overall having `LoadManifest` extract the archive one more time seems like a reasonable compromise. Or, perhaps better, it could just seek inside the archive to get the manifest file, like `tarfile.Source.readTarComponent` does, without extracting gigabytes of data onto the disk; with an uncompressed archive this would be quite cheap. With a compressed archive we might still have to decompress gigabytes of data, but at least we wouldn’t have to write it to disk.

Perhaps we could also make a special-case exception, by documenting that a reference passed to `LoadManifestDescriptor` must be _always_, and _exactly once_, passed to `ref.NewImageSource()`, and implementing the pair of these calls as a special case, without dealing with the general case of multiple, possibly concurrent, callers to `ref.NewImage{Source,Destination}`. That would be quite ugly, but ugly only locally, without impacting the rest of the code, and high-performing. I guess I’d start with the extracting-twice solution (or streaming-and-extracting solution) first, and defer implementing this special case only if the performance improvement turned out to be necessary.

oci/archive/oci_dest.go
umohnani8: @mtrmac are you saying this should return `d.unpackedDest` also?

mtrmac: `return d.unpackedDest.SupportsSignatures()`

umohnani8: nvm, I think I got what you meant. Fixed.

oci/archive/oci_transport.go
```diff
@@ -0,0 +1,226 @@
...
+
+// struct to store the ociReference and temporary directory returned by createOCIRef
+type tempDirOciRef struct {
+	tempDirectory   string
+	ociRefExtracted types.ImageReference
+}
+
+// deletes the temporary directory created
+func (t *tempDirOciRef) deleteTempDir() error {
+	if t != nil {
```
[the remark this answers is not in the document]
umohnani8: Right, it can't really ever be nil when that function is called. Took the check out - fixed.

oci/archive/oci_src.go
umohnani8: Since the `error loading index` error is more important, I didn't do what `copy.Image` does. Just took out the second `deleteTempDir()` and left it as a defer.

umohnani8:
@mtrmac are you saying that the `tempDirOciRef` struct should be attached to `ociArchiveSource` and `ociArchiveDestination` instead of `ociArchiveReference`? 
If that is the case `LoadManifestDescriptor` takes in a `types.ImageReference`, which can be typecasted to `ociArchivereference`, so it made sense to me to attach it to `ociArchiveReference`.

mtrmac:
> @mtrmac are you saying that the `tempDirOciRef` struct should be attached to `ociArchiveSource` and `ociArchiveDestination` instead of `ociArchiveReference`? 

Yes, probably—if we can live with the performance impact.

> If that is the case `LoadManifestDescriptor` takes in a `types.ImageReference`, which can be typecasted to `ociArchivereference`, so it made sense to me to attach it to `ociArchiveReference`.

Yes, and extracting the archive only once would be nice; but `types.ImageReference` has no lifetime or sharing semantics, and the deletion of directories happens in `Image{Source,Destination}.Close`, so the creation should happen in the same object or something with a closely related lifetime.

Consider
```go
ref := archive.NewReference(…)
src1 := ref.NewImageSource()  // Extracts temporary directory, sets ref.tempDirRef
src2 := ref.NewImageSource() // Does this reuse the directory from src1, or create a new one?
// Either way, src1 and src2 now both use ref.tempDirRef
… // use src1, src2
src1.Close() // This deletes ref.tempDirRef; does that break src2’s use of it?
LoadManifestDescriptor(ref) // Similar concerns as newImageSource; does the cleanup break src2 again?
src2.Close()
```

Supporting something like this, while sharing the extracted directory, would require `ref` to keep a (thread-safe?) reference count, _and_ probably a `ImageReference.Close` to clean up, and be somewhat resilient against concurrent uses of `ImageSource`/`ImageDestination` on the same `ref`.  It’s not _impossible_ but it’s a fair amount of work.

Also, if you look at what the current code _actually_ does, `createOCIRef` creates a new temporary directory on each call, and `createUntarTempDir` extracts the archive on each call; neither is reused across uses of `ref`. So, the code is _already_ effectively giving a separate temporary directory to each `Image{Source,Destination}` instance; storing it in those instances directly instead of the shared `ImageReference` object would be both simpler and eliminate the problems with `.Close()` deleting a directory used by another user of the reference.

rhatdan:
Lets go with the slow mode for now, and then if we find it to be a performace issue, we can revisit this in the future.

umohnani8:
@mtrmac attached the `tempDirOCIRef` struct to `ociArchiveImageSource and ociArchiveImageDestination`. PTAL

oci/archive/oci_transport.go
```diff
@@ -0,0 +1,226 @@
...
+// asking the backend to use a manifest from requestedManifestMIMETypes if possible.
+// The caller must call .Close() on the returned ImageSource.
+func (ref ociArchiveReference) NewImageSource(ctx *types.SystemContext, requestedManifestMIMETypes []string) (types.ImageSource, error) {
+	return newImageSource(ref, ctx, requestedManifestMIMETypes)
+}
+
+// NewImageDestination returns a types.ImageDestination for this reference.
+// The caller must call .Close() on the returned ImageDestination.
+func (ref ociArchiveReference) NewImageDestination(ctx *types.SystemContext) (types.ImageDestination, error) {
+	return newImageDestination(ref, ctx)
```
mtrmac: Please keep the `ctx` parameter in the first position in `newImage{Source,Destination}`, to be consistent with other users.

transports/alltransports/alltransports.go
mtrmac: … this is also outstanding, not blocking I guess.

oci/archive/oci_transport.go
umohnani8: Fixed.

transports/alltransports/alltransports.go
umohnani8: I added tests for oci_transport, are there any other tests that I am missing?

mtrmac: `transports/alltransports/alltransports_test.go:TestImageNameHandling`

umohnani8: added the test. @mtrmac PTAL

mtrmac:
:+1:

Thanks!

PR #43, opened 2016-07-18
WIP types: use context.Context for public API

runcom:
This patch adds a `context.Context` argument to every public function we export in our public types.
Why?
We provide a library which makes http calls to remote services which can timeout indefinitely or more than how long the user of our library is willing to wait (or is expected to).
There's an important Go concurrency pattern which is in use in the http.Request object. Specifically the Request object provides a _Cancel_ struct which is a channel that whenever is closed the request/response is closed accordingly. So, a user can provide his own context and do his stuff in the caller and if needed close the underlying http resource at his will w/o leaking anything.
As a plus, context.Context is being moved to the standard library in Go 1.7 and it's already widely accepted to have it as the first arg of public functions (see Docker, Protobuf, Docker/distribution.. and many more)

More info here:
https://blog.golang.org/context
https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/

Signed-off-by: Antonio Murdaca runcom@redhat.com

PR #316, opened 2017-07-28
allow to use context for getting signatures

mfojtik:
@mtrmac @runcom since we plan to use the containers/image for downloading signatures from external location, we want to wrap every HTTP request to any location with `context.Context` that allows to set timeouts and deadlines for the request (so we won't open a request to a black hole, etc.)

i don't want to sweep all containers/image code and address all cases for now (but somebody should ;-) just plumb the code we currently need for https://github.com/openshift/origin/pull/15494

@smarterclayton FYI

### DIFF
--- a/directory/directory_test.go
+++ b/directory/directory_test.go
@@ -2,6 +2,7 @@ package directory
 
 import (
 	"bytes"
+	"context"
 	"io/ioutil"
 	"os"
 	"testing"
@@ -145,7 +146,7 @@ func TestGetPutSignatures(t *testing.T) {
 	src, err := ref.NewImageSource(nil, nil)
 	require.NoError(t, err)
 	defer src.Close()
-	sigs, err := src.GetSignatures()
+	sigs, err := src.GetSignatures(context.Background())
 	assert.NoError(t, err)
 	assert.Equal(t, signatures, sigs)
 }
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -1,6 +1,7 @@
 package docker
 
 import (
+	"context"
 	"crypto/tls"
 	"encoding/base64"
 	"encoding/json"
@@ -254,24 +255,25 @@ func newDockerClient(ctx *types.SystemContext, ref dockerReference, write bool, 
 
 // makeRequest creates and executes a http.Request with the specified parameters, adding authentication and TLS options for the Docker client.
 // The host name and schema is taken from the client or autodetected, and the path is relative to it, i.e. the path usually starts with /v2/.
-func (c *dockerClient) makeRequest(method, path string, headers map[string][]string, stream io.Reader) (*http.Response, error) {
-	if err := c.detectProperties(); err != nil {
+func (c *dockerClient) makeRequest(ctx context.Context, method, path string, headers map[string][]string, stream io.Reader) (*http.Response, error) {
+	if err := c.detectProperties(ctx); err != nil {
 		return nil, err
 	}
 
 	url := fmt.Sprintf("%s://%s%s", c.scheme, c.registry, path)
-	return c.makeRequestToResolvedURL(method, url, headers, stream, -1, true)
+	return c.makeRequestToResolvedURL(ctx, method, url, headers, stream, -1, true)
 }
 
 // makeRequestToResolvedURL creates and executes a http.Request with the specified parameters, adding authentication and TLS options for the Docker client.
 // streamLen, if not -1, specifies the length of the data expected on stream.
 // makeRequest should generally be preferred.
 // TODO(runcom): too many arguments here, use a struct
-func (c *dockerClient) makeRequestToResolvedURL(method, url string, headers map[string][]string, stream io.Reader, streamLen int64, sendAuth bool) (*http.Response, error) {
+func (c *dockerClient) makeRequestToResolvedURL(ctx context.Context, method, url string, headers map[string][]string, stream io.Reader, streamLen int64, sendAuth bool) (*http.Response, error) {
 	req, err := http.NewRequest(method, url, stream)
 	if err != nil {
 		return nil, err
 	}
+	req = req.WithContext(ctx)
 	if streamLen != -1 { // Do not blindly overwrite if streamLen == -1, http.NewRequest above can figure out the length of bytes.Reader and similar objects without us having to compute it.
 		req.ContentLength = streamLen
 	}
@@ -323,7 +325,7 @@ func (c *dockerClient) setupRequestAuth(req *http.Request) error {
 				}
 				service, _ := challenge.Parameters["service"] // Will be "" if not present
 				scope := fmt.Sprintf("repository:%s:%s", c.scope.remoteName, c.scope.actions)
-				token, err := c.getBearerToken(realm, service, scope)
+				token, err := c.getBearerToken(req.Context(), realm, service, scope)
 				if err != nil {
 					return err
 				}
@@ -340,11 +342,12 @@ func (c *dockerClient) setupRequestAuth(req *http.Request) error {
 	return nil
 }
 
-func (c *dockerClient) getBearerToken(realm, service, scope string) (*bearerToken, error) {
+func (c *dockerClient) getBearerToken(ctx context.Context, realm, service, scope string) (*bearerToken, error) {
 	authReq, err := http.NewRequest("GET", realm, nil)
 	if err != nil {
 		return nil, err
 	}
+	authReq = authReq.WithContext(ctx)
 	getParams := authReq.URL.Query()
 	if service != "" {
 		getParams.Add("service", service)
@@ -447,14 +450,14 @@ func getAuth(ctx *types.SystemContext, registry string) (string, string, error) 
 
 // detectProperties detects various properties of the registry.
 // See the dockerClient documentation for members which are affected by this.
-func (c *dockerClient) detectProperties() error {
+func (c *dockerClient) detectProperties(ctx context.Context) error {
 	if c.scheme != "" {
 		return nil
 	}
 
 	ping := func(scheme string) error {
 		url := fmt.Sprintf(resolvedPingV2URL, scheme, c.registry)
-		resp, err := c.makeRequestToResolvedURL("GET", url, nil, nil, -1, true)
+		resp, err := c.makeRequestToResolvedURL(ctx, "GET", url, nil, nil, -1, true)
 		logrus.Debugf("Ping %s err %#v", url, err)
 		if err != nil {
 			return err
@@ -481,7 +484,7 @@ func (c *dockerClient) detectProperties() error {
 		// best effort to understand if we're talking to a V1 registry
 		pingV1 := func(scheme string) bool {
 			url := fmt.Sprintf(resolvedPingV1URL, scheme, c.registry)
-			resp, err := c.makeRequestToResolvedURL("GET", url, nil, nil, -1, true)
+			resp, err := c.makeRequestToResolvedURL(ctx, "GET", url, nil, nil, -1, true)
 			logrus.Debugf("Ping %s err %#v", url, err)
 			if err != nil {
 				return false
@@ -506,9 +509,9 @@ func (c *dockerClient) detectProperties() error {
 
 // getExtensionsSignatures returns signatures from the X-Registry-Supports-Signatures API extension,
 // using the original data structures.
-func (c *dockerClient) getExtensionsSignatures(ref dockerReference, manifestDigest digest.Digest) (*extensionSignatureList, error) {
+func (c *dockerClient) getExtensionsSignatures(ctx context.Context, ref dockerReference, manifestDigest digest.Digest) (*extensionSignatureList, error) {
 	path := fmt.Sprintf(extensionsSignaturePath, reference.Path(ref.ref), manifestDigest)
-	res, err := c.makeRequest("GET", path, nil, nil)
+	res, err := c.makeRequest(ctx, "GET", path, nil, nil)
 	if err != nil {
 		return nil, err
 	}
--- a/docker/docker_image.go
+++ b/docker/docker_image.go
@@ -1,6 +1,7 @@
 package docker
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"net/http"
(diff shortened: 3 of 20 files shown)


mtrmac:
ACK, something like that needs to happen, but `context.Context` is weird to have in `SystemContext`, I’d prefer to do that sweep to changing the API later. I'll see what I can do.

You only care about `ImageSource` for now, is that correct?

mfojtik:
@mtrmac yes, that is correct. I can add this as new argument for `reference.NewImageSource` but I didn't want to break the API.

mtrmac:
(BTW if you are planning to carry this patch, there are more places that need to be patched: `makeRequestToResolvedURL` and `getBearerToken`.)

mtrmac:
> I can add this as new argument for reference.NewImageSource but I didn't want to break the API.

https://golang.org/pkg/context/ says
> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it.

so this would be an even bigger API break.

Oh well, we can do it, or perhaps add …`WithContext` methods to the interface —  but I guess I’d prefer the API break for now; we don’t have a documented ABI commitment (@runcom WDYT?).

(For reference, previous stab at this: https://github.com/containers/image/pull/43 )

runcom:
Right we don't have an API commitment yet so I'm fine adding CTX to the first argument of interfaces so we'll be more idiomatic. I had started doing this somewhere in another PR also but not sure where it went. 

I'm +1 here

mfojtik:
i'm fine with adding `.WithContext()` into interfaces

mfojtik:
(for now we only care about signatures, so I can plumb that, leaving you guys with the rest ;-)

mfojtik:
@mtrmac @runcom my last commit add `context.Context` to the codepath we need in origin to get the autoimport of signatures done. I left bunch of TODO's for methods that have to be plumbed.

PTAL

directory/directory_test.go
```diff
@@ -145,7 +145,7 @@ func TestGetPutSignatures(t *testing.T) {
 	src, err := ref.NewImageSource(nil, nil)
 	require.NoError(t, err)
 	defer src.Close()
-	sigs, err := src.GetSignatures()
+	sigs, err := src.GetSignatures(nil)
```
runcom: I believe we want to pass `context.TODO` here even if this is a test (or context.Background better)

docker/docker_client.go
```diff
@@ -254,24 +255,27 @@ func newDockerClient(ctx *types.SystemContext, ref dockerReference, write bool,
 
 // makeRequest creates and executes a http.Request with the specified parameters, adding authentication and TLS options for the Docker client.
 // The host name and schema is taken from the client or autodetected, and the path is relative to it, i.e. the path usually starts with /v2/.
-func (c *dockerClient) makeRequest(method, path string, headers map[string][]string, stream io.Reader) (*http.Response, error) {
+func (c *dockerClient) makeRequest(ctx context.Context, method, path string, headers map[string][]string, stream io.Reader) (*http.Response, error) {
 	if err := c.detectProperties(); err != nil {
```
runcom: This function also needs ctx

docker/docker_client.go
```diff
@@ -254,24 +255,27 @@ func newDockerClient(ctx *types.SystemContext, ref dockerReference, write bool,
...
 // streamLen, if not -1, specifies the length of the data expected on stream.
 // makeRequest should generally be preferred.
 // TODO(runcom): too many arguments here, use a struct
-func (c *dockerClient) makeRequestToResolvedURL(method, url string, headers map[string][]string, stream io.Reader, streamLen int64, sendAuth bool) (*http.Response, error) {
+func (c *dockerClient) makeRequestToResolvedURL(ctx context.Context, method, url string, headers map[string][]string, stream io.Reader, streamLen int64, sendAuth bool) (*http.Response, error) {
 	req, err := http.NewRequest(method, url, stream)
 	if err != nil {
 		return nil, err
 	}
+	if ctx != nil {
```
runcom: We should make sure context is either todo or background here so I'll drop this if.
req.WithContext can accept nil?

docker/docker_client.go
```diff
@@ -454,7 +458,8 @@ func (c *dockerClient) detectProperties() error {
 
 	ping := func(scheme string) error {
 		url := fmt.Sprintf(resolvedPingV2URL, scheme, c.registry)
-		resp, err := c.makeRequestToResolvedURL("GET", url, nil, nil, -1, true)
+		// FIXME: Pass the context.Context
+		resp, err := c.makeRequestToResolvedURL(nil, "GET", url, nil, nil, -1, true)
```
runcom: Use context.TODO or Background

docker/docker_image.go
```diff
@@ -41,7 +41,8 @@ func (i *Image) SourceRefFullName() string {
 // GetRepositoryTags list all tags available in the repository. Note that this has no connection with the tag(s) used for this specific image, if any.
 func (i *Image) GetRepositoryTags() ([]string, error) {
 	path := fmt.Sprintf(tagsPath, reference.Path(i.src.ref.ref))
-	res, err := i.src.c.makeRequest("GET", path, nil, nil)
+	// FIXME: Pass the context.Context
+	res, err := i.src.c.makeRequest(nil, "GET", path, nil, nil)
```
runcom: TODO or Background

docker/docker_client.go
mfojtik: req.WithContext can't accept nil unfortunatelly

mfojtik:
@runcom all fixed :)

docker/docker_client.go
runcom: Yes please :)

mfojtik:
@mtrmac @runcom why are the travis tests broken? ;-) doesn't seems this broken them

runcom:
@mfojtik you need to sign your commits (`git commit -s`)

mfojtik:
@runcom i'm sure they are signed :-)

runcom:
@mfojtik Verified means GPG signed, you need "Signed-by"

mfojtik:
@runcom the bot still does not like me :/

runcom:
seems like everything's ok now

runcom:
you just need to rebase this I guess

mfojtik:
@runcom already did :)

mfojtik:
ok weird... i must have the link to travis cached... all green now. guess this is ready for shipping?

runcom:
LGTM but we also need @mtrmac review :)

storage/storage_image.go
```diff
@@ -537,8 +538,9 @@ func (s *storageImageSource) GetTargetManifest(digest ddigest.Digest) (manifestB
 	return nil, "", ErrNoManifestLists
 }
 
-func (s *storageImageSource) GetSignatures() (signatures [][]byte, err error) {
+func (s *storageImageSource) GetSignatures(ctx context.Context) (signatures [][]byte, err error) {
 	var offset int
+	// FIXME: pass the context.Context
```
mtrmac: contaiers/storage is generally dealing with local files, ignoring the `ctx` here should be fine.

openshift/openshift.go
```diff
@@ -226,7 +227,7 @@ func (s *openshiftImageSource) GetBlob(info types.BlobInfo) (io.ReadCloser, int6
 	return s.docker.GetBlob(info)
 }
 
-func (s *openshiftImageSource) GetSignatures() ([][]byte, error) {
+func (s *openshiftImageSource) GetSignatures(ctx context.Context) ([][]byte, error) {
```
mtrmac: Could this one be implemented now?

I’m perfectly fine with having `context` support missing in most of the API, but when it exists, it would be better for users if it reliably worked.

docker/docker_client.go
mtrmac: Why should this accept `nil`? `nil` is an invalid `Context` value, and callers who pass it should just crash. This PR is already touching all callers, so it can ensure that `nil` is never used.

Or am I missing anything?

docker/docker_client.go
```diff
@@ -447,14 +453,15 @@ func getAuth(ctx *types.SystemContext, registry string) (string, string, error)
...
+func (c *dockerClient) detectProperties(ctx context.Context) error {
 	if c.scheme != "" {
 		return nil
 	}
 
 	ping := func(scheme string) error {
 		url := fmt.Sprintf(resolvedPingV2URL, scheme, c.registry)
-		resp, err := c.makeRequestToResolvedURL("GET", url, nil, nil, -1, true)
+		// FIXME: Pass the context.Context
+		resp, err := c.makeRequestToResolvedURL(ctx, "GET", url, nil, nil, -1, true)
```
mtrmac: The FIXME seems to already have been resolved.

docker/docker_client.go
```diff
@@ -254,24 +255,29 @@ func newDockerClient(ctx *types.SystemContext, ref dockerReference, write bool,
...
+func (c *dockerClient) makeRequestToResolvedURL(ctx context.Context, method, url string, headers map[string][]string, stream io.Reader, streamLen int64, sendAuth bool) (*http.Response, error) {
 	req, err := http.NewRequest(method, url, stream)
 	if err != nil {
 		return nil, err
 	}
+	if ctx != nil {
+		req = req.WithContext(ctx)
+	} else {
+		req = req.WithContext(context.Background())
+	}
```
mtrmac: Below this, `setupRequestAuth` and `getBearerToken` also need `Context` support.

docker/docker_client.go
```diff
@@ -481,7 +488,8 @@ func (c *dockerClient) detectProperties() error {
 		// best effort to understand if we're talking to a V1 registry
 		pingV1 := func(scheme string) bool {
 			url := fmt.Sprintf(resolvedPingV1URL, scheme, c.registry)
-			resp, err := c.makeRequestToResolvedURL("GET", url, nil, nil, -1, true)
+			// FIXME: Pass the context.Context
+			resp, err := c.makeRequestToResolvedURL(ctx, "GET", url, nil, nil, -1, true)
```
mtrmac: The FIXME seems to already have been resolved.

docker/docker_image_dest.go
```diff
@@ -132,7 +133,7 @@ func (d *dockerImageDestination) PutBlob(stream io.Reader, inputInfo types.BlobI
 	// FIXME? Chunked upload, progress reporting, etc.
 	uploadPath := fmt.Sprintf(blobUploadPath, reference.Path(d.ref.ref))
 	logrus.Debugf("Uploading %s", uploadPath)
-	res, err := d.c.makeRequest("POST", uploadPath, nil, nil)
+	res, err := d.c.makeRequest(context.Background(), "POST", uploadPath, nil, nil)
```
mtrmac: All of the `.Background()` in this file should be `.TODO()` AFAICT.

docker/docker_image_dest.go
```diff
@@ -396,7 +397,7 @@ func (d *dockerImageDestination) putSignaturesToAPIExtension(signatures [][]byte
 	// always adds signatures.  Eventually we should also allow removing signatures,
 	// but the X-Registry-Supports-Signatures API extension does not support that yet.
 
-	existingSignatures, err := d.c.getExtensionsSignatures(d.ref, d.manifestDigest)
+	existingSignatures, err := d.c.getExtensionsSignatures(nil, d.ref, d.manifestDigest)
```
mtrmac: `context.TODO()`, not `nil`, please.

docker/docker_image_src.go
```diff
@@ -192,23 +193,23 @@ func (s *dockerImageSource) GetBlob(info types.BlobInfo) (io.ReadCloser, int64,
...
 	default:
 		return [][]byte{}, nil
 	}
 }
 
 // getSignaturesFromLookaside implements GetSignatures() from the lookaside location configured in s.c.signatureBase,
 // which is not nil.
-func (s *dockerImageSource) getSignaturesFromLookaside() ([][]byte, error) {
+func (s *dockerImageSource) getSignaturesFromLookaside(ctx context.Context) ([][]byte, error) {
 	if err := s.ensureManifestIsLoaded(); err != nil {
```
mtrmac: `ensureManifestIsLoaded` also needs to handle `ctx`.

docker/docker_image_src.go
```diff
@@ -254,7 +255,14 @@ func (s *dockerImageSource) getOneSignature(url *url.URL) (signature []byte, mis
 
 	case "http", "https":
 		logrus.Debugf("GET %s", url)
-		res, err := s.c.client.Get(url.String())
+		req, err := http.NewRequest("GET", url.String(), nil)
+		if err != nil {
+			return nil, false, err
+		}
+		if ctx != nil {
```
mtrmac: Drop this conditional please; `ctx == nil` is invalid.

openshift/openshift.go
mfojtik: I think containers/image should switch to use the client-go from openshift as external dependency once we cut that out. In that case the context will be available.
This change was just to satisfy the interface.

mfojtik:
@mtrmac all comments addressed

TomSweeneyRedHat:
@mfojtik looks like a rebase is needed.

docker/docker_image_src.go
```diff
@@ -150,7 +151,7 @@ func (s *dockerImageSource) getExternalBlob(urls []string) (io.ReadCloser, int64
 		err  error
 	)
 	for _, url := range urls {
-		resp, err = s.c.makeRequestToResolvedURL("GET", url, nil, nil, -1, false)
+		resp, err = s.c.makeRequestToResolvedURL(context.Background(), "GET", url, nil, nil, -1, false)
```
mtrmac: `context.TODO()`, not `.Background()`, everywhere in this file

openshift/openshift.go
mtrmac: In the mean time we would still have unreliable API, that’s not great. I’ll take a stab at this.

(As for using the OpenShift client, that would be great (I extracting those thousand lines in `openshift-copies.go` was annoying and by now it’s probably pretty behind the original, and I do hate the duplication), but the two times this was attempted, it added 60? MB to the size of skopeo, which is just too much.)

mtrmac: @mfojtik added in https://github.com/mtrmac/image/tree/add-context-openshift , can you add that into this PR?

(Disclaimer: I’m still on Go 1.6, so this is untested. I’ll have to update ~immediately, I know.)

mfojtik: added your commit.

mfojtik:
@mtrmac updated, PTAL

TomSweeneyRedHat:
LGTM fwiw.

mtrmac:
:+1:

@mfojtik Thanks; one last update after #320 , please.

mfojtik:
@mtrmac updated, lets wait for tests to pass

mfojtik:
@runcom guess this one is now good for shipping

PR #317, opened 2017-07-29
Support v3 signature packets in mechanism_openpgp

mtrmac:
They are handled similarly to the more common V4 signature packets, except that `MessageDetails.SignatureV3` is set instead of `MessageDetails.Signature`, and they don’t support expiration.

Bulk of the commit is adding tests which use V3 signature packets.

I have regenerated `unknown-key.signature` because it was generated using a temporary key created just for that purpose, and the test is simpler when fixtures for both packet variants to use the same key.

Cc: @mfojtik


mfojtik:
LGTM

(this code is already part of OpenShift 3.6 :-)

runcom:
LGTM

TomSweeneyRedHat:
LGTM

mtrmac:
👍

PR #318, opened 2017-07-31
Fix oci to save the full image name in index.json

umohnani8:
Fix oci to save the full image name `image:tag` instead of just `tag` in index.json.
Add function to retrieve the image name from index.json when loading or pulling image from oci directory.

### DIFF
--- a/oci/layout/oci_dest.go
+++ b/oci/layout/oci_dest.go
@@ -23,13 +23,16 @@ type ociImageDestination struct {
 }
 
 // newImageDestination returns an ImageDestination for writing to an existing directory.
-func newImageDestination(ref ociReference) types.ImageDestination {
+func newImageDestination(ref ociReference) (types.ImageDestination, error) {
+	if ref.image == "" {
+		return nil, errors.Errorf("cannot save image with empty image.ref.name")
+	}
 	index := imgspecv1.Index{
 		Versioned: imgspec.Versioned{
 			SchemaVersion: 2,
 		},
 	}
-	return &ociImageDestination{ref: ref, index: index}
+	return &ociImageDestination{ref: ref, index: index}, nil
 }
 
 // Reference returns the reference used to set up this destination.  Note that this should directly correspond to user's intent,
@@ -178,7 +181,7 @@ func (d *ociImageDestination) PutManifest(m []byte) error {
 	}
 
 	annotations := make(map[string]string)
-	annotations["org.opencontainers.image.ref.name"] = d.ref.tag
+	annotations["org.opencontainers.image.ref.name"] = d.ref.image
 	desc.Annotations = annotations
 	desc.Platform = &imgspecv1.Platform{
 		Architecture: runtime.GOARCH,
--- a/oci/layout/oci_transport.go
+++ b/oci/layout/oci_transport.go
@@ -36,7 +36,14 @@ func (t ociTransport) ParseReference(reference string) (types.ImageReference, er
 	return ParseReference(reference)
 }
 
-var refRegexp = regexp.MustCompile(`^([A-Za-z0-9._-]+)+$`)
+// annotation spex from https://github.com/opencontainers/image-spec/blob/master/annotations.md#pre-defined-annotation-keys
+const (
+	separator = `(?:[-._:@+]|--)`
+	alphanum  = `(?:[A-Za-z0-9]+)`
+	component = `(?:` + alphanum + `(?:` + separator + alphanum + `)*)`
+)
+
+var refRegexp = regexp.MustCompile(`^` + component + `(?:/` + component + `)*$`)
 
 // ValidatePolicyConfigurationScope checks that scope is a valid name for a signature.PolicyTransportScopes keys
 // (i.e. a valid PolicyConfigurationIdentity() or PolicyConfigurationNamespaces() return value).
@@ -44,19 +51,14 @@ var refRegexp = regexp.MustCompile(`^([A-Za-z0-9._-]+)+$`)
 // scope passed to this function will not be "", that value is always allowed.
 func (t ociTransport) ValidatePolicyConfigurationScope(scope string) error {
 	var dir string
-	sep := strings.LastIndex(scope, ":")
-	if sep == -1 {
-		dir = scope
-	} else {
-		dir = scope[:sep]
-		tag := scope[sep+1:]
-		if !refRegexp.MatchString(tag) {
-			return errors.Errorf("Invalid tag %s", tag)
-		}
-	}
+	sep := strings.SplitN(scope, ":", 2)
+	dir = sep[0]
 
-	if strings.Contains(dir, ":") {
-		return errors.Errorf("Invalid OCI reference %s: path contains a colon", scope)
+	if len(sep) == 2 {
+		image := sep[1]
+		if !refRegexp.MatchString(image) {
+			return errors.Errorf("Invalid image %s", image)
+		}
 	}
 
 	if !strings.HasPrefix(dir, "/") {
@@ -64,7 +66,7 @@ func (t ociTransport) ValidatePolicyConfigurationScope(scope string) error {
 	}
 	// Refuse also "/", otherwise "/" and "" would have the same semantics,
 	// and "" could be unexpectedly shadowed by the "/" entry.
-	// (Note: we do allow "/:sometag", a bit ridiculous but why refuse it?)
+	// (Note: we do allow "/:someimage", a bit ridiculous but why refuse it?)
 	if scope == "/" {
 		return errors.New(`Invalid scope "/": Use the generic default scope ""`)
 	}
@@ -85,28 +87,26 @@ type ociReference struct {
 	// (But in general, we make no attempt to be completely safe against concurrent hostile filesystem modifications.)
 	dir         string // As specified by the user. May be relative, contain symlinks, etc.
 	resolvedDir string // Absolute path with no symlinks, at least at the time of its creation. Primarily used for policy namespaces.
-	tag         string
+	image       string // If image=="", it means the only image in the index.json is used
 }
 
 // ParseReference converts a string, which should not start with the ImageTransport.Name prefix, into an OCI ImageReference.
 func ParseReference(reference string) (types.ImageReference, error) {
-	var dir, tag string
-	sep := strings.LastIndex(reference, ":")
-	if sep == -1 {
-		dir = reference
-		tag = "latest"
-	} else {
-		dir = reference[:sep]
-		tag = reference[sep+1:]
+	var dir, image string
+	sep := strings.SplitN(reference, ":", 2)
+	dir = sep[0]
+
+	if len(sep) == 2 {
+		image = sep[1]
 	}
-	return NewReference(dir, tag)
+	return NewReference(dir, image)
 }
 
-// NewReference returns an OCI reference for a directory and a tag.
+// NewReference returns an OCI reference for a directory and a image.
 //
 // We do not expose an API supplying the resolvedDir; we could, but recomputing it
 // is generally cheap enough that we prefer being confident about the properties of resolvedDir.
-func NewReference(dir, tag string) (types.ImageReference, error) {
+func NewReference(dir, image string) (types.ImageReference, error) {
 	resolved, err := explicitfilepath.ResolvePathToFullyExplicit(dir)
 	if err != nil {
 		return nil, err
@@ -114,12 +114,12 @@ func NewReference(dir, tag string) (types.ImageReference, error) {
 	// This is necessary to prevent directory paths returned by PolicyConfigurationNamespaces
 	// from being ambiguous with values of PolicyConfigurationIdentity.
 	if strings.Contains(resolved, ":") {
-		return nil, errors.Errorf("Invalid OCI reference %s:%s: path %s contains a colon", dir, tag, resolved)
+		return nil, errors.Errorf("Invalid OCI reference %s:%s: path %s contains a colon", dir, image, resolved)
 	}
-	if !refRegexp.MatchString(tag) {
-		return nil, errors.Errorf("Invalid tag %s", tag)
+	if len(image) > 0 && !refRegexp.MatchString(image) {
+		return nil, errors.Errorf("Invalid image %s", image)
 	}
-	return ociReference{dir: dir, resolvedDir: resolved, tag: tag}, nil
+	return ociReference{dir: dir, resolvedDir: resolved, image: image}, nil
 }
 
 func (ref ociReference) Transport() types.ImageTransport {
@@ -132,7 +132,7 @@ func (ref ociReference) Transport() types.ImageTransport {
 // e.g. default attribute values omitted by the user may be filled in in the return value, or vice versa.
 // WARNING: Do not use the return value in the UI to describe an image, it does not contain the Transport().Name() prefix.
 func (ref ociReference) StringWithinTransport() string {
-	return fmt.Sprintf("%s:%s", ref.dir, ref.tag)
+	return fmt.Sprintf("%s:%s", ref.dir, ref.image)
 }
 
 // DockerReference returns a Docker reference associated with this reference
@@ -150,7 +150,10 @@ func (ref ociReference) DockerReference() reference.Named {
 // not required/guaranteed that it will be a valid input to Transport().ParseReference().
 // Returns "" if configuration identities for these references are not supported.
 func (ref ociReference) PolicyConfigurationIdentity() string {
-	return fmt.Sprintf("%s:%s", ref.resolvedDir, ref.tag)
+	// NOTE: ref.image is not a part of the image identity, because "$dir:$someimage" and "$dir:" may mean the
+	// same image and the two can’t be statically disambiguated.  Using at least the repository directory is
+	// less granular but hopefully still useful.
+	return fmt.Sprintf("%s", ref.resolvedDir)
 }
 
 // PolicyConfigurationNamespaces returns a list of other policy configuration namespaces to search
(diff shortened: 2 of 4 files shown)


runcom:
I don't like this, we should have a way to have users specify additional annotations. That annotation is specified in the OCI image-spec here https://github.com/opencontainers/image-spec/blob/710038243d857231f17df1c3f4c10850154bd1f7/image-layout.md#indexjson-file (and later paragraph).
That stuff in the spec isn't a must of course, but I want to leave it that way (it's just a tag)

> Implementor's Note: A common use case of descriptors with a "org.opencontainers.image.ref.name" annotation is representing a "tag" for a container image. For example, an image may have a tag for different versions or builds of the software. In the wild you often see "tags" like "v1.0.0-vendor.0", "2.0.0-debug", etc. Those tags will often be represented in an image-layout repository with matching "org.opencontainers.image.ref.name" annotations like "v1.0.0-vendor.0", "2.0.0-debug", etc.

Again, we should have a way to save additional annotations instead and save the full name iif needed.

mtrmac:
@runcom OTOH see https://github.com/opencontainers/image-spec/blob/master/annotations.md explicitly documenting a specific grammar.

As for additional annotations, dunno; there’s an inherent tension between c/image being general and treating all transports the same, and exposing everything there is to each transport. It’s not at all obvious to me how to generalize OCI annotations to anything else.

umohnani8:
@runcom we need the full image name when doing kpod load or kpod pull so that it is saved with the actual image name in containers-storage. Which Annotation would you suggest I use to store the full image name because "org.opencontainers.image.ref.name" seems the best one as the full image name is used to get the reference of the image.

runcom:
LGTM since nothing in the spec prevents this https://github.com/opencontainers/image-spec/blob/master/annotations.md#pre-defined-annotation-keys

also, as @mtrmac pointed out, abstracting OCI annotation is something not-trivial right now so let's defer that.

runcom:
@umohnani8 you need to fix tests failures though

umohnani8:
@runcom ok thanks, working on the test failures

umohnani8:
@mtrmac PTAL

umohnani8:
@mtrmac PTAL

oci/layout/oci_transport.go
```diff
@@ -196,26 +198,44 @@ func (ref ociReference) getManifestDescriptor() (imgspecv1.Descriptor, error) {
...
-		return imgspecv1.Descriptor{}, fmt.Errorf("no descriptor found for reference %q", ref.tag)
+		return imgspecv1.Descriptor{}, fmt.Errorf("no descriptor found for reference %q", ref.image)
 	}
 	return *d, nil
 }
 
+// LoadManifestDescriptor loads the manifest descriptor to be used to retrieve the image name
+// when pulling an image
+func LoadManifestDescriptor(imgRef types.ImageReference) (imgspecv1.Descriptor, error) {
+	return imgRef.(ociReference).getManifestDescriptor()
```
mtrmac: So this exists _only_ to get the name from _one-item_ indexes? Or is this intended to be similar to `docker.tarfile.Source.LoadTarManifest`, which returns all of the images?

The forced `.(ociReference)` cast here is a bit ugly… at least it should be a checked cast, returning an error when the caller uses a different kind of `ImageReference`; but perhaps the functionality could instead be sliced differently, with the underlying implementation getting a `(path string, image string)`, and `getManifestDescriptor()` extracting the two strings from an `ociReference`, and then (if the goal is really to only read one-image indexes) a `LoadManifestDescriptor(path string)` just setting `image = ""`.

oci/layout/oci_transport_test.go
```diff
@@ -38,9 +38,9 @@ func TestTransportValidatePolicyConfigurationScope(t *testing.T) {
 		"/has/./dot",
 		"/has/dot/../dot",
 		"/trailing/slash/",
-		"/etc:invalid'tag!value@",
```
mtrmac: (Note: first the semantics need to be decided, the comments on the tests may become irrelevant if the semantics were to be changed. And yes, I am rather paranoid about test coverage of the naming — it is relevant to signature verification.)

The list of valid cases above this one should include a name with multiple colons, and a name with some slashes.

oci/layout/oci_transport_test.go
```diff
@@ -64,48 +64,45 @@ func testParseReference(t *testing.T, fn func(string) (types.ImageReference, err
...
-	_, err = NewReference(tmpDir+"/thisparentdoesnotexist/something", tagValue)
+	_, err = NewReference(tmpDir+"/thisparentdoesnotexist/something", imageValue)
 	assert.Error(t, err)
 
-	_, err = NewReference(tmpDir+"/has:colon", tagValue)
+	_, err = NewReference(tmpDir+"/has:colon", imageValue)
 	assert.Error(t, err)
 
-	_, err = NewReference(tmpDir, "invalid'tag!value@")
+	_, err = NewReference(tmpDir, "invalid'image!value@")
```
mtrmac: This also should have a test for the `image == ""` case.

oci/layout/oci_transport_test.go
```diff
@@ -38,9 +38,9 @@ func TestTransportValidatePolicyConfigurationScope(t *testing.T) {
 		"/has/./dot",
 		"/has/dot/../dot",
 		"/trailing/slash/",
-		"/etc:invalid'tag!value@",
+		"/etc:invalid'image!value@",
 		"/path:with/colons",
-		"/path:with/colons/and:tag",
+		"/path:with/colons/and:image",
```
mtrmac: Also should have a test for a trailing colon (i.e. an explicitly empty $image).

oci/layout/oci_transport.go
```diff
@@ -36,22 +36,21 @@ func (t ociTransport) ParseReference(reference string) (types.ImageReference, er
 	return ParseReference(reference)
 }
 
-var refRegexp = regexp.MustCompile(`^([A-Za-z0-9._-]+)+$`)
+var refRegexp = regexp.MustCompile(`^([A-Za-z0-9:._-]+)+$`)
```
mtrmac: Do we want to explicitly mirror the spec at https://github.com/opencontainers/image-spec/blob/master/annotations.md#pre-defined-annotation-keys here?
(@runcom?)

runcom: Yes please, let's follow it.

oci/layout/oci_transport.go
```diff
@@ -85,41 +84,44 @@ type ociReference struct {
...
+	var dir, image string
+	sep := strings.SplitN(reference, ":", 2)
+	dir = sep[0]
+
+	if len(sep) == 2 {
+		if strings.Contains(sep[1], ":") {
+			image = sep[1]
+		} else {
+			image = sep[1] + ":latest"
+		}
```
runcom: Yes, we must keep 1 as well

oci/layout/oci_dest.go
```diff
@@ -178,7 +178,7 @@ func (d *ociImageDestination) PutManifest(m []byte) error {
 	}
 
 	annotations := make(map[string]string)
-	annotations["org.opencontainers.image.ref.name"] = d.ref.tag
+	annotations["org.opencontainers.image.ref.name"] = d.ref.image
```
runcom: I'd say error out if name is empty, spec forbids it also

oci/layout/oci_transport.go
```diff
@@ -150,7 +152,7 @@ func (ref ociReference) DockerReference() reference.Named {
 // not required/guaranteed that it will be a valid input to Transport().ParseReference().
 // Returns "" if configuration identities for these references are not supported.
 func (ref ociReference) PolicyConfigurationIdentity() string {
-	return fmt.Sprintf("%s:%s", ref.resolvedDir, ref.tag)
+	return fmt.Sprintf("%s:%s", ref.resolvedDir, ref.image)
```
runcom: Agree to use 3 as it seems the less-ugly case

oci/layout/oci_dest.go
mtrmac: >  spec forbids it also

Where? I can’t see it; in fact the index specification says that all of `annotations` is optional.

oci/layout/oci_transport.go
umohnani8: Fixed to use the spec.

oci/layout/oci_transport.go
```diff
@@ -85,41 +84,44 @@ type ociReference struct {
 	// (But in general, we make no attempt to be completely safe against concurrent hostile filesystem modifications.)
 	dir         string // As specified by the user. May be relative, contain symlinks, etc.
 	resolvedDir string // Absolute path with no symlinks, at least at the time of its creation. Primarily used for policy namespaces.
-	tag         string
+	image       string
```
umohnani8: Documented it.

oci/layout/oci_transport.go
umohnani8: Not sure what you mean here, should it not add the `latest` tag when a tag is not given for the image?

oci/layout/oci_transport.go
umohnani8: Fixed, returning only ref.resolvedDir.

umohnani8:
@mtrmac @runcom I made some changes based on your feedback, PTAL. Working on the tests and waiting for some clarification on the other feedback.

oci/layout/oci_transport.go
mtrmac: Sorry, you’re right this is unclear. Yes, that’s what I mean by 1: treat the image name as a string which as no further structure, unlike `docker/reference` where, in the extreme, `busybox` is a shorthand for `docker.io/library/busybox:latest`.

We’ve been treating the ref.name annotation as a Docker-like “tag”, and those default to `latest`; but now there’s the actual spec-by-regexp, which says nothing about existence or default values of tags.

(Not adding the `:latest` also keeps the behavior more, though not completely, compatible with the behavior before this PR.)

[Or is there pre-existing behavior, perhaps in umoci, we should co-opt here? I haven’t checked.]

oci/layout/oci_transport.go
mtrmac: OK; then `ValidatePolicyConfigurationScope` should also accept only directory values, not dir`:`image (but enforce that the directories don’t contain `:`, so that they can actually match `resolvedDir` values) and `PolicyConfigurationNamespaces` should, to be strictly correct, return only parent directories of `resolvedDir`, not `resolvedDir` itself (though, admittedly, technically, returning `resolvedDir` here as well does not actually hurt anything IIRC).

oci/layout/oci_transport.go
mtrmac: No need to use reflection here:
```go
ociRef, ok := imgRef.(ociReference)
if !ok { fail }
return ociRef.getManifestDescriptor()
```

oci/layout/oci_transport.go
owtaylor: It seems to me that the OCI transport should be able to read and write images with any ref-name allowed by the specification from an image-layout. Doesn't appending :latest here break that?

If compatibility with what the oci transport used to do is desired, the old behavior where oci:<path> was treated as oci:<path>:latest entirely compatible with the intended use case of this patch. In the end second component of the scope isn't a tag or an image name, it's simply the value of the org.opencontainers.image.ref.name annotation - if you don't want to get 'latest' stuck in there, just pass in your own value.

But if you want to be forward looking, and consider trying to handle image layouts with no ref-names at all (you could havedifferent images distinguished by platform like a Docker manifestlist) then I'd think it would be better to break compat and not introduce 'latest' at all.

oci/layout/oci_transport.go
```diff
@@ -36,22 +37,28 @@ func (t ociTransport) ParseReference(reference string) (types.ImageReference, er
...
-	sep := strings.LastIndex(scope, ":")
-	if sep == -1 {
-		dir = scope
-	} else {
-		dir = scope[:sep]
-		tag := scope[sep+1:]
-		if !refRegexp.MatchString(tag) {
-			return errors.Errorf("Invalid tag %s", tag)
+	sep := strings.SplitN(scope, ":", 2)
+	dir = sep[0]
```
owtaylor: Note that with the change to `Split(N)`, the code that follows

       if strings.Contains(dir, ":") {
               return errors.Errorf("Invalid OCI reference %s: path contains a colon", scope)
       }

Will never be hit - 'dir' cannot contain a ':' - a policy configuration scope of '/foo:bar:baz' is treated as '/foo' with a tag of 'bar:baz'.

owtaylor:
See also the comment about paths with ':' I left at https://github.com/containers/image/pull/323#issuecomment-321977838 in response to a comment by @cyphar - it's as relevant to this PR as that one.

oci/layout/oci_dest.go
cyphar: In the EBNF:

```
ref       ::= component ("/" component)*
component ::= alphanum (separator alphanum)*
alphanum  ::= [A-Za-z0-9]+
separator ::= [-._:@+] | "--"
```

You need to have at least one component which needs to have at least one alphanum.

oci/layout/oci_transport.go
umohnani8: @mtrmac I get it now. The reason I was adding `latest` to the image name if a tag was not given was because of the way kpod and docker do things. For example, if a user downloads the `alpine:latest` image from a registry and decides to save to oci on disk using the command `kpod save --oci -o alp.tar alpine`, the image name that will be saved in index.json based on your comment will be `alpine`. 

Therefore, when `alp.tar` is loaded back into containers-storage using the `kpod load` command, it will be stored as `alpine` instead of `alpine:latest`. I feel like this alters some form of the identity of the image as there would be no tag associated with it.

If we want to give the user the freedom of setting the image name to whatever they want it to be, we need to make it very clear in the documentation since all the kpod commands allow the user to access images with `latest` tag by just passing in the name.

owtaylor: @umohnani8 what kpod does is, of course, separate from what the containers/image code does. Despite the name of this pull request, I don't think containers/image should be assuming that the ref-name in the OCI is "a full image name" - it's just a string following certain rules. The behavior that the string is the "full image name" is being added by kpod, so if you want the behavior that `kpod save --oci -o alp.tar alpine` stores a ref-name of` `alpine:latest`, then kpod would be the logical place to do that. To me, the command seems to be saying "look up the name 'alpine' in storage and save that to alp.tar" and I'd expect the :latest to be added during the lookup process, not during the saving.

umohnani8: @owtaylor agreed, just realised that this was already being handled elsewhere - my bad. Will take out the image name modification from OCI.

oci/layout/oci_transport.go
umohnani8: Right, fixed.

umohnani8:
Fixing skopeo tests https://github.com/projectatomic/skopeo/pull/399

umohnani8:
@mtrmac PTAL

oci/layout/oci_dest.go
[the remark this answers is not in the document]
mtrmac: That says that the annotation, _if it is present_, must not be empty; but the annotation can AFAICS legally be completely omitted. The complete dictionary can be missing:

> **`annotations`** *string-string map*
> 
>     This OPTIONAL property contains arbitrary metadata for the image index.

and even if the `annotations` dictionary exists, I can’t see anything requiring it to contain the `org.opencontainers.image.ref.name` key. _Only if_ the dictionary and the `ref.name` key exists, does the EBNF require a non-empty name.

mtrmac:
> See also the comment about paths with ':' I left at #323 (comment) in response to a comment by @cyphar - it's as relevant to this PR as that one.

Yeah, could more people take a look at that? (Notably, `:` is a required character in `C:\Users\Guest\`-like paths on Windows.)

Arguably we’ve been using `:`, and forbidding `:` in paths, before this PR, and this PR does not really change anything about that, so it should not be blocked on that decision… but _eventually_ we need to decide whether to handle that, and if so, how.

oci/layout/oci_transport_test.go
```diff
@@ -64,48 +64,53 @@ func testParseReference(t *testing.T, fn func(string) (types.ImageReference, err
 		"relativepath",
 		tmpDir + "/thisdoesnotexist",
 	} {
-		for _, tag := range []struct{ suffix, tag string }{
-			{":notlatest", "notlatest"},
-			{"", "latest"},
+		for _, image := range []struct{ suffix, image string }{
+			{":notlatest:image", "notlatest:image"},
+			{":latestimage", "latestimage"},
+			{":", ""},
```
mtrmac: Please add a `{"", ""},` case here

oci/layout/oci_transport_test.go
```diff
@@ -64,48 +64,53 @@ func testParseReference(t *testing.T, fn func(string) (types.ImageReference, err
...
-	_, err = NewReference(tmpDir+"/thisparentdoesnotexist/something", tagValue)
-	assert.Error(t, err)
+	ref, err = NewReference(tmpDir, noImageValue)
+	require.NoError(t, err)
+	ociRef, ok = ref.(ociReference)
+	require.True(t, ok)
+	assert.Equal(t, tmpDir, ociRef.dir)
+	assert.Equal(t, noImageValue, ociRef.image)
 
-	_, err = NewReference(tmpDir+"/has:colon", tagValue)
```
mtrmac: (The `has:colon`-in-path test should be reinstated after `NewReference` again starts refusing such paths.)

oci/layout/oci_transport_test.go
```diff
@@ -151,8 +156,9 @@ func TestReferenceStringWithinTransport(t *testing.T) {
 	defer os.RemoveAll(tmpDir)
 
 	for _, c := range []struct{ input, result string }{
-		{"/dir1:notlatest", "/dir1:notlatest"}, // Explicit tag
-		{"/dir2", "/dir2:latest"},              // Default tag
+		{"/dir1:notlatest:notlatest", "/dir1:notlatest:notlatest"}, // Explicit image
+		{"/dir2:image", "/dir2:image"},                             // Default image
```
mtrmac: There is actually nothing “default” about this `:image` now that `:latest` is not treated specially; this line can be deleted.

oci/layout/oci_transport.go
```diff
@@ -196,26 +192,48 @@ func (ref ociReference) getManifestDescriptor() (imgspecv1.Descriptor, error) {
...
 	if d == nil {
-		return imgspecv1.Descriptor{}, fmt.Errorf("no descriptor found for reference %q", ref.tag)
+		return imgspecv1.Descriptor{}, fmt.Errorf("no descriptor found for reference %q", ref.image)
 	}
 	return *d, nil
 }
 
+// LoadManifestDescriptor loads the manifest descriptor to be used to retrieve the image name
+// when pulling an image
+func LoadManifestDescriptor(imgRef types.ImageReference) (imgspecv1.Descriptor, error) {
```
mtrmac: It would be nice to have tests for this (both with a known image name and for the `""` case), but it is not relevant for signing, so absolutely non-blocking for me.

oci/layout/oci_dest.go
```diff
@@ -177,8 +177,12 @@ func (d *ociImageDestination) PutManifest(m []byte) error {
 		return err
 	}
 
+	if d.ref.image == "" {
```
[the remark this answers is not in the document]
umohnani8: Are you saying that this check should be in either `NewImageDestination` or in `newImageDestination`?

oci/layout/oci_transport.go
```diff
@@ -85,41 +87,35 @@ type ociReference struct {
...
+func NewReference(dir, image string) (types.ImageReference, error) {
 	resolved, err := explicitfilepath.ResolvePathToFullyExplicit(dir)
 	if err != nil {
 		return nil, err
 	}
-	// This is necessary to prevent directory paths returned by PolicyConfigurationNamespaces
-	// from being ambiguous with values of PolicyConfigurationIdentity.
-	if strings.Contains(resolved, ":") {
-		return nil, errors.Errorf("Invalid OCI reference %s:%s: path %s contains a colon", dir, tag, resolved)
-	}
```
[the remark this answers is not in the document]
umohnani8: added the check back.

oci/layout/oci_transport.go
```diff
@@ -150,7 +146,7 @@ func (ref ociReference) DockerReference() reference.Named {
 // not required/guaranteed that it will be a valid input to Transport().ParseReference().
 // Returns "" if configuration identities for these references are not supported.
 func (ref ociReference) PolicyConfigurationIdentity() string {
-	return fmt.Sprintf("%s:%s", ref.resolvedDir, ref.tag)
+	return fmt.Sprintf("%s", ref.resolvedDir)
```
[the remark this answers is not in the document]
umohnani8: added the comment.

oci/layout/oci_transport_test.go
umohnani8: added this test case

oci/layout/oci_transport_test.go
umohnani8: added test back

oci/layout/oci_transport_test.go
umohnani8: removed default line

oci/layout/oci_dest.go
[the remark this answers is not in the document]
mtrmac: Yes; by that time we already know that the upload will fail, so we can skip copying blobs if we report this to the user early.

(Both `ref.NewImageDestination` and `newImageDestination` work; `newImageDestination` is a bit better because it is more difficult to unintentionally bypass.)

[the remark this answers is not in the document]
umohnani8: Put the check in `newImageDestinatioin`, fixed.

umohnani8:
@mtrmac PTAL

mtrmac:
:+1:

mtrmac:
@umohnani8 Please update #399 to use the master branch again, per https://github.com/projectatomic/skopeo/blob/master/README.md#dependencies-management .

umohnani8:
@mtrmac done!

PR #320, opened 2017-08-01
Don’t load the manifest in GetSignatures if the digest is known

mtrmac:
If we are supplied a digested reference, use it instead of possibly fetching the manifest.  This allows using signatures which are not protected even if the manifest is, in some situations.

To ensure consistent behavior, only do this if the reference uses the `digest.Canonical` algorithm, which has been used so far.  This could be perhaps relaxed in the future.

Note that this does not promise that GetSignatures will never need authentication—e.g. for `X-Registry-Supports-Signatures` this already depends on the server, and for the sigstore lookaside authentication
support may be added in the future.

Cc: @mfojtik

### DIFF
--- a/docker/docker_image_src.go
+++ b/docker/docker_image_src.go
@@ -206,13 +206,24 @@ func (s *dockerImageSource) GetSignatures() ([][]byte, error) {
 	}
 }
 
+// manifestDigest returns a digest of the manifest, either from the supplied reference or from a fetched manifest.
+func (s *dockerImageSource) manifestDigest() (digest.Digest, error) {
+	if digested, ok := s.ref.ref.(reference.Digested); ok {
+		d := digested.Digest()
+		if d.Algorithm() == digest.Canonical {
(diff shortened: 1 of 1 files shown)


mfojtik:
LGTM, thanks @mtrmac 

This means that using digest as part of the image pull spec will pickup that digest without manifest retrieval, right?
Also to make sure, the manifest has to be provided in order to verify the signature still right?

mtrmac:
> This means that using digest as part of the image pull spec will pickup that digest without manifest retrieval, right?

Yes, AFAICS.

> Also to make sure, the manifest has to be provided in order to verify the signature still right?

Yes.

mfojtik:
@runcom lgty? need this and context to complete the signature import for 3.6.1

runcom:
LGTM

runcom:
@mtrmac feel free to LGTM yourself and merge

mtrmac:
:+1:

PR #321, opened 2017-08-01
 Deprecate atomic transport 

rhatdan:
This will indicate if the interface should be used any longer.  The
patch removes the transport for the list of transports.

### DIFF
--- a/transports/transports.go
+++ b/transports/transports.go
@@ -71,13 +71,19 @@ func ImageName(ref types.ImageReference) string {
 	return ref.Transport().Name() + ":" + ref.StringWithinTransport()
 }
 
-// ListNames returns a list of transport names
+// ListNames returns a list of non deprecated transport names.
+// Deprecated transports can be used, but are not presented to users.
 func ListNames() []string {
 	kt.mu.Lock()
 	defer kt.mu.Unlock()
+	deprecated := map[string]bool{
+		"atomic": true,
+	}
 	var names []string
 	for _, transport := range kt.transports {
-		names = append(names, transport.Name())
+		if !deprecated[transport.Name()] {
+			names = append(names, transport.Name())
+		}
 	}
 	sort.Strings(names)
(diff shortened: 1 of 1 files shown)


transports/transports.go
```diff
@@ -77,7 +77,9 @@ func ListNames() []string {
 	defer kt.mu.Unlock()
 	var names []string
 	for _, transport := range kt.transports {
-		names = append(names, transport.Name())
+		if !transport.Deprecated() {
```
runcom: Can be actually depreciating transports in this package directly by filtering them out here? Cluttering the interface with a deprecated method doesn't sound right to me. @mtrmac what do you think? Transport is an interface and depreciating it in the library means just documenting that it's deprecated. Users of the interface (this package I'm commenting on) should filter what they think it's deprecated.

rhatdan: Bottom line is to get "atomic" out of the list of all transports, while still continuing to support it.
If you want to add a constant that lists deprecated transports and then to see if the transport.Name() is in the list, I am fine with that.

mtrmac: It does feel a bit weird for a transport to declare _itself_ deprecated (deprecated for what purpose? etc.); though with the `transports` package intentionally not knowing all other transports, the alternatives are also ugly.

I guess I would mildly prefer to keep this information inside the `transports` package (I guess a set of strings as transport names?), not in the individual transports themselves.

---

Separately, the `ListNames` method’s documentation needs to be updated to say that it is not exhaustive / does not show deprecated transports, and that the purpose is to offer recommended alternatives in an UI.  [Either way, a human needs to read the transport’s documentation to figure out what name to use within the transport, so this is unsuitable for automated processing.]

rhatdan:
@runcom @mtrmac PTAL

runcom:
LGTM

mtrmac:
:+1:

PR #322, opened 2017-08-02
Add registries.conf file and build library to parse it

baude:
registries.conf is a system-wide configuration file initially
developed in https://github.com/projectatomic/registries.  We
eventually want to ship the configuration file with this project;
hence the addition.  However, we also needed a library like
approach that gave this and other projects the ability to
parse the file in the same manner.

Signed-off-by: baude <bbaude@redhat.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,7 +66,7 @@ EPOCH_TEST_COMMIT ?= e68e0e1110e64f906f9b482e548f17d73e02e6b1
 .gitvalidation:
 	@which git-validation > /dev/null 2>/dev/null || (echo "ERROR: git-validation not found. Consider 'make clean && make tools'" && false)
 ifeq ($(TRAVIS),true)
-	@git-validation -q -run DCO,short-subject,dangling-whitespace
+	git-validation -q -run DCO,short-subject,dangling-whitespace
 else
-	@git-validation -q -run DCO,short-subject,dangling-whitespace -range $(EPOCH_TEST_COMMIT)..HEAD
+	git-validation -q -run DCO,short-subject,dangling-whitespace -range $(EPOCH_TEST_COMMIT)..HEAD
 endif
--- a/docs/registries.conf.5.md
+++ b/docs/registries.conf.5.md
@@ -1,0 +1,41 @@
+% registries.conf(5) System-wide registry configuration file
+% Brent Baude
+% Aug 2017
+
+# NAME
+registries.conf - Syntax of System Registry Configuration File
+
+# DESCRIPTION
+The REGISTRIES configuration file is a system-wide configuration file for container image
+registries. The file format is TOML.
+
+# FORMAT
+The TOML_format is used to build simple list format for registries under two
+categories: `search` and `insecure`. You can list multiple registries using
+as a comma separated list.
+
+Search registries are used when the caller of a container runtime does not fully specify the
+container image that they want to execute.  These registries are prepended onto the front
+ of the specified container image until the named image is found at a registry.
+
+Insecure Registries.  By default container runtimes use TLS when retrieving images
+from a registry.  If the registry is not setup with TLS, then the container runtime
+will fail to pull images from the registry. If you add the registry to the list of
+insecure registries then the container runtime will attempt use standard web protocols to
+pull the image.  It also allows you to pull from a registry with self-signed certificates.
+Note insecure registries can be used for any registry, not just the
+registries listed under search.
+
+The following example configuration defines two searchable registries and one
+insecure registry.
+
+```
+[registries.search]
+registries = ["registry1.com", "registry2.com"]
+
+[registries.insecure]
+registries = ["registry3.com"]
+```
+
+# HISTORY
+Aug 2017, Originally compiled by Brent Baude <bbaude@redhat.com>
--- a/pkg/sysregistries/system_registries.go
+++ b/pkg/sysregistries/system_registries.go
@@ -1,0 +1,76 @@
+package sysregistries
+
+import (
+	"github.com/BurntSushi/toml"
+	"io/ioutil"
+)
+
+// systemRegistriesConfPath is the path to the system-wide registry configuration file
+// and is used to add/subtract potential registries for obtaining images.
+// You can override this at build time with
+// -ldflags '-X github.com/containers/image/sysregistries.systemRegistriesConfPath=$your_path'
+var systemRegistriesConfPath = builtinRegistriesConfPath
+
+// builtinRegistriesConfPath is the path to registry configuration file
+// DO NOT change this, instead see systemRegistriesConfPath above.
+const builtinRegistriesConfPath = "/etc/containers/registries.conf"
+
+type registries struct {
+	Registries []string `toml:"registries"`
+}
+
+type tomlConfig struct {
+	Registries struct {
+		Search   registries `toml:"search"`
+		Insecure registries `toml:"insecure"`
+		Block    registries `toml:"block"`
+	} `toml:"registries"`
+}
+
+// Reads the global registry file from the filesystem. Returns
+// a byte array
+func readRegistryConf() ([]byte, error) {
+	configBytes, err := ioutil.ReadFile(systemRegistriesConfPath)
+	return configBytes, err
+}
+
+// For mocking in unittests
+var readConf = readRegistryConf
+
+// Loads the registry configuration file from the filesystem and
+// then unmarshals it.  Returns the unmarshalled object.
+func loadRegistryConf() (*tomlConfig, error) {
+	config := &tomlConfig{}
+
+	configBytes, err := readConf()
+	if err != nil {
+		return nil, err
+	}
+
+	err = toml.Unmarshal(configBytes, &config)
+	return config, err
+}
+
+// Returns an array of strings that contain the names
+// of the registries as defined in the system-wide
+// registries file.  it returns an empty array if none are
+// defined
+func GetRegistries() ([]string, error) {
+	config, err := loadRegistryConf()
+	if err != nil {
+		return nil, err
+	}
+	return config.Registries.Search.Registries, nil
+}
+
+// Returns an array of strings that contain the names
+// of the insecure registries as defined in the system-wide
+// registries file.  it returns an empty array if none are
+// defined
+func GetInsecureRegistries() ([]string, error) {
+	config, err := loadRegistryConf()
+	if err != nil {
+		return nil, err
+	}
+	return config.Registries.Insecure.Registries, nil
+}
--- a/pkg/sysregistries/system_registries_test.go
+++ b/pkg/sysregistries/system_registries_test.go
@@ -1,0 +1,60 @@
+package sysregistries
+
+import (
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+var testConfig = []byte("")
+
+func init() {
+	readConf = func() ([]byte, error) {
+		return testConfig, nil
+	}
+}
+
+func TestGetRegistriesWithBlankData(t *testing.T) {
+	testConfig = []byte("")
+	registriesConfig, _ := GetRegistries()
+	assert.Nil(t, registriesConfig)
+}
+
+func TestGetRegistriesWithData(t *testing.T) {
+	answer := []string{"one.com"}
+	testConfig = []byte(`[registries.search]
+registries= ['one.com']
+`)
+	registriesConfig, err := GetRegistries()
+	assert.Nil(t, err)
+	assert.Equal(t, registriesConfig, answer)
+}
+
+func TestGetRegistriesWithBadData(t *testing.T) {
+	testConfig = []byte(`registries:
+    - one.com
+    ,`)
+	_, err := GetRegistries()
+	assert.Error(t, err)
+}
+
+func TestGetInsecureRegistriesWithBlankData(t *testing.T) {
+	answer := []string(nil)
+	testConfig = []byte("")
+	insecureRegistriesConfig, err := GetInsecureRegistries()
+	assert.Nil(t, err)
+	assert.Equal(t, insecureRegistriesConfig, answer)
+}
+
+func TestGetInsecureRegistriesWithData(t *testing.T) {
+	answer := []string{"two.com", "three.com"}
+	testConfig = []byte(`[registries.search]
+registries = ['one.com']
+[registries.insecure]
+registries = ['two.com', 'three.com']
+`)
+	insecureRegistriesConfig, err := GetInsecureRegistries()
+	if err != nil {
+		t.Fail()
+	}
+	assert.Equal(t, insecureRegistriesConfig, answer)
+}
--- a/registries.conf
+++ b/registries.conf
@@ -1,0 +1,21 @@
+# For more information on this configuration file, see registries.conf(5).
+#
+# Registries to search for images that are not fully-qualified.
+# i.e. foobar.com/my_image:latest vs my_image:latest
+[registries.search]
+registries = []
+
+# Registries that do not use TLS when pulling images or uses self-signed
+# certificates.
+[registries.insecure]
+registries = []
+
+# Blocked Registries, blocks the `docker daemon` from pulling from the blocked registry.  If you specify
+# "*", then the docker daemon will only be allowed to pull from registries listed above in the search
+# registries.  Blocked Registries is deprecated because other container runtimes and tools will not use it.
+# It is recommended that you use the trust policy file /etc/containers/policy.json to control which
+# registries you want to allow users to pull and push from.  policy.json gives greater flexibility, and
+# supports all container runtimes and tools including the docker daemon, cri-o, buildah ...
+# The atomic CLI `atomic trust` can be used to easily configure the policy.json file.
+[registries.block]
+registries = []


runcom:
Thanks @baude - I'll take a look at this tomorrow

rhatdan:
@nalind PTAL
I have a bit of problem with this since it is in YAML while the containers/storage storage.conf is in TOML.  We need to be consistent.  I think CRI-O.conf is also TOML.  If we change the format, we would need to change atomic command and potentially have an upgrade script from YAML to TOML.

Also is there a decent library for handling TOML in C?

baude:
We could also crio.conf to yaml.  What @rhatdan didn't mention is this configuration file is already shipping.  And finally, if we decide to switch things, can we please merge this PR? I will have to fix the registries C code first, and then circle back to this one which I would prefer to do separately.

runcom:
Toml is so nice for config. I'm wondering why don't we port that C code to python at least where Toml libs are available instead.

pkg/sysregistries/system_registries_test.go
```diff
@@ -0,0 +1,68 @@
...
+	readConf = func() ([]byte, error) {
+		return testConfig, nil
+	}
+}
+
+func TestGetRegistriesWithBlankData(t *testing.T) {
+	answer := []string(nil)
+	testConfig = []byte("")
+	registriesConfig, err := GetRegistries()
+	if err != nil {
```
jwhonce: Since you've already pulled in the package: assert.nil(t, err) would result in less code.

registries.conf
```diff
@@ -0,0 +1,25 @@
+# This is a system-wide configuration file used to
+# keep track of registries for various container backends.
+# It adheres to YAML format and does not support recursive
+# lists of registries.
+
+# The default location for this configuration file is /etc/containers/registries.conf.
+
+# The only valid categories are: 'registries', 'insecure_registies',
```
jwhonce: insecure_registies should be insecure_registries

pkg/sysregistries/system_registries.go
```diff
@@ -0,0 +1,99 @@
...
+*/
+func GetInsecureRegistries() ([]string, error) {
+	config, err := loadRegistryConf()
+	if err != nil {
+		return blank_string_array, err
+	}
+
+	return config.InsecureRegistries, err
+
+}
```
jwhonce: Should there be a GetBlockRegistries(), as well?

runcom: Not needed actually, we'll accomplish the same with atomic trust as @rhatdan said in another thread

baude: yeah that is up for debate ... I thiknk that will be EOL'd.

baude:
@runcom because @rhatdan wanted the registries code to be a basic C executable and small.

pkg/sysregistries/system_registries.go
```diff
@@ -0,0 +1,99 @@
...
+*/
+var systemRegistriesConfPath = builtinRegistriesConfPath
+
+/*
+builtinRegistriesConfPath is the path to registry configuration file
+DO NOT change this, instead see systemRegistriesConfPath above.
+*/
+const builtinRegistriesConfPath = "/etc/containers/registries.conf"
+
+var blank_string_array []string
```
runcom: drop this, just `return nil` when you need to return an empty slice.

pkg/sysregistries/system_registries.go
```diff
@@ -0,0 +1,99 @@
...
+
+type registries struct {
+	Registries         []string `json:"registries"`
+	InsecureRegistries []string `json:"insecure_registries"`
+	BlockRegistries    []string `json:"block_registries"`
+}
+
+var blankFile = []byte("")
```
runcom: drop this, just `return nil` when you need to return an empty slice.

pkg/sysregistries/system_registries.go
```diff
@@ -0,0 +1,99 @@
...
+
+/*
+Reads the global registry file from the filesystem. Returns
+a byte array
+*/
+func readRegistryConf() ([]byte, error) {
+	configBytes, err := ioutil.ReadFile(systemRegistriesConfPath)
+	if err != nil {
+		return blankFile, err
```
runcom: as said above, this should be, `return nil, err`

pkg/sysregistries/system_registries.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	return configBytes, err
```
runcom: just `return configBytes, nil` (`err` is never not-nil here so no reason to return it)

pkg/sysregistries/system_registries.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+var blankRegistries registries
```
runcom: drop this, use `config` directly in `loadRegistryConf`

pkg/sysregistries/system_registries.go
```diff
@@ -0,0 +1,99 @@
...
+/*
+Loads the registry configuration file from the filesystem and
+then unmarshals it.  Returns the unmarshalled object.
+*/
+func loadRegistryConf() (registries, error) {
+	var config registries
+
+	configBytes, err := readConf()
+	if err != nil {
+		return blankRegistries, nil
```
runcom: `return config, err` <- notice you're not returning the error here also!!!

pkg/sysregistries/system_registries.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	err = yaml.Unmarshal(configBytes, &config)
+
+	if err != nil {
+		return blankRegistries, err
```
runcom: `return config, err`

pkg/sysregistries/system_registries.go
```diff
@@ -0,0 +1,99 @@
...
+/*
+Returns an array of strings that contain the names
+of the registries as defined in the system-wide
+registries file.  it returns an empty array if none are
+defined
+*/
+func GetRegistries() ([]string, error) {
+	config, err := loadRegistryConf()
+	if err != nil {
+		return blank_string_array, err
```
runcom: `return nil, err`

pkg/sysregistries/system_registries.go
```diff
@@ -0,0 +1,99 @@
...
+/*
+Returns an array of strings that contain the names
+of the insecure registries as defined in the system-wide
+registries file.  it returns an empty array if none are
+defined
+*/
+func GetInsecureRegistries() ([]string, error) {
+	config, err := loadRegistryConf()
+	if err != nil {
+		return blank_string_array, err
```
runcom: `return nil, err`

pkg/sysregistries/system_registries.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	return config.InsecureRegistries, err
```
runcom: replace `err` with just `nil`

docs/registries.conf.5.md
```diff
@@ -0,0 +1,28 @@
...
+The following example configuration defines two registries and one insecure registries.
+
+```
+registries:
+    - registry1.com
+    - registry2.com
+
+insecure_registries:
+    - registry3.com
+```
```
TomSweeneyRedHat: @runcom showed a registry in his blog that lives at:  /etc/crio/crio.conf and has the format of 
registries = [
“registry.example.xyz”,
]

Is it possible for us to use one common file?
https://docs.google.com/document/d/1cBT5qPxK-ooSfGsCYan7hRAzIlhZP-JFrvvohWQcwk4/edit

pkg/sysregistries/system_registries.go
```diff
@@ -0,0 +1,99 @@
+package sysregistries
+
+import (
+	"github.com/ghodss/yaml"
+	"io/ioutil"
+)
+
+/*
+systemRegistriesConfPath is the path to the system-wide registry configuration file
+and is used to add/subtract potential registries for obtaining images
```
TomSweeneyRedHat: way micro micro nit.  Missing period

docs/registries.conf.5.md
runcom: So, we'll keep having "registries" in crio.conf for now, but the plan is to move to this (/etc/containers/registries.conf) once we have this API in place I guess.

TomSweeneyRedHat: FWIW.  I think you ought to hold off on your blog post then.  I'd hate to have the blog go out and then a month from now the file you talk about configuring not be valid....

runcom: Well, we can mention in the blog post that the configuration isn't definitive and we'll likely move it. There's no worry behind this as nobody is using CRI-O anywhere ;) most people run cri-o to test it out so a configuration change won't affect anybody IMO

TomSweeneyRedHat: What I'm concerned about is not the now, but the later.  A year from now someone finds your blog via googling, is trying out CRI-O for the first time and the registries stuff doesn't work.  Bad taste in the mouth.  If you still want to get your blog out ASAP, then I think we should at least mention the coming change.  FWIW that's my $.02.

runcom: I think later we can just edit my post and mention the post no longer _fully_ stands linking to an updated one (l'll write later)

rhatdan: crio.conf will be allowed to override the registries list in registries.conf, but it should be commented out by default so it can share the content with other tools.   We are doing the same thing with storage.conf.

rhatdan: So crio.conf should work for now on, but we want to eventually add a comment telling the user that it is better to edit the other file.

pkg/sysregistries/system_registries_test.go
baude: Fixed

registries.conf
baude: fixed

pkg/sysregistries/system_registries.go
baude: leaving block in the conf file but no function in the Go code. left only for docker.

pkg/sysregistries/system_registries.go
baude: fixed.

pkg/sysregistries/system_registries.go
baude: fixed.

pkg/sysregistries/system_registries.go
baude: fixed

pkg/sysregistries/system_registries.go
baude: fixed

pkg/sysregistries/system_registries.go
baude: done

pkg/sysregistries/system_registries.go
baude: fixed

pkg/sysregistries/system_registries.go
baude: fixed

pkg/sysregistries/system_registries.go
baude: fixed

pkg/sysregistries/system_registries.go
baude: fixed

pkg/sysregistries/system_registries.go
baude: fixed

pkg/sysregistries/system_registries.go
baude: fixed

baude:
Reworked with review comments and switched to TOML.

pkg/sysregistries/system_registries_test.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+
+func TestGetRegistriesWithData(t *testing.T) {
+	answer := []string{"one.com"}
+	testConfig = []byte(`[registries]
+secure = ['one.com']
+`)
+	registriesConfig, err := GetRegistries()
+	if err != nil {
```
jwhonce: Should this be assert.NotNil()

pkg/sysregistries/system_registries_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+func TestGetInsecureRegistriesWithData(t *testing.T) {
+	answer := []string{"two.com", "three.com"}
+	testConfig = []byte(`[registries]
+secure = ['one.com']
+insecure = ['two.com', 'three.com']
+`)
+	insecureRegistriesConfig, err := GetInsecureRegistries()
+	if err != nil {
+		t.Logf("answer:%s\n", answer)
```
jwhonce: You log expected and actual but not the error

registries.conf
```diff
@@ -0,0 +1,22 @@
+# This is a system-wide configuration file used to
+# keep track of registries for various container backends.
+# It adheres to TOML format.
+
+# The default location for this configuration file is /etc/containers/registries.conf.
+
+# The only valid categories are: 'secure', 'insecure' and 'block'.
+
+
+[registries:]
```
jwhonce: You don't want the colon in the table name

pkg/sysregistries/system_registries_test.go
```diff
@@ -0,0 +1,67 @@
...
+    ,`)
+	_, err := GetRegistries()
+	assert.Error(t, err)
+}
+
+func TestGetInsecureRegistriesWithBlankData(t *testing.T) {
+	answer := []string(nil)
+	testConfig = []byte("")
+	insecureRegistriesConfig, err := GetInsecureRegistries()
+	if err != nil {
```
jwhonce: assert.NotNil()

pkg/sysregistries/system_registries.go
```diff
@@ -0,0 +1,98 @@
...
+/*
+Returns an array of strings that contain the names
+of the insecure registries as defined in the system-wide
+registries file.  it returns an empty array if none are
+defined
+*/
+func GetInsecureRegistries() ([]string, error) {
+	config, err := loadRegistryConf()
+	if err != nil {
+		return nil, err
```
jwhonce: In idiomatic golang:
if config, err := loadRegistryConf(); err != nil {
    return nil, err
}

pkg/sysregistries/system_registries_test.go
baude: leftover, nice catch

pkg/sysregistries/system_registries_test.go
baude: did assert.nil on the error

pkg/sysregistries/system_registries_test.go
baude: did assert.nil on the error

registries.conf
baude: fixed.

docs/registries.conf.5.md
```diff
@@ -0,0 +1,25 @@
...
+categories: `registries` and `insecure_registries`.
+
+The following example configuration defines two registries and one insecure registries.
+
+```
+[registries]
+secure = ["registry1.com", "registry2.com"]
+insecure = ["registry3.com"]
+```
+
```
TomSweeneyRedHat: You've 'blocked' below too.  Should that be talked about?  If so, it might be a good example for showing an empty set.

pkg/sysregistries/system_registries.go
TomSweeneyRedHat: 👍 This is something I need to remember to do more of myself.

registries.conf
```diff
@@ -0,0 +1,22 @@
+# This is a system-wide configuration file used to
+# keep track of registries for various container backends.
+# It adheres to TOML format.
+
+# The default location for this configuration file is /etc/containers/registries.conf.
+
+# The only valid categories are: 'secure', 'insecure' and 'block'.
+
```
TomSweeneyRedHat: Perhaps add something about multiples?:  "If you need to list multiple registries in a category, each registry should be quoted and separated with a comma.  As an example ["registry.mydomain.com", "registry.access.redhat.com"].  The registries will be searched in order until the requested image is located.

TomSweeneyRedHat:
LGTM.  A couple of documentation suggestions to consider.

docs/registries.conf.5.md
rhatdan: Lets add documentation that this support is deprecated, and only suppored with docker daemon.  We recommend even in the docker case that the users setup atomic trust.  Using /etc/containers/policy.conf

pkg/sysregistries/system_registries.go
```diff
@@ -0,0 +1,98 @@
+package sysregistries
+
+import (
+	"github.com/BurntSushi/toml"
+	"io/ioutil"
+)
+
+/*
```
runcom: comments should be `//` usually

pkg/sysregistries/system_registries.go
runcom: seems like this is not fixed. it's still `return blankFile, err` - drop `blankFile` altogether also

pkg/sysregistries/system_registries.go
```diff
@@ -0,0 +1,98 @@
...
+Loads the registry configuration file from the filesystem and
+then unmarshals it.  Returns the unmarshalled object.
+*/
+func loadRegistryConf() (*tomlConfig, error) {
+	var config tomlConfig
+
+	configBytes, err := readConf()
+	if err != nil {
+		return &config, err
```
runcom: you can return `nil` directly since this is a pointer now.

pkg/sysregistries/system_registries.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	err = toml.Unmarshal(configBytes, &config)
+
+	if err != nil {
+		return &config, err
```
runcom: as above, `return nil, err`

pkg/sysregistries/system_registries.go
```diff
@@ -0,0 +1,98 @@
...
+
+// For mocking in unittests
+var readConf = readRegistryConf
+
+/*
+Loads the registry configuration file from the filesystem and
+then unmarshals it.  Returns the unmarshalled object.
+*/
+func loadRegistryConf() (*tomlConfig, error) {
+	var config tomlConfig
```
runcom: also, initialize this as `config := &tomlConfig{}` so this is already a pointer and  you don't have to use `&`

pkg/sysregistries/system_registries.go
```diff
@@ -0,0 +1,98 @@
...
+of the registries as defined in the system-wide
+registries file.  it returns an empty array if none are
+defined
+*/
+func GetRegistries() ([]string, error) {
+	config, err := loadRegistryConf()
+	if err != nil {
+		return nil, err
+	}
+	return config.Registries.SecureRegistries, nil
```
runcom: I still don't understand what `Secure...` means here - this is not about secure vs insecure registries, this is just _additional registries_. This notion of SecureRegistries must go away as it's misleading. There's no say as "if it's not in insecure then it's considered secure" - this argument doesn't stand at all here, those are just "Additional Registries" which we call Registries here.

runcom: Please take care of this one

docs/registries.conf.5.md
```diff
@@ -0,0 +1,25 @@
...
+# FORMAT
+The [TOML_format][toml] is used to build simple list format for registries under two
+categories: `registries` and `insecure_registries`.
+
+The following example configuration defines two registries and one insecure registries.
+
+```
+[registries]
+secure = ["registry1.com", "registry2.com"]
```
runcom: PLEASE DROP this `secure` key, this is not about secure vs insecure. Those are additional registries. These and insecure registries are totally orthogonal concepts. Those should say "Additional" not secure.

registries.conf
```diff
@@ -0,0 +1,22 @@
...
+
+[registries]
+secure = ["registry.access.redhat.com"]
```
runcom: DROP THIS secure

registries.conf
```diff
@@ -0,0 +1,22 @@
+# This is a system-wide configuration file used to
+# keep track of registries for various container backends.
+# It adheres to TOML format.
+
+# The default location for this configuration file is /etc/containers/registries.conf.
+
+# The only valid categories are: 'secure', 'insecure' and 'block'.
```
runcom: drop secure

docs/registries.conf.5.md
runcom: The additional registries feature doesn't say anything about secureness - you can add a bogus, untrusted, dangerous registry in here just fine.
The `insecure` list is instead for registries that do not have TLS verification. They can be as secure as possible within a development environment. This is just to say that the `secure` keyword is wrong and doesn't reflect what the additional reigstries key is truly for

pkg/sysregistries/system_registries.go
baude: Fixed

pkg/sysregistries/system_registries.go
baude: Fixed

pkg/sysregistries/system_registries.go
baude: fixed

pkg/sysregistries/system_registries.go
baude: fixed

pkg/sysregistries/system_registries.go
baude: fixed

docs/registries.conf.5.md
baude: fixed

registries.conf
baude: fixed

registries.conf
baude: fixed

registries.conf
baude: ok

docs/registries.conf.5.md
baude: done, added to the conf file

pkg/sysregistries/system_registries.go
baude: yup, got it

baude:
updated with more review comments and name changes to the toml categories.

rhatdan:
@baude What do you think about adding some of the content I just wrote up in the man page for registies.conf

docs/registries.conf.5.md
```diff
@@ -0,0 +1,26 @@
...
+registry.conf - Syntax of System Registry Configuration File
+
+# DESCRIPTION
+The REGISTRIES configuration file is a system-wide configuration file for container image
+registries. The file format is TOML.
+
+# FORMAT
+The [TOML_format][toml] is used to build simple list format for registries under two
+categories: `registries` and `insecure_registries`. You can list multiple registries using
+as comma separated list.
```
nalind: Typo: `as` -> `a`.

docs/registries.conf.5.md
```diff
@@ -0,0 +1,26 @@
...
+
+The following example configuration defines two registries and one insecure registries.
```
nalind: Typo: that second `registries` should be `registry`.

docs/registries.conf.5.md
```diff
@@ -0,0 +1,26 @@
...
+# NAME
+registry.conf - Syntax of System Registry Configuration File
+
+# DESCRIPTION
+The REGISTRIES configuration file is a system-wide configuration file for container image
+registries. The file format is TOML.
+
+# FORMAT
+The [TOML_format][toml] is used to build simple list format for registries under two
+categories: `registries` and `insecure_registries`. You can list multiple registries using
```
nalind: Shouldn't we be mentioning the `block` category here, too?

nalind: Should the list should be `additional`, `insecure`, and `block`?

baude: We really want to deprecate that notion so I am only putting it in the conf file.

docs/registries.conf.5.md
baude: fixed, nice catch

docs/registries.conf.5.md
baude: fixed

docs/registries.conf.5.md
TomSweeneyRedHat: I know there's been several loops on this and I think additional, secure and registries have all been used.  Can/should we rename these (and possibly drive Brent completely bonkers) to 
public_registries and authenticated_registries?

I really, really don't like additional.  It's "additional" to what? Do you get docker.io by default if you don't specify anything in additional?  

My $.02, I won't harp, but did want to raise the thought.

baude: @nalid the block thing is only used by a different parser (aka atomic-registries) so the functions ro read and handle are actually there ... because we want to deprecate it, i never added a function into this parsing library.

baude: the naming thing could be avoided if we didn't use registries as a toml heading.  then you can have three key value pairs.

registries = []
insecure_registries = []
....

I know this is a bit against the TOML approach which likes headers, but the header is someone benign here and not used.

runcom: Tom, public and authenticated don't work here. Those registeies aren't public nor anything related to authentication. 
If the header is the bikeshed here, let's just call the header "image_registries" and then "registries" and "insecure_registries"

TomSweeneyRedHat: image_registries, registries and insecure_registries would work for me.

I think I'm still a bit fuzzy about the difference.  Could dockerhub registry without authentication and then a private registry inside the company with authentication both be listed under "additonal/registries"?  I was thinking the dockerhub would be in the "insecure_registries" and that's where I was going with my naming.

rhatdan: Does this work
[ registries ]
registries=[`docker.io`, `registry.access.redhat.com` ]

[registries.insecure ]
registries=[ `registry.acme.inc` ]

[registries.blocked]
registries=[ `hackme.com` ]

rhatdan: Or maybe the first is 
[registries.search]

rhatdan:
I wrote a long description of the three fields.  And now can not find it.  :^(  Did It ever get posted.

docs/registries.conf.5.md
runcom: I'll write a blog post to make this clear I guess.. So much confusion going on lol

rhatdan:
registries.conf currently support three categories of registries. 

```
# Search registries are used when the caller of a container runtime does not fully specify the 
# container image that they want to execute.  These registries are prepended onto the front of the 
# specified container image until the named image is found at a registry.
[registries.search]
registries = ['docker.io']

# Insecure Registries.  By default container runtimes use TLS when retrieving images from a registry.  If 
# registry is not setup with TLS, then the container runtime will fail to pull images from the registry.  
# if you add the registry to the list of insecure registries then the container runtime will attempt use 
# standard web protocols to pull the image.  Note insecure registries can be used for any registry, not just 
# the registries listed under search
[registries.insecure]
registries = ['registry.internel.company.com'']

# Blocked Registries, blocks the `docker daemon` from pulling from the blocked registry.  If you specify
# all, then the docker daemon will only be allowed to pull from registries listed above in the search 
# registries.  Blocked Registries is deprecated because other container runtimes and tools will not use it.
# It is recommended that you use the trust policy file /etc/containers/policy.json to control which 
# registries you want to allow users to pull and push from.  policy.json gives greater flexibility, and 
# supports all container runtimes and tools including the docker daemon, cri-o, buildah ...  
# The atomic CLI `atomic trust` can be used to easily configure the policy.json file.
[registries.blocked]
registries = ['hackme.com'']
```

TomSweeneyRedHat:
@rhatdan I like the verbiage you posted in your previous reply.  If we could meld that into the manpage/doc, that would be great.  @runcom a blog topic on this after it release would be awesome.  All thanks for the info and explanations.  Hang in there Brent, this really will be pretty slick functionality for our users once complete.

baude:
@rhatdan ptal

pkg/sysregistries/system_registries.go
```diff
@@ -0,0 +1,84 @@
+package sysregistries
+
+import (
+	"github.com/BurntSushi/toml"
+	"io/ioutil"
+)
+
+//systemRegistriesConfPath is the path to the system-wide registry configuration file
```
runcom: Go format needs a space after `//`

runcom: can you actually run `gofmt -s -w THISFILEANDOTHERINTHISPR`

pkg/sysregistries/system_registries.go
```diff
@@ -0,0 +1,84 @@
...
+func loadRegistryConf() (*tomlConfig, error) {
+	config := &tomlConfig{}
+
+	configBytes, err := readConf()
+	if err != nil {
+		return nil, err
+	}
+
+	err = toml.Unmarshal(configBytes, &config)
+
```
runcom: drop this space and:
```
err = toml.Unmarshal(configBytes, &config)
return config, err
```

pkg/sysregistries/system_registries.go
```diff
@@ -0,0 +1,84 @@
...
+		Insecure registries `toml:"insecure"`
+		Block    registries `toml:"block"`
+	} `toml:"registries"`
+}
+
+//Reads the global registry file from the filesystem. Returns
+//a byte array
+func readRegistryConf() ([]byte, error) {
+	configBytes, err := ioutil.ReadFile(systemRegistriesConfPath)
+	if err != nil {
```
runcom: you can drop this `if` and:
```
return configBytes, err
```

Makefile
```diff
@@ -10,6 +10,8 @@ BUILDFLAGS := -tags "$(BUILDTAGS)"
 
 PACKAGES := $(shell go list ./... | grep -v github.com/containers/image/vendor)
 
+NO_GIT_VALIDATION = "registries.conf"
```
mtrmac: Could this have a comment explaining why this is necessary?

pkg/sysregistries/system_registries.go
baude: fixed

pkg/sysregistries/system_registries.go
```diff
@@ -0,0 +1,84 @@
...
+		return nil, err
+	}
+	return configBytes, nil
+}
+
+// For mocking in unittests
+var readConf = readRegistryConf
```
mtrmac: Totally non-blocking: A global state like this is a bit awkward, e.g. for running tests in parallel.

Could this instead be done by having `loadRegistryConfWithReader`, which is tested, and an untested `loadRegistryConf() { return loadRegistryConfWithReader(readConf) }` ?

pkg/sysregistries/system_registries.go
```diff
@@ -0,0 +1,84 @@
...
+		Search   registries `toml:"search"`
+		Insecure registries `toml:"insecure"`
+		Block    registries `toml:"block"`
+	} `toml:"registries"`
+}
+
+//Reads the global registry file from the filesystem. Returns
+//a byte array
+func readRegistryConf() ([]byte, error) {
+	configBytes, err := ioutil.ReadFile(systemRegistriesConfPath)
```
mtrmac: (This could just be a `return ioutil.…`, but this is also fine.)

pkg/sysregistries/system_registries.go
```diff
@@ -0,0 +1,84 @@
...
+//Returns an array of strings that contain the names
+//of the insecure registries as defined in the system-wide
+//registries file.  it returns an empty array if none are
+//defined
+func GetInsecureRegistries() ([]string, error) {
+	config, err := loadRegistryConf()
+	if err != nil {
+		return nil, err
+	}
+	return config.Registries.Insecure.Registries, err
```
mtrmac: `return …, nil`, please.

pkg/sysregistries/system_registries.go
baude: fixed.

registries.conf
```diff
@@ -0,0 +1,23 @@
...
+[registries.search]
+registries = []
+
+# Insecure Registries.  By default container runtimes use TLS when retrieving images from a registry.  If
+# registry is not setup with TLS, then the container runtime will fail to pull images from the registry.
+# if you add the registry to the list of insecure registries then the container runtime will attempt use
+# standard web protocols to pull the image.  Note insecure registries can be used for any registry, not just
+# the registries listed under search
+[registries.insecure]
+registries = [']
```
mtrmac: The `'` is intentional here?

pkg/sysregistries/system_registries.go
baude: fixed

pkg/sysregistries/system_registries.go
mtrmac: … scratch that, that won’t work.

Makefile
```diff
@@ -66,7 +68,8 @@ EPOCH_TEST_COMMIT ?= e68e0e1110e64f906f9b482e548f17d73e02e6b1
 .gitvalidation:
 	@which git-validation > /dev/null 2>/dev/null || (echo "ERROR: git-validation not found. Consider 'make clean && make tools'" && false)
 ifeq ($(TRAVIS),true)
-	@git-validation -q -run DCO,short-subject,dangling-whitespace
+	GIT_CHECK_EXCLUDE=${NO_GIT_VALIDATION} git-validation -q -run DCO,short-subject,dangling-whitespace
```
baude: excluding the configuration file from the gitvalidation scan due to the conf file having/and needing/ whitespace

Makefile
baude: fixed

pkg/sysregistries/system_registries.go
baude: fixed

registries.conf
baude: fixed

docs/registries.conf.5.md
mtrmac: Still needs `s/as a/a/` I think.

docs/registries.conf.5.md
```diff
@@ -0,0 +1,28 @@
+% registries.conf(5) System-wide registry configuration file
+% Brent Baude
+% Aug 2017
+
+# NAME
+registry.conf - Syntax of System Registry Configuration File
```
mtrmac: `s/registry.conf/registries.conf/`

docs/registries.conf.5.md
```diff
@@ -0,0 +1,28 @@
...
+
+# DESCRIPTION
+The REGISTRIES configuration file is a system-wide configuration file for container image
+registries. The file format is TOML.
+
+# FORMAT
+The [TOML_format][toml] is used to build simple list format for registries under two
```
mtrmac: Is the `[][]` syntax supposed to be a link?

`s/simple list format/a &/`?

docs/registries.conf.5.md
```diff
@@ -0,0 +1,28 @@
...
+as a comma separated list.
+
+The following example configuration defines two registries and one insecure registry.
+
+```
+[registries.search]
+registries = ["registry1.com", "registry2.com"]
+
+[registries.insecure]
+registries = ["registry3.com"]
```
mtrmac: The above does not really say much about what the values mean; that is, instead, contained in the comments inside the file. That’s a bit unusual.

If the substantive documentation remains in the config file comments, this man page should at least mention that, so that users don’t go looking elsewhere.

Though I’d be happier if the core of the documentation (except for the block_registries block, I guess) were in the man page, and the config file pointed to the man page, instead of the other way around: The comments in the config file can easily get lost, e.g. when a tool overwrites the file with a generated version. The man page, OTOH will end up on various man page servers / search systems.

registries.conf
```diff
@@ -0,0 +1,23 @@
...
+# Insecure Registries.  By default container runtimes use TLS when retrieving images from a registry.  If
+# registry is not setup with TLS, then the container runtime will fail to pull images from the registry.
+# if you add the registry to the list of insecure registries then the container runtime will attempt use
+# standard web protocols to pull the image.  Note insecure registries can be used for any registry, not just
+# the registries listed under search
+[registries.insecure]
+registries = []
+
+# Blocked Registries, blocks the `docker daemon` from pulling from the blocked registry.  If you specify
+# all, then the docker daemon will only be allowed to pull from registries listed above in the search
```
mtrmac: What does “specify all” mean? Literally the string `"all"` instead of a real hostname?

docs/registries.conf.5.md
baude: fixed.

docs/registries.conf.5.md
baude: was copied from another, similar man page.  I have reworked it without the []'s

docs/registries.conf.5.md
baude: This is an excellent comment and on point.  I've taken the search and insecure registry descriptions with comments from above fixed and added them to the man page.  I've put a very abbreviated summary in the configuration file instead and will mention that further description is available in the man pages.  I left the blocked description out of the man page as it is being deprecated but left the full description in the conf file.

registries.conf
baude: According to the RHEL documentation, a "*" can be used for all.  I'll update accordingly.

Makefile
```diff
@@ -66,7 +66,8 @@ EPOCH_TEST_COMMIT ?= e68e0e1110e64f906f9b482e548f17d73e02e6b1
 .gitvalidation:
 	@which git-validation > /dev/null 2>/dev/null || (echo "ERROR: git-validation not found. Consider 'make clean && make tools'" && false)
 ifeq ($(TRAVIS),true)
-	@git-validation -q -run DCO,short-subject,dangling-whitespace
+	git-validation -q -run DCO,short-subject,dangling-whitespace
 else
-	@git-validation -q -run DCO,short-subject,dangling-whitespace -range $(EPOCH_TEST_COMMIT)..HEAD
+	git-validation -q -run DCO,short-subject,dangling-whitespace -range $(EPOCH_TEST_COMMIT)..HEAD
+
```
mtrmac: Nit: please drop this empty line.

docs/registries.conf.5.md
mtrmac: Is this _intended_ to be a link? If so, _some_ markup is probably required.

docs/registries.conf.5.md
```diff
@@ -0,0 +1,40 @@
...
+Search registries are used when the caller of a container runtime does not fully specify the
+container image that they want to execute.  These registries are prepended onto the front
+ of the specified container image until the named image is found at a registry.
+
+Insecure Registries.  By default container runtimes use TLS when retrieving images
+from a registry.  If the registry is not setup with TLS, then the container runtime
+will fail to pull images from the registry. If you add the registry to the list of
+insecure registries then the container runtime will attempt use standard web protocols to
+pull the image.  Note insecure registries can be used for any registry, not just the
+registries listed under search.
```
mtrmac: Is `registries.insecure` _only_ for non-TLS registries, or does it also allow untrusted TLS (e.g. self-signed certificates)?

pkg/sysregistries/system_registries.go
```diff
@@ -0,0 +1,77 @@
...
+// of the insecure registries as defined in the system-wide
+// registries file.  it returns an empty array if none are
+// defined
+func GetInsecureRegistries() ([]string, error) {
+	config, err := loadRegistryConf()
+	if err != nil {
+		return nil, err
+	}
+	return config.Registries.Insecure.Registries, nil
+
```
mtrmac: Nit: please drop this empty line.

registries.conf
```diff
@@ -0,0 +1,20 @@
+# For more information on this configuration file, see registries.conf(5).
+#
+# Registries to search for images that are not fully-qualified.
+# i.e. foobar.com/my_image:latest vs my_image:latest
+[registries.search]
+registries = []
+
+# Registries that do not use TLS when pulling images.
```
mtrmac: Same as in the man page, does this include untrusted TLS?

Makefile
baude: Fixed

docs/registries.conf.5.md
baude: fixed

pkg/sysregistries/system_registries.go
baude: fixed.

registries.conf
baude: fixed.

mtrmac:
:+1: Thanks!

@runcom PTAL

runcom:
LGTM

mtrmac:
Is it known how `registries.conf` will get from the root of this repo into an RPM and ultimately onto an end-user’s system? Note that containers/image has no `make install` (and it wouldn’t really make sense for it to have one).

Perhaps the file should instead live in skopeo, which does have a `make install`?

rhatdan:
Right now we have a similar file in containers/image storage.conf, which skopeo just reaches in and pulls it out,  I would prefer to have the example config with the package that implements it and then have skopeo either pull it out or just use a local copy of it.  Similarly the man page describing the config should be in containers/image and container/storage.

Issue #324, opened 2017-08-11
Makefile improvements


hferentschik:
I suggest some Makefile improvements which I added while working isuee #155.

rhatdan:
@vdemeester @mtrmac @vrothberg Is this still an issue?  Can we close it?

vdemeester:
I think it's closeable indeed :wink:

PR #325, opened 2017-08-11
Issue #324 Adding build and fmt target

hferentschik:
* Add dedicated build target
* Make sure development works on macOS
* Add target to call go fmt

Fix for issue #324

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: all tools test validate lint
+.PHONY: all tools test validate lint .gitvalidation fmt
 
 # Which github repostiory and branch to use for testing with skopeo
 SKOPEO_REPO = projectatomic/skopeo
@@ -13,6 +13,8 @@ BUILDTAGS = btrfs_noversion libdm_no_deferred_remove $(DARWIN_BUILD_TAG)
 BUILDFLAGS := -tags "$(BUILDTAGS)"
 
 PACKAGES := $(shell go list $(BUILDFLAGS) ./... | grep -v github.com/containers/image/vendor)
+SOURCE_DIRS = $(shell echo $(PACKAGES) | awk 'BEGIN{FS="/"; RS=" "}{print $$4}' | uniq)
+
 # On macOS, (brew install gpgme) installs it within /usr/local, but /usr/local/include is not in the default search path.
 # Rather than hard-code this directory, use gpgme-config. Sadly that must be done at the top-level user
 # instead of locally in the gpgme subpackage, because cgo supports only pkg-config, not general shell scripts,
@@ -21,7 +23,10 @@ PACKAGES := $(shell go list $(BUILDFLAGS) ./... | grep -v github.com/containers/
 # (and the user will probably find out because the cgo compilation will fail).
 GPGME_ENV = CGO_CFLAGS="$(shell gpgme-config --cflags 2>/dev/null)" CGO_LDFLAGS="$(shell gpgme-config --libs 2>/dev/null)"
 
-all: tools .gitvalidation test validate
+all: tools test validate .gitvalidation
+
+build: vendor
+	@$(GPGME_ENV) go build $(BUILDFLAGS) $(PACKAGES)
 
 tools: tools.timestamp
 
@@ -58,6 +63,9 @@ test-skopeo:
 		$(SUDO) make BUILDTAGS="$(BUILDTAGS)" check && \
 		rm -rf $${skopeo_path}
 
+fmt:
+	@gofmt -l -s -w $(SOURCE_DIRS)
+
 validate: lint
 	@go vet $(PACKAGES)
 	@test -z "$$(gofmt -s -l . | grep -ve '^vendor' | tee /dev/stderr)"
@@ -68,8 +76,6 @@ lint:
 		echo "$$out"; \
 		exit 1; \
 	fi
-
-.PHONY: .gitvalidation
 
 EPOCH_TEST_COMMIT ?= e68e0e1110e64f906f9b482e548f17d73e02e6b1
 


hferentschik:
This is how I build on macOS:

    $ make build BUILDTAGS="containers_image_openpgp containers_image_ostree_stub exclude_graphdriver_devicemapper"

hferentschik:
Not sure what's going on, but I signed the commits now and afaict the commits are now "verified". Still the Travis build fails with:

     2fcd6fb - FAIL - does not have a valid DCO
     9b2faf5 - FAIL - does not have a valid DCO

Not sure what to do with these failing validation at this stage. 

I am wondering though whether .gitvalidation should not run last. I understand that for an actual merge of this pull request you want to ensure certain conditions for the commits, but for me this is really the last step. It would be nice to see whether things actually build and test fails.

rhatdan:
The pull requests have to have a signature inside of them
Try
git commit -a --amend -s

This will add you signature to the commit.

hferentschik:
> The pull requests have to have a signature inside of them

Got you. That worked. Thanks.

runcom:
@nalind @giuseppe ptal

Makefile
```diff
@@ -5,12 +5,16 @@ SKOPEO_REPO = projectatomic/skopeo
 SKOPEO_BRANCH = master
 # Set SUDO=sudo to run container integration tests using sudo.
 SUDO =
-BUILDTAGS   = btrfs_noversion libdm_no_deferred_remove
+BUILDTAGS  ?= btrfs_noversion libdm_no_deferred_remove
 BUILDFLAGS := -tags "$(BUILDTAGS)"
 
 PACKAGES := $(shell go list ./... | grep -v github.com/containers/image/vendor)
+SOURCE_DIRS := $(shell echo $(PACKAGES) |  tr " " "\n" | awk -F '/' '{print $$4}')
```
giuseppe: you could probably skip the tr filter here using `-vRS=' '` for awk

giuseppe:
the ostree changes LGTM

Makefile
hferentschik: let me try ..

hferentschik:
Pushed an update where I changed the 'awk' command a bit and also added uniq.

mtrmac:
> This is how I build on macOS:
>
>    $ make build BUILDTAGS="containers_image_openpgp containers_image_ostree_stub exclude_graphdriver_devicemapper"

Can some of that be automated? AFAICT libostree can’t build on macOS (ostree/src/libostree/ostree-linuxfsutil.c hard-codes Linux-specific header files and ioctls), so the `Makefile` can safely set the `containers_image_ostree_stub` tag.

(Or we could make the ostree build conditions directly conditional on Linux, but that would force us to duplicate the logic over all those files. At least setting the build tag would be a clear improvement with no downsides.)

`containers_image_openpgp` is a genuine choice AFAIK; I have no idea bout `exclude_graphdriver_devicemapper`.

mtrmac:
> I am wondering though whether .gitvalidation should not run last. I understand that for an actual merge of this pull request you want to ensure certain conditions for the commits, but for me this is really the last step. It would be nice to see whether things actually build and test fails.

Rearranging the order of the dependencies of the `all` target to make the edit-test cycle more convenient would make sense to me (`.gitvalidation` also seems much slower on macOS than Linux, so having it run before tests is fairly annoying there).

hferentschik:
Sorry for the long turn around. I was on vacation and just returned. I did rebase of this pull request on top of master and pushed an update. What do you guys think?

> Can some of that be automated? AFAICT libostree can’t build on macOS (ostree/src/libostree/ostree-linuxfsutil.c hard-codes Linux-specific header files and ioctls), so the Makefile can safely set the containers_image_ostree_stub tag.

AFAICT this has already happened if an _ifeq_ clause for Darwin.

hferentschik:
> Rearranging the order of the dependencies of the all target to make the edit-test cycle more convenient would make sense to me

I've put .gitvalidation last for now.

Makefile
```diff
@@ -7,12 +7,14 @@ SKOPEO_BRANCH = master
 SUDO =
 
 ifeq ($(shell uname),Darwin)
-DARWIN_BUILD_TAG = containers_image_ostree_stub
+	DARWIN_BUILD_TAG = containers_image_ostree_stub
```
mtrmac: Isn’t this indentation risky? If a rule somehow ended up above the `ifeq`, this would be interpreted as a shell command.

Makefile
```diff
@@ -7,12 +7,14 @@ SKOPEO_BRANCH = master
...
 endif
 BUILDTAGS = btrfs_noversion libdm_no_deferred_remove $(DARWIN_BUILD_TAG)
 BUILDFLAGS := -tags "$(BUILDTAGS)"
 
 PACKAGES := $(shell go list $(BUILDFLAGS) ./... | grep -v github.com/containers/image/vendor)
+SOURCE_DIRS := $(shell echo $(PACKAGES) | awk 'BEGIN{FS="/"; RS=" "}{print $$4}' | uniq)
```
mtrmac: If this is used in only one rule (unlike `PACKAGES`), and not intended to be overwritten by the user, it would be better to compute the list inside that single rule so that we avoid the three extra processes on every `make`.

Makefile
```diff
@@ -21,7 +23,10 @@ PACKAGES := $(shell go list $(BUILDFLAGS) ./... | grep -v github.com/containers/
 # (and the user will probably find out because the cgo compilation will fail).
 GPGME_ENV = CGO_CFLAGS="$(shell gpgme-config --cflags 2>/dev/null)" CGO_LDFLAGS="$(shell gpgme-config --libs 2>/dev/null)"
 
-all: tools .gitvalidation test validate
+all: tools build test validate .gitvalidation
+
+build: vendor
+	@$(GPGME_ENV) go build $(BUILDFLAGS) $(PACKAGES)
```
mtrmac: `go build --help`:
> When compiling multiple packages or a single non-main package,
> build compiles the packages but discards the resulting object,
> serving only as a check that the packages can be built.

What does this do in addition to the existing `test` target?

The `test` target implicitly checks that each package with tests can be built, and AFAICS even for packages with no `*_test.go` files it seems to do a full compilation, or something very close. Is there any extra benefit to a separate `go build`?

ostree/ostree_stub.go
```diff
@@ -0,0 +1,3 @@
+// +build containers_image_ostree_stub
```
mtrmac: Why is this necessary? AFAICS the `PACKAGES` definition already excludes `c/i/ostree` from the `Makefile` operations.

If an external user did refer to the `ostree` package while the stub tag is enabled, failing with `no buildable Go source files in …/github.com/containers/image/ostree` seems a bit better than failing with `undefined: ostree.SomeSymbol`.

Makefile
hferentschik: Good catch. Theoretically, you are right. I think the risk for this is happening is quite low, but let's move it back.

Makefile
hferentschik: True, but I think the overhead is minimal and we would use consistently simply expanded variables. The difference is really subtle (in this case) and I think many people are simply not aware of the difference between ':=' and '='.

If you feel strongly about it, I can change it.

Makefile
hferentschik: > What does this do in addition to the existing test target

It does less in this regards. It just complies. I find it extremely odd to have a Makefile which does not have a build/compile target. The _test_ target might do this implicitly, but imo this is not the same. If I just want to check whether something complies, it feels unnatural to call test. Also as you say, I am also wondering what the exact behaviour is of go test for packages w/o test files. Is it actually specified that it does a full compile as well?

Makefile
mtrmac: `=` vs. `:=` would minimize the run-time overhead, but it would still mean a global variable with a single user.  The `make` variable can be entirely eliminated by computing `$source_dirs` inside the shell rules for `fmt`.

Makefile
mtrmac: > I find it extremely odd to have a Makefile which does not have a build/compile target.

\*shrug\* I find Golang’s lack of shared libraries problematic for many reasons, but that’s Golang for you. The point is that `go build` does not create any artifact, so a
```make
build:
    /bin/true
```
rule would work just as well, and be equally pointless.

>  If I just want to check whether something complies, it feels unnatural to call test.

If the rule is not producing anything consumable, and it’s only effect is verifying whether the code is usable, that seems entirely appropriate for the “run self-tests of the package” target, i.e. `test`.

mtrmac: > Is it actually specified that it does a full compile as well?

`go help test`:
```
'Go test' recompiles each package along with any files with names matching
the file pattern "*_test.go".
```

That seems to specify that the package is always compiled, though it isn’t as explicit as we might like.

ostree/ostree_stub.go
hferentschik: ok

Makefile
hferentschik: > = vs. := would minimize the run-time overhead, but it would still mean a global variable with a single user. 

Right. IM, however,  it makes sense to define it globally. Partly because it creates some grouping which helps me understand the build and partly since I see this as a valuable variable for future targets.

> The make variable can be entirely eliminated by computing $source_dirs inside the shell rules for fmt.

Sure

What's about using '=' then?

mtrmac: > Partly because it creates some grouping which helps me understand the build

OTOH it splits the working of the `fmt` target into two faraway places.

> since I see this as a valuable variable for future targets.

Maybe; an uncertain bet.

> What's about using '=' then?

Fair enough, better than `:=` at least.

Makefile
hferentschik: > The point is that go build does not create any artifact, so a

> build:
>    /bin/true
> rule would work just as well, and be equally pointless.

Why is something which verifies that all packages compile pointless? Why does the outcome need to be a binary? 

Since we are quoting from the go help: `go build -f`:

> When compiling multiple packages or a single non-main package,
> build compiles the packages but discards the resulting object,
> serving only as a check that the packages can be built.

which is exactly what I want. It does explicitly what I want, whereas _go test_ does it as a side-effect. Often I just want to know whether something compiles. Running tests is not needed or overhead, forcing me to abort the command.

> That seems to specify that the package is always compiled, though it isn’t as explicit as we might like.

Right, hence `go build`

mtrmac: You started with
> I find it extremely odd to have a Makefile which does not have a build/compile target.

Well, this is a build/compile target which does produce anything built/compiled.

mtrmac: > Often I just want to know whether something compiles. Running tests is not needed or overhead

I don’t quite get what that is useful for. I guess my development process is different; my IDE immediately points out syntax errors in the current package, making a `Makefile` rule to compile all packages unnecessary.

---

Anyway, my primary concern is that having both `build` and `test` in `all` means that in a edit-compile-test cycle everything is unnecessarily compiled twice, slowing progress down.

I guess I could live with a `build` target which is not in `all`, though I would still find it unnecessary and pretty confusing.

hferentschik: > I guess I could live with a build target which is not in all,

Ok. Fair enough.

>  though I would still find it unnecessary and pretty confusing.

I think we have to agree to disagree ;-) I find having to use 'make test' counter-intuitive.

hferentschik:
@mtrmac, I've pushed another update. Not much left from my original pull request ;-) I guess several things got addressed in between me creating the initial pull request and now. WDYT?

mtrmac:
:+1: I guess.

rhatdan:
Looks like it needs some rebasing?
@runcom PTAL

hferentschik:
Rebased.

hferentschik:
Merge?

runcom:
rebase and I'll merge sorry

runcom:
lgtm

hferentschik:
> rebase and I'll merge sorry

rebased

PR #326, opened 2017-08-11
Issue #155 Making sure existing OCI index gets not overwritten

hferentschik:
Addresses issue #155

### DIFF
--- a/oci/layout/oci_dest.go
+++ b/oci/layout/oci_dest.go
@@ -27,12 +27,23 @@ func newImageDestination(ref ociReference) (types.ImageDestination, error) {
 	if ref.image == "" {
 		return nil, errors.Errorf("cannot save image with empty image.ref.name")
 	}
-	index := imgspecv1.Index{
-		Versioned: imgspec.Versioned{
-			SchemaVersion: 2,
-		},
+
+	var index *imgspecv1.Index
+	if indexExists(ref) {
+		var err error
+		index, err = ref.getIndex()
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		index = &imgspecv1.Index{
+			Versioned: imgspec.Versioned{
+				SchemaVersion: 2,
+			},
+		}
 	}
-	return &ociImageDestination{ref: ref, index: index}, nil
+
+	return &ociImageDestination{ref: ref, index: *index}, nil
 }
 
 // Reference returns the reference used to set up this destination.  Note that this should directly correspond to user's intent,
@@ -191,23 +202,20 @@ func (d *ociImageDestination) PutManifest(m []byte) error {
 		Architecture: runtime.GOARCH,
 		OS:           runtime.GOOS,
 	}
-	d.index.Manifests = append(d.index.Manifests, desc)
+	d.addManifest(&desc)
 
 	return nil
 }
 
-func ensureDirectoryExists(path string) error {
-	if _, err := os.Stat(path); err != nil && os.IsNotExist(err) {
-		if err := os.MkdirAll(path, 0755); err != nil {
-			return err
+func (d *ociImageDestination) addManifest(desc *imgspecv1.Descriptor) {
+	for i, manifest := range d.index.Manifests {
+		if manifest.Annotations["org.opencontainers.image.ref.name"] == desc.Annotations["org.opencontainers.image.ref.name"] {
+			// TODO Should there first be a cleanup based on the descriptor we are going to replace?
+			d.index.Manifests[i] = *desc
+			return
 		}
 	}
-	return nil
-}
-
-// ensureParentDirectoryExists ensures the parent of the supplied path exists.
-func ensureParentDirectoryExists(path string) error {
-	return ensureDirectoryExists(filepath.Dir(path))
+	d.index.Manifests = append(d.index.Manifests, *desc)
 }
 
 func (d *ociImageDestination) PutSignatures(signatures [][]byte) error {
@@ -230,4 +238,31 @@ func (d *ociImageDestination) Commit() error {
 		return err
 	}
 	return ioutil.WriteFile(d.ref.indexPath(), indexJSON, 0644)
+}
+
+func ensureDirectoryExists(path string) error {
+	if _, err := os.Stat(path); err != nil && os.IsNotExist(err) {
+		if err := os.MkdirAll(path, 0755); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// ensureParentDirectoryExists ensures the parent of the supplied path exists.
+func ensureParentDirectoryExists(path string) error {
+	return ensureDirectoryExists(filepath.Dir(path))
+}
+
+// indexExists checks whether the index location specified in the OCI reference exists.
+// The implementation is opinionated, since in case of unexpected errors false is returned
+func indexExists(ref ociReference) bool {
+	_, err := os.Stat(ref.indexPath())
+	if err == nil {
+		return true
+	}
+	if os.IsNotExist(err) {
+		return false
+	}
+	return true
 }
--- a/oci/layout/oci_dest_test.go
+++ b/oci/layout/oci_dest_test.go
@@ -8,6 +8,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"path/filepath"
 )
 
 // readerFromFunc allows implementing Reader by any function, e.g. a closure.
@@ -58,4 +59,65 @@ func TestPutBlobDigestFailure(t *testing.T) {
 	_, err = os.Lstat(blobPath)
 	require.Error(t, err)
 	require.True(t, os.IsNotExist(err))
+}
+
+// TestPutManifestAppendsToExistingManifest tests that new manifests are getting added to existing index.
+func TestPutManifestAppendsToExistingManifest(t *testing.T) {
+	ref, tmpDir := refToTempOCI(t)
+	defer os.RemoveAll(tmpDir)
+
+	ociRef, ok := ref.(ociReference)
+	require.True(t, ok)
+
+	// iniitally we have one manifest
+	index, err := ociRef.getIndex()
+	assert.NoError(t, err)
+	assert.Equal(t, 1, len(index.Manifests), "Unexpected number of manifests")
+
+	// create a new test reference
+	ociRef2, err := NewReference(tmpDir, "new-image")
+	assert.NoError(t, err)
+
+	putTestManifest(t, ociRef2.(ociReference), tmpDir)
+
+	index, err = ociRef.getIndex()
+	assert.NoError(t, err)
+	assert.Equal(t, 2, len(index.Manifests), "Unexpected number of manifests")
+}
+
+// TestPutManifestTwice tests that existing manifest gets updated and not appended.
+func TestPutManifestTwice(t *testing.T) {
+	ref, tmpDir := refToTempOCI(t)
+	defer os.RemoveAll(tmpDir)
+
+	ociRef, ok := ref.(ociReference)
+	require.True(t, ok)
+
+	putTestManifest(t, ociRef, tmpDir)
+	putTestManifest(t, ociRef, tmpDir)
+
+	index, err := ociRef.getIndex()
+	assert.NoError(t, err)
+	assert.Equal(t, 1, len(index.Manifests), "Unexpected number of manifests")
+}
+
+func putTestManifest(t *testing.T, ociRef ociReference, tmpDir string) {
+	imageDest, err := newImageDestination(ociRef)
+	assert.NoError(t, err)
+
+	data := []byte("abc")
+	err = imageDest.PutManifest(data)
+	assert.NoError(t, err)
+
+	err = imageDest.Commit()
+	assert.NoError(t, err)
+
+	paths := []string{}
+	filepath.Walk(tmpDir, func(path string, info os.FileInfo, err error) error {
+		paths = append(paths, path)
+		return nil
+	})
+
+	digest := "ba7816bf8f01cfea414140de5dae2223b00361a396177a9cb410ff61f20015ad"
+	assert.Contains(t, paths, filepath.Join(tmpDir, "blobs", "sha256", digest), "The OCI directory does not contain the new manifest data")
 }
--- a/oci/layout/oci_transport.go
+++ b/oci/layout/oci_transport.go
@@ -189,14 +189,25 @@ func (ref ociReference) NewImage(ctx *types.SystemContext) (types.Image, error) 
 	return image.FromSource(src)
 }
 
-func (ref ociReference) getManifestDescriptor() (imgspecv1.Descriptor, error) {
+// getIndex returns a pointer to the index references by this ociReference. If an error occurs opening an index nil is returned together
+// with an error.
+func (ref ociReference) getIndex() (*imgspecv1.Index, error) {
 	indexJSON, err := os.Open(ref.indexPath())
 	if err != nil {
-		return imgspecv1.Descriptor{}, err
+		return nil, err
 	}
 	defer indexJSON.Close()
-	index := imgspecv1.Index{}
-	if err := json.NewDecoder(indexJSON).Decode(&index); err != nil {
+
+	index := &imgspecv1.Index{}
+	if err := json.NewDecoder(indexJSON).Decode(index); err != nil {
+		return nil, err
+	}
+	return index, nil
+}
+
+func (ref ociReference) getManifestDescriptor() (imgspecv1.Descriptor, error) {
+	index, err := ref.getIndex()
+	if err != nil {
 		return imgspecv1.Descriptor{}, err
 	}
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,9 @@
 vendor
 tools.timestamp
+
+# Idea IDE
+*.iml
+.idea
+
+# Visual Studio Code
+.vscode/*


oci/layout/oci_transport_test.go
```diff
@@ -192,31 +199,37 @@ func TestReferencePolicyConfigurationIdentity(t *testing.T) {
...
-		ref, err := NewReference(path, "sometag")
-		require.NoError(t, err)
-		ns := ref.PolicyConfigurationNamespaces()
-		require.NotNil(t, ns)
-		assert.Equal(t, []string{"/etc/skel", "/etc"}, ns)
-	}
+	assert.Equal(t, fullyExplicitPath, ns[0])
+	assert.Equal(t, filepath.Dir(fullyExplicitPath), ns[1])
+
+	//// Test with a known path which should exist. Test just one non-canonical
```
hferentschik: This test does not work on macOS. I am not sure whether there is a directory which could be used across platform to make this work.

mtrmac: Yeah, this is a repeated problem with the tests I have also noticed recently. Maybe we can replace `$TMPDIR` with the expanded version of it (i.e. ending up with `/private/var/`… instead of `/var/`… IIRC); that should be harmless for any other uses of `$TMPDIR`, shouldn’t it?

oci/layout/oci_dest.go
```diff
@@ -184,23 +193,20 @@ func (d *ociImageDestination) PutManifest(m []byte) error {
...
 }
 
-func ensureDirectoryExists(path string) error {
-	if _, err := os.Stat(path); err != nil && os.IsNotExist(err) {
-		if err := os.MkdirAll(path, 0755); err != nil {
-			return err
+func (d *ociImageDestination) addManifest(desc *imgspecv1.Descriptor){
+	for i, manifest := range d.index.Manifests {
+		if manifest.Digest.String() == desc.Digest.String() {
+			// TODO Should there first be a cleanup based on the descriptor we are going to replace?
```
mtrmac: I guess this PR is still a clear improvement without the cleanup; but having a long-term TODO note here is nice.

oci/layout/oci_dest.go
```diff
@@ -184,23 +193,20 @@ func (d *ociImageDestination) PutManifest(m []byte) error {
...
 	return nil
 }
 
-func ensureDirectoryExists(path string) error {
-	if _, err := os.Stat(path); err != nil && os.IsNotExist(err) {
-		if err := os.MkdirAll(path, 0755); err != nil {
-			return err
+func (d *ociImageDestination) addManifest(desc *imgspecv1.Descriptor){
+	for i, manifest := range d.index.Manifests {
+		if manifest.Digest.String() == desc.Digest.String() {
```
mtrmac: Should this really be overwriting based on the manifest digest? If the manifest digest is the same, then the image by definition hasn’t changed (well, apart from the external metadata).

My initial impression is that this should replace images with the same `org.opencontainers.image.ref.name` annotation, if any; does that make sense?

oci/layout/oci_dest.go
```diff
@@ -224,3 +230,30 @@ func (d *ociImageDestination) Commit() error {
...
+	return nil
+}
+
+// ensureParentDirectoryExists ensures the parent of the supplied path exists.
+func ensureParentDirectoryExists(path string) error {
+	return ensureDirectoryExists(filepath.Dir(path))
+}
+
+// Exists checks whether the specified path exists.
+// The implementation is opinionated, since in case of unexpected errors false is returned
```
mtrmac: (I’m not sure this is worth it; if `os.Stat` fails, then any other use of that directory will likely failing as well, and in that case we might just as well return the first encountered error (before the filesystem gets completely confused).

OTOH a separate `exists` function only simplifies things because it returns a simple `bool`, otherwise the caller handling the three states of `exists` might just as well call `os.Stat` itself, and handle the same three states. So I guess this is fine.)

oci/layout/oci_dest.go
```diff
@@ -224,3 +230,30 @@ func (d *ociImageDestination) Commit() error {
...
+func exists(path string) bool {
```
mtrmac: Please name this `pathExists`, `dirExists` or something like that, though.

oci/layout/oci_dest_test.go
```diff
@@ -59,3 +60,59 @@ func TestPutBlobDigestFailure(t *testing.T) {
...
+	assert.NoError(t, err)
+
+	paths := []string{}
+	filepath.Walk(tmpDir, func(path string, info os.FileInfo, err error) error {
+		paths = append(paths, path)
+		return nil
+	})
+
+	digest := "ba7816bf8f01cfea414140de5dae2223b00361a396177a9cb410ff61f20015ad"
+	assert.Contains(t, paths, filepath.Join(tmpDir, "blobs", "sha256", digest), "The OCI directory does not contain the new manifest data")
```
mtrmac: (if this is checking for a single path with a known name, the code could just as well use the `exists` function and check for that specific name. But this is only a test, and it works, so whatever.)

oci/layout/oci_transport.go
```diff
@@ -186,14 +188,28 @@ func (ref ociReference) NewImage(ctx *types.SystemContext) (types.Image, error)
...
+		Versioned: imgspec.Versioned{
+			SchemaVersion: 2,
+		},
+	}
+
 	indexJSON, err := os.Open(ref.indexPath())
+	defer indexJSON.Close()
 	if err != nil {
-		return imgspecv1.Descriptor{}, err
+		return index, err
```
mtrmac: I don’t think this should silently swallow I/O errors and return an empty index in that case as if everything was just fine and any pre-existing images inexplicably disappeared.

oci/layout/oci_transport_test.go
mtrmac: … Oops, I didn’t read the related code before answering, sorry.

I hope we can find _something_, I can take a look at least for macOS and Linux. If someone wanted to run this on Windows, I have no idea.

The `$TMPDIR` issue is real, and I’d prefer a general solution to that instead of every test having magic knowledge of that like `fullyExplicitPath, err := explicitfilepath.ResolvePathToFullyExplicit(tmpDir)`.  But that’s strictly speaking unrelated to this PR.

oci/layout/oci_transport_test.go
```diff
@@ -176,12 +178,17 @@ func TestReferencePolicyConfigurationIdentity(t *testing.T) {
 	ref, tmpDir := refToTempOCI(t)
 	defer os.RemoveAll(tmpDir)
 
-	assert.Equal(t, tmpDir+":tagValue", ref.PolicyConfigurationIdentity())
+	fullyExplicitPath, err := explicitfilepath.ResolvePathToFullyExplicit(tmpDir)
```
mtrmac: I do know that this is needed for macOS right now, but this is a bit weird to have in tests. This workaround gets copied all over the place…

At the very least, a comment explaining the `/private/var` situation would help make sure a next Linux-only maintainer does not break everything again.

Can we instead add a shared workaround, something like
```go
func init() {
// comment about what this is for
  os.Setenv("TMPDIR", filepath.EvalSymlinks(os.Getenv("TMPDIR")))
}
```
or something like that? Perhaps even a helper function, IIRC there are tests in multiple subpackages which need a similar workaround.

oci/layout/oci_dest.go
```diff
@@ -224,3 +230,30 @@ func (d *ociImageDestination) Commit() error {
...
+	_, err := os.Stat(path)
+	if err == nil {
+		return true
+	}
+	if os.IsNotExist(err) {
+		return false
+	}
+	return true
```
cyphar: This can be implemented as

```
func exists(path string) bool {
    _, err := os.Stat(path)
    return err != nil && os.IsNotExist(err)
}
```

I also think that you could even remove the `err != nil` check (`os.IsNotExist(nil) == false`).

oci/layout/oci_dest.go
```diff
@@ -224,3 +230,30 @@ func (d *ociImageDestination) Commit() error {
 	}
 	return ioutil.WriteFile(d.ref.indexPath(), indexJSON, 0644)
 }
+
+func ensureDirectoryExists(path string) error {
+	if _, err := os.Stat(path); err != nil && os.IsNotExist(err) {
```
cyphar: Shouldn't this be using the `exist` function you wrote?

oci/layout/oci_dest.go
cyphar: Yeah, you're right `mtrmac`. When you update an OCI image's "tag" you need to remove any existing descriptors in the manifest with the same `org.opencontainers.image.ref.name` and replace it with the new one (with that tag added). I already do this in `umoci` (with some more complicated handling because in `umoci` we mutate blobs which requires reconstructing the merkle tree).

hferentschik: Yes, makes sense. In fact when trying some of these things out now in Minishift (where we want to use this), I realized the problem. In fact I have a couple of questions around this. 

As said, my aim is to store multiple images within the same OCI directory with the aim to share layers. Of course I need to also then be able to retrieve these images. And here is actually the rub. On a lower level, this pull request allows me to store multiple manifests to the index, but I realized now, that they won't be addressable. I think there are several reasons for that. 

Let's quickly look at some example index.json which gets created:

```json
{
   "schemaVersion":2,
   "manifests":[
      {
         "mediaType":"application/vnd.oci.image.manifest.v1+json",
         "digest":"sha256:bb1fb6e9d0509afa26e52b75bdf144faea07d2473c9d5c693427b6f09b6c3ba5",
         "size":823,
         "annotations":{
            "org.opencontainers.image.ref.name":"v3.6.0"
         },
         "platform":{
            "architecture":"amd64",
            "os":"darwin"
         }
      }
   ]
}
```

Notice, org.opencontainers.image.ref.name contains the tag of the image as it's name. Really I would expect the image name, eg openshift/origin as name and org.opencontainers.image.version should be used for the tag/version. Storing the tag in org.opencontainers.image.ref.name might work for the current single directory per image scenario, but I would think even there it is questionable. 

I think the problem comes down to the ociReference struct (https://github.com/containers/image/blob/master/oci/layout/oci_transport.go#L79), as well as how an oci transport string is parsed (https://github.com/containers/image/blob/master/oci/layout/oci_transport.go#L92). Atm OCI specifies the target directory and a tag. Somehow the image name needs to get into this. Does this make sense? If so, I am wondering what the format should be for an oci reference then?

oci/layout/oci_dest.go
hferentschik: fair enough

oci/layout/oci_dest.go
hferentschik: it should

oci/layout/oci_dest.go
hferentschik: ok

oci/layout/oci_dest.go
hferentschik: sure, I am just not sure whether more condensed code is always the better option. One needs to do much more logical operations in ones head in this version. My version is more verbose, but arguably easier to take in.

oci/layout/oci_transport.go
hferentschik: I was thinking about this one myself. So you reckon it should be nil, err?

oci/layout/oci_transport_test.go
hferentschik: > I do know that this is needed for macOS right now, but this is a bit weird to have in tests.

I know. I am also not quite sure what the best way to handle this. Unfortunately, on macOS thinks like /var /tmp and even /etc are symbolic links, so a lot of tests fail. 

> At the very least, a comment explaining the /private/var situation would help make sure a next Linux-only maintainer does not break everything again.

+1

> Can we instead add a shared workaround, something like ...

Ahhh, that's an interesting one. Let me think about this one for a bit.

oci/layout/oci_dest.go
hferentschik: I think it comes down to how one would add the image name into the oci ref? How for example would one split between oci directory and the actual image name ,eg "oci:<path to oci basedir>/<image name>:<tag>. Using '/' is problematic. How would you parse that?

hferentschik: I am happy to continue working on this, but do we agree that the way a oci reference is structured needs to be extended?

cyphar: The image name should not be part of the reference. The reference is the "tag". It doesn't make sense to include the path to the image, because you can move the image to a different path and because there's no cryptographic verification of image references.

Please, just use the current scheme. We had so many issues with references inside the OCI spec. I appreciate that the current spec doesn't forbid you from making a decision like that, I seriously would not recommend it.

hferentschik: > The image name should not be part of the reference.

But how do I then reference multiple images within the same OCI directory? I might be missing something here. Or are you suggesting that I create my own custom tag (eg combining image name with a version)?

> doesn't make sense to include the path to the image, because you can move the image 

This might be a misunderstanding. With path I mean the directory of where the OCI index lives. Then in order to push multiple images into the same index, I need to reference these images. Looking at ParseReference in oci_transport.go, tag seems to be used as the version of the image. It does not contain any name component.

So how do you suggest I use the current current _ociReference_ to store multiple images into the same OCI directory?

hferentschik: BTW, I guess this is what I am after - https://github.com/containers/image/pull/318/files

cyphar: I think #318 is not doing the right thing.

> But how do I then reference multiple images within the same OCI directory?

Use different tags, I'm not sure I understand what you mean.

> With path I mean the directory of where the OCI index lives.

There's only one OCI index at the root `index.json` (relative to the path where the OCI layout exists). I'd recommend taking a look at https://github.com/openSUSE/umoci's interface to see what I mean (disclosure, I wrote it).

hferentschik: > I think #318 is not doing the right thing.

I was just saying that it tries to address the same issue. I did not look into the exact details.

> Use different tags, I'm not sure I understand what you mean.

Let's look at a concrete example. Say I have busybox:1.27.7 and alpine:3.1 and I want to push them  into an OCI index under _/tmp/oci_

Currently I would do something like this:

    srcRef, err := alltransports.ParseImageName("oci:/tmp/oci")

In this case the org.opencontainers.image.ref.name becomes 'latest'  - https://github.com/containers/image/blob/master/oci/layout/oci_transport.go#L97

So what options do I have? 

     srcRef, err := alltransports.ParseImageName("oci:/tmp/oci:alpine:3.1")

does not work (at least not in the current impl). I could 

    srcRef, err := alltransports.ParseImageName("oci:/tmp/oci:alpine@3.1")

In this case I create my own 'tag' by artificially combining the image name and version. Is this what I should do? This would lead to a org.opencontainers.image.ref.name of alpine@3.1. Should it not be org.opencontainers.image.ref.name = alpine and org.opencontainers.image.version = 3.1? That's what I would expect looking at the OCI specs.

What if the registry matters. Say I have a custom alpine image in myrepo - myrepo/alpine:4. How do I do this? Are you suggesting something like:

    srcRef, err := alltransports.ParseImageName("oci:/tmp/oci:myrepo@alpine@3.1")

I feel it should be possible to something along the line:

    srcRef, err := alltransports.ParseImageName("oci:/tmp/oci@myrepo/alpine:3.1")

This allows me to keep the the well known notation <repo>/<image>:<version>.

There are multiple levels to this:

* From an containers/image API point of view and the need to create src and dest references based on strings, how do I denote the concepts of base directory, image name and version
* The second aspect is the concrete implementation. Right now it assumes that there is only one image per OCI directory. Depending on the outcome of the first question, the syntax and parsing of an OCI reference needs to be changed
* Lastly there is a question of spec interpretation. I would expect that if I want to push alpine:3.1 into an OCI index, the top level entry in the index should use 'org.opencontainers.image.ref.name' as name for the image, aka 'alpine' and 'org.opencontainers.image.ref.version' as version, aka 3.1. 

Does this make sense?

 > I'd recommend taking a look at https://github.com/openSUSE/umoci's interface to see what I mean (disclosure, I wrote it).

The link does not seem to work. Can you point to the actual interface?

oci/layout/oci_transport.go
mtrmac: Ah, it is actually returning a non-`nil` error value, I didn’t notice, sorry.

Anyway, yes, when the only caller ignores the index if `err != nil`, this function might as well return `nil, err` instead of pretending that returning `index` matters.

oci/layout/oci_dest.go
mtrmac: > BTW, I guess this is what I am after - https://github.com/containers/image/pull/318/files

Yes, this PR and #318 complement each other; #318 handles the permitted names and the read side, this one handles the write side.

mtrmac: @cyphar 

> The image name should not be part of the reference. The reference is the "tag". It doesn't make sense to include the path to the image, because you can move the image to a different path and because there's no cryptographic verification of image references.

Maybe we are just using different terminology? In the context of containers/image, a “reference” is most likely a `types.ImageReference`, a (local-process-only, not necessarily portable to any other environment) pointer to any kind of _single_ image; in this particular case, `ociReference` is a (filesystem directory, `org.opencontainers.image.ref.name` annotation value) pair referring to an image within an index in an on-filesystem OCI-formatted directory. The annotation value is usually called a “tag” or an “image” in the current PRs. The filesystem directory path is an obviously necessary part of `types.ImageReference`, the callers must point to a directory _somehow_, but it is in no way embedded into the contents of the OCI repo.

>
> Please, just use the current scheme. We had so many issues with references inside the OCI spec. I appreciate that the current spec doesn't forbid you from making a decision like that, I seriously would not recommend it.

What do you mean by “the current scheme” / “a decision like that”?

>I think #318 is not doing the right thing.

In what way? AFAICS all it does WRT naming is that it expands the grammar of allowed …`.name` annotation names from a tag-like regexp to the regexp in the OCI standard, without any conversions or normalizations. It is not opinionated in any way, it just allows callers to use the OCI standard as it is written.

(and, sure, with the OCI standard saying nothing about the semantics of the image names, maybe there could be interoperability problems / mismatches in expectation; But that seems to be _also_ to be an argument for containers/image just exposing the raw data and letting the caller deal with whatever the environment and semantics is for that particular image.)

mtrmac: @hferentschik 
> > I'd recommend taking a look at https://github.com/openSUSE/umoci's interface to see what I mean (disclosure, I wrote it).
>
> The link does not seem to work. Can you point to the actual interface?

Try https://github.com/openSUSE/umoci .

oci/layout/oci_transport_test.go
mtrmac: Actually, after the `/var/tmp` vs. `/private/var/tmp` ambiguity is resolved somehow, this test can just as well use `ioutil.TempDir(…)` instead of relying on a pre-existing directory in the OS.

oci/layout/oci_dest.go
cyphar: @hferentschik It looks like GitHub screwed up the link. The exact link is https://github.com/openSUSE/umoci/blob/master/oci/casext/refname.go -- which contains all of the umoci's reference handling code (including resolution and replacement of references).

hferentschik: > It looks like GitHub screwed up the link. The exact link is https://github.com/openSUSE/umoci/blob/master/oci/casext/refname.go -- which contains all of the umoci's reference handling code (including resolution and replacement of references).

Ok. Looking just at this file it is hard to tell how it is used though. Can you post the top level API call a user of this library would make to put say alpine:3.1 into the index?

mtrmac: [See also #323 for discussion on the `oci:`(path, annotation) syntax.)

mtrmac: @cyphar 

> Lastly there is a question of spec interpretation. I would expect that if I want to push alpine:3.1 into an OCI index, the top level entry in the index should use 'org.opencontainers.image.ref.name' as name for the image, aka 'alpine' and 'org.opencontainers.image.ref.version' as version, aka 3.1.

This is actually a reasonable point, and AFAICT the OCI spec is silent on this. Is it reasonable / interoperable with umoci to expect users to set …`.ref.name` to unique values (i.e. embed Docker-like version tags in there), or should every consumer be matching on a pair of `.ref.name`+`.ref.version`, or perhaps on even more annotations, because single annotations or subsets of that are not expected to be unique?

cyphar: @hferentschik What do you mean by "put alpine:3.1" in the index? Do you mean a refname of `"alpine:3.1"`? If so, you'd do something like `(casext.Engine).AddReferences(context.Background(), "alpine:3.1", <descriptor that points to the manifest>)`. But I'm not fully sure what you're asking to be honest.

cyphar: @mtrmac 

> > Lastly there is a question of spec interpretation. I would expect that if I want to push alpine:3.1 into an OCI index, the top level entry in the index should use 'org.opencontainers.image.ref.name' as name for the image, aka 'alpine' and 'org.opencontainers.image.ref.version' as version, aka 3.1.

> This is actually a reasonable point, and AFAICT the OCI spec is silent on this. Is it reasonable / interoperable with umoci to expect users to set ….ref.name to unique values (i.e. embed Docker-like version tags in there), or should every consumer be matching on a pair of .ref.name+.ref.version, or perhaps on even more annotations, because single annotations or subsets of that are not expected to be unique?

To be interoperable with `umoci`, `umoci` only looks at `org.opencontainers.image.ref.name`. It internally does support handling of multiple entries with the same name, but the UX doesn't support it (and so it does some not-very-nice things). As for the general concept, I'm not really sure which way is better. While I understand why you'd want to use different `org.opencontainers.image.ref.*` things, I would be careful about adding crucial information to annotations that other implementations might not know about. Do I think this is something that will need to be refined in future OCI updates? Most definitely, the arguments we had about issues like this before 1.0 are pages long.

Also, if you want to inter-operate with Docker (you'd have to ask @runcom to explain how the current implementation is going) then it's likely that you'll have to use `ref.name` only.

mtrmac: Great, `.ref.name`-only it is, then.

I was worried about having to invent some SQL-like query language, that would be a pretty bad UX (and a very horrible UX if every OCI consumer defined its own).

oci/layout/oci_transport_test.go
mtrmac: The `/etc/skel` problem should have been …resolved… by #336.

oci/layout/oci_transport_test.go
mtrmac: This should have been resolved by #336.

hallyn:
Hi, just wanted to check - is this still being worked on?

hferentschik:
> Hi, just wanted to check - is this still being worked on?

It is. In fact I just rebased the changes on top of master. I would love to get this integrated as well.

oci/archive/oci_transport.go
```diff
@@ -194,7 +194,7 @@ func (t *tempDirOCIRef) deleteTempDir() error {
 
 // createOCIRef creates the oci reference of the image
 func createOCIRef(image string) (tempDirOCIRef, error) {
-	dir, err := ioutil.TempDir("/var/tmp", "oci")
+	dir, err := ioutil.TempDir("", "oci")
```
mtrmac: Why does this PR need this change? It seems unrelated.

(Using `/var/tmp` on is necessary on Linux because `/tmp` may be `tmpfs`, inappropriate for storing multi-gigabyte container images.)

oci/layout/oci_dest.go
```diff
@@ -27,11 +27,21 @@ func newImageDestination(ref ociReference) (types.ImageDestination, error) {
...
 		return nil, errors.Errorf("cannot save image with empty image.ref.name")
 	}
+
 	index := imgspecv1.Index{
 		Versioned: imgspec.Versioned{
 			SchemaVersion: 2,
 		},
 	}
+
+	if indexExists(ref.dir) {
```
mtrmac: This would be cleaner as `if indexExists() { index = … } else { index = … }`, rather than always setting `index` only to overwrite it.

oci/layout/oci_dest.go
```diff
@@ -191,23 +201,20 @@ func (d *ociImageDestination) PutManifest(m []byte) error {
...
 	return nil
 }
 
-func ensureDirectoryExists(path string) error {
-	if _, err := os.Stat(path); err != nil && os.IsNotExist(err) {
-		if err := os.MkdirAll(path, 0755); err != nil {
-			return err
+func (d *ociImageDestination) addManifest(desc *imgspecv1.Descriptor) {
+	for i, manifest := range d.index.Manifests {
+		if manifest.Digest.String() == desc.Digest.String() {
```
mtrmac: The replacement needs to be based on the `org.opencontainers.image.ref.name` annotation, not on the digest; see the earlier conversation in the PR.

oci/layout/oci_dest.go
```diff
@@ -231,3 +238,30 @@ func (d *ociImageDestination) Commit() error {
...
+
+// ensureParentDirectoryExists ensures the parent of the supplied path exists.
+func ensureParentDirectoryExists(path string) error {
+	return ensureDirectoryExists(filepath.Dir(path))
+}
+
+// Exists checks whether an index file exists at the specified path.
+// The implementation is opinionated, since in case of unexpected errors false is returned
+func indexExists(path string) bool {
+	_, err := os.Stat(filepath.Join(path, "index.json"))
```
mtrmac: This should call the existing `ociReference.indexPath`, at least to make it trivial to find all uses of that path.

oci/layout/oci_transport.go
```diff
@@ -14,7 +14,9 @@ import (
 	"github.com/containers/image/transports"
 	"github.com/containers/image/types"
 	"github.com/opencontainers/go-digest"
+	imgspec "github.com/opencontainers/image-spec/specs-go"
 	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
+
```
mtrmac: (Nit: Please drop this line.)

oci/layout/oci_transport.go
mtrmac: Still outstanding.

oci/layout/oci_transport.go
```diff
@@ -189,14 +191,28 @@ func (ref ociReference) NewImage(ctx *types.SystemContext) (types.Image, error)
...
-func (ref ociReference) getManifestDescriptor() (imgspecv1.Descriptor, error) {
+func (ref ociReference) getIndex() (imgspecv1.Index, error) {
+	index := imgspecv1.Index{
+		Versioned: imgspec.Versioned{
+			SchemaVersion: 2,
+		},
+	}
+
 	indexJSON, err := os.Open(ref.indexPath())
+	defer indexJSON.Close()
```
mtrmac: Moving the `defer` up is incorrect; `indexJSON` may be `nil` at this point.

oci/layout/oci_transport.go
```diff
@@ -189,14 +191,28 @@ func (ref ociReference) NewImage(ctx *types.SystemContext) (types.Image, error)
 	return image.FromSource(src)
 }
 
-func (ref ociReference) getManifestDescriptor() (imgspecv1.Descriptor, error) {
+func (ref ociReference) getIndex() (imgspecv1.Index, error) {
+	index := imgspecv1.Index{
```
mtrmac: (After the return values are modified to return `nil, err`, this should just an empty value.)

oci/archive/oci_transport.go
hferentschik: This is indeed unrelated, however, I came across this while rebasing my original pull request. OCI is one of the few transports which would work across OS. /var/tmp will of course not work on Windows. IMO the basedir should not be specified, but I can revert this, if that means we can get this merged.

oci/layout/oci_dest.go
hferentschik: fair enough

oci/layout/oci_dest.go
hferentschik: ok

oci/layout/oci_dest.go
hferentschik: Not sure what you mean here

oci/layout/oci_transport.go
hferentschik: which one exactly?

oci/layout/oci_transport.go
hferentschik: true

oci/layout/oci_transport.go
hferentschik: ?

oci/layout/oci_transport.go
hferentschik: Ok, I'll return a *imgspecv1.Index. Not sure whether this is so much better though.

oci/layout/oci_transport.go
hferentschik: ahh,  I guess this relates to returning *imgspecv1.Index and nil in the case an error occurs. Fair enough

hferentschik:
@mtrmac - I pushed an update. I addressed your comments, except one which I am not quite clear about what it means. If it is still an issue in please clarify what you mean.

> Also see the test suite failures – locally, make should be passing for you.

I have no idea what the dangling whitespace error was in the commit. I had to drop the commit and re-create it. Seems it is gone now.

oci/archive/oci_transport.go
mtrmac: Yes, please split this out. Having as much as possible work on Windows would be very nice, but the fix should

- Be in a clearly separate commit, with an explanation of what is fixing. (At least I, with my Linux blinders on, could not guess that this improves Windows.)
- Not break Linux/macOS to fix Windows.
- Ideally, be comprehensive. See the other two instances of `temporaryDirectoryForBigFiles`; I guess this would end up as a tiny sub-package.

oci/layout/oci_transport.go
```diff
@@ -189,14 +190,29 @@ func (ref ociReference) NewImage(ctx *types.SystemContext) (types.Image, error)
...
 	if err != nil {
-		return imgspecv1.Descriptor{}, err
+		return nil, err
 	}
 	defer indexJSON.Close()
-	index := imgspecv1.Index{}
-	if err := json.NewDecoder(indexJSON).Decode(&index); err != nil {
+
+	index := &imgspecv1.Index{
+		Versioned: imgspec.Versioned{
```
mtrmac: Why is the `Versioned` field initialized? Shouldn’t that come from `indexJSON` as well?

oci/layout/oci_transport.go
mtrmac: (Returning an `imgspecv1.Index{}` would be fine as well; the point is that the common case is that all other return values are ignored on a non-`nil` `error` return, and this makes a misleading impression that the return value matters.)

oci/layout/oci_transport.go
mtrmac: (Yes, exactly as you did.)

oci/layout/oci_transport.go
mtrmac: This refers to https://github.com/containers/image/pull/326#discussion_r146985778 .

oci/layout/oci_dest.go
mtrmac:
```go
func (ref ociReference) indexExists() bool {
    _, err := os.Stat(ref.indexPath())
   …
```

mtrmac:
> I have no idea what the dangling whitespace error was in the commit.

A space/tab at the end of the line; most editors etc. don’t highlight them by default, and formatters like `gofmt` drop them, so they may seem to ”surprisingly” disappear.  IIRC the error message indicates the line, but I might be wrong. At least `git show` etc. highlight them in red.

oci/archive/oci_transport.go
hferentschik: @mtrmac - I created a new issue - https://github.com/containers/image/issues/368. I agree this should be properly tracked since it is not directly related to this pull request. 

Not sure whether I have the bandwidth to work on this, but either way I would like to see my others pull requests merged/addressed first. We need them merged in order to start using containers/image in Minishift.

mtrmac: > @mtrmac - I created a new issue - #368.

Thanks!

oci/layout/oci_dest.go
hferentschik: ahh :-)

mtrmac:
:+1:

Thanks!

runcom:
lgtm

runcom:
mmm rebase?

hferentschik:
> mmm rebase?

done

runcom:
waiting Travis to merge

hferentschik:
Awesome, thanks for the merge

PR #330, opened 2017-08-18
Support annotations on BlobInfos

williammartin:
According to the image spec [1], descriptors should be able to contain
annotations. Support them on the returned BlobInfo.

[1] https://github.com/opencontainers/image-spec/blob/fc936c78346d017dcdbc6dd220a8ff60675ce3b5/descriptor.md

Signed-off-by: Will Martin <wmartin@pivotal.io>

### DIFF
--- a/image/oci_test.go
+++ b/image/oci_test.go
@@ -29,11 +29,16 @@ func manifestOCI1FromFixture(t *testing.T, src types.ImageSource, fixture string
 }
 
 func manifestOCI1FromComponentsLikeFixture(configBlob []byte) genericManifest {
-	return manifestOCI1FromComponents(descriptorOCI1{descriptor: descriptor{
-		MediaType: imgspecv1.MediaTypeImageConfig,
-		Size:      5940,
-		Digest:    "sha256:9ca4bda0a6b3727a6ffcc43e981cad0f24e2ec79d338f6ba325b4dfd0756fb8f",
-	}}, nil, configBlob, []descriptorOCI1{
+	return manifestOCI1FromComponents(descriptorOCI1{
+		descriptor: descriptor{
+			MediaType: imgspecv1.MediaTypeImageConfig,
+			Size:      5940,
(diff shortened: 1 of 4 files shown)


tscolari:
To add a bit more of context, this is important for the [Grootfs](https://github.com/cloudfoundry/grootfs) project. We're pushing forward a feature to support buildpacks based on OCI images, and we're using annotations to differentiate normal layers from the droplets.

williammartin:
@mtrmac @runcom Any thoughts on this?

image/oci_test.go
```diff
@@ -112,12 +122,14 @@ func TestManifestOCI1ManifestMIMEType(t *testing.T) {
 
 func TestManifestOCI1ConfigInfo(t *testing.T) {
 	for _, m := range []genericManifest{
-		manifestOCI1FromFixture(t, unusedImageSource{}, "oci1.json"),
```
mtrmac: Why is this removed?

mtrmac:
> I can’t resist feeling that this will inevitably become a mess

(Of course, now that the annotations are specified by OCI, we don’t have _that much_ choice: either expose the raw annotations, or tell users to use the OCI-specific libraries directly. The former is probably still a better choice, especially in the read-only form implemented by this PR.)

image/oci_test.go
williammartin: Amateur hour mistake. Thanks!

williammartin:
@mtrmac In general we agree, the annotations are bound to be abused. However, like you said, OCI has specified them and this PR is useful to us. I'm hopeful we'll be able to remove our own dependence on annotations at some point but we do depend on them now.

mtrmac:
:+1:

Thanks!

runcom:
LGTM

runcom:
(needs rebase)

spikymonkey:
Thanks - branch rebased

mtrmac:
Rebased and merged manually, per the conversation in #337.

PR #332, opened 2017-08-25
docker: ignore EPERM when scanning for certificates

### LINKED ISSUE #297: docker:// requires root now due to certs.d
I understand that `certs.d` is needed but it means that you cannot run `skopeo` as an unprivileged user anymore due to the readability of `/etc/docker/cert.d` by normal users. This was actually a fairly big selling point of `skopeo` over `docker pull` because requiring `root` to download files of the internet doesn't make any sense.

It looks like you might be able to configure `skopeo` to not require this anymore by creating your own `certs.d` which you then point `skopeo` at, but I'm not really sure what it should contain and I don't believe there's any docs on this.

Would it be possible to disable `certs.d` usage or something? This makes the value proposition of rootless containers using `skopeo` weaker unfortunately...

cyphar:
Ignore (but log) all permission errors when trying to get the list of
additionally trusted Docker certificates and keys. This should not
result in any security issues, because additional trusted certificates
are additive (and thus the only issue that can occur is that a transport
is not trusted by skopeo when it should be).

Not ignoring permission errors would cause issues when trying to perform
a copy of a Docker image on a machine that has Docker installed (by
default Docker makes /etc/docker/certs.d not accessible). This would
manifest itself like so:

  % ./skopeo copy docker://busybox oci:busybox
  FATA[0000] Error initializing source docker://busybox:latest: open /etc/docker/certs.d/docker.io: permission denied

Fixes #297

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -155,6 +155,10 @@ func setupCertificates(dir string, tlsc *tls.Config) error {
 		if os.IsNotExist(err) {
 			return nil
 		}
+		if os.IsPermission(err) {
+			logrus.Debugf("Skipping scan of %s due to permission error: %v", dir, err)
+			return nil
+		}
 		return err
 	}
 


mtrmac:
:+1:

Thanks for working on this!

runcom:
LGTM

PR #335, opened 2017-08-30
Resolve linting errors and fix make lint target

williammartin:
The version of golint in Travis is quietly failing because it doesn't
support multiple packages as [variadic args]. This fixes the makefile
by running golint against each package individually. It also resolves
the linting error surfaced from fixing the makefile.

[variadic args]: https://github.com/golang/lint/pull/271

Signed-off-by: Gareth Clay <gclay@pivotal.io>

You can see the failure I'm talking about in this build for example https://travis-ci.org/containers/image/jobs/269932656. Between .gitvalidation and the vendor target, golint errors with:

```
open github.com/containers/image: no such file or directory
open github.com/containers/image/copy: no such file or directory
open github.com/containers/image/directory: no such file or directory
open github.com/containers/image/directory/explicitfilepath: no such file or directory
open github.com/containers/image/docker: no such file or directory
open github.com/containers/image/docker/archive: no such file or directory
open github.com/containers/image/docker/daemon: no such file or directory
open github.com/containers/image/docker/policyconfiguration: no such file or directory
open github.com/containers/image/docker/reference: no such file or directory
open github.com/containers/image/docker/tarfile: no such file or directory
open github.com/containers/image/image: no such file or directory
open github.com/containers/image/manifest: no such file or directory
open github.com/containers/image/oci: no such file or directory
open github.com/containers/image/oci/layout: no such file or directory
open github.com/containers/image/openshift: no such file or directory
open github.com/containers/image/ostree: no such file or directory
open github.com/containers/image/pkg/compression: no such file or directory
open github.com/containers/image/pkg/strslice: no such file or directory
open github.com/containers/image/pkg/sysregistries: no such file or directory
open github.com/containers/image/signature: no such file or directory
open github.com/containers/image/storage: no such file or directory
open github.com/containers/image/transports: no such file or directory
open github.com/containers/image/transports/alltransports: no such file or directory
open github.com/containers/image/types: no such file or directory
open github.com/containers/image/version: no such file or directory
```

which is comparable to the issue opened here:

https://github.com/golang/lint/issues/270

I guess the version of golint in the zesty apt repository doesn't have this variadic package args functionality yet.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,11 +52,13 @@ validate: lint
 	@test -z "$$(gofmt -s -l . | grep -ve '^vendor' | tee /dev/stderr)"
 
 lint:
-	@out="$$(golint $(PACKAGES))"; \
-	if [ -n "$$out" ]; then \
-		echo "$$out"; \
-		exit 1; \
-	fi
+	@for package in $(PACKAGES); do \
+		out="$$(golint $$package)"; \
+		if [ -n "$$out" ]; then \
+			echo "$$out"; \
+			exit 1; \
+		fi \
+	done
 
 .PHONY: .gitvalidation
 
--- a/pkg/sysregistries/system_registries.go
+++ b/pkg/sysregistries/system_registries.go
@@ -51,7 +51,7 @@ func loadRegistryConf() (*tomlConfig, error) {
 	return config, err
 }
 
-// Returns an array of strings that contain the names
+// GetRegistries returns an array of strings that contain the names
 // of the registries as defined in the system-wide
 // registries file.  it returns an empty array if none are
 // defined
@@ -63,7 +63,7 @@ func GetRegistries() ([]string, error) {
 	return config.Registries.Search.Registries, nil
 }
 
-// Returns an array of strings that contain the names
+// GetInsecureRegistries returns an array of strings that contain the names
 // of the insecure registries as defined in the system-wide
 // registries file.  it returns an empty array if none are
 // defined


mtrmac:
Thanks, great catch.

Looking at the `Makefile`, though, `make tools` should, via `go get -u`, always get the latest version of `golint`, and `.travis.yml` should be invoking the `tools` target in the tests.  It would be a bit weird to have a `Makefile` workaround for an old version of `golint` when that same `Makefile` is installing a newer version.

Maybe (I didn’t test this) the `Makefile` change could be replaced by changing
```sh
PATH=$PATH:/gopath/bin
```
in `.travis.yml` to
```sh
PATH=/gopath/bin:$PATH
```
?

williammartin:
Ah, got a little confused by the travis dockerfile which installs golint from apt: https://github.com/containers/image/blob/master/.travis.Dockerfile

I guess with:

`PATH=/gopath/bin:$PATH`

We'd pick up the `go get` `golint` first, which seems good to me. Happy to fix up the PR to try that out.

mtrmac:
> Ah, got a little confused by the travis dockerfile which installs golint from apt

Right. Perhaps that can be dropped, so that we have only one way to get `golint`, at least for development and test automation.

williammartin:
Going to close this PR in favour of #337 which leaves the Makefile alone and just ensures that the tests use the correct version of golint in Travis.

PR #336, opened 2017-08-30
Add full macOS support

mtrmac:
With this,
```sh
brew install gpgme
make
```

should “just work”.

The changes are mostly trivial, do note that this introduces a first `internal` subpackage.


mtrmac:
@hferentschik this could make your life a bit easier.

mtrmac:
RFC: Now that the tag completely disables the `ostree` package, naming it `containers_image_ostree_stub` might not be the best fit. It’s not _wrong_ (as far as `transports/alltransports` goes at least), but `containers_image_disable_ostree` would be a bit more explicit.

I’m not overall sure that changing the tag name and breaking possible existing users is worth it, though. Opinions?

rhatdan:
I would leave it alone.  I would have preferred just having ostree, similar to how we have selinux and seccomp, but I now others are using it so why break it.

runcom:
LGTM

mtrmac:
:+1:

PR #337, opened 2017-08-31
Resolve linting errors and use recent golint

williammartin:
The version of golint in Travis is quietly failing because it doesn't
support multiple packages as variadic args. This removes the version of
golint being fetched from apt in favour of that built by the tools make
target. It also resolves the linting error surfaced from fixing the makefile.

Signed-off-by: Gareth Clay <gclay@pivotal.io>

This is a replacement for the PR #335 which resolved this issue in a different way.

### DIFF
--- a/.travis.Dockerfile
+++ b/.travis.Dockerfile
@@ -1,4 +1,4 @@
 FROM ubuntu:zesty
 
 RUN apt-get -qq update && \
-    apt-get install -y sudo docker.io git make golang golint btrfs-tools libdevmapper-dev libgpgme-dev libostree-dev
+    apt-get install -y sudo docker.io git make golang btrfs-tools libdevmapper-dev libgpgme-dev libostree-dev
--- a/pkg/sysregistries/system_registries.go
+++ b/pkg/sysregistries/system_registries.go
@@ -51,7 +51,7 @@ func loadRegistryConf() (*tomlConfig, error) {
 	return config, err
 }
 
-// Returns an array of strings that contain the names
+// GetRegistries returns an array of strings that contain the names
 // of the registries as defined in the system-wide
 // registries file.  it returns an empty array if none are
 // defined
@@ -63,7 +63,7 @@ func GetRegistries() ([]string, error) {
 	return config.Registries.Search.Registries, nil
 }
 
-// Returns an array of strings that contain the names
+// GetInsecureRegistries returns an array of strings that contain the names
 // of the insecure registries as defined in the system-wide
 // registries file.  it returns an empty array if none are
 // defined


mtrmac:
:+1:

Thanks!

(Reordering the $PATH in `.travis.yml` might still be worthwhile, but this is sufficient.)

TomSweeneyRedHat:
Assuming happy tests, LGTM.

williammartin:
Yeh, I thought about whether to reorder the PATH and my gut feeling was that instead of accidentally hiding doubly installed things that could result in weird behaviour, we should seek to just install the correct version in future. These thoughts are based on entirely nothing substantive, just explaining what I was mulling around and could be nonsense.

Thanks!

mtrmac:
@runcom PTAL, this is blocking further development.

runcom:
LGTM

runcom:
needs a rebase

spikymonkey:
Wasn't sure if you wanted us to do the rebase or not, so we've gone ahead and done it anyway :) I think we were up to date with your master at the time of raising the PR, but I guess divergence is pretty routine whilst the approval goes through. Do you ever rebase PR'd branches along with the LGTM's, or always require the PR'ing team to do it if needed? Just so we know for next time. Could be worth capturing the process in a contributing.md so people can discover the processes?

mtrmac:
> …I guess divergence is pretty routine whilst the approval goes through.

It’s starting to become so now that there are more contributors (which is great!).

> Do you ever rebase PR'd branches along with the LGTM's, or always require the PR'ing team to do it if needed? Just so we know for next time.

Historically it’s been up to the submitter to rebase, and the reviewer would just click that green GitHub button. But you are very right that this is not the only possible way to do it, the reviewer could well rebase and merge manually — and now, with the higher frequency of merges and needed rebases, this does seem like a better workflow instead of one or more inherently racy roundtrips. @runcom WDYT?

> Could be worth capturing the process in a contributing.md so people can discover the processes?

Yes, also worth doing, after we know what the process is :)

mtrmac:
> But you are very right that this is not the only possible way to do it, the reviewer could well rebase and merge manually

A downside to this is that we wouldn’t have a successful Travis run for the rebased commit; the reviewer can of course run tests manually but they may not run exactly the same tests (e.g. `make` does not run `make test-skopeo`, and it is fairly easy to imagine two unrelated commits breaking the `gofmt -s -l` check when combined, for example.

This may be enough of a reason to require a fully rebased commit in GitHub, or to limit the rebases by a reviewer only to very clearly unrelated commits—or, perhaps, to move to a different CI bot system which does all of this automatically? atomic and OpenShift and and many other projects use something other than Travis, but I have no idea how it works or what it can do.

mtrmac:
Yeah, trying this with #337, rebasing+merging manually is error-prone, GitHub does not detect the PR as successfully merged. It’s not _impossible_ but a cleaner automation would be nice.

Issue #331, opened 2017-08-24
Don't expect image sources to convert images for us


owtaylor:
The point of the requestedManifestMIMETypes parameter to ImageReference.NewImageSource() is to allow the destination to affect the type of image requested from the source. It looks like the inspiration for that was that the atomic registry at one point didn't accept v2 manifests - #26 (references https://github.com/projectatomic/skopeo/pull/102#issuecomment-228747350)

However, that usage was effectively removed in #266, with with the idea that the if necessary, conversion happens inside containers/image when pushing to the registry.

The only reason to modify the list of mime types requested from the source, is if want the *source* to do the conversion rather than convert inside containers/image. But what happens inside the source is very unpredictable - it depends on the details of the source implementation, and those details may not be what we want. Examples:

 * If you pass only `Accept: application/vnd.oci.image.config.v1+json` to the docker registry, and what is stored is a schema2 manifest, it doesn't return the schema2 manifest unmodified, it converts it to schema1. And schema2 => schema1 => oci is going to be lower fidelity than schema2 => oci. (schema1 => oci is not even supported in containers/image at the moment though it wouldn't be hard to add as schema1 => schema2 => oci.)
 * If the source is the docker registry, and it has a signed schema2 manifest, then if you only pass `Accept: application/vnd.docker.distribution.manifest.v1+json`, the docker daemon happily drops the signatures, and the code in copy.Image() which is supposed to prevent conversions that drop signatures is defeated. 

Concretely, for docker/distribution, the set of conversions that happen currently are quite limited:
 * schema2 manifests are converted to schema1 manifests
 * manifest lists are converted to schema2 or vchema1 manifests by extracting the manifest for a defaul tt form.

I believe that these conversions will all happen properly with containers/image and the second is a third example of why it's better to do the conversion inside containers/image - the platform should be determined by the arch of the local system, not a default platform for the registry!

If the above argument makes, sense, I can do a PR for this - either removing requestedManifestMIMETypes parameter from ImageReference.NewImageSource() - which I guess is an API change. Or just making it do nothing locally in docker_image_src.go.

mtrmac:
I agree, removing the parameter from the API does seem like the right thing to do.

[...]

PR #338, opened 2017-08-31
Support pulling/pushing OCI manifests from a docker registry

owtaylor:
This pull request has a tiny one-line patch to add imgspecv1.MediaTypeImageManifest to the list of MIME types that a docker registry might support (as always, the exact list depends on the version of the registry we are talking to), and a bigger, though still trivial patch that prevents some breakage from the change - as discussed in #331.

(With the obvious compilation fix to skopeo, the skopeo tests all pass with these changes.)

The breakage that is avoided is that for copying from docker to oci, we don't want to pass only the imgspecv1.MediaTypeImageManifest in the Accept header - a) the particular docker registry might not in fact support imgspecv1.MediaTypeImageManifest b) and if it did, the particular manifest being requested might not be a imgspecv1.MediaTypeImageManifest, instead we want to pass in our normal list, get the manifest in it's original form and convert if necessary.

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -129,9 +129,7 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 		}
 	}()
 
-	destSupportedManifestMIMETypes := dest.SupportedManifestMIMETypes()
-
-	rawSource, err := srcRef.NewImageSource(options.SourceCtx, destSupportedManifestMIMETypes)
+	rawSource, err := srcRef.NewImageSource(options.SourceCtx)
 	if err != nil {
 		return errors.Wrapf(err, "Error initializing source %s", transports.ImageName(srcRef))
 	}
@@ -195,7 +193,7 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 
 	// We compute preferredManifestMIMEType only to show it in error messages.
 	// Without having to add this context in an error message, we would be happy enough to know only that no conversion is needed.
-	preferredManifestMIMEType, otherManifestMIMETypeCandidates, err := determineManifestConversion(&manifestUpdates, src, destSupportedManifestMIMETypes, canModifyManifest)
+	preferredManifestMIMEType, otherManifestMIMETypeCandidates, err := determineManifestConversion(&manifestUpdates, src, dest.SupportedManifestMIMETypes(), canModifyManifest)
 	if err != nil {
 		return err
 	}
--- a/docker/docker_image_src.go
+++ b/docker/docker_image_src.go
@@ -21,41 +21,23 @@ import (
 )
 
 type dockerImageSource struct {
-	ref                        dockerReference
-	requestedManifestMIMETypes []string
-	c                          *dockerClient
+	ref dockerReference
+	c   *dockerClient
 	// State
 	cachedManifest         []byte // nil if not loaded yet
 	cachedManifestMIMEType string // Only valid if cachedManifest != nil
 }
 
-// newImageSource creates a new ImageSource for the specified image reference,
-// asking the backend to use a manifest from requestedManifestMIMETypes if possible.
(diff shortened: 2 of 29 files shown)


copy/copy.go
```diff
@@ -195,6 +193,7 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 
 	// We compute preferredManifestMIMEType only to show it in error messages.
 	// Without having to add this context in an error message, we would be happy enough to know only that no conversion is needed.
+	destSupportedManifestMIMETypes := dest.SupportedManifestMIMETypes()
 	preferredManifestMIMEType, otherManifestMIMETypeCandidates, err := determineManifestConversion(&manifestUpdates, src, destSupportedManifestMIMETypes, canModifyManifest)
```
mtrmac: Nit: inlining the call directly and eliminating the variable would be shorter.

docker/docker_image_src.go
```diff
@@ -22,39 +22,21 @@ import (
...
-func newImageSource(ctx *types.SystemContext, ref dockerReference, requestedManifestMIMETypes []string) (*dockerImageSource, error) {
+func newImageSource(ctx *types.SystemContext, ref dockerReference) (*dockerImageSource, error) {
 	c, err := newDockerClient(ctx, ref, false, "pull")
 	if err != nil {
 		return nil, err
 	}
-	if requestedManifestMIMETypes == nil {
-		requestedManifestMIMETypes = manifest.DefaultRequestedManifestMIMETypes
-	}
-	supportedMIMEs := supportedManifestMIMETypesMap()
```
mtrmac: This removes the only user of `supportedManifestMIMETypesMap`; can you remove it, please, and perhaps inline the global `manifestMIMETypes` variable into `dockerImageDestination.SupportedManifestMIMETypes`?

owtaylor:
The latest set of pushes makes both code improvements you suggested, and the skopeo update is in https://github.com/projectatomic/skopeo/pull/403.

oci/layout/oci_transport.go.rej
```diff
@@ -0,0 +1,15 @@
+diff a/oci/layout/oci_transport.go b/oci/layout/oci_transport.go	(rejected hunks)
```
mtrmac: Please drop this file.

signature/#policy_reference_match_test.go#
```diff
@@ -0,0 +1,366 @@
+package signature
```
mtrmac: Please drop this file.

owtaylor:
Oy, sorry about the stray files! Should not have deviated from my normal git habits. Fixed now.

mtrmac:
```
 52ececd - FAIL - does not have a valid DCO
 be736b4 - FAIL - does not have a valid DCO
```
(`git commit --amend -s`)

owtaylor:
Added the sign-offs

owtaylor:
> 👍 pending tests

What tests are you waiting for? projectatomic/skopeo#403 seems to pass (can repush with a sign-off now, or just do that when rebasing when this lands), tests here aren't going to pass.

mtrmac:
> What tests are you waiting for?

You’re right, some failures are expected; just to make sure there are no _unexpected_ failures.  E.g. right now, please run
```sh
gofmt -s -w docker/docker_image_src.go openshift/openshift.go
```

owtaylor:
> You’re right, some failures are expected; just to make sure there are no unexpected failures. E.g. right now, please run
> 
> gofmt -s -w docker/docker_image_src.go openshift/openshift.go

Good point. Repushed with the go fmt changes.

mtrmac:
One more test failure, due to #314 landing since the original PR was prepared.

owtaylor:
This and https://github.com/projectatomic/skopeo/pull/403 are now rebased on top of the latest code and (I think) passing all the tests they are supposed to pass.

mtrmac:
Thanks!  Please re-vendor https://github.com/projectatomic/skopeo/pull/403 back to the `master` branch.

PR #340, opened 2017-09-06
pkg/sysregistries/system_registries.go: Override system registries path

baude:
pkg/sysregistries/system_registries.go: Override system registries path

For the purposes of testing, it is handy to override the default configuration
file location of the system registries configuration file.  This is particularilyi
true in the case of crio-o which consumes this library.

Signed-off-by: baude <bbaude@redhat.com>

### DIFF
--- a/pkg/sysregistries/system_registries.go
+++ b/pkg/sysregistries/system_registries.go
@@ -2,7 +2,9 @@ package sysregistries
 
 import (
 	"github.com/BurntSushi/toml"
+	"github.com/containers/image/types"
 	"io/ioutil"
+	"path/filepath"
 )
 
 // systemRegistriesConfPath is the path to the system-wide registry configuration file
@@ -29,8 +31,16 @@ type tomlConfig struct {
 
 // Reads the global registry file from the filesystem. Returns
 // a byte array
-func readRegistryConf() ([]byte, error) {
-	configBytes, err := ioutil.ReadFile(systemRegistriesConfPath)
+func readRegistryConf(ctx *types.SystemContext) ([]byte, error) {
+	dirPath := systemRegistriesConfPath
+	if ctx != nil {
+		if ctx.SystemRegistriesConfPath != "" {
+			dirPath = ctx.SystemRegistriesConfPath
+		} else if ctx.RootForImplicitAbsolutePaths != "" {
+			dirPath = filepath.Join(ctx.RootForImplicitAbsolutePaths, systemRegistriesConfPath)
+		}
+	}
+	configBytes, err := ioutil.ReadFile(dirPath)
 	return configBytes, err
 }
 
@@ -39,10 +49,10 @@ var readConf = readRegistryConf
 
 // Loads the registry configuration file from the filesystem and
 // then unmarshals it.  Returns the unmarshalled object.
-func loadRegistryConf() (*tomlConfig, error) {
+func loadRegistryConf(ctx *types.SystemContext) (*tomlConfig, error) {
 	config := &tomlConfig{}
 
-	configBytes, err := readConf()
+	configBytes, err := readConf(ctx)
 	if err != nil {
 		return nil, err
 	}
@@ -55,8 +65,8 @@ func loadRegistryConf() (*tomlConfig, error) {
 // of the registries as defined in the system-wide
 // registries file.  it returns an empty array if none are
 // defined
-func GetRegistries() ([]string, error) {
-	config, err := loadRegistryConf()
+func GetRegistries(ctx *types.SystemContext) ([]string, error) {
+	config, err := loadRegistryConf(ctx)
 	if err != nil {
 		return nil, err
 	}
@@ -67,8 +77,8 @@ func GetRegistries() ([]string, error) {
 // of the insecure registries as defined in the system-wide
 // registries file.  it returns an empty array if none are
 // defined
-func GetInsecureRegistries() ([]string, error) {
-	config, err := loadRegistryConf()
+func GetInsecureRegistries(ctx *types.SystemContext) ([]string, error) {
+	config, err := loadRegistryConf(ctx)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


pkg/sysregistries/system_registries.go
```diff
@@ -30,7 +31,17 @@ type tomlConfig struct {
 // Reads the global registry file from the filesystem. Returns
 // a byte array
 func readRegistryConf() ([]byte, error) {
-	configBytes, err := ioutil.ReadFile(systemRegistriesConfPath)
+	var configFilePath string
+	// If an environment variable for the configuration file
+	// exists, use it over the the default location.  This is
+	// useful for testing purposes in cri-o.
+	altConfPath := os.Getenv("REGISTRIES_CONF_PATH")
```
mtrmac: If this is _only_ for testing, it seems overkill to have it as an user-accessible environment variable in all installations.

Would it be possible to instead use `types.SystemContext`, e.g. the way `docker/lookaside.go:registriesDirPath` works, and for the tests to use an appropriately initialized `types.SystemContext`?

baude: I'll look into it ...

types/types.go
```diff
@@ -324,6 +324,8 @@ type SystemContext struct {
 	DockerDisableV1Ping bool
 	// Directory to use for OSTree temporary files
 	OSTreeTmpDirPath string
+	// Path to the system-wide registries configuration file
+	SystemRegistriesConfPath string
```
mtrmac: Please move the field above the os-tree related one (which should have been in its own `=== ostree.Transport overrides ==` section).

(Assuming that `sysregistries` is Docker-specific; if not, to the general section.)

pkg/sysregistries/system_registries.go
```diff
@@ -29,8 +30,12 @@ type tomlConfig struct {
 
 // Reads the global registry file from the filesystem. Returns
 // a byte array
-func readRegistryConf() ([]byte, error) {
-	configBytes, err := ioutil.ReadFile(systemRegistriesConfPath)
+func readRegistryConf(ctx *types.SystemContext) ([]byte, error) {
+	dirPath := systemRegistriesConfPath
+	if ctx != nil {
+		dirPath = ctx.SystemRegistriesConfPath
+	}
```
mtrmac: This must also check for `ctx.SystemRegistriesConfPath != ""`; a `SystemContext` may be provided by the caller to override something else in a deep call stack.

Please include support for `RootForImplicitAbsolutePaths` as well.

pkg/sysregistries/system_registries_test.go
```diff
@@ -1,21 +1,22 @@
...
+	"github.com/containers/image/types"
 	"github.com/stretchr/testify/assert"
 	"testing"
 )
 
 var testConfig = []byte("")
 
 func init() {
-	readConf = func() ([]byte, error) {
+	readConf = func(_ *types.SystemContext) ([]byte, error) {
```
mtrmac: BTW the tests could now use fixture files and use `SystemContext` to point to them, instead of having the `readConf` variable and overriding the implementation.

Not at all a blocker for this PR, only a newly available option.

pkg/sysregistries/system_registries.go
runcom: I think if caller sets to "" it's his fault, maybe this should be a `*string` instead so we can check if caller sets it?

pkg/sysregistries/system_registries_test.go
runcom: :+1: to having this indeed

types/types.go
baude: Moved to general ... it is not docker specific.

pkg/sysregistries/system_registries.go
mtrmac: Golang does not have optionals, a `string` is `""` by default, it’s not something the caller does intentionally.

A `*string` would work as well here, but `""` is a clearly invalid value for a path, and all the other users treat `""` as unspecified as well.

(The pedantic me would _love_ to have true optionals in the language, and to treat `""` as any other string value WRT code paths. But, we have Golang to work with, and all of the other users doing it strictly-speaking-wrong, so let’s at least be consistently wrong/ugly :-| )

runcom: > Golang does not have optionals, a string is "" by default, it’s not something the caller does intentionally.

I know, that's why (following opencontainers specs) I suggested `*string`. If it's `nil`,caller didn't set it, otherwise, it's her fault if she passed an empty string.

baude: Added

pkg/sysregistries/system_registries_test.go
baude: Indeed, I thought about that too, but will leave alone for now.  If I change the tests in the future, I'll likely move to that approach.

TomSweeneyRedHat:
Code LGTM, but it looks like you need a remerge and Travis isn't happy.

pkg/sysregistries/system_registries.go
```diff
@@ -29,8 +31,17 @@ type tomlConfig struct {
...
 // a byte array
-func readRegistryConf() ([]byte, error) {
-	configBytes, err := ioutil.ReadFile(systemRegistriesConfPath)
+func readRegistryConf(ctx *types.SystemContext) ([]byte, error) {
+	dirPath := systemRegistriesConfPath
+	if ctx != nil {
+		if ctx.RegistriesDirPath != "" {
+			dirPath = ctx.SystemRegistriesConfPath
+		}
+		if ctx.RootForImplicitAbsolutePaths != "" {
```
mtrmac: `else if` here (the template, `registriesDirPath`, has a `return` for each of the cases); `RegistriesDirPath` has the highest priority.

pkg/sysregistries/system_registries.go
mtrmac: `*string` would be fine with me, sure. It _is_ strictly speaking better, and we might just as well start somewhere.

(I can live with `""` as well.)

mtrmac:
:+1: if tests pass.

runcom:
LGTM

PR #342, opened 2017-09-08
ostree: fix a couple of issues

giuseppe:
a couple of issues I've observed while using `atomic pull` (which now uses Skopeo copy for ostree as well)

### DIFF
--- a/ostree/ostree_transport.go
+++ b/ostree/ostree_transport.go
@@ -84,24 +84,15 @@ func NewReference(image string, repo string) (types.ImageReference, error) {
 	// image is not _really_ in a containers/image/docker/reference format;
 	// as far as the libOSTree ociimage/* namespace is concerned, it is more or
 	// less an arbitrary string with an implied tag.
-	// We use the reference.* parsers basically for the default tag name in
-	// reference.TagNameOnly, and incidentally for some character set and length
-	// restrictions.
-	var ostreeImage reference.Named
-	s := strings.SplitN(image, ":", 2)
-
-	named, err := reference.WithName(s[0])
+	// Parse the image using reference.ParseNormalizedNamed so that we can
+	// check whether the images has a tag specified and we can add ":latest" if needed
+	ostreeImage, err := reference.ParseNormalizedNamed(image)
 	if err != nil {
 		return nil, err
 	}
 
-	if len(s) == 1 {
-		ostreeImage = reference.TagNameOnly(named)
-	} else {
-		ostreeImage, err = reference.WithTag(named, s[1])
-		if err != nil {
-			return nil, err
-		}
+	if reference.IsNameOnly(ostreeImage) {
+		image = image + ":latest"
 	}
 
 	resolved, err := explicitfilepath.ResolvePathToFullyExplicit(repo)
@@ -123,8 +114,8 @@ func NewReference(image string, repo string) (types.ImageReference, error) {
 	}
 
 	return ostreeReference{
-		image:      ostreeImage.String(),
-		branchName: encodeOStreeRef(ostreeImage.String()),
+		image:      image,
+		branchName: encodeOStreeRef(image),
 		repo:       resolved,
 	}, nil
 }
--- a/ostree/ostree_transport_test.go
+++ b/ostree/ostree_transport_test.go
@@ -44,6 +44,7 @@ var imageNameTestcases = []struct{ input, normalized, branchName string }{
 	{"busybox:notlatest", "busybox:notlatest", "busybox_3Anotlatest"},                                                  // Explicit tag
 	{"busybox", "busybox:latest", "busybox_3Alatest"},                                                                  // Default tag
 	{"docker.io/library/busybox:latest", "docker.io/library/busybox:latest", "docker.io_2Flibrary_2Fbusybox_3Alatest"}, // A hierarchical name
+	{"127.0.0.1:5000/busybox:latest", "127.0.0.1:5000/busybox:latest", "127.0.0.1_3A5000_2Fbusybox_3Alatest"},          // Port usage
 	{"UPPERCASEISINVALID", "", ""},                                                                                     // Invalid input
 	{"busybox" + sha256digest, "", ""},                                                                                 // Digested references are not supported (parsed as invalid repository name)
 	{"busybox:invalid+tag", "", ""},                                                                                    // Invalid tag value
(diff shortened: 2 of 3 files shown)


ostree/ostree_transport_test.go
```diff
@@ -46,7 +46,7 @@ var imageNameTestcases = []struct{ input, normalized, branchName string }{
 	{"docker.io/library/busybox:latest", "docker.io/library/busybox:latest", "docker.io_2Flibrary_2Fbusybox_3Alatest"}, // A hierarchical name
 	{"127.0.0.1:5000/busybox:latest", "127.0.0.1:5000/busybox:latest", "127.0.0.1_3A5000_2Fbusybox_3Alatest"},          // Port usage
 	{"UPPERCASEISINVALID", "", ""},                                                                                     // Invalid input
-	{"busybox" + sha256digest, "", ""},                                                                                 // Digested references are not supported (parsed as invalid repository name)
+	{"busybox" + sha256digest, "busybox@sha256:0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef", "busybox_40sha256_3A0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef"},
```
ashcrow: I recommend placing this above the previous invalid input to keep the valid and invalids in their own blocks.

giuseppe: amended and pushed

ashcrow:
To pass validate:
```patch
--- a/ostree/ostree_transport_test.go
+++ b/ostree/ostree_transport_test.go
@@ -47,9 +47,9 @@ var imageNameTestcases = []struct{ input, normalized, branchName string }{
        {"127.0.0.1:5000/busybox:latest", "127.0.0.1:5000/busybox:latest", "127.0.0.1_3A5000_2Fbusybox_3Alatest"},          // Port usage
        {"UPPERCASEISINVALID", "", ""},                                                                                     // Invalid input
        {"busybox" + sha256digest, "busybox@sha256:0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef", "busybox_40sha256_3A0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef"},
-       {"busybox:invalid+tag", "", ""},                                                                                    // Invalid tag value
-       {"busybox:tag:with:colons", "", ""},                                                                                // Multiple colons - treated as a tag which contains a colon, which is invalid
-       {"", "", ""},                                                                                                       // Empty input is rejected (invalid repository.Named)
+       {"busybox:invalid+tag", "", ""},     // Invalid tag value
+       {"busybox:tag:with:colons", "", ""}, // Multiple colons - treated as a tag which contains a colon, which is invalid
+       {"", "", ""},                        // Empty input is rejected (invalid repository.Named)
 }
```

ashcrow:
```diff
--- a/ostree/ostree_transport_test.go
+++ b/ostree/ostree_transport_test.go
@@ -49,7 +49,7 @@ var imageNameTestcases = []struct{ input, normalized, branchName string }{
        {"UPPERCASEISINVALID", "", ""},      // Invalid input
        {"busybox:invalid+tag", "", ""},     // Invalid tag value
        {"busybox:tag:with:colons", "", ""}, // Multiple colons - treated as a tag which contains a colon, which is invalid
-       {"", "", ""},                                                                                                       // Empty input is rejected (invalid repository.Named)
+       {"", "", ""},                        // Empty input is rejected (invalid repository.Named)
 }

```

giuseppe:
The @sha256 support might still need some changes, I have to check how it is done in the current version of atomic. I will add a note once I verify it

ostree/ostree_transport.go
```diff
@@ -87,21 +87,13 @@ func NewReference(image string, repo string) (types.ImageReference, error) {
 	// We use the reference.* parsers basically for the default tag name in
 	// reference.TagNameOnly, and incidentally for some character set and length
 	// restrictions.
-	var ostreeImage reference.Named
-	s := strings.SplitN(image, ":", 2)
-
-	named, err := reference.WithName(s[0])
+	ostreeImage, err := reference.ParseNormalizedNamed(image)
```
mtrmac: Fair enough; please update the comment to say ”to detect whether the reference has a tag” (the code below hard-codes the default value now).

Non-blocking: Consider also emphasizing that the result of the parsing is ignored (in the comment above or below when creating the `ostreeRefrence`), so the normalization done does not matter, and naming the variable something which sounds less relevant (`ostreeImage` seems more pertinent than the input `image`, but the code is intentionally throwing it away), maybe `parsedRef` or something like that.

ostree/ostree_transport_test.go
```diff
@@ -44,11 +44,12 @@ var imageNameTestcases = []struct{ input, normalized, branchName string }{
...
-	{"UPPERCASEISINVALID", "", ""},                                                                                     // Invalid input
-	{"busybox" + sha256digest, "", ""},                                                                                 // Digested references are not supported (parsed as invalid repository name)
-	{"busybox:invalid+tag", "", ""},                                                                                    // Invalid tag value
-	{"busybox:tag:with:colons", "", ""},                                                                                // Multiple colons - treated as a tag which contains a colon, which is invalid
-	{"", "", ""},                                                                                                       // Empty input is rejected (invalid repository.Named)
+	{"127.0.0.1:5000/busybox:latest", "127.0.0.1:5000/busybox:latest", "127.0.0.1_3A5000_2Fbusybox_3Alatest"},          // Port usage
+	{"busybox" + sha256digest, "busybox@sha256:0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef", "busybox_40sha256_3A0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef"},
```
mtrmac: This would probably be a bit more readable if the other two strings used `sha256DigestHex`.

giuseppe: good idea!  Going to fix that

giuseppe:
I've pushed a new version with the requested changes.

I've verified that upstream atomic (which is uses "skopeo copy" when possible) handles correctly '@sha256:' images with a Skopeo including the changes in this PR:

```
# atomic install --system busybox@sha256:b82b5740006c1ab823596d2c07f081084ecdb32fd258072707b99f52a3cb8692
Extracting to /var/lib/containers/atomic/busybox.0
Warning: /exports directory not found.  Default config files will be generated.
Warning: You may want to modify `/var/lib/containers/atomic/busybox.0/config.json` before starting the service
systemctl daemon-reload
systemctl enable busybox
# ostree refs
ociimage/add3ddb21edebb102b552fc129273216bf6312f5f1519d7c1401864a2810738b
ociimage/busybox_40sha256_3Ab82b5740006c1ab823596d2c07f081084ecdb32fd258072707b99f52a3cb8692
```

ostree/ostree_transport.go
```diff
@@ -84,24 +84,16 @@ func NewReference(image string, repo string) (types.ImageReference, error) {
...
-	// reference.TagNameOnly, and incidentally for some character set and length
-	// restrictions.
-	var ostreeImage reference.Named
-	s := strings.SplitN(image, ":", 2)
-
-	named, err := reference.WithName(s[0])
+	// Parse the image using rereference.ParseNormalizedNamed so that we can
+	// check whether the images has a tag specified and we can add ":latest" if needed
+	// A tag must be always specified in the OSTree branch name.
+	ostreeImage, err := reference.ParseNormalizedNamed(image)
```
mtrmac: “A tag must be always specified” seems incorrect, with `"busybox@sha256:…"` now supported.

`s/rereference/reference/`.

giuseppe: pushed a new version with this fixed

mtrmac:
LGTM pending tests

runcom:
LGTM

PR #115, opened 2016-10-01
manifest: enable DockerV2List

runcom:
Related to https://github.com/containers/image/issues/114

Needed for projectatomic/docker#200 - w/o this pulling an image with a manifest list is broken into docker.
Rigth now, by not supporting the manifest list media type we always got the _single_ manifest for the amd64 arch instead of the whole manifest list and this broke docker because computing the digest for that single manifest isn't the same as the digest from the manifest list.

Also note that if you pull-by-digest a manifest list in docker you get back exactly the same manifest list digest (and not the one from the target manifest) which is confusing to me but still, it's how docker works.

@mtrmac PTAL - this is blocking

Signed-off-by: Antonio Murdaca <runcom@redhat.com>

PR #344, opened 2017-09-13
Hotfix: Do not fetch manifest lists from Docker registries

mtrmac:
This allows `copy.Image` to work with Docker Hub registries, which now use manifest lists for many images; the downside is breaking non-`copy` uses for any non-default architecture.

### DIFF
--- a/manifest/manifest.go
+++ b/manifest/manifest.go
@@ -35,7 +35,7 @@ var DefaultRequestedManifestMIMETypes = []string{
 	DockerV2Schema2MediaType,
 	DockerV2Schema1SignedMediaType,
 	DockerV2Schema1MediaType,
-	DockerV2ListMediaType,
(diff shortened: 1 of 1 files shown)


mtrmac:
`test-skopeo`’s `CopySuite.TestCopyFailsWithManifestList` is expected to fail with this.

runcom:
> test-skopeo’s CopySuite.TestCopyFailsWithManifestList is expected to fail with this.

will comment it out for now

runcom:
LGTM

mtrmac:
:+1:

I have marked the test as `ExpectFailure` in https://github.com/projectatomic/skopeo/pull/417 , and the Linux build has succeeded there.

NOTE: This _must not_ be vendored into projectatomic/docker, see https://github.com/containers/image/pull/115 (pulling a manifest list by digest would give us the non-manifest-list manifest, which does not match the digest, and fails).

PR #346, opened 2017-09-16
Support copying a single image from manifest lists

mtrmac:
This teaches `copy.Image` to handle manifest lists, instead of failing, by choosing an image appropriate for the current system, and copying only that one.

Ultimately we want to copy the whole manifest list; this does not yet do that.

**NOTE:** This is an alternative to #344, and should have a similar effect for Linux users. macOS users trying to access Linux images will instead start seeing errors like
```
Error choosing an image from manifest list docker://busybox:latest: no image found in manifest list for architecture amd64, OS darwin 
```
unless they explicitly make a `types.SystemContext{OSChoice:"linux"}` override. This may be fine for some uses, but the advertised use of `skopeo copy` for copying between docker/distribution registries suffers (until we implement a copy of the whole manifest list).

`types.ImageSource` is modified to understand a set of images distinguished by an optional `instanceDigest`; (as opposed to creating new `types.ImageSource` instances for each image instance, this allows sharing state, e.g. `docker load`, or an unpacked archive, for all of the instances). Then `copy.Image` is split into a per-manifest-list initialization, and a per-image-instance copy operation, so the per-manifest-list part can choose a single instance if necessary.

**NOTE:** One of the implications of this is that signature verification now happens _only_ for the individual instances; manifest list signatures are not required and not processed at all.

The overall diff is probably difficult to review, please see the individual commits, and read the individual commit messages messages for more detailed discussion:

 - Fix an incorrect variable reference in storageImageSource.GetSignatures
 - Move the core logic of types.Image.IsMultiImage to c/i/manifest
 - Remove types.Image.IsMultiImage
 - Do the isMultiImage decision already on an UnparsedImage, not an Image
 - Rename copy.imageCopier to copier
 - Split copy.imageCopier from copy.copier
 - Use copier and imageCopier instead of parameters and variables
 - Add copier.Printf
 - FIXME Replace GetTargetManifest by GetManifest(instanceDigest)
 - FIXME: Add a instanceDigest parameter to GetSignatures
 - Support instanceDigest in types.UnparsedImage and types.Image
 - Do not Close the ImageSource in UnparsedImage/Image
 - Split copier.copyOneImage from copy.Image
 - Finally, if the copy.Image source is a manifest list, copy a single instance
 - Add architecture/OS overrides to types.SystemContext
 - Revert "Hotfix: Do not fetch manifest lists from Docker registries"

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -142,8 +142,17 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 			}
 		}
 	}()
+	multiImage, err := isMultiImage(unparsedImage)
+	if err != nil {
+		return errors.Wrapf(err, "Error determining manifest MIME type for %s", transports.ImageName(srcRef))
+	}
+	if multiImage {
+		return errors.Errorf("can not copy %s: manifest contains multiple images", transports.ImageName(srcRef))
+	}
 
 	// Please keep this policy check BEFORE reading any other information about the image.
+	// (the multiImage check above only matches the MIME type, which we have received anyway.
+	// Actual parsing of anything should be deferred.)
 	if allowed, err := policyContext.IsRunningImageAllowed(unparsedImage); !allowed || err != nil { // Be paranoid and fail if either return value indicates so.
 		return errors.Wrap(err, "Source image rejected")
 	}
@@ -160,14 +169,6 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 
 	if err := checkImageDestinationForCurrentRuntimeOS(src, dest); err != nil {
 		return err
-	}
-
-	multiImage, err := isMultiImage(src)
-	if err != nil {
-		return errors.Wrapf(err, "Error determining manifest MIME type for %s", transports.ImageName(srcRef))
-	}
-	if multiImage {
-		return errors.Errorf("can not copy %s: manifest contains multiple images", transports.ImageName(srcRef))
 	}
 
 	var sigs [][]byte
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -159,6 +159,20 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 		return errors.Errorf("can not copy %s: manifest contains multiple images", transports.ImageName(srcRef))
 	}
 
+	if err := c.copyOneImage(policyContext, options, unparsedImage); err != nil {
+		return err
+	}
+
+	if err := c.dest.Commit(); err != nil {
+		return errors.Wrap(err, "Error committing the finished image")
+	}
+
+	return nil
+}
+
+// Image copies a single (on-manifest-list) image unparsedImage, using policyContext to validate
+// source image admissibility.
+func (c *copier) copyOneImage(policyContext *signature.PolicyContext, options *Options, unparsedImage *image.UnparsedImage) (retErr error) {
 	// Please keep this policy check BEFORE reading any other information about the image.
 	// (the multiImage check above only matches the MIME type, which we have received anyway.
 	// Actual parsing of anything should be deferred.)
@@ -167,10 +181,10 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 	}
 	src, err := image.FromUnparsedImage(unparsedImage)
 	if err != nil {
-		return errors.Wrapf(err, "Error initializing image from source %s", transports.ImageName(srcRef))
+		return errors.Wrapf(err, "Error initializing image from source %s", transports.ImageName(c.rawSource.Reference()))
 	}
 
-	if err := checkImageDestinationForCurrentRuntimeOS(src, dest); err != nil {
+	if err := checkImageDestinationForCurrentRuntimeOS(src, c.dest); err != nil {
 		return err
 	}
 
@@ -187,14 +201,14 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 	}
 	if len(sigs) != 0 {
 		c.Printf("Checking if image destination supports signatures\n")
-		if err := dest.SupportsSignatures(); err != nil {
+		if err := c.dest.SupportsSignatures(); err != nil {
 			return errors.Wrap(err, "Can not copy signatures")
 		}
 	}
 
 	ic := imageCopier{
 		c:               c,
-		manifestUpdates: &types.ManifestUpdateOptions{InformationOnly: types.ManifestUpdateInformation{Destination: dest}},
+		manifestUpdates: &types.ManifestUpdateOptions{InformationOnly: types.ManifestUpdateInformation{Destination: c.dest}},
 		src:             src,
 		// diffIDsAreNeeded is computed later
 		canModifyManifest: len(sigs) == 0,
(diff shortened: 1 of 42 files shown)


mtrmac:
@runcom RFC. I still need to update/add a few tests.

mtrmac:
BTW as much as `types.ImageReference.NewImage` is convenient, what it does for manifest lists (`.Manifest()`, and the manifest digest, are the manifest list, but everything else is obtained by parsing a selected manifest instance, and notably `.UpdatedImage()` creates a new manifest from the instance, not for the list) is pretty likely to be incorrect or at least very surprising for most callers.

I’m very tempted to break the API, even if just to rename it to `NewImageForDefaultInstance` or `NewImageWithWeirdManifestListHandling`, just to force callers to think this through.

But I think even better would be eliminating the `.NewImage()` method entirely — it originally exists for the `docker.Image` and `storage.storageImage` overrides, which can be done better in other ways — and replacing all of those mostly copy&pasted implementations with a single helper, perhaps `image.FromReference` or something like that, and then we can think about handling manifests lists in there (maybe just by failing when the image is a manifest list and the caller did not expect that).

(This is all obviously for a separate PR, just noting that this PR highlights how `NewImage` is a bad fit.)

mtrmac:
See https://github.com/projectatomic/skopeo/pull/420 for the corresponding `skopeo` update.

runcom:
@cyphar @hallyn PTAL

runcom:
> NOTE: This is an alternative to #344, and should have a similar effect for Linux users. macOS users trying to access Linux images will instead start seeing errors like

I'm fine with this, macOS users now have to choose arch/os and that's fine because skopeo allows for this kind of fine-grained operations. So +1 on this and the flags you added in skopeo PR.

> Ultimately we want to copy the whole manifest list; this does not yet do that.

assuming you're talking about destinations like `oci` which supports manifest index/list, I'm fine with this either, let's tackle that problem later (unless some contributor want to take it, hint hint) (for the record CRI-O isn't affected by this.)

> NOTE: One of the implications of this is that signature verification now happens only for the individual instances; manifest list signatures are not required and not processed at all.

Good. That seems ok as a start.

@mtrmac overall the approach looks good to me. I'll make a detailed code review once fully ready

mtrmac:
> > NOTE: This is an alternative to #344, and should have a similar effect for Linux users. macOS users trying to access Linux images will instead start seeing errors like
>
> I'm fine with this, macOS users now have to choose arch/os and that's fine because skopeo allows for this kind of fine-grained operations.

Well, we used to recommend (`skopeo copy docker://… docker://…`) and it would mostly Just Work; now users would be forced to add `--override-os`, and after copying manifest lists is implemented, they may want to remove `--override-os` again.

OTOH quite a few of the Hub images now have 7 or so variants, and it’s pretty likely that most users don’t actually want to always copy of all of them. So forcing users to make an explicit choice might not be so bad.

mtrmac:
> @mtrmac overall the approach looks good to me. I'll make a detailed code review once fully ready

Done (added a few noddy tests to `dir:` and `containers-storage`, and a test for `chooseDigestFromManifestList`).

mtrmac:
@runcom PTAL

cyphar:
I will take a look over this when I get a chance (this week hopefully).

mtrmac:
@runcom PTAL.  Tests pass in https://github.com/projectatomic/skopeo/pull/420 (at least on Linux right now)

runcom:
> This may be fine for some uses, but the advertised use of skopeo copy for copying between docker/distribution registries suffers (until we implement a copy of the whole manifest list).

I'm actually afraid of this one, right now official images are now manifest lists, what would happen if I copy the redis image which is a list to another registry with this patch? (AFK right now)

mtrmac:
> > This may be fine for some uses, but the advertised use of skopeo copy for copying between docker/distribution registries suffers (until we implement a copy of the whole manifest list).
>
> I'm actually afraid of this one, right now official images are now manifest lists, what would happen if I copy the redis image which is a list to another registry with this patch? (AFK right now) 

It will select a variant for your current runtime environment (or the one you have chosen via `SystemContext`) and copy only that one image, not the whole manifest list.

For most people who only care about x86_64 this seems to be fine, perhaps even better than copying all of the architectures. macOS users wanting to copy Linux images will be forced to add an extra CLI option.

copy/copy.go
```diff
@@ -162,7 +162,11 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 		return err
 	}
 
-	if src.IsMultiImage() {
+	multiImage, err := isMultiImage(src)
+	if err != nil {
+		return errors.Wrapf(err, "Error determining manifest MIME type for %s", transports.ImageName(srcRef))
```
rhatdan: "error determining ..."

copy/copy.go
```diff
@@ -142,8 +142,17 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 			}
 		}
 	}()
+	multiImage, err := isMultiImage(unparsedImage)
+	if err != nil {
+		return errors.Wrapf(err, "Error determining manifest MIME type for %s", transports.ImageName(srcRef))
```
rhatdan: Lower case Error.

copy/copy.go
```diff
@@ -69,20 +69,24 @@ func (d *digestingReader) Read(p []byte) (int, error) {
...
+	dest             types.ImageDestination
+	rawSource        types.ImageSource
+	reportWriter     io.Writer
+	progressInterval time.Duration
+	progress         chan types.ProgressProperties
+}
+
+// imageCopier tracks state specific to a single image (possibly an item of a manifest list)
+type imageCopier struct {
+	c                 *copier
```
rhatdan: Should the object be a copier rather then a "c"

image/fixtures/schema2list.json
```diff
@@ -0,0 +1,82 @@
...
+         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
+         "size": 528,
+         "digest": "sha256:e5aa1b0a24620228b75382997a0977f609b3ca3a95533dafdef84c74cc8df642",
+         "platform": {
+            "architecture": "s390x",
+            "os": "linux"
+         }
+      }
+   ]
+}
```
rhatdan: newline

rhatdan:
Looks like throuhout the code the error messages switch from lower case to upper case and back.  The standard for the tools that use this library is lower case for errors. 
Probably should go through all error returns and lowercase them, to make them consistent.
In a different PR.

mtrmac:
> Looks like throuhout the code the error messages switch from lower case to upper case and back. The standard for the tools that use this library is lower case for errors.

The library has been fairly consistently using upper case; historically primarily because parts of `skopeo` have had minimal error handling like `if err != nil {return err}` ultimately resulting in `logrus.Fatalf(the_error_from_containers_image)`.

copy/copy.go
mtrmac: It is used all over the place, and consistent with `func (c *copier)` receivers of the per-manifest-list code.

image/fixtures/schema2list.json
mtrmac: This is an actual `busybox` manifest coming from Docker Hub; it is most representative unmodified.

runcom:
yeah I guess it's fine since it's also docker push behavior to copy just one image for now. We'll revisit this later.

runcom:
LGTM

mtrmac:
👍

PR #349, opened 2017-09-21
ostree: store the image digest in the OSTree metadata

giuseppe:


giuseppe:
The related change in atomic to use this feature: https://github.com/projectatomic/atomic/pull/1103

mtrmac:
:+1:

Hum, `_get_image_id_from_manifest` seems a bit suspicious (is it clear that it will never mix manifest digests and config digests in the same namespace?), but just recording the data here seems harmless, however it is consumed.

runcom:
LGTM

PR #348, opened 2017-09-21
Support URLs on oci layers

goonzoid:
According to the image spec [1], descriptors should be able to contain
URLs. Support them on the returned BlobInfo for layers.

[1] https://github.com/opencontainers/image-spec/blob/master/descriptor.md

Signed-off-by: Danail Branekov <danail.branekov@sap.com>

PR #147, opened 2016-11-02
[WIP] *: support layers federation

runcom:
@mtrmac PTAL

More information available at https://github.com/docker/docker/pull/27961

Signed-off-by: Antonio Murdaca <runcom@redhat.com>

PR #351, opened 2017-09-28
Support to download remote OCI layers (with TLS)

danail-branekov:
This is a follow up of #348, which added the `URL` property to the BlobInfo descriptor.

This will allow `GetBlob` to fetch external blobs when the "urls" property is present. The download is performed by an HTTP client that has its `Transport` configured depending on the newly introduced `SytemContext.OCICertPath` and `SystemContext.OCIInsecureSkipTLSVerify` properties. 

`BlobInfo.URLs` is an array and therefore the implementation would iterate over the URLs until HTTP OK response is returned.

When the `SytemContext.OCICertPath` is specified the implementation would look for CA, private key and public key certificates the same way the Docker client does. As a matter of fact they both use one and the same certificate handling implementation that is extracted in the newly introduced `tlsclientconfig` package.

(this is part of a [CloudFoundry/Grootfs](https://github.com/cloudfoundry/Grootfs) story to support OCI images based buildpacks in the future)


tscolari:
So that test broke because of the use of go < 1.8 on the tests (where the `http.Server.Close` method didn't exist). 

That's a bit misleading because the tests output says 
```
$ go version
go version go1.9 linux/amd64
``` 
but that doesn't seem to be the case.

We tried to reproduce the error with different versions of go and could only reproduce it with 1.7 and bellow. (1.9 and 1.8 was all green)

tscolari:
The PR #352 should update the version of go in the docker image

mtrmac:
Thanks.  I will take a detailed look later (hopefully tomorrow), this seem reasonable at a first glance.

(Note to self: Previous discussion in https://github.com/containers/image/pull/147 ).

oci/layout/assets/certs/accepted/cacert.crt
```diff
@@ -0,0 +1,13 @@
+-----BEGIN CERTIFICATE-----
```
mtrmac: Nit: Please use `fixtures` for the name of the subdirectory containing test fixtures, to be consistent with the other subpackages.

tscolari: :+1:
going to update it

oci/layout/oci_src_test.go
```diff
@@ -0,0 +1,133 @@
...
+		// Ensure that we only use our "CA" to validate certificates
+		ClientCAs: clientCertPool,
+	}
+
+	tlsConfig.BuildNameToCertificate()
+
+	http.HandleFunc("/", remoteLayerContent)
+
+	httpServer := &http.Server{
+		Addr:      ":12000",
```
mtrmac: Nothing guarantees that the port is available. Could this use `:0` to automatically choose an available one?

(There is also `net/http/httptest`; would it simplify things?)

oci/layout/oci_src_test.go
```diff
@@ -0,0 +1,133 @@
...
+	stopServer(httpServer, serverStopSignal)
+	os.Exit(code)
+}
+
+func TestGetBlobForRemoteLayers(t *testing.T) {
+	imageSource := createImageSource(t, &types.SystemContext{})
+	layerInfo := types.BlobInfo{
+		URLs: []string{
+			"brokenurl",
+			"https://example.com",
```
mtrmac: This is not _really_ blocking, but it would be nice for an unit test not to depend on an external service like this. 

Could this use something like `https://golang.org/src/net/http/httptest/example_test.go:ExampleServer`?

oci/layout/oci_src_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+	code := m.Run()
+
+	stopServer(httpServer, serverStopSignal)
+	os.Exit(code)
+}
+
+func TestGetBlobForRemoteLayers(t *testing.T) {
+	imageSource := createImageSource(t, &types.SystemContext{})
+	layerInfo := types.BlobInfo{
```
mtrmac: (Non-blocking: Strictly speaking, the `Digest` field is mandatory and `Size` should be `-1` if unknown. OTOH I realize that computing the digests would be fairly pointless pain.)

oci/layout/oci_src_test.go
```diff
@@ -0,0 +1,133 @@
...
+		serverStopSignal <- struct{}{}
+	}()
+
+	return httpServer, serverStopSignal, nil
+}
+
+func createImageSource(t *testing.T, context *types.SystemContext) types.ImageSource {
+	imageSource, err := newImageSource(context, ociReference{
+		dir: "assets/manifest",
+	})
```
mtrmac: (Non-blocking: Calling `NewReference` like a public user would make it obvious that there’s nothing special about constructing this reference.)

tscolari: So on this one it seems that go doesn't allow the `NewReference` return (`type.ImageReference`) to be used as an `ociReference`.

```
cannot use imageRef (type types.ImageReference) as type ociReference in argument to newImageSource: need type assertion
```

casting seems to work

oci/layout/oci_src_test.go
tscolari: Yeah, I think so, I'm going to give it a try here ^^

oci/layout/oci_src_test.go
mtrmac: The public `ociReference.NewImageSource` can be called instead.

oci/layout/oci_src_test.go
```diff
@@ -31,11 +34,20 @@ func TestMain(m *testing.M) {
...
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		fmt.Fprintln(w, "Hello world")
+	}))
+	defer ts.Close()
+
+	hash := sha256.New()
+	hash.Write([]byte("Hello world"))
 	imageSource := createImageSource(t, &types.SystemContext{})
 	layerInfo := types.BlobInfo{
+		Digest: digest.NewDigest("sha256", hash),
```
mtrmac: (FWIW this could be `digest.FromBytes([]byte(…))`. This is fine as well.)

williammartin:
@mtrmac PTAL

mtrmac:
:+1:

runcom:
LGTM

PR #354, opened 2017-10-03
Fix bug in getting external docker blob

williammartin:
This has two impacts:
1. In the case where a request succeeds, break out of the loop so as not to
continue fetching unnecessary blobs.
2. Previously it was possible for a request to fail after succeeding in a
previous iteration, which would error the function, despite a successful
fetch.

Signed-off-by: Tiago Scolari <tscolari@pivotal.io>


runcom:
@tscolari thanks!

runcom:
LGTM

mtrmac:
:+1:

Thanks; please rebase.

williammartin:
Rebased it.

PR #355, opened 2017-10-03
transports: make containers/storage stubbable

cyphar:
Currently containers/storage is only used in very specialised use-cases,
outside of the more generic usecase of converting to and from OCI
images. As a result, the fairly involved build dependencies are not
necessary for a large portion of users (for example, users that only use
skopeo as a tool for fetching OCI images and then use umoci or other
such tools to manipulate said images).

This copies the existing stubbing conventions for ostree.

### DIFF
--- a/transports/alltransports/alltransports.go
+++ b/transports/alltransports/alltransports.go
@@ -14,7 +14,7 @@ import (
 	_ "github.com/containers/image/oci/layout"
 	_ "github.com/containers/image/openshift"
 	// The ostree transport is registered by ostree*.go
-	_ "github.com/containers/image/storage"
+	// The storage transport is registered by storage*.go
 	"github.com/containers/image/transports"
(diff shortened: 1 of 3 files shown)


mtrmac:
:+1: , but tests are failing.

Cc: @nalind

cyphar:
Looks like `golint` doesn't handle imports like this:

```
// abc
import _ "blah"
```

Here's the error:

```
/gopath/src/github.com/containers/image/transports/alltransports/ostree.go:6:8: a blank import should be only in a main or test package, or have a comment justifying it
/gopath/src/github.com/containers/image/transports/alltransports/storage.go:6:8: a blank import should be only in a main or test package, or have a comment justifying it
```

I'll switch to the multi-line version I guess.

mtrmac:
Closing in favor of  #362, which is a superset. Sorry about the delay.

PR #357, opened 2017-10-11
Support MediaType on OCI Layers

williammartin:
According to the image spec [1], descriptors should contain
media types. Support them on the returned BlobInfo for layers.

[1] https://github.com/opencontainers/image-spec/blob/master/descriptor.md

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -96,6 +96,7 @@ type BlobInfo struct {
 	Size        int64         // -1 if unknown
 	URLs        []string
 	Annotations map[string]string
+	MediaType   string
 }
 
 // ImageSource is a service, possibly remote (= slow), to download components of a single image.
@@ -118,7 +119,7 @@ type ImageSource interface {
 	// out of a manifest list.
 	GetTargetManifest(digest digest.Digest) ([]byte, string, error)
 	// GetBlob returns a stream for the specified blob, and the blob’s size (or -1 if unknown).
-	// The Digest field in BlobInfo is guaranteed to be provided; Size may be -1.
+	// The Digest field in BlobInfo is guaranteed to be provided, Size may be -1 and MediaType may be optionally provided.
 	GetBlob(BlobInfo) (io.ReadCloser, int64, error)
 	// GetSignatures returns the image's signatures.  It may use a remote (= slow) service.
 	GetSignatures(context.Context) ([][]byte, error)
@@ -153,9 +154,10 @@ type ImageDestination interface {
 	AcceptsForeignLayerURLs() bool
 	// MustMatchRuntimeOS returns true iff the destination can store only images targeted for the current runtime OS. False otherwise.
 	MustMatchRuntimeOS() bool
-	// PutBlob writes contents of stream and returns data representing the result (with all data filled in).
+	// PutBlob writes contents of stream and returns data representing the result.
 	// inputInfo.Digest can be optionally provided if known; it is not mandatory for the implementation to verify it.
 	// inputInfo.Size is the expected length of stream, if known.
+	// inputInfo.MediaType describes the blob format, if known.
 	// WARNING: The contents of stream are being verified on the fly.  Until stream.Read() returns io.EOF, the contents of the data SHOULD NOT be available
 	// to any other readers for download using the supplied digest.
 	// If stream.Read() at any time, ESPECIALLY at end of input, returns an error, PutBlob MUST 1) fail, and 2) delete any data stored so far.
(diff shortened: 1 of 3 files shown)


runcom:
LGTM 

thanks @williammartin

williammartin:
This is strange, I get a 404 when I try to access this PR but when I try to
create a new one it fails. Replying via email.

Might open another PR from another branch.

williammartin:
@runcom @mtrmac It looks like maybe with Github's flakiness that nothing has poked Travis to start building, can you kick this off?

mtrmac:
ACK in principle, but looking at the references in `types.go`, quite a few assume that `BlobInfo` only has `Size` and `Digest`; could you edit them to clarify whether/how they are expected to work with the `MediaType` value?

The code base so far tends to use `MIME` instead of `Media`, unlike Moby, to be more explicit about the authority/format, unless an external format mandates e.g. a JSON field name. See e.g. `ManifestMIMEType` instances in `types.go`. (Honestly not sure whether this matters all that much.)

As for kicking the tests, unsure how to do that.

williammartin:
Sure, probably if I change something the tests will get kicked off anyway. I'll try to figure out something more consistent.

BooleanCat:
@mtrmac We've updated the BlobInfo related comments. Please take a look :)

BooleanCat:
It seems that Travis CI has marked this PR as "Abuse detected" and refuses to run tests. I've contacted Travis support.

BooleanCat:
@runcom  @mtrmac 
Could you please trigger the Travis build?

Response from Travis:

Hi Thomas,
Thanks for writing in.

We have found a bug in our abuse detection system which has been flagging users wrongly.

We're working on fixing it.

In the meantime, I have marked off the user account `containers` as trusted, so that the abuse detection checks are not run for it at all.

Could you trigger another build and let us know if it works?

Happy building!

BooleanCat:
@mtrmac Tests are green, they failed once because of a cache loading error. so I re-triggered.

tscolari:
@mtrmac could you review this after their last change?
This is blocking the last story to enable `OCI buildpacks` on Cloudfoundry! 😃

types/types.go
```diff
@@ -156,6 +157,7 @@ type ImageDestination interface {
 	// PutBlob writes contents of stream and returns data representing the result (with all data filled in).
```
mtrmac: The “all data filled in” is not currently true for return values of `PutBlob`; most (all?) implementations currently fill in only digest and size.

Same for `ReapplyBlob`.

(It _would_ be nice if it were true eventually.)

types/types.go
```diff
@@ -260,7 +262,7 @@ type ManifestUpdateOptions struct {
 // only to make writing struct literals possible.
 type ManifestUpdateInformation struct {
 	Destination  ImageDestination // and yes, UpdatedManifest may write to Destination (see the schema2 → schema1 conversion logic in image/docker_schema2.go)
-	LayerInfos   []BlobInfo       // Complete BlobInfos (size+digest) which have been uploaded, in order (the root layer first, and then successive layered layers)
+	LayerInfos   []BlobInfo       // Complete BlobInfos (size+digest+urls+mediatype+annotations) which have been uploaded, in order (the root layer first, and then successive layered layers)
```
mtrmac: This is not currently true; `copy.Image` sets `InformationOnly.LayerInfos` based on one of

- `Image.LayerInfos`, which is documented above as ”`MediaType` may be [only] optionally provided”, and says nothing about URLs/Annotations (in particular schema[12] `LayerInfos` does not set `MediaType`)
- The return value of `ReapplyBlob`/`PutBlob` (which is supposed to have “all data filled in” but actually all implementations just set digest + size)

AFAICS `LayerInfos` is currently used only in the schema1→schema2 conversion, which really only the size, so for now just documenting these two fields might be fine.

types/types.go
```diff
@@ -249,7 +251,7 @@ type Image interface {
 
 // ManifestUpdateOptions is a way to pass named optional arguments to Image.UpdatedManifest
 type ManifestUpdateOptions struct {
-	LayerInfos              []BlobInfo // Complete BlobInfos (size+digest+urls) which should replace the originals, in order (the root layer first, and then successive layered layers)
+	LayerInfos              []BlobInfo // Complete BlobInfos (size+digest+urls+mediatype+annotations) which should replace the originals, in order (the root layer first, and then successive layered layers)
```
mtrmac: This is not currently implemented that way.

Same as `ManifestUpdateInformation.LayerInfos` below, the URLs+MIME type + annotations are not currently available when called by `copy.Image`.

Also, the `UpdatedImage` implementations currently do not use the MIME type from this array (but they do use URLs+annotations). Changing the two lines would be trivial, but when `copy.Image` does not set the values, that would break things.

So, I guess, for now explicitly document that the MIME type is ignored?

BooleanCat:
@mtrmac We've addressed the comments above, let us know what you think.

mtrmac:
:+1: Thanks!

PR #358, opened 2017-10-11
Support MediaType on OCI Layers

williammartin:
According to the image spec [1], descriptors should contain
media types. Support them on the returned BlobInfo for layers.

[1] https://github.com/opencontainers/image-spec/blob/master/descriptor.md

Really strange, the last PR seems to 404 even though GitHub says it exists. Reopening from another branch. Sorry if this ends up duplicating at some point.


williammartin:
Gonna close this, just looks like github being weird.

PR #359, opened 2017-10-13
copy: avoid write() when no progress bar needed

runcom:
Observed some spikes of the progress bar while writing to `/dev/null` in CRI-O (since it has no progress bar).
This PR is a micro-optimization and will avoid a `write()` syscall when no progress bar is needed (and to keep the profiler to not show me that writing to /dev/null wastes some micro seconds).

@mrunalp @rhatdan @sameo PTAL this is needed by CRI-O post V1

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -43,7 +43,9 @@ type imageCopier struct {
 	rawSource         types.ImageSource
 	diffIDsAreNeeded  bool
 	canModifyManifest bool
-	reportWriter      io.Writer
+	writer            io.Writer
+	reportWriter      func(f string, a ...interface{})
+	hasReportWriter   bool
 	progressInterval  time.Duration
 	progress          chan types.ProgressProperties
 }
@@ -109,14 +111,20 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 		options = &Options{}
 	}
 
-	reportWriter := ioutil.Discard
+	var (
+		hasReportWriter bool
+		writer          io.Writer
+	)
 
 	if options.ReportWriter != nil {
-		reportWriter = options.ReportWriter
+		writer = options.ReportWriter
+		hasReportWriter = true
 	}
-
 	writeReport := func(f string, a ...interface{}) {
-		fmt.Fprintf(reportWriter, f, a...)
+		if !hasReportWriter {
+			return
+		}
+		fmt.Fprintf(writer, f, a...)
 	}
 
 	dest, err := destRef.NewImageDestination(options.DestinationCtx)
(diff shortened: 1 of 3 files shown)


sameo:
LGTM

TomSweeneyRedHat:
The changes look OK to me, but TRAVIS CI isn't happy with them.

mtrmac:
@runcom Could you explain this in more detail? What “spikes” in what progress bar (if it goes to `ioutil.Discard`, there is no progress bar AFAICT)

What `write()` syscall? Looking at https://golang.org/src/io/ioutil/ioutil.go?s=3442:3483#L105 `devNull.write()`, no sys call happens, so the syscall behavior should be exactly the same. The `/dev/null` path is not used at all AFAICT.

The only thing this PR actually does is avoid calling `fmt.Fprintf` and computing the output string.

Or is the important part in avoiding the `pb.New` progress bar? But even that seems to only use the provided `io.Writer`.

runcom:
> The only thing this PR actually does is avoid calling fmt.Fprintf and computing the output string.
> 
> Or is the important part in avoiding the pb.New progress bar? But even that seems to only use the provided io.Writer.

well yeah, I observed that when Discard is used, all the pb stuff were actually called and re-called. This PR basically avoids that as the go profiler showed that as being a busy path.

copy/copy.go
```diff
@@ -43,7 +43,9 @@ type imageCopier struct {
 	rawSource         types.ImageSource
 	diffIDsAreNeeded  bool
 	canModifyManifest bool
-	reportWriter      io.Writer
+	writer            io.Writer
+	reportWriter      func(f string, a ...interface{})
+	hasReportWriter   bool
```
mtrmac: It’s not immediately clear what is the difference between `writer`/`reportWriter` here. The naming is a bit confusing, as is `hasReportWriter` for a `reportWriter` which always exists

And if this is moving all all users the function, and the `io.Writer` is used _only_ for initializing the progress bar, we don’t really need `hasReportWriter`; we could just have an `io.Writer` which is either an implementation or `nil`.

copy/copy.go
```diff
@@ -109,14 +111,20 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
...
+		writer          io.Writer
+	)
 
 	if options.ReportWriter != nil {
-		reportWriter = options.ReportWriter
+		writer = options.ReportWriter
+		hasReportWriter = true
 	}
-
 	writeReport := func(f string, a ...interface{}) {
```
mtrmac: See also https://github.com/containers/image/pull/346/commits/87283c0b56acecab0f749db4fc89e4863a6db52d (“Add `copier.Printf`” from #346).  Not directly applicable, but using a real method, and using a name which is recognized by `go tool vet` would be nice.

copy/copy.go
```diff
@@ -574,14 +584,16 @@ func (ic *imageCopier) copyBlobFromStream(srcStream io.Reader, srcInfo types.Blo
...
 	// === Report progress using a pb.Reader.
-	bar := pb.New(int(srcInfo.Size)).SetUnits(pb.U_BYTES)
-	bar.Output = ic.reportWriter
-	bar.SetMaxWidth(80)
-	bar.ShowTimeLeft = false
-	bar.ShowPercent = false
-	bar.Start()
-	destStream = bar.NewProxyReader(destStream)
-	defer bar.Finish()
+	if ic.hasReportWriter {
```
mtrmac: If making the progress bar conditional is that important for performance, this needs a comment explaining why it is conditional; otherwise in a few months someone will notice that with an `ioutil.Discard` writer implementation the conditional is redundant and remove it again.

PR #360, opened 2017-10-13
ostree: support it as an image source and drop any usage of the ostree CLI tool

giuseppe:


ostree/ostree_src.go
```diff
@@ -0,0 +1,168 @@
...
+func (s *ostreeImageSource) getLayerSize(blob string) (int64, error) {
+	b := fmt.Sprintf("ociimage/%s", blob)
+	// Use golang bindings once they support to read metadata
+	out, err := exec.Command("ostree", "show", "--repo", s.ref.repo, "--print-metadata-key=docker.size", b).CombinedOutput()
+	if err != nil {
+		fmt.Printf("FAIL1\n")
+		return 0, err
+	}
+
+	size, err := strconv.ParseInt(string(out[1:len(out)-2]), 10, 64)
```
mtrmac: The `1:len(out)-2` is pretty magic, and fragile. Could this be expressed in safer terms, perhaps splitting trailing '\n' and white space?

(Applies to all instances of that.)

(I guess this would go away with using the bindings directly.)

ostree/ostree_src.go
```diff
@@ -0,0 +1,168 @@
...
+// GetManifest returns the image's manifest along with its MIME type (which may be empty when it can't be determined but the manifest is available).
+// It may use a remote (= slow) service.
+func (s *ostreeImageSource) GetManifest() ([]byte, string, error) {
+	b := fmt.Sprintf("ociimage/%s", s.ref.branchName)
+	// Use golang bindings once they support to read metadata
+	out, err := exec.Command("ostree", "show", "--repo", s.ref.repo, "--print-metadata-key=docker.manifest", b).Output()
+	if err != nil {
+		return nil, "", err
+	}
+	m := []byte(strings.Replace(string(out[1:len(out)-2]), "\\n", "\n", -1))
```
mtrmac: Is the `strings.Replace` safe? This seems ambiguous. What if the manifest contains `"ignoredKey": "Foo\nBar"`, will it be corrupted?

(Again, perhaps will go away with the golfing bindings.)

ostree/ostree_src.go
```diff
@@ -0,0 +1,168 @@
...
+	return m, manifest.GuessMIMEType(m), nil
+}
+
+func (s *ostreeImageSource) GetTargetManifest(digest digest.Digest) ([]byte, string, error) {
+	return nil, imgspecv1.MediaTypeImageManifest, nil
```
mtrmac: This should probably fail with a readable error message instead.

ostree/ostree_src.go
```diff
@@ -0,0 +1,168 @@
...
+func (s *ostreeImageSource) checkout(dest string, checksum string) error {
+	opts := otbuiltin.NewCheckoutOptions()
+	opts.UserMode = true
+	commit := fmt.Sprintf("ociimage/%s", checksum)
+	return otbuiltin.Checkout(s.ref.repo, dest, commit, opts)
+}
+
+// GetBlob returns a stream for the specified blob, and the blob's size.
+func (s *ostreeImageSource) GetBlob(info types.BlobInfo) (io.ReadCloser, int64, error) {
+	blob := strings.Replace(string(info.Digest), "sha256:", "", 1)
```
mtrmac: Should this be `info.Digest.Hex()`? Or is any non-`sha256` digest supposed to be preserved including the prefix?

ostree/ostree_src.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	// if tarsplit is nil we are looking at the manifest.  Return directly the file in layer/content
+	if tarsplit == nil {
+		stream, err := os.OpenFile(fmt.Sprintf("%s/content", layer), os.O_RDONLY, 0)
+		if err != nil {
+			return nil, 0, err
+		}
+		return ioutils.NewReadCloserWrapper(stream, func() error {
+			return os.RemoveAll(dir)
```
mtrmac: All the error paths also need to remove `dir`.

ostree/ostree_src.go
giuseppe: yes that would go away with the bindings.  Anyway, I agree it is quite hard to follow so I've pushed a new version with this part cleaned up.  The `out` string is the output of `g_variant_print` from OSTree plus a `\n`.  With this code I drop the quotes around the JSON file plus the final `\n`.

In the new version, I use `g_variant_parse` so that we don't have to deal how it is quoted.

ostree/ostree_src.go
giuseppe: this is also fixed with `g_variant_parse`.

ostree/ostree_src.go
giuseppe: yes, fixed now.

giuseppe:
I've dropped the WIP and addressed the comments

ostree/ostree_dest.go
```diff
@@ -22,6 +22,11 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/ostreedev/ostree-go/pkg/otbuiltin"
+
```
mtrmac: (Nit: Please keep the imports organized in two sections, imports from the default Golang library, and all external imports.)

ostree/ostree_src.go
```diff
@@ -0,0 +1,188 @@
...
+// #include <glib-object.h>
+// #include <gio/gio.h>
+// #include <stdlib.h>
+import "C"
+
+func parseGVariantString(in string) string {
+	cstring := C.CString(in)
+	defer C.free(unsafe.Pointer(cstring))
+
+	ptr := (*C.char)(C.g_variant_get_string(C.g_variant_parse(nil, (*C.gchar)(cstring), nil, nil, nil), nil))
```
mtrmac: Doesn’t `g_variant_unref` need to be called?

ostree/ostree_src.go
mtrmac: (Still outstanding.)

ostree/ostree_src.go
mtrmac: (This would need less repetition with a `defer func() {if !succeeded { os.RemoveAll(dir) } }()` or so. This still works I guess.)

ostree/ostree_src.go
```diff
@@ -0,0 +1,188 @@
...
+	metaUnpacker := storage.NewJSONUnpacker(mfz)
+
+	ots := asm.NewOutputTarStream(storage.NewPathFileGetter(layer), metaUnpacker)
+
+	pipeReader, pipeWriter := io.Pipe()
+	gzipWriter := gzip.NewWriter(pipeWriter)
+	go func() {
+		io.Copy(gzipWriter, ots)
+		gzipWriter.Close()
+		pipeWriter.Close()
```
mtrmac: Hum, trying to re-create a compressed blob by re-creating the uncompressed tarstream and compressing it doesn’t work long-term and is not supportable.  (We did try to do this earlier, for `containers-storage`, IIRC; it failed after a few months.)  The `gzip` bitstream is not stable over Golang versions, meaning that this will not be able to reproduce the compressed blobs of images compressed by earlier Golang versions, and after any future Golang upgrade a previously working `GetBlob` can start returning different data; in either case the blob won’t match the digest recorded in the manifest.

For me, this inability to handle compressed data is **by no means** a blocker for this PR; `containers-storage` has the same problem, and that will need to be solved by editing the manifest.  See #305 and the conversations in there.

What to do for now, though? I would be fine with this PR just skipping the compression step, meaning that the `ostree:` source produces usable images only for originally uncompressed images, and dealing with the compression as a follow-up.

That follow-up can probably be built around the tentative c/i/manifest subpackage being explored in #305.

Cc: @nalind

ostree/ostree_src.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+type ostreeImageSource struct {
+	ref    ostreeReference
+	tmpDir string
+}
+
+// newImageSource returns an ImageSource for reading from an existing directory.
+func newImageSource(ctx *types.SystemContext, tmpDir string, ref ostreeReference) (types.ImageSource, error) {
+	return &ostreeImageSource{ref: ref, tmpDir: tmpDir}, nil
```
mtrmac: (Does the `tmpDir` need to be user-specifiable? In #228, `OSTreeTmpDirPath` was introduced to affect the SELinux labels of the committed files; here the directory is used only to temporarily store a tarball, which AFAICT does not need any particular SELinux label, as long as this process can read and write the file.

Non-blocking, this does work fine.)

giuseppe: yes true.  As you say, the `tmpDir` used should not affect the result.  Also, in the long term, I'd like that the golang bindings offer some nice ways to browse/request files from the ostree repository without extracting them to a temporary directory first

ostree/ostree_src.go
giuseppe: it is something that is biting us also when we import layers from the local Docker engine, as we compute the checksum using whatever gzip implementation is used by the golang version used to build https://github.com/projectatomic/atomic/blob/master/dockertar-sha256-helper.go.  In anycase, we should probably use `skopeo copy` in atomic for accessing the local Docker engine.

giuseppe:
@mtrmac I've just rebased the PR

giuseppe:
@mtrmac there are a few changes accumulated here, anything blocking?

runcom:
LGTM

@mtrmac ?

giuseppe:
is there anything more blocking this?

ostree/ostree_src.go
```diff
@@ -0,0 +1,350 @@
...
+	return m, manifest.GuessMIMEType(m), nil
+}
+
+func (s *ostreeImageSource) GetTargetManifest(digest digest.Digest) ([]byte, string, error) {
+	return nil, "", errors.New("manifest lists are not supported by this transport")
+}
+
+func openRepo(path string) (*C.struct_OstreeRepo, error) {
+	var cerr *C.GError
+	cpath := C.CString(path)
```
mtrmac: Doesn’t `cpath` need to be freed?

ostree/ostree_src.go
```diff
@@ -0,0 +1,350 @@
...
+	pathc := C.g_file_new_for_path(cpath)
+	defer C.g_object_unref(C.gpointer(pathc))
+
+	repo := C.ostree_repo_new(pathc)
+	r := glib.GoBool(glib.GBoolean(C.ostree_repo_open(repo, nil, &cerr)))
+	if !r {
+		return nil, glib.ConvertGError(glib.ToGError(unsafe.Pointer(cerr)))
```
mtrmac: Does `repo` need to be freed here?

ostree/ostree_src.go
```diff
@@ -0,0 +1,350 @@
...
+	data := (*[1 << 30]byte)(unsafe.Pointer(C.g_bytes_get_data(b, nil)))[:count:count]
+	copy(p, data)
+	return count, nil
+}
+
+func readMetadata(repo *C.struct_OstreeRepo, commit, key string) (bool, string, error) {
+	var cerr *C.GError
+	defer func() {
+		if cerr != nil {
+			C.g_error_free(cerr)
```
mtrmac: Most other `err *C.Gerror` declarations do not have this `defer…g_error_free` block; shouldn’t they all be consistent? Or is this one somehow special?

ostree/ostree_src.go
```diff
@@ -0,0 +1,350 @@
...
+// #include <gio/gio.h>
+// #include <stdlib.h>
+// #include <ostree.h>
+// #include <gio/ginputstream.h>
+import "C"
+
+type ostreeImageSource struct {
+	ref    ostreeReference
+	tmpDir string
+	repo   *C.struct_OstreeRepo
```
mtrmac: Is `repo` ever closed?

ostree/ostree_dest.go
```diff
@@ -47,6 +64,8 @@ type ostreeImageDestination struct {
 	tmpDirPath string
 	blobs      map[string]*blobToImport
 	digest     digest.Digest
+	signatures int
+	repo       *C.struct_OstreeRepo
```
mtrmac: Is `repo` ever closed?

ostree/ostree_dest.go
```diff
@@ -47,6 +64,8 @@ type ostreeImageDestination struct {
 	tmpDirPath string
 	blobs      map[string]*blobToImport
 	digest     digest.Digest
+	signatures int
```
mtrmac: Please name this `signaturesSize`, `signaturesLen` or something like that.

ostree/ostree_dest.go
```diff
@@ -35,9 +48,13 @@ type descriptor struct {
...
 
+type fsLayersSchema1 struct {
+	BlobSum digest.Digest `json:"blobSum"`
+}
+
 type manifestSchema struct {
-	ConfigDescriptor  descriptor   `json:"config"`
-	LayersDescriptors []descriptor `json:"layers"`
+	LayersDescriptors []descriptor      `json:"layers"`
+	FSLayers          []fsLayersSchema1 `json:"fsLayers"`
```
mtrmac: It would be cleaner to use the future `c/image/manifest` subpackage to parse the manifest… but it’s doesn’t break any other transports, so non-blocking I guess.

ostree/ostree_src.go
```diff
@@ -0,0 +1,350 @@
...
+}
+
+func newOSTreePathFileGetter(repo *C.struct_OstreeRepo, commit string) (*ostreePathFileGetter, error) {
+	var cerr *C.GError
+	var parentRoot *C.GFile
+	if !glib.GoBool(glib.GBoolean(C.ostree_repo_read_commit(repo, C.CString(commit), &parentRoot, nil, nil, &cerr))) {
+		return &ostreePathFileGetter{}, glib.ConvertGError(glib.ToGError(unsafe.Pointer(cerr)))
+	}
+
+	return &ostreePathFileGetter{repo: repo, parentRoot: parentRoot}, nil
```
mtrmac: It would be a bit more GLib-like to take an extra reference to `repo`, and to `g_object_unref` it in `.Close`

ostree/ostree_src.go
```diff
@@ -0,0 +1,350 @@
...
+			return true, val, nil
+		}
+	}
+	return false, "", nil
+}
+
+func newOSTreePathFileGetter(repo *C.struct_OstreeRepo, commit string) (*ostreePathFileGetter, error) {
+	var cerr *C.GError
+	var parentRoot *C.GFile
+	if !glib.GoBool(glib.GBoolean(C.ostree_repo_read_commit(repo, C.CString(commit), &parentRoot, nil, nil, &cerr))) {
```
mtrmac: Does `C.CString(commit)` need to be freed?

ostree/ostree_src.go
```diff
@@ -0,0 +1,350 @@
...
+
+func (o ostreePathFileGetter) Get(filename string) (io.ReadCloser, error) {
+	var file *C.GFile
+	if strings.HasPrefix(filename, "./") {
+		filename = filename[2:]
+	}
+	cfilename := C.CString(filename)
+	defer C.free(unsafe.Pointer(cfilename))
+
+	file = (*C.GFile)(C.g_file_resolve_relative_path(o.parentRoot, cfilename))
```
mtrmac: Does `file` need to be freed?

giuseppe:
@mtrmac thanks for the detailed review.  I've addressed the comments in the version here ⬆️

ostree/ostree_dest.go
mtrmac: ping

ostree/ostree_src.go
mtrmac: Looking at https://github.com/ostreedev/ostree-go/blob/master/pkg/glibobject/gerror.go , `ConvertGError` on the error paths calls `g_error_free` as well. So it seems that these wrappers are designed to be used in some other way. (\*sigh\* undocumented, of course.)

ostree/ostree_src.go
```diff
@@ -0,0 +1,377 @@
...
+	stream := C.g_file_read(file, nil, &cerr)
+	if stream == nil {
+		return nil, glib.ConvertGError(glib.ToGError(unsafe.Pointer(cerr)))
+	}
+
+	return &ostreeReader{stream: stream}, nil
+}
+
+func (o ostreePathFileGetter) Close() {
+	C.g_object_ref(C.gpointer(o.repo))
```
mtrmac: This should be `g_object_unref`

ostree/ostree_src.go
```diff
@@ -0,0 +1,377 @@
...
+import "C"
+
+type ostreeImageSource struct {
+	ref    ostreeReference
+	tmpDir string
+	repo   *C.struct_OstreeRepo
+}
+
+// newImageSource returns an ImageSource for reading from an existing directory.
+func newImageSource(ctx *types.SystemContext, tmpDir string, ref ostreeReference) (types.ImageSource, error) {
```
mtrmac: (`ctx` is not used and could be dropped. Non-blocking.)

giuseppe:
@mtrmac if I simply drop the `gzipWriter` then the digest doesn't match:

```
FATA[0011] Error writing blob: Digest did not match, expected sha256:76b0264c130a031b05e695bf175147575b93271c9c838408ba63daaa7de934e9, got sha256:f4c74f205954fed1127474d1d58614639a0d560fe3bab9c9c1abd51ad7426fe7 
```

is there anything more to do to get it working with the output from `tar-split`?

mtrmac:
> @mtrmac if I simply drop the `gzipWriter` then the digest doesn't match:

It “in principle” does not match with the `gzipWriter` either; it just happens to work for your particular combination of golang version and source image:

> It is likely enough to work for a month or so, enough for users to build features on top and assume everything is fine, and break in a few golang RPM upgrades.
>
> I think I’d rather have a source which transparently always fails to reproduce compressed blobs than one that works long enough for users to rely on it when we know we can’t support this long-term. (@runcom WDYT?)

The work in #305 should make this possible one way or another; until then, I’d prefer not even trying here (but @runcom can persuade me otherwise).

giuseppe:
can this be changed once #305 is merged?  It is really unfortunate that the checksum is calculated on the compressed file, we are already using this assumption in `atomic`

mtrmac:
It _doesn’t work_. Whatever you are building that relies on this recompression doesn’t work reliably (and after #305 happens still will neither return compressed data nor return the original checksum).

giuseppe:
it doesn't depend on the compressed data, as long as we will be able to provide a modified manifest file that will match the uncompressed data.

giuseppe:
I've added another patch to handle only uncompressed layers for now.  When #305 will be merged, then I'll work on supporting compressed layers as well.

runcom:
> The work in #305 should make this possible one way or another; until then, I’d prefer not even trying here (but @runcom can persuade me otherwise).

well, #305 is needed by CRI-O as well, so it won't be that _long_ till we merge that hopefully. That PR is actually fixing 2 bugs in CRI-O, meanwhile, I don't have any issue with this PR as long as when #305 is merged, this is reworked to take into account what's missing.

@mtrmac sounds good?

giuseppe:
@mtrmac are you still fine if I drop the "ostree: support only uncompressed images for now" patch or is it necessary to get this PR approved?

mtrmac:
I’d still _prefer_ the compression not to be included; I can’t see that making an _incorrect impression_ that this works is valuable at all, and the post-#305 work will not benefit from the compression either.  But, \*shrug\* if @runcom is fine with having that in, I won’t block on that.

However you decide, please let me know when this is ready for merging.

giuseppe:
ok then, let's merge without the support for compressed images for now.  I will take another look if and how it can be supported after #305 is merged.  Thanks again for the reviews

mtrmac:
👍 

Thanks!

giuseppe:
do I need to re-vendor Skopeo or is it going to happen anyway before a new release?

mtrmac:
It is very likely to happen, but we don’t have a process that absolutely guarantees it.  If you want to be sure and submit a skopeo PR (just run `vndr` and commit), go ahead.

PR #363, opened 2017-10-20
storage: try to select an image ID when writing

nalind:
When saving a blob for an image, if an ID for the image wasn't specified, try to use the hex portion of the digest of the image's configuration blob as an ID.  In order to read that, we implement part of types.ImageSource on destination images.

### DIFF
--- a/storage/storage_image.go
+++ b/storage/storage_image.go
@@ -322,12 +322,58 @@ func (s *storageImageDestination) ReapplyBlob(blobinfo types.BlobInfo) (types.Bl
 	return s.putBlob(rc, blobinfo, false)
 }
 
+// GetBlob is only here so that it can be used to read the image's configuration blob, if there is one,
+// so it does not know how to read layers.
+func (s *storageImageDestination) GetBlob(blobinfo types.BlobInfo) (io.ReadCloser, int64, error) {
+	if blobinfo.Digest == "" {
+		return nil, -1, errors.Errorf(`Can not read a blob with unknown digest`)
+	}
+	if blob, ok := s.BlobData[blobinfo.Digest]; ok {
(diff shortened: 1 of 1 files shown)


storage/storage_image.go
```diff
@@ -330,6 +366,16 @@ func (s *storageImageDestination) Commit() error {
 			lastLayer = layerList[len(layerList)-1]
 		}
 	}
+	if s.ID == "" {
```
runcom: Why can't we always use the config blob digest? It should be present for every image afaict and will allow us to group tags by id

nalind: schema1 images, like `kubernetes/pause`, don't have config blobs.

runcom: Mmm how is docker calculating the correct content addressable image ID for schema 1 images? We could copy that

mtrmac: Docker converts the schema1 manifest into a config, and uses the digest of that as an ID.

I think the `c/i/image` conversion is very similar, but it’s not tested to generate exactly the same IDs, and I don’t even know whether there is any expectation of this conversion to be stable and frozen forever; I’d be a bit surprised if it were.

mtrmac:
Couldn’t this, like #305, instead built around the `manifest-editing`-like `c/i/manifest` subpackage, and just parsing the manifest, instead of involving all of this machinery?

nalind:
@mtrmac Yup, this is a stab at providing something for https://github.com/kubernetes-incubator/cri-o/pull/654#issuecomment-338254868, but it would be replaced by logic in #305.  If we're going to get the same IDs for schema 1 images that dockerd computes, we'll also need diffIDs, so this isn't really going to work.  Closing.

PR #374, opened 2017-11-03
Stub out the right transport name for storage

nalind:
We never did rename the `containers-storage:` transport (per https://github.com/containers/image/blob/master/storage/storage_transport.go#L70, give or take), but the stub thought that we had.


rhatdan:
LGTM

runcom:
LGTM

rhatdan:
@mtrmac PTAL
Would like to get this merged.

mtrmac:
:+1:

Doh, it never occurred to me to check this.  Thanks.

PR #376, opened 2017-11-07
Use a nil *http.Client for unix:// sockets

mtrmac:
`dockerclient.NewClient` requires `http.Client` to be set peculiarly for `unix://` sockets (like the default one).  Instead of doing that, just pass `nil` if the host is a `PF_UNIX` socket because all our overrides are TLS-related, hence inapplicable in that case.

Fixes https://github.com/projectatomic/skopeo/issues/445 .

Cc: @hferentschik

### DIFF
--- a/docker/daemon/client.go
+++ b/docker/daemon/client.go
@@ -1,11 +1,12 @@
 package daemon
 
 import (
+	"net/http"
+	"path/filepath"
+
 	"github.com/containers/image/types"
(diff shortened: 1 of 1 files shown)


runcom:
LGTM

hferentschik:
LGTM. So we should have kept the _nil_ after all ;-)

mtrmac:
:+1:

PR #377, opened 2017-11-07
Add manifest type conversion support for "dir" transport

umohnani8:
Gives the user the option to switch between the oci, v2s1, and v2s2 manifests
Add directory creation if the directory doesn't exist
Add overwrite functionality only if the directory was a container image directory
Feature added for the pulp team

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -12,8 +12,6 @@ import (
 	"strings"
 	"time"
 
-	pb "gopkg.in/cheggaaa/pb.v1"
-
 	"github.com/containers/image/image"
 	"github.com/containers/image/pkg/compression"
 	"github.com/containers/image/signature"
@@ -22,6 +20,7 @@ import (
 	"github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	pb "gopkg.in/cheggaaa/pb.v1"
 )
 
 type digestingReader struct {
@@ -95,6 +94,8 @@ type Options struct {
 	DestinationCtx   *types.SystemContext
 	ProgressInterval time.Duration                 // time to wait between reports to signal the progress channel
 	Progress         chan types.ProgressProperties // Reported to when ProgressInterval has arrived for a single artifact+offset.
+	// manifest MIME type of image set by user. "" is default and means use the autodetection to the the manifest MIME type
+	ForceManifestMIMEType string
 }
 
 // Image copies image from srcRef to destRef, using policyContext to validate
@@ -193,7 +194,7 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 
 	// We compute preferredManifestMIMEType only to show it in error messages.
 	// Without having to add this context in an error message, we would be happy enough to know only that no conversion is needed.
-	preferredManifestMIMEType, otherManifestMIMETypeCandidates, err := determineManifestConversion(&manifestUpdates, src, dest.SupportedManifestMIMETypes(), canModifyManifest)
+	preferredManifestMIMEType, otherManifestMIMETypeCandidates, err := determineManifestConversion(&manifestUpdates, src, dest.SupportedManifestMIMETypes(), canModifyManifest, options.ForceManifestMIMEType)
 	if err != nil {
 		return err
 	}
--- a/copy/manifest.go
+++ b/copy/manifest.go
@@ -41,10 +41,14 @@ func (os *orderedSet) append(s string) {
 // Note that the conversion will only happen later, through src.UpdatedImage
 // Returns the preferred manifest MIME type (whether we are converting to it or using it unmodified),
 // and a list of other possible alternatives, in order.
-func determineManifestConversion(manifestUpdates *types.ManifestUpdateOptions, src types.Image, destSupportedManifestMIMETypes []string, canModifyManifest bool) (string, []string, error) {
+func determineManifestConversion(manifestUpdates *types.ManifestUpdateOptions, src types.Image, destSupportedManifestMIMETypes []string, canModifyManifest bool, forceManifestMIMEType string) (string, []string, error) {
 	_, srcType, err := src.Manifest()
 	if err != nil { // This should have been cached?!
 		return "", nil, errors.Wrap(err, "Error reading manifest")
+	}
+
+	if forceManifestMIMEType != "" {
+		destSupportedManifestMIMETypes = []string{forceManifestMIMEType}
 	}
 
 	if len(destSupportedManifestMIMETypes) == 0 {
--- a/copy/manifest_test.go
+++ b/copy/manifest_test.go
@@ -136,7 +136,7 @@ func TestDetermineManifestConversion(t *testing.T) {
 	for _, c := range cases {
 		src := fakeImageSource(c.sourceType)
 		mu := types.ManifestUpdateOptions{}
-		preferredMIMEType, otherCandidates, err := determineManifestConversion(&mu, src, c.destTypes, true)
+		preferredMIMEType, otherCandidates, err := determineManifestConversion(&mu, src, c.destTypes, true, "")
 		require.NoError(t, err, c.description)
 		assert.Equal(t, c.expectedUpdate, mu.ManifestMIMEType, c.description)
 		if c.expectedUpdate == "" {
@@ -151,15 +151,26 @@ func TestDetermineManifestConversion(t *testing.T) {
 	for _, c := range cases {
 		src := fakeImageSource(c.sourceType)
 		mu := types.ManifestUpdateOptions{}
-		preferredMIMEType, otherCandidates, err := determineManifestConversion(&mu, src, c.destTypes, false)
+		preferredMIMEType, otherCandidates, err := determineManifestConversion(&mu, src, c.destTypes, false, "")
 		require.NoError(t, err, c.description)
 		assert.Equal(t, "", mu.ManifestMIMEType, c.description)
 		assert.Equal(t, c.sourceType, preferredMIMEType, c.description)
 		assert.Equal(t, []string{}, otherCandidates, c.description)
 	}
 
+	// With forceManifestMIMEType, the output is always the forced manifest type (in this case oci manifest)
+	for _, c := range cases {
+		src := fakeImageSource(c.sourceType)
+		mu := types.ManifestUpdateOptions{}
+		preferredMIMEType, otherCandidates, err := determineManifestConversion(&mu, src, c.destTypes, true, v1.MediaTypeImageManifest)
+		require.NoError(t, err, c.description)
+		assert.Equal(t, v1.MediaTypeImageManifest, mu.ManifestMIMEType, c.description)
+		assert.Equal(t, v1.MediaTypeImageManifest, preferredMIMEType, c.description)
+		assert.Equal(t, []string{}, otherCandidates, c.description)
+	}
+
 	// Error reading the manifest — smoke test only.
 	mu := types.ManifestUpdateOptions{}
-	_, _, err := determineManifestConversion(&mu, fakeImageSource(""), supportS1S2, true)
+	_, _, err := determineManifestConversion(&mu, fakeImageSource(""), supportS1S2, true, "")
 	assert.Error(t, err)
 }
--- a/directory/directory_dest.go
+++ b/directory/directory_dest.go
@@ -4,19 +4,77 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"path/filepath"
 
 	"github.com/containers/image/types"
 	"github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 )
 
+const version = "Directory Transport Version: 1.0\n"
+
+// ErrNotContainerImageDir indicates that the directory doesn't match the expected contents of a directory created
+// using the 'dir' transport
+var ErrNotContainerImageDir = errors.New("not a containers image directory, don't want to overwrite important data")
+
 type dirImageDestination struct {
-	ref dirReference
+	ref      dirReference
+	compress bool
 }
 
-// newImageDestination returns an ImageDestination for writing to an existing directory.
-func newImageDestination(ref dirReference) types.ImageDestination {
-	return &dirImageDestination{ref}
+// newImageDestination returns an ImageDestination for writing to a directory.
+func newImageDestination(ref dirReference, compress bool) (types.ImageDestination, error) {
+	d := &dirImageDestination{ref: ref, compress: compress}
+
+	// If directory exists check if it is empty
+	// if not empty, check whether the contents match that of a container image directory and overwrite the contents
+	// if the contents don't match throw an error
+	dirExists, err := pathExists(d.ref.resolvedPath)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error checking for path %q", d.ref.resolvedPath)
+	}
+	if dirExists {
+		isEmpty, err := isDirEmpty(d.ref.resolvedPath)
+		if err != nil {
+			return nil, err
+		}
+
+		if !isEmpty {
+			versionExists, err := pathExists(d.ref.versionPath())
+			if err != nil {
+				return nil, errors.Wrapf(err, "error checking if path exists %q", d.ref.versionPath())
+			}
+			if versionExists {
+				contents, err := ioutil.ReadFile(d.ref.versionPath())
+				if err != nil {
+					return nil, err
+				}
+				// check if contents of version file is what we expect it to be
+				if string(contents) != version {
+					return nil, ErrNotContainerImageDir
+				}
+			} else {
+				return nil, ErrNotContainerImageDir
+			}
+			// delete directory contents so that only one image is in the directory at a time
+			if err = removeDirContents(d.ref.resolvedPath); err != nil {
+				return nil, errors.Wrapf(err, "error erasing contents in %q", d.ref.resolvedPath)
+			}
+			logrus.Debugf("overwriting existing container image directory %q", d.ref.resolvedPath)
+		}
+	} else {
+		// create directory if it doesn't exist
+		if err := os.MkdirAll(d.ref.resolvedPath, 0755); err != nil {
+			return nil, errors.Wrapf(err, "unable to create directory %q", d.ref.resolvedPath)
+		}
+	}
+	// create version file
+	err = ioutil.WriteFile(d.ref.versionPath(), []byte(version), 0755)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error creating version file %q", d.ref.versionPath())
+	}
+	return d, nil
 }
 
 // Reference returns the reference used to set up this destination.  Note that this should directly correspond to user's intent,
@@ -42,7 +100,7 @@ func (d *dirImageDestination) SupportsSignatures() error {
 
 // ShouldCompressLayers returns true iff it is desirable to compress layer blobs written to this destination.
 func (d *dirImageDestination) ShouldCompressLayers() bool {
-	return false
+	return d.compress
 }
 
 // AcceptsForeignLayerURLs returns false iff foreign layers in manifest should be actually
@@ -145,5 +203,41 @@ func (d *dirImageDestination) PutSignatures(signatures [][]byte) error {
 // - Uploaded data MAY be visible to others before Commit() is called
 // - Uploaded data MAY be removed or MAY remain around if Close() is called without Commit() (i.e. rollback is allowed but not guaranteed)
 func (d *dirImageDestination) Commit() error {
+	return nil
+}
+
+// returns true if path exists
+func pathExists(path string) (bool, error) {
+	_, err := os.Stat(path)
+	if err == nil {
+		return true, nil
+	}
+	if err != nil && os.IsNotExist(err) {
+		return false, nil
+	}
+	return false, err
+}
+
+// returns true if directory is empty
+func isDirEmpty(path string) (bool, error) {
+	files, err := ioutil.ReadDir(path)
+	if err != nil {
+		return false, err
+	}
+	return len(files) == 0, nil
+}
+
+// deletes the contents of a directory
+func removeDirContents(path string) error {
+	files, err := ioutil.ReadDir(path)
+	if err != nil {
+		return err
+	}
+
+	for _, file := range files {
+		if err := os.RemoveAll(filepath.Join(path, file.Name())); err != nil {
+			return err
+		}
+	}
 	return nil
 }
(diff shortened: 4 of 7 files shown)


umohnani8:
@mtrmac @rhatdan @runcom PTAL

directory/directory_dest.go
```diff
@@ -4,19 +4,55 @@ import (
...
+	// If directory exists check if it is empty
+	// if not empty, check whether the contents match that of a container image directory and overwrite the contents
+	// if the contents don't match throw an error
+	if dirExists(d.ref.resolvedPath) {
+		isEmpty, err := isDirEmpty(d.ref.resolvedPath)
+		if err != nil {
+			return nil, err
+		}
+
+		_, err2 := os.Stat(filepath.Join(d.ref.resolvedPath, "manifest.json"))
```
rhatdan: Why use a second err?  Just use err.

directory/directory_dest.go
```diff
@@ -4,19 +4,55 @@ import (
...
+
+		if !isEmpty && err2 != nil && os.IsNotExist(err2) {
+			return nil, errors.Errorf("%q is not empty and not a container image directory. Don't want to overwrite important data, please specifiy a different directory path", d.ref.path)
```
rhatdan: errors.Wrapf(err, ...)

rhatdan: Not sure we want the second line of error.

directory/directory_dest.go
```diff
@@ -4,19 +4,55 @@ import (
...
+		}
+		if err = removeDirContents(d.ref.resolvedPath); err != nil {
+			return nil, errors.Wrapf(err, "error erasing contents in %q", d.ref.resolvedPath)
+		}
+		logrus.Infof("overwriting existing container image directory %q", d.ref.resolvedPath)
+	} else {
+		// create directory if it doesn't exist
+		if err := os.MkdirAll(d.ref.resolvedPath, 0755); err != nil {
+			return nil, err
```
rhatdan: Wrap this error with some useful information line the directory it was not able to create.

directory/directory_dest.go
```diff
@@ -4,19 +4,55 @@ import (
...
+		}
+	}
+	// create version file
+	err := ioutil.WriteFile(filepath.Join(d.ref.path, "version"), []byte(version), 0755)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error creating version file")
```
rhatdan: Add the path to the file.filepath.Join(d.ref.path, "version"),

directory/directory_dest.go
```diff
@@ -147,3 +183,40 @@ func (d *dirImageDestination) PutSignatures(signatures [][]byte) error {
 func (d *dirImageDestination) Commit() error {
 	return nil
 }
+
+// returns true if path exists
+func dirExists(path string) bool {
+	_, err := os.Stat(path)
+	if err == nil {
```
rhatdan: Simpler to write

return err == nil

rhatdan: I think you should enhance this check.  Or rename it to pathExists,
You should handle the error, because it might be a permission denied to even see if the path exists.

directory/directory_dest.go
```diff
@@ -147,3 +183,40 @@ func (d *dirImageDestination) PutSignatures(signatures [][]byte) error {
...
+
+// deletes the contents of a directory
+func removeDirContents(path string) error {
+	files, err := ioutil.ReadDir(path)
+	if err != nil {
+		return err
+	}
+
+	for _, file := range files {
+		err := os.Remove(filepath.Join(path, file.Name()))
```
rhatdan: Should use os.RemoveAll, since this could be a directory.

directory/directory_dest.go
umohnani8: Not sure we want the "directory is not empty and not a container image directory..." error?

rhatdan: What happens if the stat fails for a reason other then the directory does not exist?

directory/directory_dest.go
umohnani8: os.RemoveAll will remove the parent directory also. I just want to remove the contents of the directory and os.Remove removes both files and directories.

directory/directory_dest.go
umohnani8: oh right, I can check for that and error out.

directory/directory_dest.go
rhatdan: I don't believe os.Remove will remove a non empty subdir.  I am suggesting you use RemoveAll on the contents of the directory, not the full directory.

directory/directory_dest.go
umohnani8: fixed

directory/directory_dest.go
umohnani8: fixed

directory/directory_dest.go
umohnani8: fixed

directory/directory_dest.go
umohnani8: fixed

directory/directory_dest.go
umohnani8: fixed

directory/directory_dest.go
umohnani8: you are right, I misunderstood the docs on os.Remove. Fixed

directory/directory_dest.go
```diff
@@ -4,19 +4,64 @@ import (
...
+			return nil, errors.Wrapf(err, "%q is not empty and not a container image directory. Don't want to overwrite important data, please specifiy a different directory path", d.ref.path)
+		}
+		if err != nil {
+			return nil, errors.Wrapf(err, "error reading directory %q", d.ref.resolvedPath)
+		}
+
+		if err = removeDirContents(d.ref.resolvedPath); err != nil {
+			return nil, errors.Wrapf(err, "error erasing contents in %q", d.ref.resolvedPath)
+		}
+		logrus.Infof("overwriting existing container image directory %q", d.ref.resolvedPath)
```
rhatdan: Are these usually wrapped in some kind of --quiet wrapper?

directory/directory_dest.go
```diff
@@ -147,3 +192,43 @@ func (d *dirImageDestination) PutSignatures(signatures [][]byte) error {
...
+
+// deletes the contents of a directory
+func removeDirContents(path string) error {
+	files, err := ioutil.ReadDir(path)
+	if err != nil {
+		return err
+	}
+
+	for _, file := range files {
+		err := os.RemoveAll(filepath.Join(path, file.Name()))
```
rhatdan: Do this in one line
if err := os.RemoveAll...

types/types.go
```diff
@@ -349,6 +349,10 @@ type SystemContext struct {
 	DockerDaemonHost string
 	// Used to skip TLS verification, off by default. To take effect DockerDaemonCertPath needs to be specified as well.
 	DockerDaemonInsecureSkipTLSVerify bool
+
+	// === dir.Transport overrides ===
+	// ShouldCompress is used to compress the blobs based on user input
+	ShouldCompress bool
```
rhatdan: I never see this variable being set?

directory/directory_dest.go
umohnani8: uhh, I don't think so.

directory/directory_dest.go
umohnani8: fixed

types/types.go
umohnani8: It is set in skopeo with the value of the compress flag from skopeo copy.

rhatdan: Perhaps we should name it ForceCompress.  Since if I don't set it to true, but the source is compressed, then the dest will be compressed and the variable looks like you should not be compressed.

umohnani8: okay, changed to ForceCompress

umohnani8:
@mtrmac @runcom PTAL

copy/manifest.go
```diff
@@ -5,14 +5,16 @@ import (
 
 	"github.com/containers/image/manifest"
 	"github.com/containers/image/types"
+	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
 
 // preferredManifestMIMETypes lists manifest MIME types in order of our preference, if we can't use the original manifest and need to convert.
+// prefer oci to v2s2 because oci is given higher priority if the source manifest is not found
```
runcom: Mmm this should be ok but I fear it's going to break something, @mtrmac?

directory/directory_dest.go
runcom: Let's move this to debug I guess

umohnani8:
@mtrmac PTAL

directory/directory_dest.go
```diff
@@ -4,19 +4,64 @@ import (
...
+		}
+
+		_, err = os.Stat(filepath.Join(d.ref.resolvedPath, "version"))
+
+		if !isEmpty && err != nil && os.IsNotExist(err) {
+			return nil, errors.Wrapf(err, "%q is not empty and not a container image directory. Don't want to overwrite important data, please specifiy a different directory path", d.ref.path)
+		}
+		if !isEmpty && err != nil {
+			return nil, errors.Wrapf(err, "error reading directory %q", d.ref.resolvedPath)
+		}
```
mtrmac: This would be more readable with a single `if !isEmpty` starting above the `version` existence check and ending here.

directory/directory_dest.go
```diff
@@ -4,19 +4,64 @@ import (
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "error checking for path %q", d.ref.resolvedPath)
+	}
+	if dirExists {
+		isEmpty, err := isDirEmpty(d.ref.resolvedPath)
+		if err != nil {
+			return nil, err
+		}
+
+		_, err = os.Stat(filepath.Join(d.ref.resolvedPath, "version"))
```
mtrmac: This should use `versionPath`.

directory/directory_dest.go
```diff
@@ -4,19 +4,64 @@ import (
...
+		logrus.Debugf("overwriting existing container image directory %q", d.ref.resolvedPath)
+	} else {
+		// create directory if it doesn't exist
+		if err := os.MkdirAll(d.ref.resolvedPath, 0755); err != nil {
+			return nil, errors.Wrapf(err, "unable to create directory %q", d.ref.resolvedPath)
+		}
+	}
+	// create version file
+	versionFile := filepath.Join(d.ref.resolvedPath, "version")
+	err = ioutil.WriteFile(versionFile, []byte(version), 0755)
```
mtrmac: This should use `versionPath`.

directory/directory_dest.go
mtrmac: This can use `pathExists` instead of `os.Stat`+`os.IsNotExists`.

directory/directory_dest.go
mtrmac: `if !isEmpty`, and the `version` file exists, this should also check that it has exactly the expected contents.

(In the future we can add version parsing and compatibility logic here; for now, a straight equality amounts to the same thing.)

directory/directory_dest.go
```diff
@@ -4,19 +4,64 @@ import (
...
+
+		if !isEmpty && err != nil && os.IsNotExist(err) {
+			return nil, errors.Wrapf(err, "%q is not empty and not a container image directory. Don't want to overwrite important data, please specifiy a different directory path", d.ref.path)
+		}
+		if !isEmpty && err != nil {
+			return nil, errors.Wrapf(err, "error reading directory %q", d.ref.resolvedPath)
+		}
+
+		if err = removeDirContents(d.ref.resolvedPath); err != nil {
+			return nil, errors.Wrapf(err, "error erasing contents in %q", d.ref.resolvedPath)
```
mtrmac: Hum, is this desirable? That means previously pulled blobs, if any, will have to be pulled again. OTOH it does mean that the directory will have consistent contents of exactly one image with no junk left around.

I really don’t know. Opinions?

directory/directory_dest.go
mtrmac: (Non-blocking: It would be conceptually a bit clearer to only create the `version` file when initializing the directory, and to not remove it in `removeDireContents` if `!isEmpty`. OTOH this is much less code, and works fine.)

directory/directory_dest.go
```diff
@@ -4,19 +4,64 @@ import (
...
+	if dirExists {
+		isEmpty, err := isDirEmpty(d.ref.resolvedPath)
+		if err != nil {
+			return nil, err
+		}
+
+		_, err = os.Stat(filepath.Join(d.ref.resolvedPath, "version"))
+
+		if !isEmpty && err != nil && os.IsNotExist(err) {
+			return nil, errors.Wrapf(err, "%q is not empty and not a container image directory. Don't want to overwrite important data, please specifiy a different directory path", d.ref.path)
```
mtrmac: Nit: `s/specifiy/specify/`

mtrmac: I don’t think this should use `errors.Wrapf`: the “`"…/version"` file does not exist” explanation does add much to the information above. Non-blocking, I can see an argument for the pointer to `version` being useful as well.

copy/copy.go
```diff
@@ -95,6 +94,7 @@ type Options struct {
 	DestinationCtx   *types.SystemContext
 	ProgressInterval time.Duration                 // time to wait between reports to signal the progress channel
 	Progress         chan types.ProgressProperties // Reported to when ProgressInterval has arrived for a single artifact+offset.
+	ManifestType     string
```
mtrmac: Please name this `ManifestMIMEType`, or perhaps `ForceManifestMIMEType`, and document it at least enough to say that `""` means use the default autodetection.

directory/directory_dest.go
```diff
@@ -147,3 +192,42 @@ func (d *dirImageDestination) PutSignatures(signatures [][]byte) error {
...
+// returns true if directory is empty
+func isDirEmpty(path string) (bool, error) {
+	files, err := ioutil.ReadDir(path)
+	if err != nil {
+		return false, err
+	}
+	if len(files) > 0 {
+		return false, nil
+	}
+	return true, nil
```
mtrmac: (Non-blocking: This could be shorter as `return len(files) == 0, nil`.)

copy/copy.go
```diff
@@ -191,9 +191,16 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 		return err
 	}
 
+	var manifestMIMEType []string
+	if options.ManifestType != "" {
+		manifestMIMEType = append(manifestMIMEType, options.ManifestType)
+	} else {
+		manifestMIMEType = dest.SupportedManifestMIMETypes()
+	}
```
mtrmac: This does work, but it requires a not-entirely-obvious inspection of `determineManifestConversion` to verify that it does.  Could you instead add a `forceManifestMIMEType` option to `determineManifestConversion`, do this in there, and perhaps add some tests to ensure that we always return exactly that one MIME type?

umohnani8:
> It’s not obvious to me how this relates to the rest of this PR; why do we want both to land at the same time?

Also I don’t understand the stated rationale:

    ”if the source manifest is not found”: that can never happen during copying
    “is given higher priority”: given by what?

This does seem risky enough that I’d prefer a separate PR.

@mtrmac the reason I added the oci manifest to that list was to default to it (give it higher priority) if the source manifest type is not found (didn't know that it was not possible), but since you say that is never possible I will change it back to what it was originally.

mtrmac:
> > It’s not obvious to me how this relates to the rest of this PR; why do we want both to land at the same time?
Also I don’t understand the stated rationale:
    ”if the source manifest is not found”: that can never happen during copying
    “is given higher priority”: given by what?
This does seem risky enough that I’d prefer a separate PR.

> @mtrmac the reason I added the oci manifest to that list was to default to it (give it higher priority) if the source manifest type is not found

If the source manifest is unrecognized, we can’t parse the layer list from it and can’t copy the image at all; nor would we be able to make any conversions. (It fails through `image.FromUnparsedImage`→`image.manifestInstanceFromBlob`.)

copy/copy.go
umohnani8: fixed

copy/copy.go
umohnani8: fixed

copy/manifest_test.go
```diff
@@ -120,12 +121,12 @@ func TestDetermineManifestConversion(t *testing.T) {
...
-			"special→s2", "this needs conversion", supportS1S2OCI, manifest.DockerV2Schema2MediaType,
-			[]string{manifest.DockerV2Schema1SignedMediaType, v1.MediaTypeImageManifest, manifest.DockerV2Schema1MediaType},
+			"special→s2", "this needs conversion", supportS1S2OCI, imgspecv1.MediaTypeImageManifest,
+			[]string{manifest.DockerV2Schema2MediaType, manifest.DockerV2Schema1SignedMediaType, manifest.DockerV2Schema1MediaType},
 		},
 		{
-			"special→s1", "this needs conversion", supportS1OCI, manifest.DockerV2Schema1SignedMediaType,
-			[]string{v1.MediaTypeImageManifest, manifest.DockerV2Schema1MediaType},
+			"special→s1", "this needs conversion", supportS1OCI, imgspecv1.MediaTypeImageManifest,
+			[]string{manifest.DockerV2Schema1SignedMediaType, manifest.DockerV2Schema1MediaType},
```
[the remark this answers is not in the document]
umohnani8: removed the change to `preferredManifestMIMETypes`. fixed

directory/directory_dest.go
```diff
@@ -4,19 +4,64 @@ import (
...
 	"os"
+	"path/filepath"
 
 	"github.com/containers/image/types"
 	"github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 )
 
+const version = "Directory Transport Version: 1.0"
```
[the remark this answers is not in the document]
umohnani8: fixed

directory/directory_dest.go
umohnani8: fixed

umohnani8: fixed

directory/directory_dest.go
umohnani8: fixed

umohnani8: fixed

directory/directory_dest.go
umohnani8: fixed

directory/directory_dest.go
umohnani8: changed it a bit

directory/directory_dest.go
umohnani8: fixed

directory/directory_dest.go
umohnani8: We agreed to overwrite the directory, so deleting all contents of the directory makes sense to me. But if anyone else has another opinion, we can change things around :)

directory/directory_dest.go
umohnani8: fixed

umohnani8: left it the way it is as non blocking.

directory/directory_dest.go
umohnani8: fixed

directory/directory_transport.go
```diff
@@ -175,3 +179,8 @@ func (ref dirReference) layerPath(digest digest.Digest) string {
 func (ref dirReference) signaturePath(index int) string {
 	return filepath.Join(ref.path, fmt.Sprintf("signature-%d", index+1))
 }
+
+// versionPAth returns a path for the version file within a directory sing our conventions.
```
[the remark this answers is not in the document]
umohnani8: fixed

types/types.go
```diff
@@ -349,6 +349,10 @@ type SystemContext struct {
 	DockerDaemonHost string
 	// Used to skip TLS verification, off by default. To take effect DockerDaemonCertPath needs to be specified as well.
 	DockerDaemonInsecureSkipTLSVerify bool
+
+	// === dir.Transport overrides ===
+	// ForceCompress compresses the image layers if set to true
+	ForceCompress bool
```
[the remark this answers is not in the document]
umohnani8: fixed

umohnani8:
@mtrmac got it! Reverted the changes I made to the `preferredManifestMIMETypes` list.

directory/directory_dest.go
```diff
@@ -4,19 +4,78 @@ import (
...
+		if err != nil {
+			return nil, err
+		}
+
+		versionExists, err := pathExists(d.ref.versionPath())
+		if err != nil {
+			return nil, errors.Wrapf(err, "error checking if path exists %q", d.ref.versionPath())
+		}
+
+		if !isEmpty {
```
mtrmac: `versionExists` is also only used `if !isEmpty`; so, move the `versionExists` creation inside `if !isEmpty` as well.

copy/manifest.go
```diff
@@ -41,12 +41,17 @@ func (os *orderedSet) append(s string) {
...
-func determineManifestConversion(manifestUpdates *types.ManifestUpdateOptions, src types.Image, destSupportedManifestMIMETypes []string, canModifyManifest bool) (string, []string, error) {
+func determineManifestConversion(manifestUpdates *types.ManifestUpdateOptions, src types.Image, destSupportedManifestMIMETypes []string, canModifyManifest bool, forceManifestMIMEType string) (string, []string, error) {
 	_, srcType, err := src.Manifest()
 	if err != nil { // This should have been cached?!
 		return "", nil, errors.Wrap(err, "Error reading manifest")
 	}
 
+	if forceManifestMIMEType != "" {
+		destSupportedManifestMIMETypes = nil
+		destSupportedManifestMIMETypes = append(destSupportedManifestMIMETypes, forceManifestMIMEType)
```
mtrmac: `destSupportedManifestMIMIETypes = []string{forceManifestMIMEType}`, please.

copy/manifest_test.go
```diff
@@ -10,6 +10,7 @@ import (
 	"github.com/containers/image/manifest"
 	"github.com/containers/image/types"
 	"github.com/opencontainers/image-spec/specs-go/v1"
+	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
```
mtrmac: This is already imported as `v1` just above, please use that one. Or rename the package import if you must, but import the package only once.

copy/manifest.go
umohnani8: fixed

copy/manifest_test.go
umohnani8: fixed

directory/directory_dest.go
umohnani8: fixed

mtrmac:
:+1: pending tests.

Thanks!

umohnani8:
@runcom PTAL

rhatdan:
LGTM

runcom:
LGTM

Issue #368, opened 2017-10-26
Windows incompatibilities in Docker daemon and OCI transport 


hferentschik:
In the Docker daemon and OCI transport, there are some operations which are not compatible with Windows. For example, _ocArchiveTransport_ makes use of non-portable tmp directory in
https://github.com/containers/image/blob/master/oci/archive/oci_transport.go#L197:

```go
    func createOCIRef(image string) (tempDirOCIRef, error) {
    	dir, err := ioutil.TempDir("/var/tmp", "oci")
        //  ...
    	return tempDirRef, nil
    }
```

There are some other places where "/var/tmp" is used which should be addressed as part of the issue. 

Another problem is trying to rename a file while it is still open. On Windows the file needs to be closed first.

mtrmac:
This was fixed in #379 . Thanks!

PR #379, opened 2017-11-09
Issue #368 Making OCI and Docker daemon transport work on Windows

hferentschik:
Signed-off-by: Hardy Ferentschik <hardy@hibernate.org>

### DIFF
--- a/docker/daemon/daemon_src.go
+++ b/docker/daemon/daemon_src.go
@@ -6,12 +6,11 @@ import (
 	"os"
 
 	"github.com/containers/image/docker/tarfile"
+	"github.com/containers/image/internal/tmpdir"
 	"github.com/containers/image/types"
 	"github.com/pkg/errors"
 	"golang.org/x/net/context"
 )
-
-const temporaryDirectoryForBigFiles = "/var/tmp" // Do not use the system default of os.TempDir(), usually /tmp, because with systemd it could be a tmpfs.
 
 type daemonImageSource struct {
 	ref             daemonReference
@@ -47,7 +46,7 @@ func newImageSource(ctx *types.SystemContext, ref daemonReference) (types.ImageS
 	defer inputStream.Close()
 
 	// FIXME: use SystemContext here.
-	tarCopyFile, err := ioutil.TempFile(temporaryDirectoryForBigFiles, "docker-daemon-tar")
+	tarCopyFile, err := ioutil.TempFile(tmpdir.TemporaryDirectoryForBigFiles(), "docker-daemon-tar")
 	if err != nil {
 		return nil, err
 	}
--- a/oci/archive/oci_transport.go
+++ b/oci/archive/oci_transport.go
@@ -4,13 +4,13 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"path/filepath"
-	"regexp"
 	"strings"
 
 	"github.com/containers/image/directory/explicitfilepath"
 	"github.com/containers/image/docker/reference"
 	"github.com/containers/image/image"
+	"github.com/containers/image/internal/tmpdir"
+	"github.com/containers/image/oci/internal"
 	ocilayout "github.com/containers/image/oci/layout"
 	"github.com/containers/image/transports"
 	"github.com/containers/image/types"
@@ -48,51 +48,12 @@ func (t ociArchiveTransport) ParseReference(reference string) (types.ImageRefere
 
 // ValidatePolicyConfigurationScope checks that scope is a valid name for a signature.PolicyTransportScopes keys
 func (t ociArchiveTransport) ValidatePolicyConfigurationScope(scope string) error {
-	var file string
-	sep := strings.SplitN(scope, ":", 2)
-	file = sep[0]
-
-	if len(sep) == 2 {
-		image := sep[1]
-		if !refRegexp.MatchString(image) {
-			return errors.Errorf("Invalid image %s", image)
-		}
-	}
-
-	if !strings.HasPrefix(file, "/") {
-		return errors.Errorf("Invalid scope %s: must be an absolute path", scope)
-	}
-	// Refuse also "/", otherwise "/" and "" would have the same semantics,
-	// and "" could be unexpectedly shadowed by the "/" entry.
-	// (Note: we do allow "/:someimage", a bit ridiculous but why refuse it?)
-	if scope == "/" {
-		return errors.New(`Invalid scope "/": Use the generic default scope ""`)
-	}
-	cleaned := filepath.Clean(file)
-	if cleaned != file {
-		return errors.Errorf(`Invalid scope %s: Uses non-canonical path format, perhaps try with path %s`, scope, cleaned)
-	}
-	return nil
+	return internal.ValidateScope(scope)
 }
-
-// annotation spex from https://github.com/opencontainers/image-spec/blob/master/annotations.md#pre-defined-annotation-keys
-const (
-	separator = `(?:[-._:@+]|--)`
-	alphanum  = `(?:[A-Za-z0-9]+)`
-	component = `(?:` + alphanum + `(?:` + separator + alphanum + `)*)`
-)
-
-var refRegexp = regexp.MustCompile(`^` + component + `(?:/` + component + `)*$`)
 
 // ParseReference converts a string, which should not start with the ImageTransport.Name prefix, into an OCI ImageReference.
 func ParseReference(reference string) (types.ImageReference, error) {
-	var file, image string
-	sep := strings.SplitN(reference, ":", 2)
-	file = sep[0]
-
-	if len(sep) == 2 {
-		image = sep[1]
-	}
+	file, image := internal.SplitPathAndImage(reference)
 	return NewReference(file, image)
 }
 
@@ -102,14 +63,15 @@ func NewReference(file, image string) (types.ImageReference, error) {
 	if err != nil {
 		return nil, err
 	}
-	// This is necessary to prevent directory paths returned by PolicyConfigurationNamespaces
-	// from being ambiguous with values of PolicyConfigurationIdentity.
-	if strings.Contains(resolved, ":") {
-		return nil, errors.Errorf("Invalid OCI reference %s:%s: path %s contains a colon", file, image, resolved)
+
+	if err := internal.ValidateOCIPath(file); err != nil {
+		return nil, err
 	}
-	if len(image) > 0 && !refRegexp.MatchString(image) {
-		return nil, errors.Errorf("Invalid image %s", image)
+
+	if err := internal.ValidateImageName(image); err != nil {
+		return nil, err
 	}
+
 	return ociArchiveReference{file: file, resolvedFile: resolved, image: image}, nil
 }
 
@@ -197,7 +159,7 @@ func (t *tempDirOCIRef) deleteTempDir() error {
 
 // createOCIRef creates the oci reference of the image
 func createOCIRef(image string) (tempDirOCIRef, error) {
-	dir, err := ioutil.TempDir("/var/tmp", "oci")
+	dir, err := ioutil.TempDir(tmpdir.TemporaryDirectoryForBigFiles(), "oci")
 	if err != nil {
 		return tempDirOCIRef{}, errors.Wrapf(err, "error creating temp directory")
 	}
--- a/oci/internal/oci_util.go
+++ b/oci/internal/oci_util.go
@@ -1,0 +1,126 @@
+package internal
+
+import (
+	"github.com/pkg/errors"
+	"path/filepath"
+	"regexp"
+	"runtime"
+	"strings"
+)
+
+// annotation spex from https://github.com/opencontainers/image-spec/blob/master/annotations.md#pre-defined-annotation-keys
+const (
+	separator = `(?:[-._:@+]|--)`
+	alphanum  = `(?:[A-Za-z0-9]+)`
+	component = `(?:` + alphanum + `(?:` + separator + alphanum + `)*)`
+)
+
+var refRegexp = regexp.MustCompile(`^` + component + `(?:/` + component + `)*$`)
+var windowsRefRegexp = regexp.MustCompile(`^([a-zA-Z]:\\.+?):(.*)$`)
+
+// ValidateImageName returns nil if the image name is empty or matches the open-containers image name specs.
+// In any other case an error is returned.
+func ValidateImageName(image string) error {
+	if len(image) == 0 {
+		return nil
+	}
+
+	var err error
+	if !refRegexp.MatchString(image) {
+		err = errors.Errorf("Invalid image %s", image)
+	}
+	return err
+}
+
+// SplitPathAndImage tries to split the provided OCI reference into the OCI path and image.
+// Neither path nor image parts are validated at this stage.
+func SplitPathAndImage(reference string) (string, string) {
+	if runtime.GOOS == "windows" {
+		return splitPathAndImageWindows(reference)
+	}
+	return splitPathAndImageNonWindows(reference)
+}
+
+func splitPathAndImageWindows(reference string) (string, string) {
+	groups := windowsRefRegexp.FindStringSubmatch(reference)
+	// nil group means no match
+	if groups == nil {
+		return reference, ""
+	}
+
+	// we expect three elements. First one full match, second the capture group for the path and
+	// the third the capture group for the image
+	if len(groups) != 3 {
+		return reference, ""
+	}
+	return groups[1], groups[2]
+}
+
+func splitPathAndImageNonWindows(reference string) (string, string) {
+	sep := strings.SplitN(reference, ":", 2)
+	path := sep[0]
+
+	var image string
+	if len(sep) == 2 {
+		image = sep[1]
+	}
+	return path, image
+}
+
+// ValidateOCIPath takes the OCI path and validates it.
+func ValidateOCIPath(path string) error {
+	if runtime.GOOS == "windows" {
+		// On Windows we must allow for a ':' as part of the path
+		if strings.Count(path, ":") > 1 {
+			return errors.Errorf("Invalid OCI reference: path %s contains more than one colon", path)
+		}
+	} else {
+		if strings.Contains(path, ":") {
+			return errors.Errorf("Invalid OCI reference: path %s contains a colon", path)
+		}
+	}
+	return nil
+}
+
+// ValidateScope validates a policy configuration scope for an OCI transport.
+func ValidateScope(scope string) error {
+	var err error
+	if runtime.GOOS == "windows" {
+		err = validateScopeWindows(scope)
+	} else {
+		err = validateScopeNonWindows(scope)
+	}
+	if err != nil {
+		return err
+	}
+
+	cleaned := filepath.Clean(scope)
+	if cleaned != scope {
+		return errors.Errorf(`Invalid scope %s: Uses non-canonical path format, perhaps try with path %s`, scope, cleaned)
+	}
+
+	return nil
+}
+
+func validateScopeWindows(scope string) error {
+	matched, _ := regexp.Match(`^[a-zA-Z]:\\`, []byte(scope))
+	if !matched {
+		return errors.Errorf("Invalid scope '%s'. Must be an absolute path", scope)
+	}
+
+	return nil
+}
+
+func validateScopeNonWindows(scope string) error {
+	if !strings.HasPrefix(scope, "/") {
+		return errors.Errorf("Invalid scope %s: must be an absolute path", scope)
+	}
+
+	// Refuse also "/", otherwise "/" and "" would have the same semantics,
+	// and "" could be unexpectedly shadowed by the "/" entry.
+	if scope == "/" {
+		return errors.New(`Invalid scope "/": Use the generic default scope ""`)
+	}
+
+	return nil
+}
--- a/oci/layout/oci_dest.go
+++ b/oci/layout/oci_dest.go
@@ -112,8 +112,11 @@ func (d *ociImageDestination) PutBlob(stream io.Reader, inputInfo types.BlobInfo
 		return types.BlobInfo{}, err
 	}
 	succeeded := false
+	explicitClosed := false
 	defer func() {
-		blobFile.Close()
+		if !explicitClosed {
+			blobFile.Close()
+		}
 		if !succeeded {
 			os.Remove(blobFile.Name())
 		}
@@ -133,8 +136,15 @@ func (d *ociImageDestination) PutBlob(stream io.Reader, inputInfo types.BlobInfo
 	if err := blobFile.Sync(); err != nil {
 		return types.BlobInfo{}, err
 	}
-	if err := blobFile.Chmod(0644); err != nil {
-		return types.BlobInfo{}, err
+
+	// On POSIX systems, blobFile was created with mode 0600, so we need to make it readable.
+	// On Windows, the “permissions of newly created files” argument to syscall.Open is
+	// ignored and the file is already readable; besides, blobFile.Chmod, i.e. syscall.Fchmod,
+	// always fails on Windows.
+	if runtime.GOOS != "windows" {
+		if err := blobFile.Chmod(0644); err != nil {
+			return types.BlobInfo{}, err
+		}
 	}
 
 	blobPath, err := d.ref.blobPath(computedDigest, d.sharedBlobDir)
@@ -144,6 +154,10 @@ func (d *ociImageDestination) PutBlob(stream io.Reader, inputInfo types.BlobInfo
 	if err := ensureParentDirectoryExists(blobPath); err != nil {
 		return types.BlobInfo{}, err
 	}
+
+	// need to explicitly close the file, since a rename won't otherwise not work on Windows
+	blobFile.Close()
+	explicitClosed = true
 	if err := os.Rename(blobFile.Name(), blobPath); err != nil {
 		return types.BlobInfo{}, err
 	}
--- a/oci/layout/oci_transport.go
+++ b/oci/layout/oci_transport.go
@@ -5,12 +5,12 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"regexp"
 	"strings"
 
 	"github.com/containers/image/directory/explicitfilepath"
 	"github.com/containers/image/docker/reference"
 	"github.com/containers/image/image"
+	"github.com/containers/image/oci/internal"
 	"github.com/containers/image/transports"
 	"github.com/containers/image/types"
 	"github.com/opencontainers/go-digest"
@@ -36,45 +36,12 @@ func (t ociTransport) ParseReference(reference string) (types.ImageReference, er
 	return ParseReference(reference)
 }
 
-// annotation spex from https://github.com/opencontainers/image-spec/blob/master/annotations.md#pre-defined-annotation-keys
-const (
-	separator = `(?:[-._:@+]|--)`
-	alphanum  = `(?:[A-Za-z0-9]+)`
-	component = `(?:` + alphanum + `(?:` + separator + alphanum + `)*)`
-)
-
-var refRegexp = regexp.MustCompile(`^` + component + `(?:/` + component + `)*$`)
-
 // ValidatePolicyConfigurationScope checks that scope is a valid name for a signature.PolicyTransportScopes keys
 // (i.e. a valid PolicyConfigurationIdentity() or PolicyConfigurationNamespaces() return value).
 // It is acceptable to allow an invalid value which will never be matched, it can "only" cause user confusion.
 // scope passed to this function will not be "", that value is always allowed.
 func (t ociTransport) ValidatePolicyConfigurationScope(scope string) error {
-	var dir string
-	sep := strings.SplitN(scope, ":", 2)
-	dir = sep[0]
-
-	if len(sep) == 2 {
-		image := sep[1]
-		if !refRegexp.MatchString(image) {
-			return errors.Errorf("Invalid image %s", image)
-		}
-	}
-
-	if !strings.HasPrefix(dir, "/") {
-		return errors.Errorf("Invalid scope %s: must be an absolute path", scope)
-	}
-	// Refuse also "/", otherwise "/" and "" would have the same semantics,
-	// and "" could be unexpectedly shadowed by the "/" entry.
-	// (Note: we do allow "/:someimage", a bit ridiculous but why refuse it?)
-	if scope == "/" {
-		return errors.New(`Invalid scope "/": Use the generic default scope ""`)
-	}
-	cleaned := filepath.Clean(dir)
-	if cleaned != dir {
-		return errors.Errorf(`Invalid scope %s: Uses non-canonical path format, perhaps try with path %s`, scope, cleaned)
-	}
-	return nil
+	return internal.ValidateScope(scope)
 }
 
 // ociReference is an ImageReference for OCI directory paths.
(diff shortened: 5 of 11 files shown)


runcom:
I have unfortunately no way to verify this on windows. Changes make sense though

docker/daemon/daemon_src.go
```diff
@@ -9,9 +9,18 @@ import (
 	"github.com/containers/image/types"
 	"github.com/pkg/errors"
 	"golang.org/x/net/context"
+	"runtime"
 )
 
-const temporaryDirectoryForBigFiles = "/var/tmp" // Do not use the system default of os.TempDir(), usually /tmp, because with systemd it could be a tmpfs.
+var temporaryDirectoryForBigFiles string
+
+func init() {
```
hferentschik: Instead of doing the same thing in multiple places, I was wondering whether there should be a shared util function _GetTemporaryDirectoryForBigFiles_ somewhere. AFAICT, there is no obvious location yet for these type of utility method. Should I create one? And if so where?

oci/layout/oci_transport.go
```diff
@@ -113,7 +114,7 @@ func NewReference(dir, image string) (types.ImageReference, error) {
 	}
 	// This is necessary to prevent directory paths returned by PolicyConfigurationNamespaces
 	// from being ambiguous with values of PolicyConfigurationIdentity.
-	if strings.Contains(resolved, ":") {
+	if strings.Contains(resolved, ":") && runtime.GOOS != "windows" {
```
hferentschik: I am not so sure about this one. Obviously, on Windows, there will be a ':' in the path. Just adding this condition works for my use cases. I am not sure what the comment tries to say in terms of PolicyConfigurationNamespaces and PolicyConfigurationIdentity. I need some guidance here on whether a simple OS check is sufficient here or whether something else needs to happen. Maybe it can at least be a first step solution, since not allowing colon ':' in the path stops this to be working on Windows right in the beginning.

hferentschik:
@mtrmac, I am trying to sort out some Windows incompatibilities which stop us to move forward with integrating containers/image into Minishift. See also https://github.com/minishift/minishift/pull/1589.

hferentschik:
@runcom 
> I have unfortunately no way to verify this on windows. Changes make sense though

Sure. It would be nice to get more CI system involved and build not only on Linux, but also on other OS types. Minsihift uses CircleCI and AppVeyor for building/testing on macOS resp. Windows. 

For now it might suffice that we are testing this change as part of our integration of containers/image into Minishift - https://github.com/minishift/minishift/pull/1589. 

It would be awesome to get this in, so that we can go ahead on our side as well.

oci/layout/oci_dest.go
```diff
@@ -133,8 +133,12 @@ func (d *ociImageDestination) PutBlob(stream io.Reader, inputInfo types.BlobInfo
 	if err := blobFile.Sync(); err != nil {
 		return types.BlobInfo{}, err
 	}
-	if err := blobFile.Chmod(0644); err != nil {
-		return types.BlobInfo{}, err
+
+	// Chmod is not supported on Windows
+	if runtime.GOOS != "windows" {
+		if err := blobFile.Chmod(0644); err != nil {
```
mtrmac: How does `Chmod` fail here?  Reading `go/src/os/file_posix.go` and `go/src/syscall/syscall_windows.go` it seems to be translated to `GetFileAttributes`+`SetFileAttributes` modifying the readonly attribute, which seems plausible.

oci/layout/oci_dest.go
```diff
@@ -144,6 +148,9 @@ func (d *ociImageDestination) PutBlob(stream io.Reader, inputInfo types.BlobInfo
 	if err := ensureParentDirectoryExists(blobPath); err != nil {
 		return types.BlobInfo{}, err
 	}
+
+	// need to explicitly close the file, since a rename won't otherwise not work on Windows
+	blobFile.Close()
```
mtrmac: Per https://golang.org/pkg/io/#Closer

> The behavior of Close after the first call is undefined. 

i.e. the `defer … blobFile.Close()` must be deactivated, similarly to the `succeeded` boolean.

docker/daemon/daemon_src.go
mtrmac: Yes, please, move this into a single implementation.

Probably a subpackage in the `internal` hierarchy, something like `containers/image/internal/tmpdir`.

I wouldn’t be opposed to making this a public subpackage under `containers/image/pkg/` either, it could be useful for other software; OTOH having such a containers-independent utility under `containers/image` feels a little weird.

oci/layout/oci_transport.go
mtrmac: The original comment is no longer applicable after 4eb8db31f00f5bcb25fe55341731df33f3067768 (and should be updated, along with `ValidatePolicyConfigurationIdentity`), but this restriction is still somewhat necessary, because the same ambiguity exists in `ParseReference`:

- `/path:busybox` means `dir="/path"`, `image="busybox"`
- `/path` and `/path:` means `dir="/path"`, `image=""`
- `/path:busybox:latest` means `dir="/path"`, `image="busybox:latest`";

Note that `dir="/path:busybox"`, `image="latest"` is impossible to express in the string form consumed by `ParseReference`, because it would be parsed the other way.  Giving up on `ParseReference` for this path and allowing `dir="/path:busybox"` only via `NewReference` would also be problematic because `StringWithinTransport()` must return a valid input for `ParseReference`.

---

I think it would be very reasonable to accept a `:` on Windows (perhaps, to be more strict, only in the second position, but the output of `explicitfilepath.ResolvePathToFullyExplicit` is supposed to be a valid path, and `:` is reserved on Windows, so it should not appear anywhere but after a drive letter.)

The flip side of that, though, is that `ParseReference` should likewise accept paths starting with a drive letter and a colon, and parse that as a part of the `dir` , not as the separator starting `image`.

`ValidatePolicyConfigurationIdentity` should also be updated - to treat the input only as a directory, the way `PolicyConfigurationIdentity` formats it now, and to reject `:` on non-Windows only.

docker/daemon/daemon_src.go
hferentschik: Ok. I think I go for the internal/tmpdir

oci/layout/oci_dest.go
hferentschik: There is a difference between os.Chmod and os.File.Chmod. Go figure. See also here https://github.com/google/skicka/issues/77. As a workaround, we can also try to use os.Chmod at this point.

oci/layout/oci_dest.go
hferentschik: hmm, one still would like defer to kick in until I can explicitly close. I'll take a look.

oci/layout/oci_transport.go
hferentschik: I am aware of the problems of ParseReference. In fact, I am not using ParseReference and use NewReference directly. That works fine when using the API programmatically. It is more an issue in the case of Skopeo or other potential tools which want to create a reference from a single string. 

As I mentioned before, I think the use of ':' is unfortunate in this case. I suggested back then to use a different separator like '@' or similar. 

I feel the ParseReference problems is a bigger issue which I'd like to defer to another issue. WDYT?

In moving this along, is a OS specific check ok, together with an adjustment to ValidatePolicyConfigurationIdentity?

oci/layout/oci_dest.go
mtrmac: Oops, missed that.  The code is fine, please just add a more detailed explanation, something like
```go
// On POSIX systems, blobFile was created with mode 0600, so we need to make it readable.
// On Windows, the “permissions of newly created files” argument to syscall.Open is
// ignored and the file is already readable; besides, blobFile.Chmod, i.e. syscall.Fchmod,
// always fails on Windows.
```

oci/layout/oci_transport.go
mtrmac: > I feel the ParseReference problems is a bigger issue which I'd like to defer to another issue. WDYT?

I’d much prefer to keep of `oci_transport.go` consistent. Otherwise there’s no telling which of the alternatives will be updated in the future to make the code consistent, i.e. what use case will break.

An OS-specific check determining the treatment of `:` is OK with me.

mtrmac: Another way to put it, if you _really_ don’t want to update `ParseReference` etc. for Windows, I guess I can live with that, but then you’d have to document the code clearly and explicitly enough to keep your work working in the future.

hferentschik:
@mtrmac I pushed an update for everything except _oci_transport.go_. I will look into the latter tomorrow.

mtrmac:
> @mtrmac I pushed an update for everything except _oci_transport.go_. 

Thanks, looks good.

oci/layout/oci_transport.go
hferentschik: @mtrmac any chance you mean _ValidatePolicyConfigurationScope_? There seems to be no _ValidatePolicyConfigurationIdentity_.

If you mean _ValidatePolicyConfigurationScope_, what is the expected input to this function. The same as to _ ParseReference_? Something like <dir>:<image>? 

I am still struggling a bit with these policies and scopes. Not quite sure how everything fits together in this area of the code.

hferentschik: @mtrmac another unrelated question. There is one more problem I am seeing which is not addressed in this pull request yet. Could you have a look at https://github.com/containers/image/blob/6e47653e2d514d64af265b69b8cc02542605fccc/copy/copy.go#L290-L290. 

As you know I am exporting a bunch of images from a Docker daemon (running in a VM), store them in an OCI tree on a Windows host and then at some stage re-import into a Docker daemon. This works fine for macOS and Linux. On Windows, I get the error which is referenced above. Effectively: 

    image operating system linux cannot be used on windows

_src.OCIConfig()_ seems to return linux, I would expect Windows though. Any ideas? I have locally commented out the call to _checkImageDestinationForCurrentRuntimeOS_, since I find it questionable either way. Moving forward though, I was wondering how to solve this? If there is no bug in __src.OCIConfig()__ , I was wondering whether it would make sense to add something to SystemContext which allows to skip the check for _ImageDestinationForCurrentRuntimeOS_. WDYT?

mtrmac: > `src.OCIConfig()` seems to return `linux`, I would expect `Windows` though

The image was created by a `dockerd` on Linux, which is what is recorded inside the image’s config. It does not matter what OS is accessing the image or where the image was copied, it is designed to run on Linux.

> I was wondering whether it would make sense to add something to `SystemContext` which allows to skip the check for `ImageDestinationForCurrentRuntimeOS`. WDYT?

Yes; I have implemented that already in https://github.com/containers/image/pull/346/commits/8074c3892f3775ec876a63f8e8e2cd5bda9aea5d for https://github.com/containers/image/pull/346 .

hferentschik: > Yes; I have implemented that already in 8074c38 for #346 .

LOL :-) Seems to be exactly what I need. Except it does not seem to be merged yet :-( 
I will for now focus on the things we discussed so far and leave the destination check for #346. It would be great to get this merged though. Anything I can do do get #346 merged?

mtrmac: > Anything I can do do get #346 merged?

It needs a review by @runcom at least; any extra reviews would be welcome as well.

hferentschik:
@mtrmac I rebased the pull request and changed _ParseReference_ as well as _ValidatePolicyConfigurationScope_ to handle Windows as well. I was tempted to refactor further, but decided against, since I did not want to change any semantics/tests for the non-Windows case. WDYT?

copy/copy.go
```diff
@@ -158,9 +158,12 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 		}
 	}()
 
-	if err := checkImageDestinationForCurrentRuntimeOS(src, dest); err != nil {
-		return err
-	}
+	// Disabling checkImageDestinationForCurrentRuntimeOS since the implementation is basing decision on whether there is a match on the current OS
```
hferentschik: @mtrmac I commented this check out. I think the current implementation is not correct and w/o the ability to control this behaviour in some way, I don't think it is really useful. 

I don't think that the OS of the host can be used to determine whether a copy can occur. The destination of the copy might be remote. Or am I missing something?

mtrmac: `dest` _already_ can choose whether this check make sense, via `ImageDestination.MustMatchRuntimeOS`; and this check is actually useful for copying to local `docker-daemon:`.

Maybe a specific transport should change its value of `MustMatchRuntimeOS`? Perhaps `docker-daemon:` _when connecting to a remote host_?

hferentschik: > dest already can choose whether this check make sense, 

Right, but it is hardcoded. In the case of the Docker daemon  _daemon_dest.go_. Given that a Docker daemon can also be remote, having _MustMatchRuntimeOS()_ always return true is preventing me to copy an image from a Windows host to a Docker daemon, even though it should be possible.

> and this check is actually useful for copying to local docker-daemon:.

Maybe. 

> Maybe a specific transport should change its value of MustMatchRuntimeOS? Perhaps docker-daemon: when connecting to a remote host?

Right. So one solution would be to me _daemonImageDestination.MustMatchRuntimeOS()_ stateful. 
In the simplest case one could just use _dockerclient.DefaultDockerHost_. If the default is used, we return _true_, if the SystemContext explicitly specifies a value for _DockerDaemonHost_ we return false (unless it maybe matches `dockerclient.DefaultDockerHost`). What I want to avoid is to make any decisions based on the IP/host specified in _ctx.DockerDaemonHost_. Maybe something like 127.0.0.1 or localhost would still make sense, but everything else is not easily verifiable.  

WDYT? Make _daemonImageDestination.MustMatchRuntimeOS()_ stateful?

hferentschik: @mtrmac Any feedback regarding the other changes in this pull request? Are we good on these?

hferentschik: @mtrmac I made _MustMatchRuntimeOS()_ conditional. WDYT?

hferentschik:
@mtrmac, @runcom What do you think about this current state of the pull request?

copy/copy.go
mtrmac: > @mtrmac I made MustMatchRuntimeOS() conditional. WDYT?

Looks good. As you say, some of the cases are difficult to determine, and most users shouldn’t hit them. If someone does hits them,. we might need to make the tests a bit more precise and change the semantics; hopefully that won’t affect any previously working users.

oci/layout/oci_transport.go
mtrmac: > @mtrmac any chance you mean _ValidatePolicyConfigurationScope_? There seems to be no _ValidatePolicyConfigurationIdentity_.

Yes, sorry.

> If you mean _ValidatePolicyConfigurationScope_, what is the expected input to this function. The same as to _ ParseReference_? Something like `<dir>:<image>`? 
>
> I am still struggling a bit with these policies and scopes. Not quite sure how everything fits together in this area of the code.

1. References: A way for the user to refer to a _single image_ in text, basically a way to express a `types.ImageReference`. There may be multiple ways to refer to a single underlying image, either by definition (`docker://busybox` vs. `docker://busybox:latest`), or by the environment (`dir:pathtodir` vs. `dir:symlinktodir`). Parsed by `Transport.ParseReference`, formatted by `ImageReference.StringWithinTransport` (i.e. the output of this must be consumable by `ParseReference` and be interpreted as the same underlying image).

2. Policy configuration scopes: A way for the user to define a _hierarchical namespace_ of images. The same underlying image must have exactly one, canonical, “identity” (which is why symlinks are resolved in `NewReference`—but note that the identity fundamentally depends on the environment, i.e. the contents of the symlink), and `PolicyConfigurationNamespaces` returns the namespaces which contain that image (i.e. typically the analogue of the set of its parent directories / paths / hostnames etc.). The values are formatted by `PolicyConfigurationIdentity` and `PolicyConfigurationNamespaces`. There is no method to parse the string back into a Go value (strings in config files are compared for equality with the outputs of these two methods), but there is a `ValidatePolicyConfigurationScope` which should allow the output of `PolicyConfiguration{Identity,Scope}` and preferably little or nothing else, to allow detecting errors in the configuration.

oci/oci.go
```diff
@@ -1 +1,91 @@
...
+	if groups == nil {
+		return reference, nil
+	}
+
+	// we expect three elements. First one full match, second the capture group for the dir and
+	// the third the capture group for the image
+	if len(groups) != 3 {
+		return reference, nil
+	}
+	return groups[1], &groups[2]
```
mtrmac: In non-Windows it is possible to use relative paths, e.g. `oci:somedir:imagename`. This seems to return a non-`nil` image only if the input is an absolute path; is that intentional?

AFAICT relative paths could be allowed without ambiguity:
- `c:\dir\…` cannot mean “the relative path `.\c:\dir`” because `c:` is an invalid directory name
- `c:\dir…` cannot mean “the image `\dir…` in the relative path `c`” because `\dir` is an invalid image name.
- The troublesome case is `c:dir`, which could mean “the relative path `dir` on the current directory of `c:`” or “the image `dir` in the relative path `c`”, but at least https://blogs.msdn.microsoft.com/oldnewthing/20101011-00/?p=12563 says this concept does not exist in Windows anymore.

The above could be non-blocking, left to someone in the future to improve, because the Windows support is new; more troublesome is that this parses `somedir:busybox:latest` as path `somedir:busybox:latest`, missing image, which would then fail validation in `NewReference`; and, worse, `somedir:busybox` as path `somedir:busybox`, no image, which would succeed but be inconsistent with the POSIX version.

oci/oci.go
```diff
@@ -1 +1,91 @@
...
+}
+
+func validateScopeNonWindows(scope string, dir string) error {
+	if !strings.HasPrefix(dir, "/") {
+		return errors.Errorf("Invalid scope %s: must be an absolute path", dir)
+	}
+
+	// Refuse also "/", otherwise "/" and "" would have the same semantics,
+	// and "" could be unexpectedly shadowed by the "/" entry.
+	// (Note: we do allow "/:someimage", a bit ridiculous but why refuse it?)
```
mtrmac: The parenthesized note is no longer relevant now that identities/scopes do not contain the image name.

oci/oci.go
```diff
@@ -1 +1,91 @@
...
+	}
+	return dir, image
+}
+
+// ValidateScope takes the OCI directory and validates its scope.
+func ValidateScope(scope string, dir string) error {
+	if runtime.GOOS == "windows" {
+		return validateScopeWindows(scope, dir)
+	}
+	return validateScopeNonWindows(scope, dir)
```
mtrmac: Can the two variants share code via `filepath.IsAbs`? (That one also handles UNC paths, but that seems to be compatible enough with our formats.)

oci/oci_test.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+}
+
+func TestValidateScopeWindows(t *testing.T) {
+	tests := []testDataScopeValidation{
+		{`C:\foo:busybox`, `C:\foo`, ""},
+		{`D:\:apline`, `D:\`, ""},
+		{"C:", "C:", "Invalid scope 'C:'. Must be an absolute path"},
+		{"E", "E", "Invalid scope 'E'. Must be an absolute path"},
+		{"", "", "Invalid scope ''. Must be an absolute path"},
```
mtrmac: (`ValidatePolicyConfigurationScope` is never called with the `""` value, so this case is unnecessary. Doesn’t hurt though.)

oci/layout/oci_transport.go
```diff
@@ -111,12 +90,19 @@ func NewReference(dir, image string) (types.ImageReference, error) {
...
+
+	if runtime.GOOS == "windows" {
+		// On Windows we must allow for a ':' as part of the path
+		if strings.Count(resolved, ":") > 1 {
+			return nil, errors.Errorf("Invalid OCI reference %s:%s: path %s contains more than one colon", dir, image, resolved)
+		}
+	} else {
+		if strings.Contains(resolved, ":") {
+			return nil, errors.Errorf("Invalid OCI reference %s:%s: path %s contains a colon", dir, image, resolved)
+		}
```
mtrmac: One more thing, prompted by https://github.com/projectatomic/skopeo/issues/455 :

Nowadays we don’t treat `:` in policy scopes specially — but we _do_ treat `:` specially in the non-canonicalized input path, as output in `StringWithinTransport` and re-parsed in `ParseReference`. So, actually, this check for colons should, if I am not mistaken again, validate `dir`, not `resolved`.

(We originally didn’t validate `dir` because `image` was not allowed to contain colons, so `dir_with_colons:image_without_colons` was not ambiguous. After `image` was allowed colons, this became ambiguous and we should have started refusing such `dir` values at that time.

Conceptually independently, but at at about the same time, `image` was dropped from `PolicyConfiguration*`, and the `resolved` check was no longer necessary.)

oci/layout/oci_transport.go
```diff
@@ -36,40 +37,21 @@ func (t ociTransport) ParseReference(reference string) (types.ImageReference, er
...
-	dir = sep[0]
-
-	if len(sep) == 2 {
-		image := sep[1]
-		if !refRegexp.MatchString(image) {
-			return errors.Errorf("Invalid image %s", image)
-		}
+	dir, image := oci.SplitDirAndImage(scope)
+	if image != nil && !oci.ValidImageName(*image) {
+		return errors.Errorf("Invalid image %s", image)
```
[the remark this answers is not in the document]
mtrmac: … see the other new comment; `NewReference` maybe does not need refuse colons in `resolved` at all; in which case this should do just the check for absolute paths which are not `/`.

oci/oci_test.go
hferentschik: ok

oci/oci.go
hferentschik: I'll remove it

oci/oci.go
hferentschik: Potentially. I guess you are talking about just calling "filepath.IsAbs" w/o a OS specific switch. There would still be 

    if scope == "/" {
        return errors.New(`Invalid scope "/": Use the generic default scope ""`)
    }

to take care of which is non Windows specific (unless you suggest we drop this).

Another disadvantage of using filepath.IsAbs is that one cannot easily write tests for Windows paths. In the current code I can test _validateScopeWindows_ even w/o running the test suite on Windows. Once filepath.IsAbs is used that is not the case anymore. 

I am undecided. I see pro and cons for both and tempted to prefer the current code to start with.

oci/oci.go
```diff
@@ -1 +1,91 @@
...
+// ValidImageName returns true is the specified image matches the open-containers image name specs, false otherwise.
+func ValidImageName(image string) bool {
+	return refRegexp.MatchString(image)
+}
+
+// SplitDirAndImage tries to split the provided OCI reference into the OCI directory and image.
+// The image is returned as string pointer, allowing for nil being returned, if no image part could be extraced from the reference.
+// Neither directory nor image parts are validated at this stage.
+func SplitDirAndImage(reference string) (string, *string) {
+	if runtime.GOOS == "windows" {
```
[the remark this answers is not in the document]
hferentschik: >  OTOH a shared implementation ensures our CI is also testing most of the Windows code, so this is non-blocking.

Right, that's the intention. It allows that the core expectations are tests even on the current CI system. Using platform specific files loses this. If there were a Windows-based CI (like AppVeyor) setup, I'd prefer the platform specific files, but w/o I rather have some tests which are exercised all the time.

oci/oci.go
```diff
@@ -1 +1,91 @@
...
+
+// ValidImageName returns true is the specified image matches the open-containers image name specs, false otherwise.
+func ValidImageName(image string) bool {
+	return refRegexp.MatchString(image)
+}
+
+// SplitDirAndImage tries to split the provided OCI reference into the OCI directory and image.
+// The image is returned as string pointer, allowing for nil being returned, if no image part could be extraced from the reference.
+// Neither directory nor image parts are validated at this stage.
+func SplitDirAndImage(reference string) (string, *string) {
```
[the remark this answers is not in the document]
hferentschik: +1

oci/layout/oci_transport.go
```diff
@@ -92,14 +74,11 @@ type ociReference struct {
...
-	dir = sep[0]
-
-	if len(sep) == 2 {
-		image = sep[1]
+	dir, image := oci.SplitDirAndImage(reference)
+	if image == nil {
+		return NewReference(dir, "")
 	}
-	return NewReference(dir, image)
+	return NewReference(dir, *image)
```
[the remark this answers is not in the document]
hferentschik: I know. But as said, I wanted to keep the existing code behaving exactly how it was. And for that I need to differentiate between "" and nil. If you are ok to return "" and that an existing test gets changed, then yes.

oci/archive/oci_transport.go
```diff
@@ -48,52 +49,30 @@ func (t ociArchiveTransport) ParseReference(reference string) (types.ImageRefere
...
-
 // ParseReference converts a string, which should not start with the ImageTransport.Name prefix, into an OCI ImageReference.
 func ParseReference(reference string) (types.ImageReference, error) {
-	var file, image string
-	sep := strings.SplitN(reference, ":", 2)
-	file = sep[0]
-
-	if len(sep) == 2 {
-		image = sep[1]
+	dir, image := oci.SplitDirAndImage(reference)
```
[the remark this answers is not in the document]
hferentschik: ok

oci/archive/oci_transport.go
```diff
@@ -102,12 +81,14 @@ func NewReference(file, image string) (types.ImageReference, error) {
 	if err != nil {
 		return nil, err
 	}
+
 	// This is necessary to prevent directory paths returned by PolicyConfigurationNamespaces
 	// from being ambiguous with values of PolicyConfigurationIdentity.
 	if strings.Contains(resolved, ":") {
 		return nil, errors.Errorf("Invalid OCI reference %s:%s: path %s contains a colon", file, image, resolved)
 	}
```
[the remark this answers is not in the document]
hferentschik: ok

oci/layout/oci_transport.go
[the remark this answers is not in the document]
hferentschik: Ahh, that would make things easier indeed. Needs a bit of refactoring ....

oci/oci.go
mtrmac: Sure, the non-Windows testing is a valid concern. I suppose we don’t need to care about UNC paths at least for now.

oci/oci.go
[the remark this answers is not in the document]
mtrmac: Makes sense.

oci/layout/oci_transport.go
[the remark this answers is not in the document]
mtrmac: > I know. But as said, I wanted to keep the existing code behaving exactly how it was. And for that I need to differentiate between "" and nil.

All `ParseReference` does now is call `SplitDirAndImage`, and replace a `nil` with `""`. The difference is immediately erased.

Or are you talking (with the references to “existing test” about `ValidatePolicyConfigurationScope`)? That one does treat a trailing `:` and no `:` differently — but `ValidatePolicyConfigurationScope` should nowadays not expect a `:image` in the input at all, so that will disappear.

oci/layout/oci_transport.go
hferentschik: ahh, so check is against the unresolved path. ok

hferentschik:
@mtrmac, I pushed an update to the pull request incorporating your feedback. Hopefully, I understood all your comments correctly. Turns out things are getting easier and the code looks more regular which is usually a good sign. WDYT?

oci/archive/oci_transport.go
```diff
@@ -48,51 +48,16 @@ func (t ociArchiveTransport) ParseReference(reference string) (types.ImageRefere
...
-		return errors.New(`Invalid scope "/": Use the generic default scope ""`)
-	}
-	cleaned := filepath.Clean(file)
-	if cleaned != file {
-		return errors.Errorf(`Invalid scope %s: Uses non-canonical path format, perhaps try with path %s`, scope, cleaned)
+	err := internal.ValidateScope(scope)
+	if err != nil {
+		return err
 	}
 	return nil
```
mtrmac: (This could be just `return internal.ValidateScope(scope)`.)

oci/internal/oci_util.go
```diff
@@ -0,0 +1,121 @@
...
+	if len(image) == 0 {
+		return true
+	}
+	return refRegexp.MatchString(image)
+}
+
+// SplitDirAndImage tries to split the provided OCI reference into the OCI directory and image.
+// The image is returned as string pointer, allowing for nil being returned, if no image part could be extraced from the reference.
+// Neither directory nor image parts are validated at this stage.
+func SplitDirAndImage(reference string) (string, string) {
```
mtrmac: Please rename this `dir` to `path`, throughout.

The comment about a string pointer is no longer true.

oci/oci.go
mtrmac: This is still outstanding I think. Or are you fine with `oci:somedir:busybox` just failing?

oci/internal/oci_util.go
```diff
@@ -0,0 +1,121 @@
...
+		}
+	} else {
+		if strings.Contains(path, ":") {
+			return errors.Errorf("Invalid OCI reference: path %s contains a colon", path)
+		}
+	}
+	return nil
+}
+
+// ValidateScope takes the OCI directory and validates its scope.
```
mtrmac: “validates a policy configuration scope for an OCI transport” or something like that; for `oci-archive` this can be just an absolute path to a file.

oci/archive/oci_transport.go
```diff
@@ -102,12 +67,13 @@ func NewReference(file, image string) (types.ImageReference, error) {
 	if err != nil {
 		return nil, err
 	}
-	// This is necessary to prevent directory paths returned by PolicyConfigurationNamespaces
-	// from being ambiguous with values of PolicyConfigurationIdentity.
-	if strings.Contains(resolved, ":") {
-		return nil, errors.Errorf("Invalid OCI reference %s:%s: path %s contains a colon", file, image, resolved)
+
+	err = internal.ValidateOCIPath(file)
+	if err != nil {
```
mtrmac: (This could be `if err := … err != nil {`.  Applies to both transports.)

oci/oci.go
mtrmac: > and, worse, `somedir:busybox` as path `somedir:busybox`, no image, which would succeed but be inconsistent with the POSIX version.

Sorry, that would pass validation but `somedir:busybox` is an invalid path on Windows, so it would fail either way.

So, overall, with this PR relative paths with images just fail; that’s clearly not worse than before this PR, so non-blocking for me.

oci/archive/oci_transport.go
hferentschik: right.

oci/internal/oci_util.go
hferentschik: +1

oci/archive/oci_transport.go
hferentschik: sure. Personally, I don't like this style of the if statement, but fair enough

oci/internal/oci_util.go
hferentschik: +1

hferentschik:
@mtrmac pushed an update. 

> The question of accepting relative paths is still outstanding; but if you prefer merging the PR now without making that possible, that’s fine with me, as long as it is a conscious decision.

I'd like to get this merged, since this allows Minishift to continue.

hferentschik:
@runcom, @mtrmac - what do you think, can we merge this now?

mtrmac:
:+1:

> > The question of accepting relative paths is still outstanding; but if you prefer merging the PR now without making that possible, that’s fine with me, as long as it is a conscious decision.
>
> I'd like to get this merged, since this allows Minishift to continue. 

OK. Thanks!

@runcom PTAL.

hferentschik:
@runcom not to seem impatient, but I have two pull requests on the Minishift side depending on this :-) :

* https://github.com/minishift/minishift/pull/1685
* https://github.com/minishift/minishift/pull/1716

runcom:
lgtm

hferentschik:
thanks :-)

PR #382, opened 2017-11-13
docker: relax status code check on put manifest

### LINKED ISSUE #381: docker destination fails on quay.io
quay.io implements the V2 Docker Registry API, but not yet the V2.2 media types. containers/image is fine copying from it as a source, but when trying to copy to it as a destination, it fails [during the `PutManifest`](https://github.com/containers/image/blob/0136a925854d78a47f785ce009b5fa8ef5f7e810/docker/docker_image_dest.go#L240) because Quay returns a 202:
```
DEBU[0011] Upload of manifest type application/vnd.docker.distribution.manifest.v1+json failed: Error writing manifest: Error uploading manifest to /v2/jonboulle/busybox/manifests/latest: received unexpected HTTP status: 202 ACCEPTED 
```

The [registry API documentation](https://docs.docker.com/registry/spec/api/#pushing-an-image) doesn't explicitly call out whether 202 is supported in this case as it does for when pushing blobs (when it's expected). 

I'm on the fence about whether this is a bug in Quay or not? Docker manages to push to it just fine, so I still suspect this could be handled better in containers/image - whether the specific error handling here or in the fallback

runcom:
Fix #381 

@jonboulle @mtrmac PTAL

### DIFF
--- a/docker/docker_image_dest.go
+++ b/docker/docker_image_dest.go
@@ -236,7 +236,7 @@ func (d *dockerImageDestination) PutManifest(m []byte) error {
 		return err
 	}
 	defer res.Body.Close()
-	if res.StatusCode != http.StatusCreated {
+	if !successStatus(res.StatusCode) {
 		err = errors.Wrapf(client.HandleErrorResponse(res), "Error uploading manifest to %s", path)
 		if isManifestInvalidError(errors.Cause(err)) {
 			err = types.ManifestTypeRejectedError{Err: err}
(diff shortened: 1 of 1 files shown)


mtrmac:
:+1: seems plausible.

For the record, has this been actually tested against quay.io?

runcom:
> For the record, has this been actually tested against quay.io?

I hope @jonboulle can test it before going ahead with this :angel:

docker/docker_image_dest.go
```diff
@@ -236,7 +236,7 @@ func (d *dockerImageDestination) PutManifest(m []byte) error {
 		return err
 	}
 	defer res.Body.Close()
-	if res.StatusCode != http.StatusCreated {
+	if successStatus(res.StatusCode) {
```
mtrmac: Wait, shouldn't this be `!successStatus`?

(I do hope that that will show up in the CI.)

jonboulle:
I tested a more limited change (`if (res.StatusCode != http.StatusCreated) && (res.StatusCode != http.StatusAccepted)`) and it worked just fine, so I'd be happy enough for this to land if y'all are.

mtrmac:
:+1: pending tests then.

@jonboulle thanks!



[Approved with PullApprove](https://pullapprove.com/containers/image/pull-request/382/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=containers/image)

PR #383, opened 2017-11-20
Make registry ping in GetSignatures not fatal

mfojtik:
The detectProperties is used solery to check whether the registry
supports the signature extension. In case we don't have valid cert for
the registry, this check will error out and never give the locally
defined signature stores chance to be used.

This also reduce the log spam in OpenShift where this code is trying to detect properties for secure integrated registry and failing (not fatal, just spams the log).

### DIFF
--- a/docker/docker_image_src.go
+++ b/docker/docker_image_src.go
@@ -15,7 +15,6 @@ import (
 	"github.com/containers/image/manifest"
 	"github.com/containers/image/types"
 	"github.com/docker/distribution/registry/client"
-	"github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
@@ -181,7 +180,10 @@ func (s *dockerImageSource) GetBlob(info types.BlobInfo) (io.ReadCloser, int64, 
 // (e.g. if the source never returns manifest lists).
 func (s *dockerImageSource) GetSignatures(ctx context.Context, instanceDigest *digest.Digest) ([][]byte, error) {
 	if err := s.c.detectProperties(ctx); err != nil {
-		return nil, err
+		// Do not error here to allow the signatureBase switch to run even when the
+		// registry ping failed. In that case the supportsSignatures should be
+		// false.
+		logrus.Debugf("Detecting properties for %s failed: %v", instanceDigest, err)
 	}
 	switch {
 	case s.c.signatureBase != nil:


mfojtik:
@runcom @mtrmac FYI

TomSweeneyRedHat:
@mfojtik you need to sign your commit 'git commit --amend -s'.  Thanks!

mfojtik:
@TomSweeneyRedHat thanks, done :)

rhatdan:
LGTM

mtrmac:
This turns a temporary network connection problem from a reported error into silently returning no data; that seems very undesirable.

Can’t OpenShift just provide the valid certificates? Essentially, if the certificate is not available, the client just does not have access to the registry. Special-casing the OpenShift “only fetch signatures” case can only work so far (it already fails for non-lookaside storage anyway).

docker/docker_image_src.go
```diff
@@ -181,7 +180,10 @@ func (s *dockerImageSource) GetBlob(info types.BlobInfo) (io.ReadCloser, int64,
 // (e.g. if the source never returns manifest lists).
 func (s *dockerImageSource) GetSignatures(ctx context.Context, instanceDigest *digest.Digest) ([][]byte, error) {
 	if err := s.c.detectProperties(ctx); err != nil {
-		return nil, err
+		// Do not error here to allow the signatureBase switch to run even when the
+		// registry ping failed. In that case the supportsSignatures should be
+		// false.
+		logrus.Debugf("Detecting properties for %s failed: %v", instanceDigest, err)
```
bparees: why not check/run s.c.signatureBase first, before doing the ping?  s.c.supportsSignatures is irrelevant if s.c.signatureBase is !nil.

the caller of GetSignatures would still need to handle the ping error then, of course.  but it wouldn't get a ping error in cases where signatureBase was !nil which is the goal here.

mtrmac: Well, I carry a private branch switching the order to check `s.c.supportsSignatures` before `s.c.signatureBase`, thinking about whether it is the right thing to do. That might happen eventually, and would break this workaround.

OpenShift will need to supply the credentials eventually, won’t it?

So, ultimately, how much effort do we want to spend on workarounds and fragile navigating around the unavailability of credentials, instead of just supplying the credentials?

bparees: > Well, I carry a private branch switching the order to check s.c.supportsSignatures before s.c.signatureBase

that's not what i'm suggesting.

The code flow already checks signatureBase first.  And if signatureBase is set, there's no point in having done the ping earlier.  So why don't we just check signatureBase first, before doing the ping?  That way you do not get a failure trying to reach a secure registry, if you have a lookaside mechanism to reach that registry.

> OpenShift will need to supply the credentials eventually, won’t it?

the user may have to, but openshift itself won't need them since it may not need to talk to the registry directly.

mtrmac: > > Well, I carry a private branch switching the order to check s.c.supportsSignatures before s.c.signatureBase
>
> that's not what i'm suggesting.

I didn’t mean that you did.

> The code flow already checks signatureBase first.  And if signatureBase is set, there's no point in having done the ping earlier.

I am considering switching that order (especially for upload the system-wide default upload location in `/var` seems nonsensical when dealing with a registry which natively supports uploading via `X-Registry-Supports-Signatures`). That’s not addressing the same thing as this PR, but this PR would make such a change impossible.

> > OpenShift will need to supply the credentials eventually, won’t it?
>
> the user may have to, but openshift itself won't need them since it may not need to talk to the registry directly.

If ever `X-Registry-Support-Signatures` takes off as the generally preferred signature distribution mechanism (as I hope it does), the signature import controller will have to provide the credentials.

Generally preferred or not, if someone stands up two OpenShift clusters with one importing images from the other, it would also be much easier to use `X-Registry-Supports-Signatures` for transferring the signatures.

---

(FWIW if the certificates are available in `/etc/docker/certs.d`, containers/image should already be using them automatically. So in all of this, either the credentials are not present in `/etc/docker/certs.d`, in which case they already must be explicitly provided by the kubelet and therefore known to OpenShift somehow, or they are present in `/etc/docker/certs.d`, but the import controller for some surprising reason can’t use them (permissions? different `certs.d` on different nodes?).)

bparees: I would always expect explicit local configuration to take precedence over what the registry tells us.  If i as an admin explicitly configure the system to "get signatures from here for this registry", I don't care what the registry tries to tell me about its signatures.

mtrmac: Overall, my point remains these workarounds adding a narrow path where OpenShift can avoid supplying all the logically necessary information, and instead focusing only on making `registry.access.redhat.com` work, are IMHO just bad design. They make the code more complex, are non-obvious to casual readers, and make doing other changes more difficult, and all of that work will become redundant after OpenShift does supply the necessary credentials, which is AFAICT long-term inevitable.

I don’t think this specific priority decision is all that important. Why can’t we just do the clean thing and supply the certificates? (If I’ve been told in the past and I have forgotten, I apologize; please tell me again.)

---

(Specifically WRT the ordering: It just should never happen that the same registry has a different signatures available through the X-R-S-S and look-aside protocols; so the question which one should take precedence _for that single registry_ should ideally never come up.

The issue in there is the global default shipped in https://github.com/projectatomic/skopeo/blob/master/default.yaml , causing new signatures to be written to `/var/lib/atomic`.  That’s convenient for preparing new data to be served via the look-aside protocol (nothing needs to be set up before creating the signatures), but doing that really does not make sense for servers which do support X-R-S-S; it should never happen that both are available simultaneously, and especially that they differ. So, it seems to make sense that `PutSignatures` should prefer `X-R-S-S` to the look-aside configuration in `registries.d/*.yaml`. And then of course `GetSignatures` should have the same preference.)

rhatdan:
Are you guys going to update this PR?

rhatdan:
Since I heard no response back in January can we just close this PR?

bparees:
@mfojtik bump

rhatdan:
Since this PR seems to have gone dead, I will close it.  Please reopen if you feel we need to merge.
@vrothberg Please take a fresh look.

PR #385, opened 2017-11-29
RFC: Prefer X-Registry-Supports-Signatures to lookaside

### LINKED ISSUE #384: Try signature extensions in registry before using the docker sigstore
Now that the `docker://` transport transparently uses the registry extension for signature (in OpenShift), I think it's better to first try to push the signature on the registry before using the sigstore in the `default.yaml` file.
This came out on fedora, we have a `default-docker` for `/var/lib/atomic/sigstore` shipped in the os, but when pushing to the openshift registry which supports signatures storage remotely, I expected to use that instead of the sigstore in the `default-docker` sigstore.

@mtrmac wdyt? I think we have to first 1) check if the registry support signatures storage, 2) push signatures if it supports them 3) fall back to sigstore if it's not an openshift registry. Right now it's very misleading to have signatures on the host when I expected signatures to be pushed to the registry.

```
# This is the default signature write location for docker registries.
default-docker:
  sigstore: file:///var/lib/atomic/sigstore
#  sigstore-staging: file:///var/lib/atomic/sigstore
```
I have to comment the above to push signature to the openshift integrated registry using the `docker://` transport.

mtrmac:
In general, a single server should never have both, so the order mostly doesn't matter.  The one place where the difference is important is the ability of lookside to have a global default; if lookaside is preferred, such a global default makes X-R-S-S impossible to use.

The disadvantage of this is that simply upgrading a registry to a server version which supports X-R-S-S breaks any previously configured lookaside for the server.

Fixes #384 (?)

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -6,7 +6,7 @@ const (
 	// VersionMajor is for an API incompatible changes
 	VersionMajor = 5
 	// VersionMinor is for functionality in a backwards-compatible manner
-	VersionMinor = 6
+	VersionMinor = 7
 	// VersionPatch is for backwards-compatible bug fixes
 	VersionPatch = 1
 


TomSweeneyRedHat:
LGTM

TomSweeneyRedHat:
@runcom PTAL

runcom:
So, this patch effectively prefers the registry extension when supported right? If that's the case,  the only broken people are the one who are already using the lookaside with a pre-sig-extension registry and they'll start to use the atomic registry. It's tricky, we could look into the sigstore to see if they're using the lookaside, and if they are, prefer that... Still an hack but maybe better than totally overriding.

mtrmac:
We could also integrate the two systems, so that lookaside per-repo is prioritized over either X-R-S-or lookaside configured per-host (and having both could be a hard error), which is prioritized over the global default.

That seems clean as far as pure algorithms / correctness goes, but it also feels like more effort, and more importantly more _complexity_ that would have to be documented _and understood by users_, than this corner case would warrant.

runcom:
> We could also integrate the two systems, so that lookaside per-repo is prioritized over either X-R-S-or lookaside configured per-host (and having both could be a hard error), which is prioritized over the global default.
> 
> That seems clean as far as pure algorithms / correctness goes, but it also feels like more effort, and more importantly more complexity that would have to be documented and understood by users, than this corner case would warrant.

so maybe we can just go with this? and document _somewhere_ (TBD) what we did

rhatdan:
This PR is over a year old, @mtrmac What should we do with it?

rhatdan:
@mtrmac @vrothberg @TomSweeneyRedHat What should we do with this PR?  Should we rebase and move forward or should we close it?

mtrmac:
Rebased, but https://github.com/containers/image/pull/385#issuecomment-348349418 is still outstanding (at least to consider, if not to actually implement).

vrothberg:
@mtrmac, can we use the registry whenever it supports storing signatures **_in addition_** to a potentially configured sigstore? This way, we should remain backwards compatible and would support https://bugzilla.redhat.com/show_bug.cgi?id=1724946. We might want to compare signatures when fetching from two sources.

mtrmac:
Interesting… that’s clearly safe enough for reading (which is where we don’t have a look-aside default, though). It’s not _obviously_ the right thing to do on writes, where we do, but it might be good enough.

In particular, the OpenShift implementation of X-R-S-S has a pretty low limit on the number of allowed signatures (used to be as little as 3!), so it could cause unwanted failures.

mtrmac:
(Generally speaking, the current default/preference is mostly just wrong, and we should switch to making the sigstore opt-in for any X-R-S-S registries. Still, the potential for breaking users who rely on the current setup is there.)

sjhx:
I can only off tea and sympathy on the "breaks existing" problem however I can say that we are bringing on new user who need to take special care to reset the registries.d settings, most remove or somment out the supplied default. Progress on prioritising XRSS would help us.

rhatdan:
Where are we on this?

mtrmac:
This was done in #1035.

PR #389, opened 2017-12-12
ostree: add selinux label for each file

giuseppe:
Add a SELinux label to each file imported into OSTree.  The label given
to each file is the same it would have if installed on the host.

This has two advantages:

1) we can use in the container the same policies we use for processes on
the host.

2) we share more files in the OSTree repository with the host since
having the same SELinux label most likely will produce the same
checksum.

### DIFF
--- a/ostree/ostree_dest.go
+++ b/ostree/ostree_dest.go
@@ -14,25 +14,30 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strconv"
+	"syscall"
 	"time"
+	"unsafe"
 
 	"github.com/containers/image/manifest"
 	"github.com/containers/image/types"
 	"github.com/containers/storage/pkg/archive"
 	"github.com/opencontainers/go-digest"
+	selinux "github.com/opencontainers/selinux/go-selinux"
 	"github.com/ostreedev/ostree-go/pkg/otbuiltin"
 	"github.com/pkg/errors"
 	"github.com/vbatts/tar-split/tar/asm"
 	"github.com/vbatts/tar-split/tar/storage"
 )
 
-// #cgo pkg-config: glib-2.0 gobject-2.0 ostree-1
+// #cgo pkg-config: glib-2.0 gobject-2.0 ostree-1 libselinux
 // #include <glib.h>
 // #include <glib-object.h>
 // #include <gio/gio.h>
 // #include <stdlib.h>
 // #include <ostree.h>
 // #include <gio/ginputstream.h>
+// #include <selinux/selinux.h>
+// #include <selinux/label.h>
 import "C"
 
 type blobToImport struct {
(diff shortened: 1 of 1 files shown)


giuseppe:
@rhatdan I've tried with "SELINUX=disabled":

```
[root@rhel-atomic cloud-user]# getenforce 
Disabled
```

and it still works.  Is there any other testing I should do?

ostree/ostree_dest.go
```diff
@@ -348,14 +379,24 @@ func (d *ostreeImageDestination) Commit() error {
...
 	}
 
+	var selinuxHnd *C.struct_selabel_handle
+
+	if os.Getuid() == 0 {
+		selinuxHnd = C.selabel_open(C.SELABEL_CTX_FILE, nil, 0)
+		if selinuxHnd == nil {
+			errors.Errorf("cannot open the SELinux DB")
+		}
+		defer C.selabel_close(selinuxHnd)
```
rhatdan: Probably does not matter, but this should be 
else defer

rhatdan:
I guess the selinuxHnd ends up nil?

ostree/ostree_dest.go
giuseppe: I just realized I forgot the return at line 387.  Going to push a new version with that fixed and the defer in an else branch

giuseppe:
selinuxHnd should be nil only when the user is not root as files cannot be relabed in that case

rhatdan:
What happens if I am on an SELinux disabled system without selinux-policy installed, then the open will fail.

giuseppe:
Travis doesn't like the new else branch:
```
/gopath/src/github.com/containers/image/ostree/ostree_dest.go:388:10: if block ends with a return statement, so drop this else and outdent its block
```
so I reverted that change.

The open doesn't fail for me here with SELinux disabled:

```
# python
Python 2.7.5 (default, May  3 2017, 07:55:04) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-14)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import selinux
>>> selinux.is_selinux_enabled()
0
>>> selinux.selabel_open(selinux.SELABEL_CTX_FILE, None, 0)
<Swig Object of type 'struct selabel_handle *' at 0x7f137cfb97e0>
```

rhatdan:
Remove the file and see what happens?  Does selinux.SELABEL_CTX_FILE exists.  On a disabled system it likely does not.

giuseppe:
yes, if I completely rm `/etc/selinux` then it fails.  I added a check to verify SELinux is enabled before attempting to use it.

rhatdan:
Exellent.

runcom:
LGTM

PR #390, opened 2017-12-13
Change errors structure to handle them gracefully

ripcurld0:
Using `errors.Cause` one can understand what was the cause of the failure. Unfortunately, when the error is returned due to status code different than 203, 200 or 201 `errors.Cause` is a verbose string.

This patch suggests to set the cause of the error to be the staus text.

For example, if failed to upload a layer, instead of having the following:
`Error initiating layer upload to /path/to/upload, status 401`
This error message will be returned:
`Error initiating layer upload to /path/to/upload": Unauthorized`

It's then easier to handle this error by reading `errors.Cause`:
```Go
if errors.Cause(err) == http.StatusText(http.StatusUnauthorized) {
    return errors.New("You are not authorized. Missing credentials?")
}
```

This can help me to fix https://github.com/projectatomic/buildah/issues/254.

### DIFF
--- a/docker/docker_image_dest.go
+++ b/docker/docker_image_dest.go
@@ -131,7 +131,7 @@ func (d *dockerImageDestination) PutBlob(stream io.Reader, inputInfo types.BlobI
 	defer res.Body.Close()
 	if res.StatusCode != http.StatusAccepted {
 		logrus.Debugf("Error initiating layer upload, response %#v", *res)
-		return types.BlobInfo{}, errors.Errorf("Error initiating layer upload to %s, status %d", uploadPath, res.StatusCode)
+		return types.BlobInfo{}, errors.Wrapf(client.HandleErrorResponse(res), "Error initiating layer upload to %s", uploadPath)
 	}
 	uploadLocation, err := res.Location()
 	if err != nil {
@@ -167,7 +167,7 @@ func (d *dockerImageDestination) PutBlob(stream io.Reader, inputInfo types.BlobI
 	defer res.Body.Close()
 	if res.StatusCode != http.StatusCreated {
 		logrus.Debugf("Error uploading layer, response %#v", *res)
-		return types.BlobInfo{}, errors.Errorf("Error uploading layer to %s, status %d", uploadLocation, res.StatusCode)
+		return types.BlobInfo{}, errors.Wrapf(client.HandleErrorResponse(res), "Error uploading layer to %s", uploadLocation)
 	}
 
 	logrus.Debugf("Upload of layer %s complete", computedDigest)
@@ -447,7 +447,7 @@ sigExists:
 				logrus.Debugf("Error body %s", string(body))
 			}
 			logrus.Debugf("Error uploading signature, status %d, %#v", res.StatusCode, res)
-			return errors.Errorf("Error uploading signature to %s, status %d", path, res.StatusCode)
+			return errors.Wrapf(client.HandleErrorResponse(res), "Error uploading signature to %s", path)
 		}
 	}
 


docker/docker_client.go
```diff
@@ -363,7 +363,7 @@ func (c *dockerClient) detectProperties(ctx context.Context) error {
 		defer resp.Body.Close()
 		logrus.Debugf("Ping %s status %d", url, resp.StatusCode)
 		if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusUnauthorized {
-			return errors.Errorf("error pinging repository, response code %d", resp.StatusCode)
+			return errors.Wrapf(errors.New(http.StatusText(resp.StatusCode)), "Error pingin repository")
```
TomSweeneyRedHat: nit "pingin" to "pinging"

runcom:
LGTM

TomSweeneyRedHat:
The change LGTM with the exception of the nit, but I think I'd rather not remove anything from the original message.  I personally like the status codes, but see the value in the text.  I think I'd prefer from your example:

Error initiating layer upload to /path/to/upload, status 401: Unauthorized.

Also you didn't hit up all the occurrences in this project, should we?  @nalind WDYT?

rhatdan:
With the errors wrap won't we end up with the 401 anyways if you print out the error?

rhatdan:
LGTM with the pingin fix.

ripcurld0:
> With the errors wrap won't we end up with the 401 anyways if you print out the error?

The `http.StatusText` function takes the `res.StatusCode` and returns a text for the HTTP status code.

Here are the texts for each status-code from [`golang` source code](https://github.com/golang/go/blob/94d7c884c38561dba467feef3acf6ada50713e59/src/net/http/status.go#L77-L139)

@TomSweeneyRedHat thank you for noticing! Working on it.

@runcom thanks for the review.

nalind:
I like this.

rhatdan:
@mtrmac PTAL and get this merged.

mtrmac:
> It's then easier to handle this error by reading `errors.Cause`:
> ```Go
> if errors.Cause(err) == http.StatusText(http.StatusUnauthorized) {
>     return errors.New("You are not authorized. Missing credentials?")
> }
> ```

This still relies on a string comparison, which is fragile (it is non-obvious that the string is an API commitment), and invisible to tools.  I’d much prefer using a specific `error` value, or type, for this.

In fact, focusing on the 401 status which seems to be the core concern of https://github.com/projectatomic/buildah/issues/254 , there already are two plausible options:

1. Many of the clients for the docker/distribution API (but not signature lookaside) can, and probably should, call `github.com/docker/distribution/registry/client.HandleErrorResponse`, to get nicer and more detailed error messages. Hat function can also return well-typed `github.com/docker/distribution/registry/api/errcode.ErrorCodeUnauthorized` and `github.com/docker/distribution/registry/client.UnexpectedHTTPResponseError`.
2. We also have `c/i/docker.ErrUnauthorizedForCredentials`, which is returned if the server is using `Authorization: Bearer`.  It seems very reasonable to return the same value for `Authorization: Basic` as well.

In the worst case, a new `type c/i/docker.UnexpectedHTTPStatus int` type implementing `error` would still be cleaner than the string comparison proposed above.

TomSweeneyRedHat:
@ripcurld0 this needs to be rebased.

ripcurld0:
@mtrmac Thanks a lot for the review.
I decided to go with `HandleErrorResponse` as you suggested. Do you mind re-review it again?

By the way, we're not vendoring this package, and we're relying on other people to do that. Should I leave it as it is or we need to vendor this package?

@TomSweeneyRedHat I have rebases the patch. Thank you so much for noticing.

ripcurld0:
ping @mtrmac

mtrmac:
:+1: Thanks! Please rebase once more.

@runcom you might want to take another look.

runcom:
LGTM

ripcurld0:
Thank you so much.

mtrmac:
> By the way, we're not vendoring this package, and we're relying on other people to do that. Should I leave it as it is or we need to vendor this package?

FWIW this needs to stay as it is because c/image is a library package; types in vendored packages are not visible to users. See e.g. https://github.com/opencontainers/image-spec/pull/528#issuecomment-275621109 , and this would equally apply to the error types returned by `HandleErrorResponse`.

rhatdan:
@ripcurld0 Could you vendor this into buildah?

ripcurld0:
@rhatdan consider this done 😉

PR #391, opened 2017-12-14
Revert to containers/storage master

nalind:
We no longer need a private branch of containers/storage, so update vendor.conf to use master.


runcom:
:+1:

PR #392, opened 2017-12-14
ostree, src: support copy of compressed layers

giuseppe:
add implementation for LayerInfosForCopy so we can provide a modified manifest to be used for `copy`

### DIFF
--- a/ostree/ostree_dest.go
+++ b/ostree/ostree_dest.go
@@ -230,33 +230,36 @@ func (d *ostreeImageDestination) ostreeCommit(repo *otbuiltin.Repo, branch strin
 	return err
 }
 
-func generateTarSplitMetadata(output *bytes.Buffer, file string) error {
+func generateTarSplitMetadata(output *bytes.Buffer, file string) (digest.Digest, int64, error) {
 	mfz := gzip.NewWriter(output)
 	defer mfz.Close()
 	metaPacker := storage.NewJSONPacker(mfz)
 
 	stream, err := os.OpenFile(file, os.O_RDONLY, 0)
 	if err != nil {
-		return err
+		return "", -1, err
 	}
 	defer stream.Close()
 
 	gzReader, err := archive.DecompressStream(stream)
 	if err != nil {
-		return err
+		return "", -1, err
 	}
 	defer gzReader.Close()
 
 	its, err := asm.NewInputTarStream(gzReader, metaPacker, nil)
 	if err != nil {
-		return err
+		return "", -1, err
 	}
 
-	_, err = io.Copy(ioutil.Discard, its)
+	digester := digest.Canonical.Digester()
+
+	written, err := io.Copy(digester.Hash(), its)
 	if err != nil {
-		return err
+		return "", -1, err
 	}
-	return nil
+
+	return digester.Digest(), written, nil
 }
 
 func (d *ostreeImageDestination) importBlob(selinuxHnd *C.struct_selabel_handle, repo *otbuiltin.Repo, blob *blobToImport) error {
@@ -271,7 +274,8 @@ func (d *ostreeImageDestination) importBlob(selinuxHnd *C.struct_selabel_handle,
 	}()
 
 	var tarSplitOutput bytes.Buffer
-	if err := generateTarSplitMetadata(&tarSplitOutput, blob.BlobPath); err != nil {
+	uncompressedDigest, uncompressedSize, err := generateTarSplitMetadata(&tarSplitOutput, blob.BlobPath)
+	if err != nil {
 		return err
 	}
 
@@ -293,6 +297,8 @@ func (d *ostreeImageDestination) importBlob(selinuxHnd *C.struct_selabel_handle,
 		}
 	}
 	return d.ostreeCommit(repo, ostreeBranch, destinationPath, []string{fmt.Sprintf("docker.size=%d", blob.Size),
+		fmt.Sprintf("docker.uncompressed_size=%d", uncompressedSize),
+		fmt.Sprintf("docker.uncompressed_digest=%s", uncompressedDigest.String()),
 		fmt.Sprintf("tarsplit.output=%s", base64.StdEncoding.EncodeToString(tarSplitOutput.Bytes()))})
 
 }
--- a/ostree/ostree_dest.go
+++ b/ostree/ostree_dest.go
@@ -321,7 +321,17 @@ func (d *ostreeImageDestination) HasBlob(info types.BlobInfo) (bool, int64, erro
 	}
 	branch := fmt.Sprintf("ociimage/%s", info.Digest.Hex())
 
-	found, data, err := readMetadata(d.repo, branch, "docker.size")
+	found, data, err := readMetadata(d.repo, branch, "docker.uncompressed_digest")
+	if err != nil || !found {
+		return found, -1, err
+	}
+
+	found, data, err = readMetadata(d.repo, branch, "docker.uncompressed_size")
+	if err != nil || !found {
+		return found, -1, err
+	}
+
+	found, data, err = readMetadata(d.repo, branch, "docker.size")
 	if err != nil || !found {
 		return found, -1, err
 	}
--- a/ostree/ostree_src.go
+++ b/ostree/ostree_src.go
@@ -37,11 +37,13 @@ type ostreeImageSource struct {
 	ref    ostreeReference
 	tmpDir string
 	repo   *C.struct_OstreeRepo
+	// get the compressed layer by its uncompressed checksum
+	compressed map[digest.Digest]digest.Digest
 }
 
 // newImageSource returns an ImageSource for reading from an existing directory.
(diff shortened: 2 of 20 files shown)


runcom:
Looks ok, @mtrmac wdyt?

giuseppe:
ping

giuseppe:
cc @nalind

ostree/ostree_dest.go
```diff
@@ -226,33 +226,37 @@ func (d *ostreeImageDestination) ostreeCommit(repo *otbuiltin.Repo, branch strin
...
 	if err != nil {
-		return err
+		return "", -1, err
 	}
 
-	_, err = io.Copy(ioutil.Discard, its)
+	digester := digest.Canonical.Digester()
+	tee := io.TeeReader(its, digester.Hash())
+
+	written, err := io.Copy(ioutil.Discard, tee)
```
mtrmac: Isn’t this basically `io.Copy(digester, its)`?

ostree/ostree_src.go
```diff
@@ -348,7 +356,62 @@ func (s *ostreeImageSource) GetSignatures(ctx context.Context, instanceDigest *d
...
 func (s *ostreeImageSource) LayerInfosForCopy() []types.BlobInfo {
-	return nil
+	updatedBlobInfos := []types.BlobInfo{}
+	manifestBlob, manifestType, err := s.GetManifest(nil)
+	if err != nil {
+		return nil
+	}
+
+	var schema manifestSchema
+	if err := json.Unmarshal(manifestBlob, &schema); err != nil {
```
mtrmac: Please use the `containers/image/manifest` package (`manifest.FromBlob`), both for manifest types and for manifest parsing.

This doesn't work as expected because `LayersDescriptors` and `FSLayers` list layers in the opposite directions; besides, the `manifestSchema` franked-type containing a subset of both schema1 and schema2, and undocumented, is pretty surprising.

ostree/ostree_src.go
```diff
@@ -34,14 +36,15 @@ import (
 import "C"
 
 type ostreeImageSource struct {
-	ref    ostreeReference
-	tmpDir string
-	repo   *C.struct_OstreeRepo
+	ref        ostreeReference
+	tmpDir     string
+	repo       *C.struct_OstreeRepo
+	compressed map[digest.Digest]digest.Digest
```
mtrmac: Please at least add a comment here saying what is the key and what is the value of the map. (Ideally the field name would be self-explanatory, but that seems difficult.)

ostree/ostree_src.go
```diff
@@ -255,7 +258,12 @@ func (s *ostreeImageSource) readSingleFile(commit, path string) (io.ReadCloser,
 
 // GetBlob returns a stream for the specified blob, and the blob's size.
 func (s *ostreeImageSource) GetBlob(info types.BlobInfo) (io.ReadCloser, int64, error) {
+
 	blob := info.Digest.Hex()
+	compressedBlob, found := s.compressed[info.Digest]
```
mtrmac: It’s not in general guaranteed that `LayerInfosForCopy` will be called before `GetBlob`; if `GetBlob` depends on this map, it may need to build it on demand.

giuseppe: thanks, I've modified it to check if compressed is initialized, and use `LayerInfosForCopy` when it is `nil`

ostree/ostree_dest.go
giuseppe: correct, thanks for finding it.  Fixed it in a newer version

ostree/ostree_src.go
giuseppe: missed this comment in the last revision.  Going to address it now.

giuseppe: done now

ostree/ostree_src.go
```diff
@@ -348,7 +364,45 @@ func (s *ostreeImageSource) GetSignatures(ctx context.Context, instanceDigest *d
...
+			return nil
+		}
+		uncompressedDigest := digest.Digest(uncompressedDigestStr)
+		blobInfo := types.BlobInfo{
+			Digest:    uncompressedDigest,
+			Size:      uncompressedSize,
+			MediaType: layerBlob.MediaType,
+		}
+		s.compressed[uncompressedDigest] = layerBlob.Digest
+		updatedBlobInfos = append([]types.BlobInfo{blobInfo}, updatedBlobInfos...)
```
mtrmac: `LayerInfosForCopy` should return layers in the order of `man.LayerInfos`, not reversed.

ostree/ostree_src.go
```diff
@@ -348,7 +364,45 @@ func (s *ostreeImageSource) GetSignatures(ctx context.Context, instanceDigest *d
...
 
-// LayerInfosForCopy() returns updated layer info that should be used when reading, in preference to values in the manifest, if specified.
+// LayerInfosForCopy() returns the list of layer blobs that make up the root filesystem of
+// the image, after they've been decompressed.
 func (s *ostreeImageSource) LayerInfosForCopy() []types.BlobInfo {
-	return nil
+	updatedBlobInfos := []types.BlobInfo{}
+	manifestBlob, manifestType, err := s.GetManifest(nil)
+	if err != nil {
+		return nil
```
mtrmac: Not sure about some other cases, but at least in this one it seems clear that `err` shouldn’t be thrown away; can you add an `error` return value to `LayerInfosForCopy`?

giuseppe: this is the signature from the `ImageSource interface.  The only user of this function in `ostree_src` is `GetBlob` and I handle there the `nil` return, which is considered an error.

mtrmac: Right; please change the interface. `c/i/copy/imageCopier.copyLayers` should be able to tell the difference between “no changes needed” and “reading manifest failed” / ”manifest invalid”.

giuseppe: I've pushed a new change to change the interface

ostree/ostree_src.go
```diff
@@ -348,7 +364,45 @@ func (s *ostreeImageSource) GetSignatures(ctx context.Context, instanceDigest *d
...
+
+		found, uncompressedSizeStr, err := readMetadata(s.repo, branch, "docker.uncompressed_size")
+		if err != nil || !found {
+			return nil, nil
+		}
+
+		uncompressedSize, err := strconv.ParseInt(uncompressedSizeStr, 10, 64)
+		if err != nil {
+			return nil, nil
+		}
```
mtrmac: Is it reasonable to ignore the above errors? At least the last one seems potentially worth failing on.

Up to you.

storage/storage_image.go
```diff
@@ -177,18 +177,18 @@ func (s *storageImageSource) GetManifest(instanceDigest *digest.Digest) (manifes
 
 // LayerInfosForCopy() returns the list of layer blobs that make up the root filesystem of
 // the image, after they've been decompressed.
-func (s *storageImageSource) LayerInfosForCopy() []types.BlobInfo {
+func (s *storageImageSource) LayerInfosForCopy() ([]types.BlobInfo, error) {
```
mtrmac: At least some of the errors in here should probably cause `copy.Image` to fail.

Cc: @nalind

giuseppe: I preferred to not touch this part as I am not really familiar with it.  I can change it if it is required for this PR

ostree/ostree_src.go
giuseppe: thanks, I've changed that to return an error.  Also, I have added a patch that will trigger the layer to be pulled if this metadata is missing.

mtrmac:
Thanks!

LGTM, will wait a bit to give @nalind time to comment on `storageImageSource.LayerInfosForCopy`

storage/storage_image.go
nalind: Generally, if `storageImageSource.LayerInfosForCopy()` is returning `nil`, the copy attempt is going to fail later on when the digests don't match, since I think the majority of images use compressed layers.  So yeah, pretty much anywhere we return `nil` from there, we should start returning the error value that caused us to do so.

mtrmac:
:+1: for the record; let’s at least track `storageImageSource.LayerInfosForCopy` in an issue.

rhatdan:
@runcom PTAL

storage/storage_image.go
giuseppe: I've added a patch that changes `LayerInfosForCopy` in `storage/storage_image.go` to propagate the errors.

runcom:
> @runcom PTAL

I'm waiting on https://github.com/containers/image/pull/392#discussion_r163364190

giuseppe:
I took care of the last comment I got on this PR

rhatdan:
@runcom PTAL

giuseppe:
rebased ⬆️

giuseppe:
rebased again.  Could we please move this forward?

mtrmac:
@runcom ping

PR #393, opened 2017-12-14
storage: log image-not-resolving at debug level, not error

nalind:
Log the message about a given reference not being resolvable to an image ID at the debug level rather than as an error, and wrap the error that we return in case the caller wants to relay the reference's value back as part of the error.

### DIFF
--- a/storage/storage_reference.go
+++ b/storage/storage_reference.go
@@ -71,8 +71,8 @@ func (s *storageReference) resolveImage() (*storage.Image, error) {
 		}
 	}
 	if s.id == "" {
-		logrus.Errorf("reference %q does not resolve to an image ID", s.StringWithinTransport())
-		return nil, ErrNoSuchImage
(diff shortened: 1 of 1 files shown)


runcom:
Looks fine even if this could break someone/somewhere :)

runcom:
:+1:

runcom:
@nalind do we need this in CRI-O _right now_?

nalind:
I don't think so.  @rhatdan was thinking it was an eyesore, but it isn't breaking anything in its current state.

rhatdan:
@runcom No need for this in CRI-O other then putting messages in syslog. The problem is when you run kpod or buildah these messages flood the screen.

There is another one a few lines below, that I had in My PR, but I am not sure that is common.  I am fine with this one.

LGTM

rhatdan:
@runcom can you get this in, or do we need to wait for @mtrmac for approval?

PR #394, opened 2017-12-14
Drop missing images logrus Errors to Warning

rhatdan:
These messages get blasted to the screen in kpod run and
probably on buildah, even though they are expected.  If the
image is not local then I will pull it, but don't blast an
error message at the screen.

### DIFF
--- a/storage/storage_reference.go
+++ b/storage/storage_reference.go
@@ -71,7 +71,7 @@ func (s *storageReference) resolveImage() (*storage.Image, error) {
 		}
 	}
 	if s.id == "" {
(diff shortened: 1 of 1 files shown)


runcom:
I think @nalind made a similar PR here #393

rhatdan:
@nalind @runcom @mtrmac PTAL
This does not look good.

kpod run -q -ti fedora ls -l /dev
ERRO[0000] reference "[overlay@/var/lib/containers/storage+/var/run/containers/storage]docker.io/library/fedora:latest" does not resolve to an image ID

rhatdan:
Will cover this in https://github.com/containers/image/pull/393

PR #380, opened 2017-11-10
image: implement Docker v1 -> OCI conversion

jonboulle:
Since we already know how to convert from Docker V1 to V2 and from
Docker V2 to OCI, we simply do a transitive manifest conversion.

Signed-off-by: Jonathan Boulle <jonathanboulle@gmail.com>

PR #395, opened 2017-12-19
image: implement OCI -> Docker v1 conversion

jonboulle:
This is the inverse of #380 and unfortunately still necessary in order
to be able to use skopeo to push OCI images to a repository like
quay.io.

As in that issue, we simply do a transitive manifest conversion via
the V2 format, which should be lossless.

Signed-off-by: Jonathan Boulle <jonathanboulle@gmail.com>

### DIFF
--- a/image/oci.go
+++ b/image/oci.go
@@ -149,6 +149,16 @@ func (m *manifestOCI1) UpdatedImage(options types.ManifestUpdateOptions) (types.
 
 	switch options.ManifestMIMEType {
 	case "": // No conversion, OK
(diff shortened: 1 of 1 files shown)


runcom:
LGTM

rhatdan:
LGTM
@mtrmac PTAL

mtrmac:
:+1: Thanks!

(As in #380, it would be nice to have at least a smoke test to ensure this keeps working, the manifest conversions are more or less the trickiest code in the repo. Still, non-blocking, better to have an untested conversion than no conversion.)

PR #403, opened 2018-01-18
Update Ubuntu from Zesty to Atful

mtrmac:
Primarily because Zesty is EOL and the repositories are no longer available.

The Ubuntu EOL manifests in failures like
```
W: The repository 'http://security.ubuntu.com/ubuntu zesty-security Release' does not have a Release file.
W: The repository 'http://archive.ubuntu.com/ubuntu zesty Release' does not have a Release file.
W: The repository 'http://archive.ubuntu.com/ubuntu zesty-updates Release' does not have a Release file.
W: The repository 'http://archive.ubuntu.com/ubuntu zesty-backports Release' does not have a Release file.
E: Failed to fetch http://security.ubuntu.com/ubuntu/dists/zesty-security/universe/source/Sources  404  Not Found [IP: 91.189.88.152 80]
E: Failed to fetch http://archive.ubuntu.com/ubuntu/dists/zesty/universe/source/Sources  404  Not Found [IP: 91.189.88.161 80]
E: Failed to fetch http://archive.ubuntu.com/ubuntu/dists/zesty-updates/universe/source/Sources  404  Not Found [IP: 91.189.88.161 80]
E: Failed to fetch http://archive.ubuntu.com/ubuntu/dists/zesty-backports/main/binary-amd64/Packages  404  Not Found [IP: 91.189.88.161 80]
E: Some index files failed to download. They have been ignored, or old ones used instead.
The command '/bin/sh -c apt-get -qq update &&     apt-get install -y sudo docker.io git make btrfs-tools libdevmapper-dev libgpgme-dev libostree-dev' returned a non-zero code: 100
```

The Go update is, strictly speaking, unrelated, but we don’t have a process to keep it updated and this is as good an opportunity as any.

PR #404, opened 2018-01-18
Golang update

mtrmac:
This
- Updates Golang to 1.9.2
- ~~Takes advantage of the 1.9 semantics of `./...` to simplify `Makefile` a bit.~~

Both are RFCs:
- We could instead update to 1.8.5 from the current 1.8.3.  Per #403, F26 has 1.8 (with about 7 months til EOL), F27 and EPEL have 1.9.
- ~~The latter commit will break, or at least make life difficult for, 1.8 users. Without it, we would be testing on 1.9 but still probably making it possible to use the repo with 1.8.  And honestly the benefit of that commit is extremely weak.~~

~~Do not merge as is, it includes https://github.com/containers/image/pull/403 and should be rebased on top of it.~~

### DIFF
--- a/.travis.Dockerfile
+++ b/.travis.Dockerfile
@@ -3,8 +3,8 @@ FROM ubuntu:artful
 RUN apt-get -qq update && \
     apt-get install -y sudo docker.io git make btrfs-tools libdevmapper-dev libgpgme-dev libostree-dev
 
-ADD https://storage.googleapis.com/golang/go1.8.3.linux-amd64.tar.gz /tmp
+ADD https://storage.googleapis.com/golang/go1.9.2.linux-amd64.tar.gz /tmp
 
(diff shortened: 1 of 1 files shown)


runcom:
> The latter commit will break, or at least make life difficult for, 1.8 users. Without it, we would be testing on 1.9 but still probably making it possible to use the repo with 1.8. And honestly the benefit of that commit is extremely weak.

I'd drop that commit for now indeed

mtrmac:
Dropped the `./...` change.

runcom:
> We could instead update to 1.8.5 from the current 1.8.3. Per #403, F26 has 1.8 (with about 7 months til EOL), F27 and EPEL have 1.9.

not sure about this, we don't really have a matrix for such testing, I'd just move to latest upstream (1.9.x) as you did here

runcom:
LGTM

mtrmac:
👍 OK, let’s go with 1.9.

PR #405, opened 2018-01-18
Make token parsing compatible with Docker's Token Authentication Spec

yulianedyalkova:
According to the docker registry spec when you request a bearer token from a registry, the response JSON could contain an access_token or token key. Currently containers/image only supports the second option and is therefore incompatible with Azure container registries.

Signed-off-by: Tom Godkin tgodkin@pivotal.io
Signed-off-by: Ed King eking@pivotal.io
Signed-off-by: Will Martin wmartin@pivotal.io

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -66,9 +66,10 @@ type extensionSignatureList struct {
 }
 
 type bearerToken struct {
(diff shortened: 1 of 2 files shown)


yulianedyalkova:
We accidentally closed

PR #406, opened 2018-01-18
Add functionality to search registries

umohnani8:
podman search searches a registry for a matching image
this adds the functionality to support that
some registries respond to the v2 endpoint while others
only respond to the v1 endpoint.
This checks both endpoints for a result, and if none is given
the user is informed.

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -8,7 +8,9 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"time"
 
@@ -24,8 +26,9 @@ import (
 )
 
 const (
-	dockerHostname = "docker.io"
-	dockerRegistry = "registry-1.docker.io"
+	dockerHostname   = "docker.io"
+	dockerV1Hostname = "index.docker.io"
+	dockerRegistry   = "registry-1.docker.io"
 
 	systemPerHostCertDirPath = "/etc/docker/certs.d"
 
@@ -219,6 +222,100 @@ func CheckAuth(ctx context.Context, sCtx *types.SystemContext, username, passwor
 	default:
 		return errors.Errorf("error occured with status code %q", resp.StatusCode)
 	}
+}
+
+// SearchResult holds the information of each matching image
+// It matches the output returned by the v1 endpoint
+type SearchResult struct {
+	Name        string `json:"name"`
+	Description string `json:"description"`
+	// StarCount states the number of stars the image has
+	StarCount int  `json:"star_count"`
+	IsTrusted bool `json:"is_trusted"`
+	// IsAutomated states whether the image is an automated build
+	IsAutomated bool `json:"is_automated"`
+	// IsOfficial states whether the image is an official build
+	IsOfficial bool `json:"is_official"`
+}
+
+// SearchRegistry queries a registry for images that contain "image" in their name
+// The limit is the max number of results desired
+// Note: The limit value doesn't work with all registries
+// for example registry.access.redhat.com returns all the results without limiting it to the limit value
+func SearchRegistry(ctx context.Context, sCtx *types.SystemContext, registry, image string, limit int) ([]SearchResult, error) {
+	type V2Results struct {
+		// Repositories holds the results returned by the /v2/_catalog endpoint
+		Repositories []string `json:"repositories"`
+	}
+	type V1Results struct {
+		// Results holds the results returned by the /v1/search endpoint
+		Results []SearchResult `json:"results"`
+	}
+	v2Res := &V2Results{}
+	v1Res := &V1Results{}
+
+	// The /v2/_catalog endpoint has been disabled for docker.io therefore the call made to that endpoint will fail
+	// So using the v1 hostname for docker.io for simplicity of implementation and the fact that it returns search results
+	if registry == dockerHostname {
+		registry = dockerV1Hostname
+	}
+
+	client, err := newDockerClientWithDetails(sCtx, registry, "", "", "", nil, "")
+	if err != nil {
+		return nil, errors.Wrapf(err, "error creating new docker client")
+	}
+
+	logrus.Debugf("trying to talk to v2 search endpoint\n")
+	resp, err := client.makeRequest(ctx, "GET", "/v2/_catalog", nil, nil)
+	if err != nil {
+		logrus.Debugf("error getting search results from v2 endpoint %q: %v", registry, err)
+	} else {
+		defer resp.Body.Close()
+		if resp.StatusCode != http.StatusOK {
+			logrus.Debugf("error getting search results from v2 endpoint %q, status code %q", registry, resp.StatusCode)
+		} else {
+			if err := json.NewDecoder(resp.Body).Decode(v2Res); err != nil {
+				return nil, err
+			}
+			searchRes := []SearchResult{}
+			for _, repo := range v2Res.Repositories {
+				if strings.Contains(repo, image) {
+					res := SearchResult{
+						Name: repo,
+					}
+					searchRes = append(searchRes, res)
+				}
+			}
+			return searchRes, nil
+		}
+	}
+
+	// set up the query values for the v1 endpoint
+	u := url.URL{
+		Path: "/v1/search",
+	}
+	q := u.Query()
+	q.Set("q", image)
+	q.Set("n", strconv.Itoa(limit))
+	u.RawQuery = q.Encode()
+
+	logrus.Debugf("trying to talk to v1 search endpoint\n")
+	resp, err = client.makeRequest(ctx, "GET", u.String(), nil, nil)
+	if err != nil {
+		logrus.Debugf("error getting search results from v1 endpoint %q: %v", registry, err)
+	} else {
+		defer resp.Body.Close()
+		if resp.StatusCode != http.StatusOK {
+			logrus.Debugf("error getting search results from v1 endpoint %q, status code %q", registry, resp.StatusCode)
+		} else {
+			if err := json.NewDecoder(resp.Body).Decode(v1Res); err != nil {
+				return nil, err
+			}
+			return v1Res.Results, nil
+		}
+	}
+
+	return nil, errors.Wrapf(err, "couldn't search registry %q", registry)
 }
 
 // makeRequest creates and executes a http.Request with the specified parameters, adding authentication and TLS options for the Docker client.


umohnani8:
@mtrmac @runcom PTAL

docker/docker_client.go
```diff
@@ -357,7 +440,7 @@ func (c *dockerClient) getBearerToken(ctx context.Context, realm, service, scope
 
 // detectProperties detects various properties of the registry.
 // See the dockerClient documentation for members which are affected by this.
-func (c *dockerClient) detectProperties(ctx context.Context) error {
+func (c *dockerClient) detectProperties(ctx context.Context, isSearch bool) error {
```
mtrmac: Please use a name for the parameter which describes what the parameter means/does, not what it is used for; `allowV1` or something like that.

docker/docker_client.go
```diff
@@ -221,10 +221,93 @@ func CheckAuth(ctx context.Context, sCtx *types.SystemContext, username, passwor
...
+		return nil, nil, errors.Wrapf(err, "error creating new docker client")
+	}
+	resp, err := newLoginClient.makeRequest(ctx, "GET", path, nil, nil, true)
+	if err != nil {
+		return nil, nil, err
+	}
+	if resp.StatusCode != http.StatusOK {
+		return nil, nil, errors.Errorf("error getting response from endpoint")
+	}
+	if isV2 {
```
mtrmac: The way this is split, with the URL path hard-coded in the caller and the data format hard-coded in the callee, is weird.  Maybe instead the front part of `getRegistryData` should become a “GET + check for StatusOK” helper?

Why do the V1 and V2 requests use a separate `dockerClient` anyway?

docker/docker_client.go
```diff
@@ -221,10 +221,93 @@ func CheckAuth(ctx context.Context, sCtx *types.SystemContext, username, passwor
...
+		searchResults[reg] = results
+	}
+	return searchResults, nil
+}
+
+// getREgistryData talks to either the v2 or v1 endpoint to get the results of the search query
+func getRegistryData(ctx context.Context, sCtx *types.SystemContext, registry, path string, isV2 bool) (*V2Data, *V1Data, error) {
+	v2Data := &V2Data{}
+	v1Data := &V1Data{}
+	newLoginClient, err := newDockerClientWithDetails(sCtx, registry, "", "", "", nil, "")
```
mtrmac: “login” seems incorrect. Maybe just `client`, or even `c`?

docker/docker_client.go
```diff
@@ -407,7 +490,8 @@ func (c *dockerClient) detectProperties(ctx context.Context) error {
 		if !isV1 && c.ctx != nil && c.ctx.DockerInsecureSkipTLSVerify {
 			isV1 = pingV1("http")
 		}
-		if isV1 {
+		// can talk to v1 registry if doing a search
+		if isV1 && !isSearch {
```
mtrmac: Maybe I’m being stupid but I can’t see how this works; here `err != nil` necessarily because we got into this branch, so this only changes what is the reported error but not that the method fails, i.e. `newLoginClient.makeRequest` should AFAICT fail for V1.

docker/docker_client.go
TomSweeneyRedHat: +1 for client

umohnani8: fixed

docker/docker_client.go
umohnani8: restructured it quite a bit. Should be fixed now.

docker/docker_client.go
umohnani8: fixed

docker/docker_client.go
```diff
@@ -221,10 +221,87 @@ func CheckAuth(ctx context.Context, sCtx *types.SystemContext, username, passwor
...
+	for _, reg := range registries {
+		url := reg
+		if strings.Contains(reg, "docker") {
+			url = "index.docker.io"
+		}
+
+		logrus.Debugf("pinging v2 endpoint\n")
+		data, err := getRegistryData(ctx, sCtx, url, image, limit)
+		if err != nil {
+			fmt.Printf("couldn't search registry %q: %v", reg, err)
```
mtrmac: Please use `logrus`.

docker/docker_client.go
```diff
@@ -221,10 +221,87 @@ func CheckAuth(ctx context.Context, sCtx *types.SystemContext, username, passwor
...
+				results = append(results, res)
+			}
+		}
+		searchResults[reg] = results
+	}
+	return searchResults, nil
+}
+
+// getREgistryData talks to either the v2 or v1 endpoint to get the results of the search query
+func getRegistryData(ctx context.Context, sCtx *types.SystemContext, registry, image, limit string) (*Data, error) {
```
mtrmac: (If this function continues to exist, the name should be somehow related to searching; `getRegistryData` could be pretty much anything.)

docker/docker_client.go
```diff
@@ -221,10 +221,87 @@ func CheckAuth(ctx context.Context, sCtx *types.SystemContext, username, passwor
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "error creating new docker client")
+	}
+	logrus.Debugf("pinging v2 endpoint\n")
+	resp, err := client.makeRequest(ctx, "GET", "/v2/_catalog", nil, nil, true)
+	if err != nil || resp.StatusCode != http.StatusOK {
+		logrus.Debugf("pinging v1 endpoint\n")
+		resp, err = client.makeRequest(ctx, "GET", "/v1/search?q="+image+"&n="+limit, nil, nil, true)
+		if err != nil || resp.StatusCode != http.StatusOK {
+			return nil, errors.Errorf("error getting response from endpoint")
```
mtrmac: This should preserve the available data about the failure, i.e. either `err` or `res.StatusCode`.

docker/docker_client.go
mtrmac: … Ah, I see what’s going on: The currently relevant registries like `docker.io` and `registry.access.redhat.com` also/primarily serve content using the docker/distribution `/v2/` API as well, so we _never_ get here.

If we are happy enough to limit ourselves such V2 + `/v1/search` registries, the `allowV1` boolean does not need to exist at all.

---

BTW the https://docs.docker.com/v1.4/reference/api/docker-io_api/ document you found says that only `Authorization: Basic` is supported, making both the ping and maybe all of `setupRequestAuth` seemingly unnecessary. OTOH https://github.com/docker/docker-registry/blob/master/docker_registry/toolkit.pyhttps://github.com/docker/docker-registry/blob/master/docker_registry/toolkit.py#L276L276 does something entirely different, an `X-Signature` header or `Authorization: Token` – and `/v1/search` does not require any authentication anyway.

Similarly `/v1/search` requires authentication on neither `index.docker.io` nor `r.a.rh.com`.

moby/moby seems to support _both_ v2-like and v1-like authentication for the `/v1/search` URL. I guess we’ll eventually find out which ones, if any, matter; that doesn't need to block this PR.

docker/docker_client.go
```diff
@@ -221,10 +221,87 @@ func CheckAuth(ctx context.Context, sCtx *types.SystemContext, username, passwor
...
+		}
+
+		if len(data.Results) > 0 {
+			searchResults[reg] = data.Results
+			continue
+		}
+
+		results := []Results{}
+		for _, repo := range data.Repositories {
+			if strings.Contains(repo, image) {
```
mtrmac: Should this respect `limit`? (I don’t have a strong opinion.)

docker/docker_client.go
```diff
@@ -221,10 +221,87 @@ func CheckAuth(ctx context.Context, sCtx *types.SystemContext, username, passwor
 	}
 }
 
+// Data holds the output of both the v2 and v1 endpoints
+type Data struct {
```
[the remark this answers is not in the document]
umohnani8: Fixed using the first method you suggested.

docker/docker_client.go
```diff
@@ -221,10 +221,87 @@ func CheckAuth(ctx context.Context, sCtx *types.SystemContext, username, passwor
...
+	// Repositories holds the results returned by the /v2/_catalog endpoint
+	Repositories []string `json:"repositories"`
+	// Results holds the results returned by the /v1/search endpoint
+	Results []Results `json:"results"`
+}
+
+// Results holds the information of each matching image returned by the v1 endpoint
+type Results struct {
```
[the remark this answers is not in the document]
umohnani8: fixed

docker/docker_client.go
```diff
@@ -221,10 +221,87 @@ func CheckAuth(ctx context.Context, sCtx *types.SystemContext, username, passwor
...
+	Name        string `json:"name"`
+	Description string `json:"description"`
+	StarCount   int    `json:"star_count"`
+	IsTrusted   bool   `json:"is_trusted"`
+	IsAutomated bool   `json:"is_automated"`
+	IsOfficial  bool   `json:"is_official"`
```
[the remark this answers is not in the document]
umohnani8: Added a bit of description to the fields. fixed

docker/docker_client.go
```diff
@@ -221,10 +221,87 @@ func CheckAuth(ctx context.Context, sCtx *types.SystemContext, username, passwor
...
+}
+
+// SearchRegistries queries a list of registries for a matching image
+func SearchRegistries(ctx context.Context, sCtx *types.SystemContext, registries []string, image, limit string) (map[string][]Results, error) {
+	searchResults := make(map[string][]Results)
+	for _, reg := range registries {
```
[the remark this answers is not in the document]
umohnani8: Fixed, changed to SearchRegistry, where only 1 registry is being searched.

docker/docker_client.go
```diff
@@ -221,10 +221,87 @@ func CheckAuth(ctx context.Context, sCtx *types.SystemContext, username, passwor
...
+		url := reg
+		if strings.Contains(reg, "docker") {
+			url = "index.docker.io"
```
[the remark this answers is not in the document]
umohnani8: Took this out of here and doing it in podman. User has to send the actual registry url for the search to happen. Fixed.

docker/docker_client.go
```diff
@@ -221,10 +221,87 @@ func CheckAuth(ctx context.Context, sCtx *types.SystemContext, username, passwor
...
+		}
+
+		logrus.Debugf("pinging v2 endpoint\n")
```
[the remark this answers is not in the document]
umohnani8: fixed

docker/docker_client.go
umohnani8: Not needed anymore. Fixed

docker/docker_client.go
```diff
@@ -221,10 +221,87 @@ func CheckAuth(ctx context.Context, sCtx *types.SystemContext, username, passwor
...
+		data, err := getRegistryData(ctx, sCtx, url, image, limit)
+		if err != nil {
+			fmt.Printf("couldn't search registry %q: %v", reg, err)
+			continue
```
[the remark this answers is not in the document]
umohnani8: Fixed

docker/docker_client.go
umohnani8: Removed this function. Fixed

docker/docker_client.go
```diff
@@ -221,10 +221,87 @@ func CheckAuth(ctx context.Context, sCtx *types.SystemContext, username, passwor
...
+			}
+		}
+		searchResults[reg] = results
+	}
+	return searchResults, nil
+}
+
+// getREgistryData talks to either the v2 or v1 endpoint to get the results of the search query
+func getRegistryData(ctx context.Context, sCtx *types.SystemContext, registry, image, limit string) (*Data, error) {
+	client, err := newDockerClientWithDetails(sCtx, registry, "", "", "", nil, "")
```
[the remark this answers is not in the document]
umohnani8: So from what I read online, looks like no form of authentication is required to get a response from the search endpoint, that is why I have not incorporated credentials in here. Do you thing I should?

docker/docker_client.go
```diff
@@ -221,10 +221,87 @@ func CheckAuth(ctx context.Context, sCtx *types.SystemContext, username, passwor
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "error creating new docker client")
+	}
+	logrus.Debugf("pinging v2 endpoint\n")
```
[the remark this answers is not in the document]
umohnani8: Fixed

docker/docker_client.go
```diff
@@ -221,10 +221,87 @@ func CheckAuth(ctx context.Context, sCtx *types.SystemContext, username, passwor
...
+	resp, err := client.makeRequest(ctx, "GET", "/v2/_catalog", nil, nil, true)
```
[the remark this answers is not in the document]
umohnani8: Removed allowV1 based on your explanation below. Makes more sense now, not sure why initially it was failing guess it was a flake. Fixed.

docker/docker_client.go
```diff
@@ -221,10 +221,87 @@ func CheckAuth(ctx context.Context, sCtx *types.SystemContext, username, passwor
...
+	if err != nil || resp.StatusCode != http.StatusOK {
+		logrus.Debugf("pinging v1 endpoint\n")
+		resp, err = client.makeRequest(ctx, "GET", "/v1/search?q="+image+"&n="+limit, nil, nil, true)
```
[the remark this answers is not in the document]
umohnani8: Fixed

docker/docker_client.go
umohnani8: Fixed.

docker/docker_client.go
```diff
@@ -221,10 +221,87 @@ func CheckAuth(ctx context.Context, sCtx *types.SystemContext, username, passwor
...
+		if err != nil || resp.StatusCode != http.StatusOK {
+			return nil, errors.Errorf("error getting response from endpoint")
+		}
+	}
+	defer resp.Body.Close()
```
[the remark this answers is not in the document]
umohnani8: Fixed.

docker/docker_client.go
umohnani8: The limit is only for the v1 endpoint. I am changing the output in podman based on the number of results I get. Could change the variable to `v1Limit`, however the limit query doesn't always work. For example it doesn't work for the redhat registry.

docker/docker_client.go
```diff
@@ -221,10 +221,87 @@ func CheckAuth(ctx context.Context, sCtx *types.SystemContext, username, passwor
...
+type Results struct {
+	Name        string `json:"name"`
+	Description string `json:"description"`
+	StarCount   int    `json:"star_count"`
+	IsTrusted   bool   `json:"is_trusted"`
+	IsAutomated bool   `json:"is_automated"`
+	IsOfficial  bool   `json:"is_official"`
+}
+
+// SearchRegistries queries a list of registries for a matching image
```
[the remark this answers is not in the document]
mtrmac: If the _semantics_ of the `limit` field is an integer, it would be cleaner for the caller to pass an integer. Converting an integer value into a string is for `SearchRegistries` to worry about, just like URL-encoding the query and asking over HTTP are things internal to `SearchRegistries` and not exposed to callers.

`s/number of queries/number of results/` perhaps.

docker/docker_client.go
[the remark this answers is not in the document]
mtrmac: How does mapping the hostname in `podman` work with the certificate and username/password lookup?

(I guess it doesn’t strictly speaking matter because we assume the search/catalog endpoints to be completely unauthenticated… still, it seems unclean.)

docker/docker_client.go
mtrmac: AFAICT the error, and non-200 status, is still thrown away. Sure, the error message will have to be ugly if both v1 and v2 fail, but for debugging it would be nice to return something a bit useful.

docker/docker_client.go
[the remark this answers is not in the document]
mtrmac: … OTOH `resp.Body.Close()` needs to be `defer`red only _after_ checking `err`; in that case, `resp` may be `nil`.

docker/docker_client.go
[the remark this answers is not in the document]
mtrmac: If we don’t know of a registry that requires authentication, it makes sense not to bother implementing it (and have it completely untested).

Still, I’d prefer the external API to allow for authentication to be added in the future (see the `index.docker.io` mapping discussion.)

docker/docker_client.go
mtrmac: Fair enough, we can let the caller deal with excess results. (Or we could truncate them in this function, for both v1 and v2; both would be consistent. But throwing away received data seems wasteful.)

docker/docker_client.go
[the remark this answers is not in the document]
umohnani8: So you are saying that `SearchRegistry` should have username and password as parameters and if in future we find out that registries need authentication to search we can use that? I don't think it is something we need in podman yet, but I could definitely add it to the search registry function.

docker/docker_client.go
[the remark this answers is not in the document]
umohnani8: fixed

docker/docker_client.go
umohnani8: fixed. Logging the errors and response code now.

docker/docker_client.go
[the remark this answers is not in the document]
umohnani8: fixed

docker/docker_client.go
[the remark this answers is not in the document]
mtrmac: No, I was thinking that the username/password could in the future come from `config.GetAuthentication`; that already allows the caller to override the username/password (which should be rare) via `types.SystemContext.DockerAuthConfig`.

So, no need to add unimplemented username/password parameters to the API now; but the hypothetical `config.GetAuthentication` lookup would be easier if the `registry` parameter for Docker Hub were `docker.io`, with the `index.docker.io` mapping performed somewhere inside this code, instead of being user-visible.

docker/docker_client.go
mtrmac: Good idea, logging the two results is much clearer than trying to cram the data into an `error`.

docker/docker_client.go
```diff
@@ -221,6 +223,84 @@ func CheckAuth(ctx context.Context, sCtx *types.SystemContext, username, passwor
...
+	logrus.Debugf("trying to talk to v1 search endpoint\n")
+	resp, err = client.makeRequest(ctx, "GET", u.String(), nil, nil)
+	if err == nil || resp.StatusCode == http.StatusOK {
+		if err := json.NewDecoder(resp.Body).Decode(v1Res); err != nil {
+			return nil, err
+		}
+		return v1Res.Results, nil
+	}
+	defer resp.Body.Close()
+	logrus.Errorf("error getting search results from v2 endpoint %q, status code %q: %v", registry, resp.StatusCode, err)
```
mtrmac: `s/v2/v1/`

docker/docker_client.go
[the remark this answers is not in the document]
umohnani8: ah okay, I get it now. Yeah I can do the mapping in here instead.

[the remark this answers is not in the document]
umohnani8: fixed.

docker/docker_client.go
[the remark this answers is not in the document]
mtrmac: I still don’t think it’s correct. Now, with
```go
	resp, err := client.makeRequest(…)
	if err == nil && resp.StatusCode == http.StatusOK {
                …
		return …
	}
	defer resp.Body.Close()
```

1. `resp.Body` is not closed on the `StatusOK` path
2. `defer resp.Body.Close` is still called if `err != nil`, and will crash if `err != nil && resp == nil`.

This needs to be something like
```go
resp, err := client.makeRequest(…)
if err != nil {
    // Failed, log err
} else {
    defer resp.Body.Close()
    if resp.StatusCode != http.StatusOK {
        // Failed, log resp
    } else {
        // OK, process resp
    }
}
```
, however clumsy that looks; which is also why a `makeGETRequest` which hides the error handling might be worthwhile (not sure).

docker/docker_client.go
umohnani8: I changed it from logrus.Errorf to logrus.Debugf because the error was being logged always and I want it to only show if someone enables debugging. Hope that is acceptable.

mtrmac: The principal case which I’m thinking about here is the user mistyping a hostname, or the network being down (guessing with absolutely no evidence that these are the most likely kinds of failures); in that case it would be nice to tell the user something useful about the failure by default.

Still, a `Debugf` is acceptable.

docker/docker_client.go
```diff
@@ -221,6 +223,88 @@ func CheckAuth(ctx context.Context, sCtx *types.SystemContext, username, passwor
...
+	type V1Results struct {
+		// Results holds the results returned by the /v1/search endpoint
+		Results []SearchResult `json:"results"`
+	}
+	v2Res := &V2Results{}
+	v1Res := &V1Results{}
+
+	if registry == "docker.io" {
+		registry = "index.docker.io"
+	}
```
mtrmac: (Non-blocking: Consider using the existing `dockerHostname` constant, and adding a new one for the `index.docker.io` host name.)

mtrmac: (This still does not give the original host name to the `dockerCertDir` lookup `newDockerClientWithDetails`. But it seems we don’t need to authenticate that way for `docker.io`, and reworking the code to do the mapping later seems pointlessly difficult right now, so I guess let’s leave it as it is. The API caller has a simple interface, and that’s the important thing.)

docker/docker_client.go
```diff
@@ -221,6 +223,88 @@ func CheckAuth(ctx context.Context, sCtx *types.SystemContext, username, passwor
...
+	logrus.Debugf("trying to talk to v1 search endpoint\n")
+	resp, err = client.makeRequest(ctx, "GET", u.String(), nil, nil)
+	if err == nil || resp.StatusCode == http.StatusOK {
+		if err := json.NewDecoder(resp.Body).Decode(v1Res); err != nil {
+			return nil, err
+		}
+		return v1Res.Results, nil
+	}
+	defer resp.Body.Close()
+	logrus.Debugf("error getting search results from v2 endpoint %q, status code %q: %v", registry, resp.StatusCode, err)
```
mtrmac: `s/v2/v1/`

docker/docker_client.go
mtrmac: (This also causes us to use the V2 API against the V1 host name. Again, it works, and doing it right is more difficult. Might be worth at least a comment noting that this is not strictly following the API, and that this is for simplicity of the implementation [and not a workaround for some obscure bug which needs to be left unmodified in the future].)

docker/docker_client.go
umohnani8: Right, so using a Debugf to log any errors occurring, but if both the v1 and v2 endpoint fail an error is returned. The user can then enable debugging for further information.

docker/docker_client.go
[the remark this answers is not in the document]
umohnani8: Okay I see what you mean. I am creating a makeGetRequest function to handle the error check more gracefully.

docker/docker_client.go
umohnani8: fixed

docker/docker_client.go
umohnani8: fixed

umohnani8: Added comment.

docker/docker_client.go
umohnani8: fixed.

docker/docker_client.go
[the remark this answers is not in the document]
umohnani8: created the makeGETRequest function.

docker/docker_client.go
```diff
@@ -221,6 +224,102 @@ func CheckAuth(ctx context.Context, sCtx *types.SystemContext, username, passwor
...
+}
+
+// makeGETRequest makes a get request and checks for the error and status code
+// if the status code is non 200 an error is returned
+func (c *dockerClient) makeGETRequest(ctx context.Context, path string, headers map[string][]string, stream io.Reader) (*http.Response, error) {
+	resp, err := c.makeRequest(ctx, "GET", path, headers, stream)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error making request")
+	}
+	defer resp.Body.Close()
```
mtrmac: Does this work? The `defer` should execute when `makeGETRequest` returns, preventing further reads from `resp.Body`.

docker/docker_client.go
```diff
@@ -221,6 +224,102 @@ func CheckAuth(ctx context.Context, sCtx *types.SystemContext, username, passwor
...
+	if resp.StatusCode != http.StatusOK {
+		return nil, errors.Errorf("not OK, status code %q", resp.StatusCode)
```
mtrmac: (Would this be better with `client.HandleErrorResponse`? Maybe it doesn’t matter, but see #390 / #409 for other users who care.)

docker/docker_client.go
umohnani8: yup it didn't work, my bad. I decided to do it the way you gave an example of with a bunch of if-else statements. It looks cluttered but that seems the best and easiest way of doing it for now.

docker/docker_client.go
umohnani8: I don't think it would make a difference.

mtrmac:
:+1:

Thanks!

docker/docker_client.go
```diff
@@ -221,6 +224,100 @@ func CheckAuth(ctx context.Context, sCtx *types.SystemContext, username, passwor
...
+	u.RawQuery = q.Encode()
+
+	logrus.Debugf("trying to talk to v1 search endpoint\n")
+	resp, err = client.makeRequest(ctx, "GET", u.String(), nil, nil)
+	if err != nil {
+		logrus.Debugf("error getting search results from v1 endpoint %q: %v", registry, err)
+	} else {
+		defer resp.Body.Close()
+		if resp.StatusCode != http.StatusOK {
+			logrus.Debugf("error getting search results from v1 endpoint %q, status code %q", registry, resp.StatusCode)
```
TomSweeneyRedHat: I'm going back/forth, should this be an Error rather than Debugf?  Just wondering what the end user would see if this is hit and if they'd know what was up.

TomSweeneyRedHat: FWIW, I know they get the error below, just wondering if it would help to get the Status code there too....  Like I said, I'm on the fence.

umohnani8: I have it as a debug instead because the error was being printed out everytime this failed. And since it is possible for us to fail with the v2 endpoint and then move on to try the v1 endpoint, we don't want to fill the output with a bunch of error statements.
If both the v2 and v1 endpoints fail, an error is returned and then the user can enable debugging for further information on what is actually going wrong.

umohnani8: I think the status code will be gibberish if the error is not nil, not sure though.

TomSweeneyRedHat: Ah, the age old too little vs too much info problem.  Thanks for the 411 and you've tipped me to the debug side.  At some point in time it might be nice to get a blog together showing how you can use --debug to get more info if things are going south.

umohnani8:
@runcom PTAL

runcom:
LGTM

PR #408, opened 2018-01-23
alltransports: ostree is linux only

vbatts:
Signed-off-by: Vincent Batts <vbatts@hashbangbash.com>

### DIFF
--- a/transports/alltransports/ostree.go
+++ b/transports/alltransports/ostree.go
@@ -1,4 +1,4 @@
-// +build !containers_image_ostree_stub
+// +build !containers_image_ostree_stub,linux
 
 package alltransports
 
--- a/transports/alltransports/ostree_stub.go
+++ b/transports/alltransports/ostree_stub.go
@@ -1,4 +1,4 @@
-// +build containers_image_ostree_stub
+// +build containers_image_ostree_stub !linux
 
 package alltransports
 


mtrmac:
Please update `ostree_stub.go` in the same directory as well.

vbatts:
updated PTAL

vbatts:
let me just double check that a comma here means AND

vbatts:
yea https://golang.org/pkg/go/build/#hdr-Build_Constraints carry on

mtrmac:
The idea is that the stub is compiled in exactly when the true registration isn’t, so shouldn’t the negation of (`!containers_image_ostree_stub` and `linux`) be (`containers_image_ostree_stub` or `!linux`)?

vbatts:
@mtrmac yea, i think you're right. Updated. PTAL

vbatts:
@mtrmac off-topic of this PR, but similar needs to be done between `./signature/mechanism_gpgme.go` and `./signature/mechanism_openpgp.go` but there is something more tricky in that arrangement, including undefined functions like:
```
vendor/github.com/containers/image/signature/mechanism.go:48:9: undefined: newGPGSigningMechanismInDirectory
vendor/github.com/containers/image/signature/mechanism.go:56:9: undefined: newEphemeralGPGSigningMechanism
```

mtrmac:
:+1: Thanks!

mtrmac:
> @mtrmac off-topic of this PR, but similar needs to be done between `./signature/mechanism_gpgme.go` and `./signature/mechanism_openpgp.go`

Not exactly the same thing; GpgME can be installed on macOS (and we do use it there by default). Maybe we might want to disable GpgME on Windows, or autodetect its availability.

> but there is something more tricky in that arrangement, including undefined functions like:
> ```
> vendor/github.com/containers/image/signature/mechanism.go:48:9: undefined: newGPGSigningMechanismInDirectory
> vendor/github.com/containers/image/signature/mechanism.go:56:9: undefined: newEphemeralGPGSigningMechanism
> ```
That looks like you have disabled both of the mechanisms, which is not supported; at least one of them must be enabled.  (The `openpgp` one does not require `cgo`, so it should be an acceptable universal fallback.)

vbatts:
i can't figure out how to get the mechanism to just choose openpgp when
building on windows. :-\

vbatts:
updated. PTAL

rhatdan:
LGTM

mtrmac:
@runcom ping

vbatts:
@runcom review please?

runcom:
lgtm

PR #410, opened 2018-01-28
Use client.HandleErrorResponse in HasBlob when unauthorized

### LINKED ISSUE #409: Return errcode from HasBlob rather than error
The code in docker_image_dest.go:HasBlob() (https://github.com/containers/image/blob/master/docker/docker_image_dest.go#L193-L205) returns an error which then has to be parsed to determine if the problem was an authorization error, read error from the destination, or other error.  It would be good if it could also (or maybe instead) return an error code.   Then code like what was done in https://github.com/projectatomic/buildah/pull/406 for GetFailureCauses() could check the code rather than parsing the returned error.  This was a suggestion from @ripcurld0.

ripcurld0:
closes #409

### DIFF
--- a/docker/docker_image_dest.go
+++ b/docker/docker_image_dest.go
@@ -196,7 +196,7 @@ func (d *dockerImageDestination) HasBlob(info types.BlobInfo) (bool, int64, erro
 		return true, getBlobSize(res), nil
 	case http.StatusUnauthorized:
 		logrus.Debugf("... not authorized")
-		return false, -1, errors.Errorf("not authorized to read from destination repository %s", reference.Path(d.ref.ref))
+		return false, -1, client.HandleErrorResponse(res)
 	case http.StatusNotFound:
 		logrus.Debugf("... not present")
 		return false, -1, nil


mtrmac:
Note that the resulting error message is a bit ugly:

```console
$ ./skopeo --debug --override-os linux --policy ./default-policy.json copy docker://busybox docker://redhat/busybox:mitr
…
DEBU[0002] HEAD https://registry-1.docker.io/v2/redhat/busybox/blobs/sha256:57310166fe88e0dc63a80ca5c219283a932db0f3969712e2f8a86ada143bf566 
DEBU[0003] ... not authorized                           
FATA[0003] Error checking for blob sha256:57310166fe88e0dc63a80ca5c219283a932db0f3969712e2f8a86ada143bf566 at destination: errors:
error parsing HTTP 401 response body: unexpected end of JSON input: ""
```

where the error returned by `HasBlob` starts at `errors:` and includes two newlines; its actual structure is
```go
errcode.Errors{errcode.Error{Code:1003, Message:"requested access to the resource is denied", Detail:interface {}(nil)}, (*client.UnexpectedHTTPResponseError)(0xc420622390)} 
```

Is that good enough for you? (Cc: @TomSweeneyRedHat )

---

(The ”unexpected end of JSON input“ is because of https://tools.ietf.org/html/rfc2616#section-9.4 :
> The HEAD method is identical to GET except that the server MUST NOT return a message-body in the response
and `HandleErrorResponse` not prepared to handle this. Maybe the best way to improve upon this would be to change `HandleErrorResponse`)

ripcurld0:
Yep, that looks a bit ugly but IMHO I must say that it produces a better error message than "not authorized".

mtrmac:
👍 

FWIW the old version is
```
FATA[0003] Error checking for blob sha256:57310166fe88e0dc63a80ca5c219283a932db0f3969712e2f8a86ada143bf566 at destination: not authorized to read from destination repository redhat/busybox 
```

Anyway, the new message _is_ better-typed at least, so +1 from me, and I’ll leave the final decision on @TomSweeneyRedHat and @runcom

docker/docker_image_dest.go
```diff
@@ -196,7 +196,7 @@ func (d *dockerImageDestination) HasBlob(info types.BlobInfo) (bool, int64, erro
 		return true, getBlobSize(res), nil
 	case http.StatusUnauthorized:
 		logrus.Debugf("... not authorized")
-		return false, -1, errors.Errorf("not authorized to read from destination repository %s", reference.Path(d.ref.ref))
+		return false, -1, client.HandleErrorResponse(res)
```
TomSweeneyRedHat: I like the old and the new message both.  Could we wrap the new and old together?  Something like:

```
errors.Wrapf(client.HandleErrorResponse(res), "\nnot authorized to read from destination repository %s", reference.Path(d.ref.ref))
```
so we'd end up with something like:

```
FATA[0003] Error checking for blob sha256:57310166fe88e0dc63a80ca5c219283a932db0f3969712e2f8a86ada143bf566 at destination: errors:
error parsing HTTP 401 response body: unexpected end of JSON input: ""
not authorized to read from destination repository redhat/busybox
```
I like the idea of having the repo that's not reachable be in the error message, but am on the fence about that perhaps being too long.  Yeah, I'm greedy and want both.  What do you all think?  I'm fine with rewording that last line if we can make it fit/feel better, the biggy for me is getting the name out there.

mtrmac: The format string passed to `Wrapf` is prepended in front of, not appended to, the wrapped message. So, a
```go
errors.Wrapf(client.HandleErrorResponse(res), "not authorized to read from destination repository %s", reference.Path(d.ref.ref))
```
would end up as
```
FATA[0003] Error checking for blob sha256:57310166fe88e0dc63a80ca5c219283a932db0f3969712e2f8a86ada143bf566 at destination: not authorized to read from destination repository redhat/busybox: errors:
error parsing HTTP 401 response body: unexpected end of JSON input: ""
```

runcom:
LGTM

PR #411, opened 2018-01-29
Add /etc/containers/certs.d as default certs directory

umohnani8:
Adding /etc/containers/certs.d as another default certs directory
The code will first check /etc/containers/certs.d for the certificates
and if not found at this path it will fall back to /etc/docker/certs.dir

Taking over https://github.com/containers/image/pull/388

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -9,6 +9,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -30,8 +31,6 @@ const (
 	dockerV1Hostname = "index.docker.io"
 	dockerRegistry   = "registry-1.docker.io"
 
-	systemPerHostCertDirPath = "/etc/docker/certs.d"
-
 	resolvedPingV2URL       = "%s://%s/v2/"
 	resolvedPingV1URL       = "%s://%s/v1/_ping"
 	tagsPath                = "/v2/%s/tags/list"
@@ -52,6 +51,7 @@ var (
 	ErrV1NotSupported = errors.New("can't talk to a V1 docker registry")
 	// ErrUnauthorizedForCredentials is returned when the status code returned is 401
 	ErrUnauthorizedForCredentials = errors.New("unable to retrieve auth token: invalid username/password")
+	systemPerHostCertDirPaths     = [2]string{"/etc/containers/certs.d", "/etc/docker/certs.d"}
 )
 
 // extensionSignature and extensionSignatureList come from github.com/openshift/origin/pkg/dockerregistry/server/signaturedispatcher.go:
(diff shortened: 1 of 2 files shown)


umohnani8:
@mtrmac @rhatdan PTAL

docker/docker_client.go
```diff
@@ -133,16 +133,34 @@ func dockerCertDir(ctx *types.SystemContext, hostPort string) string {
...
 	}
 	var hostCertDir string
-	if ctx != nil && ctx.DockerPerHostCertDirPath != "" {
-		hostCertDir = ctx.DockerPerHostCertDirPath
-	} else if ctx != nil && ctx.RootForImplicitAbsolutePaths != "" {
-		hostCertDir = filepath.Join(ctx.RootForImplicitAbsolutePaths, systemPerHostCertDirPath)
-	} else {
-		hostCertDir = systemPerHostCertDirPath
+	for _, systemPerHostCertDirPath := range systemPerHostCertDirPaths {
+		if ctx != nil && ctx.DockerPerHostCertDirPath != "" {
```
mtrmac: This case is independent of the loop variable, and should be moved out of the loop; if the caller specifies a path explicitly, we should use that one regardless of existence.

docker/docker_client.go
```diff
@@ -133,16 +133,34 @@ func dockerCertDir(ctx *types.SystemContext, hostPort string) string {
...
+// pathExists returns true if path exists otherwise false
+func pathExists(path string) bool {
+	_, err := os.Stat(path)
+	if err == nil {
+		return true
+	}
+	if os.IsNotExist(err) {
+		return false
+	}
+	return true
```
mtrmac: This is a bit suspect; the error may be ”a parent directory (`/etc/containers/certs.d`) is unreadable”, for example (note that `/etc/docker/certs.d` is only readable by `root` by default).

Treating other errors as `true` may still be the right thing to do, but it’s not immediately obvious to me that it is.

umohnani8: Should the error be returned instead? This way the user might have an idea of why its failing.

docker/docker_client.go
umohnani8: fixed

umohnani8:
@mtrmac I don't understand why two tests are failing on Travis but they pass locally on my machine. Any idea?

docker/docker_client.go
mtrmac: Constraints:

- If the directory exists _and is readable_, it must be used (the happy “things work” path)
- If the directory definitely does not exist, we must try the other one
- If a permission is denied accessing the `certs.d` directory or the `hostPort` subdirectory, we must not fail entirely (the “unprivileged user with root-only certificates” case); we should try with the default TLS configuration.

Lower-priority concerns:
- I/O errors should not be silently ignored; preferably we should abort and return the I/O error.
- If a permission is denied accessing `/` or `/etc`, or other unlikely cases, it would be nice to treat them same as I/O errors — but we can’t easily distinguish this from the “`certs.d` unreadable” case, which is more important.
- _At least_ the debug logs should make it clear what is going on and why (see how `SetupCertificates` logs its decisions)

In theory it could be possible for `/etc/containers/certs.d` to exist and be unaccessible, and `/etc/docker/certs.d` to exist and be usable; I don’t think this inconsistent case is worth worrying about too much. Same for `/etc/containers/certs.d/$hostPort` existing but being unaccessible (i.e. `Stat` returning no error but `syscall.Access` reporting the directory as being not accessible).

---

With the limited error information easily available in the `os` package, the above probably translates to

- `err == nil`: use the path
- `os.IsNotExist(err)`: try the next path
- `os.IsPermission(err)`: log this, try the next path
- other `err != nil` cases: fail, return the `err`

---

Given the above semantics requirements, and especially the need to log the reasons for the decisions, it seems to me that the decision can’t be neatly abstracted/packaged into something `certs.d`-agnostic like `pathExists`; it may be cleaner to inline the decision into the caller, even if that makes it a big bigger.

mtrmac:
> @mtrmac I don't understand why two tests are failing on Travis but they pass locally on my machine. Any idea?

I guess the Travis container, as built by `.travis.Dockerfile`, does not have a `/etc/containers/certs.d/localhost:5000` directory, but your machine does.

One way to make the test behave reasonably (but not quite, strictly speaking) predictably would be to change the `registryHostPort` inside `TestDockerCertDir` to `thishostdefinitelydoesnotexist:5000`, and update the expected values.

docker/docker_client.go
umohnani8: fixed.

docker/docker_client.go
```diff
@@ -128,19 +128,38 @@ func serverDefault() *tls.Config {
...
+		}
+		if os.IsPermission(err) {
+			logrus.Debugf("error accessing certs directory due to permissions: %v", err)
+			continue
+		}
+		if err != nil {
+			return "", err
+		}
+	}
+	return filepath.Join(hostCertDir, hostPort), nil
```
mtrmac: This path is already computed above; store it in a variable and reuse it?

docker/docker_client_test.go
```diff
@@ -23,17 +23,17 @@ func TestDockerCertDir(t *testing.T) {
 	const nondefaultPerHostDir = "/this/is/not/the/default/certs.d"
 	const variableReference = "$HOME"
 	const rootPrefix = "/root/prefix"
-	const registryHostPort = "localhost:5000"
+	const registryHostPort = "thishostdefinitelydoesnotexist:5000"
 
-	systemPerHostResult := filepath.Join(systemPerHostCertDirPath, registryHostPort)
+	systemPerHostResultDocker := filepath.Join(systemPerHostCertDirPaths[1], registryHostPort)
```
mtrmac: Maybe `systemPerHostCertDirPaths[len(systemPerHostCertDirPaths)-1]`, and rename the variable, to be both future-proof and not hard-code a magic `1` constant in here? (Very low-priority, the tests would reliably break if this ever got out of sync.)

docker/docker_client_test.go
```diff
@@ -85,7 +85,8 @@ func TestDockerCertDir(t *testing.T) {
 			filepath.Join(variableReference, registryHostPort),
 		},
 	} {
-		path := dockerCertDir(c.ctx, registryHostPort)
+		path, err := dockerCertDir(c.ctx, registryHostPort)
+		assert.Equal(t, err, nil)
```
mtrmac: `require.Equal` would more accurate; if `err` is non-`nil`, comparing the returned `path` value is meaningless.

docker/docker_client.go
umohnani8: fixed

docker/docker_client_test.go
umohnani8: fixed

docker/docker_client_test.go
umohnani8: fixed

umohnani8:
@runcom PTAL

umohnani8:
@mtrmac looks like your approval didn't get in. Could you please approve again?
@runcom PTAL.

mtrmac:
👍

runcom:
:+1:

PR #414, opened 2018-02-09
Revert "image: make imageInspectInfo() use manifest.Inspect()"

### LINKED ISSUE #413: Regression in the returning of inspected labels
Since the commit 82b213b, the following command had returned an empty dict instead of fully populated Labels:

`skopeo --debug inspect docker://registry.access.redhat.com/rhosp12/openstack-nova-api:latest`

Since this is a regression and the revert is clean I'm going to propose a PR with the revert.

Whatever the fix, this affects builds of skopeo-0.1.28 so revendoring will be required.

steveb:
This reverts commit 82b213b830bd67ee1a71a145aea7b171117aecd2.

This change caused a regression where the Labels dict is empty when
inspecting images in the registry registry.access.redhat.com. For
example:

  skopeo inspect docker://registry.access.redhat.com/rhosp12/openstack-nova-api:latest

Returns "Labels": {} instead of the expected 22 entries

Signed-off-by: Steve Baker <sbaker@redhat.com>
Closes #413

### DIFF
--- a/image/docker_schema1.go
+++ b/image/docker_schema1.go
@@ -88,7 +88,21 @@ func (m *manifestSchema1) EmbeddedDockerReferenceConflicts(ref reference.Named) 
 }
 
 func (m *manifestSchema1) imageInspectInfo() (*types.ImageInspectInfo, error) {
-	return m.m.Inspect(nil)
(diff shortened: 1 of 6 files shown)


mtrmac:
This was fixed in #416.

PR #416, opened 2018-02-13
#305 cleanup

### LINKED ISSUE #413: Regression in the returning of inspected labels

mtrmac:
This is a set of quite independent cleanups and minor bug fixes on top of #305 .  Notably resolves #413, and adds tests for the schema1 code to decrease risk of similar breakage in the future.

See individual commit messages for more details. I’ll be happy to break this up into several PRs if any part needs more discussion than others.

Cc: @nalind 

- Simplify `types.Image.Inspect` call chain
- Make `ImageInspectInfo.Created` optional
- Move `LayerInfosForCopy` from `UnparsedImage` to `Image`/`sourcedImage`
- Populate labels in `manifest.Schema1.Inspect`
- Clean up manifest types
- Return `""` in `computeID` for unknown manifest types
- Instead of a type switch + cast + error, let the type switch do the conversion
- Rename `manifest.Schema1.ToSchema2` to `ToSchema2Config`
- Load a `*storage.Image` only once in `storageImageSource`
- Drop unused `storageImageDestination.{image,systemContext}`
- Create a new slice in `Schema1.UpdateLayerInfos`
- Add tests for `image.manifestSchema1`

### DIFF
--- a/manifest/docker_schema1.go
+++ b/manifest/docker_schema1.go
@@ -223,26 +223,18 @@ func (m *Schema1) ToSchema2(diffIDs []digest.Digest) ([]byte, error) {
 	if len(m.History) == 0 {
 		return nil, errors.New("image has no layers")
 	}
-	s2 := struct {
-		Schema2Image
-		ID        string `json:"id,omitempty"`
-		Parent    string `json:"parent,omitempty"`
-		ParentID  string `json:"parent_id,omitempty"`
-		LayerID   string `json:"layer_id,omitempty"`
-		ThrowAway bool   `json:"throwaway,omitempty"`
-		Size      int64  `json:",omitempty"`
-	}{}
+	s1 := Schema2V1Image{}
 	config := []byte(m.History[0].V1Compatibility)
-	err := json.Unmarshal(config, &s2)
+	err := json.Unmarshal(config, &s1)
 	if err != nil {
 		return nil, errors.Wrapf(err, "error decoding configuration")
 	}
 	// Images created with versions prior to 1.8.3 require us to re-encode the encoded object,
 	// adding some fields that aren't "omitempty".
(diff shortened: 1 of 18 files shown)


mtrmac:
This breaks API, so tests are going to fail; corresponding skopeo PR is https://github.com/projectatomic/skopeo/pull/477 .

mtrmac:
@runcom ping

runcom:
LGTM, @nalind please ack and I'll merge

manifest/docker_schema1.go
```diff
@@ -202,44 +203,39 @@ func (m *Schema1) Inspect(_ func(types.BlobInfo) ([]byte, error)) (*types.ImageI
...
 		DockerVersion: s1.DockerVersion,
-		Labels:        make(map[string]string),
 		Architecture:  s1.Architecture,
 		Os:            s1.OS,
 		Layers:        LayerInfosToStrings(m.LayerInfos()),
-	}, nil
+	}
+	if s1.Config != nil {
+		i.Labels = s1.Config.Labels
+	}
```
nalind: Does this leave `i.Labels` set to `nil` if the compat blob doesn't specify any labels?

nalind:
One nit, otherwise LGTM.

manifest/docker_schema1.go
mtrmac: “A `nil` map is equivalent to an empty map except that no elements may be added.”, which seems acceptable here (and consistent with schema2).

mtrmac:
👍

PR #418, opened 2018-02-19
Normalize source manifest MIME type before determining possible conversions

mtrmac:
Without this, images from misguided registries which return `text/plain` for s1 were treated as needing conversion to any MIME-type-restricted destination (and choosing to convert to the destination's preferred type, usually s2).

Now, if the destination accepts s1, we treat `text/plain` equivalently to specifying s1, and do no conversion.

### DIFF
--- a/copy/manifest_test.go
+++ b/copy/manifest_test.go
@@ -111,21 +111,28 @@ func TestDetermineManifestConversion(t *testing.T) {
 		{"s1→s1s2", manifest.DockerV2Schema1SignedMediaType, supportS1S2, "", []string{manifest.DockerV2Schema2MediaType, manifest.DockerV2Schema1MediaType}},
 		{"s2→s1s2", manifest.DockerV2Schema2MediaType, supportS1S2, "", supportOnlyS1},
 		{"s1→s1", manifest.DockerV2Schema1SignedMediaType, supportOnlyS1, "", []string{manifest.DockerV2Schema1MediaType}},
+		// text/plain is normalized to s1, and if the destination accepts s1, no conversion happens.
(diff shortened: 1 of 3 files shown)


copy/manifest_test.go
```diff
@@ -111,21 +111,26 @@ func TestDetermineManifestConversion(t *testing.T) {
...
 		// Conversion necessary, try the preferred formats in order.
+		// We abuse manifest.DockerV2ListMediaType here as a MIME type which is not in supportS1S2OCI,
+		// but is still recognized by manifest.NormalizedMIMEType and not normalized to s1
 		{
-			"special→s2", "this needs conversion", supportS1S2OCI, manifest.DockerV2Schema2MediaType,
+			"special→s2", manifest.DockerV2ListMediaType, supportS1S2OCI, manifest.DockerV2Schema2MediaType,
 			[]string{manifest.DockerV2Schema1SignedMediaType, v1.MediaTypeImageManifest, manifest.DockerV2Schema1MediaType},
 		},
 		{
-			"special→s1", "this needs conversion", supportS1OCI, manifest.DockerV2Schema1SignedMediaType,
```
runcom: do we want cases like that to still be covered? or a generic `""` catch-all test to return v2s1 right?

mtrmac: This section of the test cases does not test normalization, but that the hard-coded priority list along with the list of destination-supported types are interpreted correctly.  For that, we need the source type to just be something which is different from any types in the priority order or in the list of destination-supported types.

This was not really intended to cover the cases when the source declares a nonsensical MIME type; this is now tested by entirely new cases added in new lines 114–116.

runcom:
LGTM

mtrmac:
@runcom Updated, PTAL once more please.

runcom:
still LGTM, thanks for the extra tests

mtrmac:
👍

PR #412, opened 2018-02-06
Fix error handling in dirImageSource.GetBlob

mtrmac:
Return the error values instead of `nil`.

(Also modifies the returned size to be `-1` instead of `0` on the error paths, purely out of paranoia.)

PR #419, opened 2018-02-19
Remove .tar extension from blob and config file names

umohnani8:
The config file was being saved as digest.tar for the directory transport.
This is misleading as the config file is not a tar archive.
Dropped the .tar extension from all files now, including blobs.
The user can use a tool like file to determine the format of the files in the directory.

Fixes issue https://github.com/projectatomic/buildah/issues/481

### DIFF
--- a/directory/directory_dest.go
+++ b/directory/directory_dest.go
@@ -12,7 +12,7 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-const version = "Directory Transport Version: 1.0\n"
+const version = "Directory Transport Version: 1.1\n"
 
 // ErrNotContainerImageDir indicates that the directory doesn't match the expected contents of a directory created
 // using the 'dir' transport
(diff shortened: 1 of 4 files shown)


umohnani8:
@mtrmac @runcom PTAL

mtrmac:
Maks sense overall, but this is an incompatible change, so it should increase the version number.

Can you prepare a parallel PR to fix the skopeo tests, please?

(Highlighting: This will break existing automated consumers of the `dir:` format, they will need to adjust.)

rhatdan:
LGTM

TomSweeneyRedHat:
FYI Buildah uses the dire: format pretty heavily in its bats tests.

TomSweeneyRedHat:
Change LGTM, Travis is unhappy and we'll need to roll into the other projects.

umohnani8:
@mtrmac I had to incorporate the changes from your PR https://github.com/containers/image/pull/412 into this to fix the error occurring in the skopeo tests.

umohnani8:
Increased the version number to 1.1 from 1.0

mtrmac:
LGTM pending tests. Thanks!

umohnani8:
@runcom PTAL

rhatdan:
@umohnani8 Why is test blowing up?

runcom:
LGTM

runcom:
@umohnani8 go ahead and fix https://github.com/projectatomic/skopeo/pull/479

Issue #312, opened 2017-07-20
copy does not use mount support for registry v2


chlunde:
Registry v2 has an option for [mounting](https://docs.docker.com/registry/spec/api/#cross-repository-blob-mount) a blob instead of re-uploading a blob:

    POST /v2/<name>/blobs/uploads/?mount=<digest>&from=<repository name>
    Content-Length: 0

This feature could be used when copying within the same registry, for example:

    skopeo copy docker://rexample.com/prj/img1:tag docker://example.com/prj/img2:tag

With skopeo, this downloads and uploads all the data.  With docker, if I have ever pulled img1:tag, running docker tag img2 and docker push on img2 will only copy metadata.

It would also be nice if it could be used to avoid uploading other blobs when copying across registries.

Docker handles this by having a local cache of known digests pushed/pulled from this machine.
Another option could be to query the registry to see what layers the user has access to, but for a small push I guess that might be slower than just pushing directly.  However, it looks like at least for OCR no users have access to the `/v2/_catalog` service.

rhatdan:
@chlunde @mtrmac @vrothberg Is this two year old issue still valid?

vrothberg:
https://github.com/containers/image/pull/536 fixed that.

mtrmac:
Not quite; with an empty blob info cache,
```
skopeo copy docker://localhost:5000/{foo,baz}
```
won’t use mounts, because the “known location” is recorded into the cache only after reading of a blob starts, or writing it finishes, and the attempt to mount happens before either of these.

i.e. we _first_ decide that we know no existing location that could be used for mounting to `baz/`_digest_, and _then_ open `foo/`_digest_ for reading and realize that that there is a location that could possibly be used for mounting.

After one such read/write, the blob info cache for that registry does record at least one location of the blob, and future copies will use mounts.


Handling the special case of intra-registry copies with no cache seems non-trivial (Would we call `TryReusingBlob` one last time in `copyBlobFromStream`? Only if `GetBlob` indicated that it is worth trying again? Only if the source/destination transport is the same? Do we need an extra “this is a second attempt” flag for `TryReusingBlob` to keep the layer indexes to be introduced by #611 in sync?).

Still, I think it’s worth keeping a low-priority issue open to perhaps account for this in some future redesign that could make this easier… somehow…

mtrmac:
OCI distribution spec has also added a “mount blob without specifying the source repo” API, which also might be relevant here.

PR #421, opened 2018-02-28
Add cross-repository mount support to copy.

dlorenc:
This is mostly just a prototype as I'm not sure the best way to go about making this change, but it works for me.

Please feel free to suggest other approaches.

Signed-off-by: dlorenc <lorenc.d@gmail.com>

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -12,9 +12,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/containers/image/docker"
-	"github.com/containers/image/docker/reference"
-
 	"github.com/containers/image/image"
 	"github.com/containers/image/pkg/compression"
 	"github.com/containers/image/signature"
@@ -103,7 +100,6 @@ type Options struct {
 	Progress         chan types.ProgressProperties // Reported to when ProgressInterval has arrived for a single artifact+offset.
 	// manifest MIME type of image set by user. "" is default and means use the autodetection to the the manifest MIME type
 	ForceManifestMIMEType string
-	Mounts                []reference.Named // Optional cross-repository mounts to use. Only works with docker registry destinations.
 }
 
 // Image copies image from srcRef to destRef, using policyContext to validate
@@ -127,15 +123,6 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 	dest, err := destRef.NewImageDestination(options.DestinationCtx)
 	if err != nil {
 		return errors.Wrapf(err, "Error initializing destination %s", transports.ImageName(destRef))
-	}
-	if options.Mounts != nil {
-		c, ok := dest.(docker.Mountable)
-		if !ok {
-			return errors.Wrapf(err, "Destination %s is not mountable, but mount option %s specified", dest, options.Mounts)
-		}
-		if err := c.SetMounts(options.Mounts); err != nil {
-			return errors.Wrapf(err, "Error setting mounts: %s", options.Mounts)
-		}
 	}
 	defer func() {
 		if err := dest.Close(); err != nil {
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -12,6 +12,9 @@ import (
 	"strings"
 	"time"
 
+	"github.com/containers/image/docker"
+	"github.com/containers/image/docker/reference"
+
 	"github.com/containers/image/image"
 	"github.com/containers/image/pkg/compression"
 	"github.com/containers/image/signature"
@@ -100,6 +103,7 @@ type Options struct {
 	Progress         chan types.ProgressProperties // Reported to when ProgressInterval has arrived for a single artifact+offset.
 	// manifest MIME type of image set by user. "" is default and means use the autodetection to the the manifest MIME type
 	ForceManifestMIMEType string
+	Mounts                []reference.Named // Optional cross-repository mounts to use. Only works with docker registry destinations.
 }
 
 // Image copies image from srcRef to destRef, using policyContext to validate
@@ -123,6 +127,15 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 	dest, err := destRef.NewImageDestination(options.DestinationCtx)
 	if err != nil {
 		return errors.Wrapf(err, "Error initializing destination %s", transports.ImageName(destRef))
+	}
+	if options.Mounts != nil {
+		c, ok := dest.(docker.Mountable)
+		if !ok {
+			return errors.Wrapf(err, "Destination %s is not mountable, but mount option %s specified", dest, options.Mounts)
+		}
+		if err := c.SetMounts(options.Mounts); err != nil {
+			return errors.Wrapf(err, "Error setting mounts: %s", options.Mounts)
+		}
 	}
 	defer func() {
 		if err := dest.Close(); err != nil {
--- a/docker/docker_image_dest.go
+++ b/docker/docker_image_dest.go
@@ -12,7 +12,6 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
-	"strings"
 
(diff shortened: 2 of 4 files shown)


dlorenc:
Ref #312

copy/copy.go
```diff
@@ -124,6 +128,15 @@ func Image(policyContext *signature.PolicyContext, destRef, srcRef types.ImageRe
 	if err != nil {
 		return errors.Wrapf(err, "Error initializing destination %s", transports.ImageName(destRef))
 	}
+	if options.Mounts != nil {
+		c, ok := dest.(docker.Mountable)
```
mtrmac: No explicitly transport-specific code/imports in the generic `copy` interface, please.

In principle, this could just be another method in the `ImageDestination` interface.

But in this case, adding this as a specifically `docker`-specific option to `SystemContext`, completely invisible to `copy.Image`, would probably be more appropriate.

docker/docker_image_dest.go
```diff
@@ -104,6 +122,19 @@ func (c *sizeCounter) Write(p []byte) (n int, err error) {
 	return len(p), nil
 }
 
+func prepareMountQuery(mounts []reference.Named, digest string) string {
+	var repositories []string
+	for _, m := range mounts {
+		// Mount only takes the path, not the hostname
+		mount := reference.Path(m)
```
mtrmac: Still, shouldn’t this filter out any references where the registry hostname+port doesn’t match the current destination?

docker/docker_image_dest.go
```diff
@@ -104,6 +122,19 @@ func (c *sizeCounter) Write(p []byte) (n int, err error) {
...
+		repositories = append(repositories, "from="+mount)
+	}
+	if len(repositories) > 0 {
+		return fmt.Sprintf("mount=%s&%s", digest, strings.Join(repositories, "&"))
```
mtrmac: AFAICT, looking both at the official API spec and at the docker/distribution implementation, the `from` parameter does not support multiple values in a single request.

docker/docker_image_dest.go
```diff
@@ -120,15 +151,25 @@ func (d *dockerImageDestination) PutBlob(stream io.Reader, inputInfo types.BlobI
 			return types.BlobInfo{Digest: inputInfo.Digest, Size: size}, nil
 		}
 	}
+	mountQuery := prepareMountQuery(d.mounts, inputInfo.Digest.String())
 
 	// FIXME? Chunked upload, progress reporting, etc.
 	uploadPath := fmt.Sprintf(blobUploadPath, reference.Path(d.ref.ref))
+	if mountQuery != "" {
```
mtrmac: Nit: it would be cleaner to have a single chunk dealing with mounts, and maybe to have all of it conditional on `len(d.mounts != 0)`, instead of treating the `""` value a bit magically.

docker/docker_image_dest.go
```diff
@@ -120,15 +151,25 @@ func (d *dockerImageDestination) PutBlob(stream io.Reader, inputInfo types.BlobI
...
+	}
 	logrus.Debugf("Uploading %s", uploadPath)
 	res, err := d.c.makeRequest(context.TODO(), "POST", uploadPath, nil, nil)
 	if err != nil {
 		return types.BlobInfo{}, err
 	}
 	defer res.Body.Close()
+	if res.StatusCode == http.StatusCreated {
+		logrus.Debugf("Layer mounted: %s", inputInfo.Digest)
+		return types.BlobInfo{Digest: inputInfo.Digest}, nil
```
mtrmac: `PutBlob`, if successful, must also set the `Size` member of `types.BlobInfo` (so that schema2 manifests can be created).

docker/docker_image_dest.go
mtrmac: … or maybe they should be refused, with an error, already in `SetMounts.

Anyway, if this moves into `SystemContext`, the location of the verification would be different.

copy/copy.go
dlorenc: Thanks for the review, SystemContext makes much more sense. Changed to that.

docker/docker_image_dest.go
dlorenc: Ack, this changed a bit. LMK what you think.

docker/docker_image_dest.go
dlorenc: Strange, GCR definitely supports multiple, but I'm not aware of any actual use-cases for this so I dropped it.

docker/docker_image_dest.go
dlorenc: This was simplified a bit by dropping support for multiple mounts.

docker/docker_image_dest.go
dlorenc: Fixed.

docker/docker_image_dest.go
```diff
@@ -38,12 +39,29 @@ func newImageDestination(ctx *types.SystemContext, ref dockerReference) (types.I
 	if err != nil {
 		return nil, err
 	}
+	if ctx.DockerMount != nil {
```
mtrmac: A `*types.SystemContext` can always be `nil`, and should be treated as if it were `&types.SystemContext{}`.

docker/docker_image_dest.go
```diff
@@ -120,15 +145,26 @@ func (d *dockerImageDestination) PutBlob(stream io.Reader, inputInfo types.BlobI
 			return types.BlobInfo{Digest: inputInfo.Digest, Size: size}, nil
 		}
 	}
+	mountQuery := prepareMountQuery(d.mount, inputInfo.Digest.String())
```
mtrmac: `inputInfo.Digest` may be unknown (`""`) at this time (e.g. when the blob is being compressed).  In that case the mounting should probably not be attempted. (The `docker/distribution` implementation ignores the `mount` parameter if it is empty, but that is not a documented part of the API AFAICS.)

docker/docker_image_dest.go
```diff
@@ -120,15 +145,26 @@ func (d *dockerImageDestination) PutBlob(stream io.Reader, inputInfo types.BlobI
...
+		uploadPath = fmt.Sprintf("%s?%s", uploadPath, mountQuery)
+	}
 	logrus.Debugf("Uploading %s", uploadPath)
 	res, err := d.c.makeRequest(context.TODO(), "POST", uploadPath, nil, nil)
 	if err != nil {
 		return types.BlobInfo{}, err
 	}
 	defer res.Body.Close()
+	if res.StatusCode == http.StatusCreated {
+		size := getBlobSize(res)
```
mtrmac: https://github.com/docker/distribution/blob/master/docs/spec/api.md#mount-blob explicitly says that `Content-Length` of the response must be zero, so I don’t think this will work. `d.HasBlob` will probably be necessary to get the size.

types/types.go
```diff
@@ -378,6 +378,9 @@ type SystemContext struct {
 	// === dir.Transport overrides ===
 	// DirForceCompress compresses the image layers if set to true
 	DirForceCompress bool
+
+	// Optional cross-repository mount to use for Docker registry destionations.
+	DockerMount *reference.Named
```
mtrmac: Nits:
- Please move this into the `=== docker.Transport overrides ===` section.
- `s/destionations/destinations/`

rhatdan:
@dlorenc are you still interested in this?  Working on it?

rhatdan:
@dlorenc Closing for lack of activity.  @mtrmac @vrothberg PTAL if you think this is something useful, please reopen or take it over.

mtrmac:
FWIW the recent #536 should do the right thing in many cases (but, I think, not in the case of a `skopeo copy docker://same-host/never-seen-image docker://same-host/different-repo` — the blob mounting data is only recorded on first actual read/write of the layer, which happens _after_ the check whether a blob could be mounted into `different-repo`).

PR #422, opened 2018-02-28
ostree: provide DockerReference when the image is fully qualified

giuseppe:
skopeo copy "--sign-by=" will work after this change when ostree is used
as a destination.

### DIFF
--- a/ostree/ostree_transport.go
+++ b/ostree/ostree_transport.go
@@ -142,6 +142,13 @@ func (ref ostreeReference) StringWithinTransport() string {
 // (fully explicit, i.e. !reference.IsNameOnly, but reflecting user intent,
 // not e.g. after redirect or alias processing), or nil if unknown/not applicable.
 func (ref ostreeReference) DockerReference() reference.Named {
+	img, err := reference.ParseNormalizedNamed(ref.image)
+	if err != nil {
+		return nil
+	}
+	if !reference.IsNameOnly(img) {
+		return img
+	}
 	return nil
 }
 
--- a/ostree/ostree_transport_test.go
+++ b/ostree/ostree_transport_test.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/containers/image/docker/reference"
 	_ "github.com/containers/image/internal/testing/explicitfilepath-tmpdir"
 	"github.com/containers/image/types"
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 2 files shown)


mtrmac:
Most importantly, is `--sign-by=` when copying _into_ ostree actually valuable?

AFAICS it will create a signature for the original manifest, with probably compressed layers; and extract the layers and record the uncompressed files. Then reading out of ostree (e.g. `skopeo copy ostree:… …`) will return the original manifest with compressed layers, along with `LayerInfosForCopy` returning uncompressed layers, and `imageCopier.copyLayer` will abort because the signature would be broken by using `LayerInfosForCopy`.

I don’t know, maybe there is some value in allowing images to be signed “on pull”, only to be able o set up a policy enforcing that all running images have a signature, but I’d expect that signing the image on the registry, for everyone, would be a much more natural workflow.

- This will create a signature for `docker.io/library/busybox` when copying into `ostree:busybox`; is that desirable and consistent with the `ostree:` naming semantics?
- Tests need updating

giuseppe:
I get this error when I try to copy an image with signatures from ostree to a registry:

```
Checking if image destination supports signatures
FATA[0000] Internal error: copyLayers() needs to use an updated manifest but that was known to be forbidden 
```

so I guess it is not really useful.  I've added it only for completeness as I was testing different combinations and I saw the ostree backend failing on `--sign-by`

mtrmac:
Yeah, that’s the failure I expected.  That’s not a blocking problem in itself – the image in the `ostree:` storage _is_ signed and can be validated, at least with something like https://github.com/containers/image/pull/288 .

The naming and normalization semantics (`ostree:busybox`) do need to be consciously decided on before we add the support, though.  (@baude ’s work on a shared name resolution library for libpod etc. may be relevant here, the semantics should be at least compatible if not exactly the same.)

giuseppe:
even if names like "busybox" are allowed, atomic raises a warning as that is not suggested and we rather want users to use the fully qualified name "docker.io/busybox" as we don't have a "default registry" concept.  So we could expect that the destination will rather be something like `ostree:docker.io/busybox`

mtrmac:
> even if names like "busybox" are allowed, atomic raises a warning as that is not suggested and we rather want users to use the fully qualified name "docker.io/busybox"

[nit: The fully qualified variant is `docker.io/library/busybox`]

> So we could expect that the destination will rather be something like ostree:docker.io/busybox

Well, the code does accept `busybox`. It should either refuse such input, or accept it and give it some definite semantics (of course, semantics can still be “there is no corresponding Docker reference for such names”).

Accepting an input and not knowing/caring/defining what it means is a bit unsatisfactory.

Another aspect of this is `PolicyConfiguration{Identity,Namespaces}`, which treats any top-level word equivalently to a hostname, i.e. a top-level container: a `docker.io/library/busybox` name can also be configured in `docker.io` and `docker.io/library` higher-level containers; similarly `busybox` can be a higher-level container for a `busybox/foo/bar/baz` name. This matters in the sense that `PolicyConfigurationIdentity` should map all aliases of the same underlying name to the same canonical identity—though that can be also achieved by changing the `PolicyConfigurationIdentity` mapping.

giuseppe:
thanks for the explanation.

I am going to close this PR for now, as there is not really an use case yet.

PR #423, opened 2018-03-05
Support gzip for docker-archive files

umohnani8:
The docker-archive transport pulls in images from tar file
but would fail if the tar file was compressed using gzip.
This adds the functionality of detecting whether the file is
compressed, and if it is the contents are copied to a temp
uncompressed file.
This was supported for the docker-daemon transport, so moving
that functionality to tarfile/src.go for both docker-archive
and docker-daemon.

### DIFF
--- a/docker/archive/src.go
+++ b/docker/archive/src.go
@@ -13,26 +13,24 @@ type archiveImageSource struct {
 
 // newImageSource returns a types.ImageSource for the specified image reference.
 // The caller must call .Close() on the returned ImageSource.
-func newImageSource(ctx *types.SystemContext, ref archiveReference) types.ImageSource {
+func newImageSource(ctx *types.SystemContext, ref archiveReference) (types.ImageSource, error) {
 	if ref.destinationRef != nil {
 		logrus.Warnf("docker-archive: references are not supported for sources (ignoring)")
 	}
-	src := tarfile.NewSource(ref.path)
+	src, err := tarfile.NewSourceFromFile(ref.path)
+	if err != nil {
+		return nil, err
+	}
 	return &archiveImageSource{
 		Source: src,
 		ref:    ref,
-	}
+	}, nil
 }
 
 // Reference returns the reference used to set up this source, _as specified by the user_
 // (not as the image itself, or its underlying storage, claims).  This can be used e.g. to determine which public keys are trusted for this image.
 func (s *archiveImageSource) Reference() types.ImageReference {
 	return s.ref
-}
-
-// Close removes resources associated with an initialized ImageSource, if any.
-func (s *archiveImageSource) Close() error {
-	return nil
 }
 
 // LayerInfosForCopy() returns updated layer info that should be used when reading, in preference to values in the manifest, if specified.
--- a/docker/daemon/daemon_src.go
+++ b/docker/daemon/daemon_src.go
@@ -1,12 +1,7 @@
 package daemon
 
 import (
-	"io"
-	"io/ioutil"
-	"os"
-
 	"github.com/containers/image/docker/tarfile"
-	"github.com/containers/image/internal/tmpdir"
 	"github.com/containers/image/types"
 	"github.com/pkg/errors"
 	"golang.org/x/net/context"
@@ -15,7 +10,6 @@ import (
 type daemonImageSource struct {
 	ref             daemonReference
 	*tarfile.Source // Implements most of types.ImageSource
-	tarCopyPath     string
 }
 
 type layerInfo struct {
@@ -45,29 +39,13 @@ func newImageSource(ctx *types.SystemContext, ref daemonReference) (types.ImageS
 	}
 	defer inputStream.Close()
 
-	// FIXME: use SystemContext here.
-	tarCopyFile, err := ioutil.TempFile(tmpdir.TemporaryDirectoryForBigFiles(), "docker-daemon-tar")
+	src, err := tarfile.NewSourceFromStream(inputStream)
 	if err != nil {
 		return nil, err
 	}
-	defer tarCopyFile.Close()
-
-	succeeded := false
-	defer func() {
-		if !succeeded {
-			os.Remove(tarCopyFile.Name())
-		}
-	}()
-
-	if _, err := io.Copy(tarCopyFile, inputStream); err != nil {
-		return nil, err
-	}
-
-	succeeded = true
 	return &daemonImageSource{
-		ref:         ref,
-		Source:      tarfile.NewSource(tarCopyFile.Name()),
-		tarCopyPath: tarCopyFile.Name(),
+		ref:    ref,
+		Source: src,
 	}, nil
 }
 
@@ -75,11 +53,6 @@ func newImageSource(ctx *types.SystemContext, ref daemonReference) (types.ImageS
 // (not as the image itself, or its underlying storage, claims).  This can be used e.g. to determine which public keys are trusted for this image.
 func (s *daemonImageSource) Reference() types.ImageReference {
 	return s.ref
-}
-
-// Close removes resources associated with an initialized ImageSource, if any.
-func (s *daemonImageSource) Close() error {
-	return os.Remove(s.tarCopyPath)
 }
 
 // LayerInfosForCopy() returns updated layer info that should be used when reading, in preference to values in the manifest, if specified.
--- a/docker/tarfile/src.go
+++ b/docker/tarfile/src.go
@@ -3,6 +3,7 @@ package tarfile
 import (
 	"archive/tar"
 	"bytes"
+	"compress/gzip"
 	"context"
 	"encoding/json"
 	"io"
@@ -10,6 +11,7 @@ import (
 	"os"
 	"path"
 
+	"github.com/containers/image/internal/tmpdir"
 	"github.com/containers/image/manifest"
 	"github.com/containers/image/pkg/compression"
 	"github.com/containers/image/types"
@@ -19,7 +21,8 @@ import (
 
 // Source is a partial implementation of types.ImageSource for reading from tarPath.
 type Source struct {
-	tarPath string
+	tarPath              string
+	removeTarPathOnClose bool // Remove temp file on close if true
 	// The following data is only available after ensureCachedDataIsPresent() succeeds
 	tarManifest       *ManifestItem // nil if not available yet.
 	configBytes       []byte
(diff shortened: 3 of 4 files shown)


umohnani8:
@mtrmac @runcom PTAL

umohnani8:
Fixes https://github.com/projectatomic/libpod/issues/404

docker/archive/src.go
```diff
@@ -13,15 +14,18 @@ type archiveImageSource struct {
...
 // The caller must call .Close() on the returned ImageSource.
-func newImageSource(ctx *types.SystemContext, ref archiveReference) types.ImageSource {
+func newImageSource(ctx *types.SystemContext, ref archiveReference) (types.ImageSource, error) {
 	if ref.destinationRef != nil {
 		logrus.Warnf("docker-archive: references are not supported for sources (ignoring)")
 	}
-	src := tarfile.NewSource(ref.path)
+	src, err := tarfile.NewSource(ref.path, nil)
+	if err != nil {
+		return nil, errors.Wrap(err, "error getting source")
```
mtrmac: This wrapping seems not all that useful. The caller of `NewImageSource` should already be providing a similar context about setting up a source.

(Non-blocking: Maybe some of the low-level errors in `tarfile.NewSource` should be wrapped with explanations of what/why the code is doing, e.g. when creating a temporary file fails “why is the code complaining about a path I never asked to be used?!”.)

docker/daemon/daemon_src.go
```diff
@@ -45,29 +42,14 @@ func newImageSource(ctx *types.SystemContext, ref daemonReference) (types.ImageS
...
-		return nil, err
-	}
-
-	succeeded = true
 	return &daemonImageSource{
 		ref:         ref,
-		Source:      tarfile.NewSource(tarCopyFile.Name()),
-		tarCopyPath: tarCopyFile.Name(),
+		Source:      src,
+		tarCopyPath: src.GetTarPath(),
```
mtrmac: `tarCopyPath` only exists for deleting the temporary file on `Close`.  Shouldn’t that functionality also be moved from `docker/daemon` to `docker/tarfile`?

docker/tarfile/src.go
```diff
@@ -35,14 +38,63 @@ type layerInfo struct {
...
+// NewSource returns a tarfile.Source for the specified path or inputStream.
+// Cannot set path and inputStream at the same time.
+// inputStream is mainly used for the docker-daemon transport.
+func NewSource(path string, inputStream io.ReadCloser) (*Source, error) {
 	// TODO: We could add support for multiple images in a single archive, so
 	//       that people could use docker-archive:opensuse.tar:opensuse:leap as
 	//       the source of an image.
+
+	if path != "" && inputStream != nil {
+		return nil, errors.Errorf("Cannot set both path and inputStream")
```
mtrmac: If the two parameters can never be set at the same time, there’s no point in allowing them to be passed simultaneously. It would be cleaner to have a `NewFileSource`/`NewStreamSource` (`NewSourceFromFile`/`NewSourceFromStream`? I don’t care) APIs, and factor the common code, if any, into functions private to this module.

docker/tarfile/src.go
```diff
@@ -35,14 +38,63 @@ type layerInfo struct {
...
+			defer reader.Close()
+			isGzip = true
+			inputStream = reader
+		}
+	}
+
+	if isGzip || inputStream != nil {
+		// FIXME: use SystemContext here.
+		// Save inputStream to a temporary file
+		tarCopyFile, err := ioutil.TempFile(tmpdir.TemporaryDirectoryForBigFiles(), "docker-tar")
```
mtrmac: (Nothing removes `tarCopyFile` in this package – see the other comment in `docker/daemon`.)

docker/tarfile/src.go
```diff
@@ -247,6 +299,11 @@ func (s *Source) prepareLayerData(tarManifest *ManifestItem, parsedConfig *manif
 	return knownLayers, nil
 }
 
+// GetTarPath returns the tar path of s
+func (s *Source) GetTarPath() string {
```
mtrmac: Should not be necessary.

docker/tarfile/src.go
```diff
@@ -35,14 +38,63 @@ type layerInfo struct {
...
 	//       that people could use docker-archive:opensuse.tar:opensuse:leap as
 	//       the source of an image.
+
+	if path != "" && inputStream != nil {
+		return nil, errors.Errorf("Cannot set both path and inputStream")
+	}
+
+	// Check if path is a gzip file and read its contents if it is
+	isGzip := false
+	if path != "" {
```
mtrmac: (This is another argument for splitting `NewSource`: it would be awkward if `NewSource` accepted compressed input for files but not for streams; with two separate constructors it’s more acceptable.)

docker/tarfile/src.go
```diff
@@ -35,14 +38,63 @@ type layerInfo struct {
...
+		file, err := os.Open(path)
+		if err != nil {
+			return nil, err
+		}
+		defer file.Close()
+
+		reader, err := gzip.NewReader(file)
```
mtrmac: (Non-blocking: This could also use `pkg/compression.DetectCompression`.)

docker/archive/src.go
umohnani8: Fixed. wrapped the errors in `NewSourceFromFile` and `NewSourceFromStream`.

docker/daemon/daemon_src.go
umohnani8: Moved to docker/tarfile

docker/tarfile/src.go
umohnani8: Separated to two functions, `NewSourceFromFile` and `NewSourceFromStream`.

docker/tarfile/src.go
umohnani8: Moved and being deleted in Source.Close() in docker/tarfile

docker/tarfile/src.go
umohnani8: Removed

docker/tarfile/src.go
umohnani8: Fixed

docker/tarfile/src.go
umohnani8: Still using gzip.NewReader()

docker/archive/src.go
```diff
@@ -32,7 +35,7 @@ func (s *archiveImageSource) Reference() types.ImageReference {
 
 // Close removes resources associated with an initialized ImageSource, if any.
 func (s *archiveImageSource) Close() error {
-	return nil
+	return s.Source.Close()
```
mtrmac: The `Close` method can be completely removed, Go will automatically use the method in the anonymous `Source` member.

docker/tarfile/src.go
```diff
@@ -35,14 +38,55 @@ type layerInfo struct {
...
+			tarPath: path,
+		}, nil
+	}
+	defer reader.Close()
+
+	return NewSourceFromStream(reader)
+}
+
+// NewSourceFromStream returns a tarfile.Source for the specified inputStream.
+func NewSourceFromStream(inputStream io.ReadCloser) (*Source, error) {
```
mtrmac: `inputStream` can be just an `io.Reader` AFAICT.

Please document that the input must be uncompressed.

It would also be nice to document that it is safe for the caller to close the stream immediately after this function returns. (Otherwise, passing a stream to a “read something” object might make the impression that the stream will be consumed over time, and somehow closed at a later time.)

docker/tarfile/src.go
```diff
@@ -35,14 +38,55 @@ type layerInfo struct {
...
 	// TODO: We could add support for multiple images in a single archive, so
```
mtrmac: This `TODO` applies to both functions equally, and is now a bit out of place.  Maybe move it to the doc string for `Source` itself, or as a stand-alone comment near the top of the file.

docker/daemon/daemon_src.go
```diff
@@ -79,7 +57,7 @@ func (s *daemonImageSource) Reference() types.ImageReference {
 
 // Close removes resources associated with an initialized ImageSource, if any.
 func (s *daemonImageSource) Close() error {
-	return os.Remove(s.tarCopyPath)
+	return s.Source.Close()
```
mtrmac: The `Close` method can be completely removed, Go will automatically use the method in the anonymous `Source` member.

docker/tarfile/src.go
```diff
@@ -35,14 +38,55 @@ type layerInfo struct {
...
+	}()
+
+	if _, err := io.Copy(tarCopyFile, inputStream); err != nil {
+		return nil, errors.Wrapf(err, "error copying contents to temporary file %q", tarCopyFile.Name())
+	}
+	succeeded = true
+
+	return &Source{
+		tarPath:     tarCopyFile.Name(),
+		tarCopyPath: tarCopyFile.Name(),
```
mtrmac: (Non-blocking: instead of two strings with the same value there could be a `removeTarPathOnClose` boolean.)

docker/tarfile/src.go
```diff
@@ -35,14 +38,55 @@ type layerInfo struct {
 	size int64
 }
 
-// NewSource returns a tarfile.Source for the specified path.
-func NewSource(path string) *Source {
+// NewSourceFromFile returns a tarfile.Source for the specified path
```
mtrmac: Please document that compressed input is accepted.

docker/daemon/daemon_src.go
umohnani8: Should it be removed from archiveImageSource as well?

umohnani8: nvm, I just saw you mentioned it in the review above.

docker/archive/src.go
umohnani8: fixed

docker/tarfile/src.go
umohnani8: fixed

docker/tarfile/src.go
umohnani8: moved it above the two source functions

docker/daemon/daemon_src.go
umohnani8: fixed

docker/tarfile/src.go
umohnani8: Fixed to use a boolean instead

docker/tarfile/src.go
umohnani8: fixed

docker/tarfile/src.go
```diff
@@ -35,14 +38,58 @@ type layerInfo struct {
...
+		}, nil
 	}
+	defer reader.Close()
+
+	return NewSourceFromStream(reader)
+}
+
+// NewSourceFromStream returns a tarfile.Source for the specified inputStream, which must be uncompressed.
+// The caller can close the inputStream immediately after NewSourceFromFile returns.
+func NewSourceFromStream(inputStream io.ReadCloser) (*Source, error) {
```
mtrmac: Nit: `io.Reader` is sufficient, and makes it clear that this function won’t close the stream.

mtrmac:
👍 again for `pullapprove`

docker/tarfile/src.go
umohnani8: changed it to io.Reader. Thanks for the review!

umohnani8:
@runcom PTAL

runcom:
lgtm

runcom:
@umohnani8 please rebase this and we'll merge it

umohnani8:
rebased

PR #424, opened 2018-03-06
Fix failure to make auth directory

umohnani8:
Whenever the auth directory was defaulting to /run/usr/UID
it was assuming that /run/usr/UID existed. The change creates the
directory if it doesn't exist.
We were also creating the XDG_RUNTIME_DIR deirectory if the variable
was set but the directory didn't exist. Not a good idea so we error out
now when that happens.

Fixes https://github.com/projectatomic/libpod/issues/427

### DIFF
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -27,7 +27,7 @@ type dockerConfigFile struct {
 }
 
 const (
-	defaultPath       = "/run/user"
+	defaultPath       = "/run"
 	authCfg           = "containers"
 	authCfgFileName   = "auth.json"
 	dockerCfg         = ".docker"
@@ -64,7 +64,11 @@ func GetAuthentication(ctx *types.SystemContext, registry string) (string, strin
 	}
 
 	dockerLegacyPath := filepath.Join(homedir.Get(), dockerLegacyCfg)
-	paths := [3]string{getPathToAuth(ctx), filepath.Join(homedir.Get(), dockerCfg, dockerCfgFileName), dockerLegacyPath}
+	pathToAuth, err := getPathToAuth(ctx)
+	if err != nil {
+		return "", "", err
+	}
+	paths := [3]string{pathToAuth, filepath.Join(homedir.Get(), dockerCfg, dockerCfgFileName), dockerLegacyPath}
 
 	for _, path := range paths {
 		legacyFormat := path == dockerLegacyPath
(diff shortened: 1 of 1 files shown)


umohnani8:
@mtrmac @runcom PTAL

pkg/docker/config/config.go
```diff
@@ -132,8 +133,18 @@ func getPathToAuth(ctx *types.SystemContext) string {
...
 	runtimeDir := os.Getenv("XDG_RUNTIME_DIR")
-	if runtimeDir == "" {
+	if runtimeDir != "" {
+		_, err := os.Stat(runtimeDir)
+		if os.IsNotExist(err) {
+			// This means the user set the XDG_RUNTIME_DIR variable and either forgot to create the directory
+			// or made a typo while setting the environment variable
+			// so we log the error and return an empty string as the path
+			logrus.Warnf("%q does not exist", runtimeDir)
+			return ""
```
mtrmac: There are four callers of this, and the PR only fixes one of them.

Why not return an ordinary `error` value and propagate it as usual, instead of hard-coding the knowledge that `""` must mean something about `XDG_RUNTIME_DIR` in all callers?

umohnani8: Overlooked the other callers, my bad. Will return an error instead.

umohnani8:
@mtrmac @rhatdan as per the discussion in https://github.com/projectatomic/libpod/issues/427 the auth file will be created in `/run/containers/UID` if `XDG_RUNTIME_DIR` is not set. The only thing now is for that to successfully happen, the user has to be root for permissions to create a directory under `/run`.
PTAL

pkg/docker/config/config.go
umohnani8: Returning an error now.

rhatdan:
LGTM

pkg/docker/config/config.go
```diff
@@ -123,20 +130,30 @@ func RemoveAllAuthentication(ctx *types.SystemContext) error {
...
 	runtimeDir := os.Getenv("XDG_RUNTIME_DIR")
-	if runtimeDir == "" {
-		runtimeDir = filepath.Join(defaultPath, strconv.Itoa(os.Getuid()))
+	if runtimeDir != "" {
+		_, err := os.Stat(runtimeDir)
+		if os.IsNotExist(err) {
+			// This means the user set the XDG_RUNTIME_DIR variable and either forgot to create the directory
+			// or made a typo while setting the environment variable
+			// so we log the error and return an empty string as the path
+			return "", errors.Wrapf(err, "%q directory set by $XDG_RUNTIME_DIR does not exist. Either create the directory or unset $XDG_RUNTIME_DIR.", runtimeDir)
```
TomSweeneyRedHat: :+1:

mtrmac:
LGTM based on https://github.com/projectatomic/libpod/issues/427#issuecomment-373508361 .

[Approved with PullApprove](https://pullapprove.com/containers/image/pull-request/424/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=containers/image)

umohnani8:
@runcom PTAL

runcom:
LGTM

PR #425, opened 2018-03-07
Match container/storage adding extra parameter to PutLayer.

adelton:
This catches up to change in https://github.com/containers/storage/pull/156.

### DIFF
--- a/storage/storage_image.go
+++ b/storage/storage_image.go
@@ -556,7 +556,7 @@ func (s *storageImageDestination) Commit() error {
 			return errors.Errorf("error applying blob %q: content not found", blob.Digest)
 		}
 		// Build the new layer using the diff, regardless of where it came from.
-		layer, _, err := s.imageRef.transport.store.PutLayer(id, lastLayer, nil, "", false, diff)
+		layer, _, err := s.imageRef.transport.store.PutLayer(id, lastLayer, nil, "", false, nil, diff)
 		if err != nil {
 			return errors.Wrapf(err, "error adding layer with blob %q", blob.Digest)
 		}


TomSweeneyRedHat:
@adelton need to sign this one.  Otherwise the change LGTM.

adelton:
Signed-off-by added -- sorry about that.

adelton:
> You need to change the actual interface as well.

Could you elaborate? The https://github.com/containers/storage/pull/156 was not merged yet -- I assume that once it gets merged, it will start to be used automatically. Won't it?

storage/storage_image.go
```diff
@@ -556,7 +556,8 @@ func (s *storageImageDestination) Commit() error {
 			return errors.Errorf("error applying blob %q: content not found", blob.Digest)
 		}
 		// Build the new layer using the diff, regardless of where it came from.
-		layer, _, err := s.imageRef.transport.store.PutLayer(id, lastLayer, nil, "", false, diff)
+		options := (*storage.LayerOptions)(nil)
+		layer, _, err := s.imageRef.transport.store.PutLayer(id, lastLayer, nil, "", false, options, diff)
```
mtrmac: Why not just `…, false, nil, diff)`?

Or if you wanted to make it easy to add fields in the future, `options := &storage.LayerOptions{}`.  Setting the typed `nil` is a surprising mix of the two ideas which is more complex than either of them.

mtrmac:
> I assume that once it gets merged, it will start to be used automatically. Won't it?

It will in this package (`vendor.conf` specifies `master` branch from c/storage); but we will need an actual updating PR for `skopeo`, which also specifies `master` but actually does vendoring and keeps the vendored in the git repository; c/image runs `skopeo` tests for each PR.

I guess I can take care of that bit of procedure after it becomes necessary.

storage/storage_image.go
adelton: I thought having the `nil` typed would make it more explicit in showing what we try to achieve.

I've changed it to untyped `nil` now -> 1efad7d255b3e1f0250beccbd8d3dc2608e02365.

adelton:
> I guess I can take care of that bit of procedure after it becomes necessary.

Thank you.

This PR is waiting for https://github.com/containers/storage/pull/156 to be merged, so feel free to set any tags / anything about this one to put it aside until the parent one is resolved.

mtrmac:
:+1: pending containers/storage#156 and tests. Thanks!

mtrmac:
Ended up merging equivalent #436 instead.

PR #426, opened 2018-03-09
#305 cleanup 2: Reference handling

mtrmac:
This is a follow-up, and on top of, #416.  Unlike that one, which should be overwhelmingly uncontroversial, this one involves a few design choices.

See individual commit messages for more details. (…and not to read #416 again; start at “Make TestStorageReferenceDockerDeference table-driven“.); please pay particular attention to the “RFC” and “UNTESTED” marked ones. I’ll be happy to break this up into several PRs if any part needs more discussion than others.

General themes:
- Increase test coverage of various accepted, and invalid, reference formats
- Fix various uncovered corner cases, e.g. reject `@`_digest_ without a name, or _name_`@`_digest_`@`_digest_`@`_ID_
- Support _name_`:`_tag_`@`_digest_, which is actually a recognized syntax.  (Or would we refuse it instead?)
- Eliminate `verboseName`
- Rewrite `storageReference` to be a (store, `reference.Named`, ID) triplet, without three redundant  string members, over about 20 tiny commits.
- Significantly simplify `ParseStoreReference` (unless I’m missing a reason for the code to be the way it is)
- Add `types.ImageDestination.EmbeddedDockerReference`, instead of inconsistently lying in `storageImageDestination.Reference.DockerReference()`.

@nalind @runcom PTAL.

### DIFF
--- a/ostree/ostree_dest.go
+++ b/ostree/ostree_dest.go
@@ -125,6 +125,13 @@ func (d *ostreeImageDestination) MustMatchRuntimeOS() bool {
 	return true
 }
 
+// IgnoresEmbeddedDockerReference returns true iff the destination does not care about Image.EmbeddedDockerReferenceConflicts(),
+// and would prefer to receive an unmodified manifest instead of one modified for the destination.
+// Does not make a difference if Reference().DockerReference() is nil.
+func (d *ostreeImageDestination) IgnoresEmbeddedDockerReference() bool {
+	return false // N/A, DockerReference() returns nil.
+}
+
 func (d *ostreeImageDestination) PutBlob(ctx context.Context, stream io.Reader, inputInfo types.BlobInfo, isConfig bool) (types.BlobInfo, error) {
 	tmpDir, err := ioutil.TempDir(d.tmpDirPath, "blob")
 	if err != nil {
--- a/storage/storage_image.go
+++ b/storage/storage_image.go
@@ -703,6 +703,13 @@ func (s *storageImageDestination) MustMatchRuntimeOS() bool {
 	return true
 }
 
+// IgnoresEmbeddedDockerReference returns true iff the destination does not care about Image.EmbeddedDockerReferenceConflicts(),
+// and would prefer to receive an unmodified manifest instead of one modified for the destination.
+// Does not make a difference if Reference().DockerReference() is nil.
+func (s *storageImageDestination) IgnoresEmbeddedDockerReference() bool {
+	return true // Yes, we want the unmodified manifest
+}
+
 // PutSignatures records the image's signatures for committing as a single data blob.
 func (s *storageImageDestination) PutSignatures(ctx context.Context, signatures [][]byte) error {
 	sizes := []int{}
--- a/storage/storage_image.go
+++ b/storage/storage_image.go
@@ -248,6 +248,7 @@ func newImageDestination(imageRef storageReference) (*storageImageDestination, e
 	// digest, and that makes the image unusable if we subsequently try to access it using a
 	// reference that mentions the no-longer-correct digest.
 	publicRef := imageRef
+	publicRef.breakDockerReference = true // FIXME: this does not change .StringWithinTransport(), but modifies DockerReference()
 	publicRef.name = nil
 	image := &storageImageDestination{
 		imageRef:       imageRef,
--- a/storage/storage_image.go
+++ b/storage/storage_image.go
@@ -613,7 +613,7 @@ func (s *storageImageDestination) Commit(ctx context.Context) error {
 	if name := s.imageRef.DockerReference(); len(oldNames) > 0 || name != nil {
 		names := []string{}
 		if name != nil {
-			names = append(names, verboseName(name))
+			names = append(names, name.String())
 		}
 		if len(oldNames) > 0 {
 			names = append(names, oldNames...)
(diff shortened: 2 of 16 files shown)


mtrmac:
More questions:

- Why is `GetStoreImage` doing its own `store.Image(sref.id)` lookup just before a `resolveImage` call? AFAICT that bypasses the repo name match check in `resolveImage`; is that behavior difference intentional?
- And why is it using the `DockerReference.String()` lookup in preference to the `storageReference` special case? That e.g. ignores `storageReference.id`, if it happens to be available.

mtrmac:
The `skopeo `test failure is inherited from #416, the fix is the same https://github.com/projectatomic/skopeo/pull/477 .

storage/storage_image.go
```diff
@@ -51,8 +51,7 @@ type storageImageSource struct {
 }
 
 type storageImageDestination struct {
-	imageRef       storageReference                // The reference we'll use to name the image
-	publicRef      storageReference                // The reference we return when asked about the name we'll give to the image
+	imageRef       storageReference
```
rhatdan: nit: Do we need a comment here?

storage/storage_reference.go
```diff
@@ -9,63 +9,72 @@ import (
...
-		tag:       tag,
-		digest:    digest,
+	}, nil
+}
+
+// imageMatchesRepo returns true iff image.Names contains an element with the same repo as ref
+func imageMatchesRepo(image *storage.Image, ref reference.Named) bool {
+	repo := ref.Name()
+	for _, name := range image.Names {
+		if named, err := reference.ParseNormalizedNamed(name); err == nil {
```
rhatdan: Is it ok to ignore the err?  Should we logrus.Debugf this or is this expected.

rhatdan:
@umohnani8 @runcom @giuseppe PTAL

storage/storage_image.go
mtrmac: “A reference to the image this destination writes to” should be clear enough without a comment; the previous two values needed comments (even more comments, arguably) to explain when to use which value.

storage/storage_reference.go
mtrmac: `imageMatchesRepo` is clearly(?) false for values which can’t be parsed. I don’t know whether such values can happen, or how likely that is. @nalind ?

(This code is not new, it’s only moved out of the calling function (commit “UNTESTED: Extract duplicate code from resolveImage”).)

ostree/ostree_dest.go
```diff
@@ -124,6 +124,13 @@ func (d *ostreeImageDestination) MustMatchRuntimeOS() bool {
 	return true
 }
 
+// IgnoresEmbeddedDockerReference returns true iff the destination does not care about Image.EmbeddedDockerReferenceConflicts(),
+// and would prefer to receive an unmodified manifest instead of one modified for the destination.
+// Does not make a difference if Reference().DockerReference() is nil.
+func (d *ostreeImageDestination) IgnoresEmbeddedDockerReference() bool {
+	return false // N/A, DockerReference() returns nil.
```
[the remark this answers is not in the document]
mtrmac: @giuseppe I’m not _too_ worried about saving a few cycles during an image copy which writes gigabytes of data to the disk.

If anything, it’s more a matter of semantics: does the `ostree` backend prefer a manifest tailored for it, or an unmodified one? Looking at how `ostreeImageDestination.Commit` records the manifest digest in the `ostree` metadata, maybe it really would prefer the unmodified one, and this should return `true` for that reason.

If it does not matter at all, I’d rather leave it at the default `false`; we may eventually want to to have defaults for many of the `ImageDestination` flags without copies in every transport, and that will result in shorter code if more transports use the default values.

rhatdan:
@mtrmac Needs rebase,  What do we need to do to get this moving forward?

mtrmac:
At least one LGTM would be nice.  I can technically force-merge it without one.

rhatdan:
LGTM

giuseppe:
do we want to drop the RFC and UNTESTED tags from the patches before merging?

mtrmac:
> do we want to drop the RFC and UNTESTED tags from the patches before merging?

“RFC: Refuse un-named @digest reference input” is pretty fundamental to this PR, the final form of `storageReference` can not express un-named digests.

“RFC: Simplify parsing of image IDs/digests” is not essential, but it is a prerequisite for “Don’t explicitly parse a digest.”; together they are a fairly notable simplification.  (The _functional_ aspects of them have pretty good test coverage; I just don’t know whether there was any other deliberate intent in structuring the code that way, maybe for error reporting reasons.)

“RFC UNTESTED: Only load an image once in resolveImage” can be dropped without impact on the rest.

It’s been 3 months without any response to the RFCs, so I am not too keen on continuing to wait for them.

---

“UNTESTED: Fix name@digest@IDprefix parsing” could, I suppose, be dropped, by re-creating the presumed bug in other versions of the code — but the result would be just as untested in that case; or we could drop the “RFC: Simplify parsing of image IDs/digests”+“Don’t explicitly parse a digest.” commits as well and give up on simplifying `ParseStoreReference`.

I can easily enough drop “UNTESTED: Match repositories using .Name() instead of FamiliarName()” and “UNTESTED: Extract duplicate code from resolveImage”, though I am fairly confident about them—they are trivial refactorings.

---

As an alternative, would it be easy enough tor un the libpod / buildah test suites against this PR’s branch of containers/image? Is there a test harness that can do that, or can I do that in a VM easily?

mtrmac:
> As an alternative, would it be easy enough to run the libpod / buildah test suites against this PR’s branch of containers/image? Is there a test harness that can do that, or can I do that in a VM easily?

I was being stupid, of course I can just take advantage of the GitHub CI integration:

- libpod: https://github.com/projectatomic/libpod/pull/902
- CRI-O: https://github.com/kubernetes-incubator/cri-o/pull/1593
- buildah: https://github.com/projectatomic/buildah/pull/752

All of them are green, so I’ll merge this as is.

mtrmac:
👍

Issue #397, opened 2018-01-12
support registry mirrors


vrothberg:
Currently, we are working on adding support for private-registry mirrors to moby/docker (see https://github.com/moby/moby/pull/34319). The core semantics are that users can specify mirrors for a given registry, which will be selected as endpoints prior to the registry when pulling an image. Prior to our change, mirroring worked only for the docker hub.

The ability to mirror registries (i.e., pull-through caches) has various use-cases and has been requested to moby/docker upstream for a long while. It is likely that once https://github.com/moby/moby/pull/34319 is merged, to face an increasing demand for a mirror feature beyond moby, but for tools using the `containers` libraries (e.g., skopeo, podmon, cri-o). _Hence, I want to kick off a discussion and ask for feedback and ideas if there is interest to support registry mirrors in `containers/image` and how it should be implemented._

One possibility could be to extend https://github.com/containers/image/blob/master/pkg/sysregistries/system_registries.go#L24 with a mapping `registry -> [mirrors]` in the tomlconfig that can be used during endpoint/refSrc lookup.

Cc: @mtrmac
Cc: @cyphar 
Cc: @flavio

mtrmac:
It does sound interesting, and `pkg/sysregistries` does look like the natural place to keep the configuration.

(It is a bit annoying that we anticipate having two separate configuration locations from the outset, but, oh well, it’s only like the millionth time :) )

vrothberg:
@rhatdan @mtrmac Thanks for your comments!

Currently, there is a clear separation between secure and insecure registries in `pkg/sysregistries`. While the most straight-forward way of supporting mirrors might be adding another `[mirrors]` table, I am not sure if that's the most elegant one.
I see some beauty the way https://github.com/moby/moby/pull/34319 configures Docker's `daemon.json`. Now, there's an array of `Registry` structs, where the `URL`'s scheme determines if it's a secure (`https`) registry or an insecure (`http`) one. Each registry has an optional array of mirrors, with the same URL semantics as registries.

I wonder if we can achieve something similar for `pkg/sysregistries` as exemplified in the snippet below. As far as I can see, such a change would still be backwards compatible. What's your take on this?

```toml
[[registry]]
  url = "http://insecure.registry.com"

[[registry]]
  url = "https://blocked.registry.com"
  blocked = true

[[registry]]
  url = "https://secure.registry.com"

  [[registry.mirror]]
    url = "https://mirror1.secure.registry.cm"

  [[registry.mirror]]
[...]

PR #429, opened 2018-03-19
introduce sysregistriesv2

vrothberg:
Introduce sysregistriesv2, which changes the format of the
`registries.conf` TOML configuration.  Instead of having different lists
to specify search registries, blocked and insecure ones, all data is
encapsulated into one registry type.  The registry type allows to
specify a list of mirrors, which can be used in the endpoint lookup to
serve, for instance, as pull through caches for the associated registry.

An example configuration may look as follows:

```toml
[[registry]]
url = "https://registry.com"
prefix = "another-registry.com"

[[registry.mirror]]
url = "http://registry-mirror.com"
insecure = true
```

The upper example shows the configuration for `https://registry.com`.
The prefix is used for matching images, and to translate one namespace
to another.  If `prefix="example.com/bar"`, `url="https://example.com/foo/bar"`
and we pull from `example.com/bar/myimage:latest`, the image will
effectively be pulled from `example.com/foo/bar/myimage:latest`. If no
prefix is specified, it defaults to the specified URL.

Ease migration from sysregistries v1 to v2 by also loading
configurations in the v1 TOML format.  Throw an error in case a config
tries to mix both formats.  This allows a smoother migration for
developers, maintainers and the user, who can switch to the new config
format once all tools have been updated to v2.

Signed-off-by: Valentin Rothberg <vrothberg@suse.com>
Closes: https://github.com/containers/image/issues/397

### DIFF
--- a/pkg/sysregistriesv2/system_registries_v2.go
+++ b/pkg/sysregistriesv2/system_registries_v2.go
@@ -1,0 +1,322 @@
+package sysregistriesv2
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net/url"
+	"path/filepath"
+	"strings"
+
+	"github.com/BurntSushi/toml"
+	"github.com/containers/image/types"
+)
+
+// systemRegistriesConfPath is the path to the system-wide registry
+// configuration file and is used to add/subtract potential registries for
+// obtaining images.  You can override this at build time with
+// -ldflags '-X github.com/containers/image/sysregistries.systemRegistriesConfPath=$your_path'
+var systemRegistriesConfPath = builtinRegistriesConfPath
+
+// builtinRegistriesConfPath is the path to the registry configuration file.
+// DO NOT change this, instead see systemRegistriesConfPath above.
+const builtinRegistriesConfPath = "/etc/containers/registries.conf"
+
+// tomlURL is an abstraction required to unmarshal the non-primitive and
+// non-local url.URL type when loading the toml config.
+type tomlURL struct {
+	url url.URL
+}
+
+// UnmashalText interprets and parses text as a net.URL type and assigns it to
+// the tomlURL's url field.
+func (r *tomlURL) UnmarshalText(text []byte) (err error) {
+	r.url, err = parseURL(string(text))
+	return err
+}
+
+// mirror is an internal type including all mirror data but the URL.
+type mirror struct {
+	// If true, certs verification will be skipped and HTTP (non-TLS)
+	// connections will be allowed.
+	Insecure bool `toml:"insecure"`
+}
+
+// Mirror represents a mirror. Mirrors can be used as pull-through caches for
+// registries.
+type Mirror struct {
+	// The mirror's URL.
+	URL url.URL
+	mirror
+}
+
+// tomlMirror is a serializable Mirror.
+type tomlMirror struct {
+	// Serializable mirror URL.
+	URL tomlURL `toml:"url"`
+	mirror
+}
+
+// toMirror transforms tmirror to a Mirror.
+func (tmir *tomlMirror) toMirror() (Mirror, error) {
+	if len(tmir.URL.url.String()) == 0 {
+		return Mirror{}, fmt.Errorf("mirror must include a URL")
+	}
+	mir := Mirror{URL: tmir.URL.url, mirror: tmir.mirror}
+	return mir, nil
+}
+
+// registry is an internal type including all registry data but the URL and the
+// array of associated mirrors.
+type registry struct {
+	// If true, pulling from the registry will be blocked.
+	Blocked bool `toml:"blocked"`
+	// If true, certs verification will be skipped and HTTP (non-TLS)
+	// connections will be allowed.
+	Insecure bool `toml:"insecure"`
+	// If true, the registry can be used when pulling an unqualified image.
+	Search bool `toml:"unqualified-search"`
+	// Prefix is used for matching images, and to translate one namespace to
+	// another.  If `Prefix="example.com/bar"`, `URL="https://example.com/foo/bar"`
+	// and we pull from "example.com/bar/myimage:latest", the image will
+	// effectively be pulled from https://example.com/foo/bar/myimage:latest.
+	// If no Prefix is specified, it defaults to the specified URL.
+	Prefix string `toml:"prefix"`
+}
+
+// Registry represents a registry.
+type Registry struct {
+	// Serializable registry URL.
+	URL url.URL
+	// The registry's mirrors.
+	Mirrors []Mirror
+	registry
+}
+
+// tomlRegistry is serializable Registry.
+type tomlRegistry struct {
+	URL     tomlURL      `toml:"url"`
+	Mirrors []tomlMirror `toml:"mirror"`
+	registry
+}
+
+// stripURIScheme strips the URI scheme from the given URL and returns it as
+// as string.
+func stripURIScheme(url url.URL) string {
+	return strings.TrimPrefix(url.String(), url.Scheme+"://")
+}
+
+// toRegistry transforms treg to a Registry.
+func (treg *tomlRegistry) toRegistry() (Registry, error) {
+	if len(treg.URL.url.String()) == 0 {
+		return Registry{}, fmt.Errorf("registry must include a URL")
+	}
+	reg := Registry{URL: treg.URL.url, registry: treg.registry}
+	// if no prefix is specified, default to the specified URL with
+	// stripped URI scheme
+	if reg.Prefix == "" {
+		reg.Prefix = stripURIScheme(reg.URL)
+	}
+
+	for _, tmir := range treg.Mirrors {
+		mir, err := tmir.toMirror()
+		if err != nil {
+			return Registry{}, err
+		}
+		reg.Mirrors = append(reg.Mirrors, mir)
+	}
+	return reg, nil
+}
+
+// backwards compatability to sysregistries v1
+type v1TOMLregistries struct {
+	Registries []string `toml:"registries"`
+}
+
+// tomlConfig is the data type used to unmarshal the toml config.
+type tomlConfig struct {
+	TOMLRegistries []tomlRegistry `toml:"registry"`
+	// backwards compatability to sysregistries v1
+	V1Registries struct {
+		Search   v1TOMLregistries `toml:"search"`
+		Insecure v1TOMLregistries `toml:"insecure"`
+		Block    v1TOMLregistries `toml:"block"`
+	} `toml:"registries"`
+}
+
+// parseURL parses the input string, performs some sanity checks and returns
+// a url.URL.  The input must be a valid URI with an "http" or "https" scheme,
+// a specified host and an empty URI user.  Otherwise, an error is returned.
+func parseURL(input string) (url.URL, error) {
+	input = strings.TrimRight(input, "/")
+
+	uri, err := url.Parse(input)
+	if err != nil {
+		return url.URL{}, fmt.Errorf("error parsing URL %s: %v", input, err)
+	}
+
+	// only https and http are valid URI schemes
+	if uri.Scheme == "" {
+		return url.URL{}, fmt.Errorf("unspecified URI scheme: %s", input)
+	}
+	if uri.Scheme != "https" && uri.Scheme != "http" {
+		return url.URL{}, fmt.Errorf("unsupported URI scheme: %s", input)
+	}
+
+	// a host must be specified
+	if uri.Host == "" {
+		return url.URL{}, fmt.Errorf("unspecified URI host: %s", input)
+	}
+
+	// user must be empty
+	if uri.User != nil {
+		// strip password for security reasons
+		uri.User = url.UserPassword(uri.User.Username(), "xxxxxx")
+		return url.URL{}, fmt.Errorf("unsupported username/password: %q", uri)
+	}
+
+	return *uri, nil
+}
+
+// getV1Registries transforms v1 registries in the config into an array of v2
+// registries of type Registry.
+func getV1Registries(config *tomlConfig) ([]Registry, error) {
+	regMap := make(map[string]*Registry)
+
+	getRegistry := func(s string) (*Registry, error) { // Note: _pointer_ to a long-lived object
+		url, err := parseURL(s)
+		if err != nil {
+			return nil, err
+		}
+		prefix := stripURIScheme(url)
+		reg, exists := regMap[prefix]
+		if !exists {
+			reg = &Registry{URL: url,
+				Mirrors:  []Mirror{},
+				registry: registry{Prefix: prefix}}
+			regMap[prefix] = reg
+		}
+		return reg, nil
+	}
+
+	for _, search := range config.V1Registries.Search.Registries {
+		reg, err := getRegistry(search)
+		if err != nil {
+			return nil, err
+		}
+		reg.Search = true
+	}
+	for _, blocked := range config.V1Registries.Block.Registries {
+		reg, err := getRegistry(blocked)
+		if err != nil {
+			return nil, err
+		}
+		reg.Blocked = true
+	}
+	for _, insecure := range config.V1Registries.Insecure.Registries {
+		reg, err := getRegistry(insecure)
+		if err != nil {
+			return nil, err
+		}
+		reg.Insecure = true
+	}
+
+	registries := []Registry{}
+	for _, reg := range regMap {
+		registries = append(registries, *reg)
+	}
+	return registries, nil
+}
+
+// GetRegistries loads and returns the registries specified in the config.
+func GetRegistries(ctx *types.SystemContext) ([]Registry, error) {
+	config, err := loadRegistryConf(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	registries := []Registry{}
+	for _, treg := range config.TOMLRegistries {
+		reg, err := treg.toRegistry()
+		if err != nil {
+			return nil, err
+		}
+		registries = append(registries, reg)
+	}
+
+	// backwards compatibility for v1 configs
+	v1Registries, err := getV1Registries(config)
+	if err != nil {
+		return nil, err
+	}
+	if len(v1Registries) > 0 {
+		if len(registries) > 0 {
+			return nil, fmt.Errorf("mixing sysregistry v1/v2 is not supported")
+		}
+		registries = v1Registries
+	}
+
+	return registries, nil
+}
+
+// FindUnqualifiedSearchRegistries returns all registries that are configured
+// for unqualified image search (i.e., with Registry.Search == true).
+func FindUnqualifiedSearchRegistries(registries []Registry) []Registry {
+	unqualified := []Registry{}
+	for _, reg := range registries {
+		if reg.Search {
+			unqualified = append(unqualified, reg)
+		}
+	}
+	return unqualified
+}
+
+// FindRegistry returns the Registry with the longest prefix for ref.  If no
+// Registry prefixes the image, nil is returned.
+func FindRegistry(ref string, registries []Registry) *Registry {
+	reg := Registry{}
+	prefixLen := 0
+	for _, r := range registries {
+		if strings.HasPrefix(ref, r.Prefix) {
+			length := len(r.Prefix)
+			if length > prefixLen {
+				reg = r
+				prefixLen = length
+			}
+		}
+	}
+	if prefixLen != 0 {
+		return &reg
+	}
+	return nil
+}
+
+// Reads the global registry file from the filesystem. Returns a byte array.
+func readRegistryConf(ctx *types.SystemContext) ([]byte, error) {
+	dirPath := systemRegistriesConfPath
+	if ctx != nil {
+		if ctx.SystemRegistriesConfPath != "" {
+			dirPath = ctx.SystemRegistriesConfPath
+		} else if ctx.RootForImplicitAbsolutePaths != "" {
+			dirPath = filepath.Join(ctx.RootForImplicitAbsolutePaths, systemRegistriesConfPath)
+		}
+	}
+	configBytes, err := ioutil.ReadFile(dirPath)
+	return configBytes, err
+}
+
+// Used in unittests to parse custom configs without a types.SystemContext.
+var readConf = readRegistryConf
+
+// Loads the registry configuration file from the filesystem and then unmarshals
+// it.  Returns the unmarshalled object.
+func loadRegistryConf(ctx *types.SystemContext) (*tomlConfig, error) {
+	config := &tomlConfig{}
+
+	configBytes, err := readConf(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	err = toml.Unmarshal(configBytes, &config)
+	return config, err
+}
--- a/pkg/sysregistriesv2/system_registries_v2_test.go
+++ b/pkg/sysregistriesv2/system_registries_v2_test.go
@@ -1,0 +1,292 @@
+package sysregistriesv2
+
+import (
+	"github.com/containers/image/types"
+	"github.com/stretchr/testify/assert"
+	"net/url"
+	"testing"
+)
+
+var testConfig = []byte("")
+
+func init() {
+	readConf = func(_ *types.SystemContext) ([]byte, error) {
+		return testConfig, nil
+	}
+}
+
+func TestParseURL(t *testing.T) {
+	var err error
+	var url url.URL
+
+	// invalid URLs
+	_, err = parseURL("unspecified.scheme")
+	assert.NotNil(t, err)
+	assert.Contains(t, err.Error(), "unspecified URI scheme:")
+
+	_, err = parseURL("httpx://unsupported.scheme")
+	assert.NotNil(t, err)
+	assert.Contains(t, err.Error(), "unsupported URI scheme:")
+
+	_, err = parseURL("http://")
+	assert.NotNil(t, err)
+	assert.Contains(t, err.Error(), "unspecified URI host:")
+
+	_, err = parseURL("https://user:password@unsupported.com")
+	assert.NotNil(t, err)
+	assert.Contains(t, err.Error(), "unsupported username/password:")
+
+	// valid URLs
+	url, err = parseURL("http://example.com")
+	assert.Nil(t, err)
+	assert.Equal(t, "http://example.com", url.String())
+
+	url, err = parseURL("http://example.com/")
+	assert.Nil(t, err)
+	assert.Equal(t, "http://example.com", url.String())
+
+	url, err = parseURL("http://example.com//////")
+	assert.Nil(t, err)
+	assert.Equal(t, "http://example.com", url.String())
+
+	url, err = parseURL("http://example.com:5000/with/path")
+	assert.Nil(t, err)
+	assert.Equal(t, "http://example.com:5000/with/path", url.String())
+}
+
+func TestEmptyConfig(t *testing.T) {
+	testConfig = []byte(``)
+
+	registries, err := GetRegistries(nil)
+	assert.Nil(t, err)
+	assert.Equal(t, 0, len(registries))
+}
+
+func TestMirrors(t *testing.T) {
+	testConfig = []byte(`
+[[registry]]
+url = "https://registry.com"
+
+[[registry.mirror]]
+url = "https://mirror-1.registry.com"
+
+[[registry.mirror]]
+url = "https://mirror-2.registry.com"
+insecure = true
+
+[[registry]]
+url = "https://blocked.registry.com"
+blocked = true`)
+
+	registries, err := GetRegistries(nil)
+	assert.Nil(t, err)
+	assert.Equal(t, 2, len(registries))
+
+	var reg *Registry
+	reg = FindRegistry("registry.com/image:tag", registries)
+	assert.NotNil(t, reg)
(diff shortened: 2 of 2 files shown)


vrothberg:
I had a close look at how we can proceed. With the current implementation of the `Registry` type (see https://github.com/vrothberg/image/blob/b1a645d862102aa1fc1933a9b9ecbd01608dd90b/pkg/sysregistries/system_registries.go#L30) we can achieve full backward compat, which has a price as we're stuck with old API that is based on arrays of strings.

I see three options:

1. Implement backward compat. The old types can easily be mapped to the new ones, but we're stuck with the functions returning `[]string`. How could we come up with new functions to make use of the new types?
2. Implement a `sysregistriesV2` package (or something similar) to have a clear separation. This would entail many benefits, such as implementing sanity checks, using `net.URL` for parsing registries, etc.
3. Do *not* change the config as suggested above, but only a new field `registries.mirrors`, which is a map for `registry->[mirrors]`.

I am in favor of 2), as we add sanity checks, parsing of URL's, change from string to `net.URL` etc, but I am very curious to hear more opinions.

vrothberg:
Ping @mtrmac, @baude, @rhatdan.

rhatdan:
I am for you option 2, gives us a way forward, as long as we don't have to update all packages at the same time.

baude:
yeah #2 is fine to me as well.

vrothberg:
@rhatdan @baude The latest commit is a concrete suggestion. I did not add any unit tests yet as I want to wait for your opinions and for an agreement on the design.

The `Registry` type includes all data to support the old semantics (search, block, insecure) and it also implements prefix semantics (similar to the open Moby PR). I implemented a `FindRegistries` function, which demonstrates how the new API *could* be used:
``` golang
// FindRegistries returns all registries matching the specified image reference.
// In case of an unqualified image reference, all registries that are configured
// for unqualified image search are returned.  In case of a fully qualified
// image reference, the registry with the longest prefix is returned.  An empty
// array is returned in case no matching reference is found.
func FindRegistries(imgRef reference.Reference, registries []Registry) ([]Registry, error)
```

baude:
I like the direction you are headed ... good work

mtrmac:
> 2. Implement a `sysregistriesV2` package (or something similar) to have a clear separation. This would entail many benefits, such as implementing sanity checks, using `net.URL` for parsing registries, etc.

Out of the three alternatives, I prefer 2. as well.

But I’d rather have a single implementation, not two that will inevitably drift and differ.  Either provide them from a single package, share the underlying implementation in an `internal` subpackage, or—maybe even preferably—just break the existing API. For better or worse, c/image does not currently have a stable API commitment, and the existing `sysregistries` package only exports two functions, so that should not be an extreme burden on consumers.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -0,0 +1,220 @@
...
+	return err
+}
+
+// Registry is a generalized type representing a registry host and a registry
+// mirror, whereas the mirror cannot further be mirrored.
+type Registry struct {
+	URL      RegURL     `toml:"url"`
+	Blocked  bool       `toml:"blocked"`
+	Insecure bool       `toml:"insecure"`
+	Search   bool       `toml:"unqualified-search"`
```
mtrmac: What does this mean/do? Please add documentation to the field. (Others could benefit from that as well, but for the new ones it’s especially necessary to record the intent and consensus.)

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -0,0 +1,220 @@
...
+	Prefix   string     `toml:"prefix"`
```
mtrmac: What does this mean/do? Please add documentation to the field.

(And about `url = "foo"; prefix = "bar"`? Is that allowed? Reasonable?)

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -0,0 +1,220 @@
...
+	return unqualified
+}
+
+// FindRegistries returns all registries matching the specified image reference.
+// In case of an unqualified image reference, all registries that are configured
+// for unqualified image search are returned.  In case of a fully qualified
+// image reference, the registry with the longest prefix is returned.  An empty
+// array is returned in case no matching registry is found.
+func FindRegistries(imgRef reference.Reference, registries []Registry) ([]Registry, error) {
+	domain := reference.Domain(imgRef.(reference.Named))
```
mtrmac: If the function crashes with a non-`reference.Named`, why not make the input parameter a `reference.Named` in the first place?

---

I’m not sure this is going to actually work, though; `c/image/docker/reference.Reference` is not the Fedora `/usr/bin/docker`-like `docker/docker/reference` package which implements unqualified names as a first-class concept, its the upstream-like one where a reference always implicitly includes a domain. `ParseNormalizedNamed` will add a `docker.io` host name; `ParseNamed` will refuse an unqualified name.

`reference.Parse` does make it possible to create a `domain`-less value, but that value would be dangerous to pass to any other user in c/image which expects normalization to happen. And it incorrectly parses `notlocalhost/busybox` as domain=`notlocalhost`, path=`busybox`, instead of domain=unknown, path=`notlocalhost/busybox`.

[The search semantics is _really problematic_ for security: with it, and a `pull fedora` command, the user never knows what image will be downloaded.]

If this needs to exist at all (which, sadly, seems to be the case), the `imgRef` should probably just be a `string`; `reference.Named` would only be created by the result of combining the domain and the input string.

---

And maybe the public API should not combine the unqualified/qualified cases at all; AFAICT the caller needs to treat them differently anyway (for unqualified, combine the search registry and the input; for qualified, use the input as specified, and only use the flags), so the caller can just as well call two different public API endpoints here (`GetUnqualifiedSearchRegistries`/`GetRegistryForReference` or something like that); then, the qualified case could return a `*Registry` instead of an array, and the caller would not need to worry about what happens when there is >1 element.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -0,0 +1,220 @@
...
+		if strings.HasPrefix(imgRef.String(), reg.Prefix) {
+			length := len(reg.Prefix)
+			if length > prefixLen {
+				reg = r
+				prefixLen = length
+			}
+		}
+	}
+	if prefixLen != 0 {
+		regs = append(regs, reg)
```
mtrmac: (Non-blocking: The `regs` variable seems ultimately unnecessary; it has either zero or one member, so a `return` here would do just as well.)

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -0,0 +1,220 @@
...
+// checkRegistries performs several sanity checks on the specified registries.
+// An error is returned in case a mirror is further mirrored.  Warnings are
+// printed if the a mirror's security level doesn't match the one of the
+// corresponding registry, and if a mirror has a specified Prefix, which has no
+// effect.
+func checkRegistries(registries []Registry) error {
+	for _, reg := range registries {
+		// default to the URL if Prefix isn't specified
+		if reg.Prefix == "" {
+			reg.Prefix = reg.String()
```
mtrmac: If this function modifies its input, that needs to be clear from its name and description.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -0,0 +1,220 @@
...
+		}
+		for _, mir := range reg.Mirrors {
+			if len(mir.Mirrors) > 0 {
+				return fmt.Errorf("mirror '%s' of registry '%s' cannot be mirrored", mir.String(), reg.String())
+			}
+			if mir.Prefix != "" {
+				logrus.Warnf("specified prefix for mirror '%s' has no effect", mir.String())
+			}
+			if mir.Insecure != reg.Insecure {
```
mtrmac: Is this really undesirable? A private mirror on a private network is exactly the kind of thing many (sure, irresponsibly too many) users run with self-signed certificates.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -0,0 +1,220 @@
...
+// RegURL represents a Registry's URl.  This abstraction is required to
+// unmarshal the non-primitive and non-local url.URL type when loading the
+// toml config.
+type RegURL struct {
+	URL url.URL
+}
+
+// String returns host:port/path of the RegURL's URL.
+func (r *RegURL) String() string {
+	return r.URL.Host + r.URL.Path
```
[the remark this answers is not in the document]
mtrmac: (Do the `String` methods even need to exist? It may be better to force the caller to think about formatting than being surprising about it.)

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -0,0 +1,220 @@
...
+		// default to the URL if Prefix isn't specified
+		if reg.Prefix == "" {
+			reg.Prefix = reg.String()
+		}
+		for _, mir := range reg.Mirrors {
+			if len(mir.Mirrors) > 0 {
+				return fmt.Errorf("mirror '%s' of registry '%s' cannot be mirrored", mir.String(), reg.String())
+			}
+			if mir.Prefix != "" {
+				logrus.Warnf("specified prefix for mirror '%s' has no effect", mir.String())
```
mtrmac: It would be cleaner to add a separate `Mirror` type than to try to handle these inconsistencies in behavior after the fact. It really isn’t a recursive structure, so let’s not declare it like one.

Already this consistency check is costing about as many lines as reusing the `Registry` type has saved.

vrothberg:
@mtrmac 

> Does this need to be coordinated with the moby/moby PR introducing similar functionality? Should this me merged independently, or only after the two arrive on some shared consensus?
> Is there any chance at all of sharing the file between the two? Would we rely on the atomic-registries conversion? Or not even that?

It would be ideal to go in the same direction and share the configuration at best, but I do not see any realistic chance. The moby/moby PR is blocked by upstream. There have been multiple attempts over the past years to implement registry mirroring but all have been silenced with blinding grenades. The current one came far, but I am losing hope it will be merged. The current status is to wait until upstream has an agreement on what they actually want. Hence, I suggest paving the way forward here. We use mirroring already in production (for Docker) and backport the upstream PR. As we're working on offering crio as a tech preview, having the necessary pieces upstream here would be perfect.

---

>The API change is not the big worry, we can deal with that. This completely changes the format of the configuration file, and affects users. Is everyone OK with that? @rhatdan @baude 
> Or do we need to still accept the old input and transparently convert?

It's just my 2 cents but I would not accept the old input anymore.  The old API is based entirely on strings, so a clean cut and clean separation of v1 and v2 of the API may just avoid confusion.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -0,0 +1,220 @@
...
+// -ldflags '-X github.com/containers/image/sysregistries.systemRegistriesConfPath=$your_path'
+var systemRegistriesConfPath = builtinRegistriesConfPath
+
+// builtinRegistriesConfPath is the path to registry configuration file
+// DO NOT change this, instead see systemRegistriesConfPath above.
+const builtinRegistriesConfPath = "/etc/containers/registries.conf"
+
+// RegURL represents a Registry's URl.  This abstraction is required to
+// unmarshal the non-primitive and non-local url.URL type when loading the
+// toml config.
```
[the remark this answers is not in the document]
vrothberg: I will look into this.

pkg/sysregistriesv2/system_registries_v2.go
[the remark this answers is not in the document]
vrothberg: It's currently used to match an image reference against registries, but this doesn't need to be public at all.

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: In v1 of the API, search registries are used to allow unqualified image pulling. When pulling an unqualified image, the image will effectively be appended to all specified search registries. The first hit wins. The `Search` field does exactly that. If a registry has this switch set to true, then the registry (and it's mirrors) will be used to pull unqualified images.

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: I will add documentation. Sorry, in case that's confusing. Some discussion happened in IRC.

This is something similar to the moby/moby PR. The Prefix effectively allows to translate one namespace into another. So `url = "foo"; prefix = "bar";` would be allowed.

Notice that specifying a prefix is *not* mandatory.

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: I don't think it's undesirable. I have such a thing running 24/7 :rofl: But to avoid accidents I think that it's worth throwing a warning.

pkg/sysregistriesv2/system_registries_v2.go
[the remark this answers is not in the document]
vrothberg: done

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: :+1:

vrothberg:
@mtrmac I just updated the PR, with the following changes:

- Refined the types: add `Mirror`, changed to generalized `tomlURL` for Mirror and Registry.
- Added documentation to all functions, methods, types, and fields.
- Added `GetRegistries` function to load registries. If no prefix is specified, `Registry.Prefix` will default to the specified URL without the URI scheme. This is required when searching for matching registries based on an image reference.
- The upper change implies that there's no String() methods anymore.
- Splitted `FindRegistries` into `FindUnqualifiedSearchRegistries` and `FindRegistry`.

That's basically it. One thing I am not very sure about is the `Registry.Insecure` field. May we just rename it to `SkipVerify`?

mtrmac:
> >The API change is not the big worry, we can deal with that. This completely changes the format of the configuration file, and affects users. Is everyone OK with that? @rhatdan @baude 
> > Or do we need to still accept the old input and transparently convert?
>
> It's just my 2 cents but I would not accept the old input anymore.  The old API is based entirely on strings, so a clean cut and clean separation of v1 and v2 of the API may just avoid confusion.

Again, I don’t care so much about the **API**; that affects maybe 10 people maintaining the Golang programs that import this package. I care about the **file format** as an **end-user interface**; the worst case here is ignoring, or rejecting, configuration files managed by someone’s config management system, and bringing a few (thousand?) 100-node clusters down. The file format and the API are not the same thing.

@rhatdan @baude This is probably up to you (or does someone else own the `registries.conf` file?)

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -0,0 +1,199 @@
...
+// non-local url.URL type when loading the toml config.
+type tomlURL struct {
+	url url.URL
+}
+
+// Mirror represents a mirror. Mirrors can be used as pull-through caches for
+// registries.
+type Mirror struct {
+	// The mirror's URL
+	URL tomlURL `toml:"url"`
```
mtrmac: This use of `tomlURL` is still visible to consumers (in that they have to say `Mirror.URL.url` - and is the private `url` field even visible?

I was thinking something like
```go
// All existing docstrings [this is the API provided to Golang callers]
type Mirror struct {
    URL url.URL
    Insecure bool
}

// [this is the raw format of the config file]
type tomlMirror struct {
    URL tomlURL
    Insecure bool
}
```
unmarshal into `tomlMirror`, and then initialize a `Mirror` from it. (Maybe the `toml` package supports a nicer way to achieve the same thing.)

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -0,0 +1,199 @@
...
+	Insecure bool `toml:"insecure"`
+}
+
+// Registry represents a registry.
+type Registry struct {
+	// The registry's URL
+	URL tomlURL `toml:"url"`
+	// If true, pulling from the registry will be blocked
+	Blocked bool `toml:"blocked"`
+	// If true, certs verification will be skipped
```
mtrmac: … and HTTP (non-TLS) connections will be allowed.

(for both `Insecure` members.)

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -0,0 +1,199 @@
...
+	Insecure bool `toml:"insecure"`
+	// If true, the registry can be used when pulling an unqualified image
+	Search bool `toml:"unqualified-search"`
+	// If it prefixes an image, the registry will be used for pulling
+	// (defaults to the URL without the URI scheme)
+	Prefix string `toml:"prefix"`
```
mtrmac: Is the `Prefix` used only for matching, or stripped? E.g. with `URL`=`https://example.com/foo` and `Prefix`=`example.net/bar`, is `example.net/bar/baz` downloaded from `https://example.com/foo/bar/baz` or from `https://example.com/foo/baz` ?

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -0,0 +1,199 @@
...
+// GetRegistries loads and returns the registries specified in the config.
+func GetRegistries(ctx *types.SystemContext) ([]Registry, error) {
+	config, err := loadRegistryConf(ctx)
+	if err != nil {
+		return []Registry{}, err
+	}
+	// if no prefix is specified default to specified URL without the URI scheme
+	for i, reg := range config.Registries {
+		if reg.Prefix == "" {
+			prefix := strings.TrimPrefix(reg.URL.url.String(), reg.URL.url.Scheme)
```
mtrmac: Doesn’t this strip only `http` from `http://example.com`, leaving `prefix` set to `://example.com`?

mtrmac:
>  One thing I am not very sure about is the `Registry.Insecure` field. May we just rename it to `SkipVerify`?

- It does not just skip HTTPS verification, it also allows use of unencrypted/unauthenticated HTTP.
- There’s some value in making the effect (”this is insecure! you probably shouldn’t be doing this without a good reason!”) very visible.

Still, the “insecure” name is not perfect, e.g. it’s pretty non-descriptive in what it does. I’m not sure that there is any other short and descriptive alternative, and the “insecure” naming is at least somewhat traditional for distribution/docker registries by now.

I don’t feel strongly about this, up to @rhatdan / @baude again.

vrothberg:
Ping @rhatdan, @baude : Could you share your thoughts on the upper comments?

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: I will look into this.

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: The `Prefix` is used only for matching. If we have `URL=https://example.com/foo` and `Prefix=example.net/bar`, then it depends on what the user of the API decides to do. I would suggest downloading `baz:latest` from `https://example.com/foo/baz:latest`. If we want to have the other mapping, we are free to configure `URL=https://example.com/foo/bar` and `Prefix=example.net/bar`.

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: Good catch, I will check this. Once we have an agreement on the design, I will make sure to add unit tests.

pkg/sysregistriesv2/system_registries_v2.go
mtrmac: > then it depends on what the user of the API decides to do

I’d much prefer having it explicitly defined. Having two different applications on the same machine interpret the same configuration file differently would not do.

Your example shows stripping the prefix, and you’re right that that gives the user more freedom. So, just explicitly document that, please.

vrothberg:
@mtrmac I have updated the PR.

- Split `Mirror` and `Registry` into internal types (transparent to the user), which are then converted to API types.

- The main API to interact with is `GetRegistries`, `FindUnqualifiedSearchRegistries` and `FindRegistry`. I'd like to discuss if the latter two should return an error in case no registries are found. I think we could add a `NoRegistryFound` error.

- Added unit tests.

@rhatdan @baude PTAL. To continue, we need to agree on how an upgrade path from v1 to v2 should look like regarding backward compat, etc.

baude:
I'm cool with this.  It's V2 so immediate impact is minimal.  @rhatdan final approval by you?

mtrmac:
> It's V2 so immediate impact is minimal.

That’s a name of the _Go package_, not of the _config file_. This proposes a completely different format of `/etc/containers/registries.conf`, so the two Go packages can’t very well live side by side.

mtrmac:
@baude ^^^

baude:
My understanding is that v2 wouldn't be required until we implement it (i.e. ship a config file with the new format).  That's all.

vrothberg:
I guess that one thing to consider is that multiple tools are using the `/etc/containers/registries.conf` at the moment. If we update this config to the new format, other tools might not be able to read it anymore.

baude:
yeah all the tools must be updated and packaged first.

rhatdan:
So we will have no backwards compatability?  I am fine with the change. But we will need to update lots of packages at once.  Atomic, Skopeo, Buildah, Podman, CRI-O.

vrothberg:
> I’m not entirely sure about forcing the caller to keep a []Registry variable; the Find* functions could load the configuration automatically. Do you expect the calls to be so frequent that caching the []Registry array would be noticeable?

@mtrmac Yes, I expect it to be noticeable. There are quite some syscalls required to load everything, but it depends on the use-case. When being used by a daemon, I would expect to reload the daemon (or send sighup) to reload the config. In case of libpod, a user might do frequent queries etc.

> So we will have no backwards compatability? I am fine with the change. But we will need to update lots of packages at once. Atomic, Skopeo, Buildah, Podman, CRI-O.

@rhatdan I am about to add backwards compatibility. What I suggest (soon updated in the PR) is to be able to also read a `registries.conf` in the v1 format. If a config uses both notations, we can throw an error. This would allow a smoother migration from v1 to v2 with respect to the configuration files. Once all tools on a system are migrated to using v2, the user can use the new TOML format.

vrothberg:
@rhatdan I have just added another commit adding the backwards compatibility to a v1 config. The commit includes more unit tests.

@mtrmac @baude @rhatdan If I see it correctly, the last "open" topic is the signature of the API functions: Should the config get loaded for each call? Should we operate on a `[]Registry` or encapsulate that into another type, e.g. `Config`, which exposes the API functions as methods? I am okay with the current state, but I just want to make sure we discussed it.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -152,18 +170,91 @@ func parseURL(input string) (url.URL, error) {
...
+
+	addRegistry := func(reg Registry) {
+		if up, exists := regs[reg.Prefix]; exists {
+			up.Search = up.Search || reg.Search
+			up.Blocked = up.Blocked || reg.Blocked
+			up.Insecure = up.Insecure || reg.Insecure
+			regs[reg.Prefix] = up
+			return
+		}
+		reg.Mirrors = []Mirror{}
```
mtrmac: This line seems to be a better fit with the rest of the intialization in `makeRegistry`

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -152,18 +170,91 @@ func parseURL(input string) (url.URL, error) {
...
+	makeRegistry := func(s string) (Registry, error) {
+		var reg Registry
+		url, err := parseURL(s)
+		if err != nil {
+			return reg, err
+		}
+		reg.URL = url
+		reg.Prefix = stripURIScheme(url)
+		return reg, nil
+	}
```
mtrmac: This could be simpler, roughly like:

```go
regs := make(map[string]*Registry)
getRegistry := func(s string) (*Registry, error) { // Note: _pointer_ to a long-lived object
    prefix = …
    reg, exists := regs[prefix]
    if !exists {
        reg := &Registry{}
        // initialize reg
        regs[prefix] = reg
    }
   return reg, nil
}
…
for _, search := range config.V1Registries.Search.Registries {
    reg, err := getRegistry(search)
    if err != nil {
        return []Registry{}, err
    }
    reg.Search = true
}
```

notably all of the `||` updates go away, and we don’t need to worry about what parts of `reg` have been discarded in the `exists` case of `addRegistry`.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -152,18 +170,91 @@ func parseURL(input string) (url.URL, error) {
...
+			return
+		}
+		reg.Mirrors = []Mirror{}
+		regs[reg.Prefix] = reg
+	}
+
+	for _, search := range config.V1Registries.Search.Registries {
+		reg, err := makeRegistry(search)
+		if err != nil {
+			return []Registry{}, err
```
mtrmac: (Non-blocking nit: error cases like this could be `return nil, err` throughout.)

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -152,18 +170,91 @@ func parseURL(input string) (url.URL, error) {
...
 	}
+
+	registries := []Registry{}
 	for _, treg := range config.TOMLRegistries {
 		reg := treg.toRegistry()
 		registries = append(registries, reg)
 	}
 
+	// backwards compatibility for v1 configs
+	v2Registries, err := getV1Registries(config)
```
mtrmac: Shouldn’t this be `v1Registries`?

pkg/sysregistriesv2/system_registries_v2_test.go
```diff
@@ -159,3 +159,63 @@ unqualified-search = true`)
 	reg = FindRegistry("registry-c.com/foo:bar", unqRegs)
 	assert.NotNil(t, reg)
 }
+
+func TestV1BackwardsCompatibility(t *testing.T) {
```
mtrmac: Maybe also preserve the original, ~unchanged `system_registries_test.go` as well?

mtrmac:
> @mtrmac Yes, I expect it to be noticeable. … When being used by a daemon, I would expect to reload the daemon (or send sighup) to reload the config. In case of libpod, a user might do frequent queries etc.

All good points, let’s leave the load+filtering separate then.

> Should we operate on a []Registry or encapsulate that into another type, e.g. Config, which exposes the API functions as methods? I am okay with the current state, but I just want to make sure we discussed it.

I don’t feel strongly about this. It would be a tiny bit simpler for the caller to have an opaque “state”, but e.g. dumping the full configuration for debugging would be easier with the full data available. Either is fine with me.

pkg/sysregistriesv2/system_registries_v2_test.go
vrothberg: I extended the v2 ones a bit, which test more cases now, so I don't think we need that.

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: Nice catch :)

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: Done

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: Thanks for your great review. This one made the code _much_ simpler and more readable.

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: Done

vrothberg:
The PR is updated including @mtrmac's latest comments. I stashed the changes into one commit with a proper description.

The CI has some hiccups from skopeo at the moment.

mtrmac:
👍 Thanks! The code LGTM.

> The CI has some hiccups from skopeo at the moment.

Sorry about that, I left the two repos inconsistent overnight. I have restarted the tests now.


@rhatdan / @baude can you ACK the new format, please?

Also, RFC on the transition plan. This introduces a new subpackage, meaning that any consumer can migrate at its leisure, but also nothing is pushing them to migrate. Alternatively we could replace the existing `sysregistries` package, breaking all users and forcing them to accept the new format soon. I don’t have a strong opinion.

vrothberg:
@mtrmac Wonderful, thanks a lot for your great reviews. I really like the outcome. My plan would be to update libpod soon to make use of the new format in order to support mirrors. I suggest to use this as a first user, see how things are going and once that's done and "known to work", we can port other tools. What do you think?

rhatdan:
LGTM

vrothberg:
We need another approval to get this PR merged. Ping @mtrmac @baude.

mtrmac:
@vrothberg Please rebase one last time and I’ll merge this.

vrothberg:
@mtrmac I just rebased on the latest master. Thanks!

mtrmac:
(The test failure is almost certainly unrelated; I may not be able to debug it today I’m afraid.)

rhatdan:
@vrothberg Could you just amend you PR and push again to retrigger tests.

vrothberg:
@rhatdan Done.

vrothberg:
CI seems to be stable now. @mtrmac, can you approve this PR?

rhatdan:
@mtrmac @runcom PTAL

runcom:
LGTM

PR #431, opened 2018-03-20
Put context.Context arguments on almost everything

novas0x2a:
This is a pretty huge change, and it breaks api, so consider this first steps :)

After this change:
 * Network IO paths should react to cancels now.
 * File IO paths generally still won't.
 * `SystemContext` objects have been renamed to `sys` to leave `ctx` available for the stdlib context objects.

### DIFF
--- a/oci/layout/oci_dest.go
+++ b/oci/layout/oci_dest.go
@@ -1,6 +1,7 @@
 package layout
 
 import (
+	"context"
 	"encoding/json"
 	"io"
 	"io/ioutil"
@@ -24,7 +25,7 @@ type ociImageDestination struct {
 }
 
 // newImageDestination returns an ImageDestination for writing to an existing directory.
-func newImageDestination(ctx *types.SystemContext, ref ociReference) (types.ImageDestination, error) {
+func newImageDestination(ctx context.Context, sys *types.SystemContext, ref ociReference) (types.ImageDestination, error) {
 	if ref.image == "" {
 		return nil, errors.Errorf("cannot save image with empty image.ref.name")
 	}
@@ -45,8 +46,8 @@ func newImageDestination(ctx *types.SystemContext, ref ociReference) (types.Imag
 	}
 
 	d := &ociImageDestination{ref: ref, index: *index}
-	if ctx != nil {
-		d.sharedBlobDir = ctx.OCISharedBlobDirPath
+	if sys != nil {
+		d.sharedBlobDir = sys.OCISharedBlobDirPath
 	}
 
 	if err := ensureDirectoryExists(d.ref.dir); err != nil {
@@ -80,7 +81,7 @@ func (d *ociImageDestination) SupportedManifestMIMETypes() []string {
 
 // SupportsSignatures returns an error (to be displayed to the user) if the destination certainly can't store signatures.
 // Note: It is still possible for PutSignatures to fail if SupportsSignatures returns nil.
-func (d *ociImageDestination) SupportsSignatures() error {
+func (d *ociImageDestination) SupportsSignatures(ctx context.Context) error {
 	return errors.Errorf("Pushing signatures for OCI images is not supported")
 }
 
@@ -105,7 +106,7 @@ func (d *ociImageDestination) MustMatchRuntimeOS() bool {
 // WARNING: The contents of stream are being verified on the fly.  Until stream.Read() returns io.EOF, the contents of the data SHOULD NOT be available
 // to any other readers for download using the supplied digest.
 // If stream.Read() at any time, ESPECIALLY at end of input, returns an error, PutBlob MUST 1) fail, and 2) delete any data stored so far.
-func (d *ociImageDestination) PutBlob(stream io.Reader, inputInfo types.BlobInfo, isConfig bool) (types.BlobInfo, error) {
+func (d *ociImageDestination) PutBlob(ctx context.Context, stream io.Reader, inputInfo types.BlobInfo, isConfig bool) (types.BlobInfo, error) {
 	blobFile, err := ioutil.TempFile(d.ref.dir, "oci-put-blob")
 	if err != nil {
 		return types.BlobInfo{}, err
@@ -168,7 +169,7 @@ func (d *ociImageDestination) PutBlob(stream io.Reader, inputInfo types.BlobInfo
 // Unlike PutBlob, the digest can not be empty.  If HasBlob returns true, the size of the blob must also be returned.
 // If the destination does not contain the blob, or it is unknown, HasBlob ordinarily returns (false, -1, nil);
 // it returns a non-nil error only on an unexpected failure.
-func (d *ociImageDestination) HasBlob(info types.BlobInfo) (bool, int64, error) {
+func (d *ociImageDestination) HasBlob(ctx context.Context, info types.BlobInfo) (bool, int64, error) {
 	if info.Digest == "" {
 		return false, -1, errors.Errorf(`"Can not check for a blob with unknown digest`)
 	}
@@ -186,7 +187,7 @@ func (d *ociImageDestination) HasBlob(info types.BlobInfo) (bool, int64, error) 
 	return true, finfo.Size(), nil
 }
 
-func (d *ociImageDestination) ReapplyBlob(info types.BlobInfo) (types.BlobInfo, error) {
+func (d *ociImageDestination) ReapplyBlob(ctx context.Context, info types.BlobInfo) (types.BlobInfo, error) {
 	return info, nil
 }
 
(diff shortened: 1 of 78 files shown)


vendor.conf
```diff
@@ -1,3 +1,5 @@
+github.com/containers/image
```
mtrmac: One-minute look, will review in detail a later:
- ^^ This seems unexpected.
- Overall having a `context.Context` in most places is definitely valuable.
- I’m not sure about passing `types.SystemContext` and `context.Context` by a single parameter, notably because `SystemContext` currently intentionally violates the “Do not store `Context`s inside a struct type” rule. Combining them may still be the right thing to do, I’ll need to take a deeper look.

novas0x2a: Just replying to the first point; the first line of the trash vendor.conf identifies the package root ([e.g.](https://github.com/rancher/trash/blob/master/vendor.conf#L2)); trash gives me a warning without it (presumably because I use a multiple-GOPATH and it can't identify the canonical parent on its own).

mtrmac: Ah, I had no idea that was a thing. Thanks.

mtrmac:
> that's slightly gauche to pass a `nil` to something that takes a `context.Context`, but in this case (because the type system can't catch it) I've made it so passing a `nil` to `GetSystemContext` behaves the way passing a `nil` as a `SystemContext` used to.

That’s not sufficient; if anything actually tries to use the `context.Context` as a `context.Context` (e.g. calling `Done()` or `WithCancel()`, it will crash on `nil`.

oci/layout/oci_dest.go
```diff
@@ -106,7 +108,7 @@ func (d *ociImageDestination) MustMatchRuntimeOS() bool {
 // WARNING: The contents of stream are being verified on the fly.  Until stream.Read() returns io.EOF, the contents of the data SHOULD NOT be available
 // to any other readers for download using the supplied digest.
 // If stream.Read() at any time, ESPECIALLY at end of input, returns an error, PutBlob MUST 1) fail, and 2) delete any data stored so far.
-func (d *ociImageDestination) PutBlob(stream io.Reader, inputInfo types.BlobInfo) (types.BlobInfo, error) {
+func (d *ociImageDestination) PutBlob(ctx context.Context, stream io.Reader, inputInfo types.BlobInfo) (types.BlobInfo, error) {
```
mtrmac: The
```go
	size, err := io.Copy(blobFile, tee)
```
can copy many gigabytes of data, and should be cancelable.

oci/layout/oci_src.go
```diff
@@ -93,7 +94,7 @@ func (s *ociImageSource) GetManifest(instanceDigest *digest.Digest) ([]byte, str
 }
 
 // GetBlob returns a stream for the specified blob, and the blob's size.
-func (s *ociImageSource) GetBlob(info types.BlobInfo) (io.ReadCloser, int64, error) {
+func (s *ociImageSource) GetBlob(ctx context.Context, info types.BlobInfo) (io.ReadCloser, int64, error) {
 	if len(info.URLs) != 0 {
 		return s.getExternalBlob(info.URLs)
```
mtrmac: `getExternalBlob` needs to receive `ctx` and implement cancelation.

ostree/ostree_dest.go
```diff
@@ -123,7 +124,7 @@ func (d *ostreeImageDestination) MustMatchRuntimeOS() bool {
 	return true
 }
 
-func (d *ostreeImageDestination) PutBlob(stream io.Reader, inputInfo types.BlobInfo) (types.BlobInfo, error) {
+func (d *ostreeImageDestination) PutBlob(ctx context.Context, stream io.Reader, inputInfo types.BlobInfo) (types.BlobInfo, error) {
```
mtrmac: The
```go
	size, err := io.Copy(blobFile, tee)
```
can copy many gigabytes of data and should be cancelable.

ostree/ostree_dest.go
```diff
@@ -389,7 +390,7 @@ func (d *ostreeImageDestination) PutSignatures(signatures [][]byte) error {
 	return nil
 }
 
-func (d *ostreeImageDestination) Commit() error {
+func (d *ostreeImageDestination) Commit(ctx context.Context) error {
```
mtrmac: `importBlob`, `fixFiles`, `generateTarSplitMetadata`, and maybe more, can handle many-gigabyte files or directory structures and should be cancelable.

ostree/ostree_src.go
```diff
@@ -256,13 +256,13 @@ func (s *ostreeImageSource) readSingleFile(commit, path string) (io.ReadCloser,
 }
 
 // GetBlob returns a stream for the specified blob, and the blob's size.
-func (s *ostreeImageSource) GetBlob(info types.BlobInfo) (io.ReadCloser, int64, error) {
+func (s *ostreeImageSource) GetBlob(ctx context.Context, info types.BlobInfo) (io.ReadCloser, int64, error) {
```
mtrmac: The goroutine should probably be canceled by `ctx`.

storage/storage_image.go
```diff
@@ -284,7 +284,7 @@ func (s storageImageDestination) ShouldCompressLayers() bool {
 
 // PutBlob stores a layer or data blob in our temporary directory, checking that any information
 // in the blobinfo matches the incoming data.
-func (s *storageImageDestination) PutBlob(stream io.Reader, blobinfo types.BlobInfo) (types.BlobInfo, error) {
+func (s *storageImageDestination) PutBlob(ctx context.Context, stream io.Reader, blobinfo types.BlobInfo) (types.BlobInfo, error) {
```
mtrmac: The
```go
	_, err = io.Copy(diffID.Hash(), decompressed)
```
can copy many gigabytes of data and should be cancelable.

storage/storage_image.go
```diff
@@ -460,7 +460,7 @@ func (s *storageImageDestination) getConfigBlob(info types.BlobInfo) ([]byte, er
 	return nil, errors.New("blob not found")
 }
 
-func (s *storageImageDestination) Commit() error {
+func (s *storageImageDestination) Commit(ctx context.Context) error {
```
mtrmac: The
```go
		layer, _, err := s.imageRef.transport.store.PutLayer(id, lastLayer, nil, "", false, diff)
```
line can copy many gigabytes of data, and should be cancelable.  (I’m not sure that can be done right now.)

tarball/tarball_src.go
```diff
@@ -34,7 +34,7 @@ type tarballImageSource struct {
 	manifest   []byte
 }
 
-func (r *tarballReference) NewImageSource(ctx *types.SystemContext) (types.ImageSource, error) {
+func (r *tarballReference) NewImageSource(ctx context.Context) (types.ImageSource, error) {
```
mtrmac: The
```go
		n, err := io.Copy(ioutil.Discard, reader)
```
can copy many gigabytes of data and should be cancelable.

novas0x2a:
Hm, I'm willing to follow through with changes like splitting SystemContext and Context apart (which would also resolve the nil problem by making it a compile failure), and I can easily use Background instead of TODO in the tests (generally, I use TODO because those tests are the ones that would catch failures to use ctx in the code, but if you prefer Background, that's fine, too).

I'm probably not willing to take on the task of fixing the whole library re: context behavior, although there's clear value in that, my PR was basically me scratching my itch along the code paths I needed, and then propagating those changes as needed :)

Does that work for you?

mtrmac:
I’m fine with not handling the local FS copies; in the worst case I can add the FIXMEs in there myself.

I’d insist on having the two cases of blocking on a channel handled; that should be roughly
```go
select {
case <-ctx.Done():
	return ctx.Err()
case theVariable <- theChannel:
}
```

but I guess I can trade that: if you are willing to split the `SystemContext` and `Context`, I’ll add the FIXMEs and the channel aborts — and the `skopeo` updates needed so that tests pass :). Does that make sense?

---

> (generally, I use TODO because those tests are the ones that would catch failures to use ctx in the code, but if you prefer Background, that's fine, too)

I’m not sure what you mean; looking at the implementation, `TODO` and `Background` are instances of the same object that differ only in their `String()` return value. How does using `TODO` help catch failures to use `ctx` in the code?

novas0x2a:
Yeah, SGTM. It'll probably take me a few days, but I'll get to it.

> I’m not sure what you mean; looking at the implementation, TODO and Background are instances of the same object that differ only in their String() return value. How does using TODO help catch failures to use ctx in the code?

Ah, sorry, I didn't mean the TODOs would, I meant that that they're real TODOs in the sense of indicating the need for future work; the tests they're in would be the things meant to catch failures to use the context. The next step on those TODOs might be to have a test-specific context in the suite that makes sure it cancels the context after each test, so you can't leak threads during the test (for example).

novas0x2a:
I think I've handled all the comments now, except for the file io paths. How's it look?

mtrmac:
@runcom RFC, please at least ACK the general direction and `types.go`.

mtrmac:
:+1: Thanks!

I have prepared a skopeo PR at https://github.com/projectatomic/skopeo/pull/493 .

Note to self: also check a few of the `newImage*` for unnecessary parameters.

runcom:
LGTM (just checked types/types.go, it's awesome!)

runcom:
skopeo's PR passing, so merging here

PR #433, opened 2018-03-29
Add functions to return Registries paths, tools can use them to help users debug

rhatdan:
We are seeing lots of issues with people taking podman and buildah and just
building them and installing them without greating registries.conf file.

They end up doing

buildah from fedora
or
podman from fedora

And we report that fedora does not exist.

I would like to make podman/buildah smarter so that they could state something like
podman from fedora
fedora does not exist.  
/etc/containers/registries.conf and /etc/containers/registries.d directory are missing.

### DIFF
--- a/pkg/sysregistries/system_registries.go
+++ b/pkg/sysregistries/system_registries.go
@@ -42,16 +42,7 @@ func normalizeRegistries(regs *registries) {
 // Reads the global registry file from the filesystem. Returns
 // a byte array
 func readRegistryConf(sys *types.SystemContext) ([]byte, error) {
-	dirPath := systemRegistriesConfPath
-	if sys != nil {
-		if sys.SystemRegistriesConfPath != "" {
-			dirPath = sys.SystemRegistriesConfPath
-		} else if sys.RootForImplicitAbsolutePaths != "" {
-			dirPath = filepath.Join(sys.RootForImplicitAbsolutePaths, systemRegistriesConfPath)
-		}
-	}
-	configBytes, err := ioutil.ReadFile(dirPath)
-	return configBytes, err
+	return ioutil.ReadFile(RegistriesConfPath(sys))
 }
 
 // For mocking in unittests
@@ -96,4 +87,17 @@ func GetInsecureRegistries(sys *types.SystemContext) ([]string, error) {
 		return nil, err
 	}
 	return config.Registries.Insecure.Registries, nil
+}
+
+// RegistriesConfPath is the path to the system-wide registry configuration file
+func RegistriesConfPath(ctx *types.SystemContext) string {
+	path := systemRegistriesConfPath
+	if ctx != nil {
+		if ctx.SystemRegistriesConfPath != "" {
+			path = ctx.SystemRegistriesConfPath
+		} else if ctx.RootForImplicitAbsolutePaths != "" {
+			path = filepath.Join(ctx.RootForImplicitAbsolutePaths, systemRegistriesConfPath)
+		}
+	}
+	return path
 }


TomSweeneyRedHat:
Changes LGTM , but Travis isn't happy with you.

mtrmac:
If either of the two `system`… and `builtin`… variables should be made public, it should be the `system`… ones, so that in builds which use the documented `-ldflags` override `podman`/`buildah` report the path that is actually used.

But I think that `podman`/`buildah` might actually benefit from making `lookaside.go:registriesPath()` (and an analogous new function split from `sysregistries.go:readRegistryConf`), not the variables, public; that way, the returned value always corresponds to overrides done via `SystemContext`, e.g. set up from other CLI options, without having to account for any possible (future?) CLI options of `podman`/`buildah` in the callers.

mtrmac:
(Also `registries.d` is not used in the non-qualified name search code AFAICT; I guess I’m fine with making that path public just for symmetry, as long as we can minimize the number of independently maintained parsers.)

rhatdan:
@mtrmac Changed to just functions calls to return the data.

mtrmac:
@rhatdan The reason why I suggested functions could be better was not to have _functions_ per se (I don’t think that is inherently better), but to correctly and fully take into account all of `types.SystemContext`. See the existing `registriesPath`.

rhatdan:
I don't see registriesPath?

mtrmac:
https://github.com/containers/image/blob/master/docker/lookaside.go#L77 , and equivalently all but the two last lines of https://github.com/containers/image/blob/master/pkg/sysregistries/system_registries.go#L44 .

mtrmac:
> Any idea why this is failing?

That was a weird failure. I have kicked off new test runs, which seem to continue past that point.

Unfortunately skopeo’s CI is currently broken anyway, something to do with `gpg`.

pkg/sysregistries/system_registries.go
```diff
@@ -39,9 +39,9 @@ func normalizeRegistries(regs *registries) {
 	}
 }
 
-// Reads the global registry file from the filesystem. Returns
+// ReadRegistryConf reads the global registry file from the filesystem. Returns
 // a byte array
-func readRegistryConf(sys *types.SystemContext) ([]byte, error) {
+func ReadRegistryConf(sys *types.SystemContext) ([]byte, error) {
```
mtrmac: Why make this public?

pkg/sysregistries/system_registries.go
```diff
@@ -97,3 +97,16 @@ func GetInsecureRegistries(sys *types.SystemContext) ([]string, error) {
 	}
 	return config.Registries.Insecure.Registries, nil
 }
+
+// RegistriesConfPath is the path to the system-wide registry configuration file
+func RegistriesConfPath(ctx *types.SystemContext) string {
```
mtrmac: Please make `readRegistryConf` call this instead of having two copies of the code that will almost certainly get out of sync.

pkg/sysregistries/system_registries.go
```diff
@@ -97,3 +97,16 @@ func GetInsecureRegistries(sys *types.SystemContext) ([]string, error) {
...
+	dirPath := systemRegistriesConfPath
```
mtrmac: (Non-blocking: BTW I don’t know why this was called `dirPath` when it does not point to a directory. No need to qualify it really, `path` would work just as well in this small function. Or, well, `return` and no variable.)

pkg/sysregistries/system_registries.go
rhatdan: That way buildah or podman could give the location of the registries.conf message.

podman pull foobar
foobar image not found in any registry,  you can add registries to /etc/containers/registries.conf

mtrmac: If you mean the path, that’s available through the newly added `RegistriesConfPath` below. This returns the _contents_ of the file, without revealing the path at all.

pkg/sysregistries/system_registries.go
rhatdan: Fixed

pkg/sysregistries/system_registries.go
rhatdan: Fixed

mtrmac:
👍 pending tests. Thanks!

rhatdan:
@runcom PTAL

mtrmac:
@rhatdan Tests seem to progress past the failure point after https://github.com/projectatomic/skopeo/pull/510 and #459, please rebase.

rhatdan:
@mtrmac PTAL

PR #434, opened 2018-04-03
Allow disabling TLS entirely when talking to docker daemons

jlsalmon:
To support communication with docker daemons listening on plain HTTP only, this PR makes it possible to specify the scheme as part of the host URL, e.g. `http+tcp://localhost:2375`

I'm not sure if this is a good idea - I needed a way to stop the httpClient from
trying to use TLS, and this seemed like a reasonable approach, similar to how the
python docker client behaves. Happy to discuss and iterate on this.

Relates to projectatomic/skopeo#244


mtrmac:
I’m not really thrilled about `http+tcp`, it _looks_ entirely redundant and it’s not immediately obvious that the `+tcp` means “do not use HTTPS”.

The `docker/engine-api` behavior at https://github.com/docker/engine-api/blob/master/client/transport/client.go#L38 and https://github.com/docker/go-connections/blob/master/sockets/sockets.go is really surprising (essentially `anything://`, apart from a few exceptions like `unix` and `npipe`, is accepted—only at the `/usr/bin/docker` CLI level is the protocol restricted to a few options like `tcp`), and maybe we don’t want to do exactly the same thing-**especially** not the way it defaults to not using TLS.  

The `docker-py` client does not currently support `http+tcp` either, AFAICS, but it _does_ recognize ordinary `http` and `https`.  Using these two to mean “definitely not TLS” and “definitely TLS” would be instantly understood. (And because we default to TLS already, and the engine-api code does accept `https`, that works as expected right now.)


So, how about using plain `http` instead of `http+tcp`? Is there a downside I am missing?

rhatdan:
I like the idea of http also.

jlsalmon:
Agree completely, `http` makes much more sense. PR updated

mtrmac:
👍  Thanks!

PR #435, opened 2018-04-03
Reformat policy.json.md to be used as a man page

rhatdan:

### DIFF
--- a/docs/policy.json.md
+++ b/docs/policy.json.md
@@ -2,7 +2,10 @@
 % Miloslav Trmač
 % September 2016
 
-# Signature verification policy file format
+# NAME
+policy.json - syntax for the signature verification policy file
+
+## DESCRIPTION
 
 Signature verification policy files are used to specify policy, e.g. trusted keys,
 applicable when deciding whether to accept an image, or individual signatures of that image, as valid.
@@ -10,7 +13,7 @@ applicable when deciding whether to accept an image, or individual signatures of
 The default policy is stored (unless overridden at compile-time) at `/etc/containers/policy.json`;
 applications performing verification may allow using a different policy instead.
 
-## Overall structure
+## FORMAT
 
 The signature verification policy file, usually called `policy.json`,
 uses a JSON format.  Unlike some other JSON files, its parsing is fairly strict:
@@ -271,3 +274,8 @@ selectively allow individual transports and scopes as desired.
     "default": [{"type": "insecureAcceptAnything"}]
 }
 ```
+## SEE ALSO
+  atomic(1)
+
+## HISTORY
+September 2016, Originally compiled by Miloslav Trmač <mitr@redhat.com>


docs/policy.json.5.md
```diff
@@ -2,15 +2,18 @@
 % Miloslav Trmač
 % September 2016
 
-# Signature verification policy file format
+# NAME
+policy.json - Syntax for the Signature Verification Configuration File
```
TomSweeneyRedHat: Should this be "Signature Verification Policy File"?

rhatdan:
I actually think the option in buildah should be hidden.  We don't want to expose this to users.

rhatdan:
@mtrmac @runcom PTAL

docs/policy.json.5.md
```diff
@@ -2,15 +2,18 @@
 % Miloslav Trmač
 % September 2016
 
-# Signature verification policy file format
+# NAME
+policy.json - Syntax for the Signature Verification Policy File
```
mtrmac: `man 7 man-pages` says
> All words in this line (including the word immediately following the "\-") should be in lowercase, except where English or technical terminological convention dictates otherwise.

docs/policy.json.5.md
```diff
@@ -271,3 +274,8 @@ selectively allow individual transports and scopes as desired.
     "default": [{"type": "insecureAcceptAnything"}]
 }
 ```
+# SEE ALSO
+  atomic(1)
+
+# HISTORY
```
mtrmac: Nit: These two should probably be `##`, i.e. the `<h2>` level.

rhatdan: Fixed.

docs/policy.json.5.md
rhatdan: Fixed, I bet we have a lot of man pages that fail that test.

rhatdan:
I copied this file from here into the Fedora Repo to get a man page generated for the package.
We need to work with @lsm5 to see if there is a easy way to automatically pull the man pages from the REPO at build time.  But I don't think that is legal.  We could add this file to the vendord version in skopeo package and skopeo-containers could get the man page from there.

rhatdan:
@mtrmac I renamed it back to the original name, so everything will just work.

mtrmac:
> We could add this file to the vendord version in skopeo package and skopeo-containers could get the man page from there.

I didn’t expect it, but that’s actually fairly easy:
```sh
vndr -whitelist '^github.com/containers/image/docs/.*'
```
(we would actually need to automate this through a `Makefile`, instead of listing commands in `README.md`).

That would be definitely better than a manually-copied file in the source RPM, notably because the man page version would be always in sync with the vendored implementation.

(It would remain awkward for CRI-O, which does not strictly speaking depends on having `skopeo` around, but it would not be any worse than the current situation.)

mtrmac:
👍 pending tests. Thanks!

mtrmac:
> We need to work with @lsm5 to see if there is a easy way to automatically pull the man pages from the REPO at build time

Koji builders are intentionally cut off from the internet.

PR #437, opened 2018-04-06
Modify transports.ListNames() to exclude some transport names

umohnani8:
Add a paramter "transportsNotToAdd" to this function so that we get
a list of transports without the transport names in "transportsNotToAdd".
Useful in cases where we don't support all the transports, for example
buildha push does not support the tarball transport.

### DIFF
--- a/transports/transports.go
+++ b/transports/transports.go
@@ -71,13 +71,19 @@ func ImageName(ref types.ImageReference) string {
 	return ref.Transport().Name() + ":" + ref.StringWithinTransport()
 }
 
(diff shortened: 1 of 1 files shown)


umohnani8:
@mtrmac @runcom PTAL

rhatdan:
Why wouldn't we just trim out the tarball transport in buldah push?

umohnani8:
closed this because something was messed up with my branch. Opened another one https://github.com/containers/image/pull/438

PR #438, opened 2018-04-06
Modify transports.ListNames() to exclude some transport names

umohnani8:
Add a paramter "transportsNotToAdd" to this function so that we get
a list of transports without the transport names in "transportsNotToAdd".
Useful in cases where we don't support all the transports, for example
buildha push does not support the tarball transport.

### DIFF
--- a/transports/transports.go
+++ b/transports/transports.go
@@ -71,13 +71,19 @@ func ImageName(ref types.ImageReference) string {
 	return ref.Transport().Name() + ":" + ref.StringWithinTransport()
 }
 
-// ListNames returns a list of non deprecated transport names.
+// ListNames returns a list of transports except transportsToNotAdd and deprecated transport names.
 // Deprecated transports can be used, but are not presented to users.
-func ListNames() []string {
+func ListNames(transportsToNotAdd []string) []string {
 	kt.mu.Lock()
 	defer kt.mu.Unlock()
 	deprecated := map[string]bool{
 		"atomic": true,
+	}
+	// It is posibble for some transports to not be supported
+	// Add transportsToNotAdd to the deprecated map and
+	// ListNames will return a list of transports without those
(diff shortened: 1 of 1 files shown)


umohnani8:
> Why wouldn't we just trim out the tarball transport in buldah push?

@rhatdan I just thought it would be better to put it in this function for other instances where we might not support all the transports.

mtrmac:
>> Why wouldn't we just trim out the tarball transport in buldah push?
>
> @rhatdan I just thought it would be better to put it in this function for other instances where we might not support all the transports.

Most importantly, that should rarely if ever happen; the whole idea of `ImageReference` is to _abstract_ the transports from the calling applications, so that users can use all kinds of images interchangeably throughout the ecosystem (sure, sometimes, at a performance cost).

(E.g. in the case of `buildah push`, `tarball:` destinations should already fail with ’"tarball:" locations can only be read from, not written to’ without `buildah` needing to understand anything about `tarball:`.  Maybe that fails too late for a good UI, and we need to expand the `ImageReference`/`ImageTransport` interfaces; extending them that way would be fine—we already do something similar with `ImageDestination.SupportsSignatures` to allow an early exit before copying all of the image layers.)

Also, it’s not just `ListNames` in isolation; the `alltransports.ParseImageName` function should have an analogous interface, accepting the same exclusions data structure—but that one is really better done at the application level. `alltransports.parseImageName(…, excludeTarball)` can only fail with a generic ”tarball: transport is not rejected by the caller” or something like that, but the application can give a much better error message (“tarball: is not supported because $good_reason”). So, maybe we could design the excludeTarball data structure to also supply an error message… but at that point it seems to me that both the caller and the callee would be simpler if this were handled purely in the caller.

rhatdan:
Yes I think this is best handled in the caller.

PR #439, opened 2018-04-09
Add more information to error returned for oci

umohnani8:
Add what the syntax of the oci and oci-archive transport should be if an error
is returned when preparing the destination.

### DIFF
--- a/oci/layout/oci_dest.go
+++ b/oci/layout/oci_dest.go
@@ -9,13 +9,12 @@ import (
 	"path/filepath"
 	"runtime"
 
-	"github.com/pkg/errors"
-
 	"github.com/containers/image/manifest"
 	"github.com/containers/image/types"
-	"github.com/opencontainers/go-digest"
+	digest "github.com/opencontainers/go-digest"
 	imgspec "github.com/opencontainers/image-spec/specs-go"
 	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/pkg/errors"
 )
 
 type ociImageDestination struct {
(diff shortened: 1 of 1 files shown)


umohnani8:
@mtrmac PTAL

oci/layout/oci_dest.go
```diff
@@ -26,7 +26,7 @@ type ociImageDestination struct {
 // newImageDestination returns an ImageDestination for writing to an existing directory.
 func newImageDestination(ctx *types.SystemContext, ref ociReference) (types.ImageDestination, error) {
 	if ref.image == "" {
-		return nil, errors.Errorf("cannot save image with empty image.ref.name")
+		return nil, errors.Errorf("cannot save image with empty image.ref.name (must be of form <path>:<reference:tag>)")
```
mtrmac: - This reads as if all of `image.ref.name` should be of form `<path>:<reference:tag>`, when `<path>:` is parsed into a different member.
- The `image.ref.name` annotation, per https://github.com/opencontainers/image-spec/blob/master/annotations.md#pre-defined-annotation-keys , does not have a “tag” concept at all.
- While you are cleaning this up, perhaps the error should say something easier to read/understand than `image.ref.name` (a _subset_ of the official annotation name).  `getManifestDescriptor` just calls it a “reference”, and the spec says “name of the reference” and “reference”.

umohnani8: replaced image.ref.name with reference name and reworded what the syntax should look like. Should we clarify what we mean by reference name? the man page for buildah and podman does a good job at noting what the syntax should be like with examples also.

umohnani8:
Tests are being fixed in https://github.com/projectatomic/skopeo/pull/494

oci/layout/oci_dest.go
mtrmac: > Should we clarify what we mean by reference name? 

If you can think of a concise way, sure.  It’s awkward that both the `image.ref.name` and the input to `ParseReference` are called a “reference”, and I can’t think of a nice way to differentiate in a memorable way (e.g. because they are really the same thing in the `docker://` transport, OTOH this error message may be from a caller which uses `oci/layout.ParseReference`, i.e. the user input may not contain a “transport” at all.)

mtrmac:
👍 

Thanks!

PR #441, opened 2018-04-10
tags is a paginated api, need to follow Link headers

novas0x2a:
This probably isn't actually the best way to do this; in the docker client api, this is basically
```
repo, err := NewRepository(ctx, ref, baseURL, transport)
return repo.Tags().All()
```
But that API can't really use the SystemContext configuration, at least, not as-is.

### DIFF
--- a/docker/docker_image.go
+++ b/docker/docker_image.go
@@ -5,6 +5,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"net/url"
+	"strings"
 
 	"github.com/containers/image/docker/reference"
 	"github.com/containers/image/image"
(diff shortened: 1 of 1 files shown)


novas0x2a:
Hm, I just realized this doesn't work on repos that don't have a :latest tag (`gcr.io/google_containers/kube-dnsmasq-amd64` in this case) because it fetches the ref manifest in order to create the Img Source

docker/docker_image.go
```diff
@@ -39,25 +41,49 @@ func (i *Image) SourceRefFullName() string {
...
+		if res.StatusCode != http.StatusOK {
+			// print url also
+			return nil, errors.Errorf("Invalid status code returned when fetching tags list %d", res.StatusCode)
+		}
+		if err := json.NewDecoder(res.Body).Decode(tagsHolder); err != nil {
+			return nil, err
+		}
+		tags = append(tags, tagsHolder.Tags...)
+
+		if link := res.Header.Get("Link"); link != "" {
```
mtrmac: (Non-blocking nit: This would be a bit easier to read with a `if link == "" { break }`.  The `continue}break}` loop end is unusual.)

docker/docker_image.go
```diff
@@ -39,25 +41,49 @@ func (i *Image) SourceRefFullName() string {
...
-	}
+
+	tags := make([]string, 0)
 	type tagsRes struct {
 		Tags []string
 	}
-	tags := &tagsRes{}
-	if err := json.NewDecoder(res.Body).Decode(tags); err != nil {
-		return nil, err
+	tagsHolder := &tagsRes{}
```
mtrmac: (Non-blocking nit: Please move this immediately before the use of the variable to minimize the amount of context the reader needs to keep in mind.)

novas0x2a:
Yeah, I have a version I'm playing with, i'll update this with it soon when it feels right :)

novas0x2a:
(updated)

docker/docker_image.go
```diff
@@ -39,25 +41,70 @@ func (i *Image) SourceRefFullName() string {
...
-	if res.StatusCode != http.StatusOK {
-		// print url also
-		return nil, errors.Errorf("Invalid status code returned when fetching tags list %d", res.StatusCode)
+	return GetRepositoryTags(ctx, i.src.c.sys, i.src.ref)
+}
+
+// GetRepositoryTags list all tags available in the repository. Note that this
+// has minimal connection with the tag(s) used for this specific
+// ImageReference; the tag will be used to determine policy-based access (see
+// PolicyConfigurationIdentity), but will otherwise be ignored.
```
mtrmac: I’m not sure what “determine policy-based access” refers to; the only use of `.Tag()` invoked through this which I can see is in `lookaside.go`, and yes, that involves `PolicyConfigurationIdentity`, but ultimately the `signatureBase` value is not used for anything in this function.  Am I overlooking something?

(Applies to the shim as well.)

docker/docker_image.go
```diff
@@ -39,25 +41,70 @@ func (i *Image) SourceRefFullName() string {
 	return i.src.ref.ref.Name()
 }
 
-// GetRepositoryTags list all tags available in the repository. Note that this has no connection with the tag(s) used for this specific image, if any.
+// GetRepositoryTags list all tags available in the repository. Note that this
+// has minimal connection with the tag(s) used for this specific
+// ImageReference; the tag will be used to determine policy-based access (see
+// PolicyConfigurationIdentity), but will otherwise be ignored. (This is a
+// backward-compatible shim method which calls the module-level GetRepositoryTags)
 func (i *Image) GetRepositoryTags(ctx context.Context) ([]string, error) {
```
mtrmac: (I wouldn’t mind dropping this entirely, but that would require updating `skopeo` and be more work. Keeping this shim is perfectly fine.)

docker/docker_image.go
novas0x2a: I admit I don't understand the policy stuff much, that comment was an attempt to explain why a tagged reference has to be passed in even though the tag is ignored, perhaps it could be better stated somehow?

rhatdan:
@novas0x2a I think this is in your court to fix up the last couple of nits and then to rebase.

novas0x2a:
Updated and rebased.

mtrmac:
👍 Thanks!

PR #444, opened 2018-04-12
Remove unused dependencies from vendor.conf

### LINKED ISSUE #443: Some of the dependencies in vendor.conf are not needed
It has come to my attention that you have `gorilla/mux` inside the `vendor.conf` but it seems that it's not being used at all in the library.

Probably there are some other dependencies that you are tracking in the vendor file that are not being used either.

agonzalezro:
This commit closes #443, however, some of the dependencies listed in the issue were transitive dependencies that were still needed.

I have check that everything still works without these deps running a `make test`, if you require something else please let me know.

### DIFF
--- a/vendor.conf
+++ b/vendor.conf
@@ -10,11 +10,9 @@ github.com/docker/go-connections 3ede32e2033de7505e6500d6c868c2b9ed9f169d
 github.com/docker/go-units 0dadbb0345b35ec7ef35e228dabb8de89a65bf52
 github.com/docker/libtrust aabc10ec26b754e797f9028f4589c5b7bd90dc20
 github.com/ghodss/yaml 04f313413ffd65ce25f2541bfd2b2ceec5c0908c
-github.com/gorilla/context 08b5f424b9271eedf6f9f0ce86cb9396ed337a42
 github.com/gorilla/mux 94e7d24fd285520f3d12ae998f7fdd6b5393d453
 github.com/imdario/mergo 50d4dbd4eb0e84778abe37cefef140271d96fade
 github.com/mattn/go-runewidth 14207d285c6c197daabb5c9793d63e7af9ab2d50
-github.com/mattn/go-shellwords 005a0944d84452842197c2108bd9168ced206f78
 github.com/mistifyio/go-zfs c0224de804d438efd11ea6e52ada8014537d6062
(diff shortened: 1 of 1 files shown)


rhatdan:
@agonzalezro you need to sign your PR.

git commit -a --amend -s
git push --force

Should fix this.

mtrmac:
From a quick look it seems that most, if not all, of the dependencies are indeed transitively required. I guess we’ll see…

agonzalezro:
Signed off @rhatdan

mtrmac:
Yeah, tests fail because some of the packages are actually needed.

agonzalezro:
@mtrmac it's green now, however, just three dependencies were removed from the vendor file at the end. I don't know if it's worthy to merge, feel free to decline the PR if you think it's not.

rhatdan:
I think this is valid to merge, not sure if it shrinks any packages sized, but it is cleaner to only put the required vendor packages in.

LGTM

mtrmac:
I do think cleaning up the dependencies is worthwhile.

- The gorilla/context dependency is indeed no longer necessary with go ≥ 1.7.
- `github.com/mattn/go-shellwords` is interesting: it is needed by `vendor/github.com/containers/storage/pkg/parsers/kernel/kernel_darwin.go` (macOS-only), but that is needed by `vendor/github.com/containers/storage/drivers/devmapper/deviceset.go` (Linux-only).  So, ultimately, not necessary—this is a bit fragile, but we can always re-add it again..
- `golang.org/x/sys/unix`, a subpackage of `golang.org/x/sys/` is clearly required by `vendor/github.com/containers/storage/lockfile_unix.go ` and others. The reason the tests don’t fail is that `make tools`, while installing the tools, also downloads the dependencies into `$GOPATH`, so a missing package in `vendor` is not noticed. So, that one should stay.

agonzalezro:
@mtrmac, do you think it's OK to force Go >= 1.7?

agonzalezro:
I've added back `x/sys/unix`.

rhatdan:
I believe we should drop support for older go.  Especially < 1.8.

mtrmac:
👍 

Thanks!

> @mtrmac, do you think it's OK to force Go >= 1.7?

We already require ≥1.7, at the very least by using `import "context"` everywhere; the package has been added in 1.7.

PR #448, opened 2018-04-24
Replace docker.io as the default registry with <local>

rhatdan:

### DIFF
--- a/docker/daemon/daemon_transport_test.go
+++ b/docker/daemon/daemon_transport_test.go
@@ -51,8 +51,8 @@ func testParseReference(t *testing.T, fn func(string) (types.ImageReference, err
 		{"sha256:XX23456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef", "", ""}, // Invalid digest value
 		{"UPPERCASEISINVALID", "", ""},                                                      // Invalid reference input
 		{"busybox", "", ""},                                                                 // Missing tag or digest
-		{"busybox:latest", "", "docker.io/library/busybox:latest"},                          // Explicit tag
-		{"busybox@" + sha256digest, "", "docker.io/library/busybox@" + sha256digest},        // Explicit digest
+		{"busybox:latest", "", "<local>/library/busybox:latest"},                            // Explicit tag
+		{"busybox@" + sha256digest, "", "<local>/library/busybox@" + sha256digest},          // Explicit digest
 		// A github.com/distribution/reference value can have a tag and a digest at the same time!
 		// Most versions of docker/reference do not handle that (ignoring the tag), so we reject such input.
 		{"busybox:latest@" + sha256digest, "", ""},                                   // Both tag and digest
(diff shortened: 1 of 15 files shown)


rhatdan:
@baude @TomSweeneyRedHat @mheon @runcom @mtrmac WDYT

This would change the default for buildah commit to create
localhost/library/IMAGENAME:latest
rather then 
docker.io/library/IMAGENAME:latest

rhatdan:
@nalind WDYT

runcom:
`localhost` may refer to a locally deployed container registry which is ... wrong maybe since the intention is to have _local_ images in the sense they're not on any registry

rhatdan:
Sure I would be fine with image:latest as well. but the code is putting a "/" in in that case and gets an invalid name.

rhatdan:
@runcom @mtrmac PTAL

rhatdan:
@runcom Do you know why git is failing?

rhatdan:
@nalind @umohnani8 @mheon @baude FYI

mheon:
Is there a reason it is `<local>/library/$IMAGENAME` as opposed to just `<local>/$IMAGENAME`?

rhatdan:
@mheon I would say lets get this in first and then we can look at that.

mtrmac:
FWIW the `library` behavior is specialized to `docker.io`, it does not happen for other registry hostnames, at least nowadays.

docker/docker_transport_test.go
```diff
@@ -44,9 +44,9 @@ func TestParseReference(t *testing.T) {
 func testParseReference(t *testing.T, fn func(string) (types.ImageReference, error)) {
 	for _, c := range []struct{ input, expected string }{
 		{"busybox", ""}, // Missing // prefix
-		{"//busybox:notlatest", "docker.io/library/busybox:notlatest"},           // Explicit tag
-		{"//busybox" + sha256digest, "docker.io/library/busybox" + sha256digest}, // Explicit digest
-		{"//busybox", "docker.io/library/busybox:latest"},                        // Default tag
+		{"//busybox:notlatest", "<local>/library/busybox:notlatest"},           // Explicit tag
+		{"//busybox" + sha256digest, "<local>/library/busybox" + sha256digest}, // Explicit digest
+		{"//busybox", "<local>/library/busybox:latest"},                        // Default tag
```
mtrmac: This still essentially breaks the world; `skopeo inspect docker://busybox` will no longer do anything useful.

docker/daemon/daemon_transport_test.go
```diff
@@ -51,8 +51,8 @@ func testParseReference(t *testing.T, fn func(string) (types.ImageReference, err
 		{"sha256:XX23456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef", "", ""}, // Invalid digest value
 		{"UPPERCASEISINVALID", "", ""},                                                      // Invalid reference input
 		{"busybox", "", ""},                                                                 // Missing tag or digest
-		{"busybox:latest", "", "docker.io/library/busybox:latest"},                          // Explicit tag
-		{"busybox@" + sha256digest, "", "docker.io/library/busybox@" + sha256digest},        // Explicit digest
+		{"busybox:latest", "", "<local>/library/busybox:latest"},                            // Explicit tag
+		{"busybox@" + sha256digest, "", "<local>/library/busybox@" + sha256digest},          // Explicit digest
```
mtrmac: See the long comment in `docker/tarfile/Destination.PutManifest`, this would break interoperability with projectatomic/docker.

signature/policy_reference_match_test.go
```diff
@@ -136,7 +136,7 @@ var prmExactMatchTestTable = []prmSymmetricTableTest{
 	{"busybox" + digestSuffix, "busybox" + digestSuffix, true}, // NOTE: This is not documented; signing digests is not recommended at this time.
 	// Non-canonical reference format is canonicalized
 	{"library/busybox:latest", "busybox:latest", true},
-	{"docker.io/library/busybox:latest", "busybox:latest", true},
+	{"<local>/library/busybox:latest", "busybox:latest", true},
```
mtrmac: This redefines the meaning of existing signatures, e.g. https://rhelblog.redhat.com/2016/07/22/container-image-signing/ shows a reference that does not contain a host name.

(_Recent_ code uses the fully explicit format in created signatures, I’m afraid that hasn’t been originally the case.)

rhatdan:
CLosing this PR in favor of https://github.com/projectatomic/buildah/pull/648

PR #449, opened 2018-04-24
[WIP] Replace docker.io as the default registry with registries[0]

rhatdan:

### DIFF
--- a/docker/reference/normalize.go
+++ b/docker/reference/normalize.go
@@ -5,12 +5,12 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/containers/image/types"
 	"github.com/opencontainers/go-digest"
 )
 
 var (
 	legacyDefaultDomain = "index.docker.io"
-	defaultDomain       = "docker.io"
 	officialRepoName    = "library"
 	defaultTag          = "latest"
 )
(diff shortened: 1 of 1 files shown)


rhatdan:
This is an alternative to https://github.com/containers/image/pull/448

docker/reference/normalize.go
```diff
@@ -29,11 +29,18 @@ type normalizedNamed interface {
...
+func ParseNormalizedNamed(sys *types.SystemContext, s string) (Named, error) {
 	if ok := anchoredIdentifierRegexp.MatchString(s); ok {
 		return nil, fmt.Errorf("invalid repository name (%s), cannot specify 64-byte hexadecimal strings", s)
 	}
-	domain, remainder := splitDockerDomain(s)
+	registries, err := GetRegistries(sys)
+	if err != nil {
+		return nil, err
+	}
+	if len(registries) == 0 {
```
TomSweeneyRedHat: Should we default to quay.io in this case?  Or fedora?

docker/reference/normalize.go
```diff
@@ -29,11 +29,18 @@ type normalizedNamed interface {
...
+		return nil, errors.New("no registries specified")
```
TomSweeneyRedHat: "no registries specified in /etc/containers/registries.conf" perhaps?

TomSweeneyRedHat:
This is more in line with what I was thinking you were going to do.  Looks like you've a whitespace error somewhere.  I'd go forward with this and kill the earlier one.

docker/reference/normalize.go
rhatdan: The problem is I am not sure if that is the path.

docker/reference/normalize.go
rhatdan: I really don't think the containers/image should have a default.  In the perfect world we would just allow a "" as the registry.

mtrmac: A minimal comment for now:

`c/i/docker/reference.Named`, `!IsNameOnly`, is a type representing a _specific_ image in a _specific_ registry.

Maybe there needs to be a type representing the “unqualified” inputs to search.

Maybe some uses of `reference.Named` should not use it at all, and accept more lax input. (In particular I was wrong about `containers-storage:`, it uses `reference.ParseNormalizedNamed` and does not allow hostname-less references.)

But, if hostname-qualified and hostname-less references should live side by side, they _need_ to be distinct types. The experience, _both_ with projectatomic/docker’s `reference`, and with podman’s `string` uses, is that mixing them together makes it all to easy to be careless about the semantics and results in surprising and unexpected corner case behaviors.

rhatdan:
I think I prefer the original.  No registry at all.

PR #451, opened 2018-04-26
Don't fail if XDG_RUNTIME_DIR is set to a non-existent path

umohnani8:
If the image is a public image, credentials are not needed to pull it,
try pulling the image even if XDG_RUNTIME_DIR is set to a non-existent path.

Addresses https://github.com/projectatomic/libpod/issues/659

### DIFF
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -15,6 +15,7 @@ import (
 	"github.com/docker/docker-credential-helpers/credentials"
 	"github.com/docker/docker/pkg/homedir"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 )
 
 type dockerAuthConfig struct {
@@ -64,11 +65,17 @@ func GetAuthentication(sys *types.SystemContext, registry string) (string, strin
 	}
 
 	dockerLegacyPath := filepath.Join(homedir.Get(), dockerLegacyCfg)
+	var paths []string
 	pathToAuth, err := getPathToAuth(sys)
-	if err != nil {
-		return "", "", err
+	if err == nil {
+		paths = append(paths, pathToAuth)
+	} else {
+		// Error means that the path set for XDG_RUNTIME_DIR does not exist
+		// but we don't want to completely fail in the case that the user is pulling a public image
+		// Logging the error as a warning instead and moving on to pulling the image
+		logrus.Warnf("%v: Trying to pull image in the event that it is a public image.", err)
 	}
-	paths := [3]string{pathToAuth, filepath.Join(homedir.Get(), dockerCfg, dockerCfgFileName), dockerLegacyPath}
+	paths = append(paths, filepath.Join(homedir.Get(), dockerCfg, dockerCfgFileName), dockerLegacyPath)
 
 	for _, path := range paths {
 		legacyFormat := path == dockerLegacyPath


umohnani8:
@mtrmac @rhatdan PTAL

pkg/docker/config/config.go
```diff
@@ -66,7 +67,10 @@ func GetAuthentication(sys *types.SystemContext, registry string) (string, strin
 	dockerLegacyPath := filepath.Join(homedir.Get(), dockerLegacyCfg)
 	pathToAuth, err := getPathToAuth(sys)
 	if err != nil {
-		return "", "", err
+		// Error means that the path set for XDG_RUNTIME_DIR does not exist
+		// but we don't want to completely fail in the case that the user is pulling a public image
+		// Logging the error as a warning instead and moving on to pulling the image
+		logrus.Warnf("%v: Trying to pull image in the event that it is a public image.", err)
```
umohnani8: For this part, should I immediately `return "", "", nil` after logging the warning? Or should it go ahead and check the other (docker) auth files to see if it can find creds?

rhatdan: How about

```
	paths := [3]string{filepath.Join(homedir.Get(), dockerCfg, dockerCfgFileName), dockerLegacyPath}
 	pathToAuth, err := getPathToAuth(sys)
 	if err == nil {
                paths = append(pathToAuth, paths)
       else {
      }
```

rhatdan: I think you should check for the docker auth files as well.

umohnani8: okay, fixed.

pkg/docker/config/config.go
```diff
@@ -64,11 +65,17 @@ func GetAuthentication(sys *types.SystemContext, registry string) (string, strin
 	}
 
 	dockerLegacyPath := filepath.Join(homedir.Get(), dockerLegacyCfg)
+	var paths []string
 	pathToAuth, err := getPathToAuth(sys)
-	if err != nil {
-		return "", "", err
+	if err == nil {
+		paths = append(paths, pathToAuth)
```
rhatdan: I think you reversed it.  We want pathToAuth to be first.
Does this work?
```
paths = append([]string{pathToAuth}, paths ...)
```

umohnani8: yup, it will be the first thing in `paths`. The other file paths are appended later on.

rhatdan: Ok I'm an idiot, I thouht you had just taken my example, but you did it differently.

LGTM

rhatdan:
LGTM
@mtrmac @runcom PTAL
We need this to fix a bug in podman

runcom:
LGTM

umohnani8:
@mtrmac PTAL

pkg/docker/config/config.go
```diff
@@ -64,11 +65,17 @@ func GetAuthentication(sys *types.SystemContext, registry string) (string, strin
...
+	} else {
+		// Error means that the path set for XDG_RUNTIME_DIR does not exist
+		// but we don't want to completely fail in the case that the user is pulling a public image
+		// Logging the error as a warning instead and moving on to pulling the image
+		logrus.Warnf("%v: Trying to pull image in the event that it is a public image.", err)
 	}
-	paths := [3]string{pathToAuth, filepath.Join(homedir.Get(), dockerCfg, dockerCfgFileName), dockerLegacyPath}
+	paths = append(paths, []string{filepath.Join(homedir.Get(), dockerCfg, dockerCfgFileName), dockerLegacyPath}...)
```
mtrmac: The `[]string{}...` wrap+unwrap is unnecessary:
```go
paths = append(paths, filepath.Join(homedir.Get(), dockerCfg, dockerCfgFileName), dockerLegacyPath)
```

umohnani8: Agreed, fixed.

mtrmac:
👍  Thanks!

umohnani8:
Tests passed and approved. Ready to merge :)

PR #453, opened 2018-04-30
getPathToAuth(): bail on any error, not just ENOENT

edsantiago:
It's possible for stat() to fail in ways other than ENOENT.
If it does, bail out instead of silently cascading an
error condition.

Signed-off-by: Ed Santiago <santiago@redhat.com>

### DIFF
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -155,6 +155,8 @@ func getPathToAuth(sys *types.SystemContext) (string, error) {
 			// or made a typo while setting the environment variable
 			// so we log the error and return an empty string as the path
 			return "", errors.Wrapf(err, "%q directory set by $XDG_RUNTIME_DIR does not exist. Either create the directory or unset $XDG_RUNTIME_DIR.", runtimeDir)
+		} else if err != nil {
+			return "", err
 		}
 		runtimeDir = filepath.Join(runtimeDir, authCfg)
 	} else {


rhatdan:
LGTM

rhatdan:
Not sure about the validation error, does not like your signature or does not like you.

edsantiago:
>I guess this is a formatting error

That was it! It didn't like my parentheses. Fixed & force-pushed. Many thanks.

mtrmac:
> It's possible for stat() to fail in ways other than ENOENT.
If it does, bail out instead of silently cascading an
error condition.

In which case does this make any difference? AFAICT all users of `getPathToAuth` would fail when accessing the file inside the `XDG_RUNTIME_DIR` directory very soon after this function returns. I must be missing something.

(The callers of `getPathToAuth` are not in general promised that the returned path is accessible; only the `XDG_RUNTIME_DIR` case does any checking at all. So they need to be able to handle that case anyway.)

edsantiago:
You're not missing anything; it's just that I'm (sometimes overly) neurotic about catching errors and failing early. If there were no error checking in the function, I wouldn't mind letting the caller deal with it. But since there is, and it's incomplete, the fallthrough bothers me. It's just the sort of thing a future maintainer might question, costing him/her time.

mtrmac:
The incomplete-looking error handling being confusing is a fair point; OTOH the other paths have no access checks at all, which is also inconsistent.

Perhaps we could just add a comment in there, something like
```go
// We don’t need to handle the error here, callers will fail accessing the config file at the returned path
```
, maybe also adding something like “The access would fail anyway; we are checking for `os.IsNotExist` only to give the user better guidance what to do in this special case” to the comment above.

mtrmac:
I have hopefully clarified the situation in comments in #457.

PR #455, opened 2018-05-06
errors.Wrapf doesn't return error when err is nil

agonzalezro:
In this commit you can see how at the beginning of the function
`getManifestDescriptor` it was being checked for an error was different than
nil and returning in that case, however, few lines later that error (remember,
with nil value) was being wrapped. Since the original error was nil, the
wrapping was as well returning a nil.

A regression tests was added.

This commit should close the following issue in skopeo as per @mtrmac
suggestion: https://github.com/projectatomic/skopeo/issues/496

### DIFF
--- a/oci/layout/oci_transport_test.go
+++ b/oci/layout/oci_transport_test.go
@@ -13,6 +13,18 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// TestGetManifestDescriptor is testing a regression issue where a nil error was being wrapped,
+// this causes the returned error to be nil as well and the user wasn't getting a proper error output.
+//
+// More info: https://github.com/projectatomic/skopeo/issues/496
+func TestGetManifestDescriptor(t *testing.T) {
+	imageRef, err := NewReference("fixtures/two_images_manifest", "")
+	require.NoError(t, err)
+
+	_, err = imageRef.(ociReference).getManifestDescriptor()
+	assert.EqualError(t, err, ErrMoreThanOneImage.Error())
+}
+
 func TestTransportName(t *testing.T) {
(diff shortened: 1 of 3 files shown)


agonzalezro:
I know why the tests are failing, I have added an extra manifest to the fixture but it seems it was being used in some other place (I just run locally the tests for the file I was changing). I will take a look this afternoon.

oci/layout/oci_transport_test.go
```diff
@@ -13,6 +13,17 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestMoreThanOneImageError(t *testing.T) {
+	ref, tmpDir := refToTempOCI(t)
+	defer os.RemoveAll(tmpDir)
+
+	ociRef := ref.(ociReference)
+	ociRef.image = ""
```
mtrmac: Thanks!

Rather than affecting `refToTempOCI` for everyone, and the type casts here, could the test create its own manifest? And because this is read-only, there isn’t really a need to use temporary directories; the test case data could be an independent subdirectory in the expected format (and potentially editable using standard OCI-processing tools) in a `fixtures` subdirectory, like `oci_src_test.go` uses a `fixtures/manifest` subdirectory for its tests.

And maybe, to be consistent with the rest of this file, name this `TestGetManifestDescriptor`, with a note that it tests against only one regression and the rest of the function is still untested.

rhatdan:
@agonzalezro Needs a rebase.
@mtrmac PTAL

agonzalezro:
@rhatdan done

runcom:
LGTM @mtrmac PTAL

rhatdan:
@agonzalezro One more update please.

rhatdan:
:+1:

PR #456, opened 2018-05-07
tmpdir: unixTempDirForBigFiles can be overridden at build time

nlewo:
This commit exposes a variable `unixTempDirForBigFiles` to allow some
Linux distributions to use `/tmp`.
On NixOS for instance, `/tmp` is not a tmpfs and Nix sandboxed builds
do not allow to write to `/var/tmp`, only to `/tmp`.

Note the `tmpdir.TemporaryDirectoryForBigFiles()` is now also used in
the `storage` package to create temporary directories.

Signed-off-by: Antoine Eiche <lewo@abesis.fr>

### DIFF
--- a/internal/tmpdir/tmpdir.go
+++ b/internal/tmpdir/tmpdir.go
@@ -5,6 +5,16 @@ import (
 	"runtime"
 )
 
+// unixTempDirForBigFiles is the directory path to store big files on non Windows systems.
+// You can override this at build time with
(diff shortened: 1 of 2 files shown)


mtrmac:
:+1: pending tests.

Thanks!  Please rebase and I’ll merge.

mtrmac:
The test suite failure is unrelated, Fedora was updated to F28, which includes golang 1.10.  I’m working on a fix.

nlewo:
Thanks @mtrmac!

PR #457, opened 2018-05-12
Clarify that err != nil && !os.IsNotExist(err) does not need to be handled

mtrmac:
... and that this is not an oversight, per the conversation in https://github.com/containers/image/pull/453 .


mtrmac:
👍

PR #458, opened 2018-05-14
Allow image to be saved without a reference name

umohnani8:
The docker-archive and oci-archive transport should allow the
destination of the image to be valid without the reference part also.
Format is transport:path[:reference] where reference is optional.
This is for the case where a user just wants to save/push an image with
the image ID only and not the name.
Creates archives with empty repotags.

Fixes one of the issues in https://github.com/projectatomic/libpod/issues/649

### DIFF
--- a/docker/archive/transport.go
+++ b/docker/archive/transport.go
@@ -41,7 +41,9 @@ func (t archiveTransport) ValidatePolicyConfigurationScope(scope string) error {
 
 // archiveReference is an ImageReference for Docker images.
 type archiveReference struct {
-	destinationRef reference.NamedTagged // only used for destinations
+	// only used for destinations
+	// archiveReference.destinationRef is optional and can be nil for destinations as well.
+	destinationRef reference.NamedTagged
 	path           string
 }
 
--- a/docker/archive/transport_test.go
+++ b/docker/archive/transport_test.go
@@ -169,7 +169,8 @@ func TestReferenceNewImageDestination(t *testing.T) {
 	ref, err := ParseReference(filepath.Join(tmpDir, "no-reference"))
 	require.NoError(t, err)
 	dest, err := ref.NewImageDestination(context.Background(), nil)
-	assert.Error(t, err)
+	assert.NoError(t, err)
+	dest.Close()
 
 	ref, err = ParseReference(filepath.Join(tmpDir, "with-reference") + "busybox:latest")
 	require.NoError(t, err)
--- a/docker/tarfile/dest.go
+++ b/docker/tarfile/dest.go
@@ -23,10 +23,9 @@ import (
 
 // Destination is a partial implementation of types.ImageDestination for writing to an io.Writer.
 type Destination struct {
-	writer    io.Writer
-	tar       *tar.Writer
-	reference reference.NamedTagged
-	repoTags  []reference.NamedTagged
+	writer   io.Writer
+	tar      *tar.Writer
+	repoTags []reference.NamedTagged
 	// Other state.
 	blobs  map[digest.Digest]types.BlobInfo // list of already-sent blobs
 	config []byte
@@ -34,12 +33,15 @@ type Destination struct {
 
 // NewDestination returns a tarfile.Destination for the specified io.Writer.
 func NewDestination(dest io.Writer, ref reference.NamedTagged) *Destination {
+	repoTags := []reference.NamedTagged{}
+	if ref != nil {
+		repoTags = append(repoTags, ref)
+	}
 	return &Destination{
-		writer:    dest,
-		tar:       tar.NewWriter(dest),
-		reference: ref,
-		repoTags:  []reference.NamedTagged{ref},
-		blobs:     make(map[digest.Digest]types.BlobInfo),
+		writer:   dest,
+		tar:      tar.NewWriter(dest),
+		repoTags: repoTags,
+		blobs:    make(map[digest.Digest]types.BlobInfo),
 	}
 }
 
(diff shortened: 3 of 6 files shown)


umohnani8:
@mtrmac, @runcom PTAL

oci/layout/oci_dest.go
```diff
@@ -217,10 +213,6 @@ func (d *ociImageDestination) PutManifest(ctx context.Context, m []byte) error {
 		return err
 	}
 
-	if d.ref.image == "" {
-		return errors.Errorf("cannot save image with empyt image.ref.name")
-	}
-
 	annotations := make(map[string]string)
 	annotations["org.opencontainers.image.ref.name"] = d.ref.image
```
mtrmac: AFAICT an empty string is not a valid value for this annotation; it should not be added if `d.ref.image == ""`.

docker/tarfile/dest.go
```diff
@@ -168,8 +172,10 @@ func (d *Destination) ReapplyBlob(ctx context.Context, info types.BlobInfo) (typ
 }
 
 func (d *Destination) createRepositoriesFile(rootLayerID string) error {
-	repositories := map[string]map[string]string{
-		d.reference.Name(): {d.reference.Tag(): rootLayerID}}
+	repositories := map[string]map[string]string{}
+	if d.reference != nil {
+		repositories[d.reference.Name()] = map[string]string{d.reference.Tag(): rootLayerID}
+	}
```
mtrmac: Actually this should have been extended in #450 to handle `repoTags` as well.  Could you do that, please? And then `d.reference` can be completely removed in favor of `d.repoTags`.

oci/layout/oci_dest.go
umohnani8: Fixed.
Added comments too for the `archiveReference.destinationRef` and `ociReference.image`.

docker/archive/transport.go
```diff
@@ -40,6 +40,7 @@ func (t archiveTransport) ValidatePolicyConfigurationScope(scope string) error {
 }
 
 // archiveReference is an ImageReference for Docker images.
+// archiveReference.destinationRef is optional and can be nil for destinations as well.
```
mtrmac: Non-blocking nit: it would be nice to combine it with the field-specific comment below.

docker/archive/transport_test.go
```diff
@@ -169,7 +169,7 @@ func TestReferenceNewImageDestination(t *testing.T) {
 	ref, err := ParseReference(filepath.Join(tmpDir, "no-reference"))
 	require.NoError(t, err)
 	dest, err := ref.NewImageDestination(context.Background(), nil)
-	assert.Error(t, err)
+	assert.NoError(t, err)
```
mtrmac: `dest.Close()` should be called here.

docker/tarfile/dest.go
```diff
@@ -34,11 +34,15 @@ type Destination struct {
...
 // NewDestination returns a tarfile.Destination for the specified io.Writer.
 func NewDestination(dest io.Writer, ref reference.NamedTagged) *Destination {
+	repoTags := []reference.NamedTagged{}
+	if ref != nil {
+		repoTags = append(repoTags, ref)
+	}
 	return &Destination{
 		writer:    dest,
 		tar:       tar.NewWriter(dest),
 		reference: ref,
```
mtrmac: Please drop the `reference` field, it is no longer used by anything.

docker/tarfile/dest.go
```diff
@@ -168,8 +172,12 @@ func (d *Destination) ReapplyBlob(ctx context.Context, info types.BlobInfo) (typ
 }
 
 func (d *Destination) createRepositoriesFile(rootLayerID string) error {
-	repositories := map[string]map[string]string{
-		d.reference.Name(): {d.reference.Tag(): rootLayerID}}
+	repositories := map[string]map[string]string{}
+	if d.repoTags != nil {
```
mtrmac: (Non-blocking: `d.repoTags != nil` is always true due to how `NewDestination` initializes it AFAICT; also, `nil` is a valid slice value, so `range d.repoTags` does the right thing if it were `nil`.  So, this conditional does not need to exist.)

docker/tarfile/dest.go
```diff
@@ -168,8 +172,12 @@ func (d *Destination) ReapplyBlob(ctx context.Context, info types.BlobInfo) (typ
...
+		for _, repoTag := range d.repoTags {
+			repositories[repoTag.Name()] = map[string]string{repoTag.Tag(): rootLayerID}
```
mtrmac: I don't think this works for `repoTags` which share a name, e.g. "busybox:tag1", "busybox:tag2"; only one tag would get recorded.

docker/archive/transport.go
umohnani8: fixed.

docker/archive/transport_test.go
umohnani8: fixed.

docker/tarfile/dest.go
umohnani8: fixed.

docker/tarfile/dest.go
umohnani8: right, my bad. Fixed.

docker/tarfile/dest.go
umohnani8: Fixed, should record both tags  now.

mtrmac:
ACK, thanks!  Can you prepare a skopeo update as well, please?

umohnani8:
Skopeo tests fixed in https://github.com/projectatomic/skopeo/pull/505

mtrmac:
👍

PR #459, opened 2018-05-22
Add github.com/syndtr/gocapability to vendor.conf

mtrmac:
This is a newly required dependency of containers/storage.


rhatdan:
Do you actually have to vendor any code into containers/image?

mtrmac:
containers/image does not actually vendor anything in the repo, because $user`/vendor/…/foo` and $user`/vendor/github.com/containers/image/vendor/…/foo` are incompatible types; $user must make all dependencies of c/image available, e.g. by its own vendoring.

The `vendor.conf` in this repo is only actually used for running tests.

PR #461, opened 2018-05-30
ostree: use the same thread for ostree operations

giuseppe:
Since https://github.com/ostreedev/ostree/pull/1555, locking is
enabled by default in OSTree.  Unfortunately it uses thread-private
data and it breaks the Golang bindings.  Force the same thread for the
write operations to the OSTree repository.


cgwalters:
LGTM

mtrmac:
👍 Thanks!

PR #463, opened 2018-06-06
make and travis: cross platform compile check

vbatts:
Signed-off-by: Vincent Batts <vbatts@hashbangbash.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,10 +6,18 @@ SKOPEO_BRANCH = master
 # Set SUDO=sudo to run container integration tests using sudo.
 SUDO =
 
+# when cross compiling _for_ a Darwin or windows host, then we must use openpgp
+BUILD_TAGS_WINDOWS_CROSS = containers_image_ostree_stub containers_image_openpgp
+BUILD_TAGS_DARWIN_CROSS = containers_image_ostree_stub containers_image_openpgp
+# when compiling _on_ a Darwin host, then we can link against gpgme
+BUILD_TAGS_DARWIN_NATIVE = containers_image_ostree_stub
+
 ifeq ($(shell uname),Darwin)
-DARWIN_BUILD_TAG = containers_image_ostree_stub
+PLATFORM_BUILD_TAG = $(BUILD_TAGS_DARWIN_NATIVE)
 endif
-BUILDTAGS = btrfs_noversion libdm_no_deferred_remove $(DARWIN_BUILD_TAG)
+
+
+BUILDTAGS = btrfs_noversion libdm_no_deferred_remove $(PLATFORM_BUILD_TAG)
 BUILDFLAGS := -tags "$(BUILDTAGS)"
 
 PACKAGES := $(shell go list $(BUILDFLAGS) ./... | grep -v github.com/containers/image/vendor)
@@ -25,8 +33,14 @@ GPGME_ENV = CGO_CFLAGS="$(shell gpgme-config --cflags 2>/dev/null)" CGO_LDFLAGS=
 
 all: tools test validate .gitvalidation
 
-build: vendor
-	@$(GPGME_ENV) go build $(BUILDFLAGS) $(PACKAGES)
+build: vendor build-internal
+
+build-internal:
+	$(GPGME_ENV) go build $(BUILDFLAGS) $(PACKAGES)
+
+cross: vendor
+	GOOS=windows $(MAKE) build-internal BUILDTAGS="$(BUILDTAGS) $(BUILD_TAGS_WINDOWS_CROSS)"
+	GOOS=darwin $(MAKE) build-internal BUILDTAGS="$(BUILDTAGS) $(BUILD_TAGS_DARWIN_CROSS)"
 
 tools: tools.timestamp
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,4 +24,4 @@ script: >
   -e GOPATH=/gopath -e TRASH_CACHE=/gopath/.trashcache
   -v /etc/passwd:/etc/passwd -v /etc/sudoers:/etc/sudoers -v /etc/sudoers.d:/etc/sudoers.d
   -v /var/run:/var/run:z -v $HOME/gopath:/gopath:Z
-  -w /gopath/src/github.com/containers/image image-test bash -c "PATH=$PATH:/gopath/bin make tools .gitvalidation validate test test-skopeo SUDO=sudo BUILDTAGS=\"$BUILDTAGS\""
+  -w /gopath/src/github.com/containers/image image-test bash -c "PATH=$PATH:/gopath/bin make cross tools .gitvalidation validate test test-skopeo SUDO=sudo BUILDTAGS=\"$BUILDTAGS\""
--- a/vendor.conf
+++ b/vendor.conf
@@ -40,3 +40,5 @@ github.com/ostreedev/ostree-go aeb02c6b6aa2889db3ef62f7855650755befd460
 github.com/gogo/protobuf fcdc5011193ff531a548e9b0301828d5a5b97fd8
 github.com/pquerna/ffjson master
 github.com/syndtr/gocapability master
+github.com/Microsoft/go-winio ab35fc04b6365e8fcb18e6e9e41ea4a02b10b175
+github.com/Microsoft/hcsshim eca7177590cdcbd25bbc5df27e3b693a54b53a6a


vbatts:
hmmm. `make cross` works for me on my laptop, but not sure why it's getting a permission denied in this travis docker environment.

rhatdan:
Looks like it is actually attempting to create a file on disk as opposed to creating it in the test area?

Makefile
```diff
@@ -7,12 +7,18 @@ SKOPEO_BRANCH = master
 SUDO =
 
 ifeq ($(shell uname),Darwin)
-DARWIN_BUILD_TAG = containers_image_ostree_stub
+PLATFORM_BUILD_TAG = containers_image_ostree_stub containers_image_openpgp
```
mtrmac: No, non-OpenPGP builds are expected to work natively.  Maybe you want to define a separate cross-build flags variable instead?

Makefile
```diff
@@ -7,12 +7,18 @@ SKOPEO_BRANCH = master
...
+endif
+ifeq ($(GOOS),windows)
+PLATFORM_BUILD_TAG = containers_image_ostree_stub containers_image_openpgp
+endif
+BUILDTAGS = btrfs_noversion libdm_no_deferred_remove $(PLATFORM_BUILD_TAG)
 BUILDFLAGS := -tags "$(BUILDTAGS)"
 
-PACKAGES := $(shell go list $(BUILDFLAGS) ./... | grep -v github.com/containers/image/vendor)
+PACKAGES ?= $(shell go list $(BUILDFLAGS) ./... | grep -v github.com/containers/image/vendor)
```
mtrmac: Why is this necessary?

Makefile
```diff
@@ -7,12 +7,18 @@ SKOPEO_BRANCH = master
...
 ifeq ($(shell uname),Darwin)
-DARWIN_BUILD_TAG = containers_image_ostree_stub
+PLATFORM_BUILD_TAG = containers_image_ostree_stub containers_image_openpgp
 endif
-BUILDTAGS = btrfs_noversion libdm_no_deferred_remove $(DARWIN_BUILD_TAG)
+ifeq ($(GOOS),darwin)
+PLATFORM_BUILD_TAG = containers_image_ostree_stub containers_image_openpgp
+endif
+ifeq ($(GOOS),windows)
+PLATFORM_BUILD_TAG = containers_image_ostree_stub containers_image_openpgp
```
mtrmac: Yeah, at three copies, a separate cross-build variable seems much cleaner.

Makefile
```diff
@@ -26,7 +32,11 @@ GPGME_ENV = CGO_CFLAGS="$(shell gpgme-config --cflags 2>/dev/null)" CGO_LDFLAGS=
 all: tools test validate .gitvalidation
 
 build: vendor
-	@$(GPGME_ENV) go build $(BUILDFLAGS) $(PACKAGES)
+	$(GPGME_ENV) go build $(BUILDFLAGS) $(PACKAGES)
```
mtrmac: Is echoing the hard-to-parse list of packages that useful?

Makefile
```diff
@@ -26,7 +32,11 @@ GPGME_ENV = CGO_CFLAGS="$(shell gpgme-config --cflags 2>/dev/null)" CGO_LDFLAGS=
...
+
+cross:
+	GOOS=windows GOPATH=$(shell mktemp -d):$(GOPATH) $(MAKE) build
+	GOOS=darwin GOPATH=$(shell mktemp -d):$(GOPATH) $(MAKE) build
```
mtrmac: Wouldn’t it be more accurate to run native builds? At least for macOS this seems possible, e.g. skopeo does that.  Maybe not worth the effort for Windows, sure.

Makefile
vbatts: only when compiling on a darwin host, it can link to gpgme. But if you're cross compiling for darwin then it can't do the gpgme linking

mtrmac: Right, but this is the `shell uname` condition, when we _are_ on a Darwin host.

Makefile
vbatts: oh it was when i was going to set `PACKAGES` earlier in the makefile, but didn't need to as the build tags filtered the packages that were causing issues.

Makefile
vbatts: i'll refactor a bit

Makefile
vbatts: debugging. I'll add it back.

Makefile
vbatts: at least for a smoke test of the compile. We want this in travis.

vbatts:
updated for @mtrmac comments, but the travis step is still failing. I'm guessing it's the `vendor` target dependencies for the cross-platform targets?

Makefile
vbatts: look at how i've revised it

mtrmac:
> I'm guessing it's the vendor target dependencies for the cross-platform targets?

The `tools.timestamp` target, yes. Just run `go build` inside the `cross` target directly, without recursing into `make build`, perhaps?

vbatts:
nope, it needs the vendor dependency. :grimacing:

Makefile
```diff
@@ -6,10 +6,15 @@ SKOPEO_BRANCH = master
 # Set SUDO=sudo to run container integration tests using sudo.
 SUDO =
 
+ifneq ($(GOOS),linux)
```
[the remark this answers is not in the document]
mtrmac: Maybe optimize for clarity instead of number of lines?
```
DARWIN_NATIVE_TAGS = …ostree…
DARWIN_CROSS_TAGS = …ostree… …openpgp…
WINDOWS_CROSS_TAGS =…

…

ifeq ($(shell uname),Darwin)
   PLATFORM_BUILD_TAGS = $(DARWIN_NATIVE_TAGS)
endif

…

build: vendor build-internal

build-internal:
	@$(GPGME_ENV) go build $(BUILDFLAGS) $(PACKAGES)

cross:
	make build-internal GOOS=windows PLATFORM_BUILD_TAGS="$(WINDOWS_CROSS_TAGS)"
	make build-internal GOOS=darwin PLATFORM_BUILD_TAGS="$(DARWIN_CROSS_TAGS)"
```

(or with `go build` instead of the new `build-internal` target; all completely untested)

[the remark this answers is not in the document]
vbatts: i'll give that a try

vbatts:
progress. adding the missing deps now to vendor.conf

vbatts:
weird. The PLATFORM_BUILD_TAGS pass through when i `make cross` locally, but they do not for this travis container build. :confused:

vbatts:
so dumb. The travis setting `BUILDTAGS` is overriding what the makefile has, so the `PLATFORM_BUILD_TAGS` don't get included ...

vbatts:
PTAL

Makefile
```diff
@@ -6,10 +6,18 @@ SKOPEO_BRANCH = master
...
 
+# when cross compiling _for_ a Darwin or windows host, then we must use openpgp
+BUILD_TAG_WINDOWS_CROSS = containers_image_ostree_stub containers_image_openpgp
+BUILD_TAG_DARWIN_CROSS = containers_image_ostree_stub containers_image_openpgp
+# when compiling _on_ a Darwin host, then we can link against gpgme
+BUILD_TAG_DARWIN_NATIVE = containers_image_ostree_stub
+
 ifeq ($(shell uname),Darwin)
-DARWIN_BUILD_TAG = containers_image_ostree_stub
+PLATFORM_BUILD_TAG = BUILD_TAG_DARWIN_NATIVE
```
mtrmac: Shouldn’t this be `$(BUILD_TAG_DAWIN_NATIVE)`?

vbatts:
Damn you're right. I'll have to fix this in a bit

Makefile
```diff
@@ -25,8 +33,14 @@ GPGME_ENV = CGO_CFLAGS="$(shell gpgme-config --cflags 2>/dev/null)" CGO_LDFLAGS=
...
-build: vendor
-	@$(GPGME_ENV) go build $(BUILDFLAGS) $(PACKAGES)
+build: vendor build-internal
+
+build-internal:
+	$(GPGME_ENV) go build $(BUILDFLAGS) $(PACKAGES)
+
+cross: vendor
+	GOOS=windows $(MAKE) build-internal BUILDTAGS="$(BUILDTAGS) $(BUILD_TAG_WINDOWS_CROSS)"
+	GOOS=darwin $(MAKE) build-internal BUILDTAGS="$(BUILDTAGS) $(BUILD_TAG_DARWIN_CROSS)"
```
mtrmac: (Non-blocking: This is a tiny bit imprecise, basically it works only because with the current set of tags, adding more tags never breaks a build.

Maybe `.travis.yml` should be overriding `PLATFORM_BUILD_TAG` instead of `BUILDTAGS`?

Non-blocking, it’s likely enough that testing this in all the platforms is not really worth the effort, at least right now.)

Makefile
```diff
@@ -6,10 +6,18 @@ SKOPEO_BRANCH = master
 # Set SUDO=sudo to run container integration tests using sudo.
 SUDO =
 
+# when cross compiling _for_ a Darwin or windows host, then we must use openpgp
+BUILD_TAG_WINDOWS_CROSS = containers_image_ostree_stub containers_image_openpgp
+BUILD_TAG_DARWIN_CROSS = containers_image_ostree_stub containers_image_openpgp
+# when compiling _on_ a Darwin host, then we can link against gpgme
+BUILD_TAG_DARWIN_NATIVE = containers_image_ostree_stub
```
mtrmac: Oh, and `s/TAG/TAGS/` please.

Makefile
vbatts: i tried a couple of permutations like that. I don't like the way the travis job is just passing through to a docker container, but the current way runs.

mtrmac:
👍 Thanks!

PR #464, opened 2018-06-08
Cleanup sysregsv2

vrothberg:
This PR is based on a discussion at projectatomic/libpod#721 and includes the following changes:

- sysregistriesv2: forbid URI schemes
- sysregistriesv2: improve config validation

### DIFF
--- a/pkg/sysregistriesv2/system_registries_v2.go
+++ b/pkg/sysregistriesv2/system_registries_v2.go
@@ -21,53 +21,22 @@ var systemRegistriesConfPath = builtinRegistriesConfPath
 // DO NOT change this, instead see systemRegistriesConfPath above.
 const builtinRegistriesConfPath = "/etc/containers/registries.conf"
 
-// tomlURL is an abstraction required to unmarshal the non-primitive and
-// non-local url.URL type when loading the toml config.
-type tomlURL struct {
-	url url.URL
-}
-
-// UnmashalText interprets and parses text as a net.URL type and assigns it to
-// the tomlURL's url field.
-func (r *tomlURL) UnmarshalText(text []byte) (err error) {
-	r.url, err = parseURL(string(text))
-	return err
-}
-
-// mirror is an internal type including all mirror data but the URL.
-type mirror struct {
+// Mirror represents a mirror. Mirrors can be used as pull-through caches for
+// registries.
+type Mirror struct {
+	// The mirror's URL.
+	URL string `toml:"url"`
 	// If true, certs verification will be skipped and HTTP (non-TLS)
 	// connections will be allowed.
 	Insecure bool `toml:"insecure"`
 }
 
-// Mirror represents a mirror. Mirrors can be used as pull-through caches for
-// registries.
-type Mirror struct {
-	// The mirror's URL.
-	URL url.URL
-	mirror
-}
-
-// tomlMirror is a serializable Mirror.
-type tomlMirror struct {
-	// Serializable mirror URL.
-	URL tomlURL `toml:"url"`
-	mirror
-}
-
-// toMirror transforms tmirror to a Mirror.
-func (tmir *tomlMirror) toMirror() (Mirror, error) {
-	if len(tmir.URL.url.String()) == 0 {
-		return Mirror{}, fmt.Errorf("mirror must include a URL")
-	}
-	mir := Mirror{URL: tmir.URL.url, mirror: tmir.mirror}
-	return mir, nil
-}
-
-// registry is an internal type including all registry data but the URL and the
-// array of associated mirrors.
-type registry struct {
+// Registry represents a registry.
+type Registry struct {
+	// Serializable registry URL.
+	URL string `toml:"url"`
+	// The registry's mirrors.
+	Mirrors []Mirror `toml:"mirror"`
 	// If true, pulling from the registry will be blocked.
 	Blocked bool `toml:"blocked"`
 	// If true, certs verification will be skipped and HTTP (non-TLS)
@@ -76,55 +45,11 @@ type registry struct {
 	// If true, the registry can be used when pulling an unqualified image.
 	Search bool `toml:"unqualified-search"`
 	// Prefix is used for matching images, and to translate one namespace to
-	// another.  If `Prefix="example.com/bar"`, `URL="https://example.com/foo/bar"`
+	// another.  If `Prefix="example.com/bar"`, `URL="example.com/foo/bar"`
 	// and we pull from "example.com/bar/myimage:latest", the image will
-	// effectively be pulled from https://example.com/foo/bar/myimage:latest.
+	// effectively be pulled from "example.com/foo/bar/myimage:latest".
 	// If no Prefix is specified, it defaults to the specified URL.
 	Prefix string `toml:"prefix"`
-}
-
-// Registry represents a registry.
-type Registry struct {
-	// Serializable registry URL.
-	URL url.URL
-	// The registry's mirrors.
-	Mirrors []Mirror
-	registry
-}
-
-// tomlRegistry is serializable Registry.
-type tomlRegistry struct {
-	URL     tomlURL      `toml:"url"`
-	Mirrors []tomlMirror `toml:"mirror"`
-	registry
-}
-
-// stripURIScheme strips the URI scheme from the given URL and returns it as
-// as string.
-func stripURIScheme(url url.URL) string {
-	return strings.TrimPrefix(url.String(), url.Scheme+"://")
-}
-
-// toRegistry transforms treg to a Registry.
-func (treg *tomlRegistry) toRegistry() (Registry, error) {
-	if len(treg.URL.url.String()) == 0 {
-		return Registry{}, fmt.Errorf("registry must include a URL")
-	}
-	reg := Registry{URL: treg.URL.url, registry: treg.registry}
-	// if no prefix is specified, default to the specified URL with
-	// stripped URI scheme
-	if reg.Prefix == "" {
-		reg.Prefix = stripURIScheme(reg.URL)
-	}
-
-	for _, tmir := range treg.Mirrors {
-		mir, err := tmir.toMirror()
-		if err != nil {
-			return Registry{}, err
-		}
-		reg.Mirrors = append(reg.Mirrors, mir)
-	}
-	return reg, nil
 }
 
 // backwards compatability to sysregistries v1
@@ -134,7 +59,7 @@ type v1TOMLregistries struct {
 
 // tomlConfig is the data type used to unmarshal the toml config.
 type tomlConfig struct {
-	TOMLRegistries []tomlRegistry `toml:"registry"`
+	Registries []Registry `toml:"registry"`
 	// backwards compatability to sysregistries v1
 	V1Registries struct {
 		Search   v1TOMLregistries `toml:"search"`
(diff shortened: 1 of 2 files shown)


rhatdan:
@mtrmac @runcom PTAL

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -256,12 +218,12 @@ func GetRegistries(ctx *types.SystemContext) ([]Registry, error) {
 	}
 	if len(v1Registries) > 0 {
 		if len(registries) > 0 {
-			return nil, fmt.Errorf("mixing sysregistry v1/v2 is not supported")
+			return nil, errors.Wrap(&InvalidRegistries{}, "mixing sysregistry v1/v2 is not supported")
```
mtrmac: Shouldn’t this, and all other uses of `InvalidRegistries`, be `return nil, InvalidRegistries{s:"…"}` ?  It’s surprising for no user of `InvalidRegistries` to set the `s` field (why does it exist at all, then?), and if I’m reading it correctly, the `errors.Wrapf` use would result in an extra trailing `": "`.

mtrmac: (It might also be more convenient to use `type InvalidRegistries string`, because the error case would be just `return nil, InvalidRegistries("…")`.  OTOH it would be easier to accidentally mix the types than when a `struct` is involved.  Either is perfectly fine.)

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -233,21 +145,71 @@ func getV1Registries(config *tomlConfig) ([]Registry, error) {
...
+		var err error
+
+		// make sure URL and Prefix are valid
+		reg.URL, err = parseURL(reg.URL)
+		if err != nil {
+			return nil, err
+		}
+
+		if reg.Prefix == "" {
+			reg.Prefix = reg.URL
```
mtrmac: `s/validateRegistries/cleanAndValidateRegistries/`, or something like that, maybe?

It’s surprising for a “validate” function to modify its input.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -233,21 +145,71 @@ func getV1Registries(config *tomlConfig) ([]Registry, error) {
...
+	// Given a registry can be mentioned multiple times (e.g., to have
+	// multiple prefixes backed by different mirrors), we need to make sure
+	// there are no conflicts among them.
+	//
+	// Note: we need to iterate over the registries array to ensure a
+	// deterministic behavior which is not guaranteed by maps.
+	for _, reg := range registries {
+		others, exists := regMap[reg.URL]
+		if !exists {
+			continue
```
mtrmac: How can this ever happen? And if it can’t, shouldn’t the function fail with `Internal error` instead of silently continuing?

vrothberg: Good catch. I basically forgot to remove those lines; they're debugging artifacts.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -143,63 +68,50 @@ type tomlConfig struct {
...
+// github.com/containers/image/docker/reference.
+func parseURL(input string) (string, error) {
+	trimmed := strings.TrimRight(input, "/")
 
-	// user must be empty
-	if uri.User != nil {
-		// strip password for security reasons
-		uri.User = url.UserPassword(uri.User.Username(), "xxxxxx")
-		return url.URL{}, fmt.Errorf("unsupported username/password: %q", uri)
+	if _, err := reference.Parse(trimmed); err != nil {
```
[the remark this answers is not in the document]
vrothberg: I will correct the comment. It's a bit frustrating how tough parsing URI's without scheme can be. `net/url.URL` really didn't help much as, for instance, the scheme of `example.com:5000` would be `example.com`.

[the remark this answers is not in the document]
vrothberg: The `test_com` issue did not satisfy me, so I had another look at `net/url` and found the following comment:

```
URLs that do not start with a slash after the scheme are interpreted as:

scheme:opaque[?query][#fragment]
```

By considering the `URL.Opaque`, we can cover get a thorough validation. I will update the code and you can check if you prefer `docker/reference` or that.

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: Good catch, changed to `&InvalidRegistries{s: ...}`.

vrothberg:
@mtrmac I addressed the PR according to your comments. I also squashed the last two commits into one.

mtrmac:
Man, that `url.Parse` use is some non-obvious code.  @vrothberg , can you check that the comment added in the top commit of https://github.com/mtrmac/image/tree/sysregistriesv2-comments describes the situation correctly, please?

If so, LGTM, and I’ll add the comment after merging this PR.

pkg/sysregistriesv2/system_registries_v2.go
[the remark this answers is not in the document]
mtrmac: > The `test_com` issue did not satisfy me, so I had another look at net/url and found the following comment

FWIW the current `parseURL` still accepts `test_com`; the URL parser is not in the business of enforcing DNS formatting of host names.

As with the previous version, I think accepting that is good enough.

vrothberg:
> Man, that url.Parse use is some non-obvious code. @vrothberg , can you check that the comment added in the top commit of https://github.com/mtrmac/image/tree/sysregistriesv2-comments describes the situation correctly, please?

Yes, the comment is correct. I'll make sure to have more verbose comments for such wild cases in the future. As always, thanks a lot, @mtrmac!

vrothberg:
@baude @mheon @mtrmac @rhatdan @runcom @TomSweeneyRedHat (please add more to list if needed)

As you might know, my current plan is to port libpod to use sysregistriesv2 to allow the usage of mirrors. However, there is one thing I noticed that might hurt us in the future:

libpod is vendoring buildah, and both are using the sysregistries package (libpod provides even a wrapper package for it). This means that I'd need to update buildah first and then libpod (with vendoring in the updated buildah). It's not only my laziness speaking to avoid chain updates (now and in the future), but also from an architectural point of view:

**Does it make sense to refactor the pull/push code from buildah and libpod into c/images?**

buildah and libpod already use c/images, but what we need is an easier to use API with the notion of registries (and mirrors) as we're having with the sysregistriesv2 package. The API certainly requires some flexibility, and we might hit some rough edges here and there but we had the ultimate gain of fast delivery and reduced redundancy. Testing efforts could also be centralized in c/images.

@mtrmac, sorry for hijacking this PR to kickoff the discussion, but I couldn't figure out a better way.

mheon:
For my part, I'd like to see push/pull code move out into a library. We've already split it into a separate package in the `libpod` codebase to try and keep it separate from everything else, and I'd be happy to see it move completely out of our repo into a place where it can be shared with other projects more easily.

mtrmac:
> Does it make sense to refactor the pull/push code from buildah and libpod into c/images?

As mentioned earlier, I’d _love_ for c/image/docker to transparently handle the `Prefix`/`URL` mapping, mirrors support, and “insecure”/“blocked” configuration.

The one thing I’d like to currently keep out is using the search list for unqualified names, because as it is right now it breaks name matching in signature verification.  Or figure out some way to define reasonable the name matching semantics for the unqualified names.

mtrmac:
👍 Thanks again!

rhatdan:
I am fine with working on getting a library that could be shared, between the two tools.

vrothberg:
> The one thing I’d like to currently keep out is using the search list for unqualified names, because as it is right now it breaks name matching in signature verification. Or figure out some way to define reasonable name matching semantics for the unqualified names.

@mtrmac Can you elaborate a bit on this issue? Isn't the search list used already in both, Podman/libpod and Buildah?

mtrmac:
> > The one thing I’d like to currently keep out is using the search list for unqualified names, because as it is right now it breaks name matching in signature verification. Or figure out some way to define reasonable name matching semantics for the unqualified names.
>
> @mtrmac Can you elaborate a bit on this issue? Isn't the search list used already in both, Podman/libpod and Buildah?

Yes :(

Imprecisely, signature verification takes the image reference to be pulled, looks up the relevant public key, and then requires a signature with 1) that public key 2) matching the image reference.

The lookup uses a hierarchical namespace looking like Docker references, i.e. something like hostname[/ns/…[/repo]]. This namespace design does not fit the search behavior, where input can be an unqualified `fedora` (which does not mean “host name `fedora`”).

With the current search implementations, `podman`/`buildah` take the unqualified input (`fedora`), and iteratively try to prepend various search registries, and use that result as “the image reference to be pulled”: {`docker.io/`,`registry.fedoraproject.org`,`registry.access.redhat.com`}`/fedora` — which means that the signature checking expects the (search hostname + user input) pair to match, and that’s not what the user actually wanted.

So, suppose the search list is (`registry.fedoraproject.org`, `docker.io`), and the user pulls unqualified `fedora`.  Usually, `r.fp.org` is available, the image is pulled from there, and a signature is verified using the `r.fp.org/fedora` public key and expects the `r.fp.org/fedora` name to match.

But a MITM attacker can make `r.fp.org` unavailable, and the search will continue trying `docker.io/fedora`, and a signature will be verified using the `docker.io/fedora` public key and expect the `docker.io/fedora` name to match.  Or, if `docker.io` is configured not to require signatures, just making `r.fp.org` seem down allows the attacker to supply any arbitrary unsigned image and it will be accepted (assuming they can get `docker.io` to serve it).

(In a sense, the problem here is that for unqualified `fedora` the user genuinely does not know, or does not care, what server/identity/public key to use.  But that’s on us for designing the UI the way that the user can get away with not knowing/caring. I don’t know what the _best_ solution here is, maybe we should replace the general search list with an alias mechanism (so that `fedora` always means `r.fp.org/fedora`, whether or not the server is up), maybe the signature policy mechanism needs somehow extending. But until then I would prefer c/image not to accept such input and pretend to do signature verification when it is completely unclear what it is verifying.)

vrothberg:
As always, thanks a lot, @mtrmac for the details.

To be honest, I am not quite sure how to continue from here. I cannot start to port libpod to sysregsv2 unless buildah uses it as well, so I think breaking the push/pull code into a library makes sense at that point. But I don't think that we cannot *not* support the search registries as I consider that a really important usability feature. However, you have valid arguments against adding the push/pull API to containers/image (see above).

@mtrmac How could we proceed? I am also okay with porting Buildah first and leave the situation as is.

mtrmac:
> I cannot start to port libpod to sysregsv2 unless buildah uses it as well

(Why? The code can read the old format, so AFAICT as long the config file is not updated, the code can be updated in both independently.)

> But I don't think that we cannot not support the search registries as I consider that a really important usability feature.

\*shrug\* Yeah, I’m not fighting that in `libpod`/`buildah`.

> @mtrmac How could we proceed? I am also okay with porting Buildah first and leave the situation as is.

Hopefully not controversial:
- Teach `c/image/docker` to read `sysregistriesv2` for the “insecure” flag.
  - (We might also want to modify or augment `c/image/types.SystemContext.DockerInsecureSkipVerify` to be a `*bool` so that a caller can opt into full verification even if the config says “insecure”. Not sure.)
  - Then drop the `GetInsecureRegistries` callers from `libpod` (`buildah` currently doesn’t have any AFAICS).
- Implement the `URL`/`Prefix` mapping, and `Mirror` support of `sysregistriesv2` in `c/image/docker`. (No need to update `libpod`/`buildah`, these features are new and have no pre-existing equivalent.) (This assumes that callers will not want detailed control of the mirror search process; hopefully that is a reasonable assumption.)

As for the search behavior, either it can directly ported from the old to the new API in both packages without modification and kept separate, or the two can share it in some subpackage.  They are vendoring each other, so I’m not sure what is the best place—up to the maintainers of these repositories.

vrothberg:
> (Why? The code can read the old format, so AFAICT as long the config file is not updated, the code can be updated in both independently.)

That's right. What I meant was using mirrors actually.

> - Teach c/image/docker to read sysregistriesv2 for the “insecure” flag.
> [...]
> - Implement the URL/Prefix mapping and Mirror features of sysregistriesv2 in c/image/docker.

Sounds like a good plan to me, thanks for the proposal.

There is just one thing in the current design of sysregsv2 that I really want to change: currently, the API is built to be wrapped by another one (e.g., `FindUnqualifiedSerachRegistries` expects registries as an input). With the design you described above, the config is used at many different places, so to avoid continuously reloading the same config file, I'd like to change the API to a **singleton**. This aoids redundantly loading the cofig file, which would be expensive for daemons (e.g, CRI-O).

@mtrmac If you agree, I will prepare a bigger PR including your proposed changes (see above) and the change of v2 to a singleton design.

> As for the search behavior, either it can directly ported from the old to the new API in both packages without modification and kept separate, or the two can share it in some subpackage. They are vendoring each other, so I’m not sure what is the best place—up to the maintainers of these repositories.

Sounds good.

mtrmac:
> With the design you described above, the config is used at many different places, so to avoid continuously reloading the same config file, I'd like to change the API to a **singleton**. This avoids redundantly loading the cofig file, which would be expensive for daemons (e.g, CRI-O).

(Is it that expensive, if the kernel is likely to cache everything in memory? A quick benchmark on a SSD workstation shows about 71 us per load, when done in a tight loop, for me, which is negligible compared to the network accesses done by users of the data. I wouldn’t bother unless there was a known benchmark which suffered from this, but that’s no reason to stop you.)

Fine with me, as long as the implementation accounts for multi-threading and for multiple users using different config paths within the same process (i.e. a singleton lock-protected hash table or something like that); the support for different config paths via `types.SystemContext` serves as an alternative to detailed Golang controls overriding the config contents, so that needs to work.

Alternatively, the data could be cached within a `types.SystemContext` (maybe still with some protection against multi-threaded uses and path changes); most “serious” users are going to have a single long-term instance of `SystemContext`, and it is passed through the call stack (and unlike `context.Context`, a callee can modify it so that it affects the caller). OTOH it _is_ possible for `types.SystemContext` to be `nil`, and in that case the caching would be impossible or have to rely on the global variables.

vrothberg:
> (Is it that expensive, if the kernel is likely to cache everything in memory? A quick benchmark on a SSD workstation shows about 71 us per load, when done in a tight loop, for me, which is negligible compared to the network accesses done by users of the data. I wouldn’t bother unless there was a known benchmark which suffered from this, but that’s no reason to stop you.)

I think it doesn't hurt *per load* but when accumulating over time, I'd love my CPU to not waste those cycles. I think it's rather straight forward to implement and other users might benefit from it.

> Fine with me, as long as the implementation accounts for multi-threading and for multiple users using different config paths within the same process (i.e. a singleton lock-protected hash table or something like that); the support for different config paths via types.SystemContext serves as an alternative to detailed Golang controls overriding the config contents, so that needs to work.

I love this idea. As for multi-threading, I think the only critical section is writing the hash table (e.g., when reading a path for the first time in parallel). All following accesses are reads. Kudos for your patience discussing designs, much appreciated!

PR #466, opened 2018-06-12
Document the subtleties of using url.Parse for validating registry prefixes

mtrmac:
This follows up on #464 to clarify why/how we expect both `url.Parse(input)` and `url.parse("http://"+input)` to succeed.


mtrmac:
👍 

ACKed by @vrothberg in #464.

PR #467, opened 2018-06-19
sysregistriesv2: cache registries

vrothberg:
Cache previously loaded registries with the path to the specific config
file as the key.  Such kind of caching avoids redundantly loading the
config file.  The cache is thread safe and thus qualifies to be used
concurrently.

Signed-off-by: Valentin Rothberg <vrothberg@suse.com>

### DIFF
--- a/pkg/sysregistriesv2/system_registries_v2.go
+++ b/pkg/sysregistriesv2/system_registries_v2.go
@@ -6,6 +6,7 @@ import (
 	"net/url"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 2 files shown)


vrothberg:
@mtrmac I updated the comment and changed error control path in GetRegistries(). Thanks for the nice review!

mtrmac:
👍 Thanks!

rhatdan:
@runcom PTAL and merge.

rhatdan:
LGTM

PR #468, opened 2018-06-20
dockerClient: skip TLS verification if configured in registries

vrothberg:
Based on a discussion from: https://github.com/containers/image/pull/464#issuecomment-397744605

- sysregistriesv2: adjust API to SystemContext
- dockerClient: skip TLS verification if configured in registries

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/containers/image/docker/reference"
 	"github.com/containers/image/pkg/docker/config"
+	"github.com/containers/image/pkg/sysregistriesv2"
 	"github.com/containers/image/pkg/tlsclientconfig"
 	"github.com/containers/image/types"
 	"github.com/docker/distribution/registry/client"
@@ -78,11 +79,13 @@ type bearerToken struct {
 // dockerClient is configuration for dealing with a single Docker registry.
 type dockerClient struct {
 	// The following members are set by newDockerClient and do not change afterwards.
-	sys           *types.SystemContext
-	registry      string
+	sys                   *types.SystemContext
+	registry              string
+	client                *http.Client
+	insecureSkipTLSVerify bool
+	// The following members are not set by newDockerClient and must be set by callers if needed.
 	username      string
 	password      string
-	client        *http.Client
 	signatureBase signatureStorageBase
 	scope         authScope
 	// The following members are detected registry properties:
@@ -194,13 +197,26 @@ func newDockerClientFromRef(sys *types.SystemContext, ref dockerReference, write
 	if err != nil {
 		return nil, err
 	}
-	remoteName := reference.Path(ref.ref)
 
-	return newDockerClientWithDetails(sys, registry, username, password, actions, sigBase, remoteName)
+	client, err := newDockerClient(sys, registry, ref.ref.Name())
+	if err != nil {
+		return nil, err
+	}
+	client.username = username
+	client.password = password
+	client.signatureBase = sigBase
+	client.scope.actions = actions
+	client.scope.remoteName = reference.Path(ref.ref)
+	return client, nil
 }
 
-// newDockerClientWithDetails returns a new dockerClient instance for the given parameters
-func newDockerClientWithDetails(sys *types.SystemContext, registry, username, password, actions string, sigBase signatureStorageBase, remoteName string) (*dockerClient, error) {
+// newDockerClient returns a new dockerClient instance for the given registry
+// and reference.  The reference is used to query the registry configuration
+// and can either be a registry (e.g, "registry.com[:5000]"), a repository
+// (e.g., "registry.com[:5000][/some/namespace]/repo").
+// Please note that newDockerClient does not set all members of dockerClient
+// (e.g., username and password); those must be set by callers if necessary.
+func newDockerClient(sys *types.SystemContext, registry, reference string) (*dockerClient, error) {
 	hostName := registry
 	if registry == dockerHostname {
 		registry = dockerRegistry
@@ -221,33 +237,43 @@ func newDockerClientWithDetails(sys *types.SystemContext, registry, username, pa
 		return nil, err
 	}
 
-	if sys != nil && sys.DockerInsecureSkipTLSVerify {
-		tr.TLSClientConfig.InsecureSkipVerify = true
+	// Check if TLS verification shall be skipped (default=false) which can
+	// either be specified in the sysregistriesv2 configuration or via the
+	// SystemContext, whereas the SystemContext is prioritized.
+	skipVerify := false
+	if sys != nil && sys.DockerInsecureSkipTLSVerify != types.OptionalBoolUndefined {
+		// Only use the SystemContext if the actual value is defined.
+		skipVerify = sys.DockerInsecureSkipTLSVerify == types.OptionalBoolTrue
+	} else {
+		reg, err := sysregistriesv2.FindRegistry(sys, reference)
+		if err != nil {
+			return nil, errors.Wrapf(err, "error loading registries")
+		}
+		if reg != nil {
+			skipVerify = reg.Insecure
+		}
 	}
+	tr.TLSClientConfig.InsecureSkipVerify = skipVerify
 
 	return &dockerClient{
-		sys:           sys,
-		registry:      registry,
-		username:      username,
-		password:      password,
-		client:        &http.Client{Transport: tr},
-		signatureBase: sigBase,
-		scope: authScope{
-			actions:    actions,
-			remoteName: remoteName,
-		},
+		sys:                   sys,
+		registry:              registry,
+		client:                &http.Client{Transport: tr},
+		insecureSkipTLSVerify: skipVerify,
 	}, nil
 }
 
 // CheckAuth validates the credentials by attempting to log into the registry
 // returns an error if an error occcured while making the http request or the status code received was 401
 func CheckAuth(ctx context.Context, sys *types.SystemContext, username, password, registry string) error {
-	newLoginClient, err := newDockerClientWithDetails(sys, registry, username, password, "", nil, "")
+	client, err := newDockerClient(sys, registry, registry)
 	if err != nil {
 		return errors.Wrapf(err, "error creating new docker client")
 	}
+	client.username = username
+	client.password = password
 
-	resp, err := newLoginClient.makeRequest(ctx, "GET", "/v2/", nil, nil, v2Auth)
+	resp, err := client.makeRequest(ctx, "GET", "/v2/", nil, nil, v2Auth)
 	if err != nil {
 		return err
 	}
@@ -299,16 +325,21 @@ func SearchRegistry(ctx context.Context, sys *types.SystemContext, registry, ima
 		return nil, errors.Wrapf(err, "error getting username and password")
 	}
 
-	// The /v2/_catalog endpoint has been disabled for docker.io therefore the call made to that endpoint will fail
-	// So using the v1 hostname for docker.io for simplicity of implementation and the fact that it returns search results
+	// The /v2/_catalog endpoint has been disabled for docker.io therefore
+	// the call made to that endpoint will fail.  So using the v1 hostname
+	// for docker.io for simplicity of implementation and the fact that it
+	// returns search results.
+	hostname := registry
 	if registry == dockerHostname {
-		registry = dockerV1Hostname
+		hostname = dockerV1Hostname
 	}
 
-	client, err := newDockerClientWithDetails(sys, registry, username, password, "", nil, "")
+	client, err := newDockerClient(sys, hostname, registry)
 	if err != nil {
 		return nil, errors.Wrapf(err, "error creating new docker client")
 	}
+	client.username = username
+	client.password = password
 
 	// Only try the v1 search endpoint if the search query is not empty. If it is
 	// empty skip to the v2 endpoint.
@@ -530,7 +561,7 @@ func (c *dockerClient) detectProperties(ctx context.Context) error {
 		return nil
 	}
 	err := ping("https")
-	if err != nil && c.sys != nil && c.sys.DockerInsecureSkipTLSVerify {
+	if err != nil && c.insecureSkipTLSVerify {
 		err = ping("http")
 	}
 	if err != nil {
@@ -554,7 +585,7 @@ func (c *dockerClient) detectProperties(ctx context.Context) error {
 			return true
 		}
 		isV1 := pingV1("https")
-		if !isV1 && c.sys != nil && c.sys.DockerInsecureSkipTLSVerify {
+		if !isV1 && c.insecureSkipTLSVerify {
 			isV1 = pingV1("http")
 		}
 		if isV1 {
--- a/pkg/sysregistriesv2/system_registries_v2.go
+++ b/pkg/sysregistriesv2/system_registries_v2.go
@@ -368,7 +368,7 @@ func FindRegistry(ctx *types.SystemContext, ref string) (*Registry, error) {
 	reg := Registry{}
 	prefixLen := 0
 	for _, r := range registries {
-		if strings.HasPrefix(ref, r.Prefix) {
+		if strings.HasPrefix(ref, r.Prefix+"/") || ref == r.Prefix {
 			length := len(r.Prefix)
 			if length > prefixLen {
 				reg = r
--- a/pkg/sysregistriesv2/system_registries_v2.go
+++ b/pkg/sysregistriesv2/system_registries_v2.go
@@ -3,7 +3,6 @@ package sysregistriesv2
 import (
 	"fmt"
 	"io/ioutil"
-	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
@@ -83,8 +82,8 @@ func (e *InvalidRegistries) Error() string {
 }
 
 // parseURL parses the input string, performs some sanity checks and returns
-// the sanitized input string.  An error is returned in case parsing fails or
-// or if URI scheme or user is set.
+// the sanitized input string.  An error is returned if the input string is
+// empty or if contains an "http{s,}://" prefix.
 func parseURL(input string) (string, error) {
 	trimmed := strings.TrimRight(input, "/")
 
@@ -92,46 +91,8 @@ func parseURL(input string) (string, error) {
 		return "", &InvalidRegistries{s: "invalid URL: cannot be empty"}
 	}
 
-	// Ultimately, we expect input of the form example.com[/namespace/…], a prefix
-	// of a fully-expended reference (containers/image/docker/Reference.String()).
-	// c/image/docker/Reference does not currently provide such a parser.
-	// So, we use url.Parse("http://"+trimmed) below to ~verify the format, possibly
-	// letting some invalid input in, trading that off for a simpler parser.
-	//
-	// url.Parse("http://"+trimmed) is, sadly, too permissive, notably for
-	// trimmed == "http://example.com/…", url.Parse("http://http://example.com/…")
-	// is accepted and parsed as
-	// {Scheme: "http", Host: "http:", Path: "//example.com/…"}.
-	//
-	// So, first we do an explicit check for an unwanted scheme prefix:
-
-	// This will parse trimmed=="http://example.com/…" with Scheme: "http".  Perhaps surprisingly,
-	// it also succeeds for the input we want to accept, in different ways:
-	// "example.com" -> {Scheme:"", Opaque:"", Path:"example.com"}
-	// "example.com/repo" -> {Scheme:"", Opaque:"", Path:"example.com/repo"}
-	// "example.com:5000" -> {Scheme:"example.com", Opaque:"5000"}
-	// "example.com:5000/repo" -> {Scheme:"example.com", Opaque:"5000/repo"}
-	uri, err := url.Parse(trimmed)
-	if err != nil {
-		return "", &InvalidRegistries{s: fmt.Sprintf("invalid URL '%s': %v", input, err)}
-	}
-
-	// Check if a URI Scheme is set.
-	// Note that URLs that do not start with a slash after the scheme are
-	// interpreted as `scheme:opaque[?query][#fragment]`; see above for examples.
-	if uri.Scheme != "" && uri.Opaque == "" {
+	if strings.HasPrefix(trimmed, "http://") || strings.HasPrefix(trimmed, "https://") {
 		msg := fmt.Sprintf("invalid URL '%s': URI schemes are not supported", input)
-		return "", &InvalidRegistries{s: msg}
-	}
-
-	uri, err = url.Parse("http://" + trimmed)
-	if err != nil {
-		msg := fmt.Sprintf("invalid URL '%s': sanitized URL did not parse: %v", input, err)
-		return "", &InvalidRegistries{s: msg}
-	}
-
-	if uri.User != nil {
-		msg := fmt.Sprintf("invalid URL '%s': user/password are not supported", trimmed)
 		return "", &InvalidRegistries{s: msg}
 	}
 
--- a/pkg/sysregistriesv2/system_registries_v2.go
+++ b/pkg/sysregistriesv2/system_registries_v2.go
@@ -280,7 +280,18 @@ var configMutex = sync.Mutex{}
 // are synchronized via configMutex.
 var configCache = make(map[string][]Registry)
 
+// InvalidateCache invalidates the registry cache.  This function is meant to be
+// used for long-running processes that need to reload potential changes made to
+// the cached registry config files.
+func InvalidateCache() {
+	configMutex.Lock()
+	defer configMutex.Unlock()
+	configCache = make(map[string][]Registry)
+}
+
 // GetRegistries loads and returns the registries specified in the config.
+// Note the parsed content of registry config files is cached.  For reloading,
+// use `InvalidateCache` and re-call `GetRegistries`.
 func GetRegistries(ctx *types.SystemContext) ([]Registry, error) {
 	configPath := getConfigPath(ctx)
 
--- a/pkg/sysregistriesv2/system_registries_v2.go
+++ b/pkg/sysregistriesv2/system_registries_v2.go
@@ -323,19 +323,29 @@ func GetRegistries(ctx *types.SystemContext) ([]Registry, error) {
 
 // FindUnqualifiedSearchRegistries returns all registries that are configured
 // for unqualified image search (i.e., with Registry.Search == true).
-func FindUnqualifiedSearchRegistries(registries []Registry) []Registry {
+func FindUnqualifiedSearchRegistries(ctx *types.SystemContext) ([]Registry, error) {
+	registries, err := GetRegistries(ctx)
+	if err != nil {
+		return nil, err
+	}
+
 	unqualified := []Registry{}
 	for _, reg := range registries {
 		if reg.Search {
 			unqualified = append(unqualified, reg)
 		}
 	}
-	return unqualified
+	return unqualified, nil
 }
 
 // FindRegistry returns the Registry with the longest prefix for ref.  If no
 // Registry prefixes the image, nil is returned.
-func FindRegistry(ref string, registries []Registry) *Registry {
+func FindRegistry(ctx *types.SystemContext, ref string) (*Registry, error) {
+	registries, err := GetRegistries(ctx)
+	if err != nil {
+		return nil, err
+	}
+
 	reg := Registry{}
 	prefixLen := 0
 	for _, r := range registries {
@@ -348,9 +358,9 @@ func FindRegistry(ref string, registries []Registry) *Registry {
 		}
 	}
 	if prefixLen != 0 {
-		return &reg
+		return &reg, nil
 	}
-	return nil
+	return nil, nil
 }
 
 // Reads the global registry file from the filesystem. Returns a byte array.
--- a/pkg/sysregistriesv2/system_registries_v2.go
+++ b/pkg/sysregistriesv2/system_registries_v2.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/url"
+	"os"
 	"path/filepath"
 	"strings"
 	"sync"
@@ -293,6 +294,13 @@ func GetRegistries(ctx *types.SystemContext) ([]Registry, error) {
 	// load the config
 	config, err := loadRegistryConf(configPath)
 	if err != nil {
+		// Return an empty []Registry if we use the default config,
+		// which implies that the config path of the SystemContext
+		// isn't set.  Note: if ctx.SystemRegistriesConfPath points to
+		// the default config, we will still return an error.
+		if os.IsNotExist(err) && (ctx == nil || ctx.SystemRegistriesConfPath == "") {
+			return []Registry{}, nil
+		}
 		return nil, err
 	}
 
--- a/pkg/sysregistriesv2/system_registries_v2_test.go
+++ b/pkg/sysregistriesv2/system_registries_v2_test.go
@@ -25,8 +25,7 @@ func TestParseURL(t *testing.T) {
 	assert.Contains(t, err.Error(), "invalid URL 'https://example.com': URI schemes are not supported")
 
 	_, err = parseURL("john.doe@example.com")
-	assert.NotNil(t, err)
-	assert.Contains(t, err.Error(), "invalid URL 'john.doe@example.com': user/password are not supported")
+	assert.Nil(t, err)
 
 	// valid URLs
 	url, err = parseURL("example.com")
--- a/pkg/sysregistriesv2/system_registries_v2_test.go
+++ b/pkg/sysregistriesv2/system_registries_v2_test.go
@@ -76,8 +76,8 @@ blocked = true`)
 	assert.Nil(t, err)
 	assert.Equal(t, 2, len(registries))
 
-	var reg *Registry
-	reg = FindRegistry("registry.com/image:tag", registries)
+	reg, err := FindRegistry(nil, "registry.com/image:tag")
+	assert.Nil(t, err)
 	assert.NotNil(t, reg)
 	assert.Equal(t, 2, len(reg.Mirrors))
 	assert.Equal(t, "mirror-1.registry.com", reg.Mirrors[0].URL)
@@ -147,23 +147,26 @@ prefix = ""`)
 	assert.Nil(t, err)
 	assert.Equal(t, 5, len(registries))
 
-	var reg *Registry
-	reg = FindRegistry("simple-prefix.com/foo/bar:latest", registries)
+	reg, err := FindRegistry(nil, "simple-prefix.com/foo/bar:latest")
+	assert.Nil(t, err)
 	assert.NotNil(t, reg)
 	assert.Equal(t, "simple-prefix.com", reg.Prefix)
 	assert.Equal(t, reg.URL, "registry.com:5000")
 
-	reg = FindRegistry("complex-prefix.com:4000/with/path/and/beyond:tag", registries)
+	reg, err = FindRegistry(nil, "complex-prefix.com:4000/with/path/and/beyond:tag")
+	assert.Nil(t, err)
 	assert.NotNil(t, reg)
 	assert.Equal(t, "complex-prefix.com:4000/with/path", reg.Prefix)
 	assert.Equal(t, "another-registry.com:5000", reg.URL)
 
-	reg = FindRegistry("no-prefix.com/foo:tag", registries)
+	reg, err = FindRegistry(nil, "no-prefix.com/foo:tag")
+	assert.Nil(t, err)
 	assert.NotNil(t, reg)
 	assert.Equal(t, "no-prefix.com", reg.Prefix)
 	assert.Equal(t, "no-prefix.com", reg.URL)
 
-	reg = FindRegistry("empty-prefix.com/foo:tag", registries)
+	reg, err = FindRegistry(nil, "empty-prefix.com/foo:tag")
+	assert.Nil(t, err)
 	assert.NotNil(t, reg)
 	assert.Equal(t, "empty-prefix.com", reg.Prefix)
 	assert.Equal(t, "empty-prefix.com", reg.URL)
@@ -201,7 +204,8 @@ unqualified-search = true
 	assert.Nil(t, err)
 	assert.Equal(t, 4, len(registries))
 
-	unqRegs := FindUnqualifiedSearchRegistries(registries)
+	unqRegs, err := FindUnqualifiedSearchRegistries(nil)
+	assert.Nil(t, err)
 	assertSearchRegistryURLsEqual(t, []string{"registry-a.com", "registry-c.com", "registry-d.com"}, unqRegs)
 }
 
@@ -301,11 +305,13 @@ registries = ["registry-d.com", "registry-e.com", "registry-a.com"]`)
 	assert.Nil(t, err)
 	assert.Equal(t, 5, len(registries))
 
-	unqRegs := FindUnqualifiedSearchRegistries(registries)
+	unqRegs, err := FindUnqualifiedSearchRegistries(nil)
+	assert.Nil(t, err)
 	assertSearchRegistryURLsEqual(t, []string{"registry-a.com", "registry-c.com", "registry-d.com"}, unqRegs)
 
 	// check if merging works
-	reg := FindRegistry("registry-a.com/bar/foo/barfoo:latest", registries)
+	reg, err := FindRegistry(nil, "registry-a.com/bar/foo/barfoo:latest")
+	assert.Nil(t, err)
 	assert.NotNil(t, reg)
 	assert.True(t, reg.Search)
 	assert.True(t, reg.Insecure)
--- a/pkg/sysregistriesv2/system_registries_v2_test.go
+++ b/pkg/sysregistriesv2/system_registries_v2_test.go
@@ -153,6 +153,21 @@ prefix = ""`)
 	assert.Equal(t, "simple-prefix.com", reg.Prefix)
 	assert.Equal(t, reg.URL, "registry.com:5000")
 
+	// path match
+	reg, err = FindRegistry(nil, "simple-prefix.com/")
+	assert.Nil(t, err)
+	assert.NotNil(t, reg)
+
+	// hostname match
+	reg, err = FindRegistry(nil, "simple-prefix.com")
+	assert.Nil(t, err)
+	assert.NotNil(t, reg)
+
+	// invalid match
+	reg, err = FindRegistry(nil, "simple-prefix.comx")
+	assert.Nil(t, err)
+	assert.Nil(t, reg)
+
 	reg, err = FindRegistry(nil, "complex-prefix.com:4000/with/path/and/beyond:tag")
 	assert.Nil(t, err)
 	assert.NotNil(t, reg)
--- a/pkg/sysregistriesv2/system_registries_v2_test.go
+++ b/pkg/sysregistriesv2/system_registries_v2_test.go
@@ -386,3 +386,47 @@ insecure = true`)
 	assert.Nil(t, err)
 	assert.Equal(t, 4, len(registries))
 }
+
+func TestInvalidateCache(t *testing.T) {
+	testConfig = []byte(`
+[[registry]]
+url = "registry.com"
+
+[[registry.mirror]]
+url = "mirror-1.registry.com"
+
+[[registry.mirror]]
+url = "mirror-2.registry.com"
+
+
+[[registry]]
+url = "blocked.registry.com"
+blocked = true
+
+
+[[registry]]
+url = "insecure.registry.com"
+insecure = true
+
+
+[[registry]]
+url = "untrusted.registry.com"
+insecure = true`)
+
+	ctx := &types.SystemContext{}
+
+	configCache = make(map[string][]Registry)
+	registries, err := GetRegistries(ctx)
+	assert.Nil(t, err)
+	assert.Equal(t, 4, len(registries))
+	assertSearchRegistryURLsEqual(t, []string{"registry.com", "blocked.registry.com", "insecure.registry.com", "untrusted.registry.com"}, registries)
+
+	// invalidate the cache, make sure it's empty and reload
+	InvalidateCache()
+	assert.Equal(t, 0, len(configCache))
+
+	registries, err = GetRegistries(ctx)
+	assert.Nil(t, err)
+	assert.Equal(t, 4, len(registries))
+	assertSearchRegistryURLsEqual(t, []string{"registry.com", "blocked.registry.com", "insecure.registry.com", "untrusted.registry.com"}, registries)
+}
--- a/types/types.go
+++ b/types/types.go
@@ -324,6 +324,30 @@ type DockerAuthConfig struct {
 	Password string
 }
 
+// OptionalBool is a boolean with an additional undefined value, which is meant
+// to be used in the context of user input to distinguish between a
+// user-specified value and a default value.
+type OptionalBool byte
+
+const (
+	// OptionalBoolUndefined indicates that the OptionalBoolean hasn't been written.
+	OptionalBoolUndefined OptionalBool = iota
+	// OptionalBoolTrue represents the boolean true.
+	OptionalBoolTrue
+	// OptionalBoolFalse represents the boolean false.
+	OptionalBoolFalse
+)
+
+// NewOptionalBool converts the input bool into either OptionalBoolTrue or
+// OptionalBoolFalse.  The function is meant to avoid boilerplate code of users.
+func NewOptionalBool(b bool) OptionalBool {
+	o := OptionalBoolFalse
+	if b == true {
+		o = OptionalBoolTrue
+	}
+	return o
+}
+
 // SystemContext allows parameterizing access to implicitly-accessed resources,
 // like configuration files in /etc and users' login state in their home directory.
 // Various components can share the same field only if their semantics is exactly
@@ -376,7 +400,7 @@ type SystemContext struct {
 	// Ignored if DockerCertPath is non-empty.
 	DockerPerHostCertDirPath string
 	// Allow contacting docker registries over HTTP, or HTTPS with failed TLS verification. Note that this does not affect other TLS connections.
-	DockerInsecureSkipTLSVerify bool
+	DockerInsecureSkipTLSVerify OptionalBool
 	// if nil, the library tries to parse ~/.docker/config.json to retrieve credentials
 	DockerAuthConfig *DockerAuthConfig
 	// if not "", an User-Agent header is added to each request when contacting a registry.


vrothberg:
@mtrmac The API change hasn't been discussed but I think it makes the API easier to use and ultimately avoids boilerplate code for its users.

docker/docker_client.go
```diff
@@ -208,8 +209,19 @@ func newDockerClientWithDetails(sys *types.SystemContext, registry, username, pa
 		return nil, err
 	}
 
+	// If specified in the SystemContext, use its InsecureSkipVerify flag.
+	// Otherwise, query if the remoteName has a correspoding registry in
+	// the registry config and set the flag accordingly.
 	if sys != nil && sys.DockerInsecureSkipTLSVerify {
 		tr.TLSClientConfig.InsecureSkipVerify = true
+	} else if sys != nil {
```
TomSweeneyRedHat: maybe I need more tea, but if the user had set DockerInsecureSkipTLSVerify to false, meaning to use tls-verify, should we be potentially setting InsecureSkipVerify to true if that's the setting in reg.Insecure?

vrothberg: The idea behind this code is to allow users of the API to opt into full verification even if the `registries.conf` sets the given registry to be insecure (see discussion at https://github.com/containers/image/pull/464#issuecomment-397744605).

@mtrmac proposed this semantic and he usually has a good intuition when it comes to such details. The nasty thing with booleans is that there is no way to check if a given item is "unset". Some tristate would be much nicer to resolve such issues.

rhatdan: Right if I as a user of podman specify --tlsverify=TRUE, that means ignore the settings in the registries.conf file and only use https.

pkg/sysregistriesv2/system_registries_v2_test.go
```diff
@@ -76,7 +76,8 @@ blocked = true`)
 	assert.Equal(t, 2, len(registries))
 
 	var reg *Registry
-	reg = FindRegistry("registry.com/image:tag", registries)
+	reg, err = FindRegistry(nil, "registry.com/image:tag")
+	assert.Nil(t, err)
 	assert.NotNil(t, reg)
```
mtrmac: (Some of these should be `require.` so that future asserts don’t fail with `nil` references. But that’s trivial to retrofit if the tests ever fail, so not blocking.

A fair amount of this would probably be smaller if the tests were table-driven, but again, this works just fine, absolutely non-blocking.)

pkg/sysregistriesv2/system_registries_v2_test.go
```diff
@@ -186,15 +191,15 @@ unqualified-search = true`)
 	assert.Nil(t, err)
 	assert.Equal(t, 3, len(registries))
 
-	unqRegs := FindUnqualifiedSearchRegistries(registries)
+	var unqRegs []Registry
```
mtrmac: (Non-blocking: Is the declaration necessary? I thought `unqregs, err := Find…` works if one of the variables is new.)

docker/docker_client.go
mtrmac: > The idea behind this code is to allow users of the API to opt into full verification even if the `registries.conf` sets the given registry to be insecure

I can’t see how this code is doing that, unless you mean “by setting `sys == nil`”; but `types.SystemContext` documentation says that `nil` and `&types.SystemContext{}` must be treated equivalently everywhere, so that conditional would not be appropriate.

If we need a tri-state, then let’s actually set one up.  Probably an `enum` undefined/insecure/secure instead of a `*bool`—and enum will require changes in all users, but it’s easier to set an enum than a `*bool` in a struct initializer.

Or maybe add a `bool DockerForceSecureTLSVerification`, and fail if both this and `DockerInsecureSkipTLSVerify` is set — that would be compatible, but it is _very_ ugly.

But…

---

… I’m not actually convinced we _need_ this “force verification on even if the config turns it off” override; I brought it up primarily because the `--tls-verify` behavior of libpod/buildah works like that.  In which situations would the flag be used like that, though?

The only use I can think of is setting up certificates for a previously insecure registry, to check whether it is safe to change the configuration; in that case it would be better to create the new configuration, and test with `SystemContext.SystemRegistriesConfPath` pointing to that new configuration file; that tests _both_ the TLS setup of the registry and correctness of the contemplated file edits.

(It is specifically not very reasonable for the machine admin to set a registry to insecure, and one user using `--tls-verify=true` because the user disagrees, for whatever reason: it’s difficult to imagine what reason that could be, and it’s way too easy to forget to use `--tls-verify=true`, or to run a third-party script which does not use it.)

---

Yet again, if the flag syntax is `--tls-verify=false`, then `--tls-verify=true` needs to do _something_, and doing no verification (when the config file says so) would seem pretty surprising.

.travis.Dockerfile
```diff
@@ -5,6 +5,8 @@ RUN apt-get -qq update && \
 
 ADD https://storage.googleapis.com/golang/go1.9.2.linux-amd64.tar.gz /tmp
 
+COPY registries.conf /etc/containers/registries.conf
```
mtrmac: Hum, do we want c/image to _require_ that file to be present?  It seems reasonable to me for `sysregistriesv2` to silently use an empty configuration if the file does not exist (but still fail if it is unreadable for other reasons, e.g. permission or I/O error).

OTOH maybe the file will always be present in practice and this is not worth worrying about; @rhatdan ?

docker/docker_client.go
vrothberg: > I can’t see how this code is doing that, unless you mean “by setting sys == nil”; but types.SystemContext documentation says that nil and &types.SystemContext{} must be treated equivalently everywhere, so that conditional would not be appropriate.

Good catch, thanks!

> Yet again, if the flag syntax is --tls-verify=false, then --tls-verify=true needs to do something, and doing no verification (when the config file says so) would seem pretty surprising.

Given Skopeo has a similar flag for source and destination registries, I feel that we should follow the tri-state idea. Then we can also error out if the `registries.conf` and SystemContext{} are in conflict.

@mtrmac If you agree, I will update the PR accordignly. I'd make the tri-state generic with /undefined/set/unset/ to make it applicable for other use-cases.

mtrmac: > Given Skopeo has a similar flag for source and destination registries, I feel that we should follow the tri-state idea.

ACK.

> Then we can also error out if the registries.conf and SystemContext{} are in conflict.

So far CLI flags, and therefore `SystemContext`, have been overriding the configuration files; I’d prefer to keep that model instead of failing on inconsistency.

I can understand the argument for forbidding `--tls-verify=false` if the config file says to use TLS, but we have had the option without the semantics, so changing it is a bit unattractive. Also, we still allow callers to change the config file path, so enforcing consistency would not give the system admin any reliable enforcement mechanism.

As for forbidding `--tls-verify=true` if the config file says not to use TLS, that would entirely negate the point of introducing the tri-state mechanism: `true` is a default for registries not in the config file, and for registries in the config file, the `--tls-verify` value would have to match, so `--tls-verify=true` could never be used in a way that would make any difference.

.travis.Dockerfile
rhatdan: No we should make that file optional.

docker/docker_client.go
vrothberg: > So far CLI flags, and therefore SystemContext, have been overriding the configuration files; I’d prefer to keep that model instead of failing on inconsistency.

That makes things much nicer and will delay a heart attack by many years.

### toml and custom types
I have been looking into the tri-state idea and it turns out that toml makes it a bit hard/impossible to implement a tri-state type the way I imagined it (and @mtrmac I think as well). By all means, I wanted to allow users to keep using `true` and `false` in the configuration and to map it to a `type TriState int` when unmarshalling.

It turns out that's not possible with toml (and the available toml go libs). Both, `true` and `false` are always mapped to a `bool` regardless if we specified an unmarshal method for `TriState` and hence always leads to a type-conversion error (bool <-> int). I've been trying hard and skimmed old issues etc. and came to the conclusion that there is no user-friendly way to make a simple TriState type that can be used in toml, CLI flags, etc. Note, that this is possible when using `encoding.json` though. There's more such peculiarities when using toml (int <-> float when "." is missing etc.).

That being sad...
> So far CLI flags, and therefore SystemContext, have been overriding the configuration files; I’d prefer to keep that model instead of failing on inconsistency.

... makes the upper issue much simpler as we only have to care about `SystemContext.DockerInsecureSkipTLSVerify` and how we can check if the default or a user-specified value is used.

@mtrmac Hence, I suggest to make only `DockerInsecureSkipTLSVerify` a `TriState` or `TriBool` and leave the configuration as is. WDYT? (I could have updated the commit with those changes, but I figured that the upper issue is worth discussing before.)

mtrmac: > @mtrmac Hence, I suggest to make only DockerInsecureSkipTLSVerify a TriState or TriBool and leave the configuration as is. 

I agree.  (`OptionalBool` maybe?)

We genuinely have three kinds of semantics for the CLI options: “force `true`” / “force `false`” / “use configuration files”.

We don’t _really_ have three kinds of semantics for the configuration: we have “configured `true`” / “configured `false`” / “use default”, but immediately after parsing the configuration, “use default” amounts to the same thing as “configured `false`”; we don’t need to represent “use default” anywhere outside parsing a single `Registry` / `Mirror` entry (and even in there, probably only implicitly).

vrothberg: > I agree. (OptionalBool maybe?)
+1

vrothberg:
```
cmd/skopeo/utils.go:21:30: cannot use !c.GlobalBoolT("tls-verify") (type bool) as type types.OptionalBool in field value
cmd/skopeo/utils.go:31:35: cannot use !c.BoolT(flagPrefix + "tls-verify") (type bool) as type types.OptionalBool in assignment
```

Tests fail expectedly as skopeo needs to be updated. I suggest to merge this PR, then update Skopeo and after that I'll take care of the pull code.

docker/docker_client.go
```diff
@@ -533,7 +548,7 @@ func (c *dockerClient) detectProperties(ctx context.Context) error {
 			return true
 		}
 		isV1 := pingV1("https")
-		if !isV1 && c.sys != nil && c.sys.DockerInsecureSkipTLSVerify {
+		if !isV1 && c.sys != nil && c.sys.DockerInsecureSkipTLSVerify == types.OptionalBoolTrue {
```
mtrmac: The two uses of `DockerInsecureSkipTLSVerify` should use `sysregistriesv2` as well (presumably looking up the config only once).

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -285,6 +286,13 @@ func GetRegistries(ctx *types.SystemContext) ([]Registry, error) {
 	// load the config
 	config, err := loadRegistryConf(configPath)
 	if err != nil {
+		// Return an empty []Registry if we use the default config,
+		// which implies that the config path of the SystemContext
+		// isn't set.  Note: if ctx.SystemRegistriesConfPath points to
+		// the default config, we will return an error.
+		if os.IsNotExist(err) && ctx != nil && ctx.SystemRegistriesConfPath == "" {
```
mtrmac: `ctx == nil` and `ctx != nil && ctx.SystemRegistriesConfPath` should be treated equivalently.

So, `os.IsNotExist(err) && (ctx == nil || ctx.SystemRegistriesConfPath == "")` I think.

(BTW, reporting an error on an explicitly supplied path is a thoughtful nuance. Nice.)

docker/docker_client.go
mtrmac: (only once == when creating the `dockerClient`, not once in `newDockerClient…` and once in `detectProperties`.)

docker/docker_client.go
```diff
@@ -208,9 +209,23 @@ func newDockerClientWithDetails(sys *types.SystemContext, registry, username, pa
...
-		tr.TLSClientConfig.InsecureSkipVerify = true
+	// Check if TLS verification shall be skipped (default=false) which can
+	// either be specified in the sysregistriesv2 configuration or via the
+	// SystemContext, whereas the SystemContext is prioritized.
+	skipVerify := false
+	if sys != nil && sys.DockerInsecureSkipTLSVerify != types.OptionalBoolUndefined {
+		// Only use the SystemContext if the actual value is defined.
+		skipVerify = sys.DockerInsecureSkipTLSVerify == types.OptionalBoolTrue
+	} else {
+		reg, err := sysregistriesv2.FindRegistry(sys, remoteName)
```
mtrmac: `remoteName` is `reference.Path(ref)`, i.e. something like `library/busybox`; or it may be `""` with other callers.

I assume this should end up with `ref.Name()` when called by `newDockerClientFromRef`.  But there is also `CheckAuth` and `SearchRegistry`, which only have the hostname.

And, worse, for `docker.io` we have three different strings (`dockerHostname`, `dockerV1Hostname`, and `dockerRegistry`) in various variables, at various points in various code paths.

I’m not sure how to handle this entirely cleanly. Yet another parameter to `newDockerClientWithDetails`?

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: you're right, good catch!

docker/docker_client.go
vrothberg: A shame I didn't see this. Fixed now.

docker/docker_client.go
vrothberg: > And, worse, for docker.io we have three different strings (dockerHostname, dockerV1Hostname, and dockerRegistry) in various variables, at various points in various code paths.

I am not sure how we should deal with that. There is a varying degree of craziness in how Docker deals with. `index.docker.io` is mapped to `docker.io` while `registry-1.docker.io` is treated differently (with respect to the RepoTag of the image) and all effectively point to the same image.

```terminal
$ docker pull busybox
$ docker pull library/busybox
$ docker pull docker.io/busybox
$ docker pull docker.io/library/busybox
$ docker pull index.docker.io/busybox
$ docker pull index.docker.io/library/busybox
$ docker pull registry-1.docker.io/library/busybox 

$ docker images
REPOSITORY                             TAG      IMAGE ID [...]
busybox                                latest   f9b6f7f7b9d3 [...]
registry-1.docker.io/library/busybox   latest   f9b6f7f7b9d3 [...]
```

I hope this is something we can figure out at a later point and put it on our todo list?

> I’m not sure how to handle this entirely cleanly. Yet another parameter to newDockerClientWithDetails?

I fear it depends on how we want to deal with `Registry.Prefix` and how we "expect" a registry to be configured. Let's consider the following example (simplified config):
```
url="example.com"
prefix="example.com/foo"
insecure=true
```
If we use `FindRegistry("example.com/bar")` what is the expected outcome? Currently, it wouldn't return any `Registry` as the prefix doesn't match although a user might expect example.com is implicitly set to be insecure? I think there are various (valid?) interpretations of it and I don't think that's good.

I think that TLS verification should always be host/registry/url-based but that doesn't work with the current design of sysregsv2, prefixes in particular, so I want to propose a change that should solve the upper problem.

### Make Prefixes a sub-table
An exemplary config:
```toml
[[registry]]
url = "registry.com"
insecure = true
unqualified-search = true
[[registry.mirror]]
url = "mirror-1.registry.com"
[[registry.prefix]]
path="/bar"
prefix="registry.com/foo"
[[registry.prefix]]
path="/paste"
prefix="registry.com/copy"
trim=true
[[registry.prefix]]
prefix="another-registry.com:5000"
```

The idea is to detach the prefix (and namespace translation) from the registry to a sub-table as already done with mirrors. With the upper example:

- `pull registry.com/foo/image:latest` -> `registry.com/bar/foo/image:latest`
- `pull registry.com/copy/image:latest` -> `registry.com/paste/image:latest` (the prefix is trimmed  before being prepended)
- `pull another-registry.com:5000/image:latest` -> `registry.com/image:latest` (no path specified -> `path="/"`)

@mtrmac This would mean yet another change to `sysregsv2` but, to be honest, I haven't thought about this particular problem. We can add `registry.prefix.trim` at a later point but it should demonstrate that we are more flexible with prefixes being a sub-table. This way, the sysregsv2 configuration is way less ambiguous than before, and we have clear separation between registry-specific config switches and the namespace translations induced by prefixes.

docker/docker_client.go
mtrmac: >  Fixed now.

(I still see the same version of the code.)

docker/docker_client.go
mtrmac: ### Hostname Mapping

>> And, worse, for docker.io we have three different strings (`dockerHostname, `dockerV1Hostname`, and `dockerRegistry`) in various variables, at various points in various code paths.
>
> I am not sure how we should deal with that. There is a varying degree of craziness in how Docker deals with that. `index.docker.io` is mapped to `docker.io` while `registry-1.docker.io` is treated differently (with respect to the RepoTag of the image) and all effectively point to the same image.

It’s not _that_ bad actually.  For _images_ we support only the syntax `docker.io/*` (via `dockerClientFromRef` and `c/image/docker/reference.ParseNormalizedNamed`, which may make the `docker.io[/library]` prefixes implicit). I don’t think we need to normalize image names to support any of the other values; _users_ always say `docker.io` (or nothing when it is implied), the other `docker.io` hostnames are only set inside this one file.

The underlying `registry-1.docker.io` host name is mapped inside `newDockerClientWithDetails`, so handling that case is just a matter of being careful to use the right variable when calling `FindRegistry`, or something like that.

The V1 `index.docker.io` hostname is used only in the `SearchRegistry` case, which, like `CheckAuth`, works with hostnames only and does not have the `namespace…/repo` subnamespaces.

So, just making sure that `newDockerClientWithDetails` actually looks up the correct registry namespace for all three callers will force the function interface to somehow represent either a hostname-only or a hostname/path input to `FindRegistry`, and at that point the V1/`index.`/`docker.io` distinctions will be reasonably easy to handle.

---

### “Insecure” Flag Duplication.

> If we use FindRegistry("example.com/bar") what is the expected outcome? Currently, it wouldn't return any Registry as the prefix doesn't match although a user might expect example.com is implicitly set to be insecure? I think there are various (valid?) interpretations of it and I don't think that's good.

I’m not _too_ worried about this; for the most part, either a registry hostname is “real” (an underlying server that we want to reach), in which case there will be no configuration or at most a single entry for the whole server; or the hostname is intended to not be used directly, and is mirrored locally (e.g. a local mirror of a tiny subset of `docker.io`, with `Prefix` entries `docker.io/$repo` and `URL` or `Mirrors` pointing to the local mirror), in which case there may be many entries but possibly pointing at different mirrors, and there’s no reason to assume that all the mirrors will use TLS in the same way.

I.e. I’d expect that there is either no entry for `example.com[/anything]`, a single entry for all of `example.com`, or several entries for `example.com/foo` which are likely to match; a request for `example.com/bar` which does not match, but with an existing `example.com/foo` configuration seems unlikely to me.

Still, sure, not repeating the same boolean many times for a single hostname would be simpler, especially when trying to _turn off_ the `insecure` flag.

---

### Prefix Trimming

> We can add `registry.prefix.trim` at a later point but it should demonstrate that we are more flexible with prefixes being a sub-table. 

I don’t think that’s actually necessary; we can always trim the prefix from the match, and the user can undo that by repeating the trimmed part in the URL (i.e. configure `path=/bar/foo` in the `prefix=registry.com/foo` example).

---

### Changing the structure

> I think that TLS verification should always be host/registry/url-based but that doesn't work with the current design of sysregsv2, prefixes in particular, so I want to propose a change that should solve the upper problem.

\*shrug\* We can do so, but in that case let’s get as close to a clean solution as we can.

```
[[registry]]
url = "registry.com"
insecure = true
unqualified-search = true
[[registry.mirror]]
url = "mirror-1.registry.com"
[[registry.prefix]]
path="/bar"
prefix="registry.com/foo"
[[registry.prefix]]
path="/paste"
prefix="registry.com/copy"
trim=true
[[registry.prefix]]
prefix="another-registry.com:5000"
```

That eliminates the duplication of `insecure` and `unqualified-search`, sure; OTOH I’m not sure whether the `blocked` flag is intended to be per-hostname or per-prefix (@rhatdan ?).

WRT usability, it seems a bit confusing the way the mapping for `another-registry.com:5000` is hidden nested somewhere inside the `registry.com` entries; the way I think about this, the “prefix” is the primary lookup key and should be prominent.

Syntactically, the `prefix=[hostname]/path` form is ambiguous because namespace components within a registry can contain dots as well.

Also, I’m not entirely sure that there is a meaningful distinction between the `prefix`→`URL` mapping and between having mirrors; am I overlooking an essential difference?


Can we maybe talk about the configuration as conceptual data structures first, and then figure out the best way to represent it in TOML?

Is this the data we are trying to represent?
```go
type CompleteConfig struct {
  // May be empty. NOTE: Order matters
  UnqualifiedSearchOrder []Hostname
  // Order does not matter
  HostnameConfig map[Hostname]struct {
    // Defaults to false if not specified or there is no match for a Hostname
    Insecure bool
  }
  // Order does not matter as such, but longest match wins. If we match on a longer prefix, the entries for shorter prefixes, if any, are ignored.
  Prefixes map[HostnameRepoPrefix]struct {
    // Defaults to false if not specified or there is no HostnameRepoPrefix match.
    Blocked bool 
    // Defaults to the key of this structure (i.e. identity mapping) if not specified or there is no HostnameRepoPrefix match. Must not be empty.
    PhysicalLocations []HostnameRepoPrefix 
  }
}
```

mtrmac: > Syntactically, the `prefix=[hostname]/path` form is ambiguous because namespace components within a registry can contain dots as well.

I’m sorry, actually, if the `/path` variant has a mandatory leading `/`, there is no ambiguity.

vrothberg: > I’m not too worried about this; for the most part, either a registry hostname is “real” (an underlying server that we want to reach), in which case there will be no configuration or at most a single entry for the whole server; or the hostname is intended to not be used directly, and is mirrored locally (e.g. a local mirror of a tiny subset of docker.io, with Prefix entries docker.io/$repo and URL or Mirrors pointing to the local mirror), in which case there may be many entries but possibly pointing at different mirrors, and there’s no reason to assume that all the mirrors will use TLS in the same way.
>
>I.e. I’d expect that there is either no entry for example.com[/anything], a single entry for all of example.com, or several entries for example.com/foo which are likely to match; a request for example.com/bar which does not match, but with an existing example.com/foo configuration seems unlikely to me.
>
>Still, sure, not repeating the same boolean many times for a single hostname would be simpler, especially when trying to turn off the insecure flag.

Alright, you convinced me, let's push things forward as they are now. I guess I was overthinking the issue.

vrothberg: > I’m not sure how to handle this entirely cleanly. Yet another parameter to newDockerClientWithDetails?

I reduced the list of parameters now as those can (or should?) be set by the callers. However, it's a bit of a deja vu as we run into the same problem we hit in `sysregistriesv2` when parsing strings that can be a fully-qualified image reference, or a registry or a registry plus path. For now, I tried to solve that be having two parameters (i..e, registry and reference) where the reference is used in `FindRegistry(..)`.

mtrmac: > >I’m not too worried about this; for the most part, either a registry hostname is “real” (an underlying server that we want to reach), in which case there will be no configuration or at most a single entry for the whole server; or the hostname is intended to not be used directly, and is mirrored locally (e.g. a local mirror of a tiny subset of docker.io, with Prefix entries docker.io/$repo and URL or Mirrors pointing to the local mirror), in which case there may be many entries but possibly pointing at different mirrors, and there’s no reason to assume that all the mirrors will use TLS in the same way.
> 
> Alright, you convinced me, let's push things forward as they are now. I guess I was overthinking the issue.

(FWIW the “not too worried about this” comment was about the narrow case of having a configuration for prefix example.com/foo and using example.com/bar with no configuration.  I do agree about the general ugliness of having possibly several “insecure TLS” flags for the single hostname — it’s not a _blocker_ issue for me, but it is a bit worrying.)

mtrmac: >  I reduced the list of parameters now as those can (or should?) be set by the callers.

OK… it’s now easier for a caller to forget to set one up, but on the plus side the values have names, so it’s definitely not worse.

docker/docker_client.go
```diff
@@ -181,13 +183,22 @@ func newDockerClientFromRef(sys *types.SystemContext, ref dockerReference, write
...
+	client.signatureBase = sigBase
+	client.scope.actions = actions
+	client.scope.remoteName = reference.Path(ref.ref)
+	return client, nil
 }
 
-// newDockerClientWithDetails returns a new dockerClient instance for the given parameters
-func newDockerClientWithDetails(sys *types.SystemContext, registry, username, password, actions string, sigBase signatureStorageBase, remoteName string) (*dockerClient, error) {
+// newDockerClient returns a new dockerClient instance for the given registry
+// and reference.  The reference is used to query the registry configuration.
```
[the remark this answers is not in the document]
mtrmac: (Also it would be nice to mention that the caller is responsible for setting up `username`, `password`, `signatureBase` and `authScope`, if any of them are necessary.)

docker/docker_client.go
```diff
@@ -292,10 +313,12 @@ func SearchRegistry(ctx context.Context, sys *types.SystemContext, registry, ima
 		registry = dockerV1Hostname
 	}
 
-	client, err := newDockerClientWithDetails(sys, registry, username, password, "", nil, "")
+	client, err := newDockerClient(sys, registry, registry)
```
mtrmac: One more step to the fix: the `reference` argument should be the original value of `registry`, before it gets mapped to `dockerV1Hostname`.

docker/docker_client.go
```diff
@@ -85,6 +86,7 @@ type dockerClient struct {
 	client        *http.Client
 	signatureBase signatureStorageBase
 	scope         authScope
+	skipTLSVerify bool
```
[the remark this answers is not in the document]
vrothberg: Done

docker/docker_client.go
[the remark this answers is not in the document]
vrothberg: Done. I updated the comments in `dockerClient` as well.

docker/docker_client.go
vrothberg: Done

vrothberg:
> I have prepared projectatomic/skopeo#521 for this.

Thanks a lot! I've been testing it locally but wasn't sure about the process.

docker/docker_client.go
```diff
@@ -181,13 +184,27 @@ func newDockerClientFromRef(sys *types.SystemContext, ref dockerReference, write
...
+	client.scope.remoteName = reference.Path(ref.ref)
+	return client, nil
 }
 
-// newDockerClientWithDetails returns a new dockerClient instance for the given parameters
-func newDockerClientWithDetails(sys *types.SystemContext, registry, username, password, actions string, sigBase signatureStorageBase, remoteName string) (*dockerClient, error) {
+// newDockerClient returns a new dockerClient instance for the given registry
+// and reference.  The reference is used to query the registry configuration
+// and can either be registry (.e.g, "registry.com:5000"), a fully-qualified
+// image reference (e.g., "docker.io/library/alpine:latest") or pointing to a
```
mtrmac: Wait… this actually works… because `FindRegistry` just uses `strings.HasPrefix` and does not require the matched `Prefix` to be followed by any kind of semantic separator (like `/` or maybe `:`.

Thus, `Prefix` `example.com/1` matches `example.com/1/foo` as well as `example.com/100/foo`. Is that what we want?

Just below the comment goes on discussing a registries namespace — but a string prefix and a namespace are not the same thing.

(The `strings.HasPrefix` match has existed before this PR, so this is not blocking, I can split it into a separate PR for discussion.)

docker/docker_client.go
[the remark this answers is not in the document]
mtrmac: > I updated the comments in `dockerClient` as well.

Thanks, good catch.

docker/docker_client.go
vrothberg: > Thus, Prefix example.com/1 matches example.com/1/foo as well as example.com/100/foo. Is that what we want?
> Just below the comment goes on discussing a registries namespace — but a string prefix and a namespace are not the same thing.

That's definitely an issue. Given our timezones are quite distant, I have enough time and can add a commit ontop of this PR.

vrothberg: > That's definitely an issue. Given our timezones are quite distant, I have enough time and can add a commit ontop of this PR.

@mtrmac I added another commit. A prefix match is now limited to the following cases:

- full match (`ref, prefix = "docker.io/alpine:latest"`)
- namespace match (`ref = "docker.io/alpine/namespace"` and `prefix = "docker.io/alpine"`)
  - the remainder of `ref` starts with a `/`
- image match (`ref = "docker.io/alpine:latest"` and `prefix = "docker.io/alpine"`)
  - the remainder of `ref` starts with a `:`

vrothberg: ^ a namespace match certainly also applies to matching a registry (e.g., `prefix = "registry.com:5000")

mtrmac:
> I have prepared projectatomic/skopeo#521 for this.

FWIW, the failed tests in there prompted me to write unit tests for `contextFromGlobalOptions`, and I’m not at all happy with how that handles the `--tls-verify` flag (notably if one of the subcommands does not declare the flag, it is silently set to the insecure value right now).  So I’m trying to overhaul the CLI parsing in there to actually work reliably, I’m not yet quite sure where that will lead me (WIP https://github.com/mtrmac/skopeo/tree/cli-parsing ).  

There might be a simple way to make the specifically the `--tls-verify` reliable enough without that overhaul, I haven’t really been looking to be honest.

vrothberg:
> FWIW, the failed tests in there prompted me to write unit tests for contextFromGlobalOptions, and I’m not at all happy with how that handles the --tls-verify flag (notably if one of the subcommands does not declare the flag, it is silently set to the insecure value right now). So I’m trying to overhaul the CLI parsing in there to actually work reliably, I’m not yet quite sure where that will lead me (WIP https://github.com/mtrmac/skopeo/tree/cli-parsing ).

Ouch. There are quite some unexpected pain points in the stack. CLI parsing sucks big time.
Thanks for doing that!

> There might be a simple way to make the specifically the --tls-verify reliable enough without that overhaul, I haven’t really been looking to be honest.

I may find time next week to help you untangle the code.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -315,23 +324,61 @@ func GetRegistries(ctx *types.SystemContext) ([]Registry, error) {
...
+			return true
+		}
+		// The regex expects strings that either start with a '/' or
+		// that start with a ':'.  Strings starting with a ':' have a
+		// subsequent lookahead for ':' and '/' to avoid matching
+		// registries at port boundaries.
+		//
+		// This way, it matches "/foobar", "/foobar:latest" and ":latest"
+		// but it doesn't match ":5000/foobar" or ":latest:foobar".
+		reg := regexp.MustCompile(`(?m)^(\/.*|:.[^:\/]*)$`)
```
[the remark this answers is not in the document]
vrothberg: > It should be possible to ultimately implement the desired semantics, with the current inputs, exactly correctly (e.g. by exporting DomainWithOptionalName in c/image/docker/registry, or something like that [...].

I was thinking about this issue for a while, and believe we can reduce complexity by just requiring the suffix/remainder to either be empty or to start either with `/` or with `:`. This would also match at port boundaries, but that might not be such a big deal. Some users might even find a usecase for it when wanting to mirror multiple registries running on one host.

We can still introduce a `DomainWithOptionalName` regex (or something similar) for input validation, but I don't think that's necessary now can be added at a later point.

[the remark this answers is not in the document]
mtrmac: > > Ultimately, matching on the suffix only seems insufficient, because `:5000` can either be a valid tag or a valid port number. So, `example.com:5000` should not match `example.com`, but `example.com/image:5000` should match `example.com/image`.
>
> This is already working and the regex detects that (minus the digest case).

How? AFAICS it can’t work because the suffix is the same, and testing confirms it: https://play.golang.org/p/0QHfsmaoApO .

Am I overlooking something that can distinguish the two cases?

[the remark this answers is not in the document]
mtrmac: > I believe we can reduce complexity by just requiring the suffix/remainder to either be empty or to start either with `/` or with `:`. This would also match at port boundaries, but that might not be such a big deal. Some users might even find a usecase for it when wanting to mirror multiple registries running on one host.

I’d rather do this precisely correctly; it seems to me it should not be that much code, maybe I could try to do that next week.

(Either way, the problematic matching code already exists, so it is not strictly blocking this PR.  The skopeo test failures are — and I’m getting close to finishing the CLI rework in there, projectatomic/skopeo#523 — and that rework seems mostly unnecessary for the purpose of this PR, after all; I will know for sure after I finish the tests in there.)

mtrmac:
> I have prepared projectatomic/skopeo#521 for this.
> FWIW, the failed tests in there prompted me to write unit tests for contextFromGlobalOptions, and I’m not at all happy with how that handles the --tls-verify flag (notably if one of the subcommands does not declare the flag, it is silently set to the insecure value right now). So I’m trying to overhaul the CLI parsing in there to actually work reliably, I’m not yet quite sure where that will lead me (WIP https://github.com/mtrmac/skopeo/tree/cli-parsing ).
> 
> There might be a simple way to make the specifically the --tls-verify reliable enough without that overhaul, I haven’t really been looking to be honest.

It turns out that while `cli.Context.BoolT` is pretty unreliable/risky, `IsSet`/`GlobalIsSet` work as necessary, and the original test failures were pointing out my stupid thinko when updating skopeo for this API.  projectatomic/skopeo#521 now passes tests independently of the larger CLI rework, with added flag parsing tests to ensure that the results are as expected.

So, only the namespace matching is outstanding. Arguably this PR should not block on it at all and that part could be split — OTOH enabling this implicit behavior with the `strings.HasPrefix` seems risky. I’ll hopefully return to this in a few days.

vrothberg:
@mtrmac Thanks for info!

> So, only the namespace matching is outstanding. Arguably this PR should not block on it at all and that part could be split — OTOH enabling this implicit behavior with the strings.HasPrefix seems risky. I’ll hopefully return to this in a few days.

Yesterday, we hit a pretty interesting detail with the prefix/mirror downstream patch we maintain for Docker-ce. The issue is that registries can be nested under some path on the host (see `"prefix"` at https://docs.docker.com/registry/configuration/#http), which can lead to auth issues when doing a namespace translation. 

An example is a registry running at `example.com/reg-a/` and we configure `URL: "example.com/reg-a/docker.io"` and `Prefix: "docker.io/"`. What we want is to have docker.io mirrored at `example.com/reg-a/docker.io` where the registry root is `example.com/reg-a` ... in this case we need to *know* what the registry is, as we cannot *parse* it from a given string/reference. In other words, that needs to be covered by the `registries.conf` and `sysregistriesv2` and will influence our `isPrefix(..)` issue above.

mtrmac:
> Yesterday, we hit a pretty interesting detail with the prefix/mirror downstream patch we maintain for Docker-ce. The issue is that registries can be nested under some path on the host (see `"prefix"` at https://docs.docker.com/registry/configuration/#http),

The ”Docker reference” format just can’t express the difference between an image available through the API paths `example.com/v2/manifests/ns/repo` and through `example.com/ns/v2/manifests/repo`.  Only the former one can be specified.

So, such registries can’t be used with `c/image/docker` at all; it always uses `${reference.Domain(ref.ref)}/v2/`.

Similarly, looking at `moby/moby`, `lookupV2Endpoints` can’t turn a Docker reference into something referring to a registry running at a non-root path.  It can only return a non-hostname-only path only if it is present in `s.config.Mirrors` — but `registry.ValidateMirror` rejects mirror URLs with paths.

So, this seems to be a server-side capability with no client-side implementation.

Or are you saying that you want to add support for this in clients?

> which can lead to auth issues when doing a namespace translation. 

What does “auth issues” mean? (Overall, it seems reasonable to me that clients should use credentials they have stored for the mirror, and not for the underlying primary source, if they connect to a mirror. I can’t see how using non-empty path prefixes changes that, except maybe for the need to support them in `pkg/docker/config`.)

(If the example below is supposed to be the “auth issue”, I can’t see any connection to authentication.)

> An example is a registry running at `example.com/reg-a/` and we configure `URL: "example.com/reg-a/docker.io"` and `Prefix: "docker.io/"`. What we want is to have docker.io mirrored at `example.com/reg-a/docker.io` where the registry root is `example.com/reg-a` ... in this case we need to *know* what the registry is, as we cannot *parse* it from a given string/reference. 

Which of the strings in this paragraph are Docker references (hostname/ns…/repo) and which are API URLs? There isn’t a single `/v2/` above, so they all seem to be just Docker references; in that case, mapping `docker.io` to `example.com/reg-a/docker.io` seems trivial and already supported by the format.

Or are you saying that you want to deploy a registry at the URL `example.com/reg-a/v2/`, and want to map `docker.io` inside that registry as a `/docker.io` namespace? Sure, that would be new, and we would need to differentiate between the target API URL and the target namespace — but even if so…

> In other words, that needs to be covered by the `registries.conf` and `sysregistriesv2` and will influence our `isPrefix(..)` issue above.

… I can’t see how; `isPrefix` is used to match Docker reference strings, not URLs containing the `/v2/` fragment, against the Docker reference namespace hierarchy.

vrothberg:
I'll catch up soon on this, @mtrmac. Sorry for the silence.

vrothberg:
> Or are you saying that you want to deploy a registry at the URL example.com/reg-a/v2/, and want to map docker.io inside that registry as a /docker.io namespace? Sure, that would be new, and we would need to differentiate between the target API URL and the target namespace — but even if so…

@mtrmac Yes, that's something we would like to cover. I propose to delete the last commit of this PR and continue from there? Maybe, we can come up with some action items? Currently, I see the following open topics:

- Add a `PullImage(...)` API based on sysregistriesv2 that Podman, Skopeo, Buildah and friends can use.
- Refine the sysregistriesv2 config in a way that we can distinguish between the root and a namespace of a registry.

I know you have been working a lot on libpod and c/images lately, but I couldn't follow all the changes. Maybe you have some more action items to add or feedback on the proposed ones above?

vrothberg:
ping @mtrmac

vrothberg:
Ping @mtrmac. How shall we proceed with this PR? I think we blew it a bit but the initial fix to skip TLS verification is still something we need.

mtrmac:
I’m terribly sorry about this; my expectation that I can get a proper parser for the hostname/ns prefix syntax done and API designed quickly turned out to be horribly wrong.  Let’s just go with a minimal implementation of the prefix match, then (but not commit to supporting all the accidental matches that that implementation allows.)

docker/docker_client.go
mtrmac: This is an internal API; so please document only the two actually necessary inputs: either only a hostname[:port], or a hostname:[port]/[ns/…]/repo ; with hostname[:port] a single unit that must match exactly (although the implementation may do more than that for now).

(This does not directly affect the implementation, but avoids even a pretend commitment to support unnecessary formats, making future refactoring easier.)

pkg/sysregistriesv2/system_registries_v2.go
mtrmac: For now let’s go with `remainder == "" || remainder[0] isOneOf(":/@")` then, but document that the valid `prefix` valuers are _only_
- hostname[:port] as a single unit (not supporting a match of hostname prefix against a hostname:port input)
- hostname[:port]/…/ns
- hostname[:port]/…/repo
- hostname[:port]/…/repo:tag
- hostname[:port]/…/repo@digest


The implementation would match a few more inputs, of which only the hostname vs. hostname:port case would actually be valid inputs, and running multiple registries on a single host seems unlikely enough for now.

vrothberg:
> Let’s just go with a minimal implementation of the prefix match, then (but not commit to supporting all the accidental matches that that implementation allows.)

I agree :+1: Thanks for catching up!

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: If we follow where the Moby PR goes, we can just match at paths which makes the logic (and entailing problems) simpler. If the prefix doesn't end with a '/' then we append one. It must be documented but it avoid duplicating the quite hairy parsing.

vrothberg:
@mtrmac, the current state of this PR contains some changes wher I am unsure if they're still desired. I'll list them here, to make {,n}acking them a bit easier. The details of those changes and the discussion are still all in the comments here, but let's have a clean start here. Please ack/nack the changes:

### sysregistriesv2: adjust API to SystemContext
Commit: https://github.com/containers/image/pull/468/commits/5f91e53691f449605a4b0883ce9e4f6764952f77

Extends the API with a systemContext to make use of the internal cache. I just noticed that daemons like CRI-O need to be careful with the migration in case changes to this file are propagated directly.

###  sysregistriesv2: do not require default config to exist
Commit: https://github.com/containers/image/pull/468/commits/ebef31445a2a780c0decdd845915b6626638f39f

Return an empty registry instead of erroring out when the default config doesn't exist.

### dockerClient: skip TLS verification if configured in registries
Commit: https://github.com/containers/image/pull/468/commits/cdabcb717d0ec4e18f4d164282df3240f60e1783

"Unless specified in the SystemContext, the TLS
config's InsecureSkipVerify flag now adheres to the registries Insecure
field (if specified in the registries.conf).  Notice, that the default
value remains false (i.e., TLS verification is performed)."

There's also a new type `types.OptionalBool` to distinguish between the default value and a user-specified one.

### sysregistriesv2: limit prefix to namespace/image matches
Commit: https://github.com/containers/image/pull/468/commits/b80a785e1028251fabe0a8a1686b99a0b8c65b9f

I prefer to break this out into a separate PR. As mentioned above, the Moby PR is going for matching at path boundaries which simplifies many things.

pkg/sysregistriesv2/system_registries_v2.go
mtrmac: If you mean essentially `reference == prefix || strings.HasPrefix(reference, prefix+"/")`, that would make it impossible to match (e.g. configure mirrors) at the level of an individual repo — and it seems to me that that should be possible.

mtrmac:
> the current state of this PR contains some changes where I am unsure if they're still desired. 

All of the components make sense to me, both individually and as a whole.

(I’ve also marked various review comments as resolved to hopefully make this easier to read.)


> ### sysregistriesv2: adjust API to SystemContext
> Commit: [5f91e53](https://github.com/containers/image/commit/5f91e53691f449605a4b0883ce9e4f6764952f77)

ACK. (Absolutely non-blocking comments on tests.)

>I just noticed that daemons like CRI-O need to be careful with the migration in case changes to this file are propagated directly.

(Do you mean just because it changes API? That’s not unusual for c/image right now. (I don’t know, maybe that should change? Up to the important callers, really.) Or is there some problematic behavior change I have missed?)

> ### sysregistriesv2: do not require default config to exist
> Commit: [ebef314](https://github.com/containers/image/commit/ebef31445a2a780c0decdd845915b6626638f39f)

ACK.

> ### dockerClient: skip TLS verification if configured in registries
> Commit: [cdabcb7](https://github.com/containers/image/commit/cdabcb717d0ec4e18f4d164282df3240f60e1783)

ACK, but please update the comment per 
https://github.com/containers/image/pull/468#discussion_r229355474 .


> ### sysregistriesv2: limit prefix to namespace/image matches
> Commit: [b80a785](https://github.com/containers/image/commit/b80a785e1028251fabe0a8a1686b99a0b8c65b9f)
> 
> I prefer to break this out into a separate PR. As mentioned above, the Moby PR is going for matching at path boundaries which simplifies many things.

We do need to fix this, but I don’t understand how the current regex version works.  The recentish conversation was about using a simpler approximation for now.

Both a separate PR and just a commit in this one is fine with me — but a separate one would have to be merged before this one anyway, so that we don’t incorrectly disable TLS because of broken prefix  matching.

I suppose breaking it out would give us a fresh space to discuss it, at least…

pkg/sysregistriesv2/system_registries_v2_test.go
vrothberg: You're absolutely right. Cleaned up with the next update to this PR. Thanks!

docker/docker_client.go
vrothberg: Done.

vrothberg:
> (Do you mean just because it changes API? That’s not unusual for c/image right now. (I don’t know, maybe that should change? Up to the important callers, really.) Or is there some problematic behavior change I have missed?)

I was more referring to the caching of the `registries.conf` which was implemented prior to this PR. I think CRI-O is loading the registries.conf for each pull and hence always gets the latest changes. I'd argue that sending a HUP should be done, similar to dockerd.

> I suppose breaking it out would give us a fresh space to discuss it, at least…

I agree and will remove this commit from the PR.

rhatdan:
@vrothberg I am fine with reloading via a signal, or just stating the file to see if it has changed since the last time the file was read.

mtrmac:
> I updated the PR. CI breaks due to the optBool change.

Thanks!  The existing code LGTM.

I have updated https://github.com/containers/skopeo/pull/521 against this, I expect the test to pass there.

The outstanding concerns:

- Fixing the prefix matching in `FindRegistry`.
- > I was more referring to the caching of the registries.conf which was implemented prior to this PR. I think CRI-O is loading the registries.conf for each pull and hence always gets the latest changes. I'd argue that sending a HUP should be done, similar to dockerd.

  That makes sense — what should happen for this PR?
  - Add a stat + time check to the existing cache code?
  - Add an explicit “invalidate cache” API, and let CRI-O handle SIGHUP by calling it?
  - (Temporarily?) disable the existing cache? 
  - Merge as is anyway, and defer this to a separate PR before updating CRI-O? (A bit risky.)
  - Something else?

vrothberg:
In contrast to my previous statement, I just added the `InvalidateCache` API.

The remaining bits regarding prefix matching etc. require some more thought. Maybe we can set up a meeting in the near future to share some ideas?

What I mostly want to check with you is if we could rewire the sysregsv2 config. I find some beauty in the last design of the PR for Moby, which looks as follows:
```
{
    "registries": [
        {
          "prefix": "docker.io/",
          "pull": [
            {"url": "http://localhost:5000"}, // Will be used as `http://localhost:5000/v2/`
            {"url": "https://registry-1.docker.io"} // Will be used as `https://registry-1.docker.io/v2/`
          ],
          "push": [
            {"url": "https://registry-1.docker.io"} // Will be used as `https://registry-1.docker.io/v2/`
          ]
        },
    "registries": [
        {
          "prefix": "docker.io/library/",
          "pull": [
            {"url": "http://localhost:5000/v2/library"},
            {"url": "https://registry-1.docker.io/v2/library"}
          ],
          "push": [
            {"url": "https://registry-1.docker.io/v2/"} 
          ]
        },
    ]
}
```

We can do the same with TOML. Prefixes must adhere to paths and paths only, so either host, namespace or registry. Another nice improvement is that the concept applies to pushing AND pulling, each configured separately.

mtrmac:
> In contrast to my previous statement, I just added the `InvalidateCache` API.

ACK, though this is primarily up to CRI-O maintainers. @runcom ? Or who can comment on this?

> The remaining bits regarding prefix matching etc. require some more thought. Maybe we can set up a meeting in the near future to share some ideas?

Yeah, a meeting could be quite useful.

> What I mostly want to check with you is if we could rewire the sysregsv2 config.

_The one thing_ that is blocking this PR is incorrectly matching prefixes and unwanted configurations; changing the config format is, I think, quite unrelated (both the current and any proposed different config format would need the prefix matching fixed).

---

That said…

The sysregistriesv2 code has already been vendored into buildah and presumably deployed in released binaries. I’ve heard people actively asking for mirror support. What is the status of the file format? Is it unstable until documented in https://github.com/containers/image/blob/master/docs/containers-registries.conf.5.md ?

Sure, as long as it is compatible with the previous syntax, we can just keep pointing people at the old syntax and keep tweaking this one — as long as everyone understands that this is the status.

---

> I find some beauty in the last design of the PR for Moby, which looks as follows:
> 
> ```
> {
>     "registries": [
>         {
>           "prefix": "docker.io/",
>           "pull": [
>             {"url": "http://localhost:5000"}, // Will be used as `http://localhost:5000/v2/`
>             {"url": "https://registry-1.docker.io"} // Will be used as `https://registry-1.docker.io/v2/`
>           ],
>           "push": [
>             {"url": "https://registry-1.docker.io"} // Will be used as `https://registry-1.docker.io/v2/`
>           ]
> …
>           {"url": "https://registry-1.docker.io/v2/library"}
> ```
> We can do the same with TOML. Prefixes must adhere to paths and paths only, so either host, namespace or registry. Another nice improvement is that the concept applies to pushing AND pulling, each configured separately.

This is nice, an improvement over the earlier discussion around https://github.com/containers/image/pull/468#discussion_r199617603 .

- Eliminating the “mirror” concept is convenient (I assume that `push` and `pull` can be omitted to default to the standard URL).
- Forcing the users to say `/v2/library` looks rather confusing — though I appreciate how that unifies the `/v2/library` and `/different-deploment/within-a-hostname/v2/` cases.
- This still duplicates the secure/insecure configurations across all the mirrored repositories.

mtrmac:
> _The one thing_ that is blocking this PR is incorrectly matching prefixes and unwanted configurations

For now, something like #534 ?

vrothberg:
> The sysregistriesv2 code has already been vendored into buildah and presumably deployed in released binaries. I’ve heard people actively asking for mirror support. What is the status of the file format? Is it unstable until documented in https://github.com/containers/image/blob/master/docs/containers-registries.conf.5.md ?

I think I am the wrong audience for this statement given I am trying to push those things forward here and for Moby for quite some time :^)

> Sure, as long as it is compatible with the previous syntax, we can just keep pointing people at the old syntax and keep tweaking this one — as long as everyone understands that this is the status.

The progress of this package certainly depends on the priority we give it to be finished. I think we're close for it to be useful in production. One thing I think we should have is an API to abstract from using those endpoints/mirrors for pushing and pulling. I wouldn't like libpod to keep all abstractions and others to start using those.

> Yeah, a meeting could be quite useful.

This week is packed for me and next week I'll be on vacation. Maybe we can set something up for early December? I always assumed you're located at the West Coast, are you?

vrothberg:
> For now, something like #534 ?

I'll comment over there :)

vrothberg:
> The sysregistriesv2 code has already been vendored into buildah [...]

Turns out, it was a mistake doing that. Podman isn't using sysregsv2 yet but vendors buildah, so `podman pull` uses entirely different code for pulling than `podman build`. A user just reported this in #podman on IRC. 

The registries.conf looks as follows:
```
[registries.search]
registries = ['docker.io', 'registry.fedoraproject.org', 'quay.io', 'registry.access.redhat.com', 'registry.centos.org']

[registries.insecure]
registries = ['10.1.8.88:8787']

[registries.block]
registries = []
```

`podman pull` from 10.1.8.88:8787 works while a `podman build` breaks immediately with:
`invalid URL '123.456.789.0:5000/with/path': parse 123.456.789.0:5000/with/path: first path segment in URL cannot contain colon`

It's again the URL parsing and the fact that we're not using URI prefixes which confuses net/url.

vrothberg:
> :) Central Europe. But given >24-hr advance notice, and no conflicts, I can be available any time at all.

Perfect! How about Tuesday 4th of Dec at 2PM (CET)? Maybe you can join #podman to setup a conference tool of our choice?

vrothberg:
> > > The sysregistriesv2 code has already been vendored into buildah and presumably deployed in released binaries. I’ve heard people actively asking for mirror support. What is the status of the file format? Is it unstable until documented in https://github.com/containers/image/blob/master/docs/containers-registries.conf.5.md ?
> > 
> > 
> > I think I am the wrong audience for this statement given I am trying to push those things forward here and for Moby for quite some time :^)
> 
> In one sense, this is primarily your work so it is perfectly up to you to declare it stable or not yet stable; in another, as soon as a few packages vendor in a particular implementation, no matter which one, and start relying on the new features, I’d expect significant pushback to future unstable changes.
> 
> Basically I don’t know where we are in this process (as you say, Buildah has already vendored the subpackage) and whether this is widely understood; I’d like to prevent a misunderstanding. @rhatdan ?

Buildah only uses the package but not the new format which gives us the change to still change the format (i.e., extend it). I would really love to integrate the push/pull endpoints semantics similar to the ones of the Moby PR. This would also enable users to configure "push proxies". Starting from next week, I will have time to finish the remaining bits.

@mtrmac, can we merge this PR? I think it does now what we wanted it to do. The remaining bits and potential extensions can be sorted out in a fresh PR.

mtrmac:
> @mtrmac, can we merge this PR? I think it does now what we wanted it to do. The remaining bits and potential extensions can be sorted out in a fresh PR.

The one blocker for me is the prefix matching – sure, we can tell people not to use namespace/repo prefixes for now, but this matches prefixes of simple hostnames as well.

I suppose we can do ~what you propose in #534 , and as a quick fix use
```go
strings.HasPrefix(ref, r.Prefix + "/") || ref == r.Prefix
```
; the first part would work for named references vs. hostnames, and the second part for `CheckAuth` — though, warning, I didn’t test this one bit.

mtrmac:
> Perfect! How about Tuesday 4th of Dec at 2PM (CET)? Maybe you can join #podman to setup a conference tool of our choice?

Sure.  FWIW RH has a BlueJeans instance which should allow external connections.

rhatdan:
I scheduled a BlueJeans Meeting for Dec 4th 2pm (CET) 8am (EST) for containers/image discussions.
You should get an invite.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -82,58 +82,20 @@ func (e *InvalidRegistries) Error() string {
...
+// the sanitized input string.  An error is returned if the input string is
+// empty or if contains an "http{s,}://" prefix.
 func parseURL(input string) (string, error) {
 	trimmed := strings.TrimRight(input, "/")
 
 	if trimmed == "" {
 		return "", &InvalidRegistries{s: "invalid URL: cannot be empty"}
 	}
 
-	// Ultimately, we expect input of the form example.com[/namespace/…], a prefix
```
mtrmac: Fair enough, I guess we need this relaxed urgently enough that waiting for a correct parser is not reasonable.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -222,6 +184,12 @@ func postProcessRegistries(regs []Registry) ([]Registry, error) {
 			}
 		}
 
+		// make sure a Prefix ends with "/" to ensure matching at
+		// path and hence at namespace boundaries
+		if !strings.HasSuffix(reg.Prefix, "/") {
```
mtrmac: Couldn't this be just moved to the check in `FindRegistry` instead of leaking into user-visible data?

Also, AFAICS this breaks the `CheckAuth` case (`FindRegistry(…, hostnameWithoutAnySlashes)`).

mtrmac: … sure, independently it’s perfectly reasonable to define that canonical `Registry.Prefix` values always end with exactly one slash, and this is just normalizes the values to confirm to this definition, but if `FindRegistry`, the primary user, ends up having to undo that, then it’s a bit impractical.

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: Yes, that's unfortunate but necessary. Moby doesn't hit the issue as it enforces the URI scheme to either be `http` or `https`.

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: > Also, AFAICS this breaks the CheckAuth case (FindRegistry(…, hostnameWithoutAnySlashes)).

I think that `hostnameWithoutAnySlashes` should be appended with a "/" in `FindRegistry(...)`? I didn't (yet) add code to throw an error if a Prefix doesn't end with slash but likely that's going to happen (to make it more transparent to the user). Once that happens, `hostnameWithoutAnySlashes` is problematic.

mtrmac: > I think that `hostnameWithoutAnySlashes` should be appended with a "/" in `FindRegistry(...)`?

Would that involve adding an extra check whether the input is a hostname without any slashes, or would a slash be unconditionally added (even if the input is a named ref, possibly ending with a tag or digest)?

Even to implement precisely the proposed semantics (`.Prefix` matches hostnames and namespaces, not repos or tagged/digested images — at this point I’m fine with that, I’m _not_ arguing about what the semantics should be, except that `CheckAuth` should work), adding the trailing slash to the `.Prefix` value just seems to make things more complex instead of simpler.

Or, of course, I’m being stupid — I haven’t actually tested anything about this in practice, so I may well be misunderstanding how the code actually works.

vrothberg: No at all. You have a very valid point. Then let's go with your initial proposal (similar to `strings.HasPrefix(ref, r.Prefix + "/") || ref == r.Prefix`).

vrothberg: Pushed the changes.

mtrmac:
Looks good, but `make validate` fails.

`gofmt -s -w pkg/sysregistriesv2/system_registries_v2.go`, please.

mtrmac:
👍 

Tests in https://github.com/containers/skopeo/pull/521 pass . 🎉 

Thanks again!

vrothberg:
Awesome, thanks a lot!

PR #470, opened 2018-06-22
Search was not looking for credentials in authfile

umohnani8:
docker.SearchRegistry was not calling config.GetAuthentication
to get the credentials of a registry from the authfile.

Fixes https://github.com/projectatomic/libpod/issues/976

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -280,13 +280,19 @@ func SearchRegistry(ctx context.Context, sys *types.SystemContext, registry, ima
 	v2Res := &V2Results{}
 	v1Res := &V1Results{}
 
+	// Get credentials from authfile for the underlying hostname
+	username, password, err := config.GetAuthentication(sys, registry)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error getting username and password")
+	}
+
 	// The /v2/_catalog endpoint has been disabled for docker.io therefore the call made to that endpoint will fail
 	// So using the v1 hostname for docker.io for simplicity of implementation and the fact that it returns search results
 	if registry == dockerHostname {
 		registry = dockerV1Hostname
 	}
 
-	client, err := newDockerClientWithDetails(sys, registry, "", "", "", nil, "")
+	client, err := newDockerClientWithDetails(sys, registry, username, password, "", nil, "")
 	if err != nil {
 		return nil, errors.Wrapf(err, "error creating new docker client")
 	}


umohnani8:
@mtrmac @runcom @rhatdan  PTAL
@haircommander for your reference.

docker/docker_client.go
```diff
@@ -229,7 +230,7 @@ func newDockerClientWithDetails(sys *types.SystemContext, registry, username, pa
 // CheckAuth validates the credentials by attempting to log into the registry
 // returns an error if an error occcured while making the http request or the status code received was 401
 func CheckAuth(ctx context.Context, sys *types.SystemContext, username, password, registry string) error {
-	newLoginClient, err := newDockerClientWithDetails(sys, registry, username, password, "", nil, "")
+	newLoginClient, err := newDockerClientWithDetails(sys, registry, "", nil, "")
```
mtrmac: How does this work? `username` and `password` are now ignored, and validating the two is IIRC the whole purpose of `CheckAuth`.

umohnani8: They are not ignored. `newDockerClientWithDetails` calls `GetAuthentication` and retrieves the username and password. Initially `newDockerClientFromRef` was doing the get, but `newDockerClientFromRef` is not used anywhere in the search path. Plus `newDockerClientFromRef` calls `newDockerClientWithDetails` so doing the `GetAuthentication` in there makes more sense.

mtrmac: > newDockerClientWithDetails calls GetAuthentication and retrieves the username and password.

Retrieves them _from the auth file_.

Compare with https://github.com/projectatomic/libpod/blob/c08a1e0b1124cf0b7c4bb80e09717134540ec593/cmd/podman/login.go#L83 , which first calls `CheckAuth`, and only on success writes the credentials to the auth file.

mtrmac: (by “`username` and `password` are now ignored” I refer to the parameters to `CheckAuth`, not to `newDockerClient*`)

umohnani8: ohh, I see it now. My bad, fixing...

umohnani8:
Should be fixed now. @mtrmac PTAL

docker/docker_client.go
```diff
@@ -212,6 +208,13 @@ func newDockerClientWithDetails(sys *types.SystemContext, registry, username, pa
 		tr.TLSClientConfig.InsecureSkipVerify = true
 	}
 
+	if username == "" && password == "" {
+		username, password, err = config.GetAuthentication(sys, registry)
```
mtrmac: The handling of the various `*.docker.io` host names does not look right, or at least it is changing:

- Before, for `newDockerClientFromRef`, `GetAuthentication` was called with `docker.io`; now it is done with `dockerRegistry` == `registry-1.docker.io`.  That could be solved by moving this code a bit higher.
- `SearchRegistry` sets `registry` to `dockerV1Hostname`, so that’s what `GetAuthentication` will be called with. That’s more difficult.

`GetAuthentication` does eventually normalize all three to a single value — but only for the `AuthConfigs` lookup, not for `CredHelpers`; so I think we don’t want to change the inputs.

So, either `newDockerClientWithDetails` could get a separate “semantic host name” / “configuration lookup key” (I don’t know how to call that) parameter, which remains `docker.io`, or we could add a ~duplicate `GetAuthentication` call into `SearchRegistry` instead of consolidating both into `newDockerClientWithDetails`. 

Both are somewhat ugly, the latter one maybe a bit less: the distinction between the use of the three `docker.io` hostnames concisely seems difficult; whereas a simple duplication of a call to an external method is just three extra _simple_ lines and the distinction between the use of the hostnames is captured directly in straight-line code.

Or there well could be some much cleaner way to express it, one I’m missing.

TomSweeneyRedHat:
LGTM assuming happy tests

docker/docker_client.go
mtrmac: This still calls `GetAuthentication` after the two remappings, i.e. with `dockerRegistry` / `dockerV1Hostname`.  (It now won’t affect `newDockerClientFromRef` because that does its own lookup, but `SearchRegistry` will cause `dockerV1Hostname` to be looked up.)

mtrmac: (I don’t know, maybe it would make sense for `GetAuthentication` to always normalize these three inputs to the same key; it’s not at all obvious that it should accept all the pre-normalized kinds of input — https://docs.docker.com/engine/reference/commandline/login/#extended-description says that the keys are “registry domains”, which seems to rule out at least the _schema_`://`_host_ format accepted by `normalizeRegistry`.

If `GetAuthentication` mapped all of the different `docker.io`-related hostnames and specs to a single value, the callers in this file could be more lenient and we could have a single `GetAuthentication` call like this one. But I haven’t actually looked into the details of what docker does, nor what the documentation says, for the corner cases.)

mtrmac: > says that the keys are “registry domains”

I meant the keys in `credHelpers`.

umohnani8: Should `GetAuthentication` be called right at the beginning of the function then? Before any of the remapping is done?

mtrmac: That helps with `dockerRegistry`, but not with `dockerV1Hostname`.

umohnani8: finally fixed now. Added a secong `GetAuthentication` call in `SearchRegistry` before the registry hostname change happens.

mtrmac:
👍 pending tests.  Thanks!

PR #472, opened 2018-06-30
Don't physically store “empty layers” in v2s1 images in containers-storage:

mtrmac:
`buildah` image export uses the layer tree to build a list of DiffID values; these correspond to non-empty layers in the manifest (in v2s1, which represents empty layers; v2s2 and OCI don't list them in manifests at all).  So far we have been storing the empty layers as well, which broke buildah assumptions ( https://github.com/projectatomic/buildah/issues/684 ).

To make `buildah` work, don't store empty layers to the underlying storage. Then, as an `ImageSource`, replace the discarded empty layers with a canonical version, which is served by hard-coded code in `GetBlob`.

**NOTE**: This will make the current problems with v2s1 images A BIT WORSE for pre-existing containers/storage stores; users with such images will have to delete (or at least untag) the v2s1 images, and pull them again:

In principle, we could support images which don't physically store any empty layers, or images which physically store all of them, in a single codebase and be compatible with the pre-existing stores.  (In the intermediate case, where only a few empty layers are stored, matching the physical layout (which is uncompressed, i.e. has DiffIDs) to the v2s1 manifest (which only contains compressed blob digests) is not possible in general because `storage.Layer.CompressedDigest` is optional.  So, it's either all or nothing.)

All of the code needs to support the model of not physically storing empty layers, because that's how v2s2/OCI always work.  As for supporting also the other model with physical empty layers (i.e. existing stores with pre-existing v2s1 images), it seems not worth it overall:

- So far, copying v2s1 images to `containers-storage:` has created physical versions of the empty layers, and such images could be run; but they couldn't be copied out (`LayerInfosForCopy` returned a different number of entries, making `UpdateLayerInfos` abort) and they couldn't be used as `buildah` base images.  So, they were never all that well supported anyway.

- Supporting both formats would commit us to maintain two different code paths in anything that walked the layer tree and expected some correspondence with config `RootFS.DiffIDs`; we expect the proportion of v2s1 images to trend to zero over time, so the long-term maintenance burden (and risk of v2s1 being broken again in new code because noone notices) is not proportionate.

- `storageImage.Commit()` enforces that storing the same image (~same manifest) always ends up with the same top layer.  That code would have to be taught to support both possible top layer IDs - to compute, without physically building the layer tree, the layer ID sequence for both alternative layouts, check for a pre-existing image and top layer ID match, determine which of the layouts to use, and finally build it.  That would be _even more_ brittle and difficult-to-test code to support pre-existing images which never worked all that well.

So, anyone copying v2s1 images out of `containers-storage:` will need to delete them and create/pull them again.  This should primarily (only?) impact `buildah` users, who haven’t been able to work with such images until now anyway.

Plain deployment-only uses of images (e.g. k8s nodes with CRI-O) can continue to use the previously pulled images, and they will continue to work as well as they ever did.  BUT, force-pulling the same v2s1 image again, like `buildah from --pull-always`, will fail with a top layer mismatch; again, the workaround is to delete/untag the v2s1 image, and pull it again using the new codebase.

(The above describes the last commit; the previous ones are minor cleanups and building infrastructure to make it possible.  See individual commit messages for details.)

@nalind @runcom PTAL


rhatdan:
I have no problem with making the existing store bad rather then complicating the code.  We have a one time problem, versus years of carrying complicated code.

rhatdan:
@giuseppe PTAL

rhatdan:
@runcom PTAL

rhatdan:
@nalind PTAL

manifest/docker_schema1.go
```diff
@@ -95,14 +98,20 @@ func Schema1Clone(src *Schema1) *Schema1 {
 	return &copy
 }
 
-// initialize verifies invariants so that the rest of this code can assume a minimally healthy manifest.
+// initialize initializes ExtracedV1Compatibility and verifies invariants, so that the rest of this code can assume a minimally healthy manifest.
```
nalind: Has an extra "initialize" in this comment.

image/manifest_test.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+func TestManifestLayerInfosToBlobInfos(t *testing.T) {
+	blobs := manifestLayerInfosToBlobInfos([]manifest.LayerInfo{})
+	assert.Equal(t, []types.BlobInfo{}, blobs)
+
+	blobs = manifestLayerInfosToBlobInfos([]manifest.LayerInfo{
+		{
+			BlobInfo: types.BlobInfo{
+				MediaType: "application/vnd.docker.image.rootfs.diff.tar.gzip",
```
nalind: Not a blocker: this can be `manifest.DockerV2Schema2LayerMediaType`.

manifest/docker_schema1.go
mtrmac: This is “$function $does_something”; maybe `s/initializes/prepares/` to be prettier?  I will fix the typo in `ExtracedV1Compatibility` at least.

image/manifest_test.go
mtrmac: Yes, that would minimize the repetition.

On balance, I think I prefer the tests of interoperability with external protocols to hard-code values, at least some of the time, so that they detect unwanted changes to definitions of these constants — but it’s a weak preference, the repetition is not great either.

In this case it might even make more sense to use `"a"`,`"b"`,`"c"` or something similar to ensure that the values are passed through unchanged in the right order, but that would be even further from real-world usage.

mtrmac:
@nalind Thanks for the review.

> Assuming that we don't have to worry about empty uncompressed layers, this all LGTM.

Where should we worry about that? In an `ImageDestination`, we do not read the empty layer blob at all, so it should not matter whether it is compressed or uncompressed. In an `ImageSource`, `LayerInfosForCopy` replaces all empty layers with the canonical `manifest.GzippedEmptyLayer` representation, so again the original representation should not matter.

The only new thing I can think of is that pulling+pushing an image turns an uncompressed empty layer into a compressed one (unlike most layers, where pulling+pushing causes compressed layers to be uncompressed, or possibly re-compressed if the destination prefers compressed layers). Is that a problem?

nalind:
Ah, I forgot that we only ever have compressed layers in schema 1, as I was wondering about the `MediaType` field that `buildLayerInfosForCopy()` produces for empty layers possibly being incorrect.  The manifest `LayerInfos()` implementations don't appear to return media type information, though, so I think we end up just setting an empty value there.

Issue #475, opened 2018-07-06
Docker daemon source transport layer doesn't check for existing blob in case of oci destination


praveenkumar:
Is there any specific reason behind this?

```
$ skopeo copy docker-daemon:fedora:latest oci:fedora
Getting image source signatures
Copying blob sha256:891e1e4ef82ad02a4ea1f030831f942d722c7694c4db64ca3239c8163b811c58
 250.51 MB / 250.51 MB [===================================================] 16s
Copying config sha256:5415dde2eeab8b18b37d43db94983e3c8f5e3f79c9c5f64ded360ae0c38204d6
 1.31 KB / 1.31 KB [========================================================] 0s

$ skopeo copy docker-daemon:fedora:latest oci:fedora <= run it again and this doesn't check for existing blob but just start copy.
Getting image source signatures
Copying blob sha256:891e1e4ef82ad02a4ea1f030831f942d722c7694c4db64ca3239c8163b811c58
 250.51 MB / 250.51 MB [===================================================] 16s
Copying config sha256:5415dde2eeab8b18b37d43db94983e3c8f5e3f79c9c5f64ded360ae0c38204d6
 1.31 KB / 1.31 KB [========================================================] 0s
Writing manifest to image destination

$ sudo skopeo copy docker-daemon:fedora:latest containers-storage:fedora <= works as expected.
Getting image source signatures
Skipping fetch of repeat blob sha256:891e1e4ef82ad02a4ea1f030831f942d722c7694c4db64ca3239c8163b811c58
Copying config sha256:cc510acfcd701a409014118d5f417f0022520802a26c650866b8a9594d75f3a7
 2.29 KB / 2.29 KB [========================================================] 0s
Writing manifest to image destination
```

mtrmac:
“Yes” :)

Writing to OCI compresses the layers, which makes it impossible to check whether the original uncompressed layer is present when doing a second copy.

praveenkumar:
@mtrmac Is there any option to put un-compresses layer when using with oci destination or it is just a limitation for us right now?

mtrmac:
There isn't one right now.  What do you ultimately _need to do_?

praveenkumar:
> There isn't one right now. What do you ultimately need to do?

[...]

PR #478, opened 2018-07-12
Issue #475 Add flag for image compress layer for oci destination

praveenkumar:
@mtrmac have a look, hope that what you suggested on the issue.

### DIFF
--- a/oci/layout/oci_dest.go
+++ b/oci/layout/oci_dest.go
@@ -18,9 +18,10 @@ import (
 )
 
 type ociImageDestination struct {
-	ref           ociReference
-	index         imgspecv1.Index
-	sharedBlobDir string
+	ref                      ociReference
+	index                    imgspecv1.Index
+	sharedBlobDir            string
+	acceptUncompressedLayers bool
 }
 
 // newImageDestination returns an ImageDestination for writing to an existing directory.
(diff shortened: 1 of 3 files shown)


oci/layout/oci_dest.go
```diff
@@ -81,6 +83,9 @@ func (d *ociImageDestination) SupportsSignatures(ctx context.Context) error {
 }
 
 func (d *ociImageDestination) DesiredLayerCompression() types.LayerCompression {
+	if d.uncompress {
+		return types.PreserveOriginal
```
mtrmac: `types.PreserveOriginal` is _not_ the same thing as “decompress” (that would be `types.Decompress`). (And if I understand the situation correctly, you’d prefer “preserve original”, because that allows duplicate detection to work in more cases.)

So, name the boolean maybe `OCIAcceptUncompressedLayers`? `OCIDoNotForceLayerCompression`? And the private boolean accordingly.

praveenkumar: @mtrmac updated with `OCIAcceptUncompressedLayers` name.

mtrmac:
👍 

The tests are failing, please sign the commit per https://github.com/containers/image/blob/master/CONTRIBUTING.md (`git commit --amend -s`).

PR #481, opened 2018-07-14
Docker archive auto compression

mtrmac:
DO NOT MERGE YET: Completely untested in practice.

This is based on #427 , primarily adding the code to `.Close()` a `gzip.Reader`; also with a few more tests, and some stylistic touch-ups.


mtrmac:
FIXME: `tarfile.Source.GetBlob` fails to `underlyingStream.Close` on error paths.

mtrmac:
@runcom @rhatdan PTAL.  Correponding skopeo tests are in https://github.com/projectatomic/skopeo/pull/525 .

docker/tarfile/src.go
```diff
@@ -373,6 +373,12 @@ func (s *Source) GetBlob(ctx context.Context, info types.BlobInfo) (io.ReadClose
 		if err != nil {
 			return nil, 0, err
 		}
+		succeeded := false
+		defer func() {
+			if !succeeded {
```
rhatdan: Why not define return err error on the call and then just check if err != nil?

mtrmac: That’s elegant :) This particular method has another successful exit path (`info.Digest == s.configDigest`), so a bit risky.

For the same reason, the `succeeded` variable name is inaccurate, so I’ve at least renamed it to be right.

mtrmac:
> Tests are failing.

`make test-skopeo` failures are caused by adding a dependency; they should succeed in https://github.com/projectatomic/skopeo/pull/525 after Travis gets around to running them.

cyphar:
LGTM. Thanks for carrying this @mtrmac. :D

PR #482, opened 2018-07-19
Add DockerV2Schema2LayerUncompressedMediaType to manifest

TomSweeneyRedHat:

Adds the DockerV2Schema2LayerUncompressedMediaType , which is the uncompressed variant of DockerV2Schema2LayerMediaType and is used by Buildah.  Moving here to keep the manifest definitions for the tools in one central location.

Initial Buildah PR: https://github.com/projectatomic/buildah/pull/870

### DIFF
--- a/manifest/manifest.go
+++ b/manifest/manifest.go
@@ -24,6 +24,8 @@ const (
 	DockerV2Schema2ConfigMediaType = "application/vnd.docker.container.image.v1+json"
 	// DockerV2Schema2LayerMediaType is the MIME type used for schema 2 layers.
 	DockerV2Schema2LayerMediaType = "application/vnd.docker.image.rootfs.diff.tar.gzip"
+	// DockerV2Schema2LayerUncompressedMediaType is the Uncompressed MIME type used for schema 2 layers.
+	DockerV2Schema2LayerUncompressedMediaType = "application/vnd.docker.image.rootfs.diff.tar"
 	// DockerV2ListMediaType MIME type represents Docker manifest schema 2 list
 	DockerV2ListMediaType = "application/vnd.docker.distribution.manifest.list.v2+json"
 	// DockerV2Schema2ForeignLayerMediaType is the MIME type used for schema 2 foreign layers.


mtrmac:
The official spec at https://github.com/docker/distribution/blob/master/docs/spec/manifest-v2-2.md does not document such a type; why is it interoperable to use it at all?

[I’m not saying this is a bad idea; just that it’s not at all obvious to me.]

`docker/distribution` does define a `MediaTypeUncompressedLayer`, but it is only used internally during the push process to cause layers to be compressed when pushing them; when the image actually exists, it is always compressed.

(https://github.com/projectatomic/buildah/pull/870#issuecomment-406281346 refers to a “docker vendor” not that this is `buildah/docker`, the file modified in https://github.com/projectatomic/buildah/pull/870 , not `GitHub.com/docker/*`.)

---

`copy.Image` can compress/decompress layers, but does not adjust the MIME types at all.  That’s certainly a bug, but using encouraging any wider use of uncompressed MIME types in v2s2 images would make the situation even more pronounced; even if the `copy.Image` code were fixed to adjust MIME types, it’s arguable that it should not use or adjust the undocumented “uncompressed” MIME type.

mtrmac:
> Anywho, I did not look at the documentation, only the code in Docker: https://github.com/docker/distribution/blob/master/manifest/schema2/manifest.go#L33. They define the MediaTypeUncompressedLayer MediaType and I tried to massage the name in containers/image to emulate what's in that file.

Yes, that’s the one I’ve been looking at; I can’t see that the value ever exists outside the docker daemon process. I may well have overlooked something.

> So basically I just wanted to keep the MediaTypes that Buildah uses all in one central location. If you think adding this one will cause angst, I can keep it defined in Buildah.

I realize that this is an one-liner that would improve buildah consistency, and opening the wider conversation about whether to use the value at all may seem like an overreaction; still, when the usage were dubious, it would have to be be corrected at _some_ point. OTOH maybe someone knows much more about this MIME type and its safety/necessity than me (was it standardized in the past? Is there some other user in practice?) — in that case because `copy.Image` will need to learn about it, so having it documented would be valuable.

Either way c/image would benefit from knowing.

rhatdan:
I'm afraid that if you, @mtrmac, don't know then no one in here will.

mtrmac:
Let’s ask @nalind , who introduced it into `buildah`.

TomSweeneyRedHat:
reping @nalind

nalind:
buildah picked up the type because it was following the implementation more closely than the spec, and it's useful there because not compressing a new layer when we commit a new image to disk is visibly faster than compressing it, as its contents are going to have to get decompressed anyway. There's nothing more to that story, I'm afraid.

TomSweeneyRedHat:
This PR has withered on the vine too long, I'm going to close it and will leave the declarations as is.

PR #484, opened 2018-07-26
Accept a registry prefix of the form 'IP:port'

nalind:
My not-using-TLS registry is located at 172.30.1.1:5000, and I can't push images to it because attempts to parse a specification that uses the port number fail with a `error reading system registries configuration: invalid URL '172.30.0.1:5000': parse 172.30.0.1:5000: first path segment in URL cannot contain colon` error.  (The error may be specific to the standard library from Go 1.10.3.)  Removing the port number causes the entry in the insecure registries list to fail to be matched.

This patch checks for the case where the trimmed value is of the form 'IP:port', and skips the scheme check for those cases.

### DIFF
--- a/pkg/sysregistriesv2/system_registries_v2.go
+++ b/pkg/sysregistriesv2/system_registries_v2.go
@@ -3,8 +3,10 @@ package sysregistriesv2
 import (
 	"fmt"
 	"io/ioutil"
+	"net"
 	"net/url"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

rhatdan:
@mtrmac @runcom PTAL

rhatdan:
@nalind @mtrmac What is the latest on this PR?

mtrmac:
> What is the latest on this PR?

WIP https://github.com/mtrmac/image/tree/reference-additions got preempted by the copy speedup work. (Here the hard part is a good API, because changing the semantics after it is adopted would have been a nightmare; so far that branch adds the raw regexps, but we really need at least one utility function to go along with that.)

nalind:
This is a partial fix at best anyway - `atomic-registries` processes our `registries.conf` file to generate command line arguments to pass to `dockerd`, so we need to be able to not error out when we encounter something that would be valid if it were passed to `dockerd`.  That includes CIDR notation, which this doesn't fix.  For OpenShift cases, I think wildcards for hostnames and IP addresses would be useful.

vrothberg:
> @mtrmac @nalind @vrothberg What is the state of this PR? Is it valid? Is it replaced by some of the newer registry PRs?

Yes, see https://github.com/containers/image/pull/534#issuecomment-440559821. This is currently a real issue btw. Podman uses the old `sysregsitries` package while Buildah uses `sysregistriesv2` but as Podman vendors Buildah for `podman-build` the pull code for build is entirely different than the one for `podman-pull` ending up in something like this https://github.com/containers/image/pull/468#issuecomment-440667190.

rhatdan:
@vrothberg So lets get podman to sysregistriesv2.

vrothberg:
> @vrothberg So lets get podman to sysregistriesv2.

Absolutely. @mtrmac and I are currently putting our heads together to fix the remaining bits in `containers/image`, among them the issue of this PR.

mtrmac:
Actually we don’t have a direct replacement for this PR (which relaxes the parsing/validation); the other PRs deal with using the configuration _after_ it has been parsed.  
https://github.com/mtrmac/image/tree/reference-additions is a possible start but
 1. it is missing tests, at the very least
 2. arguably should expose a function instead of a regex variable
 3. does not change the parsing/validation code to take advantage of the added regex yet

rhatdan:
@nalind Needs a rebase.

PR #485, opened 2018-07-26
Define PolicyRequirementError as struct for better error detection

TomSweeneyRedHat:

Changes PolicyRequirementError from a string to a struct to allow for easier detection in Buildah.  Currently I need to do something like this in Buildah to detect the policy error:
```
if strings.Contains(err.Error(), "rejected by policy") {
```

With this change I can do something like:

```
      switch err.(type) {
                case *signature.PolicyRequirementError:
                        logrus.Debugf("TOM: 1  FOUND POLICY ERROR")
```
Which I think is cleaner/safer, I can then check the contains to verify it's the policy error I was expecting if I need to differentiate.

If I missed an easier/safer way to detect the error from Buildah with the current setup, please feel free to slap me silly.

### DIFF
--- a/signature/policy_eval.go
+++ b/signature/policy_eval.go
@@ -14,10 +14,20 @@ import (
 )
 
 // PolicyRequirementError is an explanatory text for rejecting a signature or an image.
-type PolicyRequirementError string
+type PolicyRequirementError struct {
+	message string
+}
 
-func (err PolicyRequirementError) Error() string {
-	return string(err)
+// NewPolicyRequirementError is used to create a new PolicyRequirementError
+func NewPolicyRequirementError(message string) *PolicyRequirementError {
+	return &PolicyRequirementError{
+		message: message,
+	}
+}
+
+// Error returns the message from a PolicyRequirementError
+func (e *PolicyRequirementError) Error() string {
+	return e.message
 }
 
 // signatureAcceptanceResult is the principal value returned by isSignatureAuthorAccepted.
@@ -271,7 +281,7 @@ func (pc *PolicyContext) IsRunningImageAllowed(ctx context.Context, image types.
 	reqs := pc.requirementsForImageRef(image.Reference())
 
 	if len(reqs) == 0 {
-		return false, PolicyRequirementError("List of verification policy requirements must not be empty")
+		return false, NewPolicyRequirementError("List of verification policy requirements must not be empty")
 	}
 
 	for reqNumber, req := range reqs {
--- a/signature/policy_eval_baselayer.go
+++ b/signature/policy_eval_baselayer.go
@@ -16,5 +16,5 @@ func (pr *prSignedBaseLayer) isSignatureAuthorAccepted(ctx context.Context, imag
 func (pr *prSignedBaseLayer) isRunningImageAllowed(ctx context.Context, image types.UnparsedImage) (bool, error) {
 	// FIXME? Reject this at policy parsing time already?
 	logrus.Errorf("signedBaseLayer not implemented yet!")
-	return false, PolicyRequirementError("signedBaseLayer not implemented yet!")
+	return false, NewPolicyRequirementError("signedBaseLayer not implemented yet!")
 }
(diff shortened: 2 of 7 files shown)


mtrmac:
Yeah, probably.

I was originally thinking something like this should work:
```go
package main

import (
	"errors"
	"fmt"
)

type PolicyRequirementError string

func (err PolicyRequirementError) Error() string {
	return string(err)
}

func returnPRE() error {
	return PolicyRequirementError("PRE")
}

type OtherString string

func (err OtherString) Error() string {
	return string(err)
}

func returnOtherString() error {
	return PolicyRequirementError("OtherString")
}

func returnOther() error {
	return errors.New("Other")
}

func main() {
	for _, fn := range []func() error{returnPRE, returnOtherString, returnOther} {
		err := fn()
		_, isPRE := err.(PolicyRequirementError)
		fmt.Printf("%s %v\n", err.Error(), isPRE)
	}
}
```
but as that example shows, it can’t tell the difference between `PRE` and `OtherString`.  It’s a bit surprising, then, how https://blog.golang.org/error-handling-and-go advocates for naming raw types like this.

(I’ll take a closer look later; for now, tests are failing.)

signature/policy_eval.go
```diff
@@ -14,10 +14,20 @@ import (
 )
 
 // PolicyRequirementError is an explanatory text for rejecting a signature or an image.
-type PolicyRequirementError string
+type PolicyRequirementError struct {
+	message string
+}
+
+// NewPolicyRequirementError is used to create a new PolicyRequirementError
+func NewPolicyRequirementError(message string) *PolicyRequirementError {
```
mtrmac: I don’t think external callers need to create their own instances; they can define their own error types.  So, make this private, please, or perhaps drop it completely and let everyone create `PolicyRequirementError{"…"}`.

mtrmac: Oh, and this can be a value type (not a pointer). A single-member struct is hopefully just as cheap as the member.

That way the `PolicyRequirementError` can be used in callers directly without having to know about the pointer.

TomSweeneyRedHat:
@mtrmac thx for the review comments.  I'm going to try to get the tests to green first with my original code, will address the comments with a push afterwards.

mtrmac:
> ```go
> func returnOtherString() error {
> 	return PolicyRequirementError("OtherString")
> }
> ```

Oops; that was a really stupid brain fart.

So, actually, `err.(signaturePolicyRequirementError)` should _already work_, and so does in my testing `switch err.(type) { case signature.PolicyRequirementError: }`.

Am I being stupid again and missing something?

TomSweeneyRedHat:
@mtrmac I thought I'd tried that combo that you're sugfesting and it failed.  But I went through a number of iterations and will verify that.  Thanks for the suggestion.

TomSweeneyRedHat:
OK, more playing.  I tried 
```
switch err.(type) { case signature.PolicyRequirementError: }
```
without the changes in this PR and it failed.  After digging around I discovered the value in err.(type) in this instance was "*errors.withStack".  So I switched to the new code in this PR and I still got that error back and did not expect that.  That led to more digging.

I think I found the solution, if I keep the code prior  to this change and do this:

```
       switch err := errors.Cause(err).(type) {      case signature.PolicyRequirementError: }
```

It works.  With the code in this PR it also works, but I need to add an '*' before the signature.

So long story short, I don't think I need to make these changes in order to get Buildah to recognize the error type and react appropriately after all.  

However it might be nice to spiff up the error raising around these routines a bit more.  Should I keep digging, or would you prefer to just let it rest as it's not a need?

Either way, thanks a bunch for the review and comments.  I've learn a bit more about golang errors than I knew last week.

mtrmac:
> However it might be nice to spiff up the error raising around these routines a bit more. Should I keep digging, or would you prefer to just let it rest as it's not a need?

I’m sure the reporting could be improved, but I can’t easily have an opinion about an unspecified direction.  Basically: If an application, as a caller of this code, needs more information, we should probably provide it.

rhatdan:
@TomSweeneyRedHat What do you want to do with this PR?

TomSweeneyRedHat:
Oops, I probably should have closed this a while back.  Doing so now.

PR #211, opened 2017-01-17
docker: mimic docker upstream registry authentication

runcom:
Signed-off-by: Antonio Murdaca <runcom@redhat.com>

PR #487, opened 2018-08-07
Fix requested token actions when deleting images

mtrmac:
We need `"pull"` to be able to read the manifest, and `"delete"` to delete it.

Note that until docker/distribution commit ccb839e0e30c3b6992fb4084dfd6550d0ddd4d1a (Jan 3 2017), the action checked for deleting manifests was `"*"`; this does not add it.

(Also, neither of these actions are documented at https://github.com/docker/distribution/blob/master/docs/spec/auth/scope.md . At least https://github.com/docker/distribution/blob/master/docs/spec/auth/jwt.md says that requesting permissions that the server does not grant should not be an error, so this kind of non-compliance is hopefully safe enough.)

This should not affect OpenShift, where the token is an API token and does not actually include any scopes or permissions; that is also probably why noone has noticed before.

@jwhonce @runcom PTAL

### DIFF
--- a/docker/docker_image_src.go
+++ b/docker/docker_image_src.go
@@ -310,7 +310,14 @@ func (s *dockerImageSource) getSignaturesFromAPIExtension(ctx context.Context, i
 
 // deleteImage deletes the named image from the registry, if supported.
 func deleteImage(ctx context.Context, sys *types.SystemContext, ref dockerReference) error {
-	c, err := newDockerClientFromRef(sys, ref, true, "push")
+	// docker/distribution does not document what action should be used for deleting images.
+	//
+	// Current docker/distribution requires "pull" for reading the manifest and "delete" for deleting it.
+	// quay.io requires "push" (an explicit "pull" is unnecessary), does not grant any token (fails parsing the request) if "delete" is included.
+	// OpenShift ignores the action string (both the password and the token is an OpenShift API token identifying a user).
+	//
+	// We have to hard-code a single string, luckily both docker/distribution and quay.io support "*" to mean "everything".
+	c, err := newDockerClientFromRef(sys, ref, true, "*")
 	if err != nil {
 		return err
 	}


TomSweeneyRedHat:
Maybe I messed something up in my compilation, but the change doesn't seem to fix the problem for me, it only changes the error.

```
(In Skopeo)
# make binary-local
# make install
 skopeo --tls-verify=true delete docker://docker.io/tomsweeneyredhat/testing:first
WARN[0000] '--tls-verify' is deprecated, please set this on the specific subcommand 
FATA[0000] Failed to delete /v2/tomsweeneyredhat/testing/manifests/sha256:184a1460ad7ff5a861e572da10087c32a60adc97fdb2a6df5694df483e24daec: {"errors":[{"code":"UNSUPPORTED","message":"The operation is unsupported."}]}
 (405 Method Not Allowed) 
```

mtrmac:
Yes, this is known, as noted in https://github.com/containers/skopeo/issues/531 . I’m sorry, I forgot to mention it here. When the server does not implement deletes at all, there’s not much we can do; the operation can succeed only against a registry which implements deleting images.

Running `skopeo --debug delete` should show that with this PR the code at least gets to the point of actually making a `DELETE` request, instead of failing on a `GET`.

We do at least one `skopeo delete` as part of skopeo’s integration tests (with deletion enabled), so this PR should hopefully not be _completely_ broken; OTOH that registry does not require authentication, so it is not useful for verifying this PR is an improvement. Maybe I should add a real test.

rhatdan:
LGTM

TomSweeneyRedHat:
Thanks for the info @mtrmac , I threw --debug on the command line and did indeed see the DELETE command get attempted when deleting from dockerhub.  It still failed though as you suspected.  I tried Quay.io too and got a 400 failure there and did not see the DELETE command.  Thought I'd throw it by you in case it's of concern for skopeo and/or something we should try adjusting on quay.io.  If I replace delete with inspect I see the info on the image fwiw.

```
skopeo --debug delete --creds tomsweeneyredhat:{mypassword} docker://quay.io/tomsweeneyredhat/myalpine:latest
DEBU[0000] Using registries.d directory /etc/containers/registries.d for sigstore configuration 
DEBU[0000]  Using "default-docker" configuration        
DEBU[0000]   Using file:///var/lib/atomic/sigstore      
DEBU[0000] Looking for TLS certificates and private keys in /etc/docker/certs.d/quay.io 
DEBU[0000] GET https://quay.io/v2/                      
DEBU[0000] Ping https://quay.io/v2/ err <nil>           
DEBU[0000] Ping https://quay.io/v2/ status 401          
FATA[0000] unexpected http code: 400, URL: https://quay.io/v2/auth?account=tomsweeneyredhat&scope=repository%3Atomsweeneyredhat%2Fmyalpine%3Apull%2Cdelete&service=quay.io 
```

runcom:
LGTM

rhatdan:
@runcom how do I get on the approved list for merging?

mtrmac:
>  I tried Quay.io too and got a 400 failure there and did not see the DELETE command. 

Yes, that does not look too good.  I’ll look into it.

mtrmac:
> I tried Quay.io too and got a 400 failure there and did not see the DELETE command.

So (keep in mind that we are in undocumented territory in all of this), it turns out that Quay requires the `push` action to be in the token, and it chokes on `delete`.  Meanwhile, https://github.com/containers/image/pull/211 forces us to hard-code the actions in the code, because `gcr.io` does not return an useful `WWW-Authenticate` challenge.

Luckily, it seems that both new docker/distribution and quay.io support `*` to mean “all actions”; and old docker/distribution explicitly required `*`. So, let’s try that.

mtrmac:
@runcom PTAL

runcom:
@mtrmac LGTM, up to you to merge now

mtrmac:
👍

PR #488, opened 2018-08-08
Get token to search V2 registries without creds

umohnani8:
Update the search functionality to be able to search registries
with a V2 endpoint by generating a token.

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -100,6 +100,14 @@ type authScope struct {
 	actions    string
 }
 
+type sendAuth int
+
+const (
+	v2Auth sendAuth = iota
+	v1Search
+	v1Basic
+)
+
 func newBearerTokenFromJSONBlob(blob []byte) (*bearerToken, error) {
 	token := new(bearerToken)
 	if err := json.Unmarshal(blob, &token); err != nil {
@@ -234,7 +242,7 @@ func CheckAuth(ctx context.Context, sys *types.SystemContext, username, password
 		return errors.Wrapf(err, "error creating new docker client")
 	}
 
-	resp, err := newLoginClient.makeRequest(ctx, "GET", "/v2/", nil, nil)
+	resp, err := newLoginClient.makeRequest(ctx, "GET", "/v2/", nil, nil, v2Auth)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


docker/docker_client.go
```diff
@@ -292,7 +293,7 @@ func SearchRegistry(ctx context.Context, sys *types.SystemContext, registry, ima
 		registry = dockerV1Hostname
 	}
 
-	client, err := newDockerClientWithDetails(sys, registry, username, password, "", nil, "")
+	client, err := newDockerClientWithDetails(sys, registry, username, password, "*", nil, registry)
```
umohnani8: @mtrmac what is the best way of generating the `remoteName` from the registry given by the user?

docker/docker_client.go
```diff
@@ -479,12 +480,13 @@ func (c *dockerClient) getBearerToken(ctx context.Context, realm, service, scope
...
 	}
-	tokenBlob, err := ioutil.ReadAll(res.Body)
+	tokenBlob := new(bytes.Buffer)
+	_, err := tokenBlob.ReadFrom(res.Body)
 	if err != nil {
 		return nil, err
 	}
 
-	return newBearerTokenFromJSONBlob(tokenBlob)
+	return newBearerTokenFromJSONBlob(tokenBlob.Bytes())
```
mtrmac: What does this change do? `ioutil.ReadAll` internally uses a `bytes.Buffer` in a very similar way.

docker/docker_client.go
mtrmac:
```console
$ curl -i  https://registry-1.docker.io/v2/_catalog
HTTP/1.1 401 Unauthorized
…
Www-Authenticate: Bearer realm="https://auth.docker.io/token",service="registry.docker.io",scope="registry:catalog:*"
…
```

Also looking at https://github.com/docker/distribution/blob/4a75b72fd3621c1892143614474ff2b05d30e4b3/registry/handlers/app.go#L942-L953 , `registry:catalog:*` seems to be the scope we want.

The current `remoteName`/`actions` format can not produce a `scope="registry:…`, so the `newDockerClientWithDetails` API will need to be changed.

Maybe, to reduce the number of unnamed parameters, make `authScope` a struct of `type, name, actions string` (to follow https://github.com/docker/distribution/blob/master/docs/spec/auth/scope.md), and have a pass a single `authScope` parameter to `newDockerClientWithDetails`.  (Or there could be a single `scope string` parameter, but the callers would have to worry about formatting; well-typed data concentrates that into a single place.)

mtrmac:
> make `authScope` a struct of `type, name, actions string`

Or maybe even `type, name string; actions []string`, to be extra correct.

docker/docker_client.go
mtrmac: > make `authScope` a struct of `type, name, actions string`

Or maybe even `type, name string; actions []string`, to be extra correct.

docker/docker_client.go
umohnani8: @mtrmac it was breaking and throwing an error "unknown end of JSON input" and when I changed it to read in the token like this, it stopped throwing the error.

mtrmac: That’s surprising.  Can you compare the two values passed to `newBearerTokenFromJSONBlob`?

umohnani8:
Closed in favour of https://github.com/containers/image/pull/502

PR #489, opened 2018-08-13
Add all certs.d/$host/*.crt CAs instead of only the last one

mtrmac:
(Really should add tests as well.)


rhatdan:
LGTM

TomSweeneyRedHat:
Change LGTM.
Maybe it's a lack of tea, but I'm not seeing how the title relates to the change.  It looks like you're still getting the same certs, but now only if they've not already been retrieved.
Test case or two would be good.

mtrmac:
> Maybe it's a lack of tea, but I'm not seeing how the title relates to the change. It looks like you're still getting the same certs, but now only if they've not already been retrieved.

The idea is supposed to be that
```go
tlsc.RootCAs = systemPool
```
overwrites the field, discarding CA certificates loaded by previous iterations of the loop.

But to be fair I haven’t actually tested that; so please hold on merging this until I write tests. This really is too important code not to have tests.

mtrmac:
Tests added (and they do show that this change is necessary).

runcom:
travis' complaining (???)

mtrmac:
> travis' complaining (???)

Yes, 
https://github.com/containers/storage/pull/210#issuecomment-416671810 . I can take a look tomorrow.

mtrmac:
👍

PR #491, opened 2018-08-21
Add manifest.Schema2HealthConfig.StartPeriod

nalind:
Add the StartPeriod field to the manifest.Schema2HealthConfig type.  It was added to the Docker API in 2016, but we picked up our definitions from a version that predated it.


mtrmac:
👍 pending tests.  Thanks!

PR #494, opened 2018-08-23
[WIP][RFC] torrent: allow pull/seed of OCI images using bittorent

giuseppe:
add support for pulling/seeding images through BitTorrent.

BitTorrent is used through the library: https://github.com/anacrolix/torrent

The quay.io registry has support for it, base the implementation on
how it is used there.

More details in the Skopeo PR that uses this branch: https://github.com/containers/skopeo/pull/540

### DIFF
--- a/docker/docker_image_src.go
+++ b/docker/docker_image_src.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/containers/image/docker/reference"
 	"github.com/containers/image/manifest"
+	"github.com/containers/image/torrent"
 	"github.com/containers/image/types"
 	"github.com/docker/distribution/registry/client"
 	"github.com/opencontainers/go-digest"
@@ -23,21 +24,35 @@ import (
 type dockerImageSource struct {
 	ref dockerReference
 	c   *dockerClient
+	t   *torrent.Client
 	// State
 	cachedManifest         []byte // nil if not loaded yet
 	cachedManifestMIMEType string // Only valid if cachedManifest != nil
+
+	trackers []string
 }
 
 // newImageSource creates a new ImageSource for the specified image reference.
 // The caller must call .Close() on the returned ImageSource.
 func newImageSource(sys *types.SystemContext, ref dockerReference) (*dockerImageSource, error) {
+	var t *torrent.Client
 	c, err := newDockerClientFromRef(sys, ref, false, "pull")
 	if err != nil {
 		return nil, err
 	}
+
+	if sys.DockerTryTorrent {
+		t, err = torrent.MakeClient(sys, logrus.GetLevel() == logrus.DebugLevel, false, 0)
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	return &dockerImageSource{
-		ref: ref,
-		c:   c,
+		ref:      ref,
+		c:        c,
+		t:        t,
+		trackers: sys.DockerTorrentTrackers,
 	}, nil
 }
 
@@ -49,6 +64,9 @@ func (s *dockerImageSource) Reference() types.ImageReference {
 
 // Close removes resources associated with an initialized ImageSource, if any.
 func (s *dockerImageSource) Close() error {
+	if s.t != nil {
+		s.t.Close()
+	}
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


docker/docker_image_src.go
```diff
@@ -23,21 +24,35 @@ import (
 type dockerImageSource struct {
 	ref dockerReference
 	c   *dockerClient
+	t   *torrent.Client
 	// State
 	cachedManifest         []byte // nil if not loaded yet
-	cachedManifestMIMEType string // Only valid if cachedManifest != nil
+	cachedManifestMIMEType string // Only valid if cachedManifest != ni
```
rhatdan: You seem to have dropped a character on the comment?

torrent/no_torrent.go
```diff
@@ -0,0 +1,39 @@
+// +build !containers_image_include_torrent
```
rhatdan: Why such a long name?   Why not just  !torrent?

torrent/torrent.go
```diff
@@ -0,0 +1,295 @@
...
+		return nil, err
+	}
+
+	return &Client{
+		c:   c,
+		dir: dir,
+		sys: sys,
+	}, nil
+}
+func (t *Client) getTorrentUrl(ctx context.Context, info types.BlobInfo, registry string, ref reference.Named) url.URL {
```
rhatdan: Nit: add line.

types/types.go
```diff
@@ -385,6 +385,11 @@ type SystemContext struct {
 	// Note that this field is used mainly to integrate containers/image into projectatomic/docker
 	// in order to not break any existing docker's integration tests.
 	DockerDisableV1Ping bool
+	// If set, then try to pull layers using bittorent
+	DockerTryTorrent bool
```
rhatdan: Do we have to call this Docker?

rhatdan:
@mtrmac PTAL

types/types.go
giuseppe: I've called it Docker as it is used by the "docker://" transport as the other options.  Will we keep calling it "docker://"? :-)

torrent/no_torrent.go
giuseppe: updated to `torrent`

types/types.go
rhatdan: Yes,  until their is a standard...

torrent/no_torrent.go
mtrmac: Because c/image is a library, and the tags share the namespace all over the resulting binary, incl. the 100MB ones OpenShift creates. I’d much prefer namespacing the tags for that reason.

(Stand-alone apps like podman do not have to worry about this, and can live with short tags.)

TomSweeneyRedHat:
LGTM, assuming happy tests

giuseppe:
I was waiting for more feedbacks, I think this can be very helpful for pulling the same image(s) to different nodes at the same time, OpenShift/Kubernetes can take advantage of it.  To be useful in a local network, it is required to setup a local tracker.  Who could we ask on how this can be used within OpenShift?  @wking?

rhatdan:
How about @smartclayton

rhatdan:
@giuseppe Needs a rebase.  Also @vrothberg PTAL

vrothberg:
Before going into the code, I'd like to discuss this feature a bit. I join the discussion very late, please excuse that.

@giuseppe, what is the motivation / use case for this PR? I've not heard of torrent being used in such context before. For caching images (also locally or in air-gapped environments) we can use mirrors and soon any kind of registry (when sysregistriesv2 is properly integrated into the tools).

Besides increasing the complexity of Skopeo's cli, my biggest fear is that adding (and maintaining) this feature might turn into a burden.

In other words: given the many alternatives in the wild (and the distribution spec), how would we promote this feature?

giuseppe:
@vrothberg I see it useful from a local network, where the same image(s) is pulled multiple times from different nodes.  Think of OpenShift where the same images are pulled at the same time by multiple nodes.

What I don't like so far and stopped me for pushing it further:

- need to setup a local tracker, use the tracker on quay.io is not very helpful.  Local setup of quay.io?
- only the `dir` storage can be used as a cache.  Although we could probably be using the local layers cache that is being worked on for Buildah.  @nalind would it work for this use case?

vrothberg:
> @vrothberg I see it useful from a local network, where the same image(s) is pulled multiple times from different nodes. Think of OpenShift where the same images are pulled at the same time by multiple nodes.

I don't think that this discriminates enough from what can be achieved by using mirrors or configuring the `registries.conf` via the prefix semantics. There's also ways to use proxies. Please, don't consider me as a blocker, I just want to share my concerns.

giuseppe:
> I don't think that this discriminates enough from what can be achieved by using mirrors or configuring the `registries.conf` via the prefix semantics. There's also ways to use proxies. Please, don't consider me as a blocker, I just want to share my concerns.

I agree, I am also not convinced of how useful BitTorrent can be compared to the cost of setting it up.

Although, if we take into account a  a mirror then we still need an additional entity in the network and configure every node to use it, similarly to having a BitTorrent tracker.  Then with a mirror all the load will be on the single registry, while with BitTorrent it will be shared among all the nodes that share the image(s).

rhatdan:
Is there any way we could setup a test, where we pull down a large image via bittorant to 10 different nodes and see if it improves the speed?

rhatdan:
@giuseppe @mtrmac @vrothberg What is the state of this PR?  Should we move forward or close it?

PR #498, opened 2018-08-27
Only run git-validation on commits not already present upstream

mtrmac:
Right now, `make .gitvalidation` takes 2 minutes on my system, mostly re-validating commits that have been validated many times before. That means slow `make`, and slow CI.

Instead, inspired by `skopeo`'s `hack/make/.validate`, only validate commits not present in the upstream master brach.  This, even with the added network access, takes less than two seconds on short branches.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,13 +91,13 @@ lint:
 		exit 1; \
 	fi
 
-EPOCH_TEST_COMMIT ?= e68e0e1110e64f906f9b482e548f17d73e02e6b1
-
 # When this is running in travis, it will only check the travis commit range
 .gitvalidation:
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
LGTM
Neat change!

rhatdan:
We should get this change into libpod and buildah.
LGTM

mheon:
For libpod, we update the validation epoch every weekly release, so we're
never testing more than a week's worth of commits at a time - seems to work
well.

mtrmac:
> We should get this change into libpod and buildah.

Turns out this is a bit noisy right now (https://github.com/vbatts/git-validation/issues/36 ) — not a showstopper, but probably nice to fix before copying this into another project.

PR #504, opened 2018-09-12
Return error when response code is not 200

umohnani8:
After trying the v2 search endpoint, if the response code is not
200, return an error instead of logging it.


umohnani8:
@mtrmac PTAL

umohnani8:
Nevermind, don't actually need this.

PR #505, opened 2018-09-17
bump(github.com/containers/storage)

nalind:
Update `github.com/containers/storage`, and modify `storageImageDestination.Commit()` to account for the change which makes it so that the `io.ReadCloser` that `store.Diff()` returns now holds a lock until it's closed, and because that's the same lock that `store.PutLayer()` takes, we need to finish reading it before we can do anything with its data, and we can't stream a layer's contents directly from one to an other.


rhatdan:
LGTM
@mtrmac PTAL

runcom:
LGTM

storage/storage_image.go
```diff
@@ -568,7 +568,35 @@ func (s *storageImageDestination) Commit(ctx context.Context) error {
 			if err2 != nil {
 				return errors.Wrapf(err2, "error reading layer %q for blob %q", layer, blob.Digest)
 			}
-			defer diff.Close()
+			// Copy the layer diff to a file.  Diff() takes a lock that it holds
+			// until the ReadCloser that it returns is closed, and PutLayer() wants
+			// the same lock, so the diff can't just be directly streamed from one
+			// to the other.
```
runcom: dumb image, are we going to be slowed down by this sort of buffering to disk?

runcom: I mean, slowed down to a point we care

nalind: Hard to say.  My guess is that computing the diff was/is the more expensive part of reading and reusing a layer's contents.  We hit this path when a layer that we have on disk matches the contents of a layer that we'd otherwise have to pull, and we can't just use the layer where it is because its inheritance chain isn't the same as the one we want.
I think that scenario becomes less and less likely as the size of the diff grows (both in the number of files and the size of their contents), and with smaller diffs, I hope the added cost of spooling the diff to disk and reading it back later is not going to be noticeable.

storage/storage_image.go
```diff
@@ -568,7 +568,35 @@ func (s *storageImageDestination) Commit(ctx context.Context) error {
...
+			// ctx.Done().
+			_, err = io.Copy(file, diff)
+			diff.Close()
+			file.Close()
+			if err != nil {
+				return errors.Wrapf(err, "error storing blob to file %q", filename)
+			}
+			// Make sure that we can find this file later, should we need the layer's
+			// contents again.
+			s.filenames[blob.Digest] = filename
```
mtrmac: From this point on the code is exactly the same as in the `// Check if we cached a file with that blobsum` case.

Could it be reused instead of copied, i.e. start with a
```go
if _, ok := s.filenames[blob.Digest]; !ok {
  // this code to create a temporary file
}
// Just always open s.filenames[blob.Digest]
```
?

storage/storage_image.go
```diff
@@ -568,7 +568,35 @@ func (s *storageImageDestination) Commit(ctx context.Context) error {
 			if err2 != nil {
 				return errors.Wrapf(err2, "error reading layer %q for blob %q", layer, blob.Digest)
 			}
-			defer diff.Close()
+			// Copy the layer diff to a file.  Diff() takes a lock that it holds
+			// until the ReadCloser that it returns is closed, and PutLayer() wants
+			// the same lock, so the diff can't just be directly streamed from one
+			// to the other.
+			filename := filepath.Join(s.directory, fmt.Sprintf("%d", atomic.AddInt32(&s.nextTempFileID, 1)))
```
mtrmac: (Absolutely non-blocking: consider adding a helper function so that there is only one place managing the `nextTempFileID` atomic value.)

vendor.conf
```diff
@@ -1,7 +1,7 @@
 github.com/containers/image
 
 github.com/sirupsen/logrus v1.0.0
-github.com/containers/storage master
+github.com/containers/storage 02db7cbb0bb082681f9855131fec8c8e1a3f4cb6
```
mtrmac: This will freeze the version of c/storage in our CI. Do you want that, and for c/image CI to only update by an explicit decision? (I don’t have a strong opinion either way.)

storage/storage_image.go
nalind: Yup, changing it.

storage/storage_image.go
nalind: Sounds simple enough.  Adding a `computeNextBlobCacheFile()` method.

vendor.conf
nalind: I don't have a strong opinion, either.  Dropping this.

nalind:
Updated.

rhatdan:
@runcom @mtrmac Can we get this merged?

runcom:
ping @mtrmac for lgtm and merge. we need this to unblock PRs in crio as well, if nothing is blocking, please merge this :)

nalind:
The certificates under `oci/layout/fixtures/accepted_certs` and `oci/layout/fixtures/rejected_certs` expired.  Added scripts that can be used to generate fresh ones, and generated some fresh ones.

rhatdan:
LGTM

storage/storage_image.go
```diff
@@ -533,19 +537,11 @@ func (s *storageImageDestination) Commit(ctx context.Context) error {
 			lastLayer = layer.ID
 			continue
 		}
-		// Check if we cached a file with that blobsum.  If we didn't already have a layer with
```
mtrmac: One last nit, please drop the
```go
var diff io.ReadCloser
```
about 30 lines above, to make it clear that not using the variable outside of the “Try to find the layer with contents matching that blobsum.” block is not an oversight.

LGTM otherwise.

mtrmac:
Also, thanks for handling the expired certificates!

storage/storage_image.go
nalind: Dropped it.  Thanks!

mtrmac:
LGTM pending tests (feel free to merge if I somehow didn’t). Thanks!

PR #508, opened 2018-09-28
dockerClient.detectProperties: debug log the error string

nalind:
When logging a debug message about an error that we get back when trying to ping a registry, log the error text, too.


rhatdan:
LGTM

mtrmac:
👍 Thanks!

PR #511, opened 2018-10-04
sysregistriesv2: rework registry matching

nalind:
For registries that are marked as insecure or blocked, don't bother checking if their URLs are valid or not.  Instead, expect to use them as patterns against which we'll match values that are passed to FindRegistry().  This is an alternate take on #484 which should help with https://github.com/containers/buildah/issues/1047.

### DIFF
--- a/pkg/sysregistriesv2/system_registries_v2.go
+++ b/pkg/sysregistriesv2/system_registries_v2.go
@@ -3,13 +3,18 @@ package sysregistriesv2
 import (
 	"fmt"
 	"io/ioutil"
+	"net"
 	"net/url"
+	"path"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"sync"
+	"unicode"
 
 	"github.com/BurntSushi/toml"
 	"github.com/containers/image/types"
+	"github.com/sirupsen/logrus"
 )
 
 // systemRegistriesConfPath is the path to the system-wide registry
@@ -142,11 +147,13 @@ func parseURL(input string) (string, error) {
 func getV1Registries(config *tomlConfig) ([]Registry, error) {
 	regMap := make(map[string]*Registry)
 
-	getRegistry := func(url string) (*Registry, error) { // Note: _pointer_ to a long-lived object
-		var err error
-		url, err = parseURL(url)
+	getRegistry := func(prefix string, isURL bool) (*Registry, error) { // Note: _pointer_ to a long-lived object
+		url, err := parseURL(prefix)
 		if err != nil {
-			return nil, err
+			if isURL {
+				return nil, err
+			}
+			url = prefix
 		}
 		reg, exists := regMap[url]
 		if !exists {
@@ -161,21 +168,21 @@ func getV1Registries(config *tomlConfig) ([]Registry, error) {
 	}
 
 	for _, search := range config.V1Registries.Search.Registries {
-		reg, err := getRegistry(search)
+		reg, err := getRegistry(search, true)
 		if err != nil {
 			return nil, err
 		}
 		reg.Search = true
 	}
 	for _, blocked := range config.V1Registries.Block.Registries {
-		reg, err := getRegistry(blocked)
+		reg, err := getRegistry(blocked, false)
 		if err != nil {
 			return nil, err
 		}
 		reg.Blocked = true
 	}
 	for _, insecure := range config.V1Registries.Insecure.Registries {
-		reg, err := getRegistry(insecure)
+		reg, err := getRegistry(insecure, false)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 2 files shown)


rhatdan:
@mtrmac PTAL

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -189,6 +196,235 @@ func getV1Registries(config *tomlConfig) ([]Registry, error) {
 	return registries, nil
 }
 
+// registryIsMatch returns true if flags in the registry entry indicate that
+// its URL might not be used as-is, but will have URLs matched against it in
+// order to selectively apply configuration flags to an unlisted URL.
+func registryIsMatch(reg Registry) bool {
+	return reg.Insecure || reg.Blocked
```
mtrmac: AFAICT this is essentially an “ignore parsing errors for this entry” predicate; that’s just not viable in the v2 model, where there is a single registry object with flags.  A registry being marked as `Insecure` does not at all mean that the `Prefix` or `URL` field won’t be used for pulls (see how `postProcessRegistries` enforces the `Insecure`/`Blocked` values to be consistent across all instances with the same `URL`.)

Unless I am missing something, this essentially removes any guarantees that the `URL`/`Prefix` are well-formed.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -338,6 +578,21 @@ func FindRegistry(ref string, registries []Registry) *Registry {
...
+		if registryIsMatch(r) {
+			matches, err := registryMatches(r.Prefix, ref)
+			if err != nil {
+				logrus.Debugf("error checking if %q matches %q: %v", ref, r.Prefix, err)
+			} else {
+				if matches {
+					length := len(r.Prefix)
+					if length > prefixLen {
+						reg = r
+						reg.URL = ref
```
mtrmac: This would break the `Prefix`/`URL` mapping mechanism.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -189,6 +196,235 @@ func getV1Registries(config *tomlConfig) ([]Registry, error) {
...
+// its URL might not be used as-is, but will have URLs matched against it in
+// order to selectively apply configuration flags to an unlisted URL.
+func registryIsMatch(reg Registry) bool {
+	return reg.Insecure || reg.Blocked
+}
+
+// registryMatches returns true if the candidate location matches the pattern.
+// Recognized patterns include:
+// * IPv4[/class][:port][/path]
+// * IPv6[/class][:port][/path]
```
mtrmac: (For better or worse, the `[ipv6]` syntax is not valid in Docker references; it will never match any acceptable input.)

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -189,6 +196,235 @@ func getV1Registries(config *tomlConfig) ([]Registry, error) {
...
+// * hostname[:port][/path]
+// where IPv4, IPv6, and hostname may include wildcard characters.  The path
+// is implicitly terminated with "/", and individual components may also
+// include wildcard characters.
+func registryMatches(pattern, candidate string) (bool, error) {
+	defaultRegistryPort := "5000"
```
mtrmac: The actual default is 443/80 - and Docker’s matching _does not_ treat the implied port and an explicit expression of the same port value as equivalent.

It can’t really treat them as equivalent because the 443/80 distinction only happens way later, after the “insecure” flag is 1) looked up 2) determined 3) an actual https connection attempt is made to try 443 even if marked as insecure.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -189,6 +196,235 @@ func getV1Registries(config *tomlConfig) ([]Registry, error) {
...
+		return false, nil
+	}
+	// Compare paths, only so far as the pattern supplies one
+	pdir, ppath := path.Split(ppath)
+	cdir, cpath := path.Split(cpath)
+	for pdir != "" {
+		if cdir == "" {
+			return false, nil
+		}
+		matches, err := path.Match(pdir, cdir)
```
mtrmac: Why introduce this at all? “just in case”? I can’t see how this can work with the `Prefix`/`URL` mapping.

pkg/sysregistriesv2/system_registries_v2_test.go
```diff
@@ -381,3 +382,290 @@ insecure = true`)
...
+		{"172.31.0.1/23/foo", "172.31.0.1/23", "", "/foo", false},
+		{"172.31.0.1/23:5000", "172.31.0.1/23", "5000", "", false},
+		{"172.31.0.1/23:5000/foo", "172.31.0.1/23", "5000", "/foo", false},
+
+		{"[::1/64]", "::1/64", "", "", false},
+		{"[::1/64]/foo", "::1/64", "", "/foo", false},
+		{"[::1/64]:5000", "::1/64", "5000", "", false},
+		{"[::1/64]:5000/foo", "::1/64", "5000", "/foo", false},
+
+		{"172.31.0.*", "172.31.0.*", "", "", false},
```
mtrmac: Why??

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -189,6 +196,235 @@ func getV1Registries(config *tomlConfig) ([]Registry, error) {
...
+		// but if they do, and they're invalid values, that's a problem
+		if _, err := strconv.ParseUint(i, 10, 8); err != nil {
+			return true
+		}
+	}
+	return false
+}
+
+// parseMatchPattern splits up matching information for registries.
+func parseMatchPattern(pattern string) (host, port, path string, err error) {
```
mtrmac: (These parsers are simultaneously a lot of code to maintain / modify, and very non-obvious about allowing/rejecting corner cases and invalid inputs.)

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -189,6 +196,235 @@ func getV1Registries(config *tomlConfig) ([]Registry, error) {
...
+// registryIsMatch returns true if flags in the registry entry indicate that
+// its URL might not be used as-is, but will have URLs matched against it in
+// order to selectively apply configuration flags to an unlisted URL.
+func registryIsMatch(reg Registry) bool {
+	return reg.Insecure || reg.Blocked
+}
+
+// registryMatches returns true if the candidate location matches the pattern.
+// Recognized patterns include:
+// * IPv4[/class][:port][/path]
```
mtrmac: Hum; `192.168.31.0/24` is a valid repository named `24` on the host `192.168.31.0`, so this is ambiguous.
(And `192.168.31.0/24:443` is an image tagged `443` in that repository!)

Also, the `Registry` object has not been designed with any patterns/groups in mind.

Arguably, the CIDR match is an useful feature that should be possible regardless — but then again, 1) this code does CIDR matches only against explicit IP addresses (rightly, because matching DNS names, at least with the Go TLS stack, can only work with a TOCTOU vulnerability), while Docker matches even DNS names, so it is not compatible anyway, and 2) marking whole networks insecure is just lazy and I we should not encourage that.

If this should be supported at all, it probably needs to be a separate `CIDRRange` type or something like that, not a `Registry` value.

@vrothberg WDYT?

vrothberg:
I guess I am missing some context as multiple issues seem to be addressed by this PR, which makes it a bit hard to discuss specific issues.

Overall, I am oposed to the proposed change as the complexity increases too much. Hence, I'd like to propose discussing if there are other ways. Many problems are linked to parsing, so I am more and more tempted to have barely any input validation and let things break at other places of the stack. Another pain point is that the current format doesn't allow URI schemes which complicates parsing.

I don't know of any tool using the "new" config format, so we might be able to change a few things to release us from some pain and to keep the code maintainable. Can we collect a set of requirements that the package needs to meet? Also a list of specific issues with the current version would be very helpful.

The sysregistriesv2 package partially originated from an older version of an upstream PR against Moby that we have open for a while. The design of the moby PR has changed in the meanwhile (see https://github.com/moby/moby/pull/34319#issuecomment-416317925). You can find the WIP patch implementing the new design here: https://github.com/vrothberg/moby/commit/b573b5e4d229887d486f94a39935f72d501588a5

Maybe, that's a direction we can go to or at least some source of inspiration.

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: > If this should be supported at all, it probably needs to be a separate CIDRRange type or something like that, not a Registry value.

I agree. Docker has dedicated config switches for CIDRs (see https://github.com/vrothberg/moby/commit/b573b5e4d229887d486f94a39935f72d501588a5#diff-73caf4b121449e3cc9df8ab7554b31baL11).

vrothberg:
> @mtrmac @nalind @vrothberg Is this PR Still valid or are the other registry works replacing it?

I think that the other works replace this proposal, see https://github.com/containers/image/pull/534#issuecomment-440559821

mtrmac:
Yeah, #534 (or maybe something even simpler) is closer to the path forward. That one does not add CIDR matching yet, but (if it should exist at all), as argued elsewhere here, CIDR matching and prefix matching can’t share the same config objects because the strings are ambiguous.

rhatdan:
Closing to concentrate on #534

PR #512, opened 2018-10-04
Ignore but log an error on IsNotExist errors reading root CAs

nalind:
When we encounter an IsNotExist error when reading root CA certificates as part of setting up a TLS client, log the error but continue.  Continue to bail if we encounter errors loading client credentials.  This should resolve https://github.com/containers/buildah/issues/1065.

### DIFF
--- a/pkg/tlsclientconfig/tlsclientconfig.go
+++ b/pkg/tlsclientconfig/tlsclientconfig.go
@@ -34,17 +34,25 @@ func SetupCertificates(dir string, tlsc *tls.Config) error {
 	for _, f := range fs {
 		fullPath := filepath.Join(dir, f.Name())
 		if strings.HasSuffix(f.Name(), ".crt") {
+			logrus.Debugf(" crt: %s", fullPath)
+			data, err := ioutil.ReadFile(fullPath)
+			if err != nil {
+				if os.IsNotExist(err) {
+					// Dangling symbolic link?
+					// Race with someone who deleted the
+					// file after we read the directory's
+					// list of contents?
+					logrus.Warnf("error reading certificate %q: %v", fullPath, err)
+					continue
+				}
+				return err
+			}
 			if tlsc.RootCAs == nil {
 				systemPool, err := tlsconfig.SystemCertPool()
 				if err != nil {
 					return errors.Wrap(err, "unable to get system cert pool")
 				}
 				tlsc.RootCAs = systemPool
-			}
-			logrus.Debugf(" crt: %s", fullPath)
-			data, err := ioutil.ReadFile(fullPath)
-			if err != nil {
-				return err
 			}
 			tlsc.RootCAs.AppendCertsFromPEM(data)
 		}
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
@mtrmac @runcom PTAL

pkg/tlsclientconfig/tlsclientconfig.go
```diff
@@ -44,6 +44,10 @@ func SetupCertificates(dir string, tlsc *tls.Config) error {
 			logrus.Debugf(" crt: %s", fullPath)
 			data, err := ioutil.ReadFile(fullPath)
 			if err != nil {
+				if os.IsNotExist(err) {
+					logrus.Errorf("error reading certificate %q: %v", fullPath, err)
```
TomSweeneyRedHat: I'm waffling between Warnf and Errorf here....  Thought I'd throw it by you for consideration.

TomSweeneyRedHat:
LGTM assuming happy tests.

pkg/tlsclientconfig/tlsclientconfig.go
rhatdan: Since this does not fail, I believe it should be a Warnf.

pkg/tlsclientconfig/tlsclientconfig.go
```diff
@@ -44,6 +44,10 @@ func SetupCertificates(dir string, tlsc *tls.Config) error {
 			logrus.Debugf(" crt: %s", fullPath)
 			data, err := ioutil.ReadFile(fullPath)
 			if err != nil {
+				if os.IsNotExist(err) {
```
mtrmac: \*shrug\* there is a reasonable case for the new behavior, but please at least add a comment above this noting that this can happen if `fullPath` is a dangling symlink (or during a race against something removing a file from the directory?); otherwise, because `fullPath` was created by listing the parent directory, it’s not immediately obvious how this can happen.

pkg/tlsclientconfig/tlsclientconfig_test.go
```diff
@@ -90,7 +90,7 @@ func TestSetupCertificates(t *testing.T) {
 	// Unreadable CA certificate
 	tlsc = tls.Config{}
 	err = SetupCertificates("testdata/unreadable-ca", &tlsc)
-	assert.Error(t, err)
+	assert.NoError(t, err)
```
mtrmac: Please add also tests for `tlsc` having the expected contents (either reuse the “RootCAs include SystemCertPool” condition from earlier in the test, or rearrange `SetupCertificates` to keep `tlsc.RootCAs == nil` and test for that.

pkg/tlsclientconfig/tlsclientconfig.go
nalind: Changed it.

pkg/tlsclientconfig/tlsclientconfig_test.go
nalind: Hmm, I don't think I quite understand what you're describing in that first option, so made changes to the second.

pkg/tlsclientconfig/tlsclientconfig.go
nalind: Added a comment mentioning both possibilities.

mtrmac:
👍 pending tests. Thanks!

nalind:
@runcom PTAL

PR #516, opened 2018-10-10
Include http.StatusText with every HTTP status code in errors

mtrmac:
e.g. `504 (Gateway timeout)` instead of just `504`.

Note that for unknown status codes this will output `99999 ()`; this seems not worth worrying about and building a helper for right now.

(Also fixes one instance of using `%q` for integer parameters, which results in gibberish.)


mtrmac:
> Note that for unknown status codes this will output `99999 ()`

(with the appropriate number, but extra parentheses; not a constant `99999` value, to be clear.)

rhatdan:
I still have no idea why we only have three people in the code-review/pullapprove, Where is this list being set?

mtrmac:
👍

PR #517, opened 2018-10-10
In debug logs of Docker v1 ping, only call err.Error() if err != nil

mtrmac:
For me, this seems to work and only results in a weird
```
Ping https://registry-1.docker.io/v1/_ping err <nil>
```
(with the last `(%#v)` element completely lost;

OTOH https://github.com/containers/buildah/issues/1079 reports a crash most likely attributable to this.


TomSweeneyRedHat:
LGTM

nalind:
LGTM

mtrmac:
👍

Issue #518, opened 2018-10-10
Need a way to learn the digest of (the manifest of) a copy of an image


nalind:
When we copy an image from one place to another, in cases where a format conversion is performed, we don't have a way to learn the digest of the written image's manifest.  OpenShift will want to know that value in order to be able to tell a runtime engine to use that exact version of the image.

While we can attempt to immediately use the destination reference to obtain an ImageSource, read the manifest from it, and compute its digest, if we do that we can't be really sure that there wasn't another image copied to the same reference in between.

I think we can work around this in buildah by wrapping reference and image destination types so that it can record the contents of manifests that are written to destinations, but it's probably simpler to extend copy.Image() to communicate that information to its caller somehow.

nalind:
@bparees FYI

mtrmac:
Makes sense; `copy.Image` could, I think, very reasonably just return the manifest digest as an extra return value. (Or maybe the raw manifest as a `[]byte`, to avoid the overhead of computing the manifest digest if the caller does not care.)

Right now we only copy a single image instance, so this is trivial. After true support for multi-arch images is added, it still seems reasonable to return the top-level manifest list, and the caller could then dig in to extract the per-arch manifests if necessary.

rhatdan:
@nalind @mtrmac @vrothberg What is the state of this issue? Since the referenced PR was merged can we close it?

vrothberg:
https://github.com/containers/image/pull/519 fixed that, closing.

PR #519, opened 2018-10-11
copy.Image(): return the image's manifest

nalind:
When copying an image, return the manifest of the new copy of the image, which may be different from the manifest in the image referred to by the source reference if a format conversion was required.  One possible approach for #518.

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -102,8 +102,9 @@ type Options struct {
 }
 
 // Image copies image from srcRef to destRef, using policyContext to validate
-// source image admissibility.
-func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef, srcRef types.ImageReference, options *Options) (retErr error) {
+// source image admissibility.  It returns the manifest which was written to
+// the new copy of the image.
+func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef, srcRef types.ImageReference, options *Options) (manifest []byte, retErr error) {
 	// NOTE this function uses an output parameter for the error return value.
 	// Setting this and returning is the ideal way to return an error.
 	//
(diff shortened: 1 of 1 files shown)


nalind:
@mtrmac, is this close to what you had in mind?  If it's way off, I can close this.

copy/copy.go
```diff
@@ -237,43 +238,43 @@ func (c *copier) copyOneImage(ctx context.Context, policyContext *signature.Poli
...
+	manifest, err = ic.copyUpdatedConfigAndManifest(ctx)
 	if err != nil {
 		logrus.Debugf("Writing manifest using preferred type %s failed: %v", preferredManifestMIMEType, err)
 		// … if it fails, _and_ the failure is because the manifest is rejected, we may have other options.
 		if _, isManifestRejected := errors.Cause(err).(types.ManifestTypeRejectedError); !isManifestRejected || len(otherManifestMIMETypeCandidates) == 0 {
 			// We don’t have other options.
 			// In principle the code below would handle this as well, but the resulting  error message is fairly ugly.
 			// Don’t bother the user with MIME types if we have no choice.
-			return err
+			return manifest, err
```
mtrmac: It makes a bit more sense to me to always return `nil` on errors — although in the one case when the manifest upload succeeds and only `PutSignatures` fails it could be argued the other way.

nalind: No worries, I was kind of ambivalent about it, too.  I'll change it.

nalind:
Happy to: https://github.com/containers/skopeo/pull/558

mtrmac:
👍 , thanks!

Let’s do the rebase/merge tomorrow to minimize the CI breakage time.

mtrmac:
@nalind This needs one more CI run after the `golint` failures anyway — could you rebase, please?

nalind:
Okay, rebased.

mtrmac:
Thanks! Please re-vendor 
https://github.com/containers/skopeo/pull/558 against `master`.

PR #522, opened 2018-10-16
Fix build failures regarding error functions

dongsupark:
There have been multiple build failures found when running `go test` through the git repo. Fix them to avoid test failures.

Signed-off-by: Dongsu Park <dongsu@kinvolk.io>

PR #523, opened 2018-10-22
travis: upgrade golang to 1.11.1

dongsupark:
Since Go 1.11, `go vet` [checks format strings in Printf wrappers a little stricter than previous versions][govet]. Also `go test` started to run `go vet` automatically. As a result, with Go 1.11 or newer, running `go test` causes `go vet` to fail in some cases.

That's why several format strings bugs had not been discovered until https://github.com/containers/image/pull/522. Travis CI did not catch the bugs, because the container image for travis CI is built based on Go 1.9.

To avoid such cases in the future, we should upgrade Golang to 1.11.1.

[govet]: https://golang.org/doc/go1.11#vet


TomSweeneyRedHat:
The changes LGTM in general, but thought there'd be some additional code changes account for the changes in go vet and gofmt.  Are there no coded changes necessary for this?

rhatdan:
LGTM

mtrmac:
> The changes LGTM in general, but thought there'd be some additional code changes account for the changes in go vet and gofmt. Are there no coded changes necessary for this?

That happened in #522 already.

mtrmac:
@rhatdan Is there a minimum Golang version we need to support, and if so, should that still be tested in CI?

rhatdan:
I think we should only worry about gofmt issues on the latest released version (Currently golang 1.11)

mtrmac:
Sure, but this PR does not affect just `gofmt`; it affects the entire test suite.

dongsupark:
Rebased, and added another commit to fix gofmt issues.
Now travis builds pass.

rhatdan:
LGTM

vrothberg:
Oups, I think this broke the build due to `go vet` being more strict.

vrothberg:
https://github.com/containers/image/pull/541 should fix it

dongsupark:
@vrothberg Oh, there was no problem when I rebased it two weeks ago.
That happened due to the recent changes from https://github.com/containers/image/pull/536 .
Thanks for catching & fixing it. :+1:

PR #524, opened 2018-10-23
Added OpenTracing for podman functions

sjug:
This is the `containers/image` side of the initial OpenTracing support [added to podman](https://github.com/containers/libpod/pull/1692).

My goal was to add spans to all common functions across the various libraries.

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -23,6 +23,7 @@ import (
 	"github.com/containers/image/types"
 	"github.com/klauspost/pgzip"
 	digest "github.com/opencontainers/go-digest"
+	opentracing "github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/vbauerster/mpb"
@@ -128,6 +129,9 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
 	//
 	// the defers in this routine will wrap the error return with its own errors
 	// which can be valuable context in the middle of a multi-streamed copy.
+	span, _ := opentracing.StartSpanFromContext(ctx, "copyImage")
+	defer span.Finish()
+
 	if options == nil {
 		options = &Options{}
 	}
@@ -216,6 +220,10 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
 // Image copies a single (on-manifest-list) image unparsedImage, using policyContext to validate
 // source image admissibility.
 func (c *copier) copyOneImage(ctx context.Context, policyContext *signature.PolicyContext, options *Options, unparsedImage *image.UnparsedImage) (manifestBytes []byte, retErr error) {
+	span, _ := opentracing.StartSpanFromContext(ctx, "copyOneImage")
+	span.SetTag("ref", "copier")
+	defer span.Finish()
+
 	// The caller is handling manifest lists; this could happen only if a manifest list contains a manifest list.
 	// Make sure we fail cleanly in such cases.
 	multiImage, err := isMultiImage(ctx, unparsedImage)
@@ -383,6 +391,9 @@ func (c *copier) Printf(format string, a ...interface{}) {
 }
 
 func checkImageDestinationForCurrentRuntimeOS(ctx context.Context, sys *types.SystemContext, src types.Image, dest types.ImageDestination) error {
+	span, _ := opentracing.StartSpanFromContext(ctx, "checkImageDestinationForRuntime")
+	defer span.Finish()
+
 	if dest.MustMatchRuntimeOS() {
 		wantedOS := runtime.GOOS
 		if sys != nil && sys.OSChoice != "" {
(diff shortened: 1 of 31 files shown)


TomSweeneyRedHat:
I think you need to add opentracing to the vendor.conf file and vendor it in.

```
image/sourced.go:10:2: cannot find package "github.com/opentracing/opentracing-go" in any of:
	/gopath/src/github.com/containers/image/vendor/github.com/opentracing/opentracing-go (vendor tree)
	/usr/local/go/src/github.com/opentracing/opentracing-go (from $GOROOT)
	/gopath/src/github.com/opentracing/opentracing-go (from $GOPATH)
Makefile:39: recipe for target 'build-internal' failed
make[1]: *** [build-internal] Error 1
```

TomSweeneyRedHat:
It looks like you set a tag for about three fourths of the functions.  I'm curious as to when you need a tag and when you don't.  I'd thought you'd want to tag all.

sjug:
Thanks @TomSweeneyRedHat, the initial goal was to differentiate the same function between various implementations. If it only had one then it wouldn't be unclear as to what function is being called. Also, in general I tried to tag all the methods, for the same general rationale.

I have added a new commit with a few more tags, but I'm open to suggestions as to how I should tag and how you'd suggest they be labeled. Cheers.

TomSweeneyRedHat:
@sjug The tagging that you've done seemed reasonable to me.  I couldn't see why the few that were untagged were left untagged.  I think it makes sense to tag most if not all of them.

TomSweeneyRedHat:
I'm seeing the change to vendor.conf, but don't see any vendor/github.com/opentracing/ files?   I think you'll need to run vndr and pull them in here too like in libpod.

rhatdan:
Correct the vendor should happen automatically.

TomSweeneyRedHat:
I'd trust @rhatdan over me on the vendor question here.  
LGTM, assuming happy tests then.

mtrmac:
> I'm seeing the change to vendor.conf, but don't see any vendor/github.com/opentracing/ files? I think you'll need to run vndr and pull them in here too like in libpod.

Not having a `vendor` is expected; libraries having their own `vendor` subdirectories make imported types not interoperable with types from “the same” packages vendored elsewhere. So far we’ve been able to get away with not having library-private vender packages at all.

mtrmac:
I’m completely unfamiliar with this and I’ll need to read up on this – for now, my first impression is “could we have the tracing only in the `copy` subpackage instead of copy&pasted in every single transport? I do realize that the transports can be used without involving the `copy` package.

Also, looking at https://play.golang.org/p/tHtcBGASxtz, it seems possible to get this down to one line with a helper—though I’m not at all sure that it’s worth the readability impact.

sjug:
@mtrmac, sure you can have it only at the top level function except then you have no visibility into the functions beneath it. The idea is to have the child spans all visible throughout the trace.

mtrmac:
> @mtrmac, sure you can have it only at the top level function except then you have no visibility into the functions beneath it. The idea is to have the child spans all visible throughout the trace.

I wasn’t suggesting to drop the tracing of e.g. `PutBlob` entirely, but to move it from the callee to the caller — because the codebase has, for the `types.ImageTransport` implementations, notably fewer callers than implementations.

mtrmac:
… but that does lose visibility when the same implementation is called from other callers, true.

sjug:
Getting a `containers/storage` vendor error, can I trigger retests here?

mtrmac:
I can trigger them, but the failure is not a fluke and would just reproduce. Let me try to fix it separately…

mtrmac:
#525 should fix the test failure, running tests again.

sjug:
@mtrmac The skopeo builds are failing due to the missing dependency. Does that mean I need to add it to skopeo as well?

mtrmac:
> The skopeo builds are failing due to the missing dependency. Does that mean I need to add it to skopeo as well?

Eventually, yes, per https://github.com/containers/skopeo#dependencies-management .

sjug:
@mtrmac Given the CI build here is failing building skopeo due to missing dependencies `opentracing/opentracing-go` shoudln't I add it now in order to get it to pass?

rhatdan:
@mtrmac Can you answer @sjug ?

rhatdan:
@sjug Needs a rebase.

rhatdan:
@sjug is this still needed, if so please rebase, otherwise I will close.

mtrmac:
> @mtrmac Given the CI build here is failing building skopeo due to missing dependencies `opentracing/opentracing-go` shoudln't I add it now in order to get it to pass?

(Add it to skopeo, to make tests pass here, I assume): That won’t work; `make vendor` in skopeo will just ignore a `vendor.conf` entry that has no users in the codebase. https://github.com/containers/skopeo#dependencies-management is the necessary process.

sjug:
Rebased, this arguably is the most important part part of the tracing as the binary components (in podman/buildah) just initiate the spans but without the ones in the library we don't know whats happening for most of the execution.

@mtrmac Could you link me to the necessary process, that anchor doesn't seem to exist (anymore)?

mtrmac:
> @mtrmac Could you link me to the necessary process, that anchor doesn't seem to exist (anymore)?

https://github.com/containers/skopeo/blob/master/CONTRIBUTING.md#dependencies-management

PR #526, opened 2018-10-26
Don't copy duplicate layers from containers-storage to docker, oci

steveb:
This adds an argument to LayerInfosForCopy so that the image source
can return compressed digests *if* that source stores compressed
digests and implements returning them.

For compressed destinations like docker and oci, copy calls
LayerInfosForCopy() twice, and uses both compressed and uncompressed
digests to call HasBlob. This allows layer copy operations to be
skipped when the target has the compressed layer.

In this commit only containers-storage honors the requested
LayerCompression, but other sources can implement this in later
changes.

Fixes https://github.com/containers/skopeo/issues/532
Fixes https://github.com/containers/image/issues/475
Fixes https://github.com/containers/libpod/issues/1233

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -128,8 +128,10 @@ type ImageSource interface {
 	GetSignatures(ctx context.Context, instanceDigest *digest.Digest) ([][]byte, error)
 	// LayerInfosForCopy returns either nil (meaning the values in the manifest are fine), or updated values for the layer blobsums that are listed in the image's manifest.
 	// The Digest field is guaranteed to be provided; Size may be -1.
+	// Implementations may ignore desiredLayerCompression and always return
+	// digests for there internal compression format
 	// WARNING: The list may contain duplicates, and they are semantically relevant.
-	LayerInfosForCopy(ctx context.Context) ([]BlobInfo, error)
+	LayerInfosForCopy(ctx context.Context, desiredLayerCompression LayerCompression) ([]BlobInfo, error)
 }
 
 // LayerCompression indicates if layers must be compressed, decompressed or preserved
@@ -259,8 +261,8 @@ type Image interface {
 	// LayerInfosForCopy returns either nil (meaning the values in the manifest are fine), or updated values for the layer blobsums that are listed in the image's manifest.
 	// The Digest field is guaranteed to be provided, Size may be -1 and MediaType may be optionally provided.
 	// WARNING: The list may contain duplicates, and they are semantically relevant.
-	LayerInfosForCopy(context.Context) ([]BlobInfo, error)
-	// EmbeddedDockerReferenceConflicts whether a Docker reference embedded in the manifest, if any, conflicts with destination ref.
+	LayerInfosForCopy(context.Context, LayerCompression) ([]BlobInfo, error)
+	// EmbeddedDockerReferenceConflicts whether a Dockeur reference embedded in the manifest, if any, conflicts with destination ref.
 	// It returns false if the manifest does not embed a Docker reference.
 	// (This embedding unfortunately happens for Docker schema1, please do not add support for this in any new formats.)
 	EmbeddedDockerReferenceConflicts(ref reference.Named) bool
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,7 @@ cross: vendor
 tools: tools.timestamp
 
 tools.timestamp: Makefile
-	@go get -u $(BUILDFLAGS) github.com/golang/lint/golint
+	@go get -u $(BUILDFLAGS) golang.org/x/lint/golint
 	@go get $(BUILDFLAGS) github.com/vbatts/git-validation
 	@go get -u github.com/rancher/trash
 	@touch tools.timestamp
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -369,7 +369,12 @@ func (ic *imageCopier) copyLayers(ctx context.Context) error {
 	srcInfos := ic.src.LayerInfos()
 	destInfos := []types.BlobInfo{}
 	diffIDs := []digest.Digest{}
-	updatedSrcInfos, err := ic.src.LayerInfosForCopy(ctx)
+	updatedSrcInfos, err := ic.src.LayerInfosForCopy(ctx, types.PreserveOriginal)
+	if err != nil {
+		return err
+	}
+
+	compressSrcInfos, err := ic.src.LayerInfosForCopy(ctx, types.Compress)
 	if err != nil {
 		return err
 	}
@@ -381,7 +386,7 @@ func (ic *imageCopier) copyLayers(ctx context.Context) error {
 		srcInfos = updatedSrcInfos
 		srcInfosUpdated = true
 	}
-	for _, srcLayer := range srcInfos {
+	for i, srcLayer := range srcInfos {
 		var (
 			destInfo types.BlobInfo
 			diffID   digest.Digest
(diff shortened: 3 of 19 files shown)


mtrmac:
Thanks for the PR.

## Limitations

> This allows layer copy operations to be
skipped when the target has the compressed layer.

… _in the destination repo_; even if the underlying storage is deduplicated, blob existence is actually referenced per-repository.  E.g.
```console
$ sudo docker run -p 5000:5000 registry # Leave it running
$ skopeo --debug copy --dest-tls-verify=false docker://busybox docker://localhost:5000/reg1
$ skopeo --debug copy --dest-tls-verify=false docker://busybox docker://localhost:5000/reg1
$ skopeo --debug copy --dest-tls-verify=false docker://busybox docker://localhost:5000/reg2
```
shows that the second copy to `reg1` detects that the blob exists at the destination and does not upload it, but the following copy to `reg2` does not detect the blobs as present, and uploads them again.

So, this PR does improve that one limited case of a duplicate upload to the _same_ repo from a `containers-storage:` source, but it
- does not fully fix https://github.com/containers/skopeo/issues/532 , which includes the case of a single blob in multiple repositories in the same registry ; same for https://github.com/containers/libpod/issues/1233 , which asks for exactly the same feature.
- does not AFAICS fix https://github.com/containers/image/issues/475 , which is primarily asks for handling this with the `docker-daemon:` source, at all.

## Possible alternative

A full solution also needs to, at the very least, handle cross-repo mounts within the destination registry (e.g. when pulling a base image from $registry/$base as part of a build process, and pushing back the result to $registry/$product), and that requires introducing a new database storing data about blobs known to exist in various repos.

And, in turn, such a database can then support reusing differently-compressed blobs with the same uncompressed digest.

Both is (too slowly, I’m afraid) taking shape in https://github.com/mtrmac/image/tree/blob-info-caching — but that implementations looks quite different from this.

I’m not sure investing into this smaller but limited approach in the mean time is worthwhile: sure, an anticipated/hypothetical more complete solution in the future can always revert/rework this code, that part is trivially cheap. But at least working out the signing threat model would need to happen for this PR as well before it’s merged.

So it’s a trade-off of doing some work (which may or may not be reusable) twice, to quickly get a very-limited version of the feature, vs. just doing it once and get a full version later. Given how limited this version is, this does not look quite attractive, but it’s always hard to guess about the future…

## Specific concerns

if the overall preference were to merge this right now instead of waiting for the fuller solution:

- At the very least, this needs to respect `canModifyManifest`.
- In general, interactions of these features with signing are pretty worrisome; it’s plausible that a remote registry contains a compressed payload designed to attack _some_ implementations, this one happens not to be exploitable, and pull+(push with signing) would cause the current user to reuse and sign the malicious version. So far I’m thinking that _both_ receiving signed images and sending signed images should deactivate any blob substitution mechanisms, as a conservative default; we can perhaps loosen this up later, for specific cases, after carefully considering the threats.
- AFAICS this breaks the `cachedDiffIDs` mechanism.

types/types.go
```diff
@@ -128,8 +128,10 @@ type ImageSource interface {
 	GetSignatures(ctx context.Context, instanceDigest *digest.Digest) ([][]byte, error)
 	// LayerInfosForCopy returns either nil (meaning the values in the manifest are fine), or updated values for the layer blobsums that are listed in the image's manifest.
 	// The Digest field is guaranteed to be provided; Size may be -1.
+	// Implementations may ignore desiredLayerCompression and always return
+	// digests for there internal compression format
 	// WARNING: The list may contain duplicates, and they are semantically relevant.
-	LayerInfosForCopy(ctx context.Context) ([]BlobInfo, error)
+	LayerInfosForCopy(ctx context.Context, desiredLayerCompression LayerCompression) ([]BlobInfo, error)
```
mtrmac: (`copy.Image` is really the only user of `LayerInfosForCopy`, so if that user needs to call this function twice, the API should just be changed to return two sets of values in a single call.)

steveb: How about something like the following, I can make the change when I get some positive feedback on the API:

type CopyLayerInfo struct {
  blobInfo BlobInfo
  compressedBlobInfo BlobInfo
}
LayerInfosForCopy(ctx context.Context) ([]CopyLayerInfo, error)

Implementations could choose to populate compressedBlobInfo or not

steveb:
We're doing our own cross-repo mount calls in the python which calls skopeo copy, so I'm not too concerned about relying on skopeo for that just yet.

We'd like to switch to buildah in our CI as soon as we can, but we can't without a fix for this issue. For this reason I'm hoping this shorter-term fix will be accepted.

> At the very least, this needs to respect canModifyManifest

I think I see where to fix this

> In general, interactions of these features with signing are pretty worrisome; it’s plausible that a remote registry contains a compressed payload designed to attack some implementations, this one happens not to be exploitable, and pull+(push with signing) would cause the current user to reuse and sign the malicious version. So far I’m thinking that both receiving signed images and sending signed images should deactivate any blob substitution mechanisms, as a conservative default; we can perhaps loosen this up later, for specific cases, after carefully considering the threats.

Am I right in thinking that the current HasBlob check has the same issue, so this doesn't make it worse?

> AFAICS this breaks the cachedDiffIDs mechanism.

Probably because I have no idea what cachedDiffIDs is for ;) could you elaborate enough to put me on the right track?

steveb:
Any thoughts on my comments? thanks

mtrmac:
> > In general, interactions of these features with signing are pretty worrisome; it’s plausible that a remote registry contains a compressed payload designed to attack some implementations, this one happens not to be exploitable, and pull+(push with signing) would cause the current user to reuse and sign the malicious version. So far I’m thinking that both receiving signed images and sending signed images should deactivate any blob substitution mechanisms, as a conservative default; we can perhaps loosen this up later, for specific cases, after carefully considering the threats.
> 
> Am I right in thinking that the current HasBlob check has the same issue, so this doesn't make it worse?

No; `HasBlob`/`ReapplyBlob` only reuse _exactly the same_ blob the user wanted to sign already.  (And if the remote server modified the blob, it would not match the expected digest.)

> > AFAICS this breaks the cachedDiffIDs mechanism.
> 
> Probably because I have no idea what cachedDiffIDs is for ;) could you elaborate enough to put me on the right track?

(The `diffID` computation, if `ic.diffIDsAreNeeded` is very expensive, so this attempts not to compute the value twice for the same blob.

The PR as is seems to be inconsistent about the keys to the `cachedDiffIDs` map; e.g. it’s possible to end up with `!haveBlob && !diffIDIsNeeded` (because `cachedDiffIDs[checkSrcInfo.Digest]` exists), and then the code reads and returns `cachedDiffIDs[srcInfo.Digest]` instead (which at a first glance may well be unavailable). Similarly the code setting `cachedDiffIDs[]` uses a different key than the code reading it on the `ReapplyBlob` path.

At the very least, it’s not immediately obvious that the code is consistent and correct, and the design of which keys are used in `cachedDiffIDs` needs to be carefully documented.

types/types.go
mtrmac: I’d just have `LayerInfosForCopy` return `(blobInfos, uncompressedBlobInfos, error)` so that transports that don’t implement it only need an one-word modification.

gilligan:
Hey @steveb , are you still working on this? I ended up here through an skopeo issue - skopeo reuploading all layers all the time is really painful so it would be great to see progress on this - kudos for your work on this so far!

rhatdan:
@vrothberg ptal

steveb:
I see the blob-info-cache branch has just landed https://github.com/containers/image/pull/536 which is a better framework for dedup fixes like this. Maybe @mtrmac or @vrothberg could give a summary of what further might be needed for containers-storage -> docker deduplication?

gilligan:
I’ve navigated between several of the involved issues but i’m afraid they go over my head - do the merged changes mean we are closer to skopeo being able to avoid pushing layers multiple times or does it just mean it’ll be done in different ways now?

Thanks 🙏

rhatdan:
@gilligan Yes that is the idea.  Try it out, although I am not sure if we shipped a version of skopeo with the support yet.  @runcom @mtrmac ?

gilligan:
I see vendor.conf from skopeo using containers/image from master so I suppose i could just build skopeo from master and give it a try..

rhatdan:
Excellent.

PR #527, opened 2018-10-30
docker: export functions and variables for a distribution-spec cert tests

dongsupark:
Export the following functions and variables to be used by tests for certification of the distribution-spec.

* DockerReference
* MakeRequest
* NewDockerClientFromRef
* SendAuth, V2Auth, NoAuth
* TagOrDigest

Does it make sense? If not, please let me know.
See also my work-in-progress pull request: https://github.com/kinvolk/ocicert/pull/1

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -100,17 +100,18 @@ type authScope struct {
 	actions    string
 }
 
-// sendAuth determines whether we need authentication for v2 or v1 endpoint.
-type sendAuth int
+// SendAuth determines whether we need authentication for v2 or v1 endpoint.
+type SendAuth int
 
 const (
-	// v2 endpoint with authentication.
-	v2Auth sendAuth = iota
+	// V2Auth is v2 endpoint with authentication.
+	V2Auth SendAuth = iota
 	// v1 endpoint with authentication.
 	// TODO: Get v1Auth working
 	// v1Auth
-	// no authentication, works for both v1 and v2.
-	noAuth
+
+	// NoAuth is no authentication, works for both v1 and v2.
+	NoAuth
 )
 
 func newBearerTokenFromJSONBlob(blob []byte) (*bearerToken, error) {
@@ -182,9 +183,9 @@ func dockerCertDir(sys *types.SystemContext, hostPort string) (string, error) {
 	return fullCertDirPath, nil
 }
 
-// newDockerClientFromRef returns a new dockerClient instance for refHostname (a host a specified in the Docker image reference, not canonicalized to dockerRegistry)
+// NewDockerClientFromRef returns a new dockerClient instance for refHostname (a host a specified in the Docker image reference, not canonicalized to dockerRegistry)
 // “write” specifies whether the client will be used for "write" access (in particular passed to lookaside.go:toplevelFromSection)
-func newDockerClientFromRef(sys *types.SystemContext, ref dockerReference, write bool, actions string) (*dockerClient, error) {
+func NewDockerClientFromRef(sys *types.SystemContext, ref DockerReference, write bool, actions string) (*dockerClient, error) {
 	registry := reference.Domain(ref.ref)
 	username, password, err := config.GetAuthentication(sys, reference.Domain(ref.ref))
 	if err != nil {
@@ -247,7 +248,7 @@ func CheckAuth(ctx context.Context, sys *types.SystemContext, username, password
 		return errors.Wrapf(err, "error creating new docker client")
 	}
 
-	resp, err := newLoginClient.makeRequest(ctx, "GET", "/v2/", nil, nil, v2Auth)
+	resp, err := newLoginClient.MakeRequest(ctx, "GET", "/v2/", nil, nil, V2Auth)
 	if err != nil {
 		return err
 	}
@@ -323,7 +324,7 @@ func SearchRegistry(ctx context.Context, sys *types.SystemContext, registry, ima
 		u.RawQuery = q.Encode()
 
 		logrus.Debugf("trying to talk to v1 search endpoint\n")
-		resp, err := client.makeRequest(ctx, "GET", u.String(), nil, nil, noAuth)
+		resp, err := client.MakeRequest(ctx, "GET", u.String(), nil, nil, NoAuth)
 		if err != nil {
 			logrus.Debugf("error getting search results from v1 endpoint %q: %v", registry, err)
 		} else {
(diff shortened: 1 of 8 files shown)


runcom:
makes sense to me, where we could learn more about the certification of distribution-spec?

dongsupark:
@runcom Recent discussions: https://github.com/opencontainers/distribution-spec/issues/24

runcom:
alright, it does makes sense to me. I just want to avoid exposing the low level of the docker client we use. Otherwise it's like using any other http client like curl. In case of docker/distribution, how are you dealing with that since they don't expose a raw client like us?

dongsupark:
@runcom Ok, I see.
To avoid exposing the internal docker client, how about doing like this:

1. Define a new wrapper function and a context struct in `docker/docker_client.go`
```
type SendRequestContext struct {
	Ref     types.ImageReference
	Method  string
	ReqPath string
	Headers map[string][]string
	Stream  io.Reader
	Auth    SendAuth
}

func ClientSendRequest(ctx context.Context, sys *types.SystemContext, reqCtx *SendRequestContext) (*http.Response, error) {
	dr, ok := reqCtx.Ref.(DockerReference)
	if !ok {
		return nil, fmt.Errorf("cannot get docker reference from %v", reqCtx.Ref)
	}

	client, err := newDockerClientFromRef(sys, dr, false, "pull")
	if err != nil {
		return nil, fmt.Errorf("%v", err)
	}

	res, err := client.MakeRequest(ctx, reqCtx.Method, reqCtx.ReqPath, reqCtx.Headers, reqCtx.Stream, reqCtx.Auth)
	if err != nil {
		return nil, fmt.Errorf("%v", err)
	}
	defer res.Body.Close()

	return res, nil
}
```

2. Then a caller would call it like this:
```
reqPath := filepath.Join("/v2", reference.Path(dr.DockerReference()), "manifests", refTail)

res, err := docker.ClientSendRequest(ctx, sys,
	&docker.SendRequestContext{
		Ref:     img.Reference(),
		Method:  "GET",
		ReqPath: reqPath,
		Headers: nil,
		Stream:  nil,
		Auth:    docker.V2Auth,
	})
```

Though some functions and variables like `DockerReference`, `V2Auth` still need to be exposed.
Would it be better than the current one?

mtrmac:
Thanks for the PR.

I am a bit unclear on what _exactly_ the tests are supposed to verify, and depending on that the desired API changes significantly; and the linked discussion does not really say.

- If the tests are intended to ensure that the tested registry implementation is interoperable with this `containers/image` library, as it is used by real-world callers, the tests should just use the public API (e.g., modulo error handling, `containers/image.docker.NewReference().NewImageSource().GetManifest()` as is; not the currently-private `makeRequest` that no real-world caller uses).
- If the tests are intended to ensure that a registry implementation conforms exactly to the HTTP API spec as written, then the tests should probably work directly at the HTTP API level; `containers/image` does, in general and where practical, accommodate / work around non-conforming server implementations (e.g. note how we don’t `DELETE` an aborted upload object, because that requires additional permissions with docker/distribution, or how `deleteImage` uses a not-correct-in-principle-but-working-in-practice action `*` instead of the probably right `delete`. And if further incompatibilities that are practical to work around arise in the future, containers/image could be extended to work around them as well, making some test results invalid.

If you want to test interoperability with c/image, that is _very_ welcome, but just use the public API as it would be called in practice, please.

(It is quite reasonable to contemplate that the API should be extended in some way, e.g. to make error values easier to semantically process in callers so that they can, just for illustration, reliably distinguish e.g. a missing image from a network timeout error. Quite likely that could be useful for the test suite as well. But this would _not_ go as far as exposing the whole HTTP communication so that a HTTP-level test suite could verify handling of corner cases, because real-world callers would not need that, and again, any error handling code would include workarounds for interoperability; a particular known pain point is that different registries report “image not found” in somewhat inconsistent ways, and c/image would probably prefer to report all of them as a single kind of Go error value so that the callers don’t have to care.)

Or, maybe — it somewhat looks as if what you really wanted was raw HTTP-level access for testing the behavior, but you’d like containers/image to handle authentication for you, so that each individual test can focus on its specific part of the API — and that you don’t care for the whole `ImageReference` abstraction, or for automatically looking up credentials and TLS configuration in `/etc` and elsewhere, because they make the test suite change behavior depending on where it’s run? That would be a fairly different conversation (more or less, expose `setupRequestAuth`).

mtrmac:
> That would be a fairly different conversation (more or less, expose setupRequestAuth).

… and while I can clearly see how that could allow code reuse, it would not quite be at the level of “using containers/image to test the registry”, because it would be using only a very small part of it.

Or, maybe (wildly speculating here), `c/image/docker` could get a new boolean in `types.SystemContext.DockerNoWorkarounds` that would be enabled in the test suite and disabled everywhere else? That might be practical to maintain.

Overall, though, https://github.com/opencontainers/distribution-spec/issues/24 really needs to decide whether they want to test whether _real-world_ interoperability with c/image, or strict spec conformance (or, sure, both independently).

runcom:
please all take a look at my concerns here as well https://github.com/opencontainers/distribution-spec/issues/24#issuecomment-435025421

dongsupark:
@mtrmac 
>> That would be a fairly different conversation (more or less, expose setupRequestAuth).
> Hum, that may not be quite what you want either, because setupRequestAuth does not use the scopes from the challenge, but client-provided ones: https://github.com/containers/image/blob/4b6d9cf918c1e79a605265568c93cb35185ad25d/docker/docker_client.go#L416-L423

Hmm, Interesting. The [scope provided by client is used](https://github.com/containers/image/blob/4b6d9cf918c1e79a605265568c93cb35185ad25d/docker/docker_client.go#L443).
Looking into https://github.com/containers/image/pull/211, I suppose it was done for supporting docker and gcr. 
I wanted to use `setupRequestAuth` to avoid writing duplicated lots of auth code on my own. I wanted to do so even if the actual communication was done on HTTP layer.
But in this case, I'm not sure any more how it can be done.

> … `DockerNoWorkarounds`, while possible in principle, is not trivial to pull off because it’s not immediately obvious which parts of the code are conformant and which parts are a workaround.

Thanks. Let me think.

rhatdan:
@dongsupark Needs a rebase.  Is this still something you are working on?

dongsupark:
No, I'm not working on it any more. I came up with a pure HTTP-based approach.
Will close. Thanks!

PR #528, opened 2018-10-31
Copy layers concurrently

steveb:
Copy layers concurrently                                                              
                                                                                      
This implements concurrent copying by using a simplified worker                       
pattern. Unless the source and destination both support concurrency,                  
only a single worker will perform blob transfers.                                     
                                                                                      
Concurrency support for transports are as follows:                                    
                                                                                      
Concurrency disabled:                                                                 
containers-storage, docker-daemon, ostree                                             
                                                                                      
Concurrency enabled:                                                                  
dir, docker, oci                                                                      
                                                                                      
Concurrency for source only:                                                          
docker-archive                                                                        
                                                                                      
Currently worker count is hard-coded to 4. It is hard to know what the                
optimal worker count is since its not obvious if a copy will be CPU,                  
network or storage bound. Maybe an OK compromise is min(4, cpucount).

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -324,6 +324,22 @@ func (c *copier) Printf(format string, a ...interface{}) {
 	fmt.Fprintf(c.reportWriter, format, a...)
 }
 
+// updateBarPrefix updates the progress bar prefix for a blob, optionally
+// finishing the operation for that bar
+func updateBarPrefix(bar *pb.ProgressBar, srcInfo types.BlobInfo, reason string, finish bool) {
+	if bar == nil {
+		return
+	}
+	if reason == "" {
+		reason = "    "
+	}
+	bar.Prefix(srcInfo.Digest.Encoded()[:12] + ": " + reason)
+	if finish {
+		bar.Finish()
+	}
+}
+
+
 func checkImageDestinationForCurrentRuntimeOS(ctx context.Context, sys *types.SystemContext, src types.Image, dest types.ImageDestination) error {
 	if dest.MustMatchRuntimeOS() {
 		wantedOS := runtime.GOOS
@@ -382,12 +398,33 @@ func (ic *imageCopier) copyLayers(ctx context.Context) error {
 		srcInfos = updatedSrcInfos
 		srcInfosUpdated = true
 	}
+
+	var bars []*pb.ProgressBar
 	for _, srcLayer := range srcInfos {
+		bar := pb.New(int(srcLayer.Size)).SetUnits(pb.U_BYTES)
+		bar.SetMaxWidth(80)
+		bar.ShowTimeLeft = false
+		bar.ShowPercent = false
+		updateBarPrefix(bar, srcLayer, "", false)
+		bar.Format("[\x00=\x00>\x00 \x00]")
+		bar.Start()
+		bars = append(bars, bar)
+	}
+	// only create a pool when not printing debug messages
+	var pool *pb.Pool
+	if logrus.GetLevel() < logrus.DebugLevel {
+		pool, err = pb.StartPool(bars...)
+		if err != nil {
+			return err
+		}
+	}
+	for i, srcLayer := range srcInfos {
 		var (
 			destInfo types.BlobInfo
 			diffID   digest.Digest
 			err      error
 		)
+		bar := bars[i]
 		if ic.c.dest.AcceptsForeignLayerURLs() && len(srcLayer.URLs) != 0 {
 			// DiffIDs are, currently, needed only when converting from schema1.
 			// In which case src.LayerInfos will not have URLs because schema1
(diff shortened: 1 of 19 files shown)


steveb:
This is just a work in progress, but some feedback would be greatly appreciated

mtrmac:
Thanks for the PR.

The transports, as they currently exist, have been implemented under the assumption that they are stateful and only one call is made at a time; just blindly running them in parallel is not going to work (as you have discovered), and making this reliable isn’t a matter of just testing this, it’s a matter of carefully reviewing/restructuring the code to be safe to run in parallel.

That could mean either adding quite a few locks all over the transports, or, probably more plausibly, splitting the `GetBlob`/`PutBlob` operations into a start+transfer+close components, where the start+close parts run serialized (or at least serialized within a transport), and the transfer component can be (maybe?) parallelized. Or maybe just letting transports describe themselves as parallel-safe/unsafe, but with `containers-storage:` clearly parallel-unsafe, that would not help the major use cases.

See also https://github.com/containers/image/pull/283 and I think there has been at least one or two other PRs with essentially the same idea.

BTW, this PR does address the progress bar problem, and that’s definitely useful bit of code that we should be able to take advantage of. (Disclaimer: I haven’t reviewed the code in detail.)

steveb:
I've just tested with a single worker and it seems to be as reliable as non-concurrent.

Yes, I think if sources and destinations can declare if they are parallel-capable, then workers will only be greater than 1 if both support it.

As a start I think docker, oci, and dir could be parallel capable. The registry API is just REST calls which are definitely concurrent capable. oci and dir are in theory just reading and writing files.

I realise containers-storage will take more work, but concurrent registry->registry copies are genuinely useful.

steveb:
Here is a revision which implements transport specific concurrency support

mtrmac:
> I've just tested with a single worker and it seems to be as reliable as non-concurrent.

Testing does not guarantee all cases are hit, and does nothing to ensure future updates won’t break the tested properties. The safety to run in concurrent needs to be an explicitly designed property.
 
> Yes, I think if sources and destinations can declare if they are parallel-capable, then workers will only be greater than 1 if both support it.
> 
> As a start I think docker, oci, and dir could be parallel capable. The registry API is just REST calls which are definitely concurrent capable.

That’s not even true right now (`dockerClient.tokenExpiration`), and with https://github.com/mtrmac/image/tree/blob-info-caching reading and writing blobs will include updating a local database. The same database uses would certainly be useful for `oci`, and arguably for `dir` as well (OTOH it’s being used as a debugging tool, so maybe not for `dir`).

If `docker` and `oci` and `docker-daemon` and `containers-storage` can’t be, or soon won’t be able to be, concurrent-capable using this approach, there will be very little benefit.

steveb:
Hopefully the new local database is designed to be concurrent safe.

I still think that if at any point in the future there is a goal to implement concurrent layer transfers, a reasonable first step would be landing something like this with every transport declaring that they don't support concurrency, otherwise how can there be progress?

Let me know if I should pursue this pull request, or propose the progress-bar commit on its own.

As an aside, in theory could a lockless containers-storage copy source be implemented just by doing a tar-split with overlay/<hash>/diff and overlay-layers/<hash>-tar-split.gz? The only issue I see with this would be deleting an image mid-copy. However I don't understand why mounts are needed to copy a layer out of containers-storage so I may be missing something in the design.

runcom:
> I still think that if at any point in the future there is a goal to implement concurrent layer transfers, a reasonable first step would be landing something like this with every transport declaring that they don't support concurrency, otherwise how can there be progress?

That, along with the progress bar being parallel-aware, is a nice addition. It just lacks context as, as Miloslav pointed out, the transports (src and destination) aren't meant to be parallel. An example is the docker_image_source GetBlob which hits the network to gather auth/properties and then proceed to fetch content. That isn't goroutine safe right and maybe that's just the simplest example to fix by first locking struct fields access when it needs to.
Apart from that, containers-storage is surely a big beast to tackle but, from a conversation I had with Miloslav, the PutBlob/HasBlob implementations already use temporary files which is good if we want to introduce parallelization as the final layers apply comes only on Commit(), which runs at the very end of the GetBlob/PutBlob/HasBlob code.

My take on this, to start making progress, and having in mind some priorities as well would be:

- re-use this PR as a backbone (progress bar + parallel capable introspections by transport)
- make docker-registry -> containers-storage parallel (docker_image_source GetBlob + containers-storage PutBlob/HasBlob plus some idea I have to speed up Commit() as well and process by applying layers in order as they come from netowork/store) (I've already started working on this and I'm now using this PR as backbone as well)
- the above can be viewed as a usual, normal `docker pull` to local storage, so the obvious next step would be being able to support pushes in both containers-storage and docker transport.
- meanwhile tackle your use case as well of registry->registry parallelization
- finally tackling the other transports.

Let me know your thoughts

mtrmac:
> It just lacks context as, as Miloslav pointed out, the transports (src and destination) aren't meant to be parallel.

Yes—and re:
> Hopefully the new local database is designed to be concurrent safe.

Even if the underlying persistent storage, both c/storage and the blob reuse database, are designed 100%safely, the in-process data structure is not automatically safe to share across goroutines just because the persistent file has locks. The two are different objects and need their own locking or other concurrency design.

---

BTW one other aspect to this: It needs to be _very clear_, both to reviewers and to people updating the code in the future, which data structures and parts of the code (at least to the function level, and ultimately to the level of individual lines) are or aren’t known/reviewed/intended to be safe to use concurrently. Sure, the per-transport opt-ins make it somewhat easier to build this piecewise, but even then, the code must _first_ be safe at that moment, and documented to be safe in the future, and _then_ the transport can opt in.

rhatdan:
@steveb @vrothberg @runcom @mtrmac Could we consolidate on one PR to try to move parallel pulls forward?

vrothberg:
> @steveb @vrothberg @runcom @mtrmac Could we consolidate on one PR to try to move parallel pulls forward?

I am currently working on exactly that :) But while testing I got side-tracked into vendor in new code into libpod for testing purposes.

steveb:
I won't be making any further changes to this branch, feel free to use the first commit for the progress bar behaviour

vrothberg:
> I won't be making any further changes to this branch, feel free to use the first commit for the progress bar behaviour

Thanks a lot for the great work!

PR #532, opened 2018-11-13
ostree: detect when running in an user namespace

giuseppe:
when we are running in an user namespace we cannot create devices, and
they are automatically skipped by the extractor.


mtrmac:
See 
https://github.com/containers/storage/pull/237#issuecomment-438517286 .

mtrmac:
https://github.com/containers/storage/pull/237 has made an equivalent change inside c/storage, so this is no longer necessary.

PR #534, opened 2018-11-20
Fix prefix matching in FindRegistry

mtrmac:
… so that `example.com/ns` does not match `example.com/n`.

This still allows `example.com:5000` to match `example.com`, which is unintended; that will get fixed eventually, DON'T RELY ON THE CURRENT BEHAVIOR.


vrothberg:
I think we can simplify many things by using some ideas from the Moby PR, namely to force prefixes to end with a "/". This way, we force users to specify prefixes at path boundaries (or just garbage that won't match anything) and no further complex parsing/validation is required.

Hence a proposal:
- Force prefixes to end with a "/". If they don't return an error.
- Use `strings.HasPrefix()` for matching.
- When reading the config, calculate `len(prefix)` once to avoid repeatedly calculating it for each call.

> This still allows example.com:5000 to match example.com, which is unintended; [...]

... is implicitly solved with the upper proposal.

mtrmac:
> I think we can simplify many things by using some ideas from the Moby PR, namely to force prefixes to end with a "/". 

(Couldn’t that be an internal implementation detail of `refMatchesPrefix`?)

AFAICS that would make it impossible to set up a mirror for one important repo (`docker.io/library/wordpress`) without having to set up a mirror for the whole parent namespace. That seems like a useful feature to support — though, to be fair, I don’t know for sure that it is necessary.

---

> * When reading the config, calculate `len(prefix)` once to avoid repeatedly calculating it for each call.

(BTW (I got curious:) , as can be seen e.g. looking at https://go.googlesource.com/go/+/master/src/cmd/cgo/out.go#1558 , `len(string)` is just a single memory access inside the string object, and checking something like
  ```
  $ go tool compile -d all -d=ssa/all/dump=g -S -W -asmhdr asmhdr -m -o t.output -v -w t.go
  ```
the compiler is perfectly capable of eliminating repeated calls of `len(s)`, so local variables like in this PR don’t buy anything (unless the memory changes in a compiler-opaque way, I suppose), and extra struct members to record string sizes are actually worse than just calling `len( struct.member)`.  So, I’ve been prematurely/incorrectly optimizing in this PR and the notably less readable code is not at all worth it; I’ll simplify it.)

vrothberg:
> > I think we can simplify many things by using some ideas from the Moby PR, namely to force prefixes to end with a "/".
> 
> (Couldn’t that be an internal implementation detail of `refMatchesPrefix`?)
> 
> AFAICS that would make it impossible to set up a mirror for one important repo (`docker.io/library/wordpress`) without having to set up a mirror for the whole parent namespace. That seems like a useful feature to support — though, to be fair, I don’t know for sure that it is necessary.

I am not oposed to this idea at all. Let's put that on the agenda for next week's meeting!

> > * When reading the config, calculate `len(prefix)` once to avoid repeatedly calculating it for each call.
> 
> (BTW (I got curious:) , as can be seen e.g. looking at https://go.googlesource.com/go/+/master/src/cmd/cgo/out.go#1558 , `len(string)` is just a single memory access inside the string object, and checking something like
> 
> ```
> $ go tool compile -d all -d=ssa/all/dump=g -S -W -asmhdr asmhdr -m -o t.output -v -w t.go
> ```
> the compiler is perfectly capable of eliminating repeated calls of `len(s)`, so local variables like in this PR don’t buy anything (unless the memory changes in a compiler-opaque way, I suppose), and extra struct members to record string sizes are actually worse than just calling `len( struct.member)`. So, I’ve been prematurely/incorrectly optimizing in this PR and the notably less readable code is not at all worth it; I’ll simplify it.)

Amazing job digging into that! I didn't expect the compiler to actually do that based on memory addresses but only on local variables. Awesome!

mtrmac:
(It turns out the compiler is not _quite_ as smart as I thought, so the `switch` is still valuable, but at least I got rid of the `prefixLen`/`refLen` variables. Anyway, that was enough fun…)

rhatdan:
@vrothberg Is this ready to merge?

vrothberg:
> @vrothberg Is this ready to merge?

This LGTM, yes. Thanks for the ping. If you are cool with it, feel free to merge.

vrothberg:
Commit d9d3e83 broke the build. Trying to fix it with https://github.com/containers/image/pull/541.

vrothberg:
Now we're good to go. @rhatdan WDYT?

PR #535, opened 2018-11-30
Return password of user logged in as well

umohnani8:
This will be used to authenticate any existing credentials
in podman login.


umohnani8:
@mtrmac @runcom PTAL

pkg/docker/config/config.go
```diff
@@ -88,16 +88,16 @@ func GetAuthentication(sys *types.SystemContext, registry string) (string, strin
 // GetUserLoggedIn returns the username logged in to registry from either
```
TomSweeneyRedHat: The comment mentions returning the username, but doesn't mention returning the password too and probably should.

umohnani8: Fixed. Thanks for pointing it out :)

rhatdan:
LGTM
@mtrmac @runcom @vrothberg PTAL

TomSweeneyRedHat:
LGTM too

mtrmac:
The implementation is straightforward enough, but the function starts looking rather similar to `GetAuthentication`.  Sure, there _is_ a difference in the set of files used for search, still…

While the reason for returning the credential is https://github.com/containers/libpod/pull/1905 , I don’t understand how that works / what it does: https://github.com/containers/libpod/pull/1905#discussion_r238068974 — and that might influence what the API of this module should be.

PR #536, opened 2018-12-03
Blob info caching + mount/reuse

mtrmac:
This adds types.BlobInfoCache recording the encountered compressed/uncompressed digest relationships and `docker://` blob locations, and using them to:
- Mount blobs on `docker://` destinations instead of copying the data again
- Reuse local layers on `containers-storage:` destinations instead of pulling them
- Avoid blob decompression when converting from schema1 just to figure out the DiffID values

The blob info cache is stored in `/var/lib/containers/blob-info-cache-v1.boltdb` for EUID=0 users, and `$XDG_DATA_HOME/containers/blob-info-cache-v1.boltdb` for unprivileged users.

See individual commit messages for more details.

WIP items:
- <s>When `canSubstitute`, `dockerImageDestination` makes a mount request even within the same repo, which is unnecessary.</s>
- <s>Documentation of the `BlobInfoCache` interface and possibly other code</s>
- Tests for corner cases of the BoltDB code
- Logging on failures in the BoltDB code

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/containers/image/image"
+	"github.com/containers/image/pkg/blobinfocache"
 	"github.com/containers/image/pkg/compression"
 	"github.com/containers/image/signature"
 	"github.com/containers/image/transports"
@@ -77,6 +78,7 @@ type copier struct {
 	reportWriter     io.Writer
 	progressInterval time.Duration
 	progress         chan types.ProgressProperties
+	blobInfoCache    types.BlobInfoCache
 }
 
 // imageCopier tracks state specific to a single image (possibly an item of a manifest list)
@@ -147,6 +149,10 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
 		reportWriter:     reportWriter,
 		progressInterval: options.ProgressInterval,
 		progress:         options.Progress,
+		// FIXME? The cache is used for sources and destinations equally, but we only have a SourceCtx and DestinationCtx.
+		// For now, use DestinationCtx (because blob reuse changes the behavior of the destination side more); eventually
+		// we might want to add a separate CommonCtx — or would that be too confusing?
+		blobInfoCache: blobinfocache.DefaultCache(options.DestinationCtx),
 	}
 
 	unparsedToplevel := image.UnparsedInstance(rawSource, nil)
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -633,7 +633,7 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
 	// === Process input through digestingReader to validate against the expected digest.
 	// Be paranoid; in case PutBlob somehow managed to ignore an error from digestingReader,
 	// use a separate validation failure indicator.
-	// Note that we don't use a stronger "validationSucceeded" indicator, because
+	// Note that for this check we don't use the stronger "validationSucceeded" indicator, because
 	// dest.PutBlob may detect that the layer already exists, in which case we don't
 	// read stream to the end, and validation does not happen.
 	digestingReader, err := newDigestingReader(srcStream, srcInfo.Digest)
@@ -669,8 +669,10 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
 
 	// === Deal with layer compression/decompression if necessary
 	var inputInfo types.BlobInfo
+	var compressionOperation types.LayerCompression
 	if canModifyBlob && c.dest.DesiredLayerCompression() == types.Compress && !isCompressed {
 		logrus.Debugf("Compressing blob on the fly")
+		compressionOperation = types.Compress
 		pipeReader, pipeWriter := io.Pipe()
 		defer pipeReader.Close()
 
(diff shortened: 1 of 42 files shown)


rhatdan:
@vrothberg @runcom PTAL

rhatdan:
@nalind PTAL

mtrmac:
Tests are failing here due to the API change, updated Skopeo is running the tests in https://github.com/containers/skopeo/pull/570 .

mtrmac:
> The blob info cache is stored in `/var/lib/containers/blob-info-cache-v1.boltdb` for EUID=0 users, and `$XDG_DATA_HOME/containers/blob-info-cache-v1.boltdb` for unprivileged users.

This might not be the right thing to do - apparently rootless buildah runs in a user namespace, which causes it to see EUID==0, and try to write to `/var/lib/containers`.  That does not _break_ builds because any error causes the code to fall back to memory-only caches, but it makes the optimizations unavailable.

pkg/blobinfocache/boltdb.go
```diff
@@ -0,0 +1,310 @@
...
+					}
+				}
+				if uncompressedDigestValue != primaryDigest {
+					res = bdc.appendReplacementCandidates(res, scopeBucket, uncompressedDigestValue)
+				}
+			}
+		}
+		return nil
+	}); err != nil { // Including os.IsNotExist(err)
+		return []types.BICReplacementCandidate{} // FIXME? Log?
```
rhatdan: Shouldn't we log everything except IsNotExist?

copy/copy.go
```diff
@@ -502,28 +502,19 @@ type diffIDResult struct {
...
+
+	// If we already have the blob, and we don't need to compute the diffID, then we don't need to read it from the source.
+	if !diffIDIsNeeded {
+		reused, blobInfo, err := ic.c.dest.TryReusingBlob(ctx, srcInfo)
 		if err != nil {
-			return types.BlobInfo{}, "", errors.Wrapf(err, "Error reapplying blob %s at destination", srcInfo.Digest)
+			return types.BlobInfo{}, "", errors.Wrapf(err, "Error trying to reuse blob %s at destination", srcInfo.Digest)
+		}
+		if reused {
+			ic.c.Printf("Skipping fetch of repeat blob %s\n", srcInfo.Digest)
```
rhatdan: Can you reword this?

Skipping blob %s fetch, already exists?
Is this something we actually need to see, or would it be better to just see what is pulled?

rhatdan:
@vrothberg @runcom @nalind @umohnani8 PTAL

rhatdan:
@mtrmac Is this still WIP?  Can we get it vendored into skopeo so we can get tests to pass?

mtrmac:
> @mtrmac Is this still WIP?

Yes, at least for documentation.

> Can we get it vendored into skopeo so we can get tests to pass?

Already exists in 
https://github.com/containers/skopeo/pull/570 .

pkg/blobinfocache/boltdb.go
```diff
@@ -0,0 +1,310 @@
...
+		}
+		b, err = b.CreateBucketIfNotExists([]byte(uncompressed.String()))
+		if err != nil {
+			return err
+		}
+		if err := b.Put([]byte(anyDigest.String()), []byte{}); err != nil { // Possibly writing the same []byte{} presence marker again.
+			return err
+		}
+		return nil
+	}) // FIXME? Log error?
```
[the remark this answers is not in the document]
mtrmac: Yes, probably. That’s one of the WIP parts.

copy/copy.go
mtrmac: (“fetch” in this message includes pushes.)

This can definitely be improved, at least `s/blob/layer/`, and maybe we don’t need the digests in the output at all. That’s all fairly independent from the functionality of this PR.

mtrmac: > (“fetch” in this message includes pushes.)

i.e. all of pull/push/copy go through this code path and may skip transferring a blob between transports here.

rhatdan: Does this just become noise on the screen, would the user care?  Seems to be more debugging information.

rhatdan:
I'm willing to take it without Documentation, so we can move the testing along. Then open a separate PR for Docs

Why is the tests failing?

pkg/blobinfocache/boltdb.go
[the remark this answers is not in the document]
mtrmac: (The a bit tricky aspect to this is that failures using the cache are _never_ fatal (which might not be a good idea, to be fair), so a failure could result in dozens of identical or similar error messages per copy; we probably want to throttle them somehow, or maybe even mark the cache as broken after a few failures.)

mtrmac:
Updated to avoid a no-op mount within a single repo, and added documentation to the interface/implementations; I can live with this being merged.

Tests fail here due to the API change, should eventually pass in https://github.com/containers/skopeo/pull/570 (and after merging this, a variant of https://github.com/containers/skopeo/pull/570 that vendors from `master` of this repo will need to be quickly merged).

rhatdan:
LGTM
@giuseppe @vrothberg @nalind @runcom @umohnani8 PTAL

runcom:
Reviewing this in a bit

copy/copy.go
runcom: what about `Blob transfer skip` - more generic than fetch and more valid in the context of pushes

runcom:
I had just a comment which we can live with for now/beginning of January.
This PR also passes our demo and code is pretty solid as well.

I'm gonna merge this, if you have anything against, please fill out issues/PRs

runcom:
Tests in skopeo pass as well https://github.com/containers/skopeo/pull/570

PR #539, opened 2018-12-05
sysregistriesv2: support CIDRs for insecure registries

umohnani8:
Add CIDR notation parsing support to the sysregistriesv2 pkg.
Will be used in CRI-O to support existinf Docker workloads.


umohnani8:
@mrunalp @runcom @mtrmac PTAL

Aside from the security vulnerability that we discussed @mtrmac, please let me know if there is a case I missed.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -162,13 +184,35 @@ func getV1Registries(config *tomlConfig) ([]Registry, error) {
...
+	indexConfigs := make(map[string]*IndexInfo)
 	regMap := make(map[string][]Registry)
 
 	for _, reg := range regs {
 		var err error
 
+		// Check if insecure registry is either CIDR notation or of form "host:port"
+		// and add it to either insecureRegistryCIDRs or indexConfigs
+		if reg.Insecure {
+			_, ipnet, err := net.ParseCIDR(reg.URL)
```
mtrmac: Cc: @vrothberg who was involved in earlier discussions.

`127.0.0.1/8` is, sadly, a valid namespace or repository name.  It also does not make sense to have the search/mirror functionality associated with CIDR ranges. So, at least in the _new_ format, CIDR configurations should be an explicit, separate component of the config file (i.e. in `tomlConfig`, add a separate `CIDRRanges` top-level field, or something like that — don’t make the semantics of `Registries` ambiguous.

(It’s acceptable for the compatibility code reading v1 to use such a heuristic.)

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -53,6 +54,27 @@ type Registry struct {
 	Prefix string `toml:"prefix"`
 }
 
+// IndexInfo holds information of registries in the form of "host:port"
+// and stores whether it is secure or not
+type IndexInfo struct {
```
mtrmac: There already is a `Registry`.

BTW the “index” name is a bit confusing — yes, I know Docker uses it.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -365,3 +413,42 @@ func loadRegistryConf(configPath string) (*tomlConfig, error) {
 	err = toml.Unmarshal(configBytes, &config)
 	return config, err
 }
+
+// IsSecureIndex returns true if indexName matches an element of CIDRs. indexName can be of form `host:port` or `host`
+// where the `host` part can be either a domain name or an IP address. If it is a domain name, then it will be
+// resolved to IP addresses for matching. If resolution fails, false is returned.
+func (s *SysRegistries) IsSecureIndex(indexName string) bool {
```
mtrmac: We have _just_ modified consumers of `sysregistriesv2` to use `FindRegistry` (note the prefix semantics - it’s not just host[:port] values!) and check `.Insecure` on the result.

It’s fine to change the API, but we can’t have two different ways to ask the same question which return different results, the way right now `IsSecureIndex` and `FindRegistry().Insecure` do.

My initial guess would be to somehow embed this into the `FindRegistry` prefix matching, with specific repo/namespace configs taking precedence, as they currently do, and then adding the CIDR matching roughly at the level of host:port matches (I guess, as a last fallback after that) — but that might well be the wrong way to do it.

pkg/sysregistriesv2/system_registries_v2.go
umohnani8: Added a separate component in the tomlConfig for CIDRRanges.

pkg/sysregistriesv2/system_registries_v2.go
umohnani8: Not sure what to name it, renamed to HostPortConfig for now.

pkg/sysregistriesv2/system_registries_v2.go
umohnani8: Doing this in `FindRegistry()` might be a tad bit complicated. If we want to be able to use `FindRegistry().Insecure` instead of `IsSecureIndex`, then we would have to convert the CIDR notation and host:port notation to type Registry. One was I was thinking was to add a field `isCIDR` and if that is true then you have a field `IPNet`, which would be populated. Just a thought.
I am on PTO from today onwards and will have limited access to internet as I am travelling. I can try to get back to this but since this is high priority could @mtrmac or @vrothberg please take over.

pkg/sysregistriesv2/system_registries_v2.go
mtrmac: I don’t think this needs to exist at all; host[:port][/suffix] entries can already be represented via `tomlConfig.Registries` / `SysRegistries.Registries`.

Non-CIDR values in `tomlConfig.CIDRRanges` should be treated as as syntax errors rather than recorded in a separate `SysRegistries.HostPortConfigs`.

pkg/sysregistriesv2/system_registries_v2.go
mtrmac: > Doing this in `FindRegistry()` might be a tad bit complicated. If we want to be able to use `FindRegistry().Insecure` instead of `IsSecureIndex`, then we would have to convert the CIDR notation and host:port notation to type Registry.

This is a great point; yes, `FindRegistry` would have to manufacture a `Registry` value for the host that matched the CIDR range (rather than return the whole range). That’s not _obviously_ the right thing to do, but it also seems not obviously unreasonable to me.

Alternatively we could instead design the API to “answer specific questions”, the way `IsSecureIndex` works (add `GetMirrors`/`IsBlocked`…), and probably eliminate `FindRegistry` and hide the `Registry` type altogether.

I slightly prefer the `FindRegistry` approach because it clearly centralizes the lookup semantics, and with plausible client implementations, the O(N) lookup of the matching configuration entry can happen only once instead on each `IsSecure`/`IsBlocked`/`GetMirrors`/… call.

But it’s not obviously the right answer, hoping for others to weigh in.

> One was I was thinking was to add a field `isCIDR` and if that is true then you have a field `IPNet`, which would be populated. Just a thought.

Do the callers of `FindRegistry` actually need to know whether the data matched via a CIDR at all? e.g. the current / soon-to-be-introduced callers just consult `.Insecure` and don’t use the `URL` entry at all; and `FindRegistry` could, when matching a CIDR, generate a `Registry` value that contains a value based on the specific IP address (not even the range) in `URL`/`Prefix`.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -365,3 +416,42 @@ func loadRegistryConf(configPath string) (*tomlConfig, error) {
...
+		return index.Secure
+	}
+
+	host, _, err := net.SplitHostPort(indexName)
+	if err != nil {
+		// assume indexName is of the form `host` without the port and go on.
+		host = indexName
+	}
+
+	addrs, err := net.LookupIP(host)
```
mtrmac: To reiterate, this is a TOCTOU vulnerability: an attacker with network access can override DNS to point to a known- or expected-to-be “insecure CIDR range” (127.0.0.1) during this lookup of any host name, causing `IsSecureIndex` to return false — and subsequently the `docker://` transport will make a non-TLS connection with another DNS lookup, at which point the attacker can override DNS to point at the attacker’s server and the client can’t tell that it is not talking to the legitimate registry.

I’m very tempted to demand gating this behind a `sysregistriesv2.YesIKnowCIDRLookupsAreInsecureAndBreakTLSAndIAcceptFullResponsibility` (💀) variable that CRI-O would have to set to opt into using this.

rhatdan: @mrunalp WDYT?

rhatdan:
@umohnani8 @mtrmac @vrothberg @mrunalp What is the state of this PR?  Should we rebase and move forward with it, or should we close it?

mtrmac:
The core idea of CIDR matching on IP addresses is fundamentally insecure with the Go HTTP stack; I’d very strongly prefer for this feature not to exist.

OTOH, me not wanting it in c/image does not quite stop CRI-O, or anyone else, from implementing it separately, just as insecurely (https://github.com/cri-o/cri-o/pull/577 ), and centralizing all of them in one implementation with one knob that could at least in principle be turned off is _a bit_ attractive — but it would also tempt other projects to add the CIDR knob just because it exists.

[It’s almost tempting to accept this PR as long as there is the earlier-suggested `sysregistriesv2.YesIKnowCIDRLookupsAreInsecureAndBreakTLSAndIAcceptFullResponsibility` opt-in, just to see CRI-O explicitly use that symbol. 💀]

On balance, CRI-O doesn’t need this to exist in c/image right now, and we can’t implement this securely, so, closing.

PR #545, opened 2018-12-14
parallel copying

vrothberg:
### Parallel progress bars and changed output
```
$ podman pull docker.io/nginx:latest
Trying to pull docker.io/nginx:latest...Getting image source signatures
Copying blob a5a6f2f73cd8: 21.44 MiB / 21.44 MiB [==========================] 8s
Copying blob 1ba02017c4b2: 21.18 MiB / 21.18 MiB [==========================] 8s
Copying blob 33b176c904de: 203 B / 203 B [==================================] 8s
Copying config 568c4670fa80: 5.88 KiB / 5.88 KiB [==========================] 0s
Storing signatures
568c4670fa800978e08e4a51132b995a54f8d5ae83ca133ef5546d092b864acf
```

### Performance improvement
Tests on my notebook show improvements between **6 and 16 percent** when pulling an image (tested with a remote and a local registry and varying sizes and amounts of layers).

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -13,6 +13,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/containers/image/docker/reference"
@@ -97,6 +98,11 @@ type dockerClient struct {
 	supportsSignatures bool
 	// Private state for setupRequestAuth
 	tokenCache map[string]bearerToken
+	// detectPropertiesError caches the initial error.
+	detectPropertiesError error
+	// detectPropertiesOnce is used to execuute detectProperties() at most once in
+	// in makeRequest().
+	detectPropertiesOnce sync.Once
 }
 
 type authScope struct {
@@ -545,9 +551,9 @@ func (c *dockerClient) getBearerToken(ctx context.Context, challenge challenge, 
 	return newBearerTokenFromJSONBlob(tokenBlob)
 }
 
-// detectProperties detects various properties of the registry.
-// See the dockerClient documentation for members which are affected by this.
-func (c *dockerClient) detectProperties(ctx context.Context) error {
+// detectPropertiesHelper performs the work of detectProperties which executes
+// it at most once.
+func (c *dockerClient) detectPropertiesHelper(ctx context.Context) error {
 	if c.scheme != "" {
 		return nil
 	}
@@ -602,6 +608,13 @@ func (c *dockerClient) detectProperties(ctx context.Context) error {
 		}
 	}
 	return err
+}
+
+// detectProperties detects various properties of the registry.
+// See the dockerClient documentation for members which are affected by this.
+func (c *dockerClient) detectProperties(ctx context.Context) error {
+	c.detectPropertiesOnce.Do(func() { c.detectPropertiesError = c.detectPropertiesHelper(ctx) })
+	return c.detectPropertiesError
 }
 
 // getExtensionsSignatures returns signatures from the X-Registry-Supports-Signatures API extension,
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -85,6 +85,7 @@ type dockerClient struct {
 	registry              string
 	client                *http.Client
 	insecureSkipTLSVerify bool
+
 	// The following members are not set by newDockerClient and must be set by callers if needed.
 	username      string
 	password      string
@@ -96,12 +97,12 @@ type dockerClient struct {
 	scheme             string // Empty value also used to indicate detectProperties() has not yet succeeded.
 	challenges         []challenge
 	supportsSignatures bool
-	// Private state for setupRequestAuth
-	tokenCache map[string]bearerToken
+
+	// Private state for setupRequestAuth (key: string, value: bearerToken)
+	tokenCache sync.Map
 	// detectPropertiesError caches the initial error.
 	detectPropertiesError error
-	// detectPropertiesOnce is used to execuute detectProperties() at most once in
-	// in makeRequest().
+	// detectPropertiesOnce is used to execuute detectProperties() at most once in in makeRequest().
 	detectPropertiesOnce sync.Once
 }
 
@@ -268,7 +269,6 @@ func newDockerClient(sys *types.SystemContext, registry, reference string) (*doc
 		registry:              registry,
 		client:                &http.Client{Transport: tr},
 		insecureSkipTLSVerify: skipVerify,
-		tokenCache:            map[string]bearerToken{},
 	}, nil
 }
 
@@ -479,14 +479,18 @@ func (c *dockerClient) setupRequestAuth(req *http.Request) error {
 				cacheKey = fmt.Sprintf("%s:%s", c.extraScope.remoteName, c.extraScope.actions)
 				scopes = append(scopes, *c.extraScope)
 			}
-			token, ok := c.tokenCache[cacheKey]
-			if !ok || time.Now().After(token.expirationTime) {
+			var token bearerToken
+			t, inCache := c.tokenCache.Load(cacheKey)
+			if inCache {
+				token = t.(bearerToken)
+			}
+			if !inCache || time.Now().After(token.expirationTime) {
 				t, err := c.getBearerToken(req.Context(), challenge, scopes)
 				if err != nil {
 					return err
 				}
 				token = *t
-				c.tokenCache[cacheKey] = token
+				c.tokenCache.Store(cacheKey, token)
 			}
 			req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", token.Token))
 			return nil
--- a/docker/docker_image_dest.go
+++ b/docker/docker_image_dest.go
@@ -111,6 +111,11 @@ func (c *sizeCounter) Write(p []byte) (n int, err error) {
 	return len(p), nil
 }
 
+// HasThreadSafePutBlob indicates whether PutBlob can be executed concurrently.
+func (d *dockerImageDestination) HasThreadSafePutBlob() bool {
+	return false
+}
+
 // PutBlob writes contents of stream and returns data representing the result (with all data filled in).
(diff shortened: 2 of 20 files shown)


vrothberg:
Skopeo and CRI-O CI are happy. I will perform some cleanups on each commit and once @giuseppe's PRs are merged, we can merge this one - assuming reviewers are happy.

docker/docker_client.go
```diff
@@ -95,8 +97,13 @@ type dockerClient struct {
 	scheme             string // Empty value also used to indicate detectProperties() has not yet succeeded.
 	challenges         []challenge
 	supportsSignatures bool
-	// Private state for setupRequestAuth
-	tokenCache map[string]bearerToken
+
+	// Private state for setupRequestAuth (key: string, value: bearerToken)
+	tokenCache sync.Map
```
mtrmac: “The Map type is specialized. Most code should use a plain Go map instead, with separate locking or coordination, for better type safety and to make it easier to maintain other invariants along with the map content.”

docker/docker_image_dest.go
```diff
@@ -119,11 +123,13 @@ func (c *sizeCounter) Write(p []byte) (n int, err error) {
 // to any other readers for download using the supplied digest.
 // If stream.Read() at any time, ESPECIALLY at end of input, returns an error, PutBlob MUST 1) fail, and 2) delete any data stored so far.
 func (d *dockerImageDestination) PutBlob(ctx context.Context, stream io.Reader, inputInfo types.BlobInfo, cache types.BlobInfoCache, isConfig bool) (types.BlobInfo, error) {
+	d.mutex.Lock()
+	defer d.mutex.Unlock()
```
mtrmac: This serializes `PutBlob` again, AFAICS; so what’s the point?

Or is it just to parallelize `dockerImageSource` but not `dockerImageDestination`? In that case the `IsThreadSafe` property might be a better fit for the `ImageSource`/`ImageDestination` interfaces instead of `ImageTransport`.

docker/docker_image_dest.go
```diff
@@ -31,6 +32,7 @@ type dockerImageDestination struct {
 	c   *dockerClient
 	// State
 	manifestDigest digest.Digest // or "" if not yet known.
+	mutex          sync.Mutex
```
mtrmac: Please always explicitly document what data is protected by each individual synchronization mechanism.

internal/testing/mocks/imagetransport.go
```diff
@@ -10,6 +10,11 @@ func (name NameImageTransport) Name() string {
 	return string(name)
 }
 
+// IsThreadSafe indicates whether the transport can be used concurrently.
+func (name NameImageTransport) IsThreadSafe() bool {
+	return false
```
mtrmac: Probably `panic` here instead.

storage/storage_image.go
```diff
@@ -717,16 +739,17 @@ func (s *storageImageDestination) Commit(ctx context.Context) error {
...
 		if _, err2 := s.imageRef.transport.store.DeleteImage(img.ID, true); err2 != nil {
 			logrus.Debugf("error deleting incomplete image %q: %v", img.ID, err2)
 		}
 		logrus.Debugf("error saving manifest for image %q: %v", img.ID, err)
 		return err
 	}
 	// Save the signatures, if we have any.
-	if len(s.signatures) > 0 {
-		if err := s.imageRef.transport.store.SetImageBigData(img.ID, "signatures", s.signatures); err != nil {
+	signatures := s.getSignatures()
```
mtrmac: This seems overzealous and/or insufficient. What is the lock in `getSignatures` protecting against?

Just adding locks around individual field accesses does not automatically make code thread-safe:

If there could be any user possibly racing against `Commit` at this time and concurrently modifying `s.signatures`, that would mean `Commit` might be working with possibly obsolete / incorrect data (the other user might be adding more signatures after `Commit` has read them — whether or not the point-in-time _read_ of `s.signatures` was atomic or not. The fix in that case must be to wait for that other user to finish, not to add a lock.

(Similar concerns about a few other fields.)

As mentioned elsewhere, please document precisely what fields is each lock is intended to protect, and perhaps in which situations. I’d _maybe_ rather have 50 lines of correct comments than 5 lines of unnecessary/misleading code.

vendor.conf
```diff
@@ -16,7 +16,7 @@ github.com/imdario/mergo 50d4dbd4eb0e84778abe37cefef140271d96fade
 github.com/mattn/go-runewidth 14207d285c6c197daabb5c9793d63e7af9ab2d50
 github.com/mistifyio/go-zfs c0224de804d438efd11ea6e52ada8014537d6062
 github.com/mtrmac/gpgme b2432428689ca58c2b8e8dea9449d3295cf96fc9
-github.com/opencontainers/go-digest aa2ec055abd10d26d539eb630a92241b781ce4bc
+github.com/opencontainers/go-digest c9281466c8b2f606084ac71339773efd177436e7
```
mtrmac: Is this really necessary for this PR? I can’t quickly see anything that makes a difference for concurrent execution in the changes.

vrothberg: I usually try not to apply unnecessary changes. We need this one to get the hash of the digest (see https://github.com/opencontainers/go-digest/blob/c9281466c8b2f606084ac71339773efd177436e7/digest.go#L135).

storage/storage_image.go
mtrmac: I.e., I suspect that the `IsTreadSafe` property without further qualifications is not something that can _ever reasonably_ be answered ”yes”, it’s never “safe” to call `PutSignatures` concurrently with `Commit`, locking or not, or to call _anything_ concurrently with `Close`.

So, maybe restrict the “transport is tread-safe” property only to an explicit subset of the interface (`GetBlob`/`PutBlob`/`TryReusingBlob`), explicitly document that (both in the documentation of the `IsTreadSafe` method, which might benefit from a more restrictive name then, and in the documentation of the methods that may need to be safe when executed concurrently).

vendor.conf
mtrmac: Ah, thanks. I missed that use of the new API.

storage/storage_image.go
vrothberg: It's only protecting against PutSignatues(). I will document things more precisely/carefully.
 
Note that by *thread-safe* I did not mean that anything can be executed in parallel. The order of execution must be controlled by the users. The biggest change is that PutBlob can be executed concurrently without screwing up state.

docker/docker_image_dest.go
vrothberg: I understand your questions as I should have made that clearer in the PR/commit messages. At the moment I focused on the `src: docker, dest:storage`. I preferred to make `IsThreadSafe()` per transport to keep things simpler but I can move it to `Source/Destination` or try to cover the vice versa case.

docker/docker_client.go
vrothberg: I am not sure if you like or dislike the change. I read the docs but my impression is that the code is easier to maintain and read as those maps are only used in this file at very few places. But for sure I can change it back and add per-map locks if you prefer.

vrothberg:
Thanks for the review @mtrmac!

docker/docker_client.go
vrothberg: I leave it for now as is. We can still move to use per-invariant locks or one to guard accesses to all which might be cheaper.

storage/storage_image.go
vrothberg: I agree. Giving a false impression would be pretty bad. I'll update accordingly.

docker/docker_image_dest.go
vrothberg: I was reflecting more on your concerns and believe that you really have a point. As we have discussed in other threads in this PR, the transports are only **partially thread-safe**. An `IsThreadSafe()` method is likely to trick users into believing those transports can be used concurrently. I decided to go for an `HasThreadSafe{Get,Put}Blob()` as this is what really matters in `Copy()` for now. This should limit the focus of this PR and aid in understanding what can be done with the transports and what not. A nice side-effect is that some insufficient locks can be removed as well.

vrothberg:
Alright, based on the feedback from @mtrmac, I rewrote parts of the PR.

- Instead of a per-transport, image destinations and image sources now indicate whether `{Get,Put}Blob()` can be executed concurrently. This makes the level of thread-safety more explicit and allows a more granular transition to concurrency.

- The commits are still split into small chunks to make reviewing a bit easier and to preserve the history.

- Pulling from a registry works in parallel while pushing to a registry is currently serialized by locks in containers/storage which must be addressed there.

@rhatdan @giuseppe PTAL

vrothberg:
For the record. The CI test PRs for ...
- Skopeo: https://github.com/containers/skopeo/pull/573
- CRI-O: https://github.com/kubernetes-sigs/cri-o/pull/1964

runcom:
LGTM

rhatdan:
Do we need to get the Travis to pass?  Since skopeo is updated?

runcom:
skopeo is not updated https://github.com/containers/skopeo/pull/573

talked to Valentin, he's taking care of all this

vrothberg:
I am currently making sure that all branches are up to date. I can coordinate with @runcom for merging :)

vrothberg:
Skopeo is updated and green. The last push here was just to update the commit message explaining that `golang.org/x/sync 42b317875d0fa942474b76e1b46a6060d720ae6e` is needed.

TomSweeneyRedHat:
Unhappy Travis here @vrothberg :^(

vrothberg:
> Unhappy Travis here @vrothberg :^(

It's the chicken-egg thing. I have a CI test PR over at Skopeo which passes with this c/image PR: containers/skopeo#573

runcom:
As long as skopeo passes with this PR, we're good and I can merge this one. Confirm? :)

vrothberg:
> As long as skopeo passes with this PR, we're good and I can merge this one. Confirm? :)

Yes, confirm :)

Issue #432, opened 2018-03-27
Make libselinux optional dependency


barthalion:
SELinux labeling has been introduced in 9b4510f6d1627c8e53c3303a8fe48ca7842c2ace for ostree, making libselinux required to build the library (and projects like buildah and skopeo that vendor it). Would it be possible to make it optional or hidden behind a build tag?

mtrmac:
All of `ostree:` is already optional, via the `containers_image_ostree_stub` tag. Is that sufficient for you, or are you actually trying to use `ostree:` without SELinux?

If the latter, I’m not sure whether it makes sense to make SELinux optional. @giuseppe ?

barthalion:
Rather the latter, unless ostree in context of skopeo/buildah is a niche feature. For now I just reverted linked commit, but I can't realistically keep doing it forever.

giuseppe:
The ostree backend (at least how it is used by atomic system containers) depends on SELinux as we need to label each file in the image.

@barthalion are you using the ostree backend in a different way?

Otherwise, as @mtrmac pointed out, I'd suggest you disable ostree altogether.

rhatdan:
Wouldn't the buildtag "selinux" control whether or not you need SELinux?

mtrmac:
Not in the c/image code as it exists today.

barthalion:
@giuseppe: My goal is to package podman for Arch (already did skopeo and buildah with a reverted change), so I guess it's up to the user.

I only scratched the surface of how things work. Do I understand correctly that labeling each file is needed for atomic containers for Atomic Host, but not necessarily for everyone?

giuseppe:
@barthalion yes it is needed only for atomic containers, not only on Atomic Host as they work on a non-ostree Fedora/CentOS/RHEL system as well.  I was just wondering what is the use case of the ostree backend if not used for the atomic containers and if in your case it would make more sense to just disable it.

barthalion:
I'm not sure if it's possible to solve that with cgo involved (from my cheer lack of knowledge about it) – but labels could be just skipped/ignored for distributions without selinux support. Security suffers, but I guess that atomic containers would still work that way?

giuseppe:
if the rest of the system has no selinux enabled/supported then atomic containers should still be able to work in theory, but that must be tested.

Could you give it a try and see if it works without selinux?

barthalion:
Sorry it took so long, got busy with life™ and other projects. With SELinux parts reverted, containers that ship with labels work fine.

[...]

PR #546, opened 2018-12-17
ostree: make libselinux optional

Silvanoc:
This pull request is more a draft for addressing issue #432 than a final realization, therefore please consider it a request for comments (RFC). I'd refactor a couple of places to have easier to maintain and more elegant code if I get positive feedback.

I don't have either a testing setup or use-case for SElinux features.

Decouple adding support for `libselinux` on build-time from requiring it on run-time.

More controlled linking of the library enables not failing if the library is supported, but missing.

Initial request done by @barthalion.

### DIFF
--- a/ostree/ostree_dest.go
+++ b/ostree/ostree_dest.go
@@ -23,6 +23,7 @@ import (
 	"github.com/containers/image/manifest"
 	"github.com/containers/image/types"
 	"github.com/containers/storage/pkg/archive"
+	"github.com/coreos/pkg/dlopen"
 	"github.com/klauspost/pgzip"
 	"github.com/opencontainers/go-digest"
 	selinux "github.com/opencontainers/selinux/go-selinux"
@@ -32,7 +33,7 @@ import (
 	"github.com/vbatts/tar-split/tar/storage"
 )
 
-// #cgo pkg-config: glib-2.0 gobject-2.0 ostree-1 libselinux
+// #cgo pkg-config: glib-2.0 gobject-2.0 ostree-1
 // #include <glib.h>
 // #include <glib-object.h>
 // #include <gio/gio.h>
@@ -41,6 +42,85 @@ import (
 // #include <gio/ginputstream.h>
 // #include <selinux/selinux.h>
 // #include <selinux/label.h>
+//
+// struct selabel_handle *
+//   my_selabel_open(
+//		void *function,
+//		unsigned int backend,
+//		const struct selinux_opt *opts,
+//		unsigned nopts)
+// {
+//		struct selabel_handle *(*ptr_selabel_open)(
+//			unsigned int,
+//			const struct selinux_opt *,
+//			unsigned);
+//		ptr_selabel_open = (struct selabel_handle *(*)(
+//				unsigned int,
+//				const struct selinux_opt *,
+//				unsigned))function;
+//		return ptr_selabel_open(backend, opts, nopts);
+// }
+//
+// void
+//   my_selabel_close(
+//		void *function,
+//		struct selabel_handle *handle)
+// {
+//		void (*ptr_selabel_close)(
+//			struct selabel_handle *);
+//		ptr_selabel_close = (void (*)(
+//				struct selabel_handle *))function;
+//		return ptr_selabel_close(handle);
+// }
+//
+// int
+//   my_selabel_lookup_raw(
+//		void *function,
+//		struct selabel_handle *handle,
+//		char **con,
+//		const char *key,
+//		int type)
+// {
+//		int (*ptr_selabel_lookup_raw)(
+//				struct selabel_handle *,
+//				char **,
+//				const char*,
+//				int);
+//		ptr_selabel_lookup_raw = (int (*)(
+//				struct selabel_handle *,
+//				char **,
+//				const char*,
+//				int))function;
+//		return ptr_selabel_lookup_raw(handle, con, key, type);
+// }
+//
+// int
+//   my_lsetfilecon_raw(
+//		void *function,
+//		const char *path,
+//		const char *con)
+// {
+//		int (*ptr_lsetfilecon_raw)(
+//				const char *,
+//				const char *);
+//		ptr_lsetfilecon_raw = (int (*)(
+//				const char *,
+//				const char *))function;
+//		return ptr_lsetfilecon_raw(path, con);
+// }
+//
+// void
+//   my_freecon(
+//		void *function,
+//		char *con)
+// {
+//		void (*ptr_freecon)(
+//				char *);
+//		ptr_freecon = (void (*)(
+//				char *))function;
+//		return ptr_freecon(con);
+// }
+//
 import "C"
 
 type blobToImport struct {
@@ -73,6 +153,8 @@ type ostreeImageDestination struct {
 	signaturesLen int
 	repo          *C.struct_OstreeRepo
 }
+
+var selinuxFunctions map[string]unsafe.Pointer
 
 // newImageDestination returns an ImageDestination for writing to an existing ostree.
 func newImageDestination(ref ostreeReference, tmpDirPath string) (types.ImageDestination, error) {
@@ -208,15 +290,17 @@ func fixFiles(selinuxHnd *C.struct_selabel_handle, root string, dir string, user
 			defer C.free(unsafe.Pointer(relPathC))
 			var context *C.char
 
-			res, err := C.selabel_lookup_raw(selinuxHnd, &context, relPathC, C.int(info.Mode()&os.ModePerm))
+			res, err := C.my_selabel_lookup_raw(selinuxFunctions["selabel_lookup_raw"],
+				selinuxHnd, &context, relPathC, C.int(info.Mode()&os.ModePerm))
 			if int(res) < 0 && err != syscall.ENOENT {
 				return errors.Wrapf(err, "cannot selabel_lookup_raw %s", relPath)
 			}
 			if int(res) == 0 {
-				defer C.freecon(context)
+				defer C.my_freecon(selinuxFunctions["freecon"], context)
 				fullpathC := C.CString(fullpath)
 				defer C.free(unsafe.Pointer(fullpathC))
-				res, err = C.lsetfilecon_raw(fullpathC, context)
+				res, err := C.my_lsetfilecon_raw(selinuxFunctions["lsetfilecon_raw"],
+					fullpathC, context)
 				if int(res) < 0 {
 					return errors.Wrapf(err, "cannot setfilecon_raw %s", fullpath)
 				}
@@ -433,12 +517,35 @@ func (d *ostreeImageDestination) Commit(ctx context.Context) error {
 	var selinuxHnd *C.struct_selabel_handle
 
 	if os.Getuid() == 0 && selinux.GetEnabled() {
-		selinuxHnd, err = C.selabel_open(C.SELABEL_CTX_FILE, nil, 0)
+		// dlopen.GetHandle will fail if libselinux.so is not available
+		// following block will be completely jump over and
+		// selinuxHnd = nil (important in fixFiles)
+		// so the absence of SElinux will be simply ignored
+		libselinuxHnd, err := dlopen.GetHandle([]string{"libselinux.so.1"})
+		if err != nil {
+			return errors.Wrapf(err, "SELinux enabled, but libselinux not available")
+		}
+
+		defer libselinuxHnd.Close()
+
+		selinuxFunctionNames := []string{"selabel_open", "selabel_close",
+			"selabel_lookup_raw", "lsetfilecon_raw", "freecon"}
+
+		for _, function := range selinuxFunctionNames {
+			ptrFunction, err := libselinuxHnd.GetSymbolPointer(function)
+			selinuxFunctions[function] = ptrFunction
+			if err != nil {
+				return fmt.Errorf(`couldn't get symbol %q: %v`, function, err)
+			}
+		}
+
+		selinuxHnd, err = C.my_selabel_open(selinuxFunctions["selabel_open"],
+			C.SELABEL_CTX_FILE, nil, 0)
 		if selinuxHnd == nil {
 			return errors.Wrapf(err, "cannot open the SELinux DB")
 		}
 
-		defer C.selabel_close(selinuxHnd)
+		defer C.my_selabel_close(selinuxFunctions["selabel_close"], selinuxHnd)
 	}
 
 	checkLayer := func(hash string) error {
--- a/ostree/ostree_dest.go
+++ b/ostree/ostree_dest.go
@@ -33,94 +33,98 @@ import (
 	"github.com/vbatts/tar-split/tar/storage"
 )
 
-// #cgo pkg-config: glib-2.0 gobject-2.0 ostree-1
-// #include <glib.h>
-// #include <glib-object.h>
-// #include <gio/gio.h>
-// #include <stdlib.h>
-// #include <ostree.h>
-// #include <gio/ginputstream.h>
-// #include <selinux/selinux.h>
-// #include <selinux/label.h>
-//
-// struct selabel_handle *
-//   my_selabel_open(
-//		void *function,
-//		unsigned int backend,
-//		const struct selinux_opt *opts,
-//		unsigned nopts)
-// {
-//		struct selabel_handle *(*ptr_selabel_open)(
-//			unsigned int,
-//			const struct selinux_opt *,
-//			unsigned);
-//		ptr_selabel_open = (struct selabel_handle *(*)(
-//				unsigned int,
-//				const struct selinux_opt *,
-//				unsigned))function;
-//		return ptr_selabel_open(backend, opts, nopts);
-// }
-//
-// void
-//   my_selabel_close(
-//		void *function,
-//		struct selabel_handle *handle)
-// {
-//		void (*ptr_selabel_close)(
-//			struct selabel_handle *);
-//		ptr_selabel_close = (void (*)(
-//				struct selabel_handle *))function;
-//		return ptr_selabel_close(handle);
-// }
-//
-// int
-//   my_selabel_lookup_raw(
-//		void *function,
-//		struct selabel_handle *handle,
-//		char **con,
-//		const char *key,
-//		int type)
-// {
-//		int (*ptr_selabel_lookup_raw)(
-//				struct selabel_handle *,
-//				char **,
-//				const char*,
-//				int);
-//		ptr_selabel_lookup_raw = (int (*)(
-//				struct selabel_handle *,
-//				char **,
-//				const char*,
-//				int))function;
-//		return ptr_selabel_lookup_raw(handle, con, key, type);
-// }
-//
-// int
-//   my_lsetfilecon_raw(
-//		void *function,
-//		const char *path,
-//		const char *con)
-// {
-//		int (*ptr_lsetfilecon_raw)(
-//				const char *,
-//				const char *);
-//		ptr_lsetfilecon_raw = (int (*)(
-//				const char *,
-//				const char *))function;
-//		return ptr_lsetfilecon_raw(path, con);
-// }
-//
-// void
-//   my_freecon(
-//		void *function,
-//		char *con)
-// {
-//		void (*ptr_freecon)(
-//				char *);
-//		ptr_freecon = (void (*)(
-//				char *))function;
-//		return ptr_freecon(con);
-// }
-//
+// Following block comment contains inline C code to make functions of
+// libselinux usable.
+
+/*
+#cgo pkg-config: glib-2.0 gobject-2.0 ostree-1
+#include <glib.h>
+#include <glib-object.h>
+#include <gio/gio.h>
+#include <stdlib.h>
+#include <ostree.h>
+#include <gio/ginputstream.h>
+#include <selinux/selinux.h>
+#include <selinux/label.h>
+
+struct selabel_handle *
+  my_selabel_open(
+		void *function,
+		unsigned int backend,
+		const struct selinux_opt *opts,
+		unsigned nopts)
+{
+		struct selabel_handle *(*ptr_selabel_open)(
+			unsigned int,
+			const struct selinux_opt *,
+			unsigned);
+		ptr_selabel_open = (struct selabel_handle *(*)(
+				unsigned int,
+				const struct selinux_opt *,
+				unsigned))function;
+		return ptr_selabel_open(backend, opts, nopts);
+}
+
+void
+  my_selabel_close(
+		void *function,
+		struct selabel_handle *handle)
+{
+		void (*ptr_selabel_close)(
+			struct selabel_handle *);
+		ptr_selabel_close = (void (*)(
+				struct selabel_handle *))function;
+		return ptr_selabel_close(handle);
+}
+
+int
+  my_selabel_lookup_raw(
+		void *function,
+		struct selabel_handle *handle,
+		char **con,
+		const char *key,
+		int type)
+{
+		int (*ptr_selabel_lookup_raw)(
+				struct selabel_handle *,
+				char **,
+				const char*,
+				int);
+		ptr_selabel_lookup_raw = (int (*)(
+				struct selabel_handle *,
+				char **,
+				const char*,
+				int))function;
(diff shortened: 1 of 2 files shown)


Silvanoc:
Unfortunately I sent the PR without rebasing on the top of the current master. I've done it and therefore I force-pushed.

rhatdan:
LGTM, although tests are failing.

ostree/ostree_dest.go
```diff
@@ -432,13 +516,34 @@ func (d *ostreeImageDestination) Commit(ctx context.Context) error {
...
-	if os.Getuid() == 0 && selinux.GetEnabled() {
-		selinuxHnd, err = C.selabel_open(C.SELABEL_CTX_FILE, nil, 0)
-		if selinuxHnd == nil {
-			return errors.Wrapf(err, "cannot open the SELinux DB")
-		}
+	// dlopen.GetHandle will fail if libselinux.so is not available
+	// following block will be completely jump over and
+	// selinuxHnd = nil (important in fixFiles)
+	// so the absence of SElinux will be simply ignored
+	libselinuxHnd, err := dlopen.GetHandle([]string{"libselinux.so"})
```
mtrmac: This must use a versioned soname.

Also, are we comfortable with making this a silent fail-open situation? I’m not really, I’d much prefer a build tag — but I can see the convenience.

(A random drive-by comment without a full review.)

Silvanoc: @mtrmac Thank you for your feedback!

> This must use a versioned soname.

You're fully right. Adding it.

> Also, are we comfortable with making this a silent fail-open situation? I’m not really, I’d much prefer a build tag — but I can see the convenience.

That's the kind of discussion that I wanted to trigger with this PR :smiley:

Unfortunately there is no way (like an entry in the configuration, for example) to detect if the user intended to use SELinux or not. Therefore I'll add a warning because it's a low-hanging fruit.

Another possibility would be adding a configuration entry like `selinux` defaulting to `true`. The an error would be yielded if the library is missing with the hint that the configuration entry should be changed if the absence of `libselinux` is on purpose.

I find that the second option has the best usability (less possibilities for unexpected behavior). But it also involves wider changes on the code than the 1st one, because it involves changes in `containers/storage`.

Any opinion on this?

Silvanoc:
The tests are currently failing due to issue #552.

ostree/ostree_dest.go
mtrmac: > Unfortunately there is no way (like an entry in the configuration, for example) to detect if the user intended to use SELinux or not. 

I don’t think we need to answer whether the user _intended to use_ SELinux; The thing _this implementation_ needs to know first, and the thing that is the most painful point for your use, is knowing whether `libselinux.so.1` _is available_ on the target system, whether or not the user enabled it:
- if it is not available, the implementation can’t link to it.
- if it is available, I think the implementation should _always_ link to it, and perhaps use the `libselinux`-determined “is SELinux enabled” configuration value to change its functionality (as the code already does). That’s the definite statement of user intent; we shouldn’t need an extra configuration option.

And “is `libselinux.so.1` expected to be available on the target system” is basically a build tag.  The downside is that every library linking to c/image (that cares about non-SELinux systems) will need to have a `Makefile` possibly setting the build tag.

Silvanoc: > I don’t think we need to answer whether the user intended to use SELinux; The thing this implementation needs to know first, and the thing that is the most painful point for your use, is knowing whether libselinux.so.1 is available on the target system, whether or not the user enabled it:
> 
> if it is not available, the implementation can’t link to it.
> if it is available, I think the implementation should always link to it, and perhaps use the libselinux-determined “is SELinux enabled” configuration value to change its functionality (as the code already does). That’s the definite statement of user intent; we shouldn’t need an extra configuration option.

I agree with the conditional that you are stating. But currently the decision "is SELinux enabled" is being taken at build-time: if buildtag `selinux` is not enabled, then `vendor/github.com/opencontainers/selinux/go-selinux/selinux_stub.go` gets used and `selinux.GetEnabled() ` returns false. So it doesn't have anything to do with the user intent, it's the builder/packager intent.

That's exactly the point of my patch, avoiding that *implementation needs to know first*.
1. With the current code, SELinux usage is expected if [it was built with support for it](https://github.com/containers/image/pull/546/files#diff-952ad9dc01673ce7143c20dd5808adf7R527). Which is a build-time decision of the builder/packager.
1. With my patch and additional condition is needed to get `containers/image` using SELinux: [`libselinux` is available in the system](https://github.com/containers/image/pull/546/files#diff-952ad9dc01673ce7143c20dd5808adf7R524). Which is a run-time decision of the user.

> And “is libselinux.so.1 expected to be available on the target system” is basically a build tag. The downside is that every library linking to c/image (that cares about non-SELinux systems) will need to have a Makefile possibly setting the build tag.

This sentence is showing exactly what I mean, with the current code it's the builder/packager decision to use SELinux or not, the user's opinion doesn't play a role. But OStree works like a charm without SELinux, so why making it a MUST on build-time? Why taking the decision of using it or not from the user?

Another possibility is providing different packages with and without SELinux support, but I find it's an ugly and confusing solution. If this repository enforces the decision on build-time, then those distributions and integrators not providing SELinux per default and not wanting to enforce their users to use it will just simply patch it out.

Silvanoc: @mtrmac the **main goal** of this PR is enabling the usage of **OStree without SELinux** without having to patch `ostree/ostree_dest.go` on build-time.

The goal is not really letting the user take the decision. In the end the decision if SELinux is in place or not is mostly being taken by the distributions and users just accept them, although they might change it (e.g. you use Red Hat, you have it, you use Debian, you don't have it). Letting the user decide is more a positive side-effect of the proposed design, than a driver.

In that sense, another possibility to have OStree without SELinux is having another file with the build flag
`// +build !containers_image_ostree_stub,!selinux`,
in that case some refactoring in the code of `ostree/ostree_dest.go` would be needed to avoid code duplication. It's probably the easiest way to approach it without letting the user decide him-/herself.

rhatdan:
@Silvanoc You could import github.com/opencontainers/selinux 
And check if SELinux is enabled on the host, and report an error if it is enabled and the library does not exist.
Otherwise we probably should not can about the library being missing.

rhatdan:
No I think we need to do the dlopen, but only if SELinux is enabled.  Then fail if SELinux is enabled and the library does not exist.  Which should force the user to install the library or disable SELinux.

rhatdan:
Correct the checking for selinux being enabled does nothing for the vendoring issue.

Silvanoc:
PR updated according @rhatdan to check availability of SELinux before trying to load `libselinux`.

Tests still failing due to dependency missing in `skopeo` build for testing, as documented in issue #552

rhatdan:
I like it.

Silvanoc:
The first commit of this PR is PR #554. That commit is needed here to get the tests passing.

Silvanoc:
The first commit of this PR is PR #554. That commit is needed here to get the tests passing.

ostree/ostree_dest.go
```diff
@@ -433,12 +517,35 @@ func (d *ostreeImageDestination) Commit(ctx context.Context) error {
...
+		defer libselinuxHnd.Close()
+
+		selinuxFunctionNames := []string{"selabel_open", "selabel_close",
+			"selabel_lookup_raw", "lsetfilecon_raw", "freecon"}
+
+		for _, function := range selinuxFunctionNames {
+			ptrFunction, err := libselinuxHnd.GetSymbolPointer(function)
+			selinuxFunctions[function] = ptrFunction
+			if err != nil {
+				return fmt.Errorf(`couldn't get symbol %q: %v`, function, err)
```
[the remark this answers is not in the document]
Silvanoc: I'm not an experienced Go programmer, I didn't know that "interpreted string literals" (what @vrothberg  calls "normal strings", right?) are preferred over "raw string literals".

I don't really have a preference for "raw string literals", I will replace it.

ostree/ostree_dest.go
```diff
@@ -32,7 +33,7 @@ import (
 	"github.com/vbatts/tar-split/tar/storage"
 )
 
-// #cgo pkg-config: glib-2.0 gobject-2.0 ostree-1 libselinux
```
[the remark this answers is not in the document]
Silvanoc: The inclusion of the headers enables loading `libselinux`, it only makes the constants and functions known. In fact **not including the header would lead to build errors**.

The difference between using `pkg-config` and `dlopen` to load `libselinux` is when you are getting a **runtime error**.

## Runtime behavior

### Current code
- If building without the flag `selinux`, github.com/opencontainers/selinux gets built to [return `false` to the selinux check](https://github.com/opencontainers/selinux/blob/master/go-selinux/selinux_stub.go#L34).
- If building with the flag `selinux`, github.com/opencontainers/selinux gets built so that [the function `GetEnabled` returns `true` only if SELinux is effectively enabled](https://github.com/opencontainers/selinux/blob/51c6c0a5dbc675792e953298cb9871819d6f9bb8/go-selinux/selinux_linux.go#L204).

In any case since `pkg-config` automatically tries to load `libselinux`, **you get an error if it's missing**. Therefore with the current code **`libselinux` has to be present if building with OStree support**, no matter the status of the buildflag `selinux`.

- If building without the flag `selinux`, `libselinux` has to be present, but won't be used!
- If building with the flag `selinux`, `libselinux` has to be present, and will be effectibly used.

The value of `selinuxHnd` is guarding any code involving SELinux from getting executed, if it's not enabled. And it happens independently from the `selinux` build flag.

### Patch 

The usage of [`dlopen`](https://github.com/containers/image/pull/546/files#diff-952ad9dc01673ce7143c20dd5808adf7R525) tries to load `libselinux`, but **doesn't return error if `libselinux` is missing**.

It does return **error if `libselinux` is expected to get loaded and cannot be loaded**. But what does it "expected to get loaded" means?

If it was built with the flag `selinux` and SELinux is enabled ([GetEnabled](https://github.com/opencontainers/selinux/blob/51c6c0a5dbc675792e953298cb9871819d6f9bb8/go-selinux/selinux_linux.go#L204) returns `true`).

In any case the patch doesn't change the guarding behavior of the value of `selinuxHnd` to avoid executing SELinux code, if not needed.

*Sorry for the long explanation, but I think that it will be helpful to understand the expected behavior*.

ostree/ostree_dest.go
```diff
@@ -41,6 +42,85 @@ import (
...
+//		void *function,
+//		char *con)
+// {
+//		void (*ptr_freecon)(
+//				char *);
+//		ptr_freecon = (void (*)(
+//				char *))function;
+//		return ptr_freecon(con);
+// }
+//
```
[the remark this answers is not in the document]
Silvanoc: Fixing it.

ostree/ostree_dest.go
mtrmac: I’m sorry about returning to this so late; I was away for most of December.

> But currently the decision "is SELinux enabled" is being taken at build-time: if buildtag `selinux` is not enabled, then `vendor/github.com/opencontainers/selinux/go-selinux/selinux_stub.go` gets used and `selinux.GetEnabled() ` returns false. So it doesn't have anything to do with the user intent, it's the builder/packager intent.

Sure, it is a builder/packager decision whether to have `libselinux.so` on the target system. The hypothetical user is not expected to rebuild any binaries; the user has `/etc/selinux/config` to express the desired configuration.

> 1. With the current code, SELinux usage is expected if [it was built with support for it](https://github.com/containers/image/pull/546/files#diff-952ad9dc01673ce7143c20dd5808adf7R527). Which is a build-time decision of the builder/packager.
> 2. With my patch and additional condition is needed to get `containers/image` using SELinux: [`libselinux` is available in the system](https://github.com/containers/image/pull/546/files#diff-952ad9dc01673ce7143c20dd5808adf7R524). Which is a run-time decision of the user.

Users are, in general, _not_ supposed to be randomly removing libraries from the system! “Configuration by removing a library” is not how most systems work.

Silently disabling SELinux operation just because a library is missing, or unreadable (perhaps due to a FS error) introduces brittleness into a critical security system, and it seems like something that should be avoided if possible:

Even if you are building an embedded system or a special-purpose system, you just can’t general be pulling binaries from a `libselinux`-dependent OS and expect them to work in a different environment.  From a quick check, all of the following (just a small subset!) directly links to `libselinux` (on a Fedora system): OpenSSH, PAM, shadow-utils, DBus, `sed`, `tar`, OpenVPN, `device-mapper`, `sudo`, `libmount`, `httpd`, `iproute` — as well as, to the point here, `GLib2` and `libostree`.

So, you already need a build system, or a configurable OS with build options, that allow SELinux-less builds of any of the above that your embedded system uses — and `c/image` and/or its users can likewise be a part of that build system with an option that has a SELinux-less build.

Overall, it seems quite possible to make the dependency on `libselinux` a build-time configuration option; and when this is both desirable and possible, why not do it that way?

 
> > And “is libselinux.so.1 expected to be available on the target system” is basically a build tag. The downside is that every library linking to c/image (that cares about non-SELinux systems) will need to have a Makefile possibly setting the build tag.
> 
> This sentence is showing exactly what I mean, with the current code it's the builder/packager decision to use SELinux or not, the user's opinion doesn't play a role. But OStree works like a charm without SELinux, so why making it a MUST on build-time? Why taking the decision of using it or not from the user?

OSTree _does not_ “work like a charm without SELinux” in the way this PR proposes; either you have a `--with-selinux` build or a `--without-selinux` build.  You can’t just take a `--with-selinux` build of OSTree, put it on a system where `libselinux`; it won’t work.

ostree/ostree_dest.go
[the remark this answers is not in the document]
mtrmac: This is another reason why actually using the `selinux` build tag instead of run-time `dlopen` would be better:

If the SELinux consumers were gated behind a build tag, a non-`selinux` build could succeed without `libselinux` headers and development libraries being present on the build system at all.  Considering how many C projects use something like autoconf and change their build options depending on whether the development files for `libselinux` are present, not including them in the build environment seem quite desirable.

And if you are considering a special-case embedded system, it would certainly be nice not to have to build `libselinux` for the target platform at all, if you don’t plan to use it!

ostree/ostree_dest.go
```diff
@@ -74,6 +158,9 @@ type ostreeImageDestination struct {
 	repo          *C.struct_OstreeRepo
 }
 
+// map to hold references to the required libselinux functions
+selinuxFunctions := make(map[string]unsafe.Pointer)
```
mtrmac: (FWIW this could use an integer index (e.g. a typed enum) instead of strings, for better performance and to make sure no typos happen.)

(But still I’d strongly prefer a build tag.)

ostree/ostree_dest.go
mtrmac: > Silently disabling SELinux operation just because a library is missing, or unreadable (perhaps due to a FS error) introduces brittleness into a critical security system, and it seems like something that should be avoided if possible:

… To be fair, this does not quite apply here because, surprisingly enough, `opencontainers/selinux/go-selinux` does not actually link to `libselinux`, so the current implementation (where `go-selinux` first decides whether SELinux is enabled, and then this code can hard-fail if `libselinux` is needed and missing) is not brittle that way.

Still, the `dlopen`s are overall more work than a build tag, and the wider ecosystem needs an “build with `libselinux` switch _anyway_”, so using `dlopen` does not seem worthwhile.

ostree/ostree_dest.go
```diff
@@ -433,12 +517,35 @@ func (d *ostreeImageDestination) Commit(ctx context.Context) error {
...
+		// selinuxHnd = nil (important in fixFiles)
+		// so the absence of SElinux will be simply ignored
+		libselinuxHnd, err := dlopen.GetHandle([]string{"libselinux.so.1"})
+		if err != nil {
+			return errors.Wrapf(err, "SELinux enabled, but libselinux not available")
+		}
+
+		defer libselinuxHnd.Close()
+
+		selinuxFunctionNames := []string{"selabel_open", "selabel_close",
```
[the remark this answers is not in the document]
vrothberg: Ouch,  I really missed that. Thanks!

rhatdan:
@cgwalters How much work would it be to add a build tag to ostree to make libselinux optional in ostree?

cgwalters:
> @cgwalters How much work would it be to add a build tag to ostree to make libselinux optional in ostree?

```
AC_ARG_WITH(selinux,
	    AS_HELP_STRING([--without-selinux], [Do not use SELinux]),
	    :, with_selinux=maybe)
```

Right?

Silvanoc:
Before further discussing about this PR, let me try to summarize the different approaches that I can think of with the different combinations of build and runtime setups.

If we can agree upon one of them, then we will be able to better focus on the implementation and less on the approach.

In general I prefer pr546 (the one that I'm contributing), because of it's flexibility (that's why I'm contributing it). But the buildtags approach would suffice for my use-case.

The build setup can be changed by passing or not passing the build tags `containers_image_ostree_stub` [1] or `selinux`.

The build option `containers_image_ostree_stub` disables OStree no matter if the build option `selinux` has been given or not (SELinux is only needed if using OStree, as a simple [search for SELinux](https://github.com/containers/image/search?q=selinux&unscoped_q=selinux) shows).

### Legend

💥 = runtime error

~~OStree~~/~~SELinux~~ = OStree/SELinux feature doesn't work

**OStree**/**SELinux** = OStree/SELinux feature works

## Current code (AKA current)

| ⬇️ Build setup \\ Libraries availability ➡️ | `-libostree -libselinux` | `+libostree -libselinux` | `-libostree +libselinux` | `+libostree +libselinux` |
| :------ | :------: | :------: | :------: | :------: |
| `!selinux` | 💥[3] | 💥[3] | 💥[3] | **OStree** ~~SELinux~~[2] |
| `selinux` | 💥[3] | 💥[3] | 💥[3] | **OStree** **SELinux** |

The table shows that `libselinux` has to be present to have OStree support, no matter if SELinux is enabled or not.

Buildtime:

1. **with OStree** (`!containers_image_ostree_stub`) enforcing presence of *libselinux*
1. **without OStree**

Runtime:

* **it fails**, if built with OStree and either libostree or libselinux is missing
* **it runs**, if built without OStree or with OStree and libselinux is present

## With patch from PR #546 (AKA pr546)

| ⬇️ Build setup \\ Libraries availability ➡️ | `-libostree -libselinux` | `+libostree -libselinux` | `-libostree +libselinux` | `+libostree +libselinux` |
| :------ | :------: | :------: | :------: | :------: |
| `!selinux` | 💥[3] | **OStree** ~~SELinux~~[2] | 💥[3] | **OStree** ~~SELinux~~[2] |
| `selinux` | 💥[3] | **OStree** ~~SELinux~~[5] | 💥[3] | **OStree** **SELinux** |

The table shows that building with OStree and SELinux support, the presence of `libselinux` (e.g. a *dependency package* on Arch or a *recommended package* on Debian) enables the user to easily switch on/off with a configuration flag `/etc/selinux/config`.

Buildtime:

1. **with OStree** (`!containers_image_ostree_stub`)
1. **without OStree**

Runtime:

* **it fails**, if built with OStree and libostree is missing
* **with OStree and SELinux**, if built with OStree, SELinux enabled and libselinux present
* **with OStree and without SELInux**, if built with OStree and libselinux missing or SELinux disabled

## Enabling OStree without SELinux with build tags (AKA buildtags)

No current proposal, but it would require something like a `ostree_dest_noselinux.go` file with following build tags `// +build !containers_image_ostree_stub,!selinux`.

| ⬇️ Build setup \\ Libraries availability ➡️ | `-libostree -libselinux` | `+libostree -libselinux` | `-libostree +libselinux` | `+libostree +libselinux` |
| :------ | :------: | :------: | :------: | :------: |
| `!selinux` | 💥[3] | **OStree** ~~SELinux~~[2] | 💥[3] | **OStree** ~~SELinux~~[2] |
| `selinux` | 💥[3] | 💥 | 💥[3] | **OStree** **SELinux** |

This approach enables OStree without requiring the presence of `libselinux`. But it misses the possibility of enabling SELinux by installing `libselinux` and configuring appropriately, what is possible with pr546.

Buildtime:

1. **with OStree and SELinux** (`!containers_image_ostree_stub,selinux`)
1. **with OStree and without SELinux** (`!containers_image_ostree_stub,selinux`)
1. **without OStree**

Runtime:

* **it fails**, if built with OStree and libostree is missing
* **with OStree and SELinux**, if built with OStree, SELinux enabled and libselinux present
* **with OStree and without SELinux**, if built with OStree, SELinux disabled or libselinux missing

---
[1] https://github.com/containers/image/blob/753958120182c9ffdaa2ec58b81dae4e05f518aa/README.md#L65

[2] https://github.com/opencontainers/selinux/blob/98291d4e0d5f03baed73256f9142282c19a163bf/go-selinux/selinux_stub.go#L1

[3] https://github.com/containers/image/blob/3dda079ae672eb11d7b62e6f85a3283c5b560153/ostree/ostree_dest.go#L35

[4] https://github.com/containers/image/blob/3dda079ae672eb11d7b62e6f85a3283c5b560153/ostree/ostree_dest.go#L197

[5] https://github.com/containers/image/pull/546/files#diff-952ad9dc01673ce7143c20dd5808adf7R529

mtrmac:
> But it misses the possibility of enabling SELinux by installing `libselinux` and configuring appropriately, what is possible with pr546.

It’s really not, because `libostree`, as well as _much else on the system_, is built either `--with-selinux` (in which case “not installing libselinux” is not an option) or `--without-selinux` (in which case enabling SELinux for c/image only does not really make sense.

Silvanoc:
> > But it misses the possibility of enabling SELinux by installing `libselinux` and configuring appropriately, what is possible with pr546.
> 
> It’s really not, because `libostree`, as well as _much else on the system_, is built either `--with-selinux` (in which case “not installing libselinux” is not an option) or `--without-selinux` (in which case enabling SELinux for c/image only does not really make sense).

But building libostree with SELinux doesn't require `libselinux1-dev` to be present, if not it would fail in any [Debian system withouth SELinux](https://salsa.debian.org/debian/ostree/blob/debian/master/debian/rules) (very frequent case).

I mean, what is the motivation for building without SELinux support?

On the other side, [Arch Linux builds without SELinux support](https://git.archlinux.org/svntogit/packages.git/tree/trunk/PKGBUILD?h=packages/ostree).

pr546 would enable the user activation/deactivation on systems built with SELinux support but without requirements on installed packages/libraries (e.g. for Debian).

Anyway, since my main motivation is removing the `libselinux` requirement for using OStree and there are other possibilities to accomplish it, I'll create another PR for the buildtags approach.

mtrmac:
> But building libostree with SELinux doesn't require `libselinux1-dev` to be present, if not it would fail in any [Debian system withouth SELinux](https://salsa.debian.org/debian/ostree/blob/debian/master/debian/rules) (very frequent case).

I don’t really know Debian packaging, but looking at https://github.com/ostreedev/ostree/blob/5d031ae78b2e6b6da57fd6dca2171e2471a04586/src/libostree/ostree-sepolicy.c#L24-L27 , how can that be true?

And, per 
https://salsa.debian.org/debian/ostree/blob/debian/master/debian/control#L37 and https://packages.debian.org/en/stretch/libostree-1-1 , it seems not to be true; the Debian build of `libostree` strictly requires `libselinux` to be present, both at build and run time.

Silvanoc:
> > But building libostree with SELinux doesn't require `libselinux1-dev` to be present, if not it would fail in any [Debian system withouth SELinux](https://salsa.debian.org/debian/ostree/blob/debian/master/debian/rules) (very frequent case).
> 
> I don’t really know Debian packaging, but looking at https://github.com/ostreedev/ostree/blob/5d031ae78b2e6b6da57fd6dca2171e2471a04586/src/libostree/ostree-sepolicy.c#L24-L27 , how can that be true?
> 
> And, per
> https://salsa.debian.org/debian/ostree/blob/debian/master/debian/control#L37 and https://packages.debian.org/en/stretch/libostree-1-1 , it seems not to be true; the Debian build of `libostree` strictly requires `libselinux` to be present, both at build and run time.

You are fully right, I oversaw that. You brought the killer argument against this approach :wink: 

It probably can be addressed the same way in OStree, but I don't think it's worth...

As mentioned before, I'll provide a PR for the buildtags approach. Therefore killing this approach and closing this PR.

Silvanoc:
@vrothberg @mtrmac thank you for your detailed reviews and openness for discussion!

I can ensure you that you didn't waste your time, since you educated me and got me hopefully closer to a PR that fixes the issue.

Issue #544, opened 2018-12-14
make localhost insecure by default


vrothberg:
Docker marks localhost as insecure by default. I assume that the reasoning behind that is to enable an easy and straight-forward way to develop and test against a local registry running on the same host. Note that there doesn't seem to be a restriction regarding ports.

I think containers/image should follow the example as I think that users expect this behaviour.

vrothberg:
@runcom @rhatdan @giuseppe @TomSweeneyRedHat WDYT?

mtrmac:
What is the _exact_ set of hosts this would apply to? Precisely `localhost`? Precisely `127.0.0.1`? Both? More?

At a first glance this makes sense because a CA-signed certificate (even if signed by a private CA) for `localhost` is a dubious idea.

rhatdan:
Makes sense to me. localhost/image has no meaning to Docker, they default to docker.io/image
I think 127.0.0.1 would.
But this works for me.

vrothberg:
What I would like to support is `localhost:$PORT` and `127.0.0.1`, which afaik aligns with what Docker does.

vrothberg:
Closing. It can easily be configured in the registries.conf.

PR #547, opened 2018-12-18
docker: mark localhost as insecure

vrothberg:
Mark localhost as insecure to enable communicating with a registry
running on localhost without requiring TLS or explicitly marking it as
insecure in the registries.conf.  This behaviour aligns with Docker.

Notice, that localhost can still be marked to not skip TLS verification
either by setting it via the SystemContext or by configuring it
explicitly in the registries.conf.  A registry is considered to be
running on localhost if it either starts with "localhost:" or with
"127.0.0.1".

Fixes: #544

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -241,7 +241,9 @@ func newDockerClient(sys *types.SystemContext, registry, reference string) (*doc
 
 	// Check if TLS verification shall be skipped (default=false) which can
 	// either be specified in the sysregistriesv2 configuration or via the
-	// SystemContext, whereas the SystemContext is prioritized.
+	// SystemContext, whereas the SystemContext is prioritized.  If neither
+	// sets it, check if the registry is running on the localhost and mark
+	// it as insecure.
 	skipVerify := false
 	if sys != nil && sys.DockerInsecureSkipTLSVerify != types.OptionalBoolUndefined {
 		// Only use the SystemContext if the actual value is defined.
@@ -253,6 +255,13 @@ func newDockerClient(sys *types.SystemContext, registry, reference string) (*doc
 		}
 		if reg != nil {
 			skipVerify = reg.Insecure
+		} else {
+			// unless specified in the SystemContext or in the
+			// config, set skipVerify to true if the registry
+			// is running on the localhost
+			if strings.HasPrefix(registry, "localhost:") || strings.HasPrefix(registry, "127.0.0.1") {
+				skipVerify = true
+			}
 		}
 	}
 	tr.TLSClientConfig.InsecureSkipVerify = skipVerify
--- a/docker/docker_client_test.go
+++ b/docker/docker_client_test.go
@@ -20,6 +20,23 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+func TestSkipTLSVerifyOnLocalhost(t *testing.T) {
+	client, err := newDockerClient(nil, "localhost:5000", "localhost:5000")
+	assert.Nil(t, err)
+	assert.NotNil(t, client)
+	assert.True(t, client.insecureSkipTLSVerify)
+
+	client, err = newDockerClient(nil, "localhost", "localhost")
+	assert.Nil(t, err)
+	assert.NotNil(t, client)
+	assert.False(t, client.insecureSkipTLSVerify)
+
+	client, err = newDockerClient(nil, "127.0.0.1", "127.0.0.1")
+	assert.Nil(t, err)
+	assert.NotNil(t, client)
+	assert.True(t, client.insecureSkipTLSVerify)
+}
+
 func TestDockerCertDir(t *testing.T) {
 	const nondefaultFullPath = "/this/is/not/the/default/full/path"
 	const nondefaultPerHostDir = "/this/is/not/the/default/certs.d"


vrothberg:
@giuseppe @runcom @rhatdan PTAL

vrothberg:
Ah, we need to merge the pgzip PR in Skopeo to make c/image CI run again.

runcom:
This should work at CIDR level at least https://github.com/kubernetes-sigs/cri-o/blob/master/pkg/storage/image.go#L610 (take a look here as well https://github.com/kubernetes-sigs/cri-o/blob/master/pkg/storage/image.go#L488-L522)

docker/docker_client.go
```diff
@@ -253,6 +255,13 @@ func newDockerClient(sys *types.SystemContext, registry, reference string) (*doc
 		}
 		if reg != nil {
 			skipVerify = reg.Insecure
+		} else {
+			// unless specified in the SystemContext or in the
+			// config, set skipVerify to true if the registry
+			// is running on the localhost
+			if strings.HasPrefix(registry, "localhost:") || strings.HasPrefix(registry, "127.0.0.1") {
```
mtrmac: This does not match `localhost`, and does match `127.0.0.19`.

mtrmac:
> This should work at CIDR level at least https://github.com/kubernetes-sigs/cri-o/blob/master/pkg/storage/image.go#L610 (take a look here as well https://github.com/kubernetes-sigs/cri-o/blob/master/pkg/storage/image.go#L488-L522)

Until someone figures out how to do it securely, I strongly object to making the whole 127.0.0.0/8 network insecure. “Any user on the local network who can inject a DNS reply can defeat TLS” is not a reasonable default behavior.

runcom:
I know, I understand that, but we're trying to match Docker so having at least a knob would be desiderabile

docker/docker_client.go
vrothberg: I am not sure if `localhost` would actually work since `localhost/image:latest` would be seen as an unqualified image and be transformed to `docker.io/localhost/image:latest` - Docker behaves like this as well.

vrothberg:
> This location works, but I wonder whether instead we should inject these values in sysregistriesv2 (and V1) — If we are making the config parsers available at all, users of them should probably behave consistently.

@mtrmac, I have no objections against doing that in the `sysregistriesv2`. It may even be easier to handle from a maintenance point of view.

docker/docker_client.go
mtrmac: 1. No, the `docker.io[/library]` normalization special-cases the `localhost` host name (other strings must contain a dot to be treated as host names, `localhost` is accepted as a special case – search in `c/image/docker/reference`)
2. This is checking `registry`, not `reference`; the caller must already be setting `registry` to the right host:port (not host:port/ns/repo) in order for connections to work.

To clarify, the review comment was primarily pointing out the inconsistent treatment of `:` (i.e. both ports, and non-port suffixes) in the two cases of the condition.

mtrmac:
> > This location works, but I wonder whether instead we should inject these values in sysregistriesv2 (and V1) — If we are making the config parsers available at all, users of them should probably behave consistently.
>
> @mtrmac, I have no objections against doing that in the sysregistriesv2. It may even be easier to handle from a maintenance point of view.

Overall, no users of c/image should _interpret_ the `Registry.Insecure` field now that c/image/docker does that automatically (and we have already dropped that logic from buildah and podman), and I can’t find any strong argument for why the code should live in one or the other package.

Looking at the _remaining_ uses of sysregistries{v2,} that care about the “insecure” flag, there is basically `podman info`, listing the configured secure/insecure registries. That necessarily uses `sysregistriesv2.GetRegistries` to list _all_ configured registries, and we can’t return “localhost:$anyport” in there. It seems slightly more consistent to me, then, to have `FindRegistry` return nothing by default for `localhost:$anyport` as well, and to do the mapping in c/image/docker, as this PR does.

rhatdan:
@vrothberg @mtrmac Are we moving forward with this?

vrothberg:
> @vrothberg @mtrmac Are we moving forward with this?

Not yet but a workaround is to add, for instance, `localhost:5000` to the `registries.conf`. We have a few issues in this PR:
- When/where/how can we support CIDRs?
- Shall we special case `c/image/docker`, add default registries in `c/image/pkg/sysregistriesv2` or shall recommend shipping a preset `registries.conf`?

IMHO, I prefer shipping a preset `registries.conf` and figure out how we can support CIDRs. It's straight-forward, easy to understand as visible in the config and without any internal magic that needs to be documented etc.

mtrmac:
> * When/where/how can we support CIDRs?

We can support CIDRs matching registries user-specified by an IP address (`docker://127.0.0.1`).  CIDRs matching DNS names is insecure, and AFAICS not reasonably possible to do securely with the Go HTTP/TLS/net stack.

> * Shall we special case `c/image/docker`, add default registries in `c/image/pkg/sysregistriesv2` or shall we recommend shipping a preset `registries.conf`?
> 
> IMHO, I prefer shipping a preset `registries.conf` and figure out how we can support CIDRs. It's straight-forward, easy to understand as visible in the config and without any internal magic that needs to be documented etc.

We can’t currently configure “localhost, either the implied default port, or any other explicit port” in `registries.conf`; so either that happens implicitly in code, or we would have to extend `registries.conf` (and the earlier discussion about wildcards, which I can’t find now, has shown that it’s trivial to add the wildcard matching, but the semantics WRT mirrors and precedence are rather non-obvious).

docker/docker_client.go
vrothberg: Coming back to this issue. @mtrmac, shall we support the entire range from 127.0.0.1 through 127.255.255.255?

> This does not match localhost, and does match 127.0.0.19.

^ made me wonder.

mtrmac: I’d rather not; all of these “don’t be secure” defaults are risky.

We get (guessing with no evidence) 98% of the convenience benefit matching just `localhost`, 1% matching `127.0.0.1`, with hopefully manageable risk.

Anyone in a special enough situation that there are different servers running on `127.0.0.1`, `127.0.0.19` and `127.255.255.254` has been doing enough manual and unusual configuration that having to explicitly configure `127.0.0.19` as insecure is negligible extra burden, and that we don’t quite know what/why is going on and whether the risk is still acceptable.

rhatdan:
@vrothberg @mtrmac What is the latest on this PR, is this still something you guys seeing us need, or should we close it?

vrothberg:
Let's close it. It can easily be configured in the registries.conf and avoids some more internal "magic". Thanks for catching up, Dan!

PR #550, opened 2018-12-19
Check for existing blobs on other registries too. Ignore domain mismatches.

42wim:
See https://github.com/containers/buildah/issues/1048

We're using artifactory as docker registry and we have multiple registries on it (with multiple hostnames, registry1.domain.tld, registry2.domain.tld), the current code will only check blobs from the same registry although they're also available on the other.

with docker doing a
```
docker pull registry1.domain.tld/image:tag
docker tag registry1.domain.tld/image:tag registry2.domain.tld/image2:tag 
docker push registry2.domain.tld/image2:tag
```
will not push the image to registry2.domain.tld because it's already there (i'm not sure if this is default registry behaviour, but it does so with artifactory and docker).

Doing the same with buildah will cause a repush of the image as it's not checking for the existing blob.

Removing the registry sanity checks and setting always the same scope.Opaque in the boltdb cache fixes this for me and now buildah behaves the same as docker.

Is there a reason why this isn't the default action, are there security implications ? (as it are sha256 hashes creating collisions there are not feasible)

This PR is probably not the best way to fix this, but it does the trick.


pkg/blobinfocache/boltdb.go
```diff
@@ -220,6 +220,7 @@ func (bdc *boltDBCache) RecordDigestUncompressedPair(anyDigest digest.Digest, un
 // RecordKnownLocation records that a blob with the specified digest exists within the specified (transport, scope) scope,
 // and can be reused given the opaque location data.
 func (bdc *boltDBCache) RecordKnownLocation(transport types.ImageTransport, scope types.BICTransportScope, blobDigest digest.Digest, location types.BICLocationReference) {
+	scope.Opaque = "all"
```
mtrmac: (If this should happen, it should happen inside the `docker` transport directly, by not setting a per-domain value, instead of stomping over the set value here in the transport-agnostic code.)

rhatdan:
@42wim Any movement on this?

42wim:
Oops sorry, holidays etc,
Did you look further in the docker behaviour @mtrmac ?
Or should I just try to implement your existence check?
(I can take a deeper look in this on thursday/friday)

mtrmac:
I’m not working on this right now; if you want to, that would be great.

42wim:
@mtrmac Tried to follow your suggestions:
* mounting isn't done anymore on those blobs.
* bicTransportScope is being set `everywhere` as value, maybe more clear then an empty string? 
* behaviour is now set using the `checkBlobEverywhere´ boolean. 

I'm wondering if there's a reason to just not have this `true` by default.

42wim:
Rebased on master, any feedback?
Got this also working with buildah 1.7.1

rhatdan:
@42wim @mrunalp @vrothberg Any thoughts on this PR?  Last activity was over a month ago.

42wim:
Still got a need for it, I can rebase

42wim:
Any updates, just rebased it. We're using this in our ci/cd since buildah 1.6
I can also do a PR for buildah which add support for this.

rhatdan:
@42wim You need to sign your PR and then hopefully the tests will pass.

git commit -a -s --amend
git push --force

42wim:
updated again, tests pass

rhatdan:
@mtrmac @vrothberg @nalind PTAL

docker/docker_image_dest.go
```diff
@@ -297,10 +299,15 @@ func (d *dockerImageDestination) TryReusingBlob(ctx context.Context, info types.
 
 		// Sanity checks:
 		if reference.Domain(candidateRepo) != reference.Domain(d.ref.ref) {
-			logrus.Debugf("... Internal error: domain %s does not match destination %s", reference.Domain(candidateRepo), reference.Domain(d.ref.ref))
-			continue
+			if d.checkBlobEverywhere {
+				logrus.Debugf("... Ignoring domain %s does not match destination %s. checkBlobEverywhere enabled", reference.Domain(candidateRepo), reference.Domain(d.ref.ref))
```
vrothberg: I prefer to abstract from the internals a bit to make this log easier to understand and suggest:

"... Cross repository blob check: ignoring domain mismatch between %s and %s"

types/types.go
```diff
@@ -511,6 +511,9 @@ type SystemContext struct {
 	// === dir.Transport overrides ===
 	// DirForceCompress compresses the image layers if set to true
 	DirForceCompress bool
+
+	// Check blobs across registries
+	CheckBlobEverywhere bool
```
vrothberg: I prefer calling it `CrossRepositoryBlobCheck`

42wim: Yes, that's a better name

42wim:
Anyone can take a look?
Rebased again, and I'm using it with buildah 1.16.4.
ping @rhatdan @mtrmac

mtrmac (commented):
AFAICS the way this could be implemented:
- Keep the current `bicTransportScope`, and write all of the data as before.
- In addition to `CandidateLocations2` (as it is now available in `internal/blobinfocache`), add `internal/blobinfocache.BlobInfoCache2.CandidateDigests`, which has the basic structure of computing the digests `CandidateLocations2` implementations pass to `appendReplacementCandidates`, but just records the digests without looking up the (scope, digest) → location mappings.
- Then, in `dockerImageDestination.TryReusingBlob`, after the current use of `CandidateLocations2`, call `CandidateDigests` and just check using `blobExists` for pre-existing blobs _in that destination repo_.
  - I don’t know whether we need an opt-in at all for this. Maybe just try up to `prioritize.replacementAttempts` entries _total_, starting with the `candidateLocations2` ones.

rhatdan:
@42wim Is this still something you want to work on?

42wim:
@rhatdan sorry, seems to be gone of my radar, we still need this feature. Looking at the thumbs above at least two other people also need it :-)

If it's going to be implemented by someone else, by all means, otherwise I'll try to pick it up in the next few weeks if that's okay.

mtrmac:
#1645 is probably the right approach for implementing this feature.

mtrmac:
#1645 was merged.

PR #551, opened 2018-12-20
copy.Copy: progressbars: set correct output

vrothberg:
Set the output of the progress bars and the progress pool, respectively,
to the specified io.Writer passed through the copy options.  Otherwise,
the output will be printed to Stdout breaking users which are parsing
the output and expect the progress bars to be printed on Stderr.

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -395,12 +395,13 @@ func shortDigest(d digest.Digest) string {
 }
 
 // createProgressBar creates a pb.ProgressBar.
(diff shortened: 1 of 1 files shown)


vrothberg:
Please, do not merge this until we are sure it's fixing the flakes over at Buildah (https://github.com/containers/buildah/pull/1244/).

rhatdan:
LGTM

vrothberg:
It is fixing the flakes :) The current failure relates to buildah's blobcache behaving slightly different with pzgip.

runcom:
LGTM

PR #553, opened 2018-12-21
dockerImageDestination: HasThreadSafePutBlob -> false

vrothberg:
Mark dockerImageDestination to not be thread safe as there is a race
condition in dockerClient.extraScope causing copy.Copy() to error out
with `[...]dockerClient.extraScope was set before TryReusingBlob`,
which must be addressed in future changes.


runcom:
LGTM

rhatdan:
LGTM

PR #557, opened 2019-01-14
Parallel push

vrothberg:
* make dockerImageDestination's `PutBlob` "thread safe" by moving the `extraScope` member into parameters
* tested locally

### DIFF
--- a/docker/docker_image_dest.go
+++ b/docker/docker_image_dest.go
@@ -140,7 +140,7 @@ func (d *dockerImageDestination) PutBlob(ctx context.Context, stream io.Reader, 
 	// FIXME? Chunked upload, progress reporting, etc.
 	uploadPath := fmt.Sprintf(blobUploadPath, reference.Path(d.ref.ref))
 	logrus.Debugf("Uploading %s", uploadPath)
-	res, err := d.c.makeRequest(ctx, "POST", uploadPath, nil, nil, v2Auth)
+	res, err := d.c.makeRequest(ctx, "POST", uploadPath, nil, nil, v2Auth, nil)
 	if err != nil {
 		return types.BlobInfo{}, err
 	}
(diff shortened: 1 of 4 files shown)


vrothberg:
@mtrmac @mtrmac PTAL

Note that we're still serialized by the locks in c/storage but I think I can come up with something.

rhatdan:
LGTM
@runcom PTAL

docker/docker_image_dest.go
```diff
@@ -314,7 +305,12 @@ func (d *dockerImageDestination) TryReusingBlob(ctx context.Context, info types.
 		}
 
 		// Whatever happens here, don't abort the entire operation.  It's likely we just don't have permissions, and if it is a critical network error, we will find out soon enough anyway.
-		d.c.extraScope = &authScope{
+
+		// Checking candidateRepo, and mounting from it, requires an
+		// expanded token scope.  We still want to reuse the ping
+		// information and other aspects of the client, so rather than
+		// make a fresh copy, there is this a bit ugly extraScope hack.
```
mtrmac: The “extraScope hack” originally referred to the stateful `.extraScope` member and the way it was _not_ passed through `makeRequest`*; this PR passes the data in a more accurate straightforward way, so I think the reference to the “ugly hack” can be dropped.

Unless you want to leave it around to refer to the fair amount of difficult to read parameters to `makeRequest`— but that’s mostly the fault of Go not having named parameters :)

vrothberg: :rofl: dropping sounds fine. Thanks!

vrothberg: Done :-)

vrothberg:
Updated and re-pushed.

mtrmac:
LGTM pending tests. Thanks!

vrothberg:
Rebased.

vrothberg:
Green, @mtrmac @runcom @rhatdan PTAL.

PR #558, opened 2019-01-15
copy: print singles lines when reportWriter is not a tty

vrothberg:
If the reportWriter is not a tty then print a single line for copy
operations and discard the output of the progress poll and the
progress bars.  This avoid long and hard to parse output when not
using a tty, for instance, when piping the output to a file.

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"os"
 	"reflect"
 	"runtime"
 	"strings"
@@ -22,6 +23,7 @@ import (
 	"github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/crypto/ssh/terminal"
 	"golang.org/x/sync/semaphore"
 	pb "gopkg.in/cheggaaa/pb.v1"
 )
@@ -84,6 +86,7 @@ type copier struct {
 	dest             types.ImageDestination
 	rawSource        types.ImageSource
 	reportWriter     io.Writer
+	progressOutput   io.Writer
 	progressInterval time.Duration
 	progress         chan types.ProgressProperties
 	blobInfoCache    types.BlobInfoCache
@@ -152,11 +155,19 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
 		}
 	}()
 
+	// If reportWriter is not a TTY (e.g., when piping to a file), do not
+	// print the progress bars to avoid long and hard to parse output.
+	// createProgressBar() will print a single line instead.
+	progressOutput := reportWriter
+	if !isTTY(reportWriter) {
+		progressOutput = ioutil.Discard
+	}
 	copyInParallel := dest.HasThreadSafePutBlob() && rawSource.HasThreadSafeGetBlob()
 	c := &copier{
 		dest:             dest,
 		rawSource:        rawSource,
 		reportWriter:     reportWriter,
+		progressOutput:   progressOutput,
 		progressInterval: options.ProgressInterval,
 		progress:         options.Progress,
 		copyInParallel:   copyInParallel,
@@ -394,15 +405,28 @@ func shortDigest(d digest.Digest) string {
 	return d.Encoded()[:12]
 }
 
-// createProgressBar creates a pb.ProgressBar.
-func createProgressBar(srcInfo types.BlobInfo, kind string, writer io.Writer) *pb.ProgressBar {
+// createProgressBar creates a pb.ProgressBar.  Note that if the copier's
+// reportWriter is ioutil.Discard, the progress bar's output will be discarded
+// and a single line will be printed instead.
+func (c *copier) createProgressBar(srcInfo types.BlobInfo, kind string) *pb.ProgressBar {
 	bar := pb.New(int(srcInfo.Size)).SetUnits(pb.U_BYTES)
 	bar.SetMaxWidth(80)
 	bar.ShowTimeLeft = false
 	bar.ShowPercent = false
 	bar.Prefix(fmt.Sprintf("Copying %s %s:", kind, shortDigest(srcInfo.Digest)))
-	bar.Output = writer
+	bar.Output = c.progressOutput
+	if bar.Output == ioutil.Discard {
+		c.Printf("Copying %s %s\n", kind, srcInfo.Digest)
+	}
 	return bar
+}
+
+// isTTY returns true if the io.Writer is a file and a tty.
+func isTTY(w io.Writer) bool {
+	if f, ok := w.(*os.File); ok {
+		return terminal.IsTerminal(int(f.Fd()))
+	}
+	return false
 }
 
 // copyLayers copies layers from ic.src/ic.c.rawSource to dest, using and updating ic.manifestUpdates if necessary and ic.canModifyManifest.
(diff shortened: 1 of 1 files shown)


vrothberg:
@mtrmac @runcom @rhatdan PTAL. This fixes an issue reported by Ed on IRC that test scripts get confused as the output has changed with the parallel pulling.

vrothberg:
> It does not really make sense to me to just silently ignore a user-specified option value. If it is _that_ invalid, `copy.Image` should fail with an error, not ignore the invalid input.

This would certainly break users but I like your proposal below.

> I suspect that instead of this, we still want to keep code like
> 
> ```go
> 	ic.c.Printf("Writing manifest to image destination\n")
> ```
> 
> working, and only turn off the progress bars.

That sounds fair.

vrothberg:
I updated the commit according to @mtrmac 's comments and the Skopeo CI PR is green.
Ready for the next review round.

copy/copy.go
```diff
@@ -394,17 +396,36 @@ func shortDigest(d digest.Digest) string {
...
+		bar.Output = c.reportWriter
+	} else {
+		bar.Output = ioutil.Discard
+		c.Printf("Copying %s %s\n", kind, srcInfo.Digest)
+	}
 	return bar
 }
 
+// isFileButNoTTY returns true if the io.Writer is a file but not a tty.
+func isFileButNoTTY(w io.Writer) bool {
```
mtrmac: Yet the implementation actually returns `true` if `IsTerminal`??!

copy/copy.go
```diff
@@ -394,17 +396,36 @@ func shortDigest(d digest.Digest) string {
...
+	if reflect.TypeOf(w) == reflect.TypeOf((*os.File)(nil)) {
+		f := w.(*os.File)
```
mtrmac: Aren’t these two lines basically
```go
if f, ok := w.(*os.File); ok {
```
?

copy/copy.go
```diff
@@ -469,12 +490,19 @@ func (ic *imageCopier) copyLayers(ctx context.Context) error {
...
+		bar := ic.c.createProgressBar(srcInfo, "blob")
 		progressBars[i] = bar
 	}
 
 	progressPool := pb.NewPool(progressBars...)
 	progressPool.Output = ic.c.reportWriter
+
+	// reset the output if it's a file but not a tty to avoid long and
+	// hard to parse indeterministic output from the progress bars
+	if !isFileButNoTTY(progressPool.Output) {
```
mtrmac: (Non-blocking: It would be a tiny bit more readable for this to look the same as the condition in `createProgressBar` (i.e. probably for both to use the same `if`/`else`.)

copy/copy.go
```diff
@@ -394,17 +396,36 @@ func shortDigest(d digest.Digest) string {
...
+// discarded and a single line will be printed instead.
+func (c *copier) createProgressBar(srcInfo types.BlobInfo, kind string) *pb.ProgressBar {
 	bar := pb.New(int(srcInfo.Size)).SetUnits(pb.U_BYTES)
 	bar.SetMaxWidth(80)
 	bar.ShowTimeLeft = false
 	bar.ShowPercent = false
-	bar.Prefix(fmt.Sprintf("Copying %s %s:", kind, shortDigest(srcInfo.Digest)))
-	bar.Output = writer
+	prefix := fmt.Sprintf("Copying %s %s:", kind, shortDigest(srcInfo.Digest))
+	bar.Prefix(prefix)
```
mtrmac: (The new `prefix` variable seems to have no other users, were there supposed to be any?)

copy/copy.go
```diff
@@ -19,6 +20,7 @@ import (
 	"github.com/containers/image/transports"
 	"github.com/containers/image/types"
 	"github.com/klauspost/pgzip"
+	"github.com/mattn/go-isatty"
```
mtrmac: (Absolutely non-blocking: It’s not _great_ how this uses `syscall.Syscall6` when there is a ~type-safe `golang.org/x/sys/unix.IoctlGetTermios` (used by the progress bar code) — but, oh well, it’s still much better to depend on a reusable package than to open-code it.

[OTOH the progress bar’s `isatty` implementation is not really answering that question at all…])

copy/copy.go
vrothberg: Yes, that's much nicer. I keep forgetting that go supports type assertions.

copy/copy.go
vrothberg: I like your proposal. The print decision can be based on `== ioutil.Discard` or we default `.progressBarOutput` to nil and set it there to `iotul.Discard` (maybe faster).

copy/copy.go
vrothberg: Artefact of tiredness, sorry for that.

copy/copy.go
vrothberg: I can check if https://godoc.org/golang.org/x/crypto/ssh/terminal#IsTerminal works when casting the fd to int.

vrothberg:
And another round (https://github.com/containers/skopeo/pull/593 is green).

copy/copy.go
```diff
@@ -394,17 +405,32 @@ func shortDigest(d digest.Digest) string {
 	return d.Encoded()[:12]
 }
 
-// createProgressBar creates a pb.ProgressBar.
-func createProgressBar(srcInfo types.BlobInfo, kind string, writer io.Writer) *pb.ProgressBar {
+// createProgressBar creates a pb.ProgressBar.  Note that if the copier's
+// reportWriter is iotuil.Discard, the progress bar's output will be discarded
```
mtrmac: `s/iotuil/ioutil/`

copy/copy.go
```diff
@@ -394,17 +405,32 @@ func shortDigest(d digest.Digest) string {
...
+	}
 	return bar
 }
 
+// isTTY returns true if the io.Writer is a file and a tty.
+func isTTY(w io.Writer) bool {
+	if f, ok := w.(*os.File); ok {
+		if terminal.IsTerminal(int(f.Fd())) {
+			return true
+		}
```
mtrmac: (Non-blocking: This can be `return terminal.IsTerminal(…)`)

copy/copy.go
mtrmac: Absolutely non-blocking: I have actually looked into this a bit, there’s also `logrus.IsTerminal`; it also uses the `Syscall6` implementation, but we are already depending on that exact package so using it would not bring any net new code.

Not changing this is _perfectly fine_.

edsantiago:
FWIW I've built and tested podman with b29f90ac and 87a0a32f applied; all resolve the issue I was seeing. I'm sorry this turned out more complicated than expected; thank you for your efforts to address it.

vrothberg:
> FWIW I've built and tested podman with [b29f90a](https://github.com/containers/image/commit/b29f90ac418b0370fa303f90fd0a3e9f88958f0f) and [87a0a32](https://github.com/containers/image/commit/87a0a32f649889143880c9214404699af8ac2ada) applied; all resolve the issue I was seeing. I'm sorry this turned out more complicated than expected; thank you for your efforts to address it.

Thanks a lot for proposing that, @edsantiago! My brainfarts made it more complicated than it actually was :) I will integrate @mtrmac 's last comments and then we can merge it.

copy/copy.go
vrothberg: done

copy/copy.go
vrothberg: done

mtrmac:
LGTM. Thanks!

vrothberg:
@mtrmac / @runcom  can you merge this? I feel uncomfortable merging my own PR :angel: We also need to create a new release of c/image after the merge to get the fix into Podman v1.0.X.

mtrmac:
Merging, please update 
https://github.com/containers/skopeo/pull/593 .

PR #563, opened 2019-01-21
set correct media types during manifest updates and conversions

vrothberg:
When copying an image, record the compression in the BlobInfo and use
the information when updating the manifest's layer infos to set the
layers' media types correctly.

Fixes: github.com/containers/libpod/issues/2013
Fixes: github.com/containers/buildah/issues/1589


vrothberg:
Tested locally and maybe incomplete w.r.t. to real mime type of the destination layers.

vrothberg:
@mtrmac @runcom can you have a look at this? Skopeo CI is passing but I need your pairs of eyes :) Note that I only changed directory destination to return the mediaType. There might be other cases where it needs to be updated as well.

copy/copy.go
```diff
@@ -813,6 +814,17 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
 		return types.BlobInfo{}, errors.Wrap(err, "Error writing blob")
 	}
 
+	switch compressionOperation {
+	case types.Compress:
+		uploadedInfo.MediaType = manifest.DockerV2Schema2LayerMediaType
+	case types.Decompress:
+		uploadedInfo.MediaType = manifest.DockerV2SchemeLayerMediaTypeUncompressed
```
mtrmac: This needs to be specific to the schema (`manifest.Manifest`/`types.Image`) implementation: OCI uses differently-named types.

(And it probably only should change MIME types if the previous value is exactly what is expected, so that a config MIME type is not replaced by a blob MIME type.)

copy/copy.go
```diff
@@ -813,6 +814,17 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
...
+	case types.PreserveOriginal:
+		// preserve whatever the destination sets by default
+	default:
+		panic("unexpected compression operation")
```
mtrmac: `return fmt.Errorf("Internal error: invalid compression operation %v")` or something like that, please; if it is not possible to cleanly enough redesign the code to make such a path impossible to reach.

directory/directory_dest.go
```diff
@@ -172,7 +173,7 @@ func (d *dirImageDestination) PutBlob(ctx context.Context, stream io.Reader, inp
 		return types.BlobInfo{}, err
 	}
 	succeeded = true
-	return types.BlobInfo{Digest: computedDigest, Size: size}, nil
+	return types.BlobInfo{Digest: computedDigest, Size: size, MediaType: manifest.DockerV2SchemeLayerMediaTypeUncompressed}, nil
```
mtrmac: Why is this the right value?
- `DesiredLayerCompression` is `PreserveOriginal`
- OCI has different MIME types
- configs have different MIME types

Also, it’s not at all obvious to me that `PutBlob` has any business deciding what the content of the input is; guessing is unreliable, and `PutBlob` has no knowledge of = can’t support `!imageCopier.canModifyManifest`.

directory/directory_dest.go
```diff
@@ -194,7 +195,7 @@ func (d *dirImageDestination) TryReusingBlob(ctx context.Context, info types.Blo
 	if err != nil {
 		return false, types.BlobInfo{}, err
 	}
-	return true, types.BlobInfo{Digest: info.Digest, Size: finfo.Size()}, nil
+	return true, types.BlobInfo{Digest: info.Digest, Size: finfo.Size(), MediaType: manifest.DockerV2SchemeLayerMediaTypeUncompressed}, nil
```
mtrmac: See `PutBlob` above.

mtrmac:
Also, `Buildah` has some code to set up / modify manifest MIME types, please check whether this c/image feature can replace it, or to what extent.

copy/copy.go
vrothberg: > This needs to be specific to the schema (`manifest.Manifest`/`types.Image`) implementation: OCI uses differently-named types.

:+1: 

> (And it probably only should change MIME types if the previous value is exactly what is expected, so that a config MIME type is not replaced by a blob MIME type.)

I don't think that's possible, as the code in question only applies to blobs.

copy/copy.go
vrothberg: Will change, thanks. It's more a precaution in case new types are added.

directory/directory_dest.go
vrothberg: I guess we need to leave this decision to copy then :) Thanks!

copy/copy.go
mtrmac: `copyConfig` calls `copyBlobFromStream` — though you are right that it does not matter, because it sets `canModifyBlob` to `false`.

vrothberg:
> Also, `Buildah` has some code to set up / modify manifest MIME types, please check whether this c/image feature can replace it, or to what extent.

Thanks for the pointer!

vrothberg:
> Also, `Buildah` has some code to set up / modify manifest MIME types, please check whether this c/image feature can replace it, or to what extent.

I think that's the code in question, right?
https://github.com/containers/buildah/blob/4bcddb7cbec960f93f4fea229cc45d9192349b17/cmd/buildah/push.go#L149-L161

mtrmac:
> > Also, `Buildah` has some code to set up / modify manifest MIME types, please check whether this c/image feature can replace it, or to what extent.
> 
> I think that's the code in question, right?
> https://github.com/containers/buildah/blob/4bcddb7cbec960f93f4fea229cc45d9192349b17/cmd/buildah/push.go#L149-L161

No (no mention of compression). 

https://github.com/containers/buildah/blob/4bcddb7cbec960f93f4fea229cc45d9192349b17/image.go#L112 and callers.  (My overall impression is that this should not be reused, but it should mostly  go away; at least in the case of pushes to `docker://`, buildah should not be creating temporary files and compressing layers into them, for `GetBlob` to then stream the temporary files — it should just provide the underlying uncompressed data and let `copy.Image` etc. do the compression and manifest updates. OTOH I’m not sure this is 100% applicable to all the possible destination transports.)

vrothberg:
> https://github.com/containers/buildah/blob/4bcddb7cbec960f93f4fea229cc45d9192349b17/image.go#L112 and callers. (My overall impression is that this should not be reused, but it should mostly go away; at least in the case of pushes to `docker://`, buildah should not be creating temporary files and compressing layers into them, for `GetBlob` to then stream the temporary files — it should just provide the underlying uncompressed data and let `copy.Image` etc. do the compression and manifest updates. OTOH I’m not sure this is 100% applicable to all the possible destination transports.)

@TomSweeneyRedHat @nalind @rhatdan FYI, we might need to revisit that code in buildah.

rhatdan:
I think this is @nalind code.  I am fine with changing it.

rhatdan:
@vrothberg Any movement on this?

rhatdan:
@vrothberg Ping, looking to create a new release of containers image,  any movement?

vrothberg:
@mtrmac, PTAL. The current approach works in local smoke tests but _maybe_ I miss something.

manifest/docker_schema2.go
```diff
@@ -207,7 +207,14 @@ func (m *Schema2) UpdateLayerInfos(layerInfos []types.BlobInfo) error {
...
 	for i, info := range layerInfos {
-		m.LayersDescriptors[i].MediaType = original[i].MediaType
+		switch info.Compression {
+		case types.PreserveOriginal:
+			m.LayersDescriptors[i].MediaType = original[i].MediaType
+		case types.Decompress:
+			m.LayersDescriptors[i].MediaType = DockerV2SchemaLayerMediaTypeUncompressed
+		case types.Compress:
+			m.LayersDescriptors[i].MediaType = DockerV2SchemaLayerMediaType
+		}
```
mtrmac: Yes, this implementation approach accounts for the different MIME types in different manifest formats, but this will also soon need to handle different compression algorithms (gzip vs. Zstd).  Very broadly like `compression.Algorithm` in #639 , but that does not seem to provide constants for known formats yet.

vrothberg: @mtrmac @giuseppe, so once #639 is merged, we need to add a new MIME type for `zstd` (e.g., `application/vnd.docker.image.rootfs.diff.tar.zstd` and `application/vnd.oci.image.layer.v1.tar+zstd`), right?

mtrmac: Well, that, and `BlobInfo.Compression` will either need to be replaced or augmented to also record the algorithm.

image/docker_schema2.go
```diff
@@ -207,12 +208,15 @@ func (m *manifestSchema2) convertToManifestOCI1(ctx context.Context) (types.Imag
...
+		switch m.m.LayersDescriptors[idx].MediaType {
+		case manifest.DockerV2Schema2ForeignLayerMediaType:
 			layers[idx].MediaType = imgspecv1.MediaTypeImageLayerNonDistributable
-		} else {
-			// we assume layers are gzip'ed because docker v2s2 only deals with
-			// gzip'ed layers. However, OCI has non-gzip'ed layers as well.
+		case manifest.DockerV2Schema2LayerMediaType:
 			layers[idx].MediaType = imgspecv1.MediaTypeImageLayerGzip
+		case manifest.DockerV2SchemaLayerMediaTypeUncompressed:
+			layers[idx].MediaType = imgspecv1.MediaTypeImageLayer
```
vrothberg: @mtrmac, FYI

storage/storage_image.go
```diff
@@ -345,10 +345,10 @@ func (s *storageImageDestination) Close() error {
...
 
 func (s *storageImageDestination) DesiredLayerCompression() types.LayerCompression {
-	// We ultimately have to decompress layers to populate trees on disk,
-	// so callers shouldn't bother compressing them before handing them to
-	// us, if they're not already compressed.
-	return types.PreserveOriginal
+	// We ultimately have to decompress layers to populate trees on disk
+	// and need to explicitly ask for it here, so that the layers' MIME
+	// types can be set accordingly.
+	return types.Decompress
```
vrothberg: @mtrmac, FYI - this was also necessary to fix the (docker -> storage -> dir) cycle

mtrmac: This is definitely conceptually correct, but it also significantly changes the assumptions of the code, e.g. see the `RecordDigestUncompressedPair` call in `PutBlob`, and I’m sure the `LayersByCompressedDigest` calls get the “compressed digest” value from something in this file. I don’t think this one-liner alone suffices (but I haven’t actually re-read all of the code now).

mtrmac: Or maybe it would be easier to keep the current approach, but remap the MIME types in the source’s `LayerInfosForCopy`. Easier, not necessarily cleaner.

vrothberg: I'll check that. It feels like there are a bunch of traps but we're slowly getting there. I will add tests to Skopeo to make sure we're not regressing etc.

image/docker_schema2.go
mtrmac: What is the part I should notice? That the code now uses `manifest.DockerV2SchemaLayerMediaTypeUncompressed`?

vrothberg: Yes. Wanted to have your pair of eyes on it.

mtrmac: ACK. It’s not in the spec, but containerd recognizes it (and most other tools just ignore MIME types).

storage/storage_image.go
mtrmac: Also, IIRC, decompressing the layers in generic code causes `PutManifest` to be called with an updated manifest, and that in turn breaks digest references using the digest from the original registry.

It’s all one big knot of entangled wool…

One way out of that could be to just not validate the digests (see also https://github.com/cri-o/cri-o/issues/2157 ); I also have an old branch that differentiates `{Put,Get}{,Original}Manifest` so that c/storage could record _both_ the original manifest (with the original digest, and which can be verified using signatures) and a modified manifest that actually corresponds to the uncompressed layers lying somewhere. In the meantime the `LayerInfosForCopy` hack emerged instead, and I’m pretty torn WRT continuing to rely on it more and more vs. the significant rework for the two-manifests approach.

types/types.go
```diff
@@ -91,14 +91,29 @@ type ImageReference interface {
...
-	URLs        []string
-	Annotations map[string]string
-	MediaType   string
+	Digest               digest.Digest // "" if unknown.
+	Size                 int64         // -1 if unknown
+	URLs                 []string
+	Annotations          map[string]string
+	MediaType            string
+	Compression          LayerCompression
+	CompressionAlgorithm string // as defined pkg/compression.{GzipCompression, ZstdCompression, ...}
```
mtrmac: Can’t this be the native `compression.Algorithm` type directly? (I guess that runs into visibility issues?)

copy/copy.go
```diff
@@ -899,6 +899,12 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
 		return types.BlobInfo{}, errors.Wrap(err, "Error writing blob")
 	}
 
+	// If we can modify the layer's blob, set the desired compression for it to be set in the manifest.
+	if canModifyBlob && !isConfig {
+		uploadedInfo.Compression = c.dest.DesiredLayerCompression()
```
mtrmac: Shouldn’t this be directly based on `compressionOperation` and/or the logic that set it instead of second-guessing the logic above here? It’s going to be difficult to keep the two in sync.

copy/copy.go
```diff
@@ -899,6 +899,12 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
...
+		uploadedInfo.CompressionAlgorithm = desiredCompressionFormat.Name()
+	}
```
mtrmac: In the `else` case, is `Compression` is still set and/or consumed? I can’t see anything in here setting the values on every manifest read.

We either need the values to be always filled in and present (by every single producer), or there will need to be an “unknown”/“don’t edit” value.  (Abstractly, I’d weakly prefer the former, but that would probably benefit from wrapping `BlobInfo` in constructors to enforce the field presence. Adding a “don’t edit” state would probably be easier right now.)

manifest/docker_schema2.go
```diff
@@ -207,7 +209,23 @@ func (m *Schema2) UpdateLayerInfos(layerInfos []types.BlobInfo) error {
 	original := m.LayersDescriptors
 	m.LayersDescriptors = make([]Schema2Descriptor, len(layerInfos))
 	for i, info := range layerInfos {
-		m.LayersDescriptors[i].MediaType = original[i].MediaType
+		switch info.Compression {
```
mtrmac: (Non-blocking, probably only theoretical at this point:

Should this only override known `MediaType` values? I have no idea, maybe not; it’s at least not immediately obvious.

Suppose someone uses an `com.example.org/completely-standards-violating-x-this-is-a-pretty-icon-for-the-container` MIME type and lists it in the manifest’s layer list; is it OK to just overwrite it because we decided to compress the layer? Arguably we should not have compressed a layer with an unknown MIME type in the first place…)

image/oci.go
```diff
@@ -187,7 +188,18 @@ func (m *manifestOCI1) convertToManifestSchema2() (types.Image, error) {
 	layers := make([]manifest.Schema2Descriptor, len(m.m.Layers))
 	for idx := range layers {
 		layers[idx] = schema2DescriptorFromOCI1Descriptor(m.m.Layers[idx])
-		layers[idx].MediaType = manifest.DockerV2Schema2LayerMediaType
+		switch layers[idx].MediaType {
+		case imgspecv1.MediaTypeImageLayerNonDistributable:
```
mtrmac: Note that https://github.com/opencontainers/image-spec/pull/788 adds also the compressed variants for non-distributable layers. Are there (should there be) corresponding schema2 variants as well?

pkg/compression/compression.go
```diff
@@ -13,6 +13,17 @@ import (
 	"github.com/ulikunitz/xz"
 )
 
+const (
+	// GzipCompression indicates gzip compression.
+	GzipCompression = "gzip"
```
mtrmac: I kind-of expected these constants to directly be `compression.Algorithm` values. (But I still need to catch up on #639 , so I’m not sure it’s completely viable.)

mtrmac: `compression.GzipCompression` seems unnecessarily redundant.

mtrmac: Now that the constants exist, could the hard-coded default in `copy.Image` use the symbolic constant instead of a string?

storage/storage_image.go
mtrmac: (I don’t think this has been solved yet.)

types/types.go
```diff
@@ -91,14 +91,29 @@ type ImageReference interface {
...
+	// no compression or decompression.
+	PreserveOriginal LayerCompression = iota
+	// Decompress indicates the layer must be decompressed
+	Decompress
+	// Compress indicates the layer must be compressed
+	Compress
+)
+
 // BlobInfo collects known information about a blob (layer/config).
 // In some situations, some fields may be unknown, in others they may be mandatory; documenting an “unknown” value here does not override that.
```
mtrmac: Does the rest of the interface need updating for the presence/validity of `Compression` and `CompressionAlgorithm`?  (I suppose, let’s start with the code and use the documentation as the final check that everything is accounted for.)

manifest/manifest.go
```diff
@@ -24,10 +24,17 @@ const (
...
+	DockerV2Schema2LayerMediaTypeZstd = "application/vnd.docker.image.rootfs.diff.tar.zstd"
+	// DockerV2SchemaLayerMediaTypeUncompressed is the mediaType used for uncompressed layers.
+	DockerV2SchemaLayerMediaTypeUncompressed = "application/vnd.docker.image.rootfs.diff.tar"
 	// DockerV2ListMediaType MIME type represents Docker manifest schema 2 list
 	DockerV2ListMediaType = "application/vnd.docker.distribution.manifest.list.v2+json"
 	// DockerV2Schema2ForeignLayerMediaType is the MIME type used for schema 2 foreign layers.
 	DockerV2Schema2ForeignLayerMediaType = "application/vnd.docker.image.rootfs.foreign.diff.tar.gzip"
+	// OCIV1MediaTypeImageLayerZstd is the MIME types used for OCI v1 layers compressed with zstd.
+	// Note that this MIME type is not (yet) part of the OCI v1 image specification.
+	OCIV1MediaTypeImageLayerZstd = "application/vnd.oci.image.layer.v1.tar+zstd"
```
mtrmac: Hum, https://github.com/opencontainers/image-spec/pull/788 has added the textual specification but not the Go constant.

vrothberg: Opened a PR to add them.

types/types.go
vrothberg: Yes, let's get the final polish when the code is ready.

pkg/compression/compression.go
vrothberg: Agreed. Will update with the next push.

types/types.go
vrothberg: That should be possible and I prefer your suggestion (in combination with a "don't edit" switch).

copy/copy.go
vrothberg: I think we can get the "don't edit" state by having a `Compression *compression.Algorithm` with `nil` being a "don't edit".

copy/copy.go
vrothberg: It should be `compressionOperation`. Good catch!

storage/storage_image.go
vrothberg: I think we can keep lying :^) The (docker -> storage -> dir) cycle works without the change. Am I missing some potential side effect?

copy/copy.go
mtrmac: Note that we need to differentiate between “decompress” and “don’t edit”. Would ”don’t edit” be (`LayerCompression.Compress`, `nil`)? That is good enough for a computer but rather confusing for people.

---

Actually the code already implements “don’t edit” using `PreserveOriginal`. That’s semantically… not _quite_ precise but close enough, and at least I can’t imagine what else the code should do with a `PreserveOriginal` value.

storage/storage_image.go
mtrmac: AFAICS `LayerInfosForCopy` does return the uncompressed MIME type, so keeping lying should be transparent to both the current compress/decompress decision logic and the new MIME type editing logic.

copy/copy.go
vrothberg: > Actually the code already implements “don’t edit” using PreserveOriginal.

That's what I thought as well. I think the combination of operation and algorithm works well - although I think we shouldn't use a `compress.Algorithm` as it might trick users into believing they could specify custom (de)compression functions (we're only interested in the name).

mtrmac: True, the `Compression` field is used in `Image.UpdatedImage`, which could be confusing to mean that the contents will get automatically recompressed, but the immediate container is called *`Manifest`*`UpdateOptions`…

Maybe name the field not `Compression`, but some smarter variant of `MIMETypeCompressionAdjustment`, then?

---

WRT the users passing in an unknown algorithm, that’s actually much less hassle with strings than with `compression.Algorithm` values (right now `Algorithm` is a private struct, so it’s completely impossible; even assuming it becomes a public interface, implementing an interface is more work. And maybe we can add a private function to the interface to make it exclusive, not sure whether Go allows that). And after all, your `UpdateLayerInfos` implementation already rejects unknown values, so I don’t think it matters either way.

---

This is basically the fallout of being lax before, and abusing `types.BlobInfo` as an universal carrier of various subsets of data instead of being explicit in the type system about what is / isn’t available/expected in which situations. Maybe there should be a `BlobInfo` and `BlobInfoEdits`. Or move the compression edits to be a standalone array in `ManifestUpdateInformation`?  

In particular it’s inconsistent / it does not make all that much sense to have _both_ `MediaType` and the two compression _edits_ to `MediaType` in a single structure… and then have `ManifestUpdateOptions` say that “`MediaType` fields are ignored”.

Yet fixing this cleanly has ABI implications, which immediately leads to thinking about which parts of the write path should be an ABI / “unstable public API” / …, and I suspect there isn’t much appetite for solving that this week.

vrothberg: We're getting close! I will reflect on Monday about that. Thanks for sharing your thoughts in such great detail!

image/oci.go
vrothberg: I have mixed feelings about that. We're already lying about schema2 knowing about zstd, so it would make sense to add a foreign zstd type as well.

However, I am not sure if we shouldn't just throw zstd-errors for schema2 images and focus on OCI?

manifest/docker_schema2.go
vrothberg: I'll add a `// TODO` in the code. It sounds like a topic for another PR - potentially bigger since it brings up the question of which mime types does format X support and how and where do we enforce it.

image/oci.go
vrothberg: I added them for now and updated the code conversion and update paths to take them into consideration as well.

pkg/compression/compression.go
vrothberg: Very good idea. That's changed accordingly now.

copy/copy.go
vrothberg: I left it as is (for now). As you mention above, the compression function of `compression.Algorithm` is private, so there's no real risk at the moment.

vrothberg:
@mtrmac, if you have some cycles, PTAL. Notable changes from today:
* Added tests to `image` (image conversions) and `manifest` (media type changes)
* Completed non-distributable media types, their treatment during manifest updates and during image conversions.
* Some more documentation.
* Addressed previous comments.

I'll do some more smoke tests tomorrow but a code-review would be great.

image/oci.go
mtrmac: Arguably we should not be compressing schema2 images to ZSTD in the first place, at least until the official schema2 specification adds that.

OTOH that’s yet more infrastructure to build, and to an extent defeats the point of adding the feature, as long as OCI is not the prevalent format.

I don’t know; it’s just a thing that I don’t think has been actually discussed, and probably should have been.

Cc: @giuseppe

vrothberg: Shall we error out for now? I don't want to add on-the-fly recompression to this PR as we need to get that in rather soon. 

(@giuseppe is on PTO this week.)

vrothberg:
Smoke tests are looking good. CI is green. We're getting closer.

The last remaining issue is whether we should pretend that Schema 2 supported Zstd or not (and error out during conversion).

@estesp, once that's in and re-vendored, buildah/containerd will like each other again.

vrothberg:
@mtrmac, @nalind, PTAL.

rhatdan:
https://github.com/containers/libpod/pull/3457

Sadly it is being blocked on `podman volume` fixes.

vrothberg:
> [containers/libpod#3457](https://github.com/containers/libpod/pull/3457)
> 
> Sadly it is being blocked on `podman volume` fixes.

To avoid confusion: I think this comment relates to https://github.com/containers/buildah/pull/1822#pullrequestreview-281264039

vrothberg:
Rebased to pull in the go module.

@mtrmac @nalind, can we get this in? Other PRs are depending on the changes and there are some warts when building containers letting containerd (validly) reject the images.

rhatdan:
I think we should throw an error unless you are using OCI images.

vrothberg:
> I think we should throw an error unless you are using OCI images.

@mtrmac @nalind @giuseppe , do you agree on limiting zstd support to OCI and error out for non-OCI ones?

giuseppe:
@vrothberg that works for me

vrothberg:
@giuseppe @mtrmac @rhatdan, I updated the PR to limit zstd support to OCIv1. This changed the compat matrix during conversion, so I added some more aggressive checks for supported media types.

PTAL, I really want to get this PR off my table.

rhatdan:
LGTM
@mtrmac @giuseppe PTAL

vrothberg:
@nalind, mind taking a look?

We are receiving more and more requests to get those fixes finally in and I am getting impatient.

vrothberg:
Folks, we need this in. Not only to unblock other PRs but also to ship these fixes.

@rhatdan @mtrmac @giuseppe @nalind PTAL

giuseppe:
LGTM

PR #568, opened 2019-01-28
Make docker/tarfile's GetBlob thread-safe

### LINKED ISSUE #567: Allow parallel copying from docker-archives
Is there any reason that `HasThreadSafeGetBlob` for `docker/tarball` is marked `false`?

From what I can tell, all IO in `GetBlob` seems to go through `openTarComponent` which opens a new file descriptor for each call, which should be thread-safe. `ensureCachedDataIsPresent` mutates state, but should be idempotent from what I can tell.

/cc @vrothberg

teozkr:
This fixes #567, and *should* be safe from what I can tell (according to the
reasoning I mention in that issue).

### DIFF
--- a/docker/tarfile/src.go
+++ b/docker/tarfile/src.go
@@ -399,7 +399,7 @@ func (r uncompressedReadCloser) Close() error {
 
 // HasThreadSafeGetBlob indicates whether GetBlob can be executed concurrently.
 func (s *Source) HasThreadSafeGetBlob() bool {
-	return false
+	return true
 }
 
 // GetBlob returns a stream for the specified blob, and the blob’s size (or -1 if unknown).
--- a/docker/tarfile/src.go
+++ b/docker/tarfile/src.go
@@ -9,6 +9,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
+	"sync"
 
 	"github.com/containers/image/internal/tmpdir"
 	"github.com/containers/image/manifest"
@@ -21,8 +22,10 @@ import (
 // Source is a partial implementation of types.ImageSource for reading from tarPath.
 type Source struct {
 	tarPath              string
-	removeTarPathOnClose bool // Remove temp file on close if true
+	removeTarPathOnClose bool      // Remove temp file on close if true
+	cacheDataLock        sync.Once // Atomic way to ensure that ensureCachedDataIsPresent is only invoked once
 	// The following data is only available after ensureCachedDataIsPresent() succeeds
+	cacheDataResult   error         // The return value of ensureCachedDataIsPresent, since it should be as safe to cache as the side effects
 	tarManifest       *ManifestItem // nil if not available yet.
 	configBytes       []byte
 	configDigest      digest.Digest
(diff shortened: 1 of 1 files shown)


mtrmac:
Thanks for the PR.  You are right that this should be possible to make thread-safe.

I’m afraid this PR is not doing that, though, and actual locking (or an “initialize once” mechanism, or maybe moving the initialization into creation of the `ImageSource`) will be necessary.

You are right that `ensureCachedDataIsPresent` is _conceptually_ idempotent, but the _implementation_ is not:
- Per https://golang.org/ref/mem , nothing seems to guarantee that individual field reads/writes are atomic; it’s not in general OK to read a value while another goroutine may be writing into it.
- `ensureCachedDataIsPresent` executes depending on `s.tarManifest`, which is, in the source code order, set _first_ of all the output values, so concurrent consumers may read `tarManifest`, and think that they can rely on the other values being present
- Per https://golang.org/ref/mem , again, the ordering of reads/writes observed from a different goroutine is not necessarily the same as the one in the source code order, anyway, so just moving the setting of `s.tarManifest` to the end of the function would not help.

teozkr:
Fair enough, I was afraid that there were bigger dragons hiding here.

It looks like it should be enough to wrap `ensureCachedDataIsPresent` in [`Once.Do`](https://golang.org/pkg/sync/#Once.Do), from what I can tell?

teozkr:
5f7d385 adds locks around `ensureCachedDataIsPresent`, but as mentioned in the commit message it also means that any errors are also cached.

mtrmac:
I think moving this into the constructor would be simpler — any real user is going to cause the `ensureCachedDataIsPresent` call to be called eventually, and we would not need any locks / synchronization mechanism; any fully instantiated object has the cached data available.

teozkr:
That would change the external interface, since some errors would be moved from usage to construction. Then again, `docker/daemon`'s constructor seems to be able to raise errors anyway, so clients should already be able to handle it.

teozkr:
The Travis failure seems to be a fluke about not being able to update the Fedora repos.

mtrmac:
I’m very sorry, I have missed the case of public `Source.LoadTarManifest`, which should complete quickly without having to decompress all the layers (it’s used in `libpod` to decide how to process the output of `c/image/docker/archive.NewImageReference`; so, the `Once.Do` approach is, after all, preferable to doing all that work in the constructors.

Do you want to restore https://github.com/containers/image/commit/5f7d385a0eb7b854c77ac53b21e48e8ba37ba04b , which now LTGM?

Or, if you want, I can do handle the PR process so that I don’t burden you with even more work.

teozkr:
Done. Good to know I'm not the only one who makes mistakes! :D

mtrmac:
LGTM pending tests. Thanks!

teozkr:
Thanks! Should I submit a corresponding PR to update Skopeo's dependency, or is there a process for batching that up occasionally?

mtrmac:
There’s no regular process, but c/image will certainly be updated before another Skopeo release.

If you for some reason want this in the Skopeo `master` branch now, a PR (update `vendor.conf`, run `make vendor`) would be nice.

PR #569, opened 2019-01-29
sysregistriesv2: Export registries struct for v1 format

umohnani8:
Need the v1 registries structs to be public so that I can
use it in the CRD and controller stuff for container runtime
config in the machine-config-operator repo.

### DIFF
--- a/pkg/sysregistriesv2/system_registries_v2.go
+++ b/pkg/sysregistriesv2/system_registries_v2.go
@@ -53,20 +53,23 @@ type Registry struct {
 	Prefix string `toml:"prefix"`
 }
 
-// backwards compatability to sysregistries v1
-type v1TOMLregistries struct {
+// V1TOMLregistries is for backwards compatibility to sysregistries v1
+type V1TOMLregistries struct {
 	Registries []string `toml:"registries"`
+}
+
+// V1TOMLConfig is for backwards compatibility to sysregistries v1
+type V1TOMLConfig struct {
+	Search   V1TOMLregistries `toml:"search"`
+	Insecure V1TOMLregistries `toml:"insecure"`
+	Block    V1TOMLregistries `toml:"block"`
 }
 
 // tomlConfig is the data type used to unmarshal the toml config.
 type tomlConfig struct {
 	Registries []Registry `toml:"registry"`
 	// backwards compatability to sysregistries v1
-	V1Registries struct {
-		Search   v1TOMLregistries `toml:"search"`
-		Insecure v1TOMLregistries `toml:"insecure"`
-		Block    v1TOMLregistries `toml:"block"`
-	} `toml:"registries"`
+	V1TOMLConfig `toml:"registries"`
 }
 
 // InvalidRegistries represents an invalid registry configurations.  An example
@@ -129,21 +132,21 @@ func getV1Registries(config *tomlConfig) ([]Registry, error) {
 
 	// Note: config.V1Registries.Search needs to be processed first to ensure registryOrder is populated in the right order
 	// if one of the search registries is also in one of the other lists.
-	for _, search := range config.V1Registries.Search.Registries {
+	for _, search := range config.V1TOMLConfig.Search.Registries {
 		reg, err := getRegistry(search)
 		if err != nil {
 			return nil, err
 		}
 		reg.Search = true
 	}
-	for _, blocked := range config.V1Registries.Block.Registries {
+	for _, blocked := range config.V1TOMLConfig.Block.Registries {
 		reg, err := getRegistry(blocked)
 		if err != nil {
 			return nil, err
 		}
 		reg.Blocked = true
 	}
-	for _, insecure := range config.V1Registries.Insecure.Registries {
+	for _, insecure := range config.V1TOMLConfig.Insecure.Registries {
 		reg, err := getRegistry(insecure)
 		if err != nil {
 			return nil, err


umohnani8:
@mtrmac @mrunalp PTAL

TomSweeneyRedHat:
@vrothberg PTAL

vrothberg:
Can elaborate a bit on it? We had a longer discussion in Brno about the future of this package and wanted to push it into a direction where only APIs are being called. The reasoning behind is that many users (Buildah, Podman, CRI-O) are more or less working around the package's API which makes update hard, entails boilerplate code etc.

In other words, can we tweak the API instead of exposing internal types?

umohnani8:
@vrothberg what is happening the CRD is that I need to read in a template we have with the default values, so need to convert the TOML file into its corresponding Go structs. Then I will merge in the changes the user provides from their CR, convert that back to the correct Go structs and then write that to the registries.conf file. And since right now the registries.conf file is still using the v1 registries format, I need the v1 structs exposed. I guess I can redefine those structs in the CRD and controller code, but was trying to reuse what was already available.

vrothberg:
Thanks for providing the details, @umohnani8 !

>  I guess I can redefine those structs in the CRD and controller code, but was trying to reuse what was already available.

 If that's possible and doesn't cause problems, I would prefer doing that. @mtrmac, what do you think?

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -53,19 +53,19 @@ type Registry struct {
...
 	Registries []string `toml:"registries"`
 }
 
-// tomlConfig is the data type used to unmarshal the toml config.
-type tomlConfig struct {
+// TomlConfig is the data type used to unmarshal the toml config.
+type TomlConfig struct {
 	Registries []Registry `toml:"registry"`
 	// backwards compatability to sysregistries v1
 	V1Registries struct {
```
mtrmac: I don’t think this structure, which combines both v1 and v2, and is really an implementation detail of how this package does parsing, should be made public.

If only the v1 structure is necessary, would it be possible to make a public `V1TOMLConfig` which contains only the `V1Registries` field (possibly renamed named `Registries`), and embed that `V1TOMLConfig` as an unnamed field within (private) `tomlConfig`?

mtrmac:
> > I guess I can redefine those structs in the CRD and controller code, but was trying to reuse what was already available.
> 
> If that's possible and doesn't cause problems, I would prefer doing that. @mtrmac, what do you think?

\*shrug\* We’ve run into this situation in a few other times (notably `podman trust` editing `policy.json`) — there’s the ideal of only exposing evaluation and setting of a configuration through an API with well-defined semantics, vs. the convenience of making the data structures public and allowing anyone to easily consume/produce it, hard-coding the semantics in the process.

I think it would be quite fine to provide constructors for objects which have private members (i.e. can’t be interpreted by out-of-package users) but can be formatted as valid config files.

But if you need to read and update… To be mostly flippant but 10% serious, make the definitions public in a `c/image/pkg/sysregistriesv2/onlyuseinreadwriteeditors` subpackage?

Overall, I think the case for making the data structure public is a bit stronger for `registries.conf` than for `policy.json`, where for signature checking we very strongly want to consolidate everyone to the same implementation; `registries.conf` does include security-relevant data, and consistent handling of it _is_ important, but it’s somewhat less urgent than with the crypto, the scope lookup mechanisms, and the need to avoid non-FIPS-validatable independent reimplementations.

It’s not quite plausible that the burden of copying the 20 lines of data structures will ever stop anyone from independently interpreting `registries.conf` :), so perhaps we might just as well provide a version, and surround it with comments directing users to use the API if they at all can.

---

> what is happening the CRD is that I need to read in a template we have with the default values, so need to convert the TOML file into its corresponding Go structs.

Is reading a “template” truly necessary? It sounds a tiny bit inconsistent to have “default values” in a variable template (OTOH, sure, hierarchical systems of defaults/fallbacks are common). Would a hard-coded set of values within Go work just as conveniently for you? (That would allow us to use the “public constructors of private data structures” approach, the way https://github.com/openshift/origin/pull/21864 does in `createBuildSignaturePolicyData` .)

Or, perhaps, is this all intended to be used in https://github.com/openshift/origin/pull/21864 ? That one only needs unmarshaling in tests, and I wouldn’t object to hard-coding the data structures in tests at all.

---

(Another aspect to this is whether the user should use the v1 or v2 format — it’s conceptually awkward to use the older one, but in practice it’s the only reasonable choice, when we are still (weakly?) considering possibly changing v2 for mirror support; v2 would right now only be practical through an API.)

umohnani8:
@mtrmac thanks for the feedback. I will just redefine these structs in the CRD and Controller code. I noticed that the tomlConfig structs for crio.conf and storage.conf are private as well and will have to do the same for those anyway. Closing this PR now.

mtrmac:
FWIW, I guess it got lost in that long comment — on balance, I think making the v1 format structs public is acceptable.

pkg/sysregistriesv2/system_registries_v2.go
mtrmac: Did you want public only the definition of the `V1TOMLRegistries` element, or all of `V1Registries` with the top-level `search`/`insecure`/`block` fields?

umohnani8: Yup, just did that. I followed what you sort of described in the comment above. Let me know if I missed something.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -53,20 +53,23 @@ type Registry struct {
 	Prefix string `toml:"prefix"`
 }
 
-// backwards compatability to sysregistries v1
-type v1TOMLregistries struct {
+// V1TOMLregistries is for backwards compatability to sysregistries v1
```
mtrmac: (Nit: s/compatability/compatibility/)

umohnani8:
@mtrmac tests are green

PR #570, opened 2019-01-29
update openshift API paths to groupified path

deads2k:
In 3.7, openshift moved to use groupified kubernetes APIs.  This pull updates the containers/image project to use these new APIs.  The old-style `oapi` endpoint is being removed in openshift v4.

### DIFF
--- a/openshift/openshift.go
+++ b/openshift/openshift.go
@@ -136,7 +136,7 @@ func (c *openshiftClient) doRequest(ctx context.Context, method, path string, re
 // getImage loads the specified image object.
 func (c *openshiftClient) getImage(ctx context.Context, imageStreamImageName string) (*image, error) {
 	// FIXME: validate components per validation.IsValidPathSegmentName?
-	path := fmt.Sprintf("/oapi/v1/namespaces/%s/imagestreamimages/%s@%s", c.ref.namespace, c.ref.stream, imageStreamImageName)
+	path := fmt.Sprintf("/apis/image.openshift.io/v1/namespaces/%s/imagestreamimages/%s@%s", c.ref.namespace, c.ref.stream, imageStreamImageName)
 	body, err := c.doRequest(ctx, "GET", path, nil)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


deads2k:
@mtrmac Can I get a little help finding the failing tests? `TestCopyDirSignatures` doesn't appear in my grep of this codebase.  Also, the forbidden message leads me to wonder if this is being tested against an old level of openshift.

mtrmac:
@deads2k The context to this is that c/image/openshift, == the `atomic:` transport, are deprecated; since https://github.com/openshift/origin/pull/12504 (before v1.5.0-alpha.3), the registry instead provides access by an extension of the docker/distribution API, which is more convenient because it can work without an extra `oc login`.

If I am checking correctly, the `/apis/image.openshift.io` namespace was not yet hooked up in 1.5.0 (it only came in 3.6), i.e. we would be breaking users of registries not even two years old.

Is it possible to autodetect which of the two namespaces is supported? That would be ideal.

(I don’t know; maybe we can break support for <3.7 registries (@rhatdan ?), but it’s not very attractive at a glance.)

---

FWIW the tests are in https://github.com/containers/skopeo/blob/master/integration/copy_test.go , and https://github.com/containers/skopeo/blob/master/Dockerfile indeed installs the very first v1.5.0-alpha.3 version.

(From the testing perspective, it’s convenient that these old versions can reasonably practically run the master and the registry within a single container, i.e. all of our tests run in a single container and we don’t have to worry about networking , actually running  the tests on Ubuntu hosts in Travis, or anything else.)

mtrmac:
> @deads2k The context to this is that c/image/openshift, == the `atomic:` transport, are deprecated; since [openshift/origin#12504](https://github.com/openshift/origin/pull/12504) (before v1.5.0-alpha.3), the registry instead provides access by an extension of the docker/distribution API, which is more convenient because it can work without an extra `oc login`.

… and users can refer to the images using `docker://`, as with any other images, ignoring the fact that the registry is an OpenShift one.

rhatdan:
@deads2k @mtrmac @vrothberg What is the state of this PR?  Is this still something we want or should we close it?

mtrmac:
Ideally, it would be nice for the transport to auto-detect the API path. If we have to hard-code, it makes more sense to me to keep hard-coding the old version, because there’s a workaround/improvement available for using the newer servers.

(FWIW it turns out that OpenShift has merged this change in its fork of c/image, instead of changing whatever code depends on `atomic:` to use `docker://` [— and I can’t find any such uses in the openshift/origin repo, except a CLI invocation of skopeo in a test that was disabled, anyway]. https://github.com/containers/image/compare/master...openshift:openshift-4.0-image-v1.3 )

PR #571, opened 2019-01-30
switch to openshift image api group instead from deprecated oapi

mfojtik:
@mtrmac @rhatdan the `/oapi` is going to be removed in 4.0, we need to switch this to use the proper openshift api group.


runcom:
is this a dup of https://github.com/containers/image/pull/570?

mfojtik:
yeah, David was faster...

PR #573, opened 2019-01-31
Don't update manifests when writing to digested references

nalind:
Don't make changes to manifests or the blobs they reference when we're writing to a location that's specified using a digested reference, since changing the manifest will break the reference.

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -13,7 +13,9 @@ import (
 	"sync"
 	"time"
 
+	"github.com/containers/image/docker/reference"
 	"github.com/containers/image/image"
+	"github.com/containers/image/manifest"
 	"github.com/containers/image/pkg/blobinfocache"
 	"github.com/containers/image/pkg/compression"
 	"github.com/containers/image/signature"
@@ -212,7 +214,7 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
 
 // Image copies a single (on-manifest-list) image unparsedImage, using policyContext to validate
 // source image admissibility.
-func (c *copier) copyOneImage(ctx context.Context, policyContext *signature.PolicyContext, options *Options, unparsedImage *image.UnparsedImage) (manifest []byte, retErr error) {
+func (c *copier) copyOneImage(ctx context.Context, policyContext *signature.PolicyContext, options *Options, unparsedImage *image.UnparsedImage) (manifestBytes []byte, retErr error) {
 	// The caller is handling manifest lists; this could happen only if a manifest list contains a manifest list.
 	// Make sure we fail cleanly in such cases.
 	multiImage, err := isMultiImage(ctx, unparsedImage)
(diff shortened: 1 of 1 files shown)


mtrmac:
There’s definitely some merit to this — using digested references for destinations can’t work if `copy.Image` can just modify the manifest.

Strictly speaking it’s only enforcing that the destination specifies _some_ digest, not the digest of the original manifest.

More to the point, I don’t think this can fix the underlying cause of the mismatches on pull: consider (private) https://bugzilla.redhat.com/show_bug.cgi?id=1669096#c5 .  The primary cause is that c/storage deduplicates by config, and returns the last pushed manifest for that config (which is the only one recorded), not a manifest that matches the requested digest (which would require recording all manifests).

nalind:
Ah, revised to enforce an exact match.

copy/copy.go
```diff
@@ -548,7 +563,7 @@ func layerDigestsDiffer(a, b []types.BlobInfo) bool {
 
 // copyUpdatedConfigAndManifest updates the image per ic.manifestUpdates, if necessary,
 // stores the resulting config and manifest to the destination, and returns the stored manifest.
-func (ic *imageCopier) copyUpdatedConfigAndManifest(ctx context.Context) ([]byte, error) {
+func (ic *imageCopier) copyUpdatedConfigAndManifest(ctx context.Context, expectedDigest *digest.Digest) ([]byte, error) {
```
mtrmac: Check for `expectedDigest` all the way down here is unnecessarily late.

Instead, please do it just after `image.FromUnparsedImage`, in the same block that sets `destisDigestedReference`.  That will mean that users submitting digests that don’t match the source will see an immediate failure instead of having to wait for all layers to be copied.

And if `!c.canModifyManifest`, `copyUpdatedConfigAndManifest` already uses `ic.src.Manifest`, which is exactly the source manifest unmodified, so an extra check that it matches the digest at this point is not necessary (if the manifest was checked before copying layers).

copy/copy.go
```diff
@@ -569,20 +584,29 @@ func (ic *imageCopier) copyUpdatedConfigAndManifest(ctx context.Context) ([]byte
...
+	manifestBytes, _, err := pendingImage.Manifest(ctx)
 	if err != nil {
 		return nil, errors.Wrap(err, "Error reading manifest")
 	}
+	if expectedDigest != nil {
+		manifestDigest, err := manifest.Digest(manifestBytes)
+		if err != nil {
+			return nil, errors.Wrapf(err, "Error computing digest of pending manifest")
+		}
+		if manifestDigest != *expectedDigest {
```
mtrmac: (in addition to moving the code, as discussed elsewhere)

Please use `manifest.MatchesDigest`, to have some chance of supporting multiple different digest algorithms in the future.

copy/copy.go
```diff
@@ -262,14 +277,14 @@ func (c *copier) copyOneImage(ctx context.Context, policyContext *signature.Poli
...
-		canModifyManifest: len(sigs) == 0,
+		canModifyManifest: len(sigs) == 0 && !destIsDigestedReference,
 		// Ensure _this_ copy sees exactly the intended data when either processing a signed image or signing it.
 		// This may be too conservative, but for now, better safe than sorry, _especially_ on the SignBy path:
 		// The signature makes the content non-repudiable, so it very much matters that the signature is made over exactly what the user intended.
 		// We do intend the RecordDigestUncompressedPair calls to only work with reliable data, but at least there’s a risk
 		// that the compressed version coming from a third party may be designed to attack some other decompressor implementation,
 		// and we would reuse and sign it.
-		canSubstituteBlobs: len(sigs) == 0 && options.SignBy == "",
+		canSubstituteBlobs: len(sigs) == 0 && options.SignBy == "" && !destIsDigestedReference,
```
mtrmac: (Hum, at this point it might be more maintainable to have a separate
```go
ic.canSubstituteBlobs = ic.canModifyManifest && options.SignBy == ""
```
)

copy/copy.go
nalind: Ah, right.  Changing it.

copy/copy.go
```diff
@@ -224,6 +226,19 @@ func (c *copier) copyOneImage(ctx context.Context, policyContext *signature.Poli
...
+	if ref := c.dest.Reference(); ref != nil {
+		if named := ref.DockerReference(); named != nil {
+			if digested, ok := named.(reference.Digested); ok {
+				destIsDigestedReference = true
+				referenceDigest := digested.Digest()
+				expectedDigest = &referenceDigest
+			}
+		}
+	}
+
```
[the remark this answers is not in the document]
nalind: Moving this bit.

copy/copy.go
nalind: Moved it.

copy/copy.go
nalind: Made the change as you described.

PR #574, opened 2019-01-31
storageImageDestination: look up by compressed digests

nalind:
In storageImageDestination.TryReusingBlob(), if the blobinfo cache says that the digest of the desired blob is that of a compressed blob, and the blobinfo cache knows the corresponding uncompressed blob digest, try to use that uncompressed digest to locate matching layers, and if any of them have a recorded compressed digest that matches the blobinfo's digest, use the corresponding recorded size for the compressed blob to avoid having to re-pull the blob.

### DIFF
--- a/storage/storage_image.go
+++ b/storage/storage_image.go
@@ -464,15 +464,29 @@ func (s *storageImageDestination) TryReusingBlob(ctx context.Context, blobinfo t
 		}, nil
 	}
 
-	// Does the blob correspond to a known DiffID which we already have available?
-	// Because we must return the size, which is unknown for unavailable compressed blobs, the returned BlobInfo refers to the
-	// uncompressed layer, and that can happen only if canSubstitute.
-	if canSubstitute {
-		if uncompressedDigest := cache.UncompressedDigest(blobinfo.Digest); uncompressedDigest != "" && uncompressedDigest != blobinfo.Digest {
-			layers, err := s.imageRef.transport.store.LayersByUncompressedDigest(uncompressedDigest)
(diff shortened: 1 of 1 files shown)


storage/storage_image.go
```diff
@@ -464,15 +464,29 @@ func (s *storageImageDestination) TryReusingBlob(ctx context.Context, blobinfo t
...
+	if uncompressedDigest := cache.UncompressedDigest(blobinfo.Digest); uncompressedDigest != "" && uncompressedDigest != blobinfo.Digest {
+		layers, err := s.imageRef.transport.store.LayersByUncompressedDigest(uncompressedDigest)
+		if err != nil && errors.Cause(err) != storage.ErrLayerUnknown {
+			return false, types.BlobInfo{}, errors.Wrapf(err, `Error looking for layers with digest %q`, uncompressedDigest)
+		}
+		// If one of the layers with the matching diffID was initialized using a compressed
+		// blob that matches the input, and we recorded the size of that blob, we can report
+		// that digest+size pair.
+		for _, layer := range layers {
+			if layer.CompressedDigest == blobinfo.Digest {
```
mtrmac: How can this ever be reached?  If `layer.CompressedDigest == blobinfo.Digest`, `s.imageRef.transport.store.LayersByCompressedDigest(blobinfo.Digest)` above, around line 453, should have returned a non-empty list.

nalind: Hmm, you're right, that wouldn't give us the size of the compressed version of the layer.  I think we should throw this one out.

PR #575, opened 2019-02-01
Make DefaultPolicyPath a public interface

rhatdan:
We want to tell users to install or edit /etc/config/policy.path
so we need to expose the interface from containers/image.

### DIFF
--- a/signature/policy_config.go
+++ b/signature/policy_config.go
@@ -48,11 +48,11 @@ func (err InvalidPolicyFormatError) Error() string {
 // NOTE: When this function returns an error, report it to the user and abort.
 // DO NOT hard-code fallback policies in your application.
 func DefaultPolicy(sys *types.SystemContext) (*Policy, error) {
-	return NewPolicyFromFile(defaultPolicyPath(sys))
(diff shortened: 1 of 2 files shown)


rhatdan:
@mtrmac @runcom @vrothberg PTAL

mtrmac:
> We want to tell users to install or edit /etc/config/policy.path so we need to expose the interface from containers/image.

Is this talking about an already-existing file (“we’ve used configuration loaded from `$loaded.path`”, in which case the path could be a value reported only after/while loading it), or does it need to also support installing a new one when loading a file fails?

I’d like to make it … not easy … to write independent parsers / interpreters of the policy, and making the default path is one small obstacle to that.  I realize that this is probably untenable long-term, still, worth at least asking.

mtrmac:
If this is about https://github.com/containers/buildah/issues/1307 , couldn’t that just call `signature.DefaultPolicy()` and report if it fails?

rhatdan:
Ok if we call that we should get an error back like:

 open /etc/containers/policy.conf: no such file or directory

Which probably is enough.

PR #576, opened 2019-02-04
Clean up docker/tarfile a bit

mtrmac:
- Move single-method members into the right section
- Use a separate helper which returns err as usual, to centralize the `cacheDataResult` usage and make sure future updates don’t have to deal with less unusual error reporting mechanism.

The second commit is RFC — I do think centralizing the `cacheDataResult` use is worth it, but I’m less certain about making the inner function public, even with a `…Private` name. It could instead be created as a variable inside `ensureCachedDataIsPresent` (`actualImplementation := func() error { … }` or so).

### DIFF
--- a/docker/tarfile/src.go
+++ b/docker/tarfile/src.go
@@ -15,24 +15,24 @@ import (
 	"github.com/containers/image/manifest"
 	"github.com/containers/image/pkg/compression"
 	"github.com/containers/image/types"
-	"github.com/opencontainers/go-digest"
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

rhatdan:
@vrothberg @mtrmac What is the state of this PR?  Should we merge it?

mtrmac:
Yes please. @vrothberg PTAL, **afterwards**.

PR #577, opened 2019-02-06
docker-image-src/getExternalBlob: handle nil resp

arjun024:
makeRequestToResolvedURL() can return {nil, err} for all
given urls. In that case, resp.Body can cause SIGSEGV
because resp is nil.

Signed-off-by: Arjun Sreedharan <arjun024@gmail.com>

### DIFF
--- a/docker/docker_image_src.go
+++ b/docker/docker_image_src.go
@@ -147,10 +147,10 @@ func (s *dockerImageSource) getExternalBlob(ctx context.Context, urls []string) 
 			break
 		}
 	}
-	if resp.Body != nil && err == nil {
-		return resp.Body, getBlobSize(resp), nil
+	if err != nil {
+		return nil, 0, err
 	}
-	return nil, 0, err
+	return resp.Body, getBlobSize(resp), nil
 }
 
 func getBlobSize(resp *http.Response) int64 {


docker/docker_image_src.go
```diff
@@ -147,7 +147,7 @@ func (s *dockerImageSource) getExternalBlob(ctx context.Context, urls []string)
 			break
 		}
 	}
-	if resp.Body != nil && err == nil {
+	if resp != nil && resp.Body != nil && err == nil {
```
mtrmac: Thanks; would a more conventional
```go
if err != nil {
	return nil, 0, err
}
return resp.Body, getBlobSize(resp), nil
```
work just as well? AFAICS `http.Client.Do` always returns a valid `resp.Body` if `err == nil`, so that check was always redundant if any connection was tried — and returning `nil, 0, nil` would not be handled by callers.

Maybe, to be extra careful, add a separate check for
```go
if len(urls) == 0 {
	return errors.New("Internal error: getExternalBlobs called with no URLs")
}
```
at the top, but I don’t think it’s necessary.

arjun024: Yes, that looks better.

mtrmac:
Thanks; LGTM pending tests.

rhatdan:
LGTM

arjun024:
Ok to merge?

mtrmac:
Yes; thanks!

PR #580, opened 2019-02-06
Add exported ValidatePrefix function

umohnani8:
This function calls parseURL to validate the registry input
but discards the normalized registry value. Mainly to be used
by the CRD and controller to validate registry values given
by the user in the container runtime config.

### DIFF
--- a/pkg/sysregistriesv2/system_registries_v2.go
+++ b/pkg/sysregistriesv2/system_registries_v2.go
@@ -84,6 +84,13 @@ func (e *InvalidRegistries) Error() string {
 	return e.s
 }
 
+// ValidatePrefix calls parseURL but discards the normalized registry value returned
+// It just checks that the registry value given as an input is valid
+func ValidatePrefix(input string) error {
+	_, err := parseURL(input)
+	return err
+}
+
 // parseURL parses the input string, performs some sanity checks and returns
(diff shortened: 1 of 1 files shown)


umohnani8:
@mtrmac PTAL

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -84,6 +84,13 @@ func (e *InvalidRegistries) Error() string {
 	return e.s
 }
 
+// ValidatePrefix calls parseURL but discards the normalized registry value returned
```
mtrmac: (Absolutely non-blocking: This is an implementation detail, the public callers don’t care.)

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -84,6 +84,13 @@ func (e *InvalidRegistries) Error() string {
...
+// It just checks that the registry value given as an input is valid
```
mtrmac: This _should_ be more precise, though, so that it is used correctly in callers: it’s a registry[/namespace…[/repo]], valid values for the `Registry.{Prefix,URL}` and `Mirror.URL` fields.

(@vrothberg without a digest / :tag suffix, right?)

umohnani8:
Not really needed, closing.

PR #581, opened 2019-02-08
progress bars: use github.com/vbauerster/mpb

vrothberg:
Use the github.com/vbauerster/mpb library for displaying the progress
bars.  It is actively maintained, is designed for multi-bar usage and
just gets the job done.

This will fix issues with single bars being displayed multiple times.
Note that updating the bar's prefix does not work anymore.  We're now
logging those entries, for instance, when we skip a blob as it's
already present.

The output now looks as follows (e.g., when pulling jboss/wildfly):

```
Copying blob aeb7866da422 [======================================] 71.2 MiB / 71.2 MiB
Copying blob 157601a0b538 [======================================] 10.3 MiB / 10.3 MiB
Copying blob 642f4164f381 [======================================] 1.9 KiB / 1.9 KiB
Copying blob bda512e97517 [======================================] 74.6 MiB / 74.6 MiB
Copying blob 4cccaafdae21 [======================================] 170.9 MiB / 170.9 MiB
Copying config 2602b48525 [======================================] 6.5KiB / 6.5KiB
Storing signatures
```

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -25,9 +25,10 @@ import (
 	"github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	"github.com/vbauerster/mpb"
+	"github.com/vbauerster/mpb/decor"
 	"golang.org/x/crypto/ssh/terminal"
 	"golang.org/x/sync/semaphore"
-	pb "gopkg.in/cheggaaa/pb.v1"
 )
 
 type digestingReader struct {
@@ -90,6 +91,8 @@ type copier struct {
 	reportWriter     io.Writer
 	progressOutput   io.Writer
 	progressInterval time.Duration
+	progressPool     *mpb.Progress
+	progressWG       *sync.WaitGroup
 	progress         chan types.ProgressProperties
 	blobInfoCache    types.BlobInfoCache
 	copyInParallel   bool
@@ -165,12 +168,15 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
 		progressOutput = ioutil.Discard
 	}
 	copyInParallel := dest.HasThreadSafePutBlob() && rawSource.HasThreadSafeGetBlob()
+	wg := new(sync.WaitGroup)
 	c := &copier{
 		dest:             dest,
 		rawSource:        rawSource,
 		reportWriter:     reportWriter,
 		progressOutput:   progressOutput,
 		progressInterval: options.ProgressInterval,
+		progressPool:     mpb.New(mpb.WithWidth(40), mpb.WithOutput(progressOutput), mpb.WithWaitGroup(wg)),
+		progressWG:       wg,
 		progress:         options.Progress,
 		copyInParallel:   copyInParallel,
 		// FIXME? The cache is used for sources and destinations equally, but we only have a SourceCtx and DestinationCtx.
@@ -427,22 +433,6 @@ func shortDigest(d digest.Digest) string {
 	return d.Encoded()[:12]
 }
 
-// createProgressBar creates a pb.ProgressBar.  Note that if the copier's
-// reportWriter is ioutil.Discard, the progress bar's output will be discarded
-// and a single line will be printed instead.
-func (c *copier) createProgressBar(srcInfo types.BlobInfo, kind string) *pb.ProgressBar {
-	bar := pb.New(int(srcInfo.Size)).SetUnits(pb.U_BYTES)
-	bar.SetMaxWidth(80)
-	bar.ShowTimeLeft = false
-	bar.ShowPercent = false
-	bar.Prefix(fmt.Sprintf("Copying %s %s:", kind, shortDigest(srcInfo.Digest)))
-	bar.Output = c.progressOutput
-	if bar.Output == ioutil.Discard {
-		c.Printf("Copying %s %s\n", kind, srcInfo.Digest)
-	}
-	return bar
-}
-
 // isTTY returns true if the io.Writer is a file and a tty.
 func isTTY(w io.Writer) bool {
 	if f, ok := w.(*os.File); ok {
@@ -477,6 +467,7 @@ func (ic *imageCopier) copyLayers(ctx context.Context) error {
 	// copyGroup is used to determine if all layers are copied
 	copyGroup := sync.WaitGroup{}
 	copyGroup.Add(numLayers)
+
 	// copySemaphore is used to limit the number of parallel downloads to
 	// avoid malicious images causing troubles and to be nice to servers.
 	var copySemaphore *semaphore.Weighted
@@ -487,8 +478,7 @@ func (ic *imageCopier) copyLayers(ctx context.Context) error {
 	}
 
 	data := make([]copyLayerData, numLayers)
-	copyLayerHelper := func(index int, srcLayer types.BlobInfo, bar *pb.ProgressBar) {
-		defer bar.Finish()
+	copyLayerHelper := func(index int, srcLayer types.BlobInfo, bar *mpb.Bar) {
 		defer copySemaphore.Release(1)
 		defer copyGroup.Done()
 		cld := copyLayerData{}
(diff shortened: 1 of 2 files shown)


vrothberg:
@rhatdan @mtrmac @baude @mheon This will fix the messed up progress bars in Podman master. The new library makes a very good impression on me and most importantly, it works :^)

vrothberg:
I'll spin up a Skopeo CI PR, where you can also test.

rhatdan:
Works for me, but need the skopeo change.

vrothberg:
> Works for me, but need the skopeo change.

Had two deadlocks in the code which are now fixed and Skopeo is green (see https://github.com/containers/skopeo/pull/602).
@rhatdan @mtrmac @runcom PTAL

rhatdan:
LGTM

vrothberg:
(updated to latest master)

copy/copy.go
```diff
@@ -615,9 +604,24 @@ func (c *copier) copyConfig(ctx context.Context, src types.Image) error {
...
+			mpb.PrependDecorators(
+				decor.Name(fmt.Sprintf("Copying config %s", name)),
+			),
+			mpb.AppendDecorators(
+				decor.CountersKibiByte("%.1f / %.1f"),
+			),
+		)
+		if c.progressOutput == ioutil.Discard {
+			c.Printf("Copying config %s\n", srcInfo.Digest)
+		}
```
mtrmac: A quick initial impression: The mechanics of the setup are not entirely trivial, sharing them in a helper (very roughly similar to the `createProgressBar` one) might be more readable.

vrothberg: https://github.com/containers/skopeo/pull/602 is green with the updated PR.

giuseppe:
I've left a comment in the Skopeo's PR, I am seeing a problem when I try to pull `jboss/wildfly`

giuseppe:
LGTM

copy/copy.go
```diff
@@ -615,13 +613,16 @@ func (c *copier) copyConfig(ctx context.Context, src types.Image) error {
...
-		bar.Start()
+
+		// make the short digest only 10 characters long to make it align with the blob output
+		bar := c.createProgressBar(srcInfo, shortDigest(srcInfo.Digest)[0:10], "config")
 		destInfo, err := c.copyBlobFromStream(ctx, bytes.NewReader(configBlob), srcInfo, nil, false, true, bar)
 		if err != nil {
 			return err
 		}
+		if bar != nil {
+			bar.SetTotal(0, true)
```
mtrmac: Should this be 0 or `len(configBlob)`?

copy/copy.go
```diff
@@ -753,7 +751,9 @@ func computeDiffID(stream io.Reader, decompressor compression.DecompressorFunc)
 // and returns a complete blobInfo of the copied blob.
 func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, srcInfo types.BlobInfo,
 	getOriginalLayerCopyWriter func(decompressor compression.DecompressorFunc) io.Writer,
-	canModifyBlob bool, isConfig bool, bar *pb.ProgressBar) (types.BlobInfo, error) {
+	canModifyBlob bool, isConfig bool, bar *mpb.Bar) (types.BlobInfo, error) {
+	c.progressWG.Add(1)
+	defer c.progressWG.Done()
```
mtrmac: Does `c.progressWG` do anything we need? The synchronization of parallel layer copy AFAICS happens independently via `copyGroup`, and the only effect of `mpb.WithWaitGroups` seems to be that `ic.c.progressPool.Wait()` blocks on `c.progressWG` in addition to doing everything else it does.

copy/copy.go
```diff
@@ -600,13 +580,31 @@ func (ic *imageCopier) copyUpdatedConfigAndManifest(ctx context.Context) ([]byte
 		return nil, err
 	}
 
+	ic.c.progressPool.Wait()
```
mtrmac: The documentation of `Progress.Wait` says
> After this method has been called, there is no way to reuse *Progress instance.

The middle of `copyUpdatedConfigAndManifest` is not a good place for such a call, because `copyUpdatedConfigAndManifest` can be called multiple times per image (as well, as, strictly speaking, zero times if an error happens before).  Either the pool should live across all of those calls (and in that case the “Writing manifest to image destination” message just below needs to happen in some way that does not interfere with the `progressPool`), or it should be closed before even starting to deal with configs/manifests (using just a message, not progress bar, for the config, or maybe creating an one-use single-bar progress pool just for each attempt to copy a manifest).

(Similar in nature is a concern about attaching `progressPool` to `copier` vs. `imageCopier`; right now that makes ~zero difference; longer-term after #400 is resurrected, `copier` might live across copies of multiple single-arch images, whereas `imageCopier` would carry state related to one of them.)

(Purely aesthethically it would be nice for the “closing” of the progress pool to happen (or, to be guaranteed to happen as a last-resort backstop) at the same level of the call stack that creates the progress pool.)

copy/copy.go
```diff
@@ -498,32 +488,22 @@ func (ic *imageCopier) copyLayers(ctx context.Context) error {
...
 				logrus.Debugf("Skipping foreign layer %q copy to %s", cld.destInfo.Digest, ic.c.dest.Reference().Transport().Name())
-				bar.Prefix(fmt.Sprintf("Skipping blob %s (foreign layer):", shortDigest(srcLayer.Digest)))
-				bar.Add64(bar.Total)
-				bar.Finish()
 			}
 		} else {
 			cld.destInfo, cld.diffID, cld.err = ic.copyLayer(ctx, srcLayer, bar)
 		}
 		data[index] = cld
+		if bar != nil {
```
mtrmac: `bar` can never be `nil` AFAICS, so drop this check. (The code would crash in `bar.ProxyReader` anyway.)

copy/copy.go
```diff
@@ -615,13 +613,16 @@ func (c *copier) copyConfig(ctx context.Context, src types.Image) error {
...
-		defer bar.Finish()
-		bar.Start()
+
+		// make the short digest only 10 characters long to make it align with the blob output
+		bar := c.createProgressBar(srcInfo, shortDigest(srcInfo.Digest)[0:10], "config")
 		destInfo, err := c.copyBlobFromStream(ctx, bytes.NewReader(configBlob), srcInfo, nil, false, true, bar)
 		if err != nil {
 			return err
 		}
+		if bar != nil {
```
mtrmac: `bar` can never be `nil` AFAICS, so drop this check. (The code would crash in `bar.ProxyReader` anyway.)

copy/copy.go
vrothberg: 0 is intended. It has no impact on the displayed total size but it will trigger the bar to be completed.

copy/copy.go
vrothberg: Again, that's intended. We need to make sure the progress bar has finished rendering before printing, see the next line `ic.c.Printf("Writing manifest to image destination\n")`. I understand it's not ideal when looking at the assumptions on the code, but I could not find any other way and am very hesitant to rewrite `copy.go`.

vrothberg: > (Similar in nature is a concern about attaching `progressPool` to `copier` vs. `imageCopier`; right now that makes ~zero difference; longer-term after #400 is resurrected, `copier` might live across copies of multiple single-arch images, whereas `imageCopier` would carry state related to one of them.)

It's possible to add new bars dynamically and it will render correctly as long as nothing prints in between.

copy/copy.go
vrothberg: There was one test where exactly this case happened while copying a config. It's really just a defense for corner cases.

copy/copy.go
vrothberg: There was one test case where this happened. I could not figure out why and also had a look at the mpb code but figured to just add the check to defend from the corner case. I think it's related to non-TTY as the test case (i.e., `TestCopyOCIRoundTrip`) worked when doing it manually.

copy/copy.go
vrothberg: That's intended as well. While testing this PR, we've experienced situations where the progress was still rendering although the copy operations have completed (including the one of the config). It's a background process with a specific refresh rate and the copier's WaitGroup makes sure that we progress only when it's done.

vrothberg:
> I’m sorry about getting to this so late. I can do the smaller cleanups myself.

All your comments are valid concerns, but I couldn't find any other way without refactoring parts of `copy.go`.

> @vrothberg PTAL at least at the `Progress.Wait` comment — am I missing anything?

You missed a lot of pain :^) The progress bars are always "rendered" in the background with a specific refresh rate, which makes them subject to race conditions with the code. This is why the "old" library had issues and mpb avoids such situations by integration a `sync.WaitGroup`. The place where we do this is constrained by the calls where we're printing to the reportWriter. Any print will cause issues for the progress bars.

copy/copy.go
vrothberg: > (Purely aesthethically it would be nice for the “closing” of the progress pool to happen (or, to be guaranteed to happen as a last-resort backstop) at the same level of the call stack that creates the progress pool.)

I share that desire but mpb doesn't expose a `Nuke()` API. It is possible though to nuke each bar indivually which would allow to work around that.

copy/copy.go
mtrmac: > and the copier's WaitGroup makes sure that we progress only when it's done.

I can’t see how that works; the _only_ use of `progressWG` inside the code is `p.uwg.Wait()` immediately at the start of `Progress.Wait`, which AFAICS is immediately satisfied because `copyGroup.Wait` ensured that all layer copies are done, and `copyConfig` has returned.

`Progress.Wait` does wait for the separate render process to finish, via `pState.zeroWait` (“please exit”) and `Progress.done` (“exited”); I can’t see that this depends on `progressWG` in any way.

copy/copy.go
mtrmac: Second, working around crashes without understanding how they happen is not great for reliability. AFAICS the only way `bar` can be `nil` is if `AddBar` returns `nil`, which in turn can happen only if `Progress.Wait` has been called.

… which probably points to 
https://github.com/containers/image/pull/581#discussion_r257475486 , calling `Wait` and looping to `copyUpdatedConfigAndManifest` as the underlying cause: the code tried schema2, failed, tried schema1, and `copyConfig` did not use the `bar` (and did not trigger the `bar.ProgressReader` crash) because schema1 does not have a separate config object.

I’ll take a stab at fixing this; if you can remember what the failing situation was (or is it just straightforward `make test-skopeo` failing?), that would be useful.

vrothberg: You're absolutely right, and I was/am not happy with it either. The failure is straightforward to reproduce with integration tests (i.e., `TestCopyOCIRoundTrip`).

I think a more or less easy solution is to break out the parts printing (including related code) and move the Wait() call to the end of copyOneImage.

mtrmac: Thanks, reproduced. I was not quite right about this happening with schema1 fallback:it is copying an OCI image and falling back to schema2.  For the record, the reason it does not crash in `bar.ProgressReader` is that creating the `ProgressReader` does not access `bar`, only non-empty reads do, and `PutBlob` detects that the config is already present at the destination, so no reads happen.

I think I know how to deal with this…

vrothberg: > I think I know how to deal with this…

Thank you so much for looking into this :pray:

PR #582, opened 2019-02-08
WIP DO NOT MERGE: Support multiple manifests in c/storage

mtrmac:
WIP, do not merge: Right now only to have a fallback implementation and to get it through CI.

When writing an image, always record the manifest in a `manifest-$digest` item, in addition to the existing `manifest` item.

To make this consistent, fail in `PutManifest` if the destination reference includes a digest and the manifest does not match.

When reading an image, return the manifest corresponding to the user-specified digest (if such a manifest has been recorded), falling back to the existing use of `manifest` (i.e. the latest one written).

### DIFF
--- a/storage/storage_image.go
+++ b/storage/storage_image.go
@@ -14,6 +14,7 @@ import (
 	"sync"
 	"sync/atomic"
 
+	"github.com/containers/image/docker/reference"
 	"github.com/containers/image/image"
 	"github.com/containers/image/internal/tmpdir"
 	"github.com/containers/image/manifest"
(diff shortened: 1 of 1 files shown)


mtrmac:
Note to self: #573 is also necessary because `storageImageDestionation.TryReusingBlob` can cause the manifest to be edited.

nalind:
LGTM.  #585 includes this patch, so we can either close this one, or rebase and merge this one, after which I can rebase #585.

mtrmac:
Let’s discuss this in #585.

PR #583, opened 2019-02-09
Try harder in storageImageDestination.TryReusingBlob

mtrmac:
Because a `storage.Layer` only records a single `CompressedDigest` and `CompressedSize`, we may need to depend on the `BlobInfoCache`.

In that case, `TryReusingBlob` currently returns correct data about the uncompressed blob, causing the manifest ultimately sent to `PutManifest` to be updated, and potentially breaking digest references, or disabling the reuse if manifest updates are forbidden.

The extra information necessary to avoid the manifest update is just the compressed layer size, and with non-schema1 manifests that size is a part of the manifest and provided by the caller; so, we can
just return the caller's `BlobInfo` unmodified.

To make this work, `Commit` also needs to be updated to call `LayersByUncompressedDigest` on a `diffID` recorded in `blobDiffIDs` because the kinds of layers we are discussing here can't be found by `LayersByCompressedDigest`.

### DIFF
--- a/storage/storage_image.go
+++ b/storage/storage_image.go
@@ -491,14 +491,21 @@ func (s *storageImageDestination) TryReusingBlob(ctx context.Context, blobinfo t
 
 	// Does the blob correspond to a known DiffID which we already have available?
 	// Because we must return the size, which is unknown for unavailable compressed blobs, the returned BlobInfo refers to the
-	// uncompressed layer, and that can happen only if canSubstitute.
-	if canSubstitute {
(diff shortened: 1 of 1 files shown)


mtrmac:
@nalind RFC; #573 sets `!canUpdateManifest` in more cases, reducing the blob reuse opportunities on pull, but this should restore the previous possibilities for schema2 and OCI manifests.

rhatdan:
@nalind @mtrmac @vrothberg What is the latest on this PR, Can we get it rebased and merged?

mtrmac:
@nalind (@vrothberg ) PTAL.

PR #584, opened 2019-02-11
Fallback to v2s2 if OCI manifest upload fails on ECR

markmc:
Uploading a spec-compliant OCI manifest which contains no top-level media type fails on AWS ECR with an Unsupported error code and  an 'Invalid JSON syntax' error message.

As suggested in containers/libpod#1719 catch this case as an invalid manifest error so that we automatically fallback to other accepted media types in copyOneImage().


mtrmac:
LGTM. Thanks!

PR #585, opened 2019-02-11
Storage multiple manifests

nalind:
Now that containers/storage can handle storing multiple manifests for a single image record, name manifests that we save to an image using their digests, repurposing the original name that we used for the manifest (when it only supported holding one) to hold a copy of the most-recently saved one.  This is done by directly incorporating the patch from #582.

When we're asked for a specific manifest by including its digest in an image reference, look for the manifest that matches that digest and set the returned Image's Digest field to a matching value.  Otherwise, try to return the most-recently-saved manifest's digest in the field.

### DIFF
--- a/storage/storage_image.go
+++ b/storage/storage_image.go
@@ -686,15 +686,9 @@ func (s *storageImageDestination) Commit(ctx context.Context) error {
 		lastLayer = layer.ID
 	}
 
-	manifestDigest, err := manifest.Digest(s.manifest)
-	if err != nil {
-		return errors.Wrapf(err, "error computing manifest digest")
-	}
 	// If one of those blobs was a configuration blob, then we can try to dig out the date when the image
 	// was originally created, in case we're just copying it.  If not, no harm done.
-	options := &storage.ImageOptions{
-		Digest: manifestDigest,
-	}
+	options := &storage.ImageOptions{}
 	if inspect, err := man.Inspect(s.getConfigBlob); err == nil && inspect.Created != nil {
 		logrus.Debugf("setting image creation date to %s", inspect.Created)
 		options.CreationDate = *inspect.Created
@@ -767,6 +761,10 @@ func (s *storageImageDestination) Commit(ctx context.Context) error {
 	// Save the manifest.  Allow looking it up by digest by using the key convention defined by the Store.
 	// Record the manifest twice: using a digest-specific key to allow references to that specific digest instance,
 	// and using storage.ImageDigestBigDataKey for future users that don’t specify any digest and for compatibility with older readers.
+	manifestDigest, err := manifest.Digest(s.manifest)
+	if err != nil {
+		return errors.Wrapf(err, "error computing manifest digest")
+	}
 	if err := s.imageRef.transport.store.SetImageBigData(img.ID, manifestBigDataKey(manifestDigest), s.manifest); err != nil {
 		if _, err2 := s.imageRef.transport.store.DeleteImage(img.ID, true); err2 != nil {
 			logrus.Debugf("error deleting incomplete image %q: %v", img.ID, err2)
--- a/storage/storage_image.go
+++ b/storage/storage_image.go
@@ -14,6 +14,7 @@ import (
 	"sync"
 	"sync/atomic"
 
+	"github.com/containers/image/docker/reference"
 	"github.com/containers/image/image"
 	"github.com/containers/image/internal/tmpdir"
 	"github.com/containers/image/manifest"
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM
@mtrmac @vrothberg PTAL

storage/storage_reference.go
```diff
@@ -97,6 +104,24 @@ func (s *storageReference) resolveImage() (*storage.Image, error) {
 			return nil, ErrNoSuchImage
 		}
 	}
+	// Default to having the image digest that we hand back match the most recently
+	// added manifest...
+	if digest, ok := loadedImage.BigDataDigests[storage.ImageDigestBigDataKey]; ok {
+		loadedImage.Digest = digest
```
mtrmac: What benefits from the `store.Image.Digest` assignment? I can see no user in c/image.

Is this for external users via `GetStoreImage`? I can find
- `RepoTags` computation from names + single digest (conceptually incorrect anyway) in libpod and CRI-O
- `Digest` field included in libpod’s “inspect” functionality, which is only used in `podman inspect` JSON output.

I can see how using the right `Digest` is conceptually nice, OTOH isn’t this the very first case where `store.Image` contents differ from the c/storage backing store? If it is indeed the first case, doing this seems fairly risky. Or are such edits already happening regularly?

Am I missing some other user that benefits?

Or is this primarily motivated by wanting not to record the `Digest` field in the storage (per the `Commit` change), making this read-side edit necessary so that current users have _some_ value in there?

storage/storage_reference.go
```diff
@@ -97,6 +104,24 @@ func (s *storageReference) resolveImage() (*storage.Image, error) {
...
+	// sake of older consumers that don't know there's a whole list in there now.
+	if s.digest != "" {
+		for _, digest := range loadedImage.Digests {
+			if digest == s.digest {
+				loadedImage.Digest = digest
+				break
+			}
+		}
+	} else {
+		s.digest = loadedImage.Digest
```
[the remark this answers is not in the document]
nalind: Okay, I'll drop the field.

storage/storage_reference.go
nalind: The storage library will populate the `Digest` field with a value if we don't supply one when we go to create an image record, and since it's digesting the manifest using our function, we don't need to specify a value that'd be different from the one it used to compute using the `digest` package.
In the newer version of storage, we make sure that the field contains one of the values from `Digests` (which is not itself stored, but is built from the stored `Digest` value and `BigDataDigests` map), but it doesn't have a rule for which of multiple values that'll be, since it treats them all as equally valid.  In practice, it's either the value that we set when creating the image record, or the digest of the first one that came up when traversing the map.
Having `Digest` match the data item named `manifest`, which is what we hand back from `GetManifest()` when the named reference doesn't include a digest, keeps things consistent for current users who locate the record by way of the reference, which is a behavior that I'm pretty sure current CRI-O implicitly expects.

mtrmac: > The storage library will populate the `Digest` field with a value if we don't supply one when we go to create an image record

(If you are talking about `CreateImage`, it can’t, because the “big data” entry for the manifest does not exist at that time. The two `SetImageBigData` calls soon after do cause `.Digest` to be set, though.)

> In the newer version of storage, we make sure that the field contains one of the values from `Digests` (which is not itself stored, but is built from the stored `Digest` value and `BigDataDigests` map), but it doesn't have a rule for which of multiple values that'll be, since it treats them all as equally valid. In practice, it's either the value that we set when creating the image record, or the digest of the first one that came up when traversing the map.

AFAICS it’s either the value set when creating the image, or the first manifest recorded after creating the image, because `SetBigData` calls `recomputeDigests`, populating `Digest`, which is then saved into the storage. So, subsequent changes to `BigDataDigests`, or changes in the order of enumerating the map, should not make a difference.

> Having `Digest` match the data item named `manifest`, which is what we hand back from `GetManifest()` when the named reference doesn't include a digest, keeps things consistent for current users who locate the record by way of the reference, which is a behavior that I'm pretty sure current CRI-O implicitly expects.

Not AFAICS; `GetManifest()` must match the digest in the reference, if any for `copy.Image` to work, but that’s not what this does.

CRI, and the kubelet, does expect that `RepoDigests[0]` returned from `ImageStatus` (which is computed by combining names and `.Digest`) can be used to refer back to the image, but `resolveImage` matches digested references loosely enough (name separately and digest separately) that this should always work. Arguably we should tighten that, eventually — but  we would probably fix `RepoDigests` computation in CRI-O first, and then tighten the lookup.

---

Ultimately, \*shrug\* This does worry me, but nothing in c/image depends on the value, so…

storage/storage_image.go
```diff
@@ -660,16 +685,18 @@ func (s *storageImageDestination) Commit(ctx context.Context) error {
...
 	// If one of those blobs was a configuration blob, then we can try to dig out the date when the image
 	// was originally created, in case we're just copying it.  If not, no harm done.
 	options := &storage.ImageOptions{}
 	if inspect, err := man.Inspect(s.getConfigBlob); err == nil && inspect.Created != nil {
 		logrus.Debugf("setting image creation date to %s", inspect.Created)
 		options.CreationDate = *inspect.Created
 	}
-	if manifestDigest, err := manifest.Digest(s.manifest); err == nil {
-		options.Digest = manifestDigest
-	}
```
mtrmac: Tiny nit: Now that `manifestDigest` is not necessary here, please move the computation of the value right before its use.

nalind: Moved it.  Thanks!

nalind:
Rebased.

nalind:
Merging.

PR #586, opened 2019-02-18
Fix progress bar cleanup, and other small cleanups

mtrmac:
Primarily, this fixes the `.Wait` call in `copyUpdatedConfigAndManifest`, terminating the pool before all its possible uses, by using several shorter-lived pools.  Also ensures that the pools are always cleaned up, and other small cleanups.

See the individual commit descriptions for details.

@vrothberg PTAL.

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -495,7 +495,7 @@ func (ic *imageCopier) copyLayers(ctx context.Context) error {
 
 	progressBars := make([]*mpb.Bar, numLayers)
 	for i, srcInfo := range srcInfos {
-		progressBars[i] = ic.c.createProgressBar(srcInfo, "blob")
+		progressBars[i] = ic.c.createProgressBar(ic.c.progressPool, srcInfo, "blob")
 	}
 
 	for i, srcLayer := range srcInfos {
@@ -580,9 +580,9 @@ func (ic *imageCopier) copyUpdatedConfigAndManifest(ctx context.Context) ([]byte
 	return manifest, nil
 }
 
-// createProgressBar creates a mpb.Bar.  Note that if the copier's reportWriter
+// createProgressBar creates a mpb.Bar in pool.  Note that if the copier's reportWriter
 // is ioutil.Discard, the progress bar's output will be discarded
-func (c *copier) createProgressBar(info types.BlobInfo, kind string) *mpb.Bar {
+func (c *copier) createProgressBar(pool *mpb.Progress, info types.BlobInfo, kind string) *mpb.Bar {
 	// shortDigestLen is the length of the digest used for blobs.
 	const shortDigestLen = 12
 
@@ -593,7 +593,7 @@ func (c *copier) createProgressBar(info types.BlobInfo, kind string) *mpb.Bar {
 		prefix = prefix[:maxPrefixLen]
 	}
 
-	bar := c.progressPool.AddBar(info.Size,
+	bar := pool.AddBar(info.Size,
 		mpb.PrependDecorators(
 			decor.Name(prefix),
 		),
@@ -616,7 +616,7 @@ func (c *copier) copyConfig(ctx context.Context, src types.Image) error {
 			return errors.Wrapf(err, "Error reading config blob %s", srcInfo.Digest)
 		}
 
-		bar := c.createProgressBar(srcInfo, "config")
+		bar := c.createProgressBar(c.progressPool, srcInfo, "config")
 		destInfo, err := c.copyBlobFromStream(ctx, bytes.NewReader(configBlob), srcInfo, nil, false, true, bar)
 		if err != nil {
 			return err
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -22,7 +22,7 @@ import (
 	"github.com/containers/image/transports"
 	"github.com/containers/image/types"
 	"github.com/klauspost/pgzip"
-	"github.com/opencontainers/go-digest"
+	digest "github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/vbauerster/mpb"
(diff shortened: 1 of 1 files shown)


mtrmac:
https://github.com/containers/skopeo/pull/605 if you want to test this in practice.

copy/copy.go
```diff
@@ -580,20 +572,40 @@ func (ic *imageCopier) copyUpdatedConfigAndManifest(ctx context.Context) ([]byte
...
 // is ioutil.Discard, the progress bar's output will be discarded
-func (c *copier) createProgressBar(info types.BlobInfo, name, kind string) *mpb.Bar {
-	bar := c.progressPool.AddBar(info.Size,
+func (c *copier) createProgressBar(pool *mpb.Progress, info types.BlobInfo, kind string) *mpb.Bar {
+	// shortDigestLen is the length of the digest used for blobs.
+	const shortDigestLen = 12
+
+	prefix := fmt.Sprintf("Copying %s %s", kind, info.Digest.Encoded())
+	// Truncate the prefix (chopping of some part of the digest) to make all progress bars aligned in a column.
+	maxPrefixLen := len("Copying blob ") + shortDigestLen
```
vrothberg: Nit: this could be made a const outside the function to avoid redundantly calculating it.

copy/copy.go
```diff
@@ -496,27 +486,29 @@ func (ic *imageCopier) copyLayers(ctx context.Context) error {
...
+		}
 
-	// Wait for all layers to be copied
-	copyGroup.Wait()
+		// Wait for all layers to be copied
+		copyGroup.Wait()
+	}()
 
+	destInfos := make([]types.BlobInfo, numLayers)
+	diffIDs := make([]digest.Digest, numLayers)
```
vrothberg: Nit: the intention behind moving those before `copyGroup.Wait()` was that the allocation can happen while we're still waiting for the copy-goroutines to finish. Maybe, I am missing something but is the `defer progressCleanup()` technically required or just means to make the code more readable?

I *think* that the following would be slightly faster:
```golang
	destInfos := make([]types.BlobInfo, numLayers)
	diffIDs := make([]digest.Digest, numLayers)
        copyGroup.Wait()
        progressCleanup()
```

copy/copy.go
```diff
@@ -580,20 +572,40 @@ func (ic *imageCopier) copyUpdatedConfigAndManifest(ctx context.Context) ([]byte
...
-// createProgressBar creates a mpb.Bar.  Note that if the copier's reportWriter
+// newProgressPool creates a *mpb.Progress and a cleanup function.
+// The caller must eventually call the returned cleanup function after the pool will no longer be updated.
+func (c *copier) newProgressPool(ctx context.Context) (*mpb.Progress, func()) {
+	ctx, cancel := context.WithCancel(ctx)
+	pool := mpb.New(mpb.WithWidth(40), mpb.WithOutput(c.progressOutput), mpb.WithContext(ctx))
+	return pool, func() {
+		cancel()
+		pool.Wait()
+	}
```
vrothberg: Very elegant! I've tried that as well but failed to make it work properly. Thanks a lot for cleaning up my mess!

copy/copy.go
mtrmac: OTOH that would move it far from the "Copying blob" string and make it easier for the two to get out of sync.

Or this could just hard-code `13` instead of `len("Copying blob ")`, but even with a comment that’s not really worth the readability impact.

This is to an extent already overthinking the previous `shortDigest()[:10]` implementation, and I’d prefer not to overthink it that much.

---

Ultimately/luckily, the compiler can already optimize `len("Copying blob ") + shortDigestLen` to `25`, so this `len` calls costs exactly zero:
```go
func prefix(digest digest.Digest, kind string) string {
	// shortDigestLen is the length of the digest used for blobs.
	const shortDigestLen = 12

	prefix := fmt.Sprintf("Copying %s %s", kind, digest.Encoded())
	// Truncate the prefix (chopping of some part of the digest) to make all progress bars aligned in a column.
	maxPrefixLen := len("Copying blob ") + shortDigestLen
	if len(prefix) > maxPrefixLen {
		prefix = prefix[:maxPrefixLen]
	}
	return prefix
}
```
(edited to remove most addresses and other garbage from the output):
```console
$ go run -gcflags '-S' foo.go
…
	CALL	fmt.Sprintf(SB)
	MOVQ	40(SP), AX  # return value of `Sprintf` == prefix, string start
	MOVQ	48(SP), CX  # return value of `Sprintf` == prefix, string length == len(prefix)
	CMPQ	CX, $25  # len(prefix) > maxPrefixLen
	JLE	291
	MOVL	$25, CX   # prefix = prefix[:maxPrefixLen]
291: 	MOVQ	AX, "".~r2+168(SP)  # returned string start
	MOVQ	CX, "".~r2+176(SP)  # returned string length
```

copy/copy.go
mtrmac: > Nit: the intention behind moving those before `copyGroup.Wait()` was that the allocation can happen while we're still waiting for the copy-goroutines to finish.

That’s thoughtful and clever, but we are going to be blocked (many times) on disk and possibly network, is moving a few memory allocations to run concurrently with that at all noticeable, and worth the readability impact? In this case it would mean two extra return values from the `func() { // A scope for defer`, or declaring the variables (as `nil` pointers) before that function starts, the function, adding two more links and stretching the scope of the variables much larger.

To me, computers seem fast enough that I’d much rather have short easy-to read code than long clever code, unless it is a known tight loop critical for the product (which wouldn’t be written in Go at all, most likely).


> Maybe, I am missing something but is the `defer progressCleanup()` technically required or just means to make the code more readable?

It also ensures that the progress bar is properly cleaned up in case of a `panic` (right now), which is admittedly pretty unlikely to happen — and also if any future updates add a `return err` somewhere within the scope of `progressPool`, which is much more likely (it already happens in `copyConfig`). So, it contributes to longer-term reliability.

---

Up to you, I can change it back if you prefer the speed-up.

copy/copy.go
mtrmac: (/me longingly looks at Python’s `with`, and Swift’s trailing closure syntax + templates + type inference  that would allow transparently emulating `with` while remaining readable; Go really is the BASIC of this decade.)

copy/copy.go
vrothberg: Makes perfect sense :) Thanks!

copy/copy.go
vrothberg: > Up to you, I can change it back if you prefer the speed-up.

I am fine with this code. The maintainability overrules the few cycles we may safe.

vrothberg:
LGTM, thanks for the great improvements, @mtrmac! Let's get it merged and cut off a new release.

rhatdan:
LGTM

PR #587, opened 2019-02-20
ostree: default is no OStree support

### LINKED ISSUE #432: Make libselinux optional dependency

Silvanoc:
Since only Skopeo uses the OStree transport for fetching images from
OStree repositories, making default no support for OStree. It
simplifies building on other components like libpod and buildah.

This PR is completely different approach to address #432 motivated by [this](https://github.com/containers/image/pull/562#issuecomment-458071484) and [this](https://github.com/containers/image/pull/562#issuecomment-458177678) comments of @giuseppe 

It will require changes in Skopeo, Buildah, CRI-O and libpod to have support for the c/image OStree backend only for Skopeo.

Closes #432.

Signed-off-by: Silvano Cirujano Cuesta <silvano.cirujano-cuesta@siemens.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,10 +62,8 @@ or use the build tags described below to avoid the dependencies (e.g. using `go 
 
 - `containers_image_openpgp`: Use a Golang-only OpenPGP implementation for signature verification instead of the default cgo/gpgme-based implementation;
 the primary downside is that creating new signatures with the Golang-only implementation is not supported.
-- `containers_image_ostree`: Import `ostree:` transport in `github.com/containers/image/transports/alltransports` to download images from an OStree repository.
-This builds the library requiring the `libostree` development libraries.
-Otherwise a stub which reports that the transport is not supported gets used.
-The `github.com/containers/image/ostree` package is completely disabled and impossible to import when this build tag is not in use.
+- `containers_image_ostree`: Import `ostree:` transport in `github.com/containers/image/transports/alltransports`. This builds the library requiring the `libostree` development libraries. Otherwise a stub which reports that the transport is not supported gets used. The `github.com/containers/image/ostree` package is completely disabled
+and impossible to import when this build tag is not in use.
 
 ## [Contributing](CONTRIBUTING.md)
 
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ or use the build tags described below to avoid the dependencies (e.g. using `go 
 
 - `containers_image_openpgp`: Use a Golang-only OpenPGP implementation for signature verification instead of the default cgo/gpgme-based implementation;
 the primary downside is that creating new signatures with the Golang-only implementation is not supported.
-- `ostree_repos`: Import `ostree:` transport in `github.com/containers/image/transports/alltransports` to download images from an OStree repository.
+- `containers_image_ostree`: Import `ostree:` transport in `github.com/containers/image/transports/alltransports` to download images from an OStree repository.
 This builds the library requiring the `libostree` development libraries.
 Otherwise a stub which reports that the transport is not supported gets used.
 The `github.com/containers/image/ostree` package is completely disabled and impossible to import when this build tag is not in use.
--- a/README.md
+++ b/README.md
@@ -62,8 +62,10 @@ or use the build tags described below to avoid the dependencies (e.g. using `go 
 
 - `containers_image_openpgp`: Use a Golang-only OpenPGP implementation for signature verification instead of the default cgo/gpgme-based implementation;
 the primary downside is that creating new signatures with the Golang-only implementation is not supported.
-- `containers_image_ostree_stub`: Instead of importing `ostree:` transport in `github.com/containers/image/transports/alltransports`, use a stub which reports that the transport is not supported. This allows building the library without requiring the `libostree` development libraries. The `github.com/containers/image/ostree` package is completely disabled
-and impossible to import when this build tag is in use.
+- `ostree_repos`: Import `ostree:` transport in `github.com/containers/image/transports/alltransports` to download images from an OStree repository.
+This builds the library requiring the `libostree` development libraries.
+Otherwise a stub which reports that the transport is not supported gets used.
+The `github.com/containers/image/ostree` package is completely disabled and impossible to import when this build tag is not in use.
 
 ## [Contributing](CONTRIBUTING.md)
 
--- a/Makefile
+++ b/Makefile
@@ -7,17 +7,10 @@ SKOPEO_BRANCH = master
 SUDO =
 
 # when cross compiling _for_ a Darwin or windows host, then we must use openpgp
-BUILD_TAGS_WINDOWS_CROSS = containers_image_ostree_stub containers_image_openpgp
-BUILD_TAGS_DARWIN_CROSS = containers_image_ostree_stub containers_image_openpgp
-# when compiling _on_ a Darwin host, then we can link against gpgme
-BUILD_TAGS_DARWIN_NATIVE = containers_image_ostree_stub
+BUILD_TAGS_WINDOWS_CROSS = containers_image_openpgp
+BUILD_TAGS_DARWIN_CROSS = containers_image_openpgp
 
-ifeq ($(shell uname),Darwin)
-PLATFORM_BUILD_TAG = $(BUILD_TAGS_DARWIN_NATIVE)
-endif
-
-
-BUILDTAGS = btrfs_noversion libdm_no_deferred_remove $(PLATFORM_BUILD_TAG)
+BUILDTAGS = btrfs_noversion libdm_no_deferred_remove
 BUILDFLAGS := -tags "$(BUILDTAGS)"
 
 PACKAGES := $(shell go list $(BUILDFLAGS) ./... | grep -v github.com/containers/image/vendor)
--- a/ostree/ostree_dest.go
+++ b/ostree/ostree_dest.go
@@ -1,4 +1,4 @@
-// +build !containers_image_ostree_stub
+// +build ostree_repos
 
 package ostree
 
--- a/ostree/ostree_dest.go
+++ b/ostree/ostree_dest.go
@@ -1,4 +1,4 @@
-// +build ostree_repos
+// +build containers_image_ostree
 
 package ostree
 
--- a/ostree/ostree_src.go
+++ b/ostree/ostree_src.go
@@ -1,4 +1,4 @@
-// +build !containers_image_ostree_stub
+// +build ostree_repos
 
 package ostree
 
--- a/ostree/ostree_src.go
+++ b/ostree/ostree_src.go
@@ -1,4 +1,4 @@
-// +build ostree_repos
+// +build containers_image_ostree
 
 package ostree
 
--- a/ostree/ostree_transport.go
+++ b/ostree/ostree_transport.go
@@ -1,4 +1,4 @@
-// +build !containers_image_ostree_stub
+// +build ostree_repos
 
 package ostree
 
--- a/ostree/ostree_transport.go
+++ b/ostree/ostree_transport.go
@@ -1,4 +1,4 @@
-// +build ostree_repos
+// +build containers_image_ostree
 
 package ostree
 
--- a/ostree/ostree_transport_test.go
+++ b/ostree/ostree_transport_test.go
@@ -1,4 +1,4 @@
-// +build !containers_image_ostree_stub
+// +build ostree_repos
 
 package ostree
 
--- a/ostree/ostree_transport_test.go
+++ b/ostree/ostree_transport_test.go
@@ -1,4 +1,4 @@
-// +build ostree_repos
+// +build containers_image_ostree
 
 package ostree
 
--- a/transports/alltransports/ostree.go
+++ b/transports/alltransports/ostree.go
@@ -1,4 +1,4 @@
-// +build !containers_image_ostree_stub,linux
+// +build ostree_repos,linux
 
 package alltransports
 
--- a/transports/alltransports/ostree.go
+++ b/transports/alltransports/ostree.go
@@ -1,4 +1,4 @@
-// +build ostree_repos,linux
+// +build containers_image_ostree,linux
 
 package alltransports
 
--- a/transports/alltransports/ostree_stub.go
+++ b/transports/alltransports/ostree_stub.go
@@ -1,4 +1,4 @@
-// +build containers_image_ostree_stub !linux
+// +build !ostree_repos !linux
 
 package alltransports
 
--- a/transports/alltransports/ostree_stub.go
+++ b/transports/alltransports/ostree_stub.go
@@ -1,4 +1,4 @@
-// +build !ostree_repos !linux
+// +build !containers_image_ostree !linux
 
 package alltransports
 


Silvanoc:
[Skopeo MR 607](https://github.com/containers/skopeo/pull/607) demonstrate the integration this MR passing the tests.

rhatdan:
This makes sense to me.
LGTM
@vrothberg @mtrmac @runcom PTAL

vrothberg:
I feel rather strongly against the proposed changes, mostly because they will break **all** current users of the library. containers/image is used by other tools than Podman and friends, and unless there is a technical reason to break backward compatibility, I am very hesitant.

In this specific case, the breakage can only be detected at runtime, which is further complicates users to detect the changes as existing code will compile just fine.

rhatdan:
Can we cause the code to blow up if the old tag is used?

vrothberg:
> Can we cause the code to blow up if the old tag is used?

Do you mean that ostree is compiled in although not necessarily needed?

giuseppe:
should it be `containers_image_ostree` instead of `ostree_repos`?

vrothberg:
> should it be `containers_image_ostree` instead of `ostree_repos`?

It would still break backward compatibility.

giuseppe:
> It would still break backward compatibility.

I know.  But that is fine, I don't think there is any other user except Skopeo.  The OSTree backend has a very specific use case and it is not designed/meant to be used in a generic way

rhatdan:
vrothberg
So the problem here  is someone that was expecting the ostree support will suddenly loose it with this switch correct?

vrothberg:
> So the problem here is someone that was expecting the ostree support will suddenly loose it with this switch correct?

Correct. Any build using the old build tag or those relying on the default support would be affected.

giuseppe:
> So the problem here is someone that was expecting the ostree support will suddenly loose it with this switch correct?

the ostree support in `containers/image` is only used by system containers

vrothberg:
> > So the problem here is someone that was expecting the ostree support will suddenly loose it with this switch correct?
> 
> the ostree support in `containers/image` is only used by system containers

That's the only user we're aware of but can we be sure it's not used somewhere else?

rhatdan:
If the used the old buildtag it will be ignored, and they will continue to not get ostree support.

vrothberg:
> If the used the old buildtag it will be ignored, and they will continue to not get ostree support.

That's right, but the other case is broken, namely when relying on the previous default that ostree support is opt-out.

Please don't get me wrong. I agree that excluding ostree by default would be nice. But this PR inverts a default behavior with the argument of making it easier for Podman and Buildah (who are already using the current default correctly) while not considering other users we might not be aware of. The tricky part is that inverting those defaults cannot be detected at compile-time.

giuseppe:
> That's the only user we're aware of but can we be sure it's not used somewhere else?

it could be used, but it has not really much value :-) Unless someone has also implemented the logic we have in https://github.com/projectatomic/atomic/blob/master/Atomic/syscontainers.py for managing images stored into OSTree

This is why I am arguing that in the end it won't really matter much.

To play safe we could increase the version number for the next containers/image release

vrothberg:
Alright, I don't want to block this PR if you really want it in but I am generally against breaking changes without a severe technical need or a security issue.

giuseppe:
@vrothberg @rhatdan the only point in favor is that once we stop support system containers, the ostree backend will probably go away and we won't have to support something nobody will use.

rhatdan:
I agree bump the release and put a message in the release notes.
An argument for this, is almost every platform that has tried to use it has run into ostree issues.

rhatdan:
@Silvanoc Could you add something to the README indicating the change and that the old behaviour is gone.

giuseppe:
and please `s|ostree_repos|containers_image_ostree_stub|`

Silvanoc:
I have to admit that this PR is not **my preferred approach to the issue, but PR #562**. Because of following reasons:

1. **No API changes**. Functionality remains untouched if no additional buildtags used.
1. **Support for OStree in c/image remains intact**. Without buildtag `mac_stub` exactly as currently. With buildtag `mac_stub` still supported, but without requiring `libselinux`.

These are my wishes ordered from higher to lower priority:

1. Support for OStree in c/storage for file deduplication without requiring `libselinux`.
1. Keep support for OStree in c/image for image fetching from OStree remote repository.

The usage of the buildtag `containers_image_ostree_stub` would be enough for 1. The comments in PR #562 gave me the impression that removing support for OStree in c/image has been already decided, that's why I have created this PR.

After realizing the impact of the hereby proposed changes and having the impression that removing support for OStree from c/image is not even decided yet, **I kind of regret having opened this PR and not further trying to get PR #562 merged**.

I'm sorry for the confusion, but I haven't found a better way to discuss the alternatives that I foresee.

mtrmac:
> I have to admit that this PR is not **my preferred approach to the issue, but PR #562**. Because of following reasons:

Yeah, my first impression is that something broadly like #562 is a better-targeted approach than this; I’m sorry I couldn’t find the time to return to that discussion.

Whatever the default is, both before and after this PR everyone can make a choice by modifying the build tag.

But on systems that do have both libostree or SELinux available, why _wouldn’t_ we have `ostree:` working by default? The code exists, and compiles, and it can be useful. Sure, we can talk about dropping it if it is a maintenance burden or known to be unreliable or insecure — but is it?

Or is the conclusion from #562 (where I still need to carefully read the discussion) that `libostree` is useful, but the `ostree:` backend is really not, and that the existence of the `ostree:` backend is confusing and detracts from `libostree` support in c/storage _so much_, that disabling the `ostree:` backend and pointing users to c/storage ’s `.ostree_repo` option is worth it?

giuseppe:
> Keep support for OStree in c/image for image fetching from OStree remote repository.

there is no image fetching from an OSTree remote, you would have to do this separately.  The OSTree backend is used only to store images locally.

The c/image backend for ostree is not a generic implementation.

I've tried to make it more generic in the past, and use it to copy layers back from the OSTree repository, but that works only for uncompressed layers.

To summarize: it is not usable without the atomic CLI tool (or whoever is going to implement that logic for managing images).  Since we are not planning to maintain the atomic CLI tool in future, there is no point in having the OSTree backend enabled by default.  Only Skopeo will need it for the time being.

If we will decide to have a generic implemention of ostree, that must be done differently and be hooked into container-storage.

mtrmac:
> I've tried to make it more generic in the past, and use it to copy layers back from the OSTree repository, but that works only for uncompressed layers.

It seems to work reasonably well, as an _image storage_ mechanism:
```console
$ ostree init --repo $(pwd)/t
$ skopeo copy docker://registry.fedoraproject.org/fedora:29 ostree:fedora:29@$(pwd)/t
$ skopeo copy docker://registry.fedoraproject.org/fedora:30 ostree:fedora:30@$(pwd)/t
$ skopeo copy --dest-compress=true ostree:fedora:29@$(pwd)/t dir:o-29
$ skopeo copy --dest-compress=true ostree:fedora:30@$(pwd)/t dir:o-30
```

I may be missing something known-problematic about this, but it does not look _completely broken_  at least.


> If we will decide to have a generic implemention of ostree, that must be done differently and be hooked into container-storage.

c/storage is already hooked into c/storage, via `ostree_repo` options of the `vfs` and `overlay` drivers; I don’t know whether that counts as “generic” — but it already drags the `libostree` dependency into buildah/podman, so removing c/image/ostree would only make buildah/podman smaller by the very small size of c/image/ostree itself.

---

> To summarize: it is not usable without the atomic CLI tool (or whoever is going to implement that logic for managing images).

The `atomic` tool seems to be necessary for creating container filesystems for ostree repos, but not for image _storage_ as such.

`ostree:` seems, at least, not _completely useless_ without `atomic`.

---

(I don’t have any idea whether `ostree:` is better/worse at image storage than c/storage, with or without ostree. Maybe we should disable it by default and point users to c/storage `ostree_repo` as a much more useful alternative?)

Ultimately, sure, it’s quite possible (maybe even overwhelmingly likely?) that no-one has ever used the `ostree:` transport for local storage stand-alone, and all users have been through `atomic` for system containers. I have absolutely no idea.

giuseppe:
> I may be missing something known-problematic about this, but it does not look _completely broken_ at least.

the issue is how they are going to be used.  You'd need to check out the layers manually, and copy the layers back (i.e. use ostree as the source) work only if the layers are uncompressed.

mtrmac:
> the issue is how they are going to be used. You'd need to check out the layers manually

Yes

> and copy the layers back (i.e. use ostree as the source) work only if the layers are uncompressed.

AFAICS (from the example above) that works just fine; `ostreeImageSource` now implements `LayerInfosForCopy` that represents `ostree:` sources as having uncompressed layers with matching digests.

mtrmac:
> > the issue is how they are going to be used. You'd need to check out the layers manually
> 
> Yes

… or, well, not do that and use `ostree:` only as a storage mechanism that stores data uncompressed, but deduplicates at a file level instead of at the compressed layer level like a docker/distribution server.  Hypothetically, if someone needed to record hundreds of images with comparatively small differences in a large base layer, `ostree:` could need less space than docker/distribution. (Again, maybe no-one ever does this; I have no idea.)

Silvanoc:
@giuseppe let me describe a simple hypothetical extreme use case where I see the value of OStree and you tell me which build configuration would be needed on c/image and c/storage to accomplish it:
* An embedded device running containers (where the file deduplication of OStree among different container images can be valuable due to storage limitation) and
* connected via GSM to a cloud (where the OStree HTTP fetching of single files can be valuable to security updates).

If building for example libpod with the flags `containers_image_ostree_stub` (for c/image) and `ostree` (for c/storage) accomplish it, then I probably don't need to understand what's the role of OStree in c/image (what I apparently don't do yet).

mtrmac:
> * connected via GSM to a cloud (where the OStree HTTP fetching of single files can be valuable to security updates).
> 
> If building for example libpod with the flags `containers_image_ostree_stub` (for c/image) and `ostree` (for c/storage) accomplish it, then I probably don't need to understand what's the role of OStree in c/image (what I apparently don't do yet).

mtrmac:
> * connected via GSM to a cloud (where the OStree HTTP fetching of single files can be valuable to security updates).

Nothing in GitHub.com/containers/{image,storage,skopeo,buildah,podman} can use the OSTree HTTP protocol AFAICS; the code always uses the comparatively more expensive docker/distribution protocol (or some other alternatives) for transport, OSTree, as it exists in c/image and c/storage, only ever affects storage.

Silvanoc:
> Nothing in GitHub.com/containers/{image,storage,skopeo,buildah,podman} can use the OSTree HTTP protocol AFAICS; the code always uses the comparatively more expensive docker/distribution protocol (or some other alternatives) for transport, OSTree, as it exists in c/image and c/storage, only ever affects storage.

What?! One of the most valuable features of OStree is not being used. That's a pity.

Silvanoc:
I'm really puzzled about the **role of OStree in c/storage and c/image**. There's really no information about it, but reading the code?

On the one side I imagine that @giuseppe saying that the usage of OStree in c/image is only needed for project-atomic has a good reason. But on the other side having the impression that @mtrmac and @vrothberg have some concerns about changing the default WRT OStree inclusion kind of unease me.

I'll take care of addressing the points mentioned by @giuseppe and @rhatdan in previous comments with the hope that the code is fit for merging once we have an agreement.

Silvanoc:
@rhatdan what about [this](https://github.com/containers/image/pull/587/files#diff-04c6e90faac2675aa89e2176d2eec7d8) WRT [this comment](https://github.com/containers/image/pull/587#issuecomment-466468966) of yours?

rhatdan:
@Silvanoc The README looks good.

vrothberg:
AFAICS, the agreement is to merge this PR. @Silvanoc, would you mind rebasing so we can merge the PR?

Silvanoc:
Rebased, tests are running.

vrothberg:
@rhatdan @giuseppe @mtrmac PTAL

giuseppe:
LGTM

vrothberg:
Travis failure was a network flake. I restarted it.

README.md
```diff
@@ -62,8 +62,8 @@ or use the build tags described below to avoid the dependencies (e.g. using `go
 
 - `containers_image_openpgp`: Use a Golang-only OpenPGP implementation for signature verification instead of the default cgo/gpgme-based implementation;
 the primary downside is that creating new signatures with the Golang-only implementation is not supported.
-- `containers_image_ostree_stub`: Instead of importing `ostree:` transport in `github.com/containers/image/transports/alltransports`, use a stub which reports that the transport is not supported. This allows building the library without requiring the `libostree` development libraries. The `github.com/containers/image/ostree` package is completely disabled
-and impossible to import when this build tag is in use.
+- `containers_image_ostree`: Import `ostree:` transport in `github.com/containers/image/transports/alltransports`. This builds the library requiring the `libostree` development libraries. Otherwise a stub which reports that the transport is not supported gets used. The `github.com/containers/image/ostree` package is completely disabled
```
mtrmac: “builds the library requiring” seems unclear (it seems to mean that the binary build of c/image requires development libraries, which is not the case).

Maybe “Building containers/image with this option requires the `libostree` development libraries”, or something like that.

mtrmac:
> AFAICS, the agreement is to merge this PR.

Fair enough.

---

What happens with the dependent programs (and, if different, with the Fedora/… RPMs)?
Looking at the proposed PRs:

- https://github.com/containers/skopeo/pull/607
- https://github.com/containers/libpod/pull/2407
- https://github.com/containers/buildah/pull/1360

all of these would continue to build the `ostree:` backend _if_ the headers etc. are present, i.e. no change to their contents, and no change to packaging?

Or do we perhaps want to drop `ostree:` from the default builds of `Buildah` and `libpod`? Or even from `skopeo` as well (upstream only, or in Fedora as well)?

(Raising this only in case there is clear consensus, so that the three PRs can be updated before merging them at all; if there is not a clear consensus, let’s merge just this PR here, and discuss the other changes in the three follow-up PRs.)

giuseppe:
> What happens with the dependent programs (and, if different, with the Fedora/… RPMs)?
> Looking at the proposed PRs:
> 
>     * [containers/skopeo#607](https://github.com/containers/skopeo/pull/607)
> 
>     * [containers/libpod#2407](https://github.com/containers/libpod/pull/2407)
> 
>     * [containers/buildah#1360](https://github.com/containers/buildah/pull/1360)
> 
> 
> all of these would continue to build the `ostree:` backend _if_ the headers etc. are present, i.e. no change to their contents, and no change to packaging?
> 
> Or do we perhaps want to drop `ostree:` from the default builds of `Buildah` and `libpod`? Or even from `skopeo` as well (upstream only, or in Fedora as well)?

I think we want to drop ostree from buildah and podman, but we still need to keep it inside skopeo as atomic system containers are using it

Silvanoc:
@giuseppe does [this comment](https://github.com/containers/image/pull/587#issuecomment-466456388) of yours imply that on the long run you want to remove the OSTree backend from Skopeo too?

> @vrothberg @rhatdan the only point in favor is that once we stop support system containers, the ostree backend will probably go away and we won't have to support something nobody will use.

Hopefully not... Although I don't currently need it.

giuseppe:
> @giuseppe does [this comment](https://github.com/containers/image/pull/587#issuecomment-466456388) of yours imply that on the long run you want to remove the OSTree backend from Skopeo too?

yes, the long term plan is to drop the ostree backend from Skopeo as well.  That depends on the atomic CLI tool which is still used.

giuseppe:
@Silvanoc could you rebase this PR and address the remaining comments?

Silvanoc:
> @Silvanoc could you rebase this PR and address the remaining comments?

Rebase done, CI tests are ongoing.

I thought, that I had already addresses all the comments... @giuseppe have I overseen something?

giuseppe:
> Rebase done, CI tests are ongoing.
> 
> I thought, that I had already addresses all the comments... @giuseppe have I overseen something?

no sorry, everything is addressed.

LGTM

PR #588, opened 2019-02-22
types: Add SystemContext.GetAuth

wking:
The previous system is difficult to configure, with several `SystemContext`` properties feeding the auth-getting logic.  It also assumed that there was either a single username/password (`DockerAuthConfig`) or that the auth information was on disk somewhere.  With this pull request, I'm pushing all of the complexity down into a user-supplied getter (falling back to the old logic if the getter is unset).  That makes it easy to plug in your own alternative without worrying about the default logic.

The Docker-config `AuthStore` has a public `Config` property to make it easy for callers to drop in their own configuration for read-only access without having to go through either the filesystem or JSON. I've made the backing types public to support that use-case, and set custom JSON serialization for the newly-public Auth structure to allow Go callers to avoid having to understand the base64 wrapping used in the JSON serialization.

I've also shifted a number of auth-getting unit tests from `docker_client_test.go` into `config_test.go`, since they only excercise `config.go` logic.

Ideally the auth interface would support challenge/response auth like [HTTP's][1], but for now it just uses Docker's "we'll be able to hard-code authorization for each [authority][2]" approach.

[1]: https://tools.ietf.org/html/rfc7235
[2]: https://tools.ietf.org/html/rfc3986#section-3.2


pkg/docker/config/config.go
```diff
@@ -37,111 +75,247 @@ var (
...
+	if sys != nil {
+		if sys.DockerAuthConfig != nil {
+			return sys.DockerAuthConfig.Username, sys.DockerAuthConfig.Password, nil
+		}
+
+		if sys.AuthFilePath != "" {
+			authStore.Path = sys.AuthFilePath
+		} else if sys.RootForImplicitAbsolutePaths != "" {
+			authStore.Path = filepath.Join(sys.RootForImplicitAbsolutePaths, fmt.Sprintf(defaultPerUIDPathFormat, os.Getuid()))
+		}
```
mtrmac: Why was this moved out of `getPathToAuth`/`getPath`? Now only `GetAuthentication`, but not `SetAuthentication`/`RemoveAuthentication` implement the `SystemContext` options.

types/types.go
```diff
@@ -401,11 +401,17 @@ type ImageInspectInfo struct {
...
+//
+// Deprecated: Use github.com/containers/image/credentials/single's AuthStore instead.
 type DockerAuthConfig struct {
 	Username string
 	Password string
 }
 
+// AuthGetter retrieves credentials for HTTP(S) access.  It returns
+// the username and secret as strings.
+type AuthGetter func(serverURL string) (string, string, error)
```
mtrmac: It’s very much not obvious to me that it makes sense to have a single callback for _all kinds of_ “HTTP(S)” access, no matter the protocol on top of HTTPS, or authentication mechanism. (Sure, the `AuthFilePath` name is similarly misleading; we originally thought that it would somehow be generic, but it clearly is not.)

WRT `serverURL`, in docker/distribution the lookup key is not an URL!

types/types.go
```diff
@@ -451,14 +457,18 @@ type SystemContext struct {
 	RegistriesDirPath string
 	// Path to the system-wide registries configuration file
 	SystemRegistriesConfPath string
-	// If not "", overrides the default path for the authentication file
+	// If not "", overrides the default path for the authentication file.  Ignored for reading if AuthConfig is not "".
+	//
+	// Deprecated: Use the GetAuth property and github.com/containers/image/pkg/docker/config's AuthStore instead.
```
mtrmac: It’s not obviously attractive to force the users to create extra objects when they could just set a clearly-named field. Sure, the consumer of the data might need to define what happens if multiple fields are set, or refuse such configurations, but that seems worth it.

pkg/docker/config/config.go
```diff
@@ -37,111 +75,247 @@ var (
...
+		return s.save()
+	}
+
+	return nil
+}
+
+// SetAuthentication stores the username and password in the auth.json file
+//
+// Deprecated: Use an AuthStore.
+func SetAuthentication(sys *types.SystemContext, registry, username, password string) error {
```
mtrmac: If `AuthStore` is supposed to be a replacement, it needs to support `SystemContext`.

pkg/docker/config/config.go
```diff
@@ -37,111 +75,247 @@ var (
...
+		Username:  username,
+		Secret:    password,
 	})
 }
 
 // GetAuthentication returns the registry credentials stored in
 // either auth.json file or .docker/config.json
 // If an entry is not found empty strings are returned for the username and password
+//
+// Deprecated: Use an AuthStore.
```
mtrmac: If `AuthStore` is supposed to be a replacement, it needs to make it easy to use all relevant fields of `SystemContext`.

types/types.go
```diff
@@ -485,7 +495,9 @@ type SystemContext struct {
 	DockerPerHostCertDirPath string
 	// Allow contacting docker registries over HTTP, or HTTPS with failed TLS verification. Note that this does not affect other TLS connections.
 	DockerInsecureSkipTLSVerify OptionalBool
-	// if nil, the library tries to parse ~/.docker/config.json to retrieve credentials
+	// if nil, the library tries to parse ~/.docker/config.json to retrieve credentials.
+	//
+	// Deprecated: Use the GetAuth property and github.com/containers/image/credentials/single's AuthStore instead.
```
mtrmac: It’s not obviously attractive to force the users to create extra objects when they could just set a clearly-named field. Sure, the consumer of the data might need to define what happens if multiple fields are set, or refuse such configurations, but that seems worth it.

Plumbing together creation of a `single.Credential`+`single.AuthStore`, and _somehow_ making an `AuthGetter` out of that, seems way too much code compared to a `DockerAuthConfig: &types.DockerAuthConfig{…}`

credentials/single/single.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"github.com/docker/docker-credential-helpers/credentials"
+	"github.com/pkg/errors"
+)
+
+// AuthStore is a basic credentials store that holds a single entry.
+type AuthStore *Credential
+
+// Add sets the store's credentials.
+func (s AuthStore) Add(cred *credentials.Credentials) error {
```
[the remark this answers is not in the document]
wking: > ... but there is no shared interface defined...

The shared interface is [`credentials.Helper`][1].  But since all we really need is a getter, I can remove `Add`, `Delete`, and `List` here.

[1]: https://godoc.org/github.com/docker/docker-credential-helpers/credentials#Helper

pkg/docker/config/config.go
```diff
@@ -1,29 +1,67 @@
...
+		return nil
+	}
+
+	auth.Username = parts[0]
+	auth.Secret = strings.Trim(parts[1], "\x00")
+	return nil
+}
+
+// Config holds the full authorization configuration.
+type Config struct {
```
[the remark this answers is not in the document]
wking: > And if the goal is to allow callers to submit a Docker-like config in memory without creating a file, they still don’t need this type exposed.

That's the goal (for the OpenShift installer, which has the pull-secret portion of this file in memory as `[]byte`).  And while we could have an internal `config` holding the structured data with a `LoadFromBytes([]byte)` or similar method, it seemed easier and more convenient to just make the structured configuration public (e.g. see the unit tests, which no longer need the `makeTestAuthConfig` helper).  Do you have concerns about having this be public?

pkg/docker/config/config.go
wking: > Why was this moved out of `getPathToAuth`/`getPath`?

Because I want the `AuthStore` struct to be consumed by `SystemContext`, and don't want a dependency cycle.

> Now only `GetAuthentication`, but not `SetAuthentication`/`RemoveAuthentication` implement the `SystemContext` options.

Fixed with 8ce2b8b -> 5e011a3.

credentials/single/single.go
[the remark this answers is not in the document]
wking: > But since all we really need is a getter, I can remove `Add`, `Delete`, and `List` here.

Done with 8ce2b8b -> 5e011a3.

types/types.go
wking: > WRT `serverURL`, in docker/distribution the lookup key is not an URL!

The interface I'm implementing is [`credentials.Helper`][1], which uses `serverURL`.  This getter is the `Get` method from that interface.  I'm fine using a different string if you want for the input argument (personally, I prefer "[authority][2]", or adjusting the docstring to whatever you want me to use instead, just tell me what to paste in ;).

> ... we originally thought that it would somehow be generic, but it clearly is not...

HTTP(S) auth [is a generic thing][3].  Unfortunately, the Docker APIs don't really conform, since they don't provide a way to feed in a challenge string.  Should we come up with our own API (or lean on one that already exists in Go) here, and then special-case the Docker approach with dummy challenge?

[1]: https://godoc.org/github.com/docker/docker-credential-helpers/credentials#Helper
[2]: https://tools.ietf.org/html/rfc3986#section-3.2
[3]: https://tools.ietf.org/html/rfc7235

types/types.go
wking: > Sure, the consumer of the data might need to define what happens if multiple fields are set, or refuse such configurations, but that seems worth it.

`SystemContext` seems complicated enough already, to me it is a clear win to factor out the auth-configuration properties into other locations.  For folks who currently use `SystemContext`, the transitions seem pretty clear to me, although I can work up migration docs if that would help address your concerns (or even just give us something more concrete to anchor this discussion).

pkg/docker/config/config.go
wking: > If `AuthStore` is supposed to be a replacement, it needs to support `SystemContext`.

I don't think so.  The Docker client needs a way to get credentials for a given authority, and that's what the new `GetAuth` property provides.  Users will need a way to create that getter, and that's what the two new `AuthStore` structures provide.  The old `SystemContext` properties either configure a single credential (which can be replaced easily with the new `single.AuthStore`) or the path lookup for the Docker-config `AuthStore` (which can be replaced by creating your own `config.AuthStore` with `Path` set however they like).  The rest of `SystemContext` seems unrelated to managing authentication, right?

types/types.go
wking: > Plumbing together creation of a `single.Credential`+`single.AuthStore`, and somehow making an `AuthGetter` out of that...

```console
$ git describe
v1.3-49-g44801dd
$ cat test.go
package main

import (
	"fmt"

	"github.com/containers/image/credentials/single"
	"github.com/docker/docker-credential-helpers/credentials"
)

func main() {
	authStore := single.AuthStore(credentials.Credentials{
		ServerURL: "example.org",
		Username: "alice",
		Secret: "secret",
	})
	sys := &types.SystemContext{
		GetAuth: authStore.Get,
	}
	username, secret, err := sys.GetAuth("example.org")
	fmt.Println(username)
	fmt.Println(secret)
	fmt.Println(err)
}
$ go run test.go 
alice
secret
<nil>
```

That doesn't seem too complicated to me, and it saves you from having to explain how `DockerAuthConfig` interacts with a `GetAuth` or other `SystemContext` properties.

wking:
I'll rebase this once #589 lands that pivot, so I can rebase this on top to shrink the diff size.

rhatdan:
Looks like it has merged.

wking:
Rebased onto master (around #589) with 44801dd -> 245b3b3, which also fixes a few bugs.

wking:
Rebased onto master (no conflicts) with 245b3b3 -> 4800205 to resolve the "This branch is out-of-date with the base branch" warning.

wking:
I'm going to hold off on rebasing this until #597 settles.

rhatdan:
@wking  #597 merged can you rebase.

wking:
I took a run at rebasing this, but the new keyring stuff that started in #619 is too confusing for me ;).  If anyone else wants to pick this up and rebase it, please go ahead without waiting on me.

PR #590, opened 2019-02-26
blobinfocache: sync via mutex

vrothberg:
Serialize method calls via a mutex to avoid concurrent accesses to the
map leading to panics.

Fixes: https://github.com/containers/libpod/issues/2420

### DIFF
--- a/pkg/blobinfocache/memory.go
+++ b/pkg/blobinfocache/memory.go
@@ -1,6 +1,7 @@
 package blobinfocache
 
 import (
+	"sync"
 	"time"
 
 	"github.com/containers/image/types"
@@ -17,6 +18,7 @@ type locationKey struct {
 
 // memoryCache implements an in-memory-only BlobInfoCache
 type memoryCache struct {
+	mutex                 *sync.Mutex // synchronizes concurrent accesses
 	uncompressedDigests   map[digest.Digest]digest.Digest
 	digestsByUncompressed map[digest.Digest]map[digest.Digest]struct{}             // stores a set of digests for each uncompressed digest
 	knownLocations        map[locationKey]map[types.BICLocationReference]time.Time // stores last known existence time for each location reference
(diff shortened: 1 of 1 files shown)


vrothberg:
@mtrmac @rhatdan @mheon PTAL

vrothberg:
@mtrmac @rhatdan "Please update billing information for containers" what shall we do with the pullapprove?

rhatdan:
LGTM
@nalind PTAL
@vrothberg How much is this going to hurt performance?

vrothberg:
> @vrothberg How much is this going to hurt performance?

This should have a low impact. I count three calls in dockerImageDest and one in dockerImageSrc (conditional). And two in storage.

giuseppe:
LGTM

mtrmac:
(ACK, for the record.)

> > @vrothberg How much is this going to hurt performance?
> 
> This should have a low impact. I count three calls in dockerImageDest and one in dockerImageSrc (conditional). And two in storage.

Typically, zero — we should be using an on-disk BoltDB cache. This implementation is used only if the on-disk cache is not accessible (and the debug log says why, search for “using a memory-only cache”).  And even with this mutex the memory cache is strictly faster than the BoltDB one (which also involves two mutexes on every access, in addition to file accesses including file locks).

mtrmac:
> @mtrmac @rhatdan "Please update billing information for containers" what shall we do with the pullapprove?

Just drop it, I think; we haven’t been _ever_ seriously enforcing the configured “two independent reviews” configuration, at best it included one self-review.

vrothberg:
> > @mtrmac @rhatdan "Please update billing information for containers" what shall we do with the pullapprove?
> 
> Just drop it, I think; we haven’t been _ever_ seriously enforcing the configured “two independent reviews” configuration, at best it included one self-review.

Sounds good. Thanks, @mtrmac !

PR #594, opened 2019-03-01
Certs.d manpage & *.5.md suffix

vrothberg:
@mtrmac @rhatdan @TomSweeneyRedHat PTAL

Context: https://bugzilla.redhat.com/show_bug.cgi?id=1677264


vrothberg:
Ah, the base branch was wrong. Rebasing now.

rhatdan:
LGTM

rhatdan:
@TomSweeneyRedHat PTAL

docs/containers-certs.d.5.md
```diff
@@ -0,0 +1,34 @@
...
+   └── ca.crt               <- Certificate authority that signed the registry certificate
+```
+
+### Creating Client Certificates
+A client certificate pair can be generated via `openssl`:
+
+```
+$ openssl genrsa -out client.key 4096
+$ openssl req -new -x509 -text -key client.key -out client.cert
+```
```
TomSweeneyRedHat: My only question is either of these examples valid to get a Container Registry up and running?  The one that I cobbled together with @edsantiago 's help is:
```
openssl req -newkey rsa:4096 -nodes -sha256 -keyout /root/auth/domain.key -x509 -days 2 -out /root/auth/domain.crt -subj "/C=US/ST=Foo/L=Bar/O=Red Hat, Inc./CN=localhost"
```
I don't know if that's entirely up to snuff, you'd definitely want to tweak the `/root/auth` and `Red Hat`, but I think it may be closer.

vrothberg: They do the same, but your command generates the in the call `-newkey rsa:4096` and does not prompt the user to enter data such as country, province, organization, etc. as those are set in the subject.

docs/containers-certs.d.5.md
```diff
@@ -0,0 +1,34 @@
...
+# NAME
+containers-certs.d - Directory for storing custom container-registry certficates
+
+# DESCRIPTION
+A custom certificate for a container registry can be configured by creating a directory under `/etc/containers/certs.d`.
+The name of the directory must correspond to the `host:port` of the registry (e.g., `my-registry.com:5000`).
+
+## Directory Structure
+A certs directory can contain one or more CA roots (i.e., `.crt` files) and one or more `<client>.{cert,key}` cert pairs.
+If more than one root or pair is specified, they will be used in alpha-numerical order.
```
mtrmac: I don’t really think trusted CAs have an “order”; if the server provides enough certificates to build a path (handwaving) from the leaf to any of the trusted CAs, it is accepted.

For the cert+key, they are primarily filtered by the CAs the server sends as acceptable for client authentication, but then yes, the order seems to be by ASCII code comparison.

docs/containers-certs.d.5.md
vrothberg: > Overall I’d suggest removing this section completely.

done

vrothberg:
Updated. @mtrmac, can you have another look?

mtrmac:
LGTM. Thanks!

PR #595, opened 2019-03-08
additional authentication files

vrothberg:
Allow specifying additional authentication files passed through
`types.SystemContext`.  Those additional files are only used for reading
credentials.  The `AuthFilePath` remains the center of auth management
and is the only authentication file used for storing new and removing
old credentials.

Having the ability to specify additional authentication files can come
in handy to unify multiple entities that have dedicated auth files, for
instance, the K8s kubelet.  See the below bug for additional details.

Also add debug logs to `findAuthentication()` to ease debugging
potential future regressions.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1686556

### DIFF
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -51,6 +51,14 @@ func SetAuthentication(sys *types.SystemContext, registry, username, password st
 	})
 }
 
+func additionalAuthFiles(sys *types.SystemContext) []string {
+	paths := []string{}
+	if sys != nil && sys.AdditionalAuthFiles != nil && len(sys.AdditionalAuthFiles) > 0 {
+		paths = append(paths, sys.AdditionalAuthFiles...)
+	}
+	return paths
+}
+
 // GetAuthentication returns the registry credentials stored in
 // either auth.json file or .docker/config.json
 // If an entry is not found empty strings are returned for the username and password
@@ -70,6 +78,8 @@ func GetAuthentication(sys *types.SystemContext, registry string) (string, strin
 		// Logging the error as a warning instead and moving on to pulling the image
 		logrus.Warnf("%v: Trying to pull image in the event that it is a public image.", err)
 	}
+
+	paths = append(paths, additionalAuthFiles(sys)...)
 	paths = append(paths, filepath.Join(homedir.Get(), dockerHomePath), dockerLegacyPath)
 
 	for _, path := range paths {
@@ -93,9 +103,14 @@ func GetUserLoggedIn(sys *types.SystemContext, registry string) (string, error) 
 	if err != nil {
 		return "", err
 	}
-	username, _, _ := findAuthentication(registry, path, false)
-	if username != "" {
-		return username, nil
+	paths := []string{path}
+	paths = append(paths, additionalAuthFiles(sys)...)
+
+	for _, path := range paths {
+		username, _, _ := findAuthentication(registry, path, false)
+		if username != "" {
+			return username, nil
+		}
 	}
 	return "", nil
 }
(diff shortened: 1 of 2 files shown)


pkg/docker/config/config.go
```diff
@@ -51,6 +51,14 @@ func SetAuthentication(sys *types.SystemContext, registry, username, password st
 	})
 }
 
+func getAdditionalAuthFiles(sys *types.SystemContext) []string {
```
rhatdan: Nit: Golang frowns on beginning commands with get. Just call it additionalAuthFiles

rhatdan:
Small nit, but LGTM

pkg/docker/config/config.go
vrothberg: Most functions in this file start with `{get,set}` but I apply the change.

pkg/docker/config/config.go
```diff
@@ -51,6 +51,14 @@ func SetAuthentication(sys *types.SystemContext, registry, username, password st
 	})
 }
 
+func additionalAuthFiles(sys *types.SystemContext) []string {
```
mtrmac: I don’t quite think there four lines are with a function… it’s three lines in the caller, which already has a `paths` variable, and we have ~5 lines of overhead with defining the function.

Rather, the full computation of all candidate paths should probably be unified between `GetAuthentication` and `GetUserLoggedIn`; it seems a bit unlikely that the existing divergence was intentional (OTOH I haven’t checked the past discussions.)

mtrmac: > I don’t quite think there four lines are with a function

“are worth a function”… macOS autocorrection is infuriating sometimes.

ashcrow:
Travis report:
```
 github.com/containers/image/storage
storage/storage_image.go:735:55: not enough arguments in call to s.imageRef.transport.store.SetImageBigData
	have (string, string, []byte)
	want (string, string, []byte, func([]byte) (digest.Digest, error))
storage/storage_image.go:768:54: not enough arguments in call to s.imageRef.transport.store.SetImageBigData
	have (string, string, []byte)
	want (string, string, []byte, func([]byte) (digest.Digest, error))
storage/storage_image.go:775:54: not enough arguments in call to s.imageRef.transport.store.SetImageBigData
	have (string, string, []byte)
	want (string, string, []byte, func([]byte) (digest.Digest, error))
storage/storage_image.go:784:55: not enough arguments in call to s.imageRef.transport.store.SetImageBigData
	have (string, string, []byte)
	want (string, string, []byte, func([]byte) (digest.Digest, error))
Makefile:39: recipe for target 'build-internal' failed
```

vrothberg:
> Travis report:
> 
> ```
>  github.com/containers/image/storage
> storage/storage_image.go:735:55: not enough arguments in call to s.imageRef.transport.store.SetImageBigData
> 	have (string, string, []byte)
> 	want (string, string, []byte, func([]byte) (digest.Digest, error))
> storage/storage_image.go:768:54: not enough arguments in call to s.imageRef.transport.store.SetImageBigData
> 	have (string, string, []byte)
> 	want (string, string, []byte, func([]byte) (digest.Digest, error))
> storage/storage_image.go:775:54: not enough arguments in call to s.imageRef.transport.store.SetImageBigData
> 	have (string, string, []byte)
> 	want (string, string, []byte, func([]byte) (digest.Digest, error))
> storage/storage_image.go:784:55: not enough arguments in call to s.imageRef.transport.store.SetImageBigData
> 	have (string, string, []byte)
> 	want (string, string, []byte, func([]byte) (digest.Digest, error))
> Makefile:39: recipe for target 'build-internal' failed
> ```

Thanks! That comes from containers/storage and containers/image not liking each other over at Skopeo. We need to fix this as well.

mtrmac:
> Thanks! That comes from containers/storage and containers/image not liking each other over at Skopeo.

This is the fallout from https://github.com/containers/storage/pull/299 .

vrothberg:
Closing as we found another solution for CRI-O :)

wking:
I think #588 would give an API that makes this sort of thing convenient, even if we may be going with openshift/machine-config-operator#535 in this particular case ;).

Issue #596, opened 2019-03-08
Incorrect/inconsistent docker/config.json path lookup


mtrmac:
- If the user specifies `SystemContext.AuthFilePath`, we should not be checking the home directory
- <s>It’s surprising that `GetAuthentication` and `GetUserLoggedIn` use different sets of paths.</s>

rhatdan:
Since the PR Is merged, I am closing this issue.

mtrmac:
The home directory aspect is still outstanding.

PR #333, opened 2017-08-29
Add authentication package to containers/image/pkg

umohnani8:
This package is used in authenticating a user for kpod login
and can be used for authentication in kpod push, pull etc.

Signed-off-by: umohnani8 <umohnani@redhat.com>

PR #597, opened 2019-03-08
pkg/docker/config: Deprecate GetUserLoggedIn

wking:
And refactor it to use `GetAuthentication`, which is just as convenient for callers.  This also removes the lookup differences between lookup logic between the two functions (e.g. `GetUserLoggedIn` used to ignore `DockerAuthConfig`).

Spun off from #588 [here][1] and #596.

[1]: https://github.com/containers/image/commit/480020500db2b2ca5ab850c6657a5df4d0b98904#diff-45d77e7dc3a84219aeb2aaad81b6bf96R298

### DIFF
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -85,21 +85,6 @@ func GetAuthentication(sys *types.SystemContext, registry string) (string, strin
 	return "", "", nil
 }
 
-// GetUserLoggedIn returns the username logged in to registry from either
-// auth.json or XDG_RUNTIME_DIR
-// Used to tell the user if someone is logged in to the registry when logging in
-func GetUserLoggedIn(sys *types.SystemContext, registry string) (string, error) {
-	path, err := getPathToAuth(sys)
-	if err != nil {
-		return "", err
-	}
-	username, _, _ := findAuthentication(registry, path, false)
-	if username != "" {
-		return username, nil
-	}
-	return "", nil
-}
-
 // RemoveAuthentication deletes the credentials stored in auth.json
 func RemoveAuthentication(sys *types.SystemContext, registry string) error {
 	return modifyJSON(sys, func(auths *dockerConfigFile) (bool, error) {


vrothberg:
Need to update containers/image to reflect the latest changes in containers/storage:
```
# github.com/containers/image/storage
storage/storage_image.go:735:55: not enough arguments in call to s.imageRef.transport.store.SetImageBigData
	have (string, string, []byte)
	want (string, string, []byte, func([]byte) (digest.Digest, error))
```

Code LGTM :+1: Thanks, @wking !

mtrmac:
If we are doing this, I’d just remove the method; a `logrus.Warn` is not really wanted by the callers and it’s not actionably by end users.

I do need to check the earlier history of the code (notably https://github.com/containers/image/pull/333 ), the different handling and/or the `.docker` fallback mentioned in #596 might actually have been intentional — I can’t remember any more.

wking:
> If we are doing this, I’d just remove the method...

I've pushed ee8c757 -> e64a79b, removing it.

> I do need to check the earlier history of the code (notably #333 ), the different handling and/or the `.docker` fallback mentioned in #596 might actually have been intentional...

I checked, and saw no discussion in either the commit message or pull request that motivated the different lookup logic.  You can obviously still check for yourself too ;).

vrothberg:
@wking would you rebase, so we can merge it? Note that `podman-login` uses `GetUserLoggedIn()` and must be updated accordingly when we vendor containers/image over there (Cc @baude @mheon).

rhatdan:
@wking ping...

vrothberg:
I am updating this PR, so we can merge it.

vrothberg:
Note to ourselves, we must update Podman (and Buildah?) afterward.

vrothberg:
Fixes: https://github.com/containers/image/issues/596

vrothberg:
LGTM, @mtrmac @rhatdan PTAL

rhatdan:
LGTM

lio24:
I need window 10 to download docker on my PC

PR #52, opened 2016-08-17
Implement a lookaside storage for signatures of images in Docker registries

mtrmac:
Allows configuring `sigstore` (for RW) and `sigstore-write` (write-only) for images/repositories/namespaces/registries in `atomic.conf`.
    
If configured, uses them to store signatures in the `docker:` transport.

This works fine for `file://` sigstore URLs, for read-only `http://` and `https://` `sigstore` URLs, and is at best a skeleton for writing to `http://` / `https://` (no authentication, to start). So, with this, pushing signatures to remote Docker registries is still not reasonably possible, but at least you can push to a `file://` pointing to a webroot of a sigstore served to the public over HTTP. (Or you can configure `sigstore` to point to HTTP to pull from the same sources as your users, and `sigstore-write` to the webroot.)

Does not deal with any kind of discovery, the lookaside needs to be manually configured in advance.

This pulls in #49 for `types.SystemContext` and significantly extends it, adding a generic `RootForImplicitAbsolutePaths` and adding a `SystemContext` parameter to `types.NewImage{,Source,Destination}` as tentatively discussed in #41 . (Also note that we now implicitly use the `ATOMIC_CONF` environment variable, for compatibility with `projectatomic/atomic`). I will rebase as necessary.

Parsing `atomic.conf` here in generic is admittedly fairly ugly, but perhaps still better than defining an interface for this and requiring `skopeo` to implement that…

More details are in commit messages.

PR #598, opened 2019-03-09
docker/docker_client: Drop redundant Domain(ref.ref) call

wking:
There have been redundant calls here since two `ref.ref.Hostname()` calls were added in aaedc642 (#52).  At that point the two calls were separated by a `dockerHostname` check which could have been shifted by two lines to avoid the doubled function calls.  But in f28367e1 (#333) the `dockerHostname` check moved to a separate function entirely (`newDockerClientWithDetails`) while the `Domain()` calls remained together in `newDockerClientFromRef`.  So now there is no longer any reason for the second call, and this pull request drops it.


vrothberg:
Good catch, @wking!
LGTM

mtrmac:
LGTM pending tests.  Thanks!

wking:
Heh, Travis couldn't find the build :p.  Rebased onto master with 5683b91 -> 9e941c2 to trigger another round of tests.

PR #474, opened 2018-07-05
Make it easier for CONTRIBUTORS to find information

rhatdan:
Signed-off-by: Daniel J Walsh <dwalsh@redhat.com>

PR #600, opened 2019-03-09
README: Drop '**' from "Contributing" header

wking:
These are from f4d6188f (#474).  I dunno if they were intentional or not, but I thought they might be footnote markers so I scrolled down to the bottom of the README looking for a footnote.  Remove them to save other people from that same confusion ;).


vrothberg:
LGTM, thanks @wking!

CI flakes are entirely unrelated to this PR.

PR #602, opened 2019-03-10
pkg/blobinfocache: Split implementations into subpackages

wking:
This allows consumers who don't need the defaulting logic to pick their desired implementation without pulling in unrelated dependencies.  For example, it allows you to consume the memory or `NoCache` implementations without pulling in `github.com/boltdb`.

The blobinfocache API remains unchanged, but its `NoCache` variable and `New*` getters have been deprecated in favor of the new subpackages.  Once consumers have migrated, we can drop the deprecated API.

Having single-implementation subpackages also allows for more generic names (e.g. `NewMemoryCache` is now just `New`), because `memory.New` is clear enough while `memory.NewMemoryCache` stutters.


pkg/blobinfocache/none/none.go
```diff
@@ -1,4 +1,5 @@
-package blobinfocache
+// Package none implements a dummy BlobInfoCache which records no data.
+package none
```
vrothberg: `none.NoCache` seems redundant. Could we leave this in the `blobinfocache` package? It also seems convenient to have a dummy implementation in the same package as the interface.

wking: > `none.NoCache` seems redundant.

I'm fine renaming `NoCache`, but couldn't think of anything I liked better.

> Could we leave this in the `blobinfocache` package?

No, because that package has `DefaultCache`, which pulls in all the implementations.  So `NoCache` there would require `NoCache` consumers vendor (or have non-vendor access to) the BoltDB libraties.

> It also seems convenient to have a dummy implementation in the same package as the interface.

The *interface* is over in `types` ;).

mtrmac:
> This allows consumers who don't need the defaulting logic to pick their desired implementation without pulling in unrelated dependencies. For example, it allows you to consume the memory or `NoCache` implementations without pulling in `github.com/boltdb`.

Sure, this is conceptually cleaner, but what are you ultimately trying to do?

The cache works much better when _everyone_ updates the same one, so almost everyone should be using `DefaultCache` (that the in-memory cache is currently used only as a fallback, and in tests). Providing the individual cache implementations as user-selectable components is a bit counter-productive in that sense.

In particular, I really don’t want users of this library choosing the memory cache just because they can avoid a dependency on BoltDB. 

Do you have a special use case that really shouldn’t be using the persistent cache? (Sure, there could well be one; I just want to make sure that the dependency is not the only motivation.)

> The blobinfocache API remains unchanged, but its `NoCache` variable and `New*` getters have been deprecated in favor of the new subpackages. Once consumers have migrated, we can drop the deprecated API.

(FWIW so far there hasn’t been any strong demand for not breaking the API (I did ask; and sure, a string demand can come at any time, including from you), so we’ve historically been preferring a clean API to compatibility wrappers.)

directory/directory_test.go
```diff
@@ -58,7 +58,7 @@ func TestGetPutManifest(t *testing.T) {
 func TestGetPutBlob(t *testing.T) {
 	ref, tmpDir := refToTempDir(t)
 	defer os.RemoveAll(tmpDir)
-	cache := blobinfocache.NewMemoryCache()
+	cache := memory.New()
```
mtrmac: (Absolutely non-blocking: `memory.New()` can be anything, and `boltdb.New()` is even a bit confusing. If we are retaining `NoCache`, maybe `NewCache` would still be worth it?  )

pkg/blobinfocache/deprecated.go
```diff
@@ -0,0 +1,35 @@
+package blobinfocache
```
mtrmac: So far we’ve not been maintaining API compatibility; is this a good time to start? Is there anyone who would care?

pkg/blobinfocache/prioritize/prioritize.go
```diff
@@ -1,4 +1,6 @@
-package blobinfocache
+// Package prioritize provides utilities for prioritizing locations in
+// types.BlobInfoCache.CandidateLocations.
+package prioritize
```
mtrmac: I think no external user needs to care; move this into a `pkg/blobinfocache/internal` (or `pkg/blobinfocache/internal/prioritize`? I don’t have a opinion really).

pkg/blobinfocache/test/test.go
```diff
@@ -1,4 +1,5 @@
-package blobinfocache
+// Package test provides generic BlobInfoCache test helpers.
```
mtrmac: `pkg/blobinfocache/internal/test`, please.

wking:
> In particular, I really don’t want users of this library choosing the memory cache just because they can avoid a dependency on BoltDB.

Nope, that's exactly why I worked this up ;).

> Do you have a special use case that really shouldn’t be using the persistent cache? (Sure, there could well be one; I just want to make sure that the dependency is not the only motivation.)

openshift/installer#1286 is fetching a release image, looking in that filesystem for a `release-manifests/image-references` file, fetching the referenced `machine-os-content` image, and looking at its annotations.  This is effectively a handful of HTTP requests, unmarshalling into manifests, and untar'ing some layers.  The associated vendor bump is crazy though, pulling in BoltDB and even Prometheus (openshift/installer@e189de133ffc).  Caching is nice, but in this case we're pulling some small manifests and one single, smallish layer:

```console
$ podman pull quay.io/openshift-release-dev/ocp-release:4.0.0-0.7
$ podman save -o release.tar quay.io/openshift-release-dev/ocp-release:4.0.0-0.7
$ tar xvf release.tar
$ jq -r '.rootfs.diff_ids[]' 2f361c5c569b16577e706fa542ce415604d48c7fee50696f479eb02647e72d72.json 
sha256:20b48f7ee69e4b7f0633e7c6201b9dc54373747e8875d75cd99851e1f8186bc2
sha256:5975baa1f989ae582ded9d754602ed73e606cc2243078d31f89657b70fba4168
sha256:8e52d812c30eb9a92d03b931d769d62f24f26bc5362d31e0291d4131a49c1d31
sha256:2c53f1a826d3020c53d2eddf9a34f4c5e6e6200a2ebd3ef747b0cdd78fd4972f
sha256:7900974d6b5b9209c525f97f0e68d214168004d7082e9d3d45a0c9d1da473fbd
$ ls -lh 7900974d6b5b9209c525f97f0e68d214168004d7082e9d3d45a0c9d1da473fbd.tar 
-r--r--r--. 1 trking trking 471K Dec 31  1969 7900974d6b5b9209c525f97f0e68d214168004d7082e9d3d45a0c9d1da473fbd.tar
```

Caching seems like an excessive complication when we're that lightweight.

> (FWIW so far there hasn’t been any strong demand for not breaking the API (I did ask; and sure, a string demand can come at any time, including from you), so we’ve historically been preferring a clean API to compatibility wrappers.)

I'm fine dropping the deprecated endpoints now if we have no compatibility requirements.  Do we publish a change-log or something to help consumers migrate over breaking changes?

directory/directory_test.go
wking: > Absolutely non-blocking: memory.New() can be anything...

But the import path is `github.com/containers/image/pkg/blobinfocache/memory`, which includes "cache".  And the receiving `cache` here also reminds readers about what this is.  Folks who want to include more information are free to name their import:

```go
import (
  blobinfomemorycache "github.com/containers/image/pkg/blobinfocache/memory"
)
...
cache := blobinfomemorycache.New()
```

So I like `New`.  But I'm not a maintainer; if you want me to make it `NewCache` or whatever, just tell me to change it ;).

pkg/blobinfocache/prioritize/prioritize.go
wking: > I think no external user needs to care; move this into a `pkg/blobinfocache/internal` (or `pkg/blobinfocache/internal/prioritize`? I don’t have a opinion really).

Part of me thinks "external users might want access to these utilities when they write their own `BlobInfoCache` implementation" and part of me things "nobody else is ever going to write a `BlobInfoCache` implementation" ;).  So I don't have a strong opinion either; moved under `internal/` with 15dfbbc -> ca54d11 (which also rebases onto master).

pkg/blobinfocache/test/test.go
wking: > `pkg/blobinfocache/internal/test`, please.

Done with 15dfbbc -> ca54d11, some more discussion [here][1].

[1]: https://github.com/containers/image/pull/602#discussion_r264415662

pkg/blobinfocache/deprecated.go
wking: I've dropped the deprecated methods with ca54d11 -> 4cd979b.

directory/directory_test.go
mtrmac: I did say that it is absolutely non-blocking :)

Just something to consider, and somewhat a counterweight to Valentin’s naming concerns.

pkg/blobinfocache/boltdb/boltdb_test.go
```diff
@@ -1,30 +1,31 @@
...
 	"path/filepath"
 	"testing"
 
+	"github.com/containers/image/pkg/blobinfocache/internal/test"
 	"github.com/containers/image/types"
 	"github.com/stretchr/testify/require"
 )
 
-func newTestBoltDBCache(t *testing.T) (types.BlobInfoCache, func(t *testing.T)) {
+func newTest(t *testing.T) (types.BlobInfoCache, func(t *testing.T)) {
```
mtrmac: Rename this to `newCache` or `newTestCache`, please (or keep the original name).

pkg/blobinfocache/memory/memory.go
```diff
@@ -16,21 +18,21 @@ type locationKey struct {
...
 	uncompressedDigests   map[digest.Digest]digest.Digest
 	digestsByUncompressed map[digest.Digest]map[digest.Digest]struct{}             // stores a set of digests for each uncompressed digest
 	knownLocations        map[locationKey]map[types.BICLocationReference]time.Time // stores last known existence time for each location reference
 }
 
-// NewMemoryCache returns a BlobInfoCache implementation which is in-memory only.
+// New returns a BlobInfoCache implementation which is in-memory only.
 // This is primarily intended for tests, but also used as a fallback if DefaultCache
 // can’t determine, or set up, the location for a persistent cache.
 // Manual users of types.{ImageSource,ImageDestination} might also use this instead of a persistent cache.
```
mtrmac: Please add something like
```
// Most users should call DefaultCache instead.
```

pkg/blobinfocache/memory/memory_test.go
```diff
@@ -0,0 +1,16 @@
+package memory
+
+import (
+	"testing"
+
+	"github.com/containers/image/pkg/blobinfocache/internal/test"
+	"github.com/containers/image/types"
+)
+
+func newTest(t *testing.T) (types.BlobInfoCache, func(t *testing.T)) {
```
mtrmac: Rename this to `newCache` or `newTestCache`, please (or keep the original `newTestMemoryCache` name).

mtrmac:
> > In particular, I really don’t want users of this library choosing the memory cache just because they can avoid a dependency on BoltDB.
> 
> Nope, that's exactly why I worked this up ;).

I knew it! ;)

> > Do you have a special use case that really shouldn’t be using the persistent cache? (Sure, there could well be one; I just want to make sure that the dependency is not the only motivation.)
> 
> [openshift/installer#1286](https://github.com/openshift/installer/pull/1286) is fetching a release image, looking in that filesystem for a `release-manifests/image-references` file, fetching the referenced `machine-os-content` image, and looking at its annotations.
…
> Caching is nice, but in this case we're pulling some small manifests and one single, smallish layer:
> 
> ```
> $ ls -lh 7900974d6b5b9209c525f97f0e68d214168004d7082e9d3d45a0c9d1da473fbd.tar 
> -r--r--r--. 1 trking trking 471K Dec 31  1969 7900974d6b5b9209c525f97f0e68d214168004d7082e9d3d45a0c9d1da473fbd.tar
> ```
> 
> Caching seems like an excessive complication when we're that lightweight.

In principle, the cache could still help avoid re-downloading a differently-compressed version of that blob, but that could happen only if the consumer decompressing that layer explicitly used `RecordDigestUncompressedPair`, and besides, a single release process for a single coherent product is not all that likely to compress that blob in different ways.

Fair enough; please make sure all the individual constructors include
```go
// Most users should call DefaultCache instead.
```
(or something even more strongly worded) and I’ll merge this.


---

> I'm fine dropping the deprecated endpoints now if we have no compatibility requirements. Do we publish a change-log or something to help consumers migrate over breaking changes?

We haven’t been maintaining sufficiently detailed release notes so far.

pkg/blobinfocache/prioritize/prioritize.go
mtrmac: > Part of me thinks "external users might want access to these utilities when they write their own `BlobInfoCache` implementation" and part of me things "nobody else is ever going to write a `BlobInfoCache` implementation" ;)

Yes, both.  If anyone felt a strong need to write a different cache implementation, that would primarily be a reason improve the default one.

pkg/blobinfocache/boltdb/boltdb.go
```diff
@@ -1,4 +1,5 @@
-package blobinfocache
+// Package boldtb implements a BlobInfoCache backed by BoltDB.
```
mtrmac: Tests fail with:
```
/gopath/src/github.com/containers/image/pkg/blobinfocache/boltdb/boltdb.go:1:1: package comment should be of the form "Package boltdb ..."
```

`s/boldtb/boltdb/` it seems.

wking: > `s/boldtb/boltdb/` it seems.

Done with b2f8be9 -> d545a19.

pkg/blobinfocache/memory/memory_test.go
wking: > Rename this to `newCache` or `newTestCache`

`newTestCache` with b2f8be9 -> d545a19.

pkg/blobinfocache/memory/memory.go
wking: >     // Most users should call DefaultCache instead.

Done with b2f8be9 -> d545a19.

pkg/blobinfocache/boltdb/boltdb_test.go
wking: >     Rename this to `newCache` or `newTestCache`

`newTestCache` with b2f8be9 -> d545a19.

wking:
> Fair enough; please make sure all the individual constructors include
>
>     // Most users should call DefaultCache instead.

Done with b2f8be9 -> d545a19.

mtrmac:
Thanks!

PR #603, opened 2019-03-11
copy: progress bar: show messages on completion

vrothberg:
Show messages on completion of the progress bars:
```
Copying blob 6292a0d81fc0 completed
```
...when the copy operation is completed.

```
Copying blob 6292a0d81fc0 completed: already exists
```
...when the blob already exists in the local storage.

As we cannot update progress bars on the fly, we need to defer their
creation to the point we know if a copy operations must be executed
or nor (i.e., to copyLayer()).


vrothberg:
@rhatdan @mtrmac PTAL.
Context: https://github.com/containers/libpod/issues/2600

rhatdan:
LGTM

copy/copy.go
```diff
@@ -604,11 +598,12 @@ func (c *copier) createProgressBar(pool *mpb.Progress, info types.BlobInfo, kind
 	}
 
 	bar := pool.AddBar(info.Size,
+		mpb.BarClearOnComplete(),
 		mpb.PrependDecorators(
-			decor.Name(prefix),
+			decor.OnComplete(decor.Name(prefix), fmt.Sprintf("%s %s", prefix, onComplete)),
```
mtrmac: (With my past localization experience (which, I realize, has no relevance to what we are doing at all), I’m conceptually not thrilled about just appending a “completed” word to an existing sentence. Already the `"%s", kind` is pushing it.)

Also, it just seems conceptually cleaner to add the final status to the `AppendDecorators`: it used to show progress, and that is replaced by a “done” message. Removing the progress and moving “done” to the end of prefix seems more convoluted to me.

(AFAICS the progress bar just concatenates prefix and suffix, but I might be missing something.)

copy/copy.go
```diff
@@ -668,10 +663,15 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, ba
 		}
 		if reused {
 			logrus.Debugf("Skipping blob %s (already present):", srcInfo.Digest)
+			bar := ic.c.createProgressBar(pool, srcInfo, "blob", "completed: already exists")
+			bar.SetTotal(0, true)
```
mtrmac: `bar.SetTotal(srcInfo.Size, True)`, please, the behavior of `0` is undocumented.

copy/copy.go
```diff
@@ -668,10 +663,15 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, ba
...
 			return blobInfo, cachedDiffID, nil
 		}
 	}
 
+	bar := ic.c.createProgressBar(pool, srcInfo, "blob", "completed")
+	defer bar.SetTotal(srcInfo.Size, true)
```
mtrmac: Can we just do this on the success path? Marking the progress as “everything succeeded” on the error paths seems conceptually incorrect.

copy/copy.go
```diff
@@ -668,10 +663,15 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, ba
 		}
 		if reused {
 			logrus.Debugf("Skipping blob %s (already present):", srcInfo.Digest)
+			bar := ic.c.createProgressBar(pool, srcInfo, "blob", "completed: already exists")
```
mtrmac: Maybe “skipped” instead of “completed”?

copy/copy.go
mtrmac: … never mind, we have literally copied 0 bytes, so `0` is semantically correct.

(Although it does not actually set the total to 0… but we are removing the bar and the counters on completion anyway, so no-one can ever notice.)

copy/copy.go
vrothberg: Makes perfect sense, good catch!

copy/copy.go
vrothberg: Good catch!

copy/copy.go
vrothberg: I like the idea! Moving it to the `AppendDecorators` seems more idiomatic.

> [...] I’m conceptually not thrilled about just appending a “completed” word to an existing sentence [...]

Any suggestion? I couldn't find an alternative tbh.

copy/copy.go
vrothberg: Nice suggestion. "skipped" is closer to what's actually happening!

vrothberg:
> ACK overall, mostly minor comments.

Thanks for your comments, @mtrmac! I will push an update tomorrow morning.

copy/copy.go
mtrmac: (I don’t really have a suggestion WRT sentence pasting; moving the state to `AppendDecorator`… conceptually works, but the string on the screen ends up being a single sentence anyway.

A full l10n-safe fix would be not do do splitting at all, and have localizable strings like
- Copying blob $digest
- Copying blob $digest done
- Copying blob $digest skipped: already exists
and two more for configs.

We are not doing l10n at all, so this would be an absolute overkill right now. Let’s not worry about it.)

vrothberg:
@mtrmac please have another look

copy/copy.go
```diff
@@ -668,10 +663,15 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, ba
 		}
 		if reused {
 			logrus.Debugf("Skipping blob %s (already present):", srcInfo.Digest)
+			bar := ic.c.createProgressBar(pool, srcInfo, "blob", "skipped: already exists")
+			bar.SetTotal(0, true)
 			return blobInfo, cachedDiffID, nil
 		}
 	}
 
+	// XXX: call bar.SetTotal() only if err != nil
```
mtrmac: This seems to be a WIP comment

copy/copy.go
```diff
@@ -695,9 +695,11 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, ba
 			// This is safe because we have just computed diffIDResult.Digest ourselves, and in the process
 			// we have read all of the input blob, so srcInfo.Digest must have been validated by digestingReader.
 			ic.c.blobInfoCache.RecordDigestUncompressedPair(srcInfo.Digest, diffIDResult.digest)
+			bar.SetTotal(srcInfo.Size, true)
 			return blobInfo, diffIDResult.digest, nil
 		}
 	} else {
+		bar.SetTotal(srcInfo.Size, true)
```
mtrmac: (Non-blocking: It would be a tiny bit more readable to consolidate the success return paths again, so that this happens only in one place.)

vrothberg: I did that and removed the XXX indicator. Thanks!

mtrmac: Almost… there are still two `SetTotal` calls.

vrothberg: :joy_cat: omg, thanks for bearing with me. Updated.

mtrmac:
LGTM pending tests, again. Thanks!

rhatdan:
LGTM

PR #148, opened 2016-11-02
docker: add docker-archive transport

cyphar:
This transport is effectively just a copy of the docker-daemon://
transport but replacing all of the Docker client code with simple
filesystem code. It currently works as well as docker-daemon:// (from my
testing).

Further refactoring is necessary, this was just a first pass.

Implements #140.

Signed-off-by: Aleksa Sarai <asarai@suse.com>

PR #604, opened 2019-03-17
fix Bug 1688041 - podman image save removes existing image

QiWang19:
fix [Bug 1688041 - podman image save removes existing image ](https://bugzilla.redhat.com/show_bug.cgi?id=1688041)

docker doesn't return error `docker-archive:img1: docker-archive doesn't support modifying existing images`
`docker save` can replace the existing file.
`podman save` should overwrite the existing file.

### DIFF
--- a/docker/archive/dest.go
+++ b/docker/archive/dest.go
@@ -27,15 +27,6 @@ func newImageDestination(sys *types.SystemContext, ref archiveReference) (types.
 		return nil, errors.Wrapf(err, "error opening file %q", ref.path)
 	}
 
-	fhStat, err := fh.Stat()
-	if err != nil {
-		return nil, errors.Wrapf(err, "error statting file %q", ref.path)
-	}
-
-	if fhStat.Mode().IsRegular() && fhStat.Size() != 0 {
-		return nil, errors.New("docker-archive doesn't support modifying existing images")
-	}
-
 	tarDest := tarfile.NewDestination(fh, ref.destinationRef)
 	if sys != nil && sys.DockerArchiveAdditionalTags != nil {
 		tarDest.AddRepoTags(sys.DockerArchiveAdditionalTags)


rhatdan:
@QiWang19 Could you update the commit with a message similar to what you put into the git message, with spelling corrected.

QiWang19:
docker directly overwrite the existing image. Should I match with docker or make it as the expected result?

<https://github.com/containers/libpod/blob/41019f747280ba2c47df2ce6f83ddebfc2000745/libpod/image/image.go#L1155> removes the existing or created file no matter what kind of error occurs. I think another way to fix this is to not remove when the error is `docker-archive:img1: docker-archive doesn't support modifying existing images`

mtrmac:
> docker directly overwrite the existing image. Should I match with docker or make it as the expected result?

\*shrug\* I don’t have a strong opinion. @cyphar originally wrote that code in https://github.com/containers/image/pull/148 , he may have a preference.

(Noting also the `Path.wipe(filename)` in https://github.com/SUSE/kiwi/pull/1013/files#diff-9c91fd11fc96f62d60e3a950087f948fR109 I’ve just happened to notice; overwrite may well be useful. OTOH now that the code does not overwrite, changing it to overwrite may break something that relies on the current behavior.)

mtrmac:
> https://github.com/containers/libpod/blob/41019f747280ba2c47df2ce6f83ddebfc2000745/libpod/image/image.go#L1155 removes the existing or created file no matter what kind of error occurs. I think another way to fix this is to not remove when the error is `docker-archive:img1: docker-archive doesn't support modifying existing images`

… and when the error is “I/O error reading the input“ and when `getPolicyContext` fails and… … … many other possible failures that could happen somewhere on the call stack. It’s just not safe enough to delete files unless the code _knows_ that it is safe.

(It _might_ make sense for `docker-archive:` to delete the output files on error (i.e. if `Commit` is never called) if it _knows_ that it has created a new file. But it’s not _obviously_ the right thing to do, notably it could make it more difficult to diagnose the issue in some cases — and I don’t think deleting partial output files is _generally_ a common/assumed behavior.

See also https://github.com/containers/image/pull/313 which allows writing to a named pipe or a special device: such files should not be deleted on error in any case, and once data is sent to a stream, it can’t be unsent.)

mheon:
From the perspective of a consumer of the c/image API, I would really prefer that we not delete existing files here - if we end up deciding to go that course in Podman, we can detect that the file exists and remove it there, without locking every tool that uses c/image into that potentially undesirable behavior.

PR #607, opened 2019-03-20
Do not merge - Copy detection

vrothberg:
https://github.com/containers/storage/pull/307 must be merged before. Then we can update this PR and merge.

This PR is rather big and complex but I tried to cut it into easier to digest commits where possible (the final commit being the big chunk I couldn't figure out to split).

FYI @rhatdan @wking @mtrmac 

I will open a PR over a Skopeo that you can you use to play around.

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -43,6 +43,11 @@ type digestingReader struct {
 // downloads.  Let's follow Firefox by limiting it to 6.
 var maxParallelDownloads = 6
 
+// progressBarRefreshRate defines the refresh rate of the mpb progress bars.
+// 120 ms aligns with the initial upstream defaults.  Setting the rate here is
+// more idiomatic than using some magic number in the code below.
+var progressBarRefreshRate = 120 * time.Millisecond
+
 // newDigestingReader returns an io.Reader implementation with contents of source, which will eventually return a non-EOF error
 // or set validationSucceeded/validationFailed to true if the source stream does/does not match expectedDigest.
 // (neither is set if EOF is never reached).
@@ -467,11 +472,9 @@ func (ic *imageCopier) copyLayers(ctx context.Context) error {
 		copySemaphore = semaphore.NewWeighted(int64(1))
 	}
 
-	data := make([]copyLayerData, numLayers)
-	copyLayerHelper := func(index int, srcLayer types.BlobInfo, pool *mpb.Progress) {
+	copyLayerHelper := func(index int, srcLayer types.BlobInfo, pool *mpb.Progress, cld *copyLayerData) {
 		defer copySemaphore.Release(1)
 		defer copyGroup.Done()
-		cld := copyLayerData{}
 		if ic.c.dest.AcceptsForeignLayerURLs() && len(srcLayer.URLs) != 0 {
 			// DiffIDs are, currently, needed only when converting from schema1.
 			// In which case src.LayerInfos will not have URLs because schema1
@@ -485,16 +488,26 @@ func (ic *imageCopier) copyLayers(ctx context.Context) error {
 		} else {
 			cld.destInfo, cld.diffID, cld.err = ic.copyLayer(ctx, srcLayer, pool)
 		}
-		data[index] = cld
 	}
 
+	// digestToCopyData maps a digest to a corresponding copyLayerData and
+	// avoids to redundantly copy the same layer.
+	digestToCopyData := make(map[digest.Digest]*copyLayerData)
 	func() { // A scope for defer
 		progressPool, progressCleanup := ic.c.newProgressPool(ctx)
 		defer progressCleanup()
 
 		for i, srcLayer := range srcInfos {
+			cld, ok := digestToCopyData[srcLayer.Digest]
+			if ok {
+				// digest is already being copied
+				copyGroup.Done()
+				continue
+			}
 			copySemaphore.Acquire(ctx, 1)
-			go copyLayerHelper(i, srcLayer, progressPool)
+			cld = &copyLayerData{}
+			digestToCopyData[srcLayer.Digest] = cld
+			go copyLayerHelper(i, srcLayer, progressPool, cld)
 		}
 
 		// Wait for all layers to be copied
@@ -503,7 +516,11 @@ func (ic *imageCopier) copyLayers(ctx context.Context) error {
 
 	destInfos := make([]types.BlobInfo, numLayers)
 	diffIDs := make([]digest.Digest, numLayers)
-	for i, cld := range data {
+	for i, srcLayer := range srcInfos {
+		cld, ok := digestToCopyData[srcLayer.Digest]
+		if !ok {
+			return fmt.Errorf("no copy data found for layer %q", srcLayer.Digest)
+		}
 		if cld.err != nil {
 			return cld.err
 		}
@@ -577,16 +594,25 @@ func (ic *imageCopier) copyUpdatedConfigAndManifest(ctx context.Context) ([]byte
 // The caller must eventually call the returned cleanup function after the pool will no longer be updated.
 func (c *copier) newProgressPool(ctx context.Context) (*mpb.Progress, func()) {
 	ctx, cancel := context.WithCancel(ctx)
-	pool := mpb.New(mpb.WithWidth(40), mpb.WithOutput(c.progressOutput), mpb.WithContext(ctx))
+	pool := mpb.New(mpb.WithWidth(40), mpb.WithOutput(c.progressOutput), mpb.WithContext(ctx), mpb.WithRefreshRate(progressBarRefreshRate))
 	return pool, func() {
 		cancel()
 		pool.Wait()
 	}
 }
 
+// createReplacementProgressBar is wrapper around createProgressBar and creates
+// a progress bar to replace the `toReplace` bar.
+func (c *copier) createReplacementProgressBar(pool *mpb.Progress, toReplace *mpb.Bar, info types.BlobInfo, kind string, filler string, barOpts ...mpb.BarOption) *mpb.Bar {
+	barOpts = append(barOpts, mpb.BarReplaceOnComplete(toReplace))
+	bar := c.createProgressBar(pool, info, kind, filler, barOpts...)
+	toReplace.SetTotal(0, true)
+	return bar
+}
+
 // createProgressBar creates a mpb.Bar in pool.  Note that if the copier's reportWriter
 // is ioutil.Discard, the progress bar's output will be discarded
-func (c *copier) createProgressBar(pool *mpb.Progress, info types.BlobInfo, kind string, onComplete string) *mpb.Bar {
+func (c *copier) createProgressBar(pool *mpb.Progress, info types.BlobInfo, kind string, filler string, barOpts ...mpb.BarOption) *mpb.Bar {
 	// shortDigestLen is the length of the digest used for blobs.
 	const shortDigestLen = 12
 
@@ -597,15 +623,35 @@ func (c *copier) createProgressBar(pool *mpb.Progress, info types.BlobInfo, kind
 		prefix = prefix[:maxPrefixLen]
 	}
 
-	bar := pool.AddBar(info.Size,
-		mpb.BarClearOnComplete(),
+	options := []mpb.BarOption{
 		mpb.PrependDecorators(
 			decor.Name(prefix),
 		),
-		mpb.AppendDecorators(
-			decor.OnComplete(decor.CountersKibiByte("%.1f / %.1f"), " "+onComplete),
-		),
-	)
+	}
+	options = append(options, barOpts...)
+
+	// Fillers are used for placeholder bars that are later replaced by a
+	// common progress bar.  Hence, only append a counter decorator when
+	// filler == "".
+	if filler == "" {
+		options = append(options,
+			mpb.AppendDecorators(
+				decor.OnComplete(decor.CountersKibiByte("%.1f / %.1f"), " done"),
+			))
+		options = append(options, mpb.BarClearOnComplete())
+	}
+
+	var bar *mpb.Bar
+	if filler != "" {
+		barFiller := mpb.FillerFunc(
+			func(w io.Writer, width int, st *decor.Statistics) {
+				fmt.Fprint(w, filler)
+			})
+		bar = pool.Add(0, barFiller, options...)
+	} else {
+		bar = pool.AddBar(info.Size, options...)
+	}
+
 	if c.progressOutput == ioutil.Discard {
 		c.Printf("Copying %s %s\n", kind, info.Digest)
 	}
@@ -624,7 +670,7 @@ func (c *copier) copyConfig(ctx context.Context, src types.Image) error {
 		destInfo, err := func() (types.BlobInfo, error) { // A scope for defer
 			progressPool, progressCleanup := c.newProgressPool(ctx)
 			defer progressCleanup()
-			bar := c.createProgressBar(progressPool, srcInfo, "config", "done")
+			bar := c.createProgressBar(progressPool, srcInfo, "config", "")
 			destInfo, err := c.copyBlobFromStream(ctx, bytes.NewReader(configBlob), srcInfo, nil, false, true, bar)
 			if err != nil {
 				return types.BlobInfo{}, err
@@ -649,23 +695,119 @@ type diffIDResult struct {
 	err    error
 }
 
+// tryReusingBlob is wrapper around the destination's TryReusingBlob() method.
+// If the blob is present, it returns immediately.  Otherwise, if the
+// destination supports blob locking, it will acquire the blob lock and perform
+// another call to TryReusingBlob() to see if another process has already
+// copied the blob.
+func (ic *imageCopier) tryReusingBlob(ctx context.Context, srcInfo types.BlobInfo, pool *mpb.Progress, bar *mpb.Bar) (bool, types.BlobInfo, error) {
+	reused, blobInfo, err := ic.c.dest.TryReusingBlob(ctx, srcInfo, ic.c.blobInfoCache, ic.canSubstituteBlobs)
+	if err != nil {
+		return false, types.BlobInfo{}, errors.Wrapf(err, "Error trying to reuse blob %s at destination", srcInfo.Digest)
+	}
+	if reused {
+		logrus.Debugf("Skipping blob %s (already present):", srcInfo.Digest)
+		newBar := ic.c.createReplacementProgressBar(pool, bar, srcInfo, "blob", "skipped: already exists")
+		newBar.SetTotal(0, true)
+		return true, blobInfo, nil
+	}
+
+	if !ic.c.dest.SupportsBlobLocks() {
+		return false, types.BlobInfo{}, nil
+	}
+
+	// Check if another process might be copying the layer at the moment.
+	// If so, wait for it's completion and try to reuse the blob again.
+	logrus.Debugf("Checking if blob %s is already being copied by another process", srcInfo.Digest)
+
+	// LockBlob() does not support try-lock semantics which would be a
+	// straight-forward mechanism to detect if another process is currently
+	// copying a specific layer.
+	//
+	// The trick we are using here is to acquire the lock in a separate
+	// goroutine and use a channel for signaling when the lock has been
+	// acquired.  To detect if this goroutine is blocking on the lock, we
+	// spin up a second one sending a message via another channel after a
+	// time out.
+	chanDone := make(chan bool, 1)
+	chanBlockedByAnotherProcess := make(chan bool, 1)
+
+	var lockError error
+	// Acquire the blob lock
+	go func() {
+		lockError = ic.c.dest.LockBlob(srcInfo)
+		chanDone <- true
+	}()
+	// Spin up the timer
+	go func() {
+		// The initial sleep time is `progressBarRefreshRate`
+		// to make sure the bar is properly rendered before
+		// replacing it.
+		time.Sleep(progressBarRefreshRate)
+		chanBlockedByAnotherProcess <- true
+	}()
+
+	copiedByAnotherProcess := false
+	select {
+	case <-chanDone:
+		break
+	case <-chanBlockedByAnotherProcess:
+		logrus.Debugf("Blob %s is already being copied by another process", srcInfo.Digest)
+		copiedByAnotherProcess = true
+		bar = ic.c.createReplacementProgressBar(pool, bar, srcInfo, "blob", "paused: being copied by another process", mpb.BarRemoveOnComplete())
+		// Wait until we acquired the blob lock
+		<-chanDone
+	}
+
+	if lockError != nil {
+		return false, types.BlobInfo{}, errors.Wrapf(err, "Error acquiring lock for blob %s", srcInfo.Digest)
+	}
+
+	// If the blob has not been copied by another process, we need to copy
+	// it on our own and can return immediately.
+	if !copiedByAnotherProcess {
+		logrus.Debugf("Blob %s is not being copied by another process", srcInfo.Digest)
+		return false, types.BlobInfo{}, nil
+	}
+
+	// If the blob has been copied by another process, it
+	// must be present in the in the destintation and be
+	// available for reuse.
+	reused, blobInfo, err = ic.c.dest.TryReusingBlob(ctx, srcInfo, ic.c.blobInfoCache, ic.canSubstituteBlobs)
+	if err != nil {
+		return false, types.BlobInfo{}, errors.Wrapf(err, "Error trying to reuse blob %s at destination", srcInfo.Digest)
+	}
+
+	if reused {
+		ic.c.dest.UnlockBlob(srcInfo)
+		logrus.Debugf("Blob %s has been successfully copied by another process", srcInfo.Digest)
+		bar = ic.c.createReplacementProgressBar(pool, bar, srcInfo, "blob", "done: copied by another process")
+		bar.SetTotal(0, true)
+		return true, blobInfo, nil
+	}
+
+	logrus.Debugf("Blob %s is not present: copy operation of other process must have failed", srcInfo.Digest)
+	return false, types.BlobInfo{}, nil
+}
+
 // copyLayer copies a layer with srcInfo (with known Digest and possibly known Size) in src to dest, perhaps compressing it if canCompress,
 // and returns a complete blobInfo of the copied layer, and a value for LayerDiffIDs if diffIDIsNeeded
 func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, pool *mpb.Progress) (types.BlobInfo, digest.Digest, error) {
 	cachedDiffID := ic.c.blobInfoCache.UncompressedDigest(srcInfo.Digest) // May be ""
 	diffIDIsNeeded := ic.diffIDsAreNeeded && cachedDiffID == ""
 
+	// Create the initial placeholder bar which is later being replaced by
+	// another one.
+	bar := ic.c.createProgressBar(pool, srcInfo, "blob", " ", mpb.BarRemoveOnComplete())
+
 	// If we already have the blob, and we don't need to compute the diffID, then we don't need to read it from the source.
 	if !diffIDIsNeeded {
-		reused, blobInfo, err := ic.c.dest.TryReusingBlob(ctx, srcInfo, ic.c.blobInfoCache, ic.canSubstituteBlobs)
+		reused, blobInfo, err := ic.tryReusingBlob(ctx, srcInfo, pool, bar)
 		if err != nil {
-			return types.BlobInfo{}, "", errors.Wrapf(err, "Error trying to reuse blob %s at destination", srcInfo.Digest)
+			return types.BlobInfo{}, "", err
 		}
 		if reused {
-			logrus.Debugf("Skipping blob %s (already present):", srcInfo.Digest)
-			bar := ic.c.createProgressBar(pool, srcInfo, "blob", "skipped: already exists")
-			bar.SetTotal(0, true)
-			return blobInfo, cachedDiffID, nil
+			return blobInfo, cachedDiffID, err
 		}
 	}
 
@@ -676,7 +818,7 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, po
 	}
 	defer srcStream.Close()
 
-	bar := ic.c.createProgressBar(pool, srcInfo, "blob", "done")
+	bar = ic.c.createReplacementProgressBar(pool, bar, srcInfo, "blob", "")
 
 	blobInfo, diffIDChan, err := ic.copyLayerFromStream(ctx, srcStream, types.BlobInfo{Digest: srcInfo.Digest, Size: srcBlobSize}, diffIDIsNeeded, bar)
 	if err != nil {
--- a/storage/storage_image.go
+++ b/storage/storage_image.go
@@ -64,6 +64,11 @@ type storageImageDestination struct {
 	fileSizes      map[digest.Digest]int64         // Mapping from layer blobsums to their sizes
 	filenames      map[digest.Digest]string        // Mapping from layer blobsums to names of files we used to hold them
 	SignatureSizes []int                           `json:"signature-sizes,omitempty"` // List of sizes of each signature slice
+
+	// data for {Lock,Unlock}Blob
+	lockedDigests   map[digest.Digest]storage.Locker // Bookkeeping of digests that have been locked
+	unlockedDigests map[digest.Digest]bool           // Bookkeeping of digests that have been unlocked. Required to unlock yet unlocked digests in Close()
+	digestLockMutex sync.Mutex                       // Serializes accesses to lockedDigests
 }
 
 type storageImageCloser struct {
@@ -323,12 +328,14 @@ func newImageDestination(imageRef storageReference) (*storageImageDestination, e
 		return nil, errors.Wrapf(err, "error creating a temporary directory")
 	}
 	image := &storageImageDestination{
-		imageRef:       imageRef,
-		directory:      directory,
-		blobDiffIDs:    make(map[digest.Digest]digest.Digest),
-		fileSizes:      make(map[digest.Digest]int64),
-		filenames:      make(map[digest.Digest]string),
-		SignatureSizes: []int{},
+		imageRef:        imageRef,
+		directory:       directory,
+		blobDiffIDs:     make(map[digest.Digest]digest.Digest),
+		fileSizes:       make(map[digest.Digest]int64),
+		filenames:       make(map[digest.Digest]string),
+		lockedDigests:   make(map[digest.Digest]storage.Locker),
+		unlockedDigests: make(map[digest.Digest]bool),
+		SignatureSizes:  []int{},
 	}
 	return image, nil
 }
@@ -339,8 +346,14 @@ func (s *storageImageDestination) Reference() types.ImageReference {
 	return s.imageRef
 }
 
-// Close cleans up the temporary directory.
+// Close cleans up the temporary directory and unlocks all unlocked digest locks and removes them from `lockedDigests`.
 func (s *storageImageDestination) Close() error {
+	// Make sure that all lockes are unlocked to avoid potential deadlocks.
+	for digest, unlocked := range s.unlockedDigests {
+		if !unlocked {
+			s.unlockDigest(digest)
+		}
+	}
 	return os.RemoveAll(s.directory)
 }
 
@@ -362,23 +375,54 @@ func (s *storageImageDestination) HasThreadSafePutBlob() bool {
 
 // SupportsBlobLocks indicates whether the ImageDestination supports blob
 // locking.
-func (d *storageImageDestination) SupportsBlobLocks() bool {
-	return false
+func (s *storageImageDestination) SupportsBlobLocks() bool {
+	return true
 }
 
 // LockBlob can be used to synchronize operations on it (e.g., copying).
+// Note that any unlocked lock will be unlocked in Close().
 func (s *storageImageDestination) LockBlob(b types.BlobInfo) error {
-	// NOOP for this type.
+	locker, err := func() (storage.Locker, error) {
+		// anonymous function defer the Unlock() and not worry about
+		// error paths
+		s.digestLockMutex.Lock()
+		defer s.digestLockMutex.Unlock()
+		if locker, ok := s.lockedDigests[b.Digest]; ok {
+			return locker, nil
+		}
+		locker, err := s.imageRef.transport.store.GetDigestLock(b.Digest)
+		if err != nil {
+			return nil, err
+		}
+		s.lockedDigests[b.Digest] = locker
+		return locker, nil
+	}()
+	if err != nil {
+		return err
+	}
+	locker.Lock()
+	s.digestLockMutex.Lock()
+	s.unlockedDigests[b.Digest] = false
+	s.digestLockMutex.Unlock()
 	return nil
 }
 
-// UnlockBlob unlocks the blob.  Note that it is safe to call UnlockBlob()
-// multiple times.  Only the first call is unlocking the blob.  This is
-// required to unlock a blob in the presence of errors or panics during copy
-// operations.
-func (s *storageImageDestination) UnlockBlob(b types.BlobInfo) error {
-	// NOOP for this type.
+func (s *storageImageDestination) unlockDigest(d digest.Digest) error {
+	s.digestLockMutex.Lock()
+	defer s.digestLockMutex.Unlock()
+
+	locker, ok := s.lockedDigests[d]
+	if !ok {
+		return errors.Errorf("trying to unlock non existent lock for digest %q", d)
+	}
+	locker.Unlock()
+	s.unlockedDigests[d] = true
 	return nil
+}
+
+// UnlockBlob unlocks the blob.
+func (s *storageImageDestination) UnlockBlob(b types.BlobInfo) error {
+	return s.unlockDigest(b.Digest)
 }
 
 // PutBlob writes contents of stream and returns data representing the result.
--- a/storage/storage_image.go
+++ b/storage/storage_image.go
@@ -360,6 +360,27 @@ func (s *storageImageDestination) HasThreadSafePutBlob() bool {
 	return true
 }
 
+// SupportsBlobLocks indicates whether the ImageDestination supports blob
+// locking.
+func (d *storageImageDestination) SupportsBlobLocks() bool {
+	return false
+}
+
+// LockBlob can be used to synchronize operations on it (e.g., copying).
+func (s *storageImageDestination) LockBlob(b types.BlobInfo) error {
+	// NOOP for this type.
+	return nil
+}
+
+// UnlockBlob unlocks the blob.  Note that it is safe to call UnlockBlob()
+// multiple times.  Only the first call is unlocking the blob.  This is
+// required to unlock a blob in the presence of errors or panics during copy
+// operations.
+func (s *storageImageDestination) UnlockBlob(b types.BlobInfo) error {
+	// NOOP for this type.
+	return nil
+}
+
 // PutBlob writes contents of stream and returns data representing the result.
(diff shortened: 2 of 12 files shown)


copy/copy.go
```diff
@@ -655,17 +696,108 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, po
...
+				ic.c.dest.LockBlob(srcInfo)
+				acquiredBlobLock = true
+			}()
+
+			if !acquiredBlobLock {
+				// The initial sleep time is `progressBarRefreshRate`
+				// to make sure the bar is properly rendered before
+				// replacing it.
+				time.Sleep(progressBarRefreshRate)
+				if !acquiredBlobLock {
```
mtrmac: Is it at all specified what happens when reading a value concurrently with a writer in a different goroutine, without any synchronization? I can’t find anything to that effect skimming https://golang.org/ref/mem .

copy/copy.go
```diff
@@ -655,17 +696,108 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, po
...
+			// the blob again.
+			//
+			logrus.Debugf("Checking if blob %s is already being copied by another process", srcInfo.Digest)
+
+			// LockBlob() does not support try-lock semantics which would
+			// be a straight-forward mechanism to detect if another process
+			// is currently copying a specific layer.  However, with
+			// LockBlob() we are constrained by the file locks in
+			// containers/storage which cannot support try-lock semantics
+			// as fcntl(2) does not support that.
```
mtrmac: `F_SETLK` vs. `F_SETLKW` does support that AFAICS. (The `sync.Mutex` objects used in `c/storage.lockfile` seemingly don’t, OTOH.)

copy/copy.go
```diff
@@ -655,17 +696,108 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, po
...
+			// separate goroutine, which sets a boolean signal-variable as
+			// soon as the lock is acquired.  The main-routine is then
+			// sleeping for a certain amount of time.  If the lock has not
+			// been acquired until then, the progress bar will be updated
+			// with a message indicating that the blob is being copied by
+			// another process and wait until the signal-variable is set
+			// true.
+			acquiredBlobLock := false
+			go func() {
+				ic.c.dest.LockBlob(srcInfo)
```
mtrmac: This looks fairly risky; e.g. is there anything that guarantees that `LockBlob` will be called before `ic.c.dest.Close()` at all?

… well, the code below does do that, only if nothing panics.

copy/copy.go
```diff
@@ -655,17 +696,108 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, po
...
+	bar := ic.c.createProgressBar(pool, srcInfo, "blob", " ", mpb.BarRemoveOnComplete())
+
 	// If we already have the blob, and we don't need to compute the diffID, then we don't need to read it from the source.
 	if !diffIDIsNeeded {
-		reused, blobInfo, err := ic.c.dest.TryReusingBlob(ctx, srcInfo, ic.c.blobInfoCache, ic.canSubstituteBlobs)
+		// tryReusingBlob is an anonymous function to avoid adding
+		// another method with a complex parameter list and semantics.
+		// It's sole purpose is to simplify the control with an early
+		// return if the image destination does not blob locking.
+		tryReusingBlob := func() (bool, types.BlobInfo, digest.Digest, error) {
```
mtrmac: (My initial impression is: can’t this all happen inside `storageImageDestination.TryReusingBlob`?

(Consider also the mirroring code and the possible need to report progress with trying mirrors from `GetBlob` in the progress bars.)

OTOH I haven’t actually read any of the code at all, yet; this may well be the right design.)

storage/storage_image.go
```diff
@@ -360,6 +361,22 @@ func (s *storageImageDestination) HasThreadSafePutBlob() bool {
 	return true
 }
 
+// SupportsBlobLocks indicates whether the ImageDestination supports blob
+// locking.
+func (d *storageImageDestination) SupportsBlobLocks() bool {
+	return false
```
mtrmac: Well :)

Maybe this function is not necessary at all, after all, if this went through unnoticed.

copy/copy.go
nalind: Looking at https://github.com/containers/storage/pull/307, this creates a new file under the storage root's new "$driver-locks" subdirectory, named using the digest of the blob, and keeps it open until the calling process exits.
How do callers that are using this storage root keep the number of digest lock files from growing without bound?  How does a long-lived caller like CRI-O avoid tying up an open file descriptor, until it exits, for each blob that it pulls down from a registry?

storage/storage_image.go
```diff
@@ -674,16 +695,19 @@ func (s *storageImageDestination) Commit(ctx context.Context) error {
...
 		defer file.Close()
 		// Build the new layer using the diff, regardless of where it came from.
 		// TODO: This can take quite some time, and should ideally be cancellable using ctx.Done().
 		layer, _, err := s.imageRef.transport.store.PutLayer(id, lastLayer, nil, "", false, nil, file)
 		if err != nil && errors.Cause(err) != storage.ErrDuplicateID {
+			s.UnlockBlob(blob.BlobInfo)
 			return errors.Wrapf(err, "error adding layer with blob %q", blob.Digest)
 		}
 		lastLayer = layer.ID
+		s.UnlockBlob(blob.BlobInfo)
```
mtrmac: Way too many of these; can the code be restructured so that there is only one `UnlockBlob` and no-one forgets to add another one in the future?

(As a first impression, it’s a bit troubling that this is unlocking something locked by a quite different layer in the call stack.)

storage/storage_image.go
```diff
@@ -360,6 +361,22 @@ func (s *storageImageDestination) HasThreadSafePutBlob() bool {
...
+func (d *storageImageDestination) SupportsBlobLocks() bool {
+	return false
+}
+
+// NOOP for this type.
+func (s *storageImageDestination) LockBlob(b types.BlobInfo) error {
+	return s.imageRef.transport.store.LockDigest(b.Digest)
+}
+
+// NOOP for this type.
```
mtrmac: (This is not a documentation that would help the caller; if it should exist, it probably should be inside the function body, so that `go doc` does not generate it.)

… besides being incorrect in this case. If you are going to copy&paste something everywhere, copy&paste the interface documentation.

copy/copy.go
```diff
@@ -458,18 +484,13 @@ func (ic *imageCopier) copyLayers(ctx context.Context) error {
...
-	// avoid malicious images causing troubles and to be nice to servers.
-	var copySemaphore *semaphore.Weighted
-	if ic.c.copyInParallel {
-		copySemaphore = semaphore.NewWeighted(int64(maxParallelDownloads))
-	} else {
-		copySemaphore = semaphore.NewWeighted(int64(1))
-	}
+	// sem reduces the number of parallel goroutines and protects
+	// against malicious images with many layers.
+	sem := semaphore.NewWeighted(24)
```
mtrmac: (Make this a top-level constant, please.)

copy/copy.go
```diff
@@ -655,17 +696,108 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, po
...
+					// Note that we call SetTotal() after the new
+					// bar is created to avoid potential flickering
+					// if the bar removal and replacement happens
+					// across refresh rates.
+					toReplaceBar := bar
+					bar = ic.c.createProgressBar(pool, srcInfo, "blob", "stopped: copied by another process", mpb.BarRemoveOnComplete(), mpb.BarReplaceOnComplete(bar))
+					toReplaceBar.SetTotal(0, true)
+
+					for !acquiredBlobLock {
+						time.Sleep(50 * time.Millisecond)
```
mtrmac: Can’t this use a channel to block instead?

That would also allow:
```go
lockingDoneChan := startALockingGoroutine()
weAreProbablyBlockedChan := startATimerForRefreshInterval
select {
case <- lockingDoneChan: // We are the only one, continue as usual:
case <- weAreProbablyBlockedChan: // We are blocked
    // tell the user
    <- lockingDoneChan // Block on the lock
}
// either way, lockingDoneChan has been waited on here
// (FIXME: do we need to clean up weAreProbablyBlockedChan and the timer? channels can be tricky.)
```

which does not require waiting for the whole refresh interval if we can acquire the lock faster.

copy/copy.go
```diff
@@ -309,6 +323,18 @@ func (c *copier) copyOneImage(ctx context.Context, policyContext *signature.Poli
...
 
+	// Unblock all blobs as a precation to avoid potential deadlocks in the
+	// presence of errors or even panics.  Note that it is safe to unblock
+	// here as locks are only unblocked if we're the owner of the lock;
+	// only the first call to UnlockBlob() will effectively unlock the file.
+	if ic.c.dest.SupportsBlobLocks() {
+		defer func() {
+			srcInfos := ic.src.LayerInfos()
+			for _, blob := range srcInfos {
+				ic.c.dest.UnlockBlob(blob)
```
mtrmac: Why isn’t every layer unlocked as soon as that one blob is copied? 

(… because only `storageImageDestination.Commit` actually moves creates a c/storage layer from the temporary file, if I’m not mistaken. Maybe that should be changed first? Maybe not. At the very least, it should be explicitly and clearly documented why there isn’t a single function / level of the call stack both locking and unlocking.)

storage/storage_image.go
vrothberg: Lol. No, that was stupid mistake on my side (see https://github.com/containers/skopeo/pull/619#issuecomment-475138951). This was the last optimization to the code, required some rebasing and I missed committing the last change to the PR -.- Apologies for the confusion.

copy/copy.go
vrothberg: Exactly. Committing the layer happens at a different part of the stack and I am more than afraid of the complexity changing that. It's definitely possible to change the way we commit but this would imply rewriting a huge amount of code.

I had the very same thought when thinking about that and from an architectural point of view I'd prefer to have those tasks tied together but I currently don't have the time to refactor that.

vrothberg: I will update the comments accordingly and be verbose about that. Thanks for bringing that up!

copy/copy.go
vrothberg: Sure, we can do that. Usually, I am hesitant using channels as they seem to be a common source for bugs but this case seems small enough to brain-parse the code.

storage/storage_image.go
vrothberg: > Way too many of these; can the code be restructured so that there is only one `UnlockBlob` and no-one forgets to add another one in the future?

Sounds good. I will update it.

> (As a first impression, it’s a bit troubling that this is unlocking something locked by a quite different layer in the call stack.)

I agree. But without rewriting how we're committing, that's not possible (see https://github.com/containers/image/pull/607#discussion_r267653934).

copy/copy.go
vrothberg: I think it's possible to extend the `TryReusingBlob` interface with a channel to signal if we're waiting for another process to finish copying. I will look into it.

copy/copy.go
vrothberg: The manpage states: _"If a conflicting lock is held by another process, this call returns -1 and sets errno to EACCES or EAGAIN."_

This means that we only get a _EGAIN_ if we change the type of the lock (reader <-> writer). Here we need a write-lock and hence never change the type. In this case, it would block every time.

copy/copy.go
vrothberg: A bool should be safe but it's definitely a pain for the caches. I will follow your suggestion of using channels which are designed for such a use case.

copy/copy.go
vrothberg: The idea is to have per-digest-lock. The locks have a size of 0 as we're only syncing on the file descriptor. I agree that this is a very ugly part of the current state of the PR but I don't yet have answer to that.

vrothberg: Ideally, an unlock would also delete the file but I didn't yet find a solution to avoid a race for this case.

storage/storage_image.go
vrothberg: Done.

copy/copy.go
vrothberg: Done. I made it consistent with the copySemaphore. I renamed them the fields to be more expressive about their use.

copy/copy.go
vrothberg: The code currently does it but I can add a check to LockBlob() if you want to.

storage/storage_image.go
vrothberg: This has been changed. Now, there are three parts where a previously locked blob is unlocked:

1. When we detect that the other process has successfully copied the blob.
2. When we stored the layer in containers/storage (during Commit()).
3. When we call `dest.Close()` which seems an idiomatic way of cleaning up the resources (including locks).

copy/copy.go
vrothberg: This is still outstanding.

copy/copy.go
vrothberg: This has been changed as remaining locks are now unlocked in `dest.Close()` while the storage destination keeps track of which blobs/digests have actually been locked.

copy/copy.go
vrothberg: We're now doing this via semaphore which has a `TryAcquire`.

copy/copy.go
vrothberg: This remained unchanged.

copy/copy.go
vrothberg: This is now being done via semaphore.

copy/copy.go
vrothberg: (for the record)
There are two issues that we address as follows:

1. Accumulating lock files in the storage: we address it by moving them into the tmp storage which won't persist reboots.
2. Too many open FDs: we'll extend the lockfiles in containers/storage to be able to close the file descriptors and perform reference counting. The last reference will close the FD. This will make long-living users such as CRI-O happier.

vrothberg: Now fixed.

copy/copy.go
```diff
@@ -664,30 +689,123 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, po
...
+					// block until we acquired the blob lock
+					blobLockSem.Acquire(ctx, 1)
+				}
+			}
+
+			if lockError != nil {
+				return false, types.BlobInfo{}, "", errors.Wrapf(err, "Error acquiring lock for blob %s", srcInfo.Digest)
+			}
+
+			reused, blobInfo, err = ic.c.dest.TryReusingBlob(ctx, srcInfo, ic.c.blobInfoCache, ic.canSubstituteBlobs)
```
vrothberg: Add another comment why we're calling `TryReusingBlob()` again.

copy/copy.go
```diff
@@ -669,6 +667,8 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, po
 		}
 	}
 
+	//
+
```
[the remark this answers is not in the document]
vrothberg: Done

copy/copy.go
vrothberg: Actually, we shouldn't just fall through but record in a variable if another process copied the blob.

vrothberg: done

vrothberg:
@rhatdan @nalind @mtrmac PTAL

giuseppe:
LGTM

storage/storage_image.go
```diff
@@ -360,6 +368,57 @@ func (s *storageImageDestination) HasThreadSafePutBlob() bool {
...
+
+func (s *storageImageDestination) unlockDigest(d digest.Digest) error {
+	s.digestLockMutex.Lock()
+	defer s.digestLockMutex.Unlock()
+
+	locker, ok := s.lockedDigests[d]
+	if !ok {
+		return errors.Errorf("trying to unlock non existent lock for digest %q", d)
+	}
+	defer delete(s.lockedDigests, d)
```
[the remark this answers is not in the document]
vrothberg: Good catch. I think the defer could actually cause a bug here. We need to delete within the critical section guarded by the mutex. Thanks!

vrothberg:
> one question and a nit

Fixed both. Thanks, @giuseppe !

storage/storage_image.go
```diff
@@ -54,16 +54,19 @@ type storageImageSource struct {
...
+	directory       string                           // Temporary directory where we store blobs until Commit() time
+	nextTempFileID  int32                            // A counter that we use for computing filenames to assign to blobs
+	manifest        []byte                           // Manifest contents, temporary
+	signatures      []byte                           // Signature contents, temporary
+	putBlobMutex    sync.Mutex                       // Mutex to sync state for parallel PutBlob executions
+	blobDiffIDs     map[digest.Digest]digest.Digest  // Mapping from layer blobsums to their corresponding DiffIDs
+	fileSizes       map[digest.Digest]int64          // Mapping from layer blobsums to their sizes
+	filenames       map[digest.Digest]string         // Mapping from layer blobsums to names of files we used to hold them
+	SignatureSizes  []int                            `json:"signature-sizes,omitempty"` // List of sizes of each signature slice
+	lock            storage.Locker                   // Lockfile for the blob locking
```
mtrmac: Is this used anywhere?

vrothberg: Good catch. No, that's an artifact I missed to remove.

vrothberg:
@mtrmac can you have a look at the latest commit I added? It's passing Buildah and Podman CI, and the code looks fairly straight forward.

Rather sooner than later, I would love to allocate some time and reflect how we can refactor the code. `PutBlob()` and `Commit()` could be moved into a single operation. I did not get into the details of how this can be done but seeing all the potentials of deadlocks, I am worried (not yet scared).

mtrmac:
> Rather sooner than later, I would love to allocate some time and reflect how we can refactor the code. `PutBlob()` and `Commit()` could be moved into a single operation.

In general, or specifically for the way c/storage does this? In general, all the `PutBlob`s need to be finished before the manifest can be updated and written to the destination (incl. possibly attempting more than one manifest format if the destination protocol like d/distribution can support it in principle, but that particular server does not accept some formats), so I don’t know how that would work. OTOH the code can certainly be structured in some other way.

copy/copy.go
mtrmac: > This means that we only get a EGAIN if we change the type of the lock (reader <-> writer).

FWIW, I don’t think that’s what the man page means; see elsewhere in there
> Conflicting lock combinations (i.e., a read lock and a write lock or two write locks) 

(admittedly in a quite different section).

Compare also fcntl(3p)
> If a shared or exclusive lock cannot be set, fcntl() shall return immediately with a return value of −1.

or (info libc)
> [F_SETLKW] is just like the ‘F_SETLK’ command, but causes the process to block (or wait) until the request can be specified.

---

Anyway, as long as the in-process `rwMutex` does not support the same operation, this does not change much — only maybe the comment could be corrected.

storage/storage_image.go
```diff
@@ -360,6 +367,57 @@ func (s *storageImageDestination) HasThreadSafePutBlob() bool {
...
+// locking.
+func (s *storageImageDestination) SupportsBlobLocks() bool {
+	return true
+}
+
+// LockBlob can be used to synchronize operations on it (e.g., copying).
+// Note that any unlocked lock will be unlocked in Close().
+func (s *storageImageDestination) LockBlob(b types.BlobInfo) error {
+	// anonymous function to not worry about errors paths and panics
+	getLocker := func() (storage.Locker, error) {
```
mtrmac: (_Absolutely_ non-blocking:

1. consider
    ```go
    locker, err := func() (storage.Locker, error) {
       …
    }()
    ```
    I am honestly not sure whether it is more readable — eliminating the one-use identifier is a plus, OTOH one needs to read up to the end of the function definition to see what the code does with the lambda, in either case.
2. ”to not worry about error paths“ is a bit unclear about what the worry is exactly; if you mean “to ensure `s.digestLockMutex` is unlocked on all paths”, say so.

)

storage/storage_image.go
```diff
@@ -360,6 +367,57 @@ func (s *storageImageDestination) HasThreadSafePutBlob() bool {
...
+		locker, err := s.imageRef.transport.store.GetDigestLock(b.Digest)
+		if err != nil {
+			return locker, err
+		}
+		s.lockedDigests[b.Digest] = locker
+		return locker, nil
+	}
+
+	locker, err := getLocker()
+	if err != nil {
```
mtrmac: FIXME:

At this point, `s.digestLockMutex` is unlocked; so, this thread can be suspended, and some other thread can concurrently call _both_ `LockBlob` and `UnlockBlob` before this thread resumes; that `UnlockBlob` will remove `locker` from `s.lockedDigests`; then this thread could resume, call `locker.Lock()`, and we will end up with a locked `locker` that is not present in `s.lockedDigests`.

copy/copy.go
```diff
@@ -771,6 +948,8 @@ func computeDiffID(stream io.Reader, decompressor compression.DecompressorFunc)
 func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, srcInfo types.BlobInfo,
 	getOriginalLayerCopyWriter func(decompressor compression.DecompressorFunc) io.Writer,
 	canModifyBlob bool, isConfig bool, bar *mpb.Bar) (types.BlobInfo, error) {
+	c.semMaxParallelCopies.Acquire(ctx, 1)
```
mtrmac: Moving the enforcement of `copyInParallel` all the way down here does not work:
- `GetBlob` has already been called, in parallel, even if `!HasThreadSafeGetBlob`
- Similarly, `TryReusingBlob` has been called, in parallel; that’s usually not safe. (Yes, `HasThreadSafePutBlob` is documented incorrectly.)

Sure, it could be a good idea to allow more goroutines and fewer actual parallel data transfers, _if_ parallelism is supported; but that should _both_ be conditional on that condition, and either way it’s quite independent from the purpose of this PR AFAICS.

storage/storage_image.go
mtrmac: I can’t see the code for 2. in `Commit()` anywhere.

(I was thinking that the body loop in `Commit()` could be split into a separate function, with a single `defer s.UnlockBlob()` or its current equivalent.)

OTOH, sure, as long as `Close` unlocks the locks, this works fine; unlocking during `Commit` could allow the waiter to finish earlier if the two processes shared only a few base layers.

types/types.go
```diff
@@ -287,6 +287,15 @@ type ImageDestination interface {
 	// - Uploaded data MAY be visible to others before Commit() is called
 	// - Uploaded data MAY be removed or MAY remain around if Close() is called without Commit() (i.e. rollback is allowed but not guaranteed)
 	Commit(ctx context.Context) error
+	// SupportsBlobLocks indicates whether the ImageDestination supports blob locking.
+	SupportsBlobLocks() bool
+	// LockBlob can be used to synchronize operations on it (e.g., copying).
```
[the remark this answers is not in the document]
vrothberg: Done.

storage/storage_image.go
```diff
@@ -54,16 +54,18 @@ type storageImageSource struct {
...
+	nextTempFileID  int32                            // A counter that we use for computing filenames to assign to blobs
+	manifest        []byte                           // Manifest contents, temporary
+	signatures      []byte                           // Signature contents, temporary
+	putBlobMutex    sync.Mutex                       // Mutex to sync state for parallel PutBlob executions
+	blobDiffIDs     map[digest.Digest]digest.Digest  // Mapping from layer blobsums to their corresponding DiffIDs
+	fileSizes       map[digest.Digest]int64          // Mapping from layer blobsums to their sizes
+	filenames       map[digest.Digest]string         // Mapping from layer blobsums to names of files we used to hold them
+	SignatureSizes  []int                            `json:"signature-sizes,omitempty"` // List of sizes of each signature slice
+	lockedDigests   map[digest.Digest]storage.Locker // Bookkeeping of locked digests that have not yet been unlocked
+	digestLockMutex sync.Mutex                       // Serializes accesses to lockedDigests
```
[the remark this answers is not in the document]
vrothberg: Done.

storage/storage_image.go
```diff
@@ -360,6 +367,57 @@ func (s *storageImageDestination) HasThreadSafePutBlob() bool {
...
+	// anonymous function to not worry about errors paths and panics
+	getLocker := func() (storage.Locker, error) {
+		s.digestLockMutex.Lock()
+		defer s.digestLockMutex.Unlock()
+		if locker, ok := s.lockedDigests[b.Digest]; ok {
+			return locker, nil
+		}
+		locker, err := s.imageRef.transport.store.GetDigestLock(b.Digest)
+		if err != nil {
+			return locker, err
```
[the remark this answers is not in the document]
vrothberg: I tried that before but `storage.Locker` is an interface on which we can't use pointers.

storage/storage_image.go
vrothberg: Nice idea! Done.

mtrmac:
Oops, I have somehow submitted an incomplete review.  I’m still reading the code, and I planned to revisit the FIXMEs afterwards.

storage/storage_image.go
vrothberg: Good catch!

I solved it by removing the entries from `s.lockedDigests` only in `Close()`.

copy/copy.go
```diff
@@ -485,20 +497,58 @@ func (ic *imageCopier) copyLayers(ctx context.Context) error {
...
+			idx, ok := digestToIndex[srcInfos[i].Digest]
+			if !ok {
+				// Record an error if we have no entry in data
+				// AND no recorded index in digestToIndex.
+				data[i] = &copyLayerData{
+					err: errors.Errorf("copy data for blob %q should be cached but is not", srcInfos[i].Digest.String()),
+				}
+				continue
+			}
+			data[i] = data[idx]
```
mtrmac: This is quite a bit of code only to keep the original semantics of `data`.  Why not make `data` a `map[digest.Digest]*copyLayerData` directly, and consume it directly below when building `destInfos` etc.?

That might require allocating a new `copyLayerData`, and setting `data[srcInfos[I].Digest]` in this goroutine, outside of `copyLayerHelper`, and passing the `copyLayerData` into the helper — but that does not look all that bad, it’s been non-obvious that the `data[index] = &cld` from concurrent goroutines is safe.

copy/copy.go
vrothberg: I'll delete the commits from the PR.

vrothberg:
> Oops, I have somehow submitted an incomplete review. I’m still reading the code, and I planned to revisit the FIXMEs afterwards.

Let me know when you're done. I will repush after that.

copy/copy.go
```diff
@@ -655,17 +725,122 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, po
...
+		// return if the image destination does not blob locking.
+		tryReusingBlob := func() (bool, types.BlobInfo, digest.Digest, error) {
+			reused, blobInfo, err := ic.c.dest.TryReusingBlob(ctx, srcInfo, ic.c.blobInfoCache, ic.canSubstituteBlobs)
+			if err != nil {
+				return false, types.BlobInfo{}, "", errors.Wrapf(err, "Error trying to reuse blob %s at destination", srcInfo.Digest)
+			}
+			if reused {
+				logrus.Debugf("Skipping blob %s (already present):", srcInfo.Digest)
+				bar.SetTotal(0, true)
+				ic.c.createProgressBar(pool, srcInfo, "blob", "skipped: already exists", mpb.BarReplaceOnComplete(bar))
```
mtrmac: Does this bar ever complete? If not, how does `pool.Wait()` eventually stop waiting?

copy/copy.go
```diff
@@ -655,17 +725,122 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, po
...
+				ic.c.createProgressBar(pool, srcInfo, "blob", "done: copied by another process", mpb.BarReplaceOnComplete(bar))
+				toReplaceBar.SetTotal(0, true)
+				return true, blobInfo, cachedDiffID, nil
+			}
+
+			logrus.Debugf("Blob %s is not present: copy operation of other process must have failed", srcInfo.Digest)
+			return false, types.BlobInfo{}, "", nil
+		}
+
+		reused, blob, diff, err := tryReusingBlob()
```
mtrmac: `diff` is, if it is used, always `cachedDiffID`; and in the “reuse” path that’s logically always the case (we are not reading / re-computing the value); so it does not have to be a return value from this function.

(Consider also `s/blob/blobInfo/`.)

copy/copy.go
```diff
@@ -655,17 +725,122 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, po
...
+				lockError = ic.c.dest.LockBlob(srcInfo)
+				blobLockSem.Release(1)
+			}()
+
+			copiedByAnotherProcess := false
+			if !blobLockSem.TryAcquire(1) {
+				// The initial sleep time is `progressBarRefreshRate`
+				// to make sure the bar is properly rendered before
+				// replacing it.
+				time.Sleep(progressBarRefreshRate)
```
mtrmac: (I somehow can’t reply in previous threads?)

I’d still prefer a `select` instead of having to wait here.

copy/copy.go
```diff
@@ -655,17 +725,122 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, po
...
+				if !blobLockSem.TryAcquire(1) {
+					logrus.Debugf("Blob %s is already being copied by another process", srcInfo.Digest)
+					copiedByAnotherProcess = true
+					// If we still don't have the lock, replace the
+					// initial progress bar with a new one
+					// indicating that the blob is copied by
+					// another process.
+					//
+					// Note that we call SetTotal() after the new
+					// bar is created to avoid potential flickering
```
mtrmac: (This feels like something that could be encapsulated in a helper; `ic.c.createReplacementProgressBar`?)

copy/copy.go
```diff
@@ -655,17 +725,122 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, po
...
+					// if the bar removal and replacement happens
+					// across refresh rates.
+					toReplaceBar := bar
+					bar = ic.c.createProgressBar(pool, srcInfo, "blob", "paused: being copied by another process", mpb.BarRemoveOnComplete(), mpb.BarReplaceOnComplete(bar))
```
mtrmac: Is `bar` ever completed? If not, how does `pool.Wait` finish?

copy/copy.go
```diff
@@ -655,17 +725,122 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, po
...
+			reused, blobInfo, err = ic.c.dest.TryReusingBlob(ctx, srcInfo, ic.c.blobInfoCache, ic.canSubstituteBlobs)
+			if err != nil {
+				return false, types.BlobInfo{}, "", errors.Wrapf(err, "Error trying to reuse blob %s at destination", srcInfo.Digest)
+			}
+
+			if reused {
+				ic.c.dest.UnlockBlob(srcInfo)
+				logrus.Debugf("Blob %s has been successfully copied by another process", srcInfo.Digest)
+				toReplaceBar := bar
+				ic.c.createProgressBar(pool, srcInfo, "blob", "done: copied by another process", mpb.BarReplaceOnComplete(bar))
```
mtrmac: Is the return value ever completed? If not, how does `pool.Wait` finish?

copy/copy.go
```diff
@@ -655,17 +725,122 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, po
...
+				return false, types.BlobInfo{}, "", errors.Wrapf(err, "Error trying to reuse blob %s at destination", srcInfo.Digest)
+			}
+			if reused {
+				logrus.Debugf("Skipping blob %s (already present):", srcInfo.Digest)
+				bar.SetTotal(0, true)
+				ic.c.createProgressBar(pool, srcInfo, "blob", "skipped: already exists", mpb.BarReplaceOnComplete(bar))
+				return true, blobInfo, cachedDiffID, nil
+			}
+
+			if !ic.c.dest.SupportsBlobLocks() {
```
mtrmac: This is too big for a single (nested!) function.

Maybe extract the code, starting roughly from this line, until the `!copiedByAnotherProcess` into a separate function? The logic to block and wait seems fairly self-contained.

---

Or, well, I’d still rather prefer for all of this to go into `storageImageDestination.TryReusingBlob`, so that the locking/unlocking logic is localized to that transport.

copy/copy.go
```diff
@@ -597,15 +647,35 @@ func (c *copier) createProgressBar(pool *mpb.Progress, info types.BlobInfo, kind
...
+		options = append(options,
+			mpb.AppendDecorators(
+				decor.OnComplete(decor.CountersKibiByte("%.1f / %.1f"), " done"),
+			))
+		options = append(options, mpb.BarClearOnComplete())
+	}
+
+	var bar *mpb.Bar
+	if filler != "" {
+		barFiller := mpb.FillerFunc(
```
mtrmac: (BTW, if I understand the idea correctly, we don’t need to replace the bars only to change the message; the `FillerFunc` could just `fmt.Fprint(*someVariable)` and we could update the variable (+ locking in both the reader and writer).

That does not look like an improvement right now, but if the logic moved into `storageImageDestination.TryReuseBlob`, it could make the API to update the progress bar with status messages notably easier.)

types/types.go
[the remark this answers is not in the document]
mtrmac: (To say who locks it, who unlocks it, maybe when, what is expected after unlocking.  … that feels like a lot.

Or at least what is the purpose, “to synchronize” is not very helpful.

Hiding all of this in the storage transport looks more and more attractive.)

storage/storage_image.go
mtrmac: (no further updates to the FIXME.)

mtrmac:
> Let me know when you're done. I will repush after that.

Done now.

storage/storage_image.go
```diff
@@ -360,6 +367,57 @@ func (s *storageImageDestination) HasThreadSafePutBlob() bool {
...
+	return nil
+}
+
+func (s *storageImageDestination) unlockDigest(d digest.Digest) error {
+	s.digestLockMutex.Lock()
+	defer s.digestLockMutex.Unlock()
+
+	locker, ok := s.lockedDigests[d]
+	if !ok {
+		return errors.Errorf("trying to unlock non existent lock for digest %q", d)
```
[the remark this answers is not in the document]
vrothberg: The documentation is incorrect. I missed updating that. Sorry for that!

storage/storage_image.go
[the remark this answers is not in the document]
vrothberg: Done.

copy/copy.go
```diff
@@ -485,20 +497,58 @@ func (ic *imageCopier) copyLayers(ctx context.Context) error {
...
 	func() { // A scope for defer
 		progressPool, progressCleanup := ic.c.newProgressPool(ctx)
 		defer progressCleanup()
 
+		// digestToIndex records which index in `data` corresponds to
+		// which digest.  The intention behind is to avoid that we are
+		// copying the same blob multiple times which would yield a
+		// deadlock since `copyLayer()` is locking blobs which are
+		// unlocked at a very different part of the stack, after
+		// `Image()`.  Ideally `copyLayer()` would lock and unlock a
```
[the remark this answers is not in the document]
vrothberg: Image() has a deferred dest.Close() but I re-reading it, I understand that it must be reworded. Thanks!

copy/copy.go
vrothberg: Wonderful idea! I will follow your advise and refine that.

copy/copy.go
vrothberg: It will complete. `createProgressBar()` will set the size to 0 if a filler is specified.

copy/copy.go
vrothberg: It will complete. createProgressBar() will set the size to 0 if a filler is specified.

copy/copy.go
vrothberg: It will complete. createProgressBar() will set the size to 0 if a filler is specified.

copy/copy.go
vrothberg: Done.

copy/copy.go
vrothberg: I saw a similar behavior yesterday in the GitHub UI. I was under the impression that I couldn't comment on threads where I replied with "Done.".

> I’d still prefer a select instead of having to wait here.

I will look into it.

vrothberg: Done. Thanks for pushing on this, it's much cleaner now.

copy/copy.go
vrothberg: Will do :+1:

vrothberg: Done.

copy/copy.go
vrothberg: Done.

copy/copy.go
vrothberg: I think it's possible to do that but this would not work when replacing a "placeholder" with the real progress bar, which does not specify a filler. I prefer to keep it as is (now with a dedicated method to create a replacement bar).

copy/copy.go
vrothberg: Done.

copy/copy.go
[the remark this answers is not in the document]
vrothberg: Done (changed the structs altogether as suggested).

storage/storage_image.go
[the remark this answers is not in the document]
vrothberg: Done.

storage/storage_image.go
vrothberg: Update: we leave deletion up to the garbage collector and use a 2nd map `unlockedDigest` that keeps track of the state.

vrothberg:
@mtrmac I updated the PR according to your feedback. I vendored the latest state into the CI PRs:
- https://github.com/containers/buildah/pull/1445
- https://github.com/containers/skopeo/pull/619
- https://github.com/containers/libpod/pull/2763

At the moment, I cannot run it through CRI-O as a) the K8s e2e tests regress and b) https://github.com/kubernetes-sigs/cri-o/pull/2195 must be merged before.

vrothberg:
CRI-O is now unblocked, the CI PR can be found here: https://github.com/containers/libpod/pull/2763

vrothberg:
The PR for c/storage has been merged and I just updated the `vendor.conf` here to depend on c/storage (not on my fork anymore).

vrothberg:
There is one scenario I did not see until now, which is a big blocker to merge the code.

Assume we have an image with two layers (L1, L2) and two processes (P1, P2) pulling it. We can run into a situation where P1 is locking L1 and P2 is locking L2, both waiting for the other process to release the locks: deadlock. I do not know how I could not see this issue before...

wking:
Re: deadlocks, I think you need a timeout on waiting for locks, after which you pull in parallel anyway.  See the "denial of service" paragraph [here][1].

[1]: https://github.com/containers/libpod/issues/1911#issue-386427543

vrothberg:
> Re: deadlocks, I think you need a timeout on waiting for locks, after which you pull in parallel anyway. See the "denial of service" paragraph here.

Thanks for the feedback! I think these are two orthogonal issues. If a user goes mad and (for some reason) can access our storage, we are doomed in any case. There are various other locks which could lead to a DOS. Hence, this scenario is not (yet) scope of this PR.

The problem we have in the PR now is that a DOS can currently currently happen by design without any attack, which is more than unfortunate. I have some ideas that I want to evaluate with @mtrmac who has a much better judgment on the code than I have.

However, I think that a time out would make sense for the blob locking to be on the safe side. But I really prefer to first fix the underlying problem of the potential for deadlocks.

wking:
> We can run into a situation where P1 is locking L1 and P2 is locking L2, both waiting for the other process to release the locks: deadlock.

If you assume well-behaved actors, won't P1 soon finish pulling L1 and then release the lock, after which P2 can use the layer?  That's why we have layer-level locking and not image-level locking.  I'm not seeing a case where this deadlocks with no work being done unless a bad actor acquires a layer lock and then sits on its hands instead of working to fetch the layer (which is where lock-waiting timeouts would come in).

vrothberg:
> If you assume well-behaved actors, won't P1 soon finish pulling L1 and then release the lock, after which P2 can use the layer? That's why we have layer-level locking and not image-level locking. I'm not seeing a case where this deadlocks with no work being done unless a bad actor acquires a layer lock and then sits on its hands instead of working to fetch the layer (which is where lock-waiting timeouts would come in).

The devil lies in the detail. Currently, the downloading of a layer is decoupled from committing the layer to the storage, and we commit the layers *after* all layers downloaded. Only when a layer is comitted, the associated lock will be released. Hence, in the upper example. P1 and P2 will be deadlocked during downloading as both are waiting for the other process to release the lock.

Certainly, the deadlock can avoided by combining the download and commit of a given layer, which is something I am doing at the moment. This will also speed up the entire process as layers will be committed as soon as they are available.

vrothberg:
> Certainly, the deadlock can be avoided by combining the download and commit of a given layer, which is something I am doing at the moment. This will also speed up the entire process as layers will be committed as soon as they are available.

I opened https://github.com/containers/image/pull/611 for that. It's not yet in a state for merging but it passes local smoke tests.

storage/storage_image.go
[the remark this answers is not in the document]
mtrmac: (All the other copies of the documentation need updating as well. — but let’s first see if we need to make this public at all.)

copy/copy.go
mtrmac: If I am reading the code correctly, a zero-sized bar does not automatically complete; `toComplete` is set only by `Bar.SetTotal` or `Bar.IncrBy`.

mtrmac: Anyway, this was fixed in the meantime.

copy/copy.go
mtrmac: If I am reading the code correctly, a zero-sized bar does not automatically complete; `toComplete` is set only by `Bar.SetTotal` or `Bar.IncrBy`.

Anyway, in the `!reused` case below, this progress bar should somehow be returned to the caller, so that it can be replaced before `copyLayerBlobFromStream`.

(In the `reused` case, this progress bar is replaced, which completes it.)

copy/copy.go
mtrmac: The
```go
// Note that we call SetTotal() after the new
// bar is created to avoid potential flickering
```
comment is hard-earned knowledge that seems worth preserving.

copy/copy.go
mtrmac: Sure, the real progress bar can be a different different animal.

The `fmt.Fprintf(*someVariable)` approach seems attractive in that it would give `GetBlob` (for status about mirrors) and `TryReusingBlob` (for all of this code, if we manage to move it into the transport) a pretty nice way to report progress: we could give them a
```go
type UserProgressNotifier interface {
   Update(string message)
}
```

storage/storage_image.go
mtrmac: #611 and the later deadlock conversation will take care of this.

storage/storage_image.go
```diff
@@ -64,6 +64,11 @@ type storageImageDestination struct {
 	fileSizes      map[digest.Digest]int64         // Mapping from layer blobsums to their sizes
 	filenames      map[digest.Digest]string        // Mapping from layer blobsums to names of files we used to hold them
 	SignatureSizes []int                           `json:"signature-sizes,omitempty"` // List of sizes of each signature slice
+
+	// data for {Lock,Unlock}Blob
+	lockedDigests   map[digest.Digest]storage.Locker // Bookkeeping of digests that have been locked
+	unlockedDigests map[digest.Digest]bool           // Bookkeeping of digests that have been unlocked. Required to unlock yet unlocked digests in Close()
```
mtrmac: (Absolutely non-blocking: The symmetry between `lockedDigests` and `unlockedDigests`, when are so different, makes me nervous. Maybe name the first `digestLockers`?

Also, while the `unlockedDigests` idea does work correctly AFAICT, I can’t quite formulate a _simple_ explanation why. If you have one, writing it down (here or somewhere around `LockBlob`) would be nice.)

vrothberg:
Closing in favour of containers/image#611.

PR #609, opened 2019-03-21
Create new version with semver

rhatdan:


rhatdan:
@vrothberg @mtrmac PTAL
We need to get this merged so that we can fix skopeo and then vendor into buildah/podman and cri-o.

mtrmac:
- I expect this will fail without https://github.com/containers/skopeo/pull/620 updated and merged
- Is this transitioning to Go modules, and effectively deciding that _right now_ we are freezing the API and making it all public and long-term compatible?

vrothberg:
> Is this transitioning to Go modules, and effectively deciding that right now we are freezing the API and making it all public and long-term compatible?

I don't think that's the plan. We can continue rolling fast but need to reflect a bit more about the nature of the merged changes before cutting a release to adhere to the rules of semver regarding what the major.minor.patch parts mean.

rhatdan:
@mtrmac Yes I don't think we are committing to anything, just that if we break the API we move the major version up one.

rhatdan:
As soon as I get this merged, I can update skopeo with the updated version. Can we get this merged?

vrothberg:
LGTM

mtrmac:
> As soon as I get this merged, I can update skopeo with the updated version.

I’d prefer to get https://github.com/containers/skopeo/pull/620 updated _first_, so that tests in here pass the ordinary way; yes, this is likely to be low-risk, but I’d rather not get into the habit of making unnecessary exceptions. (Or, well, anyone who disagrees can just merge this.)

rhatdan:
Skopeo merge has happened.

vrothberg:
Thanks, @rhatdan! Tests are green again -> merging.

PR #612, opened 2019-04-05
blobinfocache: correctly detect the UID in a namespace

giuseppe:
for tools in containers/ we are using the env variable
_CONTAINERS_ROOTLESS_UID when running in a user namespace to refer to
the rootless UID that created the user namespace.

Closes: https://github.com/containers/libpod/issues/2510#issuecomment-480066763


rhatdan:
LGTM
@vrothberg @mtrmac PTAL

vrothberg:
LGTM

PR #613, opened 2019-04-06
storage: fix UID detection in a namespace

giuseppe:
correctly detect the UID when running in rootless mode.

### DIFF
--- a/vendor.conf
+++ b/vendor.conf
@@ -1,7 +1,7 @@
 github.com/containers/image
 
 github.com/sirupsen/logrus v1.0.0
-github.com/containers/storage v1.12.1
+github.com/containers/storage v1.12.2
 github.com/davecgh/go-spew 346938d642f2ec3594ed81d874461961cd0faa76
 github.com/docker/docker-credential-helpers d68f9aeca33f5fd3f08eeae5e9d175edf4e731d1
 github.com/docker/distribution 5f6282db7d65e6d72ad7c2cc66310724a57be716
--- a/storage/storage_transport.go
+++ b/storage/storage_transport.go
@@ -4,7 +4,6 @@ package storage
 
 import (
 	"fmt"
-	"os"
 	"path/filepath"
 	"strings"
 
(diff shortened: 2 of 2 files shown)


vendor.conf
```diff
@@ -1,7 +1,7 @@
 github.com/containers/image
 
 github.com/sirupsen/logrus v1.0.0
-github.com/containers/storage v1.12.1
+github.com/containers/storage 1.12.2
```
TomSweeneyRedHat: Think you need a v in front of this.
```
FATA[0009] `git checkout -f --detach 1.12.2` failed:
fatal: git checkout: --detach does not take a path argument '1.12.2'
```

rhatdan:
You forgot to do a `make vendor`, or to include the vendor code in this PR,

giuseppe:
> You forgot to do a `make vendor`, or to include the vendor code in this PR,

I got confused by that when I've tried it.  It looks like there is no `vendor` anymore in the repo and we use only `vendor.conf`, in facts vendor is added to .gitignore.

giuseppe:
do we need to update Skopeo vendor first or is it fine that the test is failing?

giuseppe:
Skopeo WIP branch works fine: https://github.com/containers/skopeo/pull/624

mtrmac:
> do we need to update Skopeo vendor first or is it fine that the test is failing?

We typically want a ready-to-merge Skopeo PR with tests passing in there before merging a c/image PR, so that the c/image CI is not broken long. ( https://github.com/containers/skopeo/blob/master/CONTRIBUTING.md#dependencies-management )

https://github.com/containers/skopeo/pull/624 can work fine for this purpose, or if you want to, create a separate skopeo PR that only adds the vendored packages and can be merged immediately, in order not to block this one on the other work in https://github.com/containers/skopeo/pull/624 .

mtrmac:
LGTM pending tests (passing in skopeo, as long as they fail in here only in `test-skopeo`).

Issue #540, opened 2018-12-06
create manpage for transports


vrothberg:
As mentioned [here](https://github.com/containers/buildah/pull/1221#issuecomment-444566913), containers/image should have a manpage describing all supported transports. This manpage can then be referenced by other's (e.g., buildah/podman push/pull etc. and skopeo) and, if needed, mention which transports are not supported by a given command.

PR #615, opened 2019-04-12
docs: containers-transports(5) manpage

vrothberg:
Fixes: https://github.com/containers/image/issues/540
Fixes: github.com/containers/skopeo/issues/590

### DIFF
--- a/docs/containers-transports.5.md
+++ b/docs/containers-transports.5.md
@@ -1,0 +1,109 @@
+% CONTAINERS-TRANSPORTS(5) Containers Transports Man Page
+% Valentin Rothberg
+% April 2019
+
+## NAME
+
+containers-transports - description of supported transports for copying and storing container images
+
+## DESCRIPTION
+
+Tools which use the containers/image library, including skopeo(1), buildah(1), podman(1), all share a common syntax for referring to container images in various locations.
+The general form of the syntax is _transport:details_, where details are dependent on the specified transport, which are documented below.
+
+### **containers-storage:** [storage-specifier]{image-id|docker-reference[@image-id]}
+
+An image located in a local containers storage.
+The format of _docker-reference_ is described in detail in the **docker** transport.
+
+The _storage-specifier_ allows for referencing storage locations on the file system and has the format `[[driver@]root[+run-root][:options]]` where the optional `driver` refers to the storage driver (e.g., overlay or btrfs) and where `root` is an absolute path to the storage's root directory.
+The optional `run-root` can be used to specify the run directory of the storage where all temporary writable content is stored.
+The optional `options` are a comma-separated list of driver-specific options.
+Please refer to containers-storage.conf(5) for further information on the drivers and supported options.
+
+### **dir:**_path_
+
+An existing local directory _path_ storing the manifest, layer tarballs and signatures as individual files.
+This is a non-standardized format, primarily useful for debugging or noninvasive container inspection.
+
+### **docker://**_docker-reference_
+
+An image in a registry implementing the "Docker Registry HTTP API V2".
+By default, uses the authorization state in `$XDG_RUNTIME_DIR/containers/auth.json`, which is set using podman-login(1).
+If the authorization state is not found there, `$HOME/.docker/config.json` is checked, which is set using docker-login(1).
+The containers-registries.conf(5) further allows for configuring various settings of a registry.
+
+Note that a _docker-reference_ has the following format: `name[:tag|@digest]`.
+While the docker transport does not support both a tag and a digest at the same time some formats like containers-storage do.
+Digests can also be used in an image destination as long as the manifest matches the provided digest.
+The digest of images can be explored with skopeo-inspect(1).
+If `name` does not contain a slash, it is treated as `docker.io/library/name`.
+Otherwise, the component before the first slash is checked if it is recognized as a `hostname[:port]` (i.e., it contains either a . or a :, or the component is exactly localhost).
+If the first component of name is not recognized as a `hostname[:port]`, `name` is treated as `docker.io/name`.
+
+### **docker-archive:**_path[:docker-reference]_
+
+An image is stored in the docker-save(1) formatted file.
+_docker-reference_ is only used when creating such a file, and it must not contain a digest.
+It is further possible to copy data to stdin by specifying `docker-archive:/dev/stdin` but note that the used file must be seekable.
+
+### **docker-daemon:**_docker-reference|algo:digest_
+
+An image stored in the docker daemon's internal storage.
+The image must be specified as a _docker-reference_ or in an alternative _algo:digest_ format when being used as an image source.
+The _algo:digest_ refers to the image ID reported by docker-inspect(1).
+
+### **oci:**_path[:tag]_
+
+An image compliant with the "Open Container Image Layout Specification" at _path_.
+Using a _tag_ is optional and allows for storing multiple images at the same _path_.
+
+### **oci-archive:**_path[:tag]_
+
+An image compliant with the "Open Container Image Layout Specification" stored as a tar(1) archive at _path_.
+
+### **ostree:**_docker-reference[@/absolute/repo/path]_
+
+An image in the local ostree(1) repository.
+_/absolute/repo/path_ defaults to _/ostree/repo_.
+
+## Examples
+
+The following examples demonstrate how some of the containers transports can be used.
+The examples use skopeo-copy(1) for copying container images.
+
+**Copying an image from one registry to another**:
+```
+$ skopeo copy docker://docker.io/library/alpine:latest docker://localhost:5000/alpine:latest
+```
+
+**Copying an image from a running Docker daemon to a directory in the OCI layout**:
+```
+$ mkdir alpine-oci
+$ skopeo copy docker-daemon:alpine:latest oci:alpine-oci
+$ tree alpine-oci
+test-oci/
+├── blobs
+│   └── sha256
+│       ├── 83ef92b73cf4595aa7fe214ec6747228283d585f373d8f6bc08d66bebab531b7
+│       ├── 9a6259e911dcd0a53535a25a9760ad8f2eded3528e0ad5604c4488624795cecc
+│       └── ff8df268d29ccbe81cdf0a173076dcfbbea4bb2b6df1dd26766a73cb7b4ae6f7
+├── index.json
+└── oci-layout
+
+2 directories, 5 files
+```
+
+**Copying an image from a registry to the local storage**:
+```
+$ skopeo copy docker://docker.io/library/alpine:latest containers-storage:alpine:latest
+```
+
+## SEE ALSO
+
+docker-login(1), docker-save(1), ostree(1), podman-login(1), skopeo-copy(1), skopeo-inspect(1), tar(1), container-registries.conf(5), containers-storage.conf(5)
+
+## AUTHORS
+
+Miloslav Trmač <mitr@redhat.com>
+Valentin Rothberg <rothberg@redhat.com>


vrothberg:
@mtrmac @rhatdan @giuseppe @TomSweeneyRedHat PTAL

(shamelessly copied from skopeo(1))

docs/containers-transports.5.md
```diff
@@ -0,0 +1,53 @@
...
+**oci:**_path_**:**_tag_
+
+An image _tag_ in a directory compliant with "Open Container Image Layout Specification" at _path_.
+
+**ostree:**_image_[**@**_/absolute/repo/path_]
+
+An image in local OSTree repository.
+_/absolute/repo/path_ defaults to _/ostree/repo_.
+
+## SEE ALSO
```
rhatdan: Should we have some examples using skopeo and podman?

This man page is going to be rather confusing without some examples/use cases.

rhatdan:
@TomSweeneyRedHat PTAL

docs/containers-transports.5.md
vrothberg: Sure, we could add an #EXAMPLES section showing some examples with podman, buildah and skopeo.

docs/containers-transports.5.md
```diff
@@ -0,0 +1,53 @@
...
+
+An image in a registry implementing the "Docker Registry HTTP API V2".
+By default, uses the authorization state in either `$XDG_RUNTIME_DIR/containers/auth.json`, which is set using `(podman login)`.
+If the authorization state is not found there, `$HOME/.docker/config.json` is checked, which is set using `(docker login)`.
+
+**docker-archive:**_path_[**:**_docker-reference_]
+
+An image is stored in the `docker save` formatted file.
+_docker-reference_ is only used when creating such a file, and it must not contain a digest.
+Note that it is possible to copy data to stdin by specifying `docker-archive:/dev/stdin`.
```
giuseppe: my mistake here, as it doesn't really work with not seekable files.  Could you just add a note saying that a seekable file is needed?

docs/containers-transports.5.md
```diff
@@ -0,0 +1,53 @@
...
+
+**dir:**_path_
+
+An existing local directory _path_ storing the manifest, layer tarballs and signatures as individual files.
+This is a non-standardized format, primarily useful for debugging or noninvasive container inspection.
+
+**docker://**_docker-reference_
+
+An image in a registry implementing the "Docker Registry HTTP API V2".
+By default, uses the authorization state in either `$XDG_RUNTIME_DIR/containers/auth.json`, which is set using `(podman login)`.
```
TomSweeneyRedHat: remove "either' as you don't have 'or' later in the sentence.  Your next sentence gets the point across just fine.

docs/containers-transports.5.md
```diff
@@ -0,0 +1,53 @@
...
+
+**docker-archive:**_path_[**:**_docker-reference_]
+
+An image is stored in the `docker save` formatted file.
+_docker-reference_ is only used when creating such a file, and it must not contain a digest.
+Note that it is possible to copy data to stdin by specifying `docker-archive:/dev/stdin`.
+
+**docker-daemon:**_docker-reference_
+
+An image _docker-reference_ stored in the docker daemon internal storage.
```
TomSweeneyRedHat: "A _docker-reference_ formatted image stored" perhaps?

docs/containers-transports.5.md
```diff
@@ -0,0 +1,53 @@
...
+
+**ostree:**_image_[**@**_/absolute/repo/path_]
+
+An image in local OSTree repository.
+_/absolute/repo/path_ defaults to _/ostree/repo_.
+
+## SEE ALSO
+podman-login(1), containers-storage.conf(5)
+
+# AUTHORS
```
TomSweeneyRedHat: I think this should have a double hash `##`

docs/containers-transports.5.md
```diff
@@ -0,0 +1,53 @@
+% CONTAINERS-TRANSPORTS(5) Containers Transports Man Page
+% Valentin Rothberg
+% April 2019
+
+# NAME
+containers-transports - description of supported transports for copying container images
+
+# DESCRIPTION
```
TomSweeneyRedHat: Double hash this please `##`

docs/containers-transports.5.md
mtrmac: This is a reference of the details; let’s not turn this into a copy/pull/push tutorial. Sure, let’s add links to `skopeo-copy(1)` and `podman-pull(1)` here; but examples, if any, should primarily show the references, not the tools.

docs/containers-transports.5.md
```diff
@@ -0,0 +1,53 @@
...
+
+**containers-storage:**_docker-reference_
```
mtrmac: If only…

[`[`_complex options_`]`] (_docker reference with a tag or/**and** a digest_ | _hex digest prefix_) [`@`_hex digest prefix_]

where “prefix” includes the case of a full digest value.

docs/containers-transports.5.md
```diff
@@ -0,0 +1,53 @@
...
+
+An image located in a local containers storage image store.
+Please refer to containers-storage.conf(5) for further information.
+
+**dir:**_path_
+
+An existing local directory _path_ storing the manifest, layer tarballs and signatures as individual files.
+This is a non-standardized format, primarily useful for debugging or noninvasive container inspection.
+
+**docker://**_docker-reference_
```
mtrmac: _docker-reference_ may not include both a tag and a digest.

(Separately, _somewhere_ this should include an explanation of the reference format, incl. the `docker.io[/library]` canonicalization and the hostname vs. namespace in docker.io heuristic.)

docs/containers-transports.5.md
mtrmac: The image is not “reference-formatted”

docs/containers-transports.5.md
rhatdan: Right I would just like to see examples, probably in skopeo,

Copying from one registry to another,
Copying from a Docker-daemon to a oci image directory
Copying from a registry into container storage.

I don't care what tool you do it with.  To me examples speak much better then words.

docs/containers-transports.5.md
```diff
@@ -0,0 +1,53 @@
...
+
+An image is stored in the `docker save` formatted file.
+_docker-reference_ is only used when creating such a file, and it must not contain a digest.
+Note that it is possible to copy data to stdin by specifying `docker-archive:/dev/stdin`.
+
+**docker-daemon:**_docker-reference_
+
+An image _docker-reference_ stored in the docker daemon internal storage.
+_docker-reference_ must contain either a tag or a digest.
+Alternatively, when reading images, the format can be docker-daemon:algo:digest (an image ID).
```
mtrmac: This should be visible in the synopsis above.

docs/containers-transports.5.md
```diff
@@ -0,0 +1,53 @@
...
+**docker-archive:**_path_[**:**_docker-reference_]
+
+An image is stored in the `docker save` formatted file.
+_docker-reference_ is only used when creating such a file, and it must not contain a digest.
+Note that it is possible to copy data to stdin by specifying `docker-archive:/dev/stdin`.
+
+**docker-daemon:**_docker-reference_
+
+An image _docker-reference_ stored in the docker daemon internal storage.
+_docker-reference_ must contain either a tag or a digest.
```
mtrmac: digests are only acceptable when reading images, not when used as a destination.

docs/containers-transports.5.md
```diff
@@ -0,0 +1,53 @@
...
+Alternatively, when reading images, the format can be docker-daemon:algo:digest (an image ID).
+
+**oci:**_path_**:**_tag_
```
mtrmac: `:`_tag_ is optional (if it is missing, a source can only contain one image).

mtrmac: `oci-archive:` is missing.

docs/containers-transports.5.md
```diff
@@ -0,0 +1,53 @@
...
+
+An image _tag_ in a directory compliant with "Open Container Image Layout Specification" at _path_.
+
+**ostree:**_image_[**@**_/absolute/repo/path_]
+
+An image in local OSTree repository.
+_/absolute/repo/path_ defaults to _/ostree/repo_.
+
```
mtrmac: (`atomic:` is missing (but deprecated and probably broken against new OpenShift versions, so OK to leave out).)

docs/containers-transports.5.md
```diff
@@ -0,0 +1,53 @@
...
+An image _docker-reference_ stored in the docker daemon internal storage.
+_docker-reference_ must contain either a tag or a digest.
+Alternatively, when reading images, the format can be docker-daemon:algo:digest (an image ID).
+
+**oci:**_path_**:**_tag_
+
+An image _tag_ in a directory compliant with "Open Container Image Layout Specification" at _path_.
+
+**ostree:**_image_[**@**_/absolute/repo/path_]
```
mtrmac: _image_ uses the docker-reference format.

docs/containers-transports.5.md
mtrmac: Not a “digest prefix”, but an “image ID (prefix)”. And the above seems to allow _image ID prefix_`@`_image ID prefix`, which is forbidden.

(In hindsight, we should have insisted on a parser generator before starting this textual syntax business…)

mtrmac: So,

[`[`_complex options_`]`] ( _image ID prefix_  | _docker reference with a tag or/**and** a digest_ [`@`_image ID prefix_] )

I think.

docs/containers-transports.5.md
vrothberg: Done.

docs/containers-transports.5.md
vrothberg: Done.

docs/containers-transports.5.md
vrothberg: Done.

docs/containers-transports.5.md
vrothberg: Done.

docs/containers-transports.5.md
vrothberg: Done.

docs/containers-transports.5.md
vrothberg:
```
FATA[0000] Error determining manifest MIME type for docker-daemon:alpine@sha256:28ef97b8686a0b5399129e9b763d5b7e5ff03576aa5580d6f4182a49c5fe1913: Manifest does not match provided manifest digest sha256:28ef97b8686a0b5399129e9b763d5b7e5ff03576aa5580d6f4182a49c5fe1913
```

Using the digest as source does not work atm (Cc @nalind).

docs/containers-transports.5.md
vrothberg: Rewritten in hope to make things clearer.

docs/containers-transports.5.md
vrothberg: Lucky me :)

docs/containers-transports.5.md
vrothberg: Done.

docs/containers-transports.5.md
vrothberg: done

docs/containers-transports.5.md
vrothberg: > [[complex options]] ( image ID prefix | docker reference with a tag or/and a digest [@image ID prefix] )
> I think.

This looks good :+1:

vrothberg:
Updated / refined / rewritten :-)

@rhatdan @mtrmac @TomSweeneyRedHat @giuseppe PTAL

It would also be nice to have a manpage for containers-storage-drivers(5) :^)

rhatdan:
LGTM, but @TomSweeneyRedHat and @mtrmac Are better at this wordsmithing them me.

docs/containers-transports.5.md
```diff
@@ -0,0 +1,102 @@
...
+Note that a _docker-reference_ has the following format: `"name[:tag|@digest]"`.
+The name can either be a `domain:port` (e.g., quay.io) or be a non-domain string without "." or ":" (e.g., alpine). Non-domain names will be expanded to use the docker.io registry while those without a "/" will be prefixed with "docker.io/library" (e.g., "alpine" to "docker.io/library/alpine"). Non-domain containing a "/" will be prefixed with "docker.io" assuming that the name already refers to a repository (e.g., "circleci/ruby" to "docker.io/circleci/ruby").
+
+A _docker-reference_ may also include a digest. The digest of remote images can be explored with skopeo-inspect(1). Digests are only allowed when being used as an image source cannot be combined with a tag (e.g., tag@digest)
+
+### **docker-archive:**_path[:docker-reference]_
+
+An image is stored in the docker-save(1) formatted file.
+_docker-reference_ is only used when creating such a file, and it must not contain a digest.
+It is further possible to copy data to stdin by specifying `docker-archive:/dev/stdin` but note that the used file must be seekable.
```
mtrmac: (Absolutely non-blocking: I’m not sure this sentence is really adding much — to be flippant, this is pretty close to “you can use any file by specifying its name”. Maybe the point about the path having to refer to a seekable file should be retained, but the reference /dev/stdin feels a bit pointless to me.)

docs/containers-transports.5.md
```diff
@@ -0,0 +1,102 @@
+% CONTAINERS-TRANSPORTS(5) Containers Transports Man Page
+% Valentin Rothberg
+% April 2019
+
+
+# NAME
+containers-transports - description of supported transports for copying and storing container images
+
+
+# DESCRIPTION
```
mtrmac: (Non-blocking: Having at least one sentence to say what this page is all about, and where this syntax can/can’t be used, might be useful.

Maybe
> Tools which use the containers/image library, including skopeo(1), buildah(1), podman(1), all share a common syntax for referring to container images in various locations. The general form of the syntax is _transport_`:`:_details_, where _details_ are dependent on the specified _transport_; the individual transports are documented below.

)

docs/containers-transports.5.md
```diff
@@ -0,0 +1,102 @@
...
+This is a non-standardized format, primarily useful for debugging or noninvasive container inspection.
+
+### **docker://**_docker-reference_
+
+An image in a registry implementing the "Docker Registry HTTP API V2".
+By default, uses the authorization state in `$XDG_RUNTIME_DIR/containers/auth.json`, which is set using podman-login(1).
+If the authorization state is not found there, `$HOME/.docker/config.json` is checked, which is set using docker-login(1).
+The containers-registries.conf(5) further allows for configuring various settings of a registry.
+
+Note that a _docker-reference_ has the following format: `"name[:tag|@digest]"`.
```
mtrmac: While `docker://` does not, some formats, like c/storage, do support both a tag and digest at the same time.

docs/containers-transports.5.md
```diff
@@ -0,0 +1,102 @@
...
+The name can either be a `domain:port` (e.g., quay.io) or be a non-domain string without "." or ":" (e.g., alpine). Non-domain names will be expanded to use the docker.io registry while those without a "/" will be prefixed with "docker.io/library" (e.g., "alpine" to "docker.io/library/alpine"). Non-domain containing a "/" will be prefixed with "docker.io" assuming that the name already refers to a repository (e.g., "circleci/ruby" to "docker.io/circleci/ruby").
```
mtrmac: The name never “is” a domain:port. Port is optional. `localhost`.

Maybe 
> If _name_ does not contain a slash, it is treated as `docker.io/library/`_name_.
> If _name_ does contain a slash, the component before the first slash is checked whether it is recognized as a _hostname_[`:`_port_] (it contains either a `.` or a `:`, or the component is exactly `localhost`). If the first component of _name_ is not recognized as a _hostname_[`:`_port_], _name_ is treated as `docker.io/`_name_.

I’m not too happy with the “its recognized as a _hostname_[`:`_port_]“ wording, though.)

docs/containers-transports.5.md
mtrmac: Oops, I never thought of that; and the daemon does not record the original manifest (and can’t provide the original compressed blobs), so this can’t work unless the transport could just opt out of digest verification (or lie and say that the reference did not contain a digest).

Anyway, the point was primarily that _destinations_ must not use digests.

docs/containers-transports.5.md
```diff
@@ -0,0 +1,102 @@
...
+
+
+# DESCRIPTION
+
+### **containers-storage:** [storage-reference]image-id|docker-reference  [@image-id]}
+
+An image located in a local containers storage.
+The format of _docker-reference_ is described in detail in the **docker** transport.
+
+The _storage-reference_ allows for referencing storage locations on the file system and has the format `"[driver@root]"` where driver refers to the storage driver (e.g., overlay or btrfs) and where root is an absolute or a relative path to the storage's root directory.
```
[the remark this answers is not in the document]
TomSweeneyRedHat: style, YMMV.  I think I'd just go with the backticks around [driver@root] and would drop the double quotes.

docs/containers-transports.5.md
TomSweeneyRedHat: ditto backtick and double quote thought here for name[:tag|@digest]

docs/containers-transports.5.md
```diff
@@ -0,0 +1,102 @@
...
+
+An image in a registry implementing the "Docker Registry HTTP API V2".
+By default, uses the authorization state in `$XDG_RUNTIME_DIR/containers/auth.json`, which is set using podman-login(1).
+If the authorization state is not found there, `$HOME/.docker/config.json` is checked, which is set using docker-login(1).
+The containers-registries.conf(5) further allows for configuring various settings of a registry.
+
+Note that a _docker-reference_ has the following format: `"name[:tag|@digest]"`.
+The name can either be a `domain:port` (e.g., quay.io) or be a non-domain string without "." or ":" (e.g., alpine). Non-domain names will be expanded to use the docker.io registry while those without a "/" will be prefixed with "docker.io/library" (e.g., "alpine" to "docker.io/library/alpine"). Non-domain containing a "/" will be prefixed with "docker.io" assuming that the name already refers to a repository (e.g., "circleci/ruby" to "docker.io/circleci/ruby").
+
+A _docker-reference_ may also include a digest. The digest of remote images can be explored with skopeo-inspect(1). Digests are only allowed when being used as an image source cannot be combined with a tag (e.g., tag@digest)
```
TomSweeneyRedHat: s/image source cannot/image source and cannot/

docs/containers-transports.5.md
```diff
@@ -0,0 +1,102 @@
...
+### **docker-archive:**_path[:docker-reference]_
+
+An image is stored in the docker-save(1) formatted file.
+_docker-reference_ is only used when creating such a file, and it must not contain a digest.
+It is further possible to copy data to stdin by specifying `docker-archive:/dev/stdin` but note that the used file must be seekable.
+
+### **docker-daemon:**_docker-reference|algo:digest_
+
+An image stored in the docker daemon's internal storage.
+The image must be specified as a _docker-reference_ or in an alternative _algo:digest_ format when being used an image source.
```
TomSweeneyRedHat: s/used an image/used as an image/

docs/containers-transports.5.md
```diff
@@ -0,0 +1,102 @@
...
+The _algo:digest_ refers to the image ID reported by docker-inspect(1).
+
+### **oci:**_path[:tag]_
+
+An image in a directory compliant with "Open Container Image Layout Specification" at _path_.
```
TomSweeneyRedHat: suggest:
An image compliant with "Open Container Image Layout Specification" in a directory at _path_.

docs/containers-transports.5.md
```diff
@@ -0,0 +1,102 @@
...
+Using a _tag_ is optional but allows for storing multiple images at the same _path_.
+
+### **oci-archive:**_path[:tag]_
+
+An image in a directory compliant with "Open Container Image Layout Specification" stored as a tar(1) archive at _path_.
```
TomSweeneyRedHat: if you use suggestion from 51, do similar here.

docs/containers-transports.5.md
vrothberg: There was a user request to document that this is possible, so I would like to keep the reference.

docs/containers-transports.5.md
vrothberg: Added, thanks!

docs/containers-transports.5.md
vrothberg: LGTM, thanks!

docs/containers-transports.5.md
```diff
@@ -0,0 +1,102 @@
...
+% April 2019
+
+
+# NAME
+containers-transports - description of supported transports for copying and storing container images
+
+
+# DESCRIPTION
+
+### **containers-storage:** [storage-reference]image-id|docker-reference  [@image-id]}
```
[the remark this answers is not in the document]
vrothberg: It's actually `[storage-reference]{image-id|docker-reference[@image-id]}`

The problem was that when using `(` instead of `{`, the man page adds a line break which entirely breaks readability.

docs/containers-transports.5.md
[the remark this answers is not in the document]
vrothberg: Done.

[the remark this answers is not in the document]
vrothberg: Corrected, thanks.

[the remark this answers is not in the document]
vrothberg: Changed to `[driver@]root[+run-root][:options]` with an explanations.

docs/containers-transports.5.md
vrothberg: Done, thanks!

docs/containers-transports.5.md
vrothberg: Nice catch, thanks!

docs/containers-transports.5.md
vrothberg: Done.

docs/containers-transports.5.md
vrothberg: I went with `An image compliant with THE "Open [...]`

docs/containers-transports.5.md
vrothberg: Done, thanks!

vrothberg:
Updated. @TomSweeneyRedHat @mtrmac, PTanotherL

docs/containers-transports.5.md
rhatdan: I don't see it?

docs/containers-transports.5.md
mtrmac: I think https://github.com/containers/skopeo/issues/590 was predicated on that actually working with pipes, but, sure, let’s keep it.

docs/containers-transports.5.md
```diff
@@ -0,0 +1,111 @@
...
+
+Note that a _docker-reference_ has the following format: `name[:tag|@digest]`.
+While the docker transport does not support both, a tag and a digest, at the same time, some formats, like containers-storage, do.
+If name does not contain a slash, it is treated as `docker.io/library/name`.
+Otherwise, the component before the first slash is checked whether it is recognized as a `hostname[:port]` (it contains either a . or a :, or the component is exactly localhost).
+If the first component of name is not recognized as a `hostname[:port]`, name is treated as `docker.io/name`.
+
+A _docker-reference_ may also include a digest.
+The digest of remote images can be explored with skopeo-inspect(1).
+Digests are only allowed when being used as an image source and cannot be combined with a tag (e.g., tag@digest)
```
mtrmac: - The text should be clearer about whether it talks about the generic reference format or about `docker://`
- The `:`_tag_`@`_digest_ combination permissibility has been mentioned above.
- Both `containers-storage:` and `docker://` actually support using a digest as a destination (as long as the manifest matches the digest)

docs/containers-transports.5.md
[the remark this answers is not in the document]
mtrmac: Ah, OK.  (The pedantic me would like consistent formatting differentiation between literal syntax elements like `:` and `@`, replaceable identifiers like _tag_, and syntax grouping characters like ( | ) .)

docs/containers-transports.5.md
mtrmac: Actually, shouldn’t the standard man-pages(7) (NAME, DESCRIPTION, SEE ALSO, AUTHORS) section headers all be at the same level?  https://github.com/cpuguy83/go-md2man/blob/master/go-md2man.1.md uses a single has sign.

rhatdan: Yes It should be
```
## NAME
and
## DESCRIPTION
```

docs/containers-transports.5.md
vrothberg: Right, those sentences are a bit redundant to the paragraph before, so I merged them.

vrothberg:
Repushed.

docs/containers-transports.5.md
```diff
@@ -0,0 +1,109 @@
...
+Tools which use the containers/image library, including skopeo(1), buildah(1), podman(1), all share a common syntax for referring to container images in various locations.
+The general form of the syntax is _transport:details_, where details are dependent on the specified transport, which are documented below.
+
+### **containers-storage:** [storage-specifier]{image-id|docker-reference[@image-id]}
+
+An image located in a local containers storage.
+The format of _docker-reference_ is described in detail in the **docker** transport.
+
+The _storage-specifier_ allows for referencing storage locations on the file system and has the format `[[driver@]root[+run-root][:options]]` where the optional `driver` refers to the storage driver (e.g., overlay or btrfs) and where `root` is an absolute path to the storage's root directory.
+`run-root` allows to further specify the run directory of the storage where all temporary writable content is stored.
```
TomSweeneyRedHat: Suggest: "The optional `run-root` can be used to specify the run directory of the storage where all temporary writable content is stored.

docs/containers-transports.5.md
```diff
@@ -0,0 +1,109 @@
...
+
+### **docker://**_docker-reference_
+
+An image in a registry implementing the "Docker Registry HTTP API V2".
+By default, uses the authorization state in `$XDG_RUNTIME_DIR/containers/auth.json`, which is set using podman-login(1).
+If the authorization state is not found there, `$HOME/.docker/config.json` is checked, which is set using docker-login(1).
+The containers-registries.conf(5) further allows for configuring various settings of a registry.
+
+Note that a _docker-reference_ has the following format: `name[:tag|@digest]`.
+While the docker transport does not support both, a tag and a digest, at the same time, some formats, like containers-storage, do.
```
TomSweeneyRedHat: Suggest dropping a number of commas here: "While the docker transport does not support both a tag and a digest at the same time, some formats like containers-storage do."

docs/containers-transports.5.md
```diff
@@ -0,0 +1,109 @@
...
+Digests can also be used in an image destination as long as the manifest matches the provided digest.
+The digest of images can be explored with skopeo-inspect(1).
+If name does not contain a slash, it is treated as `docker.io/library/name`.
```
TomSweeneyRedHat: suggest: "if `name`" (Backtick name)

docs/containers-transports.5.md
```diff
@@ -0,0 +1,109 @@
...
+Otherwise, the component before the first slash is checked whether it is recognized as a `hostname[:port]` (it contains either a . or a :, or the component is exactly localhost).
+If the first component of name is not recognized as a `hostname[:port]`, name is treated as `docker.io/name`.
```
TomSweeneyRedHat: suggest ", `name` is treated"

docs/containers-transports.5.md
```diff
@@ -0,0 +1,109 @@
...
+By default, uses the authorization state in `$XDG_RUNTIME_DIR/containers/auth.json`, which is set using podman-login(1).
+If the authorization state is not found there, `$HOME/.docker/config.json` is checked, which is set using docker-login(1).
+The containers-registries.conf(5) further allows for configuring various settings of a registry.
+
+Note that a _docker-reference_ has the following format: `name[:tag|@digest]`.
+While the docker transport does not support both, a tag and a digest, at the same time, some formats, like containers-storage, do.
+Digests can also be used in an image destination as long as the manifest matches the provided digest.
+The digest of images can be explored with skopeo-inspect(1).
+If name does not contain a slash, it is treated as `docker.io/library/name`.
+Otherwise, the component before the first slash is checked whether it is recognized as a `hostname[:port]` (it contains either a . or a :, or the component is exactly localhost).
```
TomSweeneyRedHat: When you have the word "whether" it's used to contrast two thoughts and the word "or" is expected later in the sentence and here the 'or' is inside the parens.  Suggest: "Otherwise, the component before the first slash is checked whether it is recognized as a `hostname[:port]` (it contains either a period (.) or a semi-colon (:)), or the component is exactly localhost.

docs/containers-transports.5.md
```diff
@@ -0,0 +1,109 @@
...
+### **docker-daemon:**_docker-reference|algo:digest_
+
+An image stored in the docker daemon's internal storage.
+The image must be specified as a _docker-reference_ or in an alternative _algo:digest_ format when being used as an image source.
+The _algo:digest_ refers to the image ID reported by docker-inspect(1).
+
+### **oci:**_path[:tag]_
+
+An image compliant with the "Open Container Image Layout Specification" at _path_.
+Using a _tag_ is optional but allows for storing multiple images at the same _path_.
```
TomSweeneyRedHat: suggest "but allows" to "and allows"

docs/containers-transports.5.md
```diff
@@ -0,0 +1,109 @@
...
+
+### **oci-archive:**_path[:tag]_
+
+An image compliant with the "Open Container Image Layout Specification" stored as a tar(1) archive at _path_.
+
+### **ostree:**_docker-reference[@/absolute/repo/path]_
+
+An image in the local ostree(1) repository.
+_/absolute/repo/path_ defaults to _/ostree/repo_.
```
TomSweeneyRedHat: suggest adding a comma after path_, "_/absolute/repo/path_, defaults to _/ostree/repo_."

vrothberg:
Thanks @TomSweeneyRedHat! Pushed another version. I think we're good to go now unless there is some technical issue.

TomSweeneyRedHat:
LGTM assuming happy tests.

docs/containers-transports.5.md
mtrmac: The condition being checked here is whether “_name_ is recognized…” (which includes `.`, `:`, and `localhost`), [or whether it is “not recognized” because it does not match that rule].

If I understand you correctly, moving the `localhost` case outside of the parenthesis would mean something else (treating `.`/`:` as one case, `localhost` as the other one, and being silent about other values).

Maybe keep the original text, and `s/whether/if`; I’m not at all sure.

docs/containers-transports.5.md
mtrmac: I read this as `/absolute/repo/path` being the subject of the sentence.

docs/containers-transports.5.md
mtrmac: (Most commas were dropped here but a different one was left.)

mtrmac:
Contents LGTM; I’ll leave English syntax and commas to others.

@vrothberg / @TomSweeneyRedHat feel free to merge if/whenever you are happy with this.

vrothberg:
Integrated @mtrmac's last comments. Thanks @mtrmac and @TomSweeneyRedHat for the great reviews. The result is a pretty good man page!

rhatdan:
Now we need to add this man page to containers-commont in the skopeo package.

PR #564, opened 2019-01-21
Add registry mirror support to docker client

saschagrunert:
This change relates to #559 and enables registry mirror support for the docker client.

### Major changes
* The first working mirror will be used for the pull
* The main registry will be the fallback if no mirror works
* The `Insecure` flags will be used from system context and now from the corresponding registry / mirror as fallback too
* Configuring a prefix will be considered (see test cases)

## Test cases:

The following test cases has been verified by manual testing via cri-o:

<details><summary><strong>It should succeed to pull an image from a working registry mirror</strong></summary>
<p>

*Given*
`/etc/containers/registry.conf`
```toml
[[registry]]
url = "gcr.local"
mirror = [
    { url = "wrong.local" },
    { url = "gcr.io" },
]
```
*When*
```fish
> sudo crictl pull gcr.local/google-containers/hyperkube:v1.13.2
```
*Then*
```fish
exit 0
```
</p></details>

<details><summary><strong>It should succeed to pull an image unqualified from a working registry mirror</strong></summary>
<p>

*Given*
`/etc/containers/registry.conf`
```toml
[[registry]]
url = "gcr.local"
unqualified-search = true
mirror = [
    { url = "wrong.local" },
    { url = "gcr.io" },
]
```
*When*
```fish
> sudo crictl pull google-containers/hyperkube:v1.13.2
```
*Then*
```fish
exit 0
```
</p></details>

<details><summary><strong>It should succeed to pull an image from a working registry mirror if unqualified search is enabled but not used</strong></summary>
<p>

*Given*
`/etc/containers/registry.conf`
```toml
[[registry]]
url = "gcr.local"
unqualified-search = true
mirror = [
    { url = "wrong.local" },
    { url = "gcr.io" },
]
```
*When*
```fish
> sudo crictl pull gcr.local/google-containers/hyperkube:v1.13.2
```
*Then*
```fish
exit 0
```
</p></details>

<details><summary><strong>It should succeed to pull an image from a working registry mirror with  prefix</strong></summary>
<p>

*Given*
`/etc/containers/registry.conf`
```toml
[[registry]]
url = "gcr.local"
prefix = "gcr.local/foo/bar"
mirror = [
    { url = "wrong.local" },
    { url = "gcr.io" },
]
```
*When*
```fish
> sudo crictl pull gcr.local/foo/bar/google-containers/hyperkube:v1.13.2
```
*Then*
```fish
exit 0
```
</p></details>

[...]

PR #617, opened 2019-04-24
Mirror documentation changes

mtrmac:
This follows up on #564 ; see the individual commit messages for details, I can drop any part you don’t like.

Notable parts:
- This drops most of the big comment before `newImageSource`. @vrothberg you asked for that documentation, or more documentation; is this OK with you?
- The documentation has been rewritten to hopefully make more sense to a first-time reader (notably it starts with the “prefix” value and how it is used to find the relevant entry), but I’m still not really happy with how it turned out. Maybe we should use a bulleted/definition list. Or maybe it’s because I’m not all that familiar with TOML and I struggle with the terminology. Either way, comments and suggestions (including “drop that, the original was better”) are very welcome.

Cc: @vrothberg @saschagrunert

### DIFF
--- a/docs/containers-registries.conf.5.md
+++ b/docs/containers-registries.conf.5.md
@@ -18,68 +18,127 @@ VERSION 2 is the latest format of the `registries.conf` and is currently in
 beta. This means in general VERSION 1 should be used in production environments
 for now.
 
-Every registry can have its own mirrors configured.  The mirrors will be tested
-in order for the availability of the remote manifest.  This happens currently
-only during an image pull.  If the manifest is not reachable due to connectivity
-issues or the unavailability of the remote manifest, then the next mirror will
-be tested instead.  If no mirror is configured or contains the manifest to be
-pulled, then the initially provided reference will be used as fallback.  It is
-possible to set the `insecure` option per mirror, too.
+### GLOBAL SETTINGS
 
-Furthermore it is possible to specify a `prefix` for a registry.  The `prefix`
-is used to find the relevant target registry from where the image has to be
-pulled.  During the test for the availability of the image, the prefixed
-location will be rewritten to the correct remote location.  This applies to
-mirrors as well as the fallback `location`.  If no prefix is specified, it
-defaults to the specified `location`.  For example, if
-`prefix = "example.com/foo"`, `location = "example.com"` and the image will be
-pulled from `example.com/foo/image`, then the resulting pull will be effectively
-point to `example.com/image`.
+`unqualified-search-registries`
+: An array of _host_[`:`_port_] registries to try when pulling an unqualified image, in order.
 
-By default container runtimes use TLS when retrieving images from a registry.
-If the registry is not setup with TLS, then the container runtime will fail to
-pull images from the registry. If you set `insecure = true` for a registry or a
-mirror you overwrite the `insecure` flag for that specific entry.  This means
-that the container runtime will attempt use unencrypted HTTP to pull the image.
-It also allows you to pull from a registry with self-signed certificates.
+### NAMESPACED `[[registry]]` SETTINGS
 
-If `blocked = true` then it is not allowed to pull images from that registry.
+The bulk of the configuration is represented as an array of `[[registry]]`
+TOML tables; the settings may therefore differ among different registries
+as well as among different namespaces/repositories within a registry.
 
-If `mirror-by-digest-only = true`, mirrors will only be used during pulling if
-the image is referred to by digest.  Referencing an image by digest enforces to
-always return the same image, independent if it is being pulled from the main
-registry or a mirror.  Pulling an image referenced by tag could yield different
-results, depending on which mirror or registry it is being pulled from.  Setting
-this attribute makes sure to avoid such a situation but requires the primary
-registry to always be available if references by tag are used.
+#### Choosing a `[[registry]]` TOML table
 
-In addition, at the top level, an `unqualified-search-registries` array can be
-specified; this is a list of host[:port] (not more general prefix!) entries to
-use when pulling an unqualified image, in order.
+Given an image name, a single `[[registry]]` TOML table is chosen based on its `prefix` field.
+
+`prefix`
+: A prefix of the user-specified image name, i.e. using one of the following formats:
+    - _host_[`:`_port_]
+    - _host_[`:`_port_]`/`_namespace_[`/`_namespace_…]
+    - _host_[`:`_port_]`/`_namespace_[`/`_namespace_…]`/`_repo_
+    - _host_[`:`_port_]`/`_namespace_[`/`_namespace_…]`/`_repo_(`:`_tag|`@`_digest_)
+
+    The user-specified image name must start with the specified `prefix` (and continue
+    with the appropriate separator) for a particular `[[registry]]` TOML table to be
+    considered; (only) the TOML table with the longest match is used.
+
+    As a special case, the `prefix` field can be missing; if so, it defaults to the value
+    of the `location` field (described below).
+
+#### Per-namespace settings
+
+`insecure`
+: `true` or `false`.
+    By default, container runtimes require TLS when retrieving images from a registry.
+    If `insecure` is set to `true`, unencrypted HTTP as well as TLS connections with untrusted
+    certificates are allowed.
+
+`blocked`
+: `true` or `false`.
+    If `true`, pulling images with matching names is forbidden.
+
+#### Remapping and mirroring registries
+
+The user-specified image reference is, primarily, a "logical" image name, always used for naming
+the image.  By default, the image reference also directly specifies the registry and repository
+to use, but the following options can be used to redirect the underlying accesses
+to different registry servers or locations (e.g. to support configurations with no access to the
+internet without having to change `Dockerfile`s, or to add redundancy).
+
+`location`
+: Accepts the same format as the `prefix` field, and specifies the physical location
+    of the `prefix`-rooted namespace.
+
+    By default, this equal to `prefix` (in which case `prefix` can be omitted and the
+    `[[registry]]` TOML table can only specify `location`).
+
+    Example: Given
+    ```
+    prefix = "example.com/foo"
+    location = "internal-registry-for-example.net/bar"
+    ```
+    requests for the image `example.com/foo/myimage:latest` will actually work with the
+    `internal-registry-for-example.net/bar/myimage:latest` image.
+
+`mirror`
+: An array of TOML tables specifiying (possibly-partial) mirrors for the
+    `prefix`-rooted namespace.
+
+    The mirrors are attempted in the specified order; the first one that can be
+    contacted and contains the image will be used (and if none of the mirrors contains the image,
+    the primary location specified by the `registry.location` field, or using the unmodified
+    user-specified reference, is tried last).
+
+    Each TOML table in the `mirror` array can contain the following fields, with the same semantics
+    as if specified in the `[[registry]]` TOML table directly:
+    - `location`
+    - `insecure`
+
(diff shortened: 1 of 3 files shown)


docs/containers-registries.conf.5.md
```diff
@@ -18,42 +18,49 @@ VERSION 2 is the latest format of the `registries.conf` and is currently in
...
-It also allows you to pull from a registry with self-signed certificates.
-
-If you set the `unqualified-search = true` for the registry, then it is possible
-to omit the registry hostname when pulling images.  This feature does not work
-together with a specified `prefix`.
+The configuration is a sequence of `[[registry]]` entries.  Given an input image
+name, a single entry is chosen; that entry must have a `prefix` field set to be
+a prefix of the input image name (taking into account namespace/repo/tag/digest
+separators), and (only) the entry with the longest matching `prefix` field
+is used.
```
vrothberg: I still find it hard to brain parse, as we immediately dive into some very specific detail. My suggestion would be to first explain the corner stones of such a table entry.

* Each entry allows us to configure specific attributes of a registry. What are these attributes?
* Location is mandatory.
* Then describe `insecure`.
* Then describe `blocked`.
* Then describe `unqualified-search`.
* Then explain `prefix` and how it can be used (and that it defaults to location).

Now we should have all pieces to explain a mirror (which also has the insecure field) and how they fit into the picture for pulling and what they mean when `unqualified-search` is set.

I also think it would be very beneficial to give examples *in line* for some of these attributes (and how they can be combined).

docs/containers-registries.conf.5.md
```diff
@@ -18,42 +18,49 @@ VERSION 2 is the latest format of the `registries.conf` and is currently in
...
+in the order they appear in the configuration file.
 
 If `blocked = true` then it is not allowed to pull images from that registry.
 
 ### EXAMPLE
 
 ```
 [[registry]]
-location = "example.com"
+location = "example.com/foo"
```
vrothberg: I prefer to give examples in the descriptions of the each registry attribute. This specific example will teach users how the syntax looks like but not yet the semantics.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -39,26 +39,18 @@ type Endpoint struct {
 // endpoints `location` from the `ref` and creates a new named reference from it.
 // The function errors if the newly created reference is not parsable.
 func (e *Endpoint) RewriteReference(ref reference.Named, prefix string) (reference.Named, error) {
-	if ref == nil {
```
saschagrunert: I'm personally not a fan of removing this check since it is a public API and crashing would be no option for me. But this is just my personal non-blocking opinion.

docs/containers-registries.conf.5.md
```diff
@@ -18,42 +18,49 @@ VERSION 2 is the latest format of the `registries.conf` and is currently in
...
-If you set the `unqualified-search = true` for the registry, then it is possible
-to omit the registry hostname when pulling images.  This feature does not work
-together with a specified `prefix`.
+The configuration is a sequence of `[[registry]]` entries.  Given an input image
+name, a single entry is chosen; that entry must have a `prefix` field set to be
+a prefix of the input image name (taking into account namespace/repo/tag/digest
+separators), and (only) the entry with the longest matching `prefix` field
+is used.
+
+As a special case, the `prefix` field can be missing; if so, it defaults to the value
```
saschagrunert: For me this should be the default case, because the feature is way more easy to understand when initially omitting the `prefix` semantics.

mtrmac:
Note to self: I’ll separate the code changes, which seem less controversial, into a separate PR to be merged sooner, and I’ll continue improving the text in this one.

mtrmac:
> Note to self: I’ll separate the code changes, which seem less controversial, into a separate PR to be merged sooner, and I’ll continue improving the text in this one.

Done: Code is now in #628; this merges in that code, and adds documentation on top; don’t merge this before #628 is merged and this is rebased on top of `master` again.

pkg/sysregistriesv2/system_registries_v2.go
mtrmac: Moved to https://github.com/containers/image/pull/628#discussion_r283525478 .

docs/containers-registries.conf.5.md
mtrmac: The `location` field documentation now has a clearly-identifiable example (using the real-world formatting instead of in-line text inside a paragraph), and the example is more practical (actually saying `internal-registry-for-example` instead of an unmotivated mapping from `example.com` to `example.net`).

docs/containers-registries.conf.5.md
mtrmac: I thought it makes more sense to focus on `prefix` rather than `location`, because _everyone_ editing the configuration needs to understand the semantics of selecting the entry, and that it is `prefix`-based, whereas actual remapping using a `prefix`/`location` pair is only useful for, presumably, much  rarer cases of sites disconnected from the internet.

So, we can have the discussion of `location` and mirrors in a separate section relevant only for people who care, and allow, notably, the `insecure` flag to usable without having to understand `location` at all.

docs/containers-registries.conf.5.md
mtrmac: > * Each entry allows us to configure specific attributes of a registry. What are these attributes?
…
> * Then describe `insecure`.
…
> I also think it would be very beneficial to give examples _in line_ for some of these attributes (and how they can be combined).

In total, that seems to suggest having everything documented 2-3 times, and 2 examples for every item (one in-line and one at the end).

This is primarily reference, authoritative, documentation; it should not repeat itself nor does it necessarily have to be useful for a complete newcomer (although it’s _good_ if it is readable for a newcomer, it’s not the _first_ priority; the _first_ priority is to be an authoritative reference).

---

Let me try coming at this from a completely different angle, see below.

mtrmac:
Updated, trying to make this readable using a completely different approach:
- Instead of a sequence of full-sentence paragraphs “If `option = true` then …”, use definition lists to make it much easier to skim the page and find the authoritative section for any option.
- Structure the v2 documentation more, using explicit headings:
  - GLOBAL SETTINGS
  - NAMESPACED [[registry]] SETTINGS
       - Choosing a [[registry]] entry
       - Per-namespace settings
       - Remapping and mirroring registries
- Use a more realistic example for the mirroring/remapping throughout, instead of meaningless `example.com`/`example.net`

Note: This is optimized for the output of `go-md2man` (and even then it is not quite great), notably the use of definition lists is not recognized by the GitHub Markdown formatter, which makes the rendered version rather ugly.

(Grumble… how did we end up with Markdown, with _both_ fewer typographical capabilities, worse nesting/structuring capabilities, and arguably not any better readability, than 1970’s ROFF and GML?!)

docs/containers-registries.conf.5.md
```diff
@@ -18,80 +18,141 @@ VERSION 2 is the latest format of the `registries.conf` and is currently in
...
+
+`unqualified-search-registries`
+: An array of host[:port] (not more general prefix!) entries to use when pulling
+an unqualified image, in order.
+
+### NAMESPACED `[[registry]]` SETTINGS
+
+The following settings may differ among different registries, and namespaces/repositories
+within a registry.  The configuration is represented as a sequence of `[[registry]]`
+entries containging the specific settings.
```
vrothberg: Non-blocking: This description seems specific to the implications of the `prefix` which may confuse readers. I'd prefer something more general describing that registries (and namespaces) can be configured in a TOML table array, each entry starting with `[[registry]]`. Then we could merge the following sections and go through the fields and sub-tables one after another.

docs/containers-registries.conf.5.md
```diff
@@ -18,80 +18,141 @@ VERSION 2 is the latest format of the `registries.conf` and is currently in
...
+
+#### Choosing a `[[registry]]` entry
+
+A `[[registry]]` entry is chosen based on its `prefix` field; only one `[[registry]]`
+entry, the one with the best-matching `prefix`, is used.
```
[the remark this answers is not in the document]
vrothberg: Maybe we can merge this sentence into the description of `prefix`?

docs/containers-registries.conf.5.md
```diff
@@ -18,80 +18,141 @@ VERSION 2 is the latest format of the `registries.conf` and is currently in
...
-this attribute makes sure to avoid such a situation but requires the primary
-registry to always be available if references by tag are used.
-
-In addition, at the top level, an `unqualified-search-registries` array can be
-specified; this is a list of host[:port] (not more general prefix!) entries to
-use when pulling an unqualified image, in order.
+### GLOBAL SETTINGS
+
+`unqualified-search-registries`
+: An array of host[:port] (not more general prefix!) entries to use when pulling
```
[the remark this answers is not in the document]
TomSweeneyRedHat: I'd actually remove the whole parens thing entirely.  and just go with "... host[:port] entries"

docs/containers-registries.conf.5.md
TomSweeneyRedHat: s/containging/containing/

docs/containers-registries.conf.5.md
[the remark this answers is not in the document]
mtrmac: Thanks, I dropped the second sentence here (but still left a bit as an introduction).

docs/containers-registries.conf.5.md
```diff
@@ -18,80 +18,141 @@ VERSION 2 is the latest format of the `registries.conf` and is currently in
...
+    Example: Given
+    ```
+    prefix = "example.com/foo"
+    location = "internal-registry-for-example.net/bar"
+    ```
+    requests for the image `example.com/foo/myimage:latest` will actually work with the
+    `internal-registry-for-example.net/bar/myimage:latest` image.
+
+`mirror`
+: An array of TOML entries specifiying (possibly-partial) mirrors for the
```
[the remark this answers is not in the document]
mtrmac: I have (finally) read the TOML spec (“obvious, minimal”… not), and updated the text throughout to say array / TOML table.  (“TOML table” instead of just “table” because the underlying meaning of “hash table” is implementation-focused and completely divorced from the general meaning of the “table” word).

docs/containers-registries.conf.5.md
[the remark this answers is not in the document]
mtrmac: Using “_host_[`:`_port_] registries” (not “entries”), to be specific about the semantics of the entries.

docs/containers-registries.conf.5.md
mtrmac: Reworded to start with saying that there is an array of `[[registry]]` TOML tables, and to let the “Choosing a [[registry]] TOML table” section speak for itself.

For now I’m leaving the “choosing table” / “misc. settings” / “remapping and mirroring” sections separate, because the “choosing table” one must be understood by everyone, while the other two are optional — and in particular the whole concept of remapping/mirroring can be skipped by users who don’t need that; mixing `location` among all the other options without further context seems confusing to me.

umohnani8:
Changes LGTM.
@mtrmac @vrothberg are we waiting on anything else before getting this in? Also can we cut a new contaiers/image version once this gets in to unblock the buildah and cri-o PRs?

mtrmac:
> @mtrmac @vrothberg are we waiting on anything else before getting this in?

Tests need to pass.  Current iteration of that is https://github.com/containers/skopeo/pull/673 .
> Also can we cut a new containers/image version once this gets in to unblock the buildah and cri-o PRs?

That’s the plan.

rhatdan:
We have a RHEL8.1 deadline on Friday, for Podman and Buildah,  So the sooner the better.

PR #619, opened 2019-04-25
Add keyring package

QiWang19:
Add package from https://github.com/jsipprell/keyctl to store authentication for `podman login` to kernel user keyring.
close [#2167](https://github.com/containers/libpod/issues/2167)

### DIFF
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -35,6 +35,8 @@ var (
 	// ErrNotLoggedIn is returned for users not logged into a registry
 	// that they are trying to logout of
 	ErrNotLoggedIn = errors.New("not logged in")
+	// ErrNotSupported is returned for unsupported methods
+	ErrNotSupported = errors.New("not supported")
 )
 
 // SetAuthentication stores the username and password in the auth.json file
@@ -43,6 +45,13 @@ func SetAuthentication(sys *types.SystemContext, registry, username, password st
 		if ch, exists := auths.CredHelpers[registry]; exists {
 			return false, setAuthToCredHelper(ch, registry, username, password)
 		}
+
+		err := setAuthToKernelKeyring(registry, username, password)
+		if err == nil {
+			logrus.Debugf("credentials for (%s, %s) were stored in the kernel keyring\n", registry, username)
+			return false, nil
+		}
+		logrus.Debugf("failed to authenticate with the kernel keyring, falling back to authfiles.")
 
 		creds := base64.StdEncoding.EncodeToString([]byte(username + ":" + password))
 		newCreds := dockerAuthConfig{Auth: creds}
@@ -58,6 +67,12 @@ func GetAuthentication(sys *types.SystemContext, registry string) (string, strin
 	if sys != nil && sys.DockerAuthConfig != nil {
 		logrus.Debug("Returning credentials from DockerAuthConfig")
 		return sys.DockerAuthConfig.Username, sys.DockerAuthConfig.Password, nil
+	}
+
+	username, password, err := getAuthFromKernelKeyring(registry)
+	if err == nil {
+		logrus.Debug("returning credentials from kernel keyring")
+		return username, password, nil
 	}
 
 	dockerLegacyPath := filepath.Join(homedir.Get(), dockerLegacyHomePath)
@@ -96,6 +111,14 @@ func RemoveAuthentication(sys *types.SystemContext, registry string) error {
 		if ch, exists := auths.CredHelpers[registry]; exists {
 			return false, deleteAuthFromCredHelper(ch, registry)
 		}
+
+		// Next try kernel keyring
+		err := deleteAuthFromKernelKeyring(registry)
+		if err == nil {
+			logrus.Debugf("credentials for %s were deleted from the kernel keyring", registry)
+			return false, nil
+		}
+		logrus.Debugf("failed to delete credentials from the kernel keyring, falling back to authfiles")
 
 		if _, ok := auths.AuthConfigs[registry]; ok {
 			delete(auths.AuthConfigs, registry)
--- a/pkg/docker/config/config_linux.go
+++ b/pkg/docker/config/config_linux.go
@@ -1,0 +1,79 @@
+package config
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/containers/image/pkg/keyctl"
+	"github.com/pkg/errors"
+)
+
+func getAuthFromKernelKeyring(registry string) (string, string, error) {
+	userkeyring, err := keyctl.UserKeyring()
+	if err != nil {
+		return "", "", err
+	}
+	key, err := userkeyring.Search(genDescription(registry))
+	if err != nil {
+		return "", "", err
+	}
+	authData, err := key.Get()
+	if err != nil {
+		return "", "", err
+	}
+	parts := strings.SplitN(string(authData), "\x00", 2)
+	if len(parts) != 2 {
+		return "", "", nil
+	}
+	return parts[0], parts[1], nil
+}
+
+func deleteAuthFromKernelKeyring(registry string) error {
+	userkeyring, err := keyctl.UserKeyring()
+
+	if err != nil {
+		return err
+	}
+	key, err := userkeyring.Search(genDescription(registry))
+	if err != nil {
+		return err
+	}
+	return key.Unlink()
+}
+
+func setAuthToKernelKeyring(registry, username, password string) error {
+	keyring, err := keyctl.SessionKeyring()
+	if err != nil {
+		return err
+	}
+	id, err := keyring.Add(genDescription(registry), []byte(fmt.Sprintf("%s\x00%s", username, password)))
+	if err != nil {
+		return err
+	}
+
+	// sets all permission(view,read,write,search,link,set attribute) for current user
+	// it enables the user to search the key after it linked to user keyring and unlinked from session keyring
+	err = keyctl.SetPerm(id, keyctl.PermUserAll)
+	if err != nil {
+		return err
+	}
+	// link the key to userKeyring
+	userKeyring, err := keyctl.UserKeyring()
+	if err != nil {
+		return errors.Wrapf(err, "error getting user keyring")
+	}
+	err = keyctl.Link(userKeyring, id)
+	if err != nil {
+		return errors.Wrapf(err, "error linking the key to user keyring")
+	}
+	// unlink the key from session keyring
+	err = keyctl.Unlink(keyring, id)
+	if err != nil {
+		return errors.Wrapf(err, "error unlinking the key from session keyring")
+	}
+	return nil
+}
+
+func genDescription(registry string) string {
+	return fmt.Sprintf("container-registry-login:%s", registry)
+}
--- a/pkg/docker/config/config_unsupported.go
+++ b/pkg/docker/config/config_unsupported.go
@@ -1,0 +1,16 @@
+// +build !linux
+// +build !386 !amd64
+
+package config
+
+func getAuthFromKernelKeyring(registry string) (string, string, error) {
+	return "", "", ErrNotSupported
+}
+
+func deleteAuthFromKernelKeyring(registry string) error {
+	return ErrNotSupported
+}
+
+func setAuthToKernelKeyring(registry, username, password string) error {
+	return ErrNotSupported
+}
--- a/pkg/keyctl/key.go
+++ b/pkg/keyctl/key.go
@@ -1,0 +1,65 @@
+// Copyright 2015 Jesse Sipprell. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// +build linux
+// +build 386 amd64
+
+package keyctl
+
+import (
+	"unsafe"
+)
+
+// Key represents a single key linked to one or more kernel keyrings.
+type Key struct {
+	Name string
+
+	id, ring keyID
+	size     int
+}
+
+// ID returns the 32-bit kernel identifier for a specific key
+func (k *Key) ID() int32 {
+	return int32(k.id)
+}
+
+// Get the key's value as a byte slice
+func (k *Key) Get() ([]byte, error) {
+	var (
+		b        []byte
+		err      error
+		sizeRead int
+	)
+
+	if k.size == 0 {
+		k.size = 512
+	}
+
+	size := k.size
+
+	b = make([]byte, int(size))
+	sizeRead = size + 1
+	for sizeRead > size {
+		r1, _, err := keyctl(keyctlRead, uintptr(k.id), uintptr(unsafe.Pointer(&b[0])), uintptr(size))
+		if err != nil {
+			return nil, err
+		}
+
+		if sizeRead = int(r1); sizeRead > size {
+			b = make([]byte, sizeRead)
+			size = sizeRead
+			sizeRead = size + 1
+		} else {
+			k.size = sizeRead
+		}
+	}
+	return b[:k.size], err
+}
+
+// Unlink a key from the keyring it was loaded from (or added to). If the key
+// is not linked to any other keyrings, it is destroyed.
+func (k *Key) Unlink() error {
+	_, _, err := keyctl(keyctlUnlink, uintptr(k.id), uintptr(k.ring))
+	return err
+}
--- a/pkg/keyctl/keyring.go
+++ b/pkg/keyctl/keyring.go
@@ -1,0 +1,73 @@
+// Copyright 2015 Jesse Sipprell. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// +build linux
+// +build 386 amd64
+
+// Package keyctl is a Go interface to linux kernel keyrings (keyctl interface)
+package keyctl
+
+// Keyring is the basic interface to a linux keyctl keyring.
+type Keyring interface {
+	ID
+	Add(string, []byte) (*Key, error)
+	Search(string) (*Key, error)
+}
+
+type keyring struct {
+	id keyID
+}
+
+// ID is unique 32-bit serial number identifiers for all Keys and Keyrings have.
+type ID interface {
+	ID() int32
+}
+
+// Add a new key to a keyring. The key can be searched for later by name.
+func (kr *keyring) Add(name string, key []byte) (*Key, error) {
+	r, err := addkey("user", name, key, int32(kr.id))
+	if err == nil {
+		key := &Key{Name: name, id: keyID(r), ring: kr.id}
+		return key, nil
+	}
+	return nil, err
+}
+
+// Search for a key by name, this also searches child keyrings linked to this
+// one. The key, if found, is linked to the top keyring that Search() was called
+// from.
+func (kr *keyring) Search(name string) (*Key, error) {
+	id, err := searchKeyring(kr.id, name, "user")
+	if err == nil {
+		return &Key{Name: name, id: id, ring: kr.id}, nil
+	}
+	return nil, err
+}
+
+// ID returns the 32-bit kernel identifier of a keyring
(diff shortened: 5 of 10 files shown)


rhatdan:
@QiWang19 Looks like you are missing some vendors.

vrothberg:
https://github.com/zalando/go-keyring requires dbus which does not seem very portable. Imagine this running in a container.

vrothberg:
> https://github.com/zalando/go-keyring requires dbus which does not seem very portable. Imagine this running in a container.

@rhatdan WDYT about the dbus requirement? I think we can use the zalando lib for OS X and Windows, but for Linux I would prefer using the syscalls directly to avoid a dbus dependency. We need only a very limited set of the entire keyrings functionality, so I'd go for writing a package here in containers/image.

Also adding @mtrmac to share thoughts.

rhatdan:
Yes I am not interested in the dbus bindings, I would rather to it directly.

mtrmac:
More importantly, https://github.com/containers/libpod/issues/2167 asks for support of the _kernel keyring_ (**keyctl**(2) etc.), which is completely different from the per-user userspace keyring (FreeDesktop secrets) supported by https://github.com/zalando/go-keyring ; it’s not just a matter of using/not using D-Bus, it’s a completely different store.

As far as the implementation goes, it’s not obvious to me that this should be a hard-coded silently-used deafault like this; using the `credHelpers` configuration, or even more likely, a global `credsStore` option, would clearly express users’ preference where to store the credentials.  (E.g., if the user wanted to use the kernel keyring, and expected the credentials never to hit the disk, and some permission error caused the use of the kernel keyring to fail, the code automatically falls back to files.)

And then we would talk about whether to implement the helpers out-of-process using that protocol, or in-process by specially recognizing some keywords; both is plausible.

QiWang19:
still need this PR https://github.com/zalando/go-keyring  to support OS X and Windows?

mtrmac:
OS X and Windows don’t have **keyctl**(2), so containers/libpod#2167 is fundamentally impossible for them.

(The kernel keyring is fundamentally ephemeral, a reboot, or probably a log out, clears it; the keyrings supported by go-keyring are persistent and intended for ~single-sign-on (logging into the computer makes the credentials available).

Sure, we can also have a conversation about which one of these is more valuable (and I personally lean towards ephemeral login state being too much hassle for not that much security), but if this PR is supposed to support the kernel keyring or something equivalent, and it just doesn’t.

vrothberg:
> still need this PR using https://github.com/zalando/go-keyring to support OS X and Windows?

We can defer that to another PR and focus just on Linux here. I think it would nice to also support the OS X keyring (and whatever Windows does) but that might be out of scope.

@QiWang19, I suggest to first add the code to support the kernel keyring. There are go libraries in the wild that offer bindings for keyctl but I would prefer to have a dedicated (and minimal) package somewhere here in containers/image. Once that's in place, we can check the details how we can integrate this into the stack (see @mtrmac's comment https://github.com/containers/image/pull/619#issuecomment-491948636).

QiWang19:
yes, I think I can add a package to use kernel keyring first and then decide how to apply it.

QiWang19:
Grab files from https://github.com/jsipprell/keyctl to add session keyring. Is this kind of keyring that we should use?

rhatdan:
Yes I believe so.

rhatdan:
I would prefer this to be primary and not rely on external helpers.

mtrmac:
> I don't want kernel keyring to be configured under `credHelpers`. Can I use the kernel keyring as a default choice, if credHelpers is not configured, then add username &password to kernel keyring, if an error occurs they will be added to the authfile?

I am, hypothetically, a bit worried that this “if an error occurs” state could cause the user to end up with two credential locations at the same time, and unclear rules about when data is recorded in which one and how to manage it — but that is hypothetical and strongly depends on the actual implementation; arguably kernel facilities should very rarely fail, and if they do, they should fail with a clear reason.

QiWang19:
@mtrmac PTAL

mtrmac:
- This code must be automatically disabled on non-Linux platforms; AFAICS the current implementation won’t compile.

rhatdan:
@mtrmac - @vrothberg Asked her to copy the limited functions into this Repo rather then vendoring in the entire repo. (I believe)

vrothberg:
> @mtrmac - @vrothberg Asked her to copy the limited functions into this Repo rather then vendoring in the entire repo. (I believe)

Yes, I asked to have a minimal package here. The required code for our purpose is small in size and complexity. Having it here allows for changing it to our needs if needed and removes the burden of creating new source packages for downstream packaging (mostly Debian).

vrothberg:
I thought keyrings have been namespaced with v5.1 but it seems that David Howell's patch set hasn't been merged yet. Meaning, I browsed the kernel git history and could not find the patches or related changes to the code. @rhatdan, do you know something about the current state? Maybe, I just failed reading the history.

In any case, we need to somehow find a solution for kernels that don't namespace keyrings where we must not store credentials there to avoid another user NS (e.g., containers) from accessing them. I *think* that a config option could help here. Some enterprise kernels may receive a backport, where a version check would not work. Once the patch set is merged, we could add a kernel-version check.

Another thing: according to https://patchwork.kernel.org/patch/10813273/, each user namespace has its own keyring list, which gets deleted with the user NS. This means that we need to read/write credentials **before** joining the user NS when running rootless. I am really not sure how much work that would be.

@giuseppe, what's your take on that?

rhatdan:
They are not namespaced yet, but they are blocked from access inside of non privileged containers.  I am the one on David, spuring him to allow the use of namespaced separated kerings.

rhatdan:
BTW There are other uses of keyrings like Kerberos that we want to block containers from accessing.   This is nothing particular about this PR.

@QiWang19 I continued to work on this last night and found that it was not working on it the way I liked. So she will be updating this PR.

Basically this PR makes the keyring a session keyring, which means that when you went to another terminal the user would not have access to the keys.  We need to make it a User keyring, but this added some complexity, that we were trying to work out late yesterday.

QiWang19:
Updated and made changes in the package to add the key to the user keyring.

rhatdan:
@QiWang19 Can you verify that you can see the keyring from multiple different sessions.
Easy way to check is to run
podman login from gnome terminal and then to try to use it from your vscode or from a `terminator window`
Try same tests for two sudo sessions.
FInally make sure that if you run the commands as non root, and then run sudo, that you don't see the credentials.  And the opposite.  Do podman login as root and make sure you can't use the credentials as non-root.

QiWang19:
I've tried this in podman. It works the way as you commented.
- the credential can be used in the different terminal session
- non-root podman can't pull if only login with root, and vice versa.

QiWang19:
This package can now create a user key and link it to user keyring. Do I need to keep this open? or needs more review

rhatdan:
@mtrmac @vrothberg PTAL

pkg/docker/config/config_linux.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	key, err := userkeyring.Search(genDescription(registry))
+	if err != nil {
+		return err
+	}
+	return key.Unlink()
+}
+
+func setAuthToKernelKeyring(registry, username, password string) error {
+	keyring, err := keyctl.SessionKeyring()
```
vrothberg: Why are we ignoring the error here? If that's desired, I suggest adding a comment and do `if err != nil { return }` instead.

pkg/docker/config/config_linux.go
```diff
@@ -0,0 +1,81 @@
...
+
+		// Set the user permisson to link the key to userkeyring later
+		err = keyctl.SetPerm(id, keyctl.PermUserAll)
+		if err != nil {
+			return err
+		}
+		// link the key to userKeyring
+		userKeyring, err := keyctl.UserKeyring()
+		if err != nil {
+			fmt.Println("user kring", err)
```
vrothberg: This looks like an in-development artefact? If we want to keep them, we need to use `logrus` and be a bit more descriptive to make it easier to spot when reading the wall of debugging messages in our stack.

pkg/docker/config/config_linux.go
```diff
@@ -0,0 +1,81 @@
...
+			return err
+		}
+		err = keyctl.Link(userKeyring, id)
+		if err != nil {
+			fmt.Println("link err", err)
```
vrothberg: Same here.

pkg/docker/config/config_linux.go
```diff
@@ -0,0 +1,81 @@
...
+			return err
+		}
+		// unlink the key from session keyring
+		err = keyctl.Unlink(keyring, id)
+		if err != nil {
+			fmt.Println("unlink session", err)
```
vrothberg: And here.

pkg/docker/config/config_linux.go
vrothberg: Reading the code above, this error shouldn't be ignored here but by the callers.

pkg/docker/config/config.go
```diff
@@ -44,6 +46,11 @@ func SetAuthentication(sys *types.SystemContext, registry, username, password st
 			return false, setAuthToCredHelper(ch, registry, username, password)
 		}
 
+		err := setAuthToKernelKeyring(registry, username, password)
```
vrothberg: Can we add a `logrus.Debug()` here to indicate what's happening? We just recently added some logging for where the creds come from, and I think it can be useful here as well.

pkg/docker/config/config.go
```diff
@@ -97,6 +104,11 @@ func RemoveAuthentication(sys *types.SystemContext, registry string) error {
 			return false, deleteAuthFromCredHelper(ch, registry)
 		}
 
+		err := deleteAuthFromKernelKeyring(registry)
```
vrothberg: Same here: logging would be great.

pkg/docker/config/config.go
```diff
@@ -252,6 +264,11 @@ func findAuthentication(registry, path string, legacyFormat bool) (string, strin
 		return getAuthFromCredHelper(ch, registry)
 	}
 
+	username, password, err := getAuthFromKernelKeyring(registry)
```
vrothberg: And definitely some logging here.

pkg/docker/config/config_linux.go
```diff
@@ -0,0 +1,81 @@
...
+	keyring, err := keyctl.SessionKeyring()
+
+	if err == nil {
+		id, err := keyring.Add(genDescription(registry), []byte(fmt.Sprintf("%s:%s", username, password)))
+		if err != nil {
+			return err
+		}
+
+		// Set the user permisson to link the key to userkeyring later
+		err = keyctl.SetPerm(id, keyctl.PermUserAll)
```
vrothberg: I think this call is worth a comment. `UserAll` can trick a reader into believing that all users are affected but it means that the current user has all rights for this specific key (view, read, write, search, link, set attribute).

pkg/docker/config/config.go
vrothberg: The call to `setAuthToKernelKeyring(...)` should be moved to `GetAuthentication(...)`. `findAuthentication(...)` is meant for lookups in specific files. At the moment, we contact the keyring for each item in the slice of paths.

vrothberg: And lets add a comment explaining the fall-through to using the authfiles, in case an error occured (and `logrus.Debug(...)` the error).

pkg/docker/config/config_linux.go
```diff
@@ -0,0 +1,81 @@
...
+		return err
+	}
+	return key.Unlink()
+}
+
+func setAuthToKernelKeyring(registry, username, password string) error {
+	keyring, err := keyctl.SessionKeyring()
+
+	if err == nil {
+		id, err := keyring.Add(genDescription(registry), []byte(fmt.Sprintf("%s:%s", username, password)))
```
vrothberg: The format of `user:password` is dangerous as `user` and `password` could contain `:`. Luckily, we are using a byte slice, so we can use a null byte (i.e., `\0`) for splitting. We can be sure that the null byte won't be used in username or password.

pkg/docker/config/config_linux.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	key, err := userkeyring.Search(genDescription(registry))
+	if err != nil {
+		return "", "", err
+	}
+	authData, err := key.Get()
+	if err != nil {
+		return "", "", err
+	}
+	parts := strings.SplitN(string(authData), ":", 2)
```
vrothberg: Splitting at `:` is dangerous (see comment below).

pkg/docker/config/config.go
QiWang19: I think you mean that I should move `getAuthFromKernelKeyring`, not `setAuthToKernelKeyring(...)` :slightly_smiling_face: 

> The call to `setAuthToKernelKeyring(...)` should be moved to `GetAuthentication(...)`. `findAuthentication(...)` is meant for lookups in specific files. At the moment, we contact the keyring for each item in the slice of paths.

vrothberg: :rofl: wow, I scrolled and copied the wrong one. Yes, absolutely, it is `getAuthFromKernelKeyring` what I meant.

QiWang19:
@vrothberg :slightly_smiling_face: thanks for reviewing. Fixed the comments above. PTAL

vrothberg:
Apologies, @QiWang19. This one slipped through my emails! I will have another look tomorrow.

pkg/keyctl/sys_linux.go
```diff
@@ -0,0 +1,126 @@
...
+const (
+	keyctlGetKeyringID keyctlCommand = 0
+	keyctlSetPerm      keyctlCommand = 5
+	keyctlLink         keyctlCommand = 8
+	keyctlUnlink       keyctlCommand = 9
+	keyctlSearch       keyctlCommand = 10
+	keyctlRead         keyctlCommand = 11
+)
+
+var debugSyscalls bool
```
vrothberg: We can remove this along with its users. Looks like development artefact that can only be controlled in the code.

pkg/keyctl/keyring.go
```diff
@@ -0,0 +1,75 @@
...
+type keyring struct {
+	id keyID
+}
+
+// ID is unique 32-bit serial number identifiers for all Keys and Keyrings have.
+type ID interface {
+	ID() int32
+	// Info() (Info, error)
+
+	// private()
```
vrothberg: Those comments can be removed.

pkg/keyctl/keyring_test.go
```diff
@@ -0,0 +1,35 @@
...
+		t.Fatal(err)
+	}
+	data, err := key.Get()
+	if err != nil {
+		t.Fatal(err)
+	}
+	if string(data) != string(token) {
+		t.Errorf("Expected data %v, but get %v", token, data)
+	}
+}
```
vrothberg: Parts of the API are not yet tested (e.g., Unlink, Link, ID). Let's add some more tests to test them in separate functions to increase the test coverage.

vrothberg:
@rhatdan, one thing that still worries me a bit is that the keyrings are not yet namespaced. So they can leak into and from a container. Just tried that with Podman.

rhatdan:
@vrothberg Were you able to see the keyring from inside of podman, or was it blocked by SELinux and other security mechanisms?

vrothberg:
> @vrothberg Were you able to see the keyring from inside of podman, or was it blocked by SELinux and other security mechanisms?

It's blocked by SELinux _and_ SECCOMP, which I consider to be "good enough" for a general use but I am no security expert and leave the decision up to you :^)

rhatdan:
Well we already have this issue, so this just adds a little to the problem.  Hopefully the patches to the kernel will finally get merged which will at least allow us to isolate Keyring via User Namespace in the future.
I think it is good enough.

pkg/docker/config/config.go
```diff
@@ -44,6 +46,13 @@ func SetAuthentication(sys *types.SystemContext, registry, username, password st
 			return false, setAuthToCredHelper(ch, registry, username, password)
 		}
 
+		err := setAuthToKernelKeyring(registry, username, password)
+		if err == nil {
+			logrus.Debugf("credentials for (%s, %s) was stored in the kernel keyring\n", registry, username)
```
vrothberg: s/was/were/

pkg/docker/config/config.go
```diff
@@ -97,6 +112,14 @@ func RemoveAuthentication(sys *types.SystemContext, registry string) error {
 			return false, deleteAuthFromCredHelper(ch, registry)
 		}
 
+		// Next try kernel keyring
+		err := deleteAuthFromKernelKeyring(registry)
+		if err == nil {
+			logrus.Debugf("credentials for %s were deleted from the kernel keyring", registry)
```
vrothberg: The log here looks different to the one from SetAuthentication. Let's add the user here as well to have consistent logs.

pkg/keyctl/sys_linux.go
vrothberg: Let's remove this.

pkg/keyctl/sys_linux.go
```diff
@@ -0,0 +1,122 @@
+// Copyright 2015 Jesse Sipprell. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// +build linux
```
vrothberg: Sorry for not noticing this earlier: we need to add `386` and `amd64` to the build constraints as well. The syscall numbers are hard-coded and currently don't support arm, ppc, etc.

pkg/docker/config/config.go
QiWang19: I noticed the difference, but the username is not available in this `func RemoveAuthentication`. :thinking: 
or would it be better if `deleteAuthFromKernelKeyring(registry)` returns (username, pass, error)?

QiWang19:
@vrothberg PTAL

pkg/docker/config/config.go
vrothberg: Good point! Let´s keep it as is.

pkg/keyctl/sys_linux.go
```diff
@@ -0,0 +1,121 @@
...
+// +build 386 amd64
```
vrothberg: Actually, we need this at a few other places as well :( I'll drop comments at the places.

pkg/keyctl/keyring_test.go
```diff
@@ -0,0 +1,123 @@
+// +build linux
```
vrothberg: Constraints here.

pkg/keyctl/keyring.go
```diff
@@ -0,0 +1,72 @@
+// Copyright 2015 Jesse Sipprell. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// +build linux
```
vrothberg: Constraints here.

pkg/keyctl/key.go
```diff
@@ -0,0 +1,64 @@
+// Copyright 2015 Jesse Sipprell. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// +build linux
```
vrothberg: Constraints here.

pkg/docker/config/config_unsupported.go
```diff
@@ -0,0 +1,15 @@
+// +build !linux
```
vrothberg: // +build !386 !amd64

... to be sure it's not selected for ARM and friends

pkg/keyctl/key.go
QiWang19: do we need constraints in pkg/keyctl/perm.go? :thought_balloon:

QiWang19:
@vrothberg  PTAL :slightly_smiling_face:

rhatdan:
LGTM

PR #621, opened 2019-04-26
Makefile: don't assume that GOBIN in PATH

### LINKED ISSUE #620: Go binary path being assumed to be part of PATH
The [`vendor` target of the `Makefile`](https://github.com/containers/image/blob/54d6f626f672b295efc65f9336ce200d207dc264/Makefile#L54) is assuming that the directory where the Go binaries get installed (`GOBIN` or `GOPATH/bin`) is part of the `PATH`.

But at least that target doesn't work if that's not the case, like in one of my build environments.

Silvanoc:
The execution of 'trash' in the 'vendor' target of the 'Makefile'
assumes that GOBIN is part of PATH and therefore 'trash' can be called
directly. This commit enables executing 'trash' if neither GOBIN nor
GOPATH/bin are in PATH.

Closes #620

Signed-off-by: Silvano Cirujano Cuesta <silvano.cirujano-cuesta@siemens.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,15 @@ tools.timestamp: Makefile
 	@touch tools.timestamp
 
 vendor: tools.timestamp vendor.conf
-	@trash
+	@if which trash > /dev/null ; then \
+		trash ; \
+	else \
+		if [ "$(GOBIN)" ] ; then \
+			$(GOBIN)/trash ; \
+		else \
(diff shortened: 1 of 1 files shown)


vrothberg:
Thanks! LGTM but needs a rebase.

@mtrmac @rhatdan PTAL

rhatdan:
LGTM

PR #622, opened 2019-04-29
Rename atomic-signatures to containers-signatures

rhatdan:
We no longer use the term atomic, renaming atomic to containers.

### DIFF
--- a/docs/containers-signature.5.md
+++ b/docs/containers-signature.5.md
@@ -1,10 +1,10 @@
-% atomic-signature(5) Atomic signature format
+% container-signature(5) Container signature format
 % Miloslav Trmač
 % March 2017
 
-# Atomic signature format
+# Container signature format
 
-This document describes the format of “atomic” container signatures,
+This document describes the format of container signatures,
 as implemented by the `github.com/containers/image/signature` package.
 
 Most users should be able to consume these signatures by using the `github.com/containers/image/signature` package
@@ -21,7 +21,7 @@ an automated build system as a result of an automated build,
 a company IT department approving the image for production) under a specified _identity_
 (e.g. an OS base image / specific application, with a specific version).
 
-An atomic container signature consists of a cryptographic signature which identifies
+A container signature consists of a cryptographic signature which identifies
 and authenticates who signed the image, and carries as a signed payload a JSON document.
 The JSON document identifies the image being signed, claims a specific identity of the
 image and if applicable, contains other information about the image.
@@ -34,7 +34,7 @@ associating more than one signature with an image.
 
 ## The cryptographic signature
 
-As distributed, the atomic container signature is a blob which contains a cryptographic signature
+As distributed, the container signature is a blob which contains a cryptographic signature
 in an industry-standard format, carrying a signed JSON payload (i.e. the blob contains both the
 JSON document and a signature of the JSON document; it is not a “detached signature” with
 independent blobs containing the JSON document and a cryptographic signature).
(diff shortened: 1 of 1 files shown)


docs/containers-signature.5.md
```diff
@@ -21,7 +21,7 @@ an automated build system as a result of an automated build,
 a company IT department approving the image for production) under a specified _identity_
 (e.g. an OS base image / specific application, with a specific version).
 
-An atomic container signature consists of a cryptographic signature which identifies
+An container signature consists of a cryptographic signature which identifies
```
TomSweeneyRedHat: s/An container/A container/

docs/containers-signature.5.md
```diff
@@ -89,7 +89,7 @@ and if the semantics of the invalid document, as created by such an implementati
 The top-level value of the JSON document MUST be a JSON object with exactly two members, `critical` and `optional`,
 each a JSON object.
 
-The `critical` object MUST contain a `type` member identifying the document as an atomic container signature
+The `critical` object MUST contain a `type` member identifying the document as an container signature
```
TomSweeneyRedHat: s/an container/a container/

docs/containers-signature.5.md
```diff
@@ -112,7 +112,7 @@ and MUST reject unexpected data in the `critical` object, or in the top-level ob
 ```json
 {
     "critical": {
-        "type": "atomic container signature",
+        "type": "container signature",
```
mtrmac: This would immediately break all existing signatures, and is completely impractical.

Introducing a new value, and figuring out some migration plan (still breaking old consumers), is IMHO not at all worth it.

docs/containers-signature.5.md
```diff
@@ -135,7 +135,7 @@ Consumers MUST reject any signature where the `critical` object contains any unr
 
 ### `critical.type`
 
-This MUST be a string with a string value exactly equal to `atomic container signature` (three words, including the spaces).
+This MUST be a string with a string value exactly equal to `container signature` (three words, including the spaces).
```
mtrmac: This would immediately break all existing signatures, and is completely impractical.

Introducing a new value, and figuring out some migration plan (still breaking old consumers), is IMHO not at all worth it.

docs/containers-signature.5.md
```diff
@@ -1,10 +1,10 @@
-% atomic-signature(5) Atomic signature format
+% container-signature(5) Container signature format
 % Miloslav Trmač
 % March 2017
 
-# Atomic signature format
+# Container signature format
 
-This document describes the format of “atomic” container signatures,
+This document describes the format of “container” signatures,
```
mtrmac: I don’t think we need the quotes after dropping the distinguishing qualifier.

docs/containers-signature.5.md
rhatdan: Ok removed this change.

docs/containers-signature.5.md
rhatdan: Removed change.

docs/containers-signature.5.md
rhatdan: Fixed

docs/containers-signature.5.md
rhatdan: Fixed

docs/containers-signature.5.md
rhatdan: Fixed

rhatdan:
Yes the combination of the change to skopeo to install containers/image man pages and this change might need to rework skopeo spec file.

PR #623, opened 2019-05-03
Add comment DockerAuthConfig

QiWang19:
Add comment for DockerAuthConfig. In SystemContext, the fileds value of DockerAuthConfig can be empty.

Related PR [#646](https://github.com/containers/skopeo/pull/646)

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -401,6 +401,7 @@ type ImageInspectInfo struct {
 }
 
 // DockerAuthConfig contains authorization information for connecting to a registry.
+// the value of Username and Password can be empty for accessing the registry anonymously
 type DockerAuthConfig struct {
 	Username string
 	Password string


QiWang19:
@mtrmac What else need to do to pass the CI?

types/types.go
```diff
@@ -486,6 +484,7 @@ type SystemContext struct {
 	// Allow contacting docker registries over HTTP, or HTTPS with failed TLS verification. Note that this does not affect other TLS connections.
 	DockerInsecureSkipTLSVerify OptionalBool
 	// if nil, the library tries to parse ~/.docker/config.json to retrieve credentials
+	// the value of Username and Password in DockerAuthConfig can be empty
```
mtrmac: This would be better placed as a comment in the `DockerAuthConfig` structure, and it should also say what _does it mean_ for the values to be empty.

mtrmac:
@QiWang19 One of the failing CI runs correctly fails because the PR removes two necessary imports from the `types.go` file.

(The other run seems to be an infrastructure failure, pushing an updated version will restart it anyway.)

types/types.go
QiWang19: fixed ✋

PR #593, opened 2019-03-01
pkg/docker/config/testdata: Add credential-file fixtures

wking:
We're primarily interested in maintaing compatibility with user's on-disk configurations, and using fixtures makes it harder to get that wrong.

While I'm touching these tests, also use `t.Run` and `assert.Equal` to make it easier to read both the source and any test failure messages.

PR #624, opened 2019-05-11
Revert "pkg/docker/config/testdata: Add credential-file fixtures"

mtrmac:
This reverts commit 1c0fb51d9b489433a5831972654a62a3fde27b23 (#593 ).

That commit introduces a flake (depending on in which order normalizedAuths are processed):

```
--- FAIL: TestGetAuth (0.01s)
    config_test.go:71: using temporary XDG_RUNTIME_DIR directory: "/var/folders/14/d3kqjd4d57q1bff3sph57ks80000gn/T/test_docker_client_get_auth799396144"
    config_test.go:87: using temporary home directory: "/var/folders/14/d3kqjd4d57q1bff3sph57ks80000gn/T/test_docker_client_get_auth026490575"
    --- FAIL: TestGetAuth/normalize_registry#01 (0.00s)
        Error Trace:    config_test.go:218
        Error:      	Not equal:
                    	expected: "docker"
                    	received: "index"
        Error Trace:    config_test.go:219
        Error:      	Not equal:
                    	expected: "io"
                    	received: "docker.io"
FAIL
FAIL	github.com/containers/image/pkg/docker/config	0.121s
```
While the commit says “add fixtures”, it seems that not a single affected test case has actually been just ported without modification, and there’s no rationale for those modifications. So, while moving to fixtures instead of autogenerating data, and the use of t.Run and assert.Equal, are, as such, certainly valuable, this is all, along with the test suite changes, conflated in a single commit, so it’s not trivially easy to preserve the useful parts without the parts that may need more discussion; I’m afraid I’m going to revert all of this for now.


wking:
Fix for the flake in #626, as an alternative to the revert here.

mtrmac:
@vrothberg PTAL; consider also #626.

PR #625, opened 2019-05-11
Reset SystemContext.DockerAuthConfig when using mirrors at different registries

mtrmac:
This was brought up sometime during the reviews, but was apparently lost.

Consider `registries.conf`:
```
[[registry]]
location = "example.com"
mirror = [
	{ location = "docker.io" },
	{ location = "registry.access.redhat.com" },
	{ location = "registry.redhat.io" },
	{ location = "registry.fedoraproject.org" },
]
```

Then
```
skopeo --registries-conf registries.conf copy --src-creds foo:bar docker://example.com/foo/bar dir:t
```
currently sends the `foo:bar` to every mirror that requires authentication, not just to `example.com`; that seems clearly undesirable.

So, clear `SystemContext.DockerAuthConfig` when using a mirror at a different registry than the one specified by the user; if the mirrors require credentials, they have to be set up in the auth file, e.g. by `podman login`.

@vrothberg @saschagrunert PTAL.

(#588 may be relevant for more discussion about `DockerAuthConfig`)

### DIFF
--- a/docker/docker_image_src.go
+++ b/docker/docker_image_src.go
@@ -16,7 +16,7 @@ import (
 	"github.com/containers/image/pkg/sysregistriesv2"
 	"github.com/containers/image/types"
 	"github.com/docker/distribution/registry/client"
-	"github.com/opencontainers/go-digest"
+	digest "github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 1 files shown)


saschagrunert:
LGTM

vrothberg:
LGTM. I opened https://github.com/containers/image/pull/627 to make CI pass again. `ubuntu:artful` (i.e., 17:10) hit EOL.

PR #628, opened 2019-05-13
Mirror cleanups - code only

mtrmac:
This is the code subset of #617; #617 will be rebased on top of this, to include only the documentation, which needs more work.

Cc: @vrothberg @saschagrunert

### DIFF
--- a/pkg/sysregistriesv2/system_registries_v2.go
+++ b/pkg/sysregistriesv2/system_registries_v2.go
@@ -46,12 +46,10 @@ func (e *Endpoint) RewriteReference(ref reference.Named, prefix string) (referen
 
 	newNamedRef := strings.Replace(refString, prefix, e.Location, 1)
 	newParsedRef, err := reference.ParseNamed(newNamedRef)
-	if newParsedRef != nil {
-		logrus.Debugf("reference rewritten from '%v' to '%v'", refString, newParsedRef.String())
-	}
 	if err != nil {
 		return nil, errors.Wrapf(err, "error rewriting reference")
 	}
+	logrus.Debugf("reference rewritten from '%v' to '%v'", refString, newParsedRef.String())
 	return newParsedRef, nil
 }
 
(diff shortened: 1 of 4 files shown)


pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -39,26 +39,18 @@ type Endpoint struct {
 // endpoints `location` from the `ref` and creates a new named reference from it.
 // The function errors if the newly created reference is not parsable.
 func (e *Endpoint) RewriteReference(ref reference.Named, prefix string) (reference.Named, error) {
-	if ref == nil {
```
mtrmac: From https://github.com/containers/image/pull/617#discussion_r278058332 by @saschagrunert :
> I'm personally not a fan of removing this check since it is a public API and crashing would be no option for me. But this is just my personal non-blocking opinion.

mtrmac: Rationale from the commit message:

> This "should never happen" (famous last words), and even if it did, a panic with a backtrace is more useful than an error returned up the stack and not indicating the exact incorrect caller.

We don’t have many such checks in this repo; a crash (which can, ultimately, still be caught in Go using `recover`) works a bit better for fixing the bug, and this is not C where a risk of silent memory corruption can make users blame the library for being unreliable instead of assigning the blame correctly.

Anyway, I wouldn’t mind dropping this change if others prefer to keep the check.

saschagrunert:
The overall code changes are looking good to me, but one unit test does not to seem happy. :+1:

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: I concur with @mtrmac that such checks are very uncommon in the stack.

vrothberg:
I restarted the job and it's green now. Was just a flake. Going to merge.

Thanks!

PR #629, opened 2019-05-14
registries: add attribute to only pull from mirrors by digests

vrothberg:


vrothberg:
@mtrmac, @mrunalp  PTAL. I named the attribute `mirror-digests-only` but that's certainly up for debate. I did not want to name it "no-tags" or something similar as I find it more idiomatic to describe what something is rather than what it is not.

vrothberg:
... interesting that VSCode likes the tests but `go test -v` fails reproducibly

mtrmac:
> ... interesting that VSCode likes the tests but `go test -v` fails reproducibly

There is one flaky test, and `go test` caches successful results (but supposedly not failures?!) Use `go test -count=1` to disable the cache.

vrothberg:
The issue was that I forget to clean the registries cache in the test. Certainly, it was working when just executing it in isolation but not when testing the entire package.

vrothberg:
@mtrmac, good to merge?

vrothberg:
@mrunalp @mtrmac PTAL

mrunalp:
:+1:

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -71,6 +75,19 @@ type Registry struct {
 	Prefix string `toml:"prefix"`
 }
 
+// EndpointsFromReference returns a slice of Endpoints based on the passed
+// reference.
+func (r *Registry) EndpointsFromReference(ref reference.Named) []Endpoint {
```
mtrmac: Please combine this with `RewriteReference` to have something like
```go
type RewrittenEndpoint struct {
    Endpoint *Endpoint // Or not a pointer, not sure
    Reference reference.Named
}

func (r *Registry) EndpointsForPulling(ref reference.Named) ([]RewrittenEndpoint, error)
```
, ideally with a better name than `RewrittenEndpoint` (`PullSource`? Not much better.)
(And then remove the public `RewriteReference`.)

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -63,6 +63,10 @@ type Registry struct {
 	Blocked bool `toml:"blocked"`
 	// If true, the registry can be used when pulling an unqualified image.
 	Search bool `toml:"unqualified-search"`
+	// If true, mirrors will only be used for digest pulls. Pulling images by
+	// tag can potentially yield different images, depending on which endpoint
+	// we pull from.  Forcing digest-pulls for mirrors avoids that issue.
+	MirrorDigestsOnly bool `toml:"mirror-digests-only"`
```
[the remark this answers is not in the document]
vrothberg: `MirrorByDigestOnly` and the config counter-part `mirror-by-digest-only` sounds good :+1:

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: Nothing that can't be solved with another level of indirection :-)

vrothberg:
Thanks for review. I will update the PR accordingly.

pkg/sysregistriesv2/system_registries_v2.go
mtrmac: Is the “level of indirection” comment targeted at the naming? I’m not sure I understand.

vrothberg: I should not make jokes in comments, apologies.

docs/containers-registries.conf.5.md
```diff
@@ -49,6 +49,13 @@ together with a specified `prefix`.
 
 If `blocked = true` then it is not allowed to pull images from that registry.
 
+If `mirror-digests-only = true`, mirrors will only be used during pulling if the
+image is referred to by digest.  Referencing an image by digest enforces to
+always return the same image, independent if it is being pulled from the main
+registry or a mirror.  Pulling an image referenced by tag could yield different
+results, depending on which mirror or registry it is being pulled from.  Setting
+this attribute makes sure to avoid such a situation.
```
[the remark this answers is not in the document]
vrothberg: Done

pkg/sysregistriesv2/system_registries_v2.go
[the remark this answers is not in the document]
vrothberg: Done

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: `PullSource` sounds good. I could not come up with anything better.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -71,6 +75,42 @@ type Registry struct {
...
+// rewritten according to the registries prefix and the Endpoint's location.
+type PullSource struct {
+	Endpoint  Endpoint
+	Reference reference.Named
+}
+
+// PullSourcesFromReference returns a slice of PullSource's based on the passed
+// reference.
+func (r *Registry) PullSourcesFromReference(ref reference.Named) ([]PullSource, error) {
+	if r.MirrorByDigestOnly {
```
vrothberg: Reviewing myself: This can be refactored to avoid redundant code.

vrothberg: Nope, it's fine.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -71,6 +75,42 @@ type Registry struct {
...
+		// Return the endpoint if we only allow mirror digest pulls and the
+		// passed reference includes a digest.
+		if _, isDigested := ref.(reference.Canonical); isDigested {
```
mtrmac: Shouldn’t this be `!isDigested`? The tests also show that the logic is reversed.

pkg/sysregistriesv2/system_registries_v2.go
mtrmac: I would rather prefer refactoring this to have the conditional only determine the list of endpoints, and to share the `rewriteReference` loop.

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: Complete brain fart on my side. Thanks for catching that!

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -71,6 +75,41 @@ type Registry struct {
...
+
+	if r.MirrorByDigestOnly {
+		// Only use mirrors when the reference is a digest one.
+		if _, isDigested := ref.(reference.Canonical); isDigested {
+			endpoints = append(r.Mirrors, r.Endpoint)
+		} else {
+			endpoints = []Endpoint{r.Endpoint}
+		}
+	} else {
+		endpoints = append(r.Mirrors, r.Endpoint)
```
mtrmac: LGTM overall, one last non-blocking nit: Couldn’t this bit of logic also not be duplicated? E.g.
```go
if _, isDigested := ref.(reference.Canonical); isDigested || !r.MirrorByDigestOnly
```
is a bit ugly but IMHO still better than the duplication.

Feel free to merge this as it is if you prefer the current three-branch version.

vrothberg: For whatever reason the comment did not show up in my emails. I thought about that but was hesitant as the cast is only necessary when MirrorByDigestOnly == true. I'll merge, so we can unblock the other pending PRs.

PR #630, opened 2019-05-14
Add API function `TryUpdatingCache`

saschagrunert:
The function can be used to reload the registries without harming the
cache on failure. On success, the cache will be updated as intended.
This function can be reused by the current `GetRegistries` API.

### DIFF
--- a/pkg/sysregistriesv2/system_registries_v2.go
+++ b/pkg/sysregistriesv2/system_registries_v2.go
@@ -338,11 +338,24 @@ func getConfig(ctx *types.SystemContext) (*V2RegistriesConf, error) {
 	configPath := ConfigPath(ctx)
 
 	configMutex.Lock()
-	defer configMutex.Unlock()
 	// if the config has already been loaded, return the cached registries
 	if config, inCache := configCache[configPath]; inCache {
+		configMutex.Unlock()
 		return config, nil
 	}
+	configMutex.Unlock()
+
+	return TryUpdatingCache(ctx)
+}
+
+// TryUpdatingCache loads the configuration from the provided `SystemContext`
+// without using the internal cache. On success, the loaded configuration will
+// be added into the internal registry cache.
+func TryUpdatingCache(ctx *types.SystemContext) (*V2RegistriesConf, error) {
+	configPath := ConfigPath(ctx)
+
+	configMutex.Lock()
+	defer configMutex.Unlock()
 
 	// load the config
 	config, err := loadRegistryConf(configPath)
--- a/pkg/sysregistriesv2/system_registries_v2_test.go
+++ b/pkg/sysregistriesv2/system_registries_v2_test.go
@@ -445,3 +445,22 @@ func TestPullSourcesFromReference(t *testing.T) {
 	assert.Nil(t, err)
 	assert.Equal(t, 1, len(pullSources))
 }
+
+func TestTryUpdatingCache(t *testing.T) {
+	ctx := &types.SystemContext{
+		SystemRegistriesConfPath: "testdata/try-update-cache-valid.conf",
+	}
+	configCache = make(map[string]*V2RegistriesConf)
+	registries, err := TryUpdatingCache(ctx)
+	assert.Nil(t, err)
+	assert.Equal(t, 1, len(registries.Registries))
+	assert.Equal(t, 1, len(configCache))
+
+	ctxInvalid := &types.SystemContext{
+		SystemRegistriesConfPath: "testdata/try-update-cache-invalid.conf",
+	}
+	registries, err = TryUpdatingCache(ctxInvalid)
+	assert.NotNil(t, err)
+	assert.Nil(t, registries)
+	assert.Equal(t, 1, len(configCache))
+}
--- a/pkg/sysregistriesv2/testdata/try-update-cache-invalid.conf
+++ b/pkg/sysregistriesv2/testdata/try-update-cache-invalid.conf
@@ -1,0 +1,1 @@
+invalid
--- a/pkg/sysregistriesv2/testdata/try-update-cache-valid.conf
+++ b/pkg/sysregistriesv2/testdata/try-update-cache-valid.conf
@@ -1,0 +1,2 @@
+[[registry]]
+location = "registry.com"


vrothberg:
This seems redundant to invalidating the cache and then calling GetRegistries() (minus the error case). Is there a concrete use case for that? I prefer keeping the API small but if there's a non-hypothetical use case, I would not block.

saschagrunert:
> This seems redundant to invalidating the cache and then calling GetRegistries() (minus the error case). Is there a concrete use case for that? I prefer keeping the API small but if there's a non-hypothetical use case, I would not block.

My use case is that I only want to override the cache on successful parsing of the configuration, not in every case.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -278,11 +278,24 @@ func GetRegistries(ctx *types.SystemContext) ([]Registry, error) {
...
 		return registries, nil
 	}
+	configMutex.Unlock()
+
+	return GetRegistriesUncached(ctx)
+}
+
+// GetRegistriesUncached loads the registries from the provided `SystemContext`
+// without using the internal cache. On success, the found registries will be
+// memoized into the internal registry cache.
```
TomSweeneyRedHat: nit;  s/memoized/memorized/ or perhaps "added"

vrothberg:
> My use case is that I only want to override the cache on successful parsing of the configuration, not in every case.

Who would be the consumer of the API?

saschagrunert:
> > My use case is that I only want to override the cache on successful parsing of the configuration, not in every case.
> 
> Who would be the consumer of the API?

The CRI-O live config reload feature. The thing is that I don't want to destroy the cache unless the config is valid.

vrothberg:
Thinking about it, I guess you're pointing to CRI-O's reload function where it *could* make sense. If @mtrmac is on board, I would prefer to add a parameter to `GetRegistries(..)` (e.g., `updateCache bool`). With the next release of c/image, consumers of sysregistriesv2 are very likely to be broken (s/URL/Location and friends).

vrothberg:
... [to be broken], so we could add the parameter.

pkg/sysregistriesv2/system_registries_v2.go
mtrmac: https://en.wikipedia.org/wiki/Memoization ; though maybe that indicates that using simpler words would be easier to understand.

mtrmac:
> Thinking about it, I guess you're pointing to CRI-O's reload function where it _could_ make sense. If @mtrmac is on board, I would prefer to add a parameter to `GetRegistries(..)` (e.g., `updateCache bool`).

Neither of the two APIs match how I imagine the callers work (nor how I see CRI-O working today).

Most callers should _never_ call `GetRegistries`; they would (have c/image/docker implicitly) call `FindRegistry`; what value of `updateCache` should `FindRegistry` use for `GetRegistries(updateCache)`?

For a daemon, I imagine there’s a SIGHUP or similar mechanism to reload configuration (or, perhaps, “reload before starting to process every request, but not every single time while processing it); in that case “reload” is what the caller wants, “reload and return the results” seems like pointless extra return value.

Alternatively, I could imagine a “never cache, reload all the time if the file is valid” configuration — besides being a bit risky, that looks like an extra bit in `SystemContext`, so that it applies to all the implicit `FindRegistry` calls as well.

What _precisely_ is the caller doing and what does it want?

saschagrunert:
> For a daemon, I imagine there’s a SIGHUP or similar mechanism to reload configuration (or, perhaps, “reload before starting to process every request, but not every single time while processing it); in that case “reload” is what the caller wants, “reload and return the results” seems like pointless extra return value.

This API would be for the configuration reload of `registries.conf` during CRI-O runtime. This works by listening on SIGHUP as you already mentioned. My initial approach was to call `InvalidateCache()`, but this API does not return any result and immediately destroys the cache. Afterwards I would have to call `GetRegistries()`, which is currently the only function which repopulates the cache and loads the config from the TOML file. For debugging purposes I would give the user the feedback that the registries are verified and which registries were found. As a user I would like to see that my newly configured registry has been added successfully.

> 
> What _precisely_ is the caller doing and what does it want?

The caller mainly wants to recreate the registries cache, but only on successful parsing of `registries.conf`. Beside this, the function should return a meaningful error and the result of the registry cache repopulation.

vrothberg:
> This API would be for the configuration reload of registries.conf during CRI-O runtime. This works by listening on SIGHUP as you already mentioned. My initial approach was to call InvalidateCache(), but this API does not return any result and immediately destroys the cache. Afterwards I would have to call GetRegistries(), which is currently the only function which repopulates the cache and loads the config from the TOML file. For debugging purposes I would give the user the feedback that the registries are verified and which registries were found. As a user I would like to see that my newly configured registry has been added successfully.

That's pretty much how the API is intended to be used.

> The caller mainly wants to recreate the registries cache, but only on successful parsing of registries.conf.

That is the key difference to the current API but the question remains _why_ that is needed instead of using the current API? Could an empty cache lead to issues?

In case of CRI-O, I can imagine that the SIGHUP reload aims at try-reload semantics, meaning that the data will only be updated if it's valid and without error to avoid race-conditions.

saschagrunert:
> That is the key difference to the current API but the question remains _why_ that is needed instead of using the current API? Could an empty cache lead to issues?

Yes, I think so. Lets imagine the following scenario:

1. Kubernetes pulls images via CRI-O the cache of `registries.conf` is populated and everything works as intended.
2. The admin changes `registries.conf`, all background pulls are fine because the cache serves the registries
3. The admin sends SIGHUP to CRI-O.
- If we now invalidate the cache without pre-checking the configuration, then all pulls from that point in time are failing.
- If we pre-check the configuration and only invalidate the cache on successful validation, then everything should be fine.

> In case of CRI-O, I can imagine that the SIGHUP reload aims at try-reload semantics, meaning that the data will only be updated if it's valid and without error to avoid race-conditions.

Yes, that's exactly the case. I really want to be sure that everything works before changing the internal application state. For the log_level it works the same way right now:
https://github.com/cri-o/cri-o/blob/0c50025d5f0351ae9acbc1fc46d1afd38c25f953/server/config.go#L218-L232

mtrmac:
> > For a daemon, I imagine there’s a SIGHUP or similar mechanism to reload configuration (or, perhaps, “reload before starting to process every request, but not every single time while processing it); in that case “reload” is what the caller wants, “reload and return the results” seems like pointless extra return value.
> 
> This API would be for the configuration reload of `registries.conf` during CRI-O runtime. This works by listening on SIGHUP as you already mentioned. My initial approach was to call `InvalidateCache()`, but this API does not return any result and immediately destroys the cache. Afterwards I would have to call `GetRegistries()`, which is currently the only function which repopulates the cache and loads the config from the TOML file. For debugging purposes I would give the user the feedback that the registries are verified and which registries were found. As a user I would like to see that my newly configured registry has been added successfully.

So, maybe: `TryReloadingConfiguration` (`TryRefreshingCache`? whatever, naming is hard) which either reloads the configuration and succeeds (and if we wanted to complicate things significantly, would return also an indication whether the configuration has changed[1]); or keeps the current cached configuration, if any, and fails.

On success, the code can _optionally_ call `GetRegistries` to see what the changed configuration is, and log that.

---

[1] This is not something an external caller can quite trivially do, because adding an extra member to a structure does not break the API/ABI, but it can invalidate external code which tries to compare two `Registry` values.

mtrmac:
A further thought — do we even want `InvalidateCache` to exist at all, or should there only be that `Try…` API, which preserves the previous cache on failure?

I can see an argument for actually strictly&reliably failing if the configuration is currently invalid, instead of continuing to use some hard-to-describe old configuration and hiding the problem. It’s not very strong, though.

I suppose keeping `InvalidateCache` around doesn’t hurt, so we can default to keeping it, and give callers both options.

pkg/sysregistriesv2/system_registries_v2.go
saschagrunert: Yes, that's what I meant, but I changed to to `added`.

saschagrunert:
> So, maybe: `TryReloadingConfiguration` (`TryRefreshingCache`? whatever, naming is hard) which either reloads the configuration and succeeds (...); or keeps the current cached configuration, if any, and fails.

Yeah, that's what the current implementaion already does, right? Or should I remove the `[]Registries` from the return parameters?
 
> I suppose keeping `InvalidateCache` around doesn’t hurt, so we can default to keeping it, and give callers both options.

I agree and would not remove it.

vrothberg:
> Yeah, that's what the current implementaion already does, right? Or should I remove the []Registries from the return parameters? (From my POV this would over complicate things and double the work done internally.)

I'd go with a `TryUpdatingCache(..) ([]Registries, error)`. Callers can still ignore the first return value if they don't need it.

I also agree with keeping `InvalidateCache()`. It doesn't hurt to keep it around.

saschagrunert:
> I'd go with a `TryUpdatingCache(..) ([]Registries, error)`. Callers can still ignore the first return value if they don't need it.

Renamed the function according to your suggestion.

saschagrunert:
Rebased on top of the latest master branch.

vrothberg:
@mrunalp @mtrmac PTAL

rhatdan:
@saschagrunert Are you still interested in this, can  you rebase?

saschagrunert:
> @saschagrunert Are you still interested in this, can you rebase?

Yes, rebased on top of the latest master branch

rhatdan:
@mtrmac @vrothberg PTAL

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -338,11 +338,24 @@ func getConfig(ctx *types.SystemContext) (*V2RegistriesConf, error) {
...
 	}
+	configMutex.Unlock()
+
+	return TryUpdatingCache(ctx)
+}
+
+// TryUpdatingCache loads the registries from the provided
+// `SystemContext` without using the internal cache. On success, the found
+// registries will be added into the internal registry cache.
+func TryUpdatingCache(ctx *types.SystemContext) ([]Registry, error) {
```
mtrmac: The code now returns `v2Config` value (which is a `V2RegistriesConf`), not `[]Registry`.

BTW it also seems to me that the `TryUpdatingCache` might want to return the full `V2RegistriesConf` value; the caller can always discard unwanted data but getting the desired data would otherwise be not quite as easy. OTOH the downside is that the callers, even callers that don’t care, might need to hard-code knowledge about the config file format. I don’t have a strong preference in this choice, just highlighting something to consider.

pkg/sysregistriesv2/system_registries_v2_test.go
```diff
@@ -445,3 +445,24 @@ func TestPullSourcesFromReference(t *testing.T) {
 	assert.Nil(t, err)
 	assert.Equal(t, 1, len(pullSources))
 }
+
+func TestTryUpdatingCache(t *testing.T) {
+	testConfig = []byte(`
```
mtrmac: The `testConfig` override variable no longer exists; this will need to actually load the data from a fixture file first.

saschagrunert: Makes sense, added the testate and fixed the test.

pkg/sysregistriesv2/system_registries_v2.go
saschagrunert: Fixed the return value. I have no preference about the second point :)

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -338,11 +338,24 @@ func getConfig(ctx *types.SystemContext) (*V2RegistriesConf, error) {
...
 	}
+	configMutex.Unlock()
+
+	return TryUpdatingCache(ctx)
+}
+
+// TryUpdatingCache loads the registries from the provided
+// `SystemContext` without using the internal cache. On success, the found
+// registries will be added into the internal registry cache.
+func TryUpdatingCache(ctx *types.SystemContext) (*V2RegistriesConf, error) {
```
mtrmac: `s/the registries/the configuration/; s/the found registries/the loaded configuration/`, please.

LGTM otherwise (assuming tests pass).

saschagrunert: Changed as suggested :)

saschagrunert:
The CI needs https://github.com/containers/image/pull/670 to succeed.

saschagrunert:
Rebased.

vrothberg:
Green, pulling the trigger. Thanks @saschagrunert :+1:

PR #631, opened 2019-05-17
Add TransportFromImageName procedure

TristanCacqueray:
This change adds a new transports.TransportFromImageName procedure
to get the transport's name of an image without parsing the store
reference.

### DIFF
--- a/transports/alltransports/alltransports.go
+++ b/transports/alltransports/alltransports.go
@@ -22,6 +22,7 @@ import (
 
 // ParseImageName converts a URL-like image name to a types.ImageReference.
 func ParseImageName(imgName string) (types.ImageReference, error) {
+	// Keep this in sync with TransportFromImageName!
 	parts := strings.SplitN(imgName, ":", 2)
 	if len(parts) != 2 {
 		return nil, errors.Errorf(`Invalid image name "%s", expected colon-separated transport:reference`, imgName)
@@ -31,4 +32,15 @@ func ParseImageName(imgName string) (types.ImageReference, error) {
 		return nil, errors.Errorf(`Invalid image name "%s", unknown transport "%s"`, imgName, parts[0])
 	}
 	return transport.ParseReference(parts[1])
+}
+
+// TransportFromImageName converts an URL-like name to a types.ImageTransport or nil when
+// the transport is unknown or when the input is invalid.
+func TransportFromImageName(imageName string) types.ImageTransport {
+	// Keep this in sync with ParseImageName!
+	parts := strings.SplitN(imageName, ":", 2)
+	if len(parts) == 2 {
+		return transports.Get(parts[0])
+	}
+	return nil
 }
--- a/transports/alltransports/alltransports_test.go
+++ b/transports/alltransports/alltransports_test.go
@@ -3,6 +3,7 @@ package alltransports
 import (
 	"testing"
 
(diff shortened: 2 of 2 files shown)


transports/transports_test.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"testing"
+
+	"github.com/containers/image/directory"
+	"github.com/containers/image/storage"
+	"github.com/containers/image/transports"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestTransportFromImageName(t *testing.T) {
```
mtrmac: Having this test is great, but
```
import cycle not allowed in test
package github.com/containers/image/transports (test)
	imports github.com/containers/image/directory
	imports github.com/containers/image/transports
```
Maybe move the test to the c/image/transports/alltransports subpackage.  (Moving the implementation in there as well would also be acceptable; this function does not logically require the alltransports subpackage to be useful, OTOH skopeo imports it anyway, so it makes no difference for it.)

Or, I guess, maybe just drop the test altogether.

transports/transports_test.go
```diff
@@ -0,0 +1,21 @@
...
+	dirTransport := transports.TransportFromImageName("dir:/tmp/test")
+	assert.Equal(t, dirTransport.Name(), directory.Transport.Name())
+	storageTransport := transports.TransportFromImageName("containers-storage:ref:test")
```
mtrmac: Note that the storage transport is optional (can be compiled out using a build tag); the test should not rely on it.

transports/transports_test.go
```diff
@@ -0,0 +1,21 @@
...
+	assert.Equal(t, storageTransport.Name(), storage.Transport.Name())
+	badTransport := transports.TransportFromImageName("unknown:ref:test")
+	assert.Equal(t, badTransport, nil)
+	badTransport2 := transports.TransportFromImageName("unknown")
```
mtrmac: It’s a bit ugly that “valid format, unknown transport” and “invalid format” result in the same return value.

OTOH, this is likely to be very rarely used, and fits the skopeo use pretty well… at the very least, the function documentation should say that it returns `nil` on invalid input as well.

transports/transports.go
```diff
@@ -88,3 +89,12 @@ func ListNames() []string {
 	sort.Strings(names)
 	return names
 }
+
+// TransportFromImageName converts an URL-like name into a types.ImageTransport or nil when unavailable.
+func TransportFromImageName(imageName string) types.ImageTransport {
+	parts := strings.SplitN(imageName, ":", 2)
```
mtrmac: Please add a
```go
// Keep this in sync with alltransports.ParseImageName!
```
comment at the top of the function body.  (And maybe a similar comment to the `ParseImageName` function.)

transports/transports_test.go
TristanCacqueray: I moved the code to alltransports.

transports/transports_test.go
TristanCacqueray: Alright, I dropped that case.

transports/transports_test.go
TristanCacqueray: Done.

transports/transports.go
TristanCacqueray: Done.

PR #632, opened 2019-05-18
Update Version for release of skopeo.

rhatdan:


rhatdan:
@vrothberg @mtrmac PTAL

vrothberg:
I want https://github.com/containers/image/pull/630 and https://github.com/containers/image/pull/629 in this release. Let's wait for them to be merged.

TomSweeneyRedHat:
LGTM assuming happy tests.

rhatdan:
Ok, I also found a bad crash after this. So I think we should wait.

version/version.go
```diff
@@ -6,7 +6,7 @@ const (
 	// VersionMajor is for an API incompatible changes
 	VersionMajor = 1
 	// VersionMinor is for functionality in a backwards-compatible manner
-	VersionMinor = 7
+	VersionMinor = 8
```
mtrmac: BTW if we are at all serious about semantic versioning, this needs to be 2.0.0.

rhatdan: Did API change?

mtrmac: Yes.

rhatdan: Well lets move to 2.0.0 when we have mirroring.

PR #633, opened 2019-05-18
Reimplement sysregistries by calling sysregistriesv2

mtrmac:
Eliminate the old `sysregistries` implementation, call v2 instead (which handles both the v1 and v2 format). This is almost, but not quite, a prerequisite to changing the v2 format to specify search registries explicitly.
    
RFC (AKA “tell me I’m being too lazy and this should have been fixed already”)
- `sysregistries.RegistriesConfPath` continues to exist independently, because `sysregistriesv2` doesn't expose this.  I don't quite like making this value public (it tempts callers to do their own parsing), but it seems valuable enough for error reporting and the like. So, now v1 has an independent implementation that is likely to grow out of sync.  We probably need to make the v2 implementation public, and then call it from here.
- `sysregistries.GetInsecureRegistries` can now return v2 host[:port]/[ns/...]repo values, in addition to the v1 host[:port] values. Looking at existing CRI-O and podman uses, this seems more useful, but strictly speaking this breaks the API.  Alternatively, we could filter out the values here, and port the callers to use v2 directly.

The bulk of the PR is actually modifying the tests in both packages to use ordinary files instead of hooking into the internals of the config parsers (so that we can rip out the internals in v1). See individual commit messages for details.

### DIFF
--- a/pkg/sysregistries/system_registries.go
+++ b/pkg/sysregistries/system_registries.go
@@ -1,11 +1,9 @@
 package sysregistries
 
 import (
-	"io/ioutil"
 	"path/filepath"
-	"strings"
 
-	"github.com/BurntSushi/toml"
+	"github.com/containers/image/pkg/sysregistriesv2"
 	"github.com/containers/image/types"
 )
 
@@ -19,53 +17,12 @@ var systemRegistriesConfPath = builtinRegistriesConfPath
 // DO NOT change this, instead see systemRegistriesConfPath above.
 const builtinRegistriesConfPath = "/etc/containers/registries.conf"
 
-type registries struct {
-	Registries []string `toml:"registries"`
-}
-
-type tomlConfig struct {
-	Registries struct {
-		Search   registries `toml:"search"`
-		Insecure registries `toml:"insecure"`
-		Block    registries `toml:"block"`
-	} `toml:"registries"`
-}
-
-// normalizeRegistries removes trailing slashes from registries, which is a
-// common pitfall when configuring registries (e.g., "docker.io/library/).
-func normalizeRegistries(regs *registries) {
-	for i := range regs.Registries {
-		regs.Registries[i] = strings.TrimRight(regs.Registries[i], "/")
-	}
-}
-
-// Loads the registry configuration file from the filesystem and
-// then unmarshals it.  Returns the unmarshalled object.
-func loadRegistryConf(sys *types.SystemContext) (*tomlConfig, error) {
-	config := &tomlConfig{}
-
-	configBytes, err := ioutil.ReadFile(RegistriesConfPath(sys))
-	if err != nil {
-		return nil, err
-	}
-
-	err = toml.Unmarshal(configBytes, &config)
-	normalizeRegistries(&config.Registries.Search)
-	normalizeRegistries(&config.Registries.Insecure)
-	normalizeRegistries(&config.Registries.Block)
-	return config, err
-}
-
 // GetRegistries returns an array of strings that contain the names
 // of the registries as defined in the system-wide
 // registries file.  it returns an empty array if none are
 // defined
 func GetRegistries(sys *types.SystemContext) ([]string, error) {
-	config, err := loadRegistryConf(sys)
-	if err != nil {
-		return nil, err
-	}
-	return config.Registries.Search.Registries, nil
+	return sysregistriesv2.UnqualifiedSearchRegistries(sys)
 }
 
 // GetInsecureRegistries returns an array of strings that contain the names
@@ -73,11 +30,20 @@ func GetRegistries(sys *types.SystemContext) ([]string, error) {
 // registries file.  it returns an empty array if none are
 // defined
 func GetInsecureRegistries(sys *types.SystemContext) ([]string, error) {
-	config, err := loadRegistryConf(sys)
+	registries, err := sysregistriesv2.GetRegistries(sys)
 	if err != nil {
 		return nil, err
 	}
-	return config.Registries.Insecure.Registries, nil
+	res := []string(nil)
+	for _, reg := range registries {
+		if reg.Insecure {
+			// NOTE: Traditionally, could only contain host[:port] values; this will include arbitrary prefixes as well.
+			// Strictly speaking, it would be more correct to skip namespace/repository/… values here.
+			// Current known users either don't mind (CRI-O) or benefit from returning the full list (podman info).
+			res = append(res, reg.Prefix)
+		}
+	}
+	return res, nil
 }
 
 // RegistriesConfPath is the path to the system-wide registry configuration file
--- a/pkg/sysregistries/system_registries.go
+++ b/pkg/sysregistries/system_registries.go
@@ -1,12 +1,12 @@
 package sysregistries
 
 import (
+	"io/ioutil"
+	"path/filepath"
 	"strings"
 
 	"github.com/BurntSushi/toml"
 	"github.com/containers/image/types"
-	"io/ioutil"
-	"path/filepath"
 )
 
 // systemRegistriesConfPath is the path to the system-wide registry configuration file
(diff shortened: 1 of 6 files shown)


vrothberg:
I don't understand why `sysregistries` needs to call `sysregistriesv2`. I'd very much prefer users of the API to move on to v2 and we can mark `sysregistries` as deprecated.

I see a short-term convenience for users but it feels a bit too much spaghetti to me. Moreover, `sysregistries` would suddenly understand v2-formatted config files and I wouldn't know how to explain the compatibility matrix to users anymore.

Back when we added v2, we wanted a clean cut (at least on the API level) which this PR would revert. The backwards compat to the v1-formatted configs was needed as we would have broken existing deployments otherwise. But we have control over the API.

> sysregistries.RegistriesConfPath continues to exist independently [...]

If we want to keep this for debugging or for error messages, we could add a `sysregistriesv2.ConfPath(*types.SystemContext)`, so users don't have to figure out own their own if the one from the context or the default one is being used.

mtrmac:
> I don't understand why `sysregistries` needs to call `sysregistriesv2`. I'd very much prefer users of the API to move on to v2 and we can mark `sysregistries` as deprecated.

If we are at all serious about a stable API, this kind of forwarding (and more, e.g. using interfaces more often than structures with public fields) is what it will take, everywhere where we would currently just change or delete a function.

Sure, we should mark the v1 functions as deprecated _as well_ [I do wish there was a language-native way to do that instead of a comment convention].

In this particular case, yes, we could leave around an independent unmaintained implementation, but that would be a rare special case. (My original motivation for this is changing the search list configuration, i.e. changing the v1 compatibility code in v2 — and that in turn benefits from having robust v1 tests, and that in turn is better when the tests are shared.)

BTW, because the next version of c/image is going to be v2.0.0, of course we _can_ just drop the functions instead of carrying them around — but if we are ever going to get a stable API, we need to get into the _habit_ of working towards that somehow, and this seems like a small and quick step towards that, and, of all the interfaces we have, one of the very easiest ones to implement seamlessly. I’ve seen enough “we will keep the API stable after this complete rewrite to fix everything that’s wrong with it” promises that I don’t believe them at all (yet, ironically, I do think we will need at least one such highly-visible change, to make at least `ImageDestination` either private or explicitly unstable.)
 
> Moreover, `sysregistries` would suddenly understand v2-formatted config files and I wouldn't know how to explain the compatibility matrix to users anymore.

I don’t think that any user _wants_ v2 configs to be not read by any existing programs. “$app does not understand NS/repo prefixes in v2 config” is, as far as functionality goes, strictly better than “$app refuses to work if you have v2 config”. 

(This may all be hypothetical; is there any non-trivial v1 user around?)

> > sysregistries.RegistriesConfPath continues to exist independently [...]
> 
> If we want to keep this for debugging or for error messages, we could add a `sysregistriesv2.ConfPath(*types.SystemContext)`, so users don't have to figure out own their own if the one from the context or the default one is being used.

I think I said as much, but I’m depending on you or someone else to say that this needs to / does not need to be a part of this PR.

vrothberg:
> > I don't understand why `sysregistries` needs to call `sysregistriesv2`. I'd very much prefer users of the API to move on to v2 and we can mark `sysregistries` as deprecated.
> 
> If we are at all serious about a stable API, this kind of forwarding (and more, e.g. using interfaces more often than structures with public fields) is what it will take, everywhere where we would currently just change or delete a function.

The proposed forwarding here would change the previous behavior of `sysregistries`, so I don't see how this contributes to keeping the API stable (including the parseable formats into the API).

> > Moreover, `sysregistries` would suddenly understand v2-formatted config files and I wouldn't know how to explain the compatibility matrix to users anymore.
> 
> I don’t think that any user _wants_ v2 configs to be not read by any existing programs. “$app does not understand NS/repo prefixes in v2 config” is, as far as functionality goes, strictly better than “$app refuses to work if you have v2 config”.

Existing programs had to be recompiled/updated in any case. As soon as an admin switches /etc/containers/registries.conf to the new format, existing users of `registries` will break regardless of this PR.

Despite my disagreement, I don't want to block this PR. If you want this in, we can go ahead.

> > If we want to keep this for debugging or for error messages, we could add a `sysregistriesv2.ConfPath(*types.SystemContext)`, so users don't have to figure out own their own if the one from the context or the default one is being used.
> 
> I think I said as much, but I’m depending on you or someone else to say that this needs to / does not need to be a part of this PR.

I think Podman was using it until recently but am not aware of other users. I am cool with not having that in v2.

mtrmac:
> > > I don't understand why `sysregistries` needs to call `sysregistriesv2`. I'd very much prefer users of the API to move on to v2 and we can mark `sysregistries` as deprecated.
> > 
> > 
> > If we are at all serious about a stable API, this kind of forwarding (and more, e.g. using interfaces more often than structures with public fields) is what it will take, everywhere where we would currently just change or delete a function.
> 
> The proposed forwarding here would change the previous behavior of `sysregistries`,

How? The v1 tests were not substantially changed at all (only reimplemented so that they allow changing the implementation). Nothing is taken away. A bit of extra forward-compatibility is added, yes; is that bad?.


> > > Moreover, `sysregistries` would suddenly understand v2-formatted config files and I wouldn't know how to explain the compatibility matrix to users anymore.
> > 
> > 
> > I don’t think that any user _wants_ v2 configs to be not read by any existing programs. “$app does not understand NS/repo prefixes in v2 config” is, as far as functionality goes, strictly better than “$app refuses to work if you have v2 config”.
> 
> Existing programs had to be recompiled/updated in any case. As soon as an admin switches /etc/containers/registries.conf to the new format, existing users of `registries` will break regardless of this PR.

Sure, but _assuming a stable API_, an update of c/image can be very transparent; e.g. a high-level project can bump the required c/image version in `go.mod`, without the _immediate_ caller of c/image being any the wiser.

> Despite my disagreement, I don't want to block this PR. If you want this in, we can go ahead.

I don’t feel that strongly about _this_ case; leaving v1 rot and not changing it (with a deprecation warning?), or just removing it for 2.0.0, would be acceptable to me. It’s more a matter of 1) being confident about the planned v2 search path change, and 2) getting into the habit of keeping a stable API while minimizing the amount of maintained code, and gaining experience with the necessary techniques and API design guidelines, on a very low-risk initial instance.

vrothberg:
> > > [...]
> > The proposed forwarding here would change the previous behavior of `sysregistries`,
> 
> How? The v1 tests were not substantially changed at all (only reimplemented so that they allow changing the implementation). Nothing is taken away. A bit of extra forward-compatibility is added, yes; is that bad?.

Not bad at all. I am not arguing against the change per-se but against tagging it as _the_ way to achieve API stability. The nun-functional behavior should be considered as well and until this change v1 API has failed parsing v2 config.

> > > [...]
> > Despite my disagreement, I don't want to block this PR. If you want this in, we can go ahead.
> 
> I don’t feel that strongly about _this_ case; leaving v1 rot and not changing it (with a deprecation warning?), or just removing it for 2.0.0, would be acceptable to me. 

Same here. In case we don't merge this PR, I think that a deprecation warning (and package-level comment) would be a good first step. It could be removed with a following major version in the future.

> It’s more a matter of 1) being confident about the planned v2 search path change, and 2) getting into the habit of keeping a stable API while minimizing the amount of maintained code, and gaining experience with the necessary techniques and API design guidelines, on a very low-risk initial instance.

* I am okay with 1) but would deferring it to a future minor version.
* :100: for 2)

mtrmac:
OK, I’ll separate the v2 test changes (which should be non-controversial, and are a reasonable code cleanup anyway) into a separate PR, and we can keep this PR open for discussions about what to do with v1 in a future version.

mtrmac:
> OK, I’ll separate the v2 test changes (which should be non-controversial, and are a reasonable code cleanup anyway) into a separate PR

https://github.com/containers/image/pull/636

rhatdan:
@mtrmac What is going on with this PR?

mtrmac:
\*shrug\*

We are obviously not removing the subpackage before v2.0.0 now.

We can either merge this (for which there is no consensus), or at least mark the v1 subpackage as deprecated, and add `sysregistriesv2.RegistriesConfPath`. Given all the conversation in here, it seems more practical to keep this PR open than to copy it all into a newly filed issue.

mtrmac:
@vrothberg Now that there will be a v3.0.0 per #658 , shall I updated this to forward even `RegistriesConfPath` to v2 now that it exists, or should I prepare a PR to completely remove the v1 subpackage?

vrothberg:
@mtrmac, I would prefer to drop v1 entirely but feel free to follow your preference.

mtrmac:
@vrothberg https://github.com/containers/image/pull/663 then.

mtrmac:
#663 has removed the package instead.

PR #636, opened 2019-05-27
Sysregistriesv2 test fixtures

mtrmac:
This is a hopefully-noncontroversial subset of #633: Modify `sysregistriesv2` tests to use ordinary files instead of hooking into the internals of the config parsers.


vrothberg:
LGTM

PR #639, opened 2019-06-07
compression: add support for the zstd algorithm

giuseppe:
zstd is a compression algorithm that has a very fast decoder, while
providing also good compression ratios.  The fast decoder makes it
suitable for container images, as decompressing the tarballs is a very
expensive operation.  This is a first step at supporting zstd as we We
don't yet generate zstd layers.

In my testing, copying the Fedora image from a local dir: repository,
the wall clock time passed from ~8s with gzip compression to ~4.5s
with zstd.


giuseppe:
PR for c/storage: https://github.com/containers/storage/pull/363

pkg/compression/compression.go
```diff
@@ -43,6 +49,7 @@ var compressionAlgos = map[string]struct {
 	"gzip":  {[]byte{0x1F, 0x8B, 0x08}, GzipDecompressor},                 // gzip (RFC 1952)
 	"bzip2": {[]byte{0x42, 0x5A, 0x68}, Bzip2Decompressor},                // bzip2 (decompress.c:BZ2_decompress)
 	"xz":    {[]byte{0xFD, 0x37, 0x7A, 0x58, 0x5A, 0x00}, XzDecompressor}, // xz (/usr/share/doc/xz/xz-file-format.txt)
+	"zstd":  {[]byte{0x25, 0xb5, 0x2f, 0xfd}, ZstdDecompressor},           // zstd (http://www.zstd.net)
```
mtrmac: The current spec says
```c
#define ZSTD_MAGICNUMBER            0xFD2FB528    /* valid since v0.8.0 */
```
and the spec also says …8; apparently the last hexadecimal digit has historically been a protocol number, and the last update was 3 years ago.

What do we really want to do here? Using _only_ the 0.5 format seems rather surprising (although I know basically nothing about Zstd).

giuseppe: good point, I've added only the version I've seen generated.  Should we list them all?  We probably want to add at last the current version (0.8.0)

@terrelln do you have any suggestions for us?

terrelln: I recommend only using the latest version. Versions before 0.8 were "prototype" formats. We stabilized zstd's format 3 years ago with version 1.0.0, and it will never change again.

Sadly, Ubuntu has zstd-0.5 in one of its older LTS releases, but we got them to update it to the latest zstd a year ago. I would start with only the latest zstd, and add 0.5 support if requested, which I think is unlikely.

giuseppe:
@mtrmac fixed now, it is using the 0.8 format.

How could we start generating zstd tarballs?  Should `copyBlobFromStream` be extended to allow changing compression type?

mtrmac:
> @mtrmac fixed now, it is using the 0.8 format.

ACK.  Tests fail on Windows, though.

> How could we start generating zstd tarballs? Should `copyBlobFromStream` be extended to allow changing compression type?

Yes, an option in `types.SystemContext` allowing the user to choose the compression format (if compression is used) would work for now; I suppose defaulting to gzip for interoperability.

giuseppe:
Skopeo WIP PR here: https://github.com/containers/skopeo/pull/671

pkg/compression/compression.go
```diff
@@ -43,24 +78,44 @@ var compressionAlgos = map[string]struct {
...
 	"xz":    {[]byte{0xFD, 0x37, 0x7A, 0x58, 0x5A, 0x00}, XzDecompressor}, // xz (/usr/share/doc/xz/xz-file-format.txt)
+	"zstd":  {[]byte{0x28, 0xb5, 0x2f, 0xfd}, ZstdDecompressor},           // zstd (http://www.zstd.net)
+}
+
+// compressors maps an algorithm to its compression function
+var compressors = map[string]CompressorFunc{
+	"gzip":  GzipCompressor,
+	"bzip2": Bzip2Compressor,
+	"xz":    XzCompressor,
+	"zstd":  ZstdCompressor,
```
mtrmac: If the callers need to deal with the strings, please turn them into a typed Go enum.

… and then `DecompressorFunc` and `CompressorFunc` can be removed, and become methods on that typed enum.  (Well, `DecompressorFunc` needs to continue to exist so that we don’t break the API.)

copy/copy.go
```diff
@@ -814,7 +823,25 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
 		// If this fails while writing data, it will do pipeWriter.CloseWithError(); if it fails otherwise,
 		// e.g. because we have exited and due to pipeReader.Close() above further writing to the pipe has failed,
 		// we don’t care.
-		go compressGoroutine(pipeWriter, destStream) // Closes pipeWriter
+		go compressGoroutine(pipeWriter, destStream, desiredCompressionFormat, c.compressionLevel) // Closes pipeWriter
```
mtrmac: (Does the compression level really need to be modifiable in the first place? Most implementations ignore the value, that’s going to be annoying to document and deal with.)

pkg/compression/compression.go
```diff
@@ -43,24 +78,44 @@ var compressionAlgos = map[string]struct {
...
+	if !found {
+		return nil, fmt.Errorf("cannot find compressor for '%s'", name)
+	}
+	return c, nil
 }
 
 // DetectCompression returns a DecompressorFunc if the input is recognized as a compressed format, nil otherwise.
 // Because it consumes the start of input, other consumers must use the returned io.Reader instead to also read from the beginning.
-func DetectCompression(input io.Reader) (DecompressorFunc, io.Reader, error) {
+func DetectCompression(input io.Reader) (string, DecompressorFunc, io.Reader, error) {
```
mtrmac: This breaks API, and would require a major version bump. (Yes, dealing with that is annoying.)

Please turn this into something like a `DetectCompressionFormat` that returns the above-mentioned enum, and a compatibility wrapper.

(I’m still a bit hoping that the recompression can just not be implemented instead :) )

mtrmac: (It would also be worth considering deprecating the public `pkg/compression` package and intending to make it `internal/compression` in 3.0 — but we should still keep the API in the meantime.

OTOH the package is, in principle, reusable and potentially useful outside of c/image; yet again, I can’t remember any external user.

I don’t have a strong opinion on this, and it’s obviously not directly related to this PR.)

copy/copy.go
giuseppe: I've seen very significant differences when using higher values with zstd (i.e. ~60M vs ~90M), but higher compression levels are quite CPU intensive.

Also the compression buffer size has a noticeable impact on the compression

copy/copy.go
```diff
@@ -814,7 +823,25 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
...
+		if err != nil {
+			return types.BlobInfo{}, err
+		}
+		defer s.Close()
+
+		pipeReader, pipeWriter := io.Pipe()
+		defer pipeReader.Close()
+
+		go compressGoroutine(pipeWriter, s, desiredCompressionFormat, c.compressionLevel) // Closes pipeWriter
+
```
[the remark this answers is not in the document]
giuseppe: most of the images I've tried are already gzip compressed, so ` --dest-compress-format zstd` would not have any effect

pkg/compression/compression.go
giuseppe: are you fine with the last version where I changed the function to look like:

```
func GetCompressor(dest io.Writer, name string, level int) (io.WriteCloser, error)
```

so now we don't expose `CompressorFunc` at all.  Do you still want to have a Go enum for the compression algorithms?

giuseppe:
rebased :arrow_up:

giuseppe:
@vrothberg @rhatdan PTAL

vrothberg:
@giuseppe needs another rebase (will review now)

copy/copy.go
```diff
@@ -814,7 +826,25 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
 		// If this fails while writing data, it will do pipeWriter.CloseWithError(); if it fails otherwise,
 		// e.g. because we have exited and due to pipeReader.Close() above further writing to the pipe has failed,
 		// we don’t care.
-		go compressGoroutine(pipeWriter, destStream) // Closes pipeWriter
+		go c.compressGoroutine(pipeWriter, destStream, desiredCompressionFormat) // Closes pipeWriter
+		destStream = pipeReader
+		inputInfo.Digest = ""
+		inputInfo.Size = -1
+	} else if canModifyBlob && c.dest.DesiredLayerCompression() == types.Compress && isCompressed && desiredCompressionFormat != compressionFormat {
```
vrothberg: Luckily the identifiers are pretty expressive but a comment describing the condition(s) would be nice.

vendor.conf
```diff
@@ -49,3 +49,4 @@ github.com/klauspost/cpuid v1.2.0
 github.com/vbauerster/mpb v3.3.4
 github.com/mattn/go-isatty v0.0.4
 github.com/VividCortex/ewma v1.1.1
+github.com/DataDog/zstd 1.x
```
vrothberg: `s/1.x/1.4.0/` to have an explicit version in order to avoid "accidental" updates.

types/types.go
```diff
@@ -510,6 +510,11 @@ type SystemContext struct {
 	// === dir.Transport overrides ===
 	// DirForceCompress compresses the image layers if set to true
 	DirForceCompress bool
+
+	// CompressionFormat is the format to use for the compression of the blobs
+	CompressionFormat string
```
vrothberg: Non-blocking: I somehow would like to have a dedicated type for that and constants for the compression algorithms to have a more idiomatic way of referencing them e.g., `if compressionFormat == compression.Zstd { ... }`.

vrothberg: Instead of `== "zstd"`

giuseppe: added a patch that defines constants for the various supported algorithms

rhatdan:
If we add support for zstd?  Would an image created with this algorythm, still be able to be used on Docker?  Older versions of CRI-O, Buildah, Podman?

types/types.go
vrothberg: LGTM, thanks!

vrothberg:
@giuseppe, could you add tests? Besides that, LGTM.

giuseppe:
> If we add support for zstd? Would an image created with this algorythm, still be able to be used on Docker? Older versions of CRI-O, Buildah, Podman?

no, unfortunately it won't be compatible with other versions of the tools that have no support for zstd.


> @giuseppe, could you add tests? Besides that, LGTM.

I've added some new tests to `copy/copy_test.go`.  Would you like me to test something in particular?

rhatdan:
What error will docker report if it tries to use an image created with this compression?

rhatdan:
@giuseppe Should we open up a request to add this compression type to OCI?

giuseppe:
> @giuseppe Should we open up a request to add this compression type to OCI?

I am not sure if that is covered by OCI.   @vbatts, is there anything we should do in OCI?


> What error will docker report if it tries to use an image created with this compression?

it fails with this error:

```
# docker pull quay.io/giuseppe/fedora:zstd
Trying to pull repository quay.io/giuseppe/fedora ... 
sha256:5a65f2dc778a61a145ed63a27918314e86d40603aea9aedb2214c56e47d97314: Pulling from quay.io/giuseppe/fedora
9fc33571d686: Extracting [==================================================>] 78.68 MB/78.68 MB
failed to register layer: Error processing tar file(exit status 1): archive/tar: invalid tar header
```

giuseppe:
We can address OCI separately, is there anything more blocking the PR?

rhatdan:
Passing travis?

giuseppe:
> Passing travis?

the Skopeo tests are failing as Skopeo is missing the new dependency

rhatdan:
LGTM
@mtrmac PTAL

vbatts:
as for OCI, it _should_ only be needed in the `+` suffix of the mimetype/mediaType, just like `+gzip` is currently

giuseppe:
> as for OCI, it _should_ only be needed in the `+` suffix of the mimetype/mediaType, just like `+gzip` is currently

thanks.  We are producing the wrong manifest right now even for gzip :/  we will need to address that

rebased, PTAL

giuseppe:
anything more blocking this PR?

rhatdan:
@mtrmac PTAL
@giuseppe Can we get the tests to pass, by opening a PR for skopeo?

giuseppe:
@rhatdan tests are passing on Skopeo: https://github.com/containers/skopeo/pull/671

giuseppe:
If you'd like, I can open a PR for Skopeo to add only the dependency

giuseppe:
@mtrmac would you mind taking a look?

giuseppe:
rebased

giuseppe:
@rhatdan @vrothberg how can we move this forward?

pkg/compression/compression.go
mtrmac: Yes, please

- let’s not have an untyped
  ```go
  desiredCompressionFormat = "gzip"
  ```
  when it could use a `compression.Gzip` constant
- A named Go type would prevent accidentally/unnecessarily mixing raw strings and format values.
- Depending on what the type exactly ends up being, making `GetCompressor` a method on the format type might eliminate the
  ```go
  c, found := compressors[name]
		if !found {
  ```
  failure path.  (If not, no harm done.)
- OTOH, that might force us to add a `compression.GetFormatFromString()`, to set types.SystemContext.CompressionFormat` with the typed value — but that would be a good thing, because invalid values would have to be recognized early at the CLI level, instead of deep inside the `compressGoroutine`.

pkg/compression/compression.go
```diff
@@ -35,32 +36,85 @@ func XzDecompressor(r io.Reader) (io.ReadCloser, error) {
 	return ioutil.NopCloser(r), nil
 }
 
+// compressorFunc returns the compressed stream, given a compressed stream.
```
mtrmac: This description does not make sense.

pkg/compression/compression.go
```diff
@@ -35,32 +36,85 @@ func XzDecompressor(r io.Reader) (io.ReadCloser, error) {
...
+// The caller must call Close() on the compressed stream (even if the compressed input stream does not need closing!).
```
mtrmac: The parameter is not a “compressed input stream”.

pkg/compression/compression.go
```diff
@@ -35,32 +36,85 @@ func XzDecompressor(r io.Reader) (io.ReadCloser, error) {
...
+type compressorFunc func(io.Writer, int) (io.WriteCloser, error)
+
+// GzipCompressor is a CompressorFunc for the gzip compression algorithm.
+func GzipCompressor(r io.Writer, level int) (io.WriteCloser, error) {
```
mtrmac: Given the new API compatibility constraints, I’d prefer making as few parts of the API public as possible. We can always make them public in the future if it turns out to be necessary, the reverse is now problematic.

AFAICS we don’t need the individual compression handlers to be public. (Nor, ultimately, `ZstdDecompressor`, although in that case arguably it’s cleaner to keep the shape of the API the same for all compression algorithms.)

pkg/compression/compression.go
```diff
@@ -35,32 +36,85 @@ func XzDecompressor(r io.Reader) (io.ReadCloser, error) {
...
-	"xz":    {[]byte{0xFD, 0x37, 0x7A, 0x58, 0x5A, 0x00}, XzDecompressor}, // xz (/usr/share/doc/xz/xz-file-format.txt)
+	Gzip:  {[]byte{0x1F, 0x8B, 0x08}, GzipDecompressor},                 // gzip (RFC 1952)
+	Bzip2: {[]byte{0x42, 0x5A, 0x68}, Bzip2Decompressor},                // bzip2 (decompress.c:BZ2_decompress)
+	Xz:    {[]byte{0xFD, 0x37, 0x7A, 0x58, 0x5A, 0x00}, XzDecompressor}, // xz (/usr/share/doc/xz/xz-file-format.txt)
+	Zstd:  {[]byte{0x28, 0xb5, 0x2f, 0xfd}, ZstdDecompressor},           // zstd (http://www.zstd.net)
 }
 
-// DetectCompression returns a DecompressorFunc if the input is recognized as a compressed format, nil otherwise.
+// compressors maps an algorithm to its compression function
+var compressors = map[string]compressorFunc{
```
mtrmac: This could just as well be merged with `compressionAlgos` — but see the other comments about turning this into methods.

pkg/compression/compression.go
```diff
@@ -35,32 +36,85 @@ func XzDecompressor(r io.Reader) (io.ReadCloser, error) {
...
+	Gzip:  GzipCompressor,
+	Bzip2: Bzip2Compressor,
+	Xz:    XzCompressor,
+	Zstd:  ZstdCompressor,
+}
+
+// GetCompressor returns the compressor by its name
+func GetCompressor(dest io.Writer, name string, level int) (io.WriteCloser, error) {
```
mtrmac: This does not just return a compressor, it immediately calls it, and hides it from the caller of this function completely.

I’d prefer something like `compression.Format.CompressStream()` or `compression.Format.NewCompressingWriter` (or whatever name fits best), or something like that.

pkg/compression/compression.go
```diff
@@ -35,32 +36,85 @@ func XzDecompressor(r io.Reader) (io.ReadCloser, error) {
...
+	if !found {
+		return nil, fmt.Errorf("cannot find compressor for '%s'", name)
+	}
+	return c(dest, level)
+}
+
+// DetectCompressionFormat returns a DecompressorFunc if the input is recognized as a compressed format, nil otherwise.
 // Because it consumes the start of input, other consumers must use the returned io.Reader instead to also read from the beginning.
-func DetectCompression(input io.Reader) (DecompressorFunc, io.Reader, error) {
+func DetectCompressionFormat(input io.Reader) (string, DecompressorFunc, io.Reader, error) {
```
mtrmac: The returned format should be a named type, not a raw string.

Assuming the “start compressing a stream” operation is turned into a method on the format type, the “start decompressing a stream” operation should also become a method, and in that case `DetectCompressionFormat` should not return a `DecompressorFunc` separately.

pkg/compression/compression_nocgo.go
```diff
@@ -0,0 +1,18 @@
...
+)
+
+// ZstdDecompressor is a DecompressorFunc for the zstd compression algorithm.
+func ZstdDecompressor(r io.Reader) (io.ReadCloser, error) {
+	return nil, fmt.Errorf("zstd not supported on this platform")
+}
+
+// ZstdCompressor is a CompressorFunc for the zstd compression algorithm.
+func ZstdCompressor(r io.Writer, level int) (io.WriteCloser, error) {
+	return nil, fmt.Errorf("zstd not supported on this platform")
```
mtrmac: `zstd is not supported`, in both cases, please.

pkg/compression/compression.go
```diff
@@ -35,32 +36,85 @@ func XzDecompressor(r io.Reader) (io.ReadCloser, error) {
 	return ioutil.NopCloser(r), nil
 }
 
+// compressorFunc returns the compressed stream, given a compressed stream.
+// The caller must call Close() on the compressed stream (even if the compressed input stream does not need closing!).
+type compressorFunc func(io.Writer, int) (io.WriteCloser, error)
+
+// GzipCompressor is a CompressorFunc for the gzip compression algorithm.
+func GzipCompressor(r io.Writer, level int) (io.WriteCloser, error) {
+	if level > 0 {
```
mtrmac: In `pgzip`, unlike `zstd`, AFAICS `0` is a valid level value, different from “use the built-in default”.

Either define a custom scale, or make the level a `*int` or something similar, to clearly express the distinction between the default and a user-specified value, regardless of the semantics of the algorithm.

giuseppe:
@mtrmac thanks for the review.  Comments addressed

giuseppe:
rebased

vrothberg:
@giuseppe, can we use `use github.com/klauspost/compress directly` instead of the DataDog one? We stripped 2.3MB from c/storage in  https://github.com/containers/storage/commit/895d59cf1866e870b18e2508b635fa1d05908903 and I would avoid re-adding it.

giuseppe:
> @giuseppe, can we use `use github.com/klauspost/compress directly` instead of the DataDog one? We stripped 2.3MB from c/storage in [containers/storage@895d59c](https://github.com/containers/storage/commit/895d59cf1866e870b18e2508b635fa1d05908903) and I would avoid re-adding it.

done.  I've used the same wrappers you added for c/storage

TomSweeneyRedHat:
@giuseppe rebase needed.

pkg/compression/compression.go
```diff
@@ -35,32 +36,82 @@ func XzDecompressor(r io.Reader) (io.ReadCloser, error) {
...
 }
 
-// DetectCompression returns a DecompressorFunc if the input is recognized as a compressed format, nil otherwise.
+// Name returns the name for the compression algorithm.
+func (c Algorithm) Name() string {
+	return c.name
+}
+
+// compressionAlgos is an internal implementation detail of DetectCompression
+var compressionAlgos = []Algorithm{
```
mtrmac: I’m afraid I haven’t had time for a full review yet, but #563 would probably (@vrothberg ?) benefit from having public constants for the known algorithms.

vrothberg: Thanks for bringing this up, @mtrmac. Yes, this would be beneficial so we don't need to replicate constants.

mtrmac: Other callers would also benefit from being able to hard-code `types.SystemContext.CompressionFormat = …` without having to use `AlgorithmFromName` and somehow handling errors.

(Yes, I know I asked for making the compressor functions non-public — that was under the assumption that the algorithm IDs would be public. Sorry.)

giuseppe: how would you like it to look like?  Is it just about defining const for the different names?

Is it something that could be done with #563 so we have at least one user for the new consts?

mtrmac: > Is it something that could be done with #563 so we have at least one user for the new consts?

That works for me.

giuseppe:
> @giuseppe, please also open a PR for Skopeo to avoid the CI being red for too long.

I've opened it here: https://github.com/containers/skopeo/pull/671

I'll revendor containers/image and drop the WIP once this is merged

rhatdan:
@mtrmac Can we merge this?  Once this is merged, should we open a PR for Moby to support the new compression algorithm, so we can maintain compatibility.
I would love to have some numbers on this?

Does this increase speed of builds?  Pulls? Pushes?

vrothberg:
I agree with @rhatdan. This screams for a blog! @giuseppe, are you interested in writing one? Once https://github.com/containers/image/pull/563 gets merged (after this one), we can fully support zstd.

giuseppe:
so these are some numbers I've collected:

```bash
#!/bin/sh
sudo rm -rf /tmp/nginx
mkdir /tmp/nginx

for i in gzip zstd; do
./skopeo copy --dest-compress --dest-compress-format $i --dest-compress-level=20 docker://docker.io/nginx dir:/tmp/nginx/$i
done > /dev/null

du --si -s /tmp/nginx/*

for i in gzip zstd; do
./skopeo copy --dest-compress-format $i dir:/tmp/nginx/$i docker://quay.io/giuseppe/nginx:$i
done > /dev/null

printf "\npull...\n"

for i in gzip zstd; do
printf "\n%s\n" $i
time sudo unshare -m ./skopeo copy docker://quay.io/giuseppe/nginx:$i "containers-storage:[overlay@/tmp/nginx/storage/$i+/tmp/nginx/storage/run]localhost/nginx:$i" > /dev/null
done
```

I've got:

```
51M	/tmp/nginx/gzip
44M	/tmp/nginx/zstd

pull...

gzip

real	0m6.314s
user	0m5.486s
sys	0m0.913s

zstd

real	0m5.254s
user	0m3.893s
sys	0m0.772s

```

rhatdan:
Not bad at all.

rhatdan:
@giuseppe Should tests be passing now?

giuseppe:
The Skopeo failures are expected as the dependency is not there yet

vrothberg:
@giuseppe, can you open a PR at Skopeo using this branch of c/image? This way, we can make sure to not introduce a regression. Once the PR here in c/image is merged, you can update the PR at Skopeo.

giuseppe:
> @giuseppe, can you open a PR at Skopeo using this branch of c/image? This way, we can make sure to not introduce a regression. Once the PR here in c/image is merged, you can update the PR at Skopeo.

The PR for Skopeo is here: https://github.com/containers/skopeo/pull/671

vrothberg:
I can't see if  containers/skopeo#671 passes, as Travis builds don't seem to be triggered due to the merge conflicts.

Other than that, LGTM. @mtrmac, @rhatdan, @mrunalp PTAL

giuseppe:
> I can't see if [containers/skopeo#671](https://github.com/containers/skopeo/pull/671) passes as Travis builds don't seem to be triggered due to the merge conflicts.

I've rebased it now

rhatdan:
@giuseppe Still failing?  
@mtrmac PTAL

giuseppe:
@rhatdan tests for the Skopeo PR are passing.

If everything is fine, please let's move this forward

rhatdan:
LGTM

copy/copy.go
```diff
@@ -178,6 +182,9 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
 		// For now, use DestinationCtx (because blob reuse changes the behavior of the destination side more); eventually
 		// we might want to add a separate CommonCtx — or would that be too confusing?
 		blobInfoCache: blobinfocache.DefaultCache(options.DestinationCtx),
+
+		compressionFormat: *options.DestinationCtx.CompressionFormat,
```
vrothberg: The dereference breaks consumers of c/image. We need to change that to avoid nil dereference.

PR #640, opened 2019-06-13
ostree: ignore EINVAL on lsetfilecon_raw

giuseppe:
ignore EINVAL errors when setting the file label.

Closes: https://github.com/containers/libpod/issues/3202

### DIFF
--- a/ostree/ostree_dest.go
+++ b/ostree/ostree_dest.go
@@ -217,7 +217,7 @@ func fixFiles(selinuxHnd *C.struct_selabel_handle, root string, dir string, user
 				fullpathC := C.CString(fullpath)
 				defer C.free(unsafe.Pointer(fullpathC))
 				res, err = C.lsetfilecon_raw(fullpathC, context)
-				if int(res) < 0 {
+				if int(res) < 0 && err != syscall.EINVAL {
 					return errors.Wrapf(err, "cannot setfilecon_raw %s", fullpath)
 				}
 			}


mtrmac:
Why is it safe to silently not set the right SELinux label?

mtrmac:
Note that the labels come from the SELinux policy of the host machine, they are not a part of the container image.

rhatdan:
Is EINVAL correct or was it supposed to be ENOSUP?

rhatdan:
I would think this should be done at a higher level, and this should be passed in as an option, to ignore fixfiles.

mtrmac:
> I would think this should be done at a higher level, and this should be passed in as an option, to ignore fixfiles.

\*shrug\* I suppose that would absolve c/image of the responsibility of doing the right thing, but still, in which case would that “higher lever” caller set that option? The libpod issue seems to suggest that this can happen if the SELinux policy file contexts settings name an non-existent type; in that case the SELinux policy just needs to be fixed, I can’t see why anyone would want to keep using that policy and opt into ignoring that it is broken.

giuseppe:
what we do here is to use the host policy to set a label for file inside of a container image. The error seems to come from setting the label for `/root/sbin/depmod`, is it safe to assume we get EINVAL if there is no policy set?

rhatdan:
@giuseppe I think the issue is probably that the ostree has read in a filecontext based on the image rather then the hosts labelling and then is attempting to set a label that the host kernel does not understand.  In my opinion we should not be doing any labeling in ostree in this case.

mtrmac:
@rhatdan My reading of https://github.com/containers/image/blob/master/ostree/ostree_dest.go does support that, but then I have never actually read the API docs of libselinux and I don’t really have any idea what this line does:
```go
selinuxHnd, err = C.selabel_open(C.SELABEL_CTX_FILE, nil, 0)
```
It does not include any image-dependent value, at least.

rhatdan:
We trick containers into thinking SELinux is disabled so that they will not attempt to set labels while in the container.  I think we need something similar for ostree here.

mtrmac:
> We trick containers into thinking SELinux is disabled so that they will not attempt to set labels while in the container. I think we need something similar for ostree here.

#389 explicitly and intentionally added the current code to do the opposite.

Now, maybe that was wrong, but then we should be removing/replacing the code/feature and not silently ignoring failures the way this PR does.

---

I know just enough about SELinux to think that he two issues about the `EINVAL` have too many unverified assumptions and quick guesses (apparently no-one even knows for sure what the `EINVAL` error complains about?!).  It bothers me that we are already discussing a code change to fix an underproduced, issue with an unknown root cause.

I’d be much more comfortable with reading a coherent, complete, story of 1) what the original code did 2) why it was desirable and reasonable at the time 3) what it got wrong, whether about the use case or about the implementation 4) what we need the code to do now, instead 5) how to get there.

giuseppe:
I've tried to reproduce the issue with different images but I had no luck with that.  I don't think it is worth to spend more time on the ostree backend.  I've opened https://github.com/containers/image/pull/644 that just makes the error clearer.

PR #643, opened 2019-06-13
Release 2.0

mtrmac:
This is a major version bump because the API changes; while we will try to minimize API changes in the future, past experience shows that it is still extremely likely that they will be very frequent, unless the API is significantly restructured and/or large parts of it hidden or declared out of bounds.  Expect the major number to increase several times before this can settle.


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM

TomSweeneyRedHat:
Awesome!  If someone could create a v2.0 release, that would be great.  I just tried but it looks like I don't have privs to do so.

PR #646, opened 2019-06-17
docker: allow deleting OCI images

vrothberg:
Fixes: containers/skopeo/issues/667

### DIFF
--- a/docker/docker_image_src.go
+++ b/docker/docker_image_src.go
@@ -17,6 +17,7 @@ import (
 	"github.com/containers/image/types"
 	"github.com/docker/distribution/registry/client"
 	digest "github.com/opencontainers/go-digest"
+	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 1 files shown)


vrothberg:
CI PR for Skopeo: https://github.com/containers/skopeo/pull/680

@mtrmac @umohnani8 @haircommander @rhatdan PTAL

TomSweeneyRedHat:
LGTM

haircommander:
LGTM

rhatdan:
LGTM

docker/docker_image_src.go
```diff
@@ -384,7 +385,7 @@ func deleteImage(ctx context.Context, sys *types.SystemContext, ref dockerRefere
 	// When retrieving the digest from a registry >= 2.3 use the following header:
 	//   "Accept": "application/vnd.docker.distribution.manifest.v2+json"
 	headers := make(map[string][]string)
-	headers["Accept"] = []string{manifest.DockerV2Schema2MediaType}
+	headers["Accept"] = []string{manifest.DockerV2Schema2MediaType, imgspecv1.MediaTypeImageManifest}
```
mtrmac: This works for that specific bug, but shouldn’t this use all of `manifest.DefaultRequestedManifestMIMETypes`?  Or maybe just call `fetchManifest` explicitly.

mtrmac: Notably this will get out of sync the next time a MIME type is added again.

rhatdan: Is there a function call to get the list of all mimetypes?

mtrmac: This needs to be ~hard-coded in the client; what the server supports does not quite matter if the client can’t deal with it.

See above for the variable hard-coding it in c/image.

vrothberg: Very good point, @mtrmac. I'll spin a new PR here and update https://github.com/containers/skopeo/pull/680.

vrothberg: Eugh ... no idea what's going on. The last commit vndr is 11 days ago, so I can't be caused by some outdated caches in Travis. I think it's time for go modules :^)

vrothberg: ^ the comment was meant for another thread. Mixed up tabs...

PR #650, opened 2019-06-19
Trivial c/image/docker cleanups

mtrmac:
Primarily removes a no-longer-accurate comment; also uses a literal to hopefully be a bit more readable.


mtrmac:
It’s tempting to just use `fetchManifest` in `DeleteImage` to share the code in full, but handling the 404 error was not quite appealing at the moment (so many registries report missing images in so many different ways).

Eventually, consolidating some smarter “image not found” detection would make it very attractive to call `fetchManifest` from `DeleteImage`.

rhatdan:
LGTM

PR #651, opened 2019-06-21
config.go: log where credentials come from

### LINKED ISSUE #510: Add debug logs to show where docker/distribution credentials came from
- Were they provided in `types.SystemContext`?
- Were they read from a config file? If so, from which one?
- Are we using no credentials at all?

jvanz:
Updates the GetAuthentication method adding debug messages showing
from where the credentials come from.

Close #510

Signed-off-by: José Guilherme Vanz <jguilhermevanz@suse.com>

### DIFF
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -56,7 +56,7 @@ func SetAuthentication(sys *types.SystemContext, registry, username, password st
 // If an entry is not found empty strings are returned for the username and password
 func GetAuthentication(sys *types.SystemContext, registry string) (string, string, error) {
 	if sys != nil && sys.DockerAuthConfig != nil {
-		logrus.Debug("Returning credentials from docker auth config")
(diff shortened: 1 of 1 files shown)


pkg/docker/config/config.go
```diff
@@ -76,12 +77,15 @@ func GetAuthentication(sys *types.SystemContext, registry string) (string, strin
 		legacyFormat := path == dockerLegacyPath
 		username, password, err := findAuthentication(registry, path, legacyFormat)
 		if err != nil {
+			logrus.Debugf("Credentials not found")
```
vrothberg: Could we remove this log. It's redundant to the one in line 88 and there's a potential for confusion.

rhatdan: Well he returns right away, so line 88 will never happen?

vrothberg: Thanks for clarifying, @rhatdan. Must have been the Monday morning...

Issue #111, opened 2016-09-30
docker credentials retrieval should check external stores


runcom:

mtrmac:
#339 added support for `credHelpers`; we do not support `credsStore` yet.

greglanthier:
@mtrmac in reply to https://github.com/containers/image/issues/111#issuecomment-339796455 - is this something you'd consider accepting a PR for?


When [Docker-for-Mac](https://docs.docker.com/docker-for-mac/) is installed the `~/.docker/config.json` file contains something similar to the following:

```json
{
        "auths": {
        },
        "HttpHeaders": {
                "User-Agent": "Docker-Client/18.09.0-ce-beta1 (darwin)"
        },
        "credsStore": "osxkeychain"
}
```

which means the `docker-credential-osxkeychain` credential helper is used - at least by the `docker` cli.  That isn't the case when `containers/image` loads `~/.docker/config.json` - which I _think_ can lead to confusion when a Mac user tries to use `skopeo` with their osxkeychain (at least it confused me).

Modifying [dockerConfigFile](https://github.com/containers/image/blob/master/pkg/docker/config/config.go#L24-L27) to look something like this:

```go
type dockerConfigFile struct {
	AuthConfigs map[string]dockerAuthConfig `json:"auths"`
	CredHelpers map[string]string           `json:"credHelpers,omitempty"`
	CredsStore  string                      `json:"credsStore,omitempty"`
}
```

plus some shuffling in the `func findAuthentication(registry, path string, legacyFormat bool) (string, string, error)` seemed like a reasonable place to start - but it looks like some refactoring of associated unit tests would be required as well.

Can you think of a different way to handle this - one that you'd prefer before I look for a way to make the tests pass?  If I understand correctly vendoring more of the Docker cli is less than ideal (which is completely understandable).

mtrmac:
> @mtrmac in reply to [#111 (comment)](https://github.com/containers/image/issues/111#issuecomment-339796455) - is this something you'd consider accepting a PR for?

(support for `credsStore`); sure, that would be great.

> Modifying [dockerConfigFile]
…
> plus some shuffling in the `func findAuthentication(registry, path string, legacyFormat bool) (string, string, error)` seemed like a reasonable place to start
Yes, plus `SetAuthentication`/`RemoveAuthentication`/`RemoveAllAuthentication`; after one works, the rest should be straightforward. `credsStore` and `credHelpers` AFAICS use the same protocol.

[...]

PR #656, opened 2019-07-16
Credentials Store

sj26:
This implementes global credentials stores per #111 by respecting `credsStore` when set in the docker config file. In the docker cli, this always overrides the file-based fetch and save whenever set (docker always decides on which store is appropriate for a registry and [returns only that store](https://github.com/docker/cli/blob/master/cli/command/registry/login.go#L143) - the helper-based store), so the same behaviour is implemented — it short circuits.

I didn't add tests yet because the existing helper config isn't tested, and it sounded like there was some movement. But the code works great for me when [integrated with skopeo](https://github.com/sj26/skopeo/tree/creds-store) and configured with osxkeychain a la Docker for Mac, which was my motivation.

### DIFF
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -24,6 +24,7 @@ type dockerAuthConfig struct {
 type dockerConfigFile struct {
 	AuthConfigs map[string]dockerAuthConfig `json:"auths"`
 	CredHelpers map[string]string           `json:"credHelpers,omitempty"`
+	CredsStore  string                      `json:"credsStore,omitempty"`
 }
 
 var (
@@ -45,6 +46,7 @@ var (
 func SetAuthentication(sys *types.SystemContext, registry, username, password string) error {
 	return modifyJSON(sys, func(auths *dockerConfigFile) (bool, error) {
 		if ch, exists := auths.CredHelpers[registry]; exists {
+			logrus.Debugf("storing credentials in docker credential helper (%s)", ch)
 			return false, setAuthToCredHelper(ch, registry, username, password)
 		}
 
@@ -60,6 +62,12 @@ func SetAuthentication(sys *types.SystemContext, registry, username, password st
 			}
 			logrus.Debugf("failed to authenticate with the kernel keyring, falling back to authfiles. %v", err)
 		}
+
+		if auths.CredsStore != "" {
+			logrus.Debugf("storing credentials in docker credential store (%s)", auths.CredsStore)
+			return false, setAuthToCredHelper(auths.CredsStore, registry, username, password)
+		}
+
 		creds := base64.StdEncoding.EncodeToString([]byte(username + ":" + password))
 		newCreds := dockerAuthConfig{Auth: creds}
 		auths.AuthConfigs[registry] = newCreds
@@ -118,6 +126,7 @@ func RemoveAuthentication(sys *types.SystemContext, registry string) error {
 	return modifyJSON(sys, func(auths *dockerConfigFile) (bool, error) {
 		// First try cred helpers.
 		if ch, exists := auths.CredHelpers[registry]; exists {
+			logrus.Debugf("deleting credentials from docker credential helper (%s)", ch)
 			return false, deleteAuthFromCredHelper(ch, registry)
 		}
 
@@ -129,6 +138,12 @@ func RemoveAuthentication(sys *types.SystemContext, registry string) error {
 				return false, nil
 			}
 			logrus.Debugf("failed to delete credentials from the kernel keyring, falling back to authfiles")
+		}
+
+		// Then try global credential store, if set.
+		if auths.CredsStore != "" {
+			logrus.Debugf("deleting credentials from docker credential store (%s)", auths.CredsStore)
+			return false, deleteAuthFromCredHelper(auths.CredsStore, registry)
 		}
 
 		if _, ok := auths.AuthConfigs[registry]; ok {
@@ -284,6 +299,11 @@ func findAuthentication(registry, path string, legacyFormat bool) (string, strin
 	// First try cred helpers. They should always be normalized.
 	if ch, exists := auths.CredHelpers[registry]; exists {
 		return getAuthFromCredHelper(ch, registry)
+	}
+
+	// Use the global credential helper, if set
+	if auths.CredsStore != "" {
+		return getAuthFromCredHelper(auths.CredsStore, registry)
 	}
 
 	// I'm feeling lucky


TomSweeneyRedHat:
@sj26 Thanks for the PR, you'll need to sign it to make Travis CI happy, `git commit --amend -s` and then repush.

pkg/docker/config/config.go
```diff
@@ -24,6 +24,7 @@ type dockerAuthConfig struct {
 type dockerConfigFile struct {
 	AuthConfigs map[string]dockerAuthConfig `json:"auths"`
 	CredHelpers map[string]string           `json:"credHelpers,omitempty"`
+	CredsStore  string                      `json:"credsStore,omitempty"`
```
TomSweeneyRedHat: This looks to work for `docker login` but not for `podman login` or `buildah login` as the creds for those are stored in `$XDG_RUNTIME_DIR/containers/auth.json`?  Any chance to add checks for that file too?

sj26:
Ah, I was wondering what that "DCO" error was. I sign every commit, but signing off is different? In any case, done!

pkg/docker/config/config.go
sj26: Sorry, that goes a little deeper than my familiarity — I don't use podman or buildah. But if the config files are separate won't they respect each files' contents — including `credsStore`?

sj26: Oh, do you mean that `auths` can come from that other file when using those utilities? That sounds like another pull request to me, unrelated to this immediate change?

TomSweeneyRedHat:
@sj26 Thanks for the sign and repush.  That looked to fix the DCO issue, however, it looks like you need to rebase now as I think someone snuck a conflicting PR in on your changes.

pkg/docker/config/config.go
TomSweeneyRedHat: Docker only knows about its auth file that lives in `~/.docker/config.json`.  Buildah and Podman first check `$XDG_RUNTIME_DIR/containers/auth.json` and if not found there, they then check the  `~/.docker/config.json` file.  So the contents are the same, the location differs.

I'd be fine with letting this in without changes for Podman/Buildah assuming a test or two is added.  However, if this goes through without the changes for Podman/Buildah I'd like to keep the issue open and have a note added about the other work still outstanding.

sj26:
I see there has been an [independent effort to add keyring](https://github.com/containers/image/commit/94e0f0e2e754ba45be6da695c02e3b0df036c37c), regardless of docker credential stores. Does skopeo/buildah/podman intend to diverge from docker? Can this change also exist? Which takes priority — docker credential store or keyring?

sj26:
I can see that credential helpers are still preferred above keyring, so I've rebased this and made sure the global credential store is preferred above keyring as well.

rhatdan:
@mtrmac @vrothberg PTAL

rhatdan:
@sj26 Needs a rebase.  We now look to see if authfile is set and use it, if not set then we use kernel keyring.

pkg/docker/config/config.go
mtrmac: The trouble is that looking at `modifyJSON`, edits only use `…/containers/auth.json`, and `~/.docker/config.json` is never consulted.

So, if the user sets up a `credsStore` in `~/.docker/config.json`, accessing credentials recorded in there will work fine util the first `podman login` for a registry — that will perform edits on `auth.json`, not see any `credsStore` configuration in there and just write into that file.

This is particularly painful because the default `…/containers/auth.json` locations are all temporary (per-user , cleared on reboot or even earlier), so it is not all that practical to configure `credsStore` in them.


OTOH, the current `credHelpers` code has exactly the same issue, so I’m fine with merging this PR without having to solve this here.

sj26: Personally I don't use podman or buildah, I'm just using skopeo to supplement docker usage, so I don't have a `…/containers/auth.json`.

pkg/docker/config/config.go
```diff
@@ -275,6 +289,11 @@ func findAuthentication(registry, path string, legacyFormat bool) (string, strin
 		return getAuthFromCredHelper(ch, registry)
 	}
 
+	// Use the global credential helper, if set
+	if auths.CredsStore != "" {
+		return getAuthFromCredHelper(auths.CredsStore, registry)
+	}
```
mtrmac: FIXME (I need to update/research): Is this right, or should we use existing entries in the config file first?

sj26:
Rebased again.

I don't know how to test this. The code hits the filesystem and looks for binaries pretty directly. Creating mock binaries with expectations on input and crafted outputs is going to be pretty heavy-weight, and there is no precedent for the existing credential helper or keyring support.

I've used [bintest](https://github.com/buildkite/bintest) for testing with mock binaries. Would something like that be okay?

pkg/docker/config/config.go
sj26: > In the docker cli, this always overrides the file-based fetch and save whenever set (docker always decides on which store is appropriate for a registry and [returns only that store](https://github.com/docker/cli/blob/master/cli/command/registry/login.go#L143) - the helper-based store), so the same behaviour is implemented — it short circuits.

mtrmac:
WRT testing this, it would be possible to
1. wrapp the `*FromCredHelper` functions in a `credHelperCaller` interface
2. move all of top-level code from the `*GetAuthentication` functions into private helper functions that take an extra `credHelperCaller` parameter
3. from the public API, call these private helper functions with the current helper call implementation; from the tests, call them with some other implementation that does not actually execute processes.

That should be much lighter-weight and more appropriate for unit tests.

Ultimately, for real-world testing with something close to real-world credential helpers, this needs to be an integration test in a container, or something like that. That would be ideal as integration tests in skopeo (which are run on every PR in this repo as well), OTOH skopeo only invokes the `GetAuthentication` functionally. The rest (add/remove/remove all) would be better tested in podman ( https://github.com/containers/libpod ), which has `podman {login, logout}` etc. — OTOH those tests are not run on every PR in this repo.

Overall, there’s no perfect place.

---

> I've used [bintest](https://github.com/buildkite/bintest) for testing with mock binaries. Would something like that be okay?

I can’t quite see how that helps — AFAICT it can create the mocks either in an user-specified path or in a temporary directory; in itself it does nothing to ensure that `exec.LookPath` and `exec.Command` use that mock instead of the global command located in `PATH`, if any. If we had that redirection mechanism (which is possible with `PATH` manipulation I suppose), we wouldn’t need the mocking part of that package — we could either use shell scripts or build some other binaries that directly return expected data.

The only thing that buys us AFAICS is that the mocks could verify that the “expected” calls to the helpers happen, but that’s going to hard-code expectations… at this scale I guess it’s OK, I’ve seen horrors when mocks are overused.

But all of this seems pretty heavy-weight for unit tests, and difficult to make portable and safe to run by unsuspecting users who just do `go get`. I’d be less worried about doing this in the skopeo/podman integration tests, which are guaranteed to run 1) on Linux 2) in a container.

---

Ultimately, the current calls to helpers don’t have tests either, so not having tests for the new functionality is not a blocker for this PR. (Of course, contributing tests is the best way to maximize the likelihood that the functionality will keep working.)

sj26:
Sorry, I'm not going to have capacity to pick this up again. I'll close, but anybody is welcome to grab this work and run with it.

PR #657, opened 2019-07-16
progress bar: use spinners for -1 blob sizes

vrothberg:
Schema1 manifests don't include blob sizes causing the BlobInfo sizes to
be -1 and ultimately preventing the progress bars from displaying
meaningful data.  Use a spinner for such cases.  The spinner renders a
sequence of dots (".", "..", "...", "....", "").

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -597,15 +597,32 @@ func (c *copier) createProgressBar(pool *mpb.Progress, info types.BlobInfo, kind
 		prefix = prefix[:maxPrefixLen]
 	}
 
-	bar := pool.AddBar(info.Size,
-		mpb.BarClearOnComplete(),
-		mpb.PrependDecorators(
-			decor.Name(prefix),
-		),
-		mpb.AppendDecorators(
-			decor.OnComplete(decor.CountersKibiByte("%.1f / %.1f"), " "+onComplete),
-		),
-	)
+	// Use a normal progress bar when we know the size (i.e., size > 0).
+	// Otherwise, use a spinner to indicate that something's happening.
+	var bar *mpb.Bar
+	if info.Size > 0 {
+		bar = pool.AddBar(info.Size,
+			mpb.BarClearOnComplete(),
+			mpb.PrependDecorators(
+				decor.Name(prefix),
+			),
+			mpb.AppendDecorators(
+				decor.OnComplete(decor.CountersKibiByte("%.1f / %.1f"), " "+onComplete),
+			),
+		)
+	} else {
+		bar = pool.AddSpinner(info.Size,
+			mpb.SpinnerOnLeft,
+			mpb.BarClearOnComplete(),
+			mpb.SpinnerStyle([]string{".", "..", "...", "....", ""}),
+			mpb.PrependDecorators(
+				decor.Name(prefix),
+			),
+			mpb.AppendDecorators(
+				decor.OnComplete(decor.Name(""), " "+onComplete),
+			),
+		)
+	}
 	if c.progressOutput == ioutil.Discard {
 		c.Printf("Copying %s %s\n", kind, info.Digest)
 	}
--- a/vendor.conf
+++ b/vendor.conf
@@ -46,6 +46,6 @@ github.com/etcd-io/bbolt v1.3.2
 github.com/klauspost/pgzip v1.2.1
 github.com/klauspost/compress v1.4.1
 github.com/klauspost/cpuid v1.2.0
-github.com/vbauerster/mpb v3.3.4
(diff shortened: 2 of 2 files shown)


vrothberg:
@rhatdan @mtrmac @edsantiago PTAL

You can easily test it with `go mod edit -replace github.com/containers/image=github.com/vrothberg/image@use-spinners` and revendoring in Skopeo or Podman.

copy/copy.go
```diff
@@ -597,15 +597,32 @@ func (c *copier) createProgressBar(pool *mpb.Progress, info types.BlobInfo, kind
...
-			decor.Name(prefix),
-		),
-		mpb.AppendDecorators(
-			decor.OnComplete(decor.CountersKibiByte("%.1f / %.1f"), " "+onComplete),
-		),
-	)
+	// If we don't know the size of the blob (size == -1), just display a
+	// spinner. Otherwise, display a proper progress bar.
+	var bar *mpb.Bar
+	if info.Size == -1 {
```
edsantiago: Is there a guarantee that the only invalid size will be `-1`, and never zero or -2 or anything else?

And, minor point, I tend to like coding for the usual-case first (vs rare-case). What would you think of:
```go
    if info.Size > 0 {
        bar = pool.AddBar(...)
    } else {
        // Unknown or invalid size
        bar = pool.AddSpinner()
    }
```

TomSweeneyRedHat: +1 to the `<0' vs `==-1' thought form @edsantiago

vendor.conf
```diff
@@ -46,6 +46,6 @@ github.com/etcd-io/bbolt v1.3.2
 github.com/klauspost/pgzip v1.2.1
 github.com/klauspost/compress v1.4.1
 github.com/klauspost/cpuid v1.2.0
-github.com/vbauerster/mpb v3.3.4
+github.com/vbauerster/mpb v3.4.0
```
TomSweeneyRedHat: Is this change intentional?

copy/copy.go
mtrmac: The definition of `BlobInfo.Size` explicitly says `-1`.

(The flip side for having the usual case first is that one then has to explicitly track (or, as in the above suggestion, comment) what the unusual case is. In this case I don’t care either way at all, neither about the ordering nor about the condition)

copy/copy.go
```diff
@@ -597,15 +597,32 @@ func (c *copier) createProgressBar(pool *mpb.Progress, info types.BlobInfo, kind
...
+	} else {
+		bar = pool.AddBar(info.Size,
+			mpb.BarClearOnComplete(),
+			mpb.PrependDecorators(
+				decor.Name(prefix),
+			),
+			mpb.AppendDecorators(
+				decor.OnComplete(decor.CountersKibiByte("%.1f / %.1f"), " "+onComplete),
+			),
+		)
```
mtrmac: It seems attractive to build the shared `[]mpb.BarOption` options outside of the `if`/`else` — but it might well be so much uglier than the current code that it would not be worth it. Up to you.

vendor.conf
mtrmac: Yes, that version adds the `AddSpinner` call.

copy/copy.go
vrothberg: 2 pro and 1 neutral, so I'll change it as proposed.

copy/copy.go
vrothberg: Let's keep it as is but I see wisdom in rethinking this code when more conditions are being added.

vrothberg:
Here's the PR for Skopeo: https://github.com/containers/skopeo/pull/689

vrothberg:
Note that this fix was motivated by https://bugzilla.redhat.com/show_bug.cgi?id=1695236.

vrothberg:
Skopeo PR is green. Force merging here, to revendor into Skopeo.

copy/copy.go
```diff
@@ -597,15 +597,32 @@ func (c *copier) createProgressBar(pool *mpb.Progress, info types.BlobInfo, kind
...
-			decor.Name(prefix),
-		),
-		mpb.AppendDecorators(
-			decor.OnComplete(decor.CountersKibiByte("%.1f / %.1f"), " "+onComplete),
-		),
-	)
+	// Use a normal progress bar when we know the size (i.e., size > 0).
+	// Otherwise, use a spinner to indicate that something's happening.
+	var bar *mpb.Bar
+	if info.Size > 0 {
```
mtrmac: Pedantic nitpick, just to get it off my chest: `≥0`. Not really worth fixing.

PR #658, opened 2019-07-17
docker client: error if registry is blocked

vrothberg:
Return an error whenever we attempt to create a docker client when the
underlying registry is configured to be blocked in the registries.conf.
Until now, users of the c/image library implemented this behaviour but
it has not been enforced.

Now, it's time to pull the trigger and enforce the configuration.

---

Entirely untested. Just want to throw it at Skopeo and continue working on it later.

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -254,6 +254,9 @@ func newDockerClient(sys *types.SystemContext, registry, reference string) (*doc
 		return nil, errors.Wrapf(err, "error loading registries")
 	}
 	if reg != nil {
+		if reg.Blocked {
+			return nil, fmt.Errorf("registry %s is blocked in %s", reg.Prefix, sysregistriesv2.ConfigPath(sys))
+		}
 		skipVerify = reg.Insecure
 	}
 	tlsClientConfig.InsecureSkipVerify = skipVerify
--- a/pkg/sysregistriesv2/system_registries_v2.go
+++ b/pkg/sysregistriesv2/system_registries_v2.go
@@ -303,9 +303,8 @@ func (config *V2RegistriesConf) postProcess() error {
 	return nil
 }
 
-// getConfigPath returns the system-registries config path if specified.
-// Otherwise, systemRegistriesConfPath is returned.
-func getConfigPath(ctx *types.SystemContext) string {
+// ConfigPath returns the path to the system-wide registry configuration file.
+func ConfigPath(ctx *types.SystemContext) string {
 	confPath := systemRegistriesConfPath
 	if ctx != nil {
 		if ctx.SystemRegistriesConfPath != "" {
@@ -336,7 +335,7 @@ func InvalidateCache() {
 
 // getConfig returns the config object corresponding to ctx, loading it if it is not yet cached.
 func getConfig(ctx *types.SystemContext) (*V2RegistriesConf, error) {
-	configPath := getConfigPath(ctx)
+	configPath := ConfigPath(ctx)
 
 	configMutex.Lock()
 	defer configMutex.Unlock()
--- a/pkg/sysregistriesv2/system_registries_v2_test.go
+++ b/pkg/sysregistriesv2/system_registries_v2_test.go
@@ -115,7 +115,7 @@ func TestRefMatchesPrefix(t *testing.T) {
 	}
 }
 
-func TestGetConfigPath(t *testing.T) {
+func TestConfigPath(t *testing.T) {
 	const nondefaultPath = "/this/is/not/the/default/registries.conf"
 	const variableReference = "$HOME"
 	const rootPrefix = "/root/prefix"
@@ -146,7 +146,7 @@ func TestGetConfigPath(t *testing.T) {
 		// No environment expansion happens in the overridden paths
 		{&types.SystemContext{SystemRegistriesConfPath: variableReference}, variableReference},
 	} {
-		path := getConfigPath(c.sys)
+		path := ConfigPath(c.sys)
 		assert.Equal(t, c.expected, path)
 	}
 }


docker/docker_client.go
```diff
@@ -254,6 +254,9 @@ func newDockerClient(sys *types.SystemContext, registry, reference string) (*doc
 		return nil, errors.Wrapf(err, "error loading registries")
 	}
 	if reg != nil {
+		if reg.Blocked {
+			return nil, fmt.Errorf("registry %s is blocked in %s", reg.Prefix, sysregistriesv2.GetConfigPath(sys))
+		}
```
mtrmac: Yes, moving this from the various higher-level callers (except for the search path impact, I suppose) is great.

vrothberg: That's a good point.

I have no strong opinion on whether search should be allowed. It's currently allowed by Podman, which counts for allowing to search a registry even when being configured as blocked. On the other hand, cutting it off seems cleaner/stricter. Maybe I am over thinking it but if we allow search, would login make sense as well?

@mtrmac @rhatdan @mheon WDYT?

vrothberg: I think it breaks down to allowing to opt-out of blocking via the `types.SystemContext`.

rhatdan: I think we should block the search if it is blocked in the config. Podman should be fixed.

mtrmac: > I have no strong opinion on whether search should be allowed. It's currently allowed by Podman, which counts for allowing to search a registry even when being configured as blocked.

Both Buildah and CRI-O ignore search path elements that are blocked (though with different semantics, Buildah looks up the registry host, CRI-O the result of combining the registry host with the unqualified name).

Ultimately, with this PR, the access is going to fail anyway, so it mostly does not matter whether upper layers do or don’t process the block in the search implementation (except for one rare case, where a search path registry is blocked but the mirror is not: in that case this PR lets the mirror access through, but applying the block at search path level would not allow the mirror to be attempted).

> On the other hand, cutting it off seems cleaner/stricter. Maybe I am over thinking it but if we allow search, would login make sense as well?

I was in no way suggesting that search results should be _exempt_ from blocking (I see no reason for that, let’s keep the features simple and orthogonal if we can); I was just pointing out that there is _some_ code that deals with blocking in some callers of c/image that we might not want to just delete after this PR.

---

As for login, that’s a bit awkward because login is registry-wide but blocking is namespace-wide now. The current behavior (preventing login when the whole registry is blocked, allowing it if only some namespaces are blocked) seems right to me.

---

A kind of opt-out of blocking can be done by setting `SystemRegistriesConfPath` to `/dev/null`. Do we actually need the more detailed granularity? This is supposed to be system-wide policy, it is a bit suspect to provide an easy “let me do it anyway” bit. Of course we can’t _stop_ users from accessing the content (as long as they can compile / execute their own binaries), but I’d wait for an explicit request with a well-reasoned use case before adding that to the stable API.

vrothberg: @mtrmac, I interpreted "except for the search path impact, I suppose" as an opt-out. As you said, users could already set it to `/dev/null` and I wouldn't like to add more than that.

> As for login, that’s a bit awkward because login is registry-wide but blocking is namespace-wide now. The current behavior (preventing login when the whole registry is blocked, allowing it if only some namespaces are blocked) seems right to me.

I concur.

> in that case this PR lets the mirror access through, but applying the block at search path level would not allow the mirror to be attempted

I will fix that as I think it would be a bug.

mtrmac: Fix which part? OpenShift apparently has some opinions about the intersection of blocking and mirrors… IIRC the idea was to configure mirrors for the primary location _and_ block the primary location, but I’d need to find the earlier conversations or ask again to be sure.

(OTOH the search path aspect is not that relevant for OpenShift.)

vrothberg: @mtrmac, to also block here: https://github.com/containers/image/blob/master/docker/docker_image_src.go#L35

I need to test this PR and also need to add new tests for this case to make sure we're not regressing now or later.

> IIRC the idea was to configure mirrors for the primary location and block the primary location

This can already be done with the prefix/location or am I missing something?

> [...] but I’d need to find the earlier conversations or ask again to be sure.

Awesome, thanks!

mtrmac: > @mtrmac, to also block here: https://github.com/containers/image/blob/master/docker/docker_image_src.go#L35

Yes, that’s not known to be desirable right now. [Also, as a matter of elegance, having one location to enforce a “blocking” policy is clear; having two risks having three and … Of course, requirements are requirements and elegance is secondary.]

> > IIRC the idea was to configure mirrors for the primary location and block the primary location
> 
> This can already be done with the prefix/location or am I missing something?

That allows _redirecting_ the primary but not disabling its use.

One could convert (Prefix: A, Mirrors: BC) into (Prefix: A, Location: C, Mirrors: B), but that’s pretty nonintuitive. OTOH it might be good enough for OpenShift, where the configuration is going to be machine-generated.

vrothberg: > Also, as a matter of elegance, having one location to enforce a “blocking” policy is clear

Yes, that's convincing.

I added integration tests to Skopeo: https://github.com/containers/skopeo/pull/690

vrothberg:
@mtrmac rebased

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -303,9 +303,9 @@ func (config *V2RegistriesConf) postProcess() error {
 	return nil
 }
 
-// getConfigPath returns the system-registries config path if specified.
-// Otherwise, systemRegistriesConfPath is returned.
-func getConfigPath(ctx *types.SystemContext) string {
+// GetConfigPath returns the system-registries config path if specified.
+// Otherwise, the default registry path is returned.
+func GetConfigPath(ctx *types.SystemContext) string {
```
mtrmac: Nits, before this becomes a stable API:
- The `Get` prefix is not idiomatic for getters in Go.
- “if specified / default” are from the implementation point of view, the caller does not quite care; the caller just wants the path that is used for that `ctx` value.

Maybe use both the name and the doc string from the v1 package. (I don’t have any preference WRT including the “registries.conf” file name in the function name.)

vrothberg: Good point. I chose `ConfigPath` as an identifier to avoid redundancy wrt. "registries" at call sites (i.e., `sysregistriesv2.RegistriesConfPath` vs `sysregistriesv2.ConfigPath`).

mtrmac:
LGTM, thanks! Please rebase, and feel free to merge.

vrothberg:
@mtrmac, @rhatdan , to make sure we're on the same page: I think this change qualifies for a major version bump for the next release. In case you disagree, please share your thoughts.

mtrmac:
A major version bump because it is an API break, or because it is such a big feature that it deserves to be announced?

Either way, increasing the major number does not hurt much, so I can live with that (and we still “owe” a 2.1.0 bump from the `Env` addition, so I’m perfectly fine with a minor bump).  Still…

I can see an argument that this may break callers, and thus requires a major version bump, OTOH it could be also considered a bug in that the blocked flag was documented but ignored (in the context that c/image transports are generally expected to do the right thing in the callers’ environment without requiring an explicit opt-in for every single feature and config file). Yes, the two views are pretty much opposite extremes, yet I’m not at all sure which one is correct — intuitively I’m fairly strongly leaning towards “bug fix”, but I have no valid counterargument to the “breaks callers” reason.

WRT announcing the feature by a major version bump, I feel a bit more decided: Given that we have adopted semantic versioning, which is based on the API compatibility behavior, diluting that with not following the rules because we want to use the version number for an entirely different purpose is undermining the promise of the semantic versioning semantics.

So, I’m not too enthusiastic about bumping the major number here — but I won’t stand in the way.

vrothberg:
> A major version bump because it is an API break, [... ]?

@mtrmac, because of the API break.

> [...] or because it is such a big feature that it deserves to be announced?

No no, I would not exploit it in such a way. I expect users to check the changelog and users should expect such changes to be mentioned there.

mtrmac:
OK, I really don’t have a good reason to say this is not an API break.

ACK.

(As long as we are doing this, would it make sense to move at least `ImageDestination` into `types/unstable` or something like that, to try to except it from the API guarantee, because it is changing so often in hard-to-predict ways? Maybe `BlobInfoCache` as well. (Probably not `ImageSource` — it changes from time to time as well, but not as often, and it’s known to have external callers, unlike `ImageDestination` to my knowledge.)

mtrmac:
> (As long as we are doing this, would it make sense to move at least `ImageDestination` into `types/unstable` or something like that

Hum, no, not exactly this, that would create a circular `types.ImageReference` vs. `types/unstable.ImageDestination` dependency. Just renaming `ImageDestination` to `UnstableImageDestination` would be pretty ugly (and could be misunderstood to refer to instability of the image). I guess this will need more thought and probably should not block the next release.

vrothberg:
> I guess this will need more thought and probably should not block the next release.

I concur :+1:

PR #661, opened 2019-07-21
Add ability to install man pages

rhatdan:
We want to create a containers-image package so that container-common can
include all of these man pages in the distribution.

Replaces: https://github.com/containers/skopeo/pull/645

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,12 @@ BUILDFLAGS := -tags "$(BUILDTAGS)"
 PACKAGES := $(shell go list $(BUILDFLAGS) ./... | grep -v github.com/containers/image/vendor)
 SOURCE_DIRS = $(shell echo $(PACKAGES) | awk 'BEGIN{FS="/"; RS=" "}{print $$4}' | uniq)
 
+PREFIX ?= ${DESTDIR}/usr
+MANINSTALLDIR=${PREFIX}/share/man
+GOMD2MAN ?= $(shell command -v go-md2man || echo '$(GOBIN)/go-md2man')
+MANPAGES_MD = $(wildcard docs/*.5.md)
+MANPAGES ?= $(MANPAGES_MD:%.md=%)
+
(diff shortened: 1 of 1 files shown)


rhatdan:
@lsm5 @mtrmac @vrothberg PTAL

Makefile
```diff
@@ -31,6 +37,17 @@ build: vendor build-internal
 build-internal:
 	$(GPGME_ENV) go build $(BUILDFLAGS) $(PACKAGES)
 
+$(MANPAGES): %: %.md
+	@sed -e 's/\((.*\.5.md)\)//' -e 's/\[\(skopeo.*\)\]/\1/' $<  | $(GOMD2MAN) -in /dev/stdin -out $@
```
mtrmac: I can’t see that the two sed commands match anything in existing man pages. The skopeo reference is especially suspect.

Makefile
```diff
@@ -56,7 +73,7 @@ vendor: tools.timestamp vendor.conf
 	@touch vendor
 
 clean:
-	rm -rf vendor tools.timestamp
+	rm -rf vendor tools.timestamp docs/*.5
```
mtrmac: `$(MANPAGES)` please, to minimize duplication.

lsm5:
> ACK overall, but this needs to also be ACKed by @lsm5 , because new RPMs will be necessary. (Or is that Jindřich now?)

Jindřich for RHEL. Me for everything else :) . I'll check this out and get back in a bit.

Makefile
rhatdan: Yup cut & paste from skopeo Makefile (fixed)

Makefile
rhatdan: Fixed

rhatdan:
I am going to merge this and @lsm5 and @jnovy can work on packaging up container-storage and containers-image separately.

PR #663, opened 2019-07-25
Remove pkg/sysregistries

mtrmac:
Everyone should use `pkg/sysregistriesv2` instead.

Migration:
- `GetRegistries` -> `sysregistriesv2.UnqualifiedSearchRegistries`
- `GetInsecureRegistries` -> `sysregistriesv2.GetRegistries` + check `Insecure` flag
  ... or, usually, `sysregistriesv2.FindRegistry` + check `Insecure` flag
- `RegistriesConfPath` -> `sysregistriesv2.ConfigPath`


TomSweeneyRedHat:
The changes look OK, I just worry a bit about other projects that might expect the code to be there still?  Any worries on your end about that?

mtrmac:
@TomSweeneyRedHat See the earlier conversation in #633.

Practically, this is going to break the whole c/image stack with the removal of `sysregistries.RegistriesConfPath`, but the fix is trivial and it makes the code, pedantically speaking, more correct (using the same subpackage for both reading the configuration and reporting the config file used to read the configuration). That part is no-op in practice and an improvement in theory.

If there are any remaining “real” uses of the v1 package, migration to v2 may require some changes (to adapt to the idea that config is per-prefix, not per-host), it may simplify the code (because `FindRegistry` removes the need to do lookups manually, and because c/image uses most of the configuration automatically), and it is also probably more urgently necessary (e.g. new OpenShift versions generate v2 configuration unreadable with the v1 package).

rhatdan:
LGTM

Issue #666, opened 2019-07-29
containers-transports man page ambiguous


thmo:
In the containers-transports man page, the syntax for the containers-storage is noted as:
```
containers-storage: [storage-specifier]{image-id|docker-reference[@image-id]}
```
However, there's two problems:
- an extra space after the first colon
- the `[` `]` around `storage-specifier` are *literals*, i.e. have to be given, and are not denoting an optional part of the syntax.

TomSweeneyRedHat:
I'd assign this to myself, but it looks like I don't have the juice to do so.

rhatdan:
I do...
Thanks @thmo for pointing this out.

PR #667, opened 2019-07-29
Touch up transport man page

TomSweeneyRedHat:
Touch up the containers-transports storage specificiation in the man page.

Fixes: #666

### DIFF
--- a/docs/containers-transports.5.md
+++ b/docs/containers-transports.5.md
@@ -11,7 +11,7 @@ containers-transports - description of supported transports for copying and stor
 Tools which use the containers/image library, including skopeo(1), buildah(1), podman(1), all share a common syntax for referring to container images in various locations.
 The general form of the syntax is _transport:details_, where details are dependent on the specified transport, which are documented below.
 
-### **containers-storage:** [storage-specifier]{image-id|docker-reference[@image-id]}
+### **containers-storage**:[**[**storage-specifier**]**]{image-id|docker-reference[@image-id]}
 
 An image located in a local containers storage.
 The format of _docker-reference_ is described in detail in the **docker** transport.


thmo:
That doesn't seem right, now it looks like this:
```
containers-storage{:storage-specifier} {image-id|docker-reference[@image-id]}
```
- the colon should be directly after `containers-storage`
- now there's a new space between the `}` and the `{` that shouldn't be there
- the original problem of the ticket is not finally solved, because in the text below, it still says
   ```
   The storage-specifier allows for referencing storage locations on the file system
   and has the format [[driver@]root[+run-root][:options]]
   ```
   where you cannot tell the difference between the outer `[` `]` (literal) and in the inner (option).

TomSweeneyRedHat:
@thmo then I misunderstood what you were saying, and I'm still not clear what the final change should look like.  If you add it as a reply to this issue, I'll happily change it on the page.

docs/containers-transports.5.md
```diff
@@ -11,7 +11,7 @@ containers-transports - description of supported transports for copying and stor
 Tools which use the containers/image library, including skopeo(1), buildah(1), podman(1), all share a common syntax for referring to container images in various locations.
 The general form of the syntax is _transport:details_, where details are dependent on the specified transport, which are documented below.
 
-### **containers-storage:** [storage-specifier]{image-id|docker-reference[@image-id]}
+### **containers-storage**{:storage-specifier} {image-id|docker-reference[@image-id]}
```
mtrmac: I think something like
```suggestion
### **containers-storage:**[[**storage-specifier**]**]{image-id|docker-reference[@image-id]}
```
would be accurate (I haven’t checked whether it renders correctly, neither in Markdown nor in the man page).

TomSweeneyRedHat: This shows both brackets and everything highlighted.  I've pushed that out, please let me know if further adjustments are necessary.

rhatdan:
LGTM

docs/containers-transports.5.md
mtrmac: Yeah, sadly the `###` format causes a `.SS` directive, which is bold by default. We could make the replacements italic, but a regular non-bold seems might not be possible to do with Markdown.

Oh well, I suppose we at least got rid of the space.

Issue #662, opened 2019-07-22
keyring ignores --authfile


vrothberg:
Libpod CI caught a regression from the recent keyring changes (see https://github.com/containers/libpod/pull/3598/checks?check_run_id=174300929). We shouldn't use the keyring when the user explicitly requests to use an auth file.

@QiWang19 PTAL

QiWang19:
@vrothberg looking :eye: :eye:

vrothberg:
For the record. It looks like the libpod CI was just flaky and is green now. Nonetheless, we should adhere to what the user has specified in the system context.

mtrmac:
> We shouldn't use the keyring when the user explicitly requests to use an auth file.

… why? The --authfile option has historically been documented to configure the _path_ not to _turn on/off usage_.

(I’m not quite saying that using --authfile presence as an on/off switch is unreasonable, just that it is _not obvious_.)

vrothberg:
If a user relies on the credentials being stored there (and maybe move the file around), it would now break.

However, you're right that it's not obvious and a quick man pages skim did not reveal a statement that the `--auth-file` will eventually be used or not.

mtrmac:
Arguably the user could have relied on the credentials being stored in `~/.docker/config.json` just as well without explicitly specifying `--authfile`; we don’t know for sure.

OTOH forcing users to possibly specify two options on every command is also pretty unattractive… which is why I was arguing for an explicit `credHelper`/`credsStore` representation in the existing JSON. (I still need to catch up with all the conversation in #619 I’m afraid.)

vrothberg:
Alright, just drop a note here.

@QiWang19, let's wait until we have a consensus how to proceed.

mtrmac:
I feel reasonably confident that _if_ we want to allow users to explicitly choose to use a config file, we probably want to also allow them to explicitly choose to use the keyring.

OTOH how that option should be designed has another matter — should it just be a `credsStore` value (when neither “config file” nor ”keyring” are `credsStore` values at all). So that does not simplify things.

(See also how https://docs.docker.com/engine/reference/commandline/login/#credentials-store requires explicit opt-in into the macOS keychain.)

rhatdan:
As a user, if I typed `--authfile $AUTHFILEPATH`  I would expect it to be honored, and we should document this behaviour.

vrothberg:
> As a user, if I typed `--authfile $AUTHFILEPATH` I would expect it to be honored, and we should document this behaviour.

[...]

PR #668, opened 2019-07-30
use keyring when authfile empty

QiWang19:
Fixes: #662
Store user credentials to kernel keyring if AuthFilePath is "", otherwise set the credentials to the authfile path.

Related to my libpod PR https://github.com/containers/libpod/pull/3674

### DIFF
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -46,13 +46,18 @@ func SetAuthentication(sys *types.SystemContext, registry, username, password st
 			return false, setAuthToCredHelper(ch, registry, username, password)
 		}
 
-		err := setAuthToKernelKeyring(registry, username, password)
-		if err == nil {
-			logrus.Debugf("credentials for (%s, %s) were stored in the kernel keyring\n", registry, username)
-			return false, nil
+		// Set the credentials to kernel keyring if sys.AuthFile is not specified.
+		// The keyring might not work in all environments (e.g., missing capability) and isn't supported on all platforms.
+		// Hence, we want to fall-back to using the authfile in case the keyring failed.
+		// However, if the sys.AuthFilePath is set, we want adhere to the user specification and not use the keyring.
+		if sys.AuthFilePath == "" {
+			err := setAuthToKernelKeyring(registry, username, password)
+			if err == nil {
+				logrus.Debugf("credentials for (%s, %s) were stored in the kernel keyring\n", registry, username)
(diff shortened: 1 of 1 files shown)


pkg/docker/config/config.go
```diff
@@ -46,13 +46,14 @@ func SetAuthentication(sys *types.SystemContext, registry, username, password st
 			return false, setAuthToCredHelper(ch, registry, username, password)
 		}
 
-		err := setAuthToKernelKeyring(registry, username, password)
-		if err == nil {
-			logrus.Debugf("credentials for (%s, %s) were stored in the kernel keyring\n", registry, username)
-			return false, nil
+		if sys.AuthFilePath == "" {
```
vrothberg: Please add a comment here. It's a bit a of an inverted logic to achieve the fall-through and readers of the code might get stuck here.

What could help is describing the intended behaviour:
* The keyring might not work in all environments (e.g., missing capability) and isn't supported on all platforms. Hence, we want to fall-back to using the authfile in case the keyring failed.
* However, if the `sys.AuthFilePath` is set, we want adhere to the user specification and not use the keyring.

pkg/docker/config/config.go
```diff
@@ -46,13 +46,14 @@ func SetAuthentication(sys *types.SystemContext, registry, username, password st
...
+			err := setAuthToKernelKeyring(registry, username, password)
+			if err == nil {
+				logrus.Debugf("credentials for (%s, %s) were stored in the kernel keyring\n", registry, username)
+				return false, nil
+			}
+			logrus.Debugf("failed to authenticate with the kernel keyring, falling back to authfiles.")
```
vrothberg: I didn't notice it before but could we add remove the trailing dot and replace it with `: %v` and print the error. It would be helpful in debug mode.

vrothberg:
@mtrmac PTAL

rhatdan:
I like this patch LGTM once you fix @vrothberg comments.

mtrmac:
LGTM. Thanks!

rhatdan:
@vrothberg @mtrmac can we tag a new release for this?  I think this would be v2.1.0  Do you agree?  We need this to allow podman login to work with kernel keyring.

mtrmac:
After #658 it would be a v3.0.0.

rhatdan:
SGTM

vrothberg:
SGTM

PR #655, opened 2019-07-15
Use docker client TLS config for getting bearer token

linkvt:
Fixes #654 

I also removed the `TODO` as we use the parameters value now which is set on the `dockerClient`s TLS config.
I don't know what runcom means or if this was forced to true, so I would be thankful for some explanations if I missed something there.
The fix is quite trivial as it is already done like this a few lines below:
https://github.com/containers/image/blob/master/docker/docker_client.go#L560
By looking at the blame it seems like this was just forgotten as the old line was 3 years old.

I tested logging in with podman, logging out and reusing an existing login from `docker`:
```
 vincent@vincent-pc  libpod   master ●  bin/podman logout docker-reg --log-level debug
INFO[0000] running as rootless                          
DEBU[0000] Looking for TLS certificates and private keys in /etc/containers/certs.d/docker-reg 
DEBU[0000]  crt: /etc/containers/certs.d/docker-reg/ca.crt 
DEBU[0000]  cert: /etc/containers/certs.d/docker-reg/client.cert 
DEBU[0000]  key: /etc/containers/certs.d/docker-reg/client.key 
DEBU[0000] GET https://docker-reg/v2/           
DEBU[0000] Ping https://docker-reg/v2/ status 401 
DEBU[0000] GET https://docker-reg/artifactory/api/docker/null/v2/token?service=docker-reg 
Not logged into docker-reg
 vincent@vincent-pc  libpod   master ●  bin/podman login docker-reg --log-level debug
INFO[0000] running as rootless                          
Username: uid
Password: 
DEBU[0004] Looking for TLS certificates and private keys in /etc/containers/certs.d/docker-reg 
DEBU[0004]  crt: /etc/containers/certs.d/docker-reg/ca.crt 
DEBU[0004]  cert: /etc/containers/certs.d/docker-reg/client.cert 
DEBU[0004]  key: /etc/containers/certs.d/docker-reg/client.key 
DEBU[0004] GET https://docker-reg/v2/           
DEBU[0004] Ping https://docker-reg/v2/ status 401 
DEBU[0004] GET https://docker-reg/artifactory/api/docker/null/v2/token?account=uid&service=docker-reg 
DEBU[0004] GET https://docker-reg/v2/           
Login Succeeded!
 vincent@vincent-pc  libpod   master ●  bin/podman logout docker-reg --log-level debug
INFO[0000] running as rootless                          
Removed login credentials for docker-reg
 vincent@vincent-pc  libpod   master ●  docker login docker-reg                  
Username: uid
[...]

PR #669, opened 2019-08-01
Use the same HTTP client for contacting the bearer token server and the registry

### LINKED ISSUE #654: podman login not working with certificate authentication
Hi,

we have an Artifactory running behind a load balancer that uses certificate authentication.
When we try to `podman login` we get the error `Error: error authenticating creds for "domain": pinging docker registry returned: Get https://domain/v2/: remote error: tls: handshake failure`

When using `docker login` the login succeeds.
When using `podman login` it tries to reuse the existing login but also fails again:
```
$ sudo podman login domain --log-level debug
Authenticating with existing credentials...
DEBU[0000] Looking for TLS certificates and private keys in /etc/containers/certs.d/domain
DEBU[0000]  crt: /etc/containers/certs.d/domain/ca.crt
DEBU[0000]  cert: /etc/containers/certs.d/domain/client.cert
DEBU[0000]  key: /etc/containers/certs.d/domain/client.key
DEBU[0000] GET https://domain/v2/
DEBU[0000] Ping https://domain/v2/ status 401
DEBU[0000] GET https://domain/artifactory/api/docker/null/v2/token?account=userid&service=domain
Existing credentials are invalid, please enter valid username and password
Username (userid):
Password:
DEBU[0008] Looking for TLS certificates and private keys in /etc/containers/certs.d/domain
DEBU[0008]  crt: /etc/containers/certs.d/domain/ca.crt
DEBU[0008]  cert: /etc/containers/certs.d/domain/client.cert
DEBU[0008]  key: /etc/containers/certs.d/domain/client.key
DEBU[0008] GET https://domain/v2/
DEBU[0008] Ping https://domain/v2/ status 401
DEBU[0008] GET https://domain/artifactory/api/docker/null/v2/token?account=userid&service=domain
ERRO[0009] error authenticating creds for "domain": Get https://domain/artifactory/api/docker/null/v2/token?account=userid&service=domain: remote error: tls: handshake failure
```

I guess that the error is probably in this area as the certificates from the dockerClient are not used: https://github.com/containers/image/blob/5136fb6b50bde99319d73345947cf028bef9f3e9/docker/docker_client.go#L526-L530 
I will try to fix it in the next days but I'm not very familiar with the code base yet so if you see an easy fix we would appreciate that, otherwise I would also appreciate some guidance.

Thanks!

mtrmac:
Fixes #654.

Alternative to #655.

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -526,11 +526,7 @@ func (c *dockerClient) getBearerToken(ctx context.Context, challenge challenge, 
 		authReq.SetBasicAuth(c.username, c.password)
 	}
 	logrus.Debugf("%s %s", authReq.Method, authReq.URL.String())
-	tr := tlsclientconfig.NewTransport()
-	// TODO(runcom): insecure for now to contact the external token service
-	tr.TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
-	client := &http.Client{Transport: tr}
-	res, err := client.Do(authReq)
+	res, err := c.client.Do(authReq)
 	if err != nil {
 		return nil, err
 	}


mrunalp:
@vrothberg @rhatdan ptal

rhatdan:
LGTM

PR #670, opened 2019-08-02
ci: change `$GOCACHE` to a writeable path

saschagrunert:
The `$GOCACHE` seems not writeable by the user executing the tests
inside the docker container. This PR should fix this and unblock the CI
pipeline. To make skopeo build with go modules we have to rely on the
latest dot release of go 1.11.


saschagrunert:
@vrothberg seems to work again :+1:

PR #673, opened 2019-08-02
Added Container Image Encryption/Decryption support

lumjjb:
This PR contains a POC for image encryption/decryption. Based on encryption work as presented at [KubeCon CN 2019](https://www.youtube.com/watch?v=bzHPnlSfM_8). Currently, we have POCs for cri-o and skopeo that can consume this.

Available consumer examples:
- Skopeo: https://github.com/harche/skopeo/tree/pr_branch
- Cri-o: https://github.com/harche/cri-o/tree/pr_branch

Below is a summary of what is modified. We can squash the commits if preferred. 

### Types - `types/types.go`

- Added `SupportsEncryption` as part of `Image` interface (backed by `genericManifest`)
- Added `CryptoConfig` as part of `SystemContext`

### Manifest Changes - `manifest/`

- Added encryption mediatypes consts
- Modified `OCI1.UpdateLayerInfos` to also update Mediatype if provided
- Added `SupportsEncryption` into the `genericManifest` interface

### Copy - `copy/`

- In general: Added ability to encrypt/decrypt in `copy.Image`
- Added parameter to `newDigestingReader` to skip checks in the case where digest is not known (i.e. during encryption, etc.)
- Used `CryptoConfig` from `SystemContext`
- Added `EncryptConfig` and `EncryptLayers` to `copy.Options` to specify encryption when calling `Image`
- Modified `imageCopier` and `copier` constructs to propagate crypto configuration
- All of the work for actual encryption/decryption of streams are handled in `copyLayer`, with an added `toEncrypt` parameter that is passed into the function.
 - Decryption is done on source if encrypted
 - Encryption is done on dest if specified in `copy.Options`
- Note: Currently encryption supports OCI image as a source since `copy.copyAndUpdateManifest` applies changes to the src manifest, and docker schema1/2 lacks a way to define encryption via MediaTypes and Annotations.


mrunalp:
@vrothberg @mtrmac @rhatdan ptal.

vrothberg:
Hi @lumjjb, thanks for the PR! I won't find time to review this week to review but will have a look next week.

One thing I noticed while quickly skimming the code is that we add a lot of code from containerd. I wonder if we couldn't vendor in the packages from containerd directly (assuming no code changes needed)? The intention behind is to avoid redundancy and duplicate work, for instance, when fixing a regression.

mrunalp:
I think such code should probably live in a library which can be imported into the CRI implementations as needed. We are happy to host it at github.com/containers if that works for you.

lumjjb:
Yup - i am actually working on a commit now to remove that dependency. there are no hard dependencies on containerd. We were only using some `errdefs` which we didn't clean up properly. Pushing a commit now

lumjjb:
>I think such code should probably live in a library which can be imported into the CRI implementations as needed. We are happy to host it at github.com/containers if that works for you.

yep! That sounds good. We have the code living in one of our repositories now... in https://github.com/stefanberger/imgcrypt/tree/master/pkg/encryption... Yea - i think it would be perfect if we could have one living in the `containers/` path, and when we get this merged into OCI, we can decide if we should migrate it there.

mrunalp:
@lumjjb sounds good :+1: I can create the repository for you if give me the desired name for the repo.

lumjjb:
>@lumjjb sounds good 👍 I can create the repository for you if give me the desired name for the repo.

@mrunalp Awesome, `ocicrypt` should be a good name! Thanks for hosting us :D

rhatdan:
Yes my first response to this would be to get it into a vendor and then it would be easier to review.  We don't want any exposed interfaces to this code in containers/image other then the ones we support.

mrunalp:
@lumjjb I have created https://github.com/containers/ocicrypt. Who all should be added as maintainers? :)

mrunalp:
@lumjjb I have given you admin rights, so you should be able to manage the repo :)

lumjjb:
@mrunalp Awesome Thanks!!

mrunalp:
I have added Stefan as well but now you should be able to add other people as needed. Let me know if something doesn't work.

lumjjb:
Updated the PR to vendor `containers/ocicrypt` and re-did commits to help reviewing

copy/copy.go
```diff
@@ -797,7 +885,42 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
...
 	// read stream to the end, and validation does not happen.
-	digestingReader, err := newDigestingReader(srcStream, srcInfo.Digest)
+
+	var decrypted bool
+	var err error
+	if srcInfo.MediaType == manifest.DockerV2Schema2LayerGzipEncMediaType ||
+		srcInfo.MediaType == manifest.DockerV2Schema2LayerEncMediaType {
+
+		if c.decryptConfig == nil {
+			return types.BlobInfo{}, errors.New("Necessary DecryptParameters not present")
```
rhatdan: Can this be made a constant error, so callers could check?

image/docker_schema1.go
```diff
@@ -200,3 +200,8 @@ func (m *manifestSchema1) convertToManifestSchema2(uploadedLayerInfos []types.Bl
 
 	return manifestSchema2FromComponents(configDescriptor, nil, configJSON, layers), nil
 }
+
+// SupportsEncryption returns an error if encryption is not supported for the manifest type
+func (m *manifestSchema1) SupportsEncryption(context.Context) error {
+	return errors.New("Docker Schema v1 does not support encryption")
```
rhatdan: Can this be a constant?

image/docker_schema2.go
```diff
@@ -349,3 +349,8 @@ func v1ConfigFromConfigJSON(configJSON []byte, v1ID, parentV1ID string, throwawa
 	}
 	return json.Marshal(rawContents)
 }
+
+// SupportsEncryption returns an error if encryption is not supported for the manifest type
+func (m *manifestSchema2) SupportsEncryption(context.Context) error {
+	return errors.New("Docker Schema v2 does not support encryption")
```
rhatdan: Can this be a constant?

image/manifest.go
```diff
@@ -44,6 +44,8 @@ type genericManifest interface {
 	// UpdatedImage returns a types.Image modified according to options.
 	// This does not change the state of the original Image object.
 	UpdatedImage(ctx context.Context, options types.ManifestUpdateOptions) (types.Image, error)
+	// SupportsEncryption returns an error if encryption is not supported for the manifest type
+	SupportsEncryption(ctx context.Context) error
```
rhatdan: Should just return a bool rahter then an error.

lumjjb:
Addressed @rhatdan 's comments in https://github.com/containers/image/pull/673/commits/25b2de6fb088895b7ff605d553c732fccf6612f3

copy/copy.go
```diff
@@ -851,6 +979,49 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
...
+			s, fin, err := ocicrypt.EncryptLayer(c.encryptConfig, destStream, desc)
+			if err != nil {
+				return types.BlobInfo{}, errors.Wrapf(err, "Image encryption failed for the digest %+v", srcInfo.Digest)
+			}
+
+			destStream = s
+			finalizer = fin
+			inputInfo.Digest = ""
+			inputInfo.Size = -1
+			inputInfo.MediaType = encryptMediaType
```
vrothberg: Ah, so we're setting a `DockerV2Schema2` prefix to an OCI mime type. I very much prefer to use the OCI prefix for it.

copy/copy.go
```diff
@@ -668,7 +733,32 @@ type diffIDResult struct {
...
 // and returns a complete blobInfo of the copied layer, and a value for LayerDiffIDs if diffIDIsNeeded
-func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, pool *mpb.Progress) (types.BlobInfo, digest.Digest, error) {
+func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, toEncrypt bool, pool *mpb.Progress) (types.BlobInfo, digest.Digest, error) {
+	// This is used to authenticate the encrypted images that are already
+	// present in the image cached. For every layer we will try to unwrap
+	// the symmetric key with the provided private keys. If we fail, we will
+	// not allow the image to be provisioned.
+	if ic.checkAuthorization {
+		if srcInfo.MediaType == manifest.DockerV2Schema2LayerGzipEncMediaType ||
+			srcInfo.MediaType == manifest.DockerV2Schema2LayerEncMediaType {
```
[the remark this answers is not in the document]
lumjjb: I think this is a mistake on our part. We should rename `DockerV2Schema2LayerGzipEncMediaType` to use OCI instead. 

>I prefer adding an Encrypted bool field to BlobInfo and let the Manifest update it
I think this would work as well!

Kind of related to this - there is one issue that we were running into that is bugging us a little.

`copyUpdatedConfigAndManifest`(https://github.com/containers/image/blob/2178abd5f9b1b9e70c0aca05780d1defba350a53/copy/copy.go#L554) updates the manifest of the SRC and then copies it over to DST. Which is a bit tricky here because `dockerv2` doesn't support encryption, but it should not prevent encrypting a `dockerv2` image into an OCI encrypted image. I was wondering if it could be done by copying SRC to DST then applying the layer updates? What are your thoughts about this?

lumjjb:
Generate RSA keys for testing
```
openssl genrsa -out private.key 1024
openssl rsa -in private.key -pubout > public.key
```

Skopeo (to make, we use `make binary-local` due to some issue with container build). 
```
# Encrypting image
./skopeo copy --dest-tls-verify=false --dest-recipient jwe:./public.key oci:nginx_local:latest  oci:skopeo_nginx_jwe:latest

# Decrypting image
./skopeo copy --src-tls-verify=false --src-key ./private.key  oci:skopeo_nginx_jwe:latest oci:nginx_ocilayout:latest
```

We have a PR open to update that branch to use the new changes with the `ocicrypt` refactor, PR is as follows: https://github.com/harche/skopeo/pull/3.

copy/copy.go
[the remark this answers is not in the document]
vrothberg: I think this can be solved by doing something similar to https://github.com/containers/image/pull/563/files#diff-b3c9ca9ec48c31df0619e70595f4bd62R84-R91. While updating the manifests, we check each BlobInfo and update the mime types accordingly.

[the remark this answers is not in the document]
lumjjb: I think the `src` vs `dst` update would still be an issue - would you think we should move the code block from https://github.com/containers/image/blob/2178abd5f9b1b9e70c0aca05780d1defba350a53/copy/copy.go#L541-L559 

down, and so do something like

```
manifest, err := ic.c.src.Manifest(...)
...
if err := ic.c.dest.PutManifest(ctx, manifest); err != nil {
		return nil, errors.Wrap(err, "Error writing manifest")
}

_, err := ic.c.dest.UpdatedImage(ctx, *ic.manifestUpdates)
if err != nil {
	return nil, errors.Wrap(err, "Error creating an updated image manifest")
}
manifest, err := ic.c.src.Manifest(...)
```

lumjjb:
@vrothberg addressed the nits and most of the comments in https://github.com/lumjjb/image/pull/4. I think we can discuss the other ones during the call tomorrow!

rhatdan:
Looks like this needs a rebase again.

vrothberg:
@lumjjb, apologies for the longer silence. I wanted to run another review last week but didn't find the time.

How is the OCI upstream work going? I saw some discussions around the mediatype naming. Has that settled?

lumjjb:
Rebased! Also - we will have a skopeo PR with added tests in the CI in a bit (later today or tomorrow). And I think we should be good to merge the image library and skopeo if all is good.

>How is the OCI upstream work going? I saw some discussions around the mediatype naming. Has that settled?

@vrothberg The conclusion was that we would support generic suffixes, and have this be defined as part of https://github.com/opencontainers/artifacts. This would then be extended to encryption. The only change decided was that we should use `+encrypted` instead of `+enc`. We have made the changes in `containers/ocicrypt`. So there are a couple action items to move this forward now.

lumjjb:
Rebased again.

lumjjb:
@vrothberg  The CI seems to be failing for a unrelated failed test. Do you have any clues on what's going on here?

lumjjb:
It seems to pass now after rerunning the CI

copy/copy.go
```diff
@@ -687,9 +771,32 @@ type diffIDResult struct {
 
 // copyLayer copies a layer with srcInfo (with known Digest and Annotations and possibly known Size) in src to dest, perhaps compressing it if canCompress,
 // and returns a complete blobInfo of the copied layer, and a value for LayerDiffIDs if diffIDIsNeeded
-func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, pool *mpb.Progress) (types.BlobInfo, digest.Digest, error) {
+func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, toEncrypt bool, pool *mpb.Progress) (types.BlobInfo, digest.Digest, error) {
+	// This is used to authenticate the encrypted images that are already
+	// present in the image cached. For every layer we will try to unwrap
```
vrothberg: s/cached/cache/

copy/copy.go
```diff
@@ -816,7 +926,32 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
...
+		}
+
+		newDesc := ocispec.Descriptor{
+			Annotations: srcInfo.Annotations,
+		}
+
+		var d digest.Digest
+		srcStream, d, err = ocicrypt.DecryptLayer(c.decryptConfig, srcStream, newDesc, false)
+		if err != nil {
+			return types.BlobInfo{}, errors.Wrapf(err, "Error decrypting layer %s", srcInfo.Digest)
```
vrothberg: s/layer/blob/ since it could be a config as well. Same applies to other error locations.

copy/copy.go
```diff
@@ -894,6 +1029,44 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
...
+			desc := ocispec.Descriptor{
+				MediaType:   srcInfo.MediaType,
+				Digest:      srcInfo.Digest,
+				Size:        srcInfo.Size,
+				Annotations: annotations,
+			}
+
+			s, fin, err := ocicrypt.EncryptLayer(c.encryptConfig, destStream, desc)
+			if err != nil {
+				return types.BlobInfo{}, errors.Wrapf(err, "Image encryption failed for the digest %+v", srcInfo.Digest)
```
vrothberg: This error message has a quite different wording. I prefer keeping it consistent.

copy/copy.go
```diff
@@ -918,6 +1091,32 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
...
+		// to enable check authorization at a later time.
+		// TODO: Copy over only encryption related annotations
+		if uploadedInfo.Annotations == nil {
+			uploadedInfo.Annotations = map[string]string{}
+		}
+		for k, v := range srcInfo.Annotations {
+			uploadedInfo.Annotations[k] = v
+		}
+	}
+	if encrypted {
```
vrothberg: Micro optimization: `} else if encrypted {`

copy/copy.go
```diff
@@ -918,6 +1091,32 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
...
+
+		// Ensure that encryption annotations are present in the new manifest
+		// to enable check authorization at a later time.
+		// TODO: Copy over only encryption related annotations
+		if uploadedInfo.Annotations == nil {
+			uploadedInfo.Annotations = map[string]string{}
+		}
+		for k, v := range srcInfo.Annotations {
+			uploadedInfo.Annotations[k] = v
+		}
```
vrothberg: This seems redundant to line 1087 (same for the `encrypted` case).

lumjjb:
Thanks for the review! Will address the comments

lumjjb:
>https://github.com/containers/image/pull/673#discussion_r331953769

I can't seem to comment on this from the UI for some reason... This was taken out of `copier` is that okay as well? 

```
// copier allows us to keep track of diffID values for blobs, and other
// data shared across one or more images in a possible manifest list.
type copier struct {
    dest              types.ImageDestination
    rawSource         types.ImageSource
    reportWriter      io.Writer
    progressOutput    io.Writer
    progressInterval  time.Duration
    progress          chan types.ProgressProperties
    blobInfoCache     types.BlobInfoCache
    copyInParallel    bool
    compressionFormat compression.Algorithm
    compressionLevel  *int
    decryptConfig     *encconfig.DecryptConfig
    encryptConfig     *encconfig.EncryptConfig
}

type imageCopier struct {
    c                  *copier
    manifestUpdates    *types.ManifestUpdateOptions
    src                types.Image
    diffIDsAreNeeded   bool
    canModifyManifest  bool
    canSubstituteBlobs bool
    checkAuthorization bool
    decryptConfig      *encconfig.DecryptConfig
    encryptConfig      *encconfig.EncryptConfig
    encryptLayers      *[]int
}```

File: copy/copy.go [hunk 12]
```diff
@@ -121,6 +144,23 @@ type Options struct {
 	Progress         chan types.ProgressProperties // Reported to when ProgressInterval has arrived for a single artifact+offset.
 	// manifest MIME type of image set by user. "" is default and means use the autodetection to the the manifest MIME type
 	ForceManifestMIMEType string
+	// CheckAuthorization is an indication if the image does not need to be decrypted, but only
+	// requires checking if the it can be decrypted.
+	// This is an optimization for cached images to see if provided keys have the authorization
+	// to use the cached content without having to decrypt the layers again.
+	CheckAuthorization bool
```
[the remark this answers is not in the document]
lumjjb: >The Skopeo PR containers/skopeo#732 does not use the field. Is that something we really need? How is it intended to be used?

This is intended to be used by cri-o. In kubernetes, the creation of container should be given the option to check if the image is decryptable if it already exists in the cache. Just to make sure the user is able to use the image.

copy/copy.go
```diff
@@ -121,6 +144,23 @@ type Options struct {
 	Progress         chan types.ProgressProperties // Reported to when ProgressInterval has arrived for a single artifact+offset.
 	// manifest MIME type of image set by user. "" is default and means use the autodetection to the the manifest MIME type
 	ForceManifestMIMEType string
+	// CheckAuthorization is an indication if the image does not need to be decrypted, but only
+	// requires checking if the it can be decrypted.
+	// This is an optimization for cached images to see if provided keys have the authorization
```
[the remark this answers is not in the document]
lumjjb: Cached in this case means that the image exists in the store. This is more relevant in the context of the container runtime, cri-o. In this case, the scenario is where an image had been downloaded and unpacked during `ImagePull`, but we should be given the option to check if the image is decryptable at runtime as a security check. This is to prevent user B from using user A's encrypted image, just because they are on the same host and the image has already been unpacked.

copy/copy.go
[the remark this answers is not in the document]
vrothberg: Thanks! We really need a test for this in the Skopeo PR to prevent us from silently regressing in the future. But before doing that, let's find consensus on the API.

vrothberg:
> I can't seem to comment on this from the UI for some reason... This was taken out of copier is that okay as well?

Sure, both ways is fine :+1: Thanks!

copy/copy.go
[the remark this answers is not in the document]
lumjjb: Linking the relevant KEP to the thread: https://github.com/kubernetes/enhancements/pull/1066

This picture is helpful:

image

copy/copy.go
lumjjb: I believe it is possible to do decryption and re-encrypting an image in a single pass of `copy.Image`. So i think this should be okay?

lumjjb: But in that case, let me add a statement to set it to PreserveOriginal.

vrothberg: > I believe it is possible to do decryption and re-encrypting an image in a single pass of copy.Image. So i think this should be okay?

Very fair point! We should add a test to Skopeo for that.

copy/copy.go
```diff
@@ -806,7 +916,7 @@ func computeDiffID(stream io.Reader, decompressor compression.DecompressorFunc)
 // and returns a complete blobInfo of the copied blob.
 func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, srcInfo types.BlobInfo,
```
mtrmac: This is now far too big to be a single function. I don’t know what’s a _good_ way to restructure it yet (basically a pipeline of the `// ===`-demarcated `Reader` transformations, each one in a single function?), but this is no longer maintainable.

And we really need to have tests for at least individual pieces of this process, if not all the combinations.

copy/copy.go
```diff
@@ -894,6 +1029,44 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
...
+			s, fin, err := ocicrypt.EncryptLayer(c.encryptConfig, destStream, desc)
+			if err != nil {
+				return types.BlobInfo{}, errors.Wrapf(err, "Image encryption failed for the digest %+v", srcInfo.Digest)
+			}
+
+			destStream = s
+			finalizer = fin
+			inputInfo.Digest = ""
+			inputInfo.Size = -1
+			inputInfo.MediaType = encryptMediaType
```
mtrmac: This should not be something `copy.Image` does; let `UpdatedImage` handle that, like compression.  Instead of `GetEncryptedMediaType`, have `types.Image.CanEncryptLayer`==`genericManifest.CanEncryptLayer`, and leave the MIME types to the manifest implementations.

copy/copy.go
```diff
@@ -257,6 +297,10 @@ func (c *copier) copyOneImage(ctx context.Context, policyContext *signature.Poli
 		return nil, errors.Wrapf(err, "Error initializing image from source %s", transports.ImageName(c.rawSource.Reference()))
 	}
 
+	if options.EncryptLayers != nil && !src.SupportsEncryption(ctx) {
```
mtrmac: Well… what _actually_ matters is whether encryption is supported by the manifest type that ends up being written to the destination (accounting for `otherManifestMIMETypeCandidates`).

This might end up being good enough — or it might be a suggestion that the whole capability/edit process needs a deeper look.

(There are strong parallels to the compression decisions, which have not dealt with this correctly I think.)

image/manifest.go
```diff
@@ -44,6 +44,8 @@ type genericManifest interface {
 	// UpdatedImage returns a types.Image modified according to options.
 	// This does not change the state of the original Image object.
 	UpdatedImage(ctx context.Context, options types.ManifestUpdateOptions) (types.Image, error)
+	// SupportsEncryption returns if encryption is supported for the manifest type
+	SupportsEncryption(ctx context.Context) bool
```
mtrmac: See the 
> Well… what actually matters is whether encryption is supported by the manifest type that ends up being written to the destination 

comment elsewhere, I’m not sure this is a good long-term API — this way one can only ask about support for an existing image _soruce_, not for a hypothetical _destination_.

manifest/oci.go
```diff
@@ -241,3 +251,34 @@ func (m *OCI1) ImageID([]digest.Digest) (string, error) {
 	}
 	return m.Config.Digest.Hex(), nil
 }
+
+// IsEncryptedLayer indicates whether the blob is is encrypted
+func IsEncryptedLayer(b types.BlobInfo) bool {
```
mtrmac: Does this _have_ to be a public function like this, with knowledge about all possible kinds of manifest types?

Maybe it should be something the image/manifest implementation returns (in _some_ way), with each format implementation only dealing with its own formats.

manifest/oci.go
```diff
@@ -241,3 +251,34 @@ func (m *OCI1) ImageID([]digest.Digest) (string, error) {
...
+	// Check if encrypted mediatype
+	return b.MediaType == ociencspec.MediaTypeLayerGzipEnc ||
+		b.MediaType == ociencspec.MediaTypeLayerEnc
+}
+
+// GetEncryptedMediaType will return the mediatype to its encrypted counterpart and return
+// an error if the mediatype does not support encryption
+func GetEncryptedMediaType(mediatype string) (string, error) {
+	switch mediatype {
+	case DockerV2Schema2LayerMediaType, imgspecv1.MediaTypeImageLayerGzip:
```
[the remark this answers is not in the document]
lumjjb: >Don't we need the same for zstd-compressed layers as well? I am thinking whether we could make use of the artifact idea here and use +encryption as a general suffix for supported mime types (in contrast to having dedicated variables for all combinations).

Yea, I'll update this to check for the suffix instead.

manifest/oci.go
```diff
@@ -241,3 +251,34 @@ func (m *OCI1) ImageID([]digest.Digest) (string, error) {
...
+// IsEncryptedLayer indicates whether the blob is is encrypted
+func IsEncryptedLayer(b types.BlobInfo) bool {
+	// Check if encrypted mediatype
+	return b.MediaType == ociencspec.MediaTypeLayerGzipEnc ||
+		b.MediaType == ociencspec.MediaTypeLayerEnc
+}
+
+// GetEncryptedMediaType will return the mediatype to its encrypted counterpart and return
+// an error if the mediatype does not support encryption
+func GetEncryptedMediaType(mediatype string) (string, error) {
```
mtrmac: I’d much prefer for this not to be a public part of the stable API.

manifest/oci.go
```diff
@@ -241,3 +251,34 @@ func (m *OCI1) ImageID([]digest.Digest) (string, error) {
...
+		return ociencspec.MediaTypeLayerGzipEnc, nil
+	case imgspecv1.MediaTypeImageLayer:
+		return ociencspec.MediaTypeLayerEnc, nil
+	}
+	return "", errors.Errorf("unsupported mediatype to encrypt: %v", mediatype)
+}
+
+// GetDecryptedMediaType will return the mediatype to its decrypted counterpart and return
+// an error if the mediatype does not support encryption
+func GetDecryptedMediaType(mediatype string) (string, error) {
```
mtrmac: I’d much prefer for this not to be a public part of the stable API. Is there even anything that calls it?

copy/copy.go
```diff
@@ -816,7 +926,32 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
...
+		if c.decryptConfig == nil {
+			return types.BlobInfo{}, ErrDecryptParamsMissing
+		}
+
+		newDesc := ocispec.Descriptor{
+			Annotations: srcInfo.Annotations,
+		}
+
+		var d digest.Digest
+		srcStream, d, err = ocicrypt.DecryptLayer(c.decryptConfig, srcStream, newDesc, false)
```
mtrmac: (Why is it OK for the generic `copy` code to hard-code OCI-specific format implementations?

It _might_ be, but it’s not trivially clear.

If the code hard-codes OCI-specific handling here, the user-visible parameters in `SystemContext` should probably also explicitly say OCI in the name. … but then, can we legitimately name that OCI without being OCI-accepted?)

types/types.go
```diff
@@ -120,6 +133,9 @@ type BlobInfo struct {
 	// MIME type for compressed layers (e.g., gzip or zstd). This field MUST be
 	// set when `CompressionOperation == Compress`.
 	CompressionAlgorithm *compression.Algorithm
+	// CryptoOperation is used in Image.UpdateLayerInfos to instruct
+	// whether the original layer was encrypted/decrypted
+	CryptoOperation LayerCrypto
```
mtrmac: This really needs cleaning up; “edits to blob” are not “properties of a blob”, and should be separated / only maintained for `ManifestUpdateOptions`.

(Yes, compression added the parameters here.)

types/types.go
```diff
@@ -365,6 +381,8 @@ type Image interface {
 	// Everything in options.InformationOnly should be provided, other fields should be set only if a modification is desired.
 	// This does not change the state of the original Image object.
 	UpdatedImage(ctx context.Context, options ManifestUpdateOptions) (Image, error)
+	// SupportsEncryption returns an indicator that the image supports encryption
+	SupportsEncryption(ctx context.Context) bool
```
mtrmac: It’s unclear that this makes sense, see the other comments.

types/types.go
```diff
@@ -498,6 +516,8 @@ type SystemContext struct {
 	DockerInsecureSkipTLSVerify OptionalBool
 	// if nil, the library tries to parse ~/.docker/config.json to retrieve credentials
 	DockerAuthConfig *DockerAuthConfig
+	// if not nil, CryptoConfig will be used to encrypt/decrypt images
+	CryptoConfig *encconfig.CryptoConfig
```
mtrmac: Is this really appropriate for `SystemContext`, or just for `Options`?

Also see the “OCI” naming speculation elsewhere.

copy/copy.go
mtrmac: … “This” being editing the MIME type value.

mtrmac:
> I’m afraid this is not a full review, just a quick pass through the public API aspects.

(I realize that some of the design predates this PR, and it’s not something you’ve really signed up to do.

Also, some of those changes (the compression approach updates / splitting `copyBlobFromStream`) would probably be better to deal with as separate, more focused PRs.)

copy/copy.go
lumjjb: I discussed this with @harche and seems like for now it is more straightforward to do either encryption or decryption in a single pass - there isn't a usecase to how it is being done now to do both at the same time. I'll do some checking to ensure that in the library as well.

copy/copy.go
lumjjb: Gotcha - i think this is an artifact from before we had added fields to BlobInfo, so this can be removed.

copy/copy.go
lumjjb: I agree with this! I think the challenge that we ran into here was the way [`(ic *imageCopier) copyUpdatedConfigAndManifest`](https://github.com/containers/image/blob/master/copy/copy.go#L558) worked. I'm guessing we need to rewrite that function to copy to dest then update instead of the other way around.

mtrmac: Well… the layers must be uploaded before it’s valid to upload a manifest, and it’s only when uploading the manifest that we can tell whether the registry supports that manifest type or not.

So, it’s not something the code can always predict correctly, without possibly encrypting all that data and then failing; therefore, it _must_ be fine for the encryption decision to happen before that `otherManifestMIMETypeCandidates` loop, and for that encryption decision to restrict which of the other candidates have a chance of succeeding.

It’s more a matter of correctly dealing with `copy.Options.ForceManifestMIMEType` right now.

mtrmac: (To an extent, it’s quite fine to leave some of the complex cases unimplemented, if they clearly fail with an error. I’m more worried about embedding the incorrect assumptions into the stable API.)

mtrmac: Separately, there should exist something to reject copying encrypted blobs (whether unmodified, or because the caller requests encryption), to transports / formats that can’t support them.

That’s at least `containers-storage:`, and there may be others (at least the `docker/tarfile` two).

copy/copy.go
```diff
@@ -816,7 +926,32 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
 	// Note that for this check we don't use the stronger "validationSucceeded" indicator, because
 	// dest.PutBlob may detect that the layer already exists, in which case we don't
 	// read stream to the end, and validation does not happen.
-	digestingReader, err := newDigestingReader(srcStream, srcInfo.Digest)
+
+	var decrypted bool
+	var err error
+	if manifest.IsEncryptedLayer(srcInfo) {
```
mtrmac: This can’t copy encrypted layers unmodified from one registry to another?

copy/copy.go
lumjjb: Removed

types/types.go
lumjjb: I think the thought here is that like `ImagePullSecrets`, `ImageDecryptSecrets` would be represented in a similar way. We looked at integrating this into `cri-o`, and using systemContext made things a lot cleaner - so as not to need to add the parameter too every struct which the flow of a request made its way to.

mtrmac: The `SystemContext` …unwieldy type… exists primarily to insulate generic consumers of `types.ImageReference` from transport-specific parameters (so that `copy.Image` does not need to have explicit knowledge of “`docker://` transport parameters”, “`containers-storage:` parameters, “`dir:` parameters” and all the individual details directly in its API.

(Of course the transport dependency does not quite _go away_, it is here in `SystemContext`, and `copy.Image` needs to do fair amount of logic that is parametrized e.g. by the various `bool`-returning methods in `ImageDestination`, formulated as a generic facility, but transport-specific in intent/origin / choice of _those_ parameters and not others. But still, the `copy.Image` code is completely unaware of things like `DockerAuthConfig`, and in that sense `SystemContext` does its job well.)

Still, the way it collects _completely unrelated_ options, including options that could never actually be used at the same time, is unwieldy, impossible to generalize (e.g. to third-party transports, like some of the things Buildah does), and it makes option flow / usage difficult to reason about:

> using systemContext made things a lot cleaner - so as not to need to add the parameter too every struct which the flow of a request made its way to.

Another way to view the same thing is is that `SystemContext` made things a lot more confusing — the data was propagated everywhere without any ability to rely on the compiler to locate actual uses, or to report that a required value was not set by a caller.


Nothing in this PR adds any consumer of the encryption configuration in the transports or their helpers like `sysregistries2`, so this does not _need_ to be in `SystemContext` to preserve the current API. AFAICS the only consumer of the options is `copy.Image`, and `copy.Image` has `copy.Options`, a private, much better-scoped way to add options.

lumjjb: Okay - yea this sounds good. I think I had some misconceptions about the role of systemContext. @harche and I had this discussion a while back as well, it seems clearer now that having an option is preferred.

manifest/oci.go
lumjjb: This can be private - ditto for the other 2 functions `GetEnc/DecMediatype`

copy/copy.go
mtrmac: > I think having an encryption abstraction is fine, but at the moment, there isn't an alternative implementation, so defining the interface for the abstraction will most likely end up in something similar or identical to what we already have.

Not if the encryption systems are different, and require different formats/kinds of keys and other parameters. We would probably end up with a `SystemContext`-like way to hide the differences… or maybe require the caller to pass an encryption-system-specific key set object that has already been initialized, and somehow match it to the mechanisms supported by various manifest types.

The “generic” interface like that, however it ends up looking, would be quite different from the current one. Hence the suggestion that the API parameters to this encryption mechanism should explicitly say something like “OCI”, to be differentiated from the hypothetical future one.

lumjjb: Alright sounds good! - so for now I will add OCI prefixed naming for encryption specific variables.

copy/copy.go
lumjjb: I've taken a stab at what could be done to avoid this - minus the need to change the way the actual manifest updates are done still on the source than the destination. 

What do you think of something like this?

https://github.com/containers/image/pull/673/commits/78c7bece4c6f6e977aec63b832b0d0adf2e6e9b4

lumjjb: Added this in https://github.com/containers/image/pull/673/commits/78c7bece4c6f6e977aec63b832b0d0adf2e6e9b4

copy/copy.go
lumjjb: We've just added support for this. You can now copy images that are encrypted to another registry (remaining encrypted). Providing decryption keys allows decryption as well if one would like to decrypt the image. We will add some examples to the Skopeo tests.

copy/copy.go
lumjjb: If this is something that works, I can look into how we can move the manifest update to be based on destination manifest instead so we can remove that source mimetype dependency.

lumjjb:
@mtrmac Addressed most of the comments. I think the outstanding ones have to deal with the existing mechanisms in `containers/image`. I'd be happy to help implement some of those changes, but I think we can talk though it. Shall we schedule some time to chat about what we can help with here?

copy/copy.go
lumjjb: @mtrmac Any thoughts on this idea?

mtrmac: (Looking at the linked commit https://github.com/containers/image/commit/78c7bece4c6f6e977aec63b832b0d0adf2e6e9b4 ):

- There is `determineManifestConversion`, `supportsEncryption` should not duplicate (insufficiently) its logic
- `manifestSupportsEncryption` looks like something that should be provided by `c/image/manifest` (directly by the manifest type, or in some other abstraction?), not hard-coded in `c/image/copy`
  - (The same, to an extent, applies to `isOCIEncrypted`)

copy/copy.go
```diff
@@ -1074,7 +1151,33 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
...
+		srcInfo.Size = -1
+		for k := range srcInfo.Annotations {
+			if strings.HasPrefix(k, "org.opencontainers.image.enc") {
+				delete(srcInfo.Annotations, k)
+			}
+		}
+		decrypted = true
+	}
+
+	validateDigest := srcInfo.Digest != ""
```
mtrmac: Disabling digest validation should _never_ be possible, and it should validate digests of the raw blob inputs as they are referred to in the (possibly signed) manifest and as they are stored on the (possibly malicious) registry, not on some intermediate result after processing that (possibly malicious) input.

copy/copy.go
mtrmac: - Is the “TODO: Remove src.SupportsEncryption call…” item supposed to stay there, or should that already have been removed?

mtrmac: > It can't be removed right now because of the way manifests are updated, which use the source manifest instead of destination manifest.

Ah, right. If that isn’t fixed before this PR, the comment should at least clearly explain that this is the reason the check needs to exist.

mtrmac: > > There is determineManifestConversion, supportsEncryption should not duplicate (insufficiently) its logic
> 
> This would mean that we would have to pass in `src` image to `determineManifestConversion`.

(Discussed separately — that’s not what the concern was, but moving all of the logic to `determineManifestConversion` might cause `supportsEncryption` to be unnecessary anyway.)

> I initial thought that we don't want to create public interfaces. Does this change the initial concern? [#673 (comment)](https://github.com/containers/image/pull/673#discussion_r332082882)

Public APIs are a long-term commitment (or, so far, a reason to very frequently bump major versions, which is something I’d very much prefer to stop, and why I’m so worried about them). At the time of that comment it seemed that the function could be private, so making it private would have easily made the concern moot and would have given us complete flexibility to redesign that later as necessary.

If we do need a public function, or if avoiding a public function would lead to worse design (in this case tightly coupling c/image/copy and c/image/manifest) and breaking abstractions), the need to commit to it is not a _show-stopper_, but it’s a good reason to try hard to make the design general and long-lasting (which is why https://github.com/containers/image/pull/673#discussion_r332082882 tentatively suggests making it a per-format method of `c/image/manifest.Manifest` instead of a global function that understands all formats).

The “manifestSupportsEncryption” question is a bad match for `c/image/manifest.Manifest`, so maybe we need a `c/image/manifest.ManifestFormat`? No idea. Or maybe having a `MIMETypeSupportsEncryption` function very similar to the `manifestSupportsEncryption` one would work well enough (there’s at least one precedent for that already, and all of this somewhat hard-codes the _OCI_ encryption format anyway).

copy/copy.go
mtrmac: Basically.  (In the call, we’ve also identified questions about how the blob info cache works when layers can be encrypted, in particular whether layer reuse on push can be always safely used.)

lumjjb:
@mtrmac Updated with the changes we talked about during the call. - I think there's somethign we can do with the blob info cache, but I think it would be practically useful only when we get to cri-o, so we can look at that when we look at cri-o and KEP integration.

lumjjb:
rebased

lumjjb:
rebased

lumjjb:
As per discussions. PR is up to date and ready to merge!

Testing has been configured in https://github.com/containers/skopeo/pull/732 as `CopySuite.TestCopyEncryption `, which we can merge right after this (CI passing).
 
In addition, several manual tests have also been carried out for additional verification: https://gist.github.com/lumjjb/bda1faf8086a57c1a9261d951b9ddb6a

The outstanding discussions are documented in:
- https://github.com/containers/image/issues/755: Re-design BlobInfo *Operation
- https://github.com/containers/image/issues/746: Have copyUpdatedConfigAndManifest update not rely on src manifest restrictions

rhatdan:
Looks like we have one more rebase...

lumjjb:
rebased and poked CI, seems to be stuck on "receiving job"

lumjjb:
>Looks like we have one more rebase...

Rebased and CI ran success!

vrothberg:
As discussed, merging now. Thanks @lumjjb for this great work!

lumjjb:
Awesome! Thank you! 👍 👍 👍

PR #675, opened 2019-08-02
Update keyctl for other arches

nalind:
Use definitions from `golang.org/x/sys/unix` to get the `keyctl` package building on other Linux architectures.


nalind:
@QiWang19 PTAL.  I'm not sure why we have a copy of github.com/jsipprell/keyctl instead of vendoring it as a dependency.

mtrmac:
WRT having a copy, see the earlier conversation in https://github.com/containers/image/pull/619 ; things did not out quite as painless as expected.

nalind:
Ah, okay.  I missed that detail while skimming it earlier.  Thanks!

pkg/keyctl/key.go
```diff
@@ -20,8 +19,8 @@ type Key struct {
 }
 
 // ID returns the 32-bit kernel identifier for a specific key
-func (k *Key) ID() int32 {
-	return int32(k.id)
+func (k *Key) ID() int {
+	return int(k.id)
```
mtrmac: Strictly speaking, this is an API break. Do we need to change this? `int` would certainly be more idiomatic Go; still, the kernel `key_serial_t` is an `int32_t`, so this does not seem to hurt anything.

Same for `ID.ID` / `Keyring.ID`.

pkg/keyctl/sys_linux.go
```diff
@@ -3,119 +3,27 @@
...
-		uintptr(id),
-		0)
-
-	if errno != 0 {
-		err = errno
-		return 0, err
-	}
-	return int32(r1), nil
+func (id keyID) ID() int {
+	return int(id)
```
mtrmac: Non-blocking: AFAICS this method can be safely dropped; all `keyID` values are private, so no external caller can call this (and internal ones just use `int(keyID)` AFAICS.)

pkg/keyctl/keyring.go
```diff
@@ -3,11 +3,14 @@
 // license that can be found in the LICENSE file.
 
 // +build linux
-// +build 386 amd64
 
 // Package keyctl is a Go interface to linux kernel keyrings (keyctl interface)
```
mtrmac: Non-blocking, RFC: Maybe add a note here that the package is deprecated, and most callers should call `x/sys/unix` directly.

rhatdan:
LGTM with @mtrmac comments.

TomSweeneyRedHat:
Addresses:  https://github.com/containers/buildah/issues/1762

vrothberg:
Thanks for the fix, @nalind!

@mtrmac, maybe we can add cross-compilation to the CI?

vrothberg:
@mtrmac 's comments were not yet addressed.

TomSweeneyRedHat:
Once the dust settles I'm assuming we're respinning 3.0.0 or 3.0.1?

vrothberg:
3.0.1 ... 3.0.0 is already out

pkg/keyctl/key.go
nalind: Ah, you're right.  The APIs in `golang.org/x/sys/unix` use `int` for key types, so I changed the package to match without considering that.

pkg/keyctl/sys_linux.go
nalind: I think you're right - the only consumer I'm aware of is elsewhere in the library.  Would you like a follow-up PR?

mtrmac:
Follow-up: #676. PTAL.

mtrmac:
> @mtrmac, maybe we can add cross-compilation to the CI?

We do have cross-OS compilation in there. Cross-architecture is not trivial:
```shell
$ GOOS=linux GOARCH=arm64 go build -tags containers_image_openpgp ./...
vendor/github.com/containers/storage/drivers/overlay/overlay.go:21:2: build constraints exclude all Go files in $GOPATH/src/github.com/containers/image/vendor/github.com/containers/storage/drivers/quota
# exclude_graphdriver_overlay is not really representative, but let’s try:
macOS $ GOOS=linux GOARCH=arm64 go build -tags 'containers_image_openpgp exclude_graphdriver_overlay' ./...
# github.com/containers/image/vendor/github.com/containers/storage/pkg/loopback
vendor/github.com/containers/storage/pkg/loopback/attach_loopback.go:21:36: undefined: LoNameSize
# and many other failures
```

I guess the macOS failures could go away on a Linux box, and c/image CI maybe does not need to worry about representative c/storage builds.

PR #679, opened 2019-08-07
add AuthEnableKeyring  to disable/enable keyring

QiWang19:
Add AuthEnableKeyring field to SystemContext and set/get credentials to/from kernel keyring only if AuthEnableKeyring is true


rhatdan:
@vrothberg @mtrmac PTAL

PR #680, opened 2019-08-07
use enableKeyring(false) to disable keyring

QiWang19:
not set/get credentials to/from kernel keyring

### DIFF
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -32,6 +32,8 @@ var (
 	dockerHomePath          = filepath.FromSlash(".docker/config.json")
 	dockerLegacyHomePath    = ".dockercfg"
 
+	enableKeyring = false
+
 	// ErrNotLoggedIn is returned for users not logged into a registry
 	// that they are trying to logout of
 	ErrNotLoggedIn = errors.New("not logged in")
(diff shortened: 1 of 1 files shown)


QiWang19:
@mtrmac PTAL

pkg/docker/config/config.go
```diff
@@ -46,11 +48,11 @@ func SetAuthentication(sys *types.SystemContext, registry, username, password st
 			return false, setAuthToCredHelper(ch, registry, username, password)
 		}
 
-		// Set the credentials to kernel keyring if sys.AuthFile is not specified.
+		// Set the credentials to kernel keyring if enableKeyring is true.
 		// The keyring might not work in all environments (e.g., missing capability) and isn't supported on all platforms.
 		// Hence, we want to fall-back to using the authfile in case the keyring failed.
-		// However, if the sys.AuthFilePath is set, we want adhere to the user specification and not use the keyring.
```
mtrmac: Non-review note, for the record: This behavior was requested in #662 , and may (or may not) be desirable to reinstate.

(Right now it does not matter because `enableKeyring` is always `false`, so an extra condition would not make a difference.)

pkg/docker/config/config.go
```diff
@@ -46,11 +48,11 @@ func SetAuthentication(sys *types.SystemContext, registry, username, password st
...
-		if sys.AuthFilePath == "" {
+		// However, if the enableKeyring is false, we want adhere to the user specification and not use the keyring.
```
mtrmac: Non-blocking: `enableKeyring` is not really “user specification”

mtrmac:
@QiWang19 Please rebase.

TomSweeneyRedHat:
LGTM

PR #682, opened 2019-08-11
Replace `vendor.conf` by `go.mod`

praseodym:
Through `go mod init github.com/containers/image` followed by
`go mod tidy`.


praseodym:
Unfortunately CI fails, but I'm not quite sure why.

mtrmac:
Thanks for the PR.

The CI fails because it now runs the tests as root instead of an unprivileged user.

Overall, this PR is doing rather too many quite unrelated changes. They may well be desirable, but they need to be described, reviewed, and discussed one at a time — if not in separate PRs, at the very least in separate commits.

(Cc: @vrothberg , who has more experience with Go modules.)

vrothberg:
I concur with @mtrmac. Let's break the changes into easier to digest commits. The main part is replacing trash with go mod. Once that's done and the CI passes, we can improve from there.

praseodym:
> The CI fails because it now runs the tests as root instead of an unprivileged user.
> 
> Overall, this PR is doing rather too many quite unrelated changes. They may well be desirable, but they need to be described, reviewed, and discussed one at a time — if not in separate PRs, at the very least in separate commits.

I made those changes to simplify the CI flow in an attempt to fix it, so I would prefer to say that it's failing despite of those changes :slightly_smiling_face: 

I've now removed these changes and instead added `GO111MODULE="on"` in the `Makefile` to force Go modules to be enabled even within `GOPATH`. Unfortunately Travis has been stuck at "Job received" for about 45 minutes now, so I'll check back later.

Makefile
```diff
@@ -87,9 +72,7 @@ test-skopeo:
 	@echo === Testing skopeo build
 	@export GOPATH=$$(mktemp -d) && \
 		skopeo_path=$${GOPATH}/src/github.com/containers/skopeo && \
-		vendor_path=$${skopeo_path}/vendor/github.com/containers/image && \
 		git clone -b $(SKOPEO_BRANCH) https://github.com/$(SKOPEO_REPO) $${skopeo_path} && \
-		rm -rf $${vendor_path} && cp -r . $${vendor_path} && rm -rf $${vendor_path}/vendor && \
```
mtrmac: This probably needs to be updated, but it must not be just removed; the goal is to run the `skopeo` tests against the containers/image version being tested.

mtrmac:
> > The CI fails because it now runs the tests as root instead of an unprivileged user.
> > Overall, this PR is doing rather too many quite unrelated changes. They may well be desirable, but they need to be described, reviewed, and discussed one at a time — if not in separate PRs, at the very least in separate commits.
> 
> I made those changes to simplify the CI flow in an attempt to fix it, so I would prefer to say that it's failing despite of those changes 🙂

Well, the flow can only be simplified after it is understood :)

Makefile
praseodym: Reverted :)

storage/storage_test.go
```diff
@@ -178,9 +178,9 @@ func TestParse(t *testing.T) {
 func TestParseWithGraphDriverOptions(t *testing.T) {
 	optionLists := [][]string{
 		{},
-		{"unused1"},
-		{"unused1", "unused2"},
-		{"unused1", "unused2", "unused3"},
+		{"unused1=unused1"},
```
praseodym: CI now fails in `go test` with this error:
```
--- FAIL: TestParseWithGraphDriverOptions (0.00s)
    storage_test.go:95: vfs driver does not support unused1 options
```
Without this change, the test failed with a parse error.

I suspect this is due to changes in the updated containers/storage version I had to use to get cross-compilation to work, but I couldn’t find any related changes in that library. Any suggestions?

mtrmac: My first instinct would be to pare this down; get either cross-compilation, or `go mod`, working _separately_. I don’t think they have to land at exactly the same time.

Sure, if we do have to upgrade, (and we do, fairly soon anyway), and if the upgrade breaks this code, this will have to be dealt with one way or another eventually.

praseodym: 👍 Removed the containers/storage upgrade and the cross target from Travis CI and now it works (except for Travis CI flaking again).

mtrmac: > My first instinct would be to pare this down; get either cross-compilation, or `go mod`, working _separately_. I don’t think they have to land at exactly the same time.

Ah, sorry, I missed that we already run `make cross` in the CI. We do want to keep that working, but at least now we are one step closer.

Makefile
mtrmac: (If I’m reading this correctly, the final `rm -rf $${vendor_path}/vendor` could be removed, but it also does not hurt leaving it in, and it could be beneficial in case someone ran `go mod vendor`.)

praseodym: Let's keep it in just in case :)

storage/storage_test.go
praseodym: Fixed cross build by updating containers/storage to v1.12.10.

rhatdan:
LGTM

Makefile
```diff
@@ -48,35 +46,22 @@ install-docs: docs
...
+	GOOS=windows $(MAKE) build BUILDTAGS="$(BUILDTAGS) $(BUILD_TAGS_WINDOWS_CROSS)"
+	GOOS=darwin $(MAKE) build BUILDTAGS="$(BUILDTAGS) $(BUILD_TAGS_DARWIN_CROSS)"
 
 tools: tools.timestamp
 
 tools.timestamp: Makefile
 	@go get -u $(BUILDFLAGS) golang.org/x/lint/golint
 	@go get $(BUILDFLAGS) github.com/vbatts/git-validation
-	@go get -u github.com/rancher/trash
 	@touch tools.timestamp
```
mtrmac: Looking at the build logs:
```
go get: warning: modules disabled by GO111MODULE=auto in GOPATH/src;
	ignoring go.mod;
	see 'go help modules'
go get: warning: modules disabled by GO111MODULE=auto in GOPATH/src;
	ignoring go.mod;
	see 'go help modules'
```
It would be nice to clean that up; AFAICS it comes from the `tools.timestamp` rule.

---

More importantly, the `go get` commands above, when run outside of `GOPATH` actually add the three commands as dependencies in `go.mod`; that seems quite undesirable (consumers of the library neither need nor want to include our testing infrastructure).  Sadly, there isn’t an obvious way to bypass that (at least quickly skimming https://github.com/golang/go/issues/27643 and several of the linked discussions).

I guess the closest to working is the https://stackoverflow.com/questions/53368187/go-modules-installing-go-tools solution of using an unrelated temporary working directory (except that it should use `mktemp -d` instead of hard-coding a single shared=possibly malicious directory).

go.mod
```diff
@@ -0,0 +1,65 @@
...
+	github.com/opencontainers/go-digest v0.0.0-20180430190053-c9281466c8b2
+	github.com/opencontainers/image-spec v1.0.0
+	github.com/opencontainers/runc v0.0.0-20170216015123-6b1d0e76f239 // indirect
+	github.com/opencontainers/selinux v0.0.0-20181018180444-077c8b6d1c18
+	github.com/ostreedev/ostree-go v0.0.0-20181204105935-56f3a639dbc0
+	github.com/pborman/uuid v0.0.0-20170112150404-1b00554d8222 // indirect
+	github.com/pkg/errors v0.0.0-20161029093637-248dadf4e906
+	github.com/pquerna/ffjson v0.0.0-20181028064349-e517b90714f7 // indirect
+	github.com/satori/go.uuid v1.2.0 // indirect
+	github.com/sirupsen/logrus v1.4.1
```
mtrmac: (Note for the record: This is an upgrade from v1.0.0.)

go.mod
```diff
@@ -0,0 +1,65 @@
...
+	github.com/stretchr/testify v1.2.2
```
mtrmac: (Note for the record: This is an update from an untagged post-v1.1.4 commit.)

go.mod
```diff
@@ -0,0 +1,65 @@
...
+	github.com/vbatts/tar-split v0.10.2
+	github.com/vbauerster/mpb v3.4.0+incompatible
+	github.com/xeipuuv/gojsonpointer v0.0.0-20190809123943-df4f5c81cb3b // indirect
+	github.com/xeipuuv/gojsonreference v0.0.0-20180127040603-bd5ef7bd5415 // indirect
+	github.com/xeipuuv/gojsonschema v0.0.0-20190626083556-16a6735db808
+	go.etcd.io/bbolt v1.3.3 // indirect
+	golang.org/x/crypto v0.0.0-20190701094942-4def268fd1a4
+	golang.org/x/net v0.0.0-20190724013045-ca1201d0de80
+	golang.org/x/sync v0.0.0-20190423024810-112230192c58
+	golang.org/x/sys v0.0.0-20190804053845-51ab0e2deafa
```
mtrmac: (Note for the record: all of the /x/* above have been upgraded, fairly significantly.)

go.mod
```diff
@@ -0,0 +1,65 @@
+module github.com/containers/image
+
+go 1.12
```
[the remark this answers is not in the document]
praseodym: I used Go 1.12 to generate the `go.mod` file, but can change it to 1.11 if wanted.

Makefile
praseodym: I think adding an explicit `GO111MODULE=off` should fix it.

go.mod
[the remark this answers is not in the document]
mtrmac: Works for me :)

Makefile
mtrmac: For now; https://github.com/containers/skopeo/pull/695 says that Go 1.13 ignores that variable. But this is not my area of expertise.

praseodym: Oh great :(

praseodym: It looks like Go 1.13 only changes the behaviour of `GO111MODULE=auto` though:
https://github.com/golang/go/blob/ad4ed87f80c33f23bdd3767ef3208f15a1fb5c90/doc/go1.13.html#L181-L190

So explicitly setting `GO111MODULE=off` should still work with Go 1.13.

mtrmac: Thanks, looking at the code `GO111MODULE=off` does seem to set at least the local flags in Go’s `src/cmd/go/internal/modload/init.go` and `src/go/build/build.go` to turn off modules, so this code should be fine.

Yet, `GO111MODULE=off` was exactly what Skopeo was using before https://github.com/containers/skopeo/pull/695 , so I don’t quite know what is going on. Hopefully @vrothberg can help.

mtrmac:
LGTM, but I’d love to see an ACK by @vrothberg , especially (but not exclusively) WRT the `GO111MODULE` variable discussed above.

Makefile
```diff
@@ -48,35 +46,22 @@ install-docs: docs
...
 
 tools.timestamp: Makefile
-	@go get -u $(BUILDFLAGS) golang.org/x/lint/golint
-	@go get $(BUILDFLAGS) github.com/vbatts/git-validation
-	@go get -u github.com/rancher/trash
+	@GO111MODULE="off" go get -u $(BUILDFLAGS) golang.org/x/lint/golint
+	@GO111MODULE="off" go get $(BUILDFLAGS) github.com/vbatts/git-validation
 	@touch tools.timestamp
 
-vendor: tools.timestamp vendor.conf
```
vrothberg: I very much prefer keeping the vendor target. All other projects on `github.com/containers` do something as follows:

https://github.com/containers/skopeo/blob/481bb94c5f8e65bc83b97539e7e4e6ccb8428a73/Makefile#L178-L182

Makefile
vrothberg: Starting with Go v1.13 the variable will be ignored (since go modules are declared stable) but we still need it for backwards compat with 1.11 and 1.12.

As @mtrmac has mentioned, any `go get` and other go commands can potentially alter the module, so we need to be careful. What we do in buildah, for instance, is to keep the CI dependencies in a separate (sub) go module: https://github.com/containers/buildah/tree/master/tests/tools

What we do there is have a dedicated go module containing the tools for the CI that will be compiled on-demand. The code could also be downloaded on-demand, if desired. To trick go into believing we have dependencies on the tools, we have a dummy `.go` file with `import _ "github.com/.../..."` lines (https://github.com/containers/buildah/blob/master/tests/tools/tools.go).

I can take care of doing that once this PR is merged.

Makefile
```diff
@@ -99,8 +84,8 @@ fmt:
 	@gofmt -l -s -w $(SOURCE_DIRS)
 
 validate: lint
-	@go vet $(PACKAGES)
-	@test -z "$$(gofmt -s -l . | grep -ve '^vendor' | tee /dev/stderr)"
```
vrothberg: Please keep the `vendor` filter. We might have checked out the code locally via `go mod vendor`.

Makefile
vrothberg: I also want to preserve the having the code in `./vendor` (via `go mod vendor`). The workflow of many folks is used to that. If some don't want it, they are free to not use `make vendor`.

mtrmac:
> I very much prefer keeping the `./vendor` directory and the dependencies locally.

In what cases does that make a difference? I could guess non-module-aware IDEs, perhaps.

> Currently, all `github.com/containers/` projects use `go mod` only for dependency management and not for deduping and caching builds.

Note that the others are applications, and we distribute the `vendor` subdirectories in official releases; c/image is a library and it never did that.

Makefile
mtrmac: > Starting with Go v1.13 the variable will be ignored (since go modules are declared stable) but we still need it for backwards compat with 1.11 and 1.12.

That’s not what the code actually does AFAICT. There are quite a few checks for `os.Getenv("GO111MODULE") == "off"` in the codebase, even on `master`. Or have they dropped that check from some important place that affects these projects?

vrothberg:
> > I very much prefer keeping the `./vendor` directory and the dependencies locally.
> 
> In what cases does that make a difference? I could guess non-module-aware IDEs, perhaps.

Some IDEs are going wild and I found it helpful when updating/analyzing dependencies; sometimes tracking down bugs.

> > Currently, all `github.com/containers/` projects use `go mod` only for dependency management and not for deduping and caching builds.
> 
> Note that the others are applications, and we distribute the `vendor` subdirectories in official releases; c/image is a library and it never did that.

I wasn't suggesting to commit the `./vendor` folder but to preserve the `make vendor` target which downloads the dependencies - and preserves the previous behaviour. I feel it makes sense to have an opinionated way for dealing with `go mod` as there are many ways to do something wrong.

Makefile
vrothberg: Thanks for clarifying. I didn't actually check the go code but based my statement on the various build issues in Fedora when flipping it to v1.13. So it seems that `GO111MODULE` can still be turned off explicitly ... although it did not help with `go build` apparently.

mtrmac:
Sure, I’m fine with preserving the `make vendor` target, as an option.

OTOH consumers of c/image will just list it in their `go.mod` files, so the CI should run a module-aware build without relying on any `vendor` subdirectory, the way consumers would do.

I don’t know whether that means we should preserve the `build`/`build-internal` dichotomy, and modify the CI to call the non-vendoring versions, or whether we should never call the `vendor` target automatically. I don’t feel strongly about it, so far I’m leaning towards the latter for simplicity.

vrothberg:
For the build to use `./vendor` we need to add `-mod=vendor` to `go build`. So we could add `make build-vendor` target and use both in the CI?

mtrmac:
I don’t quite see the point of doing that in the CI, the CI is going to start from a fresh checkout and it is going to have to populate `./vendor` immediately before relying on it. We would be mostly testing whether `go build` and `go mod vendor` are internally consistent—and, well, that the `Makefile` rules were not broken; I suppose that’s something.

vrothberg:
@mtrmac, I am not sure I can follow then. Do you want preserve `make vendor` and do vendor-builds or fully rely on go mod?

mtrmac:
I think that definitely the CI, and probably the primary local build method, should rely on `go mod` the way our consumers do.

I don’t personally see preserving `make vendor` of anything about the `vendor` directory as important, but if you want `make vendor` (and `make build-vendor`?) to exist, that’s perfectly fine with me.

As for testing the vendor build in the CI in addition to the module-based build, I see that as even less important than preserving the vendor targets at all, but I’m fine with that as well, as long as it does not significantly increase the CI run time.

vrothberg:
Thanks for clarifying! Let's drop the `make vendor` target then. I prefer having one way rather than multiple ways to build the project.

vrothberg:
@praseodym, are you still working on it?

praseodym:
> @praseodym, are you still working on it?

Oops, I missed the review comments in rest of the discussion. I've just rebased this and addressed one comment.

Makefile
praseodym: Done!

mtrmac:
LGTM. Thanks!

@vrothberg I’ll leave this up to you to merge, in case there were some tricky aspects WRT updating consumers of c/image that already use Go modules.

vrothberg:
LGTM, thanks for contributing, @praseodym!

> @vrothberg I’ll leave this up to you to merge, in case there were some tricky aspects WRT updating consumers of c/image that already use Go modules.

To the best of my knowledge, nothing will change. Consumers of c/image already marked it as `+incompatible` since we are not doing the version-path dance. The nice thing now is that vendoring c/image will be a bit easier as go mod will fetch the potentially updated dependencies as well.

After the merge here, I will iterate over the dependencies and check for new versions to make sure we're releasing an up to date module.

PR #683, opened 2019-08-12
move keyctl to internal & func remove auth from keyring

QiWang19:
Move pkg/keyctl to internal/pkg/keyctl.
Add method removeAllAuthFromKeyring.
Get key describes from keyring using KEYCTL_READ and KEYCTL_DESCRIBE, and remove them from keyring if the description has prefix 'container-registry-login:'.


QiWang19:
@mtrmac @vrothberg PTAL :slightly_smiling_face:

pkg/docker/config/config_linux.go
```diff
@@ -6,8 +6,11 @@ import (
 
 	"github.com/containers/image/pkg/keyctl"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 )
 
+const keyDescribePrefix = "container-registry-login:"
```
mtrmac: It’s good that there is a constant; `genDescription` should use it as well.

pkg/docker/config/config_linux.go
```diff
@@ -41,6 +44,37 @@ func deleteAuthFromKernelKeyring(registry string) error {
 	return key.Unlink()
 }
 
+func removeAllAuthFromKernelKeyring() error {
```
mtrmac: Nothing calls this AFAICS?!

pkg/keyctl/key.go
```diff
@@ -62,3 +62,12 @@ func (k *Key) Unlink() error {
 	_, err := unix.KeyctlInt(unix.KEYCTL_UNLINK, int(k.id), int(k.ring), 0, 0)
 	return err
 }
+
+// Describe returns a string describing the attributes of a specified key
+func Describe(kID ID) (string, error) {
```
mtrmac: No more public functions in this subpackage, please; we don’t want to maintain it as a stable public API.  Make them private in `pkg/docker/config`.

pkg/keyctl/keyring.go
```diff
@@ -77,3 +82,68 @@ func Link(parent Keyring, child ID) error {
...
+	*(*uint16)(unsafe.Pointer(&buf[0])) = uint16(0xABCD)
+
+	switch buf {
+	case [2]byte{0xCD, 0xAB}:
+		nativeEndian = binary.LittleEndian
+	case [2]byte{0xAB, 0xCD}:
+		nativeEndian = binary.BigEndian
+	default:
+		panic("Could not determine native endianness.")
+	}
```
mtrmac: Ouch…

I mean, if this is going to be using `unsafe` at all, I guess it could just as well use something vaguely like (completely untested!)
```go
tempId := *(*int32)(unsafe.Pointer(&buf[idx:idx+idSize])
```
below, without bothering with the `encoding/binary` package at all, but finding a way to not use `unsafe` would be much nicer. Not that I know how to do that right now.

pkg/keyctl/key.go
vrothberg: Maybe we could (should?) move this package into `./internal` before the next major bump?

QiWang19: why Unlink() can stay in pkg/keyctl/key.go but not Describe(kID ID)..
I think move keyctk package to `./internal` is better?

pkg/keyctl/keyring.go
QiWang19: I can use `*(*int32)(unsafe.Pointer(&byteKeyIDs[idx]))`

pkg/keyctl/key.go
mtrmac: The `pkg/keyctl` package was present in a release, so we can’t remove it, or any public functions/methods/types in it, without a major version bump.

Still, it’s not something users are expected to look for in c/image, and we don’t really want  to maintain or enhance it for such hypothetical external users — after all, it’s just a pretty small façade for `golang.org/x/sys/unix` now.

https://github.com/containers/image/pull/676 has already marked `pkg/keyctl` as deprecated: at the time, all callers could use either `golang.org/x/sys/unix` or `github.com/jsipprell/keyctl` without involving c/image.

This is a bit less true now that `ReadUserKeyring` seems to have no direct equivalent in either of those packages; still, c/image is not the obvious place to maintain that as a public utility.

Having the new code directly in `pkg/docker/config` (or maybe `pkg/docker/config/internal` or something like  `pkg/docker/config/internal/keyring`) would work well enough, I think; nothing else in the repo needs to care.

Before v4.0.0 we can then either moe the current `pkg/keyctl` to that same location, or just discard it replace its users with direct calls to `golang.org/x/sys/unix`.

internal/keyctl/keyring.go
```diff
@@ -11,6 +11,8 @@
 package keyctl
```
mtrmac: (Just so that it isn’t forgotten, we can’t _right now_ move the package to `internal/…`, unless we want the next release to be 4.0.0.)

pkg/docker/config/config.go
```diff
@@ -142,9 +142,14 @@ func RemoveAuthentication(sys *types.SystemContext, registry string) error {
...
 
-// RemoveAllAuthentication deletes all the credentials stored in auth.json
+// RemoveAllAuthentication deletes all the credentials stored in auth.json and kernel keyring
 func RemoveAllAuthentication(sys *types.SystemContext) error {
 	return modifyJSON(sys, func(auths *dockerConfigFile) (bool, error) {
+		// first try to remove credentials from kernel keyring
+		err := removeAllAuthFromKernelKeyring()
+		if err != nil {
+			logrus.Debugf("error removing credentials from kernel keyring")
+		}
```
mtrmac: Should this _only_ remove data from the kernel keyring, or the data in the JSON as well?

The code as is removes _both_ `CredHelpers` and `AuthConfigs`, even if they both contain entries for the same registry; so it seems consistent for the keyring to be cleared as well as the two other locations.

OTOH, `RemoveAuthentication` does remove the data from exactly one location, not from all.

I don’t know. What do others think?

QiWang19: podman logout -a will use RemoveAllAuthentication, so I think this should just clear all credentials from keyring, CredHelpers and AuthConfigs.

mtrmac: Oops, I misread this for some reason; I thought the code is exiting on success. Sorry.

Nevertheless, let’s see what others think.

internal/keyctl/keyring.go
vrothberg: Right, that's why I was saying "before the next major bump" :^)

pkg/docker/config/config.go
vrothberg: Unless there's a clear mechanism to control which authentication mechanism to use, removing all sounds good to me.

pkg/keyctl/key.go
QiWang19: I added the new code to pkg/docker/config_linux.go, should I move it to pkg/docker/config.go?
Do I need to add test `readUserKeyring` in  pkg/docker/config/config_test.go?

pkg/docker/config/config.go
QiWang19: I think this related the criteria of the card [Podman: Complete kernel keyring auth](https://jira.coreos.com/browse/RUN-487?workflowName=Software+Simplified+Workflow+for+Project+TEMPLATE&stepId=6).
If `podman logout -a` need to control which storage to remove, c/image need to support the control.

QiWang19: I want this PR only to complete the keyctl to remove all keyring. Just add function `removeAllAuthFromKernelKeyring()` without calling it.

There should be another PR to support Podman using keyring. `removeAllAuthFromKernelKeyring()` should be used in that PR. But I can't make the decision that how c/image support Podman using keyring. Add `AuthEnableKeyring` field to `SystemContext` ?

@rhatdan @mtrmac @vrothberg can you decide that?

vrothberg: @QiWang19, once we have consensus, can you open a PR over at libpod vendoring this branch (i.e., `rm_all_keyring`)? I want to make sure that libpod's CI is green with the new changes before merging them into c/image.

> Add AuthEnableKeyring field to SystemContext ?

I support that as users are able to opt-in. If `AuthEnableKeyring == true` and the keyring doesn't work (e.g., privileges) we will still fall back to using the authfile, right?

QiWang19: >If AuthEnableKeyring == true and the keyring doesn't work (e.g., privileges) we will still fall back to using the authfile
from this comment, no fallback. https://jira.coreos.com/browse/RUN-487?focusedCommentId=118874&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-118874
Podman will not use XDG_RUNTIME_DIR as default, no file to fall back

vrothberg: Thanks for clarifying @QiWang19. @mtrmac WDYT?

pkg/keyctl/key.go
mtrmac: `removeAllAuthFromKernelKeyring` is Linux-specific, so it needs to be in a `*_linux.go` file (or a differently-named file that uses an explicit build tag).

Both `config_linux.go` and a separate `keyring_linux.go` would be fine with me.

If unit tests exist, having them (again, in a Linux-specific file) in the package would be nice, sure. (If they don’t, it’s not quite blocking the PR … but this is not quite trivial code, so tests would be nice.)

pkg/docker/config/config.go
mtrmac: Ah, we do currently have the `enableKeyring` bool. I suppose, then, that this PR should make the code conditional on that bool as well, to be consistent with everything else.

We can then discuss the public API for opting in/out in a separate PR.

Does that work?

vrothberg: Works for me :+1:

pkg/docker/config/config_linux.go
```diff
@@ -75,5 +111,46 @@ func setAuthToKernelKeyring(registry, username, password string) error {
 }
 
 func genDescription(registry string) string {
-	return fmt.Sprintf("container-registry-login:%s", registry)
+	return fmt.Sprintf("%s%s", keyDescribePrefix, registry)
+}
+
+// readUserKeyring reads user keyring and returns slice of key id(key_serial_t) representing the IDs of all the keys that are linked to it
+func readUserKeyring() ([]int32, error) {
```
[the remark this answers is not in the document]
QiWang19: I can export the id of `Key struct` and set a Key variable with id `KEY_SPEC_USER_KEYRING` to call Key.Get() and remove `readUserKeyring()`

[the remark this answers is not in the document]
QiWang19: > I can export the id of `Key struct` and set a Key variable with id `KEY_SPEC_USER_KEYRING` to call Key.Get() and remove `readUserKeyring()`

Does this sound better?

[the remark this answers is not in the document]
mtrmac: I think the code as is good enough; not worth adding extra public interfaces to `pkg/keyring` for me, at least.

rhatdan:
Is this ready to merge then?

pkg/docker/config/config_linux.go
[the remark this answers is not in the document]
mtrmac: @QiWang19 Could you still revert the recent change, keeping the read loop in here so that we don’t make the `Key.Id` value public?

QiWang19:
revert the code and create a PR in libpod to test if can pass the CI https://github.com/containers/libpod/pull/3860

QiWang19:
@mtrmac @vrothberg PTAL

rhatdan:
@mtrmac @vrothberg Is this ready to merge.  We need to finish this up and get back to enabling the kernel keyring support.

vrothberg:
Thanks for clarifying, @mtrmac! Anything against moving the entire `pkg/keyctl` into `internal/`? #563 would entail a major version bump as well (unless we consider all behavioral changes as bug fixes).

rhatdan:
I think we should move it.

QiWang19:
PTAL @vrothberg @mtrmac

rhatdan:
LGTM

PR #684, opened 2019-08-21
copy: compression: default to gzip

vrothberg:
Unless specified otherwise, default to using the gzip compressor.  If
left nil, we'll run into a nil dereference breaking existing consumers
of containers/image.

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -170,6 +170,7 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
 		progressOutput = ioutil.Discard
 	}
 	copyInParallel := dest.HasThreadSafePutBlob() && rawSource.HasThreadSafeGetBlob()
+
 	c := &copier{
 		dest:             dest,
 		rawSource:        rawSource,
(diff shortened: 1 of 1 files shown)


vrothberg:
@giuseppe @mtrmac @rhatdan PTAL

That's fixing a regression and prevents consumers from breaking.

rhatdan:
LGTM
Can we get a PR into skopeo to fix the test suite.

vrothberg:
Skopeo PR https://github.com/containers/skopeo/pull/706

giuseppe:
there is also https://github.com/containers/skopeo/pull/671 that is ready for review.  If there is nothing holding it, let's merge it so we can get zstd into Skopeo

giuseppe:
The Skopeo PR passes the tests.  I'll merge

copy/copy.go
```diff
@@ -181,10 +182,18 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
 		// FIXME? The cache is used for sources and destinations equally, but we only have a SourceCtx and DestinationCtx.
 		// For now, use DestinationCtx (because blob reuse changes the behavior of the destination side more); eventually
 		// we might want to add a separate CommonCtx — or would that be too confusing?
-		blobInfoCache: blobinfocache.DefaultCache(options.DestinationCtx),
-
-		compressionFormat: *options.DestinationCtx.CompressionFormat,
-		compressionLevel:  options.DestinationCtx.CompressionLevel,
+		blobInfoCache:    blobinfocache.DefaultCache(options.DestinationCtx),
+		compressionLevel: options.DestinationCtx.CompressionLevel,
```
mtrmac: `options.DestinationCtx` can be `nil`, per the common semantics of `*types.SystemContext`.

copy/copy.go
```diff
@@ -181,10 +182,18 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
...
+	}
+	// Default to using gzip compression unless specified otherwise.
+	if options.DestinationCtx.CompressionFormat == nil {
```
mtrmac: `options.DestinationCtx` can be `nil`.

copy/copy.go
vrothberg: @giuseppe, can you take care of the remaining clean ups for that? And line 189?

copy/copy.go
vrothberg: https://github.com/containers/image/pull/688

vrothberg: https://github.com/containers/image/pull/688

PR #686, opened 2019-08-21
keyctl: clean up after tests

nalind:
A number of the `pkg/keyctl` tests link keys to keyrings, but weren't unlinking them at the end of the test, causing subsequent runs of the same test to fail.


rhatdan:
LGTM

mtrmac:
LGTM _assuming tests pass_. Thanks!

This probably needs to only be rebased now.

PR #687, opened 2019-08-21
pkg/blobinfocache: fixup unit tests for euid 0

nalind:
When testing, force our rootless ID to -1 to avoid using the hard-coded default location when the tests are being run by a privileged user.  Turn off CAP_DAC_OVERRIDE in the test so that when we expect to be denied the ability to create a subdirectory under a directory to which DAC says we can't write, that's what will actually happen, even for root.
I don't think we run tests as root in CI, so this may not be worth merging.  If so, I'm fine with that.

### DIFF
--- a/pkg/blobinfocache/default_test.go
+++ b/pkg/blobinfocache/default_test.go
@@ -13,6 +13,7 @@ import (
 	"github.com/containers/image/pkg/blobinfocache/memory"
 	"github.com/containers/image/types"
 	"github.com/stretchr/testify/assert"
+	"github.com/syndtr/gocapability/capability"
 )
 
 func TestBlobInfoCacheDir(t *testing.T) {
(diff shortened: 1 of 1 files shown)


pkg/blobinfocache/default_test.go
```diff
@@ -147,6 +148,12 @@ func TestDefaultCache(t *testing.T) {
 	}()
 	os.Unsetenv("HOME")
 	os.Unsetenv("XDG_DATA_HOME")
+	os.Setenv("_CONTAINERS_ROOTLESS_UID", "-1")
+	cap, err := capability.NewPid(0)
+	require.NoError(t, err)
+	cap.Unset(capability.EFFECTIVE|capability.PERMITTED|capability.INHERITABLE|capability.BOUNDING|capability.AMBIENT, capability.CAP_DAC_OVERRIDE)
+	err = cap.Apply(capability.EFFECTIVE | capability.PERMITTED | capability.INHERITABLE | capability.BOUNDING | capability.AMBIENT)
+	require.NoError(t, err)
```
mtrmac: How does this work for non-Linux systems?

With `_CONTAINERS_ROOTLESS_UID=-1` and both `XDG_DATA_HOME` and `HOME` unset, AFAICS `blobInfoCacheDir` should fail and then capabilities don’t matter; am I missing something?

(I suppose it would be natural and reasonably portable to just skip this test if EUID is 0, but now with `_CONTAINERS_ROOTLESS_UID`, this seems to actually trigger exactly the failure that ordinary unprivileged users trigger.)

vrothberg:
Ping to move it back up the mailboxes :^)

mtrmac:
Is all of this about the “	// Error running blobInfoCacheDir:” case not failing when run as root? It seems far simpler to just skip that part of the test if UID == 0. `TestBlobInfoCacheDir` has tested what path is used in that case, and the “success” case in `TestDefaultCache` has tested the code flow. That’s not _perfect_ code coverage but probably good enough; anyone modifying this will hopefully run tests locally as an unprivileged user.

nalind:
Yeah, I think that's what it was, though I can't remember where exactly I was running into an error.  I think it's been obsoleted by other changes that have landed since I opened this one.  Closing.

PR #688, opened 2019-08-22
copy: add nil checks

vrothberg:
`DestinationCtx` can conventionally be nil, so add some checks around
references to avoid potential nil dereferences.

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -182,11 +182,10 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
 		// FIXME? The cache is used for sources and destinations equally, but we only have a SourceCtx and DestinationCtx.
 		// For now, use DestinationCtx (because blob reuse changes the behavior of the destination side more); eventually
 		// we might want to add a separate CommonCtx — or would that be too confusing?
-		blobInfoCache:    blobinfocache.DefaultCache(options.DestinationCtx),
-		compressionLevel: options.DestinationCtx.CompressionLevel,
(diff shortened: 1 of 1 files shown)


vrothberg:
Addresses https://github.com/containers/image/pull/684#discussion_r316280790.

@giuseppe @mtrmac @rhatdan PTAL

vrothberg:
PR to throw Skopeo CI at it: https://github.com/containers/skopeo/pull/706

giuseppe:
LGTM

rhatdan:
LGTM

copy/copy.go
```diff
@@ -195,6 +194,10 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
 	} else {
 		c.compressionFormat = *options.DestinationCtx.CompressionFormat
 	}
+	if options.DestinationCtx == nil {
```
mtrmac: Shouldn’t this have been `!=`?

vrothberg: :man_facepalming: yes ... doing work in a hurry has it's consequences. I'll add a fix-fix-fix.

mtrmac:
@rhatdan Why was this merged with failing CI (and not even an acknowledgement that it is failing)?

Is there a story to it I’m not aware of?

copy/copy.go
vrothberg: Thanks for checking, @mtrmac! PR is here https://github.com/containers/image/pull/690. Apologies for the sloppiness.

PR #696, opened 2019-09-10
Use the same HTTP client for contacting the bearer token server and t…

umohnani8:
…he registry

Signed-off-by: Miloslav Trmač <mitr@redhat.com>

### DIFF
--- a/.travis.Dockerfile
+++ b/.travis.Dockerfile
@@ -1,10 +1,10 @@
-FROM ubuntu:artful
+FROM ubuntu:18.04
 
 RUN apt-get -qq update && \
     apt-get install -y sudo docker.io git make btrfs-tools libdevmapper-dev libgpgme-dev libostree-dev
 
-ADD https://storage.googleapis.com/golang/go1.11.1.linux-amd64.tar.gz /tmp
+ADD https://storage.googleapis.com/golang/go1.11.12.linux-amd64.tar.gz /tmp
 
-RUN tar -C /usr/local -xzf /tmp/go1.11.1.linux-amd64.tar.gz && \
-    rm /tmp/go1.11.1.linux-amd64.tar.gz && \
+RUN tar -C /usr/local -xzf /tmp/go1.11.12.linux-amd64.tar.gz && \
+    rm /tmp/go1.11.12.linux-amd64.tar.gz && \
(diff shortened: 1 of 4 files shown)


mtrmac:
ACK, but
```
The repository 'http://security.ubuntu.com/ubuntu artful-security Release' does not have a Release file.
```
will probably require updating `.travis.Dockerfile` (and hopefully not much more)

.travis.Dockerfile
```diff
@@ -1,10 +1,10 @@
...
 
 RUN apt-get -qq update && \
     apt-get install -y sudo docker.io git make btrfs-tools libdevmapper-dev libgpgme-dev libostree-dev
 
-ADD https://storage.googleapis.com/golang/go1.11.1.linux-amd64.tar.gz /tmp
+ADD https://storage.googleapis.com/golang/go1.11.12.linux-amd64.tar.gz /tmp
 
-RUN tar -C /usr/local -xzf /tmp/go1.11.1.linux-amd64.tar.gz && \
-    rm /tmp/go1.11.1.linux-amd64.tar.gz && \
-    ln -s /usr/local/go/bin/* /usr/local/bin/
```
mtrmac: (Nitpick absolutely not worth another re-spin: removing this indentation looks unintentional.)

umohnani8:
@mtrmac do I need to vendor in the package travis is failing on?

mtrmac:
I guess so — or freeze c/storage at the version relevant to CRI-O 1.13 not to pull in updates.

umohnani8:
Travis is still not happy :/

mtrmac:
I guess that’s the other commit from https://github.com/containers/image/pull/670/files … but eventually I’d expect `make test-skopeo` to break because the current version of skopeo will expect current versions of c/image APIs; so, the tests would have to freeze skopeo at a corresponding version, as well.

Or, I guess, just patch `make test-skopeo` out, if the result of the backport will be sufficiently tested in CRI-O.  I don’t expect any other users to use this branch, and this branch is completely new, so if we got the backport wrong (which is in itself pretty unlikely), we can go back and fix this branch without too much trouble.

umohnani8:
@mtrmac tests are green now. I remove `make test-skopeo` from the .travis.yml

PR #701, opened 2019-09-15
make container tools work with go-openssl

lsm5:

@mtrmac @mrunalp @rhatdan PTAL.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -51,3 +51,5 @@ require (
 	gopkg.in/yaml.v2 v2.0.0 // indirect
 	k8s.io/client-go v0.0.0-20170217214107-bcde30fb7eae
 )
+
+replace github.com/docker/libtrust => github.com/containers/libtrust v0.0.0-20190913040956-14b96171aa3b
--- a/go.sum
+++ b/go.sum
@@ -11,6 +11,8 @@ github.com/VividCortex/ewma v1.1.1 h1:MnEK4VOv6n0RSY4vtRe3h11qjxL3+t0B8yOL8iMXdc
 github.com/VividCortex/ewma v1.1.1/go.mod h1:2Tkkvm3sRDVXaiyucHiACn4cqf7DpdyLvmxzcbUokwA=
 github.com/containerd/continuity v0.0.0-20180216233310-d8fb8589b0e8 h1:ZZOFPzvZO3N0f4LIQvZi68F2XDAMl/gqBfFMVjY6B3Y=
 github.com/containerd/continuity v0.0.0-20180216233310-d8fb8589b0e8/go.mod h1:GL3xCUCBDV3CZiTSEKksMWbLE66hEyuu9qyDOOqM47Y=
+github.com/containers/libtrust v0.0.0-20190913040956-14b96171aa3b h1:Q8ePgVfHDplZ7U33NwHZkrVELsZP5fYj9pM5WBZB2GE=
+github.com/containers/libtrust v0.0.0-20190913040956-14b96171aa3b/go.mod h1:9rfv8iPl1ZP7aqh9YA68wnZv2NUDbXdcdPHVz0pFbPY=
 github.com/containers/storage v1.13.2 h1:UXZ0Ckmk6+6+4vj2M2ywruVtH97pnRoAhTG8ctd+yQI=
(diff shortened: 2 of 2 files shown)


lsm5:
hmm, seems like i need to create a module for @containers/libtrust first.

rhatdan:
LGTM

lsm5:
given that https://github.com/containers/skopeo/pull/718 is all green, I think this should be good to merge. @vrothberg ptal when you get the chance.

vrothberg:
Just one last request: can we also add a replace `docker/libtrust` -> `containers/libtrust` to `go.mod`. This way, all consumers of `c/image` will inherit the change. It's just an extra security net to make sure that even transitive dependencies will all use `containers/libtrust`.

Besides that, LGTM :+1: Thanks, @lsm5 !

mtrmac:
> Just one last request: can we also add a replace `docker/libtrust` -> `containers/libtrust` to `go.mod`? This way, all consumers of `c/image` will inherit the change. It's just an extra security net to make sure that even transitive dependencies will all use `containers/libtrust`.

https://github.com/golang/go/wiki/Modules says
> exclude and replace directives only operate on the current (“main”) module. exclude and replace directives in modules other than the main module are ignored when building the main module. The replace and exclude statements, therefore, allow the main module complete control over its own build, without also being subject to complete control by dependencies.

mtrmac:
> given that [containers/skopeo#718](https://github.com/containers/skopeo/pull/718) is all green, I think this should be good to merge. @vrothberg ptal when you get the chance.

AFAICS #718 does not actually run the tests with the code updated by this PR.

vrothberg:
> https://github.com/golang/go/wiki/Modules says
> 
> > exclude and replace directives only operate on the current (“main”) module. exclude and replace directives in modules other than the main module are ignored when building the main module. The replace and exclude statements, therefore, allow the main module complete control over its own build, without also being subject to complete control by dependencies.

Good catch, @mtrmac! I was convinced replacements get inherited - definitely something for the release notes to mention.

mtrmac:
Why does this need a release note? This PR replaces one dependency with another, the callers of c/image only need to update the module and should automatically get the new one; in particular, they shouldn’t need a `replace` directive (at least for c/image; they may want it for other software that uses the original subpackage and also needs to not crash).

vrothberg:
> Why does this need a release note? This PR replaces one dependency with another, the callers of c/image only need to update the module and should automatically get the new one; 

Imagine a transitive dependency on docker/libtrust that it outside of our immediate control. If parts of the code go through docker/libtrust and others go through c/libtrust, users might end up with unexpected surprises. 

> in particular, they shouldn’t need a `replace` directive (at least for c/image; they may want it for other software that uses the original subpackage and also needs to not crash).

Right. The replace for c/image isn't needed. Others _might_ need it which is why I think we should add a note to the release notes.

mtrmac:
> Imagine a transitive dependency on docker/libtrust that it outside of our immediate control. If parts of the code go through docker/libtrust and others go through c/libtrust, users might end up with unexpected surprises.

The library has basically no state to speak of, and c/image uses it in a very stateless manner. I don’t expect any surprises (other than that FIPS breakage, sure).

Of course announcing that the fork is available can be useful; I was just wondering about the connection with a c/image release.

lsm5:
@mrunalp https://github.com/containers/skopeo/pull/728 is all green with the latest c/image master + docker/libtrust -> c/libtrust replacement as @vrothberg suggested. Mind taking a look at the test failures here please, not sure if that's related with the c/libtrust change.

lsm5:
I see the same failures on the merge commit @ https://travis-ci.org/containers/image/jobs/590019172 so hopefully it's not a blocker for this.

mrunalp:
LGTM

mtrmac:
@lsm5 (@mrunalp ) Per https://github.com/containers/image/pull/701#issuecomment-532107501 the `Replace` directive does not affect consumers of the library, and looking at the “downloading” lines in https://travis-ci.org/containers/skopeo/jobs/590039859 , as well as at the contents of the `skopeo/vendor` directory, skopeo continues to use the docker/libtrust library (as will all other consumers of c/image).

So, this still needs fixing, properly.

mtrmac:
> So, this still needs fixing, properly.

#715.

PR #702, opened 2019-09-17
Add AuthEnableKeyring to SystemContext

QiWang19:
Add boolean AuthEnableKeyring to SystemContex to enable podman set auth to keyring.

### DIFF
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -37,8 +37,6 @@ var (
 	dockerHomePath          = filepath.FromSlash(".docker/config.json")
 	dockerLegacyHomePath    = ".dockercfg"
 
-	enableKeyring = false
-
 	// ErrNotLoggedIn is returned for users not logged into a registry
 	// that they are trying to logout of
 	ErrNotLoggedIn = errors.New("not logged in")
@@ -57,7 +55,7 @@ func SetAuthentication(sys *types.SystemContext, registry, username, password st
 		// The keyring might not work in all environments (e.g., missing capability) and isn't supported on all platforms.
 		// Hence, we want to fall-back to using the authfile in case the keyring failed.
 		// However, if the enableKeyring is false, we want adhere to the user specification and not use the keyring.
-		if enableKeyring {
+		if sys != nil && sys.AuthEnableKeyring {
 			err := setAuthToKernelKeyring(registry, username, password)
 			if err == nil {
 				logrus.Debugf("credentials for (%s, %s) were stored in the kernel keyring\n", registry, username)
@@ -81,7 +79,7 @@ func GetAuthentication(sys *types.SystemContext, registry string) (string, strin
 		return sys.DockerAuthConfig.Username, sys.DockerAuthConfig.Password, nil
 	}
 
-	if enableKeyring {
+	if sys != nil && sys.AuthEnableKeyring {
 		username, password, err := getAuthFromKernelKeyring(registry)
 		if err == nil {
 			logrus.Debug("returning credentials from kernel keyring")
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
@mtrmac This allows the callers like podman/buildah/skopeo to specify whether or not they want to use the keyring.
@mtrmac @vrothberg @giuseppe PTAL

TomSweeneyRedHat:
LGTM

mtrmac:
> Otherwise, the common "`podman login` + use some other tool" use case will break if the "other tool" doesn't support key rings yet.

Taking this PR as isolated, to me, this strongly suggests that this should not primarily be a caller option at all, especially if Podman intends to default this to on (which would, via `podman login`, force everyone else to default this to on as well.). And yet, we did add the `enableKeyring` option with the explicit intent for the behavior to be optional.

Is there a consolidated write-up of the design / interactions / defaults / fallbacks, or can anyone remember enough of the old previous discussions to produce one? I do remember https://github.com/containers/image/pull/680 , but that does not directly refer to previous discussions for _that_ design decision, so if no-one remembered, reconstructing this might involve some GitHub archeology.

rhatdan:
I am not sure what you want here the plan is to get `podman login` and `buildah login` to default to kernel keyring.  Both will have override features to fall back to the previous versions.  Something like
podman login --nokeyring or --authfile=PATH

The tools that look up keys should look in kernel keyring first, then in the authorization file paths.

QiWang19:
I have PR in libpod CLI flag `--authfile` https://github.com/containers/libpod/pull/3827 to use/test the change in this patch.

rhatdan:
@mtrmac Ok lets turn on keyring by default, then and have podman opt out.  We had a PR for that but it was dropped.  I want keyring by default and then we can get all of the tools to use it.  There will be a short time where the defaults will not work together, until we get all tools updated.   So we will need to move quickly to get skopeo out first.

mtrmac:
Can we try to plan for what _all_ the pieces of the design are, and merge them at approximately the same time, instead of figuring the design out on-the-go only based on bug reports, after we are already committed to parts of the UI and design changes are difficult, please?

We’ve tried the “just flip a bit, what could go wrong” approach, and it didn’t go spectacularly well.

(Note that none of the non-trivial interactions currently have tests in c/image; `make test-skopeo` also only tests trivial reads of an existing config file. Podman does _at least_ have tests that login+logout work _in the default configuration_, but https://github.com/containers/libpod/pull/3827 1) does not add any extra tests for the complex cases either, and 2) fails for probably unrelated reasons. I can’t see that much reason to be confident about the design as is.)

- What is the default? (Forcing every caller of c/image to hard-code the same defaults would not really be maintainable.)
- Do the users have (explicit-option or implicit) ways to force use of the alternatives?
  - If there are multiple options, how do they interact?
    - Notably the `--authfile` behavior should, probably, live in c/image, not in Podman like https://github.com/containers/libpod/pull/3827 .
- Difficulties with the way we read multiple files, but only write one:
  - Do the credential helpers or the keyring have precedence? Right now `GetAuthentication` and the editor functions are inconsistent about that, causing `SetAuthentication` to, it seems, not edit the right value if both exist.
  - Does `SetAuthentication` preserve the current storage mechanism for a specific record (always/by default), or write to the currently-default mechanism (possibly leaving previous values in other storage mechanisms)? Some other behavior?
  - Can the user create a _persistent_ configuration to always use a credential helper (e.g. the system keyring)? Right now that seems not to be possible in `config.json` (because the first `podman login` creates a file in `/run` that effectively shadows any configuration in ~/.docker/config.json`
- Shouldn’t the keyring be treated just like any other credential helper? Or maybe like a global credential helper (`credsStore` in https://docs.docker.com/engine/reference/commandline/login/ )?

QiWang19:
@rhatdan we already had this PR to enable keyring. And it will fallback to authfile if Get/Set auth to keyring fails

Issue #618, opened 2019-04-24
gracefully handle HTTP 429 return codes (Too Many Requests)


miabbott:
Downstream BZ - https://bugzilla.redhat.com/show_bug.cgi?id=1702519

In the above BZ, the RHCOS nodes are attempting to pull a container from quay.io via `skopeo` and the registry is returning HTTP 429 (Too Many Requests).  `skopeo` reacts with:

`msg="Error determining repository tags: Invalid status code returned when fetching tags list 429 (Too Many Requests)"`

I think the error message is misleading since 429 is a valid code (https://tools.ietf.org/html/rfc6585#section-4), so at a minimum the message should be changed.

cc: @runcom

vrothberg:
Hi @miabbott, thanks for opening the issue. 

> [...], so at a minimum the message should be changed.

How about "Error determining repository tags: too many requests"? Alternatively, we can do a simple "stop DOSing the registry" ;-)

miabbott:
@vrothberg I think that message is a good start; I think it might be useful to include the HTTP code, too.

Do you think there is anything else that can be done in this scenario besides just error more cleanly?

vrothberg:
> Do you think there is anything else that can be done in this scenario besides just error more cleanly?

I don't know the penalty (i.e., how long the client is rejected) but if that's ~predictable we could try to play smart and retry after a certain amount of time. @mtrmac do you know the (default) penalty?

miabbott:
The spec says:

`...MAY include a Retry-After header indicating how long to wait before making a new request.`

So it is unfortunately up to the registry/server if they decide to tell the client how long to back-off.  Maybe try to retry according to `Retry-After` if it is present in the response?

wking:
And probably do sime sort of exponential backoff (60s, doubling the wait between attempts, give up after... 3? attempts) if there is no `Retry-After`.

josephschorr:
We can add a `Retry-After` header too, if that helps

mtrmac:
> > [...], so at a minimum the message should be changed.
> 
> How about "Error determining repository tags: too many requests"? Alternatively, we can do a simple "stop DOSing the registry" ;-)

This is not raw HTTP; the docker/distribution protocol has an integrated error reporting mechanism: https://github.com/docker/distribution/blob/master/docs/spec/api.md#on-failure-too-many-requests-1 .

[...]

PR #703, opened 2019-09-17
docker: handle http 429 status codes

vrothberg:
Consolidate checking the http-status codes to allow for a more uniform
error handling.  Also treat code 429 (too many requests) as a known
error instead of an invalid status code.

When hitting 429, perform an exponential back off starting a 2 seconds
for at most 5 iterations.

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -21,7 +21,7 @@ import (
 	"github.com/containers/image/v4/pkg/sysregistriesv2"
 	"github.com/containers/image/v4/pkg/tlsclientconfig"
 	"github.com/containers/image/v4/types"
-	"github.com/docker/distribution/registry/client"
+	clientLib "github.com/docker/distribution/registry/client"
 	"github.com/docker/go-connections/tlsconfig"
 	digest "github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
@@ -47,14 +47,7 @@ const (
 	extensionSignatureTypeAtomic    = "atomic" // extensionSignature.Type
 )
 
-var (
-	// ErrV1NotSupported is returned when we're trying to talk to a
-	// docker V1 registry.
-	ErrV1NotSupported = errors.New("can't talk to a V1 docker registry")
-	// ErrUnauthorizedForCredentials is returned when the status code returned is 401
-	ErrUnauthorizedForCredentials = errors.New("unable to retrieve auth token: invalid username/password")
-	systemPerHostCertDirPaths     = [2]string{"/etc/containers/certs.d", "/etc/docker/certs.d"}
-)
+var systemPerHostCertDirPaths = [2]string{"/etc/containers/certs.d", "/etc/docker/certs.d"}
 
 // extensionSignature and extensionSignatureList come from github.com/openshift/origin/pkg/dockerregistry/server/signaturedispatcher.go:
 // signature represents a Docker image signature.
@@ -284,14 +277,7 @@ func CheckAuth(ctx context.Context, sys *types.SystemContext, username, password
 	}
 	defer resp.Body.Close()
 
-	switch resp.StatusCode {
-	case http.StatusOK:
-		return nil
-	case http.StatusUnauthorized:
-		return ErrUnauthorizedForCredentials
-	default:
-		return errors.Errorf("error occured with status code %d (%s)", resp.StatusCode, http.StatusText(resp.StatusCode))
-	}
+	return httpResponseToError(resp)
 }
 
 // SearchResult holds the information of each matching image
@@ -365,7 +351,7 @@ func SearchRegistry(ctx context.Context, sys *types.SystemContext, registry, ima
 		} else {
 			defer resp.Body.Close()
 			if resp.StatusCode != http.StatusOK {
-				logrus.Debugf("error getting search results from v1 endpoint %q, status code %d (%s)", registry, resp.StatusCode, http.StatusText(resp.StatusCode))
+				logrus.Debugf("error getting search results from v1 endpoint %q: %v", registry, httpResponseToError(resp))
 			} else {
 				if err := json.NewDecoder(resp.Body).Decode(v1Res); err != nil {
 					return nil, err
@@ -382,7 +368,7 @@ func SearchRegistry(ctx context.Context, sys *types.SystemContext, registry, ima
 	} else {
 		defer resp.Body.Close()
 		if resp.StatusCode != http.StatusOK {
-			logrus.Errorf("error getting search results from v2 endpoint %q, status code %d (%s)", registry, resp.StatusCode, http.StatusText(resp.StatusCode))
+			logrus.Errorf("error getting search results from v2 endpoint %q: %v", registry, httpResponseToError(resp))
 		} else {
 			if err := json.NewDecoder(resp.Body).Decode(v2Res); err != nil {
 				return nil, err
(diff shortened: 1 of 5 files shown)


vrothberg:
Fixes: https://github.com/containers/image/issues/618

docker/docker_client.go
```diff
@@ -43,18 +42,13 @@ const (
 
 	minimumTokenLifetimeSeconds = 60
 
+	retryAfterDelay = 20 // 20 seconds of delay after the request
```
MartinBasti: That's too much. For example rate limiting in quay.io applies on seconds base.

I'd start with 2 seconds?

vrothberg: 2, 4, 8? What should the sequence look like?

docker/docker_client.go
```diff
@@ -419,6 +406,26 @@ func (c *dockerClient) makeRequest(ctx context.Context, method, path string, hea
...
+	var (
+		res   *http.Response
+		err   error
+		delay time.Duration
+	)
+	delay = retryAfterDelay
+	for i := 1; i < 3; i++ {
+		res, err = c.makeRequestToResolvedURLHelper(ctx, method, url, headers, stream, streamLen, auth, extraScope)
+		if res != nil && res.StatusCode == http.StatusTooManyRequests {
+			logrus.Errorf("%s: sleeping for %d seconds before next attempt", http.StatusText(res.StatusCode), delay)
```
MartinBasti: I don't think that this should be ERROR level. INFO or DEBUG should be sufficient.

vrothberg: Good catch. Used that for debugging. Will make it a `Debugf`.

docker/docker_client.go
MartinBasti: 2, 4, 8, 16, 32? 5 retries

MartinBasti: @twaugh WDYT?

twaugh: Yes, if there is no Retry-After supplied in the header, the best behaviour would be to back off exponentially like that. 2, 4, 8, 16, 32, fail, seems about right.

docker/docker_client.go
```diff
@@ -419,6 +406,26 @@ func (c *dockerClient) makeRequest(ctx context.Context, method, path string, hea
...
+			time.Sleep(delay * time.Second)
```
twaugh: When `i` is 2 I think this will be a useless sleep as we won't retry after it.

vrothberg: Good catch!

docker/docker_client.go
```diff
@@ -419,6 +404,30 @@ func (c *dockerClient) makeRequest(ctx context.Context, method, path string, hea
...
+		delay time.Duration
+	)
+	// Do an exponential back off starting at 2 seconds for at most 5
+	// iterations.
+	delay = 2
+	for i := 0; i < 4; i++ {
+		res, err = c.makeRequestToResolvedURLHelper(ctx, method, url, headers, stream, streamLen, auth, extraScope)
+		if res != nil && res.StatusCode == http.StatusTooManyRequests {
+			logrus.Debugf("%s: sleeping for %d seconds before next attempt", http.StatusText(res.StatusCode), delay)
+			if i < 4 {
```
twaugh: I think this wants to be `i < 3`, i.e. 'is this the final loop?'.

twaugh:
>     * What should the maximum delay in seconds be?

Not sure. 120? 60?

>     * Should we attempt exponential back offs for the values from the header as well?

I don't mind. I don't think it's mandatory, but it's possible it will give better behaviour.

docker/docker_client.go
vrothberg: The look should go until `5`. Need to fix the `for` statement.

docker/docker_client.go
```diff
@@ -419,6 +404,30 @@ func (c *dockerClient) makeRequest(ctx context.Context, method, path string, hea
...
+	var (
+		res   *http.Response
+		err   error
+		delay time.Duration
+	)
+	// Do an exponential back off starting at 2 seconds for at most 5
+	// iterations.
+	delay = 2
+	for i := 0; i < 5; i++ {
+		res, err = c.makeRequestToResolvedURLHelper(ctx, method, url, headers, stream, streamLen, auth, extraScope)
```
mtrmac: `stream` can be only consumed once; this fundamentally can’t be done, at least not at this level of abstraction.

docker/docker_client.go
```diff
@@ -627,9 +636,11 @@ func (c *dockerClient) getExtensionsSignatures(ctx context.Context, ref dockerRe
 		return nil, err
 	}
 	defer res.Body.Close()
-	if res.StatusCode != http.StatusOK {
-		return nil, errors.Wrapf(client.HandleErrorResponse(res), "Error downloading signatures for %s in %s", manifestDigest, ref.ref.Name())
```
mtrmac: Dropping `client.HandleErrorResponse(res)` seems rather undesirable.

docker/docker_client.go
mtrmac: Don’t use individual magic numbers in the first place; `numIterations := 5`, and here perhaps check for `numIterations -1` or so.

The logic may well stay the same, but make the relationships obvious, please.

docker/docker_client.go
```diff
@@ -419,6 +404,30 @@ func (c *dockerClient) makeRequest(ctx context.Context, method, path string, hea
...
+				delay *= 2
+			}
+			continue
+		}
+		break
+	}
+	return res, err
+}
+
+func (c *dockerClient) makeRequestToResolvedURLHelper(ctx context.Context, method, url string, headers map[string][]string, stream io.Reader, streamLen int64, auth sendAuth, extraScope *authScope) (*http.Response, error) {
```
mtrmac: `Helper` says nothing, especially without any documentation about its purpose. `…WithoutRetries`, maybe?

docker/docker_client.go
vrothberg: Could you elaborate why? I was found the output of `HandleErrorResponse` not very user friendly and hard to brain parse. Is there some behavior or output you'd like to keep?

vrothberg: In case, I'd switch all checks to that to have ~consistent error messaging.

mtrmac: - It’s the only way to get actual server-produced error messages; dumping the raw HTTP response body would be much worse. Sure, they are often useless, but at least they _can_ be improved.
- The failures are machine-identifiable and converted to specific Go types. See how many different failures in  See `vendor/github.com/docker/distribution/registry/api/v2/errors.go` just map to `http.StatusBadRequest`, and `isManifestInvalidError`.

mtrmac: > In case, I'd switch all checks to that to have ~consistent error messaging.

Sure, as long as the API documents that this format of error reporting is used. That’s not the case for look-aside signatures, at the very least; I haven’t checked the other cases.

docker/docker_client.go
vrothberg: Good point. Looking at the call sites, we should not perform any retry when `stream != nil` which avoids the issue.

docker/docker_client.go
vrothberg: Added a comment and renamed it to `makeRequestToResolvedURLOnce`.

docker/docker_client.go
vrothberg: > In case, I'd switch all checks to that to have ~consistent error messaging.

That won't work for all responses:
```
error parsing HTTP 429 response body: invalid character '<' looking for beginning of value: "<html>\r\n<head><title>429 Too Many Requests</title></head>\r\n<body>\r\n<center><h1>429 Too Many Requests</h1></center>\r\n<hr><center>nginx/1.17.3</center>\r\n</body>\r\n</html>\r\n"
```

vrothberg:
Updated. @mtrmac, mind having another look?

I'll tackle the `Retry-After` handling now.

docker/docker_client.go
mtrmac: Quay.io is known to be non-compliant, what else is new?

vrothberg: That's news for me at least. I'll try with nginx pointed to another registry...

vrothberg:
We're now handling the `Retry-After` header for both, numerical values _and_ dates.
@twaugh @MartinBasti @mtrmac PTAL.

twaugh:
LGTM. Thanks!

docker/docker_client.go
mtrmac: I’m sorry, this was too flippant and not helpful.  If we are centralizing error handling, adding a centralized workaround for Quay (e.g. “body starts with `<html>`”) would of course be an option. (Sure, technically it was an option before as well, but sprinkling that workaround all over the package would be too ugly.)

vrothberg: I left this unchanged. Improving and further centralizing error handling sounds like a good idea but is not the intention of this PR. The new function was added to avoid too much boilerplate.

vrothberg:
Rebased on latest master.

vrothberg:
CI hiccups from Skopeo have been fixed, and we're green again.

@nalind @rhatdan @mrunalp @baude @giuseppe PTAL

rhatdan:
LGTM

docker/errors.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+var (
+	// ErrV1NotSupported is returned when we're trying to talk to a
+	// docker V1 registry.
+	ErrV1NotSupported = errors.New("can't talk to a V1 docker registry")
+	// ErrUnauthorizedForCredentials is returned when the status code returned is 401
+	ErrUnauthorizedForCredentials = errors.New("unable to retrieve auth token: invalid username/password")
+	// ErrTooManyRequests is returned when the status code returned is 429
+	ErrTooManyRequests = errors.New("too many request to registry")
```
nalind: `github.com/pkg/errors.New()` records a stack trace that leads back here.  I'm guessing that the standard library's `errors.New()` would work better, perhaps using `github.com/pkg/errors.WithStack()` to wrap them before returning them in `httpResponseToError()`.

docker/docker_client.go
```diff
@@ -364,8 +350,8 @@ func SearchRegistry(ctx context.Context, sys *types.SystemContext, registry, ima
 			logrus.Debugf("error getting search results from v1 endpoint %q: %v", registry, err)
 		} else {
 			defer resp.Body.Close()
-			if resp.StatusCode != http.StatusOK {
-				logrus.Debugf("error getting search results from v1 endpoint %q, status code %d (%s)", registry, resp.StatusCode, http.StatusText(resp.StatusCode))
+			if resp.StatusCode == http.StatusOK {
```
nalind: Is this (and the similar change in 17 lines below it) reversal doing the right thing?

docker/errors.go
vrothberg: Sounds like a nice improvement!

vrothberg:
Updated

nalind:
LGTM

PR #706, opened 2019-09-18
zstd: support docker v2s2 

vrothberg:
Support zstd for docker v2s2 images.  That's required to copy and
convert a v2s2 image into an OCI images, for instance, via

   `skopeo copy --dest-compress --dest-compress-format=zstd docker://...  oci:...`

When compressing or decompressing a layer, we first update the current
manifest and later convert it.  Hence, supporting v2s2 is currently
required to support compressing and converting in one call.

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -861,7 +861,7 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
 		// re-compressed using the desired format.
 		logrus.Debugf("Blob will be converted")
 
-		compressionOperation = types.PreserveOriginal
+		compressionOperation = types.Compress
 		s, err := decompressor(destStream)
 		if err != nil {
 			return types.BlobInfo{}, err
--- a/image/docker_schema2.go
+++ b/image/docker_schema2.go
@@ -213,10 +213,14 @@ func (m *manifestSchema2) convertToManifestOCI1(ctx context.Context) (types.Imag
 			layers[idx].MediaType = imgspecv1.MediaTypeImageLayerNonDistributable
 		case manifest.DockerV2Schema2ForeignLayerMediaTypeGzip:
 			layers[idx].MediaType = imgspecv1.MediaTypeImageLayerNonDistributableGzip
+		case manifest.DockerV2Schema2ForeignLayerMediaTypeZstd:
+			layers[idx].MediaType = imgspecv1.MediaTypeImageLayerNonDistributableZstd
 		case manifest.DockerV2SchemaLayerMediaTypeUncompressed:
 			layers[idx].MediaType = imgspecv1.MediaTypeImageLayer
 		case manifest.DockerV2Schema2LayerMediaType:
 			layers[idx].MediaType = imgspecv1.MediaTypeImageLayerGzip
+		case manifest.DockerV2Schema2LayerMediaTypeZstd:
+			layers[idx].MediaType = imgspecv1.MediaTypeImageLayerZstd
 		default:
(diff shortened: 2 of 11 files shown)


copy/copy.go
```diff
@@ -861,7 +861,7 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
 		// re-compressed using the desired format.
 		logrus.Debugf("Blob will be converted")
 
-		compressionOperation = types.PreserveOriginal
+		compressionOperation = types.Compress
```
mtrmac: I can’t see how this makes all that much sense; when the semantics of the enum can not represent all necessary operations, my first instinct would be to extend the enum to match the real world, not to double down on the semantic mismatch.

I suppose we should have separate enums for the return value of `DesiredLayerCompression` and for the `BlobInfo.CompressionOperation` values. The two are never compared directly, so we don’t _have_ to have the same for both.

vrothberg:
> The test failures seem to point to a design error: The set of supported encryption algorithms need to be considered _before_ deciding to compress, not only at the point when the blob is already uploaded and manifests need updating.
> 
> (Sure, that error is, mostly, not all that harmful; it results in an orphaned blob in the registry, and the user can retry with a different compression option.)

Yup. This needs some more rework.

copy/copy.go
vrothberg: I agree that there's much space for improving the code but the issue here is that we first update the manifest and then convert it. Since we decided that Docker doesn't support zstd, we fail at updating the manifest although we ultimately want to convert.

vrothberg: It's unfortunate that the Skopeo tests passed without actually checking the manifest :(

vrothberg:
@rhatdan @mtrmac @giuseppe @lsm5 PTAL

I know this is not ideal since we did not plan to "support" zstd for v2s2 images but I don't find any other way to support something similar to `skopeo copy --dest-compress --dest-compress-format=zstd docker://... oci:...` where we change the compression and the image type at the same time. The problem is that we update and image when the compression changes; if we use zstd then docker needs to support it (assuming we keep the current logic).

I considered the idea to have some kind of inter-mediate stage that can be used during the conversion later one but we don't know at all times that we will actually convert. Supporting that seems like a bigger rewrite that I think nobody has time for at the moment.

Maybe I've missed something, so PTAL.

rhatdan:
Can we enforce this only on the source image.  IE You can not create the new s2 images with compression zstd?

vrothberg:
> Can we enforce this only on the source image. IE You can not create the new s2 images with compression zstd?

I don't think so but @mtrmac may know a trick.

copy/copy.go
mtrmac: > I agree that there's much space for improving the code but the issue here is that we first update the manifest and then convert it.

That’s just code structure we can alter.

vrothberg:
For the record: this PR is _not_ needed for Buildah. https://github.com/containers/buildah/issues/1589  is already fixed with the latest c/image. I've used a wrong binary for testing Buildah and followed a red herring.

In other words, we can take our time discussing a proper design to set the correct mime types when recompressing what this PR here attempts to do.

mtrmac:
For the record, from a separate conversation:

I think we should have some kind of a check for valid compression formats before copies start; then the later manifest conversions can ~legitimately be more lenient about unknown MIME types (or maybe we should have an explicit ("layer type", compression not representable) representation? probably not necessary).

WRT the problem of changing the manifest format and compression at the same time, it seems to me that `.UpdatedImage` should, in principle, be able to just do the conversions first, and `UpdatedLayerInfos` afterwards — that would need good test coverage.

vrothberg:
Closing this one, so we can take a fresh approach in the future.

PR #708, opened 2019-09-20
make test-skopeo: replace c/image module instead of copying code

vrothberg:
Instead of copying the code, replace the c/image module in the cloned
branch of Skopeo.  This will allow for injecting and testing changed
dependencies as well and prevent us from running into certain
dependency-induced regressions.


vrothberg:
I expect this PR to fail. Once we figure out which commit introduced the regressions for Skopeo's CI, we can update it here.

vrothberg:
@mtrmac PTAL. This fixes the regressions we faced in Skopeo when vendoring the c/image go module.

I cannot tell _exactly_ which dependency caused the issue or if it was bad combination. That's not very satisfying but at least we're back to a working state. Any future updates will catch regressions early which will hopefully help in understanding which dependencies are responsible.

lsm5:
skopeo with c/libtrust vendored along with this fork of c/image is all green as well. https://github.com/containers/skopeo/pull/728

Makefile
```diff
@@ -76,8 +78,10 @@ test-skopeo:
 		skopeo_path=$${GOPATH}/src/github.com/containers/skopeo && \
 		vendor_path=$${skopeo_path}/vendor/github.com/containers/image && \
 		git clone -b $(SKOPEO_BRANCH) https://github.com/$(SKOPEO_REPO) $${skopeo_path} && \
-		rm -rf $${vendor_path} && cp -r . $${vendor_path} && rm -rf $${vendor_path}/vendor && \
 		cd $${skopeo_path} && \
+		env GO111MODULE="on" \
+		go mod edit -replace github.com/containers/image=$(PROJECT_PATH) && \
```
mtrmac: (Absolutely non-blocking nits:
- The `env` is probably not necessary (at least it’s not used in most of the other `GO111MODULE` invocations)
- The single-use `PROJECT_PATH` variable seems not worth the overhead of introducing an extra name, but opinions on that may well differ.

)

go.mod
```diff
@@ -16,10 +16,10 @@ require (
 	github.com/etcd-io/bbolt v1.3.3
 	github.com/ghodss/yaml v0.0.0-20161207003320-04f313413ffd
 	github.com/gogo/protobuf v0.0.0-20170815085658-fcdc5011193f // indirect
-	github.com/gorilla/context v1.1.1 // indirect
-	github.com/gorilla/mux v0.0.0-20170217192616-94e7d24fd285 // indirect
+	github.com/gorilla/context v0.0.0-20140604161150-14f550f51af5 // indirect
+	github.com/gorilla/mux v0.0.0-20140926153814-e444e69cbd2e // indirect
 	github.com/gotestyourself/gotestyourself v2.2.0+incompatible // indirect
-	github.com/imdario/mergo v0.0.0-20160517064435-50d4dbd4eb0e
+	github.com/imdario/mergo v0.0.0-20141206190957-6633656539c1
```
mtrmac: There are several downgrades here; why are they the right thing to do?

E.g. this one in particular goes over 1 year back; at least some of the others revert the upgrades done in https://github.com/containers/image/pull/682 , which is still a bit suspect but at least obviously motivated, but this import used the 50d4… commit before that PR as well, so this is a straight downgrade.

(I’m not at all saying that the downgrade is incorrect, but it is at least suspect, I can’t take a detailed look _right now_, and there is no explanation in the PR nor the commit message. I’ll try to look into this later myself.)

vrothberg: The explanation is ...

> I cannot tell exactly which dependency caused the issue or if it was a bad combination. That's not very satisfying but at least we're back to a working state. Any future updates will catch regressions early which will hopefully help in understanding which dependencies are responsible.

I don't have much time to look into the details but at least wanted to get us back into a working state.

vrothberg: > That’s fair, it’s just that this is not AFAICS moving us “back” to c/image as it was. Or is it intended to be “back to c/Skopeo as it was”?

The latter after noticing that the previous ones c/image didn't work either. I used the version that c/skopeo was using since those are the ones used in the CI. The `go.mod` change actually caused c/skopeo to inherit the dependencies.

vrothberg: Thanks for checking, @mtrmac !

mrunalp:
I am merging this as it is blocking FIPS work. We can come back to update deps as needed.

mrunalp:
LGTM

go.mod
mtrmac: So, based on tests in https://github.com/containers/image/pull/710/commits , this seems to be caused by an `imdario/mergo` upgrade; there’s already a comment pointing to that behavior change in https://github.com/containers/image/blob/ddb11c057155dceb36e44ca82146b2df93b4dd3f/openshift/openshift-copies.go#L209 .

I appreciate that `openshift-copies.go` is a horrible tangle of abstractions (even after cutting down the original a bit); still…

(In the future, please clearly document both what _and why_ has happened, even if the underlying cause is not known; nothing links the the test failures in #724 , nor records them, to this PR, and reverse-engineering the causes for this downgrade would be pretty difficult in a few weeks.)

PR #709, opened 2019-09-24
add http response to log

QiWang19:
fix https://github.com/containers/libpod/issues/3884
Add http response message log to show server-side error message.

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -533,6 +533,8 @@ func (c *dockerClient) getBearerToken(ctx context.Context, challenge challenge, 
 	defer res.Body.Close()
 	switch res.StatusCode {
 	case http.StatusUnauthorized:
+		err := client.HandleErrorResponse(res)
+		logrus.Debugf("Server response when trying to obtain an access token: \n%q", err.Error())
 		return nil, ErrUnauthorizedForCredentials
 	case http.StatusOK:
 		break


docker/docker_client.go
```diff
@@ -533,6 +533,11 @@ func (c *dockerClient) getBearerToken(ctx context.Context, challenge challenge,
 	defer res.Body.Close()
 	switch res.StatusCode {
 	case http.StatusUnauthorized:
+		resBody, err := ioutil.ReadAll(res.Body)
+		if err != nil {
+			return nil, err
+		}
+		logrus.Debugf("%s", string(resBody))
```
mtrmac: Ideally, this should parse the body to extract the actual error message, if any, without all the cruft; from a quick look it seems that at least docker.io and registry.redhat.io (but not quay.io) use some kind of JSON format.

At the very least, the log should clearly say what this data is (“Server response when trying to obtain an access token: %s”), or something like that, and maybe deal with newlines.

docker/docker_client.go
```diff
@@ -533,6 +533,11 @@ func (c *dockerClient) getBearerToken(ctx context.Context, challenge challenge,
 	defer res.Body.Close()
 	switch res.StatusCode {
 	case http.StatusUnauthorized:
+		resBody, err := ioutil.ReadAll(res.Body)
+		if err != nil {
+			return nil, err
```
mtrmac: Shouldn’t this still return `ErrUnauthorizedForCredentials`? Not sure.

docker/docker_client.go
QiWang19: in that issue, `registry.gitlab.com` was used, from the log I saw different json format. I just update the log print format.

mtrmac: So, looking at https://github.com/docker/distribution/blob/14b96e55d84c367ebab6caf9f0086de0876eec72/registry/client/auth/session.go#L441 , the body should be processed with `client.HandleErrorResponse`, like most registry API accesses. That accounts for both the docker.io and registry.redhat.io JSON responses I’ve seen.

Quay.io does not conform to that expectation, but that’s separately discussed in #703.

docker/docker_client.go
```diff
@@ -533,7 +533,8 @@ func (c *dockerClient) getBearerToken(ctx context.Context, challenge challenge,
 	defer res.Body.Close()
 	switch res.StatusCode {
 	case http.StatusUnauthorized:
-		return nil, ErrUnauthorizedForCredentials
+		err := client.HandleErrorResponse(res)
+		return nil, err
```
mtrmac: This is not implausible, but both Buildah and Podman handle `ErrUnauthorizedForCredentials` specially. What about that?

Either `ErrUnauthorizedForCredentials` should be replaced/extended with some type that can preserve the server-provided status (and then all users need to be updated, for an API break), _and_ the other user of `ErrUnauthorizedForCredentials` must be updated a well — which is the long-term right thing to do;

Or do the small non-fix discussed in https://github.com/containers/libpod/issues/3884 , decode the server output, log it to `logrus`, but keep returning `ErrUnauthorizedForCredentials` to callers.

QiWang19: how to update the usage of ErrUnauthorizedForCredentials? can I replace the  ErrUnauthorizedForCredentials with actual server side error and remove the check  `case docker.ErrUnauthorizedForCredentials:` from Podman and Buildah?

mtrmac: To an extent, that depends on what Podman/Buildah want to do.  Right now, both use the `ErrUnauthorizedForCredentials` value to provide a “nice” error message, but throw away the underlying cause entirely.

It’s plausible that they could just report the original server-side value while removing that extra check, but that also depends on that value being nice enough for _most_ registries. 
Alternatively, we could replace the `ErrUnauthorizedForCredentials` error _value_ with an error _type_ that can still carry the underlying error returned by `client.HandleErrorResponse`. Then Podman/Buildah would still need updating, but they could still choose to use their own error texts.

I suppose we should collect sample responses from the major registries, and decide whether they are good enough to present without extra context.

docker/docker_client.go
```diff
@@ -533,6 +533,8 @@ func (c *dockerClient) getBearerToken(ctx context.Context, challenge challenge,
 	defer res.Body.Close()
 	switch res.StatusCode {
 	case http.StatusUnauthorized:
+		err := client.HandleErrorResponse(res)
+		logrus.Errorf("Server response when trying to obtain an access token: \n%q", err)
```
mtrmac: … wait, `Errorf`? https://github.com/containers/libpod/issues/3884 talks about debug level.

It wouldn’t be the _first_ time the library calls `logrus.Errorf`, but it’s only the third, and it can be pretty surprising for callers who don’t expect any output.

I don’t feel _too_ strongly about this, if there is consensus, but I do wonder whether it was intentional or maybe just an artifact from testing.

QiWang19: sorry, changed to debug level

docker/docker_client.go
rhatdan: Could we just wrap the error and continue to return the original errors.  Then Podman/Buildah could use errors.Cause and still get the detailed information.

mtrmac: Not directly, because `ErrUnauthorizedForCredentials` is an error _value_; replacing it with an `errors.Wrap` would be a different value and an API break.

(Sure, we _can_ consider such an API break.)

rhatdan: Let's break the API, for this, since this is useful information for the tools to report.

docker/docker_client.go
```diff
@@ -533,6 +533,8 @@ func (c *dockerClient) getBearerToken(ctx context.Context, challenge challenge,
 	defer res.Body.Close()
 	switch res.StatusCode {
 	case http.StatusUnauthorized:
+		err := client.HandleErrorResponse(res)
+		logrus.Debugf("Server response when trying to obtain an access token: \n%q", err.Error())
 		return nil, ErrUnauthorizedForCredentials
```
QiWang19: what does break API mean?
```
return nil, errors.Wrapf(err, "Server response when trying to obtain an access token")
```

rhatdan: In current version Buildah and Podman can just look at the return code and compare it to ErrUnauthorizedForCredentials
If we wrap the error with the message, then podman and buildah will no longer get ErrUnauthorizedForCredentials back, they would have to take the error returned and change it to 
if errors.Cause(err) == ErrUnauthorizedForCredentials

Any time the users of a library have to change their code, then you have broken ABI, if the callers do not have to change their code, then ABI remained the same. 

https://en.wikipedia.org/wiki/Application_binary_interface

QiWang19: Ok, Thanks, PTAL https://github.com/containers/image/pull/719

docker/docker_client.go
mtrmac: Would have been nice delay the v4.0.0 for that :)

PR #711, opened 2019-09-26
Update test certificates

nalind:
The unit tests in `oci/layout` use prebuilt certificates to run a test https server, but they've expired, so CI tests are failing at that step.  Run their respective `gencert.sh` scripts to generate new ones and commit the result.


nalind:
Rebased.

mrunalp:
LGTM

PR #712, opened 2019-09-26
[DO NOT MERGE] check if tests pass without openssl

lsm5:


lsm5:
so I created this PR with docker/libtrust replaced by lsm5/libtrust commit a6ab7ae https://github.com/lsm5/libtrust/commit/a6ab7aebae83eb02114bb77385f7054c8f856526 which removes the openssl stuff, still the same failures, also seen for the most recent merge commit https://travis-ci.org/containers/image/jobs/590019172

mtrmac:
@lsm5 Assuming this is no longer needed, please re-open if that’s not the case.

PR #713, opened 2019-09-26
[DO NOT MERGE] README change test

lsm5:


mtrmac:
@lsm5 Assuming this is no longer needed, please re-open and mention the motivation if that’s not the case.

PR #714, opened 2019-09-27
Revert the downgrade from #708, and upgrade Mergo

mtrmac:
This:
- Adds some smoke tests for OpenShift (`atomic:`) configuration parsing.

  (Given how deprecated this is, I won’t bother with any non-default paths. Also, the test fixture currently contains invalid certificate/key data to make the current tests nicer to read, but that also prevents testing `restClientFor` and its callers using that fixture.)
- Based on the confidence provided by these tests, and K8s original  https://github.com/kubernetes/kubernetes/pull/65779 , updates Mergo to v0.3.5 and adjusts the callers
- Reverts the dependency downgrades from #708 

@vrothberg @lsm5 PTAL and verify that this fixes the failures you’ve been seeing, I’m only guessing what is in scope based on test failures in  #708 and containers/skopeo#724; nothing explicitly says what the concern is.

(For the record, the test failures I’ve been fixing show up in `test-skopeo` as variants of
```
Error determining manifest MIME type for atomic:localhost:5000/…: Get http://localhost:8080/…: dial tcp 127.0.0.1:8080: connect: connection refused
```
where the `localhost:8080` part does not correspond to the actual installed cluster, but to hard-coded defaults in `openshift-copies.go:defaultCluster`.)


mrunalp:
LGTM

rhatdan:
LGTM

PR #715, opened 2019-09-27
Explicitly replace docker/libtrust with containers/libtrust

mtrmac:
<s>**Do not merge:** Includes unmerged #714.</s>

... instead of using the `replace` directive in `go.mod`, which changes absolutely nothing for consumers of this library.

Cleans up after #701.

(`docker/libtrust` remains listed in `go.mod` because it is a dependency of the `docker/distribution` _module_; it’s not used by the `docker/distribution` _subpackages_ we call. See https://github.com/golang/go/wiki/Modules#why-does-go-mod-tidy-record-indirect-and-test-dependencies-in-my-gomod .)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,12 +7,13 @@ require (
 	github.com/BurntSushi/toml v0.3.1
 	github.com/VividCortex/ewma v1.1.1 // indirect
 	github.com/containerd/continuity v0.0.0-20180216233310-d8fb8589b0e8 // indirect
+	github.com/containers/libtrust v0.0.0-20190913040956-14b96171aa3b
 	github.com/containers/storage v1.13.4
 	github.com/docker/distribution v0.0.0-20170817175659-5f6282db7d65
 	github.com/docker/docker v0.0.0-20180522102801-da99009bbb11
 	github.com/docker/docker-credential-helpers v0.6.0
 	github.com/docker/go-connections v0.0.0-20180212134524-7beb39f0b969
-	github.com/docker/libtrust v0.0.0-20160708172513-aabc10ec26b7
+	github.com/docker/libtrust v0.0.0-20160708172513-aabc10ec26b7 // indirect
 	github.com/etcd-io/bbolt v1.3.3
 	github.com/ghodss/yaml v0.0.0-20161207003320-04f313413ffd
 	github.com/gogo/protobuf v0.0.0-20170815085658-fcdc5011193f // indirect
@@ -47,5 +48,3 @@ require (
 	gopkg.in/check.v1 v1.0.0-20190902080502-41f04d3bba15 // indirect
 	k8s.io/client-go v0.0.0-20170217214107-bcde30fb7eae
 )
-
-replace github.com/docker/libtrust => github.com/containers/libtrust v0.0.0-20190913040956-14b96171aa3b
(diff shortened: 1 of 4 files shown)


rhatdan:
LGTM
@vrothberg Lets get this merged as soon as tests pass.  Then cut v4.0.0 release.

rhatdan:
@lsm5 FYI

mtrmac:
> @vrothberg Lets get this merged as soon as tests pass. Then cut v4.0.0 release.

Note that @lsm5 was reporting problems when vendoring into Podman:
```
Error: Error copying image to the remote destination: Error creating an updated image manifest: Error preparing updated manifest: unsupported media type for compression: "application/vnd.docker.image.rootfs.diff.tar.gzip"
```

Wildly guessing that this is Buildah doing its own thing for manifests and compression, but I haven’t actually checked. That may be not quite trivial to untangle.

lsm5:
> **Do not merge:** Includes unmerged #714.
> 
> ... instead of using the `replace` directive in `go.mod`, which changes absolutely nothing for consumers of this library.
> 
> Cleans up after #701.
> 
> (`docker/libtrust` remains listed in `go.mod` because it is a dependency of the `docker/distribution` _module_; it’s not used by the `docker/distribution` _subpackages_ we call. See https://github.com/golang/go/wiki/Modules#why-does-go-mod-tidy-record-indirect-and-test-dependencies-in-my-gomod .)

ack, though for my own understanding, don't think what you're saying lines up with @vrothberg's [comment here](https://github.com/containers/image/pull/701#issuecomment-532106084), correct me if I'm wrong.

vrothberg:
Yes, I had no problem when using `go get c/image@master` and running the affected test.

lsm5:
> Yes, I had no problem when using `go get c/image@master` and running the affected test.

ack, thanks.

mtrmac:
I don’t know how that happened, yet:
- https://github.com/containers/skopeo/pull/718, the test PR for #701, shows a _different_ c/image change than #701 (the Skopeo PR was actually testing what _this_ c/image PR does, not what #701 (now) does).
- The test log of #701 shows `make test-skopeo` using docker/libtrust, not c/libtrust.

> Yes, I had no problem when using `go get c/image@master` and running the affected test.

Right now, when I do either of
```
GO111MODULE=on go mod edit -replace github.com/containers/image=github.com/containers/image@9911c94ebb54856ef501311ee088
GO111MODULE=on go get github.com/containers/image@9911c94ebb54856ef501311ee08822e796dccb0322e796dccb03 
```
in the `skopeo`, both `make vendor` and `go mod graph` show that skopeo is using docker/libtrust. So, I don’t know what is going on. Is that not representative?

vrothberg:
Trying to reproduce now.

vrothberg:
@mtrmac, I think something went south in the vendoring process. Maybe the `replace c/image=c/image@...` (trying to replace itself)?

I did as follows:
```bash
$ go mod edit -replace github.com/containers/image=github.com/mtrmac/image@containers-libtrust
$ make vendor
$ stat vendor/github.com/docker/libtrust/
stat: cannot stat 'vendor/github.com/docker/libtrust/': No such file or directory
```

... everything's in c/libtrust now.

mtrmac:
Why is that wrong? This PR replaces the dependency on docker/libtrust with a dependency on c/libtrust, nothing else in the vendored subpackages requires docker/libtrust, so `make vendor` removes docker/libtrust from the vendored tree. You should see, in addition to that removal, also addition of `vendor/github.com/containers/ilbtrust`.

rhatdan:
What the heck is wrong with Travis?  Why is this taking so long to test?

rhatdan:
Ok this finally passed, merging.

vrothberg:
> What the heck is wrong with Travis? Why is this taking so long to test?

That's the issue I've raised a few times. High(er) traffic on c/image will slow down Travis and sometimes it takes hours until a job starts. It happens at Buildah as well. It would be great if we could migrate to Cirrus eventually.

vrothberg:
> Why is that wrong? This PR replaces the dependency on docker/libtrust with a dependency on c/libtrust, nothing else in the vendored subpackages requires docker/libtrust, so `make vendor` removes docker/libtrust from the vendored tree. You should see, in addition to that removal, also addition of `vendor/github.com/containers/libtrust`.

```
go mod edit -replace github.com/containers/image=github.com/containers/image@9911c94ebb54856ef501311ee088
```
^ this replace c/image with c/image. It's just a guess that this might confuse go.

mtrmac:
> > What the heck is wrong with Travis? Why is this taking so long to test?
> 
> That's the issue I've raised a few times. High(er) traffic on c/image will slow down Travis and sometimes it takes hours until a job starts.

I’ve committed a bunch of rebases on Friday, that might have triggered rate limiting I guess, but for it to carry on through Monday is surprising.

PR #659, opened 2019-07-18
Release v2.0.1

vrothberg:
The main motivation is to propagate the latest progress-bar fixes (i.e., using spinners for unknown blob sizes) and to ultimately close https://bugzilla.redhat.com/show_bug.cgi?id=1695236 once it's in Fedora.

@mtrmac @rhatdan PTAL

Issue #692, opened 2019-08-27
Importing c/image/types drags in all the compression algorithm implementations


mtrmac:
`c/image/types` imports `pkg/compression` for `compression.Algorithm`, and `pkg/compression` depends on all the compression implementations.

This makes it unnecessarily costly to import pretty much any small part of c/image ; the `compression.Algorithm` type should be split from the implementation (moving it to `c/image/types`? `pkg/compression/types`? not sure yet).

mtrmac:
This was fixed by #716 .

PR #716, opened 2019-09-27
Don’t drag in compression implementations into c/image/types

mtrmac:
<s>**Don’t merge:** Includes all of #714.</s>

Fixes #659, see the commit message as for how.

@vrothberg @giuseppe RFC. This is a bit baroque, but it is a local thing that does not require any discussion about how to redesign the interfaces in some other way; AFAICT all existing public indentifiers continue to work with the existing semantics.

We can redesign pkg/compression in some other way in some future major version, if there is ever a good enough reason, and we obviously can completely tear down what this PR does at that time.

### DIFF
--- a/pkg/compression/internal/types.go
+++ b/pkg/compression/internal/types.go
@@ -1,0 +1,57 @@
+package internal
+
+import "io"
+
+// CompressorFunc writes the compressed stream to the given writer using the specified compression level.
+// The caller must call Close() on the stream (even if the input stream does not need closing!).
+type CompressorFunc func(io.Writer, *int) (io.WriteCloser, error)
+
+// DecompressorFunc returns the decompressed stream, given a compressed stream.
+// The caller must call Close() on the decompressed stream (even if the compressed input stream does not need closing!).
+type DecompressorFunc func(io.Reader) (io.ReadCloser, error)
+
+// Algorithm is a compression algorithm that can be used for CompressStream.
+type Algorithm struct {
+	name         string
+	prefix       []byte
+	decompressor DecompressorFunc
+	compressor   CompressorFunc
+}
+
+// NewAlgorithm creates an Algorithm instance.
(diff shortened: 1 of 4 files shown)


mtrmac:
Any suggestions how to simplify this (without turning this PR into an unbounded design discussion?) are very welcome

vrothberg:
Great work tracking down the root cause of the CI flakes, @mtrmac! I skimmed the changes and they look reasonable to me. I'll have a close look on Monday morning.

vrothberg:
> Any suggestions how to simplify this (without turning this PR into an unbounded design discussion?) are very welcome

I did not find any other way than what the PR does. Good catch!

pkg/compression/internal/types.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+// AlgorithmDecompressor returns the decompressor field of algo.
+// This is a function instead of a public method so that it is only callable from by code
+// that is allowed to import this internal subpackage.
+func AlgorithmDecompressor(algo Algorithm) DecompressorFunc {
+	return algo.decompressor
+}
+
+// AlgorithmPrefix returns the prefix field of algo.// This is a function instead of a public method so that it is only callable from by code
```
TomSweeneyRedHat: nit, did you mean to have a carriage return before the second `//` in this line?

TomSweeneyRedHat:
LGTM assuming happy tests.  Looks like you need to rebase.

pkg/compression/internal/types.go
mtrmac: Yes, thanks!

rhatdan:
Is travis dead?

rhatdan:
Ok this finally passed Merging.

mtrmac:
> Fixes #659, see the commit message as for how.

https://github.com/containers/image/issues/692 actually, for the record.

PR #719, opened 2019-10-01
return resp error message

QiWang19:
follow https://github.com/containers/image/pull/709
return and wrap http response message to show server-side error message.

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -591,7 +591,7 @@ func (c *dockerClient) getBearerToken(ctx context.Context, challenge challenge, 
 	case http.StatusUnauthorized:
 		err := clientLib.HandleErrorResponse(res)
 		logrus.Debugf("Server response when trying to obtain an access token: \n%q", err.Error())
-		return nil, ErrUnauthorizedForCredentials
+		return nil, ErrUnauthorizedForCredentials{Err: err}
 	case http.StatusOK:
 		break
 	default:
--- a/docker/errors.go
+++ b/docker/errors.go
@@ -2,8 +2,10 @@ package docker
 
 import (
 	"errors"
+	"fmt"
 	"net/http"
 
+	"github.com/docker/distribution/registry/client"
 	perrors "github.com/pkg/errors"
 )
 
@@ -11,11 +13,18 @@ var (
 	// ErrV1NotSupported is returned when we're trying to talk to a
 	// docker V1 registry.
 	ErrV1NotSupported = errors.New("can't talk to a V1 docker registry")
-	// ErrUnauthorizedForCredentials is returned when the status code returned is 401
-	ErrUnauthorizedForCredentials = errors.New("unable to retrieve auth token: invalid username/password")
 	// ErrTooManyRequests is returned when the status code returned is 429
 	ErrTooManyRequests = errors.New("too many request to registry")
 )
+
+// ErrUnauthorizedForCredentials is returned when the status code returned is 401
+type ErrUnauthorizedForCredentials struct { // We only use a struct to allow a type assertion, without limiting the contents of the error otherwise.
+	Err error
+}
+
+func (e ErrUnauthorizedForCredentials) Error() string {
+	return fmt.Sprintf("unable to retrieve auth token: invalid username/password: %s", e.Err.Error())
+}
 
 // httpResponseToError translates the https.Response into an error. It returns
 // nil if the response is not considered an error.
@@ -26,7 +35,8 @@ func httpResponseToError(res *http.Response) error {
 	case http.StatusTooManyRequests:
 		return ErrTooManyRequests
 	case http.StatusUnauthorized:
-		return ErrUnauthorizedForCredentials
+		err := client.HandleErrorResponse(res)
+		return ErrUnauthorizedForCredentials{Err: err}
 	default:
 		return perrors.Errorf("invalid status code from registry %d (%s)", res.StatusCode, http.StatusText(res.StatusCode))
 	}


QiWang19:
wrap the server-side message with ErrUnauthorizedForCredentials. even when the error was not caused by ErrUnauthorizedForCredentials = errors.New("unable to retrieve auth token: invalid username/password")?
do I need to change the definition of ErrUnauthorizedForCredentials?

QiWang19:
@rhatdan @mtrmac PTAL

docker/docker_client.go
```diff
@@ -535,7 +535,7 @@ func (c *dockerClient) getBearerToken(ctx context.Context, challenge challenge,
 	case http.StatusUnauthorized:
 		err := client.HandleErrorResponse(res)
 		logrus.Debugf("Server response when trying to obtain an access token: \n%q", err.Error())
-		return nil, ErrUnauthorizedForCredentials
+		return nil, errors.Wrapf(ErrUnauthorizedForCredentials, err.Error())
```
mtrmac: This is not really how `Wrapf` is supposed to be used: `Wrapf` only adds a string.

I suppose we want _both_:
- A well-typed `ErrUnauthorizedForCredentials` that can be recognized by callers
- Preserving the original typed error value returned by `client.HandleErrorResponse`, not just its string representation.

The way `types.ManifestTypeRejectedError` is defined seems like the simplest way to do it.

QiWang19:
Thanks @mtrmac, follow the types.ManifestTypeRejectedError. not sure if it's correct. PTAL
Do I still need to wrap the ErrUnauthorizedForCredentials  https://github.com/containers/image/pull/709#discussion_r329353908?

docker/docker_client.go
```diff
@@ -271,12 +278,12 @@ func newDockerClient(sys *types.SystemContext, registry, reference string) (*doc
 // CheckAuth validates the credentials by attempting to log into the registry
 // returns an error if an error occurred while making the http request or the status code received was 401
 func CheckAuth(ctx context.Context, sys *types.SystemContext, username, password, registry string) error {
-	client, err := newDockerClient(sys, registry, registry)
+	dockerclient, err := newDockerClient(sys, registry, registry)
```
mtrmac: `dockerClient`, please.

docker/docker_client.go
```diff
@@ -50,12 +50,19 @@ const (
...
+	ErrV1NotSupported         = errors.New("can't talk to a V1 docker registry")
+	systemPerHostCertDirPaths = [2]string{"/etc/containers/certs.d", "/etc/docker/certs.d"}
 )
 
+// ErrUnauthorizedForCredentials is returned when the status code returned is 401
+type ErrUnauthorizedForCredentials struct { // We only use a struct to allow a type assertion, without limiting the contents of the error otherwise.
+	Err error
+}
+
+func (e ErrUnauthorizedForCredentials) Error() string {
```
mtrmac: This works, but we might prefer to return only the previous hard-coded string, or an `errors.Wrapf`-like text (“invalid username/password: %s"; “invalid username/password, server reported: %s") — or something similar that works well for typical server-returned error values, including for Quay.

Ideally, the callers should not _have_ to reformat the error string at all.

QiWang19: can I use this?
```go
func (e ErrUnauthorizedForCredentials) Error() string {
	return errors.Wrapf(e.Err, "unable to retrieve auth token: invalid username/password").Error()
}
```

QiWang19:
PTAL @mtrmac

docker/docker_client.go
mtrmac: That’s a very expensive way to say
```go
return fmt.Sprintf("unable to retrieve auth token: invalid username/password: %s", e.Err.Error())
```
Not really blocking, but I don’t see much reason to drag in `Wrapf` into it.

QiWang19:
PTAL @mtrmac

vrothberg:
@QiWang19, needs a rebase.

> LGTM. Now, how shall we handle the API break?

We had a very quick chat about that on Slack yesterday.

1. One idea was to batch such PRs and merge them when cutting a new major release. It would ultimately solve the current problem of having an API break just two days after the last major bump. I am not sure though how we could batch them. If it implies keeping the PRs open then I strongly object the idea as it is not very friendly for community contributions - PRs need to be rebased, merge conflicts can come up etc. and some folks might not be reachable when we want to merge. If we find a community-friendly way, I am cool with it but it would mean more work for us - one way or another (see below).

1. Another idea was to create release branches. This way, master can continue running and we had to cherry-pick commits to the release branch. It would solve the problem of having to release API breaking commits along with bug fixes. BUT it increases the work for the c/image release management. On the other hand, having release branches could help projects with stable and long-term releases such as CRI-O and libpod.

1. We could also just release as soon as we break the API. Looking at the release dates of 2.0.0, 3.0.0, 4.0.0 we have a major bump every ~2 month. Having a break after just two days is certainly unfortunate. 

I can't find a nice process to implement 1). 2) is my favorite as it solves multiple problems at once (i.e., proper release engineering, branches for major version to easily backport fixes for CRI-O/libpod etc.). If there's a majority for 3), I would not block it.

Thanks for kicking off the conversation, @mtrmac !

@rhatdan @giuseppe @nalind PTAL. Note that we will have a very similar problem with c/storage as soon as we hit 2.0.0. It would be ideal to have the same release process/engineering for both projects.

vrothberg:
@mrunalp FYI

rhatdan:
I would go with `2` as I don't really see their being the need to back port many fixes, except for CRI-O and we re doing that now, I believe.

mtrmac:
It’s worse than all that, because if we branch the API, we have to immediately branch Skopeo as well in order to keep the integration tests working. (Sure, one more good reason to move the tests.)

Really the difference between 2/3 is only in the amount/frequency/proportion of fixes that we expect to commit to both branches. We can always create branches on-demand, so the difference between 2/3 is not really in whether the branch exists first, but in whether we would only backport very rare critical fixes (once a year, say), or whether 90% of PRs would be expected to land on both the current and next major branch.

mtrmac:
To some extent, the answer must be to just not break the APIs. That is actually an option in various cases, but it’s also frequently pretty costly, both for this repo and for consumers.

(E.g., in this case, callers could opt in via `types.SystemContext` to receive different types of errors, and old callers could continue to receive the old ones. I’m not _seriously_ suggesting that, at least not at this point in the discussion.)

But a fair amount of the current code was created with very little regard to those processes, and we export too much in various places.

rhatdan:
@QiWang19 Lets rebase and get it merged since we are about to rebase the API and Major version
@giuseppe @mtrmac FYI

rhatdan:
LGTM

PR #721, opened 2019-10-01
Correctly use a c/image/v4 module namespace

mtrmac:
<s>**\<blink>DO NOT MERGE THIS until this has been proven to work as expected in all consumers\</blink>**</s>

... so that major-version-aware Go module import (as opposed to vX.Y.Z+incompatible, which does not allow different packages to use different versions) works right.


vrothberg:
If go down this path, we force all consumers of c/image to use go modules (we had a similar issues with the `mpb` library). It looks good for all projects under our direct influence. All have migrated to go modules and CRI-O v1.14 uses https://github.com/containers/image/tree/cri-o-release-1.14 which is still using the old vendoring model.

I'm supportive of doing the move. Go 1.13 has been released over a month ago and modules are now the defacto standard for dependency management, so it makes sense to me to apply to their rules.

vrothberg:
Snap. After catching up on the other PRs and having re-read the go wiki, this PR seems to be the only way forward:

`If the module is version v2 or higher, the major version of the module must be included as a /vN at the end of the module paths used in go.mod files (e.g., module github.com/my/mod/v2, require github.com/my/mod/v2 v2.0.1) and in the package import path (e.g., import "github.com/my/mod/v2/mypkg"). `

The `+incompatible` trick only seems to work for non-module dependencies with version >= v2.0.0.

vrothberg:
Opened https://github.com/containers/skopeo/pull/733 to test this PR just to be sure and having exercised the migration to v4 on my own.

rhatdan:
LGTM

rhatdan:
Do we need to update skopeo for this to get the tests to pass, or do we need to just merge this with the failing tests and then get it vendored into Skopeo?

mtrmac:
Yes, we need to update skopeo (per Valentin’s https://github.com/containers/skopeo/pull/733 ), but given how disruptive it is, I’d really prefer to see the _whole stack_ working before merging this.

— and an _explicit_ consensus that we are fine with burdening all consumers with the need to keep updating `image/v`_Number_ . We still have some chance to just revert to a non-module library. (I don’t quite think reverting is the best choice, but it is _an option_ and the downsides with this PR are non-trivial. Once we merge this, going back would be more painful.)

mtrmac:
Actually these test failures (`recipe for target 'build' failed`) are local, I’m afraid I must have forgotten to run the local tests on the final result. Fixing…. Still, Skopeo will need updating.

mtrmac:
`make test-skopeo` fails here, tests pass in https://github.com/containers/skopeo/pull/736 .

PR #722, opened 2019-10-02
allow for .dockercfg files to reside in non-home directories (facilit…

gabemontero:
…ate openshift secret mount)

From work with @mtrmac in https://github.com/openshift/builder/pull/102 where we are trying to satisfy docker authentication in an mirrored registry scenario as part of the openshift disconnected support that just shipped in 4.2

openshift can mount `.dockercfg` files via secrets in non-home dir locations

### DIFF
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -26,6 +26,11 @@ type dockerConfigFile struct {
 	CredHelpers map[string]string           `json:"credHelpers,omitempty"`
 }
 
+type authPath struct {
+	path         string
+	legacyFormat bool
+}
+
 var (
 	defaultPerUIDPathFormat = filepath.FromSlash("/run/containers/%d/auth.json")
 	xdgRuntimeDirPath       = filepath.FromSlash("containers/auth.json")
@@ -84,28 +89,28 @@ func GetAuthentication(sys *types.SystemContext, registry string) (string, strin
 		}
 	}
 
-	dockerLegacyPath := filepath.Join(homedir.Get(), dockerLegacyHomePath)
-	var paths []string
-	pathToAuth, err := getPathToAuth(sys)
+	paths := []authPath{}
+	pathToAuth, lf, err := getPathToAuth(sys)
 	if err == nil {
-		paths = append(paths, pathToAuth)
+		paths = append(paths, authPath{path: pathToAuth, legacyFormat: lf})
 	} else {
 		// Error means that the path set for XDG_RUNTIME_DIR does not exist
 		// but we don't want to completely fail in the case that the user is pulling a public image
 		// Logging the error as a warning instead and moving on to pulling the image
 		logrus.Warnf("%v: Trying to pull image in the event that it is a public image.", err)
 	}
-	paths = append(paths, filepath.Join(homedir.Get(), dockerHomePath), dockerLegacyPath)
+	paths = append(paths,
+		authPath{path: filepath.Join(homedir.Get(), dockerHomePath), legacyFormat: false},
+		authPath{path: filepath.Join(homedir.Get(), dockerLegacyHomePath), legacyFormat: true})
 
 	for _, path := range paths {
-		legacyFormat := path == dockerLegacyPath
-		username, password, err := findAuthentication(registry, path, legacyFormat)
+		username, password, err := findAuthentication(registry, path.path, path.legacyFormat)
 		if err != nil {
 			logrus.Debugf("Credentials not found")
 			return "", "", err
 		}
 		if username != "" && password != "" {
-			logrus.Debugf("Returning credentials from %s", path)
+			logrus.Debugf("Returning credentials from %s", path.path)
 			return username, password, nil
 		}
 	}
@@ -163,13 +168,16 @@ func RemoveAllAuthentication(sys *types.SystemContext) error {
 // The path can be overriden by the user if the overwrite-path flag is set
 // If the flag is not set and XDG_RUNTIME_DIR is set, the auth.json file is saved in XDG_RUNTIME_DIR/containers
 // Otherwise, the auth.json file is stored in /run/containers/UID
-func getPathToAuth(sys *types.SystemContext) (string, error) {
+func getPathToAuth(sys *types.SystemContext) (string, bool, error) {
 	if sys != nil {
 		if sys.AuthFilePath != "" {
-			return sys.AuthFilePath, nil
+			return sys.AuthFilePath, false, nil
+		}
+		if sys.LegacyFormatAuthFilePath != "" {
+			return sys.LegacyFormatAuthFilePath, true, nil
 		}
 		if sys.RootForImplicitAbsolutePaths != "" {
-			return filepath.Join(sys.RootForImplicitAbsolutePaths, fmt.Sprintf(defaultPerUIDPathFormat, os.Getuid())), nil
+			return filepath.Join(sys.RootForImplicitAbsolutePaths, fmt.Sprintf(defaultPerUIDPathFormat, os.Getuid())), false, nil
 		}
 	}
 
@@ -182,11 +190,11 @@ func getPathToAuth(sys *types.SystemContext) (string, error) {
 			// This means the user set the XDG_RUNTIME_DIR variable and either forgot to create the directory
 			// or made a typo while setting the environment variable,
 			// so return an error referring to $XDG_RUNTIME_DIR instead of xdgRuntimeDirPath inside.
-			return "", errors.Wrapf(err, "%q directory set by $XDG_RUNTIME_DIR does not exist. Either create the directory or unset $XDG_RUNTIME_DIR.", runtimeDir)
+			return "", false, errors.Wrapf(err, "%q directory set by $XDG_RUNTIME_DIR does not exist. Either create the directory or unset $XDG_RUNTIME_DIR.", runtimeDir)
 		} // else ignore err and let the caller fail accessing xdgRuntimeDirPath.
-		return filepath.Join(runtimeDir, xdgRuntimeDirPath), nil
+		return filepath.Join(runtimeDir, xdgRuntimeDirPath), false, nil
 	}
-	return fmt.Sprintf(defaultPerUIDPathFormat, os.Getuid()), nil
+	return fmt.Sprintf(defaultPerUIDPathFormat, os.Getuid()), false, nil
 }
 
 // readJSONFile unmarshals the authentications stored in the auth.json file and returns it
@@ -220,7 +228,7 @@ func readJSONFile(path string, legacyFormat bool) (dockerConfigFile, error) {
 
 // modifyJSON writes to auth.json if the dockerConfigFile has been updated
 func modifyJSON(sys *types.SystemContext, editor func(auths *dockerConfigFile) (bool, error)) error {
-	path, err := getPathToAuth(sys)
+	path, legacyFormat, err := getPathToAuth(sys)
 	if err != nil {
 		return err
 	}
@@ -232,6 +240,9 @@ func modifyJSON(sys *types.SystemContext, editor func(auths *dockerConfigFile) (
 		}
 	}
 
+	if legacyFormat {
+		return fmt.Errorf("writes to %s using legacy format are not supported", path)
+	}
 	auths, err := readJSONFile(path, false)
 	if err != nil {
 		return errors.Wrapf(err, "error reading JSON file %q", path)
--- a/pkg/docker/config/config_test.go
+++ b/pkg/docker/config/config_test.go
@@ -33,31 +33,34 @@ func TestGetPathToAuth(t *testing.T) {
 	}()
 
 	for _, c := range []struct {
-		sys      *types.SystemContext
-		xrd      string
-		expected string
+		sys          *types.SystemContext
+		xrd          string
+		expected     string
+		legacyFormat bool
 	}{
 		// Default paths
-		{&types.SystemContext{}, "", "/run/containers/" + uid + "/auth.json"},
-		{nil, "", "/run/containers/" + uid + "/auth.json"},
+		{&types.SystemContext{}, "", "/run/containers/" + uid + "/auth.json", false},
+		{nil, "", "/run/containers/" + uid + "/auth.json", false},
 		// SystemContext overrides
-		{&types.SystemContext{AuthFilePath: "/absolute/path"}, "", "/absolute/path"},
-		{&types.SystemContext{RootForImplicitAbsolutePaths: "/prefix"}, "", "/prefix/run/containers/" + uid + "/auth.json"},
+		{&types.SystemContext{AuthFilePath: "/absolute/path"}, "", "/absolute/path", false},
+		{&types.SystemContext{LegacyFormatAuthFilePath: "/absolute/path"}, "", "/absolute/path", true},
+		{&types.SystemContext{RootForImplicitAbsolutePaths: "/prefix"}, "", "/prefix/run/containers/" + uid + "/auth.json", false},
 		// XDG_RUNTIME_DIR defined
-		{nil, tmpDir, tmpDir + "/containers/auth.json"},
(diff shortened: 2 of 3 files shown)


pkg/docker/config/config.go
```diff
@@ -98,7 +98,9 @@ func GetAuthentication(sys *types.SystemContext, registry string) (string, strin
 	paths = append(paths, filepath.Join(homedir.Get(), dockerHomePath), dockerLegacyPath)
 
 	for _, path := range paths {
-		legacyFormat := path == dockerLegacyPath
+		// in addition to the home dir, openshift may mount .dockercfg files (via secret mount)
+		// in locations other than the home dir
+		legacyFormat := path == dockerLegacyPath || strings.HasSuffix(path, dockerLegacyHomePath)
```
mtrmac: `filepath.Base`, not `strings.HasSuffix`, please.

gabemontero:
> A quick take after a one-minute look: I’m not too enthusiastic about making a decision based on file name, introducing something like a `types.SystemContext.LegacyFormatAuthFilePath` with explicit caller control seems cleaner to me.
> 
> (Maybe turn `paths` into a `[]struct{path string, legacy bool}` and if `LegacyFormatAuthFilePath != ""`, add it to `paths` just after `pathToAuth` if any.)

The code today works off of file name ... unless you are also calling out the existing implementation as well in your comment.

That said, while more work on the openshift/builder end, as we have to track legacy vs. non-legacy format in order to seed the right field, I do see value in creating a new path the consumer chooses from and leaving existing behavior for $HOME/.dockercfg

I'll prototype in my openshift/builder branch/env and copy here if/when I get something working.

mtrmac:
> The code today works off of file name ... unless you are also calling out the existing implementation as well in your comment.

A little bit of both — the current code is only _expected_ to match the implicitly-added `dockerLegacyPath`, and it should probably always have been a `[]struct{path string, legacy bool}` instead of that wasteful string comparison. The current behavior, where the user can set `SystemContext.AuthFilePath` to exactly `dockerLegacyPath` and cause that file to be read first, using the legacy format, was never explicitly designed, and is arguably a bug (because `AuthFilePath` is expected to use the new format).

pkg/docker/config/config.go
```diff
@@ -85,27 +90,36 @@ func GetAuthentication(sys *types.SystemContext, registry string) (string, strin
...
-	paths = append(paths, filepath.Join(homedir.Get(), dockerHomePath), dockerLegacyPath)
+	homeDirDockerCFG := authPath{
+		path:         dockerLegacyPath,
+		legacyFormat: true,
+	}
+	homeDirConfigJSON := authPath{
+		path:         filepath.Join(homedir.Get(), dockerHomePath),
+		legacyFormat: false,
+	}
+	paths = append(paths, homeDirConfigJSON, homeDirDockerCFG)
```
mtrmac: Please use the same order for the `paths = ` build and for declaring the two variables above — or, maybe even better, inline it without adding the extra variables:
```go
paths = append(paths,
    authPath{path:…, legacyFormat: false},
    authPath{path:…. legacyFormat: true}
)
```

gabemontero:
An implementation of your alternative approach from https://github.com/containers/image/pull/722#issuecomment-537551966 has been pushed @mtrmac 

PTAL / thanks

pkg/docker/config/config.go
```diff
@@ -85,27 +90,36 @@ func GetAuthentication(sys *types.SystemContext, registry string) (string, strin
 	}
 
 	dockerLegacyPath := filepath.Join(homedir.Get(), dockerLegacyHomePath)
```
mtrmac: `dockerLegacyPath` can now be inlined into its only user.

pkg/docker/config/config.go
```diff
@@ -85,27 +90,36 @@ func GetAuthentication(sys *types.SystemContext, registry string) (string, strin
...
+		legacyFormat: false,
+	}
+	paths = append(paths, homeDirConfigJSON, homeDirDockerCFG)
 
 	for _, path := range paths {
-		legacyFormat := path == dockerLegacyPath
-		username, password, err := findAuthentication(registry, path, legacyFormat)
+		// in addition to the home dir, openshift may mount .dockercfg files (via secret mount)
+		// in locations other than the home dir
+		username, password, err := findAuthentication(registry, path.path, path.legacyFormat)
```
mtrmac: Recording the motivation is definitely useful, but it’s unclear to me what the reader of the comment is supposed to do about it _at this point_. It would be a better fit in the `getPathToAuth` that deals with the `LegacyFormatAuthFilePath` option, in `types.go` declaring the option, or maybe only in a commit message.

pkg/docker/config/config.go
```diff
@@ -232,7 +249,7 @@ func modifyJSON(sys *types.SystemContext, editor func(auths *dockerConfigFile) (
 		}
 	}
 
-	auths, err := readJSONFile(path, false)
+	auths, err := readJSONFile(path, legacyFormat)
```
mtrmac: This won’t work; `readJSONFile` would convert the data from the legacy format, and the code below would write new-format data to the legacy-format path.

I don’t at all think we need to add write support for the legacy format here.

Either `getPathToAuth` could get an `acceptLegacy` (`readOnly`?) parameter, or `LegacyFormatAuthFilePath` could be handled only in the read path, in `GetAuthentication` changing the path used by `homeDirDockerCFG`.

types/types.go
```diff
@@ -462,8 +462,10 @@ type SystemContext struct {
 	RegistriesDirPath string
 	// Path to the system-wide registries configuration file
 	SystemRegistriesConfPath string
-	// If not "", overrides the default path for the authentication file
+	// If not "", overrides the default path for the authentication file, but only new format files
 	AuthFilePath string
+	// if not "", overrides the default path for the authentication file, but with the legacy format
+	LegacyFormatAuthFilePath string
```
mtrmac: This should document that the value is
- only used for reading (if that’s how the code ends up being structured)
- ignored if `AuthFilePath` is set (or whatever the behavior ends up being)

pkg/docker/config/config_test.go
```diff
@@ -33,32 +33,37 @@ func TestGetPathToAuth(t *testing.T) {
...
-		res, err := getPathToAuth(c.sys)
+		res, lf, err := getPathToAuth(c.sys)
 		if c.expected == "" {
 			assert.Error(t, err)
 		} else {
 			require.NoError(t, err)
 			assert.Equal(t, c.expected, res)
 		}
+		if c.legacyFormat != lf {
+			t.Errorf("should not have received legacyFormat==%v for %#v", lf, c.sys)
```
mtrmac:
```suggestion
			assert.Equal(t, c.legacyFormat, lf)
```
and only in the `c.expected != ""` case.

mtrmac:
BTW openshift/builder is using Glide - can that deal with versioned Go modules? #721 is considered for imminent merging.

gabemontero:
> BTW openshift/builder is using Glide - can that deal with versioned Go modules? #721 is considered for imminent merging.

Specific commit levels can be referenced via glide ... this is what we'll have to do until either
a) a new tag is crafted in containers/image with this change
b) openshift/builder switches off of glide .... that is systematically happening as part of upgrading various openshift repos to k8s 1.16 .... we'll just have to seen when that work lands on the docket ... is not necessarily a 4.3 item

mtrmac:
… ah, right; Glide can probably completely ignore the `c/image/v4/…` prefix that Go modules introduce; the vendored copy would be in `vendor/github.com/c/image/…`, and can be referenced as such.

pkg/docker/config/config.go
gabemontero: part of next push

pkg/docker/config/config.go
gabemontero: part of next push

pkg/docker/config/config.go
gabemontero: I moved it to types.go

IMO it will be much harder to find these details if they are only in a commit

pkg/docker/config/config_test.go
gabemontero: part of next push

pkg/docker/config/config.go
gabemontero: or how about throw an error if this is called with a SystemContext that has `LegacyFormatAuthFilePath` set ?

types/types.go
gabemontero: part of next push

pkg/docker/config/config.go
mtrmac: … at this precise point, `if legacyFormat` (assuming `AuthFilePath` is documented to take precedence over `LegacyFormatAuthFilePath`, and we won’t pedantically throw an error every time both are specified).

Yes, that would work nicely as well, and allow us to add support for editing the legacy format in the future if it ever became necessary.

gabemontero:
updates pushed in separate commit @mtrmac ptal thanks

gabemontero:
PR rebased to address merge conflict, commits squashed @mtrmac ptal when time permits

gabemontero:
the travis failures seem unrelated to my change

rhatdan:
@gabemontero They are related to your PR.

Please

git commit -a -s --amend
Then change your first line description to be 
`allow for .dockercfg files to reside in non-home directories`
Followed by an empty line
Then add some more text describing the change.

git push --force

And your PR should be ready for review/merge.

gabemontero:
thanks for the help @rhatdan  .... sorry, I did not infer the need for that change based on the travis output.

gabemontero:
yeah travis is happy :-)

rhatdan:
Now we will see if @mtrmac is happy.

pkg/docker/config/config.go
```diff
@@ -232,6 +240,9 @@ func modifyJSON(sys *types.SystemContext, editor func(auths *dockerConfigFile) (
 		}
 	}
 
+	if legacyFormat {
+		return fmt.Errorf("writes to auth.json based on legacy format docker config like %s is not supported", path)
```
mtrmac: (Non-blocking:
- `writes… is not` is inconsistent
- could just say `writes to %s using legacy format` without mentioning the `auth.json` file name that is not all that likely to be used in this case.
)

gabemontero: updated error message to `writes to %s using legacy format` @mtrmac

mtrmac: …`are not supported`, I shouldn’t have been lazy. Sorry.

gabemontero: no worries ... update pushed

gabemontero:
> LGTM, with a non-blocking nit.
> 
> How urgent is merging this right now? #721 + #725 are pending, necessary for Podman, and there is some pressure to get Podman updated; so, adding ever more pending changes to that pipeline introduces some extra risk.
> 
> @gabemontero @rhatdan I have no idea what the relative priorities of the two (4.2. disconnected vs. ? FIPS) are at this very moment.

I don't have sufficient perspective neither on the prioritizing those two against each other @mtrmac 

Pending @rhatdan weighing in, I talked to my manager @pweil- and he had these points:

1) disconnected is higher because 
  a) it is a current feature that is not complete vs. a new feature like FIPS
  b) possibly, not sure, disconnected has more customer enablement behind it

2) if after more thought, the risk of merging this now compared to 721 and 725 is still a concern, pull in the respective owners to decide:

Disconnected: Manager: Steve Cuppett
Tech: Urvashi Mohnani
PM: Katherine Dubé

FIPS: Mgr: Neelesh Agrawal
Eng: Mrunal Patel
PM: Barrett

rhatdan:
Sophie's Choice.
I would say disconnected is more important, followed by fips mode.

pweil-:
> Sophie's Choice.

:+1: :smile:

mtrmac:
This does look safe enough to merge in the meantime, but we’ve been in the “safe enough, tests pass” situation with surprising bugs showing up only in Podman tests for a few days now, so I might be too paranoid.

For now, I’m mostly hoping that the Podman tests succeed fast enough so that it becomes a non-issue.

Let’s see how the current test run in https://github.com/containers/libpod/pull/4165 fares — if that succeeds, we can merge everything in quick order; if not, let’s get this one in and work on Podman afterwards.

mtrmac:
Update: https://github.com/containers/libpod/pull/4165 was green, baby-sitting the sequence of PRs to merge everything through the stack now. This can be merged after #721 , which should be in a few hours at most.

Do note that #721 introduces the Go module major versioning namespace. And…
> Glide can probably completely ignore the c/image/v4/… prefix that Go modules introduce; the vendored copy would be in vendor/github.com/c/image/…, and can be referenced as such.

this is incorrect: all intra-repo references now say `c/image/v4/…`, the vendored copy must actually be placed into `vendor/github.com/c/image/v4/…`.

gabemontero:
https://github.com/containers/image/pull/721 has merged

Need to rebase as a result per the indication above.

gabemontero:
ok rebased and pushed @mtrmac

mtrmac:
Tests were failing because Skopeo was not importing v4 yet, let’s see now that https://github.com/containers/skopeo/pull/736 is in.

mtrmac:
… yeah, that’s not going to work correctly with `test-skopeo` doing
```
GO111MODULE="on" go mod edit -replace github.com/containers/image=$${project_path} && \
```
Fixing that separately.

mtrmac:
@gabemontero Please rebase one more time, to explicitly bring in https://github.com/containers/image/pull/727 so that `test-skopeo` works. (Apparently Travis fetches the PR commit as is, without any regard to anything else that has happened on the branch in the meantime.)

gabemontero:
rebased @mtrmac

gabemontero:
travis is happy again @mtrmac :-)

gabemontero:
bump @mtrmac - can we merge ?

mtrmac:
Yup - thanks again!

gabemontero:
Awesome.  Likewise thanks !!

gabemontero:
fyi @mtrmac  looks like I was incorrect about glide handling *versioned* go modules

the `v4`  path injection is confusing it

I'll hunt around a bit more for work arounds but otherwise will probably have to accelerator openshift/builder's switch from glide to gomod to consume this change for disconnected builds

@adambkaplan @nalind @openshift/openshift-team-developer-experience FYI

mtrmac:
> fyi @mtrmac looks like I was incorrect about glide handling _versioned_ go modules
> 
> the `v4` path injection is confusing it

FWIW none of the trivial workarounds (GO111MODULE=on) worked for me either.

(Also, because openshift/builder creates `c/image/types.SystemContext` values that require the post-v4.0.1 feature from this PR, it will have to use version of c/buildah that uses the v4 c/image major version, i.e. the just-released 1.11.3, so that that c/buildah calls the v4.0.1 c/image.)

gabemontero:
OK @mtrmac I was able to get openshift/builder switched to go mod and have it referencing v4.0.1 of c/image and v1.11.3 of c/buildah and have it compiling and running unit tests.

However, I cannot seem to get it to reference v4.0.1 plus https://github.com/containers/image/commit/98c3ba15ed7e75fc4cd1b3a8364d2090350281d4

I've tried a variety of syntaxes, starting off of both v4.0.1 and v0.0.0, but `go mod tidy` and `go mod vendor` are not happy with any of them.

Unless you know what the secret sauce for that, any chance you guys can cut a v4.0.2 so I can just reference that ?

thanks

gabemontero:
ah ha ... got it to work @mtrmac 

that said, if you are of the inclination to cut a v4.0.2, I'd prefer to reference that

mtrmac:
I was hoping that `GO111MODULE=on go get github.com/containers/image/v4@master` would work, but it doesn’t — not sure whether it just doesn’t work in general, or whether it is confused by how broken our v4.0.0 is.

I’m not fundamentally opposed to cutting a release, but I’d slightly prefer to work on something else right now. Maybe I should just get over it.

(For the record, the next release would be v4.1.0 because this PR adds an API.)

PR #723, opened 2019-10-02
manifest update: consider that storage lies

vrothberg:
The manifests in containers storage lie since the layers are stored
uncompressed while, in some cases, being displayed as compressed in
the manifest.

Consider this case when updating the manifest to avoid immediate
errors.

### DIFF
--- a/manifest/docker_schema2.go
+++ b/manifest/docker_schema2.go
@@ -211,6 +211,28 @@ func (m *Schema2) LayerInfos() []LayerInfo {
 	return blobs
 }
 
+// isSchema2ForeignLayer is a convenience wrapper to check if a given mime type
+// is a compressed or decompressed schema 2 foreign layer.
+func isSchema2ForeignLayer(mimeType string) bool {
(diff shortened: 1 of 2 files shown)


mtrmac:
> but the design issue is really generic (affecting decompression/recompression) just as well. 

(Arguably storage should be primarily storing a manifest that corresponds to the actual uncompressed layers, and recording the original one in as an optional attachment, aka `ImageSource.GetOriginalManifest` of #288; then `LayerInfosForCopy` should not exist and all of this would go away. But we are definitely not doing that kind of work right now.)

vrothberg:
Ack! Will update and add functions for v2 and OCI with comments. In the meantime, I opened https://github.com/containers/skopeo/issues/734 so we could verify at least parts of the affected code paths.

vrothberg:
Updated. @mtrmac, feel free to take over this PR in case something's missing. I'll be leaving the keyboard in a bit.

rhatdan:
LGTM

PR #724, opened 2019-10-02
DO NOT MERGE: Integrating #721 and #723 for testing

mtrmac:
**Do not merge:** contains unmerged #721 and #723.

… so that it can be tested in consumers.


mtrmac:
#723 was merged, retiring this; any testing can use #721 directly.

mtrmac:
Now testing #721 + #725 .

PR #725, opened 2019-10-03
GuessMIMEType: recognize OCI images with no layers

nalind:
Don't require OCI image manifests to have layers in order for us to be able to recognize that we're looking at an OCI image manifest.

This should help with https://github.com/containers/libpod/pull/4165.

### DIFF
--- a/manifest/manifest.go
+++ b/manifest/manifest.go
@@ -127,12 +127,11 @@ func GuessMIMEType(manifest []byte) string {
 			Config struct {
 				MediaType string `json:"mediaType"`
 			} `json:"config"`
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

mtrmac:
@nalind Great find.

OTOH https://github.com/opencontainers/image-spec/blob/master/manifest.md does require at least one layer — is the libpod test really _correct_ to pass?

mtrmac:
\*shrug\* This is a guessing heuristic anyway, and being restrictive _here_ (unlike, potentially, the OCI manifest parser) does not really help anything in any case.  Let’s take it through the test PRs at least.

PR #729, opened 2019-10-11
pkg/docker/config: correct defaultPerUIDPathFormat

nalind:
If I'm reading https://github.com/containers/libpod/issues/4227 right, we should make `defaultPerUIDPathFormat` correspond to the value that we'd compute by combining the default value of `${XDG_RUNTIME_DIR}` (typically `/run/user/$(id -u)`) and `xdgRuntimeDirPath`.

### DIFF
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -32,7 +32,7 @@ type authPath struct {
 }
 
 var (
-	defaultPerUIDPathFormat = filepath.FromSlash("/run/containers/%d/auth.json")
+	defaultPerUIDPathFormat = filepath.FromSlash("/run/user/%d/containers/auth.json")
 	xdgRuntimeDirPath       = filepath.FromSlash("containers/auth.json")
 	dockerHomePath          = filepath.FromSlash(".docker/config.json")
 	dockerLegacyHomePath    = ".dockercfg"
--- a/pkg/docker/config/config_test.go
+++ b/pkg/docker/config/config_test.go
@@ -39,12 +39,12 @@ func TestGetPathToAuth(t *testing.T) {
 		legacyFormat bool
 	}{
 		// Default paths
-		{&types.SystemContext{}, "", "/run/containers/" + uid + "/auth.json", false},
-		{nil, "", "/run/containers/" + uid + "/auth.json", false},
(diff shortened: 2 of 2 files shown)


mtrmac:
Thanks.  The non-`XDG_RUNTIME_DIR` default was, IIRC, intentionally different, e.g. https://github.com/containers/image/pull/424 says:
> Looking at `pam_systemd(8)`, it **removes** `/run/user` when “the last concurrent session of a user ends”; so, if we created the directory without `systemd` knowing about the session, it can delete the directory at an unpredictable time when some other session terminates.

so it may not be all quite as simple (but I haven’t reread all of the previous issues/PRs/bugs filed).

One clear problem with the current code is that it does not work at all for non-`root` users who are not allowed to create subdirectories in `/run/containers` (OTOH that’s also the case for `/run/$UID`), so I’m not at all saying that the current code is right, just that the intent of this code path was not to match $XDG_RUNTIME_DIR.

nalind:
> One clear problem with the current code is that it does not work at all for non-`root` users who are not allowed to create subdirectories in `/run/containers` (OTOH that’s also the case for `/run/user/$UID`), so I’m not at all saying that the current code is right, just that the intent of this code path was not to match $XDG_RUNTIME_DIR.

Yeah, the "not being able to create `/run/containers` or subdirectories under it" was what had me thinking this was an error.  If it's intentional, though, should I just close this one?

mtrmac:
See also the conversation in https://github.com/containers/libpod/pull/4232 , not sure how that will turn out.

mtrmac:
The original motivation for this PR, https://github.com/containers/podman/issues/4227 , was resolved by https://github.com/containers/podman/pull/4337 , having Podman mostly defer to this code for the defaults.

That does not mean that these non-$XDG_RUNTIME_DIR defaults work all that well (they really don’t), OTOH I don’t think we have any good choices.

The longer-term plan is to have (system-wide and per-user) `registries.conf` set up a set of credential helpers, and to use such a helper instead of a flat file for the credentials. That should make the common/ideal case better.

OTOH that does nothing for users who choose to use files, or have to use files because nothing else is available, and do not have $XDG_RUNTIME_DIR; still, I don’t think we have good choices.

jankaluza:

PR #732, opened 2019-10-22
Improve progress reader to report start/end and offset update

saschagrunert:
The progress reader will be now created via the new `newProgressReader`
function. This way we have the possibility to trigger a new event called
`ProgressEventNewArtifact` and `ProgressEventDone` to indicate the
boundaries of the artifact download to the API consumer. During the
download, we additionally provide the `OffsetUpdate`, which can be used
to report the downloaded data during the last elapsed time interval.

Unit tests have been added as well to the progress reader, whereas the
documentation for the new types has been enhanced as well.

Needed for metrics improvements for CRI-O: https://github.com/cri-o/cri-o/pull/2912

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -1154,13 +1154,13 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
 
 	// === Report progress using the c.progress channel, if required.
 	if c.progress != nil && c.progressInterval > 0 {
-		destStream = &progressReader{
-			source:   destStream,
-			channel:  c.progress,
-			interval: c.progressInterval,
-			artifact: srcInfo,
-			lastTime: time.Now(),
-		}
+		destStream := newProgressReader(
+			destStream,
+			c.progress,
+			c.progressInterval,
+			srcInfo,
+		)
+		defer destStream.reportDone()
 	}
 
 	// === Finally, send the layer stream to dest.
--- a/copy/progress_reader.go
+++ b/copy/progress_reader.go
@@ -9,20 +9,71 @@ import (
 
 // progressReader is a reader that reports its progress on an interval.
 type progressReader struct {
-	source   io.Reader
-	channel  chan types.ProgressProperties
-	interval time.Duration
-	artifact types.BlobInfo
-	lastTime time.Time
-	offset   uint64
+	source       io.Reader
+	channel      chan<- types.ProgressProperties
+	interval     time.Duration
+	artifact     types.BlobInfo
+	lastUpdate   time.Time
+	offset       uint64
+	offsetUpdate uint64
 }
 
+// newProgressReader creates a new progress reader for:
+// `source`:   The source when internally reading bytes
+// `channel`:  The reporter channel to which the progress will be sent
+// `interval`: The update interval to indicate how often the progress should update
+// `artifact`: The blob metadata which is currently being progressed
+func newProgressReader(
+	source io.Reader,
+	channel chan<- types.ProgressProperties,
+	interval time.Duration,
+	artifact types.BlobInfo,
+) *progressReader {
+	// The progress reader constructor informs the progress channel
+	// that a new artifact will be read
+	channel <- types.ProgressProperties{
+		Event:    types.ProgressEventNewArtifact,
+		Artifact: artifact,
+	}
+	return &progressReader{
+		source:       source,
+		channel:      channel,
+		interval:     interval,
+		artifact:     artifact,
+		lastUpdate:   time.Now(),
+		offset:       0,
+		offsetUpdate: 0,
+	}
+}
+
+// reportDone indicates to the internal channel that the progress has been
+// finished
+func (r *progressReader) reportDone() {
+	r.channel <- types.ProgressProperties{
+		Event:        types.ProgressEventDone,
+		Artifact:     r.artifact,
+		Offset:       r.offset,
+		OffsetUpdate: r.offsetUpdate,
+	}
+}
+
+// Read continuously reads bytes into the progress reader and reports the
+// status via the internal channel
 func (r *progressReader) Read(p []byte) (int, error) {
 	n, err := r.source.Read(p)
 	r.offset += uint64(n)
-	if time.Since(r.lastTime) > r.interval {
-		r.channel <- types.ProgressProperties{Artifact: r.artifact, Offset: r.offset}
-		r.lastTime = time.Now()
+	r.offsetUpdate += uint64(n)
+
+	// Fire the progress reader in the provided interval
+	if time.Since(r.lastUpdate) > r.interval {
+		r.channel <- types.ProgressProperties{
+			Event:        types.ProgressEventRead,
+			Artifact:     r.artifact,
+			Offset:       r.offset,
+			OffsetUpdate: r.offsetUpdate,
+		}
+		r.lastUpdate = time.Now()
+		r.offsetUpdate = 0
 	}
 	return n, err
 }
--- a/copy/progress_reader_test.go
+++ b/copy/progress_reader_test.go
@@ -1,0 +1,80 @@
+package copy
+
+import (
+	"bytes"
+	"io"
+	"testing"
+	"time"
+
+	"github.com/containers/image/v5/types"
+	"github.com/stretchr/testify/assert"
+)
+
+func newSUT(
+	t *testing.T,
+	reader io.Reader,
+	duration time.Duration,
+	channel chan types.ProgressProperties,
+) *progressReader {
+	artifact := types.BlobInfo{Size: 10}
+
+	go func() {
+		res := <-channel
+		assert.Equal(t, res.Event, types.ProgressEventNewArtifact)
+		assert.Equal(t, res.Artifact, artifact)
+	}()
+	res := newProgressReader(reader, channel, duration, artifact)
+
+	return res
+}
+
+func TestNewProgressReader(t *testing.T) {
+	// Given
+	channel := make(chan types.ProgressProperties)
+	sut := newSUT(t, nil, time.Second, channel)
+	assert.NotNil(t, sut)
+
+	// When/Then
+	go func() {
+		res := <-channel
+		assert.Equal(t, res.Event, types.ProgressEventDone)
+	}()
+	sut.reportDone()
+}
+
+func TestReadWithoutEvent(t *testing.T) {
+	// Given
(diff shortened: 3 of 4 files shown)


saschagrunert:
> Overall, sure, this works, but I’m a bit worried that it may not do quite what the caller expects.

Hey @mtrmac, thanks for the first look. My main intention was to have something which changes
the library as little as possible, so I'm totally open to some rework here.

>     * Is “the first time `Read` returns” really exactly the right moment? If the source is very slow, the caller won’t receive any report about trying to work with the blob until _after_ the first data/timeout arrives. Should the first progress indication about the artifact happen at the very start of `copyBlobFromStream`, before _asking_ for the first byte? (Or, actually, all the way up before `TryReusingBlob` and the like, so that it captures the delays communicating over HTTP and the like, and setting up the streams, not just waiting for data?)

I would expect (as a user) that the progress fires the first time a connection with the server (stream) has been successfully setup and no data got transferred. Maybe a progress state (`CONNECTED`) would be a good indicator for this.

>     * Note that `PutBlob` may never actually read all of the data (well, it may not read _any_), e.g. if the blob already exists on the destination (well, if it was uploaded _after_ `TryReusingBlob` fails). Should the last update be conditioned on `r.artifact.Size`, or be determined by code flow using a `defer` or something like that? (`r.artifact.Size` can even be -1 to mean “unknown”!) And should we, in that case, somehow tell the caller that the artifact is done (by lying and saying `Offset = Size`, or by decreasing `Size`)?

The caller needs the information that data has been skipped, because this will be an additional metric in CRI-O.

> Neither is really a big show-stopper to me, and I’m not sure that CRI-O cares about either of the two at all. Thinking _too far_ in this direction would lead to a complete redesign of the progress interface, which is currently rather a crude hack, to be much more detailed about the semantics, and I suppose that’s not all that needed right now — or is it?

Yeah, I'm not sure. The first time I used this API I must honestly admit that I expected something different from a usage perspective. I would assume that the progress bar display is handled by a more higher level library (like podman), whereas the API here works like:

1. Tell the library that I want to copy an image from A to B, where the `copy` function only returns a channel
2. I can now listen on that channel and get events (connected, pulling, interrupted, done), whereas the `pulling` events get triggered periodically and the library calculates the overall download status for me (packed in events data).

mtrmac:
> My main intention was to have something which changes
> the library as little as possible

That’s _very_ fair, and I don’t at at all want to insist on a big (API-breaking) redesign. It’s perfectly fine to do a small incremental improvement, as long as it’s clear that it still has limitations.

> > Neither is really a big show-stopper to me, and I’m not sure that CRI-O cares about either of the two at all. Thinking _too far_ in this direction would lead to a complete redesign of the progress interface, which is currently rather a crude hack, to be much more detailed about the semantics, and I suppose that’s not all that needed right now — or is it?
> 
> Yeah, I'm not sure. The first time I used this API I must honestly admit that I expected something different from a usage perspective. I would assume that the progress bar display is handled by a more higher level library (like podman), whereas the API here works like:

The current progress bars are handled by `copy.Image` directly, and that is managed in more places than `progressReader`. Yes, it would be conceptually attractive for both to be built on the same abstraction. OTOH, that’s a fair amount of work — especially knowing about #611 , which among other things, included fairly significant extensions to the progress bar model as well (to be able to switch between status updates / progress status, for the same blob / unit of work).

---

- I would expect (as a user) that the progress fires the first time a connection with the server (stream) has been successfully setup and no data got transferred.
- The caller needs the information that data has been skipped, because this will be an additional metric in CRI-O.

Doing _just_ these two might be easy enough in the ~current model, see the code comments:

copy/copy.go
```diff
@@ -1159,7 +1159,7 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
 			channel:  c.progress,
 			interval: c.progressInterval,
 			artifact: srcInfo,
-			lastTime: time.Now(),
+			lastTime: time.Time{},
```
mtrmac: `copyBlobFromStream` should ideally not need to know about the `time.Time{}` special case; this rather calls for a `newProgressReader()` constructor that handles that.

- That constructor could legitimately fire a “dealing with new artifact, 0% progress” event, without making _this_ code any more complex.
- It would then be very reasonable to add a `defer func() { progressReader.reportDone(didReadAll, errorIfAny) }` or something like that (let’s not add data that no-one needs), and ensure that there really _is_ a 100%/skipped report.
- Perhaps we could also split the constructor from a `progressReader.assignSource(destStream)` call, and call the constructor earlier than here, to get the 0% indication in the right place.

mtrmac:
> OTOH, that’s a fair amount of work — especially knowing about #611 , which among other things, included fairly significant extensions to the progress bar model as well (to be able to switch between status updates / progress status, for the same blob / unit of work).

(To be clear, that work was good and it would be a shame to lose it — and splitting that work into a smaller PR focused just on progress bars could help. Still, designing a well-typed stable API with detailed events about the progress is much more work than this PR.)

types/types.go
```diff
@@ -552,4 +552,5 @@ type SystemContext struct {
 type ProgressProperties struct {
 	Artifact BlobInfo
 	Offset   uint64
+	LastRead uint64
```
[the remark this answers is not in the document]
saschagrunert: I agree, added documentation and changed the name to `OffsetUpdate`.

saschagrunert:
Ready for review, @mtrmac and @vrothberg PTAL.

copy/progress_reader.go
```diff
@@ -9,20 +10,81 @@ import (
...
+	offset       uint64
+	offsetUpdate uint64
 }
 
+// newProgressReader creates a new progress reader for:
+// `channel`:  The reporter channel to which the progress will be sent
+// `interval`: The update interval to indicate how often the progress should update
+// `artifact`: The blob metadata which is currently being progressed
+func newProgressReader(
+	channel chan types.ProgressProperties,
```
mtrmac: Can’t this be a one-directional channel? (It should have been like that all the way in `copy.Options`, sure, but we can fix it at least the new code.)

copy/progress_reader.go
```diff
@@ -9,20 +10,81 @@ import (
...
+		artifact:     artifact,
+		lastUpdate:   time.Now(),
+		offset:       0,
+		offsetUpdate: 0,
+	}
+}
+
+// withSource sets the provided source reader to the progressReader and creates
+// a new io.Reader from it
+func (r *progressReader) withSource(source io.Reader) io.Reader {
```
mtrmac: “creates a new `io.Reader`” is incorrect.

I tend to read `withSource` as “make a copy with a changed source”.

If the caller calls `newProgressReader` and `withSource` immediately afterwards, there’s little benefit to splitting them — so if you don’t plan to move the initial step elsewhere, maybe just fold this into the constructor and we can avoid a naming discussion. (`progressReader` is package-private, so we can always split this out again if necessary.)

copy/progress_reader.go
```diff
@@ -9,20 +10,81 @@ import (
...
 	n, err := r.source.Read(p)
 	r.offset += uint64(n)
-	if time.Since(r.lastTime) > r.interval {
-		r.channel <- types.ProgressProperties{Artifact: r.artifact, Offset: r.offset}
-		r.lastTime = time.Now()
+	r.offsetUpdate += uint64(n)
+
+	// Fire the progress reader for the first and last time as well as the
+	// provided interval
+	if time.Since(r.lastUpdate) > r.interval {
```
mtrmac: The comment no longer matches the code.

copy/progress_reader_test.go
```diff
@@ -0,0 +1,94 @@
...
+)
+
+func newSUT(
+	t *testing.T,
+	duration time.Duration,
+	channel chan types.ProgressProperties,
+) *progressReader {
+	artifact := types.BlobInfo{Size: 10}
+
+	go func() {
```
mtrmac: Note to self: does the testing framework know it has to wait for all spawned goroutines?

copy/progress_reader_test.go
```diff
@@ -0,0 +1,94 @@
+package copy
```
mtrmac: Yay, tests. Great!

types/types.go
```diff
@@ -547,9 +547,35 @@ type SystemContext struct {
 	CompressionLevel *int
 }
 
+// ProgressEvent is the type of events a progress reader can produce
```
mtrmac:
```suggestion
// ProgressEvent is the type of events a progress reader can produce
// Warning: new event types may be added any time.
```
(… although, right now, we can’t really add new events that don’t set `Artifact`+`Offset` without breaking users. Eventually this should probably be a complex typed API that provides only the right fields for the right events, but not now — this is great because it keeps existing consumers working.)

types/types.go
```diff
@@ -547,9 +547,35 @@ type SystemContext struct {
...
+
+const (
+	// ProgressEventNewArtifact will be fired on progress reader setup
+	ProgressEventNewArtifact ProgressEvent = iota
+
+	// ProgressEventRead indicates that the artifact download is currently in
+	// progress
+	ProgressEventRead
+
+	// ProgressEventDone is fired when the data transfer has been finished
```
mtrmac: … for the specified artifact?

copy/progress_reader.go
saschagrunert: Yes, I make it send-only.

copy/progress_reader.go
saschagrunert: Alright I will put it back into the `newProgressReader` function and remove this one. This should be now easier and more straight forward to use.

copy/progress_reader.go
saschagrunert: Fixed :heavy_check_mark:

types/types.go
saschagrunert: Alright, I will apply the suggestion :+1:

types/types.go
saschagrunert: Yes, I added a note about that.

vrothberg:
Needs rebase.

saschagrunert:
Rebased on top of the latest master branch.

TomSweeneyRedHat:
LGTM and happy green test buttons.  Would like a final head nod from @mtrmac

rhatdan:
Hopefully this does not require a new version? :^(

vrothberg:
> Hopefully this does not require a new version? :^(

Only a minor bump is required :)

PR #734, opened 2019-10-22
copy.Image(): select the CopySystemImage image using the source context

nalind:
Go back to selecting which image instance we copy in CopySystemImage mode using the source context rather than the destination context, as we did before #400.  Using the destination context made more sense to me, but it's probably worth breaking that out that as a separate decision.


nalind:
We have a few places where callers set `OSChoice` and `ArchitectureChoice` in a source `SystemContext` to affect which manifest we choose while pulling from a registry, but don't do the same for a destination `SystemContext`, and this change in #400 would break them.

rhatdan:
LGTM

PR #735, opened 2019-10-25
Release v5.0.0

mtrmac:
In addition to the minimal bump, this also includes a commit that beautifies imports a bit.


rhatdan:
LGTM
@vrothberg PTAL

version/version.go
```diff
@@ -4,11 +4,11 @@ import "fmt"
 
 const (
 	// VersionMajor is for an API incompatible changes
-	VersionMajor = 4
+	VersionMajor = 5
 	// VersionMinor is for functionality in a backwards-compatible manner
 	VersionMinor = 0
 	// VersionPatch is for backwards-compatible bug fixes
-	VersionPatch = 2
+	VersionPatch = 1
```
nalind: Should this be 0 for 5.0.0?

nalind:
LGTM give or take a question about `VersionPatch`.

version/version.go
mtrmac: Compare https://github.com/containers/image/pull/735/commits/194eb017e848758aeccedd725c4c1a49fb2f7f67 and https://github.com/containers/image/pull/735/commits/a40e6b800d77a0603389ab5f1a78e430601b2975 .

mtrmac:
**Do not merge yet**: Apparently changing this to be a /v5 is happily ignored by `make test-skopeo` now, because it continues to reference a /v4. That’s… a win for Go modules, but loss for usefulness of the CI.


Integration tests should pass in https://github.com/containers/skopeo/pull/747 .

version/version.go
nalind: Doh.  Missed that it was multiple commits.  Looks good.

rhatdan:
Tests are passing,   Merging.

PR #736, opened 2019-10-29
transports: drop ostree

giuseppe:
drop support for the ostree backend.  The only known user for the
backend is the atomic CLI tool that is not maintained anymore
upstream.


types/types.go
```diff
@@ -524,8 +524,6 @@ type SystemContext struct {
 	// Note that this field is used mainly to integrate containers/image into projectatomic/docker
 	// in order to not break any existing docker's integration tests.
 	DockerDisableV1Ping bool
-	// Directory to use for OSTree temporary files
-	OSTreeTmpDirPath string
```
mtrmac: This is an API break and it would require a major version bump.

For now, keeping this value around and ignored would work well enough, as far as _this_ type is concerned…

… but, really, dropping an entire subpackage, and the `ostree:` syntax, is an API break and requires a major version bump, either way.

giuseppe:
fix for the CI failure: https://github.com/containers/skopeo/pull/752

types/types.go
giuseppe: I've pushed another patch to bump the version to 6

rhatdan:
@giuseppe I don't see this as critical, as long as the other tools have dropped support.  This can wait until we have another reason to update containers/image API, do you agree?

giuseppe:
> @giuseppe I don't see this as critical, as long as the other tools have dropped support. This can wait until we have another reason to update containers/image API, do you agree?

yes I am fine with that.  Should we just wait for the next release and I rebase, or any other way to move this forward?

rhatdan:
If we merge now, then if we want to get a minor fix out for V5 we would need to reverse the patch or should we just create a V5 branch now, and then merge.

vrothberg:
> If we merge now, then if we want to get a minor fix out for V5 we would need to reverse the patch or should we just create a V5 branch now, and then merge.

I think it's worth branching v5 (given it will be supported for a long time with Podman in RHEL) but I will leave the decision to @mtrmac.

However, this doesn't really solve the issue that merging this PR will mean yet another major bump. If we want to ship new features we either need to cherry-pick to v5 or do the vendoring dance with a new major version. Or, we just hold this PR as I don't see an immediate need to merge it. The timing is unfortunate so close after having released v5.

mtrmac:
> > If we merge now, then if we want to get a minor fix out for V5 we would need to reverse the patch or should we just create a V5 branch now, and then merge.
> 
> I think it's worth branching v5 (given it will be supported for a long time with Podman in RHEL) but I will leave the decision to @mtrmac.

(Will Podman really be frozen at v5? 8.0 had older versions originally, so if we bumped to c/image/v5, wouldn’t we bump to /v6 in the future as well?)

> However, this doesn't really solve the issue that merging this PR will mean yet another major bump. If we want to ship new features we either need to cherry-pick to v5 or do the vendoring dance with a new major version.

This is really the question. Creating a branch is cheap, but are we setting the expectation that every PR has to be done two or more times (like in CRI-O), or are we going to only add features and most bug fixes on the master branch?

vrothberg:
> (Will Podman really be frozen at v5? 8.0 had older versions originally, so if we bumped to c/image/v5, wouldn’t we bump to /v6 in the future as well?)

The new one for RHEL will be frozen with /v5 but can also create a branch on-demand.

> > However, this doesn't really solve the issue that merging this PR will mean yet another major bump. If we want to ship new features we either need to cherry-pick to v5 or do the vendoring dance with a new major version.
> 
> This is really the question. Creating a branch is cheap, but are we setting the expectation that every PR has to be done two or more times (like in CRI-O), or are we going to only add features and most bug fixes on the master branch?

I have no preference. There are so many different ways to approach the versioning and release process. In this specific case, I would like to avoid another major bump in the near future.

rhatdan:
@mtrmac There are two streams on RHEL8  For RHEL8.2 we will be freezing on a version of Podman for 2 years with V5.  The update stream will continue to be updated every three months with the latest podman.  But if we have CVEs or required back ports for containers/image we would go to the V5 branch.

cristim:
Is this PR still relevant nowadays?

giuseppe:
> Is this PR still relevant nowadays?

yes, I still think we should drop the ostree backend at the next major version bump

PR #737, opened 2019-11-01
add support for default credential store

koshatul:
#### What change does this introduce?

Added a check for `credsStore` in the `${HOME}/.docker/config.json` file, at least on OSX the `auths` list is just URIs and there isn't a `credHelpers` list.

#### Why make this change?

To implement authentication when using applications that import this library on OSX and possibly other BSD/*nix OS

#### What approach will be taken?

Implement the check for a default credentials store past the registry specific `credHelpers` list.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,7 @@ clean:
 	rm -rf tools.timestamp $(MANPAGES)
 
 test:
-	@PATH="$(PATH):$(shell pwd)/pkg/docker/config/testdata" $(GPGME_ENV) GO111MODULE="on" go test $(BUILDFLAGS) -cover ./...
+	@$(GPGME_ENV) GO111MODULE="on" go test $(BUILDFLAGS) -cover ./...
 
 # This is not run as part of (make all), but Travis CI does run this.
 # Demonstrating a working version of skopeo (possibly with modified SKOPEO_REPO/SKOPEO_BRANCH, e.g.
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,7 @@ clean:
 	rm -rf tools.timestamp $(MANPAGES)
 
 test:
-	@$(GPGME_ENV) GO111MODULE="on" go test $(BUILDFLAGS) -cover ./...
+	@PATH="$(PATH):$(shell pwd)/pkg/docker/config/testdata" $(GPGME_ENV) GO111MODULE="on" go test $(BUILDFLAGS) -cover ./...
 
 # This is not run as part of (make all), but Travis CI does run this.
 # Demonstrating a working version of skopeo (possibly with modified SKOPEO_REPO/SKOPEO_BRANCH, e.g.
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -24,6 +24,7 @@ type dockerAuthConfig struct {
 type dockerConfigFile struct {
 	AuthConfigs map[string]dockerAuthConfig `json:"auths"`
 	CredHelpers map[string]string           `json:"credHelpers,omitempty"`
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -63,7 +63,7 @@ clean:
 	rm -rf tools.timestamp $(MANPAGES)
 
 test:
-	@$(GPGME_ENV) GO111MODULE="on" go test $(BUILDFLAGS) -cover ./...
+	@PATH="$(PATH):$(shell pwd)/pkg/docker/config/testdata" $(GPGME_ENV) GO111MODULE="on" go test $(BUILDFLAGS) -cover ./...
```
mtrmac: Please don’t change global configuration for a single test like this. See e.g. how `XDG_RUNTIME_DIR` is handled in other tests, including the comments about `t.Parallel()`.

pkg/docker/config/config.go
```diff
@@ -305,6 +306,11 @@ func findAuthentication(registry, path string, legacyFormat bool) (string, strin
 		return getAuthFromCredHelper(ch, registry)
 	}
 
+	// Second try default credential store
+	if cs := auths.CredStore; cs != "" {
+		return getAuthFromCredHelper(cs, registry)
+	}
+
```
mtrmac: At least `SetAuthentication` and `RemoveAuthentication`, and maybe also `RemoveAllAuthentication` (not at all sure now) needs to be updated as well.

pkg/docker/config/config_test.go
```diff
@@ -225,6 +225,73 @@ func TestGetAuth(t *testing.T) {
 	}
 }
 
+func TestGetHelperAuth(t *testing.T) {
```
mtrmac: This looks like something that should be fairly easy to add as one more test case in the existing `TestGetAuth`, without so much code duplication.

pkg/docker/config/testdata/helper.json
```diff
@@ -0,0 +1,6 @@
+{
+    "auths": {
+        "foobar.example.org": {}
```
mtrmac: Hum, do we actually want commit to a design in which `credsStore` takes precedence over explicit entries for registries? (Note to self: What does Docker do?)

koshatul: This is the (sanitised) docker config from my macbook, it seems to put a dummy entry in the auths and rely on the credstore to do it.

```json
{
  "auths" : {
    "private.registry.yoda" : {

    },
  },
  "HttpHeaders" : {
    "User-Agent" : "Docker-Client/19.03.2 (darwin)"
  },
  "stackOrchestrator" : "swarm",
  "credsStore" : "desktop"
}
```

looking at the source for docker/cli it looks like it uses the default `credStore` before falling back to the `auths` list.
https://github.com/docker/cli/blob/master/cli/config/configfile/file.go#L290

But I did have some trouble following the docker/cli code.

Makefile
koshatul: Fixed, implemented like `XDG_RUNTIME_DIR` in the unit test itself.

pkg/docker/config/config.go
koshatul: Updated, fixed the order of importance as well.

pkg/docker/config/config_test.go
koshatul: I had some issues with adding it, turns out whatever is the last thing there will leave its config in the "other config path", which works with the current tests because of their layout.

pkg/docker/config/testdata/helper.json
koshatul: I changed the order of importance so it is more logical.

jankaluza:

PR #744, opened 2019-11-01
Update the Ubuntu host to get newer kernel

mtrmac:
See https://github.com/containers/image/pull/744#issuecomment-557291005 below.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: go
 sudo: required
 notifications:
   email: false
-dist: trusty
+dist: xenial
 services:
   - docker
 os:


mtrmac:
Finally passed with
> ```
> SKOPEO_REPO = mtrmac/skopeo
> SKOPEO_BRANCH = test-failures
> ```

which I’ll now drop. This PR will show test failures until https://github.com/containers/skopeo/pull/760 is merged; afterwards, it should be re-tested, and tests will hopefully succeed, unblocking c/image CI.

mtrmac:
Update the Ubuntu host to get newer kernel because `crun` (used in newer `fedora` containers) calls `statx()`, which is available since kernel 4.11, but Ubuntu Trusty only has 4.4. The glibc emulation of `statx()` fails with `EINVAL` on unrecognized flags, which `crun` does not recognize, so it does not fall back to older syscalls.
    
https://github.com/containers/crun/issues/189

Based on previous testing, this should unblock c/image CI after https://github.com/containers/skopeo/pull/760 is merged.

@vrothberg PTAL.

PR #747, opened 2019-11-06
Added possibility to overwrite default tmp dir for big files

### LINKED ISSUE #495: Enable overriding TemporaryDirectoryForBigFiles in runtime
I need to pull big images - hundreds of GBs. I don't want to mount  my containers related device to /var/tmp . I don't want to rebuild this project just to change the tmp directory.

Would it be possible to introduce an environment variable (e.g. CONTAINERS_IMAGE_TMP_DIR?) or a configuration option where I could specify the directory for big files?

Zyqsempai:
Fixed #495 
Added possibility to overwrite default tmp dir for big files.
The problem in this PR is in that fact that not in all usage places it is possible to use SystemContext.
Signed-off-by: bpopovschi <zyqsempai@mail.ru>

### DIFF
--- a/docker/tarfile/dest.go
+++ b/docker/tarfile/dest.go
@@ -33,7 +33,13 @@ type Destination struct {
 }
 
 // NewDestination returns a tarfile.Destination for the specified io.Writer.
+// Deprecated: please use NewDestinationWithContext instead
 func NewDestination(dest io.Writer, ref reference.NamedTagged) *Destination {
+	return NewDestinationWithContext(nil, dest, ref)
+}
+
+// NewDestinationWithContext returns a tarfile.Destination for the specified io.Writer.
+func NewDestinationWithContext(sys *types.SystemContext, dest io.Writer, ref reference.NamedTagged) *Destination {
 	repoTags := []reference.NamedTagged{}
 	if ref != nil {
 		repoTags = append(repoTags, ref)
@@ -43,6 +49,7 @@ func NewDestination(dest io.Writer, ref reference.NamedTagged) *Destination {
 		tar:      tar.NewWriter(dest),
 		repoTags: repoTags,
 		blobs:    make(map[digest.Digest]types.BlobInfo),
+		sysCtx:   sys,
 	}
 }
 
@@ -95,8 +102,6 @@ func (d *Destination) HasThreadSafePutBlob() bool {
 // WARNING: The contents of stream are being verified on the fly.  Until stream.Read() returns io.EOF, the contents of the data SHOULD NOT be available
 // to any other readers for download using the supplied digest.
 // If stream.Read() at any time, ESPECIALLY at end of input, returns an error, PutBlob MUST 1) fail, and 2) delete any data stored so far.
-// Deprecated: Please use PutBlobWithSystemContext which will allows you to configure temp directory
-// for big files through SystemContext.BigFilesTemporaryDir
 func (d *Destination) PutBlob(ctx context.Context, stream io.Reader, inputInfo types.BlobInfo, cache types.BlobInfoCache, isConfig bool) (types.BlobInfo, error) {
 	// Ouch, we need to stream the blob into a temporary file just to determine the size.
 	// When the layer is decompressed, we also have to generate the digest on uncompressed datas.
--- a/docker/tarfile/dest.go
+++ b/docker/tarfile/dest.go
@@ -29,6 +29,7 @@ type Destination struct {
 	// Other state.
 	blobs  map[digest.Digest]types.BlobInfo // list of already-sent blobs
 	config []byte
+	sysCtx *types.SystemContext
 }
 
 // NewDestination returns a tarfile.Destination for the specified io.Writer.
@@ -94,12 +95,14 @@ func (d *Destination) HasThreadSafePutBlob() bool {
 // WARNING: The contents of stream are being verified on the fly.  Until stream.Read() returns io.EOF, the contents of the data SHOULD NOT be available
 // to any other readers for download using the supplied digest.
 // If stream.Read() at any time, ESPECIALLY at end of input, returns an error, PutBlob MUST 1) fail, and 2) delete any data stored so far.
+// Deprecated: Please use PutBlobWithSystemContext which will allows you to configure temp directory
+// for big files through SystemContext.BigFilesTemporaryDir
 func (d *Destination) PutBlob(ctx context.Context, stream io.Reader, inputInfo types.BlobInfo, cache types.BlobInfoCache, isConfig bool) (types.BlobInfo, error) {
 	// Ouch, we need to stream the blob into a temporary file just to determine the size.
 	// When the layer is decompressed, we also have to generate the digest on uncompressed datas.
 	if inputInfo.Size == -1 || inputInfo.Digest.String() == "" {
 		logrus.Debugf("docker tarfile: input with unknown size, streaming to disk first ...")
-		streamCopy, err := ioutil.TempFile(tmpdir.TemporaryDirectoryForBigFiles(), "docker-tarfile-blob")
+		streamCopy, err := ioutil.TempFile(tmpdir.TemporaryDirectoryForBigFiles(d.sysCtx), "docker-tarfile-blob")
 		if err != nil {
 			return types.BlobInfo{}, err
 		}
--- a/docker/tarfile/src.go
+++ b/docker/tarfile/src.go
@@ -46,7 +46,14 @@ type layerInfo struct {
 // 	To do for both the NewSourceFromFile and NewSourceFromStream functions
 
 // NewSourceFromFile returns a tarfile.Source for the specified path.
+// Deprecated: Please use NewSourceFromFileWithContext which will allows you to configure temp directory
+// for big files through SystemContext.BigFilesTemporaryDir
 func NewSourceFromFile(path string) (*Source, error) {
+	return NewSourceFromFileWithContext(nil, path)
+}
+
+// NewSourceFromFileWithContext returns a tarfile.Source for the specified path.
+func NewSourceFromFileWithContext(sys *types.SystemContext, path string) (*Source, error) {
 	file, err := os.Open(path)
 	if err != nil {
 		return nil, errors.Wrapf(err, "error opening file %q", path)
@@ -65,16 +72,25 @@ func NewSourceFromFile(path string) (*Source, error) {
 			tarPath: path,
 		}, nil
 	}
-	return NewSourceFromStream(stream)
+	return NewSourceFromStreamWithSystemContext(sys, stream)
 }
 
 // NewSourceFromStream returns a tarfile.Source for the specified inputStream,
 // which can be either compressed or uncompressed. The caller can close the
 // inputStream immediately after NewSourceFromFile returns.
+// Deprecated: Please use NewSourceFromStreamWithSystemContext which will allows you to configure
+// temp directory for big files through SystemContext.BigFilesTemporaryDir
 func NewSourceFromStream(inputStream io.Reader) (*Source, error) {
+	return NewSourceFromStreamWithSystemContext(nil, inputStream)
+}
+
+// NewSourceFromStreamWithSystemContext returns a tarfile.Source for the specified inputStream,
+// which can be either compressed or uncompressed. The caller can close the
+// inputStream immediately after NewSourceFromFile returns.
+func NewSourceFromStreamWithSystemContext(sys *types.SystemContext, inputStream io.Reader) (*Source, error) {
 	// FIXME: use SystemContext here.
 	// Save inputStream to a temporary file
-	tarCopyFile, err := ioutil.TempFile(tmpdir.TemporaryDirectoryForBigFiles(), "docker-tar")
+	tarCopyFile, err := ioutil.TempFile(tmpdir.TemporaryDirectoryForBigFiles(sys), "docker-tar")
 	if err != nil {
 		return nil, errors.Wrap(err, "error creating temporary file")
 	}
--- a/internal/tmpdir/tmpdir.go
+++ b/internal/tmpdir/tmpdir.go
@@ -3,6 +3,8 @@ package tmpdir
 import (
 	"os"
 	"runtime"
+
+	"github.com/containers/image/v5/types"
 )
 
 // unixTempDirForBigFiles is the directory path to store big files on non Windows systems.
@@ -18,7 +20,10 @@ const builtinUnixTempDirForBigFiles = "/var/tmp"
 // TemporaryDirectoryForBigFiles returns a directory for temporary (big) files.
 // On non Windows systems it avoids the use of os.TempDir(), because the default temporary directory usually falls under /tmp
 // which on systemd based systems could be the unsuitable tmpfs filesystem.
-func TemporaryDirectoryForBigFiles() string {
+func TemporaryDirectoryForBigFiles(sys *types.SystemContext) string {
+	if sys != nil && sys.BigFilesTemporaryDir != "" {
+		return sys.BigFilesTemporaryDir
+	}
 	var temporaryDirectoryForBigFiles string
 	if runtime.GOOS == "windows" {
 		temporaryDirectoryForBigFiles = os.TempDir()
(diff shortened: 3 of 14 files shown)


internal/tmpdir/tmpdir.go
```diff
@@ -15,10 +15,15 @@ var unixTempDirForBigFiles = builtinUnixTempDirForBigFiles
...
+// SetTemporaryDirectoryForBigFiles overwrites default temporary directory.
+func SetTemporaryDirectoryForBigFiles(customTmpDir string) {
+	unixTempDirForBigFiles = customTmpDir
+}
+
+// GetTemporaryDirectoryForBigFiles returns a directory for temporary (big) files.
 // On non Windows systems it avoids the use of os.TempDir(), because the default temporary directory usually falls under /tmp
 // which on systemd based systems could be the unsuitable tmpfs filesystem.
-func TemporaryDirectoryForBigFiles() string {
+func GetTemporaryDirectoryForBigFiles() string {
```
mtrmac: https://golang.org/doc/effective_go.html#Getters , the `Get` prefix is not conventional in Go.

internal/tmpdir/tmpdir.go
```diff
@@ -15,10 +15,15 @@ var unixTempDirForBigFiles = builtinUnixTempDirForBigFiles
 // DO NOT change this, instead see unixTempDirForBigFiles above.
 const builtinUnixTempDirForBigFiles = "/var/tmp"
 
-// TemporaryDirectoryForBigFiles returns a directory for temporary (big) files.
+// SetTemporaryDirectoryForBigFiles overwrites default temporary directory.
+func SetTemporaryDirectoryForBigFiles(customTmpDir string) {
+	unixTempDirForBigFiles = customTmpDir
```
mtrmac: This shouldn’t be a per-library global state, _especially_ not set on per-call private state like `types.SystemContext`.

Instead, just add a `*types.SystemContext` parameter to the `TemporaryDirectoryForBigFiles` function. (That is safe to do because the `internal` package is not a part of the API contract.)

oci/archive/oci_src.go
```diff
@@ -38,12 +38,12 @@ func newImageSource(ctx context.Context, sys *types.SystemContext, ref ociArchiv
 }
 
 // LoadManifestDescriptor loads the manifest
-func LoadManifestDescriptor(imgRef types.ImageReference) (imgspecv1.Descriptor, error) {
+func LoadManifestDescriptor(imgRef types.ImageReference, sys *types.SystemContext) (imgspecv1.Descriptor, error) {
```
mtrmac: This is an API break, and requires a major version bump :(

mtrmac: By convention, the parameters go `context.Context, *types.SystemContext`, then others.

oci/archive/oci_transport.go
```diff
@@ -159,8 +159,11 @@ func (t *tempDirOCIRef) deleteTempDir() error {
 }
 
 // createOCIRef creates the oci reference of the image
-func createOCIRef(image string) (tempDirOCIRef, error) {
-	dir, err := ioutil.TempDir(tmpdir.TemporaryDirectoryForBigFiles(), "oci")
+func createOCIRef(image string, sys *types.SystemContext) (tempDirOCIRef, error) {
```
mtrmac: By convention, the parameters go `context.Context, *types.SystemContext`, then others.

oci/archive/oci_transport.go
```diff
@@ -174,8 +177,8 @@ func createOCIRef(image string) (tempDirOCIRef, error) {
 }
 
 // creates the temporary directory and copies the tarred content to it
-func createUntarTempDir(ref ociArchiveReference) (tempDirOCIRef, error) {
-	tempDirRef, err := createOCIRef(ref.image)
+func createUntarTempDir(ref ociArchiveReference, sys *types.SystemContext) (tempDirOCIRef, error) {
```
mtrmac: By convention, the parameters go `context.Context, *types.SystemContext`, then others.

storage/storage_image.go
```diff
@@ -341,8 +341,11 @@ func (s *storageImageSource) GetSignatures(ctx context.Context, instanceDigest *
 
 // newImageDestination sets us up to write a new image, caching blobs in a temporary directory until
 // it's time to Commit() the image
-func newImageDestination(imageRef storageReference) (*storageImageDestination, error) {
-	directory, err := ioutil.TempDir(tmpdir.TemporaryDirectoryForBigFiles(), "storage")
+func newImageDestination(imageRef storageReference, sys *types.SystemContext) (*storageImageDestination, error) {
```
mtrmac: By convention, the parameters go `context.Context, *types.SystemContext`, then others.

types/types.go
```diff
@@ -545,6 +545,9 @@ type SystemContext struct {
 	CompressionFormat *compression.Algorithm
 	// CompressionLevel specifies what compression level is used
 	CompressionLevel *int
+
+	// Allows overwrite default temporary directory for storing big files,
+	BigFilesTemporaryDir string
```
mtrmac: - This should be in the “global configuration overrides” section.
- The comment should be `If not "", overrides the temporary directory to use for storing big files` or something like that, to be consistent with the others and to clearly how to not override the default.

docker/tarfile/dest.go
```diff
@@ -99,7 +99,7 @@ func (d *Destination) PutBlob(ctx context.Context, stream io.Reader, inputInfo t
 	// When the layer is decompressed, we also have to generate the digest on uncompressed datas.
 	if inputInfo.Size == -1 || inputInfo.Digest.String() == "" {
 		logrus.Debugf("docker tarfile: input with unknown size, streaming to disk first ...")
-		streamCopy, err := ioutil.TempFile(tmpdir.TemporaryDirectoryForBigFiles(), "docker-tarfile-blob")
+		streamCopy, err := ioutil.TempFile(tmpdir.GetTemporaryDirectoryForBigFiles(), "docker-tarfile-blob")
```
mtrmac: This also needs to be enhanced to be overridable, probably by passing a `*types.SystemContext` to `NewDestination`.

… except that it would be an API break.

docker/tarfile/src.go
```diff
@@ -74,7 +74,7 @@ func NewSourceFromFile(path string) (*Source, error) {
 func NewSourceFromStream(inputStream io.Reader) (*Source, error) {
 	// FIXME: use SystemContext here.
 	// Save inputStream to a temporary file
-	tarCopyFile, err := ioutil.TempFile(tmpdir.TemporaryDirectoryForBigFiles(), "docker-tar")
+	tarCopyFile, err := ioutil.TempFile(tmpdir.GetTemporaryDirectoryForBigFiles(), "docker-tar")
```
mtrmac: This also needs to be enhanced to be overridable, by adding an extra parameter to `NewSourceFromStream`.

… except that it would be an API break.

Zyqsempai:
@mtrmac Thank you for your review, PTAL.

internal/tmpdir/tmpdir.go
```diff
@@ -18,7 +20,10 @@ const builtinUnixTempDirForBigFiles = "/var/tmp"
 // TemporaryDirectoryForBigFiles returns a directory for temporary (big) files.
 // On non Windows systems it avoids the use of os.TempDir(), because the default temporary directory usually falls under /tmp
 // which on systemd based systems could be the unsuitable tmpfs filesystem.
-func TemporaryDirectoryForBigFiles() string {
+func TemporaryDirectoryForBigFiles(sys *types.SystemContext) string {
+	if sys != nil && sys.BigFilesTemporaryDir != "" {
+		unixTempDirForBigFiles = sys.BigFilesTemporaryDir
```
mtrmac: - This still overwrites a global variable; it shouldn’t.
- Why not use the value on Windows as well?

i.e. in here, just `return sys.BigFilesTemporaryDir`.

oci/archive/oci_transport.go
```diff
@@ -159,8 +159,16 @@ func (t *tempDirOCIRef) deleteTempDir() error {
 }
 
 // createOCIRef creates the oci reference of the image
+// Deprecated: Please use createOCIRefWithSystemContext which will allows you to configure temp directory
+// for big files through SystemContext.BigFilesTemporaryDir
 func createOCIRef(image string) (tempDirOCIRef, error) {
```
mtrmac: This is a package-private function, so it can just be modified without maintaining a compatibility shim.

types/types.go
```diff
@@ -494,6 +494,9 @@ type SystemContext struct {
 	// Additional tags when creating or copying a docker-archive.
 	DockerArchiveAdditionalTags []reference.NamedTagged
 
+	// If not "", overrides the temporary directory to use for storing big files
```
mtrmac: (Non-blocking nit: Maybe drop the empty line, and the one below `BlobInfoCacheDir`.)

Zyqsempai:
@mtrmac ping

docker/tarfile/dest.go
```diff
@@ -94,12 +94,26 @@ func (d *Destination) HasThreadSafePutBlob() bool {
...
+}
+
+// PutBlobWithSystemContext writes contents of stream and returns data representing the result (with all data filled in).
+// inputInfo.Digest can be optionally provided if known; it is not mandatory for the implementation to verify it.
+// inputInfo.Size is the expected length of stream, if known.
+// May update cache.
+// WARNING: The contents of stream are being verified on the fly.  Until stream.Read() returns io.EOF, the contents of the data SHOULD NOT be available
+// to any other readers for download using the supplied digest.
+// If stream.Read() at any time, ESPECIALLY at end of input, returns an error, PutBlob MUST 1) fail, and 2) delete any data stored so far.
+// If SystemContext.BigFilesTemporaryDir not "", overrides the temporary directory to use for storing big files
```
mtrmac: (Absolutely nonblocking: `BigFilesTemporaryDir` does not need to be documented here. But it does not hurt either.)

docker/tarfile/dest.go
```diff
@@ -94,12 +94,26 @@ func (d *Destination) HasThreadSafePutBlob() bool {
...
+func (d *Destination) PutBlobWithSystemContext(ctx context.Context, sys *types.SystemContext, stream io.Reader, inputInfo types.BlobInfo, cache types.BlobInfoCache, isConfig bool) (types.BlobInfo, error) {
```
mtrmac: This won’t work; `PutBlob` is a method in the `ImageDestination` interface, and callers won’t call this `PutBlobWithSystemContext`.

The context needs to be passed in `NewDestination`, and recorded in the `Destination` object to be used in `PutBlob`. (So, we also don’t need a new `PutBlobWithSystemContext` method at all.)

docker/tarfile/src.go
```diff
@@ -65,16 +72,26 @@ func NewSourceFromFile(path string) (*Source, error) {
...
+// Deprecated: Please use NewSourceFromStreamWithSystemContext which will allows you to configure
+// temp directory for big files through SystemContext.BigFilesTemporaryDir
 func NewSourceFromStream(inputStream io.Reader) (*Source, error) {
+	return NewSourceFromStreamWithSystemContext(nil, inputStream)
+}
+
+// NewSourceFromStreamWithSystemContext returns a tarfile.Source for the specified inputStream,
+// which can be either compressed or uncompressed. The caller can close the
+// inputStream immediately after NewSourceFromFile returns.
+// If SystemContext.BigFilesTemporaryDir not "", overrides the temporary directory to use for storing big files
```
mtrmac: (Absolutely nonblocking: `BigFilesTemporaryDir` does not need to be documented here. But it does not hurt either.)

oci/archive/oci_transport.go
```diff
@@ -158,9 +158,10 @@ func (t *tempDirOCIRef) deleteTempDir() error {
 	return os.RemoveAll(t.tempDirectory)
 }
 
-// createOCIRef creates the oci reference of the image
-func createOCIRef(image string) (tempDirOCIRef, error) {
-	dir, err := ioutil.TempDir(tmpdir.TemporaryDirectoryForBigFiles(), "oci")
+// createOCIRefWithSystemContext creates the oci reference of the image
+// If SystemContext.BigFilesTemporaryDir not "", overrides the temporary directory to use for storing big files
+func createOCIRefWithSystemContext(sys *types.SystemContext, image string) (tempDirOCIRef, error) {
```
mtrmac: This can keep the older `createOCIRef` name.

oci/archive/oci_transport.go
```diff
@@ -175,7 +176,7 @@ func createOCIRef(image string) (tempDirOCIRef, error) {
 
 // creates the temporary directory and copies the tarred content to it
 func createUntarTempDir(ref ociArchiveReference) (tempDirOCIRef, error) {
-	tempDirRef, err := createOCIRef(ref.image)
+	tempDirRef, err := createOCIRefWithSystemContext(nil, ref.image)
```
mtrmac: This should pass along a real context instead of `nil`.

(Apparently that already existed, with `LoadManifestDescriptor` modified previously. Where did it disappear to?)

storage/storage_image.go
```diff
@@ -341,8 +341,9 @@ func (s *storageImageSource) GetSignatures(ctx context.Context, instanceDigest *
 
 // newImageDestination sets us up to write a new image, caching blobs in a temporary directory until
 // it's time to Commit() the image
-func newImageDestination(imageRef storageReference) (*storageImageDestination, error) {
-	directory, err := ioutil.TempDir(tmpdir.TemporaryDirectoryForBigFiles(), "storage")
+// If SystemContext.BigFilesTemporaryDir not "", overrides the temporary directory to use for storing big files
```
mtrmac: (Absolutely nonblocking: `BigFilesTemporaryDir` does not need to be documented here. But it does not hurt either.)

Zyqsempai:
@mtrmac Done, PTAL

docker/tarfile/dest.go
```diff
@@ -29,6 +29,7 @@ type Destination struct {
 	// Other state.
 	blobs  map[digest.Digest]types.BlobInfo // list of already-sent blobs
 	config []byte
+	sysCtx *types.SystemContext
```
mtrmac: Nothing initializes this; this needs to be set in a new `NewDestinationWithContext` (and callers updated)

docker/tarfile/dest.go
```diff
@@ -94,12 +95,14 @@ func (d *Destination) HasThreadSafePutBlob() bool {
 // WARNING: The contents of stream are being verified on the fly.  Until stream.Read() returns io.EOF, the contents of the data SHOULD NOT be available
 // to any other readers for download using the supplied digest.
 // If stream.Read() at any time, ESPECIALLY at end of input, returns an error, PutBlob MUST 1) fail, and 2) delete any data stored so far.
+// Deprecated: Please use PutBlobWithSystemContext which will allows you to configure temp directory
+// for big files through SystemContext.BigFilesTemporaryDir
```
mtrmac: This comment no longer applies, there is no `PutBlobWithSystemContext`.

oci/archive/oci_src.go
```diff
@@ -43,7 +43,7 @@ func LoadManifestDescriptor(imgRef types.ImageReference) (imgspecv1.Descriptor,
 	if !ok {
 		return imgspecv1.Descriptor{}, errors.Errorf("error typecasting, need type ociArchiveReference")
 	}
-	tempDirRef, err := createUntarTempDir(ociArchRef)
+	tempDirRef, err := createUntarTempDir(nil, ociArchRef)
```
mtrmac: This should pass a real context instead of `nil`, https://github.com/containers/image/pull/747#discussion_r346954903 still outstanding.

Zyqsempai: The problem is that `LoadManifestDescriptor` is not used, so i see not way how to pass the systemContext here

mtrmac: `LoadManifestDescriptor` exists to be called from other software, notably https://github.com/containers/libpod/blob/c673ff8cb63c71a39e1a1d561373af42bfea7f24/libpod/image/pull.go#L178 (which is what #495 wants to improve).

The technique should be the same as for other public functions - introduce new `LoadManifestDescriptorWithContext`, provide a compatibility wrapper `LoadManifestDescriptor` that calls the new one (and eventually update Podman to call the new one and benefit from the feature).

rhatdan:
LGTM

mtrmac:
@rhatdan please don’t merge PRs that haven’t passed tests.

rhatdan:
I guess I mis interpreted your message.

PR #752, opened 2019-11-14
Untested: Enforce runtime architecture

mtrmac:
When pulling images for consumption by the current runtime, reject images that target a different architecture: https://bugzilla.redhat.com/show_bug.cgi?id=1753019 .

WIP:
- Absolutely untested, beyond that it compiles.
- I have no idea how this interacts with manifest lists yet.
- We might want to look for images that would be broken by this (is it possible to run i386 images on x86_64?)
- This should ideally be tested directly in CRI-O to ensure it behaves reasonably, and the error is propagated all the way up to the pod status.
- I’d really love to hear opinions on the semantics change of `MustMatchRuntimeOS`.

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -663,7 +663,7 @@ func checkImageDestinationForCurrentRuntimeOS(ctx context.Context, sys *types.Sy
 			wantedOS = sys.OSChoice
 		}
 		if (wantedOS == "windows" && c.OS == "linux") || (wantedOS != "windows" && c.OS == "windows") {
-			return fmt.Errorf("image operating system %q cannot be used on %q", c.OS, wantedOS)
+			return fmt.Errorf("Image operating system mismatch: image uses %q, expecting %q", c.OS, wantedOS)
 		}
 	}
 	return nil
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -662,7 +662,7 @@ func checkImageDestinationForCurrentRuntimeOS(ctx context.Context, sys *types.Sy
 		if sys != nil && sys.OSChoice != "" {
 			wantedOS = sys.OSChoice
 		}
-		if (wantedOS == "windows" && c.OS == "linux") || (wantedOS != "windows" && c.OS == "windows") {
+		if wantedOS != c.OS {
 			return fmt.Errorf("Image operating system mismatch: image uses %q, expecting %q", c.OS, wantedOS)
 		}
 	}
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -524,7 +524,7 @@ func (c *copier) copyOneImage(ctx context.Context, policyContext *signature.Poli
 		}
 	}
 
-	if err := checkImageDestinationForCurrentRuntimeOS(ctx, options.DestinationCtx, src, c.dest); err != nil {
+	if err := checkImageDestinationForCurrentRuntime(ctx, options.DestinationCtx, src, c.dest); err != nil {
 		return nil, "", "", err
 	}
 
@@ -651,7 +651,8 @@ func (c *copier) Printf(format string, a ...interface{}) {
 	fmt.Fprintf(c.reportWriter, format, a...)
 }
 
-func checkImageDestinationForCurrentRuntimeOS(ctx context.Context, sys *types.SystemContext, src types.Image, dest types.ImageDestination) error {
+// checkImageDestinationForCurrentRuntime enforces dest.MustMatchRuntimeOS, if necessary.
+func checkImageDestinationForCurrentRuntime(ctx context.Context, sys *types.SystemContext, src types.Image, dest types.ImageDestination) error {
 	if dest.MustMatchRuntimeOS() {
 		c, err := src.OCIConfig(ctx)
 		if err != nil {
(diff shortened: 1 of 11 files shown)


copy/copy.go
```diff
@@ -651,21 +651,28 @@ func (c *copier) Printf(format string, a ...interface{}) {
...
+		}
+
 		wantedOS := runtime.GOOS
 		if sys != nil && sys.OSChoice != "" {
 			wantedOS = sys.OSChoice
 		}
-		c, err := src.OCIConfig(ctx)
-		if err != nil {
-			return errors.Wrapf(err, "Error parsing image configuration")
+		if (wantedOS == "windows" && c.OS == "linux") || (wantedOS != "windows" && c.OS == "windows") {
```
vrothberg: Why not `if wantedOS != c.OS`?

copy/copy.go
```diff
@@ -651,21 +651,28 @@ func (c *copier) Printf(format string, a ...interface{}) {
 	fmt.Fprintf(c.reportWriter, format, a...)
 }
 
-func checkImageDestinationForCurrentRuntimeOS(ctx context.Context, sys *types.SystemContext, src types.Image, dest types.ImageDestination) error {
+// checkImageDestinationForCurrentRuntime enforces dest.MustMatchRuntimeOS, if necessary.
+func checkImageDestinationForCurrentRuntime(ctx context.Context, sys *types.SystemContext, src types.Image, dest types.ImageDestination) error {
 	if dest.MustMatchRuntimeOS() {
+		c, err := src.OCIConfig(ctx)
+		if err != nil {
+			return errors.Wrapf(err, "Error parsing image configuration")
```
vrothberg: Errors should start with lower caps (for wrapping).

copy/copy.go
```diff
@@ -651,21 +651,28 @@ func (c *copier) Printf(format string, a ...interface{}) {
...
+		if sys != nil && sys.ArchitectureChoice != "" {
+			wantedArch = sys.ArchitectureChoice
 		}
-		osErr := fmt.Errorf("image operating system %q cannot be used on %q", c.OS, wantedOS)
-		if wantedOS == "windows" && c.OS == "linux" {
-			return osErr
-		} else if wantedOS != "windows" && c.OS == "windows" {
-			return osErr
+		if wantedArch != c.Architecture {
+			return fmt.Errorf("Image for architecture %q cannot be used on %q", c.Architecture, wantedArch)
```
vrothberg: Errors should start with lower caps (for wrapping).

copy/copy.go
```diff
@@ -651,21 +651,28 @@ func (c *copier) Printf(format string, a ...interface{}) {
...
+
 		wantedOS := runtime.GOOS
 		if sys != nil && sys.OSChoice != "" {
 			wantedOS = sys.OSChoice
 		}
-		c, err := src.OCIConfig(ctx)
-		if err != nil {
-			return errors.Wrapf(err, "Error parsing image configuration")
+		if (wantedOS == "windows" && c.OS == "linux") || (wantedOS != "windows" && c.OS == "windows") {
+			return fmt.Errorf("image operating system %q cannot be used on %q", c.OS, wantedOS)
```
vrothberg: The error message can be misleading as we could force to exchange the image and host via the SystemContext, such that it claims that the "image operating system "linux" cannot be used on "windows"" (when running a linux host). Maybe we can change the wording to "mismatching image operating system: received %q, expected %q"? Same applies to the architecture.

copy/copy.go
mtrmac: This keeps the previous logic unchanged. I didn’t look into why it is written exactly this way.

copy/copy.go
mtrmac: c/image has always capitalized them (compare the rest of this package), if nothing else because various CLIs have reported the underlying errors without any wrapping/meaningful prefix.

copy/copy.go
mtrmac: c/image has always capitalized them (compare the rest of this package), if nothing else because various CLIs have reported the underlying errors without any wrapping/meaningful prefix.

copy/copy.go
mtrmac: Will fix, though I expect the implementation, and so possibly the reporting, will change drastically after the full scope of #753 .

copy/copy.go
vrothberg: It dates back to https://github.com/containers/image/commit/0c75281602badef6f375ee7b55528f4d50cc6710 but I can't imagine why it was written that way. According to spec, we should understand all values of GOOS (https://golang.org/doc/install/source#environment) which is currently not the case - at least for this specific check.

copy/copy.go
mtrmac: Improved the texts now.

copy/copy.go
vrothberg: This is somehow still outstanding. The logic predates this PR but I'm convinced it's wrong. @mtrmac, if you prefer, I can open a PR after and take ownership of the consequences.

mtrmac: \*shrug\* added a commit to simplify it now:
> Looking back at older images, `docker://busybox:1.21-ubuntu` from 2015 does already have `"os"` and `"architecture"` filled in, so this seems safe enough.

I’m not exactly sure how we are going to test that this PR does not break compatibility with obscure images… but if we figure that out, we might just as well test both changes at once.

rhatdan:
LGTM

PR #753, opened 2019-11-18
Extended SystemContext to include VariantChoice.

grooverdan:
Like ArchitectureChoice and OSChoice, the user has a choice of architecture variant that can be used.

### DIFF
--- a/manifest/docker_schema2_list.go
+++ b/manifest/docker_schema2_list.go
@@ -100,13 +100,17 @@ func (list *Schema2List) ChooseInstance(ctx *types.SystemContext) (digest.Digest
 	if ctx != nil && ctx.OSChoice != "" {
 		wantedOS = ctx.OSChoice
 	}
+	wantedVariant := ""
+	if ctx != nil && ctx.VariantChoice != "" {
+		wantedVariant = ctx.VariantChoice
+	}
 
 	for _, d := range list.Manifests {
-		if d.Platform.Architecture == wantedArch && d.Platform.OS == wantedOS {
+		if d.Platform.Architecture == wantedArch && d.Platform.OS == wantedOS && (d.Platform.Variant == wantedVariant || wantedVariant == "") {
 			return d.Digest, nil
 		}
 	}
-	return "", fmt.Errorf("no image found in manifest list for architecture %s, OS %s", wantedArch, wantedOS)
+	return "", fmt.Errorf("no image found in manifest list for architecture %s(variant \"%s\"), OS %s", wantedArch, wantedVariant, wantedOS)
 }
 
 // Serialize returns the list in a blob format.
--- a/manifest/list_test.go
+++ b/manifest/list_test.go
@@ -100,10 +100,14 @@ func TestChooseInstance(t *testing.T) {
 		require.NoError(t, err)
 		// Match found
 		for arch, expected := range manifestList.matchedInstances {
-			digest, err := list.ChooseInstance(&types.SystemContext{
+			ctx := &types.SystemContext{
 				ArchitectureChoice: arch,
 				OSChoice:           "linux",
-			})
+			}
+			if arch == "arm" {
+				ctx.VariantChoice = "v5"
+			}
+			digest, err := list.ChooseInstance(ctx)
 			require.NoError(t, err, arch)
 			assert.Equal(t, expected, digest)
 		}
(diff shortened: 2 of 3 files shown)


manifest/docker_schema2_list.go
```diff
@@ -100,13 +100,17 @@ func (list *Schema2List) ChooseInstance(ctx *types.SystemContext) (digest.Digest
...
+	}
 
 	for _, d := range list.Manifests {
-		if d.Platform.Architecture == wantedArch && d.Platform.OS == wantedOS {
+		if d.Platform.Architecture == wantedArch && d.Platform.OS == wantedOS && (d.Platform.Variant == wantedVariant || wantedVariant == "") {
 			return d.Digest, nil
 		}
 	}
-	return "", fmt.Errorf("no image found in manifest list for architecture %s, OS %s", wantedArch, wantedOS)
+	return "", fmt.Errorf("no image found in manifest list for architecture %s(variant \"%s\"), OS %s", wantedArch, wantedVariant, wantedOS)
```
TomSweeneyRedHat: nit, could you add a space after the %s and before the new paren?  `%s (variant`

TomSweeneyRedHat:
Some unhappy tests at the moment

mtrmac:
The test failures are unrelated, should be eventually fixed in https://github.com/containers/image/pull/744 and https://github.com/containers/skopeo/pull/760/files .

grooverdan:
> Thanks!
> 
>     * The *`Choice` options are _overrides_; first we should implement the default (choosing the right variant for the current runtime)

Is including https://github.com/containerd/containerd/blob/master/platforms/cpuinfo.go#L76 exposed as [the API](https://github.com/containerd/containerd/blob/master/platforms/defaults.go#L31) acceptable?

>     * This should be implemented for `manifest/oci_index.go` as well.

Quite right. Is in [the spec](https://github.com/opencontainers/image-spec/blob/master/specs-go/v1/descriptor.go#L63)

>     * … and `copy.CheckImageDestinationForCurrentRuntime`* (after #752 ).

Thanks for the heads up, I hadn't checked PRs.

>     * … so, the runtime variant detection, compatibility rules, overrides and matching logic should probably be factored out into a shared helper now that it is bigger.

Like https://github.com/containerd/containerd/blob/master/platforms/compare.go ?

mtrmac:
> > Thanks!
> > ```
> > * The *`Choice` options are _overrides_; first we should implement the default (choosing the right variant for the current runtime)
> > ```
> 
> Is including https://github.com/containerd/containerd/blob/master/platforms/cpuinfo.go#L76 

It’s not like https://github.com/opencontainers/image-spec/blob/master/image-index.md is all that specific about the details. That’s as good a starting point as any other, I guess.

(It does vaguely feel that user-space code should be able to detect CPU variants without OS help via `/proc`, but I know too little about ARM, and if this code works, it can be used just fine until someone cares to it better.)

> exposed as [the API](https://github.com/containerd/containerd/blob/master/platforms/defaults.go#L31) acceptable?

(We’ve found it difficult to keep APIs stable in c/image, so unless strictly necessary, this shared implementation etc. should be an either private within some package, or an internal subpackage (e.g. c/image/internal/…).  With an internal-only API, the API design should be _correct_ but is not _that much_ of a concern for it to be long-term extensible and stable.)

I’m not 100% sure how the compatibility checks (that v6 images can be run on v8 systems, and that v8 images are preferred to v6 images on such systems) should be structured in the API, but, given the above, possibly.

> > ```
> > * … so, the runtime variant detection, compatibility rules, overrides and matching logic should probably be factored out into a shared helper now that it is bigger.
> > ```
> 
> Like https://github.com/containerd/containerd/blob/master/platforms/compare.go ?

Form a quick skim that seems rather over-engineered (`orderedPlatformComparer` / `anyPlatformComparer`)… given a runtime platform, we can have a hard-coded database of platform → list of supported architectures+variants in order of preference (~ a more compact version of the knowledge in `platforms.Only()`), and use that in the straightforward way for both the `CheckImageDestinationForCurrentRuntime` check and image selection from multi-arch lists.

(The `os.version`/`os.features`/`features` field in OCI may require significantly more complex logic, but they are underspecified enough, and the relative ordering of preference if many images match is vague enough, that I’d rather not worry generalizing any of this now. As long as the API is internal, we can always make it more complex; but if it _starts_ complex, we will have to pay the maintenance cost unnecessarily.)

mtrmac:
For the record, another attempt is in https://github.com/containers/image/pull/786 .

mtrmac:
This functionality was added in #854.

PR #760, opened 2019-11-27
Make: split out installing tools

vrothberg:
Split the installation of required tools into dedicated .install.$tool 
targets and only install them if they're absent.  Remove the sudo call 
when installing golangci-lint by properly setting GOBIN.               

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,0 +1,13 @@
+container:
+  image: golang:1.13
+
+env:
+  GOPROXY: https://proxy.golang.org
+
+validate_task:
+  validate_script:
+    - apt-get update -qq
+    - apt-get install -qq -y libgpgme-dev libdevmapper-dev btrfs-tools libbtrfs-dev
+    - make tools
+    - make .gitvalidation
+    - make validate
--- a/Makefile
+++ b/Makefile
@@ -96,7 +96,7 @@ test-skopeo:
 		cd $${skopeo_path} && \
 		GO111MODULE="on" go mod edit -replace $${project_module}=$${project_path} && \
 		make vendor && \
-		make BUILDTAGS="$(BUILDTAGS)" binary-local test-all-local && \
+		make BUILDTAGS="$(BUILDTAGS)" binary-local test-unit-local && \
 		$(SUDO) make BUILDTAGS="$(BUILDTAGS)" check && \
 		rm -rf $${skopeo_path}
 
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,11 @@ SKOPEO_BRANCH = master
 # Set SUDO=sudo to run container integration tests using sudo.
 SUDO =
 
+GOBIN := $(shell go env GOBIN)
+ifeq ($(GOBIN),)
+GOBIN := $(GOPATH)/bin
+endif
+
 # when cross compiling _for_ a Darwin or windows host, then we must use openpgp
 BUILD_TAGS_WINDOWS_CROSS = containers_image_openpgp
 BUILD_TAGS_DARWIN_CROSS = containers_image_openpgp
@@ -23,6 +28,8 @@ MANINSTALLDIR=${PREFIX}/share/man
 GOMD2MAN ?= $(shell command -v go-md2man || echo '$(GOBIN)/go-md2man')
 MANPAGES_MD = $(wildcard docs/*.5.md)
 MANPAGES ?= $(MANPAGES_MD:%.md=%)
+
+export PATH := $(PATH):${GOBIN}
 
 # On macOS, (brew install gpgme) installs it within /usr/local, but /usr/local/include is not in the default search path.
 # Rather than hard-code this directory, use gpgme-config. Sadly that must be done at the top-level user
@@ -52,18 +59,26 @@ cross:
 	GOOS=windows $(MAKE) build BUILDTAGS="$(BUILDTAGS) $(BUILD_TAGS_WINDOWS_CROSS)"
 	GOOS=darwin $(MAKE) build BUILDTAGS="$(BUILDTAGS) $(BUILD_TAGS_DARWIN_CROSS)"
 
-tools: tools.timestamp
+tools: .install.gitvalidation .install.golangci-lint .install.golint
 
-tools.timestamp: Makefile
-	@GO111MODULE="off" go get -u $(BUILDFLAGS) golang.org/x/lint/golint
-	if [ ! -x "$(GOBIN)/golangci-lint" ]; then \
-		curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sudo sh -s -- -b $(GOBIN)/ v1.21.0; \
+.install.gitvalidation:
+	if [ ! -x "$(GOBIN)/git-validation" ]; then \
+		GO111MODULE="off" go get $(BUILDFLAGS) github.com/vbatts/git-validation; \
 	fi
-	@GO111MODULE="off" go get $(BUILDFLAGS) github.com/vbatts/git-validation
-	@touch tools.timestamp
+
+.install.golangci-lint:
+	if [ ! -x "$(GOBIN)/golangci-lint" ]; then \
+		curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s -- -b $(GOBIN) v1.21.0; \
+	fi
+
+.install.golint:
+	# Note, golint is only needed for Skopeo's tests.
+	if [ ! -x "$(GOBIN)/golint" ]; then \
+		GO111MODULE="off" go get -u $(BUILDFLAGS) golang.org/x/lint/golint; \
+	fi
 
 clean:
-	rm -rf tools.timestamp $(MANPAGES)
+	rm -rf $(MANPAGES)
 
 test:
 	@$(GPGME_ENV) GO111MODULE="on" go test $(BUILDFLAGS) -cover ./...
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -52,18 +52,20 @@ cross:
...
-	@GO111MODULE="off" go get -u $(BUILDFLAGS) golang.org/x/lint/golint
+.install.gitvalidation:
+	if [ ! -x "$(GOBIN)/git-validation" ]; then \
+		GO111MODULE="off" go get $(BUILDFLAGS) github.com/vbatts/git-validation; \
+	fi
+
+.install.golangci-lint: .gopathok
 	if [ ! -x "$(GOBIN)/golangci-lint" ]; then \
-		curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sudo sh -s -- -b $(GOBIN)/ v1.21.0; \
+		curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $(GOBIN)/ v1.21.0; \
```
TomSweeneyRedHat: do you still need sudo here before 'sh -s'?

Makefile
```diff
@@ -52,18 +52,20 @@ cross:
...
+tools: .install.gitvalidation .install.golangci-lint
 
-tools.timestamp: Makefile
-	@GO111MODULE="off" go get -u $(BUILDFLAGS) golang.org/x/lint/golint
+.install.gitvalidation:
+	if [ ! -x "$(GOBIN)/git-validation" ]; then \
+		GO111MODULE="off" go get $(BUILDFLAGS) github.com/vbatts/git-validation; \
+	fi
+
+.install.golangci-lint: .gopathok
```
TomSweeneyRedHat: Tests are failing with:
'make: *** No rule to make target '.gopathok', needed by '.install.golangci-lint'.  Stop.'

Makefile
vrothberg: If needed, I want user to use `sudo make ...` over prompting for a PW.

Makefile
vrothberg: Thanks for checking!

Makefile
```diff
@@ -52,18 +52,20 @@ cross:
 	GOOS=windows $(MAKE) build BUILDTAGS="$(BUILDTAGS) $(BUILD_TAGS_WINDOWS_CROSS)"
 	GOOS=darwin $(MAKE) build BUILDTAGS="$(BUILDTAGS) $(BUILD_TAGS_DARWIN_CROSS)"
 
-tools: tools.timestamp
+tools: .install.gitvalidation .install.golangci-lint
 
-tools.timestamp: Makefile
-	@GO111MODULE="off" go get -u $(BUILDFLAGS) golang.org/x/lint/golint
+.install.gitvalidation:
+	if [ ! -x "$(GOBIN)/git-validation" ]; then \
```
mtrmac: This is not quite a new issue, but what guarantees `GOBIN` is set?

Makefile
```diff
@@ -52,18 +52,20 @@ cross:
 	GOOS=windows $(MAKE) build BUILDTAGS="$(BUILDTAGS) $(BUILD_TAGS_WINDOWS_CROSS)"
 	GOOS=darwin $(MAKE) build BUILDTAGS="$(BUILDTAGS) $(BUILD_TAGS_DARWIN_CROSS)"
 
-tools: tools.timestamp
+tools: .install.gitvalidation .install.golangci-lint
 
-tools.timestamp: Makefile
-	@GO111MODULE="off" go get -u $(BUILDFLAGS) golang.org/x/lint/golint
```
mtrmac: So, this removes the installation of `golint`? Would be nice to mention it in the description and commit message.

vrothberg: Sure can add :+1:

Makefile
vrothberg: Added some checks to make it more robust

vrothberg:
Either I don't take enough time to think this through and figure out what's going wrong; I'll focus on only running golangci-lint in the CI now. Some larger refactoring would be nice in the future but I don't have time for a whack-a-mole at the moment.

vrothberg:
Don't merge. Cirrus is not yet firing.

vrothberg:
That wasn't pretty but oh well, it's done.

@rhatdan @giuseppe @TomSweeneyRedHat PTAL. Please refer to the individual commit messages.

.cirrus.yml
```diff
@@ -0,0 +1,13 @@
...
+env:
+  GOPROXY: https://proxy.golang.org
+
+validate_task:
+  validate_script:
+    - apt-get update -qq
+    - apt-get install -qq -y libgpgme-dev libdevmapper-dev btrfs-tools libbtrfs-dev
+    - make tools
+    - make .gitvalidation
+    - make validate
```
mtrmac: Moving to Cirrus is a pretty big deal, not even mentioning it in the CI description is rather sneaky. Do I understand correctly that we now need _both_ systems to work to keep the current test coverage?

Can you explain how the move to Cirrus helps? Naively, both are Ubuntu-based containers with roughly similar configuration, aren’t they? (I’m sure the many iterations showed interesting or bizarre failures, but none of that knowledge is preserved in this PR.)

Especially the move of `.givalidation`, where, AFAICS, nothing changes about the _implementation_ of the test, and the previous Travis system seemed to work fine (didn’t it?), is rather curious.

(I’m not at all _insisting_ on Travis, it’s just that from the PR alone, it’s hard to tell whether there is really something wrong with Travis or whether all of this only makes a difference because it happens to paper over  some not-well-understood problem that is going to bite us even worse soon.)

.travis.Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM ubuntu:18.04
 
 RUN apt-get -qq update && \
-    apt-get install -y sudo docker.io git make btrfs-tools libdevmapper-dev libgpgme-dev libostree-dev curl
+    apt-get install -y sudo golint docker.io git make btrfs-tools libdevmapper-dev libgpgme-dev libostree-dev curl
```
mtrmac: Is the addition of `golint` necessary? The `Makefile` is still installing that, isn‘t it?

Makefile
```diff
@@ -8,6 +8,12 @@ SKOPEO_BRANCH = master
 # Set SUDO=sudo to run container integration tests using sudo.
 SUDO =
 
+GO ?= go
```
mtrmac: So much of the `Makefile` uses `go` directly, do we need this variable? If we do, in which cases it should / shouldn’t be used?

Makefile
```diff
@@ -52,18 +60,26 @@ cross:
...
-	@GO111MODULE="off" go get -u $(BUILDFLAGS) golang.org/x/lint/golint
+.install.golangci-lint:
 	if [ ! -x "$(GOBIN)/golangci-lint" ]; then \
-		curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sudo sh -s -- -b $(GOBIN)/ v1.21.0; \
+		curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s -- -b $(GOBIN) v1.21.0; \
+	fi
+
+.install.golint:
+	# Note, golint is only needed for Skopeo's tests.
+	if [ ! -x "$(GOBIN)/golangci-lint" ]; then \
```
mtrmac: This should probably refer to `golint`.

Makefile
```diff
@@ -81,7 +97,7 @@ test-skopeo:
 		cd $${skopeo_path} && \
 		GO111MODULE="on" go mod edit -replace $${project_module}=$${project_path} && \
 		make vendor && \
-		make BUILDTAGS="$(BUILDTAGS)" binary-local test-all-local && \
+		make BUILDTAGS="$(BUILDTAGS)" binary-local test-unit-local && \
```
mtrmac: Doesn’t removing the Skopeo integration + system tests pretty much defeat the point of `test-skopeo`? What tests that the code works against real-world registries now?

pkg/docker/config/config_linux.go
```diff
@@ -4,7 +4,7 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/containers/image/v5/internal/pkg/keyctl"
+	"github.com/containers/image/v5/internal/pkg/keyctl" //nolint
```
mtrmac: Now that the subpackage is internal anyway, and this is the only user, maybe we can just remove the “Deprecated:” comment instead.

Makefile
vrothberg: `test-all-local` is triggering `validate-local` and `test-unit-local` with
```Makefile
validate-local:
        hack/make.sh validate-git-marks validate-gofmt validate-lint validate-vet
```
Maybe I am missing something but I don't see the integration tests being triggered by `make.sh` as we specify arguments.

mtrmac: Oh, you’re right, sorry. The important tests are run in `make check` one line below, I just reacted to the `all` word without reading the details.

(I’m still mildly curious why `validate-local` broke, but it’s not nearly as important.)

pkg/docker/config/config_linux.go
vrothberg: Will do, thanks!

mtrmac:
> notably the restriction of `test-skopeo`, are rather worrying.

(As discussed elsewhere, this was my mistake; `test-skopeo` still does what it needs to do.)

Makefile
vrothberg: It better should :)

Makefile
vrothberg: Valid point. Will drop it and use `go` for setting `GOBIN`.

.travis.Dockerfile
vrothberg: Not necessary anymore, thanks!

.cirrus.yml
vrothberg: > Moving to Cirrus is a pretty big deal, not even mentioning it in the CI description is rather sneaky. Do I understand correctly that we now need _both_ systems to work to keep the current test coverage?

Yes, both need to work. I see using Cirrus as a _first_ step to become less dependent on Travis. Most containers/ projects are using Cirrus already and share a common GCP billing where we have _full_ control over the VM images. However, using Container tasks is for free in Cirrus, so I wanted to break the validation steps out into a dedicated task. I tried running golangci-lint in Travis but failed (with our setup) - more below.

> Can you explain how the move to Cirrus helps? Naively, both are Ubuntu-based containers with roughly similar configuration, aren’t they? (I’m sure the many iterations showed interesting or bizarre failures, but none of that knowledge is preserved in this PR.)

Some reasons are mentioned above. Both are container based and both run Ubuntu-based images, so there is no functional change yet. I could have very well also added another task to Travis but given the containers/ umbrella is all on Cirrus and that we can have full control over the VMs there, I see this as a _first step_ toward Cirrus.

> Especially the move of `.gitvalidation`, where, AFAICS, nothing changes about the _implementation_ of the test, and the previous Travis system seemed to work fine (didn’t it?), is rather curious.

The main driver for Cirrus was that I was unable to make golangci-lint run. The other PR didn't actually enable it as it silently failed. While trying to do that, I hit all kinds of errors in our setup which I attribute mostly to the (rather exotic?) way we run the container. Once I managed to setup a home dir, sudo fell apart and didn't manage to fix it. After a while I found the current setup to fragile to extend and decided to break out the validation step.

Ultimately, it's very easy to setup new (container-based) tasks in Cirrus that can run in parallel or in sequence with specified dependencies.

> (I’m not at all _insisting_ on Travis, it’s just that from the PR alone, it’s hard to tell whether there is really something wrong with Travis or whether all of this only makes a difference because it happens to paper over some not-well-understood problem that is going to bite us even worse soon.)

In a nutshell: I tried enabling golangci-lint but hit too many issues wiring that in our Travis setup. Breaking it out into a separate task made sense to me at some point; not only to have a clearer separation of CI tasks (e.g., validation, unit testing, integration testing) but also to not invest _too_ much time into this rather small linting adventure.

mtrmac: > However, using Container tasks is for free in Cirrus, so I wanted to break the validation steps out into a dedicated task.

Thanks, that makes sense.

If the `.gitvalidation` step is for cleanliness, not to work around breakage (and now that there is a written record of it), ACK to the general concept of this PR. (I haven’t reviewed / studied the Cirrus setup it in detail.)

vrothberg:
Acked by @mtrmac (https://github.com/containers/image/pull/760#discussion_r353275046), so I'll go ahead and merge.

PR #762, opened 2019-11-28
storage_test.go: check storage interfaces

vrothberg:
Revert a previous change that removed the checks for storage interfaces
as golangci-lint complained about them.  Those checks are still desired
to catch errors early in case storage doesn't implement those
interfaces.  Silence golangci-lint by explicitly ingoring any lint
report for these specific lines.

### DIFF
--- a/storage/storage_test.go
+++ b/storage/storage_test.go
@@ -35,7 +35,11 @@ import (
 )
 
 var (
-	topwd = ""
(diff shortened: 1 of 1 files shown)


vrothberg:
@mtrmac PTAL

mtrmac:
@vrothberg Thanks!

PR #769, opened 2019-12-06
Add registries warning to registries.conf

TomSweeneyRedHat:
Add a warning about fully specifying registries
to the search list.

### DIFF
--- a/registries.conf
+++ b/registries.conf
@@ -8,6 +8,27 @@
 #
 # Registries to search for images that are not fully-qualified.
 # i.e. foobar.com/my_image:latest vs my_image:latest
+#
+# NOTE: RISK OF USING UNQUALIFIED IMAGE NAMES.
+# Pulling an image that is not fully qualified, i.e., one that includes the
+# image name but does not include the registry or  tag, is not recommended.
+# There is a risk that the image being pulled could be spoofed. An example
+# of this would be if a user wanted to pull an image named `foobar` from a
+# registry and expect it to come from myregistry.com.  If myregistry.com is
+# not first in the search list, an attacker could place a different `foobar`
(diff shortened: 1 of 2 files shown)


TomSweeneyRedHat:
@rhatdan PTAL

registries.conf
```diff
@@ -8,6 +8,11 @@
 #
 # Registries to search for images that are not fully-qualified.
 # i.e. foobar.com/my_image:latest vs my_image:latest
+# NOTE: pulling an image that is not fully qualified, i.e. one that
```
rhatdan: I think this needs more description.  Let's use a google doc to start discussion.

mtrmac:
I’m perfectly fine with adding a warning here, but note that the vast majority of users of short names will never read this config file, so warning _only_ here would not be sufficient.

rhatdan:
Agreed, we also need to add this to container-registires.conf.md man page.

TomSweeneyRedHat:
As the comments in the googledoc have died down, I've just updated this PR with the latest changes from there.  I did make some line feed changes for it to fit under 80 char, but left the wording intact.

TomSweeneyRedHat:
And I think I made it past the validation test finally.

rhatdan:
Can you add the same information into the man page for container-registries.conf.5.md

vrothberg:
Once we consolidated the search-code into c/image, I'd love to throw a warning in the logs for that as well.

TomSweeneyRedHat:
I've added to the man page too.  FWIW, I made one small change, at the start of the second paragraph we had "Note that it is recommended".  I changed that to "It is recommended".  We'd used the word in another place or two and it just sounded klunky to me.

rhatdan:
LGTM

PR #770, opened 2019-12-09
Provide additional OverrideManifestMIMETypeList option for copy.Image

sseago:
While the existing ForceManifestMIMEType option forces a single, specific
MIME type for the destination image, OverrideManifestMIMETypeList allows
for a list of one or more MIME types. Passing in a list with a single
mime type is equivalent to setting ForceManifestMIMEType, but this allows
for narrowing the set of MIME types used while not requiring conversion
for images which use MIME types in the list. For example, this would allow
for copying images with forced conversion of V2Schema1 images while allowing
both OCI and V2Schema2 images to be preserved as-is.

The particular use case of immediate concern is to support copying images
to a v2.7.1 Docker registry with Schema1 disabled. Without overriding
MIME types, schema1 images fail with a 500 error from the registry.

To copy an image to a v2.7.1 docker registry which only converts schema1 images,
leaving all other images unchanged, the following would be added to copy.Options:

	OverrideManifestMIMETypeList: []string{
		imgspecv1.MediaTypeImageManifest,
		manifest.DockerV2Schema2MediaType,
		imgspecv1.MediaTypeImageIndex,
		manifest.DockerV2ListMediaType,
	},

Signed-off-by: Scott Seago <sseago@redhat.com>

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -168,8 +168,11 @@ type Options struct {
 	Progress         chan types.ProgressProperties // Reported to when ProgressInterval has arrived for a single artifact+offset.
 	// manifest MIME type of image set by user. "" is default and means use the autodetection to the the manifest MIME type
 	ForceManifestMIMEType string
-	ImageListSelection    ImageListSelection // set to either CopySystemImage (the default), CopyAllImages, or CopySpecificImages to control which instances we copy when the source reference is a list; ignored if the source reference is not a list
-	Instances             []digest.Digest    // if ImageListSelection is CopySpecificImages, copy only these instances and the list itself
+	// manifest MIME type list for image set by user. Ignored if ForceManifestMIMEType is set. Setting to a single value
+	// is equivalent to setting ForceManifestMIMEType. If source manifest MIME type is in list, it takes priority
+	OverrideManifestMIMETypeList []string
+	ImageListSelection           ImageListSelection // set to either CopySystemImage (the default), CopyAllImages, or CopySpecificImages to control which instances we copy when the source reference is a list; ignored if the source reference is not a list
+	Instances                    []digest.Digest    // if ImageListSelection is CopySpecificImages, copy only these instances and the list itself
 	// If OciEncryptConfig is non-nil, it indicates that an image should be encrypted.
 	// The encryption options is derived from the construction of EncryptConfig object.
 	// Note: During initial encryption process of a layer, the resultant digest is not known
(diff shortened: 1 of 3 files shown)


copy/copy.go
```diff
@@ -383,6 +386,9 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
 
 	// Determine if we'll need to convert the manifest list to a different format.
 	forceListMIMEType := options.ForceManifestMIMEType
+	if forceListMIMEType == "" && len(options.OverrideManifestMIMETypeList) == 1 {
+		forceListMIMEType = options.OverrideManifestMIMETypeList[0]
```
mtrmac: This special case does not make much sense to me — if a single-element array affects `forceListMIMEType`, why does two-element array not do so?

“`OverrideManifestMIMETypeList` overrides the transport-determined list of possible options, but does not change the selection logic otherwise” seems a much simpler story to me.

mtrmac:
For the record, the related issue is https://github.com/containers/skopeo/issues/651 .

Ideally, we should be able to detect such registries automatically and not introduce any options at all; last time I have looked into this (per the linked issue), that was not possible.

If we _must_ have an option, at this point I’d prefer a Docker-destination-specific option in `types.SystemContext` affecting the return value of `dockerImageDestination.SupportedManifestMIMETypes`.

(@vrothberg WDYT?)

copy/copy.go
sseago: This is one part that I was less clear on. I wanted to make sure that providing a single value in `OverrideManifestMIMETypeList` was equivalent to setting `ForceManifestMIMEType` which required this. I considered a more general approach here. It seems like the more general approach would be that if the list included *only* Docker types, then we'd want to set this to `manifest.DockerV2ListMediaType`, and if the list included *only* OCI types, then we'd set it to `imgspecv1.MediaTypeImageIndex`. I decided not to do this since I realized that if `dest.SupportedManifestMIMETypes` limits to a similar subset, `forceListMIMEType` doesn't get set.

I could just eliminate this, leaving the list conversion alone, although then a single element override list would no longer be equivalent to setting `ForceManifestMIMEType`. Perhaps that's OK, though.

sseago:
I'm not sure exactly how to approach a docker-specific way of modifying `SupportedManifestMIMETypes`. Adding an element to `copy.Options` for `DockerIncludeSchema1` is clear enough, but is there an existing pattern for passing optional configuration into image destinations? We could do it in the args to `SupportedManifestMIMETypes`, adding an options arg which would be empty for everything but dockerImageDestination, where it would optionally include this boolean. Otherwise we'd need to get this into the image destination itself via `newImageDestination`.

mtrmac:
See how e.g. `DirForceCompress` or `OSTreeTmpDirPath` work.

mtrmac:
> If `dockerImageDestination` was an exported type, I could have done it all in my application code by implementing a custom `ImageDestination` type which delegated to `dockerImageDestination` for everything except `SupportedManifestMIMETypes`.

In principle, it should be possible to create an `ImageReference` implementation that implements `NewImageDestination` by wrapping the result of a `docker.Transport.NewReference().NewImageDestination()`. But in order to work with signatures and UI and the like, it would have to call itself `docker`, making it difficult to differentiate between the c/image implementation and the wrapper. I’d rather add an option to this library so that such confusing wrappers are not necessary — just noting that this should be possible in principle, and there are (sadly?) some precedents to this.

sseago:
https://github.com/containers/image/pull/771 implements the docker-specific configuration for disabling Schema1 destination MIME types approach. If that's a better way forward than this PR, then this one can be closed.

vrothberg:
Looks like we're going with your alternative solution in #771, so I am closing this one to avoid confusion. Thanks a lot for working on this!

PR #771, opened 2019-12-10
Allow for disabling schema1 MIME types for docker destinations

sseago:
This commit adds DockerDisableDestSchema1MIMETypes to types.Context
which, if set to true, will omit DockerV2Schema1SignedMediaType and
DockerV2Schema1MediaType from
dockerImageDestination.SupportedManifestMIMETypes.

Signed-off-by: Scott Seago <sseago@redhat.com>

### DIFF
--- a/docker/docker_image_dest.go
+++ b/docker/docker_image_dest.go
@@ -58,14 +58,16 @@ func (d *dockerImageDestination) Close() error {
 }
 
 func (d *dockerImageDestination) SupportedManifestMIMETypes() []string {
-	return []string{
+	mimeTypes := []string{
 		imgspecv1.MediaTypeImageManifest,
 		manifest.DockerV2Schema2MediaType,
 		imgspecv1.MediaTypeImageIndex,
 		manifest.DockerV2ListMediaType,
-		manifest.DockerV2Schema1SignedMediaType,
-		manifest.DockerV2Schema1MediaType,
 	}
+	if d.c.sys == nil || !d.c.sys.DockerDisableDestSchema1MIMETypes {
+		mimeTypes = append(mimeTypes, manifest.DockerV2Schema1SignedMediaType, manifest.DockerV2Schema1MediaType)
+	}
+	return mimeTypes
 }
 
 // SupportsSignatures returns an error (to be displayed to the user) if the destination certainly can't store signatures.
--- a/types/types.go
+++ b/types/types.go
@@ -547,6 +547,8 @@ type SystemContext struct {
 	// Note that this field is used mainly to integrate containers/image into projectatomic/docker
 	// in order to not break any existing docker's integration tests.
 	DockerDisableV1Ping bool
+	// If true, dockerImageDestination.SupportedManifestMIMETypes will omit the Schema1 media types from the supported list
+	DockerDisableDestSchema1MIMETypes bool
 	// Directory to use for OSTree temporary files
 	OSTreeTmpDirPath string
 


docker/docker_image_dest.go
```diff
@@ -58,14 +58,16 @@ func (d *dockerImageDestination) Close() error {
...
-	return []string{
+	mimeTypes := []string{
 		imgspecv1.MediaTypeImageManifest,
 		manifest.DockerV2Schema2MediaType,
 		imgspecv1.MediaTypeImageIndex,
 		manifest.DockerV2ListMediaType,
-		manifest.DockerV2Schema1SignedMediaType,
-		manifest.DockerV2Schema1MediaType,
 	}
+	if !d.c.sys.DockerDisableDestSchema1MIMETypes {
```
mtrmac: `if d.c.sys == nil || !d.c.…`

types/types.go
```diff
@@ -547,6 +547,8 @@ type SystemContext struct {
 	// Note that this field is used mainly to integrate containers/image into projectatomic/docker
 	// in order to not break any existing docker's integration tests.
 	DockerDisableV1Ping bool
+	// if true, dockerImageDestination.SupportManifestMIMETypes will omit the Schema1 media types from the supported list
+	DockerDisableDestSchema1MIMETypes bool
```
mtrmac: `s/Support/Supported/; s/if/If/`, please.

mtrmac:
LGTM (but I’d like to hear @vrothberg ’s thoughts).

The test failure probably requires a `gofmt -s -w` pass on the file.

sseago:
Ahh, yes I'll fix that. I ran gofmt for the earlier PR but forgot it with this one.

mtrmac:
I don’t know.

The way I think about it, the user wants to say “schema1 is broken”, and schema1 is _two_ MIME types. Forcing the user to list two explicitly, or to even know about two existing, is not great — really this PR compared to #770 was intended to let the user express the intent instead of the implementation details and hard-code implementation knowledge. (Sure, callers of c/image like Skopeo can hide the existence of two different MIME types… but then why have the distinction in the c/image API).

OTOH, we do run into registries that pretend to accept OCI and fail (but then again, those might be easier to detect than the schema1-rejecting ones, I don’t think we have looked into that in detail), so “schema1 and OCI are both broken on this registry” might be useful.

In general I lean towards being specific and under-generalizing until there is a clear pattern (API specifically targeting a single weird corner case is easier to transparently re-implement in a completely different backend implementation; a fully-general API that may be used for many, possibly unknown, use cases, essentially forces that design to be a part of any future backend implementations in the full generality. A specific API is therefore better if the problem space is not completely understood or likely to evolve), but that’s a vague weak preference, not a clear well-agreed-upon best practice.

So, I’m leaning towards the current special-purpose boolean, but I don’t feel strongly about it and I may well be missing something.

sseago:
So from the last comments, @vrothberg suggested an alternative approach but @mtrmac preferred things as-is in the PR. Any final call as to whether this is fine as-is or whether I should implement the suggested alternative?

vrothberg:
> So from the last comments, @vrothberg suggested an alternative approach but @mtrmac preferred things as-is in the PR. Any final call as to whether this is fine as-is or whether I should implement the suggested alternative?

I am totally fine with merging as is. Thanks a lot for double checking!

mtrmac:
Thanks again!

Issue #775, opened 2019-12-12
docker.io registry rewrites still apply the `library` suffix


saschagrunert:
Hey, I'm facing an issue where I'd like to discuss the solution approach before actually fixing the problem.

Let's assume that users push their docker.io mirrored images to `localhost:5000/docker.io` 

Then they rewrite a registry like this:
```toml
[[registry]]
prefix = "docker.io"
location = "localhost:5000/docker.io"
```

Now they expect that images pulled from `docker.io/image` get rewritten to `localhost:5000/docker.io/image`. Unfortunately, the internal docker ref normalization translates it to `localhost:5000/docker.io/library/image`.

What do we want to do? I see that we have this behavior everywhere when it comes to docker references so we might just document it appropriately. WDYT?

mtrmac:
Yes, `prefix` (quite intentionally) does not do the `docker.io[/library]` normalization.
This should work, in general, if configured as
```
# For non-library namespaces
[[registry]]
prefix = "docker.io"
location = "localhost:5000/docker.io"

# The library namespace is more specific, so this stanza is preferred to the above
[[registry]]
prefix = "docker.io/library"
location = "localhost:5000/docker.io/library"
```
(and making appropriate modifications to the mirror’s configuration and mirroring scripts); does it?

---

Looking at the two-item example above, this also shows why we can’t drop `library` automatically: there isn’t a consistently correct way to mirror _both_ `docker.io/library/busybox` and `docker.io/busybox/foo`: if the mapping placed `docker.io/busybox/foo` to `localhost:5000/docker.io/busybox/foo`, but `docker.io/library/busybox` to `localhost:5000/docker.io/busybox`, the two hierarchies/administrative domains would be mixed.

(In practice, that is not _quite_ impossible right now because it isn’t possible to use a `docker://` reference to the `docker.io/[not library]/busybox` repository, so there isn’t currently an actual conflict in image _identities_, but it makes it impossible to e.g. treat the `library` set of images and a top-level `busybox` namespace on the mirror as separate, with e.g. separate admins / mirror scripts / permissions. And it could get worse in the future, AFAIK in principle nothing prevents `docker.io/library/busybox/foo` from existing, and at that point the conflict would be irreconcilable.)

---

I do agree this should be properly documented; it’s kind of implied reading between the lines if the reader knows about the `docker.io/library` mechanism at all, but many probably don’t.

mtrmac:
> ```
> # For non-library namespaces
> [[registry]]
[...]

PR #777, opened 2019-12-13
Add documentation about rewriting docker.io registries

saschagrunert:
We now document the already existing internal `/library` suffix for
docker.io mirrors and provide an example how to deal with them.

I also fixed two typos in `containers-registries.d.5.md`.

Closes https://github.com/containers/image/issues/775

### DIFF
--- a/docs/containers-registries.conf.5.md
+++ b/docs/containers-registries.conf.5.md
@@ -109,6 +109,26 @@ internet without having to change `Dockerfile`s, or to add redundancy).
 *Note*: Redirection and mirrors are currently processed only when reading images, not when pushing
 to a registry; that may change in the future.
 
+#### Normalization of docker.io references
+
+The Docker Hub `docker.io` is handled in a special way: every push and pull
+operation gets internally normalized with `/library` if no other specific
+namespace is defined (for example on `docker.io/namespace/image`).
+
+(Note that the above-described normalization happens to match the behavior of
+Docker.)
+
+This means that a pull of `docker.io/alpine` will be internally translated to
+`docker.io/library/alpine`. A pull of `docker.io/user/alpine` will not be
+rewritten because this is already the correct remote path.
+
+Therefore, to remap or mirror the `docker.io` images in the (implied) `/library`
+namespace (or that whole namespace), the prefix and location fields in this
+configuration file must explicitly include that `/library` namespace. For
+example `prefix = "docker.io/library/alpine"` and not `prefix =
+"docker.io/alpine"`. The latter would match the `docker.io/alpine/*`
+repositories but not the `docker.io/[library/]alpine` image).
+
 ### EXAMPLE
 
 ```
--- a/docs/containers-registries.d.5.md
+++ b/docs/containers-registries.d.5.md
@@ -24,7 +24,7 @@ don’t matter.
 The contents of these files are merged together; to have a well-defined and easy to understand
 behavior, there can be only one configuration section describing a single namespace within a registry
 (in particular there can be at most one one `default-docker` section across all files,
-and there can be at most one instance of any key under the the `docker` section;
+and there can be at most one instance of any key under the `docker` section;
 these sections are documented later).
 
 Thus, it is forbidden to have two conflicting configurations for a single registry or scope,
@@ -98,7 +98,7 @@ docker:
 For developers in `example.com`:
 
 - Consume most container images using the public servers also used by clients.
-- Use a separate sigure storage for an container images in a namespace corresponding to the developers' department, with a staging storage used before publishing signatures.
+- Use a separate signature storage for an container images in a namespace corresponding to the developers' department, with a staging storage used before publishing signatures.
 - Craft an individual exception for a single branch a specific developer is working on locally.
 
 ```yaml


docs/containers-registries.conf.5.md
```diff
@@ -109,6 +109,26 @@ internet without having to change `Dockerfile`s, or to add redundancy).
...
 
+#### Mirroring and rewriting the docker.io registry
+
+The default Docker registry `docker.io` is handled in a special way, because
+every push and pull operation gets internally suffixed with `/library`. This
+means that a pull of `docker.io/image` will be internally translated to
+`docker.io/library/image`. This has to be considered when rewriting registries
+under that domain.
+
+For example, if a Docker registry mirror go setup under `localhost/docker.io`,
```
vrothberg: * s/Docker registry mirror/registry mirror/
* s/go setup/is setup/

docs/containers-registries.conf.5.md
```diff
@@ -109,6 +109,26 @@ internet without having to change `Dockerfile`s, or to add redundancy).
...
+then the appropriate rewrite semantics would look like:
```
vrothberg: s/like/as follows/

docs/containers-registries.conf.5.md
```diff
@@ -109,6 +109,26 @@ internet without having to change `Dockerfile`s, or to add redundancy).
...
+
+```toml
+[[registry]]
+prefix = "docker.io/library"
+location = "localhost/docker.io"
+```
+
+This means if an image get pulled from `docker.io/my-image`, then it gets correctly
```
vrothberg: s/get pulled/is pulled/

docs/containers-registries.conf.5.md
```diff
@@ -109,6 +109,26 @@ internet without having to change `Dockerfile`s, or to add redundancy).
 *Note*: Redirection and mirrors are currently processed only when reading images, not when pushing
 to a registry; that may change in the future.
 
+#### Mirroring and rewriting the docker.io registry
+
+The default Docker registry `docker.io` is handled in a special way, because
+every push and pull operation gets internally suffixed with `/library`. This
+means that a pull of `docker.io/image` will be internally translated to
+`docker.io/library/image`. This has to be considered when rewriting registries
+under that domain.
```
vrothberg: I think it's worth highlighting two more things:
* We need this to match Docker's behavior. Some folks might be wondering why there's such special treatment.
* It's only rewritten, if the specified reference is without a repository. For instance, `docker.io/opensuse/leap` will not be rewritten. I think that's worth mentioning to avoid misinterpreting the example.

docs/containers-registries.conf.5.md
```diff
@@ -109,6 +109,26 @@ internet without having to change `Dockerfile`s, or to add redundancy).
 *Note*: Redirection and mirrors are currently processed only when reading images, not when pushing
 to a registry; that may change in the future.
 
+#### Mirroring and rewriting the docker.io registry
+
+The default Docker registry `docker.io` is handled in a special way, because
+every push and pull operation gets internally suffixed with `/library`. This
```
vrothberg: That's not entirely correct (see below). Not _every_ operation but only those missing a repository.

docs/containers-registries.conf.5.md
saschagrunert: Alright, thanks for the detailed review. I'll adapted the examples to match a real world use case and explicitly mentioned that it only affects the unqualified docker paths.

docs/containers-registries.conf.5.md
```diff
@@ -109,6 +109,31 @@ internet without having to change `Dockerfile`s, or to add redundancy).
...
+`docker.io/alpine` will be internally translated to `docker.io/library/alpine`.
+A pull of `docker.io/user/alpine` will not be rewritten because this is already
+the correct remote path.
+
+This behavior has to be considered when generally rewriting registries under
+the unqualified `docker.io` domain, because containers/image has to match
+Docker’s behavior in that case.
+
+For example, if a registry mirror is setup under `localhost/docker.io`,
+then the appropriate rewrite semantics would look as follows:
```
mtrmac: This is not _always_ the right thing to do, e.g. when mirroring the users’ own Docker Hub images.

It’s great to have this example; this should be a bit clearer about when it is appropriate.

docs/containers-registries.conf.5.md
```diff
@@ -109,6 +109,31 @@ internet without having to change `Dockerfile`s, or to add redundancy).
 *Note*: Redirection and mirrors are currently processed only when reading images, not when pushing
 to a registry; that may change in the future.
 
+#### Mirroring and rewriting the docker.io registry
+
+The default Docker registry `docker.io` is handled in a special way, because
+every push and pull operation gets internally suffixed with `/library`, but only
+if no other specific repository is defined. This means that a pull of
```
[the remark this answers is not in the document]
saschagrunert: I'll renamed it to `namespace` and gave an example for an specific namespace to make it more clear.

docs/containers-registries.conf.5.md
```diff
@@ -109,6 +109,31 @@ internet without having to change `Dockerfile`s, or to add redundancy).
...
+`docker.io/alpine` will be internally translated to `docker.io/library/alpine`.
+A pull of `docker.io/user/alpine` will not be rewritten because this is already
+the correct remote path.
+
+This behavior has to be considered when generally rewriting registries under
+the unqualified `docker.io` domain, because containers/image has to match
```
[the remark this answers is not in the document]
saschagrunert: I decided to rename it to URL to be a bit more abstract, but I'm not sure if that is really better. (waiting for your input in January).

TomSweeneyRedHat:
Looks like my PR beat you in @saschagrunert , you need to rebase.

saschagrunert:
Rebased on top of the latest master :+1:

TomSweeneyRedHat:
@saschagrunert did you sign this one?

TomSweeneyRedHat:
LGTM otherwise

vrothberg:
Note - please don't merge until @mtrmac's ACK

docs/containers-registries.conf.5.md
```diff
@@ -109,6 +109,32 @@ internet without having to change `Dockerfile`s, or to add redundancy).
 *Note*: Redirection and mirrors are currently processed only when reading images, not when pushing
 to a registry; that may change in the future.
 
+#### Mirroring and rewriting the docker.io registry
+
+The default Docker registry `docker.io` is handled in a special way, because
+every push and pull operation gets internally suffixed with `/library` if no
+other specific namespace is defined (for example on `docker.io/username/image`)
```
vrothberg: missing dot

vrothberg:
> If the mirror is a pull-through cache, it will very likely even break pulling when doing such a rewrite.

Yes, it would break pull-through caches (i.e., "proxy" in the registry config).

docs/containers-registries.conf.5.md
saschagrunert: Fixed :)

saschagrunert:
> My previous main concern has not yet been addressed: Mirroring `docker.io/library/` to `$mirror/docker.io` is imho _not_ the appropriate or correct way to mirror the Docker Hub. I really think it's wrong and misleading. As far as I understand, this doc change is motivated by some users being surprised about this behavior. Hence, I think we should only educate the users and explain _why_ the normalization with `/library/` happens. If the mirror is a pull-through cache, it will very likely even break pulling when doing such a rewrite.

Okay, I removed the mirror example to avoid further confusion by users.

docs/containers-registries.conf.5.md
```diff
@@ -109,6 +109,20 @@ internet without having to change `Dockerfile`s, or to add redundancy).
 *Note*: Redirection and mirrors are currently processed only when reading images, not when pushing
 to a registry; that may change in the future.
 
+#### Mirroring and rewriting the docker.io registry
+
+The default Docker registry `docker.io` is handled in a special way, because
+every push and pull operation gets internally suffixed with `/library` if no
```
vrothberg: Can we s/suffixed/normalized/? In case of `docker.io/foo` -> `docker.io/library/foo`, the "suffix" only relates to the namespace.

docs/containers-registries.conf.5.md
```diff
@@ -109,6 +109,20 @@ internet without having to change `Dockerfile`s, or to add redundancy).
...
+other specific namespace is defined (for example on `docker.io/username/image`).
+
+This means that a pull of `docker.io/alpine` will be internally translated to
+`docker.io/library/alpine`. A pull of `docker.io/user/alpine` will not be
+rewritten because this is already the correct remote path.
+
+This behavior has to be considered when generally rewriting registries under
+the unqualified `docker.io` URL, because containers/image has to match
+Docker’s behavior in that case.
```
vrothberg: This sentence is very specific to the c/image library that many users might not be aware of (or even care about). I suggest to turn this into a more general: _"Note that the upper normalization happens to match the behavior of Docker."_

docs/containers-registries.conf.5.md
```diff
@@ -109,6 +109,20 @@ internet without having to change `Dockerfile`s, or to add redundancy).
 *Note*: Redirection and mirrors are currently processed only when reading images, not when pushing
 to a registry; that may change in the future.
 
+#### Mirroring and rewriting the docker.io registry
```
vrothberg: Maybe "Normalization of references to docker.io"? Maybe I need more coffee though :^)

docs/containers-registries.conf.5.md
```diff
@@ -109,6 +109,20 @@ internet without having to change `Dockerfile`s, or to add redundancy).
...
+
+The default Docker registry `docker.io` is handled in a special way, because
```
vrothberg: s/default Docker registry/Docker Hub/

docs/containers-registries.conf.5.md
saschagrunert: Sure, change it to `Normalization of docker.io references`

rhatdan:
@mtrmac Please comment on this PR?

docs/containers-registries.conf.5.md
```diff
@@ -109,6 +109,18 @@ internet without having to change `Dockerfile`s, or to add redundancy).
...
+
+The Docker Hub `docker.io` is handled in a special way, because every push and
+pull operation gets internally normalized with `/library` if no other specific
+namespace is defined (for example on `docker.io/username/image`).
+
+This means that a pull of `docker.io/alpine` will be internally translated to
+`docker.io/library/alpine`. A pull of `docker.io/user/alpine` will not be
+rewritten because this is already the correct remote path.
+
+Note that the upper normalization happens to match the behavior of Docker.
```
mtrmac: s/the upper normalization/the above-described normalization/ ?

Possibly move it to the first paragraph in a parenthesis.

docs/containers-registries.conf.5.md
```diff
@@ -109,6 +109,18 @@ internet without having to change `Dockerfile`s, or to add redundancy).
...
 
+#### Normalization of docker.io references
+
+The Docker Hub `docker.io` is handled in a special way, because every push and
+pull operation gets internally normalized with `/library` if no other specific
+namespace is defined (for example on `docker.io/username/image`).
+
+This means that a pull of `docker.io/alpine` will be internally translated to
+`docker.io/library/alpine`. A pull of `docker.io/user/alpine` will not be
+rewritten because this is already the correct remote path.
```
mtrmac: This version may have overshot the “let’s not give the users a misleading example” direction, in particular I’m a bit worried that the users who read this will not connect this statement to its impact on the mirror/redirection features.

Maybe add a paragraph here vaguely to the effect of
> Therefore, to remap/mirror the `docker.io` images in the implied `/library` namespace (or that whole namespace), the `prefix` and `location` fields in this configuration file must explicitly include that `/library` namespace (e.g. `prefix = "docker.io/library/alpine"`, not `prefix = "docker.io/alpine"`; the latter would match the `docker.io/alpine/*` repositories but not the `docker.io/[library/]alpine` image).

OTOH this one is probably over-explaining yet again, please edit as desired.

docs/containers-registries.conf.5.md
```diff
@@ -109,6 +109,18 @@ internet without having to change `Dockerfile`s, or to add redundancy).
 *Note*: Redirection and mirrors are currently processed only when reading images, not when pushing
 to a registry; that may change in the future.
 
+#### Normalization of docker.io references
+
+The Docker Hub `docker.io` is handled in a special way, because every push and
```
mtrmac: ( `s|, because|:|` ? The rest does not give a reason. )

docs/containers-registries.conf.5.md
```diff
@@ -109,6 +109,18 @@ internet without having to change `Dockerfile`s, or to add redundancy).
...
+pull operation gets internally normalized with `/library` if no other specific
+namespace is defined (for example on `docker.io/username/image`).
```
mtrmac: This feels a little opaque about what a “namespace” means.  I don’t think we should give a formal syntax here, changing the example to `docker.io/namespace/image` should be enough.

(Absolutely non-blocking: I might find it a little easier to read with the condition (“no namespace is explicitly specified”) before the consequence (“the image is normalized to use the default /library namespace”). )

docs/containers-registries.conf.5.md
saschagrunert: Fixed :+1:

docs/containers-registries.conf.5.md
saschagrunert: I adapted your paragraph and changed it a bit.

saschagrunert:
Alright, I added the review notes and gave it another rebase on top of the latest master branch.

TomSweeneyRedHat:
LGTM

Issue #783, opened 2019-12-29
Allow pulling from a docker reference with tag and digest


mgoltzsche:
Docker's image pull syntax `repo/image:tag@sha` should be supported as well for compatibility reasons - despite its ambiguity.
I came across this problem using [skaffold](https://github.com/GoogleContainerTools/skaffold/issues/3220) and [CRI-O](https://github.com/cri-o/cri-o/issues/2351).

The current code [already handles](https://github.com/containers/image/blob/94de853cb65ab465f3530d33ff3ff91454e70569/docker/docker_image_src.go#L178) pulling such image references by digest properly. However a `dockerReference` construction explicitly [fails](https://github.com/containers/image/blob/94de853cb65ab465f3530d33ff3ff91454e70569/docker/docker_transport.go#L80) when both tag and digest are specified.

mgoltzsche:
Closed in favour of [a better solution](https://github.com/cri-o/cri-o/pull/3060) within CRI-O.

PR #784, opened 2019-12-29
WIP: Allow image specified by both tag and digest.

mgoltzsche:
Refers to #783.

Changes:
- Moved the code that returns an error when both tag and digest are
  present from `newReference` to `reference.newImageDestination` in
  order to allow pulls but no pushes.
- If present use digest instead of tag as `DockerReferenceIdentity`
  or rather for policy matching.
- Adjusted tests.

Signed-off-by: Max Goltzsche <max.goltzsche@gmail.com>

Question remains how well this works with the other tools in the chain - needs to be tested first which  is why this is still WIP.

### DIFF
--- a/docker/docker_image_dest.go
+++ b/docker/docker_image_dest.go
@@ -36,6 +36,15 @@ type dockerImageDestination struct {
 
 // newImageDestination creates a new ImageDestination for the specified image reference.
 func newImageDestination(sys *types.SystemContext, ref dockerReference) (types.ImageDestination, error) {
+	// A github.com/distribution/reference value can have a tag and a digest at the same time!
+	// While some tools (namely skaffold) rely on this docker behaviour to pull the correct image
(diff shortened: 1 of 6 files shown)


mgoltzsche:
I realize now that it may indeed make more sense to handle this case in CRI-O directly/only - especially because allowing the syntax only for pulling but not for pushing is inconsistent. For pushing such a tag the `DockerReferenceIdentity` would need behave differently than for a pull because it would need to check the tag instead of the digest.

mgoltzsche:
I am closing this in favour of [a simpler solution](https://github.com/cri-o/cri-o/pull/3060) within CRI-O (where the problem appears) to avoid polluting this library.
To preserve them for historical/documentation purposes I moved this PR's changes into a separate [branch](https://github.com/mgoltzsche/image/tree/pull-by-both-tag-and-digest-attempt).

PR #786, opened 2020-01-01
Adds the possibility to choose the image by the variant field in a manifest

JirkaChadima:
This is part of an https://github.com/containers/skopeo/issues/790 and improves the `ChooseImage` method so that it respects the `variant` field which is used on ARM architecture. It also exposes the `VariantChoice` to the outside world, so it can be passed as a command line argument.

I'm not a Golang person, so it might be a little rusty. I intentionally kept the original behaviour that when there is no match for Variant, the first record matching the platform and architecture is returned. It might be more transparent to fail in that case though.

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -705,16 +705,21 @@ func checkImageDestinationForCurrentRuntime(ctx context.Context, sys *types.Syst
 		}
 		wantedPlatforms, err := platform.WantedPlatforms(sys)
 		if err != nil {
-			return errors.Wrapf(err, "error getting platform information %#v", sys)
+			return errors.Wrapf(err, "error getting current platform information %#v", sys)
 		}
-		wantedPlatform := wantedPlatforms[0] // TODO fixme
-		if wantedPlatform.OS != c.OS {
-			return fmt.Errorf("Image operating system mismatch: image uses %q, expecting %q", c.OS, wantedPlatform.OS)
+		for _, wantedPlatform := range wantedPlatforms {
+			if wantedPlatform.OS != c.OS {
+				return fmt.Errorf("Image operating system mismatch: image uses %q, expecting %q", c.OS, wantedPlatform.OS)
+			}
+			if wantedPlatform.Architecture != c.Architecture {
+				return fmt.Errorf("Image architecture mismatch: image uses %q, expecting %q", c.Architecture, wantedPlatform.Architecture)
+			}
+			/*
+				// TODO Waiting for https://github.com/opencontainers/image-spec/pull/777
+				if wantedPlatform.Variant != "" && c.Variant != "" && wantedPlatform.Variant != c.Variant {
+					return fmt.Errorf("Image variant mismatch: image uses %q, expecting %q", c.Variant, wantedPlatform.Variant)
+				}*/
 		}
-		if wantedPlatform.Architecture != c.Architecture {
-			return fmt.Errorf("Image architecture mismatch: image uses %q, expecting %q", c.Architecture, wantedPlatform.Architecture)
-		}
-		// TODO also check for variant
 	}
 	return nil
 }
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -702,10 +702,11 @@ func checkImageDestinationForCurrentRuntime(ctx context.Context, sys *types.Syst
 		if err != nil {
 			return errors.Wrapf(err, "Error parsing image configuration")
 		}
-		wantedPlatform, err := manifest.WantedPlatform(sys)
+		wantedPlatforms, err := manifest.WantedPlatforms(sys)
 		if err != nil {
 			return errors.Wrapf(err, "error getting platform information %#v", sys)
 		}
+		wantedPlatform := wantedPlatforms[0] // TODO fixme
 		if wantedPlatform.OS != c.OS {
 			return fmt.Errorf("Image operating system mismatch: image uses %q, expecting %q", c.OS, wantedPlatform.OS)
 		}
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -8,7 +8,6 @@ import (
 	"io/ioutil"
 	"os"
 	"reflect"
-	"runtime"
 	"strings"
 	"sync"
 	"time"
@@ -703,22 +702,17 @@ func checkImageDestinationForCurrentRuntime(ctx context.Context, sys *types.Syst
 		if err != nil {
 			return errors.Wrapf(err, "Error parsing image configuration")
 		}
-
-		wantedOS := runtime.GOOS
-		if sys != nil && sys.OSChoice != "" {
-			wantedOS = sys.OSChoice
+		wantedPlatform, err := manifest.WantedPlatform(sys)
+		if err != nil {
+			return errors.Wrapf(err, "error getting platform information %#v", sys)
 		}
-		if wantedOS != c.OS {
-			return fmt.Errorf("Image operating system mismatch: image uses %q, expecting %q", c.OS, wantedOS)
+		if wantedPlatform.OS != c.OS {
+			return fmt.Errorf("Image operating system mismatch: image uses %q, expecting %q", c.OS, wantedPlatform.OS)
 		}
-
-		wantedArch := runtime.GOARCH
-		if sys != nil && sys.ArchitectureChoice != "" {
-			wantedArch = sys.ArchitectureChoice
+		if wantedPlatform.Architecture != c.Architecture {
+			return fmt.Errorf("Image architecture mismatch: image uses %q, expecting %q", c.Architecture, wantedPlatform.Architecture)
 		}
-		if wantedArch != c.Architecture {
-			return fmt.Errorf("Image architecture mismatch: image uses %q, expecting %q", c.Architecture, wantedArch)
-		}
+		// TODO also check for variant
 	}
 	return nil
 }
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/image/v5/image"
+	platform "github.com/containers/image/v5/internal/pkg/platform"
 	"github.com/containers/image/v5/manifest"
 	"github.com/containers/image/v5/pkg/blobinfocache"
 	"github.com/containers/image/v5/pkg/compression"
@@ -702,7 +703,7 @@ func checkImageDestinationForCurrentRuntime(ctx context.Context, sys *types.Syst
 		if err != nil {
 			return errors.Wrapf(err, "Error parsing image configuration")
 		}
-		wantedPlatforms, err := manifest.WantedPlatforms(sys)
+		wantedPlatforms, err := platform.WantedPlatforms(sys)
 		if err != nil {
 			return errors.Wrapf(err, "error getting platform information %#v", sys)
 		}
--- a/internal/pkg/platform/platform_matcher.go
+++ b/internal/pkg/platform/platform_matcher.go
@@ -1,4 +1,4 @@
-package manifest
+package platform
 
 // Largely based on
 // https://github.com/moby/moby/blob/bc846d2e8fe5538220e0c31e9d0e8446f6fbc022/distribution/cpuinfo_unix.go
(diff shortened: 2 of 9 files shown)


manifest/docker_schema2_list.go
```diff
@@ -101,11 +101,25 @@ func (list *Schema2List) ChooseInstance(ctx *types.SystemContext) (digest.Digest
 		wantedOS = ctx.OSChoice
 	}
 
+	var fallback digest.Digest
+
 	for _, d := range list.Manifests {
 		if d.Platform.Architecture == wantedArch && d.Platform.OS == wantedOS {
-			return d.Digest, nil
+			// TODO It should be possible to somehow use runtime.GOARM to construct a default VariantChoice
+			if ctx.VariantChoice == "" {
```
rhatdan: Why not
if ctx == "" || d.Platform.Variant == ctx.VariantChoice {
 				return d.Digest, nil
}

manifest/oci_index.go
```diff
@@ -84,11 +84,25 @@ func (index *OCI1Index) ChooseInstance(ctx *types.SystemContext) (digest.Digest,
 		wantedOS = ctx.OSChoice
 	}
 
+	var fallback digest.Digest
+
 	for _, d := range index.Manifests {
 		if d.Platform != nil && d.Platform.Architecture == wantedArch && d.Platform.OS == wantedOS {
-			return d.Digest, nil
+			// TODO It should be possible to somehow use runtime.GOARM to construct a default VariantChoice
+			if ctx.VariantChoice == "" {
```
rhatdan: Same question as above.

manifest/docker_schema2_list.go
JirkaChadima: That slipped, I'll fix it

rhatdan:
LGTM
But we need @mtrmac and @vrothberg to review.

manifest/docker_schema2_list.go
```diff
@@ -101,11 +101,23 @@ func (list *Schema2List) ChooseInstance(ctx *types.SystemContext) (digest.Digest
 		wantedOS = ctx.OSChoice
 	}
 
+	var fallback digest.Digest
+
 	for _, d := range list.Manifests {
 		if d.Platform.Architecture == wantedArch && d.Platform.OS == wantedOS {
-			return d.Digest, nil
+			// TODO It should be possible to somehow use runtime.GOARM to construct a default VariantChoice
+			if ctx.VariantChoice == "" || d.Platform.Variant == ctx.VariantChoice {
```
mtrmac: `ctx` can be `nil`, see how it is used above.

manifest/oci_index.go
```diff
@@ -84,11 +84,23 @@ func (index *OCI1Index) ChooseInstance(ctx *types.SystemContext) (digest.Digest,
 		wantedOS = ctx.OSChoice
 	}
 
+	var fallback digest.Digest
+
 	for _, d := range index.Manifests {
 		if d.Platform != nil && d.Platform.Architecture == wantedArch && d.Platform.OS == wantedOS {
-			return d.Digest, nil
+			// TODO It should be possible to somehow use runtime.GOARM to construct a default VariantChoice
+			if ctx.VariantChoice == "" || d.Platform.Variant == ctx.VariantChoice {
```
mtrmac: `ctx` can be `nil`, see how it is used above.

types/types.go
```diff
@@ -510,6 +510,8 @@ type SystemContext struct {
 	ArchitectureChoice string
 	// If not "", overrides the use of platform.GOOS when choosing an image or verifying OS match.
 	OSChoice string
+	// If not "", it is used when choosing an image or verifying OS match
+	VariantChoice string
```
mtrmac: “it is used when” does not say what the value means.

JirkaChadima:
Thanks for pointing out #753, I will check that out. Do you have any experience with [GOARM](https://github.com/golang/go/wiki/GoArm)? That might be used for default detection.

mtrmac:
#753 does point at _some_ pre-existing implementations; at this point, I’m happy to defer to ARM experts as long as the proposed mechanism is not excessively invasive and minimally plausible. We can always improve on the detection if there is interest.

JirkaChadima:
I did some digging around. The GOARM seems to be primarily used when compiling GO programs and is [a little controversial](https://github.com/golang/go/issues/9737), so I would not use it at all, the [support table](https://github.com/golang/go/wiki/GoArm#supported-architectures) refers to the ability to compile GO programs for those architectures/variants. And from my tests on RPi, it is not even a public property during runtime.

The go to option seems to be the parsing of `/proc/cpuinfo`. The whole approach is discussed in great detail [in this presentation](https://events19.linuxfoundation.org/wp-content/uploads/2017/12/Improve-the-container-image-compatibility-on-Arm_Penny-Zheng.pdf) from ARM.

So I'm going to try to port the containerd code while taking a look into moby/moby as well. This is apparently not an easy problem.

copy/copy.go
```diff
@@ -703,22 +702,17 @@ func checkImageDestinationForCurrentRuntime(ctx context.Context, sys *types.Syst
...
 			return errors.Wrapf(err, "Error parsing image configuration")
 		}
-
-		wantedOS := runtime.GOOS
-		if sys != nil && sys.OSChoice != "" {
-			wantedOS = sys.OSChoice
-		}
-		if wantedOS != c.OS {
-			return fmt.Errorf("Image operating system mismatch: image uses %q, expecting %q", c.OS, wantedOS)
+		wantedPlatform, err := manifest.WantedPlatform(sys)
```
mtrmac: I don’t think a single value returned here is going to work long-term (a new variant of an architecture may support several previous variants).

That’s not a strong reason not to do it this way _for now_ at all, just a reason not to make it a public API (see below).

manifest/platform_matcher.go
```diff
@@ -0,0 +1,120 @@
...
+		wantedOS = ctx.OSChoice
+	}
+
+	wantedPlatform := imgspecv1.Platform{
+		OS:           wantedOS,
+		Architecture: wantedArch,
+	}
+
+	if wantedArch == "arm" || wantedArch == "arm64" {
+		if ctx != nil && ctx.VariantChoice != "" {
```
mtrmac: Why not switch the nesting of these two `if`, so that user input is always used, and guessing happens only on ARM? Silently ignoring user input would be surprising.

manifest/platform_matcher.go
```diff
@@ -0,0 +1,120 @@
...
+	case "3":
+		variant = "v3"
+	default:
+		variant = "unknown"
+	}
+
+	return variant
+}
+
+func WantedPlatform(ctx *types.SystemContext) (imgspecv1.Platform, error) {
```
mtrmac: This should be a private subpackage; we don’t need to make this public (no external users are known AFAIK, at least), and at this point we are probably not ready to commit to a stable ABI.

(We’ve not historically done that in c/image, but that’s why are at v5.x.y now, so we really need to change that.)

`containers/image/internal/pkg/platform` would be a possible place. (@vrothberg RFC)

copy/copy.go
JirkaChadima: Ah, I see where you are going with this. It is probably a good place to put the "fallback" variant logic inside the platform matcher rather than in the image selection process itself.

manifest/platform_matcher.go
JirkaChadima: Sounds reasonable.

manifest/platform_matcher.go
JirkaChadima: I am not really familiar of how the internal/external API works in GO, so just tell me where to put the code and I'll do it. Thanks.

JirkaChadima:
One thing I'm still not sure about is `copy.CheckImageDestinationForCurrentRuntime`. What does it do exactly? I wasn't able to get the `Variant` field out from the `Image` so I have nothing to compare it to, but maybe I was just looking in the wrong place.

manifest/platform_matcher.go
JirkaChadima: Oh, actually no. I think that we need to always guess variant on arm if no user input is specified to download the proper variant. On any other architecture, variant does not seem to be used, so it should be ignored or nulled. Or maybe fail if there is user input for `VariantChoice`.

mtrmac: That `variant` is only used on ARM is just the current state, not an inherent property of the universe.

Allowing `VariantChoice` on other architectures could, in the future, allow using this version of the code (unaware of variants on those architectures) by newer callers (aware of variants), explicitly making a choice that c/image cannot (yet) make.

> Or maybe fail if there is user input for VariantChoice.

Yes; it should _not_ be ignored (because, again, in some future version it may start to matter, and code writing nonsense in there would start to fail).

Given an architecture that does not have variants, and an user-provided `VariantChoice`, I would expect the architecture-independent part of this PR to _naturally_, without special-casing, reject any image — exactly the same as if the caller asked for ARM variant `this-is-not-an-ARM-variant-at-all`.

mtrmac: > reject any image

reject any image that does not have such a variant explicitly present, that is.

manifest/platform_matcher.go
mtrmac: e.g. (`containers/image`)/`internal/pkg/platform/platform_matcher.go`.

That will allow using the platform matcher from any subpackage in c/image  , (exactly the same as if the subpackage were named differently), but it won’t be visible / possible to import from code outside of c/image.

See https://golang.org/doc/go1.4#internalpackages .

mtrmac:
> One thing I'm still not sure about is `copy.CheckImageDestinationForCurrentRuntime`. What does it do exactly?

This exists so that copies of images that are intended to be _run_ (e.g. “pulls” from remote registries to the local filesystem, as opposed to copies of images from one registry to another, without intending to directly run the image) fail early, i.e. that `podman pull ARM-only-image` or `podman pull Windows-only-image` fail on an x86 Linux.

Without this check, it’s perfectly possible to download images and extract them to a filesystem, it would only be _running_ containers from such images that would fail. This allows alerting the user to this earlier, when it’s easy to detect and it can avoid costly/long pulls of unusable data.

> I wasn't able to get the `Variant` field out from the `Image` so I have nothing to compare it to, but maybe I was just looking in the wrong place.

Indeed; see https://github.com/opencontainers/image-spec/pull/777 . Until that lands (and images start using it), I suppose missing image variant has to be treated as “unknown, assumed compatible” in that code.

manifest/platform_matcher.go
JirkaChadima: This approach totally makes sense. I'll see what I can do.

manifest/platform_matcher.go
JirkaChadima: Thanks, I'll check that out.

JirkaChadima:
> Until that lands (and images start using it), I suppose missing image variant has to be treated as “unknown, assumed compatible” in that code.

I've added the code, but it's commented out. I'm not sure if there is another way in a typed language.

Otherwise I think it's doing what it should be and I've hopefully addressed all of the (very helpful!) remarks. I'm pretty sure there is still some code that's not very go-like, so let me know what to change. 

Thanks for your feedback, it's great to contribute to a project with such kind and welcoming maintainers.

JirkaChadima:
I'm sorry if my last comment sounded definitive. Is there anything else I can do to push this closer to being merged? If not, is there any kind of timeline for this being merged and/or released? Thanks!

mtrmac:
Don’t worry, the comment is perfectly fine; I’m afraid I just got busy with other work.

mtrmac:
FWIW there were a few user reports of broken user cases by the architecture check in `checkImageDestinationForCurrentRuntime`, so we are going to remove it again.

rhatdan:
@mtrmac Whats the scoop on this one?

manifest/oci_index.go
```diff
@@ -75,26 +76,31 @@ func (index *OCI1Index) UpdateInstances(updates []ListUpdate) error {
...
 		}
 	}
+
 	for _, d := range index.Manifests {
 		if d.Platform == nil {
 			return d.Digest, nil
 		}
 	}
-	return "", fmt.Errorf("no image found in image index for architecture %s, OS %s", wantedArch, wantedOS)
+	return "", fmt.Errorf("no image found in manifest list for architecture %s, variant %s, OS %s", wantedPlatforms[0].Architecture, wantedPlatforms[0].Variant, wantedPlatforms[0].OS)
```
mtrmac: “image index” is the OCI terminology, and should be preserved.

internal/pkg/platform/platform_matcher.go
```diff
@@ -0,0 +1,184 @@
...
+
+var compatibility = map[string][]string{
+	"arm":   {"v7", "v6", "v5"},
+	"arm64": {"v8"},
+}
+
+// Returns all compatible platforms with the platform specifics possibly overriden by user,
+// the most compatible platform is first.
+// If some option (arch, os, variant) is not present, a value from current platform is detected.
+func WantedPlatforms(ctx *types.SystemContext) ([]imgspecv1.Platform, error) {
```
mtrmac: Non-blocking: This can currently never fail, and it’s an internal API. The `error` return value could be dropped, which would simplify callers a bit.

copy/copy.go
```diff
@@ -703,21 +703,22 @@ func checkImageDestinationForCurrentRuntime(ctx context.Context, sys *types.Syst
...
+			return errors.Wrapf(err, "error getting current platform information %#v", sys)
 		}
-		if wantedArch != c.Architecture {
-			return fmt.Errorf("Image architecture mismatch: image uses %q, expecting %q", c.Architecture, wantedArch)
+		for _, wantedPlatform := range wantedPlatforms {
+			if wantedPlatform.OS != c.OS {
+				return fmt.Errorf("Image operating system mismatch: image uses %q, expecting %q", c.OS, wantedPlatform.OS)
+			}
+			if wantedPlatform.Architecture != c.Architecture {
+				return fmt.Errorf("Image architecture mismatch: image uses %q, expecting %q", c.Architecture, wantedPlatform.Architecture)
```
mtrmac: This now needs to just use `logrus.Infof` instead of failing.

JirkaChadima:
@mtrmac I think it would be more efficient if you take over. I don't remember much about this anymore TBH. And I wasn't following any other changes here, so I'm totally out of context anyway.

mtrmac:
Updated in #854. Thanks again!

PR #788, opened 2020-01-07
Improve error reporting on mirror failures

### LINKED ISSUE #674: Improve error reporting on mirror failures
`docker.newImageSource` only reports the last failure (= the failure of the primary endpoint), without mentioning anything about the failures to use mirrors.

That’s acceptable for mirrors that exist for performance improvements only, but unhelpful when the mirrors exist for redirection (i.e. the primary endpoint is by design unreachable and will never be reachable) and the mirror failures are the only ones that matter.

mtrmac:
Primarily, ensure that errors reported by `docker.newImageSource` (i.e. most errors trying to access registries) also contain failures when accessing _all_ mirrors, not just the last error = the error accessing the primary endpoint.

The resulting error is extremely ugly (see below) but hopefully informative, and we can tinker with that later after this is not as urgent (in two ways: by dropping some context and shortening the error in common failure cases, and by by heuristically prioritizing some of the errors vs. others, e.g. a TLS / user authentication failure for a mirror is more likely to be relevant than a DNS not found for the primary).

In addition:
- Some errors (notably an endpoint being configured as blocked) are now not immediately fatal, the other endpoints continue to be attempted.
- The debug logs, in addition to starting each attempt with `Trying to pull…` (now `Trying to access…`), also contain the failure for each endpoint as soon as that endpoint fails.

Fixes #674.

### DIFF
--- a/docker/docker_image_src.go
+++ b/docker/docker_image_src.go
@@ -60,9 +60,9 @@ func newImageSource(ctx context.Context, sys *types.SystemContext, ref dockerRef
 	}
 	for _, pullSource := range pullSources {
 		logrus.Debugf("Trying to pull %q", pullSource.Reference)
-		testImageSource, err := newImageSourceAttempt(ctx, sys, pullSource, primaryDomain)
+		s, err := newImageSourceAttempt(ctx, sys, pullSource, primaryDomain)
 		if err == nil {
-			return testImageSource, nil
+			return s, nil
 		}
 		manifestLoadErr = err
 	}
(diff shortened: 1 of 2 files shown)


mtrmac:
Examples:
```console
$ ./skopeo inspect docker://busybox:thisdoesnotexist
FATA[0001] Error parsing image name "docker://busybox:thisdoesnotexist": Error reading manifest thisdoesnotexist in docker.io/library/busybox: manifest unknown: manifest unknown 
```
With a `docker.io` "mirror" set to `docker.io`, primary redirected to non-existent
```console
$ ./skopeo --debug --registries-conf ./registries.conf inspect docker://busybox:thisdoesnotexist
FATA[0001] Error parsing image name "docker://busybox:thisdoesnotexist": (First tried mirrors: [docker.io/library/busybox:thisdoesnotexist: Error reading manifest thisdoesnotexist in docker.io/library/busybox: manifest unknown: manifest unknown]): does-not-exist-primary.example.com/library/busybox:thisdoesnotexist: error pinging docker registry does-not-exist-primary.example.com: Get https://does-not-exist-primary.example.com/v2/: dial tcp: lookup does-not-exist-primary.example.com on 192.168.64.1:53: no such host 
```

mtrmac:
Tested and documented above now. PTAL.

docker/docker_image_src.go
```diff
@@ -73,10 +71,10 @@ func newImageSource(ctx context.Context, sys *types.SystemContext, ref dockerRef
 
 // newImageSourceAttempt is an internal helper for newImageSource. Everyone else must call newImageSource.
 // Given a specified source and primaryDomain, return a dockerImage source, or an error, and an indication whether the error is fatal.
```
vrothberg: Non-blocking nit: the comment needs an update.

docker/docker_image_src.go
```diff
@@ -53,20 +54,42 @@ func newImageSource(ctx context.Context, sys *types.SystemContext, ref dockerRef
...
+	default:
+		// Don’t just build a string, try to preserve the typed error.
+		primary := &attempts[len(attempts)-1]
+		extras := []string{}
+		for i := 0; i < len(attempts)-1; i++ {
+			// This is difficult to fit into a single-line string, when the error can contain arbitrary strings including any metacharacters we decide to use.
+			// The paired [] at least have some chance of being unambiguous.
+			extras = append(extras, fmt.Sprintf("[%s: %v]", attempts[i].ref.String(), attempts[i].err))
+		}
+		return nil, errors.Wrapf(primary.err, "(First tried mirrors: %s): %s", strings.Join(extras, "\n"), primary.ref.String())
```
vrothberg: * Isn't `primary.ref.String()` already part of `primary.err`?
* "First tried mirrors:" is hard to understand. Maybe "Other contacted endpoints:"?

docker/docker_image_src.go
mtrmac: Fixed, thanks!

docker/docker_image_src.go
mtrmac: - Not always, or at least not in full, see e.g. the “error pinging” example in https://github.com/containers/image/pull/788#issuecomment-572133188 .
- 
  - We don’t use the “endpoint” abstraction name in user-facing text so far; “mirror” directly corresponds to the configuration.
  - The “first tried” wording is attempting to convey order; “other contacted” feels to me a bit as if the primary was contacted first, and the others later. I do agree that this is hard to understand. Maybe I should give up on trying to explain the order in here?
  - (This is also somewhat constrained by the `errors.Wrapf` mechanism only adding text to the _start_ of the message. Ideally, it should be possible to say: “mirror 1 not found in DNS, LOOK HERE: mirror 2 found but authentication failed, primary not found in DNS” with both the appropriate order and the appropriate focus on the likely-to-be-important part, _and_ returning the correct typed error for “authentication failed”. This is anyway nowhere near that.)

mtrmac: Just to throw something out there, maybe “Mirrors also failed:”?

vrothberg: Sounds good, thanks!

mtrmac: To be explicit, I’ll go with `Mirrors also failed:`, then.

PR #796, opened 2020-01-20
Always use SigningMechanism implementations by reference

mtrmac:
… instead of by value.

AFAICS this did not have any actual undesirable impact (the mechanism object is either not used at all, or used only to read another object reference), but it was not intentional and passing the data by value could lead to surprises later.

### DIFF
--- a/signature/mechanism_gpgme.go
+++ b/signature/mechanism_gpgme.go
@@ -139,7 +139,7 @@ func (m *gpgmeSigningMechanism) Sign(input []byte, keyIdentity string) ([]byte, 
 }
 
 // Verify parses unverifiedSignature and returns the content and the signer's identity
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
Does this cause a change in API?

vrothberg:
> Does this cause a change in API?

I don't think so. The type is package internal and implements a public interface, so I can't imagine a scenario where it would break consumers.

LGTM

TomSweeneyRedHat:
LGTM and all happy green tests buttons.

mtrmac:
> Does this cause a change in API?

Good question; no – the only constructor returns an interface and you can’t legally / in a supported way get at the underlying pointer or make a copy of the value:
```go
	f, x, _ := newEphemeralGPGSigningMechanism(nil)
	y := *f // invalid indirect of f (type SigningMechanism)
```

PR #801, opened 2020-01-27
oci manifest: do not require whitelist of layer types

tych0:
69aa1e85460f ("media type checks") introduced a whitelist of required mime
types for individual layers.

However, this actually violates the OCI spec,
https://github.com/opencontainers/image-spec/blob/master/manifest.md which
says:

"An encountered mediaType that is unknown to the implementation MUST be ignored."

Indeed, we have code that uses manifest types not defined in this
whitelist, so it breaks our code.

### DIFF
--- a/image/oci_test.go
+++ b/image/oci_test.go
@@ -427,5 +427,5 @@ func TestConvertToV2S2WithInvalidMIMEType(t *testing.T) {
 	require.NoError(t, err)
 
 	_, err = manifestOCI1FromManifest(originalSrc, manifest)
-	require.Error(t, err)
+	require.NoError(t, err)
 }
--- a/manifest/docker_schema2_list.go
+++ b/manifest/docker_schema2_list.go
@@ -81,9 +81,6 @@ func (list *Schema2List) UpdateInstances(updates []ListUpdate) error {
 		if updates[i].MediaType == "" {
 			return errors.Errorf("update %d of %d passed to Schema2List.UpdateInstances had no media type (was %q)", i+1, len(updates), list.Manifests[i].MediaType)
 		}
-		if err := SupportedSchema2MediaType(updates[i].MediaType); err != nil && SupportedOCI1MediaType(updates[i].MediaType) != nil {
-			return errors.Wrapf(err, "update %d of %d passed to Schema2List.UpdateInstances had an unsupported media type (was %q): %q", i+1, len(updates), list.Manifests[i].MediaType, updates[i].MediaType)
-		}
 		list.Manifests[i].MediaType = updates[i].MediaType
 	}
 	return nil
--- a/manifest/oci.go
+++ b/manifest/oci.go
@@ -32,7 +32,14 @@ type OCI1 struct {
 	imgspecv1.Manifest
 }
 
-// SupportedOCI1MediaType checks if the specified string is a supported OCI1 media type.
+// SupportedOCI1MediaType checks if the specified string is a supported OCI1
+// media type.
+//
+// Deprecated: blindly rejecting unknown MIME types when the consumer does not
+// need to process the input just reduces interoperability (and violates the
+// standard) with no benefit, and that this function does not check that the
+// media type is appropriate for any specific purpose, so it’s not all that
+// useful for validation anyway.
 func SupportedOCI1MediaType(m string) error {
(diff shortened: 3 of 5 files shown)


image/oci_test.go
```diff
@@ -420,12 +420,3 @@ func TestConvertToManifestSchema2AllMediaTypes(t *testing.T) {
...
 	require.Error(t, err) // zstd compression is not supported for docker images
 }
-
-func TestConvertToV2S2WithInvalidMIMEType(t *testing.T) {
-	originalSrc := newOCI1ImageSource(t, "httpd-copy:latest")
-	manifest, err := ioutil.ReadFile(filepath.Join("fixtures", "oci1-invalid-media-type.json"))
-	require.NoError(t, err)
-
-	_, err = manifestOCI1FromManifest(originalSrc, manifest)
-	require.Error(t, err)
```
vrothberg: It wouldn't hurt to keep the test around but adjust it to not fail. This will make sure we're not regressing in the future.

manifest/docker_schema2_list.go
```diff
@@ -81,7 +81,7 @@ func (list *Schema2List) UpdateInstances(updates []ListUpdate) error {
 		if updates[i].MediaType == "" {
 			return errors.Errorf("update %d of %d passed to Schema2List.UpdateInstances had no media type (was %q)", i+1, len(updates), list.Manifests[i].MediaType)
 		}
-		if err := SupportedSchema2MediaType(updates[i].MediaType); err != nil && SupportedOCI1MediaType(updates[i].MediaType) != nil {
```
vrothberg: Removing the OCI checks while keeping the schema2 ones is causing the CI failures:

time="2020-01-27T23:43:21Z" level=fatal msg="Error updating manifest list: update 1 of 6 passed to Schema2List.UpdateInstances had an unsupported media type (was \"application/vnd.docker.distribution.manifest.v2+json\"): \"application/vnd.oci.image.manifest.v1+json\": unsupported docker v2s2 media type: \"application/vnd.oci.image.manifest.v1+json\""

I guess we have to relax the checks altogether.

tych0:
Sure, sorry for the delay. I went a head and restored the test and inverted the error code, and dropped the other check. Let's hope the tests pass now :)

vrothberg:
Thanks, @tych0!

@mtrmac @nalind @rhatdan PTAL

tych0:
Seems like travis might be stuck? :(

manifest/oci.go
```diff
@@ -32,31 +31,12 @@ type OCI1 struct {
 	imgspecv1.Manifest
 }
 
-// SupportedOCI1MediaType checks if the specified string is a supported OCI1 media type.
-func SupportedOCI1MediaType(m string) error {
```
mtrmac: Removing a public function would be an API break.

Please keep it around, _including tests_, with a https://github.com/golang/go/wiki/Deprecated comment (saying something to the effect that blindly rejecting unknown MIME types when the consumer does not need to _process_ the input just reduces interoperability with no benefit, and that this function does not check that the media type is appropriate for any specific purpose, so it’s not all that useful for validation anyway).

manifest/docker_schema2_list.go
mtrmac: If we are relaxing the schema2 checks _here_, why not do it in the other places as well (i.e. do the same edits for `SupportedSchema2MediaType`)? That way OCI and schema2 remain consistent.

mtrmac:
> Seems like travis might be stuck? :(

Not stuck, just very slow; I’m afraid multi-hour delays are common. We do plan to move to a different infrastructure, eventually.

tych0:
I somehow can't reply to the comment about why the check is in docker, seems like a github bug. Anyway, I think it *is* valid in the docker case, because you can't upload just arbitrary layers to docker. It's not valid in OCI, because the spec says that arbitrary layers are allowed.

manifest/oci.go
tych0: Adding this check was also an API break, but ok :). I've re-added it.

rhatdan:
@tych0 @mtrmac Where are we on this PR?

tych0:
Should be good to go; I believe I've addressed everything.

PR #803, opened 2020-01-29
restrict sizes of inmem processed blobs

vrothberg:
Restrict the sizes of blobs which are copied into memory such as the
manifest, the config, signatures, etc.  This will protect consumers of
c/image from rogue or hijacked registries that return too big blobs in
hope to cause an OOM DOS attack.

Note that error messages should be improved in a future change to make
sure that it's clear in which code path we hit a limit.

Fixes: CVE-2020-1702
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1792796

### DIFF
--- a/docker/docker_image_src.go
+++ b/docker/docker_image_src.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 
 	"github.com/containers/image/v5/docker/reference"
+	"github.com/containers/image/v5/internal/iolimits"
 	"github.com/containers/image/v5/manifest"
 	"github.com/containers/image/v5/pkg/sysregistriesv2"
 	"github.com/containers/image/v5/types"
@@ -191,7 +192,8 @@ func (s *dockerImageSource) fetchManifest(ctx context.Context, tagOrDigest strin
 	if res.StatusCode != http.StatusOK {
 		return nil, "", errors.Wrapf(client.HandleErrorResponse(res), "Error reading manifest %s in %s", tagOrDigest, s.ref.ref.Name())
 	}
-	manblob, err := ioutil.ReadAll(res.Body)
+
+	manblob, err := iolimits.ReadAtMost(res.Body, iolimits.MaxManifestBodySize)
 	if err != nil {
 		return nil, "", err
 	}
@@ -377,7 +379,7 @@ func (s *dockerImageSource) getOneSignature(ctx context.Context, url *url.URL) (
 		} else if res.StatusCode != http.StatusOK {
 			return nil, false, errors.Errorf("Error reading signature from %s: status %d (%s)", url.String(), res.StatusCode, http.StatusText(res.StatusCode))
 		}
-		sig, err := ioutil.ReadAll(res.Body)
+		sig, err := iolimits.ReadAtMost(res.Body, iolimits.MaxSignatureBodySize)
 		if err != nil {
 			return nil, false, err
 		}
@@ -436,7 +438,7 @@ func deleteImage(ctx context.Context, sys *types.SystemContext, ref dockerRefere
 		return err
 	}
 	defer get.Body.Close()
-	manifestBody, err := ioutil.ReadAll(get.Body)
+	manifestBody, err := iolimits.ReadAtMost(get.Body, iolimits.MaxManifestBodySize)
 	if err != nil {
 		return err
 	}
@@ -459,7 +461,7 @@ func deleteImage(ctx context.Context, sys *types.SystemContext, ref dockerRefere
 	}
 	defer delete.Body.Close()
 
-	body, err := ioutil.ReadAll(delete.Body)
+	body, err := iolimits.ReadAtMost(delete.Body, iolimits.MaxErrorBodySize)
 	if err != nil {
 		return err
 	}
--- a/internal/iolimits/iolimits.go
+++ b/internal/iolimits/iolimits.go
@@ -1,0 +1,57 @@
+package iolimits
+
+import (
+	"io"
+	"io/ioutil"
+
+	"github.com/pkg/errors"
+)
+
+// All constants below are intended to be used as limits for `ReadAtMost`. The
+// immediate use-case for limiting the size of in-memory copied data is to
+// protect against OOM DOS attacks as described inCVE-2020-1702. Instead of
+// copying data until running out of memory, we error out after hitting the
+// specified limit.
+const (
+	// megaByte denotes one megabyte and is intended to be used as a limit in
+	// `ReadAtMost`.
+	megaByte = 1 << 20
+	// MaxManifestBodySize is the maximum allowed size of a manifest. The limit
+	// of 4 MB aligns with the one of a Docker registry:
+	// https://github.com/docker/distribution/blob/a8371794149d1d95f1e846744b05c87f2f825e5a/registry/handlers/manifests.go#L30
+	MaxManifestBodySize = 4 * megaByte
+	// MaxAuthTokenBodySize is the maximum allowed size of an auth token.
+	// The limit of 1 MB is considered to be greatly sufficient.
+	MaxAuthTokenBodySize = megaByte
+	// MaxSignatureListBodySize is the maximum allowed size of a signature list.
+	// The limit of 4 MB is considered to be greatly sufficient.
+	MaxSignatureListBodySize = 4 * megaByte
+	// MaxSignatureBodySize is the maximum allowed size of a signature.
+	// The limit of 4 MB is considered to be greatly sufficient.
+	MaxSignatureBodySize = 4 * megaByte
+	// MaxErrorBodySize is the maximum allowed size of an error-response body.
+	// The limit of 1 MB is considered to be greatly sufficient.
+	MaxErrorBodySize = megaByte
+	// MaxConfigBodySize is the maximum allowed size of a config blob.
+	// The limit of 4 MB is considered to be greatly sufficient.
+	MaxConfigBodySize = 4 * megaByte
+	// MaxOpenShiftStatusBody is the maximum allowed size of an OpenShift status body.
+	// The limit of 4 MB is considered to be greatly sufficient.
+	MaxOpenShiftStatusBody = 4 * megaByte
+)
+
+// ReadAtMost reads from reader and errors out if the specified limit (in bytes) is exceeded.
+func ReadAtMost(reader io.Reader, limit int) ([]byte, error) {
+	limitedReader := io.LimitReader(reader, int64(limit+1))
+
+	res, err := ioutil.ReadAll(limitedReader)
+	if err != nil && err != io.EOF {
+		return nil, err
+	}
+
+	if len(res) > limit {
+		return nil, errors.Errorf("exceeded maximum allowed size of %d bytes", limit)
+	}
+
+	return res, nil
+}
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -6,7 +6,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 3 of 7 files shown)


rhatdan:
LGTM

pkg/util/util.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"bufio"
+	"bytes"
+	"io"
+
+	"github.com/pkg/errors"
+)
+
+// MegaByte denotes one megabyte and is intended to be used as a limit in `ReadAtMost`.
+const MegaByte = 1 << 20
```
TomSweeneyRedHat: Should/can this be a conf file entry somewhere?

TomSweeneyRedHat:
The changes LGTM, would like a head nod from @mtrmac and/or @nalind.  My only concern are the outliers for the size,  I wonder if we could/should make the size allowed a configurable option by the end user.

docker/docker_image_src.go
```diff
@@ -377,7 +379,7 @@ func (s *dockerImageSource) getOneSignature(ctx context.Context, url *url.URL) (
 		} else if res.StatusCode != http.StatusOK {
 			return nil, false, errors.Errorf("Error reading signature from %s: status %d (%s)", url.String(), res.StatusCode, http.StatusText(res.StatusCode))
 		}
-		sig, err := ioutil.ReadAll(res.Body)
+		sig, err := util.ReadAtMost(res.Body, 4*util.MegaByte)
```
saschagrunert: I recommend to re-declare two consts which describe their purpose.

vrothberg: I had that initially and didn't like it, as we needed more than two consts. Basically for each individual blob (manifest, config, authtoken, errorresponse size, signature, signature list, etc.). I much prefer to specify the limit in each call and use the generic `util.MegaByte`.

mtrmac: Non-blocking, I’d also prefer a named constant — to make it clear that this value is an arbitrary choice, not something protocol-defined that has to be hard-coded to that exact value.

`notBigMetadataSize` or something like that (covering all of errors, manifests, configs, signatures,…) would still be clearer in purpose than `MegaByte`.

OTOH that can’t be a public API of the util package, it would have to be a separate constant in each package.

docker/docker_image_src.go
```diff
@@ -436,7 +438,7 @@ func deleteImage(ctx context.Context, sys *types.SystemContext, ref dockerRefere
 		return err
 	}
 	defer get.Body.Close()
-	manifestBody, err := ioutil.ReadAll(get.Body)
+	manifestBody, err := util.ReadAtMost(get.Body, 4*util.MegaByte)
```
mtrmac: There really should be a comment somewhere pointing out that docker/distribution has the same limit.

pkg/util/util.go
mtrmac: No, users shouldn’t ever end up having to make a choice.

pkg/util/util.go
```diff
@@ -0,0 +1,37 @@
...
+				break
+			}
+			return nil, err
+		}
+		writeBuffer.WriteByte(b)
+		bytesRead++
+	}
+
+	if bytesRead > limit {
+		return nil, errors.Errorf("exceeded maximum allowed size of %d bytes", limit)
```
mtrmac: If this is going to be a public API, it should be an error recognizable by callers (error type, I guess).

pkg/util/util.go
```diff
@@ -0,0 +1,37 @@
+package util
```
mtrmac: Should this be a part of the public API at all? On the one hand it clearly can be useful elsewhere, OTOH it requires extra care to design it to last.

If this should be public, please make it a separate subpackage, `pkg/iolimits` (??) or something like that; an unspecific `pkg/util` that grows over time does not _quite_ hurt but it will drag in extra code and dependencies for every caller that only wants one of the utilities.

pkg/util/util.go
mtrmac: (It feels weird to have this definition in an, overall, pretty random package. It’s not really a concept specific to this :)

OTOH it works just fine.)

pkg/util/util.go
```diff
@@ -0,0 +1,37 @@
...
+	"io"
+
+	"github.com/pkg/errors"
+)
+
+// MegaByte denotes one megabyte and is intended to be used as a limit in `ReadAtMost`.
+const MegaByte = 1 << 20
+
+// ReadAtMost reads from reader and errors out if the specified limit (in bytes) is exceeded.
+func ReadAtMost(reader io.Reader, limit int) ([]byte, error) {
```
mtrmac: Note to self: Actually review the code.

docker/docker_image_src.go
mtrmac: … “really should be” is, on reflection, too strong. We can live without it — but if there are two different values, it will cause anyone new to the code to wonder about reasons for the difference.

pkg/util/util.go
vrothberg: Good point. I think we should move it to `./internal`. There we could define the constants and don't need to care (much) about providing something ~generic.

pkg/util/util.go
vrothberg: I think it should be `./internal` see https://github.com/containers/image/pull/803#discussion_r372444310

pkg/util/util.go
mtrmac: Sure, `internal` sounds good. Still a subpackage in there, I guess, but `internal/util` would be perfectly fine.

pkg/util/util.go
vrothberg: Now in `./internal/iolimits` :+1:

docker/docker_image_src.go
vrothberg: Everything's in `internal/iolimits` now with individual limits.

vrothberg:
PR updates:
* The new package now resides in `internal/iolimits`
* There are specific constants in the `iolimits` package for each individual blob we're pulling
* `openshift/` is using the new package
* Tar ball has _not_ been changed; I don't think it's an immediate threat

internal/iolimits/iolimits.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+// ReadAtMost reads from reader and errors out if the specified limit (in bytes) is exceeded.
+func ReadAtMost(reader io.Reader, limit int) ([]byte, error) {
+	readBuffer := bufio.NewReader(reader)
+	writeBuffer := bytes.Buffer{}
+
+	bytesRead := 0
+	for bytesRead < limit+1 {
+		b, err := readBuffer.ReadByte()
```
mtrmac: Doing this one byte at a time seems _rather_ expensive with all the bounds checks on every call, and the extra `bufio` buffer is another copy.

Looking at how `ReadAll` is implemented, maybe something like
```go
   limited := io.LimitReader(reader, limit + 1)
   res, err := ioutil.ReadAll(limited)
   if err != nil { … }
   if len >= limit { … }
   return res, nil
```
?

That’s less code, much more likely to be similar to `ioutil.ReadAll` in quirks/behaviors (`WriteByte`, but not `ReadAll`, reports failures by panic), and only does bounds/limit checks only once per `bytes.MinRead` bytes.

internal/iolimits/iolimits.go
```diff
@@ -0,0 +1,67 @@
...
+	// MaxConfigBodySize is the maximum allowed size of a config blob.
+	// The limit of 4 MB is considered to be greatly sufficient.
+	MaxConfigBodySize = 4 * megaByte
+	// MaxOpenShiftStatusBody is the maximum allowed size of an OpenShift status body.
+	// The limit of 4 MB is considered to be greatly sufficient.
+	MaxOpenShiftStatusBody = 4 * megaByte
+)
+
+// ReadAtMost reads from reader and errors out if the specified limit (in bytes) is exceeded.
+func ReadAtMost(reader io.Reader, limit int) ([]byte, error) {
```
mtrmac: (Absolutely non-blocking: It would be nice to have at least a smoke test. Let me write one in the meantime…)

internal/iolimits/iolimits.go
vrothberg: Oh nice! Great idea.

go.sum
```diff
@@ -24,11 +24,13 @@ github.com/containerd/containerd v1.2.10/go.mod h1:bC6axHOhabU15QhwfG7w5PipXdVtM
 github.com/containerd/containerd v1.3.0-beta.2.0.20190828155532-0293cbd26c69/go.mod h1:bC6axHOhabU15QhwfG7w5PipXdVtMXFTttgp+kVtyUA=
 github.com/containerd/continuity v0.0.0-20180216233310-d8fb8589b0e8 h1:ZZOFPzvZO3N0f4LIQvZi68F2XDAMl/gqBfFMVjY6B3Y=
 github.com/containerd/continuity v0.0.0-20180216233310-d8fb8589b0e8/go.mod h1:GL3xCUCBDV3CZiTSEKksMWbLE66hEyuu9qyDOOqM47Y=
+github.com/containerd/continuity v0.0.0-20190426062206-aaeac12a7ffc h1:TP+534wVlf61smEIq1nwLLAjQVEK2EADoW3CX9AuT+8=
 github.com/containerd/continuity v0.0.0-20190426062206-aaeac12a7ffc/go.mod h1:GL3xCUCBDV3CZiTSEKksMWbLE66hEyuu9qyDOOqM47Y=
 github.com/containerd/fifo v0.0.0-20190226154929-a9fb20d87448/go.mod h1:ODA38xgv3Kuk8dQz2ZQXpnv/UZZUHUCL7pnLehbXgQI=
 github.com/containerd/go-runc v0.0.0-20180907222934-5a6d9f37cfa3/go.mod h1:IV7qH3hrUgRmyYrtgEeGWJfWbgcHL9CSRruz2Vqcph0=
 github.com/containerd/ttrpc v0.0.0-20190828154514-0e0f228740de/go.mod h1:PvCDdDGpgqzQIzDW1TphrGLssLDZp2GuS+X5DkEJB8o=
 github.com/containerd/typeurl v0.0.0-20180627222232-a93fcdb778cd/go.mod h1:Cm3kwCdlkCfMSHURc+r6fwoGH6/F1hH3S4sg0rLFWPc=
+github.com/containers/image v3.0.2+incompatible h1:B1lqAE8MUPCrsBLE86J0gnXleeRq8zJnQryhiiGQNyE=
```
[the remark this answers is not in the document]
vrothberg: Will remove.

internal/iolimits/iolimits.go
mtrmac:
```go
package iolimits

import (
	"bytes"
	"math/rand"
	"testing"

)

func TestReadAtMost(t *testing.T) {
	for _, c := range []struct {
		input, limit  int
		shouldSucceed bool
	}{
		{0, 0, true},
		{0, 1, true},
		{1, 0, false},
		{1, 1, true},
		{bytes.MinRead*5 - 1, bytes.MinRead * 5, true},
		{bytes.MinRead * 5, bytes.MinRead * 5, true},
		{bytes.MinRead*5 + 1, bytes.MinRead * 5, false},
	} {
		input := make([]byte, c.input)
		_, err := rand.Read(input)
		require.NoError(t, err)
		result, err := ReadAtMost(bytes.NewReader(input), c.limit)
		if c.shouldSucceed {
			assert.NoError(t, err)
			assert.Equal(t, result, input)
		} else {
			assert.Error(t, err)
		}
	}
}
```

internal/iolimits/iolimits.go
vrothberg: Updated.

internal/iolimits/iolimits.go
vrothberg: Thanks!

mtrmac:
Continues here: #805.

vrothberg:
Thanks for pushing it over the finish line, @mtrmac !

Issue #799, opened 2020-01-23
Drop the architecture check in checkImageDestinationForCurrentRuntime


mtrmac:
Per recent Buildah / Podman reports, apparently pulling different-architecture images is fairly common; so, revert/undo #752.

vrothberg:
I suggest to not revert the previous changes entirely but to `logrus.Info` instead of returning the error. The motivation would be to make users aware of the issue. I am somehow quite surprised that the k8s pause image issue hasn't been detected earlier.

mtrmac:
Sure.

vrothberg:
Fixed with https://github.com/containers/image/pull/804. Sorry for missing that in the commit message!

PR #804, opened 2020-01-31
copy: relax platform checks

vrothberg:
Relax the platform checks of the source images architecture and OS when
copying.  Some images do not set these attributes correctly in their
config which unexpectedly broke previously running deployments.

Instead of returning an error, just info-log about the mismatch.

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -709,7 +709,7 @@ func checkImageDestinationForCurrentRuntime(ctx context.Context, sys *types.Syst
 			wantedOS = sys.OSChoice
 		}
 		if wantedOS != c.OS {
-			return fmt.Errorf("Image operating system mismatch: image uses %q, expecting %q", c.OS, wantedOS)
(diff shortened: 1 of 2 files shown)


vrothberg:
@baude @mheon @mtrmac PTAL

copy/copy.go
```diff
@@ -709,15 +709,15 @@ func checkImageDestinationForCurrentRuntime(ctx context.Context, sys *types.Syst
 			wantedOS = sys.OSChoice
 		}
 		if wantedOS != c.OS {
-			return fmt.Errorf("Image operating system mismatch: image uses %q, expecting %q", c.OS, wantedOS)
+			logrus.Infof("Image operating system mismatch: image uses %q, expecting %q", c.OS, wantedOS)
```
mtrmac: Absolutely non-blocking: We’ve had the OS check for a long time without complaints… not sure that one needs to be demoted. OTOH, it’s pretty hard to _accidentally_ refer to a Windows image on Linux computers, and the consistency with architecture is nice.

mtrmac:
For the record, fixes #799.

PR #809, opened 2020-02-04
copyMultipleImages: re-serialize the list if we only want to change its MIME type

nalind:
If the only change we were making to a manifest list was to change its MIME type (i.e., other updates to the list didn't materially change its contents), we wouldn't re-serialize the list, so we wouldn't actually change its MIME type.

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -380,6 +380,7 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
 			return nil, "", errors.Wrap(err, "Can not copy signatures")
 		}
 	}
+	canModifyManifestList := (len(sigs) == 0)
 
 	// Determine if we'll need to convert the manifest list to a different format.
 	forceListMIMEType := options.ForceManifestMIMEType
@@ -394,7 +395,6 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
 		return nil, "", errors.Wrapf(err, "Error determining manifest list type to write to destination")
 	}
 	if selectedListType != list.MIMEType() {
-		canModifyManifestList := (len(sigs) == 0)
 		if !canModifyManifestList {
 			return nil, "", errors.Errorf("Error: manifest list must be converted to type %q to be written to destination, but that would invalidate signatures", selectedListType)
 		}
@@ -451,12 +451,6 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
 		return nil, "", errors.Wrapf(err, "Error updating manifest list")
 	}
 
-	// Check if the updates meaningfully changed the list of images.
-	listIsModified := false
-	if !reflect.DeepEqual(list.Instances(), originalList.Instances()) {
-		listIsModified = true
-	}
-
 	// Perform the list conversion.
 	if selectedListType != list.MIMEType() {
 		list, err = list.ConvertToMIMEType(selectedListType)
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

copy/copy.go
```diff
@@ -463,6 +463,7 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
 		if err != nil {
 			return nil, "", errors.Wrapf(err, "Error converting manifest list to list with MIME type %q", selectedListType)
 		}
+		listIsModified = true
```
mtrmac: This is easy to forget in any future updates, and we are already using `reflect.DeepEqual` above. Could the `if listIsModified` be replaced with something like `reflect.DeepEqual(list, original)`? That would solve the problem once for all.

Using `reflect` is ugly in that it will also compare private fields, but at least it fails safe (preferring a conversion even if not necessary); we could also add a `List.Equals()` method, but that would have the same risk of forgetting to compare some members.

copy/manifest.go
```diff
@@ -127,14 +127,14 @@ func isMultiImage(ctx context.Context, img types.UnparsedImage) (bool, error) {
 // forced value, and returns the MIME type to which we should convert the list
 // of manifests, whether we are converting to it or using it unmodified.
 func (c *copier) determineListConversion(currentListMIMEType string, destSupportedMIMETypes []string, forcedListMIMEType string) (string, error) {
-	// If we're forcing it, we prefer the forced value over everything else.
-	if forcedListMIMEType != "" {
-		return forcedListMIMEType, nil
```
mtrmac: (Aside from adding debug output, this has the side effect of rejecting unrecognized `forcedListMIMEType` values. That’s not bad, just surprising given the commit description.)

nalind: Hmm, yeah, I'll take another look at this bit.

mtrmac: I’m sorry, to be clear, I’m fine with the change itself — it would fail anyway, and this way it fails before copying all of the images. It’s just the mismatch with the description that surprised me.

copy/copy.go
nalind: Would it be simpler to serialize both, and then compare the byte slices?

mtrmac: Good idea, that would work nicely. (The Go implementation does not _promise_ that, but the JSON marshaller serializes fields in the order they are declared, and anyway, it would fail safe as well.)

nalind: Trying that.

copy/manifest.go
nalind: Oh, okay.  I'd switched that up a bit, but I think I can change it back and fix the comment and commit log.

copy/copy.go
```diff
@@ -465,12 +459,24 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
...
+	// Check if the updates or a type conversion meaningfully changed the list of images.
+	updatedManifestList, err := list.Serialize()
+	if err != nil {
+		return nil, "", errors.Wrapf(err, "Error encoding updated manifest list (%q: %#v)", list.MIMEType(), list.Instances())
+	}
+	originalManifestList, err := originalList.Serialize()
+	if err != nil {
+		return nil, "", errors.Wrapf(err, "Error encoding original manifest list for comparison (%q: %#v)", originalList.MIMEType(), originalList.Instances())
+	}
+	listIsModified := !bytes.Equal(updatedManifestList, originalManifestList)
```
mtrmac: (Absolutely non-blocking: This variable could be inlined into its only use.)

copy/copy.go
```diff
@@ -465,12 +459,24 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
...
+
 	// If we can't use the original value, but we have to change it, flag an error.
 	if listIsModified {
-		manifestList, err = list.Serialize()
-		if err != nil {
-			return nil, "", errors.Wrapf(err, "Error encoding updated manifest list (%q: %#v)", list.MIMEType(), list.Instances())
+		canModifyManifestList := (len(sigs) == 0)
```
mtrmac: This sanity check is nice.

`canModifyManifestList` is computed earlier as well, can it be done only once to make sure the logic doesn’t diverge?

copy/copy.go
nalind: Changed it.

copy/copy.go
nalind: Switched to reusing a `canModifyManifestList`.

PR #814, opened 2020-02-07
Add NewReference to a few transports, + storage.GetStoreIfSet

mtrmac:
`New*Reference` allows creating references from well-typed data without string formatting and parsing; this adds it to a few transports that were missing the functionality.

Also adds `storage.Transport.GetStoreIfSet` (naming?) to allow checking that the default store is not set to two different values in a single application.

### DIFF
--- a/storage/storage_reference.go
+++ b/storage/storage_reference.go
@@ -30,6 +30,14 @@ func newReference(transport storageTransport, named reference.Named, id string) 
 	if named == nil && id == "" {
 		return nil, ErrInvalidReference
 	}
+	if named != nil && reference.IsNameOnly(named) {
+		return nil, errors.Wrapf(ErrInvalidReference, "reference %s has neither a tag nor a digest", named.String())
+	}
+	if id != "" {
+		if err := validateImageID(id); err != nil {
+			return nil, errors.Wrapf(ErrInvalidReference, "invalid ID value %q: %v", id, err)
+		}
+	}
 	// We take a copy of the transport, which contains a pointer to the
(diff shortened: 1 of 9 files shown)


storage/storage_reference.go
```diff
@@ -30,6 +30,16 @@ func newReference(transport storageTransport, named reference.Named, id string)
 	if named == nil && id == "" {
 		return nil, ErrInvalidReference
 	}
+	if named != nil {
```
rhatdan: Nit: Why not use the && here?  Used it above.

if named != nil && reference.IsNameOnly(named) {
 			return nil, errors.Wrapf(ErrInvalidReference, "reference %s has neither a tag nor a digest", named.String())
}

storage/storage_reference.go
```diff
@@ -30,6 +30,16 @@ func newReference(transport storageTransport, named reference.Named, id string)
...
+		if reference.IsNameOnly(named) {
+			return nil, errors.Wrapf(ErrInvalidReference, "reference %s has neither a tag nor a digest", named.String())
+		}
+	}
+	if id != "" {
```
rhatdan: Same comment as above.

mtrmac: It’s not possible to do `if (expression) && newVariable := …; expression(newVariable)`.

storage/storage_reference.go
mtrmac: Thanks, fixed.

rhatdan:
LGTM

PR #817, opened 2020-02-10
Add $HOME/.config/containers/certs.d to perHostCertDirPath

rhatdan:
We want to allow users to store certs in their homedir when running in rootless mode.
We want rootless podman and rootless buildah to add $HOME/.config/containers/certs.d
to the search path for certificates by default.

Currently there is no way for a non privileged user to get certs without being root on
the system or specify the certs dir on ever call.

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -21,6 +21,7 @@ import (
 	"github.com/containers/image/v5/pkg/sysregistriesv2"
 	"github.com/containers/image/v5/pkg/tlsclientconfig"
 	"github.com/containers/image/v5/types"
+	"github.com/containers/storage/pkg/homedir"
 	clientLib "github.com/docker/distribution/registry/client"
 	"github.com/docker/go-connections/tlsconfig"
 	digest "github.com/opencontainers/go-digest"
@@ -51,7 +52,18 @@ const (
 	backoffMaxDelay      = 60 * time.Second
 )
 
-var systemPerHostCertDirPaths = [2]string{"/etc/containers/certs.d", "/etc/docker/certs.d"}
+type certPath struct {
+	path     string
+	absolute bool
+}
+
+var (
+	homeCertDir     = filepath.FromSlash(".config/containers/certs.d")
+	perHostCertDirs = []certPath{
+		{path: "/etc/containers/certs.d", absolute: true},
+		{path: "/etc/docker/certs.d", absolute: true},
+	}
+)
 
 // extensionSignature and extensionSignatureList come from github.com/openshift/origin/pkg/dockerregistry/server/signaturedispatcher.go:
 // signature represents a Docker image signature.
@@ -166,11 +178,12 @@ func dockerCertDir(sys *types.SystemContext, hostPort string) (string, error) {
 		hostCertDir     string
 		fullCertDirPath string
 	)
-	for _, systemPerHostCertDirPath := range systemPerHostCertDirPaths {
-		if sys != nil && sys.RootForImplicitAbsolutePaths != "" {
-			hostCertDir = filepath.Join(sys.RootForImplicitAbsolutePaths, systemPerHostCertDirPath)
+
+	for _, perHostCertDir := range append([]certPath{{path: filepath.Join(homedir.Get(), homeCertDir), absolute: false}}, perHostCertDirs...) {
+		if sys != nil && sys.RootForImplicitAbsolutePaths != "" && perHostCertDir.absolute {
+			hostCertDir = filepath.Join(sys.RootForImplicitAbsolutePaths, perHostCertDir.path)
 		} else {
-			hostCertDir = systemPerHostCertDirPath
+			hostCertDir = perHostCertDir.path
 		}
 
 		fullCertDirPath = filepath.Join(hostCertDir, hostPort)
--- a/docker/docker_client_test.go
+++ b/docker/docker_client_test.go
@@ -18,7 +18,7 @@ func TestDockerCertDir(t *testing.T) {
 	const rootPrefix = "/root/prefix"
 	const registryHostPort = "thishostdefinitelydoesnotexist:5000"
 
-	systemPerHostResult := filepath.Join(systemPerHostCertDirPaths[len(systemPerHostCertDirPaths)-1], registryHostPort)
+	systemPerHostResult := filepath.Join(perHostCertDirs[len(perHostCertDirs)-1].path, registryHostPort)
 	for _, c := range []struct {
 		sys      *types.SystemContext
 		expected string
--- a/docs/containers-certs.d.5.md
+++ b/docs/containers-certs.d.5.md
@@ -4,7 +4,7 @@
 containers-certs.d - Directory for storing custom container-registry TLS configurations
 
 # DESCRIPTION
-A custom TLS configuration for a container registry can be configured by creating a directory under `/etc/containers/certs.d`.
+A custom TLS configuration for a container registry can be configured by creating a directory under `$HOME/.config/containers/certs.d` or `/etc/containers/certs.d`.
 The name of the directory must correspond to the `host:port` of the registry (e.g., `my-registry.com:5000`).
 
 ## Directory Structure
--- a/docs/containers-registries.conf.5.md
+++ b/docs/containers-registries.conf.5.md
@@ -207,6 +207,9 @@ the destination registry is unambiguous. Pulling by digest
 (i.e., quay.io/repository/name@digest) further eliminates the ambiguity of
 tags.
 
+# SEE ALSO
+  containers-certs.d(5)
+
 # HISTORY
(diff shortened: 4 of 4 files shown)


mtrmac:
If it makes sense to have per-user configuration of certificates (which is not 100% clear but it probably does), why should it be only applicable to rootless, and not an universal default? (Compare https://github.com/containers/skopeo/issues/800 ).

And even if it had to be rootless-specific, is the right design having Podman+Buildah hard-code a path, or should that ultimately just be a boolean? (Or maybe the hard-coded path should exist but live in c/common, I guess.)

rhatdan:
It does not need to be rootless specific, I was just suggesting the reason I was adding it.  Eventually we might want this customizable via containers.conf, but for now, we are just placing content in ~/.config/containers for use by rootless users.

mtrmac:
I’d much prefer for the per-user path to be used by c/image by default without having every consumer of c/image to opt in, and to independently hard-code the path. I can’t see much downside to having per-user configuration, even if run as root (I suppose determining which configuration is used could be a bit confusing with `sudo`).

WRT `containers.conf`, are you suggesting that it should point at locations of other config files, by explicit paths inside `containers.conf` (i.e. that the file is not portable from one user account to another)? Or maybe the other config files should be located in exactly the same directory (i.e. use one lookup path/logic for config files)? Or that the TLS blobs (keys, certificates) should be recorded inside `containers.conf` instead of using separate files on the filesystem?

rhatdan:
I am fine with hard coding this to $HOME/.config/containers if this is what you want.

docker/docker_client.go
```diff
@@ -165,8 +165,13 @@ func dockerCertDir(sys *types.SystemContext, hostPort string) (string, error) {
 	var (
 		hostCertDir     string
 		fullCertDirPath string
+		homeCertDir     string
 	)
-	for _, systemPerHostCertDirPath := range systemPerHostCertDirPaths {
+	home := os.Getenv("HOME")
+	if home != "" {
+		homeCertDir = filepath.Join(home, ".config", "containers", ".certs.d")
```
mtrmac: Please make this a global constant defined near `systemPerHostCertDirPaths` (compare `dockerHomePath`).

docker/docker_client.go
```diff
@@ -165,8 +165,13 @@ func dockerCertDir(sys *types.SystemContext, hostPort string) (string, error) {
...
+	}
+	for _, systemPerHostCertDirPath := range append([]string{homeCertDir}, systemPerHostCertDirPaths...) {
```
mtrmac: - I don’t think this works correctly for `homeCertDir == ""`
- `RootForImplicitAbsolutePaths` should not be used for the home directory.
- We use `docker/docker/pkg/homedir` when looking up credentials, it would be nice to be consistent.

docker/docker_client.go
rhatdan: Well made it a global var, which matches dockerHomePath

docker/docker_client.go
rhatdan: Fixed 1 and 3, not sure we should hack around 2.

rhatdan:
@mtrmac @vrothberg PTAL
Is there anything else I need to do to move this forward, would close a libpod issue

rhatdan:
Fixes: https://github.com/containers/libpod/issues/5006

docker/docker_client.go
```diff
@@ -22,6 +22,7 @@ import (
 	"github.com/containers/image/v5/pkg/tlsclientconfig"
 	"github.com/containers/image/v5/types"
 	clientLib "github.com/docker/distribution/registry/client"
+	"github.com/docker/docker/pkg/homedir"
```
vrothberg: Please use `c/storage/pkg/homedir`. It's not yet released though but merged.

docker/docker_client.go
mtrmac: Yes, 2. still needs handling correctly. (See `legacyFormat` in `GetAuthentication` for one way to structure the code without repeating the consuming loop.)

docker/docker_client.go
```diff
@@ -51,7 +52,10 @@ const (
 	backoffMaxDelay      = 60 * time.Second
 )
 
-var systemPerHostCertDirPaths = [2]string{"/etc/containers/certs.d", "/etc/docker/certs.d"}
+var (
+	homeCertDir               = filepath.FromSlash(".config/containers/certs.d")
+	systemPerHostCertDirPaths = []string{filepath.Join(homedir.Get(), homeCertDir), "/etc/containers/certs.d", "/etc/docker/certs.d"}
```
mtrmac: Evaluating `$HOME` at this point isn’t great — e.g. it makes it impossible to write unit tests with overridden $HOME.

docker/docker_client.go
rhatdan: Updated

docker/docker_client.go
rhatdan: Ok moved the call out of init and just before it is used.

docker/docker_client.go
```diff
@@ -51,7 +52,10 @@ const (
 	backoffMaxDelay      = 60 * time.Second
 )
 
-var systemPerHostCertDirPaths = [2]string{"/etc/containers/certs.d", "/etc/docker/certs.d"}
+var (
+	homeCertDir               = filepath.FromSlash(".config/containers/certs.d")
+	systemPerHostCertDirPaths = []string{"/etc/containers/certs.d", "/etc/docker/certs.d"}
```
TomSweeneyRedHat: are these documented anywhere?

docker/docker_client.go
rhatdan: I think I have it the way you want it now.

TomSweeneyRedHat:
Would be nice to fix the "ssytem" typo in the PR title.

docker/docker_client.go
```diff
@@ -51,7 +52,18 @@ const (
 	backoffMaxDelay      = 60 * time.Second
 )
 
-var systemPerHostCertDirPaths = [2]string{"/etc/containers/certs.d", "/etc/docker/certs.d"}
+type certPath struct {
+	path         string
+	legacyFormat bool
```
mtrmac: `legacyFormat` is not what that boolean means — maybe `absolute[Path]`.

docker/docker_client.go
```diff
@@ -166,11 +178,13 @@ func dockerCertDir(sys *types.SystemContext, hostPort string) (string, error) {
 		hostCertDir     string
 		fullCertDirPath string
 	)
-	for _, systemPerHostCertDirPath := range systemPerHostCertDirPaths {
-		if sys != nil && sys.RootForImplicitAbsolutePaths != "" {
-			hostCertDir = filepath.Join(sys.RootForImplicitAbsolutePaths, systemPerHostCertDirPath)
+
+	systemPerHostCertDirs = append([]certPath{{path: filepath.Join(homedir.Get(), homeCertDir), legacyFormat: false}}, systemPerHostCertDirs...)
+	for _, systemPerHostCertDir := range systemPerHostCertDirs {
```
mtrmac: These paths are no longer all `system`. Maybe `s/systemP/p/`, for both variables.

docker/docker_client.go
```diff
@@ -166,11 +178,13 @@ func dockerCertDir(sys *types.SystemContext, hostPort string) (string, error) {
 		hostCertDir     string
 		fullCertDirPath string
 	)
-	for _, systemPerHostCertDirPath := range systemPerHostCertDirPaths {
-		if sys != nil && sys.RootForImplicitAbsolutePaths != "" {
-			hostCertDir = filepath.Join(sys.RootForImplicitAbsolutePaths, systemPerHostCertDirPath)
+
+	systemPerHostCertDirs = append([]certPath{{path: filepath.Join(homedir.Get(), homeCertDir), legacyFormat: false}}, systemPerHostCertDirs...)
```
mtrmac: Doesn’t this keep appending to the global variable? (Renaming, see below, would help solve this.)

docker/docker_client.go
mtrmac: Good point, updating `docs/containers-certs.d.5.md` would be nice.

rhatdan: container-certs.d.conf has info.

docker/docker_client.go
rhatdan: Fixed

TomSweeneyRedHat:
@rhatdan your finger just won't let you type system in the title will they?

TomSweeneyRedHat:
LGTM

rhatdan:
@mtrmac @vrothberg I think this is ready to go.  PTAL

docker/docker_client.go
rhatdan: fixed

rhatdan:
@mtrmac @vrothberg PTAL  Can we get this merged?

rhatdan:
@vrothberg @mtrmac Tests are passing again, can we get this merged?

PR #819, opened 2020-02-14
Add client support for opencontainers/distribution-spec

rchincha:
Add a new transport with URL prefix called "dist://"

This is an implementation of an OCI image repository client, based on the
opencontainers/distribution-spec.

For a full commit history, please see https://github.com/anuvu/image/commits/master

Consequently, a container image tool like skopeo when linked with this
version of containers/image can do:

"skopeo copy dist://src-server dist://dest-server"

where src-server and dest-server are both OCI dist-spec compliant
servers.

For the current state of server-side implementations of opencontainers/distribution-spec,
please see https://oci.bloodorange.io/

Signed-off-by: Ramkumar Chinchani <rchincha@cisco.com>
Signed-off-by: Serge Hallyn <shallyn@cisco.com>

### DIFF
--- a/dist/dist_client.go
+++ b/dist/dist_client.go
@@ -1,0 +1,500 @@
+package dist
+
+import (
+	"bytes"
+	"crypto/sha256"
+	"crypto/tls"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"os"
+	"path/filepath"
+	"strconv"
+
+	"github.com/containers/image/v5/pkg/docker/config"
+	"github.com/containers/image/v5/pkg/tlsclientconfig"
+	"github.com/containers/image/v5/types"
+	"github.com/opencontainers/go-digest"
+	ispec "github.com/opencontainers/image-spec/specs-go/v1"
+
+	"github.com/pkg/errors"
+)
+
+type OciRepo struct {
+	url       url.URL
+	ref       *distReference
+	authCreds string
+	client    *http.Client
+}
+
+//nolint (funlen)
+func NewOciRepo(ref *distReference, sys *types.SystemContext) (r OciRepo, err error) {
+	server := "127.0.0.1"
+	port := "8080"
+	hostName := ""
+
+	if ref.server != "" {
+		server = ref.server
+		hostName = server
+	}
+
+	if ref.port != -1 {
+		port = fmt.Sprintf("%d", ref.port)
+		hostName += ":" + port
+	}
+
+	insecureSkipVerify := false
+	if sys != nil {
+		insecureSkipVerify = (sys.DockerInsecureSkipTLSVerify == types.OptionalBoolTrue)
+	}
+
+	tlsClientConfig := &tls.Config{
+		MinVersion:               tls.VersionTLS10,
+		PreferServerCipherSuites: true,
+		InsecureSkipVerify:       insecureSkipVerify,
+	}
+
+	certDir, err := ociCertDir(sys, hostName)
+	if err != nil {
+		return r, err
+	}
+	if err := tlsclientconfig.SetupCertificates(certDir, tlsClientConfig); err != nil {
+		return r, err
+	}
+
+	transport := &http.Transport{TLSClientConfig: tlsClientConfig}
+	client := &http.Client{Transport: transport}
+	creds := ""
+	if sys != nil {
+		if sys.DockerAuthConfig != nil {
+			a := sys.DockerAuthConfig
+			creds = base64.StdEncoding.EncodeToString([]byte(a.Username + ":" + a.Password))
+		} else {
+			registry := fmt.Sprintf("%s:%s", server, port)
+			if username, password, err := config.GetAuthentication(sys, registry); err == nil {
+				creds = base64.StdEncoding.EncodeToString([]byte(username + ":" + password))
+			}
+		}
+	}
+
+	r = OciRepo{ref: ref, authCreds: creds, client: client}
+
+	ping := func(scheme string) error {
+		u := url.URL{Scheme: scheme, Host: fmt.Sprintf("%s:%s", server, port), Path: fmt.Sprintf("/v2/")}
+		resp, err := client.Get(u.String())
+		if err != nil {
+			return err
+		}
+		defer resp.Body.Close()
+		if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusUnauthorized {
+			return errors.Errorf("Error pinging registry %s:%s, response code %d (%s)",
+				server, port, resp.StatusCode, http.StatusText(resp.StatusCode))
+		}
+		return nil
+	}
+
+	scheme := "https"
+	err = ping(scheme)
+	if err != nil && insecureSkipVerify {
+		scheme = "http"
+		err = ping(scheme)
+	}
+	if err != nil {
+		return r, errors.Wrap(err, "unable to ping registry")
+	}
+
+	r.url = url.URL{Scheme: scheme, Host: fmt.Sprintf("%s:%s", server, port)}
+	return r, nil
+}
+
+func (o *OciRepo) GetManifest() ([]byte, *ispec.Manifest, error) {
+	name := o.ref.name
+	tag := o.ref.tag
+	m := &ispec.Manifest{}
+
+	var body []byte
+
+	manifestURI, err := url.Parse(fmt.Sprintf("/v2/%s/manifests/%s", name, tag))
+	if err != nil {
+		return body, m, errors.Wrapf(err, "couldn't parse manifest url for repo %s and image %s", name, tag)
+	}
+
+	manifestURI = o.url.ResolveReference(manifestURI)
+	req, err := http.NewRequest(http.MethodGet, manifestURI.String(), nil)
+
+	if err != nil {
+		return body, m, errors.Wrapf(err, "Couldn't create DELETE request for %v", manifestURI)
+	}
+
+	if o.authCreds != "" {
+		req.Header.Add("Authorization", "Basic "+o.authCreds)
+	}
+
+	req.Header.Add("Accept", "application/vnd.oci.image.manifest.v1+json")
+	resp, err := o.client.Do(req)
+
+	if err != nil {
+		return body, m, errors.Wrapf(err, "Error getting manifest %s %s from %v", name, tag, o.url)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return body, m, fmt.Errorf("bad return code %d getting manifest", resp.StatusCode)
+	}
+
+	body, err = ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return body, m, errors.Wrapf(err, "Error reading response body for %s", tag)
+	}
+
+	err = json.Unmarshal(body, m)
+
+	if err != nil {
+		return body, m, errors.Wrap(err, "Failed decoding response")
+	}
+
+	return body, m, nil
+}
+
+func (o *OciRepo) RemoveManifest() error {
+	name := o.ref.name
+	tag := o.ref.tag
+	manifestURI, err := url.Parse(fmt.Sprintf("/v2/%s/manifests/%s", name, tag))
+
+	if err != nil {
+		return errors.Wrapf(err, "couldn't parse manifest url for repo %s and image %s", name, tag)
+	}
+
+	manifestURI = o.url.ResolveReference(manifestURI)
+	req, err := http.NewRequest(http.MethodDelete, manifestURI.String(), nil)
+
+	if err != nil {
+		return errors.Wrapf(err, "Couldn't create DELETE request for %v", manifestURI)
+	}
+
+	if o.authCreds != "" {
+		req.Header.Add("Authorization", "Basic "+o.authCreds)
+	}
+
+	resp, err := o.client.Do(req)
+	if err != nil {
+		return errors.Wrapf(err, "Error deleting manifest")
+	}
+
+	if resp.StatusCode != http.StatusAccepted {
+		return fmt.Errorf("server returned unexpected code %d", resp.StatusCode)
+	}
+
+	defer resp.Body.Close()
+
+	return nil
+}
+
+func (o *OciRepo) PutManifest(body []byte) error {
+	name := o.ref.name
+	tag := o.ref.tag
+	manifestURI, err := url.Parse(fmt.Sprintf("/v2/%s/manifests/%s", name, tag))
+
+	if err != nil {
+		return errors.Wrapf(err, "couldn't parse manifest url for repo %s and image %s", name, tag)
+	}
+
+	manifestURI = o.url.ResolveReference(manifestURI)
+	req, err := http.NewRequest(http.MethodPut, manifestURI.String(), bytes.NewReader(body))
+
+	if err != nil {
+		return errors.Wrapf(err, "Couldn't create PUT request for %v", manifestURI)
+	}
+
+	if o.authCreds != "" {
+		req.Header.Add("Authorization", "Basic "+o.authCreds)
+	}
+
+	req.Header.Set("Content-Type", "application/vnd.oci.image.manifest.v1+json")
+	resp, err := o.client.Do(req)
+
+	if err != nil {
+		return errors.Wrapf(err, "Error posting manifest")
+	}
+
+	if resp.StatusCode != http.StatusCreated {
+		return fmt.Errorf("server returned unexpected code %d", resp.StatusCode)
+	}
+
+	defer resp.Body.Close()
+
+	return nil
+}
+
+//HEAD /v2/<name>/blobs/<digest>  -> 200 (has layer)
+func (o *OciRepo) HasLayer(ldigest string) bool {
+	name := o.ref.name
+	blobURI, err := url.Parse(fmt.Sprintf("/v2/%s/blobs/%s", name, ldigest))
+
+	if err != nil {
+		return false
+	}
+
+	blobURI = o.url.ResolveReference(blobURI)
+	req, err := http.NewRequest(http.MethodHead, blobURI.String(), nil)
+
+	if err != nil {
+		return false
+	}
+
+	if o.authCreds != "" {
+		req.Header.Add("Authorization", "Basic "+o.authCreds)
+	}
+
+	resp, err := o.client.Do(req)
+
+	if err != nil {
+		return false
+	}
+
+	defer resp.Body.Close()
+
+	return resp.StatusCode == http.StatusOK
+}
+
+func (o *OciRepo) GetLayer(ldigest string) (io.ReadCloser, int64, error) {
+	name := o.ref.name
+	blobURI, err := url.Parse(fmt.Sprintf("/v2/%s/blobs/%s", name, ldigest))
+
+	if err != nil {
+		return nil, -1, errors.Wrapf(err, "couldn't parse URL for repo %s blob %s", name, ldigest)
+	}
+
+	blobURI = o.url.ResolveReference(blobURI)
+	req, err := http.NewRequest("GET", blobURI.String(), nil)
+
+	if err != nil {
+		return nil, -1, errors.Wrapf(err, "Couldn't create GET request for %v", blobURI)
+	}
+
+	if o.authCreds != "" {
+		req.Header.Add("Authorization", "Basic "+o.authCreds)
+	}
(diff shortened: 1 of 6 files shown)


rchincha:
Additional discussion about this topic at:
https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/avpDcOncnIE

tych0:
It might be nice to call this "oci-dist" or something; it's not really clear what "dist" means.

mtrmac:
Thanks for the PR. I’ll make a quick pass through the code and add a few notes in a while.

Most importantly, though, are there actually any real differences between `docker://` and the proposed transport?
- You yourself say in the quoted thread that
    > That said the actual differences between docker:// and proposed dist:// transports are very slim
- The original motivation for the PR, that `Content-Type` used when writing manifests is Docker-specific, is simply incorrect; `dockerImageDestination.SupportedManifestMIMETypes` does support both OCI manifests _and_ indexes (unlike this PR, which only considers manifests, i.e. can’t handle multi-arch images), and the OCI types will be used by `c/image/copy.Image` if:
  - the original image is OCI, or
  - OCI conversion was explicitly required by `copy.Options.ForceManifestMIMEType`, or
  - the source image is not OCI, but the destination registry rejects Docker manifest formats.
- Looking at the commit history https://github.com/opencontainers/distribution-spec/commits/master/spec.md , the changes are overwhelmingly trivial clean-ups with no functional impact (the only non-trivial change I have noticed is the renaming of the session ID parameter during uploads, but the previous code didn’t use that at all anyway).

From a maintenance point of view, it is **rather burdensome** to have two significantly different implementations of substantially the same functionality, including one of the two transports that are most important and most often updated, for little end-user benefit. I don’t doubt that _eventually_ there will be a fairly significant divergence, and I’m sure the `docker://` transport name should eventually be retired in favor of a name that references the OCI distribution specification in some way, but in my view, neither of that, at least currently justifies:
  -  ~doubling the maintenance overhead
  - introducing the OCI transport in a way that has noticeably fewer features — and that makes it unattractive for users to even start to use the new transport, when they can use `docker://` that auto-detects OCI support in most cases, and has more features. E.g.: missing support for  `registries.conf` (registry blocking, redirection, mirrors), bearer token authentication (required for anonymous access to Docker Hub at least), `Retry-After`, blob info cache, multi-arch image, external layer

If it makes sense to introduce the new transport name and syntax right now, at all (which is unclear, especially WRT bothering users with the transition, given ~zero end-user benefit),but I’m open to considering that), I’m **very insistent** that it should share 99% of the code with the existing `docker://` transport (probably by moving most of the implementation into a shared `c/image/internal/…` subpackage, and using it in both). Taking partial copies of the `docker://` transport that are stripped of much of the functionality, and linking it to a separate HTTP client implementation that does mostly the same things, but differently, and having to maintain both is just not at all attractive, in my personal opinion, right now. (Note that other maintainers like @vrothberg might well disagree.)

But my first instinct is that the current `docker://` transport can technically do, or should be easy to modify to do, everything that is necessary, and there’s just no _technical_ need for this; all that I can see so far is the PR aspect of using a different name. Sure, that’s something but it’s also not all that compelling from a technical point of view, and I’d prefer to discuss that before delving into implementation details.

dist/dist_client.go
```diff
@@ -0,0 +1,500 @@
...
+	if resp.StatusCode != http.StatusOK {
+		return body, m, fmt.Errorf("bad return code %d getting manifest", resp.StatusCode)
+	}
+
+	body, err = ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return body, m, errors.Wrapf(err, "Error reading response body for %s", tag)
+	}
+
+	err = json.Unmarshal(body, m)
```
mtrmac: Why? Nothing uses the value AFAICS, and it’s, in general, not the job of the transport to worry about contents of the data (unless the transport is simulating something or doing internal conversions, I guess, like the `docker/tarfile` ones, and arguably that code should be moved into the generic conversion mechanism instead).

dist/dist_client.go
```diff
@@ -0,0 +1,500 @@
...
+	}
+
+	manifestURI = o.url.ResolveReference(manifestURI)
+	req, err := http.NewRequest(http.MethodDelete, manifestURI.String(), nil)
+
+	if err != nil {
+		return errors.Wrapf(err, "Couldn't create DELETE request for %v", manifestURI)
+	}
+
+	if o.authCreds != "" {
```
mtrmac: (I can’t quite see the point of having code like this copy&pasted all over the place, when the original has this factored out to a single `dockerClient.makeRequest*` path.)

dist/dist_client.go
```diff
@@ -0,0 +1,500 @@
...
+//nolint (gocognit)
+func (o *OciRepo) CompleteLayer(session *url.URL, stream io.Reader) (digest.Digest, int64, error) {
+	digester := sha256.New()
+	hashReader := io.TeeReader(stream, digester)
+	// using "chunked" upload
+	count := int64(0)
+	for {
+		const maxSize = 10 * 1024 * 1024
+		var buf bytes.Buffer
+		size, err := io.CopyN(&buf, hashReader, maxSize)
```
mtrmac: What’s the benefit of using chunked upload instead of a single long-lived request? AFAICS this only adds overhead with no benefit I can see (e.g. I could imagine that chunked upload might allow restarting on partial failure, but this code doesn’t do that).

dist/dist_client.go
```diff
@@ -0,0 +1,500 @@
...
+}
+
+// ociCertDir returns a path to a directory to be consumed by
+// tlsclientconfig.SetupCertificates() depending on ctx and hostPort.
+func ociCertDir(sys *types.SystemContext, hostPort string) (string, error) {
+	if sys != nil && sys.DockerCertPath != "" {
+		return sys.DockerCertPath, nil
+	}
+
+	if sys != nil && sys.DockerPerHostCertDirPath != "" {
```
mtrmac: Using the `Docker*` configuration options does not quite look good, especially not without any documentation about that effect.

dist/image_dest.go
```diff
@@ -0,0 +1,121 @@
...
+	return o.ref
+}
+
+func (o *distImageDest) Close() error {
+	return nil
+}
+
+func (o *distImageDest) SupportedManifestMIMETypes() []string {
+	return []string{
+		ispec.MediaTypeImageManifest,
```
mtrmac: No OCI indices = no multi-arch support?!

dist/image_dest.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func (o *distImageDest) TryReusingBlob(ctx context.Context, info types.BlobInfo,
+	cache types.BlobInfoCache, canSubstitute bool) (bool, types.BlobInfo, error) {
+	if info.Digest == "" {
+		return false, types.BlobInfo{}, errors.Errorf(`"Can not check for a blob with unknown digest`)
+	}
+
+	if o.s.HasLayer(info.Digest.String()) {
+		return true, types.BlobInfo{Digest: info.Digest, Size: -1}, nil
```
mtrmac: `Size` is required to be set in this case.  (See why the function documentation from `types.go` is redundantly copy&pasted, and laboriously updated everywhere on any change, in all implementations?)

dist/image_dest.go
```diff
@@ -0,0 +1,121 @@
...
+
+	return false, types.BlobInfo{}, nil
+}
+
+func (o *distImageDest) PutManifest(ctx context.Context, m []byte, d *digest.Digest) error {
+	return o.s.PutManifest(m)
+}
+
+func (o *distImageDest) PutSignatures(ctx context.Context, signatures [][]byte, d *digest.Digest) error {
+	return nil
```
mtrmac: This should explicitly fail instead of silently ignoring data if `len(signatures) != 0`.

dist/transport.go
```diff
@@ -0,0 +1,186 @@
+package dist
```
mtrmac: Because naming has extremely important impact on security, and because it’s so easy to miss cases when parsing ambiguous strings, we typically require close to 100% unit test code coverage for the `ImageTransport` and `ImageReference` implementations.

dist/transport.go
```diff
@@ -0,0 +1,186 @@
...
+var Transport = distTransport{}
+
+type distTransport struct{}
+
+func (s distTransport) Name() string {
+	return "dist"
+}
+
+func splitReference(ref string) (fullname, server string, port int, err error) {
+	port = 8080
```
mtrmac: This value should be hard-coded in _exactly one_ place, and in a top-level constant, not both here and in `NewOciRepo`.

dist/transport.go
```diff
@@ -0,0 +1,186 @@
...
+// NOTE - the transport interface is defined in types/types.go.
+// Valid uris are:
+//    dist:///name1/name2/tag
+//    dist://server/name1/name2/name3/tag
+// The tag can be separated by either / or :
+//    dist://server:port/name1/name2/name3/tag
+//    dist://server:port/name1/name2/name3:tag
+// So the reference passed in here would be e.g.
+//    ///name1/name2/tag
+//    //server:port/name1/name2/tag
```
mtrmac: - Overwhelming experience with the Docker references is that it is way too attractive to create them using `Sprintf`, and parse them using `strings.Split`, which _almost always_ handles corner cases incorrectly. If this defines a new reference type of similar importance, it seems _very necessary_ to provide a Go package that does all the parsing, validation and formatting for users, so that users at least have a _chance_ to use something safer. I.e. I feel that something similar to `{docker/distribution,containers/image}/docker/reference` is absolutely necessary.
- The `//` part is pointless visual noise. The values are not URLs (notably no URL escaping), it only leads to extra typing and confused users.
- Why allow the `/tag` vs. `:tag` ambiguity? What’s the end-user benefit?
- Are digest references not supported? Or does “tag” somehow include digests without documentation?

dist/transport.go
```diff
@@ -0,0 +1,186 @@
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "Failed parsing reference: '%s'", reference)
+	}
+
+	// support : for tag separateion
+	var name, tag string
+
+	fields := strings.Split(fullname, ":")
+
+	if len(fields) != 2 || len(fields[0]) == 0 || len(fields[1]) == 0 {
```
mtrmac: So, `:tag` is mandatory, and the `/tag` syntax is invalid, after all?

No `:latest` default? (I could well live with this decision, but it’s an important and I imagine controversial one.)

dist/transport.go
```diff
@@ -0,0 +1,186 @@
...
+		return nil, fmt.Errorf("no tag specified in '%s'", fullname)
+	}
+
+	name = fields[0]
+	tag = fields[1]
+
+	return distReference{
```
mtrmac: At no point are contents of the fields validated in any substantial way, except for the one or two separators implicitly used when splitting the input string.

E.g. this allows the user to specify `?`, `%`, or any other URL metacharacters, and they will be silently accepted and injected into URLs. That’s not completely impossible (garbage-in/garbage-out), but it makes the system susceptible to unnecessary risks. Given any kind of untrusted input, _something_ should validate it and reject unwanted data, and it’s all too easy for every part of the chain to imagine that it’s someone else’s problem.

dist/transport.go
```diff
@@ -0,0 +1,186 @@
...
+
+	if ref.port != -1 {
+		port = fmt.Sprintf("%d/", ref.port)
+	}
+
+	return fmt.Sprintf("//%s:%s%s", ref.server, port, ref.fullname)
+}
+
+func (ref distReference) DockerReference() reference.Named {
+	return nil
```
mtrmac: Why? The references, AFAIK, use exactly the same namespace, and are pretty much 100% interoperable, and the `DockerReference` value is important for signing and for naming images in local storage (e.g. so that `podman pull` tags the result with the location of the image on the registry).

dist/transport.go
```diff
@@ -0,0 +1,186 @@
...
+
+func (ref distReference) Transport() types.ImageTransport {
+	return Transport
+}
+
+func (ref distReference) StringWithinTransport() string {
+	port := ""
+
+	if ref.port != -1 {
+		port = fmt.Sprintf("%d/", ref.port)
```
mtrmac: This does not look right; `.port == -1` will result in `"//server:fullname"`.

dist/transport.go
```diff
@@ -0,0 +1,186 @@
...
+
+	return fmt.Sprintf("//%s:%s%s", ref.server, port, ref.fullname)
+}
+
+func (ref distReference) DockerReference() reference.Named {
+	return nil
+}
+
+func (ref distReference) PolicyConfigurationIdentity() string {
+	return ref.StringWithinTransport()
```
mtrmac: `PolicyConfigurationIdentity` is supposed to be canonical, i.e. adjust for the defaulted server/port values.

go.sum
```diff
@@ -30,6 +30,7 @@ github.com/containerd/fifo v0.0.0-20190226154929-a9fb20d87448/go.mod h1:ODA38xgv
 github.com/containerd/go-runc v0.0.0-20180907222934-5a6d9f37cfa3/go.mod h1:IV7qH3hrUgRmyYrtgEeGWJfWbgcHL9CSRruz2Vqcph0=
 github.com/containerd/ttrpc v0.0.0-20190828154514-0e0f228740de/go.mod h1:PvCDdDGpgqzQIzDW1TphrGLssLDZp2GuS+X5DkEJB8o=
 github.com/containerd/typeurl v0.0.0-20180627222232-a93fcdb778cd/go.mod h1:Cm3kwCdlkCfMSHURc+r6fwoGH6/F1hH3S4sg0rLFWPc=
+github.com/containers/image v3.0.2+incompatible h1:B1lqAE8MUPCrsBLE86J0gnXleeRq8zJnQryhiiGQNyE=
```
mtrmac: Please drop this, to make sure the non-v5 imports never get in.

dist/transport.go
```diff
@@ -0,0 +1,186 @@
...
+	transports.Register(Transport)
+}
+
+// nolint (gochecknoglobals)
+var Transport = distTransport{}
+
+type distTransport struct{}
+
+func (s distTransport) Name() string {
+	return "dist"
```
mtrmac: I suppose we will need bike-shedding over this :)

mtrmac:
For reference, some previous conversation about the `c/image/docker` client vs. OCI spec compliance: https://github.com/containers/image/pull/527 .

rhatdan:
If the two distribution mechanisms get merged, it would be nice if the "docker" was changed to "dist" so we could remove a company name from a distribution type.

mtrmac:
Maybe one way to think about this: what’s the _end-user_ use case / mechanism? So far, it seems to be possible to support Docker schema1+schema2 + OCI with a single syntax, single workflow, single transport, single set of mostly consistent features, _mostly_ automatically doing the right thing (there are definitely a few cases where the code doesn’t detect limited-capability registries, even if it should be technically possible, and users have to explicitly choose a manifest format — but none of them are fundamentally impossible to fix, AFAIK).

If there are two different transports, how do users interact with this? Are they completely interchangeable? If users use CLIs that deemphasize the transport: syntax (like `podman pull`/`podman push`), or outright don’t support it, which transport is used? For direct users of the transport: syntax (e.g. users of `skopeo copy`), when should they switch to the OCI transport, and what is the benefit they get for that work?

Starting with a protocol implementation without being clear on how the software is going to be used seems backwards to me. Users, ultimately, care very little about specifications, specification names, and fairly little about conformance of any particular implementation to any particular standard — working UIs that don’t get in the way so often trump all that.

vrothberg:
I am quite excited about conforming with the OCI dist spec. Thanks a lot for tackling that!

I generally agree with @mtrmac's assessment.

Certainly, using `docker://` for OCI distribution can be confusing but that's orthogonal to this PR and can be addressed by aliasing to a more generic term (e.g., registry://) that would continue supporting the various distribution protocols.

In this PR, we should wire in the OCI support into `docker://` for all aforementioned reasons (see https://github.com/containers/image/pull/819#issuecomment-586533486). Duplicating the two _might_ work for Skopeo and users who know _exactly_ what they want, but I guess that most users either don't care or simply don't know these implementation details. I don't want user to bother about these details and let consumers of c/image continue working transparently without having to "try" different transports.

rchincha:
@mtrmac and others, thanks for the detailed comments on the PR.

As per suggestions, I am working on another PR to incorporate OCI support into docker:// transport code itself and let it evolve from there.

rchincha:
After some more investigation, it appears that perhaps no additional changes are needed in containers/image itself.

https://github.com/containers/skopeo/blob/master/docs/skopeo-copy.1.md
```
--format, -f manifest-type Manifest type (oci, v2s1, or v2s2) to use when saving image to directory using the 'dir:' transport (default is manifest type of source)
```

```
$ skopeo copy --format=oci docker://busybox:latest docker://<dist-spec-server>/busybox:latest
```

works just fine.

The support for "format" option appears to have been added a while back.
```
$ git tag --contains 485a7aa33
v0.1.33
v0.1.34
v0.1.35
v0.1.36
v0.1.37
v0.1.38
v0.1.39
v0.1.40
v0.1.41
```

tych0:
So it seems the goal should be to get thinks working in skopeo automatically, i.e. autodetecting the --format argument?

mtrmac:
c/image _is_ autodetecting[1] whether the destination supports an image format, and a conversion happens if the destination rejects it.

In base case, in which the registry supports the source image format without modification, that image is copied without conversion unless the user explicitly asks otherwise. (That way signatures, if any, and manifest digests are preserved if possible.)

What _behavior_ are you looking for? Always converting an image to OCI even if the registry supports storing the original format does not seem to be obviously and unambiguously the right thing.



[1] There are a few known deficiencies, IIRC e.g. one registry implementation that accepts OCI without error but can’t return it: that one could be detected but currently isn’t. But those are implementation deficiencies, not deliberate design decisions.

tych0:
> What behavior are you looking for? Always converting an image to OCI even if the registry supports storing the original format does not seem to be obviously and unambiguously the right thing.

For starters, if the source is OCI and the destination speaks OCI, it seems reasonable to keep it as OCI.

But I'm not sure I generally agree with this statement. Why not use the standard instead of a custom format where available?

mtrmac:
> > What behavior are you looking for? Always converting an image to OCI even if the registry supports storing the original format does not seem to be obviously and unambiguously the right thing.
> 
> For starters, if the source is OCI and the destination speaks OCI, it seems reasonable to keep it as OCI.

Isn’t that already happening? If it isn’t, that’s most likely a bug.

> But I'm not sure I generally agree with this statement. Why not use the standard instead of a custom format where available?

Because every conversion has some small risk of breaking the image itself, and a much bigger risk of changing how the image is treated by its consumers. (And there are other small reasons like schema2, but not OCI, supporting health check commands — or maybe that is just the Podman implementation, not sure.)

tych0:
> Isn’t that already happening? If it isn’t, that’s most likely a bug.

Possibly, I don't know. I do know that you can't write to zot without --format=oci, so some part of the detection is broken.

mtrmac:
(Also, the distinction between ”standard” and ”a custom format” is much less clear-cut that that wording implies, given how much software out there (still) doesn’t support OCI in full/correctly. Sure, that’s a chicken-and-egg problem, but forcing the issue by defaulting to OCI if the support is not seamless is not _that_ likely to help — it’s more likely to motivate users to sprinkle scripts with `--format=docker` (compare https://issues.sonatype.org/browse/NEXUS-16947 ), cementing the previous format even further.)

rhatdan:
@rchincha @vrothberg @mtrmac @tych0 Any update on this?

rchincha:
> 
> 
> @rchincha @vrothberg @mtrmac @tych0 Any update on this?

@rhatdan: couple of things going on:
1) in the dist-spec sphere, we are organizing conformance tests as per client workflows  (@jdolitsky and @pmengelbert) are driving this [1], [2]. The idea being not all client tools need everything and any changes needed are small and contained.
2) evaluating client tools for some tasks - e.g, skopeo (docker driver) appears to work fine for image push/pull. Others are buildah, oras, etc. Evaluating those.

References:
[1] https://github.com/opencontainers/distribution-spec/issues/108
[2] https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/_GP6IqefKcg

mtrmac:
I’m going to close this, at least for now. If at all possible the existing c/image/docker code should be able to automatically work with both OCI and old docker/distribution registries, without user involvement.

I’m fairly sure it’s not a 100% compliant client right now (IIRC in the authentication behavior, if nothing else); it may turn out that the goal of a single automagic implementation is not achievable, and in that case we may need to revisit this.

PR #826, opened 2020-02-19
Add $HOME/.config/containers/registries.conf to config path

QiWang19:
provide per-user configuration of registries.conf under $HOME/.config/containers/registries.conf for other tools.

### DIFF
--- a/docs/containers-registries.conf.5.md
+++ b/docs/containers-registries.conf.5.md
@@ -9,7 +9,7 @@ containers-registries.conf - Syntax of System Registry Configuration File
 The CONTAINERS-REGISTRIES configuration file is a system-wide configuration
 file for container image registries. The file format is TOML.
 
-By default, the configuration file is located at `/etc/containers/registries.conf`.
+Container engines will use the `$HOME/.config/containers/registries.conf` if it exists, otherwise they will use `/etc/containers/registries.conf`
 
 # FORMATS
 
--- a/pkg/sysregistriesv2/system_registries_v2.go
+++ b/pkg/sysregistriesv2/system_registries_v2.go
@@ -11,6 +11,7 @@ import (
 	"github.com/BurntSushi/toml"
 	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/image/v5/types"
+	"github.com/containers/storage/pkg/homedir"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
@@ -43,6 +44,9 @@ type Endpoint struct {
 	// connections will be allowed.
 	Insecure bool `toml:"insecure,omitempty"`
 }
+
+// userRegistriesFile is the path to the per user registry configuration file.
+var userRegistriesFile = filepath.FromSlash(".config/containers/registries.conf")
 
 // rewriteReference will substitute the provided reference `prefix` to the
 // endpoints `location` from the `ref` and creates a new named reference from it.
@@ -316,15 +320,20 @@ func (config *V2RegistriesConf) postProcess() error {
 
 // ConfigPath returns the path to the system-wide registry configuration file.
 func ConfigPath(ctx *types.SystemContext) string {
-	confPath := systemRegistriesConfPath
-	if ctx != nil {
-		if ctx.SystemRegistriesConfPath != "" {
-			confPath = ctx.SystemRegistriesConfPath
-		} else if ctx.RootForImplicitAbsolutePaths != "" {
-			confPath = filepath.Join(ctx.RootForImplicitAbsolutePaths, systemRegistriesConfPath)
-		}
+	if ctx != nil && ctx.SystemRegistriesConfPath != "" {
+		return ctx.SystemRegistriesConfPath
 	}
-	return confPath
+
+	userRegistriesFilePath := filepath.Join(homedir.Get(), userRegistriesFile)
+	if _, err := os.Stat(userRegistriesFilePath); err == nil {
+		return userRegistriesFilePath
+	}
+
+	if ctx != nil && ctx.RootForImplicitAbsolutePaths != "" {
+		return filepath.Join(ctx.RootForImplicitAbsolutePaths, systemRegistriesConfPath)
+	}
+
+	return systemRegistriesConfPath
 }
 
 // configMutex is used to synchronize concurrent accesses to configCache.
--- a/pkg/sysregistriesv2/system_registries_v2_test.go
+++ b/pkg/sysregistriesv2/system_registries_v2_test.go
@@ -118,20 +118,37 @@ func TestConfigPath(t *testing.T) {
 	const nondefaultPath = "/this/is/not/the/default/registries.conf"
 	const variableReference = "$HOME"
 	const rootPrefix = "/root/prefix"
+	oldHomeEnv, hasHomeEnv := os.LookupEnv("HOME")
+	tempHome, err := ioutil.TempDir("", "tempHome")
+	require.NoError(t, err)
+	err = os.Setenv("HOME", tempHome)
+	require.NoError(t, err)
+	defer func() {
+		os.RemoveAll(tempHome)
+		if hasHomeEnv {
+			os.Setenv("HOME", oldHomeEnv)
+		} else {
+			os.Unsetenv("HOME")
+		}
+	}()
+	var userRegistriesFile = filepath.FromSlash(".config/containers/registries.conf")
+	userRegistriesFilePath := filepath.Join(tempHome, userRegistriesFile)
 
 	for _, c := range []struct {
-		sys      *types.SystemContext
-		expected string
+		sys             *types.SystemContext
+		userfilePresent bool
+		expected        string
 	}{
 		// The common case
-		{nil, systemRegistriesConfPath},
+		{nil, false, systemRegistriesConfPath},
 		// There is a context, but it does not override the path.
-		{&types.SystemContext{}, systemRegistriesConfPath},
+		{&types.SystemContext{}, false, systemRegistriesConfPath},
 		// Path overridden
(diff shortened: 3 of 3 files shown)


QiWang19:
@rhatdan @mtrmac @vrothberg  PTAL

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -315,6 +319,9 @@ func ConfigPath(ctx *types.SystemContext) string {
 			confPath = filepath.Join(ctx.RootForImplicitAbsolutePaths, systemRegistriesConfPath)
 		}
 	}
+	if _, err := os.Stat(filepath.Join(homedir.Get(), userRegistriesFile)); err == nil {
```
mtrmac: - `SystemContext`, if it set, has precedence over everything
- Then per-user, if exists (and taking into account `RootForImplicitAbsolutePaths`)
- Then the system-wide.

This seems to prefer the per-user file to everything else.

(Non-blocking, but updating `TestConfigPath` would be nice, and would have demonstrated this.)

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -35,6 +36,9 @@ type Endpoint struct {
 	Insecure bool `toml:"insecure,omitempty"`
 }
 
+// userRegistriesFile is the path to the per user registry configuration file.
+const userRegistriesFile = ".config/containers/registries.conf"
```
mtrmac: `filepath.FromSlash`, please.

docs/containers-registries.conf.5.md
```diff
@@ -10,7 +10,7 @@ The CONTAINERS-REGISTRIES configuration file is a system-wide configuration
 file for container image registries. The file format is TOML.
 
 By default, the configuration file is located at `/etc/containers/registries.conf`.
-
+Or can be configured by creating a file at `$HOME/.config/containers/registries.conf`.
```
mtrmac: This should be clear about which one is used if both exist.

pkg/sysregistriesv2/system_registries_v2.go
TomSweeneyRedHat: I'd suggest defining a new variable, perhaps secureConfPath and set it using  securejoin.SecureJoin() before doing the os.Stat then just return it.

mtrmac: Why `SecureJoin`? We know what `userRegistriesFile` is, and that it won’t try to escape from the home directory.

mtrmac: It _would_ be useful, though, to have the `filepath.Join()` computation not repeated twice.

TomSweeneyRedHat: "Why SecureJoin?" too many burn marks.  You're right, should be good with filepath.join, but would like to see whichever join done just once.

docs/containers-registries.conf.5.md
```diff
@@ -10,6 +10,8 @@ The CONTAINERS-REGISTRIES configuration file is a system-wide configuration
 file for container image registries. The file format is TOML.
 
 By default, the configuration file is located at `/etc/containers/registries.conf`.
+If the default configuration file is not set in system context, will use user created file at `$HOME/.config/containers/registries.conf`. If user's configuration file does not exist, will use `/etc/containers/registries.conf`
+as system-wide configuration path.
```
TomSweeneyRedHat: suggest:
"If running as a rootless user, the file at `$HOME/.config/containers/registries.conf` will be used.  If that file does not exist, then the default configuration file will be used instead."

mtrmac: This code does not contain any “running as rootless user” check (and that’s good!).

QiWang19:
@mtrmac @TomSweeneyRedHat @rhatdan PTAL

docs/containers-registries.conf.5.md
```diff
@@ -10,6 +10,7 @@ The CONTAINERS-REGISTRIES configuration file is a system-wide configuration
 file for container image registries. The file format is TOML.
 
 By default, the configuration file is located at `/etc/containers/registries.conf`.
+If user created file `$HOME/.config/containers/registries.conf` exists, will use user created file. If user's configuration file does not exist, will use `/etc/containers/registries.conf` as system-wide configuration path.
```
rhatdan: How about
```
Container engines will use the `$HOME/.config/containers/registries.conf` if it exists, otherwise they will use `/etc/containers/registries.conf`

**Comment by TomSweeneyRedHat** [comment]
LGTM

File: pkg/sysregistriesv2/system_registries_v2_test.go [hunk 6]
```diff
@@ -118,21 +118,35 @@ func TestConfigPath(t *testing.T) {
...
+	assert.Nil(t, err)
+	defer func() {
+		if hasHomeEnv {
+			os.Setenv("HOME", oldHomeEnv)
+		} else {
+			os.Unsetenv("HOME")
+		}
+	}()
+	var userRegistriesFile = filepath.FromSlash(".config/containers/registries.conf")
+	userRegistriesFilePath := filepath.Join(os.TempDir(), userRegistriesFile)
```
mtrmac: Please use `ioutil.TempDir` to create a private directory for `$HOME` (and `defer os.RemoveAll()` to clean it up).

Using hard-coded paths directly in shared `TempDir` subjects users running the tests to attacks, e.g. by placing malicious symlinks.

pkg/sysregistriesv2/system_registries_v2_test.go
```diff
@@ -118,21 +118,35 @@ func TestConfigPath(t *testing.T) {
...
 
 	for _, c := range []struct {
 		sys      *types.SystemContext
 		expected string
+		userfile bool
```
mtrmac:
```suggestion
		userFilePresent bool
```

pkg/sysregistriesv2/system_registries_v2_test.go
```diff
@@ -141,10 +155,32 @@ func TestConfigPath(t *testing.T) {
...
-		{&types.SystemContext{SystemRegistriesConfPath: variableReference}, variableReference},
+		{&types.SystemContext{SystemRegistriesConfPath: variableReference}, variableReference, false},
 	} {
+		os.Remove(userRegistriesFilePath)
+		if c.userfile {
+			err := os.MkdirAll(filepath.Dir(userRegistriesFilePath), os.ModePerm)
+			assert.Nil(t, err)
+
+			_, err = os.Create(userRegistriesFilePath)
+			assert.Nil(t, err)
```
mtrmac: `require.NoError(t, err)`, in both cases, please.

pkg/sysregistriesv2/system_registries_v2_test.go
```diff
@@ -141,10 +155,32 @@ func TestConfigPath(t *testing.T) {
...
 		// No environment expansion happens in the overridden paths
-		{&types.SystemContext{SystemRegistriesConfPath: variableReference}, variableReference},
+		{&types.SystemContext{SystemRegistriesConfPath: variableReference}, variableReference, false},
 	} {
+		os.Remove(userRegistriesFilePath)
+		if c.userfile {
+			err := os.MkdirAll(filepath.Dir(userRegistriesFilePath), os.ModePerm)
+			assert.Nil(t, err)
+
+			_, err = os.Create(userRegistriesFilePath)
```
mtrmac: (Nonblocking: closing the file object explicitly would be clearer. OTOH the implementation sets a finalizer anyway, so the descriptor does not actually leak forever.)

pkg/sysregistriesv2/system_registries_v2_test.go
```diff
@@ -141,10 +155,32 @@ func TestConfigPath(t *testing.T) {
...
+				SystemRegistriesConfPath:     nondefaultPath,
+			},
+			nondefaultPath,
+			true,
 		},
 		// No environment expansion happens in the overridden paths
-		{&types.SystemContext{SystemRegistriesConfPath: variableReference}, variableReference},
+		{&types.SystemContext{SystemRegistriesConfPath: variableReference}, variableReference, false},
 	} {
+		os.Remove(userRegistriesFilePath)
```
mtrmac: (Absolutely non-blocking: Putting this into an `else` branch might be a bit clearer.)

go.mod
```diff
@@ -39,10 +39,10 @@ require (
 	github.com/xeipuuv/gojsonpointer v0.0.0-20190809123943-df4f5c81cb3b // indirect
 	github.com/xeipuuv/gojsonschema v0.0.0-20190816131739-be0936907f66
 	go.etcd.io/bbolt v1.3.3 // indirect
-	golang.org/x/crypto v0.0.0-20191112222119-e1110fd1c708
+	golang.org/x/crypto v0.0.0-20200214034016-1d94cc7ab1c6
```
mtrmac: (Is this necessary? Or did Go somehow cause the update to happen automatically?  Not blocking, just a bit surprising.)

pkg/sysregistriesv2/system_registries_v2_test.go
mtrmac: … and please put it before the `expected` member, so that inputs go before outputs.

go.mod
QiWang19: I don't know somehow it updated automatically

vrothberg:
> Note #834 (@cc: @vrothberg ); suppose this needs to account for it somehow.
> 
> The naive do-nothing approach could result with `~/.config/containers/registries.conf` config being used, but overridden by snippets from `/etc/containers/registries.conf.d`, which seems undesirable.
> 
> It seems to me that the choice for the `.conf` and `.d` default (if not individually overridden) should be done for both in sync.

Let's move that over to #834. I am merging this PR now, so I can tackle that part as well.

PR #480, opened 2018-07-13
docker client: fix requests made against bearer token services

flavio:
Previous to this commit the requests made against a bearer token service would be lacking the `author` field.That causes the jwt token to have an empty `sub` ([Subject](https://docs.docker.com/registry/spec/auth/jwt/#getting-a-bearer-token)) field.

The docker registry relies on the `sub` field to know the author of all the requests. When a registry is configured to send [notifications](https://docs.docker.com/registry/configuration/#notifications) the resulting webhooks calls will be missing the author too.

This commit fixes the issue by reproducing the same request made by the docker client.

## Some logs

In the next sections you can find some debugging logs obtained by running docker and skopeo against a docker registry configured to use [Portus](https://port.us.org) as bearer token.

### docker client

This is the authentication request sent to the bearer token service:

```
Started GET "/v2/token?account=flavio&client_id=docker&offline_token=true&service=registry.kube.lan" for 127.0.0.1 at 2018-07-13 12:40:57 +0000
Processing by Api::V2::TokensController#show as JSON
  Parameters: {"account"=>"flavio", "client_id"=>"docker", "offline_token"=>"true", "service"=>"registry.kube.lan"}
[... snip ...]
[jwt_token] [claim] {:iss=>"portus.kube.lan", :sub=>"flavio", :aud=>"registry.kube.lan", :iat=>1531485657, :nbf=>1531485652, :exp=>1531485957, :jti=>"2xNhr4Rs2EXaK1iE6BHQbssLt9ewHwtAVZ4tdzvqWo"}
Completed 200 OK in 88ms (Views: 0.2ms | ActiveRecord: 16.0ms)
```
The `GET` request has `account` set to `flavio`, the same value is present into the final bearer token `:sub=>"flavio"`.

The logs on the docker registry are something like:

```
time="2018-07-13T13:16:45.399284684Z" level=debug msg="filesystem.Stat(\"/docker/registry/v2/blobs/sha256/74/74f634b1bc1bd74535d5209589734efbd44a25f4e2dc96d78784576a3eb5b335/data\")" auth.user.name=flavio go.version=go1.9.4 http.request.c
ontenttype="application/vnd.docker.distribution.manifest.v2+json" http.request.host=registry.kube.lan http.request.id=f477cba8-5667-4ca8-880f-d1de128d90c4 http.request.method=PUT http.request.remoteaddr=192.168.100.1 http.request.uri="/v2
[...]

PR #829, opened 2020-02-21
Support IdentityToken in registry authn

### LINKED ISSUE #748: Support oauth2 authentication mechanism
This issue is a follow up to https://github.com/containers/skopeo/issues/533 and https://github.com/containers/libpod/issues/4357.

The issue first manifested when using azure container registry. To use private azure container registry, `az acr login` updates the credential required to `$HOME/.docker/container.json`. Docker successfully consumed the auth tokens, but podman fails to use the same token. `az acr login` creates a `identityToken` field and the current authentication mechanism doesn't work with this. @mtrmac identified the issue as `container/image` doesn't support OAuth2 authentication in https://github.com/containers/skopeo/issues/533. 

Given that azure container registry uses this and to be feature parity with Docker, this issue should be fixed.

yihuaf:
Adding the support for using identitytoken in the .docker/config.json
files. It's part of oauth2 and Azure Container Registry is one of the case that uses this.

Since containers/image implemented it's own docker registry client, we want to make it working with oauth2. The identitytoken can be used to get an access/bear token in place of password. Instead of setting basicAuth in the request, we follow the oauth2 definition to send a post request with grant_type.

Reference:
1. Docker registry oauth2 documentation: https://docs.docker.com/registry/spec/auth/oauth/
2. Azure Container Registry oauth2 documentation: https://github.com/Azure/acr/blob/7eddc1c5866ce5cc80e95df37d9d119e70d8c15e/docs/AAD-OAuth.md

Test: Tested with podman and logging into azure container registry, which uses identitytoken.

Concern: This patch changed the `config.GetAuthentication` interface, which is used by podman directly. Not sure if this is intended behaviour for podman to access GetAuthentication directly.

close #748  
This is the base to fix these: containers/skopeo#533 and containers/libpod#4357.

Signed-off-by: yihuaf <fang.yihua.eric@gmail.com>

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -1,11 +1,13 @@
 package docker
 
 import (
+	"bytes"
 	"context"
 	"crypto/tls"
 	"encoding/json"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"net/url"
 	"os"
@@ -97,8 +99,7 @@ type dockerClient struct {
 	// by detectProperties(). Callers can edit tlsClientConfig.InsecureSkipVerify in the meantime.
 	tlsClientConfig *tls.Config
 	// The following members are not set by newDockerClient and must be set by callers if needed.
-	username      string
-	password      string
+	auth          types.DockerAuthConfig
 	registryToken string
 	signatureBase signatureStorageBase
 	scope         authScope
@@ -210,10 +211,11 @@ func dockerCertDir(sys *types.SystemContext, hostPort string) (string, error) {
 // “write” specifies whether the client will be used for "write" access (in particular passed to lookaside.go:toplevelFromSection)
 func newDockerClientFromRef(sys *types.SystemContext, ref dockerReference, write bool, actions string) (*dockerClient, error) {
 	registry := reference.Domain(ref.ref)
-	username, password, err := config.GetAuthentication(sys, registry)
+	auth, err := config.GetCredentials(sys, registry)
 	if err != nil {
 		return nil, errors.Wrapf(err, "error getting username and password")
 	}
+
 	sigBase, err := configuredSignatureStorageBase(sys, ref, write)
 	if err != nil {
 		return nil, err
@@ -223,8 +225,7 @@ func newDockerClientFromRef(sys *types.SystemContext, ref dockerReference, write
 	if err != nil {
 		return nil, err
 	}
-	client.username = username
-	client.password = password
+	client.auth = auth
 	if sys != nil {
 		client.registryToken = sys.DockerBearerRegistryToken
 	}
@@ -289,8 +290,10 @@ func CheckAuth(ctx context.Context, sys *types.SystemContext, username, password
 	if err != nil {
 		return errors.Wrapf(err, "error creating new docker client")
 	}
-	client.username = username
-	client.password = password
+	client.auth = types.DockerAuthConfig{
+		Username: username,
+		Password: password,
+	}
 
 	resp, err := client.makeRequest(ctx, "GET", "/v2/", nil, nil, v2Auth, nil)
 	if err != nil {
@@ -332,7 +335,7 @@ func SearchRegistry(ctx context.Context, sys *types.SystemContext, registry, ima
 	v1Res := &V1Results{}
 
 	// Get credentials from authfile for the underlying hostname
-	username, password, err := config.GetAuthentication(sys, registry)
+	auth, err := config.GetCredentials(sys, registry)
 	if err != nil {
 		return nil, errors.Wrapf(err, "error getting username and password")
 	}
@@ -350,8 +353,7 @@ func SearchRegistry(ctx context.Context, sys *types.SystemContext, registry, ima
 	if err != nil {
 		return nil, errors.Wrapf(err, "error creating new docker client")
 	}
-	client.username = username
-	client.password = password
+	client.auth = auth
 	if sys != nil {
 		client.registryToken = sys.DockerBearerRegistryToken
 	}
@@ -535,7 +537,7 @@ func (c *dockerClient) setupRequestAuth(req *http.Request, extraScope *authScope
 		schemeNames = append(schemeNames, challenge.Scheme)
 		switch challenge.Scheme {
 		case "basic":
-			req.SetBasicAuth(c.username, c.password)
+			req.SetBasicAuth(c.auth.Username, c.auth.Password)
 			return nil
 		case "bearer":
 			registryToken := c.registryToken
@@ -553,10 +555,19 @@ func (c *dockerClient) setupRequestAuth(req *http.Request, extraScope *authScope
 					token = t.(bearerToken)
 				}
 				if !inCache || time.Now().After(token.expirationTime) {
-					t, err := c.getBearerToken(req.Context(), challenge, scopes)
+					var (
+						t   *bearerToken
+						err error
+					)
+					if c.auth.IdentityToken != "" {
+						t, err = c.getBearerTokenOAuth2(req.Context(), challenge, scopes)
+					} else {
+						t, err = c.getBearerToken(req.Context(), challenge, scopes)
+					}
 					if err != nil {
 						return err
 					}
+
 					token = *t
 					c.tokenCache.Store(cacheKey, token)
 				}
@@ -572,48 +583,96 @@ func (c *dockerClient) setupRequestAuth(req *http.Request, extraScope *authScope
 	return nil
 }
 
-func (c *dockerClient) getBearerToken(ctx context.Context, challenge challenge, scopes []authScope) (*bearerToken, error) {
+func (c *dockerClient) getBearerTokenOAuth2(ctx context.Context, challenge challenge,
+	scopes []authScope) (*bearerToken, error) {
 	realm, ok := challenge.Parameters["realm"]
 	if !ok {
 		return nil, errors.Errorf("missing realm in bearer auth challenge")
 	}
 
-	authReq, err := http.NewRequest("GET", realm, nil)
+	authReq, err := http.NewRequest(http.MethodPost, realm, nil)
 	if err != nil {
 		return nil, err
 	}
+
 	authReq = authReq.WithContext(ctx)
-	getParams := authReq.URL.Query()
-	if c.username != "" {
-		getParams.Add("account", c.username)
-	}
+
+	// Make the form data required against the oauth2 authentication
+	// More details here: https://docs.docker.com/registry/spec/auth/oauth/
+	params := authReq.URL.Query()
 	if service, ok := challenge.Parameters["service"]; ok && service != "" {
-		getParams.Add("service", service)
+		params.Add("service", service)
 	}
 	for _, scope := range scopes {
 		if scope.remoteName != "" && scope.actions != "" {
-			getParams.Add("scope", fmt.Sprintf("repository:%s:%s", scope.remoteName, scope.actions))
+			params.Add("scope", fmt.Sprintf("repository:%s:%s", scope.remoteName, scope.actions))
 		}
 	}
-	authReq.URL.RawQuery = getParams.Encode()
-	if c.username != "" && c.password != "" {
-		authReq.SetBasicAuth(c.username, c.password)
-	}
+	params.Add("grant_type", "refresh_token")
+	params.Add("refresh_token", c.auth.IdentityToken)
+
+	authReq.Body = ioutil.NopCloser(bytes.NewBufferString(params.Encode()))
+	authReq.Header.Add("Content-Type", "application/x-www-form-urlencoded")
 	logrus.Debugf("%s %s", authReq.Method, authReq.URL.String())
 	res, err := c.client.Do(authReq)
 	if err != nil {
 		return nil, err
 	}
 	defer res.Body.Close()
-	switch res.StatusCode {
-	case http.StatusUnauthorized:
-		err := clientLib.HandleErrorResponse(res)
-		logrus.Debugf("Server response when trying to obtain an access token: \n%q", err.Error())
-		return nil, ErrUnauthorizedForCredentials{Err: err}
-	case http.StatusOK:
-		break
-	default:
-		return nil, errors.Errorf("unexpected http code: %d (%s), URL: %s", res.StatusCode, http.StatusText(res.StatusCode), authReq.URL)
+	if err := httpResponseToError(res, "Trying to obtain access token"); err != nil {
+		return nil, err
+	}
+
+	tokenBlob, err := iolimits.ReadAtMost(res.Body, iolimits.MaxAuthTokenBodySize)
+	if err != nil {
+		return nil, err
+	}
+
+	return newBearerTokenFromJSONBlob(tokenBlob)
+}
+
+func (c *dockerClient) getBearerToken(ctx context.Context, challenge challenge,
+	scopes []authScope) (*bearerToken, error) {
+	realm, ok := challenge.Parameters["realm"]
+	if !ok {
+		return nil, errors.Errorf("missing realm in bearer auth challenge")
+	}
+
+	authReq, err := http.NewRequest(http.MethodGet, realm, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	authReq = authReq.WithContext(ctx)
+	params := authReq.URL.Query()
+	if c.auth.Username != "" {
+		params.Add("account", c.auth.Username)
+	}
+
+	if service, ok := challenge.Parameters["service"]; ok && service != "" {
+		params.Add("service", service)
+	}
+
+	for _, scope := range scopes {
+		if scope.remoteName != "" && scope.actions != "" {
+			params.Add("scope", fmt.Sprintf("repository:%s:%s", scope.remoteName, scope.actions))
+		}
+	}
+
+	authReq.URL.RawQuery = params.Encode()
+
+	if c.auth.Username != "" && c.auth.Password != "" {
+		authReq.SetBasicAuth(c.auth.Username, c.auth.Password)
+	}
+
+	logrus.Debugf("%s %s", authReq.Method, authReq.URL.String())
+	res, err := c.client.Do(authReq)
+	if err != nil {
+		return nil, err
+	}
+	defer res.Body.Close()
+	if err := httpResponseToError(res, "Requesting bear token"); err != nil {
+		return nil, err
 	}
 	tokenBlob, err := iolimits.ReadAtMost(res.Body, iolimits.MaxAuthTokenBodySize)
 	if err != nil {
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -18,7 +18,8 @@ import (
 )
 
 type dockerAuthConfig struct {
-	Auth string `json:"auth,omitempty"`
+	Auth          string `json:"auth,omitempty"`
+	IdentityToken string `json:"identitytoken,omitempty"`
 }
 
 type dockerConfigFile struct {
@@ -72,20 +73,23 @@ func SetAuthentication(sys *types.SystemContext, registry, username, password st
 	})
 }
 
-// GetAuthentication returns the registry credentials stored in
-// either auth.json file or .docker/config.json
-// If an entry is not found empty strings are returned for the username and password
-func GetAuthentication(sys *types.SystemContext, registry string) (string, string, error) {
+// GetCredentials returns the registry credentials stored in either auth.json
+// file or .docker/config.json, including support for OAuth2 and IdentityToken.
+// If an entry is not found, an empty struct is returned.
+func GetCredentials(sys *types.SystemContext, registry string) (types.DockerAuthConfig, error) {
 	if sys != nil && sys.DockerAuthConfig != nil {
 		logrus.Debug("Returning credentials from DockerAuthConfig")
-		return sys.DockerAuthConfig.Username, sys.DockerAuthConfig.Password, nil
+		return *sys.DockerAuthConfig, nil
 	}
 
 	if enableKeyring {
 		username, password, err := getAuthFromKernelKeyring(registry)
 		if err == nil {
 			logrus.Debug("returning credentials from kernel keyring")
-			return username, password, nil
+			return types.DockerAuthConfig{
+				Username: username,
+				Password: password,
+			}, nil
 		}
 	}
 
@@ -104,18 +108,39 @@ func GetAuthentication(sys *types.SystemContext, registry string) (string, strin
 		authPath{path: filepath.Join(homedir.Get(), dockerLegacyHomePath), legacyFormat: true})
 
 	for _, path := range paths {
-		username, password, err := findAuthentication(registry, path.path, path.legacyFormat)
+		authConfig, err := findAuthentication(registry, path.path, path.legacyFormat)
 		if err != nil {
 			logrus.Debugf("Credentials not found")
-			return "", "", err
+			return types.DockerAuthConfig{}, err
 		}
-		if username != "" && password != "" {
+
+		if (authConfig.Username != "" && authConfig.Password != "") || authConfig.IdentityToken != "" {
 			logrus.Debugf("Returning credentials from %s", path.path)
-			return username, password, nil
+			return authConfig, nil
 		}
 	}
+
 	logrus.Debugf("Credentials not found")
-	return "", "", nil
+	return types.DockerAuthConfig{}, nil
+}
+
+// GetAuthentication returns the registry credentials stored in
+// either auth.json file or .docker/config.json
+// If an entry is not found empty strings are returned for the username and password
+//
+// Deprecated: This API only has support for username and password. To get the
+// support for oauth2 in docker registry authentication, we added the new
+// GetCredentials API. The new API should be used and this API is kept to
+// maintain backward compatibility.
+func GetAuthentication(sys *types.SystemContext, registry string) (string, string, error) {
+	auth, err := GetCredentials(sys, registry)
+	if err != nil {
+		return "", "", err
+	}
+	if auth.IdentityToken != "" {
+		return "", "", errors.Wrap(ErrNotSupported, "non-empty identity token found and this API doesn't support it")
+	}
+	return auth.Username, auth.Password, nil
 }
 
 // RemoveAuthentication deletes the credentials stored in auth.json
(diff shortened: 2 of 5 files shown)


docker/docker_client.go
```diff
@@ -548,34 +552,36 @@ func (c *dockerClient) setupRequestAuth(req *http.Request, extraScope *authScope
...
-	authReq, err := http.NewRequest("GET", realm, nil)
+	authReq, err := http.NewRequest(http.MethodGet, realm, nil)
 	if err != nil {
 		return nil, err
 	}
+
 	authReq = authReq.WithContext(ctx)
-	getParams := authReq.URL.Query()
-	if c.username != "" {
-		getParams.Add("account", c.username)
```
mtrmac: Why is it OK to drop this code?

docker/docker_client.go
```diff
@@ -599,6 +606,75 @@ func (c *dockerClient) getBearerToken(ctx context.Context, challenge challenge,
...
+
+	tokenBlob, err := iolimits.ReadAtMost(res.Body, iolimits.MaxAuthTokenBodySize)
+	if err != nil {
+		return nil, err
+	}
+
+	return newBearerTokenFromJSONBlob(tokenBlob)
+}
+
+func (c *dockerClient) getBearerToken(ctx context.Context, challenge challenge,
```
mtrmac: `getBearerToken` has a single caller, this single `if` doesn’t quite seem breaking out into a single function — it can be inlined into `setupRequestAuth`.

pkg/docker/config/config.go
```diff
@@ -75,17 +83,23 @@ func SetAuthentication(sys *types.SystemContext, registry, username, password st
 // GetAuthentication returns the registry credentials stored in
 // either auth.json file or .docker/config.json
 // If an entry is not found empty strings are returned for the username and password
-func GetAuthentication(sys *types.SystemContext, registry string) (string, string, error) {
+func GetAuthentication(sys *types.SystemContext, registry string) (DockerAuthConfig, error) {
 	if sys != nil && sys.DockerAuthConfig != nil {
 		logrus.Debug("Returning credentials from DockerAuthConfig")
-		return sys.DockerAuthConfig.Username, sys.DockerAuthConfig.Password, nil
+		return DockerAuthConfig{
```
mtrmac: It should be possible for a caller to specify an identity token via `SystemContext`.

So, maybe add the IdentityToken to `types.DockerAuthConfig` and don’t introduce the new `config.DockerAuthConfig` type. (Or keep the current `SystemContext.DockerAuthConfig` and add a separate `SytemContext.DockerIdentityToken` value, if they are always going to be exclusive.)

pkg/docker/config/config.go
```diff
@@ -17,8 +17,16 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+// DockerAuthConfig exposes the auth information used by docker
+type DockerAuthConfig struct {
```
mtrmac: This should document what `IdentityToken` is, and whether it can be set at the same time as `Username`/`Password`.

mtrmac:
> Concern: This patch changed the `config.GetAuthentication` interface, which is used by podman directly. Not sure if this is intended behaviour for podman to access GetAuthentication directly.

It is a public API, and most of the `config` subpackage exists for `podman {login, logout}`, so ti is intended in that sense. This PR should keep that code, unmodified, working for user name+password autentication. Podman might well want to add extra support for the identity token, and in that case it can migrate to a new API.

docker/docker_client.go
yihuaf: Base on the document "account" is not one of the parameters specified. However, "offline" and "client_id" are two params that are missing from the current implementation, but seems to be OK if not set. I didn't find a reason why this was originally added, so I dropped it here.

Ref: https://docs.docker.com/registry/spec/auth/token/

pkg/docker/config/config.go
yihuaf: Will do.

docker/docker_client.go
yihuaf: Will do.

pkg/docker/config/config.go
yihuaf: Will do.

yihuaf:
> > Concern: This patch changed the `config.GetAuthentication` interface, which is used by podman directly. Not sure if this is intended behaviour for podman to access GetAuthentication directly.
> 
> It is a public API, and most of the `config` subpackage exists for `podman {login, logout}`, so ti is intended in that sense. This PR should keep that code, unmodified, working for user name+password autentication. Podman might well want to add extra support for the identity token, and in that case it can migrate to a new API.

I can keep the "GetAuthentication" interface intact, and create a new interface for the bearer token code path. "GetAuthentication" can internally call the new function and only expose username and password. Any preference on the name of the new API? "GetAuthenticationDocker" or "GetAuthenticationOauth2"?

pkg/docker/config/config.go
yihuaf: Adding IdentityToken to `types.DockerAuthConfig`. These variables should be grouped together. It's either username + password or username + identity token.

yihuaf:
@mtrmac PTAL at the latest change. Addressed your comments.

rhatdan:
@vrothberg PTAL

yihuaf:
Force push to rebase to latest master.

docker/docker_client.go
mtrmac: Git history points at https://github.com/containers/image/pull/480 .

Specs lie, please don’t remove things unrelated to the purpose of the PR.

mtrmac:
> I can keep the "GetAuthentication" interface intact, and create a new interface for the bearer token code path. "GetAuthentication" can internally call the new function and only expose username and password. Any preference on the name of the new API? "GetAuthenticationDocker" or "GetAuthenticationOauth2"?

(I know, bikeshedding…) `GetAuthenticationStruct`? It is not _only_ for OAuth2. Or, maybe, I never really liked the “authentication” noun here - `GetCredentials`?

docker/docker_client.go
```diff
@@ -544,10 +548,21 @@ func (c *dockerClient) setupRequestAuth(req *http.Request, extraScope *authScope
...
+					t   *bearerToken
+					err error
+				)
+
+				if c.auth.IdentityToken != "" {
+					t, err = c.getBearerTokenOAuth2(req.Context(), challenge, scopes)
+				} else {
+					t, err = c.getBearerToken(req.Context(), challenge, scopes)
+				}
+
```
mtrmac: (Absolutely non-blocking: Maybe drop this line to keep the error check visually close.)

pkg/docker/config/config.go
```diff
@@ -72,20 +73,25 @@ func SetAuthentication(sys *types.SystemContext, registry, username, password st
...
-// GetAuthentication returns the registry credentials stored in
-// either auth.json file or .docker/config.json
-// If an entry is not found empty strings are returned for the username and password
-func GetAuthentication(sys *types.SystemContext, registry string) (string, string, error) {
+// GetAuthenticationOAuth2 reads the auth including identity tokens. The
+// previous API GetAuthentication only has support for username and password. To
+// maintain the backward competibility, we keep the old API. To get the support
+// for oauth2 in docker registry authentication, callers should be using this
+// API.
+func GetAuthenticationOAuth2(sys *types.SystemContext, registry string) (types.DockerAuthConfig, error) {
```
mtrmac: As mentioned elsewhere, a non-OAuth2-specific name would be better.

The doc string should say what the function does or how to use it, not describe its history (history is best kept in commit messages or PRs). Callers of _this one_ don’t care what “previous API” was or why we keep it. OTOH they _do_ care about what happens when no data is found.

pkg/docker/config/config.go
```diff
@@ -104,18 +110,28 @@ func GetAuthentication(sys *types.SystemContext, registry string) (string, strin
...
+
 	logrus.Debugf("Credentials not found")
-	return "", "", nil
+	return types.DockerAuthConfig{}, nil
+}
+
+// GetAuthentication returns the registry credentials stored in
+// either auth.json file or .docker/config.json
+// If an entry is not found empty strings are returned for the username and password
+func GetAuthentication(sys *types.SystemContext, registry string) (string, string, error) {
```
mtrmac: The doc string of _this_ function would be a good place to warn against using it, per https://github.com/golang/go/wiki/Deprecated .

pkg/docker/config/config.go
```diff
@@ -104,18 +110,28 @@ func GetAuthentication(sys *types.SystemContext, registry string) (string, strin
...
+	auth, err := GetAuthenticationOAuth2(sys, registry)
+	return auth.Username, auth.Password, err
```
mtrmac: Hum, should this return an error if the registry has an identity token configured? I’m not 100% sure, but I’m leaning rather towards “yes” — that way users are immediately told that this is not going to work.

pkg/docker/config/config.go
```diff
@@ -294,20 +310,37 @@ func deleteAuthFromCredHelper(credHelper, registry string) error {
...
+		username, password, err := decodeDockerAuth(val.Auth)
+		if err != nil {
+			return types.DockerAuthConfig{}, err
+		}
+
+		return types.DockerAuthConfig{
+			Username:      username,
+			Password:      password,
+			IdentityToken: val.IdentityToken,
+		}, nil
```
mtrmac: Move the responsibility for decoding _all_ of `val` inside `decodeDockerAuth`, please, to avoid repeating this code.

(I don’t care at all whether the function is renamed to `decodeDockerAuthConfig` at the same time or left with the current name.)

pkg/docker/config/config_test.go
```diff
@@ -110,13 +110,14 @@ func TestGetAuth(t *testing.T) {
...
-			expectedPassword string
-			expectedError    error
-			sys              *types.SystemContext
+			name                  string
+			hostname              string
+			path                  string
+			expectedUsername      string
+			expectedPassword      string
+			expectedError         error
+			expectedIdentityToken string
```
mtrmac: Non-blocking: This could be an `expected types.DockerAuthConfig` now. Also, the “error” is conceptually separate from username+password+token, and should be the last of the four.

pkg/docker/config/config_test.go
```diff
@@ -273,6 +291,12 @@ func TestGetAuthFromLegacyFile(t *testing.T) {
 				t.Fatal(err)
 			}
 
+			auth, err := GetAuthenticationOAuth2(nil, tc.hostname)
+			assert.Equal(t, tc.expectedError, err)
+			assert.Equal(t, tc.expectedUsername, auth.Username)
+			assert.Equal(t, tc.expectedPassword, auth.Password)
```
mtrmac:
```suggestion
			assert.Equal(t, tc.expectedPassword, auth.Password)
			assert.Equal(t, "", auth.IdentityToken)
```

or a single-shot
```go
			assert.Equal(t, types.DockerAuthConfig{…}, auth)
```

pkg/docker/config/config_test.go
```diff
@@ -372,20 +396,21 @@ func TestGetAuthFailsOnBadInput(t *testing.T) {
...
 
 	// no config file present
-	username, password, err := GetAuthentication(nil, "index.docker.io")
+	auth, err := GetAuthenticationOAuth2(nil, "index.docker.io")
 	if err != nil {
 		t.Fatalf("got unexpected error: %#+v", err)
 	}
-	if len(username) > 0 || len(password) > 0 {
-		t.Fatalf("got unexpected not empty username/password: %q/%q", username, password)
+	if len(auth.Username) > 0 || len(auth.Password) > 0 {
```
mtrmac: …and token? Really `if auth != types.DockerAuthConfig{}` I guess.

docker/docker_client.go
```diff
@@ -285,8 +287,11 @@ func CheckAuth(ctx context.Context, sys *types.SystemContext, username, password
 	if err != nil {
 		return errors.Wrapf(err, "error creating new docker client")
 	}
-	client.username = username
-	client.password = password
+
```
mtrmac: (Absolutely non-blocking: There wasn’t a newline here because `newDockerClient` requires callers to set the credentials separately; that’s conceptually a single operation.)

pkg/docker/config/config_test.go
```diff
@@ -216,10 +223,21 @@ func TestGetAuth(t *testing.T) {
...
+				assert.Equal(t, tc.expectedIdentityToken, auth.IdentityToken)
+
+				// Test for the previous APIs.
 				username, password, err := GetAuthentication(sys, tc.hostname)
 				assert.Equal(t, tc.expectedError, err)
 				assert.Equal(t, tc.expectedUsername, username)
 				assert.Equal(t, tc.expectedPassword, password)
+
+				if err := os.RemoveAll(configPath); err != nil {
+					t.Fatal(err)
```
mtrmac: We typically use `require.NoError(t, err)` (Yes, the top of the test does the other thing currently.)

pkg/docker/config/config_test.go
```diff
@@ -372,20 +396,21 @@ func TestGetAuthFailsOnBadInput(t *testing.T) {
...
-		t.Fatalf("got unexpected not empty username/password: %q/%q", username, password)
+	if len(auth.Username) > 0 || len(auth.Password) > 0 {
+		t.Fatalf("got unexpected not empty username/password: %q/%q", auth.Username, auth.Password)
 	}
 
 	if err := ioutil.WriteFile(configPath, []byte("Json rocks! Unless it doesn't."), 0640); err != nil {
 		t.Fatalf("failed to write file %q: %v", configPath, err)
 	}
-	username, password, err = GetAuthentication(nil, "index.docker.io")
+
```
mtrmac: Please drop this line, to keep the test a single visual unit.

pkg/docker/config/config_test.go
```diff
@@ -394,22 +419,25 @@ func TestGetAuthFailsOnBadInput(t *testing.T) {
...
 	os.RemoveAll(configPath)
 	// no config file present
-	username, password, err = GetAuthentication(nil, "index.docker.io")
+	auth, err = GetAuthenticationOAuth2(nil, "index.docker.io")
 	if err != nil {
 		t.Fatalf("got unexpected error: %#+v", err)
 	}
-	if len(username) > 0 || len(password) > 0 {
-		t.Fatalf("got unexpected not empty username/password: %q/%q", username, password)
+
```
mtrmac: Please drop this line, to keep the test a single visual unit.

pkg/docker/config/config_test.go
```diff
@@ -394,22 +419,25 @@ func TestGetAuthFailsOnBadInput(t *testing.T) {
...
+	if len(auth.Username) > 0 || len(auth.Password) > 0 {
+		t.Fatalf("got unexpected not empty username/password: %q/%q", auth.Username, auth.Password)
 	}
 
 	configPath = filepath.Join(tmpDir2, ".dockercfg")
 	if err := ioutil.WriteFile(configPath, []byte("I'm certainly not a json string."), 0640); err != nil {
 		t.Fatalf("failed to write file %q: %v", configPath, err)
 	}
-	username, password, err = GetAuthentication(nil, "index.docker.io")
+
```
mtrmac: Please drop this line, to keep the test a single visual unit.

pkg/docker/config/config_test.go
```diff
@@ -394,22 +419,25 @@ func TestGetAuthFailsOnBadInput(t *testing.T) {
...
+	auth, err = GetAuthenticationOAuth2(nil, "index.docker.io")
 	if err == nil {
-		t.Fatalf("got unexpected non-error: username=%q, password=%q", username, password)
+		t.Fatalf("got unexpected non-error: username=%q, password=%q", auth.Username, auth.Password)
 	}
+
```
mtrmac: Please drop this line, to keep the test a single visual unit.

docker/docker_client.go
```diff
@@ -561,34 +576,97 @@ func (c *dockerClient) setupRequestAuth(req *http.Request, extraScope *authScope
...
-func (c *dockerClient) getBearerToken(ctx context.Context, challenge challenge, scopes []authScope) (*bearerToken, error) {
+func (c *dockerClient) getBearerTokenOAuth2(ctx context.Context, challenge challenge,
+	scopes []authScope) (*bearerToken, error) {
 	realm, ok := challenge.Parameters["realm"]
 	if !ok {
 		return nil, errors.Errorf("missing realm in bearer auth challenge")
 	}
 
-	authReq, err := http.NewRequest("GET", realm, nil)
+	// We try to authenticate with oath2
```
mtrmac: There doesn’t strictly _needs_ to be a comment saying this, but if it exists, then the best place is a doc string for the function. By the time anyone reads this far, the function name gave it away already.

docker/docker_client.go
```diff
@@ -561,34 +576,97 @@ func (c *dockerClient) setupRequestAuth(req *http.Request, extraScope *authScope
...
+
+	authReq.Body = ioutil.NopCloser(bytes.NewBufferString(params.Encode()))
+	authReq.Header.Add("Content-Type", "application/x-www-form-urlencoded")
+	logrus.Debugf("%s %s %s", authReq.Method, authReq.URL.String(), params.Encode())
+	res, err := c.client.Do(authReq)
+	if err != nil {
+		return nil, err
+	}
+
+	defer res.Body.Close()
```
mtrmac: (Absolutely non-blocking: Move this above the newline, please, it is more relevant to the `c.client.Do` contract than to the `StatusCode` handling.)

yihuaf:
Addressed the review. Created a separate review so it's easier to track the change. I can squash the commits together if needed.

yihuaf:
Force push to rebase master.

vrothberg:
I will do another review tomorrow (unless it's merged before :-)). But before merging, @yihuaf could you squash the commits into one?

yihuaf:
> I will do another review tomorrow (unless it's merged before :-)). But before merging, @yihuaf could you squash the commits into one?

I can certainly do that :)

docker/docker_client.go
```diff
@@ -561,34 +573,100 @@ func (c *dockerClient) setupRequestAuth(req *http.Request, extraScope *authScope
...
+	authReq, err := http.NewRequest(http.MethodGet, realm, nil)
 	if err != nil {
 		return nil, err
 	}
+
 	authReq = authReq.WithContext(ctx)
-	getParams := authReq.URL.Query()
-	if c.username != "" {
-		getParams.Add("account", c.username)
+	params := make(url.Values)
```
mtrmac: This seems to drop parameters from the `realm` value. Why?

docker/docker_client.go
```diff
@@ -561,34 +573,100 @@ func (c *dockerClient) setupRequestAuth(req *http.Request, extraScope *authScope
...
+			params.Add("scope", fmt.Sprintf("repository:%s:%s", scope.remoteName, scope.actions))
+		}
+	}
+
+	params.Add("grant_type", "refresh_token")
+	params.Add("refresh_token", c.auth.IdentityToken)
+
+	authReq.Body = ioutil.NopCloser(bytes.NewBufferString(params.Encode()))
+	authReq.Header.Add("Content-Type", "application/x-www-form-urlencoded")
+	logrus.Debugf("%s %s %s", authReq.Method, authReq.URL.String(), params.Encode())
```
mtrmac: I don’t think the token value should appear in logs, at least not without an explicit opt-in. (@vrothberg WDYT?)

pkg/docker/config/config.go
```diff
@@ -104,18 +108,40 @@ func GetAuthentication(sys *types.SystemContext, registry string) (string, strin
...
+// GetCredentials API. The new API should be used and this API is kept to
+// maintain backward compatibility.
+func GetAuthentication(sys *types.SystemContext, registry string) (string, string, error) {
+	auth, err := GetCredentials(sys, registry)
+	if err != nil {
+		return "", "", err
+	}
+	if auth.IdentityToken != "" {
+		logrus.Debug("non-empty identity token found and this API doesn't support it")
+		return "", "", ErrNotSupported
```
mtrmac: Non-blocking: It would be more useful to include more end-user-focused information in the error value than a short “not supported”.  Maybe this can just be a `fmt.Errorf`, or a `errors.Wrapf(ErrNotSupported, …)`.

pkg/docker/config/config_test.go
```diff
@@ -216,10 +239,26 @@ func TestGetAuth(t *testing.T) {
...
 					sys = tc.sys
 				}
-				username, password, err := GetAuthentication(sys, tc.hostname)
+				auth, err := GetCredentials(sys, tc.hostname)
 				assert.Equal(t, tc.expectedError, err)
-				assert.Equal(t, tc.expectedUsername, username)
-				assert.Equal(t, tc.expectedPassword, password)
+				assert.Equal(t, tc.expected.Username, auth.Username)
+				assert.Equal(t, tc.expected.Password, auth.Password)
+				assert.Equal(t, tc.expected.IdentityToken, auth.IdentityToken)
```
mtrmac: Non-blocking: Doesn’t `assert.Equal(t, tc.expected, auth)` work? That was the point of using `expected` as a single unit.

pkg/docker/config/config_test.go
```diff
@@ -216,10 +239,26 @@ func TestGetAuth(t *testing.T) {
...
+
+				// Test for the previous APIs.
+				username, password, err := GetAuthentication(sys, tc.hostname)
+				switch err {
+				case nil:
+					assert.Equal(t, tc.expected.Username, username)
+					assert.Equal(t, tc.expected.Password, password)
+				case ErrNotSupported:
+					assert.Equal(t, "", username)
+					assert.Equal(t, "", password)
```
mtrmac: `assert.NotEmpty(t, tc.expected.IdentityToken)` ?

pkg/docker/config/config_test.go
```diff
@@ -216,10 +239,26 @@ func TestGetAuth(t *testing.T) {
...
+				assert.Equal(t, tc.expected.Username, auth.Username)
+				assert.Equal(t, tc.expected.Password, auth.Password)
+				assert.Equal(t, tc.expected.IdentityToken, auth.IdentityToken)
+
+				// Test for the previous APIs.
+				username, password, err := GetAuthentication(sys, tc.hostname)
+				switch err {
+				case nil:
+					assert.Equal(t, tc.expected.Username, username)
+					assert.Equal(t, tc.expected.Password, password)
```
mtrmac: `assert.empty(t, tc.expected.IdentityToken)` ?

pkg/docker/config/config_test.go
```diff
@@ -249,34 +288,44 @@ func TestGetAuthFromLegacyFile(t *testing.T) {
...
 		t.Run(tc.name, func(t *testing.T) {
 			if err := ioutil.WriteFile(configPath, contents, 0640); err != nil {
 				t.Fatal(err)
 			}
 
+			auth, err := GetCredentials(nil, tc.hostname)
+			assert.Equal(t, tc.expectedError, err)
+			assert.Equal(t, tc.expected.Username, auth.Username)
+			assert.Equal(t, tc.expected.Password, auth.Password)
+			assert.Equal(t, "", auth.IdentityToken)
```
mtrmac: Non-blocking: `assert.Equal(t, tc.expected, auth)`

pkg/docker/config/config_test.go
```diff
@@ -372,20 +421,19 @@ func TestGetAuthFailsOnBadInput(t *testing.T) {
...
-	username, password, err := GetAuthentication(nil, "index.docker.io")
+	auth, err := GetCredentials(nil, "index.docker.io")
 	if err != nil {
 		t.Fatalf("got unexpected error: %#+v", err)
 	}
-	if len(username) > 0 || len(password) > 0 {
-		t.Fatalf("got unexpected not empty username/password: %q/%q", username, password)
-	}
+
+	require.Equal(t, types.DockerAuthConfig{}, auth, "got unexpected non empty auth credential: %v", auth)
```
mtrmac: - This should be `assert` (if this fails, other parts of the test function can continue).
- Please drop the blank line above
- (Absolutely non-blocking: No need to keep the “got unexpected …” format string, `.Equal` prints something by default, including the unexpected value, IIRC.)

pkg/docker/config/config_test.go
```diff
@@ -394,21 +442,21 @@ func TestGetAuthFailsOnBadInput(t *testing.T) {
...
 	os.RemoveAll(configPath)
 	// no config file present
-	username, password, err = GetAuthentication(nil, "index.docker.io")
+	auth, err = GetCredentials(nil, "index.docker.io")
 	if err != nil {
 		t.Fatalf("got unexpected error: %#+v", err)
 	}
-	if len(username) > 0 || len(password) > 0 {
-		t.Fatalf("got unexpected not empty username/password: %q/%q", username, password)
+	if len(auth.Username) > 0 || len(auth.Password) > 0 {
```
mtrmac: This should also be a `types.DockerAuthConfig{}` comparison I think. (See the other one above).

pkg/docker/config/config_test.go
mtrmac: More directly, the switch could be based on `tc.expected.IdentityToken`  / `tc.expectedError` , so that the expectations drive what is being compared, instead of return values driving what the expectations “must” have been.

pkg/docker/config/config_test.go
mtrmac: More directly, the switch could be based on `tc.expected.IdentityToken`  / `tc.expectedError` , so that the expectations drive what is being compared, instead of return values driving what the expectations “must” have been.

docker/docker_client.go
vrothberg: Good catch! I agree. Let's not put that in the logs.

docker/docker_client.go
```diff
@@ -561,34 +573,100 @@ func (c *dockerClient) setupRequestAuth(req *http.Request, extraScope *authScope
...
+	res, err := c.client.Do(authReq)
+	if err != nil {
+		return nil, err
+	}
+	defer res.Body.Close()
+
+	switch res.StatusCode {
```
vrothberg: I think we should be using `httpResponseToError(res *http.Response) error` to avoid code duplication.

pkg/docker/config/config.go
vrothberg: I agree. The useful part is in the debug message which is better placed in the error.

docker/docker_client.go
yihuaf: fixed

pkg/docker/config/config_test.go
yihuaf: > this should be assert (if this fails, other parts of the test function can continue).

The previous behavior was t.Fatal on error condition. Changing to assert per suggestion.

docker/docker_client.go
yihuaf: Good to know.

pkg/docker/config/config_test.go
```diff
@@ -216,10 +239,24 @@ func TestGetAuth(t *testing.T) {
...
+				// Test for the previous APIs.
+				username, password, err := GetAuthentication(sys, tc.hostname)
+				if tc.expected.IdentityToken != "" {
+					assert.Equal(t, "", username)
+					assert.Equal(t, "", password)
+					assert.Error(t, err)
+				} else {
+					assert.Equal(t, tc.expected.Username, username)
+					assert.Equal(t, tc.expected.Password, password)
+					assert.Empty(t, tc.expected.IdentityToken)
```
mtrmac: (This is no longer necessary, but it’s fine to leave it in.)

pkg/docker/config/config_test.go
mtrmac: I’m sorry, I somehow thought this is a new test. You’re right that this was previously fatal, and unrelated to this PR.

yihuaf:
> One last nit.

done.

PR #831, opened 2020-02-28
When semaphore cannot acquire lock in the copy image, it should print out the error at least info level, not debug

kandaaaaa:
We rarely (<1% chance) hit the following panic when use this library. I have gone through the code and think this is the place to catch exactly why this panic happens. Due to very hard to reproduce, we need to know why Acquire() gets error
```
panic: semaphore: released more than held

goroutine 403 [running]:
golang.org/x/sync/semaphore.(*Weighted).Release(0xc000582230, 0x1)
/root/go/pkg/mod/golang.org/x/sync@v0.0.0-20190911185100-cd5d95a43a6e/semaphore/semaphore.go:98 +0x1b1
github.com/containers/image/v5/copy.(*imageCopier).copyLayers.func1(0xe, 0xc00067e8c0, 0x47, 0x2fd, 0x0, 0x0, 0x0, 0x0, 0xc0005d2780, 0x31, ...)
```

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -833,21 +833,24 @@ func (ic *imageCopier) copyLayers(ctx context.Context) error {
 		}
 	}
 
-	func() { // A scope for defer
+	if err := func() error { // A scope for defer
 		progressPool, progressCleanup := ic.c.newProgressPool(ctx)
 		defer progressCleanup()
 
 		for i, srcLayer := range srcInfos {
 			err = copySemaphore.Acquire(ctx, 1)
 			if err != nil {
-				logrus.Debug("Can't acquire semaphoer", err)
+				return errors.Wrapf(err, "Can't acquire semaphore")
 			}
 			go copyLayerHelper(i, srcLayer, encLayerBitmap[i], progressPool)
 		}
 
 		// Wait for all layers to be copied
 		copyGroup.Wait()
-	}()
+		return nil
+	}(); err != nil {
+		return err
+	}
 
 	destInfos := make([]types.BlobInfo, numLayers)
 	diffIDs := make([]digest.Digest, numLayers)


kandaaaaa:
Furthermore, i think the function should just clean up whatever downloaded and do the retry to avoid this panic when acquire hit error

mtrmac:
Thanks for the report.

AFAICS the `copySemaphore.Acquire` can only fail if `ctx` is canceled; in that case the caller does not want us to retry (and anyway it’s unclear why retrying should ever work when an intra-process lock fails).

Regardless of the reason of the failure if this did fail, we clearly shouldn’t just continue; the error should just be returned to the caller of `copyLayers`; ultimately some high-level caller will report that text to the user (or not).

Also, please sign your commits per https://github.com/containers/image/blob/master/CONTRIBUTING.md#sign-your-prs .

copy/copy.go
```diff
@@ -840,7 +840,7 @@ func (ic *imageCopier) copyLayers(ctx context.Context) error {
 		for i, srcLayer := range srcInfos {
 			err = copySemaphore.Acquire(ctx, 1)
 			if err != nil {
-				logrus.Debug("Can't acquire semaphoer", err)
+				logrus.Info("Can't acquire semaphoer", err)
```
TomSweeneyRedHat: I know it's pre-existing, but while you're in here, can you please change "semaphoer" -> "semaphore"

kandaaaaa:
@TomSweeneyRedHat when looking further, I am thinking if we can just use channel to make sure the parallel rather than addtional package to handle this. I am testing the code

kandaaaaa:
@mtrmac From the binary we have, we dont see any cancel call at all. And it is very wierd that some users saying restart the node will help... That is why i really want to see what the exact error it is

mtrmac:
> @TomSweeneyRedHat when looking further, I am thinking if we can just use channel to make sure the parallel rather than addtional package to handle this.

The ability to cancel the transfer is actually valuable, we shouldn’t lose it. (Sure, that can be done with a channel as well, but that’s more code than using the semaphore for little benefit.)

---

> From the binary we have, we dont see any cancel call at all.

Nor a `context.WithTimeout`?

> That is why i really want to see what the exact error it is

What I’m suggesting is replacing the
```go
logrus.Debug("Can't acquire semaphoer", err)
```
with
```go
	err := func() error { // A scope for defer
		…
		for … {
			err = copySemaphore.Acquire(ctx, 1)
			if err != nil {
				return errors.Wrapf(err, "Can't acquire semaphore")
			}
			go …
		}
                …
                return nil
	}()
	if err != nil { return err }
```
That way, the error reason is directly returned to the caller of `copy.Image` and can be recorded by the caller.

kandaaaaa:
Thank you for the suggestion, i will make the changes

rhatdan:
@kandaaaaa Any progress?

kandaaaaa:
@mtrmac @rhatdan Updated the code and signed. Thanks

TomSweeneyRedHat:
Needs a rebase.

rhatdan:
LGTM

PR #834, opened 2020-03-02
registries.conf.d: add stances for the registries.conf

vrothberg:
When loading the registries.conf, allow for loading additional files
from `/etc/containers/registries.conf.d`.  The files are loaded in
alpha-numerical order and specified fields will overwrite the previous
config.

### DIFF
--- a/docs/containers-registries.conf.d.5.md
+++ b/docs/containers-registries.conf.d.5.md
@@ -1,0 +1,33 @@
+% CONTAINERS-REGISTRIES.CONF.D(5)
+% Valentin Rothberg
+% Mar 2020
+
+# NAME
+containers-registries.conf.d - directory for drop-in registries.conf files
+
+# DESCRIPTION
+CONTAINERS-REGISTRIES.CONF.D is a systemd-wide directory for drop-in
+configuration files in the `containers-registries.conf(5)` format.
+
+By default, the directory is located at `/etc/containers/registries.conf.d`.
+
+# CONFIGURATION PRECEDENCE
+
+Once the main configuration at `/etc/containers/registries.conf` is loaded, the
+files in `/etc/containers/registries.conf.d` are loaded in alpha-numerical order.
+Specified fields in a config will overwrite any previous setting.
+
+For instance, setting the `unqualified-search-registries` in
+`/etc/containers/registries.conf.d/myregistries.conf` will overwrite previous
+settings in `/etc/containers/registries.conf`.
+
+Note that all files must be specified in the same version of the
+`containers-registries.conf(5)` format.  The entire `[[registry]]` table will
+always be overridden if set.
+
+# SEE ALSO
+`containers-registries.conf(5)`
+
+# HISTORY
+
+Mar 2020, Originally compiled by Valentin Rothberg <rothberg@redhat.com>
--- a/pkg/sysregistriesv2/system_registries_v2.go
+++ b/pkg/sysregistriesv2/system_registries_v2.go
@@ -2,7 +2,6 @@ package sysregistriesv2
 
 import (
 	"fmt"
-	"io/ioutil"
 	"os"
 	"path/filepath"
 	"regexp"
@@ -25,6 +24,16 @@ var systemRegistriesConfPath = builtinRegistriesConfPath
 // builtinRegistriesConfPath is the path to the registry configuration file.
 // DO NOT change this, instead see systemRegistriesConfPath above.
 const builtinRegistriesConfPath = "/etc/containers/registries.conf"
+
+// systemRegistriesConfDirPath is the path to the system-wide registry
+// configuration directory and is used to add/subtract potential registries for
+// obtaining images.  You can override this at build time with
+// -ldflags '-X github.com/containers/image/sysregistries.systemRegistriesConfDirecotyPath=$your_path'
+var systemRegistriesConfDirPath = builtinRegistriesConfDirPath
+
+// builtinRegistriesConfDirPath is the path to the registry configuration directory.
+// DO NOT change this, instead see systemRegistriesConfDirectoryPath above.
+const builtinRegistriesConfDirPath = "/etc/containers/registries.conf.d"
 
 // Endpoint describes a remote location of a registry.
 type Endpoint struct {
@@ -360,16 +369,45 @@ func TryUpdatingCache(ctx *types.SystemContext) (*V2RegistriesConf, error) {
 	defer configMutex.Unlock()
 
 	// load the config
-	config, err := loadRegistryConf(configPath)
-	if err != nil {
+	config := &tomlConfig{}
+	if err := config.loadConfig(configPath); err != nil {
 		// Return an empty []Registry if we use the default config,
 		// which implies that the config path of the SystemContext
 		// isn't set.  Note: if ctx.SystemRegistriesConfPath points to
 		// the default config, we will still return an error.
 		if os.IsNotExist(err) && (ctx == nil || ctx.SystemRegistriesConfPath == "") {
-			return &V2RegistriesConf{Registries: []Registry{}}, nil
+			config = &tomlConfig{}
+			config.V2RegistriesConf = V2RegistriesConf{Registries: []Registry{}}
+		} else {
+			return nil, err
 		}
-		return nil, err
+	}
+
+	// Load the configs from the conf directory path.
+	confDir := systemRegistriesConfDirPath
+	if ctx != nil && ctx.SystemRegistriesConfDirPath != "" {
+		confDir = ctx.SystemRegistriesConfDirPath
+	}
+	err := filepath.Walk(confDir,
+		// WalkFunc to read additional
+		func(path string, info os.FileInfo, err error) error {
+			switch {
+			case info == nil:
+				// registries.conf.d doesn't exist
+				return nil
+			case info.IsDir():
+				// ignore directories
+				return nil
+			default:
+				// expect all files to be a config
+				return config.loadConfig(path)
+			}
+		},
+	)
+	if err != nil && !os.IsNotExist(err) {
+		// Ignore IsNotExist errors: most systems won't have a
+		// registries.conf.d directory.
+		return nil, errors.Wrapf(err, "error reading registries.conf.d")
 	}
 
 	v2Config := &config.V2RegistriesConf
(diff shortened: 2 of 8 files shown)


vrothberg:
@mtrmac @baude @mrunalp PTAL

docs/containers-registries.conf.d.5.md
```diff
@@ -0,0 +1,33 @@
...
+By default, the directory is located at `/etc/containers/registries.conf.d`.
+
+# CONFIGURATION PRECEDENCE
+
+Once the main configuration at `/etc/containers/registries.conf` is loaded, the
+files in `/etc/containers/registries.conf.d` are loaded in alpha-numerical order.
+Specified fields in a config will overwrite any previous setting.
+
+For instance, setting the `unqualified-search-registries` in
+`/etc/containers/registries.conf.d/example.conf` will overwrite previous
```
TomSweeneyRedHat: suggestion only, your call.  instead of `example.conf`, perhaps `myregistries.conf`?

TomSweeneyRedHat:
I'm not sure I'm seeing the need for this, but the code LGTM.

docs/containers-registries.conf.d.5.md
vrothberg: Sounds good!

vrothberg:
> I'm not sure I'm seeing the need for this, but the code LGTM.

OpenShift wants it :)

mrunalp:
@vrothberg Thanks for putting this together so fast :)

vrothberg:
Skopeo CI has been fixed; should turn green in a bit here :heavy_check_mark:

vrothberg:
@rhatdan PTAL

rhatdan:
LGTM

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -360,16 +369,45 @@ func TryUpdatingCache(ctx *types.SystemContext) (*V2RegistriesConf, error) {
...
+			return nil, err
 		}
-		return nil, err
+	}
+
+	// Load the configs from the conf directory path.
+	confDir := systemRegistriesConfDirPath
+	if ctx != nil && ctx.SystemRegistriesConfDirPath != "" {
+		confDir = ctx.SystemRegistriesConfDirPath
+	}
```
mtrmac: `ctx.RootForImplicitAbsolutePaths` should be handled here as well.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -360,16 +369,45 @@ func TryUpdatingCache(ctx *types.SystemContext) (*V2RegistriesConf, error) {
...
+	err := filepath.Walk(confDir,
+		// WalkFunc to read additional
+		func(path string, info os.FileInfo, err error) error {
+			switch {
+			case info == nil:
+				// registries.conf.d doesn't exist
+				return nil
+			case info.IsDir():
+				// ignore directories
+				return nil
```
mtrmac: Shouldn’t this be `filepath.SkipDir` for all but the parent directory?

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -360,16 +369,45 @@ func TryUpdatingCache(ctx *types.SystemContext) (*V2RegistriesConf, error) {
...
+	if ctx != nil && ctx.SystemRegistriesConfDirPath != "" {
+		confDir = ctx.SystemRegistriesConfDirPath
+	}
+	err := filepath.Walk(confDir,
+		// WalkFunc to read additional
+		func(path string, info os.FileInfo, err error) error {
+			switch {
+			case info == nil:
+				// registries.conf.d doesn't exist
+				return nil
```
mtrmac: This (or rather `err != nil`) can also happen if the directory is not readable; that might be better reported to the caller, especially if the caller has explicitly provided a path (OTOH it’s not quite critical, and probably should not fail in the defaults case even on permission denied).

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -360,16 +369,45 @@ func TryUpdatingCache(ctx *types.SystemContext) (*V2RegistriesConf, error) {
...
+			default:
+				// expect all files to be a config
+				return config.loadConfig(path)
+			}
+		},
+	)
+	if err != nil && !os.IsNotExist(err) {
+		// Ignore IsNotExist errors: most systems won't have a
+		// registries.conf.d directory.
+		return nil, errors.Wrapf(err, "error reading registries.conf.d")
```
mtrmac: AFAICS because `case info == nil: return nil`, `filepath.Walk` will actually return `err == nil` in that case, right now.

pkg/sysregistriesv2/system_registries_v2_test.go
```diff
@@ -463,3 +463,18 @@ func TestTryUpdatingCache(t *testing.T) {
...
+		SystemRegistriesConfDirPath: "testdata/registries.conf.d",
+	}
+	configCache = make(map[string]*V2RegistriesConf)
+	registries, err := TryUpdatingCache(ctx)
+	assert.Nil(t, err)
+	assert.NotNil(t, registries)
+
+	assert.Equal(t, registries.UnqualifiedSearchRegistries, []string{"example-overwrite.com"})
+	assert.Equal(t, len(registries.Registries), 1)
+	assert.Equal(t, registries.Registries[0].Location, "2.com")
```
mtrmac: Why is this correct? Wasn’t the configuration for “`base.com` should be insecure” (as well as the empty configuration for `1.com` completely lost?

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -360,16 +369,45 @@ func TryUpdatingCache(ctx *types.SystemContext) (*V2RegistriesConf, error) {
...
+		// WalkFunc to read additional
+		func(path string, info os.FileInfo, err error) error {
+			switch {
+			case info == nil:
+				// registries.conf.d doesn't exist
+				return nil
+			case info.IsDir():
+				// ignore directories
+				return nil
+			default:
```
mtrmac: Consider restricting this to only `*.conf` files, so that random editor backups and WIP temporary files are not read.

pkg/sysregistriesv2/system_registries_v2_test.go
```diff
@@ -463,3 +463,18 @@ func TestTryUpdatingCache(t *testing.T) {
...
 }
+
+func TestRegistriesConfDirectory(t *testing.T) {
+	ctx := &types.SystemContext{
+		SystemRegistriesConfPath:    "testdata/base-for-registries.d.conf",
+		SystemRegistriesConfDirPath: "testdata/registries.conf.d",
+	}
+	configCache = make(map[string]*V2RegistriesConf)
+	registries, err := TryUpdatingCache(ctx)
+	assert.Nil(t, err)
```
mtrmac: `require.NoError(t, err)`

pkg/sysregistriesv2/system_registries_v2_test.go
vrothberg: The `[[registry]]` tables are not "merged" but entirely replaced if specified. Hence, we first replace "base.com" with "1.com" which is finally replaced with "2.com".

mtrmac: Sure, I understand that’s what the implementation does. I’m suggesting that this is not what users would expect or consider reasonable.

(It does, I suppose, work for the search path overrides this exists for… but in that case it almost might be safer to just not support `registries` overrides in `*.d` instead of overriding it as a single unit.)

vrothberg: We can extend the logic to replace items in the `[[registry]]` table (based on the location). It should be fairly straight forward.

mtrmac: That would be ideal.

(I’m not quite sure how this would work for v1, but this is a new directory — so why not make it v2 only, and do the v1→v2 conversion only on the old `.conf` file?)

vrothberg: This sounds like a sane approach. I'll spin up a new PR tomorrow morning :+1: 

@mrunalp FYI

pkg/sysregistriesv2/system_registries_v2_test.go
```diff
@@ -463,3 +463,18 @@ func TestTryUpdatingCache(t *testing.T) {
 	assert.Nil(t, registries)
 	assert.Equal(t, 1, len(configCache))
 }
+
+func TestRegistriesConfDirectory(t *testing.T) {
+	ctx := &types.SystemContext{
+		SystemRegistriesConfPath:    "testdata/base-for-registries.d.conf",
+		SystemRegistriesConfDirPath: "testdata/registries.conf.d",
+	}
+	configCache = make(map[string]*V2RegistriesConf)
```
mtrmac: One more thing: `configCache` now must be keyed by _both_ the `.conf` and the `.d/*` paths, otherwise a call with different `SystemRegistriesConfDirPath` could incorrectly use previously-cached data.

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: Good catch. I plan to return the error. We can continue discussing when opening a new PR.

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: Good catch!

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: This is fixed as we now return the errors

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: Done.

pkg/sysregistriesv2/system_registries_v2_test.go
vrothberg: I'd expect users to invalidate the cache. We already have a very similar situation when updating `registries.conf`; we don't check time stamps or anything but force users to invalidate the cache (as cri-o does).

vrothberg: Otherwise, looking up registries would get _much_ more expensive, especially with the additional `registries.conf.d.` walk.

Issue #746, opened 2019-11-06
Have copyUpdatedConfigAndManifest update not rely on src manifest restrictions


lumjjb:
Currently, the way `copyUpdatedConfigAndManifest` in `copy` package updates the manifest is to apply updates to the source manifest and call PutManifest of the modified source manifest. 

The implication is that updates to the manifest that use annotations cannot be done even if the destination supports them. An example of this is in the case of encryption - schemas like docker2 cannot be encrypted even though the destination supports encryption because docker2 struct does not have annotations, which are required.

I can look into this after https://github.com/containers/image/pull/673. Feel free to assign the issue to me.

mtrmac:
“Update destination manifest” is not really possible — the destination manifest, in general, does not exist until `PutManifest`; and we should not upload any user-visible intermediate stages only to somehow read them and edit them again.

This is all an internal in-memory matter of how `image.UpdatedImage` is implemented.

lumjjb:
Ah ok - so maybe something like creating an intermediate manifest representation that we can update in flight before actually putting it to the destination.

mtrmac:
Look at how `UpdatedImage` works, this already almost the case (the _result_ of the conversion is already an in-memory manifest representation).

Basically this amounts to doing the conversion before other edits — and having it comprehensively enough thought through / tested to handle all the corner cases, `UpdatedImageNeedsLayerDiffIDs`, and so on.

lumjjb:
alright - let me take a look at that! Thanks

lumjjb:
Resolved by https://github.com/containers/image/pull/836

PR #836, opened 2020-03-03
Make UpdatedImage update after conversion of MIME type

lumjjb:
Addressing Issue https://github.com/containers/image/issues/746

### DIFF
--- a/image/docker_schema1.go
+++ b/image/docker_schema1.go
@@ -56,7 +56,7 @@ func (m *manifestSchema1) ConfigBlob(context.Context) ([]byte, error) {
 // layers in the resulting configuration isn't guaranteed to be returned to due how
 // old image manifests work (docker v2s1 especially).
 func (m *manifestSchema1) OCIConfig(ctx context.Context) (*imgspecv1.Image, error) {
-	v2s2, err := m.convertToManifestSchema2(nil, nil)
+	v2s2, err := m.convertToManifestSchema2(ctx, types.ManifestUpdateInformation{})
 	if err != nil {
 		return nil, err
 	}
@@ -107,6 +107,24 @@ func (m *manifestSchema1) UpdatedImageNeedsLayerDiffIDs(options types.ManifestUp
 // This does not change the state of the original Image object.
 func (m *manifestSchema1) UpdatedImage(ctx context.Context, options types.ManifestUpdateOptions) (types.Image, error) {
 	copy := manifestSchema1{m: manifest.Schema1Clone(m.m)}
+
+	// We have 2 MIME types for schema 1, which are basically equivalent (even the un-"Signed" MIME type will be rejected if there isn’t a signature; so,
+	// handle conversions between them by doing nothing.
+	if options.ManifestMIMEType != manifest.DockerV2Schema1MediaType && options.ManifestMIMEType != manifest.DockerV2Schema1SignedMediaType {
+		converted, err := convertManifestIfRequiredWithUpdate(ctx, options, map[string]manifestConvertFn{
+			imgspecv1.MediaTypeImageManifest:  copy.convertToManifestOCI1,
+			manifest.DockerV2Schema2MediaType: copy.convertToManifestSchema2,
+		})
+		if err != nil {
+			return nil, err
+		}
+
+		if converted != nil {
+			return converted, nil
+		}
+	}
+
+	// No conversion required, update manifest
 	if options.LayerInfos != nil {
 		if err := copy.m.UpdateLayerInfos(options.LayerInfos); err != nil {
 			return nil, err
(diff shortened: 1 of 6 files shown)


lumjjb:
Test passed, pushing formatting nits

image/docker_schema2_test.go
```diff
@@ -485,7 +485,6 @@ func TestManifestSchema2UpdatedImage(t *testing.T) {
 		assert.NoError(t, err, mime)
 	}
 	for _, mime := range []string{
-		manifest.DockerV2Schema2MediaType, // This indicates a confused caller, not a no-op
```
mtrmac: I’m not thrilled with dropping these checks — confused callers should be detected.

image/docker_schema1.go
```diff
@@ -107,46 +107,53 @@ func (m *manifestSchema1) UpdatedImageNeedsLayerDiffIDs(options types.ManifestUp
...
+
+		return memoryImageFromManifest(&copy), nil
 	case manifest.DockerV2Schema2MediaType:
 		m2, err := copy.convertToManifestSchema2(options.InformationOnly.LayerInfos, options.InformationOnly.LayerDiffIDs)
 		if err != nil {
 			return nil, err
 		}
-		return memoryImageFromManifest(m2), nil
+
+		return m2.UpdatedImage(ctx, options)
```
mtrmac: And it would be nice to have the “no conversion” edits (`LayerInfos`+`EmbeddedDockerReference`) visually/structurally separated from the “convert depending on MIME type” path. Doing that without duplicating the “no conversion needs to happen” check rather calls for a helper to do the conversion + `UpdatedImage` call (and that helper could drop the `ManifestMIMEType` edit request so that unexpected no-ops are detected, see elsewhere). 

Maybe
```go
converted, err := convertManifestIfRequired(options, map[string]func(context.Context, types.ManifestUpdateInformation)  (types.Image, error) {
    manifest.DockerV2Schema2MediaType: copy.convertToManifestSchema2,
    imgspecv1.MediaTypeImageManifest:  copy.convertToOCI, // a new helper
})
if err != nil { fail }
if converted != nil { return converted }
// ordinary edits for LayerInfos and EmbeddedDocerReference follow
```
with a
```go
func convertManifestIfRequired(options types.ManifestUpdateOptions, converters map[string]func(context.Context, types.ManifestUpdateInformation)  (types.Image, error) {
    if options.ManifestMIMEType != "" {
        if converter, ok := converters[options.ManifestMIMEType]; ok {
            tmp, err := converter(options.InformationOnly)
            if err != nil { fail }
            optionsCopy := options
            optionsCopy.ManifestMIMEType = ""
            return tmp.UpdatedImage(ctx, optionsCopy)
        }
    }
    return nil, nil
}
```

image/docker_schema2_test.go
lumjjb: I was looking at this and there's some inconsistency with docker schema v1 where it does not have this check.. Is there a particular reason for it? Or should I change it to reflect that behavior as well?

mtrmac: The trouble in there is that the manifest looks the same for both MIME types, i.e. by the time `manifestSchema1FromManifest` is called, we can’t tell whether it is a bug trying to convert to “the same” MIME type or a generic caller just trying the next MIME type in the list without understanding that the two are really not different. (In particular, see how `determineManifestConversion` works with `dockerImageDestination.SupportedManifestMIMETypes`.)

Overall, I do think this check is valuable to detect mistakes, but not _so_ valuable that it’s worth forcing every generic caller to understand this schema1 idiosyncrasy.

image/docker_schema1.go
lumjjb: Yea this makes sense. Let me make this change.

image/docker_schema2_test.go
lumjjb: Ahh!! okay. Gotcha. Thanks, that comment in the schema1 file makes a lot more sense with this context!

lumjjb:
Am updating incrementally. Will ping you when ready for review again

lumjjb:
@mtrmac ready for another look!

lumjjb:
Rebased

image/docker_schema1.go
```diff
@@ -121,36 +122,30 @@ func (m *manifestSchema1) UpdatedImage(ctx context.Context, options types.Manife
...
+		converted, err := convertManifestIfRequired(ctx, options, map[string]manifestConvertFn{
+			imgspecv1.MediaTypeImageManifest:  copy.convertToManifestOCI1,
+			manifest.DockerV2Schema2MediaType: copy.convertToManifestSchema2,
+		})
 		if err != nil {
 			return nil, err
 		}
-		return memoryImageFromManifest(m2), nil
-	case imgspecv1.MediaTypeImageManifest:
-		// We can't directly convert to OCI, but we can transitively convert via a Docker V2.2 Distribution manifest
```
mtrmac: (Absolutely non-blocking: This comment or something like that might be worth preserving.)

image/oci.go
```diff
@@ -140,29 +140,25 @@ func (m *manifestOCI1) UpdatedImage(ctx context.Context, options types.ManifestU
...
 		if err := copy.m.UpdateLayerInfos(options.LayerInfos); err != nil {
 			return nil, err
 		}
 	}
 	// Ignore options.EmbeddedDockerReference: it may be set when converting from schema1, but we really don't care.
 
-	switch options.ManifestMIMEType {
-	case "": // No conversion, OK
-	case manifest.DockerV2Schema1MediaType, manifest.DockerV2Schema1SignedMediaType:
-		// We can't directly convert to V1, but we can transitively convert via a V2 image
```
mtrmac: (Likewise might be worth preserving.)

image/oci.go
```diff
@@ -213,6 +209,18 @@ func (m *manifestOCI1) convertToManifestSchema2() (types.Image, error) {
...
 }
 
+func (m *manifestOCI1) convertToManifestSchema1(ctx context.Context, updateInfo types.ManifestUpdateInformation) (types.Image, error) {
+	m2, err := m.convertToManifestSchema2(ctx, updateInfo)
+	if err != nil {
+		return nil, err
+	}
+
+	return m2.UpdatedImage(ctx, types.ManifestUpdateOptions{
+		ManifestMIMEType: manifest.DockerV2Schema1MediaType,
```
mtrmac: (Absolutely non-blocking: `DockerV2Schema1SignedMediaType` is a tiny bit more correct, all schema1 manifests must contain a the signature. OTOH it doesn’t show up in the data anyway…)

mtrmac:
> Aren’t the non-conversion edits now applied twice on conversions (once before the `convertManifestIfRequired` call, and once inside it)?

This admittedly doesn’t matter now, strictly speaking, because `ManifestUpdateOptions` are more or less idempotent — but there are some RFEs floating around for more editing functionality (“add a layer”) that should not be applied twice, and just being clear about the order of operations should make it easier to make future changes.

lumjjb:
Thanks for the comments @mtrmac cP:{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{hanges made!

lumjjb:
{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{

lumjjb:
Rebased! All comments should be addressed.

image/manifest.go
```diff
@@ -75,3 +76,29 @@ func manifestLayerInfosToBlobInfos(layers []manifest.LayerInfo) []types.BlobInfo
 	}
 	return blobs
 }
+
+// manifestConvertFn is used to encapsulate helper manifest converstion functions
+// to perform applying of manifest update information.
+type manifestConvertFn func(context.Context, types.ManifestUpdateInformation) (types.Image, error)
+
+// convertManifestIfRequired will run conversion functions of a manifest if
+// required and re-apply the options to the converted type.
```
mtrmac:
```suggestion
// required and re-apply the options to the converted type.
// It returns (nil, nil) if no conversion was requested.
```
(Absolutely non-blocking.)

lumjjb:
Apparently github suggestions feature doesn't add signature. repushed..

lumjjb:
rebased

lumjjb:
rebased. @vrothberg any chance you've managed to take a look at this?

vrothberg:
> rebased. @vrothberg any chance you've managed to take a look at this?

Thanks for the ping! I will have a look first thing tomorrow morning.

lumjjb:
@vrothberg done! Thanks!

PR #838, opened 2020-03-04
containers-registries.conf.d: refine semantics

vrothberg:
- Only load `.conf` suffixed files
- Enforce v2 format of registries.conf
- Don't recurse into sub-directories
- Rootless support
- Several code clean ups and minor fixes


vrothberg:
@mtrmac, let's continue the conversations here.

* I did not embed the drop-in-configs into the cache key. We already force users to invalidate the cache if they want to reload the main config. Adding new drop-in-configs seems like a very similar situation to me.

* I did not implement the idea form yesterday to "merge" the `[[registry]]` table. Mainly because I quickly ran into hairy issues: What if I want to explicitly change the order? What if I want to explicitly clear the previous table? I quickly moved back to a hard replace since this allows to do whatever a user wants to do. Changing a single item in the table requires to copy the entire one from before _but_ it allows for performing all kinds of changes.

mrunalp:
@umohnani8 @vrothberg we want this pulled into crio and backported all the way back to 4.3/1.16. Thanks!

mtrmac:
> @mtrmac, let's continue the conversations here.
(from #834 , for historical record).

> * I did not embed the drop-in-configs into the cache key. We already force users to invalidate the cache if they want to reload the main config. Adding new drop-in-configs seems like a very similar situation to me.

I didn’t mean to make the directory _contents_ (nor file content nor file list) a cache key; only to make the directory _path_ a cache key.

I do agree that if anything inside the directory changes, it’s up to the caller to explicitly `TryReloading`.

But having the path included is necessary so that two in-process users, one using `SystemContext{}`, and one using `SystemContext{SystemRegistriesConfDirPath:appSpecificOverrides}` don’t share a cache. (In this scenario, neither of the users is necessarily aware of each other, and may otherwise not care at all that this package is caching data.)

---

> * I did not implement the idea form yesterday to "merge" the `[[registry]]` table. Mainly because I quickly ran into hairy issues: What if I want to explicitly change the order?

_Approximately_, this does not apply to the v2 format (where the longest prefix wins), AFAICS. Unfortunately, “approximately” because we don’t actually reject duplicates, and in particular the current conflict detection code in in `postProcess` only detects blocked/insecure differences. OTOH, ultimately, `FindRegistry` only finds the first entry, so the others can be treated as nonexistent.

I was thinking basically to parse each of the files’ registries into a `map[prefixAfterCanonicalization]Registry`, and let later files’ entries for the same key win. That would allow overriding any entry (but not fine-tuning a small part of it). Would that not work?

> What if I want to explicitly clear the previous table?

You mean remove _everything_ configured by `registries.conf` or any other drop-in? That’s not what drop-ins typically do; they tend to be one file per purpose, and thus typically the only file dealing with a particular configuration item.

---

> I quickly moved back to a hard replace since this allows to do whatever a user wants to do. Changing a single item in the table requires to copy the entire one from before _but_ it allows for performing all kinds of changes.

That only works if the user has full knowledge of the content being overridden; if that value really is known and fixed, the user could just as well overwrite `registries.conf` directly. (In particular, this feature is motivated by the ability to drop in a single change without breaking any other contents of `registries.conf` — though, admittedly, in the limited case of the search path.)

Honestly I’m torn between leaving the hard replace in and just dropping it and making this a search-path-only thing. It’s not that the hard replace could _never_ be useful, but it seems to me that it’s going to be _far more often_ an unwelcome surprise and not usable for the expected purpose — and then it seems tempting to me to not provide the feature at all, i.e. not lock us into the hard replace semantics, and allow someone who needs a more granular feature enough to write a PR to do so in the future.

vrothberg:
> I didn’t mean to make the directory contents (nor file content nor file list) a cache key; only to make the directory path a cache key.

Done. Would have been a bad bug otherwise. Thanks for catching that!

> That’s not what drop-ins typically do; they tend to be one file per purpose, and thus typically the only file dealing with a particular configuration item.

Let's follow your idea. Will update the PR in a bit.

vrothberg:
@mtrmac, ready for the next review

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -353,13 +389,17 @@ func InvalidateCache() {
 	configCache = make(map[string]*V2RegistriesConf)
 }
 
+// cacheKey returns the key for the cache which is a concatenated string of the
+// path to registry.conf and the directory. We can do that as both are always
+// non-empty avoiding key colissions.
+func cacheKey(ctx *types.SystemContext) string {
+	return ConfigPath(ctx) + configDirPath(ctx)
```
mtrmac: `map` keys can be `struct`s, so this can be precise without having to worry about collisions or ambiguities.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -336,11 +353,30 @@ func ConfigPath(ctx *types.SystemContext) string {
 	return systemRegistriesConfPath
 }
 
+// configDirPath returns the path to the system-wide registry configuration
+// directory.
+func configDirPath(ctx *types.SystemContext) string {
```
mtrmac: Is it reasonable to decide on the `.conf` and the `.d` independently? I’m leaning towards deciding on the two as a unit (i.e., assuming there are no `SystemContext` overrides, look for per-user `.d` iff we are using a per-user `.conf`).

At least system-wide overrides for a per-user config would be surprising.  OTOH per-user overrides for a system-wide config might make sense.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -517,12 +568,66 @@ func FindRegistry(ctx *types.SystemContext, ref string) (*Registry, error) {
...
+
+	// Save the registries before decoding the file where they could be lost.
+	// We merge them later again.
+	registryMap := make(map[string]Registry)
+	for i := range c.Registries {
+		registryMap[c.Registries[i].Prefix] = c.Registries[i]
+	}
+
+	// Load the tomlConfig. Note that `DecodeFile` will overwrite set fields.
+	c.Registries = nil // important to clear the memory to prevent us from overlapping fields
```
mtrmac: FIXME:

Save-clear-unmarshal-process-overwrite feels a bit heavy-handed. Why not, for each file, unmarshall-process in a new object, and then merge? That would give us complete control over the merging policy.

OTOH risk of forgetting to merge new fields?

vrothberg: > Why not, for each file, unmarshall-process in a new object, and then merge?

Because otherwise we cannot distinguish between `unqualified-search-registies = []` and them  not being set at all. `c.Registries = nil` is needed as the "old" registries and the "new" registries start at the same memory address which could lead to unintended overlaps.

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: I think it makes sense to decide independently. Maybe a user wants to follow whatever the admin (or package) decides in `/etc/containers` but also wants to dump their stuff into `~/.config/containers/registries.d`.

vrothberg:
@mtrmac can you take another look? I want to get this merged and cut a new release, so we can wire it into CRI-O.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -315,6 +318,20 @@ func (config *V2RegistriesConf) postProcess() error {
 		config.UnqualifiedSearchRegistries[i] = registry
 	}
 
+	// Registries are ordered and the first longest prefix always wins,
+	// rendering later items with the same prefix non-existent. We cannot error
+	// out anymore as this might break existing users, so let's just ignore them
+	// to guarantee that the same prefix exists only once.
+	prefixMap := make(map[string]bool)
```
mtrmac: (Absolutely non-blocking: `prefixExists`? `seenPrefix`? `knownPrefix`?  “map” is fairly redundant, it’s easy enough to see that the type is a `map` within the scope of these <10 lines.)

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -315,6 +318,20 @@ func (config *V2RegistriesConf) postProcess() error {
...
+	uniqueRegistries := []Registry{}
+	for i := range config.Registries {
+		if _, exists := prefixMap[config.Registries[i].Prefix]; exists {
+			continue
+		}
+		uniqueRegistries = append(uniqueRegistries, config.Registries[i])
```
mtrmac: `prefixMap[…]` should be set here.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -315,6 +318,20 @@ func (config *V2RegistriesConf) postProcess() error {
...
 
+	// Registries are ordered and the first longest prefix always wins,
+	// rendering later items with the same prefix non-existent. We cannot error
+	// out anymore as this might break existing users, so let's just ignore them
+	// to guarantee that the same prefix exists only once.
+	prefixMap := make(map[string]bool)
+	uniqueRegistries := []Registry{}
+	for i := range config.Registries {
+		if _, exists := prefixMap[config.Registries[i].Prefix]; exists {
+			continue
```
mtrmac: Non-blocking: should this warn? I suppose that would be a separate PR.

(Absolutely non-blocking: just using `if …; !exists` without a `continue` might be tiny bit easier to read, maybe.)

pkg/sysregistriesv2/system_registries_v2.go
mtrmac: Still outstanding.

mtrmac: (Using a `struct` would also avoid redundant calls to `ConfigPath` / `configDirPath`; the caller could call “compute paths“ and use the return value for both the cache and accessing the files.)

pkg/sysregistriesv2/system_registries_v2_test.go
```diff
@@ -44,21 +45,30 @@ func TestParseLocation(t *testing.T) {
 }
 
 func TestEmptyConfig(t *testing.T) {
-	registries, err := GetRegistries(&types.SystemContext{SystemRegistriesConfPath: "testdata/empty.conf"})
+	registries, err := GetRegistries(&types.SystemContext{
+		SystemRegistriesConfPath:    "testdata/empty.conf",
+		SystemRegistriesConfDirPath: "invalid",
```
mtrmac: (Non-blocking: `testdata/this-does-not-exist` if that’s the kind of “invalid” this is? Also in the other cases.)

pkg/sysregistriesv2/system_registries_v2_test.go
```diff
@@ -246,11 +259,16 @@ func assertRegistryLocationsEqual(t *testing.T, expected []string, regs []Regist
 	for _, r := range regs {
 		names = append(names, r.Location)
 	}
+	sort.Strings(expected)
```
mtrmac: Update (or drop) the “and their order” comment above, please.

pkg/sysregistriesv2/system_registries_v2.go
mtrmac: > Because otherwise we cannot distinguish between `unqualified-search-registies = []` and them not being set at all.

AFAICS unset is `nil`, set to empty is `[]string{}`.

OTOH this is less code, it does work fine, and anyway it’s private.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -517,12 +568,66 @@ func FindRegistry(ctx *types.SystemContext, ref string) (*Registry, error) {
...
+		return err
+	}
+
+	// Merge the freshly loaded registries.
+	for i := range c.Registries {
+		registryMap[c.Registries[i].Prefix] = c.Registries[i]
+	}
+
+	// Clear the config's registries and append the merged set. The order will
+	// now change (go map iteration) but we never committed to that.
```
mtrmac: This is technically true, but I wonder… The callers of `GetRegistries`/`TryUpdatingCache` are going to use the list as a whole in _some_ way (if they are looking for a single entry, they should use `FindRegistry`, after all).

E.g. CRI-O logs the configuration (the return value of `TryUpdatingCache`) on reload; it seems desirable for that log line not to change gratuitously.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -393,52 +475,21 @@ func TryUpdatingCache(ctx *types.SystemContext) (*V2RegistriesConf, error) {
...
-	if config.V1RegistriesConf.Nonempty() {
-		if config.V2RegistriesConf.Nonempty() {
-			return nil, &InvalidRegistries{s: "mixing sysregistry v1/v2 is not supported"}
-		}
-		v2, err := config.V1RegistriesConf.ConvertToV2()
-		if err != nil {
+	for _, c := range dinConfigs {
+		// Enforce v2 format for drop-in-configs.
+		if err := config.loadConfig(c, true); err != nil {
 			return nil, err
```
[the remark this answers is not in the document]
mtrmac: … and maybe in the `configPath` case as well.

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: That's a very thoughtful point! We now enforce an order by sorting the registries by Prefix.

pkg/sysregistriesv2/system_registries_v2_test.go
vrothberg: Order is now enforced again, so I dropped the sorting.

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: I added a TODO and changed the code as suggested.

vrothberg:
> A full review now.

Thanks! Addressed the comments and repushed.

TomSweeneyRedHat:
@vrothberg rebase is needed

vrothberg:
> @vrothberg rebase is needed

We can rebase and merge as long as there are no merge conflicts.

go.sum
```diff
@@ -30,6 +30,7 @@ github.com/containerd/fifo v0.0.0-20190226154929-a9fb20d87448/go.mod h1:ODA38xgv
 github.com/containerd/go-runc v0.0.0-20180907222934-5a6d9f37cfa3/go.mod h1:IV7qH3hrUgRmyYrtgEeGWJfWbgcHL9CSRruz2Vqcph0=
 github.com/containerd/ttrpc v0.0.0-20190828154514-0e0f228740de/go.mod h1:PvCDdDGpgqzQIzDW1TphrGLssLDZp2GuS+X5DkEJB8o=
 github.com/containerd/typeurl v0.0.0-20180627222232-a93fcdb778cd/go.mod h1:Cm3kwCdlkCfMSHURc+r6fwoGH6/F1hH3S4sg0rLFWPc=
+github.com/containers/image v3.0.2+incompatible h1:B1lqAE8MUPCrsBLE86J0gnXleeRq8zJnQryhiiGQNyE=
```
mtrmac: (This isn’t used anywhere, is it?)

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -517,12 +583,72 @@ func FindRegistry(ctx *types.SystemContext, ref string) (*Registry, error) {
...
 
-// tomlConfig loads the unmarshals the configuration at the specified path.
-// Note that the tomlConfig's fields will be overridden if they are set in
-// specified path.  This behavior is necessary for registries.conf.d stances
-// behavior.
-func (t *tomlConfig) loadConfig(path string) error {
-	logrus.Debugf("Loading registries.conf %q", path)
-	_, err := toml.DecodeFile(path, t)
-	return err
+// tomlConfig loads and unmarshals the configuration at the specified path. Note
```
mtrmac:
```suggestion
// loadConfig loads and unmarshals the configuration at the specified path. Note
```

mtrmac: (Non-blocking, this was present before.)

go.sum
vrothberg: VSCode ... thanks for catching. I run `go mod tidy` to clean up  a bit.

vrothberg:
> One last thing, please update the error message in `docker_client.go` to include the `.d` path as well.
> 
> Feel free to merge afterwards.

Done!

PR #841, opened 2020-03-07
add missing user-agent in auth request

metallkopf:
it wasn't been set

agentless


metallkopf:
i will submit this as an issue so some else can patch it

Issue #689, opened 2019-08-22
docker: Pass Bearer Token directly


knqyf263:
Currently, this library gets the Bearer Token by using username/password .
https://github.com/containers/image/blob/5144ced37a1b21b63c6ef605e56811e29a687528/docker/docker_client.go#L428

Then, it sets the token to the Authorization header.
https://github.com/containers/image/blob/5144ced37a1b21b63c6ef605e56811e29a687528/docker/docker_client.go#L435

However, I would like to pass the Bearer Token generated elsewhere directly. This is because I am developing a Harbor plugin and this plugin receives Bearer Token directly. Any chances getting support this feature? 

If so, I'm going to implement it.

Thank you for the awesome library.

rhatdan:
@mtrmac @vrothberg WDYT?

knqyf263:
@rhatdan @mtrmac @vrothberg How is it?

lizrice:
Hi @rhatdan! If we (well, I really mean @knqyf263) came up with a proposal / PR for adding this support is that something you might accept?

mtrmac:
@knqyf263 @lizrice Sorry for the delay.  Sure, if you want to add a `types.SystemContext.DockerBearerToken` or something like that for library use, that should work nicely.

(To be extra clear, this is different from the “IdentityToken” support being added in #829, isn’t it?)

rhatdan:
Even better if you drop the "Docker" adjective.

knqyf263:
Hi @rhatdan, I've opened PR to close this issue. Would you take a look? DockerBearerToken(=DockerRegistyToken) is used to authenticate to a registry. Actually, `containers/image` already uses DockerBearerToken internally in exchange for username/password. When developing a plugin for the container registry such as Harbor, we don't have username/password but DockerRegistryToken. This PR just makes it possible to pass DockerRegistryToken from outside.

To my understanding, `IdentityToken` is used to obtain DockerRegistryToken instead of username/password.

yihuaf:
That's correct. IdentityToken is used as a refresh_token to obtain a bearer token in place of password in the oauth2 workflow. @knqyf263 intention is to pass in a bear token instead of getting one from the registry using username/password or identitytoken.

PR #842, opened 2020-03-08
feat(docker): support docker registry token

knqyf263:
Close https://github.com/containers/image/issues/689

I tried not to break backward compatibility, so I didn't change `config.GetAuthentication`. If I can do it, I'd like to make `config.GetAuthentication` return `registryToken` as well, but I'll follow your thought anyway.

Also, are there any tests against private images?  I looked for the test with `sys.DockerAuthConfig` because I wanted to add tests for `sys.DockerRegistryToken` to it, but I couldn't find it. Although I've added only the failing test, I should add happy path as well. Would you tell me how to achieve it?

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -99,6 +99,7 @@ type dockerClient struct {
 	// The following members are not set by newDockerClient and must be set by callers if needed.
 	username      string
 	password      string
+	registryToken string
 	signatureBase signatureStorageBase
 	scope         authScope
 
@@ -224,6 +225,9 @@ func newDockerClientFromRef(sys *types.SystemContext, ref dockerReference, write
 	}
 	client.username = username
 	client.password = password
+	if sys != nil {
+		client.registryToken = sys.DockerBearerRegistryToken
+	}
 	client.signatureBase = sigBase
 	client.scope.actions = actions
 	client.scope.remoteName = reference.Path(ref.ref)
@@ -348,6 +352,9 @@ func SearchRegistry(ctx context.Context, sys *types.SystemContext, registry, ima
 	}
 	client.username = username
 	client.password = password
+	if sys != nil {
+		client.registryToken = sys.DockerBearerRegistryToken
+	}
 
 	// Only try the v1 search endpoint if the search query is not empty. If it is
 	// empty skip to the v2 endpoint.
@@ -531,27 +538,31 @@ func (c *dockerClient) setupRequestAuth(req *http.Request, extraScope *authScope
 			req.SetBasicAuth(c.username, c.password)
 			return nil
 		case "bearer":
-			cacheKey := ""
-			scopes := []authScope{c.scope}
-			if extraScope != nil {
-				// Using ':' as a separator here is unambiguous because getBearerToken below uses the same separator when formatting a remote request (and because repository names can't contain colons).
-				cacheKey = fmt.Sprintf("%s:%s", extraScope.remoteName, extraScope.actions)
-				scopes = append(scopes, *extraScope)
-			}
-			var token bearerToken
-			t, inCache := c.tokenCache.Load(cacheKey)
-			if inCache {
-				token = t.(bearerToken)
-			}
-			if !inCache || time.Now().After(token.expirationTime) {
-				t, err := c.getBearerToken(req.Context(), challenge, scopes)
-				if err != nil {
-					return err
+			registryToken := c.registryToken
+			if registryToken == "" {
+				cacheKey := ""
+				scopes := []authScope{c.scope}
+				if extraScope != nil {
+					// Using ':' as a separator here is unambiguous because getBearerToken below uses the same separator when formatting a remote request (and because repository names can't contain colons).
+					cacheKey = fmt.Sprintf("%s:%s", extraScope.remoteName, extraScope.actions)
+					scopes = append(scopes, *extraScope)
 				}
-				token = *t
-				c.tokenCache.Store(cacheKey, token)
+				var token bearerToken
+				t, inCache := c.tokenCache.Load(cacheKey)
+				if inCache {
+					token = t.(bearerToken)
+				}
+				if !inCache || time.Now().After(token.expirationTime) {
+					t, err := c.getBearerToken(req.Context(), challenge, scopes)
+					if err != nil {
+						return err
+					}
+					token = *t
+					c.tokenCache.Store(cacheKey, token)
+				}
+				registryToken = token.Token
 			}
-			req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", token.Token))
+			req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", registryToken))
 			return nil
 		default:
 			logrus.Debugf("no handler for %s authentication", challenge.Scheme)
--- a/types/types.go
+++ b/types/types.go
@@ -542,7 +542,10 @@ type SystemContext struct {
 	// Allow contacting docker registries over HTTP, or HTTPS with failed TLS verification. Note that this does not affect other TLS connections.
 	DockerInsecureSkipTLSVerify OptionalBool
 	// if nil, the library tries to parse ~/.docker/config.json to retrieve credentials
+	// Ignored if DockerBearerRegistryToken is non-empty.
 	DockerAuthConfig *DockerAuthConfig
+	// if not "", the library uses this registry token to authenticate to the registry
+	DockerBearerRegistryToken string
 	// if not "", an User-Agent header is added to each request when contacting a registry.
 	DockerRegistryUserAgent string
 	// if true, a V1 ping attempt isn't done to give users a better error. Default is false.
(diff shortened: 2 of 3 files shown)


docker/docker_client.go
```diff
@@ -224,6 +225,7 @@ func newDockerClientFromRef(sys *types.SystemContext, ref dockerReference, write
 	}
 	client.username = username
 	client.password = password
+	client.registryToken = sys.DockerRegistryToken
```
mtrmac: `*types.SystemContext` can always be `nil`, and must be treated the same as `&types.SystemContext{}`.

docker/docker_client.go
```diff
@@ -348,6 +350,7 @@ func SearchRegistry(ctx context.Context, sys *types.SystemContext, registry, ima
...
```
mtrmac: `*types.SystemContext` can always be `nil`, and must be treated the same as `&types.SystemContext{}`.

types/types.go
```diff
@@ -542,7 +542,10 @@ type SystemContext struct {
 	// Allow contacting docker registries over HTTP, or HTTPS with failed TLS verification. Note that this does not affect other TLS connections.
 	DockerInsecureSkipTLSVerify OptionalBool
 	// if nil, the library tries to parse ~/.docker/config.json to retrieve credentials
+	// Ignored if DockerRegistryToken is non-empty.
 	DockerAuthConfig *DockerAuthConfig
+	// if not "", the library uses this registry token to authenticate to the registry
+	DockerRegistryToken string
```
mtrmac: Maybe `DockerRegistryBearerToken`, to differentiate this from the “identity token” discussed elsewhere.

docker/docker_client.go
```diff
@@ -531,6 +534,10 @@ func (c *dockerClient) setupRequestAuth(req *http.Request, extraScope *authScope
 			req.SetBasicAuth(c.username, c.password)
 			return nil
 		case "bearer":
+			if c.registryToken != "" {
+				req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", c.registryToken))
```
mtrmac: Could this be reformatted to share the `req.Header.Set` call with the other ways to obtain a token?

docker/docker_transport_test.go
```diff
@@ -163,6 +163,20 @@ func TestReferenceNewImage(t *testing.T) {
...
+	_, err = ref.NewImage(context.Background(), &types.SystemContext{
+		RegistriesDirPath:        "/this/doesnt/exist",
+		DockerPerHostCertDirPath: "/this/doesnt/exist",
+		ArchitectureChoice:       "amd64",
+		OSChoice:                 "linux",
+		DockerRegistryToken:      "invalid",
+	})
+	// This should fail due to invalid registry token
+	require.Error(t, err)
+}
```
mtrmac: (Non-blocking: I’m not quite sure this is worth testing here.)

knqyf263:
>Or can they store the bearer token value as well?
First of all, the bearer token is the similar stuff to IdentityToken. You can see RegistryToken in AuthConfig as well as IdentityToken.
https://github.com/moby/moby/blob/8e610b2b55bfd1bfa9436ab110d311f5e8a74dcb/api/types/auth.go#L16-L21

types/types.go
knqyf263: `moby/moby` says `RegistryToken` and `IdentityToken`, so I feel `DockerRegistryToken` and `DockerIdentityToken` is natural in `containers/image`.
https://github.com/moby/moby/blob/8e610b2b55bfd1bfa9436ab110d311f5e8a74dcb/api/types/auth.go#L16-L21

But I can follow you if you still want me to change it.

knqyf263:
>One blocker: newImageSourceAttempt could use the bearer token for a different registry, which would leak the secret; the code changing endpointSys needs to clear the value as well. 

Thank you for the review. I fixed it. https://github.com/containers/image/pull/842/commits/a2496e3ddbaa4ea3d74bd6b6661af5824cd0c160

docker/docker_client.go
knqyf263: I've fixed it.
https://github.com/containers/image/pull/842/commits/b6597e00d767628a3aed219f53b7daf32d818746

docker/docker_transport_test.go
knqyf263: I removed
https://github.com/containers/image/pull/842/commits/1ffed6145a11a60c364cf85a1c234d4f8e32c292

docker/docker_client.go
knqyf263: Do you mean like https://github.com/containers/image/pull/842/commits/90aca60274d70a498c1b348c9290f2b186a2d559? Or, you meant I should just make `req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", c.registryToken))` function?

knqyf263:
>This code is primarily tested in https://github.com/containers/skopeo/blob/master/integration/copy_test.go (see e.g. TestCopySrcWithAuth). 

I think `skopeo` doesn't have to take `RegistryToken` via the CLI option. Most of the use case for this PR is that containers/image is imported as a library. In other words, I think we can't test it by integration tests of skopeo.

docker/docker_client.go
mtrmac: This one is still outstanding AFAICS.

docker/docker_client.go
mtrmac: I was thinking
```go
var bearerToken string
if c.registryToken != "" {
     bearerToken = c.registryToken
} else {
     // existing code
}
req.Header.Set(…)
```
or maybe
```go
bearerToken := c.registryToken
if bearerToken == "" {
    // existing code
}
req.Header.Set(…)
```

types/types.go
mtrmac: I’d weakly prefer the “bearer” to be in there, but it’s not blocking for me.

Do document it at least in the comment, please. “Registry token” is ambiguous without extra context, all tokens are “for a registry”.

mtrmac:
> > This code is primarily tested in https://github.com/containers/skopeo/blob/master/integration/copy_test.go (see e.g. TestCopySrcWithAuth).
> 
> I think `skopeo` doesn't have to take `RegistryToken` via the CLI option. Most of the use case for this PR is that containers/image is imported as a library. In other words, I think we can't test it by integration tests of skopeo.

Ah, you’re right, and I so far don’t think this makes much sense as an end-user option. So, that means no tests, or unit tests here.

We still don’t want to hard-code “real” tokens for public registries in tests. I suppose it would be possible to have a unit test for this case of `setupRequestAuth` (no server is needed, just test that the values are propagated correctly), if you want to do that, but that would also be artificial enough that it doesn’t make any sense to insist on such a test.

docker/docker_client.go
knqyf263: fixed b6597e0

types/types.go
knqyf263: I got it. I fixed it.
https://github.com/containers/image/pull/842/commits/ed44669ebe2fc54a45772e965ab68b788bed4935

docker/docker_client.go
knqyf263: I see. I actually tried to avoid a long nested block because that change puts the existing code into `if` statement, but I get your point. I've fixed it.
https://github.com/containers/image/pull/842/commits/ecb74f5c21e52011d7ede140ed34d1f71ee76bfe

docker/docker_client.go
knqyf263: Oh, I'm sorry. I was confused with https://github.com/containers/image/commit/b6597e00d767628a3aed219f53b7daf32d818746. I've fixed it now. https://github.com/containers/image/pull/842/commits/4e559c3809dacade587698fc446143a0d375b506

knqyf263:
>but that would also be artificial enough that it doesn’t make any sense to insist on such a test.

I agree with you, so I just removed the test.

docker/docker_client.go
```diff
@@ -531,27 +538,31 @@ func (c *dockerClient) setupRequestAuth(req *http.Request, extraScope *authScope
...
+					// Using ':' as a separator here is unambiguous because getBearerToken below uses the same separator when formatting a remote request (and because repository names can't contain colons).
+					cacheKey = fmt.Sprintf("%s:%s", extraScope.remoteName, extraScope.actions)
+					scopes = append(scopes, *extraScope)
+				}
+				var token bearerToken
+				t, inCache := c.tokenCache.Load(cacheKey)
+				if inCache {
+					token = t.(bearerToken)
+				}
+				if c.registryToken == "" && (!inCache || time.Now().After(token.expirationTime)) {
```
mtrmac: Revert the added `c.registryToken == ""` condition here, please.

knqyf263:
@vrothberg Thanks! I squashed commits.

docker/docker_client.go
knqyf263: Thanks. I've fixed it.

PR #848, opened 2020-03-12
Updated warning about short names.

rhatdan:
Patch originally from fatherlinux <smccarty@redhat.com>

### DIFF
--- a/registries.conf
+++ b/registries.conf
@@ -9,20 +9,21 @@
 # Registries to search for images that are not fully-qualified.
 # i.e. foobar.com/my_image:latest vs my_image:latest
 #
-# NOTE: RISK OF USING UNQUALIFIED IMAGE NAMES.
-# Pulling an image that is not fully qualified, i.e., one that includes the
-# image name but does not include the registry or  tag, is not recommended.
-# There is a risk that the image being pulled could be spoofed. An example
-# of this would be if a user wanted to pull an image named `foobar` from a
-# registry and expect it to come from myregistry.com.  If myregistry.com is
-# not first in the search list, an attacker could place a different `foobar`
-# image at a registry earlier in the search list.  Now you would accidentally
-# run the attackers code rather than the intended content. Registries that
-# are added to this list should be completely controlled, i.e., not allow
-# unknown/arbitrary users being able to create accounts with arbitrary names
-# to prevent an image from being spoofed, squatted or otherwise made
-# insecure.  If it is necessary to use one of these registries, it should be
-# added at the end of the list.
+# NOTE: RISK OF USING UNQUALIFIED IMAGE NAMES
+# We recommend always using fully qualified image names including the registry
+# server (full dns name), namespace, image name, and tag
+# (e.g., registry.redhat.io/ubi8/ubi:latest). When using short names, there is
+# always an inherent risk that the image being pulled could be spoofed. For
+# example, a user wants to pull an image named `foobar` from a registry and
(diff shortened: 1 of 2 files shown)


registries.conf
```diff
@@ -9,20 +9,21 @@
...
-# run the attackers code rather than the intended content. Registries that
-# are added to this list should be completely controlled, i.e., not allow
-# unknown/arbitrary users being able to create accounts with arbitrary names
-# to prevent an image from being spoofed, squatted or otherwise made
-# insecure.  If it is necessary to use one of these registries, it should be
-# added at the end of the list.
+# NOTE: RISK OF USING UNQUALIFIED IMAGE NAMES
+# We recommend always using fully qualified image names including the registry
+# server (full dns name), namespace, image name, and tag
+# (ex. registry.redhat.io/ubi8/ubu:latest). When using short names, there is
```
TomSweeneyRedHat: ubi:latest no?

TomSweeneyRedHat:
One probable small typo, otherwise LGTM

registries.conf
```diff
@@ -9,20 +9,21 @@
...
-# run the attackers code rather than the intended content. Registries that
-# are added to this list should be completely controlled, i.e., not allow
-# unknown/arbitrary users being able to create accounts with arbitrary names
-# to prevent an image from being spoofed, squatted or otherwise made
-# insecure.  If it is necessary to use one of these registries, it should be
-# added at the end of the list.
+# NOTE: RISK OF USING UNQUALIFIED IMAGE NAMES
+# We recommend always using fully qualified image names including the registry
+# server (full dns name), namespace, image name, and tag
+# (ex. registry.redhat.io/ubi8/ubi:latest). When using short names, there is
```
vrothberg:
```suggestion
# (e.g., registry.redhat.io/ubi8/ubi:latest). When using short names, there is
```

mtrmac:
Note to self: word diff.

---

NOTE: RISK OF USING UNQUALIFIED IMAGE <s>NAMES.
Pulling an image that is not</s> <b>NAMES
We recommend always using</b> fully <s>qualified, i.e., one that includes the</s> <b>qualified</b> image <s>name but does not include</s> <b>names including</b> the registry <s>or  tag, is not recommended.
There</s>
<b>server (full dns name), namespace, image name, and tag
(e.g., registry.redhat.io/ubi8/ubi:latest). When using short names, there</b> is <s>a</s>
<b>always an inherent</b> risk that the image being pulled could be spoofed. <s>An example
of this would be if</s> <b>For
example,</b> a user <s>wanted</s> <b>wants</b> to pull an image named `foobar` from a registry and <s>expect</s>
<b>expects</b> it to come from myregistry.com. If myregistry.com is not first in the
search list, an attacker could place a different `foobar` image at a registry
earlier in the search list.  <s>Now you</s> <b>The user</b> would accidentally <b>pull and</b> run the <s>attackers</s>
<b>attacker's image and</b> code rather than the intended content. <s>Registries that</s> <b>We recommend only
adding registries which</b> are <s>added to this list should be</s> completely <s>controlled, i.e., not</s> <b>trusted, i.e. registries which don't</b>
allow <b>unknown or anonymous</b>
<s>unknown/arbitrary</s> users <s>being able</s> to create accounts with arbitrary <s>names
to</s> <b>names. This
will</b> prevent an image from being spoofed, squatted or otherwise made insecure.
If it is necessary to use one of these registries, it should be added at the
end of the list.

PR #850, opened 2020-03-13
Remove manifest interface SupportsEncryption

lumjjb:
The additional interface "SupportsEncryption" was required temporarily due to https://github.com/containers/image/issues/746 

Quote from the TODO in code:

>Remove src.SupportsEncryption call and interface once copyUpdatedConfigAndManifest does not depend on source Image manifest type. Currently, the way copyUpdatedConfigAndManifest updates the manifest is to apply updates to the source manifest and call PutManifest of the modified source manifest. The implication is that schemas like docker2 cannot be encrypted even though the destination supports encryption because docker2 struct does not have annotations, which are required.

Reference to issue: https://github.com/containers/image/issues/746

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -527,15 +527,6 @@ func (c *copier) copyOneImage(ctx context.Context, policyContext *signature.Poli
 		return nil, "", "", errors.Wrapf(err, "Error initializing image from source %s", transports.ImageName(c.rawSource.Reference()))
 	}
 
-	// TODO: Remove src.SupportsEncryption call and interface once copyUpdatedConfigAndManifest does not depend on source Image manifest type
-	// Currently, the way copyUpdatedConfigAndManifest updates the manifest is to apply updates to the source manifest and call PutManifest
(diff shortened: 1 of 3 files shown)


lumjjb:
I think there's nothing that uses this API as of yet, the only call should be from the code removed in the PR. My thought was to remove it before anyone starts using it. :)

mtrmac:
I think it’s a fair _guess_ that nothing external calls it, but it has been in a public release, and I’d prefer not to get on the “this API break is not important enough to cause all users the pain of changing import parts to /v6” slippery slope.

It would be great to remove the check in `copy/copy.go`  removed now.

The `SupportsEncryption` methods should stay, with a https://github.com/golang/go/wiki/Deprecated comment (and maybe a pending PR to remove them? I guess we should have a PR label for tracking these. OTOH it might make sense to make more of the `image.genericManifest` API private, so a PR changing the current API might be wasted. I have no idea.)

lumjjb:
Sounds good - updated this PR to just perform deprecation of `SupportsEncryption` fn

PR #853, opened 2020-03-17
Fix crash on missing RootFS in docker/tarfile configs

mtrmac:
Reported by a code scanner.

### DIFF
--- a/docker/tarfile/src.go
+++ b/docker/tarfile/src.go
@@ -249,6 +249,9 @@ func (s *Source) ensureCachedDataIsPresentPrivate() error {
 	if err := json.Unmarshal(configBytes, &parsedConfig); err != nil {
 		return errors.Wrapf(err, "Error decoding tar config %s", tarManifest[0].Config)
 	}
+	if parsedConfig.RootFS == nil {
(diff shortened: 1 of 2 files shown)


docker/tarfile/src.go
```diff
@@ -249,6 +249,9 @@ func (s *Source) ensureCachedDataIsPresentPrivate() error {
 	if err := json.Unmarshal(configBytes, &parsedConfig); err != nil {
 		return errors.Wrapf(err, "Error decoding tar config %s", tarManifest[0].Config)
 	}
+	if parsedConfig.RootFS == nil {
+		return errors.Errorf("Invalid image config %s, rootFS is not set", tarManifest[0].Config)
```
vrothberg: Maybe we can move `%s` at the end of the error message? "rootFS is not set" is the important part but it might be hard to catch that as the config string might be long.

mtrmac: The value is a path within the tar archive, the JSON contents of the config. But, sure, moving the path to the end will result in an easier-to-google error.

mtrmac: > The value is a path within the tar archive, the JSON contents of the config.

**not** the JSON contents…

PR #854, opened 2020-03-17
Add the option to choose an image by the variant field in a manifest

mtrmac:
This carries over #786 , updating it on top of the later changes to warn instead of reporting an error.

Then, the error messages were updated (now that they don’t abort) to make more sense with multiple platform options. See the added test for an example.

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -8,13 +8,13 @@ import (
 	"io/ioutil"
 	"os"
 	"reflect"
-	"runtime"
 	"strings"
 	"sync"
 	"time"
 
 	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/image/v5/image"
+	"github.com/containers/image/v5/internal/pkg/platform"
 	"github.com/containers/image/v5/manifest"
 	"github.com/containers/image/v5/pkg/blobinfocache"
 	"github.com/containers/image/v5/pkg/compression"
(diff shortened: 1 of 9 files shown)


mtrmac:
@vrothberg PTAL, I think this (and #820) are the major things worth waiting for before a release.

copy/copy.go
```diff
@@ -8,13 +8,13 @@ import (
...
 	"os"
 	"reflect"
-	"runtime"
 	"strings"
 	"sync"
 	"time"
 
 	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/image/v5/image"
+	platform "github.com/containers/image/v5/internal/pkg/platform"
```
vrothberg: This seems redundant (and smells like VSCode?).

copy/copy.go
```diff
@@ -699,24 +699,61 @@ func checkImageDestinationForCurrentRuntime(ctx context.Context, sys *types.Syst
...
+	}
+	return nil
+}
 
-		wantedArch := runtime.GOARCH
-		if sys != nil && sys.ArchitectureChoice != "" {
-			wantedArch = sys.ArchitectureChoice
+// runtimeMatchWarning returns an error message warning about the mismatch
+// between an image config and set of platforms by a runtime, or "" if the two match.
+func runtimeMatchWarning(cfg *imgspecv1.Image, platforms []imgspecv1.Platform) string {
```
vrothberg: Nit: Using an error (instead of a string) as return value seems more idiomatic to me. Can we change it?

copy/copy.go
```diff
@@ -699,24 +699,61 @@ func checkImageDestinationForCurrentRuntime(ctx context.Context, sys *types.Syst
...
+		if wantedPlatform.OS != cfg.OS {
+			osOptions.append(wantedPlatform.OS)
+			matches = false
+		}
+		if wantedPlatform.Architecture != cfg.Architecture {
+			archOptions.append(wantedPlatform.Architecture)
+			matches = false
+		}
+		if matches {
+			break
```
vrothberg: We could return here.

copy/copy.go
```diff
@@ -699,24 +699,61 @@ func checkImageDestinationForCurrentRuntime(ctx context.Context, sys *types.Syst
...
+			cfg.OS, strings.Join(osOptions.list, ", "))
+	case len(osOptions.list) == 0 && len(archOptions.list) != 0:
+		return fmt.Sprintf("Image architecture mismatch: image uses %q, expecting one of %q",
+			cfg.Architecture, strings.Join(archOptions.list, ", "))
+	default:
+		options := newOrderedSet()
+		for _, o := range platforms {
+			options.append(fmt.Sprintf("%s+%s", o.OS, o.Architecture))
 		}
+		return fmt.Sprintf("Image operating system mismatch: image uses OS %q+architecture %q, expecting one of %q",
```
vrothberg: I prefer to only keep the `default` case here. I don't like the precedence of OS over the arch as both could very well mismatch.

vrothberg: It would simply this function quite a bit as well.

internal/pkg/platform/platform_matcher.go
```diff
@@ -0,0 +1,184 @@
+package platform
+
+// Largely based on
```
vrothberg: Then we should add `// Copyright 2013-2018 Docker, Inc.`

copy/copy.go
mtrmac: Sure, the default can be good enough.

In that case I’m tempted to inline the warning generator back and drop the tests. WDYT?

vrothberg: SGTM :+1:

internal/pkg/platform/platform_matcher.go
mtrmac: (Still WIP)

vrothberg:
@mtrmac, sorry for not mentioning earlier. Could you squash the commits into one?

mtrmac:
Updated and squashed now.

rhatdan:
LGTM

Issue #864, opened 2020-03-26
no image found in manifest list for architecture amd64, variant v7, OS linux


crazy-max:
Hi,

It looks like #854 adds a new field `platform.variant` that is not available for some multi-arch image. For example image manifest of `grafana/grafana:6.7.1` does not include any `variant` field for `platform`:

```
λ docker manifest inspect grafana/grafana:6.7.1
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
   "manifests": [
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 1997,
         "digest": "sha256:9d43bebe9f3a5d27647231e1fa760644fafcc09a1ad85eeae1df40c491841bd9",
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 1575,
         "digest": "sha256:e5c033be6235bb1cb8c55153223f83ae7580c7a29adea2c17d671b2f68df9b11",
         "platform": {
            "architecture": "arm",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 1575,
         "digest": "sha256:b624750e7cdd78634b58047400a9e172b9055e1f03284afc332d7402eeaab15e",
         "platform": {
            "architecture": "arm64",
            "os": "linux"
         }
      }
   ]
}
```

Any hint to handle `platform.variant` correctly?

Logs and more info in crazy-max/diun#54

Thanks!

mtrmac:
Thanks for your report.

To confirm, you are _running_ this on ARM, but the configuration is something like https://github.com/crazy-max/diun/blob/bac975009afbc4aa802913ecc0c314957fa0a1d6/pkg/registry/registry.go#L42-L49 with an `ArchitectureChoice: "amd64"`?

That looks like an oversight with no workaround, stay tuned.

mtrmac:
Can you verify that https://github.com/containers/image/pull/865 fixes this, please?

crazy-max:
Hi @mtrmac, thanks for your quick response. Will check this out.

crazy-max:
[...]

PR #865, opened 2020-03-26
Only auto-detect platform variant if we are also auto-detecting the architecture

mtrmac:
Otherwise, on ARM, users with `ArchitectureChoice: "amd64"` still end up with a `wantedVariant: "v7"`.

Fixes: #864 .

### DIFF
--- a/internal/pkg/platform/platform_matcher.go
+++ b/internal/pkg/platform/platform_matcher.go
@@ -128,14 +128,14 @@ func WantedPlatforms(ctx *types.SystemContext) ([]imgspecv1.Platform, error) {
 	if ctx != nil && ctx.ArchitectureChoice != "" {
 		wantedArch = ctx.ArchitectureChoice
 	}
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

mtrmac:
(Hoping the original reporter will verify the fix before merging it.)

crazy-max:
@mtrmac LGTM too!

crazy-max:
@mtrmac I'm wondering if [ImageInspectInfo](https://github.com/containers/image/blob/master/types/types.go#L441-L450) should not have a `Variant` field :thinking:

mtrmac:
> @mtrmac I'm wondering if [ImageInspectInfo](https://github.com/containers/image/blob/master/types/types.go#L441-L450) should not have a `Variant` field 🤔

- Separate RFEs as separate issues, please
- It wouldn’t be universal, the OCI config does not contain a variant field. But it does seem to be possible for Docker schema2 at least.

PR #867, opened 2020-03-26
Make restTLSClientConfig in restConfig a public field

mtrmac:
imdario/mergo, despite its official documentation, does not merge public fields inside private fields as of v.0.3.9: https://github.com/imdario/mergo/issues/139

Fixing that seems non-trivial.

Instead, make the `restTLSClientConfig` a public field of `restConfig`.

The `restConfig` type itself remains private, so this does not make anything actually public outside the subpackage. This way, the calls to mergo work as expected with both 0.3.8 and 0.3.9.

### DIFF
--- a/openshift/openshift-copies.go
+++ b/openshift/openshift-copies.go
@@ -65,7 +65,7 @@ type restConfig struct {
 	BearerToken string
 
 	// TLSClientConfig contains settings to enable transport layer security
-	restTLSClientConfig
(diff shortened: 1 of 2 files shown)


mtrmac:
Yes, the commit description points at https://github.com/imdario/mergo/issues/139 . So far, it’s not clear whether this is a regression or a bug fix.

PR #874, opened 2020-03-30
Update dependencies

mtrmac:
Manually update dependencies which are not handled by Dependabot. (Some of them are now Go modules and will be handled by it.

This should eventually allow us to https://github.com/containers/skopeo/issues/796 .

DO NOT MERGE until test PRs pass in consumers of c/image to verify the upgrades don’t break anything in them.


mtrmac:
The one truly notable update is a major version update of vbauerster/mpb from v4 to v5. I haven’t yet even tried running the result to verify that progress bars still look OK.

mtrmac:
Buildah tests passed in https://github.com/containers/buildah/pull/2261 .

mtrmac:
Libpod tests passed in https://github.com/containers/libpod/pull/5682 , apart from  rdoproject.org/github-check .

mtrmac:
Skopeo tests passed in https://github.com/containers/skopeo/pull/874 (although an update to drop use of c/common was necessary).

The progress bars look plausible.

rhatdan:
Should we cut a minor release?

mtrmac:
There was one on Saturday. I wouldn’t _mind_ but maybe we can wait for more than one 5-line feature before a new release, to balance the benefit with the overhead.

mtrmac:
Actually, let me take it back. https://github.com/containers/image/pull/878 requires a synchronized update of c/image and skopeo (`{common,storage}/pkg/unshare` can’t be both in the same process), so let’s do that now (along with an update of c/storage to 1.18.1 for https://github.com/containers/storage/pull/580 .)

Issue #876, opened 2020-03-31
expose getCPUVariant/WantedPlatforms


grooverdan:
like #868 (which will probably help later - Thanks!),  I would like to see [getCPUVariant](https://github.com/containers/image/blob/master/internal/pkg/platform/platform_matcher.go#L108) exported to fill [this TODO in a PR for buildah](https://github.com/grooverdan/buildah/blob/args-target-platform/config.go#L98)

This code corresponds to the the `build --bud --platform Windows/arm64` to fill in the architecture variant correctly (after #875).

Alternately I could also take `WantedPlatforms`, with slightly more setup cost, (though passing `wantedOS` to `getCPUVariant(wantedOS, runtime.GOARCH)` seem more consistent).

I could be a bit muddled or trying to be to accepting of incomplete arguments than necessary so corrections/opinions gratefully accepted. As you've already decided to not get the Variant when provided an architecture what was the reasoning?

mtrmac:
> I could be a bit muddled or trying to be to accepting of incomplete arguments than necessary so corrections/opinions gratefully accepted. As you've already decided to not get the Variant when provided an architecture what was the reasoning?

See https://github.com/containers/image/blob/06b89332d28f4b626c308df465512c827b15c8f4/internal/pkg/platform/platform_matcher.go#L132-L137 ; does that make sense?

(To an extent it’s an artifact of the API evolution, pre-existing callers that set `ArchitectureChoice` can’t opt out of variant autodetection which didn’t exist.)

---

>  I would like to see [getCPUVariant](https://github.com/containers/image/blob/master/internal/pkg/platform/platform_matcher.go#L108) exported to fill [this TODO in a PR for buildah](https://github.com/grooverdan/buildah/blob/args-target-platform/config.go#L98)
> 
> This code corresponds to the the `build --bud --platform Windows/arm64` to fill in the architecture variant correctly (after #875).

OK, that’s a clear enough use case.

I’m not quite sure whether to expose `getVariant` (as `GetRuntimeVariant`?), all of `WantedPlatforms` (maybe renamed?), or just the front half of `WantedPlatforms` that returns the current platform without computing the compatibility list.

Also, the public functions should probably have an error return value instead of silently returning `""` if `getCPUInfo` fails.

grooverdan:
[...]

PR #882, opened 2020-04-01
WantedPlatforms: OSChoice used to determine variant

grooverdan:
If a OSChoice is specified, but the architecture/variant
is left to be determined, we determine the variant taking
into account the OSChoice, and the detection of architecture.

mentioned in #876.

### DIFF
--- a/internal/pkg/platform/platform_matcher.go
+++ b/internal/pkg/platform/platform_matcher.go
@@ -106,7 +106,7 @@ func getCPUVariantArm() string {
 }
 
 func getCPUVariant(os string, arch string) string {
-	if os == "windows" {
+	if os == "windows" || runtime.GOOS == "windows" {
 		return getCPUVariantWindows(arch)
 	}
 	if arch == "arm" || arch == "arm64" {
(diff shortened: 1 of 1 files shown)


mtrmac:
Yeah, I don’t quite see the logic in this.

If the idea is “given user-unspecified arch/variant, use the current runtime environment”, the goal is just to _detect_ “the current runtime environment”, using whatever method currently works, isn’t it? The `OSChoice` value doesn’t really factor into “whatever method currently works”

mtrmac:
(I’m sorry, the closure was a mis-click)

… doesn’t really factor into “whatever method currently works” AFAICS.

E.g. consider what happens in a few years when ARM `v9` or something similar is introduced: The current Linux logic can probably be similarly extended to learn about more values in `/proc/cpuinfo`; the Windows logic will very likely have to be significantly extended from the current trivial runtime-environment-independent implementation into something that actually asks Windows/the CPU for the right value (unless, I guess, decides that `v9` will never be supported or the new `v9` features will never be enabled — which seems unlikely).

At that point, “given user-unspecified arch/variant, use the current runtime environment” running on a v9 CPU should detect v9, and that requires using the Linux/Windows—specific logic depending on `runtime.GOOS`, not on `OSChoice`.

Am I thinking about this right?

PR #883, opened 2020-04-01
copy: avoid unnecessary re-encoding of source lists

nalind:
When copying a manifest list, if we don't change any of its contents and don't change its type, use the original serialized version instead of one that we re-encoded ourselves.


TomSweeneyRedHat:
LGTM

rhatdan:
Makes sense to me.
@mtrmac PTAL

rhatdan:
LGTM
Seems to have hung the tests. Merging.

PR #885, opened 2020-04-03
Downgrade siruspen/logrus from 1.5 to 1.4.2

umohnani8:
The upgrade of logrus forces to update the dependency in go module
consumers as well.

The new locking in logrus 1.5.0 introduces regressions in conjunction to
deadlocks when using custom hooks, which is the case in CRI-O for
example.
Having it point to v1.5.0 here is forcing cri-o to upgrade it as well
when we update c/image to 5.3.1.

Let's downgrade logrus till sirupsen/logrus#1122
is fixed.

And bump up the version to 5.4.1


umohnani8:
@rhatdan @mtrmac PTAL

mrunalp:
LGTM

PR #891, opened 2020-04-07
set client_id

sjhx:
could be elaborated to include release however this fixes a problem with missing the *required* `client_id` form field  

per https://tools.ietf.org/html/rfc6749#section-4.1.3

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -610,6 +610,7 @@ func (c *dockerClient) getBearerTokenOAuth2(ctx context.Context, challenge chall
 	}
 	params.Add("grant_type", "refresh_token")
 	params.Add("refresh_token", c.auth.IdentityToken)
+	params.Add("client_id", "containers/image")
 
 	authReq.Body = ioutil.NopCloser(bytes.NewBufferString(params.Encode()))
 	authReq.Header.Add("Content-Type", "application/x-www-form-urlencoded")


TomSweeneyRedHat:
@sjhx need a rebase for this one.

vrothberg:
Rebased via GitHub UI.

TomSweeneyRedHat:
Looking at the https://github.com/containers/skopeo/issues/533, I think we need to get this in, spin up c/image 5.4.3, then do the vendoring of the dependent projects.

mtrmac:
> Looking at the [containers/skopeo#533](https://github.com/containers/skopeo/issues/533), I think we need to get this in, spin up c/image 5.4.3, then do the vendoring of the dependent projects.

Or just leave this for c/image 5.5.0-dev and skopeo post-0.2.0. It’s a new supported server implementation for a fairly new feature. Your call.

TomSweeneyRedHat:
@mtrmac my concern is it's new now, but two years from now on RHEL will it be a problem if it's not supported in this version.  A balancing act between dragging today's release versus backporting later.  I think a slight delay today is better, but if we can't get this merged it might be a moot point.

rhatdan:
We will be releasing a new version of containers/image for RHEL8.3,
We are freezing right now for RHEL8.2.1

We should not race any new features in now.

TomSweeneyRedHat:
@rhatdan your call, I'm just concerned that if we can't authenticate to an IBM cloud container registry, we'll be spinning up 20+ BZ's that will consume a lot more time than getting this in today.

mtrmac:
I thought we _can_ authenticate using the traditional (user+password→non-OAuth2 token) flow. @sjhx ?

---

> @rhatdan your call, I'm just concerned that if we can't authenticate to an IBM cloud container registry, we'll be spinning up 20+ BZ's that will consume a lot more time than getting this in today.

I don’t think we have received them so far — and at least I personally haven’t verified that this PR doesn’t break Azure, for which the OAuth2 feature was necessary (OTOH, I have no reason to suspect this _should_ break Azure, and this is _really_ small).

I agree it’s a tough call either way.

rhatdan:
I will let @sjhx chime in.  Whether this is critical now, or can wait until RHEL8.3 in the Fall.

yihuaf:
> I thought we can authenticate using the traditional (user+password→non-OAuth2 token) flow. @sjhx ?

The condition to use the oauth2 authentication is the existence of identity token.
```go
if c.auth.IdentityToken != "" {
    t, err = c.getBearerTokenOAuth2(req.Context(), challenge, scopes)
} else {
    t, err = c.getBearerToken(req.Context(), challenge, scopes)
}
```

sjhx:
Its true that the automation case using an apikey for ICCR is not affected. The interactive user authentication flow for ICCR creates a identityToken entry in .docker/config.json, driving use of OAuth2, with increasing preference for skopeo and related tools connected to strategic use of simple image signing is presenting a significant adoption and UX problem for us. Not seeing this fixed in tools till the fall would be a major problem.

TomSweeneyRedHat:
Happy green test buttons.

rhatdan:
LGTM

Issue #898, opened 2020-04-16
Arm64: "no image found" error on arm64


Pennyzct:
Hi guys

When I launch a k8s v1.17.x/crio v1.17.3 cluster on ARM node, the following errors occurred:
```
[ERROR ImagePull]: failed to pull image k8s.gcr.io/kube-controller-manager:v1.17.3: output: time="2020-04-16T15:09:47+08:00" level=fatal msg="pulling image failed: rpc error: code = Unknown desc = Error choosing image instance: no image found in manifest list for architecture arm64, variant v8, OS linux"
, error: exit status 1
```
IMO, in https://github.com/containers/image/pull/854, we imported absolute variant check when pulling multi-arch images and it brought errors when arm image entry doesn't have variant info. Like `k8s.gcr.io/kube-controller-manager:v1.17.3`
```
$ ./manifest-tool-linux-arm64 inspect k8s.gcr.io/kube-controller-manager:v1.17.3
Name:   k8s.gcr.io/kube-controller-manager:v1.17.3 (Type: application/vnd.docker.distribution.manifest.list.v2+json)
Digest: sha256:2f0bf4d08e72a1fd6327c8eca3a72ad21af3a608283423bb3c10c98e68759844
 * Contains 5 manifest references:
1    Mfst Type: application/vnd.docker.distribution.manifest.v2+json
1       Digest: sha256:8f03391c0d22e3da8d22725178efe3c4338e1920504f7f4eb4a1c7f5f40c4c6e
1  Mfst Length: 741
1     Platform:
1           -      OS: linux
1           - OS Vers:
1           - OS Feat: []
1           -    Arch: amd64
1           - Variant:
1           - Feature:
1     # Layers: 2
         layer 1: digest = sha256:597de8ba0c30cdd0b372023aa2ea3ca9b3affbcba5ac8db922f57d6cb67db7c8
         layer 2: digest = sha256:02c23a6c0b486ac6845f53754f33dbf44c0807ac45fd7ea83f7f4c7943674888

2    Mfst Type: application/vnd.docker.distribution.manifest.v2+json
2       Digest: sha256:fd2317542b55c3603c79644bd6eb86b39721f9d20989f6fe31a5e15533e400fb
2  Mfst Length: 741
2     Platform:
2           -      OS: linux
2           - OS Vers:
2           - OS Feat: []
2           -    Arch: arm
2           - Variant:
2           - Feature:
2     # Layers: 2
[...]

PR #899, opened 2020-04-17
Platform matching cleanups

mtrmac:
This is a set of simplifications created while working on #898 , but independent from that.

Filing it as a separate PR that should not change behavior (apart from some error messages) while I continue to work on a full solution for #898 , to make the actual fix later smaller and easier to review.


rhatdan:
LGTM
@vrothberg PTAL

PR #902, opened 2020-04-22
Support layer deltas

alexlarsson:
Deltas are a way to avoid downloading a full copy if a layer tar file
if you have a previous version of the layer available locally. In
testing these deltas have been shown to be around 10x to 100x smaller
than the .tar.gz files for typical Linux base images.

In the typical client-side case we have some previous version of the
image stored in container-storage somewhere, which means that we have
an uncompressed files available, but not the actual tarball
(compressed or not).

This means we can use github.com/alexlarsson/tar-diff which takes
two tar files and produces a delta file which when applied on the
untar:ed content of the first tarfile produces the (bitwise identical)
content of the uncompressed second tarfile. It just happens that the
uncompressed tarfile is exactly what we need to reproduce, because that
is how the layers are refered to in the image config (the DiffIDs).

How this works is that we use OCI artifacts to store, for each regular
image a manifest with information about the available deltas for the
image.  This image looks like a regular manifest, except each layer
contains a tar-diff (as a blob) an uses the existing annotations key
to record which DiffIDs the layer applies to.

For example, a manifest would look like this:

```
{
  "schemaVersion": 2,
  "config": {
    "mediaType": "application/vnd.oci.image.config.v1+json",
    "digest": "sha256:ca3d163bab055381827226140568f3bef7eaac187cebd76878e0b63e9e442356",
    "size": 3
  },
  "layers": [
    {
      "mediaType": "application/vnd.redhat.tardiff",
      "digest":
"sha256:49402288de20a465616174a38aca4746f46be2c3f9519fe4d14fc7f83f44a32a",
      "size": 7059734,
      "annotations": {
          "com.redhat.deltaFrom":
"sha256:b9137868142acd7ce4d62216e2b03e63e9800e2b647bf682492d3e9c5e66277c",
          "com.redhat.deltaTo":
"sha256:c88d2d437799c2879fded33ee358429e1eb954968a25f3153e2e0e26fef7ef28"
      }
    }
  ]
}
```

The config blob is just an json file containing "{}". Ideally it
should not be of type application/vnd.oci.image.config.v1+json,
because that is reserved for docker-style images. However, as
explained in https://github.com/deislabs/oras/issues/129, docker hub
doesn't currently support any other type. For registries that support
OCI artifacts we should instead use some other type so that tooling
can know that this is not a regular image.

The way we attach the delta manifest to the image is that we store it
in the same repo and a tag name based on the manifest digest like
"delta-${shortid}".

The delta layers record which DiffID they apply to, which is what we
want to use to look up the pre-existing layers to use as delta source
material, and it is what the delta apply will generate. This means
however that using the deltas only works if we're allowed to
substitute blobs, but this doesn't seem to be an issue in the typical
case.

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"os"
 	"reflect"
+	"sort"
 	"strings"
 	"sync"
 	"time"
@@ -23,6 +24,7 @@ import (
 	"github.com/containers/image/v5/types"
 	"github.com/containers/ocicrypt"
 	encconfig "github.com/containers/ocicrypt/config"
+	"github.com/containers/tar-diff/pkg/tar-patch"
 	digest "github.com/opencontainers/go-digest"
 	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
@@ -790,6 +792,11 @@ func (ic *imageCopier) copyLayers(ctx context.Context) error {
 		srcInfosUpdated = true
 	}
 
+	deltaLayers, err := types.ImageDeltaLayers(ic.src, ctx)
+	if err != nil {
+		return err
+	}
+
 	type copyLayerData struct {
 		destInfo types.BlobInfo
 		diffID   digest.Digest
@@ -809,7 +816,7 @@ func (ic *imageCopier) copyLayers(ctx context.Context) error {
 	}
 
 	data := make([]copyLayerData, numLayers)
-	copyLayerHelper := func(index int, srcLayer types.BlobInfo, toEncrypt bool, pool *mpb.Progress) {
+	copyLayerHelper := func(index int, srcLayer types.BlobInfo, toEncrypt bool, pool *mpb.Progress, deltaLayers []types.BlobInfo) {
 		defer copySemaphore.Release(1)
 		defer copyGroup.Done()
 		cld := copyLayerData{}
@@ -824,7 +831,7 @@ func (ic *imageCopier) copyLayers(ctx context.Context) error {
 				logrus.Debugf("Skipping foreign layer %q copy to %s", cld.destInfo.Digest, ic.c.dest.Reference().Transport().Name())
 			}
 		} else {
-			cld.destInfo, cld.diffID, cld.err = ic.copyLayer(ctx, srcLayer, toEncrypt, pool)
+			cld.destInfo, cld.diffID, cld.err = ic.copyLayer(ctx, index, srcLayer, toEncrypt, pool, deltaLayers)
 		}
 		data[index] = cld
 	}
@@ -861,7 +868,7 @@ func (ic *imageCopier) copyLayers(ctx context.Context) error {
 				return errors.Wrapf(err, "Can't acquire semaphore")
 			}
 			copyGroup.Add(1)
-			go copyLayerHelper(i, srcLayer, encLayerBitmap[i], progressPool)
+			go copyLayerHelper(i, srcLayer, encLayerBitmap[i], progressPool, deltaLayers)
 		}
 
 		// A call to copyGroup.Wait() is done at this point by the defer above.
@@ -1043,9 +1050,83 @@ type diffIDResult struct {
 	err    error
 }
 
+// Get all the deltas that apply to this layer
+func (ic *imageCopier) getMatchingDeltaLayers(ctx context.Context, srcIndex int, deltaLayers []types.BlobInfo) (digest.Digest, []*types.BlobInfo) {
+	if deltaLayers == nil {
+		return "", nil
+	}
+	config, _ := ic.src.OCIConfig(ctx)
+	if config == nil || config.RootFS.DiffIDs == nil || len(config.RootFS.DiffIDs) <= srcIndex {
+		return "", nil
+	}
+
+	layerDiffId := config.RootFS.DiffIDs[srcIndex]
+
+	var matchingLayers []*types.BlobInfo
+	for i := range deltaLayers {
+		deltaLayer := &deltaLayers[i]
+		to := deltaLayer.Annotations["io.github.containers.delta.to"]
+		if to == layerDiffId.String() {
+			matchingLayers = append(matchingLayers, deltaLayer)
+		}
+	}
+
+	return layerDiffId, matchingLayers
+}
+
+// Looks at which of the matching delta froms have locally available data and picks the best one
+func (ic *imageCopier) resolveDeltaLayer(ctx context.Context, matchingDeltas []*types.BlobInfo) (io.ReadCloser, tar_patch.DataSource, types.BlobInfo, error) {
+	// Sort smallest deltas so we favour the smallest useable one
+	sort.Slice(matchingDeltas, func(i, j int) bool {
+		return matchingDeltas[i].Size < matchingDeltas[j].Size
+	})
+
+	for i := range matchingDeltas {
+		matchingDelta := matchingDeltas[i]
+		from := matchingDelta.Annotations["io.github.containers.delta.from"]
+		fromDigest, err := digest.Parse(from)
+		if err != nil {
+			continue // Silently ignore if server specified a werid format
+		}
+
+		dataSource, err := types.ImageDestinationGetLayerDeltaData(ic.c.dest, ctx, fromDigest)
+		if err != nil {
+			return nil, nil, types.BlobInfo{}, err // Internal error
+		}
+		if dataSource == nil {
+			continue // from layer doesn't exist
+		}
+
+		logrus.Debugf("Using delta %v for DiffID %v", matchingDelta.Digest, fromDigest)
+
+		deltaStream, _, err := ic.c.rawSource.GetBlob(ctx, *matchingDelta, ic.c.blobInfoCache)
+		if err != nil {
+			return nil, nil, types.BlobInfo{}, errors.Wrapf(err, "Error reading delta blob %s", matchingDelta.Digest)
+		}
+		return deltaStream, dataSource, *matchingDelta, nil
+	}
+	return nil, nil, types.BlobInfo{}, nil
+}
+
+func (ic *imageCopier) canUseDeltas(srcInfo types.BlobInfo) (bool, string) {
+	// Deltas rewrite the manifest to refer to the uncompressed digest, so we must be able to substiture blobs
+	if !ic.canSubstituteBlobs {
+		return false, ""
+	}
+
+	switch srcInfo.MediaType {
+	case manifest.DockerV2Schema2LayerMediaType, manifest.DockerV2SchemaLayerMediaTypeUncompressed:
+		return true, manifest.DockerV2SchemaLayerMediaTypeUncompressed
+	case imgspecv1.MediaTypeImageLayer, imgspecv1.MediaTypeImageLayerGzip, imgspecv1.MediaTypeImageLayerZstd:
+		return true, imgspecv1.MediaTypeImageLayer
+	}
+
+	return false, ""
+}
+
 // copyLayer copies a layer with srcInfo (with known Digest and Annotations and possibly known Size) in src to dest, perhaps compressing it if canCompress,
 // and returns a complete blobInfo of the copied layer, and a value for LayerDiffIDs if diffIDIsNeeded
-func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, toEncrypt bool, pool *mpb.Progress) (types.BlobInfo, digest.Digest, error) {
+func (ic *imageCopier) copyLayer(ctx context.Context, srcIndex int, srcInfo types.BlobInfo, toEncrypt bool, pool *mpb.Progress, deltaLayers []types.BlobInfo) (types.BlobInfo, digest.Digest, error) {
 	cachedDiffID := ic.c.blobInfoCache.UncompressedDigest(srcInfo.Digest) // May be ""
 	// Diffs are needed if we are encrypting an image or trying to decrypt an image
 	diffIDIsNeeded := ic.diffIDsAreNeeded && cachedDiffID == "" || toEncrypt || (isOciEncrypted(srcInfo.MediaType) && ic.ociDecryptConfig != nil)
@@ -1061,6 +1142,52 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, to
 			bar := ic.c.createProgressBar(pool, srcInfo, "blob", "skipped: already exists")
 			bar.SetTotal(0, true)
 			return blobInfo, cachedDiffID, nil
+		}
+	}
+
+	// First look for a delta that matches this layer and substitute the result of that
+	if ok, deltaResultMediaType := ic.canUseDeltas(srcInfo); ok {
+		// Get deltas going TO this layer
+		deltaDiffID, matchingDeltas := ic.getMatchingDeltaLayers(ctx, srcIndex, deltaLayers)
+		// Get best possible FROM delta
+		deltaStream, deltaDataSource, matchingDelta, err := ic.resolveDeltaLayer(ctx, matchingDeltas)
+		if err != nil {
+			return types.BlobInfo{}, "", err
+		}
+		if deltaStream != nil {
+			bar := ic.c.createProgressBar(pool, matchingDelta, "delta", "done")
+
+			wrappedDeltaStream := bar.ProxyReader(deltaStream)
+
+			// Convert deltaStream to uncompressed tar layer stream
+			pr, pw := io.Pipe()
+			go func() {
+				if err := tar_patch.Apply(wrappedDeltaStream, deltaDataSource, pw); err != nil {
+					// We will notice this error when failing to verify the digest, so leave it be
+					logrus.Infof("Failed to apply layer delta: %v", err)
+				}
+				deltaDataSource.Close()
+				deltaStream.Close()
+				wrappedDeltaStream.Close()
+				pw.Close()
+			}()
+			defer pr.Close()
+
+			// Copy uncompressed tar layer to destination, verifying the diffID
+			blobInfo, err := ic.c.copyBlobFromStream(ctx, pr, types.BlobInfo{Digest: deltaDiffID, Size: -1, MediaType: deltaResultMediaType, Annotations: srcInfo.Annotations}, nil, ic.canModifyManifest, false, toEncrypt, nil)
+			if err != nil {
+				return types.BlobInfo{}, "", err
+			}
+
+			bar.SetTotal(matchingDelta.Size, true)
+
+			// We verified this when streaming the applied delta above
+			diffID := deltaDiffID
+
+			// Record the fact that this blob is uncompressed
+			ic.c.blobInfoCache.RecordDigestUncompressedPair(diffID, diffID)
+
+			return blobInfo, diffID, err
 		}
 	}
 
@@ -1213,7 +1340,9 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
 		return types.BlobInfo{}, errors.Wrapf(err, "Error reading blob %s", srcInfo.Digest)
 	}
 	isCompressed := decompressor != nil
-	destStream = bar.ProxyReader(destStream)
+	if bar != nil {
+		destStream = bar.ProxyReader(destStream)
+	}
 
 	// === Send a copy of the original, uncompressed, stream, to a separate path if necessary.
 	var originalLayerReader io.Reader // DO NOT USE this other than to drain the input if no other consumer in the pipeline has done so.
@@ -1230,6 +1359,11 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
 	if canModifyBlob && isOciEncrypted(srcInfo.MediaType) {
 		// PreserveOriginal due to any compression not being able to be done on an encrypted blob unless decrypted
 		logrus.Debugf("Using original blob without modification for encrypted blob")
+		compressionOperation = types.PreserveOriginal
+		inputInfo = srcInfo
+	} else if canModifyBlob && manifest.IsNoCompressType(srcInfo.MediaType) {
+		// This is a blob we should not repack, such as a delta
+		logrus.Debugf("Using original blob without modification for no-compress type")
 		compressionOperation = types.PreserveOriginal
 		inputInfo = srcInfo
 	} else if canModifyBlob && c.dest.DesiredLayerCompression() == types.Compress && !isCompressed {
--- a/types/types.go
+++ b/types/types.go
@@ -258,6 +258,36 @@ type ImageSource interface {
 	LayerInfosForCopy(ctx context.Context, instanceDigest *digest.Digest) ([]BlobInfo, error)
 }
 
+// This is an optional extension of ImageSource
+type DeltaImageSource interface {
+	ImageSource
+	// GetDeltaManifest returns the delta manifest for the current image, as well as its type, if it exist.
+	// No error is returned if no delta manifest exists, just a nil slice
+	// It may use a remote (= slow) service.
+	// If instanceDigest is not nil, it contains a digest of the specific manifest instance to retrieve deltas for (when the primary manifest is a manifest list);
+	// this never happens if the primary manifest is not a manifest list (e.g. if the source never returns manifest lists).
+	GetDeltaManifest(ctx context.Context, instanceDigest *digest.Digest) ([]byte, string, error)
+	// GetDeltaIndex returns an ImageReference that can be used to update the delta index for deltas for Image.
+	// If deltas are not supported it will return nil
+	GetDeltaIndex(ctx context.Context) (ImageReference, error)
+}
+
+// These are wrappers for ImageSource extensions that do the dynamic type cast required to work if the type is supported
+// and the default implementation otherwise
+func ImageSourceGetDeltaManifest(src ImageSource, ctx context.Context, instanceDigest *digest.Digest) ([]byte, string, error) {
+	if d, ok := src.(DeltaImageSource); ok {
+		return d.GetDeltaManifest(ctx, instanceDigest)
+	}
+	return nil, "", nil
+}
+
+func ImageSourceGetDeltaIndex(src ImageSource, ctx context.Context) (ImageReference, error) {
+	if d, ok := src.(DeltaImageSource); ok {
+		return d.GetDeltaIndex(ctx)
+	}
+	return nil, nil
+}
+
 // ImageDestination is a service, possibly remote (= slow), to store components of a single image.
 //
 // There is a specific required order for some of the calls:
@@ -329,6 +359,23 @@ type ImageDestination interface {
 	// - Uploaded data MAY be visible to others before Commit() is called
 	// - Uploaded data MAY be removed or MAY remain around if Close() is called without Commit() (i.e. rollback is allowed but not guaranteed)
 	Commit(ctx context.Context, unparsedToplevel UnparsedImage) error
+}
+
+// This is an optional extension of ImageSource
+type DeltaImageDestination interface {
+	ImageDestination
+	// Tries to get access to the uncompressed data of a given DiffID that is locally available
+	// This data is used to apply a delta from this layer
+	// If deltas are not supported or the layer is not available, nil is returned (and no error)
+	GetLayerDeltaData(ctx context.Context, diffID digest.Digest) (DeltaDataSource, error)
+}
+
+func ImageDestinationGetLayerDeltaData(dst ImageDestination, ctx context.Context, diffID digest.Digest) (DeltaDataSource, error) {
+	if d, ok := dst.(DeltaImageDestination); ok {
+		return d.GetLayerDeltaData(ctx, diffID)
+	}
+	return nil, nil
+
 }
 
 // ManifestTypeRejectedError is returned by ImageDestination.PutManifest if the destination is in principle available,
@@ -407,6 +454,21 @@ type Image interface {
 	// Size returns an approximation of the amount of disk space which is consumed by the image in its current
 	// location.  If the size is not known, -1 will be returned.
 	Size() (int64, error)
+}
+
+// This is an optional extension of Image
+type DeltaImage interface {
+	Image
+	// Downloads and parses the delta manifest for the image, returning the available delta layers
+	// If no deltas available, returns nil without an error
+	DeltaLayers(ctx context.Context) ([]BlobInfo, error)
+}
+
+func ImageDeltaLayers(img Image, ctx context.Context) ([]BlobInfo, error) {
+	if d, ok := img.(DeltaImage); ok {
+		return d.DeltaLayers(ctx)
+	}
+	return nil, nil
 }
 
 // ImageCloser is an Image with a Close() method which must be called by the user.
@@ -622,4 +684,14 @@ type ProgressProperties struct {
 	// The additional offset which has been downloaded inside the last update
 	// interval. Will be reset after each ProgressEventRead event.
 	OffsetUpdate uint64
+}
+
+// This is an interface that allows you to access existing local data to
+// use as existing content when applying a delta file
+// It is identical to the DataSource interface in tar_diff but re-exported
+// here to avoid unncecceary dependencies
+type DeltaDataSource interface {
+	io.ReadSeeker
+	io.Closer
+	SetCurrentFile(file string) error
 }
--- a/copy/manifest_test.go
+++ b/copy/manifest_test.go
@@ -9,6 +9,7 @@ import (
 	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/image/v5/manifest"
 	"github.com/containers/image/v5/types"
+	digest "github.com/opencontainers/go-digest"
 	"github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -75,6 +76,15 @@ func (f fakeImageSource) SupportsEncryption(ctx context.Context) bool {
 	panic("Unexpected call to a mock function")
 }
 func (f fakeImageSource) Size() (int64, error) {
+	panic("Unexpected call to a mock function")
+}
+func (f fakeImageSource) GetDeltaManifest(ctx context.Context, instanceDigest *digest.Digest) ([]byte, string, error) {
+	panic("Unexpected call to a mock function")
+}
+func (f fakeImageSource) GetDeltaIndex(ctx context.Context) (types.ImageReference, error) {
+	panic("Unexpected call to a mock function")
+}
+func (f fakeImageSource) DeltaLayers(ctx context.Context) ([]types.BlobInfo, error) {
 	panic("Unexpected call to a mock function")
 }
 
--- a/docker/docker_image_src.go
+++ b/docker/docker_image_src.go
@@ -203,6 +203,44 @@ func (s *dockerImageSource) fetchManifest(ctx context.Context, tagOrDigest strin
 	return manblob, simplifyContentType(res.Header.Get("Content-Type")), nil
 }
 
+func (s *dockerImageSource) GetDeltaManifest(ctx context.Context, instanceDigest *digest.Digest) ([]byte, string, error) {
+	// Get the real manifest digest
+	srcManifestDigest, err := s.manifestDigest(ctx, instanceDigest)
+	if err != nil {
+		return nil, "", err
+	}
+
+	// Load the delta manifest index
+	ib, _, err := s.fetchManifest(ctx, "_deltaindex")
+	// Don't return error if the  manifest doesn't exist, only for internal errors
+	// Deltas are an optional optimization anyway
+	if err == nil {
+		index, err := manifest.OCI1IndexFromManifest(ib)
+		if err != nil {
+			return nil, "", err
+		}
+
+		// Look up the delta manifest in the index by the real manifest digest
+		for _, manifest := range index.Manifests {
+			if manifest.Annotations["io.github.containers.delta.target"] == srcManifestDigest.String() {
+				return s.fetchManifest(ctx, manifest.Digest.String())
+			}
+		}
+	}
+
+	// No delta
+	return nil, "", nil
+}
+
+func (s *dockerImageSource) GetDeltaIndex(ctx context.Context) (types.ImageReference, error) {
+	deltaRef, err := reference.WithTag(s.logicalRef.ref, "_deltaindex")
+	if err != nil {
+		return nil, err
+	}
+
+	return newReference(deltaRef)
+}
+
 // ensureManifestIsLoaded sets s.cachedManifest and s.cachedManifestMIMEType
 //
 // ImageSource implementations are not required or expected to do any caching,
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,7 @@ require (
 	github.com/containers/libtrust v0.0.0-20190913040956-14b96171aa3b
 	github.com/containers/ocicrypt v1.0.2
 	github.com/containers/storage v1.20.2
+	github.com/containers/tar-diff v0.1.2
 	github.com/docker/distribution v2.7.1+incompatible
 	github.com/docker/docker v1.4.2-0.20191219165747-a9416c67da9f
 	github.com/docker/docker-credential-helpers v0.6.3
--- a/go.sum
+++ b/go.sum
@@ -55,6 +55,8 @@ github.com/containers/storage v1.20.1 h1:2XE4eRIqSa6YjhAZjNwIkIKE6+Miy+5WV8l1KzY
 github.com/containers/storage v1.20.1/go.mod h1:RoKzO8KSDogCT6c06rEbanZTcKYxshorB33JikEGc3A=
 github.com/containers/storage v1.20.2 h1:tw/uKRPDnmVrluIzer3dawTFG/bTJLP8IEUyHFhltYk=
 github.com/containers/storage v1.20.2/go.mod h1:oOB9Ie8OVPojvoaKWEGSEtHbXUAs+tSyr7RO7ZGteMc=
+github.com/containers/tar-diff v0.1.2 h1:6E04zGCdCsSJ8SoApFLiYkAxqkFllcKOaOATBUGydL4=
+github.com/containers/tar-diff v0.1.2/go.mod h1:9/tnBUlqmoW1bz83CQAW9wC+EQCH+h1Wn8uq3VxLbMc=
 github.com/coreos/go-systemd v0.0.0-20190321100706-95778dfbb74e/go.mod h1:F5haX7vjVVG0kc13fIWeqUViNPyEJxv/OmvnBo0Yme4=
 github.com/cpuguy83/go-md2man/v2 v2.0.0-20190314233015-f79a8a8ca69d/go.mod h1:maD7wRr/U5Z6m/iR4s+kqSMx2CaBsrgA7czyZG/E6dU=
 github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
--- a/image/unparsed.go
+++ b/image/unparsed.go
@@ -66,6 +66,20 @@ func (i *UnparsedImage) Manifest(ctx context.Context) ([]byte, string, error) {
 	return i.cachedManifest, i.cachedManifestMIMEType, nil
 }
 
+func (i *UnparsedImage) DeltaLayers(ctx context.Context) ([]types.BlobInfo, error) {
+	// Note that GetDeltaManifest can return nil with a nil error. This is ok if no deltas exist
+	mb, mt, err := types.ImageSourceGetDeltaManifest(i.src, ctx, i.instanceDigest)
+	if mb == nil {
+		return nil, err
+	}
+
+	m, err := manifestInstanceFromBlob(ctx, nil, i.src, mb, mt)
+	if err != nil {
+		return nil, err
+	}
+	return m.LayerInfos(), nil
+}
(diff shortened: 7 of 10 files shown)


alexlarsson:
To test this I've been using "skopeo copy", it works fine if you just
use the new version of containers/image. However, to generate deltas I
have a branch that supports the generate-delta command here:

https://github.com/alexlarsson/skopeo/tree/wip/deltas

I also create a set of test images on docker hub:
  https://hub.docker.com/repository/registry-1.docker.io/alexl/fedoratest

This has a few different versions of fedora 32, including deltas between then, which were created like this:

```
$ skopeo generate-delta docker://alexl/fedoratest:32-2019-09-27 docker://alexl/fedoratest:32-2019-08-26
Generating delta for layer sha256:c88d2d437799c2879fded33ee358429e1eb954968a25f3153e2e0e26fef7ef28 from sha256:b9137868142acd7ce4d62216e2b03e63e9800e2b647bf682492d3e9c5e66277c
Downloading 'from' blob sha256:ab8d40319bf2ab4567c7e7cfa69aca0983e443e96ee67afc4e45f19aba0d07d1
Downloading 'to' blob sha256:a39edc9e7bc3a586926c94144a8c7ebc83dbfaa17c2a60f4ad56df7066cba285
Generating delta
Generated delta size 6.7 MB, original layer 66.5 MB
Uploading delta layer sha256:49402288de20a465616174a38aca4746f46be2c3f9519fe4d14fc7f83f44a32a
Uploading delta manifest
```

As an example of using this is:
```
# First a full image
$ skopeo copy docker://alexl/fedoratest:32-2019-08-26 containers-storage:fedoratest:32-2019-08-26
Copying blob ab8d40319bf2 [============>-------------------------] 23.2MiB / 66.3MiB
Copying blob ab8d40319bf2 [====================================>-] 64.3MiB / 66.3MiB
Copying blob ab8d40319bf2 done
Copying config 5895d009c5 done
Storing signatures
# Now one with delta
$ skopeo copy docker://alexl/fedoratest:32-2019-09-27 containers-storage:fedoratest:32-2019-09-27
Copying delta 49402288de2 [--------------------------------------] 0.0b / 6.7MiB
Copying delta 49402288de2 done
Copying config e13031c001 done
Storing signatures
```

There are some outstanding questions.

I put tar-diff on my github. Should we move it to the containers
organization?

The names of the media type and annotations now contain "redhat",
maybe we should try to register some more independent names for these.

The use of tags for the deltas makes it necessary to manually garbage
collect deltas. Is there some other way we could refer to them from an
image? I would like to avoid having to modify the image itself when
adding the deltas, because doing so changes the digest of it and may
invalidate things refering to it. Maybe we could use image lists? I'm
not sure how they are used in practice, should they also avoid
changing in the same way as the actual image manifests?

Also, currently the code is always using a regular OCI image config
media type for the config blob in delta manifest. It is more typical
in OCI artifact use to have a custom media type for these so that
regular container tools can filter out these images. However, docker
hub doesn't currently support such media types so for now I'm using
the old one so it can be demoed easily. We can easily try a custom
media type and use that if it works and fall back otherwise.

alexlarsson:
Force-pushed with sign-off and adding the test scaffolding for the new interface methods.

vrothberg:
Hi @alexlarsson, thanks! I am quite excited about your work.

At the moment, I am deeply buried in Podman work, so I'll have a hard time to find enough time to review such a fairly complex PR. I'll try to allocate a couple of hours next week.

@giuseppe @mtrmac @nalind PTAL

CC @fatherlinux @mrunalp @rhatdan as it's super rad.

alexlarsson:
Also, @josephschorr, you might want to take a look at this as you're involved with both quay and OCI artifacts which are related to this.

alexlarsson:
>The use of tags for the deltas makes it necessary to manually garbage
>collect deltas. Is there some other way we could refer to them from an
>image? I would like to avoid having to modify the image itself when
>adding the deltas, because doing so changes the digest of it and may
>invalidate things refering to it. Maybe we could use image lists? I'm
>not sure how they are used in practice, should they also avoid
>changing in the same way as the actual image manifests?

Maybe we can use a single tag `deltas` which maps to a image index, where each image in the index refers to a delta manifest, specifying which using the annotation fields of the index.manifests array elements. That way we only have one tag for deltas, simplifying delta garbage collection. We would still need to remove unnecessary deltas from the index, but it seems more localized.

alexlarsson:
Also, CC @owtaylor

giuseppe:
CC @vbatts

giuseppe:
ideally it should be part of the OCI specs, but since it is not a breaking change and if there are no deltas we just fallback to the existing code path, we could probably carry it without blocking on OCI.

@vbatts is https://github.com/opencontainers/artifacts the proper place to discuss and develop such stuff?

alexlarsson:
The artifacts spec is more about detailing how in *general* you would store something other than pure docker images on a registry, they are not involved in standardizing the format of individual artifacts (other than possibly to ratify/list known media types).

If anything this should be related to the image spec itself, but as you say it is quite independent of it and it doesn't necessarily have to be affected at all by this, which is nice as getting such changes in would be a very slow work...

alexlarsson:
I pushed a new commit (which hasn't shown up here yet, whats up with github?) that changes it to use a single branch "deltas" which has an OCI index and then for each image in that a `com.redhat.deltaTarget` annotation that specifies the manifest id it is for. 

The one issue with this is that updating the deltas tag seems a bit racy. We download the old one, update it and put it back, and if someone else does this at the same time we would lose one of the updates. Is there any way to do atomic updates of oci images? Like, set the tag to this new blob, but fail if the current version is not $digest. Then we could loop it until we successfully updated it.

types/types.go
```diff
@@ -256,6 +257,17 @@ type ImageSource interface {
 	// The Digest field is guaranteed to be provided; Size may be -1.
 	// WARNING: The list may contain duplicates, and they are semantically relevant.
 	LayerInfosForCopy(ctx context.Context, instanceDigest *digest.Digest) ([]BlobInfo, error)
+	// GetDeltaManifest returns the delta manifest for the current image, as well as its type, if it exist.
+	// No error is returned if no delta manifest exists, just a nil slice
+	// It may use a remote (= slow) service.
+	// If instanceDigest is not nil, it contains a digest of the specific manifest instance to retrieve deltas for (when the primary manifest is a manifest list);
+	// this never happens if the primary manifest is not a manifest list (e.g. if the source never returns manifest lists).
+	GetDeltaManifest(ctx context.Context, instanceDigest *digest.Digest) ([]byte, string, error)
```
mtrmac, line 265: Hum, an API break.

I suppose we’ll now have to figure out how to extend these interfaces without breaking API, it had to come one day.

(Let’s figure out the right approach first and deal with the API later, just highlighting that this will have to happen.)

types/types.go
```diff
@@ -256,6 +257,17 @@ type ImageSource interface {
...
+	// GetDeltaManifestDestination returns an ImageDestination that can be used to update the delta manifest for this Image.
+	// If deltas are not supported it will return nil
+	// If instanceDigest is not nil, it contains a digest of the specific manifest instance to retrieve deltas for (when the primary manifest is a manifest list);
+	// this never happens if the primary manifest is not a manifest list (e.g. if the source never returns manifest lists).
+	GetDeltaManifestDestination(ctx context.Context, instanceDigest *digest.Digest) (ImageDestination, error)
```
mtrmac, line 270: This feels intuitively weird. Why should a “source” care about creating destinations? At a first glance this might be a better first for `ImageReference` or maybe `ImageTransport` (or somehow split between “source” and `types.Image`), but I haven’t read the full PR yet.

types/types.go
```diff
@@ -5,6 +5,7 @@ import (
 	"io"
 	"time"
 
+	"github.com/alexlarsson/tar-diff/pkg/tar-patch"
```
mtrmac, line 8: `types.go` should be extremely cheap to use, without dragging in dependencies like compression implementations.

copy/copy.go
```diff
@@ -1040,9 +1047,83 @@ type diffIDResult struct {
...
+	if config == nil || config.RootFS.DiffIDs == nil || len(config.RootFS.DiffIDs) <= srcIndex {
+		return "", nil
+	}
+
+	layerDiffId := config.RootFS.DiffIDs[srcIndex]
+
+	var matchingLayers []*types.BlobInfo
+	for i := range deltaLayers {
+		deltaLayer := &deltaLayers[i]
+		to := deltaLayer.Annotations["com.redhat.deltaTo"]
```
mtrmac, line 1065: The annotation name, and the concept of “RootFS”, feels like something that should be a part of `image.genericManifest`/`types.Image` instead of assumed to be generally applicable. I’m not immediately sure how practical that is.

mtrmac (commented):
Thanks!

I need to handle a few bugs in the coming days, so for now here just a few quick thoughts on things that caught my eye.

types/types.go
alexlarsson: Yes, I think reporting a ImageReference makes more sense. In fact when using this in skopeo i had to do call ImageDestination.Reference() on it, so I already kinda planned on this change.

types/types.go
alexlarsson: The only real dependency on it is that we return a tar_patch.DataSource interface, which is a simple interface to look up stream data given a pathname. We can probably wrap this interface inside types.go to avoid the external dependency.

josephschorr:
> Also, @josephschorr, you might want to take a look at this as you're involved with both quay and OCI artifacts which are related to this.

@alexlarsson The referenced blobs in the `deltaFrom`... from where are they being served? Are these annotations the *only* reference to them?

josephschorr:
> Then we could loop it until we successfully updated it.

Gah! Please don't design calls that need to loop against registries!

If you need atomic updates, then it likely needs to be done as an extension to the registry protocol.

alexlarsson:
> > Also, @josephschorr, you might want to take a look at this as you're involved with both quay and OCI artifacts which are related to this.
> 
> @alexlarsson The referenced blobs in the `deltaFrom`... from where are they being served? Are these annotations the _only_ reference to them?

The "from" blobs need not be served at all. The setup is like this:

 * Tag `latest` points to `sha256:AAA`, which has an uncompressed id `sha256:aaa`
 * Tag `deltas` points to an image list, which points to the delta manifest
 * The delta mainfest points to the actual blobs that are tardiffs
 * The delta manifest has a key from: `sha256:bbb`, and to: `sha256:aaa`

Here, `sha256:bbb` need not be served from the registry at all, instead what is important is that the *client* may have it available locally (for example if he pulled a previous version of the `latest` tag. If so, he can use this data with the tardiff to reprocude the `sha256:aaa` layer

alexlarsson:
> > Then we could loop it until we successfully updated it.
> 
> Gah! Please don't design calls that need to loop against registries!
> 
> If you need atomic updates, then it likely needs to be done as an extension to the registry protocol.

It seems like a standard way to do an atomic update though. You get the current version of the index, update it locally, and the PUT it with an If-Match header containing the current ETAG. The PUT will then succeed only if (atomically) the current version of the tag is the same as the one in the header. If it is not it will fail, and we loop.

Its not like we're busy looping, we will only loop in the case where there is an actual conflict.

josephschorr:
> Its not like we're busy looping, we will only loop in the case where there is an actual conflict.

It could, however, result in a lot of clients looping because they are all hitting against one another. This is a real concern with any design consideration for an API that will likely be used by *millions* of clients.

> The config blob is just an json file containing "{}"

Yeah, be aware Quay will (correctly) fail to validate this.

Also, it feels like if this "delta" manifest is a manifest representing the delta, then the information about `from` and `to` should, in theory, be contained in the *config* (it could be duplicated in the annotations, I guess)

> We can easily try a custom media type and use that if it works and fall back otherwise.

Please don't fallback. If you do, someone is invariably going to push the content with the "wrong" media type to a registry that does not do validation, then they'll try to mirror that content into a registry that *does* do validation, it will fail, and the *mirroring* registry will get blamed. We should *not* be abusing the content types simply because registries do not *yet* support artifacts.

> The use of tags for the deltas makes it necessary to manually garbage
collect deltas. Is there some other way we could refer to them from an
image? I would like to avoid having to modify the image itself when
adding the deltas, because doing so changes the digest of it and may
invalidate things refering to it. Maybe we could use image lists? I'm
not sure how they are used in practice, should they also avoid
changing in the same way as the actual image manifests?

I think using an OCI Index here is the way to go: The index could contain the "normal" manifest as an entry, as well as the delta manifest as a secondary entry. That way, the delta+normal manifest can be tagged as a unit, and it'll also hold the references to things as a single GCable unit. If a new delta needs to be generated, the *previous* should be contained in the new index as well (up to some limit, after which point you could compact the deltas or somesuch).

mtrmac:
> > The use of tags for the deltas makes it necessary to manually garbage
> > collect deltas. Is there some other way we could refer to them from an
> > image? I would like to avoid having to modify the image itself when
> > adding the deltas, because doing so changes the digest of it and may
> > invalidate things refering to it. Maybe we could use image lists? I'm
> > not sure how they are used in practice, should they also avoid
> > changing in the same way as the actual image manifests?
> 
> I think using an OCI Index here is the way to go: The index could contain the "normal" manifest as an entry, as well as the delta manifest as a secondary entry. That way, the delta+normal manifest can be tagged as a unit, and it'll also hold the references to things as a single GCable unit.

(That does not quite work in full generality; OCI in theory supports many possible shapes of index trees, but in practice clients are not going to be interperable by doing a recursive search and combining the metadata from various levels of the tree using completely undefined criteria. There are going to be more restrictive “profiles” of the standard.

As far as c/image is concerned, OCI use mirrors the Docker schema2 use: index ~ manifest list, a single-level lookup for multi-arch images. Sure, adding delta pseudo-architectures or something like that to the existing single level of an OCI index would be possible, and it would contain registry-recognizable references to deltas.)

But…

---

Ultimately “garbage collection” is going to be problematic either way. _If_ the registry supports deletion at all (many don’t), it won’t by default delete untagged manifests (because digest references exist, and if nothing else there is a race between looking up an index by tag and reading the component manifest by digest vs. deleting the index tag). So, manifests that reference deltas are going to keep piling up and prevent deletion of the layer deltas, unless maybe the user explicitly opts in into deleting untagged manifests, and we shouldn’t assume such an opt in.

So:
- Never refer to any component of deltas by digest from the primary images (because primary images can be referred to by digest, and are ~never deleted). So, no reference to delta blobs in primary manifests, no reference to delta manifests in primary indices.
- By implication, create some other way to look up the current set of deltas. Because “the current set of deltas” for a static primary image can change over time, that probably implies a tag, separate from the tag of the primary image.
- Make delta manifests/indices _recognizable_ to the registry, so that it can treat older versions of the manifests, if untagged, as garbage to be collected.

owtaylor:
> It could, however, result in a lot of clients looping because they are all hitting against one another. This is a real concern with any design consideration for an API that will likely be used by millions of clients.

It doesn't seem likely to me to have millions, or a handful constantly updating a *single* repository. Two clients trying to update a single repository at once is a rare situation - and the registry load of an extra rejected HTTP request - trivial.

> I think using an OCI Index here is the way to go: The index could contain the "normal" manifest as an entry, as well as the delta manifest as a secondary entry. That way, the delta+normal manifest can be tagged as a unit, and it'll also hold the references to things as a single GCable unit. If a new delta needs to be generated, the previous should be contained in the new index as well (up to some limit, after which point you could compact the deltas or somesuch).

If I understand you, you are suggestion that the deltas appear in the image index that tags point to - `latest` points to an OCI index, that has (presumably) both the different arch images, and also the delta manifests for those images. This would mean that if you were telling someone to pull a version by digest instead of tag (install foo@sha256:<....>) the digest is the image index digest and dependent on the set of deltas. Deltas basically become *part* of the bundle you are pushing to the registry.

This isn't ideal for a couple of reasons:
 - the set of deltas that you want to include depends on context where you are publishing - you don't care what versions you have internally, you want deltas from the versions that have appeared in your public repository.
 - you might want to remove deltas archived image versions - there's no reason to keep deltas that say how to go to an old version from a yet-older version.

Maintenance of deltas might also be also be something that is offered as a service on some registries - a user could define a policy (deltas from the last 5 versions of the 'latest' tag) and let the registry or an external service handle creating the deltas. In this case, you wouldn't want the registry retagging the 

Those are reasons why we have generally thinking of deltas as being side-car information to an image rather than being *part* of the tagged image/image-manifest.

alexlarsson:
> > Its not like we're busy looping, we will only loop in the case where there is an actual conflict.
> 
> It could, however, result in a lot of clients looping because they are all hitting against one another. This is a real concern with any design consideration for an API that will likely be used by _millions_ of clients.

The way I see it there are three ways to do an atomic update API
 * Atomic compare and swap (what i proposed)
 * Locking
 * Adding server side support for updating the index

The last one seems vastly more complex to even specify. I think CAS would perform better in the non-contended case, and locking if you have contention. However, I doubt you'll actually see contention in updating a specific tag, so imho CAS is the better choice. But anyway that is beyond this discussion.

> > The config blob is just an json file containing "{}"
> 
> Yeah, be aware Quay will (correctly) fail to validate this.

If we claim it is of type `application/vnd.oci.image.config.v1+json`, yes. But what if we give it a different type?  Does quay support that?
 
> Also, it feels like if this "delta" manifest is a manifest representing the delta, then the information about `from` and `to` should, in theory, be contained in the _config_ (it could be duplicated in the annotations, I guess)

It could easily be in the config (only, or in both), and maybe that is conceptually more correct. However, in practice all that means is that we have to load and update two files (manifest and config) instead of just one.
 
> > We can easily try a custom media type and use that if it works and fall back otherwise.
> 
> Please don't fallback. If you do, someone is invariably going to push the content with the "wrong" media type to a registry that does not do validation, then they'll try to mirror that content into a registry that _does_ do validation, it will fail, and the _mirroring_ registry will get blamed. We should _not_ be abusing the content types simply because registries do not _yet_ support artifacts.

What is the current status of artifacts in quay? I'm working on this because redhat uses OCI instead of ofstree for the fedora flatpak runtime and the performance (vs ostree) is horrible due to the lack of deltas. Lots of people are complaining about this, and if the solution is going to have to wait until everyone implements oci artifacts that may be a problem.
 
> I think using an OCI Index here is the way to go: The index could contain the "normal" manifest as an entry, as well as the delta manifest as a secondary entry. That way, the delta+normal manifest can be tagged as a unit, and it'll also hold the references to things as a single GCable unit. If a new delta needs to be generated, the _previous_ should be contained in the new index as well (up to some limit, after which point you could compact the deltas or somesuch).

Yeah, this would definately work. I'm not really aware of how oci indexes are used in the wild. In ostree, a delta is a completely independent optimization and doesn't affect the digest of the content (which in ostree parlance would be the commit id). 

So, If at some point we upload a version of an image the tag will point to some digest, and then later we add a delta, this will change the digest of the index that the tag points to (even thought the final image digest is the same). Is this a problem to typical OCI use? (I honestly don't know)

mtrmac:
> So, If at some point we upload a version of an image the tag will point to some digest, and then later we add a delta, this will change the digest of the index that the tag points to (even thought the final image digest is the same). Is this a problem to typical OCI use? (I honestly don't know)

Not a breaking problem, but something to consider: Cluster orchestrators like OpenShift may well resolve tags to digests “early”, so that multi-node software deployments all use exactly the same image even if the tag moves. In that case, deltas added “later” would not be found by nodes that are pinned to a specific digest.

alexlarsson:
> > So, If at some point we upload a version of an image the tag will point to some digest, and then later we add a delta, this will change the digest of the index that the tag points to (even thought the final image digest is the same). Is this a problem to typical OCI use? (I honestly don't know)
> 
> Not a breaking problem, but something to consider: Cluster orchestrators like OpenShift may well resolve tags to digests “early”, so that multi-node software deployments all use exactly the same image even if the tag moves. In that case, deltas added “later” would not be found by nodes that are pinned to a specific digest.

Yeah, this seems bad to me, deltas are an independent transport thing an not part of the image itself. The original images are reconstructed perfectly, so nobody need to even be aware of the deltas. So, I think considering them a sidecar thing is the correct thing to do. 

Additionally, as mentioned before, referencing the deltas from the main manifests makes it very hard to later remove deltas, and deltas really only are useful for the latest versions of things. You rarely pull older versions, so its not worth wasting space on them.

alexlarsson:
> Ultimately “garbage collection” is going to be problematic either way. _If_ the registry supports deletion at all (many don’t), it won’t by default delete untagged manifests (because digest references exist, and if nothing else there is a race between looking up an index by tag and reading the component manifest by digest vs. deleting the index tag). So, manifests that reference deltas are going to keep piling up and prevent deletion of the layer deltas, unless maybe the user explicitly opts in into deleting untagged manifests, and we shouldn’t assume such an opt in.

What we could do is have the `deltas` tag (which points to an index of all current delta manifests) be the only thing that refers to delta manifests (and indirectly delta blobs). Then when we update this tag we could explicitly delete the old manifest version (and free the blobs for GC). Is that possible?

josephschorr:
> it won’t by default delete untagged manifests

@mtrmac We absolutely *do* delete untagged manifests. In fact, in Quay, if you have a manifest that is not referenced by a tag (either directly or indirectly via a manifest list), the manifest immediately becomes unpullable and will eventually be GCed.

> OCI use mirrors the Docker schema2 use

@mtrmac Based on numerous discussions, it seems that this will not last once we have artifacts supported. Many individuals I've spoke to have stated they intend to place, for example, signatures at the top level along images, or source references, or other even Helm charts, etc, all found in the same index, at the same level; client tooling would then choose to make use of whatever it supports.

> It doesn't seem likely to me to have millions, or a handful constantly updating a single repository. Two clients trying to update a single repository at once is a rare situation - and the registry load of an extra rejected HTTP request - trivial.

@owtaylor Sure... until you have automated tooling hitting 10,000 repositories on the same *multi-tenant* registry service, backed by the same database: then locking across indexes can become a major concern. We know this, because this has *already* been a concern for Quay, and I've argued strenuously against polling-based APIs as a result.

>  Deltas basically become part of the bundle you are pushing to the registry.

@owtaylor Yes, since they represent an alternative "view" of the same manifest layers. Otherwise, if we just use another specially-named tag, there is the major concern about overwriting, consistency and scaling.

Basically, if I were generating deltas for a manifest, it would make sense to pull that manifest, compute deltas, then push the manifest back with the deltas "added", for lack of a better term.

> But what if we give it a different type? Does quay support that?
> What is the current status of artifacts in quay?

@alexlarsson Experimentally, but not in production today. Artifacts spec is, after all, not yet standardized (heck, neither is OCI distribution, for that matter)

> However, in practice all that means is that we have to load and update two files (manifest and config) instead of just one.

@alexlarsson Yeah, hence why I said its better spec-wise, but not a requirement in my mind.

> deltas are an independent transport thing an not part of the image itself

So, ultimately, I think this discussion is showing why storing deltas-as-artifacts is conceptually not appealing. If the intention is to have this delta support be *generic*, then we should not be using deltas-in-manifests *at all*, but rather creating an extension to the OCI image distribution protocol that *provides* delta computation automatically based on the API. That would have the benefit of working for *all existing manifests*, it would allow for far less storage of duplicative information, and it could be added to existing registries without too much overhead; any registry that didn't want to support it, simply wouldn't.

mtrmac:
> > it won’t by default delete untagged manifests
> 
> @mtrmac We absolutely _do_ delete untagged manifests. In fact, in Quay, if you have a manifest that is not referenced by a tag (either directly or indirectly via a manifest list), the manifest immediately becomes unpullable and will eventually be GCed.

That’s good to know; still, docker/distribution doesn’t by default (although there is an option), so the design has to account for that.

> > OCI use mirrors the Docker schema2 use
> 
> @mtrmac Based on numerous discussions, it seems that this will not last once we have artifacts supported. Many individuals I've spoke to have stated they intend to place, for example, signatures at the top level along images, or source references, or other even Helm charts, etc, all found in the same index, at the same level; client tooling would then choose to make use of whatever it supports

Sure; c/image is a library for handling _container images_, not arbitrary payloads (and deltas of this kind are clearly not applicable to arbitrary payloads). It may well not make any sense to support Helm charts in that codebase. It all depends on how much they end up having in common and what contributors are interested in.

> If the intention is to have this delta support be _generic_, then we should not be using deltas-in-manifests _at all_, but rather creating an extension to the OCI image distribution protocol that _provides_ delta computation automatically based on the API call's parameters. That would have the benefit of working for _all existing manifests_, it would allow for far less storage of duplicative information, and it could be added to existing registries without too much overhead; any registry that didn't want to support it, simply wouldn't.

It doesn’t make much sense to me to make delta computation a registry responsibility: the publisher of the image is at least as much invested in making images quick to pull, and can invest the resources (delta creation is fairly resource-intensive) instead of the registry. Registries computing deltas completely on-demand seems fairly unlikely given the resource requirements anyway. (Registries may want to publish deltas even if the image publisher didn’t, because they would benefit from the deltas — but that would be a better fit as a data-based targeted process.)

It is also definitely easier and simpler to deploy deltas to and make them easy to mirror if that doesn’t require server-side support. Sure, that argument is treacherous because it can motivate all kinds of nasty hacks, but in this case I’m not sure that integrated server-side support would give us benefits that are not easily obtainable otherwise, and outweigh the limited/delayed support.

mtrmac:
> What we could do is have the `deltas` tag (which points to an index of all current delta manifests) be the only thing that refers to delta manifests (and indirectly delta blobs). Then when we update this tag we could explicitly delete the old manifest version (and free the blobs for GC). Is that possible?

That does seem possible in the API.

types/types.go
mtrmac: “image”reference… I do need to read the full PR

alexlarsson:
> > What we could do is have the `deltas` tag (which points to an index of all current delta manifests) be the only thing that refers to delta manifests (and indirectly delta blobs). Then when we update this tag we could explicitly delete the old manifest version (and free the blobs for GC). Is that possible?
> 
> That does seem possible in the API.

How would i do this in containers/image? There is a ImageReference.DeleteImage() operation, but it works on named references, and that would in this case be the `deltas` tag of the repository, but I don't want to delete that, i want to delete the previous manifests it pointed to (which i know the digest of since I replaced them).

josephschorr:
@alexlarsson You'd simply point the `deltas` tag to a new manifest, which should (in theory) disconnect the old one allowing for GC.

I'm still wondering, however, why having a single `deltas` tag (or another tag of the form `deltas-whatever`) is preferable to using an OCI Index: The image and its deltas are logically linked entities (that is exactly the purpose of an OCI Index) and it would easily solve the GC problem as well.

owtaylor:
> @josephschorr: I'm still wondering, however, why having a single deltas tag (or another tag of the form deltas-whatever) is preferable to using an OCI Index: The image and its deltas are logically linked entities (that is exactly the purpose of an OCI Index) and it would easily solve the GC problem as well.

It's been mentioned above some, but the basic idea is that the identity of the image should be distinct from deltas that might be used to accelerate downloading the image. If I delete deltas for an archived image to save space, it should still be `registry.example.com/owtaylor/myrepo@sha256:....`. This works because the tardiff file format we are using reconstructs tar files with the same DiffID's recorded in the target manifest's config, so a client can know that the image they reconstruct works the same as the target image, and any trust derived from known hashes or signatures still is applicable.

alexlarsson:
> @alexlarsson You'd simply point the `deltas` tag to a new manifest, which should (in theory) disconnect the old one allowing for GC.

This would work on quay, but as per above it wouldn't e.g. work on docker distribution in the default configuration, so if we could explicitly delete untagged manifests that might help there.

> I'm still wondering, however, why having a single `deltas` tag (or another tag of the form `deltas-whatever`) is preferable to using an OCI Index: The image and its deltas are logically linked entities (that is exactly the purpose of an OCI Index) and it would easily solve the GC problem as well.

This is the same reason why e.g. the git packfile checksum is not part of the git commit object checksum. The image (which includes all images and the index) is the immutable object you are interested in, and the deltas is just a way to more efficiently distribute them. You want to allow the administrator (or the some automatic framework) to add and remove deltas as it fits the current traffic. For example it is generally not very interesting to keep deltas to very old images (because typical deltas go from old to new, and people don't update from super-old to slightly-less-super-old). 

If the delta is part index, then you cannot do such administration of what deltas to keep around without modifying the ideally immutable old object. For example, if someone hardcoded the digest id for the old index for some old release that will break if we want to remove the delta from that.

josephschorr:
@alexlarsson @owtaylor Well, the digest of the *image* manifest would be unchanged, but you are correct that the overall manifest digest would be different.

It just feels odd to have a fragile linkage by naming convention, since it will be so easy for it to break and it doesn't allow for generic handling of GC without major problems.

mtrmac:
> > > What we could do is have the `deltas` tag (which points to an index of all current delta manifests) be the only thing that refers to delta manifests (and indirectly delta blobs). Then when we update this tag we could explicitly delete the old manifest version (and free the blobs for GC). Is that possible?
> > 
> > 
> > That does seem possible in the API.
> 
> How would i do this in containers/image?

(I meant the docker/distribution HTTP API above.)

>  There is a ImageReference.DeleteImage() operation, but it works on named references, and that would in this case be the `deltas` tag of the repository, but I don't want to delete that, i want to delete the previous manifests it pointed to (which i know the digest of since I replaced them).

AFAICS `DeleteImage` should work on tagged references as well (per the `refTail, err := ref.tagOrDigest()` use). And if not, it should be possible to fix that.

(Either way, the code should not break with docker/distribution registries in the default configuration, which have deletion completely disabled.)

---



> @alexlarsson You'd simply point the `deltas` tag to a new manifest, which should (in theory) disconnect the old one allowing for GC.

To be explicit, _in Quay_. docker/distribution does not do that kind of GC automatically.

josephschorr:
> AFAICS DeleteImage should work on tagged references as well

If it is invoking the `DELETE` in the V2 API, its only by manifest AFAIK:

From: https://docs.docker.com/registry/spec/api/
> For deletes, reference must be a digest or the delete will fail

Further, many registries do not implement the `DELETE` for blob, so I'd be careful about that.

mtrmac:
> > AFAICS DeleteImage should work on tagged references as well

(This was a thinko: the question was about digested references, and I meant that digested references should be handled by that code as well, although with an extra unnecessary HTTP request.)

> If it is invoking the `DELETE` in the V2 API, its only by manifest AFAIK:

Yeah, that code looks up the digest the tag points to, and deletes by digest. It’s not trivially obvious that this works when the input is a digest as well. (There’s a whole another can of worms WRT whether this is what the users want when they specify a tag, but that’s irrelevant for this PR.)

vrothberg:
> I put tar-diff on my github. Should we move it to the containers
organization?

Sure! I created https://github.com/containers/tar-diff/ and invited you. Feel free to move the code over there. I already enabled Cirrus CI on the repo. Using it is pretty straight forward by adding `.cirrus.yml` file to the root which can look as follows:
```
container:
  image: fedora:32

env:
  GOPROXY: https://proxy.golang.org

build_and_test_task:
  build_and_test_script:
    - dnf install -y skopeo
    - make ... 
```

Containerized CI is free. We do have VMs available but enabling them requires some special rain dance.

> The names of the media type and annotations now contain "redhat",
maybe we should try to register some more independent names for these.

I am bad at naming. @rhatdan @vbatts PTAL.

alexlarsson:
@vrothberg I'll move it over. I just want to clean the repo up a bit first, adding some tests, sign-off tags, etc.

alexlarsson:
For naming. One option is to use containers.github.io, which is the github pages for the containers org. Which is perhaps better than nothing.

vrothberg:
Addding one comment at a time:

> skopeo generate-delta ...

I'd love this to be part of Buildah rather than Skopeo. It's a great use case for Buildah and helps promoting that it's meant to be used beyond Dockerfiles.

alexlarsson:
> Addding one comment at a time:
> 
> > skopeo generate-delta ...
> 
> I'd love this to be part of Buildah rather than Skopeo. It's a great use case for Buildah and helps promoting that it's meant to be used beyond Dockerfiles.

I don't think that is a great fit. Buildah is for creating images, while delta generation is about maintaining a registry. People who build images should not need to care about deltas, that is up to the infra people running the registry, and the registry people shouldn't use image building tools in their sysadmin workflow.

alexlarsson:
That said, I think that some of the code to generate deltas that is now in the skopeo code should move to containers/images to make it easy to use from wherever.

vrothberg:
The discussion around what a delta image is and how it should be implemented reminds me of source containers/images (i.e., https://github.com/containers/BuildSourceImage). Such images contain the source code of the corresponding "binary" image. The long-term goal is to make this part of the OCI as license compliance is a sufficiently widespread and general problem to be part of a spec and standardize on it.

It seems that much we come up with now will be a "convention" rather than a spec; the simpler it is the easier a migration to a real standard gets. That's something we have been realizing for source images as well. For now, source images have a "-source" suffix and are hence identifiable by tag/reference rather than via new media types.

The beauty of deltas is that they can be (depending on the implementation) opt-in for the server and client, and as such, are easier to get into a spec.

Cc: @vbatts

alexlarsson:
New tar-diff at https://github.com/containers/tar-diff, and all references changed to it.

alexlarsson:
The annotations are now:
 * `io.github.containers.deltaTo`
 * `io.github.containers.deltaTo.deltaFrom`
 * `io.github.containers.deltaTarget`
    
This is not great, but it matches the containers.github.io dns address that is owned by github.com/containers, and seems better than having redhat there.

The tar-diff media type is now `application/vnd.tar-diff` instead of `application/vnd.redhat.tardiff`, because there seem to be no particular need to encode an organization name in the mime type.

mtrmac:
> The tar-diff media type is now `application/vnd.tar-diff` instead of `application/vnd.redhat.tardiff`, because there seem to be no particular need to encode an organization name in the mime type.

https://tools.ietf.org/html/rfc6838#section-3.2 .`vnd` alone does not make sense.

mtrmac:
… That said, the registration requirements (and, to an extent, even the format) for MIME types seem to be fairly widely ignored in practice (e.g. only one of the OCI types is registered). Still, we should at the very least make it _possible_ to do the right thing in the future.

alexlarsson:
> > The tar-diff media type is now `application/vnd.tar-diff` instead of `application/vnd.redhat.tardiff`, because there seem to be no particular need to encode an organization name in the mime type.
> 
> https://tools.ietf.org/html/rfc6838#section-3.2 .`vnd` alone does not make sense.

This says:

>   Vendor-tree registrations will be distinguished by the leading facet
>   "vnd.".  That may be followed, at the discretion of the registrant,
>   by either a media subtype name from a well-known producer (e.g.,
>   "vnd.mudpie") or by an IANA-approved designation of the producer's
>   name that is followed by a media type or product designation (e.g.,
>   vnd.bigcompany.funnypictures).

Where `vnd.tar-diff` seems exactly like the (allowed) example `vnd.mudpie` (but not like the other altarnative `vnd.bigcompany.funnypictures`.

alexlarsson:
Looking at the assignments (http://www.iana.org/assignments/media-types/media-types.xhtml) it does seem that most have some kind of organization, but far from all.

mtrmac:
That’s fair. With
>  "Vendor" and "producer" are construed very broadly in this context and are considered equivalent.

I don’t understand what does it mean to register a single MIME type for a “vendor”, but you’re right that it is allowed.

alexlarsson:
I updated the skopeo branch to now also support generating multiple deltas if the tag resolves to an index or list. I put a test of that here: https://hub.docker.com/repository/registry-1.docker.io/alexl/fedoratest2/

alexlarsson:
After some discussions with @josephschorr yesterday I added a io.github.containers.delta.target annotations to the delta manifest itself (same as would be in the delta index when looking up the delta manifest there). This would allow future work to automatically (via the db) look up the delta manfest for an image digest without needing the _deltaindex hack. Additionally it is also helpful for  doing automatic garbage collection of deltas where the target image has been removed.

alexlarsson:
Upstream discussion here: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/daBUKI3KkRk

alexlarsson:
Rebased to apply to master and combined all commits into one.

alexlarsson:
@mtrmac It would be nice to move forward on reviewing this. I guess the API break is a big issue. Do you have any ideas of how we best could solve that? I can add a new interface with the new interface methods and do runtime interface casts before using them. Would that be ok?

alexlarsson:
@mtrmac New version, I made all the changes in types.go runtime optional, which should make this a API compatible change. In addition that alllowed me to drop a bunch of

alexlarsson:
I updated this to the latest master. Any chance I can get someone to look at it again? Would be nice to get something in.

I noticed that deltas were discussed in the latest OCI meeting, but the general opinion seemed to be that OCI wants changes to be deployed in the wild and tested in actual use before possibly looking at standardizing them via OCI. So, I think we should try to get this deployed and test it in real use.

alexlarsson:
I also updated the skopeo branch (https://github.com/alexlarsson/skopeo/tree/wip/deltas)

vrothberg (commented):
Overall LGTM :+1: 

The [question](https://github.com/containers/image/pull/902/files#r413860731) of how handle the API parts is still open but I don't have an alternative at hand other than what the PR does already.

alexlarsson:
> Overall LGTM +1
> 
> The [question](https://github.com/containers/image/pull/902/files#r413860731) of how handle the API parts is still open but I don't have an alternative at hand other than what the PR does already.

I changed the API so now it shouldn't really be an API break. Existing implementers of the interfaces should keep working. Only if you want to support deltas do you need to implement the new interfaces.

alexlarsson:
I updated the skopeo patch to use custom mimetypes by default, so depending on the registry you may need to pass `--fallback-config-type` and/or `--fallback-layer-type`.

alexlarsson:
I merged the flatpak support for this, hope to get it into an unstable release so we can start using it.

rhatdan:
@alexlarsson Needs a rebase.
@mtrmac PTAL

rhatdan:
@lumjjb PTAL

copy/copy.go
```diff
@@ -1064,6 +1145,52 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, to
...
+		}
+		if deltaStream != nil {
+			bar := ic.c.createProgressBar(pool, matchingDelta, "delta", "done")
+
+			wrappedDeltaStream := bar.ProxyReader(deltaStream)
+
+			// Convert deltaStream to uncompressed tar layer stream
+			pr, pw := io.Pipe()
+			go func() {
+				if err := tar_patch.Apply(wrappedDeltaStream, deltaDataSource, pw); err != nil {
```
lumjjb, line 1165: Are the deltas always uncompressed/unencrypted?

copy/copy.go
```diff
@@ -1043,9 +1050,83 @@ type diffIDResult struct {
...
+func (ic *imageCopier) canUseDeltas(srcInfo types.BlobInfo) (bool, string) {
+	// Deltas rewrite the manifest to refer to the uncompressed digest, so we must be able to substiture blobs
+	if !ic.canSubstituteBlobs {
+		return false, ""
+	}
+
+	switch srcInfo.MediaType {
+	case manifest.DockerV2Schema2LayerMediaType, manifest.DockerV2SchemaLayerMediaTypeUncompressed:
+		return true, manifest.DockerV2SchemaLayerMediaTypeUncompressed
+	case imgspecv1.MediaTypeImageLayer, imgspecv1.MediaTypeImageLayerGzip, imgspecv1.MediaTypeImageLayerZstd:
```
lumjjb, line 1120: Interaction for encrypting path with decryption routine

Encryption mediatypes https://github.com/containers/ocicrypt/blob/master/spec/spec.go

copy/copy.go
```diff
@@ -1064,6 +1145,52 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, to
 		}
 	}
 
+	// First look for a delta that matches this layer and substitute the result of that
+	if ok, deltaResultMediaType := ic.canUseDeltas(srcInfo); ok {
```
lumjjb, line 1149: Interaction for encrypting path with `toEncrypt` flag

lumjjb (commented):
@alexlarsson This is a nice feature to have! I'm trying to think about how this would play with the OCI layer suffixes. My understanding that it is a diff on the tar level, so it will work on compressed/encrypted layers and that will be handled on the client side from the build perspective.

I would imagine that this would work very well with encryption as well... that the deltas themselves could be encrypted as well... So from a user perspective they could use deltas even on encrypted images.

@rhatdan @mtrmac  I think there may be some interactions between the two flows that need to be worked out (possibly a separate PR in the future?). I am wondering if for now we add a couple comments on encryption interactions, and maybe add a few checks to make sure the features are not used together yet. 

I've put in a few comments where I think we could check on some of these interactions.

copy/copy.go
alexlarsson: Deltas are inherently compressed. I.e. they are not wrapped in a compression layer but instead compressed on the parts inside that need it.

alexlarsson: That said, they are always unencrypted, so layering wise we might want to wrap some encryption around them.

copy/copy.go
alexlarsson: canUseDeltas() currently returns false (due to the above media checks) for encrypted manifests, because for now we don't support encrypted deltas. However, once we do these checks need to be widened, yes.

copy/copy.go
alexlarsson: Not sure what you mean here, it seem alright to me. Applying deltas naturally create the uncompressed layers, so we get the diffid etc. That said, I'm not well versed in how crypto works here, so I might be missing something.

lumjjb: Hmm... so the `toEncrypt` flag tells the copy process to encrypt the blob when passed into `copyBlobFromStream`. My concern was if any deltas are created by this process, that they would need to support encryption. If they are not, then it should probably be fine.

alexlarsson: copy never generates deltas, it just applies them to regenerate the tar. Creating deltas is currently done by a separate patch to skopeo.

lumjjb:
@alexlarsson Thanks for the responses!

Agreed on the points and design, just kind of going through the thought exercise.. In general, it looks like there are no design conflicts. I am fine with maybe having a comment or line in a docs to make this more apparent that encrypted layers are not currently supported with deltas. And once this is in, I can open up an issue to keep track of some of the TODOs for support once both technologies become mature and there is an ask.

harche:
@alexlarsson I wasn't able to test https://github.com/alexlarsson/skopeo/tree/wip/deltas using local registry.

```bash
 ./skopeo generate-delta  docker://localhost:5000/nginx:1.17.7 docker://localhost:5000/nginx:1.17.8
FATA[0000] Error creating image: error pinging docker registry localhost:5000: Get "https://localhost:5000/v2/": http: server gave HTTP response to HTTPS client 
```

Just like skopeo `copy` command maybe `generate-delta` should also support `--dest-tls-verify` and `--src-tls-verify` flags. Right now it fails with following error,

```bash
$ ./skopeo generate-delta  --src-tls-verify=false docker://localhost:5000/nginx:1.17.7 docker://localhost:5000/nginx:1.17.8
FATA[0000] unknown flag: --src-tls-verify               
[harshal@localhost skopeo]$ ./skopeo generate-delta  --dest-tls-verify=false docker://localhost:5000/nginx:1.17.7 docker://localhost:5000/nginx:1.17.8
FATA[0000] unknown flag: --dest-tls-verify    
```

alexlarsson:
@harche Added a commit to the skopeo branch that does this.

alexlarsson:
I'm off on PTO from some time next week, so might not see comments here.

alexlarsson:
I noticed that containers 1.4 is out with stargz support. Would have been cool to get delta support in before that...

tpopela:
Could this have some attention please? @mtrmac

cgwalters:
This came up again when discussing https://github.com/ostreedev/ostree-rs-ext/issues/69

I'll just write this here briefly; there are 3 distinct things which make sense separately, *and* can be combined as they're conceptually orthogonal:

- Generating images with "split reproducible blobs"; the advantage of this is it works with the existing ecosystem today.  The main disadvantage is it's hard to impossible to do this from `Dockerfile`, but that can be fixed.
- Dynamic/on-demand fetching: This is what the [zstd:chunked](https://github.com/containers/image/pull/1084) work is about.  The advantage of this is that the delta computation is dynamic, and the zstd issue aside, it will be easy for registries and operators to use this.  (Though, the number of HTTP requests issue is outstanding)
- Pre-computed deltas (bsdiff/rsync/etc).  That's what this PR is about.  Alex's work here is basically hardcoded to bsdiff, but IMO an advantage of the ostree design is it supports "rsync style" rollsum deltas which cover a fair number of binary cases too, and are *much* less CPU intensive.

The way I'd say this is that "split reproducible blobs" aids both "zstd dynamic" and "pre-computed deltas" - it means less work for deltas.  "zstd dynamic" and "pre-computed deltas" more heavily overlap, but they have somewhat opposite advantages/disadvantages, so it depends on the situation.

hdonnay:
Stumbled on this, and it occurs to me that this may be easier to do using the diff blobs as artifacts, now that OCI referrer/"artifact" support exists (or is very close to existing).

cgwalters:
@hdonnay yes see https://github.com/flatpak/flatpak-oci-specs/blob/main/image-deltas.md#locating-the-delta-manifest

hdonnay:
Ah, rad. I was briefly concerned for how people would expect this to interact with Clair, but if the invariant that "a client can always pull the original blob" is maintained, it shouldn't be an issue.

cgwalters:
To make something a bit more obvious here, since this PR is still linked in a variety of places: What ended up happening here is the flatpak developers took similar code/logic into flatpak's implementation of OCI:

- standard: https://github.com/flatpak/flatpak-oci-specs/blob/main/image-deltas.md
- code: https://github.com/flatpak/flatpak/blob/fa252b865a9305a5ff01564140a7273f05503ece/common/flatpak-oci-registry.c#L1885 (etc)

---

On a still related, but different topic, see also recent discussion in https://github.com/ostreedev/ostree/issues/3388 around chromium's zucchini.

jankaluza:

Issue #733, opened 2019-10-22
wrong mediatypes when blob is already present at destination


vrothberg:
When copying an image, the media type of a given blob can be wrong when the a blob with the same digest is already present at the destination. Ultimately, `TryReusingBlob` and semantics behind (including the caches) only care about digests but they do not take the media types into consideration.

I think we need to lift setting media types on a higher level and not embed that in each individual `BlobInfo`. Currently, the API doesn't allow for specifying the compression on the granularity of layers but only for the entire image. That's something we could (should?) make use of when updating the manifest.

vrothberg:
@rhatdan @mrunalp FYI

rhatdan:
SGTM

mtrmac:
> When copying an image, the media type of a given blob can be wrong when the a blob with the same digest is already present at the destination. Ultimately, `TryReusingBlob` and semantics behind (including the caches) only care about digests but they do not take the media types into consideration.

Well, `TryReusingBlob` _does_ return a `types.BlobInfo` and could return a valid MIME type. The difficulty is in knowing the right media type, https://github.com/docker/distribution/blob/master/docs/spec/api.md#get-blob suggests that `Content-Type: application/octet-stream` is OK and the MIME type is only carried in manifests. So, I guess we would have to always pass in the expected MIME type into `TryReusingBlob`, _and_ modify the blob info cache to record MIME types.

> I think we need to lift setting media types on a higher level and not embed that in each individual `BlobInfo`. Currently, the API doesn't allow for specifying the compression on the granularity of layers but only for the entire image. That's something we could (should?) make use of when updating the manifest.

I can’t quite see how this relates to the issue, it seems to go exactly the opposite way. It could _easily_ be the case that a single image (created as a sequence of build+push steps) can contain differently-compressed individual layers, so moving this from `BlobInfo` to “a higher level” does not make sense to me … I probably don’t understand what “a higher level” means.

---

We _really need to_ move `ImageDestination` and a fair amount of the copy infrastructure out of the stable API (`containers/image/types/unstable.ImageDestination`? or an actual embedded Go module with v0 version?), or we are going to keep bumping the major version all the time.

vrothberg:
> We _really need to_ move `ImageDestination` and a fair amount of the copy infrastructure out of the stable API (`containers/image/types/unstable.ImageDestination`? or an actual embedded Go module with v0 version?), or we are going to keep bumping the major version all the time.

[...]

PR #904, opened 2020-04-23
Improve handling of Variant: ""

### LINKED ISSUE #898: Arm64: "no image found" error on arm64

mtrmac:
- When a variant is autodetected (or specified), always try a `Variant: ""` image as well – and only as the last attempt. This is necessary for images like `docker://k8s.gcr.io/pause:3.1` .

  NOTE: This includes the case of `VariantChoice` with a completely unrecognized variant, even in that case we try `Variant: ""`.

- When a variant is not specified with `OverrideArch`, or perhaps not autodetected, try an image with a "base variant" as well. This is necessary for images like `docker://busybox:1.27.2` . If the available images contain _both_ a `Variant: ""` and a `Variant: baseVariant` image, the two are equivalent in priority and the outcome is ambiguous; in that case, prefer the `Variant: ""` image to more closely match the `VariantChoice` made; the caller can always explicitly use `VariantChoice: baseVariant` to get the other one (which would not be the case if we preferred `Variant: baseVariant`).

Fixes #898.

### DIFF
--- a/internal/pkg/platform/platform_matcher.go
+++ b/internal/pkg/platform/platform_matcher.go
@@ -119,7 +119,7 @@ func getCPUVariant(os string, arch string) string {
 // order from most capable (most restrictive) to least capable (most compatible).
 // Architectures that don’t have variants should not have an entry here.
 var compatibility = map[string][]string{
-	"arm":   {"v7", "v6", "v5"},
+	"arm":   {"v8", "v7", "v6", "v5"},
 	"arm64": {"v8"},
 }
 
--- a/internal/pkg/platform/platform_matcher.go
+++ b/internal/pkg/platform/platform_matcher.go
@@ -191,17 +191,10 @@ func WantedPlatforms(ctx *types.SystemContext) ([]imgspecv1.Platform, error) {
 	return res, nil
 }
 
+// MatchesPlatform returns true if a platform descriptor from a multi-arch image matches
+// an item from the return value of WantedPlatforms.
 func MatchesPlatform(image imgspecv1.Platform, wanted imgspecv1.Platform) bool {
-	if image.Architecture != wanted.Architecture {
-		return false
-	}
-	if image.OS != wanted.OS {
-		return false
-	}
-
-	if image.Variant == wanted.Variant {
-		return true
-	}
-
-	return false
+	return image.Architecture == wanted.Architecture &&
+		image.OS == wanted.OS &&
+		image.Variant == wanted.Variant
 }
--- a/internal/pkg/platform/platform_matcher.go
+++ b/internal/pkg/platform/platform_matcher.go
@@ -115,9 +115,20 @@ func getCPUVariant(os string, arch string) string {
 	return ""
 }
 
+// compatibility contains, for a specified architecture, a list of known variants, in the
+// order from most capable (most restrictive) to least capable (most compatible).
+// Architectures that don’t have variants should not have an entry here.
 var compatibility = map[string][]string{
 	"arm":   {"v7", "v6", "v5"},
 	"arm64": {"v8"},
+}
+
+// baseVariants contains, for a specified architecture, a variant that is known to be
+// supported by _all_ machines using that architecture.
+// Architectures that don’t have variants, or where there are possible versions without
+// an established variant name, should not have an entry here.
+var baseVariants = map[string]string{
+	"arm64": "v8",
 }
 
 // WantedPlatforms returns all compatible platforms with the platform specifics possibly overriden by user,
@@ -146,18 +157,27 @@ func WantedPlatforms(ctx *types.SystemContext) ([]imgspecv1.Platform, error) {
 	}
 
 	var variants []string = nil
-	if wantedVariant != "" && compatibility[wantedArch] != nil {
-		variantOrder := compatibility[wantedArch]
-		for i, v := range variantOrder {
-			if wantedVariant == v {
-				variants = variantOrder[i:]
-				break
+	if wantedVariant != "" {
+		if compatibility[wantedArch] != nil {
+			variantOrder := compatibility[wantedArch]
+			for i, v := range variantOrder {
+				if wantedVariant == v {
+					variants = variantOrder[i:]
+					break
+				}
 			}
 		}
-	}
-	if variants == nil {
-		// user wants a variant which we know nothing about - not even compatibility
-		variants = []string{wantedVariant}
+		if variants == nil {
+			// user wants a variant which we know nothing about - not even compatibility
+			variants = []string{wantedVariant}
+		}
+		variants = append(variants, "")
+	} else {
+		variants = append(variants, "") // No variant specified, use a “no variant specified” image if present
+		if baseVariant, ok := baseVariants[wantedArch]; ok {
+			// But also accept an image with the “base” variant for the architecture, if it exists.
+			variants = append(variants, baseVariant)
+		}
 	}
 
 	res := make([]imgspecv1.Platform, 0, len(variants))
@@ -179,7 +199,7 @@ func MatchesPlatform(image imgspecv1.Platform, wanted imgspecv1.Platform) bool {
 		return false
 	}
 
-	if wanted.Variant == "" || image.Variant == wanted.Variant {
+	if image.Variant == wanted.Variant {
 		return true
 	}
 
--- a/image/fixtures/schema2list-variants.json
+++ b/image/fixtures/schema2list-variants.json
@@ -1,0 +1,44 @@
+{
+   "schemaVersion": 2,
+   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
+   "manifests": [
+      {
+         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
+         "size": 527,
+         "digest": "sha256:59eec8837a4d942cc19a52b8c09ea75121acc38114a2c68b98983ce9356b8610",
+         "platform": {
+            "architecture": "amd64",
+            "os": "linux"
+         }
+      },
+      {
+         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
+         "size": 527,
+         "digest": "sha256:f365626a556e58189fc21d099fc64603db0f440bff07f77c740989515c544a39",
+         "platform": {
+            "architecture": "arm",
+            "variant": "v6",
+            "os": "linux"
+         }
+      },
+      {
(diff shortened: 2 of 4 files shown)


mtrmac:
@vrothberg @saschagrunert @Pennyzct @everflux @pjoomen RFC, this works for the two known test images around but I’m not 100% sure it is correct – especially the “unknown variant” case.

internal/pkg/platform/platform_matcher.go
```diff
@@ -115,11 +115,22 @@ func getCPUVariant(os string, arch string) string {
...
 	"arm":   {"v7", "v6", "v5"},
 	"arm64": {"v8"},
 }
 
+// baseVariants contains, for a specified architecture, a variant that is known to be
+// supported by _all_ machines using that architecture.
+// Architectures that don’t have variants, or where there are possible versions without
+// an established variant name, should not have an entry here.
+var baseVariants = map[string]string{
+	"arm64": "v8",
```
saschagrunert, line 131: arm/v7 does not run all 32 bit arm variants?

internal/pkg/platform/platform_matcher.go
```diff
@@ -115,11 +115,22 @@ func getCPUVariant(os string, arch string) string {
 	return ""
 }
 
+// compatibility contains, for a specified architecture, a list of known variants, in the
+// order from most capable (most restrictive) to least capable (most compatible).
+// Architectures that don’t have variants should not have an entry here.
 var compatibility = map[string][]string{
 	"arm":   {"v7", "v6", "v5"},
 	"arm64": {"v8"},
```
saschagrunert, line 123: I think arm64 runs arm as well, right?

internal/pkg/platform/platform_matcher.go
mtrmac: `baseVariants` answers a different question: “when we want an image for ARM/unknown, which variant is certainly acceptable?” Assuming v5 is a subset of v7, arm/v5 would be a better answer to that question than v7. I’m not sure whether v5 is a safe default — looking at `getCPUVariantArm`, values up to v3 seem to exist.

Hum, looking at docker://busybox:1.27.2 , we might need a `baseVariants` entry for ARM — but I really don’t know whether v5 is OK.

I’d be happy to hear from experts.

internal/pkg/platform/platform_matcher.go
mtrmac: From what I hear, that is CPU-dependent, not an architecture requirement.

Vogtinator: Correct.

Though there is "arm/v8" as well - the ARMv8 32bit instructions.

internal/pkg/platform/platform_matcher.go
Vogtinator: I would say that it should take the lowest variant (most likely to work), but prefer a matching "arm/unkown" image if present (perfect match).

Or if it's not too incompatible, erroring out would probably also work, as except for the case where an "arm/unknown" image is requested explicitly (e.g. for mirroring) that shouldn't be happening at all.

mtrmac: > I would say that it should take the lowest variant (most likely to work), but prefer a matching "arm/unkown" image if present (perfect match).

That’s what the code does, but which is the lowest variant for 32-bit ARM (on container-capable Linux systems, I guess)?

Vogtinator: The Linux kernel currently supports anything down to v4 AFAICT, though I've never seen container images for that...

mtrmac: This is all for the “non-ARM executable wants an ARM image” case, so I think at least for now (while we need to urgently fix CRI-O) it might be good enough to require the caller to explicitly specify a variant.

internal/pkg/platform/platform_matcher.go
pjoomen: `arm/v8` is the one referenced in #3626, 64 bits kernel on 32 bits userland.

mtrmac: https://github.com/cri-o/cri-o/issues/3626 should work with this PR (arm/v8 will now use the image for arm/""); but with the list above, arm/v8 would not use a arm/v7 image.

I can add v8 to the 32-bit ARM list above in this PR as well (causing the arm/v8 compatibility list to be arm/{v8,v7,v6,v5,""} ), if there is consensus on that.

Vogtinator: Yes. ARMv8 defines both a 32bit ISA (superset of ARMv7) and a 64bit ISA.

Some CPUs only support one of them, so `arm/v8` has to be treated differently from `arm64/v8`.

mtrmac: Thanks, added.

rhatdan:
@saschagrunert @vrothberg @Vogtinator  PTAL

Pennyzct:
Hi @mtrmac 

Thanks for fixing https://github.com/containers/image/issues/898, Could we backport it or make it in a stable version, then cri-o could use it?? Since for now we could not launch any cri-o/k8s v1.17.1 ~ v1.18.x on AArch64 cluster.

vrothberg:
> @vrothberg @rhatdan WDYT? I wasn’t present when the fine details of the branching design were finalized.

There are no branches yet. The format will be following CRI-O's example with `release-$major-$minor`.

CRI-O 1.18 is on 5.4.x while 1.17 is on 5.3.x. I believe we can bump both to the latest 5.4.x to get the fixes there.

I think it's a good time to cut a `release-5.4` branch. All tools are using that branch.

> There is a set of fairly important bug fixes pending in c/image, but then there is also #733 that should be fixed quickly.

I agree that it's important but given the age of it, I wouldn't like it to block. Once there is a fix, we can worry about backporting if needed.

mtrmac:
OK, preparing 5.4.4 in #921.

PR #906, opened 2020-04-24
registries.conf: remove references on v1

vrothberg:
Version 2 of the registries.conf has been used long enough in the wild
to consider it stable.  Hence, remove the references in the
registries.conf file to the previous version in hope to increase v2
adoption.

### DIFF
--- a/registries.conf
+++ b/registries.conf
@@ -1,57 +1,22 @@
 # For more information on this configuration file, see containers-registries.conf(5).
 #
-# There are multiple versions of the configuration syntax available, where the
-# second iteration is backwards compatible to the first one. Mixing up both
-# formats will result in an runtime error.
-#
-# The initial configuration format looks like this:
-#
-# Registries to search for images that are not fully-qualified.
-# i.e. foobar.com/my_image:latest vs my_image:latest
-#
 # NOTE: RISK OF USING UNQUALIFIED IMAGE NAMES
 # We recommend always using fully qualified image names including the registry
 # server (full dns name), namespace, image name, and tag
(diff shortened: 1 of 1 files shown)


registries.conf
```diff
@@ -25,33 +16,6 @@
 # If it is necessary to use one of these registries, it should be added at the
 # end of the list.
 #
-# It is recommended to use fully-qualified images for pulling as the
-# destination registry is unambiguous. Pulling by digest
-# (i.e., quay.io/repository/name@digest) further eliminates the ambiguity of
-# tags.
```
mtrmac, line 31: Shouldn’t this paragraph stay?

TomSweeneyRedHat: Although after expanding the verbiage that GitHub auto hid, I think it's OK to remove this as it's a bit redundant.

vrothberg: I squeezed it in the initial note.

vrothberg:
> `docs/containers-registries.conf.5.md` should probably be updated as well.

@mtrmac, what do you have in mind?

rhatdan:
This LGTM, I think we could eliminate mention of V1 going forward.

mtrmac:
> > `docs/containers-registries.conf.5.md` should probably be updated as well.
> 
> @mtrmac, what do you have in mind?

At the very start there is at least
> ## VERSION 2
> VERSION 2 is the latest format of the `registries.conf` and is currently in beta. This means in general VERSION 1 should be used in production environments for now.

and the rest might need updating as well.

rhatdan:
Ok lets fix the man page in a separate PR>

PR #909, opened 2020-04-24
Simplify MIME type updates

mtrmac:
In preparation to fixing #733, this simplifies MIME type updates in manifests by factoring out the code common to all cases, and by making the mapping itself table-driven instead of a large set of manually written switch statements.

Should not change behavior, apart from some error strings.

### DIFF
--- a/manifest/manifest.go
+++ b/manifest/manifest.go
@@ -256,13 +256,3 @@ func FromBlob(manblob []byte, mt string) (Manifest, error) {
 	// Note that this may not be reachable, NormalizedMIMEType has a default for unknown values.
 	return nil, fmt.Errorf("Unimplemented manifest MIME type %s (normalized as %s)", mt, nmt)
 }
-
-// layerInfosToStrings converts a list of layer infos, presumably obtained from a Manifest.LayerInfos()
-// method call, into a format suitable for inclusion in a types.ImageInspectInfo structure.
-func layerInfosToStrings(infos []LayerInfo) []string {
-	layers := make([]string, len(infos))
-	for i, info := range infos {
-		layers[i] = info.Digest.String()
-	}
-	return layers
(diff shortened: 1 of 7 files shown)


manifest/manifest.go
```diff
@@ -266,3 +268,45 @@ func layerInfosToStrings(infos []LayerInfo) []string {
 	}
 	return layers
 }
+
+// compressionMIMETypeSet describes a set of MIME type “variants” that represent differently-compressed
+// versions of “the same kind of content”.
+// The map key is the return value of compression.Algorithm.Name(), or "" to mean "uncompressed";
+// the map value is a MIME type, or "" to mean "recognized but unsupported".
```
vrothberg, line 275: Non-blocking nit: I'd prefer to make the "uncompressed" and "recognized but unsupported case" more explicit and make them consts. I was initially confused that the same value (i.e., "") can have very different meanings for keys and values.

It's well documented here but would make reading code using it more explicit.

mtrmac: That’s a good point. Would using the same `""` value but named (`mtsUncompressed`? `mtsUnsupportedMIMEType`?) work? “Reserving” a real name worries me about conflicts with actual MIME type values.

(/me longingly looks at Swift’s variant types and enum type name inference.)

mtrmac:
Updated to
- add named constants for the uncompressed/unsupported cases
- move various helpers shared across formats into `manifest/common.go`
- make the MIME type determination code even more shared across schema2/OCI.

manifest/manifest.go
vrothberg: Yes, that's what I meant. Using an empty string is fine. The consts are meant to only improve readability.

rhatdan:
LGTM

PR #914, opened 2020-04-29
Remove registries.conf VERSION 1 references from man page

rhatdan:
We don't want users using VERSION 1 any longer.  So remove all references
to depracated configuration.

### DIFF
--- a/docs/containers-registries.conf.5.md
+++ b/docs/containers-registries.conf.5.md
@@ -11,13 +11,6 @@ file for container image registries. The file format is TOML.
 
 Container engines will use the `$HOME/.config/containers/registries.conf` if it exists, otherwise they will use `/etc/containers/registries.conf`
 
-# FORMATS
-
-## VERSION 2
-VERSION 2 is the latest format of the `registries.conf` and is currently in
-beta. This means in general VERSION 1 should be used in production environments
-for now.
-
 ### GLOBAL SETTINGS
 
 `unqualified-search-registries`
@@ -154,8 +147,8 @@ Given the above, a pull of `example.com/foo/image:latest` will try:
 
 in order, and use the first one that exists.
 
-## VERSION 1
-VERSION 1 can be used as alternative to the VERSION 2, but it does not support
+## VERSION 1 FORMAT - DEPRECATED
+VERSION 1 format is still supported but it does not support
 using registry mirrors, longest-prefix matches, or location rewriting.
 
 The TOML format is used to build a simple list of registries under three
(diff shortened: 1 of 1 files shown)


rhatdan:
@mtrmac @vrothberg PTAL

rhatdan:
I disagree, I believe this just causes confusion.  Will end up 5 years from now, still carrying around this baggage.  We plan on updating the registries.conf going forward, to the new format, the only way that someone would see the old format would be if they had modified the old format, and it is pretty easy to convert.  We would not want them to stick to the old format. 

I guess the argument for the old format sticking around would be if someone came to containers-image looking for documentation and only saw the new format, that their containers/image based code did not support.

mtrmac:
> I disagree, I believe this just causes confusion. Will end up 5 years from now, still carrying around this baggage.

As long as we carry around the code that processes that format, sure.

> I guess the argument for the old format sticking around would be if someone came to containers-image looking for documentation and only saw the new format, that their containers/image based code did not support.

Yes. It’s perfectly fine to drop the old format from the sample file, from blogs, tutorials and how-to documentation; all of that can be opinionated and show the best way to do things. But the reference documentation, which is this man page, would be the place to go to see _all_ the formats and features that are supported.

docs/containers-registries.conf.5.md
```diff
@@ -154,39 +149,6 @@ Given the above, a pull of `example.com/foo/image:latest` will try:
 
 in order, and use the first one that exists.
 
-## VERSION 1
```
TomSweeneyRedHat, line 157: I'd keep this but change to:

```
## Version 1- To be DEPRECATED

**NOTE:** For backward compatibility reasons, Version 1 is still usable but will be deprecated in a release in the near future.   The Version 1 format should be converted to the Version 2 format at the earliest opportunity.
```

TomSweeneyRedHat:
If I get a vote, I'd keep v1 but warn.  Then we can remove it later with prejudice.  My $.02.

vrothberg:
I feel strongly to keep v1 references. New users won't care about it as they will get a v2 config but existing users will get updated man pages while still having a v1 registries.conf.

A couple of weeks ago, someone opened an issue on libpod as they wanted to migrate from v1 to v2. Keeping this information was essential.

> I guess the argument for the old format sticking around would be if someone came to containers-image looking for documentation and only saw the new format, that their containers/image based code did not support.

When updating, existing users will get a new podman and new man pages but their configs will likely remain untouched. We don't help these users when removing v1 entirely from the docs.

docs/containers-registries.conf.5.md
```diff
@@ -13,11 +13,6 @@ Container engines will use the `$HOME/.config/containers/registries.conf` if it
 
 # FORMATS
```
mtrmac, line 14: This heading now seems somewhat disconnected from the contents — maybe drop it?

PR #919, opened 2020-05-05
Add not about policy.json file in homedir

rhatdan:
We are now reading ~/.config/containers/policy.json if it exists
in the users homedir, when running rootless.  We should document
in the man page that this file is also used.

Signed-off-by: Daniel J Walsh <dwalsh@redhat.com>

PR #922, opened 2020-05-12
Add defaults for using the rootless policy path

QiWang19:
Add support for default rootless policy path of $HOME/.config/containers/policy.json

### DIFF
--- a/docs/containers-policy.json.5.md
+++ b/docs/containers-policy.json.5.md
@@ -10,8 +10,7 @@ containers-policy.json - syntax for the signature verification policy file
 Signature verification policy files are used to specify policy, e.g. trusted keys,
 applicable when deciding whether to accept an image, or individual signatures of that image, as valid.
 
-The default policy is stored (unless overridden at compile-time) at `/etc/containers/policy.json`;
-applications performing verification may allow using a different policy instead.
+By default, the policy is read from `$HOME/.config/containers/policy.json`, if it exists, otherwise from `/etc/containers/policy.json`;  applications performing verification may allow using a different policy instead.
 
 ## FORMAT
 
--- a/signature/policy_config.go
+++ b/signature/policy_config.go
@@ -17,11 +17,13 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"os"
 	"path/filepath"
 
 	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/image/v5/transports"
 	"github.com/containers/image/v5/types"
+	"github.com/containers/storage/pkg/homedir"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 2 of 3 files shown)


QiWang19:
replace https://github.com/containers/image/pull/919

QiWang19:
@mtrmac PTAL

docs/containers-policy.json.5.md
```diff
@@ -10,8 +10,7 @@ containers-policy.json - syntax for the signature verification policy file
 Signature verification policy files are used to specify policy, e.g. trusted keys,
 applicable when deciding whether to accept an image, or individual signatures of that image, as valid.
 
-The default policy is stored (unless overridden at compile-time) at `/etc/containers/policy.json`;
-applications performing verification may allow using a different policy instead.
+The default policy is stored (unless overridden at compile-time) at `/etc/containers/policy.json`;  applications performing verification may allow using a different policy instead. Rootless applications may also read the default policy from `$HOME/.config/containers/policy.json` if it exists.
```
TomSweeneyRedHat, line 13: I think I'd change "may also read" -> "will read"

mtrmac: “rootless” is a Podman jargon, and anyway this does not accurately describe what the code does.

QiWang19: how about mention `$HOME/.config/containers/policy.json` if it exists. Otherwise, will use the `/etc/containers/policy.json`?

rhatdan: Yes from containers image point of view, we should read the file whether we are root or rootless.
IE if we are root we should look for /root/.config/contianers/policy.json
So we don't really need to differentiate.

QiWang19: PTAL , I updated the containers-policy.json.5.md

signature/policy_config.go
```diff
@@ -53,13 +58,15 @@ func DefaultPolicy(sys *types.SystemContext) (*Policy, error) {
...
-			return filepath.Join(sys.RootForImplicitAbsolutePaths, systemDefaultPolicyPath)
-		}
+	if sys != nil && sys.SignaturePolicyPath != "" {
+		return sys.SignaturePolicyPath
+	}
+	userPolicyFilePath := filepath.Join(homedir.Get(), userPolicyFile)
+	if _, err := os.Stat(userPolicyFilePath); err == nil {
+		return userPolicyFilePath
+	}
+	if sys != nil && sys.RootForImplicitAbsolutePaths != "" {
```
rhatdan, line 68: @mtrmac how should we handle this path versus RootForImplicitAbsolutePaths?

rhatdan, line 68: @mtrmac how should we handle this path versus RootForImplicitAbsolutePaths?

PR #201, opened 2017-01-07
WIP do not merge: Dump HTTP operations

mtrmac:
Quite often it has been useful to dump the full HTTP request and response headers, and sometimes fuill bodies, for debugging; we probably should make that possible either automatically at the “debug” log level, or when the user sets an option.

This is *by no means* a workable implementation for that, but I’ve written exactly this hack two or three times now, so right now I am at least recording it for posterity, expecting that rebasing it will be easier than rewriting it.

A real implementation should, I guess, wrap a `http.RoundTripper` or perhaps `http.Client` (or just `http.Client.Do`?), or perhaps using `net/http/httptrace` somehow.

PR #924, opened 2020-05-13
Add debug line to get Content-Type from manifests

robbmanes:
Often, manifests fail to pull because of a MIME type mismatch and have
to be debugged manually to figure out what the server/proxy is sending
the client.  This is much easier to see when it's just printed for you.

Signed-off-by: Robb Manes <rmanes@redhat.com>

### DIFF
--- a/docker/docker_image_src.go
+++ b/docker/docker_image_src.go
@@ -190,6 +190,7 @@ func (s *dockerImageSource) fetchManifest(ctx context.Context, tagOrDigest strin
 	if err != nil {
 		return nil, "", err
 	}
+	logrus.Debugf("Content-Type from manifest GET is %q", res.Header.Get("Content-Type"))
 	defer res.Body.Close()
 	if res.StatusCode != http.StatusOK {
 		return nil, "", errors.Wrapf(client.HandleErrorResponse(res), "Error reading manifest %s in %s", tagOrDigest, s.physicalRef.ref.Name())


robbmanes:
Example bug where this or something like it would have been useful is [Red Hat Bugzilla 1795334](https://bugzilla.redhat.com/show_bug.cgi?id=1795334).  In essence we have a lot of situations where a proxy or something in-between is using a `Content-Type` of `prettyjws` or something else that isn't `json` on the manifest.  I debated between doing this and making a method that would dump the entire header and take the `res` from a `makeRequest` call, which we could use at various places; if you'd prefer I'd do that, I can do so, but I figured this would be easier and still directly address the number of issues we've seen with incorrect `Content-Type`'s.

docker/docker_image_src.go
```diff
@@ -187,6 +187,7 @@ func (s *dockerImageSource) fetchManifest(ctx context.Context, tagOrDigest strin
 		"Accept": manifest.DefaultRequestedManifestMIMETypes,
 	}
 	res, err := s.c.makeRequest(ctx, "GET", path, headers, nil, v2Auth, nil)
+	logrus.Debugf("Content-Type from GET to %s is %s", path, res.Header.Get("Content-Type"))
```
TomSweeneyRedHat, line 190: tests aren't happy.  I'm guessing you might want to add a nil check on ref before derefencing?

rhatdan: Yes this message should only be set on success.

robbmanes: Whoops, you're right, let me fix that.

robbmanes: Fixed to set the log only after we check success.

robbmanes: Checks fixed, terribly sorry about that, rookie mistake.

rhatdan:
LGTM
@mtrmac PTAL

docker/docker_image_src.go
```diff
@@ -190,6 +190,7 @@ func (s *dockerImageSource) fetchManifest(ctx context.Context, tagOrDigest strin
 	if err != nil {
 		return nil, "", err
 	}
+	logrus.Debugf("Content-Type from GET to %s is %s", path, res.Header.Get("Content-Type"))
```
mtrmac, line 193: Please use `%q` for the header value, it can be empty or contain unexpected garbage.

(I’m not sure including `path` is all that valuable, especially because it is not a full URL; the full URL is logged anyway in `makeRequestToResolvedURLOnce`. But it does not exactly _hurt_.)

mtrmac:
> I debated between doing this and making a method that would dump the entire header and take the `res` from a `makeRequest` call, which we could use at various places

(FWIW there’s a very rough version of that at #201 , but it’s not been exposed in a user-consumable way.)

robbmanes:
> (FWIW there’s a very rough version of that at #201 , but it’s not been exposed in a user-consumable way.)

That's brilliant and helps us out immensely if/when it is merged; I just chose the low-hanging fruit for what we see as the most common issue in `skopeo` and `podman` in this immediate moment.

mtrmac:
Sure, I think this PR as is is a great idea and would have helped debugging quite a few problem reports. #201 is by no means a replacement for this.

TomSweeneyRedHat:
LGTM once @mtrmac concern is addressed and the PR is rebased.

docker/docker_image_src.go
robbmanes: Would then just specifying "Content-Type from manifest GET" then work?  I only included the path because it'd be clear it was a GET on the manifest itself as per the path.  I'll fix just this one piece and if you want me to drop the path I'm happy to do so.

mtrmac:
```console
$ skopeo linux inspect --debug docker://registry.fedoraproject.org/fedora                                                                                            
…
DEBU[0001] Ping https://registry.fedoraproject.org/v2/ status 200 
DEBU[0001] GET https://registry.fedoraproject.org/v2/fedora/manifests/latest 
DEBU[0001] GET https://registry.fedoraproject.org/v2/fedora/manifests/sha256:119d2ca164232e4e46ee07d177504d8b98172a20ffad1c11b2433a5df9494511 
DEBU[0002] Downloading /v2/fedora/blobs/sha256:d81c91deec0dc09a3f05db3f538bdba9c49f1d51e6955078f274cd4ed4376c6f 
…
```
Yes, I think adding just a `Content-Type from manifest GET` without reiterating the path would have sufficient context, although it is not trivially visible from just this part of source code.

robbmanes: Should be addressed.  Thanks!

PR #925, opened 2020-05-13
Support credsStore config

QiWang19:
Support the "credsStore" in the auth.json file. If a credential helper configed, will save the authentication there.
Updated the containers-auth.json.5.md about the credsStore and credential helpers configuration.

### DIFF
--- a/docs/containers-auth.json.5.md
+++ b/docs/containers-auth.json.5.md
@@ -36,8 +36,26 @@ their accounts on quay.io and docker.io:
 An entry can be removed by using a `logout` command from a container
 tool such as `podman logout` or `buildah logout`.
 
+In addition, credential store and credential helpers can be configured and the credentials-helper
+software can be used to manage the credentials in a more secure way than depending on the base64 encoded authentication
+provided by `login`. If the credential helpers are configured for specific registries, the `credsStore` will not be used
+for operations concerning credentials of the specified registries.
+
+When the credential store is in use on a Linux platform, the auth.json file would contain:
+
+```
+    "auths": {
+        "localhost:5001": {}
+    },
+    "credsStore": "secretservice"
+}
+```
+
+For more information on Docker credential stores, please reference the [GitHub docker-credential-helpers project](https://github.com/docker/docker-credential-helpers/releases).
+
+
 # SEE ALSO
-    buildah-login(1), buildah-logout(1), podman-login(1), podman-logout(1)
+    buildah-login(1), buildah-logout(1), podman-login(1), podman-logout(1), skopeo-login(1), skopeo-logout(1)
 
 # HISTORY
 Feb 2020, Originally compiled by Tom Sweeney <tsweeney@redhat.com>
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -25,6 +25,7 @@ type dockerAuthConfig struct {
 type dockerConfigFile struct {
 	AuthConfigs map[string]dockerAuthConfig `json:"auths"`
 	CredHelpers map[string]string           `json:"credHelpers,omitempty"`
+	CredsStore  string                      `json:"credsStore,omitempty"`
 }
 
 type authPath struct {
@@ -52,6 +53,9 @@ func SetAuthentication(sys *types.SystemContext, registry, username, password st
 	return modifyJSON(sys, func(auths *dockerConfigFile) (bool, error) {
 		if ch, exists := auths.CredHelpers[registry]; exists {
 			return false, setAuthToCredHelper(ch, registry, username, password)
+		}
+		if auths.CredsStore != "" {
+			return false, setAuthToCredHelper(auths.CredsStore, registry, username, password)
 		}
 
 		// Set the credentials to kernel keyring if enableKeyring is true.
@@ -150,6 +154,9 @@ func RemoveAuthentication(sys *types.SystemContext, registry string) error {
 		if ch, exists := auths.CredHelpers[registry]; exists {
 			return false, deleteAuthFromCredHelper(ch, registry)
 		}
+		if auths.CredsStore != "" {
+			return false, deleteAuthFromCredHelper(auths.CredsStore, registry)
+		}
 
 		// Next if keyring is enabled try kernel keyring
 		if enableKeyring {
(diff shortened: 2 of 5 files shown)


docs/containers-auth.json.5.md
```diff
@@ -36,6 +36,24 @@ their accounts on quay.io and docker.io:
 An entry can be removed by using a `logout` command from a container
 tool such as `podman logout` or `buildah logout`.
 
+In addition, a Docker credential store and Credential helpers can be configured and the Docker credentials-helper
```
rhatdan, line 39: Remove "Docker" from the descriptions.

pkg/docker/config/config.go
```diff
@@ -318,6 +342,26 @@ func deleteAuthFromCredHelper(credHelper, registry string) error {
 	return helperclient.Erase(p, registry)
 }
 
+func deleteAuthFromCredsStore(credHelper string) error {
+	creds, err := listAuthFromCredHelper(credHelper)
+	if err != nil {
+		return err
+	}
+	for registry := range creds {
+		err = deleteAuthFromCredHelper(credHelper, registry)
```
rhatdan, line 351: Single line
if err = ...

rhatdan: Should we ignore this ENOEXIST type error?

pkg/docker/config/config.go
```diff
@@ -183,7 +190,22 @@ func RemoveAllAuthentication(sys *types.SystemContext) error {
 			}
 			logrus.Debugf("error removing credentials from kernel keyring")
 		}
+		if len(auths.CredHelpers) != 0 {
+			for cred, registry := range auths.CredHelpers {
+				if err := deleteAuthFromCredHelper(cred, registry); err != nil {
+					logrus.Debugf("error removing credentials for %s from credential helper %s: %q", registry, cred, err)
+					return false, err
```
rhatdan, line 197: Should we ignore registry does not exist errors?  Could we hit a race where two apps are removing the creds at the same time.  Do we want to  save the error and continue deleting other registries.

QiWang19: fix this with ignoring IsErrCredentialsNotFoundMessage.

pkg/docker/config/config.go
QiWang19: I think no. the code before this listAuthFromCredHelper has made sure the authentication of the registry exist

pkg/docker/config/config.go
```diff
@@ -183,7 +190,26 @@ func RemoveAllAuthentication(sys *types.SystemContext) error {
 			}
 			logrus.Debugf("error removing credentials from kernel keyring")
 		}
+		if len(auths.CredHelpers) != 0 {
+			for cred, registry := range auths.CredHelpers {
+				if err := deleteAuthFromCredHelper(cred, registry); err != nil {
+					// ingore the error if no credential stored for the registry
```
mtrmac, line 196:
```suggestion
					// ignore the error if no credential stored for the registry
```

pkg/docker/config/config.go
```diff
@@ -183,7 +190,26 @@ func RemoveAllAuthentication(sys *types.SystemContext) error {
...
+			}
+		}
+		if auths.CredsStore != "" {
+			if err := deleteAuthFromCredsStore(auths.CredsStore); err != nil {
+				logrus.Debugf("error removing credentials from credential helper %s: %q", auths.CredsStore, err)
+				return false, err
+			}
+		}
 		auths.CredHelpers = make(map[string]string)
+		auths.CredsStore = ""
```
mtrmac, line 212: I’m not immediately sure deleting this makes sense — I’ll need to revisit previous discussions.

pkg/docker/config/config.go
```diff
@@ -295,7 +321,9 @@ func getAuthFromCredHelper(credHelper, registry string) (string, string, error)
 	helperName := fmt.Sprintf("docker-credential-%s", credHelper)
 	p := helperclient.NewShellProgramFunc(helperName)
 	creds, err := helperclient.Get(p, registry)
-	if err != nil {
+	// skip the IsErrCredentialsNotFoundMessage error, caller can continue check
+	// the auth from other authfiles
+	if err != nil && !credentials.IsErrCredentialsNotFoundMessage(err.Error()) {
 		return "", "", err
 	}
 	return creds.Username, creds.Secret, nil
```
mtrmac, line 329: This would dereference a `nil` pointer if `err != nil`.

rhatdan:
Miloslav lets have a conversation today at standup for the best way to handle this.  The issue I have is that auth.json is being used both for configuring authorization and storing registry/password.  The question would be could we change this and start storing the default credential-helper in containers.conf rather then in auth.json.

QiWang19:
@vrothberg has suggested setting the configurations in registries.conf since there already supports the prefix. I think that makes sense can discuss in the standup. 

>  The question would be could we change this and start storing the default credential-helper in containers.conf rather then in auth.json.

rhatdan:
That works for me also.  Having them in auth.json is not something we want to support long run.

mtrmac:
> Miloslav lets have a conversation today at standup for the best way to handle this. The issue I have is that auth.json is being used both for configuring authorization and storing registry/password.

It makes sense that the two should be separated, especially if `auth.json` continues to be session-only. OTOH the credential setup seems to me to be much more of a per-user decision than a per-system configuration like `containers.conf` or `registries.conf` (in non-rootless situations where there is a difference).

rhatdan:
My goal is to get to the point of shipping fedora/RHEL with a default for rootless users to use the gnome-keyring.
Perhaps we should setup a search for credential helpers.  Where if gnome-keyring does not work, it falls back to flat files.

QiWang19:
> My goal is to get to the point of shipping fedora/RHEL with a default for rootless users to use the gnome-keyring.
> Perhaps we should setup a search for credential helpers. Where if gnome-keyring does not work, it falls back to flat files.

search for credential helpers and fall back to auth.json, will this be expensive if used by build, pull or other?

rhatdan:
I am willing to put the credential helper the tools want to use in the system context, if that is what @mtrmac wants.  Then we could put it in the containers.conf.  If it goes in the registries.conf then it requires the container simage to use.

If we had two fields rootlessCredentialHelper and CredentialHelper, would that be of interest.

QiWang19:
@rhatdan PTAL some reservations about the containers.conf https://docs.google.com/document/d/1mTPI9GCvy3C_c11Z1MNQzOR6N_lMBVLHW3gpOllC6rw/edit

mtrmac:
> search for credential helpers and fall back to auth.json, will this be expensive if used by build, pull or other?

Reading a few tiny local files that easily fit into cache if used frequently — I would be _very_ surprised if that were noticeable compared to all the network requests and hundreds of megabytes of data.

mtrmac:
> I am willing to put the credential helper the tools want to use in the system context, if that is what @mtrmac wants. Then we could put it in the containers.conf.

I’m not at all a fan of a tool like Podman _always_ configuring non-defaults in `SystemContext`. The default system behavior should, if at all possible, be how c/image behaves by default without the caller having to care.

Per the other discussion, I’ve been corrected that `containers.conf` actually has all the properties I have asked for (per-user configurations that override small parts of system configurations without having to keep large configuration files in sync).

c/common already depends on c/image, so this would create a dependency loop. That seems rather risky.

rhatdan:
I did not plan on putting any type of dependency on c/i on containers/common.  I was stating we could put the default in containers.conf and then have tools like podman, buildah, skopeo use the distro default  while containers/image continues to default to flat files.

We can leave the opinion-ation to the distros rather then hard coding it into containers/image.  Especially where I want the default for rootless users to be gnome-keyring.

mtrmac:
I fully agree that distributions the best place to adjust system-wide defaults (if they are inappropriate for upstreams).

That does not imply callers of c/image should do extra work to make those system-wide default effective; the system-wide defaults would actually be more reliable if they happened automatically in c/image.

PR #929, opened 2020-05-16
pkg/docker/config/ModifyJSON: fix MkdirAll usage

kolyshkin:
1. Using os.Stat() without checking that the result is a directory is
   error-prone: there might be a file with the same name as target dir.

2. There is no need to use os.Stat() before calling os.MkdirAll() as
   the latter does it as well.

3. Reverse the checks, error out early in case of legacyFormat.

Signed-off-by: Kir Kolyshkin <kolyshkin@gmail.com>

### DIFF
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -257,17 +257,15 @@ func modifyJSON(sys *types.SystemContext, editor func(auths *dockerConfigFile) (
 	if err != nil {
 		return err
 	}
-
-	dir := filepath.Dir(path)
-	if _, err := os.Stat(dir); os.IsNotExist(err) {
-		if err = os.MkdirAll(dir, 0700); err != nil {
-			return errors.Wrapf(err, "error creating directory %q", dir)
-		}
-	}
-
 	if legacyFormat {
 		return fmt.Errorf("writes to %s using legacy format are not supported", path)
 	}
+
+	dir := filepath.Dir(path)
+	if err = os.MkdirAll(dir, 0700); err != nil {
+		return err
+	}
+
 	auths, err := readJSONFile(path, false)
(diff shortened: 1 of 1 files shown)


pkg/docker/config/config.go
```diff
@@ -257,17 +257,15 @@ func modifyJSON(sys *types.SystemContext, editor func(auths *dockerConfigFile) (
...
+		return fmt.Errorf("writes to %s using legacy format are not supported", path)
+	}
 
 	dir := filepath.Dir(path)
-	if _, err := os.Stat(dir); os.IsNotExist(err) {
-		if err = os.MkdirAll(dir, 0700); err != nil {
-			return errors.Wrapf(err, "error creating directory %q", dir)
-		}
+	if err = os.MkdirAll(dir, 0700); err != nil {
+		return err
```
TomSweeneyRedHat, line 266: I'd prefer the old error  message.
```
return errors.Wrapf(err, "error creating directory %q", dir)
```

kolyshkin: so, `os.MkdirAll` already has the operation (`mkdir`) and the directory name in its error message (see `os.PathError`), so there's no need to wrap it unless you want to add some higher-level context, which does not seem to be the case here, that is why I (semi-automatically) removed it here.

rhatdan: I don't think we need to wrap.

This test will give
```
err := os.MkdirAll("/test/foo/bar", 0700)
fmt.Println(err)
```
Gives an error like:
```
go run test.go 
mkdir /test: permission denied
```

kolyshkin: ...and forgot to describe it in commit message. Fixed.

PR #933, opened 2020-05-18
sysregistriesv2: add `LocationsForReferences`

vrothberg:
Add a new API to sysregistriesv2 that returns a sorted list of all
registries that may be contacted during image pull.  This data is needed
for Podman's remote API to figure out which credentials the server might
need.


mtrmac:
To expand on this, this PR seems to suggest a Libpod workflow like:
```go
func credentialsToSend(imageName string) []registry {
    if isUnqualified(imageName) {
        return serverAPI.UnqualifiedSearchRegistriesAndMirrors() // may be too much
    else {
        return reference.getRegistryFrom(imageName) // does not include mirrors
    }
}
```

and I was expecting
```go
func credentialsToSend(imageName string) []registry {
    return serverAPI.RegistriesOfPullSourcesFromReference(imageName)
}
```

mtrmac:
Oops, that’s right for qualified but nonsense for unqualified.  This is closer:

client:
```go
func credentialsToSend(imageName) []registry {
    return serverAPI.CredentialsToSend(imageName)
}
```
server:
```go
func (* serverAPI) CredentialsToSend(imageName string) []registry {
    qualifiedNames := (isUnqualified(imageName) 
        ? UnqualifiedSearchRegistries().map(searchRegistry => combine(searchRegistry, imageName)
        : []string{imageName}
    )
    return qualifiedNames
            .map(qualified => FindRegistry(qualified).PullSourcesFromReference(qualified))
            .map(pullSource => reference.getRegistryFrom(pullSource.reference)
}
```

vrothberg:
`UnqualifiedSearchRegistriesAndMirrors` will only be called for unqualified images. But I thought that this detail doesn't belong here. The case for qualified images is easy to handle with the already existing API.

vrothberg:
On second though, it may be worth adding _one_ API here. Makes testing and maintenance easier.

vrothberg:
@mtrmac, I pushed a new flavour which AFAICS accounts for all possible situations:
```LocationsForReferences(ctx *types.SystemContext, references []string) ([]string, error)```

If references are nil or empty, all unqualified-search registries and their mirrors are returned. Otherwise, we iterate over the references:
* Unqualified -> add all unq. registries and their mirrors
* Otherwise, add the domain of the reference and, if configured, it's registry and all mirrors

**Comment by mtrmac** [comment]
> The case for qualified images is easy to handle with the already existing API.

That’s right, of course. I don’t know what I was thinking.

**Comment by vrothberg** [comment]
Tests added and ready to merge from my side.

File: pkg/sysregistriesv2/system_registries_v2.go [hunk 1]
```diff
@@ -519,6 +519,75 @@ func GetRegistries(ctx *types.SystemContext) ([]Registry, error) {
...
+
+	// If needed, add all unqualified-search registries and their mirrors.
+	if needUnqualified {
+		config, err := getConfig(ctx)
+		if err != nil {
+			return nil, err
+		}
+		for _, searchReg := range config.UnqualifiedSearchRegistries {
+			// Add the search registry.
+			addLocation(searchReg)
```
mtrmac, line 571: This also seems unnecessary, see the other call.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -519,6 +519,75 @@ func GetRegistries(ctx *types.SystemContext) ([]Registry, error) {
...
+			for _, reg := range config.Registries {
+				if !refMatchesPrefix(searchReg, reg.Prefix) {
```
mtrmac, line 573: Isn’t this just `FindRegistry(searchReg)`? That doesn’t account for mirrors of namespaces within `searchReg`.

So, this should combine `searchReg` with (each member of) `references` — and c/image does not currently provide any API to qualify unqualified names, I don’t quite want to make it a part of the core API (it should be some opt-in “Podman-like name handling” subpackage), but if we don’t, there’s a risk of the c/image implementation and the Podman implementation getting out of sync.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -519,6 +519,75 @@ func GetRegistries(ctx *types.SystemContext) ([]Registry, error) {
 	return config.Registries, nil
 }
 
+// LocationsForReferences returns a sorted list of locations (i.e., registries)
+// that may be contacted when pulling the specified references.  If the
+// references are empty, all unqualified-search registries and their mirrors
+// will be returned.
+func LocationsForReferences(ctx *types.SystemContext, references []string) ([]string, error) {
+	locations := make(map[string]bool)
+	needUnqualified := len(references) == 0
```
mtrmac, line 528: Isn’t this just a caller error? Is it necessary to provide this? (The use case seems to be some version of “We may use unqualified names but don’t know which ones”, but that doesn’t work for mirrors of individual namespaces on the search registries.)

vrothberg:
> Overall, I don’t think this is necessary to provide in c/image: Podman can combine its existing “get a list of fully qualified references from an unqualified reference” implementation” with `FindRegistry().PullSourcesFromReference()`, like [#933 (comment)](https://github.com/containers/image/pull/933#issuecomment-630328756) .
> 
> That would (for now) keep c/image out of the “interpreting unqualified image names” business.

Sounds good to me. I'll move back to just listing unqualified registries and their mirrors and move the remaining logic into `libpod` :+1:

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: We can't use `FindRegistry(...)` as we need _all_ matching registries (and mirrors) and not only the one with the longest matching prefix. I'll add more verbose comments after changing the API.

mtrmac: I thought that we need, for all search registries, the mirrors that apply for qualifying the input name with that registry = the mirrors found by `FindRegistry(combinedName)`.

Is this about the case of `len(references)==0`?

PR #936, opened 2020-05-20
Load the rootless registries.conf.d for override

QiWang19:
Document the ConfigPath() and ConfigDirPath() as deprecated API. Add new ConfigurationSourceDescription() API returning the config file and directory sources.

Load the rootless registries.conf.d for override

### DIFF
--- a/docs/containers-registries.conf.d.5.md
+++ b/docs/containers-registries.conf.d.5.md
@@ -15,7 +15,8 @@ By default, the directory is located at `/etc/containers/registries.conf.d`.
 
 Once the main configuration at `/etc/containers/registries.conf` is loaded, the
 files in `/etc/containers/registries.conf.d` are loaded in alpha-numerical
-order. Specified fields in a config will overwrite any previous setting.  Note
+order. Then the conf files in `$HOME/.config/containers/registries.conf.d` are loaded in alpha-numerical order, if they exist. If the `$HOME/.config/containers/registries.conf` is loaded, only the conf files under `$HOME/.config/containers/registries.conf.d` are loaded in alpha-numerical order.
+Specified fields in a conf file will overwrite any previous setting.  Note
 that only files with the `.conf` prefix are loaded, other files and
 sub-directories are ignored.
 
--- a/pkg/sysregistriesv2/system_registries_v2.go
+++ b/pkg/sysregistriesv2/system_registries_v2.go
@@ -338,40 +338,18 @@ func (config *V2RegistriesConf) postProcess() error {
 }
 
 // ConfigPath returns the path to the system-wide registry configuration file.
+// Deprecated: This API implies configuration is read from files, and that there is only one.
+// Please use ConfigurationSourceDescription to obtain a string usable for error messages.
 func ConfigPath(ctx *types.SystemContext) string {
-	if ctx != nil && ctx.SystemRegistriesConfPath != "" {
-		return ctx.SystemRegistriesConfPath
-	}
-
-	userRegistriesFilePath := filepath.Join(homedir.Get(), userRegistriesFile)
-	if _, err := os.Stat(userRegistriesFilePath); err == nil {
-		return userRegistriesFilePath
-	}
-
-	if ctx != nil && ctx.RootForImplicitAbsolutePaths != "" {
-		return filepath.Join(ctx.RootForImplicitAbsolutePaths, systemRegistriesConfPath)
-	}
-
-	return systemRegistriesConfPath
+	return newConfigWrapper(ctx).configPath
 }
 
 // ConfigDirPath returns the path to the system-wide directory for drop-in
 // registry configuration files.
+// Deprecated: This API implies configuration is read from directories, and that there is only one.
+// Please use ConfigurationSourceDescription to obtain a string usable for error messages.
 func ConfigDirPath(ctx *types.SystemContext) string {
-	if ctx != nil && ctx.SystemRegistriesConfDirPath != "" {
-		return ctx.SystemRegistriesConfDirPath
-	}
-
-	userRegistriesDirPath := filepath.Join(homedir.Get(), userRegistriesDir)
-	if _, err := os.Stat(userRegistriesDirPath); err == nil {
-		return userRegistriesDirPath
-	}
-
-	if ctx != nil && ctx.RootForImplicitAbsolutePaths != "" {
-		return filepath.Join(ctx.RootForImplicitAbsolutePaths, systemRegistriesConfDirPath)
-	}
-
-	return systemRegistriesConfDirPath
+	return newConfigWrapper(ctx).configDirPath
 }
 
 // configWrapper is used to store the paths from ConfigPath and ConfigDirPath
@@ -383,10 +361,35 @@ type configWrapper struct {
 
 // newConfigWrapper returns a configWrapper for the specified SystemContext.
 func newConfigWrapper(ctx *types.SystemContext) configWrapper {
-	return configWrapper{
-		configPath:    ConfigPath(ctx),
-		configDirPath: ConfigDirPath(ctx),
+	var wrapper configWrapper
+	userRegistriesFilePath := filepath.Join(homedir.Get(), userRegistriesFile)
+	userRegistriesDirPath := filepath.Join(homedir.Get(), userRegistriesDir)
+	if ctx != nil && ctx.SystemRegistriesConfPath != "" {
+		wrapper.configPath = ctx.SystemRegistriesConfPath
+	} else if _, err := os.Stat(userRegistriesFilePath); err == nil {
+		wrapper.configPath = userRegistriesFilePath
+	} else if ctx != nil && ctx.RootForImplicitAbsolutePaths != "" {
+		wrapper.configPath = filepath.Join(ctx.RootForImplicitAbsolutePaths, systemRegistriesConfPath)
+	} else {
+		wrapper.configPath = systemRegistriesConfPath
 	}
+
+	if ctx != nil && ctx.SystemRegistriesConfDirPath != "" {
+		wrapper.configDirPath = ctx.SystemRegistriesConfDirPath
+	} else if _, err := os.Stat(userRegistriesDirPath); err == nil {
+		wrapper.configDirPath = userRegistriesDirPath
+	} else if ctx != nil && ctx.RootForImplicitAbsolutePaths != "" {
+		wrapper.configDirPath = filepath.Join(ctx.RootForImplicitAbsolutePaths, systemRegistriesConfDirPath)
+	} else {
+		wrapper.configDirPath = systemRegistriesConfDirPath
+	}
+	return wrapper
+}
+
+// ConfigurationSourceDescription returns a string containres paths of registries.conf and registries.conf.d
+func ConfigurationSourceDescription(ctx *types.SystemContext) string {
+	wrapper := newConfigWrapper(ctx)
+	return strings.Join([]string{wrapper.configPath, wrapper.configDirPath}, ", ")
 }
 
 // configMutex is used to synchronize concurrent accesses to configCache.
--- a/pkg/sysregistriesv2/system_registries_v2.go
+++ b/pkg/sysregistriesv2/system_registries_v2.go
@@ -344,19 +344,27 @@ func ConfigPath(ctx *types.SystemContext) string {
 	return newConfigWrapper(ctx).configPath
 }
 
-// ConfigDirPath returns the path to the system-wide directory for drop-in
+// ConfigDirPath returns the path to the directory for drop-in
 // registry configuration files.
 // Deprecated: This API implies configuration is read from directories, and that there is only one.
 // Please use ConfigurationSourceDescription to obtain a string usable for error messages.
 func ConfigDirPath(ctx *types.SystemContext) string {
-	return newConfigWrapper(ctx).configDirPath
+	configWrapper := newConfigWrapper(ctx)
+	if configWrapper.userConfigDirPath != "" {
+		return configWrapper.userConfigDirPath
+	}
+	return configWrapper.configDirPath
 }
 
 // configWrapper is used to store the paths from ConfigPath and ConfigDirPath
 // and acts as a key to the internal cache.
 type configWrapper struct {
-	configPath    string
+	// path to the registries.conf file
+	configPath string
+	// path to system-wide registries.conf.d directory, or "" if not used
 	configDirPath string
+	// path to user specificed registries.conf.d directory, or "" if not used
+	userConfigDirPath string
 }
 
 // newConfigWrapper returns a configWrapper for the specified SystemContext.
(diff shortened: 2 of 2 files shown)


QiWang19:
@mtrmac PTAL

docs/containers-registries.conf.d.5.md
```diff
@@ -15,7 +15,8 @@ By default, the directory is located at `/etc/containers/registries.conf.d`.
 
 Once the main configuration at `/etc/containers/registries.conf` is loaded, the
 files in `/etc/containers/registries.conf.d` are loaded in alpha-numerical
-order. Specified fields in a config will overwrite any previous setting.  Note
+order. Then the files in `$HOME/.config/containers/registries.conf.d` are loaded in alpha-numerical order, if they exist.
```
TomSweeneyRedHat, line 18:
```suggestion
order. Then the conf files in `$HOME/.config/containers/registries.conf.d` are loaded in alpha-numerical order, if they exist.
```

docs/containers-registries.conf.d.5.md
```diff
@@ -15,7 +15,8 @@ By default, the directory is located at `/etc/containers/registries.conf.d`.
...
+Specified fields in a config will overwrite any previous setting.  Note
```
TomSweeneyRedHat, line 19:
```suggestion
Specified fields in a conf file will overwrite any previous setting.  Note
```

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -355,9 +355,9 @@ func ConfigPath(ctx *types.SystemContext) string {
 	return systemRegistriesConfPath
 }
 
-// ConfigDirPath returns the path to the system-wide directory for drop-in
-// registry configuration files.
-func ConfigDirPath(ctx *types.SystemContext) string {
+// UserConfigDirPath returns the user's directory for drop-in
+// registry configuration files, returns "" is not specified
+func UserConfigDirPath(ctx *types.SystemContext) string {
 	if ctx != nil && ctx.SystemRegistriesConfDirPath != "" {
```
rhatdan, line 361: Shouldn't this still be under ConfigDirPath?

QiWang19: I split the original method ConfigDirPath(), ConfigDirPath() still there.

rhatdan: No I am talking about 

if ctx != nil && ctx.SystemRegistriesConfDirPath != "" {

Should be in the ConfigDirPath() not in UserConfigDirPath()

QiWang19: Sorry, I miss read! Will this ctx.SystemRegistriesConfDirPath comes from user specification in libpod or buildah?

rhatdan: Yes.

QiWang19: So I think it should be under UserConfigDirPath, like the $HOME/.config/containers

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -358,6 +358,16 @@ func ConfigPath(ctx *types.SystemContext) string {
...
+	if ctx != nil && ctx.RootForImplicitAbsolutePaths != "" {
+		return filepath.Join(ctx.RootForImplicitAbsolutePaths, systemRegistriesConfDirPath)
+	}
+
+	return systemRegistriesConfDirPath
+}
+
+// UserConfigDirPath returns the user's directory for drop-in
+// registry configuration files, returns "" is not specified
+func UserConfigDirPath(ctx *types.SystemContext) string {
```
mtrmac, line 370: We’ve made the mistake of providing `ConfigDirPath` exposing internal logic, which is now unusable (we have up to two directories, but can only return one), once; let’s not do it again.

Please:
- Gather all the logic for deciding on paths to use into `newConfigWrapper`
- Make `ConfigPath` and `ConfigDirPath` call `newConfigWrapper` and extract the paths form from there
- Provide new `ConfigurationSourceDescription`, which is documented to return a user-readable string to be possibly included in error messages (NOT paths otherwise consumable by callers), and returns all the paths from a `newConfigWrapper`.
- Document `ConfigPath` and `ConfigDirPath` as deprecated, directing callers to the new function.

That would be ideal as a refactoring / API change commit that does not change the actual logic (does not add the second directory), and then we can use a separate commit for adding the new features.

pkg/sysregistriesv2/system_registries_v2.go
mtrmac: In general, it should be possible for the caller to disable using the system directory. But I don’t (yet?) think we need separate overrides for the system and per-user one: if the user specifies a specific directory, we can use only that configuration.

QiWang19:
@mtrmac PTAL. What error should the ConfigurationSourceDescription return? should this stats the paths from the configWrapper?

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -338,55 +338,104 @@ func (config *V2RegistriesConf) postProcess() error {
...
+		return wrapper
 	}
+	if _, err := os.Stat(userRegistriesFilePath); err == nil {
+		wrapper.configPath = userRegistriesFilePath
+		if _, err := os.Stat(userRegistriesDirPath); err == nil {
+			wrapper.userConfigDirPath = userRegistriesDirPath
+		}
+		return wrapper
+	}
+	if ctx != nil && ctx.RootForImplicitAbsolutePaths != "" {
```
mtrmac, line 398: This is very similar to the `RootForImplicitAbsolutePaths == ""` variant, please combine them and only make the `wrapper.configPath` determination conditional.

It seems to me that we really want to make a “shall we use the system configuration file or `userRegistriesFilePath`” decision first, treat `userRegistriesFilePath` specially, and in all the other cases share the code for determining directories.

But that’s also up to what others think the behavior should be in the various cases.

docs/containers-registries.conf.d.5.md
```diff
@@ -15,7 +15,8 @@ By default, the directory is located at `/etc/containers/registries.conf.d`.
 
 Once the main configuration at `/etc/containers/registries.conf` is loaded, the
 files in `/etc/containers/registries.conf.d` are loaded in alpha-numerical
-order. Specified fields in a config will overwrite any previous setting.  Note
+order. Then the conf files in `$HOME/.config/containers/registries.conf.d` are loaded in alpha-numerical order, if they exist.
+Specified fields in a conf file will overwrite any previous setting.  Note
```
mtrmac, line 19: (Eventually this will need updating after the semantics is settled.)

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -338,55 +338,104 @@ func (config *V2RegistriesConf) postProcess() error {
...
-
-	userRegistriesDirPath := filepath.Join(homedir.Get(), userRegistriesDir)
-	if _, err := os.Stat(userRegistriesDirPath); err == nil {
-		return userRegistriesDirPath
-	}
-
-	if ctx != nil && ctx.RootForImplicitAbsolutePaths != "" {
-		return filepath.Join(ctx.RootForImplicitAbsolutePaths, systemRegistriesConfDirPath)
+	wrapper := newConfigWrapper(ctx)
+	if wrapper.userConfigDirPath != "" {
```
mtrmac, line 358: To the little extent it matters, the documentation says “to the system-wide directory”, so this should probably prefer the system path if both are set.

QiWang19:
@mtrmac  PTAL

QiWang19:
@mtrmac PTAL

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -338,55 +338,81 @@ func (config *V2RegistriesConf) postProcess() error {
...
+	if ctx == nil {
+		return systemRegistriesConfDirPath
+	}
+	return filepath.Join(ctx.RootForImplicitAbsolutePaths, systemRegistriesConfDirPath)
+}
+
+// ConfigurationSourceDescription returns a string containres paths of registries.conf and registries.conf.d
+func ConfigurationSourceDescription(ctx *types.SystemContext) string {
+	wrapper := newConfigWrapper(ctx)
+	return strings.Join([]string{wrapper.configPath, wrapper.configDirPath}, ",")
```
mtrmac, line 415: (Absolutely non-blocking: Maybe
```suggestion
	return strings.Join([]string{wrapper.configPath, wrapper.configDirPath}, ", ")
```
to make this easier to read (and marginally harder to parse by machines 👿 ))

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -338,55 +338,81 @@ func (config *V2RegistriesConf) postProcess() error {
...
-	if _, err := os.Stat(userRegistriesDirPath); err == nil {
-		return userRegistriesDirPath
-	}
-
-	if ctx != nil && ctx.RootForImplicitAbsolutePaths != "" {
-		return filepath.Join(ctx.RootForImplicitAbsolutePaths, systemRegistriesConfDirPath)
-	}
-
-	return systemRegistriesConfDirPath
+	return newConfigWrapper(ctx).configDirPath
```
mtrmac, line 352: (Non-blocking: It’s not clear what this should return if `configDirPath == ""`; both `userConfigDirPath` and `""` are wrong. `userConfigDirPath` is a bit useful for error messages.)

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -338,55 +338,81 @@ func (config *V2RegistriesConf) postProcess() error {
...
+		return ctx.SystemRegistriesConfDirPath
+	}
+	userRegistriesDirPath := filepath.Join(homedir.Get(), userRegistriesDir)
+	if _, err := os.Stat(userRegistriesDirPath); err == nil {
+		return userRegistriesDirPath
+	}
+	return ""
+}
+
+func getConfigDirPath(ctx *types.SystemContext) string {
```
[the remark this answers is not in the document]
QiWang19: > (There may not yet be consensus on this?) I think if `SystemRegistriesConfDirPath` is set, we should read only that directory, and not both the system and per-user one; that way the caller can opt out of using any existing global configuration, and we don’t have to add another option.
>


@mtrmac  I might misunderstand the systemRegistriesConfDirPath from ctx, the hardcode `SystemRegistriesConfDirPath` and the socalled root config dirs, do they mean the same thing? 
It feels like only should discriminate system-wide and per-user. system-wide will be one of ctx.SystemRegistriesConfDirPath or hard code systemRegistriesConfDirPath, and the per-user means the path under $HOME

pkg/sysregistriesv2/system_registries_v2.go
QiWang19: fixed in the refactoring commit

QiWang19:
@mtrmac PTAL

QiWang19:
@mtrmac PTAL

pkg/sysregistriesv2/system_registries_v2.go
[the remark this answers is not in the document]
mtrmac: Aargh, I didn’t realize (`ctx.`)`SystemRegistriesConfDirPath` and the `systemRegistriesConfDirPath` constant only differ in one letter. I can’t quickly find a reference to “root config dirs”.

Anyway, the logic I was considering (again, not sure there is consensus on this) is roughly:
```go
if (ctx.SystemRegistriesConfPath set) { // File explicitly chosen: potentially use both dirs
    sysFilePath = ctx.SystemRegistriesConfPath
} else if (userFilePath exists) { // User file: Don’t read system dir
    if (ctx.SystemRegistriesConfDirPath set) { // Dir explicitly chosen: use that one
        return (userFilePath, dirs = [SystemRegistriesConfDirPath])
    } else {
        return (userFilePath, dirs = [userDirPath])
    }
} else if (ctx.RootForImplicitAbsolutePaths set) { // System file elsewhere: potentially use both dirs
    sysFilePath = filepath.Join(ctx.RootForImplicitAbsolutePaths, systemRegistriesConfPath)
} else { // Default system file: Potentially use both dirs
    sysFilePath = systemRegistriesConfPath
}

// The “potentially use both dirs” logic:
if (ctx.SystemRegistriesConfDirPath set) { // Dir explicitly chosen: use only that one
   dirs = [ctx.SystemRegistriesConfDirPath]
} else {
    if (ctx.RootForImplicitAbsolutePaths set) {
        dirs = [filepath.Join(ctx.RootForImplicitAbsolutePaths, systemRegistriesConfDirPath)]
    } else {
        dirs = [systemRegistriesConfDirPath]
    }
    if (userFilePath exists) {
        dirs += userFilePath
    }
return (sysFilePath, dirs = dirs)
```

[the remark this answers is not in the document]
mtrmac: (I have edited the ‘The “potentially use both dirs” logic:’ , to fix a file/dir mix-up, and to drop a useless existence check.)

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -338,55 +338,92 @@ func (config *V2RegistriesConf) postProcess() error {
...
+	if ctx == nil {
+		wrapper.configPath = systemRegistriesConfPath
+		wrapper.configDirPath = systemRegistriesConfDirPath
+		if _, err := os.Stat(userRegistriesDirPath); err == nil {
+			wrapper.userConfigDirPath = userRegistriesDirPath
+		}
+		return wrapper
 	}
+
+	if ctx.SystemRegistriesConfPath != "" {
```
mtrmac, line 380: Please consolidate the `ctx == nil` case and this one with the code below, to ensure it behaves the same for the directories. [1]

Either the somewhat weird “check for `userRegistriesFilePath` in the middle and exit early” code in the yesterday’s pseudocode, or a somewhat redundant explicit “are using the system file or the user file” check and a top-level if/else would be fine with me.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -338,55 +338,92 @@ func (config *V2RegistriesConf) postProcess() error {
...
+	if ctx.RootForImplicitAbsolutePaths != "" {
+		wrapper.configPath = filepath.Join(ctx.RootForImplicitAbsolutePaths, systemRegistriesConfPath)
+	} else {
+		wrapper.configPath = systemRegistriesConfPath
+	}
+	if ctx.SystemRegistriesConfDirPath != "" {
+		wrapper.configDirPath = ctx.SystemRegistriesConfDirPath
+	} else {
+		if ctx.RootForImplicitAbsolutePaths != "" {
+			wrapper.configDirPath = filepath.Join(ctx.RootForImplicitAbsolutePaths, systemRegistriesConfDirPath)
```
mtrmac, line 414: [1] E.g. here the code does not set `userConfigDirPath` (I think it should), although the `if ctx.SystemRegistriesConfPath` case does use it. The two should behave consistently (`RootForImplicitAbsolutePaths` does not affect home directories).

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -338,55 +338,92 @@ func (config *V2RegistriesConf) postProcess() error {
...
+			wrapper.userConfigDirPath = userRegistriesDirPath
+		}
+	}
+	return wrapper
+}
+
+// ConfigurationSourceDescription returns a string containres paths of registries.conf and registries.conf.d
+func ConfigurationSourceDescription(ctx *types.SystemContext) string {
+	wrapper := newConfigWrapper(ctx)
+	return strings.Join([]string{wrapper.configPath, wrapper.configDirPath}, ", ")
```
mtrmac, line 426: This should include `userConfigDirPath` (in the commit that introduces it).

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -338,55 +338,92 @@ func (config *V2RegistriesConf) postProcess() error {
...
-	if _, err := os.Stat(userRegistriesDirPath); err == nil {
-		return userRegistriesDirPath
-	}
-
-	if ctx != nil && ctx.RootForImplicitAbsolutePaths != "" {
-		return filepath.Join(ctx.RootForImplicitAbsolutePaths, systemRegistriesConfDirPath)
-	}
-
-	return systemRegistriesConfDirPath
+	return newConfigWrapper(ctx).userConfigDirPath
```
mtrmac, line 352: Now that _either_ of `configDirPath` and `userConfigDirPath` can be `""`, this should return the one that is not empty.

QiWang19:
@mtrmac PTAL

docs/containers-registries.conf.d.5.md
QiWang19: @mtrmac  not sure how to describe the logic in the manpage

QiWang19:
@mtrmac can this get merged?

QiWang19:
@mtrmac @rhatdan @vrothberg PTAL. can this get merged?

rhatdan:
LGTM

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -338,55 +338,76 @@ func (config *V2RegistriesConf) postProcess() error {
...
 
-// ConfigDirPath returns the path to the system-wide directory for drop-in
+// ConfigDirPath returns the path to the directory for drop-in
 // registry configuration files.
+// Deprecated: This API implies configuration is read from directories, and that there is only one.
+// Please use ConfigurationSourceDescription to obtain a string usable for error messages.
 func ConfigDirPath(ctx *types.SystemContext) string {
-	if ctx != nil && ctx.SystemRegistriesConfDirPath != "" {
-		return ctx.SystemRegistriesConfDirPath
+	if newConfigWrapper(ctx).userConfigDirPath != "" {
```
mtrmac, line 352: Non-blocking: It would be better to call `newConfigWrapper` only once, to avoid the repeated system calls.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -338,55 +338,76 @@ func (config *V2RegistriesConf) postProcess() error {
...
+		wrapper.userConfigDirPath = userRegistriesDirPath
+	}
+
+	return wrapper
+}
+
+// ConfigurationSourceDescription returns a string containres paths of registries.conf and registries.conf.d
+func ConfigurationSourceDescription(ctx *types.SystemContext) string {
+	wrapper := newConfigWrapper(ctx)
+	return strings.Join([]string{wrapper.configPath, wrapper.configDirPath, wrapper.userConfigDirPath}, ", ")
```
mtrmac, line 410: I don’t think this does the right thing if either of the `*DirPath` values is `""`.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -338,55 +338,76 @@ func (config *V2RegistriesConf) postProcess() error {
...
+	// path to user specificed registries.conf.d directory, or "" if not used
+	userConfigDirPath string
 }
 
 // newConfigWrapper returns a configWrapper for the specified SystemContext.
 func newConfigWrapper(ctx *types.SystemContext) configWrapper {
-	return configWrapper{
-		configPath:    ConfigPath(ctx),
-		configDirPath: ConfigDirPath(ctx),
+	var wrapper configWrapper
```
vrothberg, line 371: Please add comments to the code below. Three level nested branches can easily overflow my brain parser :)

mtrmac: Would not factoring out the `ctx != nil` check work better?

See https://github.com/containers/image/pull/936#discussion_r436804626 and https://github.com/containers/image/pull/936#discussion_r437601619 for the intent.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -338,55 +338,89 @@ func (config *V2RegistriesConf) postProcess() error {
...
 	}
+
+	return wrapper
+}
+
+// ConfigurationSourceDescription returns a string containres paths of registries.conf and registries.conf.d
+func ConfigurationSourceDescription(ctx *types.SystemContext) string {
+	var configSources []string
+	wrapper := newConfigWrapper(ctx)
+	if wrapper.configPath != "" {
```
mtrmac, line 414: This is supposed to always be `true`.

pkg/sysregistriesv2/system_registries_v2.go
QiWang19: @vrothberg PTAL. Comments added.

vrothberg:
@mtrmac good to merge?

TomSweeneyRedHat:
LGTM

PR #940, opened 2020-05-23
Fix crash on inspecting an OCI image with no config

mtrmac:
Don't use the schema2 config field, which is a pointer that may be nil; the OCI config is an embedded struct and accessing it is always possible.

Using the OCI config is more correct anyway.

Should fix https://github.com/containers/libpod/issues/6361 .


gasinvein:
Applied on top of podman 1.9.2, it fixes the crash for me, thanks.

rhatdan:
LGTM
@vrothberg PTAL

PR #942, opened 2020-05-25
docker/config - get all credentials 

vrothberg:
@mtrmac ... I am reopening for testing purposes. Let's discuss this tomorrow in the water cooler with @baude @jwhonce and @rhatdan but I really think we should go this path.

Querying for all credentials on the client and sending them over to the server is exponentially easier and less error prone than querying the server individually based on a set of references. On one hand, it's already quite hairy to get simple cases right as we need to combine the pull-ref-pair semantics when pulling in libpod _with_ the registries.conf registries and mirrors. On the other, there are commands where determining the set of images is not straight forward. Those commands are build and play-kube. Client-side parsing of the files would entail pulling in dependencies that @baude and @jwhonce are trying to prune from the remote-client (e.g., K8s stuff).

While non of the mentioned issues are rocket science, the amount of work to get it work plus the amount of testing exceed the benefits IMHO. Sending all credentials at once (unless user/pw are set on the CLI) doesn't run into these issues at all and reduces code complexity.

Old PR: https://github.com/containers/image/pull/928


vrothberg:
> Most (all?) of the detailed review comments from [#928 (review)](https://github.com/containers/image/pull/928#pullrequestreview-412888432) are still outstanding.

Yes, they are still open. I didn't want to invest time until there is consensus on the way forward. The current state is sufficient for testing the libpod PR though.

> Also:
> 
>     * How does this work with the plan to default credential storage for keyrings? This does nothing if all credentials are in keyrings.

The credential helper API includes a list operation for listing stored credentials. As long as the future defaults adhere to that API, we should be fine.

>     * How _will_ credential helpers be configured for Podman clients, actually? It can’t be by reading Podman server’s `registries.conf`*

I honestly have no idea of the remote client is envisioned with regards to configuration. @rhatdan and @baude PTAL.

>       * (This is BTW somewhat of a reason not store this in `registries.conf`: Podman clients need to configure credential helpers, but not the other parts of `registries.conf`, at least the way I imagine things working (OTOH I don’t want to reopen this decision)).
> 
>     * The Podman server will need a way to use the submitted credentials and otherwise ignore the system-wide default credential helper. What will that be?

I cannot look too far in the future but in the open PR against libpod, we either set the a temporary AuthFilePath OR the DockerAuthConfig of the SystemContext. However, if defaults in c/image changed, we had to keep that in the back of our heads. But other users might very well depend on that behaviour as well.

> The more I think about this, the _less_ clear the picture seems so far :-|

I share the feeling but am convinced that listing (and sending) all credentials is the lesser evil.

vrothberg:
Updated, tests added. @mtrmac @rhatdan PTAL.

rhatdan:
LGTM
@QiWang19 PTAL

QiWang19:
LGTM

pkg/docker/config/config.go
```diff
@@ -73,6 +73,59 @@ func SetAuthentication(sys *types.SystemContext, registry, username, password st
...
+				return nil, err
+			}
+			authConfigs[normalizeRegistry(registry)] = conf
+		}
+
+		// Credential helpers may override credentials from the auth file.
+		for registry, credHelper := range auths.CredHelpers {
+			username, password, err := getAuthFromCredHelper(credHelper, registry)
+			if err != nil {
+				return nil, err
```
QiWang19, line 103: should ignore the error if it's from `credentials.IsErrCredentialsNotFoundMessage`?

pkg/docker/config/config_test.go
```diff
@@ -454,3 +459,56 @@ func TestGetAuthFailsOnBadInput(t *testing.T) {
...
+			username: "local-user",
+			password: "local-password",
+		},
+	}
+
+	// Write the credentials to the authfile.
+	for _, d := range data {
+		err := SetAuthentication(&sys, d.server, d.username, d.password)
+		require.NoError(t, err)
+	}
```
mtrmac, line 500: (Absolutely non-blocking: There are already quite a few test files, including the legacy / non-normalized formats. Reading them would be more representative than creating a new config file using the latest `SetAuthentication` implementation.)

pkg/docker/config/config.go
```diff
@@ -73,6 +73,59 @@ func SetAuthentication(sys *types.SystemContext, registry, username, password st
...
+			conf := types.DockerAuthConfig{Username: username, Password: password}
+			authConfigs[normalizeRegistry(registry)] = conf
+		}
+	}
+
+	return authConfigs, nil
+}
+
+// getAuthFilePaths returns a slice of authPaths based on the system context
+// and some default and legacy paths. Note that some paths may not exist.
```
mtrmac, line 115:
```suggestion
// in the order they should be searched. Note that some paths may not exist.
```

(It might also be helpful to document in `getPathToAuth` that this is the path that should be used for writing any changes… or maybe not, because right now we can’t delete an entry from `dockerHomePath`, so the single `getPathToAuth` function may need to be replaced for that purpose anyway.)

pkg/docker/config/config.go
```diff
@@ -73,6 +73,59 @@ func SetAuthentication(sys *types.SystemContext, registry, username, password st
...
+func getAuthFilePaths(sys *types.SystemContext) []authPath {
```
mtrmac, line 116: 👍 

(Eventually we might want to go full “enterprise” and make a “credential store” interface, and include the keyring / credential helpers in the order. But that would not be a benefit for this PR.)

pkg/docker/config/config.go
mtrmac: Yes.

pkg/docker/config/config.go
```diff
@@ -73,6 +73,59 @@ func SetAuthentication(sys *types.SystemContext, registry, username, password st
 	})
 }
 
+// GetAllCredentials returns the registry credentials for all registries stored
+// in either the auth.json file or the docker/config.json.
+func GetAllCredentials(sys *types.SystemContext) (map[string]types.DockerAuthConfig, error) {
```
mtrmac, line 78: (Absolutely non-blocking, a weird thought: Would it be more useful to return a `[]byte authConfigFile`, not guaranteeing _anything_ about its format, just that it can be fed back into (…some version of?) this package via `SystemContext.AuthFilePath`?

That would discourage callers from searching in the path, and it would make it possible to add new kinds of credentials in the future or new feeds to `DockerAuthConfig` without the caller having to know — OTOH it would also require the Podman API to pass the `[]byte` as a single unit, and to make the format effectively undocumented in the Podman API, which I suppose is unacceptable.)

pkg/docker/config/config.go
vrothberg: Good catch, thanks!

pkg/docker/config/config.go
vrothberg: I concur with the `OTOH` thoughts. We might get into bigger troubles. But certainly I share the desire to not expose such low-level APIs.

pkg/docker/config/config.go
```diff
@@ -93,21 +146,7 @@ func GetCredentials(sys *types.SystemContext, registry string) (types.DockerAuth
...
 
-	paths := []authPath{}
-	pathToAuth, lf, err := getPathToAuth(sys)
-	if err == nil {
-		paths = append(paths, authPath{path: pathToAuth, legacyFormat: lf})
-	} else {
-		// Error means that the path set for XDG_RUNTIME_DIR does not exist
-		// but we don't want to completely fail in the case that the user is pulling a public image
-		// Logging the error as a warning instead and moving on to pulling the image
-		logrus.Warnf("%v: Trying to pull image in the event that it is a public image.", err)
```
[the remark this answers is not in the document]
mtrmac: This is still outstanding.

mtrmac:
(I’ll be around in 6-8 hours. I’d like https://github.com/containers/image/pull/942#discussion_r432084766 to be fixed, but it would be acceptable to merge this PR as is and retrofit that later, if a release were that urgent.)

pkg/docker/config/config.go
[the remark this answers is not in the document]
vrothberg: Ouch. Thanks for double checking! Moved it back now.

vrothberg:
> (I’ll be around in 6-8 hours. I’d like [#942 (comment)](https://github.com/containers/image/pull/942#discussion_r432084766) to be fixed, but it would be acceptable to merge this PR as is and retrofit that later, if a release were that urgent.)

Addressed now. Thanks for double checking!

PR #944, opened 2020-05-26
Add hardcode Authfile for windows and mac

QiWang19:
Add hardcode Authfile for windows and mac used for the credential store.
fix https://github.com/containers/skopeo/issues/923

### DIFF
--- a/docs/containers-auth.json.5.md
+++ b/docs/containers-auth.json.5.md
@@ -5,8 +5,9 @@ containers-auth.json - syntax for the registry authentication file
 
 # DESCRIPTION
 
-A credentials file stored at `${XDG_RUNTIME_DIR}/containers/auth.json` in
-json format used to authenticate against container image registries.
+A credentials file in JSON format used to authenticate against container image registries.
+On Linux it is stored at `${XDG_RUNTIME_DIR}/containers/auth.json`;
+on Windows and macOS, at `$HOME/.config/containers/auth.json`
 
 ## FORMAT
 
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM, but should we also make this ~/.config/containers/auth.json or ~/.containers/auth.json

pkg/docker/config/config.go
```diff
@@ -37,6 +38,7 @@ var (
 	xdgRuntimeDirPath       = filepath.FromSlash("containers/auth.json")
 	dockerHomePath          = filepath.FromSlash(".docker/config.json")
 	dockerLegacyHomePath    = ".dockercfg"
+	nonLinuxAuthFilePath    = filepath.FromSlash(".containers/auth.json")
```
mtrmac, line 41: We use `~/.config/containers` for `policy.json`, `registries.conf*`, `certs.d`, so it would make sense to me to use the same directory here as well.

docs/containers-auth.json.5.md
```diff
@@ -7,6 +7,7 @@ containers-auth.json - syntax for the registry authentication file
 
 A credentials file stored at `${XDG_RUNTIME_DIR}/containers/auth.json` in
 json format used to authenticate against container image registries.
+For windows and darwin systems the credential file is stored at `$HOME/.config/containers/auth.json`
```
mtrmac, line 10: Something like
```suggestion
A credentials file in JSON format used to authenticate against container image registries.
On Linux it is stored at `${XDG_RUNTIME_DIR}/containers/auth.json`;
on Windows and macOS, at `$HOME/.config/containers/auth.json`
```

rhatdan:
LGTM
Thanks @QiWang19

rhatdan:
@mtrmac Can you cut a new c/i release?

PR #948, opened 2020-06-03
Add support for ProgressEventSkipped

saschagrunert:
The new event indicates that a blob has been skipped. This information
can be used by CRI-O for more detailed image layer re-use metrics.

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -1060,6 +1060,14 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, to
 			logrus.Debugf("Skipping blob %s (already present):", srcInfo.Digest)
 			bar := ic.c.createProgressBar(pool, srcInfo, "blob", "skipped: already exists")
 			bar.SetTotal(0, true)
+
+			// Throw an event that the layer has been skipped
+			if ic.c.progress != nil && ic.c.progressInterval > 0 {
+				ic.c.progress <- types.ProgressProperties{
+					Event:    types.ProgressEventSkipped,
+					Artifact: srcInfo,
+				}
+			}
 			return blobInfo, cachedDiffID, nil
 		}
 	}
--- a/types/types.go
+++ b/types/types.go
@@ -604,6 +604,10 @@ const (
 	// ProgressEventDone is fired when the data transfer has been finished for
 	// the specific artifact
 	ProgressEventDone
+
+	// ProgressEventSkipped is fired when the artifact has been skipped because
+	// its already available at the destination
+	ProgressEventSkipped
 )
 
 // ProgressProperties is used to pass information from the copy code to a monitor which


copy/copy.go
```diff
@@ -1060,6 +1060,14 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, to
 			logrus.Debugf("Skipping blob %s (already present):", srcInfo.Digest)
 			bar := ic.c.createProgressBar(pool, srcInfo, "blob", "skipped: already exists")
 			bar.SetTotal(0, true)
+
+			// Throw an event that the layer has been skipped
+			if ic.c.progress != nil {
```
mtrmac, line 1065: The other condition has a `&& c.progressInterval > 0`. Is there a reason to deviate?

rhatdan:
LGTM

copy/copy.go
```diff
@@ -1060,6 +1060,14 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, to
...
+				ic.c.progress <- types.ProgressProperties{
+					Event:    types.ProgressEventSkipped,
```
mtrmac, line 1067: (Should there be a `ProgressEventNewArtifact` event as well?

I’m don’t see a strong reason to prefer either way, maybe not doing it is a tiny bit simpler. I just want to make sure this has been considered.)

copy/copy.go
```diff
@@ -1060,6 +1060,14 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, to
...
+					Artifact: blobInfo,
```
mtrmac, line 1068: Elsewhere (both in the progress bar code above, and in the event channel used when compressing/decompressing during copy), we use the original `srcInfo` data/digests; so it seems reasonable to do the same here.

types/types.go
```diff
@@ -604,6 +604,10 @@ const (
 	// ProgressEventDone is fired when the data transfer has been finished for
 	// the specific artifact
 	ProgressEventDone
+
+	// ProgressEventSkipped is fired when the artifact has been skipped because
+	// its already locally available
```
mtrmac, line 609:
```suggestion
	// its already available at the destination
```

rhatdan:
@saschagrunert Could you answer @mtrmac ?

copy/copy.go
saschagrunert: Hm, I added the check for sake of consistency in behavior. :+1:

copy/copy.go
saschagrunert: We already have it here:
https://github.com/containers/image/blob/8d71f3f4c9aafaca1952fabfa51e329c3b7d46ed/copy/progress_reader.go#L32-L37

If you want then I could follow-up by adding a new event and changing the value there. The thing is that we will break the API with something like that.

copy/copy.go
saschagrunert: Makes sense, changed it to use `srcInfo` :+1:

types/types.go
saschagrunert: Changed as suggested :+1:

rhatdan:
Thanks @mtrmac PTAL

copy/copy.go
mtrmac: I meant: should the sequence of events for a skipped blob be `NewArtifact`+`Skipped`, or `Skipped` only?

I can imagine a caller that would “register” a blob on `NewArtifact` and “mark as done” on either `Skipped` or `Done`, where it would be convenient to emit both. (Hum, in that case we probably should call `NewArtifact` before `TryReusingBlob` spends time waiting for a remote registry.)

OTOH I can well imagine a different caller that wants to differentiate between skipped blobs (e.g. no progress bar at all, the roundtrips for reuse detection are considered negligible) and copied blobs, and in that case it’s useful that skipped blobs don’t start with `NewArtifact`; we would need a `NewArtifact`+`ReallyStartingToCopy`+`Read`+`Done`.

Looking at that `ReallyStartingToCopy`… let’s not worry about that now. We can, later, add a `CheckingForReuse` phase before the `NewArtifact`/`Skipped` report, if it turned out useful to watch for the time spent in `TryReuseBlob` .

PR #952, opened 2020-06-04
Fix error messages on !canModifyManifest

mtrmac:
- If `LayerInfosForCopy` returns a modified list, report it as an user-targeted error instead of an internal one: this is the first place a `podman push` of a signed image would fail.
- In error messages, report both possible causes (signed image or a digested destination).  I took the simple approach here, listing both options , instead of determining which one applies in the current situation.


rhatdan:
LGTM
@vrothberg PTAL

PR #959, opened 2020-06-09
Print warning if not using credential helpers

QiWang19:
Print warning message if login command stores credentials as plain text without credential configuraiton.

close https://github.com/containers/libpod/issues/4119#issuecomment-641435325

### DIFF
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -72,6 +72,9 @@ func SetAuthentication(sys *types.SystemContext, registry, username, password st
 			}
 			logrus.Debugf("failed to authenticate with the kernel keyring, falling back to authfiles. %v", err)
 		}
+		fmt.Fprintf(os.Stderr, "%s\n", `WARNING! Your password will be stored unencrypted in authentication file
+Configure a credential helper to remove this warning. See
+https://github.com/containers/image/blob/master/docs/containers-auth.json.5.md`)
 		creds := base64.StdEncoding.EncodeToString([]byte(username + ":" + password))
 		newCreds := dockerAuthConfig{Auth: creds}
 		auths.AuthConfigs[registry] = newCreds


QiWang19:
@mtrmac @rhatdan @TomSweeneyRedHat PTAL. Does it make sense to print the warning here?

pkg/docker/config/config.go
```diff
@@ -72,6 +72,9 @@ func SetAuthentication(sys *types.SystemContext, registry, username, password st
 			}
 			logrus.Debugf("failed to authenticate with the kernel keyring, falling back to authfiles. %v", err)
 		}
+		fmt.Fprintf(os.Stderr, "%s\n", `WARNING! Your password will be stored unencrypted in authentication file
```
TomSweeneyRedHat, line 75:
```suggestion
		fmt.Fprintf(os.Stderr, "%s\n", `WARNING! Your password will be stored unencrypted in the authentication file.
```

TomSweeneyRedHat: With the `'`, I'm not sure if there'll be a carriage return after this line or not.  If not, you need to make sure there's at least a space between it and the first word in the next line.

TomSweeneyRedHat:
Looks like the new output has upset some of the prior tests.
The location LGTM, but if @mtrmac or @rhatdan have a differing opinion, I'd go with their thoughts.

pkg/docker/config/config.go
mtrmac, line 75: Hard-coding `os.Stderr` in a library is problematic. The closest we have for similar cases so far is `logrus.Warn*`; would that work here?

QiWang19: Can I use the logrus.Warn* explicitly printout the warning without using --log-level?

QiWang19:
At least the `credsHelper` config is supported in the current authentication file, this idea is just compatible with docker.  The current docker has this warning to suggest configuring credential helpers in config.json.

pkg/docker/config/config.go
mtrmac: The default level in Logrus is “info” (relevant for Skopeo), and both Buildah and Podman default to “error”, so that should work fine AFAICS.

(One downside is that it uses the Logrus format, with some visual noise. That’s something to decide/configure at the top-level executable level, though.)

mtrmac:
> At least the `credsHelper` config is supported in the current authentication file, this idea is just compatible with docker.

It doesn’t quite work the same way because the auth files used for updates are not persisted, yet (and credential helpers configured in `~/.docker/config.json` are ignored on the write path, IIRC). We need https://github.com/containers/image/pull/936 + the follow-up to store credential helpers there, first.

QiWang19:
ok, I will close this and rework on the idea after the credential helpers can be used as default.

PR #960, opened 2020-06-15
oci: don't overwrite tags pointing at the same manifest

tych0:
In ca5fe04cb38a ("Add manifest list support"), the code was changed from a
single step (match on the name annotation) to a two step process: 1. untag
the existing name, and 2. match on the manifest's digest and change the tag
of this match to the new tag.

However, this means that if there were two tags:

    "foo": sha256:c0ffee55
    "bar": sha256:deadbeef

And we update "bar" to point to c0ffee55, we'd end up doing:

1. remove any "bar" tags:

    "foo": sha256:c0ffee55
    "": sha256:deadbeef

2. re-tag the first c0ffee55 thing:

    "bar": sha256:c0ffee55
    "": sha256:deadbeef

Thus losing the "foo" tag entirely.

Instead, let's make sure that we only re-tag something with the empty
string, so we don't delete existing tags.

### DIFF
--- a/oci/layout/oci_dest_test.go
+++ b/oci/layout/oci_dest_test.go
@@ -106,7 +106,7 @@ func TestPutManifestTwice(t *testing.T) {
 
 	index, err := ociRef.getIndex()
 	assert.NoError(t, err)
-	assert.Equal(t, 2, len(index.Manifests), "Unexpected number of manifests")
+	assert.Len(t, index.Manifests, 2, "Unexpected number of manifests")
 }
 
 func TestPutTwoDifferentTags(t *testing.T) {
--- a/oci/layout/oci_dest_test.go
+++ b/oci/layout/oci_dest_test.go
@@ -11,6 +11,7 @@ import (
 	"github.com/containers/image/v5/pkg/blobinfocache/memory"
 	"github.com/containers/image/v5/types"
 	digest "github.com/opencontainers/go-digest"
+	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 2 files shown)


oci/layout/oci_dest_test.go
```diff
@@ -108,6 +109,27 @@ func TestPutManifestTwice(t *testing.T) {
...
+	defer os.RemoveAll(tmpDir)
+
+	ociRef, ok := ref.(ociReference)
+	require.True(t, ok)
+
+	putTestConfig(t, ociRef, tmpDir)
+	putTestManifest(t, ociRef, tmpDir)
+
+	// add the same manifest with a different tag; it shouldn't get overwritten
+	ociRef.image = "zomg"
```
mtrmac, line 123: References are conceptually immutable. This should work similarly, and be more representative, with a
```go
   ref2, … = NewReference(tmpDir, "zomg")
```

oci/layout/oci_dest_test.go
```diff
@@ -108,6 +109,27 @@ func TestPutManifestTwice(t *testing.T) {
...
+	putTestManifest(t, ociRef, tmpDir)
+
+	index, err := ociRef.getIndex()
+	assert.NoError(t, err)
+	assert.Equal(t, 3, len(index.Manifests), "Unexpected number of manifests")
```
mtrmac, line 128: Absolutely non-blocking:
```suggestion
	assert.Len(t, index.Manifests, 3)
```

tych0: Sure, I just copied this from somewhere else, so I fixed that location up too.****

oci/layout/oci_dest_test.go
tych0: Done.

oci/layout/oci_dest_test.go
```diff
@@ -105,7 +106,31 @@ func TestPutManifestTwice(t *testing.T) {
...
+
+	putTestConfig(t, ociRef, tmpDir)
+	putTestManifest(t, ociRef, tmpDir)
+
+	// add the same manifest with a different tag; it shouldn't get overwritten
+	ref, err := NewReference(tmpDir, "zomg")
+	ociRef, ok = ref.(ociReference)
+	require.True(t, ok)
+	assert.NoError(t, err)
+	putTestManifest(t, ociRef, tmpDir)
```
mtrmac, line 127:
```suggestion
	ref, err := NewReference(tmpDir, "zomg")
	require.NoError(t, err)
	ociRef, ok = ref.(ociReference)
	require.True(t, ok)
	putTestManifest(t, ociRef, tmpDir)
```
or the error check would probably not trigger.

oci/layout/oci_dest_test.go
mtrmac: Thanks

oci/layout/oci_dest_test.go
tych0: Oh, yeah, fixed thanks.

PR #963, opened 2020-06-17
When we can't store signatures, point the user at the destination.

mtrmac:
Currently the error message is:
> Can not copy signatures: X-Registry-Supports-Signatures extension not supported, and lookaside is not configured

which does not tell the user whether to check the source or the destination.

Mention the destination explicitly.


TomSweeneyRedHat:
LGTM

rhatdan:
LGTM

PR #964, opened 2020-06-17
enable search using pagination

QiWang19:
Enable search registry specifying the number of results rather than using default API 100 limit.
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1827794

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -331,7 +331,6 @@ func SearchRegistry(ctx context.Context, sys *types.SystemContext, registry, ima
 		// Results holds the results returned by the /v1/search endpoint
 		Results []SearchResult `json:"results"`
 	}
-	v2Res := &V2Results{}
 	v1Res := &V1Results{}
 
 	// Get credentials from authfile for the underlying hostname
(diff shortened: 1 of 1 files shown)


docker/docker_client.go
```diff
@@ -388,7 +388,13 @@ func SearchRegistry(ctx context.Context, sys *types.SystemContext, registry, ima
 	}
 
 	logrus.Debugf("trying to talk to v2 search endpoint")
-	resp, err := client.makeRequest(ctx, "GET", "/v2/_catalog", nil, nil, v2Auth, nil)
+	u := url.URL{
+		Path: "/v2/_catalog",
+	}
+	q := u.Query()
+	q.Set("n", strconv.Itoa(limit))
```
TomSweeneyRedHat, line 395: Does limit get set to 100 if not specified by the user?

QiWang19: I think Podman search can limit this to its default number of result 25. https://github.com/containers/libpod/blob/3eb0ad04a8b1d56866a16f1428bb8019927ccfa3/libpod/image/search.go#L136-L139

QiWang19:
The linked bug expected to receive the total result if the total number of results when there are over 100.
To support the pagination, the podman will provide a page size argument and with the current limit for total number of results. These enable to `podman search` to return the paginated result and this is the functionality that should be provided. Did I get it right?

mtrmac:
No; “pagination” means that a single HTTP request only returns up to N results (a “page”), and the client must make another HTTP request for another page, and yet another for the third page, and so on.

See https://github.com/docker/distribution/blob/master/docs/spec/api.md#pagination , and  `GetRepositoryTags` for an example.

mtrmac:
(The page size is ultimately decided by the server, which just won’t return larger pages than some internally-decided number. It doesn’t need to be a number specified by Podman, and especially users should not have to care.)

rhatdan:
If the user cares about page size then they can pipe it to less.

mtrmac:
The “page” size from the API is completely unrelated to the end-user UI, first of all because the “pages” apply to the full catalog but we only return results of the search in that catalog.

QiWang19:
@mtrmac @vrothberg @TomSweeneyRedHat @rhatdan PTAL. This uses the pagination until the search result reaches the limit.

docker/docker_client.go
```diff
@@ -388,31 +388,54 @@ func SearchRegistry(ctx context.Context, sys *types.SystemContext, registry, ima
...
 
 	logrus.Debugf("trying to talk to v2 search endpoint")
-	resp, err := client.makeRequest(ctx, "GET", "/v2/_catalog", nil, nil, v2Auth, nil)
-	if err != nil {
-		logrus.Debugf("error getting search results from v2 endpoint %q: %v", registry, err)
-	} else {
+	searchRes := []SearchResult{}
+	resultLimit := limit
+	path := "/v2/_catalog"
+	for resultLimit > 0 {
```
mtrmac, line 394: The `resultLimit` name does not really say “the number of results we still want to get”.

Instead of worrying about naming it correctly, we don’t actually need that variable: `len(searchRes) < limit` computes the same thing in the same O(1) time, and we wouldn’t have to worry about keeping `searches` and `resultLimit` in sync.

docker/docker_client.go
```diff
@@ -388,31 +388,54 @@ func SearchRegistry(ctx context.Context, sys *types.SystemContext, registry, ima
...
+		}
 		defer resp.Body.Close()
 		if resp.StatusCode != http.StatusOK {
 			logrus.Errorf("error getting search results from v2 endpoint %q: %v", registry, httpResponseToError(resp, ""))
-		} else {
-			if err := json.NewDecoder(resp.Body).Decode(v2Res); err != nil {
-				return nil, err
+			return nil, errors.Wrapf(err, "couldn't search registry %q", registry)
+		}
+		if err := json.NewDecoder(resp.Body).Decode(v2Res); err != nil {
```
mtrmac, line 405: Non-blocking: Making the `v2res` variable local to the loop body would avoid having to worry about reusing the value across loop iterations, and what `Decode` does with pre-existing values. (This does work fine, though.)

docker/docker_client.go
```diff
@@ -388,31 +388,54 @@ func SearchRegistry(ctx context.Context, sys *types.SystemContext, registry, ima
...
+		if link == "" {
+			break
+		}
+		linkURLStr := strings.Trim(strings.Split(link, ";")[0], "<>")
+		linkURL, err := url.Parse(linkURLStr)
+		if err != nil {
+			return searchRes, err
+		}
+
+		path = linkURL.Path
```
mtrmac, line 432: Non-blocking: It would be nice to preserve comment from the other implementation, or some variant of it, about explicitly deciding not to use the host name etc. from `linkURL`.

PR #970, opened 2020-06-24
Cleanup tests

tych0:
Some tests were leaking their dirs in /tmp.


TomSweeneyRedHat:
LGTM
assuming happy tests

PR #971, opened 2020-06-24
Auth configs in registries.conf

QiWang19:
Allow `credential-helpers = []` and `credential-helper = ""` configuration in registries.conf as defautl store for credentials.
```toml
credential-helpers = ["secretservice"] # general credential helpers
[[registry]]
credential-helper = "secretservice" # credhelper for specfic registry
location = "example.com/foo"
```

close https://github.com/containers/podman/issues/4119

### DIFF
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -10,6 +10,7 @@ import (
 	"runtime"
 	"strings"
 
+	"github.com/containers/image/v5/pkg/sysregistriesv2"
 	"github.com/containers/image/v5/types"
 	"github.com/containers/storage/pkg/homedir"
 	helperclient "github.com/docker/docker-credential-helpers/client"
@@ -54,9 +55,23 @@ var (
 	ErrNotSupported = errors.New("not supported")
 )
 
-// SetAuthentication stores the username and password in the auth.json file
+// SetAuthentication stores the username and password in the credential helper or file
 func SetAuthentication(sys *types.SystemContext, registry, username, password string) error {
+	helpers, err := sysregistriesv2.CredentialHelpersForRegistry(sys, registry)
+	if err != nil {
+		return err
+	}
+	for _, helper := range helpers {
+		if err = setAuthToCredHelper(helper, registry, username, password); err != nil {
+			logrus.Warnf("error storing credentials for %s to the credential helper %s, %v", registry, helper, err)
+			continue
+		}
+		logrus.Debugf("credentials for %s were stored in the credential helper %s", registry, helper)
+		return nil
+	}
+
 	return modifyJSON(sys, func(auths *dockerConfigFile) (bool, error) {
+
 		if ch, exists := auths.CredHelpers[registry]; exists {
 			return false, setAuthToCredHelper(ch, registry, username, password)
 		}
@@ -81,45 +96,66 @@ func SetAuthentication(sys *types.SystemContext, registry, username, password st
 }
 
 // GetAllCredentials returns the registry credentials for all registries stored
-// in either the auth.json file or the docker/config.json.
+// in either the auth.json file or the docker/config.json, and the keyring.
 func GetAllCredentials(sys *types.SystemContext) (map[string]types.DockerAuthConfig, error) {
-	// Note: we need to read the auth files in the inverse order to prevent
+	// Note: we need to update the authConfigs only if the auth for the registry does not exist to prevent
 	// a priority inversion when writing to the map.
 	authConfigs := make(map[string]types.DockerAuthConfig)
+	// query all credentials from the credential helpers
+	helpers, err := sysregistriesv2.AllConfiguredCredentialHelpers(sys)
+	if err != nil {
+		return nil, err
+	}
+	for _, helper := range helpers {
+		creds, err := listAuthFromCredHelper(helper)
+		if err != nil {
+			return nil, err
+		}
+		for registry := range creds {
+			if _, ok := authConfigs[normalizeRegistry(registry)]; ok {
+				continue
+			}
+			conf, err := getAuthFromCredHelper(helper, registry)
+			if err != nil {
+				return nil, err
+			}
+			authConfigs[normalizeRegistry(registry)] = conf
+		}
+	}
+
 	paths := getAuthFilePaths(sys, homedir.Get())
-	for i := len(paths) - 1; i >= 0; i-- {
-		path := paths[i]
+	for _, path := range paths {
 		// readJSONFile returns an empty map in case the path doesn't exist.
 		auths, err := readJSONFile(path.path, path.legacyFormat)
 		if err != nil {
 			return nil, errors.Wrapf(err, "error reading JSON file %q", path.path)
 		}
 
-		for registry, data := range auths.AuthConfigs {
-			conf, err := decodeDockerAuth(data)
-			if err != nil {
-				return nil, err
-			}
-			authConfigs[normalizeRegistry(registry)] = conf
-		}
-
-		// Credential helpers may override credentials from the auth file.
 		for registry, credHelper := range auths.CredHelpers {
-			username, password, err := getAuthFromCredHelper(credHelper, registry)
+			if _, ok := authConfigs[normalizeRegistry(registry)]; ok {
+				continue
+			}
+			conf, err := getAuthFromCredHelper(credHelper, registry)
 			if err != nil {
 				if credentials.IsErrCredentialsNotFoundMessage(err.Error()) {
 					continue
 				}
 				return nil, err
 			}
+			authConfigs[normalizeRegistry(registry)] = conf
+		}
 
-			conf := types.DockerAuthConfig{Username: username, Password: password}
+		for registry, data := range auths.AuthConfigs {
+			if _, ok := authConfigs[normalizeRegistry(registry)]; ok {
+				continue
+			}
+			conf, err := decodeDockerAuth(data)
+			if err != nil {
+				return nil, err
+			}
 			authConfigs[normalizeRegistry(registry)] = conf
 		}
 	}
-
-	// TODO(keyring): if we ever re-enable the keyring support, we had to
-	// query all credentials from the keyring here.
 
 	return authConfigs, nil
 }
@@ -174,6 +210,23 @@ func getCredentialsWithHomeDir(sys *types.SystemContext, registry, homeDir strin
 		return *sys.DockerAuthConfig, nil
 	}
 
+	helpers, err := sysregistriesv2.CredentialHelpersForRegistry(sys, registry)
+	if err != nil {
+		return types.DockerAuthConfig{}, err
+	}
+	for _, helper := range helpers {
+		creds, err := getAuthFromCredHelper(helper, registry)
+		if err != nil {
+			if credentials.IsErrCredentialsNotFoundMessage(err.Error()) {
+				logrus.Debugf("credentials not found in %s, %v", helper, err)
+				continue
+			}
+			return creds, err
+		}
+		logrus.Debugf("Returning credentials from credential helper %s", helper)
+		return creds, nil
+	}
+
 	if enableKeyring {
 		username, password, err := getAuthFromKernelKeyring(registry)
 		if err == nil {
@@ -188,7 +241,7 @@ func getCredentialsWithHomeDir(sys *types.SystemContext, registry, homeDir strin
 	for _, path := range getAuthFilePaths(sys, homeDir) {
 		authConfig, err := findAuthentication(registry, path.path, path.legacyFormat)
 		if err != nil {
-			logrus.Debugf("Credentials not found")
+			logrus.Debugf("Credentials not found, %v", err)
 			return types.DockerAuthConfig{}, err
 		}
 
@@ -229,6 +282,24 @@ func getAuthenticationWithHomeDir(sys *types.SystemContext, registry, homeDir st
 
 // RemoveAuthentication deletes the credentials stored in auth.json
 func RemoveAuthentication(sys *types.SystemContext, registry string) error {
+	helpers, err := sysregistriesv2.CredentialHelpersForRegistry(sys, registry)
+	if err != nil {
+		return err
+	}
+	for _, helper := range helpers {
+		if _, err = getAuthFromCredHelper(helper, registry); err != nil {
+			if credentials.IsErrCredentialsNotFoundMessage(err.Error()) {
+				logrus.Debugf("Not logged in to %s with credential helper %s", registry, helper)
+			}
+			continue
+		}
+		if err = deleteAuthFromCredHelper(helper, registry); err != nil {
+			return err
+		}
+		logrus.Debugf("Credentials were deleted from credential helper %s", helper)
+		return nil
+	}
+
 	return modifyJSON(sys, func(auths *dockerConfigFile) (bool, error) {
 		// First try cred helpers.
 		if ch, exists := auths.CredHelpers[registry]; exists {
@@ -258,6 +329,24 @@ func RemoveAuthentication(sys *types.SystemContext, registry string) error {
 
 // RemoveAllAuthentication deletes all the credentials stored in auth.json and kernel keyring
 func RemoveAllAuthentication(sys *types.SystemContext) error {
+	helpers, err := sysregistriesv2.AllConfiguredCredentialHelpers(sys)
+	if err != nil {
+		return err
+	}
+	for _, helper := range helpers {
+		creds, err := listAuthFromCredHelper(helper)
+		if err != nil {
+			return err
+		}
+		for registry := range creds {
+			if err = deleteAuthFromCredHelper(helper, registry); err != nil {
+				logrus.Warningf("error deleting credentials for %s from %s: %v", registry, helper, err)
+			} else {
+				logrus.Debugf("Credentials for %s were deleted from credential helper %s", registry, helper)
+			}
+		}
+	}
+
 	return modifyJSON(sys, func(auths *dockerConfigFile) (bool, error) {
 		if enableKeyring {
 			err := removeAllAuthFromKernelKeyring()
@@ -267,13 +356,26 @@ func RemoveAllAuthentication(sys *types.SystemContext) error {
 			}
 			logrus.Debugf("error removing credentials from kernel keyring")
 		}
+		for registry, helper := range auths.CredHelpers {
+			if err = deleteAuthFromCredHelper(helper, registry); err != nil {
+				logrus.Warningf("error deleting credentials for %s from %s: %v", registry, helper, err)
+			} else {
+				logrus.Debugf("Credentials for %s were deleted from credential helper %s", registry, helper)
+			}
+		}
 		auths.CredHelpers = make(map[string]string)
 		auths.AuthConfigs = make(map[string]dockerAuthConfig)
 		return true, nil
 	})
 }
 
-// getPathToAuth gets the path of the auth.json file used for reading and writing credentials
+func listAuthFromCredHelper(credHelper string) (map[string]string, error) {
+	helperName := fmt.Sprintf("docker-credential-%s", credHelper)
+	p := helperclient.NewShellProgramFunc(helperName)
+	return helperclient.List(p)
+}
+
+// getPathToAuth gets the path of the auth.json file used for reading and writting credentials
 // returns the path, and a bool specifies whether the file is in legacy format
 func getPathToAuth(sys *types.SystemContext) (string, bool, error) {
 	return getPathToAuthWithOS(sys, runtime.GOOS)
@@ -387,14 +489,17 @@ func modifyJSON(sys *types.SystemContext, editor func(auths *dockerConfigFile) (
 	return nil
 }
 
-func getAuthFromCredHelper(credHelper, registry string) (string, string, error) {
+func getAuthFromCredHelper(credHelper, registry string) (types.DockerAuthConfig, error) {
 	helperName := fmt.Sprintf("docker-credential-%s", credHelper)
 	p := helperclient.NewShellProgramFunc(helperName)
 	creds, err := helperclient.Get(p, registry)
 	if err != nil {
-		return "", "", err
+		return types.DockerAuthConfig{}, err
 	}
-	return creds.Username, creds.Secret, nil
+	return types.DockerAuthConfig{
+		Username: creds.Username,
+		Password: creds.Secret,
+	}, nil
 }
 
 func setAuthToCredHelper(credHelper, registry, username, password string) error {
@@ -423,15 +528,7 @@ func findAuthentication(registry, path string, legacyFormat bool) (types.DockerA
 
 	// First try cred helpers. They should always be normalized.
 	if ch, exists := auths.CredHelpers[registry]; exists {
-		username, password, err := getAuthFromCredHelper(ch, registry)
-		if err != nil {
-			return types.DockerAuthConfig{}, err
-		}
-
-		return types.DockerAuthConfig{
-			Username: username,
-			Password: password,
-		}, nil
+		return getAuthFromCredHelper(ch, registry)
 	}
 
 	// I'm feeling lucky
--- a/pkg/sysregistriesv2/system_registries_v2.go
+++ b/pkg/sysregistriesv2/system_registries_v2.go
@@ -78,6 +78,9 @@ type Registry struct {
 	// effectively be pulled from "example.com/foo/bar/myimage:latest".
 	// If no Prefix is specified, it defaults to the specified location.
 	Prefix string `toml:"prefix"`
+	// The credential helper for specified registry
+	// global default CredentialHelpers is used if CredentialHelper is not specified
+	CredentialHelper string `toml:"credential-helper,omitempty"`
 	// A registry is an Endpoint too
 	Endpoint
 	// The registry's mirrors.
@@ -151,7 +154,9 @@ func (config *V1RegistriesConf) Nonempty() bool {
 
 // V2RegistriesConf is the sysregistries v2 configuration format.
 type V2RegistriesConf struct {
-	Registries []Registry `toml:"registry"`
+	// The credential store for registries
+	CredentialHelpers []string   `toml:"credential-helpers"`
+	Registries        []Registry `toml:"registry"`
 	// An array of host[:port] (not prefix!) entries to use for resolving unqualified image references
 	UnqualifiedSearchRegistries []string `toml:"unqualified-search-registries"`
 
@@ -299,6 +304,11 @@ func (config *V2RegistriesConf) postProcessRegistries() error {
 			if err != nil {
 				return err
 			}
+		}
+
+		// ignore the credHelper if it is set for inner namespace
+		if reg.CredentialHelper != "" && strings.Contains(reg.Prefix, "/") {
+			return &InvalidRegistries{s: fmt.Sprintf("credential-helper can only be configured for registry domain, not for inner namespaces %s", reg.Prefix)}
 		}
 
 		// make sure mirrors are valid
@@ -661,6 +671,51 @@ func GetShortNameMode(ctx *types.SystemContext) (types.ShortNameMode, error) {
 		return -1, err
 	}
 	return config.shortNameMode, err
+}
+
+// AllConfiguredCredentialHelpers returns a list of all credential helpers that are configured for some registry
+// Almost all callers should use CredentialHelpersForRegistry to get configuration applicable to a specific registry.
+func AllConfiguredCredentialHelpers(sys *types.SystemContext) ([]string, error) {
+	helpers := make(map[string]bool)
+	config, err := getConfig(sys)
+	if err != nil {
+		return nil, err
+	}
+	for _, store := range config.partialV2.CredentialHelpers {
+		helpers[store] = true
+	}
+	for _, reg := range config.partialV2.Registries {
+		if reg.CredentialHelper != "" {
+			if ok := helpers[reg.CredentialHelper]; !ok {
+				helpers[reg.CredentialHelper] = true
+			}
+		}
+	}
+
+	var result []string
+	for cred := range helpers {
+		result = append(result, cred)
+	}
+
+	return result, nil
+}
+
+// CredentialHelpersForRegistry returns a list of credential helpers to try to use for the given registry,
+// in the order to try.
+func CredentialHelpersForRegistry(sys *types.SystemContext, registry string) ([]string, error) {
+	config, err := getConfig(sys)
+	if err != nil {
+		return nil, err
+	}
+	for _, reg := range config.partialV2.Registries {
+		if reg.Prefix == registry && reg.CredentialHelper != "" {
+			return []string{reg.CredentialHelper}, nil
+		}
+	}
+	if len(config.partialV2.CredentialHelpers) > 0 {
+		return config.partialV2.CredentialHelpers, nil
+	}
+	return nil, nil
 }
 
 // refMatchesPrefix returns true iff ref,
--- a/docs/containers-registries.conf.5.md
+++ b/docs/containers-registries.conf.5.md
@@ -16,6 +16,8 @@ Container engines will use the `$HOME/.config/containers/registries.conf` if it 
 `unqualified-search-registries`
 : An array of _host_[`:`_port_] registries to try when pulling an unqualified image, in order.
 
+`credential-helpers`: An array of credential helpers used as external credential store for the registries.
+
 ### NAMESPACED `[[registry]]` SETTINGS
 
 The bulk of the configuration is represented as an array of `[[registry]]`
@@ -50,6 +52,9 @@ Given an image name, a single `[[registry]]` TOML table is chosen based on its `
 `blocked`
 : `true` or `false`.
     If `true`, pulling images with matching names is forbidden.
+
+`credential-helper`
+: The credential helper for registry specified by the `prefix` field.  Global default `credential-helpers` is used if `credential-helper` is not specified. This configuration only works if the `prefix` field is a registry domain, not used for inner namespace.
 
 #### Remapping and mirroring registries
 
--- a/pkg/docker/config/config_test.go
+++ b/pkg/docker/config/config_test.go
@@ -97,6 +97,17 @@ func TestGetAuth(t *testing.T) {
 		os.Setenv("XDG_RUNTIME_DIR", origXDG)
 	}()
 
+	// override PATH for executing credHelper
+	curtDir, err := os.Getwd()
+	require.NoError(t, err)
+	origPath := os.Getenv("PATH")
+	newPath := fmt.Sprintf("%s:%s", filepath.Join(curtDir, "testdata"), origPath)
+	os.Setenv("PATH", newPath)
+	t.Logf("using PATH: %q", newPath)
+	defer func() {
+		os.Setenv("PATH", origPath)
+	}()
+
 	tmpHomeDir, err := ioutil.TempDir("", "test_docker_client_get_auth")
 	if err != nil {
 		t.Fatal(err)
@@ -236,6 +247,17 @@ func TestGetAuth(t *testing.T) {
 				name:     "match none (empty.json)",
 				hostname: "https://localhost:5000",
 				path:     filepath.Join("testdata", "empty.json"),
+			},
+			{
+				name:     "credhelper from registries.conf",
+				hostname: "registry-a.com",
+				sys: &types.SystemContext{
+					SystemRegistriesConfPath: filepath.Join("testdata", "cred-helper.conf"),
+				},
+				expected: types.DockerAuthConfig{
+					Username: "foo",
+					Password: "bar",
+				},
 			},
 		} {
 			t.Run(tc.name, func(t *testing.T) {
@@ -474,7 +496,22 @@ func TestGetAllCredentials(t *testing.T) {
 	err = tmpFile.Close()
 	require.NoError(t, err)
 	authFilePath := tmpFile.Name()
-	sys := types.SystemContext{AuthFilePath: authFilePath}
+	// override PATH for executing credHelper
+	path, err := os.Getwd()
+	require.NoError(t, err)
+	origPath := os.Getenv("PATH")
+	newPath := fmt.Sprintf("%s:%s", filepath.Join(path, "testdata"), origPath)
+	os.Setenv("PATH", newPath)
+	t.Logf("using PATH: %q", newPath)
+	defer func() {
+		os.Setenv("PATH", origPath)
+	}()
+	err = os.Chmod(filepath.Join(path, "testdata", "docker-credential-helper-registry"), os.ModePerm)
+	require.NoError(t, err)
+	sys := types.SystemContext{
+		AuthFilePath:             authFilePath,
+		SystemRegistriesConfPath: filepath.Join("testdata", "cred-helper.conf"),
+	}
 
 	data := []struct {
 		server   string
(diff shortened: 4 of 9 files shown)


pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -78,6 +78,8 @@ type Registry struct {
 	// effectively be pulled from "example.com/foo/bar/myimage:latest".
 	// If no Prefix is specified, it defaults to the specified location.
 	Prefix string `toml:"prefix"`
+	// The credential store for the registry
+	CredHelper string `toml:"cred_helper"`
```
mtrmac, line 82: If we are looking up credentials by registry host names, we should reject configurations that set this on inner namespaces.

Alternatively, we may need credentials per smaller namespaces (there’s a worrisome case of needing different Quay.io credentials for private repositories and for pulling OpenShift), but then `docker/config` API needs to be extended to accept full Docker references instead of just registry names.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -570,6 +574,15 @@ func UnqualifiedSearchRegistries(ctx *types.SystemContext) ([]string, error) {
 	return config.UnqualifiedSearchRegistries, nil
 }
 
+// GetCredsStore return the list of credetntial store configed by creds_store for registries
+func GetCredsStore(ctx *types.SystemContext) ([]string, error) {
```
mtrmac, line 578: No, the API should, if at all possible, give the user the effective configuration, not the raw pieces.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -151,6 +153,8 @@ func (config *V1RegistriesConf) Nonempty() bool {
 
 // V2RegistriesConf is the sysregistries v2 configuration format.
 type V2RegistriesConf struct {
+	// The credential store for registries
+	CredsStore []string   `toml:"creds_store"`
```
mtrmac, line 157: Why is this a list? How are the users expected to use that, what happens on errors?

pkg/docker/config/config.go
```diff
@@ -353,6 +494,31 @@ func modifyJSON(sys *types.SystemContext, editor func(auths *dockerConfigFile) (
 	return nil
 }
 
+// errCredNotConfigured indicates the credential service for the registry was not configured
+var errCredNotConfigured = errors.New("Credential service configuration not found")
+
+// getRegistryCredConfigs returns the credHelper configed for the registry
+// creds_store will not be used if cred_helper configured for specified registries
+func getRegistryCredConfigs(sys *types.SystemContext, registry string) ([]string, error) {
```
mtrmac, line 502: This should be a function in `sysregistriesv2`, so that `sysregistriesv2` is in control of the configuration format and semantics; not private logic that this subpackage imposes on `sysregistriesv2`, and that any other consumer would have to duplicate.

Also, the documentation of the function should say what it returns / how it should be used, not the details of the configuration semantics.

mtrmac, line 502: Naming: `config` does not really say anything specific; saying `CredentialHelper[s]` in the name would be useful.

pkg/docker/config/config.go
```diff
@@ -353,6 +494,31 @@ func modifyJSON(sys *types.SystemContext, editor func(auths *dockerConfigFile) (
...
+	credsStore, err := sysregistriesv2.GetCredsStore(sys)
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to parse the registries.conf file")
+	}
+	if len(credsStore) > 0 {
+		return credsStore, nil
```
mtrmac, line 508: The documentation above says per-registry configuration overrides the global one.

pkg/docker/config/config.go
```diff
@@ -353,6 +494,31 @@ func modifyJSON(sys *types.SystemContext, editor func(auths *dockerConfigFile) (
...
+	registries, err := sysregistriesv2.GetRegistries(sys)
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to parse the registries.conf file")
+	}
+	for _, reg := range registries {
+		if reg.Prefix == registry {
+			return []string{reg.CredHelper}, nil
+		}
+	}
+	return nil, errCredNotConfigured
```
mtrmac, line 519: How does this differ from just returning `nil, nil`?

pkg/sysregistriesv2/system_registries_v2.go
QiWang19: Initially, I thought this can match docker document, this configuration supports a specified registry domain. So to not set for an inner namespace, this configuration use substring before the first '/' of `Prefix` as the registry name?

pkg/sysregistriesv2/system_registries_v2.go
QiWang19: The design mentioned `failover list that would attempt to use each credential helper until one works. https://docs.google.com/document/d/1mTPI9GCvy3C_c11Z1MNQzOR6N_lMBVLHW3gpOllC6rw/edit
Changed to a string type. Maybe the error can be silently ignored this way, the usage seems underministic. I am ok to not use the list, like the credsStore in config.json

pkg/sysregistriesv2/system_registries_v2.go
QiWang19: What are effective configurations? Just want this API to export the `CredsStore` field to pkg/config/config.go

mtrmac: If we export that field _individually_, we can’t easily change it or replace it; compare how we now can’t return a value exactly as documented for `ConfigDirPath` 

Instead, give the user an API to ask a specific question and receive a single response, like `FindRegistry` that returns effective configuration for a single input (which might never be physically present in any config file), or something like `getRegistryCredConfigs` in this case.

We might _eventually_ need to export the raw data, but let’s not commit to that API until there’s a clear need.

pkg/sysregistriesv2/system_registries_v2.go
mtrmac: Just stripping namespaces from `Prefix` could result in ambiguities if there are different configurations for different namespaces of a single registry.

Instead, `postProcess` should validate the rules, and reject configurations that contain a `/`.

pkg/sysregistriesv2/system_registries_v2.go
mtrmac: Thanks, I’ll return to that document.

pkg/docker/config/config.go
```diff
@@ -55,7 +56,27 @@ var (
 
 // SetAuthentication stores the username and password in the auth.json file
 func SetAuthentication(sys *types.SystemContext, registry, username, password string) error {
+	credConfigs, err := sysregistriesv2.GetRegistryCredentialHelpers(sys, registry)
+	if err != nil {
+		return err
+	}
+	if len(credConfigs) > 0 {
+		if err := func(credConfigs []string, registry, username, password string) error {
```
mtrmac, line 64: I don’t understand why the `func() ()` parts exist. Wouldn’t it be simpler to inline fully? Even the `len(credConfigs) > 0` check would go away.

pkg/docker/config/config.go
```diff
@@ -55,7 +56,27 @@ var (
...
+			for _, credConfig := range credConfigs {
+				if err = setAuthToCredHelper(credConfig, registry, username, password); err != nil {
```
mtrmac, line 66: So this should (log? and) try the next helper if this one is not set up. Right now the second or any further helpers are never tried at all — and if any fail, the errors are silently ignored.

pkg/docker/config/config.go
```diff
@@ -55,7 +56,27 @@ var (
 
 // SetAuthentication stores the username and password in the auth.json file
 func SetAuthentication(sys *types.SystemContext, registry, username, password string) error {
+	credConfigs, err := sysregistriesv2.GetRegistryCredentialHelpers(sys, registry)
+	if err != nil {
+		return err
+	}
+	if len(credConfigs) > 0 {
+		if err := func(credConfigs []string, registry, username, password string) error {
+			for _, credConfig := range credConfigs {
```
mtrmac, line 65:
```suggestion
			for _, helper := range helpers {
```
That’s what it is, isn’t it? There isn’t any other configuration.

pkg/docker/config/config.go
```diff
@@ -104,15 +125,14 @@ func GetAllCredentials(sys *types.SystemContext) (map[string]types.DockerAuthCon
 
 		// Credential helpers may override credentials from the auth file.
 		for registry, credHelper := range auths.CredHelpers {
-			username, password, err := getAuthFromCredHelper(credHelper, registry)
+			conf, err := getAuthFromCredHelper(credHelper, registry)
```
mtrmac, line 128: `GetAllCredentials` needs to return the credentials from the `registryies.conf`-configured helpers as well, I think. (@vrothberg ?)

pkg/docker/config/config.go
```diff
@@ -55,7 +56,27 @@ var (
 
 // SetAuthentication stores the username and password in the auth.json file
```
mtrmac, line 57: Throughout, the documentation should be updated to say that the credentials are stored/read/…, but not refer to any specific files.

pkg/docker/config/config.go
```diff
@@ -152,6 +172,30 @@ func GetCredentials(sys *types.SystemContext, registry string) (types.DockerAuth
 		return *sys.DockerAuthConfig, nil
 	}
 
+	credConfigs, err := sysregistriesv2.GetRegistryCredentialHelpers(sys, registry)
+	if err != nil {
+		return types.DockerAuthConfig{}, err
+	}
+	if len(credConfigs) > 0 {
+		if auth, err := func(credConfigs []string, registry string) (types.DockerAuthConfig, error) {
```
mtrmac, line 180: Why the `func() ()`?

pkg/docker/config/config.go
```diff
@@ -152,6 +172,30 @@ func GetCredentials(sys *types.SystemContext, registry string) (types.DockerAuth
...
+			var lastError error
+			for _, credConfig := range credConfigs {
+				dockerAuthConfig, err := getAuthFromCredHelper(credConfig, registry)
```
mtrmac, line 183: So this should (log? and) try the next helper if this one is not set up.

pkg/docker/config/config.go
```diff
@@ -152,6 +172,30 @@ func GetCredentials(sys *types.SystemContext, registry string) (types.DockerAuth
...
+				if err != nil {
+					lastError = err
+					if credentials.IsErrCredentialsNotFoundMessage(err.Error()) {
```
mtrmac, line 186: `lastError` currently is always `IsErrCredentialsNotFoundMessage` — instead, `lastError` should probably be set only if `!credentials.IsErrCredentialsNotFoundMessage`. `credentials.IsErrCredentialsNotFoundMessage` should result in (mostly?) silence and trying other locations.

pkg/docker/config/config.go
```diff
@@ -152,6 +172,30 @@ func GetCredentials(sys *types.SystemContext, registry string) (types.DockerAuth
...
 	}
 
+	credConfigs, err := sysregistriesv2.GetRegistryCredentialHelpers(sys, registry)
+	if err != nil {
+		return types.DockerAuthConfig{}, err
+	}
+	if len(credConfigs) > 0 {
+		if auth, err := func(credConfigs []string, registry string) (types.DockerAuthConfig, error) {
+			var lastError error
+			for _, credConfig := range credConfigs {
```
mtrmac, line 182:
```suggestion
			for _, helper := range helpers {
```

pkg/docker/config/config.go
mtrmac, line 183: (Absolutely non-blocking: `dockerAuthConfig` can be just `authConfig`, `credentials` or even `creds`.)

pkg/docker/config/config.go
```diff
@@ -201,6 +245,33 @@ func GetAuthentication(sys *types.SystemContext, registry string) (string, strin
 
 // RemoveAuthentication deletes the credentials stored in auth.json
 func RemoveAuthentication(sys *types.SystemContext, registry string) error {
+	credConfigs, err := sysregistriesv2.GetRegistryCredentialHelpers(sys, registry)
+	if err != nil {
+		return err
+	}
+	if len(credConfigs) > 0 {
+		if err := func(credConfigs []string, registry string) error {
```
mtrmac, line 253: Why the `func() ()`?

pkg/docker/config/config.go
```diff
@@ -201,6 +245,33 @@ func GetAuthentication(sys *types.SystemContext, registry string) (string, strin
...
+			var err error
+			for _, credConfig := range credConfigs {
+				if _, err = getAuthFromCredHelper(credConfig, registry); err != nil {
```
mtrmac, line 256: So this should (log? and) try the next helper if this one is not set up.

pkg/docker/config/config.go
```diff
@@ -201,6 +245,33 @@ func GetAuthentication(sys *types.SystemContext, registry string) (string, strin
...
+					if credentials.IsErrCredentialsNotFoundMessage(err.Error()) {
```
mtrmac, line 257: This is a good reason to try the next helper, not to abort the loop.

pkg/docker/config/config.go
```diff
@@ -201,6 +245,33 @@ func GetAuthentication(sys *types.SystemContext, registry string) (string, strin
...
 // RemoveAuthentication deletes the credentials stored in auth.json
 func RemoveAuthentication(sys *types.SystemContext, registry string) error {
+	credConfigs, err := sysregistriesv2.GetRegistryCredentialHelpers(sys, registry)
+	if err != nil {
+		return err
+	}
+	if len(credConfigs) > 0 {
+		if err := func(credConfigs []string, registry string) error {
+			var err error
+			for _, credConfig := range credConfigs {
```
mtrmac, line 255:
```suggestion
			for _, helper := range helpers {
```

pkg/docker/config/config.go
```diff
@@ -230,6 +301,16 @@ func RemoveAuthentication(sys *types.SystemContext, registry string) error {
...
+	}
+	if len(credConfigs) > 0 {
+		if err := removeAllAuthFromCredConfigs(credConfigs); err == nil {
+			logrus.Debug("Credentials were deleted from credential helper")
+		}
+	}
+
 	return modifyJSON(sys, func(auths *dockerConfigFile) (bool, error) {
 		if enableKeyring {
 			err := removeAllAuthFromKernelKeyring()
```
mtrmac, line 316: I think it would be consistent to also remove the relevant credentials from the helpers configured in `auths.CredHelpers` below.

pkg/docker/config/config.go
```diff
@@ -245,6 +326,48 @@ func RemoveAllAuthentication(sys *types.SystemContext) error {
 	})
 }
 
+func removeAuthFromCredConfigs(credConfigs []string, registry string) error {
```
mtrmac, line 329: This is unused AFAICS.

pkg/docker/config/config.go
```diff
@@ -245,6 +326,48 @@ func RemoveAllAuthentication(sys *types.SystemContext) error {
...
+		if err = deleteAuthFromCredHelper(credConfig, registry); err != nil {
+			return err
+		}
+		logrus.Debugf("Credentials were deleted from credential helper %s", credConfig)
+		return nil
+	}
+	return err
+}
+
+func removeAllAuthFromCredConfigs(credConfigs []string) error {
```
mtrmac, line 348: This might also be small enough to be inlined.

pkg/docker/config/config.go
```diff
@@ -245,6 +326,48 @@ func RemoveAllAuthentication(sys *types.SystemContext) error {
...
+	var err error
+	for _, credHelper := range credConfigs {
```
mtrmac, line 350:
```suggestion
	for _, helper := range helpers {
```

pkg/docker/config/config.go
```diff
@@ -245,6 +326,48 @@ func RemoveAllAuthentication(sys *types.SystemContext) error {
...
+			return err
+		}
+		for registry := range creds {
+			if err = deleteAuthFromCredHelper(credHelper, registry); err != nil {
+				logrus.Warningf("error deleting credentials for %s from %s: %v", registry, credHelper, err)
+			}
+		}
+	}
+
+	return err
```
mtrmac, line 362: This currently fairly randomly returns an error or `nil`, depending on what happened last. That doesn’t make much sense.

I’m not quite sure how to handle this _well_, though — listing every single credential that failed might be dozens or hundreds in pathological cases. But it’s probably a reasonable starting point — let’s not lose data, and we can maybe somehow summarize later if it turns out to be necessary.

pkg/docker/config/config.go
```diff
@@ -389,15 +515,12 @@ func findAuthentication(registry, path string, legacyFormat bool) (types.DockerA
...
 		if err != nil {
-			return types.DockerAuthConfig{}, err
+			return dockerAuthConfig, err
 		}
 
-		return types.DockerAuthConfig{
-			Username: username,
-			Password: password,
-		}, nil
+		return dockerAuthConfig, nil
```
mtrmac, line 523: This is just `return getAuthFromCredHelper(…)`.

pkg/sysregistriesv2/system_registries_v2.go
mtrmac, line 157:
```suggestion
	CredentialHelpers []string   `toml:"credential-helpers"`
```
please.

pkg/sysregistriesv2/system_registries_v2.go
mtrmac, line 82:
```suggestion
	CredentialHelper string `toml:"credential-helper"`
```
we use dashes elsewhere in this file already.

Also, document that the global default is used if `CredentialHelper` is not specified, and use `omitempty`.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -275,6 +279,11 @@ func (config *V2RegistriesConf) postProcess() error {
 			}
 		}
 
+		// ignore the credHelper if it is set for inner namespace
+		if reg.CredHelper != "" && strings.Contains(reg.Prefix, "/") {
+			reg.CredHelper = ""
+		}
```
mtrmac, line 285: This should report an error instead of silently ignoring invalid input.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -570,6 +579,66 @@ func UnqualifiedSearchRegistries(ctx *types.SystemContext) ([]string, error) {
 	return config.UnqualifiedSearchRegistries, nil
 }
 
+// GetCredsStore return the list of credetntial store configed by creds_store for registries
+func GetCredsStore(ctx *types.SystemContext) ([]string, error) {
```
mtrmac, line 583: At least until we have a caller, please remove this entirely.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -570,6 +579,66 @@ func UnqualifiedSearchRegistries(ctx *types.SystemContext) ([]string, error) {
...
+	config, err := getConfig(ctx)
+	if err != nil {
+		return nil, err
+	}
+	return config.CredsStore, nil
+}
+
+// GetCredConfigs returns a slice of credential stores configred in registries.conf
+// caller should use GetRegistryCredentialHelpers to get cred helper of specified registry
+func GetCredConfigs(sys *types.SystemContext) ([]string, error) {
```
mtrmac, line 593:
```suggestion
// AllCredentialHelpers returns a list of all credential helpers that are configured for some registry.
// Almost all callers should use GetRegistryCredentialHelpers to get configuration applicable to a specific registry.
func AllCredentialHelpers(sys *types.SystemContext) ([]string, error) {
```
or something like that `AllConfiguredCredentialHelpers`?. `Get` is not idiomatic in Go.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -570,6 +579,66 @@ func UnqualifiedSearchRegistries(ctx *types.SystemContext) ([]string, error) {
...
+	var (
+		credConfigs []string
+		credHelpers = make(map[string]bool)
```
mtrmac, line 596: Non-blocking? The naming distinction between `credConfigs` and `credHelpers`, both of which reference the same data, does not really make sense, especially when there isn’t any other kind of configuration. `helperList` and `helperSet` would at least be descriptive. Or it might be good enough to just have `helpers` (the set) and a very short-lived `res`/`result` variable declared only for the last loop — the shorter the scope of a symbol, the less descriptive/distinctive the name needs to be.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -570,6 +579,66 @@ func UnqualifiedSearchRegistries(ctx *types.SystemContext) ([]string, error) {
...
+	)
+	config, err := getConfig(sys)
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to parse the registries.conf file")
```
mtrmac, line 600: The error returned by `getConfig` is already fairly descriptive (notably it includes the _correct_ path), this `Wrapf` does not really help.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -570,6 +579,66 @@ func UnqualifiedSearchRegistries(ctx *types.SystemContext) ([]string, error) {
...
+	}
+	for _, store := range config.CredsStore {
+		credHelpers[store] = true
+	}
+	registries, err := GetRegistries(sys)
```
mtrmac, line 605: Just iterate `config.Registries` directly; this function already depends on the layout of `config`, and this way we avoid the lock in `getConfig`, don’t have to worry about using different snapshots of the configuration if it is read twice, and the case of `err != nil` vanishes.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -570,6 +579,66 @@ func UnqualifiedSearchRegistries(ctx *types.SystemContext) ([]string, error) {
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to parse the registries.conf file")
+	}
+	for _, reg := range registries {
+		if ok, _ := credHelpers[reg.CredHelper]; !ok {
```
mtrmac, line 610: Only if `reg.CredHelper != ""`

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -570,6 +579,66 @@ func UnqualifiedSearchRegistries(ctx *types.SystemContext) ([]string, error) {
...
+	}
+	for cred := range credHelpers {
+		credConfigs = append(credConfigs, cred)
+	}
+	return credConfigs, nil
+}
+
+// GetRegistryCredentialHelpers returns the credHelper configured for given registry
+// credHelper configured for the registry can override list of global configuration
+func GetRegistryCredentialHelpers(sys *types.SystemContext, registry string) ([]string, error) {
```
mtrmac, line 622:
```suggestion
// CredentialHelpersForRegistry returns a list of credential helpers to try to use for the given registry,
// in the order to try.
func CredentialHelpersForRegistry(sys *types.SystemContext, registry string) ([]string, error) {
```

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -570,6 +579,66 @@ func UnqualifiedSearchRegistries(ctx *types.SystemContext) ([]string, error) {
...
+	registries, err := GetRegistries(sys)
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to parse the registries.conf file")
+	}
+	for _, reg := range registries {
+		if reg.CredHelper != "" && reg.Prefix == registry {
+			return []string{reg.CredHelper}, nil
+		}
+	}
+	config, err := getConfig(sys)
```
mtrmac, line 632: Same as elsewhere, please use `getConfig.Registries` instead of a separate `GetRegistries` call.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -570,6 +579,66 @@ func UnqualifiedSearchRegistries(ctx *types.SystemContext) ([]string, error) {
...
+
+// GetRegistryCredentialHelpers returns the credHelper configured for given registry
+// credHelper configured for the registry can override list of global configuration
+func GetRegistryCredentialHelpers(sys *types.SystemContext, registry string) ([]string, error) {
+	registries, err := GetRegistries(sys)
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to parse the registries.conf file")
+	}
+	for _, reg := range registries {
+		if reg.CredHelper != "" && reg.Prefix == registry {
```
mtrmac, line 628: Please put the `reg.Prefix` check (“is this the right entry?”) before the `CredHelper != ""` one (“does the entry specify a configuration?”)

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -570,6 +579,66 @@ func UnqualifiedSearchRegistries(ctx *types.SystemContext) ([]string, error) {
...
+		}
+	}
+	config, err := getConfig(sys)
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to parse the registries.conf file")
+	}
+	if len(config.CredsStore) > 0 {
+		return config.CredsStore, nil
+	}
+	return nil, nil
```
mtrmac, line 639: (We might eventually need the `file` fallback here.)

mtrmac:
Eventually, it would be useful to have tests for all of this; is definitely important enough. But that can certainly wait until after the structure of the code, and the details of the design, settle.

pkg/docker/config/config.go
QiWang19: is it still necessary to check `credentials.IsErrCredentialsNotFoundMessage`? Can it just log "Not logged in with helper" and continue?

mtrmac: Ideally we should be fairly quiet about “not found” and “helper not set up”, but at least somewhat louder about any other kinds of errors; it might be even OK to outright fail for other kinds of errors, if we could reliably detect the “not set up” case and not fail on that one.

pkg/sysregistriesv2/system_registries_v2.go
QiWang19: Prefer to add  ["file"] configuration in a second PR.  So if there's a "file" in the list, return the original list. If "file" does not exist, append it to the end of the list and can fall back to file when dealing with the credential configs?

mtrmac: No — if the user configures _only_ a desktop keyring, we should not be overriding that decision and writing to files. Only if `registries.conf`/`registries.d/*` don’t contain _any_ helper configuration, we would implicitly assume `["file"]`.

(With only this PR, we do always fall back to files, because we first use the helpers, and then always fall back to `readJSON`/`modifyJSON`, but that other PR would change that so that the JSON code only runs if `file` is specified.)

QiWang19:
@mtrmac PTAL

QiWang19:
@mtrmac PTAL. I added some tests in a separate commit. The CI fails with `test-skopeo`, is it a regression caused by the retry PR?

QiWang19:
@mtrmac PTAL.

mtrmac:
> The CI fails with `test-skopeo`, is it a regression caused by bumpping go-multierror from 1.0.0 to 1.1.0?

Yes; https://github.com/containers/image/pull/987 should fix that.

QiWang19:
@mtrmac PTAL

pkg/docker/config/config.go
mtrmac: Outstanding / depends on “helper installed but not set up” detection.

pkg/docker/config/config.go
mtrmac: Outstanding / depends on “helper installed but not set up” detection.

pkg/docker/config/config.go
mtrmac: Outstanding / depends on “helper installed but not set up” detection.

pkg/docker/config/config.go
mtrmac: Outstanding / depends on “helper installed but not set up” detection.

pkg/docker/config/config.go
mtrmac: Most of the functions still need updated documentation.

TomSweeneyRedHat:
Are there any changes to man pages with this?  Or will that be in a separate PR?

QiWang19:
Yes, containers-registries.conf.5.md needs update.

QiWang19:
@mtrmac PTAL

QiWang19:
we should be able to detect the error gnome-keyring not installed or keyring locked, and fall back to current auth files or use ["file"] as credential-helper after implemented in the future PR, does this make sense?

rhatdan:
Yes that makes sense.

mtrmac:
> how can I make `gnome-keyring` not installed, uninstall `gnome-keyring`?

Yes, just `dnf remove gnome-keyring`. It might not be possible on a full desktop system.

QiWang19:
Can the set/get authentication fallback to current authfile when the returned error is not handled by the docker-credential-helpers? If the credential-helper is configured, log the error to debug level and fallback.
Screenshot from 2020-08-28 13-28-11
Screenshot from 2020-08-28 13-38-20

rhatdan:
@QiWang19 Needs a rebase.

QiWang19:
Draft a branch in docker-credential-helpers https://github.com/docker/docker-credential-helpers/compare/master...QiWang19:list-error?expand=1, will this upstream fix make sense?
On the system without `gnome-keyring` installed, `docker-credential-secretservice list`,  the error is `The name is not activatable`.

QiWang19:
Is this a blocker of this PR https://github.com/docker/docker-credential-helpers/pull/191? What else we need to get this PR in?

umohnani8:
@mtrmac @rhatdan can this be completed or is there something else blocking it? (apart from the rebase requirement)

rhatdan:
I would love to get this in.
@QiWang19 Please rebase and get this to pass tests.

pkg/docker/config/config.go
```diff
@@ -104,21 +119,41 @@ func GetAllCredentials(sys *types.SystemContext) (map[string]types.DockerAuthCon
...
+			return nil, err
+		}
+		for registry := range creds {
+			conf, err := getAuthFromCredHelper(helper, registry)
+			if err != nil {
+				return nil, err
+			}
+			value, ok := authConfigs[normalizeRegistry(registry)]
+			if ok && value != conf {
+				return nil, err
```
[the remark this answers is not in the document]
rhatdan: @QiWang19 Has this change been made?

pkg/docker/config/config.go
```diff
@@ -201,6 +253,24 @@ func GetAuthentication(sys *types.SystemContext, registry string) (string, strin
...
+	helpers, err := sysregistriesv2.CredentialHelpersForRegistry(sys, registry)
+	if err != nil {
+		return err
+	}
+	for _, helper := range helpers {
+		if _, err = getAuthFromCredHelper(helper, registry); err != nil {
+			if credentials.IsErrCredentialsNotFoundMessage(err.Error()) {
+				logrus.Warnf("Not logged in to %s with credential helper %s", registry, helper)
+			}
+			continue
```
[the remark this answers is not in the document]
rhatdan: @QiWang19 Could you add the warning?

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -570,6 +580,51 @@ func UnqualifiedSearchRegistries(ctx *types.SystemContext) ([]string, error) {
 	return config.UnqualifiedSearchRegistries, nil
 }
 
+// AllConfiguredCredentialHelpers returns a list of all credential helpers that are configred for some registry
```
rhatdan, line 583: Spelling configred

pkg/sysregistriesv2/system_registries_v2.go
rhatdan: I agree with @mtrmac

rhatdan:
LGTM

QiWang19:
@mtrmac PTAL

umohnani8:
@vrothberg @mtrmac @rhatdan  PTAL, would like to get this moving along.
@QiWang19 this may need a rebase.

rhatdan:
@umohnani8 Could you take over the PR and get it rebased.

umohnani8:
Looks like @QiWang19 is still following up with this.

QiWang19:
No new reviews for a long time, is this PR a feasible solution we want to avoid keeping the solution in plaintext? Should I close this PR?

rhatdan:
@vrothberg PTAL

pkg/docker/config/config.go
```diff
@@ -258,6 +329,24 @@ func RemoveAuthentication(sys *types.SystemContext, registry string) error {
 
 // RemoveAllAuthentication deletes all the credentials stored in auth.json and kernel keyring
 func RemoveAllAuthentication(sys *types.SystemContext) error {
+	helpers, err := sysregistriesv2.AllConfiguredCredentialHelpers(sys)
```
vrothberg, line 332: Same here.  We need to collect errors and try to remove all credentials from all possible sources.

We should not return early in error cases.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -151,7 +154,9 @@ func (config *V1RegistriesConf) Nonempty() bool {
 
 // V2RegistriesConf is the sysregistries v2 configuration format.
 type V2RegistriesConf struct {
-	Registries []Registry `toml:"registry"`
+	// The credential store for registries
+	CredentialHelpers []string   `toml:"credential-helpers"`
```
vrothberg, line 158: Please move it below `Registries` or add a comment to it.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -301,6 +306,11 @@ func (config *V2RegistriesConf) postProcessRegistries() error {
 			}
 		}
 
+		// ignore the credHelper if it is set for inner namespace
```
vrothberg, line 309: Comment needs a change since we're not ignoring it but error out.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -301,6 +306,11 @@ func (config *V2RegistriesConf) postProcessRegistries() error {
...
+		if reg.CredentialHelper != "" && strings.Contains(reg.Prefix, "/") {
+			return &InvalidRegistries{s: fmt.Sprintf("credential-helper can only be configured for registry domain, not for inner namespaces %s", reg.Prefix)}
```
vrothberg, line 311:
```suggestion
			return &InvalidRegistries{s: fmt.Sprintf("credential-helper can only be configured for a registry domain, not for an inner namespaces or an image: %s", reg.Prefix)}
```

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -301,6 +306,11 @@ func (config *V2RegistriesConf) postProcessRegistries() error {
 			}
 		}
 
+		// ignore the credHelper if it is set for inner namespace
+		if reg.CredentialHelper != "" && strings.Contains(reg.Prefix, "/") {
```
vrothberg, line 310: @QiWang19, can you elaborate _why_ we have that restriction?

The `strings.Contains(reg.Prefix, "/")` check is an overestimation since a registry _may_ be running on domain.com/registry.  I haven't seen it yet but it's technically possible so I want to make sure to fully understand the restriction.  We may need to mention the reasoning also in the man page.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -663,6 +673,51 @@ func GetShortNameMode(ctx *types.SystemContext) (types.ShortNameMode, error) {
...
+	}
+	for _, reg := range config.partialV2.Registries {
+		if reg.CredentialHelper != "" {
+			if ok := helpers[reg.CredentialHelper]; !ok {
+				helpers[reg.CredentialHelper] = true
+			}
+		}
+	}
+
+	var result []string
```
vrothberg, line 695: Maps in golang are non-deterministic, so we have to move `result` up and append to it in both `for` loops. That will preserve the order _and_ make it deterministic.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -663,6 +673,51 @@ func GetShortNameMode(ctx *types.SystemContext) (types.ShortNameMode, error) {
 	return config.shortNameMode, err
 }
 
+// AllConfiguredCredentialHelpers returns a list of all credential helpers that are configured for some registry
```
vrothberg, line 676:
```suggestion
// AllConfiguredCredentialHelpers returns a list of all credential helpers including the global credential helpers and the ones configured for specific registries in the given order.
```

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -663,6 +673,51 @@ func GetShortNameMode(ctx *types.SystemContext) (types.ShortNameMode, error) {
...
+}
+
+// CredentialHelpersForRegistry returns a list of credential helpers to try to use for the given registry,
+// in the order to try.
+func CredentialHelpersForRegistry(sys *types.SystemContext, registry string) ([]string, error) {
+	config, err := getConfig(sys)
+	if err != nil {
+		return nil, err
+	}
+	for _, reg := range config.partialV2.Registries {
```
vrothberg, line 710: Please use `FindRegistry` instead.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -663,6 +673,51 @@ func GetShortNameMode(ctx *types.SystemContext) (types.ShortNameMode, error) {
...
+	var result []string
+	for cred := range helpers {
+		result = append(result, cred)
+	}
+
+	return result, nil
+}
+
+// CredentialHelpersForRegistry returns a list of credential helpers to try to use for the given registry,
+// in the order to try.
```
vrothberg, line 704:
```suggestion
// CredentialHelpersForRegistry returns the configured credential helper for the specified registry or the global credential helpers in order.
```

vrothberg (commented):
Apologies for the late review.

Mostly nits but I have one question on the design.  Currently, we can configure exactly one credhelper per registry but we can configure multiple global ones.  Should we leave the door open for more than one helper per registry and turn the `string` into a `[]string`?

I have absolutely no feeling for it but want to point out that `string` is not extensible.

rhatdan:
@vrothberg Could you take this over and drive it home.  I think @QiWang19 is going to be too busy with CRI-O issues.  is that correct @QiWang19 ?

QiWang19:
> @vrothberg Could you take this over and drive it home. I think @QiWang19 is going to be too busy with CRI-O issues. is that correct @QiWang19 ?

yes, @vrothberg feel free to take this over.

vrothberg:
> > @vrothberg Could you take this over and drive it home. I think @QiWang19 is going to be too busy with CRI-O issues. is that correct @QiWang19 ?
> 
> yes, @vrothberg feel free to take this over.

Will do. Thanks for working on this, @QiWang19 !

rhatdan:
@vrothberg Do you think we could get this in, and cut a new release?

vrothberg:
> @vrothberg Do you think we could get this in, and cut a new release?

When do you need it?

rhatdan:
Probably post podman 3.1 at this point.  But I would like to get this into podman and Buildah.

vrothberg:
> Probably post podman 3.1 at this point. But I would like to get this into podman and Buildah.

Excellent, that relaxes me a bit.  I want to get the `buildah source` PR out today and will find time tomorrow.

vrothberg:
Replaced by https://github.com/containers/image/pull/1193

Issue #972, opened 2020-06-28
"panic: assignment to entry in nil map" on SetAuthentication 


CooperLi:
**Description**

Modify `auth.json` fields leads to `panic: assignment to entry in nil map`

**Steps to reproduce the issue:**

1. success login image registry

2. change `auth.json` fileds

    ```
    $ cat auth.json           
    {
            "auths": {
                    "mydockerhub.io": {
                            "auth": "aGVsbG86d29ybGQK"
                    }
    }
    
    # after modified
    $ diff auth.json auth.json.bk
    2c2
    <       "auths": {
    ---
    >       "buths": {
    
    ```

3. continue login with same registry

4. panic

**Describe the results you received:**

```bash
panic: assignment to entry in nil map

goroutine 168 [running]:
github.com/containers/image/v5/pkg/docker/config.SetAuthentication.func1(0xc0004bc050, 0x1a, 0x12e6d00, 0x0)
	/home/GOWORK/src/cooper/cooper/vendor/github.com/containers/image/v5/pkg/docker/config/config.go:71 +0x1b2
github.com/containers/image/v5/pkg/docker/config.modifyJSON(0xc00046a8c0, 0xc000327970, 0xc37416, 0x0)
	/home/GOWORK/src/cooper/cooper/vendor/github.com/containers/image/v5/pkg/docker/config/config.go:276 +0x198
github.com/containers/image/v5/pkg/docker/config.SetAuthentication(0xc00046a8c0, 0xc000474b80, 0x18, 0xc0002e2948, 0x7, 0xc0002e2990, 0xa, 0xc000474b80, 0x18)
	/home/GOWORK/src/cooper/cooper/vendor/github.com/containers/image/v5/pkg/docker/config/config.go:52 +0xa8
```

**Describe the results you expected:**
no panic, login success with right auth info; `auth.json` will be recover to the right form or error poped out

**Additional information you deem important (e.g. issue happens only occasionally):**

- issue will `always happened` as long as the fileds in auth.json was modified

- go.mod:  `github.com/containers/image/v5 v5.4.4`

**Additional environment details (AWS, VirtualBox, physical, etc.):**
Virtual Machine (CentOS7.5)

CooperLi:
It seems start with: 
`vendor/github.com/containers/image/v5/pkg/docker/config/config.go:228`

```go
func readJSONFile(path string, legacyFormat bool) (dockerConfigFile, error) {
	var auths dockerConfigFile
 	/* should set map here?
[...]

PR #973, opened 2020-06-28
docker/config: performed map creation action in advance

CooperLi:
When auth.json is modified and try to readJSONFile, the panic will
happened because `auths.AuthConfigs` is a nil map. So we need to create
map in advance to avoid panic.

Fix: #972

Signed-off-by: CooperLi <a710905118@gmail.com>

### DIFF
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -284,11 +284,11 @@ func getPathToAuth(sys *types.SystemContext) (string, bool, error) {
 // if the file exists and is empty, readJSONFile returns an error
 func readJSONFile(path string, legacyFormat bool) (dockerConfigFile, error) {
 	var auths dockerConfigFile
+	auths.AuthConfigs = map[string]dockerAuthConfig{}
 
 	raw, err := ioutil.ReadFile(path)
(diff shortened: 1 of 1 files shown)


pkg/docker/config/config.go
```diff
@@ -284,11 +284,11 @@ func getPathToAuth(sys *types.SystemContext) (string, bool, error) {
 // if the file exists and is empty, readJSONFile returns an error
 func readJSONFile(path string, legacyFormat bool) (dockerConfigFile, error) {
 	var auths dockerConfigFile
+	auths.AuthConfigs = map[string]dockerAuthConfig{}
```
TomSweeneyRedHat, line 287: Is it possible for AuthConfigs to have a valid value at this point?  If so, I don't think we want to reset it to empty here.

CooperLi: As far as I know, we just declear `auths` at line#286 and assign value for it  at line#304
`if err = json.Unmarshal(raw, &auths)`.  So there is no valid value at this point :-)

CooperLi:
Dup  #942, closed

PR #975, opened 2020-06-29
WIP - support multi-image docker archives

vrothberg:
Add a `MultiImageArchive{Reader,Writer}` to `docker/archive` to support
docker archives with more than one image.

To allow the new archive reader/writer to be used for copying images,
add an `Image{Destination,Source}` to `copy.Options`.  When set, the
destination/source referenced will be ignored and the specified
`Image{Destination,Source}` will be used instead.

### DIFF
--- a/docker/archive/dest.go
+++ b/docker/archive/dest.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"os"
 
+	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/image/v5/docker/tarfile"
 	"github.com/containers/image/v5/types"
 	"github.com/pkg/errors"
@@ -17,32 +18,36 @@ type archiveImageDestination struct {
 }
 
 func newImageDestination(sys *types.SystemContext, ref archiveReference) (types.ImageDestination, error) {
+	return newArchiveImageDestination(sys, ref.path, ref.destinationRef)
+}
+
+func newArchiveImageDestination(sys *types.SystemContext, path string, ref reference.NamedTagged) (*archiveImageDestination, error) {
 	// ref.path can be either a pipe or a regular file
 	// in the case of a pipe, we require that we can open it for write
 	// in the case of a regular file, we don't want to overwrite any pre-existing file
 	// so we check for Size() == 0 below (This is racy, but using O_EXCL would also be racy,
 	// only in a different way. Either way, it’s up to the user to not have two writers to the same path.)
-	fh, err := os.OpenFile(ref.path, os.O_WRONLY|os.O_CREATE, 0644)
+	fh, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE, 0644)
 	if err != nil {
-		return nil, errors.Wrapf(err, "error opening file %q", ref.path)
+		return nil, errors.Wrapf(err, "error opening file %q", path)
 	}
 
 	fhStat, err := fh.Stat()
 	if err != nil {
-		return nil, errors.Wrapf(err, "error statting file %q", ref.path)
+		return nil, errors.Wrapf(err, "error statting file %q", path)
 	}
 
 	if fhStat.Mode().IsRegular() && fhStat.Size() != 0 {
 		return nil, errors.New("docker-archive doesn't support modifying existing images")
 	}
 
-	tarDest := tarfile.NewDestinationWithContext(sys, fh, ref.destinationRef)
+	tarDest := tarfile.NewDestinationWithContext(sys, fh, ref)
 	if sys != nil && sys.DockerArchiveAdditionalTags != nil {
 		tarDest.AddRepoTags(sys.DockerArchiveAdditionalTags)
 	}
 	return &archiveImageDestination{
 		Destination: tarDest,
-		ref:         ref,
+		ref:         archiveReference{path, ref},
 		writer:      fh,
 	}, nil
 }
@@ -69,4 +74,62 @@ func (d *archiveImageDestination) Close() error {
 // - Uploaded data MAY be removed or MAY remain around if Close() is called without Commit() (i.e. rollback is allowed but not guaranteed)
 func (d *archiveImageDestination) Commit(ctx context.Context, unparsedToplevel types.UnparsedImage) error {
 	return d.Destination.Commit(ctx)
+}
+
+func (m multiImageDestinationReference) NewImageDestination(_ context.Context, _ *types.SystemContext) (types.ImageDestination, error) {
+	return m.dest, nil
+}
+
+// MultiImageDestinations allows for creating and writing to docker archives
+// that include more than one image.
+type MultiImageDestination struct {
+	*archiveImageDestination
+	path string
+}
+
+// multiImageDestinationReference is a types.ImageReference embedding a MultiImageDestination.
+type multiImageDestinationReference struct {
+	*archiveReference
+	dest *MultiImageDestination
+}
+
+// NewMultiImageDestination returns a MultiImageDestination for the specified path.
+func NewMultiImageDestination(sys *types.SystemContext, path string) (*MultiImageDestination, error) {
+	dest, err := newArchiveImageDestination(sys, path, nil)
+	if err != nil {
+		return nil, err
+	}
+	return &MultiImageDestination{dest, path}, nil
+}
+
+// Reference returns an ImageReference embedding the MultiImageDestination.
+func (m *MultiImageDestination) Reference() types.ImageReference {
+	ref := &archiveReference{path: m.path}
+	return &multiImageDestinationReference{ref, m}
+}
+
+// Close is a NOP.  Please use Finalize() for committing the archive and
+// closing the underlying resources.
+func (m *MultiImageDestination) Close() error {
+	return nil
+}
+
+// Commit is a NOP.  Please use Finalize() for committing the archive and
+// closing the underlying resources.
+func (m *MultiImageDestination) Commit(_ context.Context, _ types.UnparsedImage) error {
+	return nil
+}
+
+// Finalize commits pending data and closes the underlying tarfile.
+func (m *MultiImageDestination) Finalize(ctx context.Context) (finalErr error) {
+	defer func() {
+		if err := m.writer.Close(); err != nil {
+			if finalErr == nil {
+				finalErr = err
+			} else {
+				finalErr = errors.Wrap(finalErr, err.Error())
+			}
+		}
+	}()
+	return m.Destination.Commit(ctx)
 }
--- a/docker/archive/src.go
+++ b/docker/archive/src.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"github.com/containers/image/v5/docker/tarfile"
+	ctrImage "github.com/containers/image/v5/image"
 	"github.com/containers/image/v5/types"
 	"github.com/sirupsen/logrus"
 )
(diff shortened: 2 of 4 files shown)


vrothberg:
libpod PR to illustrate how I imagine it to be used -> https://github.com/containers/libpod/pull/6811

copy/copy.go
```diff
@@ -170,6 +170,10 @@ type Options struct {
 	ForceManifestMIMEType string
 	ImageListSelection    ImageListSelection // set to either CopySystemImage (the default), CopyAllImages, or CopySpecificImages to control which instances we copy when the source reference is a list; ignored if the source reference is not a list
 	Instances             []digest.Digest    // if ImageListSelection is CopySpecificImages, copy only these instances and the list itself
+	// ImageDestination is a preset types.ImageDestination object. If not nil, the `destRef` will be ignored.
+	ImageDestination types.ImageDestination
+	// ImageSource is a preset types.ImageSource object. If not nil, the `srcRef` will be ignored.
+	ImageSource types.ImageSource
```
mtrmac, line 176: This is not really necessary; it’s, in the extreme, possible to create a do-nothing `ImageReference` that just returns the right `ImageDestination`/`ImageSource`.

But doing so would also more starkly expose the need to deal with names/identities (if nothing else, so that error messages from `copy.Image` are meaningful). I do think that needs to happen — the multi* objects should be a way to get (lifetime-limited, but otherwise pretty ordinary) `ImageReference` values.

docker/archive/multi-reader.go
```diff
@@ -0,0 +1,77 @@
...
+	"context"
+
+	"github.com/containers/image/v5/docker/tarfile"
+	"github.com/containers/image/v5/types"
+)
+
+// MultiImageArchiveReader allows for reading a docker-archive with multiple
+// images.  It can be used in subsequent copy.Image operations when set as the
+// copy.Options.ImageSource.  Note that you need to call Next() before each
+// copy.Image operation.
```
mtrmac, line 13: I don’t think this needs to be, or should be, that stateful.

Instead of embedding an `archiveImageSource`, and trying to “reset” it, this could basically wrap `docker/tarfile.Source.{tarPath,removeTarPathOnClose}` (+`openTarComponent`), and provide a way to list the manifests and just get `docker/archive.archiveReference` values that are “bound” to that `tarPath`.

That
- would allow reading from the archive in parallel
- would be clearly correct in not sharing state across images in the archive
- would force use to deal with the `ImageReference` values, not just using the right tags, but dealing with untagged images and introducing names for them (which IIRC can be created) — and that would also allow referencing any _single_ image using the string `docker-archive:` syntax by tag / untagged.

docker/archive/multi-reader.go
```diff
@@ -0,0 +1,77 @@
...
+type MultiImageArchiveReader struct {
+	*archiveImageSource
+
+	manifests     []tarfile.ManifestItem
+	manifestIndex int
+}
+
+// NewMultiImageArchiveReader returns a new MultiImageArchiveReader based on the reference.
+func NewMultiImageArchiveReader(ctx context.Context, sys *types.SystemContext, ref types.ImageReference) (*MultiImageArchiveReader, error) {
+	path, destinationRef, err := parsePathAndNamedRef(ref.StringWithinTransport())
```
mtrmac, line 26: This does not really make sense; the user could supply a `dockerReference` here, and it’s going to be accepted?!

`destinationRef` is completely ignored on the read path anyway, right now, and all the user _wanted_ to supply is a file path.

docker/archive/multi-reader.go
```diff
@@ -0,0 +1,77 @@
...
+	return &MultiImageArchiveReader{
+		archiveImageSource: src,
+		manifests:          manifests,
+		manifestIndex:      -1, // So we can call Next() before each copy.Image call.
+	}, nil
+}
+
+// Next selects the next image in the archive. False is returned when no image
+// is left.
+func (m *MultiImageArchiveReader) Next() bool {
```
mtrmac, line 54: It’s a bit awkward that the caller has to manually track the same index value to obtain the tags from a separately-obtained copy of the manifest.

docker/archive/multi-writer.go
```diff
@@ -0,0 +1,51 @@
...
+//
+// For using a MultiImageArchiveWriter in copy.Image, set it as the
+// copy.Option.ImageDestination.
+type MultiImageArchiveWriter struct {
+	*archiveImageDestination
+}
+
+func NewMultiImageArchiveWriter(ctx context.Context, sys *types.SystemContext, path string) (*MultiImageArchiveWriter, error) {
+	ref := archiveReference{path: path}
+	dst, err := newArchiveImageDestination(sys, ref)
```
mtrmac, line 22: I don’t think this works at all; how does the user specify tags for the individual images?

Again, let this object primarily track the temporary file (and more state about layers, in that case), and give out `ImageReference` objects that can be used to write a single image.

docker/tarfile/dest.go
```diff
@@ -184,22 +187,14 @@ func (d *Destination) TryReusingBlob(ctx context.Context, info types.BlobInfo, c
 }
 
 func (d *Destination) createRepositoriesFile(rootLayerID string) error {
```
mtrmac, line 189: The name needs updating.

docker/archive/multi-reader.go
vrothberg: I like your proposal. Note that I very likely will run out of time for extending the `docker-archive` syntax but we would definitely get the ground work done.

docker/archive/multi-writer.go
vrothberg: Tags can be specified via `AddRepoTags()`. I just finished rewriting the source parts and will now look into the destination.

vrothberg:
@mtrmac, mind taking another look? I've been starring at it too long.

docker/archive/dest.go
```diff
@@ -70,3 +75,61 @@ func (d *archiveImageDestination) Close() error {
...
+	dest, err := newArchiveImageDestination(sys, path, nil)
+	if err != nil {
+		return nil, err
+	}
+	return &MultiImageDestination{dest, path}, nil
+}
+
+// Reference returns an ImageReference embedding the MultiImageDestination.
+func (m *MultiImageDestination) Reference() types.ImageReference {
+	ref := &archiveReference{path: m.path}
```
mtrmac, line 107: This does not include the tag

docker/archive/src.go
```diff
@@ -34,3 +35,83 @@ func newImageSource(ctx context.Context, sys *types.SystemContext, ref archiveRe
...
+	tarSource *tarfile.Source
+}
+
+// Manifest returns the tarfile.ManifestItem.
+func (m *MultiImageSourceItem) Manifest() (*tarfile.ManifestItem, error) {
+	items, err := m.tarSource.LoadTarManifest()
+	if err != nil {
+		return nil, err
+	}
+	return &items[0], nil
```
mtrmac, line 53: So this only allows reading the first image, instead of all/any of them?

docker/tarfile/dest.go
```diff
@@ -256,20 +247,18 @@ func (d *Destination) PutManifest(ctx context.Context, m []byte, instanceDigest
...
-		return err
-	}
+	})
 
-	// FIXME? Do we also need to support the legacy format?
-	return d.sendBytes(manifestFileName, itemsBytes)
+	// Reset the repoTags to prevent them from leaking into a following
+	// image/manifest.
+	d.repoTags = []reference.NamedTagged{}
+	return nil
```
mtrmac, line 261: This is still stateful, then…

docker/archive/src.go
vrothberg: It points to _one_ image in the tarfile but the tarfile API returns a slice.

docker/archive/dest.go
vrothberg: Tags are set via `AdditionalTags`.

https://github.com/containers/libpod/pull/6811/files#diff-3303ff4a5ad91328550c7bee8df0be69R259

docker/archive/src.go
mtrmac: Oh, this is not `ImageSource.GetManifest`. My mistake.

docker/archive/dest.go
mtrmac: Yes, but not on the reference = not in `copy.Image` error messages.

mtrmac: OTOH there is still the case of saving untagged images, so references don’t _always_ have any extra data anyway.

I’ll read all of this more carefully a bit later.

vrothberg: Much appreciated, thanks!

I am on PTO tomorrow but would will go back to this on Monday morning. Ideally, we need to get the feature in next week.

mtrmac: BTW `AdditionalTags` was basically a hack to allow creating archives with multiple tags from a single image as a workaround for not having true multi-image support (where the same result would be obtained naturally by just copying the image to references with different tags); so this PR should conceptually eliminate the need for it.

vrothberg: @mtrmac, ideas how to proceed?

vrothberg: @mtrmac @rhatdan ... can we get this moving or are we blocked on something? I am getting increasing pressure to get this done.

vrothberg: Can we please proceed?

mtrmac: I’m sorry for not getting back, I did promise I would.

Still, https://github.com/containers/image/pull/975#pullrequestreview-441846261 was I think fairly clear to the general direction. See https://github.com/containers/image/pull/991 for an unfinished prototype of the read side. Yes, the PR is larger, but it already includes the ability to read any (even untagged) image in an archive using a textual reference, and a lot of the “new” code is actually only moved — `tarfile.Source` was just split into two, the only non-trivial _net new_ code at that layer is just `chooseManifest`.

vrothberg: Thanks! I am still not sure how to proceed.

Now we have two PRs and I am worried we are running out of time; in fact, we might not get it into RHEL 8.3 any more.

If you agree, I want to focus on Podman-only needs first. We can still make follow-up cards for a more generally applicable solution, in case that will buy us some time. Certainly, the API shouldn't break.

docker/archive/src.go
```diff
@@ -34,3 +35,83 @@ func newImageSource(ctx context.Context, sys *types.SystemContext, ref archiveRe
...
+// MultiImageSourceItem is a reference to _one_ image in a multi-image archive.
+// Note that MultiImageSourceItem implements types.ImageReference.  It's a
+// long-lived object that can only be closed via it's parent MultiImageSource.
+type MultiImageSourceItem struct {
+	*archiveReference
+	tarSource *tarfile.Source
+}
+
+// Manifest returns the tarfile.ManifestItem.
+func (m *MultiImageSourceItem) Manifest() (*tarfile.ManifestItem, error) {
```
mtrmac, line 48: Nothing but `RepoTags` is _really_ usable by callers, see e.g. the use of `Config` in https://github.com/containers/podman/pull/6811 ; so I’m a bit reluctant to commit to this as an API. OTOH, we clearly _can_ keep this stable.

docker/archive/dest.go
```diff
@@ -70,3 +75,61 @@ func (d *archiveImageDestination) Close() error {
...
 }
+
+func (m multiImageDestinationReference) NewImageDestination(_ context.Context, _ *types.SystemContext) (types.ImageDestination, error) {
+	return m.dest, nil
+}
+
+// MultiImageDestinations allows for creating and writing to docker archives
+// that include more than one image.
+type MultiImageDestination struct {
+	*archiveImageDestination
```
mtrmac, line 86: Doesn’t this externally expose `PutBlob` and everything else?

docker/archive/dest.go
```diff
@@ -70,3 +75,61 @@ func (d *archiveImageDestination) Close() error {
...
+
+// Reference returns an ImageReference embedding the MultiImageDestination.
+func (m *MultiImageDestination) Reference() types.ImageReference {
+	ref := &archiveReference{path: m.path}
+	return &multiImageDestinationReference{ref, m}
+}
+
+// Close is a NOP.  Please use Finalize() for committing the archive and
+// closing the underlying resources.
+func (m *MultiImageDestination) Close() error {
```
mtrmac, line 113: `Close()` is conceptually a bit different from `Finalize()` (or `ImageDestination.Commit`); it allows cleaning up the temporary files even on error.

Sure, this is a possible way to structure the API, but it’s a bit inconvenient to use: A typically caller will use something like `defer multiDest.Close()` and might not even check for errors if there is already a failure with a more important root cause to preserve, whereas on success the caller really wants to check that `multiDest.Finalize()` did succeed. Having “commit” and “deallocate” be the same operation forces every such caller to have a `committed` flag that’s checked inside the `defer`, or to have a critical part of the process in a `defer`.

docker/tarfile/dest.go
mtrmac, line 261: … it’s also not documented that the caller is supposed to use (only) `AddRepoTags` after each `PutManifest` AFAICS.

vrothberg:
Closing as @mtrmac took over. Thanks again!

PR #976, opened 2020-06-30
return err for no decryptconfig in dec. routine

lumjjb:
Addresses https://github.com/cri-o/cri-o/issues/3913


lumjjb:
Disregard PR, behavior of code is correct as defined by spec... To support copying encrypted images from registry to registry without unpacking... Change for that issue should be in crio instead.

Issue #974, opened 2020-06-29
Do not use 0755 permission for files and dirs


CooperLi:
Some files should not have 0755 permissions (especially **auth.json**). Nothing serious as of now. But, still ;-)

```bash
$ grep -Erni "0?755"                                                                                                                                                  
pkg/docker/config/config.go:348:                if err = ioutil.WriteFile(path, newData, 0755); err != nil {                                                          
Makefile:53:    install -d -m 755 ${MANINSTALLDIR}/man5                                                                                                               
oci/layout/oci_dest.go:319:             if err := os.MkdirAll(path, 0755); err != nil {                                                                               
.
.
.                                  
```

mtrmac:
Thanks for your report. You’re right about `auth.json`; but why should the directories not be 0755? Especially the man pages should be accessible by all users if installed system-wide, but even for the OCI files, the general convention is to make files group/user readable and limit that by umask, AFAIK.

CooperLi:
ping @mtrmac 
You are right about it, we need to consider the permission about `auth.json` first and leave the dir to the umask for example :-)

PR #978, opened 2020-07-08
pkg/docker/config: correct default file mode when create auth.json file

CooperLi:
`auth.json` contains sensitive info which should not be seen or modified by other people
and no need for execution right, so we change the default filemode from
0755 to 0600

Fix: #974

Signed-off-by: CooperLi <a710905118@gmail.com>


rhatdan:
LGTM

PR #991, opened 2020-07-25
Support reading any/all images from docker-archive: tarballs

mtrmac:
- Add support for choosing an image to read from a `docker-archive:`, either by reference (using the existing syntax) or by index (`docker-archive:/some/path:@0`).
- Add `docker/archive.Reader` that allows listing all images in a `docker-archive`-formatted file, and reading any/all of them, only copying/extracting the archive once.

### DIFF
--- a/docker/internal/tarfile/src.go
+++ b/docker/internal/tarfile/src.go
@@ -11,6 +11,7 @@ import (
 	"path"
 	"sync"
 
+	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/image/v5/internal/iolimits"
 	"github.com/containers/image/v5/manifest"
 	"github.com/containers/image/v5/pkg/compression"
@@ -22,7 +23,8 @@ import (
 // Source is a partial implementation of types.ImageSource for reading from tarPath.
 type Source struct {
 	archive      *Reader
-	closeArchive bool // .Close() the archive when the source is closed.
+	closeArchive bool                  // .Close() the archive when the source is closed.
+	ref          reference.NamedTagged // May be nil to indicate the only image in the archive
 	// The following data is only available after ensureCachedDataIsPresent() succeeds
 	tarManifest       *ManifestItem // nil if not available yet.
 	configBytes       []byte
@@ -40,17 +42,14 @@ type layerInfo struct {
 	size int64
 }
 
-// TODO: We could add support for multiple images in a single archive, so
-//       that people could use docker-archive:opensuse.tar:opensuse:leap as
-//       the source of an image.
-// 	To do for both the NewSourceFromFile and NewSourceFromStream functions
-
-// NewSource returns a tarfile.Source for the only image in the specified archive.
+// NewSource returns a tarfile.Source for an image in the specified archive matching ref
+// (or the only image if ref is nil).
 // The archive will be closed if closeArchive
-func NewSource(archive *Reader, closeArchive bool) *Source {
+func NewSource(archive *Reader, closeArchive bool, ref reference.NamedTagged) *Source {
 	return &Source{
 		archive:      archive,
 		closeArchive: closeArchive,
+		ref:          ref,
 	}
 }
 
@@ -66,11 +65,10 @@ func (s *Source) ensureCachedDataIsPresent() error {
 // ensureCachedDataIsPresentPrivate is a private implementation detail of ensureCachedDataIsPresent.
 // Call ensureCachedDataIsPresent instead.
 func (s *Source) ensureCachedDataIsPresentPrivate() error {
-	// Check to make sure length is 1
-	if len(s.archive.Manifest) != 1 {
-		return errors.Errorf("Unexpected tar manifest.json: expected 1 item, got %d", len(s.archive.Manifest))
+	tarManifest, err := s.archive.chooseManifestItem(s.ref)
+	if err != nil {
+		return err
 	}
-	tarManifest := &s.archive.Manifest[0]
 
 	// Read and parse config.
 	configBytes, err := s.archive.readTarComponent(tarManifest.Config, iolimits.MaxConfigBodySize)
--- a/docker/internal/tarfile/src.go
+++ b/docker/internal/tarfile/src.go
@@ -23,8 +23,10 @@ import (
 // Source is a partial implementation of types.ImageSource for reading from tarPath.
 type Source struct {
 	archive      *Reader
-	closeArchive bool                  // .Close() the archive when the source is closed.
-	ref          reference.NamedTagged // May be nil to indicate the only image in the archive
+	closeArchive bool // .Close() the archive when the source is closed.
+	// If ref is nil and sourceIndex is -1, indicates the only image in the archive.
+	ref         reference.NamedTagged // May be nil
+	sourceIndex int                   // May be -1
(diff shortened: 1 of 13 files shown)


docker/internal/tarfile/src.go
```diff
@@ -66,43 +70,67 @@ func (s *Source) ensureCachedDataIsPresent() error {
...
+	s.tarManifest = tarManifest
 	s.configBytes = configBytes
 	s.configDigest = digest.FromBytes(configBytes)
 	s.orderedDiffIDList = parsedConfig.RootFS.DiffIDs
 	s.knownLayers = knownLayers
 	return nil
 }
 
+// chooseManifest selects an item from manifests matching ref, which may be nil to mean the only image.
+func chooseManifest(manifests []ManifestItem, ref reference.NamedTagged) (*ManifestItem, error) {
```
vrothberg: I think we also need a way to chose _by index_ (or more complex by digest). I created the following `manifest.json` with `docker-save` with two untagged images.
```json
[
  {
    "Config": "1c35c441208254cb7c3844ba95a96485388cef9ccc0646d562c7fc026e04c807.json",
    "RepoTags": null,
    "Layers": [
      "9719c1ce1e2335e5b108c339b9f2b2f8aae8e62f30a831c40b9d9329678fd711/layer.tar"
    ]
  },
  {
    "Config": "a24bb4013296f61e89ba57005a7b3e52274d8edd3ae2077d04395f806b63d83e.json",
    "RepoTags": null,
    "Layers": [
      "fa27ccac46e9a5b7ff1d188e99763db1a5fb45adb34a917c1c482dfde99ad432/layer.tar"
    ]
  }
]
```

The RepoTags of both manifests is "null", so we cannot match by reference.

Maybe that's being tackled in a later commit. Will read further.

docker/archive/multi_source.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+// List returns the a set of references for images in the MultiImageSource,
+// grouped by the image the references point to.
+// The references are valid only until the MultiImageSource is closed.
+func (mis *MultiImageSource) List() ([][]types.ImageReference, error) {
+	res := [][]types.ImageReference{}
+	for imageIndex, image := range mis.archive.Manifest {
+		refs := []types.ImageReference{}
+		for _, tag := range image.RepoTags {
```
vrothberg, line 43: As mentioned in another comment, we need to account for manifests of untagged images where `RepoTags` is `nil`.

docs/containers-transports.5.md
```diff
@@ -41,11 +41,13 @@ If `name` does not contain a slash, it is treated as `docker.io/library/name`.
 Otherwise, the component before the first slash is checked if it is recognized as a `hostname[:port]` (i.e., it contains either a . or a :, or the component is exactly localhost).
 If the first component of name is not recognized as a `hostname[:port]`, `name` is treated as `docker.io/name`.
 
-### **docker-archive:**_path[:docker-reference]_
+### **docker-archive:**_path[:{docker-reference|@source-index}]_
```
vrothberg, line 44: :+1: love it!

docker/archive/multi_source.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/containers/image/v5/docker/internal/tarfile"
+	"github.com/containers/image/v5/docker/reference"
+	"github.com/containers/image/v5/types"
+	"github.com/pkg/errors"
+)
+
+// MultiImageSource manages a single Docker archive, allows listing its contents
+// and accessing individual images with less overhead than creating image references
+// individually (because the archive is, if necessary, copied or decompressed only once).
+type MultiImageSource struct {
```
mtrmac, line 13: I’m not entirely happy about this name - it’s not an `ImageSource`. `docker/archive.Reader`? `MultiReader`? `MultiImageReader`?

docker/internal/tarfile/readable.go
```diff
@@ -0,0 +1,227 @@
...
+
+	"github.com/containers/image/v5/internal/iolimits"
+	"github.com/containers/image/v5/internal/tmpdir"
+	"github.com/containers/image/v5/pkg/compression"
+	"github.com/containers/image/v5/types"
+	"github.com/pkg/errors"
+)
+
+// ReadableArchive is a ((docker save)-formatted) tar archive that allows random access to any component.
+type ReadableArchive struct {
```
mtrmac, line 19: This name feels a bit weird, but so does `internal/tarfile.Reader`. I’m very much open to suggestions.

OTOH it’s internal, so it’s not _as_ critical.

mtrmac:
@vrothberg This should now be code-complete (full transport tests, leading to fixes in reference parsing), but I still need to rebase to move some changes across commits. And it remains mostly untested, hoping Podman’s tests will handle the basics.

docker/archive/multi_source.go
mtrmac: Trying `docker/archive.Reader` — it’s not just for multi-image readers, a caller that wants to open a single-image archive, extract the tag from the manifest, and read the single image, would benefit as well.

docker/internal/tarfile/readable.go
mtrmac: Trying `tarfile.Reader`, with variable name `archive`. The `internal/tarfile.Reader`/`archive.Reader` pair is somewhat symmetrical, but it might be too confusing; I’m not sure.

mtrmac:
@vrothberg Ready for review (but still basically untested).

vrothberg:
Thanks, I'll test https://github.com/containers/image/pull/998 and will comment there.

rhatdan:
Needs a rebase

mtrmac:
Rebased, now includes all of #998.

vrothberg:
@rhatdan ... merge me :)

mtrmac:
It wouldn’t be unprecedented… but I kind of think that the caller should know that in the first place. IIRC the only really ambiguous case is `podman pull docker-archive:$path`, where I’d argue “treat the transport:name syntax always as a single image” is closest to a consistent semantics.

> Could we add means to figure out if a given reference refers to a file? Ultimately, if it's a file then we have to create a Reader and query for the reference list.
> 
> Currently, I don't find any other way than checking if `(ref).StringWithinTransport()` is on disk or not.

Well, if you want an ugly hack, the syntax is documented to split path:rest on a `:`. But I’d recommend against that.

mtrmac:
Added another commit, “Add docker/archive.NewReaderForReference”. DO NOT MERGE until it is verified it is useful for Podman.

mtrmac:
Also added `Reader.ManifestTagsForReference` to accommodate Podman’s handling of unqualified short names in the archives (which are only created by projectatomic/docker AFAIK). Ugh.

vrothberg (approved):
LGTM ... ready to merge from my side

rhatdan:
Need @mtrmac to remove the "DO NOT Merge:" from the commit message.

mtrmac:
OK, rebased and “DO NOT MERGE” dropped (although I haven’t had time to fully review the Podman PR)

rhatdan:
LGTM
Merge once it passes the tests.

vrothberg (commented):
LGTM, thanks a lot for the great work and your support, @mtrmac !

PR #992, opened 2020-07-27
Coverity found potential nil dereference

rhatdan:


mtrmac:
I’m fine with adding dead code to make other Coverity results more useful, but I’d like that dead code to be consistent with the rest.

PR #996, opened 2020-07-29
Writing multiple images to a single docker-archive: file

mtrmac:
WIP, absolutely untested, missing locking, probably incorrect about legacy metadata reuse.


mtrmac:
@vrothberg filing very early for API feedback and possible testing / experimenting.

mtrmac:
See also #998 for read+write in one PR.

vrothberg:
Thanks, I'll test #998 and will comment there.

mtrmac:
@vrothberg This should now be ready for review (although still basically untested). Changes since the early version:
- `docker/internal/tarfile.Writer.Finish` is now `Writer.Close`, similar to `archive.tar.Writer.Close`. There isn’t a (`Commit`,`Close`) difference like in `ImageDestination`, so it seems reasonable to just use the standard `Close` name.
- `docker/internal/tarfile.Writer` now uses a mutex to ensure the resulting stream is consistent, and in principle that allows copying images in parallel. I don’t expect much benefit from that, hence not setting `HasThreadSafePutBlob` either.
- The actual format implementation was changed much more than the earlier versions, to correctly create the legacy-format metadata (for images that share layers but differ in configs).

rhatdan:
Amazing work @mtrmac LGTM

henning-schild:
Can anyone point me to docs on how to actually use that with `skopeo` or `buildah` or any other tool really. I have multiple images and am trying to merge them into one repo image to deduplicate layers and save cost on storage and download.

henning-schild:
Thanks. I really do not want to know the details, just found this and thought it might be a good place to ask which of the many container tools actually uses that interface. I will take that pointer to podman and just hope that the version i can get can do that.

PR #997, opened 2020-07-29
Error message change

ParkerVR:
PR Response to https://bugzilla.redhat.com/show_bug.cgi?id=1848150

Minor change to error messages to provide more user-friendly experience

Signed-off-by: Parker Van Roy <pvanroy@redhat.com>

### DIFF
--- a/docker/docker_image_dest.go
+++ b/docker/docker_image_dest.go
@@ -83,7 +83,7 @@ func (d *dockerImageDestination) SupportsSignatures(ctx context.Context) error {
 	case d.c.supportsSignatures:
 		return nil
 	default:
-		return errors.Errorf("X-Registry-Supports-Signatures extension not supported, and lookaside is not configured")
+		return errors.Errorf("X-Registry-Supports-Signatures extension not supported. Configure sigstore-staging path (view containers-registries.d.5.md)")
 	}
 }
 
@@ -484,7 +484,7 @@ func (d *dockerImageDestination) PutSignatures(ctx context.Context, signatures [
 	case d.c.supportsSignatures:
 		return d.putSignaturesToAPIExtension(ctx, signatures, *instanceDigest)
 	default:
-		return errors.Errorf("X-Registry-Supports-Signatures extension not supported, and lookaside is not configured")
+		return errors.Errorf("Cannot store signature: sigstore not configured for destination registry")
 	}
 }
 


TomSweeneyRedHat:
@mtrmac here's the current sample output from the BZ, I did not try it myself.

```
Steps to Reproduce:
1. Configure a sigstore for a source registry
2. Do not configure a sigstore for destination registry
3. Copy from the source to the destination without specifying the "--remove-signatures" flag

Actual results:
Error received: "FATA[0003] Can not copy signatures: X-Registry-Supports-Signatures extension not supported, and lookaside is not configured"
```

mtrmac:
Yes, that’s pre-#963.

TomSweeneyRedHat:
Thanks @mtrmac , so as of skopeo v1.0.0 the new error message would look something like:

"Error received: Can not copy signatures to {destination_reference}: FATA[0003] Can not copy signatures: X-Registry-Supports-Signatures extension not supported, and lookaside is not configured"

The requestor had asked for:

"Can not copy signatures: Either configure the destination registry with X-Registry-Supports-Signatures extension or specify a sigstore-staging"

Could we change @ParkerVR verbiage change to 
Either configure the destination registry with X-Registry-Supports-Signatures extension or specify a sigstore-staging

Or do you have other thoughts?  I think the genesis of the problem is the current message hints at the problem, but doesn't hint at the solution.

ParkerVR:
Maybe it might be better to better distinguish errors between not having sigstore configured and the extension? Is this rational or would we have to rewrite the extension for something trivial?

rhatdan:
@ParkerVR Yes split them.

mtrmac:
> Either configure the destination registry with X-Registry-Supports-Signatures extension or specify a sigstore-staging

X-R-S-S support is a feature of the implementation; in the only implementation I’m aware of (OpenShift integrated registry) it’s not something that can be configured in any way. The registry either supports it it or does not. “Replace your registry with a completely different server” is… an option, but also probably not very frequently possible.

> Maybe it might be better to better distinguish errors between not having sigstore configured and the extension?

How? At that point, the extension is known to be unavailable, and no configuration exists. We can’t really tell whether the user intended one, or the other, to exist and something went wrong, or whether the user didn’t think about storing signatures at all and this is the first time the issue came up.

In a sense, it is more practical to point at sigstore-staging, because that can always be done no matter the registry — OTOH it is more _useful_ to point at X-R-S-S, _if_ the user can replace the registry setup, it results in a _much_ better workflow.

> Or do you have other thoughts? I think the genesis of the problem is the current message hints at the problem, but doesn't hint at the solution.

I think the only _substantive_ change that is a clear improvement at this point is to s/lookaside/sigstore/, and possibly mention sigstore-staging. Plus word-smithing (“is not configured” vs. “you should configure”) I don’t feel strongly about.

A sort of unorthodox option would be to include a hard-coded link to a larger external document in the message… that might help or just get in the way, I don’t know.

rhatdan:
@ParkerVR @mtrmac What is the scoop on this one?

docker/docker_image_dest.go
```diff
@@ -485,7 +485,7 @@ func (d *dockerImageDestination) PutSignatures(ctx context.Context, signatures [
 	case d.c.supportsSignatures:
 		return d.putSignaturesToAPIExtension(ctx, signatures, *instanceDigest)
 	default:
-		return errors.Errorf("X-Registry-Supports-Signatures extension not supported, and lookaside is not configured")
+		return errors.Errorf("Cannot store signature: sigstore not configured for destination registry")
```
mtrmac, line 488: This should still mention `X-R-S-S`; overall the two error messages should stay in sync.

docker/docker_image_dest.go
```diff
@@ -84,7 +84,7 @@ func (d *dockerImageDestination) SupportsSignatures(ctx context.Context) error {
 	case d.c.supportsSignatures:
 		return nil
 	default:
-		return errors.Errorf("X-Registry-Supports-Signatures extension not supported, and lookaside is not configured")
+		return errors.Errorf("X-Registry-Supports-Signatures extension not supported. Configure s/lookaside/sigstore/ or specify a sigstore-staging")
```
mtrmac, line 87: I’m sorry, the `s/`…`/`…`/` syntax is a jargon for replacing, not a literal suggestion ( http://www.catb.org/~esr/jargon/html/writing-style.html ).

Maybe
```suggestion
		return errors.Errorf("X-Registry-Supports-Signatures extension not supported. Configure a sigstore-staging path")
```
and perhaps point at the relevant man page in to say how; OTOH that may get way too long.

mtrmac (commented):
I’m sorry, I didn’t realize this was updated (@ParkerVR GitHub does not automatically send e-mail in that case).

docker/docker_image_dest.go
ParkerVR: I'm not sure of the proper way to link a page etc. but this is the doc I'd link: https://github.com/containers/image/blob/master/docs/containers-registries.d.5.md

mtrmac: Well, just include an URL in the page.

OTOH, in the meantime, there is https://github.com/containers/image/pull/1035 , which will make these error messages impossible to hit, so after/if that PR is accepted, this PR can be just closed.

---

@rhatdan (Cc: @QiWang19 ) FYI, one result of #1035 (having built-in defaults for per-user/per-system sigstores) is that copies of signed images to registries that don’t support signatures, and have no sigstore set up, will completely silently succeed, and users will have to figure out that signatures need publishing without much guidance.

That’s always been the case for root (with the configs pointing to `/var/…` for `sigstore-staging`), but now it will be the case for everyone with no way to opt out. I think it’s probably worth it for the simplicity, and anyone who cares about signatures can always configure a default sigstore path. Still, we are adding a silent failure^Wnon-success case for unprivileged users (where before, we were hard-failing unless the user explicitly provided a registries.d configuration that pointed to a writable directory).

rhatdan: Should we add an info message when they do not have a separate stage instance, telling that is what is happening?

mtrmac: That’s an interesting idea, definitely worth considering.

It would probably be way too noisy when reading signatures (most images are not signed, but we would want to tell the user about the default path exactly when no signatures are found); a reminder when writing signatures (i.e. in `PutSignatures`, when we already know there are signatures to write) could be rather helpful — even if the user knows about the configuration, just to say that the staging data must be published.

mtrmac:
#1035 was merged, making this no longer applicable.

PR #1000, opened 2020-08-06
Look for normalized paths in tarfile.

jellelicht:
I have a tarball with a "./" prefix (so: `./manifest.json`, `./config.json`, `./<layer-hash>/layer.tar` etc etc). This tarball works with docker load, but does not work with skopeo inspect, which is why I'm here now :smile: 

I added some normalization using `path.Clean` to all places where we might be looking at the path of something in the tarball; since I assume 'security' concerns (e.g. `../../`) are taken care of in other places, this should not introduce new vulnerabilities.

Let me know if I need to adapt something.

### DIFF
--- a/docker/tarfile/src.go
+++ b/docker/tarfile/src.go
@@ -184,10 +184,11 @@ func (s *Source) openTarComponent(componentPath string) (io.ReadCloser, error) {
 	return &tarReadCloser{Reader: tarReader, backingFile: f}, nil
 }
 
-// findTarComponent returns a header and a reader matching path within inputFile,
+// findTarComponent returns a header and a reader matching componentPath within inputFile,
 // or (nil, nil, nil) if not found.
-func findTarComponent(inputFile io.Reader, path string) (*tar.Reader, *tar.Header, error) {
+func findTarComponent(inputFile io.Reader, componentPath string) (*tar.Reader, *tar.Header, error) {
 	t := tar.NewReader(inputFile)
+	componentPath = path.Clean(componentPath)
 	for {
 		h, err := t.Next()
 		if err == io.EOF {
@@ -196,7 +197,7 @@ func findTarComponent(inputFile io.Reader, path string) (*tar.Reader, *tar.Heade
 		if err != nil {
 			return nil, nil, err
 		}
-		if h.Name == path {
+		if path.Clean(h.Name) == componentPath {
 			return t, h, nil
 		}
 	}
@@ -308,7 +309,7 @@ func (s *Source) prepareLayerData(tarManifest *ManifestItem, parsedConfig *manif
 			// which of the tarManifest.Layers paths is used; (docker save) actually makes the duplicates symlinks to the original.
 			continue
 		}
-		layerPath := tarManifest.Layers[i]
+		layerPath := path.Clean(tarManifest.Layers[i])
 		if _, ok := unknownLayerSizes[layerPath]; ok {
 			return nil, errors.Errorf("Layer tarfile %s used for two different DiffID values", layerPath)
 		}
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
@jellelicht needs a rebase

jellelicht:
ah sorry, I sausage-fingered the big button here, I'll rebase properly in a bit

docker/tarfile/src.go
```diff
@@ -184,9 +184,9 @@ func (s *Source) openTarComponent(componentPath string) (io.ReadCloser, error) {
 	return &tarReadCloser{Reader: tarReader, backingFile: f}, nil
 }
 
-// findTarComponent returns a header and a reader matching path within inputFile,
+// findTarComponent returns a header and a reader matching rpath within inputFile,
```
mtrmac, line 187: I don’t know what “rpath” stands for here (it is clearly not a reference to `realpath(3p)` nor to `ld -rpath`. Why not just say “path”?

docker/tarfile/src.go
```diff
@@ -196,7 +196,7 @@ func findTarComponent(inputFile io.Reader, path string) (*tar.Reader, *tar.Heade
 		if err != nil {
 			return nil, nil, err
 		}
-		if h.Name == path {
+		if path.Clean(h.Name) == path.Clean(rpath) {
```
mtrmac, line 199: `path.Clean(rpath)` can be computed once outside the loop.

rhatdan, line 199: path.Clean(rpath)  should only be done once.  Not every time though the loop

docker/tarfile/src.go
```diff
@@ -335,7 +335,7 @@ func (s *Source) prepareLayerData(tarManifest *ManifestItem, parsedConfig *manif
 		if err != nil {
 			return nil, err
 		}
-		if li, ok := unknownLayerSizes[h.Name]; ok {
+		if li, ok := unknownLayerSizes[path.Clean(h.Name)]; ok {
```
mtrmac, line 338: The same normalization must happen when adding entries to `unknownLayerSizes` a few lines above as well.

Also, the `path.Clean(h.Name)` can be only computed once.

rhatdan, line 338: Can  you do this once and save the variable to be used below.  I am not sure how much path.Clean does, but we probably don't want to call it multiple times, if we don't have to.

docker/tarfile/src.go
mtrmac: (It does seem useful to put a warning _somewhere_ to future maintainers to think about the need to use `path.Clean`. But I can’t think of a place that would work.)

jellelicht: I want to say `path.Clean(...)`. Do you have a proposal that is more clear? I don't really know the shadowing rules in Go

docker/tarfile/src.go
jellelicht: I've tried to pull this out of the loop

docker/tarfile/src.go
jellelicht: done

docker/tarfile/src.go
jellelicht: I opted for `componentPath` for the argument/parameter, so that leaves `path` to actually refer to the import.

docker/tarfile/src.go
jellelicht: done

docker/tarfile/src.go
jellelicht: done, and done

docker/tarfile/src.go
mtrmac: Ah, I didn’t realize that the package name was conflicting.

PR #1002, opened 2020-08-07
check docker-archive support encryption

QiWang19:
copy image should failed when docker-archive does not support encryption.

related issues: https://github.com/containers/skopeo/issues/982

### DIFF
--- a/copy/manifest.go
+++ b/copy/manifest.go
@@ -65,6 +65,9 @@ func (ic *imageCopier) determineManifestConversion(ctx context.Context, destSupp
 		if !requiresOciEncryption || manifest.MIMETypeSupportsEncryption(t) {
 			supportedByDest[t] = struct{}{}
 		}
+		if ic.c.dest.Reference().Transport().Name() == "docker-archive" && !manifest.MIMETypeSupportsEncryption(t) {
+			return "", nil, errors.New("docker-archive does not support encryption")
+		}
 	}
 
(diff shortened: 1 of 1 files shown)


QiWang19:
@mtrmac PTAL. Will this work for fix https://github.com/containers/skopeo/issues/982?

mtrmac:
It would also be interesting to look into the final stages of the failure — AFAICS the layer really is encrypted, and the encryption code records the metadata in `uploadedInfo`, and if that data is not supported by a format, it is just silently ignored instead of failing loudly.

If practical, we _should_ fail before starting to copy the multi-gigabyte layers, but at the point we do have the encryption metadata, the code should be robust and not just silently ignore it.

mtrmac:
#1930 fixes the original failure in `determineManifestConversion`.

mtrmac:
> It would also be interesting to look into the final stages of the failure

Still outstanding, I’ll take a look later.

TomSweeneyRedHat:
@QiWang19 looks like you need a rebase.

mtrmac:
@TomSweeneyRedHat This PR is very old, mostly just tracking the issue to be fixed; see the last few comments.

mtrmac:
> It would also be interesting to look into the final stages of the failure

(Separately from #1930 stopping such operations at the very start,) WIP #1932 will fix that part as well:
```
… creating an updated image manifest: encryption change (for layer "sha256:71daf5f79bf641926f93bdc9514288dadab2c97e99823b8e6617de046b72f518") is not supported in schema2 manifests 
```

PR #1012, opened 2020-08-09
use container/storage/pkg/homedir

rhatdan:


rhatdan:
This ends up pulling the docker dependency into containers/common and then into other repos.  We have both being used, and I would rather stick to the one we maintain.

mtrmac:
Most of users of c/common are going to pull in docker/docker via c/image/transports/alltransports.ParseImageName through c/image/docker/daemon anyway, so I don’t think this is going to make any difference in that respect.

rhatdan:
Well lets hope it does.  @vrothberg @mtrmac @TomSweeneyRedHat PTAL and Merge.

TomSweeneyRedHat:
LGTM

TomSweeneyRedHat:
I'll let @mtrmac or @vrothberg hit the big merge button

PR #1014, opened 2020-08-10
backports pagination fix #964

QiWang19:
Backports #964 to release-5.5
Enable search registry uses the pagination until the search result reaches the limit, instead of returning default 100 limit from registry API.
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1866153
~BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1827794~

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -331,7 +331,6 @@ func SearchRegistry(ctx context.Context, sys *types.SystemContext, registry, ima
 		// Results holds the results returned by the /v1/search endpoint
 		Results []SearchResult `json:"results"`
 	}
-	v2Res := &V2Results{}
 	v1Res := &V1Results{}
 
 	// Get credentials from authfile for the underlying hostname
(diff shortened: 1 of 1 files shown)


vrothberg:
Can you also cherry-pick commit 050d8f2466f7? This should make CI pass again.

QiWang19:
@vrothberg Do I need other commit to pass the test? I Also cherry-pick commit 5261342ead96b4a9c9ce5e120433050e5b5d7f9e because there's go-multierror failure.

vrothberg:
The BZ link should be https://bugzilla.redhat.com/show_bug.cgi?id=1866153.

> I don’t think it’s reasonable to have make test-skopeo chase all future changes; whenever Skopeo is updated to use a new feature of c/image, this branch would have to backport that features as well. Instead, use a SKOPEO_BRANCH that corresponds to the c/image branch, or possibly even hard-code a version of skopeo that uses the right c/image version.

How could I miss that? Thanks!

@QiWang19, the correct SKOPEO_BRANCH is https://github.com/containers/skopeo/tree/release-1.1.

Makefile
```diff
@@ -4,7 +4,7 @@ export GOPROXY=https://proxy.golang.org
 
 # Which github repository and branch to use for testing with skopeo
 SKOPEO_REPO = containers/skopeo
-SKOPEO_BRANCH = master
+SKOPEO_BRANCH = release-1.1
```
QiWang19, line 7: @vrothberg @mtrmac change Makefile like this?

mtrmac: In… principle. Except that `release-1.1` uses c/image 5.5.

AFAICS there has never been a Skopeo release that uses 5.1; we went from 5.0.0 in 0.1.40 to 5.2.1 in 0.1.41.

And yet again, the latest bug refers to Podman 2.0.[45], which use c/image 5.5.1, not 5.1.x. Is 5.1 really the right target?

QiWang19: should I commit to https://github.com/containers/image/tree/release-5.5?

vrothberg: This PR is against `release-5.5`, so I think everything's in order?

vrothberg: Apologies, I guess I caused the confusion in my comment in BZ where I asked to open against v5.1 but meant v5.5.

However, this PR is against v5.5, so we should be good :+1:

mtrmac:
@QiWang19 Travis somehow still seems to remember the old commit and the old 5.1 target. If doesn’t catch up by itself, can you try creating a fresh PR for the 5.5 branch, please?

QiWang19:
Thanks. replace this with  #1020

PR #1027, opened 2020-08-18
Support per user registries.d

QiWang19:
Support per user  ~/.config/containers/registries.d to allow rootless podman push image sign.

### DIFF
--- a/docker/lookaside.go
+++ b/docker/lookaside.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/image/v5/types"
+	"github.com/containers/storage/pkg/homedir"
 	"github.com/ghodss/yaml"
 	"github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
@@ -25,6 +26,9 @@ var systemRegistriesDirPath = builtinRegistriesDirPath
 // builtinRegistriesDirPath is the path to registries.d.
 // DO NOT change this, instead see systemRegistriesDirPath above.
 const builtinRegistriesDirPath = "/etc/containers/registries.d"
+
+// userRegistriesDirPath is the path to the per user registries.d.
+var userRegistriesDir = filepath.FromSlash(".config/containers/registries.d")
 
 // registryConfiguration is one of the files in registriesDirPath configuring lookaside locations, or the result of merging them all.
 // NOTE: Keep this in sync with docs/registries.d.md!
@@ -75,14 +79,17 @@ func configuredSignatureStorageBase(sys *types.SystemContext, ref dockerReferenc
 
 // registriesDirPath returns a path to registries.d
 func registriesDirPath(sys *types.SystemContext) string {
-	if sys != nil {
-		if sys.RegistriesDirPath != "" {
-			return sys.RegistriesDirPath
-		}
-		if sys.RootForImplicitAbsolutePaths != "" {
-			return filepath.Join(sys.RootForImplicitAbsolutePaths, systemRegistriesDirPath)
-		}
+	if sys != nil && sys.RegistriesDirPath != "" {
+		return sys.RegistriesDirPath
 	}
+	userRegistriesDirPath := filepath.Join(homedir.Get(), userRegistriesDir)
+	if _, err := os.Stat(userRegistriesDirPath); err == nil {
+		return userRegistriesDirPath
+	}
+	if sys != nil && sys.RootForImplicitAbsolutePaths != "" {
+		return filepath.Join(sys.RootForImplicitAbsolutePaths, systemRegistriesDirPath)
+	}
+
 	return systemRegistriesDirPath
 }
 
--- a/docker/lookaside_test.go
+++ b/docker/lookaside_test.go
@@ -53,20 +53,36 @@ func TestRegistriesDirPath(t *testing.T) {
 	const nondefaultPath = "/this/is/not/the/default/registries.d"
 	const variableReference = "$HOME"
 	const rootPrefix = "/root/prefix"
-
+	oldHomeEnv, hasHomeEnv := os.LookupEnv("HOME")
+	tempHome, err := ioutil.TempDir("", "tempHome")
+	require.NoError(t, err)
+	err = os.Setenv("HOME", tempHome)
+	require.NoError(t, err)
+	defer func() {
+		os.RemoveAll(tempHome)
+		if hasHomeEnv {
+			os.Setenv("HOME", oldHomeEnv)
+		} else {
+			os.Unsetenv("HOME")
+		}
+	}()
+	var userRegistriesDir = filepath.FromSlash(".config/containers/registries.d")
+	userRegistriesDirPath := filepath.Join(tempHome, userRegistriesDir)
 	for _, c := range []struct {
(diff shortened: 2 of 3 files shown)


docker/lookaside.go
```diff
@@ -75,14 +79,17 @@ func configuredSignatureStorageBase(sys *types.SystemContext, ref dockerReferenc
...
-		if sys.RegistriesDirPath != "" {
-			return sys.RegistriesDirPath
-		}
-		if sys.RootForImplicitAbsolutePaths != "" {
-			return filepath.Join(sys.RootForImplicitAbsolutePaths, systemRegistriesDirPath)
-		}
+	if sys != nil && sys.RegistriesDirPath != "" {
+		return sys.RegistriesDirPath
 	}
+	userRegistriesDirPath := filepath.Join(homedir.Get(), userRegistriesDir)
```
TomSweeneyRedHat, line 85: should we use securejoin here?

mtrmac: No reason to, `userRegistriesDir` is a constant and nothing crosses a privilege boundary.

mtrmac:
From a quick look, ACK to having this functionality, but it’s not yet enough for making rootless work, as long as the default config file contains a `sigstore-staging:` default pointing to `/var`. Are we going to change that default?

QiWang19:
Can the default sigstore-staging be pointed to ~/.local/share/containers/sigstore for rootless?

mtrmac:
Not in the config files as they exist now. As for code — we can’t tell whether the user wants to use the global config files or not, just by noticing that a per-user config file is missing.

mtrmac:
(I’m sure this was discussed previously, and there might have been an agreed solution that I can’t remember offhand. I vaguely remember a suggestion to ship a system-wide config file that doesn’t contain a sigstore-staging (but leave the old config file around for upgrading users), and then, in code, default to the per-user location if not root, but there might well have been a different consensus.)

mtrmac:
… also, none of that directly prevents merging this PR, of course; I just want to make sure the full workflow is considered before the feature is “done”.

rhatdan:
Yes I don't think we should ship a default config, but have the config built in.  It can default to one behavior for rootfull and a different one for rootless.

rhatdan:
LGTM

rhatdan:
The tools using containers/image can, and could pass in an indicator to use rootless or rootful mode.

mtrmac:
> > Yes I don't think we should ship a default config, but have the config built in. It can default to one behavior for rootfull and a different one for rootless.
> 
> Without a config file, c/cimage can't tell if the user is rootful or rootless set the path in the code.

(I’ll set aside what is “rootless”, I never understood in full what Podman means.)

c/image should be able to see whether EUID == 0, and default to `/var` vs. `~/.local/share` .  That’s assuming that the c/image/docker code is not running under a user namespace — and if it is, I think it can’t read the configuration, and read/write sigstores, in user’s home directory anyway. Or am I wrong about how user namespaces work in Podman?

docker/lookaside_test.go
```diff
@@ -75,11 +91,35 @@ func TestRegistriesDirPath(t *testing.T) {
 				RootForImplicitAbsolutePaths: rootPrefix,
 				RegistriesDirPath:            nondefaultPath,
 			},
+			false,
+			nondefaultPath,
+		},
+		// User registries.d overridden
```
mtrmac, line 97:
```suggestion
		// User registries.d present, not overridden
```

docker/lookaside_test.go
```diff
@@ -75,11 +91,35 @@ func TestRegistriesDirPath(t *testing.T) {
...
 	} {
+		if c.userFilePresent {
+			err := os.MkdirAll(filepath.Dir(userRegistriesDirPath), os.ModePerm)
+			require.NoError(t, err)
+			f, err := os.Create(userRegistriesDirPath)
+			require.NoError(t, err)
+			f.Close()
+		} else {
+			os.Remove(userRegistriesDirPath)
+			require.NoError(t, err)
```
mtrmac, line 121: `err` is not set in the `os.Remove` call above, this is checking the result of `Setenv` above or something like that.

But if two non-`userFilePresent` tests exist in a sequence, this should not fail on “does not exist” anyway. So, either use `os.RemoveAll`, or check the error here, but only if it is not “directory does not exist”.

docker/lookaside_test.go
```diff
@@ -75,11 +91,35 @@ func TestRegistriesDirPath(t *testing.T) {
...
+			},
+			true,
 			nondefaultPath,
 		},
 		// No environment expansion happens in the overridden paths
-		{&types.SystemContext{RegistriesDirPath: variableReference}, variableReference},
+		{&types.SystemContext{RegistriesDirPath: variableReference}, false, variableReference},
 	} {
+		if c.userFilePresent {
+			err := os.MkdirAll(filepath.Dir(userRegistriesDirPath), os.ModePerm)
```
mtrmac, line 114: Don’t use `0777`, that unnecessarily allows external interference. `0700` should be fine.

docker/lookaside_test.go
```diff
@@ -75,11 +91,35 @@ func TestRegistriesDirPath(t *testing.T) {
...
+			require.NoError(t, err)
+			f, err := os.Create(userRegistriesDirPath)
```
mtrmac, line 116: Why is this creating a file and not a directory? In fact, just the `os.MkdirAll` above could create that.

docs/containers-registries.d.5.md
```diff
@@ -12,7 +12,7 @@ The registries configuration directory contains configuration for various regist
 so that the configuration does not have to be provided in command-line options over and over for every command,
 and so that it can be shared by all users of containers/image.
 
-By default (unless overridden at compile-time), the registries configuration directory is `/etc/containers/registries.d`;
+By default (unless overridden at compile-time), the registries configuration directory is `$HOME/.config/containers/registries.d` if it exists, otherwise `/etc/containers/registries.d`will be used;
```
mtrmac, line 15:
```suggestion
By default, the registries configuration directory is `$HOME/.config/containers/registries.d` if it exists, otherwise `/etc/containers/registries.d` (unless overridden at compile-time);
```

rhatdan:
Rootless means that podman or buildah or skopeo started by a non root user.  The difficulty is that Podman and Buildah jump into the user namespace right away,causing the EUID to be "root" within the user namespace.  So they need to figure out whether they are in a user namespace and sometimes they need to figure out the "Real" UID.

mtrmac:
> Rootless means that podman or buildah or skopeo started by a non root user. The difficulty is that Podman and Buildah jump into the user namespace right away,causing the EUID to be "root" within the user namespace.

[So does it mean “non-root” or “root within a user namespace”?]

Ouch; that does require cooperation from the caller.

… it turns out there already is https://github.com/containers/image/blob/8ad4121296f87a366fcc198ac23ecf650f527286/pkg/blobinfocache/default.go#L51-L58 ; is that the right check? If so, we could move that into a separate internal subpackage and use that without having the caller do anything else.

rhatdan:
Yes that will tell you the UID that launched the container.  Since you can have a user namespace from root, I think it is best to check this environment variable.

mtrmac (approved):
LGTM if tests pass. Thanks!

PR #1029, opened 2020-08-19
Define an error to allow an external reference

zhangguanzhang:
Fixes: https://github.com/containers/podman/issues/7337
more detail could see https://github.com/containers/podman/issues/7337#issuecomment-675584694
Signed-off-by: zhangguanzhang <zhangguanzhang@qq.com>

### DIFF
--- a/oci/layout/oci_transport.go
+++ b/oci/layout/oci_transport.go
@@ -30,6 +30,9 @@ var (
 	// ErrMoreThanOneImage is an error returned when the manifest includes
 	// more than one image and the user should choose which one to use.
 	ErrMoreThanOneImage = errors.New("more than one image in oci, choose an image")
+
+	// ErrNotFoundDescriptor is an error when the reference cannot found in the manifests
+	ErrNotFoundDescriptor = errors.New("no descriptor found for the reference")
 )
 
 type ociTransport struct{}
@@ -209,7 +212,7 @@ func (ref ociReference) getManifestDescriptor() (imgspecv1.Descriptor, error) {
 		}
 	}
 	if d == nil {
-		return imgspecv1.Descriptor{}, fmt.Errorf("no descriptor found for reference %q", ref.image)
+		return imgspecv1.Descriptor{}, errors.Wrap(ErrNotFoundDescriptor, ref.image)
 	}
 	return *d, nil
 }


rhatdan:
@vrothberg @mtrmac PTAL

vrothberg:
I think we need to fix it in `podman`.

`$ podman load -i FOO.tar name:tag` is supposed to load the image in `FOO.tar` and the tag it as `name:tag`.

zhangguanzhang:
> I think we need to fix it in `podman`.
> 
> `$ podman load -i FOO.tar name:tag` is supposed to load the image in `FOO.tar` and the tag it as `name:tag`.

you could see this https://github.com/containers/podman/issues/7337#issuecomment-675584694
I use the dlv to debug this and found the reason, so I commit this pr

mtrmac (changes requested):
I’m afraid I have no idea what is the problem, the solution, or how this makes any sense.

- Why make a check for `len(index.Manifests)`, and use the `[0]` member _inside_ a loop iterating through array?
- If the user asks for a specific `ref.image`, how does it make sense to return an entry that _does not have that name_?
- How does it make sense for a “find” operation to edit the data?

If this is to influence Podman’s heuristics for deciding how to tag the result of a pull, those are entirely up to Podman and c/image is trying fairly hard to stay away.

zhangguanzhang:
I change it，I need this error type to judge in podman

mtrmac:
> I’m afraid I have no idea what is the problem, the solution

I can’t see how this relates to the linked Podman issue at all. If Podman is using file names instead of other data it was supplied, why can’t it just _stop doing that_?

zhangguanzhang:
```
$ podman pull alpine
$ podman save --format oci-archive alpine:latest >| /tmp/FOO.tar
```
this will be ok
```
$ podman load -i /tmp/FOO.tar 
```
and this image name will be `docker.io/library/alpine:latest`, it will run into this line https://github.com/containers/image/blob/master/oci/layout/oci_transport.go#L190


but if use a name to override, afer this , the image's name should be` localhost/foo:latest`
```
$ podman load -i /tmp/FOO.tar foo 
Error: error pulling "foo": unable to pull dir:/tmp/FOO.tar: error determining pull goal for image "dir:/tmp/FOO.tar": error parsing dest reference name "localhost/tmp/FOO.tar": error parsing named reference "localhost/tmp/FOO.tar": invalid reference format: repository name must be lowercase
```
it does not run ok in the function in https://github.com/containers/podman/blob/master/libpod/runtime_img.go#L288-L294
Because in this case,  https://github.com/containers/image/blob/master/oci/layout/oci_transport.go#L205 ref.image is localhost/foo:latest, it is not equal to `docker.io/library/alpine:latest`

mtrmac:
> Because in this case,  https://github.com/containers/image/blob/master/oci/layout/oci_transport.go#L205 ref.image is localhost/foo:latest, it is not equal to `docker.io/library/alpine:latest`

So what’s actually wrong about that situation? If the ref is for _$path_`:foo`, and `foo` does not exist in the index, that’s that. 
- As far as the _end user_ is concerned, what should happen instead?
- How does that difference in actual vs. expected behavior manifest at the c/image API boundary?
- How does adding a recognizable error type help? What is Podman going to _do_ with that error? (I guess “try again with some other reference — but then why is it trying _two ways_ instead of just the _correct way_?)

(Perhaps see also the conversation around https://github.com/containers/podman/pull/6811#discussion_r473009515 , but I suspect the questions I have are at a more basic level.)

zhangguanzhang:
I think when `len(index.Manifests) == 1` and `ref.image != ""`， the image's name will be the ref.image after load the oci-archive file

rhatdan:
@mtrmac What is your latest feeling on this?

mtrmac:
https://github.com/containers/podman/pull/7386#issuecomment-678659544 – I don’t really understand what’s going on, even what’s _wrong_, my best guess is that Podman is internally confused about the meaning of the “name” parameter.

PR #1035, opened 2020-08-26
Set default rootless sigstore

QiWang19:
Set default rootless sigstore to ~/.local/share/containers/sigstore if the caller is non-root.
Fix rootless podman push --sign-by.

### DIFF
--- a/docker/docker_image_dest.go
+++ b/docker/docker_image_dest.go
@@ -78,12 +78,12 @@ func (d *dockerImageDestination) SupportsSignatures(ctx context.Context) error {
 		return err
 	}
 	switch {
-	case d.c.signatureBase != nil:
-		return nil
 	case d.c.supportsSignatures:
 		return nil
+	case d.c.signatureBase != nil:
+		return nil
 	default:
-		return errors.Errorf("X-Registry-Supports-Signatures extension not supported, and lookaside is not configured")
+		return errors.Errorf("Internal error: X-Registry-Supports-Signatures extension not supported, and lookaside should not be empty configuration")
 	}
 }
 
@@ -479,12 +479,12 @@ func (d *dockerImageDestination) PutSignatures(ctx context.Context, signatures [
 		return err
 	}
 	switch {
-	case d.c.signatureBase != nil:
-		return d.putSignaturesToLookaside(signatures, *instanceDigest)
 	case d.c.supportsSignatures:
 		return d.putSignaturesToAPIExtension(ctx, signatures, *instanceDigest)
+	case d.c.signatureBase != nil:
+		return d.putSignaturesToLookaside(signatures, *instanceDigest)
 	default:
-		return errors.Errorf("X-Registry-Supports-Signatures extension not supported, and lookaside is not configured")
+		return errors.Errorf("Internal error: X-Registry-Supports-Signatures extension not supported, and lookaside should not be empty configuration")
 	}
 }
 
@@ -502,9 +502,6 @@ func (d *dockerImageDestination) putSignaturesToLookaside(signatures [][]byte, m
 	// NOTE: Keep this in sync with docs/signature-protocols.md!
 	for i, signature := range signatures {
 		url := signatureStorageURL(d.c.signatureBase, manifestDigest, i)
-		if url == nil {
-			return errors.Errorf("Internal error: signatureStorageURL with non-nil base returned nil")
-		}
 		err := d.putOneSignature(url, signature)
 		if err != nil {
 			return err
@@ -517,9 +514,6 @@ func (d *dockerImageDestination) putSignaturesToLookaside(signatures [][]byte, m
 	// is sufficient.
 	for i := len(signatures); ; i++ {
 		url := signatureStorageURL(d.c.signatureBase, manifestDigest, i)
-		if url == nil {
-			return errors.Errorf("Internal error: signatureStorageURL with non-nil base returned nil")
-		}
 		missing, err := d.c.deleteOneSignature(url)
 		if err != nil {
 			return err
--- a/docker/docker_image_src.go
+++ b/docker/docker_image_src.go
@@ -297,12 +297,12 @@ func (s *dockerImageSource) GetSignatures(ctx context.Context, instanceDigest *d
 		return nil, err
 	}
 	switch {
-	case s.c.signatureBase != nil:
-		return s.getSignaturesFromLookaside(ctx, instanceDigest)
 	case s.c.supportsSignatures:
 		return s.getSignaturesFromAPIExtension(ctx, instanceDigest)
+	case s.c.signatureBase != nil:
+		return s.getSignaturesFromLookaside(ctx, instanceDigest)
 	default:
-		return [][]byte{}, nil
+		return nil, errors.Errorf("Internal error: X-Registry-Supports-Signatures extension not supported, and lookaside should not be empty configuration")
 	}
 }
 
@@ -336,9 +336,6 @@ func (s *dockerImageSource) getSignaturesFromLookaside(ctx context.Context, inst
 	signatures := [][]byte{}
 	for i := 0; ; i++ {
 		url := signatureStorageURL(s.c.signatureBase, manifestDigest, i)
-		if url == nil {
-			return nil, errors.Errorf("Internal error: signatureStorageURL with non-nil base returned nil")
-		}
 		signature, missing, err := s.getOneSignature(ctx, url)
 		if err != nil {
 			return nil, err
@@ -474,24 +471,19 @@ func deleteImage(ctx context.Context, sys *types.SystemContext, ref dockerRefere
 		return errors.Errorf("Failed to delete %v: %s (%v)", deletePath, string(body), delete.Status)
 	}
 
-	if c.signatureBase != nil {
-		manifestDigest, err := manifest.Digest(manifestBody)
+	manifestDigest, err := manifest.Digest(manifestBody)
+	if err != nil {
+		return err
+	}
+
+	for i := 0; ; i++ {
+		url := signatureStorageURL(c.signatureBase, manifestDigest, i)
+		missing, err := c.deleteOneSignature(url)
 		if err != nil {
 			return err
 		}
-
-		for i := 0; ; i++ {
-			url := signatureStorageURL(c.signatureBase, manifestDigest, i)
-			if url == nil {
-				return errors.Errorf("Internal error: signatureStorageURL with non-nil base returned nil")
-			}
-			missing, err := c.deleteOneSignature(url)
-			if err != nil {
-				return err
-			}
-			if missing {
-				break
-			}
+		if missing {
+			break
 		}
 	}
 
--- a/docker/lookaside.go
+++ b/docker/lookaside.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 
 	"github.com/containers/image/v5/docker/reference"
+	"github.com/containers/image/v5/internal/rootless"
 	"github.com/containers/image/v5/types"
 	"github.com/containers/storage/pkg/homedir"
 	"github.com/ghodss/yaml"
@@ -30,6 +31,12 @@ const builtinRegistriesDirPath = "/etc/containers/registries.d"
 // userRegistriesDirPath is the path to the per user registries.d.
 var userRegistriesDir = filepath.FromSlash(".config/containers/registries.d")
 
+// defaultUserDockerDir is the default sigstore directory for unprivileged user
+var defaultUserDockerDir = filepath.FromSlash(".local/share/containers/sigstore")
+
+// defaultDockerDir is the default sigstore directory for root
+var defaultDockerDir = "/var/lib/containers/sigstore"
+
 // registryConfiguration is one of the files in registriesDirPath configuring lookaside locations, or the result of merging them all.
 // NOTE: Keep this in sync with docs/registries.d.md!
 type registryConfiguration struct {
@@ -45,11 +52,18 @@ type registryNamespace struct {
 }
 
 // signatureStorageBase is an "opaque" type representing a lookaside Docker signature storage.
-// Users outside of this file should use configuredSignatureStorageBase and signatureStorageURL below.
-type signatureStorageBase *url.URL // The only documented value is nil, meaning storage is not supported.
+// Users outside of this file should use SignatureStorageBaseURL and signatureStorageURL below.
+type signatureStorageBase *url.URL
 
-// configuredSignatureStorageBase reads configuration to find an appropriate signature storage URL for ref, for write access if “write”.
-func configuredSignatureStorageBase(sys *types.SystemContext, ref dockerReference, write bool) (signatureStorageBase, error) {
+// SignatureStorageBaseURL reads configuration to find an appropriate signature storage URL for ref, for write access if “write”.
+// the usage of the BaseURL is defined under docker/distribution registries—separate storage of docs/signature-protocols.md
+// Warning: This function only exposes configuration in registries.d;
+// just because this function returns an URL does not mean that the URL will be used by c/image/docker (e.g. if the registry natively supports X-R-S-S).
+func SignatureStorageBaseURL(sys *types.SystemContext, ref types.ImageReference, write bool) (*url.URL, error) {
+	dr, ok := ref.(dockerReference)
+	if !ok {
+		return nil, errors.Errorf("ref must be a dockerReference")
+	}
 	// FIXME? Loading and parsing the config could be cached across calls.
 	dirPath := registriesDirPath(sys)
 	logrus.Debugf(`Using registries.d directory %s for sigstore configuration`, dirPath)
@@ -58,20 +72,23 @@ func configuredSignatureStorageBase(sys *types.SystemContext, ref dockerReferenc
 		return nil, err
 	}
 
-	topLevel := config.signatureTopLevel(ref, write)
-	if topLevel == "" {
-		return nil, nil
-	}
-
-	url, err := url.Parse(topLevel)
-	if err != nil {
-		return nil, errors.Wrapf(err, "Invalid signature storage URL %s", topLevel)
+	topLevel := config.signatureTopLevel(dr, write)
+	var url *url.URL
+	if topLevel != "" {
+		url, err = url.Parse(topLevel)
+		if err != nil {
+			return nil, errors.Wrapf(err, "Invalid signature storage URL %s", topLevel)
+		}
+	} else {
+		// returns default directory if no sigstore specified in configuration file
+		url = builtinDefaultSignatureStorageDir(rootless.GetRootlessEUID())
+		logrus.Debugf(" No signature storage configuration found for %s, using built-in default %s", dr.PolicyConfigurationIdentity(), url.String())
 	}
 	// NOTE: Keep this in sync with docs/signature-protocols.md!
 	// FIXME? Restrict to explicitly supported schemes?
-	repo := reference.Path(ref.ref) // Note that this is without a tag or digest.
-	if path.Clean(repo) != repo {   // Coverage: This should not be reachable because /./ and /../ components are not valid in docker references
-		return nil, errors.Errorf("Unexpected path elements in Docker reference %s for signature storage", ref.ref.String())
+	repo := reference.Path(dr.ref) // Note that this is without a tag or digest.
+	if path.Clean(repo) != repo {  // Coverage: This should not be reachable because /./ and /../ components are not valid in docker references
+		return nil, errors.Errorf("Unexpected path elements in Docker reference %s for signature storage", dr.ref.String())
 	}
 	url.Path = url.Path + "/" + repo
 	return url, nil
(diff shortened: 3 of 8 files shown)


QiWang19:
@mtrmac PTAL

QiWang19:
Add `GetRootlessUID()` to internal subpackage according to https://github.com/containers/image/pull/1027#issuecomment-679320458, can use unshare.GetRootlessUID() from containers/storage here?

docker/lookaside.go
```diff
@@ -175,6 +182,13 @@ func (config *registryConfiguration) signatureTopLevel(ref dockerReference, writ
 	if config.DefaultDocker != nil {
 		logrus.Debugf(` Using "default-docker" configuration`)
 		if url := config.DefaultDocker.signatureTopLevel(write); url != "" {
+			if uid := uid.GetRootlessUID(); uid != 0 {
+				if url != defaultDockerDir {
+					// use user configured default-docker
+					return url
+				}
+				return filepath.Join(homedir.Get(), defaultRootlessDockerDir)
```
mtrmac, line 190: It would be rather surprising if the value visible in the config file (and only one specific value) were silently overridden. The conversation in #1027 suggested that:
- We would ship config files without a default (or no config file at all); but users upgrading from an older release could/would retain the old configuration
- The code would contain a built-in default that only applied if the config files did not specify anything.

Also, the man page should be updated.

docker/lookaside.go
```diff
@@ -30,6 +31,12 @@ const builtinRegistriesDirPath = "/etc/containers/registries.d"
 // userRegistriesDirPath is the path to the per user registries.d.
 var userRegistriesDir = filepath.FromSlash(".config/containers/registries.d")
 
+// defaultRootlessDockerDir is the default sigstore directory for per use configuration
+var defaultRootlessDockerDir = filepath.FromSlash(".local/share/containers/sigstore")
```
mtrmac, line 35: `User`, `Nonroot`, `Unprivileged`, not `Rootless`, please; that means something more specific.

internal/uid/uid.go
```diff
@@ -0,0 +1,16 @@
+package uid
```
mtrmac, line 1: (Maybe `rootless`? Then this could be `rootless.GetEuid`.)

internal/uid/uid.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	"os"
+	"strconv"
+)
+
+// GetRootlessUID returns the UID of the user in the parent userNS
+func GetRootlessUID() int {
```
mtrmac, line 9: Please call this from `pkg/blobinfocache/default.go`, and drop the copy in that file.

internal/uid/uid.go
```diff
@@ -0,0 +1,16 @@
+package uid
+
+import (
+	"os"
+	"strconv"
+)
+
+// GetRootlessUID returns the UID of the user in the parent userNS
```
mtrmac, line 8:
```suggestion
// GetRootlessUID returns the UID of the current user (in the parent userNS, if any)
//
// Podman and similar software, in “rootless” configuration, when run as a non-root
// user, very early switches to a user namespace, where Geteuid() == 0 (but does not
// switch to a limited mount namespace); so, code relying on Geteuid() would use 
// system-wide paths in e.g. /var, when the user is actually not privileged to write to
// them, and expects state to be stored in the home directory.
//
// If Podman is setting up such a user namespace, it records the original UID in an
// environment variable, allowing us to make choices based on the actual user’s identity.
```
or something like that (if the above is accurate), to give newcomers more context.

internal/uid/uid.go
mtrmac, line 9: This should all talk about EUID, in all of function/variable names and comments, because that’s what the value apparently is (except the environment variable must not be renamed).

docker/lookaside.go
QiWang19: @mtrmac  PTAL.  Use the old configuration if the default configuration exists. And when euid not 0, only use the built-in default if there's no per-user configuration.

mtrmac (commented):
Oh… the rest of the client will now need to account for the state that sigstore is no longer “opt-in”. Especially after the `/var` path is added, the various code paths that assume `signatureStorageBase` can be `nil` become dead code.
- As the rest of the code is now, X-R-S-S would never be used, because sigstore is preferred; so we need to either prefer X-R-S-S (#385), or add a new configuration semantics to explicitly say “don’t use sigstore, use X-R-S-S or nothing”. I guess we should just do #385 ; the likelihood that a registry suddenly gains X-R-S-S is getting smaller every month.

QiWang19:
@mtrmac PTAL. I added the changes from #385. But the CI is not happy, will check the test failures.

docker/lookaside.go
```diff
@@ -30,6 +31,12 @@ const builtinRegistriesDirPath = "/etc/containers/registries.d"
 // userRegistriesDirPath is the path to the per user registries.d.
 var userRegistriesDir = filepath.FromSlash(".config/containers/registries.d")
 
+// defaultUserDockerDir is the default sigstore directory for unprivileged user
+var defaultUserDockerDir = filepath.FromSlash(".local/share/containers/sigstore")
+
+// defaultDockerDir is the default sigsore directory derived from registries.d
```
mtrmac, line 37:
```suggestion
// defaultDockerDir is the default sigstore directory for root
```

docs/containers-registries.d.5.md
```diff
@@ -77,6 +77,10 @@ described above.  The configuration section is a YAML mapping, with the followin
    This key is optional; if it is missing, no signature storage is defined (no signatures
    are download along with images, adding new signatures is possible only if `sigstore-staging` is defined).
 
+The default `sigstore-staging` is `/var/lib/containers/sigstore` from `/etc/containers/registries.d/default.yaml`.
+If the user has no privilege to write the signature to the default sigstore-staging, the YAML format files
```
mtrmac, line 81: That’s not the actual condition.

docs/containers-registries.d.5.md
```diff
@@ -77,6 +77,10 @@ described above.  The configuration section is a YAML mapping, with the followin
...
+under `$HOME/.config/containers/registries.d` can be configured, otherwise, `$HOME/.local/share/containers/sigstore` will be used.
```
mtrmac, line 82: All of this should be in the ”Registries, Scopes and Search Order” section, maybe in an extra named subsection.

docs/containers-registries.d.5.md
```diff
@@ -77,6 +77,10 @@ described above.  The configuration section is a YAML mapping, with the followin
...
+
 ## Examples
```
mtrmac, line 84: Consider adding an extra “example” for no configuration.

docker/lookaside.go
```diff
@@ -178,8 +185,16 @@ func (config *registryConfiguration) signatureTopLevel(ref dockerReference, writ
...
-	return ""
+	if euid := rootless.GetRootlessEUID(); euid != 0 {
+		// returns unprivileged default directory if no sigstore specified in configuration file
+		defaultUserDockerDirPath := filepath.Join(homedir.Get(), defaultUserDockerDir)
+		u := url.URL{Scheme: "file", Path: defaultUserDockerDirPath}
+		logrus.Debugf(" No signature storage configuration found for %s, use default unprivileged directory %s", ref.PolicyConfigurationIdentity(), u.String())
+		return u.String()
+	}
+	u := url.URL{Scheme: "file", Path: defaultUserDockerDir}
+	logrus.Debugf(" No signature storage configuration found for %s, use default %s", ref.PolicyConfigurationIdentity(), u.String())
```
mtrmac, line 196:
```suggestion
	logrus.Debugf(" No signature storage configuration found for %s, using default %s", ref.PolicyConfigurationIdentity(), u.String())
```
(in both cases)

docker/lookaside.go
```diff
@@ -178,8 +185,16 @@ func (config *registryConfiguration) signatureTopLevel(ref dockerReference, writ
 			return url
 		}
 	}
-	logrus.Debugf(" No signature storage configuration found for %s", ref.PolicyConfigurationIdentity())
-	return ""
+	if euid := rootless.GetRootlessEUID(); euid != 0 {
```
mtrmac, line 188: > `// config.signatureTopLevel returns an URL string configured in config for ref, for write access if “write”.`

i.e. this could return `nil` if `config` does not specify a configuration.

Moving these defaults up to `configuredSignatureStorageBase` would also avoid the unnecessary roundtrip between `url.URL`, a `string`, and back.

--- 

Either way the documentation of the function needs updating, if only to update the “if no signature storage should be used.” part.

docker/lookaside_test.go
```diff
@@ -28,13 +29,21 @@ func TestConfiguredSignatureStorageBase(t *testing.T) {
...
 	require.NoError(t, err)
 	defer os.RemoveAll(emptyDir)
 	base, err := configuredSignatureStorageBase(&types.SystemContext{RegistriesDirPath: emptyDir},
 		dockerRefFromString(t, "//this/is/not/in/the:configuration"), false)
+	expectBase := "file://"
+	if rootless.GetRootlessEUID() == 0 {
+		expectBase += filepath.Join(defaultDockerDir, "//this/is/not/in/the")
+	} else {
+		expectBase += filepath.Join(os.Getenv("HOME"), defaultUserDockerDir, "//this/is/not/in/the")
+	}
```
mtrmac, line 43: No, please test both: split `configuredSignatureStorageBaseForEUID` from `configuredSignatureStorageBase`, document that it only exists for tests, and test the `…ForEUID` variant here, for both the root and non-root cases. Or maybe split a `builtinDefaultSignatureStorageDir` with an EUID parameter, test both only for that one small helper, and here assume that tests are not running as root.

docker/lookaside_test.go
```diff
@@ -262,10 +271,16 @@ func TestRegistryConfigurationSignaureTopLevel(t *testing.T) {
 		},
 	}
 	dr := dockerRefFromString(t, "//thisisnotmatched")
+	expectDefault := "file://"
```
mtrmac, line 274: Moving the default directories from `signatureTopLevel` to `configuredSignatureStorageBase` would also allow not changing these tests.

docker/lookaside.go
```diff
@@ -178,8 +185,16 @@ func (config *registryConfiguration) signatureTopLevel(ref dockerReference, writ
...
+		// returns unprivileged default directory if no sigstore specified in configuration file
+		defaultUserDockerDirPath := filepath.Join(homedir.Get(), defaultUserDockerDir)
+		u := url.URL{Scheme: "file", Path: defaultUserDockerDirPath}
+		logrus.Debugf(" No signature storage configuration found for %s, use default unprivileged directory %s", ref.PolicyConfigurationIdentity(), u.String())
+		return u.String()
+	}
+	u := url.URL{Scheme: "file", Path: defaultUserDockerDir}
+	logrus.Debugf(" No signature storage configuration found for %s, use default %s", ref.PolicyConfigurationIdentity(), u.String())
+	return u.String()
```
mtrmac, line 197: After all this settles and tests work, the code and documentation that assumes `configuredSignatureStorageBase` can be `nil` should be removed/updated.

mtrmac:
> I added the changes from #385. But the CI is not happy, will check the test failures.

I think that’s because this adds only half of #385; the source implementation needs the change as well, or it will try reading the sigstore instead of X-R-S-S.

docs/containers-registries.d.5.md
```diff
@@ -61,6 +61,11 @@ more general scopes is ignored.  For example, if _any_ configuration exists for
 `docker.io/library/busybox`, the configuration for `docker.io` is ignored
 (even if some element of the configuration is defined for `docker.io` and not for `docker.io/library/busybox`).
 
+### Signature Storage
```
mtrmac, line 64: Maybe
```suggestion
### Built-in Defaults
```
everything is about signature storage in this file.

docs/containers-registries.d.5.md
```diff
@@ -61,6 +61,11 @@ more general scopes is ignored.  For example, if _any_ configuration exists for
...
+
+If default-docker not configured and no matched configuration applied for the container image can be found,
+the default directory, `/var/lib/containers/sigstore` or `$HOME/.local/share/containers/sigstore`,  will be used for reading and writing signatures.
```
mtrmac, line 67: This should say how the value is chosen.

mtrmac (commented):
ACK overall.

Would it be useful to make some variant of configuredSignatureStorageBase public for Podman? That’s up mostly to you, I think.

docker/lookaside.go
```diff
@@ -45,11 +52,11 @@ type registryNamespace struct {
 }
 
 // signatureStorageBase is an "opaque" type representing a lookaside Docker signature storage.
-// Users outside of this file should use configuredSignatureStorageBase and signatureStorageURL below.
+// Users outside of this file should use ConfiguredSignatureStorageBase and signatureStorageURL below.
 type signatureStorageBase *url.URL // The only documented value is nil, meaning storage is not supported.
```
mtrmac, line 56: This does not work for external callers if we don’t export `signatureStorageURL` (can the result even be used in Podman without an explicit `*url.URL` cast?)

… but there’s not _that much_ benefit to doing just that, because any caller would hard-code the semantics of the signature storage (in particular the sequential indices) anyway. We would need a very generic ReadSignatures/WriteSignatures/DeleteSignature API, which seems more effort than is strictly necessary right now.

So…

docker/lookaside.go
```diff
@@ -45,11 +52,11 @@ type registryNamespace struct {
...
 
-// configuredSignatureStorageBase reads configuration to find an appropriate signature storage URL for ref, for write access if “write”.
-func configuredSignatureStorageBase(sys *types.SystemContext, ref dockerReference, write bool) (signatureStorageBase, error) {
+// ConfiguredSignatureStorageBase reads configuration to find an appropriate signature storage URL for ref, for write access if “write”.
+func ConfiguredSignatureStorageBase(sys *types.SystemContext, ref dockerReference, write bool) (signatureStorageBase, error) {
```
mtrmac, line 59: … maybe name this something like `SignatureStorageBaseURL` (`RegistriesDSignatureStorageBaseURL`? `RegistriesDSigstoreBaseURL`?) and have it return an URL directly?

The documentation should also link to the “docker/distribution registries—separate storage” section of `docs/signature-protocols.md` for explanation of what a “base URL” is, and explicitly say something like
> Warning:  This function only exposes configuration in registries.d; just because this function returns an URL does not mean that the URL will be used by c/image/docker (e.g. if the registry natively supports X-R-S-S).

docker/lookaside.go
```diff
@@ -59,13 +66,16 @@ func configuredSignatureStorageBase(sys *types.SystemContext, ref dockerReferenc
 	}
 
 	topLevel := config.signatureTopLevel(ref, write)
+	var url *url.URL
 	if topLevel == "" {
```
mtrmac, line 70: (Absolutely non-blocking: It would be a tiny bit more readable to start with `topLevel != ""`, so that the uses of `topLevel` are close together.)

docker/lookaside.go
mtrmac, line 56: The `nil` value can not happen any longer.

docker/lookaside_test.go
```diff
@@ -313,3 +314,13 @@ func TestSignatureStorageBaseSignatureStorageURL(t *testing.T) {
 		assert.Equal(t, expectedURL, res, c.expected)
 	}
 }
+
+func TestBuiltinDefaultSignatureStorageDir(t *testing.T) {
+	base := builtinDefaultSignatureStorageDir(0)
+	assert.NotNil(t, base)
+	assert.Equal(t, "file://"+defaultDockerDir, (*url.URL)(base).String())
```
mtrmac, line 321: Are the `*url.URL` casts truly necessary here?

docker/docker_image_dest.go
```diff
@@ -78,10 +78,10 @@ func (d *dockerImageDestination) SupportsSignatures(ctx context.Context) error {
 		return err
 	}
 	switch {
-	case d.c.signatureBase != nil:
-		return nil
 	case d.c.supportsSignatures:
 		return nil
+	case d.c.signatureBase != nil:
+		return nil
 	default:
```
mtrmac, line 85: This is now unreachable; so this function can just be `return nil`, or the error returned might be “Internal error”+description of the inconsistency.

docker/docker_image_dest.go
```diff
@@ -479,10 +479,10 @@ func (d *dockerImageDestination) PutSignatures(ctx context.Context, signatures [
 		return err
 	}
 	switch {
-	case d.c.signatureBase != nil:
-		return d.putSignaturesToLookaside(signatures, *instanceDigest)
 	case d.c.supportsSignatures:
 		return d.putSignaturesToAPIExtension(ctx, signatures, *instanceDigest)
+	case d.c.signatureBase != nil:
+		return d.putSignaturesToLookaside(signatures, *instanceDigest)
 	default:
```
mtrmac, line 486: This is now unreachable; consistently with `SupportsSignatures`, either assume `d.c.signatureBase` is non-`nil`, or report an internal error.

docker/docker_image_src.go
```diff
@@ -297,10 +297,10 @@ func (s *dockerImageSource) GetSignatures(ctx context.Context, instanceDigest *d
 		return nil, err
 	}
 	switch {
-	case s.c.signatureBase != nil:
-		return s.getSignaturesFromLookaside(ctx, instanceDigest)
 	case s.c.supportsSignatures:
 		return s.getSignaturesFromAPIExtension(ctx, instanceDigest)
+	case s.c.signatureBase != nil:
+		return s.getSignaturesFromLookaside(ctx, instanceDigest)
 	default:
```
mtrmac, line 304: This is now unreachable, handle it consistently with `docker_image_dest`.

QiWang19:
@mtrmac PTAL

docker/docker_image_src.go
```diff
@@ -474,24 +474,22 @@ func deleteImage(ctx context.Context, sys *types.SystemContext, ref dockerRefere
 		return errors.Errorf("Failed to delete %v: %s (%v)", deletePath, string(body), delete.Status)
 	}
 
-	if c.signatureBase != nil {
-		manifestDigest, err := manifest.Digest(manifestBody)
+	manifestDigest, err := manifest.Digest(manifestBody)
```
mtrmac, line 477: OK; please document for `newDockerClientFromRef` that in the return value, `signatureBase` is always set.

docker/lookaside.go
```diff
@@ -196,13 +213,9 @@ func (ns registryNamespace) signatureTopLevel(write bool) string {
 	return ""
 }
 
-// signatureStorageURL returns an URL usable for acessing signature index in base with known manifestDigest, or nil if not applicable.
-// Returns nil iff base == nil.
+// signatureStorageURL returns an URL usable for acessing signature index in base with known manifestDigest.
 // NOTE: Keep this in sync with docs/signature-protocols.md!
 func signatureStorageURL(base signatureStorageBase, manifestDigest digest.Digest, index int) *url.URL {
```
mtrmac, line 218: The checks for the return value being `nil` in callers can be removed as well.

docker/lookaside.go
QiWang19: the `dockerReference` also needs to be exported

mtrmac: Exporting that would not directly help; users of `alltransports.ParseImageName` or `docker.NewReference` only get a generic `types.ImageReference` interface.

`GetRepositoryTags` already handles that by accepting a `types.ImageReference` and making a runtime check. That’s ugly but convenient enough for callers, and impossible to misuse.

QiWang19: Thanks.  Add the same type assertion check inside `SignatureStorageBaseURL`

docker/docker_image_dest.go
QiWang19: I keep them as internal errors.

QiWang19:
@mtrmac Can I get a review before you leave for PTO?

docs/containers-registries.d.5.md
```diff
@@ -61,6 +61,11 @@ more general scopes is ignored.  For example, if _any_ configuration exists for
 `docker.io/library/busybox`, the configuration for `docker.io` is ignored
 (even if some element of the configuration is defined for `docker.io` and not for `docker.io/library/busybox`).
 
+### Built-in Defaults
+
+If default-docker not configured and no matched configuration applied for the container image can be found,
```
mtrmac, line 66: (Maybe something like
```suggestion
If no `docker` section can be found for the container image, and no `default-docker` section is configured
```
to go from the most-specific to the most general, in the “order of evaluation”)

mtrmac (commented):
ACK, please rebase on top of the recent typo fix.

(And ideally make sure Podman can benefit from `SignatureStorageBaseURL`.)

mtrmac:
LGTM after dropping the  `docker/docker_transport.go` part, actually.

mtrmac (approved):
Thanks!

Issue #1007, opened 2019-10-18
Work around docker/distribution servers that don't include the JSON body on error


bparry02:
I have a docker-distribution registry configured with `storage.maintenance.readonly.enabled: true` which prevents writes to the registry. When I run `skopeo copy`, I see an error message that fails to parse the body of the response.

Instead, I expect to see a simple message, such as "copy: Method not allowed" or something similar.

```
# skopeo copy --dest-tls-verify=false docker-daemon:registry.redhat.io/rhel7.7:latest docker://127.0.0.1:5000/rhel7.7:latest
Getting image source signatures
Copying blob 506371554037 done
Copying blob 893590530078 done
FATA[0002] Error writing blob: Error initiating layer upload to /v2/rhel7.7/blobs/uploads/ in 127.0.0.1:5000: error parsing HTTP 405 response body: invalid character 'M' looking for beginning of value: "Method not allowed\n" 
```

rhatdan:
@mtrmac Are you planning on working on this, or should we put @QiWang19 on it?

mtrmac:
@rhatdan I probably won’t work on this this week. (Most of the error handling in c/image/docker is already fairly centralized, so this should be a fairly small change.)

rhatdan:
@QiWang19 PTAL

rhatdan:
@QiWang19 PTAL

QiWang19:
What workaround is needed? Define `error parsing HTTP 405 response body: invalid character 'M' looking for beginning of value: "Method not allowed` as an error and check if the failure is caused by that kind of error?

mtrmac:
That’s already a typed error (`UnexpectedHTTPResponseError` I think), and it contains the full HTTP response body.

A workaround would be to check if that body looks like a non-JSON plain text string, or possibly HTML, and output some other kind of error. In a sense it would be nice to preserve the `UnexpectedHTTPResponseError` type and only wrap it, but that would not get rid of the visual noise, and we don’t document that type — so, probably just a plain-text error that contains the HTTP status and possibly the body if it is not too big.

PR #1037, opened 2020-08-28
Return error body if UnexpectedHTTPResponseError

QiWang19:
Format the error message use the response body if the original error is returned as client.UnexpectedHTTPResponseError type

Use this pach:

```
$ skopeo copy --dest-tls-verify=false docker://library/alpine:latest docker://127.0.0.1:5000/cp:latest

FATA[0000] Error writing blob: Error initiating layer upload to /v2/cp1/blobs/uploads/ in 127.0.0.1:5000: StatusCode: 405, Method not allowed
```

### DIFF
--- a/docker/docker_image_dest.go
+++ b/docker/docker_image_dest.go
@@ -22,7 +22,6 @@ import (
 	"github.com/containers/image/v5/types"
 	"github.com/docker/distribution/registry/api/errcode"
 	v2 "github.com/docker/distribution/registry/api/v2"
-	"github.com/docker/distribution/registry/client"
 	"github.com/opencontainers/go-digest"
 	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
@@ -154,7 +153,7 @@ func (d *dockerImageDestination) PutBlob(ctx context.Context, stream io.Reader, 
 	defer res.Body.Close()
 	if res.StatusCode != http.StatusAccepted {
 		logrus.Debugf("Error initiating layer upload, response %#v", *res)
-		return types.BlobInfo{}, errors.Wrapf(client.HandleErrorResponse(res), "Error initiating layer upload to %s in %s", uploadPath, d.c.registry)
+		return types.BlobInfo{}, errors.Wrapf(registryHTTPResponseToError(res), "Error initiating layer upload to %s in %s", uploadPath, d.c.registry)
 	}
 	uploadLocation, err := res.Location()
 	if err != nil {
@@ -175,7 +174,7 @@ func (d *dockerImageDestination) PutBlob(ctx context.Context, stream io.Reader, 
 		}
 		defer res.Body.Close()
 		if !successStatus(res.StatusCode) {
-			return nil, errors.Wrapf(client.HandleErrorResponse(res), "Error uploading layer chunked")
+			return nil, errors.Wrapf(registryHTTPResponseToError(res), "Error uploading layer chunked")
 		}
 		uploadLocation, err := res.Location()
 		if err != nil {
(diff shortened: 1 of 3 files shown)


QiWang19:
close #1007

QiWang19:
@mtrmac PTAL

mtrmac (commented):
Thanks!

- `client.HandleErrorResponse` should only be called once, because it reads `res.Body`.
- The `errors.New(string(e.Response))` is not going to be useful for callers to specifically look for anyway; we might just as well use `fmt.Errorf` or something like that.
- The error should include `err.StatusCode`
- If I’m reading the code right (I haven’t tested it), it seems that the `Response` string may be very long, and HTML (compare https://bugzilla.redhat.com/show_bug.cgi?id=1522707 or https://github.com/goharbor/harbor/issues/8203 ); maybe limit it to a few hundred characters, to be sure.

QiWang19:
Can I get the HTML instead of using the Response to avoid the long erro rmessages?

docker/docker_image_dest.go
```diff
@@ -154,7 +154,11 @@ func (d *dockerImageDestination) PutBlob(ctx context.Context, stream io.Reader,
 	defer res.Body.Close()
 	if res.StatusCode != http.StatusAccepted {
 		logrus.Debugf("Error initiating layer upload, response %#v", *res)
-		return types.BlobInfo{}, errors.Wrapf(client.HandleErrorResponse(res), "Error initiating layer upload to %s in %s", uploadPath, d.c.registry)
+		err = client.HandleErrorResponse(res)
+		if e, ok := errors.Cause(err).(*client.UnexpectedHTTPResponseError); ok {
+			return types.BlobInfo{}, errors.Wrapf(fmt.Errorf("%s, StatusCode: %d", string(e.Response), e.StatusCode), "Error initiating layer upload to %s in %s", uploadPath, d.c.registry)
```
mtrmac, line 159: Perhaps start with the status code, when `e.Response` can be arbitrarily large. “HTTP status %d: %s” or something like that.

This can be further simplified to avoid duplicating the “Error initiating layer upload” text with `err = fmt.Errorf(…)` and letting the code continue.

mtrmac:
> Can I get the HTML instead of using the Response to avoid the long erro rmessages?

(To be sure we are talking about the same thing, `e.Response` would contain the HTML) Turning a HTML page into a single one-line string is going to be difficult in general; even after stripping the various HTML tags the result can be pretty much anything. (“Something went wrong … Cute picture of a fail whale … we are sorry … [bottom-page set of links for SEO] … copyright”).

I think we should to impose a maximum length (just silently truncating, or at most indicating with …, is good enough of an improvement); parsing HTML could be nice but goes _way_ beyond the minimum necessary for dealing with the reported issue — and if we did do that, it should be used for _all_ error cases in the package, not just this particular one.

---

Actually, please factor this change out into a `client.HandleErrorResponse` wrapper/replacement used throughout c/image/docker, not just on this single HTTP request.

QiWang19:
@mtrmac PTAL

docker/errors.go
```diff
@@ -44,3 +44,16 @@ func httpResponseToError(res *http.Response, context string) error {
 		return perrors.Errorf("%sinvalid status code from registry %d (%s)", context, res.StatusCode, http.StatusText(res.StatusCode))
 	}
 }
+
+// generateHttpResponseError check if the error is UnexpectedHTTPResponseError and generate the error message
```
mtrmac, line 48:
```suggestion
// generateHttpResponseError creates a Go error from an HTTP error response of a docker/distribution
// registry
```
or something like that?

Also perhaps name this `registryHTTPResponseToError` or something similar, to be consistent but distinct form `httpResponseToError` above.

mtrmac (commented):
Thanks! ACK to the code, I’ll check the call sites tomorrow.

TomSweeneyRedHat:
LGTM

mtrmac (approved):
LGTM. Thanks again!

Issue #1017, opened 2020-08-12
oci-archive can't lchown when untarring


kitforbes:
Hi folks,

We've got a rootless image created with Buildah. The image (with a bunch of irrelevant stuff removed) looks like this
```dockerfile
FROM registry.fedoraproject.org/fedora:32

RUN set -ex; \
  useradd deploy && \
  yum update -y -d 1 && yum install -y -d 1 \
    skopeo \
  && \
  yum clean all && rm -rf /var/cache /var/log/dnf* /var/log/yum.*

USER deploy
WORKDIR /home/deploy
```
The Skopeo version is 1.1.1.

I'm running the image locally through Docker with
```bash
docker run --rm -it --privileged image-name /bin/bash
```

I've downloaded a `tar` which was previously created with `skopeo copy` to create an OCI archive. Then, within the container I'm trying to run
```bash
skopeo copy oci-archive:container.tar docker://somewhere
```
Which results in the error of 
>FATA[0000] Error initializing source oci-archive:container.tar:: error creating temp directory: error untarring file "/var/tmp/oci564737160": lchown /var/tmp/oci564737160/blobs: operation not permitted

I've tried changing the directory in which the untarring occurs with the `--tmpdir` flag, but that hasn't helped so far. I was wondering if someone could point me in the right direction to make this work, and I appreciate that this may just be a Linux issue rather than an issue with Skopeo or one of the libraries. Most likely just my lack of knowledge in this area.

Thanks for your help.

rudolpho1:
Hi @mtrmac do you have any update on this? This issue is a blocker for us as we want to use skopeo to automatically deploy new container images to the production system. And we cannot run it as root. Thanks a lot for your help!

TomSweeneyRedHat:
LGTM
but definitely want a @mtrmac or @vrothberg head nod

kitforbes:
Thanks for the fast turn around folks!

PR #1039, opened 2020-09-03
Set NoLchown to true in untar opts

QiWang19:
Pass NoLchown in the Taroptions to avoid the lchown permission denied error dealing with archive.
Fix https://github.com/containers/image/issues/1017

### DIFF
--- a/oci/archive/oci_transport.go
+++ b/oci/archive/oci_transport.go
@@ -183,7 +183,12 @@ func createUntarTempDir(sys *types.SystemContext, ref ociArchiveReference) (temp
 	src := ref.resolvedFile
 	dst := tempDirRef.tempDirectory
 	// TODO: This can take quite some time, and should ideally be cancellable using a context.Context.
-	if err := archive.UntarPath(src, dst); err != nil {
+	arch, err := os.Open(src)
+	if err != nil {
+		return tempDirOCIRef{}, err
+	}
+	defer arch.Close()
+	if err := archive.NewDefaultArchiver().Untar(arch, dst, &archive.TarOptions{NoLchown: true}); err != nil {
 		if err := tempDirRef.deleteTempDir(); err != nil {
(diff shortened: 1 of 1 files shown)


rhatdan:
@mtrmac @vrothberg PTAL
LGTM

oci/archive/oci_transport.go
```diff
@@ -183,7 +183,12 @@ func createUntarTempDir(sys *types.SystemContext, ref ociArchiveReference) (temp
 	src := ref.resolvedFile
 	dst := tempDirRef.tempDirectory
 	// TODO: This can take quite some time, and should ideally be cancellable using a context.Context.
-	if err := archive.UntarPath(src, dst); err != nil {
+	archiver, err := os.Open(src)
```
mtrmac, line 186:
```suggestion
	archive, err := os.Open(src)
```

mtrmac (commented):
Looks good to me, but I’ll defer to @nalind WRT using the API correctly / optimally.

nalind:
My guess is that we don't even want to try to preserve ownership information from the archive, and this accomplishes that.  Using one of the "WithChown" variations would just have us chowning things to same value they get when we create them, so this saves us a bit of work in comparison.

LGTM

rhatdan:
@QiWang19 Fix @mtrmac suggestion and we can get this merged.

TomSweeneyRedHat:
LGTM
once the nit is addressed.

oci/archive/oci_transport.go
QiWang19: use arch instead of archive since there's name confliction with package storage/pkg/archive

QiWang19:
The suggestion fixed.
@mtrmac @nalind @TomSweeneyRedHat @rhatdan PTAL

rhatdan:
LGTM

PR #1041, opened 2020-09-07
Compare destination image manifest copy optimization/feature

gjenkins8:
This PR optimizes image copies for the scenario where the destination image is expected to exist.

The scenario occurs for a common usage of `skopeo sync`: when periodically synchronizing two registries. Where the set of source images are (mostly) unchanged, and already exist at the destination. For this scenario, fetching the destination image's manifest, and checking if it matches the source manifest. And skipping the 'copy' (ie. layer content checks and manifest update) if equivalent provides a performance improvement.

This check should be an opt-in. As the default scenario for image copying is assumedly copying an image to a new destination. And the manifest fetch is pessimistic in this case. The copy is also never skipped if the copy requires signing or encryption (through more work could be done to skip the copy if the signing or encryption are un-changed).

More details / related issue:
https://github.com/containers/skopeo/issues/1021

Thanks!

Sample output when copy is skipped:
```
INFO[0004] Copying image tag 2/42                        from="docker://registry.company.com/subpath/foo:0.0.18" to="docker://registry.company.com/synctest/foo:0.0.18"
Skipping: image already present at destination
```

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -194,6 +194,10 @@ type Options struct {
 	OciDecryptConfig *encconfig.DecryptConfig
 	// MaxParallelDownloads indicates the maximum layers to pull at the same time.  A reasonable default is used if this is left as 0.
 	MaxParallelDownloads uint
+	// When OptimizeDestinationImageAlreadyExists is set, optimize the copy assuming that the destination image already
+	// exists (and is equivalent). Making the eventual (no-op) copy more performant for this case. Enabling the option
+	// is slightly pessimistic if the destination image doesn't exist, or is not equivalent.
+	OptimizeDestinationImageAlreadyExists bool
 }
 
 // validateImageListSelection returns an error if the passed-in value is not one that we recognize as a valid ImageListSelection value
@@ -359,6 +363,45 @@ func supportsMultipleImages(dest types.ImageDestination) bool {
 		}
 	}
 	return false
+}
+
+// compareImageDestinationManifestEqual compares the `src` and `dest` image manifests (reading the manifest from the
+// (possibly remote) destination). Returning true and the destination's manifest, type and digest if they compare equal.
+func compareImageDestinationManifestEqual(ctx context.Context, options *Options, src types.Image, targetInstance *digest.Digest, dest types.ImageDestination) (bool, []byte, string, digest.Digest, error) {
+	srcManifest, _, err := src.Manifest(ctx)
+	if err != nil {
+		return false, nil, "", "", errors.Wrapf(err, "Error reading manifest from image")
+	}
+
+	srcManifestDigest, err := manifest.Digest(srcManifest)
+	if err != nil {
+		return false, nil, "", "", errors.Wrapf(err, "Error calculating manifest digest")
+	}
+
+	destImageSource, err := dest.Reference().NewImageSource(ctx, options.DestinationCtx)
+	if err != nil {
+		logrus.Debugf("Unable to create destination image %s source: %v", dest.Reference(), err)
+		return false, nil, "", "", nil
+	}
+
+	destManifest, destManifestType, err := destImageSource.GetManifest(ctx, targetInstance)
+	if err != nil {
+		logrus.Debugf("Unable to get destination image %s/%s manifest: %v", destImageSource, targetInstance, err)
+		return false, nil, "", "", nil
+	}
+
+	destManifestDigest, err := manifest.Digest(destManifest)
+	if err != nil {
+		return false, nil, "", "", errors.Wrapf(err, "Error calculating manifest digest")
+	}
+
+	logrus.Debugf("Comparing source and destination manifest digests: %v vs. %v", srcManifestDigest, destManifestDigest)
+	if srcManifestDigest != destManifestDigest {
+		return false, nil, "", "", nil
+	}
+
+	// Destination and source manifests, types and digests should all be equivalent
+	return true, destManifest, destManifestType, destManifestDigest, nil
 }
 
 // copyMultipleImages copies some or all of an image list's instances, using
@@ -646,6 +689,26 @@ func (c *copier) copyOneImage(ctx context.Context, policyContext *signature.Poli
 	// If encrypted and decryption keys provided, we should try to decrypt
 	ic.diffIDsAreNeeded = ic.diffIDsAreNeeded || (isEncrypted(src) && ic.c.ociDecryptConfig != nil) || ic.c.ociEncryptConfig != nil
 
+	// If enabled, fetch and compare the destination's manifest. And as an optimization skip updating the destination iff equal
+	if options.OptimizeDestinationImageAlreadyExists {
+		shouldUpdateSigs := len(sigs) > 0 || options.SignBy != "" // TODO: Consider allowing signatures updates only and skipping the image's layers/manifest copy if possible
+		noPendingManifestUpdates := ic.noPendingManifestUpdates()
+
+		logrus.Debugf("Checking if we can skip copying: has signatures=%t, OCI encryption=%t, no manifest updates=%t", shouldUpdateSigs, destRequiresOciEncryption, noPendingManifestUpdates)
+		if !shouldUpdateSigs && !destRequiresOciEncryption && noPendingManifestUpdates {
+			isSrcDestManifestEqual, retManifest, retManifestType, retManifestDigest, err := compareImageDestinationManifestEqual(ctx, options, src, targetInstance, c.dest)
+			if err != nil {
+				logrus.Warnf("Failed to compare destination image manifest: %v", err)
+				return nil, "", "", err
+			}
+
+			if isSrcDestManifestEqual {
+				c.Printf("Skipping: image already present at destination\n")
+				return retManifest, retManifestType, retManifestDigest, nil
+			}
+		}
+	}
+
 	if err := ic.copyLayers(ctx); err != nil {
 		return nil, "", "", err
 	}
@@ -784,6 +847,10 @@ func (ic *imageCopier) updateEmbeddedDockerReference() error {
 	return nil
 }
 
+func (ic *imageCopier) noPendingManifestUpdates() bool {
+	return reflect.DeepEqual(*ic.manifestUpdates, types.ManifestUpdateOptions{InformationOnly: ic.manifestUpdates.InformationOnly})
+}
+
 // isTTY returns true if the io.Writer is a file and a tty.
 func isTTY(w io.Writer) bool {
 	if f, ok := w.(*os.File); ok {
@@ -904,6 +971,8 @@ func (ic *imageCopier) copyLayers(ctx context.Context) error {
 		diffIDs[i] = cld.diffID
 	}
 
+	// WARNING: If you are adding new reasons to change ic.manifestUpdates, also update the
+	// OptimizeDestinationImageAlreadyExists short-circuit conditions
 	ic.manifestUpdates.InformationOnly.LayerInfos = destInfos
 	if ic.diffIDsAreNeeded {
 		ic.manifestUpdates.InformationOnly.LayerDiffIDs = diffIDs
@@ -932,7 +1001,7 @@ func layerDigestsDiffer(a, b []types.BlobInfo) bool {
 // and its digest.
 func (ic *imageCopier) copyUpdatedConfigAndManifest(ctx context.Context, instanceDigest *digest.Digest) ([]byte, digest.Digest, error) {
 	pendingImage := ic.src
-	if !reflect.DeepEqual(*ic.manifestUpdates, types.ManifestUpdateOptions{InformationOnly: ic.manifestUpdates.InformationOnly}) {
+	if !ic.noPendingManifestUpdates() {
 		if !ic.canModifyManifest {
 			return nil, "", errors.Errorf("Internal error: copy needs an updated manifest but that was known to be forbidden")
 		}
@@ -1283,6 +1352,8 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
 	}
 
 	// === Deal with layer compression/decompression if necessary
+	// WARNING: If you are adding new reasons to change the blob, update also the OptimizeDestinationImageAlreadyExists
+	// short-circuit conditions
 	var inputInfo types.BlobInfo
 	var compressionOperation types.LayerCompression
 	uploadCompressionFormat := &c.compressionFormat


mtrmac (commented):
Thanks!

I haven’t checked how this integrates into the full process, for now just quick notes:
- Please sign your commit per https://github.com/containers/image/blob/master/CONTRIBUTING.md#sign-your-prs
- Looking at the test failures, failure of `destImageSource.GetManifest` should not be fatal (it is fine, and quite expected, for the destination not to contain any image before a copy happens).
- The opt-in to do this check should eventually be exposed in `copy.Options`.

copy/copy.go
```diff
@@ -577,6 +577,29 @@ func (c *copier) copyOneImage(ctx context.Context, policyContext *signature.Poli
 		}
 	}
 
+	destExistsCheck := true // TBD: enable configuration
```
TomSweeneyRedHat, line 580: Maybe something in [containers.conf](https://github.com/containers/common/blob/master/pkg/config/containers.conf) eventually?

copy/copy.go
```diff
@@ -577,6 +577,29 @@ func (c *copier) copyOneImage(ctx context.Context, policyContext *signature.Poli
...
+	if destExistsCheck {
+		destImageSource, err := c.dest.Reference().NewImageSource(ctx, options.DestinationCtx)
+
+		if err == nil {
+
+			retManifest, retManifestType, err = destImageSource.GetManifest(ctx, nil)
+			if err != nil {
+				return nil, "", "", errors.Wrapf(err, "Error reading manifest from dest image %s", transports.ImageName(destImageSource.Reference()))
```
TomSweeneyRedHat, line 588:
```suggestion
				return nil, "", "", errors.Wrapf(err, "Error reading manifest from destination image %s", transports.ImageName(destImageSource.Reference()))
```

copy/copy.go
```diff
@@ -577,6 +577,29 @@ func (c *copier) copyOneImage(ctx context.Context, policyContext *signature.Poli
...
+			}
+
+			sourceManifest, _, err := src.Manifest(ctx)
+			if err != nil {
+				return nil, "", "", errors.Wrapf(err, "Error reading manifest from source image")
+			}
+
+			if bytes.Compare(sourceManifest, retManifest) == 0 {
+				c.Printf("Dest image manifest compares equal\n")
```
TomSweeneyRedHat, line 597: Soft suggestion here and above for dest -> destination
```suggestion
				c.Printf("Destination image manifest compares equal\n")
```

TomSweeneyRedHat:
Will need a rebase at some point @gjenkins8 , nice start!

copy/copy.go
mtrmac: No, this is not a system-wide decision; whether to do the check (= optimistically assume that the copy is a no-op) it is individual to the specific operation (is this a push of a just-built image that does not exist anywhere else vs. a daily sync of a rarely-changing large set of repositories).

So, for c/image this should be only a Go option with no user configuration, and I guess _most_ callers should leave this off. `skopeo sync` can start with hard-coding to on. Eventually we might encounter commands or cases where the right default is not obvious based on just the command, and where the difference is big enough, and that software can expose the option in the CLI or in individual config files.

copy/copy.go
mtrmac: BTW most of the code in this file does not say source/destination at all, the image name + operation should be sufficient.

mtrmac: … but anyway we need to carefully deal with correctness first, and see what text is necessary for that, before fine-tuning it.

(e.g. — note to self — we still need to do the copy if the source contains signatures. There may be more.)

copy/copy.go
gjenkins8: Thanks @TomSweeneyRedHat , @mtrmac, I have created an option `DestinationManifestCheck` in `copy.Options` (suggestions on naming completely welcome, there is likely a more descriptive name out there)

copy/copy.go
gjenkins8: Thanks! I've also amend my initial PR note with a sample output

copy/copy.go
mtrmac: Maybe something like `AssumeDestinationExists`? The API should describe the situation / assumption / desired behavior, not details of the implementation.

(Anyway, I’m afraid I still need to look into the details of correctness.)

mtrmac:
As a very quick guess without actually checking, skipping the copy also skips the opportunity to add a signature via `--sign-by`. I guess some _earlier_ copy was optimized out, causing the image to not be signed as expected; this error complains about a different signature that does exist; if there were a valid signature, that complaint would not be reported to the user.

copy/copy.go
```diff
@@ -546,6 +551,25 @@ func (c *copier) copyOneImage(ctx context.Context, policyContext *signature.Poli
 		return nil, "", "", errors.Wrapf(err, "Error initializing image from source %s", transports.ImageName(c.rawSource.Reference()))
 	}
 
+	options.DestinationManifestCheck = true // TODO: Forced enablement for initial PR testing
+	if options.DestinationManifestCheck && options.SignBy == "" {
```
gjenkins8, line 555: Should verify that `options.DestinationManifestCheck` is not set, if `options.SignBy` is set? Since both of these options are exclusive (with the current implementation).

The alternative would be to leave this decision to the client (which might choose to only set `options.DestinationManifestCheck` if `options.SignBy` is not set). And alway skip the copy exclusively based upon `options.DestinationManifestCheck`.

I don't think `options.DestinationManifestCheck && options.SignBy == "" ` is the right approach, as it allows the client to request skipping the manifest copy when that is not possible.

mtrmac, line 555: The full set of conditions AFAICS:
- Caller opts in
- No need to edit the image/manifest (per the `reflect.DeepEqual` check in `copyUpdatedConfigAndManifest`, which is non-trivial enough that it should probably be split into a separate `ic.noPendingManifestUpdates` helper instead of copy&pasted):
  - No manifest conversion is needed
  - No need to update the embedded Docker reference
- No need to copy existing signatures (`len(sigs) == 0`, or alternatively all of these signatures are known to exist at the destination as well — that can be left to a future PR)
- No new signature (`opts.SignBy == ""`)
- No OCI encryption/decryption (`copier.oci*Config`)

I’m a bit torn on whether to have the conditions inlined here or in a helper function; I'm leaning towards inlining, to make the logic easier to see and to make it more difficult to overlook that this optimization exists.

The “destination already contains the desired manifest” check, OTOH, is conceptually trivial and would be better split into a helper function, to make `copyOneImage` a little less gargantuan.

copy/copy.go
```diff
@@ -546,6 +551,25 @@ func (c *copier) copyOneImage(ctx context.Context, policyContext *signature.Poli
...
+		destImageSource, err := c.dest.Reference().NewImageSource(ctx, options.DestinationCtx)
+		if err == nil {
+			retManifest, retManifestType, err = destImageSource.GetManifest(ctx, nil)
```
mtrmac, line 558: This should specify `targetInstance` so that it benefits multi-arch images as well (and correctly handles updates to a tag between single-arch/multi-arch images).

copy/copy.go
```diff
@@ -546,6 +551,25 @@ func (c *copier) copyOneImage(ctx context.Context, policyContext *signature.Poli
...
+			if err == nil {
+				sourceManifest, _, err := src.Manifest(ctx)
+				if err != nil {
+					return nil, "", "", errors.Wrapf(err, "Error reading manifest from image")
+				}
+
+				if bytes.Equal(sourceManifest, retManifest) {
+					c.Printf("Destination image manifest compares equal\n")
```
mtrmac, line 566: Maybe
```suggestion
					c.Printf("Destination already contains the expected manifest, not copying the image again\n")
```
or something a bit succinct — but the goal is that the reader understands why no layer/manifest copies show up in the debug output.

mtrmac (commented):
Finally a full review.

copy/copy.go
```diff
@@ -546,6 +551,25 @@ func (c *copier) copyOneImage(ctx context.Context, policyContext *signature.Poli
 		return nil, "", "", errors.Wrapf(err, "Error initializing image from source %s", transports.ImageName(c.rawSource.Reference()))
 	}
 
+	options.DestinationManifestCheck = true // TODO: Forced enablement for initial PR testing
+	if options.DestinationManifestCheck && options.SignBy == "" {
+		destImageSource, err := c.dest.Reference().NewImageSource(ctx, options.DestinationCtx)
+		if err == nil {
+			retManifest, retManifestType, err = destImageSource.GetManifest(ctx, nil)
+			if err == nil {
```
mtrmac, line 559: Non-blocking: For the two error cases above, it would probably be useful to at least log them in the debug output.

copy/copy.go
mtrmac: The way I think about it, the caller is not making a _request_ that the code here must honor — the caller does not want the `copy.Image` call to _fail_ just because a full image must be copied. The caller is only is providing `AssumeImageAlreadyExistsAtDestination` information that this code could completely ignore, or use in any way it wants to optimize the behavior. 

From that point of view, it is perfectly fine for the client to provide “image probably already exists” information, together with a request for an operation (format conversion, encryption, …) which we don’t currently support in the optimization. Maybe such an optimization can be implemented in the future; maybe not. The caller can provide that information just the same, and if some future optimization arrives, the caller will benefit at that time.

mtrmac:
> @mtrmac -- I have done the naive thing, and only attempt to skip copying iff signing is not requested (`options.SignBy == ""`).
> 
> I don't know enough about signing to know if copying can be skipped (manifests compare equal), but the destination image can be re-signed.

Signatures are “attachments” to an image, so it is possible to first create an image, and later add signatures (or even more signatures) to it. So, the presence of a manifest does not automatically imply no need to copy signatures.

For `docker://`, it would be possible to just skip the layer+config+manifest copy, and then add signatures; but right now, the `ImageDestination` interface specifies that layer/config/manifest/signature updates must all happen in that order, so it is not valid to just update signatures. Some cases do arise from time to time where it would be interesting to relax this restriction, but I think we shouldn’t block this PR on an API change like that.

It _would_ be possible, if `options.SignBy` is not set but `sigs` exists, for the optimization to read signatures at the destination (the `ImageSource` interface has no ordering requirements, unlike `ImageDestination`), and to skip the image copy if both the manifest and the sets of signatures are satisfactory. Unless this actually comes up in your use, I think it’s fine to defer that to a separate PR.

> Also, my PR is a little awkward. In order to get integration/test coverage, I need to 'force' `options.DestinationManifestCheck` to be true (as it is essentially acting as a feature flag). But of course, this would be don't by the client (`skopeo`) in practice.

ACK; that’s a great way to test it now, we just shouldn’t merge it like that.

> How would you like to proceed? I can remove the 'feature enablement' from this PR, and Skopeo can be updated to set `options.DestinationManifestCheck = true`? Or, I can make a new PR to add `options.DestinationManifestCheck`, a second PR to `skopeo` to enable the option. And this PR would be the 3rd to implement the behaviour?

The 3-PR version would still have to wait for an update of c/image in Skopeo, so I don’t think it speeds anything up. Let’s
- test this PR with the hard-coded enablement
- Remove the enablement, and merge into c/image
- (Make a new c/image release? That would be cleaner but more overhead.)
- Vendor c/image into Skopeo (either a release, or a git commit)
- In the same vendoring PR, or later, set the flag in `skopeo sync` (and add extra tests?)

The 3-PR version would be helpful if extra Skopeo tests found problems in the implementation; but in the meantime we would have a do-nothing flag in c/image, possibly even in a release. Maybe we can afford to be optimistic and assume that there won’t be huge problems only found later.

copy/copy.go
mtrmac: … a bit _more_ succinct, I mean.

copy/copy.go
gjenkins8: > The way I think about it, the caller is not making a request that the code here must honor

Thanks, nice. I *was* thinking about it like this. But I agree, this is more of an 'optimization hint'.

copy/copy.go
```diff
@@ -182,6 +182,11 @@ type Options struct {
 	// OciDecryptConfig contains the config that can be used to decrypt an image if it is
 	// encrypted if non-nil. If nil, it does not attempt to decrypt an image.
 	OciDecryptConfig *encconfig.DecryptConfig
+	// When true, DestinationManifestCheck will cause a copy to first check the destination image's manifest.
+	// If the destination image manifest exists, and compares equal to the source image manifest, the copy is skipped as the destination already matches the aource.
+	// Otherwise the copy proceeds as normal.
+	// This is a useful optimization to avoid if the destination image is expected to already exist, such as repeated `skopeo sync` between registries.
+	DestinationManifestCheck bool
```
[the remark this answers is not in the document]
gjenkins8: I came up with:
```
	// When OptimizeDestinationImageAlreadyExists is set, optimize the copy assuming that the destination image already
	// exists (and is equivalent). Making the eventual (no-op) copy more performant for this case. Enabling the option
	// is slightly pessimistic if the destination image doesn't exist, or is not equivalent.
	OptimizeDestinationImageAlreadyExists bool
```

copy/copy.go
gjenkins8: `"Destination copy skipped as expected image already present"` ?

copy/copy.go
```diff
@@ -633,6 +669,26 @@ func (c *copier) copyOneImage(ctx context.Context, policyContext *signature.Poli
 	// If encrypted and decryption keys provided, we should try to decrypt
 	ic.diffIDsAreNeeded = ic.diffIDsAreNeeded || (isEncrypted(src) && ic.c.ociDecryptConfig != nil) || ic.c.ociEncryptConfig != nil
 
+	options.OptimizeDestinationImageAlreadyExists = true // TODO: PR: Remove forced enablement for PR testing
```
gjenkins8, line 672: NOTE: Remove before merging

copy/copy.go
```diff
@@ -633,6 +669,26 @@ func (c *copier) copyOneImage(ctx context.Context, policyContext *signature.Poli
...
+			return nil, "", "", err
+		}
+
+		if isSrcDestManifestEqual {
+			c.Printf("Destination copy skipped as expected image already present\n")
+			return retManifest, retManifestType, retManifestDigest, nil
+		}
+	}
+
 	if err := ic.copyLayers(ctx); err != nil {
```
gjenkins8, line 692: Is it a concern that `ic.copyLayers()` updates `ic. manifestUpdates`, and so can change the result of `ic.noPendingManifestUpdates()`?  But the check to skip the copy on [L678](https://github.com/containers/image/pull/1041/files#diff-5f78baafa6d34438aba66ed64f4b16d5R678) will never see those updates?

I assume it isn't, since if no layers need to be copied because the manifest is equivalent, then `ic.copyLayers()` would not update `ic.manifestUpdates`.

But I also can't say I am a fan of this implicit contract/coupling between these bits of functionality.

mtrmac:
@gjenkins8 Thanks for the update. I’m on PTO this week, I’ll return to this afterwards.

ChristianCiach:
I've built Skopeo from source to include this PR, because we really need an optimization like this.

Unfortunately, when using your patch, `Destination copy skipped as expected image already present` never gets logged. After adding some debug logging I can see that `ic.noPendingManifestUpdates()` is always true. I think this is already discussed at https://github.com/containers/image/pull/1041/files#r501891650 .

After removing `!ic.noPendingManifestUpdates()` from the `if`-condition, everything works as expected and the sync is hugely accelerated.

rhatdan:
@gjenkins8 Could you rebase and try again?

ChristianCiach:
@gjenkins8 Thanks for the heads up! I will try to build it again some time next week.

copy/copy.go
mtrmac: Probably `s/Destination copy/Copy/`, and maybe even `s| as|,|`; I’m not a native speaker and I’m not sure how much the use of “as” to mean “because” is common.

copy/copy.go
mtrmac: Yes, it is “a concern”.

It would of course be nicer if we could rely on the `ic.manifestUpdates` coming out of `ic.copyLayers` — but avoiding `ic.copyLayers` is the whole point, we can’t make some of them without initiating the layer reads via `GetBlob`.

Still, it would probably be worth it to make this a bit harder to miss, maybe by adding a
```go
// WARNING: If you are adding new reasons to change ic.manifestUpdates, update also the OptimizeDestinationImageAlreadyExists short-circuit conditions
```
before the update at the end of `ic.copyLayers`, and a similar
```go
// WARNING: If you are adding new reasons to change the blob, …
```
somewhere in `copyBlobFromStream`, maybe either immediately after the “The copying happens through a pipeline of connected io.Readers.” comment or close to (above?) the “Deal with layer compression/decompression if necessary” one.

copy/copy.go
```diff
@@ -361,6 +365,44 @@ func supportsMultipleImages(dest types.ImageDestination) bool {
 	return false
 }
 
+// compareImageDestinationManifestEqual compares the `src` and `dest` image's manifests, returning the 'true' if they compare equal.
```
mtrmac, line 368: `checkIfDestinationManifestIsEqual` ?

“returning true and information about the manifest if…”

copy/copy.go
```diff
@@ -361,6 +365,44 @@ func supportsMultipleImages(dest types.ImageDestination) bool {
...
+// The destination manifest is always fetched.
```
mtrmac, line 369: This is not the case on error paths. “This usually fetches the destination manifest”? Or maybe just remove this comment.

copy/copy.go
```diff
@@ -361,6 +365,44 @@ func supportsMultipleImages(dest types.ImageDestination) bool {
...
+func compareImageDestinationManifestEqual(ctx context.Context, options *Options, src types.Image, dest types.ImageDestination) (bool, []byte, string, digest.Digest, error) {
+
```
mtrmac, line 371: (Absolutely non-blocking: Maybe drop this line.)

copy/copy.go
```diff
@@ -361,6 +365,44 @@ func supportsMultipleImages(dest types.ImageDestination) bool {
...
+	if err != nil {
+		return false, nil, "", "", errors.Wrapf(err, "Error calculating manifest digest")
+	}
+
+	destImageSource, err := dest.Reference().NewImageSource(ctx, options.DestinationCtx)
+	if err != nil {
+		return false, nil, "", "", nil
+	}
+
+	destManifest, destManifestType, err := destImageSource.GetManifest(ctx, nil)
```
mtrmac, line 387: Shouldn’t this use `targetInstance`? https://github.com/containers/image/pull/1041#discussion_r499155705

copy/copy.go
```diff
@@ -361,6 +365,44 @@ func supportsMultipleImages(dest types.ImageDestination) bool {
...
+	if err != nil {
+		return false, nil, "", "", nil
+	}
+
+	destManifestDigest, err := manifest.Digest(destManifest)
+	if err != nil {
+		return false, nil, "", "", errors.Wrapf(err, "Error calculating manifest digest")
+	}
+
+	logrus.Debugf("Comparing source and destintion manifest digests: srcManifestDigest=%v, destManifestDigest=%v", srcManifestDigest, destManifestDigest)
```
mtrmac, line 397:
```suggestion
	logrus.Debugf("Comparing source and destination manifest digests: srcManifestDigest=%v, destManifestDigest=%v", srcManifestDigest, destManifestDigest)
```
or maybe just something like
```suggestion
	logrus.Debugf("Source and destination manifest digests: %v vs. %v", srcManifestDigest, destManifestDigest)
```
to be less verbose in a many-image `skopeo sync` log.

copy/copy.go
```diff
@@ -361,6 +365,44 @@ func supportsMultipleImages(dest types.ImageDestination) bool {
...
+	if srcManifestDigest != destManifestDigest {
+		return false, nil, "", "", nil
+	}
+
+	// Destination and source manifests, types and digests should all be equivalent
+	return true, destManifest, destManifestType, srcManifestDigest, nil
```
mtrmac, line 403: (Non-blocking: Using `destManifestDigest` here is exactly equivalent, but maybe a tiny bit easier to follow.)

copy/copy.go
```diff
@@ -646,6 +688,27 @@ func (c *copier) copyOneImage(ctx context.Context, policyContext *signature.Poli
 	// If encrypted and decryption keys provided, we should try to decrypt
 	ic.diffIDsAreNeeded = ic.diffIDsAreNeeded || (isEncrypted(src) && ic.c.ociDecryptConfig != nil) || ic.c.ociEncryptConfig != nil
 
+	options.OptimizeDestinationImageAlreadyExists = true // TODO: PR: Remove forced enablement for PR testing
+	// If enabled, fetch and compare the destination's manifest. And as an optimization skip updating the destination iff equal
```
mtrmac, line 692: This does not make much sense just above the `isSigned` computation.

It might make sense to document the purpose of the whole section of the code, OTOH maybe just the `if options.Optimize…` (if it becomes top-level, as suggested below) with no comment is enough for that.

copy/copy.go
```diff
@@ -646,6 +688,27 @@ func (c *copier) copyOneImage(ctx context.Context, policyContext *signature.Poli
...
+	// To be enabled, options.OptimizeDestinationImageAlreadyExists must be true, the image must not be signed or encryted. And must
+	// not have any pending manifest updates
```
mtrmac, line 694: (I’m not sure that this comment helps much — it just repeats the code, doesn’t explain anything.)

copy/copy.go
```diff
@@ -646,6 +688,27 @@ func (c *copier) copyOneImage(ctx context.Context, policyContext *signature.Poli
...
+	isSigned := len(sigs) > 0 || options.SignBy != "" // TODO: Consider allowing signatures updates only and skipping the image's layers/manifest copy if possible
+
+	logrus.Debugf("Evaluating image manifest check optimzation: options.OptimizeDestinationImageAlreadyExists=%t !isSigned=%t !destRequiresOciEncryption=%t ic.noPendingManifestUpdates()=%t", options.OptimizeDestinationImageAlreadyExists, !isSigned, !destRequiresOciEncryption, ic.noPendingManifestUpdates())
```
mtrmac, line 697: Please wrap all of this in `if options.OptimizeDestinationImageAlreadyExists`, so that this log line does not appear if the caller does not opt in (and the presence of that log line is enough to indicate the value of that flag, so it doesn’t need to be logged separately); that way we don’t need to compute `isSigned` if the caller does not opt in.

mtrmac, line 697: That debug line could be polished a bit — e.g. probably log `isSigned` instead of `!isSigned`, and the like.

copy/copy.go
```diff
@@ -646,6 +688,27 @@ func (c *copier) copyOneImage(ctx context.Context, policyContext *signature.Poli
 	// If encrypted and decryption keys provided, we should try to decrypt
 	ic.diffIDsAreNeeded = ic.diffIDsAreNeeded || (isEncrypted(src) && ic.c.ociDecryptConfig != nil) || ic.c.ociEncryptConfig != nil
 
+	options.OptimizeDestinationImageAlreadyExists = true // TODO: PR: Remove forced enablement for PR testing
+	// If enabled, fetch and compare the destination's manifest. And as an optimization skip updating the destination iff equal
+	// To be enabled, options.OptimizeDestinationImageAlreadyExists must be true, the image must not be signed or encryted. And must
+	// not have any pending manifest updates
+	isSigned := len(sigs) > 0 || options.SignBy != "" // TODO: Consider allowing signatures updates only and skipping the image's layers/manifest copy if possible
```
mtrmac, line 695: This is not quite `isSigned`, rather  `shouldUpdateSigs` or something like that.

mtrmac (commented):
I’m sorry about the terribly late review.

The `targetInstance` point is the only major outstanding implementation concern; the rest is mostly stylistic.

copy/copy.go
gjenkins8: oops, thanks! I missed that comment.

I also want to discuss test coverage, but I'll do this in the main PR comment section (so discussion doesn't get lost when this comment becomes outdated )

copy/copy.go
gjenkins8: I've added these, but perhaps I should remove?

They will log if the destination image doesn't exist (albeit at debug level of course), which is a fairly common/expected scenario.

copy/copy.go
```diff
@@ -646,6 +689,26 @@ func (c *copier) copyOneImage(ctx context.Context, policyContext *signature.Poli
 	// If encrypted and decryption keys provided, we should try to decrypt
 	ic.diffIDsAreNeeded = ic.diffIDsAreNeeded || (isEncrypted(src) && ic.c.ociDecryptConfig != nil) || ic.c.ociEncryptConfig != nil
 
+	options.OptimizeDestinationImageAlreadyExists = true // TODO: PR: Remove forced enablement for PR testing
+	// If enabled, fetch and compare the destination's manifest. And as an optimization skip updating the destination iff equal
+	if options.OptimizeDestinationImageAlreadyExists {
+		shouldUpdateSigs := len(sigs) > 0 || options.SignBy != "" // TODO: Consider allowing signatures updates only and skipping the image's layers/manifest copy if possible
+
+		logrus.Debugf("Evaluating image manifest check optimzation: shouldUpdateSigs=%t destRequiresOciEncryption=%t ic.noPendingManifestUpdates()=%t", shouldUpdateSigs, destRequiresOciEncryption, ic.noPendingManifestUpdates())
```
mtrmac, line 697:
```suggestion
		logrus.Debugf("Evaluating image manifest check optimization: shouldUpdateSigs=%t destRequiresOciEncryption=%t ic.noPendingManifestUpdates()=%t", shouldUpdateSigs, destRequiresOciEncryption, ic.noPendingManifestUpdates())
```
at least, or
```suggestion
		logrus.Debugf("Checking if we can avoid the copy: shouldUpdateSigs=%t destRequiresOciEncryption=%t ic.noPendingManifestUpdates()=%t", shouldUpdateSigs, destRequiresOciEncryption, ic.noPendingManifestUpdates())
```
, or even
```suggestion
		logrus.Debugf("Checking if we can avoid the copy: has signatures=%t, OCI encryption=%t, no manifest updates=%t", shouldUpdateSigs, destRequiresOciEncryption, ic.noPendingManifestUpdates())
```
I don’t have a strong preference.

copy/copy.go
```diff
@@ -646,6 +689,26 @@ func (c *copier) copyOneImage(ctx context.Context, policyContext *signature.Poli
...
+		if !shouldUpdateSigs && !destRequiresOciEncryption && ic.noPendingManifestUpdates() {
```
mtrmac, line 698: `ic.noPendingManifestUpdates` is a bit expensive, it would be nice (and safer) to only compute it once.

mtrmac (commented):
@vrothberg Can I hand this off to you?

vrothberg:
> @vrothberg Can I hand this off to you?

Sure. I added it to my list, thanks!

copy/copy.go
gjenkins8: Good spot. Apologies, not enough thinking on my behalf. I've factored out.

copy/copy.go
gjenkins8: sure, I did the latter:
`		logrus.Debugf("Checking if we can avoid the copy: has signatures=%t, OCI encryption=%t, no manifest updates=%t", shouldUpdateSigs, destRequiresOciEncryption, noPendingManifestUpdates)`

mtrmac (commented):
LGTM (apart from https://github.com/containers/image/pull/1041/files#r500609569 ). Thanks again!

copy/copy.go
gjenkins8: Above added

copy/copy.go
```diff
@@ -361,6 +365,45 @@ func supportsMultipleImages(dest types.ImageDestination) bool {
 	return false
 }
 
+// compareImageDestinationManifestEqual compares the `src` and `dest` image manifests (reading the manifest from the
+// (possibly remote) destination). Returning true and the image's manifest, type and digest if they compare equal.
```
vrothberg, line 369:
```suggestion
// (possibly remote) destination). Returning true and the destination's manifest, type and digest if they compare equal.
```

copy/copy.go
```diff
@@ -361,6 +365,45 @@ func supportsMultipleImages(dest types.ImageDestination) bool {
...
+		logrus.Debugf("Unable to get destination image %s/%s manifest: %v", destImageSource, targetInstance, err)
+		return false, nil, "", "", nil
+	}
+
+	destManifestDigest, err := manifest.Digest(destManifest)
+	if err != nil {
+		return false, nil, "", "", errors.Wrapf(err, "Error calculating manifest digest")
+	}
+
+	logrus.Debugf("Source and destination manifest digests: %v vs. %v", srcManifestDigest, destManifestDigest)
```
vrothberg, line 398:
```suggestion
	logrus.Debugf("Comparing source and destination manifest digests: %v vs. %v", srcManifestDigest, destManifestDigest)
```
Would be a bit easier for me when skimming logs.

copy/copy.go
```diff
@@ -646,6 +689,26 @@ func (c *copier) copyOneImage(ctx context.Context, policyContext *signature.Poli
...
+		logrus.Debugf("Checking if we can avoid the copy: has signatures=%t, OCI encryption=%t, no manifest updates=%t", shouldUpdateSigs, destRequiresOciEncryption, noPendingManifestUpdates)
+		if !shouldUpdateSigs && !destRequiresOciEncryption && noPendingManifestUpdates {
+			isSrcDestManifestEqual, retManifest, retManifestType, retManifestDigest, err := compareImageDestinationManifestEqual(ctx, options, src, targetInstance, c.dest)
+			if err != nil {
+				logrus.Warnf("Failed to compare destination image manifest: %v", err)
+				return nil, "", "", err
+			}
+
+			if isSrcDestManifestEqual {
+				c.Printf("Copy skipped, expected image already present\n")
```
vrothberg, line 706:
```suggestion
				c.Printf("Skipping: image already present at destination\n")
```
"Skipping" for consistency with other logs on writer.

copy/copy.go
```diff
@@ -646,6 +689,26 @@ func (c *copier) copyOneImage(ctx context.Context, policyContext *signature.Poli
 	// If encrypted and decryption keys provided, we should try to decrypt
 	ic.diffIDsAreNeeded = ic.diffIDsAreNeeded || (isEncrypted(src) && ic.c.ociDecryptConfig != nil) || ic.c.ociEncryptConfig != nil
 
+	// If enabled, fetch and compare the destination's manifest. And as an optimization skip updating the destination iff equal
+	if options.OptimizeDestinationImageAlreadyExists {
+		shouldUpdateSigs := len(sigs) > 0 || options.SignBy != "" // TODO: Consider allowing signatures updates only and skipping the image's layers/manifest copy if possible
+		noPendingManifestUpdates := ic.noPendingManifestUpdates()
+
+		logrus.Debugf("Checking if we can avoid the copy: has signatures=%t, OCI encryption=%t, no manifest updates=%t", shouldUpdateSigs, destRequiresOciEncryption, noPendingManifestUpdates)
```
vrothberg, line 697:
```suggestion
		logrus.Debugf("Checking if we can skip copying: has signatures=%t, OCI encryption=%t, no manifest updates=%t", shouldUpdateSigs, destRequiresOciEncryption, noPendingManifestUpdates)
```

vrothberg (commented):
Thanks for working on this, @gjenkins8!

I only have minor nits and the request to squash the commits into a single one.

copy/copy.go
```diff
@@ -646,6 +689,26 @@ func (c *copier) copyOneImage(ctx context.Context, policyContext *signature.Poli
...
+		if !shouldUpdateSigs && !destRequiresOciEncryption && noPendingManifestUpdates {
```
vrothberg, line 698: Can you add a `// TODO: Consider compression changes` here?

Just in case we want further optimizations in the future.

mtrmac: IIRC (a drive-by comment without reading much else), requested compression changes are currently ignored anyway if `TryReusingBlob` finds a match — and if we find the full manifest, that should always be the case. So requested compression changes should not be a reason to skip this optimization.

gjenkins8: Thanks for the explanation @mtrmac

To paraphrase/If I understand correctly: further work would be needed for change in compression to be effected in a copy? Only then, would it make sense to disable the optimization (or more specifically, then it would be required to disable the optimization).

I've added the `// TODO: ..` as per @vrothberg's suggestion. But I don't see anything else required here(?)

mtrmac: No, I’m saying that if the destination manifest exists, any requested re-compression via `types.SystemContext.Compression{Operation,Algorithm}` wouldn’t happen _anyway_, so there is nothing to be done here.

OTOH I’m not 100% sure what “consider compression changes” refers to, @vrothberg can you clarify it into something more precise / actionable?

vrothberg: I am cool to drop the comment if it's causing more confusion than good.  My intention was to leave some breadcrumbs in case `skopeo sync` would support compression changes.  As far as I understand the code, compression changes would be ignored and @mtrmac's comment above confirm.

vrothberg:
> @vrothberg -- thanks for the review. I've updated with your suggested changes. And squashed the commits.
> 
> (and note: [CI is failing](https://travis-ci.org/github/containers/image/builds/758654465?utm_source=github_status&utm_medium=notification) due to #1107 -- I'll re-run shortly)

Thanks for your work!  When repushing, could you also rebase?

copy/copy.go
gjenkins8: I've remove the comment. Thanks

vrothberg (commented):
LGTM, thanks!

PR #1045, opened 2020-09-11
Fix problems found by codespell

rhatdan:

### DIFF
--- a/docker/reference/reference_test.go
+++ b/docker/reference/reference_test.go
@@ -639,7 +639,7 @@ func TestParseNamed(t *testing.T) {
 			failf("error parsing name: %s", err)
 			continue
 		} else if err == nil && testcase.err != nil {
-			failf("parsing succeded: expected error %v", testcase.err)
+			failf("parsing succeeded: expected error %v", testcase.err)
 			continue
 		} else if err != testcase.err {
 			failf("unexpected error %v, expected %v", err, testcase.err)
--- a/pkg/sysregistriesv2/system_registries_v2.go
+++ b/pkg/sysregistriesv2/system_registries_v2.go
@@ -363,7 +363,7 @@ type configWrapper struct {
 	configPath string
 	// path to system-wide registries.conf.d directory, or "" if not used
 	configDirPath string
-	// path to user specificed registries.conf.d directory, or "" if not used
+	// path to user specified registries.conf.d directory, or "" if not used
 	userConfigDirPath string
 }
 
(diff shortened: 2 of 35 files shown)


TomSweeneyRedHat:
LGTM

docker/reference/reference.go
```diff
@@ -8,8 +8,8 @@
 //	domain                          := domain-component ['.' domain-component]* [':' port-number]
 //	domain-component                := /([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9])/
 //	port-number                     := /[0-9]+/
-//	path-component                  := alpha-numeric [separator alpha-numeric]*
-// 	alpha-numeric                   := /[a-z0-9]+/
+//	path-component                  := alphanumeric [separator alphanumeric]*
+// 	alphanumeric                   := /[a-z0-9]+/
```
mtrmac, line 12: This comes from a different repo originally, and it would be easier to keep in sync if kept unmodified.

docker/reference/reference_test.go
```diff
@@ -639,7 +639,7 @@ func TestParseNamed(t *testing.T) {
 			failf("error parsing name: %s", err)
 			continue
 		} else if err == nil && testcase.err != nil {
-			failf("parsing succeded: expected error %v", testcase.err)
+			failf("parsing succeeded: expected error %v", testcase.err)
```
mtrmac, line 642: (This comes from the same repo, but was fixed the same way upstream, so please don’t revert this part.)

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -363,7 +363,7 @@ type configWrapper struct {
 	configPath string
 	// path to system-wide registries.conf.d directory, or "" if not used
 	configDirPath string
-	// path to user specificed registries.conf.d directory, or "" if not used
+	// path to user specified registries.conf.d directory, or "" if not used
```
mtrmac, line 366: (Shouldn’t this actually be “user-specified”? Either way it’s an improvement, and not adding extra changes that the tool did not suggest would make the PR better-contained.)

docker/reference/reference.go
rhatdan: Reverted.

pkg/sysregistriesv2/system_registries_v2.go
rhatdan: I think we should leave it the as `user specified`

mtrmac (approved):
LGTM. Thanks!

PR #1048, opened 2020-09-17
Search credentials under XDG_CONFIG_HOME

QiWang19:
Add XDG_CONFIG_HOME to the paths to be searched when login a registry. If XDG_CONFIG_HOME is empty, search under $HOME/.config. In the order that first search for authfile, XDG_RUNTIME_DIR, XDG_CONFIG_HOME, and docker config file.

Close: https://github.com/containers/podman/issues/7563

### DIFF
--- a/docs/containers-auth.json.5.md
+++ b/docs/containers-auth.json.5.md
@@ -6,8 +6,15 @@ containers-auth.json - syntax for the registry authentication file
 # DESCRIPTION
 
 A credentials file in JSON format used to authenticate against container image registries.
-On Linux it is stored at `${XDG_RUNTIME_DIR}/containers/auth.json`;
-on Windows and macOS, at `$HOME/.config/containers/auth.json`
+The primary (read/write) file is stored at `${XDG_RUNTIME_DIR}/containers/auth.json` on Linux;
+on Windows and macOS, at `$HOME/.config/containers/auth.json`.
+
+When searching for the credential for a registry, the following files will be read in sequence until the valid credential is found:
+first reading the primary (read/write) file, or the explicit override using an option of the calling application.
+If credentials are not present, search in `${XDG\_CONFIG\_HOME}/containers/auth.json`, `$HOME/.docker/config.json`, `$HOME/.dockercfg`.
+
+Except the primary (read/write) file, other files are read-only, unless the user use an option of the calling application explicitly points at it as an override.
+
 
 ## FORMAT
 
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -35,6 +35,7 @@ type authPath struct {
 
 var (
 	defaultPerUIDPathFormat = filepath.FromSlash("/run/containers/%d/auth.json")
+	xdgConfigHomePath       = filepath.FromSlash("containers/auth.json")
 	xdgRuntimeDirPath       = filepath.FromSlash("containers/auth.json")
 	dockerHomePath          = filepath.FromSlash(".docker/config.json")
 	dockerLegacyHomePath    = ".dockercfg"
@@ -136,6 +137,11 @@ func getAuthFilePaths(sys *types.SystemContext) []authPath {
 		// Logging the error as a warning instead and moving on to pulling the image
 		logrus.Warnf("%v: Trying to pull image in the event that it is a public image.", err)
 	}
+	xdgCfgHome := os.Getenv("XDG_CONFIG_HOME")
+	if xdgCfgHome == "" {
+		xdgCfgHome = filepath.Join(homedir.Get(), ".config")
+	}
+	paths = append(paths, authPath{path: filepath.Join(xdgCfgHome, xdgConfigHomePath), legacyFormat: false})
 	if dockerConfig := os.Getenv("DOCKER_CONFIG"); dockerConfig != "" {
 		paths = append(paths,
 			authPath{path: filepath.Join(dockerConfig, "config.json"), legacyFormat: false},


pkg/docker/config/config.go
```diff
@@ -136,7 +136,15 @@ func getAuthFilePaths(sys *types.SystemContext) []authPath {
 		// Logging the error as a warning instead and moving on to pulling the image
 		logrus.Warnf("%v: Trying to pull image in the event that it is a public image.", err)
 	}
+
+	var xdgCfgHomePath string
+	if xdgCfgHome := os.Getenv("XDG_CONFIG_HOME"); xdgCfgHome != "" {
+		xdgCfgHomePath = filepath.Join(xdgCfgHome, xdgRuntimeDirPath)
+	} else {
+		xdgCfgHomePath = filepath.Join(homedir.Get(), nonLinuxAuthFilePath)
+	}
```
TomSweeneyRedHat, line 145: could a non-root user set XDG_CONFIG_HOME to somewhere they shouldn't?  Should we use securjoin for thie first one?

pkg/docker/config/config.go
```diff
@@ -136,7 +136,15 @@ func getAuthFilePaths(sys *types.SystemContext) []authPath {
 		// Logging the error as a warning instead and moving on to pulling the image
 		logrus.Warnf("%v: Trying to pull image in the event that it is a public image.", err)
 	}
+
+	var xdgCfgHomePath string
+	if xdgCfgHome := os.Getenv("XDG_CONFIG_HOME"); xdgCfgHome != "" {
+		xdgCfgHomePath = filepath.Join(xdgCfgHome, xdgRuntimeDirPath)
```
mtrmac, line 142: Please introduce a new `xdcConfigHomePath` constant, or maybe somehow rename the old one; don’t just silently reuse it for a different purpose.

pkg/docker/config/config.go
```diff
@@ -136,7 +136,15 @@ func getAuthFilePaths(sys *types.SystemContext) []authPath {
...
+	} else {
+		xdgCfgHomePath = filepath.Join(homedir.Get(), nonLinuxAuthFilePath)
```
mtrmac, line 144: Using a “`nonLinux`” path on Linux is also confusing — and on non-Linux that path was already added.

(In this context, it’s not a “non-linux” code path; it’s a “following the XDG base dir spec, XDG_CONFIG_HOME is not set so use the spec-defined default” path; i.e. the if/else should conceptually only get the $HOME`/.config`)

mtrmac (commented):
Just a very quick skim, mostly quibbling with code clarity, not with the outcome.

Adding extra hard-coded logic when we wanted 1) not to store plain-text credentials and 2) use credential helpers, feels a bit counterproductive… yet it’s clearly useful, and as long as we don’t write to that path by default (i.e. requires an explicit opt-in), seems safe enough.

Please update `docs/containers-auth.json.5.md` as well.

pkg/docker/config/config.go
mtrmac: Looking at the other comment — this should probably end up as a “get XDG_CONFIG_HOME path” + “append a path relative to XDG_CONFIG_HOME”.

pkg/docker/config/config.go
mtrmac: The same user can already specify `--authfile` pointing anywhere else (unless the user somehow able to manipulate the environment of a future `su`/`sudo`, but not the command being run… I’m not too worried about that, and setting `XDG_RUNTIME_DIR` would allow the same attack already).

The suffix is a known constant, we don’t need `securejoin` to protect against that constant containing `../`.

QiWang19:
> 
> Please update `docs/containers-auth.json.5.md` as well.

The PR just use XDG_CONFIG_HOME, like .docker/config, to search for existing name&password when no valid credentials under XDG_RUNTIME_DIR. `docs/containers-auth.json.5.md` doesn't mention these search order.  This change can be mentioned in the login man pages when this PR applied to Podman, Buildah?

mtrmac:
I think we typically do document search order and the like in the file format man pages. (Otherwise it would have to be documented not only in the `login` page, but also in `pull` and `push` and so on, in every single consumer of c/image, with no way to tell which man page is authoritative if they differed.)

`.docker/config.json` is a bit different in that this project is not authoritative about the format specification — but mentioning it in the `containers-auth.json(5)` man page could also be useful.

OTOH with the long-planned move to prioritize credential helpers, we admittedly might need a different place for the primary documentation of the mechanism. I’m not sure what that place would be; `registries.conf` would not be easy enough to find.

QiWang19:
@mtrmac PTAL

rhatdan:
@QiWang19 Needs a rebase.

QiWang19:
@mtrmac PTAL

docs/containers-auth.json.5.md
```diff
@@ -7,7 +7,10 @@ containers-auth.json - syntax for the registry authentication file
 
 A credentials file in JSON format used to authenticate against container image registries.
 On Linux it is stored at `${XDG_RUNTIME_DIR}/containers/auth.json`;
-on Windows and macOS, at `$HOME/.config/containers/auth.json`
+on Windows and macOS, at `$HOME/.config/containers/auth.json`.
+
+When searching for the credential for a registry, the following files will be read in sequence until the valid credential is found: the system-wide file will be first searched if configured, if credentials are not present, search in `${XDG\_RUNTIME\_DIR}/containers/auth.json`, `${XDG\_CONFIG\_HOME}/containers/auth.json`, `$HOME/.docker/config.json`.
```
mtrmac, line 12: - It took me multiple tries to realize “the system-wide file … if configured” refers the `SystemContext.AuthFilePath` and the like; this should probably refer to an explicit override using an option of the calling application, or something like that (end users don’t interact with `SystemContext` directly).
- If the above _is_ configured, `XDG_RUNTIME_DIR/…` is not attempted.
- This ultimately should clearly differentiate between the one file that the code reads _and_ writes, and the others than are read-only (unless the user explicitly points at it as an override).

---

Using `os.UserConfigDir` is very interesting, but the results are not the same as `XDG_RUNTIME_DIR` on non-Linux platforms. At this point I’m not sure we have the expertise or the time/interest to commit to adding another Windows/macOS … directory (and we can do so later if the expertise/interest comes up); so maybe only add the new behavior “on traditional Unix” (i.e. if we are not using the `nonLinuxAuthFilePath` value).

docs/containers-auth.json.5.md
```diff
@@ -7,7 +7,13 @@ containers-auth.json - syntax for the registry authentication file
 
 A credentials file in JSON format used to authenticate against container image registries.
 On Linux it is stored at `${XDG_RUNTIME_DIR}/containers/auth.json`;
```
mtrmac, line 9: Maybe
```suggestion
The primary (read/write) file is stored at `${XDG_RUNTIME_DIR}/containers/auth.json` on Linux;
```
and then refer to “primary” in the last paragraph, instead of saying `XDG_RUNTIME_DIR` on Linux or … on other OS, again?

WRT reverting to explicitly use XDG_CONFIG_HOME… fine, it is equivalent to `nonLinuxAuthFilePath` so it does not introduce new behavior on Windows/macOS; works for me.

mtrmac (approved):
LGTM. Thanks!

PR #1050, opened 2020-09-22
Skip checking if the manifest difest matches for docker-daemon source. Fixes #1049

### LINKED ISSUE #1049: Error "Manifest does not match provided manifest digest" when using @sha256:digest in docker-daemon
I reported this in skopeo (see https://github.com/containers/skopeo/issues/1051) but the fix might be required in containers/image instead.

When running a `skopeo inspect docker-daemon:repo@sha256:digest` it fails with a "Manifest does not match provided manifest digest" error.

After some tests, the cause might be that the contents of the manifest are different (testing with `skopeo inspect --raw` for _docker://_ source and for _docker-daemon:_ source. My guess in here is that for _docker-daemon:_ source the manifest is automatically generated by Skopeo or the underlying libraries, while for _docker://_ source, the manifest is downloaded "as is". The _docker://_ version, when retrieved from the registry, is indented and formatted, so the hash differs with the hash of the locally generated manifest. See my tests:

```
airadier@ender:~$ docker pull alpine@sha256:a15790640a6690aa1730c38cf0a440e2aa44aaca9b0e8931a9f2b0d7cc90fd65
sha256:a15790640a6690aa1730c38cf0a440e2aa44aaca9b0e8931a9f2b0d7cc90fd65: Pulling from library/alpine
df20fa9351a1: Pull complete
Digest: sha256:a15790640a6690aa1730c38cf0a440e2aa44aaca9b0e8931a9f2b0d7cc90fd65
Status: Downloaded newer image for alpine@sha256:a15790640a6690aa1730c38cf0a440e2aa44aaca9b0e8931a9f2b0d7cc90fd65

airadier@ender:~$ ./skopeo inspect docker-daemon:alpine@sha256:a15790640a6690aa1730c38cf0a440e2aa44aaca9b0e8931a9f2b0d7cc90fd65
FATA[0000] Error parsing manifest for image: Manifest does not match provided manifest digest sha256:a15790640a6690aa1730c38cf0a440e2aa44aaca9b0e8931a9f2b0d7cc90fd65

airadier@ender:~$ ./skopeo inspect --raw docker-daemon:alpine@sha256:a15790640a6690aa1730c38cf0a440e2aa44aaca9b0e8931a9f2b0d7cc90fd65
{"schemaVersion":2,"mediaType":"application/vnd.docker.distribution.manifest.v2+json","config":{"mediaType":"application/vnd.docker.container.image.v1+json","size":1509,"digest":"sha256:a24bb4013296f61e89ba57005a7b3e52274d8edd3ae2077d04395f806b63d83e"},"layers":[{"mediaType":"application/vnd.docker.image.rootfs.diff.tar.gzip","size":5844992,"digest":"sha256:50644c29ef5a27c9a40c393a73ece2479de78325cae7d762ef3cdc19bf42dd0a"}]}

airadier@ender:~$ ./skopeo inspect --raw docker-daemon:alpine@sha256:a15790640a6690aa1730c38cf0a440e2aa44aaca9b0e8931a9f2b0d7cc90fd65 | shasum -a 256
0fe895ba713b184489c17550d67464dd67fade3f92f3f0c1904dc8df6750d609  -

airadier@ender:~$ ./skopeo inspect --raw docker://alpine@sha256:a15790640a6690aa1730c38cf0a440e2aa44aaca9b0e8931a9f2b0d7cc90fd65
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
   "config": {
      "mediaType": "application/vnd.docker.container.image.v1+json",
      "size": 1509,
      "digest": "sha256:a24bb4013296f61e89ba57005a7b3e52274d8edd3ae2077d04395f806b63d83e"
   },
   "layers": [
      {
         "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
         "size": 2797541,
         "digest": "sha256:df20fa9351a15782c64e6dddb2d4a6f50bf6d3688060a34c4014b0d9a752eb4c"
      }
   ]
}

airadier@ender:~$ ./skopeo inspect --raw docker://alpine@sha256:a15790640a6690aa1730c38cf0a440e2aa44aaca9b0e8931a9f2b0d7cc90fd65 | shasum -a 256
a15790640a6690aa1730c38cf0a440e2aa44aaca9b0e8931a9f2b0d7cc90fd65  -
```

I would expect that it is possible to do a `skopeo inspect docker-daemon:repo@sha256:digest` instead of failing. So some proposals would be:

- Fix the manifest generation, so it matches the remote manifest format. I am not sure if this is in general a good idea. Is there an standard for manifest formatting? 
- Skip the check that seems to be done in https://github.com/containers/image/blob/21244c96ad792ef415068dc1bc1ab82dffb68dc3/image/unparsed.go#L59, as the digest for the manifest generated on the fly won't probably match the real manifest in the remote repository.

The issue also applies for other commands like `skopeo copy`:

```
kopeo copy docker-daemon:alpine@sha256:a15790640a6690aa1730c38cf0a440e2aa44aaca9b0e8931a9f2b0d7cc90fd65 oci:test.tar
FATA[0000] Error determining manifest MIME type for docker-daemon:alpine@sha256:a15790640a6690aa1730c38cf0a440e2aa44aaca9b0e8931a9f2b0d7cc90fd65: Manifest does not match provided manifest digest sha256:a15790640a6690aa1730c38cf0a440e2aa44aaca9b0e8931a9f2b0d7cc90fd6
```

and anything using FromUnparsedImage (https://github.com/containers/image/blob/1a0dda734e3bed323a9674e274b379cf841fe8b1/image/sourced.go#L68) which in turn uses:

https://github.com/containers/image/blob/21244c96ad792ef415068dc1bc1ab82dffb68dc3/image/unparsed.go#L59

airadier:
As described in https://github.com/containers/container-libs/issues/261, the manifest digest specified in the repo@sha256:digest won't match the digest calculated from the locally generated manifest when using docker-daemon source.

This fix skips the verification for docker-daemon source.

### DIFF
--- a/image/unparsed.go
+++ b/image/unparsed.go
@@ -55,7 +55,7 @@ func (i *UnparsedImage) Manifest(ctx context.Context) ([]byte, string, error) {
 			if err != nil {
 				return nil, "", errors.Wrap(err, "Error computing manifest digest")
 			}
-			if !matches {
+			if i.Reference().Transport().Name() != "docker-daemon" && !matches {
 				return nil, "", errors.Errorf("Manifest does not match provided manifest digest %s", digest)
 			}
 		}


mtrmac (changes requested):
Thanks for the PR.

We shouldn’t have exceptions like this in the code, it will be very difficult to trace back unexpected behavior to this condition.

My first guess is that `.DockerReference` should just not return a value if it can’t return a working value, but I’ll need to take a closer look at how the method is used in callers of c/image (sadly it’s not very well-defined what the value is intended to be used for).

mtrmac:
Changing the `DockerReference` method of the `docker-daemon:` transport is the available way to represent the situation in the current API.

Alternatively, yes, a transport declaring that the digests from `DockerReference` are not to be validated is an option.

_Either way_ that affects semantics in caller-visible ways, so we need to at least audit existing users in the GitHub.com/containers ecosystem.

universam1:
example issue: `podman pull ghcr.io/fluxcd/flagger:1.22.1`

Any workaround possible?

mtrmac:
@universam1 That seems completely unrelated to the `docker-daemon` problem in the original PR. Please file a separate report, against Podman, and include actual/expected results and logs.

jankaluza:

PR #1056, opened 2020-09-29
use DOCKER_CONFIG if set

### LINKED ISSUE #1055: consult DOCKER_CONFIG environment variable
The docker CLI lets you set the `DOCKER_CONFIG` env var to point to the directory containing config.json. https://docs.docker.com/engine/reference/commandline/cli/#environment-variables However, right now this library just looks at `$HOME/.docker/config.json`. It would be nice if it consulted `$DOCKER_CONFIG/config.json` instead if the env var is set.

rittneje:
Use $DOCKER_CONFIG/config.json as an auth path to match the behavior of the docker CLI.
Fixes #1055

Signed-off-by: Jesse Rittner <rittneje@gmail.com>

### DIFF
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -136,8 +136,16 @@ func getAuthFilePaths(sys *types.SystemContext) []authPath {
 		// Logging the error as a warning instead and moving on to pulling the image
 		logrus.Warnf("%v: Trying to pull image in the event that it is a public image.", err)
 	}
+	if dockerConfig := os.Getenv("DOCKER_CONFIG"); dockerConfig != "" {
+		paths = append(paths,
+			authPath{path: filepath.Join(dockerConfig, "config.json"), legacyFormat: false},
+		)
+	} else {
+		paths = append(paths,
+			authPath{path: filepath.Join(homedir.Get(), dockerHomePath), legacyFormat: false},
+		)
+	}
 	paths = append(paths,
-		authPath{path: filepath.Join(homedir.Get(), dockerHomePath), legacyFormat: false},
 		authPath{path: filepath.Join(homedir.Get(), dockerLegacyHomePath), legacyFormat: true},
 	)
 	return paths


vrothberg (commented):
The changes to the certs files look unrelated.

Code changes LGTM, @mtrmac @QiWang19 WDYT?

rittneje:
@vrothberg They are unrelated. A few of the unit tests failed because the certs expired the other day, so I regenerated them.

mtrmac:
Thanks for the certificate update — it’s unrelated and urgent, so let’s split it into #1057 so that we can discuss the details of this PR here.

pkg/docker/config/config.go
```diff
@@ -136,6 +136,11 @@ func getAuthFilePaths(sys *types.SystemContext) []authPath {
 		// Logging the error as a warning instead and moving on to pulling the image
 		logrus.Warnf("%v: Trying to pull image in the event that it is a public image.", err)
 	}
+	if dockerConfig := os.Getenv("DOCKER_CONFIG"); dockerConfig != "" {
+		paths = append(paths,
+			authPath{path: filepath.Join(dockerConfig, "config.json"), legacyFormat: false},
+		)
+	}
```
mtrmac, line 143: Looking at https://github.com/projectatomic/docker/blob/master/cliconfig/config.go as well as more recent versions, the `DOCKER_CONFIG` path _replaces_ the $HOME`/.docker`, it does not add one more file to check. So this code should do the same thing.

mtrmac (commented):
In principle, it’s non-transparent for a library to accept configuration from non-obvious channels like environment variables, making it hard for callers to reason about the behavior, so I’m not too happy about this.

OTOH the whole point of reading these files is compatibility, and it’s not the very first place where we add environment variables, so… I guess reading the environment is better than not reading it.

pkg/docker/config/config.go
```diff
@@ -136,10 +136,16 @@ func getAuthFilePaths(sys *types.SystemContext) []authPath {
...
+	if dockerConfig := os.Getenv("DOCKER_CONFIG"); dockerConfig != "" {
+		paths = append(paths,
+			authPath{path: filepath.Join(dockerConfig, "config.json"), legacyFormat: false},
+		)
+	} else {
+		paths = append(paths,
+			authPath{path: filepath.Join(homedir.Get(), dockerHomePath), legacyFormat: false},
+			authPath{path: filepath.Join(homedir.Get(), dockerLegacyHomePath), legacyFormat: true},
+		)
+	}
```
mtrmac, line 148: No, the legacy file always uses `$HOME` AFAICS, and is not ignored when `DOCKER_CONFIG` is set.

rittneje: I don't know why someone would go through the effort of setting `DOCKER_CONFIG` but still want the legacy behavior. In any case, done.

mtrmac (commented):
LGTM. Thanks!

pkg/docker/config/config.go
mtrmac: Sure, but we never want to end up researching why the two behave differently when we can easily prevent it before merging the PR.

rittneje:
@mtrmac Can this be merged?

QiWang19:
LGTM

PR #1060, opened 2020-10-06
sysregistriesv2: short-name aliasing

vrothberg:
The use of unqualified-search registries entails an ambiguity as it's
unclear from which registry a given image, referenced by a short name,
may be pulled from.

There have been reports of squatting on some registries to trap users
into pulling from a registry that may have pull precedence over the
intended registry; all depending on the relative order of the entries
in the `unqualified-search-registries` field in the `registries.conf`.

Removing the feature of unqualified-search registries is not an option,
as many users depend on it.  The agreed on alternative is to introduce
aliases.  Aliases make short-name resolution explicit.  Similar to bash
aliases, the new `[aliases]` table (internally a `map[string]string`)
has a left-hand name and a right-hand value:

```TOML
[aliases]
name="registry.com/namespace/name:tag"
```

Consumers of containers/image can now use the new API to resolve aliases
and add new ones.

When adding a new alias via `(*V2RegistriesConf).AddShortNameAlias(...)`,
the alias is written to a dedicated `short-name-alias.conf` file.  The
separate file attempts to separate potentially user-edited files in
`/etc` from a machine-generated file.

Note that the `short-name-alias.conf` has precedence over
registries.conf files when looking up an alias via
`(*V2RegistriesConf).ResolveShortNameAlias(...)`.

### DIFF
--- a/pkg/sysregistriesv2/shortnames.go
+++ b/pkg/sysregistriesv2/shortnames.go
@@ -1,0 +1,316 @@
+package sysregistriesv2
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/BurntSushi/toml"
+	"github.com/containers/image/v5/docker/reference"
+	"github.com/containers/image/v5/types"
+	"github.com/containers/storage/pkg/lockfile"
+	"github.com/docker/docker/pkg/homedir"
+	"github.com/pkg/errors"
+)
+
+// defaultShortNameMode is the default mode of registries.conf files if the
+// corresponding field is left empty.
+const defaultShortNameMode = types.ShortNameModePermissive
+
+// userShortNamesFile is the user-specific config file to store aliases.
+var userShortNamesFile = filepath.FromSlash("containers/short-name-aliases.conf")
+
+// shortNameAliasesConfPath returns the path to the machine-generated
+// short-name-aliases.conf file.
+func shortNameAliasesConfPath(ctx *types.SystemContext) (string, error) {
+	if ctx != nil && len(ctx.UserShortNameAliasConfPath) > 0 {
+		return ctx.UserShortNameAliasConfPath, nil
+	}
+
+	configHome, err := homedir.GetConfigHome()
+	if err != nil {
+		return "", err
+	}
+
+	return filepath.Join(configHome, userShortNamesFile), nil
+}
+
+// alias combines the parsed value of an alias with the config file it has been
+// specified in.  The config file is crucial for an improved user experience
+// such that users are able to resolve potential pull errors.
+type alias struct {
+	// The parsed value of an alias.  May be nil if set to "" in a config.
+	value reference.Named
+	// The config file the alias originates from.
+	configOrigin string
+}
+
+// shortNameAliasConf is a subset of the `V2RegistriesConf` format.  It's used in the
+// software-maintained `userShortNamesFile`.
+type shortNameAliasConf struct {
+	// A map for aliasing short names to their fully-qualified image
+	// reference counter parts.
+	// Note that Aliases is niled after being loaded from a file.
+	Aliases map[string]string `toml:"aliases"`
+
+	// Note that an alias value may be nil iff it's set as an empty string
+	// in the config.
+	namedAliases map[string]alias
+}
+
+// ResolveShortNameAlias performs an alias resolution of the specified name.
+// The user-specific short-name-aliases.conf has precedence over aliases in the
+// assembled registries.conf.  It returns the possibly resolved alias or nil, a
+// human-readable description of the config where the alias is specified, and
+// an error. The origin of the config file is crucial for an improved user
+// experience such that users are able to resolve potential pull errors.
+// Almost all callers should use pkg/shortnames instead.
+//
+// Note that it’s the caller’s responsibility to pass only a repository
+// (reference.IsNameOnly) as the short name.
+func ResolveShortNameAlias(ctx *types.SystemContext, name string) (reference.Named, string, error) {
+	if err := validateShortName(name); err != nil {
+		return nil, "", err
+	}
+	confPath, lock, err := shortNameAliasesConfPathAndLock(ctx)
+	if err != nil {
+		return nil, "", err
+	}
+
+	// Acquire the lock as a reader to allow for multiple routines in the
+	// same process space to read simultaneously.
+	lock.RLock()
+	defer lock.Unlock()
+
+	aliasConf, err := loadShortNameAliasConf(confPath)
+	if err != nil {
+		return nil, "", err
+	}
+
+	// First look up the short-name-aliases.conf.  Note that a value may be
+	// nil iff it's set as an empty string in the config.
+	alias, resolved := aliasConf.namedAliases[name]
+	if resolved {
+		return alias.value, alias.configOrigin, nil
+	}
+
+	config, err := getConfig(ctx)
+	if err != nil {
+		return nil, "", err
+	}
+	alias, resolved = config.namedAliases[name]
+	if resolved {
+		return alias.value, alias.configOrigin, nil
+	}
+	return nil, "", nil
+}
+
+// editShortNameAlias loads the aliases.conf file and changes it. If value is
+// set, it adds the name-value pair as a new alias. Otherwise, it will remove
+// name from the config.
+func editShortNameAlias(ctx *types.SystemContext, name string, value *string) error {
+	if err := validateShortName(name); err != nil {
+		return err
+	}
+	if value != nil {
+		if _, err := parseShortNameValue(*value); err != nil {
+			return err
+		}
+	}
+
+	confPath, lock, err := shortNameAliasesConfPathAndLock(ctx)
+	if err != nil {
+		return err
+	}
+
+	// Acquire the lock as a writer to prevent data corruption.
+	lock.Lock()
+	defer lock.Unlock()
+
+	// Load the short-name-alias.conf, add the specified name-value pair,
+	// and write it back to the file.
+	conf, err := loadShortNameAliasConf(confPath)
+	if err != nil {
+		return err
+	}
+
+	if value != nil {
+		conf.Aliases[name] = *value
+	} else {
+		// If the name does not exist, throw an error.
+		if _, exists := conf.Aliases[name]; !exists {
+			return errors.Errorf("short-name alias %q not found in %q: please check registries.conf files", name, confPath)
+		}
+
+		delete(conf.Aliases, name)
+	}
+
+	f, err := os.OpenFile(confPath, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0600)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	encoder := toml.NewEncoder(f)
+	return encoder.Encode(conf)
+}
+
+// AddShortNameAlias adds the specified name-value pair as a new alias to the
+// user-specific aliases.conf.  It may override an existing alias for `name`.
+//
+// Note that it’s the caller’s responsibility to pass only a repository
+// (reference.IsNameOnly) as the short name.
+func AddShortNameAlias(ctx *types.SystemContext, name string, value string) error {
+	return editShortNameAlias(ctx, name, &value)
+}
+
+// RemoveShortNameAlias clears the alias for the specified name.  It throws an
+// error in case name does not exist in the machine-generated
+// short-name-alias.conf.  In such case, the alias must be specified in one of
+// the registries.conf files, which is the users' responsibility.
+//
+// Note that it’s the caller’s responsibility to pass only a repository
+// (reference.IsNameOnly) as the short name.
+func RemoveShortNameAlias(ctx *types.SystemContext, name string) error {
+	return editShortNameAlias(ctx, name, nil)
+}
+
+// parseShortNameValue parses the specified alias into a reference.Named.  The alias is
+// expected to not be tagged or carry a digest and *must* include a
+// domain/registry.
+//
+// Note that the returned reference is always normalized.
+func parseShortNameValue(alias string) (reference.Named, error) {
+	ref, err := reference.Parse(alias)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error parsing alias %q", alias)
+	}
+
+	if _, ok := ref.(reference.Digested); ok {
+		return nil, errors.Errorf("invalid alias %q: must not contain digest", alias)
+	}
+
+	if _, ok := ref.(reference.Tagged); ok {
+		return nil, errors.Errorf("invalid alias %q: must not contain tag", alias)
+	}
+
+	named, ok := ref.(reference.Named)
+	if !ok {
+		return nil, errors.Errorf("invalid alias %q: must contain registry and repository", alias)
+	}
+
+	registry := reference.Domain(named)
+	if !(strings.ContainsAny(registry, ".:") || registry == "localhost") {
+		return nil, errors.Errorf("invalid alias %q: must contain registry and repository", alias)
+	}
+
+	// A final parse to make sure that docker.io references are correctly
+	// normalized (e.g., docker.io/alpine to docker.io/library/alpine.
+	named, err = reference.ParseNormalizedNamed(alias)
+	return named, err
+}
+
+// validateShortName parses the specified `name` of an alias (i.e., the left-hand
+// side) and checks if it's a short name and does not include a tag or digest.
+func validateShortName(name string) error {
+	repo, err := reference.Parse(name)
+	if err != nil {
+		return errors.Wrapf(err, "cannot parse short name: %q", name)
+	}
+
+	if _, ok := repo.(reference.Digested); ok {
+		return errors.Errorf("invalid short name %q: must not contain digest", name)
+	}
+
+	if _, ok := repo.(reference.Tagged); ok {
+		return errors.Errorf("invalid short name %q: must not contain tag", name)
+	}
+
+	named, ok := repo.(reference.Named)
+	if !ok {
+		return errors.Errorf("invalid short name %q: no name", name)
+	}
+
+	registry := reference.Domain(named)
+	if strings.ContainsAny(registry, ".:") || registry == "localhost" {
+		return errors.Errorf("invalid short name %q: must not contain registry", name)
+	}
+	return nil
+}
+
+// parseAndValidate parses and validates all entries in conf.Aliases and stores
+// the results in conf.namedAliases.
+func (conf *shortNameAliasConf) parseAndValidate(path string) error {
+	if conf.Aliases == nil {
+		conf.Aliases = make(map[string]string)
+	}
+	if conf.namedAliases == nil {
+		conf.namedAliases = make(map[string]alias)
+	}
+	errs := []error{}
+	for name, value := range conf.Aliases {
+		if err := validateShortName(name); err != nil {
+			errs = append(errs, err)
+		}
+
+		// Empty right-hand side values in config files allow to reset
+		// an alias in a previously loaded config. This way, drop-in
+		// config files from registries.conf.d can reset potentially
+		// malconfigured aliases.
+		if value == "" {
+			conf.namedAliases[name] = alias{nil, path}
+			continue
+		}
+
+		named, err := parseShortNameValue(value)
+		if err != nil {
+			// We want to report *all* malformed entries to avoid a
+			// whack-a-mole for the user.
+			errs = append(errs, err)
+		} else {
+			conf.namedAliases[name] = alias{named, path}
+		}
+	}
+	var err error // nil if no errors
+	for _, e := range errs {
+		if err == nil {
+			err = e
+		} else {
+			err = errors.Wrapf(err, "%v\n", e)
+		}
+	}
+	return err
+}
+
+func loadShortNameAliasConf(confPath string) (*shortNameAliasConf, error) {
+	conf := shortNameAliasConf{}
+
+	_, err := toml.DecodeFile(confPath, &conf)
+	if err != nil && !os.IsNotExist(err) {
+		// It's okay if the config doesn't exist.  Other errors are not.
+		return nil, errors.Wrapf(err, "error loading short-name aliases config file %q", confPath)
+	}
+
+	// Better safe than sorry: validate the machine-generated config.  The
+	// file could still be corrupted by another process or user.
+	if err := conf.parseAndValidate(confPath); err != nil {
+		return nil, errors.Wrapf(err, "error loading short-name aliases config file %q", confPath)
+	}
+
+	return &conf, nil
+}
+
+func shortNameAliasesConfPathAndLock(ctx *types.SystemContext) (string, lockfile.Locker, error) {
+	shortNameAliasesConfPath, err := shortNameAliasesConfPath(ctx)
+	if err != nil {
+		return "", nil, err
+	}
+	// Make sure the path to file exists.
+	if err := os.MkdirAll(filepath.Dir(shortNameAliasesConfPath), 0700); err != nil {
+		return "", nil, err
+	}
+
+	lockPath := shortNameAliasesConfPath + ".lock"
+	locker, err := lockfile.GetLockfile(lockPath)
+	return shortNameAliasesConfPath, locker, err
+}
--- a/pkg/sysregistriesv2/shortnames_test.go
+++ b/pkg/sysregistriesv2/shortnames_test.go
@@ -1,0 +1,276 @@
+package sysregistriesv2
+
+import (
+	"io/ioutil"
+	"os"
+	"testing"
+
+	"github.com/containers/image/v5/types"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestParseShortNameValue(t *testing.T) {
+	tests := []struct {
+		input string
+		valid bool
+	}{
+		// VALID INPUT
+		{"docker.io/library/fedora", true},
+		{"localhost/fedora", true},
+		{"localhost:5000/fedora", true},
+		{"localhost:5000/namespace/fedora", true},
+		// INVALID INPUT
+		{"docker.io/library/fedora:latest", false}, // tag
+		{"docker.io/library/fedora@sha256:b87dd5f837112a9e1e9882963a6406387597698268c0ad371b187151a5dfe6bf", false}, // digest
+		{"fedora", false},                // short name
+		{"fedora:latest", false},         // short name + tag
+		{"library/fedora", false},        // no registry
+		{"library/fedora:latest", false}, // no registry + tag
+		{"$$4455%%", false},              // gargabe
+		{"docker://foo", false},          // transports are not supported
+		{"docker-archive://foo", false},  // transports are not supported
+		{"", false},                      // empty
+	}
+
+	for _, test := range tests {
+		named, err := parseShortNameValue(test.input)
+		if test.valid {
+			require.NoError(t, err, "%q should be a valid alias", test.input)
+			assert.NotNil(t, named)
+			assert.Equal(t, test.input, named.String())
+		} else {
+			require.Error(t, err, "%q should be an invalid alias", test.input)
+			assert.Nil(t, named)
+		}
+	}
+
+	// Now make sure that docker.io references are normalized.
+	named, err := parseShortNameValue("docker.io/fedora")
+	require.NoError(t, err)
+	assert.NotNil(t, named)
+	assert.Equal(t, "docker.io/library/fedora", named.String())
+
+}
+
+func TestValidateShortName(t *testing.T) {
+	tests := []struct {
+		input string
+		valid bool
+	}{
+		// VALID INPUT
+		{"library/fedora", true},
+		{"fedora", true},
+		{"1234567489", true},
+		// INVALID INPUT
+		{"docker.io/library/fedora:latest", false},
+		{"docker.io/library/fedora@sha256:b87dd5f837112a9e1e9882963a6406387597698268c0ad371b187151a5dfe6bf", false}, // digest
+		{"fedora:latest", false},
+		{"library/fedora:latest", false},
+		{"$$4455%%", false},
+		{"docker://foo", false},
+		{"docker-archive://foo", false},
+		{"", false},
+	}
+
+	for _, test := range tests {
+		err := validateShortName(test.input)
+		if test.valid {
+			require.NoError(t, err, "%q should be a valid alias", test.input)
+		} else {
+			require.Error(t, err, "%q should be an invalid alias", test.input)
+		}
+	}
+}
+
+func TestResolveShortNameAlias(t *testing.T) {
+	tmp, err := ioutil.TempFile("", "aliases.conf")
+	require.NoError(t, err)
+	defer os.Remove(tmp.Name())
+
+	sys := &types.SystemContext{
+		SystemRegistriesConfPath:    "testdata/aliases.conf",
+		SystemRegistriesConfDirPath: "testdata/this-does-not-exist",
+		UserShortNameAliasConfPath:  tmp.Name(),
+	}
+
+	configCache = make(map[configWrapper]*V2RegistriesConf)
+	conf, err := TryUpdatingCache(sys)
+	require.NoError(t, err)
+	assert.Len(t, conf.namedAliases, 4)
+	assert.Len(t, conf.Aliases, 0)
+
+	aliases := []struct {
+		name, value string
+	}{
+		{
+			"docker",
+			"docker.io/library/foo",
+		},
+		{
+			"quay/foo",
+			"quay.io/library/foo",
+		},
+		{
+			"example",
+			"example.com/library/foo",
+		},
+	}
+
+	for _, alias := range aliases {
+		value, path, err := ResolveShortNameAlias(sys, alias.name)
+		require.NoError(t, err)
+		require.NotNil(t, value)
+		assert.Equal(t, alias.value, value.String())
+		assert.Equal(t, "testdata/aliases.conf", path)
+	}
+
+	// Non-existent alias.
+	value, path, err := ResolveShortNameAlias(sys, "idonotexist")
+	require.NoError(t, err)
+	assert.Nil(t, value)
+	assert.Equal(t, "", path)
+
+	// Empty right-hand value (special case) -> does not resolve.
+	value, path, err = ResolveShortNameAlias(sys, "empty")
+	require.NoError(t, err)
+	assert.Nil(t, value)
+	assert.Equal(t, "testdata/aliases.conf", path)
+}
+
+func TestAliasesWithDropInConfigs(t *testing.T) {
+	tmp, err := ioutil.TempFile("", "aliases.conf")
+	require.NoError(t, err)
+	defer os.Remove(tmp.Name())
+
+	sys := &types.SystemContext{
+		SystemRegistriesConfPath:    "testdata/aliases.conf",
+		SystemRegistriesConfDirPath: "testdata/registries.conf.d",
+		UserShortNameAliasConfPath:  tmp.Name(),
+	}
+
+	configCache = make(map[configWrapper]*V2RegistriesConf)
+	conf, err := TryUpdatingCache(sys)
+	require.NoError(t, err)
+	assert.Len(t, conf.namedAliases, 8)
+	assert.Len(t, conf.Aliases, 0)
+
+	aliases := []struct {
+		name, value, config string
+	}{
+		{
+			"docker",
+			"docker.io/library/config1",
+			"testdata/registries.conf.d/config-1.conf",
+		},
+		{
+			"quay/foo",
+			"quay.io/library/foo",
+			"testdata/aliases.conf",
+		},
+		{
+			"config1",
+			"config1.com/image", // from config1
+			"testdata/registries.conf.d/config-1.conf",
+		},
+		{
+			"barz",
+			"barz.com/config2", // from config1, overridden by config2
+			"testdata/registries.conf.d/config-2.conf",
+		},
+		{
+			"config2",
+			"config2.com/image", // from config2
+			"testdata/registries.conf.d/config-2.conf",
+		},
+		{
+			"added1",
+			"aliases.conf/added1", // from AddShortNameAlias
+			tmp.Name(),
+		},
+		{
+			"added2",
+			"aliases.conf/added2", // from AddShortNameAlias
+			tmp.Name(),
+		},
+		{
+			"added3",
+			"aliases.conf/added3", // from config2, overridden by AddShortNameAlias
+			tmp.Name(),
+		},
+	}
+
+	require.NoError(t, AddShortNameAlias(sys, "added1", "aliases.conf/added1"))
+	require.NoError(t, AddShortNameAlias(sys, "added2", "aliases.conf/added2"))
+	require.NoError(t, AddShortNameAlias(sys, "added3", "aliases.conf/added3"))
+
+	for _, alias := range aliases {
+		value, path, err := ResolveShortNameAlias(sys, alias.name)
+		require.NoError(t, err)
+		require.NotNil(t, value, "%v", alias)
+		assert.Equal(t, alias.value, value.String())
+		assert.Equal(t, alias.config, path)
+	}
+
+	value, path, err := ResolveShortNameAlias(sys, "i/do/no/exist")
+	require.NoError(t, err)
+	assert.Nil(t, value)
+	assert.Equal(t, "", path)
+
+	// Empty right-hand value (special case) -> does not resolve.
+	value, path, err = ResolveShortNameAlias(sys, "empty") // from aliases.conf, overridden by config2
+	require.NoError(t, err)
+	assert.Nil(t, value)
+	assert.Equal(t, "testdata/aliases.conf", path)
+
+	mode, err := GetShortNameMode(sys)
+	require.NoError(t, err)
+	assert.Equal(t, types.ShortNameModePermissive, mode) // from alias.conf, overridden by config2
+
+	// Now remove the aliases from the machine config.
+	require.NoError(t, RemoveShortNameAlias(sys, "added1"))
+	require.NoError(t, RemoveShortNameAlias(sys, "added2"))
+	require.NoError(t, RemoveShortNameAlias(sys, "added3"))
+
+	// Make sure that 1 and 2 are gone.
+	for _, alias := range []string{"added1", "added2"} {
+		value, path, err := ResolveShortNameAlias(sys, alias)
+		require.NoError(t, err)
+		assert.Nil(t, value)
+		assert.Equal(t, "", path)
+	}
+
+	// 3 is still present in config2
+	value, path, err = ResolveShortNameAlias(sys, "added3")
+	require.NoError(t, err)
+	require.NotNil(t, value)
+	assert.Equal(t, "xxx.com/image", value.String())
+	assert.Equal(t, "testdata/registries.conf.d/config-2.conf", path)
+
+	require.Error(t, RemoveShortNameAlias(sys, "added3")) // we cannot remove it from config2
+}
+
+func TestInvalidAliases(t *testing.T) {
+	tmp, err := ioutil.TempFile("", "aliases.conf")
+	require.NoError(t, err)
+	defer os.Remove(tmp.Name())
+
+	sys := &types.SystemContext{
+		SystemRegistriesConfPath:    "testdata/invalid-aliases.conf",
+		SystemRegistriesConfDirPath: "testdata/this-does-not-exist",
+		UserShortNameAliasConfPath:  tmp.Name(),
+	}
+
+	configCache = make(map[configWrapper]*V2RegistriesConf)
+	_, err = TryUpdatingCache(sys)
+	require.Error(t, err)
+
+	// We validate the alias value before loading existing configuration,
+	// so this tests the validation although the pre-existing configuration
+	// is invalid.
+	assert.Error(t, AddShortNameAlias(sys, "added1", "aliases"))
+	assert.Error(t, AddShortNameAlias(sys, "added2", "aliases.conf"))
+	assert.Error(t, AddShortNameAlias(sys, "added3", ""))
+	assert.Error(t, AddShortNameAlias(sys, "added3", " "))
+	assert.Error(t, AddShortNameAlias(sys, "added3", "$$$"))
+}
--- a/pkg/sysregistriesv2/system_registries_v2.go
+++ b/pkg/sysregistriesv2/system_registries_v2.go
@@ -154,6 +154,28 @@ type V2RegistriesConf struct {
 	Registries []Registry `toml:"registry"`
 	// An array of host[:port] (not prefix!) entries to use for resolving unqualified image references
 	UnqualifiedSearchRegistries []string `toml:"unqualified-search-registries"`
+
+	// Absolut path to the configuration file that set the UnqualifiedSearchRegistries.
+	unqualifiedSearchRegistriesOrigin string
+
+	// ShortNameMode defines how short-name resolution should be handled by
+	// _consumers_ of this package.  Depending on the mode, the user should
+	// be prompted with a choice of using one of the unqualified-search
+	// registries when referring to a short name.
+	//
+	// Valid modes are: * "prompt": prompt if stdout is a TTY, otherwise
+	// use all unqualified-search registries * "enforcing": always prompt
+	// and error if stdout is not a TTY * "disabled": do not prompt and
+	// potentially use all unqualified-search registries
+	ShortNameMode string `toml:"short-name-mode"`
+
+	// TODO: separate upper format from internal data below:
+	// https://github.com/containers/image/pull/1060#discussion_r503386541
+
+	// shortNameMode is stored _once_ when loading the config.
+	shortNameMode types.ShortNameMode
+
+	shortNameAliasConf
 }
 
 // Nonempty returns true if config contains at least one configuration entry.
@@ -254,7 +276,7 @@ var anchoredDomainRegexp = regexp.MustCompile("^" + reference.DomainRegexp.Strin
 
 // postProcess checks the consistency of all the configuration, looks for conflicts,
 // and normalizes the configuration (e.g., sets the Prefix to Location if not set).
-func (config *V2RegistriesConf) postProcess() error {
+func (config *V2RegistriesConf) postProcessRegistries() error {
 	regMap := make(map[string][]*Registry)
 
 	for i := range config.Registries {
@@ -301,6 +323,7 @@ func (config *V2RegistriesConf) postProcess() error {
 				msg := fmt.Sprintf("registry '%s' is defined multiple times with conflicting 'insecure' setting", reg.Location)
 				return &InvalidRegistries{s: msg}
 			}
+
 			if reg.Blocked != other.Blocked {
 				msg := fmt.Sprintf("registry '%s' is defined multiple times with conflicting 'blocked' setting", reg.Location)
 				return &InvalidRegistries{s: msg}
@@ -385,6 +408,7 @@ func newConfigWrapper(ctx *types.SystemContext) configWrapper {
 		} else {
 			wrapper.userConfigDirPath = userRegistriesDirPath
 		}
+
 		return wrapper
 	} else if ctx != nil && ctx.RootForImplicitAbsolutePaths != "" {
 		wrapper.configPath = filepath.Join(ctx.RootForImplicitAbsolutePaths, systemRegistriesConfPath)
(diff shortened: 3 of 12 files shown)


vrothberg:
@baude @mtrmac @rhatdan ... early feedback is good feedback

I'm now moving on creating a `pkg/shortnames` to build logic around the it.

pkg/sysregistriesv2/alias.go
```diff
@@ -0,0 +1,112 @@
...
+
+// userAliasesFile is the user-specific registries.conf file to store aliases.
+var userAliasesFile = filepath.FromSlash(".config/containers/aliases.conf")
+
+// userAliasesLockFile is the lockfile for the user-specific registries.conf
+// file to store aliases.
+var userAliasesLockFile = filepath.FromSlash(".local/share/containers/aliases.conf.lock")
+
+// ResolveAlias performs an alias resolution of the specified name.  The
+// user-specific aliases.conf has precedence over aliases in `conf`.
```
mtrmac, line 21: Please document the return values.

pkg/sysregistriesv2/alias.go
```diff
@@ -0,0 +1,112 @@
...
+func (conf *V2RegistriesConf) ResolveAlias(ctx *types.SystemContext, name string) (string, bool, error) {
```
mtrmac, line 22: Could this return a `reference.Named`? I suspect inadvertent mixing of `string` values that all mean different things are a contributing factor to several of the semantic surprises in interpreting user input.

vrothberg:
Note: I'll add docs once the work has settled.

pkg/sysregistriesv2/alias.go
mtrmac, line 22: `V2RegistriesConf` is not a natural place to provide this; that type represents “a raw configuration file” basically for the purpose of `ConvertToV2`; there isn’t even a natural way for callers to _get_ a `V2RegistriesConf` value, and we don’t want them to think about individual files anyway.

I’d expect a top-level function similar to `FindRegistry`.

Also, I shouldn’t this use use the existing caching mechanism ? (I do feel fairly unsure about that, admittedly.)

pkg/sysregistriesv2/alias.go
```diff
@@ -0,0 +1,112 @@
...
+	value, resolved := aliasConf.Aliases[name]
+	if !resolved {
+		value, resolved = conf.Aliases[name]
+	}
+	return value, resolved, nil
+}
+
+// AddAliases adds the specified name-value pair as a new alias to the
+// user-specific aliases.conf.  It may override an existing alias for `name`.
+func (conf *V2RegistriesConf) AddAlias(ctx *types.SystemContext, name string, value string) error {
```
mtrmac, line 48: Same here:
- Make this a top-level function
- Consider making the `value` a `reference.Named`.

… and eventually probably validate the syntax of both the alias and the resolved value before writing it to the file.

pkg/sysregistriesv2/alias.go
mtrmac, line 22: Maybe `ResolveShortNameAlias` (and same for the other function)? Many things _could_ be aliased. OTOH that does feel rather wordy

pkg/sysregistriesv2/alias.go
```diff
@@ -0,0 +1,112 @@
...
+	"path/filepath"
+
+	"github.com/BurntSushi/toml"
+	"github.com/containers/image/v5/types"
+	"github.com/containers/storage/pkg/homedir"
+	"github.com/containers/storage/pkg/lockfile"
+)
+
+// userAliasesFile is the user-specific registries.conf file to store aliases.
+var userAliasesFile = filepath.FromSlash(".config/containers/aliases.conf")
```
mtrmac, line 14: `short-name-aliases.conf`? Users hopefully don’t need to type this, and many things in the container world could be or have aliases.

pkg/sysregistriesv2/alias.go
```diff
@@ -0,0 +1,112 @@
...
+	aliasesConfPath := ""
+	aliasesConfLockPath := ""
+
+	if ctx != nil && len(ctx.UserAliasesConfPath) > 0 {
+		aliasesConfPath = ctx.UserAliasesConfPath
+	} else {
+		aliasesConfPath = filepath.Join(homedir.Get(), userAliasesFile)
+	}
+
+	if ctx != nil && len(ctx.UserAliasesConfLockPath) > 0 {
```
mtrmac, line 104: I don’t think this should be configurable; the same configuration should always use the same lock file. Just hard-code a suffix.

pkg/sysregistriesv2/alias_test.go
```diff
@@ -0,0 +1,133 @@
...
+	sys := &types.SystemContext{
+		SystemRegistriesConfPath:    "testdata/aliases.conf",
+		SystemRegistriesConfDirPath: "testdata/this-does-not-exist",
+		UserAliasesConfPath:         tmp1.Name(),
+		UserAliasesConfLockPath:     tmp2.Name(),
+	}
+
+	configCache = make(map[configWrapper]*V2RegistriesConf)
+	conf, err := TryUpdatingCache(sys)
+	assert.Nil(t, err)
```
mtrmac, line 30:
```suggestion
	require.NoError(t)
```
an error makes `conf` `nil`, so the rest would crash without `require`. (Also elsewhere.)

pkg/sysregistriesv2/alias_test.go
```diff
@@ -0,0 +1,133 @@
...
+	assert.Equal(t, 3, len(conf.Aliases))
+
+	aliases := []struct {
+		name, value string
+	}{
+		{
+			"fooDocker",
+			"docker.io/library/foo:latest",
```
mtrmac, line 38: Do we really want to allow aliases to point to tags/digests?

My initial guess is “no” (and we can always add that later, but we won’t be able to remove it).

pkg/sysregistriesv2/alias_test.go
```diff
@@ -0,0 +1,133 @@
...
+		SystemRegistriesConfPath:    "testdata/aliases.conf",
+		SystemRegistriesConfDirPath: "testdata/this-does-not-exist",
+		UserAliasesConfPath:         tmp1.Name(),
+		UserAliasesConfLockPath:     tmp2.Name(),
+	}
+
+	configCache = make(map[configWrapper]*V2RegistriesConf)
+	conf, err := TryUpdatingCache(sys)
+	assert.Nil(t, err)
+	assert.Equal(t, 3, len(conf.Aliases))
```
mtrmac, line 31: `assert.Len` exists.  (Also elsewhere.)

pkg/sysregistriesv2/alias_test.go
```diff
@@ -0,0 +1,133 @@
...
+	assert.Nil(t, conf.AddAlias(sys, "added2", "aliases.conf/image:2"))
+	assert.Nil(t, conf.AddAlias(sys, "added3", "aliases.conf/image:3"))
+
+	for _, alias := range aliases {
+		value, resolved, err := conf.ResolveAlias(sys, alias.name)
+		assert.Nil(t, err)
+		assert.True(t, resolved)
+		assert.Equal(t, alias.value, value)
+	}
+}
```
mtrmac, line 133: At least one test for the `!resolved` case would be nice.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -149,17 +149,30 @@ func (config *V1RegistriesConf) Nonempty() bool {
 		len(config.V1TOMLConfig.Block.Registries) != 0)
 }
 
+// aliasesConf includes a map of aliases.
```
mtrmac, line 152: Maybe
```suggestion
// aliasesConf is the subset of the V2RegistriesConf format that is software-maintained in aliases.conf
```
or something vaguely like that?

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -149,17 +149,30 @@ func (config *V1RegistriesConf) Nonempty() bool {
...
+type aliasesConf struct {
+	// A map of aliases with keys being the alias names, replaced by the
+	// value.
```
mtrmac, line 155: (Eventually document syntax/semantics/format restrictions, if any.)

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -149,17 +149,30 @@ func (config *V1RegistriesConf) Nonempty() bool {
...
+	Aliases map[string]string `toml:"aliases"`
+}
+
 // V2RegistriesConf is the sysregistries v2 configuration format.
 type V2RegistriesConf struct {
 	Registries []Registry `toml:"registry"`
 	// An array of host[:port] (not prefix!) entries to use for resolving unqualified image references
 	UnqualifiedSearchRegistries []string `toml:"unqualified-search-registries"`
+	// Aliasing sets the aliasing mode (enabled, disabled, etc.).
+	Aliasing string `toml:"aliasing"`
```
mtrmac, line 165: - I’m not too happy about the “aliasing” name. If this controls non-alias behavior of Podman etc. as well, maybe `short-name-{behavior, handling}`?
- The allowed values, their semantics, and validation (rejecting typos and other values) will need to be nailed down before this is released.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -149,17 +149,30 @@ func (config *V1RegistriesConf) Nonempty() bool {
...
+
+	aliasesConf // cleanly separates aliases.conf from registries.conf
```
mtrmac, line 167: (Nitpick: The field’s purpose is not to “cleanly separate” anything. That’s the purpose of the `aliasesConf` type.)

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -640,8 +653,11 @@ func (c *tomlConfig) loadConfig(path string, forceV2 bool) error {
 		registryMap[c.Registries[i].Prefix] = c.Registries[i]
 	}
 
+	prevAliases := c.Aliases // store the aliases so they're not overridden
```
mtrmac, line 656: This quick hack has become too much. Please separate
- Load+TOML decode+format conversion into a _newly allocated `tomlConfig`→`V2RegistriesConf`.
- Merging of two `V2RegistriesConf` values into the effective config (vaguely `V2RegistriesConf.overrideByConfig(V2RegistriesConf)` or something like that)

(In a separate commit, or maybe even a separate PR, please.)

pkg/sysregistriesv2/testdata/aliases.conf
```diff
@@ -0,0 +1,4 @@
+[aliases]
+fooDocker="docker.io/library/foo:latest"
```
mtrmac, line 2: Hum, do we allow upper-case in aliases? In a sense, it would be great to have that (in the extreme _requiring_ upper-case would even make it unambiguous that the alias is not a `docker.io`/search short-name).

Either way, we need to explicitly document the allowed syntax. Maybe `regexp.go:nameComponentRegexp`, possibly with upper case allowed as well?

---

I haven’t checked but I’m somewhat sure that Kubernetes would be unhappy with non-`reference.Named` input (i.e. upper-case)

But, on second thought, It seems attractive to say that should explicitly _not_ be supported by CRI-O, even if Podman on the same host supports them , and even if CRI-O respects the short-name setting to configure what it does on search (reject short names completely / use old search behavior). Users of k8s don’t _need_ aliases as much, most of the configuration is somehow software-generated in either case (I _guess_, with no data), so let’s not introduce that concept for all of k8s to deal with?

OTOH consistency between Podman and CRI-O _is_ valuable. I don’t know.

types/types.go
```diff
@@ -509,6 +509,10 @@ type SystemContext struct {
 	SystemRegistriesConfPath string
 	// Path to the system-wide registries configuration directory
 	SystemRegistriesConfDirPath string
+	// Path to the user-specific aliases configuration file
+	UserAliasesConfPath string
```
mtrmac, line 513: (Also mentioned elsewhere:) This probably should talk about short names, to disambiguate. And …somehow… warn that aliases are only applicable to a specific subpackage, not to most of c/image.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -665,6 +681,16 @@ func (c *tomlConfig) loadConfig(path string, forceV2 bool) error {
...
+	c.Aliases = prevAliases // point back to the previous map and override
+	if c.Aliases == nil {
+		c.Aliases = make(map[string]string)
+	}
+	for name, value := range newAliases {
+		c.Aliases[name] = value
+	}
+
 	// Post process registries, set the correct prefixes, sanity checks, etc.
 	if err := c.postProcess(); err != nil {
```
mtrmac, line 695: Should aliases be validated for some format properties? Or maybe only in `Resolve`…?

mtrmac (commented):
Just a quick skim for immediate feedback, mixing comparatively important design concerns and trivial cleanups / naming quibbles, I’m afraid.

pkg/sysregistriesv2/alias.go
vrothberg: Good catch. I also just felt into the trap when working on `pkg/shortnames`.

pkg/sysregistriesv2/alias.go
vrothberg: I am torn as I enjoy the thought of preventing certain mistakes (and giving early feedback to the user in case they provided garbage in the conf) but it comes at cost of parsing each value when loading the config. If the list grew big, this could be a noticeable.

I have no feeling though how big the amount of aliases most be to have a noticeable performance impact. What's your take? I may be off.

vrothberg: I felt into the trap while working `pkg/shortnames`.

> Also, I shouldn’t this use use the existing caching mechanism ? 

Absolutely. That's the beauty of having a separate `aliases.conf` - we don't invalidate the registries.confs when adding a new alias.

pkg/sysregistriesv2/alias.go
vrothberg: Good point. I see wisdom in making that explicit in files that user will easily hit.

pkg/sysregistriesv2/alias_test.go
vrothberg: Can you elaborate on why tags may be a problem?

pkg/sysregistriesv2/alias.go
mtrmac: (I’m not worried about string copies/splits/… until given evidence to the contrary.)

`docker/reference` parsing is basically “regular expression” execution. I suspect those aren’t O(input size), but still probably good enough to support hundreds or thousands of values — especially with caching.

On my system, a `ParseNamed` takes around 3 microseconds (vs. 0.1 us for a `strings.Split`)

mtrmac: … actually, _at this point_ it does not matter. We might argue about the cost of validating all entries when reading the full config files into memory, but validating a _single_ entry in `ResolveAlias`/`AddAlias` is clearly cheap enough.

mtrmac: I was actually thinking about using the caching mechanism for `aliases.conf` as well — at least a single Podman command probably doesn’t care to, or might even actively _not_ want to, notice the alias changing over a run.

OTOH it _might_ be possible, with interactive use and indefinite user-driven delays, for users to observe that an entry was updated in one terminal but not used in another. So I’m not at all sure. (Either way, the caching is not contractual and we can add/remove it later.)

CRI-O already has the `InvalidateCache` etc. mechanism, assuming it should even care.

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: Added a TODO.

pkg/sysregistriesv2/alias.go
vrothberg: Just one is cheap but if we validate here, I also expect us to validate the configs to notify early on config errors.

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: If we want to validate I vote for validating here. If users have a malconfigured alias, I want them to know early on (ideally at start up). Otherwise it's entirely unpredictable when and if a user will hit the error when resolving to a malconfigured alias.

pkg/sysregistriesv2/testdata/aliases.conf
vrothberg: I leave this open as I don't yet have a feeling for a potential syntax.

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: Aliases are now validated on load and on `AddAlias`.

pkg/sysregistriesv2/system_registries_v2.go
rhatdan: Yes let's go with ShortName.  In case there is a future verion alias or people get confused between mirrors and alias.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -631,6 +641,9 @@ func FindRegistry(ctx *types.SystemContext, ref string) (*Registry, error) {
 // Note that specified fields in path will replace already set fields in the
 // tomlConfig.  Only the [[registry]] tables are merged by prefix.
 func (c *tomlConfig) loadConfig(path string, forceV2 bool) error {
+	// TODO: the code became hard to brain-parse and maintain. The function
+	// should separate the laoding, converting and merging steps more
```
rhatdan, line 645: laoding?

pkg/sysregistriesv2/testdata/aliases.conf
rhatdan: I think stick to lowercase, so we don't have to deal with Fedora being different then fedora.

rhatdan:
@baude @mheon @mrunalp @nalind PTAL

pkg/sysregistriesv2/testdata/aliases.conf
vrothberg: Still unadressed in the PR.

pkg/sysregistriesv2/alias.go
vrothberg: I wonder if it would make sense to have `name` be `reference.Reference`? It may avoid certain pitfalls _and_ `pkg/shortnames` will need to know if a user input includes a tag or digest and behave accordingly.

FWIW, I changed all return values to reference.Named

mtrmac: `reference.Reference` is a weaker type (it only provides .`String()`). `reference.Named` is _at least_ a repo name, and the `reference.Named` value commonly (as returned by `ParseNormalizedNamed`) includes a tag/digest, and that’s reachable by an extra type check.

vrothberg: I don't think we can use a `reference.Named` for the input as it may not include a repository. It could very well just be `foo` or `foo:tag`.

vrothberg: Argh...it doesn't make sense here but rather in `pkg/shortnames`.

go.mod
```diff
@@ -42,4 +42,5 @@ require (
 	golang.org/x/sys v0.0.0-20200810151505-1b9f1253b3ed
 	golang.org/x/time v0.0.0-20191024005414-555d28b269f0 // indirect
 	gopkg.in/check.v1 v1.0.0-20190902080502-41f04d3bba15 // indirect
+	gotest.tools v2.2.0+incompatible
```
mtrmac, line 45: What does this do?

pkg/sysregistriesv2/shortnames.go
```diff
@@ -0,0 +1,293 @@
...
+	"github.com/containers/storage/pkg/lockfile"
+	"github.com/pkg/errors"
+)
+
+// defaultShortNameMode is the default mode of registries.conf files if the
+// corresponding field is left empty.
+const defaultShortNameMode = types.ShortNameModePermissive
+
+// userShortNamesFile is the user-specific registries.conf file to store aliases.
+var userShortNamesFile = filepath.FromSlash(".config/containers/short-name-alias.conf")
```
mtrmac, line 21: Absolutely non-blocking: `s/alias/aliases/` ?

pkg/sysregistriesv2/shortnames.go
```diff
@@ -0,0 +1,293 @@
...
+
+// shortNameAliasConf is a subset of the `V2RegistriesConf` format.  It's used in the
+// software-maintained `userShortNamesFile`.
+type shortNameAliasConf struct {
+	// A map for aliasing short names to their fully-qualified image
+	// reference counter parts.
+	Aliases map[string]string `toml:"aliases"`
+
+	namedAliases map[string]reference.Named
```
mtrmac, line 30: Non-blocking? I’m not much a fan of mixing external formats and internal caches in a single type. (Are we certain the TOML unmarshal won’t allow writing into the cache? Even if we were, it is simpler to not even have to ask.)

`shortNameAliasConf.parseAndValidate` can easily return the cached data as a separate return value.

pkg/sysregistriesv2/shortnames.go
```diff
@@ -0,0 +1,293 @@
...
+func parseShortNameMode(mode string) (types.ShortNameMode, error) {
+	switch mode {
+	case "disabled":
+		return types.ShortNameModeDisabled, nil
+	case "enforcing":
+		return types.ShortNameModeEnforcing, nil
+	case "permissive", "":
+		return types.ShortNameModePermissive, nil
+	default:
+		return -1, errors.Errorf("invalid short-name mode: %q", mode)
```
mtrmac, line 44: Non-blocking? `-1` is not a valid value of the type, and there already is …`Invalid`. Could that be used?

pkg/sysregistriesv2/alias.go
mtrmac: Oops, you were asking about the _input_ to the lookup.

In my mind (which has _nothing_ to do with the actual software), only “normalized” values are `reference.Named`; short names are not (in Podman they are `imageParts`).

Basically, what I was asking for is to make the “possibly short name”, “certainly short name”, and “fully-qualified name” situations to be expressed by different Go types — certainly at least the fully-qualified from the other two. `string` vs. `reference.Named` is one way to do that — not at all an obvious way, admittedly.

pkg/sysregistriesv2/shortnames.go
```diff
@@ -0,0 +1,293 @@
...
+	}
+}
+
+// ResolveShortNameAlias performs an alias resolution of the specified name.
+// The user-specific short-name-aliases.conf has precedence over aliases in the
+// assembled registries.conf.  It returns whether the provided name could be
+// resolved to an alias, the possibly resolved alias, and an error.
```
mtrmac, line 51:
```suggestion
// resolved to an alias, the possibly resolved alias, and an error.
// Almost all callers should use pkg/shortnames instead.
```

pkg/sysregistriesv2/shortnames.go
```diff
@@ -0,0 +1,293 @@
...
+			errs = append(errs, err)
+		}
+
+		named, err := parseAlias(alias)
+		if err != nil {
+			// We want to report *all* malformed entries to avoid a
+			// whack-a-mole for the user.
+			errs = append(errs, err)
+		}
+		if len(errs) == 0 {
```
mtrmac, line 240: Non-blocking: I’d just use `else` here, the efficiency in an error case is not worth the mental overhead of figuring out that `len(errs)` implies `err == nil`, to me.

pkg/sysregistriesv2/shortnames.go
```diff
@@ -0,0 +1,293 @@
...
+	return err
+}
+
+func loadShortNameAliasConf(confPath string) (*shortNameAliasConf, error) {
+	conf := shortNameAliasConf{}
+
+	_, err := toml.DecodeFile(confPath, &conf)
+	switch err {
+	case nil:
+		if conf.Aliases == nil {
```
mtrmac, line 262: Doesn’t this need to happen in the `IsNotExist` case as well?

Why the `switch` and not `if`?

pkg/sysregistriesv2/shortnames_test.go
```diff
@@ -0,0 +1,307 @@
...
+		{"disabled", types.ShortNameModeDisabled, false},
+		{"enforcing", types.ShortNameModeEnforcing, false},
+		{"permissive", types.ShortNameModePermissive, false},
+		{"xxx", -1, true},
+	}
+
+	for _, test := range tests {
+		shortName, err := parseShortNameMode(test.input)
+		if test.mustFail {
+			require.Error(t, err)
```
mtrmac, line 29: Absolutely non-blocking: This one can be `assert.Error` because nothing will break if it is not true, we can just continue with another case. (OTOH, the test fails anyway, so, not at all worth worrying about.)

(Several other instances I won’t bother to point out individually.)

pkg/sysregistriesv2/shortnames_test.go
```diff
@@ -0,0 +1,307 @@
...
+		{"$$4455%%", false},              // gargabe
+		{"docker://foo", false},          // transports are not supported
+		{"docker-archive://foo", false},  // transport are not supported
+	}
+
+	for _, test := range tests {
+		named, err := parseAlias(test.input)
+		if test.valid {
+			require.NoError(t, err, "%q should be a valid alias", test.input)
+			assert.NotNil(t, named)
```
mtrmac, line 63: `assert.Equal(t, named.String(), test.input)` as well? Depends on whether we actually want to add that constraint (test case is `docker.io/fedora` I think).

pkg/sysregistriesv2/shortnames.go
```diff
@@ -0,0 +1,293 @@
...
+	// A map for aliasing short names to their fully-qualified image
+	// reference counter parts.
+	Aliases map[string]string `toml:"aliases"`
+
+	namedAliases map[string]reference.Named
+}
+
+// parseShortNameMode translates the string into well-typed
+// types.ShortNameMode.
+func parseShortNameMode(mode string) (types.ShortNameMode, error) {
```
mtrmac, line 35: This seems to be a better fit for the `sysregistriesv2.go` file right now; the only user is in there.

pkg/sysregistriesv2/system_registries_v2.go
mtrmac: Man pages will need to be updated.

pkg/sysregistriesv2/shortnames_test.go
```diff
@@ -0,0 +1,307 @@
...
+		err := validateShortName(test.input)
+		if test.valid {
+			require.NoError(t, err, "%q should be a valid alias", test.input)
+		} else {
+			require.Error(t, err, "%q should be an invalid alias", test.input)
+		}
+	}
+}
+
+func TestGetShortNameMode(t *testing.T) {
```
mtrmac, line 100: Move this to the test file corresponding to the source file that contains the implementation, please.

pkg/sysregistriesv2/shortnames_test.go
```diff
@@ -0,0 +1,307 @@
...
+		SystemRegistriesConfPath:    "testdata/aliases.conf",
+		SystemRegistriesConfDirPath: "testdata/this-does-not-exist",
+		UserShortNameAliasConfPath:  tmp.Name(),
+	}
+
+	configCache = make(map[configWrapper]*V2RegistriesConf)
+	conf, err := TryUpdatingCache(sys)
+	require.NoError(t, err)
+	assert.Len(t, conf.namedAliases, 3)
+	assert.Len(t, conf.Aliases, 0)
```
mtrmac, line 158: This is weird. If we are returning a value from `TryUpdatingCache`, shouldn’t it be accurate?

pkg/sysregistriesv2/shortnames_test.go
```diff
@@ -0,0 +1,307 @@
...
+	sys := &types.SystemContext{
+		SystemRegistriesConfPath:    "testdata/invalid-aliases.conf",
+		SystemRegistriesConfDirPath: "testdata/this-does-not-exist",
+	}
+
+	configCache = make(map[configWrapper]*V2RegistriesConf)
+	_, err := TryUpdatingCache(sys)
+	require.Error(t, err)
+
+	require.Error(t, AddShortNameAlias(sys, "added1", "aliases"))
```
mtrmac, line 302: Maybe use the temp file for `aliases.conf`, so that we don’t corrupt users’ configuration if any of the validations actually broke.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -154,12 +154,30 @@ type V2RegistriesConf struct {
...
+	// ShortNameMode defines how short-name resolution should be handled by
+	// _consumers_ of this package.  Depending on the mode, the user should
+	// be prompted with a choice of using one of the unqualified-search
+	// registries when referring to a short name.
+	//
+	// Valid modes are:
+	//  * "default": prompt if stdout is a TTY
+	//  * "enforcing": always prompt and error if stdout is not a TTY
+	//  * "disabled", "": do not prompt and potentially use all unqualified-search registries
+	ShortNameMode string `toml:"short-name-mode"`
```
mtrmac, line 166: Eventually the man page should be updated.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -154,12 +154,30 @@ type V2RegistriesConf struct {
...
+
+	// shortNameMode is stored _once_ when loading the config.
+	shortNameMode types.ShortNameMode
+	// ShortNameModeDisabled.
```
mtrmac, line 170: What does this comment mean?

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -154,12 +154,30 @@ type V2RegistriesConf struct {
...
+	// registries when referring to a short name.
+	//
+	// Valid modes are:
+	//  * "default": prompt if stdout is a TTY
+	//  * "enforcing": always prompt and error if stdout is not a TTY
+	//  * "disabled", "": do not prompt and potentially use all unqualified-search registries
+	ShortNameMode string `toml:"short-name-mode"`
+
+	// shortNameMode is stored _once_ when loading the config.
+	shortNameMode types.ShortNameMode
```
mtrmac, line 169: I’d prefer not mixing file formats and internal data.

Something like
```go
type parsedConfiguration struct {
  V2RegistriesConf // raw, merged - or only _some_ fields validly merged; or missing entirely
  shortNameMode
  parsedAliases 
}
```
could be generated and stored in the cache.

pkg/sysregistriesv2/testdata/aliases.conf
mtrmac: The PR’s validation currently disables upper-case because it uses `reference.Parse*`.

go.mod
vrothberg: Development artefact. Will remove it.

pkg/sysregistriesv2/shortnames.go
vrothberg: Sounds good

pkg/sysregistriesv2/system_registries_v2.go
mtrmac: This is actually a blocker for me. I feel strongly enough about this that I might even help :)

Related concerns:
- Separating internal caches from file format structs
- `TryUpdatingCache` should return the full configuration currently in force — definitely including the `*.d` files, undefined about the new alias.conf file (I vote “let’s not add this, and consider the idea of returning the full configuration deprecated”).

pkg/sysregistriesv2/shortnames.go
vrothberg: Not entirely sure yet. I leave this open to have another look when other parts of the code are done cooking.

pkg/sysregistriesv2/shortnames.go
vrothberg: Good idea

pkg/sysregistriesv2/shortnames.go
```diff
@@ -0,0 +1,293 @@
...
+	case "enforcing":
+		return types.ShortNameModeEnforcing, nil
+	case "permissive", "":
+		return types.ShortNameModePermissive, nil
+	default:
+		return -1, errors.Errorf("invalid short-name mode: %q", mode)
+	}
+}
+
+// ResolveShortNameAlias performs an alias resolution of the specified name.
```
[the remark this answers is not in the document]
vrothberg: Absolutely, thanks

pkg/sysregistriesv2/shortnames.go
```diff
@@ -0,0 +1,293 @@
...
+	if err != nil {
+		return false, nil, err
+	}
+	value, resolved = config.namedAliases[name]
+	return resolved, value, nil
+}
+
+// AddShortNameAliases adds the specified name-value pair as a new alias to the
+// user-specific aliases.conf.  It may override an existing alias for `name`.
+func AddShortNameAlias(ctx *types.SystemContext, name string, value string) error {
```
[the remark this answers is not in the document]
vrothberg: Good call! I'll leave open to revisit it later.

pkg/sysregistriesv2/shortnames.go
vrothberg: I think that's not needed any more. I'll check and simplify.

Thanks for checking!

pkg/sysregistriesv2/shortnames.go
```diff
@@ -0,0 +1,293 @@
...
+
+	named, ok := repo.(reference.Named)
+	if !ok {
+		return nil, errors.Errorf("invalid alias %q: must contain registry and repository", alias)
+	}
+
+	registry := reference.Domain(named)
+	if !(strings.ContainsAny(registry, ".:") || registry == "localhost") {
+		return nil, errors.Errorf("invalid alias %q: must contain registry and repository", alias)
+	}
```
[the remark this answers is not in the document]
vrothberg: It turns out we can't. `ParseNamed` does not accept `docker.io/fedora` but we can cast to it.

pkg/sysregistriesv2/shortnames_test.go
vrothberg: `conf.Aliases` is nil-ed when loading. I didn't like the thought of having the same data stored twice.

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: Removed. Must have forgotten to clean up.

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: Are you okay to add another TODO item (that I commit to addressing at a later point)? I think it makes sense to bundle this change together with breaking the big monolith of config loading into smaller pieces.

mtrmac: Yes, this + the `TryReloadingConfig` return value + merging are all related.

pkg/sysregistriesv2/shortnames.go
```diff
@@ -0,0 +1,261 @@
...
+	if err != nil {
+		return err
+	}
+
+	if value != nil {
+		conf.Aliases[name] = *value
+	} else {
+		// If the name does not exist, throw an error.
+		if _, exists := conf.Aliases[name]; !exists {
+			return errors.Errorf("short-name alias %q not found in %q: please check registries.conf files", name, confPath)
```
mtrmac, line 105: Should this be even more helpful, listing the currently configured alias?

mtrmac: The full UI should probably differentiate between “no alias defined at all” and “alias defined, but not in the editable file”.

pkg/sysregistriesv2/shortnames_test.go
```diff
@@ -0,0 +1,258 @@
...
+			require.NoError(t, err, "%q should be a valid alias", test.input)
+		} else {
+			require.Error(t, err, "%q should be an invalid alias", test.input)
+		}
+	}
+}
+
+func TestResolveShortNameAlias(t *testing.T) {
+	tmp, err := ioutil.TempFile("", "aliases.conf")
+	assert.Nil(t, err)
```
mtrmac, line 82: The three instances of this should be `require.NoError`.

go.mod
mtrmac: Still outstanding.

pkg/sysregistriesv2/shortnames.go
```diff
@@ -0,0 +1,293 @@
...
+		return errors.Errorf("invalid short name %q: must not contain tag", name)
+	}
+
+	named, ok := repo.(reference.Named)
+	if !ok {
+		return nil
+	}
+
+	registry := reference.Domain(named)
+	if strings.ContainsAny(registry, ".:") || registry == "localhost" {
```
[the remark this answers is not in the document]
mtrmac: We decided that `foo/bar` is a valid alias.

pkg/sysregistriesv2/shortnames_test.go
mtrmac: > I guess we can say that that doesn’t apply to these new values, and it’s not an API break, but it’s weird.

… it turns out I was advocating for exactly that in https://github.com/containers/image/pull/1060#discussion_r503391147 . So I can’t be feeling very strongly about the choice; if we decide to deprecate the returned configuration, let’s at least document that.

pkg/sysregistriesv2/shortnames.go
```diff
@@ -0,0 +1,287 @@
...
+	if _, ok := repo.(reference.Digested); ok {
+		return errors.Errorf("invalid short name %q: must not contain digest", name)
+	}
+
+	if _, ok := repo.(reference.Tagged); ok {
+		return errors.Errorf("invalid short name %q: must not contain tag", name)
+	}
+
+	named, ok := repo.(reference.Named)
+	if !ok {
```
mtrmac, line 186: This is AFAICS unreachable, but if it were reachable, it’s a good reason to fail.

pkg/sysregistriesv2/shortnames.go
```diff
@@ -0,0 +1,287 @@
...
+	// reference counter parts.
+	Aliases map[string]string `toml:"aliases"`
+
+	namedAliases map[string]reference.Named
+}
+
+// ResolveShortNameAlias performs an alias resolution of the specified name.
+// The user-specific short-name-aliases.conf has precedence over aliases in the
+// assembled registries.conf.  It returns whether the provided name could be
+// resolved to an alias, the possibly resolved alias, and an error.
```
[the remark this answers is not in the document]
vrothberg: It could be determined by `nil` check but I find the code much easier to read with:
```golang
resolved, named, err := ...

if resolved { ... }
```

pkg/sysregistriesv2/shortnames.go
```diff
@@ -0,0 +1,287 @@
...
+
+	aliasConf, err := loadShortNameAliasConf(confPath)
+	if err != nil {
+		return false, nil, err
+	}
+
+	// First look up the short-name-aliases.conf.  Note that a value may be
+	// nil iff it's set as an empty string in the config.
+	value, resolved := aliasConf.namedAliases[name]
+	if resolved && value != nil {
```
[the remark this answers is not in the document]
vrothberg: aliases.conf doesn't allow `= ""` at the moment.

pkg/sysregistriesv2/shortnames.go
vrothberg: I defer to future improvements for this one.

pkg/sysregistriesv2/shortnames.go
```diff
@@ -0,0 +1,287 @@
...
+	if err := conf.parseAndValidate(); err != nil {
+		return nil, errors.Wrapf(err, "error loading short-name aliases config file %q", confPath)
+	}
+
+	return &conf, nil
+}
+
+// ShortNameAliasesConfPath returns the path to the machine-generated
+// short-name-aliases.conf file.
+func ShortNameAliasesConfPath(ctx *types.SystemContext) (string, error) {
```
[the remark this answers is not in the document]
vrothberg: We could make it `./internal`

pkg/sysregistriesv2/shortnames_test.go
```diff
@@ -0,0 +1,258 @@
...
+		{"localhost:5000/fedora", true},
+		{"localhost:5000/namespace/fedora", true},
+		// INVALID INPUT
+		{"docker.io/library/fedora:latest", false}, // tag
+		{"docker.io/library/fedora@sha256:b87dd5f837112a9e1e9882963a6406387597698268c0ad371b187151a5dfe6bf", false}, // digest
+		{"fedora", false},                // short name
+		{"fedora:latest", false},         // short name + tag
+		{"library/fedora", false},        // no registry
+		{"library/fedora:latest", false}, // no registry + tag
+		{"$$4455%%", false},              // gargabe
```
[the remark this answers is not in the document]
vrothberg: :-) I leave it for entertainment

pkg/sysregistriesv2/shortnames.go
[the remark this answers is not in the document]
mtrmac: OTOH
```go
value, err := …
if value != nil { use value }
```
seems natural — and it makes it trivially visible that `value != nil` implies “`value` is safe to dereference”.

I guess this is not blocking, but `resolved == (named != nil)`, the documentation should clearly say so.

pkg/sysregistriesv2/shortnames.go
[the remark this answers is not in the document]
mtrmac: Ah, true, so it’s all consistent in a way. It’s not trivial to see from the code, which will parse `""` in aliases.conf just fine, though — and _especially_ given the comment above, which explicitly suggests an empty string may be used.

pkg/sysregistriesv2/shortnames.go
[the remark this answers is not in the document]
mtrmac: I was thinking just include it in `ConfigurationSourceDescription` … but now, having read #1061, I guess the `GuessPullError` API design needs to be settled first, as this is mostly an implementation detail of that.

pkg/sysregistriesv2/shortnames.go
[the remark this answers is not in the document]
vrothberg: You convinced me. I'll change :+1:

pkg/sysregistriesv2/shortnames_test.go
```diff
@@ -0,0 +1,258 @@
...
+		{"docker-archive://foo", false},  // transport are not supported
+		{"", false},                      // empty
+	}
+
+	for _, test := range tests {
+		named, err := parseShortNameValue(test.input)
+		if test.valid {
+			require.NoError(t, err, "%q should be a valid alias", test.input)
+			assert.NotNil(t, named)
+			assert.Equal(t, test.input, named.String())
```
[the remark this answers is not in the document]
vrothberg: Very thoughtful! I'll address it in `parseShortNameValue` as you suggest :+1:

pkg/sysregistriesv2/shortnames_test.go
vrothberg: Documented :+1:

vrothberg (commented):
thanks for the reviews

vrothberg:
@mtrmac ... I added means to track exactly in which config the USRs and each alias is specified. That'll help in reasoning over errors.

mtrmac:
(Immediate first look, not a review): That’s useful, of course, though maybe not strictly necessary for v1.
- Please let’s not talk about paths. “configuration origin”, “origin description”, fine, as long as it is explicitly only for human consumption and not to be read/opened.
- The separation of parsing+merging is still a blocker for me, this continues the current path.

I’ll take stab at the parsing/merging separation over the weekend, let’s see if it can be made minimally-disruptive.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -570,6 +594,40 @@ func UnqualifiedSearchRegistries(ctx *types.SystemContext) ([]string, error) {
 	return config.UnqualifiedSearchRegistries, nil
 }
 
+// UnqualifiedSearchRegistriesConfigOrigin returns the path to the config file
+// where the UnqualifiedSearchRegistries originate from.
+func UnqualifiedSearchRegistriesConfigOrigin(ctx *types.SystemContext) (string, error) {
```
mtrmac, line 599: I’d prefer a single-call `UnqualifiedSearchRegistriesConfigWithOrigin`; that would mean only one set of locks taken, only one error check in the caller, and less risk of the caller reading the two with different `SystemContext` values.

Also, this should talk about “human-readable description (if any)”, not “path”. Let’s not make any suggestions that the caller should access the file independently; also consider what happens if no config files exist, or if they exist but don’t contain any search registries configuration.

mtrmac:
> > I’ll take stab at the parsing/merging separation over the weekend, let’s see if it can be made minimally-disruptive.
> 
> FYI #1076, I’ll also try to prepare a variant on top of this PR.

#1077 . Note the semantics change for USR origin computation, otherwise it’s IMHO fairly straightforward, if tedious.

pkg/sysregistriesv2/shortnames.go
```diff
@@ -0,0 +1,313 @@
...
+// shortNameAliasConf is a subset of the `V2RegistriesConf` format.  It's used in the
+// software-maintained `userShortNamesFile`.
+type shortNameAliasConf struct {
+	// A map for aliasing short names to their fully-qualified image
+	// reference counter parts.
+	// Note that Aliases is niled after being loaded from a file.
+	Aliases map[string]string `toml:"aliases"`
+
+	// Note that a value may be nil iff it's set as an empty string in the
+	// config.
```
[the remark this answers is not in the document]
mtrmac: The comment still needs updating.

mtrmac:
@vrothberg The parsing/merging separation is still a blocker for me (there will never be “a more convenient time” to get that done). WDYT about the attempts above?

pkg/sysregistriesv2/shortnames.go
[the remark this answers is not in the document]
mtrmac: … actually, if it talks about `alias.value`, it’s accurate.

pkg/sysregistriesv2/shortnames.go
```diff
@@ -0,0 +1,313 @@
...
+	// in the config.
+	namedAliases map[string]alias
+}
+
+// ResolveShortNameAlias performs an alias resolution of the specified name.
+// The user-specific short-name-aliases.conf has precedence over aliases in the
+// assembled registries.conf.  It returns the possibly resolved alias, a path
+// to the config where the alias is specified, and an error. The path to the
+// config file is crucial for an improved user experience such that users are
+// able to resolve potential pull errors.  Almost all callers should use
```
mtrmac, line 66: s/path/human-readable description …/

pkg/sysregistriesv2/shortnames.go
mtrmac: Still outstanding: this should fail.

vrothberg:
> @vrothberg The parsing/merging separation is still a blocker for me (there will never be “a more convenient time” to get that done). WDYT about the attempts above?

They're looking good to me. Nice improvement overall!

vrothberg:
@mtrmac updated both PRs based on your latest comments.

pkg/sysregistriesv2/shortnames.go
```diff
@@ -0,0 +1,313 @@
...
+
+// ResolveShortNameAlias performs an alias resolution of the specified name.
+// The user-specific short-name-aliases.conf has precedence over aliases in the
+// assembled registries.conf.  It returns the possibly resolved alias, a
+// human-readable description of the config where the alias is specified, and
+// an error. The origin of the config file is crucial for an improved user
+// experience such that users are able to resolve potential pull errors.
+// Almost all callers should use pkg/shortnames instead.
+//
+// Note that the specified name is taken as it without normalizing it.
```
mtrmac, line 69: I’m not sure what this sentence means. That we treat `fedora` and `library/fedora` differently?

… looking back, this might be https://github.com/containers/image/pull/1060#discussion_r503347200 ; I tend to understand “normalization” as the docker.io[/library] thing.

“It’s the caller’s responsibility to pass only a repository (reference.IsNameOnly) as the short name”, or perhaps better, “… performs alias resolution of the specified short repository name”.

(All three instances.)

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -570,6 +594,42 @@ func UnqualifiedSearchRegistries(ctx *types.SystemContext) ([]string, error) {
...
+	case "enforcing":
+		return types.ShortNameModeEnforcing, nil
+	case "permissive", "":
+		return types.ShortNameModePermissive, nil
+	default:
+		return types.ShortNameModeInvalid, errors.Errorf("invalid short-name mode: %q", mode)
+	}
+}
+
+// GetShortNameMode returns the types.ShortNameMode in the config files.
```
mtrmac, line 624: Something like
```suggestion
// GetShortNameMode returns the configured types.ShortNameMode .
```
or s/configured/applicable/ ? It returns a value even if config files don’t include a value.

pkg/sysregistriesv2/shortnames.go
```diff
@@ -0,0 +1,313 @@
...
+	Aliases map[string]string `toml:"aliases"`
+
+	// Note that an alias value may be nil iff it's set as an empty string
+	// in the config.
+	namedAliases map[string]alias
+}
+
+// ResolveShortNameAlias performs an alias resolution of the specified name.
+// The user-specific short-name-aliases.conf has precedence over aliases in the
+// assembled registries.conf.  It returns the possibly resolved alias, a
```
mtrmac, line 63:
```suggestion
// assembled registries.conf.  It returns the possibly resolved alias or nil, a
```

mtrmac (commented):
To make some progress, I’m calling this the last iteration, and whatever I’ve missed, I’ve missed.

LGTM apart from the `parseShortName` built-in default; the rest is not strictly blocking.

How do you want to deal with the parsing/merging split? Do you want to include it here, or review+merge #1077 as a separate PR?

pkg/sysregistriesv2/shortnames.go
```diff
@@ -0,0 +1,313 @@
...
+	}
+	return err
+}
+
+func loadShortNameAliasConf(confPath string) (*shortNameAliasConf, error) {
+	conf := shortNameAliasConf{}
+
+	_, err := toml.DecodeFile(confPath, &conf)
+	if err != nil && !os.IsNotExist(err) {
+		// It's okay if the conig doesn't exist.  Other errors are not.
```
mtrmac, line 287:
```suggestion
		// It's okay if the config doesn't exist.  Other errors are not.
```

pkg/sysregistriesv2/shortnames.go
```diff
@@ -0,0 +1,313 @@
...
+
+	encoder := toml.NewEncoder(f)
+	return encoder.Encode(conf)
+}
+
+// AddShortNameAlias adds the specified name-value pair as a new alias to the
+// user-specific aliases.conf.  It may override an existing alias for `name`.
+//
+// Note that the specified name is taken as it without normalizing it.
+func AddShortNameAlias(ctx *types.SystemContext, name string, value string) error {
```
mtrmac, line 161: (I’d still slightly prefer the `value` parameter to Add/Remove to be a `reference.Named`, to add a minimum of type safety, OTOH passing strings probably makes it more convenient to at least one caller, and the value _is_ fairly strongly validated, so this is fair enough.)

vrothberg:
> To make some progress, I’m calling this the last iteration, and whatever I’ve missed, I’ve missed.

Cheers!

> How do you want to deal with the parsing/merging split? Do you want to include it here, or review+merge #1077 as a separate PR?

I'd prefer to merge as a separate PR.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -570,6 +594,42 @@ func UnqualifiedSearchRegistries(ctx *types.SystemContext) ([]string, error) {
 	return config.UnqualifiedSearchRegistries, nil
 }
 
+// UnqualifiedSearchRegistriesWithOrigin returns a list of host[:port] entries
```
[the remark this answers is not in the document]
vrothberg: Nice idea!

pkg/sysregistriesv2/shortnames_test.go
```diff
@@ -0,0 +1,273 @@
...
+	value, path, err := ResolveShortNameAlias(sys, "idonotexist")
+	require.NoError(t, err)
+	assert.Nil(t, value)
+	assert.Equal(t, "", path)
+
+	// Empty right-hand value (special case) -> does not resolve.
+	value, path, err = ResolveShortNameAlias(sys, "empty")
+	require.NoError(t, err)
+	assert.Nil(t, value)
+	assert.Equal(t, "", path)
```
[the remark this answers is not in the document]
vrothberg: Now it does. Good catch!

mtrmac (approved):
LGTM. Thanks for all the work!

pkg/sysregistriesv2/system_registries_v2_test.go
```diff
@@ -557,4 +558,76 @@ func TestRegistriesConfDirectory(t *testing.T) {
...
+func TestParseShortNameMode(t *testing.T) {
+	tests := []struct {
+		input    string
+		result   types.ShortNameMode
+		mustFail bool
+	}{
+		{"disabled", types.ShortNameModeDisabled, false},
+		{"enforcing", types.ShortNameModeEnforcing, false},
+		{"permissive", types.ShortNameModePermissive, false},
+		{"", types.ShortNameModePermissive, true},
```
mtrmac, line 577:
```suggestion
		{"", -1, true},
```
I guess. I’ll clean that up later.

vrothberg:
Thanks for the great reviews!

pkg/sysregistriesv2/system_registries_v2_test.go
mtrmac: https://github.com/containers/image/pull/1091

PR #1061, opened 2020-10-08
pkg/shortnames

vrothberg:
Add a new package for short-name resolution. `pkg/shortnames` is built        
around the short-name aliasing in the registries.conf and introduces two      
functions.                                                                    
                                                                              
`IsShortName(input string) (bool, error)`                                     
                                                                              
IsShortName is intended to reduce code duplication across consumers of        
the containers/image library.  Detecting short names is a common task,        
and centralizing the API will reduce the risk of bugs and/or behavioural      
differences.                                                                  
                                                                              
`Resolve(ctx *types.SystemContext, name string, clear bool) ([]string, error)`
                                                                              
Resolve resolves the specified name to either one or more                     
fully-qualified image references.  If the specified name is already a         
fully-qualified reference (i.e., not a short name), it is returned as         
is.  In case, it's a short name, it's resolved as follows:                    
                                                                              
If there is a matching short-name alias in the registries.conf, the           
alias is returned.  Otherwise, the behaviour is determined by the             
specified short-name mode as configured in the registries.conf:               
                                                                              
* "" or "disabled": append the name to all specified unqualified-search       
                    registries (USR).                                         
                                                                              
* "enforcing": prompt the user to select a USR.  Error if stdout is not a     
               TTY as we cannot prompt.                                       
                                                                              
* "permissive": prompt the user to select a USR.  If stdout is not a TTY,     
                fall back to the "disabled" behaviour.                        
                                                                              
Note that if there is only one USR, the USR is auto-selected without          
prompt.                                                                       
                                                                              
If `clear` is set, a possibly existing alias will be cleared                  
                                                                              

vrothberg:
This the PR focusing on the user-visible part that Podman, Buildah and friends can use. It bases on https://github.com/containers/image/pull/1060, so please ignore any changes outside of `pkg/shortnames`. Once https://github.com/containers/image/pull/1060 is merged, we can rebase here.

vrothberg:
I shamelessly recycled the SELinux terminology for the different modes :^)

pkg/shortnames/shortnames.go
```diff
@@ -0,0 +1,154 @@
...
+
+	registry := reference.Domain(named)
+	if strings.ContainsAny(registry, ".:") || registry == "localhost" {
+		return false, nil
+	}
+
+	return true, nil
+}
+
+func completeNameWithRegistry(name, registry string) string {
```
mtrmac, line 46: - Return a `reference.Named` (which may include a tag/digest)
- Ideally the input should be _only_ a value already processed by the “is the input a short name” question. Compare `parts.go`. Strings Are Bad™️.

pkg/shortnames/shortnames.go
```diff
@@ -0,0 +1,154 @@
...
+//                TTY as we cannot prompt.
+//
+// * "permissive": prompt the user to select a USR.  If stdout is not a TTY,
+//                 fall back to the "disabled" behaviour.
+//
+// Note that if there is only one USR, the USR is auto-selected without prompt.
+// If there is no USR, "docker.io" is selected.
+//
+// If `clear` is set, a possibly existing alias will be cleared.
+func Resolve(ctx *types.SystemContext, name string, clear bool) ([]string, error) {
```
mtrmac, line 72: Return a `reference.Named` (possibly with a digest/tag).

pkg/shortnames/shortnames.go
```diff
@@ -0,0 +1,154 @@
...
+	// If `input` is not a short name, we're done.
+	isShort, err := IsShortName(name)
```
mtrmac, line 74: (This might very well be fine, mostly a note to self: I’m wondering whether this belongs here, vs. other heuristics like “is this an image ID prefix”? I’ll need to check the uses)

pkg/shortnames/shortnames.go
mtrmac, line 72: The “clear” parameter is strange. Why is clearing an alias a part of its resolution? What does `Resolve(fully-qualified-name, clear=true)` mean?

pkg/shortnames/shortnames.go
```diff
@@ -0,0 +1,154 @@
...
+	unqualifiedSearchRegistries, err := sysregistriesv2.UnqualifiedSearchRegistries(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	candidates := []string{}
+	switch len(unqualifiedSearchRegistries) {
+	case 0:
+		// No search registries, so pick the default one.
+		return []string{completeNameWithRegistry(name, defaultRegistry)}, nil
```
mtrmac, line 106: Isn’t this wrong? See above.

pkg/shortnames/shortnames.go
```diff
@@ -0,0 +1,154 @@
...
+// short-name mode as configured in the registries.conf:
+//
+// * "" or "disabled": append the name to all specified unqualified-search
+//                     registries (USR).
+//
+// * "enforcing": prompt the user to select a USR.  Error if stdout is not a
+//                TTY as we cannot prompt.
+//
+// * "permissive": prompt the user to select a USR.  If stdout is not a TTY,
+//                 fall back to the "disabled" behaviour.
```
mtrmac, line 66: None of disabled/enforcing/permissive actually describe what the behavior is. Who would answer “Return all options” to “what does a ‘disabled’ setting mean?” ?

pkg/shortnames/shortnames_test.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+func TestIsShortName(t *testing.T) {
+	tests := []struct {
+		input       string
+		isShortName bool
+		mustFail    bool
+	}{
+		// SHORT NAMES
+		{
```
mtrmac, line 17: (Could each `{…},` be condensed to one line? That would mean less scrolling and make it easier to check what cases are covered.)

pkg/sysregistriesv2/shortnames.go
```diff
@@ -0,0 +1,233 @@
...
+	Aliases map[string]string `toml:"aliases"`
+}
+
+// ShortNameMode defines how short-name resolution should be handled by
+// _consumers_ of this package.
+type ShortNameMode int
+
+const (
+	// Do not prompt.
+	ShortNameModeDisabled = iota // we default to it for backwards compat!
```
mtrmac, line 33: `ShortNameMode{Invalid,Uninitialized} ShortNameMode = iota`, to
- Make the constants well-typed
- Detect uses of uninitialized values

pkg/shortnames/shortnames.go
```diff
@@ -0,0 +1,154 @@
...
+		Items:    candidates,
+		HideHelp: true, // do not show navigation help
+	}
+
+	_, selection, err := prompt.Run()
+	if err != nil {
+		return nil, err
+	}
+
+	if err := sysregistriesv2.AddShortNameAlias(ctx, name, selection); err != nil {
```
mtrmac, line 149: So the UI model is that the user must choose a registry to pull from, that choice is recorded, and _then_ we try pulling. If the user chose “wrong” and the image does not actually exist there, it’s going to be annoying to recover.

(Yes, I know, this is clearly more secure than letting the user guess and then record an alias for the first image that actually successfully pulls. Still, is it intentional/desired?)

---

Another aspect to this: In the “disabled” setting, I assume the full set of options is used to pull. Wouldn’t we still want to record the outcome of the pull after the search completes, so that future uses of the same short name always use the same registry? (I’m sure this was covered in the design doc. I need to look it up, link it to this PR, and then revisit.)

mtrmac (commented):
Just a _very minimal_ quick skim. I haven’t really thought through the options and interaction model at all.

pkg/shortnames/shortnames.go
vrothberg: Very fair point. Removed.

Still in a raw state.

pkg/shortnames/shortnames.go
vrothberg: > Still, is it intentional/desired?

Intentional, yes. Desired, up to debate I guess.

For now it makes testing and development easier. Recording the name/alias pair and adding them _after_ a successful pull would be a more intrusive change, and I didn't want to invest time into something that I believe is an open question.

I'll set up a meeting next week where we can go through the details :+1: 

> In the “disabled” setting, I assume the full set of options is used to pull. Wouldn’t we still want to record the outcome of the pull after the search completes, so that future uses of the same short name always use the same registry? 

I intended "disabled" to fully opt-out from aliasing. I will push the latest development state in a jiffy where the three modes should make more sense. If unspecified, we now default to "permissive" which I see as a compromise of aliasing "where possible". Taking the above point a bit further in the permissive context, it may really be better to only add an alias once it's know to be pulled successfully. This way, "permissive" would record aliases even without a TTY/prompt.

pkg/shortnames/shortnames.go
```diff
@@ -0,0 +1,225 @@
...
+
+	// Query the registry for unqualified-search registries.
+	unqualifiedSearchRegistries, err := sysregistriesv2.UnqualifiedSearchRegistries(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	// Detect which mode we're running in.
+	var mode types.ShortNameMode
+	if ctx != nil && ctx.ShortNameMode != nil {
```
mtrmac, line 127: I’d expect `GetShortNameMode` to handle this, so that it is the same in all callers of that function.

pkg/shortnames/shortnames.go
```diff
@@ -0,0 +1,225 @@
...
+		return nil, errors.Errorf("unsupported short-name mode (%v)", mode)
+	}
+
+	// Strip off the tag to normalize the short name for looking it up in
+	// the config files.
+	normalizedShortName := name
+	withTag := ""
+	if tagged, isTagged := shortRef.(reference.Tagged); isTagged {
+		withTag = tagged.Tag()
+		normalizedShortName = strings.TrimSuffix(name, ":"+withTag)
```
mtrmac, line 165: `reference.TrimNamed`, please.

pkg/shortnames/shortnames.go
```diff
@@ -0,0 +1,225 @@
...
+
+	// If there's already an alias, use it.
+	isAliased, namedAlias, err := sysregistriesv2.ResolveShortNameAlias(ctx, normalizedShortName)
+	if err != nil {
+		return nil, err
+	}
+
+	if isAliased {
+		// If the input was tagged, add it to the alias.
+		if len(withTag) > 0 {
```
mtrmac, line 176: `isTagged` was already a boolean. Keeping the `tagged, isTagged` variables visible here would allow dropping the `withTag` variable entirely.

pkg/shortnames/shortnames.go
```diff
@@ -0,0 +1,225 @@
...
+	"github.com/manifoldco/promptui"
+	"github.com/pkg/errors"
+	"golang.org/x/crypto/ssh/terminal"
+)
+
+// IsShortName returns true if the specified input is a "short name" and an
+// error that may occur when the input could not be parsed.  A "short name"
+// refers to a container image without a fully-qualified reference, and is
+// hence missing a registry (or domain).  Names including a digest are not
+// short names.
```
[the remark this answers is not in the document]
vrothberg: > I get that digested names are safe to use with search, but does that mean digested names wouldn’t use aliases?

I added this issue to today's meeting topics. I am not 100 percent sure.

> // Names including a digest are not short names.

With a fresh pair of eyes, I think that should be reverted.

pkg/shortnames/shortnames.go
```diff
@@ -0,0 +1,225 @@
...
+	isShort, shortRef, err := isShortName(name)
+	if err != nil {
+		return err
+	}
+	if !isShort {
+		return errors.Errorf("%q is not a short name", name)
+	}
+	normalizedShortName := name
+	if tagged, isTagged := shortRef.(reference.Tagged); isTagged {
+		normalizedShortName = strings.TrimSuffix(name, ":"+tagged.Tag())
```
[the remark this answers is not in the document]
vrothberg: Without it, all callers had to strip the tag on their own. I imagine callers to use the package as follows:

```golang
candidates, err := shortnames.Resolve(..., name)
...
for _, candidate := range {
 ...
}
...
// Use the same name as above and let the API normalize.
if err := shortnames.Add(..., name, succesfulCandidate); err != nil {
...
}
```

pkg/shortnames/shortnames.go
vrothberg: I don't think we can use as not all references are named.

pkg/shortnames/shortnames.go
```diff
@@ -0,0 +1,225 @@
...
+
+// Resolve resolves the specified name to either one or more fully-qualified
+// image references.  If the specified name is already a fully-qualified
+// reference (i.e., not a short name), it is returned as is.  In case, it's a
+// short name, it's resolved according to the ShortNameMode in the
+// SystemContext (is specified) or in the registries.conf.
+//
+// Note that returned slice may be empty if no matching alias and no
+// unqualified-search registries could be found.  Names including a digest are
+// returned as is.  Tags are stripped from the specified name.
```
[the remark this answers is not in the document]
vrothberg: Let's discuss this part in today's meeting. I am not entirely sure how we want that part to behave.

pkg/shortnames/shortnames.go
mtrmac: IIRC all references that can get here are named; the only way to obtain a digest-only reference is `ParseAnyReference`.

(And if the input weren’t a named reference, the above would not be correct anyway.)

vrothberg: Makes sense :+1:

pkg/sysregistriesv2/shortnames.go
```diff
@@ -0,0 +1,274 @@
...
+	"github.com/containers/storage/pkg/lockfile"
+	"github.com/pkg/errors"
+)
+
+// defaultShortNameMode is the default mode of registries.conf files if the
+// corresponding field is left empty.
+const defaultShortNameMode = types.ShortNameModePermissive
+
+// userShortNamesFile is the user-specific registries.conf file to store aliases.
+var userShortNamesFile = filepath.FromSlash(".cache/containers/short-name-aliases.conf")
```
edsantiago, line 21: Could you use `$XDG_CACHE_HOME` if set? It is the [defined convention](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html), and would make testing easier/safer as well.

vrothberg: Sounds good. I'll update the code (and the docs).

docs/containers-registries.conf.5.md
```diff
@@ -102,6 +102,68 @@ internet without having to change `Dockerfile`s, or to add redundancy).
...
+* `permissive`: Behaves as enforcing but does not lead to an error if the
+  program is not running in a terminal.  Instead, fallback to using all
+  unqualified-search registries.
+
+* `disabled`: Use all unqualified-search registries without prompting.
+
+If the user-specified short name was not alias already the `enforcing` and
+`permissive` mode will record a new alias after a successful pull.  Note that
+the recorded alias will be written to
+`$XDG_CONFIG_HOME/.config/containers/short-name-aliases.conf` to have a clear
```
edsantiago, line 159: Correction: remove `.config` (it is implicit in `$XDG_CONFIG_HOME`)

vrothberg: Thanks for catching that!

pkg/shortnames/shortnames.go
```diff
@@ -0,0 +1,472 @@
...
+}
+
+// GuessPullError returns an error that _guesses_ why a the specified short
+// name could not resolve to a successful pull candidate.  This can be helpful
+// for giving some hints to users to resolve pull errors.  Note that nil is
+// returned if it's not a short name.
+//
+// GuessPullError is expected to be called if none of the `Resolved` candidates
+// from `Resolve` lead to a successfully pulled image.
+func GuessPullError(ctx *types.SystemContext, name string) error {
```
vrothberg, line 93: @mtrmac, this could be a tricky one. @rhatdan asked to improve the errors and I think Buildah and Podman should share this code.

@rhatdan PTAL

vrothberg: Now further generalized in `func GuessPullError(ctx *types.SystemContext, name string, pullErrors []error) error {`

docs/containers-registries.conf.5.md
```diff
@@ -102,6 +102,70 @@ internet without having to change `Dockerfile`s, or to add redundancy).
...
+#### Short-Name Aliasing
+The use of unqualified-search registries entails an ambiguity as it's
+unclear from which registry a given image, referenced by a short name,
+may be pulled from.
+
+As mentioned in the note at the end of the document, using short names is
+subject to the risk of spoofed registry namespaces.  If the unqualified-search
+registries are set to `["registry1.com", "registry2.com"]` an attacker may take
+over a namespace of registry1.com such that an image may be pulled from
+registry1.com instead of the intended soruce registry2.com.
```
TomSweeneyRedHat, line 114:
```suggestion
registry1.com instead of the intended source registry2.com.
```

docs/containers-registries.conf.5.md
```diff
@@ -102,6 +102,70 @@ internet without having to change `Dockerfile`s, or to add redundancy).
 *Note*: Redirection and mirrors are currently processed only when reading images, not when pushing
 to a registry; that may change in the future.
 
+#### Short-Name Aliasing
+The use of unqualified-search registries entails an ambiguity as it's
```
TomSweeneyRedHat, line 106: suggestion: it is not wrong, but I think it's better to not use a contraction.
```suggestion
The use of unqualified-search registries entails an ambiguity as it is
```

docs/containers-registries.conf.5.md
```diff
@@ -102,6 +102,70 @@ internet without having to change `Dockerfile`s, or to add redundancy).
...
+unclear from which registry a given image, referenced by a short name,
+may be pulled from.
+
+As mentioned in the note at the end of the document, using short names is
```
TomSweeneyRedHat, line 110:
```suggestion
As mentioned in the note at the end of this man page, using short names is
```

docs/containers-registries.conf.5.md
```diff
@@ -102,6 +102,70 @@ internet without having to change `Dockerfile`s, or to add redundancy).
...
+subject to the risk of spoofed registry namespaces.  If the unqualified-search
+registries are set to `["registry1.com", "registry2.com"]` an attacker may take
+over a namespace of registry1.com such that an image may be pulled from
+registry1.com instead of the intended soruce registry2.com.
+
+While it is highly recommended to always use fully-qualified image references,
+existing deployments using short names may not be easily changed.  To
+circumvent the aforementioned ambiguity, so called short-name aliases can be
+configured that allow to alias a short name to a fully-qualified image
```
TomSweeneyRedHat, line 119:
```suggestion
configured that point to a fully-qualified image
```
or perhaps "resolve" instead of "point"

docs/containers-registries.conf.5.md
```diff
@@ -102,6 +102,70 @@ internet without having to change `Dockerfile`s, or to add redundancy).
...
+reference.
+
+Short-name aliases can be configured in the `[aliases]` table in the form of
+`"name"="value"` with the left-hand `name` being the short name (e.g., "image")
+and the right-hand `value` being the fully-qualified image reference (e.g.,
+"registry.com/namespace/image").  Note that neither "name" nor "value" can
+include a tag or digest.  Moreover, "name" must be a short name and hence
+cannot include a registry domain or refer to localhost.
```
TomSweeneyRedHat, line 127: Do we guard against duplicate aliases entered into the table?  Which one do we use if they accidentally add one and do we need to mention anything here about that?

TomSweeneyRedHat: maybe I'll discover it later, but we have `quay.io/buildah/stable:latest :v1.16.0` and other  tags.  Does the alias always pull latest?  Is there a way to specify a particular tag?  i.e.  `fedora:31` vs `fedora:32`

docs/containers-registries.conf.5.md
```diff
@@ -102,6 +102,70 @@ internet without having to change `Dockerfile`s, or to add redundancy).
...
+  select one of the specified search registries.  If the program is not running
+  in a terminal, the ambiguity cannot be resolved which will lead to an error.
+
+* `permissive`: Behaves as enforcing but does not lead to an error if the
+  program is not running in a terminal.  Instead, fallback to using all
+  unqualified-search registries.
+
+* `disabled`: Use all unqualified-search registries without prompting.
+
+If the user-specified short name was not aliased already the `enforcing` and
```
TomSweeneyRedHat, line 156:
```suggestion
If the user-specified short name was not aliased already, the `enforcing` and
```

docs/containers-registries.conf.5.md
```diff
@@ -102,6 +102,70 @@ internet without having to change `Dockerfile`s, or to add redundancy).
...
+`permissive` mode, if prompted, will record a new alias after a successful
```
TomSweeneyRedHat, line 157:
```suggestion
`permissive` mode if prompted, will record a new alias after a successful
```

docs/containers-registries.conf.5.md
```diff
@@ -102,6 +102,70 @@ internet without having to change `Dockerfile`s, or to add redundancy).
...
+pull.  Note that the recorded alias will be written to
+`$XDG_CONFIG_HOME/containers/short-name-aliases.conf` to have a clear
+separation between possibly human-edited registries.conf files and the
+machine-generated `short-name-aliases-conf`.  Note that `$HOME/.config` is used
+if `$XDG_CONFIG_HOME` is not set.  If and alias is specified in a
```
TomSweeneyRedHat, line 162:
```suggestion
if `$XDG_CONFIG_HOME` is not set.  If an alias is specified in a
```

docs/containers-registries.conf.5.md
```diff
@@ -102,6 +102,70 @@ internet without having to change `Dockerfile`s, or to add redundancy).
...
+`registries.conf` file and the machine-generated
```
TomSweeneyRedHat, line 163:
```suggestion
`registries.conf` file and also the machine-generated
```

docs/containers-registries.conf.5.md
```diff
@@ -102,6 +102,70 @@ internet without having to change `Dockerfile`s, or to add redundancy).
...
+`short-name-aliases.conf`, the latter has precedence.
```
TomSweeneyRedHat, line 164: suggestion:  latter can sometimes be confusing
```suggestion
`short-name-aliases.conf`, the `short-name-aliases.conf`file has precedence.
```

docs/containers-registries.conf.5.md
```diff
@@ -102,6 +102,70 @@ internet without having to change `Dockerfile`s, or to add redundancy).
...
+
+If `short-name-mode` is not specified at all or left empty, default to the
+`permissive` mode.
```
TomSweeneyRedHat, line 167: I'd move these lines up to be just under the 3 bullet points, somwhere near line 156

pkg/shortnames/errors.go
```diff
@@ -0,0 +1,118 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/containers/image/v5/pkg/sysregistriesv2"
+	"github.com/containers/image/v5/types"
+	"github.com/pkg/errors"
+)
+
+// GuessPullErrors returns an error that _guesses_ why a the specified short
```
TomSweeneyRedHat, line 12: nit
```suggestion
// GuessPullErrors returns an error that _guesses_ why the specified short
```

pkg/shortnames/shortnames.go
```diff
@@ -0,0 +1,396 @@
...
+// name.
+func (r *Resolved) String() string {
+	return r.Value.String()
+}
+
+// Resolve resolves the specified name to either one or more fully-qualified
+// image references that the short name may be *pulled* from.  If the specified
+// name is already a fully-qualified reference (i.e., not a short name), it is
+// returned as is.  In case, it's a short name, it's resolved according to the
+// ShortNameMode in the SystemContext (is specified) or in the registries.conf.
```
TomSweeneyRedHat, line 173:
```suggestion
// ShortNameMode in the SystemContext (if specified) or in the registries.conf.
```

pkg/shortnames/shortnames.go
```diff
@@ -0,0 +1,396 @@
...
+	unqualifiedSearchRegistries, err := sysregistriesv2.UnqualifiedSearchRegistries(sys)
+	if err != nil {
+		return nil, err
+	}
+
+	var candidates []Resolved
+	for _, reg := range unqualifiedSearchRegistries {
+		named, err := reference.ParseNormalizedNamed(fmt.Sprintf("%s/%s", reg, name))
+		if err != nil {
+			return nil, errors.Wrapf(err, "error creating reference with unqalified-serach registry %q", reg)
```
TomSweeneyRedHat, line 257:
```suggestion
			return nil, errors.Wrapf(err, "error creating reference with unqualified-serach registry %q", reg)
```

pkg/shortnames/shortnames_test.go
```diff
@@ -0,0 +1,458 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestIsShortName(t *testing.T) {
+	tests := []struct {
+		input       string
+		isShortName bool
+		mustFail    bool
+	}{
```
TomSweeneyRedHat, line 20: It might be good to add to or modify the tests to pull quay.io/skopeo/stable and a variety of it's tags.

docs/containers-registries.conf.5.md
vrothberg: I'll clarify.

Duplicate entries will cause a TOML load error. Tags and digests are not allowed in any config.

vrothberg:
Thanks a lot for the review and comments, @TomSweeneyRedHat. Much appreciated!

vrothberg:
Ready to go from my side.

rhatdan:
LGTM
@mtrmac PTAL

vrothberg:
Can we get this moving? Podman and Buildah are blocked on release.

rhatdan:
@mtrmac we are waiting for you go-ahead.

pkg/shortnames/shortnames.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	named, ok := ref.(reference.Named)
+	if !ok {
+		return false, ref, nil
+	}
+
+	registry := reference.Domain(named)
+	if strings.ContainsAny(registry, ".:") || registry == "localhost" {
+		return false, ref, nil
```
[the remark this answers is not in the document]
mtrmac: Still outstanding; at least in the `!isShort` code paths this is consumed as if it were normalized.

pkg/shortnames/shortnames.go
```diff
@@ -0,0 +1,225 @@
...
+		return false, ref, errors.Wrapf(err, "cannot parse input: %q", input)
+	}
+
+	if _, isDigested := ref.(reference.Digested); isDigested {
+		return false, ref, nil
+	}
+
+	named, ok := ref.(reference.Named)
+	if !ok {
+		return false, ref, nil
```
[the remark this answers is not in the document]
mtrmac: Still outstanding.

pkg/shortnames/errors.go
```diff
@@ -0,0 +1,118 @@
...
+// to resolve pull errors.  Note that nil is returned if it's not a short name.
+//
+// GuessPullErrors is expected to be called if none of the `Resolved`
+// candidates from `Resolve` lead to a successfully pulled image.  The returned
+// error should be displayed to users as is and not be wrapped any further.
+func GuessPullError(ctx *types.SystemContext, name string, pullErrors []error) error {
+	// If there are no pull errors and we could not pull an image,
+	// something went really wrong.
+	if len(pullErrors) == 0 {
+		return errors.Errorf("Pulling %q failed for unknown reasons", name)
```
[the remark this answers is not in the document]
vrothberg: I don't think we can reach this condition but I added it for completeness and to be sure.

If there's no USR, we should error out before. If there's at least one USR and we didn't pull an image, we should have at least one pull error.

pkg/shortnames/errors.go
```diff
@@ -0,0 +1,118 @@
...
+	var sb strings.Builder
+	if guess != nil {
+		sb.WriteString(guess.Error())
+		sb.WriteString("\n\n")
+	}
+
+	term := "error"
+	if len(pullErrors) > 1 {
+		term += "s"
+	}
```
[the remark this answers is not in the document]
vrothberg: I went the lazy path and used "error(s)"

pkg/shortnames/errors.go
```diff
@@ -0,0 +1,118 @@
...
+
+// GuessPullErrors returns an error that _guesses_ why the specified short
+// name could not resolve to a successful pull candidate along with the
+// specified pull errors.  This can be helpful for giving some hints to users
+// to resolve pull errors.  Note that nil is returned if it's not a short name.
+//
+// GuessPullErrors is expected to be called if none of the `Resolved`
+// candidates from `Resolve` lead to a successfully pulled image.  The returned
+// error should be displayed to users as is and not be wrapped any further.
+func GuessPullError(ctx *types.SystemContext, name string, pullErrors []error) error {
```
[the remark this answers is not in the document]
vrothberg: Resolve does not attempt to pull. Both Buildah and Podman tried to reason over why a pull may have failed and that's attempt to consolidate the code and give the same errors.

I spent some time thinking of embedding the logic into `Resolved` and `Resolve` but I found that would over-complicate things. In the end, it's a way to have a consistent formatting plus some hints to resolve the errors as Podman and Buildah did before.

pkg/shortnames/errors.go
```diff
@@ -0,0 +1,118 @@
...
+
+	sb.WriteString(fmt.Sprintf("%d %s occurred trying to pull %q:", len(pullErrors), term, name))
+	for _, err := range pullErrors {
+		sb.WriteString("\n * ")
+		sb.WriteString(err.Error())
+	}
+	return errors.New(sb.String())
+}
+
+func guessPullError(ctx *types.SystemContext, name string) error {
```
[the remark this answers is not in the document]
vrothberg: I was thinking about it as well but found it to be overcomplicating things.

Also we are very late and I have barely any time left to revisit fundamental design decisions.

pkg/shortnames/errors.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	// Query the registry for unqualified-search registries.
+	unqualifiedSearchRegistries, err := sysregistriesv2.UnqualifiedSearchRegistries(ctx)
+	if err != nil {
+		// No need to have a redundant log if any.
+		return nil
+	}
+
+	if len(unqualifiedSearchRegistries) == 0 {
```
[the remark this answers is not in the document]
vrothberg: Agreed. Added for completeness / extra security net.

pkg/shortnames/errors.go
```diff
@@ -0,0 +1,118 @@
...
+		return errors.Errorf("No matching short-name alias has been found for %q and no unqualified-search registries are configured in %q or %q",
+			name, configPath, configDirPath)
+	}
+
+	return errors.Errorf("No matching short-name alias has been found for %q and no unqualified-search registry in %q or %q includes a matching image",
```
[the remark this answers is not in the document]
vrothberg: Fair point! I reworded the end.

docs/containers-registries.conf.5.md
```diff
@@ -102,6 +102,68 @@ internet without having to change `Dockerfile`s, or to add redundancy).
...
+If `short-name-mode` is not specified at all or left empty, default to the
+`permissive` mode.  If the user-specified short name was not aliased already,
+the `enforcing` and `permissive` mode if prompted, will record a new alias
+after a successful pull.  Note that the recorded alias will be written to
+`$XDG_CONFIG_HOME/containers/short-name-aliases.conf` to have a clear
+separation between possibly human-edited registries.conf files and the
+machine-generated `short-name-aliases-conf`.  Note that `$HOME/.config` is used
+if `$XDG_CONFIG_HOME` is not set.  If an alias is specified in a
+`registries.conf` file and also the machine-generated
+`short-name-aliases.conf`, the `short-name-aliases.conf`file has precedence.
```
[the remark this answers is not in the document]
vrothberg: I placed it here because the `short-name-aliases.conf` relates to prompting which is tied to the modes.

docs/containers-registries.conf.5.md
```diff
@@ -102,6 +102,68 @@ internet without having to change `Dockerfile`s, or to add redundancy).
...
+be pulled from registry1.com instead of the intended source registry2.com.
+
+While it is highly recommended to always use fully-qualified image references,
+existing deployments using short names may not be easily changed.  To
+circumvent the aforementioned ambiguity, so called short-name aliases can be
+configured that point to a fully-qualified image
+reference.
+
+Short-name aliases can be configured in the `[aliases]` table in the form of
+`"name"="value"` with the left-hand `name` being the short name (e.g., "image")
```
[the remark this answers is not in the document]
vrothberg: Great catch! I added a paragraph to it also pointing to registries.conf.d.

pkg/shortnames/shortnames.go
```diff
@@ -0,0 +1,396 @@
...
+// error that may occur when the input could not be parsed.  A "short name"
+// refers to a container image without a fully-qualified reference, and is
+// hence missing a registry (or domain).  Names including a digest are not
+// short names.
+//
+// Examples:
+//  * short names: "image:tag", "library/fedora"
+//  * not short names: "quay.io/image", "localhost/image:tag",
+//                     "server.org:5000/lib/image", "image@sha256:..."
+func IsShortName(input string) (bool, error) {
```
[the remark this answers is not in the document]
vrothberg: I think the input validation has _some_ service to the users as they can detect "some" garbage earlier. We might very well just return a bool if you prefer.

I tend to add errors to APIs even if they're always nil at the beginning. That'll allow for future extensions without breaking the API.

pkg/shortnames/shortnames.go
```diff
@@ -0,0 +1,396 @@
...
+	_, _, normalizedShortName, _, _ := normalizeUserInput(shortRef)
+	return sysregistriesv2.RemoveShortNameAlias(ctx, normalizedShortName)
+}
+
+// Resolved is a resolved short name.  Once the Value has been used
+// successfully, users MUST call `(*Resolved).Record(..)` to possibly record it
+// as a new short-name alias.
+type Resolved struct {
+	// The short name.
+	Name string
```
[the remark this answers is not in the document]
vrothberg: It's only used in `Record` now, so we can make it private.

pkg/shortnames/shortnames.go
```diff
@@ -0,0 +1,396 @@
...
+	}
+	if err := Add(r.systemContext, r.Name, r.Value.String()); err != nil {
+		return errors.Wrapf(err, "error recording short-name alias (%q=%q)", r.Name, r.Value)
+	}
+	return nil
+}
+
+// String returns the fully-qualified image reference of the resolved short
+// name.
+func (r *Resolved) String() string {
```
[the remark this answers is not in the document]
vrothberg: Removed. Was only used for testing.

pkg/shortnames/shortnames.go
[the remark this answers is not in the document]
vrothberg: Can you elaborate? I don't understand what's wrong.

vrothberg (commented):
GitHub wants a top-level comment for my review-reply.

pkg/shortnames/shortnames_test.go
```diff
@@ -0,0 +1,458 @@
...
+		require.Len(t, aliases, 1)
+		assert.Equal(t, test.value, aliases[0].String())
+		assert.False(t, aliases[0].record)
+	}
+
+	// Non existant should return and empty slice as no search registries
+	// are configured in the config.
+	aliases, err := Resolve(sys, "dontexist")
+	require.NoError(t, err)
+	require.Len(t, aliases, 0)
```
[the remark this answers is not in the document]
vrothberg: I considered it but found the sequences here to be easier to digest.

pkg/shortnames/shortnames_test.go
```diff
@@ -0,0 +1,458 @@
...
+			continue
+		}
+		require.NoError(t, err, "%v", test)
+	}
+}
+
+func TestResolveAndRecord(t *testing.T) {
+	tests := []struct {
+		mode       types.ShortNameMode
+		mustRecord bool
```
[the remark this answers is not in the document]
vrothberg: Yes. It'll only be true when prompting. It was changed at some point.

This tests needs some more love (and mocking of the recording)

pkg/shortnames/shortnames_test.go
```diff
@@ -0,0 +1,458 @@
...
+	require.NoError(t, Remove(sys, "added1"))
+	require.NoError(t, Remove(sys, "added2:tag"))
+	require.NoError(t, Remove(sys, "added3@sha256:d366a4665ab44f0648d7a00ae3fae139d55e32f9712c67accd604bb55df9d05a"))
+
+	// Doesn't exit -> error.
+	require.Error(t, Remove(sys, "added1"))
+	require.Error(t, Remove(sys, "added2:tag"))
+	require.Error(t, Remove(sys, "added3@sha256:d366a4665ab44f0648d7a00ae3fae139d55e32f9712c67accd604bb55df9d05a"))
+
+	// Cannot remove entries from registries.conf files -> error.
```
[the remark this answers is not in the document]
vrothberg: I don't feel strongly. Let me know if you want it. Currently, `""` is not allowed in the aliases.conf.

vrothberg:
Note that I am attending a conference today (and tomorrow). I may need until tomorrow afternoon to wrap things up.

vrothberg (commented):
.

pkg/shortnames/errors.go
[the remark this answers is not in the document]
mtrmac: Clever :)

pkg/shortnames/errors.go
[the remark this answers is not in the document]
mtrmac: - The values returned currently don’t _at all_ guess “why a pull may have failed”. That might be very useful to build a heuristic for, but the current implementation is not trying. (I guess, if that’s the real goal, build a placeholder heuristic that fails to make a recommendation (`if true { }`?), just to show where it would go and what the output would look like.)
- The thing the implementation _is_ doing, providing context for why a particular set of qualified names was chosen, is already turning out to be difficult to get right.

pkg/shortnames/errors.go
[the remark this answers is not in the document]
mtrmac: It seems to me that replacing
```go
return []Resolved{newResolved(_, _, record, ctx)
```
with a
```go
return ResolutionResult{
    Resolved: []Resolved{Resolved{_, _}},
    Rationale: fmt.Sprintf(…)
    record: record,
    sys: ctx,
}
```
or something like that is a pretty cheap refactoring — it’s basically one reason string per return path in `Resolve`, isn’t it?

pkg/shortnames/shortnames.go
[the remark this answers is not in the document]
mtrmac: > I think the input validation has _some_ service to the users as they can detect "some" garbage earlier.

Well, that’s fine, but that requires a _definition_ of “garbage”. I guess what I mean is that the documentation should say that “input” is expected to be a valid (qualified or unqualified) Docker-like named reference. Then we can return errors on “garbage”. Without such a definition, `docker://UPPERCASEISINVALID` is just “not a short name”, not “garbage”.

> I tend to add errors to APIs even if they're always nil at the beginning. That'll allow for future extensions without breaking the API.

That’s a good general guideline sure, OTOH how is the _caller_ going to use that value? If we don’t specify different semantics for `false` and error, we can’t expect the caller to treat them differently — and so we can’t add meaningful errors anyway.

pkg/shortnames/shortnames.go
```diff
@@ -0,0 +1,396 @@
...
+		return
+	}
+
+	return
+}
+
+// Add records the specified name-value pair as a new short-name alias to the
+// user-specific aliases.conf.  It may override an existing alias for `name`.
+//
+// Note that tags and digests are stripped from the specified name and value.
```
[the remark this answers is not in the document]
mtrmac: I see the value in telling everyone to use this sub package instead of the config directly (ideally the documentation in that package should point here, then).

As for the stripping, I’m worried about `podman alias fedora:latest = fedoraproject.org/fedora:f33` succeeding without error, and then `podman pull fedora:latest` surprisingly not pulling `:f33`. What will actually benefit from silently stripping the data?

pkg/shortnames/shortnames.go
```diff
@@ -0,0 +1,396 @@
...
+// short name), it is returned as is.  In case, it's a short name, the returned
+// slice of named references looks as follows:
+//
+//  1) If present, the short-name alias
+//  2) "localhost/" as used by many container engines such as Podman and Buildah
+//  3) Unqualified-search registries from the registries.conf files
+//
+// Note that tags and digests are stripped from the specified name before
+// looking up an alias.  The returned slice contains at least one item.
+func ResolveLocally(ctx *types.SystemContext, name string) ([]reference.Named, error) {
```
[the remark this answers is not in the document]
mtrmac: Fine, if this is `Runtime.NewFromLocal` only.

What is `Runtime.New` going to do? Call `Resolve` and hard-code prepending `localhost/`? That would try to prepend `localhost` even if the input is fully-qualified. It seems to me that this should be handled in this package as well, OTOH I haven’t taken a detailed look at the consumers yet.

pkg/shortnames/shortnames_test.go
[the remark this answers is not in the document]
mtrmac: I don’t actually feel strongly about `""` in aliases.conf either — I just misread the registries code and thought that it was allowed. If it is clearly and consistently not allowed, that’s fine with me, we can always add that later.

pkg/shortnames/shortnames.go
```diff
@@ -0,0 +1,396 @@
...
+
+	// Query the registry for unqualified-search registries.
+	unqualifiedSearchRegistries, err := sysregistriesv2.UnqualifiedSearchRegistries(sys)
+	if err != nil {
+		return nil, err
+	}
+
+	var candidates []Resolved
+	for _, reg := range unqualifiedSearchRegistries {
+		named, err := reference.ParseNormalizedNamed(fmt.Sprintf("%s/%s", reg, name))
```
mtrmac, line 255: (Absolutely non-blocking: In both cases of this loop: This is of course correct, but a tiny bit non-obvious that this code path can ignore `isTagged`/`isDigested`/`tag`/`digest`. Consolidating/factoring out the re-tagging/re-digesting and using it consistently on all paths might be worth considering. OTOH this is short and simple.)

pkg/shortnames/shortnames.go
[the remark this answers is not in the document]
mtrmac: AFAICS `Resolve("docker.io/fedora")` calls `isShortName("docker.io/fedora")`, passes through here, and returns a `reference.Named` with `.String() == "docker.io/fedora"`.

If that reference is used for `c/image/docker.NewReference` without passing through a string; that causes incorrect values to be generated for policy matching, causes `reference.Path(ref)` to be `fedora` and not `library/fedora`, causing incorrect URLs to be used when talking to the registry, breaks lookups by name in local storage — and generally breaks assumptions that `reference.Named` values are all in normalized form and can be compared for equality without re-normalization all over the place.

pkg/shortnames/errors.go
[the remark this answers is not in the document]
mtrmac: Compare also the history of Podman’s `pullGoal.{usedSearchRegistries,searchedRegistries}`: That was added because the previous version was also guessing what probably happened only after a failure, and was guessing wrong and/or getting out of sync with the actual decision logic.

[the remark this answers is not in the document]
vrothberg: I guess you're right. We're late, so I am shielding bigger changes but this one makes sense to push through.

How about we:
* Record during `Resolve` how a given `Resolved` came to be (alias, USR, none)
* Add a new method `(*Resolved) WrapError(err error) error` which will add the context information to the error. This way each error will give us more information.

Future cookies/work: record during config loading where the USRs come from and where each alias comes from.

pkg/shortnames/shortnames.go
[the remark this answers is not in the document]
vrothberg: Will change to return an error only :+1:

pkg/shortnames/errors.go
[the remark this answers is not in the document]
mtrmac: How would the caller know what single `Resolved` to use? It can’t well iterate though all the `Resolved` candidate, call `WrapPullError` on all of them, and combine the result, because that would usually repeat the same explanation N times.

alias/USR/none is a single decision that feels like a single return value to me (see `ResolutionResult.Rationale` below, OTOH that has unresolved issues as to how the record/SystemContext values participate in recording outcomes).

Admittedly if (some variant of) `Resolve` starts adding `localhost/` entries, we _may_ need different explanation for entries (e.g. “`localhost/` comes from application policy, the others come from USR”).

Or maybe combine both, and turn `GuessPullError([]error)` into `ResolutionResult.WrapError(errors []error)` , requiring that `len(errors) == len(ResolutionResult.Resolved)` but otherwise giving us the option to record any amount of data that later turns out to be useful as private members of `ResolutionResult`.

(As usual, my naming suggestions suck. Please improve on them.)

pkg/shortnames/shortnames.go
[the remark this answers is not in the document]
vrothberg: Gotcha, thanks!

pkg/shortnames/errors.go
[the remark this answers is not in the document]
vrothberg: > How would the caller know what single Resolved to use? It can’t well iterate though all the Resolved candidate, call WrapPullError on all of them, and combine the result, because that would usually repeat the same explanation N times.

Consider:
```golang
candidates, _ := shortnames.Resolve(...)

for _, c := range candidates {
   _, err := pull(c)
   if err != nil {
     err = c.WrapPullError(err)
   }
}
```
AFAICS that would eliminate the need for `GuessPullErrors` entirely.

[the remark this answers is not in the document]
vrothberg: At least that would explain where a given fully-qualified reference originated from.

vrothberg:
Updated the c/image PRs to track exactly where a given USR or alias comes from. Extended `Resolved` with a `WrapPullError` method that can be used during pulling to add some context information to the pull error.

I updated the Buildah PR and really like it:
```
4 errors occured while pulling:
 * "registry.fedoraproject.org/123123:latest" has been completed with an unqualified-search registry (see /etc/containers/registries.conf): Error initializing source docker://registry.fedoraproject.org/123123:latest: Error reading manifest latest in registry.fedoraproject.org/123123: manifest unknown: manifest unknown
 * "registry.access.redhat.com/123123:latest" has been completed with an unqualified-search registry (see /etc/containers/registries.conf): Error initializing source docker://registry.access.redhat.com/123123:latest: Error reading manifest latest in registry.access.redhat.com/123123: name unknown: Repo not found
 * "registry.centos.org/123123:latest" has been completed with an unqualified-search registry (see /etc/containers/registries.conf): Error initializing source docker://registry.centos.org/123123:latest: Error reading manifest latest in registry.centos.org/123123: manifest unknown: manifest unknown
 * "docker.io/library/123123:latest" has been completed with an unqualified-search registry (see /etc/containers/registries.conf): Error initializing source docker://123123:latest: Error reading manifest latest in docker.io/library/123123: errors:
unauthorized: authentication required
```

A follow-up work could tackle the redundancy and trim down the error messages.

vrothberg:
```
"i.do/not/exist:latest" is a recorded short-name alias (see /home/vrothberg/.config/containers/short-name-aliases.conf): Error initializing source docker://i.do/not/exist:latest: error pinging docker registry i.do: Get "https://i.do/v2/": dial tcp: lookup i.do: no such host
```

vrothberg:
@mtrmac @rhatdan PTAL

pkg/shortnames/shortnames.go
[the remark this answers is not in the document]
vrothberg: `Runtime.New` would continue calling `NewFromLocal` and if needed call `Resolve`. Note sure why we had to pull from `localhost/`.

pkg/shortnames/errors.go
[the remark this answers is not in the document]
mtrmac: That creates one `err` value per loop iteration; and without further heuristics the caller will have to show all of them (the way `GuessPullError` did). So all of them will include exactly the same rationale (especially in the case of USR), the and the caller will end up with ~2-5 errors all starting with the same “%q has been completed with an unqualified-search registry” prefix.

True, it would also show the fully-qualified name, but IIRC that shows up in error messages anyway (and it’s also the part that the caller of `Resolve` can easily do itself, if truly necessary).

mtrmac:
> A follow-up work could tackle the redundancy and trim down the error messages.

A general note: The most binding constraint on c/image is Go modules + semver imposing that the API should not break callers on updates. So I’m much more concerned about the public API shape making things _possible_ in the way that we intend to than about the current implementation. I.e. for a heuristic function, it’s much more important for the function to have the right inputs+outputs+context than to actually _have_ a heuristic.

pkg/shortnames/shortnames.go
```diff
@@ -0,0 +1,474 @@
...
+// IsShortName returns true if the specified input is a "short name".  A "short
+// name" refers to a container image without a fully-qualified reference, and
+// is hence missing a registry (or domain).  Names including a digest are not
+// short names.
+//
+// Examples:
+//  * short names: "image:tag", "library/fedora"
+//  * not short names: "quay.io/image", "localhost/image:tag",
+//                     "server.org:5000/lib/image", "image@sha256:..."
+func IsShortName(input string) bool {
```
mtrmac, line 26: BTW neither https://github.com/containers/buildah/pull/2695 nor https://github.com/containers/podman/pull/7964 calls this, AFAICS. So maybe it can be dropped.

mtrmac:
> > A follow-up work could tackle the redundancy and trim down the error messages.
> 
> I.e. for a heuristic function, it’s much more important for the function to have the right inputs+outputs+context than to actually _have_ a heuristic.

So, if we want to avoid the redundancy and note the USR origin only once, there must be a way for the caller to do that. Hence the API shape suggestion to have a top-level rationale returned from `Resolve`, or a `ResolutionResult`, where the origin could be recorded, and the heuristic API to consume both that top-level output and the list of errors.

----

BTW if we are in a time pressure, it makes more sense to me to be deferring some features, not rapidly expanding the scope of the PR. If I said something that seemed to suggest that we must perfect the heuristics implementation now, I apologize.

pkg/shortnames/shortnames.go
```diff
@@ -0,0 +1,474 @@
...
+	case rationaleNone:
+		fallthrough
+	default:
+		return err
+	}
+}
+
+// FormatPullErrors is a convenience function to format errors that occured
+// while trying to pull one or more `Resolved` items.
+func FormatPullErrors(pullErrors []error) error {
```
mtrmac, line 214: This is more or less an UI aesthetics decision; it _can_ live here but it’s not all that necessary (compare https://github.com/hashicorp/go-multierror as previously used in one of the pull implementations).

As it is, it does not easily allow implementing the de-duplication (it still requires the caller to individually `WrapError` every single one, and then here we would have to un-wrap that, and we have neither a `Resolved` nor any other object to carry state, so the full state of the resolution would have to be passed through every single `Resolved` and every single error returned through `WrapError` only to be deduplicated here.

And what happens if the caller mistakenly returns `pullErrors` from more than one `Resolve` call?

Basically it seems to me that we will deprecate this after the first improvement.

---

```go
type ResolutionResult {
   Results []reference.Named // or an array of structs? Not sure, we can always add a top-level array to expand.
   unqualified-search-origin string
   itemRationale []rationale
}

func (rr *ResolutionResult) FormatPullErrors(errors []error) string { // or WrapPullErrors that returns an error? Not sure; It's not very machine-consumable
  // errors must match Canonical
  // Just list them, or add context, or prioritize, or whatever
}

// Possible FUTURE extensions (let's not expand this PR, please)

// Returns a short-form description suitable for progress displays before starting to pull (e.g. without including paths and other info that is only useful for diagnosing failures)
func (rr *ResolutionResult) ShortDescription() string

// Returns a short-form description suitable for progress displays, before starting to pull rr.Results[index] (e.g. without including paths and other info that is only useful for diagnosing failures)
func (rr *ResolutionResult) ResultContext(index index) string

// Pull in order, give appropriate progress updates, and so on
func c/common/sharedpull.PullResolvedImage(rr *ResolutionResult, progressWriter, …)
```

pkg/shortnames/shortnames.go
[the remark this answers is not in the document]
mtrmac: Works for me.

pkg/shortnames/shortnames.go
vrothberg: You're making a pretty convincing point. Encapsulating all data in a single type is a wise thing to do.

vrothberg:
@mtrmac thanks for the reviews and comments. Current state is shown below. Based on the conversation around redundancy and encapsulating into a single object. Independent of an error, it may actually be a good idea to always print the context.

```
podman (shortnames) $ ./bin/podman pull alpine                                                                                   
Resolved short name "alpine" to a recorded short-name alias (origin: /home/vrothberg/.config/containers/short-name-aliases.conf) 
Trying to pull docker.io/library/alpine:latest...                                                                                
Copying blob 188c0c94c7c5 done                                                                                                   
Copying config d6e46aa247 done                                                                                                   
Storing signatures                                                                                                               
d6e46aa2470df1d32034c6707c8041158b652f38d2a9ae3d7ad7e7532d22ebe0                                                                 
podman (shortnames) $ ./bin/podman pull nginx                                                                                    
Completed short name "nginx" with unqualified-search registries (origin: /etc/containers/registries.conf)                        
Trying to pull registry.fedoraproject.org/nginx:latest...                                                                        
  manifest unknown: manifest unknown                                                                                             
Trying to pull registry.access.redhat.com/nginx:latest...                                                                        
  unsupported: This repo requires terms acceptance and is only available on registry.redhat.io                                   
Trying to pull registry.centos.org/nginx:latest...                                                                               
  manifest unknown: manifest unknown                                                                                             
Trying to pull docker.io/library/nginx:latest...                                                                                 
Copying blob 5a9f1c0027a7 [=>------------------------------------] 1.4MiB / 25.3MiB                                              
Copying blob bb79b6b2107f [>-------------------------------------] 714.5KiB / 25.8MiB                                            
Copying blob b5c20b2b484f done                                                                                                   
Copying blob 166a2418f7e8 done                                                                                                   
Copying blob 1966ea362d23 done                                                                                                   
```

rhatdan:
Pull this would be fine, but it should NOT do this for podman run, since this could interfere with people collecting stderr and stdout.

vrothberg:
It uses the same io Writer as the progress bars, so run would work as
before. Should be stderr by default.

vrothberg:
@mtrmac are we good to go?

pkg/shortnames/shortnames.go
```diff
@@ -0,0 +1,477 @@
...
+	// Fully-qualified reference with tag or digest.
+	Value reference.Named
+	// Input from the user.
+	inputName string
+	// Control whether to record it permanently as an alias.
+	record bool
+	// The system context used when looking up the alias.  Always recording
+	// the system context allows for easier testing as we can simulate
+	// recording even if `record` would inititally be set false.
+	systemContext *types.SystemContext
```
mtrmac, line 211: Maybe a pointer back to parent `*Resolved` instead of some of the duplication? Hopefully the GC can deal with the cycles.

pkg/shortnames/shortnames.go
```diff
@@ -0,0 +1,477 @@
...
+	case rationaleUserSelection, rationaleNone:
+		fallthrough
+	default:
+		return ""
+	}
+}
+
+// FormatPullErrors is a convenience function to format errors that occured
+// while trying to pull one or more PullCandidates.
+func FormatPullErrors(pullErrors []error) error {
```
mtrmac, line 181: This should be a method on `Resolved` so that it can benefit from all the private data in there, and the API should require `len(pullErrors) == len(Resolved.PullCandidates)`.

mtrmac (commented):
From a very quick skim, this is primarily blocked on the #1060 parsing/merging separation.

docs/containers-registries.conf.5.md
```diff
@@ -102,6 +102,75 @@ internet without having to change `Dockerfile`s, or to add redundancy).
...
+When pulling a short name, the configured aliases table will be used for
+resolving the short name.  If a matching alias is found, it will be used
+without further consulting the unqualified-search registries list.  If no
+matching alias is found, the behavior can be controlled via the
+`short-name-mode` option as described below.
+
+Note that tags and digests are stripped off a user-specified short name for
+alias resolution.  Hence, "image", "image:tag" and "image@digest" all resolve
+to the same alias (i.e., "image").  Stripped off tags and digests are later
+appeneded to the resolved alias.
```
mtrmac, line 138:
```suggestion
appended to the resolved alias.
```

docs/containers-registries.conf.5.md
```diff
@@ -102,6 +102,75 @@ internet without having to change `Dockerfile`s, or to add redundancy).
...
+If `short-name-mode` is not specified at all or left empty, default to the
+`permissive` mode.  If the user-specified short name was not aliased already,
+the `enforcing` and `permissive` mode if prompted, will record a new alias
+after a successful pull.  Note that the recorded alias will be written to
+`$XDG_CONFIG_HOME/containers/short-name-aliases.conf` to have a clear
+separation between possibly human-edited registries.conf files and the
+machine-generated `short-name-aliases-conf`.  Note that `$HOME/.config` is used
+if `$XDG_CONFIG_HOME` is not set.  If an alias is specified in a
+`registries.conf` file and also the machine-generated
+`short-name-aliases.conf`, the `short-name-aliases.conf`file has precedence.
```
mtrmac, line 172:
```suggestion
`short-name-aliases.conf`, the `short-name-aliases.conf` file has precedence.
```

pkg/shortnames/shortnames_test.go
```diff
@@ -0,0 +1,473 @@
...
+		{"repo/fedora", "repo/fedora", false, false},
+		{"registry.com/fedora", "registry.com/fedora", false, false},
+		// Tags
+		{"fedora:tag", "fedora", true, false},
+		{"repo/fedora:tag", "repo/fedora", true, false},
+		{"registry.com/fedora:latest", "registry.com/fedora", true, false},
+		// Digests
+		{"fedora@sha256:d366a4665ab44f0648d7a00ae3fae139d55e32f9712c67accd604bb55df9d05a", "fedora", false, true},
+		{"repo/fedora@sha256:d366a4665ab44f0648d7a00ae3fae139d55e32f9712c67accd604bb55df9d05a", "repo/fedora", false, true},
+		{"registry.com/fedora@sha256:d366a4665ab44f0648d7a00ae3fae139d55e32f9712c67accd604bb55df9d05a", "registry.com/fedora", false, true},
```
mtrmac, line 73: It would be nice to add a “repo:tag@digest” test case; AFAICS the current code works but I have already started considering a refactoring before noticing that it would break that case.

mtrmac (commented):
I’ve tried hard for this to be the last iteration like #1060, OTOH there’s more discussion outstanding.

(Also pointing to a few earlier discussions that are not quite resolved.)

types/types.go
```diff
@@ -509,6 +539,10 @@ type SystemContext struct {
 	SystemRegistriesConfPath string
 	// Path to the system-wide registries configuration directory
 	SystemRegistriesConfDirPath string
+	// Path to the user-specific short-names configuration file
+	UserShortNameAliasConfPath string
+	// If set, short-name resolution must follow the specified mode
+	ShortNameMode *ShortNameMode
```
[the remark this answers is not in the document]
vrothberg: added "in pkg/shortnames" to make it clearer

pkg/shortnames/shortnames.go
```diff
@@ -0,0 +1,483 @@
...
+// IsShortName returns true if the specified input is a "short name".  A "short
+// name" refers to a container image without a fully-qualified reference, and
+// is hence missing a registry (or domain).  Names including a digest are not
+// short names.
+//
+// Examples:
+//  * short names: "image:tag", "library/fedora"
+//  * not short names: "quay.io/image", "localhost/image:tag",
+//                     "server.org:5000/lib/image", "image@sha256:..."
+func IsShortName(input string) bool {
```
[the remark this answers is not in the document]
vrothberg: There are a couple of places in Podman that could benefit from it. It's a reoccurring thing, and having a single source of truth would be good.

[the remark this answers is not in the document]
vrothberg: Note: they are not addressed in the shortnames PR against Podman. I want to limit the scope to remote and local resolution there.

pkg/shortnames/shortnames.go
```diff
@@ -0,0 +1,396 @@
...
+//  * not short names: "quay.io/image", "localhost/image:tag",
+//                     "server.org:5000/lib/image", "image@sha256:..."
+func IsShortName(input string) (bool, error) {
+	isShort, _, err := isShortName(input)
+	return isShort, err
+}
+
+// isShortName implements IsShortName but also returns the parsed
+// reference.Reference.
+func isShortName(input string) (bool, reference.Reference, error) {
```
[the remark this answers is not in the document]
vrothberg: Very nice!

pkg/shortnames/shortnames.go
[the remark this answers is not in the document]
vrothberg: Changed. Also now everything's a `reference.Named` (will do the same for the other PR).

vrothberg:
Rebased on master.

rhatdan:
LGTM

types/types.go
```diff
@@ -486,6 +486,36 @@ func NewOptionalBool(b bool) OptionalBool {
...
+const (
+	ShortNameModeInvalid ShortNameMode = iota
+	// Use all configured unqualified-search registries without prompting
+	// the user.
+	ShortNameModeDisabled
+	// If stdout and stdin are a TTY, prompt the user to select a configured
+	// unqualified-search registry. Otherwise, use all configured
+	// unqualified-search registries.
+	//
+	// Note that if only one unqualified-serach registry is set, it will be
```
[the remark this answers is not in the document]
mtrmac: Still two instances.

types/types.go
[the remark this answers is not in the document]
mtrmac: Works for me.

pkg/shortnames/shortnames_test.go
```diff
@@ -0,0 +1,527 @@
...
+	// Fully-qualified input will be returned as is.
+	resolved, err = Resolve(sys, "quay.io/repo/fedora")
+	require.NoError(t, err)
+	require.NotNil(t, resolved)
+	require.Len(t, resolved.PullCandidates, 1)
+	assert.Equal(t, "quay.io/repo/fedora:latest", resolved.PullCandidates[0].Value.String())
+	assert.False(t, resolved.PullCandidates[0].record)
+}
+
+func toNamed(input string, trim bool) (reference.Named, error) {
```
mtrmac, line 171: This could be `toNamed(t *testing.T, …)` and abort on failure, saving extra checks in callers.

pkg/shortnames/shortnames.go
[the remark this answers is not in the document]
mtrmac: I kind of think there shouldn’t be any such places, but, fine.

types/types.go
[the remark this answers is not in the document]
vrothberg: I'll fix it with the rebase after https://github.com/containers/image/pull/1077 is merged.

pkg/shortnames/shortnames.go
```diff
@@ -0,0 +1,458 @@
...
+
+	// Strip off tag/digest if present.
+	normalized = reference.TrimNamed(named)
+
+	return
+}
+
+// Add records the specified name-value pair as a new short-name alias to the
+// user-specific aliases.conf.  It may override an existing alias for `name`.
+func Add(ctx *types.SystemContext, name reference.Named, value reference.Named) error {
```
mtrmac, line 84: Having a `value reference.Named` is nice; using `reference.Named` for the alias loses the earlier-mentioned separation of short names vs. normalized names via the type system.

We didn’t get to a full type-system separation, still…

Could the alias name be a `string` again, please?

Same for `Remove`.

vrothberg: done

pkg/shortnames/shortnames_test.go
vrothberg: done

mtrmac (approved):
LGTM. Thanks!

Let’s merge #1077 and then this?

vrothberg:
> LGTM. Thanks!
> 
> Let’s merge #1077 and then this?

Yes, that sounds right :+1:

rhatdan:
#1077 Has merged

vrothberg:
> #1077 Has merged

Rebased. Finger's crossed that Travis is scheduling quicker today. Then we can do the vendor :man_dancing: :woman_dancing:

vrothberg:
We're green and good to go. Thanks everyone!

PR #1063, opened 2020-10-20
Add support for downloading of foreign layers to OCI destinations

wagoodman:
This partially addresses https://github.com/containers/skopeo/issues/545 by adding a configurable option to allow/disallow foreign layer URLs to the OCI layout and archive destinations. When `downloadForeignLayers` is set to `true`, then `AcceptsForeignLayerURLs` returns false, hinting to the [`copyLayerHelper` to explicitly copy the layer blob](https://github.com/containers/image/blob/v5.6.0/copy/copy.go#L816).

(When/If this is merged, I'll put in the pairing skopeo PR)


rhatdan:
LGTM
@vrothberg @mtrmac PTAL

TomSweeneyRedHat:
LGTM, but want other's head nods.
Needs a rebase regardless @wagoodman

wagoodman:
@TomSweeneyRedHat @rhatdan thanks! Let me know if there is anything else that is needed for this PR.

rhatdan:
@mtrmac @vrothberg Need one of you to approve.

wagoodman:
If this is a matter of validating that this is the right direction for the upcoming skopeo change, I can supply a skopeo branch that uses these changes if it helps?

mtrmac (commented):
Thanks.

This is not obviously OCI-specific; it might be desirable to make a full copy of foreign layers when copying to a `docker://` registry. So I think this should be an option in `c/image/copy.Options`.

Cc: @vrothberg for another pair of eyes.

vrothberg:
> This is not obviously OCI-specific; it might be desirable to make a full copy of foreign layers when copying to a `docker://` registry. So I think this should be an option in `c/image/copy.Options`.

I concur.  The spec allows for (opting into) copying foreign layers and that's actually something I strongly desire for source images.

rhatdan:
@wagoodman Are you still working on this?

wagoodman:
I plan to, yes... sorry for the radio silence! We just welcomed a new baby boy into our family, so I haven't been able to work on anything for a couple weeks. I think I'll be back in action in about 2 more weeks.

rhatdan:
Awesome, congratulations.  I am having a second grand child in the next couple of weeks.  Although I don't think Red Hat will give me any leave.  :^)

wagoodman:
@rhatdan thanks! And congrats to you as well!

Ok, it's been a while, but I'm back at this... @mtrmac @vrothberg  I see why it would be nice to make this a bit more generic and allow for even copying foreign layer contents to a registry. **This has the ramification of needing to mutate the mediaType to remove the `nondistributable` indication** --is this something that we want to do? 

For example, if we have an OCI layout directory locally with the blob already downloaded and we want to copy the image to a registry (including the layer blob for the foreign layer): 
```
# note the --keep-foreign-layer-contents does not exist on skopeo but would use the downloadForeignLayers option from this PR
skopeo copy oci:some-image-dir-with-foreign-blob docker://user/image:latest --keep-foreign-layer-contents
```
...I think a docker registry would (and should) deny this. The [OCI spec](https://github.com/opencontainers/image-spec/blob/master/layer.md#non-distributable-layers) states that `Implementations SHOULD NOT upload layers tagged with this ["nondistributable"] media type`, which indirectly implies that registries should really not accept a blob for a layer with a mediaType of "...nondistributable...".

It seems like there are two use cases:
1. The download case: I want to download the blob for a foreign layer and leave the mediaType intact (OCI spec mentions this to be fine).
2. The upload case: I want to upload the blob for a layer that was downloaded from a foreign URL reference, therefore before upload I need to remove the URL reference from the manifest, change the mediaType for that layer to NOT indicate that it is "nondistributable", recompute dependent data (e.g. manifest blob, etc.), then upload the blob to the registry.

Should this turn into two options on `copy.Options`?...
- `DownloadForeignLayers`: if a foreign layer exists in the source image, download the contents. 
- `TranslateForeignLayers`:  if a foreign layer exists in the source image, download the contents and mutate the mediaType to reflect that it is no longer a foreign layer.

vrothberg:
> ...I think a docker registry would (and should) deny this. The [OCI spec](https://github.com/opencontainers/image-spec/blob/master/layer.md#non-distributable-layers) states that `Implementations SHOULD NOT upload layers tagged with this ["nondistributable"] media type`, which indirectly implies that registries should really not accept a blob for a layer with a mediaType of "...nondistributable...".

[SHOULD NOT](https://tools.ietf.org/html/rfc2119) allows for doing that assuming the use case is understood. "MUST NOT" would be a clear no.  I haven't actually tested whether a registry would accept that but some registries should to support on-premise use case.

AFAIK the intention of this rule is to prevent leaking data unintentionally.  An opt-in case should still be possible.

wagoodman:
Does that mean that this PR should only have `DownloadForeignLayers ` and not the proposed `TranslateForeignLayers` option @vrothberg ?

vrothberg:
> Does that mean that this PR should only have `DownloadForeignLayers ` and not the proposed `TranslateForeignLayers` option @vrothberg ?

Yes.  I think that a `TranslateForeignLayers` is a different change and a possibly different use case.  Sorry for not replying to this questions earlier!

wagoodman:
@vrothberg  no problem! I updated the branch to add a single option `PersistForeignLayerContents` which the `copy.imageCopier` object takes as configuration.

vrothberg (commented):
Thanks!

The file must be `gofmt`ed.  I preferred to name it `DownloadForeignLayers` over `Persist...Contents`.

wagoodman:
@vrothberg renamed + gofmt'd :+1:

vrothberg (commented):
LGTM, thanks!

@rhatdan PTAL

rhatdan:
LGTM

PR #1069, opened 2020-10-27
subdomain matching in policy.json

lsm5:

@mtrmac @vrothberg @ashcrow PTAL

### DIFF
--- a/docker/daemon/daemon_transport_test.go
+++ b/docker/daemon/daemon_transport_test.go
@@ -32,6 +32,8 @@ func TestTransportValidatePolicyConfigurationScope(t *testing.T) {
 		"registry.example.com/ns/stream",
 		"registry.example.com/ns",
 		"registry.example.com",
+		"*.example.com",
+		"*.com",
 		sha256digestHex, // Accept also unqualified hexdigest values, they are in principle possible host names.
 	} {
 		err := Transport.ValidatePolicyConfigurationScope(scope)
@@ -218,6 +220,7 @@ func TestReferencePolicyConfigurationNamespaces(t *testing.T) {
 		"docker.io/library/busybox",
 		"docker.io/library",
 		"docker.io",
+		"*.io",
 	}, ref.PolicyConfigurationNamespaces())
 }
 
--- a/docker/docker_transport_test.go
+++ b/docker/docker_transport_test.go
@@ -30,6 +30,7 @@ func TestTransportValidatePolicyConfigurationScope(t *testing.T) {
 		"docker.io/library/busybox",
 		"docker.io/library",
 		"docker.io",
+		"*.io",
 	} {
 		err := Transport.ValidatePolicyConfigurationScope(scope)
 		assert.NoError(t, err, scope)
@@ -147,6 +148,7 @@ func TestReferencePolicyConfigurationNamespaces(t *testing.T) {
 		"docker.io/library/busybox",
 		"docker.io/library",
 		"docker.io",
+		"*.io",
 	}, ref.PolicyConfigurationNamespaces())
 }
 
--- a/docker/policyconfiguration/naming.go
+++ b/docker/policyconfiguration/naming.go
@@ -52,5 +52,26 @@ func DockerReferenceNamespaces(ref reference.Named) []string {
 		}
 		name = name[:lastSlash]
 	}
+
+	// Strip port number if any, before appending to res slice.
+	// Currently, the most compatible behavior is to return
+	// example.com:8443/ns, example.com:8443, *.com.
+	// If a port number is not specified, the expected behavior would be
+	// example.com/ns, example.com, *.com
+	portNumColon := strings.Index(name, ":")
+	if portNumColon != -1 {
+		name = name[:portNumColon]
+	}
+
+	// Append wildcarded domains to res slice
+	for {
+		firstDot := strings.Index(name, ".")
+		if firstDot == -1 {
+			break
+		}
+		name = name[firstDot+1:]
+
+		res = append(res, "*."+name)
+	}
 	return res
 }
--- a/docker/policyconfiguration/naming_test.go
+++ b/docker/policyconfiguration/naming_test.go
@@ -10,7 +10,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-// TestDockerReference tests DockerReferenceIdentity and DockerReferenceNamespaces simulatenously
+// TestDockerReference tests DockerReferenceIdentity and DockerReferenceNamespaces simultaneously
 // to ensure they are consistent.
 func TestDockerReference(t *testing.T) {
 	sha256Digest := "@sha256:0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef"
@@ -18,18 +18,19 @@ func TestDockerReference(t *testing.T) {
 	// and that DockerReferenceNamespaces starts with the expected value (fullName), i.e. that the two functions are
 	// consistent.
 	for inputName, expectedNS := range map[string][]string{
-		"example.com/ns/repo": {"example.com/ns/repo", "example.com/ns", "example.com"},
-		"example.com/repo":    {"example.com/repo", "example.com"},
+		"example.com/ns/repo": {"example.com/ns/repo", "example.com/ns", "example.com", "*.com"},
+		"example.com/repo":    {"example.com/repo", "example.com", "*.com"},
 		"localhost/ns/repo":   {"localhost/ns/repo", "localhost/ns", "localhost"},
 		// Note that "localhost" is special here: notlocalhost/repo is parsed as docker.io/notlocalhost.repo:
-		"localhost/repo":         {"localhost/repo", "localhost"},
-		"notlocalhost/repo":      {"docker.io/notlocalhost/repo", "docker.io/notlocalhost", "docker.io"},
-		"docker.io/ns/repo":      {"docker.io/ns/repo", "docker.io/ns", "docker.io"},
-		"docker.io/library/repo": {"docker.io/library/repo", "docker.io/library", "docker.io"},
-		"docker.io/repo":         {"docker.io/library/repo", "docker.io/library", "docker.io"},
-		"ns/repo":                {"docker.io/ns/repo", "docker.io/ns", "docker.io"},
-		"library/repo":           {"docker.io/library/repo", "docker.io/library", "docker.io"},
-		"repo":                   {"docker.io/library/repo", "docker.io/library", "docker.io"},
+		"localhost/repo":                       {"localhost/repo", "localhost"},
+		"notlocalhost/repo":                    {"docker.io/notlocalhost/repo", "docker.io/notlocalhost", "docker.io", "*.io"},
+		"docker.io/ns/repo":                    {"docker.io/ns/repo", "docker.io/ns", "docker.io", "*.io"},
+		"docker.io/library/repo":               {"docker.io/library/repo", "docker.io/library", "docker.io", "*.io"},
+		"docker.io/repo":                       {"docker.io/library/repo", "docker.io/library", "docker.io", "*.io"},
+		"ns/repo":                              {"docker.io/ns/repo", "docker.io/ns", "docker.io", "*.io"},
+		"library/repo":                         {"docker.io/library/repo", "docker.io/library", "docker.io", "*.io"},
+		"repo":                                 {"docker.io/library/repo", "docker.io/library", "docker.io", "*.io"},
+		"yet.another.example.com:8443/ns/repo": {"yet.another.example.com:8443/ns/repo", "yet.another.example.com:8443/ns", "yet.another.example.com:8443", "*.another.example.com", "*.example.com", "*.com"},
 	} {
 		for inputSuffix, mappedSuffix := range map[string]string{
 			":tag":       ":tag",
@@ -49,7 +50,13 @@ func TestDockerReference(t *testing.T) {
 			moreSpecific := identity
 			for i := range expectedNS {
 				assert.Equal(t, ns[i], expectedNS[i], fmt.Sprintf("%s item %d", fullInput, i))
-				assert.True(t, strings.HasPrefix(moreSpecific, ns[i]))
+				// Verify that expectedNS is ordered from most specific to least specific
+				if strings.HasPrefix(ns[i], "*.") {
+					// Check for subdomain matches if wildcard present
+					assert.True(t, strings.Contains(moreSpecific, ns[i][1:]))
+				} else {
+					assert.True(t, strings.HasPrefix(moreSpecific, ns[i]))
+				}
 				moreSpecific = ns[i]
 			}
 		}
--- a/docs/containers-policy.json.5.md
+++ b/docs/containers-policy.json.5.md
@@ -65,7 +65,8 @@ The `atomic:` transport refers to images in an Atomic Registry.
 
 Supported scopes use the form _hostname_[`:`_port_][`/`_namespace_[`/`_imagestream_ [`:`_tag_]]],
 i.e. either specifying a complete name of a tagged image, or prefix denoting
-a host/namespace/image stream.
+a host/namespace/image stream or a wildcarded expression for matching all
+subdomains. For wildcarded subdomain matching, `*.example.com` is a valid case, but `example*.*.com` is not.
 
 *Note:* The _hostname_ and _port_ refer to the Docker registry host and port (the one used
 e.g. for `docker pull`), _not_ to the OpenShift API host and port.
@@ -90,7 +91,9 @@ Scopes matching individual images are named Docker references *in the fully expa
 using a tag or digest. For example, `docker.io/library/busybox:latest` (*not* `busybox:latest`).
 
 More general scopes are prefixes of individual-image scopes, and specify a repository (by omitting the tag or digest),
-a repository namespace, or a registry host (by only specifying the host name).
+a repository namespace, or a registry host (by only specifying the host name)
+or a wildcarded expression for matching all subdomains. For wildcarded subdomain
+matching, `*.example.com` is a valid case, but `example*.*.com` is not.
 
 ### `oci:`
 
@@ -253,6 +256,8 @@ selectively allow individual transports and scopes as desired.
             /* Similarly, allow installing the “official” busybox images.  Note how the fully expanded
                form, with the explicit /library/, must be used. */
             "docker.io/library/busybox": [{"type": "insecureAcceptAnything"}]
+            /* Allow installing images from all subdomains */
+            "*.temporary-project.example.com": [{"type": "insecureAcceptAnything"}]
             /* Other docker: images use the global default policy and are rejected */
         },
         "dir": {
--- a/openshift/openshift_transport_test.go
+++ b/openshift/openshift_transport_test.go
@@ -25,6 +25,8 @@ func TestTransportValidatePolicyConfigurationScope(t *testing.T) {
 		"registry.example.com/ns/stream",
 		"registry.example.com/ns",
 		"registry.example.com",
+		"*.example.com",
+		"*.com",
 	} {
 		err := Transport.ValidatePolicyConfigurationScope(scope)
 		assert.NoError(t, err, scope)
@@ -112,6 +114,8 @@ func TestReferencePolicyConfigurationNamespaces(t *testing.T) {
 		"registry.example.com:8443/ns/stream",
 		"registry.example.com:8443/ns",
 		"registry.example.com:8443",
+		"*.example.com",
+		"*.com",
 	}, ref.PolicyConfigurationNamespaces())
 }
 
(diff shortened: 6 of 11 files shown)


docker/policyconfiguration/naming.go
```diff
@@ -43,14 +45,25 @@ func DockerReferenceNamespaces(ref reference.Named) []string {
...
+				continue
+			}
+			name = name[:lastSlash]
+		} else {
+			res = append(res, name)
+			firstDot := strings.Index(name, ".")
+			if firstDot == -1 {
+				break
+			}
+			name = fmt.Sprint("*", name[firstDot-1:])
```
mtrmac, line 65: Given a `foo.bar`, this turns `name` into `*.bar`, i.e. the same number of dots, so the loop never terminates AFAICS. The tests running out of memory are consistent with that.

mtrmac (commented):
Tests are failing (see the Travis output, or during development, run `make` in the checkout).

I’d expect `naming_test.go` to need updating, at the very least.

Given how security-critical this is, tests should aim at 100% code coverage (apart from clearly unreachable code) for the `naming.go` file and for `docker_transport.go`, and at least _some_ “integration” checks in `c/image/signature` that this has the expected effect.

lsm5:
some comments addressed, some yet to be addressed..

docker/policyconfiguration/naming.go
```diff
@@ -43,14 +45,25 @@ func DockerReferenceNamespaces(ref reference.Named) []string {
...
 		}
-		name = name[:lastSlash]
 	}
```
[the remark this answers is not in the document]
ashcrow: Could https://golang.org/pkg/net/url/#Parse be helpful here?

docker/policyconfiguration/naming.go
```diff
@@ -52,5 +52,19 @@ func DockerReferenceNamespaces(ref reference.Named) []string {
 		}
 		name = name[:lastSlash]
 	}
+
+	portNumColon := strings.Index(name, ":")
+	name = name[:portNumColon]
+	res = append(res, name)
+
+	for {
```
ashcrow, line 60:
```suggestion
        // Iterate over the URI splitting the host by dots into an array
	for {
```

docker/policyconfiguration/naming.go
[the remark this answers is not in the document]
lsm5: @ashcrow do you mean rewrite the entire splitting including namespaces/subdomains using url.Parse? Or just for stripping the port number? I used portNumColon for detecting port, but I guess I could use SplitHostPort.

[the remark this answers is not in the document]
lsm5: @ashcrow is this what you meant?

```
-       portNumColon := strings.Index(name, ":")
-       name = name[:portNumColon]
+       HostPort, err := url.Parse(name)
+       if err != nil {
+               panic(err)
+       }
+       name, _, _ = net.SplitHostPort(HostPort.Host)
```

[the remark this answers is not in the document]
mtrmac: Probably not `url.Parse`, because the inputs are not URLs (e.g. no escaping, and not most of the semantics.)

`SplitHostPort` does a fair bit of extra work, which can be ignored for the input that is actually acceptable to `reference.DomainRegexp`, but it also fails on port-less host names; so it would not simplify much.

docker/policyconfiguration/naming.go
```diff
@@ -52,5 +52,20 @@ func DockerReferenceNamespaces(ref reference.Named) []string {
 		}
 		name = name[:lastSlash]
 	}
+
+	portNumColon := strings.Index(name, ":")
+	name = name[:portNumColon]
```
mtrmac, line 57: `[:-1]` (if there is no port) is not valid.

docker/policyconfiguration/naming.go
```diff
@@ -52,5 +52,20 @@ func DockerReferenceNamespaces(ref reference.Named) []string {
...
+	res = append(res, name)
```
mtrmac, line 58: This should happen only if `name` was changed, otherwise there would be a duplicate value.

… Actually, I’m not at all sure that we should allow non-wildcard $host matching $host:port , historically that did not match. We can safely introduce new semantics for the wildcard syntax, but this might break existing users. So, I’m leaning towards just removing this.

docker/policyconfiguration/naming.go
```diff
@@ -52,5 +52,20 @@ func DockerReferenceNamespaces(ref reference.Named) []string {
...
+
+        // Iterate over the URI splitting the host by dots into an array
```
mtrmac, line 60: Well, this is not splitting “the host into an array”.

docker/docker_transport_test.go
```diff
@@ -147,6 +148,7 @@ func TestReferencePolicyConfigurationNamespaces(t *testing.T) {
 		"docker.io/library/busybox",
 		"docker.io/library",
 		"docker.io",
+		"*.io",
```
mtrmac, line 151: docs/containers-policy.json.5.md should eventually be updated.

lsm5:
> From the tests:
> 
> > [f9f19d7](https://github.com/containers/image/commit/f9f19d75cb5cb2358e4fdc9f96f71becf6bcb1ea) - FAIL - does not have a valid DCO
> 
> How did accepting a GitHub suggestion end up with “Co-authored-by: Stephen Milner [stevem@gnulinux.net](mailto:stevem@gnulinux.net)”?

I hit `commit suggestion` and looks like that's what it ended up doing. Anyway, I'll be squashing commits, so the DCO issue will go away.

Still working on other review comments, specially the tests, so I'll update this PR soon. I'll ping you once it's ready for further review and comments.

Thanks

docker/policyconfiguration/naming.go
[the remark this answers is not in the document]
ashcrow: @lsm5 yes, but it was more specifically a question as if it would help in this situation. I defer to @mtrmac's note.

mtrmac:
> How did accepting a GitHub suggestion end up with “Co-authored-by: Stephen Milner [stevem@gnulinux.net](mailto:stevem@gnulinux.net)”?

Oops, this actually does match the original author; I have confused the GitHub handle with someone else. I apologize.

docker/policyconfiguration/naming.go
lsm5: I haven't pushed the change yet, but locally I have:

```
        // Needed for TestReferencePolicyConfigurationNamepsaces
        portNumColon := strings.Index(name, ":")
        if portNumColon != -1 {
                name = name[:portNumColon]
                res = append(res, name)
        }
```

`openshift.TestReferencePolicyConfigurationNamepsaces` uses a port number, and it will fail if I include `*.example.com` and `*.com` but will pass on `*.example.com:8443` and `*.com:8443` which isn't something we want IIUC. So, do we keep this, or do we get rid of port number from the test?

mtrmac: I think the most compatible behavior is to return `registry.example.com:8443/ns/stream`, `registry.example.com:8443/ns`, `registry.example.com:8443,`*.example.com`, `*.com` — and not `registry.example.com`.

I can see the argument for including `registry.example.com` anyway, even if it changes behavior of existing configurations, for more consistency. It’s a tough call.

lsm5: > I think the most compatible behavior is to return `registry.example.com:8443/ns/stream`, `registry.example.com:8443/ns`, `registry.example.com:8443`,`*.example.com`, `*.com` — and not `registry.example.com`.

Hmm, I guess I will go with this for now. Let me know if you later prefer the one below.

> 
> I can see the argument for including `registry.example.com` anyway, even if it changes behavior of existing configurations, for more consistency. It’s a tough call.

lsm5: so if there is no port number mentioned, for example: `registry.example.com/ns/stream`, should the returned array include `registry.example.com`  in that case? or just `[registry.example.com/ns/stream registry.example.com/ns *.example.com *.com]`

mtrmac: Yes; it is included now, so it must be included in the updated version as well.

lsm5:
@mtrmac ping, looks like a few tests are failing on current origin/master with `key "1D8230F6CDB6A06716E414C1DB72F2188BB46CC8" not found`. Or is it just me?

mtrmac:
> @mtrmac ping, looks like a few tests are failing on current origin/master with `key "1D8230F6CDB6A06716E414C1DB72F2188BB46CC8" not found`. Or is it just me?

@lsm5 works for me, I think (At least `go test ./copy ./signature ` works for me, and the Travis tests fail differently ; you didn’t say what is failing).

Maybe try killing gpg-agent and/or logging in again.

lsm5:
> > @mtrmac ping, looks like a few tests are failing on current origin/master with `key "1D8230F6CDB6A06716E414C1DB72F2188BB46CC8" not found`. Or is it just me?
> 
> @lsm5 works for me, I think (At least `go test ./copy ./signature ` works for me, and the Travis tests fail differently ; you didn’t say what is failing).
> 
> Maybe try killing gpg-agent and/or logging in again.

ack, I rebooted and I don't see it anymore. Thanks!

lsm5:
@mtrmac PTAL when you get the chance, tests pass now (but rerunning because of rebase)

docs/containers-policy.json.5.md
```diff
@@ -32,6 +32,7 @@ Policy requirements can be defined for:
   such a most specific scope define namespaces of matching images.
 - A default policy for a single transport, expressed using an empty string as a scope
 - A global default policy.
+- A wildcard entry for all subdomains.
```
mtrmac, line 35: This should go into the “## Supported transports and their scopes” section, it’s not an universal mechanism (the list above basically matches `requirementsForImageRef`)

docs/containers-policy.json.5.md
```diff
@@ -229,6 +230,8 @@ selectively allow individual transports and scopes as desired.
             /* Similarly, allow installing the “official” busybox images.  Note how the fully expanded
                form, with the explicit /library/, must be used. */
             "docker.io/library/busybox": [{"type": "insecureAcceptAnything"}]
+            /* Allow installing rhel7/rhel images from all subdomains of access.redhat.com */
+            "*.access.redhat.com/rhel7/rhel": [{"type": "insecureAcceptAnything"}]
```
mtrmac, line 234: I don’t think this scope would be matched

signature/docker_test.go
```diff
@@ -22,6 +23,8 @@ func TestSignDockerManifest(t *testing.T) {
 
 	// Successful signing
 	signature, err := SignDockerManifest(manifest, TestImageSignatureReference, mech, TestKeyFingerprint)
+	fmt.Println(signature)
+	fmt.Println(err)
```
mtrmac, line 27: Probably drop this

signature/policy_eval.go
```diff
@@ -142,7 +142,7 @@ func (pc *PolicyContext) requirementsForImageRef(ref types.ImageReference) Polic
 			return req
 		}
 
-		// Look for a match of the possible parent namespaces.
+		// Look for a match of the possible parent namespaces and wildcarded subdomains.
```
mtrmac, line 145: As far as the generic code cares, wildcards are just parent PolicyConfigurationNamespaces.

mtrmac (commented):
A quick pass, I’ll need to think a bit more about the added tests vs. what is needed.

lsm5:
@mtrmac PTAL at the changes. Thanks.

lsm5:
@mtrmac ping, PTAL

ashcrow:
The updates look good to me! @mtrmac thoughts?

rhatdan:
Needs a rebase.

docker/daemon/daemon_transport_test.go
```diff
@@ -218,6 +218,7 @@ func TestReferencePolicyConfigurationNamespaces(t *testing.T) {
 		"docker.io/library/busybox",
 		"docker.io/library",
 		"docker.io",
+		"*.io",
```
mtrmac, line 221: Add this to `TestTransportValidatePolicyConfigurationScope` as well, please.

docker/policyconfiguration/naming_test.go
```diff
@@ -49,7 +50,12 @@ func TestDockerReference(t *testing.T) {
 			moreSpecific := identity
 			for i := range expectedNS {
 				assert.Equal(t, ns[i], expectedNS[i], fmt.Sprintf("%s item %d", fullInput, i))
-				assert.True(t, strings.HasPrefix(moreSpecific, ns[i]))
+				// Check for subdomain matches if wildcard present
```
mtrmac, line 53: (Nit: This comment really belongs _inside_ the `if` branch. I’m not sure that it is very necessary either way.

Here, a comment to the effect of “verify that `expectedNS` is ordered from most specific to least specific” might be useful to explain the _purpose_ of the code, and that might sufficiently explain thy the subdomain does what it does.)

docker/policyconfiguration/naming_test.go
```diff
@@ -49,7 +50,12 @@ func TestDockerReference(t *testing.T) {
...
+				if strings.HasPrefix(ns[i], "*.") {
+					assert.True(t, strings.Contains(moreSpecific, ns[i][2:]))
```
mtrmac, line 55: Non-blocking: Wouldn’t `[1:]` be a more precise match here?

docs/containers-policy.json.5.md
```diff
@@ -90,7 +90,8 @@ Scopes matching individual images are named Docker references *in the fully expa
 using a tag or digest. For example, `docker.io/library/busybox:latest` (*not* `busybox:latest`).
 
 More general scopes are prefixes of individual-image scopes, and specify a repository (by omitting the tag or digest),
-a repository namespace, or a registry host (by only specifying the host name).
+a repository namespace, or a registry host (by only specifying the host name)
+    or a wildcarded expression for matching all subdomains.
```
mtrmac, line 94: This needs to be precise enough to allow users to construct that expression (e.g. what the wildcard is), and not suggest that things like `foo*.*.bar` are valid.

Is the extra indentation intentional? It usually has semantic meaning in Markdown.

openshift/openshift_transport_test.go
```diff
@@ -112,6 +112,8 @@ func TestReferencePolicyConfigurationNamespaces(t *testing.T) {
 		"registry.example.com:8443/ns/stream",
 		"registry.example.com:8443/ns",
 		"registry.example.com:8443",
+		"*.example.com",
+		"*.com",
```
mtrmac, line 116: Add this to `TestTransportValidatePolicyConfigurationScope` as well, please.

docker/policyconfiguration/naming.go
```diff
@@ -52,5 +52,24 @@ func DockerReferenceNamespaces(ref reference.Named) []string {
 		}
 		name = name[:lastSlash]
 	}
+
+	// Needed for TestReferencePolicyConfigurationNamepsaces
+	portNumColon := strings.Index(name, ":")
+	if portNumColon != -1 {
+		name = name[:portNumColon]
+		// Do not append port-stripped name to res array
+		//res = append(res, name)
```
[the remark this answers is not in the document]
mtrmac: Now that there is a good comment above, this can just be dropped.

mtrmac (commented):
I’m sorry about taking so long.

Code looks good, clear documentation is the only pressing concern, other than that just a few implementation-private cleanups.

The current set of tests is as far as we can reasonably get with c/image. (The Skopeo integration tests could be extended with a minimal smoke test, but without a DNS infrastructure to manipulate it would be cumbersome to test much more, and anyway the existing set of tests should give us enough confidence that I don’t think testing this in Skopeo is necessary at all.)

lsm5:
@mtrmac thanks a lot for the review and comments. I'll quite likely address these along with fixing test failures (post-rebase) after I'm back from break (or maybe during the break if I'm bored). See you all soon..

lsm5:
@mtrmac PTAL, all comments should be addressed with the latest change. Thanks a lot!

lsm5:
also updated the commit message a bit.

docs/containers-policy.json.5.md
```diff
@@ -90,7 +90,9 @@ Scopes matching individual images are named Docker references *in the fully expa
 using a tag or digest. For example, `docker.io/library/busybox:latest` (*not* `busybox:latest`).
 
 More general scopes are prefixes of individual-image scopes, and specify a repository (by omitting the tag or digest),
-a repository namespace, or a registry host (by only specifying the host name).
+a repository namespace, or a registry host (by only specifying the host name)
+or a wildcarded expression for matching all subdomains. For wildcarded subdomain
+matching, `*.example.com` is a valid case, but `example*.*.com` is not.
```
mtrmac, line 95: `atomic:` above (c/image/openshift) also needs updating. (Possibly also `docker-daemon:`, but that’s currently not documented here at all, so N/A.)

docs/containers-policy.json.5.md
```diff
@@ -253,6 +255,8 @@ selectively allow individual transports and scopes as desired.
             /* Similarly, allow installing the “official” busybox images.  Note how the fully expanded
                form, with the explicit /library/, must be used. */
             "docker.io/library/busybox": [{"type": "insecureAcceptAnything"}]
+            /* Allow installing images from all subdomains of access.redhat.com */
+            "*.access.redhat.com": [{"type": "insecureAcceptAnything"}]
```
mtrmac, line 259: It’s not ideal to recommend insecure use of signed RH images. Something like `*.temporary-project.example.com`? Or maybe show how to require RH signatures for all of `*.redhat.com` (but that could detract from the `*.` point).

mtrmac (commented):
Code LGTM, a few more documentation nits. I should have noticed the `atomic:` one before, I’m sorry.

TomSweeneyRedHat:
LGTM once @mtrmac's comments are addressed.

docs/containers-policy.json.5.md
lsm5: Ack, so I could just copy-paste that paragraph on general scopes as-is to the atomic: section, correct? Or, are there any atomic specific changes required there?

mtrmac: Yes; the two differ slightly in the after-hostname part (and the mandatory `//` prefix) only.

lsm5: pushed a new change, PTAL. Thanks

lsm5: ah, noticed your comment only now. I need to change this, sorry i just copy-pasted it earlier.

lsm5: @mtrmac re-read your comment just now. So, if the difference is only in the prefix and after-hostname, we should be good to go afaict. Let me know, thanks!

docs/containers-policy.json.5.md
```diff
@@ -253,6 +260,8 @@ selectively allow individual transports and scopes as desired.
             /* Similarly, allow installing the “official” busybox images.  Note how the fully expanded
                form, with the explicit /library/, must be used. */
             "docker.io/library/busybox": [{"type": "insecureAcceptAnything"}]
+            /* Allow installing images from all subdomains of access.redhat.com */
```
mtrmac, line 263: Update the comment as well, please.

docs/containers-policy.json.5.md
```diff
@@ -70,6 +70,11 @@ a host/namespace/image stream.
 *Note:* The _hostname_ and _port_ refer to the Docker registry host and port (the one used
 e.g. for `docker pull`), _not_ to the OpenShift API host and port.
 
+More general scopes are prefixes of individual-image scopes, and specify a repository (by omitting the tag or digest),
+a repository namespace, or a registry host (by only specifying the host name)
```
mtrmac, line 74: - The `*note*` paragraph should probably be after, not in the middle of, the scope discussion.
- Hum, the more detailed scopes are “image streams”, not “repositories” — and already described above; so, maybe just continue the “host namespace/image stream” sentence with “, or a wildcarded…”.

docs/containers-policy.json.5.md
lsm5: whoops, done!

docs/containers-policy.json.5.md
lsm5: done

mtrmac (approved):
LGTM. Thanks!

rhatdan:
Is this something we should get into podman 3.0?  Or is it mainly for CRI-O?

mtrmac:
Primary is OpenShift cluster run-time, i.e. CRI-O. Podman is IIRC involved only in parts of the installation process, and I’m not immediately sure whether this is exposed there and will need to be supported.

Either way, I don’t think it is a reason to delay Podman: there will need to be OpenShift updates to accept (and document) the wildcard syntax, it’s not enabled for supported use just by shipping a new CRI-O.

PR #1072, opened 2020-11-02
Add reader/writer to for oci-archive multi image

QiWang19:
Add read/writer with helpers to allow podman save/load multi oci-archive images.
work with PR: https://github.com/containers/podman/pull/8218
close: https://github.com/containers/podman/issues/4646

### DIFF
--- a/oci/archive/oci_transport.go
+++ b/oci/archive/oci_transport.go
@@ -32,10 +32,12 @@ type ociArchiveTransport struct{}
 
 // ociArchiveReference is an ImageReference for OCI Archive paths
 type ociArchiveReference struct {
-	file         string
-	resolvedFile string
-	image        string
-	sourceIndex  int
+	file             string
+	resolvedFile     string
+	image            string
+	sourceIndex      int
+	archiveReaderRef *tempDirOCIRef
+	archiveWriterRef *tempDirOCIRef
 }
 
 func (t ociArchiveTransport) Name() string {
@@ -55,20 +57,24 @@ func (t ociArchiveTransport) ValidatePolicyConfigurationScope(scope string) erro
 
 // ParseReference converts a string, which should not start with the ImageTransport.Name prefix, into an OCI ImageReference.
 func ParseReference(reference string) (types.ImageReference, error) {
-	file, image := internal.SplitPathAndImage(reference)
-	image, index, err := internal.ParseOCIReferenceName(image)
+	file, image, index, err := internal.ParseReferenceIntoElements(reference)
 	if err != nil {
 		return nil, err
 	}
-	return newReference(file, image, index)
+	return newReference(file, image, index, nil, nil)
 }
 
 // NewReference returns an OCI reference for a file and an image.
 func NewReference(file, image string) (types.ImageReference, error) {
-	return newReference(file, image, -1)
+	return newReference(file, image, -1, nil, nil)
 }
 
-func newReference(file, image string, sourceIndex int) (types.ImageReference, error) {
+// NewIndexReference returns an OCI reference for a file and sourecIndex points to the image.
+func NewIndexReference(file string, sourceIndex int) (types.ImageReference, error) {
+	return newReference(file, "", sourceIndex, nil, nil)
+}
+
+func newReference(file, image string, sourceIndex int, archiveReaderRef, archiveWriterRef *tempDirOCIRef) (types.ImageReference, error) {
 	resolved, err := explicitfilepath.ResolvePathToFullyExplicit(file)
 	if err != nil {
 		return nil, err
@@ -85,7 +91,10 @@ func newReference(file, image string, sourceIndex int) (types.ImageReference, er
 	if sourceIndex != -1 && sourceIndex < 0 {
 		return nil, errors.Errorf("Invalid oci archive: reference: index @%d must not be negative", sourceIndex)
 	}
-	return ociArchiveReference{file: file, resolvedFile: resolved, image: image, sourceIndex: sourceIndex}, nil
+	if sourceIndex != -1 && image != "" {
+		return nil, errors.Errorf("Can not set image %s and index @%d at same time", image, sourceIndex)
+	}
+	return ociArchiveReference{file: file, resolvedFile: resolved, image: image, sourceIndex: sourceIndex, archiveReaderRef: archiveReaderRef, archiveWriterRef: archiveWriterRef}, nil
 }
 
 func (ref ociArchiveReference) Transport() types.ImageTransport {
@@ -138,7 +147,7 @@ func (ref ociArchiveReference) PolicyConfigurationNamespaces() []string {
 // verify that UnparsedImage, and convert it into a real Image via image.FromUnparsedImage.
 // WARNING: This may not do the right thing for a manifest list, see image.FromSource for details.
 func (ref ociArchiveReference) NewImage(ctx context.Context, sys *types.SystemContext) (types.ImageCloser, error) {
-	src, err := newImageSource(ctx, sys, ref)
+	src, err := newImageSource(ctx, sys, ref, ref.archiveReaderRef)
 	if err != nil {
 		return nil, err
 	}
@@ -148,13 +157,13 @@ func (ref ociArchiveReference) NewImage(ctx context.Context, sys *types.SystemCo
 // NewImageSource returns a types.ImageSource for this reference.
 // The caller must call .Close() on the returned ImageSource.
 func (ref ociArchiveReference) NewImageSource(ctx context.Context, sys *types.SystemContext) (types.ImageSource, error) {
-	return newImageSource(ctx, sys, ref)
+	return newImageSource(ctx, sys, ref, ref.archiveReaderRef)
 }
 
 // NewImageDestination returns a types.ImageDestination for this reference.
 // The caller must call .Close() on the returned ImageDestination.
 func (ref ociArchiveReference) NewImageDestination(ctx context.Context, sys *types.SystemContext) (types.ImageDestination, error) {
-	return newImageDestination(ctx, sys, ref)
+	return newImageDestination(ctx, sys, ref, ref.archiveWriterRef)
 }
 
 // DeleteImage deletes the named image from the registry, if supported.
@@ -180,9 +189,15 @@ func createOCIRef(sys *types.SystemContext, image string, sourceIndex int) (temp
 	if err != nil {
 		return tempDirOCIRef{}, errors.Wrapf(err, "error creating temp directory")
 	}
-	ociRef, err := ocilayout.NewReferenceWithIndex(dir, image, sourceIndex)
-	if err != nil {
-		return tempDirOCIRef{}, err
+	var ociRef types.ImageReference
+	if sourceIndex > -1 {
+		if ociRef, err = ocilayout.NewIndexReference(dir, sourceIndex); err != nil {
+			return tempDirOCIRef{}, err
+		}
+	} else {
+		if ociRef, err = ocilayout.NewReference(dir, image); err != nil {
+			return tempDirOCIRef{}, err
+		}
 	}
 
 	tempDirRef := tempDirOCIRef{tempDirectory: dir, ociRefExtracted: ociRef}
--- a/oci/archive/oci_transport.go
+++ b/oci/archive/oci_transport.go
@@ -35,6 +35,7 @@ type ociArchiveReference struct {
 	file         string
 	resolvedFile string
 	image        string
+	sourceIndex  int
 }
 
 func (t ociArchiveTransport) Name() string {
@@ -55,11 +56,19 @@ func (t ociArchiveTransport) ValidatePolicyConfigurationScope(scope string) erro
 // ParseReference converts a string, which should not start with the ImageTransport.Name prefix, into an OCI ImageReference.
 func ParseReference(reference string) (types.ImageReference, error) {
 	file, image := internal.SplitPathAndImage(reference)
-	return NewReference(file, image)
+	image, index, err := internal.ParseOCIReferenceName(image)
+	if err != nil {
+		return nil, err
+	}
+	return newReference(file, image, index)
 }
 
-// NewReference returns an OCI reference for a file and a image.
+// NewReference returns an OCI reference for a file and an image.
 func NewReference(file, image string) (types.ImageReference, error) {
+	return newReference(file, image, -1)
+}
+
+func newReference(file, image string, sourceIndex int) (types.ImageReference, error) {
 	resolved, err := explicitfilepath.ResolvePathToFullyExplicit(file)
 	if err != nil {
 		return nil, err
@@ -73,7 +82,10 @@ func NewReference(file, image string) (types.ImageReference, error) {
 		return nil, err
 	}
 
-	return ociArchiveReference{file: file, resolvedFile: resolved, image: image}, nil
+	if sourceIndex != -1 && sourceIndex < 0 {
+		return nil, errors.Errorf("Invalid oci archive: reference: index @%d must not be negative", sourceIndex)
+	}
+	return ociArchiveReference{file: file, resolvedFile: resolved, image: image, sourceIndex: sourceIndex}, nil
 }
 
 func (ref ociArchiveReference) Transport() types.ImageTransport {
@@ -83,7 +95,10 @@ func (ref ociArchiveReference) Transport() types.ImageTransport {
 // StringWithinTransport returns a string representation of the reference, which MUST be such that
 // reference.Transport().ParseReference(reference.StringWithinTransport()) returns an equivalent reference.
 func (ref ociArchiveReference) StringWithinTransport() string {
-	return fmt.Sprintf("%s:%s", ref.file, ref.image)
+	if ref.sourceIndex == -1 {
+		return fmt.Sprintf("%s:%s", ref.file, ref.image)
+	}
+	return fmt.Sprintf("%s:@%d", ref.file, ref.sourceIndex)
 }
 
 // DockerReference returns a Docker reference associated with this reference
@@ -160,12 +175,12 @@ func (t *tempDirOCIRef) deleteTempDir() error {
 
 // createOCIRef creates the oci reference of the image
 // If SystemContext.BigFilesTemporaryDir not "", overrides the temporary directory to use for storing big files
-func createOCIRef(sys *types.SystemContext, image string) (tempDirOCIRef, error) {
+func createOCIRef(sys *types.SystemContext, image string, sourceIndex int) (tempDirOCIRef, error) {
 	dir, err := ioutil.TempDir(tmpdir.TemporaryDirectoryForBigFiles(sys), "oci")
 	if err != nil {
 		return tempDirOCIRef{}, errors.Wrapf(err, "error creating temp directory")
 	}
-	ociRef, err := ocilayout.NewReference(dir, image)
+	ociRef, err := ocilayout.NewReferenceWithIndex(dir, image, sourceIndex)
 	if err != nil {
 		return tempDirOCIRef{}, err
 	}
@@ -176,7 +191,7 @@ func createOCIRef(sys *types.SystemContext, image string) (tempDirOCIRef, error)
 
 // creates the temporary directory and copies the tarred content to it
 func createUntarTempDir(sys *types.SystemContext, ref ociArchiveReference) (tempDirOCIRef, error) {
-	tempDirRef, err := createOCIRef(sys, ref.image)
+	tempDirRef, err := createOCIRef(sys, ref.image, ref.sourceIndex)
 	if err != nil {
 		return tempDirOCIRef{}, errors.Wrap(err, "error creating oci reference")
 	}
--- a/oci/archive/reader.go
+++ b/oci/archive/reader.go
@@ -1,0 +1,98 @@
+package archive
+
+import (
+	"context"
+	"encoding/json"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/containers/image/v5/internal/tmpdir"
+	"github.com/containers/image/v5/oci/layout"
+	"github.com/containers/image/v5/types"
+	"github.com/containers/storage/pkg/archive"
+	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/pkg/errors"
+)
+
+// Reader keeps the temp directory the oci archive will be untarred to and the manifest of the images
+type Reader struct {
+	manifest      *imgspecv1.Index
+	tempDirectory string
+}
+
+// NewReader creates the temp directory that keeps the untarred archive from src.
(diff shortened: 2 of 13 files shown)


oci/archive/reader.go
```diff
@@ -0,0 +1,93 @@
...
+	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/pkg/errors"
+)
+
+type reader struct {
+	manifest *imgspecv1.Index
+	tempDirOCIRef
+}
+
+func CreateUntarTempDirReader(src string, sys *types.SystemContext) (*reader, error) {
```
vrothberg, line 22: Docs are missing

oci/archive/reader.go
```diff
@@ -0,0 +1,93 @@
...
+	// TODO: This can take quite some time, and should ideally be cancellable using a context.Context.
```
vrothberg, line 23: Note sure what "this" refers to and how we could cancel. But if we want to do add cancellation we should change the signature of the function and add a `context.Context`. This way we don't break the API in the future.

oci/archive/reader.go
```diff
@@ -0,0 +1,93 @@
...
+
+// List returns a list of ReferenceWrapper for images in the reader.
+// the ImageReferences of the ReferenceWrapper are valid only until the Reader is closed.
+func (r *reader) List() ([]ReferenceWrapper, error) {
+	res := []ReferenceWrapper{}
+	for _, md := range r.manifest.Manifests {
+		if md.MediaType != imgspecv1.MediaTypeImageManifest && md.MediaType != imgspecv1.MediaTypeImageIndex {
+			continue
+		}
+		refName, ok := md.Annotations[imgspecv1.AnnotationRefName]
```
vrothberg, line 73: I think we need to support more than one name as we support with docker-archive.

Doing a `podman save -o foo.tar image1-a image1-b image1-c image2` with image1-* all referring to the same image, the names will be preserved.

oci/archive/reader.go
```diff
@@ -0,0 +1,93 @@
...
+
+// ReferenceWrapper includes the reference for the image and image dest name
+type ReferenceWrapper struct {
+	DestName       string
+	ImageReference types.ImageReference
+}
+
+// List returns a list of ReferenceWrapper for images in the reader.
+// the ImageReferences of the ReferenceWrapper are valid only until the Reader is closed.
+func (r *reader) List() ([]ReferenceWrapper, error) {
```
vrothberg, line 67: Looking at `docker-archive`, I think we should not create a custom `ReferenceWrapper` but return a `[][]types.ImageReference`. With the first array determining the amount of images and the second one the "names".

`docker-archive` supports loading by index and by name and I think that's what we should do here as well to be consistent. The docker-archive parts are in `docker/archive/reader.go`.

oci/archive/reader.go
QiWang19: Will look into add the `context.Context`.  I grab the same code from oci_transport.go to untar the archive. The comments are from https://github.com/containers/image/blob/f1a6f9b50d0286bf208935d4b9ab48117294ac31/oci/archive/oci_transport.go#L185

oci/archive/reader.go
QiWang19: @vrothberg Could you elaborate on what's loading by index and name?

vrothberg: A `poadman load -i` always expects a file, so we can't use the transport syntax. But we can individually pull/extract images from an archive. To extract an image (note we can do this with `podman pull` as well), we need to specify which one. We can specify the image either via a docker-reference (without digest) or by it's index in the list of manifests.

Examples are the man page (https://github.com/containers/image/blob/master/docs/containers-transports.5.md#docker-archivepathdocker-referencesource-index) or the tests in podman that exercise this code (https://github.com/containers/podman/commit/7fea46752cbf#diff-36d4ee0b4d8927b38ddbff8a5648df51ebf0e9289fcb0d37e0dedc2ffa3ee161R173).

oci/archive/reader.go
mtrmac: (Drive-by comment, I haven’t read the full PR.) These TODO comments are sprinkled around the code base from the time we’ve added `context.Context` to the transports, but we didn’t actually implement cancellation in many places because `io.Copy` and local file operations are not trivially cancellable in Go.

So it would be, I guess, consistent, to leave that s a TODO instead of implementing — but I do agree that worrying about making it eventually _possible_ through the API by at least passing (and ignoring) a `context.Context` makes sense.

oci/archive/reader.go
QiWang19: The names of the same image are preserved as `imgspecv1.AnnotationRefName`("org.opencontainers.image.ref.name") field of different entries. And I changed the `ReferenceWrapper` https://github.com/containers/image/pull/1072#discussion_r516612025, use a map to keep the name and ImageReferences.

QiWang19:
@vrothberg @mtrmac PTAL

oci/archive/oci_transport.go
```diff
@@ -69,11 +71,27 @@ func NewReference(file, image string) (types.ImageReference, error) {
...
 	}
 
-	if err := internal.ValidateImageName(image); err != nil {
-		return nil, err
+	var img string
+	index := -1
+	if strings.HasPrefix(image, "@") {
+		idx, err := strconv.Atoi(image[1:])
+		if err != nil {
+			return nil, errors.Wrapf(err, "Invalid source index %s", image)
```
vrothberg, line 79:
```suggestion
			return nil, errors.Wrapf(err, "Invalid source index @%s: not an integer", image[1:])
```

oci/layout/oci_transport.go
```diff
@@ -63,20 +63,21 @@ type ociReference struct {
...
-	return NewReference(dir, image)
+	return NewReference(dir, image, -1)
 }
 
 // NewReference returns an OCI reference for a directory and a image.
 //
 // We do not expose an API supplying the resolvedDir; we could, but recomputing it
 // is generally cheap enough that we prefer being confident about the properties of resolvedDir.
-func NewReference(dir, image string) (types.ImageReference, error) {
+func NewReference(dir, image string, sourceIndex int) (types.ImageReference, error) {
```
vrothberg, line 80: That's technically breaking the API. Can we let this default to `0`?

vrothberg: Maybe a new `NewReferenceWithIndex`?

vrothberg (commented):
Nice work, @QiWang19

oci/archive/writer.go
```diff
@@ -0,0 +1,45 @@
+package archive
+
+import (
+	"io/ioutil"
+	"os"
+
+	"github.com/containers/image/v5/internal/tmpdir"
+
```
rhatdan, line 8: Should not have blank line.

oci/layout/oci_transport.go
```diff
@@ -72,6 +73,27 @@ func ParseReference(reference string) (types.ImageReference, error) {
 	return NewReference(dir, image)
 }
 
+// NewReferenceWithIndex returns an OCI reference for a directory and a image， sourceIndex points to an image.
+//
+// We do not expose an API supplying the resolvedDir; we could, but recomputing it
+// is generally cheap enough that we prefer being confident about the properties of resolvedDir.
+func NewReferenceWithIndex(dir, image string, sourceIndex int) (types.ImageReference, error) {
```
rhatdan, line 80: NewReference() Should be calling this function, rather then duplicating the code.

rhatdan (approved):

QiWang19:
@vrothberg @mtrmac  PTAL

oci/archive/reader.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+type reader struct {
+	manifest *imgspecv1.Index
+	tempDirOCIRef
+}
+
+// CreateUntarTempDirReader creates the temp directory that keeps the untarred archive from src.
+// The caller should call .Close() on the returned object.
+func CreateUntarTempDirReader(ctx context.Context, src string, sys *types.SystemContext) (*reader, error) {
```
mtrmac, line 25: Why not just a `NewReader`, to be consistent with docker/archive?

The `reader` type should probably be public (with all members private), so that the caller can declare a temporary variable or a parameter holding the returned value.

oci/archive/reader.go
```diff
@@ -0,0 +1,90 @@
...
+// List returns a (name, reference) map for images in the reader
+// the name will be used to determin reference name of the dest image.
+// the ImageReferences are valid only until the Reader is closed.
+func (r *reader) List() (map[string]types.ImageReference, error) {
+	res := make(map[string]types.ImageReference)
+	for _, md := range r.manifest.Manifests {
+		if md.MediaType != imgspecv1.MediaTypeImageManifest && md.MediaType != imgspecv1.MediaTypeImageIndex {
+			continue
+		}
+		refName, ok := md.Annotations[imgspecv1.AnnotationRefName]
```
mtrmac, line 71: Shouldn’t all unnamed images still be accessible? Otherwise the @_int_ syntax has little purpose.

oci/archive/reader.go
```diff
@@ -0,0 +1,90 @@
...
+		if !ok {
+			continue
+		}
+		ref, err := layout.NewReference(r.tempDirOCIRef.tempDirectory, refName)
+		if err != nil {
+			continue
```
mtrmac, line 77: Why is it OK to ignore the error? Either fix that or add a comment, please.

oci/archive/reader.go
```diff
@@ -0,0 +1,90 @@
...
+		}
+		if _, ok := res[refName]; ok {
+			return nil, errors.Errorf("image descriptor confliction")
```
mtrmac, line 80: s/confliction/conflict/, and it would be nice to include the name.

oci/archive/reader.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/containers/image/v5/oci/layout"
+	"github.com/containers/image/v5/types"
+	"github.com/containers/storage/pkg/archive"
+	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/pkg/errors"
+)
+
+type reader struct {
+	manifest *imgspecv1.Index
+	tempDirOCIRef
```
mtrmac, line 20: I’d expect `tempDirOCIRef` to completely disappear, and for its current users to be implemented in terms of the `reader`.

oci/archive/oci_transport.go
```diff
@@ -69,11 +71,27 @@ func NewReference(file, image string) (types.ImageReference, error) {
 		return nil, err
 	}
 
-	if err := internal.ValidateImageName(image); err != nil {
-		return nil, err
+	var img string
+	index := -1
+	if strings.HasPrefix(image, "@") {
```
mtrmac, line 76: The idea of `NewReference` is that the inputs are already well-typed; it shouldn’t be doing string parsing.

Please move the string parsing into ParseReference (and then further mostly into the `internal` subpackage), and add a public a constructor that allows creating the index references from raw data.

---

It would also be nice for the oci/archive and oci/layout reference API, and a fair bit of the implementation, to remain the same (or shared in `internal`), although the two will necessarily diverge a bit with the support for long-term readers/writers.

oci/archive/reader.go
```diff
@@ -0,0 +1,90 @@
...
+	res := make(map[string]types.ImageReference)
+	for _, md := range r.manifest.Manifests {
+		if md.MediaType != imgspecv1.MediaTypeImageManifest && md.MediaType != imgspecv1.MediaTypeImageIndex {
+			continue
+		}
+		refName, ok := md.Annotations[imgspecv1.AnnotationRefName]
+		if !ok {
+			continue
+		}
+		ref, err := layout.NewReference(r.tempDirOCIRef.tempDirectory, refName)
```
mtrmac, line 75: I don’t think it’s quite that simple — this creates references with the temporary path , which would show up in the UI (and, if anyone ever used it, in `PolicyConfigurationIdentity`).

See how `docker.archiveReference` holds an optional `archiveReader`.

oci/archive/writer.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// Close deletes temporary files associated with the Writer
+func (w *Writer) Close() error {
+	return os.RemoveAll(w.TempDir)
+}
+
+// TarDirectory converts the directory from Writer and saves it to file
+func (w *Writer) TarDirectory(file string) error {
+	dst, err := explicitfilepath.ResolvePathToFullyExplicit(file)
```
mtrmac, line 39: Why is this useful _here_? It would make sense for creating the individual constituent references, but at this point it’s just passed to `os.Create`.

oci/archive/writer.go
```diff
@@ -0,0 +1,44 @@
...
+	if err != nil {
+		return err
+	}
+	return tarDirectory(w.TempDir, dst)
```
mtrmac, line 43: Just reading this bit of code, it’s not very clear what the `Writer` does and why it needs to exist — if the caller needs to manage the directory, manually create references for the contents (and those will, see `reader` above, leak the temporary path in the UI), and then create an archive, the caller might almost as well do this without this layer.

---

Probably this should be more directly modeled based on `docker/archive.Writer`, notably with the `NewReference` method and the way it creates archive references with a hidden writer association.

oci/layout/oci_transport.go
```diff
@@ -194,10 +203,14 @@ func (ref ociReference) getManifestDescriptor() (imgspecv1.Descriptor, error) {
 		}
 	} else {
 		// if image specified, look through all manifests for a match
-		for _, md := range index.Manifests {
+		for i, md := range index.Manifests {
 			if md.MediaType != imgspecv1.MediaTypeImageManifest && md.MediaType != imgspecv1.MediaTypeImageIndex {
 				continue
 			}
+			if ref.sourceIndex == i {
```
mtrmac, line 210: Why not special-case this and index `index.Manifests[ref.sourceIndex]` directly? That would also allow returning a detailed error (“index too large, only %d entries available”) instead of just a “not found”.

oci/layout/oci_transport.go
```diff
@@ -72,11 +73,11 @@ func ParseReference(reference string) (types.ImageReference, error) {
 	return NewReference(dir, image)
 }
 
-// NewReference returns an OCI reference for a directory and a image.
+// NewReferenceWithIndex returns an OCI reference for a directory and a image， sourceIndex points to an image.
 //
 // We do not expose an API supplying the resolvedDir; we could, but recomputing it
 // is generally cheap enough that we prefer being confident about the properties of resolvedDir.
-func NewReference(dir, image string) (types.ImageReference, error) {
+func NewReferenceWithIndex(dir, image string, sourceIndex int) (types.ImageReference, error) {
```
mtrmac, line 80: Please document the `sourceIndex == -1` case, and enforce that `sourceIndex` and `image` can’t be set simultaneously.

---

Actually docker/archive has separate `NewReference` and `NewIndexReference`, making it structurally impossible to set both at the same time, and allowing validation that refuses a negative index in the “index” case (OTOH that validation doesn’t exist).

oci/layout/oci_transport.go
```diff
@@ -63,7 +63,8 @@ type ociReference struct {
 	resolvedDir string // Absolute path with no symlinks, at least at the time of its creation. Primarily used for policy namespaces.
 	// If image=="", it means the "only image" in the index.json is used in the case it is a source
 	// for destinations, the image name annotation "image.ref.name" is not added to the index.json
-	image string
+	image       string
+	sourceIndex int
```
mtrmac, line 67: Please add documentation, on both the special-case values and the relationship with `image`.

oci/layout/oci_transport.go
```diff
@@ -72,11 +73,11 @@ func ParseReference(reference string) (types.ImageReference, error) {
 	return NewReference(dir, image)
 }
```
mtrmac, line 74: Because `StringWithinTransport` will (see elsewhere) need to contain the index, `ParseReference` in oci/layout also needs to support a string form of the index syntax (ideally sharing much of the code with oci/archive; and then docs/containers-transports.5.md will need to be updated).

oci/archive/oci_transport.go
```diff
@@ -35,6 +36,7 @@ type ociArchiveReference struct {
 	file         string
 	resolvedFile string
 	image        string
+	sourceIndex  int
```
mtrmac, line 39: This _must_ show up in `StringWithinTransport`, and the tests should have appropriate coverage.

oci/layout/oci_transport.go
mtrmac, line 67: This _must_ show up in `StringWithinTransport`, and the tests should have appropriate coverage.

mtrmac (commented):
Thanks, this is a definitely a welcome feature.

oci/layout/oci_transport.go
mtrmac, line 67: I don’t think it makes sense to use a `sourceIndex` for destinations (or does it?), so `NewImageDestination` should reject such references (in both transports).

oci/archive/reader.go
QiWang19: Could you elaborate why the `archiveReader` is needed?

mtrmac: With this implementation,
```go
ref := /* something like */archive.CreateUntarTempDirReader(… "/home/mitr/my-precious-archive" …).List()["my-tag"]
printf("Now pulling %s", transports.ImageName(ref) // prints Now pulling oci-layout:/var/tmp/ociRANDOM:my-tag
```
and similarly that identity is probably going to show up in error messages. Relating that back to the `my-precious-archive` path is non-obvious.

`PolicyConfigurationIdentity` is theoretically the most severe impact (bypassing/breaking `policy.json` because both the transport and the path is different), in practice probably never hit; the error message aspect is more important.

---

The `archiveReader` approach has another potentially useful (but admittedly not strictly _necessary_) property, that the references created by `reader.List` for accessing individual parts of a compressed archive are, as far as the string syntax shows, exactly the same as the `oci-archive:$file:$tag` references the user can write manually; the `reader` object becomes just a way to optimize accessing many such references with a single tar extraction overhead.

QiWang19:
@mtrmac PTAL

QiWang19:
@mtrmac PTAL

QiWang19:
@mtrmac PTAL.

oci/archive/writer.go
```diff
@@ -0,0 +1,54 @@
...
+		tempDir: dir,
+		path:    dst,
+	}
+	return ociWriter, nil
+}
+
+// NewReference returns an ImageReference that allows adding an image to Writer,
+// with an optional image name
+func (w *Writer) NewReference(name string) (types.ImageReference, error) {
+	return layout.NewReference(w.tempDir, name)
```
mtrmac, line 43: Still exposes `w.tempDir` in user-visible references.

oci/layout/oci_transport.go
```diff
@@ -64,19 +64,26 @@ type ociReference struct {
...
 
-// NewReference returns an OCI reference for a directory and a image.
+// NewReferenceWithIndex returns an OCI reference for a directory and a image， sourceIndex points to an image.
 //
+// If sourceIndex==-1, the index will not be valid to point out the source image, only image will be used.
 // We do not expose an API supplying the resolvedDir; we could, but recomputing it
 // is generally cheap enough that we prefer being confident about the properties of resolvedDir.
-func NewReference(dir, image string) (types.ImageReference, error) {
+func NewReferenceWithIndex(dir, image string, sourceIndex int) (types.ImageReference, error) {
```
mtrmac, line 86: Probably make this a private `newReference`; let external users call either `NewReference` or `NewIndexReference`.

It would be useful for the two transports to be structurally similar in the API and the implementation approach.

oci/layout/oci_transport.go
```diff
@@ -182,8 +209,14 @@ func (ref ociReference) getManifestDescriptor() (imgspecv1.Descriptor, error) {
 	if err != nil {
 		return imgspecv1.Descriptor{}, err
 	}
-
 	var d *imgspecv1.Descriptor
+	if ref.sourceIndex != -1 {
+		if len(index.Manifests) > ref.sourceIndex {
```
mtrmac, line 214: Non-blocking? `ref.sourceIndex < len(index.Manifests)` feels more natural, at least it is consistent with `ref.sourceIndex != -1` (which is not `-1 != ref.sourceIndex`).

oci/archive/writer.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+// Close converts the data about images in the temp directory to the archive and
+// deletes temporary files associated with the Writer
+func (w *Writer) Close() error {
```
mtrmac, line 48: Should this have a `context.Context`, to eventually allow aborting that write? OTOH that would make this not conform to `io.Closer`.

Ugh.

Alternatively, pass a `context.Context` to `NewWriter` would be an option, violating the usual rules of use of `context.Context` (OTOH it wouldn’t be the first time IIRC).

\*shrug\* Maybe we should defer solving that for now, we can always add a `CloseWithContext` or something like that when actually adding a cancellable implementation.

mtrmac (commented):
Just a quick partial read for now I’m afraid, focusing at least on the public API.

I’d still prefer to re-implement `ociArchiveImage{Source,Destination}` in terms of the `Reader`/`Writer`, rather than making `Reader`/`Writer` a completely separate path that creates oci/layout references, per https://github.com/containers/image/pull/1072#discussion_r529782470 .

oci/layout/oci_transport_test.go
```diff
@@ -23,6 +23,18 @@ func TestGetManifestDescriptor(t *testing.T) {
 
 	_, err = imageRef.(ociReference).getManifestDescriptor()
 	assert.EqualError(t, err, ErrMoreThanOneImage.Error())
+
```
mtrmac, line 26: The comment above the function now needs to be moved to the appropriate test case.

mtrmac (commented):
A full review now.

QiWang19:
@mtrmac PTAL

oci/archive/reader.go
```diff
@@ -0,0 +1,98 @@
...
+		err error
+	)
+	for i, md := range r.manifest.Manifests {
+		if md.MediaType != imgspecv1.MediaTypeImageManifest && md.MediaType != imgspecv1.MediaTypeImageIndex {
+			continue
+		}
+		refName, ok := md.Annotations[imgspecv1.AnnotationRefName]
+		if !ok {
+			refName = "@" + md.Digest.Encoded()
+			if ref, err = layout.NewIndexReference(r.tempDirectory, i); err != nil {
```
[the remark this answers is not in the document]
QiWang19: Reimplemented the ociArchiveImageSource/Destination to hide this directory to be a field under archive reference, would this a feasible way?

rhatdan:
@QiWang19 @mtrmac Any movement on this?

rhatdan:
@vrothberg PTAL
@QiWang19 Could you rebase and push.

PR #1073, opened 2020-11-02
Fix misleading network error

### LINKED ISSUE #1058: Network error message is misleading
When using `skopeo copy docker://... oci:...` from a Docker repository that closes the https connection early while transferring a layer, the error message is `Error writing blob: unexpected EOF`.

I would have expected the error to be about "reading" instead of "writing", and there wasn't much context to indicate that it was a network EOF.  In my opinion, the most helpful error message would include the layer URL that failed along with details indicating that the server closed the TCP/https connection before all data arrived.

QiWang19:
Fix misleading network error `Error writing blob: unexpected EOF`
close #1058 

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -1166,6 +1166,21 @@ func computeDiffID(stream io.Reader, decompressor compression.DecompressorFunc) 
 	return digest.Canonical.FromReader(stream)
 }
 
+// errorAnnotationReader wraps the io.Reader passed to PutBlob for annotating the error happened during read.
+// These errors are reported as PutBlob errors, so we would otherwise misleadingly attribute them to the copy destination.
+type errorAnnotationReader struct {
+	reader io.Reader
+}
+
+// Read annotates the error happened during read
+func (r errorAnnotationReader) Read(b []byte) (n int, err error) {
+	n, err = r.reader.Read(b)
+	if err != io.EOF {
+		return n, errors.Wrapf(err, "error happened during read")
+	}
+	return n, err
+}
+
 // copyBlobFromStream copies a blob with srcInfo (with known Digest and Annotations and possibly known Size) from srcStream to dest,
 // perhaps sending a copy to an io.Writer if getOriginalLayerCopyWriter != nil,
 // perhaps compressing it if canCompress,
@@ -1335,7 +1350,7 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
 	}
 
 	// === Finally, send the layer stream to dest.
-	uploadedInfo, err := c.dest.PutBlob(ctx, destStream, inputInfo, c.blobInfoCache, isConfig)
+	uploadedInfo, err := c.dest.PutBlob(ctx, &errorAnnotationReader{destStream}, inputInfo, c.blobInfoCache, isConfig)
 	if err != nil {
 		return types.BlobInfo{}, errors.Wrap(err, "Error writing blob")
 	}


mtrmac (commented):
All errors from `io.Copy` are not necessarily caused by failures on the read side (e.g. the system can be out of disk space).

AFAICS this needs to be implemented as a wrapper around the `io.Reader` to mark the read-side errors, and possibly with an extra utility to detect such marked errors.

As for whether to do this in the transports (so that non-`copy.Image` callers still get useful errors) or just in `copy.Image` (so that it happens in a single place and the transports don’t have to care), I don’t have a strong opinion — just that if this happens per-transport, it should be ~required part of the behavior, and added to _all_ transports (notably to `docker://` uploads, where we’ve been hitting this most often).

QiWang19:
@mtrmac PTAL.  Added the wrap function that apply to the oci transport.

mtrmac (commented):
Yup, this is the minimal fix. Do you plan to apply it in `copy.Image`, or make it a responsibility of every single `PutBlob`?

QiWang19:
Apply it to copyBlobFromStream to work for PutBlob.

mtrmac (commented):
ACK, works for me. Maybe rename to something a bit more specific like `errorAnnotationReader`? Every `Reader` is a “stream Reader”.

copy/copy.go
```diff
@@ -1166,6 +1166,18 @@ func computeDiffID(stream io.Reader, decompressor compression.DecompressorFunc)
...
 }
 
+type streamReader struct {
+	reader io.Reader
+}
+
+func (r streamReader) Read(b []byte) (n int, err error) {
+	n, err = r.reader.Read(b)
+	if err != nil {
+		return n, errors.Wrapf(err, "error happened during read")
```
mtrmac, line 1176: Oops, looking at tests, this must not annotate `io.EOF`, because exactly that value is a part of the `io.Reader` protocol.

copy/copy.go
```diff
@@ -1166,6 +1166,18 @@ func computeDiffID(stream io.Reader, decompressor compression.DecompressorFunc)
...
 }
 
+type errorAnnotationReader struct {
+	reader io.Reader
+}
+
+func (r errorAnnotationReader) Read(b []byte) (n int, err error) {
+	n, err = r.reader.Read(b)
+	if err != nil && err != io.EOF {
+		return n, errors.Wrapf(err, "error happened during read")
```
mtrmac, line 1176: Absolutely non-blocking: Maybe just `err = errors.Wrapf(…)` and let it fall through here

copy/copy.go
```diff
@@ -1166,6 +1166,18 @@ func computeDiffID(stream io.Reader, decompressor compression.DecompressorFunc)
 	return digest.Canonical.FromReader(stream)
 }
 
+type errorAnnotationReader struct {
```
mtrmac, line 1169: Non-blocking: A one-line comment about why this needs to exist would be nice, so that readers 3 years later don’t have to look up the history on GitHub (if it even exists.) I’m sorry, I should have said this earlier.

mtrmac (commented):
LGTM.

mtrmac:
Thanks!

copy/copy.go
```diff
@@ -1166,6 +1166,20 @@ func computeDiffID(stream io.Reader, decompressor compression.DecompressorFunc)
 	return digest.Canonical.FromReader(stream)
 }
 
+// errorAnnotationReader wraps the io.Reader passed to PutBlob for annotating the error happened during read.
```
mtrmac, line 1169:
```suggestion
// errorAnnotationReader wraps the io.Reader passed to PutBlob for annotating the error happened during read.
// These errors are reported as PutBlob errors, so we would otherwise misleadingly attribute them to the copy destination.
```
The _why_ is the more difficult part to reverse-engineer from code.

mtrmac: If you want this to be merged as-is, that’s fine as well.

PR #1075, opened 2020-11-04
Add a signedIdentity choice "type": "remapIdentity"

mtrmac:
This allows accepting signatures for a complete or partial mirror of some other repository namespace in a single step, similar in signing effect to setting up mirrors in `registries.conf`, but letting image consumers refer to the mirrors directly.

For tag/digest matching, this currently only implemnents the default `matchRepoDigestOrExact`-like semantics; it's the right choice for almost all users, and we can add other alternatives later if it turned out to be necessary.

### DIFF
--- a/signature/policy_config.go
+++ b/signature/policy_config.go
@@ -19,6 +19,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"regexp"
 
 	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/image/v5/transports"
@@ -507,6 +508,8 @@ func newPolicyReferenceMatchFromJSON(data []byte) (PolicyReferenceMatch, error) 
 		res = &prmExactReference{}
 	case prmTypeExactRepository:
 		res = &prmExactRepository{}
+	case prmTypeRemapIdentity:
+		res = &prmRemapIdentity{}
 	default:
 		return nil, InvalidPolicyFormatError(fmt.Sprintf("Unknown policy reference match type \"%s\"", typeField.Type))
 	}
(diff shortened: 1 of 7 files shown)


mtrmac:
https://github.com/containers/skopeo/pull/1121 adds a Skopeo end-to-end demonstration of the feature.

mtrmac:
@vrothberg @rhatdan PTAL.

signature/policy_reference_match_test.go
```diff
@@ -358,3 +412,193 @@ func TestPRMExactRepositoryMatchesDockerReference(t *testing.T) {
...
+	for _, c := range []struct {
+		ref, prefix string
+		expected    bool
+	}{
+		// Prefix is a reference.Domain() value
+		{"docker.io/image", "docker.io", true},
+		{"docker.io/image", "example.com", false},
+		{"example.com:5000/image", "example.com:5000", true},
+		{"example.com:50000/image", "example.com:5000", false},
+		{"example.com:5000/image", "example.com", false},
```
vrothberg, line 426: This case may be worth mentioning explicitly in the docs.  Some users may assume that "host" would prefix "host:port".

signature/policy_config.go
```diff
@@ -693,3 +696,76 @@ func (prm *prmExactRepository) UnmarshalJSON(data []byte) error {
...
+func validateIdentityRemappingPrefix(s string) error {
+	if remapIdentityDomainRegexp.MatchString(s) ||
+		(remapIdentityNameRegexp.MatchString(s) && remapIdentityDomainPrefixRegexp.MatchString(s)) {
+		// FIXME? This does not reject "shortname" nor "ns/shortname", because docker/reference
+		// does not provide an API for the short vs. long name logic.
+		// It will either not match, or fail in the ParseNamed call of
+		// prmRemapIdentity.remapReferencePrefix when trying to use such a prefix.
+		return nil
+	}
+	return InvalidPolicyFormatError(fmt.Sprintf("prefix \"%s\" is not valid", s))
```
vrothberg, line 723: Mini nit: Could use `%q` instead of quoting `%s`.

vrothberg (commented):
Mini nit and a doc suggestion. Other than that, LGTM :+1:

signature/policy_reference_match_test.go
mtrmac: Added, as well as a note that references not matching `prefix` are not remapped, and a heading to make allow faster skimming.

(The man page formatting is not good, but it is at least not worse than before.)

signature/policy_config.go
mtrmac: Good point.

(I have left one other case as is to be consistent with the rest of the code, and two other cases don’t need that because the inputs are valid references and don’t contain characters that would need escaping.)

signature/policy_reference_match_test.go
vrothberg: I don't see a new commit? Did you push?

mtrmac: Oops, pushed the wrong repo. Updated now.

vrothberg (commented):
LGTM, nice work, @mtrmac !

PR #1076, opened 2020-11-06
sysregistriesv2: Split configuration loading and merging

mtrmac:
Currently just RFC and to run tests, depending on logistics we might end up making this change as part of, or after, #1060


mtrmac:
Superseded by #1077 after #1060.

PR #1078, opened 2020-11-10
Image's digest from HEAD request

crazy-max:
### Context

Docker Hub is instituting [pull rate-limits for anonymous and free users](https://www.docker.com/increase-rate-limits) since Nov 2, 2020.

> You are receiving this email because of a policy change to Docker products and services you use. On Monday, November 2, 2020 at 9am Pacific Standard Time, Docker will begin enforcing rate limits on container pulls for Anonymous and Free users. Anonymous (unauthenticated) users will be limited to 100 container image pulls every six hours, and Free (authenticated) users will be limited to 200 container image pulls every six hours, when enforcement is fully implemented. Docker Pro and Team subscribers can pull container images from Docker Hub without restriction, as long as the quantities are not excessive or abusive.

With this limit in place, every time a [NewImageSource](https://github.com/containers/image/blob/45cb72b02d1ab6b821b5dc0c1b43972136bd6405/docker/docker_image_src.go#L126) is created, it will make a GET request on the manifest endpoint which counts as a pull and will return this body response if limit is reached:

```
{"message":"toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit"}
```

### Solution

First I have removed the [ensureManifestIsLoaded](https://github.com/containers/image/compare/master...crazy-max:head-digest?expand=1#diff-0d863bb03c7af220a36b2674f5e7f0f559c68f4d2e452816159d9a1971688b11L126) in NewImage/NewImageSource func to prevent the initial (not required?) manifest GET request.

I have also added a new interface to be able to retrieve the manifest digest through a HEAD request [which does not count as a pull on Docker Hub](https://github.com/moby/moby/pull/41607#issuecomment-723380166). [GetDigest is exposed](https://github.com/containers/image/compare/master...crazy-max:head-digest?expand=1#diff-f8d2bd35e1a8c9b86c8c4448dde00bc5480652f85ed0630efc012c3906d3a14fR235) to be able to use the digest from the HEAD request on [Diun](https://github.com/crazy-max/diun) and compared with a manifest stored in my bbolt database:

```
Tue, 10 Nov 2020 17:11:52 CET DBG [containers/image] Loading registries configuration "/etc/containers/registries.conf"
Tue, 10 Nov 2020 17:11:52 CET DBG [containers/image] Trying to access "docker.io/crazymax/diun:latest"
Tue, 10 Nov 2020 17:11:52 CET DBG [containers/image] Returning credentials from DockerAuthConfig
Tue, 10 Nov 2020 17:11:52 CET DBG [containers/image] Using registries.d directory /etc/containers/registries.d for sigstore configuration
Tue, 10 Nov 2020 17:11:52 CET DBG [containers/image]  No signature storage configuration found for docker.io/crazymax/diun:latest, using built-in default file://C:%5CUsers%5Ccrazy%5C.local%5Cshare%5Ccontainers%5Csigstore
Tue, 10 Nov 2020 17:11:52 CET DBG [containers/image] Looking for TLS certificates and private keys in \etc\docker\certs.d\docker.io
Tue, 10 Nov 2020 17:11:52 CET DBG [containers/image] GET https://registry-1.docker.io/v2/
Tue, 10 Nov 2020 17:11:53 CET DBG [containers/image] Ping https://registry-1.docker.io/v2/ status 401
Tue, 10 Nov 2020 17:11:53 CET DBG [containers/image] GET https://auth.docker.io/token?account=pharmagest&scope=repository%3Acrazymax%2Fdiun%3Apull&service=registry.docker.io
Tue, 10 Nov 2020 17:11:53 CET DBG [containers/image] HEAD https://registry-1.docker.io/v2/crazymax/diun/manifests/latest
Tue, 10 Nov 2020 17:11:54 CET DBG [containers/image] Content-Type from digest HEAD is "application/vnd.docker.distribution.manifest.list.v2+json"
```

### Further enhancement

Maybe this mechanic could be used in your internal cache to avoid unnecessary call to the manifest GET endpoint?

Let me know if I've missed something and thanks again for this great lib!

### DIFF
--- a/docker/docker_image.go
+++ b/docker/docker_image.go
@@ -4,12 +4,15 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net/http"
 	"net/url"
 	"strings"
 
 	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/image/v5/image"
+	"github.com/containers/image/v5/manifest"
 	"github.com/containers/image/v5/types"
+	"github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
 )
 
@@ -102,4 +105,48 @@ func GetRepositoryTags(ctx context.Context, sys *types.SystemContext, ref types.
 		}
 	}
 	return tags, nil
+}
+
+// GetDigest returns the image's digest
+// Use this to optimize and avoid use of an ImageSource based on the returned digest;
+// if you are going to use an ImageSource anyway, it’s more efficient to create it first
+// and compute the digest from the value returned by GetManifest.
+// NOTE: Implemented to avoid Docker Hub API limits, and mirror configuration may be
+// ignored (but may be implemented in the future)
+func GetDigest(ctx context.Context, sys *types.SystemContext, ref types.ImageReference) (digest.Digest, error) {
+	dr, ok := ref.(dockerReference)
+	if !ok {
+		return "", errors.Errorf("ref must be a dockerReference")
+	}
+
+	tagOrDigest, err := dr.tagOrDigest()
+	if err != nil {
+		return "", err
+	}
+
+	client, err := newDockerClientFromRef(sys, dr, false, "pull")
+	if err != nil {
+		return "", errors.Wrap(err, "failed to create client")
+	}
+
+	path := fmt.Sprintf(manifestPath, reference.Path(dr.ref), tagOrDigest)
+	headers := map[string][]string{
+		"Accept": manifest.DefaultRequestedManifestMIMETypes,
+	}
+
+	res, err := client.makeRequest(ctx, "HEAD", path, headers, nil, v2Auth, nil)
+	if err != nil {
+		return "", err
+	}
+
+	if res.StatusCode != http.StatusOK {
+		return "", errors.Wrapf(registryHTTPResponseToError(res), "Error reading digest %s in %s", tagOrDigest, dr.ref.Name())
+	}
+
+	dig, err := digest.Parse(res.Header.Get("Docker-Content-Digest"))
+	if err != nil {
+		return "", err
+	}
+
+	return dig, nil
 }


rhatdan:
@mtrmac PTAL

rhatdan:
Does this article help?

https://developers.redhat.com/blog/2019/08/14/best-practices-for-running-buildah-in-a-container/

Podman inside of a container is a lot more complex.  Remember if you are doing this with Docker, you need to disable the seccomp filters, or use /usr/share/containers/seccomp.json

Using podman for running the locked down Buildah container is the preferred mechanism.

mtrmac:
Thanks for the PR.

Removing `ensureManifestIsLoaded` from `newImageSourceAttempt` would completely break the mirroring implementation.

Does this actually end up reducing the number of requests? Basically every operation that can be done while reading c/image (copy an image, “inspect” it, …) will end up reading the manifest as one of the first things, and `ensureManifestIsLoaded` as well as `image.Sourced`, so AFAICT this typically only _moves_ the manifest read, not eliminates one.

Unless I guess you have an _external_ caller in mind for the new `GetDigest` call? What does it do with the value?

crazy-max:
@mtrmac 

> Removing `ensureManifestIsLoaded` from `newImageSourceAttempt` would completely break the mirroring implementation. Does this actually end up reducing the number of requests?

Argh that's what I thought but yeah it helps removing the first manifest GET request when a new image interface is setted up. Maybe `ensureManifestIsLoaded` could be called in a further func to preserve mirroring impl?

> Basically every operation that can be done while reading c/image (copy an image, “inspect” it, …) will end up reading the manifest as one of the first things, and ensureManifestIsLoaded as well as image.Sourced, so AFAICT this typically only moves the manifest read, not eliminates one.

Yeah that's my point I don't want the GET manifest endpoint to be called when I instantiate a new image.

> Unless I guess you have an external caller in mind for the new GetDigest call? What does it do with the value?

I [compare a manifest digest located in my own bolt db to the `GetDigest` HEAD req to know if I need or not to GET a fresh manifest](https://github.com/crazy-max/diun/blob/cee21b6983b441af27fe66c98a6ace5e6eb855e4/pkg/registry/manifest.go#L41-L70). This helps to reduce unnecessary calls to the GET manifest endpoint.

I understand that my implementation might be outside the scope of this library so don't hesitate to tell me if this is the case.

mtrmac:
> I [compare a manifest digest located in my own bolt db to the `GetDigest` HEAD req to know if I need or not to GET a fresh manifest](https://github.com/crazy-max/diun/blob/cee21b6983b441af27fe66c98a6ace5e6eb855e4/pkg/registry/manifest.go#L41-L70). This helps to reduce unnecessary calls to the GET manifest endpoint.
> 
> I understand that my implementation might be outside the scope of this library so don't hesitate to tell me if this is the case.

That’s not something the library _currently_ can do, OTOH there’s a clear benefit to sharing the auth/token/… infrastructure with c/image/docker. I’m interesting in enabling this, although not at any cost.

Adding methods to an interface like `ImageSource` breaks API (external implementations of that interface); it’s still _possible_ to build another way (although we don’t have a settled convention), with more code.

Primarily it’s not obvious to me that this generalizes — basically this is an optimization for a _single server on the internet_ providing two operations that are ~exactly as expensive but one is by policy rate-limited, whereas most other servers and most other `ImageTransport` implementations really don’t care either way and could just as well provide the `GetManifest` data.

---

At this point it seems more natural to me to say “this is a very Docker-specific”, so I’d prefer to provide this as a Docker-specific API integrating with the Docker auth/token infrastructure but not with `ImageTransport` (like `GetRepositoryTags` and `SearchRegistry`): `c/image/docker.GetDigest(context.Context, *types.SystemContext, types.ImageReference)` or so. If we find out later that this _does_ generalize, we can always add it to `ImageSource` (modulo the API concerns) later.

OTOH that would require setting up a separate `dockerClient` = getting a new bearer token. Is that rate-limited as well? That would make this approach unviable. The FAQ seems to suggest it’s not rate-limited, at least.

@vrothberg RFC

crazy-max:
@mtrmac 

> At this point it seems more natural to me to say “this is a very Docker-specific”, so I’d prefer to provide this as a Docker-specific API integrating with the Docker auth/token infrastructure but not with `ImageTransport` (like `GetRepositoryTags` and `SearchRegistry`): `c/image/docker.GetDigest(context.Context, *types.SystemContext, types.ImageReference)` or so. If we find out later that this _does_ generalize, we can always add it to `ImageSource` (modulo the API concerns) later.

You're absolutely right, it's a Docker-specific case. I've made changes to reflect that without breaking the API. Let me know if it looks good to you.

docker/docker_image.go
```diff
@@ -103,3 +107,50 @@ func GetRepositoryTags(ctx context.Context, sys *types.SystemContext, ref types.
...
+	path := fmt.Sprintf(manifestPath, reference.Path(dr.ref), tagOrDigest)
+	headers := map[string][]string{
+		"Accept": manifest.DefaultRequestedManifestMIMETypes,
+	}
+
+	res, err := client.makeRequest(ctx, "HEAD", path, headers, nil, v2Auth, nil)
+	if err != nil {
+		return nil, err
+	}
+	logrus.Debugf("Content-Type from manifest HEAD is %q", res.Header.Get("Content-Type"))
```
mtrmac, line 144: This exists in `ensureManifestIsLoaded` to debug conversion issues and misconfigured servers; it might not be necessary here. OTOH it doesn’t hurt anything.

docker/docker_image.go
```diff
@@ -103,3 +107,50 @@ func GetRepositoryTags(ctx context.Context, sys *types.SystemContext, ref types.
...
+
+// GetDigest returns the image's digest. (This is a
+// backward-compatible shim method which calls the module-level
+// GetDigest)
+func (i *Image) GetDigest(ctx context.Context) (*digest.Digest, error) {
+	return GetDigest(ctx, i.src.c.sys, i.src.logicalRef)
+}
+
+// GetDigest returns the image's digest
+func GetDigest(ctx context.Context, sys *types.SystemContext, ref types.ImageReference) (*digest.Digest, error) {
```
mtrmac, line 119: Typically we return `digest.Digest` by value (except for the `instanceDigest` values, which use a pointer to represent an optional value), and just return `""` in the error cases.

docker/docker_image.go
```diff
@@ -103,3 +107,50 @@ func GetRepositoryTags(ctx context.Context, sys *types.SystemContext, ref types.
...
 }
+
+// GetDigest returns the image's digest. (This is a
+// backward-compatible shim method which calls the module-level
+// GetDigest)
+func (i *Image) GetDigest(ctx context.Context) (*digest.Digest, error) {
+	return GetDigest(ctx, i.src.c.sys, i.src.logicalRef)
+}
+
+// GetDigest returns the image's digest
```
mtrmac, line 118: Maybe something like
```suggestion
// GetDigest returns the image's digest
// NOTE: This optimizes to avoid Docker Hub API limits, and mirror configuration may be ignored (but may be implemented in the future)
```
or implement the mirror lookup the same way `newImageSource` does. (As the only current user, it’s up to you.)

mtrmac, line 118: Absolutely non-blocking: Also maybe
```suggestion
// GetDigest returns the image's digest
// Use this to optimize and avoid use of an ImageSource based on the returned digest;
// if you are going to use an ImageSource anyway, it’s more efficient to create it first
// and compute the digest from the value returned by GetManifest.
```
to explain the trade-offs.

mtrmac (commented):
ACK overall.

docker/docker_image.go
crazy-max: I prefer to keep this for debugging purposes if it's ok with you.

crazy-max:
@mtrmac Done, thanks for your review.

mtrmac (commented):
Thanks! LGTM.

@vrothberg PTAL, I’d like a second pair of eyes on the API.

vrothberg (commented):
LGTM, nice work @crazy-max and @mtrmac

PR #1079, opened 2020-11-11
docker_client: Handle "invalid_scope" errors

rhafer:
podman 2.1.1 (which uses containers/images 5.7.0) fails to search some registries. E.g.

`podman --log-level=debug search registry.suse.com/`

gives:
```
[...]
DEBU[0000] trying to talk to v2 search endpoint         
DEBU[0000] GET https://registry.suse.com/v2/            
DEBU[0000] Ping https://registry.suse.com/v2/ status 401 
DEBU[0000] GET https://registry.suse.com/auth?service=SUSE+Linux+Docker+Registry 
DEBU[0000] Increasing token expiration to: 60 seconds   
DEBU[0000] GET https://registry.suse.com/v2/_catalog    
ERRO[0000] error getting search results from v2 endpoint "registry.suse.com": unable to retrieve auth token: invalid username/password: 
```

Looking into the details of the requests and response and trying to repeat it with `curl` I can see that the request to `/v2/_catalog` failes with Status 401 and the www-authenticate containing this error:

```
< www-authenticate: Bearer realm="https://registry.suse.com/auth",service="SUSE Linux Docker Registry",scope="registry:catalog:*",error="insufficient_scope"
```
This can be fixed by simply fetching a new token with the `scope` value set to the value return in that header.

This PR tries to implement a fix for the above behavior by checking for the error and retrying the request with and updated token.

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -123,8 +123,9 @@ type dockerClient struct {
 }
 
 type authScope struct {
-	remoteName string
-	actions    string
+	resourceType string
+	remoteName   string
+	actions      string
 }
 
 // sendAuth determines whether we need authentication for v2 or v1 endpoint.
@@ -233,6 +234,7 @@ func newDockerClientFromRef(sys *types.SystemContext, ref dockerReference, write
 		client.registryToken = sys.DockerBearerRegistryToken
 	}
 	client.signatureBase = sigBase
+	client.scope.resourceType = "repository"
 	client.scope.actions = actions
 	client.scope.remoteName = reference.Path(ref.ref)
 	return client, nil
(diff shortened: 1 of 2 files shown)


docker/docker_client.go
```diff
@@ -447,7 +447,37 @@ func (c *dockerClient) makeRequest(ctx context.Context, method, path string, hea
...
 	url := fmt.Sprintf("%s://%s%s", c.scheme, c.registry, path)
-	return c.makeRequestToResolvedURL(ctx, method, url, headers, stream, -1, auth, extraScope)
+	resp, err := c.makeRequestToResolvedURL(ctx, method, url, headers, stream, -1, auth, extraScope)
+
+	// As by default the challenges from the /v2/ "ping" response were used for fetching the
+	// auth token, we might be getting a error back indicating that the auth scope was insufficient
+	// check for that and update the client challenges to retry with requesting a new token
+	if retry, challenges := needsRetryWithUpdatedChallenges(err, resp); retry {
+		logrus.Debug("Detected insufficent_scope error, retrying request with updated challenges")
+		c.challenges = challenges
```
mtrmac, line 457: `dockerClient`, once initialized, can be used concurrently (`HasThreadSafe…`), so overwriting a property which is intentionally in the “… and never change afterwards” section of `dockerClient` is not safe. (This is also why `extraScope` is a parameter and not a property.)

mtrmac (commented):
Thanks!

There’s an immediate concurrency concern. Other than that, based on a very quick initial look, this seems useful and reasonable at a first glance.

(Note that `Parameters["scope"]` is not used at all, because Docker has historically ignored it, and some implementations just don’t return useful values. Using that only on a failure where the previous code would give up is almost certainly fine, but it will be necessary for me to return to the old research.)

docker/docker_client.go
rhafer: Ah, ok. Must have overlooked that :(. I had initially thought to put the updated scope into the `extraScope` argument, but as that would have required to rework the `authScope` struct a bit (as the current code always assumes the resourcetype part of it to be `repository` I went of the above approach in my first try. I guess I need to rethink that again :).

 What do you think, does it make sense to pass the updated scope via the `extraScope` argument?

rhafer:
This would need another rebase. But I am not able to spend any time on this. So I'll just close it.

mtrmac:
Sorry about that, not getting back to the PR is totally our fault. I have at least filed #1478 so that we track this action item.

josegomezr:
@rhafer Could I take over this code? I have some free time I'd like to invest in this.

rhafer:
@josegomezr Sure, go ahead. As stated I won't be able to spent any time on this. (Neither does this bug affect me personally anymore)

rhatdan:
@josegomezr Did you ever work on this?

PR #644, opened 2019-06-13
ostree: improve error message

giuseppe:
when lsetfilecon_raw fails we also print the context it tried to
set.

Signed-off-by: Giuseppe Scrivano <gscrivan@redhat.com>

PR #1084, opened 2020-11-19
pkg/compression: new zstd variant zstd:chunked

giuseppe:
add a new custom variant of the zstd compression that permits to retrieve each file separately.

The idea is based on CRFS and its stargz format for having seekable and indexable tarballs.

zstd_chunked

One disadvantage of the stargz format is that a custom file is added to the tarball to store the layer metadata, and the metadata file is part of the image itself.  Clients that are not aware of the stargz format will propagate the metadata file inside of the containers.
    
The zstd compression supports embedding additional data as part of the stream that the zstd decompressor will ignore (skippable frame), so the issue above with CRFS can be solved directly within the zstd compression format.

Beside this minor advantage, zstd is much faster and compresses better than gzip, so take this opportunity to push the zstd format further.

The zstd compression is supported by the OCI image specs since August 2019: https://github.com/opencontainers/image-spec/pull/788 and has been supported by containers/image since then.

Clients that are not aware of the zstd:chunked format, won't notice any difference when handling a blob that uses the variant.

Since access to host files is required for the deduplication, the untar doesn't happen in a separate process running in a chroot, but it happens in original mount namespace.  To mitigate possible breakages, openat2(2) is used were available.

Requires: https://github.com/containers/storage/pull/775


giuseppe:
@cyphar what do you think about this idea? :-)

I've implemented (still hacky) the deduplication also with host files, e.g. pulling fedora:33 on a Fedora 33 host gives this result:

```
$ skopeo copy docker://docker.io/gscrivano/zstd-chunked:fedora33 containers-storage:fedora:33
Copying blob 39e269c6c24a [========++++++++++++++++++++++++++++++] 68.4MiB / 68.4MiB (deduplicated: 54.0MiB = 78.87%)
Copying config 7985d34067 done  
Storing signatures
```

copy/copy.go
```diff
@@ -182,6 +183,8 @@ type Options struct {
 	// OciDecryptConfig contains the config that can be used to decrypt an image if it is
 	// encrypted if non-nil. If nil, it does not attempt to decrypt an image.
 	OciDecryptConfig *encconfig.DecryptConfig
+
+	FetchPartialBlobs bool // attempt to fetch the blob partially.  Experimental.
```
mtrmac, line 187: I have no idea what this means, and the test Skopeo PR hard-codes this to true. Is this intended to be an end-user option, or just a feature flag for a WIP feature?

copy/copy.go
```diff
@@ -1426,7 +1530,9 @@ func (c *copier) compressGoroutine(dest *io.PipeWriter, src io.Reader, compressi
 	if err != nil {
 		return
 	}
-	defer compressor.Close()
+	defer func() {
+		err = compressor.Close()
```
mtrmac, line 1534: Typically the error from `CopyBuffer` is much more important, because it probably contains an error reading from the source; it must not be just discarded. If the `Close` error becomes important as well, maybe both need to be reported.

docker/errors.go
```diff
@@ -17,6 +17,14 @@ var (
 	ErrTooManyRequests = errors.New("too many requests to registry")
 )
 
+// ErrBadRequest is returned when the status code returned is 400
+type ErrBadRequest struct {
```
mtrmac, line 21: Is there any user for this? If this is supposed to be a part of the `GetBlobAt` interface, the caller is probably not going to be checking for a docker-transport-specific error.

storage/chunked_zstd.go
```diff
@@ -0,0 +1,834 @@
...
+	}
+	return dstFile, written, nil
+}
+
+func maybeDoIDRemap(manifest []compression.FileMetadata, options *archive.TarOptions) error {
+	if options.ChownOpts == nil && len(options.UIDMaps) == 0 || len(options.GIDMaps) == 0 {
+		return nil
+	}
+
+	idMappings := idtools.NewIDMappingsFromMaps(options.UIDMaps, options.GIDMaps)
```
mtrmac, line 159: So far c/image has managed not to have to understand ID mappings.

types/types.go
```diff
@@ -331,6 +331,24 @@ type ImageDestination interface {
...
+}
+
+// ImageSourceSeekable is an image source that permits to fetch chunks of the entire blob.
+type ImageSourceSeekable interface {
+	// GetBlobAt returns a stream for the specified blob.
+	GetBlobAt(context.Context, BlobInfo, []ImageSourceChunk) (io.ReadCloser, string, error)
+}
+
+// ImageDestinationPartial is a service to store a blob by requesting the missing chunks to a ImageSourceSeekable.
+type ImageDestinationPartial interface {
```
mtrmac, line 347: We should very seriously discuss whether we want to keep any future additions to the transports a public API, or to just close it down as internal.

I do realize that there are other implementation, mainly various magic wrappers in Buildah; OTOH the need to keep these things stable has been a very notable source of overhead.

storage/chunked_zstd.go
```diff
@@ -0,0 +1,834 @@
...
+}
+
+func (d whiteoutHandler) Setxattr(path, name string, value []byte) error {
+	file, err := openFileUnderRoot(name, d.Root, d.Dirfd, unix.O_WRONLY, 0)
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+
+	return unix.Fsetxattr(int(file.Fd()), name, value, 0)
```
mtrmac, line 627: I am rather unhappy about having things like that outside c/storage, yet expected to perfectly interoperate. Why shouldn’t the existing graph drivers be in charge of every aspect of the tar -> filesystem translation?

copy/copy.go
```diff
@@ -1041,22 +1084,45 @@ type diffIDResult struct {
...
+
 // copyLayer copies a layer with srcInfo (with known Digest and Annotations and possibly known Size) in src to dest, perhaps compressing it if canCompress,
 // and returns a complete blobInfo of the copied layer, and a value for LayerDiffIDs if diffIDIsNeeded
 func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, toEncrypt bool, pool *mpb.Progress) (types.BlobInfo, digest.Digest, error) {
 	cachedDiffID := ic.c.blobInfoCache.UncompressedDigest(srcInfo.Digest) // May be ""
 	// Diffs are needed if we are encrypting an image or trying to decrypt an image
 	diffIDIsNeeded := ic.diffIDsAreNeeded && cachedDiffID == "" || toEncrypt || (isOciEncrypted(srcInfo.MediaType) && ic.c.ociDecryptConfig != nil)
 
+	srcAlgo := ""
+	if srcInfo.CompressionAlgorithm != nil {
```
mtrmac, line 1112: `srcInfo.Compression*` is supposed to never be set at this point; they really don’t belong to `BlobInfo` at all, it’s an output of `copyBlobFromStream` to be consumed by `UpdatedImage`.

copy/copy.go
```diff
@@ -1070,14 +1136,52 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, to
 		}
 	}
 
+	imgSource, okSource := ic.c.rawSource.(types.ImageSourceSeekable)
+	imgDest, okDest := ic.c.dest.(types.ImageDestinationPartial)
+	if ic.c.fetchPartialBlobs && okSource && okDest && !diffIDIsNeeded {
```
mtrmac, line 1141: ACK, this approach of a “short-cut in the generic code” makes sense.

storage/storage_image.go
```diff
@@ -459,6 +475,118 @@ func (s *storageImageDestination) PutBlob(ctx context.Context, stream io.Reader,
...
+func (s *storageImageDestination) PutBlobPartial(ctx context.Context, stream types.ImageSourceSeekable, srcInfo types.BlobInfo, cache types.BlobInfoCache) (types.BlobInfo, error) {
+	blobDigest := srcInfo.Digest
+
+	readerAt := seekableToReaderAt{
+		stream:  stream,
+		srcInfo: srcInfo,
+		ctx:     ctx,
+	}
+
+	data, err := compression.ReadChunkedZstdManifest(&readerAt, srcInfo.Size)
```
mtrmac, line 554: The `PutBlobPartial` interface looks generic. Just blindly hard-coding Zstd here seems surprising. (Maybe all it takes is a MIME type check.)

storage/chunked_zstd.go
```diff
@@ -0,0 +1,834 @@
...
+	newFileFlags           = (unix.O_CREAT | unix.O_TRUNC | unix.O_WRONLY | unix.O_EXCL)
+)
+
+type DirectDiffInput struct {
+	stream         types.ImageSourceSeekable
+	store          *storageImageDestination
+	manifest       []byte
+	ctx            context.Context
+	srcInfo        types.BlobInfo
+	layersMetadata map[string][]compression.FileMetadata
```
mtrmac, line 51: All of the fields, and especially the maps/arrays, will eventually need documentation.

storage/storage_image.go
```diff
@@ -682,6 +810,31 @@ func (s *storageImageDestination) Commit(ctx context.Context, unparsedToplevel t
...
+		if diffOutput := s.diffOutputs[blob.Digest]; diffOutput != nil {
+			layer, err := s.imageRef.transport.store.CreateLayer(id, lastLayer, nil, "", false, nil)
+			if err != nil {
+				return err
+			}
+
+			// FIXME: what to do with the uncompressed digest?
+			diffOutput.UncompressedDigest = blob.Digest
+
+			if err := s.imageRef.transport.store.ApplyDiffFromStagingDirectory(layer.ID, diffOutput.Target, diffOutput, nil); err != nil {
```
mtrmac, line 823: Is the two-stage process really _required_ for this?

We’d already like for the image pull process to happen concurrently for layers as they arrive (pull in any order, and create layers in order as soon as possible, instead of currently just staging everything and then doing a lot of work in `Commit`). If that’s the more natural approach for this feature, it would probably make sense to return to that work and get it working, and have a simpler implementation and a more straightforward c/storage API.

mtrmac (commented):
Just an extremely partial skim, I have read < 10% of the code, just to get a basic idea of the broad structure.
- At a first uninformed glance, I’d prefer all of the filesystem knowledge to be in c/storage.
- Should this wait until the c/image/storage parallel pull code (not necessarily the more complex “concurrent pull detection” part) is revived?
- Is there a relationship to #902 ? Is this strictly more powerful, or are the two independent optimizations, each with its own advantages, that can live alongside each other?

mtrmac:
Oh, and: how does digest verification happen with this, or what replaces it to obtain equivalent security?

copy/copy.go
giuseppe: thanks for noticing it, I pushed a fixed version

docker/errors.go
giuseppe: `storage/chunked_zstd.go` is using it

types/types.go
giuseppe: I agree and I thought about it as well.  I'd prefer to keep it private especially for now and have the possibility to iterate on it

copy/copy.go
giuseppe: It is hard coded just for testing purposes.  It should be an option, at least for now

giuseppe:
> * 
> 
> * Is there a relationship to #902 ? Is this strictly more powerful, or are the two independent optimizations, each with its own advantages, that can live alongside each other?

they are independent.  #902 requires deltas to be generated and they are useful when e.g. pulling an updated image.  This PR instead is to improve pulling of a generic layer and doing deduplication with all the files in the storage.

Another advantage of this PR is that it performs local deduplication as well.  If a file is present locally, it is deduplicated with reflinks when the underlying file system supports them.

The two stages "prepare staging directory" & "commit from staging directory" breaks the c/storage abstraction that everything is a tarball and can be handled transparently by different drivers (in such case the `PutBlob()` logic is still fine), but I think it is an improvement when using overlay, the "Commit()" is basically a `os.Rename()` and can be easily done in parallel.

giuseppe:
> * 
> 
> * Should this wait until the c/image/storage parallel pull code (not necessarily the more complex “concurrent pull detection” part) is revived?

I think will simplify the "parallel pull" as each "commit to staging directory" can be done separately without any interference among the different layers.  The `Commit()` can still be left sequential, but being a very cheap operation, that won't be a problem. @vrothberg What do you think?

vrothberg:
> I think will simplify the "parallel pull" as each "commit to staging directory" can be done separately without any interference among the different layers. The `Commit()` can still be left sequential, but being a very cheap operation, that won't be a problem. @vrothberg What do you think?

I only briefly skimmed this PR and do not yet understand how the two PRs relate to another.  In case an optimized commit would facilitate this work, let me know.  We can prioritize it.  If others want to tackle it earlier, I am more than happy :)

At the moment, I channel all my mental resources on getting the podman-cp rewrite done but I hope to have it finished by Monday. Thanksgiving buys me uninterrupted time in the EU :^)

docker/errors.go
mtrmac: Oops, GitHub is auto-collapsing files…

I don’t think it makes sense to have a generic interface with a specific error condition like that; compare `types.ManifestTypeRejectedError`, as an explicitly documented error case in the `PutManifest` method.

(Alternatively, it would be consistent to just say “this is totally docker+c/storage-specific, there will be no generic interfaces, and this is a `DockerGetBlobAt`, and so on. That’s probably not the best design, I expect `dir:` could easily implement `GetBlobAt`, for example.)

mtrmac:
> Another advantage of this PR is that it performs local deduplication as well. If a file is present locally, it is deduplicated with reflinks when the underlying file system supports them.

BTW if this is letting the creator of the image cause accesses to files on the host filesystem, this worries me a lot. It very likely allows a network observer to determine whether files exists at attacker-directed paths on the host filesystem, and in the extreme (but probably fixable) cases it risks hardware operations (IIRC some files in `/dev` cause hardware to do things (rewind tapes?) just on open(2)).

mtrmac:
> > Oh, and: how does digest verification happen with this, or what replaces it to obtain equivalent security?
> 
> this part is still not solved and where I'll need your help :-) Do you think it is reasonable to include signatures as part of the metadata (it can be a separate chunk)? Only the manifest must be signed.
> 
> Since the manifest has the checksum for each file in the layer, we can validate each file individually.

There needs to be some kind of “chain of authentication” going from the manifest (which is validated against a signature or user-specified digest) all the way to individual on-disk files. I really haven’t read the code; maybe that chain of authentication could be an annotation on the layer in the manifest (which would restrict this feature to manifest schemas that have annotations, but this is probably already OCI-only), with that annotation containing some kind of hash of the “top-level” metadata item in the layer (is there even such a thing?), which further authenticates the individual files / layer chunks, or some groups of individual files / layer chunks.

giuseppe:
> There needs to be some kind of “chain of authentication” going from the manifest (which is validated against a signature or user-specified digest) all the way to individual on-disk files. I really haven’t read the code; maybe that chain of authentication could be an annotation on the layer in the manifest (which would restrict this feature to manifest schemas that have annotations, but this is probably already OCI-only), with that annotation containing some kind of hash of the “top-level” metadata item in the layer (is there even such a thing?), which further authenticates the individual files / layer chunks, or some groups of individual files / layer chunks.

yes that is what I am proposing.

The manifest is a JSON file that looks like:

```
[
{"type":"DIR","name":"bin/","mode":16877,"size":0,"uid":0,"gid":0,"modTime":"2020-10-21T11:23:56+02:00","accessTime":"0001-01-01T00:00:00Z","changeTime":"0001-01-01T00:00:00Z","devMajor":0,"devMinor":0},
{"type":"SYMLINK","name":"bin/arch","linkName":"/bin/busybox","mode":41471,"size":0,"uid":0,"gid":0,"modTime":"2020-10-21T11:23:56+02:00","accessTime":"0001-01-01T00:00:00Z","changeTime":"0001-01-01T00:00:00Z","devMajor":0,"devMinor":0},
{"type":"REG","name":"bin/busybox","mode":33261,"size":841288,"uid":0,"gid":0,"modTime":"2020-06-16T13:05:54+02:00","accessTime":"0001-01-01T00:00:00Z","changeTime":"0001-01-01T00:00:00Z","devMajor":0,"devMinor":0,"checksum":"d62b5272e28473938cd6b8be12cc6260195e635968c1b02397b9e851702ccedb","startOffset":184,"endOffset":528310},
.......
]
```

so from a validated manifest we can easily validate each file in the image against their checksum for the payload and also its metadata.

giuseppe:
@ktock what do you think about this format?  Do you think it could be helpful for the remote snapshotter?

storage/chunked_zstd.go
```diff
@@ -0,0 +1,839 @@
...
+
+	for stream == nil {
+		stream, contentType, err = input.stream.GetBlobAt(input.ctx, input.srcInfo, chunksToRequest)
+		if err != nil {
+			if _, ok := err.(docker.ErrBadRequest); ok {
+				requested := len(missingChunks)
+				// If the server cannot handle at least 64 chunks in a single request, just give up.
+				if requested < 64 {
+					return err
+				}
```
ktock, line 526: Don't registries that don't support multi-part range request (e.g. gcr.io https://github.com/containerd/stargz-snapshotter/pull/81) always walks the worst code path? (i.e. everytime fails with log n retries)

ktock:
@giuseppe Thanks!

> One disadvantage of the stargz format is that a custom file is added to the tarball to store the layer metadata, and the metadata file is part of the image itself. Clients that are not aware of the stargz format will propagate the metadata file inside of the containers.

Could you elaborate this? Is this the problem from the security perspective?

ktock:
> There needs to be some kind of “chain of authentication” going from the manifest (which is validated against a signature or user-specified digest) all the way to individual on-disk files. I really haven’t read the code; maybe that chain of authentication could be an annotation on the layer in the manifest (which would restrict this feature to manifest schemas that have annotations, but this is probably already OCI-only), with that annotation containing some kind of hash of the “top-level” metadata item in the layer (is there even such a thing?), which further authenticates the individual files / layer chunks, or some groups of individual files / layer chunks.

eStargz exactly does this by leveraging layer annotations. https://github.com/containerd/stargz-snapshotter/blob/master/docs/verification.md

storage/chunked_zstd.go
giuseppe: thanks for letting me know about this issue.  I think I could raise the minimum number of chunks to minimize the worst code path in such cases.

giuseppe:
> @giuseppe Thanks!
> 
> > One disadvantage of the stargz format is that a custom file is added to the tarball to store the layer metadata, and the metadata file is part of the image itself. Clients that are not aware of the stargz format will propagate the metadata file inside of the containers.
> 
> Could you elaborate this? Is this the problem from the security perspective?

I don't think it is a security issue, at least not now with the information present in the stargz manifest file.  It just looks strange that the metadata must be part of the image itself, and that can be solved nicely with the zstd compression

ktock:
zstd:chunk can be applied to remote snapshotter but I think the way to verify the contents and the way to avoid the read overhead of on-demand fetching (e.g. [by prefetching](https://github.com/containerd/stargz-snapshotter/blob/master/docs/stargz-estargz.md)) will be needed like done by eStargz at least.

giuseppe:
> > 
> 
> eStargz exactly does this by leveraging layer annotations. https://github.com/containerd/stargz-snapshotter/blob/master/docs/verification.md

that is a great idea!  Thanks for it.  I'll use something similar for validating the manifest

manifest/common.go
```diff
@@ -63,7 +64,7 @@ func compressionVariantMIMEType(variantTable []compressionMIMETypeSet, mimeType
 			if mt == mimeType { // Found the variant
 				name := mtsUncompressed
 				if algorithm != nil {
-					name = algorithm.Name()
+					name = strings.Split(algorithm.Name(), ":")[0]
```
mtrmac, line 67: Is this really a _generic_ mechanism for MIME types? If not, I’d prefer 1) expressing this as table-driven data, or 2) an explicit exception for that particular pair of strings, not introducing a generic code behavior with unspecified semantics.

(I know, it is far too early to discuss such details.)

storage/chunked_zstd.go
mtrmac: Also consider the recent c/storage feature to force a single mode and store metadata in extended attributes.

⚠️ This _really must_ be implemented in c/storage and be subject to c/storage configuration (although I have not read nearly enough code to have an opinion on the c/storage - c/image interface.)

mtrmac:
> The manifest is a JSON file that looks like:
> …
> so from a validated manifest we can easily validate each file in the image against their checksum for the payload and also its metadata.

That makes sense.

giuseppe:
I've moved the zstd:chunked implementation to containers/storage and implemented the manifest validation using an OCI annotation.

ktock:
> > eStargz exactly does this by leveraging layer annotations. https://github.com/containerd/stargz-snapshotter/blob/master/docs/verification.md


> that is a great idea! Thanks for it. I'll use something similar for validating the manifest

@giuseppe We are planning to restart the lazy pull implementation (#644). I think we can support both of eStargz (for compatibility) & zstd:chunked (for faster decompression) for lazy pulling. 

Currently zstd:chunked's Manifest structure & verification logic is very similar to eStargz, so can we use TOC as the Manifest here? This would be much helpful for us to implement lazy pulling in podman.
https://github.com/containerd/stargz-snapshotter/blob/b1920d367a9aba1016da3c5548cf634f95f138c5/docs/stargz-estargz.md#toc-tocentries-and-footer

giuseppe:
> @giuseppe We are planning to restart the lazy pull implementation (#644). I think we can support both of eStargz (for compatibility) & zstd:chunked (for faster decompression) for lazy pulling.
> 
> Currently zstd:chunked's Manifest structure & verification logic is very similar to eStargz, so can we use TOC as the Manifest here? This would be much helpful for us to implement lazy pulling in podman.
> https://github.com/containerd/stargz-snapshotter/blob/b1920d367a9aba1016da3c5548cf634f95f138c5/docs/stargz-estargz.md#toc-tocentries-and-footer

sure, I think we can reuse the stargz TOC.

giuseppe:
> sure, I think we can reuse the stargz TOC.

I've changed the implementation to use the same metadata as the stargz TOC: https://github.com/containers/storage/blob/3ce2f867b8865b059e909c4a8abb65b3362fd1c4/pkg/zstdchunked/compression.go#L19-L41

ktock:
Thanks!

rhatdan:
@giuseppe Needs a rebase.

rhatdan:
@giuseppe needs a rebase.

cgwalters:
Regarding deduplication, I think it generally makes sense only to do it across an "image stream", i.e. given base image `registry.access.redhat.com/ubi8/ubi:8.3-289:8.3-289` and `registry.access.redhat.com/ubi8/ubi:8.2-299` we only try to dedup inside a storage area namespaced/scoped to `registry.access.redhat.com/ubi8/ubi`.  

The common case is that the host will be different than the containers too.

rhatdan:
In Openshift, the common case would be running RHEL containers on a RHEL Host, so why wouldn't we want to take advantage of this at least for the content in /usr?

cgwalters:
> In Openshift, the common case would be running RHEL containers on a RHEL Host, so why wouldn't we want to take advantage of this at least for the content in /usr?

Sure but it is likely that the host is newer, while the container images are older.  At some point in the life of OpenShift 4 for example the host will jump to rhel9, but that doesn't have anything to do with users' apps.

rhatdan:
Understood, but we still could get some improvements in performance and still take advantage of the the content in containers/storage.

cgwalters:
I am sure that pulling content from the host makes for a cool demo, particularly the first time a container is pulled, and when you set things up so the host and container are roughly the same.  But my argument is that in particular for most Kubernetes use cases that's not the common case.

So I think for most users it'd benefit more to reuse content from prior container layers (and mostly base layers).  If we really want the host bit, one way to do it would be to build a local image from some key selected host parts and use that as a dedup base.

cgwalters:
One other random argument against opportunistic path-based host content reuse like this is it will pollute the Linux kernel negative dentry cache (xref https://lwn.net/Articles/1511/ https://lwn.net/Articles/814535/ etc.).  It's going to be better to generate the index in advance.

Also another further argument against this is it's subject to race conditions because this code is doing "compute checksum, then if it works copy".  So if e.g. I'm hacking on my operating system host and do something like `sudo make install` over the running host `/usr` while something else is pulling a container image, unless that process doing the install is careful to do the "create new file then rename()" pattern instead of just `cp` then you can get incorrect binaries in the container too.

This race is a lot more likely on a non-image-based (e.g. non-ostree) system that doesn't have a default read-only bind mount over `/usr`.

pkg/compression/internal/types.go
```diff
@@ -35,6 +37,11 @@ func (c Algorithm) Name() string {
 	return c.name
 }
 
+// Name returns the MIME type to use for the compression algorithm.
+func (c Algorithm) MIME() string {
```
mtrmac, line 41: The returned values are clearly not MIME types; this needs accurately thinking through what we are actually doing. What is the _precise_ semantics of the various fields, and the relationship between `zstd` and `zstd:chunked`?

Are they two completely distinct file formats (so that `DetectCompression` must differentiate between them?), two related file formats where on is-a the other (so that we can conservatively almost always use only one of them), or is “chunked” maybe just a compressor algorithm option like `types.SystemContext.CompressionLevel` that _only_ matters for creating streams? Something else?

pkg/compression/zstd.go
```diff
@@ -1,16 +1,21 @@
...
 import (
+	"fmt"
 	"io"
 
 	"github.com/klauspost/compress/zstd"
 )
 
 type wrapperZstdDecoder struct {
 	decoder *zstd.Decoder
+	closed  bool
```
mtrmac, line 12: This is never set

pkg/compression/zstd.go
```diff
@@ -40,13 +48,13 @@ func zstdWriter(dest io.Writer) (io.WriteCloser, error) {
...
 
-func zstdWriterWithLevel(dest io.Writer, level int) (io.WriteCloser, error) {
+func zstdWriterWithLevel(dest io.Writer, level int) (*zstd.Encoder, error) {
 	el := zstd.EncoderLevelFromZstd(level)
 	return zstd.NewWriter(dest, zstd.WithEncoderLevel(el))
 }
 
 // zstdCompressor is a CompressorFunc for the zstd compression algorithm.
-func zstdCompressor(r io.Writer, level *int) (io.WriteCloser, error) {
+func zstdCompressor(r io.Writer, metadata map[string]string, level *int) (io.WriteCloser, error) {
```
mtrmac, line 57: `metadata` is never used?

storage/storage_image.go
```diff
@@ -413,6 +418,12 @@ func (s *storageImageDestination) Reference() types.ImageReference {
 
 // Close cleans up the temporary directory.
 func (s *storageImageDestination) Close() error {
+	for _, v := range s.diffOutputs {
+		if v.Target != "" {
+			os.RemoveAll(v.Target)
```
mtrmac, line 423: Are these c/image-owned temporary files? If so, it would be simpler to have them in `s.directory`

Are they c/storage-owned? If so, shouldn’t there be an API so that c/storage can manage its own locking and metadata and post-sudden-reboot cleanup? Just removing files managed by another codebase feels very risky.

Are they something else?

storage/storage_image.go
```diff
@@ -540,6 +576,39 @@ func (s *storageImageDestination) TryReusingBlobWithOptions(ctx context.Context,
...
+func (s *storageImageDestination) PutBlobPartial(ctx context.Context, stream types.ImageSourceSeekable, srcInfo types.BlobInfo, cache types.BlobInfoCache) (types.BlobInfo, error) {
+	fetcher := zstdFetcher{
+		stream:   stream,
+		ctx:      ctx,
+		blobInfo: srcInfo,
+	}
+
+	// While the PutBlobPartial() interface is generic, at this point only one format is supported (zstd:chunked), so  get a
+	// zstd:chunked differ and apply the diff using it.
+	differ, err := zstdchunked.MakeZstdChunkedDiffer(ctx, s.imageRef.transport.store, srcInfo.Size, srcInfo.Annotations, &fetcher)
```
mtrmac, line 590: How is this actually going to work in the future with multiple formats? I guess based on `srcInfo.MediaType`, but how _exactly_?
- (Is that field even consistently filled in at this point?)
- `MediaType` is manifest-format-specific. So do we actually need some kind of mapping or reuse of data from `c/image/manifest` to turn `MediaType` into a known format? Is this something this function can handle itself, or something the copy pipeline must provide? I’m digging into the weeds here because this probably impacts the “what is `zstd:chunked`, is it a compression algorithm/variant/neither” discussion elsewhere.

mtrmac:
^^^ is just a random quick skim again.

---

> One other random argument against opportunistic path-based host content reuse like this

I’m really leaning towards “this feels unsafe and should not go in unless we can build a _comprehensive argument why it is safe_ (and we probably can’t in the general case)”. I appreciate that that it demoes well, and the condition is a very high bar to clear — but it’s all too easy to _speculate_ about ways that this could possibly backfire (open a device file that does something as a side effect, like rewind a type or start a watchdog? flood audit logs by intentionally causing millions of SELinux policy denials) and it’s not at all obvious to me how this can be safe.

We don’t leave around attacker-initiated reads of unowned memory with “well we don’t have a specific reproducer how the attacker can abuse this”, and I it’s not clear to me why we shouldn’t treat attacker-initiated reads of unowned files differently.

giuseppe:
> I’m really leaning towards “this feels unsafe and should not go in unless we can build a _comprehensive argument why it is safe_ (and we probably can’t in the general case)”. I appreciate that that it demoes well, and the condition is a very high bar to clear — but it’s all too easy to _speculate_ about ways that this could possibly backfire (open a device file that does something as a side effect, like rewind a type or start a watchdog? flood audit logs by intentionally causing millions of SELinux policy denials) and it’s not at all obvious to me how this can be safe.
> 
> We don’t leave around attacker-initiated reads of unowned memory with “well we don’t have a specific reproducer how the attacker can abuse this”, and it’s not clear to me why we **EDIT shouldn’t should** treat attacker-initiated reads of unowned files differently.

it is true that the current implementation currently suffers from a race condition and if the source file is truncated or modified in place, that won't be detected while copying the file.  I will address it by calculating the checksum again after the file is copied for now.  Longer term plan is to create  the index before and not be limited to the lookup by the same path.
Said so, the host deduplication is currently limited to regular files under `/usr`.  I am fine if we restrict it even further but I don't think we should drop this optimization completely as in some cases it turns very useful (pull fedora:34 on a fedora:34 host skips ~90% of the files to be downloaded)

pkg/compression/zstd.go
giuseppe: this is just a preparatory work for https://github.com/containers/storage/pull/775/commits/22a53043dda44ef3f10289c7d2b95ee0dbc3cc65#diff-a32635dc11fd91356384da8c2a47d25c8b278bbd5182eaecd8d2ccaba3d1b670R471-R478

pkg/compression/internal/types.go
giuseppe: they are two related file formats.  The zstd decompressor can handle `zstd:chunked` but it loses some of its features.

In future we'd probably need the same differentiation between gzip and stargz.  Where gzip can handle the stargz format but it won't handle its additional features

pkg/compression/compression.go
```diff
@@ -69,26 +73,33 @@ func XzDecompressor(r io.Reader) (io.ReadCloser, error) {
...
 
 // CompressStream returns the compressor by its name
 func CompressStream(dest io.Writer, algo Algorithm, level *int) (io.WriteCloser, error) {
-	return internal.AlgorithmCompressor(algo)(dest, level)
+	m := map[string]string{}
+	return internal.AlgorithmCompressor(algo)(dest, m, level)
+}
+
+// CompressStreamWithMetadata returns the compressor by its name.  If the compression
+// generates any metadata, it is written to the provided metadata map.
```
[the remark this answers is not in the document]
giuseppe: the metadata generated here will end up in an annotation.

I am fine to specify it, although `pkg/compression` doesn't seem the best place for it.  Do you have any suggestions where it should be documented?

docker/errors.go
```diff
@@ -32,11 +33,15 @@ func httpResponseToError(res *http.Response, context string) error {
 	switch res.StatusCode {
 	case http.StatusOK:
 		return nil
+	case http.StatusPartialContent:
+		return nil
```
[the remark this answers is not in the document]
giuseppe: the HTTP status 206 is used only when a range request is performed.  Other callers should not be affected (unless we worry about broken HTTP servers?)

docker/docker_image_src.go
```diff
@@ -276,6 +276,41 @@ func (s *dockerImageSource) HasThreadSafeGetBlob() bool {
...
+// GetBlobAt returns a stream for the specified blob.
+func (s *dockerImageSource) GetBlobAt(ctx context.Context, info types.BlobInfo, chunks []types.ImageSourceChunk) (io.ReadCloser, string, error) {
+	headers := make(map[string][]string)
+
+	var rangeVals []string
+	for _, c := range chunks {
+		rangeVals = append(rangeVals, fmt.Sprintf("%d-%d", c.Offset, c.Offset+c.Length-1))
+	}
+
+	headers["Range"] = []string{fmt.Sprintf("bytes=%s", strings.Join(rangeVals, ","))}
```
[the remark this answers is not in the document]
giuseppe: the multipart MIME is not strictly HTTP (at least RFC 2046 doesn't refer to HTTP at all).

The thing I am worried the most about is that without buffering, mimepart slows significantly down the pull.

I'll try to fix it and use a goroutine to handle the multipart and return a `chan io.ReadCloser`

pkg/compression/internal/types.go
mtrmac: Yes; what I’d like to see is a precise modeling and correct naming of the individual functions/values — so that any future updates to the code can tell what was the purpose, and what are the invariants that need to be maintained, without understanding the complete codebase.

(Or a very clear one-time exception — the original version had a `strings.Split(algorithm.Name(), ":")` somewhere in the compression/MIME mapping code. The concern with that was essentially the same as with the above: introducing a data model with undocumented and unclear semantics — and at that place a specific `if algorithm.Name() == "zstd:chunked"` was one of the options discussed.

While exceptions are worse than an ideal clean code that can predict future needs, it’s fine if we don’t know what is a good long-term model, and make a very explicit, clearly delineated, and reasonably documented, one-time exception.)

pkg/compression/compression.go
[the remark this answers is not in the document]
mtrmac: We don’t necessarily have to specify the _format_ here — we don’t carry a full specification of Zstd either.

We do need a “how would the caller ever use this?” so that there even can be a legitimate caller. Maybe something vaguely like:
> CompressStreamWithIntegrityData returns … . Compressing a stream will create integrity data that [is optional to  | must be] provided on decompression along with the compressed byte stream via $some_API, and that will allow the decompression to only use subsets of the compressed data (if the compressed data is seekable and most of the uncompressed data is already present via other means), _while still protecting integrity of the compressed stream against unwanted modification_. (In OCI container images, this metadata is usually carried in manifest annotations.)

That’s very likely not accurate, but it captures:
- How is the data used by the caller
- What is the benefit / consequence of not providing it
- Purpose/Applicability outside this narrow “compression” subpackage.

docker/errors.go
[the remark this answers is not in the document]
mtrmac: Yes, we worry about broken HTTP servers that return 206 when they should have returned 200. More importantly, we worry about HTTP servers that completely ignore the range header and return 200 when they should have returned 206 — so the `GetBlobAt` code should, I think, special-case 206 in either case.

docker/docker_image_src.go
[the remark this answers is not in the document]
mtrmac: Works for me — I think worrying about buffering and the like is a bit of a better fit than doing that in c/storage. c/image/docker knows it’s talking to the network and that there are roundtrip latencies involved.

giuseppe:
pushed a new revision but still not happy with the `GetBlobAt` API, I'll make another iteration at it

cgwalters:
In OpenShift 4 today we have a "bootstrap" node that does a lot of stuff to pull up the 3 control plane nodes.  The bootstrap node is always thrown away at the end.  When you have a cluster, you *never* have to think about "what's the state of the bootstrap" because you know it's gone.

Someone once called this a "debugging firewall" in a meeting.  The bootstrap doesn't matter once it's done and you won't need to debug it.

Before this PR, pulling containers had zero interaction with the host.  Similarly, performing host updates could (should) have zero interaction with containers.

This PR is poking a hole in that "debugging firewall" - that's not something to take lightly.  The benefit should be broad and compelling.  And we need to put in as many mechanisms as possible to mitigate risk.

Ironically I am also approaching this "host/container storage" angle in https://github.com/containers/image/issues/1209 as you've probably seen (I started looking at other PRs after filing that to see what else was going on, and found this one).

Note in that issue I am very intentionally trying to preserve that "debugging firewall" for the host updates - we're just using skopeo as a data pipe, and I am going to make absolutely sure it only has read-only access to the host for *exactly* the kind of reasons discussed here.  I don't want to find out at some point a patch to skopeo (or c/image,c/storage) started modifying a file in `/run` that somehow affects a concurrent container image operation, etc.

storage/storage_image.go
giuseppe: added a new API to c/storage to deal with the cleanup

storage/storage_image.go
giuseppe: moved the logic to c/storage

giuseppe:
I've made the "host deduplication" part configurable through the `storage.conf` file and it is disabled by default.

rhatdan:
That seems like a decent compromise.

giuseppe:
I'll deal with the portability issues in c/storage and remove all the glue code I needed to add to fix build on `!linux`

giuseppe:
what are the blocking issues to deal with before we can merge?

giuseppe:
can we move this forward?

docker/errors.go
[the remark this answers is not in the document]
rhatdan: @giuseppe What about this comment?

pkg/compression/compression.go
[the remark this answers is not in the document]
rhatdan: @giuseppe Have  you fixed this comment?

pkg/compression/internal/types.go
rhatdan: @giuseppe Any response to this comment?

types/types.go
rhatdan: Have you made it private?

giuseppe: it is used by the storage package, so I cannot make it private.

I've marked all the new interface as experimental though, so we don't need to incement the major version if we make any breaking change.

mtrmac: (Random drive-by comment, not part of a full review)

> it is used by the storage package, so I cannot make it private.

So there’s a circular dependency? Historically that’s been rather problematic. (For starters, how is either of the two PRs going to actually pass CI before merging?)

> I've marked all the new interface as experimental though, so we don't need to incement the major version if we make any breaking change.

(I’m skeptical that comments matter that much. Go is going to upgrade and break things without notice. It is possible to copy&paste code without noticing documentation. Yet, we do need some way to experiment… A subpackage with a hard-to-overlook name (c/image/types/unstable?) would be a bit more explicit.)

---

I think there should be:
- A type defined somewhere in c/storage , so that there isn’t a loop, and so that c/storage itself can at compile-time enforce interface conformance. (There already is one)
- A completely private type in c/image , in the already-existing c/image/internal/types , that no external callers have business touching. We want to, over time, regain the ability to change the transports interfaces, so let’s not expose any more unless we have to. (We might need an explicit adapter between the two interfaces if they use different names/types; that’s perfectly fine.)

A possible major difficulty with this is Buildah’s (and Podman’s??) alternative/wrapping transport implementations , notably the blob cache; the above would prevent Buildah’s blob cache from using the chunked variant. That’s probably? fine, the thoughts about making transports private have always implied moving the blob cache inside c/image so that it does not break on interface changes, and can benefit from private interfaces. OTOH that blob cache move might then have to happen before this PR.

giuseppe:
> I'll deal with the portability issues in c/storage and remove all the glue code I needed to add to fix build on `!linux`

rebased and dropped the portability workarounds since the issue was fixed in c/storage

types/types.go
giuseppe: > > it is used by the storage package, so I cannot make it private.
> 
> So there’s a circular dependency? Historically that’s been rather problematic. (For starters, how is either of the two PRs going to actually pass CI before merging?)

no, with storage package I meant c/image/storage.

I try to move these definitions under `internal/types`

giuseppe:
opened a PR to add zstd support to quay: https://github.com/quay/quay/pull/801

giuseppe:
I've moved all the new types under `internal/types`

giuseppe:
rebased

giuseppe:
can we move this forward?

giuseppe:
@mtrmac PTAL

giuseppe:
@vrothberg PTAL

vrothberg (commented):
Overall LGTM. Really nice work, @giuseppe!

Pretty much all of my comments relate to adding a bit more comments to leave more breadcrumbs for reader.

One thing I noticed while playing with the PR is the progress bar, where I would like to change

``
Copying blob df748f8d6e55                        .... 81.8MiB / 81.8MiB (skipped: 219.6KiB = 0.26%)  
Copying config 7273426b03 done                                                                  
```
... to ...
```
Copying blob df748f8d6e55 done (skipped: 219.6KiB = 0.26%)  
Copying config 7273426b03 done                                                                  
```

File: copy/copy.go [hunk 22]
```diff
@@ -1154,6 +1202,23 @@ type diffIDResult struct {
 	err    error
 }
 
+type imageSourceSeekableProxy struct {
```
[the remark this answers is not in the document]
mtrmac: Also maybe move this to `progress_reader.go`, `copy.go` is getting too large.

copy/copy.go
```diff
@@ -1640,21 +1744,34 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
...
 
-	compressor, err := compression.CompressStream(dest, compressionFormat, c.compressionLevel)
+	compressor, err := compression.CompressStreamWithMetadata(dest, metadata, compressionFormat, c.compressionLevel)
 	if err != nil {
 		return
 	}
-	defer compressor.Close()
+	defer func() {
+		err2 := compressor.Close()
+		if err == nil {
```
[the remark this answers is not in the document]
mtrmac: Compare `diffIDComputationGoroutine`, written because I’m afraid of mistakes in this area:

Have the `…Goroutine` function deal with the goroutine state and enforcing the communication protocol (that an error is always sent via `dest`), make it as trivial as possible to make the communication reliable; move any non-trivial operations and control flow to a separate function that doesn’t have to care about goroutines.

giuseppe:
thanks for the review.  Addressed the comments and pushed a new version

giuseppe:
also found an issue in the storage library that prevented deduplication from other layers: https://github.com/containers/storage/pull/954

vrothberg:
It seems like the progress bar is not cleared. Can others reproduce?
```
Copying blob 8f75b7c47cae [======================================] 82.2MiB / 82.2MiB (skipped: 203.1KiB = 0.24%)
Copying config 0a485308c4 done                                                                                  
Storing signatures                                                                                              
```

giuseppe:
> It seems like the progress bar is not cleared. Can others reproduce?

it might be a side effect of using `customPartialBlobCounter`.  Although, how would we know how much was saved if we destroy the bar?

vrothberg:
I'd imagine it to look similar to that:
```
Copying blob df748f8d6e55 done (skipped: 219.6KiB = 0.26%)  
Copying config 7273426b03 done                                                                  
```

giuseppe:
fixed in the last version

giuseppe:
rebased again.  @vrothberg are you ok with the last version of the progress bar?

vrothberg:
@giuseppe, I am cool to merge as is (see below), thanks!  The race is something I wanted to tackle for a long time but never found/took the time until now.

```
skopeo (main) $ ./bin/skopeo copy docker://docker.io/gscrivano/zstd-chunked:fedora3 containers-storage:fedora-chunke
d
INFO[0000] Not using native diff for overlay, this may cause degraded performance for building images: opaque flag erroneously copied up, co
nsider update to kernel 4.8 or later to fix 
Copying blob 8f75b7c47cae [======================================] 82.0MiB / 82.2MiB (skipped: 0.0b = 0.00%)
Copying config 0a485308c4 done  
Storing signatures
skopeo (main) $ ./bin/skopeo copy docker://docker.io/gscrivano/zstd-chunked:fedora3 containers-storage:fedora-chunked
INFO[0000] Not using native diff for overlay, this may cause degraded performance for building images: opaque flag erroneously copied up, co
nsider update to kernel 4.8 or later to fix 
Copying blob 8f75b7c47cae [--------------------------------------] 0.0b / 0.0b
Copying config 0a485308c4 done  
Storing signatures
```

giuseppe:
So anything more left before we can finally merge this?  :)

giuseppe:
@rhatdan are you fine with it?

rhatdan:
LGTM

storage/storage_image.go
```diff
@@ -844,6 +908,27 @@ func (s *storageImageDestination) commitLayer(ctx context.Context, blob manifest
...
+	s.lock.Lock()
+	diffOutput, ok := s.diffOutputs[blob.Digest]
+	s.lock.Unlock()
+	if ok {
+		layer, err := s.imageRef.transport.store.CreateLayer(id, lastLayer, nil, "", false, nil)
+		if err != nil {
+			return err
+		}
+
+		// FIXME: what to do with the uncompressed digest?
```
mtrmac, line 920: @giuseppe Why is this (and the related `s.blobDiffIDs[blobDigest] = blobDigest` assignment), using the compressed values as the supposedly-unique uncompressed ones, correct and safe?

storage/chunked_zstd.go
mtrmac: @giuseppe years later, https://github.com/containers/storage/commit/3bbd19c9dcc0693bd054c51598547b92e52bb811 now loops until `requested == 1`.

Is that something to worry about, or are the registries well-behaved enough that this is fine?

giuseppe: @mtrmac I've not seen that behavior with quay.io or docker.io, but hard to say for other registries.

Given that we care more about use cases like composefs now, probably the tradeoff shifted more towards "try harder to use partial images" than it was 4 years ago when it was just an optimization.

@ktock is there any estargz image on `gcr.io` that I can use?

mtrmac: (Just to be explicit, I have no opinion on the trade-off and I’m happy to defer to experts; I have just happened to come upon this discussion, and remembered the recent code change.)

pkg/compression/compression.go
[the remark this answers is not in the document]
mtrmac: 3 years later, this will be resolved in https://github.com/containers/image/pull/2605 .

PR #1085, opened 2020-11-19
Use storage/pkg/homedir rather then docker version

rhatdan:
The Docker version only checks for $HOME being set,
the storage version falls back to looking in the /etc/passwd
file for the current UID running the command "RootlessUID"

### DIFF
--- a/pkg/sysregistriesv2/shortnames.go
+++ b/pkg/sysregistriesv2/shortnames.go
@@ -8,8 +8,8 @@ import (
 	"github.com/BurntSushi/toml"
 	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/image/v5/types"
+	"github.com/containers/storage/pkg/homedir"
 	"github.com/containers/storage/pkg/lockfile"
-	"github.com/docker/docker/pkg/homedir"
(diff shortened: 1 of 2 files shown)


giuseppe (approved):
LGTM

pkg/sysregistriesv2/shortnames.go
```diff
@@ -8,8 +8,8 @@ import (
 	"github.com/BurntSushi/toml"
 	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/image/v5/types"
+	"github.com/containers/storage/pkg/homedir"
```
TomSweeneyRedHat, line 11: Do we have this anywhere else that needs to be cleansed?

rhatdan: Not in here.

rhatdan: Once this is cleaned up then we need to vendor new versions into buildah as well.

QiWang19 (commented):
LGTM, but tests are failing.

rhatdan:
The test failures are not related to this PR, so I am going to merge.  To move this along.

PR #1088, opened 2020-11-23
copy: check our assumptions about compression

nalind:
When copying a blob where we know to associate a compression algorithm with its MIME type, check if the blob was in fact compressed using that algorithm, and log a diagnostic message at level "debug" if it isn't.  Our local storage backend is less sensitive to this than others, but it's useful to me at least to be able to tell.  I'm not too sure about just adding a table for it, though.

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -53,6 +53,14 @@ var (
 // compressionBufferSize is the buffer size used to compress a blob
 var compressionBufferSize = 1048576
 
+// expectedCompressionFormats is used to check if a blob with a specified media type is compressed
+// using the algorithm that the media type says it should be compressed with
+var expectedCompressionFormats = map[string]*compression.Algorithm{
+	imgspecv1.MediaTypeImageLayerGzip:      &compression.Gzip,
+	imgspecv1.MediaTypeImageLayerZstd:      &compression.Zstd,
+	manifest.DockerV2Schema2LayerMediaType: &compression.Gzip,
+}
+
 // newDigestingReader returns an io.Reader implementation with contents of source, which will eventually return a non-EOF error
 // or set validationSucceeded/validationFailed to true if the source stream does/does not match expectedDigest.
 // (neither is set if EOF is never reached).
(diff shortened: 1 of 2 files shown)


copy/copy.go
```diff
@@ -1234,6 +1242,10 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
 	isCompressed := decompressor != nil
 	destStream = bar.ProxyReader(destStream)
 
+	if expectedCompressionFormat, known := expectedCompressionFormats[srcInfo.MediaType]; known && compressionFormat.Name() != expectedCompressionFormat.Name() {
+		logrus.Debugf("blob %s with type %s should be compressed with %s, but compressor appears to be %s", srcInfo.Digest.String(), srcInfo.MediaType, compressionFormat.Name(), expectedCompressionFormat.Name())
```
rhatdan, line 1246: Should this be Warnf?

nalind: Maybe?

nalind:
Fixed the order of compression types in the debug message being wrong, and "knowing" gzip for foreign layers in Schema 2 instead of regular layers.

vrothberg (approved):
LGTM

copy/copy.go
```diff
@@ -1234,6 +1242,10 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
 	isCompressed := decompressor != nil
 	destStream = bar.ProxyReader(destStream)
 
+	if expectedCompressionFormat, known := expectedCompressionFormats[srcInfo.MediaType]; known && compressionFormat.Name() != expectedCompressionFormat.Name() {
```
mtrmac, line 1245: The `compressionFormat` value is invalid if `!isCompressed` (the `Name` call won’t crash given the current implementation, but calling `CompressStream` will); so `!isCompressed` needs to be checked first.

(And it turns out `DetectCompressionFormat` is under-documented.)

mtrmac (commented):
I’m not quite sure that this is worth maintaining either, but you did find it useful, and it fails “open” if new algorithms or MIME types are added, so … why not.

copy/copy.go
nalind: Added a check for `isCompressed` before calling at `compressionFormat.Name()`, and tweaked the godoc for `DetectCompressionFormat` to describe the returned Algorithm value.

mtrmac (approved):
Thanks! LGTM.

copy/copy.go
mtrmac: Either would make sense.

_Right now_, I’m thinking “What’s the user going to do about it?”
- If the resulting build/copy/use pipeline _works_, the user is not very motivated to do something about that warning, it’s just going to be ignored.
- If the build/copy/use pipeline doesn’t work, the warning could be useful for diagnosis, but (assuming the wrong MIME types are not created by the GitHub.com/containers codebase, which is a bit of a hubris, sure), really the debugging process will start with the error reported by the consumer, not by an optional copy tool in the middle.
- We AFAIK provide no tools/process to recover from such a mismatch, apart maybe from some pull/push sequences that might repair the MIME types purely by side effect, not as an intentional maintained/tested feature.

So in my view this is closer to the “most users won’t care or can’t act on it” side, i.e. debug logs, than to the “most users need to do something definitve about this with some urgency”, i.e. a public warning. But that’s a very weak judgement and opinions can certainly differ.

giuseppe (approved):
LGTM

PR #1089, opened 2020-11-23
blobinfocache: cache compression types

### LINKED ISSUE #733: wrong mediatypes when blob is already present at destination

nalind:
Extend the blob info cache to also cache the name of the type of compression used on a blob in a known location.  (An earlier version of this patch cached the full MIME type, but required additional logic to map the MIME types in cases where we converted images during copying, so the patch was larger.)

Pass an updated MIME type and information about compression changes to PutBlob() in its input BlobInfo, so that it can populate the blob info cache with correct compression information when it succeeds.  Some of the local tables we kept in `manifest/common.go` were moved to an internal package (`internal/mediatypes`) so that they could be reused there.

Make sure that when `dockerImageDestination.TryReusingBlob()` successfully uses a blob from the blob info cache, that it provides updated MIME type and compression information in the `BlobInfo` that it returns, so that converted manifests can use them to correctly update their `LayerInfo` fields.

This should fix #733.

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/image/v5/image"
+	internalblobinfocache "github.com/containers/image/v5/internal/blobinfocache"
 	"github.com/containers/image/v5/internal/pkg/platform"
 	"github.com/containers/image/v5/manifest"
 	"github.com/containers/image/v5/pkg/blobinfocache"
@@ -113,7 +114,7 @@ type copier struct {
 	progressOutput    io.Writer
 	progressInterval  time.Duration
 	progress          chan types.ProgressProperties
-	blobInfoCache     types.BlobInfoCache
+	blobInfoCache     internalblobinfocache.BlobInfoCache2
 	copyInParallel    bool
 	compressionFormat compression.Algorithm
 	compressionLevel  *int
@@ -265,7 +266,7 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
 		// FIXME? The cache is used for sources and destinations equally, but we only have a SourceCtx and DestinationCtx.
 		// For now, use DestinationCtx (because blob reuse changes the behavior of the destination side more); eventually
 		// we might want to add a separate CommonCtx — or would that be too confusing?
-		blobInfoCache:    blobinfocache.DefaultCache(options.DestinationCtx),
+		blobInfoCache:    internalblobinfocache.FromBlobInfoCache(blobinfocache.DefaultCache(options.DestinationCtx)),
 		ociDecryptConfig: options.OciDecryptConfig,
 		ociEncryptConfig: options.OciEncryptConfig,
 	}
@@ -648,13 +649,19 @@ func (c *copier) copyOneImage(ctx context.Context, policyContext *signature.Poli
 	// With docker/distribution registries we do not know whether the registry accepts schema2 or schema1 only;
 	// and at least with the OpenShift registry "acceptschema2" option, there is no way to detect the support
 	// without actually trying to upload something and getting a types.ManifestTypeRejectedError.
-	// So, try the preferred manifest MIME type. If the process succeeds, fine…
+	// So, try the preferred manifest MIME type with possibly-updated blob digests, media types, and sizes if
+	// we're altering how they're compressed.  If the process succeeds, fine…
 	manifestBytes, retManifestDigest, err := ic.copyUpdatedConfigAndManifest(ctx, targetInstance)
 	retManifestType = preferredManifestMIMEType
 	if err != nil {
 		logrus.Debugf("Writing manifest using preferred type %s failed: %v", preferredManifestMIMEType, err)
-		// … if it fails, _and_ the failure is because the manifest is rejected, we may have other options.
-		if _, isManifestRejected := errors.Cause(err).(types.ManifestTypeRejectedError); !isManifestRejected || len(otherManifestMIMETypeCandidates) == 0 {
+		// … if it fails, and the failure is either because the manifest is rejected by the registry, or
+		// because we failed to create a manifest of the specified type because the specific manifest type
+		// doesn't support the type of compression we're trying to use (e.g. docker v2s2 and zstd), we may
+		// have other options available that could still succeed.
+		_, isManifestRejected := errors.Cause(err).(types.ManifestTypeRejectedError)
+		_, isCompressionIncompatible := errors.Cause(err).(manifest.ManifestLayerCompressionIncompatibilityError)
+		if (!isManifestRejected && !isCompressionIncompatible) || len(otherManifestMIMETypeCandidates) == 0 {
 			// We don’t have other options.
 			// In principle the code below would handle this as well, but the resulting  error message is fairly ugly.
 			// Don’t bother the user with MIME types if we have no choice.
@@ -896,7 +903,7 @@ func (ic *imageCopier) copyLayers(ctx context.Context) error {
 	return nil
 }
 
-// layerDigestsDiffer return true iff the digests in a and b differ (ignoring sizes and possible other fields)
+// layerDigestsDiffer returns true iff the digests in a and b differ (ignoring sizes and possible other fields)
 func layerDigestsDiffer(a, b []types.BlobInfo) bool {
 	if len(a) != len(b) {
 		return true
@@ -951,7 +958,7 @@ func (ic *imageCopier) copyUpdatedConfigAndManifest(ctx context.Context, instanc
 		instanceDigest = &manifestDigest
 	}
 	if err := ic.c.dest.PutManifest(ctx, man, instanceDigest); err != nil {
-		return nil, "", errors.Wrap(err, "Error writing manifest")
+		return nil, "", errors.Wrapf(err, "Error writing manifest %q", string(man))
 	}
 	return man, manifestDigest, nil
 }
@@ -1058,6 +1065,12 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, to
 
 	// If we already have the blob, and we don't need to compute the diffID, then we don't need to read it from the source.
 	if !diffIDIsNeeded {
+		// TODO: at this point we don't know whether or not a blob we end up reusing is compressed using an algorithm
+		// that is acceptable for use on layers in the manifest that we'll be writing later, so if we end up reusing
+		// a blob that's compressed with e.g. zstd, but we're only allowed to write a v2s2 manifest, this will cause
+		// a failure when we eventually try to update the manifest with the digest and MIME type of the reused blob.
+		// Fixing that will probably require passing more information to TryReusingBlob() than the current version of
+		// the ImageDestination interface lets us pass in.
 		reused, blobInfo, err := ic.c.dest.TryReusingBlob(ctx, srcInfo, ic.c.blobInfoCache, ic.canSubstituteBlobs)
 		if err != nil {
 			return types.BlobInfo{}, "", errors.Wrapf(err, "Error trying to reuse blob %s at destination", srcInfo.Digest)
@@ -1253,16 +1266,23 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
 		originalLayerReader = destStream
 	}
 
-	desiredCompressionFormat := c.compressionFormat
-
 	// === Deal with layer compression/decompression if necessary
 	var inputInfo types.BlobInfo
 	var compressionOperation types.LayerCompression
+	uploadCompressionFormat := &c.compressionFormat
+	srcCompressorName := internalblobinfocache.Uncompressed
+	if isCompressed {
+		srcCompressorName = compressionFormat.Name()
+	}
+	var uploadCompressorName string
 	if canModifyBlob && isOciEncrypted(srcInfo.MediaType) {
 		// PreserveOriginal due to any compression not being able to be done on an encrypted blob unless decrypted
 		logrus.Debugf("Using original blob without modification for encrypted blob")
 		compressionOperation = types.PreserveOriginal
 		inputInfo = srcInfo
+		srcCompressorName = internalblobinfocache.UnknownCompression
+		uploadCompressorName = internalblobinfocache.UnknownCompression
+		uploadCompressionFormat = nil
 	} else if canModifyBlob && c.dest.DesiredLayerCompression() == types.Compress && !isCompressed {
 		logrus.Debugf("Compressing blob on the fly")
 		compressionOperation = types.Compress
@@ -1272,11 +1292,12 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
 		// If this fails while writing data, it will do pipeWriter.CloseWithError(); if it fails otherwise,
 		// e.g. because we have exited and due to pipeReader.Close() above further writing to the pipe has failed,
 		// we don’t care.
-		go c.compressGoroutine(pipeWriter, destStream, desiredCompressionFormat) // Closes pipeWriter
+		go c.compressGoroutine(pipeWriter, destStream, *uploadCompressionFormat) // Closes pipeWriter
 		destStream = pipeReader
 		inputInfo.Digest = ""
 		inputInfo.Size = -1
-	} else if canModifyBlob && c.dest.DesiredLayerCompression() == types.Compress && isCompressed && desiredCompressionFormat.Name() != compressionFormat.Name() {
+		uploadCompressorName = uploadCompressionFormat.Name()
+	} else if canModifyBlob && c.dest.DesiredLayerCompression() == types.Compress && isCompressed && uploadCompressionFormat.Name() != compressionFormat.Name() {
 		// When the blob is compressed, but the desired format is different, it first needs to be decompressed and finally
 		// re-compressed using the desired format.
 		logrus.Debugf("Blob will be converted")
@@ -1291,11 +1312,12 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
 		pipeReader, pipeWriter := io.Pipe()
 		defer pipeReader.Close()
 
-		go c.compressGoroutine(pipeWriter, s, desiredCompressionFormat) // Closes pipeWriter
+		go c.compressGoroutine(pipeWriter, s, *uploadCompressionFormat) // Closes pipeWriter
 
 		destStream = pipeReader
 		inputInfo.Digest = ""
 		inputInfo.Size = -1
+		uploadCompressorName = uploadCompressionFormat.Name()
 	} else if canModifyBlob && c.dest.DesiredLayerCompression() == types.Decompress && isCompressed {
 		logrus.Debugf("Blob will be decompressed")
 		compressionOperation = types.Decompress
@@ -1307,11 +1329,15 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
 		destStream = s
 		inputInfo.Digest = ""
 		inputInfo.Size = -1
+		uploadCompressorName = internalblobinfocache.Uncompressed
+		uploadCompressionFormat = nil
 	} else {
 		// PreserveOriginal might also need to recompress the original blob if the desired compression format is different.
 		logrus.Debugf("Using original blob without modification")
 		compressionOperation = types.PreserveOriginal
 		inputInfo = srcInfo
+		uploadCompressorName = srcCompressorName
+		uploadCompressionFormat = nil
 	}
 
 	// Perform image encryption for valid mediatypes if ociEncryptConfig provided
@@ -1371,9 +1397,7 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
 
 	uploadedInfo.CompressionOperation = compressionOperation
 	// If we can modify the layer's blob, set the desired algorithm for it to be set in the manifest.
-	if canModifyBlob && !isConfig {
-		uploadedInfo.CompressionAlgorithm = &desiredCompressionFormat
-	}
+	uploadedInfo.CompressionAlgorithm = uploadCompressionFormat
 	if decrypted {
 		uploadedInfo.CryptoOperation = types.Decrypt
 	} else if encrypted {
@@ -1390,7 +1414,7 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
 		}
 	}
 
-	// This is fairly horrible: the writer from getOriginalLayerCopyWriter wants to consumer
+	// This is fairly horrible: the writer from getOriginalLayerCopyWriter wants to consume
 	// all of the input (to compute DiffIDs), even if dest.PutBlob does not need it.
 	// So, read everything from originalLayerReader, which will cause the rest to be
 	// sent there if we are not already at EOF.
@@ -1422,6 +1446,12 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
 			c.blobInfoCache.RecordDigestUncompressedPair(srcInfo.Digest, uploadedInfo.Digest)
 		default:
 			return types.BlobInfo{}, errors.Errorf("Internal error: Unexpected compressionOperation value %#v", compressionOperation)
+		}
+		if uploadCompressorName != "" && uploadCompressorName != internalblobinfocache.UnknownCompression {
+			c.blobInfoCache.RecordDigestCompressorName(uploadedInfo.Digest, uploadCompressorName)
+		}
+		if srcInfo.Digest != "" && srcCompressorName != "" && srcCompressorName != internalblobinfocache.UnknownCompression {
+			c.blobInfoCache.RecordDigestCompressorName(srcInfo.Digest, srcCompressorName)
 		}
 	}
 	return uploadedInfo, nil
--- a/directory/directory_dest.go
+++ b/directory/directory_dest.go
@@ -194,7 +194,9 @@ func (d *dirImageDestination) PutBlob(ctx context.Context, stream io.Reader, inp
 // (e.g. if the blob is a filesystem layer, this signifies that the changes it describes need to be applied again when composing a filesystem tree).
 // info.Digest must not be empty.
 // If canSubstitute, TryReusingBlob can use an equivalent equivalent of the desired blob; in that case the returned info may not match the input.
-// If the blob has been successfully reused, returns (true, info, nil); info must contain at least a digest and size.
+// If the blob has been successfully reused, returns (true, info, nil); info must contain at least a digest and size, and may
+// include CompressionOperation and CompressionAlgorithm fields to indicate that a change to the compression type should be
+// reflected in the manifest that will be written.
 // If the transport can not reuse the requested blob, TryReusingBlob returns (false, {}, nil); it returns a non-nil error only on an unexpected failure.
 // May use and/or update cache.
 func (d *dirImageDestination) TryReusingBlob(ctx context.Context, info types.BlobInfo, cache types.BlobInfoCache, canSubstitute bool) (bool, types.BlobInfo, error) {
@@ -210,7 +212,6 @@ func (d *dirImageDestination) TryReusingBlob(ctx context.Context, info types.Blo
 		return false, types.BlobInfo{}, err
 	}
 	return true, types.BlobInfo{Digest: info.Digest, Size: finfo.Size()}, nil
-
 }
 
 // PutManifest writes manifest to the destination.
--- a/docker/docker_image_dest.go
+++ b/docker/docker_image_dest.go
@@ -15,6 +15,7 @@ import (
 	"strings"
 
 	"github.com/containers/image/v5/docker/reference"
+	"github.com/containers/image/v5/internal/blobinfocache"
 	"github.com/containers/image/v5/internal/iolimits"
 	"github.com/containers/image/v5/internal/uploadreader"
 	"github.com/containers/image/v5/manifest"
@@ -284,7 +285,9 @@ func (d *dockerImageDestination) mountBlob(ctx context.Context, srcRepo referenc
 // (e.g. if the blob is a filesystem layer, this signifies that the changes it describes need to be applied again when composing a filesystem tree).
 // info.Digest must not be empty.
 // If canSubstitute, TryReusingBlob can use an equivalent equivalent of the desired blob; in that case the returned info may not match the input.
-// If the blob has been successfully reused, returns (true, info, nil); info must contain at least a digest and size.
+// If the blob has been successfully reused, returns (true, info, nil); info must contain at least a digest and size, and may
+// include CompressionOperation and CompressionAlgorithm fields to indicate that a change to the compression type should be
+// reflected in the manifest that will be written.
 // If the transport can not reuse the requested blob, TryReusingBlob returns (false, {}, nil); it returns a non-nil error only on an unexpected failure.
 // May use and/or update cache.
 func (d *dockerImageDestination) TryReusingBlob(ctx context.Context, info types.BlobInfo, cache types.BlobInfoCache, canSubstitute bool) (bool, types.BlobInfo, error) {
@@ -299,17 +302,23 @@ func (d *dockerImageDestination) TryReusingBlob(ctx context.Context, info types.
 	}
 	if exists {
 		cache.RecordKnownLocation(d.ref.Transport(), bicTransportScope(d.ref), info.Digest, newBICLocationReference(d.ref))
-		return true, types.BlobInfo{Digest: info.Digest, Size: size}, nil
+		return true, types.BlobInfo{Digest: info.Digest, MediaType: info.MediaType, Size: size}, nil
 	}
 
 	// Then try reusing blobs from other locations.
-	for _, candidate := range cache.CandidateLocations(d.ref.Transport(), bicTransportScope(d.ref), info.Digest, canSubstitute) {
+	bic := blobinfocache.FromBlobInfoCache(cache)
+	candidates := bic.CandidateLocations2(d.ref.Transport(), bicTransportScope(d.ref), info.Digest, canSubstitute)
+	for _, candidate := range candidates {
 		candidateRepo, err := parseBICLocationReference(candidate.Location)
 		if err != nil {
 			logrus.Debugf("Error parsing BlobInfoCache location reference: %s", err)
 			continue
 		}
-		logrus.Debugf("Trying to reuse cached location %s in %s", candidate.Digest.String(), candidateRepo.Name())
+		if candidate.CompressorName != blobinfocache.Uncompressed {
+			logrus.Debugf("Trying to reuse cached location %s compressed with %s in %s", candidate.Digest.String(), candidate.CompressorName, candidateRepo.Name())
+		} else {
+			logrus.Debugf("Trying to reuse cached location %s with no compression in %s", candidate.Digest.String(), candidateRepo.Name())
+		}
 
 		// Sanity checks:
 		if reference.Domain(candidateRepo) != reference.Domain(d.ref.ref) {
@@ -351,8 +360,16 @@ func (d *dockerImageDestination) TryReusingBlob(ctx context.Context, info types.
 				continue
 			}
 		}
-		cache.RecordKnownLocation(d.ref.Transport(), bicTransportScope(d.ref), candidate.Digest, newBICLocationReference(d.ref))
-		return true, types.BlobInfo{Digest: candidate.Digest, Size: size}, nil
+
+		bic.RecordKnownLocation(d.ref.Transport(), bicTransportScope(d.ref), candidate.Digest, newBICLocationReference(d.ref))
+
+		compressionOperation, compressionAlgorithm, err := blobinfocache.OperationAndAlgorithmForCompressor(candidate.CompressorName)
+		if err != nil {
+			logrus.Debugf("... Failed: %v", err)
+			continue
+		}
+
+		return true, types.BlobInfo{Digest: candidate.Digest, MediaType: info.MediaType, Size: size, CompressionOperation: compressionOperation, CompressionAlgorithm: compressionAlgorithm}, nil
 	}
 
 	return false, types.BlobInfo{}, nil
--- a/internal/blobinfocache/blobinfocache.go
+++ b/internal/blobinfocache/blobinfocache.go
@@ -1,0 +1,63 @@
+package blobinfocache
+
+import (
+	"github.com/containers/image/v5/pkg/compression"
+	"github.com/containers/image/v5/types"
+	digest "github.com/opencontainers/go-digest"
+)
+
+// FromBlobInfoCache returns a BlobInfoCache2 based on a BlobInfoCache, returning the original
+// object if it implements BlobInfoCache2, or a wrapper which discards compression information
+// if it only implements BlobInfoCache.
+func FromBlobInfoCache(bic types.BlobInfoCache) BlobInfoCache2 {
+	if bic2, ok := bic.(BlobInfoCache2); ok {
+		return bic2
+	}
+	return &v1OnlyBlobInfoCache{
+		BlobInfoCache: bic,
+	}
+}
+
+type v1OnlyBlobInfoCache struct {
+	types.BlobInfoCache
+}
+
+func (bic *v1OnlyBlobInfoCache) RecordDigestCompressorName(anyDigest digest.Digest, compressorName string) {
+}
+
+func (bic *v1OnlyBlobInfoCache) CandidateLocations2(transport types.ImageTransport, scope types.BICTransportScope, digest digest.Digest, canSubstitute bool) []BICReplacementCandidate2 {
+	return nil
+}
+
+// CandidateLocationsFromV2 converts a slice of BICReplacementCandidate2 to a slice of
+// types.BICReplacementCandidate, dropping compression information.
+func CandidateLocationsFromV2(v2candidates []BICReplacementCandidate2) []types.BICReplacementCandidate {
+	candidates := make([]types.BICReplacementCandidate, 0, len(v2candidates))
+	for _, c := range v2candidates {
+		candidates = append(candidates, types.BICReplacementCandidate{
+			Digest:   c.Digest,
+			Location: c.Location,
+		})
+	}
+	return candidates
+}
+
+// OperationAndAlgorithmForCompressor returns CompressionOperation and CompressionAlgorithm
+// values suitable for inclusion in a types.BlobInfo structure, based on the name of the
+// compression algorithm, or Uncompressed, or UnknownCompression.  This is typically used by
+// TryReusingBlob() implementations to set values in the BlobInfo structure that they return
+// upon success.
+func OperationAndAlgorithmForCompressor(compressorName string) (types.LayerCompression, *compression.Algorithm, error) {
+	switch compressorName {
+	case Uncompressed:
+		return types.Decompress, nil, nil
+	case UnknownCompression:
+		return types.PreserveOriginal, nil, nil
+	default:
+		algo, err := compression.AlgorithmByName(compressorName)
+		if err == nil {
+			return types.Compress, &algo, nil
+		}
+		return types.PreserveOriginal, nil, err
+	}
+}
--- a/internal/blobinfocache/types.go
+++ b/internal/blobinfocache/types.go
@@ -1,0 +1,45 @@
+package blobinfocache
+
+import (
+	"github.com/containers/image/v5/types"
+	digest "github.com/opencontainers/go-digest"
+)
+
+const (
+	// Uncompressed is the value we store in a blob info cache to indicate that we know that
+	// the blob in the corresponding location is not compressed.
+	Uncompressed = "uncompressed"
+	// UnknownCompression is the value we store in a blob info cache to indicate that we don't
+	// know if the blob in the corresponding location is compressed (and if so, how) or not.
+	UnknownCompression = "unknown"
+)
+
+// BlobInfoCache2 extends BlobInfoCache by adding the ability to track information about what kind
+// of compression was applied to the blobs it keeps information about.
+type BlobInfoCache2 interface {
+	types.BlobInfoCache
+	// RecordDigestCompressorName records a compressor for the blob with the specified digest,
+	// or Uncompressed or UnknownCompression.
+	// WARNING: Only call this with LOCALLY VERIFIED data; don’t record a compressor for a
+	// digest just because some remote author claims so (e.g. because a manifest says so);
+	// otherwise the cache could be poisoned and cause us to make incorrect edits to type
+	// information in a manifest.
+	RecordDigestCompressorName(anyDigest digest.Digest, compressorName string)
+	// CandidateLocations2 returns a prioritized, limited, number of blobs and their locations
+	// that could possibly be reused within the specified (transport scope) (if they still
+	// exist, which is not guaranteed).
+	//
+	// If !canSubstitute, the returned cadidates will match the submitted digest exactly; if
+	// canSubstitute, data from previous RecordDigestUncompressedPair calls is used to also look
+	// up variants of the blob which have the same uncompressed digest.
+	//
+	// The CompressorName fields in returned data must never be UnknownCompression.
+	CandidateLocations2(transport types.ImageTransport, scope types.BICTransportScope, digest digest.Digest, canSubstitute bool) []BICReplacementCandidate2
+}
+
+// BICReplacementCandidate2 is an item returned by BlobInfoCache2.CandidateLocations2.
+type BICReplacementCandidate2 struct {
+	Digest         digest.Digest
+	CompressorName string // either the Name() of a known pkg/compression.Algorithm, or Uncompressed or UnknownCompression
+	Location       types.BICLocationReference
+}
--- a/manifest/common.go
+++ b/manifest/common.go
@@ -5,7 +5,6 @@ import (
 
 	"github.com/containers/image/v5/pkg/compression"
 	"github.com/containers/image/v5/types"
-	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
 
@@ -54,6 +53,12 @@ const mtsUnsupportedMIMEType = "" // A value in compressionMIMETypeSet that mean
 
 // compressionVariantMIMEType returns a variant of mimeType for the specified algorithm (which may be nil
 // to mean "no compression"), based on variantTable.
+// The returned error will be a ManifestLayerCompressionIncompatibilityError if mimeType has variants
+// that differ only in what type of compression is applied, but it can't be combined with this
+// algorithm to produce an updated MIME type that complies with the standard that defines mimeType.
+// If the compression algorithm is unrecognized, or mimeType is not known to have variants that
+// differ from it only in what type of compression has been applied, the returned error will not be
+// a ManifestLayerCompressionIncompatibilityError.
 func compressionVariantMIMEType(variantTable []compressionMIMETypeSet, mimeType string, algorithm *compression.Algorithm) (string, error) {
 	if mimeType == mtsUnsupportedMIMEType { // Prevent matching against the {algo:mtsUnsupportedMIMEType} entries
 		return "", fmt.Errorf("cannot update unknown MIME type")
@@ -70,15 +75,15 @@ func compressionVariantMIMEType(variantTable []compressionMIMETypeSet, mimeType 
 						return res, nil
 					}
 					if name != mtsUncompressed {
-						return "", fmt.Errorf("%s compression is not supported", name)
+						return "", ManifestLayerCompressionIncompatibilityError{fmt.Sprintf("%s compression is not supported for type %q", name, mt)}
 					}
-					return "", errors.New("uncompressed variant is not supported")
+					return "", ManifestLayerCompressionIncompatibilityError{fmt.Sprintf("uncompressed variant is not supported for type %q", mt)}
 				}
 				if name != mtsUncompressed {
-					return "", fmt.Errorf("unknown compression algorithm %s", name)
+					return "", ManifestLayerCompressionIncompatibilityError{fmt.Sprintf("unknown compressed with algorithm %s variant for type %s", name, mt)}
 				}
 				// We can't very well say “the idea of no compression is unknown”
-				return "", errors.New("uncompressed variant is not supported")
+				return "", ManifestLayerCompressionIncompatibilityError{fmt.Sprintf("uncompressed variant is not supported for type %q", mt)}
 			}
 		}
 	}
@@ -89,7 +94,11 @@ func compressionVariantMIMEType(variantTable []compressionMIMETypeSet, mimeType 
 }
 
 // updatedMIMEType returns the result of applying edits in updated (MediaType, CompressionOperation) to
-// mimeType, based on variantTable. It may use updated.Digest for error messages.
+// mimeType, based on variantTable.  It may use updated.Digest for error messages.
+// The returned error will be a ManifestLayerCompressionIncompatibilityError if mimeType has variants
+// that differ only in what type of compression is applied, but applying updated.CompressionOperation
+// and updated.CompressionAlgorithm to it won't produce an updated MIME type that complies with the
+// standard that defines mimeType.
 func updatedMIMEType(variantTable []compressionMIMETypeSet, mimeType string, updated types.BlobInfo) (string, error) {
 	// Note that manifests in containers-storage might be reporting the
 	// wrong media type since the original manifests are stored while layers
@@ -99,6 +108,12 @@ func updatedMIMEType(variantTable []compressionMIMETypeSet, mimeType string, upd
 	// {de}compressed.
 	switch updated.CompressionOperation {
 	case types.PreserveOriginal:
+		// Force a change to the media type if we're being told to use a particular compressor,
+		// since it might be different from the one associated with the media type.  Otherwise,
+		// try to keep the original media type.
+		if updated.CompressionAlgorithm != nil {
+			return compressionVariantMIMEType(variantTable, mimeType, updated.CompressionAlgorithm)
+		}
 		// Keep the original media type.
 		return mimeType, nil
 
@@ -115,4 +130,15 @@ func updatedMIMEType(variantTable []compressionMIMETypeSet, mimeType string, upd
 	default:
 		return "", fmt.Errorf("unknown compression operation (%d)", updated.CompressionOperation)
 	}
+}
+
+// ManifestLayerCompressionIncompatibilityError indicates that a specified compression algorithm
+// could not be applied to a layer MIME type.  A caller that receives this should either retry
+// the call with a different compression algorithm, or attempt to use a different manifest type.
+type ManifestLayerCompressionIncompatibilityError struct {
+	text string
+}
+
+func (m ManifestLayerCompressionIncompatibilityError) Error() string {
+	return m.text
 }
--- a/manifest/common_test.go
+++ b/manifest/common_test.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/containers/image/v5/pkg/compression"
 	"github.com/containers/image/v5/types"
+	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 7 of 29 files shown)


docker/docker_image_dest.go
```diff
@@ -351,8 +358,24 @@ func (d *dockerImageDestination) TryReusingBlob(ctx context.Context, info types.
...
+		if candidate.CompressorName != "" {
+			compressionOperation = types.Compress
+			algorithm, err := compression.AlgorithmByName(candidate.CompressorName)
+			if err != nil {
+				logrus.Debugf("... unknown compressor %q", candidate.CompressorName)
+				return false, types.BlobInfo{}, nil
+			}
+			compressionAlgorithm = &algorithm
+		}
+		mediaType, err := mediatypes.AnyUpdatedMIMEType(info.MediaType, types.BlobInfo{CompressionOperation: compressionOperation, CompressionAlgorithm: compressionAlgorithm})
```
mtrmac, line 374: I haven’t read the code in full at all, (and I apologize if I’m missing the whole point), but this feels misplaced at a first glance.

Originally I was expecting the cache to record the raw MIME type value, `TryReusingBlob` to report the raw MIME type value, and the _generic code_ deal with the concept of manifest-format-specific MIME types. It’s a very good insight that “MIME type of a blob” is not really a real concept with the v2s2/OCI alternatives, still, the transports should to an extent not need to worry about that.

Especially inventing a ~random MIME type based on the input (not output) format, and the compression mechanism seems superfluous — if the MIME type is really that bogus, maybe we should be deprecating/removing it completely? OTOH that seems way too much.

(I have no idea what the right answer is — at this point I’m just thinking that the difficulty of changing APIs makes it important to get this right if possible.)

---

This doesn’t handle uncompressed formats AFAICS.

mtrmac (commented):
Thanks for working on this!

(Not at all a review, just a minimal first skim. Making most of this ~unavoidably visible through the public API is the main worry.)

types/types.go
```diff
@@ -208,7 +211,7 @@ type BlobInfoCache interface {
 
 	// RecordKnownLocation records that a blob with the specified digest exists within the specified (transport, scope) scope,
 	// and can be reused given the opaque location data.
-	RecordKnownLocation(transport ImageTransport, scope BICTransportScope, digest digest.Digest, location BICLocationReference)
+	RecordKnownLocation(transport ImageTransport, scope BICTransportScope, digest digest.Digest, compressorName string, location BICLocationReference)
```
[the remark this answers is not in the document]
nalind: I didn't have a way to record the additional info that didn't end up requiring adding either a new parameter here, or a new method (which would be another write to the database) to the interface.  If you know of a way to do that, I'd be thrilled to restructure this.

pkg/blobinfocache/boltdb/boltdb.go
```diff
@@ -244,7 +245,7 @@ func (bdc *cache) RecordKnownLocation(transport types.ImageTransport, scope type
 		if err != nil {
 			return err
 		}
-		if err := b.Put([]byte(location.Opaque), value); err != nil { // Possibly overwriting an older entry.
+		if err := b.Put([]byte(location.Opaque), append(append([]byte(compressorName), 0), value...)); err != nil { // Possibly overwriting an older entry.
```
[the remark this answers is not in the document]
nalind: Oops, yeah, will fix that.

docker/docker_image_dest.go
nalind: An earlier version of the patch recorded the MIME types.  For cases where we ended up doing format conversions, that required adding a way to map from "schema 2's name for a gzip-compressed layer" in a BICReplacementCandidate to "OCI's name for a gzip-compressed layer" in copy, or in TryReusingBlob(), where the manifest's type isn't currently exposed.  I found that recording just the compressor names in the cache and passing MediaType values across the existing interfaces let me avoid adding that.

For comparison, the WIP before that change is https://github.com/nalind/image/tree/blob-info-cache-media-types-1.  It didn't move mappings from the `manifest` package to `internal/mediatypes`, so the diff stats aren't going to be very useful.  I wasn't happy with having so many ad-hoc lookup tables scattered across the code base.

The intent is to use `MTSUncompressed` as the "compressor" for uncompressed formats.  It's possible that part of that is missing and causing the test failures, but I'm thinking that if the blob cache cares more about the MediaType field than it used to, that some of those are due to the schema 1 manifests not supplying values there.

types/types.go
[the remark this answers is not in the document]
mtrmac: No _good_ way I’m afraid. `type BlobInfoCacheV2` + a type check to call V1 or V2 + probably an internal helper to simplify callers , _and_ all of the code supporting both at the same time.

pkg/blobinfocache/boltdb/boltdb.go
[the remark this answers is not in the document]
nalind: Updated to `knownLocationsBucket stores a nested structure of buckets, keyed by (transport name, scope string, blob digest), ultimately containing a bucket of (opaque location reference, compressorName+\0+BinaryMarshaller-encoded time.Time value). v2 of this format added the compressorName and the \0 separator; v1 did not include them`, let me know if you want something different.

docker/docker_image_dest.go
mtrmac: > For comparison, the WIP before that change is https://github.com/nalind/image/tree/blob-info-cache-media-types-1. 

Thanks, I’ll take a look.

>  I wasn't happy with having so many ad-hoc lookup tables scattered across the code base.

Very vaguely I guess that it should be possible to build a single master table of (manifest format, blob semantics, compression format) → MIME type in a way that allows the relevant lookups (“what is this type with compression changed? what is the type for a different manifest?” etc.) Building that for O(1) lookups and no repetition would be some work, but the tables are currently very small — linear lookups would might well be perfectly OK.

(The existence of `CompressionOperation` inside `BlobInfo` is conceptually wrong — it is a “make this edit” temporary state of the copy process just for `ManifestUpdateOptions`, just something we didn’t catch in time. Making this a field that is fair game to be manipulated by the transports complicates the analysis quite a bit, or requires setting and documenting detailed restrictive expectations.)

---

> The intent is to use `MTSUncompressed` as the "compressor" for uncompressed formats. It's possible that part of that is missing and causing the test failures, but I'm thinking that if the blob cache cares more about the MediaType field than it used to, that some of those are due to the schema 1 manifests not supplying values there.

That could work, but the way I’m reading the code (untested), `mediatypes.UpdatedMIMEType` only does a table lookup if the `CompressionOperation` is not `PreserveOriginal`, and this code only sets `CompressionOperation` to `Compress` if the compressor is non-empty. A quick `if candidate.CompressorName != "" { … } else { compressionOperation = types.Decompress }` might be all it takes.

nalind: I think you're right about the cases where we want to change the type of compression being used for a layer.

nalind: Modified `UpdatedMIMEType()` to treat `PreserveOriginal` with a non-empty compressor name differently, to update the media type.  Setting the operation to `Decompress` when there's no compressor set could have us decompress something that's already compressed when we don't intend that, and  I think `PreserveOriginal` with no compressor specified happens a lot, so holding off on that, at least for now.

types/types.go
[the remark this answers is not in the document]
nalind: > This is an API break, sadly third-party implementations of `BlobInfoCache` are possible. (The API impact is the awkward part of the problem.)

I really wish you were wrong about that, but https://github.com/containers/buildah/pull/2828 exists.

> No _good_ way I’m afraid. `type BlobInfoCacheV2` + a type check to call V1 or V2 + probably an internal helper to simplify callers , _and_ all of the code supporting both at the same time.

Is this something you'd like me to pursue here?

nalind:
Dropped the "MediaType is always correct" goal because handling switching to zstd compression while updating a schema2 manifest wasn't looking doable without restructuring how the `copy` package was handling manifests.
The updated version makes `TryReusingBlob()` do most of the work, in that if it returns a non-nil `CompressionAlgorithm` in the returned `BlobInfo`, the existing manifest `UpdateLayerInfos` logic takes care of most of the rest.  The version in the "docker" transport uses the blob info cache, while the "directory" and "oci" versions look at the blobs that they find.
The "docker" transport's `GetBlob` trusts the `MediaType` in the source `BlobInfo` that it's given.  Pulling schema1 images could be a problem here -- the bits that had it supply a fake MediaType in its LayerInfos didn't look like they were needed for passing tests, so I dropped them out.  They can come back if need be.
The "docker" transport's `PutBlob` just looks at the blob to make sure it's recording the correct information about how a blob is compressed.

types/types.go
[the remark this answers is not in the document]
nalind: Okay, I'm going to start roughing that out: a v2 interface that, if the cache we're given implements it, we'll use to record both v1 and v2 information using different files for each.  For reads, it will fall back to checking for v1 data if the v2 interface doesn't return any data.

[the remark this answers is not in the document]
nalind: Okay, I think that's going to work without being _too_ awful to look at.  This current revision reworks how we store the compression information - it's now in a bucket in the same file that mirrors the structure of the one we use for storing timestamp information.  Logic for updating two different files, and keeping track of their names, and initializing a boltdb cache using two files, got unwieldy pretty fast.  There's still some cleanup work to do to fix and flesh out the unit tests, but the functional parts are pretty much where I want them to be.

nalind:
The "docker" transport's `GetBlob()` implementation now detects compression on blobs that it retrieves for recording compression information instead of guessing based on the passed-in MIME type, so it no longer needs visibility into conversion information that the `manifest` package has.  
The current revision drops the hunk that migrates compression type information into an `internal/mediatypes` package, because it's no longer necessary.

nalind:
Rebased.

docker/docker_image_dest.go
```diff
@@ -298,18 +309,43 @@ func (d *dockerImageDestination) TryReusingBlob(ctx context.Context, info types.
 		return false, types.BlobInfo{}, err
 	}
 	if exists {
-		cache.RecordKnownLocation(d.ref.Transport(), bicTransportScope(d.ref), info.Digest, newBICLocationReference(d.ref))
-		return true, types.BlobInfo{Digest: info.Digest, Size: size}, nil
+		var compressorName string
+		if info.CompressionAlgorithm != nil {
+			compressorName = info.CompressionAlgorithm.Name()
+		}
```
nalind, line 315: I'm not completely sure we end up recording the right value here.

types/types.go
```diff
@@ -218,6 +225,23 @@ type BlobInfoCache interface {
 	CandidateLocations(transport ImageTransport, scope BICTransportScope, digest digest.Digest, canSubstitute bool) []BICReplacementCandidate
 }
 
+// BlobInfoCache2 extends BlobInfoCache by adding the ability to track information about what kind of compression was applied to
+// the blobs it keeps information about.  If a BlobInfoCache also implements BlobInfoCache2, users of the cache are encouraged to
+// use RecordKnownLocation2() instead of RecordKnownLocation(), and to use CandidateLocations2() instead of CandidateLocations().
+type BlobInfoCache2 interface {
```
mtrmac, line 231: My initial impression is that this shouldn’t be public; we might be forced to open it up again, but let’s at least try not to.

- Move both `BlobInfoCache2` and `BICReplacementCandidate2` into `c/image/internal/blobinfocache`
  - (Be extra paranoid and add a non-public method that is impossible to implement outside of c/image/internal ? That might be overdoing it to the point of being actively unhelpful.)
- Document `BlobInfoCache` as (supported but?) deprecated in the sense that library callers should use one of the cache implementations in `c/image/pkg/blobinfocache` only, if at all possible.
- In `c/image/internal/blobinfocache`, provide helper functions that abstract the difference for callers, so that every transport doesn’t have to open-code that.
  - Non-blocking: It could also be possible to abstract the difference for the implementations (i.e. provide a “forwarding” implementation of `RecordKnownLocation` that can be embedded in a cache implementation).
  - (That _might_ be also a place to implement more of the common logic, perhaps even “update `BlobInfo` from a stream” and the like. I’m not at all sure yet, naively without taking a deeper look I’d prefer most of the intelligence to be not the responsibility of the transports but of `c/image/copy`.)

types/types.go
```diff
@@ -218,6 +225,23 @@ type BlobInfoCache interface {
...
+	BlobInfoCache
+	// RecordKnownLocation2 records that a blob with the specified digest exists within the specified (transport, scope) scope,
+	// compressed with the specified compression algorithm, and can be reused given the opaque location data.
```
mtrmac, line 234: If this can represent “uncompressed” and “unknown”, the two should be documented and distinct.

(If we end up with an internal definition of the interface, perhaps we can avoid the “unknown” case.)

types/types.go
```diff
@@ -170,6 +170,13 @@ type BICReplacementCandidate struct {
 	Location BICLocationReference
 }
 
+// BICReplacementCandidate2 is an item returned by BlobInfoCache2.CandidateLocations2.
+type BICReplacementCandidate2 struct {
+	Digest         digest.Digest
+	CompressorName string
```
mtrmac, line 176: Can this be “uncompressed” or “unknown”? If so, that should be documented and distinct.

And do we just soldier on when getting “unknown”, or prefer correctness and don’t reuse? Just by asking that question I’m somewhat suggesting to prefer correctness, OTOH that would make third-party cache implementations half-ignored. Something for me to think about while I continue reading tomorrow.

pkg/blobinfocache/memory/memory_test.go
```diff
@@ -7,6 +7,9 @@ import (
 	"github.com/containers/image/v5/types"
 )
 
+var _ types.BlobInfoCache = &cache{}
+var _ types.BlobInfoCache2 = &cache{}
```
mtrmac, line 11: This would be useful to have for `NoCache` as well.

pkg/blobinfocache/memory/memory.go
```diff
@@ -18,13 +18,18 @@ type locationKey struct {
...
+}
+
 // cache implements an in-memory-only BlobInfoCache
 type cache struct {
 	mutex sync.Mutex
 	// The following fields can only be accessed with mutex held.
 	uncompressedDigests   map[digest.Digest]digest.Digest
-	digestsByUncompressed map[digest.Digest]map[digest.Digest]struct{}             // stores a set of digests for each uncompressed digest
-	knownLocations        map[locationKey]map[types.BICLocationReference]time.Time // stores last known existence time for each location reference
+	digestsByUncompressed map[digest.Digest]map[digest.Digest]struct{}                     // stores a set of digests and media types for each uncompressed digest
```
mtrmac, line 31: (The comment update is incorrect.)

pkg/blobinfocache/internal/test/test.go
```diff
@@ -103,6 +116,14 @@ func assertCandidatesMatch(t *testing.T, scopeName string, expected []candidate,
 	assert.Equal(t, e, actual)
 }
 
+func assertCandidatesMatch2(t *testing.T, scopeName string, expected []candidate, actual []types.BICReplacementCandidate2) {
```
mtrmac, line 119: If the set of new tests is the same as the old tests (is it?), maybe instead move the `cache.CandidateLocations{,2}` calls inside the `assert…` helper, have a single helper call test both methods, and we can avoid the duplication.

(This can clearly wait until after the interface is settled.)

pkg/blobinfocache/internal/test/test.go
```diff
@@ -16,6 +16,10 @@ const (
 	digestCompressedA         = digest.Digest("sha256:3333333333333333333333333333333333333333333333333333333333333333")
 	digestCompressedB         = digest.Digest("sha256:4444444444444444444444444444444444444444444444444444444444444444")
 	digestCompressedUnrelated = digest.Digest("sha256:5555555555555555555555555555555555555555555555555555555555555555")
+	compressorNameU           = "compressorName/U"
+	compressorNameA           = "compressorName/A"
+	compressorNameB           = "compressorName/B"
+	compressorNameCU          = "compressorName/CU"
 )
 
 // GenericCache runs an implementation-independent set of tests, given a
```
mtrmac, line 25: Absolutely non-blocking, it might be interesting to test _both_ the V2 implementation and the V1 implementation"
- call the V2 versions if available, which is “always” for the two users, or even just assert that V2 is available
- wrap the V2 implementation in a V1-only object to test that the V1 implementation, which is externally-observable, continues to work as expected for V1.

OTOH… “externally-observable” is, I fervently hope, such a theoretical concept, so who cares…

pkg/blobinfocache/internal/prioritize/prioritize_test.go
```diff
@@ -21,20 +21,20 @@ var (
 	// cssLiteral contains a non-trivial candidateSortState shared among several tests below.
 	cssLiteral = candidateSortState{
 		cs: []CandidateWithTime{
```
mtrmac, line 23: Add some compression values?

(_After_ the interface settles.)

mtrmac (commented):
So far basically just looking at the API, I’ll continue reading the actual implementation tomorrow.

types/types.go
nalind: Sounds good to me.  Will make those changes.

pkg/blobinfocache/memory/memory_test.go
nalind: I keep forgetting about `NoCache`.  Will add it.

pkg/blobinfocache/memory/memory.go
nalind: Whoops, leftover from a previous iteration.  Reverting this bit.

types/types.go
nalind: Yes, I think it makes sense to mark cases where we don't know because we retrieved the data from a cache that original methods.  Added constants for uncompressed and unknown to avoid ambiguity.

types/types.go
nalind: Okay, removed a lot of the "if implements v2, use it, else call the v1 version" logic in favor of wrapping cache implementations that only implement one or the other pair of methods in a cache object that provides both sets of methods, and using that everywhere.

nalind:
Rebased, and fixed cases where we incorrectly used the `io.Reader` returned by `pkg/compression.DetectCompressionFormat()`.

docker/docker_image_dest.go
nalind: Changed the `TryReusingBlob()` case where the registry indicates that it already has the blob so that we record the compression type as UnknownCompression.  The MIME type we're passed from `copy` comes from the source image's manifest, so it should already be correct for the blob with the same digest without needing any fixup from us.

types/types.go
nalind: Distinguishing cases where the compression type isn't known from uncompressed cases would keep us from actively breaking things, so I'm leaning that way as well.  Updated `dockerImageDestination.TryReusingBlob()` to skip candidate locations where the compression type isn't known.

directory/directory_dest.go
```diff
@@ -194,23 +195,38 @@ func (d *dirImageDestination) PutBlob(ctx context.Context, stream io.Reader, inp
...
 	}
-	return true, types.BlobInfo{Digest: info.Digest, Size: finfo.Size()}, nil
+	defer f.Close()
+
+	finfo, err := f.Stat()
+	if err != nil {
+		return false, types.BlobInfo{}, err
+	}
+
+	compressionOperation, compressionAlgorithm, _, _, err := blobinfocache.DetectCompressionFormat(f)
```
mtrmac, line 223: Both in this and the OCI/layout case, the code only looks for exactly the specified digest and no substitution happens. So this should never cause a change in the MIME type, and the two calls to `DetectCompressionFormat` seem unnecessary.

---

OTOH what about the `canSubstitute` case in `storageImageDestination.TryReusingBlob`? That one might need a MIME type update.

docker/docker_image_dest.go
```diff
@@ -291,25 +301,35 @@ func (d *dockerImageDestination) TryReusingBlob(ctx context.Context, info types.
...
 
 	// First, check whether the blob happens to already exist at the destination.
 	exists, size, err := d.blobExists(ctx, d.ref.ref, info.Digest, nil)
 	if err != nil {
 		return false, types.BlobInfo{}, err
 	}
 	if exists {
-		cache.RecordKnownLocation(d.ref.Transport(), bicTransportScope(d.ref), info.Digest, newBICLocationReference(d.ref))
-		return true, types.BlobInfo{Digest: info.Digest, Size: size}, nil
+		bic.RecordKnownLocation2(d.ref.Transport(), bicTransportScope(d.ref), info.Digest, blobinfocache.UnknownCompression, newBICLocationReference(d.ref))
```
mtrmac, line 312: Most of the time (but not for schema1), `info.MediaType` should be set and valid here. Recording `Unknown` and preventing the loop below from benefiting is not ideal.

So it would be useful to implement the MIME type → compression, or MIME type → [compression] → MIME type, lookup — and at that would in turn allow moving this out of the transports into the `TryReusingBlob` caller.

… except that we might not trust the contents of a manifest. Ugh.  But, see elsewhere, if we split recording the location from recording the compression, generic code can reliably detect compression and record it.

docker/docker_image_dest.go
```diff
@@ -351,8 +371,24 @@ func (d *dockerImageDestination) TryReusingBlob(ctx context.Context, info types.
...
+			if err != nil {
+				logrus.Debugf("... unknown compressor %q", candidate.CompressorName)
+				return false, types.BlobInfo{}, nil
+			}
+			compressionOperation = types.Compress
+			compressionAlgorithm = &algorithm
+		} else {
+			compressionOperation = types.Decompress
+			compressionAlgorithm = &compression.Algorithm{}
+		}
```
mtrmac, line 390: Nothing about this is transport-specific; it feels like something that should exist in the generic code.

internal/blobinfocache/types.go
```diff
@@ -0,0 +1,63 @@
...
+
+// blobInfoCache2Additions contains extended versions of the BlobInfoCache methods which add the
+// ability to track information about what kind of compression was applied to the blobs it keeps
+// information about.
+type blobInfoCache2Additions interface {
+	// RecordKnownLocation2 records that a blob with the specified digest exists within the
+	// specified (transport, scope) scope, compressed with the specified compression algorithm
+	// (one recognized by pkg/compression.AlgorithmByName()), or Uncompressed or with
+	// UnknownCompression, and can be reused given the opaque location data.
+	RecordKnownLocation2(transport types.ImageTransport, scope types.BICTransportScope, digest digest.Digest, compressorName string, location types.BICLocationReference)
```
mtrmac, line 54: (Looking at the docker loop) the compressor name is really a property of the digest, not of any specific location; so this could probably make sense as a separate “record compression for digest” API.

Or maybe the API should be a single call to have only a single BoltDB lock+use, but internally be recorded in two different ways? (Similarly, `CandidateLocations2` might still return both together.)

Recording the compression per-digest instead of per-location would also directly lead to `copyBlobFromStream`, which is _already_ detecting compression, being in charge of recording that — _both_ for the source and the destination, instead of `dockerImage{Source.GetBlob,Destination,PutBlob}`.

mtrmac (commented):
Just some half-baked thoughts, hopefully useful a bit.

directory/directory_dest.go
nalind: `storage` is a bit of a weird beast here.  It uses `LayersInfosForCopy()`, so it's never on the hook for being able to reproduce the blobs that it's given.  It's also in the business of trying to avoid causing changes to the manifest even when they're allowed, so it wants to only ever behave as though `canSubstitute` is false.

mtrmac: Yes, OTOH we already need to provide an uncompressed size so that the generic copy code is consistent, IIRC, so it makes sense for that rare code path to be fully consistent instead of partially.

Let’s see how the other paths pan out, at this point it’s probably just something to eventually verify.

internal/blobinfocache/types.go
nalind: I'm starting to rough out a `RecordDigestCompressorName()` method, which for boltdb stores compressor names (or "uncompressed" or "unknown") in a top-level "digestCompressor" bucket keyed by blob digest, to see how well that goes.

types/types.go
nalind: Added distinct values for both Uncompressed and Unknown; for Unknown we just drop the entry from the blobinfocache.

types/types.go
nalind: Okay, I _think_ we don't need to skip these any more.

docker/docker_image_dest.go
nalind: Moved to a helper function in `internal/blobinfocache` for use by `TryReusingBlob()` implementations.

nalind:
Updated to store compression information about a known blob digest separately from the blob's location.  Compression information is now stored by `copyBlobFromStream()`, as you suggested, and retrieved along with other information about candidates by the `docker` transport's `TryReusingBlob()`.

jomeier:
Docker pull rate limit ...

/retest

nalind:
/retest
Hoping https://github.com/containers/skopeo/pull/1146 made a difference.

nalind:
The bot's not here, so I've restarted the test at Travis.

nalind:
Rebased.

mtrmac (commented):
Partial notes for quicker feedback.

(Note to self, to do:
- `isCompressionIncompatible`:
- `manifest/`
- `pkg/BIC/internal/test`
- Eventually, documentation strings

internal/blobinfocache/types.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/containers/image/v5/types"
+	digest "github.com/opencontainers/go-digest"
+)
+
+const (
+	// Uncompressed is the value we store in a blob info cache to indicate that we know that the blob in
+	// the corresponding location is not compressed.
+	Uncompressed = "uncompressed"
+	// UnknownCompression s the value we store in a blob info cache to indicate that we don't know if the
+	// blob in the corresponding location is compressed or not.
```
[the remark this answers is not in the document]
nalind: Looks right, updating.

internal/blobinfocache/types.go
```diff
@@ -0,0 +1,46 @@
...
+type BICReplacementCandidate2 struct {
+	Digest         digest.Digest
+	CompressorName string // either the Name() of a known pkg/compression.Algorithm, or Uncompressed or UnknownCompression
+	Location       types.BICLocationReference
+}
+
+// BlobInfoCache2WithoutV1 contains extended versions of the BlobInfoCache methods which add the
+// ability to track information about what kind of compression was applied to the blobs it keeps
+// information about.
+type blobInfoCache2WithoutV1 interface {
```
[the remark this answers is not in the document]
nalind: Not really, now that I look at it.  Having it distinct is a vestige of an earlier revision.  Merging its methods into the only interface that was incorporating it.

pkg/blobinfocache/boltdb/boltdb.go
```diff
@@ -220,6 +226,30 @@ func (bdc *cache) RecordDigestUncompressedPair(anyDigest digest.Digest, uncompre
 	}) // FIXME? Log error (but throttle the log volume on repeated accesses)?
 }
 
+// RecordDigestCompressorName records that the blob with digest anyDigest was compressed with the specified
+// compressor, or is blobinfocache.Uncompressed.
+// WARNING: Only call this for LOCALLY VERIFIED data; don’t record a digest pair just because some remote author claims so (e.g.
+// because a manifest/config pair exists); otherwise the cache could be poisoned and allow substituting unexpected blobs.
```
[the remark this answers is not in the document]
nalind: Right, adding this:
```
WARNING: Only call this with LOCALLY VERIFIED data; don’t record a compressor for a
digest just because some remote author claims so (e.g. because a manifest says so);
otherwise the cache could be poisoned and cause us to make incorrect edits to type
information in a manifest.
```

internal/blobinfocache/blobinfocache.go
```diff
@@ -0,0 +1,72 @@
...
+func FromBlobInfoCache(bic types.BlobInfoCache) BlobInfoCache2 {
+	if bic2, ok := bic.(BlobInfoCache2); ok {
+		return bic2
+	}
+	return &wrappedBlobInfoCache{
+		BlobInfoCache: bic,
+	}
+}
+
+type wrappedBlobInfoCache struct {
```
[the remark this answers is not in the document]
nalind: Sure, renaming it.

internal/blobinfocache/blobinfocache.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func (bic *wrappedBlobInfoCache) CandidateLocations2(transport types.ImageTransport, scope types.BICTransportScope, digest digest.Digest, canSubstitute bool) []BICReplacementCandidate2 {
+	oldCandidates := bic.CandidateLocations(transport, scope, digest, canSubstitute)
+	results := make([]BICReplacementCandidate2, 0, len(oldCandidates))
+	for _, c := range oldCandidates {
+		results = append(results, BICReplacementCandidate2{
+			Digest:         c.Digest,
+			Location:       c.Location,
+			CompressorName: UnknownCompression,
```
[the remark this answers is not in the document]
nalind: Easy enough while we're in there.

pkg/blobinfocache/memory/memory.go
```diff
@@ -87,8 +94,9 @@ func (mem *cache) RecordDigestUncompressedPair(anyDigest digest.Digest, uncompre
 	anyDigestSet[anyDigest] = struct{}{} // Possibly writing the same struct{}{} presence marker again.
 }
 
-// RecordKnownLocation records that a blob with the specified digest exists within the specified (transport, scope) scope,
-// and can be reused given the opaque location data.
+// RecordKnownLocation records that a blob with the specified digest, whether compressed with the specified algorithm or
+// uncompressed if compressorName is "", exists within the specified (transport, scope) scope, and can be reused given
+// the opaque location data.
```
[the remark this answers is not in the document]
nalind: Dropping.

pkg/blobinfocache/boltdb/boltdb.go
```diff
@@ -252,20 +282,28 @@ func (bdc *cache) RecordKnownLocation(transport types.ImageTransport, scope type
 }
 
 // appendReplacementCandiates creates prioritize.CandidateWithTime values for digest in scopeBucket, and returns the result of appending them to candidates.
```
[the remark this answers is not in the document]
nalind: Adding a bit about `compressionBucket` (`appendReplacementCandiates creates prioritize.CandidateWithTime values for digest in scopeBucket with corresponding compression info from compressionBucket (if compressionBucket is not nil), and returns the result of appending them to candidates.`) to the godoc here, and reiterating why we check for `nil` when we check it (`the bucket won't exist if the cache was created by a v1 implementation and hasn't yet been updated by a v2 implementation`).

pkg/blobinfocache/boltdb/boltdb.go
```diff
@@ -220,6 +226,30 @@ func (bdc *cache) RecordDigestUncompressedPair(anyDigest digest.Digest, uncompre
...
+		key := []byte(anyDigest.String())
+		if previousBytes := b.Get(key); previousBytes != nil {
+			if string(previousBytes) != compressorName {
+				logrus.Warnf("Compressor for blob with digest %s previously recorded as %s, now %s", anyDigest, string(previousBytes), compressorName)
+			}
+		}
+		if compressorName == blobinfocache.UnknownCompression {
+			return b.Delete([]byte(anyDigest.String()))
+		}
+		return b.Put([]byte(anyDigest.String()), []byte(compressorName))
```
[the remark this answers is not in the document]
nalind: Ah, missed that.  Fixing.

pkg/blobinfocache/boltdb/boltdb.go
```diff
@@ -22,6 +23,8 @@ var (
 
 	// uncompressedDigestBucket stores a mapping from any digest to an uncompressed digest.
 	uncompressedDigestBucket = []byte("uncompressedDigest")
+	// digestCompressorBucket stores a mapping from any digest to a compressor, or blobinfocache.Uncompressed
```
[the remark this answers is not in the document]
nalind: Adding `It may not exist in caches created by older versions, even if uncompressedDigestBucket is present.`  Let me know if you prefer your suggested text instead.

copy/copy.go
```diff
@@ -1408,18 +1411,41 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
 	if inputInfo.Digest != "" && uploadedInfo.Digest != inputInfo.Digest {
 		return types.BlobInfo{}, errors.Errorf("Internal error writing blob %s, blob with digest %s saved with digest %s", srcInfo.Digest, inputInfo.Digest, uploadedInfo.Digest)
 	}
+
 	if digestingReader.validationSucceeded {
+		bic := internalblobinfocache.FromBlobInfoCache(c.blobInfoCache)
```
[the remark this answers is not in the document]
nalind: Right, changing the type of that field to a `BlobInfoCache2`, since it isn't exported.

copy/copy.go
```diff
@@ -1408,18 +1411,41 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
...
 		// If compressionOperation != types.PreserveOriginal, we now have two reliable digest values:
 		// srcinfo.Digest describes the pre-compressionOperation input, verified by digestingReader
 		// uploadedInfo.Digest describes the post-compressionOperation output, computed by PutBlob
 		// (because inputInfo.Digest == "", this must have been computed afresh).
 		switch compressionOperation {
 		case types.PreserveOriginal:
-			break // Do nothing, we have only one digest and we might not have even verified it.
+			if desiredCompressionFormat.Name() != "" && isCompressed {
+				// we recompressed it
```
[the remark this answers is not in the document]
nalind: That makes a lot of sense.  Making that change.

copy/copy.go
```diff
@@ -1408,18 +1411,41 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
...
+				bic.RecordDigestCompressorName(uploadedInfo.Digest, desiredCompressionFormat.Name())
+			} else {
+				// it wasn't, and still isn't, compressed
+				bic.RecordDigestCompressorName(uploadedInfo.Digest, internalblobinfocache.Uncompressed)
+			}
+			// Do nothing else, we have only one digest and we might not have even verified it.
 		case types.Compress:
 			c.blobInfoCache.RecordDigestUncompressedPair(uploadedInfo.Digest, srcInfo.Digest)
+			if srcInfo.Digest != "" && !isCompressed {
```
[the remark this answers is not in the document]
nalind: When `digestingReader.validationSucceeded` is true, it looks like neither of the values is empty, so removing that check around the logic that's being added to save compressor info determined in the "Deal with layer compression/decompression if necessary" section.

docker/docker_image_dest.go
```diff
@@ -351,8 +359,15 @@ func (d *dockerImageDestination) TryReusingBlob(ctx context.Context, info types.
...
 			}
 		}
-		cache.RecordKnownLocation(d.ref.Transport(), bicTransportScope(d.ref), candidate.Digest, newBICLocationReference(d.ref))
-		return true, types.BlobInfo{Digest: candidate.Digest, Size: size}, nil
+
+		bic.RecordKnownLocation(d.ref.Transport(), bicTransportScope(d.ref), candidate.Digest, newBICLocationReference(d.ref))
+
+		compressionOperation, compressionAlgorithm, err := blobinfocache.OperationAndAlgorithmForCompressor(candidate.CompressorName)
+		if err != nil {
+			return false, types.BlobInfo{}, err
```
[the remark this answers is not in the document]
nalind: Yup, making sure we don't overwrite a previously-known compressor value for a given digest makes sense to me.

nalind:
Rebased.

nalind:
Rebased.

cuppett:
@mtrmac can this one be approved now or is there additional work needed?

nalind:
Rebased.

types/types.go
mtrmac: What happens with unknown compression now? If that happens (i.e. an old BoltDB cache that knows a location but no compression, or a third-party `BlobInfoCache` implementation), don’t we fall back to the old behavior of using that blob but not updating the MIME type to match?

pkg/blobinfocache/boltdb/boltdb.go
```diff
@@ -220,6 +227,30 @@ func (bdc *cache) RecordDigestUncompressedPair(anyDigest digest.Digest, uncompre
...
+			return err
+		}
+		key := []byte(anyDigest.String())
+		if previousBytes := b.Get(key); previousBytes != nil {
+			if string(previousBytes) != compressorName {
+				logrus.Warnf("Compressor for blob with digest %s previously recorded as %s, now %s", anyDigest, string(previousBytes), compressorName)
+			}
+		}
+		if compressorName == blobinfocache.UnknownCompression {
+			return b.Delete([]byte(anyDigest.String()))
```
mtrmac, line 248: (This can use `key` as well.)

copy/copy.go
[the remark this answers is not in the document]
mtrmac: `digestingReader` is initialized with the pre-`DecryptLayer` value, afterwards `srcInfo.Digest` is changed (and if I’m reading `DecryptLayer`, always to an empty value).

(That’s pre-existing, clearly less important than this bug, and I think in the worst case we’ll record a `""` compressor name entry that is never used by `TryReusingBlob`.)

docker/docker_image_dest.go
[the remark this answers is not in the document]
mtrmac: Basically I think this should (log and) `continue`, not fail. See the `Whatever happens here, don't abort the entire operation.` comment.

One way how this could happen is if the cache is updated by some future version to contain a currently-unknown `futureAlgorithm`; then `OperationAndAlgorithmForCompressor` fails.

(BTW we do overwrite a previously-known compressor value, and the BIC tests depend on it due to the `scopeName`+`compressorKnown` loop nesting..)

copy/copy.go
```diff
@@ -1307,11 +1316,20 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
...
 		logrus.Debugf("Using original blob without modification")
 		compressionOperation = types.PreserveOriginal
 		inputInfo = srcInfo
+		if isCompressed {
+			uploadCompressorName = compressionFormat.Name()
+			srcCompressorName = compressionFormat.Name()
+		} else {
+			uploadCompressorName = internalblobinfocache.Uncompressed
+			srcCompressorName = internalblobinfocache.Uncompressed
+		}
```
mtrmac, line 1332: (It might be tiny bit simpler to set `srcCompressorName` based on `DetectCompression` universally, then this can just be `uploadCompressorName = srcCompressorName` and the three other cases above are also covered.

It would require setting the “unknown” values in the encrypted case separately, but being explicit about that could be a good thing.)

pkg/blobinfocache/boltdb/boltdb.go
nalind: Oof, fixing.

copy/copy.go
```diff
@@ -1418,11 +1438,22 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
 			break // Do nothing, we have only one digest and we might not have even verified it.
 		case types.Compress:
 			c.blobInfoCache.RecordDigestUncompressedPair(uploadedInfo.Digest, srcInfo.Digest)
+			compressorNames[srcInfo.Digest] = internalblobinfocache.Uncompressed
 		case types.Decompress:
 			c.blobInfoCache.RecordDigestUncompressedPair(srcInfo.Digest, uploadedInfo.Digest)
+			compressorNames[uploadedInfo.Digest] = internalblobinfocache.Uncompressed
```
mtrmac, line 1444: Aren’t these redundant with `srcCompressorName`/`uploadCompressorName` now?

docker/docker_image_dest.go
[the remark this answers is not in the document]
mtrmac: > (BTW we do overwrite a previously-known compressor value

Ah, but `copy.go` does not.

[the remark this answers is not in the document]
nalind: Whoops, missed that.  Changing that to log and `continue`, like the `blobExists()`-returned-an-error section.

mtrmac (commented):
A full review; overall LGTM, mostly cleaning up the documentation/… details.

IIRC the only really notable outstanding question is about handling of “unknown” compressors now.

copy/copy.go
nalind: Okay, changing it.

copy/copy.go
```diff
@@ -654,7 +655,9 @@ func (c *copier) copyOneImage(ctx context.Context, policyContext *signature.Poli
 	if err != nil {
 		logrus.Debugf("Writing manifest using preferred type %s failed: %v", preferredManifestMIMEType, err)
 		// … if it fails, _and_ the failure is because the manifest is rejected, we may have other options.
```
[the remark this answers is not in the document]
nalind: Updating.

copy/copy.go
nalind: Yes, I think they are.  Removing them.

directory/directory_dest.go
```diff
@@ -194,7 +194,9 @@ func (d *dirImageDestination) PutBlob(ctx context.Context, stream io.Reader, inp
 // (e.g. if the blob is a filesystem layer, this signifies that the changes it describes need to be applied again when composing a filesystem tree).
 // info.Digest must not be empty.
 // If canSubstitute, TryReusingBlob can use an equivalent equivalent of the desired blob; in that case the returned info may not match the input.
-// If the blob has been successfully reused, returns (true, info, nil); info must contain at least a digest and size.
+// If the blob has been successfully reused, returns (true, info, nil); info must contain at least a digest and size, and may
+// include CompressionOperation and CompressionAlgorithm fields to indicate that a change to the compression type should be
+// reflected in the manifest that will be written.
```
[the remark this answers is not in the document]
nalind: Updating.

mtrmac:
> Dropped the "MediaType is always correct" goal because handling switching to zstd compression while updating a schema2 manifest wasn't looking doable without restructuring how the `copy` package was handling manifests.

(BTW this is still not entirely correct; e.g. with a `ForceManifestMIMEType` to schema2, or a non-OCI registry, `TryReusingBlob` can choose a zstd blob, and the copy will just fail instead of trying a different blob.

But that can happen only for users that choose the non-default `zstd`, so this PR is a clear improvement in a much more important case. Let’s defer this to some later time, maybe just after/when one of the `zstd:chunked` PRs introduces a way to extend/update `ImageDestination` without visible API incompatibilities.)

manifest/common.go
```diff
@@ -70,15 +69,15 @@ func compressionVariantMIMEType(variantTable []compressionMIMETypeSet, mimeType
...
 					}
-					return "", errors.New("uncompressed variant is not supported")
+					return "", ManifestLayerCompressionIncompatibilityError{fmt.Sprintf("uncompressed variant is not supported for type %q", mt)}
 				}
 				if name != mtsUncompressed {
 					return "", fmt.Errorf("unknown compression algorithm %s", name)
 				}
 				// We can't very well say “the idea of no compression is unknown”
-				return "", errors.New("uncompressed variant is not supported")
+				return "", ManifestLayerCompressionIncompatibilityError{fmt.Sprintf("uncompressed variant is not supported for type %q", mt)}
```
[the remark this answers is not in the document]
nalind: Making both return `ManifestLayerCompressionIncompatibilityError`, then.

manifest/common.go
```diff
@@ -99,6 +98,12 @@ func updatedMIMEType(variantTable []compressionMIMETypeSet, mimeType string, upd
 	// {de}compressed.
 	switch updated.CompressionOperation {
 	case types.PreserveOriginal:
+		// Force a change to the media type if we're being told to use a particular compressor,
+		// since it might be different from the one associated with the media type.  Otherwise,
+		// try to keep the original media type.
+		if updated.CompressionAlgorithm != nil && updated.CompressionAlgorithm.Name() != mtsUncompressed {
```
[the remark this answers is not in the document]
nalind: Okay, removing the second condition check.

manifest/common.go
```diff
@@ -116,3 +121,11 @@ func updatedMIMEType(variantTable []compressionMIMETypeSet, mimeType string, upd
 		return "", fmt.Errorf("unknown compression operation (%d)", updated.CompressionOperation)
 	}
 }
+
+type ManifestLayerCompressionIncompatibilityError struct {
```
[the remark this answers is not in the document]
nalind: Okay, I'm adding to the godoc for these.

manifest/common_test.go
```diff
@@ -87,3 +88,208 @@ func TestCompressionVariantMIMEType(t *testing.T) {
...
+		{oci1CompressionMIMETypeSets, imgspecv1.MediaTypeLayoutHeader, imgspecv1.MediaTypeLayoutHeader},
+		{oci1CompressionMIMETypeSets, imgspecv1.MediaTypeImageManifest, imgspecv1.MediaTypeImageManifest},
+		{oci1CompressionMIMETypeSets, imgspecv1.MediaTypeImageIndex, imgspecv1.MediaTypeImageIndex},
+		{oci1CompressionMIMETypeSets, imgspecv1.MediaTypeImageLayer, imgspecv1.MediaTypeImageLayer},
+		{oci1CompressionMIMETypeSets, imgspecv1.MediaTypeImageLayerGzip, imgspecv1.MediaTypeImageLayerGzip},
+		{oci1CompressionMIMETypeSets, imgspecv1.MediaTypeImageLayerZstd, imgspecv1.MediaTypeImageLayerZstd},
+		{oci1CompressionMIMETypeSets, imgspecv1.MediaTypeImageLayerNonDistributable, imgspecv1.MediaTypeImageLayerNonDistributable},
+		{oci1CompressionMIMETypeSets, imgspecv1.MediaTypeImageLayerNonDistributableGzip, imgspecv1.MediaTypeImageLayerNonDistributableGzip},
+		{oci1CompressionMIMETypeSets, imgspecv1.MediaTypeImageLayerNonDistributableZstd, imgspecv1.MediaTypeImageLayerNonDistributableZstd},
+		{oci1CompressionMIMETypeSets, imgspecv1.MediaTypeImageConfig, imgspecv1.MediaTypeImageConfig},
```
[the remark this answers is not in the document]
nalind: For these yes, cleaning that up.

manifest/common_test.go
```diff
@@ -87,3 +88,208 @@ func TestCompressionVariantMIMEType(t *testing.T) {
...
+	}
+	for i, c := range preserve {
+		update := types.BlobInfo{
+			MediaType:            c.mimeType,
+			CompressionOperation: types.PreserveOriginal,
+		}
+		updatedType, err := updatedMIMEType(c.compression, c.mimeType, update)
+		require.Nilf(t, err, "%d: updatedMIMEType(%q, %+v) failed unexpectedly", i, c.mimeType, update)
```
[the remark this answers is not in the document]
nalind: Fixing.

manifest/common_test.go
```diff
@@ -87,3 +88,208 @@ func TestCompressionVariantMIMEType(t *testing.T) {
...
+		require.Equalf(t, c.updatedType, updatedType, "%d: updatedMIMEType(%q, %+v)", i, c.mimeType, update)
```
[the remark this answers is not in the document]
nalind: Fixing.

copy/copy.go
```diff
@@ -1418,11 +1438,22 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
...
 		}
+		if uploadCompressorName != "" && uploadCompressorName != internalblobinfocache.UnknownCompression {
+			compressorNames[uploadedInfo.Digest] = uploadCompressorName
+		}
+		if srcCompressorName != "" && srcCompressorName != internalblobinfocache.UnknownCompression {
+			compressorNames[srcInfo.Digest] = srcCompressorName
+		}
+		for digest, compressorName := range compressorNames {
+			c.blobInfoCache.RecordDigestCompressorName(digest, compressorName)
+		}
```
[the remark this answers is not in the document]
nalind: Okay, I'll leave that for later, then.

manifest/common.go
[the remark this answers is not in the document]
nalind: Attempting to add a description as godoc.

nalind:
I don't have a good idea of what's supposed to happen when we try to use v2s2 and zstd compression.  With this PR, `skopeo -- copy --format v2s2 --dest-compress --dest-compress-format=zstd` from `docker` to `dir` fails, but I don't know about quietly changing the type of compression used, and it doesn't fall back to trying other manifest types.

mtrmac:
I think it’s perfectly fine — indeed the most correct behavior — for `--format v2s2 --dest-compress-format=zstd` to fail because the user explicitly requested something impossible.

The problematic case is
`skopeo copy --format oci --dest-compress-format=zstd $src docker://$registry/oci` (which succeeds) + `skopeo copy --format v2s2 $src docker://$registry/v2s2` (which can find and try to reuse the `zstd` blob AFAICS, and fail).

nalind:
Ah, and at that point it's too late to go back and tell `copyLayers()` to not reuse zstd blobs.  Having `determineManifestConversion()` return a list of acceptable compression types for the list of manifest types that it returns, and then passing that all the way down to `TryReusingBlob()` would probably work, ensuring that at least one of the manifest types we're prepared to try won't fail with `ManifestLayerCompressionIncompatibilityError`, but that's an API change to ImageDestination.

mtrmac:
Right. Let’s not do that in this PR.

---

Is there some resolution to https://github.com/containers/image/pull/1089#discussion_r557551086 ?

mtrmac:
> Ah, and at that point it's too late to go back and tell `copyLayers()` to not reuse zstd blobs.

… on second thought, at least a TODO comment at the `TryReusingBlob` would be nice; this might eventually be an unpleasant surprise, let’s make finding the cause a bit easier.

mtrmac:
… at the `TryReusingBlob` call site …

types/types.go
nalind: `OperationAndAlgorithmForCompressor()` gets passed `UnknownCompression` and returns `PreserveOriginal` with no specified compressor, and `dockerImageDestination.TryReusingBlob()` returns them in a `BlobInfo` with the original media type and no compression changes.  When that eventually gets to `manifest.updatedMIMEType()`, it doesn't make any changes.  So until we actually read that blob again, we will continue to behave as before when reusing that blob.

mtrmac: Isn’t “operate as before” #733? Some version of this PR skipped the blobs with `UnknownCompression`: https://github.com/containers/image/pull/1089#discussion_r539574676 .

That might force an extra unnecessary upload for schema1 (where the MIME type does not matter), but with schema1 disabled by docker/distribution by default, fixing #733 more reliably seems preferable.

nalind:
> Is there some resolution to [#1089 (comment)](https://github.com/containers/image/pull/1089#discussion_r557551086) ?

Followed up there -- the short version is that we get the current behavior until we eventually re-read the blob for whatever reason and get a chance to detect how it's compressed.  We could change `CandidateLocations2()` to ignore entries for which the compression type isn't known -  `pkg/blobinfocache/boltdb/cache.appendReplacementCandidates()` would just return if the `compressionBucket` wasn't there or if it failed to return a `compressorNameValue` for `digestKey`.

> … on second thought, at least a TODO comment at the `TryReusingBlob` would be nice; this might eventually be an unpleasant surprise, let’s make finding the cause a bit easier.

Added `TODO: at this point we don't know whether or not a blob we end up reusing is compressed using an algorithm that is acceptable for use on layers in the manifest that we'll be writing later, so if we end up reusing a blob that's compressed with e.g. zstd, but we're only allowed to write a v2s2 manifest, this will cause a failure when we eventually try to update the manifest with the digest and MIME type of the reused blob.  Fixing that will probably require passing more information to TryReusingBlob() than the current version of the ImageDestination interface lets us pass in.`

copy/copy.go
```diff
@@ -1423,6 +1452,12 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
 		default:
 			return types.BlobInfo{}, errors.Errorf("Internal error: Unexpected compressionOperation value %#v", compressionOperation)
 		}
+		if uploadedInfo.Digest != "" && uploadCompressorName != "" && uploadCompressorName != internalblobinfocache.UnknownCompression {
```
mtrmac, line 1455: `uploadedInfo.Digest` can’t be `""` (or we won’t be able to create a manifest).

docker/docker_image_dest.go
```diff
@@ -162,6 +163,7 @@ func (d *dockerImageDestination) PutBlob(ctx context.Context, stream io.Reader,
 
 	digester := digest.Canonical.Digester()
 	sizeCounter := &sizeCounter{}
+
```
mtrmac, line 166: (Absolutely non-blocking: Maybe drop this? `digester` is a pretty local variable to compute `computedDigest` in this ~paragraph. But then here I am having opinions about newlines…)

types/types.go
```diff
@@ -96,7 +96,8 @@ type LayerCompression int
 
 const (
 	// PreserveOriginal indicates the layer must be preserved, ie
-	// no compression or decompression.
+	// no compression or decompression, including already-compressed
+	// layers for which only the compression algorithm is being changed.
```
mtrmac, line 100: An absolutely non-blocking weak opinion: I kind of think about this case as “a special exception for `BlobInfo.Compression{Operation,Algorithm}`”, not a generic property of `PreserveOriginal`. So maybe document it only there?

OTOH it’s very reasonable to argue that readers of the code are likely to read documentation of the enum value rather than the `CompressionOperation` field.

Completely up to you.

mtrmac (commented):
> We could change `CandidateLocations2()` to ignore entries for which the compression type isn't known - `pkg/blobinfocache/boltdb/cache.appendReplacementCandidates()` would just return if the `compressionBucket` wasn't there or if it failed to return a `compressorNameValue` for `digestKey`.

That’s actually a good point: a previous version had the `UnknownCompression` check in `dockerImageDestination.TryReusingBlob`, but it’s not conceptually the transports’ responsibility to care. Defining `CandidateLocations2` to never include `UnknownCompression` entries would get that knowledge out of the transports, at least, and simplify the 2 cache implementations + the compatibility wrapper a bit — OTOH it would be three places responsible for the same check, and probably require test changes.

(Conceptually the decision ideally belongs to the copy pipeline around `copyLayer`, OTOH that would require the `TryReusingBlob` change we’re trying to avoid, and it would be non-trivial — a callback from `TryReusingBlob` back into the copy code to determine which candidate locations are acceptable?) So now I don’t have a preference WRT locating the check in `dockerImageDestination` vs. `CandidateLocations2`.

As for having the `UnknownCompression` check or not — I’d prefer to include it, so that users can just upgrade without having to manually clear caches. Is it blocking? I guess not (it can be added in a follow-up, after all), but it’s very close.

nalind:
> As for having the `UnknownCompression` check or not — I’d prefer to include it, so that users can just upgrade without having to manually clear caches. Is it blocking? I guess not (it can be added in a follow-up, after all), but it’s very close.

I lean that way myself, so I can change it.

copy/copy.go
```diff
@@ -1258,11 +1271,21 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
 	// === Deal with layer compression/decompression if necessary
 	var inputInfo types.BlobInfo
 	var compressionOperation types.LayerCompression
+	srcCompressorName := compressionFormat.Name()
```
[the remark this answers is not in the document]
nalind: Changing it to initialize as `Uncompressed`, and replace with `compressionFormat.Name()` if `isCompressed`.

copy/copy.go
```diff
@@ -1258,11 +1271,21 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
...
+	if srcCompressorName == "" {
+		srcCompressorName = internalblobinfocache.Uncompressed
+	}
+	uploadCompressorName := desiredCompressionFormat.Name()
+	if uploadCompressorName == "" {
+		uploadCompressorName = internalblobinfocache.Uncompressed
+	}
```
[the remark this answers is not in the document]
nalind: Moving the setting of `uploadCompressorName` into the "Deal with layer compression/decompression if necessary" `if` blocks, on the assumption that the blob is a layer, leaving the `if !canModifyBlob || isConfig` check that follows where it is, to match the test that sets `uploadedInfo.CompressionAlgorithm` for the opposite condition.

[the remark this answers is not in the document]
nalind: ... or not, per below.

copy/copy.go
nalind: Removing that check.

docker/docker_image_dest.go
nalind: You spend more time looking at this code than I do, so of course I'll drop it.

types/types.go
nalind: Reverting this bit.

copy/copy.go
```diff
@@ -1314,6 +1337,11 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
 		inputInfo = srcInfo
 	}
 
+	// If we can't modify the layer's blob, its compressor will be the same as the source's compressor, if known.
+	if !canModifyBlob || isConfig {
```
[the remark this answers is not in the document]
nalind: Took a while to convince myself of that, but making that change.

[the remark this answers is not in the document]
mtrmac: I had to double-check this works correctly as well.

Would a change (that I can prepare in a separate PR) to force `if isConfig {canModifyBlob=false}`, somewhere at the top of the function, help?

nalind:
>  OTOH it would be three places responsible for the same check, and probably require test changes.

Yes, it did.  Hopefully I got them all.

copy/copy.go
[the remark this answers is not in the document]
nalind: It'd be less for me to remember if/when I'm in this function again, but I'm content with whatever you prefer.

pkg/blobinfocache/boltdb/boltdb.go
```diff
@@ -251,21 +282,34 @@ func (bdc *cache) RecordKnownLocation(transport types.ImageTransport, scope type
...
 	}
+	compressorName := blobinfocache.UnknownCompression
+	if compressionBucket != nil {
+		// the bucket won't exist if the cache was created by a v1 implementation and
+		// hasn't yet been updated by a v2 implementation
+		if compressorNameValue := compressionBucket.Get(digestKey); len(compressorNameValue) > 0 {
+			compressorName = string(compressorNameValue)
+		}
+	}
+	if compressorName == blobinfocache.UnknownCompression && requireCompressionInfo {
```
mtrmac, line 300: Oops, I didn’t realize the `requireCompressionInfo` flag would be necessary for V1 compatibility; so this is not any simpler after all.

pkg/blobinfocache/memory/memory.go
```diff
@@ -101,14 +108,31 @@ func (mem *cache) RecordKnownLocation(transport types.ImageTransport, scope type
...
+		if !compressorKnown && requireCompressionInfo {
+			continue
+		}
 		candidates = append(candidates, prioritize.CandidateWithTime{
-			Candidate: types.BICReplacementCandidate{
-				Digest:   digest,
-				Location: l,
+			Candidate: blobinfocache.BICReplacementCandidate2{
+				Digest:         digest,
+				CompressorName: compressorName,
```
mtrmac, line 134:
```suggestion
				CompressorName: compressorName, // May be unset if !requireCompressionInfo
```
or something like that? Changing the code to actually set it to `UnknownCompression` is also an option, OTOH it’s essentially dead code and the extra assignment doesn’t make the code much easier to follow. I don’t have a strong preference.

internal/blobinfocache/types.go
```diff
@@ -0,0 +1,43 @@
...
+	// RecordDigestCompressorName records a compressor for the blob with the specified digest,
+	// or Uncompressed or UnknownCompression.
+	// WARNING: Only call this with LOCALLY VERIFIED data; don’t record a compressor for a
+	// digest just because some remote author claims so (e.g. because a manifest says so);
+	// otherwise the cache could be poisoned and cause us to make incorrect edits to type
+	// information in a manifest.
+	RecordDigestCompressorName(anyDigest digest.Digest, compressorName string)
+	// CandidateLocations2 returns a prioritized, limited, number of blobs and their locations
+	// that could possibly be reused within the specified (transport scope) (if they still
+	// exist, which is not guaranteed).
```
mtrmac, line 30:
```suggestion
	// exist, which is not guaranteed).
	//
	// The CompressorName fields in returned data must never be UnknownCompression.	
```

internal/blobinfocache/blobinfocache.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+type v1OnlyBlobInfoCache struct {
+	types.BlobInfoCache
+}
+
+func (bic *v1OnlyBlobInfoCache) RecordDigestCompressorName(anyDigest digest.Digest, compressorName string) {
+}
+
+func (bic *v1OnlyBlobInfoCache) CandidateLocations2(transport types.ImageTransport, scope types.BICTransportScope, digest digest.Digest, canSubstitute bool) []BICReplacementCandidate2 {
```
mtrmac, line 28: This should just return `nil` now AFAICS.

mtrmac (commented):
LGTM, some very last nits.

Thanks for all your work!

pkg/blobinfocache/memory/memory.go
nalind: Setting it to `UnknownCompression` for that case for consistency if nothing else.

internal/blobinfocache/types.go
nalind: Adding it.

internal/blobinfocache/blobinfocache.go
nalind: Yeah, it would have to.  Changing it.

nalind:
Okay, updated and back in the CI queue.

copy/copy.go
```diff
@@ -1390,7 +1414,7 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
 		}
```
mtrmac, line 1414: Oops:
Above, the
```go
	// If we can modify the layer's blob, set the desired algorithm for it to be set in the manifest.
	if canModifyBlob && !isConfig {
		uploadedInfo.CompressionAlgorithm = &desiredCompressionFormat
	}
```
condition is no longer a no-op if `CompressionOperation != types.Compress` (so just a simple copy of a `zstd` or uncompressed data is now changed to a `gzip` MIME type AFAICS).

I think replacing that condition with `uploadCompressorName != internalblobinfocache.UnknownCompression && uploadCompressorName != internalblobinfocache.Uncompressed` would work, but it feels a bit that there should be a more elegant way.

mtrmac: One possible approach is that the decision made has been indeed tracked in `uploadCompressorName`, so this could be
```go
if uploadCompressorName != internalblobinfocache.UnknownCompression && uploadCompressorName != internalblobinfocache.Uncompressed {
    algo, err := compression.AlgorithmByName(uploadCompressorName);
    if err { /* ??! */ }
    uploadedInfo.CompressionAlgorithm = &algo
}
```
which is _so_ ugly, but we can’t nicely and succinctly express `uploadCompressor = enum { unknown, uncompressed, algorithm(compression.Algorithm) }` in Go. Alternatively, the `if`/`else if` cascade could set yet another variable with the right `uploadedCompressionAlgorithm` value, risking inconsistency with `uploadCompressorName`.

So here’s hoping you have a much more elegant idea.

copy/copy.go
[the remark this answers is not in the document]
mtrmac: Filed https://github.com/containers/image/pull/1119 , but noticed https://github.com/containers/image/pull/1089#discussion_r558621134 :(

internal/blobinfocache/blobinfocache.go
```diff
@@ -0,0 +1,72 @@
...
+			Location:       c.Location,
+			CompressorName: UnknownCompression,
+		})
+	}
+	return results
+}
+
+// CandidateLocationsFromV2 converts a slice of BICReplacementCandidate2 to a slice of
+// types.BICReplacementCandidate, dropping compression information.
+func CandidateLocationsFromV2(v2candidates []BICReplacementCandidate2) []types.BICReplacementCandidate {
```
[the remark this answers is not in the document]
mtrmac: No longer possible with the `requireCompressionInfo` logic in callers.

copy/copy.go
nalind: I suppose we could initialize `desiredCompressionFormat` to `&c.compressionFormat`, and clear or change its value in the "Deal with layer compression/decompression if necessary" sections where we update `uploadCompressorName` with values that would conflict with it.  Then we could assign it to `uploadedInfo.CompressionAlgorithm` unconditionally.  The final `else` clause would still need to check `isCompressed` to decide between `&compressionFormat` and `nil`.  The additional diff looks more or less like https://gist.github.com/nalind/3e12d2be6f6858720851739de6a5d6ea.

mtrmac: Works for me.
- `s/desiredCompressionFormat/uploadCompressionFormat/` , then, I guess — it now records the outcome, not the input
- in the `!canModify` case, set it always to `nil` to unambiguously trigger no manifest edits.

nalind: Okay, added the gist patch, made these changes, and rebased.

mtrmac (approved):
LGTM. Thanks again!

Romain-Geissler-1A:
Hi,

Sorry for jumping into this pull request. I think I may suffer from this issue using podman + Artifactory. I wanted to try with the latest fedora rawhide which apparently uses podman from master branch everyday, but it looks like I still have the issue. Checking podman source code, apparently the version of containers/image used there is still 5.9.0 from last december. Would you know if the container team in Red Hat plans on updating containers/image in podman in January or February, and start pushing updated podman packages with this fix in fedora rawhide ?

Cheers,
Romain

rhatdan:
We should be cutting a new release of containers/image soon, and then this will get vendored into podman Most likely podman 3.1 or later.

jomeier:
Yes could you cut a new release, please? A pluriversum of developers is waiting for that.

jomeier:
Skopeo and all images mirrored to quay for OKD and the community operators, ... are also dependent on that.

mtrmac:
https://github.com/containers/image/pull/1128 is ongoing.

rhatdan:
Yes this fix with be in podman 3.0 and newer versions of skopeo and buildah which will show up in rhel8.4.

PR #1091, opened 2020-12-03
Tidy up TestParseShortNameMode

mtrmac:
Now that empty strings are rejected, don't include a valid value in the result.  This does not actualy change what is being tested, just makes it a tiny bit more readable.


vrothberg (commented):
LGTM

PR #1092, opened 2020-12-05
fix tarball example code

rcgoodfellow:
Looks like the `copy.Image` interface has changed, small update to example documentation for new interface.

### DIFF
--- a/tarball/doc.go
+++ b/tarball/doc.go
@@ -5,11 +5,13 @@
 //	package main
 //
 //	import (
-//		"fmt"
+//		"context"
 //
 //		cp "github.com/containers/image/v5/copy"
+//		"github.com/containers/image/v5/signature"
 //		"github.com/containers/image/v5/tarball"
 //		"github.com/containers/image/v5/transports/alltransports"
+//		"github.com/containers/image/v5/types"
 //		imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
 //	)
 //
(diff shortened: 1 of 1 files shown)


rcgoodfellow:
Also added default policy context, as the example will have a nil pointer deref without this.

tarball/doc.go
```diff
@@ -39,7 +41,17 @@
 //		if err != nil {
 //			panic(err)
 //		}
-//		err = cp.Image(nil, dest, src, nil)
+//
+//		policy, err := signature.DefaultPolicy(&types.SystemContext{ /*...*/ })
```
mtrmac, line 45: `DefaultPolicy(nil)` would be the simplest use, though the example given is of course valid.

tarball/doc.go
```diff
@@ -39,7 +41,17 @@
...
+//		if err != nil {
+//			panic(err)
+//		}
+//
+//		pc, err := signature.NewPolicyContext(policy)
+//		if err != nil {
+//			panic(err)
+//		}
```
mtrmac, line 53:
```suggestion
//		}
//		defer pc.Destroy()
```
needs to be here.

mtrmac (commented):
Thanks! At least one more thing was missing.

tarball/doc.go
rcgoodfellow: Changed.

tarball/doc.go
rcgoodfellow: Added.

rcgoodfellow:
Comments, addressed. Thanks for the great library.

rhatdan:
LGTM

TomSweeneyRedHat:
Needs a rebase

TomSweeneyRedHat:
LGTM once rebased

mtrmac (approved):
LGTM.

Thanks!

mtrmac:
(Force-merging because that part of the code changes so infrequently that linear history is not much of a concern, and the CI is depressingly slow lately.)

PR #1099, opened 2020-12-09
Add logMirrorChoice to ctx for log

QiWang19:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1881694
This BZ needs change the log level of the image pull location from the debug to info.
Add DockerLogMirrorChoice to types.SystemContext for keeping log level of the physical pull source of images.
Crio could set it to log the image source if it's from a mirror.

### DIFF
--- a/docker/docker_image_src.go
+++ b/docker/docker_image_src.go
@@ -64,6 +64,11 @@ func newImageSource(ctx context.Context, sys *types.SystemContext, ref dockerRef
 	}
 	attempts := []attempt{}
 	for _, pullSource := range pullSources {
+		if sys.DockerLogMirrorChoice {
+			logrus.Infof("Trying to access %q", pullSource.Reference)
+		} else {
+			logrus.Debugf("Trying to access %q", pullSource.Reference)
+		}
 		logrus.Debugf("Trying to access %q", pullSource.Reference)
 		s, err := newImageSourceAttempt(ctx, sys, ref, pullSource)
 		if err == nil {
--- a/types/types.go
+++ b/types/types.go
@@ -600,6 +600,8 @@ type SystemContext struct {
 	DockerDisableV1Ping bool
 	// If true, dockerImageDestination.SupportedManifestMIMETypes will omit the Schema1 media types from the supported list
 	DockerDisableDestSchema1MIMETypes bool
+	// If true, the physical pull source of docker transport images logged as info level
+	DockerLogMirrorChoice bool
 	// Directory to use for OSTree temporary files
 	OSTreeTmpDirPath string
 


rhatdan:
LGTM

TomSweeneyRedHat (commented):
LGTM

umohnani8:
LGTM

mtrmac:
I get that this is an “easy quick improvement”, and I don’t have any similarly easy suggestion, but I’m not too happy about a library just logging like this without giving the caller much say in where the log goes, especially on such a frequent path that everyone is going to hit.

Conceptually this probably belongs into the `copy.Options.Progress` log stream, as a well-typed data structure — but getting the data there would involve changing the `ImageSource` interface.

QiWang19:
Could you explain why it would be better to be in `copy.Options.Progress` log stream, I am not very familiar with it, what is  `copy.Options.Progress` log stream,  related with https://github.com/containers/image/blob/master/copy/progress_reader.go?  
Will that log show up in crio logs?

mtrmac:
Yes, that’s the major current producer (+ one event after `TryReusingBlob` succeeds). The point is to make these things observable by _software_ as an API, not to produce a formatted text that’s impossible to work with because it can change any time.

Notably CRI-O is using that interface already: https://github.com/cri-o/cri-o/blob/8a05a29873456911db9a5c85a8d569e4a0db0ee3/server/image_pull.go#L171

It’s, for the most part, not the job of c/image to provide an UI, but it should allow callers to create an UI. (OTOH, admittedly, `copy.Image` does create progress bars, and it is expected to write _something undefined_ to `copy.Options.ReportWriter`.) 

“`copy.Image` is required to write important information to the log stream” just does not feel like a good fit in that context; it’s not a sustainable way to define or maintain an API.

Or maybe the mirror choice could go to `ReportWriter`, sure. That would make it user-readable but not software-consumable, not as good as `Progress` but better than just the indiscriminate log.

Either way we’re looking at some kind of API break to `ImageSource`, something that we do need to be able to do _eventually_ but much more work than this is probably worth — the customers in this case _could_ just bump CRI-O to debug mode for the time necessary to test mirroring (or, well, monitor the network externally), if I understand the situation correctly.

---

I suppose one way to shut me up is to gate this behind a `types.SystemContext.DockerLogMirrorChoice` opt-in or something like that, making it an ”obscure implementation aspect of the Docker transport that CRI-O specifically knows about”, rather than a design pattern in c/image. OTOH that also could become a very bad precedent.

docker/docker_image_src.go
```diff
@@ -67,6 +67,7 @@ func newImageSource(ctx context.Context, sys *types.SystemContext, ref dockerRef
 		logrus.Debugf("Trying to access %q", pullSource.Reference)
 		s, err := newImageSourceAttempt(ctx, sys, ref, pullSource)
 		if err == nil {
+			sys.DockerLogPullSource = s.physicalRef.ref.Name()
```
mtrmac, line 70: Hum. That’s certainly an inventive way to not break the API, but it’s too surprising a way to be maintainable. Generally code assumes that things like https://github.com/containers/image/blob/122b16af75d13b9dacda1b04111fa7fdffde30c5/docker/docker_image_src.go#L108-L111 are fine to do.

QiWang19:
@mtrmac PTAL. Change to use the `DockerLogMirrorChoice` to avoid send the sys back to cri-o.

mtrmac (commented):
LGTM. I’m not happy about this but blocking this small improvement on building infrastructure for `ImageSource` API changes is not proportionate.

PR #1109, opened 2020-12-26
Support additional layer store (patch for containers/image)

ktock:
https://github.com/containers/podman/issues/4739

## Reconsidered the design on 2021/1/12

This enables podman to create containers using layers stored in a specified directory instead of pulling them from the registry. Leveraging this feature with remotely-mountable layers provided by stargz/zstd:chunked or CVMFS, podman can achieve lazy pulling.

### Changes in contianers/storage: https://github.com/containers/storage/pull/795

That directory is named "additional layer store" (call ALS in this doc) and has the following structure.

```
<ALS root>/base64("key1=value1")/base64("key2=value2")/.../
`-- diff
`-- info
```

`diff` directory contains the extracted layer diff contents specified by the key-value pairs.
`info` file contains `*c/storage.Layer` struct that indicates the information of this layer contents (`*c/storage.Layer.ID`, `*c/storage.Layer.Parent` and `*c/storage.Layer.Created` can be empty as it'll be filled by `c/storage.Store`).

On each pull, `c/storage.Store` searches the layer diff contents from ALS using pre-configured key-value pairs.
Each key-value pair is base64 encoded.
By default, the following `key=value` pairs can be used as elements of the path in ALS,

- `reference=<image reference>`
- `layerdigest=<digest of the compressed layer contents>`

Additionally, layer annotations (defined in the image manifest) prefixed by `containers/image/target.` can be used as well.
The prefix `containers/image/target.` will be trimmed from the key when it's used in the path on ALS.

Overlay driver supports an option to specify which key-value pair to be used and how the order should they be when `c/storage.Store` searches layers in the ALS.

```
layerstore=<ALS root directory>:<key1>:<key2>:...:<keyX>
```

In the above case, on each pull, `c/storage.Store` searches the following path in ALS,

```
<ALS root>/base64("key1=value1")/base64("key2=value2")/.../base64("keyX=valueX")/
`-- diff
`-- info
```

The underlying filesystem (e.g. stargz/zstd:chunked-based filesystem or CVMFS) should show the exploded view of the target layer diff and its information at these locations.
Example filesystem implementation (currently stargz-based) is https://github.com/ktock/stargz-snapshotter/tree/als-pool-example (this must be mounted on <ALS root>)

If the layer content is found in ALS, `c/storage.Store` creates layer using `<ALS root>/.../info` as `*c/storage.Layer` and using `<ALS root>/.../diff` as its diff directory.
So `c/image`'s copier doesn't need to pull this layer from the registry.

### Changes in containers/image: https://github.com/containers/image/pull/1109

Now `c/image`'s copier leverages this store. 
Every time this pulls an image, it first tries to reuse blobs from ALS.
That copier passes each layer's OCI annotations (key-value pairs) + the following key-value to `c/storage.Store`.

- containers/image/target.reference: The image reference of an image that contains the target layer.  This is needed for supporting registry-backed filesystems (e.g. estargz, zstd:chunked).

`*c/image.storageImageDestination.TryReusingBlob()` cannot pass image reference to `c/storage.Store` so this commit adds a new API `*c/image.storageImageDestination.TryReusingBlobWithRef() for achieving this.
When this copier successfully acquires that layer, this reuses this layer without pulling.

### Command exapmle

```
podman --storage-opt "layerstore=/tmp/storage:reference:layerdigest" pull ghcr.io/stargz-containers/rethinkdb:2.3.6-esgz
podman --storage-opt "layerstore=/tmp/storage:reference:layerdigest" run --rm -it ghcr.io/stargz-containers/rethinkdb:2.3.6-esgz /bin/echo 'Hello, World!'
```

In the above cases, `c/storage.Store` looks up `/tmp/storage/base64("reference=ghcr.io/stargz-containers/rethinkdb:2.3.6-esgz")/base64(<layer digests>)/{diff, info}` in ALS.
The example filesystem implementation (https://github.com/ktock/stargz-snapshotter/tree/als-pool-example) is mounted at `/tmp/storage` shows the extracted layer at that location.
Then `rethinkdb:2.3.6-esgz` can run without pulling it from registry.

## Known limitation and request for comments

Some operations (e.g. save) requires correct value to be set to `c/storage.Layer.UncompressedSize`. This field seems to be the size of the layer but without compression (i.e. the size of the tar-archived format of that layer). For registry-backed ALS, getting this information is difficult because neither of OCI/Docker image nor registry API provides the way to get the uncompressed size of layers. We cannot get this information without actually pull and decompress the layer, which is not lazy pulling that this PR aims to.

I'll check the codebase deeper to come up with the way to get this information from somewhere or the way to safely allow `c/storage.Layer.UncompressedSize` to be *unknown* during operations. But if someone has good idea for managing this, please let me know.

cc: @siscia @giuseppe


copy/copy.go
```diff
@@ -862,6 +869,14 @@ func (ic *imageCopier) copyLayers(ctx context.Context) error {
 		}()
 
 		for i, srcLayer := range srcInfos {
+			if srcLayer.Annotations == nil {
+				srcLayer.Annotations = make(map[string]string)
+			}
+			// Add default labels containing information of the target image and layers.
+			srcLayer.Annotations[layerTargetDigest] = srcLayer.Digest.String()
+			if dref := ic.c.rawSource.Reference().DockerReference(); dref != nil {
+				srcLayer.Annotations[layerTargetReference] = dref.String()
```
mtrmac, line 878: - This exposes private information; it’s not expected that the destination of a copy is told about the name used at the source.
- Why the full Docker reference, anyway? c/storage keys things off an image ID, and a Docker reference can refer to many different images over time.

copy/copy.go
```diff
@@ -862,6 +869,14 @@ func (ic *imageCopier) copyLayers(ctx context.Context) error {
 		}()
 
 		for i, srcLayer := range srcInfos {
+			if srcLayer.Annotations == nil {
+				srcLayer.Annotations = make(map[string]string)
+			}
+			// Add default labels containing information of the target image and layers.
+			srcLayer.Annotations[layerTargetDigest] = srcLayer.Digest.String()
```
mtrmac, line 876: This seems suspect. The manifest pretty much already exists to contain digests; why duplicate that?

If the intent of that annotation is to contain the DiffID (uncompressed digest), that’s 1) undocumented, 2) not implemented that way.

mtrmac (commented):
Just a very quick initial skim to highlight concerns. I have _literally not read the PR description in full yet_.

How does this relate to #1084 ?

If this extends the image format, there should eventually be formal documentation for interoperability — maybe after the code settles, sure.

copy/copy.go
ktock: Registry-backed additional layer store filesystem (e.g. eStargz, zstd:chunked) needs to query layer data to the registry. The registry API [requires image reference](https://docs.docker.com/registry/spec/api/#pulling-a-layer) so this needs to be passed to the c/storage's `Store` and the underlying filesystem somehow.

copy/copy.go
ktock: This layer digest is stored to the annotations for being passed to the c/storage's `Store`.
ALS accessed via c/storage's `Store` can know which layers should be shown to c/storage.
This is especially needed for registry-backed ALS (e.g. eStargz, zstd:chunked).

ktock:
> How does this relate to #1084 ?

The underlying filesystem (additional layer store) accessed from c/storage can support mounting layers from the registry leveraging eStargz/zstd:chunked.

copy/copy.go
mtrmac: Having now read a bit more, https://github.com/containers/storage/pull/795#discussion_r551319228 : this is just data that should be passed as parameters, not by modifying the manifest, which affects every single `skopeo copy` including things like inter-registry copies exposing locations of private staging registries when making a product release.

(Even given that, it’s not at all clear to to me how this is going to work with authenticated registries; the credentials used by `copy.Image` are not made available to the on-demand additional store implementations by this. But that’s a secondary concern and probably solvable at least for the simple cases.)

copy/copy.go
mtrmac: Again, https://github.com/containers/storage/pull/795#discussion_r551319228 . `TryReusingBlob` already has an _authoritative_ value of this digest directly as `srcInfo.Digest`; it’s not necessary to pass another copy, and especially not in this way.

storage/storage_image.go
```diff
@@ -654,6 +667,14 @@ func (s *storageImageDestination) Commit(ctx context.Context, unparsedToplevel t
 		// this layer blob to its parent, if it has one, or the blob's hex value otherwise.
 		diffID, haveDiffID := s.blobDiffIDs[blob.Digest]
 		if !haveDiffID {
+			if lid, ok := s.blobLayerIDs[blob.Digest]; ok {
+				if layer, err := s.imageRef.transport.store.Layer(lid); layer != nil && err == nil {
+					// The exploded layer (chain) found in the store with the
+					// specified custom ID.
+					lastLayer = layer.ID
```
mtrmac, line 674: This is wrong; a `*c/storage.Layer` object refers to the full chain (all parent layers + the layer identified by `diffID` on top).

But `blobLayerIDs` is built _only_ using the annotations, i.e. an image reference and a _single layer_ digest — the key of `blobLayerIDs` is just a _single layer_ digest; i.e. it can’t differentiate between chains (root1, sameDigest); (root2, …, sameDigest); and (sameDigest) with no parent.

That’s why the existing code always creates `lastLayer` with a deterministic ID that is computed from the full chain; ignoring that chain and setting `lastLayer` based on the top layer AFAICS just doesn’t work.

If there is some backing implementation that provides overlay-formatted single-layer data, c/storage will probably have to expose an operation that, given a parent `*c/storage.Layer` and some kind of ID of that single-layer data, creates a new `*c/storage.Layer` object that represents mounting of that overlay-formatted single-layer data on top of the submitted parent.

storage/storage_image.go
```diff
@@ -759,6 +780,12 @@ func (s *storageImageDestination) Commit(ctx context.Context, unparsedToplevel t
 	if intendedID == "" {
 		intendedID = s.computeID(man)
 	}
+	for _, lid := range s.blobLayerIDs {
```
mtrmac, line 783: `blobLayerIDs` is a map, making this value non-deterministic.

storage/storage_image.go
```diff
@@ -759,6 +780,12 @@ func (s *storageImageDestination) Commit(ctx context.Context, unparsedToplevel t
...
+		// If we use custom layer IDs for creating an image, we should include this
+		// information as well. This prevents the conflict of between image ID of an image
+		// with custom ID and the one without custom ID.
+		intendedID = digest.SHA256.FromString(fmt.Sprintf("%s %s", intendedID, lid)).Encoded()
```
mtrmac, line 787: What is a “custom ID”?

Why does the presence of additional-layer-store layers matter at all? It’s all the same image, so it should be deduplicated, shouldn’t it?

mtrmac:
> > How does this relate to #1084 ?
> 
> The underlying filesystem (additional layer store) accessed from c/storage can support mounting layers from the registry leveraging eStargz/zstd:chunked.

So should we have only one of the two features? If so, which one? Or if both, why have two so different implementations working with the same format? (I don’t have any opinion yet, I assume you have thought about these things much more than me.)

copy/copy.go
giuseppe: could we use a lookaside cache for storing such information (something implemented with: https://github.com/containers/storage/pull/787)?

mtrmac: What if 3 different images that share a layer are pulled, one after another; 2 from a single registry, the third from a completely different registry; the three pulls are using different credentials, and later _some_ of the credentials are invalidated? The credentials are conceptually a property of (end user, repository), not a layer.  (And the concept of “end user” gets rather tricky once CRI-O is involved.)

I suppose just starting with unauthenticated repositories would still be useful. Getting the complex cases right might be more interesting.

ktock:
> > > How does this relate to #1084 ?

> > The underlying filesystem (additional layer store) accessed from c/storage can support mounting layers from the registry leveraging eStargz/zstd:chunked.

> So should we have only one of the two features? If so, which one? Or if both, why have two so different implementations working with the same format? (I don’t have any opinion yet, I assume you have thought about these things much more than me.)

We should have both of eStargz (fully compatible with legacy gzip-based images) for backward-compatibility to the current ecosystem and zstd:chunked (zstd-based) for "skippable frame" support & possibly better performance.

mtrmac:
I was not asking about the formats, but about the two structurally different implementations of zstd:chunked .

ktock:
This patch is for enabling "lazy pulling" (skipping pulling of layers but fetching necessary chunks on-demand). But #1084 is for deduplication of pulling so the scope is different.

copy/copy.go
ktock: > this is just data that should be passed as parameters, not by modifying the manifest, which affects every single skopeo copy including things like inter-registry copies exposing locations of private staging registries when making a product release.

Reference is now pased as parameters to `c/storage.Store`, through a new API `*c/image.storageImageDestination.TryReusingBlobWithRef()`.

ktock: > Even given that, it’s not at all clear to to me how this is going to work with authenticated registries

As initial implementation, the underlying filesystem's authentication logic (creds management) is separated from `copy.Image`.
We can come up with the way to sync creds between them but this can be following-up PRs.

copy/copy.go
ktock: > Again, containers/storage#795 (comment) . TryReusingBlob already has an authoritative value of this digest directly as srcInfo.Digest; it’s not necessary to pass another copy, and especially not in this way.

Fixed this.
Now `*c/image.storageImageDestination.TryReusingBlobWithRef()` recieves digest through `srcInfo.Digest`.

storage/storage_image.go
```diff
@@ -486,6 +489,16 @@ func (s *storageImageDestination) TryReusingBlob(ctx context.Context, blobinfo t
 		}, nil
 	}
 
+	// Check if the layer can be used from the additional layer store with fresh ID.
+	if l, err := s.imageRef.transport.store.Layer(stringid.GenerateRandomID(), storage.WithAnnotations(blobinfo.Annotations)); err == nil {
```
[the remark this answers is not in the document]
ktock: > Just add a new lookup method that receives the data relevant for the lookup, similar to the existing LayersByCompressedDigest and the like.

Introduced a new API `c/storage.Store.LayerFromAdditionalLayerStore(d digest.Digest, labels map[string]string) (*Layer, error)`.

storage/storage_image.go
```diff
@@ -486,6 +489,16 @@ func (s *storageImageDestination) TryReusingBlob(ctx context.Context, blobinfo t
...
+		s.blobLayerIDs[blobinfo.Digest] = l.ID
```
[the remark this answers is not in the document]
ktock: > With blobDiffIDs not set, this will cause computeID to fail.

Fixed this. Now filesystem (ALS) provides diffID.

storage/storage_image.go
ktock: > This is wrong; a *c/storage.Layer object refers to the full chain (all parent layers + the layer identified by diffID on top).

Fixed this.
Now `c/storage.Store` has a new API `PutFromAdditionalLayerStore(id, parent string, d digest.Digest, labels map[string]string) (*Layer, error)` for enabling `c/image` to always creates lastLayer with a ID that is computed from the full chain.

storage/storage_image.go
ktock: > blobLayerIDs is a map, making this value non-deterministic.

Now the order can be determined by the driver option `layerstore` (please also see the description of this PR in the above).

storage/storage_image.go
ktock: > What is a “custom ID”?

Now we don't have it.

copy/copy.go
```diff
@@ -1051,7 +1051,7 @@ type diffIDResult struct {
 
 // copyLayer copies a layer with srcInfo (with known Digest and Annotations and possibly known Size) in src to dest, perhaps compressing it if canCompress,
 // and returns a complete blobInfo of the copied layer, and a value for LayerDiffIDs if diffIDIsNeeded
-func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, toEncrypt bool, pool *mpb.Progress) (types.BlobInfo, digest.Digest, error) {
+func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, toEncrypt bool, pool *mpb.Progress, srcRef reference.Named) (types.BlobInfo, digest.Digest, error) {
```
mtrmac, line 1054: The function documentation should say that `srcRef` may be `nil`.

copy/copy.go
```diff
@@ -1076,6 +1076,28 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, to
 			}
 			return blobInfo, cachedDiffID, nil
 		}
+
+		rDest, ok := ic.c.dest.(types.ImageDestinationReusableWithRef)
+		if ok {
+			reused, blobInfo, err = rDest.TryReusingBlobWithRef(ctx, srcInfo, srcRef)
```
mtrmac, line 1082: We should call only one of the `TryReusingBlob` methods, so that a transport can implement both (to be usable by old clients as well as new) and we don’t try to reuse everything twice. That would also reduce duplication in the consumption of the results.

copy/copy.go
mtrmac: > We can come up with the way to sync creds between them but this can be following-up PRs.

I can’t much imagine how that would look like; so is this “will never happen”? The tradeoffs are relevant for choosing an an approach/implementation.

storage/storage_image.go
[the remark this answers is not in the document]
mtrmac: How does the filesystem obtain the diffID? By reading all of the layer? If so, wouldn’t it be more efficient to avoid it and just read all of the layer ourselves? (Or is that somehow cached across machines?)

storage/storage_image.go
```diff
@@ -682,6 +706,15 @@ func (s *storageImageDestination) Commit(ctx context.Context, unparsedToplevel t
 			lastLayer = layer.ID
 			continue
 		}
+		if ref, ok := s.blobRefs[blob.Digest]; ok {
+			layer, err := s.imageRef.transport.store.PutLayerFromAdditionalLayerStore(id, lastLayer, nil, blob.Digest, ref.String())
```
mtrmac, line 710: If the `TryReusing…` operation has already looked up the layer, wouldn’t the lookup be able to return some kind of (opaque?) identifier so that here we don’t have to look it up using (digest, ref) again?

types/types.go
```diff
@@ -331,6 +331,14 @@ type ImageDestination interface {
 	Commit(ctx context.Context, unparsedToplevel UnparsedImage) error
 }
 
+type ImageDestinationReusableWithRef interface {
+	ImageDestination
+
+	// TryReusingBlobWithRef checks whether the destination can efficiently reuse a blob, and if so, applies it to the current destination.
+	// This allows the destination to search the blob contents with taking the image reference into consideration.
+	TryReusingBlobWithRef(ctx context.Context, blobinfo BlobInfo, ref reference.Named) (bool, BlobInfo, error)
```
mtrmac, line 339: See elsewhere about callers calling only one of the two `TryReusing*` methods.

types/types.go
```diff
@@ -331,6 +331,14 @@ type ImageDestination interface {
 	Commit(ctx context.Context, unparsedToplevel UnparsedImage) error
 }
 
+type ImageDestinationReusableWithRef interface {
```
mtrmac, line 334: Probably should be `WithSrcRef`, or at the very least clearly documented.

mtrmac, line 334: Let’s at least try to stop making the write side public; define this in `internal/types` or something like that.

RFC: @nalind — is this practical to do with the blob cache layer?

copy/copy.go
ktock: Added comments.

copy/copy.go
ktock: Fixed it by making `TryReusingBlobWithSrcRef` be a superset of `TryReusingBlob`.

storage/storage_image.go
ktock: Fixed to avoid looking up multiple times, by introducing a new object `c/storage.AdditionalLayer` that points to the looked up layer.

types/types.go
ktock: Fixed it by making `TryReusingBlobWithSrcRef` be a superset of `TryReusingBlob`.

types/types.go
ktock: Fixed it.

ktock: Moved this to `internal/types`.

ktock:
>> We can come up with the way to sync creds between them but this can be following-up PRs.

> I can’t much imagine how that would look like; so is this “will never happen”? The tradeoffs are relevant for choosing an an approach/implementation.

I think the file system API isn't enough for achieving this. Another dedicated communication channel (e.g. unix socket) will need to be exposed by the filesystem (FUSE daemon) for communicating creds with the runtime (Podman, CRI-O, ...).

AkihiroSuda:
needs rebase

rhatdan:
@ktock Could you rebase this PR>]?

ktock:
Rebased.

rhatdan:
And sadly it needs to be rebased again.

ktock:
@giuseppe Do we need a new tag for https://github.com/containers/storage/pull/795 to make CI of this PR pass?

rhatdan:
I just released containers/storage v1.29.0

ktock:
> I just released containers/storage v1.29.0

Thanks!

ktock:
Error in CI (**`Skopeo`**):

```
time="2021-04-12T23:06:07Z" level=fatal msg="Error copying ref \"docker://k8s.gcr.io/k8s-dns-kube-dns:1.14.12\": Error writing blob: error happened during read: read tcp 10.88.0.6:36496->64.233.191.128:443: read: connection reset by peer"
```

Seems unrelated to this patch? :thinking:

giuseppe:
@ktock could you please rebase?

ktock:
@giuseppe
rebased :)

giuseppe (approved):
LGTM

@vrothberg @rhatdan @mtrmac PTAL

rhatdan:
I would like @mtrmac and/or @vrothberg Approval.

storage/storage_image.go
```diff
@@ -532,14 +537,37 @@ func (s *storageImageDestination) PutBlob(ctx context.Context, stream io.Reader,
...
-	reused, info, err := s.TryReusingBlob(ctx, blobinfo, options.Cache, options.CanSubstitute)
+	reused, info, err := s.tryReusingBlobWithSrcRef(ctx, blobinfo, options.Cache, options.CanSubstitute, options.SrcRef)
 	if err != nil || !reused || options.LayerIndex == nil {
 		return reused, info, err
 	}
 
 	return reused, info, s.queueOrCommit(ctx, info, *options.LayerIndex)
 }
 
+func (s *storageImageDestination) tryReusingBlobWithSrcRef(ctx context.Context, blobinfo types.BlobInfo, cache types.BlobInfoCache, canSubstitute bool, ref reference.Named) (bool, types.BlobInfo, error) {
```
vrothberg, line 548: Please add a comment to this function.

storage/storage_image.go
```diff
@@ -553,6 +581,11 @@ func (s *storageImageDestination) TryReusingBlob(ctx context.Context, blobinfo t
 	// lock the entire method as it executes fairly quickly
 	s.lock.Lock()
 	defer s.lock.Unlock()
+
+	return s.tryReusingBlobLocked(ctx, blobinfo, cache, canSubstitute)
+}
+
+func (s *storageImageDestination) tryReusingBlobLocked(ctx context.Context, blobinfo types.BlobInfo, cache types.BlobInfoCache, canSubstitute bool) (bool, types.BlobInfo, error) {
```
vrothberg, line 588: Please add a comment to this function.

vrothberg (commented):
The changes look good to me but I miss a verbose commit message that explains the details. Why is the change needed? What's an additional layer store? etc.

ktock:
@vrothberg Thanks for the review. 
Added comments and a commit message.

ktock:
CI failure (Skopeo) seems not related to this PR

```
time="2021-04-21T07:52:31Z" level=fatal msg="Error reading blob sha256:f7794a8687064c222edec1eacac34186676299758221ad448828a138c313432c: error fetching external blob from \"https://mcr.microsoft.com/v2/windows/nanoserver/blobs/sha256:f7794a8687064c222edec1eacac34186676299758221ad448828a138c313432c\": 500 (Internal Server Error)"
```

vrothberg:
> CI failure (Skopeo) seems not related to this PR
> 
> ```
> time="2021-04-21T07:52:31Z" level=fatal msg="Error reading blob sha256:f7794a8687064c222edec1eacac34186676299758221ad448828a138c313432c: error fetching external blob from \"https://mcr.microsoft.com/v2/windows/nanoserver/blobs/sha256:f7794a8687064c222edec1eacac34186676299758221ad448828a138c313432c\": 500 (Internal Server Error)"
> ```

I concur and restarted the CI job.

vrothberg (commented):
LGTM, thank you!

@mtrmac, if you have time to skim the changes. I'd feel more comfortable with your blessings.

ktock:
Can we move this forward?

vrothberg:
> Can we move this forward?

Thanks for the ping. Yes, let's get it merged. Can you rebase?

ktock:
Rebased :)

giuseppe:
@ktock do you have some instructions how I can play with the stargz-snapshotter and additional layer stores?

How do I run it and what setup to use for `storage.conf`?

ktock:
@giuseppe Does https://github.com/containerd/stargz-snapshotter/blob/v0.7.0/docs/INSTALL.md#install-stargz-store-for-cri-opodman-with-systemd help?

giuseppe:
thanks that is helpful!  I've followed the installation tutorial and I think it is running now.  Do you have any image ready I can use to lazy pull?

ktock:
@giuseppe There are pre-converted estargz images on ghcr.io/stargz-containers : https://github.com/containerd/stargz-snapshotter/blob/main/docs/pre-converted-images.md

Other methods to get estargz images: https://github.com/containerd/stargz-snapshotter#getting-estargz-images

giuseppe:
Thanks!  That worked well

giuseppe:
is there also support for zstd:chunked?  I've tried with `docker.io/gscrivano/zstd-chunked:fedora` but it failed

ktock:
That is on https://github.com/containerd/stargz-snapshotter/pull/293 but hasn't merged yet. I'll work on it soon.

rhatdan:
@ktock We are looking to packge stargz-snapshotter as a separate package in Fedora.  @lsm5 is taking the lead on this, are you interested in working on it? If we are going to truly allow users to use this method of pulling containers, we need to make it as easy as possible for them to do it.

This includes building and pushing images to registries, and then configuring container/storage to use the fuse snapshotter file system out of the box.

ktock:
@rhatdan 

> We are looking to packge stargz-snapshotter as a separate package in Fedora. @lsm5 is taking the lead on this, are you interested in working on it?

Yes, I am. Thank you!

giuseppe:
I am adding support for estargz to the chunked package in containers/storage.

@ktock what is the easiest way to create estargz images?

Is it worth to support the original stargz format as well or we can safely assume estargz is the format used by containerd?

ktock:
@giuseppe 

> I am adding support for estargz to the chunked package in containers/storage.

Thank you!

> what is the easiest way to create estargz images?

### On command line:

You can use `nerdctl image convert` ([doc](https://github.com/containerd/nerdctl/blob/v0.11.1/docs/stargz.md#building-stargz-images-using-nerdctl-build)) to convert an arbitrary image into eStarz.
You can perform the conversion using containerized nerdctl.

```
docker build -t nerdctl https://github.com/containerd/nerdctl.git
docker run -v ~/.docker/config.json -it --rm --privileged nerdctl /bin/bash
# nerdctl pull ghcr.io/stargz-containers/python:3.9-org
# nerdctl image convert --estargz --oci ghcr.io/stargz-containers/python:3.9-org ghcr.io/ktock/python:3.9-esgz
# nerdctl push ghcr.io/ktock/python:3.9-esgz
```

### On golang code:

You can use [`estargz` pkg](https://github.com/containerd/stargz-snapshotter/tree/main/estargz) of stargz snapshotter project. This provides [`estargz.Build()`](https://pkg.go.dev/github.com/containerd/stargz-snapshotter/estargz#Build) which converts a blob (gzip, zstd or plain tar) into the eStargz blob.

Stargz Snapshotter requires the following OCI annotations ([source](https://github.com/containerd/stargz-snapshotter/blob/v0.7.0/estargz/types.go#L70-L80)) in the OCI descriptor of the layer.

- `containerd.io/snapshot/stargz/toc.digest` : the digest of uncompressed TOC JSON of the layer.
- `io.containers.estargz.uncompressed-size` : uncompressed size of the layer.




> we can safely assume estargz is the format used by containerd?

Yes, we can.

lsm5:
@ktock if you would like to do the actual packaging it for fedora, let's talk via email (lsm5 AT redhat DOT com) or irc/matrix. My libera irc is `lsm5` and matrix id is @lsm5:lsm5.ems.host . Thanks a lot!

giuseppe:
@ktock opened a PR for containers/image: https://github.com/containers/image/pull/1351

With that in place, you can convert an image more easily as:

```
skopeo copy --dest-compress-format gzip:estargz docker://foo/bar docker://baz/bar
```

ktock:
@giuseppe stargz snapshotter  [v0.8.0 ](https://github.com/containerd/stargz-snapshotter/releases/tag/v0.8.0)comes with support for lazy pulling of zstd:chunked

giuseppe:
@ktock thanks for the update.  That is a great news :)

PR #1114, opened 2021-01-08
Update $HOME-dependant tests for c/storage, and c/storage itself

mtrmac:
This:
- Finally fixes pkg/docker/config tests on non-Linux systems like macOS, after a long time
- Modifies various tests not to assume that changing $HOME changes the value of `homedir.Get()`, mostly by splitting the tested code into a wrapper+internal helper, and testing only the internal helper with an explicit home directory parameter
- Afterwards, updates c/storage like #1113 to a recent version, where `pkg/homedir` only evaluates $HOME once.

i.e. this updates c/storage like #1113 without breaking tests.


vrothberg (commented):
LGTM

rhatdan:
LGTM

PR #1115, opened 2021-01-12
Set a default User-Agent if unset

jonjohnsonjr:
Currently, the User-Agent defaults to a golang default of either:

Go-http-client/1.1
Go-http-client/2.0

This makes it rather difficult to distinguish this client from any other
random golang program in registry logs.

Instead, set the default User-Agent to `"containers/$VERSION (github.com/containers/image)"`.

Context: https://github.com/containers/skopeo/issues/1109

Signed-off-by: Jon Johnson <jonjohnson@google.com>

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -23,6 +23,7 @@ import (
 	"github.com/containers/image/v5/pkg/sysregistriesv2"
 	"github.com/containers/image/v5/pkg/tlsclientconfig"
 	"github.com/containers/image/v5/types"
+	"github.com/containers/image/v5/version"
 	"github.com/containers/storage/pkg/homedir"
 	clientLib "github.com/docker/distribution/registry/client"
 	"github.com/docker/go-connections/tlsconfig"
@@ -65,6 +66,8 @@ var (
 		{path: "/etc/containers/certs.d", absolute: true},
 		{path: "/etc/docker/certs.d", absolute: true},
 	}
+
+	defaultUserAgent = "containers/" + version.Version + " (github.com/containers/image)"
 )
 
 // extensionSignature and extensionSignatureList come from github.com/openshift/origin/pkg/dockerregistry/server/signaturedispatcher.go:
@@ -92,8 +95,9 @@ type bearerToken struct {
 // dockerClient is configuration for dealing with a single Docker registry.
 type dockerClient struct {
 	// The following members are set by newDockerClient and do not change afterwards.
-	sys      *types.SystemContext
-	registry string
+	sys       *types.SystemContext
+	registry  string
+	userAgent string
 
 	// tlsClientConfig is setup by newDockerClient and will be used and updated
 	// by detectProperties(). Callers can edit tlsClientConfig.InsecureSkipVerify in the meantime.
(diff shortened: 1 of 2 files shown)


mtrmac (commented):
Thanks! This captures the most important part of https://github.com/containers/skopeo/issues/1109 .

LGTM, hoping for a second pair of eyes to review for possible risks we might have missed.

mtrmac:
> I didn't see an obvious place to add tests for this. I'm personally happy not to write any tests for a small change like this, but I'm also more than willing to if you can point me in the right direction 😄

Yeah. I wouldn’t mind a test for `newDockerClient`, even if it tests only the small aspect of setting the `userAgent` field based on input; the c/image/docker package has few unit tests and we have to start _somewhere_.

OTOH the cost/benefit ratio of writing tests for this really is pretty bad — if this will ever be broken, it will probably not be individually broken, but just moved or bypassed in a way that a unit test would not detect.  It would make more sense to incorporate a test into something a bit more end-to-end, at least like #254, but that has proven to be a nightmare to maintain and is probably not going to happen in this repo.

jonjohnsonjr:
I've added the simplest test I can imagine (that actually involves sending an http request).

rhatdan:
LGTM
@mtrmac @vrothberg PTAL

docker/docker_client.go
```diff
@@ -277,9 +279,15 @@ func newDockerClient(sys *types.SystemContext, registry, reference string) (*doc
 	}
 	tlsClientConfig.InsecureSkipVerify = skipVerify
 
+	userAgent := "containers-image/" + version.Version
```
vrothberg, line 282: Any chance we can do something like "github.com/containers/image"?  Not sure that's allowed but the source would be more explicit.

jonjohnsonjr: > Any chance we can do something like "github.com/containers/image"? Not sure that's allowed but the source would be more explicit.


I believe "something like", yes, but not exactly:

```
       User-Agent     = "User-Agent" ":" 1*( product | comment )

       product         = token ["/" product-version]

       product-version = token

       token          = 1*<any CHAR except CTLs or tspecials>

       tspecials      = "(" | ")" | "<" | ">" | "@"
                      | "," | ";" | ":" | "\" | <">
                      | "/" | "[" | "]" | "?" | "="
                      | "{" | "}" | SP | HT

       comment        = "(" *( ctext | comment ) ")"

       ctext          = <any TEXT excluding "(" and ")">

       TEXT           = <any OCTET except CTLs,
                        but including LWS>
```

From https://tools.ietf.org/html/rfc1945#section-2.2 and https://tools.ietf.org/html/rfc1945#section-3.7 and https://tools.ietf.org/html/rfc1945#section-10.15

`token` forbids `/`, but we could add it to the `comment`, which can include anything other than `(`, `)`, control characters, and newlines.

How about this?

```
User-Agent: containers/$VERSION (github.com/containers/image)
```

vrothberg: Thanks for looking into it. I really like it :)

@rhatdan @mtrmac PTAL

jonjohnsonjr:
Updated the PR description and pushed the suggestion of `containers/$VERSION (github.com/containers/image)`.

Let me know if you'd prefer I squash the commits so that the commit message makes more sense.

rhatdan:
LGTM

PR #1119, opened 2021-01-15
Clarify the canModifyBlob condition in copyBlobFromStream

mtrmac:
The `copyConfig` caller already sets `canModifyBlob` to `false`, set it explicitly inside the function just to make the relationship easier to follow.

Also fix some comments.

Motivated by https://github.com/containers/image/pull/1089#discussion_r558475506 .

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -1056,7 +1056,7 @@ type diffIDResult struct {
 	err    error
 }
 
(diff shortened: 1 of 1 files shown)


mtrmac:
Marked as draft because this is much less important than #1089, and #1089 will probably need to change the logic a bit.

rhatdan:
LGTM

nalind:
LGTM.

PR #1120, opened 2021-01-16
Cleanup description of shortname expansion

rhatdan:
When short name expansion is happening in Podman and Buildah there
are two many useless words being repeated on the screen. Also the
actual expansion is never shown. This PR shortens the text to only
the key descriptions and displays the actual long name.

podman pull ubi8
Resolved "ubi8" (/etc/containers/registries.conf.d/shortnames.conf)
...

### DIFF
--- a/pkg/shortnames/shortnames.go
+++ b/pkg/shortnames/shortnames.go
@@ -149,9 +149,9 @@ const (
 func (r *Resolved) Description() string {
 	switch r.rationale {
 	case rationaleAlias:
-		return fmt.Sprintf("Resolved short name %q to a recorded short-name alias (origin: %s)", r.userInput, r.originDescription)
+		return fmt.Sprintf("Resolved %q as an alias (%s)", r.userInput, r.originDescription)
 	case rationaleUSR:
-		return fmt.Sprintf("Completed short name %q with unqualified-search registries (origin: %s)", r.userInput, r.originDescription)
+		return fmt.Sprintf("Resolving %q using unqualified-search registries (%s)", r.userInput, r.originDescription)
 	case rationaleUserSelection, rationaleNone:
 		fallthrough
 	default:


rhatdan:
@vrothberg @mtrmac @lsm5 PTAL

pkg/shortnames/shortnames.go
```diff
@@ -149,9 +157,9 @@ const (
 func (r *Resolved) Description() string {
 	switch r.rationale {
 	case rationaleAlias:
-		return fmt.Sprintf("Resolved short name %q to a recorded short-name alias (origin: %s)", r.userInput, r.originDescription)
+		return fmt.Sprintf("Resolve %q->%q (%s)", r.userInput, r.candidates(), r.originDescription)
```
mtrmac, line 160: It doesn’t make sense to put the whole list of candidates, with possible commas, in quotes.

(Also `Description` is small enough, and the loop building the candidates list seems difficult enough to reuse, that it might be a bit cleaner to inline it.)

pkg/shortnames/shortnames.go
```diff
@@ -149,9 +157,9 @@ const (
...
 	case rationaleUSR:
-		return fmt.Sprintf("Completed short name %q with unqualified-search registries (origin: %s)", r.userInput, r.originDescription)
+		return fmt.Sprintf("Resolve %q with unqualified-search registries (%s)", r.userInput, r.originDescription)
```
mtrmac, line 162: `s/Resolve/Resolved/` in both cases?

pkg/shortnames/shortnames.go
mtrmac, line 160: The origin is currently usually a path to a config file, which can contain both `unqualified-search-registries` and `aliases`, so this probably keep saying that it found an alias.

vrothberg (commented):
The idea of these logs was not to show to what the short names resolved to but _how_.   Later logs during `podman pull` or `buildah pull` will show that information.  I think it's important to know whether an alias or an unqualified-search registry was used.

I am open to trying to reduce the number of words but I am rather objecting the idea to (redundantly) display the resolved references:
```
pull foo
foo -> [quay.io/repo/foo, registry.com/repo/foo]
Trying to pull quay.io/repo/foo
... failed
Trying to pull registry.com/repo/foo
```
... is too redundant IMHO

rhatdan:
We can discuss this at today's standup.  I tried another pass.

lsm5:
I don't have strong opinions on this as I haven't really dealt with shortnames before, so I'll defer to Miloslav and Valentin, but fwiw, I would prefer the older `Resolved ubi8 to a recorded short-name alias (origin: $conf_file_path)` as it appears clearer.

rhatdan:
@lsm5 The issue is that will get old. After you see it a few hundred times,  My goal is to get this as close to 80 chars as possible, so I don't see terminal wrap all of the time.

lsm5:
> @lsm5 The issue is that will get old. After you see it a few hundred times, My goal is to get this as close to 80 chars as possible, so I don't see terminal wrap all of the time.

ack, yes, makes sense, LGTM from my side, so I'll defer to the others.

mtrmac (commented):
Non-blocking: I’d prefer both to include the precise option name (`aliases`, `unqualified-search-registries`).

That would make the short-name line even shorter (perhaps “Resolved %q as an alias (%s)”). OTOH the USR line would get a bit longer.

mtrmac (commented):
Works for me. @vrothberg ?

pkg/shortnames/shortnames.go
```diff
@@ -149,9 +149,9 @@ const (
 func (r *Resolved) Description() string {
 	switch r.rationale {
 	case rationaleAlias:
-		return fmt.Sprintf("Resolved short name %q to a recorded short-name alias (origin: %s)", r.userInput, r.originDescription)
+		return fmt.Sprintf("Resolved %q as an alias (%s)", r.userInput, r.originDescription)
 	case rationaleUSR:
-		return fmt.Sprintf("Completed short name %q with unqualified-search registries (origin: %s)", r.userInput, r.originDescription)
+		return fmt.Sprintf("Resolving %q using unqualified-search-registries (%s)", r.userInput, r.originDescription)
```
vrothberg, line 154:
```suggestion
		return fmt.Sprintf("Resolving %q using unqualified-search registries (%s)", r.userInput, r.originDescription)
```
The 2nd dash isn't needed.

vrothberg (commented):
Other than the nit, LGTM.

@rhatdan, do you want that in 3.0?

pkg/shortnames/shortnames.go
mtrmac: The option in the config file is literally named `unqualified-search-registries`.

vrothberg: For technical reasons though.  The hyphen in English is needed to relate "search".

TomSweeneyRedHat:
@rhatdan needs a rebase.  I'd drop the second dash as suggested by @vrothberg, that matches what we do in the man pages.

rhatdan:
too late for 3.0
I am willing to wait until 3.1

mtrmac (approved):
LGTM, then.

PR #1121, opened 2021-01-18
Update `golangci-lint`

mtrmac:
… because it fails with recent Go versions.

(But first fix some problems the new version can find.)


vrothberg (commented):
LGTM
Good catch!

Issue #776, opened 2019-12-12
Limit maximum concurrent layer pull


yangm97:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description**

Although concurrent pulling usually speeds things up over broadband connections it will slow down or even fail a pull over a less than ideal connection.

The docker daemon has a `--max-concurrent-downloads` flag which allows the user to tune the image pull parallelism. Since we're daemonless that would need to be set on the config file or as a `pull/run/create` flag (or even both).

I looked into the code hoping to implement this myself and send a PR but I couldn't find where image pulling is done 😬.

rhatdan:
I see this as an issue with CRI-O, perhaps but not sure this is something we should attempt to fix.  Preventing a user from running 100 podman commands at the same time?  Pulling images from a container registry?  Seems like a lot of complexity would need to be added to the code.  To me this is similar to someone firing off 100 curl/wget commands against a we server.  It is up to the web server/registry to protect itself.

vrothberg:
I interpreted the request to limit the number of concurrent layer pulls in a single pull operation but not prevent other podmans or skopeos from pulling. At the moment, when an images has more than 1 layer we are pulling at most 6 layers in parallel. This can indeed cause problems in bad environments.

So it would not prevent other processes from pulling but just define how many go-routines can pull in parallel in one `c/image/copy` process.

rhatdan:
Ok if it is per/process then I am fine with this.  I just wanted to avoid anything more complicated.
Not sure how the user/admin would specify this, but might be a good candidate for containers.conf.

yangm97:
Yup, this issue is exactly as @vrothberg described it.

mtrmac:
I don’t think `containers.conf` is quite the perfect place, because that would set a global limit regardless of context/operation: a local user running a single `podman pull`, and CRI-O starting hundreds of pods, would run on the same system but would have different latency vs. throughput vs. available bandwidth decisions.

It would make a bit more sense to me to expose this in `c/image/copy.Options`, and let individual callers (that need to care for some reason) provide options, whether on the CLI or in caller-specific configuration files. (OTOH, this does seem to be somewhat niche use case, and doing this in every single caller may not be worth the effort; not sure.)

zgoda:
This is becoming painful now that I am mostly working from home and my connection is far from being stable. With 6 concurrent layer pulls it easily gets saturated resulting in pull failures.

[...]

PR #1124, opened 2021-01-19
Allow callers to set the MaxParallelDownloads field

rhatdan:
Users should be able to set the number of ParallelDownloads so that
they can customize how container/tools perform based on the
characteristics of their network.

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -48,7 +48,7 @@ var (
 
 	// maxParallelDownloads is used to limit the maxmimum number of parallel
 	// downloads.  Let's follow Firefox by limiting it to 6.
-	maxParallelDownloads = 6
+	maxParallelDownloads = uint(6)
 )
 
 // compressionBufferSize is the buffer size used to compress a blob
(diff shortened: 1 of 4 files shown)


copy/copy.go
```diff
@@ -191,6 +191,8 @@ type Options struct {
 	// OciDecryptConfig contains the config that can be used to decrypt an image if it is
 	// encrypted if non-nil. If nil, it does not attempt to decrypt an image.
 	OciDecryptConfig *encconfig.DecryptConfig
+	// MaxParallelDownloads indicates the maximum layers to pull at the same time
+	MaxParallelDownloads uint
```
mtrmac, line 195: This should document the “built-in default” case.

mtrmac (commented):
Design LGTM, but this doesn‘t compile.

copy/copy.go
mtrmac: To be more precise: It should say something like “A reasonable default is used if this is left as 0” (suggesting that that’s what most callers should do; it should _not_ commit to 6 as an API guarantee.

copy/copy.go
```diff
@@ -256,13 +259,14 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
...
-		progress:         options.Progress,
-		copyInParallel:   copyInParallel,
+		dest:                 dest,
+		rawSource:            rawSource,
+		reportWriter:         reportWriter,
+		progressOutput:       progressOutput,
+		maxParallelDownloads: options.MaxParallelDownloads,
+		progressInterval:     options.ProgressInterval,
+		progress:             options.Progress,
+		copyInParallel:       copyInParallel,
```
mtrmac, line 269: Please keep the field order in constructor the same as in the type definition. (I don’t care which which of the two, is updated, or perhaps both.)

rhatdan: Fixed

copy/copy.go
rhatdan: Fixed

vrothberg (commented):
LGTM

rhatdan:
First step to fix https://github.com/containers/image/issues/776

rhatdan:
@mtrmac I think this is ready to merge.

mtrmac (approved):
LGTM. Thanks!

DjangoCalendar:
@rhatdan @mtrmac has this been finally exposed to `skopeo copy` somehow ?

If not directly to cli but over some configuration that I need to pass would you be able to share how ?

thanks in advance,

mtrmac:
It’s not in yet Skopeo (but it seems to be configurable for Podman/Buildah).

Care to prepare a PR?

DjangoCalendar:
I May try but not sure where to start from. Any guidance?

PR #1127, opened 2021-01-26
Fix 'systemDefaultPolicyPath' linker flag comment

gjenkins8:
Tweak the path in the comment, to avoid the next reader hitting the same copy&paste error (but in general, it seems annoying that this comment will go out of date next module version)

### DIFF
--- a/docker/lookaside.go
+++ b/docker/lookaside.go
@@ -21,7 +21,7 @@ import (
 
 // systemRegistriesDirPath is the path to registries.d, used for locating lookaside Docker signature storage.
 // You can override this at build time with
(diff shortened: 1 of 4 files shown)


vrothberg (commented):
LGTM, thanks!

@rhatdan @mtrmac PTAL

mtrmac (commented):
Thanks.

There are 4 other comments with a similar problem. Could you update them as well, please?

> (but in general, it seems annoying that this comment will go out of date next module version)

Yes — and not just the comment, any previous configuration could just be silently ignored. Maybe we will be able to avoid that (by having v6 import v5, and indirectly use the /v5/ variables) — but primarily we should strive to avoid having to move to v6.

mtrmac (approved):
LGTM. Thanks!

PR #1132, opened 2021-01-30
Copy cleanups

mtrmac:
Primarily fixes writing of signatures if a manifest of a multi-arch image is updated.

Also cleans up the documentation of the blob copy pipeline to be a bit more consistent.


nalind:
LGTM

PR #1140, opened 2021-02-04
Cirrus: Migrate from Travis CI

cevich:


cevich:
@containers/image-maintainers Q.Q.: Does the main test-command `make cross tools test test-skopeo` ***have*** to run inside a privileged container (`.travis.yml` currently does this)?

There's no option for running simple cirrus-ci containers in privileged-mode (due to the security implications).  Assuming privileged mode is needed, the scope of this migration will need to be expanded to bring in a brand-new GCP setup for this repo.  I'm willing to do this, but would prefer to avoid it if possible.

rhatdan:
Not really sure why this would need to run privileged. Could you try to run without --privileged and see what blows up.

cevich:
> Could you try to run without --privileged and see what blows up.

Ugh, of course!  I'm a fool, I should have thought to do that...maybe something's wrong with my Coffee.

cevich:
Answer: [There are test failures](https://travis-ci.org/github/containers/image/jobs/757602064#L1700-L1714), but I don't know enough to tell if they're caused by a missing --privileged or not.

.travis.yml
```diff
@@ -17,7 +17,7 @@ env:
   - BUILDTAGS='btrfs_noversion libdm_no_deferred_remove containers_image_openpgp'
 
 script: >
-  sudo docker run --privileged -ti --rm --user $(id -u):$(id -g)
+  sudo docker run -ti --rm --user $(id -u):$(id -g)
```
vrothberg, line 20: I think we need the --privileged to get the tests green.

mtrmac:
> @containers/image-maintainers Q.Q.: Does the main test-command `make cross tools test test-skopeo` _**have**_ to run inside a privileged container (`.travis.yml` currently does this)?

I’d expect `cross tools test` to run unprivileged without much hassle. `test-skopeo` creates its own container, and runs several registry servers and (a very old version of) OpenShift inside it; I’d expect that would have to run in _some_ container. (Also, using the Cirrus CI create-container-from-Docker file feature would be great for this, because most of the test runtime is building this OpenShift, so keeping an image around that is very rarely rebuilt would speed up tests a lot.)

It is may well be possible to run the Skopeo tests in a Cirrus-managed container directly instead of a `Makefile`-managed `make check` creating a nested container inside a CI-managed container, probably avoiding the need for `--privileged`. All of that would be easier to set up / manage / test directly from the Skopeo repo, which is why I recommend migrating Skopeo to Cirrus first, and only moving c/image afterwards.

mtrmac:
> Answer: [There are test failures](https://travis-ci.org/github/containers/image/jobs/757602064#L1700-L1714), but I don't know enough to tell if they're caused by a missing --privileged or not.

If the kernel keyring is not namespaced, and unaccessible from a container (which would not be surprising), `--privileged` might help… well, Cirrus shouldn’t allow two tenants to share a keyring.

Note that the keyring use has been disabled, so we don’t actually urgently need these tests. So, modifying the tests to `t.Skip` if there is no permission to access the keyring (and documenting why we do it and why it is fine) could well be worth avoiding the `--privileged` complexity.

cevich:
Thanks for all the insights @mtrmac 

> I recommend migrating Skopeo to Cirrus first, and only moving c/image afterwards.

Oh, that's interesting/unexpected.   I'm happy to focus on c/skopeo first, just want to know more precisely what/why (so I don't miss it).  So there are some parts of the c/image testing that "vendors" (used loosely) tests or setup from c/skopeo?

> Cirrus CI create-container-from-Docker file feature

Yes, this is really handy but does have limitations.  I *think* I'm reading between the lines that you'd like to use the managed container image from one repo., in testing for another repo?   If so, I don't think that's possible.  The images are named with hashes that AFAIK, aren't consistent or discoverable from another repo's Cirrus-CI operations.  Otherwise, and if we can get around the `--privileged` need, this would be a nice feature to utilize.

> so we don’t actually urgently need these tests.

Okay, so disabling these tests for CI could be an accepted option to avoid needing to implement GCP support.  Certainly setting it up at some future date is always an option, and would help spread out the migration efforts more over time.

mtrmac:
> > I recommend migrating Skopeo to Cirrus first, and only moving c/image afterwards.
> 
> Oh, that's interesting/unexpected. I'm happy to focus on c/skopeo first, just want to know more precisely what/why (so I don't miss it). So there are some parts of the c/image testing that "vendors" (used loosely) tests or setup from c/skopeo?

Yes, c/image mostly relies on Skopeo integration tests via `make test-skopeo` to run tests for the `docker://` code and the copy implementation. The `Makefile` rule should be fairy straightforward to read.


> > Cirrus CI create-container-from-Docker file feature
> 
> Yes, this is really handy but does have limitations. I _think_ I'm reading between the lines that you'd like to use the managed container image from one repo., in testing for another repo? If so, I don't think that's possible. The images are named with hashes that AFAIK, aren't consistent or discoverable from another repo's Cirrus-CI operations. Otherwise, and if we can get around the `--privileged` need, this would be a nice feature to utilize.

I was primarily thinking of using it directly for Skopeo; you’re right that it could equally benefit c/image, where the indirection complicates things.

---

> > so we don’t actually urgently need these tests.
> 
> Okay, so disabling these tests for CI could be an accepted option to avoid needing to implement GCP support.

(Please let the tests run if the test user is allowed to use keyrings; don’t just remove them.)

.travis.yml
rhatdan: The failures are just in creating keyrings.  The rest of the tests are passing, as far as I can see.  We could potentially check if the tests are inside of a container and not run.

rhatdan:
@cevich Why can't we use the same VMs that we are using for podman to run these tests?

cevich:
> Why can't we use the same VMs that we are using for podman to run these tests?

We can (and will) use the same VM images, since those can be shared read-only.  But I'd like to keep separate GCP projects per github repo., out of an abundance of caution.  Primarily, this makes monitoring easier in case a GCP project's credentials leak, the possible damage would be limited to only that project, instead of several.  Think of it as analogous to running multiple rootless-podman containers under different user accounts, instead of running everything under root :wink:

As per usual...the trade-off between ease-of-use and security, is ever present, even in the cloud :disappointed:

.travis.yml
cevich: Yep, should be mostly trivial to skip based the `$container` env. var. being set non-empty, no?

cevich:
@vrothberg reminder: you were going to investigate adding a skip to the tests needing the kernel keyring, if they're running inside a container (i.e. `$container` is non-empty).

I see where the Makefile does pull down, build, and test skopeo but I don't see anything suggesting that Cirrus-CI must be working there first.  I'll proceed forward here, with one eye closed for safety-sake :wink:

vrothberg:
@cevich, is it possible to run the container as a non-root user inside the container? The tests are built with this assumption.

vrothberg:
SOOOOO nice to see more of Cirrus here, @cevich. Thanks a lot!

cevich:
> is it possible to run the container as a non-root user inside the container? The tests are built with this assumption.

Hmmm, this is a more complex than it appears since "being a user" normally requires/implies additional context.  I searched through the Cirrus-CI docs, but don't see any mention of running a container with a user account.  Though we do have "root" access, so conceivably can run things via `sudo` or `su`.

Note: This may be another argument to give up and go for a full-blown GCE VM setup.  There we can do the `ssh` trick to setup as root, but then execute as a user with full shell-login context.

cevich:
> SOOOOO nice to see more of Cirrus here, @cevich. Thanks a lot!

You're welcome...though writing tasks is one thing; making them pass is another :confused:

cevich:
@mtrmac I think I understand more now re: the `test-skopeo` conundrum.   The intent is to confirm that a c/image change does not impact c/skopeo, ya?

I spent quite a while pondering how to do this w/o anyone completely loosing their mind in complexity.  I think the best way to stay within containers-only Cirrus testing, is to setup `quay.io` to build the c/skopeo `Dockerfile` on every change (there), tagged with the branch name.  Using that image here (with configurable branch-name), Cirrus would clone this repo. into the container before building c/image, rebuild c/skopeo, then run tests.

Sounds like a lot of work, but it's actually not that much compared to implementing a full GCP cloud setup.

vrothberg:
@cevich, Cirrus should be able to set the user when setting the container user via the Dockerfile.

```Dockerfile
FROM $IMAGE
USER newuser
```

cevich:
> user via the Dockerfile

I'm happy to try, but I won't be surprised if it's not honored at runtime.  We may still end up doing some sudo/su shenanigans :confounded:

vrothberg:
https://github.com/cirruslabs/cirrus-ci-docs/issues/178 gives me hope *fingers crossed

cevich:
Excellent!  Working on some changes and will give it a try...

cevich:
Well :hankey: 

```
Using built-in Git...
Failed to init repository: mkdir /var/tmp/go/src/github.com/containers/image: permission denied!
```

I wasn't expecting it to break like that.  Hrmmmm.

cevich:
So it only took 9 tries, but I got the basics all green in about 1-minute :smile:

Now to get the skopeo task to run is going to be the real trick.

cevich:
**Note to self**: Need to add a 'success' (status accumulator) task, and update the branch-protection rules and make that required instead of `continuous-integration/travis-ci`

vrothberg:
Nice work, @cevich !

cevich:
@mtrmac @vrothberg I setup quay.io to build and tag a skopeo CI container for `master` and each of the `release-*` branches on repo. change.  Then I pointed the new "Skopeo" tests (here) at that image.  I also setup a matrix for the testing tasks to use two variations of BUILDTAGS (as was done under Travis).  We'll know shortly if/what new problems there are...

cevich:
...oof, the skopeo Makefile is trying and failing to build the very container it's already running inside (same as what quay.io builds now).  This must be the part where/why @mtrmac warned about doing the c/skopeo conversion first :grin: 

According to the Makefile comment, the `test-skopeo` target is entirely intended to be only run under CI.  To enforce that, I moved the target's instructions into `.cirrus.yml`.  This will also make it more readable at runtime, since Cirrus-CI prints out the script commands as it runs them.

Looking deeper, it appears there are only a few places where a skopeo Makefile `-local` target isn't available (integration and system tests).  So I'll mark those as "TODO" items for now, but hopefully the rest of the skopeo targets work...

cevich:
...lovely, it appears like github or cirrus-ci broke (no new run despite pushing changes) :confused: ...

cevich:
...there, got 'em "passing" now.

Ref. https://cirrus-ci.com/task/5483688348614656

@mtrmac @vrothberg PTAL, is it okay to leave these TODO's in the Skopeo tests for now.  I'll circle back here after mucking around with the skopeo `Makefile` and `hack/make.sh`.  I glanced at them enough to spot the problem areas, but have no idea on the real extent of needed changes until I get my hands into them.

cevich:
**Note to self**: Need to add a 'success' (status accumulator) task, and update the branch-protection rules and make that required instead of `continuous-integration/travis-ci`

cevich:
@mtrmac @vrothberg ping (are the TODO's okay?)

vrothberg:
> @mtrmac @vrothberg ping (are the TODO's okay?)

Thanks for the ping and apologies for the silence. I don't think the TODO's are okay since those are the main tests. So we shouldn't merge it now.

If I understand the comments above correctly, some changes to Skopeo Makefile should let the tests run again, would they?

Great work, btw.  I was looking forward to seeing Cirrus here :)

mtrmac:
(Drive-by comment without reading the full history, I’m afraid): Yes, Skopeo’s `test-integration` and `test-system` tests are the ~only reason we want Skopeo’s tests here (not the unit tests of Skopeo’s handling of CLI options), and almost the only tests that the `docker://` registry client works at all.

So those two need to continue to be called in c/image’s CI. (Or, well, the test suite could be migrated into c/image somehow, and Skopeo could become a trivial CLI wrapper with tests only for that trivial CLI wrapper (except for separate tests for `skopeo sync` I guess); that would make both packages easier to maintain, but it’s absolutely out of scope of this effort.)

cevich:
Thanks for the feedback guys, I wasn't aware the skopeo tests were so important here.

Unfortunately there's no simple/practical way to have all the automation inside skopeo.  This is because it's complex to have an event on one repo., provide feedback in a meaningful context (i.e. a PR on the first repo.).

I think the current setup is about the best it can get, so I'll shift focus onto the c/skopeo conversion.  I'm hopeful it's just a Makefile/script change, but won't be surprised if not.

cevich:
Update: [Got c/skopeo integration tests working, using code from a skopeo PR.](https://cirrus-ci.com/task/6547829914075136?command=main#L224)

cevich:
@mtrmac Cirrus-CI containers are run with docker.  I got the c/skopeo *integration* tests working, but it seems the *system-tests* can't work in a podman-in-docker environment (w/o `--privileged`).  Am I wrong?  Any ideas for alternatives?

cevich:
@vrothberg @mtrmac  I confirmed with Dan: It's a major PITA to pull off podman-in-docker, so some fairly extensive re-architecture of the skopeo system-test would be needed and that seems outside the scope of this PR.

Depending on https://github.com/containers/skopeo/pull/1211, I think we should move forward with the system-tests (commented out as a *TODO* (here).  It's an imperfect situation but hopefully temporary.  It would be good to provide some benefit to developers from a partial Cirrus-CI migration, no?

vrothberg:
> @vrothberg @mtrmac I confirmed with Dan: It's a major PITA to pull off podman-in-docker, so some fairly extensive re-architecture of the skopeo system-test would be needed and that seems outside the scope of this PR.
> 
> Depending on [containers/skopeo#1211](https://github.com/containers/skopeo/pull/1211), I think we should move forward with the system-tests (commented out as a _TODO_ (here). It's an imperfect situation but hopefully temporary. It would be good to provide some benefit to developers from a partial Cirrus-CI migration, no?

Thanks, @cevich!  We need _all_ Skopeo tests to run.  Otherwise, we'd be running blind for an unknown amount of time.

cevich:
>  running blind for an unknown amount of time.

Well...This problem only affects CI, developers are still able to run the tests manually...So perhaps partially blind in one eye :wink:

I opened #1216 about this.  I'm leaning toward setting up VM's in a shared GCP project for both c/image and c/skopeo.  It's a bit of kicking the can down the road, but it will at least allow the migration to happen sooner.  At the cost of greatly increased infrastructure complexity.

vrothberg:
Much appreciated, @cevich! Thank you a ton!

cevich:
> Much appreciated, @cevich! Thank you a ton!

You're welcome...though don't celebrate too loudly yet, nothing will actually work until the backend bureaucratic stuff gets sorted out.  That could take a day or two...and it's possible they'll say "no"...but I'm optimistic.

cevich:
@vrothberg @mtrmac I need some help here, [some skopeo tests are failing with a nonsense error.](https://cirrus-ci.com/task/6061681961861120?command=main#L2395).  It doesn't seem to matter if I add `root:10000:65535` to `/etc/sub{u,g}id`, the tests still try to run in rootless mode, despite being root.

Previously under Travis, the tests ran docker as root in --privileged mode, which then runs podman in rootless mode.  I was hoping to avoid all this extra complexity.  Do you think it would work if I simply ran the skopeo `test-integration` and `test-system`  targets as a regular user instead of root?

cevich:
Oof, nevermind, I'm just going to try it...

.cirrus.yml
```diff
@@ -1,13 +1,212 @@
...
-  validate_script:
-    - apt-get update -qq
-    - apt-get install -qq -y libgpgme-dev libdevmapper-dev btrfs-tools libbtrfs-dev
-    - make tools
-    - make .gitvalidation
-    - make validate
+    # The git-validation tool doesn't work well on branch or tag push,
+    # under Cirrus-CI, due to challenges obtaining the starting commit ID.
+    # Only do validation for PRs.
+    only_if: CIRRUS_PR != ''
```
edigaryev, line 56:
```suggestion
    only_if: $CIRRUS_PR != ''
```

Just passing by: you've probably meant this 😅

cevich: Woops!  Thanks!

cevich:
@vrothberg @mtrmac PTAL.  I've setup the skopeo tests to run as root and rootless (using an actual login-shell).  However, I'm **still** encountering substantially the same [oddball errors as I did last week](https://github.com/containers/image/pull/1140#issuecomment-790885345) :disappointed:  [The problems appear to come from](https://cirrus-ci.com/task/6062415126200320?command=main#L482) `copy_test.go` but don't look (to me) like https://github.com/containers/skopeo/issues/1093 :cry: 

I believe this is a real (Skopeo) problem that's been hidden by years of burying the tests inside an old-ass Ubuntu 18  environment.  When the tests are run on a F33 VM, it readily reproduces these `copy_test.go` errors.  

```bash
# as root: echo "root:10000:65535" | tee -a /etc/subuid >> /etc/subgid
# as root: dnf erase -y skopeo
export GOPATH=/var/tmp/go
export BUILDTAGS='btrfs_noversion libdm_no_deferred_remove'
export IMAGE_PATH=$GOPATH/src/github.com/containers/image
export SKOPEO_PATH=$GOPATH/src/github.com/containers/skopeo
export "PATH=$SKOPEO_PATH/bin:$PATH:$GOPATH/bin"
mkdir -p $IMAGE_PATH
mkdir -p $SKOPEO_PATH
git clone https://github.com/containers/image $IMAGE_PATH
git clone https://github.com/containers/skopeo $SKOPEO_PATH
cd $IMAGE_PATH
make tools
cd $SKOPEO_PATH
GO111MODULE="on" go mod edit -replace github.com/containers/image/v5=$IMAGE_PATH
make vendor
make bin/skopeo
make BUILDTAGS="${BUILDTAGS}" test-integration
```

Always I get `copy_test` failures with a message similar to:

```
...cut...
/usr/bin/podman run --rm -i -e TESTFLAGS=""  -t "skopeo-dev:master" bash -c 'rm -f /dev/random; ln -sf /dev/urandom /dev/random; SKOPEO_CONTAINER_TESTS=1 BUILDTAGS="btrfs_noversion libdm_no_deferred_remove" hack/make.sh test-integration'
rm: cannot remove '/dev/random': Device or resource busy
ln: failed to create symbolic link '/dev/random': Device or resource busy
...cut...
... value *exec.ExitError = &exec.ExitError{ProcessState:(*os.ProcessState)(0xc00051f540), Stderr:[]uint8(nil)} ("exit status 1")
... time="2021-03-08T21:22:57Z" level=warning msg="error reading allowed ID mappings: error reading subuid mappings for user \"root\" and subgid mappings for group \"root\": No subuid ranges found for user \"root\" in /etc/subuid"
time="2021-03-08T21:22:57Z" level=warning msg="Found no UID ranges set aside for user \"root\" in /etc/subuid."
time="2021-03-08T21:22:57Z" level=warning msg="Found no GID ranges set aside for user \"root\" in /etc/subgid."
...cut...
time="2021-03-08T21:23:00Z" level=error msg="Error while applying layer: ApplyLayer exit status 1 stdout:  stderr: potentially insufficient UIDs or GIDs available in user namespace (requested 0:12 for /var/spool/mail): Check /etc/subuid and /etc/subgid: lchown /var/spool/mail: invalid argument"
```

The error is the same no matter if I'm root or a regular user on the VM.  The test always complains about no subuid/gid for `root`. 

Can you please take a look at the test logs, test code, and advise what could possibly be going wrong?  It may be a simple fix, I just don't know the code/tests well enough to say.

cevich:
Update: I spent an hour looking through the test code, and am now more confused than when ignorant.  It seems the tests mostly execute substantially similar `skopeo` commands.  So I wonder why is it only these several tests that throw subuid/gid errors for root, and none of the others?

cevich:
@containers/image-maintainers PTAL, I think this is about as good as this will get for now.  **Note:**The large number of skipped skopeo Integration tests will be resolved once https://github.com/containers/skopeo/pull/1220 merges.

rhatdan:
LGTM
I say  we get this over to Cirrus and then continue improving it there.

vrothberg:
Let's first get https://github.com/containers/skopeo/pull/1220 in.  I think it's too dangerous to not have working tests, even for a short period of time.

cevich:
> even for a short period of time.

Yep, and since this pulls from master, I can just re-run the tests here once that merges.

cevich:
> @cevich can you rebase?

Yep.  Ref: [Tests are passing now w/o skipping.](https://cirrus-ci.com/task/5088711613349888?command=main#L201)

cevich:
Still passing., though the travis requirement will never succeed (this PR removes the config).  I'm going to preemptively change the settings to no-longer require travis.  This can be easily be un-done if it's a mistake.

cevich:
Okay, all-green now :grin:

vrothberg (commented):
LGTM
Great work, @cevich! Thank you

cevich:
Running tests outside a snowflake-container image ref: #1222

Makefile
```diff
@@ -83,40 +83,19 @@ clean:
...
-		rm -rf $${skopeo_path}
-
 fmt:
-	@gofmt -l -s -w $(SOURCE_DIRS)
+	@go fmt -l -s -w $(SOURCE_DIRS)
 
 validate: lint
 	@GO111MODULE="on" go vet ./...
-	@test -z "$$(gofmt -s -l . | grep -ve '^vendor' | tee /dev/stderr)"
+	@test -z "$$(go fmt -s -l . | grep -ve '^vendor' | tee /dev/stderr)"
```
mtrmac, line 91: Is there more context for these changes? AFAICT `go fmt` doesn’t have (and never had?) the `-s -l -w` options, this seems to break local `make`.

cevich: ugh, darn, I wish there was (in hind-sight).  It's stretching my memory quite a bit but I do remember having problems with running 'gofmt'.  Since those options are invalid, I bet the command is (incorrectly) passing because 'pipefail' defaults to disabled in bash.  In any case, I'll take a look into this and open up a new PR to fix it.  Thanks for tracking it down.

mtrmac: For the record, this is being addressed in #1255 .

PR #1142, opened 2021-02-09
short-name-aliases.conf: use cache folders instead of $HOME

vrothberg:
As shown in github.com/containers/podman/issues/8559, writing to the
$HOME directory or root may be undesired.  Using /var/cache for root and
$HOME./cache for ordinary users, however, is common practice.

### DIFF
--- a/pkg/sysregistriesv2/shortnames.go
+++ b/pkg/sysregistriesv2/shortnames.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/BurntSushi/toml"
 	"github.com/containers/image/v5/docker/reference"
+	"github.com/containers/image/v5/internal/rootless"
 	"github.com/containers/image/v5/types"
 	"github.com/containers/storage/pkg/homedir"
 	"github.com/containers/storage/pkg/lockfile"
@@ -27,12 +28,24 @@ func shortNameAliasesConfPath(ctx *types.SystemContext) (string, error) {
 		return ctx.UserShortNameAliasConfPath, nil
 	}
 
-	configHome, err := homedir.GetConfigHome()
-	if err != nil {
-		return "", err
+	if rootless.GetRootlessEUID() == 0 {
+		// Root user or in a non-conforming user NS
+		return filepath.Join("/var/cache", userShortNamesFile), nil
 	}
 
-	return filepath.Join(configHome, userShortNamesFile), nil
+	// Rootless user
+	var cacheRoot string
+	if xdgCache := os.Getenv("XDG_CACHE_HOME"); xdgCache != "" {
+		cacheRoot = xdgCache
+	} else {
+		configHome, err := homedir.GetConfigHome()
+		if err != nil {
+			return "", err
+		}
+		cacheRoot = filepath.Join(configHome, ".cache")
+	}
+
+	return filepath.Join(cacheRoot, userShortNamesFile), nil
 }
 
 // shortNameAliasConf is a subset of the `V2RegistriesConf` format.  It's used in the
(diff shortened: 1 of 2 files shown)


vrothberg:
@rhatdan @mtrmac @giuseppe PTAL

That is a breaking change but all discussed solution in https://github.com/containers/podman/issues/8559 are as well.  I was playing with several solutions yesterday and chatted with @rhatdan about them.  I know believe that using cache is the right thing to do for this kind of data.

giuseppe (commented):
Should rootless honor XDG_CACHE_HOME when it is set?

vrothberg:
> Should rootless honor XDG_CACHE_HOME when it is set?

Yes, it does that implicitly as it's using `storage/pkg/homedir.GetConfigHome()`

giuseppe:
I think that uses XDG_CONFIG_HOME. XDG_CACHE_HOME instead is the equivalent of /var/cache

vrothberg:
> I think that uses XDG_CONFIG_HOME. XDG_CACHE_HOME instead is the equivalent of /var/cache

Cheers! I honestly read CONFIG instead of CACHE.

rhatdan:
This LGTM, but we really need to get this in before we release podman 3.0.

vrothberg:
> This LGTM, but we really need to get this in before we release podman 3.0.

Agreed.

vrothberg:
@giuseppe, PTanotherL

giuseppe (approved):
LGTM

mtrmac:
- Was https://github.com/containers/podman/issues/8559#issuecomment-741836106 done? In that particular case short names should not even be used AFAICT.
- How does it make sense to use `/var/cache` for user configuration that should not be lost? (We even talked about commands like `podman alias foo=bar`). If anything, I’d expect `/var/lib`, (and maybe `/var/lib/…roots-private-short-name-aliases.conf` to make it clear it is not system-wide data).
- (I kind of accept that this has been a breaking change and moving out of `/root` is desirable even if not elegant… but it _is_ 3.0. after all, so forcing at least the automated systems-started services to use fully-qualified names seems rather attractive to me. But then I’ve always hated short names in all their forms :). )

pkg/sysregistriesv2/shortnames.go
```diff
@@ -27,12 +28,24 @@ func shortNameAliasesConfPath(ctx *types.SystemContext) (string, error) {
...
-	if err != nil {
-		return "", err
+	if rootless.GetRootlessEUID() == 0 {
+		// Root user or in a non-conforming user NS
+		return filepath.Join("/var/cache", userShortNamesFile), nil
+	}
+
+	// Rootless user
+	var cacheRoot string
+	if xdgCache := os.Getenv("XDG_CACHE_HOME"); xdgCache != "" {
```
mtrmac, line 38: Is there a reason to move the non-root configuration as well?

mtrmac: Consistency I guess.

vrothberg: Yes, that's for consistency.

vrothberg:
>     * Was [containers/podman#8559 (comment)](https://github.com/containers/podman/issues/8559#issuecomment-741836106) done? In that particular case short names should not even be used AFAICT.

No. I consider the image reference in systemd units to be a (very) different issue.  We may have a hard time getting that done since we're literally copying `os.Args` around to generate these files.

>     * How does it make sense to use `/var/cache` for user configuration that should not be lost? (We even talked about commands like `podman alias foo=bar`). If anything, I’d expect `/var/lib`, (and maybe `/var/lib/…roots-private-short-name-aliases.conf` to make it clear it is not system-wide data).

`/var/cache` must persists over reboots _but_ users are allowed to remove it and even edit. I think that reflects the intentions.

@rhatdan and I chatted about using `/var/lib` (using the storage GraphRoot in particular) but that turned out to be messy with the different tools allowing to change the GraphRoot in various ways. c/image has so far been able to ignore GraphRoot and we didn't want to introduce it.

>     * (I kind of accept that this has been a breaking change and moving out of `/root` is desirable even if not elegant… but it _is_ 3.0. after all, so forcing at least the automated systems-started services to use fully-qualified names seems rather attractive to me. But then I’ve always hated short names in all their forms :). )

We stand united in the hate for short names :) Yet, I stand guilty of being a lazy ass, so I use them all the time.

PR #1148, opened 2021-02-15
shortnames: resolution error: point to manpage if config is absent

vrothberg:
When a short name could not be resolved via `shortnames.Resolve(...)`
the error message attempts to help the users solve the issue by pointing
to short-name aliases and the registries.conf.

Until now, the error implied the existence of a registries.conf which
may not always be the case.  Hence, check whether a registries.conf
was used during short-name resolution and point to the manpage if none
was present.

@rhatdan @giuseppe  @lsm5 @umohnani8 PTAL


vrothberg:
@saschagrunert PTAL

saschagrunert (approved):
LGTM

giuseppe (approved):
/lgtm

PR #1155, opened 2021-02-18
shortnames: correctly determine cache directory for rootless

go-d:
PR #1142 placed .cache as a subdirectory below configHome, but
$HOME/.cache is what it ought te be.

Signed-off-by: G.I. Doe <gogit@pm.me>

PR #1157, opened 2021-02-19
v5.10: place shortnames in `~/.cache` not `~/.config/.cache`

vrothberg:
In case XDG_CACHE_HOME isn't set for a rootless user, fix the fallback
location that we inteded to be `~/.cache` but while refactoring placed
it in `~/.config/.cache` instead.


vrothberg:
@rhatdan @giuseppe, we need this for 8.4.  https://github.com/containers/image/pull/1155 is the counter part for the main branch.

rhatdan:
LGTM

vrothberg:
As @rhatdan opened https://github.com/containers/storage/pull/826, we may very well wait for that instead.

vrothberg:
Updated and vendored in the new c/storage

vrothberg:
Pinned the skopeo branch in the Makefile and also bumped the version to speed things up a bit.

giuseppe (approved):
/lgtm

PR #1162, opened 2021-02-24
[release-5.10] copy: compute blob compression on reused blobs based on source MediaType

nalind:
Teach BlobInfoCache2 implementations to be able to retrieve the type of compression that was applied to a blob with a given digest, if we know it.

If we successfully reuse a blob while writing an image, and we know how a blob with that reused blob's digest was compressed, use that information when updating the MIME type for the reused blob.

This covers situations where TryReusingBlob() returns a BlobInfo containing a different digest and MIME type than the one we originally asked it to reuse. copy.imageCopier.copyLayers() sets it up as part of the LayerInfos in update info that copy.imageCopier.copyUpdatedConfigAndManifest() passes to src.UpdatedImage(), and the UpdateLayerInfos() implementations discard the MIME type from the new BlobInfo in favor of the MIME type in the original layers list.

The buildah blob cache triggers this case when we use it while pushing an image where the local copy of the manifest specifies an uncompressed digest and MIME type, and the blob cache tries to substitute the digest of the compressed version in LayerInfosForCopy(), since it knows it can retrieve the compressed version from its cache. The blob cache doesn't get used very often, so we've only recently started noticing it since we started testing #1089 in openshift/origin#25830.

Cherry picked from #1138.

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -1349,7 +1349,15 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
 		compressionOperation = types.PreserveOriginal
 		inputInfo = srcInfo
 		uploadCompressorName = srcCompressorName
-		uploadCompressionFormat = nil
+		// Remember if the original blob was compressed, and if so how, so that if
+		// LayerInfosForCopy() returned something that differs from what was in the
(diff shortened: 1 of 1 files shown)


vrothberg (commented):
LGTM

vrothberg:
I'll cut a new release in a jiffy.

PR #1167, opened 2021-03-02
storage: GetBlob: write to a local tempfile

vrothberg:
When reading blobs (e.g., layers) from the storage, write the data to a
temporary file first and return its file descriptor.

The main goal is to keep the time of the storage being locked as short
as possible.  With this approach we are disk-bound which is generally
faster than being network bound.

The motivation for this change is the desire to allow for accessing the
storage during pushes.  We have attempted to optimize certain execution
paths in containers/storage but those require the callers to know
whether data will be manipulated or read, which is far from being
trivial.  The daemon-less nature of the storage also requires to write
certain data.

This approach here is entirely transparent to the storage.

### DIFF
--- a/storage/storage_image.go
+++ b/storage/storage_image.go
@@ -45,6 +45,7 @@ var (
 type storageImageSource struct {
 	imageRef        storageReference
 	image           *storage.Image
+	systemContext   *types.SystemContext    // SystemContext used in GetBlob() to create temporary files
 	layerPosition   map[digest.Digest]int   // Where we are in reading a blob's layers
 	cachedManifest  []byte                  // A cached copy of the manifest, if already known, or nil
 	getBlobMutex    sync.Mutex              // Mutex to sync state for parallel GetBlob executions
@@ -96,6 +97,7 @@ func newImageSource(ctx context.Context, sys *types.SystemContext, imageRef stor
 	// Build the reader object.
 	image := &storageImageSource{
 		imageRef:        imageRef,
+		systemContext:   sys,
 		image:           img,
 		layerPosition:   make(map[digest.Digest]int),
 		SignatureSizes:  []int{},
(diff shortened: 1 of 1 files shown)


vrothberg:
@rhatdan @giuseppe @nalind PTAL

rhatdan:
LGTM

storage/storage_image.go
```diff
@@ -93,8 +94,14 @@ func newImageSource(ctx context.Context, sys *types.SystemContext, imageRef stor
 		return nil, err
 	}
 
+	directory, err := ioutil.TempDir(tmpdir.TemporaryDirectoryForBigFiles(sys), "storage")
```
mtrmac, line 97: This is never removed? Does it even need to exist?

storage/storage_image.go
```diff
@@ -44,6 +44,7 @@ var (
 
 type storageImageSource struct {
 	imageRef        storageReference
+	directory       string // Temporary directory where we store blobs until Commit() time
```
mtrmac, line 47: Please update the comment (and, absolutely non-blocking, maybe move it below the `image` ~constant member).

mtrmac (commented):
(Just a drive-by, I don’t have an opinion on the performance trade-off.)

storage/storage_image.go
vrothberg: I'll check where we could remove it.

I think it needs to exist here, since GetBlob doesn't have a `SystemContext` that we need for the big temp dir.

mtrmac: `storageImageSource` could just hold a reference to the `SystemContext` directly.

vrothberg: Even better, thanks :)

lsm5:
@vrothberg could you perhaps suggest a test to measure this change? And would it make sense to include such a test in CI ? /cc @ashcrow

vrothberg:
> @vrothberg could you perhaps suggest a test to measure this change? And would it make sense to include such a test in CI ? /cc @ashcrow

It's rather tough to test such performance-oriented changes. A simple manual test is to for a long copy operation to a registry (e.g., podman push) and in parallel accessing the storage (e.g., podman images).  The (expected) result is that second operation is blocked only for a short period of time in contrast to being blocked for the entire copy/push operation as before.

vrothberg:
> toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit

Let's wait a bit :^)

TomSweeneyRedHat:
LGTM
assuming rate limits let the test through.

lsm5 (commented):
Change LGTM. I'll try a test with this change and get back if I see anything. Guess we gotta wait anyway for the rate-limiting issue.

vrothberg:
Green, and happy.

lsm5 (approved):
Tested locally just now, this is really nice. `podman images` and other commands don't have to wait for podman push anymore!

LGTM.

lsm5:
I want to hit merge, but just checking if it's ok to `Rebase and merge` or do you prefer a merge commit be added?

vrothberg:
> I want to hit merge, but just checking if it's ok to `Rebase and merge` or do you prefer a merge commit be added?

"Rebase + merge" is the default in c/image. Thanks for reviewing and testing!

akostadinov:
How about using a hard link to avoid unnecessary disk writes?

giuseppe:
> How about using a hard link to avoid unnecessary disk writes?

we'd need to make sure the same file system is used.  For that, we could use the new `staging directory` API that was added to c/storage to address exactly this issue

mtrmac:
> How about using a hard link to avoid unnecessary disk writes?

The data is typically a stream (formed at runtime with tar-split data), not a file that could be linked.

PR #1168, opened 2021-03-03
storageImageSource.LayerInfosForCopy(): return uncompressed MediaTypes

nalind:
Since #1138, we've been paying attention to the MediaType values returned by LayerInfosForCopy(), so lying about layer data being compressed when the data we provide isn't would cause uncompressed layers to be mistakenly marked as compressed when we copied the image without attempting to add or modify the compression of layers.


vrothberg (commented):
LGTM

Would you backport as well?

nalind:
After it passes tests and we merge it, I can open a cherry-pick PR for the release-5.10 branch.

PR #1169, opened 2021-03-03
[release-5.10] storageImageSource.LayerInfosForCopy(): return uncompressed MediaTypes

nalind:
Since v5.10.4, we've been paying attention to the MediaType values returned by LayerInfosForCopy(), so lying about layer data being compressed when the data we provide isn't would cause uncompressed layers to be mistakenly marked as compressed when we copied the image without attempting to add or modify the compression of layers.  Cherry picked from #1168.


vrothberg (commented):
LGTM

PR #1176, opened 2021-03-12
platform matcher: use compat matrix

vrothberg:
Use the entire compatibility matrix when matching the platform of a
specific image to possible candidates.

Context: github.com/containers/podman/issues/9637

### DIFF
--- a/internal/pkg/platform/platform_matcher.go
+++ b/internal/pkg/platform/platform_matcher.go
@@ -123,14 +123,6 @@ var compatibility = map[string][]string{
 	"arm64": {"v8"},
 }
 
-// baseVariants contains, for a specified architecture, a variant that is known to be
-// supported by _all_ machines using that architecture.
-// Architectures that don’t have variants, or where there are possible versions without
-// an established variant name, should not have an entry here.
-var baseVariants = map[string]string{
-	"arm64": "v8",
-}
-
 // WantedPlatforms returns all compatible platforms with the platform specifics possibly overridden by user,
 // the most compatible platform is first.
 // If some option (arch, os, variant) is not present, a value from current platform is detected.
@@ -158,6 +150,8 @@ func WantedPlatforms(ctx *types.SystemContext) ([]imgspecv1.Platform, error) {
 
 	var variants []string = nil
 	if wantedVariant != "" {
+		// If the user requested a specific variant, we'll walk down
+		// the list from most to least compatible.
 		if compatibility[wantedArch] != nil {
 			variantOrder := compatibility[wantedArch]
 			for i, v := range variantOrder {
@@ -171,12 +165,14 @@ func WantedPlatforms(ctx *types.SystemContext) ([]imgspecv1.Platform, error) {
 			// user wants a variant which we know nothing about - not even compatibility
 			variants = []string{wantedVariant}
 		}
+		// Make sure to have a candidate with an empty variant as well.
 		variants = append(variants, "")
 	} else {
-		variants = append(variants, "") // No variant specified, use a “no variant specified” image if present
-		if baseVariant, ok := baseVariants[wantedArch]; ok {
-			// But also accept an image with the “base” variant for the architecture, if it exists.
-			variants = append(variants, baseVariant)
+		// Make sure to have a candidate with an empty variant as well.
+		variants = append(variants, "")
+		// If available add the entire compatibility matrix for the specific architecture.
+		if possibleVariants, ok := compatibility[wantedArch]; ok {
+			variants = append(variants, possibleVariants...)
 		}
 	}
 
--- a/internal/pkg/platform/platform_matcher_test.go
+++ b/internal/pkg/platform/platform_matcher_test.go
@@ -20,12 +20,22 @@ func TestWantedPlatforms(t *testing.T) {
 				{OS: "linux", Architecture: "amd64", Variant: ""},
 			},
 		},
-		{ // ARM
+		{ // ARM with variant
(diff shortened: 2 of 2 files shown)


vrothberg:
@rhatdan @giuseppe @lsm5 PTAL

rhatdan:
LGTM

giuseppe (approved):
/lgtm

mtrmac:
Why is this the right thing to do? Nothing in here even says what it is fixing. Is that https://github.com/containers/podman/issues/9637#issuecomment-797398399 ?

If the manifest list lists an arm-≥v7 image, and the user says “get me an arm-unspecified image”, why is it OK to use an image that doesn’t run on v6?

If the manifest list lists an arm-≥v7 image AND an arm-≥v5 image, and the user says “get me an arm-unspecified image”, why does it make sense to prefer the ≥v7 one to the ≥v5 one?

**Assuming “no variant” is the most universally compatible, it is completely inconsistent to start with the "no variant" image and then go down the list from the most-restrictive variant to the least-restrictive**

Please write down
- expected/desired behavior
- what is wrong about the previous logic
- what is right about the new one.

mtrmac:
For starters, if this is about the `docker.io/syncthing/syncthing:latest` image, the original report said:
> `podman pull` is able to pull the right image without any further hints 
and only complained about `podman auto-update` not finding that image.

Please explain how is that report related to this PR, and explain how is https://github.com/containers/podman/issues/9637#issuecomment-797398399 related to the original problem report at all.

vrothberg:
`podman pull` does not fetch the image, it fails with the same error (see https://github.com/containers/podman/issues/9637#issuecomment-797398399).

> Please write down
> 
>     * expected/desired behavior

If `variant==""`, still match other candidates in the matrix.

>     * what is wrong about the previous logic

It does not match other variants.

>     * what is right about the new one.

It matches Docker.

mtrmac:
> `podman pull` does not fetch the image, it fails with the same error (see [containers/podman#9637 (comment)](https://github.com/containers/podman/issues/9637#issuecomment-797398399)).

That’s explicitly not representative of native on-ARM execution, and AFAICS irrelevant to the original report.

If someone is pulling on a non-ARM architecture, we can’t very well guess which variant the user intends to deploy the images on, the user must specify that. Even assuming “it matches Docker” is a good enough justification, and that using the most-restrictive variant is desirable, the produced order starting with `''"` is inconsistent with that goal.

vrothberg:
> > `podman pull` does not fetch the image, it fails with the same error (see [containers/podman#9637 (comment)](https://github.com/containers/podman/issues/9637#issuecomment-797398399)).
> 
> That’s explicitly not representative of native on-ARM execution, and AFAICS irrelevant to the original report.
> 
> If someone is pulling on a non-ARM architecture, we can’t very well guess which variant the user intends to deploy the images on, the user must specify that.

But we are not guessing on non-ARM architectures, or am I missing something?

> Even assuming “it matches Docker” is a good enough justification, and that using the most-restrictive variant is desirable, the produced order starting with `''"` is inconsistent with that goal.

That's preserving the previous behavior.  If no variant is specified and an image with non exists, than it's preferred over those with variants.

mtrmac:
> But we are not guessing on non-ARM architectures, or am I missing something?

Exactly; we can’t guess, the user _must tell us_. That’s what `VariantChoice` is for, and then we behave exactly the same as in the native case with that variant auto-detected.

> That's preserving the previous behavior. 

Well, it’s preserving the previously-tested behavior. It’s not preserving the previous _semantics_. The semantics of variants is “manifest list entry with no variant specified is more compatible than image with _any other variant_, except that it is equivalent to `baseVariants`”.

It makes no sense to the that `if wantedVariant != ""`, the order is `v7 v6 v5 ""`, but `if wantedVariant == ""`, the order is `"" v7 v6 v5`. The semantics of the manifest list in question does not change by the user specifying, or not, a `VariantChoice`.

The `WantedPlatforms` documentation, while incorrect about “most compatible”, clearly says that order of the return values matters, and the callers depend on that.

vrothberg:
> It makes no sense to the that `if wantedVariant != ""`, the order is `v7 v6 v5 ""`, but `if wantedVariant == ""`, the order is `"" v7 v6 v5`. The semantics of the manifest list in question does not change by the user specifying, or not, a `VariantChoice`.

I see some sense in that since `variant == ""` can be both, unset or explicitly set to `""`.  But I don't feel strongly as long as `arch=arm,variant=""` is able to match any other variant.

Please let me know how you want it to behave, and I can clean up on Monday.

mtrmac:
You’re right, I forgot that we can’t tell between “the user wanted `""`” and “the user made no choice”. Go’s lack of optionals again 😞 

I guess if this PR _has_ to happen, it probably does have to happen this way. But I still can’t see any connection to the original auto-update report, and it’s very non-obvious to me that following Docker here makes sense.

vrothberg:
Apologies, I should have better explained the motivations of the change in the commit message.

vrothberg:
@rhatdan @TomSweeneyRedHat, please let me know whether you want this in RHEL 8.4. I can open a BZ if needed. I think it would be good but do not feel strongly about it.

rhatdan:
I think we can wait until RHEL8.4.0.2

vrothberg:
> I think we can wait until RHEL8.4.0.2

@rhatdan, which version of Podman will that be?  I am mostly checking whether we need to cut a new c/image minor version or can backport it to the v5.10.X series.

rhatdan:
podman 3.1 or later.

PR #1179, opened 2021-03-15
Implement credsStore in credential config files

matejvasek:
Config file on macOS may contain "credsStore" property indicating what cred store should be used.
This patch checks for its presence and uses it if appropriate.

Signed-off-by: Matej Vasek <mvasek@redhat.com>

### DIFF
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -60,7 +60,7 @@ var (
 // Returns a human-redable description of the location that was updated.
 // NOTE: The return value is only intended to be read by humans; its form is not an API,
 // it may change (or new forms can be added) any time.
-func SetCredentials(sys *types.SystemContext,  key, username, password string) (string, error) {
+func SetCredentials(sys *types.SystemContext, key, username, password string) (string, error) {
 	isNamespaced, err := validateKey(key)
 	if err != nil {
 		return "", err
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -161,10 +161,13 @@ func GetAllCredentials(sys *types.SystemContext) (map[string]types.DockerAuthCon
 					addRegistry(registry)
 				}
 				if auths.CredsStore != "" {
-					if creds, err := listAuthsFromCredHelper(auths.CredsStore); err == nil {
+					creds, err := listAuthsFromCredHelper(auths.CredsStore)
+					if err == nil {
 						for registry := range creds {
 							addRegistry(registry)
 						}
+					} else {
+						return nil, err
 					}
 				}
 			}
@@ -467,10 +470,16 @@ func RemoveAllAuthentication(sys *types.SystemContext) error {
 				auths.CredHelpers = make(map[string]string)
 				auths.AuthConfigs = make(map[string]dockerAuthConfig)
 				if auths.CredsStore != "" {
-					if creds, err := listAuthsFromCredHelper(auths.CredsStore); err == nil {
+					creds, err := listAuthsFromCredHelper(auths.CredsStore)
+					if err == nil {
 						for registry := range creds {
-							_ = deleteAuthFromCredHelper(auths.CredsStore, registry)
+							err = deleteAuthFromCredHelper(auths.CredsStore, registry)
+							if err != nil {
+								return false, err
+							}
 						}
+					} else {
+						return false, err
 					}
 
 				}
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -401,6 +401,21 @@ func RemoveAuthentication(sys *types.SystemContext, key string) error {
 				if innerHelper, exists := auths.CredHelpers[key]; exists {
 					removeFromCredHelper(innerHelper)
 				}
+				if auths.CredsStore != "" {
+					c, err := getAuthFromCredHelper(auths.CredsStore, key)
+					if err != nil {
+						multiErr = multierror.Append(multiErr, errors.Wrapf(err, "removing credentials for %s from credential helper %s", key, auths.CredsStore))
+					} else {
+						if c.Username != "" || c.IdentityToken != "" {
+							err = deleteAuthFromCredHelper(auths.CredsStore, key)
+							if err != nil {
+								multiErr = multierror.Append(multiErr, errors.Wrapf(err, "removing credentials for %s from credential helper %s", key, auths.CredsStore))
+							} else {
+								isLoggedIn = true
+							}
+						}
+					}
+				}
 				if _, ok := auths.AuthConfigs[key]; ok {
 					isLoggedIn = true
 					delete(auths.AuthConfigs, key)
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -30,6 +30,7 @@ type dockerAuthConfig struct {
 type dockerConfigFile struct {
 	AuthConfigs map[string]dockerAuthConfig `json:"auths"`
 	CredHelpers map[string]string           `json:"credHelpers,omitempty"`
+	CredsStore  string                      `json:"credsStore,omitempty"`
 }
 
 type authPath struct {
@@ -59,7 +60,7 @@ var (
 // Returns a human-redable description of the location that was updated.
 // NOTE: The return value is only intended to be read by humans; its form is not an API,
 // it may change (or new forms can be added) any time.
-func SetCredentials(sys *types.SystemContext, key, username, password string) (string, error) {
+func SetCredentials(sys *types.SystemContext,  key, username, password string) (string, error) {
 	isNamespaced, err := validateKey(key)
 	if err != nil {
 		return "", err
@@ -84,6 +85,9 @@ func SetCredentials(sys *types.SystemContext, key, username, password string) (s
 						return false, unsupportedNamespaceErr(ch)
 					}
 					return false, setAuthToCredHelper(ch, key, username, password)
+				}
+				if auths.CredsStore != "" {
+					return false, setAuthToCredHelper(auths.CredsStore, key, username, password)
 				}
 				creds := base64.StdEncoding.EncodeToString([]byte(username + ":" + password))
 				newCreds := dockerAuthConfig{Auth: creds}
@@ -155,6 +159,13 @@ func GetAllCredentials(sys *types.SystemContext) (map[string]types.DockerAuthCon
 				}
 				for registry := range auths.AuthConfigs {
 					addRegistry(registry)
+				}
+				if auths.CredsStore != "" {
+					if creds, err := listAuthsFromCredHelper(auths.CredsStore); err == nil {
+						for registry := range creds {
+							addRegistry(registry)
+						}
+					}
 				}
 			}
 		// External helpers.
@@ -440,6 +451,14 @@ func RemoveAllAuthentication(sys *types.SystemContext) error {
 				}
 				auths.CredHelpers = make(map[string]string)
 				auths.AuthConfigs = make(map[string]dockerAuthConfig)
+				if auths.CredsStore != "" {
+					if creds, err := listAuthsFromCredHelper(auths.CredsStore); err == nil {
+						for registry := range creds {
+							_ = deleteAuthFromCredHelper(auths.CredsStore, registry)
+						}
+					}
+
+				}
 				return true, nil
 			})
 		// External helpers.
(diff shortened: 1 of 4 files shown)


mtrmac:
(Drive-by comment, by no means a review:)
- This is not macOS specific
- Read support without write support can have surprising behavior
- See also #737 (and #971 ).

matejvasek:
@QiWang19 does https://github.com/containers/image/pull/971 fixes this?

matejvasek:
@vrothberg do you know whether #971 fixes this?

vrothberg:
Would https://github.com/containers/image/pull/1193 help?

lkingland:
This seems to be working for me.

Given the following `~/.docker/config.json`
```
{
  "HttpHeaders" : {
    "User-Agent" : "Docker-Client/19.03.12 (darwin)"
  },
  "credsStore" : "desktop",
  "auths" : {
    "https://index.docker.io/v1/" : {

    }
  },
  "experimental" : "enabled",
  "stackOrchestrator" : "swarm"
}
```
The following program:
```
package main

import (
        "fmt"
        "github.com/containers/image/v5/pkg/docker/config"
        "log"
)

func main() {
        dauth, err := config.GetAllCredentials(nil)
        if err != nil {
                log.Fatal(err)
        }
        fmt.Printf("dauths: %+v\n", dauth)
}
```
accesses credentials from the credsStore:
```
> go run main.go
dauths: map[index.docker.io:{Username:[REDACTED] Password:[REDACTED] IdentityToken:}]
```

lkingland:
> Would #1193 help?

Unfortunately, when trying with #1193, the Username and Password output are blank

rhuss:
I have a similar issue, in that `config.GetAllCredentials(nil)` does not find my real passwords, but returns bogus (possibly stale) passwords. Even when I do a `docker logout` the passwords are still present.

I'm a bit puzzled as I checked all credentials helpers available (docker-credential-desktop,  docker-credential-ecr-login, docker-credential-osxkeychain) and none has the bogus password that `containers/images` find. 

My config.json looks the same:

```json
{
	"auths": {
		"index.docker.io": {},
		"public.ecr.aws": {},
		"quay.io": {}
	},
	"credsStore": "desktop",
	"experimental": "enabled",
	"stackOrchestrator": "swarm"
}%
```

Any help is highly appreciated.

matejvasek:
@vrothberg ^^^ any idea what could be issue for Roland?

mtrmac:
@rhuss this seems to have nothing to do with this proposed PR, please file a separate issue.

rhuss:
> @rhuss this seems to have nothing to do with this proposed PR, please file a separate issue.

I added my comments here, as I think that is the issue for implementing support for credential helpers, right ? --> https://github.com/containers/image/issues/111#issuecomment-861238976

mtrmac:
> > Would #1193 help?
> 
> Unfortunately, when trying with #1193, the Username and Password output from the aforementioned program are blank

#1193 requires the configuration to be present in a `registries.conf` option named `credential-helpers`, not in `credsStore`.

rhatdan:
@matejvasek Are you still working on this?

matejvasek:
@rhatdan for now I made a workaround in our SW. I could get back to this if there is no one else willing to work on this. For now my priority is to fix https://github.com/containers/podman/issues/10776.

matejvasek:
@mtrmac @vrothberg PTAL

pkg/docker/config/config.go
```diff
@@ -440,6 +451,14 @@ func RemoveAllAuthentication(sys *types.SystemContext) error {
 				}
 				auths.CredHelpers = make(map[string]string)
 				auths.AuthConfigs = make(map[string]dockerAuthConfig)
+				if auths.CredsStore != "" {
+					if creds, err := listAuthsFromCredHelper(auths.CredsStore); err == nil {
+						for registry := range creds {
+							_ = deleteAuthFromCredHelper(auths.CredsStore, registry)
```
mtrmac, line 457: Same here.

pkg/docker/config/config.go
```diff
@@ -646,6 +665,12 @@ func findAuthentication(ref reference.Named, registry, path string, legacyFormat
 		return getAuthFromCredHelper(ch, registry)
 	}
 
+	if auths.CredsStore != "" {
+		if cred, err := getAuthFromCredHelper(auths.CredsStore, registry); err == nil {
```
mtrmac, line 669: Why is it OK to completely ignore errors here?

pkg/docker/config/config_test.go
```diff
@@ -559,6 +559,7 @@ func TestGetAuthFailsOnBadInput(t *testing.T) {
 }
 
 func TestGetAllCredentials(t *testing.T) {
```
mtrmac, line 561: Please add tests for the other modified functions as well.

pkg/docker/config/config_test.go
```diff
@@ -636,6 +637,42 @@ func TestGetAllCredentials(t *testing.T) {
...
+	f, err := os.OpenFile(authFilePath, os.O_RDWR|os.O_TRUNC, 0644)
+	require.NoError(t, err)
+	_, err = f.Write([]byte(`{ "credsStore": "helper-registry" }`))
+	require.NoError(t, err)
+	err = f.Close()
+	require.NoError(t, err)
+	sys = types.SystemContext{
+		AuthFilePath:                authFilePath,
+		SystemRegistriesConfPath:    "",
+		SystemRegistriesConfDirPath: "",
```
mtrmac, line 650: This is (sadly?) the same as leaving the values unset, in particular it doesn’t disable any ordinary users’ configuration. So if this data matters, it should point at real files.

pkg/docker/config/config_test.go
```diff
@@ -636,6 +637,42 @@ func TestGetAllCredentials(t *testing.T) {
...
+	dir, err := ioutil.TempDir("", "test-home")
+	if err != nil {
+		t.Fatal(err)
+		return func() {}
+	}
+	oldHome, oldHomeExists := os.LookupEnv("HOME")
+	os.Setenv("HOME", dir)
+	return func() {
+		if oldHomeExists {
+			os.Setenv("HOME", oldHome)
```
mtrmac, line 670: I don’t immediately see that this is necessary (the existing `GetAllCredentials` test got by without), and anyway, AFAIK it doesn’t work; `containers/storage/pkg/homedir.Get()` caches the value throughout its lifetime, so a single-test override is either completely ineffective, or is effective only if that test happens to run first.

So, to make some of the code testable, we had to introduce `getCredentialsWithHomeDir` and I suspect something similar would be necessary in other cases that tried to override `HOME`.

pkg/docker/config/config.go
```diff
@@ -440,6 +451,14 @@ func RemoveAllAuthentication(sys *types.SystemContext) error {
 				}
 				auths.CredHelpers = make(map[string]string)
 				auths.AuthConfigs = make(map[string]dockerAuthConfig)
+				if auths.CredsStore != "" {
```
mtrmac, line 454: `RemoveAuthentication` also needs adding the support.

pkg/docker/config/config.go
```diff
@@ -85,6 +86,9 @@ func SetCredentials(sys *types.SystemContext, key, username, password string) (s
 					}
 					return false, setAuthToCredHelper(ch, key, username, password)
 				}
+				if auths.CredsStore != "" {
+					return false, setAuthToCredHelper(auths.CredsStore, key, username, password)
+				}
```
mtrmac, line 91: Please keep the order of handling `CredHelpers`/`CredsStore`/`AuthConfigs` consistent throughout the package, to make it easier to check that all operations handle the same data. (I guess in the order above, unless there’s a specific reason to do something else.)

mtrmac (commented):
Thanks.

The implementation looks reasonable reading each operation _in isolation_, but does this actually _work_ end-to-end?

- The user sets `$XDG_RUNTIME_DIR/containers/auth.json` to contain a `credsStore`; that works fine for a while, but is wiped on a reboot. Not good.
- The user sets `~/.docker/config.json` to contain a `credsStore`; that allows reading fine, and survives reboot, but the first modification (`SetCredentials`/`RemoveAuthentication` _only_ reads `auth.json`) _only_ reads `…/auth.json`, doesn’t see any `credsStore` entry in there, and happily writes credentials to `…/auth.json`. That’s not good either.

(Yes, this was already broken with `credHelpers`; in that case it’s more plausible to argue that the read-only path is good enough because the helper might point at a cloud-specific helper and credentials _for that registry_ are never updated via this package. That’s not the case if this is intended to handle all credentials.

I realize that this is not an issue on macOS, where we don’t have `$XDG_RUNTIME_DIR` and the configuration is persistent; having a macOS-only feature is somewhat a stretch, having it available and badly broken on Linux is not really acceptable.

[That’s why we have ended up implementing the `registries.conf` helper configuration.])

@vrothberg PTAL independently, this is tricky and I could well be missing something.

pkg/docker/config/config.go
```diff
@@ -59,7 +60,7 @@ var (
 // Returns a human-redable description of the location that was updated.
 // NOTE: The return value is only intended to be read by humans; its form is not an API,
 // it may change (or new forms can be added) any time.
-func SetCredentials(sys *types.SystemContext, key, username, password string) (string, error) {
+func SetCredentials(sys *types.SystemContext,  key, username, password string) (string, error) {
```
vrothberg: Nit: not sure why the extra space is needed.

vrothberg (commented):
Really just found a mini nit in addition to @mtrmac's comments.

pkg/docker/config/config_test.go
matejvasek: @mtrmac without this the test is failing on my machine because the  `GetAllCredentials` function also returns credentials from my docker config.

pkg/docker/config/config_test.go
matejvasek: @mtrmac Do I need to set it? I think that for this test `AuthFilePath` is enough, isn't it?

matejvasek:
>I realize that this is not an issue on macOS, where we don’t have $XDG_RUNTIME_DIR and the configuration is persistent; having a macOS-only feature is somewhat a stretch, having it available and badly broken on Linux is not really acceptable.

@mtrmac actually this is not just macOS specific, you can use credStore also on Linux (and Windows).

pkg/docker/config/config_test.go
matejvasek: >Please add tests for the other modified functions as well.

Are those functions properly tested right now anyway? I believe they are not tested with respect to helpers right now.

pkg/docker/config/config_test.go
matejvasek: I you don't have `$HOME/.docker/config.json` on your machine it probably works. But for me it's failing.

pkg/docker/config/config_test.go
matejvasek: >I believe they are not tested with respect to helpers right now.

For instance the mock helper does noting for the store operation so you can really assert it written anything when calling the `SetCredentials` function.

matejvasek:
>The user sets $XDG_RUNTIME_DIR/containers/auth.json to contain a credsStore; that works fine for a while, but is wiped on a reboot. Not good.

@mtrmac  so user has to login after each reboot with `podman`?

matejvasek:
>The user sets $XDG_RUNTIME_DIR/containers/auth.json to contain a credsStore; that works fine for a while, but is wiped on a reboot. Not good.

Doesn't this also apply to already used `credHelpers` field?

pkg/docker/config/config_test.go
matejvasek: >caches the value throughout its lifetime, so a single-test override is either completely ineffective, or is effective only if that test happens to run first.

I don't what to do about this.

matejvasek: The test is touch not only testing config files but also actual user config files, that's not good.

matejvasek:
Aren't most objections against `credHelpers` also valid to already implemented `credHelpers`?

matejvasek:
What about if we implemented only read only part?

matejvasek:
@mtrmac What about if we implemented only read only part?

matejvasek:
@mtrmac why is `containers/auth.json` not persistent?

jankaluza:

PR #1180, opened 2021-03-15
Update signature-protocols.md

I-Cat:
* Fix urls, http:// and https:// and File:// *


mtrmac (commented):
`file://foo/bar` means “file `/bar` on network host `foo`”; this intends to express that only _truly local_ references are supported, that’s why the `///` is important. No such strong need for a distinction exists for `http`/`https`.

PR #1185, opened 2021-03-17
copy to storage: commit layers as early as possible

vrothberg:
When copying an image to a storage destination, make sure to commit
downloaded layers as early as possible.  Since layers are committed in
sequence, we need to preserve the index of each layer: layer N can only
be committed to the storage when all N-1 previous layers are already
present in the storage, in the given order.

Two new methods are added to the storage destination to pass down the
index to TryResusingBlob and PutBlob.  The idea is fairly simple:

 * PutBlobWithOptions and TryReusingBlobWithOptions can "queue in" work.

 * Exaclty one caller of them may enter a second stage which is
   committing the previously queued in blobs to the storage while
   preserving the order.  If one caller is already in stage two,
   others will return early freeing up goroutines workers to continue
   pulling more blobs if needed.

Note: since this optimization is limited to storage destinations, we do
not need to extend the entire public API with methods that only some
types implement.  Hence, add an internal `ImageDestinationWithOptions`
interface to the new `internal/types` package.

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -16,6 +16,7 @@ import (
 	"github.com/containers/image/v5/image"
 	internalblobinfocache "github.com/containers/image/v5/internal/blobinfocache"
 	"github.com/containers/image/v5/internal/pkg/platform"
+	internalTypes "github.com/containers/image/v5/internal/types"
 	"github.com/containers/image/v5/manifest"
 	"github.com/containers/image/v5/pkg/blobinfocache"
 	"github.com/containers/image/v5/pkg/compression"
@@ -920,7 +921,7 @@ func (ic *imageCopier) copyLayers(ctx context.Context) error {
 				logrus.Debugf("Skipping foreign layer %q copy to %s", cld.destInfo.Digest, ic.c.dest.Reference().Transport().Name())
 			}
 		} else {
-			cld.destInfo, cld.diffID, cld.err = ic.copyLayer(ctx, srcLayer, toEncrypt, pool)
+			cld.destInfo, cld.diffID, cld.err = ic.copyLayer(ctx, srcLayer, toEncrypt, pool, index)
 		}
 		data[index] = cld
 	}
@@ -1116,7 +1117,7 @@ func (c *copier) copyConfig(ctx context.Context, src types.Image) error {
 			progressPool, progressCleanup := c.newProgressPool(ctx)
 			defer progressCleanup()
 			bar := c.createProgressBar(progressPool, srcInfo, "config", "done")
-			destInfo, err := c.copyBlobFromStream(ctx, bytes.NewReader(configBlob), srcInfo, nil, false, true, false, bar)
+			destInfo, err := c.copyBlobFromStream(ctx, bytes.NewReader(configBlob), srcInfo, nil, false, true, false, bar, -1)
 			if err != nil {
 				return types.BlobInfo{}, err
 			}
@@ -1142,7 +1143,7 @@ type diffIDResult struct {
 
 // copyLayer copies a layer with srcInfo (with known Digest and Annotations and possibly known Size) in src to dest, perhaps (de/re/)compressing it,
 // and returns a complete blobInfo of the copied layer, and a value for LayerDiffIDs if diffIDIsNeeded
-func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, toEncrypt bool, pool *mpb.Progress) (types.BlobInfo, digest.Digest, error) {
+func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, toEncrypt bool, pool *mpb.Progress, layerIndex int) (types.BlobInfo, digest.Digest, error) {
 	// If the srcInfo doesn't contain compression information, try to compute it from the
 	// MediaType, which was either read from a manifest by way of LayerInfos() or constructed
 	// by LayerInfosForCopy(), if it was supplied at all.  If we succeed in copying the blob,
@@ -1175,7 +1176,26 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, to
 		// a failure when we eventually try to update the manifest with the digest and MIME type of the reused blob.
 		// Fixing that will probably require passing more information to TryReusingBlob() than the current version of
 		// the ImageDestination interface lets us pass in.
-		reused, blobInfo, err := ic.c.dest.TryReusingBlob(ctx, srcInfo, ic.c.blobInfoCache, ic.canSubstituteBlobs)
+		var (
+			blobInfo types.BlobInfo
+			reused   bool
+			err      error
+		)
+		// Note: the storage destination optimizes the committing of
+		// layers which requires passing the index of the layer.
+		// Hence, we need to special case and cast.
+		dest, ok := ic.c.dest.(internalTypes.ImageDestinationWithOptions)
+		if ok {
+			options := internalTypes.TryReusingBlobOptions{
+				Cache:         ic.c.blobInfoCache,
+				CanSubstitute: ic.canSubstituteBlobs,
+				LayerIndex:    &layerIndex,
+			}
+			reused, blobInfo, err = dest.TryReusingBlobWithOptions(ctx, srcInfo, options)
+		} else {
+			reused, blobInfo, err = ic.c.dest.TryReusingBlob(ctx, srcInfo, ic.c.blobInfoCache, ic.canSubstituteBlobs)
+		}
+
 		if err != nil {
 			return types.BlobInfo{}, "", errors.Wrapf(err, "Error trying to reuse blob %s at destination", srcInfo.Digest)
 		}
@@ -1217,7 +1237,7 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, to
 
 	bar := ic.c.createProgressBar(pool, srcInfo, "blob", "done")
 
-	blobInfo, diffIDChan, err := ic.copyLayerFromStream(ctx, srcStream, types.BlobInfo{Digest: srcInfo.Digest, Size: srcBlobSize, MediaType: srcInfo.MediaType, Annotations: srcInfo.Annotations}, diffIDIsNeeded, toEncrypt, bar)
+	blobInfo, diffIDChan, err := ic.copyLayerFromStream(ctx, srcStream, types.BlobInfo{Digest: srcInfo.Digest, Size: srcBlobSize, MediaType: srcInfo.MediaType, Annotations: srcInfo.Annotations}, diffIDIsNeeded, toEncrypt, bar, layerIndex)
 	if err != nil {
 		return types.BlobInfo{}, "", err
 	}
@@ -1248,7 +1268,7 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, to
 // perhaps (de/re/)compressing the stream,
 // and returns a complete blobInfo of the copied blob and perhaps a <-chan diffIDResult if diffIDIsNeeded, to be read by the caller.
 func (ic *imageCopier) copyLayerFromStream(ctx context.Context, srcStream io.Reader, srcInfo types.BlobInfo,
-	diffIDIsNeeded bool, toEncrypt bool, bar *mpb.Bar) (types.BlobInfo, <-chan diffIDResult, error) {
+	diffIDIsNeeded bool, toEncrypt bool, bar *mpb.Bar, layerIndex int) (types.BlobInfo, <-chan diffIDResult, error) {
 	var getDiffIDRecorder func(compression.DecompressorFunc) io.Writer // = nil
 	var diffIDChan chan diffIDResult
 
@@ -1273,7 +1293,7 @@ func (ic *imageCopier) copyLayerFromStream(ctx context.Context, srcStream io.Rea
 		}
 	}
 
-	blobInfo, err := ic.c.copyBlobFromStream(ctx, srcStream, srcInfo, getDiffIDRecorder, ic.canModifyManifest, false, toEncrypt, bar) // Sets err to nil on success
+	blobInfo, err := ic.c.copyBlobFromStream(ctx, srcStream, srcInfo, getDiffIDRecorder, ic.canModifyManifest, false, toEncrypt, bar, layerIndex) // Sets err to nil on success
 	return blobInfo, diffIDChan, err
 	// We need the defer … pipeWriter.CloseWithError() to happen HERE so that the caller can block on reading from diffIDChan
 }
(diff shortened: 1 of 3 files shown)


vrothberg:
Note: I am still working on adding more code comments etc but the logic should not change.  Will open test PRs against Podman, Buildah and CRI-O as well.

vrothberg:
@nalind @giuseppe, if you find time, I'd appreciate your pairs of eyes on this change. I am now polishing up remaining bits that the other CIs reveal but I don't expect the logic to change.

copy/copy.go
```diff
@@ -124,6 +125,12 @@ type copier struct {
 	downloadForeignLayers bool
 }
 
+// hasStorageDestination returns true if the copier's `dest` points to a
+// containers-storage destination.
+func (c *copier) hasStorageDestination() bool {
+	return c.dest.Reference().Transport().Name() == internalStorage.Transport.Name()
```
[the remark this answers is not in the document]
mtrmac: Also, IIRC the blob cache or some other components may wrap transports to modify behavior, but return the original names — in that case this name check could pass but the later type check would fail instead of going down the compatibility path.

copy/copy.go
```diff
@@ -1521,7 +1550,24 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
...
+	// special case and cast.
+	if c.hasStorageDestination() {
+		storageDest, ok := c.dest.(*internalStorage.StorageImageDestination)
+		if !ok {
+			return types.BlobInfo{}, errors.Wrap(err, "Error converting to storage destination")
+		}
+		options := internalStorage.PutBlobOptions{
+			Cache:      c.blobInfoCache,
+			IsConfig:   isConfig,
+			LayerIndex: &layerIndex,
```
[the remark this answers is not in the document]
vrothberg: It'll work in both cases but it's much cleaner to set it to `nil`. Good catch!

internal/storage/storage_image.go
```diff
@@ -636,9 +701,177 @@ func (s *storageImageDestination) getConfigBlob(info types.BlobInfo) ([]byte, er
...
+	// Conceptually, the code is divided in two stages:
+	//
+	// 1) Queue in work by marking the layer as ready to be committed.
+	//    If at least one previous/parent layer with a lower index has
+	//    not yet been committed, return early.
+	//
+	// 2) Process the queued-in work by committing the "ready" layers
+	//    in sequence.  Make sure that more items can be queued-in
+	//    during the comparatively I/O expensive task of committing a
+	//    layer.
```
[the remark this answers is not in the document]
vrothberg: I considered the previous design where each caller would block until it's parent layers have been committed and then commit the layer itself.  I *think* Docker does something similar since Docker's progress bars are showing "downloaded" and "committed" message (not sure about the exact wording but conceptually they do that).

What I like about this design is:
* We have at most 6 goroutines pulling down layers in parallel
* Assume we have > 6 layers to pull

With the old design, 5 of the 6 goroutines may be blocked always waiting until they are allowed to commit. With the new design, only 1 goroutine is the "worker" committing the layers.  The other 5 can continue pulling layers and queue in work.

I like the simplicity of this design and found the code comparatively approachable.

Extended the comment to make this aspect explicit.

internal/storage/storage_image.go
```diff
@@ -666,116 +899,25 @@ func (s *storageImageDestination) Commit(ctx context.Context, unparsedToplevel t
...
 	}
 	layerBlobs := man.LayerInfos()
 	// Extract or find the layers.
-	lastLayer := ""
-	for _, blob := range layerBlobs {
-		if blob.EmptyLayer {
-			continue
+	for i, blob := range layerBlobs {
+		if err := s.commitLayer(ctx, blob, i); err != nil {
+			return err
```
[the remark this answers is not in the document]
vrothberg: I love this change as well.

copy/copy.go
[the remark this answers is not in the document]
vrothberg: Very good point, will change.

vrothberg:
> It feels a bit risky to have a storage-specific interface. 

I have no preference.  In case you feel strongly, I can follow the `internal/transport` idea. But, as you mentioned, since everything is internal, we could defer that until we have a clearer idea.

mtrmac:
> But, as you mentioned, since everything is internal, we could defer that until we have a clearer idea.

It’s not a blocker for me.

storage/storage.go
```diff
@@ -0,0 +1,32 @@
+package storage
+
+import internal "github.com/containers/image/v5/internal/storage"
+
+// Note that the storage package is merely aliasing *some* parts of the
+// `internal/storage` package.  This allows us to control the public parts of
+// the API (i.e., in this file) and the c/image internal parts of the API
+// (i.e., unaliased parts).  An immediate use case is to add new features
+// (e.g., to the `copy` package) without having to worry too much about public
+// facing changes and versioning.
```
mtrmac, line 10: BTW I don’t think this is actually true. Once a camel-case method exists, it can be called by anyone using a matching interface, whether or not the implementor is in a visible package or somewhere in `internal`.

So there isn’t much difference between
- Private interface in `c/internal/storage`; implementation in `c/image/storage` as a private type, returns `types.imageDestination`
- Private implementation in `c/internal/storage` as an exported type, returns `types.ImageDestination`

The caller is going to make a very similar check, and external users can do the same things. So I don’t think this move + forwarding is _necessary_.

OTOH a major difference is that once this is structured as a method on a type, not an interface, it’s impossible for any other internal transport to directly adopt the new API without extracting the interface.

vrothberg: I think I overstretched the statement.  If it's interface you're right on point. In this case, the extensions to `StorageImageDestination` are methods outside of the API, so I don't think any outside callers could access the `*WithOptions` methods.  Do you agree?

mtrmac: I got the syntax wrong earlier (a simple cast does statically verify that the method is known to exists), but AFAIK a dynamic check like this always works:
```go
package main

import (
	"fmt"
	"io"
)

type myType struct{}

func (myType) Close() error       { return nil }
func (myType) NotInAnyInterface() { fmt.Println("Hello!") }

func constructor() io.Closer { return &myType{} }

func writeToCloser(closer io.Closer) {
	type bypass interface{ NotInAnyInterface() }
	bypassed := closer.(bypass)
	bypassed.NotInAnyInterface()
}

func main() {
	writeToCloser(constructor())
}
```

internal/storage/storage_image.go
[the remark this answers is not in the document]
mtrmac: > With the old design, 5 of the 6 goroutines may be blocked always waiting until they are allowed to commit. With the new design, only 1 goroutine is the "worker" committing the layers. The other 5 can continue pulling layers and queue in work.

That’s a great point, thanks for writing that down.

vrothberg:
I think I made my mind up and will restructure. An `internal.ImageDestinationWithOptions` seems cleaner and more approachable than splitting `storage` in two.

vrothberg:
Done.  I did the following:
* Instead of splitting `storage` in two, I added a new internal package `internal/types`
* This new _internal_ package declares a new `ImageDestinationWithOptions` interface extending `types.ImageDestination`.

I think that's much cleaner than the initial version. Thanks, @mtrmac!

vrothberg:
Buildah (https://github.com/containers/buildah/pull/3085) and Podman (https://github.com/containers/podman/pull/9738) are green.

giuseppe (approved):
I've a feeling this is going to give me some pain with https://github.com/containers/image/pull/1084 :-) but,

LGTM

vrothberg:
> I've a feeling this is going to give me some pain with #1084 :-) but,

Thank you for reviewing.  If the "old" way would be more convenient for #1084, we could special-case the special case :)

vrothberg:
@rhatdan PTAL

internal/types/types.go
```diff
@@ -0,0 +1,53 @@
...
+// interface.
+type ImageDestinationWithOptions interface {
+	publicTypes.ImageDestination
+
+	// PutBlobWithOptions is a wrapper around PutBlob.  If
+	// options.LayerIndex is set, the blob will be committed directly.
+	// Either by the calling goroutine or by another goroutine already
+	// committing layers.
+	//
+	// Please not that TryReusingBlobWithOptions and PutBlobWithOptions
```
umohnani8, line 20: nit: s/not/note

internal/types/types.go
```diff
@@ -0,0 +1,53 @@
...
+	// *must* be used the together.  Mixing the two with the non
```
umohnani8, line 21: nit: remove extra "the"

internal/types/types.go
```diff
@@ -0,0 +1,53 @@
...
+	// "WithOptions" functions is not supported.
+	PutBlobWithOptions(ctx context.Context, stream io.Reader, blobinfo publicTypes.BlobInfo, options PutBlobOptions) (publicTypes.BlobInfo, error)
+
+	// TryReusingBlobWithOptions is a wrapper around TryReusingBlob.  If
+	// options.LayerIndex is set, the reused blob will be recoreded as
+	// already pulled.
+	//
+	// Please not that TryReusingBlobWithOptions and PutBlobWithOptions
```
umohnani8, line 29: nit: s/not/note

internal/types/types.go
```diff
@@ -0,0 +1,53 @@
...
+	// *must* be used the together.  Mixing the two with the non
```
umohnani8, line 30: nit: remove extra "the"

storage/storage_image.go
```diff
@@ -410,6 +418,26 @@ func (s *storageImageDestination) computeNextBlobCacheFile() string {
 	return filepath.Join(s.directory, fmt.Sprintf("%d", atomic.AddInt32(&s.nextTempFileID, 1)))
 }
 
+// PutBlobWithOptions is a wrapper around PutBlob.  If options.LayerIndex is
+// set, the blob will be committed directly.  Either by the calling goroutine
+// or by another goroutine already committing layers.
+//
+// Please not that TryReusingBlobWithOptions and PutBlobWithOptions *must* be
```
umohnani8, line 425: nit: s/not/note

storage/storage_image.go
```diff
@@ -410,6 +418,26 @@ func (s *storageImageDestination) computeNextBlobCacheFile() string {
...
+// used the together.  Mixing the two with the non "WithOptions" functions
```
umohnani8, line 426: nit: same here remove extra "the"

umohnani8:
Few comment nits, but LGTM

storage/storage_image.go
```diff
@@ -636,6 +680,179 @@ func (s *storageImageDestination) getConfigBlob(info types.BlobInfo) ([]byte, er
...
+
+	// Carry over the previous ID for empty non-base layers.
+	if blob.EmptyLayer && index > 0 {
+		s.indexToStorageID[index] = &lastLayer
+		return nil
+	}
+
+	// Check if there's already a layer with the ID that we'd give to the result of applying
+	// this layer blob to its parent, if it has one, or the blob's hex value otherwise.
+	diffID, haveDiffID := s.blobDiffIDs[blob.Digest]
```
mtrmac, line 758: `blobDiffIDs` is protected by `putBlobMutex`, which is not held AFAICS. Also elsewhere.

storage/storage_image.go
```diff
@@ -636,6 +680,179 @@ func (s *storageImageDestination) getConfigBlob(info types.BlobInfo) ([]byte, er
...
+	if !haveDiffID {
+		// Check if it's elsewhere and the caller just forgot to pass it to us in a PutBlob(),
+		// or to even check if we had it.
+		// Use none.NoCache to avoid a repeated DiffID lookup in the BlobInfoCache; a caller
+		// that relies on using a blob digest that has never been seen by the store had better call
+		// TryReusingBlob; not calling PutBlob already violates the documented API, so there’s only
+		// so far we are going to accommodate that (if we should be doing that at all).
+		logrus.Debugf("looking for diffID for blob %+v", blob.Digest)
+		has, _, err := s.TryReusingBlob(ctx, blob.BlobInfo, none.NoCache, false)
```
mtrmac, line 767: Please add a note that this is not `TryReusingBlobWithOptions`, i.e. that this can’t recurse.

storage/storage_image.go
```diff
@@ -636,6 +680,179 @@ func (s *storageImageDestination) getConfigBlob(info types.BlobInfo) ([]byte, er
...
+	}
+	if layer, err2 := s.imageRef.transport.store.Layer(id); layer != nil && err2 == nil {
+		// There's already a layer that should have the right contents, just reuse it.
+		lastLayer = layer.ID
+		s.indexToStorageID[index] = &lastLayer
+		return nil
+	}
+	// Check if we previously cached a file with that blob's contents.  If we didn't,
+	// then we need to read the desired contents from a layer.
+	filename, ok := s.filenames[blob.Digest]
```
mtrmac, line 791: `putBlobMutex`?

storage/storage_image.go
```diff
@@ -66,6 +67,11 @@ type storageImageDestination struct {
 	filenames       map[digest.Digest]string        // Mapping from layer blobsums to names of files we used to hold them
 	SignatureSizes  []int                           `json:"signature-sizes,omitempty"`  // List of sizes of each signature slice
 	SignaturesSizes map[digest.Digest][]int         `json:"signatures-sizes,omitempty"` // Sizes of each manifest's signature slice
+	// fields to support early commit of layers below
+	indexLock         sync.Mutex              // Mutex to synchronize access to below fields
```
mtrmac, line 71: Not `indexToStorageID`, it seems. Please make that clearer, similarly document what exactly `putBlobMutex` covers, and then review the users.

mtrmac: Alternatively, do we even need two different locks if all we are protecting (IIRC) is short non-blocking data accesses?

vrothberg: > Alternatively, do we even need two different locks if all we are protecting (IIRC) is short non-blocking data accesses?

Very good thinking. I'll investigate that idea.

vrothberg: > Not indexToStorageID, it seems. Please make that clearer, similarly document what exactly putBlobMutex covers, and then review the users.

`indexToStorageID` is protected implicitly since only _one_ goroutine can be the "worker" which in turn is guaranteed via the `indexLock`.

mtrmac: You’re right, `indexToStorageID` uses are safe. I meant that the comment “to synchronize access to below fields” was inaccurate WRT that field.

These things should be accurately documented for future readers of the code.
Or, speculatively, maybe something like
```go
type storageImageDestination struct {
    …
    locked struct {
        lock sync.Mutex
        … // THESE fields require the lock to be held, and it’s clearly visible at every use
    }
   …
}
```
? Obviously that works much better with languages with generics that can actually ~enforce that accesses only happen with a lock held, but it might be worth it even here just because comments go so easily out of sync with behavior.

rhatdan:
:+1:

storage/storage_image.go
vrothberg: I will prepare a follow-up PR to improve the comments for future generations :) Thank you, as always, for your very constructive and thoughtful review!

vrothberg: https://github.com/containers/image/pull/1188

PR #1186, opened 2021-03-22
manifest: GuessMIMEType: support OCI artifacts

vrothberg:
If a schema version 2 manifest doesn't include a top-level media type,
default to "guessing" it's an OCI v1 manifest.

Previously, we defaulted to a Docker v2s2 manifest which is causing
issues for OCI artifacts which may ship a custom config media type.
In that case, we guessed it's a Docker image which is wrong.

To harden the code a bit further, add an additional media-type check for
the config to not only check for an ordinary OCI config but also a
Docker one.

Found the issues while prototyping source containers in Buildah. @rhatdan @nalind @giuseppe @lsm5 PTAL

### DIFF
--- a/manifest/manifest.go
+++ b/manifest/manifest.go
@@ -132,9 +132,16 @@ func GuessMIMEType(manifest []byte) string {
 		if err := json.Unmarshal(manifest, &ociMan); err != nil {
 			return ""
 		}
-		if ociMan.Config.MediaType == imgspecv1.MediaTypeImageConfig {
+		switch ociMan.Config.MediaType {
+		case imgspecv1.MediaTypeImageConfig:
 			return imgspecv1.MediaTypeImageManifest
+		case DockerV2Schema2ConfigMediaType:
+			// This case should not happen since a Docker image
+			// must declare a top-level media type and
+			// `meta.MediaType` has already been checked.
+			return DockerV2Schema2MediaType
 		}
+		// Maybe an image index or an OCI artifact.
 		ociIndex := struct {
 			Manifests []imgspecv1.Descriptor `json:"manifests"`
 		}{}
@@ -145,9 +152,13 @@ func GuessMIMEType(manifest []byte) string {
 			if ociMan.Config.MediaType == "" {
 				return imgspecv1.MediaTypeImageIndex
 			}
+			// FIXME: this is mixing media types of manifests and configs.
 			return ociMan.Config.MediaType
 		}
-		return DockerV2Schema2MediaType
+		// It's most likely an OCI artifact with a custom config media
+		// type which is not (and cannot) be covered by the media-type
+		// checks cabove.
+		return imgspecv1.MediaTypeImageManifest
 	}
 	return ""
 }
--- a/oci/layout/oci_src.go
+++ b/oci/layout/oci_src.go
@@ -15,6 +15,7 @@ import (
 	"github.com/opencontainers/go-digest"
 	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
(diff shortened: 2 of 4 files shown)


manifest/manifest.go
```diff
@@ -135,6 +135,10 @@ func GuessMIMEType(manifest []byte) string {
 		if ociMan.Config.MediaType == imgspecv1.MediaTypeImageConfig {
 			return imgspecv1.MediaTypeImageManifest
 		}
+		if ociMan.Config.MediaType == DockerV2Schema2ConfigMediaType {
+			return DockerV2Schema2MediaType
+		}
```
mtrmac, line 140: Please document that this should not happen (`meta.MediaType` should have been set). 

(Use a `switch`?)

manifest/manifest.go
```diff
@@ -147,7 +151,7 @@ func GuessMIMEType(manifest []byte) string {
 			}
 			return ociMan.Config.MediaType
```
mtrmac, line 152: (Just noticed: This doesn’t make sense to me — it is mixing config and manifest media type, and an index should not have a config field. Obviously unrelated.)

manifest/manifest.go
```diff
@@ -147,7 +151,7 @@ func GuessMIMEType(manifest []byte) string {
...
 		}
-		return DockerV2Schema2MediaType
+		return imgspecv1.MediaTypeImageManifest
```
mtrmac, line 154: Please document that we are guessing this is an OCI artifact, not necessarily a container image.

mtrmac (commented):
(Mostly drive-by:)

ACK, more detailed documentation of the heuristic assumptions would be nice.

(This doesn’t get us that much closer to ”supporting OCI” artifacts — we generally assume that an OCI manifest is paired with an OCI config, and other things are probably going to fail, maybe quite silently discarding the artifact config, without a full-scale assumption review. But it’s a necessary step.)

manifest/manifest.go
vrothberg: That also jumped in my eyes but I assumed that I missed the point.  Thanks for highlighting!

lsm5 (approved):
LGTM going by problem description, tests and the diff but would defer to Miloslav :) 

/cc @ashcrow :)

manifest/manifest.go
```diff
@@ -132,9 +132,16 @@ func GuessMIMEType(manifest []byte) string {
 		if err := json.Unmarshal(manifest, &ociMan); err != nil {
 			return ""
 		}
-		if ociMan.Config.MediaType == imgspecv1.MediaTypeImageConfig {
+		switch ociMan.Config.MediaType {
+		case imgspecv1.MediaTypeImageConfig:
 			return imgspecv1.MediaTypeImageManifest
+		case DockerV2Schema2ConfigMediaType:
+			// This case should not happen since a Docker image
+			// must delcare a top-level media type and
```
mtrmac, line 140:
```suggestion
			// must declare a top-level media type and
```

mtrmac (commented):
LGTM. Thanks!

rhatdan:
LGTM

oci/layout/oci_src.go
```diff
@@ -94,6 +95,7 @@ func (s *ociImageSource) GetManifest(ctx context.Context, instanceDigest *digest
 
 	m, err := ioutil.ReadFile(manifestPath)
 	if err != nil {
+		logrus.Errorf("Error HERE")
```
mtrmac, line 98: @vrothberg Could you drop this part, please?

vrothberg: Ouch, thanks for catching!

vrothberg: https://github.com/containers/image/pull/1201

PR #1193, opened 2021-03-30
registries.conf: configure credential helpers

vrothberg:
Allow for configuring credential helpers in `registries.conf` files.

Credential helpers are configured as a top-level field
`credential-helpers`.  This is an array of strings.  Items listed in the
array are consulted in the specified order when looking up or removing
credentials.

Note that there is a built-in credential helpers `containers-auth.json`
for using auth files.  If no global defaults are specified, we will
fallback to using auth files.  This assures backwards compat and a
working default setting.

The traces of the disabled "keyring" functionality have largely been
removed.  If we ever want to re-enable support, we can follow the
example of auth files and specify a new built-in keyring helper.

Using a built-in helper simplifies the code quite a bit since the code
structure boilds down to conditionally dispatching helpers; everything's
a credential helper with some special values for built-in helpers.

Make sure that the execution paths are properly logged (debug level).

### DIFF
--- a/docs/containers-registries.conf.5.md
+++ b/docs/containers-registries.conf.5.md
@@ -16,6 +16,9 @@ Container engines will use the `$HOME/.config/containers/registries.conf` if it 
 `unqualified-search-registries`
 : An array of _host_[`:`_port_] registries to try when pulling an unqualified image, in order.
 
+`credential-helpers`
+: An array of default credential helpers used as external credential stores.  Note that "containers-auth.json" is a reserved value to use auth files as specified in containers-auth.json(5).  The credential helpers are set to `["containers-auth.json"]` if none are specified.
+
 ### NAMESPACED `[[registry]]` SETTINGS
 
 The bulk of the configuration is represented as an array of `[[registry]]`
@@ -270,7 +273,7 @@ the destination registry is unambiguous. Pulling by digest
 tags.
 
 # SEE ALSO
-  containers-certs.d(5)
+ containers-auth.json(5) containers-certs.d(5)
 
 # HISTORY
 Dec 2019, Warning added for unqualified image names by Tom Sweeney <tsweeney@redhat.com>
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -6,14 +6,17 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"runtime"
 	"strings"
 
+	"github.com/containers/image/v5/pkg/sysregistriesv2"
 	"github.com/containers/image/v5/types"
 	"github.com/containers/storage/pkg/homedir"
 	helperclient "github.com/docker/docker-credential-helpers/client"
 	"github.com/docker/docker-credential-helpers/credentials"
+	"github.com/hashicorp/go-multierror"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
@@ -41,12 +44,6 @@ var (
 	dockerLegacyHomePath    = ".dockercfg"
 	nonLinuxAuthFilePath    = filepath.FromSlash(".config/containers/auth.json")
 
-	// Note that the keyring support has been disabled as it was causing
-	// regressions. Before enabling, please revisit TODO(keyring) comments
-	// which need to be addressed if the need remerged to support the
-	// kernel keyring.
-	enableKeyring = false
-
 	// ErrNotLoggedIn is returned for users not logged into a registry
 	// that they are trying to logout of
 	ErrNotLoggedIn = errors.New("not logged in")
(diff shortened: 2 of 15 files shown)


vrothberg:
Do not merge. We need to iron out the default behaviors and fallbacks etc.

vrothberg:
@rhatdan PTAL

rhatdan:
LGTM
@QiWang19 @mtrmac PTAL

pkg/docker/config/config.go
```diff
@@ -170,41 +222,70 @@ func GetCredentials(sys *types.SystemContext, registry string) (types.DockerAuth
...
+	}
+	// If the registry doesn't set credential helpers explicitly, use the
+	// global defaults.
+	if len(helpers) == 0 {
+		helpers, err = sysregistriesv2.GlobalCredentialHelpers(sys)
 		if err != nil {
-			logrus.Debugf("Credentials not found")
 			return types.DockerAuthConfig{}, err
 		}
+	}
```
mtrmac, line 255: At a first glance this looks like something that should be a single call, not something a caller could have the option of getting wrong. (OTOH one caller is already different.)

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -663,6 +688,75 @@ func GetShortNameMode(ctx *types.SystemContext) (types.ShortNameMode, error) {
 	return config.shortNameMode, err
 }
 
+// GlobalCredentialHelpers returns the global top-level credential helpers.
+// A slice with only AuthenticationFileHelper is returned if no helpers are
+// configured.
```
mtrmac, line 693: This should not be an API promise. Document what the caller is allowed to do with the return value, not how it is formed.

pkg/docker/config/config.go
```diff
@@ -227,53 +308,146 @@ func getAuthenticationWithHomeDir(sys *types.SystemContext, registry, homeDir st
...
-			err := removeAllAuthFromKernelKeyring()
-			if err == nil {
-				logrus.Debugf("removing all credentials from kernel keyring")
-				return false, nil
+	// Query for all helpers including the built-in one for auth files.
+	helpers, err := sysregistriesv2.AllCredentialHelpers(sys)
+	if err != nil {
+		return err
+	}
+	globalHelpers, err := sysregistriesv2.GlobalCredentialHelpers(sys)
```
mtrmac, line 389: `AllCredentialHelpers` is already calling this AFAICS.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -663,6 +688,75 @@ func GetShortNameMode(ctx *types.SystemContext) (types.ShortNameMode, error) {
...
+		return reg.CredentialHelpers, nil
+	}
+	return []string{}, nil
+}
+
+// AllCredentialHelpers returns a list of all credential helpers listed either
+// as the top-level credential helpers (see `GlobalCredentialHelpers()` or
+// specific to registry (see `FindCredentialHelpers`).  The top-level
+// credential helpers are listed prior to the registry-specific ones, in the
+// specified order.
```
mtrmac, line 726: Why does the order matter? If the caller doesn’t know the boundary, it isn’t a separation mechanism. Is it intended for some kind of prioritization? What does the order _mean_ for the caller, then? That earlier entries override the latter ones on conflict? The opposite?

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -663,6 +688,75 @@ func GetShortNameMode(ctx *types.SystemContext) (types.ShortNameMode, error) {
 	return config.shortNameMode, err
 }
 
+// GlobalCredentialHelpers returns the global top-level credential helpers.
+// A slice with only AuthenticationFileHelper is returned if no helpers are
+// configured.
+func GlobalCredentialHelpers(sys *types.SystemContext) ([]string, error) {
```
mtrmac, line 694: See elsewhere, does this need to exist?

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -663,6 +688,75 @@ func GetShortNameMode(ctx *types.SystemContext) (types.ShortNameMode, error) {
...
+	}
+	return helpers, nil
+}
+
+// FindCredentialHelpers returns a list of credential helpers for the
+// specified registry.  The helpers are intended to be contacted in the
+// returned order.
+//
+// Note that the global top-level credential helpers are ignored.
+func FindCredentialHelpers(sys *types.SystemContext, registry string) ([]string, error) {
```
mtrmac, line 711: `CredentialHelpersForRegistry`? Not at all sure.

(Or should that be `…ForReference reference.Named`? Probably not, but think about the “multiple credentials for different tenants on `quay.io` problem. OTOH I don’t think we have a _solution_ for that either way, so maybe let’s add a new API only when we have one.)

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -663,6 +688,75 @@ func GetShortNameMode(ctx *types.SystemContext) (types.ShortNameMode, error) {
...
+	}
+	return []string{}, nil
+}
+
+// AllCredentialHelpers returns a list of all credential helpers listed either
+// as the top-level credential helpers (see `GlobalCredentialHelpers()` or
+// specific to registry (see `FindCredentialHelpers`).  The top-level
+// credential helpers are listed prior to the registry-specific ones, in the
+// specified order.
+func AllCredentialHelpers(sys *types.SystemContext) ([]string, error) {
```
mtrmac, line 727: Does this need to exist? It probably does, it just reminds me of the `GetRegistries` conversation in the other PR.

docs/containers-registries.conf.5.md
```diff
@@ -16,6 +16,9 @@ Container engines will use the `$HOME/.config/containers/registries.conf` if it
 `unqualified-search-registries`
 : An array of _host_[`:`_port_] registries to try when pulling an unqualified image, in order.
 
+`credential-helpers`
+: An array of default credential helpers used as external credential stores.  The credential helpers are consulted in the specified order.  The first successful store and lookup suffices.  Note that "containers-auth.json" is a reserved value to use auth files as specified in containers-auth.json(5).
```
mtrmac, line 20: If there is a hard-coded default, it probably should be documented here.

mtrmac (commented):
Just a minimal skim for the API surface.

mtrmac:
> Registry-specific credential helpers are configured as part of a
registry object and allow for overriding the global defaults.

BTW, random idle thought, do we need this? IIRC the global configuration in the config file was originally wanted so that we can system-wide default to a credential helper instead of using plaintext files (but users can revert); that doesn’t imply the need to do it per-registry.

(It might be a good idea to have the per-registry (per-reference) _API_, that doesn’t strictly require supporting per-registry _configuration_. Or, uh, in theory, even the sysregistriesv2 API could be private and we could just support accessing credentials via pkg/docker only.)

vrothberg:
I am cool to keep it simple and only use "global" cred helpers and ditch the registry-specific ones. @rhatdan WDYT?

rhatdan:
I am fine with Global also.  Not sure where the registry specific ones came from (Hopefully there is no message where I requested it.)  Simple is better.

mtrmac:
The flip side is that per-registry helpers are already possible via `.docker/config.json`=`auth.json` , and it feels a bit cumbersome to have a `registries conf` helper list point to a config that points to helpers.

But then again, do we care? Is there a _good_ reason to use a helper for one single registry? (Some kind of integration with a specific cloud provider based on other credentials in the environment, maybe?)

I don’t really have an opinion either way, I just want to make sure we know why we are adding a ~permanent feature before we are committed to maintaining it.

rhatdan:
I would leave that to the auth.json file.  I don't see people using different mechanisms for this.

vrothberg:
Less code to maintain is always good.  I took over the design and assumed there was consensus on it.  I'll clean it up now and repush.

vrothberg:
Done.  Now, there is only 1 top-level field to configure cred helpers.  Greatly simplified the API, the code and and the docs.

TomSweeneyRedHat:
LGTM
and I smell a blog post coming out of this....

rhatdan:
LGTM

PR #1204, opened 2021-04-17
Add SetCredentials which returns human-readable path for configured auth

flouthoc:
Reference Issue: https://github.com/containers/podman/issues/6717
Sister PR to: https://github.com/containers/podman/pull/10068

### DIFF
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -51,11 +51,13 @@ var (
 	ErrNotSupported = errors.New("not supported")
 )
 
-// SetAuthentication stores the username and password in the credential helper or file
-func SetAuthentication(sys *types.SystemContext, registry, username, password string) error {
+// SetCredentials stores the username and password in the credential helper or file
+// and returns path to file or helper name in format (helper:%s)
+func SetCredentials(sys *types.SystemContext, registry, username, password string) (string, error) {
+	path := ""
 	helpers, err := sysregistriesv2.CredentialHelpers(sys)
 	if err != nil {
-		return err
+		return path, err
 	}
 
 	// Make sure to collect all errors.
@@ -65,7 +67,7 @@ func SetAuthentication(sys *types.SystemContext, registry, username, password st
 		switch helper {
 		// Special-case the built-in helpers for auth files.
 		case sysregistriesv2.AuthenticationFileHelper:
-			err = modifyJSON(sys, func(auths *dockerConfigFile) (bool, error) {
+			path, err = modifyJSON(sys, func(auths *dockerConfigFile) (bool, error) {
 				if ch, exists := auths.CredHelpers[registry]; exists {
 					return false, setAuthToCredHelper(ch, registry, username, password)
 				}
@@ -76,7 +78,8 @@ func SetAuthentication(sys *types.SystemContext, registry, username, password st
 			})
 		// External helpers.
 		default:
-			err = setAuthToCredHelper(helper, registry, username, password)
+			path = fmt.Sprintf("credential helper: %s", helper)
+			err  = setAuthToCredHelper(helper, registry, username, password)
 		}
 		if err != nil {
 			multiErr = multierror.Append(multiErr, err)
@@ -84,9 +87,16 @@ func SetAuthentication(sys *types.SystemContext, registry, username, password st
 			continue
 		}
 		logrus.Debugf("Stored credentials for %s in credential helper %s", registry, helper)
-		return nil
+		return path, nil
 	}
-	return multiErr
+	return "", multiErr
+}
+
+
+// SetAuthentication stores the username and password in the credential helper or file
+func SetAuthentication(sys *types.SystemContext, registry, username, password string) error {
+	_, err := SetCredentials(sys, registry, username, password)
+	return err
 }
 
 // GetAllCredentials returns the registry credentials for all registries stored
@@ -322,7 +332,7 @@ func RemoveAuthentication(sys *types.SystemContext, registry string) error {
 		switch helper {
 		// Special-case the built-in helper for auth files.
 		case sysregistriesv2.AuthenticationFileHelper:
-			err = modifyJSON(sys, func(auths *dockerConfigFile) (bool, error) {
+			_, err = modifyJSON(sys, func(auths *dockerConfigFile) (bool, error) {
 				if innerHelper, exists := auths.CredHelpers[registry]; exists {
 					removeFromCredHelper(innerHelper)
 				}
@@ -368,7 +378,7 @@ func RemoveAllAuthentication(sys *types.SystemContext) error {
 		switch helper {
 		// Special-case the built-in helper for auth files.
 		case sysregistriesv2.AuthenticationFileHelper:
-			err = modifyJSON(sys, func(auths *dockerConfigFile) (bool, error) {
+			_, err = modifyJSON(sys, func(auths *dockerConfigFile) (bool, error) {
 				for registry, helper := range auths.CredHelpers {
 					// Helpers in auth files are expected
 					// to exist, so no special treatment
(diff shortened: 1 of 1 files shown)


mtrmac (commented):
I don’t think this is any better than adding a new API `SetCredentials` that returns a human-readable description of the new location, as discussed in https://github.com/containers/podman/issues/6717 .

Either way c/image is adding a new API; and either way we have to deal with the fairly ambiguous semantics of the returned value, in order to retain flexibility for where/how credentials are stored.

With this `ContextVerbosity`:
- The data ends up on stdout, with no way for the caller to control it
- `ContextVerbosity` makes no promises, i.e. the caller has no reason to assume that `SetAuthentication` will use the value for anything — and no reason to assume that `SetAuthentication` will not add much more other value the caller does not care about.

mtrmac:
Also this only handles the JSON case, not any of the cases where a credential helper is used.

flouthoc:
@mtrmac Fair points.Btw I only added descriptive verbose log in `modifyJSON` as starting point we can for sure add it on other places if by any chance we go forward with this approach . But regarding previous points I had few doubts just raising it in original issue https://github.com/containers/podman/issues/6717 to keep discussion at same place.

mtrmac:
> * The data ends up on stdout, with no way for the caller to control it

Notably this is not useful for remote Podman.

rhatdan:
I am fine with this being an API addition.

vrothberg:
Friendly ping, @flouthoc

flouthoc:
@vrothberg Thanks a lot reminding me 🙂 . I was unable to work on this due to reasons specified here https://github.com/containers/podman/pull/10068. Probably i'll open a new PR with relevent API changes as that makes more sense and close this one ? WDYT ?

vrothberg:
Sure, sounds good to me, thank you!

flouthoc:
@vrothberg @mtrmac @rhatdan Could we please re-open this. I have made changes and i think it'll be better if i keep same PR so we wont loose discussions.

rhatdan:
Does not look like this can be re-opend, Not sure why.  Looks like you will need a new PR.

flouthoc:
oh that's sad @rhatdan afaik repo admin could re-open a closed PR.

flouthoc:
@rhatdan Could you please check again if have reverted the commits as per github documentation. `Re-open` should show now.

flouthoc:
@mtrmac @rhatdan Could you guys  please review this again.

rhatdan:
LGTM

rhatdan:
@vrothberg PTAL

pkg/docker/config/config.go
```diff
@@ -51,11 +51,12 @@ var (
 	ErrNotSupported = errors.New("not supported")
 )
 
-// SetAuthentication stores the username and password in the credential helper or file
-func SetAuthentication(sys *types.SystemContext, registry, username, password string) error {
+// SetCredentials stores the username and password in the credential helper or file
```
mtrmac, line 54: Please document what is the return value, consistently with what the function ends up returning (and don’t promise it to be a path).

mtrmac (commented):
Thanks!

ACK overall.

pkg/docker/config/config.go
flouthoc: @mtrmac resolved in latest commit.

pkg/docker/config/config.go
```diff
@@ -76,17 +77,25 @@ func SetAuthentication(sys *types.SystemContext, registry, username, password st
 			})
 		// External helpers.
 		default:
-			err = setAuthToCredHelper(helper, registry, username, password)
+			path = fmt.Sprintf("docker-credential-%s", helper)
```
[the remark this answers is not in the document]
flouthoc: @mtrmac Tried fixing this in lastest commit, I have a suggestion a regular path for file and `helper:%s` in case helper is used.Caller doesn't have a strict contract as of now its used here https://github.com/containers/common/pull/607 but in any case if calle want to check if it was a helper he can split for `helper:`. WDYT ?

pkg/docker/config/config.go
```diff
@@ -76,17 +78,26 @@ func SetAuthentication(sys *types.SystemContext, registry, username, password st
 			})
 		// External helpers.
 		default:
-			err = setAuthToCredHelper(helper, registry, username, password)
+			credHelper := fmt.Sprintf("docker-credential-%s", helper)
```
vrothberg, line 81: I think we should just leave `helper` as is and not prefix `docker-credential`.

pkg/docker/config/config.go
```diff
@@ -76,17 +78,26 @@ func SetAuthentication(sys *types.SystemContext, registry, username, password st
...
+			path = fmt.Sprintf("helper: %s", credHelper)
```
vrothberg, line 82:
```suggestion
			path = fmt.Sprintf("credential helper: %s", credHelper)
```

vrothberg (commented):
Two nits. Other than that LGTM

pkg/docker/config/config.go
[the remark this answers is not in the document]
flouthoc: @mtrmac Added prefix `credential helper: %s` as suggested by @vrothberg.

pkg/docker/config/config.go
flouthoc: @vrothberg Thanks. Made relevant changes. PTAL

flouthoc:
@rhatdan @vrothberg I want to update `images` in containers/common for this PR https://github.com/containers/common/pull/607 to become healthy. Should i update the dependencies manually or is it automatically done by dependabot on a new release cut.

pkg/docker/config/config.go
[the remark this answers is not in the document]
mtrmac: > but in any case if calle want to check if it was a helper he can split for `helper:`. WDYT ?

If we make such a promise, we can’t add any new storage mechanisms (and this would have better been a Go type, not a string for a caller to parse _if_ we wanted to make such a promise).

Note to self: Eventually change this to make no format commitments. The text content is not an API (and doesn’t match the current code right now).

rhatdan:
@flouthoc Let's take @mtrmac advice and create a new credential struct, which has a single function to return the equivalent data that we now return as a string.  Something like:

```
        type CredHelperType int

        const (
        	File CredHelperType = iota
        )
struct CredHelper {
          type CredHelperType
          path string
}

func (c *CredHelper) Path() {
        return c.path
}
```
 func SetCredentials(sys *types.SystemContext, registry, username, password string) (CredHelper, error) {
}

flouthoc:
@rhatdan @mtrmac Sure let me open a PR for the same.

mtrmac:
> @flouthoc Let's take @mtrmac advice and create a new credential struct, which has a single function to return the equivalent data that we now return as a string.

No, I’m sorry, I was unclear. I explicitly _don’t_ want to do this because if we do that, and add some new way to store data in the future (kernel keyring, TPM, whatever), and add another item to the struct, all callers will be silently broken.

This API is only returning “a human readable description” that no caller should ever interpret, and that’s all that `podman login` needs; the API is is going to continue returning “a human readable description” if we add unpredictable future ways to store credentials.

We could build the Go type and an enum and somehow convince callers that they have to anticipate unknown cases and do … what? … in response, but if the only caller we are contemplating were going to only call a `CredHelper.HumanReadableDescription()` to implement `podman login --verbose` in a way that doesn’t break in the future, we just don’t need that struct, the other fields, the API stability commitment, and all that overhead.

I _like_ the current implementation; I’m just pedantic about documenting what it returns and what it promises. I’ll prepare a PR next week or so, to bear the cost of that pedantry personally.

rhatdan:
Ok, I misunderstood your complaint.  SGTM Lets get a release ready.

flouthoc:
@rhatdan Just a friendly reminder, when can we get a release for this.

flouthoc:
@mtrmac @rhatdan Any updates for new release of `image`.

pkg/docker/config/config.go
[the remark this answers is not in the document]
mtrmac: > Note to self: Eventually change this to make no format commitments. 

Filed #1256 .

mtrmac:
@flouthoc A few build-infrastructure PRs need to land first, and probably #1256. But it should happen within a few days, thanks for the reminder.

PR #1208, opened 2021-04-22
Don't try to realpath() on /proc/$pid/fd magic links

cgwalters:
I'm trying to do `skopeo copy docker://quay.io/cgwalters/fcos oci-archive://proc/self/fd/5`
where a higher level process passes the write end of a `pipe()`.

We can't call `realpath` on a pipe, and actually in general I think
programs should avoid running `realpath` and just operate on what
they're passed directly.  There is presumably some reason
we're doing this but it seems like it should work to pass
a pipe, so let's not error out.

### DIFF
--- a/directory/explicitfilepath/path.go
+++ b/directory/explicitfilepath/path.go
@@ -1,17 +1,29 @@
 package explicitfilepath
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/pkg/errors"
 )
 
 // ResolvePathToFullyExplicit returns the input path converted to an absolute, no-symlinks, cleaned up path.
 // To do so, all elements of the input path must exist; as a special case, the final component may be
-// a non-existent name (but not a symlink pointing to a non-existent name)
+// a non-existent name (but not a symlink pointing to a non-existent name, unless it's a /proc/self/fd "magic link").
 // This is intended as a a helper for implementations of types.ImageReference.PolicyConfigurationIdentity etc.
 func ResolvePathToFullyExplicit(path string) (string, error) {
+	// This function gets passed the URL // in addition to a potentially absolute
+	// path, i.e. we trim oci-archive:///path/to/foo.tar to ///path/to/foo.tar,
+	// relying on the kernel to trim the useless // but let's normalize it now.
+	if strings.HasPrefix(path, "///") {
+		path = path[2:]
+	}
+	// We don't resolve magic links, as they could be a pipe or something else.
+	if strings.HasPrefix(path, "/proc/") {
+		return path, nil
+	}
 	switch _, err := os.Lstat(path); {
 	case err == nil:
 		return resolveExistingPathToFullyExplicit(path)


vrothberg (commented):
Thanks, @cgwalters!

Currently it doesn't build (imported and not used: "fmt").

@giuseppe can you have a look? My guts tell me you may know an alternative trick to check for /proc magic.

giuseppe:
there are already few different places where we expect `/proc` to be a procfs so unless we care about procfs also at a different path, I think the patch is fine.
Do we need to care if there are more than 2 additional `/` in the prefix (e.g. `dir://///////////////tmp/foo` is a valid destination)?

directory/explicitfilepath/path.go
```diff
@@ -1,17 +1,29 @@
...
 // ResolvePathToFullyExplicit returns the input path converted to an absolute, no-symlinks, cleaned up path.
 // To do so, all elements of the input path must exist; as a special case, the final component may be
-// a non-existent name (but not a symlink pointing to a non-existent name)
+// a non-existent name (but not a symlink pointing to a non-existent name, unless it's a /proc/self/fd "magic link").
 // This is intended as a a helper for implementations of types.ImageReference.PolicyConfigurationIdentity etc.
 func ResolvePathToFullyExplicit(path string) (string, error) {
+	// This function gets passed the URL // in addition to a potentially absolute
+	// path, i.e. we trim oci-archive:///path/to/foo.tar to ///path/to/foo.tar,
+	// relying on the kernel to trim the useless // but let's normalize it now.
+	if strings.HasPrefix(path, "///") {
```
rhatdan, line 20: Should just loop untile path does not begin with //

mtrmac: Nothing in here are URLs, URL syntax concepts are not applicable.

mtrmac (commented):
Yeah, there is “some reason”: the paths are defined as keys to a policy (compare `ociArchiveReference.PolicyConfiguration{Identity,Namespaces}`), and that doesn’t work if various ways to refer to the same file don’t compare equal. In retrospect unnecessary, I suppose, but not something that we can just ignore. 

I guess this condition makes sense, but it definitely needs  _some_ canonicalization (`filepath.Abs` + `filepath.Clean`) to make the policy work at least enough to be able to write an “anything goes in `/proc/self/fd`” policy or the like. The `/proc` comparison is also only valid after `filepath.Abs`, strictly speaking.

Should this be limited to Linux? It’s clearly incorrect on Windows, not sure about non-Linux Unixes.

mtrmac:
Also, this is a part of the policy enforcement codebase, keep this as close to 100% test coverage as possible, please. (Compare the pre-existing “Coverage:” comments.)

cgwalters:
OK in the end this patch isn't actually useful today.   Let's discuss higher level approach in https://github.com/containers/image/issues/1209 and if it turns out we want containers/image to gracefully handle file descriptors we can revisit this.

cgwalters:
I realized there's a simple workaround: instead of using a `pipe(2)` I can use a temporary `mkfifo(2)`, i.e. by giving the pipe a name it means `realpath()` works (even though it doesn't make sense to resolve its name).

i.e. I can do:
```
$ tempdir=$(mktemp -d)
$ mkfifo ${tempdir}/p
$ skopeo copy docker://quay.io/example/image:latest docker-archive:${tempdir}/p
```

Then another process gets streaming output via `cat ${tempdir}/p`, and I can clean up the fifo afterwards.

PR #1232, opened 2021-05-19
storage: set image names last

nalind:
Set the names on an image last, after we've stored its config blob, manifests, and signatures.  This will keep callers that look up an image and then immediately try to inspect it from getting tripped up if they hit the window between when we set the names and save the data that's used when we inspect an image.

### DIFF
--- a/storage/storage_image.go
+++ b/storage/storage_image.go
@@ -1034,25 +1034,6 @@ func (s *storageImageDestination) Commit(ctx context.Context, unparsedToplevel t
 			return errors.Wrapf(err, "error saving big data %q for image %q", blob.String(), img.ID)
 		}
 	}
-	// Set the reference's name on the image.  We don't need to worry about avoiding duplicate
-	// values because SetNames() will deduplicate the list that we pass to it.
-	if name := s.imageRef.DockerReference(); len(oldNames) > 0 || name != nil {
-		names := []string{}
-		if name != nil {
-			names = append(names, name.String())
-		}
-		if len(oldNames) > 0 {
-			names = append(names, oldNames...)
-		}
-		if err := s.imageRef.transport.store.SetNames(img.ID, names); err != nil {
-			if _, err2 := s.imageRef.transport.store.DeleteImage(img.ID, true); err2 != nil {
-				logrus.Debugf("error deleting incomplete image %q: %v", img.ID, err2)
-			}
-			logrus.Debugf("error setting names %v on image %q: %v", names, img.ID, err)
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@mtrmac @vrothberg PTAL

mtrmac:
LGTM. An ID/digest reference could still see an unsigned image, but this is clearly an improvement.

Definitely not blocking: There was a discussion some months ago about an atomic commit operation instead of these separate writes; I’m afraid I can neither recall the race nor find the discussion now.  That would probably handle this case as well. Is that being considered?

vrothberg (commented):
LGTM

nalind:
> Definitely not blocking: There was a discussion some months ago about an atomic commit operation instead of these separate writes; I’m afraid I can neither recall the race nor find the discussion now. That would probably handle this case as well. Is that being considered?

I think the broad strokes there were centered on adding a slice of data items to save for a just-created image to the ImageOptions structure that we hand to the storage library, so that the storage library could notionally save the various data blobs (for us that's the config, manifests, signatures) at the same time it was creating the image record.  That would remove the "new image record has an ID but no manifest or signatures" window.

We'd still have to set the image's names in a discrete step because the storage library normally reject attempts to create an image if the ID or names are specified and any of them are already being used.  If the image record already existed, we'd have to ensure that everything we would have tried to save during its creation was there, adding the missing items (I'm thinking of manifests here, since a given image ID an have more than one that includes the same config blob), so it wouldn't be a complete solution.

storage/storage_image.go
```diff
@@ -1130,6 +1111,25 @@ func (s *storageImageDestination) Commit(ctx context.Context, unparsedToplevel t
 		}
 		logrus.Debugf("saved image metadata %q", string(metadata))
 	}
+	// Set the reference's name on the image.  We don't need to worry about avoiding duplicate
+	// values because SetNames() will deduplicate the list that we pass to it.
+	if name := s.imageRef.DockerReference(); len(oldNames) > 0 || name != nil {
```
mtrmac, line 1116: FWIW not quite related to this PR:
- AFAICS the `len(oldNames) > 0` condition is unnecessary, we would only be setting names to the current value
- If we are worried about races between the call to `CreateImage` and this point, it might make sense to move the `oldNames` computation (which no longer needs to happen at this line) inside this conditional, to decrease (but not eliminate) a time window during which names added by a concurrent caller would be overwritten by `oldNames + name`, losing the other updates.
- … So, to be fully correct, there probably should be an `AddName` that does the update atomically under the c/storage lock, instead of `SetNames`.

PR #420, opened 2018-02-26
Improve docker/* error messages

mtrmac:
especially focus on those from `client.HandleErrorResponse`, which may be very unhelpful, e.g. `"unknown: Not Found"`.

In general, make sure the error includes an expanded Docker reference, or a full URL, or a hostname + path.

Examples:
```console
$ ./skopeo-old inspect docker://www.example.com/busybox
FATA[0001] pinging docker registry returned: error pinging repository, response code 404 
$ ./skopeo inspect docker://www.example.com/busybox
FATA[0001] pinging docker registry returned: error pinging registry www.example.com, response code 404
```
```console
$ ./skopeo-old --override-os linux --policy default-policy.json copy docker://busybox:latest docker://busybox:thisisnotpermitted
…
FATA[0005] Error writing manifest: Error uploading manifest to /v2/library/busybox/manifests/thisisnotpermitted: errors:
denied: requested access to the resource is denied
unauthorized: authentication required
$ ./skopeo --override-os linux --policy default-policy.json copy docker://busybox:latest docker://busybox:thisisnotpermitted
…
FATA[0005] Error writing manifest: Error uploading manifest thisisnotpermitted to docker.io/library/busybox: errors:
denied: requested access to the resource is denied
unauthorized: authentication required
```
```console
$ ./skopeo-old --override-os linux --policy default-policy.json copy docker://alpine:latest docker://busybox:thisisnotpermitted
…
FATA[0005] Error writing blob: Error initiating layer upload to /v2/library/busybox/blobs/uploads/: errors:
denied: requested access to the resource is denied
unauthorized: authentication required
$ ./skopeo --override-os linux --policy default-policy.json copy docker://alpine:latest docker://busybox:thisisnotpermitted
…
FATA[0005] Error writing blob: Error initiating layer upload to /v2/library/busybox/blobs/uploads/ in registry-1.docker.io: errors:
denied: requested access to the resource is denied
unauthorized: authentication required
```
```console
$ ./skopeo-old inspect docker://busybox:thisdoesnotexist
FATA[0001] manifest unknown: manifest unknown           
$ ./skopeo inspect docker://busybox:thisdoesnotexist
FATA[0001] Error reading manifest thisdoesnotexist in docker.io/library/busybox: manifest unknown: manifest unknown 
```

PR #1240, opened 2021-06-01
Do not prepend Error on each wrapped error message.

rhatdan:
Podman and other tools already add Error: to the front of returned error
message, and this ends up as a stutter.

podman pull fedora.io/fred
Trying to pull fedora.io/fred:latest...
Error: Error initializing image from source docker://fedora.io/fred:latest: invalid character '<' looking for beginning of value

### DIFF
--- a/docker/docker_image_src.go
+++ b/docker/docker_image_src.go
@@ -36,7 +36,7 @@ type dockerImageSource struct {
 func newImageSource(ctx context.Context, sys *types.SystemContext, ref dockerReference) (*dockerImageSource, error) {
 	registry, err := sysregistriesv2.FindRegistry(sys, ref.ref.Name())
 	if err != nil {
-		return nil, errors.Wrapf(err, "error loading registries configuration")
+		return nil, errors.Wrapf(err, "loading registries configuration")
 	}
 	if registry == nil {
 		// No configuration was found for the provided reference, so use the
@@ -196,7 +196,7 @@ func (s *dockerImageSource) fetchManifest(ctx context.Context, tagOrDigest strin
 	logrus.Debugf("Content-Type from manifest GET is %q", res.Header.Get("Content-Type"))
 	defer res.Body.Close()
 	if res.StatusCode != http.StatusOK {
-		return nil, "", errors.Wrapf(registryHTTPResponseToError(res), "Error reading manifest %s in %s", tagOrDigest, s.physicalRef.ref.Name())
+		return nil, "", errors.Wrapf(registryHTTPResponseToError(res), "reading manifest %s in %s", tagOrDigest, s.physicalRef.ref.Name())
 	}
 
 	manblob, err := iolimits.ReadAtMost(res.Body, iolimits.MaxManifestBodySize)
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -92,7 +92,7 @@ func (d *digestingReader) Read(p []byte) (int, error) {
 			// Coverage: This should not happen, the hash.Hash interface requires
 			// d.digest.Write to never return an error, and the io.Writer interface
 			// requires n2 == len(input) if no error is returned.
-			return 0, errors.Wrapf(err, "Error updating digest during verification: %d vs. %d", n2, n)
+			return 0, errors.Wrapf(err, "updating digest during verification: %d vs. %d", n2, n)
 		}
 	}
 	if err == io.EOF {
@@ -240,7 +240,7 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
 
 	dest, err := destRef.NewImageDestination(ctx, options.DestinationCtx)
 	if err != nil {
-		return nil, errors.Wrapf(err, "Error initializing destination %s", transports.ImageName(destRef))
+		return nil, errors.Wrapf(err, "initializing destination %s", transports.ImageName(destRef))
 	}
 	defer func() {
 		if err := dest.Close(); err != nil {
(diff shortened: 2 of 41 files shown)


docker/docker_image_src.go
```diff
@@ -197,7 +197,7 @@ func (s *dockerImageSource) fetchManifest(ctx context.Context, tagOrDigest strin
 	logrus.Debugf("Content-Type from manifest GET is %q", res.Header.Get("Content-Type"))
 	defer res.Body.Close()
 	if res.StatusCode != http.StatusOK {
-		return nil, "", errors.Wrapf(registryHTTPResponseToError(res), "Error reading manifest %s in %s", tagOrDigest, s.physicalRef.ref.Name())
```
mtrmac, line 200: There’s a history and experience to this, see https://github.com/containers/image/pull/420 .

Even without that and focusing _only_ on the `copy` caller, this loses data:
- about which mirror is used
- about which manifest (for a multi-arch image) is failing

It might make sense to make this more intelligent, and add that context only in the cases where that data is not the defaults — but consider the `skopeo inspect` example from #420. This might need a clearer definition about which context is expected to be added by the caller vs. callee.

mtrmac (commented):
There’s way too many other `Wrapf("Error …` instances thought the codebase, and almost all errors start with a capital; changing only `copy/copy.go` is inconsistent. I’m not hard set against changing all of that, but then we should really change _all_ of that (and I guess get a consensus first).

---

I don’t know how I feel about removing the “Error …” part from each of those `Wrapf` calls; it seems to me not fluent / possibly confusing without it, using just the colons as separators, OTOH the text does get long, and the colons-only format is what the Go standard library does (but that’s also a good example of how it can be insufficient: `open latest: no such file or directory` is not all that great for non-programmers, with the `open` jargon and no indication that `latest` is a filename).

Similarly I’m tempted to say that Podman writing a context-less “Error:” is not very useful, and only makes sense precisely because c/image/copy is so verbose about context (consider what would happen with a different library, or in one of the less-verbose c/image cases that “can‘t happen”, just returning something like Go’s `io.ErrUnexpectedEOF`).

I guess I have a very weak preference but not at all a strong opinion about these two questions.

docker/docker_image_src.go
rhatdan: Well we could try to do this at the higher level, where the other stutter happens, but I was not sure if this would remove data for other transports.

mtrmac: That’s why I ended up musing about caller/callee expectations.

Both c/image/copy and c/image/docker added all of that context because each of those sub packages found all of that context helpful. I mean, we _could_ define a `c/image/internal/errors.ManifestReadingError{manifestReference reference.Named}` and have this case return a typed error that the copy code could recognize and process specially (it would be a special case, but a clean and IDE-visible special case) — but at that point thinking about the caller/callee context expectations almost seems easier.

rhatdan: Reverted this back, and will concentrate on removing "error" and "Error" from obvious error messages.

rhatdan:
I have a PR for skopeo that fixes all of the tests

https://github.com/containers/skopeo/pull/1333

How can we move forward?

@mtrmac @vrothberg PTAL

mtrmac:
> but then we should really change all of that (and I guess get a consensus first).

Looking at https://github.com/containers/skopeo/pull/1333/fileshttps://github.com/containers/skopeo/pull/1333/files
> parsing manifest for image: choosing image instance: no image found in manifest list for architecture $arch, variant 

I’m not really _enthusiastic_ about this form.

OTOH I’m not a native speaker, Go linters are becoming more aggressive in enforcing the Go (server-focused?) convention, and I know I am bad at UI, so I’m fine with whatever the majority decides.

rhatdan:
The link above does not work, and I am not sure what you are getting at?

mtrmac:
Thanks, I have updated the link. Just below is what I think of a representative example.

That text is clearly useful _to me_, I can easily enough find the relevant code paths using it. I’m worried that it isn’t nearly as comprehensive to casual users who are not extensively exposed to `errors.Wrapf` and similarly formatted strings as a concept. But, again, I know I’m bad at UX.

rhatdan:
I don't see how "error" or "Error" helps users understanding of something that was clearly an Error.

mtrmac:
It’s not the “this whole line is an error” part, which is probably communicated at the start of the line; it’s the “this context information is separate from that context information on the same line” that capitalization and a separate “Error …” heading might convey a bit better than just a `: `.

mtrmac:
A tie-breaker opinion would help. @vrothberg , anyone else?

Secondarily, the tests are really failing.

mtrmac:
See the non-Skopeo test failures in the GitHub report, or just run `make`.

vrothberg (approved):
I do not have a good mental model in which code paths an "{E,e}rror" prefix may be helpful, so I am cool with merging as is as I find it an improvement.

We can selectively re-add some in the future if we feel it would be helpful.

rhatdan:
@mtrmac this now passes all tests except skopeo tests, and I have a fix for the skopeo tests.  So not sure what next step is.

mtrmac:
Thanks @vrothberg 

---

@rhatdan 

> @mtrmac this now passes all tests except skopeo tests, and I have a fix for the skopeo tests. So not sure what next step is.

- Force-merge this (I can do that if you say when)
- Immediately (rebase and) update https://github.com/containers/skopeo/pull/1333 to drop the `replace` directive and update the primary dependency to a recent commit of c/image
- Hopefully tests in the Skopeo PR pass, merge ASAP because all other PRs are going to be failing test in the meantime.

mtrmac (approved):

PR #1242, opened 2021-06-02
copy: re-enable early commits

vrothberg:
The early committing of blobs has been temporarily disabled in #1205.
There was a regression in deciding whether an empty layer may be
committed to the storage or not [1].

Initially, the early committing considerred a layer to empty (and hence
not commit it to storage) when it matched the shasum of an empty GZIP
layer.  As it turned out, that is not a reliable indicator since some
images may still include these layers in the diff IDs [2].

Re-enable the early committing of blobs by passing down the emptiness
information down by querrying the image's manifest.

[1] https://github.com/containers/image/pull/1205#issuecomment-822459523
[2] https://github.com/containers/image/pull/1205#issuecomment-822458565

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -43,10 +43,6 @@ type digestingReader struct {
 	validationSucceeded bool
 }
 
-// FIXME: disable early layer commits temporarily until a solid solution to
-// address #1205 has been found.
-const enableEarlyCommit = false
-
 var (
 	// ErrDecryptParamsMissing is returned if there is missing decryption parameters
 	ErrDecryptParamsMissing = errors.New("Necessary DecryptParameters not present")
@@ -893,6 +889,18 @@ func (ic *imageCopier) copyLayers(ctx context.Context) error {
 		err      error
 	}
 
+	// The manifest is used to extract the information whether a given
+	// layer is empty.
+	manifestBlob, manifestType, err := ic.src.Manifest(ctx)
+	if err != nil {
+		return err
+	}
+	man, err := manifest.FromBlob(manifestBlob, manifestType)
+	if err != nil {
+		return err
+	}
+	manifestLayerInfos := man.LayerInfos()
+
 	// copyGroup is used to determine if all layers are copied
 	copyGroup := sync.WaitGroup{}
 
@@ -925,7 +933,7 @@ func (ic *imageCopier) copyLayers(ctx context.Context) error {
 				logrus.Debugf("Skipping foreign layer %q copy to %s", cld.destInfo.Digest, ic.c.dest.Reference().Transport().Name())
 			}
 		} else {
-			cld.destInfo, cld.diffID, cld.err = ic.copyLayer(ctx, srcLayer, toEncrypt, pool, index, srcRef)
+			cld.destInfo, cld.diffID, cld.err = ic.copyLayer(ctx, srcLayer, toEncrypt, pool, index, srcRef, manifestLayerInfos[index].EmptyLayer)
 		}
 		data[index] = cld
 	}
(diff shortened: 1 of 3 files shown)


vrothberg:
Buildah test PR: https://github.com/containers/buildah/pull/3273

vrothberg:
@mtrmac, I went with querying the manifest in `copyLayer()` for the emptiness information.  That seemed to be the least intrusive change.

storage/storage_image.go
```diff
@@ -761,10 +761,11 @@ func (s *storageImageDestination) queueOrCommit(ctx context.Context, blob types.
 	}
 
 	for info := s.indexToPulledBlob[index]; info != nil; info = s.indexToPulledBlob[index] {
+		info := s.indexToPulledBlob[index]
```
mtrmac, line 764: When does the make a difference? I can’t think of a case.

mtrmac (commented):
ACK.

I’m not quite happy about parsing all of the data _again_ when it was already available while creating `ic.src.LayerInfos`, but it’s externally-invisible and I’m also not going to work on avoiding that, this week at least. I guess I’ll file an issue to maybe optimize this later if there’s ever extra free time.

storage/storage_image.go
vrothberg: It doesn't make a difference.  Missed that line when cleaning up all the debugging statements.  Thanks for catching!

copy/copy.go
```diff
@@ -1195,10 +1203,9 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, to
 				Cache:         ic.c.blobInfoCache,
 				CanSubstitute: ic.canSubstituteBlobs,
 				SrcRef:        srcRef,
+				EmptyLayer:    emptyLayer,
 			}
-			if enableEarlyCommit {
-				options.LayerIndex = &layerIndex
-			}
+			options.LayerIndex = &layerIndex
```
mtrmac, line 1208: (Absolutely non-blocking: maybe fold this into the initializer above — OTOH this way it’s more consistent with the other use.)

BTW random unrelated notes to self for that some-time-maybe issue:
- `srcRef` does not need to be a parameter passed down the stack; it could be computed once when creating `copier` instead of per every layer
- It _almost_ seems that `copyLayer` and callees could pass around a `PutBlobOptions` instead of a long list of unnamed parameters. Making that would would probably be a bit of churn (the -1/nil difference for `emptyLayer`, probably consolidate `PutBlobOptions` and `TryReusingBlobOptions`) but might pay off in readability.

vrothberg:
Buildah CI is green. Just rebased, so we're ready to go

vrothberg:
@rhatdan PTAL

TomSweeneyRedHat:
Does this by chance address https://bugzilla.redhat.com/show_bug.cgi?id=1966872?

mtrmac:
Is there a known root cause? My vague guess is that if anything, this PR is going to make such problems more likely, not less (before this PR we first download everything to local disk, and only after all downloads succeed, create c/storage layers and an image — which is why that bug is rather puzzling to me; after this PR we are creating c/storage layers as we go, and could leave some around on unexpected download abort).

(To be explicit, there isn’t any _known reason_ why this should break anything, and I don’t think this introduces any fundamentally new situations (image pull has always created layers before the image object, only within a shorter timeframe); it just makes unexpected aborts to more likely to be visible to c/storage and thus more likely to need/cause clean-up.)

mtrmac:
@vrothberg Is the hold label is an oversight?

rhatdan:
LGTM

vrothberg:
> @vrothberg Is the hold label is an oversight?

Yes, thanks for checking.  I added the label while Buildah CI was still running but missed removing it once CI was green.

Issue #1252, opened 2021-06-12
(make validate) silently ignores (go fmt) failures


mtrmac:
E.g. from https://github.com/containers/image/runs/2795305104 , https://cirrus-ci.com/task/5935215819358208 ends with
```
${GOBIN}/git-validation -q -run DCO,short-subject,dangling-whitespace -range $(git merge-base ${DEST_BRANCH:-main} HEAD)..${CIRRUS_CHANGE_IN_REPO}
fatal: Not a valid object name main
make validate
/var/tmp/go/bin/golangci-lint run --build-tags "btrfs_noversion libdm_no_deferred_remove"
flag provided but not defined: -s
usage: go fmt [-n] [-x] [packages]
Run 'go help fmt' for details.
```
Similarly locally, `make validate` fails like that but exits with 0.

mtrmac:
This was fixed in #1255 .

PR #1253, opened 2021-06-12
Fix (make)

mtrmac:
This at least allows `make` not to fail (on non-Linux, and after the master/main rename).

https://github.com/containers/image/issues/1252 still needs addressing.


rhatdan:
LGTM

PR #1255, opened 2021-06-14
Fix missing gofmt and validate runtime environment

cevich:
An error was introduced in
https://github.com/containers/image/pull/1140 that incorrectly changed
the command.  However, it was observed the previous command contained
a hidden bug WRT pipefail being unset by default.  Fix both problems by
ensuring golang tooling is present, and calling out to a shell-script
where the environment can be more carefully controlled and verified.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,11 +84,10 @@ test:
 	@$(GPGME_ENV) GO111MODULE="on" go test $(BUILDFLAGS) -cover ./...
 
 fmt:
-	@go fmt -l -s -w $(SOURCE_DIRS)
+	@gofmt -l -s -w $(SOURCE_DIRS)
 
 validate: lint
-	@GO111MODULE="on" go vet ./...
-	@test -z "$$(go fmt -s -l . | grep -ve '^vendor' | tee /dev/stderr)"
+	@hack/validate.sh
 
 lint:
 	$(GOBIN)/golangci-lint run --build-tags "$(BUILDTAGS)"
--- a/hack/validate.sh
+++ b/hack/validate.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+set -eo pipefail
+
+eval $(go env)
+PATH="$GOPATH/bin:$PATH"
+
+die() { echo "Error: ${1:-No message provided}" > /dev/stderr; exit 1; }
+
+# Always run from the repository root
+cd $(dirname "${BASH_SOURCE[0]}")/../
+
+if [[ -z $(type -P gofmt) ]]; then
+    die "Unable to find 'gofmt' binary in \$PATH: $PATH"
+fi
+
+echo "Executing go vet"
+GO111MODULE="on" go vet ./...
+
+echo "Executing gofmt"
+OUTPUT=$(gofmt -s -l . | sed -e '/^vendor/d')
+if [[ ! -z "$OUTPUT" ]]; then
+    die "Please fix the formatting of the following files:
+$OUTPUT"
+fi
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -38,7 +38,7 @@ env:
     PRIOR_UBUNTU_NAME: "ubuntu-2010"
 
     # Google-cloud VM Images
-    IMAGE_SUFFIX: "c6032583541653504"
+    IMAGE_SUFFIX: "c4635821094469632"
     FEDORA_CACHE_IMAGE_NAME: "fedora-${IMAGE_SUFFIX}"
     PRIOR_FEDORA_CACHE_IMAGE_NAME: "prior-fedora-${IMAGE_SUFFIX}"
     UBUNTU_CACHE_IMAGE_NAME: "ubuntu-${IMAGE_SUFFIX}"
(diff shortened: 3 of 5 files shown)


Makefile
```diff
@@ -84,11 +84,11 @@ test:
 	@$(GPGME_ENV) GO111MODULE="on" go test $(BUILDFLAGS) -cover ./...
 
 fmt:
-	@go fmt -l -s -w $(SOURCE_DIRS)
+	@gofmt -l -s -w $(SOURCE_DIRS)
 
 validate: lint
 	@GO111MODULE="on" go vet ./...
-	@test -z "$$(go fmt -s -l . | grep -ve '^vendor' | tee /dev/stderr)"
+	@test -z "$$(gofmt -s -l . | grep -ve '^vendor' | tee /dev/stderr)"
```
mtrmac, line 91: Thanks!

Not the topic of this PR, but I have also noticed https://github.com/containers/image/issues/1252 . Is there a canonical / elegant way to deal with that? `set  -o pipefail` seems supported by both Bash and Zsh, but is not POSIX.

```sh
full=$(gofmt -s -l . ) && echo "$full" | grep …
```
?

mtrmac:
```
/bin/sh: gofmt: command not found
```
in the `validate` task output.

Makefile
cevich: One option is to move these commands into actual bash scripts where the environment/expectations can be more tightly controlled.  Normally not done for small one-liners like this but it would be easy to maintain.

Another option is to [explicitly set `SHELL` and `.SHELLFLAGS` in the Makefile](https://www.gnu.org/software/make/manual/html_node/Choosing-the-Shell.html).  However that would then apply to EVERY shell command, which could have unintended consequences.

cevich:
:cry: Okay, let's go with a dedicated shell script then, that way we can make sure expectations (like `$GOPATH/bin`) are set properly...

cevich:
Oh!  I see why `gofmt` is "missing".  This is going to take new VM images to fix properly.

https://github.com/containers/automation_images/pull/78

cevich:
[Okay, it's failing as expected now when formatting is bad.](https://cirrus-ci.com/task/5082072491491328?logs=main#L160)

hack/validate.sh
```diff
@@ -0,0 +1,26 @@
...
+if [[ -z $(type -P gofmt) ]]; then
+    die "Unable to find 'gofmt' binary in \$PATH: $PATH"
+fi
+
+echo "Executing go vet"
+GO111MODULE="on" go vet ./...
+
+echo "Executing gofmt"
+# grep exits non-zero when no matches found
+OUTPUT=$(gofmt -s -l . |& egrep -v '^vendor' || true)
```
mtrmac, line 22: This seems to be parsed as (`gofmt -s -l . |& egrep -v '^vendor'`) `||` `true`, and e.g. invalid Gofmt flags don’t cause a failure.

Maybe
```suggestion
OUTPUT=$(gofmt -s -l . |& { egrep -v '^vendor' || true; })
```

mtrmac (commented):
LGTM otherwise.

hack/validate.sh
cevich: Hmmm, I didn't think so, but am not entirely sure now that you mention it :confused: Rather than complicate things more, it may be cleaner to just not use grep here...

cevich: ...there.  Replaced `grep` with `sed`.

mtrmac: I wasn’t sure at all either, so I tested that:
```bash
$ set -o pipefail
$ gofmt -eee -s -l . 2>&1 | { cat; false; } || true; echo $?
flag provided but not defined: -eee
…
0
```
(with the `cat` to avoid a broken pipe failure).

The `sed` solution is inventive. I have learned a new trick.

hack/validate.sh
```diff
@@ -0,0 +1,25 @@
...
+
+if [[ -z $(type -P gofmt) ]]; then
+    die "Unable to find 'gofmt' binary in \$PATH: $PATH"
+fi
+
+echo "Executing go vet"
+GO111MODULE="on" go vet ./...
+
+echo "Executing gofmt"
+OUTPUT=$(gofmt -s -l . |& sed -r -e '/^vendor/d')
```
mtrmac, line 21: One last thing,
```suggestion
OUTPUT=$(gofmt -s -l . 2>&1 | sed -r -e '/^vendor/d')
```
please, macOS ships a very old `bash` (and probably isn’t going to upgrade), and this runs from `make` during ordinary development.

(Does the `-r` do anything here?)

cevich: Oh, didn't realize that.  It's probably not necessary, same with the `-r`...

cevich: ...okay it's simplified now.  I don't have a Mac to test it on though.

mtrmac: Tested this manually (success, invalid Gofmt flags, file not in the correct format), everything is as expected.

mtrmac (approved):
Thanks!

cevich:
> Thanks!

Thank you for noticing the problem and giving useful feedback.  OMG yes, `sed` and `awk` are both often overlooked and super handy tools.  [Books have been written about them](https://www.oreilly.com/library/view/sed-awk/1565922255/).  Ya, that "grep with no match returns non-zero" has bitten me hard before.  Someone aught to add a `--stfu` flag to turn off that behavior :grinning:

PR #1262, opened 2021-06-17
storage: PutBlobWithOptions: fix EmptyLayer logic

vrothberg:
As it turned out in containers/podman/pull/10700,
containers/image@db5fb6e introduced a bug in the put-blob logic where
the empty-layer info was used for all blobs that the worker will commit
to storage.

Fix the bug by correctly storing the empty-layer information for each
blob separately and use that inside the worker.

### DIFF
--- a/storage/storage_image.go
+++ b/storage/storage_image.go
@@ -76,12 +76,12 @@ type storageImageDestination struct {
 	indexToStorageID map[int]*string
 	// All accesses to below data are protected by `lock` which is made
 	// *explicit* in the code.
-	blobDiffIDs         map[digest.Digest]digest.Digest           // Mapping from layer blobsums to their corresponding DiffIDs
-	fileSizes           map[digest.Digest]int64                   // Mapping from layer blobsums to their sizes
-	filenames           map[digest.Digest]string                  // Mapping from layer blobsums to names of files we used to hold them
-	currentIndex        int                                       // The index of the layer to be committed (i.e., lower indices have already been committed)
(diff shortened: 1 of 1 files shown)


vrothberg:
Podman test PR https://github.com/containers/podman/pull/10708

vrothberg:
@mtrmac @rhatdan PTAL

The bug boils down to:
```
func (..., emtpyLayer bool) {
....
for ... := range otherLayers {
    commitAnotherLayer(emptyLayer)
}
}
```

mtrmac (approved):
LGTM. Thanks!

flouthoc:
@mtrmac @vrothberg Could we please cut a new release since 5.13.0 is still not propagated into podman.

mtrmac:
@flouthoc #1263 .

flouthoc:
@mtrmac Thanks :smiley:  I guess we would also need to bump `containers/common` with propagated `images` as vendor.

mtrmac:
In principle, but as long as the top-level consumers use the fixed version, the dependency of the library doesn’t quite matter. Dependabot will pick the update eventually in c/common, but I don’t think it’s urgent.

flouthoc:
@mtrmac ah yes my bad.

vrothberg:
I think it's perfectly fine to vendor commits (in contrast to a tag/version) during development.  So we can do a `go get github.com/containers/image/v5/@commit` if we need a specific new feature or fix in the consumers of c/image. We just need to make sure that we have a release once Podman, Buildah and co. get a new release.

PR #1264, opened 2021-06-19
Add override of SignaturePolicyPath

basilgello:
Using podman with paths overridden by various CONTAINERS_* environment
variables still requires 'policy.json' to be present in '/etc/containers'
for root and in '$HOME/.config/containers' for rootless set-ups.

This commit adds the environment variable 'CONTAINERS_POLICY_JSON' to
override this path like all other similar paths.

### DIFF
--- a/signature/policy_config.go
+++ b/signature/policy_config.go
@@ -65,6 +65,11 @@ func defaultPolicyPath(sys *types.SystemContext) string {
 // defaultPolicyPathWithHomeDir is an internal implementation detail of defaultPolicyPath,
 // it exists only to allow testing it with an artificial home directory.
 func defaultPolicyPathWithHomeDir(sys *types.SystemContext, homeDir string) string {
+	if envSigPath, ok := os.LookupEnv("CONTAINERS_POLICY_JSON"); ok {
+		if _, err := os.Stat(envSigPath); err == nil {
+			return envSigPath
+		}
+	}
 	if sys != nil && sys.SignaturePolicyPath != "" {
 		return sys.SignaturePolicyPath
 	}


mtrmac (changes requested):
Thanks for the PR.

Using environment variables is problematic for exactly the same reason it can be useful: it changes behavior without the immediate caller being aware.

It’s absolutely undesirable for an environment variable to override the _explicitly caller-set_ `sys.SignaturePolicyPath` (e.g. if the caller has specifically created a policy for this single operation); this PR _must not_ be merged with that behavior.

Even assuming that were fixed, this doesn’t look attractive to me; it is completely invisible to an auditor of the c/image caller that relies on signature verification. Most other paths used by c/image are in fact _not_ environment-variable controlled, we only accept pre-existing variables like `DOCKER_CONFIG` and wide conventions like `XDG_RUNTIME_DIR`.

Callers of c/image can of course define and use environment variables (and some share their choices in containers/common).

basilgello:
Thanks for explanation! I guess then it is best to shoot one commit in podman instead adding environment variable support to each caller path of this function?﻿

mtrmac:
> Thanks for explanation! I guess then it is best to shoot one commit in podman instead adding environment variable support to each caller path of this function?

Well, the code creating the _default_ values of `types.SystemContext` should be mostly centralized, and AFAICS, at least in this respect, that’s what the code intends to do, but I didn’t review all Podman code paths to see whether it is already consistently done.

basilgello:
> it is completely invisible to an auditor of the c/image caller that relies on signature verification

That's why I decided to try adding constructs like this to `cmd/podman/*`: https://github.com/basilgello/podman/commit/6a3657a6d5de6081d3fed499ac1417ca70cca5f8 It will log whenever `SignaturePolicy` gets overridden with the value from environment variable + it will not try to override the value set explicitly.

mtrmac:
> That's why I decided to try adding constructs like this to `cmd/podman/*`: [basilgello/podman@6a3657a](https://github.com/basilgello/podman/commit/6a3657a6d5de6081d3fed499ac1417ca70cca5f8) It will log whenever `SignaturePolicy` gets overridden with the value from environment variable + it will not try to override the value set explicitly.

(This is probably best discussed in a Podman PR); this implementation is contrary to the “should be mostly centralized” concept, and it will be hard to keep it consistent as new commands are added or as any individual command is edited.

I was looking at `containers/common/pkg/config.DefaultConfig`; as mentioned above, I haven’t actually read all of that code to confirm it is used everywhere, but I would expect Podman to have some centralized point of that kind.

PR #1266, opened 2021-06-19
RFC: Simplify digest references to manifest lists in c/storage

mtrmac:
- If the digest references a single-platform manifest, all matching images will have the same manifest, same config,
  and same behavior of the dummy index created by `imageMatchesSystemContext`; so don't build that dummy index and  rename the function to `multiArchImageMatchesSystemContext` .
- If the digest references a multi-platform list, just check whether `ChooseInstance` would chose the current image; don’t check _both_ `ChooseInstance` based on the multi-platform list and config values through the dummy index. The original pull did both (`ChooseInstance` to choose a per-platform image, and `copy.checkImageDestinationForCurrentRuntime()`, so they are expected to match (and we don't do a config check vs. the current `SystemContext` to refuse handling single-platform images in c/storage, in general).
- Even more verbosely document the situation, explaining how we can end up with multiple images that match the same digest, and what we want to achieve.

Cc: @nalind . Is there a pre-existing set of tests that cover this behavior?

### DIFF
--- a/storage/storage_reference.go
+++ b/storage/storage_reference.go
@@ -11,7 +11,6 @@ import (
 	"github.com/containers/image/v5/types"
 	"github.com/containers/storage"
 	digest "github.com/opencontainers/go-digest"
-	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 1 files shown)


mtrmac:
@vrothberg you mentioned you are working in this area.

vrothberg (approved):
LGTM

vrothberg (commented):
LGTM, feel free to merge

rhatdan:
LGTM

PR #1267, opened 2021-06-19
RFC: Consolidate deleting a partial c/storage image on Commit failure

mtrmac:
Do it from `defer`, so that any new error paths (and quite a few pre-existing ones that missed this) delete the image without
having to specifically care.  Also makes the code a bit shorter.

Note also the added comment; is deleting the image always the right thing to do?

Cc: @nalind . Looking at the existing code paths that didn’t clean up the image, it kind of looks as if failures to write to c/storage should cause the image to be deleted, but other failures (e.g. error computing a digest) should not delete it — but nothing documents such an intent and the code is not consistently behaving that way either.

### DIFF
--- a/storage/storage_image.go
+++ b/storage/storage_image.go
@@ -1011,6 +1011,18 @@ func (s *storageImageDestination) Commit(ctx context.Context, unparsedToplevel t
 	} else {
 		logrus.Debugf("created new image ID %q", img.ID)
 	}
+
+	// Clean up the unfinished image on any error.
+	// (Is this the right thing to do if the image has existed before?)
+	commitSucceeded := false
+	defer func() {
+		if !commitSucceeded {
+			if _, err := s.imageRef.transport.store.DeleteImage(img.ID, true); err != nil {
+				logrus.Debugf("error deleting incomplete image %q: %v", img.ID, err)
+			}
+		}
+	}()
+
 	// Add the non-layer blobs as data items.  Since we only share layers, they should all be in files, so
 	// we just need to screen out the ones that are actually layers to get the list of non-layers.
 	dataBlobs := make(map[digest.Digest]struct{})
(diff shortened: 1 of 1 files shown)


storage/storage_image.go
```diff
@@ -1011,6 +1011,18 @@ func (s *storageImageDestination) Commit(ctx context.Context, unparsedToplevel t
 	} else {
 		logrus.Debugf("created new image ID %q", img.ID)
 	}
+
+	// Clean up the unfinished image on any error.
+	// (Is this the right thing to do if the image has existed before?)
+	commitSucceeded := false
+	defer func() {
+		if !commitSucceeded {
```
vrothberg, line 1019: I feel quite tempted to do a `logrus.Errorf()` here.

If the image couldn't be committed, leaving an explicit and obvious breadcrumb to the user may save some debugging time.

vrothberg (commented):
I think deleting the image, even if it existed before, is a good thing since the image may have been corrupted before or after.

storage/storage_image.go
mtrmac: Yes, a failure to delete an image suggests storage problems that should be reported to the user.

I have changed this as a separate commit, to differentiate from the cleanup.

vrothberg: Sorry, I was vague before.  I would welcome a log indicating that the commit has failed and that the image is going to be deleted from the storage.

storage/storage_image.go
```diff
@@ -1011,6 +1011,18 @@ func (s *storageImageDestination) Commit(ctx context.Context, unparsedToplevel t
...
+			if _, err := s.imageRef.transport.store.DeleteImage(img.ID, true); err != nil {
+				logrus.Errorf("error deleting incomplete image %q: %v", img.ID, err)
```
vrothberg, line 1021:
```suggestion
				logrus.Errorf("Error deleting incomplete image %q: %v", img.ID, err)
```

storage/storage_image.go
vrothberg: If the image existed before an unsuccessful pull, some users may worry why it has been deleted.  Tracking that down can be time costly.

mtrmac: Sure, added a log line in that case, including the old names that are now untagged. (I haven’t tested this manually, though.)

vrothberg (commented):
LGTM

PR #1273, opened 2021-06-25
pkg/cli/environment: UpdateRegistriesConf

vrothberg:
Add a function that reads the env variables that Podman and Buildah are
sensitive to to set the registries conf path in the provided system
context, unless already set.

The motivation to move the functionality here is to consolidate
duplicate code across various places in Podman.

@mtrmac WDYT?

### DIFF
--- a/pkg/cli/environment/environment.go
+++ b/pkg/cli/environment/environment.go
@@ -1,0 +1,31 @@
+package environment
+
+import (
+	"errors"
+	"os"
+
+	"github.com/containers/image/v5/types"
+)
+
+// UpdateRegistriesConf sets the SystemRegistriesConfPath in the system
+// context, unless already set.  Possible values are, in priority and only if
+// set, the CONTAINERS_REGISTRIES_CONF or REGISTRIES_CONFIG_PATH environment
+// variable.
+func UpdateRegistriesConf(sys *types.SystemContext) error {
+	if sys == nil {
+		return errors.New("internal error: UpdateRegistriesConf: nil argument")
+	}
+	if sys.SystemRegistriesConfPath != "" {
+		return nil
+	}
+	if envOverride, ok := os.LookupEnv("CONTAINERS_REGISTRIES_CONF"); ok {
+		sys.SystemRegistriesConfPath = envOverride
+		return nil
+	}
+	if envOverride, ok := os.LookupEnv("REGISTRIES_CONFIG_PATH"); ok {
+		sys.SystemRegistriesConfPath = envOverride
+		return nil
+	}
+
+	return nil
+}


rhatdan:
LGTM

mtrmac (commented):
Compare https://github.com/containers/image/pull/1264 and what c/common does with --authfile and whatever else…

I’d like to stop and think for a bit about what we want the full picture to look like, before writing code and designing this one option at a time.

A half-baked conversation starter:

TL;DR: `pkg/cli` (`pkg/cli/environment`?) with `func SystemContextFromEnvironment` (`func UpdateSystemContextFromEnvironment`?), and maybe eventually more shared flag handling.

---

Historically we started with fully-integrated Skopeo (`skopeo inspect` only, at first), with global CLI flags.

That got split into c/image+Skopeo, with a `types.SystemContext` directly mirroring the CLI flags, the easiest approach to handle the ImageTransport abstraction + global CLI flags in a way that the CLI doesn’t need to know which transport is being initialized to support options for it.

(Similarly `ImageReference`, with its to/from string conversion and no state, has design driven by CLI + `policy.json`, and makes it hard to represent WIP docker-archive and the like.)

`types.SystemContext` is growing all the time, accumulating tons of options, most of which many individual callers don’t care about, all because of that CLI relationship.

If there weren’t a CLI, we probably wouldn’t design it this way: there would be separate option structures for the registry client, for `dir:`, for short names, …, in quite a few cases aggregated at a level (e.g. “platform choice”, or `BigFilesTemporaryDir` could be a struct that is referenced from various other structs). And if one built a CLI on top, there would be a separate “flat CLI flags → individual option structures” conversion layer (to a small extent Skopeo does create that that, with its `dockerImageOptions` etc. — and then flattens that back to `SystemContext`; I think libimage does something similar in places).

---

So where I’m going with this is that _ideally_ it would be nice to get rid of `SystemContext` (which isn’t going to happen without some extraordinary motivation to replace the full API) and replace it with granular options, and then to have a “CLI utility” layer that can be potentially shared across CLI utilities, dealing with flags and documented environment variables, and that is not used at all by non-CLI callers.

We aren’t going to get rid of `SystemContext`, but that “CLI utility” layer still seems a bit closer to a cleaner design than opt-in “update `SystemContext` with $this” all over the place. We can still let the caller opt in to only handle specific $this, but having all of these “options from CLI flags / environment” helpers in a single subpackage would make them all much easier to find.

vrothberg:
> TL;DR: pkg/cli (pkg/cli/environment?) with func SystemContextFromEnvironment (func UpdateSystemContextFromEnvironment?), and maybe eventually more shared flag handling.

Sounds good to me.  I also agree with all other points.

Some of these env variables were introduced for easier testing in CI, eventually used by user and then more were added on top. Unfortunately, not with c/image and the other tools in mind.

While I'd love to consolidate even further, as you suggested, I don't know when we'll find time for that.

I'd still like to get the PR (in one form or another) merged to consolidate at least all users.  c/common/libimage did already some of that but it cannot be called directly from `cmd/podman` as we want to keep the binary size of the remote client small; c/storage and others pull in quite some clutter.

mtrmac:
> Some of these env variables were introduced for easier testing in CI

(Pedantically, CI tools have little business showing up in the final product. Build a separate podman-for-CI binary with those overrides enabled? I acknowledge that it is probably not worth the effort and risk of testing the wrong thing… just highlighting the thought.)


> I'd still like to get the PR (in one form or another) merged to consolidate at least all users.

How about `c/image/pkg/cli/environment/UpdateRegistriesConf(types.SystemContext)` or something like that, then? Exactly the current implementation, let’s just carve out a new place for it. We can then maybe consolidate more of the environment variables in that package.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -60,6 +60,27 @@ var userRegistriesFile = filepath.FromSlash(".config/containers/registries.conf"
 // userRegistriesDir is the path to the per user registry configuration file.
 var userRegistriesDir = filepath.FromSlash(".config/containers/registries.conf.d")
 
+// SetRegistriesConfPath sets the SystemRegistriesConfPath in the system
+// context, unless already set.  Possible values are, in priority and only if
+// set, the CONTAINERS_REGISTRIES_CONF or REGISTRIES_CONFIG_PATH environment
+// variable.
+func SetRegistriesConfPath(sys *types.SystemContext) {
+	if sys == nil {
```
mtrmac, line 68: Maybe this should just crash — if the caller intends environment to be used, passing `nil` and causing the data to be ignored is clearly a programming error.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -60,6 +60,27 @@ var userRegistriesFile = filepath.FromSlash(".config/containers/registries.conf"
 // userRegistriesDir is the path to the per user registry configuration file.
 var userRegistriesDir = filepath.FromSlash(".config/containers/registries.conf.d")
 
+// SetRegistriesConfPath sets the SystemRegistriesConfPath in the system
+// context, unless already set.  Possible values are, in priority and only if
+// set, the CONTAINERS_REGISTRIES_CONF or REGISTRIES_CONFIG_PATH environment
+// variable.
+func SetRegistriesConfPath(sys *types.SystemContext) {
```
mtrmac, line 67: `error` return to support (optional, future, not currently implemented) validation like pointing at nonexistent or inaccessible paths?

vrothberg:
@mtrmac, updated accordingly. Turned the `nil` check into an error.

mtrmac (approved):
Thanks!

LGTM; feel free to merge or wait for other feedback, up to you.

Issue #1276, opened 2021-06-25
Support pulling from different accounts on same registry


saschagrunert:
This is the tracking issue for the feature implementation.

#### Summary

- Change the auth.json parsing behavior to allow full URL paths
- Update [the documentation](https://github.com/containers/image/blob/main/docs/containers-auth.json.5.md) to reflect the changed behavior

mtrmac:
Cross-linking https://issues.redhat.com/browse/RUN-1246 .

Do we really want URLs? I’d prefer `host[:port]/ns/…repo` , both because those values are not HTTP-reachable URLs and because it is clearly conflicting with the legacy format of `.dockercfg`.

mtrmac:
#1278 was merged.

PR #1278, opened 2021-06-28
Add support for registry paths in auth.json

saschagrunert:
This patch adds support for `host[:port]/ns/…repo` to auth.json while keeping the backwards compatible behavior.

Refers to https://github.com/containers/image/issues/1276

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -213,8 +213,7 @@ func dockerCertDir(sys *types.SystemContext, hostPort string) (string, error) {
 // “write” specifies whether the client will be used for "write" access (in particular passed to lookaside.go:toplevelFromSection)
 // signatureBase is always set in the return value
 func newDockerClientFromRef(sys *types.SystemContext, ref dockerReference, write bool, actions string) (*dockerClient, error) {
-	registry := reference.Domain(ref.ref)
-	auth, err := config.GetCredentials(sys, registry)
+	auth, err := config.GetCredentialsForRef(sys, ref.ref)
 	if err != nil {
 		return nil, errors.Wrapf(err, "error getting username and password")
 	}
@@ -224,6 +223,7 @@ func newDockerClientFromRef(sys *types.SystemContext, ref dockerReference, write
 		return nil, err
 	}
 
+	registry := reference.Domain(ref.ref)
 	client, err := newDockerClient(sys, registry, ref.ref.Name())
 	if err != nil {
 		return nil, err
@@ -343,7 +343,8 @@ func SearchRegistry(ctx context.Context, sys *types.SystemContext, registry, ima
 	v1Res := &V1Results{}
 
 	// Get credentials from authfile for the underlying hostname
-	auth, err := config.GetCredentials(sys, registry)
+	// lint:ignore SA1019 We can't use GetCredentialsForRef because we want to search the whole registry.
+	auth, err := config.GetCredentials(sys, registry) // nolint:staticcheck // https://github.com/golangci/golangci-lint/issues/741
 	if err != nil {
 		return nil, errors.Wrapf(err, "error getting username and password")
 	}
--- a/docs/containers-auth.json.5.md
+++ b/docs/containers-auth.json.5.md
@@ -23,7 +23,20 @@ user to container image registries.  The file can have zero to many entries and
 is created by a `login` command from a container tool such as `podman login`,
 `buildah login` or `skopeo login`.  Each entry includes the name of the registry and then an auth
 token in the form of a base64 encoded string from the concatenation of the
-username, a colon, and the password.
+username, a colon, and the password. The registry name can additionally contain
+a path or repository name (an image name without tag or digest). The path (or
+namespace) is matched in its hierarchical order when checking for available
+authentications. For example, an image pull for
+`my-registry.local/namespace/user/image:latest` will result in a lookup in
+`auth.json` in the following order:
+
+- `my-registry.local/namespace/user/image`
+- `my-registry.local/namespace/user`
+- `my-registry.local/namespace`
+- `my-registry.local`
+
+This way it is possible to setup multiple credentials for a single registry
+which can be distinguished by their path.
 
 The following example shows the values found in auth.json after the user logged in to
 their accounts on quay.io and docker.io:
@@ -37,6 +50,25 @@ their accounts on quay.io and docker.io:
 		"quay.io": {
 			"auth": "juQAqGmz5eR1ipzx8Evn6KGdw8fEa1w5MWczmgY="
 		}
+	}
+}
+```
+
+This example demonstrates how to use multiple paths for a single registry, while
+preserving a fallback for `my-registry.local`:
+
+```
+{
+	"auths": {
+		"my-registry.local/foo/bar/image": {
+			"auth": "…"
+		},
+		"my-registry.local/foo": {
+			"auth": "…"
+		},
+		"my-registry.local": {
+			"auth": "…"
+		},
 	}
 }
 ```
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -11,6 +11,7 @@ import (
 	"runtime"
 	"strings"
 
+	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/image/v5/pkg/sysregistriesv2"
 	"github.com/containers/image/v5/types"
 	"github.com/containers/storage/pkg/homedir"
@@ -215,13 +216,34 @@ func getAuthFilePaths(sys *types.SystemContext, homeDir string) []authPath {
 // helpers with falling back to using either auth.json
 // file or .docker/config.json, including support for OAuth2 and IdentityToken.
 // If an entry is not found, an empty struct is returned.
+//
+// Deprecated: GetCredentialsForRef should be used in favor of this API
+// because it allows different credentials for different repositories on the
+// same registry.
 func GetCredentials(sys *types.SystemContext, registry string) (types.DockerAuthConfig, error) {
-	return getCredentialsWithHomeDir(sys, registry, homedir.Get())
+	return getCredentialsWithHomeDir(sys, nil, registry, homedir.Get())
 }
 
-// getCredentialsWithHomeDir is an internal implementation detail of GetCredentials,
-// it exists only to allow testing it with an artificial home directory.
-func getCredentialsWithHomeDir(sys *types.SystemContext, registry, homeDir string) (types.DockerAuthConfig, error) {
+// GetCredentialsForRef returns the registry credentials necessary for
+// accessing ref on the registry ref points to,
+// appropriate for sys and the users’ configuration.
+// If an entry is not found, an empty struct is returned.
+func GetCredentialsForRef(sys *types.SystemContext, ref reference.Named) (types.DockerAuthConfig, error) {
+	return getCredentialsWithHomeDir(sys, ref, reference.Domain(ref), homedir.Get())
+}
+
+// getCredentialsWithHomeDir is an internal implementation detail of
+// GetCredentialsForRef and GetCredentials. It exists only to allow testing it
+// with an artificial home directory.
+func getCredentialsWithHomeDir(sys *types.SystemContext, ref reference.Named, registry, homeDir string) (types.DockerAuthConfig, error) {
+	// consistency check of the ref and registry arguments
+	if ref != nil && reference.Domain(ref) != registry {
+		return types.DockerAuthConfig{}, errors.Errorf(
+			"internal error: provided reference domain %q name does not match registry %q",
+			reference.Domain(ref), registry,
+		)
+	}
+
 	if sys != nil && sys.DockerAuthConfig != nil {
 		logrus.Debugf("Returning credentials for %s from DockerAuthConfig", registry)
 		return *sys.DockerAuthConfig, nil
@@ -230,7 +252,7 @@ func getCredentialsWithHomeDir(sys *types.SystemContext, registry, homeDir strin
 	// Anonymous function to query credentials from auth files.
 	getCredentialsFromAuthFiles := func() (types.DockerAuthConfig, error) {
 		for _, path := range getAuthFilePaths(sys, homeDir) {
-			authConfig, err := findAuthentication(registry, path.path, path.legacyFormat)
+			authConfig, err := findAuthentication(ref, registry, path.path, path.legacyFormat)
 			if err != nil {
 				return types.DockerAuthConfig{}, err
 			}
@@ -294,7 +316,7 @@ func GetAuthentication(sys *types.SystemContext, registry string) (string, strin
 // getAuthenticationWithHomeDir is an internal implementation detail of GetAuthentication,
 // it exists only to allow testing it with an artificial home directory.
 func getAuthenticationWithHomeDir(sys *types.SystemContext, registry, homeDir string) (string, string, error) {
-	auth, err := getCredentialsWithHomeDir(sys, registry, homeDir)
+	auth, err := getCredentialsWithHomeDir(sys, nil, registry, homeDir)
 	if err != nil {
 		return "", "", err
 	}
@@ -575,8 +597,10 @@ func deleteAuthFromCredHelper(credHelper, registry string) error {
 	return helperclient.Erase(p, registry)
 }
 
-// findAuthentication looks for auth of registry in path
-func findAuthentication(registry, path string, legacyFormat bool) (types.DockerAuthConfig, error) {
+// findAuthentication looks for auth of registry in path. If ref is
+// not nil, then it will be taken into account when looking up the
+// authentication credentials.
+func findAuthentication(ref reference.Named, registry, path string, legacyFormat bool) (types.DockerAuthConfig, error) {
 	auths, err := readJSONFile(path, legacyFormat)
 	if err != nil {
 		return types.DockerAuthConfig{}, errors.Wrapf(err, "error reading JSON file %q", path)
@@ -587,12 +611,32 @@ func findAuthentication(registry, path string, legacyFormat bool) (types.DockerA
 		return getAuthFromCredHelper(ch, registry)
 	}
 
-	// I'm feeling lucky
-	if val, exists := auths.AuthConfigs[registry]; exists {
-		return decodeDockerAuth(val)
+	// Support for different paths in auth.
+	// (This is not a feature of ~/.docker/config.json; we support it even for
+	// those files as an extension.)
+	var keys []string
+	if !legacyFormat && ref != nil {
+		keys = authKeysForRef(ref)
+	} else {
+		keys = []string{registry}
+	}
+
+	// Repo or namespace keys are only supported as exact matches. For registry
+	// keys we prefer exact matches as well.
+	for _, key := range keys {
+		if val, exists := auths.AuthConfigs[key]; exists {
+			return decodeDockerAuth(val)
+		}
 	}
 
 	// bad luck; let's normalize the entries first
+	// This primarily happens for legacyFormat, which for a time used API URLs
+	// (http[s:]//…/v1/) as keys.
+	// Secondarily, (docker login) accepted URLs with no normalization for
+	// several years, and matched registry hostnames against that, so support
+	// those entries even in non-legacyFormat ~/.docker/config.json.
+	// The docker.io registry still uses the /v1/ key with a special host name,
+	// so account for that as well.
 	registry = normalizeRegistry(registry)
 	normalizedAuths := map[string]dockerAuthConfig{}
 	for k, v := range auths.AuthConfigs {
@@ -604,6 +648,28 @@ func findAuthentication(registry, path string, legacyFormat bool) (types.DockerA
 	}
 
 	return types.DockerAuthConfig{}, nil
+}
+
+// authKeysForRef returns the valid paths for a provided reference. For example,
+// when given a reference "quay.io/repo/ns/image:tag", then it would return
+// - quay.io/repo/ns/image
+// - quay.io/repo/ns
+// - quay.io/repo
+// - quay.io
+func authKeysForRef(ref reference.Named) (res []string) {
+	name := ref.Name()
+
+	for {
+		res = append(res, name)
+
+		lastSlash := strings.LastIndex(name, "/")
+		if lastSlash == -1 {
+			break
+		}
+		name = name[:lastSlash]
+	}
+
+	return res
 }
 
 // decodeDockerAuth decodes the username and password, which is
--- a/pkg/docker/config/config_test.go
+++ b/pkg/docker/config/config_test.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/image/v5/types"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -134,16 +135,19 @@ func TestGetAuth(t *testing.T) {
 
 	for _, configPath := range configPaths {
 		for _, tc := range []struct {
-			name          string
-			hostname      string
-			path          string
-			expected      types.DockerAuthConfig
-			expectedError error
-			sys           *types.SystemContext
+			name            string
+			ref             string
+			hostname        string
+			path            string
+			expected        types.DockerAuthConfig
+			expectedError   error
+			sys             *types.SystemContext
+			testPreviousAPI bool
 		}{
 			{
-				name:     "no auth config",
-				hostname: "index.docker.io",
+				name:            "no auth config",
+				hostname:        "index.docker.io",
+				testPreviousAPI: true,
 			},
 			{
 				name: "empty hostname",
@@ -157,11 +161,13 @@ func TestGetAuth(t *testing.T) {
 					Username: "example",
 					Password: "org",
 				},
+				testPreviousAPI: true,
 			},
 			{
-				name:     "match none",
-				hostname: "registry.example.org",
-				path:     filepath.Join("testdata", "example.json"),
+				name:            "match none",
+				hostname:        "registry.example.org",
+				path:            filepath.Join("testdata", "example.json"),
+				testPreviousAPI: true,
 			},
 			{
 				name:     "match docker.io",
@@ -171,6 +177,7 @@ func TestGetAuth(t *testing.T) {
 					Username: "docker",
 					Password: "io",
 				},
+				testPreviousAPI: true,
 			},
 			{
 				name:     "match docker.io normalized",
@@ -180,6 +187,7 @@ func TestGetAuth(t *testing.T) {
 					Username: "index",
 					Password: "docker.io",
 				},
+				testPreviousAPI: true,
 			},
 			{
 				name:     "normalize registry",
@@ -189,6 +197,7 @@ func TestGetAuth(t *testing.T) {
 					Username: "example",
 					Password: "org",
 				},
+				testPreviousAPI: true,
 			},
 			{
 				name:     "match localhost",
@@ -198,6 +207,7 @@ func TestGetAuth(t *testing.T) {
 					Username: "local",
 					Password: "host",
 				},
+				testPreviousAPI: true,
 			},
 			{
 				name:     "match ip",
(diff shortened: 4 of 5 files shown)


docker/docker_client.go
```diff
@@ -214,7 +214,9 @@ func dockerCertDir(sys *types.SystemContext, hostPort string) (string, error) {
 // signatureBase is always set in the return value
 func newDockerClientFromRef(sys *types.SystemContext, ref dockerReference, write bool, actions string) (*dockerClient, error) {
 	registry := reference.Domain(ref.ref)
-	auth, err := config.GetCredentials(sys, registry)
+	refPath := reference.Path(ref.ref)
+
+	auth, err := config.GetCredentialsWithRefPath(sys, registry, refPath)
```
vrothberg, line 219: I think it would be more idiomatic to pass a `reference.Reference` around than splitting/joining.

pkg/docker/config/config.go
```diff
@@ -216,12 +216,25 @@ func getAuthFilePaths(sys *types.SystemContext, homeDir string) []authPath {
...
+	return getCredentialsWithHomeDir(sys, registry, "", homedir.Get())
+}
+
+// GetCredentialsWithRefPath returns the registry credentials stored in the
+// registry-specific credential helpers or in the default global credentials
+// helpers with falling back to using either auth.json
+// file or .docker/config.json, including support for OAuth2 and IdentityToken.
+// Compared to GetCredentials, this function accepts a refPath (containing the
+// full reference path including the image name) together with the registry.
+// This refPath will be taken into account when looking up the credentials in
```
vrothberg, line 228: I think we need to elaborate on what "taken into account" exactly means.

pkg/docker/config/config.go
```diff
@@ -599,13 +614,27 @@ func findAuthentication(registry, path string, legacyFormat bool) (types.DockerA
 		normalizedAuths[normalizeRegistry(k)] = v
 	}
 
-	if val, exists := normalizedAuths[registry]; exists {
+	// If a reference path is provided, then we include it in the auth registry
+	// to be looked up.
+	registryLookup := registry
+	if refPath != "" {
+		registryLookup = stripImageName(registry, refPath)
+	}
```
vrothberg, line 622: We always pass a `refPath` around. So if we have a pre-existing auth.json with a credentials for `quay.io`, would we still be able to find them when pulling `quay.io/repo/image:tag`?

Doesn't that break backwards compat as we'd always look for credentials for `quay.io/repo/image`?

vrothberg (commented):
I think we should walk the path.  Let's assume we pull `quay.io/repo/image:tag`, I'd like the code to be able to find the following credentials in order:
* quay.io/repo/image
* quay.io/repo (to separate credentials for different users)
* quay.io (also to remain backwards compat)

docker/docker_client.go
saschagrunert: I've chosen a `reference.Named`, which I hope is fine as well since it's a more concrete type.

pkg/docker/config/config.go
saschagrunert: I agree and updated the docs for this function.

pkg/docker/config/config.go
saschagrunert: Yes, with the latest implementation.

docker/docker_client.go
vrothberg: Yes, even better!

pkg/docker/config/config.go
```diff
@@ -606,6 +634,30 @@ func findAuthentication(registry, path string, legacyFormat bool) (types.DockerA
 	return types.DockerAuthConfig{}, nil
 }
 
+func findAuthenticationForRef(auths dockerConfigFile, ref reference.Named) (types.DockerAuthConfig, error) {
+	for _, registry := range []string{ // example: quay.io/repo/image:tag
+		reference.TrimNamed(ref).String(), // quay.io/repo/image
+		stripImageName(ref),               // quay.io/repo
+		reference.Domain(ref),             // quay.io (backwards compatibility)
+	} {
```
vrothberg, line 642: I think we need a dedicated (for easier unit testing) "walking" function to create all candidates.  The paths can be nested deeper, e.g. `foo.com/1/2/3/4/5/image:tag`.

saschagrunert: Done, added a dedicated function `pathsForRef` and added unit tests

pkg/docker/config/config.go
```diff
@@ -14,6 +14,7 @@ import (
 	"github.com/containers/image/v5/pkg/sysregistriesv2"
 	"github.com/containers/image/v5/types"
 	"github.com/containers/storage/pkg/homedir"
+	"github.com/docker/distribution/reference"
```
vrothberg, line 17: This should  be `github.com/containers/image/v5/docker/reference`.  I guess your IDE auto-imported this one.

saschagrunert: Thank you for the hint, fixed as suggested :+1:

pkg/docker/config/config.go
```diff
@@ -216,12 +217,29 @@ func getAuthFilePaths(sys *types.SystemContext, homeDir string) []authPath {
...
+// helpers with falling back to using either auth.json
+// file or .docker/config.json, including support for OAuth2 and IdentityToken.
+// Compared to GetCredentials, this function accepts a full ref, which will
+// be used for supporting multiple paths when looking up the credentials in
+// auth.json. Path support means, that a given ref
+// "registry.io/some/path/image:latest" will lookup:
+// - registry.io/some/path/image
+// - registry.io/some/path
+// - registry.io/some
+// - registry.io (fallback for compatibility with GetCredentials)
```
mtrmac, line 234: This should not be an API promise. Callers (unlike software _creating_ auth.json) have no need to care about this, and new features frequently enough cause us to break such promises.

(In fact I’m tempted to drop even the list of sources this function reads from, above; just say “credentials appropriate for `sys` and the users’ environment”.)

pkg/docker/config/config.go
```diff
@@ -216,12 +217,29 @@ func getAuthFilePaths(sys *types.SystemContext, homeDir string) []authPath {
...
+// If an entry is not found, an empty struct is returned.
+func GetCredentialsWithRef(sys *types.SystemContext, ref reference.Named) (types.DockerAuthConfig, error) {
```
mtrmac, line 236: (Not a native speaker) `GetCredentialsForRef`?

pkg/docker/config/config.go
```diff
@@ -216,12 +217,29 @@ func getAuthFilePaths(sys *types.SystemContext, homeDir string) []authPath {
...
+	return getCredentialsWithHomeDir(sys, ref, reference.Domain(ref), homedir.Get())
 }
 
 // getCredentialsWithHomeDir is an internal implementation detail of GetCredentials,
```
mtrmac, line 240: Update this to list both functions?

pkg/docker/config/config.go
```diff
@@ -288,13 +306,13 @@ func getCredentialsWithHomeDir(sys *types.SystemContext, registry, homeDir strin
...
 func GetAuthentication(sys *types.SystemContext, registry string) (string, string, error) {
-	return getAuthenticationWithHomeDir(sys, registry, homedir.Get())
+	return getAuthenticationWithHomeDir(sys, nil, registry, homedir.Get())
 }
 
 // getAuthenticationWithHomeDir is an internal implementation detail of GetAuthentication,
 // it exists only to allow testing it with an artificial home directory.
-func getAuthenticationWithHomeDir(sys *types.SystemContext, registry, homeDir string) (string, string, error) {
-	auth, err := getCredentialsWithHomeDir(sys, registry, homeDir)
+func getAuthenticationWithHomeDir(sys *types.SystemContext, ref reference.Named, registry, homeDir string) (string, string, error) {
```
mtrmac, line 314: I wouldn’t bother with the new parameter here , and just call `getCredentialsWithHomeDir(…, nil, …)`. This is a legacy compatibility path with no new callers expected.

pkg/docker/config/config.go
```diff
@@ -216,12 +217,29 @@ func getAuthFilePaths(sys *types.SystemContext, homeDir string) []authPath {
...
+// - registry.io (fallback for compatibility with GetCredentials)
+// If an entry is not found, an empty struct is returned.
+func GetCredentialsWithRef(sys *types.SystemContext, ref reference.Named) (types.DockerAuthConfig, error) {
+	return getCredentialsWithHomeDir(sys, ref, reference.Domain(ref), homedir.Get())
 }
 
 // getCredentialsWithHomeDir is an internal implementation detail of GetCredentials,
 // it exists only to allow testing it with an artificial home directory.
-func getCredentialsWithHomeDir(sys *types.SystemContext, registry, homeDir string) (types.DockerAuthConfig, error) {
+func getCredentialsWithHomeDir(sys *types.SystemContext, ref reference.Named, registry, homeDir string) (types.DockerAuthConfig, error) {
```
mtrmac, line 242: Hum… we now have record priorities. Should we look up per-repo credentials in config files in preference to per-registry credentials in credential helpers, even if the helpers come first in `helpers`?

I’m leaning towards no (we are explicitly not supporting per-repo in helpers now, and if we ever added that, the prioritization would be a mess — and if anyone is writing per-repo credentials to files now, I can’t see much reason to have the per-registry ones not in the same files), bringing this up to confirm with others (and to have a record of our thinking).

pkg/docker/config/config.go
```diff
@@ -587,6 +607,16 @@ func findAuthentication(registry, path string, legacyFormat bool) (types.DockerA
 		return getAuthFromCredHelper(ch, registry)
 	}
 
+	// Support for different paths in auth.
+	if ref != nil {
+		return findAuthenticationForRef(auths, ref)
```
mtrmac, line 612: Non-blocking: Is this worth the extra two functions?

If the goal is testing, splitting this into “read file+parse JSON” and “interpret in-memory data”, or maybe “read file” + “parse and interpret byte stream” would allow better testing, because that would test the prioritization of repo/registry credentials.

pkg/docker/config/config.go
```diff
@@ -606,6 +636,45 @@ func findAuthentication(registry, path string, legacyFormat bool) (types.DockerA
 	return types.DockerAuthConfig{}, nil
 }
 
+func findAuthenticationForRef(auths dockerConfigFile, ref reference.Named) (types.DockerAuthConfig, error) {
+	for _, registry := range pathsForRef(ref) {
```
mtrmac, line 640: The iteration variable is not a `registry`. (Also maybe `authKeysForRef`, not sure about that one.)

pkg/docker/config/config.go
```diff
@@ -606,6 +636,45 @@ func findAuthentication(registry, path string, legacyFormat bool) (types.DockerA
...
+		normalizedAuth := normalizeRegistry(registry)
```
mtrmac, line 641: Do we need this at all? It’s tempting to steer users towards `docker.io/foo/bar`, `docker.io/foo`, `docker.io`.

(*NOTE*: verify K8s can accept those values. I haven’t now checked. Has anyone looked into this already?)

pkg/docker/config/config.go
```diff
@@ -606,6 +636,45 @@ func findAuthentication(registry, path string, legacyFormat bool) (types.DockerA
...
+}
+
+// stripImageName removes the image name from the provided ref.
+func stripImageName(ref reference.Named) string {
+	splitRefPath := strings.Split(reference.Path(ref), "/")
+	return strings.Join(append(
+		[]string{reference.Domain(ref)},
+		splitRefPath[:len(splitRefPath)-1]...,
+	), "/")
+}
```
mtrmac, line 676: This can be expressed simpler (and, much less important, a bit more efficient), at the very least not using a different approach for the initial `TrimNamed` vs. `Domain+Path`, and the `splitRefPath` array seems also unnecessary.

 Compare `docker/policyconfiguration.DockerReferenceNamespaces`.

pkg/docker/config/config.go
```diff
@@ -629,24 +698,28 @@ func decodeDockerAuth(conf dockerAuthConfig) (types.DockerAuthConfig, error) {
...
 
-	nameParts := strings.SplitN(stripped, "/", 2)
+	if strings.HasSuffix(stripped, "/v1") {
+		stripped = strings.TrimSuffix(stripped, "/v1")
+	} else if strings.HasSuffix(stripped, "/v2") {
+		stripped = strings.TrimSuffix(stripped, "/v2")
+	}
 
-	return nameParts[0]
+	return stripped
```
mtrmac, line 718: Changing what we do with the old URLs:
- Doesn’t work anyway, this mangles user-intended `quay.io/v2ofmyproject` namespaces
- Seems unnecessarily risky for these extremely-legacy formats (remember from elsewhere that `docker login https://arbitray/url/with/unexpected/contents` was accepted for years, and worked)

I’m thinking (not sure) we don’t need to invoke `normalizeRegistry` for the new repo entries at all, and even if we did want to convert the hostname, splitting `normalizeRegistry` into two, one dealing with full `.dockercfg` URLs and one dealing with only hostnames, would be much safer.

pkg/docker/config/config.go
```diff
@@ -587,6 +607,16 @@ func findAuthentication(registry, path string, legacyFormat bool) (types.DockerA
 		return getAuthFromCredHelper(ch, registry)
 	}
 
+	// Support for different paths in auth.
+	if ref != nil {
```
mtrmac, line 611: Just to be sure, restrict this to `!legacyFormat` where URLs may exist?

pkg/docker/config/config_test.go
```diff
@@ -280,12 +309,19 @@ func TestGetAuth(t *testing.T) {
...
+					assert.Nil(t, err)
+				}
+
+				auth, err := getCredentialsWithHomeDir(sys, ref, tc.hostname, tmpHomeDir)
 				assert.Equal(t, tc.expectedError, err)
 				assert.Equal(t, tc.expected, auth)
 
 				// Test for the previous APIs.
-				username, password, err := getAuthenticationWithHomeDir(sys, tc.hostname, tmpHomeDir)
+				username, password, err := getAuthenticationWithHomeDir(sys, ref, tc.hostname, tmpHomeDir)
```
mtrmac, line 324: This is not really representative of anything. Test with `nil` and do… something, maybe just ignore the return value if the test case doesn’t record the expected data, or another field to the test case.

pkg/docker/config/config.go
mtrmac, line 242: `if ref != nil && reference.Domain(ref) != registry { fail with internal error: inconsistent data }`, to make sure we catch such situations before they lead to hard-to-debug problems.

mtrmac (commented):
We’ll also need the write path (`login`/`logout`). Is that expected to be a separate PR? (That’s fine if so, this PR is big enough already — just to make sure this is on the radar.)

Note to self: Only skimmed the tests for now.

pkg/docker/config/config.go
saschagrunert: I thought about still stripping the `https://` prefix from registries as some sort of feature. What would be the reason to drop that?

pkg/docker/config/config.go
saschagrunert: Changed the function name with the next push.

pkg/docker/config/config.go
saschagrunert: No, I just refactored them back into the original function.

pkg/docker/config/config.go
saschagrunert: Yep, fixed as suggested :+1:

pkg/docker/config/config.go
saschagrunert: Fixed as suggested :+1:

pkg/docker/config/config.go
saschagrunert: Fixed as suggested :+1:

pkg/docker/config/config.go
saschagrunert: Fixed as suggested and removed the list.

pkg/docker/config/config.go
```diff
@@ -216,12 +217,29 @@ func getAuthFilePaths(sys *types.SystemContext, homeDir string) []authPath {
 // file or .docker/config.json, including support for OAuth2 and IdentityToken.
 // If an entry is not found, an empty struct is returned.
 func GetCredentials(sys *types.SystemContext, registry string) (types.DockerAuthConfig, error) {
```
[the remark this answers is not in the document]
saschagrunert: Done :+1:

docs/containers-auth.json.5.md
```diff
@@ -23,7 +23,10 @@ user to container image registries.  The file can have zero to many entries and
 is created by a `login` command from a container tool such as `podman login`,
 `buildah login` or `skopeo login`.  Each entry includes the name of the registry and then an auth
 token in the form of a base64 encoded string from the concatenation of the
-username, a colon, and the password.
+username, a colon, and the password. The registry name can additionally contain
+a path or image name (without tag or digest). This way it is possible to setup
```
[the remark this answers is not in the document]
saschagrunert: Improved the docs as suggested and added an example.

pkg/docker/config/config.go
saschagrunert: Sounds good, changed as suggested.

pkg/docker/config/config.go
saschagrunert: Okay, I splitted the http[s] trim part into a new function and added it to the path check in findAuthentication. This should not affect the original code path.

pkg/docker/config/config_test.go
saschagrunert: Removed the added parameter from this function so we now can preserve the original test case.

pkg/docker/config/config.go
saschagrunert: Thank you for the hint, changed as suggested and it's now much simpler.

pkg/docker/config/config.go
saschagrunert: Good, added the check and returning an error in that case.

saschagrunert:
> We’ll also need the write path (`login`/`logout`). Is that expected to be a separate PR? (That’s fine if so, this PR is big enough already — just to make sure this is on the radar.)

Yes, let's follow-up on this one in another PR.

pkg/docker/config/config.go
saschagrunert: I'm not sure about this point, maybe @vrothberg can chime in here.

pkg/docker/config/config.go
```diff
@@ -587,6 +614,16 @@ func findAuthentication(registry, path string, legacyFormat bool) (types.DockerA
...
+	// Support for different paths in auth.
+	if !legacyFormat && ref != nil {
+		for _, key := range authKeysForRef(ref) {
+			if val, exists := auths.AuthConfigs[key]; exists {
+				return decodeDockerAuth(val)
+			}
+		}
+	}
+
+	// Fallback to the registry based lookup.
```
vrothberg, line 626: I think this could be moved into an `} else {` branch above.

pkg/docker/config/config.go
vrothberg: SGTM

pkg/docker/config/config.go
saschagrunert: Yep, fixed as suggested :+1:

pkg/docker/config/config.go
saschagrunert: Just to clarify: So we leave it as it is, right?

pkg/docker/config/config.go
```diff
@@ -215,22 +216,46 @@ func getAuthFilePaths(sys *types.SystemContext, homeDir string) []authPath {
...
 
-// getCredentialsWithHomeDir is an internal implementation detail of GetCredentials,
-// it exists only to allow testing it with an artificial home directory.
-func getCredentialsWithHomeDir(sys *types.SystemContext, registry, homeDir string) (types.DockerAuthConfig, error) {
+// GetCredentialsForRef returns the registry credentials stored in the
+// registry-specific credential helpers or in the default global credentials
+// helpers with falling back to using either auth.json
+// file or .docker/config.json, including support for OAuth2 and IdentityToken.
+// Compared to GetCredentials, this function accepts a full ref, which will
+// be used for supporting multiple paths when looking up the credentials
```
mtrmac, line 231: A little more polishing: don’t talk about “registry credentials”, the “compared to GetCredentials” is not something readers care about.

Note to self: propose an alternative.

pkg/docker/config/config.go
```diff
@@ -215,22 +216,46 @@ func getAuthFilePaths(sys *types.SystemContext, homeDir string) []authPath {
 // helpers with falling back to using either auth.json
 // file or .docker/config.json, including support for OAuth2 and IdentityToken.
 // If an entry is not found, an empty struct is returned.
+//
+// Deprecated: GetCredentialsForRef should be used in favor of this API
+// because it supports the registry path lookup feature.
```
mtrmac, line 221:
```suggestion
// because it allows different credentials for different repositories on the same registry.
```

mtrmac (commented):
WIP, to shorten the iteration cycle.

Overall looks good, I still need to read the tests.

pkg/docker/config/config.go
mtrmac: My mistake, “registry credentials” is actually correct.

----

How about much simpler
```go
// GetCredentialsForRef returns the registry credentials necessary for
// accessing ref on the registry ref points to,
// appropriate for sys and the users’ configuration.
// If an entry is not found, an empty struct is returned.
```
? Or maybe expand this to “configuration (based on configuration in user’ home directory, global configuration files and elsewhere)” but that, vague enough not to commit to specifics, feels vague enough to not be useful to a caller.

----

Alternatively, we don’t need spend time in _this_ PR polishing the pre-existing documentation from `GetCredentials`:
```go
// GetCredentialsForRef returns the registry credentials necessary for
// accessing ref on the registry ref points to, stored in the
// registry-specific credential helpers or in the default global credentials
// helpers with falling back to using either auth.json
// file or .docker/config.json, including support for OAuth2 and IdentityToken.
// If an entry is not found, an empty struct is returned.
```

pkg/docker/config/config.go
```diff
@@ -587,25 +614,57 @@ func findAuthentication(registry, path string, legacyFormat bool) (types.DockerA
 		return getAuthFromCredHelper(ch, registry)
 	}
 
-	// I'm feeling lucky
-	if val, exists := auths.AuthConfigs[registry]; exists {
-		return decodeDockerAuth(val)
-	}
+	// Support for different paths in auth.
```
mtrmac, line 617: Absolutely non-blocking:
```suggestion
	// Support for different paths in auth.
	// (This is not a feature of ~/.docker/config.json; we support it even for those files as an extension.)
```
I don’t know where to put this; I could see a case for writing this in the man page but that man page doesn’t (authoritatively) document `.docker` anyway. Yet again, just the existence of this PR might be enough already.

pkg/docker/config/config_test.go
```diff
@@ -566,3 +624,54 @@ func TestGetAllCredentials(t *testing.T) {
...
+				"quay.io/a/b/c/d",
+				"quay.io/a/b/c",
+				"quay.io/a/b",
+				"quay.io/a",
+				"quay.io",
+			},
+		},
+	} {
+		ref, err := reference.ParseNamed(tc.ref)
+		require.Nil(t, err, tc.name)
```
mtrmac, line 672:
```suggestion
		require.NoError(t, err, tc.name)
```

pkg/docker/config/config_test.go
```diff
@@ -280,20 +329,29 @@ func TestGetAuth(t *testing.T) {
 				if tc.sys != nil {
 					sys = tc.sys
 				}
-				auth, err := getCredentialsWithHomeDir(sys, tc.hostname, tmpHomeDir)
+
+				var ref reference.Named
+				if tc.ref != "" {
+					ref, err = reference.ParseNamed(tc.ref)
+					assert.Nil(t, err)
```
mtrmac, line 336:
```suggestion
					require.NoError(t, err)
```

mtrmac (commented):
A full review now.

Still checking for URLs even if the caller submits a `ref` (because the primary caller is going to always submit one) is the only non-trivial concern, the rest is just nits and small cleanups.

pkg/docker/config/config.go
```diff
@@ -215,22 +216,46 @@ func getAuthFilePaths(sys *types.SystemContext, homeDir string) []authPath {
...
+// GetCredentialsForRef and GetCredentials. It exists only to allow testing it
+// with an artificial home directory.
+func getCredentialsWithHomeDir(sys *types.SystemContext, ref reference.Named, registry, homeDir string) (types.DockerAuthConfig, error) {
 	if sys != nil && sys.DockerAuthConfig != nil {
 		logrus.Debugf("Returning credentials for %s from DockerAuthConfig", registry)
 		return *sys.DockerAuthConfig, nil
 	}
 
+	// consistency check of the ref and registry arguments
+	if ref != nil && reference.Domain(ref) != registry {
```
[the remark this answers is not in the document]
saschagrunert: Done.

pkg/docker/config/config.go
saschagrunert: Fixed as suggested :+1:

pkg/docker/config/config.go
saschagrunert: Added the comment as suggested

docker/docker_client.go
```diff
@@ -343,6 +343,7 @@ func SearchRegistry(ctx context.Context, sys *types.SystemContext, registry, ima
 	v1Res := &V1Results{}
 
 	// Get credentials from authfile for the underlying hostname
+	// nolint: staticcheck
```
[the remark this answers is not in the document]
saschagrunert: Yes, by adding a custom rules go `.golangci.yml`. Changed to use that instead of the comment.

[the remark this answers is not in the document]
saschagrunert: Yeah I think the deprecation will help consumers when they lint their projects.

pkg/docker/config/config.go
```diff
@@ -587,25 +614,57 @@ func findAuthentication(registry, path string, legacyFormat bool) (types.DockerA
...
-		return decodeDockerAuth(val)
-	}
+	// Support for different paths in auth.
+	if !legacyFormat && ref != nil {
+		for _, key := range authKeysForRef(ref) {
+			if val, exists := auths.AuthConfigs[key]; exists {
+				return decodeDockerAuth(val)
+			}
+		}
+	} else {
```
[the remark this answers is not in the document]
saschagrunert: Alright, modified the code path as suggested and added the comments.

pkg/docker/config/config.go
saschagrunert: Thank you for the suggestion, I decided to use your simple version (the first one).

pkg/docker/config/config.go
vrothberg: Yes

vrothberg (commented):
LGTM

pkg/docker/config/config_test.go
mtrmac: (`require.NoError` aborts the test, unlike `assert.NoError` continuing, and doing something unexpected with invalid `ref`. OTOH if the test fails, that’s quite enough and worrying about _how_ it fails can get unproductive.)

docker/docker_client.go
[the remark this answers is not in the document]
mtrmac: Ouch, that‘s worse than silencing the checker, and the ideal fix doesn’t work: https://github.com/golangci/golangci-lint/issues/741 

I have ended up with
```go
	//lint:ignore SA1019 We can't use GetCredentialsForRef because we want to search the whole registry.
	auth, err := config.GetCredentials(sys, registry) //nolint:staticcheck // https://github.com/golangci/golangci-lint/issues/741
```

so that both a direct invocation of `staticcheck` and use via `golangci-lint` don’t complain.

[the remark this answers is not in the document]
saschagrunert: Fixed as suggested, thank you for your suggestion.

mtrmac (approved):
Thanks!

PR #1285, opened 2021-06-29
Implement "non-empty configuration" using reflect.DeepEqual

mtrmac:
A RFC improvement on top of #1283 .

This is more expensive, but it automatically handles any future member additions.

Preserve the existing semantics of treating non-nil but empty slices and maps as “not non-empty”, just because the old implementation did so; I didn't check whether that actually ever happens and whether it matters for any existing callers.


rhatdan:
LGTM

vrothberg (approved):
LGTM

PR #1287, opened 2021-06-30
Change all references to docker registry to container registry

rhatdan:
Will leave references to distribution spec until there is an OCI Spec we
can point at.

### DIFF
--- a/manifest/manifest.go
+++ b/manifest/manifest.go
@@ -195,7 +195,7 @@ func MatchesDigest(manifest []byte, expectedDigest digest.Digest) (bool, error) 
 }
 
 // AddDummyV2S1Signature adds an JWS signature with a temporary key (i.e. useless) to a v2s1 manifest.
-// This is useful to make the manifest acceptable to a Docker Registry (even though nothing needs or wants the JWS signature).
+// This is useful to make the manifest acceptable to a docker/distribution registry (even though nothing needs or wants the JWS signature).
 func AddDummyV2S1Signature(manifest []byte) ([]byte, error) {
 	key, err := libtrust.GenerateECP256PrivateKey()
 	if err != nil {
--- a/openshift/openshift.go
+++ b/openshift/openshift.go
@@ -164,7 +164,7 @@ type openshiftImageSource struct {
 	// Values specific to this image
 	sys *types.SystemContext
 	// State
-	docker               types.ImageSource // The Docker Registry endpoint, or nil if not resolved yet
+	docker               types.ImageSource // The docker/distribution API endpoint, or nil if not resolved yet
 	imageStreamImageName string            // Resolved image identifier, or "" if not known yet
 }
 
@@ -316,7 +316,7 @@ func (s *openshiftImageSource) ensureImageIsResolved(ctx context.Context) error 
 
 type openshiftImageDestination struct {
 	client *openshiftClient
-	docker types.ImageDestination // The Docker Registry endpoint
+	docker types.ImageDestination // The docker/distribution API endpoint
 	// State
 	imageStreamImageName string // "" if not yet known
 }
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 containers' images and container image registries.
 
 The containers/image library allows application to pull and push images from
-container image registries, like the upstream docker registry. It also
+container image registries, like the docker.io and quay.io registries. It also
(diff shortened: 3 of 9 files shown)


rhatdan:
@vrothberg Do you want these to say `container image registry`

vrothberg (commented):
LGTM

I am cool with "container registry" (less words). @mtrmac WDYT?

internal/iolimits/iolimits.go
```diff
@@ -17,7 +17,7 @@ const (
 	// `ReadAtMost`.
 	megaByte = 1 << 20
 	// MaxManifestBodySize is the maximum allowed size of a manifest. The limit
-	// of 4 MB aligns with the one of a Docker registry:
+	// of 4 MB aligns with the one of a container registry:
```
mtrmac, line 20: This is a reference to a specific implementation, not the generic concept.

manifest/manifest.go
```diff
@@ -195,7 +195,7 @@ func MatchesDigest(manifest []byte, expectedDigest digest.Digest) (bool, error)
 }
 
 // AddDummyV2S1Signature adds an JWS signature with a temporary key (i.e. useless) to a v2s1 manifest.
-// This is useful to make the manifest acceptable to a Docker Registry (even though nothing needs or wants the JWS signature).
+// This is useful to make the manifest acceptable to a container registry (even though nothing needs or wants the JWS signature).
```
mtrmac, line 198: This is (mostly() a reference to a specific implementation.

openshift/openshift.go
```diff
@@ -164,7 +164,7 @@ type openshiftImageSource struct {
 	// Values specific to this image
 	sys *types.SystemContext
 	// State
-	docker               types.ImageSource // The Docker Registry endpoint, or nil if not resolved yet
+	docker               types.ImageSource // The container registry endpoint, or nil if not resolved yet
```
mtrmac, line 167: This is confusing; the OpenShift integrated registry is also a “container registry”, so a “container registry endpoint” is not a distinction.

Refer to the docker/distribution (distribution/distribution? the WIP OCI distribution idea?) protocol here.

Also the other instance.

internal/iolimits/iolimits.go
rhatdan: I am not sure what you means Docker distribution registry?  Docker.io?

manifest/manifest.go
rhatdan: Should we say older container registry, previous container registry?  Or is the a specific version of the Docker distribution that you want to reference?

openshift/openshift.go
rhatdan: So you want to say the docker/distribution endpoint?

internal/iolimits/iolimits.go
mtrmac: The line just below links to the implementation in question. (I don’t know what exactly runs `docker.io` but it’s attractive to guess there’s a relationship.)

manifest/manifest.go
mtrmac: The commit message for this function refers to docker/distribution.

I have no idea whether other implementations accept unsigned v2s1.

openshift/openshift.go
mtrmac: “docker/distribution API” I guess

rhatdan: Fixed

internal/iolimits/iolimits.go
rhatdan: Removed change.  Not really worth changing some of these.  I wanted to change the ones that show up in users faces.  But just grepped the code for any "docker registry" string.

openshift/openshift.go
```diff
@@ -316,7 +316,7 @@ func (s *openshiftImageSource) ensureImageIsResolved(ctx context.Context) error
 
 type openshiftImageDestination struct {
 	client *openshiftClient
-	docker types.ImageDestination // The Docker Registry endpoint
+	docker types.ImageDestination // The container registry endpoint
```
mtrmac, line 319: `docker/distribution API` here as well, please.

mtrmac (commented):
LGTM after one more change.

Thanks!

mtrmac (approved):

TomSweeneyRedHat:
LGTM

PR #1288, opened 2021-07-01
Support updating registry credentials scoped to namespaces/repos

saschagrunert:
We now error on login if repositories or repository namespaces are used
for other credential helpers than the `AuthenticationFileHelper`. On
logout we ignore them and debug log a warning that nothing has been
modified.

The functions `SetCredentials` (for login) as well as
`RemoveAuthentication` (for logout) already feature support for path
based registries for the `AuthenticationFileHelper`. This patch adds
unit tests to ensure that the support will not break in the future.

Refers to https://github.com/containers/image/issues/1276

### DIFF
--- a/docs/containers-auth.json.5.md
+++ b/docs/containers-auth.json.5.md
@@ -21,14 +21,15 @@ Except the primary (read/write) file, other files are read-only, unless the user
 The auth.json file stores encrypted authentication information for the
 user to container image registries.  The file can have zero to many entries and
 is created by a `login` command from a container tool such as `podman login`,
-`buildah login` or `skopeo login`.  Each entry includes the name of the registry and then an auth
-token in the form of a base64 encoded string from the concatenation of the
-username, a colon, and the password. The registry name can additionally contain
-a path or repository name (an image name without tag or digest). The path (or
-namespace) is matched in its hierarchical order when checking for available
-authentications. For example, an image pull for
-`my-registry.local/namespace/user/image:latest` will result in a lookup in
-`auth.json` in the following order:
+`buildah login` or `skopeo login`. Each entry either contains a single
+hostname (e.g. `docker.io`) or a namespace (e.g. `quay.io/user/image`) as a key
+and an auth token in the form of a base64 encoded string as value of `auth`. The
+token is built from the concatenation of the username, a colon, and the
+password. The registry name can additionally contain a repository name (an image
+name without tag or digest) and namespaces. The path (or namespace) is matched
+in its hierarchical order when checking for available authentications. For
+example, an image pull for `my-registry.local/namespace/user/image:latest` will
+result in a lookup in `auth.json` in the following order:
 
 - `my-registry.local/namespace/user/image`
 - `my-registry.local/namespace/user`
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -54,10 +54,17 @@ var (
 
 // SetCredentials stores the username and password in a location
 // appropriate for sys and the users’ configuration.
+// A valid key can be either a registry hostname or additionally a namespace if
+// the AuthenticationFileHelper is being unsed.
 // Returns a human-redable description of the location that was updated.
 // NOTE: The return value is only intended to be read by humans; its form is not an API,
 // it may change (or new forms can be added) any time.
-func SetCredentials(sys *types.SystemContext, registry, username, password string) (string, error) {
+func SetCredentials(sys *types.SystemContext, key, username, password string) (string, error) {
+	isNamespaced, err := validateKey(key)
+	if err != nil {
+		return "", err
+	}
+
 	helpers, err := sysregistriesv2.CredentialHelpers(sys)
 	if err != nil {
 		return "", err
@@ -72,33 +79,45 @@ func SetCredentials(sys *types.SystemContext, registry, username, password strin
 		// Special-case the built-in helpers for auth files.
 		case sysregistriesv2.AuthenticationFileHelper:
 			desc, err = modifyJSON(sys, func(auths *dockerConfigFile) (bool, error) {
-				if ch, exists := auths.CredHelpers[registry]; exists {
-					return false, setAuthToCredHelper(ch, registry, username, password)
+				if ch, exists := auths.CredHelpers[key]; exists {
+					if isNamespaced {
+						return false, unsupportedNamespaceErr(ch)
+					}
+					return false, setAuthToCredHelper(ch, key, username, password)
 				}
 				creds := base64.StdEncoding.EncodeToString([]byte(username + ":" + password))
 				newCreds := dockerAuthConfig{Auth: creds}
-				auths.AuthConfigs[registry] = newCreds
+				auths.AuthConfigs[key] = newCreds
 				return true, nil
 			})
 		// External helpers.
 		default:
-			desc = fmt.Sprintf("credential helper: %s", helper)
-			err = setAuthToCredHelper(helper, registry, username, password)
+			if isNamespaced {
+				err = unsupportedNamespaceErr(helper)
+			} else {
+				desc = fmt.Sprintf("credential helper: %s", helper)
+				err = setAuthToCredHelper(helper, key, username, password)
+			}
 		}
 		if err != nil {
 			multiErr = multierror.Append(multiErr, err)
-			logrus.Debugf("Error storing credentials for %s in credential helper %s: %v", registry, helper, err)
+			logrus.Debugf("Error storing credentials for %s in credential helper %s: %v", key, helper, err)
 			continue
 		}
-		logrus.Debugf("Stored credentials for %s in credential helper %s", registry, helper)
+		logrus.Debugf("Stored credentials for %s in credential helper %s", key, helper)
 		return desc, nil
 	}
 	return "", multiErr
 }
 
+func unsupportedNamespaceErr(helper string) error {
+	return errors.Errorf("namespaced key is not supported for credential helper %s", helper)
+}
+
 // SetAuthentication stores the username and password in the credential helper or file
-func SetAuthentication(sys *types.SystemContext, registry, username, password string) error {
-	_, err := SetCredentials(sys, registry, username, password)
+// See the documentation of SetCredentials for format of "key"
+func SetAuthentication(sys *types.SystemContext, key, username, password string) error {
+	_, err := SetCredentials(sys, key, username, password)
 	return err
 }
 
@@ -326,9 +345,16 @@ func getAuthenticationWithHomeDir(sys *types.SystemContext, registry, homeDir st
 	return auth.Username, auth.Password, nil
 }
 
-// RemoveAuthentication removes credentials for `registry` from all possible
+// RemoveAuthentication removes credentials for `key` from all possible
 // sources such as credential helpers and auth files.
-func RemoveAuthentication(sys *types.SystemContext, registry string) error {
+// A valid key can be either a registry hostname or additionally a namespace if
+// the AuthenticationFileHelper is being unsed.
+func RemoveAuthentication(sys *types.SystemContext, key string) error {
+	isNamespaced, err := validateKey(key)
+	if err != nil {
+		return err
+	}
+
 	helpers, err := sysregistriesv2.CredentialHelpers(sys)
 	if err != nil {
 		return err
@@ -338,17 +364,22 @@ func RemoveAuthentication(sys *types.SystemContext, registry string) error {
 	isLoggedIn := false
 
 	removeFromCredHelper := func(helper string) {
-		err := deleteAuthFromCredHelper(helper, registry)
-		if err == nil {
-			logrus.Debugf("Credentials for %q were deleted from credential helper %s", registry, helper)
-			isLoggedIn = true
+		if isNamespaced {
+			logrus.Debugf("Not removing credentials because namespaced keys are not supported for the credential helper: %s", helper)
 			return
+		} else {
+			err := deleteAuthFromCredHelper(helper, key)
+			if err == nil {
+				logrus.Debugf("Credentials for %q were deleted from credential helper %s", key, helper)
+				isLoggedIn = true
+				return
+			}
+			if credentials.IsErrCredentialsNotFoundMessage(err.Error()) {
+				logrus.Debugf("Not logged in to %s with credential helper %s", key, helper)
+				return
+			}
 		}
-		if credentials.IsErrCredentialsNotFoundMessage(err.Error()) {
-			logrus.Debugf("Not logged in to %s with credential helper %s", registry, helper)
-			return
-		}
-		multiErr = multierror.Append(multiErr, errors.Wrapf(err, "removing credentials for %s from credential helper %s", registry, helper))
+		multiErr = multierror.Append(multiErr, errors.Wrapf(err, "removing credentials for %s from credential helper %s", key, helper))
 	}
 
 	for _, helper := range helpers {
@@ -357,15 +388,12 @@ func RemoveAuthentication(sys *types.SystemContext, registry string) error {
 		// Special-case the built-in helper for auth files.
 		case sysregistriesv2.AuthenticationFileHelper:
 			_, err = modifyJSON(sys, func(auths *dockerConfigFile) (bool, error) {
-				if innerHelper, exists := auths.CredHelpers[registry]; exists {
+				if innerHelper, exists := auths.CredHelpers[key]; exists {
 					removeFromCredHelper(innerHelper)
 				}
-				if _, ok := auths.AuthConfigs[registry]; ok {
+				if _, ok := auths.AuthConfigs[key]; ok {
 					isLoggedIn = true
-					delete(auths.AuthConfigs, registry)
-				} else if _, ok := auths.AuthConfigs[normalizeRegistry(registry)]; ok {
-					isLoggedIn = true
-					delete(auths.AuthConfigs, normalizeRegistry(registry))
+					delete(auths.AuthConfigs, key)
 				}
 				return true, multiErr
 			})
@@ -699,16 +727,16 @@ func decodeDockerAuth(conf dockerAuthConfig) (types.DockerAuthConfig, error) {
 // to just an hostname.
 // Copied from github.com/docker/docker/registry/auth.go
 func convertToHostname(url string) string {
-	stripped := url
-	if strings.HasPrefix(url, "http://") {
-		stripped = strings.TrimPrefix(url, "http://")
-	} else if strings.HasPrefix(url, "https://") {
-		stripped = strings.TrimPrefix(url, "https://")
-	}
-
+	stripped := stripScheme(url)
 	nameParts := strings.SplitN(stripped, "/", 2)
-
 	return nameParts[0]
+}
+
+// stripScheme striped the http|https scheme from the provided URL.
+func stripScheme(url string) string {
+	stripped := strings.TrimPrefix(url, "http://")
+	stripped = strings.TrimPrefix(stripped, "https://")
+	return stripped
 }
 
 func normalizeRegistry(registry string) string {
@@ -718,4 +746,15 @@ func normalizeRegistry(registry string) string {
 		return "index.docker.io"
 	}
 	return normalized
+}
+
+// validateKey verifies that the input key does not have a prefix that is not
+// allowed and returns an indicator if the key is namespaced.
+func validateKey(key string) (isNamespaced bool, err error) {
+	if strings.HasPrefix(key, "http://") || strings.HasPrefix(key, "https://") {
+		return isNamespaced, errors.Errorf("key %s contains http[s]:// prefix", key)
+	}
+
+	// check if the provided key contains one or more subpaths.
+	return strings.ContainsRune(key, '/'), nil
 }
--- a/pkg/docker/config/config_test.go
+++ b/pkg/docker/config/config_test.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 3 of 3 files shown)


saschagrunert:
I was checking the functionality during my follow-up mentioned in https://github.com/containers/image/pull/1278#pullrequestreview-693921949. Looks like we're mostly set in c/image, whereas c/common would require some slight adaption to make login/logout work correctly.

saschagrunert:
I was checking the functionality during my follow-up mentioned in https://github.com/containers/image/pull/1278#pullrequestreview-693921949. Looks like we're mostly set in c/image, whereas c/common would require some slight adaption to make login/logout work correctly.

rhatdan:
LGTM

mtrmac:
> The functions `SetCredentials` (for login) as well as
> `RemoveAuthentication` (for logout) already feature support for path
> based registries. This patch adds unit tests to ensure that the support
> will not break in the future.

That’s not good, because the functions blindly send such values to credential helpers, which may work very badly (see https://issues.redhat.com/browse/RUN-1246?focusedCommentId=16344129&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16344129 ).

Also, if we are not adding new functions, the documentation of `SetCredentials`/`RemoveAuthentication` should be expanded to document that new input is allowed (well, _some_ new input; do we want to allow tagged/digested references?).

WRT credential helpers, for now it would be consistent to just refuse to write namespaced credentials to them (and `RemoveAuthentication` then needs to account for that by not failing if trying to remove something that we would never add/look up). Eventually we might need to either propose fixes to the helpers, or maintain a list of known good/bad helpers…

saschagrunert:
@mtrmac would it be better to normalize the input registry for every credentials helper other than `AuthenticationFileHelper` or test the input registry for being namespaced and fail in that case if the helper is not the `AuthenticationFileHelper`?

mtrmac:
If the user requests to record a login for `quay.io/foo` and later for `quay.io/bar`, it wouldn’t do to store both to `quay.io`.

IIRC from a quick skim, we already ~silently continue if a helper refuses data; so I think we can continue to do that and just refuse to even try with non-`AuthenticationFileHelper` helpers.

Non-blocking? It might be useful to provide a typed error for such a case, or at least to special-case the `multiErr` handling here, so that the `* login` commands can report a meaningful “namespaced credentials are not supported with credential helpers” error.

saschagrunert:
Added a new error type `ErrNamespacedRegistry` if the registry is namespaced and a non auth.json credentials helper is being used.

TomSweeneyRedHat:
@saschagrunert looks like you need a rebase here.

saschagrunert:
The branch is not on the latest master but I don't see any conflicts :)

vrothberg (commented):
I have troubles understanding the meaning of "namespaced registry". It mostly relates to using the "namespaced" adjective (all registries have namespaces) but also c/image usually uses "repo[sitory]" rather than "namespace".

I am also a bit worried of the implications of the new error.  Can we be certain that registry+repo/path is not supported for creds helpers?  I tried looking in the (scarse) docs (https://pkg.go.dev/github.com/docker/docker-credential-helpers@v0.6.4/credentials#Credentials) but could not find much info on what a "ServerURL" is.

mtrmac:
> I have troubles understanding the meaning of "namespaced registry". It mostly relates to using the "namespaced" adjective (all registries have namespaces) but also c/image usually uses "repo[sitory]" rather than "namespace".

Yeah the key is not a “registry”, so a “namespaces registry” is not ideal. We should think of a name and use it consistently throughout.
* host[:port] = registry
* host[:port]/[ns1/…/nsX/]name = repository
* host[:port]/[ns1/…]/nsX (meant as a parent of …/nsX/name) = namespace; sometimes namespaces might also include leaf repositories (?)
* any of the above = ???

containers-registries.conf(5) just define a similar syntax (but with leading wildcards) for the `prefix` field, and then talks about “like the `prefix` field”.

containers-policy.json(5) talks about “scopes” (but in a more generic transport-dependent way).

I don’t have a good suggestion. “Credentials scope”? “Credential [storage] key”? Both are pretty bad.

---

> I am also a bit worried of the implications of the new error. Can we be certain that registry+repo/path is not supported for creds helpers? I tried looking in the (scarse) docs (https://pkg.go.dev/github.com/docker/docker-credential-helpers@v0.6.4/credentials#Credentials) but could not find much info on what a "ServerURL" is.

See the various helper implementations in that repo:
- `osxkeychain` parses the serverURL into protocol://host:port/path, and stores all of that (but not possible URL query/fragment parts)
- `pass` and `secretservice` seemingly treat `serverURL` as an opaque byte sequence with no parsing/normalization
-  `wincred` does approximate matching where basically only the hostname must match.

For now this is already ~decided by #1278 , which does not call helpers for any non-registry keys, and this just does the same thing on the write path; if we don’t look for per-repo credentials in helpers, it doesn’t make much sense to report success on storing the credentials in those helpers. Though we can certainly revisit, and now before a release would be the best time :)

pkg/docker/config/config.go
```diff
@@ -719,3 +742,14 @@ func normalizeRegistry(registry string) string {
...
 	return normalized
 }
+
+// isNamespaced returns true if the provided registry contains one or more
+// subpaths.
+func isNamespaced(registry string) bool {
+	stripped := stripScheme(registry)
+
+	// don't take trailling slashes into account
+	stripped = strings.TrimSuffix(stripped, "/")
```
mtrmac, line 752: We are not going to find entries with trailing slashes, so we should not allow writing them either.

pkg/docker/config/config.go
```diff
@@ -719,3 +742,14 @@ func normalizeRegistry(registry string) string {
...
+
+	return len(strings.Split(stripped, "/")) > 1
```
mtrmac, line 754: This is just `containsRune(stripped, '/')`, isn’t it?

pkg/docker/config/config.go
```diff
@@ -83,7 +90,12 @@ func SetCredentials(sys *types.SystemContext, registry, username, password strin
 		// External helpers.
```
mtrmac, line 90: Above, the `CredHelpers[registry]` case also needs to be rejected. `RemoveAuthentication` also needs to handle that case.

pkg/docker/config/config.go
```diff
@@ -50,10 +50,17 @@ var (
...
+	// supported for the credential helper. Only the
+	// sysregistriesv2.AuthenticationFileHelper supports namespaced registries.
+	ErrNamespacedRegistry = errors.New("namespaced registry not supported for credential helper")
 )
 
 // SetCredentials stores the username and password in the credential helper or file
 // and returns path to file or helper name in format (helper:%s).
+// An ErrNamespacedRegistry will be part of the returned multierror if the
+// provided registry is namespaced and a credentials helper other than the
+// AuthenticationFileHelper is used.
```
mtrmac, line 63: Hum… do we want this as a multi-error component or as a single top-level error?
I.e. the caller gets one of:
- success
- error indicating a failure to write to file, _and nothing else_ (even if there are helpers; or should there be one `ErrNamespacedRegistry` in that case?)
- `ErrNamespacedRegistry` only (if files aren't used, and even if there are many helpers)

_What does c/common/pkg/auth need for a reasonable error message?_

3 repetitions of `ErrNamespacedRegistry` with nothing else are not really useful; at that point 3 untyped error messages to the effect of “helper $foo is configured, but we are not storing namespace credentials there” would work at least as well.

I imagine that c/common/pkg/auth would like a single top-level `ErrNamespacedRegistry` to check against, if that’s the _only_ reason writing fails, so that it can write a nice error message with explanations. Or maybe c/image should just return such a nice error message already, with pkg/auth not having to care, and c/image doesn’t need to define an API-visible error.

pkg/docker/config/config_test.go
```diff
@@ -675,3 +676,170 @@ func TestAuthKeysForRef(t *testing.T) {
...
+				require.NotEmpty(t, auth.AuthConfigs["quay.io"].Auth)
+				require.NotEmpty(t, auth.AuthConfigs["my-registry.local"].Auth)
+			},
+		},
+	} {
+		tmpFile, err := ioutil.TempFile("", "auth.json.set")
+		require.NoError(t, err)
+		_, err = tmpFile.WriteString("{}")
+		require.NoError(t, err)
+		defer os.RemoveAll(tmpFile.Name())
```
mtrmac, line 722: Non-blocking: This conceptually belongs just after `TempFile` succeeds.

pkg/docker/config/config_test.go
```diff
@@ -675,3 +676,170 @@ func TestAuthKeysForRef(t *testing.T) {
...
+		sys := &types.SystemContext{AuthFilePath: tmpFile.Name()}
+
+		for _, registry := range tc.registries {
+			_, err := SetCredentials(sys, registry, "username", "password")
```
mtrmac, line 726: The credentials are free-form strings, so maybe use (`username`+loopIndex, `password`+loopIndex). Then…

pkg/docker/config/config_test.go
```diff
@@ -675,3 +676,170 @@ func TestAuthKeysForRef(t *testing.T) {
...
+		},
+	} {
+		tmpFile, err := ioutil.TempFile("", "auth.json.set")
+		require.NoError(t, err)
+		_, err = tmpFile.WriteString("{}")
+		require.NoError(t, err)
+		defer os.RemoveAll(tmpFile.Name())
+		sys := &types.SystemContext{AuthFilePath: tmpFile.Name()}
+
+		for _, registry := range tc.registries {
```
mtrmac, line 725: The variable is not a registry (== host[:port]); the array is not `registries`.

pkg/docker/config/config_test.go
```diff
@@ -675,3 +676,170 @@ func TestAuthKeysForRef(t *testing.T) {
...
+			_, err := SetCredentials(sys, registry, "username", "password")
+			require.NoError(t, err)
+		}
+
+		auth, err := readJSONFile(tmpFile.Name(), false)
+		require.NoError(t, err)
+
+		tc.assert(auth)
```
mtrmac, line 733: … after this check, this could actually call `GetCredentialsForRef` to test that the two interoperate and the right credentials are found.

pkg/docker/config/config_test.go
```diff
@@ -675,3 +676,170 @@ func TestAuthKeysForRef(t *testing.T) {
 		require.Equal(t, tc.expected, result, tc.name)
 	}
 }
+
+func TestSetCredentials(t *testing.T) {
```
mtrmac, line 680: (It’s great to finally have tests for these functions. Thanks!)

pkg/docker/config/config_test.go
```diff
@@ -675,3 +676,170 @@ func TestAuthKeysForRef(t *testing.T) {
...
+	for _, tc := range []struct {
+		registries []string
+		assert     func(dockerConfigFile)
```
mtrmac, line 683: Non-blocking: This looks like it could be just an `expectedKeys []string`, or something like that.

pkg/docker/config/config_test.go
```diff
@@ -675,3 +676,170 @@ func TestAuthKeysForRef(t *testing.T) {
...
+	}{
+		{
+			registries: []string{"quay.io"},
+			assert: func(auth dockerConfigFile) {
+				require.Len(t, auth.AuthConfigs, 1)
+				require.NotEmpty(t, auth.AuthConfigs["quay.io"].Auth)
```
mtrmac, line 689: Non-blocking, throughout: `require.` for conditions that break tests, e.g. a failure setting up the environment, or an unexpected error we don’t get a value necessary for other checks; `assert.` for conditions that are a test failure but still allow continuing tests.

OTOH, sure, a test failure is a test failure, and needs to be fixed much more urgently than tinkering with how the test failure looks :)

pkg/docker/config/config_test.go
```diff
@@ -675,3 +676,170 @@ func TestAuthKeysForRef(t *testing.T) {
...
+		tc.assert(auth)
+	}
+}
+
+func TestRemoveAuthentication(t *testing.T) {
+	testAuth := dockerAuthConfig{Auth: "ZXhhbXBsZTpvcmc="}
+	for _, tc := range []struct {
+		config     dockerConfigFile
+		registries []string
+		assert     func(dockerConfigFile)
```
mtrmac, line 742: Non-blocking: This could probably be a `expectedKeys []string`.

pkg/docker/config/config_test.go
```diff
@@ -675,3 +676,170 @@ func TestAuthKeysForRef(t *testing.T) {
...
+		require.NoError(t, err)
+
+		tc.assert(auth)
+	}
+}
+
+func TestRemoveAuthentication(t *testing.T) {
+	testAuth := dockerAuthConfig{Auth: "ZXhhbXBsZTpvcmc="}
+	for _, tc := range []struct {
+		config     dockerConfigFile
```
mtrmac, line 740: (Absolutely non-blocking: This could be a `preexistingKeys []string`. OTOH it would not be that much of a win.)

pkg/docker/config/config_test.go
```diff
@@ -675,3 +676,170 @@ func TestAuthKeysForRef(t *testing.T) {
...
+		tc.assert(auth)
+	}
+}
+
+func TestIsNamespaced(t *testing.T) {
+	for _, tc := range []struct {
+		registry string
+		expected bool
+	}{
+		{
```
mtrmac, line 820: (Absolutely non-blocking: consider squeezing each entry to a single line instead of 4.)

pkg/docker/config/config_test.go
```diff
@@ -675,3 +676,170 @@ func TestAuthKeysForRef(t *testing.T) {
...
+		tmpFile, err := ioutil.TempFile("", "auth.json")
+		require.NoError(t, err)
+		defer os.RemoveAll(tmpFile.Name())
+
+		_, err = tmpFile.Write(content)
+		require.NoError(t, err)
+
+		sys := &types.SystemContext{AuthFilePath: tmpFile.Name()}
+
+		for _, registry := range tc.registries {
```
mtrmac, line 803: The variable is not a registry.

pkg/docker/config/config.go
```diff
@@ -50,10 +50,17 @@ var (
 	ErrNotLoggedIn = errors.New("not logged in")
 	// ErrNotSupported is returned for unsupported methods
 	ErrNotSupported = errors.New("not supported")
+	// ErrNamespacedRegistry is returned if namespaced registries are not
+	// supported for the credential helper. Only the
+	// sysregistriesv2.AuthenticationFileHelper supports namespaced registries.
```
mtrmac, line 55:
```suggestion
	// ErrNamespacedRegistry is returned if storing credentials for namespaced registries
	// is not supported on the current system.
```
(modulo the name for “namespaced registries”), please.

Let’s make no promises about supporting them _only_ in files, or we would break that if adding more mechanisms.

mtrmac (commented):
To be explicit, I’m treating this PR as a full “add support for editing namespaced credentials” feature addition. If that’s a bad assumption, please correct me.

pkg/docker/config/config.go
saschagrunert: Alright, removed that additional modification.

pkg/docker/config/config.go
saschagrunert: Yep, fixed as suggested :+1:

pkg/docker/config/config.go
saschagrunert: Added an additional path check for both cases.

pkg/docker/config/config_test.go
saschagrunert: Sounds good, adapted as suggested :+1:

pkg/docker/config/config_test.go
saschagrunert: Yep, fixed

pkg/docker/config/config_test.go
saschagrunert: Yes, I changed it to `input` to be more generic.

pkg/docker/config/config_test.go
saschagrunert: I decided to add it to the actual asser calls, but generally agree checking the username/password makes sense.

pkg/docker/config/config_test.go
saschagrunert: Yeah, not any more after my modification of the `assert` function. :upside_down_face:

pkg/docker/config/config_test.go
saschagrunert: Fixed :+1:

pkg/docker/config/config_test.go
saschagrunert: Fixed :+1:

pkg/docker/config/config_test.go
saschagrunert: Got it, fixed as suggested :+1:

pkg/docker/config/config.go
saschagrunert: Fixed as suggested

pkg/docker/config/config.go
saschagrunert: Hm, we can wrap the error with the name of the credentials helper to provide more context. I decided to implement that, not sure if it's ideal.

pkg/docker/config/config.go
```diff
@@ -328,6 +340,9 @@ func getAuthenticationWithHomeDir(sys *types.SystemContext, registry, homeDir st
 
 // RemoveAuthentication removes credentials for `registry` from all possible
 // sources such as credential helpers and auth files.
+// An ErrNamespacedRegistry will be part of the returned multierror if the
+// provided registry is namespaced and a credentials helper other than the
+// AuthenticationFileHelper is used.
```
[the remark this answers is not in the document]
saschagrunert: The content of the helper could be modified manually by a third party. So I'd tend to say we should check the integrity during save and remove.

pkg/docker/config/config.go
```diff
@@ -50,10 +50,17 @@ var (
...
 )
 
 // SetCredentials stores the username and password in the credential helper or file
 // and returns path to file or helper name in format (helper:%s).
+// An ErrNamespacedRegistry will be part of the returned multierror if the
+// provided registry is namespaced and a credentials helper other than the
+// AuthenticationFileHelper is used.
 // Returns a human-redable description of the location that was updated.
 // NOTE: The return value is only intended to be read by humans; its form is not an API,
 // it may change (or new forms can be added) any time.
```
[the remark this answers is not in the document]
saschagrunert: > And SetCredentials, as well as RemoveAuthentication, should start validating the format to ensure we don’t write unexpected data to the configs; at the very least reject http[s]:// prefixes, maybe something like validateIdentityRemappingPrefix.

Implemented this as suggested.

saschagrunert:
Alright, just a few points are left open, mostly about the term "namespaced" and "registry". How about:

- changing "registry" to "key", mentioning in the docs that it should be a "registry" for any credentials helper than the auth file, which supports full "repositories"
- chaning the `ErrNamespacedRegistry` to `ErrRepositoryPath` or `ErrUnsupportedRepository`, …

pkg/docker/config/config_test.go
saschagrunert: Decided to leave the assert function to have more flexibility for test additions.

pkg/docker/config/config_test.go
saschagrunert: Decided to leave the assert function to have more flexibility for test additions.

pkg/docker/config/config.go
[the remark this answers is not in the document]
mtrmac: If we are worried about someone manually adding a namespaced credential (and we think we have any business modifying an entry we never read, which I’m not at all sure about), that would be a reason to try to delete the credential but not to fail if it is missing; it’s not at all a reason to hard-fail on `registryIsNamespaced` without even looking at what the helper contains, the way the implementation currently does.

pkg/docker/config/config.go
mtrmac: Works for me.

If the caller is not going to be specifically checking for this value (and c/common currently isn’t), a `fmt.Errorf("whatever_we_decide_to_call_namespaced_registry is not supported for credential helper %s", helperName)` is easier to understand than wrapping, and we don’t need to expose an API.

pkg/docker/config/config.go
```diff
@@ -719,3 +751,20 @@ func normalizeRegistry(registry string) string {
...
+// isNamespaced returns true if the provided registry contains one or more
+// subpaths.
+func isNamespaced(registry string) bool {
+	stripped := stripScheme(registry)
+	return strings.ContainsRune(stripped, '/')
+}
+
+// validateIdentityRemappingPrefix verifies that the input registry does not
+// have any not allowed prefix.
+func validateIdentityRemappingPrefix(registry string) error {
```
mtrmac, line 764: The name does not make sense here, we are not validating an “identity remapping prefix”. (And the input is not a `registry`).

Blocked on the name for the concept, sure.

pkg/docker/config/config.go
```diff
@@ -719,3 +751,20 @@ func normalizeRegistry(registry string) string {
 	}
 	return normalized
 }
+
+// isNamespaced returns true if the provided registry contains one or more
+// subpaths.
+func isNamespaced(registry string) bool {
```
mtrmac, line 757: The input is not a registry. (Similarly the `registryIsNamespaced` variable name doesn’t make sense.)

mtrmac (commented):
So that it isn’t forgotten, we will need a solution for namespaces-only logins, maybe https://github.com/containers/common/pull/659#issuecomment-875505807 .

mtrmac:
> Alright, just a few points are left open, mostly about the term "namespaced" and "registry". How about:
> 
> * changing "registry" to "key",

\*shrug\* works for me. @vrothberg ?

> mentioning in the docs that it should be a "registry" for any credentials helper than the auth file, which supports full "repositories"

repositories and repository namespaces

> * chaning the `ErrNamespacedRegistry` to `ErrRepositoryPath` or `ErrUnsupportedRepository`, …

If nothing is checking for that value (as is currently the case), maybe we can just not provide an API and this specific instance goes away (OTOH naming still affects most of the PR throughout anyway).

vrothberg:
SGTM

pkg/docker/config/config.go
```diff
@@ -50,10 +50,17 @@ var (
 	ErrNotLoggedIn = errors.New("not logged in")
 	// ErrNotSupported is returned for unsupported methods
 	ErrNotSupported = errors.New("not supported")
+	// ErrNamespacedRegistry is returned if namespaced registries are not
+	// supported for the credential helper. Only the
+	// sysregistriesv2.AuthenticationFileHelper supports namespaced registries.
+	ErrNamespacedRegistry = errors.New("namespaced registry not supported for credential helper")
```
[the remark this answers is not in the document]
saschagrunert: Removed the error with a non-api helper function to create a human readable error string.

pkg/docker/config/config.go
saschagrunert: SGTM, improved the wrapping as suggested

pkg/docker/config/config.go
saschagrunert: Function got removed after the next push and refactored into `validateKey`

pkg/docker/config/config.go
```diff
@@ -719,3 +751,20 @@ func normalizeRegistry(registry string) string {
 	}
 	return normalized
 }
+
+// isNamespaced returns true if the provided registry contains one or more
+// subpaths.
+func isNamespaced(registry string) bool {
+	stripped := stripScheme(registry)
```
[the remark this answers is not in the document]
saschagrunert: Yes, makes sense. Changed the validation function to become `validateKey`

pkg/docker/config/config.go
saschagrunert: Yep, not it's just `validateKey`

pkg/docker/config/config.go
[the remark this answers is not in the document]
saschagrunert: Okay, I changed the behavior to not hardly fail in case the registry is namespaced, but debug log an indication that nothing will be modified.

saschagrunert:
Implemented the review suggestions, PTAL @mtrmac @vrothberg

docs/containers-auth.json.5.md
```diff
@@ -24,7 +24,7 @@ is created by a `login` command from a container tool such as `podman login`,
 `buildah login` or `skopeo login`.  Each entry includes the name of the registry and then an auth
 token in the form of a base64 encoded string from the concatenation of the
 username, a colon, and the password. The registry name can additionally contain
-a path or repository name (an image name without tag or digest). The path (or
+a repository name (an image name without tag or digest) and namespaces. The path (or
```
mtrmac, line 27: “entry includes the name of the registry [unrelated text, period]. Registry name can contain (things that are not a registry name)” is very convoluted.

namespaces are parents of repositories, and this wording seems to suggest the opposite or perhaps that their presence is orthogonal.

pkg/docker/config/config.go
```diff
@@ -54,10 +54,18 @@ var (
 
 // SetCredentials stores the username and password in a location
 // appropriate for sys and the users’ configuration.
+// An error will be part of the returned multierror if the provided key is
+// namespaced and a credentials helper other than the AuthenticationFileHelper
+// is used.
```
mtrmac, line 59: This doesn’t tell the caller anything actionable (beyond “on error the `error` return value is set”, which is not typically worth writing out), and right now, as is, can just be dropped.

OTOH the function should document the accepted formats for `key` (and that might be a good place to mention that sub-registry keys might not be supported, depending on users’ configuration).

pkg/docker/config/config.go
```diff
@@ -72,33 +80,45 @@ func SetCredentials(sys *types.SystemContext, registry, username, password strin
...
 				}
 				creds := base64.StdEncoding.EncodeToString([]byte(username + ":" + password))
 				newCreds := dockerAuthConfig{Auth: creds}
-				auths.AuthConfigs[registry] = newCreds
+				auths.AuthConfigs[key] = newCreds
 				return true, nil
 			})
 		// External helpers.
 		default:
 			desc = fmt.Sprintf("credential helper: %s", helper)
```
mtrmac, line 96: (Non-blocking: It might be cleaner to set `desc` closer to the place that actually makes the write, i.e. inside the `!isNamespaced` case.)

pkg/docker/config/config.go
```diff
@@ -72,33 +80,45 @@ func SetCredentials(sys *types.SystemContext, registry, username, password strin
...
 		return desc, nil
 	}
 	return "", multiErr
 }
 
+func unsupportedNamespaceErr(helper string) error {
+	return errors.Errorf("namespaced key is not supported for credential helper %s", helper)
+}
+
 // SetAuthentication stores the username and password in the credential helper or file
```
mtrmac, line 119: Maybe something like
```suggestion
// SetAuthentication stores the username and password in the credential helper or file
// See the documentation of SetCredentials for format of "key"
```

pkg/docker/config/config.go
```diff
@@ -326,9 +346,14 @@ func getAuthenticationWithHomeDir(sys *types.SystemContext, registry, homeDir st
 	return auth.Username, auth.Password, nil
 }
 
-// RemoveAuthentication removes credentials for `registry` from all possible
+// RemoveAuthentication removes credentials for `key` from all possible
 // sources such as credential helpers and auth files.
```
mtrmac, line 350: Document format of `key`, at least linking to `SetCredentials`.

pkg/docker/config/config.go
```diff
@@ -338,17 +363,21 @@ func RemoveAuthentication(sys *types.SystemContext, registry string) error {
...
-		if err == nil {
-			logrus.Debugf("Credentials for %q were deleted from credential helper %s", registry, helper)
-			isLoggedIn = true
-			return
-		}
-		if credentials.IsErrCredentialsNotFoundMessage(err.Error()) {
-			logrus.Debugf("Not logged in to %s with credential helper %s", registry, helper)
-			return
+		if isNamespaced {
+			logrus.Debugf("Not removing credentials: %v", unsupportedNamespaceErr(helper))
```
mtrmac, line 367: `err` is unset on this path, and still appended to `multiErr`. Shouldn’t the function just return?

mtrmac, line 367: Creating and parsing an `error` object is a rather expensive way to construct a message, and more importantly I can’t see any benefit: it just makes it harder to see that the debug line in practice explains the reason.

pkg/docker/config/config.go
```diff
@@ -357,15 +386,15 @@ func RemoveAuthentication(sys *types.SystemContext, registry string) error {
...
+				if innerHelper, exists := auths.CredHelpers[key]; exists {
 					removeFromCredHelper(innerHelper)
 				}
-				if _, ok := auths.AuthConfigs[registry]; ok {
+				if _, ok := auths.AuthConfigs[key]; ok {
 					isLoggedIn = true
-					delete(auths.AuthConfigs, registry)
-				} else if _, ok := auths.AuthConfigs[normalizeRegistry(registry)]; ok {
+					delete(auths.AuthConfigs, key)
+				} else if _, ok := auths.AuthConfigs[normalizeRegistry(key)]; ok {
```
mtrmac, line 395: Doesn’t this turn `logout quay.io/foo` into an unwanted `logout quay.io`?

(Also if we never used `normalizeRegistry` on the write path (or did we? I haven’t checked now), this might be completely unnecessary from the start.)

pkg/docker/config/config.go
```diff
@@ -719,3 +748,15 @@ func normalizeRegistry(registry string) string {
...
+
+// validateKey verifies that the input key does not have any not allowed prefix
+// and returns an indicator if the key is namespaced.
+func validateKey(key string) (isNamespaced bool, err error) {
+	if strings.HasPrefix(key, "http://") || strings.HasPrefix(key, "https://") {
+		return isNamespaced, errors.Errorf("key %s contains http[s]:// prefix", key)
+	}
+
+	// check if the provided key contains one or more subpaths.
+	stripped := stripScheme(key)
```
mtrmac, line 760: This must do nothing after the rejection above.

mtrmac (commented):
So that it isn’t forgotten, we will need a solution for namespaces-only logins, maybe containers/common#659 (comment) .

pkg/docker/config/config.go
saschagrunert: Yep, added docs about the key and dropped the comment on the error.

docs/containers-auth.json.5.md
saschagrunert: Reworked the paragraph a bit.

pkg/docker/config/config.go
saschagrunert: Yes, fixed :+1:

pkg/docker/config/config.go
saschagrunert: Opps, yes I removed that.

pkg/docker/config/config.go
saschagrunert: Yeah I think we can drop this path. I also added a unit test to cover the non removal of `quay.io` if `quay.io/some/path` is being tried to remove.

saschagrunert:
@mtrmac thank you for the review! I addressed your requested changes, PTAL again.

docs/containers-auth.json.5.md
```diff
@@ -21,14 +21,15 @@ Except the primary (read/write) file, other files are read-only, unless the user
...
-`buildah login` or `skopeo login`.  Each entry includes the name of the registry and then an auth
-token in the form of a base64 encoded string from the concatenation of the
-username, a colon, and the password. The registry name can additionally contain
-a path or repository name (an image name without tag or digest). The path (or
-namespace) is matched in its hierarchical order when checking for available
-authentications. For example, an image pull for
-`my-registry.local/namespace/user/image:latest` will result in a lookup in
-`auth.json` in the following order:
+`buildah login` or `skopeo login`. Each entry either contains a single
+hostname (e.g. `docker.io`) or a namespace (e.g. `quay.io/user/image`) as key
```
TomSweeneyRedHat, line 25:
```suggestion
hostname (e.g. `docker.io`) or a namespace (e.g. `quay.io/user/image`) as a key
```

pkg/docker/config/config.go
```diff
@@ -719,3 +747,14 @@ func normalizeRegistry(registry string) string {
 	}
 	return normalized
 }
+
+// validateKey verifies that the input key does not have any not allowed prefix
```
TomSweeneyRedHat, line 751: a nit since you've another
```suggestion
// validateKey verifies that the input key does not have a prefix that is not allowed 
```

TomSweeneyRedHat:
Couple small nits and you need to rebase @saschagrunert

docs/containers-auth.json.5.md
saschagrunert: Fixed :+1:

pkg/docker/config/config.go
saschagrunert: Fixed :+1:

saschagrunert:
Rebased and addressed @TomSweeneyRedHat's comments.

pkg/docker/config/config.go
```diff
@@ -699,18 +727,18 @@ func decodeDockerAuth(conf dockerAuthConfig) (types.DockerAuthConfig, error) {
...
-	}
-
+	stripped := stripScheme(url)
 	nameParts := strings.SplitN(stripped, "/", 2)
-
 	return nameParts[0]
 }
 
+// stripScheme striped the http|https scheme from the provided URL.
+func stripScheme(url string) string {
```
mtrmac, line 736: (Non-blocking: This can be inlined back into its only caller.)

mtrmac (commented):
LGTM. Thanks!

Are you planning to work on https://github.com/containers/common/pull/659#issuecomment-875505807 as well, to make everything fit together?

TomSweeneyRedHat:
LGTM
The skopeo test failed, looks like a flake trying to setup a connection to a registry.

saschagrunert:
> LGTM. Thanks!
> 
> Are you planning to work on [containers/common#659 (comment)](https://github.com/containers/common/pull/659#issuecomment-875505807) as well, to make everything fit together?

Yes, I’ll continue working on it next week. :)

PR #1289, opened 2021-07-02
search: catalog: return full match as first item

vrothberg:
When searching via v2/catalog make sure that a full match is returned as
the first item.  The change is motivated by a conversation in a BZ [1]
on the rather undefined nature/semantics of searching images.

While this change does not fix the BZ, it may be a small improvement
when using the v2/catalog endpoint.

[1] bugzilla.redhat.com/show_bug.cgi?id=1976283

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -423,7 +423,14 @@ func SearchRegistry(ctx context.Context, sys *types.SystemContext, registry, ima
 				res := SearchResult{
 					Name: repo,
 				}
-				searchRes = append(searchRes, res)
+				// bugzilla.redhat.com/show_bug.cgi?id=1976283
+				// If we have a full match, make sure it's listed as the first result.
+				// (Note there might be a full match we never see if we reach the result limit first.)
+				if repo == image {
+					searchRes = append([]SearchResult{res}, searchRes...)
+				} else {
+					searchRes = append(searchRes, res)
+				}
 			}
 		}
 


vrothberg:
@mtrmac WDYT?

rhatdan:
LGTM

vrothberg:
Added a number of commits on top.

docker/docker_client.go
```diff
@@ -423,7 +423,13 @@ func SearchRegistry(ctx context.Context, sys *types.SystemContext, registry, ima
 				res := SearchResult{
 					Name: repo,
 				}
-				searchRes = append(searchRes, res)
+				// bugzilla.redhat.com/show_bug.cgi?id=1976283
```
mtrmac, line 426: (Absolutely non-blocking: I’m not sure the bug number is that relevant for the future; having it in the commit message and this PR might be enough. OTOH it doesn’t really hurt.)

docker/docker_client.go
```diff
@@ -423,7 +423,13 @@ func SearchRegistry(ctx context.Context, sys *types.SystemContext, registry, ima
...
+				// If we have a full match, make sure it's listed as the first result.
```
mtrmac, line 427: Maybe
```suggestion
				// If we have a full match, make sure it's listed as the first result.
				// (Note there might be a full match we never see if we reach the result limit first.)
```

mtrmac (commented):
Thanks, this is a good find and a useful contribution to that discussion.

LGTM, feel free to merge as is or with some comment changes.

---

Note that is only an improvement if we find the exact match during the first `limit` results. So we go from completely-unpredictable to somewhat-more-predictable outcomes, but search is still ultimately not a tool to build reliable processes on.

---

(My personal appetite for heuristically tuning the result ranking is ultimately limited, but this is trivial enough to be clearly worth it, and if someone wanted to work on tuning the heuristics, and the code came with tests, sure, why not.)

docker/docker_client.go
vrothberg: I find bug and issue numbers in code comments incredibly helpful when reading code.  Browsing the git history works but takes much longer.

vrothberg:
Thanks for the review!

> Note that is only an improvement if we find the exact match during the first limit results. So we go from completely-unpredictable to somewhat-more-predictable outcomes, but search is still ultimately not a tool to build reliable processes on.

Yes, that's unfortunate.  I tried tweaking the code a bit and really go through the entire list but afaik that would break `limit`.

docker/docker_client.go
mtrmac: Bug numbers are useful if they contain a larger discussion about finding the root cause, or about trade-offs. In this case I’m not sure what it brings over the comment just below; it is mostly side-tracks.

Anyway, it doesn’t hurt.

mtrmac (approved):
Thanks!

PR #1291, opened 2021-07-05
decrypt blob when get blob for pandora

bellondr:
- decrypt blob when get blob for iaas hybrid cloud pandora

### DIFF
--- a/docker/docker_image_src.go
+++ b/docker/docker_image_src.go
@@ -2,6 +2,9 @@ package docker
 
 import (
 	"context"
+	"crypto/aes"
+	"crypto/cipher"
+	"encoding/hex"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -11,6 +14,8 @@ import (
 	"os"
 	"strconv"
 	"strings"
+
+	"encoding/base64"
 
 	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/image/v5/internal/iolimits"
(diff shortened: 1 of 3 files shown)


docker/docker_image_src.go
```diff
@@ -12,6 +15,8 @@ import (
 	"strconv"
 	"strings"
 
+	"encoding/base64"
```
rhatdan, line 18: Should be in prior section, no blank line before.

pkg/docker/config/config.go
```diff
@@ -499,6 +499,7 @@ func modifyJSON(sys *types.SystemContext, editor func(auths *dockerConfigFile) (
 	if err != nil {
 		return err
 	}
+	fmt.Printf("WARNING storage auth info into %s \n", path)
```
rhatdan, line 502: Is this a debugging message?  If not then this should be in logrus.

docker/docker_image_src.go
```diff
@@ -294,8 +307,91 @@ func (s *dockerImageSource) GetBlob(ctx context.Context, info types.BlobInfo, ca
...
+	d := DecryptReaderCloser{
+		source: source,
+		stream: stream,
+	}
+	return &d, nil
+}
+
+// FillBuffer fills the given buffer with as many bytes from the reader as possible. It returns
+// EOF if an EOF was encountered or any other error.
+func FillBuffer(reader io.Reader, buffer []byte) (int, error) {
```
rhatdan, line 389: Should this be an external function?

mtrmac (changes requested):
NAK.

AFAICS this provides absolutely no new security properties, does clearly worse than just using TLS, and the implementation quality honestly makes accepting any kind of security PR hard to contemplate.

There’s also no motivation (what is “pandora”), no explanation of the design (should this be a transport-specific thing or something general? why isn’t the experimental `containers/ocicrypt` support useful for the use case?).

Frankly I’m very tempted to close this PR outright, possibly even with prejudice against any future submissions that have any security exposure at all.

But let’s talk about it first. What does this do and why should it exist?

mtrmac:
> But let’s talk about it first. What does this do and why should it exist?

BTW given the risks, it’s not at all tempting for c/image to design or implement a new crypto system from raw low-level elements. Isn’t there an already well-established industry standard directly applicable to the problem space?

Issue #1290, opened 2021-07-05
Old config file location ~/.dockercfg overwrites newer path at ~/.docker/config.json


rhuss:
An old Docker config location `~/.dockercfg` takes precedence over the newer one at `~/.docker/config.json`. The order should be reversed.

I noted this while debugging an issue related to #111 and I had accidentally both configuration options (with different content) in my home directory.

mtrmac:
Can you provide a reproducer, please? The `~/.docker/config.json` one is preferred AFAICS, and we have a test for that (the one that broke in the current version of #1292 :) )

PR #1292, opened 2021-07-07
 We should only look at Legacy path if DOCKER_CONFIG is not set. 

rhatdan:

### DIFF
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -204,10 +204,10 @@ func getAuthFilePaths(sys *types.SystemContext, homeDir string) []authPath {
 		paths = append(paths,
 			authPath{path: filepath.Join(homeDir, dockerHomePath), legacyFormat: false},
 		)
+		paths = append(paths,
+			authPath{path: filepath.Join(homeDir, dockerLegacyHomePath), legacyFormat: true},
+		)
 	}
-	paths = append(paths,
-		authPath{path: filepath.Join(homeDir, dockerLegacyHomePath), legacyFormat: true},
-	)
 	return paths
 }
 


rhatdan:
@rhuss PTAL
@vrothberg @mtrmac PTAL

pkg/docker/config/config.go
```diff
@@ -201,13 +201,13 @@ func getAuthFilePaths(sys *types.SystemContext, homeDir string) []authPath {
 			authPath{path: filepath.Join(dockerConfig, "config.json"), legacyFormat: false},
 		)
 	} else {
+		paths = append(paths,
+			authPath{path: filepath.Join(homeDir, dockerLegacyHomePath), legacyFormat: true},
+		)
```
rhuss, line 206: So is it the last path entry 'wins', right (i.e. the latter overwrites a former entry if both are present in the filesystem). Then the change looks good to me.

rhatdan: I actually can not see where last entry wins.  BTW, on a quick look.  It looks to me the first one with a username/password would win.  Although I am not an expert on this code.

mtrmac (commented):
> We should only look at Legacy path if DOCKER_HOST is not set.

- That variable is not used
- (That’s **EDIT not** the only thing this PR is changing. Please document your intent.)
- Why? docker/cli AFAICS does not have that condition.

pkg/docker/config/config.go
mtrmac: Yes, the first entry wins. Which makes this PR hard to reconcile with the original bug report.

mtrmac (changes requested):
[See before]

pkg/docker/config/config.go
mtrmac:
```go
// getAuthFilePaths returns a slice of authPaths based on the system context
// in the order they should be searched. 
```
and `TestGetAuthPreferNewConfig`.

rhatdan: Switched to use legacy last.

rhatdan:
Sorry I meant DOCKER_CONFIG, now fixed commit message.

mtrmac:
The “why” question still remains; this now has no clear relationship to #1290 , and does something else than the original we are trying go be compatible with.

pkg/docker/config/config.go
rhuss: Sorry that my issue description is so vague.  Unfortunately, I don't have a reproducer anymore, but my problem was solved after I have removed the legacy `~/.dockercfg` that I still had parallel to `~/.docker/config.json`. I will restore it from my backup and will add a more in-depth description of the issue.

rhatdan:
@mtrmac Reworded the PR description as well as the github description.  Even though this does not fix the original problem, it still fixes a potential bug.

mtrmac:
> * Why? docker/cli AFAICS does not have that condition.

> The “why” question still remains; this now has no clear relationship to #1290 , and does something else than the original we are trying go be compatible with.

Why is the current behavior a bug?

rhatdan:
If I set DOCKER_CONFIG and the DOCKER_CONFIG file is not set, but the legacy path is set, then I believe unexpected results happen, if that is ok with you, then the non legacy path should also succeed, and it does not.  We should not treat legacy path special.

mtrmac:
“Unexpected results” by what measure? All of this exists to be compatible to a pre-existing implementation, isn’t that the thing to measure against?

rhatdan:
Unexpected by the user, I would expect to fail if I set DOCKER_CONFIG=/tmp/foobar and it did not exist, but it would succed if the legacy path exists.

But If you don't want this then close it. Most likely would never happen, but I believe this was a mistake when the code was written.

mtrmac:
> Unexpected by the user, I would expect to fail if I set DOCKER_CONFIG=/tmp/foobar and it did not exist, but it would succed if the legacy path exists.

Most importantly, I just don’t think we are in the business of defining of what `DOCKER_CONFIG` should mean. We are reading that to try to do exactly what `/usr/bin/docker` did.

(To be explicit, note that ~/.dockercfg is _not_ in `$DOCKER_CONFIG` which is ~/.docker/`. And taking this thought further, should setting `DOCKER_CONFIG` prevent use of `~/.containers/auth.json`? If not, why is `~/.dockercfg` different?)

> But If you don't want this then close it. Most likely would never happen, but I believe this was a mistake when the code was written.

AFAICS, looking at docker/cli, the current semantics is as expected. If you see some other docker/ code with different behavior, or if I’m reading it incorrectly, please reopen.

rhatdan:
Well then I would say that Docker had broken code...

mtrmac:
That’s not at all obvious to me (`DOCKER_CONFIG` is not _exclusively_ for `config.json`, e.g. the directory is used for TLS secrets of the daemon, `cli-plugins`, Notary, probably more), but even if it were true, I don’t think that’s very relevant: if we aren’t bug-for-bug compatible, we aren’t compatible. Defining a new semantics for `~/.dockercfg` that nothing writes to for the last 6 years (compare https://issues.redhat.com/browse/RUN-1246?focusedCommentId=16344129&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16344129 ) doesn’t help.

rhatdan:
Then we are not compatible since we have fixed other "docker" bugs.

mtrmac:
A good reason to un-fix them, maybe :)

rhatdan:
Nope we will be compatible to a point, but when they are clearly in error, especially ones that modify content of the host outside of the container, then we will not be compatible.

PR #1294, opened 2021-07-07
Add support for decompressing while copying to dir://

moio:
This is the accompanying PR for https://github.com/containers/skopeo/pull/1365

### DIFF
--- a/directory/directory_dest.go
+++ b/directory/directory_dest.go
@@ -21,14 +21,21 @@ const version = "Directory Transport Version: 1.1\n"
 var ErrNotContainerImageDir = errors.New("not a containers image directory, don't want to overwrite important data")
 
 type dirImageDestination struct {
-	ref        dirReference
-	compress   bool
-	decompress bool
+	ref                     dirReference
+	desiredLayerCompression types.LayerCompression
 }
 
 // newImageDestination returns an ImageDestination for writing to a directory.
-func newImageDestination(ref dirReference, compress bool, decompress bool) (types.ImageDestination, error) {
-	d := &dirImageDestination{ref: ref, compress: compress, decompress: decompress}
+func newImageDestination(sys *types.SystemContext, ref dirReference) (types.ImageDestination, error) {
+	desiredLayerCompression := types.PreserveOriginal
+	if sys != nil {
+		if sys.DirForceCompress {
+			desiredLayerCompression = types.Compress
+		} else if sys.DirForceDecompress {
+			desiredLayerCompression = types.Decompress
+		}
+	}
+	d := &dirImageDestination{ref: ref, desiredLayerCompression: desiredLayerCompression}
 
 	// If directory exists check if it is empty
 	// if not empty, check whether the contents match that of a container image directory and overwrite the contents
@@ -102,13 +109,7 @@ func (d *dirImageDestination) SupportsSignatures(ctx context.Context) error {
 }
 
 func (d *dirImageDestination) DesiredLayerCompression() types.LayerCompression {
-	if d.compress {
-		return types.Compress
-	}
-	if d.decompress {
-		return types.Decompress
-	}
-	return types.PreserveOriginal
+	return d.desiredLayerCompression
 }
 
 // AcceptsForeignLayerURLs returns false iff foreign layers in manifest should be actually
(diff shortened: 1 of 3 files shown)


types/types.go
```diff
@@ -633,6 +633,8 @@ type SystemContext struct {
 	// === dir.Transport overrides ===
 	// DirForceCompress compresses the image layers if set to true
 	DirForceCompress bool
+	// DirForceDecompress decompresses the image layers if set to true
+	DirForceDecompress bool
```
mtrmac, line 637: Cc: @vrothberg
- Should this be just a `dir:` feature or a generic `c/image/copy.Options` option? I don’t have much of an idea whether this is likely to be used elsewhere. It doesn’t _hurt_ here, doing it requires much less thinking about designing the `copy.Options` value _well_ (interaction of that new option with `DesiredLayerCompression` would be a concern), and this is cheap to implement/maintain even if we later decided to make this a `copy` feature, so I’m leaning towards this `dir:`-specific approach — but if there were already a good reason to make this a `copy` feature _now_, please speak up.
- RFC: I think I’d prefer a `DirDesiredLayerCompression *types.LayerCompression`, with the default `nil` meaning “unspecified”, and document that `DirForceCompress` is ignored (forbidden?) if `DirDesiredLayerCompression` is specified. That would naturally extend to future values of `LayerCompression`, if any, and it avoids the question of “what does `DirForceCompress ==true && DirForceDecompress == true` mean?”. I don’t feel too strongly about this, we can just reject the true && true combination instead.

directory/directory_transport.go
```diff
@@ -157,7 +157,11 @@ func (ref dirReference) NewImageDestination(ctx context.Context, sys *types.Syst
 	if sys != nil {
 		compress = sys.DirForceCompress
 	}
-	return newImageDestination(ref, compress)
+	decompress := false
+	if sys != nil {
+		decompress = sys.DirForceDecompress
+	}
+	return newImageDestination(ref, compress, decompress)
```
mtrmac, line 164: This is fine, but at this point it would be more consistent with other transports and cleaner, to pass all of `sys` down to `newImageDestination(sys, ref)`;  `dirReference` doesn’t have to care about the compression options.

directory/directory_dest.go
```diff
@@ -21,13 +21,14 @@ const version = "Directory Transport Version: 1.1\n"
 var ErrNotContainerImageDir = errors.New("not a containers image directory, don't want to overwrite important data")
 
 type dirImageDestination struct {
-	ref      dirReference
-	compress bool
+	ref        dirReference
+	compress   bool
+	decompress bool
```
mtrmac, line 26: Non-blocking: Just store a single `desiredLayerCompression types.LayerCompression` here, and compute it in the constructor; that eliminates the `compress == true && decompress == true` concern (and works however we decide to define the `SystemContext` options).

mtrmac (commented):
Thanks!

A design concern I’d like to have more opinion on, + some not-quite-blocking implementation suggestions.

mtrmac:
Looking at the failing tests, please rebase your branch on top of latest `main`; hopefully that will help.

directory/directory_transport.go
mtrmac, line 164: `compress && decompress` should fail with an error, compare https://github.com/containers/skopeo/pull/1365#discussion_r665677728 .

types/types.go
vrothberg: > but if there were already a good reason to make this a copy feature now, please speak up.

I can only think of one reasons: determinism.  Copying decompressed layers around would always yield the same result while compressing layers may yield different results depending on the (version of the) tools and the tool chains.

Maybe via `--dest-compress-level=0`?

mtrmac: `--dest-compress-level` does not have globally defined semantics, and there are no reserved values; the Gzip compressor already recognizes the `0` value (as “store in the Gzip envelope with no compression applied” it seems).

mtrmac: Deterministically storing ~twice as much data on a registry… I don’t know. A hypothetical auditor can check RootFS.DiffIDs in most cases.

moio: Do I understand correctly that this PR could in principle be merged with the current concept, the assumption being that a follow-up PR might always happen at a later point to extend the functionality to other protocols?

If so I can start working at the other comments here.

mtrmac: Yeah, let’s go with the `dir:`-only approach for now.

moio:
Hopefully it's all good now

rhatdan:
LGTM
@vrothberg @mtrmac PTAL

mtrmac (approved):
LGTM. Thanks!

PR #1302, opened 2021-07-13
docs: fix indentation in registries.conf

vrothberg:
Fix a number of indentation issues in the containers-registries.conf man
page which caused rendering issues both in the man pages and the
upstream markdown on GitHub; move all to the root indent level/scope.

### DIFF
--- a/docs/containers-registries.conf.5.md
+++ b/docs/containers-registries.conf.5.md
@@ -39,8 +39,8 @@ Given an image name, a single `[[registry]]` TOML table is chosen based on its `
 The user-specified image name must start with the specified `prefix` (and continue
 with the appropriate separator) for a particular `[[registry]]` TOML table to be
 considered; (only) the TOML table with the longest match is used. It can
-also include wildcarded subdomains in the format `*.example.com` along as mentioned
-above. The wildcard should only be present at the beginning as shown in the formats
+also include wildcarded subdomains in the format `*.example.com`.
+The wildcard should only be present at the beginning as shown in the formats
 above. Other cases will not work. For example, `*.example.com` is valid but
 `example.*.com`, `*.example.com/foo` and `*.example.com:5000/foo/bar:baz` are not.
 
--- a/docs/containers-registries.conf.5.md
+++ b/docs/containers-registries.conf.5.md
@@ -36,28 +36,28 @@ Given an image name, a single `[[registry]]` TOML table is chosen based on its `
   - _host_[`:`_port_]`/`_namespace_[`/`_namespace_…]`/`_repo_(`:`_tag|`@`_digest_)
   - [`*.`]_host_
 
-    The user-specified image name must start with the specified `prefix` (and continue
-    with the appropriate separator) for a particular `[[registry]]` TOML table to be
-    considered; (only) the TOML table with the longest match is used. It can
-    also include wildcarded subdomains in the format `*.example.com` along as mentioned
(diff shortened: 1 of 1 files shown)


vrothberg:
@lsm5 @mtrmac @rhatdan PTAL

mtrmac (commented):
TL;DR: Stalling.

My initial impression is that the result of this is a long slog of a “Remapping and mirroring registries” paragraph with no structure. The old version was _clearly wrong_ and yet ended up easier to scan. (And the raw markdown happens to be better than either of them, right now…)

It doesn’t help that go-md2man and/or its parser seems somewhat selective in what/how it supports. Just replacing the 4 leading spaces with 3 leading spaces avoids the code blocks, but loses the block structure all the same. In principle, Markdown does support multi-paragraph list items, so it’s tempting to just fix the tools to detect that properly, and to generate `.TP`/`.IP` as appropriate. But that’s also a rathole…

(Pointless rant: I just don’t see the point of Markdown at all. If you want an ASCII language with control over structure and nesting, use HTML; if you want something easier to type, use a rich text editor. What good does \*obscure\_syntax\-with\(precise\)\[sequences\] of unintuitive characters do?

Honestly as a way to write _man pages_, it’s very tempting to just write raw nroff. The syntax isn’t any more obscure than the Markdown, just see all the corner cases and lax syntax in https://github.github.com/gfm/ . But then we wouldn’t have a web page to point to…)

vrothberg:
@mtrmac, I am not sure how to proceed.  IMHO, it's an improvement over the current state at no cost.

mtrmac:
At this point I’d rather have the code blocks than no structure, though that’s an _extremely_ weak preference.

Primarily, give me a few days to look into those Markdown parsers. Perhaps we could actually achieve the right thing, and improve the ecosystem for others.

rhatdan:
LGTM

TomSweeneyRedHat:
@vrothberg needs a rebase.
FWIW, I have no idea how I managed to make a milestone, sorry!

docs/containers-registries.conf.5.md
```diff
@@ -36,28 +36,28 @@ Given an image name, a single `[[registry]]` TOML table is chosen based on its `
...
-    with the appropriate separator) for a particular `[[registry]]` TOML table to be
-    considered; (only) the TOML table with the longest match is used. It can
-    also include wildcarded subdomains in the format `*.example.com` along as mentioned
-    above. The wildcard should only be present at the beginning as shown in the formats
-    above. Other cases will not work. For example, `*.example.com` is valid but
-    `example.*.com`, `*.example.com/foo` and `*.example.com:5000/foo/bar:baz` are not.
+The user-specified image name must start with the specified `prefix` (and continue
+with the appropriate separator) for a particular `[[registry]]` TOML table to be
+considered; (only) the TOML table with the longest match is used. It can
+also include wildcarded subdomains in the format `*.example.com` along as mentioned
```
TomSweeneyRedHat, line 42: I would just drop "along as mentioned above".  I'm not sure what you are trying to say here or what you're trying to connect to.

TomSweeneyRedHat: And yeah, I realized it's pre-existing.

vrothberg: Done

vrothberg:
Rebased

rhatdan:
I am going to merge, since this is an improvement.  If Miloslav finds a better way to format the man page, we can do this in a separate PR.

PR #96, opened 2016-09-20
Be benevolent to .docker/config.json file

miminar:
Docker can store auth entries with schema prefix or even path suffix.
See an example:

    {
        "auths": {
            "10.3.10.88:5000": {
                ...
            },
            "http://10.3.10.88:5000/v2/": {
                ...
            },
            "https://10.3.10.88:5000": {
                ...
            },
            "https://index.docker.io/v1/": {
                ...
            }
        }
    }

The entries were created using command `docker login` of upstream Docker
1.12. Let's normalize the auth keys before trying to match against
hostname.

Signed-off-by: Michal Minář <miminar@redhat.com>

PR #1304, opened 2021-07-19
Disable registry fallback for repository credentials

saschagrunert:
As mentioned in containers/common#659 (comment), we now do not normalize the registry when being in non legacy format. This means we do not match "quay.io/foo" if "quay.io/bar" is provided as credentials, because both would be normalized to "quay.io" before this patch.

We also add a new corner case where auth.json has been external modified to contain an entry like "https://quay.io/v1/", which now also matches if "quay.io/foo" has been provided via the credentials.

### DIFF
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -665,10 +665,10 @@ func findAuthentication(ref reference.Named, registry, path string, legacyFormat
 	// those entries even in non-legacyFormat ~/.docker/config.json.
 	// The docker.io registry still uses the /v1/ key with a special host name,
 	// so account for that as well.
-	registry = normalizeRegistry(registry)
+	registry = normalizeAuthFileKey(registry, legacyFormat)
 	normalizedAuths := map[string]dockerAuthConfig{}
 	for k, v := range auths.AuthConfigs {
-		normalizedAuths[normalizeRegistry(k)] = v
+		normalizedAuths[normalizeAuthFileKey(k, legacyFormat)] = v
 	}
 
 	if val, exists := normalizedAuths[registry]; exists {
@@ -723,29 +723,27 @@ func decodeDockerAuth(conf dockerAuthConfig) (types.DockerAuthConfig, error) {
 	}, nil
 }
 
-// convertToHostname converts a registry url which has http|https prepended
-// to just an hostname.
-// Copied from github.com/docker/docker/registry/auth.go
-func convertToHostname(url string) string {
-	stripped := stripScheme(url)
-	nameParts := strings.SplitN(stripped, "/", 2)
-	return nameParts[0]
-}
-
-// stripScheme striped the http|https scheme from the provided URL.
-func stripScheme(url string) string {
-	stripped := strings.TrimPrefix(url, "http://")
+// normalizeAuthFileKey takes a key, converts it to a host name and normalizes
+// the resulting registry.
+func normalizeAuthFileKey(key string, legacyFormat bool) string {
+	stripped := strings.TrimPrefix(key, "http://")
 	stripped = strings.TrimPrefix(stripped, "https://")
-	return stripped
+
+	if legacyFormat || stripped != key {
+		stripped = strings.SplitN(stripped, "/", 2)[0]
+	}
+
+	return normalizeRegistry(stripped)
 }
 
+// normalizeRegistry converts the provided registry if a known docker.io host
+// is provided.
 func normalizeRegistry(registry string) string {
-	normalized := convertToHostname(registry)
-	switch normalized {
+	switch registry {
 	case "registry-1.docker.io", "docker.io":
 		return "index.docker.io"
 	}
-	return normalized
+	return registry
 }
 
 // validateKey verifies that the input key does not have a prefix that is not
--- a/pkg/docker/config/config_test.go
+++ b/pkg/docker/config/config_test.go
@@ -886,3 +886,94 @@ func TestValidateKey(t *testing.T) {
 		assert.Equal(t, tc.isNamespaced, isNamespaced)
 	}
 }
+
+func TestSetGetCredentials(t *testing.T) {
+	const (
+		username = "username"
+		password = "password"
+	)
+
+	tmpDir, err := ioutil.TempDir("", "auth-test-")
+	require.NoError(t, err)
+	defer os.RemoveAll(tmpDir)
+
+	for _, tc := range []struct {
+		name         string
+		set          string
+		get          string
+		useLegacyAPI bool
+		shouldAuth   bool
+	}{
+		{
+			name:       "Should match namespace",
+			set:        "quay.io/foo",
+			get:        "quay.io/foo/a",
+			shouldAuth: true,
+		},
+		{
+			name:       "Should match registry if repository provided",
+			set:        "quay.io",
+			get:        "quay.io/foo",
+			shouldAuth: true,
+		},
+		{
+			name:       "Should not match different repository",
+			set:        "quay.io/foo",
+			get:        "quay.io/bar",
+			shouldAuth: false,
+		},
+		{
+			name:       "Should match legacy registry entry (new API)",
+			set:        "https://quay.io/v1/",
+			get:        "quay.io/foo",
+			shouldAuth: true,
+		},
+		{
+			name:         "Should match legacy registry entry (legacy API)",
+			set:          "https://quay.io/v1/",
+			get:          "quay.io",
+			shouldAuth:   true,
+			useLegacyAPI: true,
+		},
+	} {
+
+		// Create a new empty SystemContext referring an empty auth.json
+		tmpFile, err := ioutil.TempFile("", "auth.json-")
+		require.NoError(t, err)
+		defer os.RemoveAll(tmpFile.Name())
+
+		sys := &types.SystemContext{}
+		if tc.useLegacyAPI {
+			sys.LegacyFormatAuthFilePath = tmpFile.Name()
+			_, err = tmpFile.WriteString(fmt.Sprintf(
+				`{"%s":{"auth":"dXNlcm5hbWU6cGFzc3dvcmQ="}}`, tc.set,
+			))
+		} else {
+			sys.AuthFilePath = tmpFile.Name()
+			_, err = tmpFile.WriteString(fmt.Sprintf(
+				`{"auths":{"%s":{"auth":"dXNlcm5hbWU6cGFzc3dvcmQ="}}}`, tc.set,
+			))
+		}
+		require.NoError(t, err)
+
+		// Try to authenticate against them
+		var auth types.DockerAuthConfig
+		if !tc.useLegacyAPI {
+			ref, err := reference.ParseNamed(tc.get)
+			require.NoError(t, err)
+			auth, err = getCredentialsWithHomeDir(sys, ref, reference.Domain(ref), tmpDir)
+			require.NoError(t, err)
+		} else {
+			auth, err = getCredentialsWithHomeDir(sys, nil, tc.get, tmpDir)
+			require.NoError(t, err)
+		}
+
+		if tc.shouldAuth {
+			assert.Equal(t, username, auth.Username, tc.name)
+			assert.Equal(t, password, auth.Password, tc.name)
+		} else {
+			assert.Empty(t, auth.Username, tc.name)
+			assert.Empty(t, auth.Password, tc.name)
+		}
+	}
+}


pkg/docker/config/config.go
```diff
@@ -697,7 +696,15 @@ func authKeysForRef(ref reference.Named) (res []string) {
 		name = name[:lastSlash]
 	}
 
-	return res
+	// Named references always contain at least one "/", so in theory we never
+	// return nil in this function. We still preserve this sanity check to
+	// avoid accessing out of bounds in case if future reference
```
TomSweeneyRedHat, line 701: nit of a nit if you've other changes.
```suggestion
	// avoid accessing out of bounds in case of future reference
```

TomSweeneyRedHat:
LGTM

pkg/docker/config/config.go
```diff
@@ -697,7 +696,15 @@ func authKeysForRef(ref reference.Named) (res []string) {
...
+	// implementations change.
+	if len(res) > 1 {
+		return res[:len(res)-1]
```
mtrmac, line 704: The code could just not add such a value in the first place, instead of stripping it afterwards.

mtrmac (changes requested):
I may be missing something, but this doesn’t look right at all.

A `login quay.io; pull quay.io/foo` should find the registry-level credentials; so dropping the registry name from `authKeysForRef` does not make sense to me. (And also I can’t see that it makes any difference, because the `normalizedAuths` lookup will find `quay.io` anyway!)

The originally-reported problem was that a `login quay.io/foo; pull quay.io/bar` finds the `foo` credentials, and shouldn’t. AFAICS the `normalizedAuths` code will find that both before and after this PR.

And for completeness, `login https://quay.io/v1/; pull quay.io/foo` (with the login done not using our current/future tools, but something that writes such keys into `auth.json` directly) should continue to find the URL-keyed credential.

The original proposal talked about how this is governed by the format of the key in the config file, not by format of the query. So that’s another reason why changing `authKeysForRef`, but not changing the `keys = []string{registry}` is surprising.

Am I missing something about how this works?

Either way this is a bit tricky, so could you add tests for the three cases above to one of the `GetAuth*` functions, please?

saschagrunert:
@mtrmac I think I misunderstood what was required but now added the unit tests and aligned the current implementation to it. Sorry for that, please take a look again.

pkg/docker/config/config_test.go
```diff
@@ -886,3 +886,92 @@ func TestValidateKey(t *testing.T) {
...
+		auth, err := GetCredentialsForRef(sys, ref)
+		require.NoError(t, err)
+		return auth
+	}
+
+	// Unset the home dir to make the test not fallback to other auth files
+	// there.
+	tmpDir, err := ioutil.TempDir("", "auth-test-")
+	require.NoError(t, err)
+	os.Setenv("HOME", tmpDir)
```
mtrmac, line 908: Doesn’t this have to be restored after the test? Also the comment above says “unset”, not “change”.

Either way, wouldn’t using `getCredentialsWithHomeDir` allow testing this without affecting $HOME at all?

pkg/docker/config/config_test.go
```diff
@@ -886,3 +886,92 @@ func TestValidateKey(t *testing.T) {
...
+	defer os.RemoveAll(tmpDir)
+
+	for _, tc := range []struct {
+		name       string
+		set        string
+		get        string
+		setByAPI   bool
```
mtrmac, line 915: Absolutely non-blocking: I’m not sure we need this; just testing the lookup finds the right entries in a fixture might test the lookup path enough. Compare `TestGetAuth`.

Alternatively test both a fixture, and the set/lookup API round-trip.

But, sure, these tests are already infinitely better than the previous no tests.

pkg/docker/config/config.go
```diff
@@ -739,13 +752,27 @@ func stripScheme(url string) string {
...
 func normalizeRegistry(registry string) string {
 	normalized := convertToHostname(registry)
-	switch normalized {
+	return convertDockerHost(normalized)
+}
+
+// stripRegistry removes known prefixes as well as suffixes from the registry.
+func stripRegistry(registry string) string {
+	stripped := stripScheme(registry)
+	stripped = strings.TrimSuffix(stripped, "/v1/")
```
mtrmac, line 764: AFAICS this matches `quay.io/v1/` and `quay.io/foo/v1/`, differentiating from scoped keys `quay.io/v1` and `quay.io/foo/v1` only using the trailing slash.

And historically `normalizeRegistry` has matched arbitrary URLs, including `https://example.com/an/ignored/comment`, so relying on a single trailing slash feels rather brittle.

Why not normalize only keys that start with a scheme`://` (and then probably ignore the path part completely)?

pkg/docker/config/config.go
```diff
@@ -649,30 +649,43 @@ func findAuthentication(ref reference.Named, registry, path string, legacyFormat
...
+		strippedAuths[stripRegistry(k)] = v
+	}
+
 	// Repo or namespace keys are only supported as exact matches. For registry
 	// keys we prefer exact matches as well.
 	for _, key := range keys {
 		if val, exists := auths.AuthConfigs[key]; exists {
 			return decodeDockerAuth(val)
 		}
+		if val, exists := strippedAuths[key]; exists {
```
mtrmac, line 666: The comment above talks about exact matches, while this adds a non-exact match.

I’d expect only the “bad luck” code below to need changing.

pkg/docker/config/config.go
```diff
@@ -649,30 +649,43 @@ func findAuthentication(ref reference.Named, registry, path string, legacyFormat
...
-	// several years, and matched registry hostnames against that, so support
-	// those entries even in non-legacyFormat ~/.docker/config.json.
-	// The docker.io registry still uses the /v1/ key with a special host name,
-	// so account for that as well.
-	registry = normalizeRegistry(registry)
-	normalizedAuths := map[string]dockerAuthConfig{}
-	for k, v := range auths.AuthConfigs {
-		normalizedAuths[normalizeRegistry(k)] = v
-	}
+	if legacyFormat || ref == nil {
```
mtrmac, line 671: The `https://index.docker.io/v1/` form is used by Docker in non-legacy formats, and must be found if `ref != nil`. (Another test case?)

https://github.com/containers/common/pull/659#issuecomment-875505807 talks about conditional normalization based on `legacyFormat` for the scheme-less `quay.io/foo` keys, but AFAICS we must normalize the URL keys in all cases.

pkg/docker/config/config.go
```diff
@@ -739,13 +752,27 @@ func stripScheme(url string) string {
 	return stripped
 }
 
+// normalizeRegistry converts the provided registry to contain only the hostname.
 func normalizeRegistry(registry string) string {
```
mtrmac, line 756: The input is not a registry; and the output can be a host:port, not just a hostname.

(Yes, the function is already misnamed, but let’s not double down.)

Maybe
- `normalizeRegistry`+`stripRegistry` → `normalizeAuthFileKey(key, legacyFormat)` (I don’t think we need two different ones)
- `convertDockerHost` → `normalizeRegistry`

pkg/docker/config/config.go
```diff
@@ -739,13 +752,27 @@ func stripScheme(url string) string {
...
 	normalized := convertToHostname(registry)
-	switch normalized {
+	return convertDockerHost(normalized)
+}
+
+// stripRegistry removes known prefixes as well as suffixes from the registry.
+func stripRegistry(registry string) string {
```
mtrmac, line 762: So the input is not a registry, otherwise there would be nothing to strip.

pkg/docker/config/config.go
```diff
@@ -649,30 +649,43 @@ func findAuthentication(ref reference.Named, registry, path string, legacyFormat
...
+	if legacyFormat || ref == nil {
+		// bad luck; let's normalize the entries first
+		// This primarily happens for legacyFormat, which for a time used API URLs
+		// (http[s:]//…/v1/) as keys.
+		// Secondarily, (docker login) accepted URLs with no normalization for
+		// several years, and matched registry hostnames against that, so support
+		// those entries even in non-legacyFormat ~/.docker/config.json.
+		// The docker.io registry still uses the /v1/ key with a special host name,
+		// so account for that as well.
+		registry = normalizeRegistry(registry)
```
mtrmac, line 680: BTW I think the key we are looking up only needs to go through the (current) `convertDockerHost`, `registry` should already be a host[port], never an URL.

mtrmac (commented):
Thanks, this looks closer.

I’m not sure whether some of the design decisions different from https://github.com/containers/common/pull/659#issuecomment-875505807 are accidents or intentional because that proposal is broken; if that’s the case, please write down the rationale, at least here in the PR.

pkg/docker/config/config_test.go
saschagrunert: Yes, switched to use `getCredentialsWithHomeDir`

pkg/docker/config/config_test.go
saschagrunert: Removed that one and always stick to create a file by coinstructing a plain string.

pkg/docker/config/config.go
saschagrunert: Alright, switched to the conversion approach which makes the whole function obsolete.

pkg/docker/config/config.go
saschagrunert: Removed that code part.

pkg/docker/config/config.go
saschagrunert: Removed the `if`

pkg/docker/config/config.go
saschagrunert: Renamed the function and changed the docs.

pkg/docker/config/config.go
saschagrunert: This function does not exist any more.

pkg/docker/config/config.go
saschagrunert: Yep, changed as suggetest.

saschagrunert:
@mtrmac PTAL, I think I addressed all of your comments.

pkg/docker/config/config.go
```diff
@@ -739,13 +739,24 @@ func stripScheme(url string) string {
 	return stripped
 }
 
+// normalizeAuthFileKey takes a key, converts it to a host name and normalizes
+// the resulting registry.
+func normalizeAuthFileKey(key string, legacyFormat bool) string {
+	stripped := stripScheme(key)
+	if legacyFormat || stripped != key {
+		stripped = convertToHostname(key)
```
mtrmac, line 747: This is the only caller of `convertToHostname`, and we have already stripped the scheme, only for `convertToHostname` to try doing it again. If `convertToHostname` were inlined, we could remove that redundancy.

(And then, absolutely optionally, inline `stripScheme` into the only caller.)

mtrmac (commented):
Thanks! Looks good, some possible cleanups.

pkg/docker/config/config.go
```diff
@@ -665,10 +665,10 @@ func findAuthentication(ref reference.Named, registry, path string, legacyFormat
 	// those entries even in non-legacyFormat ~/.docker/config.json.
 	// The docker.io registry still uses the /v1/ key with a special host name,
 	// so account for that as well.
-	registry = normalizeRegistry(registry)
+	registry = normalizeAuthFileKey(registry, legacyFormat)
```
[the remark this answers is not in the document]
saschagrunert: This would result failing of the test case `TestGetAuthFromLegacyFile/ignore_schema_and_path`

pkg/docker/config/config.go
saschagrunert: Yep makes sense, fixed as suggested :+1:

pkg/docker/config/config_test.go
```diff
@@ -886,3 +886,96 @@ func TestValidateKey(t *testing.T) {
...
+	const (
+		username = "username"
+		password = "password"
+	)
+
+	tmpDir, err := ioutil.TempDir("", "auth-test-")
+	require.NoError(t, err)
+	defer os.RemoveAll(tmpDir)
+
+	getAuth := func(sys *types.SystemContext, input string, legacyAPI bool) (auth types.DockerAuthConfig) {
```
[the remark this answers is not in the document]
saschagrunert: Fixed :+1:

pkg/docker/config/config.go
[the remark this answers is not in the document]
mtrmac: Thanks for pointing that out.

AFAICS those tests look wrong from the start: they were added in https://github.com/containers/image/pull/96 , at a time where the only caller to `docker.getAuth` only submitted a host[:port] — so it was testing an impossible state at the time.

But that’s not something _this PR_ should be dealing with.

mtrmac (approved):
LGTM.

pkg/docker/config/config.go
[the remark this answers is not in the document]
mtrmac: Filed #1318, for the record.

PR #1307, opened 2021-07-23
Use quay.io in docker unit tests instead of docker.io

mtrmac:
... to avoid docker.io pull limits.

`NewImageSource` and `NewImage` actually access the remote server (because `NewImageSource` is looking for a working mirror); update `NewImageDestination` basically for symmetry.

### DIFF
--- a/docker/docker_transport_test.go
+++ b/docker/docker_transport_test.go
@@ -153,7 +153,7 @@ func TestReferencePolicyConfigurationNamespaces(t *testing.T) {
 }
 
 func TestReferenceNewImage(t *testing.T) {
-	ref, err := ParseReference("//busybox")
+	ref, err := ParseReference("//quay.io/libpod/busybox")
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

cevich:
LGTM also.  Those images we copied to quay are used by nearly all of our CI everywhere and everybody understands not to touch them :D

The test-failures here will go away after we synchronize the "skopeo_test" task to the new c/skopeo version...

cevich:
...rebase and this will fix the skopeo tests: https://github.com/containers/image/pull/1309

TomSweeneyRedHat:
LGTM
Looks like the skopeo test is unhappy, but it's calling for a rebase.  Perhaps the rebase will pick that up?

cevich:
> Perhaps the rebase will pick that up?

Ya it's a known issue, rebase will indeed fix it.

PR #1308, opened 2021-07-23
Don't run skopeo's runner.sh validate

mtrmac:
It was removed in https://github.com/containers/skopeo/pull/1374 .

Fixes a failure visible in https://github.com/containers/image/pull/1307 .


cevich:
@mtrmac this is correct fix, but i would actually prefer https://github.com/containers/image/pull/1309 because it keeps the blocks 100% identical.  Only the prefix comment is different between the two.

mtrmac:
Thanks, closing in favor of #1309.

PR #1311, opened 2021-07-23
Fix unpredictable outcome of compression.DetectCompression*

mtrmac:
Because `compressionAlgorithms` is a map, it is traversed in an unpredictable order, so using the same prefix for `zstd` and `zstd:chunked` meant that the caller could get either.

Define empty prefix as "opt out of detection", to always return `Zstd`.


rhatdan:
LGTM
@vrothberg PTAL

vrothberg (approved):
LGTM

PR #1314, opened 2021-07-26
Stop caching encrypted layers in docker destination

hshiina:
In a case where a non-encrypted images is committed to a registry after
an encrypted image is committed, this library may try to reuse
encrypted layers for the non-encrypted images if the images have common
layers. When the non-encrypted image is committed, there is no
information to recognize that cached layers are encrypted. As a result,
the non-encrypted image has encrypted layers unexpectedly in the
registry. This causes an error when the non-encrypted image is pulled
from the registry.

To avoid reusing encrypted layers for another image, this change stops
caching a layer if it is encrypted.

Signed-off-by: Hironori Shiina <shiina.hironori@jp.fujitsu.com>

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -1620,6 +1620,7 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
 			finalizer = fin
 			inputInfo.Digest = ""
 			inputInfo.Size = -1
+			inputInfo.CryptoOperation = types.Encrypt
 			encrypted = true
 		}
 	}
--- a/docker/docker_image_dest.go
+++ b/docker/docker_image_dest.go
@@ -205,7 +205,9 @@ func (d *dockerImageDestination) PutBlob(ctx context.Context, stream io.Reader, 
 	}
 
 	logrus.Debugf("Upload of layer %s complete", computedDigest)
-	cache.RecordKnownLocation(d.ref.Transport(), bicTransportScope(d.ref), computedDigest, newBICLocationReference(d.ref))
+	if inputInfo.CryptoOperation != types.Encrypt {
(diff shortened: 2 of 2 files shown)


mtrmac (changes requested):
Thanks for the PR.

Targeting `RecordKnownLocation` like this seems unlikely to be correct
- As a trivial matter, `RecordKnownLocation` is called elsewhere for other reasons; e.g. just copying an encrypted image from one registry to another can create a record, so this fix doesn’t work.
- Layer reuse can happen without a `RecordKnownLocation` at all.
- It’s awkward for the transport to deal with this; this should be handled in the transport-independent copy logic, somehow.
- Conceptually, the _location record_ is not the issue at all; we want to avoid the reuse of encrypted layers when the original is not encrypted (or differently-encrypted); i.e. the issue is the digest/uncompressed-digest associations don’t record the information.

mtrmac:
> we want to avoid the reuse of encrypted layers when the original is not encrypted (or differently-encrypted); i.e. the issue is the digest/uncompressed-digest associations don’t record the information.

E.g. consider what happens when an already-encrypted image is copied to a registry (which does not trigger this code path at all); and then is copied _decrypting_ it, into a `docker-archive:` destination; that might create a digest/uncompressed record as well, and a known location. Copying the same image back into the same repo could again trigger reuse.

I haven’t really thought about how to actually cleanly fix this yet, though, I’m afraid.

mtrmac:
> I haven’t really thought about how to actually cleanly fix this yet, though, I’m afraid.

https://github.com/containers/image/pull/1533 is probably not the most accurate logic possible, but it at least seems to conservatively avoid the problematic layer reuse (after one-time manual removal of the blob info cache).

mtrmac:
#1533 was merged, and is effectively a superset of this PR. Thanks!

PR #1315, opened 2021-07-27
Return an actual error instead of nil when registry returns bad status code

divolgin:
`SearchRegistry` can return nil error when registry returns a status code that indicates an error, such as 401, because `err` is nil, which is ignored by `errors.Wrap` function.


rhatdan:
Thanks @divolgin 
LGTM
@mtrmac @vrothberg PTAL

mtrmac (approved):
LGTM. Thanks!

PR #1318, opened 2021-07-27
Simplify credential normalization, fix tests

mtrmac:
- Don't build a map just to access one element. NOTE: This might change behavior with ambiguous files, we now use the first matching entry, not the last one.
- Don't unnecessarily strip schema:// and /path from registry on search, and adjust tests


vrothberg (commented):
LGTM
@rhatdan @saschagrunert PTAL

saschagrunert (approved):
LGTM

PR #1320, opened 2021-07-29
copy: drop FetchPartialBlobs option

giuseppe:
drop FetchPartialBlobs option and always try to pull the partial
image.

The partial pull is performed only when the image has an annotation,
so there is no need to request this information to the user.


mtrmac (commented):
LGTM. Thanks.

(See #1321 for a bit more on why we are doing this API break.)

rhatdan:
Should we cut  a new release for this, so we limit the length of the "breaking change"?

mtrmac:
This is already included in v5.15.0.

PR #1321, opened 2021-07-29
Remove copy.Options.FetchPartialBlobs

mtrmac:
... and enable the option by default.  Now that it is enabled by default, the option can't be used to opt out because we can't distinguish between "unset" and "opt out"; so we would need a different option to support opt outs anyway.

This is technically an API break, which we try really hard not to do, but it was in a published release only for 7 days,
has no known users, and it was documented as experimental, so hopefully we can get away with it.

@giuseppe RFC


mtrmac:
Closing in favor of #1320.

PR #1322, opened 2021-07-30
backport + v5.14.1

vrothberg:
Backport commit 3556c3159a5f237a57dc6c7c72355825e0e640c0 and cut a new v5.14.1.  This way we can easily cut a new Skopeo from main including the feature from 3556c3159a5f237a57dc6c7c72355825e0e640c0.

@mtrmac @TomSweeneyRedHat PTAL

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -101,26 +101,22 @@ test_task:
               BUILDTAGS: 'btrfs_noversion libdm_no_deferred_remove'
         - name: "Test w/ opengpg"
           env:
-              BUILDTAGS: 'btrfs_noversion libdm_no_deferred_remove containers_image_openpgp'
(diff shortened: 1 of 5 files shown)


mtrmac:
@vrothberg Why not tag the release on the main branch, and branch from the release tag? I’d like to include things like #1305 #1307 #1320 . OTOH note that the main branch includes several vendor updates.

---

Actually https://github.com/containers/image/commit/3556c3159a5f237a57dc6c7c72355825e0e640c0 adds a new API, so I think this should be a v5.15.0 anyway — and that really should be from the main branch. I might be missing some reason to branch specifically from that point, and to backport specifically that feature.

vrothberg:
Sure, let's cut from main.

PR #1328, opened 2021-08-04
Update docker/docker to v20.10.7+incompatible

mtrmac:
It seems Dependabot can't handle the current version requirement as expected; the latest now is v20.10.8.

Intentionally update to v20.10.7 to verify that Dependabot will now find ...8.


mtrmac:
Success: #1330 .

PR #1333, opened 2021-08-05
Don't recompress non-gzip data without an explicit request

mtrmac:
With destination transports where `DesiredLayerCompression() == Compress`, notably c/image/docker, only recompress layers if the user has explicitly requested a specific format in `SystemContext`; don't default to recompressing `zstd` layers to `gzip`.

See individual commit messages for more details.

Should address https://github.com/containers/skopeo/issues/1406 .


mtrmac:
#1334 might help with the test failures.

TomSweeneyRedHat:
LGTM

vrothberg (commented):
LGTM

PR #1340, opened 2021-08-10
accept identity tokens from credential helpers

terinjokes:
When calling a credential helper, there are two types of credentials
that may be returned: a username and password pair, or an identity
token. The credential helper uses the same struct for returning both
types of credentials. From the `docker login` [documentation]:

> If the secret being stored is an identity token, the Username should
> be set to `<token>`.

When an identity token is provided, the Docker client should
authenticate using OAuth2 rather than the standard bearer challenge
response.

This changeset modifies the credential helper code to recognize this
special username, and sets the IdentityToken field of DockerAuthConfig
instead of Username and Password.

[documentation]: https://docs.docker.com/engine/reference/commandline/login/#credential-helper-protocol

Signed-off-by: Terin Stock <terinjokes@gmail.com>

### DIFF
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -601,10 +601,18 @@ func getAuthFromCredHelper(credHelper, registry string) (types.DockerAuthConfig,
 	if err != nil {
 		return types.DockerAuthConfig{}, err
 	}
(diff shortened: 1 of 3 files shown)


mtrmac (approved):
LGTM. Thanks!

Issue #1160, opened 2021-02-23
Image pull computes 2–3× more digests than necessary


mtrmac:
On the most common pull path, we compute:
- `c/image/copy`:
  - `digestingReader` = digest of the compressed layer (for authenticity)
- `c/image/storage/storageImageDestination.PutBlob`:
  - digest of the compressed layer (for deduplication of metadata during pull + for API conformance of the “write” e.g. when decrypting)
  - digest of the uncompressed layer (for the “lookup side” layer deduplication / reuse across images)
- `c/storage/layerStore.ApplyDiff`:
  - digest of the compressed layer (to record data that allows avoiding layer pulls)
  - digest of the uncompressed layer (to record data that allows avoiding layer pulls)

i.e. each digest is computed at least twice; that should be avoidable by adding some guarantees to the API, and a new c/storage interface that allows the caller to submit pre-computed digest values.

mrunalp:
@vrothberg @rhatdan @nalind @umohnani8 ptal

vrothberg:
I'll be tackling push & pull optimizations this sprint.  Please let me know in case this issue has a higher priority.

rhatdan:
@vrothberg @QiWang19 Any progress on this?  People are noticing...

vrothberg:
I didn't work on it, no.

vrothberg:
@mtrmac, to avoid duplicate work: are you tackling the c/storage parts as well?

mtrmac:
The Jira card is assigned to me, so I’ll work on that.

mtrmac:
Scope extension: Many transports’ `PutBlob` methods unconditionally recompute digests; they shouldn’t. (And we should make the requirement that `PutBlob` callers are required to only submit validated digests official.)

mtrmac:
https://github.com/containers/storage/pull/1003 + https://github.com/containers/image/pull/1352 is the second part.

Scope extension work pending.

mtrmac:
> Scope extension work pending.

The final part will eventually be https://github.com/containers/image/pull/1355 .

rhatdan:
Thanks @mtrmac

PR #1350, opened 2021-08-19
Don’t compute a digest of the compressed blob twice when pulling

mtrmac:
This is the smaller half of #1160.

Please see individual commit messages for details.

Cc:
- @nalind : Is there some subtlety about using `digest.Digest.Validate()`? Are there callers of `PutBlob()` that can’t be relied on to provide correct digests?
- @mrunalp

### DIFF
--- a/storage/storage_image.go
+++ b/storage/storage_image.go
@@ -532,9 +532,9 @@ func (s *storageImageDestination) PutBlob(ctx context.Context, stream io.Reader,
 
 	// Record information about the blob.
 	s.lock.Lock()
-	s.blobDiffIDs[hasher.Digest()] = diffID.Digest()
-	s.fileSizes[hasher.Digest()] = counter.Count
-	s.filenames[hasher.Digest()] = filename
+	s.blobDiffIDs[blobDigest] = diffID.Digest()
+	s.fileSizes[blobDigest] = counter.Count
+	s.filenames[blobDigest] = filename
 	s.lock.Unlock()
 	// This is safe because we have just computed both values ourselves.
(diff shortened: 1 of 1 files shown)


storage/storage_image.go
```diff
@@ -482,55 +484,66 @@ func (s *storageImageDestination) PutBlob(ctx context.Context, stream io.Reader,
...
 		Size:   -1,
 	}
-	// Set up to digest the blob and count its size while saving it to a file.
-	hasher := digest.Canonical.Digester()
-	if blobinfo.Digest.Validate() == nil {
-		if a := blobinfo.Digest.Algorithm(); a.Available() {
-			hasher = a.Digester()
+	if blobinfo.Digest != "" {
+		if err := blobinfo.Digest.Validate(); err != nil {
+			return errorBlobInfo, fmt.Errorf("invalid digest %#v: %w", blobinfo.Digest.String(), err)
```
mtrmac, line 489: FWIW this introduces the first use of Go 1.13-style standard-library errors with `Unwrap()` instead of `github.com/pkg/errors.Wrapf`. We require Go 1.13 anyway, and pkg/errors includes call stack information which we never used, so moving closer to the standard library will hopefully be more efficient as well as, eventually, more familiar to newcomers.

I don’t think we want to do any kind of large-scale migration, just maybe start using this in new code.

@vrothberg RFC.

rhatdan:
LGTM
@vrothberg PTAL

rhatdan:
@nalind PTAL

vrothberg (approved):
LGTM

+1 for breaking it into such easy to digest commits ... almost surgical precision

PR #1351, opened 2021-08-20
compression: add support for estargz

giuseppe:
add support for the estargz compression format from
https://github.com/containerd/stargz-snapshotter

### DIFF
--- a/pkg/compression/compression.go
+++ b/pkg/compression/compression.go
@@ -35,6 +35,9 @@ var (
 	// Zstd:chunked compression.
 	ZstdChunked = internal.NewAlgorithm(types.ZstdChunkedAlgorithmName, types.ZstdAlgorithmName, /* Note: InternalUnstableUndocumentedMIMEQuestionMark is not ZstdChunkedAlgorithmName */
 		nil, ZstdDecompressor, compressor.ZstdCompressor)
+	// Gzip:estargz compression.
+	GzipEstargz = internal.NewAlgorithm(types.GzipEstargzAlgorithmName, types.GzipAlgorithmName, /* Note: InternalUnstableUndocumentedMIMEQuestionMark is not EstargzAlgorithmName */
+		nil, GzipDecompressor, compressor.EstargzCompressor)
 
 	compressionAlgorithms = map[string]Algorithm{
 		Gzip.Name():        Gzip,
@@ -42,6 +45,7 @@ var (
 		Xz.Name():          Xz,
 		Zstd.Name():        Zstd,
 		ZstdChunked.Name(): ZstdChunked,
+		GzipEstargz.Name(): GzipEstargz,
 	}
 )
 
--- a/pkg/compression/types/types.go
+++ b/pkg/compression/types/types.go
@@ -38,4 +38,9 @@ const (
 	// will actually be available. (In fact it is intended for this types package not to depend
 	// on any of the implementations.)
 	ZstdChunkedAlgorithmName = "zstd:chunked"
+	// GzipEstargzAlgorithmName is the name used by pkg/compression.Estargz.
+	// NOTE: Importing only this /types package does not inherently guarantee a GzipEstargz algorithm
+	// will actually be available. (In fact it is intended for this types package not to depend
+	// on any of the implementations.)
+	GzipEstargzAlgorithmName = "gzip:estargz"
 )


pkg/compression/compression.go
```diff
@@ -35,13 +37,17 @@ var (
 	// Zstd:chunked compression.
 	ZstdChunked = internal.NewAlgorithm(types.ZstdChunkedAlgorithmName, types.ZstdAlgorithmName, /* Note: InternalUnstableUndocumentedMIMEQuestionMark is not ZstdChunkedAlgorithmName */
 		nil, ZstdDecompressor, compressor.ZstdCompressor)
+	// Estargz compression.
+	Estargz = internal.NewAlgorithm(types.EstargzChunkedAlgorithmName, types.GzipAlgorithmName, /* Note: InternalUnstableUndocumentedMIMEQuestionMark is not ZstdChunkedAlgorithmName */
```
mtrmac, line 41: The comment is wrong.

(Also, please actually define and document the semantics of that field?)

pkg/compression/compression.go
```diff
@@ -165,3 +171,57 @@ func AutoDecompress(stream io.Reader) (io.ReadCloser, bool, error) {
 	}
 	return res, decompressor != nil, nil
 }
+
+type estarCompressor struct {
```
mtrmac, line 175: Move this into a separate file, please; let’s keep `compression.go` somewhat generic.

pkg/compression/compression.go
```diff
@@ -165,3 +171,57 @@ func AutoDecompress(stream io.Reader) (io.ReadCloser, bool, error) {
...
+	}
+
+	blob, err := estargz.Build(sr, estargz.WithCompressionLevel(compressionLevel))
+	if err != nil {
+		return err
+	}
+	defer blob.Close()
+
+	c.metadata["containerd.io/snapshot/stargz/toc.digest"] = string(blob.TOCDigest())
+	c.metadata["io.containers.estargz.uncompressed-size"] = fmt.Sprintf("%v", st.Size())
```
mtrmac, line 206: Is there a Go package providing these strings as constants?

pkg/compression/compression.go
```diff
@@ -165,3 +171,57 @@ func AutoDecompress(stream io.Reader) (io.ReadCloser, bool, error) {
...
+
+	_, err = io.Copy(c.writer, blob)
+	return err
+}
+
+// estargzCompressor is a CompressorFunc for the estargz compression algorithm.
+func estargzCompressor(r io.Writer, metadata map[string]string, level *int) (io.WriteCloser, error) {
+	file, err := ioutil.TempFile("", "estargz-layer")
```
mtrmac, line 214: `TemporaryDirectoryForBigFiles`

pkg/compression/types/types.go
```diff
@@ -38,4 +38,6 @@ const (
 	// will actually be available. (In fact it is intended for this types package not to depend
 	// on any of the implementations.)
 	ZstdChunkedAlgorithmName = "zstd:chunked"
+	// EstargzChunkedAlgorithmName is the name used by containerd/stargz-snapshotter.
```
mtrmac, line 41: Add the `NOTE` as well.

mtrmac (commented):
How large is the added dependency?

Through some lack of foresight and carelessness, we have ended up with various components that really shouldn’t care, like c/image/docker, depending on this subpackage.

At some point, we need to resolve that, either by a significant reimplementation of those dependency chains, or by making the new compression formats optional.

In this case at least, where we don’t need any new implementation to decompress the data, and where compression using this format can only ever happen by an explicit user request (i.e. `types.SystemContext.CompressionFormat` providing a `*compressiontypes.Algorithm`), it might make sense to move this into a subpackage. And then, to allow simple implementations of CLIs, we might need a `transports/alltransports`-like split of the algorithm registry and the algorithm dependencies.

pkg/compression/compression.go
ktock: Yes, there is at https://github.com/containerd/stargz-snapshotter/blob/v0.7.0/estargz/types.go#L70-L80 .

pkg/compression/compression.go
mtrmac, line 175: BTW at least a smoke test that this produces usable results would be nice.

giuseppe:
@ktock do you think we can move some of the logic to github.com/containerd/stargz-snapshotter?

ktock:
@giuseppe I think we can move `estarCompressor` to github.com/containerd/stargz-snapshotter/estargz . Could you open up the PR?

pkg/compression/compression.go
ktock, line 206: This needs to be the uncompressed size of the *result* estargz blob (not the source tar).

pkg/compression/compression.go
```diff
@@ -165,3 +171,57 @@ func AutoDecompress(stream io.Reader) (io.ReadCloser, bool, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	// Unlink immediately the file so we won't leak it.
+	os.Remove(file.Name())
```
AkihiroSuda, line 219: Can we return err here?

giuseppe:
I've moved the implementation to `containers/storage`: https://github.com/containers/storage/pull/1001

docker/docker_image_src.go
```diff
@@ -305,53 +305,92 @@ func (s *dockerImageSource) GetBlobAt(ctx context.Context, info types.BlobInfo,
...
-
-	streams := make(chan io.ReadCloser)
-	errs := make(chan error)
+	switch res.StatusCode {
+	case http.StatusOK:
+		// if the server replied with a 200 status code, convert the full body response to a series of
+		// streams as it would have been done with 206.
+		streams := make(chan io.ReadCloser)
+		errs := make(chan error)
+		go func() {
```
mtrmac, line 315: This is becoming rather large (e.g. how far down the “unknown `res.StatusCode` case is); please split the two `streams`-producing goroutines into two separate functions.

Unit tests also look worthwhile, at least for those `streams` producers; without testing actual pulls of this format with high frequency (the way Skopeo and Podman tests trigger at least the default pull paths), we are pretty unlikely to notice breakage.

giuseppe:
@mtrmac moved a patch to a different PR: https://github.com/containers/image/pull/1359

PR #1353, opened 2021-08-20
Don't call digest.Digester.Hash() on every digeringReader.Read()

mtrmac:
`digester.Hash()` just copies an interface value, but that apparently involves a heap allocation, and actually shows up fairly high in profiles.  So do that once per layer instead of once per copied buffer.

---

It turns out it’s a dangerous idea to start looking at profile data, one never knows what one could find.


rhatdan:
LGTM
@vrothberg PTAL

vrothberg (commented):
LGTM, great catch!

PR #1354, opened 2021-08-21
Run (gofmt -s -w)

mtrmac:
Go 1.17 introduces a much more reasonable build constraint format, and `gofmt` now fails without using it.

Sadly we still need the old format _as well_, to support <1.17 builds.


rhatdan:
LGTM

PR #1355, opened 2021-08-21
Only compute digests in PutBlob if truly necessary

### LINKED ISSUE #1160: Image pull computes 2–3× more digests than necessary

mtrmac:
Document that `PutBlob` callers must only provide validated digests. This was always sort of implied, and we now rely on that explicitly to avoid computing the digests; it turns out the digest computation can consume a very noticeable amount of CPU time.

Then Introduce `internal/putblobdigest.Digester` and use it to let `PutBlob` use caller-provided digest values, if any.

Fixes #1160 (the last part).

### DIFF
--- a/directory/directory_dest.go
+++ b/directory/directory_dest.go
@@ -171,9 +171,9 @@ func (d *dirImageDestination) PutBlob(ctx context.Context, stream io.Reader, inp
 	if err != nil {
 		return types.BlobInfo{}, err
 	}
-	computedDigest := digester.Digest()
+	blobDigest := digester.Digest()
 	if inputInfo.Size != -1 && size != inputInfo.Size {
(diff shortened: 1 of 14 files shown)


giuseppe (approved):
LGTM

vrothberg (commented):
LGTM

PR #1356, opened 2021-08-23
copy.Options: semaphore to limit parallel pulls globaly

vrothberg:
Add a weighted semaphore to the copy options which allows for setting a
limit of simultaneously pulled layers within the current process space.

This allows for long-running processes, such as CRI-O, to set custom
limits based on the execution environment.

@mrunalp @mtrmac PTAL ... as so often, I feel like naming the option is the hardest part.

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -63,20 +63,19 @@ var expectedCompressionFormats = map[string]*compressiontypes.Algorithm{
 // copier allows us to keep track of diffID values for blobs, and other
 // data shared across one or more images in a possible manifest list.
 type copier struct {
-	dest                  types.ImageDestination
-	rawSource             types.ImageSource
-	reportWriter          io.Writer
-	progressOutput        io.Writer
-	progressInterval      time.Duration
-	progress              chan types.ProgressProperties
-	blobInfoCache         internalblobinfocache.BlobInfoCache2
-	copyInParallel        bool
-	compressionFormat     *compressiontypes.Algorithm // Compression algorithm to use, if the user explicitly requested one, or nil.
-	compressionLevel      *int
-	ociDecryptConfig      *encconfig.DecryptConfig
-	ociEncryptConfig      *encconfig.EncryptConfig
-	maxParallelDownloads  uint
-	downloadForeignLayers bool
+	dest                          types.ImageDestination
+	rawSource                     types.ImageSource
+	reportWriter                  io.Writer
+	progressOutput                io.Writer
+	progressInterval              time.Duration
+	progress                      chan types.ProgressProperties
+	blobInfoCache                 internalblobinfocache.BlobInfoCache2
+	compressionFormat             *compressiontypes.Algorithm // Compression algorithm to use, if the user explicitly requested one, or nil.
+	compressionLevel              *int
+	ociDecryptConfig              *encconfig.DecryptConfig
+	ociEncryptConfig              *encconfig.EncryptConfig
+	concurrentBlobCopiesSemaphore *semaphore.Weighted // Limits the amount of concurrently copied blobs
+	downloadForeignLayers         bool
 }
 
 // imageCopier tracks state specific to a single image (possibly an item of a manifest list)
@@ -149,7 +148,10 @@ type Options struct {
 	// encrypted if non-nil. If nil, it does not attempt to decrypt an image.
 	OciDecryptConfig *encconfig.DecryptConfig
 
-	// MaxParallelDownloads indicates the maximum layers to pull at the same time.  A reasonable default is used if this is left as 0.
+	// A weighted semaphore to limit the amount of concurrently copied layers and configs. Applies to all copy operations using the semaphore. If set, MaxParallelDownloads is ignored.
+	ConcurrentBlobCopiesSemaphore *semaphore.Weighted
+
+	// MaxParallelDownloads indicates the maximum layers to pull at the same time. Applies to a single copy operation. A reasonable default is used if this is left as 0. Ignored if ConcurrentBlobCopiesSemaphore is set.
 	MaxParallelDownloads uint
 
 	// When OptimizeDestinationImageAlreadyExists is set, optimize the copy assuming that the destination image already
@@ -222,7 +224,6 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
 	if !isTTY(reportWriter) {
 		progressOutput = ioutil.Discard
 	}
-	copyInParallel := dest.HasThreadSafePutBlob() && rawSource.HasThreadSafeGetBlob()
 
 	c := &copier{
 		dest:             dest,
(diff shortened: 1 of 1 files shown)


copy/copy.go
```diff
@@ -231,16 +232,29 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
...
+	if dest.HasThreadSafePutBlob() && rawSource.HasThreadSafeGetBlob() {
+		if c.layerCopySemaphore == nil {
+			max := options.MaxParallelDownloads
+			if max == 0 {
+				max = maxParallelDownloads
+			}
+			c.layerCopySemaphore = semaphore.NewWeighted(int64(max))
+		}
+	} else {
+		c.layerCopySemaphore = semaphore.NewWeighted(int64(1))
```
mtrmac, line 255: With this implementation, with a global semaphore for 5 copies, initiating 20 copies will cause only 5 layer copies if in parallel, but 20 if not parallel. I think this should acquire the global semaphore here, and create a local 1-weight semaphore for local synchronization.

OTOH I can see the argument for “you started 20 copies, you get 20 copies”…

vrothberg: Fair point. Guess we should do what you suggest :+1:

vrothberg:
@mtrmac, I guess we should also acquire the semaphore when copying the config, right?

vrothberg:
> @mtrmac, I guess we should also acquire the semaphore when copying the config, right?

done

mtrmac:
> @mtrmac, I guess we should also acquire the semaphore when copying the config, right?

I concur.

It is copied with `canModifyBlob = false`, and uncompressed, so it is not going to trigger the 4 MB allocation that initiated this work. OTOH “limit parallel object copies / HTTP streams / storage writes” is a much clearer concept for callers than “limit actions that _might_ trigger a large memory allocation depending on an implementation (and we are not even checking the blog format to see whether that blob triggers the allocation, but count that as a parallel layer anyway)”.

Either way, the caller will have to make some kind of heuristic decision about how many parallel streams to allow, and this allows limiting the worst case but makes it not at all easy for the caller to achieve _exactly optimal_ performance given a specific resource constraint (no-one knows how much bandwidth or memory or CPU a blob is going to require, and that data changes with every dependency update).

So doing the simple and consistent thing seems better.

copy/copy.go
```diff
@@ -149,9 +148,12 @@ type Options struct {
 	// encrypted if non-nil. If nil, it does not attempt to decrypt an image.
 	OciDecryptConfig *encconfig.DecryptConfig
 
-	// MaxParallelDownloads indicates the maximum layers to pull at the same time.  A reasonable default is used if this is left as 0.
+	// MaxParallelDownloads indicates the maximum layers to pull at the same time. Applies to a single copy operation. A reasonable default is used if this is left as 0.
 	MaxParallelDownloads uint
```
mtrmac, line 153: (Nit: Maybe drop this blank line to make the “`MaxParallelDownloads` is ignored” comment a bit more visible. And/or reverse the order so that newcomers read about `GlobalParallelDownloadsSemaphore` first.)

copy/copy.go
```diff
@@ -231,16 +232,35 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
...
+			if max == 0 {
+				max = maxParallelDownloads
+			}
+			c.copySemaphore = semaphore.NewWeighted(int64(max))
+		}
+	} else {
+		c.copySemaphore = semaphore.NewWeighted(int64(1))
+		if options.GlobalParallelDownloadsSemaphore != nil {
+			if err := options.GlobalParallelDownloadsSemaphore.Acquire(ctx, 1); err != nil {
+				return nil, errors.Wrap(err, "acquiring global parallel downloads semaphore")
```
mtrmac, line 258: Non-blocking:
```suggestion
				return nil, fmt.Errorf("acquiring global parallel downloads semaphore: %w", err)
```

copy/copy.go
```diff
@@ -149,9 +148,12 @@ type Options struct {
 	// encrypted if non-nil. If nil, it does not attempt to decrypt an image.
 	OciDecryptConfig *encconfig.DecryptConfig
 
-	// MaxParallelDownloads indicates the maximum layers to pull at the same time.  A reasonable default is used if this is left as 0.
+	// MaxParallelDownloads indicates the maximum layers to pull at the same time. Applies to a single copy operation. A reasonable default is used if this is left as 0.
 	MaxParallelDownloads uint
 
+	// A weighted semaphore to limit the global maximum of simultaneously pulled layers and configs. Applies to all copy operations using the semaphore. If set, MaxParallelDownloads is ignored.
+	GlobalParallelDownloadsSemaphore *semaphore.Weighted
```
mtrmac, line 155: Non-blocking, just something to consider. This works well enough, entirely up to you.
- It’s not really `Downloads`; it applies to pushes equally. OTOH accurate `BlobCopies` is not all that great, and of course there’s a precedent.
- I’m not sure `Global` is necessary; pedantically it’s “caller-provided” and the caller is free to set up a few concurrent scopes with their own semaphores, so it’s not _always_ global (and “caller-provided” is of course implied).

copy/copy.go
```diff
@@ -70,12 +70,11 @@ type copier struct {
 	progressInterval      time.Duration
 	progress              chan types.ProgressProperties
 	blobInfoCache         internalblobinfocache.BlobInfoCache2
-	copyInParallel        bool
 	compressionFormat     *compressiontypes.Algorithm // Compression algorithm to use, if the user explicitly requested one, or nil.
 	compressionLevel      *int
 	ociDecryptConfig      *encconfig.DecryptConfig
 	ociEncryptConfig      *encconfig.EncryptConfig
-	maxParallelDownloads  uint
+	copySemaphore         *semaphore.Weighted
```
mtrmac, line 77: Non-blocking?

The `copySemaphore` name in `copier.copySemaphore *semaphore` adds no information over the parent and type, and the scope is now larger so there is more ambiguity.

`blobCopySemaphore`? `parallelBlobCopySemaphore`? Or just move something like the original comment here?

mtrmac (commented):
Thanks!

Implementation LGTM, just some weakly-held stylistic observations.

copy/copy.go
vrothberg: Renamed to `ConcurrentBlobCopiesSemaphore` ... naming is hard

copy/copy.go
vrothberg: Renamed to `concurrentBlobCopiesSemaphore`

copy/copy.go
```diff
@@ -912,9 +919,9 @@ func (ic *imageCopier) copyLayers(ctx context.Context) error {
 		defer copyGroup.Wait()
 
 		for i, srcLayer := range srcInfos {
-			err = copySemaphore.Acquire(ctx, 1)
+			err = ic.c.copySemaphore.Acquire(ctx, 1)
 			if err != nil {
-				return errors.Wrapf(err, "Can't acquire semaphore")
+				return errors.Wrap(err, "copying layer")
```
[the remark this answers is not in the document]
vrothberg: I should have moved it into a separate commit.  I found it useful to add context to the errors so that we can see where things broke without a stack trace.

vrothberg:
Thanks for reviewing, @mtrmac, PTanotherL

copy/copy.go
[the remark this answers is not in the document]
mtrmac: Oh, that’s my mistake — this change differentiates “failed acquiring the semaphore _while copying a layer vs. config_”, which is definitely useful and I completely failed to think about. The part I was focusing on was that this change loses the context that the failure is _acquiring semaphore vs. other failures during copy_, which is not actually necessary in the end I think (but it’s not trivial).

I wasn’t asking for a comment justifying the use of error wrapping just in this one place; I was asking for a comment justifying not including the “acquiring semaphore” context, maybe something like `// This can only fail with ctx.Err(), so no need to blame acquiring the semaphore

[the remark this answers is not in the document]
vrothberg: Done :heavy_check_mark:

copy/copy.go
mtrmac: It totally is.

mtrmac (approved):
LGTM. Thanks again!

PR #1364, opened 2021-08-27
Add image destination transport support for sigstore's cosign image signing

font:
This is a very rough WIP PR to add image destination transport support for signing container images using the [cosign signature spec](https://github.com/sigstore/cosign/blob/main/specs/SIGNATURE_SPEC.md). There are some TODOs throughout and summarized here:

### TODO:
- [ ] xdg-open does not open browser and URL is not printed to terminal. This prevents successful keyless OIDC auth flow in rootless mode. Additionally, callback server at `http://localhost:5556/auth/callback` is not successfully launched. Need to continue investigating this issue.
- [ ] Push the signature image and manifest (`putSignature`) to the registry.
- [ ] Eventually (not necessarily in this PR) remove use of env vars for Rekor and Fulcio server URLs and replace with policy configuration.
- [ ] Add signing to `copyMultipleImages`.
- [ ] Use debug log in sigstore mechanism.
- [ ] `registries.d/default.yaml` field name and location for storing Sigstore's cosign signatures.
- [ ] Consider adding optional claims to the `Optional` simple signing field.
- [ ] Consistent usage of sigstore/cosign naming convention throughout. The c/* tooling already has a defined meaning for a *sigstore* concept. Converge on proper naming to differentiate both.
- [ ] Consolidate interfaces and APIs with existing GPG signing mechanism.
- [ ] Any useful tests.
- [ ] Add missing godoc comments.
- [ ] Anything else?


signature/signature.go
```diff
@@ -199,6 +200,95 @@ func (s untrustedSignature) sign(mech SigningMechanism, keyIdentity string) ([]b
 	return mech.Sign(json, keyIdentity)
 }
 
+// cosignSignature is the payload format for sigstore signing an image using
+// the simple signing format.
+type cosignSignature struct {
```
mtrmac, line 205: I’m paranoid about misuse — I’d prefer to prefix all the names in this with `untrusted`, compare `untrustedSignature` vs. `Signature`.

---

BTW shouldn’t the internal types like `untrustedSignature` be now renamed to refer to a singing format? We can’t remove the public ones, but at least the internal ones should use clear names. For the public part, we might be able to provide new aliases and deprecate old names.

signature/signature.go
```diff
@@ -199,6 +200,95 @@ func (s untrustedSignature) sign(mech SigningMechanism, keyIdentity string) ([]b
...
+func (s cosignSignature) MarshalJSON() ([]byte, error) {
+	if s.DockerManifestDigest == "" || s.DockerReference == "" {
+		return nil, errors.New("Unexpected empty signature content")
+	}
+	critical := map[string]interface{}{
+		"type":     cosignSignatureType,
+		"image":    map[string]string{"docker-manifest-digest": s.DockerManifestDigest.String()},
+		"identity": map[string]string{"docker-reference": s.DockerReference},
+	}
+	optional := map[string]interface{}{} // TODO: add annotations i.e. claims
```
mtrmac, line 234: I do think there’s non-trivial long-term value with including the name and version number of the producer inside the signature, so that future versions can add workarounds for known bugs in old software.

signature/signature.go
```diff
@@ -15,7 +15,8 @@ import (
 )
 
 const (
-	signatureType = "atomic container signature"
+	signatureType       = "atomic container signature"
+	cosignSignatureType = "cosign container image signature"
```
mtrmac, line 19: Very tentative: Maybe implement the two formats in separate files with a similar structure?

I’m not at all sure how much reuse (or a shared interface?) makes sense.

copy/sign.go
```diff
@@ -29,3 +40,53 @@ func (c *copier) createSignature(manifest []byte, keyIdentity string) ([]byte, e
...
+	}
+
+	fmt.Println("Generating certificate from Fulcio")
+	if err := mech.GenerateCertificate(); err != nil {
+		return nil, errors.Wrap(err, "getting key from Fulcio")
+	}
+
+	fmt.Println("Marshalling payload into JSON")
+	cosignSignature := signature.NewCosignSignature(manifestDigest, dockerReference.String())
+	sigPayload, err := cosignSignature.MarshalJSON()
```
mtrmac, line 72: This feels like something that should be a bit higher-level; at this high level of abstraction, the caller shouldn’t need to do 4 calls exactly in sequence, and definitely not care that the format is JSON. Maybe, very vaguely

- SignDockerManifestWithCosign(manifest, imageIdentity, whoSignsThis) -> blob(s)ToStoreWithTheImage, dataForUpload
- UploadTransparencyLogEntry(dataForUpload, credentialsIfAny)

(Should the mechanism even have _any_ public methods apart from `Close`? The `Sign…WithCosign` might be the only supported caller for now. We can always make private methods public, but not the other way.)

signature/mechanism.go
```diff
@@ -34,6 +35,21 @@ type SigningMechanism interface {
...
+type SigstoreSigningMechanism interface {
+	// Close removes resources associated with the mechanism, if any.
+	Close() error
+	// SupportsSigning returns nil if the mechanism supports signing, or a SigningNotSupportedError.
+	SupportsSigning() error
+	GenerateCertificate() error
+	// Sign creates a (non-detached) signature of input using ephemeral keys.
+	// Fails with a SigningNotSupportedError if the mechanism does not support signing.
+	Sign(payload []byte) (signature []byte, err error)
+	Upload(signature, payload []byte) (*models.LogEntryAnon, error)
```
mtrmac, line 50: Should this be public at all? See elsewhere.

signature/mechanism_sigstore.go
```diff
@@ -0,0 +1,325 @@
...
+	}
+
+	fClient := fulcioClient.New(fulcioServer)
+	if fClient == nil {
+		return errors.New("error initializing fulcio client")
+	}
+
+	// TODO: use debug log
+	fmt.Println("Generating ephemeral keys")
+	priv, err := generatePrivateKey()
```
mtrmac, line 106: Besides being a good practice, AFAIK FIPS requires explicit zeroization of all secret material after it is no longer used. How is that going to work?

`crypto/ecdsa` doesn’t seem to attempt to do this at all.

(Yeah, it’s hard. That’s why the previous implementation only allowed signing via a separate third-party non-Go process.)

signature/signature.go
```diff
@@ -199,6 +200,95 @@ func (s untrustedSignature) sign(mech SigningMechanism, keyIdentity string) ([]b
...
+// cosignSignature is the payload format for sigstore signing an image using
+// the simple signing format.
+type cosignSignature struct {
+	DockerManifestDigest digest.Digest
+	DockerReference      string
+}
+
+// newCosignSignature returns a cosignSignature object with
+// the specified primary contents and appropriate metadata.
+func NewCosignSignature(dockerManifestDigest digest.Digest, dockerReference string) cosignSignature {
```
mtrmac, line 212: I’d prefer the public API to be higher-level, see elsewhere.

signature/mechanism_sigstore.go
```diff
@@ -0,0 +1,325 @@
...
+	return certPem, chainPem, nil
+}
+
+// Sign creates a signature using the payload input.
+// Fails with a SigningNotSupportedError if the mechanism does not support signing.
+func (s *sigstoreSigningMechanism) Sign(payload []byte) ([]byte, error) {
+	return s.SignMessage(bytes.NewReader(payload), options.WithContext(s.ctx))
+}
+
+func (s *sigstoreSigningMechanism) Upload(signature, payload []byte) (*models.LogEntryAnon, error) {
```
mtrmac, line 182: This doesn’t use a `context.Context` for timeouts/cancellation?

signature/mechanism_sigstore.go
```diff
@@ -0,0 +1,325 @@
...
+	return ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+}
+
+func getCertForOauthID(priv *ecdsa.PrivateKey, ops operations.ClientService, oidcIssuer string, oidcClientID string) (certPem, chainPem []byte, err error) {
+	pubBytes, err := x509.MarshalPKIXPublicKey(&priv.PublicKey)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	tok, err := oauthflow.OIDConnect(oidcIssuer, oidcClientID, "", oauthflow.DefaultIDTokenGetter)
```
mtrmac, line 138: This doesn’t use a `context.Context` for timeouts/cancellation?

signature/mechanism_sigstore.go
```diff
@@ -0,0 +1,325 @@
...
+		&fulcioModels.CertificateRequest{
+			PublicKey: &fulcioModels.CertificateRequestPublicKey{
+				Algorithm: fulcioModels.CertificateRequestPublicKeyAlgorithmEcdsa,
+				Content:   &content,
+			},
+			SignedEmailAddress: &signedChallenge,
+		},
+	)
+
+	resp, err := ops.SigningCert(params, bearerAuth)
```
mtrmac, line 165: This doesn’t use a `context.Context` for timeouts/cancellation?

mtrmac (commented):
Just a fairly surface-level skim for now.

I’m strongly focusing on the public API because Go module semantics requires us to not break it, so it’s fairly important to get it right the first time.

As far as the internal impact on c/image’s `copy.Image`/`PutSignature`/`policy.json` and so on, there isn’t much here yet, it seems.

signature/mechanism_sigstore.go
```diff
@@ -0,0 +1,325 @@
...
+	LogIndex       int64       `json:"logIndex"`
+	LogID          string      `json:"logID"`
+}
+
+func VerifySET(bundlePayload BundlePayload, signature []byte, pub *ecdsa.PublicKey) error {
+	contents, err := json.Marshal(bundlePayload)
+	if err != nil {
+		return errors.Wrap(err, "marshaling")
+	}
+	canonicalized, err := jsoncanonicalizer.Transform(contents)
```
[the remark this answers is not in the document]
vrothberg: I'd love avoid new dependencies if possible.  This one seems to be maintained by a single person which is something we should consider.  In case we need to canonicalize, maybe we can use another library; ideally one we already use.

signature/mechanism_sigstore.go
```diff
@@ -0,0 +1,325 @@
...
+}
+
+func (s *sigstoreSigningMechanism) Upload(signature, payload []byte) (*models.LogEntryAnon, error) {
+	rekorClient, err := rekorClient.GetRekorClient(rekorServer())
+	if err != nil {
+		return nil, err
+	}
+	re := rekorEntry(payload, signature, s.cert)
+	returnVal := models.Rekord{
+		APIVersion: swag.String(re.APIVersion()),
```
vrothberg, line 189: `swag.String` seems to only return a pointer to the input string.  Could we do this manually and get rid of adding `swag` as a new dependency.

Note that I am trying to avoid new dependencies unless urgently necessary to keep our footprint small; it'll add some extra work downstream packaging as well (e.g., Debian has to devendor all dependencies and put them into extra source packages).

vrothberg (commented):
Thanks!

By no means a full review (I ran out of time for today).

copy/copy.go
```diff
@@ -751,6 +752,14 @@ func (c *copier) copyOneImage(ctx context.Context, policyContext *signature.Poli
 		sigs = append(sigs, newSig)
 	}
 
+	if options.Sign {
```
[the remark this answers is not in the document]
mtrmac: I think it makes enough sense to publish an image and sign it using both systems, to allow both kinds of consumers to consume it. Eventually we’ll probably recommend one of them, but that’s not a reason to prohibit signing using both.

signature/mechanism.go
mtrmac: `internal` is an interesting point — it didn’t even exist when `c/image/signature` was first written.

OTOH we might want to expose _something_ lower-level than `copy.Image`; e.g. `podman image sign` exists (as much as I’d like for that not to be the case), and other callers might have their own production/publishing/… workflows that don’t want `copy.Image`, while we’d still prefer them to use a library for actual signing (and much more importantly, verification).

But, short-term, doing things in `internal` makes the API design bar much lower, and we can always delay adding a public API until after there is a known user/need.

font (commented):
@mtrmac @vrothberg Thank you for your early review! There is still lots of work to do and there are many areas that need to be scrubbed. But I think we need to discuss a few things before I can continue moving forward:
1. The naming convention we want to use i.e. sigstore vs cosign, to avoid confusion and conflicts with existing naming scheme. This will likely impact user visible help strings and manpages for correct usage.
2. UX and semantics for `--sign` vs `--sign-by`. Originally I thought maybe `--sign-by` being `nil` could indicate ephemeral key (aka "keyless') signing instead of having to introduce a new flag. But that could be confusing. In any case, we should settle on the right flag and options to consider for future-proofing this in case we want to add bring-your-own-keys, KMS, etc. type of workflows in the future.
3. The right level of abstraction for introducing this feature's APIs and how to separate it with existing simple signing approach.

signature/mechanism_sigstore.go
```diff
@@ -0,0 +1,325 @@
...
+			fmt.Println("Signature already exists. Displaying proof")
+			uriSplit := strings.Split(existsErr.Location.String(), "/")
+			uuid := uriSplit[len(uriSplit)-1]
+			return verifyTLogEntry(rekorClient, uuid)
+		}
+		return nil, err
+	}
+	// UUID is at the end of location
+	for _, p := range resp.Payload {
+		return &p, nil
```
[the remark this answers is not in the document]
font: Mostly, except `resp.Payload` is a `map[string]`.

signature/mechanism_sigstore.go
```diff
@@ -0,0 +1,325 @@
...
+		return errors.Wrap(err, "generating private key")
+	}
+
+	signer, err := signature.LoadECDSASignerVerifier(priv, crypto.SHA256)
+	if err != nil {
+		return err
+	}
+
+	fmt.Println("Retrieving signed certificate")
+	cert, chain, err := getCertForOauthID(priv, fClient.Operations, oidcIssuer, oidcClientID)
```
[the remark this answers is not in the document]
mtrmac: That must moves the question one level higher — where does the list of OIDC providers come from, then? From the certificate-signing Fulcio server, and _that_ one is user-controlled? I think that works.

---

I wasn’t really thinking about what a browser in the picture means, to be honest.

But also, will this allow automated use without a browser? I don’t think that blocks a first version, but it seems like something that should be possible _eventually_, so a public API (if any) should make sure to allow passing that kind of information. (Making the API not public could well be a very acceptable way to bypass this question.)

signature/mechanism_sigstore.go
mtrmac: Well, https://github.com/sigstore/rekor/blob/main/pkg/generated/models/rekord.go depends on `swag` anyway, so I don’t think this would buy us anything.

signature/mechanism_sigstore.go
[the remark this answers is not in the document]
mtrmac: Ah, thanks, that explains the need for a loop, though not quite why the data structure looks that way.

font:
Still isn't ready for review necessarily, but I've pushed another update to start addressing some of the feedback.

mtrmac:
@font I realize you are pressed for time; the part that would help us most most of all to see how c/image needs to change would be to see a verification implementation (what are the from-registry inputs, what are the local-client trust configurations and roots, and how do they combine into a yes/no decision).

mtrmac:
#1785 , and pre-requisite PRs, is now a doing everything from this PR that we want to do (an example exception: don’t care to cryptographically validate the Rekor-returned data against that Rekor’s public key).

Thanks for helping with this!

PR #1191, opened 2021-03-28
Enable subdomain matching in registries.conf

lsm5:
This commit allows the prefix field in registries.conf to be of the
form: `prefix = "*.example.com" for wildcard subdomain matching.

refMatchesPrefix now returns the length of the prefix if there's a match
and the prefix doesn't contain `*.`. If prefix contains `*.` and there's
a match, then refMatchesPrefix returns the length of the refString
without the image. This change removes the need for
any additional string comparison in `rewriteReference`.

Co-authored-by: Valentin Rothberg <rothberg@redhat.com>
Signed-off-by: Lokesh Mandvekar <lsm5@fedoraproject.org>

/cc @ashcrow

PR #1366, opened 2021-08-30
Fix subdomain matching FIXMEs in  commit 373440662

lsm5:
FIXMEs addressed:

 1. https://github.com/containers/image/pull/1191#discussion_r610621608

 2. https://github.com/containers/image/pull/1191#discussion_r610122617

 3. https://github.com/containers/image/pull/1191#discussion_r610623216

 4. https://github.com/containers/image/pull/1191#discussion_r610623829
 
5. https://github.com/containers/image/pull/1191#discussion_r610622495

@vrothberg @mtrmac @ashcrow PTAL

### DIFF
--- a/pkg/sysregistriesv2/system_registries_v2.go
+++ b/pkg/sysregistriesv2/system_registries_v2.go
@@ -75,14 +75,10 @@ func (e *Endpoint) rewriteReference(ref reference.Named, prefix string) (referen
 	}
 	// In the case of an empty `location` field, simply return the original
 	// input ref as-is.
-	//
-	// FIXME: already validated in postProcessRegistries, so check can probably
-	// be dropped.
-	// https://github.com/containers/image/pull/1191#discussion_r610621608
-	if e.Location == "" {
-		if prefix[:2] != "*." {
-			return nil, fmt.Errorf("invalid prefix '%v' for empty location, should be in the format: *.example.com", prefix)
-		}
+	locationPrefixCheck := EmptyLocationWildcardedPrefixCheck(e.Location, prefix)
+	if locationPrefixCheck == -1 {
+		return nil, fmt.Errorf("invalid prefix '%v' for empty location, should be in the format: *.example.com", prefix)
+	} else if locationPrefixCheck == 1 {
 		return ref, nil
 	}
 	newNamedRef = e.Location + refString[prefixLen:]
@@ -271,18 +267,22 @@ func (e *InvalidRegistries) Error() string {
 func parseLocation(input string) (string, error) {
 	trimmed := strings.TrimRight(input, "/")
 
-	// FIXME: This check needs to exist but fails for empty Location field with
-	// wildcarded prefix. Removal of this check "only" allows invalid input in,
-	// and does not prevent correct operation.
-	// https://github.com/containers/image/pull/1191#discussion_r610122617
-	//
-	//	if trimmed == "" {
-	//		return "", &InvalidRegistries{s: "invalid location: cannot be empty"}
-	//	}
-	//
-
 	if strings.HasPrefix(trimmed, "http://") || strings.HasPrefix(trimmed, "https://") {
 		msg := fmt.Sprintf("invalid location '%s': URI schemes are not supported", input)
+		return "", &InvalidRegistries{s: msg}
+	}
+
+	return trimmed, nil
+}
+
+// parsePrefix parses the input string, performs some sanity checks and returns
+// the sanitized input string.  An error is returned if the input string is
+// empty or if contains an "http{s,}://" prefix.
+func parsePrefix(input string) (string, error) {
+	trimmed := strings.TrimRight(input, "/")
+
+	if strings.HasPrefix(trimmed, "http://") || strings.HasPrefix(trimmed, "https://") {
+		msg := fmt.Sprintf("invalid prefix '%s': URI schemes are not supported", input)
 		return "", &InvalidRegistries{s: msg}
 	}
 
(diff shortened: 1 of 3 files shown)


lsm5:
ready for review (afaict)

vrothberg:
Before I start looking at the changes: the commit message isn't really helpful since it merely points to 5 comments "somewhere". The reader has to put the puzzle pieces together of what the changes are intended to do but I really want the author to explain the changes in detail. Since there are 5 different comments, it may be worth splitting them into 5 different commits.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -343,6 +343,21 @@ func (config *V1RegistriesConf) ConvertToV2() (*V2RegistriesConf, error) {
 // anchoredDomainRegexp is an internal implementation detail of postProcess, defining the valid values of elements of UnqualifiedSearchRegistries.
 var anchoredDomainRegexp = regexp.MustCompile("^" + reference.DomainRegexp.String() + "$")
 
+// WildcardedPrefixCheck is the only place to check if a prefix contains "*."
+func WildcardedPrefixCheck(prefix string) bool {
```
vrothberg, line 347: Please make this private.  We try to keep new public APIs at the bare minimum to prevent bumping the major version.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -343,6 +343,21 @@ func (config *V1RegistriesConf) ConvertToV2() (*V2RegistriesConf, error) {
...
+	return prefix[:2] == "*."
+}
+
+func EmptyLocationWildcardedPrefixCheck(location string, prefix string) int {
```
vrothberg, line 351: Same here.  Please add a comment on the return value.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -357,36 +372,36 @@ func (config *V2RegistriesConf) postProcessRegistries() error {
...
+		if reg.Prefix != "" {
+			reg.Prefix, err = parsePrefix(reg.Prefix)
 			if err != nil {
 				return err
 			}
-			// FIXME: allow config authors to always use Prefix.
-			// https://github.com/containers/image/pull/1191#discussion_r610622495
-			if reg.Prefix[:2] != "*." && reg.Location == "" {
-				return &InvalidRegistries{s: "invalid condition: location is unset and prefix is not in the format: *.example.com"}
+			if reg.Location != "" {
```
vrothberg, line 380: This looks suspicious.  Why are we allowing a config with an empty location?

If that's desired, we need tests for that case and probably some sanity checks and some massaging of the docs.  For instance, unless it's blocked, it would require at least one mirror to make sense which in turn would bring up the question: why lift the requirement to specify a location,

lsm5:
> Before I start looking at the changes: the commit message isn't really helpful since it merely points to 5 comments "somewhere". The reader has to put the puzzle pieces together of what the changes are intended to do but I really want the author to explain the changes in detail. Since there are 5 different comments, it may be worth splitting them into 5 different commits.

sgtm, will do.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -357,36 +372,36 @@ func (config *V2RegistriesConf) postProcessRegistries() error {
...
+				return err
 			}
+			reg.Location = reg.Prefix
+		} else {
+			return &InvalidRegistries{s: "invalid condition: both location and prefix are unset"}
 		}
 
 		// make sure mirrors are valid
 		for _, mir := range reg.Mirrors {
+			if mir.Location == "" {
```
mtrmac, line 398: - I don’t see that the caller of `parseLocation` should need to specifically reject `""`
- Moving the check for `""` _before_ `parseLocation` changes semantics; `"/"` was previously rejected, now it is allowed.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -357,36 +372,36 @@ func (config *V2RegistriesConf) postProcessRegistries() error {
...
+				return &InvalidRegistries{s: "invalid condition: mirror location is unset"}
+			}
 			mir.Location, err = parseLocation(mir.Location)
 			if err != nil {
 				return err
 			}
-
-			//FIXME: unqualifiedSearchRegistries now also accepts empty values
```
mtrmac, line 384: Has anything been done to fix that?

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -919,17 +934,6 @@ func loadConfigFile(path string, forceV2 bool) (*parsedConfig, error) {
 		res.shortNameMode = types.ShortNameModeInvalid
 	}
 
-	// Valid wildcarded prefixes must be in the format: *.example.com
-	// FIXME: Move to postProcessRegistries
-	// https://github.com/containers/image/pull/1191#discussion_r610623829
-	for i := range res.partialV2.Registries {
-		prefix := res.partialV2.Registries[i].Prefix
-		if prefix[:2] == "*." && strings.ContainsAny(prefix, "/@:") {
```
mtrmac, line 927: What replaces the validation?

pkg/sysregistriesv2/system_registries_v2.go
mtrmac: See all the “prefix is a wildcard” and “prefix is empty” conditions in https://github.com/openshift/runtime-utils/pull/11/files `EditRegistriesConfig`.

Right now, if the user wants to set properties (insecure/blocked/mirrors) on a registry _without redirecting it_, there are two different syntaxes: (prefix=*.domain, location="") or (prefix="", location="non-wildcard.domain/…"), with little reason to for the two to differ. It would conceptually be much simpler if we always required `prefix` to exist as a lookup key, and made `location` entirely optional for redirection only; we can’t do that any more due to compatibility, but at least _allowing_ software that writes entries from scratch to always use `prefix` could simplify it.

lsm5:
Thanks for all the comments @mtrmac  @vrothberg . Closing this. I'll open a separate PR for each FIXME. Starting with: https://github.com/containers/image/pull/1368

PR #1368, opened 2021-09-01
Allow config authors to always use Prefix

lsm5:
Previously an unset Location was only allowed for wildcarded Prefixes.
This commit will allow any valid Prefix.

Fixes: https://github.com/containers/image/pull/1191#discussion_r610622495

@vrothberg @mtrmac @ashcrow  PTAL

### DIFF
--- a/docs/containers-registries.conf.5.md
+++ b/docs/containers-registries.conf.5.md
@@ -82,8 +82,7 @@ location = "internal-registry-for-example.net/bar"
 requests for the image `example.com/foo/myimage:latest` will actually work with the
 `internal-registry-for-example.net/bar/myimage:latest` image.
 
-With a `prefix` containing a wildcard in the format: "*.example.com" for subdomain matching,
-the location can be empty. In such a case,
+With any valid `prefix`, the location can be emtpy. In such a case,
 prefix matching will occur, but no reference rewrite will occur. The
 original requested image string will be used as-is. But other settings like
 `insecure` / `blocked` / `mirrors` will be applied to matching images.
@@ -91,6 +90,10 @@ original requested image string will be used as-is. But other settings like
 Example: Given
 ```
 prefix = "*.example.com"
+```
+OR
+```
+prefix = "blah.example.com"
 ```
 requests for the image `blah.example.com/foo/myimage:latest` will be used
 as-is. But other settings like insecure/blocked/mirrors will be applied to matching images
--- a/pkg/sysregistriesv2/system_registries_v2.go
+++ b/pkg/sysregistriesv2/system_registries_v2.go
@@ -75,14 +75,7 @@ func (e *Endpoint) rewriteReference(ref reference.Named, prefix string) (referen
 	}
 	// In the case of an empty `location` field, simply return the original
 	// input ref as-is.
-	//
-	// FIXME: already validated in postProcessRegistries, so check can probably
-	// be dropped.
-	// https://github.com/containers/image/pull/1191#discussion_r610621608
 	if e.Location == "" {
-		if prefix[:2] != "*." {
-			return nil, fmt.Errorf("invalid prefix '%v' for empty location, should be in the format: *.example.com", prefix)
-		}
 		return ref, nil
 	}
 	newNamedRef = e.Location + refString[prefixLen:]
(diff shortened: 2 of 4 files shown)


vrothberg (commented):
Needs tests and changes to docs.  AFAIR the docs state the location must be set, and we need to explain how rewriting works when location is empty.

The tests should make sure that rewriting the reference will return the input.

lsm5:
Done. PTAL.

lsm5:
@mtrmac removing the entire check for empty location from rewriteReference seemed to cause failures in `TestRewriteReferenceFailedDuringParseNamed` . So, let me know if the FIXME comment still needs to exist there.

docs/containers-registries.conf.5.md
```diff
@@ -82,7 +82,7 @@ location = "internal-registry-for-example.net/bar"
 requests for the image `example.com/foo/myimage:latest` will actually work with the
 `internal-registry-for-example.net/bar/myimage:latest` image.
 
-With a `prefix` containing a wildcard in the format: "*.example.com" for subdomain matching,
+With any valid `prefix` with or without a wildcard in the format: "*.example.com" for subdomain matching,
 the location can be empty. In such a case,
```
vrothberg, line 86: This sentence seems a bit too complicated. I suggest: "The `location` can be empty if the `prefix` is set."

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -357,21 +350,26 @@ func (config *V2RegistriesConf) postProcessRegistries() error {
 			return err
 		}
 
-		if reg.Prefix == "" {
-			if reg.Location == "" {
-				return &InvalidRegistries{s: "invalid condition: both location and prefix are unset"}
-			}
-			reg.Prefix = reg.Location
-		} else {
+		// Allow config authors to always use Prefix.
```
vrothberg, line 353: I don't find the comment helpful. I'd prefer a comment stating that either prefix OR location OR both must be set.

vrothberg (commented):
I'd love a test with at least one registries.conf file with an empty location and set prefix.

rhatdan:
@lsm5 any update on this PR?

lsm5:
@rhatdan working on it..

lsm5:
@mtrmac @vrothberg PTAL

vrothberg (commented):
LGTM
@mtrmac PTAL

lsm5:
@mtrmac ping PTAL

vrothberg:
> @mtrmac ping PTAL

On PTO until next week.

lsm5:
@mtrmac PTAL

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -357,21 +350,27 @@ func (config *V2RegistriesConf) postProcessRegistries() error {
...
 
-		if reg.Prefix == "" {
-			if reg.Location == "" {
-				return &InvalidRegistries{s: "invalid condition: both location and prefix are unset"}
-			}
-			reg.Prefix = reg.Location
-		} else {
+		// Prefix and Location cannot both be empty.
+		// Either one at least must be set.
+		if reg.Prefix != "" {
```
mtrmac, line 355: This could be simpler as
```go
if reg.Prefix == "" && reg.Location == "" { fail }
if reg.Location != "" { reg.Location, err = parseLocation(…); if err … }
if reg.Prefix == "" { reg.Prefix = reg.Location }
else { reg.Prefix, err = parseLocation(…); if err … } 
```

or perhaps
```go
// Validate individual fields
if reg.Location != "" { reg.Location, err = parseLocation(…); if err … }
if reg.Prefix != "" { reg.Prefix, err = parseLocation(…); if err … } 
// Validate the combination, normalize to have reg.Prefix always set
if reg.Prefix == "" {
    if reg.Location == "" { fail }
    else { reg.Prefix = reg.Location }
}
```

Either way, please remove the `parseLocation` call above this line, and make `parseLocation` fail on `""` again.

pkg/sysregistriesv2/system_registries_v2_test.go
```diff
@@ -383,6 +383,12 @@ func TestFindRegistry(t *testing.T) {
 	assert.Equal(t, "empty-prefix.com", reg.Prefix)
 	assert.Equal(t, "empty-prefix.com", reg.Location)
 
+	reg, err = FindRegistry(sys, "empty-location.set-prefix.example.com/namespace/repo")
+	assert.Nil(t, err)
+	assert.NotNil(t, reg)
```
mtrmac, line 388: (Non-blocking: These should be `require.`. OTOH this is consistent with the rest, and only matters on test failures.)

mtrmac (commented):
Thanks for the PR, and I apologize for the late review.

ACK overall; a few more simplifications seem possible.

pkg/sysregistriesv2/system_registries_v2_test.go
mtrmac: I honestly don’t think this is worth much worrying about; if you find updating pre-existing test cases fun, sure, go for it, but it’s not work that results in anything users could notice and consider valuable while _using_ the software. As far as “clean code” goes, `require` is a bit more correct but consistency is also nice, so to me that’s not a strong reason to prefer either.

(Just to be explicit: `require` should be used for testing prerequisites that must be true for other checks to be even valid to do, e.g.
```go
ptr, err := testSubject()
require.NoError(t, err)
assert.Equal(t, …, *ptr) // would crash if testSubject returned (nil, err) and the above were assert.NoError(…)
```

But there is a _minimal_ practical difference; with `assert.NoError(t, err)` the test crashes with a backtrace, with `require.NoError(t, err)` the test fails with an error message containing `err`. It doesn’t matter at all if the test completely succeeds. If the test fails because `err` is set, in both cases the test fails, it’s easy enough to identify the location of the failure, and probably something about `testSubject` needs to be updated. `require` is only better in that it includes the actual error text in the test failure output, and that could help with a inconsistent rarely-occurring failure in CI.)

jankaluza:

PR #1379, opened 2021-09-17
[RFC] oci: allow zstd for Docker images

giuseppe:
Allow to use the zstd compression for docker images now that Docker
supports it: https://github.com/moby/moby/pull/41759

Marked as RFC as I am not completely sure about its implications.

Without this patch though I am not able to build any container that uses a zstd compressed image as base:

```
$ cat Dockerfile 
FROM docker.io/gscrivano/zstd-chunked:fedora
RUN dnf install -y nginx && dnf clean all -y

$ buildah bud --format oci --squash  .
STEP 1/2: FROM docker.io/gscrivano/zstd-chunked:fedora
error creating build container: error preparing image configuration: error converting image "containers-storage:[overlay@~/.local/share/containers/storage+/run/user/1000/containers]@c309a2d1ba2e8026019aa4c987684a1d8f84d6bc598ff88d6afe1b4cbde4d70d" from "application/vnd.oci.image.manifest.v1+json" to "application/vnd.docker.distribution.manifest.v2+json": Error during manifest conversion: "application/vnd.oci.image.layer.v1.tar+zstd": zstd compression is not supported for docker images 2
```

### DIFF
--- a/image/oci.go
+++ b/image/oci.go
@@ -216,7 +216,7 @@ func (m *manifestOCI1) convertToManifestSchema2(_ context.Context, _ *types.Mani
 		case imgspecv1.MediaTypeImageLayerGzip:
 			layers[idx].MediaType = manifest.DockerV2Schema2LayerMediaType
 		case imgspecv1.MediaTypeImageLayerZstd:
-			return nil, fmt.Errorf("Error during manifest conversion: %q: zstd compression is not supported for docker images", layers[idx].MediaType)
+			layers[idx].MediaType = manifest.DockerV2Schema2LayerMediaTypeZstd
 		default:
 			return nil, fmt.Errorf("Unknown media type during manifest conversion: %q", layers[idx].MediaType)
 		}
--- a/image/oci_test.go
+++ b/image/oci_test.go
@@ -486,7 +486,7 @@ func TestConvertToManifestSchema2AllMediaTypes(t *testing.T) {
 	_, err := original.UpdatedImage(context.Background(), types.ManifestUpdateOptions{
 		ManifestMIMEType: manifest.DockerV2Schema2MediaType,
 	})
-	require.Error(t, err) // zstd compression is not supported for docker images
+	require.NoError(t, err)
 }
 
 func TestConvertToV2S2WithInvalidMIMEType(t *testing.T) {
--- a/manifest/manifest.go
+++ b/manifest/manifest.go
@@ -24,6 +24,8 @@ const (
 	DockerV2Schema2ConfigMediaType = "application/vnd.docker.container.image.v1+json"
 	// DockerV2Schema2LayerMediaType is the MIME type used for schema 2 layers.
 	DockerV2Schema2LayerMediaType = "application/vnd.docker.image.rootfs.diff.tar.gzip"
+	// DockerV2Schema2LayerMediaTypeZstd is the MIME type used for schema 2 layers that use the zstd compression.
+	DockerV2Schema2LayerMediaTypeZstd = "application/vnd.docker.image.rootfs.diff.tar.zst"
 	// DockerV2SchemaLayerMediaTypeUncompressed is the mediaType used for uncompressed layers.
 	DockerV2SchemaLayerMediaTypeUncompressed = "application/vnd.docker.image.rootfs.diff.tar"
 	// DockerV2ListMediaType MIME type represents Docker manifest schema 2 list


rhatdan:
LGTM
@vrothberg @mtrmac PTAL

manifest/manifest.go
```diff
@@ -24,6 +24,8 @@ const (
 	DockerV2Schema2ConfigMediaType = "application/vnd.docker.container.image.v1+json"
 	// DockerV2Schema2LayerMediaType is the MIME type used for schema 2 layers.
 	DockerV2Schema2LayerMediaType = "application/vnd.docker.image.rootfs.diff.tar.gzip"
+	// DockerV2Schema2LayerMediaTypeZstd is the MIME type used for schema 2 layers that use the zstd compression.
+	DockerV2Schema2LayerMediaTypeZstd = "application/vnd.docker.image.rootfs.diff.tar.zst"
```
mtrmac, line 28: If this is introduced, it must also exist in `schema2CompressionMIMETypeSets`.

image/oci.go
```diff
@@ -216,7 +216,7 @@ func (m *manifestOCI1) convertToManifestSchema2(_ context.Context, _ *types.Mani
 		case imgspecv1.MediaTypeImageLayerGzip:
 			layers[idx].MediaType = manifest.DockerV2Schema2LayerMediaType
 		case imgspecv1.MediaTypeImageLayerZstd:
-			return nil, fmt.Errorf("Error during manifest conversion: %q: zstd compression is not supported for docker images", layers[idx].MediaType)
+			layers[idx].MediaType = manifest.DockerV2Schema2LayerMediaTypeZstd
```
mtrmac, line 219: The other direction needs to exist (and be tested) as well.

---

What about `MediaTypeImageLayerNonDistributableZstd`? Intentionally unsupported? Actually supported? Forgotten about?

image/oci_test.go
```diff
@@ -486,7 +486,7 @@ func TestConvertToManifestSchema2AllMediaTypes(t *testing.T) {
 	_, err := original.UpdatedImage(context.Background(), types.ManifestUpdateOptions{
 		ManifestMIMEType: manifest.DockerV2Schema2MediaType,
 	})
-	require.Error(t, err) // zstd compression is not supported for docker images
+	require.NoError(t, err)
```
mtrmac, line 489: I’d expect a non-trivial test implementation here, detailing the expected results.

(Similarly for any other new conversions.

Overall most of the code in `c/image/manifest` and `c/image/image` should have good test coverage, because it can be tricky, it’s easy to miss corner cases in testing, and it has no state so it is fairly easy to write tests.)

mtrmac (changes requested):
Most importantly, where did
```go
	DockerV2Schema2LayerMediaTypeZstd = "application/vnd.docker.image.rootfs.diff.tar.zst"
```
come from? I can’t see anything in https://github.com/moby/moby/pull/41759 that would introduce, generate, or support such a value.

IOW, what did https://github.com/moby/moby/pull/41759 actually _define_ as the new format? I realize that it now _happens to accept_ Zstd via auto-detection, but what formats are valid for Buildah to generate?

Without defining that, this PR feels premature. (Or, alternatively, this might be the place where that gets defined, sure, but I suspect schema2 MIME types should at least involve the schema2 format owners.)

mtrmac:
> Without this patch though I am not able to build any container that uses a zstd compressed image as base:
> 
> ```
> …
> error creating build container: error preparing image configuration: error converting image "containers-storage:[overlay@~/.local/share/containers/storage+/run/user/1000/containers]@c309a2d1ba2e8026019aa4c987684a1d8f84d6bc598ff88d6afe1b4cbde4d70d" from "application/vnd.oci.image.manifest.v1+json" to "application/vnd.docker.distribution.manifest.v2+json": Error during manifest conversion: "application/vnd.oci.image.layer.v1.tar+zstd": zstd compression is not supported for docker images 2
> ```

To me, this just primarily looks like a long-standing Buildah design decision (problem?), in that it expects almost everything about the images to be simultaneously representable as schema2 _and_ OCI. See the two `unmarshalConvertedConfig` calls in `Builder.initConfig`, and how this fails converting to schema2 even if the user explicitly asked for an OCI-formatted output. If we are going to support any kind of long-term format evolution/divergence, I think that Buildah design constraint needs to be eliminated Buildah-side. (To be clear, it’s not at all obvious to me _how_ to cleanly design the build process, without this constraint; I just think that it needs to happen, somehow, eventually.)

There may well be other contributing factors to how this fails right now (e.g. it’s of dubious value to even talk about the compressed MIME types in manifests for in-c/storage decompressed values); I didn’t look into it closely.

giuseppe:
> > ```
> > 
> > ```
> 
> 
> To me, this just primarily looks like a long-standing Buildah design decision (problem?), in that it expects almost everything about the images to be simultaneously representable as schema2 _and_ OCI. See the two `unmarshalConvertedConfig` calls in `Builder.initConfig`, and how this fails converting to schema2 even if the user explicitly asked for an OCI-formatted output. If we are going to support any kind of long-term format evolution/divergence, I think that Buildah design constraint needs to be eliminated Buildah-side. (To be clear, it’s not at all obvious to me _how_ to cleanly design the build process, without this constraint; I just think that it needs to happen, somehow, eventually.)

thanks for the hint.  I tried to remove the `unmarshalConvertedConfig` to `manifest.DockerV2Schema2MediaType` when building an OCI image and I could build successfully the image (I also had to drop a check for number of layers == number of items in the history).

I am not familiar with this part of Buildah.  I don't know what the solution should be and what are the implications if I drop these check.  @nalind do you have any suggestions how we can solve this issue?

PR #1384, opened 2021-10-04
WIP - add shortnames.Prompt()

vrothberg:
Expose an API to render the prompt to select a pull candidate.
This API is needed to allow for podman-remote to share the same
prompting code with a local podman.


vrothberg:
@mtrmac, I started working on remote prompting. https://github.com/containers/common/pull/798 is doing the dance for generating pull candidates, and this PR here is meant to do the prompting in podman-remote.

vrothberg:
Eeek, doesn't even compile yet but I am more interested in a general ACK/NACK.

mtrmac:
Sure, it makes sense to reuse that code.

My first instinct was to expect a `Resolved`, or something like that, as an input, instead of just strings — but the libimage `options.copyFromRegistryDryRun` code paths seems to do more filtering than that. I’m not immediately sure why - AFAICS in the local pull case the user would be prompted for all options, including those excluded by `PullPolicyNewwer`, so why is the `PullCandidates` code filtering them out?

vrothberg:
> Sure, it makes sense to reuse that code.
> 
> My first instinct was to expect a `Resolved`, or something like that, as an input, instead of just strings —

I didn't feel comfortable throwing an entire `Resolved` over the REST API.

> but the libimage `options.copyFromRegistryDryRun` code paths seems to do more filtering than that. I’m not immediately sure why - AFAICS in the local pull case the user would be prompted for all options, including those excluded by `PullPolicyNewer`, so why is the `PullCandidates` code filtering them out?

That is a sharp finding.  I was overthinking it, I guess.

vrothberg:
> I didn't feel comfortable throwing an entire Resolved over the REST API.

Ah, and (maybe more importantly), the image may resolve to a local one (e.g., via ID or pull-never). In this case, we can treat it like a short name and hence cannot return a `Resolved`.

pkg/shortnames/shortnames.go
```diff
@@ -386,6 +380,39 @@ func Resolve(ctx *types.SystemContext, name string) (*Resolved, error) {
 	return resolved, nil
 }
 
+// Prompt opens a prompt in the terminal to let the user choose *one* of the
+// specified items.
```
mtrmac, line 384: “… of the specified fully-qualified image references”, or something like that? It’s not a general-purpose prompting utility.

pkg/shortnames/shortnames.go
```diff
@@ -386,6 +380,39 @@ func Resolve(ctx *types.SystemContext, name string) (*Resolved, error) {
...
+func Prompt(items []string) (string, error) {
+	switch len(items) {
+	case 0:
+		return fmt.Errorf("cannot prompt without items")
+	case 1:
+		return items[0]
+	default:
+		if !term.IsTerminal(int(os.Stdout.Fd())) || !term.IsTerminal(int(os.Stdin.Fd())) {
```
mtrmac, line 392: If we take the effort to avoid repeating the two syscalls for this check for the in-package caller, it might be worth it factoring the TTY check into a small helper function, to ensure both callers of `promptNoTTYCheck` do _exactly_ the same check and they don’t drift apart over time.

mtrmac:
> I didn't feel comfortable throwing an entire Resolved over the REST API.

Ah, that’s a good point, I was thinking in terms of the server-side `PullCandidates`. And the private fields of `Resolved` make no sense on the client.

mtrmac (commented):
(A few totally premature nits)

mtrmac:
A thought: maybe this should be its own subpackage, so that the remote client doesn’t grow any other dependencies (right now that’s `sysregistriesv2` and infrastructure like `internal/rootless` and `c/storage/pkg/homedir` — it’s not much but it’s not _zero_, and more importantly `pkg/shortnames` might grow more dependencies in the future).

I’m not sure splitting this is really _necessary_ but I don’t think it would make anything worse. We can always merge the implementations by just forwarding calls, but splitting the functionality once we have a stable API is not possible.

pkg/shortnames/shortnames.go
```diff
@@ -386,6 +384,55 @@ func Resolve(ctx *types.SystemContext, name string) (*Resolved, error) {
...
+		return "", fmt.Errorf("cannot prompt without items")
+	case 1:
+		// NOTE: return the original will make sure that information on
+		// how the (possibly short) name is resolved afterwards. For
+		// instance, `podman-remote pull short` will then print if
+		// `short` resolved to an alias.
+		//
+		// Returning `items[0]` would return a FQN and hence hide the
+		// fact that it's an alias when actually pulling down the
+		// image.
```
mtrmac, line 406: Thanks for the large comment, I would have missed this.

This feels unintuitive.
- Basically, the question is “what is the semantics of the return value?” “It’s one of the fully-qualified options in `items` that is safe to send to the remote server” is a nice and simple story; “it can be a fully-qualified option or a short name that needs to be processed $somehow” is unclear.
- The caller already has both `original` and `items`, so it can make whatever policy decisions it wants, based on any of `len(items) == 1`, `original \in items`, `original == Prompt()`, or other conditions.
  - (I wouldn’t even mind if `Prompt` just failed on `len(items) <= 1`, with the rationale that it’s a prompting utility, not a policy mechanism.)
- Compare how `Resolve` does return a `resolved` with full candidates, not some single-candidate special case.
- It seems to me that the caller already must have at least a _suspicion_ that the input is a short name or an alias, to even trigger a `PullCandidates` operation and to decide to invoke the `Prompt` UI. Whether the input is an alias or a short name is not a question `Prompt` can answer; OTOH the return value of `PullCandidates` should clearly indicate this already. (Maybe `PullCandidates` could need to return a `shortnames.rationale`-like value, but again I don’t think it’s too likely `Prompt` can help with reconstructing that if it is not provided by the server.)

There’s a probably a good reason for this, so I’m curious to see the caller PR which needs this semantics.

pkg/shortnames/shortnames.go
```diff
@@ -386,6 +384,55 @@ func Resolve(ctx *types.SystemContext, name string) (*Resolved, error) {
...
+		return original, nil
+	default:
+		if !canPromptWithTTY() {
+			if mode == types.ShortNameModeEnforcing {
+				return "", errNoTTY
+			}
+			return original, nil
```
mtrmac, line 413: OK, that’s sort of an argument to support returning `original`, but what’s the caller going to do when the remote server refuses to prompt locally, and the local client can’t prompt?

(I wouldn’t mind `Prompt` not dealing with `mode` at all, but that just moves the conversation to another PR.)

How can the code path returning `original` succeed?

pkg/shortnames/shortnames.go
vrothberg: The _main_ reason to have this logic in Prompt() is to avoid leaking any short-name mode logic outside of `pkg/shortnames`.  The logic of dealing with the short-name mode could be moved into Podman but that would scatter it a bit too much for my taste.

Maybe I am overly cautious?

pkg/shortnames/shortnames.go
vrothberg: The passed in `mode` is the one from the server.  Since we already now it's not enforcing, the server would then fall back to walking the search registries.

pkg/shortnames/shortnames.go
mtrmac: OK, so this is not just a prompting utility.

Ignore the above for now, and I’ll re-visit after understanding the full picture.

vrothberg: https://github.com/containers/podman/pull/11867 is putting the pieces together

pkg/shortnames/shortnames_test.go
```diff
@@ -515,3 +515,45 @@ func TestResolveLocally(t *testing.T) {
...
+	tests := []struct {
+		mode           types.ShortNameMode
+		original       string
+		items          []string
+		expected       string
+		mustFail       bool
+		errorSubstring string // no API guarantee for checking behavior
+	}{
+		{types.ShortNameModeEnforcing, "image:tag", []string{"fqn.com/image:tag"}, "image:tag", false, ""},
+		{types.ShortNameModeEnforcing, "image:tag", []string{"fqn.com/image:tag", "fqn.com/another/image:tag"}, "image:tag", true, errNoTTY.Error()},
```
mtrmac, line 542: Just to confirm, does this fail with `errNoTTY` when _interactively_ running `go test ./pkg/shortnames`?

pkg/shortnames/shortnames.go
mtrmac: Reading the Podman piece now, I don’t think this works quite correctly now: it _always_ tries to prompt (but perhaps fails in enforcing mode) if there are multiple options, but that’s not what happens locally with `ShortNameDisabled`, or non-interactive `ShortNamePermissive`.

So this should be closer to including the `switch mode` code from `Resolve`, but at a first glance the requirements look different enough that I’m not sure we can actually share that switch.

---

The `ShortName{Permissive,Disabled}` case does, though, convince me that the remote client might need to submit an unresolved short name to the server, at least in that case. (I’m not sure that the client should submit a short name in the alias case, but it might turn out to be the simplest approach.)

(I haven’t thought through what the `Resolved.Description` logging should like in the remote case; that might be a relevant design constraint; for now I’m just pointing out the need to handle `ShortName{Permissive,Disabled}`.)

mtrmac: > The main reason to have this logic in Prompt() is to avoid leaking any short-name mode logic outside of pkg/shortnames.

(FWIW if the logic ends up more complex, as it probably will, it totally makes sense to have the client-side the logic here; even if it had to be duplicated with no code sharing, better to have it duplicated in a single file than across repositories.)

PR #1385, opened 2021-10-04
Tag v5.16.1

mtrmac:
The v5.16.0 release has been triggering the https://github.com/vbauerster/mpb/issues/100 hang, causing unaware users who just update c/image (in particular CRI-O today) to encounter these hangs.

So, tag a new version where we depend on a newer MPB, to avoid breaking any other users.


vrothberg (approved):
LGTM

Issue #1392, opened 2021-10-10
Support precomputing digests prior to destination registry upload


paulnivin:
c/image does not support deduplicating layers when digests are computed by `PutBlob` (during layer push to destination registry). These duplicate layers are always transferred over the network, stored under a uuid on the registry, and then removed by the registry once c/image completes the upload by sending the registry the computed digest.

We have a common CI/CD use case running `skopeo copy` where images are sourced from `docker-daemon:` and copied to a destination of `docker:` with c/image compressing "on the fly" -- the vast majority of source layers already exist on the destination registry as compressed layers. Without a pre-populated BlobInfoCache containing compressed digests, skopeo is unable to skip these duplicate layers and they are always uploaded. c/image should have the ability to precompute digests and skip layers that already exist on the destination registry.

PR #1393, opened 2021-10-10
Precompute digests option prior to registry upload

paulnivin:
This ensures layers are not uploaded that already exist on the
destination registry, in exchange for streaming layers to temporary
files when digests are unknown (ex. compressing "on the fly").

Fixes https://github.com/containers/image/issues/1392

Signed-off-by: Paul Fisher <pfisher@lyft.com>


docker/docker_image_dest.go
```diff
@@ -131,15 +132,49 @@ func (d *dockerImageDestination) HasThreadSafePutBlob() bool {
...
+	stream = io.TeeReader(stream, sizeCounter)
+
+	// If requested, precompute the blob digest to prevent uploading layers that already exist on the registry.
+	// This functionality is particularly useful when BlobInfoCache has not been populated with compressed digests,
+	// the source blob is uncompressed, and the destination blob is being compressed "on the fly".
+	if inputInfo.Digest == "" && d.c.sys.DockerRegistryPushPrecomputeDigests {
+		logrus.Debugf("Precomputing digest layer for %s", reference.Path(d.ref.ref))
+		diskBlob, err := ioutil.TempFile(tmpdir.TemporaryDirectoryForBigFiles(d.c.sys), "docker-layer-blob")
+		if err != nil {
+			return types.BlobInfo{}, errors.Wrap(err, "creating temporary on-disk layer")
```
vrothberg, line 146: Can you turn `errors.Wrap(err, "")` into `fmt.Errorf("...: %w", err)`?  Old code is still using the errors package but we want new code to use the native error handling in Go.

docker/docker_image_dest.go
```diff
@@ -131,15 +132,49 @@ func (d *dockerImageDestination) HasThreadSafePutBlob() bool {
...
+		}
+		defer diskBlob.Close()
+		// immediately unlink for removal on fd close
+		err = os.Remove(diskBlob.Name())
```
mtrmac, line 150: This works on Linux but isn’t portable to Windows, AFAIK; so it should use `defer`.

docker/docker_image_dest.go
```diff
@@ -131,15 +132,49 @@ func (d *dockerImageDestination) HasThreadSafePutBlob() bool {
 // to any other readers for download using the supplied digest.
 // If stream.Read() at any time, ESPECIALLY at end of input, returns an error, PutBlob MUST 1) fail, and 2) delete any data stored so far.
 func (d *dockerImageDestination) PutBlob(ctx context.Context, stream io.Reader, inputInfo types.BlobInfo, cache types.BlobInfoCache, isConfig bool) (types.BlobInfo, error) {
+	digester, stream := putblobdigest.DigestIfCanonicalUnknown(stream, inputInfo)
+	sizeCounter := &sizeCounter{}
+	stream = io.TeeReader(stream, sizeCounter)
```
mtrmac, line 137: The way these values are applied/computed in different locations is non-obvious and doesn‘t look necessary to me, e.g. the way `digester` _exists_ during the upload but might actually only be computed on the digest pre-computation path.

I’d prefer for these to be initialized where they were, and for the new code to have its own digester (which can just be a `digest.Canonical.Digester` without using a redundant `DigestIfCanonicalUnknown` condition); we also don’t need two locations of `sizeCounter` at all, AFAICS.

docker/docker_image_dest.go
```diff
@@ -131,15 +132,49 @@ func (d *dockerImageDestination) HasThreadSafePutBlob() bool {
...
+		}
+		stream = diskBlob
+		inputInfo.Digest = digester.Digest()
+		inputInfo.Size = sizeCounter.size
+	}
+
 	if inputInfo.Digest != "" {
-		// This should not really be necessary, at least the copy code calls TryReusingBlob automatically.
-		// Still, we need to check, if only because the "initiate upload" endpoint does not have a documented "blob already exists" return value.
+		// The copy code calls TryReusingBlob automatically; however, we still need to perform an existing layer check for only this blob.
```
mtrmac, line 168: The copy code is not the _only_ anticipated caller of this method (`types.ImageDestination` is, for better or worse, a public type), so the comment should not be changed to assume only that caller.

Or if the change is trying to say something else non-obvious, I have missed the intent of that.

---

… is this trying to say that we have to try _again_ because we didn’t know the digest and now we maybe do because we computed it ourselves? It’s not actually saying that.

docker/docker_image_dest.go
```diff
@@ -131,15 +132,49 @@ func (d *dockerImageDestination) HasThreadSafePutBlob() bool {
...
+		// The "initiate upload" endpoint does not have a documented "blob already exists" return value.
 		// But we do that with NoCache, so that it _only_ checks the primary destination, instead of trying all mount candidates _again_.
 		haveBlob, reusedInfo, err := d.TryReusingBlob(ctx, inputInfo, none.NoCache, false)
 		if err != nil {
 			return types.BlobInfo{}, err
 		}
 		if haveBlob {
+			logrus.Debugf("Layer %s already exists, skipping upload", inputInfo.Digest)
+			cache.RecordKnownLocation(d.ref.Transport(), bicTransportScope(d.ref), inputInfo.Digest, newBICLocationReference(d.ref))
```
mtrmac, line 177: ?

`TryReusingBlob` already _both_ logs debug information, _and_ does `RecordKnownLocation`. Why is this necessary?

mtrmac (commented):
Thanks for the PR!

The _implementation_ approach is reasonable, but I’d very much like to discuss the motivating situation and explore alternatives first.

My general intuition is if at all possible, the reuse detection done by `TryReusingBlob` should succeed; can we explore that more?

In particular, in various past benchmarks the slowest part of an upload was almost always not the network cost of the transfer, but the CPU cost of compressing the blob; so this PR saves on the relatively cheap part (which can be done almost entirely in parallel with the compression) but not the expensive one — in most situations, I’d expect `DockerRegistryPushPrecomputeDigests` to be _slower even if it fully succeeds_ than the pre-existing code, although, yes, it would use significantly less upload bandwidth. That, to me, already significantly decreases the utility of this feature and the benefit that should motivate maintaining that extra, (because of the limited utility) rare code path long-term.

It’s quite possible that this is really the only reasonable approach to the problem in question, but at least right now it’s not obvious to me.

With #1392 filed, maybe we should discuss alternatives there?

docker/docker_image_dest.go
```diff
@@ -131,15 +132,49 @@ func (d *dockerImageDestination) HasThreadSafePutBlob() bool {
...
 // If stream.Read() at any time, ESPECIALLY at end of input, returns an error, PutBlob MUST 1) fail, and 2) delete any data stored so far.
 func (d *dockerImageDestination) PutBlob(ctx context.Context, stream io.Reader, inputInfo types.BlobInfo, cache types.BlobInfoCache, isConfig bool) (types.BlobInfo, error) {
+	digester, stream := putblobdigest.DigestIfCanonicalUnknown(stream, inputInfo)
+	sizeCounter := &sizeCounter{}
+	stream = io.TeeReader(stream, sizeCounter)
+
+	// If requested, precompute the blob digest to prevent uploading layers that already exist on the registry.
+	// This functionality is particularly useful when BlobInfoCache has not been populated with compressed digests,
+	// the source blob is uncompressed, and the destination blob is being compressed "on the fly".
+	if inputInfo.Digest == "" && d.c.sys.DockerRegistryPushPrecomputeDigests {
```
mtrmac, line 142: We already have almost the same code in `docker/internal/tarfile.PutBlob`.

Maybe that could be factored out into a separate helper (`internal/{streamdigest,savestream,…}`, I don’t know what’s a good name but I also don’t worry about internal names too much) — that would make the size of the feature in this file < 5 lines and pretty much make my concerns about taking on long-term maintenance a rarely-used feature go completely away.

paulnivin: Sounds good. I'll work on a refactor and will update the PR.

docker/docker_image_dest.go
paulnivin: Done

internal/streamdigest/stream_digest.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/containers/image/v5/internal/tmpdir"
+	"github.com/containers/image/v5/types"
+	digest "github.com/opencontainers/go-digest"
+)
+
+// ComputeBlobInfo streams a blob to a temporary file and populates Digest and Size in inputInfo.
+// The temporary file is returned with the file position set to the beginning of the file.
+// It is the caller's responsibility to remove the file when no longer needed.
+// If an error occurs, inputInfo is not modified.
+func ComputeBlobInfo(sys *types.SystemContext, stream io.Reader, inputInfo *types.BlobInfo) (*os.File, error) {
```
mtrmac, line 18: Non-blocking, this is an internal interface we can change any time if this became important:

A `func ComputeBlobInfo(…) (stream io.Reader, cleanup func(), error)` (with `cleanup()` doing  `streamCopy.Close()` and `os.Remove(streamCopy.Name())`) would encapsulate the implementation, and more importantly simplify callers, further. (The `cleanup` closure could even simplify error handling inside this function.)

paulnivin: Changed the interface to return a cleanup closure.

vrothberg (commented):
LGTM
Nice work!

paulnivin:
> LGTM as well — please squash the commits and we’ll merge.

Done. Thanks for the assistance with upstreaming. I'll submit the follow-on PR for skopeo shortly.

PR #1402, opened 2021-11-06
sif: initial sif transport implementation

yhcote:
This implements the support for inbound SIF files in the image package.
 - Outbound support coming soon.
 - This version requires fakeroot and squashfs-tools distribution packages. Near future intent is to do all this in go, starting with replacing the use of fakeroot.
 - Go tests coming next.

### DIFF
--- a/sif/internal/sif_util.go
+++ b/sif/internal/sif_util.go
@@ -1,0 +1,254 @@
+package internal
+
+import (
+	"bufio"
+	"bytes"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+
+	"github.com/opencontainers/go-digest"
+	"github.com/pkg/errors"
+	uuid "github.com/satori/go.uuid"
+	"github.com/yhcote/sif/pkg/sif"
+
+	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+type SifImage struct {
+	fimg      sif.FileImage
+	rootfs    *sif.Descriptor
+	deffile   *sif.Descriptor
+	defReader *io.SectionReader
+	cmdlist   []string
+	runscript *bytes.Buffer
+	env       *sif.Descriptor
+	envReader *io.SectionReader
+	envlist   []string
+	diffID    digest.Digest
+}
+
+func LoadSIFImage(path string) (image SifImage, err error) {
+	// open up the SIF file and get its header
+	image.fimg, err = sif.LoadContainer(path, true)
+	if err != nil {
+		return
+	}
+
+	// check for a system partition and save it
+	image.rootfs, _, err = image.fimg.GetPartPrimSys()
+	if err != nil {
+		return SifImage{}, errors.Wrap(err, "looking up rootfs from SIF file")
+	}
+
+	// look for a definition file object
+	searchDesc := sif.Descriptor{Datatype: sif.DataDeffile}
+	resultDescs, _, err := image.fimg.GetFromDescr(searchDesc)
+	if err == nil && resultDescs != nil {
+		// we assume in practice that typical SIF files don't hold multiple deffiles
+		image.deffile = resultDescs[0]
+		image.defReader = io.NewSectionReader(image.fimg.Fp, image.deffile.Fileoff, image.deffile.Filelen)
+	}
+	if err = image.generateConfig(); err != nil {
+		return SifImage{}, err
+	}
+
+	// look for an environment variable set object
+	searchDesc = sif.Descriptor{Datatype: sif.DataEnvVar}
+	resultDescs, _, err = image.fimg.GetFromDescr(searchDesc)
+	if err == nil && resultDescs != nil {
+		// we assume in practice that typical SIF files don't hold multiple EnvVar sets
+		image.env = resultDescs[0]
+		image.envReader = io.NewSectionReader(image.fimg.Fp, image.env.Fileoff, image.env.Filelen)
+	}
+
+	return image, nil
+}
+
+func (image *SifImage) parseEnvironment(scanner *bufio.Scanner) error {
+	for scanner.Scan() {
+		s := strings.TrimSpace(scanner.Text())
+		if s == "" || strings.HasPrefix(s, "#") {
+			continue
+		}
+		if strings.HasPrefix(s, "%") {
+			return nil
+		}
+		image.envlist = append(image.envlist, s)
+	}
+	if err := scanner.Err(); err != nil {
+		return errors.Wrap(err, "parsing environment from SIF definition file object")
+	}
+	return nil
+}
+
+func (image *SifImage) parseRunscript(scanner *bufio.Scanner) error {
+	for scanner.Scan() {
+		s := strings.TrimSpace(scanner.Text())
+		if strings.HasPrefix(s, "%") {
+			return nil
+		}
+		image.cmdlist = append(image.cmdlist, s)
+	}
+	if err := scanner.Err(); err != nil {
+		return errors.Wrap(err, "parsing runscript from SIF definition file object")
+	}
+	return nil
+}
+
+func (image *SifImage) generateRunscript() error {
+	base := `#!/bin/bash
+`
+	image.runscript = bytes.NewBufferString(base)
+	for _, s := range image.envlist {
+		_, err := image.runscript.WriteString(fmt.Sprintln(s))
+		if err != nil {
+			return errors.Wrap(err, "writing to runscript buffer")
+		}
+	}
+	for _, s := range image.cmdlist {
+		_, err := image.runscript.WriteString(fmt.Sprintln(s))
+		if err != nil {
+			return errors.Wrap(err, "writing to runscript buffer")
+		}
+	}
+	return nil
+}
+
+func (image *SifImage) generateConfig() error {
+	if image.deffile == nil {
+		image.cmdlist = append(image.cmdlist, "bash")
+		return nil
+	}
+
+	// extract %environment/%runscript from definition file
+	var err error
+	scanner := bufio.NewScanner(image.defReader)
+	for scanner.Scan() {
+		s := strings.TrimSpace(scanner.Text())
+	again:
+		if s == `%environment` {
+			if err = image.parseEnvironment(scanner); err != nil {
+				return err
+			}
+		} else if s == `%runscript` {
+			if err = image.parseRunscript(scanner); err != nil {
+				return err
+			}
+		}
+		s = strings.TrimSpace(scanner.Text())
+		if s == `%environment` || s == `%runscript` {
+			goto again
+		}
+	}
+	if err := scanner.Err(); err != nil {
+		return errors.Wrap(err, "reading lines from SIF definition file object")
+	}
+
+	if len(image.cmdlist) == 0 && len(image.envlist) == 0 {
+		image.cmdlist = append(image.cmdlist, "bash")
+	} else {
+		image.generateRunscript()
+		image.cmdlist = []string{"/podman/runscript"}
+	}
+
+	return nil
+}
+
+func (image SifImage) GetConfig(config *imgspecv1.Image) error {
+	config.Config.Cmd = append(config.Config.Cmd, image.cmdlist...)
+	return nil
+}
+
+func (image SifImage) UnloadSIFImage() (err error) {
+	err = image.fimg.UnloadContainer()
+	return
+}
+
+func (image SifImage) GetSIFID() string {
+	return image.fimg.Header.ID.String()
+}
+
+func (image SifImage) GetSIFArch() string {
+	return sif.GetGoArch(string(image.fimg.Header.Arch[:sif.HdrArchLen-1]))
+}
+
+const squashFilename = "rootfs.squashfs"
+const tarFilename = "rootfs.tar"
+
+func runUnSquashFSTar(tempdir string) (err error) {
+	script := `
+#!/bin/sh
+unsquashfs -f ` + squashFilename + ` && tar --acls --xattrs -C ./squashfs-root -cpf ` + tarFilename + ` ./
+`
+
+	if err = ioutil.WriteFile(filepath.Join(tempdir, "script"), []byte(script), 0755); err != nil {
+		return err
+	}
+	cmd := []string{"fakeroot", "--", "./script"}
+
+	xcmd := exec.Command(cmd[0], cmd[1:]...)
+	xcmd.Stderr = os.Stderr
+	xcmd.Dir = tempdir
+	err = xcmd.Run()
+	return
+}
+
+func (image *SifImage) writeRunscript(tempdir string) (err error) {
+	if image.runscript == nil {
+		return nil
+	}
+	rsPath := filepath.Join(tempdir, "squashfs-root", "podman")
+	if err = os.MkdirAll(rsPath, 0755); err != nil {
+		return
+	}
+	if err = ioutil.WriteFile(filepath.Join(rsPath, "runscript"), image.runscript.Bytes(), 0755); err != nil {
+		return errors.Wrap(err, "writing /podman/runscript")
+	}
+	return nil
+}
+
+func (image SifImage) SquashFSToTarLayer(tempdir string) (tarpath string, err error) {
+	if _, err = image.fimg.Fp.Seek(image.rootfs.Fileoff, 0); err != nil {
+		return
+	}
+	f, err := os.Create(filepath.Join(tempdir, squashFilename))
+	if err != nil {
+		return
+	}
+	defer f.Close()
+	if _, err = io.CopyN(f, image.fimg.Fp, image.rootfs.Filelen); err != nil {
+		return
+	}
+	if err = f.Sync(); err != nil {
+		return
+	}
+	if err = image.writeRunscript(tempdir); err != nil {
+		return
+	}
+	if err = runUnSquashFSTar(tempdir); err != nil {
+		return
+	}
+	return filepath.Join(tempdir, tarFilename), nil
+}
+
+func createSIF() error {
+	cinfo := sif.CreateInfo{
+		Pathname:   "container.sif",
+		Launchstr:  sif.HdrLaunch,
+		Sifversion: sif.HdrVersion,
+		ID:         uuid.NewV4(),
+	}
+
+	image, err := sif.CreateContainer(cinfo)
+	if err != nil {
+		return err
+	}
+	fmt.Printf("SIF: %+v\n", image)
+
+	return nil
+}
--- a/sif/internal/sif_util.go
+++ b/sif/internal/sif_util.go
@@ -11,10 +11,10 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/containers/image/v5/sif/sif"
 	"github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
 	uuid "github.com/satori/go.uuid"
-	"github.com/yhcote/sif/pkg/sif"
 
 	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
 )
--- a/sif/internal/sif_util.go
+++ b/sif/internal/sif_util.go
@@ -14,9 +14,7 @@ import (
 	"strings"
 
 	"github.com/containers/image/v5/sif/sif"
-	"github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
-	uuid "github.com/satori/go.uuid"
 
 	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
 )
@@ -31,7 +29,6 @@ type SifImage struct {
 	env       *sif.Descriptor
 	envReader *io.SectionReader
 	envlist   []string
-	diffID    digest.Digest
 }
 
 func LoadSIFImage(path string) (image SifImage, err error) {
@@ -154,7 +151,9 @@ func (image *SifImage) generateConfig() error {
 	if len(image.cmdlist) == 0 && len(image.envlist) == 0 {
 		image.cmdlist = append(image.cmdlist, "bash")
 	} else {
-		image.generateRunscript()
+		if err = image.generateRunscript(); err != nil {
+			return errors.Wrap(err, "generating runscript")
+		}
 		image.cmdlist = []string{"/podman/runscript"}
 	}
 
@@ -236,21 +235,4 @@ func (image SifImage) SquashFSToTarLayer(tempdir string) (tarpath string, err er
 		return
 	}
 	return filepath.Join(tempdir, tarFilename), nil
-}
-
-func createSIF() error {
-	cinfo := sif.CreateInfo{
-		Pathname:   "container.sif",
-		Launchstr:  sif.HdrLaunch,
-		Sifversion: sif.HdrVersion,
-		ID:         uuid.NewV4(),
-	}
-
-	image, err := sif.CreateContainer(cinfo)
-	if err != nil {
-		return err
-	}
-	fmt.Printf("SIF: %+v\n", image)
-
-	return nil
 }
--- a/sif/internal/sif_util.go
+++ b/sif/internal/sif_util.go
@@ -1,3 +1,5 @@
+// +build linux
+
 package internal
 
 import (
--- a/sif/sif/init.go
+++ b/sif/sif/init.go
@@ -1,0 +1,21 @@
+// Copyright (c) 2018, Sylabs Inc. All rights reserved.
+// Copyright (c) 2017, SingularityWare, LLC. All rights reserved.
+// Copyright (c) 2017, Yannick Cote <yhcote@gmail.com> All rights reserved.
+// This software is licensed under a 3-clause BSD license. Please consult the
+// LICENSE file distributed with the sources of this project regarding your
+// rights to use or distribute this software.
+
+package sif
+
+import (
+	"bytes"
+	"log"
+)
+
+var (
+	sifLoggerBuf bytes.Buffer
+	siflog       = log.New(&sifLoggerBuf, "", log.Ldate|log.Ltime|log.Lshortfile)
+)
+
+func init() {
+}
--- a/sif/sif/init.go
+++ b/sif/sif/init.go
@@ -5,6 +5,8 @@
 // LICENSE file distributed with the sources of this project regarding your
 // rights to use or distribute this software.
 
+// +build linux
+
 package sif
 
 import (
--- a/sif/sif/sif.go
+++ b/sif/sif/sif.go
@@ -5,6 +5,8 @@
 // LICENSE file distributed with the sources of this project regarding your
 // rights to use or distribute this software.
 
+// +build linux
+
 // Package sif implements data structures and routines to create
 // and access SIF files.
 // 	- sif.go contains the data definition the file format.
@@ -18,8 +20,9 @@ package sif
 
 import (
 	"bytes"
-	"github.com/satori/go.uuid"
 	"os"
+
+	uuid "github.com/satori/go.uuid"
 )
 
 // Layout of a SIF file (example)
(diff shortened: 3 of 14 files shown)


sif/internal/sif_util.go
```diff
@@ -0,0 +1,254 @@
...
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+
+	"github.com/opencontainers/go-digest"
+	"github.com/pkg/errors"
+	uuid "github.com/satori/go.uuid"
+	"github.com/yhcote/sif/pkg/sif"
```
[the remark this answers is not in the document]
mtrmac: This is a fork of a repo that claims to be a “reference implementation” of SIF. If that’s true, it would be ideal (I have no idea if feasible) to actually use that implementation, possibly contributing improvements there, rather than commit c/image to maintaining a long-term internal fork (“This branch is 314 commits behind hpcng:master.”)

(I’m afraid I haven’t yet reviewed this PR, nor the changes in that fork, so I don’t have an opinion on what’s the right approach yet, at all. At this point I just want to make sure the trade-offs are recognized.)

tri-adam:
@yhcote @vrotberg @mtrmac would love to collaborate with you all on this. We're actively maintaining a [`github.com/sylabs/sif/v2`](https://github.com/sylabs/sif) module and would welcome a discussion about re-use of (at a minimum) the core SIF implementation.

yhcote:
> @yhcote @Vrotberg @mtrmac would love to collaborate with you all on this. We're actively maintaining a [`github.com/sylabs/sif/v2`](https://github.com/sylabs/sif) module and would welcome a discussion about re-use of (at a minimum) the core SIF implementation.

@tri-adam I checked quickly and saw two different repos with SIF updates. Since the current feature requires simple data object dumps from the SIF file the current import is sufficient. The best course of action right now to get the ball rolling is to at least get this initial PR in and let the community decide which improvements (some already listed in the PR comment) they want to see next. That PR needs a few touch ups but it should be done this week.

tri-adam:
> @tri-adam I checked quickly and saw two different repos with SIF updates.

Yes, there is `github.com/sylabs/sif/v2` and `github.com/hpcng/sif/v2`. `github.com/hpcng/sif/v2` is not actively maintained, other than occasionally releasing code from `github.com/sylabs/sif/v2` (ex. https://github.com/hpcng/sif/releases/tag/v2.0.0).

`github.com/sylabs/sif/v2` contains both features and bug fixes not yet found in the other module. For that reason, I'd expect `github.com/sylabs/sif/v2` to be the natural choice if there's interest in code re-use.

> Since the current feature requires simple data object dumps from the SIF file the current import is sufficient. The best course of action right now to get the ball rolling is to at least get this initial PR in and let the community decide which improvements (some already listed in the PR comment) they want to see next.

Unless I'm missing something, using `github.com/sylabs/sif/v2` rather than duplicating the code in `containers/image/sif/sif` would be straight forward, and mean less code to maintain in `containers/image`. I've taken a stab at that here, as a practical example: https://github.com/tri-adam/image/commit/1dd4610f563f37a4b8461e9f2f6e23f51698cdd6.

Anyways, just want to make it crystal clear that we're open to collaboration on this. Excited to see where it goes, in any case!

rhatdan:
@mtrmac @vrothberg Can we get a fresh review on this?

sif/internal/sif_util.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+	f, err := os.Create(filepath.Join(tempdir, squashFilename))
+	if err != nil {
+		return
+	}
+	defer f.Close()
+	if _, err = io.CopyN(f, image.fimg.Fp, image.rootfs.Filelen); err != nil {
+		return
+	}
+	if err = f.Sync(); err != nil {
```
mtrmac, line 228: Why is this necessary?

sif/internal/sif_util.go
```diff
@@ -0,0 +1,238 @@
...
+		return errors.Wrap(err, "writing /podman/runscript")
+	}
+	return nil
+}
+
+func (image SifImage) SquashFSToTarLayer(tempdir string) (tarpath string, err error) {
+	if _, err = image.fimg.Fp.Seek(image.rootfs.Fileoff, 0); err != nil {
+		return
+	}
+	f, err := os.Create(filepath.Join(tempdir, squashFilename))
```
mtrmac, line 220: Shouldn’t the file be removed by the time this function returns, so that we don’t unnecessarily use up disk space for another copy of the data?

---

(Do we even need a copy? `unsquashfs` has an `-offset` option — but I know nothing about the format and I haven’t tried using it.)

sif/internal/sif_util.go
```diff
@@ -0,0 +1,238 @@
...
+	xcmd.Dir = tempdir
+	err = xcmd.Run()
+	return
+}
+
+func (image *SifImage) writeRunscript(tempdir string) (err error) {
+	if image.runscript == nil {
+		return nil
+	}
+	rsPath := filepath.Join(tempdir, "squashfs-root", "podman")
```
mtrmac, line 206: If `runUnSquashFSTar` depends on the `squashfs-root` name, it would be nice to make it a Go constant to make that dependency visible. (Also see elsewhere.)

sif/internal/sif_util.go
```diff
@@ -0,0 +1,238 @@
...
+unsquashfs -f ` + squashFilename + ` && tar --acls --xattrs -C ./squashfs-root -cpf ` + tarFilename + ` ./
+`
+
+	if err = ioutil.WriteFile(filepath.Join(tempdir, "script"), []byte(script), 0755); err != nil {
+		return err
+	}
+	cmd := []string{"fakeroot", "--", "./script"}
+
+	xcmd := exec.Command(cmd[0], cmd[1:]...)
+	xcmd.Stderr = os.Stderr
```
mtrmac, line 196: All of the output should be captured and provided to the caller (on error only, I guess); a library like c/image shouldn’t just write unexpected data to the standard streams. Probably using `xcmd.CombinedOutput()`.

sif/internal/sif_util.go
```diff
@@ -0,0 +1,238 @@
...
+func runUnSquashFSTar(tempdir string) (err error) {
+	script := `
+#!/bin/sh
+unsquashfs -f ` + squashFilename + ` && tar --acls --xattrs -C ./squashfs-root -cpf ` + tarFilename + ` ./
+`
+
+	if err = ioutil.WriteFile(filepath.Join(tempdir, "script"), []byte(script), 0755); err != nil {
+		return err
+	}
+	cmd := []string{"fakeroot", "--", "./script"}
```
mtrmac, line 193: Non-blocking: I don’t think we need this single-use variable; this can be just constants passed to `exec.Command`.

And then `s/xcmd/cmd/`.

sif/internal/sif_util.go
```diff
@@ -0,0 +1,238 @@
...
+	return sif.GetGoArch(string(image.fimg.Header.Arch[:sif.HdrArchLen-1]))
+}
+
+const squashFilename = "rootfs.squashfs"
+const tarFilename = "rootfs.tar"
+
+func runUnSquashFSTar(tempdir string) (err error) {
+	script := `
+#!/bin/sh
+unsquashfs -f ` + squashFilename + ` && tar --acls --xattrs -C ./squashfs-root -cpf ` + tarFilename + ` ./
```
mtrmac, line 187: Shouldn’t `shquashfs-root` be removed immediately afterwards, to decrease disk space usage?

sif/internal/sif_util.go
```diff
@@ -0,0 +1,238 @@
...
+		return
+	}
+	defer f.Close()
+	if _, err = io.CopyN(f, image.fimg.Fp, image.rootfs.Filelen); err != nil {
+		return
+	}
+	if err = f.Sync(); err != nil {
+		return
+	}
+	if err = image.writeRunscript(tempdir); err != nil {
```
mtrmac, line 231: (Non-blocking? It’s non-obvious that `writeRunscripts` affects the output of `runUnSquashFSTar`, which at a first glance seems to create the complete input for the tar file. That seems with at least a comment somewhere, or possibly some kind of restructuring — maybe inlining `writeRunscript` into `runUnSquashFSTar`, and renaming the latter one to something like `createTar`.)

sif/internal/sif_util.go
```diff
@@ -0,0 +1,238 @@
...
+
+func (image SifImage) GetSIFID() string {
+	return image.fimg.Header.ID.String()
+}
+
+func (image SifImage) GetSIFArch() string {
+	return sif.GetGoArch(string(image.fimg.Header.Arch[:sif.HdrArchLen-1]))
+}
+
+const squashFilename = "rootfs.squashfs"
```
mtrmac, line 181: (Absolutely non-blocking: Below is not _too_ much code that it would be unmanageable, still, the way every function gets a `tempdir` and implicitly knowns its full structure has unnecessary implicit dependencies.

A top-level `SquashFSToTarLayer` “driver” that determines `squashPath`/`tarPath`/`extractedPath` and passes those to its subroutines, which don’t make any assumptions about the complete structure of the directory, would be a bit more explicit.)

sif/internal/sif_util.go
```diff
@@ -0,0 +1,238 @@
...
+}
+
+func (image SifImage) GetConfig(config *imgspecv1.Image) error {
+	config.Config.Cmd = append(config.Config.Cmd, image.cmdlist...)
+	return nil
+}
+
+func (image SifImage) UnloadSIFImage() (err error) {
+	err = image.fimg.UnloadContainer()
+	return
```
mtrmac, line 170:
```suggestion
func (image SifImage) UnloadSIFImage() error {
	return image.fimg.UnloadContainer()
```

sif/internal/sif_util.go
```diff
@@ -0,0 +1,238 @@
...
+		if s == `%environment` {
+			if err = image.parseEnvironment(scanner); err != nil {
+				return err
+			}
+		} else if s == `%runscript` {
+			if err = image.parseRunscript(scanner); err != nil {
+				return err
+			}
+		}
+		s = strings.TrimSpace(scanner.Text())
```
mtrmac, line 142: This seems to be redundant with the line at the top of the loop at a first glance, and is very non-obvious. At the very least
```go
for scanner.Scan() {
again:
	// parseEnvironment and parseRunscript, will, on success, scan one line past its section, so we need to
	// return here without going through scanner.Scan()
	s := strings.TrimSpace(scanner.Text())
	if s == `%environment` {
		if err = image.parseEnvironment(scanner); err != nil {
			return err
		}
		goto again
	} else if s == `%runscript` {
		if err = image.parseRunscript(scanner); err != nil {
			return err
		}
		goto again
	}
}
```
including the comment, but that AFAICS doesn’t handle EOF in `parseEnvironment`/`parseRunscript` correctly, calling `.Text()` after EOF. That’s not _invalid_ but the value is not obvious — I’d expect a trailing `%environment` line to cause an infinite loop, without checking in detail.

So, how about something vaguely like
```go
// enum state = parsingEnvironment/parsingRunscript/parsingOther
for scanner.Scan() {
	s := strings.TrimSpace(scanner.Text())
	switch {
	case s == "%environment": state = parsingEnvironment
	case s == "%runscript": state = parsingRunscript
	case strings.HasPrefix(s, "%"): state = parsingOther // Warn on unrecognized?
	case state == parsingEnvironment: image.parseEnvironmentEntry(s)
	case state == parsingRunscript: image.parseRunscriptEntry(s)
	default: // ignore s, or warn?
    }
}
```
?

sif/internal/sif_util.go
```diff
@@ -0,0 +1,238 @@
...
+`
+	image.runscript = bytes.NewBufferString(base)
+	for _, s := range image.envlist {
+		_, err := image.runscript.WriteString(fmt.Sprintln(s))
+		if err != nil {
+			return errors.Wrap(err, "writing to runscript buffer")
+		}
+	}
+	for _, s := range image.cmdlist {
+		_, err := image.runscript.WriteString(fmt.Sprintln(s))
```
mtrmac, line 113: ⚠️ Is it intentional that `Sprintln` interprets Go format strings in the data? That seems very surprising.

sif/internal/sif_util.go
```diff
@@ -0,0 +1,238 @@
...
+	image.rootfs, _, err = image.fimg.GetPartPrimSys()
+	if err != nil {
+		return SifImage{}, errors.Wrap(err, "looking up rootfs from SIF file")
+	}
+
+	// look for a definition file object
+	searchDesc := sif.Descriptor{Datatype: sif.DataDeffile}
+	resultDescs, _, err := image.fimg.GetFromDescr(searchDesc)
+	if err == nil && resultDescs != nil {
+		// we assume in practice that typical SIF files don't hold multiple deffiles
```
mtrmac, line 51: Should there be a warning, or at least a debug log, in that case?

sif/internal/sif_util.go
```diff
@@ -0,0 +1,238 @@
...
+		image.deffile = resultDescs[0]
+		image.defReader = io.NewSectionReader(image.fimg.Fp, image.deffile.Fileoff, image.deffile.Filelen)
```
mtrmac, line 53: `defReader` is only used in `generateConfig` immediately below, so maybe it can just be a local variable.

sif/internal/sif_util.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+	if err = image.generateConfig(); err != nil {
+		return SifImage{}, err
+	}
+
+	// look for an environment variable set object
+	searchDesc = sif.Descriptor{Datatype: sif.DataEnvVar}
+	resultDescs, _, err = image.fimg.GetFromDescr(searchDesc)
+	if err == nil && resultDescs != nil {
+		// we assume in practice that typical SIF files don't hold multiple EnvVar sets
```
mtrmac, line 63: Should there be a warning, or at least a debug log, in that case?

sif/internal/sif_util.go
```diff
@@ -0,0 +1,238 @@
...
+		image.env = resultDescs[0]
+		image.envReader = io.NewSectionReader(image.fimg.Fp, image.env.Fileoff, image.env.Filelen)
+	}
```
mtrmac, line 66: I can’t see any use of `.env` or `.envReader`. What does this code do?

sif/sif/init.go
```diff
@@ -0,0 +1,23 @@
...
+
+package sif
+
+import (
+	"bytes"
+	"log"
+)
+
+var (
+	sifLoggerBuf bytes.Buffer
```
mtrmac, line 18: Is this available in any way? I can’t see a way to read this data, which makes the logger pointless.

(c/image otherwise uses `logrus`).

mtrmac (commented):
I didn’t check the `sif/sif` package in detail yet — I do default to thinking we should use the maintained external dependency instead, but I’ll check the code etc. first.

Highlights:
- <s>`Sprintf` on untrusted input</s>
- Do we want a temporary compressed representation at all?
- Removing on-disk data immediately after use
- Policy namespaces are inconsistent (and must have tests even the final design is trivial)

sif/internal/sif_util.go
```diff
@@ -0,0 +1,238 @@
+// +build linux
+
+package internal
```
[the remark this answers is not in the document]
yhcote: I am following the pattern of previous transport here.

sif/internal/sif_util.go
yhcote: have you tested it?

sif/internal/sif_util.go
yhcote: how so?

sif/internal/sif_util.go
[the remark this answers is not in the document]
mtrmac: `oci/internal/oci_util.go` exists to share code between `oci/layout` and `oci/archive`; that’s not quite the situation here.

sif/internal/sif_util.go
mtrmac: No. I’m not claiming that the code as is doesn’t work (well, apart from a trailing directive?), just that it looks very non-obvious.

If there are even more non-obvious gotchas that I didn’t notice by reading the code as is, those gotchas should be commented in detail and/or covered by unit tests.

sif/internal/sif_util.go
mtrmac: What if cmdlist contains `Hello %s%50d%n`? That’s not going to result in anything _useful_ in the generated script AFAICS.

Luckily, with Go, it’s not a trivially-exploited arbitrary-code-execution vulnerability, but it’s both useless and it exposes the format string processing code to arbitrary potentially-malicious input, something it’s not really intended for.

yhcote: After two months I need to confess that I need to sit down and get back into the context, and I will do that sometime soon. Thanks for the thorough review BTW, but are we not talking about an Sprint(a) vs an Sprintf(format, a) construct here, where Sprintln(a) does not interpret anything? Or are you trying to make another point?

mtrmac: You’re right, my mistake; this is not interpreting as a format string, so it is safe.

---

(It doesn’t quite make sense to me to do it this way — either performance and memory use is critical, in which case `runscript.WriteString(entry); runscript.WriteByte('\n')` (avoiding an allocated copy for the string with newline, and all the reflection implied by `Sprintln`)), or it isn’t so much, and those loops can be replaced by something like `strings.Join(image.cmdlist, "\n")`.

Assuming the performance of this not _that_ much of a concern, that can be deferred for some indeterminate future.)

sif/internal/sif_util.go
```diff
@@ -0,0 +1,238 @@
...
+		return errors.Wrap(err, "reading lines from SIF definition file object")
+	}
+
+	if len(image.cmdlist) == 0 && len(image.envlist) == 0 {
+		image.cmdlist = append(image.cmdlist, "bash")
+	} else {
+		if err = image.generateRunscript(); err != nil {
+			return errors.Wrap(err, "generating runscript")
+		}
+		image.cmdlist = []string{"/podman/runscript"}
```
mtrmac, line 157: `image.cmdlist` serves two different purposes — a list of bash commands (e.g. input to `generateRunscript`, and a single command to be by the runtime. Those should be two separate variables.

sif/internal/sif_util.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+
+	return nil
+}
+
+func (image SifImage) GetConfig(config *imgspecv1.Image) error {
```
mtrmac, line 163: The name of this function doesn’t match what it does. Maybe a `) Command() []string` that just returns the command to run?

sif/sif/sif.go
```diff
@@ -0,0 +1,297 @@
+// Copyright (c) 2018, Sylabs Inc. All rights reserved.
+// Copyright (c) 2017, SingularityWare, LLC. All rights reserved.
+// Copyright (c) 2017, Yannick Cote <yhcote@gmail.com> All rights reserved.
+// This software is licensed under a 3-clause BSD license. Please consult the
+// LICENSE file distributed with the sources of this project regarding your
+// rights to use or distribute this software.
+
+// +build linux
```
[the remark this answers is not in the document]
mtrmac: Just removing these build conditionals, the code seems to at least build fine on macOS.

Of course things like `unsquashfs` or `fakeroot` might not be available — but the error handling needs to be user-acceptable for a Linux only version as well.  At least from a very quick search, both seem to be packaged in Homebrew.

sif/internal/sif_util.go
mtrmac: So I’m not sure which “it” I was supposed to test, but looking at https://github.com/mtrmac/image/tree/sif-wip :
- A trailing `%`… section header does not cause an infinite loop, because `scanner.Scan()`, after EOF, returns an empty string in `scanner.Text()`; AFAICS that’s totally undocumented (and possibly even contrary to the strict text of the existing documentation). So it’s good that there isn’t an infinite loop but we shouldn’t rely on that, unless there’s some guarantee I’m missing.
- A state-machine approach like the above does work, at least enough to pass a trivial smoke test I added.

mtrmac:
@yhcote @tri-adam Looking at this PR’s
```go
	if string(fimg.Header.Version[:HdrVersionLen-1]) > HdrVersion {
 		return fmt.Errorf("invalid SIF file: Version %s want <= %s", fimg.Header.Version, HdrVersion)
 	}
…
	HdrVersion      = "02"        // SIF SPEC VERSION
```
(≤ , and 2)

vs. https://github.com/sylabs/sif/blob/e807689a1a94304812a6c20573d0d11484b4e2af/pkg/sif/load.go#L29-L31 and https://github.com/sylabs/sif/blob/e807689a1a94304812a6c20573d0d11484b4e2af/pkg/sif/sif.go#L108-L114 (strict equality, and 1).

What’s that about?

Obviously a file with version "1" passes both checks, so at least that’s good. Is that the only one that matters?

I see https://github.com/sylabs/sif/commit/f68560dba9be69367c9e226b94752d86e4f4c4b4 exists in various repos, and then there’s https://github.com/sylabs/sif/commit/74413918f0817350ccd5879b49a886fd2c46aa82 .

(Speaking purely for myself in personal capacity, I don’t really want to dig up historical controversies, if any, and especially I’d prefer if c/image didn’t end up having to play arbiter between competing variants of a format, or worse, having to maintain its own private copy that has to do non-obvious workarounds to accept all of those competing variants at the same time. I’m really hoping for something like “$answer is the obviously correct one, some of the code is just behind”…)

dtrudg:
Hi @mtrmac @yhcote (@tri-adam)

The HdrVersion '01' is the one that really makes sense here. Regrettably during the early 3.x releases of Singularity there were some changes like this one (the reversion over the increment) that are not great, in hindsight...

The SIF v1.0.3 tag that contains the reversion from 02->01 in https://github.com/sylabs/sif/commit/74413918f0817350ccd5879b49a886fd2c46aa82 was first included in Singularity 3.2.0. Versions since then, including current versions of Singularity, are creating SIFs with a header version of `01`.

```
04:47 PM $ singularity sif header test.sif 
Launch Script:        #!/usr/bin/env run-singularity
Version:              01
```

Essentially all SIF containers you would find in the wild will say `01`.

dtrudg:
Further to my previous comment....

Singularity v3.1.1 used SIF v1.0.2, in which HdrVersion '01' was in effect (before the increment).
https://github.com/sylabs/singularity/blob/a83e020c14d6ee6af84831ff0c68acb2f176f0d4/Gopkg.toml#L51

Singularity v3.2.0 used SIF v1.0.3, in which HdrVersion '01' was in effect (after the reversion).
https://github.com/sylabs/singularity/blob/8ed39ade65934bf8cc202c7c5d3bf6ac2ae17c9d/go.mod#L75

So, we had no official tagged release that should have given a HdrVersion '02' SIF image. They should only appear if someone had built from development code pulling in SIF at that point. The code pulled into containers/image in this PR wasn't used in a release of Singularity.

mtrmac:
> So, we had no official tagged release that should have given a HdrVersion '02' SIF image.

Thanks, that’s very helpful.

mtrmac:
> `github.com/hpcng/sif/v2` is not actively maintained, other than occasionally releasing code from `github.com/sylabs/sif/v2` (ex. https://github.com/hpcng/sif/releases/tag/v2.0.0).

For the record, looking at the v1.7.0, v2.0.0, and latest common that seems true enough — there are _some_ differences between the corresponding tags, but almost all of the new code seems cherry-picked from the Sylabs repo, and the differences I 
> 
> `github.com/sylabs/sif/v2` contains both features and bug fixes not yet found in the other module. For that reason, I'd expect `github.com/sylabs/sif/v2` to be the natural choice if there's interest in code re-use.
> 
> > Since the current feature requires simple data object dumps from the SIF file the current import is sufficient. The best course of action right now to get the ball rolling is to at least get this initial PR in and let the community decide which improvements (some already listed in the PR comment) they want to see next.
> 
> Unless I'm missing something, using `github.com/sylabs/sif/v2` rather than duplicating the code in `containers/image/sif/sif` would be straight forward, and mean less code to maintain in `containers/image`. I've taken a stab at that here, as a practical example: [tri-adam@1dd4610](https://github.com/tri-adam/image/commit/1dd4610f563f37a4b8461e9f2f6e23f51698cdd6).
> 
> Anyways, just want to make it crystal clear that we're open to collaboration on this. Excited to see where it goes, in any case!

mtrmac:
> `github.com/hpcng/sif/v2` is not actively maintained, other than occasionally releasing code from `github.com/sylabs/sif/v2` (ex. https://github.com/hpcng/sif/releases/tag/v2.0.0).

For the record, looking at the v1.7.0, v2.0.0, and latest common v2.2.3 tags, that seems true enough — there are _some_ differences between the corresponding tags, but mostly just a timing difference (a commit missing from one version but added later, or the like), and almost all of the new code seems cherry-picked from the Sylabs repo.

mtrmac:
> I've taken a stab at that here, as a practical example: [tri-adam@1dd4610](https://github.com/tri-adam/image/commit/1dd4610f563f37a4b8461e9f2f6e23f51698cdd6).

Comparing with the current PR, I think `LoadContainerFromPath` should come with `OptLoadWithFlag(os.O_RDONLY)`. Otherwise the two look pretty similar — the only not-quite-trivial behavior difference I saw is things like `string(fimg.Header.Version[:HdrVersionLen-1])` vs. comparing the complete array against a value that includes a trailing 0; an implementation difference is that it the Sylabs version doesn’t try to use `mmap`.

Using the external library would also not drag in any extra dependencies.

sif/sif/sif.go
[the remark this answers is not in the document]
mtrmac: `fakeroot` does exist in homebrew, but it installs with
> Warning: fakeroot has been deprecated because it does not build!

and it also doesn’t work — in wildly unpredictable ways (sometimes files are not created, sometimes chown fails with EPERM, sometimes chown doesn’t fail but doesn’t do anything).

So making this Linux-only seems reasonable for now. (Ideally we shouldn’t need `fakeroot`, and do the conversion from SquashFS to tar most in-memory.)

mtrmac:
> Unless I'm missing something, using `github.com/sylabs/sif/v2` rather than duplicating the code in `containers/image/sif/sif` would be straight forward, and mean less code to maintain in `containers/image`. I've taken a stab at that here, as a practical example: [tri-adam@1dd4610](https://github.com/tri-adam/image/commit/1dd4610f563f37a4b8461e9f2f6e23f51698cdd6).

@tri-adam Could you sign this commit (`git commit --amend -s`) per https://github.com/containers/image/blob/main/CONTRIBUTING.md#sign-your-prs , and submit that branch as a PR, please? Then we can integrate it, along with other finishing touches I’m working on.

sif/sif_src.go
```diff
@@ -0,0 +1,287 @@
...
+
+	// populate the rootfs section of the config
+	rootfs := imgspecv1.RootFS{
+		Type:    "layers",
+		DiffIDs: []digest.Digest{imgSrc.diffID},
+	}
+	created := imgSrc.blobTime
+	history := []imgspecv1.History{
+		{
+			Created:   &created,
```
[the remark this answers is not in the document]
mtrmac: Using `.ModifiedAt()` from the header as a starting point.

sif/internal/sif_util.go
mtrmac: > Do we even need a copy? `unsquashfs` has an `-offset` option

This does work, but sadly isn’t available in RHEL 8.

mtrmac:
To test this, e.g.
```shell
# This is probably not the right way to download the image, it’s just a guess that seems to work
curl -L 'https://library.sylabs.io/v1/imagefile/library/default/lolcow:latest?arch=amd64' -o lolcow.sif
bin/skopeo copy --debug sif:lolcow.sif dir:t
podman run --privileged dir:t
```

or, of course, directly `podman pull sif:…` / `podman run sif:…`.

tri-adam:
> @tri-adam Could you sign this commit (`git commit --amend -s`) per https://github.com/containers/image/blob/main/CONTRIBUTING.md#sign-your-prs , and submit that branch as a PR, please? Then we can integrate it, along with other finishing touches I’m working on.

@mtrmac done, please see #1436. I haven't re-based or otherwise modified the original branch, but happy to do so if it'd help. Cheers.

mtrmac:
I have taken over this effort in #1438 . Thank you very much, @yhcote and @tri-adam !

PR #1403, opened 2021-11-09
Fix c/image fails to pull OCI image with non-`http(s)://` urls

ktock:
Fixes https://github.com/containers/podman/issues/12231

Currently, c/image cannot pull OCI images that contain non-`http(s)://` urls.
This commit fixes this issue by allowing fallback when pull fails on non-`http(s)://` urls.

### DIFF
--- a/docker/docker_image_src.go
+++ b/docker/docker_image_src.go
@@ -236,6 +236,9 @@ func (s *dockerImageSource) ensureManifestIsLoaded(ctx context.Context) error {
 	return nil
 }
 
+// getExternalBlob returns the reader of the first available blob URL from urls, which must not be empty.
+// This function can return nil reader when no url is supported by this function. In this case, the caller
+// should fallback to fetch the non-external blob (i.e. pull from the registry).
 func (s *dockerImageSource) getExternalBlob(ctx context.Context, urls []string) (io.ReadCloser, int64, error) {
 	var (
 		resp *http.Response
@@ -244,20 +247,26 @@ func (s *dockerImageSource) getExternalBlob(ctx context.Context, urls []string) 
 	if len(urls) == 0 {
 		return nil, 0, errors.New("internal error: getExternalBlob called with no URLs")
 	}
-	for _, url := range urls {
+	for _, u := range urls {
+		if u, err := url.Parse(u); err != nil || (u.Scheme != "http" && u.Scheme != "https") {
+			continue // unsupported url. skip this url.
+		}
 		// NOTE: we must not authenticate on additional URLs as those
 		//       can be abused to leak credentials or tokens.  Please
 		//       refer to CVE-2020-15157 for more information.
-		resp, err = s.c.makeRequestToResolvedURL(ctx, http.MethodGet, url, nil, nil, -1, noAuth, nil)
+		resp, err = s.c.makeRequestToResolvedURL(ctx, http.MethodGet, u, nil, nil, -1, noAuth, nil)
 		if err == nil {
 			if resp.StatusCode != http.StatusOK {
-				err = errors.Errorf("error fetching external blob from %q: %d (%s)", url, resp.StatusCode, http.StatusText(resp.StatusCode))
+				err = errors.Errorf("error fetching external blob from %q: %d (%s)", u, resp.StatusCode, http.StatusText(resp.StatusCode))
 				logrus.Debug(err)
 				resp.Body.Close()
 				continue
 			}
 			break
 		}
+	}
+	if resp == nil && err == nil {
+		return nil, 0, nil // fallback to non-external blob
 	}
 	if err != nil {
 		return nil, 0, err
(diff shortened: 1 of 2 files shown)


flouthoc:
LGTM. I think a pull test libimage would be nice but afaik `ghcr.io` will rate limit CI pulls.

@vrothberg @giuseppe @rhatdan PTAL

oci/layout/oci_src.go
```diff
@@ -132,6 +137,16 @@ func (s *ociImageSource) GetBlob(ctx context.Context, info types.BlobInfo, cache
 	return r, fi.Size(), nil
 }
 
+func containsValidHTTPURLs(urls []string) bool {
+	for _, s := range urls {
+		u, err := url.Parse(s)
+		if err == nil && (u.Scheme == "http" || u.Scheme == "https") {
```
vrothberg, line 143: I'd prefer to have `getExternalBlob` only attempt to fetch the URL if it's a supported scheme.  Could we instead extend `getExternalBlob` to do that?  Same for the other case.

ktock: Thank you for the review. Fixed `getExternalBlob`.

docker/docker_image_src.go
```diff
@@ -236,13 +236,23 @@ func (s *dockerImageSource) ensureManifestIsLoaded(ctx context.Context) error {
...
+	urls = supportedHTTPURLs(urls)
 	var (
 		resp *http.Response
 		err  error
 	)
 	if len(urls) == 0 {
-		return nil, 0, errors.New("internal error: getExternalBlob called with no URLs")
+		return nil, 0, true, errors.New("internal error: getExternalBlob called with no URLs")
 	}
 	for _, url := range urls {
```
vrothberg, line 257: I think the url check should be done here.  This way, we iterate over the `urls` slice once but not twice.

ktock: Thank you for the comment. Fixed.

docker/docker_image_src.go
```diff
@@ -236,33 +236,41 @@ func (s *dockerImageSource) ensureManifestIsLoaded(ctx context.Context) error {
...
 
-func (s *dockerImageSource) getExternalBlob(ctx context.Context, urls []string) (io.ReadCloser, int64, error) {
+func (s *dockerImageSource) getExternalBlob(ctx context.Context, urls []string) (_ io.ReadCloser, _ int64, fallback bool, _ error) {
 	var (
 		resp *http.Response
 		err  error
 	)
 	if len(urls) == 0 {
-		return nil, 0, errors.New("internal error: getExternalBlob called with no URLs")
+		return nil, 0, true, errors.New("internal error: getExternalBlob called with no URLs")
```
vrothberg, line 245: Shouldn't this be `false` since it's a fatal error?

vrothberg: Same for the other case.  It would be helpful to add a comment above the function to describe what the (boolean) return values mean.

vrothberg (commented):
LGTM

@mtrmac PTAL

mtrmac (commented):
ACK overall.

(Reading the OCI _spec_, it might be possible to just always fall back to querying the registry. OTOH the `docker/docker` implementation fails the same way, and we can make the IPFS URLs possible without changing this, so let’s do the conservative thing now and not change that for HTTP.)

docker/docker_image_src.go
```diff
@@ -236,33 +236,46 @@ func (s *dockerImageSource) ensureManifestIsLoaded(ctx context.Context) error {
...
+// This parameter becomes true only if no url is supported by this function.
+// Do not pass zero-length "urls". This is considered a fatal error and returns an error without allowing fallback.
+func (s *dockerImageSource) getExternalBlob(ctx context.Context, urls []string) (_ io.ReadCloser, _ int64, fallback bool, _ error) {
 	var (
 		resp *http.Response
 		err  error
 	)
 	if len(urls) == 0 {
-		return nil, 0, errors.New("internal error: getExternalBlob called with no URLs")
+		return nil, 0, false, errors.New("internal error: getExternalBlob called with no URLs") // fatal error
```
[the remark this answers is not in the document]
ktock: Thank you for the review. Fixed.

docker/docker_image_src.go
```diff
@@ -236,33 +236,46 @@ func (s *dockerImageSource) ensureManifestIsLoaded(ctx context.Context) error {
 	return nil
 }
 
-func (s *dockerImageSource) getExternalBlob(ctx context.Context, urls []string) (io.ReadCloser, int64, error) {
+// getExternalBlob returns the reader of the first available blob URL.
+// This returns a bool value (named "fallback") which indicates whether the caller can fallback to the pull of
+// the non-external blob (i.e. pull from the registry).
+// This parameter becomes true only if no url is supported by this function.
+// Do not pass zero-length "urls". This is considered a fatal error and returns an error without allowing fallback.
```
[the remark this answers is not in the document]
ktock: Fixed the comment.

docker/docker_image_src.go
```diff
@@ -236,33 +236,46 @@ func (s *dockerImageSource) ensureManifestIsLoaded(ctx context.Context) error {
...
 	var (
 		resp *http.Response
 		err  error
 	)
 	if len(urls) == 0 {
-		return nil, 0, errors.New("internal error: getExternalBlob called with no URLs")
+		return nil, 0, false, errors.New("internal error: getExternalBlob called with no URLs") // fatal error
 	}
-	for _, url := range urls {
+	noSupportedURL := true
```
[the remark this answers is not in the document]
ktock: Fixed.

docker/docker_image_src.go
```diff
@@ -236,33 +236,46 @@ func (s *dockerImageSource) ensureManifestIsLoaded(ctx context.Context) error {
...
 				logrus.Debug(err)
 				resp.Body.Close()
 				continue
 			}
 			break
 		}
 	}
+	if noSupportedURL {
+		return nil, 0, true, errors.New("no supported url is specified") // fallback to non-external blob
+	}
```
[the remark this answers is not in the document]
ktock: Thank you for the suggestion. Fixed the logic.

ktock:
@mtrmac Thank you for the review. Fixed.

ktock:
All green. Thank you for the review!

Could we move this forward?

vrothberg (commented):
LGTM

rhatdan:
LGTM

PR #1409, opened 2021-11-17
Reject ambiguous manifest formats

mtrmac:
Refuse to process manifest / manifest list data that could possibly be interpreted as two different manifest formats, because differences in how those ambiguities are resolved could be used to bypass image verification or review mechanisms.

Fixes CVE-2021-41190 / GHSA-77vh-xpmg-72qh .

### DIFF
--- a/manifest/common.go
+++ b/manifest/common.go
@@ -1,6 +1,7 @@
 package manifest
 
 import (
+	"encoding/json"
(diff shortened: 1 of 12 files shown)


mtrmac:
Previously reviewed by @vrothberg ; extra eyes would be very welcome.

Issue #1407, opened 2021-11-16
Separate mirror list for digest only and tag allowed image pull


QiWang19:
`mirror-by-digest-only` in registry.conf is configured for the original registry.
It is expected to configure this feature for mirrors, keep two lists of mirrors for the original registry: one list of mirrors is digest only, the other list of mirrors is tag allowed.
This feature is requested by ImageContentSourcePolicy CRD [enhancement](https://github.com/openshift/enhancements/pull/929/files#diff-eebeb781f4d3603039f4db588d8bd2ad7f35467593652f558463bf1c10312a3aR260) supporting the allow mirror by tags. This underlying change in registry.conf can simplify the syntax checking and validation of API implementation.

Enhancement PR: https://github.com/openshift/enhancements/pull/929

vrothberg:
Here's the specific location in the enhancement: https://github.com/openshift/enhancements/pull/929/files#diff-eebeb781f4d3603039f4db588d8bd2ad7f35467593652f558463bf1c10312a3aR260

vrothberg:
Before opening a PR, let's have a conversation on how to achieve that.  As @QiWang19 mentions above, `mirror-by-digest-only` is an attribute of the main registry. We must not change that for backwards compat.

I think that we may need two attributes for mirrors: `mirror-by-digest-only`  and `mirror-by-tag-only`.  They cannot both be set to `true`.  If non is set, the mirror will be used for tags and for digests.

mtrmac:
No, the goal is not to have two _booleans_, but two _lists_. I.e. turn `mirrors` into `digest-mirrors` and `tag-mirrors`, or maybe more likely(?) `digest-mirrors` and `mirrors` (I need to re-read the enhancement and earlier discussions carefully.) 

I.e. allow defining digest-only mirrors, and tag+digest mirrors, both for the same scope (where the digest-only mirrors can be on quite untrusted infrastructure, but the tag mirrors need to be more trusted for unsigned images, and possibly carefully synchronized if mutable tags like `:latest` are used in deployment configurations). The `mirror-by-digest-only` boolean would remain but probably effectively become deprecated, in favor of using the `digest-mirrors` list.

(Naming is hard. `digest-reference-mirrors`? More accurate, but probably too long to type.)

vrothberg:
[...]

PR #1411, opened 2021-11-19
Add pull-from-mirror for adding per-mirror level restrictions

QiWang19:
Close: https://github.com/containers/image/issues/1407
Add pull-from-mirror: all, digest-only, tag-only for adding per-mirror level restrictions

The `mirror-by-digest-only` for primary is still allowed configuring,
and it is honored for compatibility

### DIFF
--- a/docs/containers-registries.conf.5.md
+++ b/docs/containers-registries.conf.5.md
@@ -106,20 +106,26 @@ contacted and contains the image will be used (and if none of the mirrors contai
 the primary location specified by the `registry.location` field, or using the unmodified
 user-specified reference, is tried last).
 
-Each TOML table in the `mirror` array can contain the following fields, with the same semantics
-as if specified in the `[[registry]]` TOML table directly:
-- `location`
-- `insecure`
+Each TOML table in the `mirror` array can contain the following fields:
+- `location`： same semantics
+as specified in the `[[registry]]` TOML table
+- `insecure`： same semantics
+as specified in the `[[registry]]` TOML table
+- `pull-from-mirror`: `all`, `digest-only` or `tag-only`.  If "digest-only"， mirrors will only be used for digest pulls. Pulling images by tag can potentially yield different images, depending on which endpoint we pull from.  Restricting mirrors to pulls by digest avoids that issue.  If "tag-only", mirrors will only be used for tag pulls.  For a more up-to-date and expensive mirror that it is less likely to be out of sync if tags move, it should not be unnecessarily used for digest references.  Default is "all" (or left empty), mirrors will be used for both digest pulls and tag pulls unless the mirror-by-digest-only is set for the primary registry.
+Note that this per-mirror setting is allowed only when `mirror-by-digest-only` is not configured for the primary registry.
 
 `mirror-by-digest-only`
 : `true` or `false`.
 If `true`, mirrors will only be used during pulling if the image reference includes a digest.
+Note that if all mirrors are configured to be digest-only, images referenced by a tag will only use the primary
+registry.
+If all mirrors are configured to be tag-only, images referenced by a digest will only use the primary
+registry.
+
 Referencing an image by digest ensures that the same is always used
 (whereas referencing an image by a tag may cause different registries to return
 different images if the tag mapping is out of sync).
 
-Note that if this is `true`, images referenced by a tag will only use the primary
-registry, failing if that registry is not accessible.
 
 *Note*: Redirection and mirrors are currently processed only when reading images, not when pushing
 to a registry; that may change in the future.
--- a/pkg/sysregistriesv2/system_registries_v2.go
+++ b/pkg/sysregistriesv2/system_registries_v2.go
@@ -43,6 +43,16 @@ const builtinRegistriesConfDirPath = "/etc/containers/registries.conf.d"
 // helper.
 const AuthenticationFileHelper = "containers-auth.json"
 
+const (
+	// configuration values for "pull-from-mirror"
+	// mirrors will be used for both digest pulls and tag pulls
+	MirrorAll = "all"
+	// mirrors will only be used for digest pulls
+	MirrorByDigestOnly = "digest-only"
+	// mirrors will only be used for tag pulls
+	MirrorByTagOnly = "tag-only"
+)
+
 // Endpoint describes a remote location of a registry.
 type Endpoint struct {
 	// The endpoint's remote location. Can be empty iff Prefix contains
@@ -53,6 +63,18 @@ type Endpoint struct {
 	// If true, certs verification will be skipped and HTTP (non-TLS)
 	// connections will be allowed.
 	Insecure bool `toml:"insecure,omitempty"`
+	// PullFromMirror is used for adding restrictions to image pull through the mirror.
+	// Set to "all", "digest-only", or "tag-only".
+	// If "digest-only"， mirrors will only be used for digest pulls. Pulling images by
+	// tag can potentially yield different images, depending on which endpoint
+	// we pull from.  Restricting mirrors to pulls by digest avoids that issue.
+	// If "tag-only", mirrors will only be used for tag pulls.  For a more up-to-date and expensive mirror
+	// that it is less likely to be out of sync if tags move, it should not be unnecessarily
+	// used for digest references.
+	// Default is "all" (or left empty), mirrors will be used for both digest pulls and tag pulls unless the mirror-by-digest-only is set for the primary registry.
+	// This can only be set in a registry's Mirror field, not in the registry's primary Endpoint.
+	// This per-mirror setting is allowed only when mirror-by-digest-only is not configured for the primary registry.
+	PullFromMirror string `toml:"pull-from-mirror,omitempty"`
 }
 
 // userRegistriesFile is the path to the per user registry configuration file.
@@ -115,7 +137,7 @@ type Registry struct {
 	Blocked bool `toml:"blocked,omitempty"`
 	// If true, mirrors will only be used for digest pulls. Pulling images by
 	// tag can potentially yield different images, depending on which endpoint
-	// we pull from.  Forcing digest-pulls for mirrors avoids that issue.
+	// we pull from.  Restricting mirrors to pulls by digest avoids that issue.
 	MirrorByDigestOnly bool `toml:"mirror-by-digest-only,omitempty"`
 }
 
@@ -130,17 +152,29 @@ type PullSource struct {
 // reference.
 func (r *Registry) PullSourcesFromReference(ref reference.Named) ([]PullSource, error) {
 	var endpoints []Endpoint
-
+	_, isDigested := ref.(reference.Canonical)
 	if r.MirrorByDigestOnly {
-		// Only use mirrors when the reference is a digest one.
-		if _, isDigested := ref.(reference.Canonical); isDigested {
-			endpoints = append(r.Mirrors, r.Endpoint)
-		} else {
-			endpoints = []Endpoint{r.Endpoint}
+		// Only use mirrors when the reference is a digested one.
+		if isDigested {
+			endpoints = append(endpoints, r.Mirrors...)
 		}
 	} else {
-		endpoints = append(r.Mirrors, r.Endpoint)
+		for _, mirror := range r.Mirrors {
+			// skip the mirror if per mirror setting exists but reference does not match the restriction
+			switch mirror.PullFromMirror {
+			case MirrorByDigestOnly:
+				if !isDigested {
+					continue
+				}
+			case MirrorByTagOnly:
+				if isDigested {
+					continue
+				}
+			}
+			endpoints = append(endpoints, mirror)
+		}
 	}
+	endpoints = append(endpoints, r.Endpoint)
 
 	sources := []PullSource{}
 	for _, ep := range endpoints {
@@ -374,6 +408,10 @@ func (config *V2RegistriesConf) postProcessRegistries() error {
 			}
 		}
 
+		// validate the mirror usage settings does not apply to primary registry
+		if reg.PullFromMirror != "" {
+			return fmt.Errorf("pull-from-mirror must not be set for a non-mirror registry %q", reg.Prefix)
+		}
 		// make sure mirrors are valid
 		for _, mir := range reg.Mirrors {
 			mir.Location, err = parseLocation(mir.Location)
@@ -386,6 +424,14 @@ func (config *V2RegistriesConf) postProcessRegistries() error {
 			// https://github.com/containers/image/pull/1191#discussion_r610623216
 			if mir.Location == "" {
 				return &InvalidRegistries{s: "invalid condition: mirror location is unset"}
+			}
+
+			if reg.MirrorByDigestOnly && mir.PullFromMirror != "" {
+				return &InvalidRegistries{s: fmt.Sprintf("cannot set mirror usage mirror-by-digest-only for the registry (%q) and pull-from-mirror for per-mirror (%q) at the same time", reg.Prefix, mir.Location)}
+			}
+			if mir.PullFromMirror != "" && mir.PullFromMirror != MirrorAll &&
+				mir.PullFromMirror != MirrorByDigestOnly && mir.PullFromMirror != MirrorByTagOnly {
+				return &InvalidRegistries{s: fmt.Sprintf("unsupported pull-from-mirror value %q for mirror %q", mir.PullFromMirror, mir.Location)}
 			}
 		}
 		if reg.Location == "" {
--- a/pkg/sysregistriesv2/system_registries_v2_test.go
+++ b/pkg/sysregistriesv2/system_registries_v2_test.go
@@ -671,6 +671,122 @@ func TestPullSourcesFromReference(t *testing.T) {
 	assert.Equal(t, 1, len(pullSources))
 }
 
+func TestPullSourcesMirrorFromReference(t *testing.T) {
+	sys := &types.SystemContext{
+		SystemRegistriesConfPath:    "testdata/pull-sources-mirror-reference.conf",
+		SystemRegistriesConfDirPath: "testdata/this-does-not-exist",
+	}
+	registries, err := GetRegistries(sys)
+	require.NoError(t, err)
+	assert.Equal(t, 7, len(registries))
+
+	digest := "@sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+	tag := ":aaa"
+	for _, tc := range []struct {
+		registry      string
+		digestSources []string
+		tagSources    []string
+	}{
+		// Registry A has mirrors allow any kind of pull
+		{
+			"registry-a.com/foo/image",
+			[]string{"mirror-1.registry-a.com", "mirror-2.registry-a.com", "registry-a.com/bar"},
+			[]string{"mirror-1.registry-a.com", "mirror-2.registry-a.com", "registry-a.com/bar"},
+		},
+		// Registry B has mirrors allow digests pull only
+		{
+			"registry-b.com/foo/image",
+			[]string{"mirror-1.registry-b.com", "mirror-2.registry-b.com", "registry-b.com/bar"},
+			[]string{"registry-b.com/bar"},
+		},
+		// Registry C has a mirror allows digest pull only and a mirror allows any kind of pull
+		{
+			"registry-c.com/foo/image",
+			[]string{"mirror-1.registry-c.com", "mirror-2.registry-c.com", "registry-c.com/bar"},
+			[]string{"mirror-1.registry-c.com", "registry-c.com/bar"},
+		},
+		// Registry D set digest-only for registry level, allows only digest pulls
+		// Registry D has no digest-only set for mirrors table
+		{
+			"registry-d.com/foo/image",
+			[]string{"mirror-1.registry-d.com", "mirror-2.registry-d.com", "registry-d.com/bar"},
+			[]string{"registry-d.com/bar"},
+		},
+		// Registry E has mirrors only allows tag pull
+		{
+			"registry-e.com/foo/image",
+			[]string{"registry-e.com/bar"},
+			[]string{"mirror-1.registry-e.com", "mirror-2.registry-e.com", "registry-e.com/bar"},
+		},
+		// Registry F has one tag only mirror does not allow digest pull
+		{
+			"registry-f.com/foo/image",
+			[]string{"mirror-1.registry-f.com", "registry-f.com/bar"},
+			[]string{"mirror-1.registry-f.com", "mirror-2.registry-f.com", "registry-f.com/bar"},
+		},
+		// Registry G has one digest-only pull and one tag only pull
+		{
+			"registry-g.com/foo/image",
+			[]string{"mirror-1.registry-g.com", "mirror-3.registry-g.com", "mirror-4.registry-g.com", "registry-g.com/bar"},
+			[]string{"mirror-2.registry-g.com", "mirror-3.registry-g.com", "mirror-4.registry-g.com", "registry-g.com/bar"},
+		},
+	} {
+		// Digest
+		digestedRef := toNamedRef(t, tc.registry+digest)
+		registry, err := FindRegistry(sys, digestedRef.Name())
+		require.NoError(t, err)
+		require.NotNil(t, registry)
+		pullSource, err := registry.PullSourcesFromReference(digestedRef)
+		require.NoError(t, err)
+		for i, s := range tc.digestSources {
+			assert.Equal(t, s, pullSource[i].Endpoint.Location)
+		}
+		// Tag
+		taggedRef := toNamedRef(t, tc.registry+tag)
+		registry, err = FindRegistry(sys, taggedRef.Name())
+		require.NoError(t, err)
+		require.NotNil(t, registry)
+		pullSource, err = registry.PullSourcesFromReference(taggedRef)
+		require.NoError(t, err)
+		for i, s := range tc.tagSources {
+			assert.Equal(t, s, pullSource[i].Endpoint.Location)
+		}
+	}
+}
+
+func TestInvalidMirrorConfig(t *testing.T) {
+	for _, tc := range []struct {
+		sys       *types.SystemContext
+		expectErr string
+	}{
+		{
+			sys: &types.SystemContext{
+				SystemRegistriesConfPath:    "testdata/invalid-config-level-mirror.conf",
+				SystemRegistriesConfDirPath: "testdata/this-does-not-exist",
+			},
+			expectErr: fmt.Sprintf("pull-from-mirror must not be set for a non-mirror registry %q", "registry-a.com/foo"),
+		},
+		{
+			sys: &types.SystemContext{
+				SystemRegistriesConfPath:    "testdata/invalid-conflict-mirror.conf",
+				SystemRegistriesConfDirPath: "testdata/this-does-not-exist",
+			},
+			expectErr: fmt.Sprintf("cannot set mirror usage mirror-by-digest-only for the registry (%q) and pull-from-mirror for per-mirror (%q) at the same time", "registry-a.com/foo", "mirror-1.registry-a.com"),
+		},
+		{
+			sys: &types.SystemContext{
+				SystemRegistriesConfPath:    "testdata/invalid-value-mirror.conf",
+				SystemRegistriesConfDirPath: "testdata/this-does-not-exist",
+			},
+			expectErr: fmt.Sprintf("unsupported pull-from-mirror value %q for mirror %q", "notvalid", "mirror-1.registry-a.com"),
+		},
+	} {
+		_, err := GetRegistries(tc.sys)
+		assert.ErrorContains(t, err, tc.expectErr)
+	}
+
+}
+
 func TestTryUpdatingCache(t *testing.T) {
 	ctx := &types.SystemContext{
 		SystemRegistriesConfPath:    "testdata/try-update-cache-valid.conf",
--- a/pkg/sysregistriesv2/testdata/invalid-config-level-mirror.conf
+++ b/pkg/sysregistriesv2/testdata/invalid-config-level-mirror.conf
@@ -1,0 +1,11 @@
+[[registry]]
+prefix = "registry-a.com/foo"
+location = "registry-a.com/bar"
+pull-from-mirror = "digest-only"
+
+[[registry.mirror]]
+location = "mirror-1.registry-a.com"
+
+[[registry.mirror]]
+location = "mirror-2.registry-a.com"
+insecure = true
--- a/pkg/sysregistriesv2/testdata/invalid-conflict-mirror.conf
+++ b/pkg/sysregistriesv2/testdata/invalid-conflict-mirror.conf
@@ -1,0 +1,12 @@
+[[registry]]
+prefix = "registry-a.com/foo"
+location = "registry-a.com/bar"
+mirror-by-digest-only = true
+
+[[registry.mirror]]
+pull-from-mirror = "digest-only"
+location = "mirror-1.registry-a.com"
+
+[[registry.mirror]]
+location = "mirror-2.registry-a.com"
+insecure = true
--- a/pkg/sysregistriesv2/testdata/invalid-value-mirror.conf
+++ b/pkg/sysregistriesv2/testdata/invalid-value-mirror.conf
@@ -1,0 +1,11 @@
+[[registry]]
+prefix = "registry-a.com/foo"
+location = "registry-a.com/bar"
+
+[[registry.mirror]]
+pull-from-mirror = "notvalid"
+location = "mirror-1.registry-a.com"
+
+[[registry.mirror]]
+location = "mirror-2.registry-a.com"
+insecure = true
--- a/pkg/sysregistriesv2/testdata/pull-sources-mirror-reference.conf
+++ b/pkg/sysregistriesv2/testdata/pull-sources-mirror-reference.conf
@@ -1,0 +1,86 @@
+[[registry]]
+prefix = "registry-a.com/foo"
+location = "registry-a.com/bar"
+
+[[registry.mirror]]
+location = "mirror-1.registry-a.com"
+
+[[registry.mirror]]
+location = "mirror-2.registry-a.com"
+insecure = true
+
+[[registry]]
+prefix = "registry-b.com/foo"
+location = "registry-b.com/bar"
+
+[[registry.mirror]]
+pull-from-mirror = "digest-only"
+location = "mirror-1.registry-b.com"
+
+[[registry.mirror]]
+pull-from-mirror = "digest-only"
+location = "mirror-2.registry-b.com"
+
+[[registry]]
+prefix = "registry-c.com/foo"
+location = "registry-c.com/bar"
+
+[[registry.mirror]]
+location = "mirror-1.registry-c.com"
+
+[[registry.mirror]]
+pull-from-mirror = "digest-only"
+location = "mirror-2.registry-c.com"
+
+[[registry]]
+prefix = "registry-d.com/foo"
+location = "registry-d.com/bar"
+mirror-by-digest-only = true
+
+[[registry.mirror]]
+location = "mirror-1.registry-d.com"
+
+[[registry.mirror]]
+location = "mirror-2.registry-d.com"
+
+[[registry]]
+prefix = "registry-e.com/foo"
+location = "registry-e.com/bar"
+
+[[registry.mirror]]
+pull-from-mirror = "tag-only"
+location = "mirror-1.registry-e.com"
+
+[[registry.mirror]]
+pull-from-mirror = "tag-only"
+location = "mirror-2.registry-e.com"
+
+[[registry]]
+prefix = "registry-f.com/foo"
+location = "registry-f.com/bar"
+
+[[registry.mirror]]
+location = "mirror-1.registry-f.com"
+
+[[registry.mirror]]
+pull-from-mirror = "tag-only"
+location = "mirror-2.registry-f.com"
+
+[[registry]]
+prefix = "registry-g.com/foo"
+location = "registry-g.com/bar"
+
+[[registry.mirror]]
+pull-from-mirror = "digest-only"
+location = "mirror-1.registry-g.com"
+
+[[registry.mirror]]
+pull-from-mirror = "tag-only"
+location = "mirror-2.registry-g.com"
+
+[[registry.mirror]]
+location = "mirror-3.registry-g.com"
+
+[[registry.mirror]]
+pull-from-mirror = "all"
+location = "mirror-4.registry-g.com"


QiWang19:
@mtrmac @vrothberg Could you review?

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -110,7 +120,7 @@ type Registry struct {
 	// A registry is an Endpoint too
 	Endpoint
 	// The registry's mirrors.
-	Mirrors []Endpoint `toml:"mirror,omitempty"`
+	Mirrors []Mirror `toml:"mirror,omitempty"`
```
mtrmac, line 123: Hum, this is probably an API break. And putting `MirrorByDigestOnly` into `Endpoint` means there are two instances of that flag in a `Registry`.

This is going to be “fun”. (I didn’t spend time thinking how to deal with this. The nuclear option is to leave the public data types as is, and build a parallel implementation with private data types. Hopefully we can do something easier.)

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -130,17 +140,18 @@ type PullSource struct {
 // reference.
 func (r *Registry) PullSourcesFromReference(ref reference.Named) ([]PullSource, error) {
 	var endpoints []Endpoint
-
-	if r.MirrorByDigestOnly {
```
mtrmac, line 134: This still needs to honor the `Registry.MirroByDigestOnly` flag, for compatibility with older config files and older OpenShift.

mtrmac (commented):
(A quick first glance only.)

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: I don't think we need a new struct.  Moving `MirrorByDigestOnly bool` from `type Registry` to `type Endpoint` should do the trick.  This way, we continue supporting the top-level field of the registry and add support for mirrors as well.

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: I agree, see my comment in https://github.com/containers/image/pull/1411/files#r753042127

pkg/sysregistriesv2/system_registries_v2.go
mtrmac: Just moving it would break existing users like https://github.com/openshift/runtime-utils/blob/8b8348d80d1d1e7b6cf06fb009d5965e0b55baa2/pkg/registries/registries.go#L162 . Go API stability rules.

mtrmac: … Oh, I see what you mean; that one would work. But
```go
	v := Registry {
		MirrorByDigestOnly: true,
	}
```
```
./prog.go:50:3: cannot use promoted field Endpoint.MirrorByDigestOnly in struct literal of type Registry
```

vrothberg: I think that would still work since the type is embedded, isn't it?  Only initialization wouldn't work, e.g.:
```go
foo := Registry {
      MirrorByDigestOnly: true,
}
```

(Lawyer talking): the types are meant to be read, not written, which makes me OK to do this change.

mtrmac: Heh, I would argue the types are published only to be written (because an old writer can reasonably write a subset of the fields, if any new fields are added in the future), _not_ to be read (because an old reader will misinterpret configuration with unknown fields).

vrothberg: An alternative would be not use the very same name in the structs and have an `Endpoint.DigestOnly`.  That would practically then also apply to use the main registry only for digests which seems like a legit use case.

mtrmac: > An alternative would be not use the very same name in the structs and have an `Endpoint.DigestOnly`.

Yes.

> That would practically then also apply to use the main registry only for digests which seems like a legit use case.

That doesn’t make much sense to me. The primary registry can be unavailable, but to prohibit its use for _some_ accesses if it is available, and to prohibit especially those accesses for which that registry is the _only_ actual authority, is strange.

I’d prefer to reject such configurations, if only to avoid the confusion between the existing field (which affects all mirrors) and the new field (if it affects that single endpoint), at least until we have a clear use case.

vrothberg: Sounds good to me :heavy_check_mark:

docs/containers-registries.conf.5.md
```diff
@@ -108,16 +108,25 @@ Each TOML table in the `mirror` array can contain the following fields, with the
...
+- `digest-only`
 
 `mirror-by-digest-only`
 : `true` or `false`.
 If `true`, mirrors will only be used during pulling if the image reference includes a digest.
+Note that if this is `true`, images referenced by a tag will only use the primary
+registry, failing if that registry is not accessible.
+
+`digest-only`: `true` or `false`.
+: for each mirror in the `mirror` array, the mirror has `digest-only` set to `true` will only be used during pulling if the image reference includes a digest.
```
vrothberg, line 120:
```suggestion
: if `true` an individual mirror will only be used during pulling if the image reference includes a digest.
```

vrothberg: The text below is a bit hard to read.  I'd keep it short with: "Note that `digest-only` will be overridden if `mirror-by-digest` is set for the entire registry."

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -129,17 +135,31 @@ type PullSource struct {
 // PullSourcesFromReference returns a slice of PullSource's based on the passed
 // reference.
 func (r *Registry) PullSourcesFromReference(ref reference.Named) ([]PullSource, error) {
+	if r.DigestOnly {
```
vrothberg, line 138: Not sure if that's necessary here since it should be rejected when _loading_ the configuration files.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -129,17 +135,31 @@ type PullSource struct {
...
 		} else {
 			endpoints = []Endpoint{r.Endpoint}
 		}
 	} else {
-		endpoints = append(r.Mirrors, r.Endpoint)
+		// check digest-only for each mirror if non mirror-by-digest-only set, since mirror-by-digest-only is honored
+		for _, mirror := range r.Mirrors {
+			if mirror.DigestOnly {
+				if isDigested {
+					// Only use the mirror when the reference is a digest one.
```
vrothberg, line 155:
```suggestion
					// Only use the mirror when the reference is a digested one.
```

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -129,17 +135,31 @@ type PullSource struct {
...
 // reference.
 func (r *Registry) PullSourcesFromReference(ref reference.Named) ([]PullSource, error) {
+	if r.DigestOnly {
+		return nil, fmt.Errorf("digest-only does not apply to primary registry")
+	}
 	var endpoints []Endpoint
-
+	_, isDigested := ref.(reference.Canonical)
 	if r.MirrorByDigestOnly {
 		// Only use mirrors when the reference is a digest one.
```
vrothberg, line 144:
```suggestion
		// Only use mirrors when the reference is a digested one.
```

pkg/sysregistriesv2/testdata/pull-sources-from-reference.conf
```diff
@@ -12,10 +12,33 @@ insecure = true
 [[registry]]
 prefix = "registry-b.com/foo"
 location = "registry-b.com/bar"
-mirror-by-digest-only = true
```
vrothberg, line 15: I am a afraid of a potential (future) regression if alter this existing test.  We should leave this as is but probably add a new one to test for `mirror-by-digest-only = {true,false}` and mirrors set `digest-only={true,false}` (mixed)

pkg/sysregistriesv2/system_registries_v2_test.go
```diff
@@ -675,6 +675,84 @@ func TestPullSourcesFromReference(t *testing.T) {
...
+	referenceDDigest := toNamedRef(t, "registry-d.com/foo/image@sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa")
+	pullSources, err = registryD.PullSourcesFromReference(referenceDDigest)
+	assert.Nil(t, err)
+	assert.Equal(t, 3, len(pullSources))
+	assert.Equal(t, "mirror-1.registry-d.com/image@sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa", pullSources[0].Reference.String())
+	// Tag
+	referenceDTag := toNamedRef(t, "registry-d.com/foo/image:aaa")
+	pullSources, err = registryD.PullSourcesFromReference(referenceDTag)
+	assert.Nil(t, err)
+	assert.Equal(t, 1, len(pullSources))
```
vrothberg, line 753: Non-blocking nit: a table-driven would be easier to brain parse.

vrothberg (commented):
Other than the nit, LGTM

QiWang19:
@vrothberg  PTAL.

vrothberg (commented):
LGTM
@mtrmac PTAL

docs/containers-registries.conf.5.md
```diff
@@ -108,16 +108,24 @@ Each TOML table in the `mirror` array can contain the following fields, with the
...
 `mirror-by-digest-only`
 : `true` or `false`.
 If `true`, mirrors will only be used during pulling if the image reference includes a digest.
+Note that if this is `true`, images referenced by a tag will only use the primary
+registry, failing if that registry is not accessible.
+
+`digest-only`: `true` or `false`.
+: if `true` an individual mirror will only be used during pulling if the image reference includes a digest.
+Note that digest-only will be overridden if mirror-by-digest is set for the entire registry.
+Note that if all mirrors in the `mirror` array have `digest-only = true`, images referenced by a tag will only use the primary
```
TomSweeneyRedHat, line 122: A very soft suggestion if you need to clean anything else up.
```suggestion
Also note that if all mirrors in the `mirror` array have `digest-only = true`, images referenced by a tag will only use the primary
```

rhatdan:
@mtrmac PTAL

QiWang19:
@mtrmac Could you review?

QiWang19:
@mtrmac Could you review?

pkg/sysregistriesv2/system_registries_v2_test.go
```diff
@@ -675,6 +675,72 @@ func TestPullSourcesFromReference(t *testing.T) {
 	assert.Equal(t, 1, len(pullSources))
 }
 
+func TestPullSourcesMirrorFromReference(t *testing.T) {
```
mtrmac, line 678: Most importantly, thanks for the tests!

pkg/sysregistriesv2/system_registries_v2_test.go
```diff
@@ -675,6 +675,72 @@ func TestPullSourcesFromReference(t *testing.T) {
...
+	sys := &types.SystemContext{
```
mtrmac, line 679: To make it easier for the future, could `TestPullSourcesFromReference` be modified to use the same table-driven format, and then these tests added to that function, without adding a new config file?

It’s a bit out of scope (especially with the testing for `Location`/`Insecure`), so I’m perfectly fine with doing that work myself later, in a separate PR.

pkg/sysregistriesv2/system_registries_v2_test.go
```diff
@@ -675,6 +675,72 @@ func TestPullSourcesFromReference(t *testing.T) {
...
+	registries, err := GetRegistries(sys)
+	assert.Nil(t, err)
+	assert.Equal(t, 4, len(registries))
+
+	for _, tc := range []struct {
+		registry      string
+		digestRef     string
+		digestSources int
+		tagRef        string
+		tagSources    int
```
mtrmac, line 692: This could be simplified to just have `repo string`, `digestSources int`, `tagSources int`. Then…

pkg/sysregistriesv2/system_registries_v2_test.go
```diff
@@ -675,6 +675,72 @@ func TestPullSourcesFromReference(t *testing.T) {
...
+		// Digest
+		referenceDigest := toNamedRef(t, tc.digestRef)
+		pullSource, err := registry.PullSourcesFromReference(referenceDigest)
+		assert.Nil(t, err)
+		assert.Equal(t, tc.digestSources, len(pullSource))
+		// Tag
+		referenceTag := toNamedRef(t, tc.tagRef)
+		pullSource, err = registry.PullSourcesFromReference(referenceTag)
+		assert.Nil(t, err)
+		assert.Equal(t, tc.tagSources, len(pullSource))
```
mtrmac, line 740:
```go
digestedName := tc.registry + "@sha256:…"
registry, err := FindRegistry(sys, digestedRef)
…
digestedRef := toNamedRef(t, digestedName)
pullSources, err := …
```
and similarly test behavior of a tagged reference; no need to repeat the digest and tag values in the test case array, no need for a separate `tc.registry` with an ignored tag.

pkg/sysregistriesv2/system_registries_v2_test.go
mtrmac, line 692: `digestSources, tagSources []string`, listing the exact expected values, would be nice.

mtrmac (commented):
I apologize for the late response.

This does work fine, and it’s perfectly sufficient for correctness.

OTOH it doesn’t implement https://github.com/openshift/enhancements/pull/929 as currently proposed. That enhancement has a separate list of per-digest mirrors and per-tag mirrors, i.e. it is possible to have a tag-only mirror.

---

I can _imagine_,  with absolutely no data, that some users might want to use “cheap” mirrors for digested references, and only use “more expensive”, better-secured, mirrors for tagged references. But even in that hypothetical case, I find it fairly hard to argue that they’d rather have a failed pull than use the “more expensive” mirror.

So, for c/image, I think this could well be sufficient — and users might set up the configuration so that digest-only mirrors are listed before the digest+tag mirrors, if they wanted to avoid use of a “costly” digest+tag mirror.

But it’s pretty surprising to introduce the OpenShift feature, and this underlying design for an implementation of the OpenShift feature, at the same time, but not have them do _exactly_ the same thing. So either c/image should allow configuring tag-only mirrors, or the OpenShift CR should be changed so that the “tag” mirrors are documented to _also_ be used for digested references.

---

It’s somewhat likely that there was an earlier conversation about this that I can’t remember/find; is there?

Except I can find https://github.com/containers/image/issues/1407#issuecomment-971361471 , so it’s quite possible that the current inconsistent state of things is my fault. If that’s the case, I’m sorry.

pkg/sysregistriesv2/system_registries_v2_test.go
mtrmac: Actually, to be even closer to real-world callers,
```go
digestedRef := toNamedRef(t, tc.registry + "@sha256:…")
registry, err := FindRegistry(sys, digestedRef.Name())
```

QiWang19:
> So either c/image should allow configuring tag-only mirrors, or the OpenShift CR should be changed so that the “tag” mirrors are documented to _also_ be used for digested references.

Does the tag-only mirrors mean the registry reference without tag should be rejected in the case? The  openshift/enhancements#929 proposed to allow using tags, not to limit the use to tag-only.

pkg/sysregistriesv2/system_registries_v2_test.go
QiWang19: ok, I'd leave that to you.

mtrmac:
By tag-only mirrors I mean mirrors that aren’t used for digested references. Digested references would only use digest-only mirrors (if any), not the tag-only mirrors, and the primary (non-mirror) location.

The enhancement’s `ImageSourceTagPolicy` AFAICS only documents affecting tagged references; with this PR, that can’t be implemented as specified because the ISTP mirrors would be used for digest-only references as well.

Given ISDP (`source` -> `mirrorDigest`), ISTP (`source` -> `mirrorTag`), AFAICS OpenShift documents that `source:latest` should be looked up in (`mirrorTag:latest`, `source:latest`), and `source@sha256…` should be looked up in (`mirrorDigest@sha256…,` `source@sha256…`). With this PR, the former can’t be configured, the best we could do is (`mirrorTag:latest`, `mirrorDigest:latest`, `source:latest`).



I don’t have a very strong opinion on whether to change this implementation or the OpenShift CR.

QiWang19:
> By tag-only mirrors I mean mirrors that aren’t used for digested references. Digested references would only use digest-only mirrors (if any), not the tag-only mirrors, and the primary (non-mirror) location.
>
>With this PR, the former can’t be configured, the best we could do is (`mirrorTag:latest`, `mirrorDigest:latest`, `source:latest`).

The former one follows current implementation, like the [testcase](https://github.com/containers/image/pull/1411/files#diff-0c9be19c5fad22b2374b337e8dcf077e6ab87bddf59aa9b13dc018f36903a701R710)

I think you mean the digest example, `source@sha256… should be looked up in (mirrorDigest@sha256…, source@sha256…)` cannot be configured,
Current PR does (mirrorDigest@sha256, mirrorTag@sha256, source@sha256).

mtrmac:
> I think you mean the digest example

Yes, my mistake. Thanks.

QiWang19:
@mtrmac I fixed the reviews, PTAL.

docs/containers-registries.conf.5.md
```diff
@@ -108,16 +108,24 @@ Each TOML table in the `mirror` array can contain the following fields, with the
 as if specified in the `[[registry]]` TOML table directly:
 - `location`
 - `insecure`
+- `digest-only`
```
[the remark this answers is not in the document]
mtrmac: The new text is fine, but there’s another entry for `digest-only` that is now unnecessary or even misleading (it suggests that it can be set on `[[registry]]`.

docs/containers-registries.conf.5.md
```diff
@@ -108,16 +108,24 @@ Each TOML table in the `mirror` array can contain the following fields, with the
...
 : `true` or `false`.
 If `true`, mirrors will only be used during pulling if the image reference includes a digest.
+Note that if this is `true`, images referenced by a tag will only use the primary
+registry, failing if that registry is not accessible.
+
+`digest-only`: `true` or `false`.
+: if `true` an individual mirror will only be used during pulling if the image reference includes a digest.
+Note that digest-only will be overridden if mirror-by-digest is set for the entire registry.
+Also note that if all mirrors in the `mirror` array have `digest-only = true`, images referenced by a tag will only use the primary
+registry, failing if that registry is not accessible.
```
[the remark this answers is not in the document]
mtrmac: The `digest-only` copy will, I assume be removed — then the only copy of this note should be a separate paragraph, not a part of the `mirrors-by-digest-only` paragraph.

pkg/sysregistriesv2/system_registries_v2_test.go
```diff
@@ -675,6 +675,71 @@ func TestPullSourcesFromReference(t *testing.T) {
...
+		// Registry D set mirror-by-digest-only, honor this configuration for registry
+		// Regsitry D has no digest-only set for mirrors table
+		{
+			"registry-d.com/foo/image:latest",
+			[]string{"mirror-1.registry-d.com", "mirror-2.registry-d.com", "registry-d.com/bar"},
+			[]string{"registry-d.com/bar"},
+		},
+	} {
+		// Digest
+		digestedRef := toNamedRef(t, tc.registry+digest)
```
mtrmac, line 721: This uses values like `"registry-a.com/foo/image:latest@sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"`. Those are technically valid references per the c/image/docker/reference parser, but aren’t representative (c/image/docker rejects them).

…

pkg/sysregistriesv2/system_registries_v2_test.go
```diff
@@ -675,6 +675,71 @@ func TestPullSourcesFromReference(t *testing.T) {
...
+		registry, err := FindRegistry(sys, digestedRef.Name())
+		assert.Nil(t, err)
+		assert.NotNil(t, registry)
+		pullSource, err := registry.PullSourcesFromReference(digestedRef)
+		assert.Nil(t, err)
+		for i, s := range tc.digestSources {
+			assert.Equal(t, s, pullSource[i].Endpoint.Location)
+		}
+		// Tag
+		taggedRef := toNamedRef(t, tc.registry+tag)
```
mtrmac, line 731: … This uses values like `"registry-a.com/foo/image:latestaaa"`. That’s valid but pretty surprising.

Drop the `:latest` from `tc.registry`, use `tag := ":aaa"`, and this will be both simpler and more representative.

pkg/sysregistriesv2/system_registries_v2_test.go
```diff
@@ -675,6 +675,71 @@ func TestPullSourcesFromReference(t *testing.T) {
...
+		{
+			"registry-d.com/foo/image:latest",
+			[]string{"mirror-1.registry-d.com", "mirror-2.registry-d.com", "registry-d.com/bar"},
+			[]string{"registry-d.com/bar"},
+		},
+	} {
+		// Digest
+		digestedRef := toNamedRef(t, tc.registry+digest)
+		registry, err := FindRegistry(sys, digestedRef.Name())
+		assert.Nil(t, err)
```
mtrmac, line 723: Non-blocking, these tests do work well enough, and it can be improved if we ever run into a case where this matters:
- `NoError` instead of `Nil`, that produces better failure text
- For error values, usually `require.` instead of `assert.`, because the rest of the test case would almost always fail on a `nil` reference if the code just continued and tried using the other return values.

In all cases.

pkg/sysregistriesv2/system_registries_v2_test.go
```diff
@@ -675,6 +675,71 @@ func TestPullSourcesFromReference(t *testing.T) {
...
+		assert.NotNil(t, registry)
+		pullSource, err := registry.PullSourcesFromReference(digestedRef)
+		assert.Nil(t, err)
+		for i, s := range tc.digestSources {
+			assert.Equal(t, s, pullSource[i].Endpoint.Location)
```
mtrmac, line 728: Absolutely non-blocking: Testing `pullSource[i].Reference` would be more directly related to the end-user-wanted behavior.

But this works just fine, and I can clean that up later when unifying the two tests.

mtrmac (commented):
Thanks! The implementation LGTM, just some test and documentation nits.

---

I don’t think it would be reasonable to block _this PR_ on the inconsistency with OpenShift plans, per https://github.com/containers/image/pull/1411#pullrequestreview-853516067 . But I do hope there’s _some_ plan to address that.

I suppose another c/image PR that adds another per-mirror option can land later, or something.

QiWang19:
> inconsistency with OpenShift plans

Question about the inconsistency with openshift enhancement openshift/enhancements#929. The plan hasn't pointed out the `ImageTagMirrorSet` configured mirror is only for tag pull. The plan documented it is used to allow using tags that has the underlying meaning the digest pull is allowed by default. Are you suggesting the plan should mention the `ImageTagMirrorSet` mirrors will only be used for tag pulls and will not be used by digest pull?

mtrmac:
I understand
```go
	// repositoryTagMirrors allows images referenced by image tags in pods to be
 	// pulled from alternative mirrored repository locations. 
```
to mean that it doesn’t affect digested references at all.

Paying more attention, the distinctions like
```markdown
- ImageDigestMirrorSet: holds mirror configurations that are used to pull image from mirrors by digest specification only.
- ImageTagMirrorSet: holds mirror configurations that allow pulling image from mirrors using tag specification.
```
(with the “only” word only in the digest case) might be understood to mean that ITMS affects digested references as well, that was extremely easy to miss.

There’s also
```go
// Images can be pulled from these mirrors only if they are referenced by their digests.
```
even in the ITMS section, also suggesting exclusivity.

---

If the intent of the enhancement is to make ITMS affect both tags and digests, I think that’s a reasonable design but this design choice should be made more explicit in the proposal (and I’m not sure why naming it _Tag_Mirror is appropriate in that case).

mtrmac (commented):
Code LGTM. Thanks!

Leaving this open for a bit to let the design discussion settle.

QiWang19:
Thanks. I will start looking at having support for per-mirror option tag-only.
Following the idea form https://github.com/containers/image/issues/1407#issuecomment-971355574 Add a new field for each mirror:
```
[[registry]]
# mirror-by-digest-only = true digest only for all mirrors if set true
prefix = "registry-b.com/foo"
location = "registry-b.com/bar"

[[registry.mirror]]
digest-only = true # conflict if set both digest-only and tag-only
location = "mirror-1.registry-b.com"

[[registry.mirror]]
tag-only = true
location = "mirror-2.registry-b.com"

[[registry.mirror]]
location = "generic.mirror.com"
```

QiWang19:
@mtrmac PTAL. Added the tag-only mirrors by `mirror-by-tag-only` and `tag-only`.

QiWang19:
@mtrmac @vrothberg Could you review? The  `mirror-by-tag-only` and `tag-only` is added to support the mirrors for tag only to stay consistent with https://github.com/openshift/enhancements/pull/929

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -115,8 +127,12 @@ type Registry struct {
 	Blocked bool `toml:"blocked,omitempty"`
 	// If true, mirrors will only be used for digest pulls. Pulling images by
 	// tag can potentially yield different images, depending on which endpoint
-	// we pull from.  Forcing digest-pulls for mirrors avoids that issue.
+	// we pull from.  Restricting mirrors to pulls for mirrors avoids that issue.
 	MirrorByDigestOnly bool `toml:"mirror-by-digest-only,omitempty"`
+	// If true, mirrors will only be used for tag pulls. Pulling images by
+	// tag can potentially yield different images, depending on which endpoint
+	// we pull from.  Restricting mirrors to pulls for mirrors avoids that issue.
+	MirrorByTagOnly bool `toml:"mirror-by-tag-only,omitempty"`
```
mtrmac, line 135: … Do we need this setting at all? I don’t really see why one would ever enable this. AFAIK, tag-only mirrors only make sense if there are other digest-only mirrors.

@vrothberg WDYT?

docs/containers-registries.conf.5.md
```diff
@@ -104,20 +104,32 @@ contacted and contains the image will be used (and if none of the mirrors contai
...
-- `insecure`
+Each TOML table in the `mirror` array can contain the following fields:
+- `location`： same semantics
+as specified in the `[[registry]]` TOML table
+- `insecure`： same semantics
+as specified in the `[[registry]]` TOML table
+- `digest-only`： `true` or `false`. If `true` an individual mirror will only be used during pulling if the image reference includes a digest.
+- `tag-only`： `true` or `false`. If `true` an individual mirror will only be used during pulling if the image reference includes a tag.
+Note that digest-only will be overridden if mirror-by-digest-only is set for the entire registry, tag-only will not be overridden.
+tag-only will be overridden if mirror-by-tag-only is set for the entire registry, digest-only will not be overridden.
```
mtrmac, line 115: Quickly reading this, it’s unclear to me what is the resulting behavior of “will (not) be overridden”.

Especially the way how a mirror can “opt out” of the registry-level $something-only by setting $opposite-only is to me, completely non-obvious from this text.

I’m not even sure we need to define these things — alternatively we could forbid mirror-level *-only options if the registry-level *-only is set, and thus allow the administrator to either define registry-level, or mirror-level, settings, and avoid these conflicts. (@vrothberg ?)

docs/containers-registries.conf.5.md
```diff
@@ -104,20 +104,32 @@ contacted and contains the image will be used (and if none of the mirrors contai
...
-as if specified in the `[[registry]]` TOML table directly:
-- `location`
-- `insecure`
+Each TOML table in the `mirror` array can contain the following fields:
+- `location`： same semantics
+as specified in the `[[registry]]` TOML table
+- `insecure`： same semantics
+as specified in the `[[registry]]` TOML table
+- `digest-only`： `true` or `false`. If `true` an individual mirror will only be used during pulling if the image reference includes a digest.
+- `tag-only`： `true` or `false`. If `true` an individual mirror will only be used during pulling if the image reference includes a tag.
```
mtrmac, line 113: This probably should indicate that `false` is the default somehow.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -53,6 +53,18 @@ type Endpoint struct {
...
 	Insecure bool `toml:"insecure,omitempty"`
+	// If true, the mirror will only be used for digest pulls. Pulling images by
+	// tag can potentially yield different images, depending on which endpoint
+	// we pull from.  Restricting mirrors to pulls by digest avoids that issue.
+	// This can only be set in a registry's Mirror field, not in the registry's primary Endpoint.
+	// If mirror-by-digest-only is set to true for the primary registry, this per-mirror setting is ignored.
+	DigestOnly bool `toml:"digest-only,omitempty"`
+	// If true, the mirror will only be used for tag pulls. Pulling images by
+	// tag can potentially yield different images, depending on which endpoint
+	// we pull from.  Restricting mirrors to pulls by digest avoids that issue.
```
mtrmac, line 64: The `Pulling … issue` sentences are relevant as motivation for `DigestOnly`, and don’t seem that helpful for `TagOnly`.

I suppose that should be replaced by an explanation for why anyone would ever want `TagOnly` (= to have a more-up-to-date mirror that is less likely to be out of sync if tags move, but which is more expensive, so it should not be unnecessarily used for digest references).

… but it’s a bit unexpected that all this usage/rationale is in code comments, but not available in the end-user-facing `registries.conf.5.md`. If anything, I’d expect it the other way around.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -115,8 +127,12 @@ type Registry struct {
 	Blocked bool `toml:"blocked,omitempty"`
 	// If true, mirrors will only be used for digest pulls. Pulling images by
 	// tag can potentially yield different images, depending on which endpoint
-	// we pull from.  Forcing digest-pulls for mirrors avoids that issue.
+	// we pull from.  Restricting mirrors to pulls for mirrors avoids that issue.
```
mtrmac, line 130: I don’t understand what “Restricting mirrors to pulls for mirrors” means. The previous version was not quite correct either, to be fair — but it should be possible to do better than both.

(See also elsewhere about making the rationale / recommendations use-visible.)

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -115,8 +127,12 @@ type Registry struct {
...
 	MirrorByDigestOnly bool `toml:"mirror-by-digest-only,omitempty"`
+	// If true, mirrors will only be used for tag pulls. Pulling images by
+	// tag can potentially yield different images, depending on which endpoint
+	// we pull from.  Restricting mirrors to pulls for mirrors avoids that issue.
```
mtrmac, line 134: I don’t understand what “Restricting mirrors to pulls for mirrors” means — and this should, if anything, explain why one would use `MirrorByTagOnly`, not `MirrorByDigestOnly` (see elsewhere).

But…

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -387,6 +415,11 @@ func (config *V2RegistriesConf) postProcessRegistries() error {
 			if mir.Location == "" {
 				return &InvalidRegistries{s: "invalid condition: mirror location is unset"}
 			}
+
+			if mir.DigestOnly && mir.TagOnly {
+				msg := fmt.Sprintf("invalid condition: set digest-only and tag-only for the same mirror %s", mir.Location)
```
mtrmac, line 420: I’m not sure the `invalid condition: ` part is useful here. (I realize the error above starts the same way.)

Probably “digest-only and tag-only (is) set on the same mirror”, I tend to interpret “set …” as a command.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -130,17 +146,19 @@ type PullSource struct {
...
 	var endpoints []Endpoint
-
-	if r.MirrorByDigestOnly {
-		// Only use mirrors when the reference is a digest one.
-		if _, isDigested := ref.(reference.Canonical); isDigested {
-			endpoints = append(r.Mirrors, r.Endpoint)
+	_, isDigested := ref.(reference.Canonical)
+	for _, mirror := range r.Mirrors {
+		if isDigested {
+			if !mirror.TagOnly && !r.MirrorByTagOnly || mirror.DigestOnly {
```
mtrmac, line 152: If the model isn’t simplified, and these conditions remain (see elsewhere): Please use at least parentheses to be explicit about the operator priority / intent here.
Maybe even something like
```go
if mirror.DigestOnly // explicitly opted in
  || (!mirror.TagOnly && !r.MirrorByTagOnly) { // not opted out
```

Admittedly the `(!A && !B) || C` and `!A && (!B | C)` expressions differ only in the forbidden case of `mirror.TagOnly && mirror.DigestOnly`, but I ended up actually making a table of all 8 possible inputs to the two expressions and evaluating the two expressions to figure that out (admittedly it can also be done symbolically) — either way I think that’s a bit too much to ask of anyone modifying this code in the future, possibly in a hurry.

---

Same in the other condition.

pkg/sysregistriesv2/system_registries_v2.go
QiWang19: My mistake. It should be `Restricting mirrors to pulls by digest mirrors avoids that issue.`

pkg/sysregistriesv2/system_registries_v2.go
vrothberg: IIRC, I was asking to add it but I am not feeling strongly.  _But_ I am worried about the ambiguity of a mixed list of mirrors. Let's consider the following:
* Mirror A: digest-only=true
* Mirror B: digest-only=false
* Mirror C: digest-only=true

Now, I am pulling `foo@digest`.  Would we attempt a pull from `Mirror B`?  `digest-only=false` would intuitively imply that I can safely pull a digest from it but in the context of the other mirrors it does not really makes sense.

Maybe `mirror-by-tag-only` should only be set internally if a "sibling mirror" sets `digest-only=true`?


Making configuration settings user-friendly is hard.

docs/containers-registries.conf.5.md
vrothberg: > and thus allow the administrator to either define registry-level, or mirror-level, settings, and avoid these conflicts

That sounds like a clean model to me.

pkg/sysregistriesv2/system_registries_v2.go
mtrmac: My mind is sufficiently twisted by programming that “`digest-only = false` allows working with digests” feels obvious and intuitive to me, but I know how I am at UI…

---

How about reformulating this as:
- `pull-from-mirror=all[-references]` (the default)
- `pull-from-mirror=digest-only`
- `pull-from-mirror=tag-only`

That’s the same set of choices, but 1) it is impossible to set digest-only and tag-only at the same code, avoiding a corner case, and 2) it might make it clearer what the default is (and 3) it allows us to define a different set of mirrors for pushing, if we ever added that feature).

vrothberg: > How about reformulating this as:

That is by far the cleanest version IMHO. @QiWang19, what do you think?

QiWang19:
> That is by far the cleanest version IMHO. @QiWang19, what do you think?

@vrothberg @mtrmac  Reformulating it as `pull-from-mirror` sounds good.  I have updated this PR could you review it?
I am not sure should we still keep the logic for `mirror-by-digest-only` or only use the  `pull-from-mirror`?

QiWang19:
@mtrmac @vrothberg PTAL

docs/containers-registries.conf.5.md
```diff
@@ -104,20 +104,25 @@ contacted and contains the image will be used (and if none of the mirrors contai
...
-as if specified in the `[[registry]]` TOML table directly:
-- `location`
-- `insecure`
-
+Each TOML table in the `mirror` array can contain the following fields:
+- `location`： same semantics
+as specified in the `[[registry]]` TOML table
+- `insecure`： same semantics
+as specified in the `[[registry]]` TOML table
+- `pull-from-mirror`: `all`, `digest-only` or `tag-only`.  If "digest-only"， mirrors will only be used for digest pulls. Pulling images by tag can potentially yield different images, depending on which endpoint we pull from.  Restricting mirrors to pulls by digest avoids that issue.  If "tag-only", mirrors will only be used for tag pulls.  For a more up-to-date and expensive mirror that it is less likely to be out of sync if tags move, it should not be unnecessarily used for digest references.  Default is "all", mirror will be used for both digest pull and tag pull unless the mirror-by-digest-only is set for primary registry .
```
vrothberg, line 112:
```suggestion
- `pull-from-mirror`: `all`, `digest-only` or `tag-only`.  If "digest-only"， mirrors will only be used for digest pulls. Pulling images by tag can potentially yield different images, depending on which endpoint we pull from.  Restricting mirrors to pulls by digest avoids that issue.  If "tag-only", mirrors will only be used for tag pulls.  For a more up-to-date and expensive mirror that it is less likely to be out of sync if tags move, it should not be unnecessarily used for digest references.  Default is "all", mirrors will be used for both digest pulls and tag pulls unless the mirror-by-digest-only is set for the primary registry . 
```

docs/containers-registries.conf.5.md
```diff
@@ -104,20 +104,25 @@ contacted and contains the image will be used (and if none of the mirrors contai
...
+- `pull-from-mirror`: `all`, `digest-only` or `tag-only`.  If "digest-only"， mirrors will only be used for digest pulls. Pulling images by tag can potentially yield different images, depending on which endpoint we pull from.  Restricting mirrors to pulls by digest avoids that issue.  If "tag-only", mirrors will only be used for tag pulls.  For a more up-to-date and expensive mirror that it is less likely to be out of sync if tags move, it should not be unnecessarily used for digest references.  Default is "all", mirror will be used for both digest pull and tag pull unless the mirror-by-digest-only is set for primary registry . 
+Note that pull-from-mirror can only be set in a registry's Mirror field, not in the registry's primary Endpoint. This per-mirror setting is allowed only when mirror-by-digest-only is not configured for the primary registry.
 `mirror-by-digest-only`
 : `true` or `false`.
 If `true`, mirrors will only be used during pulling if the image reference includes a digest.
+Note that if all mirrors are configured to be digest-only, images referenced by a tag will only use the primary
+registry, failing if that registries are not accessible.
```
vrothberg, line 118:
```suggestion
registry.
```

docs/containers-registries.conf.5.md
```diff
@@ -104,20 +104,25 @@ contacted and contains the image will be used (and if none of the mirrors contai
...
+If all mirrors are configured to be tag-only, images referenced by a digest will only use the primary
+registry, failing if that registries are not accessible.
```
vrothberg, line 120:
```suggestion
registry.
```

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -43,6 +43,14 @@ const builtinRegistriesConfDirPath = "/etc/containers/registries.conf.d"
 // helper.
 const AuthenticationFileHelper = "containers-auth.json"
 
+type PullFromMirror string
+
+const (
+	MirrorAll          PullFromMirror = "all"
+	MirrorByDigestOnly PullFromMirror = "digest-only"
+	MirrorByTagOnly    PullFromMirror = "tag-only"
+)
```
vrothberg, line 52: Do we really need to export them?  I'd expect a simple string (no extra type).

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -130,17 +150,25 @@ type PullSource struct {
...
-			endpoints = []Endpoint{r.Endpoint}
+		// Only use mirrors when the reference is a digested one.
+		if isDigested {
+			endpoints = append(endpoints, r.Mirrors...)
 		}
 	} else {
-		endpoints = append(r.Mirrors, r.Endpoint)
+		for _, mirror := range r.Mirrors {
+			// skip the mirror if per mirror setting exists but reference does not match the restriction
+			if mirror.PullFromMirror != "" && mirror.PullFromMirror == MirrorByDigestOnly && !isDigested {
```
vrothberg, line 162:
```suggestion
			if mirror.PullFromMirror == MirrorByDigestOnly && !isDigested {
```

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -130,17 +150,25 @@ type PullSource struct {
...
+				continue
+			}
+			if mirror.PullFromMirror != "" && mirror.PullFromMirror == MirrorByTagOnly && isDigested {
```
vrothberg, line 165:
```suggestion
			if mirror.PullFromMirror == MirrorByTagOnly && isDigested {
```

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -387,6 +419,14 @@ func (config *V2RegistriesConf) postProcessRegistries() error {
 			if mir.Location == "" {
 				return &InvalidRegistries{s: "invalid condition: mirror location is unset"}
 			}
+
+			if reg.MirrorByDigestOnly && mir.PullFromMirror != "" {
+				return &InvalidRegistries{s: "mirror usage set for non-mirror registry and per-mirror at the same time."}
```
vrothberg, line 424: The error message should include the fields that conflict so that users know directly what to look for.

Maybe something like: `cannot set option A for the registry and option B for the mirror`

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -387,6 +419,14 @@ func (config *V2RegistriesConf) postProcessRegistries() error {
...
+			}
+			if mir.PullFromMirror != "" && mir.PullFromMirror != MirrorAll &&
+				mir.PullFromMirror != MirrorByDigestOnly && mir.PullFromMirror != MirrorByTagOnly {
+				return &InvalidRegistries{s: fmt.Sprintf("invalid pull-from-mirror %s is set on the mirror %s", mir.PullFromMirror, mir.Location)}
```
vrothberg, line 428:
```suggestion
				return &InvalidRegistries{s: fmt.Sprintf("unsupported pull-from-mirror value %q for mirror %q", mir.PullFromMirror, mir.Location)}
```

QiWang19:
@vrothberg @mtrmac PTAL.

QiWang19:
@mtrmac Could you review it?

pkg/sysregistriesv2/system_registries_v2.go
QiWang19: > Could you add comments to them? I am still not sure whether introducing a new type is needed though.

yes, the new type is not necessary. Removed the new type.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -387,6 +419,14 @@ func (config *V2RegistriesConf) postProcessRegistries() error {
...
 				return &InvalidRegistries{s: "invalid condition: mirror location is unset"}
 			}
+
+			if reg.MirrorByDigestOnly && mir.PullFromMirror != "" {
+				return &InvalidRegistries{s: "cannot set mirror usage mirror-by-digest-only for the registry and pull-from-mirror for per-mirror at the same time."}
+			}
+			if mir.PullFromMirror != "" && mir.PullFromMirror != MirrorAll &&
+				mir.PullFromMirror != MirrorByDigestOnly && mir.PullFromMirror != MirrorByTagOnly {
+				return &InvalidRegistries{s: fmt.Sprintf("unsupported pull-from-mirror value %q for mirror %q", mir.PullFromMirror, mir.Location)}
+			}
```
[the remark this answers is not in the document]
QiWang19: test added

QiWang19:
@vrothberg @mtrmac PTAL

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -130,17 +152,32 @@ type PullSource struct {
...
+				}
+				endpoints = append(endpoints, mirror)
+			case MirrorByTagOnly:
+				if isDigested {
+					continue
+				}
+				endpoints = append(endpoints, mirror)
+			default:
+				endpoints = append(endpoints, mirror)
+			}
```
vrothberg, line 177:
```suggestion
			switch mirror.PullFromMirror {
			case MirrorByDigestOnly:
				if !isDigested {
					continue
				}
			case MirrorByTagOnly:
				if isDigested {
					continue
				}
			}
			endpoints = append(endpoints, mirror)
```

vrothberg (commented):
Other than the nit, LGTM

pkg/sysregistriesv2/system_registries_v2.go
mtrmac: This is still outstanding.

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -387,6 +425,14 @@ func (config *V2RegistriesConf) postProcessRegistries() error {
 			if mir.Location == "" {
 				return &InvalidRegistries{s: "invalid condition: mirror location is unset"}
 			}
+
+			if reg.MirrorByDigestOnly && mir.PullFromMirror != "" {
+				return &InvalidRegistries{s: "cannot set mirror usage mirror-by-digest-only for the registry and pull-from-mirror for per-mirror at the same time"}
```
mtrmac, line 430: Non-blocking: It would be useful to identify the problematic registry/mirror (via `reg.Prefix`/`mir.Location`) in the error message.

pkg/sysregistriesv2/system_registries_v2_test.go
```diff
@@ -671,6 +671,122 @@ func TestPullSourcesFromReference(t *testing.T) {
...
+			"registry-g.com/foo/image",
+			[]string{"mirror-1.registry-g.com", "mirror-3.registry-g.com", "mirror-4.registry-g.com", "registry-g.com/bar"},
+			[]string{"mirror-2.registry-g.com", "mirror-3.registry-g.com", "mirror-4.registry-g.com", "registry-g.com/bar"},
+		},
+	} {
+		// Digest
+		digestedRef := toNamedRef(t, tc.registry+digest)
+		registry, err := FindRegistry(sys, digestedRef.Name())
+		require.NoError(t, err)
+		assert.NotNil(t, registry)
```
mtrmac, line 738: Non-blocking: `require.NotNil`, otherwise the next line would crash.

The test fails either way, so it can be fixed up if it ever happens.

Same in the `taggedRef` case below.

docs/containers-registries.conf.5.md
```diff
@@ -106,20 +106,26 @@ contacted and contains the image will be used (and if none of the mirrors contai
...
-as if specified in the `[[registry]]` TOML table directly:
-- `location`
-- `insecure`
+Each TOML table in the `mirror` array can contain the following fields:
+- `location`： same semantics
+as specified in the `[[registry]]` TOML table
+- `insecure`： same semantics
+as specified in the `[[registry]]` TOML table
+- `pull-from-mirror`: `all`, `digest-only` or `tag-only`.  If "digest-only"， mirrors will only be used for digest pulls. Pulling images by tag can potentially yield different images, depending on which endpoint we pull from.  Restricting mirrors to pulls by digest avoids that issue.  If "tag-only", mirrors will only be used for tag pulls.  For a more up-to-date and expensive mirror that it is less likely to be out of sync if tags move, it should not be unnecessarily used for digest references.  Default is "all" (or left empty), mirrors will be used for both digest pulls and tag pulls unless the mirror-by-digest-only is set for the primary registry.
+Note that pull-from-mirror can only be set in a registry's Mirror field, not in the registry's primary Endpoint. This per-mirror setting is allowed only when mirror-by-digest-only is not configured for the primary registry.
```
mtrmac, line 115: The readers of this document are not aware of the `Endpoint` type shared between registries and mirrors, that’s an invisible implementation detail to them; so, the sentence about the `Mirror` field can be just dropped.

mtrmac (commented):
Design and overall implementation LGTM.

QiWang19:
@vrothberg @mtrmac PTAL. This is ready to get merged.

mtrmac (approved):
LGTM. Thanks!

vrothberg (approved):
LGTM

QiWang19:
@mtrmac @vrothberg can we have a new release for this feature?

mtrmac:
Sure, a new release sounds fine. @vrothberg WDYT? Is there any reason not to, or anything urgent we should finish first?

vrothberg:
Releasing now SGTM. Thanks for working on this, @QiWang19!

mtrmac:
https://github.com/containers/image/pull/1514 filed to release v5.21.0.

PR #1413, opened 2021-11-21
Add option to preserve digests

Jamstah:
A digest-stable copy seems popular, even when not copying signed images.
Using --all can still change digests. Adding an option to ensure digests
are preserved.

Will follow up with a skopeo PR if/when this is merged to use it.

See:
https://github.com/containers/skopeo/issues/1440
https://github.com/containers/skopeo/issues/1378
https://github.com/containers/skopeo/issues/1102
https://github.com/containers/skopeo/issues/1451

Signed-off-by: James Hewitt <james.hewitt@uk.ibm.com>

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -80,13 +80,13 @@ type copier struct {
 
 // imageCopier tracks state specific to a single image (possibly an item of a manifest list)
 type imageCopier struct {
-	c                  *copier
-	manifestUpdates    *types.ManifestUpdateOptions
-	src                types.Image
-	diffIDsAreNeeded   bool
-	canModifyManifest  bool
-	canSubstituteBlobs bool
-	ociEncryptLayers   *[]int
+	c                          *copier
+	manifestUpdates            *types.ManifestUpdateOptions
+	src                        types.Image
+	diffIDsAreNeeded           bool
+	cannotModifyManifestReason string // The reason the manifest cannot be modified, or an empty string if it can
+	canSubstituteBlobs         bool
+	ociEncryptLayers           *[]int
 }
 
 const (
@@ -129,10 +129,14 @@ type Options struct {
 	DestinationCtx   *types.SystemContext
 	ProgressInterval time.Duration                 // time to wait between reports to signal the progress channel
 	Progress         chan types.ProgressProperties // Reported to when ProgressInterval has arrived for a single artifact+offset.
+
+	// Preserve digests, and fail if we cannot.
+	PreserveDigests bool
 	// manifest MIME type of image set by user. "" is default and means use the autodetection to the the manifest MIME type
 	ForceManifestMIMEType string
 	ImageListSelection    ImageListSelection // set to either CopySystemImage (the default), CopyAllImages, or CopySpecificImages to control which instances we copy when the source reference is a list; ignored if the source reference is not a list
 	Instances             []digest.Digest    // if ImageListSelection is CopySpecificImages, copy only these instances and the list itself
+
 	// If OciEncryptConfig is non-nil, it indicates that an image should be encrypted.
 	// The encryption options is derived from the construction of EncryptConfig object.
 	// Note: During initial encryption process of a layer, the resultant digest is not known
@@ -410,7 +414,35 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
 			return nil, errors.Wrapf(err, "Can not copy signatures to %s", transports.ImageName(c.dest.Reference()))
 		}
 	}
-	canModifyManifestList := (len(sigs) == 0)
+
+	// If the destination is a digested reference, make a note of that, determine what digest value we're
+	// expecting, and check that the source manifest matches it.
+	destIsDigestedReference := false
+	if named := c.dest.Reference().DockerReference(); named != nil {
+		if digested, ok := named.(reference.Digested); ok {
+			destIsDigestedReference = true
+			matches, err := manifest.MatchesDigest(manifestList, digested.Digest())
+			if err != nil {
+				return nil, errors.Wrapf(err, "computing digest of source image's manifest")
+			}
+			if !matches {
+				return nil, errors.New("Digest of source image's manifest would not match destination reference")
+			}
+		}
+	}
+
+	// Determine if we're allowed to modify the manifest list.
+	// If we can, set to the empty string. If we can't, set to the reason why.
+	cannotModifyManifestListReason := ""
+	if len(sigs) > 0 {
+		cannotModifyManifestListReason = "Would invalidate signatures"
+	}
+	if destIsDigestedReference {
+		cannotModifyManifestListReason = "Destination specifies a digest"
+	}
+	if options.PreserveDigests {
+		cannotModifyManifestListReason = "Instructed to preserve digests"
+	}
 
 	// Determine if we'll need to convert the manifest list to a different format.
 	forceListMIMEType := options.ForceManifestMIMEType
@@ -425,8 +457,8 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
 		return nil, errors.Wrapf(err, "determining manifest list type to write to destination")
 	}
 	if selectedListType != originalList.MIMEType() {
-		if !canModifyManifestList {
-			return nil, errors.Errorf("manifest list must be converted to type %q to be written to destination, but that would invalidate signatures", selectedListType)
+		if cannotModifyManifestListReason != "" {
+			return nil, errors.Errorf("Manifest list must be converted to type %q to be written to destination, but we cannot modify it: %q", selectedListType, cannotModifyManifestListReason)
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


mtrmac (commented):
Thanks! This is great to have.

One thing I’m wondering about are the error messages … ”we are preserving digests” is not going to be necessarily obvious to users a few levels up a call stack, e.g. if the image is signed. Maybe that should be something like “we can’t change the digest (image is singled)”, with the part in parenthesis determined when making the `canModify… = false` decision and stored in `copier`/`imageCopier`. (Or do others think this is good enough and we don’t need that over-engineering?)

copy/copy.go
```diff
@@ -410,7 +414,7 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
 			return nil, errors.Wrapf(err, "Can not copy signatures to %s", transports.ImageName(c.dest.Reference()))
 		}
 	}
-	canModifyManifestList := (len(sigs) == 0)
+	canModifyManifestList := (len(sigs) == 0) && !options.PreserveDigests
```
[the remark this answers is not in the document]
Jamstah: Sure, have added.

copy/copy.go
```diff
@@ -84,7 +84,7 @@ type imageCopier struct {
 	manifestUpdates    *types.ManifestUpdateOptions
 	src                types.Image
 	diffIDsAreNeeded   bool
-	canModifyManifest  bool
+	canModifyManifest  string
```
mtrmac, line 87: (Others: WDYT?)

Hum, I didn’t anticipate just using a string without a bool for this. But it works pretty well? Except that it’s non-obvious without understanding the full picture.

Either split this into `canModifyManifest bool` `cannotModifyManifestReason string`, or at least rename and document the strings to something like
```go
cannotModifyManifestReason string // Either "" if modification is allowed, or a reason to include in error messages
```

(Similarly for the `canModifyManifestList` variable.)

copy/copy.go
```diff
@@ -410,7 +414,34 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
...
+	// Determine if we're allowed to modify the manifest and why not
+	canModifyManifestList := ""
+	if len(sigs) > 0 {
+		canModifyManifestList = "Would invalidate signatures"
+	}
+	if destIsDigestedReference {
+		canModifyManifestList = "Destination specifies a digest"
+	}
+	if options.PreserveDigests {
+		canModifyManifestList = "Digests are being preserved"
```
mtrmac, line 443: Maybe something like “Instructed to preserve digests”, to point at some caller?

I’m not at all sure this is much better, we still can’t tell the user what this means exactly (e.g. what CLI option that was).

(This should be the same in both cases.)

copy/copy.go
```diff
@@ -813,9 +855,9 @@ func (ic *imageCopier) updateEmbeddedDockerReference() error {
 		return nil // No reference embedded in the manifest, or it matches destRef already.
 	}
 
-	if !ic.canModifyManifest {
-		return errors.Errorf("Copying a schema1 image with an embedded Docker reference to %s (Docker reference %s) would change the manifest, which is not possible (image is signed or the destination specifies a digest)",
-			transports.ImageName(ic.c.dest.Reference()), destRef.String())
+	if ic.canModifyManifest != "" {
+		return errors.Errorf("Copying a schema1 image with an embedded Docker reference to %s (Docker reference %s) would change the manifest, but cannot change the manifest: %q",
```
mtrmac, line 859: One of the others says “…, which we cannot do”, that would work nicely here.

Jamstah:
> One thing I’m wondering about are the error messages … ”we are preserving digests” is not going to be necessarily obvious to users a few levels up a call stack, e.g. if the image is signed. Maybe that should be something like “we can’t change the digest (image is singled)”, with the part in parenthesis determined when making the `canModify… = false` decision and stored in `copier`/`imageCopier`. (Or do others think this is good enough and we don’t need that over-engineering?)

I considered this, but thought it would be a bit messy. You're right though, it will be more usable.

I've had a go at an impl, see what you think.

mtrmac (commented):
Looks good overall, just nits for readability

copy/copy.go
Jamstah: I feel like it's fairly obvious by reading the code, but am happy to change it up to suit. I couldn't think of a _neat_ way to do it, which is one of the reasons I didn't do it in the first commit :)

Which would you prefer, two variables, or one with a better name and comment?

mtrmac: Basically I’m worried `if canModifyManifest != ""` is non-obvious (boolean naming + string type, “true” indicated by empty string) for busy people seeing a few lines in isolation, e.g. in some search results.

On balance, I’d probably prefer just the `cannot…Reason` name (or some better name… `fixedManifestReason`??? I’m bad at naming) + comment, because that way the two can’t get out of sync. I didn’t originally realize the …reason variable would probably be named the same in both cases.

(But this is all a somewhat weak preference, if others feel differently, please speak up.)

Jamstah: Yep, and I share your concern, it could be more obvious.

Have gone with one variable that is more appropriately named and better commented.

copy/copy.go
```diff
@@ -710,10 +755,10 @@ func (c *copier) copyOneImage(ctx context.Context, policyContext *signature.Poli
 		}
 		// If the original MIME type is acceptable, determineManifestConversion always uses it as preferredManifestMIMEType.
 		// So if we are here, we will definitely be trying to convert the manifest.
-		// With !ic.canModifyManifest, that would just be a string of repeated failures for the same reason,
+		// With !ic.cannotModifyManifestReason, that would just be a string of repeated failures for the same reason,
```
mtrmac, line 758:
```suggestion
		// With ic.cannotModifyManifestReason != "", that would just be a string of repeated failures for the same reason,
```

copy/copy.go
```diff
@@ -833,7 +878,7 @@ func isTTY(w io.Writer) bool {
 	return false
 }
 
-// copyLayers copies layers from ic.src/ic.c.rawSource to dest, using and updating ic.manifestUpdates if necessary and ic.canModifyManifest.
+// copyLayers copies layers from ic.src/ic.c.rawSource to dest, using and updating ic.manifestUpdates if necessary and ic.cannotModifyManifestReason.
```
mtrmac, line 881:
```suggestion
// copyLayers copies layers from ic.src/ic.c.rawSource to dest, using and updating ic.manifestUpdates if necessary and ic.cannotModifyManifestReason == "".
```

mtrmac (commented):
LGTM apart from the trivial nits. Thanks!

@vrothberg I’d appreciate a second opinion.

mtrmac (approved):

copy/copy.go
```diff
@@ -410,7 +414,35 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
...
+			if !matches {
+				return nil, errors.New("Digest of source image's manifest would not match destination reference")
+			}
+		}
+	}
+
+	// Determine if we're allowed to modify the manifest list.
+	// If we can, set to the empty string. If we can't, set to the reason why.
+	cannotModifyManifestListReason := ""
+	if len(sigs) > 0 {
```
vrothberg, line 437: I'd prefer this to be a `switch { case: }` or is it intended to possibly override the reason?

copy/copy.go
```diff
@@ -629,21 +661,34 @@ func (c *copier) copyOneImage(ctx context.Context, policyContext *signature.Poli
...
+	cannotModifyManifestReason := ""
+	if len(sigs) > 0 {
+		cannotModifyManifestReason = "Would invalidate signatures"
+	}
+	if destIsDigestedReference {
+		cannotModifyManifestReason = "Destination specifies a digest"
+	}
+	if options.PreserveDigests {
+		cannotModifyManifestReason = "Instructed to preserve digests"
+	}
```
vrothberg, line 675: This looks like a code clone above. It could be worth breaking that into a separate function to prevent diverging in the future.

vrothberg (commented):
Thanks, this is really nice! I just have some mini nits.

copy/copy.go
mtrmac: Any of the three conditions, possibly all three, can be true simultaneously.

I guess we could build a list here, but I’m personally not really worried about that, and not at all sure it’s worth the complexity.

copy/copy.go
mtrmac: (It seems likely, but not quite obvious, to me, that the two are going to stay in sync.)

@vrothberg did you have in mind a function with three boolean parameters? A struct with named fields, to avoid three boolean values in random order? Or something with a `copier, options, sigs` parameters?

At 2 iterations, just comments linking the two (“See also $function”) might be a low-tech approach.

mtrmac: It does feel that there is a cleaner version hidden, waiting to be discovered (e.g. the “get signatures that we want to copy” code can almost certainly be consolidated between the two paths), but at least the `destIsDigestedReference` check is legitimately different — so many intermediate “correct” versions are going to be more complex, and refactoring this to be _clean and simple_ feels rather out of scope of this contributed PR.

vrothberg: Agreed, let's merge and keep it in the back of our heads.

vrothberg (approved):
LGTM

copy/copy.go
mtrmac: Just to tug on this thread, #1417 is a tentative idea. I don’t think that one is actually worth it as is.

mtrmac: … and #1418 for the comment pair.

PR #1415, opened 2021-11-23
shortnames: mechanism to enforce resolving to Docker Hub

vrothberg:
Podman's Docker-compatible REST API is in need of a mechanism to enforce
resolving to Docker Hub only.  Yet there is the desire for the rest of
the stack to continue honoring the system's registries.conf.

We recently added a new field to containers.conf [1] which allows for
opting out from enforcing Docker Hub for Podman's compat API but we
still lack a way of enforcement when resolving short names; which
ultimately is *the* place to do that.

This change does the necessary plumbing.  The compat REST handlers will
set the new field in the `types.SystemContext` and pass that down to
libimage and buildah.

[1] https://github.com/containers/common/commit/e698b8caca9413a437a3d5ade87d690ba0969f22

Context: containers/podman/issues/12320

### DIFF
--- a/pkg/shortnames/shortnames.go
+++ b/pkg/shortnames/shortnames.go
@@ -118,6 +118,7 @@ type Resolved struct {
 }
 
 func (r *Resolved) addCandidate(named reference.Named) {
+	named = reference.TagNameOnly(named) // Make sure to add ":latest" if needed
 	r.PullCandidates = append(r.PullCandidates, PullCandidate{named, false, r})
 }
 
@@ -138,6 +139,8 @@ const (
 	rationaleUSR
 	// Resolved value has been selected by the user (via the prompt).
 	rationaleUserSelection
+	// Resolved value has been enforced to use Docker Hub (via SystemContext).
+	rationaleEnforcedDockerHub
 )
 
 // Description returns a human-readable description about the resolution
@@ -152,6 +155,8 @@ func (r *Resolved) Description() string {
 		return fmt.Sprintf("Resolved %q as an alias (%s)", r.userInput, r.originDescription)
 	case rationaleUSR:
 		return fmt.Sprintf("Resolving %q using unqualified-search registries (%s)", r.userInput, r.originDescription)
+	case rationaleEnforcedDockerHub:
+		return fmt.Sprintf("Resolving %q to docker.io (%s)", r.userInput, r.originDescription)
 	case rationaleUserSelection, rationaleNone:
 		fallthrough
 	default:
@@ -265,8 +270,20 @@ func Resolve(ctx *types.SystemContext, name string) (*Resolved, error) {
 		return nil, err
 	}
 	if !isShort { // no short name
-		named := reference.TagNameOnly(shortRef) // Make sure to add ":latest" if needed
+		resolved.addCandidate(shortRef)
+		return resolved, nil
+	}
+
+	// Resolve to docker.io only if enforced by the caller (e.g., Podman's
+	// Docker-compatible REST API).
+	if ctx != nil && ctx.PodmanOnlyShortNamesIgnoreRegistriesConfAndForceDockerHub {
+		named, err := reference.ParseNormalizedNamed(name)
+		if err != nil {
+			return nil, errors.Wrapf(err, "cannot normalize input: %q", name)
+		}
 		resolved.addCandidate(named)
+		resolved.rationale = rationaleEnforcedDockerHub
+		resolved.originDescription = "enforced by caller"
 		return resolved, nil
 	}
 
@@ -295,9 +312,6 @@ func Resolve(ctx *types.SystemContext, name string) (*Resolved, error) {
 				return nil, err
 			}
 		}
-		// Make sure to add ":latest" if needed
-		namedAlias = reference.TagNameOnly(namedAlias)
-
 		resolved.addCandidate(namedAlias)
 		resolved.rationale = rationaleAlias
 		resolved.originDescription = aliasOriginDescription
@@ -325,9 +339,6 @@ func Resolve(ctx *types.SystemContext, name string) (*Resolved, error) {
 		if err != nil {
 			return nil, errors.Wrapf(err, "creating reference with unqualified-search registry %q", reg)
 		}
-		// Make sure to add ":latest" if needed
-		named = reference.TagNameOnly(named)
-
 		resolved.addCandidate(named)
 	}
 
@@ -412,6 +423,23 @@ func ResolveLocally(ctx *types.SystemContext, name string) ([]reference.Named, e
 
 	var candidates []reference.Named
 
+	// Complete the candidates with the specified registries.
+	completeCandidates := func(registries []string) ([]reference.Named, error) {
+		for _, reg := range registries {
+			named, err := reference.ParseNormalizedNamed(fmt.Sprintf("%s/%s", reg, name))
+			if err != nil {
+				return nil, errors.Wrapf(err, "creating reference with unqualified-search registry %q", reg)
+			}
+			named = reference.TagNameOnly(named) // Make sure to add ":latest" if needed
+			candidates = append(candidates, named)
+		}
+		return candidates, nil
+	}
+
+	if ctx != nil && ctx.PodmanOnlyShortNamesIgnoreRegistriesConfAndForceDockerHub {
+		return completeCandidates([]string{"docker.io"})
+	}
+
 	// Strip off the tag to normalize the short name for looking it up in
 	// the config files.
 	isTagged, isDigested, shortNameRepo, tag, digest := splitUserInput(shortRef)
@@ -434,9 +462,7 @@ func ResolveLocally(ctx *types.SystemContext, name string) ([]reference.Named, e
 				return nil, err
 			}
 		}
-		// Make sure to add ":latest" if needed
-		namedAlias = reference.TagNameOnly(namedAlias)
-
+		namedAlias = reference.TagNameOnly(namedAlias) // Make sure to add ":latest" if needed
 		candidates = append(candidates, namedAlias)
 	}
 
@@ -447,16 +473,5 @@ func ResolveLocally(ctx *types.SystemContext, name string) ([]reference.Named, e
 	}
 
 	// Note that "localhost" has precedence over the unqualified-search registries.
-	for _, reg := range append([]string{"localhost"}, unqualifiedSearchRegistries...) {
-		named, err := reference.ParseNormalizedNamed(fmt.Sprintf("%s/%s", reg, name))
-		if err != nil {
-			return nil, errors.Wrapf(err, "creating reference with unqualified-search registry %q", reg)
-		}
-		// Make sure to add ":latest" if needed
-		named = reference.TagNameOnly(named)
-
-		candidates = append(candidates, named)
-	}
-
-	return candidates, nil
+	return completeCandidates(append([]string{"localhost"}, unqualifiedSearchRegistries...))
 }
--- a/pkg/shortnames/shortnames_test.go
+++ b/pkg/shortnames/shortnames_test.go
@@ -105,35 +105,86 @@ func TestResolve(t *testing.T) {
 		UserShortNameAliasConfPath:  tmp.Name(),
 	}
 
+	sysResolveToDockerHub := &types.SystemContext{
+		SystemRegistriesConfPath:                                  "testdata/aliases.conf",
+		SystemRegistriesConfDirPath:                               "testdata/this-does-not-exist",
+		UserShortNameAliasConfPath:                                tmp.Name(),
+		PodmanOnlyShortNamesIgnoreRegistriesConfAndForceDockerHub: true,
+	}
+
 	_, err = sysregistriesv2.TryUpdatingCache(sys)
 	require.NoError(t, err)
 
+	digest := "@sha256:d366a4665ab44f0648d7a00ae3fae139d55e32f9712c67accd604bb55df9d05a"
+
 	tests := []struct {
-		name, value string
+		input, value, dockerHubValue string
 	}{
-		{"docker", "docker.io/library/foo:latest"},
-		{"docker:tag", "docker.io/library/foo:tag"},
-		{
-			"docker@sha256:d366a4665ab44f0648d7a00ae3fae139d55e32f9712c67accd604bb55df9d05a",
-			"docker.io/library/foo@sha256:d366a4665ab44f0648d7a00ae3fae139d55e32f9712c67accd604bb55df9d05a",
+		{ // alias
+			"docker",
+			"docker.io/library/foo:latest",
+			"docker.io/library/docker:latest",
 		},
-		{"quay/foo", "quay.io/library/foo:latest"},
-		{"quay/foo:tag", "quay.io/library/foo:tag"},
-		{
-			"quay/foo@sha256:d366a4665ab44f0648d7a00ae3fae139d55e32f9712c67accd604bb55df9d05a",
-			"quay.io/library/foo@sha256:d366a4665ab44f0648d7a00ae3fae139d55e32f9712c67accd604bb55df9d05a",
+		{ // alias tagged
+			"docker:tag",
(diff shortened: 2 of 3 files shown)


types/types.go
```diff
@@ -561,6 +561,8 @@ type SystemContext struct {
 	UserShortNameAliasConfPath string
 	// If set, short-name resolution in pkg/shortnames must follow the specified mode
 	ShortNameMode *ShortNameMode
+	// If set, short names will resolve docker.io, and unqualified-search registries and short-name alias are ignored.
```
vrothberg, line 564:
```suggestion
	// If set, short names will resolve docker.io, and unqualified-search registries and short-name aliases are ignored.
```

vrothberg:
@mtrmac @rhatdan PTAL

rhatdan:
LGTM

mtrmac (commented):
I’m not really a fan.

To the extent short names are a primarily a UI feature, with user-targeted interactions and semantics, it’s user-hostile for those names to have option-dependent semantics.

The way I think about it, the case this is targeting we claim that the users didn’t want to use short names with alias/search semantics, their client software didn’t want to represent users’s wishes over the AF_UNIX socket with alias/search semantics, and the API server understands / is configured to understand the AF_UNIX request to not have alias/search semantics. So it seems natural to me that the API server shouldn’t call the alias/search semantics code, then.

We don’t have `docker://` call `shortnames.Resolve(ResolveShortNamesToDockerHub)`. Why is the API server client a different case?

(Yeah, to an extent this is just aesthetics; we are going to have the ugly special case _somewhere_. Still… we don’t want _anyone else_, **ever**, to turn this option on, do we?)

The one reason I can see for doing it this way is the `ResolveLocally` path, where `localhost/` is _still_ preferred. I’m not sure that is a good enough reason, though opinions on that can certainly differ — and come to think of it, I’m not even sure the API should prefer `localhost/` in that case, then. If, in the API, “short names mean `docker.io`”, shouldn’t that equally apply to _everything_, including names of pulled images, outcomes of local builds, local tag canonicalization, and everything else? And shouldn’t the `localhost/` lookup then be unnecessary at least, or quite possibly actively unwanted?

types/types.go
mtrmac: Also … resolve to …

vrothberg:
> I’m not really a fan.

Me neither but I do not see any other long-term maintainable solution.

The code is scattered across at least three different repositories: c/podman, c/common, c/buildah.  I want to avoid having the logic (and the decision of running in compat-mode) scattered across all these repositories. c/image, and the `shortnames` package in particular, is the heart of short-names resolution so it seemed natural to me to move it. Setting one bool in the SystemsContext (which is used across the stack) in the compat API and leave the rest of the stack untouched seems very attractive to me.

Note that  am still convinced users (or distributions) should edit registries.conf if they desire the specific behavior.

> (Yeah, to an extent this is just aesthetics; we are going to have the ugly special case _somewhere_. Still… we don’t want _anyone else_, **ever**, to turn this option on, do we?)
> 
> The one reason I can see for doing it this way is the `ResolveLocally` path, where `localhost/` is _still_ preferred. I’m not sure that is a good enough reason, though opinions on that can certainly differ — and come to think of it, I’m not even sure the API should prefer `localhost/` in that case, then. If, in the API, “short names mean `docker.io`”, shouldn’t that equally apply to _everything_, including names of pulled images, outcomes of local builds, local tag canonicalization, and everything else? And shouldn’t the `localhost/` lookup then be unnecessary at least, or quite possibly actively unwanted?

Yes, good point.  `localhost/` should be excluded as well and the tag/untag behavior in the compat API and the commit logic in Buildah had to be updated as well.

mtrmac:
> Me neither but I do not see any other long-term maintainable solution.

I’d expect this to be a boolean handled by all the API handlers. Sure, there are quite a few, but it’s routine work that only needs to be done once because the API is frozen, isn’t it?

In particular, a “build” request which contains a plain-text `Dockerfile` with `FROM fedora` is _not_ a case of the “in the API, a short name might mean docker.io” situation, because there isn’t a Docker Engine API client passing explicitly fully-qualified user input through `FamiliarName` to turn it into a short name against user wishes.

If that case should also interpret short names as `docker.io` (which I assume/guess is the primary hard case motivating putting this into `SystemContext`), it’s completely unclear to me why that should be so.

mtrmac:
> and the tag/untag behavior in the compat API and the commit logic in Buildah had to be updated as well.

Using `pkg/shortnames` on those paths really doesn’t make sense to me in the general case, because `Resolve` returns a list of candidates. It would work _better_ with this proposed new flag that _without_!

vrothberg:
> I’d expect this to be a boolean handled by all the API handlers. Sure, there are quite a few, but it’s routine work that only needs to be done once because the API is frozen, isn’t it?

That's what I hoped for yes, but as you suspect, build is the hairy part.
 
> In particular, a “build” request which contains a plain-text `Dockerfile` with `FROM fedora` is _not_ a case of the “in the API, a short name might mean docker.io” situation, because there isn’t a Docker Engine API client passing explicitly fully-qualified user input through `FamiliarName` to turn it into a short name against user wishes.

I am not sure I understand you correctly.  Do you mean that building should continue using search registries and only the parts of the REST API with images in parameters would be updated?

mtrmac:
> Do you mean that building should continue using search registries and only the parts of the REST API with images in parameters would be updated?

Yes. I see a reason for interpreting the API-native reference values (where e.g. `docker/docker/client` accepts `reference.Named` values and calls `FamiliarName`) to imply `docker.io`; I don’t see one for the plain text `Dockerfile`s. (Users can always set U-S-R to `docker.io` to get that behavior, sure.)

It’s admittedly all very messy whatever we do.

vrothberg:
I would feel very uncomfortable with build resolving differently than the rest. Note sure if it could be security relevant but at the very least it would be inconsistent and probably annoying. I wanted to avoid doing this (or a similar) PR here in c/image but I do not see an alternative without introducing new fields/parameters to large parts Podman's call stack. 

Especially, I don't want future generations having to worry too much about that stuff. "Flip this knob in types.SystemContext" seems very attractive. The only alternative I see is trying to wire it into libimage but that would require updating many APIs that I really want to avoid.

> It’s admittedly all very messy whatever we do.

I know :sob:

pkg/shortnames/shortnames.go
```diff
@@ -246,6 +250,21 @@ func Resolve(ctx *types.SystemContext, name string) (*Resolved, error) {
 	}
 	resolved.systemContext = ctx
 
+	// Resolve to docker.io only if enforced by the caller (e.g., Podman's
+	// Docker-compatible REST API).
+	if ctx.ResolveShortNamesToDockerHub {
+		named, err := reference.ParseNormalizedNamed(name)
```
mtrmac, line 256: I think this should only happen after the `!isShort` case, to make sure truly fully-qualified names behave exactly the same everywhere. We can re-normalize at that point (similarly to what `parseUnnormalizedShortName`) does but we shouldn’t re-parse the string input from scratch.

pkg/shortnames/shortnames.go
```diff
@@ -246,6 +250,21 @@ func Resolve(ctx *types.SystemContext, name string) (*Resolved, error) {
...
+		if err != nil {
+			return nil, errors.Wrapf(err, "cannot normalize input: %q", name)
+		}
+		// Make sure to add ":latest" if needed
+		named = reference.TagNameOnly(named)
```
mtrmac, line 261: Please make the `TagNameOnly` call a responsibility of `addCandidate`, to reduce duplication now that there are going to be 4 copies.

pkg/shortnames/shortnames.go
```diff
@@ -412,6 +431,26 @@ func ResolveLocally(ctx *types.SystemContext, name string) ([]reference.Named, e
...
+			// Make sure to add ":latest" if needed
+			named = reference.TagNameOnly(named)
+
+			candidates = append(candidates, named)
+		}
+		return candidates, nil
+	}
+
+	if ctx != nil && ctx.ResolveShortNamesToDockerHub {
+		return completeCandidates([]string{"docker.io"})
```
mtrmac, line 451: So per the earlier conversation, we should’t be including `localhost/` here, I think. That also probably implies no `completeCandidates` helper.

pkg/shortnames/shortnames.go
```diff
@@ -138,6 +138,8 @@ const (
 	rationaleUSR
 	// Resolved value has been selected by the user (via the prompt).
 	rationaleUserSelection
+	// Resolved value has been enforced to use Docker Hub (via SystemContext).
+	rationaleEnforcedDockerHub
```
mtrmac, line 142: (I’ve been wondering if we even need this, or if we should just treat the input _exactly_ the same as fully-qualified.

Ideally (by hypothesis that there really are no short-name users, at least) users are _never_ going to see the short names and worry how they became `docker.io`-qualified.

OTOH it doesn’t hurt, sure, and if we _did_ ever encounter such users, the traceability is certainly nice.)

pkg/shortnames/shortnames_test.go
```diff
@@ -479,13 +505,25 @@ func TestResolveLocally(t *testing.T) {
 	assert.Equal(t, "quay.io/repo/image:latest", aliases[2].String())      // registry 0
 	assert.Equal(t, "registry.com/repo/image:latest", aliases[3].String()) // registry 0
 
+	aliases, err = ResolveLocally(sysResolveToDockerHub, "repo/image") // alias match but enforced
+	require.NoError(t, err)
+	require.Len(t, aliases, 2)                                          // localhost + docker.io enforced
+	assert.Equal(t, "localhost/repo/image:latest", aliases[0].String()) // localhost
+	assert.Equal(t, "docker.io/repo/image:latest", aliases[1].String()) // docker.io enforced
+
```
mtrmac, line 513: It seems to me that turning this long repetitive function into a table-driven test, with a `[]struct { input string; expected []string }`, is warranted enough to be a review blocker at this point, though I don’t feel _extremely_ strongly about it.

pkg/shortnames/shortnames_test.go
```diff
@@ -105,29 +105,39 @@ func TestResolve(t *testing.T) {
...
+		UserShortNameAliasConfPath:   tmp.Name(),
+		ResolveShortNamesToDockerHub: true,
+	}
+
 	_, err = sysregistriesv2.TryUpdatingCache(sys)
 	require.NoError(t, err)
 
 	tests := []struct {
-		name, value string
+		name, value, dockerHubValue string
```
mtrmac, line 119: While here, please `s/name/input/`, or something similar.

pkg/shortnames/shortnames_test.go
```diff
@@ -141,6 +151,16 @@ func TestResolve(t *testing.T) {
...
+	// Now another run with enforcing resolution to Docker Hub.
+	for _, test := range tests {
+		resolved, err := Resolve(sysResolveToDockerHub, test.name)
+		require.NoError(t, err, "%v", test)
+		require.NotNil(t, resolved)
+		require.Len(t, resolved.PullCandidates, 1)
+		assert.Equal(t, test.dockerHubValue, resolved.PullCandidates[0].Value.String())
+		assert.False(t, resolved.PullCandidates[0].record)
+	}
+
```
mtrmac, line 163: The above only tests defined aliases AFAICS. The `doesnotexist` and fully-qualified cases below might also be worth expanding (and perhaps worth making table-driven).

pkg/shortnames/shortnames_test.go
```diff
@@ -470,6 +490,12 @@ func TestResolveLocally(t *testing.T) {
 		SystemRegistriesConfDirPath: "testdata/this-does-not-exist",
 		UserShortNameAliasConfPath:  tmp.Name(),
 	}
+	sysResolveToDockerHub := &types.SystemContext{
```
mtrmac, line 493: Note to self: I didn’t actually review the test cases, I expect them to change if `localhost/` is not used in the Docker case.

types/types.go
```diff
@@ -561,6 +561,8 @@ type SystemContext struct {
...
+	ResolveShortNamesToDockerHub bool
```
mtrmac, line 565: So how scary can we make this name?
`PodmanDockerEngineAPICompatOnlyResolveShortNamesToDockerHub` 😛 ? I guess not _that_, but it should stick out at least as much as …InsecureSkipTLSVerify does.

Also the documentation should
- refer to pkg/shortname
- explicitly record that this is only intended to help implementing the Docker Engine API,
- warn that this should NEVER be used by any new features and for any other purpose

mtrmac (commented):
So per an off-GitHub conversation, this is considered to the easiest way to centralize the logic (because the “always `docker.io`” semantics will be configurable — and maybe c/common/libimage will read the new `SystemContext` boolean as a switch for its own code that doesn’t involve pkg/shortnames), and we are probably doing it…

types/types.go
mtrmac: … and possibly even  warn that the exact semantics may change to support that use case only, and only as implemented by Podman + c/common. I worry a bit that e.g. the `localhost/` part is not yet 100% certain, and we might discover that we need to change the behavior only after making a c/image release.

Alternatively, name it `UnstableUnsupported…` and explicitly exclude it from the general c/image API stability effort.

Either way, if we did need to change the semantics, c/image + the callers would end up with tightly coupled versions. Or would we actually drop this field, and introduce a differently-named field for the new semantics?

vrothberg:
Thanks for reviewing, @mtrmac.  Here's an updated version.

vrothberg:
Converting to draft for now.  I need to spend some more time in `libimage`

pkg/shortnames/shortnames.go
```diff
@@ -446,17 +472,5 @@ func ResolveLocally(ctx *types.SystemContext, name string) ([]reference.Named, e
 		return nil, err
 	}
 
-	// Note that "localhost" has precedence over the unqualified-search registries.
```
mtrmac, line 449: Was removing this comment intentional?

types/types.go
```diff
@@ -561,6 +561,11 @@ type SystemContext struct {
 	UserShortNameAliasConfPath string
 	// If set, short-name resolution in pkg/shortnames must follow the specified mode
 	ShortNameMode *ShortNameMode
+	// If set, short names will resolve to docker.io only, and unqualified-search registries and short-name aliases in
+	// registries.conf are ignored.  Note that this field should only be used in pkg/shortnames and is only intended
```
mtrmac, line 565:
```suggestion
	// If set, short names will resolve in pkg/shortnames to docker.io only, and unqualified-search registries and short-name aliases in
	// registries.conf are ignored.  Note that this is only intended
```

mtrmac (commented):
(I know another approach is being investigated; just getting this off my plate.)

vrothberg:
> (I know another approach is being investigated; just getting this off my plate.)

Thanks!  I am _almost_ done with the REST API and am currently working on the last tests which are on builds.

vrothberg:
Ready from my side

mtrmac (commented):
LGTM. So is the approach that is going to be used? A clean split between the build code and the rest? Some kind of ad-hoc mix?

vrothberg:
> LGTM. So is the approach that is going to be used? 

Yes, I opened https://github.com/containers/podman/pull/12435.

> A clean split between the build code and the rest? Some kind of ad-hoc mix?

I have a hard time considering it to be "clean" since I'd still prefer a change in `registries.conf` but "ad-hoc mix" sounds accurate to me.  At the time of writing, https://github.com/containers/podman/pull/12435 normalizes all parameters it can and leaves the rest to c/buildah via an updated `types.SystemContext`.

vrothberg:
@mtrmac good to merge?

mtrmac:
> > LGTM. So is the approach that is going to be used?
> 
> Yes, I opened [containers/podman#12435](https://github.com/containers/podman/pull/12435).
> 
> > A clean split between the build code and the rest? Some kind of ad-hoc mix?
> 
> I have a hard time considering it to be "clean"

Very fair :)

mtrmac (approved):
\*sigh\*

PR #1416, opened 2021-11-23
be more explicit about multiple image references for oci

lsm5:

@mtrmac @vrothberg @rhatdan PTAL.

RE: https://github.com/containers/skopeo/issues/1513

### DIFF
--- a/docs/containers-transports.5.md
+++ b/docs/containers-transports.5.md
@@ -60,7 +60,7 @@ The _algo:digest_ refers to the image ID reported by docker-inspect(1).
 ### **oci:**_path[:reference]_
 
 An image compliant with the "Open Container Image Layout Specification" at _path_.
-Using a _reference_ is optional and allows for storing multiple images at the same _path_.
+Using a _reference_ is optional for a single image reference but required for storing multiple images at the same _path_.
 
 ### **oci-archive:**_path[:reference]_
 


lsm5:
does [CI:DOCS] need to be in the commit title? I added it to the commit message body, but it's still going through CI.

docs/containers-transports.5.md
```diff
@@ -60,7 +60,7 @@ The _algo:digest_ refers to the image ID reported by docker-inspect(1).
 ### **oci:**_path[:reference]_
 
 An image compliant with the "Open Container Image Layout Specification" at _path_.
-Using a _reference_ is optional and allows for storing multiple images at the same _path_.
+Using a _reference_ is optional for a single image reference but required for storing multiple images at the same _path_.
```
mtrmac, line 63: This is not quite precise. It’s _accessing_ an image from a multi-image archive that requires providing a disambiguating name; the code will happily enough add (untagged/unnamed) images to an archive, as the motivating bug report demonstrates.

lsm5: @mtrmac would you prefer an additional sentence and/or an example in the docs for clarifying the github issue case? Something like `pulling multiple images without specification of reference will not preserve all references` ?

mtrmac: > pulling multiple images without specification of reference will not preserve all references ?

I’m not quite sure what that means. (It’s a general principle of `skopeo copy` that names used in the source reference don’t affect names in the destination reference; it’s not something specific to this transport. So, copies by `skopeo copy` never “preserve” names/references in that sense — OTOH Podman _does_ have various heuristics of that kind.)

---

The general kinds information we probably want to convey for the two transports:
* (Maybe: The _reference_ value corresponds to the `org.opencontainers.image.ref.name` annotation.)
* When reading images from `oci:` or `oci-archive:`, if _path_ contains multiple images, a _reference_ must be specified to disambiguate; otherwise, the only image is implied.
* When writing images, _reference_ is optional and set on the written image [to allow disambiguating reads]
* Repeated image writes to `oci:` don’t overwrite pre-existing images, whether or not _reference_ is provided (but every _reference_ will only point to one image).
* Repeated image writes to `oci-archive:` overwrite the whole archive; only one image, the last written one, is included. (Do we want to _guarantee_ that? Probably, ambiguity doesn’t help anyone, but I’m not 100% sure.)

(Warning: I didn’t actually test any of this right now, just reading the code.)

mtrmac:
The core of this PR has since been documented by https://github.com/containers/image/pull/2042 ; https://github.com/containers/image/pull/1416#discussion_r755389115 contains a few more items that might be worth highlighting.

mtrmac:
> [#1416 (comment)](https://github.com/containers/image/pull/1416#discussion_r755389115) contains a few more items that might be worth highlighting.

I have filed #2468 for the missing part.

Thanks for this PR!

PR #1417, opened 2021-11-29
RFC: Consolidate getting signatures

mtrmac:
Consolidate figuring out which signatures from the source to copy, and whether that can work, into a helper.

@vrothberg RFC — I’m honestly _rather_ unhappy about passing the progress messages into the helper (and, at the same time, unwilling to make the messages less descriptive just to share a few lines of code); basically just to show that something like this is possible, and solicit better ideas.

At this point I’d really prefer only the first commit to be merged.

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -411,7 +411,7 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
 		sigs = []internalsig.Signature{}
 	} else {
 		c.Printf("Getting image list signatures\n")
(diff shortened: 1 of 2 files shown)


vrothberg (commented):
Both look like an improvement to me.

LGTM

rhatdan:
LGTM
Is this still a draft?

mtrmac:
I’m not _that_ happy about the end state, but I don’t think I will think of a better approach with any more time, so let’s do that.

PR #1419, opened 2021-11-29
manifest.GuessMIMEType(): recognize self-described OCI manifests

nalind:
Newer versions of the OCI spec include the mediaType field in image manifests and image indexes, and if they include them, save ourselves some work by using that information.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -26,7 +26,7 @@ require (
 	github.com/morikuni/aec v1.0.0 // indirect
 	github.com/mtrmac/gpgme v0.1.2
 	github.com/opencontainers/go-digest v1.0.0
-	github.com/opencontainers/image-spec v1.0.2-0.20210819154149-5ad6f50d6283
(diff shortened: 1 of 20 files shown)


vrothberg (commented):
Good idea, LGTM!

mtrmac (commented):
Please update also the comment below, something like per https://github.com/mtrmac/image/commit/665a421426384646db45e0df6e41d77198b2ba39 .

(I’m not sure when I’ll finish that wider work, so merging this part now makes perfect sense.)

nalind:
Oof, I need to add tests for this.

nalind:
Added "nomime" variants to the existing OCI test cases for `GuessMIMEType()`, and started setting the `MediaType` field by default in places where we construct OCI manifests.  Basically swiped @mtrmac's updates from https://github.com/mtrmac/image/commit/665a421426384646db45e0df6e41d77198b2ba39.

mtrmac (commented):
LGTM assuming tests pass. Thanks!

rhatdan:
LGTM

PR #1427, opened 2021-12-07
docker/config: handle credentials not found errors

vrothberg:
A credential helper may not have the credentials we are looking for and
exits with 1 (and specific error).  Make sure we're treating this case
correctly and don't error but return empty credentials.

Fixes: containers/podman/issues/11636

@rhatdan @giuseppe @umohnani8 PTAL

### DIFF
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -620,6 +620,10 @@ func getAuthFromCredHelper(credHelper, registry string) (types.DockerAuthConfig,
 	p := helperclient.NewShellProgramFunc(helperName)
 	creds, err := helperclient.Get(p, registry)
 	if err != nil {
+		if credentials.IsErrCredentialsNotFoundMessage(err.Error()) {
+			logrus.Debugf("Not logged in to %s with credential helper %s", registry, credHelper)
+			err = nil
+		}
 		return types.DockerAuthConfig{}, err
 	}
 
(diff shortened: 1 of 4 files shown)


vrothberg:
@saschagrunert PTAL

rhatdan:
LGTM

umohnani8:
LGTM

mtrmac:
That was definitely a bug for `GetCredentials`, but that should affect only helpers configured for the repository in question. The https://github.com/containers/podman/issues/11636 report (pulling `docker.io` but an `amazonaws.com` credential configured) could AFAICS only happen via `GetAllCredentials` — but that call already had a check for `credentials.IsErrCredentialsNotFoundMessage`. 

What am I missing? The above suggests that the check for `credentials.IsErrCredentialsNotFoundMessage` in `GetAllCredentials` was broken, which seems likely (there is a `multierror` interposed) — so that check:
- needs to be fixed somehow
- is dead code and should be removed
- needs to be replaced by a check for the correct return value `(DockerAuthConfig{}, nil)` – probably this one.

Or is something entirely different going on?

vrothberg:
> is dead code and should be removed

Definitely.

> needs to be replaced by a check for the correct return value (DockerAuthConfig{}, nil) – probably this one.

Probably a good idea to prevent adding empty structs to the map :+1:

mtrmac:
> The https://github.com/containers/podman/issues/11636 report (pulling `docker.io` but an `amazonaws.com` credential configured) could AFAICS only happen via `GetAllCredentials` — but that call already had a check for `credentials.IsErrCredentialsNotFoundMessage`. 

#1432, for the record.

PR #1432, opened 2022-01-04
Fix handling of missing data in GetAllCredentials

mtrmac:
- Don't check for `credentials.IsErrCredentialsNotFoundMessage`, which is handled by `GetCredentials` and is not reachable here
- Do check for an empty struct, which is the outcome of that handling. Don't add that empty struct, meaning "nothing found", to the returned data.
- A somewhat-related cleanup.

This is a follow-up to https://github.com/containers/image/pull/1427#issuecomment-991109516 .

### DIFF
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -285,7 +285,7 @@ func getCredentialsWithHomeDir(sys *types.SystemContext, key, homeDir string) (t
 				return types.DockerAuthConfig{}, "", err
 			}
 
-			if (authConfig.Username != "" && authConfig.Password != "") || authConfig.IdentityToken != "" {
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
@flouthoc PTAL

rhatdan:
@lsm5 PTAL

TomSweeneyRedHat:
LGTM
would like a head nod from @flouthoc too

flouthoc (approved):
LGTM

Since we return empty `DockerAuthConfig{}` when nothing is found: https://github.com/containers/image/blob/main/pkg/docker/config/config.go#L338

vrothberg (commented):
LGTM

PR #1434, opened 2022-01-06
RFC: Update golang.org/x/crypto, and silence warnings about openpgp

mtrmac:
This is a fallback code; the primary recommendation is to use the gpgme mechanism implementation, which is out-of-process and more appropriate for handling long-term private key material than any Go implementation.

For this verify-only fallback, we haven't reviewed any of the existing alternatives to choose; so, for now, continue to use this frozen deprecated implementation.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -39,8 +39,8 @@ require (
 	github.com/xeipuuv/gojsonschema v1.2.0
 	go.etcd.io/bbolt v1.3.6
 	go.opencensus.io v0.23.0 // indirect
-	golang.org/x/crypto v0.0.0-20210322153248-0c34fe9e7dc2
-	golang.org/x/net v0.0.0-20211005001312-d4b1ae081e3b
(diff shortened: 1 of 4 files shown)


mtrmac:
RFC. This is by no means a well-researched PR.

The primary reason for doing this is updating golang.org/x/crypto to a recent version breaks the c/image build with an “golang.org/x/crypto/openpgp is deprecated” warning, and that could happen to us on any future dependency update. So we are going to have to deal with this soon anyway.

For the reasons quoted above, I think continuing to use a frozen version is just about acceptable, but I could well be wrong about that.

Alternatively, we would need to find another, maintained, OpenPGP implementation (and probably decide which one, among several forks, is most viable to bet on). I haven’t done any research on this, at all.

rhatdan:
LGTM

TomSweeneyRedHat:
Changes LGTM, but the tests aren't hip.

vrothberg (approved):
LGTM

PR #1436, opened 2022-01-10
sif: initial sif transport implementation

tri-adam:
Initial SIF transport introduced by @yhcote in #1402, adapted to use upstream SIF implementation (as requested [here](https://github.com/containers/image/pull/1402#issuecomment-1008005231)).

Includes change to use `OptLoadWithFlag(os.O_RDONLY)` as suggested in https://github.com/containers/image/pull/1402#issuecomment-1006678214.


mtrmac:
Thanks! I have included this work in #1438 .

PR #1437, opened 2022-01-10
Release c/image v5.18.0

mtrmac:
Cc: @rhatdan @mheon

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -6,12 +6,12 @@ const (
 	// VersionMajor is for an API incompatible changes
 	VersionMajor = 5
 	// VersionMinor is for functionality in a backwards-compatible manner
-	VersionMinor = 17
+	VersionMinor = 18
 	// VersionPatch is for backwards-compatible bug fixes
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

mtrmac:
v5.18.0 tagged.

vrothberg:
Don't we still need the SIF work for 4.0? Once that merges, we'd need to do another bump to 5.19.

rhatdan:
Yup, but I would like to get everything tested now.

vrothberg:
`go get github.com/containers/image/v5@main` would be enough for testing. I prefer less releases.

rhatdan:
I was hoping to trigger dependabots, but they seem to be not working.

mtrmac:
I’m not sure, but it seems to me that Dependabot is reluctant to upgrade from specific git commit pseudo-versions — and both Buildah and Podman use such a git commit reference.

rhatdan:
I am not seeing them in common either.

PR #1438, opened 2022-01-10
Initial sif transport implementation

mtrmac:
This is a follow-up to #1402 and #1436 (many thanks to both @yhcote and @tri-adam), self-addressing my review from #1402.

Major changes:
- don't create a compressed image
- clean up various temporary files/directories ASAP to decrease disk space usage
- eliminate a single-use object created by newImageSource, instead just use a function. I think that’s conceptually cleaner but it might well not be worth it in practice (accepting and returning several unnamed values), I’d be happy to change this back again.

Marking as draft because I this is a 109-commit version, primarily to allow easy fixes/updates if I broke anything during those changes (which did happen during development, several times), but we probably don’t want that in the history. In the end, because this is entirely new code to the project, this can be the contributed commits, minimal updates to make the code build, and maybe a single large commit to update.

Only minimally tested, per https://github.com/containers/image/pull/1402#issuecomment-1008988259 .


sif/load.go
```diff
@@ -0,0 +1,206 @@
...
+	if err := ioutil.WriteFile(filePath, injectedScript, 0755); err != nil {
+		return fmt.Errorf("writing %s to %s: %w", injectedScriptTargetPath, filePath, err)
+	}
+	return nil
+}
+
+// createTarFromSIFInputs creates a tar file at tarPath, using a squashfs image at squashFSPath.
+// It can also use extractedRootPath and scriptPath, which are allocated for its exclusive use,
+// if necessary.
+func createTarFromSIFInputs(ctx context.Context, tarPath, squashFSPath string, injectedScript []byte, extractedRootPath, scriptPath string) error {
```
vrothberg, line 114: This function can take _a lot of_ time. What I would love to have here is a progress-bar spinner to indicate that we're doing some very heavy lifting.

Unfinished thought: I wonder if we could add an io.Writer to `types.SystemContext` similar as in `copy.Options`.

vrothberg (commented):
Really great work! In addition to the progress-bar comment, I would appreciate a commit adding debug messages.

sif/load.go
mtrmac: Yeah, a progress indication would definitely be useful. It’s not reasonably possible in the current API (and several other transports that e.g. make temporary files on disk would benefit).  We now have `internal/types`, so it’s something that can be built (at least for `c/image/copy`, where we don’t have to worry whether to expose MBP as a public API commitment).

Just an `io.Writer` would not be great, this needs to account for interactive use (frequent progress updates), completely non-interactive use (no progress updates, just a log of successes), and (per a RFE) a middle ground of plain-text updates a few times a minute. We might need to build an internal progress abstraction, and I don’t think we can commit to a public `io.Writer` API in `types.SystemContext`, just like that, at this point.

Given the Podman 4.0 timing, would it be sufficient to just wrap that operation in a pair of debug logs?

mtrmac:
I have added debug messages around the long-running operations.

We could also log size+offset of the found descriptors to get a bit more tracing, OTOH that can be read using `siftool` at any time.

vrothberg (approved):
LGTM

vrothberg:
@mtrmac feel free to merge at will.

mtrmac:
I have left the original commits, some clearly stand-alone things (build fixes, allowing non-Linux builds), file moves, and squashed the rest into one commit.

I‘ll keep the equivalent https://github.com/mtrmac/image/tree/sif-many-commits around for a bit (a few weeks?), in case we needed to revisit any part.

vrothberg (commented):
LGTM

Issue #751, opened 2019-11-14
API stability review


mtrmac:
WIP list of packages/interfaces to be kept / made private / changed.
- S = we can commit to keeping this **S**table
- I = this should become **I**nternal
- U = **U**nstable, do we somehow declare it as such or change it to be able to keep it stable?
- D = Marked as **D**eprecated. TO DO: Find these, and decide whether to keep them.
- ? = unreviewed, undecided, or may need changes before committing to it

X | Element | Comment
--|--|--
S | c/i | N/A, docs only
S | c/i/copy | This _must_ be stable, not sure about the current API though
? | c/i/directory | 
I | c/i/directory/explicitfilepath | 
? | c/i/docker | 
? | c/i/docker/archive | 
? | c/i/docker/daemon | 
I | c/i/docker/policyconfiguration | 
S | c/i/docker/reference | This must have stable semantics, has a long enough history, but do we need _all_ of it
? | c/i/docker/tarfile | Should be internal, but libpod needs a part
? | c/i/image | Careful!
? | c/i/manifest | Careful!
S | c/i/oci | N/A
? | c/i/oci/archive | 
? | c/i/oci/layout | 
? | c/i/openshift | Deprecated already
? | c/i/pkg/blobinfocache | 
? | c/i/pkg/blobinfocache/boltdb | 
? | c/i/pkg/blobinfocache/memory | 
? | c/i/pkg/blobinfocache/none | 
? | c/i/pkg/compression | Would be nice to make this internal, just export opaque types
? | c/i/pkg/compression/types | 
? | c/i/pkg/docker/config | 
S | c/i/pkg/strslice | 
? | c/i/pkg/sysregistriesv2 | 
? | c/i/pkg/tlsclientconfig | 
? | c/i/signature | 
? | c/i/storage | 
? | c/i/tarball | 
? | c/i/transports | 
? | c/i/transports/alltransports | Need this to be stable
? | c/i/types | Must provide _something_… but often opaque types would be better than full contents.
? | c/i/version | Need this to be stable.

---

Transport subpackage concerns:

For all the per-transport subpackages, we probably need to provide the `Transport` variable (but does that need to publicly support anything besides `Name()`?) Or maybe make public transport-specific reference (source? destination?) types so that callers that _must_ distinguish transports after-the-fact can do so using a type check instead of name comparison, and discourage `Name()` comparison.

[...]

PR #1439, opened 2022-01-11
WIP RFC: internal/external interface separation/adapters

mtrmac:
This is a RFC for a hopefully more manageable approach to evolving the internal interfaces, motivated by the progress discussion in #1438 .

I think this looks fairly clean, but boy it’s a lot of _toil_.

Major themes:
- Add `imagedestination.FromPublic`, which allows an internal caller to always use the fresh `internal/types` interface, instead of dealing with the compat versions. See how this benefits `copy`.
   - This might eventually turn into a next-generation public interface, now that `types.ImageDestination` is effectively frozen and deprecated: instead of having `internal/imagedestination.FromPublic()` returning an internal interface, we could have `stable/imagedestination.Destination()` returning a struct with newly-designed and newly-named public functions (e.g. using the "functional options" pattern, or anything else), while still letting us have an internal interface that changes over time. 
   - ⚠️ As is, this is not really _done_, notably because it’s not really clear how this wrapping this way is compatible with interface checks like `ImageDestinationPartial`. I guess we’ll want to fold `ImageDestinationPartial` into `ImageDestinationWithOptions` and use a `SupportsPutBlobPartial() bool`, but it’s not obviously the only possible approach. Maybe just robust self tests would be enough.
- Add `imagedestination/impl.AddCompat`, which allows an internal transport to implement the new `internal/types` interface, and rely on provided stubs for to still implement the compat version. See how this benefits `storage` (although that required a fair bit of code movement).
    - (Also extend this by making `internal/types` use `BlobInfoCache2` instead of the public frozen version; then `docker` can just assume that version, and delegate dealing with old callers to the generic infrastructure. Note that this has some run-time cost.)

@vrothberg RFC. There might be a much simpler design I’m missing.

I wanted to prototype this, using actually existing internal features and code, to see what it would take to have something similar for _new_ features, e.g. have `NewImageSource(…options{Progress:…})`, or adding progress bars to those `GetBlob`/`PutBlob` implementations that create extra copies/compress/decompress and the like. I think this is broadly viable and correct — but it’s also a fair amount of somewhat tedious work I don’t think I want to commit to do in a rush before Podman 4.0.

### DIFF
--- a/docker/docker_image_dest_test.go
+++ b/docker/docker_image_dest_test.go
@@ -6,9 +6,12 @@ import (
 	"net/http"
 	"testing"
 
+	"github.com/containers/image/v5/internal/private"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
+
+var _ private.ImageDestination = (*dockerImageDestination)(nil)
 
 func TestIsManifestInvalidError(t *testing.T) {
 	// Sadly only a smoke test; this really should record all known errors exactly as they happen.
--- a/internal/imagedestination/wrapper.go
+++ b/internal/imagedestination/wrapper.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 
+	"github.com/containers/image/v5/internal/blobinfocache"
 	"github.com/containers/image/v5/internal/private"
 	"github.com/containers/image/v5/types"
 )
@@ -53,7 +54,7 @@ func (w *wrapped) PutBlobWithOptions(ctx context.Context, stream io.Reader, inpu
 // It is available only if SupportsPutBlobPartial().
 // Even if SupportsPutBlobPartial() returns true, the call can fail, in which case the caller
 // should fall back to PutBlobWithOptions.
-func (w *wrapped) PutBlobPartial(ctx context.Context, chunkAccessor private.BlobChunkAccessor, srcInfo types.BlobInfo, cache types.BlobInfoCache) (types.BlobInfo, error) {
+func (w *wrapped) PutBlobPartial(ctx context.Context, chunkAccessor private.BlobChunkAccessor, srcInfo types.BlobInfo, cache blobinfocache.BlobInfoCache2) (types.BlobInfo, error) {
 	return types.BlobInfo{}, fmt.Errorf("internal error: PutBlobPartial is not supported by the %q transport", w.Reference().Transport().Name())
 }
 
--- a/docker/docker_image_dest.go
+++ b/docker/docker_image_dest.go
@@ -238,7 +238,7 @@ func (d *dockerImageDestination) SupportsPutBlobPartial() bool {
 // It is available only if SupportsPutBlobPartial().
 // Even if SupportsPutBlobPartial() returns true, the call can fail, in which case the caller
 // should fall back to PutBlobWithOptions.
-func (d *dockerImageDestination) PutBlobPartial(ctx context.Context, chunkAccessor private.BlobChunkAccessor, srcInfo types.BlobInfo, cache types.BlobInfoCache) (types.BlobInfo, error) {
+func (d *dockerImageDestination) PutBlobPartial(ctx context.Context, chunkAccessor private.BlobChunkAccessor, srcInfo types.BlobInfo, cache blobinfocache.BlobInfoCache2) (types.BlobInfo, error) {
 	return types.BlobInfo{}, fmt.Errorf("internal error: PutBlobPartial is not supported by the %q transport", d.Reference().Transport().Name())
 }
 
@@ -316,7 +316,7 @@ func (d *dockerImageDestination) mountBlob(ctx context.Context, srcRepo referenc
 // tryReusingExactBlob is a subset of TryReusingBlob which _only_ looks for exactly the specified
 // blob in the current repository, with no cross-repo reuse or mounting; cache may be updated, it is not read.
 // The caller must ensure info.Digest is set.
-func (d *dockerImageDestination) tryReusingExactBlob(ctx context.Context, info types.BlobInfo, cache types.BlobInfoCache) (bool, types.BlobInfo, error) {
+func (d *dockerImageDestination) tryReusingExactBlob(ctx context.Context, info types.BlobInfo, cache blobinfocache.BlobInfoCache2) (bool, types.BlobInfo, error) {
 	exists, size, err := d.blobExists(ctx, d.ref.ref, info.Digest, nil)
 	if err != nil {
 		return false, types.BlobInfo{}, err
@@ -350,8 +350,7 @@ func (d *dockerImageDestination) TryReusingBlobWithOptions(ctx context.Context, 
 	}
 
 	// Then try reusing blobs from other locations.
-	bic := blobinfocache.FromBlobInfoCache(options.Cache)
-	candidates := bic.CandidateLocations2(d.ref.Transport(), bicTransportScope(d.ref), info.Digest, options.CanSubstitute)
+	candidates := options.Cache.CandidateLocations2(d.ref.Transport(), bicTransportScope(d.ref), info.Digest, options.CanSubstitute)
 	for _, candidate := range candidates {
 		candidateRepo, err := parseBICLocationReference(candidate.Location)
 		if err != nil {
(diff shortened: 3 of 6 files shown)


mtrmac:
On second thought, we will need basically all of this, both for `ImageSource` and `ImageDestination`, for Cosign, to add support for the new signature format.

So solving this is, one way or another, is not _urgent_ but it’s going to have to happen soonish.

docker/docker_image_dest_test.go
```diff
@@ -6,10 +6,13 @@ import (
 	"net/http"
 	"testing"
 
+	internalTypes "github.com/containers/image/v5/internal/types"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
+var _ internalTypes.ImageDestinationWithOptions = &dockerImageDestination{}
```
mtrmac, line 14: Because of the type checks, we need to have some mechanisms ensure the conformances hold, but it would be nice to be able to do that without relying on manually writing tests like this. (I did break `ImageDestinationPartial` in one version of this PR already.)

Hopefully a new internal `Transport` (with `ParseInternalReference` returning an internal `Reference`, and `NewInternalSource` returning an internal `ImageSource`, will do that automatically, or something like that.

internal/imagedestination/wrapper.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+// PutBlobWithOptions writes contents of stream and returns data representing the result.
+// inputInfo.Digest can be optionally provided if known; if provided, and stream is read to the end without error, the digest MUST match the stream contents.
+// inputInfo.Size is the expected length of stream, if known.
+// inputInfo.MediaType describes the blob format, if known.
+// WARNING: The contents of stream are being verified on the fly.  Until stream.Read() returns io.EOF, the contents of the data SHOULD NOT be available
+// to any other readers for download using the supplied digest.
+// If stream.Read() at any time, ESPECIALLY at end of input, returns an error, PutBlob MUST 1) fail, and 2) delete any data stored so far.
+func (w *wrapped) PutBlobWithOptions(ctx context.Context, stream io.Reader, inputInfo publicTypes.BlobInfo, options types.PutBlobOptions) (publicTypes.BlobInfo, error) {
```
mtrmac, line 41: It would be interesting to have the internal names clean, without the `WithOptions` baggage. We can’t quite do that directly in the returned `ImageDestination` objects, and doing it here would require wrapping the `publicTypes.ImageDestination` in another non-zero-cost layer of indirection. Or, alternatively, we could have all of the internal code clean and work directly with internal objects and internal names, but always proxy external callers though the external types in separate objects (so that we can have direct access to internal `PutBlob(new)` and external callers get a proxy object with `PutBlob(old)`).

A cheap compromise would be to just name the internal methods `PutBlob2` or `PutBlobX` (with `X` reserved for all internal names), admittedly that’s also pretty ugly.

internal/types/types.go
```diff
@@ -5,58 +5,68 @@ import (
...
+	// If the blob has been successfully reused, returns (true, info, nil); info must contain at least a digest and size, and may
+	// include CompressionOperation and CompressionAlgorithm fields to indicate that a change to the compression type should be
+	// reflected in the manifest that will be written.
+	// If the transport can not reuse the requested blob, TryReusingBlob returns (false, {}, nil); it returns a non-nil error only on an unexpected failure.
+	TryReusingBlobWithOptions(ctx context.Context, info publicTypes.BlobInfo, options TryReusingBlobOptions) (bool, publicTypes.BlobInfo, error)
+}
+
 // ImageDestinationWithOptions is an internal extension to the ImageDestination
 // interface.
 type ImageDestinationWithOptions interface {
```
mtrmac, line 36: This should probably end up just `internalTypes.ImageDestination` (or maybe `…Destination2`/`DestinationX`??).

For internal callers, it is almost attractive to alias every public type through this package, so that internal callers only ever import one `types` package and we don’t have the `internalTypes`/`publicTypes` noise.

Or maybe even `c/image/internal/image.Destination`??? (OTOH there’s `c/image/image`)

There are way too many minimally-plausible ways to slice this. Luckily this is internal, so we don’t need to get this _perfect_ immediately.

copy/copy.go
```diff
@@ -1657,24 +1643,15 @@ func (c *copier) copyBlobFromStream(ctx context.Context, srcStream io.Reader, sr
...
-			IsConfig:   isConfig,
-			EmptyLayer: emptyLayer,
-		}
-		if !isConfig {
-			options.LayerIndex = &layerIndex
-		}
-		uploadedInfo, err = dest.PutBlobWithOptions(ctx, &errorAnnotationReader{destStream}, inputInfo, options)
-	} else {
-		uploadedInfo, err = c.dest.PutBlob(ctx, &errorAnnotationReader{destStream}, inputInfo, c.blobInfoCache, isConfig)
+	options := internalTypes.PutBlobOptions{
```
mtrmac, line 1646: It _almost_ seems clear that `copyBlobFromStream` should have a `PutBlobOptions` parameter, rather than individual ones; the large number of parameters to `copyBlobFromStream` are not nice at all.

Except for `Cache`, and for the fact that we might, in that case, want to pass `PutBlobOptions` by reference instead of by value.

Oh well, we can always introduce a single-purpose struct just for naming parameters to `copyBlobFromStream`. But that’s out of scope.

vrothberg (commented):
Very clever! I am cool with the changes but I also begin to wonder whether the added complexity of indirection outweighs the "benefits" of breaking the API and redesigning the API; just a thought/feeling.

mtrmac:
> I also begin to wonder whether the added complexity of indirection outweighs the "benefits" of breaking the API and redesigning the API; just a thought/feeling.

One possible answer to that is to make the external API purely a utility/wrapper that we mostly don’t care about (everything internal uses an internal API), e.g. use `internal/types` everywhere, exclusively. (This PR leans in the direction of internal-API-primacy, so far without trying to avoid the public API.) OTOH that carries a fair bit of risk of breaking external users and not noticing.

E.g. if we, for SIF, add some kind of progress facility to `NewImageSource`, a transport that uses that would not actually need to return/implement `publicTypes.ImageSource` _ at all_; we would just have an object that implements `privateTypes.ImageSource`, and callers that expect only that interface, in ideal theory nice and consistent and clean. (In practice actual backward/forward compatibility concerns and trade-offs _would_ actually have to manifest somewhere.)

There just aren’t any _great_ options.
- Just drop most of the interfaces, basically provide just `c/image/copy`. I don’t think a c/image that can’t implement `skopeo inspect` is very useful at all; also Podman needs various tools to expand/list/create multi-image archives and the like.
- Just don’t bother with API stability, work on features, and break the API. Because we have frozen on an API that isn’t easy to evolve, we would very likely keep bumping the major version _quite_ frequently, the way we went from v1 to v5 in the course of just 7 months! That’s painful for consumers, especially anything that passes a `types.ImageReference` around (the one pretty stable concept that we I publicly strongly recommend using instead of strings) across repos like c/common would need to coordinate major version bumps.
  - The same, but don’t bump the API major version. Broadly the same kind of mess with worse tooling.
- Spend extra time to design a new API, with little feedback or experience using/evolving it, and announce a v6. Fair bit of risk that we’ll not get that v6 _exactly_ right, and no-one has the time to just stop doing everything else on a rewrite.
- Freeze the problematic parts of the external API (most of it, arguably), add internal variants _as we need them for more features_, and eventually expose them publicly if there is a user (and only after we have some internal experience).
  - Have every consumer, and every transport, deal with the internal/external dichotomy individually.
  - Do something like this PR, a feature-less tedious work that hides that from most of the code.

Of all those options, it seems to me that something very broadly like this PR is the least bead way to keep moving forward. But I think all of them are sad, and I might very well be wrong about preferring this one.

Admittedly this PR does not _feel_ productive, compared to actual features, and yes, those two interface wrappers are tedious and not elegant. OTOH it was a prerequisite for improving the call stack and naming in `storageImageDestination`, and I’m quite pleased about that part, if I say so myself.

mtrmac:
A reminder from earlier #751 : `buildah/pkg/blobcache` is external to this package, and we need to handle it _somehow_.

Probably ideally to move that feature inside this repo.

Failing that, we would at least have a real-world test for all the compatibility infrastructure 😟

internal/types/types.go
mtrmac: It’s `private.ImageDestination` now.

mtrmac:
Superseded by #1589.

PR #1445, opened 2022-01-18
Log every credential path we consult

mtrmac:
... to make it easier for users to diagnose situations where no credentials are found.


vrothberg (commented):
LGTM

Looks like a network flake

rhatdan:
LGTM

PR #1446, opened 2022-01-19
GPGME: support passphrase for prompt-less signing

vrothberg:
To support signing images via gpgme without user prompt, allow for
providing a passphrase via the copy options.  Add a new *WithOptions API
to the `signature` package and extend its interface.

To prevent breaking the API, extend the signature API with an internal
type as has already been done for other types and interfaces in c/image.


vrothberg:
Skopeo PR: https://github.com/containers/skopeo/pull/1540

mtrmac (commented):
A process-wide environment variable shouldn’t be the only way to expose this (e.g. it is unusable from any multi-threaded server managing multiple private keys); a caller should be able to just pass a Go string to a `NewGPGSigningMechanism$differentiator` constructor. (Assuming $differentiator is `WithOptions`, should that just be a struct, or “functional options” with all that boilerplate? I don’t have a strong opinion at this point)

Similarly, up the stack, the passphrase should come from `copy.Options`.

Afterwards, _if_ passing this through the environment is a good idea (AFAIK it’s not _unsafe_ but it might be unergonomic — notably the way https://github.com/containers/skopeo/pull/1540 indiscriminately makes that available to all child processes immediately makes me pause), we can add a function to `pkg/cli/environment`.

signature/mechanism_gpgme.go
```diff
@@ -133,6 +134,30 @@ func (m *gpgmeSigningMechanism) Sign(input []byte, keyIdentity string) ([]byte,
...
+
+	if path := os.Getenv(PASSPHRASE_FILE_ENVIRONMENT); path != "" {
+		logrus.Debugf("Reading user-specified passphrase for gpgme signing from %s=%s", PASSPHRASE_FILE_ENVIRONMENT, path)
+		buf, err := ioutil.ReadFile(path)
+		if err != nil {
+			return nil, fmt.Errorf("reading passphrase file: %w", err)
+		}
+
+		// Callback to write the passphrase to the specified file descriptor.
+		callback := func(uidHint string, prevWasBad bool, f *os.File) error {
```
mtrmac, line 146: IIRC this should fail `if prevWasBad`.

signature/mechanism_gpgme.go
```diff
@@ -133,6 +134,30 @@ func (m *gpgmeSigningMechanism) Sign(input []byte, keyIdentity string) ([]byte,
...
+			buf = append(buf, '\n') // passphrase must terminate with a new line
```
mtrmac, line 147: (This would add one more newline per invocation, AFAICS.)

signature/mechanism_gpgme.go
```diff
@@ -133,6 +134,30 @@ func (m *gpgmeSigningMechanism) Sign(input []byte, keyIdentity string) ([]byte,
 	if err != nil {
 		return nil, err
 	}
+
+	if path := os.Getenv(PASSPHRASE_FILE_ENVIRONMENT); path != "" {
+		logrus.Debugf("Reading user-specified passphrase for gpgme signing from %s=%s", PASSPHRASE_FILE_ENVIRONMENT, path)
+		buf, err := ioutil.ReadFile(path)
```
mtrmac, line 140: Should this strip a trailing newline, so that `echo password > file` works? Only read the first line?

Probably the same behavior as `gpg2`.

mtrmac (commented):
See also the previous review — this should be configurable in Go.

signature/mechanism_gpgme.go
vrothberg: Very good though.  `gpg` reads the first line only, I'll change:

> --passphrase-file file
       Read  the  passphrase  from file file. Only the first line will be read from file file. This can only be used if only one
       passphrase is supplied. Obviously, a passphrase stored in a file is of questionable security if other users can read this
       file. Don't use this option if you can avoid it.

mtrmac: Yeah, it’s not particularly pretty, but
- Careful users can use a Go API
- Humans who are likely to store the passphrase in files are, I think, fairly likely to use the same file for `gig` and for this.

I wouldn’t object to this being _stricter_ than `gpg` (e.g. rejecting files that contain any non-empty lines after the first one); I do think it makes sense for that first line to be, if possible, interpreted the same.

vrothberg:
Good to go from my side.  Had to do the usual `*WithOptions` dance.

TomSweeneyRedHat (commented):
LGTM
Documentation targeted towards somewhere in Podman and/or Buildah?

vrothberg:
> LGTM Documentation targeted towards somewhere in Podman and/or Buildah?

Skopeo for now.  If needed, we can add the `--sign-passphrase-file` flag to Podman/Buildah as well.  Would require some plumbing in c/common/libimage.

signature/mechanism.go
```diff
@@ -28,6 +34,9 @@ type SigningMechanism interface {
 	// Sign creates a (non-detached) signature of input using keyIdentity.
 	// Fails with a SigningNotSupportedError if the mechanism does not support signing.
 	Sign(input []byte, keyIdentity string) ([]byte, error)
+	// Sign creates a (non-detached) signature of input using keyIdentity.
+	// Fails with a SigningNotSupportedError if the mechanism does not support signing.
+	SignWithOptions(input []byte, keyIdentity string, options *SignOptions) ([]byte, error)
```
mtrmac, line 39: This is an API break.

The simplest way to handle this would be to provide a single passphrase to `NewGPGSigningMechanism`; that’s somewhat inaccurate (in that a single `SigningMechanism` can be used to `Sign` with different `keyIdentity` values, i.e. different passphrases, but not _too_ likely to be a problem in practice.

Alternatively, provide a `map[keyID]string`, or a `func(keyId)string`, as a parameter to `NewGPGSigningMechanism`. That would actually be better in that the passphrase callback is a per-GPGME-context state, so we can’t have two at a time, and concurrent calls of `Sign` with different `keyIdentity` values would be, _in principle_ possible — but it would also be very noticeably worse in that the `keyIdentity` is _supposed_ to be a key ID but in practice shorter values and email addresses work right now, and none of that would work with a generic callback AFAICT.

Or, well, go all the way, define a `SigningMechanism2` (and make it private), and all that …

signature/mechanism_gpgme.go
```diff
@@ -133,12 +137,56 @@ func (m *gpgmeSigningMechanism) Sign(input []byte, keyIdentity string) ([]byte,
 	if err != nil {
 		return nil, err
 	}
+
+	passphraseFile := os.Getenv(PASSPHRASE_FILE_ENVIRONMENT)
```
mtrmac, line 141: I still wish we didn’t add any environment variable like that to the mandatory code path.

signature/mechanism_gpgme.go
```diff
@@ -133,12 +137,56 @@ func (m *gpgmeSigningMechanism) Sign(input []byte, keyIdentity string) ([]byte,
...
+		if err != nil {
+			return nil, err
+		}
+		defer ppf.Close()
+
+		// Read the *first* line in the passphrase file, just as gpg(1) does.
+		buf, err := bufio.NewReader(ppf).ReadBytes('\n')
+		if err != nil && !errors.Is(err, io.EOF) {
+			return nil, fmt.Errorf("reading passphrase file: %w", err)
+		}
```
mtrmac, line 157: This is not GPGME-specific, so it should probably be in the shared `mechanism.go` (and ideally have tests).

Or move it out of `signature` to `pkg/cli` altogether, and then I don’t care about the tests quite as much :)

signature/mechanism_gpgme.go
```diff
@@ -133,12 +137,56 @@ func (m *gpgmeSigningMechanism) Sign(input []byte, keyIdentity string) ([]byte,
...
+				return fmt.Errorf("bad passphrase")
+			}
+			_, err := gpgmeFD.Write(buf)
+			return err
+		}
+		if err := m.ctx.SetCallback(callback); err != nil {
+			return nil, fmt.Errorf("setting gpgme passphrase callback: %w", err)
+		}
+
+		// Loopback mode prevents gpgme from prompting and using the callback.
```
mtrmac, line 172: This reads “prevents … from … using the callback”, when we are doing it exactly to make the callback applicable.

signature/mechanism_gpgme.go
```diff
@@ -133,12 +137,56 @@ func (m *gpgmeSigningMechanism) Sign(input []byte, keyIdentity string) ([]byte,
...
+
+		// Callback to write the passphrase to the specified file descriptor.
+		callback := func(uidHint string, prevWasBad bool, gpgmeFD *os.File) error {
+			if prevWasBad {
+				return fmt.Errorf("bad passphrase")
+			}
+			_, err := gpgmeFD.Write(buf)
+			return err
+		}
+		if err := m.ctx.SetCallback(callback); err != nil {
```
mtrmac, line 168: For most of `signature`, code should have tests with pretty good coverage — at least for “correct passphrase” and “incorrect passphrase” I guess.

signature/signature.go
```diff
@@ -16,6 +16,9 @@ import (
 
 const (
 	signatureType = "atomic container signature"
+
+	// Path pointing to a passphrase file when signing via GPGME.  Specified as an evironment variable.
```
mtrmac, line 20: GPGME should not be a part of the API. “Name of an environment variable that contains …”.

vrothberg:
@mtrmac, PTanotherL:
* We're passing a passphrase only (no path) around
* `pkg/cli` received a convenience function for parsing a passphrase file
* added a new `signature/internal/signature` package to extend the interface without breaking the API

signature/internal/signature/mechanism.go
```diff
@@ -0,0 +1,34 @@
+package signature
```
mtrmac, line 1: This can be just `signature/internal`, and avoid the `internalSignature` naming.

---

… But, actually, that still makes the interface only reachable by the `c/image/signature` namespace (e.g. not from `c/image/copy`). So I don’t think we need an internal sub package at all — just name the private interface using a private symbol, e.g. `c/image/signature.signingMechanismWithPassphrase`.

IIRC that also makes the documentation better: a type alias in the public package to the internal package makes the interface _usable_ but `go doc` etc. can’t follow it and don’t show the reachable public methods. So keeping `c/image/signature.SigningMechanism` in place, and using a private interface in the same subpackage, would work a bit better.

OTOH, I could well accept the argument that we want to deprecate any external uses of `SigningMechanism` anyway, and hading the methods from documentation is actually a feature :) In that case an explicit comment that the interface is deprecated (like `types.BlobInfoCache`) might be useful.

signature/mechanism_openpgp.go
```diff
@@ -110,10 +110,16 @@ func (m *openpgpSigningMechanism) SupportsSigning() error {
 
 // Sign creates a (non-detached) signature of input using keyIdentity.
 // Fails with a SigningNotSupportedError if the mechanism does not support signing.
-func (m *openpgpSigningMechanism) Sign(input []byte, keyIdentity string) ([]byte, error) {
+func (m *openpgpSigningMechanism) SignWithOptions(input []byte, keyIdentity string, options *SignOptions) ([]byte, error) {
```
mtrmac, line 113: It’s `SignWithPassphrase` now.

Please make `newGPGSigningMechanismInDirectory` and `newEphemeralGPGSigningMechanism` (in both implementations) return the private interface, that will prevent such mistakes in the future.

signature/mechanism_gpgme.go
```diff
@@ -133,12 +134,41 @@ func (m *gpgmeSigningMechanism) Sign(input []byte, keyIdentity string) ([]byte,
 	if err != nil {
 		return nil, err
 	}
+
+	if passphrase != "" {
+		if !strings.HasSuffix(passphrase, "\n") { // passphrase needs to have a trailing new line
+			passphrase += "\n"
```
mtrmac, line 140: AFAICS  a `'\n'` inside the passphrase is always interpreted as a terminator inside the callback, so I’d rather explicitly reject any such input, instead of silently using a different passphrase value. If we ever implemented a Go-native mechanism, any hypothetical users passing a string with trailing `'\n'` would be broken.

signature/mechanism_gpgme.go
```diff
@@ -133,12 +134,41 @@ func (m *gpgmeSigningMechanism) Sign(input []byte, keyIdentity string) ([]byte,
...
+		}
+		// Callback to write the passphrase to the specified file descriptor.
+		callback := func(uidHint string, prevWasBad bool, gpgmeFD *os.File) error {
+			if prevWasBad {
+				return fmt.Errorf("bad passphrase")
```
mtrmac, line 145: (`fmt.Errorf(constant)` == `errors.New`. The caller ignores the specific value anyway, as long as it is not `nil`.)

signature/signature.go
```diff
@@ -190,12 +194,16 @@ func (s *untrustedSignature) strictUnmarshalJSON(data []byte) error {
...
-func (s untrustedSignature) sign(mech SigningMechanism, keyIdentity string) ([]byte, error) {
+func (s untrustedSignature) sign(mech SigningMechanism, keyIdentity string, passphrase string) ([]byte, error) {
 	json, err := json.Marshal(s)
 	if err != nil {
 		return nil, err
 	}
 
+	if newMech, ok := mech.(internalSignature.SigningMechanismWithPassphrase); ok {
+		return newMech.SignWithPassphrase(json, keyIdentity, passphrase)
+	}
```
mtrmac, line 205: This should fail if `passphrase` is supplied and the mechanism doesn’t implement that.

signature/signature.go
```diff
@@ -9,13 +9,17 @@ import (
...
 	"github.com/containers/image/v5/version"
 	digest "github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
 )
 
 const (
 	signatureType = "atomic container signature"
+
+	// Path pointing to a passphrase file when signing via GPGME.  Specified as an evironment variable.
+	PASSPHRASE_FILE_ENVIRONMENT = "CONTAINERS_SIGNATURE_PASSPHRASE_FILE"
```
mtrmac, line 22: Nothing uses this AFAICS. Add a helper to `pkg/cli/environment`, and move the constant there (or just don’t make it public)?

pkg/cli/parse.go
```diff
@@ -0,0 +1,35 @@
+package cli
```
mtrmac, line 1: (Absolutely non-blocking: Name this `passphrase.go`?)

mtrmac (commented):
ACK to the overall approach.

signature/mechanism_gpgme.go
mtrmac: I can’t see any added tests. (Probably add a passphrase-protected key to `GNUPGHOME=signature/fixtures`, and use it in `TestGPGSigningMechanismSign`, in addition to the existing unprotected key.)

signature/internal/signature/mechanism.go
vrothberg: Went with `c/image/signature.signingMechanismWithPassphrase`

signature/signature.go
vrothberg: It's not used/needed anymore, so we can safely remove it.

signature/docker.go
```diff
@@ -3,22 +3,45 @@
...
 		return nil, err
 	}
 	sig := newUntrustedSignature(manifestDigest, dockerReference)
-	return sig.sign(mech, keyIdentity)
+
+	var passphrase string
+	if options != nil {
+		passphrase = options.Passphrase
+		if strings.Contains(passphrase, "\n") {
+			return nil, errors.New("invalid passphrase: must not contain a line break")
```
mtrmac, line 34:
```suggestion
			// The gpgme implementation can’t use passphrase with \n ; reject it here for consistent behavior.
			return nil, errors.New("invalid passphrase: must not contain a line break")
```
or alternatively move the check into `mechanism_gpgme.go`.

mtrmac:
@vrothberg I’d really like to have a test for the `gpgmeSigningMechanism.SignWithPassphrase` code. Are you working on one, or should I take that on?

signature/docker_test.go
```diff
@@ -44,6 +44,42 @@ func TestSignDockerManifest(t *testing.T) {
...
+	require.NoError(t, err)
+
+	// Successful signing
+	signature, err := SignDockerManifestWithOptions(manifest, TestImageSignatureReference, mech, TestKeyFingerprintWithPassphrase, &SignOptions{Passphrase: TestPassphrase})
+	require.NoError(t, err)
+
+	verified, err := VerifyDockerManifestSignature(signature, manifest, TestImageSignatureReference, mech, TestKeyFingerprintWithPassphrase)
+	assert.NoError(t, err)
+	assert.Equal(t, TestImageSignatureReference, verified.DockerReference)
+	assert.Equal(t, TestImageManifestDigest, verified.DockerManifestDigest)
```
mtrmac, line 66: (If this is the only part that’s new, that could be added to the existing `TestSignDockerManifest`, [possibly renaming it to `TestSignDockerManifestWithOptions`], without copy&pasting the other parts.)

signature/docker_test.go
```diff
@@ -44,6 +44,42 @@ func TestSignDockerManifest(t *testing.T) {
...
+	}
+
+	manifest, err := ioutil.ReadFile("fixtures/image.manifest.json")
+	require.NoError(t, err)
+
+	// Successful signing
+	signature, err := SignDockerManifestWithOptions(manifest, TestImageSignatureReference, mech, TestKeyFingerprintWithPassphrase, &SignOptions{Passphrase: TestPassphrase})
+	require.NoError(t, err)
+
+	verified, err := VerifyDockerManifestSignature(signature, manifest, TestImageSignatureReference, mech, TestKeyFingerprintWithPassphrase)
```
mtrmac, line 63: One more test case for `VerifyDockerManifestSignature(…, "this passphrase is invalid")`, please.

vrothberg: Do you mean `SignDockerManifestWithOptions`?

mtrmac: Yup, I’m sorry.

signature/docker_test.go
```diff
@@ -44,6 +44,47 @@ func TestSignDockerManifest(t *testing.T) {
...
+
+	manifest, err := ioutil.ReadFile("fixtures/image.manifest.json")
+	require.NoError(t, err)
+
+	// Successful signing
+	signature, err := SignDockerManifestWithOptions(manifest, TestImageSignatureReference, mech, TestKeyFingerprintWithPassphrase, &SignOptions{Passphrase: TestPassphrase})
+	require.NoError(t, err)
+
+	// Invalid passphrase
+	_, err = SignDockerManifestWithOptions(manifest, TestImageSignatureReference, mech, TestKeyFingerprintWithPassphrase, &SignOptions{Passphrase: TestPassphrase + "\n"})
```
mtrmac, line 64: This exercises the “reject `\n`” logic, but not the failure path of the actual callback via gpgme.

signature/docker_test.go
```diff
@@ -44,6 +44,47 @@ func TestSignDockerManifest(t *testing.T) {
...
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "invalid passphrase")
+
+	verified, err := VerifyDockerManifestSignature(signature, manifest, TestImageSignatureReference, mech, TestKeyFingerprintWithPassphrase)
+	assert.NoError(t, err)
+	assert.Equal(t, TestImageSignatureReference, verified.DockerReference)
+	assert.Equal(t, TestImageManifestDigest, verified.DockerManifestDigest)
```
mtrmac, line 71: Please keep this together just after generating the successful signature, it’s still a part of validating the return value of that call.

signature/docker_test.go
```diff
@@ -2,12 +2,22 @@ package signature
...
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
+// Kill the running gpg-agent to drop cached keys. This allows for rerunning the tests with a cold cache.
+func killGPGAgent(t *testing.T) {
+	cmd := exec.Command("gpgconf", "--kill", "gpg-agent")
+	cmd.Env = append(os.Environ(), "GNUPGHOME=signature/fixtures")
```
mtrmac, line 16: Not `./fixtures` aka `testGPGHomeDirectory` ?

signature/docker_test.go
```diff
@@ -2,12 +2,22 @@ package signature
...
 	"io/ioutil"
+	"os"
+	"os/exec"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
+// Kill the running gpg-agent to drop cached keys. This allows for rerunning the tests with a cold cache.
```
mtrmac, line 13:
```suggestion
// Kill the running gpg-agent to drop unlocked keys. This allows for testing handling of invalid passphrases.
```

signature/docker_test.go
```diff
@@ -44,6 +54,57 @@ func TestSignDockerManifest(t *testing.T) {
 	assert.Error(t, err)
 }
 
+func TestSignDockerManifestWithPassphrase(t *testing.T) {
+	defer killGPGAgent(t)
```
mtrmac, line 58: Shouldn’t the test just do it at the _start_? That way it would work from a known state.

signature/docker_test.go
```diff
@@ -44,6 +54,57 @@ func TestSignDockerManifest(t *testing.T) {
...
+	require.Error(t, err)
+
+	// Successful signing
+	signature, err := SignDockerManifestWithOptions(manifest, TestImageSignatureReference, mech, TestKeyFingerprintWithPassphrase, &SignOptions{Passphrase: TestPassphrase})
+	require.NoError(t, err)
+
+	verified, err := VerifyDockerManifestSignature(signature, manifest, TestImageSignatureReference, mech, TestKeyFingerprintWithPassphrase)
+	assert.NoError(t, err)
+	assert.Equal(t, TestImageSignatureReference, verified.DockerReference)
+	assert.Equal(t, TestImageManifestDigest, verified.DockerManifestDigest)
```
mtrmac, line 91: Non-blocking: https://github.com/containers/image/pull/1446#discussion_r791653760

signature/docker_test.go
mtrmac: Needs to be `./fixtures` for `go test ./signature` to work.

signature/docker_test.go
vrothberg: Sure, that sounds much cleaner.

signature/docker_test.go
vrothberg: Left untouched; I prefer the symmetry of the two tests but don't feel strongly.

mtrmac: ACK.

mtrmac (approved):

PR #1461, opened 2022-02-09
docker, BlobInfoCache: try to reuse compressed blobs from `all known` compressed digests

flouthoc:
It seems we try to reuse blobs only for the specified registry, however
we can have valid known compressed digests across registry as well
following pr attempts to use that by doing following steps.

* Move all known blobs into common data scope i.e `all` instead of
  storing them in seperate registry.

* Increase the sample set of potential blob reuse to all known
  compressed digests , also involving the one which do not belong to
current registry.

* If a blob is found match it against the registry where we are
  attempting to push. If blob is already there consider it a `CACHE
HIT!` and reply skipping blob, since its already there.

#### Pre-steps

* Remove all images `buildah rmi --all` // needed so all new pulled blobs can
  be tagged again in common bucket
* Remove any previous `blob-info-cache` by

```console
rm /home/<user>/.local/share/containers/cache/blob-info-cache-v1.boltdb
```
### Actual use-case

```console
$ skopeo copy docker://registry.fedoraproject.org/fedora-minimal docker://quay.io/fl/test:some-tag
$ buildah pull registry.fedoraproject.org/fedora-minimal
$ buildah tag registry.fedoraproject.org/fedora-minimal quay.io/fl/test
$ buildah push quay.io/fl/test
```
#### Expected Output

```console
Copying blob a3497ca15bbf skipped: already exists
Copying config f7e02de757 done
Storing signatures
```

### DIFF
--- a/docker/cache.go
+++ b/docker/cache.go
@@ -7,8 +7,9 @@ import (
 
 // bicTransportScope returns a BICTransportScope appropriate for ref.
 func bicTransportScope(ref dockerReference) types.BICTransportScope {
-	// Blobs can be reused across the whole registry.
-	return types.BICTransportScope{Opaque: reference.Domain(ref.ref)}
+	// Blobs can be reused across multiple registries
+	// therefore bucket all blobs in same scope
+	return types.BICTransportScope{Opaque: "all"}
 }
 
 // newBICLocationReference returns a BICLocationReference appropriate for ref.
--- a/docker/docker_image_dest.go
+++ b/docker/docker_image_dest.go
@@ -336,6 +336,7 @@ func (d *dockerImageDestination) TryReusingBlob(ctx context.Context, info types.
 	bic := blobinfocache.FromBlobInfoCache(cache)
 	candidates := bic.CandidateLocations2(d.ref.Transport(), bicTransportScope(d.ref), info.Digest, canSubstitute)
 	for _, candidate := range candidates {
+		crossRegistry := false
 		candidateRepo, err := parseBICLocationReference(candidate.Location)
 		if err != nil {
 			logrus.Debugf("Error parsing BlobInfoCache location reference: %s", err)
@@ -350,7 +351,8 @@ func (d *dockerImageDestination) TryReusingBlob(ctx context.Context, info types.
 		// Sanity checks:
 		if reference.Domain(candidateRepo) != reference.Domain(d.ref.ref) {
 			logrus.Debugf("... Internal error: domain %s does not match destination %s", reference.Domain(candidateRepo), reference.Domain(d.ref.ref))
-			continue
+			// most likely this blob is on a different registry
+			crossRegistry = true
 		}
 		if candidateRepo.Name() == d.ref.ref.Name() && candidate.Digest == info.Digest {
 			logrus.Debug("... Already tried the primary destination")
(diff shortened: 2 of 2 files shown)


flouthoc:
@mtrmac @nalind @vrothberg PTAL. Following PR is in draft since i did not add any tests but will add tests after initial review and discussion.

docker/cache.go
```diff
@@ -7,8 +7,9 @@ import (
 
 // bicTransportScope returns a BICTransportScope appropriate for ref.
 func bicTransportScope(ref dockerReference) types.BICTransportScope {
-	// Blobs can be reused across the whole registry.
-	return types.BICTransportScope{Opaque: reference.Domain(ref.ref)}
+	// Blobs can be reused across multiple registries
+	// therefore bucket all blobs in same scope
+	return types.BICTransportScope{Opaque: "all"}
```
mtrmac, line 12: This looses most of the nuance/intelligence. We do want `CandidateLocations2` to primarily try to use blobs that are known to exist on that registry, not just recently-used digests — _only if_ we have no candidates at that location, blindly try the compressed version (and only a few such digests — compare `prioritize.replacementAttempts`).

docker/docker_image_dest.go
```diff
@@ -361,6 +363,13 @@ func (d *dockerImageDestination) TryReusingBlob(ctx context.Context, info types.
 
 		// Checking candidateRepo, and mounting from it, requires an
 		// expanded token scope.
+
+		if crossRegistry {
+			// found following blob in different registry but we need to check blob presence against the registry
+			// where we are planning to push, hence switch back the candidate repo to the one where we are planning to push
+			candidateRepo, _ = parseBICLocationReference(types.BICLocationReference{Opaque: string(d.ref.ref.Name())})
```
mtrmac, line 370: (This could just be `candidateRepo = d.ref.ref.Name()` — and it should have happened instead of the `crossRegistry = true` assignment already, so that the “Already tried the primary destination” check applies to the cross-registry situation as well.)

mtrmac:
FWIW we differentiate between “blob info cache” (`c/image/internal/blobinfocache`, metadata-only, always used) and “blob cache” (`buildah/pkg/blobcache`, includes blob data, strictly opt-in).

docker/cache.go
flouthoc: One issue is that we will perform lookup twice one iteration for `known` blobs via `CandidateLocations2` and another if we create a new function to perform lookup on all available `blobs`. So there is a overlap of while searching for `blobs` of course we could minimize that by some checks.

So i agree we can split the matching of `all` away from `CandidateLocations2` so original logic remains intact. Let try something around this.

docker/docker_image_dest.go
flouthoc: ackd.

docker/cache.go
mtrmac: `CandidateLocations2` is an internal interface, we can reshape it any way we like.

flouthoc: @mtrmac just re-visiting this since BZ showed interest in this RFE again, do you like the idea if we modify `CandidateLocations2` to return `all` blobs but give these newly added blobs will have least priority so maybe some modifications in `DestructivelyPrioritizeReplacementCandidates`.

mtrmac: - I agree with the lowest priority part for “location-unknown” candidates. (We should also limit their number.)
- `Opaque` does mean `Opaque` to the generic code: that field is exclusively up to the transport to define and the generic cache code has no business checking for specific string values. In the returned `BICReplacementCandidate2`, introduce a new way to represent “location unknown” values (an extra bool or turn `Location` into a pointer).

https://github.com/containers/image/pull/550#pullrequestreview-579816214 is an old description how I imagine this could work — now I don’t think we need a separate `CandidateDigests`, just let `CandidateLocations2` include the location-unknown results. But that’s just a guess and the structure might end up different based on the final code.

flouthoc: @mtrmac Got it i have a patch ready, do you prefer if i open a new PR cause i think it might have different review comments altogether and close this PR ? But i can also force push on this PR if that is preferred.

flouthoc:
Closing this in favor of https://github.com/containers/image/pull/1645

PR #1462, opened 2022-02-11
Establish internal/private.ImageDestination

mtrmac:
Based on the #1439 experiment, this is trying to establish the common type, so that we can then separately work on the client-side wrapper, and the implementation-side compat/helper infrastructure.

This includes some commits from #1439; the new part is a proposal to, instead of using `publicTypes "…/v5/types" and `internalTypes "…/v5/internal/types`, with `internalTypes.ImageDestinationWithOptions`, to use `types "…/v5/types"` (as we used to) and `private "…/v5/internal/private"`, with `private.ImageDestination`.

The “private” name, as well as all other aspects, is certainly open for debate.

@vrothberg WDYT?


vrothberg (approved):
LGTM, let's do this :scream:

rhatdan:
LGTM

PR #1463, opened 2022-02-12
Wrap public transport implementations to consistently use a private API

mtrmac:
Another part of the #1439 concept.

- Add `private.ImageSource`, similar to the existing `private.ImageDestination`.
- Introduce `internal/imagesource.FromPublic` and `internal/imagedestination.FromPublic`; the copy pipeline can now always use the private API without cluttering the code with compatibility paths.
- Replace the use of optional `ImageSourceSeekable` and `PutBlobPartial` interfaces with mandatory methods on the private interfaces, along with "Supports…” methods. For now, this is more wordy and risks inconsistency between the supports flags and actual implementations, but we’ll eventually make it cleaner and type-safe again, by providing mix-ins (so that a transport that doesn’t support a feature, or does support a feature, needs just to add a mix-in).

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -1272,8 +1272,8 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, to
 	// of the source file are not known yet and must be fetched.
 	// Attempt a partial only when the source allows to retrieve a blob partially and
 	// the destination has support for it.
-	imgSource, okSource := ic.c.rawSource.(private.BlobChunkAccessor)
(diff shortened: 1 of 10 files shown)


vrothberg (commented):
LGTM, really nice work, @mtrmac

vrothberg (approved):

vrothberg:
@giuseppe PTAL

mtrmac:
BTW I’d like to add #1459 next, before converting transports to private-only implementations.

giuseppe (approved):
LGTM

PR #1465, opened 2022-02-12
Eliminate a goroutine

mtrmac:
Have `imageSourceSeekableProxy` directly update the progress bar, instead of sending the data to a channel and updating the progress bar in a separate goroutine.

In theory, this might slow down `GetBlobAt` a bit, because it will block on a progress bar update (but not on a repaint IIRC). We already pay the same price for `bar.ProxyReader()` on ordinary copies, so it's almost certainly not worth worrying about, and worth the simplification.


mtrmac:
Rebased.

vrothberg (approved):

Issue #1139, opened 2021-02-04
http.Transport configuration enforces http1.1 and causes issues with istio


jhart1685:
This code https://github.com/containers/image/blob/master/pkg/tlsclientconfig/tlsclientconfig.go#L100-L105 

Creates an http transport with a custom dialler, which means that by default, Go will not allow the connection to be upgraded to http2 (this is explained in the documentation of the [ForceAttemptHTTP2](https://github.com/golang/go/blob/master/src/net/http/transport.go#L277-L282) option).

This, coupled with the `DisableKeepAlives: true` option, means that the resulting connection is susceptible to being chopped quickly by the server.
See https://github.com/istio/istio/issues/30577 for background of this.. 
But essentially, if you are doing (for example) a skopeo copy of an image between two registries, and the source registry is hosted behind an istio ingress, the image read is likely to get terminated early with an EOF. 

I've seen other issues in this repo around EOFs, improving diagnostics and logging of them for example.
But the frequency of these EOFs can be much reduced by implementing one of the following options:
- Allowing keepalives on the http(1.1) client connection.
- Allowing the connection to be upgraded to http2 , either by not using a custom dialler, or by setting the `ForceAttemptHTTP2` option.

vrothberg:
Thanks for reaching out and the great summary, @jhart1685!

I browsed a bit around and had a look. Docker is also disabling the keep alives while containerd does not. I need to look a bit closer to check whether changing that in c/image would have other implications.

@mtrmac @rhatdan WDYT?

mtrmac:
I’ll try to check in more detail later — just as a very general point, the `pkg/tlsclientconfig` was created for compatibility, but we mostly have neither the interest nor the expertise to fine-tune TCP and similar low-level layers, and especially to keep up with the changing best practices.

We probably can’t just blindly drop everything in there; overall, it does seem attractive to decrease the deviation from defaults when we don’t have a specific reason to keep deviating.

jhart1685:
Hi @vrothberg @mtrmac - Thanks for looking. I take your point about fine-tuning TCP not being the core function of this project. Likewise, it's not my area of expertise either.

But in this case, we've tracked down one of the causes of the EOFs that make Skopeo unusable in certain situations. It looks like a relatively small modification here (albeit one still to be made carefully and with due diligence wrt testing etc.) 
could make the client much more tolerant of different server configurations, and provide a big win in terms of reliability.

mtrmac:
[...]

PR #1469, opened 2022-02-15
Use the default KeepAlive interval (15s) instead of 30s

nalind:
Remove our overriding of the default KeepAlive interval.  We previously set it to 30 seconds, but now we'll leave it unset to get the default, which has been 15 seconds since Go 1.12.

This should fix https://bugzilla.redhat.com/show_bug.cgi?id=2052672.

### DIFF
--- a/pkg/tlsclientconfig/tlsclientconfig.go
+++ b/pkg/tlsclientconfig/tlsclientconfig.go
@@ -94,7 +94,6 @@ func hasFile(files []os.FileInfo, name string) bool {
 func NewTransport() *http.Transport {
 	direct := &net.Dialer{
 		Timeout:   30 * time.Second,
-		KeepAlive: 30 * time.Second,
 		DualStack: true,
 	}
 	tr := &http.Transport{


rhatdan:
LGTM
@mtrmac @vrothberg PTAL

mtrmac:
(I mean, https://datatracker.ietf.org/doc/html/rfc1122#section-4.2.3.6 is still a valid standard. “MUST default to no less than two hours.”, not that 30-second or a 15-second default makes that much of a difference in this respect. \*grumble\*)

---

What’s actually happening here? Somehow the CPU load to decompress gzip, and local disk, is so slow, that there’s a 30-second stall with no network activity at all? I mean, that’s not impossible but how would a 30-second burst of decompress+write happen with the default buffer sizes on the order of dozens/hundreds of kilobytes?

```
DEBU[0040] docker-daemon: Closing tar stream to abort loading
```
What’s that about? That’s not a pull at all, and AFAICS not relevant to Buildah or CRI-O.

I could well imagine a 30-second stall on e.g. a registry/registry copy, where one of the registries is connected using a _significantly_ slower network (compare e.g. https://github.com/containers/image/issues/1139 and especially https://github.com/containers/image/issues/1139#issuecomment-774284293 ).

mtrmac:
To be clear, I’m not at all saying that 30 seconds is a _good_ value that needs to stay, and this PR might well be the right fix.

It’s just that the data I’ve seen so far, at a _very_ quick skim (I’ve taken maybe one minute to look at the TCP trace), don’t look at all definitive. E.g. two of the tickets contain rather different `reset by peer` error messages, another doesn’t contain the word “reset” at all; there isn’t any specific command/reproducer, at best an error message without any context. I might well have missed something, and I’m afraid I won’t be able to dig deep into this today (nor, likely, any time this week).

> That error would happen while attempting to copy an image to a running dockerd, right?

Yes — and IIRC the last time we did that was in `/usr/bin/atomic pull`. `projectatomic/docker` would never do that.

vrothberg (commented):
LGTM
The theory makes sense to me. Any chance we can provide a test build/RPM/binary before doing the backporting dance? Just be sure.

mtrmac:
I really don’t understand the situation and I don’t want to hold up a fix, so if y’all are confident about it, feel free to merge, definitely don’t wait for me.

---

The linked S3 article says that connections _idle_ for 20 seconds will be closed. It’s not obvious to me that TCP-level keepalives make the connection count as active (“was not read from or written to within the timeout period”). (BTW it also _only_ talks about uploads, which is not the case here at all.)

---

One more completely half-baked theory:
```[TCP Keep-Alive] 46988 → 443 [ACK] Seq=1059 Ack=9560374 Win=3144704 Len=0```

Doesn’t the large `Win=` value mean that the recipient is _not_ struggling to keep up with the data? The connection seems to be just idle (to be fair, we don’t see the sequence numbers from the last data packet, and we don‘t see the sender’s view of the situation[1]). The previous incoming data packets are, also, _always_, not full-sized. That could be a random occurrence (some buffer in the sender not a multiple of the MTU), or it could be the end of the data.

Both in the  `docker-daemon` and c/storage cases, that’s somewhat consistent with a fast copy of the data to a temporary file, and then `PutBlob` spends a fair bit of time doing something _else_ (c/storage extracting and committing a layer, or docker-daemon sending the data over to the daemon, which is doing something similar). During all of that time, the source stream is _open_ and idle (we only close it after `PutBlob` returns).

So in these situations, I would not at all be surprised by a connection idle for tens of seconds, and I wouldn’t be surprised by that connection being reset by an aggressive server for being idle (well, sending keep-alive packets might actually _trigger_ a visible RST, that’s also consistent with the trace).

It’s completely unclear to me how that would result in an user-observable failure at all (we don’t check the error from `srcStream.Close()` at all, and `PutBlob` would be in the idle local-processing phase only after successfully observing an EOF (that might be a TLS application-side EOF, not visible in a TCP trace, but still), at which point `PutBlob` has no reason to continue reading or to see any “connection reset” state — but if that _did_ result in an user-observable failure, the fix would be to allow for the connection to be terminated (or actively terminate it) sooner, not to keep it idle and alive.

---

[1] Alternatively, the sender has sent a fair bit of data that was not just lost due to a bad link, and … didn’t retransmit at all, as is the sender’s responsibility? Gave up retransmitting? It’s not obvious that sending a keep-alive ping helps here, unless that somehow fixes the link, or somehow triggers the sender to retransmit in time, when it otherwise wouldn’t. This might well be material… it’s just that I can’t see much _evidence_ for that.

mtrmac:
FWIW:
> We previously set it to 30 seconds, but now we'll leave it unset to get the default, which has been 15 seconds since Go 1.12.

That’s correct for `net.Dialer` (https://github.com/golang/go/commit/5bd7e9c54f946eec95d32762e7e9e1222504bfc1 ), but a more relevant comparison might be `http.DefaultTransport`, where the default has been 30 seconds since Go 1.3 (https://github.com/golang/go/commit/49beb23ba73eab22c470e68cac87d51a0d12d97b ). I can’t find any detailed rationale for the 30 value; for 15, the conversation around https://github.com/golang/go/issues/23459#issuecomment-374777402 basically says “we don’t have a precise target and we don’t care to that detail, preferring a simple implementation instead”.

Now, that’s not at all an argument for hard-coding either value; I suppose we should be doing something like  `http.DefaultTransport.Clone`, and change only the TLS configuration we have a good reason to change.

---

Ultimately I think the real change will be the elsewhere-requested change to the pull process, so that unexpected EOFs (after pulling some unspecified sufficient amount of data?) are transparently resumed, on registries that support range requests.

mtrmac:
Per internal support tickets:

> In Feb 2022, $REDACTED reported that they were no longer experiencing the issue, but no further details were provided on what might have changed to resolve it.

So, closing; please reopen if this is still relevant.

PR #1476, opened 2022-02-23
docker/reference: add IsFullIdentifier

vrothberg:
containers/common/libimage is in need to check whether a given input
looks like a full image ID (i.e., a 64-byte hex string).  Since
compiling regexes has a negative impact on init- and run-time, let's
expose a function to check whether a given string matches the already
compiled regex in docker/reference.

@mtrmac PTAL
I want it for optimizing the fixes for https://github.com/containers/podman/issues/12761.

### DIFF
--- a/docker/reference/regexp-additions.go
+++ b/docker/reference/regexp-additions.go
@@ -1,0 +1,6 @@
+package reference
+
+// Return true if the specified string fully matches `IdentifierRegexp`.
+func IsFullIdentifier(s string) bool {
+	return anchoredIdentifierRegexp.MatchString(s)
+}
--- a/docker/reference/regexp_test.go
+++ b/docker/reference/regexp_test.go
@@ -545,6 +545,9 @@ func TestIdentifierRegexp(t *testing.T) {
 
 	for i := range fullCases {
 		checkRegexp(t, anchoredIdentifierRegexp, fullCases[i])
+		if IsFullIdentifier(fullCases[i].input) != fullCases[i].match {
+			t.Errorf("Expected match for %q to be %v", fullCases[i].input, fullCases[i].match)
+		}
 	}
 
 	for i := range shortCases {


mtrmac (commented):
I realize I’m probably being a pest, still… what’s the full story here?

(_Conceptually_, I’d expect a “parse input of $syntax” function that is the ~authoritative implementation of $syntax; not an API of “try parsing the input as $syntax1; if it failed, try $syntax2; if it failed, try $syntax3”. This new function could well be a part of the first “authoritative parser” implementation, it just raises my suspicion :) )

Specifically WRT https://github.com/containers/podman/issues/12761 , compare https://github.com/containers/image/blob/6ab45a64f62a13c19c748fe02d007cf3b621204c/storage/storage_transport.go#L145 and https://github.com/containers/image/blob/6ab45a64f62a13c19c748fe02d007cf3b621204c/storage/storage_transport.go#L160 enforcing a `minimumTruncatedIDLength` (OTOH the contents check is fairly weak, and using a stricter regex might be valuable). That’s, AFAICT, not the case in `lookupImageInLocalStorage`, where `store.Image()` accepts an unambiguous prefix of any length.

I’m also not sure how detecting a _full_ identifier changes handling of _short_ identifiers, perhaps that’s only tangentially related?

Is this just a locally-scoped change to modify the one
```go
                if strings.HasPrefix(name, "sha256:") || (len(name) == 64 && !strings.ContainsAny(name, "/.:@")) {
```
check in `Runtime.Pull`? Consider `ParseAnyReference` — that would work fine for the ID syntaxes, OTOH the `Named` returned would not be usable, so that’s probably not better.

(Completely unrelated; The `HasPrefix(name, "sha256:")` check seems to be always false.)

---

Implementation looks OK, of course. From a maintenance POV, I’d prefer adding it to a separate file (`regexp-additions.go`)? so that the correspondence with distribution/distribution/reference is clear and easier to maintain — OTOH the test is really much better when integrated into the existing one, and placing the test and the function differently is not all that attractive.

vrothberg:
> I realize I’m probably being a pest, still… what’s the full story here?

:rofl: 

> I’m also not sure how detecting a full identifier changes handling of short identifiers, perhaps that’s only tangentially related?

I want it for an optimization. https://github.com/containers/podman/issues/12761 happens because we're doing a `lookupImageInLocalStorage()` with the input _before_ doing the short-name expansion.  I want to keep looking up an image with it's full ID as fast as possible (i.e., immediately) without it being subject to short-name expansion with N look ups. In order to do that, I need check whether the input is a 64-byte hex string.

Here's the PR for c/common that should make my intention clearer: https://github.com/containers/common/pull/939

vrothberg:
There's another issue with tagging that I noticed:
`podman tag $image $64-byte-hex` does not error out but should -.-

mtrmac:
> [containers/podman#12761](https://github.com/containers/podman/issues/12761) happens because we're doing a `lookupImageInLocalStorage()` with the input _before_ doing the short-name expansion. I want to keep looking up an image with it's full ID as fast as possible (i.e., immediately) without it being subject to short-name expansion with N look ups. In order to do that, I need check whether the input is a 64-byte hex string.

OK, that’s actually getting us closer to a consistent parser — where the 64-byte hex strings are only ever interpreted one way. Nice.

LGTM; consider moving it to a separate file, but that’s quite a weak preference.

vrothberg:
> LGTM; consider moving it to a separate file, but that’s quite a weak preference.

Thanks! Done :heavy_check_mark:

mtrmac (approved):

mtrmac:
Looks like CI is frequently broken now…

vrothberg:
> Looks like CI is frequently broken now…

> --- FAIL: TestReadKeyring (0.00s)
    keyring_test.go:185: expected to read 1 userkeyring, but get 2

I saw this one a number of times this week.

mtrmac:
> Looks like CI is frequently broken now…

https://github.com/containers/image/pull/1477 , for the record.

vrothberg:
green

TomSweeneyRedHat:
@vrothberg you may need to rebase
changes LGTM otherwise

PR #1477, opened 2022-02-23
Remove keyring support

mtrmac:
The package is not used, no-one is currently working on wiring it up, and it started failing in CI (apparently because it assumes the user's keyring is empty when tests start, which is now not reliably the case).

Instead of fixing the tests, just remove it; we can always restore the code from Git history.


vrothberg (commented):
LGTM

> +0 −614 

I love such commits ... less work is good work :^)

rhatdan:
LGTM

PR #1479, opened 2022-02-24
Remove a direct dependency on golang.org/x/sys

mtrmac:
After https://github.com/containers/image/pull/1477 has removed the last direct user of the package, update also the dependency.

Don't completely remove it so that updating c/image can't cause a downgrade of `golang.org/x/sys` in consumers;
eventually some dependency will start requiring a newer version of `golang.org/x/sys` and the entry will completely
drop out (or we eventually update to the newer `go.mod` format that lists indirect dependencies as well).


vrothberg (commented):
Good catch, LGTM

TomSweeneyRedHat:
LGTM

PR #1480, opened 2022-02-24
storage: use race-free `AddNames` instead of `SetNames`

flouthoc:
Commits from parallel builds using `SetNames` removes `names` from
storage for other builds.

Use race-free atomic `AddNames` to prevent breaking of parallel builds.

Uses `c/storage` from PR: https://github.com/containers/storage/pull/1153

Fixes issue: 

**Builds fail with image not known.**
### Reproducer
```console
#!/bin/bash
x=1
sudo ./podman rmi -af
rm -f log*.*
sudo ./podman build -t first . &> logfirst.log
while [ $x -le 30 ]
do
  echo "$x times"
  sudo ./podman build --log-level debug -t $x . &> log$x.log &
  x=$(( $x + 1 ))
#  sleep 1
done
```

### Dockerfile
```console
FROM quay.io/jitesoft/alpine
```

PS: I am not sure how to ensure parallel builds in CI and produce a race so not sure how to test this in CI.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/containerd/containerd v1.5.9 // indirect
 	github.com/containers/libtrust v0.0.0-20200511145503-9c3a6c22cd9a
 	github.com/containers/ocicrypt v1.1.2
-	github.com/containers/storage v1.38.2
+	github.com/containers/storage v1.38.3-0.20220301151551-d06b0f81c0aa
 	github.com/docker/distribution v2.8.0+incompatible
 	github.com/docker/docker v20.10.12+incompatible
 	github.com/docker/docker-credential-helpers v0.6.4
--- a/storage/storage_image.go
+++ b/storage/storage_image.go
@@ -1197,21 +1197,13 @@ func (s *storageImageDestination) Commit(ctx context.Context, unparsedToplevel t
 		}
 		logrus.Debugf("saved image metadata %q", string(metadata))
 	}
-	// Set the reference's name on the image.  We don't need to worry about avoiding duplicate
-	// values because SetNames() will deduplicate the list that we pass to it.
-	if name := s.imageRef.DockerReference(); len(oldNames) > 0 || name != nil {
-		names := []string{}
-		if name != nil {
-			names = append(names, name.String())
+	// Adds the reference's name on the image.  We don't need to worry about avoiding duplicate
+	// values because AddNames() will deduplicate the list that we pass to it.
+	if name := s.imageRef.DockerReference(); name != nil {
+		if err := s.imageRef.transport.store.AddNames(img.ID, []string{name.String()}); err != nil {
+			return errors.Wrapf(err, "adding names %v to image %q", name, img.ID)
 		}
-		if len(oldNames) > 0 {
-			names = append(names, oldNames...)
-		}
-		if err := s.imageRef.transport.store.SetNames(img.ID, names); err != nil {
-			logrus.Debugf("error setting names %v on image %q: %v", names, img.ID, err)
-			return errors.Wrapf(err, "setting names %v on image %q", names, img.ID)
-		}
-		logrus.Debugf("set names of image %q to %v", img.ID, names)
+		logrus.Debugf("added name %q to image %q", name, img.ID)
 	}
 
 	commitSucceeded = true
--- a/go.sum
+++ b/go.sum
@@ -78,6 +78,8 @@ github.com/Microsoft/go-winio v0.4.17/go.mod h1:JPGBdM1cNvN/6ISo+n8V5iA4v8pBzdOp
 github.com/Microsoft/go-winio v0.5.0/go.mod h1:JPGBdM1cNvN/6ISo+n8V5iA4v8pBzdOpzfwIujj1a84=
 github.com/Microsoft/go-winio v0.5.1 h1:aPJp2QD7OOrhO5tQXqQoGSJc+DjDtWTGLOmNyAm6FgY=
 github.com/Microsoft/go-winio v0.5.1/go.mod h1:JPGBdM1cNvN/6ISo+n8V5iA4v8pBzdOpzfwIujj1a84=
+github.com/Microsoft/go-winio v0.5.2 h1:a9IhgEQBCUEk6QCdml9CiJGhAws+YwffDHEMp1VMrpA=
+github.com/Microsoft/go-winio v0.5.2/go.mod h1:WpS1mjBmmwHBEWmogvA2mj8546UReBk4v8QkMxJ6pZY=
 github.com/Microsoft/hcsshim v0.8.6/go.mod h1:Op3hHsoHPAvb6lceZHDtd9OkTew38wNoXnJs8iY7rUg=
 github.com/Microsoft/hcsshim v0.8.7-0.20190325164909-8abdbb8205e4/go.mod h1:Op3hHsoHPAvb6lceZHDtd9OkTew38wNoXnJs8iY7rUg=
 github.com/Microsoft/hcsshim v0.8.7/go.mod h1:OHd7sQqRFrYd3RmSgbgji+ctCwkbq2wbEYNSzOYtcBQ=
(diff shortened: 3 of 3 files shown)


flouthoc:
I have to `point` to actual `c/storage` version right now its my fork.

storage/storage_image.go
```diff
@@ -1207,7 +1207,7 @@ func (s *storageImageDestination) Commit(ctx context.Context, unparsedToplevel t
 		if len(oldNames) > 0 {
```
mtrmac, line 1207: The `oldNames` variable is now conceptually unnecessary and should be eliminated (_maybe_ apart from the error message about cleanup).

storage/storage_image.go
```diff
@@ -1207,7 +1207,7 @@ func (s *storageImageDestination) Commit(ctx context.Context, unparsedToplevel t
...
 			names = append(names, oldNames...)
```
mtrmac, line 1208: … Then the `names` variable will have a trivial value and can be eliminated/inlined.

storage/storage_image.go
```diff
@@ -1207,7 +1207,7 @@ func (s *storageImageDestination) Commit(ctx context.Context, unparsedToplevel t
...
 		}
-		if err := s.imageRef.transport.store.SetNames(img.ID, names); err != nil {
+		if err := s.imageRef.transport.store.AddNames(img.ID, names); err != nil {
 			logrus.Debugf("error setting names %v on image %q: %v", names, img.ID, err)
 			return errors.Wrapf(err, "setting names %v on image %q", names, img.ID)
 		}
```
mtrmac, line 1213: All of these texts should only list the newly added names now; they shouldn’t list a set of names that might never actually exist in storage.

storage/storage_image.go
mtrmac, line 1207: Also update the `SetNames()` comment above.

flouthoc: Actually i wanted to be change to be minimum like just `s/SetNames/AddNames` but you are right conceptually we don't need `oldNames` anymore.

storage/storage_image.go
```diff
@@ -1197,21 +1197,18 @@ func (s *storageImageDestination) Commit(ctx context.Context, unparsedToplevel t
...
+	// values because AddNames() will deduplicate the list that we pass to it.
+	if name := s.imageRef.DockerReference(); name != nil {
 		names := []string{}
 		if name != nil {
 			names = append(names, name.String())
 		}
-		if len(oldNames) > 0 {
-			names = append(names, oldNames...)
+		if err := s.imageRef.transport.store.AddNames(img.ID, names); err != nil {
+			logrus.Debugf("error adding names %v on image %q: %v", names, img.ID, err)
```
vrothberg, line 1208: No need for a log when returning the same error.

storage/storage_image.go
```diff
@@ -1197,21 +1197,18 @@ func (s *storageImageDestination) Commit(ctx context.Context, unparsedToplevel t
...
 		logrus.Debugf("saved image metadata %q", string(metadata))
 	}
-	// Set the reference's name on the image.  We don't need to worry about avoiding duplicate
-	// values because SetNames() will deduplicate the list that we pass to it.
-	if name := s.imageRef.DockerReference(); len(oldNames) > 0 || name != nil {
+	// Adds the reference's name on the image.  We don't need to worry about avoiding duplicate
+	// values because AddNames() will deduplicate the list that we pass to it.
+	if name := s.imageRef.DockerReference(); name != nil {
 		names := []string{}
 		if name != nil {
```
vrothberg, line 1204: This check can be removed; we already know that `name != nil`.

storage/storage_image.go
```diff
@@ -1197,21 +1197,18 @@ func (s *storageImageDestination) Commit(ctx context.Context, unparsedToplevel t
...
+		if err := s.imageRef.transport.store.AddNames(img.ID, names); err != nil {
+			logrus.Debugf("error adding names %v on image %q: %v", names, img.ID, err)
+			return errors.Wrapf(err, "adding names %v on image %q", names, img.ID)
 		}
-		if err := s.imageRef.transport.store.SetNames(img.ID, names); err != nil {
-			logrus.Debugf("error setting names %v on image %q: %v", names, img.ID, err)
-			return errors.Wrapf(err, "setting names %v on image %q", names, img.ID)
-		}
-		logrus.Debugf("set names of image %q to %v", img.ID, names)
+		logrus.Debugf("add names of image %q to %v", img.ID, names)
```
vrothberg, line 1211:
```suggestion
		logrus.Debugf("add name %q to image %s",  name, img.ID)
```

storage/storage_image.go
```diff
@@ -1197,21 +1197,15 @@ func (s *storageImageDestination) Commit(ctx context.Context, unparsedToplevel t
...
-	// values because SetNames() will deduplicate the list that we pass to it.
-	if name := s.imageRef.DockerReference(); len(oldNames) > 0 || name != nil {
+	// Adds the reference's name on the image.  We don't need to worry about avoiding duplicate
+	// values because AddNames() will deduplicate the list that we pass to it.
+	if name := s.imageRef.DockerReference(); name != nil {
 		names := []string{}
-		if name != nil {
-			names = append(names, name.String())
+		names = append(names, name.String())
+		if err := s.imageRef.transport.store.AddNames(img.ID, names); err != nil {
```
vrothberg, line 1205: With the two lines above deleted.
```suggestion
		if err := s.imageRef.transport.store.AddNames(img.ID, string[]{name.String()}); err != nil {
```

vrothberg (commented):
LGTM

storage/storage_image.go
```diff
@@ -1197,21 +1197,13 @@ func (s *storageImageDestination) Commit(ctx context.Context, unparsedToplevel t
...
-	// values because SetNames() will deduplicate the list that we pass to it.
-	if name := s.imageRef.DockerReference(); len(oldNames) > 0 || name != nil {
-		names := []string{}
-		if name != nil {
-			names = append(names, name.String())
+	// Adds the reference's name on the image.  We don't need to worry about avoiding duplicate
+	// values because AddNames() will deduplicate the list that we pass to it.
+	if name := s.imageRef.DockerReference(); name != nil {
+		if err := s.imageRef.transport.store.AddNames(img.ID, []string{name.String()}); err != nil {
+			return errors.Wrapf(err, "adding names %v on image %q", name, img.ID)
```
mtrmac, line 1204:
```suggestion
			return errors.Wrapf(err, "adding names %v to image %q", name, img.ID)
```

storage/storage_image.go
```diff
@@ -1197,21 +1197,13 @@ func (s *storageImageDestination) Commit(ctx context.Context, unparsedToplevel t
...
 		}
-		if len(oldNames) > 0 {
-			names = append(names, oldNames...)
-		}
-		if err := s.imageRef.transport.store.SetNames(img.ID, names); err != nil {
-			logrus.Debugf("error setting names %v on image %q: %v", names, img.ID, err)
-			return errors.Wrapf(err, "setting names %v on image %q", names, img.ID)
-		}
-		logrus.Debugf("set names of image %q to %v", img.ID, names)
+		logrus.Debugf("add name %q to image %s", name, img.ID)
```
mtrmac, line 1206:
```suggestion
		logrus.Debugf("added name %q to image %s", name, img.ID)
```
Also, `img.ID` originally used `%q`, and I don’t currently see a reason to change that. (I don’t have a strong opinion on use of `%q` for `name`.)

flouthoc:
Wait for https://github.com/containers/storage/pull/1153 and tests to pass here: https://github.com/containers/podman/pull/13339

storage/storage_image.go
mtrmac: FWIW the way I think about this, there will never be more time or more motivation to clean up and simplify the code.

Keeping the code as simple and as easy to maintain as possible is an everyday job.

mtrmac (approved):
LGTM — just so that it isn‘t forgotten, the `replace` in `go.mod` should be replaced.

After that is done, and tests pass, feel free to merge without another review.

mtrmac:
Also: Thanks!

vrothberg (approved):
LGTM

go.mod
```diff
@@ -42,3 +42,5 @@ require (
 	golang.org/x/sys v0.0.0-20220128215802-99c3d69c2c27 // indirect
 	golang.org/x/term v0.0.0-20210927222741-03fcf44c2211
 )
+
+replace github.com/containers/storage => github.com/containers/storage v1.38.3-0.20220301151551-d06b0f81c0aa
```
mtrmac, line 46: The dependency above needs to be updated, without using `replace`; `replace` is ignored when importing c/image into other users.

vrothberg: thanks for catching that, @mtrmac !

flouthoc:
@mtrmac @vrothberg done.

mtrmac:
LGTM. Can you rebase it over the release, please, so that it’s trivially clear which commits part of 5.20.0 and which commits come after? Feel free to merge afterwards.

vrothberg (approved):
LGTM

flouthoc:
@containers/image-maintainers @mtrmac @vrothberg  I want to backport this commit to `v5.16.0` but I was unable to find any maintenance branch for that. Is there any branch where i could raise a PR to get this backported. 

Thanks

vrothberg:
@flouthoc, I just created one: https://github.com/containers/image/tree/release-5.16

When backporting, please make sure to bump the version accordingly in `version/version.go`.

mtrmac (commented):
ACK to the code (but blocked on a c/storage release).

WRT the test failures, is it possible to fix that by updating `SKOPEO_BRANCH` on this branch to an appropriate consumer of c/image 5.16?

PR #1482, opened 2022-03-01
Release v5.20.0

mtrmac:
- `docker/referece`: add `IsFullIdentifier`
- Changed oci layout transport to thread-safe destination
- add `pkg/blobcache` from Buildah
- blobcache: drop import on `buildah/docker`
- blobcache: drop history comment
- blobcache: make `ClearCache()` private
- blobcache: remove `CacheLookupReferenceFunc`
- blobcache: turn `BlobCache` into a struct
- blobcache: export `clearCache`
- Remove (unused and unreachable) keyring support
- Eliminate a goroutine
- Also introduces internal-only interfaces to allow extending the transport feature set in the future

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -6,12 +6,12 @@ const (
 	// VersionMajor is for an API incompatible changes
 	VersionMajor = 5
 	// VersionMinor is for functionality in a backwards-compatible manner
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

PR #1483, opened 2022-03-04
bug fix: handle CopySpecificImages correctly

philipp-rosenkranz-form3:
Currently the `ImageListSelection` setting `copy.CopySpecificImages` in `copy.Options` is useless since any digest slice passed to `Instances` must have the same length as the slice of manifests in the source list. 

Therefore, the only way to use `copy.CopySpecificImages` is to pass into `Instances` all digests which are present in the source list of manifests. Thus, `copy.CopySpecificImages` behaves like a really clunky version of `copy.CopyAllImages`.

This PR ensures that `copy.CopySpecificImages` is actually useful (again?) by removing any manifest from the list which cannot be matched with a digest listed in `Instances`.

### DIFF
--- a/manifest/docker_schema2_list.go
+++ b/manifest/docker_schema2_list.go
@@ -86,6 +86,30 @@ func (list *Schema2List) UpdateInstances(updates []ListUpdate) error {
 	return nil
 }
 
+func (list *Schema2List) UpdateInstancesAndDiscardOthers(listUpdate []ListUpdate) error {
+	existingManifest := map[digest.Digest]Schema2ManifestDescriptor{}
+	for _, m := range list.Manifests {
+		for _, u := range listUpdate {
+			if m.Digest == u.Digest {
+				existingManifest[u.Digest] = m
+			}
+		}
+	}
+	list.Manifests = []Schema2ManifestDescriptor{}
+	for _, u := range listUpdate {
+		updatedManifest, ok := existingManifest[u.Digest]
+		if !ok {
+			return errors.New("digest needs to exist in manifest")
+		}
+		updatedManifest.MediaType = u.MediaType
+		updatedManifest.Digest = u.Digest
+		updatedManifest.Size = u.Size
+		list.Manifests = append(list.Manifests, updatedManifest)
+	}
+
+	return nil
+}
+
 // ChooseInstance parses blob as a schema2 manifest list, and returns the digest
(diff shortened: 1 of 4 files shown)


manifest/docker_schema2_list.go
```diff
@@ -86,6 +86,30 @@ func (list *Schema2List) UpdateInstances(updates []ListUpdate) error {
...
 }
 
+func (list *Schema2List) UpdateInstancesAndDiscardOthers(listUpdate []ListUpdate) error {
+	existingManifest := map[digest.Digest]Schema2ManifestDescriptor{}
+	for _, m := range list.Manifests {
+		for _, u := range listUpdate {
+			if m.Digest == u.Digest {
+				existingManifest[u.Digest] = m
+			}
+		}
```
mtrmac, line 96: AFAICS this could unconditionally build `existingManifest[m.Digest] = m`; the entries to be removed would just be ignored. That turns this from O(N^2) into O(N).

Same in the other implementation.

manifest/docker_schema2_list.go
```diff
@@ -86,6 +86,30 @@ func (list *Schema2List) UpdateInstances(updates []ListUpdate) error {
...
+	}
+	list.Manifests = []Schema2ManifestDescriptor{}
+	for _, u := range listUpdate {
+		updatedManifest, ok := existingManifest[u.Digest]
```
mtrmac, line 100: I don’t think this works at all in the case where the copy had to convert the format of the original manifest, so `u.Digest` doesn‘t match any pre-existing value in `list.Manifests`.

mtrmac (changes requested):
Thanks for the PR.

For better or worse, the current implementation (essentially a sparse copy) is international (e.g. documented in `skopeo copy --multi-arch=index-only`). We can’t just break it.

A feature to strip the image of unwanted architectures (while still keeping it a multi-arch image, not just extracting a _single_ architecture) does sound useful, but it would be a new feature. There’s also a more generic request to add manifest editing support (e.g. https://github.com/containers/skopeo/issues/1136 ), and ideally we would want that to replace the c/common / Buildah / Podman implementations.

So that would be interesting, but a significantly larger work than this.

The need to keep the API of c/image stable is also a fairly significant constraint. One we _can_ work with, but it requires more infrastructure, and thinking carefully about any new API additions.

PR #1299, opened 2021-07-10
Improve registry client error handling

mtrmac:
See https://github.com/containers/image/issues/1293#issuecomment-875791392 and the added comment.

The primary expected benefit is that this should turn
```
Error initializing source docker://foobar:latest: Error reading manifest latest in docker.io/library/foobar: errors:
denied: requested access to the resource is denied
unauthorized: authentication required
```
into
```
Error initializing source docker://foobar:latest: Error reading manifest latest in docker.io/library/foobar: denied: requested access to the resource is denied
```

⚠️ 
- <s>Currently absolutely untested. We should see what this changes in practice for a variety of registries and situations.</s>
- <s>This could potentially _silently_ break callers; notably https://github.com/containers/common/blob/a38eda676e2adea08a5138aa639f873537cac335/pkg/retry/retry.go#L60 almost certainly needs to add a case for `ErrorCodeDenied` because the second `ErrorCodeUnauthorized` will not be visible.</s>

@vrothberg PTAL.

---

Tests pass in https://github.com/containers/skopeo/pull/1522 .

PR #1487, opened 2022-03-09
Remove github.com/docker/distribution/registry/client package

Luap99:
Using the github.com/docker/distribution/registry/client package will
import many huge prometheus dependencies, e.g.
 * github.com/prometheus/client_golang/prometheus/promhttp
 * github.com/prometheus/client_golang/prometheus
 * github.com/prometheus/procfs
and even more...

All of these dependencies are completely unused AFAICT but will still end
up in a binary because they are imported transitive.

github.com/docker/distribution/registry/client is only used to check
http errors so I think it makes sense to copy only the required code
into the docker package.

I vendored this commit into podman and it saves over 700KB in binary
size.

### DIFF
--- a/docker/distribution_error.go
+++ b/docker/distribution_error.go
@@ -1,0 +1,148 @@
+// Code below is taken from https://github.com/distribution/distribution/blob/a4d9db5a884b70be0c96dd6a7a9dbef4f2798c51/registry/client/errors.go
+// Copyright 2022 github.com/distribution/distribution authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package docker
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+
+	"github.com/docker/distribution/registry/api/errcode"
+	dockerChallenge "github.com/docker/distribution/registry/client/auth/challenge"
+)
+
+// errNoErrorsInBody is returned when an HTTP response body parses to an empty
+// errcode.Errors slice.
+var errNoErrorsInBody = errors.New("no error details found in HTTP response body")
+
+// unexpectedHTTPStatusError is returned when an unexpected HTTP status is
+// returned when making a registry api call.
+type unexpectedHTTPStatusError struct {
+	Status string
+}
+
+func (e *unexpectedHTTPStatusError) Error() string {
+	return fmt.Sprintf("received unexpected HTTP status: %s", e.Status)
+}
+
+// unexpectedHTTPResponseError is returned when an expected HTTP status code
+// is returned, but the content was unexpected and failed to be parsed.
+type unexpectedHTTPResponseError struct {
+	ParseErr   error
+	StatusCode int
+	Response   []byte
+}
+
+func (e *unexpectedHTTPResponseError) Error() string {
+	return fmt.Sprintf("error parsing HTTP %d response body: %s: %q", e.StatusCode, e.ParseErr.Error(), string(e.Response))
+}
+
+func parseHTTPErrorResponse(statusCode int, r io.Reader) error {
+	var errors errcode.Errors
+	body, err := ioutil.ReadAll(r)
+	if err != nil {
+		return err
+	}
+
+	// For backward compatibility, handle irregularly formatted
+	// messages that contain a "details" field.
+	var detailsErr struct {
+		Details string `json:"details"`
+	}
+	err = json.Unmarshal(body, &detailsErr)
+	if err == nil && detailsErr.Details != "" {
+		switch statusCode {
+		case http.StatusUnauthorized:
+			return errcode.ErrorCodeUnauthorized.WithMessage(detailsErr.Details)
+		case http.StatusTooManyRequests:
+			return errcode.ErrorCodeTooManyRequests.WithMessage(detailsErr.Details)
+		default:
+			return errcode.ErrorCodeUnknown.WithMessage(detailsErr.Details)
+		}
+	}
+
+	if err := json.Unmarshal(body, &errors); err != nil {
+		return &unexpectedHTTPResponseError{
+			ParseErr:   err,
+			StatusCode: statusCode,
+			Response:   body,
+		}
+	}
+
+	if len(errors) == 0 {
+		// If there was no error specified in the body, return
+		// UnexpectedHTTPResponseError.
+		return &unexpectedHTTPResponseError{
+			ParseErr:   errNoErrorsInBody,
+			StatusCode: statusCode,
+			Response:   body,
+		}
+	}
+
+	return errors
+}
+
+func makeErrorList(err error) []error {
+	if errL, ok := err.(errcode.Errors); ok {
+		return []error(errL)
+	}
+	return []error{err}
+}
+
+func mergeErrors(err1, err2 error) error {
+	return errcode.Errors(append(makeErrorList(err1), makeErrorList(err2)...))
+}
+
+// handleErrorResponse returns error parsed from HTTP response for an
+// unsuccessful HTTP response code (in the range 400 - 499 inclusive). An
+// UnexpectedHTTPStatusError returned for response code outside of expected
+// range.
+func handleErrorResponse(resp *http.Response) error {
+	if resp.StatusCode >= 400 && resp.StatusCode < 500 {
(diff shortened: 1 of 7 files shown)


rhatdan:
LGTM
@mtrmac @vrothberg PTAL

docker/errors.go
```diff
@@ -59,3 +63,144 @@ func registryHTTPResponseToError(res *http.Response) error {
 	}
 	return err
 }
+
+// Code below is taken from https://github.com/distribution/distribution/blob/a4d9db5a884b70be0c96dd6a7a9dbef4f2798c51/registry/client/errors.go
```
TomSweeneyRedHat, line 67: @vrothberg and @mtrmac 's call, but this almost seems like this new code should be in a separate file.  I"m going back and forth on it.

vrothberg: I concur. It should be a separate file.

In a second commit, we should also unexport the API and symbols.

vrothberg (commented):
Dropping 700k from the binary is sufficiently attractive to me.

We definitely need to add the tests as well but let's wait for @mtrmac's blessing before.

docker/errors.go
mtrmac: +1 to both, if this would end up happening.

docker/errors_test.go
```diff
@@ -8,7 +8,6 @@ import (
 	"testing"
```
mtrmac, line 8: If we are going to copy the code, we should also copy `registry/client/errors_test.go`, or any other relevant tests from the original repo.

mtrmac (commented):
As is, I’m not at all sure that this is worth it.
- (None of the `client.*Error` values are explicitly documented API, but they are soft of an API in practice. Though that’s much more the case for `github.com/docker/distribution/registry/api/errcode`, with various uses in c/common, Buildah, Podman.)
- Right now this would remove 700 kB, but soon https://github.com/sigstore/fulcio/blob/main/pkg/api/metrics.go would just drag it all back in, and we would we back where we started.

Basically I think we need to decide whether we are going to make no use of Prometheus (or only an opt-in use) a universal rule, and insist/contribute in upstreams to make that viable. Is that reasonable to aim for?

Alternatively, an (admittedly somewhat desperate) approach might be for the top-level users to use `go.mod`’s `replace` directive to just replace all of the Prometheus client with something dummy that does nothing and has no dependencies. (Compare https://github.com/prometheus/client_golang/issues/196 .) That would allow us to drop that dependency chain, whether upstreams cooperate or not.

---

(Now, moving _just_ the parser could be useful just because it would allow us to integrate better — `registryHTTPResponseToError` is just plumbing to interface with docker/distribution, and with the ability to modify `HandleErrorResponse`, most of that plumbing could almost entirely disappear. But that would be a different PR, with much more focus on interoperability/test coverage.)

Luap99:
My goal is to reduce the podman binary size. As it looks there is really not much to do there, the prometheus code is large and we do not need it. Many dependencies are deeply wired and cannot be removed, this one seems like an easy way to safe space. Most of my other findings will only reduce size by < 100 KB. So I would prefer if we could get this in.

> Right now this would remove 700 kB, but soon https://github.com/sigstore/fulcio/blob/main/pkg/api/metrics.go would just drag it all back in, and we would we back where we started.

Yeah that would be unfortunate.

> (None of the client.*Error values are explicitly documented API, but they are soft of an API in practice. Though that’s much more the case for github.com/docker/distribution/registry/api/errcode, with various uses in c/common, Buildah, Podman.)

Do you fear that this change is breaking consumers?

> Alternatively, an (admittedly somewhat desperate) approach might be for the top-level users to use go.mod’s replace directive to just replace all of the Prometheus client with something dummy that does nothing and has no dependencies. (Compare https://github.com/prometheus/client_golang/issues/196 .) That would allow us to drop that dependency chain, whether upstreams cooperate or not.

I doubt that maintenance would be worth it. It would be just a workaround for the actual issue and it would still take some space for the stub interface.

I leave this up for the maintainers to decide if you want to go forward with this before I apply the requested changes.

vrothberg:
> Right now this would remove 700 kB, but soon https://github.com/sigstore/fulcio/blob/main/pkg/api/metrics.go would just drag it all back in, and we would we back where we started.

Great catch, @mtrmac!  I don't think that it should block this PR here though.  Taming the dependencies will never be finished, I guess, and we can tackle them as they come in.  In case of fulcio, I am sure we can introduce a build-tag upstream to debloat - other callers may benefit as well.

mtrmac:
> I don't think that it should block this PR here though. Taming the dependencies will never be finished, I guess, and we can tackle them as they come in.

There’s a thin line between yak shaving an endless, but clearly useful work, and doing something that is ultimately known not to make a difference. I’m honestly not sure on which side we are here.

So I’m somewhat skeptical but not strongly opposed to this PR, and I’m fine with doing this if there is otherwise consensus that there is worth trying.

>  In case of fulcio, I am sure we can introduce a build-tag upstream to debloat - other callers may benefit as well.

Should we try to do that with docker/distribution = distribution/distribution as well, then? (Maybe even the same build tag?)

rhatdan:
I think we should go forward with this PR, and open an issue with Docker.

vrothberg:
> I think we should go forward with this PR, and open an issue with Docker.

That seems contradictory to me.

>  Should we try to do that with docker/distribution = distribution/distribution as well, then? (Maybe even the same build tag?)

Sounds like a good idea to me.  @Luap99, do you want to give it a shot?

Luap99:
> > Should we try to do that with docker/distribution = distribution/distribution as well, then? (Maybe even the same build tag?)
> 
> Sounds like a good idea to me. @Luap99, do you want to give it a shot?

I am not against trying this but size gained will be less, the current import path seems to be `github.com/docker/distribution/registry/client` -> `github.com/docker/distribution/registry/storage/cache` -> `github.com/docker/distribution/metrics` -> `github.com/docker/go-metrics` -> `github.com/prometheus/client_golang/prometheus` -> and more prometheus stuff

When we add a build tag to disable prometheus, we would only gain this one dependency. In theory the other packages in this list could import other big dependencies that we do not need. This is not be the case today but it could be in the future. 

If we want to change upstream I think it would make more sense to split the error handling into a separate package which does not need further imports and we could import this package instead.

tyler92:
Is there any chance to merge this PR? I find it very useful too.

Luap99:
> Right now this would remove 700 kB, but soon https://github.com/sigstore/fulcio/blob/main/pkg/api/metrics.go would just drag it all back in, and we would we back where we started.

@mtrmac Is the sigstore work fully merged?

I do not see any new references to `github.com/prometheus/client_golang/prometheus` so I think it would make sense to try to remove it from here.

I can also try to push https://github.com/distribution/distribution/pull/3608 forward if this option would be preferred?

mtrmac:
> > Right now this would remove 700 kB, but soon https://github.com/sigstore/fulcio/blob/main/pkg/api/metrics.go would just drag it all back in, and we would we back where we started.
>
> @mtrmac Is the sigstore work fully merged?

The Fulcio part does not exist in c/image yet — but I have also removed that dependency in https://github.com/sigstore/fulcio/pull/616 . So this specific concern no longer applies, and that turns this PR from ineffective to useful. (OTOH some more users of docker/distribution have been added. AFAICS those don’t include the registry/client subpackage right now.)

That makes this PR quite a bit more attractive to me. I read @vrothberg as being supportive of the idea from the start, so, let’s do it.

Could you address the earlier review feedback, please?

---

> I can also try to push https://github.com/distribution/distribution/pull/3608 forward if this option would be preferred?

It would be nice for the wider ecosystem, to spread the idea that simple clients should not have hard-coded dependencies on infrastructure for serving metrics over HTTP. Personally, I wouldn’t say that I _prefer_ that PR any more — d/distribution releases have been very infrequent, and moving the HTTP parser in-package so that we don’t need to override its outcomes in docker/errors.go (as in #1299 ) could — eventually, _after_ we have much more test coverage and confidence — result in a simpler implementation overall.

Luap99:
> Could you address the earlier review feedback, please?

done

rhatdan:
LGTM

docker/distribution_error.go
```diff
@@ -0,0 +1,148 @@
...
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+
+	"github.com/docker/distribution/registry/api/errcode"
+	dockerChallange "github.com/docker/distribution/registry/client/auth/challenge"
```
mtrmac, line 27:
```suggestion
	dockerChallenge "github.com/docker/distribution/registry/client/auth/challenge"
```
(or just leave it as `challenge` without renaming?)

mtrmac (commented):
LGTM other than that one nit, feel free to merge without another review round.

Thanks!

(FWIW, this removes about 3 % from Skopeo’s size on macOS.)

docker/distribution_error.go
Luap99: oh yeah lets not rename imports if it is not required

Luap99: looks like we have to rename since challenge is already declared here: https://github.com/containers/image/blob/3aed337418462460c111a3beade6723832af5be0/docker/wwwauthenticate.go#L13
but will fix the typo.

mtrmac (approved):
LGTM. Thanks!

PR #174, opened 2016-11-30
copy: provide src&dest contexts

runcom:
Signed-off-by: Antonio Murdaca <runcom@redhat.com>

PR #1499, opened 2022-03-21
Clarified optional.creator section from the signature specs.

ipanova:
Added a clarification that creator is the low-level software
implementation.

### DIFF
--- a/docs/containers-signature.5.md
+++ b/docs/containers-signature.5.md
@@ -211,6 +211,7 @@ Consumers still SHOULD reject any signature where a member of an `optional` obje
 ### `optional.creator`
 
 If present, this MUST be a JSON string, identifying the name and version of the software which has created the signature.
+This should be the low-level software implementation and not the top-level caller.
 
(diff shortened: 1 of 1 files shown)


docs/containers-signature.5.md
```diff
@@ -211,6 +211,7 @@ Consumers still SHOULD reject any signature where a member of an `optional` obje
 ### `optional.creator`
 
 If present, this MUST be a JSON string, identifying the name and version of the software which has created the signature.
+This should be the low-level software implementation and not the top-level caller.
```
mtrmac, line 214: Non-blocking: Maybe “refer to” or “identify” instead of “be”?

mtrmac (commented):
Thanks!

Please use upper-case SHOULD to be consistent with the rest of the document.

mtrmac:
#174 attempts to clarify this.

mtrmac:
Thanks again for pointing this out!

PR #1500, opened 2022-03-21
Added documentation for containers-storage transport.

ipanova:
It is important to mention it when the default policy is in restrictive
mode.
During push operation image is being copied from local storage to the
docker transport. This action should be allowed for a successfull push.

### DIFF
--- a/docs/containers-policy.json.5.md
+++ b/docs/containers-policy.json.5.md
@@ -110,6 +110,12 @@ The `tarball:` transport refers to tarred up container root filesystems.
 
 Scopes are ignored.
 
+### `containers-storage:`
+
+The `containers-storage` transport refers to the image location in a local containers storage.
+
+Supported scopes are for a specific image(s) or `""` for all images located in the containers-storage.
+
 ## Policy Requirements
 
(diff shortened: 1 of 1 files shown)


docs/containers-policy.json.5.md
```diff
@@ -110,6 +110,12 @@ The `tarball:` transport refers to tarred up container root filesystems.
 
 Scopes are ignored.
 
+### `containers-storage:`
+
+The `containers-storage` transport refers to the image location in a local containers storage.
+
+Supported scopes are for a specific image(s) or `""` for all images located in the containers-storage.
```
mtrmac, line 117: The supported scopes are actually quite different (notably they always start with a `[…]` store specification), see `storageReference.PolicyConfiguration{Identity,Namespaces}`.

mtrmac (commented):
Thanks!

I’m afraid we have not kept this documentation up, I have at least filed https://github.com/containers/image/issues/1501 for that.

It’s admittedly unfair to block this PR (documenting the important `podman push` case) on getting the containers-storage: scopes _exactly_ right. I guess I could live with a cop-out “documentation of scopes TBD”, if no-one has time to document it correctly now.

PR #1506, opened 2022-03-24
pkg/docker: lock auth.json before reading/writing

vrothberg:
Read/Write lock each auth.json file before reading/writing
it to prevent race coditions.

Fixes: #1365

### DIFF
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -15,6 +15,7 @@ import (
 	"github.com/containers/image/v5/pkg/sysregistriesv2"
 	"github.com/containers/image/v5/types"
 	"github.com/containers/storage/pkg/homedir"
+	"github.com/containers/storage/pkg/lockfile"
 	helperclient "github.com/docker/docker-credential-helpers/client"
 	"github.com/docker/docker-credential-helpers/credentials"
 	"github.com/hashicorp/go-multierror"
@@ -541,6 +542,29 @@ func getPathToAuthWithOS(sys *types.SystemContext, goOS string) (string, bool, e
 // or returns an empty dockerConfigFile data structure if auth.json does not exist
 // if the file exists and is empty, readJSONFile returns an error
 func readJSONFile(path string, legacyFormat bool) (dockerConfigFile, error) {
+	if _, err := os.Stat(path); err != nil {
+		if os.IsNotExist(err) {
+			var auths dockerConfigFile
+			auths.AuthConfigs = map[string]dockerAuthConfig{}
+			return auths, nil
+		}
+		return dockerConfigFile{}, err
+	}
+
+	// Make sure to read-lock the file before reading it.
+	lock, err := lockfile.GetLockfile(path)
+	if err != nil {
+		return dockerConfigFile{}, fmt.Errorf("creating lock: %w", err)
+	}
+	lock.RLock()
+	defer lock.Unlock()
+
+	return readJSONFileLocked(path, legacyFormat)
+}
+
+// Implements readJSONFile but without locking the specified path.
+// The caller is responsible for synchronizing access to the path.
+func readJSONFileLocked(path string, legacyFormat bool) (dockerConfigFile, error) {
 	var auths dockerConfigFile
 
 	raw, err := ioutil.ReadFile(path)
@@ -550,6 +574,14 @@ func readJSONFile(path string, legacyFormat bool) (dockerConfigFile, error) {
 			return auths, nil
 		}
 		return dockerConfigFile{}, err
+	}
+
+	// When creating a file for the first time, it may be empty at this
+	// point as it was created by the write lock. Hence, return early here
+	// to prevent an unmarshal error; an empty file is not valid JSON.
+	if len(raw) == 0 {
+		auths.AuthConfigs = map[string]dockerAuthConfig{}
+		return auths, nil
 	}
 
 	if legacyFormat {
@@ -590,7 +622,15 @@ func modifyJSON(sys *types.SystemContext, editor func(auths *dockerConfigFile) (
 		return "", err
 	}
 
-	auths, err := readJSONFile(path, false)
+	// Make sure to write-lock the file before writing to it.
+	lock, err := lockfile.GetLockfile(path)
+	if err != nil {
+		return "", fmt.Errorf("creating lock: %w", err)
+	}
+	lock.Lock()
+	defer lock.Unlock()
+
+	auths, err := readJSONFileLocked(path, false)
 	if err != nil {
 		return "", errors.Wrapf(err, "reading JSON file %q", path)
 	}


rhatdan:
Where is the lock for writing?

vrothberg:
> Where is the lock for writing?

Each auth.json file is used as a lock. This way, we don't have to worry about finding a path for a lockfile for (root)less on Mac, Win, and Linux.

vrothberg:
Doesn't seem to work yet as I want to though.

rhatdan:
Not what I meant. I would think there is a code path when writing the auth.json file that should take the lock, but this PR only handles the read paths.  But I am not overly familiar with this code.

vrothberg:
> Not what I meant. I would think there is a code path when writing the auth.json file that should take the lock, but this PR only handles the read paths. But I am not overly familiar with this code.

The writing part is buried at the bottom: https://github.com/containers/image/pull/1506/files#diff-893e8c3da19873ab36cfae5eb73229819aed89b5115aca90e1aa7960ab661454R616-R624

rhatdan:
Ok I missed the change in function name.

LGTM

vrothberg:
Here we go ... the CI jobs are back

pkg/docker/config/config.go
```diff
@@ -541,6 +542,28 @@ func getPathToAuthWithOS(sys *types.SystemContext, goOS string) (string, bool, e
...
+		return dockerConfigFile{}, fmt.Errorf("creating lock: %w", err)
+	}
+	lock.RLock()
+	defer lock.Unlock()
+
+	return readJSONFileNoLock(path, legacyFormat)
+}
+
+// Implements readJSONFile but without locking the specified path.
+func readJSONFileNoLock(path string, legacyFormat bool) (dockerConfigFile, error) {
```
mtrmac, line 566: Something like
```suggestion
// Implements readJSONFile but without locking the specified path. The caller is responsible for locking the lock.
func readJSONFileLocked(path string, legacyFormat bool) (dockerConfigFile, error) {
```
to make it very clear that it’s not OK to just call this method without holding a lock?

At least the `…Locked` naming convention is used in other places in c/image.

pkg/docker/config/config.go
```diff
@@ -541,6 +542,28 @@ func getPathToAuthWithOS(sys *types.SystemContext, goOS string) (string, bool, e
...
+		if os.IsNotExist(err) {
+			var auths dockerConfigFile
+			auths.AuthConfigs = map[string]dockerAuthConfig{}
+			return auths, nil
+		}
+		return dockerConfigFile{}, err
+	}
+
+	// Make sure to read-lock the file before reading it.
+	lock, err := lockfile.GetLockfile(path)
```
mtrmac, line 555: Isn’t the path parameter a path of the _lock_? Which contains a random value to support `.Modified`? How can `auth.json` be a JSON and a random string at the same time?

pkg/docker/config/config.go
```diff
@@ -552,6 +575,14 @@ func readJSONFile(path string, legacyFormat bool) (dockerConfigFile, error) {
 		return dockerConfigFile{}, err
 	}
 
+	// When creating a file for the first time, it may be empty at this
+	// point as it was created by the write lock. Hence, return early here
```
mtrmac, line 579: That’s surprising to me (and incompatible with older versions of consumers, though that’s a small concern — they could easily race with `WriteFile` and see an empty file, and fail, for that reason).

See elsewhere about sharing the lock with JSON.

mtrmac (commented):
Thanks! Using `c/storage/pkg/lockfile`, when we already depend on it, is a great idea.

pkg/docker/config/config.go
vrothberg: `Touch()` and `Modified()` don't work in this scenario but I think that's OK in this scenario since we're controlling it here.

pkg/docker/config/config.go
vrothberg: I think it's a valid concern but I do not have a good/definitive answer/idea.  We could use one central (similar to the short-names lock) _but_ old client would still compete with newer ones.

vrothberg:
@mtrmac updated the name and comment. Let's continue the conversation on the remaining locking issue in the threads.

pkg/docker/config/config.go
mtrmac: I was thinking about using a separate `auth.json.lock` file; then the concerns about the lock file object trying to own file contents would go away, and so would this situation, AFAICS.

pkg/docker/config/config.go
mtrmac: Ah, I have not been reading that carefully enough; the contents are not actually _used_ if we call neither `Touch()` nor `Modified()`.

Still, that feels like an implementation detail that could change any time. I suppose our tests would catch it…

pkg/docker/config/config.go
mtrmac: The underlying `fcntl` lock is unlocked when _any_ file descriptor to that file is closed, isn’t it?

If so, `ioutil.ReadFile` of the JSON always releases the lock, making the `modifyJSON` lock ineffective — and we do need to use a different path.

vrothberg: > I was thinking about using a separate `auth.json.lock` file; then the concerns about the lock file object trying to own file contents would go away, and so would this situation, AFAICS.

I was considering that but didn't like the user experience.  I would find it surprising/annoying if I do a `--authfile file` and would find a `file.lock` afterwards.

> The underlying `fcntl` lock is unlocked when _any_ file descriptor to that file is closed, isn’t it?

Holy frog, that is true and buried in the man pages. That indeed is a K.O. for this approach - we could do ref-counting on the path but that seems a bit too hacky for my taste.

mtrmac: Just to be explicit,`pkg/lockfile` is already using a global hash table and refcounting the objects to work correctly with those locks, so that works for exclusively-used lockfiles (and as long as c/storage doesn’t get a separate `/v2` with a separate hash table); but it breaks if there is any _other_ user of that file, like here.

mtrmac: … the point about `file.lock` files on user-specified `--authfile` paths (outside of dot files, in user-maintained project repositories and the like) being annoying is quite valid, though; we might not even necessarily be able to create a lockfile (if the `auth.json` is in a read-only configuration directory).

flock(2) might be an option, but it is distinctly less portable (and the Linux man pages suggest that some implementations implement it on top of the `fcntl` locks. I’m sure there are good write-ups about the various locking mechanisms and their trade-offs, I haven’t now taken the time to find and read one.

mtrmac: A possible _partial_ alternative: use `c/storage/pkg/ioutils.AtomicWriteFile`. That could still lose data with concurrent writers, but at least it wouldn’t break readers concurrent with a single writer, who can now see empty files or partial contents (= invalid JSON).

vrothberg: Seems like we're between a rock and hard place.  I will close this PR as I ran out of time and really do not have a good idea yet.

mtrmac: Let’s at least do the `AtomicWriteFile` part, that should be safe and make _some_ difference: https://github.com/containers/image/pull/1515 .

Issue #1284, opened 2021-06-29
Make sure unknown keys in config files (typos, keys for future versions) are rejected


mtrmac:
We tend to decode config files using the easiest method available, which is typically server-API-focused, expects version drift, and just ignores mismatches.

If at all possible, we should detect invalid config files (and have tests for that), so that typos are detected and reported immediately.

E.g. the TOML package we use actually supports that:
```go
	_, err := toml.DecodeFile(path, &combinedTOML)
```
vs.
```go
	meta, err := toml.DecodeFile(path, &combinedTOML)
	if err != nil {
		return nil, err
	}
	// meta.Undecoded()
```
also we might be able to use `meta.IsDefined` to handle present-but-empty vs. not-present values, I haven’t yet looked into the details.

mtrmac:
Once the APIs are researched and tested, it might be worth looking at this throughout the containers ecosystem at the same time.

rhatdan:
Yes I like the idea for verifying config files.

PR #1507, opened 2022-03-25
toml decoding: report unknown keys

vrothberg:
Despite the desire in #1284 to report unknown-key errors (e.g., for
typos), only log them on the debug level.  We have made rather bad
experiences with containers.conf which initially logged these errors
on the warning level which caused an immense noise for users.

The underlying problem is that not all tools are updated in sync,
especially for non-{CentOS,Fedora,RHEL} distributions.  Config files
changes for tools with an updated conf will hence cause noise for others
etc.

However, logging these error on the debug level will still improve user
experience and be especially helpful in analyzing users issues.

Fixes: #1284

@mtrmac @rhatdan PTAL


rhatdan:
LGTM

pkg/sysregistriesv2/shortnames.go
```diff
@@ -315,11 +316,14 @@ func (c *shortNameAliasCache) updateWithConfigurationFrom(updates *shortNameAlia
...
 	conf := shortNameAliasConf{}
 
-	_, err := toml.DecodeFile(confPath, &conf)
+	meta, err := toml.DecodeFile(confPath, &conf)
 	if err != nil && !os.IsNotExist(err) {
 		// It's okay if the config doesn't exist.  Other errors are not.
 		return nil, nil, errors.Wrapf(err, "loading short-name aliases config file %q", confPath)
 	}
+	if keys := meta.Undecoded(); len(keys) > 0 {
+		logrus.Debugf("Failed to decode the keys %q from %q", keys, confPath)
```
mtrmac, line 325: `s/the keys/keys/`? I’m not a native speaker.

(In both cases.)

mtrmac (commented):
Thanks! LGTM.

Feel free to merge as is or after tweaking the working, without another review by me.

mtrmac (approved):
Thanks again!

PR #1503, opened 2022-03-23
[v5.19, backport] storage: use race-free `AddNames` instead of `SetNames`

flouthoc:
Backport of: https://github.com/containers/image/pull/1480

Commits from parallel builds using `SetNames` removes `names` from
storage for other builds.

Use race-free atomic `AddNames` to prevent breaking of parallel builds.

PR #1512, opened 2022-04-01
[v5.16, backport] storage: use race-free `AddNames` instead of `SetNames`

flouthoc:
Backport of: https://github.com/containers/image/pull/1480
Similar to: https://github.com/containers/image/pull/1503

Commits from parallel builds using `SetNames` removes names from
storage for other builds.

Use race-free atomic `AddNames` to prevent breaking of parallel builds.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/containerd/containerd v1.5.4 // indirect
 	github.com/containers/libtrust v0.0.0-20190913040956-14b96171aa3b
 	github.com/containers/ocicrypt v1.1.2
-	github.com/containers/storage v1.37.0
+	github.com/containers/storage v1.36.3
 	github.com/docker/distribution v2.7.1+incompatible
 	github.com/docker/docker v20.10.8+incompatible
 	github.com/docker/docker-credential-helpers v0.6.4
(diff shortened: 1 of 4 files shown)


flouthoc:
@mtrmac PTAL. I am just concerned about why `c/storage` was pointing to `v1.37.0` on previous branch when both `podman` and `buildah` using this release are using `c/storage` `v1.36.1`

go.mod
```diff
@@ -8,7 +8,7 @@ require (
 	github.com/containerd/containerd v1.5.4 // indirect
 	github.com/containers/libtrust v0.0.0-20190913040956-14b96171aa3b
 	github.com/containers/ocicrypt v1.1.2
-	github.com/containers/storage v1.37.0
+	github.com/containers/storage v1.36.3
```
mtrmac, line 11: This is not acceptable.

(Alternative approaches are being discussed, just noting this here so that it doesn’t get merged as is.)

mtrmac (changes requested):

go.mod
flouthoc: @TomSweeneyRedHat @mheon FYI this is sort of a trouble where we are stuck and discussing an alternative approach.

TomSweeneyRedHat: @flouthoc looks like a good Cabal topic.  Could you add it to the list please?

flouthoc: @TomSweeneyRedHat I created this PR for `v3.4.2-rhel` but afaik the backport for `v3.4.2-rhel` is not decided so we might not need this PR anymore.

mtrmac:
Closing in favor of #1519 ; please speak up if that’s incorrect.

PR #1515, opened 2022-04-04
Use c/storage/pkg/ioutils.AtomicWriteFiles to update auth.json

mtrmac:
... so that users can't observe an intermediate empty, or partial, file.

(Note that there is still no locking, so two simultaneous updates to `auth.json` might lose some updates.)

Fixes at least the worst impact of https://github.com/containers/image/issues/1365 .


mtrmac:
`c/image/storage` already depends on `c/storage/pkg/ioutils`, so for those users this is not a new dependency. Even pure users of `c/image/storage` should only need to depend on a few very small sub packages more.

vrothberg (approved):
LGTM

Issue #1523, opened 2022-04-21
Clarify usage of registries.conf with different mirrors


telmich:
I hope this is the right cri-o repo to report to, if not, please advise.

## General Situation

* k8s clusters usually pull images from multiple registries like quay.io and dockerhub
* To be able to locally cache/proxy images, one needs to create a specific proxy *per registry* - resulting in https://proxy-for-docker-hub and https://proxy-for-quay-io
* You might have multiple proxies for each registry for redundancy like https://proxy-for-docker-hub-1 and https://proxy-for-docker-hub-2
* In the registries.conf the mirror list applies for *all* registries, which leads to a lot of false hits

## Problem

We cannot specify that a certain mirror is only for certain paths. Thus resulting in a lot of incorrect accesses over the proxies.

## Pratical example

The following configuration is taken from a sample kubernetes cluster:

```
[09:29] server27.place10:~# cat /etc/containers/registries.conf
unqualified-search-registries = ["docker.io"]

[[registry]]
prefix = "docker.io"
location = "registry.hub.docker.com"

[[registry]]
prefix = "quay.io"
location = "quay.io"

# List of our mirrors
[[registry.mirror]]
location = "harbor.ungleich.svc.p10.k8s.ooo/quayio"

[[registry.mirror]]
location = "harbor.ungleich.svc.p10.k8s.ooo/dockerhub"

[[registry.mirror]]
location = "harbor.ungleich.svc.c2.k8s.ooo/dockerhub"

[[registry.mirror]]
location = "harbor.ungleich.svc.c2.k8s.ooo/quayio"
```

As you can see there are 2 quayio mirrors and 2 dockerhub mirrors. We plan to add 3 mirrors per official source for redundancy reasons. Thus per pull request cri-o might fall through 3 incorrect mirrors and then finally access the correct one. 

## Improvement

We suggest to add a new field, `prefix` to the mirror list:

```
[[registry.mirror]]
location = "harbor.ungleich.svc.c2.k8s.ooo/quayio"
prefix="quay.io"
```

This way only the mirrors with the right prefix would be contacted.

vrothberg:
Thanks for reaching out, @telmich!  You've found the right place for this kind of question.

If you have a dedicated mirror for Docker Hub and a dedicated mirror Quay, an exemplary config may look as follows:
```TOML
[[registry]]
location = "docker.io"
[[registry.mirror]]
location = "docker.mirror.com"

[[registry]]
location = "quay.io"
[...]

PR #1525, opened 2022-04-21
man containers-registries.conf: clarify mirror association

vrothberg:
Clarify that a `[[registry.mirror]]` is associated only with the
previous `[[registry]]`.

Fixes: #1523

### DIFF
--- a/docs/containers-registries.conf.5.md
+++ b/docs/containers-registries.conf.5.md
@@ -99,7 +99,7 @@ as-is. But other settings like insecure/blocked/mirrors will be applied to match
 
 `mirror`
 : An array of TOML tables specifying (possibly-partial) mirrors for the
-`prefix`-rooted namespace.
+`prefix`-rooted namespace (i.e., the current `[[registry]]` TOML table).
 
 The mirrors are attempted in the specified order; the first one that can be
 contacted and contains the image will be used (and if none of the mirrors contains the image,
@@ -236,6 +236,12 @@ location = "example-mirror-0.local/mirror-for-foo"
 [[registry.mirror]]
 location = "example-mirror-1.local/mirrors/foo"
 insecure = true
+
+[[registry]]
+location = "registry.com"
+
+[[registry.mirror]]
+location = "mirror.registry.com"
 ```
 Given the above, a pull of `example.com/foo/image:latest` will try:
     1. `example-mirror-0.local/mirror-for-foo/image:latest`
@@ -243,6 +249,8 @@ Given the above, a pull of `example.com/foo/image:latest` will try:
     3. `internal-registry-for-example.net/bar/image:latest`
 
 in order, and use the first one that exists.
+
+Note that a mirror is associated only with the current `[[registry]]` TOML table. If using the example above, pulling the image `registry.com/image:latest` will hence only reach out to `mirror.registry.com`, and the mirrors associated with `example.com/foo` will not be considered.
 
 ## VERSION 1 FORMAT - DEPRECATED
 VERSION 1 format is still supported but it does not support


vrothberg:
@mtrmac PTAL

rhatdan:
LGTM
@TomSweeneyRedHat PTAL

docs/containers-registries.conf.5.md
```diff
@@ -99,7 +99,7 @@ as-is. But other settings like insecure/blocked/mirrors will be applied to match
 
 `mirror`
 : An array of TOML tables specifying (possibly-partial) mirrors for the
-`prefix`-rooted namespace.
+`prefix`-rooted namespace (i.e., the previous registry).
```
mtrmac, line 102: Is that the “current” registry, actually?

It’s not even a _“registry”_ at all, it’s that `prefix`-delineated space, which might be smaller than a registry.

How about “the current `[[registry]]` TOML table`?

docs/containers-registries.conf.5.md
```diff
@@ -244,6 +250,8 @@ Given the above, a pull of `example.com/foo/image:latest` will try:
 
 in order, and use the first one that exists.
 
+Note that a mirror is associated only with the previous registry.  Pulling `registry.com/image:latest` will hence only reach out to `mirror.registry.com` and not use the one associated to the other registry.
```
mtrmac, line 253: Again, consider something like “`[[registry]]` TOML table” rather than just “registry” (in both sentences). Also `s/previous/current/` — `example.com/foo` is the “previous” one.

Or maybe say something to express the idea that “a [[registry.mirror]] TOML table is, per TOML semantics, logically nested inside the previous [[registry]] table, and only applies to that [[registry]] table” — to talk about the options and their nesting, instead of relative ordering of tables as if they were conceptually unrelated. I don’t have a good idea of which one is easier to understand.

mtrmac, line 253: Would it be clearer (or much less clear??) to explicitly list the [`"example-mirror-0.local"`, `"example-mirror-1.local"`] names instead of saying “the other registry”? Or maybe both, “and use neither `"example-mirror-0.local"` nor `"example-mirror-1.local"`, which only apply to the `example.com/foo` namespace”— OTOH that seems _way_ too long.

mtrmac (commented):
Thanks!

Just some non-blocking suggestions to consider. I’ll fully defer to native speakers in all respects.

docs/containers-registries.conf.5.md
```diff
@@ -236,6 +236,12 @@ location = "example-mirror-0.local/mirror-for-foo"
...
 insecure = true
+
+[[registry]]
+location = "registry.com"
+
+[[registry.mirror]]
+location = "mirror.registry.com"
 ```
 Given the above, a pull of `example.com/foo/image:latest` will try:
     1. `example-mirror-0.local/mirror-for-foo/image:latest`
```
TomSweeneyRedHat, line 247: Not part of this change, but it looks like there's an extra leading space on this line.

docs/containers-registries.conf.5.md
vrothberg: Good idea!  We can keep it shorter by saying "the mirror of `example.com/foo`"

vrothberg:
@mtrmac @TomSweeneyRedHat PTAL

mtrmac (approved):
LGTM. Thanks!

docs/containers-registries.conf.5.md
mtrmac: Let’s fix that: https://github.com/containers/image/pull/1527 (_after_ this PR).

rhatdan:
LGTM

PR #1527, opened 2022-04-22
Use correct list syntax in containers-registries.conf(5)

mtrmac:
Four leading spaces are interpreted by go-md2man as a code block.

Add a new line to start a new paragraph, so that go-md2man recognizes the list syntax.


rhatdan:
LGTM
@vrothberg PTAL

vrothberg (approved):
LGTM

PR #1530, opened 2022-04-26
Improve progress bars, and update mpb

### LINKED ISSUE #1464: audit Bar.SetTotal calls in copy.go
- Some seem to be unnecessary — `createProgressBar` already uses `blobInfo.Size`
- `copyLayer`, after a copy is done, calls `SetTotal` with the source info (which might not contain a size), when it certainly has a size from the updated result.

It’s probably just bugs, but review also the history in case some of those are non-obvious and necessary.

mtrmac:
Resolves #1464 (see there for rationale), and replaces #1473 . The general principle is that progress bars track object size at the _source_ transport now, consistently everywhere.

Also various other related cleanups in this area.


rhatdan:
LGTM 
There were a couple of typos in your commit messages, but nothing requiring a repush.  Excellent job.
@vrothberg PTAL

vrothberg (approved):
Changes LGTM.

Something I would love to see in the UX is to draw all progress bars _before_ pulling. Currently, a bar is drawn the moment we start pulling the blob.

It would be nice to have them all drawn before.  Unrelated to this PR though.

@mtrmac feel free to merge.

PR #1533, opened 2022-04-27
Fix unwanted reuse of encrypted layers

mtrmac:
Per https://github.com/containers/image/pull/1314 :

Operations that involve encryption/decryption are already restricted e.g. not to use `TryReusingBlob`; but operations that don't themselves involve encryption could still find encrypted blobs in the blob info cache, and potentially use them in other contexts.
    
To avoid that, use a somewhat big hammer of just not calling `RecordDigestUncompressedPair` on that.  Note that this does not help if the blob info cache has already added such entries before this change; it only makes a difference for the future.
    
We continue to call `RecordKnownLocation` with encrypted data; simple copies of encrypted images from one registry to another (which don't encrypt/decrypt as part of the copy) can benefit from e.g. cross-repo blob reuse just fine.
    
It seems likely that a more precise logic which records more data and allows more blob reuse could be built, but it's not trivially obvious to me that it would be safe, so this change only does the conservative thing to avoid known breakage.

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -1253,9 +1253,17 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, to
 					return types.BlobInfo{}, "", errors.Wrap(diffIDResult.err, "computing layer DiffID")
 				}
 				logrus.Debugf("Computed DiffID %s for layer %s", diffIDResult.digest, srcInfo.Digest)
-				// This is safe because we have just computed diffIDResult.Digest ourselves, and in the process
-				// we have read all of the input blob, so srcInfo.Digest must have been validated by digestingReader.
-				ic.c.blobInfoCache.RecordDigestUncompressedPair(srcInfo.Digest, diffIDResult.digest)
+				// Don’t record any associations that involve encrypted data. This is a bit crude,
+				// some blob substitutions (replacing pulls of encrypted data with local reuse of known decryption outcomes)
+				// might be safe, but it’s not trivially obvious, so let’s be conservative for now.
+				// This crude approach also means we don’t need to record whether a blob is encrypted
+				// in the blob info cache (which would probably be necessary for any more complex logic),
+				// and the simplicity is attractive.
+				if !encryptingOrDecrypting {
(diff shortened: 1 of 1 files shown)


mtrmac:
@vrothberg @nalind @hshiina @lumjjb PTAL.

I haven’t tested this PR in practice yet, still I’d like to maximize the opportunity for others to find problems in this PR.

vrothberg (commented):
Changes LGTM

mtrmac:
Updated now to remove one more place where “we have to compute DiffIDs” was set when encryption is involved. (I can find no record of why there were two places doing this.)

---

Tested with
```sh
rm ~/.local/share/containers/cache/blob-info-cache-v1.boltdb
rm -rf t t-{encrypted,unencrypted,decrypted}
# --dest-decompress to make sure the (uncompressed digest, compressed digest)-association-based reuse logic is involved on pushes
bin/skopeo --debug --override-os linux copy --format oci --dest-decompress docker://quay.io/libpod/alpine:latest dir:t
bin/skopeo --debug --override-os linux copy --encryption-key jwe:./encryption-test-public.pem dir:t docker://$registry/first-encrypted
bin/skopeo --debug --override-os linux copy docker://$registry/first-encrypted dir:t-encrypted
# This used to reuse the encrypted blob from first-encrypted, and is no longer doing so.
bin/skopeo --debug --override-os linux copy dir:t docker://$registry/second-unencrypted
bin/skopeo --debug --override-os linux copy docker://$registry/second-unencrypted dir:t-unencrypted
# Sanity check that the encrypted content can be decrypted
bin/skopeo --debug  --override-os linux copy --decryption-key encryption-test-private.pem dir:t-encrypted dir:t-decrypted
# (And manually inspect the 4 t* directories)
```

hshiina:
I reproduced this bug in buildah tests(containers/buildah#3941) and confirmed the test succeeds with this fix. Thanks!

mtrmac:
@rhatdan, @vrothberg this was reported as a RHEL bug, so it would be useful to include into the upcoming vendor dance.

lumjjb (approved):
LGTM! @mtrmac this is an awesome PR! It opens this up for more granular encryption policy!

PR #1534, opened 2022-04-29
Bump github.com/containers/ocicrypt to 566b808

lsm5:
The two latest commits in c/ocicrypt get rid of OAEPDefaultHash envvar
and default to sha256.

They are waiting for confirmation on their rust side before cutting a
release. So, hoping to get this in assuming dependabot could take a
while.

@mtrmac @vrothberg @rhatdan PTAL.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/BurntSushi/toml v1.1.0
 	github.com/containerd/containerd v1.6.1 // indirect
 	github.com/containers/libtrust v0.0.0-20200511145503-9c3a6c22cd9a
-	github.com/containers/ocicrypt v1.1.3
+	github.com/containers/ocicrypt v1.1.4-0.20220428134531-566b808bdf6f
 	github.com/containers/storage v1.38.3-0.20220301151551-d06b0f81c0aa
 	github.com/docker/distribution v2.8.1+incompatible
 	github.com/docker/docker v20.10.14+incompatible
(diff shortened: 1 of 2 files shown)


mtrmac (commented):
(I don’t think Dependabot is bringing in commits that aren’t a tagged release, at least in our configuration.)

LGTM. (This changes default behavior on encryption, but at least from the diff, no changes were necessary on the recipient side, so the new behavior should still be interoperable with older consumers.)

Thanks!

cevich:
Yep, exactly.  https://github.com/containers/image/pull/1535

lsm5:
@mtrmac should be good to go now.

mtrmac (approved):

lsm5:
@mtrmac is this worth a new version for dependabot's sake, or would you be cool if I vendor the commit into our usual tools? No rush of course, enjoy your weekend :)

mtrmac:
@lsm5 I think @rhatdan is planning a vendor dance for start of next week – maybe you want it in the same timeframe for the same purpose?

If so, I’d prefer to wait until Monday to give others (and me, for #1533) a chance to get a perhaps some last-minute PRs in.

mtrmac:
… WRT vendoring by commit, the code should be in good enough shape to do that if you want to e.g. benefit from it even in advance of that vendor dance.

lsm5:
I'm cool with waiting, thanks a lot @mtrmac

PR #1539, opened 2022-05-04
[release-5.19 backport] Bump github.com/containers/ocicrypt to 566b808

lsm5:
The two latest commits in c/ocicrypt get rid of OAEPDefaultHash envvar
and default to sha256.

They are waiting for confirmation on their rust side before cutting a
release. So, hoping to get this in assuming dependabot could take a
while.

@vrothberg @mtrmac @rhatdan PTAL. This is for subsequent vendoring into podman v4.0-rhel branch.

RE: https://github.com/containers/podman/pull/14102#issuecomment-1117204950

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/BurntSushi/toml v1.0.0
 	github.com/containerd/containerd v1.5.9 // indirect
 	github.com/containers/libtrust v0.0.0-20190913040956-14b96171aa3b
-	github.com/containers/ocicrypt v1.1.2
+	github.com/containers/ocicrypt v1.1.4
 	github.com/containers/storage v1.38.3
 	github.com/docker/distribution v2.7.1+incompatible
 	github.com/docker/docker v20.10.12+incompatible
@@ -39,6 +39,5 @@ require (
 	golang.org/x/crypto v0.0.0-20211215153901-e495a2d5b3d3
 	golang.org/x/net v0.0.0-20211112202133-69e39bad7dc2
 	golang.org/x/sync v0.0.0-20210220032951-036812b2e83c
-	golang.org/x/sys v0.0.0-20220114195835-da31bd327af9
 	golang.org/x/term v0.0.0-20201126162022-7de9c90e9dd1
 )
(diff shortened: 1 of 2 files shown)


mtrmac (approved):
LGTM.

lsm5:
Depends on https://github.com/containers/image/pull/1540

cevich:
> backporting that change as is is good enough, and probably worth doing

Ya, I generally avoid changing out VM images on release branches, and this specific repo. has extra complicating factors.  However there are sometimes exceptions.  If this branch is reasonably new and sufficiently related to the skopeo-repo-state where F36 was introduced, then yes.  Update the VM images here so they match the skopeo branch referenced for testing ([which is also currently/probably set wrong](https://github.com/containers/image/blob/release-5.19/.cirrus.yml#L8-L11)).

cevich:
> It’s about a month of difference so far

Oh okay, that's not super bad, maybe it can be done.

> the number of trivial changes being discovered one at a time is inching towards looking concerning.

Yes, there are differences in the images too (i.e. tooling script differences).  Another reason generally I say they only move  forward.  I do have a super-squirrel-secret way to recover images within a 29-59-day window, but if the images age past 60-days, they're irrecoverable.  I only mention that in case it's an option here, I haven't checked any CI-run dates on this branch.

go.mod
```diff
@@ -42,3 +42,5 @@ require (
 	golang.org/x/sys v0.0.0-20220114195835-da31bd327af9
 	golang.org/x/term v0.0.0-20201126162022-7de9c90e9dd1
 )
+
+replace github.com/containers/ocicrypt => github.com/containers/ocicrypt v1.1.4-0.20220428134531-566b808bdf6f
```
mtrmac, line 46: Is this line really necessary? I think just bumping the version above should be enough. (And it kind of _has_ to be enough because `replace` directives only work at the top level, i.e. this wouldn’t apply to a Skopeo/Podman build.)

lsm5: `go mod download` didn't let me fetch `v1.1.4` so I kept the line with the version-commit as-is and got rid of the replace directive. PTAL.

mtrmac: `go mod download` isn’t supposed to update `go.mod` I think…

FWIW `go get github.com/containers/ocicrypt@v1.1.4` worked for me, resulting in
```
github.com/containers/ocicrypt v1.1.4
```
. I didn’t realize this pre-release version actually points at a tagged version — in that case, I think using the 1.1.4 tag would be a bit safer WRT any other consumers of that.

mtrmac (approved):
LGTM.

Thanks for all the work to make this possible!

lsm5:
Thanks @mtrmac .

Btw, on rawhide, I got this:
```
$ go get github.com/containers/ocicrypt@v1.1.4
go: github.com/containers/ocicrypt@v1.1.4: invalid version: resolves to version v1.1.5-0.20220428134531-566b808bdf6f (v1.1.4 is not a tag)
```

But then I simply edited go.mod to have `v1.1.4` against ocicrypt, followed by only running `go mod tidy`  (no `go get`), and seems to work well.

Maybe that's a new rawhide change, anyway PTAL at the new change.

mtrmac:
Or it could be something with GOPROXY, or a pre-existing local download? I find the tool’s behavior a bit hard to understand sometimes.

lsm5:
> Or it could be something with GOPROXY, or a pre-existing local download? I find the tool’s behavior a bit hard to understand sometimes.

ack, I'll look further. Thanks a lot for the merge!!

PR #1540, opened 2022-05-04
[release-5.19 backport] Cirrus: Update to F36 VMs

lsm5:
(cherry picked from commit 90eec841dfd667eaec52ab1ec1e5d8aef8f570e5)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ export GOPROXY=https://proxy.golang.org
 
 # Which github repository and branch to use for testing with skopeo
 SKOPEO_REPO = containers/skopeo
-SKOPEO_BRANCH ?= master
+SKOPEO_BRANCH ?= main
 # Set SUDO=sudo to run container integration tests using sudo.
 SUDO =
 
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -8,7 +8,7 @@ env:
     # Name of the ultimate destination branch for this CI run
     DEST_BRANCH: "main"
     # CI container image tag (c/skopeo branch name)
-    SKOPEO_CI_TAG: "main"
+    SKOPEO_CI_TAG: "release-1.6"
     # Use GO module mirror (reason unknown, travis did it this way)
     GOPROXY: https://proxy.golang.org
     # Overrides default location (/tmp/cirrus) for repo clone
@@ -32,12 +32,12 @@ env:
     ####
     #### Cache-image names to test with (double-quotes around names are critical)
     ####
-    FEDORA_NAME: "fedora-35"
-    PRIOR_FEDORA_NAME: "fedora-34"
+    FEDORA_NAME: "fedora-36"
+    PRIOR_FEDORA_NAME: "fedora-35"
     UBUNTU_NAME: "ubuntu-2110"
 
     # Google-cloud VM Images
-    IMAGE_SUFFIX: "c6226133906620416"
+    IMAGE_SUFFIX: "c4955393725038592"
     FEDORA_CACHE_IMAGE_NAME: "fedora-${IMAGE_SUFFIX}"
     PRIOR_FEDORA_CACHE_IMAGE_NAME: "prior-fedora-${IMAGE_SUFFIX}"
     UBUNTU_CACHE_IMAGE_NAME: "ubuntu-${IMAGE_SUFFIX}"
--- a/contrib/cirrus/runner.sh
+++ b/contrib/cirrus/runner.sh
@@ -17,12 +17,6 @@ else
     ) > /dev/stderr
 fi
 
-OS_RELEASE_ID="$(source /etc/os-release; echo $ID)"
-# GCE image-name compatible string representation of distribution _major_ version
-OS_RELEASE_VER="$(source /etc/os-release; echo $VERSION_ID | tr -d '.')"
-# Combined to ease some usage
-OS_REL_VER="${OS_RELEASE_ID}-${OS_RELEASE_VER}"
-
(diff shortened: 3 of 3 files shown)


lsm5:
```
CGO_CFLAGS="" CGO_LDFLAGS="-L/usr/lib64 -lgpgme" GO111MODULE=on go build -mod=vendor "-buildmode=pie" -ldflags '-X main.gitCommit=1fe8da63a9ed49de41e8c185e6c9970fafdb13dc ' -gcflags "" -tags "btrfs_noversion libdm_no_deferred_remove" -o bin/skopeo ./cmd/skopeo
# github.com/containers/skopeo/cmd/skopeo
cmd/skopeo/copy.go:249:4: unknown field 'SignIdentity' in struct literal of type copy.Options
make: *** [Makefile:140: bin/skopeo] Error 2

Exit status: 2
```

does this need a change in the skopeo version used in cirrus?

mtrmac:
Yes — update `SKOPEO_BRANCH`.

Makefile
```diff
@@ -4,7 +4,7 @@ export GOPROXY=https://proxy.golang.org
 
 # Which github repository and branch to use for testing with skopeo
 SKOPEO_REPO = containers/skopeo
-SKOPEO_BRANCH ?= master
+SKOPEO_BRANCH ?= main
```
mtrmac, line 7: Uh, that was apparently left over — I guess we were saved by a redirect.

But in this case the 5.19 version of c/image needs to be testing with a 5.19-using version of Skopeo, something like `v1.6.2.

lsm5: hmm, so IIUC, that would be handled in `SKOPEO_CIDEV_CONTAINER_FQIN: "quay.io/libpod/skopeo_cidev:${IMAGE_SUFFIX}"` ?? I see that the IMAGE_SUFFIX is a quay.io repo tag and I don't think there's an easy mapping of repo tag <-> skopeo version, or am I looking in the wrong place?

/cc @cevich

mtrmac: Oops, my mistake, `SKOPEO_REPO` and `SKOPEO_BRANCH` seems to be completely unused now.

The right variable is `SKOPEO_CI_TAG` in `.cirrus.yml`; that’s consumed in `contrib/cirrus/runner.sh` to check out the relevant Skopeo version.

mtrmac: Those unused variables will be fixed on the main branch by https://github.com/containers/image/pull/1541.

mtrmac:
> sed: can't read /etc/containers/storage.conf: No such file or directory

… So Skopeo will need (some parts of) https://github.com/containers/skopeo/pull , backported to the relevant branch.

The perils of being the first change on an older branch that is not being regularly tested…

mtrmac:
@lsm5 https://github.com/containers/skopeo/pull/1637 should help with the `storage.conf` failure; then _this_ PR should change again, to target Skopeo’s (unreleased) `release-1.6` branch, rather than the specific `v1.6.2` tag.

Makefile
cevich: Ya, I think you're on the right track.  There is no hard correlation between branch, tag, and $IMAGE_SUFFIX.  So long as the `$IMAGE_SUFFIX` is consistent between VMs and container images, as was tested when the `$IMAGE_SUFFIX`.  The tricky thing for release branches is, we cannot easily go back and build new "old" images (for example if they were to get pruned due to disuse).  Hopefully that makes sense.

cevich:
> The perils of being the first change on an older branch that is not being regularly tested…

Yep, you said it correctly :disappointed:   I'm seriously thinking this may need to change.  In recent memory, I think this is the third or fourth time there's been a problem(?).  While I can't do anything about old images that were already pruned. If there are release branches we should preserve going-forward (i.e. with "not yet dead" VM images), we can do something about that.  Just send me the branch names and/or [add them into the cirrus-cron schedule yourself (config page)](https://cirrus-ci.com/settings/repository/6429752730058752), the [docs are short and helpful too](https://cirrus-ci.org/guide/writing-tasks/#cron-builds).

mtrmac:
> While I can't do anything about old images that were already pruned. If there are release branches we should preserve going-forward (i.e. with "not yet dead" VM images), we can do something about that.

If you asked _me_ yesterday, I wouldn’t have, from memory, pointed at this one as necessary…

Ideally I’d dream of RHEL needs driving this, so that every time we make a release that could plausibly need back ported fixes, we immediately branch _and_ mark the CI infrastructure for preservation (and periodic CI validation tests?). Assuming we want to pay for so much infrastructure, that is — IIRC that was a concern at some point.

cevich:
> and periodic CI validation tests?

Technically the tests don't really even need to run.  It's that "meta" task that's the important thing.  At least that needs to run probably weekly, and be monitored (so we know if it stops running).  I'm not worried about the infrastructure or cost (pennies) at all if it's important/critical to RHEL.  It is an extra maintenance burden, but sometimes working CI for backports is necessary and  important.  I'm simply not qualified to be the judge of that.

Adding the e-mail monitoring is a pain, but it's a one-time up-front time-cost.
Adding cron-jobs for new branches is trivial.  It's just an entry in that web-form I linked.

lsm5:
@mtrmac PTAL.

lsm5:
ugh ..i thought i saw all green, nvm, looking ..

mtrmac:
@lsm5 The keyring thing is probably fixed by backporting https://github.com/containers/image/pull/1477 . The other failure looks like an infrastructure flake.

lsm5:
Closing this in favour of https://github.com/containers/image/pull/1542

PR #1544, opened 2022-05-08
Config files should live in /usr/local on FreeBSD

dfr:
Signed-off-by: Doug Rabson <dfr@rabson.org>

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -61,8 +61,8 @@ type certPath struct {
 var (
 	homeCertDir     = filepath.FromSlash(".config/containers/certs.d")
 	perHostCertDirs = []certPath{
-		{path: "/etc/containers/certs.d", absolute: true},
-		{path: "/etc/docker/certs.d", absolute: true},
+		{path: etcDir + "/containers/certs.d", absolute: true},
+		{path: etcDir + "/docker/certs.d", absolute: true},
 	}
 
 	defaultUserAgent = "containers/" + version.Version + " (github.com/containers/image)"
--- a/docker/lookaside.go
+++ b/docker/lookaside.go
@@ -25,7 +25,7 @@ var systemRegistriesDirPath = builtinRegistriesDirPath
 
 // builtinRegistriesDirPath is the path to registries.d.
 // DO NOT change this, instead see systemRegistriesDirPath above.
-const builtinRegistriesDirPath = "/etc/containers/registries.d"
+const builtinRegistriesDirPath = etcDir + "/containers/registries.d"
 
 // userRegistriesDirPath is the path to the per user registries.d.
 var userRegistriesDir = filepath.FromSlash(".config/containers/registries.d")
--- a/docker/paths_common.go
+++ b/docker/paths_common.go
@@ -1,0 +1,6 @@
+//go:build !freebsd
+// +build !freebsd
+
+package docker
+
+const etcDir = "/etc"
--- a/docker/paths_freebsd.go
+++ b/docker/paths_freebsd.go
@@ -1,0 +1,6 @@
+//go:build freebsd
+// +build freebsd
+
+package docker
+
+const etcDir = "/usr/local/etc"
--- a/pkg/sysregistriesv2/paths_common.go
+++ b/pkg/sysregistriesv2/paths_common.go
@@ -1,0 +1,12 @@
+//go:build !freebsd
+// +build !freebsd
+
+package sysregistriesv2
+
+// builtinRegistriesConfPath is the path to the registry configuration file.
+// DO NOT change this, instead see systemRegistriesConfPath above.
+const builtinRegistriesConfPath = "/etc/containers/registries.conf"
+
+// builtinRegistriesConfDirPath is the path to the registry configuration directory.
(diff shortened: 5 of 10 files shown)


pkg/sysregistriesv2/system_registries_v2_darwin.go
```diff
@@ -0,0 +1,9 @@
+package sysregistriesv2
```
vrothberg, line 1: Could we use build conditionals instead of file suffixes?

Two files should be sufficient, one with `// +build freebsd` the other one with `!freebsd`.

signature/policy_config_darwin.go
```diff
@@ -0,0 +1,5 @@
+package signature
+
```
vrothberg, line 2: Same here re: build conditionals.

vrothberg (commented):
Thanks for contributing!

pkg/sysregistriesv2/system_registries_v2_common.go
```diff
@@ -0,0 +1,12 @@
+//go:build !freebsd
```
mtrmac, line 1: (Absolutely non-blocking: Naming this `pkg/sysregistriesv2/paths_{common,freebsd}.go` would work just as well, and be a tiny bit more descriptive.)

signature/policy_config_common.go
```diff
@@ -0,0 +1,8 @@
+//go:build !freebsd
```
mtrmac, line 1: (Absolutely non-blocking: Naming this `pkg/sysregistriesv2/policy_paths_{common,freebsd}.go` would work just as well, and be a tiny bit more descriptive.)

mtrmac (commented):
There are three other `/etc` paths in `c/image/docker`; should those be updated as well?

---

Compare https://github.com/containers/skopeo/pull/1298#issuecomment-849529062 / https://github.com/containers/skopeo/pull/1507 . Changing the default in c/image instead of Skopeo is definitely better, still…

- Changing the default is going to immediately break Skopeo’s `Makefile` that installs config files to `/etc`.
- How are upgrades going to work? Why isn’t this going to, potentially silently, ignore existing user’s configuration (with security implications like ignoring `blocked` registry configurations)? At least for users that used to compile themselves using e.g. Skopeo’s `Makefile`, this seems to me to be an unavoidable compatibility break. Or is there some reason why that isn’t actually possible, so it isn’t relevant?

dfr:
> There are three other `/etc` paths in `c/image/docker`; should those be updated as well?

I don't think there is a fully functioning docker for FreeBSD at this point. I know some people were working on porting moby but I'm not sure of the status there.

> 
> Compare [containers/skopeo#1298 (comment)](https://github.com/containers/skopeo/pull/1298#issuecomment-849529062) / [containers/skopeo#1507](https://github.com/containers/skopeo/pull/1507) . Changing the default in c/image instead of Skopeo is definitely better, still…
> 
> * Changing the default is going to immediately break Skopeo’s `Makefile` that installs config files to `/etc`.
> * How are upgrades going to work? Why isn’t this going to, potentially silently, ignore existing user’s configuration (with security implications like ignoring `blocked` registry configurations)? At least for users that used to compile themselves using e.g. Skopeo’s `Makefile`, this seems to me to be an unavoidable compatibility break. Or is there some reason why that isn’t actually possible, so it isn’t relevant?

This is a good point but perhaps not a big issue - I think there are only a handful of people using skopeo on FreeBSD at this point since its not available as a pre-build package. I can follow up with a PR on skopeo to install to the correct path.

mtrmac:
> > * How are upgrades going to work? Why isn’t this going to, potentially silently, ignore existing user’s configuration (with security implications like ignoring `blocked` registry configurations)? At least for users that used to compile themselves using e.g. Skopeo’s `Makefile`, this seems to me to be an unavoidable compatibility break. Or is there some reason why that isn’t actually possible, so it isn’t relevant?
> 
> This is a good point but perhaps not a big issue - I think there are only a handful of people using skopeo on FreeBSD at this point since its not available as a pre-build package

I am unhappy about this in principle, but ultimately I do think it’s appropriate for FreeBSD contributors to decide on the FreeBSD tradeoffs…

RFC @mateuszkwiatkowski (also @vrothberg @lsm5 )

mtrmac:
> Also, do we need something FreeBSD specific in CI going forrward?

At least to the extent Red Hat is paying for the CI, I wouldn’t say we _need_ FreeBSD coverage. Still, it’s a good idea to explore this, probably in a separate PR. (I guess adding FreeBSD to the `cross` test would be a cheap and easy smoke test. Modifying all of Skopeo’s integration/system testing to run on FreeBSD seems intuitively hard, but I know very little about FreeBSD capabilities.)

rhatdan:
LGTM

mtrmac:
Thanks again!

mtrmac:
> Are there release notes for skopeo releases? Perhaps the change can be mentioned there.

GitHub does have a release notes mechanism. We don’t currently have automated mechanism to collect/create them for c/image or Skopeo. I have, at least, re-titled the PRs; hopefully that helps us remember when the time comes.

PR #1548, opened 2022-05-11
docker: add workaround for CloudFront

giuseppe:
CloudFront returns an unquoted ':' in the boundary, special handle it.

It is needed for partial pulls from quay.io.
    
### DIFF
--- a/docker/docker_image_src.go
+++ b/docker/docker_image_src.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


docker/docker_image_src.go
```diff
@@ -399,7 +405,22 @@ func (s *dockerImageSource) GetBlobAt(ctx context.Context, info types.BlobInfo,
 	case http.StatusPartialContent:
 		mediaType, params, err := mime.ParseMediaType(res.Header.Get("Content-Type"))
 		if err != nil {
-			return nil, nil, err
+			if err == mime.ErrInvalidMediaParameter {
+				// CloudFront returns an invalid MIME type, that contains an unquoted ":", let's
+				// handle it here.
+				re := regexp.MustCompile("multipart/byteranges; boundary=([A-Za-z-0-9:]+)")
```
mtrmac, line 411: This can be compiled just once per process lifetime, can’t it?

docker/docker_image_src.go
```diff
@@ -399,7 +405,22 @@ func (s *dockerImageSource) GetBlobAt(ctx context.Context, info types.BlobInfo,
 	case http.StatusPartialContent:
 		mediaType, params, err := mime.ParseMediaType(res.Header.Get("Content-Type"))
 		if err != nil {
-			return nil, nil, err
+			if err == mime.ErrInvalidMediaParameter {
+				// CloudFront returns an invalid MIME type, that contains an unquoted ":", let's
```
mtrmac, line 409: I don’t quite see how this comment `:` relates to the code, but, I guess we’ll see from the test case…

Do you mean that it is invalid to have a `;` in the _boundary_? There is no `:` in `multipart/byteranges`.

mtrmac (commented):
Thanks.

Most importantly, please add a unit test, ideally with an unmodified response of the potential server. (If that is easier when splitting the content-type parser into a separate function, sure, why not.)

We’ve already been in a situation where we have a workaround and no way to maintain it because we don’t know what the workaround is looking for how what are the essential parts, let’s not add any such situations.

docker/docker_image_src.go
mtrmac: Ah, the PR description does say “boundary”.

giuseppe:
@mtrmac thanks for the review.  I addressed your comments in the last version :arrow_up:

rhatdan:
LGTM

mtrmac (commented):
LGTM. Thanks again!

PR #1555, opened 2022-05-19
Stop calling gpgme-config

mtrmac:
As of the just-updated `github.com/proglottis/gpgme` 0.1.2, the `gpgme` subpackage uses CGo’s native `#cgo pkg-config` support to find the relevant libraries, and we no longer need to manually set `CGO_CFLAGS` and `CGO_LDFLAGS`. So stop doing that.

Note that the `proglottis/gpgme` update (already before this commit) means the minimal supported version of GPGME is 1.13.0.


rhatdan:
LGTM

PR #1559, opened 2022-05-26
Make sure github.com/opencontainers/runc >= 1.1.2 is used

mtrmac:
... to silence Dependabot alerts about CVE-2022-29162 = GHSA-f3fp-gc8g-vw66.

Note that nothing in c/image actually calls this package, so the vulnerability could only be triggered by some other code path.


rhatdan:
LGTM
@vrothberg PTAL

vrothberg (approved):
LGTM, thanks

PR #1568, opened 2022-06-06
Update github.com/containers/ocicrypt

mtrmac:
> go get github.com/containers/ocicrypt@latest

It seems Dependabot doesn't update from commit references.


rhatdan:
LGTM

PR #1571, opened 2022-06-08
Update to benefit from Go 1.17

mtrmac:
Per https://github.com/containers/image/pull/1570#issuecomment-1149876963 , Go 1.17 will be required; so, update c/image to benefit from it.

### DIFF
--- a/pkg/docker/config/config_test.go
+++ b/pkg/docker/config/config_test.go
@@ -25,18 +25,7 @@ func TestGetPathToAuth(t *testing.T) {
 
 	tmpDir := t.TempDir()
 
-	// Environment is per-process, so this looks very unsafe; actually it seems fine because tests are not
-	// run in parallel unless they opt in by calling t.Parallel().  So don’t do that.
-	oldXRD, hasXRD := os.LookupEnv("XDG_RUNTIME_DIR")
-	defer func() {
-		if hasXRD {
-			os.Setenv("XDG_RUNTIME_DIR", oldXRD)
-		} else {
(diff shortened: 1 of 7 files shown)


TomSweeneyRedHat:
LGTM

pkg/docker/config/config_test.go
```diff
@@ -61,40 +50,35 @@ func TestGetPathToAuth(t *testing.T) {
...
-		if c.expected == "" {
-			assert.Error(t, err)
-		} else {
-			require.NoError(t, err)
-			assert.Equal(t, c.expected, res)
-			assert.Equal(t, c.legacyFormat, lf)
-		}
+		t.Run(fmt.Sprintf("%d", caseIndex), func(t *testing.T) {
+			t.Setenv("XDG_RUNTIME_DIR", c.xrd)
+			if c.xrd == "" {
```
rhatdan, line 55: This seems a little weird, why not only set XDG_RUNTIME_DIR if c.xrd is set? Unless you are worried about XDG_RUNTIME_DIR already being set.

mtrmac: Thanks for pointing this out, this really deserves a comment. Updated now, in both places.

(`t.Setenv`, unlike `os.Setenv`, ensures the environment variable is always restored to the original value (or original unset state) after the test finishes; and there is no `t.Unsetenv`. So we call `t.Setenv` unconditionally to get get the cleanup guarantee, and only then worry about possibly unsetting the variable if wanted by the test.)

rhatdan:
LGTM

Issue #1150, opened 2021-02-16
Generic OCI artifact transport?


cfiderer:
Hello all,
 I need to transfer non-image artifacts (that adhere to the OCI Artifact spec) between repositories and the file system (repo -> repo, repo -> file system, file system -> repo), I thought about using the `containers/image` library.
But unfortunately, the `docker://` prefix is the only prefix that I've found for repositories (on the network), and the code related to Docker repositories handles only Docker images (and refuses everything else).
I'd like to ask:
- Are there other transports for OCI artifacts that I've overlooked until now?
- Are there projects on the way that address this feature?

 Kind regards, Christian

vrothberg:
Thanks for reaching out!

Did you try copying an artifact that failed?  `containers/image` is OCI conformant and hence supports artifacts. The transport is called `docker://` for historical reasons.  We are considering to alias `registry://` to it to avoid confusion. That being said, `docker://` supports images in the OCI format and Docker v2 schema 1 and schema 2.

cfiderer:
Hi Valentin,

thank you for your reply. The artifacts that I am talking about are not (OCI or Docker) images, but configuration stuff that adheres to the [Artifact Authors Guidance](https://github.com/opencontainers/artifacts/blob/master/artifact-authors.md).

When I try to inspect such an artifact, I get an `unsupported docker v2s2 media type:` error from the `manifest.SupportedSchema2MediaType()` function. And yes, our layer.mediaType is not one of the known media types for Docker images.

``` bash
$ skopeo inspect docker://eu.gcr.io/gardener-project/cc/cnudie-test-3/component-descriptors/github.com/gardener/cc-utils@sha256:95df958f73f74e81e14ce83d199e4f3a29dd2957e832c7971a28b302c0b595f4
time="2021-02-17T12:32:35+01:00" level=fatal msg="Error parsing manifest for image: unsupported docker v2s2 media type: \"application/vnd.oci.gardener.cloud.cnudie.component-descriptor.config.v2+yaml+tar\""
```

A sample manifest is `eu.gcr.io/v2/gardener-project/cc/cnudie-test-3/component-descriptors/github.com/gardener/cc-utils/manifests/sha256:95df958f73f74e81e14ce83d199e4f3a29dd2957e832c7971a28b302c0b595f4` (a bit redacted for readability):

``` bash
[...]

Issue #1370, opened 2020-05-27
Support other OCI artifact types


gaocegege:
> skopeo can work with OCI images as well as the original Docker v2 images.

It will be great if we could support Helm Chart, CNAB and other [OCI artifact](https://github.com/opencontainers/artifacts) types.

nevermore-muyi:
Support other OCI types are very useful.

bt-lemery:
A Harbor user here who would also find this very useful.

mtrmac:
With #1574 , c/image should now support copying OCI artifacts. Some operations, like inspecting an image or format conversions, are rejected.

Note that the only _specific_ example of a failure in this issue, https://github.com/containers/image/issues/1370#issuecomment-705125081 , was referring to a non-OCI manifest/index type. Those objects are not OCI artifacts, and are not enabled by #1574 .

PR #1574, opened 2022-06-11
Support OCI artifacts

mtrmac:
This broadly follows the plan from https://github.com/containers/image/pull/1408#issuecomment-1029487477 :

- Reject image-specific operations (like inspect, image ID) on OCI artifacts
- Don’t change compression of artifact layers
- FIXME: Should we also reject encryption/decryption operations?

See the individual commits for details.

FIXME: At this point I’m rather unhappy about the `CanChangeLayerCompression` API. It’s good enough for internal use for now, exposing it in `c/image/manifest` is not great. OTOH I’d rather not spend time investigating the more general which-image-formats-can-support-which-compression-and-what-to-do-if-they-don’t problem space.

At this point absolutely untested apart from unit tests (which don’t cover most of the `copy` code); this will be the first run against Skopeo’s integration test and a real artifact.

See individual commit messages for details. Quite a lot of the refactoring could be split into a separate PR… if an existing OCI artifact didn’t break CI right now. Something to consider later.

### DIFF
--- a/internal/manifest/errors.go
+++ b/internal/manifest/errors.go
@@ -1,0 +1,32 @@
+package manifest
+
+import "fmt"
+
+// NonImageArtifactError (detected via errors.As) is used when asking for an image-specific operation
+// on an object which is not a “container image” in the standard sense (e.g. an OCI artifact)
+//
+// This is publicly visible as c/image/manifest.NonImageArtifactError (but we don’t provide a public constructor)
+type NonImageArtifactError struct {
+	// Callers should not be blindly calling image-specific operations and only checking MIME types
+	// on failure; if they care about the artifact type, they should check before using it.
+	// If they blindly assume an image, they don’t really need this value; just a type check
+	// is sufficient for basic "we can only pull images" UI.
+	//
+	// Also, there are fairly widespread “artifacts” which nevertheless use imgspecv1.MediaTypeImageConfig,
+	// e.g. https://github.com/sigstore/cosign/blob/main/specs/SIGNATURE_SPEC.md , which could cause the callers
+	// to complain about a non-image artifact with the correct MIME type; we should probably add some other kind of
+	// type discrimination, _and_ somehow make it available in the API, if we expect API callers to make decisions
+	// based on that kind of data.
+	//
+	// So, let’s not expose this until a specific need is identified.
+	mimeType string
+}
+
+// NewNonImageArtifactError returns a NonImageArtifactError about an artifact with mimeType.
+func NewNonImageArtifactError(mimeType string) error {
+	return NonImageArtifactError{mimeType: mimeType}
+}
+
+func (e NonImageArtifactError) Error() string {
+	return fmt.Sprintf("unsupported image-specific operation on artifact with type %q", e.mimeType)
+}
--- a/copy/blob.go
+++ b/copy/blob.go
@@ -18,8 +18,6 @@ import (
 func (ic *imageCopier) copyBlobFromStream(ctx context.Context, srcReader io.Reader, srcInfo types.BlobInfo,
 	getOriginalLayerCopyWriter func(decompressor compressiontypes.DecompressorFunc) io.Writer,
 	isConfig bool, toEncrypt bool, bar *progressBar, layerIndex int, emptyLayer bool) (types.BlobInfo, error) {
-	canModifyBlob := !isConfig && ic.cannotModifyManifestReason == ""
-
 	// The copying happens through a pipeline of connected io.Readers;
 	// that pipeline is built by updating stream.
 	// === Input: srcReader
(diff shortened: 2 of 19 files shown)


mtrmac:
Note to self: See whether this resolves _all_ of #1150, #1370, #1408 .

mtrmac:
Parts of this have now been split into #1576 and #1577.

internal/manifest/errors.go
```diff
@@ -0,0 +1,25 @@
...
+
+import "fmt"
+
+// NonImageArtifactError (detected via errors.As) is used when asking for an image-specific operation
+// on an object which is not a “container image” in the standard sense (e.g. an OCI artifact)
+//
+// This is publicly visible as c/image/manifest.NonImageArtifactError (but we don’t provide a public constructor)
+type NonImageArtifactError struct {
+	// Callers should not be blindly calling image-specific operations and only checking MIME types
+	// on failure; if they care about the artifact type, they should check before using it.
```
vrothberg, line 11: Note that it's non-trivial to figure out whether an image is an artifact or not.  An artifact may have a custom config and/or custom layers.

vrothberg: Maybe we can help callers in figuring that out.

vrothberg (commented):
LGTM, nice work

internal/manifest/errors.go
mtrmac: https://github.com/opencontainers/artifacts/blob/main/artifact-authors.md#visualizing-artifacts explicitly says that the `config.mediaType` is how they are supposed to be differentiated.

But then _both_ the primary motivations for this work are not compliant: #1408 wants to use images with custom layers, and the https://github.com/sigstore/cosign/blob/main/specs/SIGNATURE_SPEC.md explicitly defines the signature to use an image MIME type.

In the cases where this PR adds this failure mode, we really care about the thing being an _image_ (e.g. we inspect the config or even convert the config’s format). In cases of abuse of that MIME type, it seems … technically correct… to complain about an “invalid image” more than about an “unexpected artifact”.

But you’re completely right, in practice it might be much more practical to detect the major violators and provide good error messages for them. That’s actually a pretty good reason not to expose the `mimeType` value of the error (which I was quite on the fence about) until we can provide something useful to callers.

mtrmac: https://github.com/opencontainers/artifacts/blob/main/artifact-authors.md#visualizing-artifacts explicitly says that the `config.mediaType` is how they are supposed to be differentiated.

But then _both_ the primary motivations for this work are not compliant: #1408 wants to use images with custom layers, and the https://github.com/sigstore/cosign/blob/main/specs/SIGNATURE_SPEC.md explicitly defines the signature to use an image MIME type.

In the cases where this PR adds this failure mode, we really care about the thing being an _image_ (e.g. we inspect the config or even convert the config’s format). In cases of abuse of that MIME type, it seems … technically correct… to complain about an “invalid image” more than about an “unexpected artifact”.

But you’re completely right, in practice it might be much more practical to detect the major violators and provide good error messages for them. That’s actually a pretty good reason not to expose the `mimeType` value of the error (which I was quite on the fence about) until we can provide something useful to callers.

mtrmac:
- Fixes #1408 : Layers in non-image OCI artifacts, and layers in images with unknown MIME types, are now never compressed or decompressed, and we don’t update their MIME types even if we detect a specific compression format.
- Fixes #1150: (Parts of that report are an artifact-like object with v2s2 schema, which is invalid.) Same as the above, the copy was failing on trying to compress the layers, and that’s now handled correctly.
- #1370 is unclear. (WRT the conversation in that issue, this PR correctly refuses conversions from OCI artifacts to other non-artifact image types, and fails on attempts to query image data on non-image artifacts.) But the original request was for `oci/layout` to somehow support directories where the top level is nether an OCI index nor an OCI manifest — those are _not_ OCI artifacts and this PR doesn’t add support.

mtrmac:
OK, I’m done with this.

One of the intermediate versions (motivating #1579 ) added a `SourcedImage` method to check whether the _current_ manifest format supports the chosen compression algorithm. That’s incorrect, because we can convert v2s2 to OCI and compress using Zstd (not supported in v2s2) in one step, i.e. that depends on the destination manifest format, not on the source one. Extending the logic for that is certainly possible in principle (ideally already in `determineManifestConversion`?), but that’s really a different feature from “don’t break when syncing repos with OCI artifacts”, and a rather lower-priority one.

So, this PR does not handle that any more — and now we need only a single new `SourcedImage` method instead of three in the last prototype, which  simpler and less baggage to carry around if we ended up needing a completely different design.

---

This also updates the CI image to a newer registry that can accept OCI artifacts, although we don’t currently test with any.

Then, https://github.com/containers/skopeo/pull/1680 re-enables tests of syncing the repo with a Cosign signature; see that for at least a smoke test of this PR.

mtrmac:
> This also updates the CI image to a newer registry that can accept OCI artifacts, although we don’t currently test with any.

Oh, actually, that requires https://github.com/containers/skopeo/pull/1679 . Dropping that commit again.

So I think the merging order should be:
- This PR, and https://github.com/containers/skopeo/pull/1679 , in either order.
- After both are in, the revert part of https://github.com/containers/skopeo/pull/1680 .

vrothberg (approved):
LGTM

mtrmac:
> So I think the merging order should be:
> 
> * This PR, and [Use an updated CI image with OCI-capable registry skopeo#1679](https://github.com/containers/skopeo/pull/1679) , in either order.

Here, c/image needs to switch to the updated CI image here as well, or c/image CI is going to break: https://github.com/containers/image/pull/1581 .

> * After both are in, the revert part of [DO NOT MERGE: Add OCI artifact support, test syncing Cosign signatures again skopeo#1680](https://github.com/containers/skopeo/pull/1680) .

PR #1576, opened 2022-06-11
Make c/image/image available as c/image/internal/image

mtrmac:
This is a part of #1574, per https://github.com/containers/image/pull/1408#issuecomment-1029487477 :
- Move the implementation of c/image/image to c/image/internal/image, so that we can adjust the API without affecting external callers (notably without changing `types.Image`). Keep c/image/image with trivial forwarders, for compatibility.
- Add `image.FromReference`, use it to simplify transports’ implementation of `NewImage` (and to fix a bug in most of them).


rhatdan:
LGTM
@vrothberg PTAL

vrothberg (commented):
LGTM, nice work!

PR #1577, opened 2022-06-11
Consolidate interface mocks in a testing subpackage

mtrmac:
This is a part of #1574 .

<s>This currently depends on #1576 for the file moves that have happened there, but it could just as well be merged in the other order — let me know which one you prefer.</s>

Various tests set up “implementations” of `types.Image*` interfaces that just `panic` when called, or `panic` on all but a few methods. Consolidate that into `internal/testing/mocks.Forbidden*` implementations that just `panic`, and let the tests only implement the methods they care.

Only modifies tests, should not change behavior.


mtrmac:
Rebased and ready for review.

rhatdan:
LGTM
@vrothberg PTAL

vrothberg (approved):
LGTM

PR #1578, opened 2022-06-14
Use a SourcedImage struct in c/image/copy

mtrmac:
Another part of #1574 (but note that this approach is rather different from an earlier version):

Use an `internal/image.SourcedImage` directly instead of a `types.Image` interface for most of c/image/copy. The implementation/tests of `copy.determineManifestConversion` currently mock the `types.Image` interface; instead, turn `determineManifestConversion` into a ~pure function, with all inputs and outputs as fields in a struct.

Also add the first benefit of using the non-public `SourcedImage`: access the manifest and manifest’s MIME type directly without having to worry about (impossible) failures.

Should not change behavior overall. See individual commit messages for details.

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -82,7 +82,7 @@ type copier struct {
 type imageCopier struct {
 	c                          *copier
 	manifestUpdates            *types.ManifestUpdateOptions
-	src                        types.Image
+	src                        *image.SourcedImage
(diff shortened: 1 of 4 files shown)


vrothberg (approved):
LGTM
merge at will

rhatdan:
LGTM

PR #1590, opened 2022-07-02
Implement private.ImageSource in all transports, improve transport helpers

mtrmac:
This is the `ImageSource` counterpart to #1589 , pretty much the same things.

It is conceptually independent from it, but it splits `storage_image.go` and `openshift.go` into three files, and thus depends on pretty much anything that could touch it, so for now it includes #1589. I can rebase it to be independent, but either way I’d prefer #1589 to be reviewed first, with the design discussed there. I can then adjust this PR correspondingly.


vrothberg (commented):
LGTM

Feel free to merge at will.

PR #1592, opened 2022-07-05
Implement private.ImageSource / ImageDestination in blobcache

mtrmac:
… so that we can extend the signature representation (and possibly add other features), which is why we moved `pkgblobcache` to c/image in the first place.


vrothberg (approved):
LGTM

PR #1595, opened 2022-07-06
Add support for reading and writing Cosign attachments, to c/image/docker

mtrmac:
**⚠️ Warning:** This is write-only code, as in I haven’t read it after myself, and it has no tests yet. Might be completely broken. It really needs an integration test, and interoperability testing.

**Note:** design decisions:
- We can read Cosign data from lookaside
- We _only_ write Cosign data to Cosign attachments, never to lookaside; because lookaside is set up by default, that
      would be too confusing.
- We _only_ use Cosign attachments at all if the user opts in via `registries.d`, option `use-cosign-attachments`.
    
  One concern is performance impact of the extra round-trip for large-scale operations like (skopeo sync).
    
  Short-term, a much more worrying is the risk that we probably have the "is this failure just a missing atachment manifest, or a real failure reading it?" heuristic wrong, so without an opt-in, _all_ image reads are going to fail.  This might eventually go away after more testing.

Depends on unmerged #1594.


docker/docker_image_dest.go
```diff
@@ -596,6 +637,140 @@ func (d *dockerImageDestination) putOneSignature(url *url.URL, signature []byte)
...
+func (d *dockerImageDestination) putSignaturesToCosignAttachments(ctx context.Context, signatures []signature.Cosign, manifestDigest digest.Digest) error {
+	if !d.c.useCosignAttachments {
+		return errors.New("writing Cosign attachments is disabled by configuration")
+	}
+
+	ociManifest, err := d.c.getCosignAttachmentManifest(ctx, d.ref, manifestDigest)
+	if err != nil {
+		return nil
+	}
+	var ociConfig imgspecv1.Image // Most fields empty by default
```
mtrmac, line 649: TO DO: Only fetch the manifest if we actually have a new signature to add (and if we continue to update the DiffID list).

docker/docker_image_dest.go
```diff
@@ -596,6 +637,140 @@ func (d *dockerImageDestination) putOneSignature(url *url.URL, signature []byte)
...
+		if err != nil {
+			return err
+		}
+		sigDesc.Annotations = annotations
+		ociManifest.Layers = append(ociManifest.Layers, sigDesc)
+		ociConfig.RootFS.DiffIDs = append(ociConfig.RootFS.DiffIDs, sigDesc.Digest)
+		logrus.Debugf("Adding new signature, digest %s", sigDesc.Digest.String())
+	}
+
+	configBlob, err := json.Marshal(ociConfig)
```
mtrmac, line 705: TO DO: Only upload an updated config and manifest if we added at least one new signature.

docker/docker_client.go
```diff
@@ -801,6 +807,166 @@ func (c *dockerClient) detectProperties(ctx context.Context) error {
...
+	defer reader.Close()
+	payload, err := iolimits.ReadAtMost(reader, iolimits.MaxSignatureBodySize)
+	if err != nil {
+		return nil, fmt.Errorf("reading blob %s in %s: %w", desc.Digest.String(), ref.ref.Name(), err)
+	}
+	return payload, nil
+}
+
+// isManifestUnknownError returns true iff err from client.HandleErrorResponse is a “manifest unknown” error.
+func isManifestUnknownError(err error) bool {
```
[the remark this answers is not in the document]
mtrmac: Fixed

mtrmac:
> So, calling this ready for review.

Note the most important risk mitigation: The code reading/writing Cosign attachments is opt-in, so it hopefully won’t break another users.

vrothberg (commented):
> One concern is performance impact of the extra round-trip
  for large-scale operations like (skopeo sync).

>  Short-term, a much more worrying is the risk that we probably
  have the "is this failure just a missing atachment manifest,
  or a real failure reading it?" heuristic wrong, so without an
  opt-in, _all_ image reads are going to fail.  This might eventually
  go away after more testing.
  
 Code and design LGTM, nice work!

vrothberg:
Feel free to merge

rhatdan:
LGTM

PR #1596, opened 2022-07-06
Add signature.untrustedCosignPayload, with JSON marshaling

mtrmac:
The code and tests are basically copy&pasted from simple signing.

Note that this just adds a _private_ type with no users, to be used by both the signing and verification implementations. So, do not merge until those exist at least in draft form.

(The paranoid JSON validation should be mostly contributed upstream to `sigstore/sigstore/pkg/signature/payload`.)

**Note:** the comments about `dockerReference`.


vrothberg (commented):
LGTM, merge at will

rhatdan:
LGTM

PR #1597, opened 2022-07-06
Cosign sign

mtrmac:
**⚠️ Warning:** This is write-only code, as in I haven’t read it after myself, it has never been run, and it has no tests yet. Might be completely broken. It really needs unit and an integration tests, and interoperability testing.

Private key only, no Fulcio/Rekor. Figuring out a sane UI for non-default Fulcio servers is quite a bit of more work than this (probably commits us to defining a new config file format).

Depends on unmerged #1594 and #1596.


signature/cosign/sign.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+	// AARGH. sigstore/cosign mostly ignores dockerReference for actual policy decisions.
+	// That’s bad enough, BUT they also:
+	// - Record the repo (but NOT THE TAG) in the value; without the tag we can’t detect version rollbacks.
+	// - parse dockerReference @ dockerManifestDigest and expect that to be valid.
+	// - It seems (FIXME: TEST THIS) that putting a repo:tag in would pass the current implementation.
+	//   And signing digest references is _possible_ but probably rare (because signing typically happens on push, when
+	//   the digest reference is not known in advance).
+	//   SO: We put the full value in, which is not interoperable for signed digest references right now,
+	//   and TODO: Talk sigstore/cosign to relax that.
```
mtrmac, line 51: For the record: There are two types (`payload.SimpleContainerImage` and `payload.Cosign`).

`payload.Cosign` is currently used to _create_ signature payloads. It has an `UnmarshalJSON` implementation which is, AFAICS, currently unused.

`payload.SimpleContainerImage` is the one used to validate signatures (via `cosign.SimpleClaimVerifier`). It does not impose any requirements on `dockerReference` at all. (`cosign.SimpleClaimVerifier` only verifies that the manifest digest matches, and optionally that user-specified optional fields match.)

mtrmac:
In manual testing, this produces signatures that Cosign accepts.

So, calling this ready for review. It could definitely do with unit tests, and integration tests in the Skopeo repo.

vrothberg (commented):
LGTM

rhatdan:
LGTM

PR #1598, opened 2022-07-06
DO NOT MERGE IRRESPONSIBLE: Cosign verify

mtrmac:
# 💀💀💀DO NOT MERGE THIS WITHOUT ADDING TEST COVERAGE

**⚠️ Warning:** This is write-only code, as in I haven’t read it after myself, it has never been run, and it has no tests yet. Might be completely broken. It really needs unit and an integration tests, and interoperability testing.

`type: cosignSigned`, with the usual `keyData`/`keyPath`. Fulcio/Rekor is plausible _for the off-line Rekor log entry proofs_, but not currently implemented. Tests first.
    
**Note:** This only allows a single public key, not a keyring, unlike simple signing. That seems problematic, there are known users of that. But we can fix that later by adding `keyDirectory` and the like.

Depends on unmerged #1594 and #1596.

PR #1599, opened 2022-07-06
DO NOT MERGE IRRESPONSIBLE: Cosign integration

mtrmac:
This is a single integration branch of the Cosign prototypes.

# 💀💀💀DO NOT MERGE THIS WITHOUT ADDING TEST COVERAGE

See #1598.

---

(Outstanding) merge/review order:
- Three independent parts:
  - #1595 
  - #1597
  - #1598


copy/copy.go
```diff
@@ -122,15 +123,17 @@ type ImageListSelection int
...
-	ReportWriter     io.Writer
-	SourceCtx        *types.SystemContext
-	DestinationCtx   *types.SystemContext
-	ProgressInterval time.Duration                 // time to wait between reports to signal the progress channel
-	Progress         chan types.ProgressProperties // Reported to when ProgressInterval has arrived for a single artifact+offset.
+	RemoveSignatures                 bool            // Remove any pre-existing signatures. SignBy will still add a new signature.
+	SignBy                           string          // If non-empty, asks for a signature to be added during the copy, and specifies a key ID, as accepted by signature.NewGPGSigningMechanism().SignDockerManifest(),
+	SignPassphrase                   string          // Passphare to use when signing with the key ID from `SignBy`.
+	SignByCosignPrivateKeyFile       string          // If non-empty, asks for a signature to be added during the copy, using a Cosign private key file at the provided path.
+	SignByCosignPrivateKeyPassphrase []byte          // Passphare to use when signing with `SignByCosignPrivateKeyFile`.
```
vrothberg, line 130:
```suggestion
	SignCosignPrivateKeyPassphrase []byte          // Passphare to use when signing with `SignByCosignPrivateKeyFile`.
```
Just to be consistent with `SignBy` and `SignPassphrase`

docker/docker_image_dest.go
```diff
@@ -596,6 +637,125 @@ func (d *dockerImageDestination) putOneSignature(url *url.URL, signature []byte)
...
+		sigDesc, err := d.putBlobBytesAsOCI(ctx, sig.UntrustedPayload, sig.UntrustedMIMEType, private.PutBlobOptions{
+			Cache:      none.NoCache, // FIXME: Pass BlobInfoCache to signature accessors? Or is this just pointless noise??
+			IsConfig:   false,
+			EmptyLayer: false,
+			LayerIndex: nil,
+		})
+		if err != nil {
+			return err
+		}
+		sigDesc.Annotations = sig.UntrustedAnnotations
```
vrothberg, line 689: I _think_ that using `sig` is dangerous.  If there are multiple signatures, wouldn't all users of `UntrustedAnnotations` point to the field of the last item in the slice?

docker/docker_image_dest.go
```diff
@@ -596,6 +637,125 @@ func (d *dockerImageDestination) putOneSignature(url *url.URL, signature []byte)
...
+				found = true
+				break
+			}
+		}
+		if found {
+			continue
+		}
+
+		sigDesc, err := d.putBlobBytesAsOCI(ctx, sig.UntrustedPayload, sig.UntrustedMIMEType, private.PutBlobOptions{
+			Cache:      none.NoCache, // FIXME: Pass BlobInfoCache to signature accessors? Or is this just pointless noise??
```
vrothberg, line 681: @mtrmac, do you think the blob cache could speed things up?

docker/docker_image_dest.go
```diff
@@ -596,6 +637,125 @@ func (d *dockerImageDestination) putOneSignature(url *url.URL, signature []byte)
...
+			return err
+		}
+		if err := json.Unmarshal(configBlob, &ociConfig); err != nil {
+			return fmt.Errorf("parsing Cosign config %s in %s: %w", ociManifest.Config.Digest.String(),
+				d.ref.ref.Name(), err)
+		}
+	}
+
+	for _, sig := range signatures {
+		found := false
```
vrothberg, line 669: Nit: can we rename `found` to something similar to `alreadyOnRegistry`?  It adds some more context that would help me brain-parse the code.

docker/docker_image_dest.go
```diff
@@ -596,6 +637,125 @@ func (d *dockerImageDestination) putOneSignature(url *url.URL, signature []byte)
...
+			IsConfig:   false,
+			EmptyLayer: false,
+			LayerIndex: nil,
+		})
+		if err != nil {
+			return err
+		}
+		sigDesc.Annotations = sig.UntrustedAnnotations
+		ociManifest.Layers = append(ociManifest.Layers, sigDesc)
+		ociConfig.RootFS.DiffIDs = append(ociConfig.RootFS.DiffIDs, sigDesc.Digest)
```
vrothberg, line 691: Are the diff IDs needed?  I couldn't find something in the [cosign spec](https://github.com/sigstore/cosign/blob/main/specs/SIGNATURE_SPEC.md) but recall that cosign just ships an empty config.

docker/docker_image_dest.go
mtrmac: Nevertheless, doing more explicit copies when creating/consuming signatures would be worth the peace of mind.

docker/docker_image_dest.go
mtrmac: I don’t know.

It’s _quite_ possible they aren’t needed. I also thought cosign ships a completely empty config, but when looking what to write into that config, I found DiffID entries in a Cosign-generated file. That might be a `go-containerregistry` artifact, the API has concepts like “add a layer” which adds it both to the manifest and to DiffIDs.

At least distribution/distribution seems not to care, _except if the client is schema1-only_.

Historically Quay has been the registry with the strictest validation we’ve been hearing about, I have no idea what they do.

vrothberg: Thanks for checking!

I also suspect `go-containerregistry` to be the source.  I am totally fine to add the DiffIDs.

mtrmac: It’s a good thing to point out and to track — if we don’t need to add DiffIDs, we would eliminate a blob write for the config on every copy that adds signatures to an existing image with signatures.

docker/docker_image_dest.go
mtrmac: Replaced the FIXMEs with comments in #1595 .

copy/copy.go
mtrmac: Done in #1597 .

mtrmac:
There is now just one outstanding PR - #1598 , so closing this.

Outstanding unfinished tasks and the like are tracked in #1601.

PR #1600, opened 2022-07-07
Use "io.ReadAll" instead of "os.ReadAll"

rockdrilla:
Fixes FTBFS with go 1.18.3

Fixes: "Update users of deprecated io/ioutil" 7152f888
Signed-off-by: Konstantin Demin <rockdrilla@gmail.com>

### DIFF
--- a/ostree/ostree_src.go
+++ b/ostree/ostree_src.go
@@ -378,7 +378,7 @@ func (s *ostreeImageSource) GetSignatures(ctx context.Context, instanceDigest *d
 		}
 		defer sigReader.Close()
 
-		sig, err := os.ReadAll(sigReader)
+		sig, err := io.ReadAll(sigReader)
 		if err != nil {
 			return nil, err
 		}


rhatdan:
LGTM
@vrothberg @mtrmac PTAL

mtrmac:
Oops. So `ostree` didn’t compile for almost three months. Thanks!

@rockdrilla Are you actually using the `ostree:` transport?

---

(Right now, we would probably only notice on end-user releases for RPM builds.

Skopeo doesn’t include `ostree` since https://github.com/containers/skopeo/pull/751 , Podman since https://github.com/containers/podman/pull/4376 , Buildah since https://github.com/containers/buildah/pull/1954 .

c/image #736 is tracked to happen on a future major version bump of c/image… which we are trying very hard to avoid.

OTOH one thing we probably should do is to at least compile-test the `ostree:` transport in regular CI.)

mtrmac (approved):
LGTM.

PR #1604, opened 2022-07-08
fix skopeo copy can't  decrypt to docker-daemon image

ningmingxiao:

### DIFF
--- a/internal/image/docker_schema2.go
+++ b/internal/image/docker_schema2.go
@@ -161,11 +161,25 @@ func (m *manifestSchema2) UpdatedImageNeedsLayerDiffIDs(options types.ManifestUp
 // The returned error will be a manifest.ManifestLayerCompressionIncompatibilityError
 // if the CompressionOperation and CompressionAlgorithm specified in one or more
 // options.LayerInfos items is anything other than gzip.
-func (m *manifestSchema2) UpdatedImage(ctx context.Context, options types.ManifestUpdateOptions) (types.Image, error) {
+func (m *manifestSchema2) UpdatedImage(ctx context.Context, options types.ManifestUpdateOptions) (image types.Image, retErr error) {
 	copy := manifestSchema2{ // NOTE: This is not a deep copy, it still shares slices etc.
 		src:        m.src,
 		configBlob: m.configBlob,
 		m:          manifest.Schema2Clone(m.m),
+	}
+
+	manifestTmp := m
+	defer func() {
+		if retErr != nil {
+			m = manifestTmp
+		}
+	}()
+
+	// No conversion required, update manifest
+	if options.LayerInfos != nil {
+		if err := copy.m.UpdateLayerInfos(options.LayerInfos); err != nil {
+			return nil, err
+		}
 	}
 
 	converted, err := convertManifestIfRequiredWithUpdate(ctx, options, map[string]manifestConvertFn{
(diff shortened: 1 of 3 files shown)


ningmingxiao:
fix https://github.com/containers/skopeo/issues/1703

mtrmac (changes requested):
Thanks for the PR.

This is not accurate: `manifest.DockerV2Schema2ForeignLayerMediaTypeGzip` just isn’t an encrypted MIME type, and this conversion logic is correct to reject encrypted inputs.

Besides the conceptual cleanliness point, this would also drop the encrypted MIME type even if not decrypting at all! That case must continue to fail.

I suppose the change from encrypted to non-encrypted MIME types should happen before even considering a format conversion.

ningmingxiao:
will this new commit be ok?  @mtrmac

mtrmac:
I’m sorry about the previous very brief comment.

Just moving `UpdateLayerInfos` would, I think, work here, but the flip side is that the order of `convertManifestIfRequiredWithUpdate` vs. `UpdateLayerInfos` would be different in OCI vs. schema2, and we couldn’t make it consistent, otherwise a single step “convert from schema2 to OCI, and encrypt” would break. The code is messy and rife with with similar corner cases, so consistency of the implementation is valuable for maintenance.

---

Also, I suspect (but didn’t verify) that this implementation modifies the original image, which must not happen. That’s best ensured with a unit test within the existing test function — that would also help with making sure the use case in question doesn’t break.

---

I think we have to bite the conceptual bullet, accept, and express more the complexity … in this case I think `manifestOCI1.UpdatedImage` should do a separate check for `info.CryptoOperation == types.Decrypt` (and maybe, _eventually_, not in this PR, some other edits) and make an update of that kind first, so that it moves the image towards a “more interoperable” variant, then call `convertManifestIfRequiredWithUpdate`, and finally the original `UpdateLayerInfos`.

mtrmac:
> Also, I suspect (but didn’t verify) that this implementation modifies the original image

oh, maybe not, `UpdateLayerInfos` allocates a new slice. Either way, a test would be the best way to make sure.

internal/image/oci.go
```diff
@@ -143,13 +143,26 @@ func (m *manifestOCI1) UpdatedImageNeedsLayerDiffIDs(options types.ManifestUpdat
...
 		configBlob: m.configBlob,
 		m:          manifest.OCI1Clone(m.m),
 	}
 
+	manifestTmp := m
+	defer func() {
+		if retErr != nil {
+			m = manifestTmp
+		}
+	}()
```
mtrmac, line 158: I don’t understand at all what this is intended to do. `m` is local to this function, and just a pointer. Assigning to it, AFAICS, does nothing visible to the caller.

internal/image/oci.go
```diff
@@ -143,13 +143,26 @@ func (m *manifestOCI1) UpdatedImageNeedsLayerDiffIDs(options types.ManifestUpdat
...
+
+	// No conversion required, update manifest
```
mtrmac, line 160: This comment is not correct an this place.

internal/image/docker_schema2.go
```diff
@@ -161,13 +161,27 @@ func (m *manifestSchema2) UpdatedImageNeedsLayerDiffIDs(options types.ManifestUp
...
+	manifestTmp := m
+	defer func() {
+		if retErr != nil {
+			m = manifestTmp
+		}
+	}()
+
+	// No conversion required, update manifest
+	if options.LayerInfos != nil {
+		if err := copy.m.UpdateLayerInfos(options.LayerInfos); err != nil {
```
mtrmac, line 180: As mentioned previously, this would break one-step ”convert to OCI and encrypt”.

We need to split that, and do the decryption MIME type changes before format conversions, and encryption MIME type changes after format conversions.

mtrmac (commented):
Also, without unit tests we can’t be confident this will keep working.

mtrmac:
#1932 should make all of this work. Thanks for your work in this area!

PR #1606, opened 2022-07-11
s/sigstore/lookaside/

mtrmac:
… per recent discussions.

<s>Only tested by that one updated unit test so far.</s>

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -108,7 +108,7 @@ type dockerClient struct {
 	// The following members are not set by newDockerClient and must be set by callers if needed.
 	auth                 types.DockerAuthConfig
 	registryToken        string
-	signatureBase        signatureStorageBase
+	signatureBase        lookasideStorageBase
 	useCosignAttachments bool
 	scope                authScope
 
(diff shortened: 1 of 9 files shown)


mtrmac:
Somewhat tested by https://github.com/containers/skopeo/pull/1701 , with integration tests using `lookaside` and system tests `sigstore`.

vrothberg (approved):
LGTM

ChristianCiach:
Sorry, but what is "lookaside"? You seem to use it interchangeably with "sigstore". Is "sigstore" a bad word around here?

 > … per recent discussions.
 
 Which discussions?

ktdreyer:
It's worth putting this important historical information into https://github.com/containers/image/blob/main/docs/signature-protocols.md

The deprecated `sigstore` setting is still widely used, eg. https://src.fedoraproject.org/rpms/containers-common/blob/rawhide/f/registry.redhat.io.yaml

mtrmac:
`signature-protocols.md` primarily documents the inter-computer APIs and protocols, and _a bit_ the recommended end-user workflows. If the old fields should be documented, it would be in the reference documentation for the file, `containers-registries.d(5)`.

In principle, yes, documenting those fields is valuable; doing that without confusing new GitHub.com/sigstore users seems a bit hard and it would probably require a wider rewrite of that page.

ktdreyer:
I took a look at rewriting the man page and gave up. Here's the change to edit Fedora's config file instead: https://src.fedoraproject.org/rpms/containers-common/pull-request/6

PR #1608, opened 2022-07-12
enhance skopeo  inspect

ningmingxiao:

### DIFF
--- a/internal/image/docker_schema1_test.go
+++ b/internal/image/docker_schema1_test.go
@@ -273,56 +273,71 @@ func TestManifestSchema1Inspect(t *testing.T) {
 		manifestSchema1FromFixture(t, "schema1.json"),
 		manifestSchema1FromComponentsLikeFixture(t),
 	} {
+
+		labels := map[string]string{
+			"Kolla-SHA":                "5.0.0-39-g6f1b947b",
+			"architecture":             "x86_64",
+			"authoritative-source-url": "registry.access.redhat.com",
+			"build-date":               "2018-01-25T00:32:27.807261",
+			"com.redhat.build-host":    "ip-10-29-120-186.ec2.internal",
+			"com.redhat.component":     "openstack-nova-api-docker",
+			"description":              "Red Hat OpenStack Platform 12.0 nova-api",
+			"distribution-scope":       "public",
+			"io.k8s.description":       "Red Hat OpenStack Platform 12.0 nova-api",
+			"io.k8s.display-name":      "Red Hat OpenStack Platform 12.0 nova-api",
+			"io.openshift.tags":        "rhosp osp openstack osp-12.0",
+			"kolla_version":            "stable/pike",
+			"name":                     "rhosp12/openstack-nova-api",
+			"release":                  "20180124.1",
+			"summary":                  "Red Hat OpenStack Platform 12.0 nova-api",
+			"tripleo-common_version":   "7.6.3-23-g4891cfe",
+			"url":                      "https://access.redhat.com/containers/#/registry.access.redhat.com/rhosp12/openstack-nova-api/images/12.0-20180124.1",
+			"vcs-ref":                  "9b31243b7b448eb2fc3b6e2c96935b948f806e98",
+			"vcs-type":                 "git",
+			"vendor":                   "Red Hat, Inc.",
+			"version":                  "12.0",
+			"version-release":          "12.0-20180124.1",
+		}
+
+		Layers := []string{
+			"sha256:9cadd93b16ff2a0c51ac967ea2abfadfac50cfa3af8b5bf983d89b8f8647f3e4",
+			"sha256:4aa565ad8b7a87248163ce7dba1dd3894821aac97e846b932ff6b8ef9a8a508a",
+			"sha256:f576d102e09b9eef0e305aaef705d2d43a11bebc3fd5810a761624bd5e11997e",
+			"sha256:9e92df2aea7dc0baf5f1f8d509678d6a6306de27ad06513f8e218371938c07a6",
+			"sha256:62e48e39dc5b30b75a97f05bccc66efbae6058b860ee20a5c9a184b9d5e25788",
+			"sha256:e623934bca8d1a74f51014256445937714481e49343a31bda2bc5f534748184d",
+		}
+
+		Env := []string{
+			"PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
+			"container=oci",
+			"KOLLA_BASE_DISTRO=rhel",
+			"KOLLA_INSTALL_TYPE=binary",
+			"KOLLA_INSTALL_METATYPE=rhos",
+			"PS1=$(tput bold)($(printenv KOLLA_SERVICE_NAME))$(tput sgr0)[$(id -un)@$(hostname -s) $(pwd)]$ ",
+		}
+
 		ii, err := m.Inspect(context.Background())
 		require.NoError(t, err)
 		created := time.Date(2018, 1, 25, 0, 37, 48, 268558000, time.UTC)
-		assert.Equal(t, types.ImageInspectInfo{
-			Tag:           "latest",
-			Created:       &created,
-			DockerVersion: "1.12.6",
-			Labels: map[string]string{
-				"Kolla-SHA":                "5.0.0-39-g6f1b947b",
-				"architecture":             "x86_64",
-				"authoritative-source-url": "registry.access.redhat.com",
-				"build-date":               "2018-01-25T00:32:27.807261",
-				"com.redhat.build-host":    "ip-10-29-120-186.ec2.internal",
-				"com.redhat.component":     "openstack-nova-api-docker",
-				"description":              "Red Hat OpenStack Platform 12.0 nova-api",
-				"distribution-scope":       "public",
-				"io.k8s.description":       "Red Hat OpenStack Platform 12.0 nova-api",
-				"io.k8s.display-name":      "Red Hat OpenStack Platform 12.0 nova-api",
-				"io.openshift.tags":        "rhosp osp openstack osp-12.0",
-				"kolla_version":            "stable/pike",
-				"name":                     "rhosp12/openstack-nova-api",
-				"release":                  "20180124.1",
-				"summary":                  "Red Hat OpenStack Platform 12.0 nova-api",
-				"tripleo-common_version":   "7.6.3-23-g4891cfe",
-				"url":                      "https://access.redhat.com/containers/#/registry.access.redhat.com/rhosp12/openstack-nova-api/images/12.0-20180124.1",
-				"vcs-ref":                  "9b31243b7b448eb2fc3b6e2c96935b948f806e98",
-				"vcs-type":                 "git",
-				"vendor":                   "Red Hat, Inc.",
-				"version":                  "12.0",
-				"version-release":          "12.0-20180124.1",
-			},
-			Architecture: "amd64",
-			Os:           "linux",
-			Layers: []string{
-				"sha256:9cadd93b16ff2a0c51ac967ea2abfadfac50cfa3af8b5bf983d89b8f8647f3e4",
-				"sha256:4aa565ad8b7a87248163ce7dba1dd3894821aac97e846b932ff6b8ef9a8a508a",
-				"sha256:f576d102e09b9eef0e305aaef705d2d43a11bebc3fd5810a761624bd5e11997e",
-				"sha256:9e92df2aea7dc0baf5f1f8d509678d6a6306de27ad06513f8e218371938c07a6",
-				"sha256:62e48e39dc5b30b75a97f05bccc66efbae6058b860ee20a5c9a184b9d5e25788",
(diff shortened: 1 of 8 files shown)


ningmingxiao:
skopeo inspect can't see whether an image's MediaType,I want to use MediaType to check whether it is enctryed.

types/types.go
```diff
@@ -464,8 +464,15 @@ type ImageInspectInfo struct {
...
 	Variant       string
 	Os            string
-	Layers        []string
+	Layers        []Layer
 	Env           []string
+	Cmd           []string
+}
+
+type Layer struct {
+	MediaType string
```
mtrmac, line 473: Despite old precedents, we generally prefer naming this a `MIMEType` nowadays.

mtrmac (changes requested):
Thanks for the PR.

c/image is following the Go semantic versioning rules, so we must not incompatibly change the externally-visible API like this (and we have no appetite for increasing the major version number either).

This needs to be done by adding new fields, without modifying any of the existing ones.

manifest/oci.go
```diff
@@ -225,6 +225,44 @@ func (m *OCI1) Inspect(configGetter func(types.BlobInfo) ([]byte, error)) (*type
 	return i, nil
 }
 
+// Inspect returns various information for (skopeo inspect) parsed from the manifest and configuration.
+func (m *OCI1) InspectV2(configGetter func(types.BlobInfo) ([]byte, error)) (*types.ImageInspectInfoV2, error) {
```
mtrmac, line 229: I’m sorry about the lack of detail - it _is_ API-compatible to add new fields to the returned struct; we just can’t modify any existing ones.

So an entire new `InspectV2` isn’t necessary — and in fact the copy&pasted duplication would be harder to maintain.

types/types.go
```diff
@@ -466,6 +466,18 @@ type ImageInspectInfo struct {
 	Os            string
 	Layers        []string
 	Env           []string
+	LayersDetail  []LayerDetail
+	History       []v1.History
+	Author        string
+	Size          int64
+	Config        v1.ImageConfig
```
mtrmac, line 473: Adding a full config here is redundant with `Image.OCIConfig`, and probably unwanted for the `Inspect` caller, who is calling `Inspect` exactly not to deal with the details of the config format.

types/types.go
```diff
@@ -466,6 +466,18 @@ type ImageInspectInfo struct {
...
+}
+
+type LayerDetail struct {
```
mtrmac, line 476: `ImageInspectLayer`?

types/types.go
```diff
@@ -466,6 +466,18 @@ type ImageInspectInfo struct {
 	Os            string
 	Layers        []string
 	Env           []string
+	LayersDetail  []LayerDetail
+	History       []v1.History
```
mtrmac, line 470: As this very PR shows, it’s valuable to be able to add extra fields — so a separate `ImageInspectHistory` where we can add fields to would be better than forever tying this to OCIv1, and potentially having to add a `HistoryDetail`.

manifest/oci.go
```diff
@@ -220,7 +220,12 @@ func (m *OCI1) Inspect(configGetter func(types.BlobInfo) ([]byte, error)) (*type
 		Architecture:  v1.Architecture,
 		Os:            v1.OS,
 		Layers:        layerInfosToStrings(m.LayerInfos()),
+		LayersDetail:  imgInspectLayersFromLayerInfos(m.LayerInfos()),
 		Env:           v1.Config.Env,
+		History:       v1.History,
+		Author:        v1.Author,
+		Size:          d1.Size,
```
mtrmac, line 227: Size is not defined in OCI — so the `ImageInspectInfo` documentation for the field must document how “unknown” is represented, and this code needs to set that value.

(`0` is not a good “unknown” value for a size field, in principle… We typically use -1.)

manifest/docker_schema2.go
```diff
@@ -151,11 +152,11 @@ type Schema2History struct {
...
 type Schema2Image struct {
 	Schema2V1Image
-	Parent     digest.Digest    `json:"parent,omitempty"`
-	RootFS     *Schema2RootFS   `json:"rootfs,omitempty"`
-	History    []Schema2History `json:"history,omitempty"`
-	OSVersion  string           `json:"os.version,omitempty"`
-	OSFeatures []string         `json:"os.features,omitempty"`
+	Parent     digest.Digest       `json:"parent,omitempty"`
+	RootFS     *Schema2RootFS      `json:"rootfs,omitempty"`
+	History    []imgspecv1.History `json:"history,omitempty"`
```
mtrmac, line 157: This is an externally-visible breaking change to the API, so not something we can do. (Also, inaccurate, if either of the formats added more fields.)

manifest/docker_schema2.go
```diff
@@ -215,6 +216,19 @@ func (m *Schema2) LayerInfos() []LayerInfo {
 	return blobs
 }
 
+// imgInspectLayersFromLayerInfos converts a list of layer infos, presumably obtained from a Manifest.LayerInfos()
+// method call, into a format suitable for inclusion in a types.ImageInspectInfo structure.
+func imgInspectLayersFromLayerInfos(infos []LayerInfo) []types.LayerDetail {
```
mtrmac, line 221: When this function is used in all the format handlers, `manifest/common.go` would be a better location.

manifest/docker_schema2.go
```diff
@@ -279,10 +293,26 @@ func (m *Schema2) Inspect(configGetter func(types.BlobInfo) ([]byte, error)) (*t
...
+		i.Config.User = s2.Config.User
+		i.Config.Volumes = s2.Config.Volumes
+		i.Config.Entrypoint = s2.Config.Entrypoint
+		for key, value := range s2.Config.ExposedPorts {
+			exposedPorts := make(map[string]struct{})
+			exposedPorts[string(key)] = value
+			i.Config.ExposedPorts = exposedPorts
+		}
+		i.Config.StopSignal = s2.Config.StopSignal
+		i.Config.WorkingDir = s2.Config.WorkingDir
```
mtrmac, line 315: See the `OCIConfig` comment… this feels like a code that shouldn’t need to exist.

internal/image/docker_schema2.go
```diff
@@ -326,7 +326,7 @@ func (m *manifestSchema2) convertToManifestSchema1(ctx context.Context, options
 			ID:        v1ID,
 			Parent:    parentV1ID,
 			Comment:   historyEntry.Comment,
-			Created:   historyEntry.Created,
+			Created:   *historyEntry.Created,
```
mtrmac, line 329: (This will probably go away because we just can’t change the type of `historyEntry` to the OCI struct.)

Blindly dereferencing like this can crash. Sure, the field is not optional in the format definition, but invalid/malicious configs still shouldn’t cause the code to crash.

manifest/docker_schema1.go
```diff
@@ -228,10 +228,26 @@ func (m *Schema1) Inspect(_ func(types.BlobInfo) ([]byte, error)) (*types.ImageI
 		Architecture:  s1.Architecture,
 		Os:            s1.OS,
 		Layers:        layerInfosToStrings(m.LayerInfos()),
+		LayersDetail:  imgInspectLayersFromLayerInfos(m.LayerInfos()),
+		History:       nil,
```
mtrmac, line 232: If `nil` is means “history is unavailable” and not “a zero-length history”, that needs to be documented in `types.go`.

Actually history exists in the image, compare `Schema1.ToSchema2Config`.

manifest/oci.go
```diff
@@ -220,7 +220,12 @@ func (m *OCI1) Inspect(configGetter func(types.BlobInfo) ([]byte, error)) (*type
 		Architecture:  v1.Architecture,
 		Os:            v1.OS,
 		Layers:        layerInfosToStrings(m.LayerInfos()),
+		LayersDetail:  imgInspectLayersFromLayerInfos(m.LayerInfos()),
```
mtrmac, line 223: Non-blocking: It would be nice to call `LayerInfos` only once.

(In all three format handlers.)

types/types.go
```diff
@@ -466,6 +466,18 @@ type ImageInspectInfo struct {
...
+	Author        string
+	Size          int64
```
mtrmac, line 472: Size of what? The compressed representation? Uncompressed representation? Either, depending on circumstances?

In OCI this can apparently be unknown.

mtrmac (commented):
This PR keeps growing…

While the additions are certainly valuable, perhaps it might be better to get one feature _done_, and then add others similarly, in one or more smaller PRs.

mtrmac:
Unit test coverage, to make sure this does not accidentally break, would also be nice. (And, looking at how the existing tests fail, it’s in a sense unavoidable at least to that little extent.)

types/types.go
ningmingxiao: I think ImageInspectLayer  and  Layers are same meaning, at .ImageInspectInfo struct

types/types.go
ningmingxiao: you means I should use another func  ImageInspectHistory to show history?  When I use podman inspect an image, I can see history

types/types.go
ningmingxiao: I use podman  or docker inspect an image, they will show config, so I think we should stay the same.

internal/image/docker_schema1_test.go
```diff
@@ -273,56 +273,71 @@ func TestManifestSchema1Inspect(t *testing.T) {
...
+		assert.Equal(t, &created, ii.Created)
+		assert.Equal(t, "1.12.6", ii.DockerVersion)
+		assert.Equal(t, labels, ii.Labels)
+		assert.Equal(t, "amd64", ii.Architecture)
+		assert.Equal(t, "linux", ii.Os)
+		assert.Equal(t, Layers, ii.Layers)
+		assert.Equal(t, Env, ii.Env)
+		assert.NotEqual(t, nil, ii.Config)
+		assert.Equal(t, int64(0), ii.Size)
+		assert.Equal(t, "", ii.Author)
```
mtrmac, line 333: Please maintain the original structure.

A single `Equal` over the whole structure is more useful because it ensures that nothing was left out in the tests. Doing this item by item requires much more manual inspection and maintenance.

internal/image/docker_schema1_test.go
```diff
@@ -273,56 +273,71 @@ func TestManifestSchema1Inspect(t *testing.T) {
...
+		for _, layer := range ii.LayersDetail {
+			assert.NotEqual(t, nil, layer)
```
mtrmac, line 335: AFAICS that’s always true (members of the slice are not pointers at all), and not a useful test. Same for the other loop.

types/types.go
mtrmac: But then are other callers like `skopeo inspect`, or automation that just wants a single element (like looking up a single label) that don’t need both at the same time. Paying the cost of the full conversion and JSON formatting is not all that great for them.

As a maintainer, I’d also _much_ prefer not to have to maintain logic to convert to OCI format in two separate implementations for each format.

types/types.go
mtrmac: The `types` subpackage contains quite a few types … and it’s useful to make it clear that the purpose of this struct is _only_ as an inspect output.

We already have `types.BlobInfo` (and `manifest.LayerInfo`); being clear about the purposes of each (especially when there is no written documentation) is helpful to users of the API.

(Especially when may users are just going to use `Inspect().LayersDetail…` without having to name the type, we need the type name more to be unambiguous than we need it to be short.)

types/types.go
mtrmac: I mean that the _type_ of the field should be `History []ImageInpectHistory`.

manifest/oci.go
mtrmac: Still outstanding: The -1 value needs documenting.

types/types.go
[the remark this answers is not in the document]
mtrmac: Outstanding.

types/types.go
mtrmac: Still outstanding; the added comment doesn’t answer this question.

manifest/docker_schema1.go
```diff
@@ -215,23 +215,59 @@ func validateV1ID(id string) error {
...
+		Author:        s1.Author,
+		Size:          s1.Size,
 	}
+
+	diffIDs := []digest.Digest{digest.FromString(s1.ID)}
+	schema2Config, err := m.ToSchema2Config(diffIDs)
+	if err != nil {
+		return nil, err
+	}
+	imageConfig := &schema2Image{}
```
mtrmac, line 251: (Why is this using a custom type and not the existing schema2 `manifest.Schema2Image`? But see elsewhere, we shouldn’t need to involve JSON for this.)

types/types.go
```diff
@@ -466,6 +466,19 @@ type ImageInspectInfo struct {
 	Os            string
 	Layers        []string
 	Env           []string
+	LayersDetail  []LayerDetail
```
mtrmac, line 469: I think we can keep `LayersDetail` right next to `Layers`.

TomSweeneyRedHat:
@ningmingxiao needs a rebase

mtrmac:
#1626 follows up on this PR (with a subset of the functionality).

jankaluza:

PR #1615, opened 2022-07-20
copy: print copy info once when writer==io.Discard

vrothberg:
When the writer is set to `io.Discard` make sure to print the
`Copying ...` output once.  The partial pull attempt caused
the line to be printed twice rendering the output of
`podman-remote pull ...` a bit too verbose.


copy/copy.go
```diff
@@ -1253,6 +1253,7 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, to
 			logrus.Debugf("Failed to retrieve partial blob: %v", err)
 			return false, types.BlobInfo{}
 		}(); reused {
+			ic.c.printCopyInfo("blob", srcInfo)
```
mtrmac, line 1256: Isn’t this too late? I understand the point of of “copying…” to be to show what the code _will_ be doing afterwards = which part is taking too long. Printing this only after a successful copy is not what that means.

mtrmac: - A very minimal local improvement would be to change the message to be “Trying to use a partial pull…” just so that the two aren’t duplicated. OTOH that’s clutter that is, most of the time, not useful.
- Really, after `TryReusingBlob…` fails, we know we are going to copy a layer, so we could `Printf` at that point, instead of tying that to progress bars. (A notable downside is that this would make the two output kinds less connected).
- Ideally, I think `private.ImageDestination` should have an `LayerAllowsPutBlobPartial` check (which does the annotation comparisons from `c/storage/pkg/chunked.GetDiffer`); `copyLayer` would then call that first, and not bother with the progress bars, or any user-visible activity, at all, for blobs where this isn’t an option. (Combined with the message change to “Trying to use a partial pull”.)

vrothberg: I do not know another way.  As far as I understood the code, the bar would just get removed if it fails but we cannot revert the "copying ..." line once it's out.

I figured it's less ugly to have a line later than twice.

vrothberg: The updated version should address the problem.  In the `writer == io.Discard` case we never really cared whether we skipped the copy or not, so I moved it out of the progress bar creation and made it explicit for copying a config and a layer.

mtrmac (commented):
Nice, this works.

Please also update the `createProgressBar` comment in `copy.Image` to point at the new function.

vrothberg:
Done :heavy_check_mark: Thanks, @mtrmac !

mtrmac:
Please update this comment as well: https://github.com/containers/image/blob/55ea76c7db702ed1af60924a0b57c8da533d9e5a/copy/copy.go#L238-L244

mtrmac:
Feel free to merge after tests pass.

vrothberg:
@mtrmac any objections against cutting a patch release tomorrow?

mtrmac:
> @mtrmac any objections against cutting a patch release tomorrow?

On `main` it wouldn’t be a small patch release — it would be a 5.22.0 with all the sigstore work and other activity since start of May (+10514-5646 lines).

If you are fine with that, sure, the repo should always be in a usable state and a release should be an option.

If we need a small patch release, this PR needs to be backported to some branch.

PR #1617, opened 2022-07-20
Silence a "potentially unused parameter" warning

mtrmac:
`configBlobImageSource.f` is only called if the digest is exactly one value, so the function doesn't need to use it (and none of them do).  So, remove it; this also drops a recently-introduced warning in VS Code.

Should not change (test) behavior.


rhatdan:
LGTM
@vrothberg PTAL

vrothberg (approved):
LGTM

PR #1620, opened 2022-07-22
Manually update various old dependencies

mtrmac:
… both direct and indirect.

This is a random pass through primarily because GitHub’s Dependabot doesn’t update from commit references to later releases, and also to possibly compensate for direct dependencies that depend on other packages but don’t check for updates.

A few commit dependencies that are already just a month or two old were left alone; those much older than that were updated.

(I don’t think it makes sense to commit to keeping all commit references updated — some projects apparently do do that using (some version of?) Dependabot, but that’s a _lot_ of noise for repos like `golang.org/x/…` that never make releases. A manual pass from time to time seems a reasonable compromise to me.)


rhatdan:
LGTM

PR #1621, opened 2022-07-25
Cirrus: Update CI VM images

cevich:
**Depends on**: https://github.com/containers/skopeo/pull/1711

Note: Removed disused `PRIOR_FEDORA*` and `UBUNTU_*` references since
they're not actually used in this CI.  Further, F35 VM images were not
built as part of `c6013173500215296` due to a missing golang 1.18
requirement for podman.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -33,18 +33,13 @@ env:
     #### Cache-image names to test with (double-quotes around names are critical)
     ####
     FEDORA_NAME: "fedora-36"
-    PRIOR_FEDORA_NAME: "fedora-35"
-    UBUNTU_NAME: "ubuntu-2204"
 
     # Google-cloud VM Images
-    IMAGE_SUFFIX: "c6340043416535040"
+    IMAGE_SUFFIX: "c5495735033528320"
     FEDORA_CACHE_IMAGE_NAME: "fedora-${IMAGE_SUFFIX}"
-    PRIOR_FEDORA_CACHE_IMAGE_NAME: "prior-fedora-${IMAGE_SUFFIX}"
-    UBUNTU_CACHE_IMAGE_NAME: "ubuntu-${IMAGE_SUFFIX}"
 
     # Container FQIN's (include bleeding-edge development-level container deps.)
     FEDORA_CONTAINER_FQIN: "quay.io/libpod/fedora_podman:${IMAGE_SUFFIX}"
-    PRIOR_FEDORA_CONTAINER_FQIN: "quay.io/libpod/prior-fedora_podman:${IMAGE_SUFFIX}"
     UBUNTU_CONTAINER_FQIN: "quay.io/libpod/ubuntu_podman:${IMAGE_SUFFIX}"
     # Built along with the standard PR-based workflow in c/automation_images
     SKOPEO_CIDEV_CONTAINER_FQIN: "quay.io/libpod/skopeo_cidev:${IMAGE_SUFFIX}"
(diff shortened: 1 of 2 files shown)


mtrmac (commented):
Purely locally, the `.cirrus.yml` changes LGTM.

(Marking the PR as draft in GitHub, we don’t have a bot for `/hold` currently.)

cevich:
This should be ready to go now, after https://github.com/containers/skopeo/pull/1711

mtrmac:
https://github.com/containers/skopeo/pull/1711 was merged, but `validate` is still unhappy in https://cirrus-ci.com/task/6214691553804288?logs=main#L184 . Is that specific to c/image?

cevich:
> is still unhappy

I'm assuming by "still" you tried re-running it?  Then, no, it's probably not caused by the skopeo update :disappointed: 

```
make validate
/var/tmp/go/bin/golangci-lint run --build-tags "btrfs_noversion libdm_no_deferred_remove"
panic: load embedded ruleguard rules: rules/rules.go:13: can't load fmt
```

Hrmmm, so the `Makefile` for the above is quite simple, literally calling that `golangci-lint` command directly.  It's configuration is also quite sparse:

```yaml
---
run:
  concurrency: 6
  deadline: 5m
```

Curiously, I see skopeo doesn't use `golangci-lint`, it runs something called `golint`.  So I'd ascribe the errors here to new versions of either golang or golang-ci-lint in the environment...ahhh!  The `validate` task uses an Ubuntu container image: `quay.io/libpod/ubuntu_podman:c6013173500215296` in this case.  ***THAT*** was definitely updated in this environment from golang `1.17` (IIRC) -> `1.18.4`.  As for `golang-ci-lint`, that would have come from podman's Makefile, looking...it says `VERSION=1.46.2`.

@mtrmac any of this spark any ideas?

mtrmac:
> I'm assuming by "still" you tried re-running it?

I’m sorry, I should have said this: Yes, I triggered a re-run after merging the Skopeo PR.

---

> So I'd ascribe the errors here to new versions of either golang or golang-ci-lint in the environment.

Looking for that text finds an upstream report that does suggest that this is some kind of a conflict between the (upstream-compiled) `golangci-lint` binary and the system’s Golang binaries.

The release notes of `golangci-lint` document that Go 1.18 is supported starting in 1.45. So, let me try upgrading… #1624 .

---

(As for using Ubuntu, I think this is a decision that seems to have been inherited all the way back from 2016 or so. I’m not aware of any specific reason why it _has_ to be Ubuntu, if migrating would make things easier. But I guess that should happen in some other PR, to minimize risk.)

mtrmac:
@cevich So #1624 fixes the `validate` step, but system tests now fail with
```
  `arch=$(podman info --format '{{.Host.Arch}}')' failed with status 125
# time="2022-07-29T11:39:26-05:00" level=warning msg="Storage configuration is unset - using hardcoded default graph root \"/var/lib/containers/storage\""
# time="2022-07-29T11:39:26-05:00" level=warning msg="Storage configuration is unset - using hardcoded default graph root \"/var/lib/containers/storage\""
# Error: no storage runroot or graphroot specified: missing necessary StoreOptions
```
. Skopeo used to fail the same way in https://github.com/containers/skopeo/pull/1711 , is there some other known problem that was fixed for Skopeo and needs fixing in c/image? Or is it a new flake that just happened to pass in Skopeo? (It doesn’t quite look like a flake to me, I’d guess a missing `storage.conf` in the `skopeo_cidev` container.)

This PR uses c6013173500215296 ; the Skopeo one started with that value, but was updated to c5495735033528320 later.

cevich:
> that would have come from podman's Makefile, looking...it says VERSION=1.46.2.

Great, glad you found something.

> but was updated to c5495735033528320 later.

Oh woops, did I not update it here?  My bad...

cevich:
> Or is it a new flake that just happened to pass in Skopeo?

Note: I'm not sure about that.  But the old image had some known broken dependencies (runc IIRC).  And in any case, the image ID has to match what's in skopeo CI.  Let's see what happens...

mtrmac:
Tests succeeded now. Just to be a bit more certain it wasn’t a flake, I have hit the “Re-run all checks” button, and I think we can merge if that passes.

cevich:
Great!

mtrmac:
(“Re-run all checks” seems to have done nothing, but re-running the two Skopeo test directly in Cirrus did schedule a re-run.)

cevich:
Oh ya, the big re-run doesn't work if everything passed.  I forgot about that.

mtrmac:
Looks good, merging. Thanks!

PR #1626, opened 2022-08-01
enhance inspect

ningmingxiao:

### DIFF
--- a/internal/image/docker_schema1_test.go
+++ b/internal/image/docker_schema1_test.go
@@ -276,6 +276,7 @@ func TestManifestSchema1Inspect(t *testing.T) {
 		ii, err := m.Inspect(context.Background())
 		require.NoError(t, err)
 		created := time.Date(2018, 1, 25, 0, 37, 48, 268558000, time.UTC)
+		var emptyAnnotations map[string]string
 		assert.Equal(t, types.ImageInspectInfo{
 			Tag:           "latest",
 			Created:       &created,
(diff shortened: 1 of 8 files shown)


rhatdan:
Thanks @ningmingxiao 
LGTM
@mtrmac @vrothberg PTAL

types/types.go
```diff
@@ -466,6 +466,15 @@ type ImageInspectInfo struct {
 	Os            string
 	Layers        []string
 	Env           []string
+	LayersInspect  []ImageInspectLayer
+	Author        string
+}
+
+type ImageInspectLayer struct {
```
vrothberg, line 473: Please add a top-level comment to this new struct.

manifest/docker_schema2.go
```diff
@@ -278,7 +279,9 @@ func (m *Schema2) Inspect(configGetter func(types.BlobInfo) ([]byte, error)) (*t
 		Architecture:  s2.Architecture,
 		Variant:       s2.Variant,
 		Os:            s2.OS,
-		Layers:        layerInfosToStrings(m.LayerInfos()),
+		Layers:        layerInfosToStrings(layerInfos),
+		LayersInspect:  imgInspectLayersFromLayerInfos(layerInfos),
```
vrothberg, line 283: `LayersData`?

vrothberg (commented):
Thanks!

Please also add a description to the commit message that explains what the commit is about.

internal/image/docker_schema1_test.go
```diff
@@ -269,13 +269,15 @@ func TestManifestSchema1EmbeddedDockerReferenceConflicts(t *testing.T) {
 }
 
 func TestManifestSchema1Inspect(t *testing.T) {
+
```
mtrmac, line 272: (Non-blocking: drop this change.)

internal/image/oci_test.go
```diff
@@ -275,7 +275,7 @@ func TestManifestOCI1EmbeddedDockerReferenceConflicts(t *testing.T) {
 func TestManifestOCI1Inspect(t *testing.T) {
 	configJSON, err := os.ReadFile("fixtures/oci1-config.json")
 	require.NoError(t, err)
-
+	var annotations map[string]string
```
mtrmac, line 278: It would be a bit easier to maintain if the three tests were updated consistently. Right now, each of the three adds the `annotations` field differently, making the reader wonder if there is any difference.

internal/image/docker_schema1_test.go
```diff
@@ -269,13 +269,15 @@ func TestManifestSchema1EmbeddedDockerReferenceConflicts(t *testing.T) {
...
 	for _, m := range []genericManifest{
 		manifestSchema1FromFixture(t, "schema1.json"),
 		manifestSchema1FromComponentsLikeFixture(t),
 	} {
 		ii, err := m.Inspect(context.Background())
 		require.NoError(t, err)
 		created := time.Date(2018, 1, 25, 0, 37, 48, 268558000, time.UTC)
+		var annotations map[string]string
```
mtrmac, line 280: Non-blocking:

If this is only used to avoid repetition in the constant, naming this `emptyAnnotations` would be a bit clearer. Personally I’d just use a literal `map[string]string{}` in data.

Either is fine.

---

… in all three formats.

types/types.go
```diff
@@ -466,6 +466,16 @@ type ImageInspectInfo struct {
 	Os            string
 	Layers        []string
 	Env           []string
+	LayersData    []ImageInspectLayer
```
mtrmac, line 469: Please move this one line up, to be close to `Layers` (and consistent with the `Inspect` implementations).

Or alternatively (less preferable?) update the `Inspect` implementations to match the order of fields here.

types/types.go
```diff
@@ -466,6 +466,16 @@ type ImageInspectInfo struct {
...
+	Author        string
+}
+
+//ImageInspectLayer is a set of metadata describing an image layers' detail
+type ImageInspectLayer struct {
+	MIMEType    string
```
mtrmac, line 475:
```suggestion
	MIMEType    string // "" if unknown.
```

types/types.go
```diff
@@ -466,6 +466,16 @@ type ImageInspectInfo struct {
...
+	Digest      digest.Digest
+	Size        int64
```
mtrmac, line 477:
```suggestion
	Size        int64 // -1 if unknown.
```

manifest/docker_schema2.go
```diff
@@ -267,6 +267,7 @@ func (m *Schema2) Inspect(configGetter func(types.BlobInfo) ([]byte, error)) (*t
 	if err != nil {
 		return nil, err
 	}
+	layerInfos := m.LayerInfos()
```
mtrmac, line 270: `LayerInfos` are computed after the `Unmarshal` in the other two implementations, this one differs for no obvious reason.

mtrmac (commented):
Thanks! Overall implementation LGTM.

The documentation of “unknown” values in `ImageInspectLayer` is blocking; the other stylistic/consistency nits are a weaker preference.

mtrmac (approved):
LGTM. Thanks again!

PR #1630, opened 2022-08-15
Use c/image's reference package

mtrmac:
… instead of the external dependency, which snuck in by IDE autocomplete.


rhatdan:
LGTM

PR #1631, opened 2022-08-18
Run gofmt

mtrmac:
The previous PR was tested with an earlier Golang version, so it was merged despite violating current formatting validation.

This should fix breakage when working on unrelated code.


vrothberg (approved):
LGTM

PR #1632, opened 2022-08-22
Use openpgp signature if cgo is disabled

paralin:
Fixes build when CGO_ENABLED=0 and no tags are set.

The openpgp implementation is used only if the containers_image_openpgp tag is set, otherwise the cgo-dependent libgpgme implementation is used.

This patch also uses openpgp if cgo is unavailable (the cgo tag is not set).

### DIFF
--- a/signature/mechanism_gpgme.go
+++ b/signature/mechanism_gpgme.go
@@ -1,5 +1,5 @@
-//go:build !containers_image_openpgp
-// +build !containers_image_openpgp
+//go:build !containers_image_openpgp && cgo
+// +build !containers_image_openpgp,cgo
 
 package signature
 
--- a/signature/mechanism_gpgme_test.go
+++ b/signature/mechanism_gpgme_test.go
@@ -1,5 +1,5 @@
-//go:build !containers_image_openpgp
-// +build !containers_image_openpgp
+//go:build !containers_image_openpgp && cgo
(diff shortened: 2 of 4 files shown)


mtrmac:
Thanks for the PR, but this is not desirable.

The in-process openpgp implementation is basically unmaintained, and we don’t support signing with it at all. So it needs to be explicitly opted into.

paralin:
I thought you might close this with that kind of message. But, the thing doesn't compile when cgo is disabled... How do you plan to address that?

mtrmac:
Existing callers that disable CGo add the `containers_image_openpgp` build tag, e.g. https://github.com/openshift/machine-config-operator/blob/b6cfd42372427eb5d7aee44ce30c81c74ae12b08/Makefile#L23 .

Yes, it’s fairly ugly that the decision needs to propagate all the way to the top level of building the binary, but that’s where the decision to enable or disable CGo happens, so that’s also the right place to make an informed decision and opt in.

paralin:
@mtrmac Adding a build tag conditional on cgo support is not always easy, we don't know if cgo is available... the standard practice is to use a build tag for this in the projects I've worked on.

The expectation is that "go build" never fails if all the tags are unset.

Maybe it would be better to just stub it on platforms that don't support cgo and return errors?

PR #1633, opened 2022-08-22
Delete partial docker-archive files, to allow retries

mtrmac:
Per https://github.com/containers/skopeo/issues/1708 ,  `skopeo copy --retry-times 3 docker://… docker-archive:…` cannot retry because the first failing copy creates a partial archive, and the subsequent retry refuses to modify it (we only support creating archives from scratch).

So, if we are creating a regular file, and we have not _successfully_ written at least one image, delete the destination file.

_Note:_ If the destination already exists as a zero-sized file, this deletes it. That’s a weird corner case — right now we always open the destination with `O_CREATE`, and just use that file whether it is a pipe or a regular file. Conceptually, we could do the check for non-regular files and non-empty destinations separately, and then use a temporary file + `Rename()` on success if the destination is a regular file. That’s a possible future enhancement — it would be clearer but I’m not sure if there is any situation where that actually _matters_ — at the very least it’s not worse than the current version.

### DIFF
--- a/docker/archive/writer.go
+++ b/docker/archive/writer.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"sync"
 
 	"github.com/containers/image/v5/docker/internal/tarfile"
 	"github.com/containers/image/v5/docker/reference"
@@ -13,9 +14,14 @@ import (
 
 // Writer manages a single in-progress Docker archive and allows adding images to it.
 type Writer struct {
-	path    string // The original, user-specified path; not the maintained temporary file, if any
-	archive *tarfile.Writer
-	writer  io.Closer
+	path        string // The original, user-specified path; not the maintained temporary file, if any
+	regularFile bool   // path refers to a regular file (e.g. not a pipe)
+	archive     *tarfile.Writer
+	writer      io.Closer
+
+	// The following state can only be acccessed with the mutex held.
+	mutex     sync.Mutex
+	hadCommit bool // At least one successful commit has happened
 }
 
 // NewWriter returns a Writer for path.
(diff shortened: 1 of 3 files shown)


docker/archive/writer.go
```diff
@@ -63,6 +79,19 @@ func (w *Writer) Close() error {
...
+	if err == nil && w.regularFile && !w.hadCommit {
+		// Writing to the destination never had a success; delete the destination if we created it.
+		// This is done primarily because we don’t implement adding another image to a pre-existing image, so if we
+		// left a partial archive around (notably because reading from the _source_ has failed), we couldn’t retry without
+		// the caller manually deleting the partial archive. So, delete it instead.
+		//
+		// Archives with at least one successfully created image are left around; they might still be valuable.
+		//
+		// Note a corner case: If there _originally_ was an empty file (which is not a valid archive anyway), this deletes it.
+		if err2 := os.Remove(w.path); err2 != nil {
```
rhatdan, line 91: Why create err2?  Why not just return os.Remove(w.path)? Or set it to 
if err = os.Remote ...

mtrmac: I’m actually a bit struggling with the logic of which errors should or should not cause this `Remove` … e.g. if writing the final metadata stream failed, we might still want to remove the unusable archive, but report primarily the metadata write failure back to the user.

I suppose I should just do the extra bit and use a temporary file with `Rename` here, either way…

vrothberg: I am much in favor in naming errors rather than enumerating them in such cases. `$subjectError` is much easier for me to brain parse than `err$Number`.

vrothberg (commented):
LGTM

rhatdan:
Not worth worrying about.

LGTM

PR #1635, opened 2022-08-23
Document limitations of transports for remote podman client

graywolf-at-work:
When using remote podman client, not all transports work as expected. So
document this limitation.

Fixes: containers/podman#15141
Signed-off-by: Tomas Volf <tomas.volf@showmax.com>

### DIFF
--- a/docs/containers-transports.5.md
+++ b/docs/containers-transports.5.md
@@ -11,6 +11,11 @@ containers-transports - description of supported transports for copying and stor
 Tools which use the containers/image library, including skopeo(1), buildah(1), podman(1), all share a common syntax for referring to container images in various locations.
 The general form of the syntax is _transport:details_, where details are dependent on the specified transport, which are documented below.
 
+The semantics of the image names ultimately depend on the environment where
+they are evaluated. For example: if evaluated on a remote server, image names
+might refer to paths on that server; relative paths are relative to the current
+directory of the image consumer.
+
 ### **containers-storage**:[**[**storage-specifier**]**]{image-id|docker-reference[@image-id]}
 
 An image located in a local containers storage.


vrothberg (commented):
Thanks!  I think this should be documented rather on the Podman side.

mtrmac:
I’m kind of torn. On the one hand, this is certainly a problem users might run into.

On the other hand, “might not work as expected” is very much depending on one’s point of view — I could very well argue that `podman --remote … transport:/local/path` is “expected” to refer to files on the server. So we end up getting closer to “warning: the meaning of the image name might depend on the context/location where it is evaluated”, which is… kind of stating something that is always true. (This is certainly an exaggeration … but clarifying what we mean exactly might take a bit more thought.)

---

Actually, do we really want to _commit to_ these things referring to the server, by documenting the behavior to be like that? Or is it conceptually a bug that we might want to fix, changing the semantics of these names in `podman --remote` operation again?

Even if it is a known bug we don’t anticipate quickly fixing, it might be worth documenting — but the documentation would be quite different in the two cases.

---

Alternatively, if this is conceptually a known bug, it might make sense for Podman remote to explicitly reject the problematic transports. It’s already doing that, after all, in https://github.com/containers/podman/blob/ee2f8155cc85705787900db71f4154d358bd419d/pkg/api/handlers/libpod/images_pull.go#L55 .

docs/containers-transports.5.md
```diff
@@ -11,6 +11,10 @@ containers-transports - description of supported transports for copying and stor
 Tools which use the containers/image library, including skopeo(1), buildah(1), podman(1), all share a common syntax for referring to container images in various locations.
 The general form of the syntax is _transport:details_, where details are dependent on the specified transport, which are documented below.
 
+Some transports might not work as expected when used in combination with a
+remote podman client. For example, `from oci-archive:/x.tar` in Containerfile
```
TomSweeneyRedHat, line 15:
```suggestion
remote podman client. For example, `from oci-archive:/x.tar` in the Containerfile
```

vrothberg:
In order to fix https://github.com/containers/podman/issues/15141 I expected only a docs change for `podman build` stating that the file must be part of the build context.

I agree with @mtrmac that "does not work as expected" needs to be more precise.  In general, file-based transports will resolve to paths on the _server_.  This information look worth adding to the transports man page.

mtrmac:
Something vaguely like
> The semantics of the image names ultimately depend on the environment where they are evaluated. For example: if evaluated on a remote server, image names might refer to paths on that server; relative paths are relative to the current directory of the image consumer.

?

vrothberg:
SGTM :heavy_check_mark:

mtrmac (approved):
LGTM. Thanks!

mtrmac:
@graywolf-at-work Can you rebase this on top of the current `main` branch, please? That is likely to help with the Skopeo test failures.

vrothberg (commented):
LGTM, thank you

PR #1636, opened 2022-08-23
Update a test dependency to avoid override problems

mtrmac:
> go get github.com/honeycombio/beeline-go@latest

github.com/honeycombio/beeline-go@v1.1.1 depends on github.com/gobuffalo/pop/v5@v5.3.1 , which depends on github.com/mattn/go-sqlite3@v2.0.3+incompatible , which has been untagged. pop/v5 mitigates that by using an override directive; that, in turn, only matters for top-level builds and not for libraries.

This ultimately causes failures when vendoring users of c/image with older versions of Go (not with Go 1.19) while trying to build the complete module graph.

Updating beeline-go changes the pop/v5 dependency to a pop/v6 version that depends on the correct github.com/mattn/go-sqlite3@v1.14.9 . Because beeline-go is only a test dependency, that doesn't really
affect most users of c/image (other than those that actually use some part of the dependency subtree),
but the unresolvable go-sqlite3@v2.0.3 reference should go away.


rhatdan:
LGTM

mtrmac:
So much of _fixing_ problems:
```
github.com/containers/skopeo/cmd/skopeo imports
…
	github.com/honeycombio/libhoney-go loaded from github.com/honeycombio/libhoney-go@v1.15.2,
	but go 1.16 would select v1.15.8
```

Oh well, let’s call it an investment: https://github.com/containers/skopeo/pull/1737

mtrmac:
Skopeo tests passed in https://github.com/containers/skopeo/pull/1737 .

PR #1638, opened 2022-08-24
signing: clarify build error when cgo is not available

### LINKED ISSUE #1634: Build fails if CGO_ENABLED=0
Cgo is not available on many platforms. libgpgme is used for signing.

#1632 (closed) proposed to enable openpgp when cgo is unavailable.

The package should compile when `CGO_ENABLED=0` - with some kind of alternative implementation to libgpgme that is go-only.

paralin:
The build fails if CGO_ENABLED=0:

```
# github.com/containers/image/v5/signature
signature/mechanism_gpgme.go:156:41: undefined: gpgme.PinEntryLoopback
signature/mechanism_gpgme.go:161:31: undefined: gpgme.Key
signature/mechanism_gpgme.go:161:31: too many errors
```

Make it clear that cgo is required by adding a comment & undefined variable:

```
# github.com/containers/image/v5/signature
./mechanism_other.go:10:21: undefined: ContainersImageSignatureRequiresCgo
```

Anyone who encounters this error (like me) will immediately understand the issue and can
read the comment on that line to understand that the alternative build tag is
not actively maintained and considered not secure.

```go
// CgoIsDisabled indicates as a compiler error that this package requires cgo.
//
// You may enable the containers_image_openpgp build tag but beware that this
// implementation is not actively maintained and is considered insecure.
var CgoIsDisabled = ContainersImageSignatureRequiresCgo
```

Fixes #1634

### DIFF
--- a/signature/mechanism_none.go
+++ b/signature/mechanism_none.go
@@ -1,0 +1,19 @@
+//go:build containers_image_disable_signing
+// +build containers_image_disable_signing
+
+package signature
+
+import "errors"
+
+// errSigningDisabled is returned if container image signing is disabled.
+var errSigningDisabled = errors.New("container image signing is disabled in this build")
+
+// newGPGSigningMechanismInDirectory returns an error indicating signing is disabled.
+func newGPGSigningMechanismInDirectory(optionalDir string) (SigningMechanism, error) {
+	return nil, errSigningDisabled
+}
+
+// newEphemeralGPGSigningMechanism returns an error indicating signing is disabled.
+func newEphemeralGPGSigningMechanism(blobs [][]byte) (SigningMechanism, []string, error) {
+	return nil, nil, errSigningDisabled
+}
--- a/signature/mechanism_other.go
+++ b/signature/mechanism_other.go
@@ -1,0 +1,10 @@
+//go:build !containers_image_openpgp && !cgo
+// +build !containers_image_openpgp,!cgo
+
+package signature
+
+// CgoIsDisabled indicates as a compiler error that this package requires cgo.
+//
+// You may enable the containers_image_openpgp build tag but beware that this
+// implementation is not actively maintained and is considered insecure.
+var CgoIsDisabled = ContainersImageSignatureRequiresCgo
--- a/signature/mechanism_other.go
+++ b/signature/mechanism_other.go
@@ -1,10 +1,26 @@
-//go:build !containers_image_openpgp && !cgo
-// +build !containers_image_openpgp,!cgo
+//go:build !containers_image_openpgp && !containers_image_disable_signing && !cgo
+// +build !containers_image_openpgp,!containers_image_disable_signing,!cgo
 
 package signature
 
 // CgoIsDisabled indicates as a compiler error that this package requires cgo.
 //
-// You may enable the containers_image_openpgp build tag but beware that this
-// implementation is not actively maintained and is considered insecure.
+// The containers_image_openpgp build tag enables the OpenPGP implementation but
+// beware that it is not actively maintained and is considered insecure.
+//
+// The containers_image_disable_signing build tag alternatively disables signing
+// and will return an error when attempting to sign an image at runtime.
+//
+// # github.com/containers/image/v5/signature
+// ./mechanism_other.go:16:21: undefined: ContainersImageSignatureRequiresCgo
 var CgoIsDisabled = ContainersImageSignatureRequiresCgo
+
+// newGPGSigningMechanismInDirectory cannot be compiled without Cgo.
+func newGPGSigningMechanismInDirectory(optionalDir string) (SigningMechanism, error) {
+	return nil, CgoIsDisabled
+}
+
+// newEphemeralGPGSigningMechanism cannot be compiled without Cgo.
+func newEphemeralGPGSigningMechanism(blobs [][]byte) (SigningMechanism, []string, error) {
+	return nil, nil, CgoIsDisabled
+}
--- a/signature/mechanism_test.go
+++ b/signature/mechanism_test.go
@@ -1,3 +1,6 @@
+//go:build cgo || containers_image_openpgp
+// +build cgo containers_image_openpgp
+
 package signature
 
 // These tests are expected to pass unmodified for _both_ mechanism_gpgme.go and mechanism_openpgp.go.
@@ -10,10 +13,6 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-)
-
-const (
-	testGPGHomeDirectory = "./fixtures"
 )
 
 // Many of the tests use two fixtures: V4 signature packets (*.signature), and V3 signature packets (*.signature-v3)
--- a/signature/policy_eval_sigstore_test.go
+++ b/signature/policy_eval_sigstore_test.go
@@ -1,6 +1,9 @@
-// Policy evaluation for prCosignSigned.
+//go:build cgo || containers_image_openpgp
+// +build cgo containers_image_openpgp
 
 package signature
+
+// Policy evaluation for prCosignSigned.
 
 import (
 	"context"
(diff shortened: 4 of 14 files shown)


paralin:
The second commit in this PR adds a new build tag to disable signing:

Possible use cases:

 - default: use cgo and enable libgpgme
 - CGO_ENABLED=0 -> build fails
 - CGO_ENABLED=0 and containers_image_openpgp -> build succeeds
 - CGO_ENABLED=0 and containers_image_disable_signing -> build succeeds but returns errors at runtime

The default is to compile in the endorsed cgo implementation and fail the build otherwise.

rhatdan:
@vrothberg @mtrmac WDYT?

signature/mechanism_test.go
```diff
@@ -1,3 +1,6 @@
+//go:build cgo || containers_image_openpgp
+// +build cgo containers_image_openpgp
```
mtrmac, line 2: If the goal is to make tests pass with `containers_image_disable_signing` (and I do think those tests should pass, because parts of c/image/signature are completely independent of the simple signing implementation), this doesn’t do that, and from a quick check it would be a rather larger effort — because individual expected results would have to be made conditional, correctly.

signature/mechanism_other.go
```diff
@@ -0,0 +1,13 @@
...
+package signature
+
+// CgoIsDisabled indicates as a compiler error that this package requires cgo.
+//
+// The containers_image_openpgp build tag enables the OpenPGP implementation but
+// beware that it is not actively maintained and is considered insecure.
+//
+// The containers_image_disable_signing build tag disables signing and will
+// return an error when attempting to sign an image at runtime.
+var CgoIsDisabled = ContainersImageSignatureRequiresCgo
```
mtrmac, line 13: Non-blocking: If we are in the business of grooming error messages, implementing `newGPGSigningMechanismInDirectory` and `newEphemeralGPGSigningMechanism` here, just to make this one error message stand out more, might be worthwhile.

signature/mechanism_none.go
```diff
@@ -0,0 +1,19 @@
+//go:build containers_image_disable_signing
+// +build containers_image_disable_signing
```
mtrmac, line 2: Something like
```suggestion
// +build containers_image_disable_simple_signing
```
(updated everywhere) because this does not prevent use of sigstore. (Cc: @rhatdan )

signature/mechanism_other.go
mtrmac, line 13: RequiresCGoAndGPGMEDevelopmentHeaders , or maybe ContainersImageSignatureReadREADMEOfContainersImage ?

I guess I don’t really care to polish this that much…

mtrmac (commented):
I’m fine with doing this, overall, though it does seem quite a bit of effort for what I think is a fairly small gain. Only users familiar enough with Go to use build tags are going to benefit, and those might find their way to the README either way.

I’m especially uncertain about the `disable_signing` part. If that tag is added:
- It should be documented in `README.md`, so that it has some chance of being useful to anyone else
- … and then I guess we need it to keep working, i. e. the tests must work (easily quadrupling the size of this PR), and the tests should be run in CI — and maybe we don’t want to pay for a complete separate CI step, so just run unit tests of `./signature/...` with that tag?

paralin:
> though it does seem quite a bit of effort for what I think is a fairly small gain

The PR in its current form is not a lot of effort and fixes a number of issues for us downstream in package managers like Buildroot.org.

Many of your comments are requesting changes that would increase the effort. I agree with you that those particular additions are a lot of extra effort for a very small gain. I commented above.

signature/mechanism_test.go
paralin: The goal was to make `go test` pass by excluding the signing tests unless a signing mechanism is compiled in.

Why would we need to run the signing tests when the entire signature mechanism is not compiled in (in the case of containers_image_disable_signing) and will always return errors?

signature/mechanism_other.go
paralin: Will do.

paralin: I've seen a few errors with this format - "RequresCgo" - in the past.

Main place I've seen a build failure enforce something about the version or go tags is here:

https://github.com/protocolbuffers/protobuf-go/blob/v1.28.1/runtime/protoimpl/version.go#L58

It used to generate something like `var ProtoVersionIsTooOld = undefined` but was recently changed to use Go's uint overflow checking to enforce the build failure.

signature/mechanism_none.go
paralin: @mtrmac I suppose what I meant by `disable_signing` is "the signing mechanism is disabled so you won't be able to create new signatures" - am I correct that disabling `mechanism*.go` does this?

paralin:
@mtrmac I've fixed the PR according to your comments - can update the name of the tag & the other few pending points as well if/when needed.

signature/mechanism_other.go
paralin: Okay, I've done that now. The error is now just:

```
./mechanism_other.go:16:21: undefined: ContainersImageSignatureRequiresCgo
```

mtrmac: The pedantic point was that CGo is not the _only_ requirement. But this works well enough already.

signature/policy_eval_sigstore_test.go
```diff
@@ -1,7 +1,10 @@
-// Policy evaluation for prCosignSigned.
+//go:build cgo || containers_image_openpgp
```
mtrmac, line 1: This is the other signing implementation (“sigstore”) and needs to work, so it needs to be tested, even if “simple signing” is disabled.

signature/policy_eval_simple_test.go
```diff
@@ -1,3 +1,6 @@
+//go:build cgo || containers_image_openpgp
```
mtrmac, line 1: This contains things like “reject all images from this repo” and needs to work, so it needs to be tested, even if “simple signing”  is disabled.

signature/policy_eval_test.go
```diff
@@ -1,3 +1,6 @@
+//go:build cgo || containers_image_openpgp
```
mtrmac, line 1: This needs to work in all cases.

signature/mechanism_none.go
mtrmac: c/image/signature does not _only_ do cryptography; it includes policies like “always allow” and “always reject”.

Of the cryptographic mechanisms, it now actually implements _two: “simple signing”, which requires GPG, and “sigstore”, which doesn’t. All of the discussion so far was about disabling / choosing a mechanism underlying the “simple signing” part; “sigstore” is unaffected.

So, the tag name should refer (in some way) to simple signing in particular, and not to the concept of signing in general.

mtrmac (commented):
On second thought (which should have come with the first review already, and I apologize for not thinking this through), do we want the disable_signing tag at all?

If I understand it correctly, you don’t intend to use it.
I don’t know of anyone who wants to use it either.

And making that work without compromising the testing of the rest of the `policy.json` evaluation would probably be more work than just disabling test files; and _some_ ongoing maintenance cost (if nothing else, the cost of computers running tests). If no-one benefits, is that work worthwhile?

So maybe let’s just add the `CGoIsDisabled` error reporting mechanism that you truly want, and not the other part.

---

WRT the tag name, (see https://github.com/containers/image/pull/1638#discussion_r956591992 for background), assuming the tag remains, I think `…disable_simple_singing` would be an accurate name but I’d prefer a second opinion to confirm.

rhatdan:
@paralin Still working on this?

paralin:
@rhatdan I'll revisit this shortly

mtrmac:
Over a year later … `containers_image_disable_signing` in such generality seems unnecessary; and things like `prReject`/`prInsecureAcceptAnything` _must_ keep working.

Sigstore keeps getting more popular; it’s not conceivable that we might eventually want to allow disabling the simple signing features — but this PR didn’t get us nearer to a working implementation.

PR #1646, opened 2022-09-02
Update comment to conform to Go 1.19 formatting

mtrmac:
Also start with the function name, as linters prefer.


mtrmac:
This is necessary to allow local development of any other work with Go 1.19.

PR #1653, opened 2022-09-12
Log invalid and empty credential helper entries

mtrmac:
In these cases we would not have _any_ log entry about the file, confusing log readers who expected to see something.

Also, both of these situations are non-obvious and worth calling attention to.


rhatdan:
LGTM
@vrothberg PTAL

vrothberg (commented):
LGTM

Feel free to merge

PR #1672, opened 2022-10-05
WIP: docker/docker_image_src: Support partial signature lookups

wking:
Sometimes we find some signatures, but hit an error before deciding we've retrieved "all the signatures".  But "all the signatures" isn't a well-defined thing.  Additional signatures may become available at any time.  This commit adjusts the signature-fetching paths to allow for "I had some trouble, so I'm not completely satisfied, but these are the signatures I did find...", and maybe one of those will be acceptable to the caller.

### DIFF
--- a/docker/docker_image_src.go
+++ b/docker/docker_image_src.go
@@ -400,26 +400,23 @@ func (s *dockerImageSource) GetSignaturesWithFormat(ctx context.Context, instanc
 	switch {
 	case s.c.supportsSignatures:
 		sigs, err := s.getSignaturesFromAPIExtension(ctx, instanceDigest)
-		if err != nil {
-			return nil, err
-		}
 		res = append(res, sigs...)
+		if err != nil {
+			return res, err
+		}
 	case s.c.signatureBase != nil:
 		sigs, err := s.getSignaturesFromLookaside(ctx, instanceDigest)
-		if err != nil {
-			return nil, err
-		}
 		res = append(res, sigs...)
+		if err != nil {
+			return sigs, err
+		}
 	default:
 		return nil, errors.New("Internal error: X-Registry-Supports-Signatures extension not supported, and lookaside should not be empty configuration")
 	}
 
 	sigstoreSigs, err := s.getSignaturesFromSigstoreAttachments(ctx, instanceDigest)
-	if err != nil {
-		return nil, err
-	}
 	res = append(res, sigstoreSigs...)
-	return res, nil
+	return res, err
 }
 
 // manifestDigest returns a digest of the manifest, from instanceDigest if non-nil; or from the supplied reference,
--- a/internal/image/unparsed.go
+++ b/internal/image/unparsed.go
@@ -108,7 +108,7 @@ func (i *UnparsedImage) Signatures(ctx context.Context) ([][]byte, error) {
 
 // UntrustedSignatures is like ImageSource.GetSignaturesWithFormat, but the result is cached; it is OK to call this however often you need.
 func (i *UnparsedImage) UntrustedSignatures(ctx context.Context) ([]signature.Signature, error) {
-	if i.cachedSignatures == nil {
+	if i.cachedSignatures == nil { // FIXME: this is going to be sticky...
 		sigs, err := i.src.GetSignaturesWithFormat(ctx, i.instanceDigest)
 		if err != nil {
 			return nil, err
--- a/internal/imagesource/impl/compat.go
+++ b/internal/imagesource/impl/compat.go
@@ -42,14 +42,11 @@ func (c *Compat) GetSignatures(ctx context.Context, instanceDigest *digest.Diges
 	// Alternatively, we could possibly define the old GetSignatures to use the multi-format
 	// signature.Blob representation now, in general, but that could silently break them as well.
 	sigs, err := c.src.GetSignaturesWithFormat(ctx, instanceDigest)
-	if err != nil {
-		return nil, err
-	}
 	simpleSigs := [][]byte{}
 	for _, sig := range sigs {
 		if sig, ok := sig.(signature.SimpleSigning); ok {
 			simpleSigs = append(simpleSigs, sig.UntrustedSignature())
 		}
 	}
-	return simpleSigs, nil
+	return simpleSigs, err
 }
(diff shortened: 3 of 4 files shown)


docker/docker_image_src.go
```diff
@@ -400,26 +400,23 @@ func (s *dockerImageSource) GetSignaturesWithFormat(ctx context.Context, instanc
 	switch {
 	case s.c.supportsSignatures:
 		sigs, err := s.getSignaturesFromAPIExtension(ctx, instanceDigest)
+		res = append(res, sigs...)
 		if err != nil {
-			return nil, err
+			return res, err
```
wking, line 405: and we may not even want to fail-fast here, in case `getSignaturesFromSigstoreAttachments` has any luck finding signatures a few lines down.  If we go that route, we would want to collect and aggregate errors from any troubled strategies.

wking:
A larger refactor could move us to an interface like [this callback system][1] I wrote for the cluster-version operator, allowing the caller to view and judge verified-ness as signatures are detected, instead of waiting until "all" the signatures are returned.

[1]: https://github.com/openshift/library-go/blob/0d713c6520e279077b8988a4643e1c343a26ad9b/pkg/verify/store/store.go#L34

mtrmac (commented):
I totally accept that there are situations where something like this would be desirable — where signatures _only ever_ exist for signature presence enforcement, and where finding one valid signature is enough, and failures reading other signatures can then be ignored.

The trouble is:

- This code does not know that:

   `dockerImageSource` is used both for “pulls” and for “copies” (e.g. registry-to-registry mirroring); and for “copies”, an unreadable signature really is exactly as hard a failure as an unreadable layer.

  That’s, to an extent, fixable, though pedantically speaking the difference between a “pull” and a “copy” is only in the users’ intent.

- Even for pulls, it’s _usually_ not the right thing to do.

  On pulls, signatures are not used _only_ for enforcement of `policy.json`; all of them are also recorded along with the image in local storage, even if `policy.json` is not enforcing any signature existence at all.

   The signature provides “nonrepudiation”, i.e. a proof that some specific party has created the image. So, it can be the case that the policy requires an internal-IT “approved to run” signature, without specifically enforcing the vendor’s (e.g. RH) signature — but if later a compromised computer is forensically examined, and the image turns out to have a provable backdoor, it is very _relevant_ to have a copy of the RH signature on record locally (where RH can’t retroactively delete it).

   This will, probably, in the future similarly apply to other sigstore-style attachments, like SBOM files, attestations, and other metadata, where a consumer might not necessarily (fully, or at all?) enforce their validity but they nevertheless should be recorded.

- This is not, observably, a server failure, it’s a file format failure.

   ```
   parsing signature https://…/signature-5: unrecognized signature format, starting with binary 0x3c
   ```

  Per https://github.com/containers/image/blob/d0d5eef85d803da8cdc5dfe1acaac0b7efc8a275/docker/docker_image_src.go#L500 , signatures are only parsed if the server returns HTTP status 200. So we got 200 and an invalid file contents… and then it gets ugly. What is the right `Signature.FormatID` value of this file? Where should it be stored, for destinations that store different formats differently?

   A software dealing with that _can_ be built, sure. But the context is a container image built with content-addressed layers, where a server returning 200 + invalid data would _definitely_ result in a digest validation failure (and conceivably an alert about a possible attack), with no-one contemplating that the software should be build to accept, copy, and preserve, that known-invalid data. So it’s rather unattractive to me to build that special-case code for signatures; I think of signatures as an “integral part of the image”.

---


(This is not a review of the details of the code at all.)

mtrmac:
Notes to follow up on from the Slack discussion:
- Heuristically look for `Content-type: text/html`, and warn. (Unsure whether to just skip the signature and continue anyway.)
- A design goal of `policy.json`’s signature verification was that unrecognized signatures, or even signatures with a recognized format but clearly invalid, are not a reason to reject an image that has other valid signatures. That property was lost with the new signature format abstraction — it’s not _immediately_ clear whether that’s a good thing. Maybe we should build all that “completely unknown signature format, but preserve it” infrastructure after all. Or maybe not, we really want signatures to have a format ID of _some_ kind, and the current “raw blob which is assumed to be GPG but could be HTML” can be argued to just be a historical mistake.

mtrmac:
> to follow up

@jnovy please consider updating the `containers-common` files `registry.redhat.io.yaml` and `registry.access.redhat.com.yaml` to `s/sigstore/lookaside/g` .

mtrmac:
> Notes to follow up on from the Slack discussion:

https://github.com/containers/image/pull/1673

mtrmac:
On

> A design goal of policy.json’s signature verification was that unrecognized signatures, or even signatures with a recognized format but clearly invalid, are not a reason to reject an image that has other valid signatures.

vs.

> On pulls, signatures are not used only for enforcement of policy.json; all of them are also recorded along with the image in local storage

:

---

The “unrecognized signatures are not a reason to reject an image” goal is not because being lax in format handling is valuable (in security, anything unrecognized is suspect and more likely an attack than “nothing to worry about”; as a matter of principle weird-looking data should be rejected by default, and only allowed with a good reason); it’s for forward-compatibility with future file formats unknown to the current implementation.

With that perspective, we now have two “signature” formats: the old one is a blob with no format identifier (where we can either find a valid GPG signature or assume the contents to be “something else”), the new one is a (format ID, blob) pair, with a single-blob serialization defined (and an _optionally-used_ parser that can consume both old-format blobs and new-format blobs).

The new _format_ is inherently forward compatible (while individual sources/destinations might restrict the “format ID” values they support). Our current _implementation_ of that format is not forward-compatible; it flatly refuses to parse data with unknown format IDs. We should change that (https://github.com/containers/image/issues/1674 ), but it’s fairly low-priority compared to other things going on currently.

Now that we _have_ the new format with an explicit format ID, it doesn’t really make much sense to treat the old format as a forward-compatibility mechanism. At least to my knowledge, there were never any users of the signature storage with non-GPG data; any new users with new kinds of image attachments/metadata would be directed to the new format.

So, based on the reasoning above, I don’t think it makes sense, any more, to be “forward-compatible” by allowing arbitrary blobs with unknown format that don’t contain a format ID. So, the current situation (a lookaside server returning 200 + HTML) is much closer to “a server serving flatly incorrect data” or perhaps “a server clearly violating protocol / file format definitions” than “a server using a legitimate forward compatibility mechanism, we are just unlucky that there was a failure and the data is unusable”.

And with all that, because we do want to record all signatures for forensic purposes, I think not failing when a signature can’t be parsed this way is not the right tradeoff.

PR #1675, opened 2022-10-06
Introduce oci/{archive,layout}.ImageNotFoundError

mtrmac:
… for https://github.com/containers/skopeo/pull/1757 .

Warning: Absolutely untested.

### DIFF
--- a/oci/archive/oci_src.go
+++ b/oci/archive/oci_src.go
@@ -17,6 +17,17 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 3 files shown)


cgwalters (commented):
Awesome, thanks for doing this!

mtrmac:
(It would be fair for this to only go in _after_ #1381, which has been pending for a long time — so that that PR is not forced to deal with one more feature. Let me see if I can make progress on that one now.)

mtrmac:
Looking at #1381, it needs a few more changes — so let’s not block this one on it.

PR #1683, opened 2022-10-13
Recognize invalid error responses of registry.redhat.io

mtrmac:
... when checking for missing images.

In particular, this is necessary for `use-sigstore-attachments` not to cause failures when pulling from registry.redhat.io.

Red Hat internal reference: RITM1310318

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -1,6 +1,7 @@
 package docker
 
 import (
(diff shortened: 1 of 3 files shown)


vrothberg (commented):
LGTM

rhatdan:
LGTM

PR #1481, opened 2022-02-24
Change OCI layout transport to thread-safe destination

Madeeks:
This PR follows up the discussion in #1466.

I did another round of practical tests with Skopeo 1.6.1 built on top of this change.
I was able to successfully copy to OCI layout format the following images: 

| Image reference                                                        | Number of blobs | Size on registry |
|------------------------------------------------------------------------|-----------------|------------------|
| `quay.io/madeeks/pyfr:1.10.0-cuda11.2-openmpi4.1.0-ubuntu20.04`          |        18       |       2.3GB      |
| `quay.io/madeeks/horovod:0.22.0-tf2.5.0-cuda11.2-mpich3.1.4-ubuntu18.04` |        23       |       4.1GB      |
| `quay.io/madeeks/gromacs:2020.4-cuda11.2.0-mpich3.1.4-centos7`           |        15       |       2.3GB      |
| `nvcr.io/nvidia/tensorflow:22.01-tf2-py3`                                |        44       |      6.66GB      |
| `nvcr.io/nvidia/nvhpc:22.1-devel-cuda_multi-ubuntu20.04`                 |        13       |      8.14GB      |

Closes #1466

PR #1685, opened 2022-10-13
Set Directory Transport Destination as Thread-Safe

t0rr3sp3dr0:
Just like #1481 did for OCI Layout Destination, this patch marks Directory Transport Destination as thread-safe. The `PutBlobWithOptions` implementation of both transports are almost identical, differing only on the path where the blob will be stored.

```diff
- 	blobPath := d.ref.layerPath(blobDigest)
+ 	blobPath, err := d.ref.blobPath(blobDigest, d.sharedBlobDir)
+	if err != nil {
+		return types.BlobInfo{}, err
+	}
+	if err := ensureParentDirectoryExists(blobPath); err != nil {
+		return types.BlobInfo{}, err
+	}
```


mtrmac (approved):
LGTM. Thanks!

PR #1690, opened 2022-10-17
[CI:DOCS] Update renovate config. comment

cevich:
Discovered the golang-version update behavior is recent and up for some debate.  Reference open issue on the topic in case the config. needs to be corrected in the future due to a fix.


mtrmac (approved):
Thanks!

PR #1691, opened 2022-10-18
[CI:DOCS] Move boiler-plate golang config to automation repo

cevich:
This saves having to repeat the same options over and over in every
repo.

Ref: https://github.com/containers/automation/commit/56579d17501230b1b9a230c903fe6acf18872033


cevich:
drafting this: I think maybe this rule is general enough I should just add it into the preset instead of copy-pasting it everywhere.

cevich:
force-push: Add reference to moved configuration.

PR #1704, opened 2022-10-26
Fix a panic on scheme-less lookaside base

mtrmac:
… because we shadow the `net/url` package by a `url` local variable. To be extra sure, rename all `url` variables.

Fixes https://github.com/containers/podman/issues/16301 .


vrothberg (commented):
LGTM

A nice lesson on interface design in Go. Having a function match the name of the method of an object that in turn has the same name as the package invites for the trap we fell into.

mtrmac:
Sadly, none of the linters of golangci-lint seems to warn about variable shadowing of this kind. So I don’t know how to systematically prevent it.

We could probably, somehow, prohibit variables named `url`, but that seems way too specific to be worth the effort and maintenance hassle.

rhatdan:
LGTM

mtrmac:
> A nice lesson on interface design in Go.

Also: Nullable types by default are a mistake.

PR #1707, opened 2022-10-30
Add copy option to strip sparse manifests

### LINKED ISSUE #1661: Support stripping dropped instances from manifests when making sparse multi-platform copies
Hi,

Not sure if this is a bug or the intended behaviour so asking for clarification. When using the `CopySpecifcImages` it will copy the images listed in the `Instances` field but then try to put the entire manifest list into the destination which will only have the updated details for those copied as part of the `Instances`. 

This will result in seeing `manifest blob unknown: blob unknown to registry` error, because there will be blobs referenced in the manifest that have not been included in the copy because they are not in the `Instances` list.

The language in the documentation for `CopySpecifcImages` does have the line "along with the list itself". Is this useful behaviour?

The usecase for this is to be able to copy only specific architectures from a multi-arch manifest rather than copying across all which can be costly storage. 1-2 GB wasted from unrequired archs.

bertbaron:
Not all registries accept manifests where some images are missing. Therefore an option is added to strip missing images from the manifest when only a subset of images is copied.

This is a first change toward adding a filter on architecture in skopey copy and sync. See https://github.com/containers/skopeo/issues/1694.

I'd be happy to add documentation and tests when this change is acceptable.

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -169,6 +169,9 @@ type Options struct {
 	// Download layer contents with "nondistributable" media types ("foreign" layers) and translate the layer media type
 	// to not indicate "nondistributable".
 	DownloadForeignLayers bool
+
+	// When not all images of a manifest list are copied, strip the other images from the list. The digest will differ from the original if so.
+	StripManifestList bool
 }
 
 // validateImageListSelection returns an error if the passed-in value is not one that we recognize as a valid ImageListSelection value
@@ -467,6 +470,7 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
 	}
 	c.Printf("Copying %d of %d images in list\n", imagesToCopy, len(instanceDigests))
 	updates := make([]manifest.ListUpdate, len(instanceDigests))
+	var skipped []digest.Digest
 	instancesCopied := 0
 	for i, instanceDigest := range instanceDigests {
 		if options.ImageListSelection == CopySpecificImages {
@@ -485,6 +489,7 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
 				logrus.Debugf("Skipping instance %s (%d/%d)", instanceDigest, i+1, len(instanceDigests))
 				// Record the digest/size/type of the manifest that we didn't copy.
 				updates[i] = update
+				skipped = append(skipped, instanceDigest)
 				continue
 			}
 		}
@@ -508,6 +513,15 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
 	// Now reset the digest/size/types of the manifests in the list to account for any conversions that we made.
 	if err = updatedList.UpdateInstances(updates); err != nil {
 		return nil, fmt.Errorf("updating manifest list: %w", err)
+	}
+
+	// Remove skipped images from the manifest if StripManifestList == true
+	if options.StripManifestList {
+		for _, d := range skipped {
+			logrus.Debugf("Removeing instance %s from manifest list", d)
+
+			updatedList.RemoveInstance(d)
+		}
 	}
 
 	// Iterate through supported list types, preferred format first.
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -119,6 +119,22 @@ const (
 // specific images from the source reference.
 type ImageListSelection int
 
+const (
+	// SparseManifestListAction is the default value which, when set in
+	// Options.SparseManifestListAction, indicates that the manifest is kept
+	// as is even though some images from the list may be missing. Some
+	// registries may not support this.
+	KeepSparseManifestList SparseManifestListAction = iota
+
+	// StripSparseManifestList will strip missing images from the manifest
+	// list. When images are stripped the digest will differ from the original.
+	StripSparseManifestList
+)
+
+// SparseImageListAction is one of KeepSparseManifestList or StripSparseManifestList
+// to control the behavior when only a subset of images from a manifest list is copied
+type SparseManifestListAction int
+
 // Options allows supplying non-default configuration modifying the behavior of CopyImage.
 type Options struct {
 	RemoveSignatures                 bool            // Remove any pre-existing signatures. SignBy will still add a new signature.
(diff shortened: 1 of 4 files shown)


manifest/list.go
```diff
@@ -33,6 +33,10 @@ type List interface {
 	// must be specified.
 	UpdateInstances([]ListUpdate) error
 
+	// RemoveInstance removes the instance with the given digest from the list. If no such instance
+	// exists the list is left unchanged.
+	RemoveInstance(digest.Digest)
```
mtrmac, line 38: This should probably return an error on a missing digest (but because it start as an internal-only interface, it’s not critical to get right).

copy/copy.go
```diff
@@ -467,6 +487,7 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
 	}
 	c.Printf("Copying %d of %d images in list\n", imagesToCopy, len(instanceDigests))
 	updates := make([]manifest.ListUpdate, len(instanceDigests))
+	var skipped []digest.Digest
```
mtrmac, line 490: Identifying instances by digest is fine right now.

For the per-arch filtering anticipated in #1694, I wonder if that is sufficient; it could, in principle, be the case that two different instances (most likely the same basic platforms, but different variants) point at the same instance digest — and in that case we might end up wanting to remove one pointer to that digest but not the other one. In that case we might need to identify instances by the index, not by the digest (and remove them in the right order).

I suppose that’s not a concern for _this_ PR, and a future PR that adds the per-arch filter could deal with that… but it might be simpler to change the design right now.

mtrmac (changes requested):
Thanks. Sadly `manifest.List` is a public API and adding a method to it could break external callers which implement this interface, so we can’t just add a method.

This is why we have been moving towards making most interfaces private instead — compare `image` vs. `internal/image`. So adding this method would first require introducing something like `internal/manifest.List`, moving at least the code that refer to the internal interface to `internal/manifest` (and I’m not sure about the two actual implementations), while preserving the current external API in a way that minimizes duplication.

copy/copy.go
bertbaron: Indeed, I had the same concern. But this could also affect List.Instance(digest.Digest). Its because of that method that I implemented it this way.
As an alternative I could remove it by passing a slice of integer indices or something like that.

manifest/list.go
bertbaron: I'm afraid my knowledge of go and this project make it difficult for me to perform such a refactoring. So instead I moved the functionality to a private method using a type switch. That way it has no impact on the public interface and can it safely be refactored in a future PR.

bertbaron: > This should probably return an error on a missing digest (but because it will start as an internal-only interface, it’s not critical to get right).

I personally prefer the more idempotent type of behavior (why return an error if the end result is what you want it to be), but I'm fine of course to implement it differently. Since this is internal now indeed I left it as is.

copy/copy.go
bertbaron: Images are removed by index now

copy/copy.go
```diff
@@ -169,6 +185,10 @@ type Options struct {
 	// Download layer contents with "nondistributable" media types ("foreign" layers) and translate the layer media type
 	// to not indicate "nondistributable".
 	DownloadForeignLayers bool
+
+	// When only a subset of images of a list is copied, this action indicates if the manifest should be kept or stript.
```
TomSweeneyRedHat, line 189: Not often that someone sneaks a new word past me, but you did with "stript".   It's fine, but I wonder if more people would understand "stripped" even though they're equivalent....

TomSweeneyRedHat:
Needs a rebase @bertbaron

copy/copy.go
bertbaron: Fixed. It wasn't intentional and I used 'stripped' elsewhere. But it's good to know that others may learn from my lack of knowledge of the English language.

bertbaron:
Is there a reason why this PR still can not be merged?
I did want to add some tests, but since I couldn't find existing tests for the copy function that could serve as examples this would require more effort than anticipated and I didn't find the time yet.

copy/copy.go
```diff
@@ -390,6 +410,33 @@ func compareImageDestinationManifestEqual(ctx context.Context, options *Options,
 	return true, destManifest, destManifestType, destManifestDigest, nil
 }
 
+// removeInstanceFromList removes the given image from the list
+// this should probably move to an internal API
+func removeInstancesFromList(manifestList manifest.List, imageIndices map[int]bool) error {
```
mtrmac, line 415: No; let’s build the infrastructure the right way rather than piling on tech debt.

It will never be any easier to get that done, and each such “small temporary thing” makes it harder.

mtrmac (commented):
I’m sorry, I’m afraid I’m mostly just busy with other work.

copy/copy.go
bertbaron: I appreciate your way of thinking, but its just that I'm a bit lost at how to do it. I can't move the public interface of course to not break it. I can extend it (embed it in an internal one). But I also can't move the implementation structs I guess because those also seem to be public. I can extend those with the internal interface, but then the methods would still be public on the struct and I'm not sure if that's ok. And is it possible in go to do something like return type overloading to get methods like Clone() right?
With some pointers I might be able to perform the refactoring but at this point I'm a bit stuck.

mtrmac: Yes, this is the hard part of all this work.

Unlike interfaces, adding new methods to structs is ABI-safe. So my first guess is that the `manifest.List` interface, and the methods that create that (`ListFromBlob`) should be ~duplicated in internal packages. The implementations would _probably_ also be moved to internal-only, with public aliases (`type OCI1Index = internal.OCI1Index`), but I’m _much_ less certain about this part.

---

Alternatively… I think we’ll need to make these refactors for other purposes (adding/removing zstd-compressed variants) fairly soon anyway (but I’m not exactly sure when). So it might be an option to wait with this PR until that infrastructure work is done by other people.

Cc: @vrothberg .

bertbaron: I tried but didn't succeed. Apparently go requires return types of a method to be equal in order to implement an interface, a subtype is not supported (I think its the lack of covariance support as described here: https://github.com/golang/go/issues/30602). So the `Clone` method for example has to return a public `List` and not an internal one. But it also doesn't allow a more specific `ListUpdate` to be returned for example.
One option I considered was to add methods like `CloneInternal()` to the private interface. But that seems to be impossible without introducing package cycles because the public interface and public creator method are in the same package. The only workaround/hack I can think of is to do some dynamic registration to redirect the public creator to the private creator.
On the other hand, while I learned more about go with this exercise there's still a lot more for me to learn so I might have overlooked some other options. So unless someone can give me a clear pointer to something I missed I think I'll give up for now.

jankaluza:

PR #1711, opened 2022-10-31
fix sigstore verify failed

hongxu-jia:
While sigstore (cosign) enabled, sign a container image in registry R1, and then copy it to registry R2, if you pull it from registry R2 with sigstore (cosign) verify enabled, the verification will fail

While cosign signing, it saves reference info to sigstore attachment manifest as the meta data (payload) of signed container image.

While skopeo or podman applying cosign verify, it needs the reference info to fetch signed data.

If a container image is signed in registry R1, it records reference R1 to sigstore attachment manifest; if the container image is copied to R2, reference R1 in sigstore attachment manifest is obsolete.

Due to the sigstore attachment manifest could not be edited and it is along with container image in registry, in this situation, if reference from sigstore attachment manifest does not work, try to use image reference to fetch signed data again

Signed-off-by: Hongxu Jia <hongxu.jia@windriver.com>

### DIFF
--- a/signature/policy_eval_sigstore.go
+++ b/signature/policy_eval_sigstore.go
@@ -52,8 +52,10 @@ func (pr *prSigstoreSigned) isSignatureAccepted(ctx context.Context, image priva
 
 	signature, err := internal.VerifySigstorePayload(publicKey, sig.UntrustedPayload(), untrustedBase64Signature, internal.SigstorePayloadAcceptanceRules{
 		ValidateSignedDockerReference: func(ref string) error {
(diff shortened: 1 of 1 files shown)


mtrmac (changes requested):
Thanks for the PR.

This would pretty much _completely break_ the point of reference verification, which is to match the signed identity (`ref`) with the user-desired identity (`image.Refrence().DockerReference()`).

Matching the user-desired identity against the user-desired identity is completely pointless.

The policy configuration already contains sufficient configuration features (`signedIdentity: {exactReference,exactRepository,remapIdentity`) that allow this to work.

PR #1725, opened 2022-11-11
signature: Export the default policy path

dfr:
This will allow using the default policy path from here in c/podman and c/common instead of duplicating the constant in those projects.

Signed-off-by: Doug Rabson <dfr@rabson.org>

### DIFF
--- a/signature/policy_config.go
+++ b/signature/policy_config.go
@@ -50,15 +50,15 @@ func (err InvalidPolicyFormatError) Error() string {
 // NOTE: When this function returns an error, report it to the user and abort.
 // DO NOT hard-code fallback policies in your application.
 func DefaultPolicy(sys *types.SystemContext) (*Policy, error) {
-	return NewPolicyFromFile(defaultPolicyPath(sys))
+	return NewPolicyFromFile(DefaultPolicyPath(sys))
 }
 
-// defaultPolicyPath returns a path to the default policy of the system.
-func defaultPolicyPath(sys *types.SystemContext) string {
+// DefaultPolicyPath returns a path to the default policy of the system.
+func DefaultPolicyPath(sys *types.SystemContext) string {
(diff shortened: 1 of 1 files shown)


mtrmac (commented):
Thanks.

I’d prefer not to do this.

- c/common AFAICS has really no reason to set the defaults; it should just set the field unset, for `DefaultPolicy` to work as expected. That would also ensure that the c/common defaults and c/image don’t diverge (as they apparently already did?)

- What Podman’s pkg/trust does should not really be _easy_, because that copy of code does not keep up with the actual policy language, and it can misinterpret it. That’s happened this year. If anything should happen, I’d prefer the relevant logic of pkg/trust to be moved inside this package … in some unknown way. We would need to worry about designing a stable API that can actually evolve over time — which seems extremely hard (and we did, also, change the semantics of `podman trust show` this year).

dfr:
All of what you say makes sense. I'm pretty sure I caused the divergence back in May when I changed the default in c/image.

Looking at c/common, I don't see any direct uses of DefaultSignaturePolicyPath but this copy of the value was definitely the cause of 'podman commit' failing for me in FreeBSD testing, probably via the DefaultConfig function. This was changed yesterday to sync with the value from c/image but it would be nice not to have the copy at all.

For podman's pkg/trust, I'm not at all clear on what the best way forward is - I don't really understand how this all works.

mtrmac:
pkg/trust is the flip side of the argument: _Because_ c/image doesn’t want to expose a stable API, pkg/trust contains copies some of the code and data structures instead. So I understand the argument for exposing more as a c/image stable API — it’s just that this argument doesn’t make it any easier to actually design.

dfr:
Ok then - since the immediate problem of getting 'podman commit' working is resolved by changing the constant in c/common I will abandon this PR and perhaps open an issue here to remind that there is still a little tech-debt associated with the location of policy.json.

PR #1730, opened 2022-11-21
Update to distribution/distribtion in master branch

rhatdan:
Improvements on https://github.com/containers/podman/issues/16205

### DIFF
--- a/docker/reference/regexp.go
+++ b/docker/reference/regexp.go
@@ -3,141 +3,167 @@ package reference
 import "regexp"
 
 var (
-	// alphaNumericRegexp defines the alpha numeric atom, typically a
+	// alphaNumeric defines the alpha numeric atom, typically a
 	// component of names. This only allows lower case characters and digits.
-	alphaNumericRegexp = match(`[a-z0-9]+`)
+	alphaNumeric = `[a-z0-9]+`
 
-	// separatorRegexp defines the separators allowed to be embedded in name
+	// separator defines the separators allowed to be embedded in name
 	// components. This allow one period, one or two underscore and multiple
-	// dashes.
-	separatorRegexp = match(`(?:[._]|__|[-]*)`)
+	// dashes. Repeated dashes and underscores are intentionally treated
+	// differently. In order to support valid hostnames as name components,
+	// supporting repeated dash was added. Additionally double underscore is
+	// now allowed as a separator to loosen the restriction for previously
+	// supported names.
+	separator = `(?:[._]|__|[-]*)`
 
-	// nameComponentRegexp restricts registry path component names to start
+	// nameComponent restricts registry path component names to start
 	// with at least one letter or number, with following parts able to be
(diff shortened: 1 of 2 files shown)


rhatdan:
@mtrmac @alexlarsson Is this what you had in mind?

docker/reference/normalize.go
```diff
@@ -89,16 +88,21 @@ func ParseDockerRef(ref string) (Named, error) {
 // needs to be already validated before.
 func splitDockerDomain(name string) (domain, remainder string) {
 	i := strings.IndexRune(name, '/')
-	if i == -1 || (!strings.ContainsAny(name[:i], ".:") && name[:i] != "localhost") {
+	if i == -1 || (!strings.ContainsAny(name[:i], ".:") && name[:i] != "localhost" && strings.ToLower(name[:i]) == name[:i]) {
```
mtrmac, line 91: It’s not obvious we want this.

docker/reference/reference.go
```diff
@@ -5,7 +5,9 @@
 //
 //	reference                       := name [ ":" tag ] [ "@" digest ]
 //	name                            := [domain '/'] path-component ['/' path-component]*
-//	domain                          := domain-component ['.' domain-component]* [':' port-number]
+//	domain                          := host [':' port-number]
+//	host                            := domain-name | IPv4address | \[ IPv6address \]	; rfc3986 appendix-A
```
mtrmac, line 9: It’s _not at all_ obvious that we want this. c/image/storage uses a `[storageOptions]name` syntax already, how do we disambiguate?

mtrmac (commented):
Maaaayyybe? It’s not at all obvious we can safely update all of it, and I personally don’t want to deal with this now.

docker/reference/normalize.go
rhatdan: Do you think we have tests to reveal that we want this?  If not I am fine with just removing the change.

docker/reference/reference.go
rhatdan: No idea, I was just force updating, I have no problem removing the host definition from this.

mtrmac: … and then we need to somehow manage the divergence from the upstream package, if one emerges.

I’m fine with that happening in principle, I just can’t prioritize this.

mtrmac: A possible path forward: Update not to the head of the branch, but to some specific commit that contains exactly what we want, but nothing that we would have to think about. That way we would get the improvements we want, we would stay in sync, and we could still bring in the other changes when there’s more capacity.

_If_ such a commit exists.

Alternatively, get this reviewed by someone else who can investigate in enough detail.

mtrmac: That is an effective fork, selectively choosing features… It’s an option, sure, but it turns the cost of making a decision on those other changes into technical debt to be paid later. When will that “later” ever happen?

rhatdan:
Will go with @vrothberg version.

PR #1733, opened 2022-11-28
docker/reference: reduce regex compilations

vrothberg:
Improve the code such that a given raw regex string is compiled only once.

Before:
init 1341880 bytes, 11453 allocs

After:
init 485600 bytes, 4501 allocs

@mtrmac @rhatdan 

I did all changes manually (i.e., not copied over).  While it would be best to avoid the redundancy in the packages, I do not see a viable solution at the moment.  This is a smaller improvement but an improvement nonetheless.

### DIFF
--- a/docker/reference/regexp.go
+++ b/docker/reference/regexp.go
@@ -120,13 +120,7 @@ var re = regexp.MustCompile
 // literal compiles s into a literal regular expression, escaping any regexp
 // reserved characters.
 func literal(s string) string {
-	re := re(regexp.QuoteMeta(s))
-
-	if _, complete := re.LiteralPrefix(); !complete {
-		panic("must be a literal")
-	}
-
-	return re.String()
+	return regexp.QuoteMeta(s)
 }
 
 // expression defines a full expression, where each regular expression must
--- a/docker/reference/regexp.go
+++ b/docker/reference/regexp.go
@@ -2,7 +2,7 @@ package reference
 
 import "regexp"
 
-var (
+const (
 	// alphaNumeric defines the alpha numeric atom, typically a
 	// component of names. This only allows lower case characters and digits.
 	alphaNumeric = `[a-z0-9]+`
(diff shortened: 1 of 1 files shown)


mtrmac (commented):
So we are hard-forking the package? At least for now? Fine, I guess...

Is this equivalent to upstream commit 1c89ce5fc1cec85e0726f598bccc12b873d04de3 ? If so, would it be possible to use it directly, and very clearly document that? From a very quick look at the history, it seems that it might be applicable as is. (Or is the difference between `const` and `var` visible?)

---

_Eventually_, purely hypothetically, I suppose it could be possible to abandon the regexes altogether by introducing `docker/reference2` which does not contain them, and forward all v1 implementations to v2. OTOH, for Podman to benefit, every single user of the`docker/reference` subpackage would need to be migrated.

vrothberg:
> So we are hard-forking the package? At least for now? Fine, I guess...

No, it's not a fork.  I just picked the idea but performed all the changes manually.  The other package has ipv6 stuff in that may cause undesired side-effects.  So instead of cherry-picking the commit and removing unwanted changes, I started with a clean slate.

I investigated ways to "replace" the distribution/reference package but there seems no way without forking - very unattractive and then prone to circular imports on the repo level.

rhatdan:
LGTM

docker/reference/regexp.go
```diff
@@ -1,143 +1,168 @@
...
+	anchoredShortIdentifierRegexp = match(anchored(shortIdentifierStr))
 )
 
 // match compiles the string to a regular expression.
 var match = regexp.MustCompile
 
 // literal compiles s into a literal regular expression, escaping any regexp
 // reserved characters.
-func literal(s string) *regexp.Regexp {
+func literal(s string) string {
```
mtrmac, line 122: AFAICS this could be basically eliminated (by making it `return regexp.QuoteMeta` or just hard-coding the result), avoiding another bunch of compiles.

mtrmac:
> > So we are hard-forking the package? At least for now? Fine, I guess...
> 
> No, it's not a fork. I just picked the idea but performed all the changes manually.

That’s what I meant: there is no longer any direct correspondence between the c/image/docker/reference and distribution/distribution/reference codebases. (You’re right that we are not creating a fork of either of the two whole repos, to be used in a `go.mod` `replace`.)

mtrmac:
> Is this equivalent to upstream commit 1c89ce5fc1cec85e0726f598bccc12b873d04de3 ?

@vrothberg WDYT about this part? We probably can’t get to a 1-1 commit-commit correspondence either way, but maybe a cherry-pick of this + additions on top would make it a bit easier to catch up / merge in the future.

vrothberg:
It sounds like a lot of extra work. I can do it if you prefer but what is
the expected benefit over the current state? I could add a link to the
other commit and add it to the commit message in case you’re worried about
preserving the origin. The latest state of the code is already quite
different than the commit in question, so I didn’t bother about keeping it
close.

mtrmac:
> what is the expected benefit over the current state?

Assuming we will ever want to catch up with the distribution/distribution definition (which is not _certain_ but seems likely), I think it’s useful to be able to tell where we differ, and which parts are intentional deviations vs. us just being behind. So, direct 1-1 correspondence in changes, and minimizing user-irrelevant differences, seems valuable.

> The latest state of the code is already quite different than the commit in question, so I didn’t bother about keeping it
close.

https://github.com/mtrmac/image/tree/regex-improvements doesn’t look so bad: That’s just copying the file from 1c89ce5fc1cec85e0726f598bccc12b873d04de3 , + your changes, resolving merge conflicts in favor of using the distribution/distribution variable naming.

vrothberg:
Updated accordingly. @mtrmac PTanotherL

vrothberg:
@alexlarsson FYI

mtrmac:
@vrothberg do you want to also do https://github.com/containers/image/pull/1733#discussion_r1033921229 , or would you prefer to merge as is?

vrothberg:
> @vrothberg do you want to also do [#1733 (comment)](https://github.com/containers/image/pull/1733#discussion_r1033921229) , or would you prefer to merge as is?

Apologies.  I forgot about that comment but addressed it now.  I added another commit ontop to use strings.Join() to carve out 2 more allocs.  We're now down to 4411.

mtrmac:
Oh… for some reason I thought that regex compilation of that literal would be more expensive, and it would be a bigger win.

Either way, it’s good to have that improvement.

PR #1746, opened 2022-12-08
WIP: Add a `/etc/containers/auth.json`

cgwalters:
A long-running tension in the docker/podman land is around running as a system service versus being executed by a user. (Specifically a "login user", i.e. a Unix user that can be logged
 into via `ssh` etc.)

 For login users, it makes total sense to configure the container
 runtime in `$HOME`.

 But for system services (e.g. code executed by systemd) it
 is generally a bad idea to access or read the `/root` home
 directory.  On image based systems, `/root` may be dynamically
 mutable state in contrast to `/etc` which may be managed
 by OS upgrades, or even be read-only.

 For these reasons, let's introduce `/etc/contaners/auth.json`.
 If it is present, and the current process is executing in
 systemd, it will be preferred.  (There's some further logic
 around this that is explained in the manpage; please see that
 for details)

cc https://github.com/coreos/rpm-ostree/issues/4180

### DIFF
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -32,6 +33,8 @@ type dockerConfigFile struct {
 	CredHelpers map[string]string           `json:"credHelpers,omitempty"`
 }
 
+// systemPath is the global auth path preferred for systemd services.
+var systemPath = authPath{path: filepath.FromSlash("/etc/containers/auth.json"), legacyFormat: false, requireUserOnly: true}
 var (
 	defaultPerUIDPathFormat = filepath.FromSlash("/run/containers/%d/auth.json")
 	xdgConfigHomePath       = filepath.FromSlash("containers/auth.json")
@@ -53,6 +56,8 @@ var (
 type authPath struct {
 	path         string
 	legacyFormat bool
+	// requireUserOnly will cause the file to be ignored if it is readable by group or other
+	requireUserOnly bool
 }
 
 // newAuthPathDefault constructs an authPath in non-legacy format.
@@ -215,7 +220,21 @@ func GetAllCredentials(sys *types.SystemContext) (map[string]types.DockerAuthCon
 // The homeDir parameter should always be homedir.Get(), and is only intended to be overridden
 // by tests.
 func getAuthFilePaths(sys *types.SystemContext, homeDir string) []authPath {
+	runningInSystemd := os.Getenv("INVOCATION_ID") != ""
+	runningAsRoot := os.Getuid() == 0
+	runningSystemdPrivileged := runningInSystemd && runningAsRoot
+
 	paths := []authPath{}
+
+	haveExplicitConfig := sys != nil && (sys.AuthFilePath != "" || sys.LegacyFormatAuthFilePath != "")
+
+	// If we're in systemd, prefer the global auth path first.
+	insertedGlobalPath := false
+	if !haveExplicitConfig && runningSystemdPrivileged {
+		paths = append(paths, systemPath)
+		insertedGlobalPath = true
+	}
+
 	pathToAuth, err := getPathToAuth(sys)
 	if err == nil {
 		paths = append(paths, pathToAuth)
@@ -225,7 +244,7 @@ func getAuthFilePaths(sys *types.SystemContext, homeDir string) []authPath {
 		// Logging the error as a warning instead and moving on to pulling the image
 		logrus.Warnf("%v: Trying to pull image in the event that it is a public image.", err)
 	}
-	if sys == nil || (sys.AuthFilePath == "" && sys.LegacyFormatAuthFilePath == "") {
+	if !haveExplicitConfig {
 		xdgCfgHome := os.Getenv("XDG_CONFIG_HOME")
 		if xdgCfgHome == "" {
 			xdgCfgHome = filepath.Join(homeDir, ".config")
@@ -241,6 +260,12 @@ func getAuthFilePaths(sys *types.SystemContext, homeDir string) []authPath {
 		paths = append(paths,
 			authPath{path: filepath.Join(homeDir, dockerLegacyHomePath), legacyFormat: true},
 		)
+		// If we didn't already insert the global path, do it at the end if we're running as root.
+		// This will ensure the same semantics for code executed as systemd units and run
+		// from an interactive shell (as root) as long as there's no user-root owned configs.
+		if !insertedGlobalPath && runningAsRoot {
+			paths = append(paths, systemPath)
+		}
 	}
 	return paths
 }
@@ -552,13 +577,28 @@ func getPathToAuthWithOS(sys *types.SystemContext, goOS string) (authPath, error
 func (path authPath) parse() (dockerConfigFile, error) {
 	var auths dockerConfigFile
 
-	raw, err := os.ReadFile(path.path)
+	f, err := os.Open(path.path)
 	if err != nil {
 		if os.IsNotExist(err) {
 			auths.AuthConfigs = map[string]dockerAuthConfig{}
 			return auths, nil
 		}
 		return dockerConfigFile{}, err
+	}
+	defer f.Close()
+	if path.requireUserOnly {
+		st, err := f.Stat()
+		if err != nil {
+			return dockerConfigFile{}, fmt.Errorf("stat %s: %w", path.path, err)
+		}
+		perms := st.Mode().Perm()
+		if (perms & 044) > 0 {
+			return dockerConfigFile{}, fmt.Errorf("refusing to process %s with group or world read permissions", path.path)
+		}
+	}
+	raw, err := io.ReadAll(f)
+	if err != nil {
+		return dockerConfigFile{}, fmt.Errorf("reading %s: %w", path.path, err)
 	}
 
 	if path.legacyFormat {
--- a/docs/containers-auth.json.5.md
+++ b/docs/containers-auth.json.5.md
@@ -6,15 +6,20 @@ containers-auth.json - syntax for the registry authentication file
 # DESCRIPTION
 
 A credentials file in JSON format used to authenticate against container image registries.
-The primary (read/write) file is stored at `${XDG_RUNTIME_DIR}/containers/auth.json` on Linux;
+The primary (read/write) per-user file is stored at `${XDG_RUNTIME_DIR}/containers/auth.json` on Linux;
 on Windows and macOS, at `$HOME/.config/containers/auth.json`.
 
-When searching for the credential for a registry, the following files will be read in sequence until the valid credential is found:
+There is also a system-global `/etc/containers/auth.json` path.  When the current process is executing inside systemd as root, this path will be preferred.
+
+When running as a user and searching for the credential for a registry, the following files will be read in sequence until the valid credential is found:
 first reading the primary (read/write) file, or the explicit override using an option of the calling application.
 If credentials are not present, search in `${XDG_CONFIG_HOME}/containers/auth.json` (usually `~/.config/containers/auth.json`), `$HOME/.docker/config.json`, `$HOME/.dockercfg`.
 
+If the current process is not running in systemd, but is running as root, the system global path will be read last.
+
 Except the primary (read/write) file, other files are read-only, unless the user use an option of the calling application explicitly points at it as an override.
 
+Note that the `/etc/containers/auth.json` file must not be readable by group or world (i.e. mode `044`), or a fatal error will occur.
 
 ## FORMAT
 


cgwalters:
Specifically for example I plan to ensure that bootc to reads this file and *not* `/root/.docker` etc by default.  For podman to ensure maximum compatibility I think we need to *add* this file into the set.

cgwalters:
(Lightly tested with an override to point skopeo at this)

cgwalters:
For OpenShift I think it would make sense to have `/var/lib/kubelet/config.json` be a symlink to `/etc/containers/auth.json`.  This would fix a pile of [long-running hacks that hardcode `podman run --authfile=/var/lib/kubelet/config.json`](https://github.com/openshift/machine-config-operator/blob/357b1a792721ec86ebbacb0e19b1d0f292399b45/templates/common/_base/units/machine-config-daemon-pull.service.yaml#L20).

pkg/docker/config/config.go
```diff
@@ -204,7 +208,17 @@ func GetAllCredentials(sys *types.SystemContext) (map[string]types.DockerAuthCon
...
 // by tests.
 func getAuthFilePaths(sys *types.SystemContext, homeDir string) []authPath {
+	runningInSystemd := os.Getenv("INVOCATION_ID") != ""
+	runningAsRoot := os.Getuid() == 0
+	runningSystemdPrivileged := runningInSystemd && runningAsRoot
+
 	paths := []authPath{}
+
+	// If we're in systemd, prefer the global auth path first.
+	if runningSystemdPrivileged {
```
mtrmac, line 218: The explicit caller’s configuration from `types.SystemContext` should be preferred over any built-in defaults.

pkg/docker/config/config.go
```diff
@@ -204,7 +208,17 @@ func GetAllCredentials(sys *types.SystemContext) (map[string]types.DockerAuthCon
 // The homeDir parameter should always be homedir.Get(), and is only intended to be overridden
 // by tests.
 func getAuthFilePaths(sys *types.SystemContext, homeDir string) []authPath {
```
mtrmac, line 210: Adding the new paths to this function means that writers would never write to `systemPath` without a specific override; is that intended?

I think that’s actually a good idea, so that `root`’s interactive login doesn’t write to a noninteractively-used file, but it should be documented.

pkg/docker/config/config.go
```diff
@@ -35,9 +36,12 @@ type dockerConfigFile struct {
 type authPath struct {
 	path         string
 	legacyFormat bool
+	// requireUserOnly will cause the file to be ignored if it is readable by group or other
+	requireUserOnly bool
```
mtrmac, line 40: This is never set AFAICS

mtrmac (commented):
Thanks! Using a PR to discuss the details is definitely helpful. Just a quick first impression for now.

---

> For OpenShift I think it would make sense to have `/var/lib/kubelet/config.json` be a symlink to `/etc/containers/auth.json`.

Yes, if there is any kind of computer where having a system-wide credentials file makes sense, an OpenShift node should be that. OTOH does this mean we are giving up on some SELinux restrictions on `root`? The `/etc/containers` path is necessarily unrestricted. I didn’t look into how the Kubelet is, or isn’t restricted.

Maybe do it the other way, and symlink `/etc/containers/auth.json` to the Kubelet path.

cgwalters:
> OTOH does this mean we are giving up on some SELinux restrictions on root? The /etc/containers path is necessarily unrestricted. I didn’t look into how the Kubelet is, or isn’t restricted.

Good question.  Today,

```
$ ll -Z /etc/containers/
total 12
drwxr-xr-x. 1 root root system_u:object_r:etc_t:s0        0 Oct  6 15:46 certs.d
drwx------. 1 root root unconfined_u:object_r:etc_t:s0   26 Oct 30 17:55 networks
drwxr-xr-x. 1 root root system_u:object_r:etc_t:s0       14 Oct  6 15:46 oci
-rw-r--r--. 1 root root system_u:object_r:etc_t:s0      569 Oct  6 15:46 policy.json
-rw-r--r--. 1 root root system_u:object_r:etc_t:s0     3897 Oct  6 15:46 registries.conf
drwxr-xr-x. 1 root root system_u:object_r:etc_t:s0       38 Oct  6 15:46 registries.conf.d
drwxr-xr-x. 1 root root system_u:object_r:etc_t:s0      132 Nov 17 19:53 registries.d
-rw-r--r--. 1 root root system_u:object_r:etc_t:s0      699 Dec  9  2021 toolbox.conf
```

It could make sense to introduce a dedicated `container_auth_t` or so.  But we can do that even inside `/etc/containers`.  Or, we could have it be in `/etc/containers/secrets` and label the directory for future expansion - and if we do that it'd make sense to have the dir *also* mode `0700` for example for stock DAC protections.

cgwalters:
A variant of this is that we could:

- Just add a change to the man page here which documents the possible existence of `/etc/containers/auth.json`
- I change bootc to use it by default

This I think would get about 45% of the value with 1% of the work.

mtrmac:
I’m not thinking so much a matter of whether we could isolate this in `root` with SELinux in _principle_, it’s more that we can’t do that _effectively_: once `/etc/containers/…auth.json` becomes the Podman default for `root`, all plausible SELinux contexts that could run Podman must be allowed to access that file. The only way to truly isolate is to have more isolated files, i.e. not the system-wide `/etc/containers/*`.

But that’s just me rehashing the previous discussion, I guess. It seems to me that single-purpose appliances can use `/etc/containers`, traditional multi-user multi-purpose UNIX servers shouldn’t set it up at all, and should continue to use whatever they are doing now.

mtrmac:
> A variant of this is that we could: …

Overall, so far I’m not convinced that adding this default path is _clearly worthwhile_. But it’s useful enough in some contexts that I’m fine with it being added.

cgwalters:
> I’m not thinking so much whether we could isolate this in root with SELinux in principle, it’s more that we can’t do that effectively: once /etc/containers/…auth.json becomes the Podman default for root, all plausible SELinux contexts that could run Podman must be allowed to access that file.

I think it is possible because today most invocations of podman are going to start from `unconfined_t` or `init_t`, and we'd be sure that those contexts are allowed to access the file.  The value of the distinct context is that lesser-privileged domains (say `NetworkManager_t`) would not be able to read it by default.

> But that’s just me rehashing the previous discussion, I guess. It seems to me that single-purpose appliances can use /etc/containers, traditional multi-user multi-purpose UNIX servers shouldn’t set it up at all, and should continue to use whatever they are doing now.

Well, we have to account for mixed setups.  For example, a valid user story here is a company/organization deploying Linux desktop systems using `bootc` and having that system upgrade use one pull secret, which would be distinct from the secret they use (today) in `$HOME` or `/root` for their application containers.

So I see this as just adding a dedicated file for "the system" without conflicting with any optional setup.  

> Overall, so far I’m not convinced that adding this default path is clearly worthwhile. But it’s useful enough in some contexts that I’m fine with it being added.

Thanks; this is a useful debate to have for sure because once added, it can't be removed and so there's long term maintenance implications for sure.

mtrmac:
> I think it is possible because today most invocations of podman are going to start from `unconfined_t` or `init_t`, and we'd be sure that those contexts are allowed to access the file. The value of the distinct context is that lesser-privileged domains (say `NetworkManager_t`) would not be able to read it by default.

(BTW OpenShift, at least in some versions, runs Podman from inside a NetworkManager hook… it’s containers all the way down.)

cgwalters:
> At this point it seems useful to make readJSONFile, findCredentialsInFile (and other functions?) methods on authPath, instead of passing the individual parameters; ideally in a separate commit.

Done in https://github.com/containers/image/pull/1765
(unit tests don't quite pass for me locally, but they fail in the same way without this.  They seem to assume some global state)

mtrmac:
> 
(unit tests don't quite pass for me locally, but they fail in the same way without this. They seem to assume some global state)

I wouldn’t be too surprised if they were insufficiently isolated from the per-user state. If so, that’s not intentional.

pkg/docker/config/config.go
cgwalters: It looks like you partially addressed this with https://github.com/containers/image/pull/1751 - I built on that, though this is getting messier.

pkg/docker/config/config.go
```diff
@@ -538,17 +560,32 @@ func getPathToAuthWithOS(sys *types.SystemContext, goOS string) (string, bool, e
...
 	}
+	defer f.Close()
+	if requireUserOnly {
+		st, err := f.Stat()
+		if err != nil {
+			return dockerConfigFile{}, fmt.Errorf("stat %s: %w", path, err)
+		}
+		perms := st.Mode().Perm()
+		if (perms & 04) > 0 {
+			return dockerConfigFile{}, fmt.Errorf("refusing to process %s with world read permissions", path)
```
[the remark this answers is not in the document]
cgwalters: For now I chose to error on group/world read permissions.  Forcing 0600 also forces write permissions, but I can imagine cases where one might not want that.  (Not that it really matters because root usually has `CAP_DAC_OVERRIDE`)

cgwalters:
> (BTW OpenShift, at least in some versions, runs Podman from inside a NetworkManager hook… it’s containers all the way down.)

Yeah true :cry: 

But an attempt to default SELinux confine NetworkManager hooks IIRC failed, and I think the status quo there is the default is unconfined.

cgwalters:
OK I rebased :surfer: this - it's definitely cleaner on top of https://github.com/containers/image/pull/1765

Fixed some comments and issues, but still more to do.

pkg/docker/config/config.go
```diff
@@ -552,14 +577,29 @@ func getPathToAuthWithOS(sys *types.SystemContext, goOS string) (authPath, error
...
 		return dockerConfigFile{}, err
 	}
+	defer f.Close()
+	if path.requireUserOnly {
+		st, err := f.Stat()
+		if err != nil {
+			return dockerConfigFile{}, fmt.Errorf("stat %s: %w", path.path, err)
+		}
+		perms := st.Mode().Perm()
+		if (perms & 044) > 0 {
```
mtrmac, line 595: `!= 0` seems a better fit with the “bit set” format.

pkg/docker/config/config.go
mtrmac: Looking back at #1751, that duplication really wasn’t a good idea. #1772 should help.

cgwalters:
We might also want to investigate https://systemd.io/CREDENTIALS/ here.

cgwalters:
I recently came across https://access.redhat.com/solutions/7002142 which has a really bad suggestion for how to make `podman login` persist.  I do think this PR still makes sense; will try to rebase at some point soon and pick this back up.

vrothberg:
@cgwalters, what are your thoughts on this issue today?

I am carefully walking over the bootc docs and think that the [status quo](http://localhost:8080/bootc/container-pull-secrets/#_unifying_bootc_and_podman_credentials) does not offer a good UX.  Having an `/etc/containers/auth.json` seems like an intuitive way forward but maybe some conversations have changed direction in the meantime.

I didn't read the code yet but would expect that `/etc/containers/auth.json` is only used for reading credentials?

rhatdan:
What would the protections on that file be by default 600 would be my suggestion, and then people who wanted to share it with non root users could open it up, since it has a non encrypted password embedded in it.

cgwalters:
> @cgwalters, what are your thoughts on this issue today?

Oh yes I think we should definitely still do this.

mtrmac:
> I am carefully walking over the bootc docs and think that the status quo does not offer a good UX.

@vrothberg That link isn’t publicly available

mtrmac:
My thoughts in general:

- I don’t really see value in a systemd-service-specific feature, the way this PR works now. A systemd service can be reasonably (and with things like Quadlet, quite easily and in an automated way!) set up to have its own service-specific secret file, and to pass `--authfile` to _only_ use those credentials instead of whatever else is in the environment. Making it the default workflow to use an explicit `--authfile` is attractive to me.
- I don’t feel strongly about a system-wide `/etc/containers/auth.json` read by _everything_. There are systems where “everything has access to a credential” can be valuable, and certainly easier than manually managing credentials in all relevant contexts. OTOH it needs to be constrained to such specific single-purpose systems, it should never be the _default_ way flexible systems are built. (In particular, ISV’s software installation instructions for generally-applicable software should _never_ tell users to put credentials into a system-readable file.) So… this would make work a tiny bit easier for users who are already doing a _lot_ of work to build a single-purpose appliance anyway.
- We’ve had a conversation about RHEL subscription management automatically setting up registry credentials. Now for that, we _do_ need a system-wide config file — but it also probably should not be a _single_ file, but a `*.d` directory, so that it can be managed per-repo (per product subscription) without conflicts. Or, alternatively, we would have to figure out local locking (#1365 ), and ensure that _everything_ follows the rules — supporting `*.d` seems easier.

cgwalters:
I'd agree that in the general case of per-image/per-service pull secrets makes total sense. 

But I do feel for people trying out container-based systems at a smaller scale (especially the important proof-of-concept phase), it's highly likely they start with just one authenticated registry and they'd be perfectly fine with a global secret to start (especially one that applied to bootc and podman), with the ability to override for individual images being the next step.

> but it also probably should not be a single file, but a *.d directory, so that it can be managed per-repo (per product subscription) without conflicts. 

This would definitely be nice indeed. For clarity and completeness, this would hopefully really be having an auth file stack that implemented the https://uapi-group.org/specifications/specs/configuration_files_specification/

mtrmac:
> For clarity and completeness, this would hopefully really be having an auth file stack that implemented the https://uapi-group.org/specifications/specs/configuration_files_specification/

As long as they are at the end of the search list = do not incur a cost for applications with more explicit credential locations, why not.

cgwalters:
I randomly came across https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/8/html-single/composing_a_customized_rhel_system_image/index#con_the-container-registry-credentials_assembly_pushing-a-container-to-a-register-and-embedding-it-into-a-image today which created another path for pull secrets - if I had to guess at the rationale it'd be similar.

mtrmac:
That looks like a single application having a place to configure credentials used for that application (and, of course, it’s fine for such things to exist); if that expresses a need for need for a system-wide credential store, I don’t know enough to see it.

cgwalters:
I do continue to think that many simple use cases would be happy with a default global path and
if we had a global .d directory, we would have at least referenced it from that doc as an option versus an app-specific path.

cgwalters:
Closing in favor of https://github.com/containers/image/pull/2600 where development continues

Issue #161, opened 2016-11-10
Docker TLS configuration outstanding question


mtrmac:
Outstanding comments from #160:

* Why are we using a `ServerDefault` `tls.Config` in a client?
* Why are we applying it only to a single path, not when we override `tls.Config` for our purposes?
* In `newTransport`, the `proxyDialer` setup should happen before we create a `http.Transport`, so that we don’t need to override `tr.Dial` in a created object.

rhatdan:
@mtrmac @vrothberg What is the state of this issue. It is two years old, can we close it?

vrothberg:
I'll leave that up to @mtrmac.

mtrmac:
Updated the description — the second item was cleaned up.

The third one is a trivial cleanup; the first one would require actual research, it would be nice for it to eventually happen.

mtrmac:
The third item will go away, so I’m now re-titling this to track making an opinion on TLS and HTTP
- Do we want to deviate from the standard library’s defaults at all?
- If so, how and why? And document that in the code.

PR #1748, opened 2022-12-08
Remove go-conections uses in tlsclientconfig

mtrmac:
A small cleanup that should not affect behavior; see individual commit messages for details.

Motivated by https://github.com/containers/image/issues/161 .

Note that this does not change anything about the TLS configuration, or express any opinions about the HTTP configuration. Also, the dependency is still used indirectly due to other sub packages.


vrothberg (approved):
LGTM

rhatdan:
LGTM

PR #1749, opened 2022-12-08
Use io.SeekStart instead of a hard-coded 0

mtrmac:
The hard-coded 0 is defined in the API, so this works fine; this is just a readability improvement.

Should not change behavior.


vrothberg (approved):
LGTM

PR #1751, opened 2022-12-08
Don't use any default path fallbacks if the user specified a path

### LINKED ISSUE #596: Incorrect/inconsistent docker/config.json path lookup

mtrmac:
This means that users setting up AuthFilePath can rely on that config file, without influences from the external environment.

OTOH it might break existing users.

Fixes #596 — but I’m not really sure we want to make this change after all these years.

### DIFF
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -217,23 +217,25 @@ func getAuthFilePaths(sys *types.SystemContext, homeDir string) []authPath {
 		// Logging the error as a warning instead and moving on to pulling the image
 		logrus.Warnf("%v: Trying to pull image in the event that it is a public image.", err)
 	}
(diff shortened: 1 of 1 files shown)


vrothberg (commented):
SGTM

Hard to predict whether it'll break something.  Yet it seems more intuitive to only use what the user has provided.

rhatdan:
I see the current settings as a bug.  If a calling app specifies the path, then it should use it, just because the tool finds it in a random location, can cause random failures that could be difficult to debug.

mtrmac:
All right, let’s do this.

Issue #755, opened 2019-11-24
Re-design BlobInfo *Operation


lumjjb:
>“edits to blob” are not “properties of a blob”, and should be separated / only maintained for ManifestUpdateOptions.

During discussions of https://github.com/containers/image/pull/673, one of the discussion points was that there should be a re-design  of `BlobInfo` to not use the recently introduced `CompressionOperation` field. 

https://github.com/containers/image/blob/039b763daf9582d170d1b9eba1d7bbc4f574332f/types/types.go#L107-L118

mtrmac:
#1752 will at least document the intent; actual redesign will probably happen only when a new need to edit the fields comes up.

mtrmac:
We can’t change the semantics of a public API any more… but it seems possible that the layer edits will need a new API, and that one will probably start out internal-only.

mtrmac:
We now have the infrastructure to do image operations in unexported API, e.g. `internal/image.Sourced.CanChangeLayerCompression` is not publicly reachable. Similarly there is `private.ImageSource` and `private.ImageDestination` which might migrate from public `BlobInfo` to a smaller object.

But we are very unlikely to refactor existing users of `BlobInfo` just as a cleanup, without another reason to touch the code; so, there’s not much benefit to tracking it as an issue.

`types.BlobInfo` contains fairly loud comments warning against adding more “edit” fields, hopefully that’s enough.

PR #1752, opened 2022-12-09
Add comments to BlobInfo to warn against adding more edit fields

mtrmac:
This is not #755, but it should help prevent the problem from spreading.

The _next_ user who needs to add an edit operation should be forced to provide a new internal edit API, and then we can add the existing fields to that API and deprecate the old fields.


vrothberg (approved):
LGTM

rhatdan:
LGTM

PR #1753, opened 2022-12-09
Reject files mixing v1 and v2 registries.conf, even with empty fields

mtrmac:
We already rely on the TOML parser's behavior of setting specified arrays to `[]`, but missing arrays to `nil`; so just use it to be more strict.

This can help situations like

```toml
[registries.search]
registries = []

[registries.insecure]
registries = []

[registries.block]
registries = []

unqualified-search-registries = ["docker.io"]

[[registry]]
...
```

although only indirectly: From the TOML parser's point of view, the `unqualified-search-registries` field is `block.unqualified-search-registries` and therefore invisible to us. Conceptually, the same thing could happen with
```toml
[arbitrary-unknown]
unqualified-search-registries = ["docker.io"]


[[registry]]
...
```

and we wouldn't warn about that.

Even so, the config file author specifying even empty fields is suspect and suggests a problem to be resolved.

---

Note that this could possibly break current users that generate slightly-nonsensical files.

### DIFF
--- a/pkg/sysregistriesv2/system_registries_v2.go
+++ b/pkg/sysregistriesv2/system_registries_v2.go
@@ -198,6 +198,7 @@ type V1RegistriesConf struct {
 }
 
 // Nonempty returns true if config contains at least one configuration entry.
+// Empty arrays are treated as missing entries.
 func (config *V1RegistriesConf) Nonempty() bool {
 	copy := *config // A shallow copy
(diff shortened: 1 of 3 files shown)


mtrmac:
Fixes #1054 .

vrothberg (approved):
LGTM

rhatdan:
LGTM

PR #1756, opened 2022-12-09
Fix comments about exponential backoff with Retry-After

### LINKED ISSUE #1565: Misleading comment in dockerClient.makeRequestToResolvedURL
The comment in https://github.com/containers/image/blob/dabbab219651a7abfdb37308fbf5dd590c762b7b/docker/docker_client.go#L530  is somewhat misleading: we primarily use the delay value from a `Retry-After` header, and only use an exponential backoff if that header is missing or we can’t parse it.

mtrmac:
Clarify that exponential backoff only happens if the registry does not specify otherwise.

And in `parseRetryAfter`, do not talk about backoff at all, because that function is not involved in that.

Fixes #1565 .

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -522,9 +522,8 @@ func parseRetryAfter(res *http.Response, fallbackDelay time.Duration) time.Durat
 		return time.Duration(num) * time.Second
 	}
 	// Second, check if we have an HTTP date.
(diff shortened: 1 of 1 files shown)


vrothberg (approved):
LGTM

rhatdan:
LGTM

PR #1765, opened 2022-12-15
config: use `authPath` struct consistently

cgwalters:
config: use `authPath` struct consistently

Instead of packing and unpacking it into two variables in
various places.  Prep for adding a third member of this struct
as part of https://github.com/containers/image/pull/1746

---

config: Make parsing function a method on authFile

Just a general cleanup.

---

### DIFF
--- a/pkg/docker/config/config.go
+++ b/pkg/docker/config/config.go
@@ -157,8 +157,8 @@ func GetAllCredentials(sys *types.SystemContext) (map[string]types.DockerAuthCon
 		// Special-case the built-in helper for auth files.
 		case sysregistriesv2.AuthenticationFileHelper:
 			for _, path := range getAuthFilePaths(sys, homedir.Get()) {
-				// readJSONFile returns an empty map in case the path doesn't exist.
-				auths, err := readJSONFile(path)
+				// parse returns an empty map in case the path doesn't exist.
+				auths, err := path.parse()
 				if err != nil {
 					return nil, fmt.Errorf("reading JSON file %q: %w", path.path, err)
 				}
@@ -546,10 +546,10 @@ func getPathToAuthWithOS(sys *types.SystemContext, goOS string) (authPath, error
 	return newAuthPathDefault(fmt.Sprintf(defaultPerUIDPathFormat, os.Getuid())), nil
 }
 
-// readJSONFile unmarshals the authentications stored in the auth.json file and returns it
+// parse unmarshals the authentications stored in the auth.json file and returns it
 // or returns an empty dockerConfigFile data structure if auth.json does not exist
-// if the file exists and is empty, readJSONFile returns an error
-func readJSONFile(path authPath) (dockerConfigFile, error) {
+// if the file exists and is empty, this function returns an error.
+func (path authPath) parse() (dockerConfigFile, error) {
 	var auths dockerConfigFile
 
 	raw, err := os.ReadFile(path.path)
@@ -602,7 +602,7 @@ func modifyJSON(sys *types.SystemContext, editor func(auths *dockerConfigFile) (
 		return "", err
 	}
 
-	auths, err := readJSONFile(path)
+	auths, err := path.parse()
 	if err != nil {
 		return "", fmt.Errorf("reading JSON file %q: %w", path.path, err)
 	}
@@ -678,7 +678,7 @@ func deleteAuthFromCredHelper(credHelper, registry string) error {
 // findCredentialsInFile looks for credentials matching "key"
 // (which is "registry" or a namespace in "registry") in "path".
 func findCredentialsInFile(key, registry string, path authPath) (types.DockerAuthConfig, error) {
-	auths, err := readJSONFile(path)
+	auths, err := path.parse()
 	if err != nil {
 		return types.DockerAuthConfig{}, fmt.Errorf("reading JSON file %q: %w", path.path, err)
 	}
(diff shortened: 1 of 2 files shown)


pkg/docker/config/config.go
```diff
@@ -502,26 +505,26 @@ func listAuthsFromCredHelper(credHelper string) (map[string]string, error) {
...
-func getPathToAuthWithOS(sys *types.SystemContext, goOS string) (string, bool, error) {
+func getPathToAuthWithOS(sys *types.SystemContext, goOS string) (authPath, error) {
 	if sys != nil {
 		if sys.AuthFilePath != "" {
-			return sys.AuthFilePath, false, nil
+			return newAuthPathDefault(sys.AuthFilePath), nil
 		}
 		if sys.LegacyFormatAuthFilePath != "" {
-			return sys.LegacyFormatAuthFilePath, true, nil
+			return authPath{path: sys.LegacyFormatAuthFilePath, legacyFormat: true}, nil
```
cgwalters, line 520: This is the only place that actually set `legacyFormat: true` that I saw, but please double check.  Everything else should be using `newAuthPathDefault`.

mtrmac: There’s also `dockerLegacyHomePath`.

cgwalters: OK right, which is also correctly not changed.

pkg/docker/config/config.go
```diff
@@ -222,14 +226,13 @@ func getAuthFilePaths(sys *types.SystemContext, homeDir string) []authPath {
...
 			xdgCfgHome = filepath.Join(homeDir, ".config")
 		}
-		paths = append(paths, authPath{path: filepath.Join(xdgCfgHome, xdgConfigHomePath), legacyFormat: false})
+		paths = append(paths, newAuthPathDefault(filepath.Join(xdgCfgHome, xdgConfigHomePath)))
 		if dockerConfig := os.Getenv("DOCKER_CONFIG"); dockerConfig != "" {
-			paths = append(paths,
-				authPath{path: filepath.Join(dockerConfig, "config.json"), legacyFormat: false},
-			)
+			path := newAuthPathDefault(filepath.Join(dockerConfig, "config.json"))
+			paths = append(paths, path)
```
mtrmac, line 232: Inline the `path` value into its single user? All the surrounding code inlines `append(paths, newAuthPathDefault(…))` as well, this one being different invites worrying what’s special about it.

pkg/docker/config/config.go
```diff
@@ -533,20 +536,20 @@ func getPathToAuthWithOS(sys *types.SystemContext, goOS string) (string, bool, e
 			// This means the user set the XDG_RUNTIME_DIR variable and either forgot to create the directory
 			// or made a typo while setting the environment variable,
 			// so return an error referring to $XDG_RUNTIME_DIR instead of xdgRuntimeDirPath inside.
-			return "", false, fmt.Errorf("%q directory set by $XDG_RUNTIME_DIR does not exist. Either create the directory or unset $XDG_RUNTIME_DIR.: %w", runtimeDir, err)
+			return newAuthPathDefault(""), fmt.Errorf("%q directory set by $XDG_RUNTIME_DIR does not exist. Either create the directory or unset $XDG_RUNTIME_DIR.: %w", runtimeDir, err)
```
mtrmac, line 539:
```suggestion
			return authPath{}, fmt.Errorf("%q directory set by $XDG_RUNTIME_DIR does not exist. Either create the directory or unset $XDG_RUNTIME_DIR.: %w", runtimeDir, err)
```
to make it a bit more similar that this is “nothing” vs. similar to a valid value.

(/me is longing for a better-designed language again…)

pkg/docker/config/config.go
```diff
@@ -596,7 +600,7 @@ func modifyJSON(sys *types.SystemContext, editor func(auths *dockerConfigFile) (
 		return "", err
 	}
 
-	auths, err := readJSONFile(path, false)
+	auths, err := readJSONFile(pathV)
```
mtrmac, line 603: This invites asking what `pathV` stands for…

`path, err := getPathToAuth`, and using `path.path` in the 3 users of that value, seems easier than answering that question.

pkg/docker/config/config_test.go
```diff
@@ -58,13 +58,13 @@ func TestGetPathToAuth(t *testing.T) {
...
-			res, lf, err := getPathToAuthWithOS(c.sys, c.os)
+			res, err := getPathToAuthWithOS(c.sys, c.os)
 			if c.expected == "" {
 				assert.Error(t, err)
 			} else {
 				require.NoError(t, err)
-				assert.Equal(t, c.expected, res)
-				assert.Equal(t, c.legacyFormat, lf)
+				assert.Equal(t, c.expected, res.path)
+				assert.Equal(t, c.legacyFormat, res.legacyFormat)
```
mtrmac, line 67: Absolutely non-blocking:
```go
assert.Equal(t, authPath{path: c.expected, legacyFormat: c.legacyFormat}, res)
```
would work, and going further `c.expected` could be a whole `authPath` value instead of individual fields.

This works just fine as is.

pkg/docker/config/config.go
```diff
@@ -52,6 +52,10 @@ var (
 	ErrNotSupported = errors.New("not supported")
 )
 
+func newAuthPathDefault(path string) authPath {
```
mtrmac, line 55: Non-blocking: Moving the `authPath` type just above this constructor would be a tiny bit easier to follow.

mtrmac (commented):
Thanks!

LGTM, just a few possible cleanups to ease future maintenance.

TomSweeneyRedHat (commented):
LGTM
once @mtrmac comments are addressed.

cgwalters:
Thanks, addressed those comments, and also added a further cleanup commit.

PR #1769, opened 2022-12-30
storage: Immediately unlink tmpfile

cgwalters:
All we need in this case is an anonymous temporary file, so unlink the file immediately after creating it.  This avoids leaking space if the caller forgets to call `Close()` (and the process exits).

This would have mitigated https://github.com/containers/skopeo/pull/1837 which was a missing `Close()` in the skopeo proxy code.

(It'd be better to use Linux `O_TMPFILE`, but this path is
 portable and avoids a new dependency)

### DIFF
--- a/storage/storage_src.go
+++ b/storage/storage_src.go
@@ -23,7 +23,6 @@ import (
 	"github.com/containers/image/v5/types"
 	"github.com/containers/storage"
 	"github.com/containers/storage/pkg/archive"
-	"github.com/containers/storage/pkg/ioutils"
 	digest "github.com/opencontainers/go-digest"
 	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/sirupsen/logrus"
@@ -129,21 +128,27 @@ func (s *storageImageSource) GetBlob(ctx context.Context, info types.BlobInfo, c
 	if err != nil {
 		return nil, 0, err
 	}
+	success := false
+	defer func() {
+		if !success {
+			tmpFile.Close()
+		}
+	}()
+	// On Unix and modern Windows (2022 at least) we can eagerly unlink the file to ensure it's automatically
+	// cleaned up on process termination (or if the caller forgets to invoke Close())
+	if err := os.Remove(tmpFile.Name()); err != nil {
+		return nil, 0, err
+	}
 
 	if _, err := io.Copy(tmpFile, rc); err != nil {
 		return nil, 0, err
 	}
-
 	if _, err := tmpFile.Seek(0, io.SeekStart); err != nil {
 		return nil, 0, err
 	}
 
-	wrapper := ioutils.NewReadCloserWrapper(tmpFile, func() error {
-		defer os.Remove(tmpFile.Name())
-		return tmpFile.Close()
-	})
-
-	return wrapper, n, err
+	success = true
+	return tmpFile, n, nil
 }
 
 // getBlobAndLayer reads the data blob or filesystem layer which matches the digest and size, if given.


rhatdan:
LGTM
@vrothberg @mtrmac PTAL

mtrmac:
We do have the general ambition to make the code work on Windows, where this AFAICS isn’t possible.

(I don’t think c/storage itself can work on Windows right now, admittedly…)

cgwalters:
Hmm, OK.  Added an (untested) Windows code path.  

That said...instead of this tempfile path, it seems like it'd be better to have the copy operation do this spooling/buffering.  I assume this is to handle slow remote registries, but I think it's a pessimization for local callers that could just parse the stream directly.

vrothberg (commented):
LGTM
@mtrmac PTAL

mtrmac:
(Re: using a temporary file)
> ```go
> // The intention is to keep the time we own the storage lock
> // as short as possible to allow other processes to access the storage.”
> ````

I don’t have a strong opinion on whether this is important (it might be a _bit_ less important soon in the future where this could possibly be a truly read-only shared lock on the layer store, without any exclusive locking intervals) — but it’s not a general feature of the copy process: registry-to-registry copies don’t need a temporary file, they are truly streamed. So this needs to be somehow specific to the storage transport.

That doesn’t mean that it needs to be done directly in the transport, it could, in principle, be a boolean feature flag that drives the general copy code. I don’t know what that would mean for non-copy callers that use `ImageSource` directly — I can imagine an argument that stopping the temporary buffering would be a caller-visible API breakage, but it’s a bit tenuous.

But then again, the quoted argument about reducing locking applies equally to copies and for direct `ImageSource` callers; it’s just not obvious to me that `ImageSource` callers should get a different behavior.

Ultimately, we can always add another option to `SystemContext`… but this feels a bit in the weeds, not something most callers should be tweaking. If anything, it would probably best fit as a per-store property (optimize single-user latency vs. multi-access concurrency).

storage/storage_src.go
```diff
@@ -129,7 +130,13 @@ func (s *storageImageSource) GetBlob(ctx context.Context, info types.BlobInfo, c
...
 	}
-
+	// On Unix we can eagerly unlink the file to ensure it's automatically
+	// cleaned up on process termination (or if the caller forgets to invoke Close())
+	if runtime.GOOS != "windows" {
+		if err := os.Remove(tmpFile.Name()); err != nil {
+			return nil, 0, err
+		}
+	}
 	if _, err := io.Copy(tmpFile, rc); err != nil {
```
mtrmac, line 140: Looking at the code… these error paths are buggy, they should be deleting the file.

Could you fix that as well, please? Something very vaguely like
```go
tmpFile := …
readCloser := tmpFile
if !windows { os.Remove(…) }
else {
    readCloser = NewReadCloserWrapper(…)
}
defer func() {
    if !succeeded { readCloser.Close() }
}
```

would fix that, _and_ move the “ensure the file is eventually removed” logic into one place instead of two as it is now.

(We could even extract that “create a temporary file and ensure it is removed by `Close` at the latest” into a separate utility subpackage, but that can easily wait until we have more users.)

cgwalters: Done.

This code is IMO a perfect example of the "superficial simplicity" of Go materializing as bugs.  In Rust, ownership semantics (instead of potentially arbitrarily delayed GC) would ensure the file is closed by default (in `Drop`) reliably.  Second, conditional compilation is *much* nicer with things like `if cfg!()` - splitting the code into separate files is a big hammer here, and so here I just chose to dynamically dispatch at runtime on the platform, which one finds even in the Go standard library.

This is very familiar to me because I did https://github.com/bytecodealliance/cap-std/commit/a521444939b35b7ef30bdc445b7381fc7ee58a95

(Hmm though now that I think about it, I guess `new_anonymous()` likely doesn't work on Windows...)

cgwalters:
OK in https://github.com/bytecodealliance/cap-std/pull/287 we discovered that newer Windows actually *does* support unlinking files this way, so I would lean a bit towards changing the code to do so unconditionally; by the time we care about Windows support for containers-storage, I'd assume we would require a relatively modern Windows.  But, I dunno.

storage/storage_src.go
mtrmac: > "superficial simplicity" of Go

Oh yes.

mtrmac:
> we discovered that newer Windows actually does support unlinking files this way

Oh, that’s a great find, I didn’t know. I can’t find an official reference but some comments suggest that this has worked as of Windows 10 1909 .

I agree, assuming modern Windows and deleting immediately seems better than that extra complexity.

cgwalters:
> I agree, assuming modern Windows and deleting immediately seems better than that extra complexity.

Done

mtrmac (approved):
Thanks!

cgwalters:
Hm, is there a reason "Require branches to be up to date before merging" is enabled on this repo?  It causes IMO unnecessary rebase churn.

mtrmac:
The repos is small enough, and we typically don’t have that much activity for rebases to be a _frequent_ hassle, so it seems worth the extra confidence about avoiding semantic merge conflicts.

That trade-off would change with a larger project; I don’t think we are not at that scale yet.

Trivial cases can be force-merged.

PR #1776, opened 2023-01-05
RFC: Introduce opaque signer.Signer objects

mtrmac:
This adds a new, very opaque, `signer.Signer` type, allowing users to construct it for the existing simple signing and sigstore implementations, and to pass them to `copy.Image` in `copy.Options.Signers`.

The goal is to:
- Front-load initialization of the signer
  - To allow interactive initialization (passphrase prompts, web browser interactions) immediately, not waiting for layer copies
  - To possibly initialize a signer once before a sequence of multiple `copy.Image` calls (like in `skopeo sync`)
  - To report some failures (incorrect paths) immediately
- Remove the dependency of `copy.Image` on individual signing implementations:
  - Notably just including `copy.Image` won’t necessarily drag in all of the Rekor / Fulcio RPC infrastructure for specialized callers that don’t need it.
  - We can add new options to constructors of the individual `Signer` implementations without having to mirror every single option in a new `copy.Options` field
- Not commit us to any details of the signing API, at least for now.

This is missing tests, and hasn’t actually been used for Fulcio / Rekor yet (compare _extremely_ premature https://github.com/mtrmac/image/tree/fulcio-signing ); filing it early to discuss the API and general approach.

@vrothberg PTAL


copy/sign.go
```diff
@@ -60,12 +51,24 @@ func (c *copier) createSignature(manifest []byte, keyIdentity string, passphrase
...
+	if passphrase != "" {
+		opts = append(opts, simple.WithPassphrase(passphrase))
+	}
+	signer, err := simple.NewSigner(opts...)
+	if err != nil {
+		return nil, err
+	}
+	defer signer.Close()
+
+	c.Printf("%s\n", internalSigner.ProgressMessage(signer))
```
vrothberg: Note to self: why does this work?  Shouldn't it be `signer.ProgressMessage()`?

signature/simple/signer.go
```diff
@@ -0,0 +1,104 @@
...
+	"strings"
+
+	"github.com/containers/image/v5/docker/reference"
+	internalSig "github.com/containers/image/v5/internal/signature"
+	internalSigner "github.com/containers/image/v5/internal/signer"
+	"github.com/containers/image/v5/signature"
+	"github.com/containers/image/v5/signature/signer"
+)
+
+// simpleSigner is a signer.Signer implementation for simple signing signatures.
```
vrothberg, line 15: Nit: Could we embed GPG in the name?  It would be more expressive to me when reading.

copy/copy.go
```diff
@@ -124,17 +125,30 @@ type ImageListSelection int
...
-	RemoveSignatures                 bool            // Remove any pre-existing signatures. SignBy will still add a new signature.
-	SignBy                           string          // If non-empty, asks for a signature to be added during the copy, and specifies a key ID, as accepted by signature.NewGPGSigningMechanism().SignDockerManifest(),
-	SignPassphrase                   string          // Passphare to use when signing with the key ID from `SignBy`.
-	SignBySigstorePrivateKeyFile     string          // If non-empty, asks for a signature to be added during the copy, using a sigstore private key file at the provided path.
-	SignSigstorePrivateKeyPassphrase []byte          // Passphare to use when signing with `SignBySigstorePrivateKeyFile`.
+	RemoveSignatures bool // Remove any pre-existing signatures. SignBy will still add a new signature.
+
+	// Signers to use to add signatures during the copy.
+	// Callers are still responsible for closing these Signer objects; they can be reused for multiple copy.Image operations in a row.
+	Signers []*signer.Signer
```
vrothberg, line 132: I don't have a strong opinion on that.  One worry I have is boilerplate code for callers.  skopeo, libimage and CRI-O would all need updates, wouldn't they?

vrothberg (commented):
LGTM

copy/sign.go
vrothberg: Ah, OK got it now :heavy_check_mark:

mtrmac: What would be a good place to document how this is intended to work?

copy/copy.go
mtrmac: Totally, this would need a common utility. (Not CRI-O, it doesn’t create signatures.)

Right now there is a WIP https://github.com/mtrmac/image/tree/fulcio-signing/pkg/cli/sigstore : just importing `copy.Image` adds dependencies enough to _validate_ Fulcio + Rekor signatures, and to sign using private keys, but importing this `cli/sigstore` subpackage drags in the Fulcio + Rekor RPC clients.

I actually fairly strongly think that we need to go further than that, and to expose a config file for signing with Fulcio+Rekor (because there are like 6 individual options to set). And then `NewSignerFromSigstoreSigningConfigFile` would be a natural utility to share across various `copy.Image` callers to implement that config file.


---

If the comment is about the deprecation of existing `SignBy…` options, I now think that’s something that is eventually a good idea, but should _not_ be done in this PR; it’s an internal cleanup for no end-user benefit and the end-user functionality is much more important short-term.

signature/simple/signer.go
mtrmac: Should it even be `signature/simple`? `signature/simplesigning` would be consistent with `internal/signature.SimpleSigning`.

mtrmac: Moved to `signature/simplesigning` now, but that’s not intended to stop the search for another name.

signature/simplesigning/signer_test.go
```diff
@@ -0,0 +1,234 @@
...
+
+		sig, err := internalSigner.SignImageManifest(s, manifest, testImageSignatureReference)
+		require.NoError(t, err, c.name)
+		simpleSig, ok := sig.(internalSig.SimpleSigning)
+		require.True(t, ok)
+
+		// FIXME FIXME: gpgme_op_sign with a passphrase succeeds, but somehow confuses the GPGME internal state
+		// so that gpgme_op_verify below never completes (it polls on an already closed FD).
+		// That’s probably a GPG bug, and needs investigating and fixing, but it isn’t related to this “signer” implementation.
+		if len(c.opts) == 0 {
```
mtrmac, line 191: This is a bug that certainly needs tracking down: https://github.com/containers/image/issues/1777 .

“Luckily” it probably only affects long-term processes that sign (and verify?), i.e. not Podman and not CRI-O.

signature/simplesigning/signer_test.go
```diff
@@ -0,0 +1,234 @@
...
+	"github.com/opencontainers/go-digest"
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	// testImageManifestDigest is the Docker manifest digest of "image.manifest.json"
+	testImageManifestDigest = digest.Digest("sha256:20bf21ed457b390829cdbeec8795a7bea1626991fda603e0d01b4e7f60427e55")
+	testGPGHomeDirectory    = "./testdata"
```
mtrmac, line 21: See the commit message about using separate GPG home directories.

There _still_ is some flake that is fairly frequently triggered by two subpackages testing GPG concurrently, resulting in a signing operation failing with “end of file”. I haven’t tracked that one down. Short—term, I’d rather have a flake than an untested signing code, but that’s assuming the pain won’t be that big in CI.

mtrmac (commented):
Now updated:
- With tests
- The deprecation was moved to another branch, to be done (much?) later
- Renaming `signature/simple` to `signature/simplesigning`
- GPG tests now terminate `gpg-agent` processes so that we don’t leave several around.

Still a draft because final users don’t exist, but I think this PR is fairly close to finished.

copy/sign.go
vrothberg: I have no idea.  I think it's OK as is and it was immediately obvious in the code. Just when reading the commits one-by-one I was surprised for a moment.

signature/simple/signer.go
vrothberg: `simplesigning` SGTM
It's somehow a known terminology in our universe.

copy/copy.go
vrothberg: SGTM

I am mostly worried to end up like in a pre-libimage period where the tools have duplicate code that must be maintained separately.

A config file sounds good to me and makes perfect sense.  It seems more user friendly (and code friendly) to have one central place to configure the setup rather than passing things from the CLI through the entire call stack.

vrothberg (approved):
LGTM, nice work!

rhatdan:
@mtrmac is this still Draft?

mtrmac:
Yes, I want to have a testable Skopeo binary with the anticipated features before committing to an API.

Notably the `signature/sigstore.NewSignerWithPrivateKeyFileUnstable` should not ship like that.

mtrmac:
Updated now:
- Split some of the conceptually test changes into #1779 
- Replaced `signature/sigstore.NewSignerWithPrivateKeyFileUnstable` (and the current `signature/sigstore.SignDockerManifestWithPrivateKeyFileUnstable`) with `signature/sigstore.NewSigner` and the functional-options pattern, which allows adding Rekor and Fulcio as _optional_ dependencies. See https://github.com/mtrmac/image/tree/fulcio-signing for intended use
- The signing unit tests in `copy` now just test identity using a stub `Signer`, instead of actually running GPG. That way they are truly unit tests, and cheaper.

I think this is ready for final review, marked as draft due to the #1779 dependency.

vrothberg (commented):
LGTM

vrothberg (approved):
LGTM

PR #1780, opened 2023-01-07
Speed up pkg/blobcache tests

mtrmac:
- Use an invalid SystemContext.BlobInfoCacheDir to trigger use of blobinfocache/memory instead of blobinfocache/boltdb, thus avoiding some Fsync calls. This alone more than halves the total test time.
- Also don't repeat the test with 5 different file sizes; file contents are not _that_ much of a focus of the tests that we need to test so many variants.

On macOS, this reduces test time by 81% (measured by -count=3), from (in this particular instance) 8.8s to 1.6s per iteration.

### DIFF
--- a/pkg/blobcache/blobcache_test.go
+++ b/pkg/blobcache/blobcache_test.go
@@ -69,9 +69,9 @@ func makeLayer(filename string, repeat int, compression archive.Compression) ([]
 func TestBlobCache(t *testing.T) {
 	cacheDir := t.TempDir()
 
-	systemContext := types.SystemContext{}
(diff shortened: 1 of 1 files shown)


mtrmac:
Of course, this not that big a deal. It’s just that `pkg/blobcache` is retested on every `c/image/copy` change because the tests call `copy.Image`, which makes the delay a big more annoying than it would be for other subpackages.

rhatdan:
LGTM
@vrothberg PTAL

vrothberg (approved):
LGTM, nice!

PR #1782, opened 2023-01-09
Make sure value types also implement json.Marshaler

mtrmac:
... otherwise `json.Marshal(value)` ignores the custom implementation.

Luckily the implementations were fine, so just add the extra safeguards.

(The `_` values are, AFAICS, optimized out and do not increase binary size. Usually I'd move the assertions to test code anyway, but for security-critical code it seems better to have the assertion nearby.)


mtrmac:
As ever, Go is a worse-is-better language that must be compensated for by manual effort.

rhatdan:
LGTM
@vrothberg PTAL

vrothberg (approved):
LGTM

Issue #1601, opened 2022-07-07
Cosign technical debt


mtrmac:
Captured from PRs

- [ ] Unit tests for #1595 , at least the config file handling
- [ ] Unit tests for #1597
- [ ] Integration tests for #1595
- [ ] Integration tests for #1597
- [ ] Integration tests for #1598
- [ ] Only fetch the attachment config if we actually have a new signature to add (and if we continue to update the DiffID list).
- [ ] Only upload an updated attachment config and manifest if we added at least one new signature.
- [ ] Consider accepting any of a set of public keys
- [ ] Possibly test `isManifestUnknown` from #1595 against various registries?
- [ ] Possibly consider accepting an image with a signed manifest list but not the individual images (i.e. `cosign sign` without `--recursive`)

mtrmac:
I have moved the Skopeo-specific items to https://github.com/containers/skopeo/issues/1704 .

mtrmac:
https://github.com/letsencrypt/boulder/pull/6651 decreases Skopeo binary size on macOS by 295 kB.

PR #1784, opened 2023-01-09
Add support for Rekor (“transparency log”) uploads to sigstore signing

mtrmac:
Tested manually with Skopeo; unit tests to hopefully come later per https://github.com/containers/image/issues/1601 .

This provides an API; hooking up to the CLI or other code will be added later.

To test, see the commands at the end of https://github.com/mtrmac/skopeo/blob/c-image-integration-experiment/cmd/skopeo/cosign_test.go (a branch in a horrible state, which nevertheless contains all pending pieces):
`skopeo copy --sign-by-sigstore=path/to/file`, demonstrates both Rekor + private key, and Rekor + Fulcio.


signature/internal/rekor_set.go
```diff
@@ -0,0 +1,98 @@
+package internal
+
+import (
+	"encoding/json"
+)
+
+// This is the github.com/sigstore/rekor/pkg/generated/models.Hashedrekord.APIVersion for github.com/sigstore/rekor/pkg/generated/models.HashedrekordV001Schema.
+// We could alternatively use github.com/sigstore/rekor/pkg/types/hashedrekord.APIVERSION, but that subpackage adds too many dependencies.
+const HashedRekordV001APIVersion = "0.0.1"
```
rhatdan, line 9: Can you open an issue with them to not pull in imports when defining types?

mtrmac: The `APIVERSION` constant lives in a code-focused package that implements various aspects of that type.

The actual type definition lives in https://github.com/sigstore/rekor/blob/main/openapi.yaml , the package we import is Swagger-generated, so that’s already the smaller package. (In fact it’s the smaller one, which is why it doesn’t carry the constant.)

It might be possible to change the YAML file so that a constant is included in the generated model, I couldn’t immediately find a way (if the `version` field were an enum, it would reject unknown versions completely).

Really, if anything, we’d like to avoid using even the generated types package, because that already adds too many dependencies. So asking upstream to change packages we don’t want to use feels a bit unreasonable.

vrothberg (approved):
LGTM

haydentherapper:
fyi @asraa

rhatdan:
LGTM

PR #1785, opened 2023-01-09
Fulcio signing implementation

mtrmac:
This adds API to use Fulcio-generated short-lived certificates.

Depends on #1784 .

I still need to test one part but it seems broadly ready.

I don’t feel _too_ confident about the API, but it seems close enough, and we can always add more option functions, or a new `signature/sigstore/fulciov2` package (still operating on `signature/sigstore/internal.Signer`)

PR #1787, opened 2023-01-11
Add a sigstore signing parameter file format, and CLI utility

mtrmac:
This defines a new file format (`containers-sigstore-signing-params.yaml(5)`), and provides an API to integrate it into callers.

That way, `skopeo` and `podman` can add a single `--sigstore-sign-by=`_param-file_ option, instead of 9 options; and users can maintain their site-specific configuration in one file, instead of manually copy&pasting 6 options in a typical case.

The parameter file format definition is a separate package that does not drag in the Rekor and Fulcio dependency tree; only actually _using_ the file has that cost.

I’m not too happy about the “containers sigstore signing parameters” name, nor about the `sigstore/params` subpackage name; I’ll certainly welcome suggestions.

(One possibility would be to make this a “containers signing parameters file”, not sigstore-specific, and to move all options in that file into a `sigstore` sub-object; then we could add a “simple signing” option to that same file as well.)

I also didn’t bother with using “functional options” for the `NewSignerFromParameterFile` function, nor with hiding the actual file format and only providing a constructor or an edit API. This is a non-essential utility and we can always provide a `cli/fulciov2` without impacting any existing callers, so keeping the API opaque and stable over time is not _that extremely_ important in this case.

Depends on #1785 .

@vrothberg PTAL. @rhatdan FYI.

### DIFF
--- a/docs/containers-sigstore-signing-params.yaml.5.md
+++ b/docs/containers-sigstore-signing-params.yaml.5.md
@@ -1,0 +1,117 @@
+% CONTAINERS-SIGSTORE-SIGNING-PARAMS.YAML 5 sigstore signing parameters Man Page
+% Miloslav Trmač
+% January 2023
+
+# NAME
+containers-sigstore-signing-params.yaml - syntax for the sigstore signing parameter file
+
+# DESCRIPTION
+
+Sigstore signing parameter files are used to store options that may be required to create sigstore signatures.
+There is no default location for these files; they are user-managed, and used as inputs to a container image signing operation,
+e.g. `skopeo copy --sign-by-sigstore=`_param-file_`.yaml` or `podman push --sign-by-sigstore=`_param-file_`.yaml` .
+
+## FORMAT
+
+Sigstore signing parameter files use YAML.
+
+Many parameters are optional, but the file must specify enough to create a signature;
+in particular either a private key, or Fulcio.
+
+### Signing with Private Keys
+
+- `privateKeyFile:` _path_
+
+   Create a signature using a private key at _path_.
+   Existence of this field triggers the use of a private key.
+
+- `privateKeyPassphraseFile:` _passphrasePath_
+
+   Read the passphrase required to use `privateKeyFile` from _passphrasePath_.
+   Optional: if this is not set, the user must provide the passphrase interactively.
+
+### Signing with Fulcio-generated Certificates
+
+Instead of a static private key, the signing process generates a short-lived key pair
+and requests a Fulcio server to issue a certificate for that key pair,
+based on the user authenticating to an OpenID Connect provider.
+
+To specify Fulcio, include a `fulcio` sub-object with one or more of the following keys.
+In addition, a Rekor server must be specified as well.
+
+- `fulcioURL:` _URL_
+
+  Required. URL of the Fulcio server to use.
+
+- `oidcMode:` `interactive` | `deviceGrant` | `staticToken`
+
+  Required. Specifies how to obtain the necessary OpenID Connect credential.
+
+  `interactive` opens a web browser on the same machine, or if that is not possible,
+  asks the user to open a browser manually and to type in the provided code.
+  It requires the user to be able to directly interact with the signing process.
+
+  `deviceGrant` uses a device authorization grant flow (RFC 8628).
+  It requires the user to be able to read text printed by the signing process, and to act on it reasonably promptly.
+
+  `staticToken` provides a pre-existing OpenID Connect “ID token”, which must have been obtained separately.
+
+- `oidcIssuerURL:` _URL_
+
+  Required for `oidcMode:` `interactive` or `deviceGrant`. URL of an OpenID Connect issuer server to authenticate with.
+
+- `oidcClientID:` _client ID_
+
+  Used for `oidcMode:` `interactive` or `deviceGrant` to identify the client when contacting the issuer.
+  Optional but likely to be necessary in those cases.
+
+- `oidcClientSecret:` _client secret_
+
+  Used for `oidcMode:` `interactive` or `deviceGrant` to authenticate the client when contacting the issuer.
+  Optional.
+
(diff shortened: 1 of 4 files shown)


mtrmac:
Tested manually in Skopeo https://github.com/containers/skopeo/pull/1849 , basically with the provided example configs.

Adding unit tests is tracked in #1601.

mtrmac:
Podman WIP: https://github.com/containers/common/pull/1288 https://github.com/containers/podman/pull/17088/files

vrothberg (approved):
LGTM

mtrmac:
… and I do want to encourage changing the name of the file. I’m not happy at all with the current one.

mtrmac:
Updated to quote YAML strings, motivated by https://ruudvanasseldonk.com/2023/01/11/the-yaml-document-from-hell .

(I chose to avoid TOML because of https://github.com/containers/image/issues/1054 , and YAML’s expression of nesting is much nicer. But I’m fine with any other format, to be consistent with the rest of the ecosystem.)

vrothberg:
> Updated to quote YAML strings, motivated by https://ruudvanasseldonk.com/2023/01/11/the-yaml-document-from-hell .
> 
> (I chose to avoid TOML because of #1054 , and YAML’s expression of nesting is much nicer. But I’m fine with any other format, to be consistent with the rest of the ecosystem.)

LGTM

I don't have an opinion on the file format.  Made me think of Mutt's slog: "All mail clients suck. This one just sucks less."

docs/containers-sigstore-signing-params.yaml.5.md
```diff
@@ -0,0 +1,117 @@
+% CONTAINERS-SIGSTORE-SIGNING-PARAMS.YAML 5 sigstore signing parameters Man Page
+% Miloslav Trmač
+% January 2023
+
+# NAME
+containers-sigstore-signing-params.yaml - syntax for the sigstore signing parameter file
+
+# DESCRIPTINO
```
TomSweeneyRedHat, line 8:
```suggestion
# DESCRIPTION
```

docs/containers-sigstore-signing-params.yaml.5.md
```diff
@@ -0,0 +1,117 @@
...
+
+Sigstore signing parameter files are used to store options that may be required to create sigstore signatures.
```
TomSweeneyRedHat, line 10: I thought this was going to be a single file?  If so, suggest:
```suggestion
The sigstore signing parameter file is  used to store options that may be required to create sigstore signatures.
```

TomSweeneyRedHat: Ignore, I see where you were getting at down below.

docs/containers-sigstore-signing-params.yaml.5.md
```diff
@@ -0,0 +1,117 @@
...
+There is no standard location for these files; they are user-managed, and used as inputs to a container image signing operation,
```
TomSweeneyRedHat, line 11: On the fence, your call.  "standard location" or "default location"?

mtrmac: (I’m not very sure that sentence even needs to exist… I just wanted to give _some_ context for users who stumble on the page directly, without starting at a `--sign-by-sigstore` option documentation.)

Sure, “default location” works.

docs/containers-sigstore-signing-params.yaml.5.md
```diff
@@ -0,0 +1,117 @@
...
+  Required. Specifies how to obtain the necessary OpenID Connect credential.
+
+  `interactive` opens a web browser on the same machine, or if that is not possible,
+  asks the user to open a browser manually and to type in the provided code.
+  It requires the user to be able to directly interact with the signing process.
+
+  `deviceGrant` uses a device authorization grant flow (RFC 8628).
+  It requires the user to be able to read text printed by the signing process, and to act on it reasonably promptly.
+
+  `staticToken` provides a pre-existing Open ID Connect ID token which must have been obtained separately.
```
TomSweeneyRedHat, line 57: Based solely on grokking Google, it looks like it should be "OpenID Connect".  (https://openid.net/connect/).  Please change here and elsewhere.  I've added quotes to set it off from the rest of the text and to let the reader know that's a thing that could be looked up if they're unfamiliar as I was.  That said, I'm still on the fence about having the quotes, so no foul if you want to drop them.

```suggestion
  `staticToken` provides a pre-existing "OpenID Connect" ID token which must have been obtained separately.
```

TomSweeneyRedHat: I'll hop off the fence, drop the quotes.  I see you have a number of instances and it would look odd all over.

mtrmac: Thanks! The quotes are a great idea for separating the two terms.

(I didn’t find any other instance of “Open ID” with a space.)

TomSweeneyRedHat:
Couple doc nits which I think @mtrmac has tended to already, otherwise, LGTM.

docs/containers-sigstore-signing-params.yaml.5.md
mtrmac: How about quoting the “ID token” part instead? That would be less useful as a stand-alone search term, but assuming the readers figure out `OpenID Connect` is a term to search from other parts of the page, making it clear “ID token” is a single phrase might be worthwhile.

mtrmac:
@TomSweeneyRedHat Thanks a lot!

docs/containers-sigstore-signing-params.yaml.5.md
TomSweeneyRedHat: Works for me!

mtrmac: Thanks, changed to quote “ID token”.

TomSweeneyRedHat (commented):
LGTM

PR #1788, opened 2023-01-11
Do not preallocate regex in init program

rhatdan:


TomSweeneyRedHat:
LGTM
would like a head nod from @vrothberg or @mtrmac

mtrmac (commented):
For each of these subpackages, when would that subpackage be actually imported (i.e. not compiled out) _and_ not trigger compilation of the regexp?

- Sure, `ostree`, but that’s already gated by a build tag.
- `oci/internal`, if dragged in only via `transports/alltransports` and not used at runtime, I guess.
- `signature`, if the policy never use `remapIdentity`, OK…
- Clearly not `docker` or `docker/reference`, nor `sysregistriesv2`, every use is going to trigger this.

Is there a measurable improvement? I wouldn’t be surprised if the extra `sync.Once` on every use _increased_ the total run time.

---

Aesthetically, I’m not enthusiastic. This is tediously doing something the compiler should do for us — _and_ if the compiler ever gets better, we probably won’t benefit because the compiler won’t figure the opportunistic initialization out.

I guess the closest we actually have from the compiler as it exists is `go:generate` — but I couldn’t find any regexp package using that.

---

My first impression:
- Let’s not touch the subpackages where the regexp is going to be used anyway.
- That leaves `oci/internal`, `ostree`, and `signature` I think.
- For those, I’d prefer a less repetitive code. Factor the `sync.Once` into a subpackage, either
  ```go
  // global
  var remapIdentityDomainRegexp = delayedRegexp.New("…")
  // use
  remapIdentityDomainRegexp.Compile().MatchString(…)
  ```
  keeping the existing shape of the code, or maybe even
  ```go
   // global
   var remapIdentityDomainRegexp = delayedRegexp.Cache()
   // use
   remapIdentityDomainRegexp.Compile("…").MatchString()
   ```
   which would move the actual regexp closer to the user, a nice improvement (OTOH only for single-user regexes).
- And with that utility, we could plausibly, without too much hassle, delay-compile _individual_ regexes within the `docker/reference` package (OTOH I didn’t actually check whether there are any independent subsets).

But, really, is this measurable? https://github.com/containers/image/pull/1733 made a difference because there were _so many_ individual compilations. But these few?

mtrmac:
(I have updated the review comment which was submitted accidentally.)

mtrmac:
> For each of these subpackages, when would that subpackage be actually imported (i.e. not compiled out) _and_ not trigger compilation of the regexp?

I completely forgot about `c/storage/pkg/reexec`, which means we pay the `init` cost for every package in the binary on various executions that don’t come _close_ to running anything else from the package.

rhatdan:
@mtrmac @vrothberg PTAL

vrothberg (commented):
LGTM

I don't expect this PR to have a significant impact on Podman though.  Image references are almost always parsed.

mtrmac (approved):
LGTM. Thanks!

mtrmac:
>  Image references are almost always parsed.

I think `regexp.Delayed` primarily speeds up `c/storage/pkg/reexec`, where we don’t get to use the compiled objects at all.

PR #1789, opened 2023-01-12
manifest,zstd: give priority to `zstd` compressed images when pulling image from a manifest list

flouthoc:
While perform a pull from manifest list of type `application/vnd.oci.image.index.v1+json` make sure that if more than one image exists for the same platform then give priority to the one compressed with `zstd`.

Annotation `io.github.containers.compression.zstd` is expected to be set for the image compressed with `zstd`.

For users who wish to test this feature with buildah/podman

```console
# create manifest list
buildah manifest create host/repo:latest
# create image 
buildah build -t test1 .
# create another image
buildah build -t test2 .

# add test1 to manifest list
buildah manifest add host/repo:latest sha:<test1>
# add test2 to manifest list
buildah manifest add host/repo:latest sha:<test2>

# tag zstd compressed image
buildah manifest annotate --annotation io.github.containers.compression.zstd=true host/repo:latest sha256:<test1>
# push to the repo
buildah manifest push --all host/repo:latest docker://host/repo:latest

# Fresh system !!! Pull from manifest list should give priority to zstd compressed image
buildah pull host/repo:latest

```


manifest/oci_index.go
```diff
@@ -89,11 +98,20 @@ func (index *OCI1Index) ChooseInstance(ctx *types.SystemContext) (digest.Digest,
 				Variant:      d.Platform.Variant,
 			}
 			if platform.MatchesPlatform(imagePlatform, wantedPlatform) {
-				return d.Digest, nil
+				matchedDigest = d.Digest
+				// Always give priority to zstd if more than one image
```
vrothberg, line 102: I am not sure about doing that unconditionally.  In theory, this can break some use cases that rely on always picking gzip.  Shouldn't there be (at least) a way to opt out of this behavior?  Could be config knob?

internal/image/fixtures/schemazstdselection.json
```diff
@@ -0,0 +1,27 @@
+{
+    "schemaVersion": 2,
+    "mediaType": "application/vnd.oci.image.index.v1+json",
+    "manifests": [
+        {
+            "mediaType": "application/vnd.oci.image.manifest.v1+json",
+            "digest": "sha256:ac9bbf71dacfad2bfbe48f7d46ca279b1b3b351a6c814db760030015167405b8",
+            "size": 772,
+            "annotations": {
+                "io.github.containers.compression.zstd": "true"
```
vrothberg, line 10: We need more tests.  In this particular case, the first image would always be picked - with or without the code changes.  The crux of the matter is to walk the entire list and pick the ZSTD matching one.

In the end, the gzip compressed ones must always be listed before the zstd ones to remain backwards compatible.

manifest/oci_index.go
vrothberg: That would imply walking the entire list, recording the matching ones but the pick the first ZSTD one.

flouthoc: A config knob sounds good to me , let me think of a way without breaking any API.

> That would imply walking the entire list, recording the matching ones but the pick the first ZSTD one.

True

internal/image/fixtures/schemazstdselection.json
flouthoc: Added another test where gzip is the first listed image , and added more images in unordered manner. Could you suggest  some more tests if i missed anything.

vrothberg: There could be a manifest with multiple matching gzip-compressed and multiple matching zstd-compressed images.  We should test this case and make sure to select the expected one(s).  Either the first gzip or the first zstd one - depending on what we're looking for.

manifest/oci_index.go
```diff
@@ -12,6 +12,13 @@ import (
 	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
 )
 
+const (
+	// Following annotation is used while performing instance selection.
+	// If more than one image exists for the same platform then the image
+	// compressed with `zstd` will be given priority.
+	ZSTDInstanceAnnotation = "io.github.containers.compression.zstd"
```
mtrmac, line 19: This either shouldn’t be a part of the public API, or if we want to make a Go constant, it should be in a separate subpackage (`c/image/manifest/compression`?) so that it is usable without dragging in all of the dependencies of `c/image/copy`.

---

Also, the name (together with the package name) should be sufficiently descriptive. “instance” of what? But let’s figure out the public/nonpublic decision first.

manifest/oci_index.go
```diff
@@ -89,11 +98,20 @@ func (index *OCI1Index) ChooseInstance(ctx *types.SystemContext) (digest.Digest,
 				Variant:      d.Platform.Variant,
 			}
 			if platform.MatchesPlatform(imagePlatform, wantedPlatform) {
-				return d.Digest, nil
+				matchedDigest = d.Digest
```
mtrmac, line 101: AFAICS this changes behavior: before were using the first match from equal (non-std ones), now we are using the last one.

The original behavior needs to be preserved (it’s essential to the design of the compatibility mechanism).

manifest/oci_index.go
```diff
@@ -89,11 +98,20 @@ func (index *OCI1Index) ChooseInstance(ctx *types.SystemContext) (digest.Digest,
...
+				}
 			}
 		}
 	}
 
+	if matchedDigest != "" {
+		return matchedDigest, nil
+	}
+
 	for _, d := range index.Manifests {
```
mtrmac, line 115: Shouldn’t we use the same logic here as well?

That probably suggests defining some kind of `instanceBetterMatchesPlatforms` predicate (which can be unit-tested separately), and this function reducing to a single loop that finds the best entry.

manifest/oci_index.go
mtrmac: - _Typically_ `CopySystemImage` is used for pulls. In that case we don’t need an option at all.
- Pushes don’t have a Zstd source, so they don’t need an option either.
- That leaves `skopeo copy docker://… docker://…`, which indeed uses `CopySystemImage` by default. I’m not too sure we need an option for this _now_ (as opposed to some later `copy --compression={ensure:zstd, strip:zstd, only:gzip}` mechanism) , but it wouldn’t hurt.

I’d expect a new option in `copy.Option` to be the natural way to express this.

And yes, that would require adding an `internal/manifest.List`, so that we can add/change a method, somewhat similarly to how we now have `internal/image`. This will be necessary to allow adding / removing compression variants in the future, either way.

And that `internal/manifest` work probably be a separate (fairly large) PR just to build that infrastructure, with no new features.

internal/image/fixtures/schemazstdselection.json
mtrmac: (Note to self: I didn’t look at the test coverage.)

mtrmac:
> # Fresh system !!! Pull from manifest list should give priority to zstd compressed image

I’d expect `skopeo copy [--multi-arch=system] docker://… docker://…` to expose this logic without needing to wipe c/storage.

manifest/oci_index.go
mtrmac: FWIW the `internal/manifest` could only contain the `List` parts we need, no need to migrate everything.

manifest/oci_index.go
flouthoc: Matched now and added a test.

manifest/oci_index.go
flouthoc: Made a new const `OCI1InstanceAnnotationCompressionZSTD`

manifest/oci_index.go
```diff
@@ -89,16 +98,45 @@ func (index *OCI1Index) ChooseInstance(ctx *types.SystemContext) (digest.Digest,
...
 		if d.Platform == nil {
-			return d.Digest, nil
+			if matchedDigest == "" {
+				matchedDigest = d.Digest
+			}
+			// Always give priority to zstd if more than one image
+			// exists for the same platform.
+			// TODO: In future this check will be configurable and
+			// will be relayed via `ctx`.
```
vrothberg, line 128: I am not thrilled about delaying making it conditional.  There's a risk this commit could make it into a release and that may break someone on something somewhere.

manifest/oci_index.go
vrothberg: A new field in `copy.Option` sounds good.  This would nicely cover CLI use cases - however it will be exposed on the CLI.

But I also desire an option in a configuration file.  I think that registries.conf can be a nice host for that.  This could, for now, be sufficient to make the ZSTD-preference conditional (and ideally opt-in).  This way, all users of c/image would be covered (incl. CRI-O if OpenShift decided to default to ZSTD in the future).

flouthoc: I have made is configurable in `systemContext` and as discussed yesterday later on we can extend `copy.Option` to propagate it to systemContext

manifest/oci_index.go
flouthoc: Made it configurable in a new commit.

manifest/oci_index.go
flouthoc: Resolving this so we can discuss on the new config option.

flouthoc:
@vrothberg @mtrmac PTAL again as discussed before I have made it configurable options and later on it can be relayed to `SystemContext` via `copyOptions`.

TomSweeneyRedHat:
@flouthoc, I think you need a rebase.

manifest/oci_index.go
```diff
@@ -69,36 +76,60 @@ func (index *OCI1Index) UpdateInstances(updates []ListUpdate) error {
...
 				Variant:      d.Platform.Variant,
 			}
-			if platform.MatchesPlatform(imagePlatform, wantedPlatform) {
-				return d.Digest, nil
+		}
+		if (platformCheck && platform.MatchesPlatform(imagePlatform, wantedPlatform)) || (!platformCheck && d.Platform == nil) {
+			if matchedDigest == "" {
+				matchedDigest = d.Digest
+			}
+			if ctx.GivePriorityToZstdImages != types.OptionalBoolFalse {
```
TomSweeneyRedHat, line 100: maybe not enough tea yet, but should there be an else here?  It looks like we will at best just set `matchedDigest` to `d.Digest` again in this if....

mtrmac:
I’m afraid I’ll only return to this carefully next week.

Just to give a brief (aesthetic ?) preference: Eventually most of the compression conversion options (add a variant, drop a variant, ensure a variant exists) are going to exist in `copy.Options`, and make no sense outside of `copy.Options`. So I’m not too enthusiastic about adding an option to `types.SystemContext` now — we would keep it forever mostly just as an artifact of the order in which we develop things; that seems _entirely avoidable_. Or are we in some rush to deliver this immediately? (OTOH, admittedly, adding an option next to things like `ArchitectureChoice` would be fairly clean.)

So I’d prefer starting with making `manifest.List` private, and an option in `copy.Options`, not `SystemContext`. (In general, `SystemContext` is a bit of a mistake.)

---

> But I also desire an option in a configuration file. I think that registries.conf can be a nice host for that.

But wouldn’t such an option be necessary also for the two OCI transports? That wouldn’t work with `registries.conf`. Especially if the thinking is that this should be a fail-safe that is disabled by default, that option seems like something that shouldn’t be transport-specific.

It might well make sense to add a `registries.conf`, per-scope, override mechanism, but that seems to me to be a more advanced feature, not essential for the very first version. (It would also require passing that choice through `private.ImageDestination`, because `registries.conf` is `docker://`-specific.)

manifest/oci_index.go
flouthoc: @TomSweeneyRedHat This is needed to preserve the original logic of image selection , also added a test to verify that we preserve the original logic. But please unresolve if you think something is missing :)

flouthoc:
@mtrmac adding it to `copy.Options` and passing it to `manifest.List` SGTM. Just a small doubt since we had a discussion on the before and there was a preference to do both of these in a different PR. Should I create a new PR for it or are you fine in this PR itself ?

>  Or are we in some rush to deliver this immediately? 

I don't think there is a rush, I think getting this feature in the right way would be better :)

mtrmac:
I’d prefer a separate PR to introduce (probably) `internal/manifest.List` as an extension / replacement (not 100% sure which one, preferably an extension _if possible_) for the stable public `manifest.List`; with no new added functionality.

Then this PR can add/change the internal `ChooseInstance` (which no longer has public API restrictions) to add a compression option, and use that from `copy`.

manifest/oci_index.go
TomSweeneyRedHat: Just to make sure we're talking the same change, my suggestion is:

```
		if matchedDigest == "" {
				matchedDigest = d.Digest
		} else {
			     if ctx.GivePriorityToZstdImages != types.OptionalBoolFalse {
                                // Always give priority to zstd if more than one image
                  	       // exists for the same platform.
				if _, ok := d.Annotations[OCI1InstanceAnnotationCompressionZSTD]; ok {
					matchedDigest = d.Digest
					return matchedDigest
				}
			}
		}
```

(spacing, tabs, brackets way off).  It's not much of a times saving, but I don't think `d.Digest` changes at all between the two lines that assign its value to matchedDigest, so no need to spend time on the two if statements if matchedDigest is already set.

But perhaps I'm misreading?

types/types.go
```diff
@@ -671,6 +671,9 @@ type SystemContext struct {
 	CompressionFormat *compression.Algorithm
 	// CompressionLevel specifies what compression level is used
 	CompressionLevel *int
+	// Give priority to pulling zstd images if multiple images are present unless explicitly
+	// configured false.
+	GivePriorityToZstdImages OptionalBool
```
giuseppe, line 676: how would it work if we ever add another compression level?

Isn't it better to have something like `PreferGzip` to maintain the current behavior?

I think the new zstd priority code should be the default, and have a way to opt-out.  In the long term we want everyone to move to using zstd so IMO it makes more sense to make it the default and have a possibility to opt-out when needed

flouthoc: @giuseppe `PreferGzip` sounds munch better but as discussed above, not using this `OptionalBool` anymore , I'll add a new option to `copy.Options` and will pass it down to the `internal/manifest.List`'s ChooseInstance method so will add a `PreferGzip` there.

rhatdan: I agree.

flouthoc:
@mtrmac PR for flipping dependency and introducing `internal.manifest.List` https://github.com/containers/image/pull/1791 PTAL

manifest/oci_index.go
```diff
@@ -69,36 +76,60 @@ func (index *OCI1Index) UpdateInstances(updates []ListUpdate) error {
...
-	wantedPlatforms, err := platform.WantedPlatforms(ctx)
-	if err != nil {
-		return "", fmt.Errorf("getting platform information %#v: %w", ctx, err)
-	}
-	for _, wantedPlatform := range wantedPlatforms {
-		for _, d := range index.Manifests {
+func instanceBetterMatchesPlatforms(ctx *types.SystemContext, manifests []imgspecv1.Descriptor, platformCheck bool, wantedPlatform imgspecv1.Platform) digest.Digest {
+	var matchedDigest digest.Digest
+	matchedDigest = ""
+	for _, d := range manifests {
```
mtrmac, line 82: This structure feels _very_ unnatural to me.

There is now a loop in  `ChooseInstance`; that loop terminates on a first match, and calls `instanceBetterMatchesPlatform`, which initiates _another_ loop, and always finds an image, and returns it, and that terminates `ChooseIntance`, AFAICS. Why have two loops at all?

The specific check for `GivePriority`, and the _immediate_ return in that case, also doesn’t generalize to the future at all. We might have more compression formats.

---

The suggestion with `instanceBetterMatchesPlatforms` was to define it as a predicate that defines an ordering on platforms; and then `ChooseInstance` can just find a minimum = best match.

Conceptually:
```swift
func ChooseInstance(ctx: SystemContext) -> Digest {
    let wantedPlatforms = platform.WantedPlatforms(ctx: ctx)
    let relevant = index.Manifests.filter({ manifest in
        platformMatches(manifest: manifest, wantedPlatforms: wantedPlatforms)
    })
    let best := relevant.min(by: { x, y in instanceComparator(x, y, platformOrder: wantedPlatforms) == .orderedAscending })
    return best?.Digest
}

func instanceComparator(x, y Descriptor, platformOrder: …) -> ComparisonResult {
    if let v = platformComparator(x, y, orderingBy: platformOrder); v != .orderedSame { return v }
    if let v = compressionComparator(x, y, preferenceBasedOn: compressionPreferenceOptions); v != .orderedSame { return v }
    return .orderedSame
}
```

Now, the actual implementation would probably look a bit different — e.g. finding out the platform ordering means a linear search through `wantedPlatforms`, so we would only want to do that once per item, instead of deep in the `instanceComparator`, but the _rough idea_ was to separate the “is this a better match” and the iteration code.

types/types.go
mtrmac, line 676: Just to explicitly repeat this, this should be controlled in `copy.Options`, not in `types.SystemContext`.

mtrmac: I think, at this point, we are somewhat experimenting and the options are therefore very specific. `PreferZstdOverGzip` is not a good long-term API, but it’s something we can do now without too much design.

Alternatively, we _could_ draft what this should look like in some end state, generalized to multiple compression formats, and the like:
```go
    EnsureCompressionVariantsExist Set<CompressionVariant> // add these if they don’t exist
    StripCompressionVariants Set<CompressionVariant> // remove these if they exist
    StripToOneCompressionVariant bool // if there are multiple variants, only preserve one
    CompressionVariantPreferenceOrder []CompressionVariant // relevant for StripToOneCompressionVariant and for CopySystemImage
```

---

But right now, I’m not too tempted to overthink this. Note that the `GivePriorityToZstdImages` / `PreferGzip` is currently _primarily_ desired by @vrothberg just as a feature flag, to allow us to revert to the previous behavior, not because we expect users to _long_ term make these choices at all. Such a feature flag is inherently short-term and not worth _too_ much design, apart from the fact that we will need to keep it around forever due to API stability.

OTOH it’s not _quite_ short-term: we do need to give users some kind of choice for `CopySystemImage` to destinations other than local c/storage … or, um, maybe to recommend against using `CopySystemImage` in such situations completely.

manifest/oci_index.go
flouthoc: @mtrmac I have modified code so no unnecessary loop is added and original code is kept intact, I like the idea of adding comparator but while I was implementing that I saw it requires us to add additional sorting infra while our feature set of `ChooseInstance` and `copy.Options` does not have choices it only supports `zstd` or `gzip` so code for doing priority selection looked early optimization with additional sorting infrastructure, hence I refrained from adding that and was thinking to add that complexity when one more choice arise. 

But anyways I am okay with whatever you prefer so could you please take a look a new push and let me know ?

mtrmac: We don’t really need _sorting_, `min` is just:
```go
var best *Manifest
for _, m := range manifests {
    if /* manifest is acceptable at all */ && (best == nil || instanceBetterMatchesPlatforms(manifest, best) {
        best = manifest
    }
}
```

mtrmac: … I’ll take a more detailed look in a few hours.

copy/copy.go
```diff
@@ -180,6 +180,10 @@ type Options struct {
 	// Download layer contents with "nondistributable" media types ("foreign" layers) and translate the layer media type
 	// to not indicate "nondistributable".
 	DownloadForeignLayers bool
+
+	// Give priority to pulling zstd images if multiple images are present unless explicitly
+	// configured true.
+	PreferGzip types.OptionalBool
```
mtrmac, line 186: - Conceptually this belongs in the image selection block, maybe after `Instances`
- The documentation must say that this affects only `CopySystemImage`
- … and, I think, that we
  - prefer gzip (not zstd, as the comment says now) images if this is set to `OptionalBoolTrue`
  - prefer the best compression supported by the system (which may change over time) if this set to `OptionalBoolFalse`
  - choose automatically (and the choice may change over time) if this is set to `OptionalBoolUndefined` (which is the default behavior, and recommended for most callers).

Cc: @vrothberg

internal/manifest/oci_index.go
```diff
@@ -93,19 +100,49 @@ func (index *OCI1IndexPublic) ChooseInstance(ctx *types.SystemContext) (digest.D
...
 			}
 			if platform.MatchesPlatform(imagePlatform, wantedPlatform) {
-				return d.Digest, nil
+				if matchedDigest == "" {
+					matchedDigest = d.Digest
+				}
+				if compressionAnnotation == "" {
+					return d.Digest, nil
+				} else {
+					if _, ok := d.Annotations[compressionAnnotation]; ok {
```
mtrmac, line 109: Hum, should this enforce that the value is `"true"`, and fail on other values?

I think it should, we have _one_ chance to enforce all creators to use that syntax correctly. We can always relax this later.

OTOH “be liberal in what you accept” is a long-standing idea with some attraction to it, so I don’t want to _completely_ dismiss it. What do others think?

rhatdan: I agree lets be strict in the beginning until we are forced to loosen.

internal/manifest/oci_index.go
```diff
@@ -93,19 +100,49 @@ func (index *OCI1IndexPublic) ChooseInstance(ctx *types.SystemContext) (digest.D
...
 		}
 	}
 	return "", fmt.Errorf("no image found in image index for architecture %s, variant %q, OS %s", wantedPlatforms[0].Architecture, wantedPlatforms[0].Variant, wantedPlatforms[0].OS)
 }
 
+// ChooseInstanceByCompression parses blob as an oci v1 manifest index, and returns the digest
+// of the image which is appropriate for the current environment but if multiple digests matches
+// for a given platform it gives priority to the one which is configured in policy, as of now
+// only two options are allowed either prefer gzip or zstd.
+func (index *OCI1IndexPublic) ChooseInstanceByCompression(ctx *types.SystemContext, preferGzip types.OptionalBool) (digest.Digest, error) {
```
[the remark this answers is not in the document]
flouthoc: Schema2List does not actually do anything so it just calls `ChooseInstance(`

internal/manifest/oci_index.go
```diff
@@ -93,19 +100,49 @@ func (index *OCI1IndexPublic) ChooseInstance(ctx *types.SystemContext) (digest.D
...
+				}
 			}
 		}
 	}
-
-	for _, d := range index.Manifests {
+	if matchedDigest != "" {
+		return matchedDigest, nil
+	}
+	for _, d := range manifests {
```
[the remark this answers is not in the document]
flouthoc: I have created a new function specific to this task.

internal/manifest/oci_index_test.go
```diff
@@ -45,3 +49,101 @@ func TestOCI1IndexFromManifest(t *testing.T) {
...
+		for _, match := range manifestList.matchedInstances {
+			testName := fmt.Sprintf("%s %q+%q", manifestList.listFile, match.arch, match.variant)
+			digest, err := list.ChooseInstanceByCompression(&types.SystemContext{
+				ArchitectureChoice: match.arch,
+				VariantChoice:      match.variant,
+				OSChoice:           "linux",
+			}, types.NewOptionalBool(false))
+			require.NoError(t, err, testName)
+			assert.Equal(t, match.instanceDigest, digest, testName)
+		}
```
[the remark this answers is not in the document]
flouthoc: Added a test for this.

internal/manifest/list_test.go
```diff
@@ -112,7 +112,7 @@ func TestChooseInstance(t *testing.T) {
 			},
 		},
 	} {
-		rawManifest, err := os.ReadFile(filepath.Join("..", "..", "internal", "image", "fixtures", manifestList.listFile))
+		rawManifest, err := os.ReadFile(filepath.Join("..", "..", "internal", "manifest", "testdata", manifestList.listFile))
```
mtrmac, line 115: This is just `filepath.Join("testdata", manifestList.listFile)`, isn’t it?

internal/manifest/list_test.go
```diff
@@ -137,3 +137,96 @@ func TestChooseInstance(t *testing.T) {
 		}
 	}
 }
+
+func TestChooseInstanceByCompressionAndListInternalFromBlob(t *testing.T) {
```
mtrmac, line 141: All of this is just copied `TestChooseInstance` to ensure that `ChooseInstanceByCompression` behaves the same, and doesn’t depend on the compression preference, isn’t it?

If so, please don‘t duplicate such a large piece of code; just loop inside the existing `TestChooseInstance`:
```go
chooseInstanceCalls := []func(sys *types.SystemContext, rawManifest []byte) (digest.Digest, error) {
   func(sys *types.SystemContext, rawManifest []byte) (digest.Digest, error) {
      list, err := ListPublicFromBlob(rawManifest, GuessMIMEType(rawManifest))
      …
      return list.ChooseInstance(sys)
   },
   func /* use ChooseInstanceByCompression … OptionalBoolTrue */,
   func /* use ChooseInstanceByCompression … OptionalBoolFalse */,
   func /* use ChooseInstanceByCompression … OptionalBoolUndefined, adding that is now cheap */,
}
…
for _, manifestList := range []struct {…} {
		for _, match := range manifestList.matchedInstances {
		    for callIndex, chooseInstance := range chooseInstanceCalls {
		        testName := …
		        digest, err = chooseInstance(&systemContext{…}, rawManifest
```

internal/manifest/oci_index_test.go
```diff
@@ -45,3 +49,113 @@ func TestOCI1IndexFromManifest(t *testing.T) {
 	// Extra fields are rejected
 	testValidManifestWithExtraFieldsIsRejected(t, parser, validManifest, []string{"config", "fsLayers", "history", "layers"})
 }
+
+func TestOCI1IndexChooseInstanceByCompressionAndListInternalFromBlob(t *testing.T) {
```
mtrmac, line 53: `TestOCI1IndexChooseInstanceByCompression`, please; it’s a unit test targeting that method.

internal/manifest/oci_index_test.go
```diff
@@ -45,3 +49,113 @@ func TestOCI1IndexFromManifest(t *testing.T) {
...
+			matchedInstances: []expectedMatch{
+				{"amd64", "", "sha256:5b0bcabd1ed22e9fb1310cf6c2dec7cdef19f0ad69efa1f392e94a4333501270"},
+				{"ppc64le", "", "sha256:e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7fc331f"},
+			},
+			unmatchedInstances: []string{
+				"unmatched",
+			},
+		},
+	} {
+		rawManifest, err := os.ReadFile(filepath.Join("..", "..", "internal", "manifest", "testdata", manifestList.listFile))
```
mtrmac, line 126: This is just `Join("testdata", …)`, isn’t it?

internal/manifest/oci_index_test.go
```diff
@@ -45,3 +49,113 @@ func TestOCI1IndexFromManifest(t *testing.T) {
...
+		// Not found and Gzip preference false
+		for _, arch := range manifestList.unmatchedInstances {
+			_, err := list.ChooseInstanceByCompression(&types.SystemContext{
+				ArchitectureChoice: arch,
+				OSChoice:           "linux",
+			}, types.NewOptionalBool(false))
+			assert.Error(t, err)
+		}
+		// Match found and Gzip preference true
+		for _, match := range manifestList.prefferedGzipInstances {
```
mtrmac, line 150: Make the `preferGzip` input a new field in `expectedMatch` (https://github.com/containers/image/pull/1789#discussion_r1101831378); then this loop, and `preferredGzipInstances`, don’t need to exist, _and_ we can also, possibly, test `OptionalBoolUndefined` more easily.

internal/manifest/oci_index.go
```diff
@@ -93,19 +115,54 @@ func (index *OCI1IndexPublic) ChooseInstance(ctx *types.SystemContext) (digest.D
...
+}
+
+func (index *OCI1Index) ChooseInstanceByCompression(ctx *types.SystemContext, preferGzip types.OptionalBool) (digest.Digest, error) {
+	return index.chooseInstance(ctx, preferGzip)
+}
+
+// ChooseInstance parses blob as an oci v1 manifest index, and returns the digest
+// of the image which is appropriate for the current environment.
+func (index *OCI1IndexPublic) ChooseInstance(ctx *types.SystemContext) (digest.Digest, error) {
+	return index.chooseInstance(ctx, types.NewOptionalBool(false))
```
mtrmac, line 163: Everywhere: instead of `types.NewOptionalBool(constant)`, we can just use `types.OptionalBoolConstant`.

internal/manifest/testdata/oci1.index.zstd-selection.json
```diff
@@ -0,0 +1,66 @@
...
+            "digest": "sha256:88cb27f194cce3b06994b4af2dd8a5642f4f294b71bda5b74a3fbb74z9522511",
+            "size": 758,
+            "platform": {
+                "architecture": "s390x",
+                "os": "linux"
+            }
+        },
+        {
+            "mediaType": "application/vnd.oci.image.manifest.v1+json",
+            "digest": "sha256:ac9bbf71dacfad2bfbe48m7d46ca279b1b3b351a6c814db760030015167405b8",
```
mtrmac, line 55: `m`, and `z` elsewhere in this file, is not a valid hexadecimal digit; that makes it long-term uncertain whether the unit test is testing what we want it to.

---

Having digests that only differ by one latter somewhere in the middle is _seriously_ unhelpful for casual readers of the test. “Oh, the first 8 or so characters match, it looks random, so this looks like the one I’m looking for”.

Please either use completely random values that visually differ, or something _explicitly_ structured like `sha256:aaaa…a` and `sha256:bbb…b`

internal/manifest/testdata/oci1.index.zstd-selection.json
```diff
@@ -0,0 +1,66 @@
...
+            "digest": "sha256:88cb27f194cce3b06994b4af2dd8a5642f4f294b71bda5b74a3fbb74z9522511",
+            "size": 758,
+            "platform": {
+                "architecture": "arm64",
+                "os": "linux"
+            }
+        },
+        {
+            "mediaType": "application/vnd.oci.image.manifest.v1+json",
+            "digest": "sha256:88cb27f194cce3b06994b4af2dd8a5642f4f294b71bda5b74a3fbb74z9522511",
```
mtrmac, line 46: These two digests are _exactly_ the same.

1. That’s unrealistic, and
2. Unless this is explicitly testing for duplicate digest handling (which is not relevant to our code), this makes for an unhelpful test case because we don’t know which one matched.

internal/manifest/oci_index_test.go
```diff
@@ -45,3 +49,113 @@ func TestOCI1IndexFromManifest(t *testing.T) {
...
+				{"arm64", "", "sha256:88cb27f194cce3b06994b4af2dd8a5642f4f294b71bda5b74a3fbb74a9522511"},
+				// select a signle gzip s390x image
+				{"s390x", "", "sha256:88cb27f194cce3b06994b4af2dd8a5642f4f294b71bda5b74a3fbb74z9522511"},
+			},
+			unmatchedInstances: []string{
+				"unmatched",
+			},
+			// When preffredGzip: true
+			prefferedGzipInstances: []expectedMatch{
+				// out of gzip and zstd in amd64 select the first zstd image
```
mtrmac, line 79: Incorrect comment

internal/manifest/oci_index_test.go
```diff
@@ -45,3 +49,113 @@ func TestOCI1IndexFromManifest(t *testing.T) {
...
+				{"arm", "v7", "sha256:f365626a556e58189fc21d099fc64603db0f440bff07f77c740989515c544a39"},
+				{"arm", "v6", "sha256:f365626a556e58189fc21d099fc64603db0f440bff07f77c740989515c544a39"},
+				{"arm", "v5", "sha256:c84b0a3a07b628bc4d62e5047d0f8dff80f7c00979e1e28a821a033ecda8fe53"},
+				{"arm", "", "sha256:c84b0a3a07b628bc4d62e5047d0f8dff80f7c00979e1e28a821a033ecda8fe53"},
+				{"arm", "unrecognized-present", "sha256:bcf9771c0b505e68c65440474179592ffdfa98790eb54ffbf129969c5e429990"},
+				{"arm", "unrecognized-not-present", "sha256:c84b0a3a07b628bc4d62e5047d0f8dff80f7c00979e1e28a821a033ecda8fe53"},
+			},
+			unmatchedInstances: []string{
+				"unmatched",
+			},
```
mtrmac, line 113: This is not relevant to the function under test at all.

mtrmac (commented):
Ran out of time today…

internal/manifest/oci_index_test.go
mtrmac: … but what _would_ be relevant is doing a similar variant preference test against OCI, and making sure that it interacts _correctly_ with the compression preference — i.e. that the variant preference is more important.

flouthoc:
Not a complete push but push only for https://github.com/containers/image/pull/1789#discussion_r1105132442.

internal/manifest/oci_index.go
```diff
@@ -93,19 +147,51 @@ func (index *OCI1IndexPublic) ChooseInstance(ctx *types.SystemContext) (digest.D
 				Variant:      d.Platform.Variant,
 			}
 			if platform.MatchesPlatform(imagePlatform, wantedPlatform) {
-				return d.Digest, nil
+				if bestMatchedManifest.Digest == "" || instanceBetterMatchesCompression(d, bestMatchedManifest, didPreferGzip) {
```
mtrmac, line 150: Surprisingly enough, AFAICS nothing here enforces that entries of `manifests` all have `.Digest != ""` (and this function is probably not the place to do that). _Conceptually_, this condition therefore does not quite express the intent. A pedantically more correct alternative would be to have `bestMatch *….Descriptor`, with `nil` as the “no match yet” case.

OTOH, sure, an item with `Digest == ""` would be unusable… and anyway I this is, at best, an item to deal with the last tune-up.

---

(The `bestMatchedManifest.Digest == ""` only needs to be in _one_ of `instanceBetterMatchesCompression` and its caller. I think ideally it should be in the caller, but that makes more sense to make if there is only one call site.)

internal/manifest/oci_index.go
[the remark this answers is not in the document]
mtrmac: Works for me.

internal/manifest/list_test.go
mtrmac: I’d still advocate for this loop structure — or even shorter
```go
…
		for callIndex, chooseInstance := range chooseInstanceCalls {
			for _, match := range manifestList.matchedInstances {
			for _, arch := range manifestList.unmatchedInstances {
```
so that there is only one `chooseInstanceCalls` loop.

In my experiment it is shorter (-16 lines), it has more coverage (`OptionalBoolUndefined`), and it is much clearer about how the repeated instances of the code do / don’t differ, making it less likely that future updates will only change one of the sites.

I guess non-blocking.

mtrmac (commented):
So the test suggested in https://github.com/containers/image/pull/1789#discussion_r1105187412 really is necessary, and it currently fails; because the current search loop accounts for the three relevant variables differently:
- platform preference, by loop order
- compression preference, by explicit code
- manifest order preference, by loop order

and it can’t do two different choices from the same information.

---

Prototyping this, moving to an explicit predicate really results in, dare I say, quite straightforward code that _very directly_ captures the rules we have chosen:
```go
type instanceCandidate struct {
	platformIndex    int // Index of the candidate in platform.WantedPlatforms: lower numbers are preferred; or math.maxInt if the candidate doesn’t have a platform
	isZstd           bool // or some other expression of the compression, maybe just the annotations hash
	manifestPosition int // A zero-based index of the instance in the manifest list
}

func (ic instanceCandidate) isPreferredOver(other instanceCandidate, preferGzip bool) bool {
	switch {
	case ic.platformIndex != other.platformIndex:
		return ic.platformIndex < other.platformIndex
	case ic.isZstd != other.isZstd:
		if !preferGzip {
			return ic.isZstd
		} else {
			return !ic.isZstd
		}
		// Or use a compressionScore as suggested elsewhere, for even more regular structure
	case ic.manifestPosition != other.manifestPosition:
		return ic.manifestPosition < other.manifestPosition
	}
	panic("return an error instead") // return internal error: this should not be reachable
}
```

`chooseInstance` then needs to to create this `instanceCandidate`, primarily by determining `isZstd` and by finding the right `platformIndex`; it can do that itself, or in small helper functions.

---

And _ultimately_, all of this `instanceCandidate` logic could then end up in a separate `instance_choice.go` or something; then the format-specific `ChooseInstance` code would _only_ be responsible for turning the specific instance format into an `[]instanceCandidate`, and a generic code in `instance_choice.go` could make the actual choice, ensuring that we are consistent for all formats

That is quite likely overkill, because it would make too many assumptions that the multi-arch structure is structurally the same between formats (e.g. that the manifest order matters the same way). And the extra code to create `instanceCandidate` structures would not actually be shorter than the current `Schema2ListPublic.ChooseInstance`.

So, I think let’s not do this part now… just noting that that is a choice that opens up by making `instanceCandidate` completely format-independent.

---

(This is now a full review, including the tests.)

internal/manifest/oci_index_test.go
flouthoc: Added a test for this.

internal/manifest/list_test.go
flouthoc: Fixed.

internal/manifest/oci_index.go
```diff
@@ -73,15 +83,78 @@ func (index *OCI1IndexPublic) UpdateInstances(updates []ListUpdate) error {
...
+func instanceIsZstd(manifest imgspecv1.Descriptor) bool {
+	if value, ok := manifest.Annotations[OCI1InstanceAnnotationCompressionZSTD]; ok && value == "true" {
+		return true
+	}
+	return false
+}
+
+type instanceCandidate struct {
+	platformIndex    int  // Index of the candidate in platform.WantedPlatforms: lower numbers are preferred; or math.maxInt if the candidate doesn’t have a platform
+	isZstd           bool // or some other expression of the compression, maybe just the annotations hash
```
mtrmac, line 124: “or some other expression” is not a good documentation of the field. I guess just drop that comment.

internal/manifest/oci_index.go
```diff
@@ -73,15 +83,78 @@ func (index *OCI1IndexPublic) UpdateInstances(updates []ListUpdate) error {
...
+	switch {
+	case ic.platformIndex != other.platformIndex:
+		return ic.platformIndex < other.platformIndex
+	case ic.isZstd != other.isZstd:
+		if !preferGzip {
+			return ic.isZstd
+		} else {
+			return !ic.isZstd
+		}
+		// Or use a compressionScore as suggested elsewhere, for even more regular structure
```
mtrmac, line 138: Here, as well, either replace the comment with the implementation of that comment, or remove it: it will no be useful to anyone reading the code in the future without referring to this PR.

mtrmac (commented):
Thanks!

That’s broadly the correct behavior, and sufficient test coverage assuming a yet more simplified implementation.

And the pesky second loop still needs fixing — so let’s eliminate it instead.

flouthoc:
@mtrmac PTAL

internal/manifest/oci_index.go
```diff
@@ -73,15 +83,78 @@ func (index *OCI1IndexPublic) UpdateInstances(updates []ListUpdate) error {
...
+	case ic.manifestPosition != other.manifestPosition:
+		return ic.manifestPosition < other.manifestPosition
+	case ic.platformIndex == other.platformIndex && ic.manifestPosition == other.manifestPosition:
+		return false
+	}
+	panic("return an error instead") // return internal error: this should not be reachable
```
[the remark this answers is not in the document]
mtrmac: Non-blocking: It would be somewhat better to really return an error instead of using `panic` here — if this really ended up being reachable, it might turn into a registry-triggerable DoS.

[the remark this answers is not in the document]
mtrmac: … at the very least the comment is, right now, incorrect.

internal/manifest/oci_index.go
```diff
@@ -93,19 +166,56 @@ func (index *OCI1IndexPublic) ChooseInstance(ctx *types.SystemContext) (digest.D
...
+				}
 			}
 		}
 	}
+	if bestMatchedManifest.Digest != "" {
+		return bestMatchedManifest.Digest, nil
+	}
 
-	for _, d := range index.Manifests {
+	// process specifically for cases when platform == nil
```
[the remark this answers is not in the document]
mtrmac: Actually, test coverage for `d.Platform == ""` (both for processing the zstd/gzip preference correctly, and for correctly preferring a defined platform over no platform) would still be nice.

internal/manifest/oci_index.go
```diff
@@ -93,19 +136,38 @@ func (index *OCI1IndexPublic) ChooseInstance(ctx *types.SystemContext) (digest.D
 				Variant:      d.Platform.Variant,
 			}
 			if platform.MatchesPlatform(imagePlatform, wantedPlatform) {
-				return d.Digest, nil
+				if bestMatchedInstance.digest == "" || instanceCandidate.isPreferredOver(bestMatchedInstance, didPreferGzip) {
```
mtrmac, line 139: This is now unnecessarily called `len(manifests)*len(wantedPlatforms)` times.
- If `d.Platform == nil`, we don’t need to iterate through `wantedPlatforms` at all
- Otherwise, just search for the first match, and use that to set `instanceCandidate.platformIndex`; call `isPreferredOver` once per manifest.

Sure, the current implementation works and is not asymptotically any more costly.

Compare https://github.com/containers/image/pull/1789#discussion_r1113453162 suggesting that the inner loop could be a helper function; I don’t care right now about actually factoring that function out.

My primary goal is that the structure should cleanly separate the `platformIndex` determination from the rest of the logic, so that we can, in the future, replace `wantedPlatforms` with something that allows O(1) lookup.

mtrmac (commented):
Thanks! pretty good now.

Outstanding:
- https://github.com/containers/image/pull/1789#discussion_r1105135778 (at the very least because the `instanceMatchesPlatformAndCompression` function name is not correctly describing the operation at all)
- https://github.com/containers/image/pull/1789#discussion_r1113444610 

Non-blocking but preferred:
- Test coverage: https://github.com/containers/image/pull/1789#discussion_r1105132824 
- Test coverage of https://github.com/containers/image/pull/1789#discussion_r1113458096 
- https://github.com/containers/image/pull/1789#discussion_r1109330713 - one clean way to do this is to have `var bestMatchedInstance *instanceCandidate` and use `nil` as the sentinel instead of `.digest = ""`

Non-blocking:
- https://github.com/containers/image/pull/1789#discussion_r1105142584

internal/manifest/oci_index.go
mtrmac: Yes, that commit gets the core point right.

internal/manifest/oci_index_test.go
```diff
@@ -45,3 +49,114 @@ func TestOCI1IndexFromManifest(t *testing.T) {
...
+				{"amd64", "", "sha256:cccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc", false},
+				{"amd64", "", "sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa", true},
+				// must return first gzip even if the first entry is zstd
+				{"arm64", "", "sha256:qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqq", true},
+				// must return first zstd even if the first entry for same platform is gzip
+				{"arm64", "", "sha256:mmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm", false},
+				// must return first zstd instance agnostic of platform
+				{"", "", "sha256:cccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc", false},
+				// must return first gzip instance agnostic of platform
+				{"", "", "sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa", true},
```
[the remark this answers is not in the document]
flouthoc: done added a test for this.

internal/manifest/oci_index.go
```diff
@@ -73,39 +84,86 @@ func (index *OCI1IndexPublic) UpdateInstances(updates []ListUpdate) error {
...
-	}
-
-	for _, d := range index.Manifests {
-		if d.Platform == nil {
-			return d.Digest, nil
+		if bestMatchedInstance == nil || instanceCandidate.isPreferredOver(bestMatchedInstance, didPreferGzip) {
+			bestMatchedInstance = &instanceCandidate
 		}
 	}
+	if bestMatchedInstance != nil && bestMatchedInstance.platformIndex != math.MaxInt {
```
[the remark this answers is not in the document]
flouthoc: not needed, it was needed in the core logic not here.

internal/manifest/oci_index.go
```diff
@@ -73,39 +84,89 @@ func (index *OCI1IndexPublic) UpdateInstances(updates []ListUpdate) error {
...
-				Architecture: d.Platform.Architecture,
-				OS:           d.Platform.OS,
-				OSVersion:    d.Platform.OSVersion,
-				OSFeatures:   slices.Clone(d.Platform.OSFeatures),
-				Variant:      d.Platform.Variant,
-			}
-			if platform.MatchesPlatform(imagePlatform, wantedPlatform) {
-				return d.Digest, nil
+		}
+		if d.Platform == nil || candidate.platformIndex != math.MaxInt {
```
mtrmac, line 148: This works, but it was not obvious to me without your comment elsewhere alluding to this.

This is repetitive for the `d.Platform` case, and _just far enough_ from the `wantedPlatform` loop for me to miss the connection with that loop.

Maybe
```go
if d.Platform != nil {
  for … { … }
  if candidate.platformIndex == math.MaxInt { // No platform match found
    continue
  }
}
```
or all the way to full formal tracking:
```go
if d.Platform != nil {
  foundPlatform := false 
  for … { if … { foundPlatform = true; candidate.platformIndex = …; break } }
  if !foundPlatform {
    continue
  }
}
```
in both cases, I think this `if` statement can be removed afterwards.

internal/manifest/oci_index_test.go
```diff
@@ -45,3 +49,126 @@ func TestOCI1IndexFromManifest(t *testing.T) {
...
+			testName := fmt.Sprintf("%s %q+%q", manifestList.listFile, match.arch, match.variant)
+			digest, err := list.ChooseInstanceByCompression(&types.SystemContext{
+				ArchitectureChoice: match.arch,
+				VariantChoice:      match.variant,
+				OSChoice:           "linux",
+			}, types.NewOptionalBool(match.preferGzip))
+			require.NoError(t, err, testName)
+			assert.Equal(t, match.instanceDigest, digest, testName)
+		}
+		// Not found and Gzip preference false
```
mtrmac, line 161: The two comments about Gzip preference are inapplicable now that that is driven by `match.preferGzip`.

internal/manifest/oci_index_test.go
```diff
@@ -45,3 +49,126 @@ func TestOCI1IndexFromManifest(t *testing.T) {
...
+		for _, instance := range manifestList.unmatchedInstances {
+			digest, err := list.ChooseInstanceByCompression(&types.SystemContext{
+				ArchitectureChoice: instance.arch,
+				OSChoice:           "linux",
+			}, types.NewOptionalBool(instance.preferGzip))
+			if instance.instanceDigest != "" {
```
mtrmac, line 167: This new `instanceDigest != ""` case is AFAICS structurally equal to using a `matchedInstances` entry; so those new tests for `d.Platform == ""` can, I think, be expressed using a `matchedInstances` entry, and `unmatchedInstances` can remain the way it was.

(`unmatchedInstances` is literally meant to mean “these are not expected to match and to be found”, so having them match and return a value is contradictory to the name. `unmatchedInstances` does not mean “the platform names are unknown”, we are using the `unmatched` text only to be _very clear_ to the reader that this is not going to match anything.

So, also, for the new tests that _are_ expected to find something, it would be more descriptive not to name them `unmatched`; something like `matchesImageWithNoPlatform` (or a shorter version of that) would tell the reader what is going on.)

internal/manifest/oci_index.go
```diff
@@ -73,39 +84,94 @@ func (index *OCI1IndexPublic) UpdateInstances(updates []ListUpdate) error {
...
-				OSVersion:    d.Platform.OSVersion,
-				OSFeatures:   slices.Clone(d.Platform.OSFeatures),
-				Variant:      d.Platform.Variant,
+			if !foundPlatform {
+				continue
 			}
-			if platform.MatchesPlatform(imagePlatform, wantedPlatform) {
-				return d.Digest, nil
+		}
+		if d.Platform == nil || candidate.platformIndex != math.MaxInt {
```
mtrmac, line 153: The `foundPlatform` change was meant to make this unnecessary, so please drop it.

mtrmac (commented):
One last line I think

internal/manifest/oci_index.go
flouthoc: whoops i missed this.

mtrmac: The whole condition needs removing - otherwise images with `Platform` defined would not be found. (And tests find that, which is great.)

mtrmac (approved):
LGTM, assuming tests pass.

Thanks!

flouthoc:
@giuseppe @rhatdan @vrothberg @TomSweeneyRedHat PTAL

vrothberg (approved):
LGTM, nice work!

PR #1791, opened 2023-01-17
manifest: introduce `internal/manifest` with private types and freeze public `manifest.List`

flouthoc:
Flips dependency of internal and public manifest.List API and introduces private manifest.List API which will be extended in future for new features.


flouthoc:
Upon discussion with @mtrmac  I think this is something which he was suggesting. But @mtrmac could you PTAL and confirm if this is on the right track. Thanks

internal/manifest/list.go
```diff
@@ -0,0 +1,24 @@
...
+}
+
+// ListFromBlob parses a list of manifests.
+func ListFromBlob(manifest []byte, manifestMIMEType string) (List, error) {
+	return publicManifest.ListFromBlob(manifest, manifestMIMEType)
+}
+
+// ConvertListToMIMEType converts the passed-in manifest list to a manifest
+// list of the specified type.
+func ConvertListToMIMEType(list List, manifestMIMEType string) (List, error) {
```
mtrmac, line 22: There are no callers of this trivial wrapper, so I think that can stay in the public package.

manifest/errors.go
```diff
@@ -0,0 +1,32 @@
...
+	// to complain about a non-image artifact with the correct MIME type; we should probably add some other kind of
+	// type discrimination, _and_ somehow make it available in the API, if we expect API callers to make decisions
+	// based on that kind of data.
+	//
+	// So, let’s not expose this until a specific need is identified.
+	mimeType string
+}
+
+// NewNonImageArtifactError returns a NonImageArtifactError about an artifact with mimeType.
+func NewNonImageArtifactError(mimeType string) error {
```
mtrmac, line 26: Really the whole point of the pre-existing `internal/manifest` package was to make this function internal-only (so that we can potentially add more fields, and we don’t need to worry about compatibility with external callers who don’t supply those), so moving this function to the public package is problematic.

mtrmac (commented):
Could you prototype this a bit further, please?

In a new commit, add a do-nothing method to `internal/manifest.List`, and call it from `copy.Image`. Build the necessary infrastructure to do that; ensure that the new method can not be called via `c/image/manifest.ListFromBlob`; and then remove that do-nothing method.

In particular, I expect that will require at least creating `internal/manifest.OCI1Index` and `internal/manifest.Schema2List`; and that, in turn, might require moving quite a bit more types/constants/variables from `manifest` to `internal/manifest`.

The goal of this is to do _all_ of the refactoring in this PR, so that the Zstd PR can just add a method without any more refactoring.

---

I guess the `NonImageArtifactError` type might need to be moved into a separate subpackage. Or, alternatively, move almost all of the implementation of `c/image/manifest` to `internal/manifest, so that `internal/manifest` has never a reason to call the public one.

mtrmac:
> ensure that the new method can not be called via `c/image/manifest.ListFromBlob`

On second thought — callers that _explicitly cast_ are not interesting; those deserve any breakage they get. But the new method should not be directly visible in the API (i.e. it should not be a public method on a publicly visible `struct`).

flouthoc:
> I guess the NonImageArtifactError type might need to be moved into a separate subpackage. Or, alternatively, move almost all of the implementation of c/image/manifest to internal/manifest, so that internal/manifest` has never a reason to call the public one.

@mtrmac Do you have any priority on these two suggested solution ? Like which one do you think could be better so I can do that part. IMO maybe moving all of the `manifest` implementation to `internal/manifest` is easier but it creates redundancy in codebase which we will have to slowly reduce.

mtrmac:
> but it creates redundancy in codebase which we will have to slowly remove/reduce.

I was thinking that the public `manifest` package would not contain a duplicate implementation, just forwarding stubs, similarly to c/image/image.
d
---

My hunch is to try moving the `NonImageArtifactError` code to a separate subpackage; at a first glance it _seems_ quite possible to move only the multi-arch code (implementations of `manifest.List`) without substantially affecting the single-instance code (s1/s2/OCI); that way, we wouldn’t need to add all the forwarding stubs now. It’s quite possible I have overlooked something and moving all of the single-instance code would be necessary as well.

flouthoc:
Wrapping of structs is breaking the calling pattern for public API, I wonder if  redundancy of structs is the only option ? https://go.dev/play/p/1Q_4X9hNB5j

flouthoc:
> In a new commit, add a do-nothing method to internal/manifest.List, and call it from copy.Image.

@mtrmac PTAL I think this is what you meant, I have also added a `DoNothing()` to private implementation in a new commit for proof-of-concept. If this looks good then I will drop the `DoNothing` commit.

mtrmac:
(This is _not_ a review nor a request for any change, just a note about Go language features at this point.)

> Wrapping of structs is breaking the calling pattern for public API, I wonder if  redundancy of structs is the only option ?

```go
type OCI1 = internalPackage.OCI1Public.
```

Compare e.g. `pkg/compression/types.Algorithm`.

mtrmac:
Thinking back to the goals of this infrastructure:
- Primary: we _must not_ change the public `manifest.List` interface, or other interfaces visible in `manifest`. This PR works fine for that.
- Not 100% necessary but desirable: minimize duplication between the public and internal `manifest` package, in particular to decrease the cost of maintenance.
- Not 100% necessary bu very desirable: Don’t expose the new APIs we need, at least while they are being developed for the full Zstd variants feature and still being shaped, as public API in `manifest` that we need to keep stable.

This PR works fine for the first point; but looking at the second two
- the existing public `ChooseInstance` methods, and the new methods with more options, should share code if at all possible (with the existing methods probably just calling to the new more capable implementation)
- that shared code should not live in the public `manifest` package

That’s a bit awkward to do with the current internal → public dependency direction. It would require moving the actual implementation logic into a third `$implementation` subpackage, so that the public package does not have new functions, but so that the actual implementation can still be called from `manifest`. And to move the actual implementation to `$implementation`, we would have to also move the data type to `$implementation`. (That would not be _strictly_ necessary for `OCI1`, with a third-party package that contains the actual data definitions, but it _would_ be necessary for schema2list).

The net effect of that would be `internal/manifest` (basically trivial wrappers) → `manifest` (trivial wrappers) → `$implementation`…

… so I think it would actually be _much_ simpler to reverse the dependency:
- Make `manifest` depend on `internal/manifest`. No `$implementation` package is necessary.`
- Move all of the actual implementation from `manifest` to `internal/manifest` (`manifest.OCI1Index = internal/manifest.OCI1IndexPublic`)
- Let `internal/manifest` provide `internal/manifest.OCI1` that contains an `internal/manifest.OCI1IndexPublic` member
- Existing public `OCI1IndexPublic` methods can stay public methods.
- Other shared implementations would end up as separate public functions. In particular,
  -  A new `internal/manifest.OCI1ChooseInstance(internal/manifest.OCI1Public` function would be created contain the actual implementation
  - `manifest/OCI1.ChooseInstance` = `internal/manifest.OCI1Public.ChooseInstance` just calls `OCI1ChooseInstance` with the right parameters
  - A future new `internal/manifest/OCI1.ChooseInstance[TheNewOne]` is a trivial forwarder to `internal/manifest.OCI1ChooseInstance`.

Does that make sense? I think this would work and do everything necessary, but I didn’t actually do the work so I’m not 100% sure there isn‘t a snag.

mtrmac:
>   * A new `internal/manifest.OCI1ChooseInstance(internal/manifest.OCI1Public` function would be created contain the actual implementation

Actually that can just be a  package-private `OCI1Public.chooseInstance` method.

flouthoc:
@mtrmac PTAL I agree moving everything to internal is much better. Also I have kept naming scheme as `OCI1Index` and `OCI1IndexInternal` for consistency with other packages and it automatically explains that the one with suffix `Internal` is the private type while other is public.

Also added a dummy commit to show how private API can be extended.

mtrmac (commented):
At this point let’s just discuss it, please don’t immediately make changes:

---

Is there a dependency that forces us to move the non-multi arch implementations as well?

---

> Also I have kept naming scheme as `OCI1Index` and `OCI1IndexInternal` for consistency with other packages

We don’t really have that precedent AFAICS.

> and it automatically explains that the one with suffix `Internal` is the private type while other is public.

The way I think about this is, _most_ of the c/image code is going to use the internal API so saying “internal” all over the place is not particularly useful. Pointing out public parts of an otherwise-nonpublic API might be a thing to pay attention to.

copy/copy.go
```diff
@@ -420,7 +422,7 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
 	if err != nil {
 		return nil, fmt.Errorf("reading manifest list: %w", err)
 	}
-	originalList, err := manifest.ListFromBlob(manifestList, manifestType)
+	originalList, err := internalManifest.ListInternalFromBlob(manifestList, manifestType)
 	if err != nil {
 		return nil, fmt.Errorf("parsing manifest list %q: %w", string(manifestList), err)
 	}
```
mtrmac, line 428: Just below, `updatedList.DoNothing()` will need to work for us to be able to add/remove Zstd variants.

mtrmac (commented):
Toying with https://github.com/mtrmac/image/tree/list-experiment (which is by no means complete), it does seem that we don’t actually _need_ to move the non-multi-arch types. The only real dependency is a `FromBlob` in `list_test.go`, which can be avoided by keeping the original test in `manifest`, and dropping that one test part from a copy in `internal/manifest`.

So, yes, I’d prefer minimizing the churn and dealing with only the two multi-arch objects.

flouthoc:
@mtrmac I was playing with your branch it correctly moves the two needed types and leaving everything else intact, what do else is left which you think is missing from that and needs to be done.

internal/manifest/docker_schema2.go
```diff
@@ -0,0 +1,13 @@
+package manifest
+
+import (
+	"github.com/opencontainers/go-digest"
+)
+
+// Schema2Descriptor is a “descriptor” in docker/distribution schema 2.
```
mtrmac, line 7:
```suggestion
// Schema2Descriptor is a “descriptor” in docker/distribution schema 2.
//
// This is publicly visible as c/image/manifest.Schema2Descriptor.
```

… and similar comments to the other types and public functions that are referenced from c/image/manifest.

---

All edits to `internal/manifest/docker_schema2.go` should apply to `internal/manifest/oci_index.go` analogously.

internal/manifest/fixtures_info_test.go
```diff
@@ -0,0 +1,12 @@
+package manifest
```
mtrmac, line 1: Non-blocking: Now that the files are in the Go-supported `testdata` subdirectory, this could probably be renamed to `testdata_info_test.go`.

manifest/docker_schema2_list_test.go
```diff
@@ -1,28 +0,0 @@
-package manifest
```
mtrmac, line 1: Let’s preserve this, to have _something_ to exercise the public interface. The more we move our own callers to the internal-only API, the more useful this is, even if it is a small smoke test.

Same for `oci_index_test.go`.

internal/manifest/list.go
```diff
@@ -0,0 +1,80 @@
...
+	Serialize() ([]byte, error)
+
+	// ConvertToMIMEType returns the list rebuilt to the specified MIME type, or an error.
+	ConvertToMIMEType(mimeType string) (ListPublic, error)
+
+	// Clone returns a deep copy of this list and its contents.
+	Clone() ListPublic
+}
+
+type List interface {
```
mtrmac, line 46:
```suggestion
// List is an interface for parsing, modifying lists of image manifests.
// Callers can either use this abstract interface without understanding the details of the formats,
// or instantiate a specific implementation (e.g. manifest.OCI1Index) and access the public members
// directly.
type List interface {
```

internal/manifest/list.go
```diff
@@ -0,0 +1,80 @@
...
+	ListPublic
+	CloneInternal() List
```
mtrmac, line 48:
```suggestion
	// CloneInternal returns a deep copy of this list and its contents.
	CloneInternal() List
```

\+ copy&paste this to the 2 implementations.

internal/manifest/docker_schema2_list.go
```diff
@@ -0,0 +1,249 @@
...
+	Features     []string `json:"features,omitempty"` // removed in OCI
+}
+
+// Schema2ManifestDescriptor references a platform-specific manifest.
+type Schema2ManifestDescriptor struct {
+	Schema2Descriptor
+	Platform Schema2PlatformSpec `json:"platform"`
+}
+
+// Schema2ListPublic is a list of platform-specific manifests.
```
mtrmac, line 30:
```suggestion
// Schema2ListPublic is a list of platform-specific manifests.
//
// This is publicly visible …
// Internal users should usually use Schema2List instead.
```

internal/manifest/docker_schema2_list.go
```diff
@@ -0,0 +1,249 @@
...
+		return list.ToOCI1Index()
+	case DockerV2Schema1MediaType, DockerV2Schema1SignedMediaType, imgspecv1.MediaTypeImageManifest, DockerV2Schema2MediaType:
+		return nil, fmt.Errorf("Can not convert manifest list to MIME type %q, which is not a list type", manifestMIMEType)
+	default:
+		// Note that this may not be reachable, NormalizedMIMEType has a default for unknown values.
+		return nil, fmt.Errorf("Unimplemented manifest list MIME type %s", manifestMIMEType)
+	}
+}
+
+type Schema2List struct {
```
mtrmac, line 222:
```suggestion
// Schema2List is a list of platform-specific manifests.
type Schema2List struct {
```

internal/manifest/list.go
```diff
@@ -0,0 +1,80 @@
...
+
+	"github.com/containers/image/v5/types"
+	digest "github.com/opencontainers/go-digest"
+	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+// ListPublic is an interface for parsing, modifying lists of image manifests.
+// Callers can either use this abstract interface without understanding the details of the formats,
+// or instantiate a specific implementation (e.g. manifest.OCI1Index) and access the public members
+// directly.
```
mtrmac, line 14:
```suggestion
// ListPublic is a subset of List which is a part of the public API;
// so no methods can be added, removed or changed.
//
// Internal users should usually use List instead.
```

mtrmac (commented):
It seems to me that the structure works overall … so just a final cleanup:

- Explicitly note, everywhere, which symbols in `internal/manifest` are publicly visible
- Add individual documentation to the new symbols, and update documentation of the old symbols
- Keep the existing tests in `c/image/manifest`

flouthoc:
@mtrmac PTAL again, I'll drop `DoNothing(` commit once PR is approved.

internal/manifest/manifest.go
```diff
@@ -0,0 +1,163 @@
...
+// GuessMIMEType guesses MIME type of a manifest and returns it _if it is recognized_, or "" if unknown or unrecognized.
+// FIXME? We should, in general, prefer out-of-band MIME type instead of blindly parsing the manifest,
+// but we may not have such metadata available (e.g. when the manifest is a local file).
+func GuessMIMEType(manifest []byte) string {
+	// A subset of manifest fields; the rest is silently ignored by json.Unmarshal.
+	// Also docker/distribution/manifest.Versioned.
+	meta := struct {
+		MediaType     string      `json:"mediaType"`
+		SchemaVersion int         `json:"schemaVersion"`
+		Signatures    interface{} `json:"signatures"`
```
mtrmac, line 44: This seems to have lost a recent `s/interface{}/any/g` during conflict resolution.

internal/manifest/oci_index.go
```diff
@@ -0,0 +1,263 @@
...
+func (index *OCI1Index) CloneInternal() List {
+	return oci1IndexFromPublic(OCI1IndexPublicClone(&index.OCI1IndexPublic))
+}
+
+func (index *OCI1Index) Clone() ListPublic {
+	return index.CloneInternal()
+}
+
+// OCI1IndexFromManifest creates a Schema2 manifest list instance from marshalled
+// JSON, presumably generated by encoding a Schema2 manifest list.
```
mtrmac, line 256: `s/Schema2/OCI1/g`, my mistake.

internal/manifest/common.go
```diff
@@ -0,0 +1,96 @@
...
+func dupStringStringMap(m map[string]string) map[string]string {
+	if len(m) == 0 {
+		return nil
+	}
+	result := make(map[string]string)
+	for k, v := range m {
+		result[k] = v
+	}
+	return result
+}
```
mtrmac, line 30: These disappeared on `upstream/main` recently; a mistake in conflict resolution

internal/manifest/docker_schema2_list.go
```diff
@@ -0,0 +1,252 @@
...
+	if err != nil {
+		return "", fmt.Errorf("getting platform information %#v: %w", ctx, err)
+	}
+	for _, wantedPlatform := range wantedPlatforms {
+		for _, d := range list.Manifests {
+			imagePlatform := imgspecv1.Platform{
+				Architecture: d.Platform.Architecture,
+				OS:           d.Platform.OS,
+				OSVersion:    d.Platform.OSVersion,
+				OSFeatures:   dupStringSlice(d.Platform.OSFeatures),
```
mtrmac, line 103: This is missing a recent upstream/main change.

internal/manifest/oci_index.go
```diff
@@ -0,0 +1,263 @@
...
+				Variant:      component.Platform.Variant,
+			}
+		}
+		m := imgspecv1.Descriptor{
+			MediaType:   component.MediaType,
+			Size:        component.Size,
+			Digest:      component.Digest,
+			URLs:        dupStringSlice(component.URLs),
+			Annotations: dupStringStringMap(component.Annotations),
+			Platform:    platform,
```
mtrmac, line 145: This is missing recent `upstream/main` changes.

internal/manifest/common.go
```diff
@@ -0,0 +1,96 @@
...
+	// to also reject case-insensitive matches (which would be used by Go when really decoding
+	// the manifest).
+	// (It is expected that as manifest formats are added or extended over time, more fields will be added
+	// here.)
+	detectedFields := struct {
+		Config    interface{} `json:"config"`
+		FSLayers  interface{} `json:"fsLayers"`
+		History   interface{} `json:"history"`
+		Layers    interface{} `json:"layers"`
+		Manifests interface{} `json:"manifests"`
```
mtrmac, line 67: This is missing a recent `upstream/main` change.

mtrmac (commented):
- Please add the “This is publicly available” comments to _all_ such top-level items. Both data types like `Schema2PlatformSpec` and functions like `MatchesDigest`
- I’m afraid there was recently some churn on the main branch to benefit from Go 1.18, the recent rebase of this branch is reverting some of those. Please update. (FWIW that can be compared using e.g things like `git diff {upstream/main:,other-branch:internal/}manifest/oci_index_test.go `).
- The second commit can be dropped now; we have proven the architecture out already, so let’s avoid another review iteration.

mtrmac:
> * I’m afraid there was recently some churn on the main branch to benefit from Go 1.18, the recent rebase of this branch is reverting some of those. Please update.

Similarly #1824 has a small intersection with this PR. I’m sorry about that hassle, it was necessary to allow building with Go 1.20 and enabling other work.

internal/manifest/docker_schema2_list.go
```diff
@@ -0,0 +1,254 @@
...
+type Schema2PlatformSpec struct {
+	Architecture string   `json:"architecture"`
+	OS           string   `json:"os"`
+	OSVersion    string   `json:"os.version,omitempty"`
+	OSFeatures   []string `json:"os.features,omitempty"`
+	Variant      string   `json:"variant,omitempty"`
+	Features     []string `json:"features,omitempty"` // removed in OCI
+}
+
+// Schema2ManifestDescriptor references a platform-specific manifest.
```
mtrmac, line 26:
```suggestion
// Schema2ManifestDescriptor references a platform-specific manifest.
// This is publicly visible as c/image/manifest.Schema2ManifestDescriptor.
```

internal/manifest/list.go
```diff
@@ -0,0 +1,84 @@
...
+type List interface {
+	ListPublic
+	// CloneInternal returns a deep copy of this list and its contents.
+	CloneInternal() List
+}
+
+// ListUpdate includes the fields which a List's UpdateInstances() method will modify.
```
mtrmac, line 56:
```suggestion
// ListUpdate includes the fields which a List's UpdateInstances() method will modify.
// This is publicly visible as c/image/manifest.ListUpdate.
```

internal/manifest/list.go
```diff
@@ -0,0 +1,84 @@
...
+type ListUpdate struct {
+	Digest    digest.Digest
+	Size      int64
+	MediaType string
+}
+
+// ListPublicFromBlob parses a list of manifests.
```
mtrmac, line 63:
```suggestion
// ListPublicFromBlob parses a list of manifests.
// This is publicly visible as c/image/manifest.ListFromBlob.
```

internal/manifest/oci_index.go
```diff
@@ -0,0 +1,262 @@
...
+func (index *OCI1IndexPublic) Serialize() ([]byte, error) {
+	buf, err := json.Marshal(index)
+	if err != nil {
+		return nil, fmt.Errorf("marshaling OCI1Index %#v: %w", index, err)
+	}
+	return buf, nil
+}
+
+// OCI1IndexPublicFromComponents creates an OCI1 image index instance from the
+// supplied data.
```
mtrmac, line 120: Publicly visible…

internal/manifest/oci_index.go
```diff
@@ -0,0 +1,262 @@
...
+			URLs:        slices.Clone(component.URLs),
+			Annotations: maps.Clone(component.Annotations),
+			Platform:    platform,
+		}
+		index.Manifests[i] = m
+	}
+	return &index
+}
+
+// OCI1IndexPublicClone creates a deep copy of the passed-in index.
```
mtrmac, line 154: Publicly visible…

internal/manifest/oci_index.go
```diff
@@ -0,0 +1,262 @@
...
+			},
+		}
+		components = append(components, converted)
+	}
+	s2 := Schema2ListPublicFromComponents(components)
+	return s2, nil
+}
+
+// OCI1IndexPublicFromManifest creates an OCI1 manifest index instance from marshalled
+// JSON, presumably generated by encoding a OCI1 manifest index.
```
mtrmac, line 197: Publicly visible…

mtrmac (commented):
A few more comment updates, please.

LGTM aftewards, feel free to merge after tests pass.

flouthoc:
@vrothberg @giuseppe @mtrmac PTAL

vrothberg (approved):
LGTM

mtrmac:
@flouthoc Thanks!

PR #1794, opened 2023-01-17
manifest: pull Variant from an OCI config

nalind:
When reading fields from an OCI config blob, pull the `Variant` field's value, too.  Stumbled across it while looking for something else.  Read it from a v2s1 config blob, too, even if it's almost certainly going to be empty.


mtrmac (approved):
Oops.

LGTM. Thanks!

PR #1797, opened 2023-01-18
Remove left-over logging from test development

mtrmac:


vrothberg (commented):
LGTM

Sorry for not catching during review

mtrmac:
The logs aren’t actually visible during ordinary `go test` (only with `-v`), so this is mostly just for conceptual cleanliness.

rhatdan:
LGTM

PR #1816, opened 2023-01-25
Reconnecting blob reader

mtrmac:
Not production-ready at this point; filing this PR early to run through CI, and to generate log entries so that this can be
verified, improved and tuned.

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -472,12 +472,22 @@ func (c *dockerClient) makeRequest(ctx context.Context, method, path string, hea
 		return nil, err
 	}
 
-	urlString := fmt.Sprintf("%s://%s%s", c.scheme, c.registry, path)
-	requestURL, err := url.Parse(urlString)
+	requestURL, err := c.resolveRequestURL(path)
 	if err != nil {
 		return nil, err
 	}
 	return c.makeRequestToResolvedURL(ctx, method, requestURL, headers, stream, -1, auth, extraScope)
+}
+
+// resolveRequestURL turns a path for c.makeRequest into a full URL.
+// Most users should call makeRequest directly, this exists basically to make the URL available for debug logs.
+func (c *dockerClient) resolveRequestURL(path string) (*url.URL, error) {
+	urlString := fmt.Sprintf("%s://%s%s", c.scheme, c.registry, path)
+	res, err := url.Parse(urlString)
+	if err != nil {
+		return nil, err
+	}
+	return res, nil
 }
 
 // Checks if the auth headers in the response contain an indication of a failed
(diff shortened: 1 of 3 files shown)


docker/docker_client.go
```diff
@@ -472,14 +472,24 @@ func (c *dockerClient) makeRequest(ctx context.Context, method, path string, hea
...
 		return nil, err
 	}
 	return c.makeRequestToResolvedURL(ctx, method, requestURL, headers, stream, -1, auth, extraScope)
 }
 
+// resolveRequestURL turns a path for c.makeRequest into a full URL.
+// Most users should call makeRequest directly, this exists basically to make the URL available for debug logs.
+func (c *dockerClient) resolveRequestURL(path string) (*url.URL, error) {
+	urlString := fmt.Sprintf("%s://%s%s", c.scheme, c.registry, path)
+	res, err := url.Parse(urlString)
```
vrothberg, line 486: Micro nit: could return `url.Parse()` directly to avoid the error being checked twice.

mtrmac:
@vrothberg The “unexpected EOF” failures have been a real annoyance recently; https://github.com/containers/podman/pull/17221 has now seen at least 2 cases where (an earlier version of this PR) worked, so I suggest merging this soonish. I think now it’s clearly better to have _some_ improvement quickly than going for perfection.

This will make errors fetching blobs quite verbose and loud, possibly annoying for users, but it would allow us to better tune the heuristics in the future.

mtrmac:
(FWIW the _logic_ is mostly unchanged, but the actual code is changed quite a bit; IIRC the original version of this PR was just crashing instead of reconnecting…)

mtrmac:
Test failure not handled by the heuristic:
```
happened during read: (heuristic tuning data: last retry 0, current offset 117460; 2 ms total, 0 ms since progress): unexpected EOF
```
2 ms since connection start, 100 kB downloaded … that feels rather tight. We certainly can’t go reconnecting every 2 ms.

Maybe we need some kind of budget, like 1 reconnection per 30 seconds (but allowing _one_ almost-immediate one)?

vrothberg (approved):
LGTM

vrothberg:
@mtrmac I leave it up to you whether you want to backport it to Podman v4.4

mtrmac:
quay.io:
```
(heuristic tuning data: last retry 0, current offset 752782; 69.893 ms total, 0.187 ms since progress): unexpected EOF
```

mtrmac:
> @mtrmac I leave it up to you whether you want to backport it to Podman v4.4

https://github.com/containers/podman/pull/17462 for the record (`v4.4`, not `v4.4.1-rhel`)

PR #1821, opened 2023-01-31
Update to github.com/vbauerster/mpb/v8

mtrmac:
Renovate filed #1786, which isn’t sufficient; actually update the users.

I have reviewed the API for changes, but only minimally smoke-tested this in practice with a single copy.

(Note to self: Podman should migrate its users as well.)


rhatdan:
LGTM

mtrmac:
Podman PR: https://github.com/containers/podman/pull/17342 .

PR #1824, opened 2023-02-02
Update to golangci-lint 1.51.0

mtrmac:
This is required for support of Go 1.20.

Also avoid a call to math/rand.Read, which is deprecated in Go 1.20.

---

Then, manually enable all linters, and fix up things that seem either relevant, or easier to fix than to check again.


vrothberg (commented):
LGTM
Needs a rebase

mtrmac:
Rebased. Merging immediately so that `make` with Go 1.20 installed works.

Apologies for the churn.

PR #1826, opened 2023-02-02
Merge pullSource tests as table driven format

QiWang19:
Follow-up PR: https://github.com/containers/image/pull/1411#discussion_r785275923 
Merge the test and test data of TestPullSourcesMirrorFromReference to TestPullSourcesFromReference


QiWang19:
@vrothberg @mtrmac PTAL

pkg/sysregistriesv2/system_registries_v2_test.go
```diff
@@ -650,101 +650,68 @@ func TestPullSourcesFromReference(t *testing.T) {
...
-	registryA, err := FindRegistry(sys, "registry-a.com/foo/image:latest")
-	assert.Nil(t, err)
-	assert.NotNil(t, registryA)
-	// Digest
-	referenceADigest := toNamedRef(t, "registry-a.com/foo/image@sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa")
-	pullSources, err := registryA.PullSourcesFromReference(referenceADigest)
-	assert.Nil(t, err)
-	assert.Equal(t, 3, len(pullSources))
-	assert.Equal(t, "mirror-1.registry-a.com/image@sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa", pullSources[0].Reference.String())
-	assert.True(t, pullSources[1].Endpoint.Insecure)
```
mtrmac, line 666: The tests for.Reference.String() and .Insecure` have not been fully migrated:

The test data should probably add an optional `tagInsecure, digestInsecure []bool` arrays that contain the expected values. Similarly for `Reference`, I guess the arrays could contain the repo, so that the test appends `digest` or `tag` appropriately.

mtrmac (commented):
Thanks for following up on this!

pkg/sysregistriesv2/system_registries_v2_test.go
mtrmac: (I don’t think it’s necessary to add this data to to existing table entries dealing with digest/tag mirroring: if the arrays are `nil` in the test data, that part of the comparison can be skipped.)

pkg/sysregistriesv2/system_registries_v2_test.go
```diff
@@ -650,101 +650,68 @@ func TestPullSourcesFromReference(t *testing.T) {
...
-		SystemRegistriesConfDirPath: "testdata/this-does-not-exist",
-	}
-	registries, err := GetRegistries(sys)
 	require.NoError(t, err)
-	assert.Equal(t, 7, len(registries))
+	assert.Equal(t, 9, len(registries))
 
 	digest := "@sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
-	tag := ":aaa"
+	tag := "aaa"
```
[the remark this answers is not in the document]
QiWang19: oh it was a typo :face_exhaling:

pkg/sysregistriesv2/system_registries_v2_test.go
QiWang19: added the test data

QiWang19:
@mtrmac PTAL

pkg/sysregistriesv2/system_registries_v2_test.go
```diff
@@ -650,124 +650,125 @@ func TestPullSourcesFromReference(t *testing.T) {
...
 
 	digest := "@sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
 	tag := ":aaa"
 	for _, tc := range []struct {
-		registry      string
-		digestSources []string
-		tagSources    []string
+		registry       string
+		repo           string
+		digestSources  []string
```
mtrmac, line 661: Looking at how these fields (and `tagSources`) are used, they are not really `registry`/`repo`: we have `registry` fields naming namespaces.

So maybe hame the fields (`matchedPrefix`, `repo` (`subRepo`??? probably not), `digestPrefixes`, `tagPrefixes`)?

pkg/sysregistriesv2/system_registries_v2_test.go
```diff
@@ -650,124 +650,125 @@ func TestPullSourcesFromReference(t *testing.T) {
...
 		{
-			"registry-g.com/foo/image",
+			"registry-g.com/baz",
+			"image",
 			[]string{"mirror-1.registry-g.com", "mirror-3.registry-g.com", "mirror-4.registry-g.com", "registry-g.com/bar"},
+			[]bool{false, false, false, false},
 			[]string{"mirror-2.registry-g.com", "mirror-3.registry-g.com", "mirror-4.registry-g.com", "registry-g.com/bar"},
+			[]bool{false, false, false, false},
 		},
 	} {
```
mtrmac, line 748: Throughout, `filepath.Join` is a bad fit:
- It uses `os.PathSeparator` instead of `/`
- `filepath.Join` calls `filepath.Clean`. That doesn’t quite _hurt_ but it is not semantically correct.

Maybe hard-coding `/` (with `+"/"+` or `Sprintf` to combine all the elements, I don’t care which one) would be simplest.

mtrmac (commented):
Thanks! Looks good overall, last round of smaller things.

TomSweeneyRedHat:
LGTM, once @mtrmac comments are addressed.  Looks like you need a rebase too @QiWang19

mtrmac (approved):
Thanks! LGTM.

PR #1843, opened 2023-02-10
Remove a direct use of golang.org/x/net/http2

mtrmac:
A version of `golang.org/x/net/http2` is bundled into the standard library: https://github.com/golang/go/blob/master/src/net/http/h2_bundle.go and here we certainly don't care about the specific version or details.

So just enable use of the built-in one.

Note that this is very unlikely to actually remove the dependency from binaries, because it is required by `google.golang.org/grpc/internal/transport`, and that one has quite a few uses; so most of the benefit right now is just
one less package to get dependency update PRs from.

---

Also replace deprecated `Dial` by `DialContext`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -39,7 +39,6 @@ require (
 	go.etcd.io/bbolt v1.3.7
 	golang.org/x/crypto v0.6.0
 	golang.org/x/exp v0.0.0-20230206171751-46f607a40771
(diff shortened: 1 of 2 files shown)


TomSweeneyRedHat:
LGTM

vrothberg (approved):
LGTM

PR #1845, opened 2023-02-11
UNTESTED: Replace sigstore/rekor/pkg/client with a manually-created client

mtrmac:
This removes 4.728 MB from a macOS Skopeo binary.

The costs are:
- A few dozen lines of new code
- A _major_ loss of flexibility. All the removed layers had many options, this code hard-codes the choices we were making. Reintroducing the options could be pretty costly in extra code.
- We lose the ability to debug Rekor API accesses via environment variables (something I had no idea exists; so there’s that.)
- The previous code was able to decode responses in many formats, e.g. YAML, based on the `Content-Type:` header in the response. This one asks for JSON and expects JSON.

Absolutely untested, pretty likely to be broken due to some some fairly trivial but critical typo. Filing early to give more opportunity to review / discuss.


rhatdan:
@lukehinds PTAL

rhatdan:
I like the idea of shrinking the executable, hopefully this same shrinkage would apply to Podman.
I don't see the loss of functionality that you describe as been that drastic.

mtrmac:
Note to self: This should probably only be merged after we have integration tests against a Rekor server.

rhatdan:
Needs a rebase

Luap99:
replaced by https://github.com/containers/image/pull/2873

PR #1847, opened 2023-02-13
Relax retry heuristics

mtrmac:
Even if we didn't make much progress, allow one retry per minute anyway. Notably this allows one retry ~immediately, we see failures a few dozen milliseconds after the connection is set up.

### DIFF
--- a/docker/body_reader.go
+++ b/docker/body_reader.go
@@ -198,19 +198,20 @@ func (br *bodyReader) Read(p []byte) (int, error) {
 	}
 }
 
-// millisecondsSince is like time.Since(tm).Milliseconds, but it returns a floating-point value
-func millisecondsSince(tm time.Time) float64 {
+// millisecondsSinceOptional is like time.Since(tm).Milliseconds, but it returns a floating-point value.
+// If the input time is time.Time{}, it returns math.NaN()
+func millisecondsSinceOptional(tm time.Time) float64 {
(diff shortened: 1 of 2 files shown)


mtrmac:
Untested in practice; I’ll at least open a Podman PR to have a chance to notice stupid crashes.

I should also add unit tests…

mtrmac:
Some unit tests added — for the heuristics, not for the retry implementation itself. (The `Read` implementation has been exercised and is not changing in this PR, so there’s less risk to it.)

docker/body_reader.go
```diff
@@ -17,22 +17,27 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-// bodyReaderMinimumProgress is the minimum progress we want to see before we retry
-const bodyReaderMinimumProgress = 1 * 1024 * 1024
+const (
+	// bodyReaderMinimumProgress is the minimum progress we consider a good reason to retry
+	bodyReaderMinimumProgress = 1 * 1024 * 1024
+	// bodyReaderMSSinceLastRetry is the minimum time since a last retry we consider a good reason to retry
+	bodyReaderMSSinceLastRetry = 60 * 1_000
```
rhatdan, line 24: I did not know about the _ support in 1_000, cool.

vrothberg (approved):
LGTM

PR #1849, opened 2023-02-13
Update letsencrypt/boulder after https://github.com/letsencrypt/boulder/pull/6651

mtrmac:
This removes quite a bit of dependent code (295 kB on macOS).

It also adds a new github.com/go-jose/go-jose.v2 , but we'll get rid of that again after https://github.com/letsencrypt/boulder/pull/6581 lands.


rhatdan:
LGTM

Issue #1852, opened 2023-02-15
Ambiguous discussion of detached signature in containers-signature.5.md


dwmarshall:
Consider a GET to a sigstore that returns content such as:
```
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA256

{"critical":{"identity":{"docker-reference":"my.registry.com/one/two:3."},"image":{"docker-manifest-digest":"sha256:8dce522effa05639e3c65b68e019bf75abc48c2c7275988c03726168304f3f1e"},"type":"atomic container signature"}}
-----BEGIN PGP SIGNATURE-----

wsBcBAEBCAAQBQJj0+0UCRBQSsigm7VTpwAAnx0IAEtGq8n072SPh+ov3x1715Mn
...
+nnbo/ROceOYiuUYOqsx/SM77mZpOv5+dBhxBURqwy5CQfLFy+zwjg8LtT231w8=
=JzjO
-----END PGP SIGNATURE-----
```

Somewhere between v5.19.4 and v5.22.1, the handling of that signature changed from working to not, with a message such as
```
unrecognized signature format, starting with binary 0x2d
```

Reasonable people may disagree whether the above response is a detached signature. There are indeed two blobs within the response that could be considered independent, but they are provided to containers/image as just one blob.

As much as I would like to ask for (or provide) handling for content of this sort, it would helpful to have something added to containers-signature-5.md that such signatures are not valid.

mtrmac:
Thanks for your report.

Pedantically, I’d argue that the spec already implies this by referring to RFC 4880 section 11.3. , which refers to specific binary packet formats, and does not talk about ASCII armor; but it’s clearly not explicit enough.

Pragmatically, much more important is that the implementation chosen by the `containers_image_openpgp` build tag has always rejected ASCII-armored signatures, so they were not widely interoperable (E.g. CRI-O is rejecting them in the default build).

So, I am proposing https://github.com/containers/image/pull/1854 .

---

> As much as I would like to ask for (or provide) handling for content of this sort,

Why is it useful?

To me it seems to only add unnecessary ambiguity, implementation complexity, and data size, for no end-user benefit; if anything, it adds _risk_ because it makes it easier for users to see, and potentially rely on, the contents of the signature without actually verifying its authenticity.

I may well be missing something; what is it?

(Arguably, “we have tons of content signed that way” could count as a good enough reason alone… but, it’s, shall we say, not an _elegant_ reason.)

mtrmac:
[...]

PR #1854, opened 2023-02-17
Make it clear that cleartext signatures are not accepted in simple signing

mtrmac:
The current implementation transparently accepts them in `gpgmeSigningMechanism` , but `openpgpSigningMechanism` rejects them; so make it clear that they should be rejected.

See #1852 .


vrothberg (approved):
LGTM

PR #1855, opened 2023-02-17
Make some symbols in internal packages package-private

mtrmac:
… just to express more precisely how they are used, and to possibly help compilers/linters with scope analysis.

Should not change behavior. All of these types are internal, so we can trivially make them public again if necessary.


vrothberg (commented):
LGTM

PR #1861, opened 2023-02-21
Merge release branch into `main`

mtrmac:
… so that users of the c/image main branch are not offered downgrades, and so that they can continue to use `main` without having to use a `replace` directive.


vrothberg (commented):
LGTM

Note that `go get ...@main` did the trick for me in Podman.

PR #1867, opened 2023-02-27
client: enable HTTP(S) keep-alive

giuseppe:
Enable HTTP(S) keep-alive to improve network performance and reduce latency.  We need several HTTP(S) requests before we get to request the blob and each of them requires a new HTTP(S) connection and that slows down significantly pulling images, especially on networks with a higher latency (e.g. wifi).
This will allow multiple requests to be sent over a single connection, reducing the overhead of establishing new connections for each request.

### DIFF
--- a/docker/daemon/client_test.go
+++ b/docker/daemon/client_test.go
@@ -19,6 +19,8 @@ func TestDockerClientFromNilSystemContext(t *testing.T) {
 
 	assert.Equal(t, dockerclient.DefaultDockerHost, client.DaemonHost(), "The default docker host should have been used")
 	assert.Equal(t, defaultAPIVersion, client.ClientVersion(), "The default api version should have been used")
+
+	assert.NoError(t, client.Close())
 }
 
 func TestDockerClientFromCertContext(t *testing.T) {
@@ -38,6 +40,8 @@ func TestDockerClientFromCertContext(t *testing.T) {
 
 	assert.Equal(t, host, client.DaemonHost())
 	assert.Equal(t, "1.22", client.ClientVersion())
+
+	assert.NoError(t, client.Close())
 }
 
 func TestTlsConfigFromInvalidCertPath(t *testing.T) {
@@ -94,6 +98,7 @@ func TestSpecifyPlainHTTPViaHostScheme(t *testing.T) {
 	assert.NotNil(t, client, "A Docker client reference should have been returned")
 
 	assert.Equal(t, host, client.DaemonHost())
+	assert.NoError(t, client.Close())
 }
 
 func testDir(t *testing.T) string {
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -213,6 +213,7 @@ func dockerCertDir(sys *types.SystemContext, hostPort string) (string, error) {
 // newDockerClientFromRef returns a new dockerClient instance for refHostname (a host a specified in the Docker image reference, not canonicalized to dockerRegistry)
 // “write” specifies whether the client will be used for "write" access (in particular passed to lookaside.go:toplevelFromSection)
 // signatureBase is always set in the return value
+// The caller must call .Close() on the returned client when done.
 func newDockerClientFromRef(sys *types.SystemContext, ref dockerReference, registryConfig *registryConfiguration, write bool, actions string) (*dockerClient, error) {
 	auth, err := config.GetCredentialsForRef(sys, ref.ref)
 	if err != nil {
@@ -247,6 +248,7 @@ func newDockerClientFromRef(sys *types.SystemContext, ref dockerReference, regis
 // (e.g., "registry.com[:5000][/some/namespace]/repo").
 // Please note that newDockerClient does not set all members of dockerClient
 // (e.g., username and password); those must be set by callers if necessary.
+// The caller must call .Close() on the returned client when done.
 func newDockerClient(sys *types.SystemContext, registry, reference string) (*dockerClient, error) {
 	hostName := registry
 	if registry == dockerHostname {
@@ -302,6 +304,7 @@ func CheckAuth(ctx context.Context, sys *types.SystemContext, username, password
 	if err != nil {
 		return fmt.Errorf("creating new docker client: %w", err)
 	}
+	defer client.Close()
 	client.auth = types.DockerAuthConfig{
 		Username: username,
 		Password: password,
@@ -371,6 +374,7 @@ func SearchRegistry(ctx context.Context, sys *types.SystemContext, registry, ima
 	if err != nil {
 		return nil, fmt.Errorf("creating new docker client: %w", err)
 	}
+	defer client.Close()
 	client.auth = auth
 	if sys != nil {
 		client.registryToken = sys.DockerBearerRegistryToken
@@ -1083,4 +1087,12 @@ func (c *dockerClient) getExtensionsSignatures(ctx context.Context, ref dockerRe
 // sigstoreAttachmentTag returns a sigstore attachment tag for the specified digest.
 func sigstoreAttachmentTag(d digest.Digest) string {
 	return strings.Replace(d.String(), ":", "-", 1) + ".sig"
+}
+
+// Close removes resources associated with an initialized dockerClient, if any.
+func (c *dockerClient) Close() error {
+	if c.client != nil {
+		c.client.CloseIdleConnections()
+	}
+	return nil
 }
--- a/docker/docker_image_src.go
+++ b/docker/docker_image_src.go
@@ -153,6 +153,7 @@ func newImageSourceAttempt(ctx context.Context, sys *types.SystemContext, logica
 	s.Compat = impl.AddCompat(s)
 
 	if err := s.ensureManifestIsLoaded(ctx); err != nil {
+		client.Close()
 		return nil, err
 	}
 	return s, nil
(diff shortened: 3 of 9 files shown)


mtrmac (commented):
I would expect at least the
```
	IdleConnTimeout:       90 * time.Second,
```
setting from `http.DefaultTransport` to be relevant. (Arguably we should just use `http.DefaultTransport` instead of customizing, but that would be a bigger change than this PR aims for, notably enabling HTTP/2.)

----

Currently, we use many short-lived `http.Transport` and `http.Client` objects (basically one per accessed image).

In addition, calling `http.Client.CloseIdleConnections` after the client is done (e.g. when closing the `ImageSource`/`ImageDestination` seems appropriate; idle connections created by one `http.Transport` can’t be reused by a separate `http.Transport` AFAICS from reading the implementation.

(I would expect that _eventually_ the server times out and closes the connection, so even without these two things, there should be no long-term _leak_ — but both the server and the client are probably tying up more resources than necessary.)

mtrmac:
(To be clear, I am in favor of doing this, in general. Thanks for taking the time!)

docker/daemon/client_test.go
```diff
@@ -19,6 +19,8 @@ func TestDockerClientFromNilSystemContext(t *testing.T) {
 
 	assert.Equal(t, dockerclient.DefaultDockerHost, client.DaemonHost(), "The default docker host should have been used")
 	assert.Equal(t, defaultAPIVersion, client.ClientVersion(), "The default api version should have been used")
+
+	assert.Nil(t, client.Close())
```
mtrmac, line 23: (Absolutely non-blocking: `assert.NoError` results in better output on failure. That can be changed if the test ever fails, so this is fine for now.)

docker/docker_client.go
```diff
@@ -1084,3 +1086,9 @@ func (c *dockerClient) getExtensionsSignatures(ctx context.Context, ref dockerRe
 func sigstoreAttachmentTag(d digest.Digest) string {
 	return strings.Replace(d.String(), ":", "-", 1) + ".sig"
 }
+
+// Close removes resources associated with an initialized dockerClient, if any.
```
mtrmac, line 1090: Please update the documentation of `newDockerClient` and `newDockerClientFromRef` to say that the caller must call `.Close()` when done.

docker/docker_image_src.go
```diff
@@ -166,7 +166,7 @@ func (s *dockerImageSource) Reference() types.ImageReference {
 
 // Close removes resources associated with an initialized ImageSource, if any.
 func (s *dockerImageSource) Close() error {
-	return nil
+	return s.c.Close()
```
mtrmac, line 169: `dockerImageDestination` needs this as well.

mtrmac, line 169: `newImageSourceAttempt` should ideally close it on the failure path(s) as well.

pkg/tlsclientconfig/tlsclientconfig.go
```diff
@@ -96,8 +96,7 @@ func NewTransport() *http.Transport {
 		Proxy:               http.ProxyFromEnvironment,
 		DialContext:         direct.DialContext,
 		TLSHandshakeTimeout: 10 * time.Second,
-		// TODO(dmcgowan): Call close idle connections when complete and use keep alive
-		DisableKeepAlives: true,
+		IdleConnTimeout:     90 * time.Second,
```
mtrmac, line 99: Adding
```go
		MaxIdleConns:        100,
```
from `DefaultTransport` would be an extra layer of security.

oci/layout/oci_src.go
```diff
@@ -94,6 +94,9 @@ func (s *ociImageSource) Reference() types.ImageReference {
 
 // Close removes resources associated with an initialized ImageSource, if any.
 func (s *ociImageSource) Close() error {
+	if t, ok := s.client.Transport.(*http.Transport); ok {
```
mtrmac, line 97: Isn’t `s.client.CloseIdleConnections()` sufficient?

mtrmac (commented):
Testing _something_ like this, I still see one connection kept open until the `IdleConnTimeout`… I’m not sure why.

The need to close `docker/client.Client` is a great catch.

mtrmac:
<details><summary>My test patch on top of https://github.com/containers/skopeo/pull/1928</summary>

```diff
diff --git a/cmd/skopeo/inspect.go b/cmd/skopeo/inspect.go
index e969caa4..0e02959a 100644
--- a/cmd/skopeo/inspect.go
+++ b/cmd/skopeo/inspect.go
@@ -1,11 +1,16 @@
 package main
 
 import (
+       "bytes"
        "encoding/json"
        "errors"
        "fmt"
        "io"
+       "os"
+       "os/exec"
+       "regexp"
        "strings"
+       "time"
 
        "github.com/containers/common/pkg/report"
        "github.com/containers/common/pkg/retry"
@@ -100,11 +105,11 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error)
                return fmt.Errorf("Error parsing image name %q: %w", imageName, err)
        }
 
-       defer func() {
-               if err := src.Close(); err != nil {
-                       retErr = noteCloseFailure(retErr, "closing image", err)
-               }
-       }()
+       // defer func() {
+       //      if err := src.Close(); err != nil {
+       //              retErr = noteCloseFailure(retErr, "closing image", err)
+       //      }
+       // }()
 
        if err := retry.IfNecessary(ctx, func() error {
                rawManifest, _, err = src.GetManifest(ctx, nil)
@@ -214,7 +219,49 @@ func (opts *inspectOptions) run(args []string, stdout io.Writer) (retErr error)
                        logrus.Warnf("Registry disallows tag list retrieval; skipping")
                }
        }
-       return opts.writeOutput(stdout, outputData)
+       if err := opts.writeOutput(stdout, outputData); err != nil {
+               return err
+       }
+       if err := src.Close(); err != nil {
+               return err
+       }
+
+       start := time.Now()
+       last := ""
+       myPID := fmt.Sprintf("%d", os.Getpid())
+       for i := 0; i < 300; i++ {
+               out, err := exec.Command("lsof", "-p", myPID).CombinedOutput()
+               if err != nil {
+                       return err
+               }
+               cleaned := clean(out)
+               if last != cleaned {
+                       fmt.Fprintf(stdout, "==========@ %f\n%s\n", float64(time.Since(start).Milliseconds())/1000.0, cleaned)
+                       last = cleaned
+               }
+               time.Sleep(1 * time.Second)
+       }
+       return nil
+}
+
+var cleanPairs = []struct {
+       regexp      *regexp.Regexp
+       replacement []byte
+}{
+       {regexp.MustCompile("PIPE[ \t]+"), []byte("PIPE ")},  // Account for PIPE entry alignment differences
+       {regexp.MustCompile("0x[0-9a-f]+"), []byte("0xHEX")}, // Various file IDs
+       {regexp.MustCompile("0t[0-9]+"), []byte("0tTTY")},    // TTY offset
+}
+
+func clean(in []byte) string {
+       b := strings.Builder{}
+       for _, line := range bytes.SplitAfter(in, []byte{'\n'}) {
+               for _, pair := range cleanPairs {
+                       line = pair.regexp.ReplaceAll(line, pair.replacement)
+               }
+               b.Write(line)
+       }
+       return b.String()
 }
 
 // writeOutput writes data depending on opts.format to stdout
 diff --git a/vendor/github.com/containers/image/v5/docker/docker_image_src.go b/vendor/github.com/containers/image/v5/docker/docker_image_src.go
index a115268d..0b193be4 100644
--- a/vendor/github.com/containers/image/v5/docker/docker_image_src.go
+++ b/vendor/github.com/containers/image/v5/docker/docker_image_src.go
@@ -166,6 +166,7 @@ func (s *dockerImageSource) Reference() types.ImageReference {
 
 // Close removes resources associated with an initialized ImageSource, if any.
 func (s *dockerImageSource) Close() error {
+       s.c.client.CloseIdleConnections()
        return nil
 }
 
diff --git a/vendor/github.com/containers/image/v5/pkg/tlsclientconfig/tlsclientconfig.go b/vendor/github.com/containers/image/v5/pkg/tlsclientconfig/tlsclientconfig.go
index 96af5ace..106fc14c 100644
--- a/vendor/github.com/containers/image/v5/pkg/tlsclientconfig/tlsclientconfig.go
+++ b/vendor/github.com/containers/image/v5/pkg/tlsclientconfig/tlsclientconfig.go
@@ -97,8 +97,10 @@ func NewTransport() *http.Transport {
                Proxy:               http.ProxyFromEnvironment,
                DialContext:         direct.DialContext,
                TLSHandshakeTimeout: 10 * time.Second,
+               MaxIdleConns:        100,
+               IdleConnTimeout:     90 * time.Second,
                // TODO(dmcgowan): Call close idle connections when complete and use keep alive
-               DisableKeepAlives: true,
+               // DisableKeepAlives: true,
        }
        return tr
 }
```
</details>

mtrmac:
> Testing _something_ like this, I still see one connection kept open until the `IdleConnTimeout`… I’m not sure why.

Turns out I forgot to patch `GetRepositoryTags` to close its client. With that, all sockets are closed immediately after `inspect` is done.

giuseppe:
@mtrmac thanks for you review.  I've addressed your comments and pushed a new version

mtrmac (approved):
Thanks!

LGTM.

giuseppe:
@vrothberg PTAL

vrothberg (approved):
Nice!
LGTM

PR #1877, opened 2023-03-08
copy,refactor: split `copyOneImage` and `copyMultipleImages`

flouthoc:
As discussed in https://github.com/containers/image/pull/1875 it is suggested to keep the two implementation in different files since there independent complexity is increasing.


flouthoc:
@mtrmac PTAL, the only callers of `copyOneImage` and `copyMultipleImage` is `func Image(` which i kept intact but we can also move parts of it to `single` and `multiple`.

mtrmac (commented):
Thanks!

Please also move
- Smaller utility functions only used in the moved functions (e.g. `compareImageDestinationManifestEqual`). In particular all of the layer code that still is in `copy.go` should move to `single.go`
- Similarly, the global constants/variables that only are used in the moved functions (e.g. `compressionBufferSize`), to the same files
- `imageCopier` itself, to `single.go` (and probably all methods on `imageCopier`).
- Unit tests should move with the tested functions.

---

(If you’d like, I can prepare a PR for these moves — the choice basically depends on how it would be more convenient for you to rebase/update the other outstanding work.)

flouthoc:
> (If you’d like, I can prepare a PR for these moves — the choice basically depends on how it would be more convenient for you to rebase/update the other outstanding work.)

@mtrmac Sure sounds good to me, I'll rebase on top of these changes.

mtrmac:
> (If you’d like, I can prepare a PR for these moves

https://github.com/containers/image/pull/1878

mtrmac:
#1878 was merged.

PR #1878, opened 2023-03-09
Split `copy/copy.go` into three files

mtrmac:
Move the multi-arch code to `copy/multi.go`, and the single-instance code (including handling configs and layers) to `copy/single.go`.

Other than renaming `copyOneImage` to `copySingleImage`, only moves unchanged code, so this should not change behavior.

Cc: @flouthoc


flouthoc (approved):
LGTM Thanks @mtrmac

PR #1881, opened 2023-03-10
copy: move `c.compression*` to `imageCopier`

flouthoc:
There may be needs to modify of selected compression for a single instance so allow setting compression to imageCopier which can be altered while operation is being performed on a single image.

Part of discussion here: https://github.com/containers/image/pull/1875

### DIFF
--- a/copy/compression.go
+++ b/copy/compression.go
@@ -129,13 +129,13 @@ func (ic *imageCopier) bpcCompressUncompressed(stream *sourceStream, detected bp
 	if ic.c.dest.DesiredLayerCompression() == types.Compress && !detected.isCompressed {
 		logrus.Debugf("Compressing blob on the fly")
 		var uploadedAlgorithm *compressiontypes.Algorithm
-		if ic.c.compressionFormat != nil {
-			uploadedAlgorithm = ic.c.compressionFormat
+		if ic.compressionFormat != nil {
+			uploadedAlgorithm = ic.compressionFormat
 		} else {
 			uploadedAlgorithm = defaultCompressionFormat
 		}
 
-		reader, annotations := ic.c.compressedStream(stream.reader, *uploadedAlgorithm)
+		reader, annotations := ic.compressedStream(stream.reader, *uploadedAlgorithm)
 		// Note: reader must be closed on all return paths.
 		stream.reader = reader
 		stream.info = types.BlobInfo{ // FIXME? Should we preserve more data in src.info?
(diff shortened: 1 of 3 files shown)


flouthoc:
@mtrmac @giuseppe @vrothberg PTAL

copy/compression.go
```diff
@@ -129,13 +129,13 @@ func (ic *imageCopier) bpcCompressUncompressed(stream *sourceStream, detected bp
...
-		if ic.c.compressionFormat != nil {
-			uploadedAlgorithm = ic.c.compressionFormat
+		if ic.compressionFormat != nil {
+			uploadedAlgorithm = ic.compressionFormat
 		} else {
 			uploadedAlgorithm = defaultCompressionFormat
 		}
 
-		reader, annotations := ic.c.compressedStream(stream.reader, *uploadedAlgorithm)
+		reader, annotations := ic.c.compressedStream(stream.reader, *uploadedAlgorithm, ic.compressionLevel)
```
mtrmac, line 138: Please make `compressedStream` (and `compressGoroutine`) a method on `imageCopier` instead of adding a parameter.

(Historically, IIRC the code was written to use the more general `copier` when available, for possible generalization to non-`imageCopier` uses, but that doesn’t really reflect any coherent design thought; that generalization would now hinder us, so let’s specialize. Arguably many of these single-purpose utilities only used by `copySingleImage` could be methods of `imageCopier`, not `copier`.)

copy/single.go
```diff
@@ -31,6 +31,8 @@ import (
 // imageCopier tracks state specific to a single image (possibly an item of a manifest list)
 type imageCopier struct {
 	c                          *copier
+	compressionFormat          *compressiontypes.Algorithm // Compression algorithm to use, if the user explicitly requested one, or nil.
+	compressionLevel           *int
```
mtrmac, line 35: (Absolutely non-blocking: Maybe move it near the encryption field, so that the ”important” `manifestUpdates` and `src` come first. This works just fine.)

mtrmac (commented):
Thanks!  One more simplification…

copy/compression.go
flouthoc: Done.

copy/single.go
flouthoc: Moved below.

giuseppe (approved):
LGTM

mtrmac (approved):
Thanks!

PR #1883, opened 2023-03-13
copy/multiple: use more flexible `EditInstances` instead of `UpdateInstances`

flouthoc:
After https://github.com/containers/image/pull/1896 `c/image` now supports new API `EditInstances` which is more flexible in terms of `Adding` new instances or `Modifying` existing instances so lets use that.

### DIFF
--- a/copy/multiple.go
+++ b/copy/multiple.go
@@ -93,18 +93,12 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
 		imagesToCopy = len(options.Instances)
 	}
 	c.Printf("Copying %d of %d images in list\n", imagesToCopy, len(instanceDigests))
-	updates := make([]manifest.ListUpdate, len(instanceDigests))
+	instanceEdits := []internalManifest.ListEdit{}
 	instancesCopied := 0
 	for i, instanceDigest := range instanceDigests {
 		if options.ImageListSelection == CopySpecificImages &&
 			!slices.Contains(options.Instances, instanceDigest) {
-			update, err := updatedList.Instance(instanceDigest)
-			if err != nil {
-				return nil, err
-			}
 			logrus.Debugf("Skipping instance %s (%d/%d)", instanceDigest, i+1, len(instanceDigests))
-			// Record the digest/size/type of the manifest that we didn't copy.
-			updates[i] = update
 			continue
 		}
 		logrus.Debugf("Copying instance %s (%d/%d)", instanceDigest, i+1, len(instanceDigests))
@@ -116,16 +110,16 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
 		}
 		instancesCopied++
 		// Record the result of a possible conversion here.
-		update := manifest.ListUpdate{
-			Digest:    updatedManifestDigest,
-			Size:      int64(len(updatedManifest)),
-			MediaType: updatedManifestType,
-		}
-		updates[i] = update
+		instanceEdits = append(instanceEdits, internalManifest.ListEdit{
+			ListOperation:   internalManifest.ListOpUpdate,
+			UpdateOldDigest: instanceDigest,
+			UpdateDigest:    updatedManifestDigest,
+			UpdateSize:      int64(len(updatedManifest)),
+			UpdateMediaType: updatedManifestType})
 	}
 
 	// Now reset the digest/size/types of the manifests in the list to account for any conversions that we made.
-	if err = updatedList.UpdateInstances(updates); err != nil {
+	if err = updatedList.EditInstances(instanceEdits); err != nil {
 		return nil, fmt.Errorf("updating manifest list: %w", err)
 	}
 


flouthoc:
@mtrmac I was thinking a split like this, I will add unit tests for this new functions once the idea looks good to you, although I have a doubt if this is created additional loop.

cc @giuseppe @vrothberg

copy/multiple.go
```diff
@@ -12,11 +12,46 @@ import (
...
+}
+
+// instancesToCopy returns the list of digests which actually need to be copied to the new manifest list, ignoring the ones which are not needed.
+func (c *copier) instancesToCopy(options *Options, list internalManifest.List) (instances []instanceCopy, copyCount int) {
+	result := []instanceCopy{}
+	instanceDigests := list.Instances()
+	imagesToCopy := len(instanceDigests)
+	if options.ImageListSelection == CopySpecificImages {
+		imagesToCopy = len(options.Instances)
+	}
```
mtrmac, line 43: (I wouldn’t mind counting the items one by one as the relevant operations are chosen, or possibly even doing one more iteration over the loop just to count.; I think we’ll end up doing one of these a future version anyway, and it’s more obviously correct without having to think — so we might as well start now.)

copy/multiple.go
```diff
@@ -12,11 +12,46 @@ import (
...
+	//CopyAndReplicate
+)
+
+type instanceCopy struct {
+	digest digest.Digest
+	op     CopyOperation
+}
+
+// instancesToCopy returns the list of digests which actually need to be copied to the new manifest list, ignoring the ones which are not needed.
+func (c *copier) instancesToCopy(options *Options, list internalManifest.List) (instances []instanceCopy, copyCount int) {
```
mtrmac, line 37: (Absolutely non-blocking: I’d prefer not to use named return values; they tend to result in non-obvious bugs caused just by a missing `:` in `:=` somewhere, especially with `err` and `defer` involved.

I expect this will all change anyway…)

mtrmac (commented):
Thanks!

The design of this is fairly strongly related to the design of the expected future `internal/manifest.List.EditInstances`.

So let’s think what that would look like.

(It might even make sense to prototype that — there’s _some_ chance that we don’t need a package-private `instanceCopy` struct at all, that we can just use a partially-filled input to the `EditInstances` call. But, anyway, this is all internal, so we can change that data structure in when working on `EditInstances` easily enough.)

---

One option is for the `EditInstances` operation to accept “the desired end state”, i.e. the list of images that the resulting manifest list should contain, and how to get them.

Another option is for the `EditInstances` operation to accept “updates”, i.e. what to replace, what to add, what to delete.

The differences are, I think:
- Primarily, how deletes are represented. With “end state”, the caller must make a list of all remaining instances, with “updates”, it’s one entry
- How instance ordering / preference is handled. Currently it is possible for a manifest list to contain several possible “matching” instances (compare the recent work on `ChooseInstance` and all that logic), and the generic `copy` code _probably_ doesn’t understand everything relevant about the ordering / preference logic (in particular, I’m not sure we _want_ to hard-code “earlier match is preferred” here in the generic code.) With “end state”, the caller might end up being responsible for the correct order when adding an instance, with “updates”, it’s the callee.

It seems to me that these suggest that the “updates” model is probably better; it potentially allows calling `EditInstances` in something like `buildah manifest` without involving `c/image/copy` at all, while still benefiting from the full ordering knowledge.

Does that make sense to you? (I didn’t actually _read_ how `buildah manifest` / `podman manifest` works in detail. Maybe I should do that first?)

---

Now… the “updates” data is something we can build, but we don’t have a good way to _consume_ it in the existing `UpdateInstances`; so, actually, maybe we should build some subset of the new `List.EditInstances` _first_?

Alternatively, we could build a minimal/restricted version of the consumption of the updates here in this PR, and move it into EditInstances later:


```go
updates := map(instanceDigests, updatedList.Instance) // preserve by default
for _, edit := range decidedUpdateEdits {
    switch edit {
    case Replace:
        i := index(updates, update.sourceDigest); assert i != nil
        updates[i].Digest = updatedDigest
        // Update other fields…
    }
}
```

---

(The detailed review comments are probably premature.)

mtrmac:
Sure, the `copy` refactor and adding `EditInstances` are conceptually separate but they need to be at least designed concurrently. So working on `EditInstances` SGTM. Or possibly do the two in one PR, although I think smaller PRs make it easier to keep track of discussions.

(The PRs also don’t need to be _perfectly_ clean — it’s fine to refactor `copy` here, and then in the `EditInstances` one, to adjust `copy` further. It just seems like an unnecessary effort to design/implement one approach first, including the tests, and then to completely abandon that and to go in a different direction in the next PR.)

Ultimately this is up to you — I think the smaller PRs are easier, but you might have a different opinion on that.

flouthoc:
It seems https://github.com/containers/image/pull/1896 is about to get merged, i am getting back to this.

copy/multiple.go
flouthoc: Not sure if this is relevent anymore `EditInstances` directly take list of updates which can be directly updated.

flouthoc:
@mtrmac I think after https://github.com/containers/image/pull/1896 my old approach is not even valid anymore  since `EditInstances` directly supports `ListOperation` which does everything this. 

I think we will just need another case which can handle case for `ListOpAdd` but that needs to be done later on. Do you think we need anything else in this PR ?

mtrmac (commented):
The original PR was intended to split the _logic deciding what to copy_ (which will become a bit more complex with the need to see which instances need a new copy added), so that we can write and unit-test that in isolation. I.e. to split the _decision_ what to copy and the _execution_ of that.

I think that’s still something that would be valuable to build — the ability to cheaply unit test that logic, as well as splitting `copyMultipleImages` into smaller pieces, seem useful.

In the meantime, what this PR is now, just a migration from `UpdateInstances` to `EditInstances` is independently a valuable step. So let’s do it!

---

Code for this LGTM, just some naming / stylistic nits.

mtrmac (commented):
The original PR was intended to split the _logic deciding what to copy_ (which will become a bit more complex with the need to see which instances need a new copy added), so that we can write and unit-test that in isolation. I.e. to split the _decision_ what to copy and the _execution_ of that.

I think that’s still something that would be valuable to build — the ability to cheaply unit test that logic, as well as splitting `copyMultipleImages` into smaller pieces, seem useful.

In the meantime, what this PR is now, just a migration from `UpdateInstances` to `EditInstances` is independently a valuable step. So let’s do it!

---

Code for this LGTM, just some naming / stylistic nits.

flouthoc:
@mtrmac PTAL again.

mtrmac (approved):
Thanks!

flouthoc:
@mtrmac I created a new PR to split this https://github.com/containers/image/pull/1982

PR #1884, opened 2023-03-13
Don't claim auth.json contains encrypted information

mtrmac:
... to decrease the risk that users might think it doesn't need protecting.


rhatdan:
LGTM
@vrothberg @TomSweeneyRedHat PTAL

vrothberg (approved):
LGTM

PR #1886, opened 2023-03-14
Cirrus: Replace Ubuntu container w/ Debian VM

cevich:
Previously a pre-built Ubuntu-based container image was used for several CI tasks.  With the general move toward Debian however, a container image is no-longer available.  Remove the old (CI-only) Dockerfile and convert the tasks to run on Debian SID VMs.

Ref: https://github.com/containers/automation_images/pull/266


contrib/cirrus/runner.sh
```diff
@@ -43,6 +45,15 @@ _run_setup() {
     "${SKOPEO_PATH}/${SCRIPT_BASE}/runner.sh" setup
 }
 
+_run_debian_setup() {
+    msg "Installing previously downloaded/cached Docker packages"
+    ooe.sh dpkg -i \
+        $PACKAGE_DOWNLOAD_DIR/containerd.io*.deb \
+        $PACKAGE_DOWNLOAD_DIR/docker-ce*.deb
```
mtrmac, line 52: Is this necessary for c/image / Skopeo? I wouldn’t expect either one to need `docker` if `podman` is installed, and I don’t think anything uses `containerd`.

contrib/cirrus/runner.sh
```diff
@@ -43,6 +45,15 @@ _run_setup() {
...
+    git remote update
+    make tools
```
mtrmac, line 54: Keeping this in the Cirrus `validate_task`, and not doing that for the other ones, might speed up the other ones a tiny bit.

Arguably we should fix the `Makefile` so that it installs tools automatically on-demand, so that users of `make validate` don’t need to care.

And I guess ideally the tools could be pre-installed in the VM, though that’s definitely not blocking this PR and it can be done when/if someone has extra time.

mtrmac (commented):
(Just a quick bikeshed to follow what’s going on… I’ll gladly leave the _actual_ decisions like Debian/Ubuntu/Fedora and container/VM to your expertise.)

contrib/cirrus/runner.sh
cevich: I dunno. I only put it here because the old `.cirrus.Dockerfile` installed some package called `docker.io`.  If you know for sure it's unnecessary, I'll gladly strip it out.

contrib/cirrus/runner.sh
cevich: Good suggestion about avoiding `make tools` beyond validate, I'll do that.

contrib/cirrus/runner.sh
mtrmac: AFAICS that originally comes from c000d6b2673e3c8a07dcd93d628ad2f2e0d2e4c3 , when we were running `test-skopeo` inside that container (and Podman was before 1.0).

Anyway, it’s certainly fine for this PR to only migrate the current behavior , and not do other improvements just because they touch the same areas.

cevich: Thanks for confirming and checking the history.  I'll strip it out because it means entirely avoiding a call to the runner script.

cevich:
**note**: Tests will still fail, new images aren't fully built yet.

cevich:
Attempt at a fix.  Total guess.

cevich:
@mtrmac PTAL, [this seems like a LOT of test failures](https://cirrus-ci.com/task/5230339853910016) for just updating from Ubuntu -> Debian.  Perhaps due to a big jump in golang version maybe?

mtrmac:
@cevich it seems that the tests now run as root, not as an unprivileged user. That explains at least some of the failures (which try to intentionally trigger a permission denial).

Also, perhaps `HOME` is not set at all?! That could break quite a lot of things, or at least trigger rarely-used code paths. I wouldn’t quite say that the tests need a full login session, but something a bit closer to that would be useful.

cevich:
> Also, perhaps HOME is not set at all?!

I know for a fact it's not.  I can set it easy enough.

> I agree setting XDG_CACHE_HOME can make this work, but setting HOME is probably a more general fix.

oh okay, yep you're right that probably would be better.

> the tests now run as root, not as an unprivileged user

Yeah, I was afraid that might lead to trouble.  So, since the tests expect to be running as a user, maybe it's worth me implementing the `ssh testuser@localhost` mechanism.  That will fix the unset `$HOME` problem as well.

cevich:
This is ready now.

mtrmac (approved):
Thanks! LGTM.

(Please rebase.)

PR #1890, opened 2023-03-17
refactor(docs): Disambiguate sigstoreSigned documentation

Delet0r:
From the current documentation, one could mistakenly assume, that `use-sigstore-attachments` needs to be enabled by the registry and not in the client. This commit rephrases the part of the documentation to make it more clear, that this needs to be done on the client side.

    Signed-off-by: Lucas Humfeldt <lucas@humfeldt.de>

### DIFF
--- a/docs/containers-policy.json.5.md
+++ b/docs/containers-policy.json.5.md
@@ -282,7 +282,7 @@ signed by the provided public key.
 The `signedIdentity` field has the same semantics as in the `signedBy` requirement described above.
 Note that `cosign`-created signatures only contain a repository, so only `matchRepository` and `exactRepository` can be used to accept them (and that does not protect against substitution of a signed image with an unexpected tag).
 
-To use this with images hosted on image registries, the relevant registry or repository must have the `use-sigstore-attachments` option enabled in containers-registries.d(5).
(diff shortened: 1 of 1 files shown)


docs/containers-policy.json.5.md
```diff
@@ -282,7 +282,7 @@ signed by the provided public key.
 The `signedIdentity` field has the same semantics as in the `signedBy` requirement described above.
 Note that `cosign`-created signatures only contain a repository, so only `matchRepository` and `exactRepository` can be used to accept them (and that does not protect against substitution of a signed image with an unexpected tag).
 
-To use this with images hosted on image registries, the relevant registry or repository must have the `use-sigstore-attachments` option enabled in containers-registries.d(5).
+To use this with images hosted on image registries, the `use-sigstore-attachments` option needs to be enabled for any registry specifically in the client's containers-registries.d(5).
```
mtrmac, line 285: “the relevant registry or repository” is right; “any registry” is not saying the same thing, and I don’t understand “any registry specifically” at all.

mtrmac (commented):
Thanks, this is definitely worth clarifying.

docs/containers-policy.json.5.md
Delet0r: Yeah, you are right. That sounded better when is was writing it. I'll change that back.

mtrmac (approved):
LGTM. Thanks!

PR #1892, opened 2023-03-21
Don’t discard annotations on blob reuse and partial pulls

### LINKED ISSUE #1888: Image Copy looses layer annotations when reusing blob
I've noticed an issue when pushing an image to a container registry where the layer annotations are lost if the layer already exists in the registry

As an example a layer with this digest
```
{
    "mediaType": "application/vnd.oci.image.layer.v1.tar+zstd",
    "digest": "sha256:14e05236c928e6b9583038025d14979e596210b54c4a2966bfff41172e7acab4",
    "size": 1199,
    "annotations": {
        "io.github.containers.zstd-chunked.manifest-checksum": "sha256:6a716ddd20d5e0a811ed8530a6fd5ed10eac5fd06a80d4690a0b973857a940b4",
        "io.github.containers.zstd-chunked.manifest-position": "748:403:1976:1"
    }
}
```
results in
```
{
    "mediaType": "application/vnd.oci.image.layer.v1.tar+zstd",
    "digest": "sha256:14e05236c928e6b9583038025d14979e596210b54c4a2966bfff41172e7acab4",
    "size": 1199
}
```

After digging through the code I believe it's down to this function not knowing about the original annotations https://github.com/containers/image/blob/0c6d07e02a9a0c8119885cc461c89b7452684ba5/docker/docker_image_dest.go#L321

mtrmac:
Primarily, this modifies the `TryReuseBlob*` and `PutBlobPartial` code paths to preserve annotations in the source manifest; fixes #1888 .

To do that, the private `PutBlob*` now return a `private.UploadedBlob`, and `TryReusingBlobWithOptions` returns a `ReusedBlob`, with a much more specific and explicitly documented set of fields and requirements, with the idea that adding a new struct is a small price to pay for the ability to be more precise about field semantics.

Also, this fixes a possible infinite recursion (when storageImageDestination is used incorrectly)


vrothberg (approved):
LGTM

TomSweeneyRedHat:
CHanges LGTM, but looks like you need a rebase.

PR #1894, opened 2023-03-23
Add FIXMEs about handling of zstd:chunked blob annotations on blob changes

mtrmac:
It's completely undefined whether the OCI blob annotations apply to the object as a concept, regardless of representation, or to the specific binary representation. So it's unclear whether we should preserve or drop them when compressing/decompressing/substituting blobs.

In particular, we currently don't truly correctly handle the zstd:chunked annotations on:
- decompression (should be dropped)
- recompression (should be dropped)
- substitution (should be replaced by data about the other blob, if any; we don't record that)

Right now, we drop all annotations on decompression and recompression (which happens to work fine), and preserve annotations on substitution (which is technically invalid).

Luckily, the zstd:chunked use is opportunistic, and if the annotations are invalid or not applicable, the manifest checksum fails, and we fall through to an ordinary pull; so, that is not quite a deal breaker.

So, for now, just add FIXMEs recording the pain points.

To fix this truly correctly, we would need:
- a new `metadataCleaner` field in `pkg/compression/internal.Algorithm`
- a new `pkg/compression.CleanMetadata`
- turning public `manifest.Manifest` into `internal/manifest.Manifest` where we can add methods
- adding `internal/manifest.Manifest.LayerInfosWithCompression` that turns MIME types into `compression.Algorigthm values
- (using that in `copy.copyLayer` instead of the current hard-coded switch)
- then either defining a new alternative to `UpdatedImage` that can handle these annotations naturally,
  or all the marked users that need to clean the annotations themselves calling `LayerInfosWithCompression`
  and `CleanMetadata` on the affected blobs.
- recording the zstd annotations in the blob info cache
- reading those annotations when substituting blobs based on the cache

We should do all that long-term, but that's quite a lot of work to fix a metadata inconsistency which we can currently silently, with moderate cost, hide from the user.

---

Also includes a fix for some typos discovered when scoping the “correct” fix.

### DIFF
--- a/copy/compression.go
+++ b/copy/compression.go
@@ -176,7 +176,7 @@ func (ic *imageCopier) bpcRecompressCompressed(stream *sourceStream, detected bp
 		recompressed, annotations := ic.compressedStream(decompressed, *ic.compressionFormat)
 		// Note: recompressed must be closed on all return paths.
 		stream.reader = recompressed
-		stream.info = types.BlobInfo{ // FIXME? Should we preserve more data in src.info?
+		stream.info = types.BlobInfo{ // FIXME? Should we preserve more data in src.info? Notably the current approach correctly removes zstd:chunked metadata annotations.
 			Digest: "",
 			Size:   -1,
 		}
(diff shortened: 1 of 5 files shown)


mtrmac:
@giuseppe FYI, to make sure you have the full picture. If this handling is unreasonable, please speak up.

TomSweeneyRedHat:
if @giuseppe is hip with these changes, LGTM

rhatdan:
@giuseppe waiting on you?

giuseppe (approved):
LGTM

PR #1896, opened 2023-03-25
manfiest: prepare internal `EditInstances`

flouthoc:
Introduces a new private API which is modification of UpdateInstances so it can allow and support more that one operation on exisiting manifest list.


flouthoc:
@mtrmac I was thinking of introducing a new `ListEdit` type which can contain `instanceOperation`. WDYT ?

internal/manifest/list.go
```diff
@@ -65,6 +69,22 @@ type ListUpdate struct {
 	MediaType string
 }
 
+type instanceOperation int
```
mtrmac, line 72: Absolutely non-blocking: In an internal package, this can be public — that would allow callers to store that value in a variable (and in particular, it could allow `c/image/copy` to use this enum, instead of inventing its own, to represent “that kind of an update will need to happen” when creating the copy plan, when the full `ListEdit` value can not yet be constructed.

But we can change that when/if a user is added.

internal/manifest/list.go
```diff
@@ -65,6 +69,22 @@ type ListUpdate struct {
...
+const (
+	op_ignore instanceOperation = iota + 1
+	op_update
+)
+
+// ListEdit includes the fields which a List's EditInstances() method will modify.
+type ListEdit struct {
+	ListUpdate
+	Platform            *imgspecv1.Platform
+	Schema2ListPlatform *Schema2PlatformSpec
```
mtrmac, line 83: I think there should be just one field; it’s the job of this subpackage to deal with the individual formats. (Or do the two have different capabilities such that the OCI one can’t represent schema2?)

internal/manifest/list.go
```diff
@@ -65,6 +69,22 @@ type ListUpdate struct {
...
+	Annotations         map[string]string
+	op                  instanceOperation
```
mtrmac, line 85: Typically I’d expect the operation that determines the semantics, and requirements on, the other fields, to be the first field in a struct.

---

If this is a private field, how is the `copy` caller going to set it? Do you plan to provide specialized constructors for every value of `instanceOperation`? That would work fine but might not be worth the effort for internal-only code.

internal/manifest/oci_index.go
```diff
@@ -84,6 +84,45 @@ func (index *OCI1IndexPublic) UpdateInstances(updates []ListUpdate) error {
...
+			}
+			index.Manifests[i].Size = updates[i].Size
+			if updates[i].MediaType == "" {
+				return fmt.Errorf("update %d of %d passed to OCI1Index.UpdateInstances had no media type (was %q)", i+1, len(updates), index.Manifests[i].MediaType)
+			}
+			index.Manifests[i].MediaType = updates[i].MediaType
+			if updates[i].Platform != nil {
+				index.Manifests[i].Platform = updates[i].Platform
+			}
+			if len(updates[i].Annotations) != 0 {
```
mtrmac, line 114: This doesn’t allow deleting all annotations; checking for `….Annotations != nil` would make that possible.

internal/manifest/oci_index.go
```diff
@@ -84,6 +84,45 @@ func (index *OCI1IndexPublic) UpdateInstances(updates []ListUpdate) error {
...
+			index.Manifests[i].Digest = updates[i].Digest
+			if updates[i].Size < 0 {
+				return fmt.Errorf("update %d of %d passed to OCI1Index.UpdateInstances had an invalid size (%d)", i+1, len(updates), updates[i].Size)
+			}
+			index.Manifests[i].Size = updates[i].Size
+			if updates[i].MediaType == "" {
+				return fmt.Errorf("update %d of %d passed to OCI1Index.UpdateInstances had no media type (was %q)", i+1, len(updates), index.Manifests[i].MediaType)
+			}
+			index.Manifests[i].MediaType = updates[i].MediaType
+			if updates[i].Platform != nil {
```
mtrmac, line 111: https://github.com/containers/image/pull/1875/files#r1128485829

internal/manifest/oci_index.go
```diff
@@ -84,6 +84,45 @@ func (index *OCI1IndexPublic) UpdateInstances(updates []ListUpdate) error {
 	return nil
 }
 
+func (index *OCI1Index) EditInstances(updates []ListEdit) error {
+	if len(updates) != len(index.Manifests) {
+		//return fmt.Errorf("incorrect number of update entries passed to OCI1Index.UpdateInstances: expected %d, got %d", len(index.Manifests), len(updates))
+		if len(updates) > len(index.Manifests) {
```
mtrmac, line 90: https://github.com/containers/image/pull/1875/files#r1128482848

internal/manifest/oci_index.go
```diff
@@ -84,6 +84,45 @@ func (index *OCI1IndexPublic) UpdateInstances(updates []ListUpdate) error {
 	return nil
 }
 
+func (index *OCI1Index) EditInstances(updates []ListEdit) error {
```
mtrmac, line 87: It would be useful for `UpdateInstances` to delegate to the new more general `EditInstances`, so that we only have one copy of the “business logic”.

But let’s only add that to this PR only after the design of `EditInstances` settles.

mtrmac (commented):
Thanks! We do need a new API, and it’s better to add it private first.

---

This doesn’t deal in any way with the design questions in https://github.com/containers/image/pull/1883#pullrequestreview-1337397386 ; the API in this PR right now requires the caller to handle ordering (which I think this subpackage should be handling), and eventually we need to think about deletes.

(It’s perfectly fine to not _implement_ deletes now when we don’t need them, not even the enum value needs to be defined now — but it seems worthwhile to make it possible to add that without another large refactoring; and anyway the ordering point suggests a delete-friendly design.

---

Compare https://github.com/containers/image/pull/1883#discussion_r1134219100 .)

So, maybe structurally something like
```swift
enum ListEdit {
// By default, all existing entries are preserved unmodified, unless .update or .delete instructs otherwise
// .update never changes the relative priority (i.e. position) of an entry.
case .update(oldDigest: Digest, newDigest: Digest?, newSize: Int64?, newMIMEType: String?, …)
// The `List` instance is responsible for adding the new entry with the right priority (position), compared to other entries. The caller just specifies what to add.
case .add(digest: Digest, size: Int64, mimeType: String, …)
// case .delete(digestToDelete: Digest) // unimplemented
}
```

---

(The comments on OCI apply symmetrically to schema2.)

internal/manifest/oci_index.go
flouthoc: @mtrmac Could you rephrase this comment, i did not get it from previous line.

internal/manifest/oci_index.go
flouthoc: Added a function below to remove element, it should resize the slice as well.

mtrmac: We do need the delete operation, eventually. I’m not sure that we need it _now_ (though adding it, by itself, wouldn’t hurt), and more importantly that doesn’t resolve the issues with this code branch.

- The API documentation still says the number of updates must match (which is what the _original_ review comment was pointing out.)
- Adding `ListOpRemove` directly causes the `newIndexes` reallocation to be incorrect, because `len(updates)` now counts `ListOpRemove` items.
- The update implementation is now _also_ broken by this: e.g. a two-item manifest list with entries (`A`, `B`), and edits (`ListOpRemove`, `ListOpUpdate(B')`), which _should_ ideally result in a single-item (`B'`), instead results in (`B`, `B'`), because the indices get out of sync (and if the `newIndexes` allocation size were computed correctly, it would cause a panic due to out-of-bounds indexing).
- Also consider how much work it would be to remove one instance but keep all others unchanged: the caller would have to create a `[]ListEdit` with N-1 `ListOpIgnore` entries and one `ListOpRemove` entry, when ideally it, IMHO, should be just a single “remove that one instance“ instruction.

I think we definitely need a design that can be extended to support deletions; but I don’t really think that’s _this_ design, so dealing with a correct _implementation_ of deletions seems to be a bit premature. Or, maybe, do you want the code in precisely so that we can game out all the design and implementation concerns, even at the cost of extra coding effort?

internal/manifest/oci_index.go
```diff
@@ -84,6 +84,54 @@ func (index *OCI1IndexPublic) UpdateInstances(updates []ListUpdate) error {
...
+		}
+	}
+	return nil
+}
+
+// removeManifests is internal function which takes the original slice and an index,
+// as a result it will return slice of manifests with element removed and order kept
+// intact.
+func removeManifests(slice []imgspecv1.Descriptor, s int) []imgspecv1.Descriptor {
+	return append(slice[:s], slice[s+1:]...)
```
mtrmac, line 132: BTW I’d prefer this to use https://pkg.go.dev/golang.org/x/exp/slices#Delete instead of open-coding.

internal/manifest/oci_index.go
flouthoc: >  I’m not sure that we need it now (though adding it, by itself, wouldn’t hurt)

I think we don't need `delete` as of now but I thought why not add it early, OTOH its just increasing the coding overhead should we do `delete` later on one a minimal MVP of the API is already in and working ?

mtrmac: I really don’t mind either way, entirely up to you.

If deletion is a part of this PR, it will need to work, correctly, and probably with tests. That’s extra effort, but having the code makes it easier to discuss / imagine the parts of the API that we do need right now, sure, why not. Personally I think I can imagine these parts of the design well enough without actually preparing a production-quality quality.

flouthoc: Okay I have removed `delete` for now to speed things up, I think it'll be easier to add it later on once major infrastructure is built.

mtrmac:
I think we need something like https://github.com/containers/image/pull/1896#pullrequestreview-1359605527 :

```go
type ListEdit {
  Op ListEditOperation

  // if Op == ListEditUpdate (basically the previous UpdateInstances). All fields must be set.
  UpdateOldDigest digest.Digest
  UpdateDigest digest.Digest
  UpdateSize int64
  UpdateMediaType string

  // If Op = ListEditAdd. All fields must be set.
  AddDigest digest.Digest
  AddSize int64
  AddMediaType string
  AddPlatform imgspecv1.Platform
  AddAnnotations map[string]string
}

func (index *…) EditInstaces(edits []ListEdit) error {
  addedEntries := []Descriptor{}
  for _, edit := range edits {
    switch edit.op {
    case ListEditUpdate:
      i := /*index of entry matching UpdateOldDigest in index.Manifests */
      // Existing UpdateInstances code on index.Manifests[i]
    case ListEditAdd:
      addedEntries = append(addedEntries, /* data from edit */)
  }
  // SOMEHOW sort addedEntries, and index.Manifests, to a reasonable resulting list of instances, and update index.Manifests.
}
```

with the “somehow sort” part being really the hard problem. When adding a Zstd instance to an existing gzip instance, sort the new one after; when adding a gzip instance to an existing Zstd one, sort the new one before. Hypothetically, if we get 3 different compression formats, the pre-existing manifest could already contain 2 of them in the “wrong” order, and I don’t know what we would do when adding the third one.

Maybe the only long-term sustainable thing is to just ignore the pre-existing order of `index.Manifests`, and whenever we are adding and entry, to re-sort them all (first by platform, just for human readability; then for platform variants - least compatible first?, then for compression variants).

flouthoc:
@mtrmac One small question ( since I am not sure :grin:  ) why is sorting important here ? We are anyways doing the comparison in `ChooseInstance` while instance is selected from the manifest list https://github.com/containers/image/blob/main/internal/manifest/oci_index.go#L153

mtrmac:
> why is sorting important here

To allow other implementations unaware of Zstd to choose the right instance. (And potentially to similarly deal with other “dimensions” of priority that might emerge in the future.)

internal/manifest/oci_index.go
```diff
@@ -33,6 +34,74 @@ type OCI1IndexPublic struct {
...
+			m1IsZstd = true
+		}
+	}
+	isZstd, ok = m2.Annotations["io.github.containers.compression.zstd"]
+	if ok {
+		if isZstd == "true" {
+			m2IsZstd = true
+		}
+	}
+	return !m2IsZstd && m1IsZstd
```
mtrmac, line 60: Doesn’t this sort Zstd before gzip? (Ultimately there should be a test of `EditInstances`)

For compatibility, IIRC we want gzip first, so that old implementations use that one.

internal/manifest/list.go
```diff
@@ -65,6 +69,32 @@ type ListUpdate struct {
 	MediaType string
 }
 
+type ListOp int
+
+const (
+	ListOpIgnore ListOp = iota + 1
```
mtrmac, line 75: After `ListOpUpdate` is modified to use `UpdateOldDigest`, there’s no reason for `ListOpIgnore` to exist.

Maybe make the first filled `listOpInvalid` (and then handle it in the `default` case of `switch` statements, just to detect uninitialized fields.

mtrmac (commented):
Thanks, that’s some progress.

There is still some fundamental disconnect about the “update” operation — the caller should only need to edit just the wanted updates, not replacements for everything.

Note also the part about not sorting when no updates happen.

---

The schema2 comments apply to OCI similarly.

mtrmac (commented):
Thanks, that’s some progress.

There is still some fundamental disconnect about the “update” operation — the caller should only need to edit just the wanted updates, not replacements for everything.

Note also the part about not sorting when no updates happen.

---

The schema2 comments apply to OCI similarly.

internal/manifest/docker_schema2_list.go
```diff
@@ -90,6 +90,40 @@ func (list *Schema2ListPublic) UpdateInstances(updates []ListUpdate) error {
...
 
+func (index *Schema2List) EditInstances(updates []ListEdit) error {
+	//addedEntries := []Schema2ManifestDescriptor{}
+	for i, editInstance := range updates {
+		switch editInstance.ListOperation {
+		case ListOpUpdate:
+			if err := editInstance.UpdateOldDigest.Validate(); err != nil {
+				return fmt.Errorf("update %d of %d passed to OCI1Index.UpdateInstances contained an invalid digest: %w", i+1, len(updates), err)
+			}
+			index.Manifests[i].Digest = editInstance.UpdateDigest
```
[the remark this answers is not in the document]
mtrmac: > Instead, it should find the instance by `UpdateOldDigest`

(`slices.IndexFunc`)

internal/manifest/oci_index.go
```diff
@@ -84,6 +85,40 @@ func (index *OCI1IndexPublic) UpdateInstances(updates []ListUpdate) error {
 	return nil
 }
 
+func (index *OCI1Index) EditInstances(updates []ListEdit) error {
+	addedEntries := []imgspecv1.Descriptor{}
+	for i, editInstance := range updates {
```
mtrmac, line 90: `editInstance` vs. `updates`, see similarly to the previous schema2 case

internal/manifest/oci_index.go
```diff
@@ -84,6 +85,40 @@ func (index *OCI1IndexPublic) UpdateInstances(updates []ListUpdate) error {
...
+			if editInstance.UpdateSize < 0 {
+				return fmt.Errorf("update %d of %d passed to OCI1Index.UpdateInstances had an invalid size (%d)", i+1, len(updates), editInstance.UpdateSize)
+			}
+			index.Manifests[i].Size = editInstance.UpdateSize
+			if editInstance.UpdateMediaType == "" {
+				return fmt.Errorf("update %d of %d passed to OCI1Index.UpdateInstances had no media type (was %q)", i+1, len(updates), index.Manifests[i].MediaType)
+			}
+			index.Manifests[i].MediaType = updates[i].UpdateMediaType
+		case ListOpAdd:
+			addedEntries = append(addedEntries, imgspecv1.Descriptor{MediaType: editInstance.AddMediaType,
```
mtrmac, line 106:
```suggestion
			addedEntries = append(addedEntries, imgspecv1.Descriptor{
						MediaType: editInstance.AddMediaType,
```
also elsewhere

mtrmac (commented):
(nice work on the recent changes.)

internal/manifest/docker_schema2_list.go
```diff
@@ -90,6 +99,50 @@ func (list *Schema2ListPublic) UpdateInstances(updates []ListUpdate) error {
...
+			}
+			instanceToBeModified.Digest = editInstance.UpdateDigest
+			if editInstance.UpdateSize < 0 {
+				return fmt.Errorf("update %d of %d passed to Schema2List.UpdateInstances had an invalid size (%d)", i+1, len(editInstances), editInstance.UpdateSize)
+			}
+			instanceToBeModified.Size = editInstance.UpdateSize
+			if editInstance.UpdateMediaType == "" {
+				return fmt.Errorf("update %d of %d passed to Schema2List.UpdateInstances had no media type (was %q)", i+1, len(editInstances), index.Manifests[i].MediaType)
+			}
+			instanceToBeModified.MediaType = editInstance.UpdateMediaType
```
mtrmac, line 125: How does editing the local `…ToBeModified` variable make any difference?

internal/manifest/docker_schema2_list.go
```diff
@@ -44,6 +44,15 @@ func (list *Schema2ListPublic) MIMEType() string {
 	return list.MediaType
 }
 
+func (list *Schema2ListPublic) Descriptor(instanceDigest digest.Digest) Schema2ManifestDescriptor {
```
mtrmac, line 47: This can be private, can’t it?

internal/manifest/docker_schema2_list.go
```diff
@@ -90,6 +99,50 @@ func (list *Schema2ListPublic) UpdateInstances(updates []ListUpdate) error {
...
+					OS:           editInstance.AddPlatform.OS,
+					Architecture: editInstance.AddPlatform.Architecture,
+					OSVersion:    editInstance.AddPlatform.OSVersion,
+					OSFeatures:   editInstance.AddPlatform.OSFeatures,
+					Variant:      editInstance.AddPlatform.Variant,
+				},
+			}
+			addedEntries = append(addedEntries, addInstance)
+		default:
+			return fmt.Errorf("internal error: invalid operation: %d", ListOpInvalid)
```
mtrmac, line 139: The euro should include the actually invalid `ListOperation` value, not a constant.

mtrmac (commented):
There are several earlier unaddressed review comments.

internal/manifest/docker_schema2_list.go
flouthoc: I meant to pass by reference.

internal/manifest/docker_schema2_list.go
flouthoc: I meant to pass by reference.

internal/manifest/docker_schema2_list.go
flouthoc: Done.

internal/manifest/docker_schema2_list.go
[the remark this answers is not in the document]
flouthoc: @mtrmac Isn't this addressed in recent commit ?

internal/manifest/list.go
```diff
@@ -55,6 +55,10 @@ type List interface {
 	// SystemContext ( or for the current platform if the SystemContext doesn't specify any detail ) and preferGzip for compression which
 	// when configured to OptionalBoolTrue and chooses best available compression when it is OptionalBoolFalse or left OptionalBoolUndefined.
 	ChooseInstanceByCompression(ctx *types.SystemContext, preferGzip types.OptionalBool) (digest.Digest, error)
+	// Edit information about the list's instances.  The length of the passed-in slice must
+	// match the length of the list of instances which the list already contains, and every field
+	// must be specified.
```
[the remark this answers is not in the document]
flouthoc: I have removed the comment.

internal/manifest/list.go
```diff
@@ -65,6 +69,32 @@ type ListUpdate struct {
 	MediaType string
 }
 
+type ListOp int
+
+const (
+	ListOpInvalid ListOp = iota + 1
```
mtrmac, line 75: The idea of explicitly reserving `ListOpInvalid` is that is the _zero_ element, not `1`. We don’t need to treat 1 specially. With `listOpInvalid = iota`, callers that forget to initialize the field would get a known-invalid value.

Also this enum value can be private; we make no external promises about what using it would do.

mtrmac (commented):
Collecting the earlier review comments:
- `ListEdit` with modifications irrelevant to matching should not sort
- Then make `UpdateInstances` a wrapper over `EditInstances`
- (`x/exp/slices.SortStableFunc` would be a tiny bit more concise than `sort.SliceStable`)

flouthoc:
Updated

> (x/exp/slices.SortStableFunc would be a tiny bit more concise than sort.SliceStable)

I just avoided this part cause `x/exp` mentioned that all functions are experimental/deprecated and they can be removed anytime.

internal/manifest/list.go
flouthoc: Done.

internal/manifest/docker_schema2_list.go
```diff
@@ -44,6 +44,15 @@ func (list *Schema2ListPublic) MIMEType() string {
 	return list.MediaType
 }
 
+func (list *Schema2ListPublic) descriptor(instanceDigest digest.Digest) *Schema2ManifestDescriptor {
```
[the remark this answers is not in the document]
flouthoc: `x/exp` mentions that functions are experimental or deprecated and they can be removed without any warnings, hence I chose to return index from `descriptorIndex`.

internal/manifest/docker_schema2_list.go
```diff
@@ -69,23 +78,55 @@ func (list *Schema2ListPublic) Instance(instanceDigest digest.Digest) (ListUpdat
...
+	for i, editInstance := range editInstances {
+		switch editInstance.ListOperation {
+		case listOpUpdate:
+			if err := editInstance.UpdateOldDigest.Validate(); err != nil {
+				return fmt.Errorf("Schema2List.EditInstances: Attempting to update %s which is an invalid digest: %w", editInstance.UpdateOldDigest, err)
+			}
+			if err := editInstance.UpdateDigest.Validate(); err != nil {
+				return fmt.Errorf("Schema2List.EditInstances: Modified digest %s is an invalid digest: %w", editInstance.UpdateDigest, err)
+			}
+			instanceToBeModifiedIdx := index.descriptorIndex(editInstance.UpdateOldDigest)
```
mtrmac, line 101: (Absolutely non-blocking: four words for `instanceToBeModifiedIdx` seems a bit excessive. `targetIndex`?)

internal/manifest/docker_schema2_list.go
```diff
@@ -69,23 +78,55 @@ func (list *Schema2ListPublic) Instance(instanceDigest digest.Digest) (ListUpdat
...
-		if updates[i].Size < 0 {
-			return fmt.Errorf("update %d of %d passed to Schema2List.UpdateInstances had an invalid size (%d)", i+1, len(updates), updates[i].Size)
-		}
-		list.Manifests[i].Size = updates[i].Size
-		if updates[i].MediaType == "" {
-			return fmt.Errorf("update %d of %d passed to Schema2List.UpdateInstances had no media type (was %q)", i+1, len(updates), list.Manifests[i].MediaType)
+	return index.EditInstances(editInstances)
+}
+
+func (index *Schema2ListPublic) EditInstances(editInstances []ListEdit) error {
```
mtrmac, line 90: `EditInstances` should not be a public method on `*ListPublic`, at least for now; compare how `chooseInstance` is defined and exposed.

internal/manifest/list.go
```diff
@@ -65,6 +69,32 @@ type ListUpdate struct {
 	MediaType string
 }
 
+type ListOp int
+
+const (
+	listOpInvalid ListOp = iota
+	listOpAdd
+	listOpUpdate
```
mtrmac, line 77: Uh, these two (unlike `listOpInvalid`) must be public for the new functionality to be usable at all.

internal/manifest/docker_schema2_list.go
```diff
@@ -69,23 +78,55 @@ func (list *Schema2ListPublic) Instance(instanceDigest digest.Digest) (ListUpdat
...
+					OSFeatures:   editInstance.AddPlatform.OSFeatures,
+					Variant:      editInstance.AddPlatform.Variant,
+				},
+			}
+			addedEntries = append(addedEntries, addInstance)
+		default:
+			return fmt.Errorf("internal error: invalid operation: %d", editInstance.ListOperation)
 		}
-		list.Manifests[i].MediaType = updates[i].MediaType
+		index.Manifests = append(index.Manifests, addedEntries...)
```
mtrmac, line 129: (I’d prefer the handling of `addedEntries` here to directly correspond to the OCI handling, whatever that will be.)

internal/manifest/oci_index.go
```diff
@@ -64,22 +74,52 @@ func (index *OCI1IndexPublic) Instance(instanceDigest digest.Digest) (ListUpdate
...
+			addedEntries = append(addedEntries, imgspecv1.Descriptor{
+				MediaType:   editInstance.AddMediaType,
+				Size:        editInstance.AddSize,
+				Digest:      editInstance.AddDigest,
+				Platform:    editInstance.AddPlatform,
+				Annotations: editInstance.AddAnnotations})
+			index.Manifests = append(index.Manifests, addedEntries...)
+			sort.SliceStable(index.Manifests, func(i, j int) bool {
+				return !instanceIsZstd(index.Manifests[i]) && instanceIsZstd(index.Manifests[j])
+			})
```
mtrmac, line 119: Sorting on every add does work, but it ends up being O(N^2 long N) instead of O(N log N), unnecessarily.

Just collect `addedEntries` while processing edits, and at the end, if `len(addedEntries) != 0`, actually append and sort once.

(Adding `addedEntries` at the very end also has the benefit of failing on a request to update a just-added entry, a little bit of extra validation.)

internal/manifest/docker_schema2_list.go
```diff
@@ -69,23 +78,55 @@ func (list *Schema2ListPublic) Instance(instanceDigest digest.Digest) (ListUpdat
...
 // UpdateInstances updates the sizes, digests, and media types of the manifests
 // which the list catalogs.
-func (list *Schema2ListPublic) UpdateInstances(updates []ListUpdate) error {
-	if len(updates) != len(list.Manifests) {
-		return fmt.Errorf("incorrect number of update entries passed to Schema2List.UpdateInstances: expected %d, got %d", len(list.Manifests), len(updates))
+func (index *Schema2ListPublic) UpdateInstances(updates []ListUpdate) error {
+	editInstances := []ListEdit{}
+	for i, instance := range updates {
+		editInstance := ListEdit{UpdateOldDigest: index.Manifests[i].Digest, UpdateDigest: instance.Digest, UpdateSize: instance.Size, UpdateMediaType: instance.MediaType, ListOperation: listOpUpdate}
+		editInstances = append(editInstances, editInstance)
```
mtrmac, line 85: One line per field, in the order of the declaration (i.e. with `ListOperation` going first), please.

(Possibly just
```go
… = append(editInstances, ListEdit{
  …
}
```
without the extra variable.)

mtrmac (commented):
Thanks, looks pretty good!

Some tests — basic smoke tests, and something for the “no changes do not sort” aspect and for Zstd prioritization, would be nice.

flouthoc:
Added a test with `zstdpriority` check for `OCI1Index`.

internal/manifest/docker_schema2_list.go
flouthoc: We don't add zstd annotations to docker schema right, hence not sorting here.

internal/manifest/oci_index.go
flouthoc: done.

internal/manifest/oci_index_test.go
```diff
@@ -50,6 +51,61 @@ func TestOCI1IndexFromManifest(t *testing.T) {
...
+		ListOperation:   ListOpUpdate})
+	err = list.EditInstances(editInstances)
+	require.NoError(t, err)
+	assert.Equal(t, list.Instances()[0], editInstances[0].UpdateDigest, "editInstanceTest")
+	instance, err := list.Instance(list.Instances()[0])
+	require.NoError(t, err)
+	assert.Equal(t, instance.MediaType, "something", "editInstanceTest")
+	assert.Equal(t, instance.Size, int64(32), "editInstanceTest")
+
+	// Verfiy correct zstd sorting
```
mtrmac, line 74: Please start with a fresh
```go
	list, err = ListFromBlob(validManifest, GuessMIMEType(validManifest))
	require.NoError(t, err)
```

here, so that the data from this operation is unaffected by the previous test.

mtrmac (commented):
The schema2 part is incorrect now.

internal/manifest/docker_schema2_list.go
mtrmac: You’re right, the two would differ in sorting.

But the rest of the logic can directly correspond — see how the schema2 variant now turned out to be incorrect.

Having the two implementations very similar decreases the risk of such problems, and generally allows future maintenance to be done with half the analysis.

mtrmac:
> > (x/exp/slices.SortStableFunc would be a tiny bit more concise than sort.SliceStable)
>
> I just avoided this part cause `x/exp` mentioned that all functions are experimental/deprecated and they can be removed anytime.

See https://tip.golang.org/doc/go1.21 / https://pkg.go.dev/slices@master — it is coming in Go 1.21.

*shrug*, I can update that afterwards.

internal/manifest/docker_schema2_list.go
flouthoc: Fixed.

internal/manifest/oci_index_test.go
flouthoc: Done.

internal/manifest/oci_index_test.go
```diff
@@ -50,6 +51,75 @@ func TestOCI1IndexFromManifest(t *testing.T) {
...
+	require.NoError(t, err)
+
+	expectedDigests := list.Instances()
+	expectedDigests[0] = editInstances[0].UpdateDigest
+	// order of old elements must remain same.
+	assert.Equal(t, list.Instances(), expectedDigests)
+
+	instance, err := list.Instance(list.Instances()[0])
+	// We only edited the elements so list size must remain same.
+	assert.Equal(t, listSize, len(list.Instances()))
```
mtrmac, line 78: (Non-blocking: This seems unnecessary; the `assert.Equal(… list.Instances())` also implies a length comparison.)

internal/manifest/oci_index_test.go
```diff
@@ -50,6 +51,75 @@ func TestOCI1IndexFromManifest(t *testing.T) {
...
+	require.NoError(t, err)
```
mtrmac, line 79: Please keep the `err` check immediately after the call that produces `err`; don’t interleave two checks.

internal/manifest/docker_schema2_list_test.go
```diff
@@ -27,6 +30,68 @@ func TestSchema2ListPublicFromManifest(t *testing.T) {
...
+		AddDigest:     "sha256:cccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc",
+		AddSize:       32,
+		AddMediaType:  "application/vnd.oci.image.manifest.v1+json",
+		AddPlatform:   &imgspecv1.Platform{Architecture: "amd64", OS: "linux", OSFeatures: []string{"sse4"}},
+		ListOperation: ListOpAdd})
+	err = list.EditInstances(editInstances)
+	require.NoError(t, err)
+
+	// Verify list size with two newly added instances
+	assert.Equal(t, len(list.Instances()), oldListSize+2)
```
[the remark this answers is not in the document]
flouthoc: Done.

internal/manifest/docker_schema2_list_test.go
```diff
@@ -27,6 +30,68 @@ func TestSchema2ListPublicFromManifest(t *testing.T) {
...
+
+	// Add new elements to the end of old list to maintain order
+	oldListOrder = append(oldListOrder, digest.Digest("sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"))
```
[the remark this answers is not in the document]
flouthoc: Done.

internal/manifest/oci_index_test.go
```diff
@@ -50,6 +51,75 @@ func TestOCI1IndexFromManifest(t *testing.T) {
...
+		UpdateOldDigest: list.Instances()[0],
+		UpdateDigest:    "sha256:bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb",
+		UpdateSize:      32,
+		UpdateMediaType: "something",
+		ListOperation:   ListOpUpdate})
+	err = list.EditInstances(editInstances)
+	require.NoError(t, err)
+
+	expectedDigests := list.Instances()
+	expectedDigests[0] = editInstances[0].UpdateDigest
```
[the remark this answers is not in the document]
flouthoc: Done.

flouthoc:
@mtrmac PTAL again

mtrmac (approved):
Thanks!

PR #1902, opened 2023-03-30
Allow using cosign-generated private keys with a "SIGSTORE" type

mtrmac:
Don't (yet) generate them, no released version of cosign can work with them.

Compare Cosign’s change: https://github.com/sigstore/cosign/pull/2735


vrothberg (approved):
LGTM

PR #1903, opened 2023-03-30
Update, and support, Fulcio v1.2.0

mtrmac:
v1.2.0 deprecates an existing certificate extension, and introduces a new one. So, support that.

Tested _only_ using unit tests, not against a real certificate; at least the public Fulcio server does not yet generate the new extension.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,7 +25,7 @@ require (
 	github.com/opencontainers/selinux v1.11.0
 	github.com/ostreedev/ostree-go v0.0.0-20210805093236-719684c64e4f
 	github.com/proglottis/gpgme v0.1.3
-	github.com/sigstore/fulcio v1.1.0
+	github.com/sigstore/fulcio v1.2.0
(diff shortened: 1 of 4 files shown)


vrothberg (approved):
LGTM

Nice work splitting the commits up so nicely

TomSweeneyRedHat:
LGTM, do we need to update QE on test methodologies for these changes?

mtrmac:
I don’t think that setting up two versions of the Fulcio server just to test this is quite worth it — we can test this in a small isolated unit test, and I think that would be just as representative.

(Admittedly we don’t _have_ exactly that unit test, with two versions of a production-Fulcio-generated certificate recorded in the repo and testing that we accept both. But we are quite close to that; if we should track QE improvements, I think an issue here in c/image for improving the unit test would be better than bothering QE with this.)

TomSweeneyRedHat:
SGTM, thx @mtrmac

PR #1904, opened 2023-03-30
Add ability to verify a signature with a set of fingerprints

Jamstah:
To support https://github.com/containers/skopeo/pull/1950

### DIFF
--- a/signature/docker.go
+++ b/signature/docker.go
@@ -11,6 +11,7 @@ import (
 	"github.com/containers/image/v5/manifest"
 	"github.com/containers/image/v5/signature/internal"
 	"github.com/opencontainers/go-digest"
+	"golang.org/x/exp/slices"
 )
 
 // SignOptions includes optional parameters for signing container images.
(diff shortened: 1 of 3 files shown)


Jamstah:
I should probably make it return the selected fingerprint. If I don't do it now it's hard to change the API later.

signature/docker.go
```diff
@@ -49,16 +50,25 @@ func SignDockerManifest(m []byte, dockerReference string, mech SigningMechanism,
 // VerifyDockerManifestSignature checks that unverifiedSignature uses expectedKeyIdentity to sign unverifiedManifest as expectedDockerReference,
 // using mech.
 func VerifyDockerManifestSignature(unverifiedSignature, unverifiedManifest []byte,
-	expectedDockerReference string, mech SigningMechanism, expectedKeyIdentity string) (*Signature, error) {
+	expectedDockerReference string, mech SigningMechanism, expectedKeyIdentity string) (sig *Signature, err error) {
```
mtrmac, line 53: Don’t use named return values please; they might hide errors (e.g. failing to set `err` on an error path), and interact in surprising ways with `defer` ( https://github.com/containers/storage/pull/1483#discussion_r1154252787  ).

(Yes, paranoia is rampant throughout the `signature` subpackage…)

signature/docker.go
```diff
@@ -49,16 +50,25 @@ func SignDockerManifest(m []byte, dockerReference string, mech SigningMechanism,
...
+	sig, _, err = VerifyDockerManifestSignatureUsingKeyIdentityList(unverifiedSignature, unverifiedManifest, expectedDockerReference, mech, []string{expectedKeyIdentity})
+	return
+}
+
+// VerifyDockerManifestSignatureUsingKeyIdentityList checks that unverifiedSignature uses one of the expectedKeyIdentities
+// to sign unverifiedManifest as expectedDockerReference, using mech.
+func VerifyDockerManifestSignatureUsingKeyIdentityList(unverifiedSignature, unverifiedManifest []byte,
```
mtrmac, line 60: `VerifyImageManifest`… please, the preferred terminology has shifted since the previous function was added.

signature/docker.go
```diff
@@ -49,16 +50,25 @@ func SignDockerManifest(m []byte, dockerReference string, mech SigningMechanism,
...
+	expectedDockerReference string, mech SigningMechanism, expectedKeyIdentities []string) (sig *Signature, matchedKeyIdentity string, err error) {
```
mtrmac, line 61: No named return values, please. Then, ideally, document the “matched key identity” value in the doc comment above.

mtrmac (commented):
Thanks! Looks good, just a few nits.

signature/docker.go
Jamstah: Should I duplicate/deprecate the existing method signature too for good measure, or are you happy for them to mismatch?

mtrmac: I think leaving the old one as is is fine — almost everyone should use `signature.Policy` anyway.

signature/docker_test.go
```diff
@@ -160,3 +160,77 @@ func TestVerifyDockerManifestSignature(t *testing.T) {
 	assert.Error(t, err)
 	assert.Nil(t, sig)
 }
+
+func TestVerifyDockerManifestSignatureUsingKeyIdentityList(t *testing.T) {
```
mtrmac, line 164:
```suggestion
func TestVerifyImageManifestSignatureUsingKeyIdentityList(t *testing.T) {
```

mtrmac (commented):
LGTM otherwise.

signature/docker_test.go
Jamstah: Nice spot :)

Jamstah:
OK, I think its done :)

TomSweeneyRedHat:
LGTM

mtrmac (approved):
Thanks!

Issue #1919, opened 2023-04-12
Copy:  error when copy an image with encryption  to the local directory 


ChengyuZhu6:
When copy an image with encryption  to the format of local directory,  there is an error:
```
Internal error: no candidate MIME types
```
And this was reported in https://github.com/containers/skopeo/issues/1971

mtrmac:
#1930 should fix this.

PR #1920, opened 2023-04-12
copy/manifest: skip conversion if destSupportedManifestMIMETypes=nil

ChengyuZhu6:
As described in https://github.com/containers/image/issues/1919,  the `destSupportedManifestMIMETypes`  is empty when the destination format is local directory, which results in `prioritizedTypes.list` is empty and causing the error: `no candidate MIME types`. Maybe we can skip the conversion if the `destSupportedManifestMIMETypes` is empty.

Fixed: #1919 

Signed-off-by: ChengyuZhu6 [chengyu.zhu@intel.com](mailto:chengyu.zhu@intel.com)

### DIFF
--- a/copy/manifest.go
+++ b/copy/manifest.go
@@ -76,7 +76,7 @@ func determineManifestConversion(in determineManifestConversionInputs) (manifest
 		destSupportedManifestMIMETypes = []string{in.forceManifestMIMEType}
 	}
 
-	if len(destSupportedManifestMIMETypes) == 0 && (!in.requiresOCIEncryption || manifest.MIMETypeSupportsEncryption(srcType)) {
+	if len(destSupportedManifestMIMETypes) == 0 {
 		return manifestConversionPlan{ // Anything goes; just use the original as is, do not try any conversions.
 			preferredMIMEType:       srcType,
 			otherMIMETypeCandidates: []string{},


mtrmac (changes requested):
Thanks for the PR.

No, that check, or _some_ extra logic, is necessary. This check prevents trying to use encryption in image formats that don’t support it. That’s, AFAICS, exactly the case in https://github.com/containers/skopeo/issues/1971 : the source image is Docker schema2, which just can’t be encrypted without conversion (and AFAICS what would happen with this PR is that the layers would be encrypted, but the required metadata would not be saved anywhere, making the image unusable).

In this situation, the logic needs to trigger conversion to OCI.

---

Also, please make sure that any added cases have test coverage in `TestDetermineManifestConversion`

ChengyuZhu6:
> Thanks for the PR.
> 
> No, that check, or _some_ extra logic, is necessary. This check prevents trying to use encryption in image formats that don’t support it. That’s, AFAICS, exactly the case in [containers/skopeo#1971](https://github.com/containers/skopeo/issues/1971) : the source image is Docker schema2, which just can’t be encrypted without conversion (and AFAICS what would happen with this PR is that the layers would be encrypted, but the required metadata would not be saved anywhere, making the image unusable).
> 
> In this situation, the logic needs to trigger conversion to OCI.
> 
> Also, please make sure that any added cases have test coverage in `TestDetermineManifestConversion`

 Hi [mtrmac](https://github.com/mtrmac), thank you for taking the time to review your code and provide feedback. I was wrong and I ignored the problem with metadata. But when the destination accepts all manifest formats, maybe we can set a default format to avoid the problem of not being able to trigger conversion, such as OCI ?

mtrmac:
`dirImageDestination` is correctly saying that it supports all imaginable schemas; it shouldn’t need to incorrectly say that it only supports a subset.

In this case, I think `DetermineManifestConversion` should decide to trigger a conversion.

ChengyuZhu6:
> `dirImageDestination` is correctly saying that it supports all imaginable schemas; it shouldn’t need to incorrectly say that it only supports a subset.
> 
> In this case, I think `DetermineManifestConversion` should decide to trigger a conversion.

I'm sorry I didn't express my thoughts clearly. I mean that after checking that the manifest conversion is required with decryption https://github.com/containers/image/blob/main/copy/manifest.go#:~:text=if%20len(destSupportedManifestMIMETypes)%20%3D%3D%200%20%26%26%20(!in.requiresOCIEncryption%20%7C%7C%20manifest.MIMETypeSupportsEncryption(srcType))%20%7B,  if `destSupportedManifestMIMETypes` is still empty(This only happens when the destination mimtype is `dirImageDestination`), maybe it should be assigned a default mimetype, such as oci

mtrmac:
#1930 should fix this. I’d appreciate an independent confirmation.

mtrmac:
#1930 was merged.

PR #1922, opened 2023-04-14
Don't completely silently ignore non-OCI manifests in OCI layouts

mtrmac:
Instead, report that we did find a match and that it is not supported.

Partially fixes https://github.com/containers/skopeo/issues/1975 .

Also, significantly improve unit tests of that code.


rhatdan:
LGTM
@vrothberg PTAL

vrothberg (approved):
LGTM

PR #1923, opened 2023-04-14
Simplify the tarball: transport

mtrmac:
Track layer data in a single `map[digest]struct` instead of 6 separate arrays, some completely unused, and other fields.

Also simplify how that data is created, and fix pedantically-invalid uses of `bytes.NewBuffer`.

(Motivated by https://github.com/containers/podman/issues/18193 but this doesn’t do anything to actually fix it.)

### DIFF
--- a/tarball/tarball_src.go
+++ b/tarball/tarball_src.go
@@ -67,7 +67,7 @@ func (r *tarballReference) NewImageSource(ctx context.Context, sys *types.System
 		} else {
 			file, err := os.Open(filename)
 			if err != nil {
-				return nil, fmt.Errorf("error opening %q for reading: %w", filename, err)
+				return nil, err
 			}
 			defer file.Close()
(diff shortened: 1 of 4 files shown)


tarball/tarball_src.go
```diff
@@ -57,7 +55,7 @@ func (r *tarballReference) NewImageSource(ctx context.Context, sys *types.System
 			blobTime = time.Now()
 			reader = bytes.NewReader(r.stdin)
 		} else {
-			file, err = os.Open(filename)
+			file, err := os.Open(filename)
 			if err != nil {
 				return nil, fmt.Errorf("error opening %q for reading: %w", filename, err)
```
rhatdan, line 70: This is a stutter, can you just return err.

vrothberg (approved):
LGTM

Issue #1768, opened 2022-12-17
pull by digest broken on aes encrypted image pushed to dockerhub


vans163:

mtrmac:
Note: https://github.com/containers/image/pull/1930#discussion_r1177040737

PR #1930, opened 2023-04-25
Fix conversion determination when encrypting

mtrmac:
- When encrypting, actually make a choice even if the destination does not care about manifest formats
- Don't try converting to formats that don't support encryption at all
- Fail with a readable error message instead of "internal error"  if we filter out all candidates.
- Explicitly fail encryption/decryption if we can't change the manifest, instead of failing later with an internal error

See individual commit messages for a bit more detail.

### DIFF
--- a/copy/encryption.go
+++ b/copy/encryption.go
@@ -33,12 +33,15 @@ type bpDecryptionStepData struct {
 // blobPipelineDecryptionStep updates *stream to decrypt if, it necessary.
 // srcInfo is only used for error messages.
 // Returns data for other steps; the caller should eventually use updateCryptoOperation.
-func (c *copier) blobPipelineDecryptionStep(stream *sourceStream, srcInfo types.BlobInfo) (*bpDecryptionStepData, error) {
-	if isOciEncrypted(stream.info.MediaType) && c.ociDecryptConfig != nil {
+func (ic *imageCopier) blobPipelineDecryptionStep(stream *sourceStream, srcInfo types.BlobInfo) (*bpDecryptionStepData, error) {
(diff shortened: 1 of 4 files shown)


copy/encryption.go
```diff
@@ -33,28 +33,33 @@ type bpDecryptionStepData struct {
...
-		})
+func (ic *imageCopier) blobPipelineDecryptionStep(stream *sourceStream, srcInfo types.BlobInfo) (*bpDecryptionStepData, error) {
+	if !isOciEncrypted(stream.info.MediaType) || ic.c.ociDecryptConfig == nil {
 		return &bpDecryptionStepData{
-			decrypting: true,
+			decrypting: false,
 		}, nil
 	}
+
+	if ic.cannotModifyManifestReason != "" {
```
mtrmac, line 43: I’m not immediately sure how this interacts with `podman pull` of digest references, I need to test that. (It seems to me that that never worked, in which case this is, I guess, good enough…)

mtrmac:
This also replaces #1920, and partially #1002.

copy/encryption.go
mtrmac: Actually, no need to test, this is tracked as https://github.com/containers/image/issues/1768 . I’ll just add a comment here.

mtrmac: … not even a comment; #1768 would probably be cleanest to fix by recording the original manifest, and the updated manifest, separately; and in that case `cannotModifyManifestReason` would not be set.

rhatdan (approved):
LGTM

vrothberg (approved):
LGTM

PR #1938, opened 2023-05-02
copy.Options: add an InstancePlatforms field

nalind:
Add an `InstancePlatforms` field to the `copy.Options` structure.  When asked to copy only a specific subset of instances, combine the `Instances` list with the instances which match the platforms given in the `InstancePlatforms` list, returning an error if we're unable to find an instance which matches any of the `InstancePlatforms`.  I think this is a necessary step in implementing RUN-1803 in skopeo cleanly.

### DIFF
--- a/copy/multiple.go
+++ b/copy/multiple.go
@@ -10,11 +10,13 @@ import (
 	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/image/v5/internal/image"
 	internalManifest "github.com/containers/image/v5/internal/manifest"
+	"github.com/containers/image/v5/internal/set"
 	"github.com/containers/image/v5/manifest"
 	"github.com/containers/image/v5/signature"
+	"github.com/containers/image/v5/types"
+	digest "github.com/opencontainers/go-digest"
 	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/sirupsen/logrus"
-	"golang.org/x/exp/slices"
 )
 
 // copyMultipleImages copies some or all of an image list's instances, using
@@ -89,15 +91,19 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
 	// Copy each image, or just the ones we want to copy, in turn.
 	instanceDigests := updatedList.Instances()
 	imagesToCopy := len(instanceDigests)
+	var specificImages *set.Set[digest.Digest]
 	if options.ImageListSelection == CopySpecificImages {
-		imagesToCopy = len(options.Instances)
+		if specificImages, err = determineSpecificImages(options, updatedList); err != nil {
+			return nil, err
+		}
+		imagesToCopy = specificImages.Size()
 	}
 	c.Printf("Copying %d of %d images in list\n", imagesToCopy, len(instanceDigests))
 	updates := make([]manifest.ListUpdate, len(instanceDigests))
 	instancesCopied := 0
 	for i, instanceDigest := range instanceDigests {
 		if options.ImageListSelection == CopySpecificImages &&
-			!slices.Contains(options.Instances, instanceDigest) {
+			!specificImages.Contains(instanceDigest) {
 			update, err := updatedList.Instance(instanceDigest)
 			if err != nil {
 				return nil, err
@@ -195,4 +201,31 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
 	}
 
 	return manifestList, nil
+}
+
+// determineSpecificImages returns a set of images to copy based on the
+// Instances and InstancePlatforms fields of the passed-in options structure
+func determineSpecificImages(options *Options, updatedList internalManifest.List) (*set.Set[digest.Digest], error) {
+	// Start with the instances that were listed by digest.
+	specificImages := set.New[digest.Digest]()
+	for _, instanceDigest := range options.Instances {
+		specificImages.Add(instanceDigest)
+	}
+	if len(options.InstancePlatforms) > 0 {
+		// Choose the best match for each platform we were asked to
+		// also copy, and add it to the set of instances to copy.
+		for _, platform := range options.InstancePlatforms {
+			platformContext := types.SystemContext{
+				OSChoice:           platform.OS,
+				ArchitectureChoice: platform.Architecture,
+				VariantChoice:      platform.Variant,
+			}
+			instanceDigest, err := updatedList.ChooseInstanceByCompression(&platformContext, options.PreferGzipInstances)
+			if err != nil {
+				return nil, fmt.Errorf("While choosing the instance for platform spec %v: %w", platform, err)
+			}
+			specificImages.Add(instanceDigest)
+		}
+	}
+	return specificImages, nil
 }
--- a/copy/multiple_test.go
+++ b/copy/multiple_test.go
@@ -1,0 +1,170 @@
+package copy
+
+import (
+	"io/ioutil"
+	"strings"
+	"testing"
+
+	internalManifest "github.com/containers/image/v5/internal/manifest"
+	digest "github.com/opencontainers/go-digest"
+	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestDetermineSpecificImages(t *testing.T) {
+	testCases := []struct {
+		id                  string
+		fixture             string
+		instanceDigests     []digest.Digest
+		instancePlatforms   []imgspecv1.Platform
+		expected            []digest.Digest
+		expectedErrIncludes string
+	}{
+		{
+			id:      "no inputs no outputs",
+			fixture: "../manifest/fixtures/v2list.manifest.json",
+		},
+		{
+			id:      "instances only out of order",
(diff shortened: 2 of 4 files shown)


copy/multiple.go
```diff
@@ -87,17 +89,43 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
...
+				VariantChoice:      platform.Variant,
+			}
+			instanceDigest, err := updatedList.ChooseInstance(&platformContext)
+			if err != nil {
+				return nil, fmt.Errorf("While choosing the instance for platform spec %q: %w", platform, err)
+			}
+			instanceDigestMap[instanceDigest] = struct{}{}
+		}
+		// Deduplicate the set of digests.
+		optionsInstances = nil
```
mtrmac, line 114: It’s might be a bit misleading to name this `optionsInstances` when it depends on the other option as well.

Arguably `optionsInstances` includes `options.Instances*`, but it seems simple to avoid this uncertainty.

Maybe `[instance]digestsToCopy`? That would also say what the data is for.

copy/multiple.go
```diff
@@ -87,17 +89,43 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
...
-		imagesToCopy = len(options.Instances)
+		imagesToCopy = len(optionsInstances)
 	}
 	c.Printf("Copying %d of %d images in list\n", imagesToCopy, len(instanceDigests))
 	updates := make([]manifest.ListUpdate, len(instanceDigests))
 	instancesCopied := 0
 	for i, instanceDigest := range instanceDigests {
 		if options.ImageListSelection == CopySpecificImages &&
-			!slices.Contains(options.Instances, instanceDigest) {
+			!slices.Contains(optionsInstances, instanceDigest) {
```
mtrmac, line 128: This would ideally (at least in the asymptotic sense) be a hash table lookup, instead of an O(N) scan.

Currently we turn a list into a map, and then a map into a list; that second conversion could be just skipped, building a `digestsToCopy set.Set[digest.Digest]` from the combination of `options.Instances` and `options.InstancePlatforms`, and consuming it here directly. (We’d need a `set.Size()` added.)

copy/multiple.go
```diff
@@ -87,17 +89,43 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
 	}
 
 	// Copy each image, or just the ones we want to copy, in turn.
+	optionsInstances := options.Instances
 	instanceDigests := updatedList.Instances()
```
mtrmac, line 93: `instanceDigests` is, AFAICS, not used inside the added code, so please set it just before the first user.

---

All of the new code computing the `optionsInstances` data structure from `options.*` seems like a good candidate for a separate function; it is clearly “functional”, unlike the more imperative surrounding code, and potentially easy to unit-test if split. (I don’t think a unit test is blocking this PR.)

mtrmac (commented):
Thanks!

The implementation looks good overall.

WRT the API:

Looking at the example in https://github.com/containers/skopeo/issues/1935 , and the code in https://github.com/osbuild/osbuild/blob/b92168bc16d8251256ec34df938d616bca99cd22/sources/org.osbuild.skopeo#L105 , it seems that the primary use case revolves around the `CopySystemImage` choice, what https://github.com/containers/skopeo/issues/1935 calls `--multi-arch=sparse:[system]`. (Cc: @achilleas-k to keep me honest.)
 
How is that going to be represented? I’m a bit torn.
- One approach is to model it in this API explicitly (and therefore inherit the way `platform.WantedPlatforms` consumes `types.SystemContext`): seems more consistent, OTOH the API could be more complex, either with the members of `InstancePlatforms` some kind of sum type (`enum{ case system; case platform(platform) }`), or with an extra `InstanceForSystemPlatform` field. Not quite elegant.
- Another is to do this just in the Skopeo CLI. Probably simpler all around, OTOH I think it’s valuable to be fairly strict about Skopeo features all being available to other callers of c/image.

Ideally I think _some_ kind of c/image API this would be better, assuming we can design a practical one that doesn’t grow to an absurd size.

This PR doesn’t say, so I assume it plans to use the Skopeo approach? Or a separate c/image PR?

---

(For the record, and mostly because I wrote this for the initial implementation which was destructive. Not a concern for the code as is.

- There are two kinds of users in this area
  - Some want to keep the original manifest list, and create a sparse copy (with some instances missing), so that the digest references are not broken — that’s what RUN-1803 wants, some registries reject such lists.
  - Some want to strip the modify the manifest list, so that the copy is accepted by all registries. That’s e.g. #1707 .
  
The API to choose instances should (for the future where the “strip” variant exists) allow both, with a single orthogonal switch between the two behaviors. That’s probably going to happen “by default” in many implementations (the API of `EditInstances` kind of forces that on us :) ), just something to keep an eye on.)

---

Cc: @flouthoc for the interaction with the Zstd work — I think this should be mostly orthogonal, just a heads-up about potential merge conflicts.

TomSweeneyRedHat:
The VM img.keepalive test had failed, it look like a google instance issue.  I've restarted.

mtrmac:
> This PR doesn’t say, so I assume it plans to use the Skopeo approach?

https://github.com/containers/skopeo/pull/1987 implements that in Skopeo.

mtrmac:
Note to self: the `--override-arch=amd64 --multi-arch=sparse:system` case in https://github.com/containers/skopeo/pull/1987 needs a more detailed look.

mtrmac:
Oh… The decision on _what to filter on_ is actually non-obvious. Do we want platform triplets, or separately OSes and architectures? All three? https://github.com/containers/skopeo/pull/1987#discussion_r1183115878 for a starting point, to continue tomorrow…

copy/multiple.go
```diff
@@ -88,6 +90,31 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
 
 	// Copy each image, or just the ones we want to copy, in turn.
 	instanceDigests := updatedList.Instances()
+	if len(options.InstancePlatforms) > 0 {
+		// Build a map for helping us avoid duplication of instance digests.
+		instanceDigestMap := make(map[digest.Digest]struct{})
```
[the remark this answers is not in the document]
nalind: Sure, changing it.

copy/multiple.go
```diff
@@ -88,6 +90,31 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
...
+			if err != nil {
+				return nil, fmt.Errorf("No instance in list matched platform spec %v", platform)
+			}
+			instanceDigestMap[instanceDigest] = struct{}{}
+		}
+		// Deduplicate the set of digests.
+		instanceDigests = nil
+		for instanceDigest := range instanceDigestMap {
+			instanceDigests = append(instanceDigests, instanceDigest)
+		}
```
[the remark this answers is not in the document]
nalind: Yeah, I think preserving the order would have been desirable even if it wasn't a requirement, and using the set `Contains()` method to decided whether to copy an instance while we're going through that should make it trivial.

copy/multiple.go
nalind: The set is tied to the `CopySpecificImages` choice, so how about `specificImages`?

copy/multiple.go
nalind: `len()` of the set's `Values()` will work until then.  I wouldn't complain about a `Size()` or `Len()`, though.

mtrmac: This PR can add the `Size`/`Len`, it’s an internal subpackage we can change with no risk.

copy/multiple.go
nalind: Split it off into a `determineSpecificImages()` function.

copy/multiple.go
nalind: You suggested `Size()` first, so I've added it as that.

mtrmac (commented):
Implementation of the API LGTM.

Outstanding design questions:
- How to represent `system`
- Do we want the user to specify whole platforms? Or just arch[+variant]?

copy/multiple_test.go
```diff
@@ -0,0 +1,195 @@
...
+				},
+			},
+			expectedErrIncludes: "no image found in manifest list for",
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.id, func(t *testing.T) {
+			listBlob, err := ioutil.ReadFile(tc.fixture)
+			if err != nil {
+				t.Fatalf("unexpected error reading fixture %q: %v", tc.fixture, err)
```
mtrmac, line 164: (Absolutely non-blocking: {`assert`,`require`}`.NoError(t, err)` is a bit more concise. This works just fine as is.)

flouthoc (approved):
LGTM

internal/set/set.go
```diff
@@ -44,3 +44,7 @@ func (s *Set[E]) Empty() bool {
 func (s *Set[E]) Values() []E {
 	return maps.Keys(s.m)
 }
+
+func (s Set[E]) Size() int {
```
mtrmac, line 48:
```suggestion
func (s *Set[E]) Size() int {
```
please.

The current version works just fine, but I managed to inattentively screw up the `Add`/`Delete` methods (https://github.com/containers/image/pull/1951 ), so using a pointer receiver everywhere would hopefully make it harder for me to make the same mistake again.

nalind: Changed it.

jankaluza:

PR #1954, opened 2023-05-11
Propagate `PolicyRequirementError` to API user

saschagrunert:
Wrapping the error in `copy/single.go` will make a simple check like this not work any more:

```go
var e signature.PolicyRequirementError
if errors.As(err, &e) { }
```

Means we now avoid wrapping the error with another static message.

Beside that, we now wrap the internal errors from
`req.isRunningImageAllowed` as `PolicyRequirementError` to let API consumers know that the signature verification failed.

### DIFF
--- a/copy/single.go
+++ b/copy/single.go
@@ -59,7 +59,7 @@ func (c *copier) copySingleImage(ctx context.Context, policyContext *signature.P
 	// (The multiImage check above only matches the MIME type, which we have received anyway.
 	// Actual parsing of anything should be deferred.)
 	if allowed, err := policyContext.IsRunningImageAllowed(ctx, unparsedImage); !allowed || err != nil { // Be paranoid and fail if either return value indicates so.
-		return nil, "", "", fmt.Errorf("Source image rejected: %w", err)
+		return nil, "", "", err
 	}
 	src, err := image.FromUnparsedImage(ctx, options.SourceCtx, unparsedImage)
 	if err != nil {
--- a/signature/policy_eval.go
+++ b/signature/policy_eval.go
@@ -253,7 +253,7 @@ func (pc *PolicyContext) GetSignaturesWithAcceptedAuthor(ctx context.Context, pu
 	return res, nil
 }
 
-// IsRunningImageAllowed returns true iff the policy allows running the image.
+// IsRunningImageAllowed returns true if the policy allows running the image.
 // If it returns false, err must be non-nil, and should be an PolicyRequirementError if evaluation
 // succeeded but the result was rejection.
 // WARNING: This validates signatures and the manifest, but does not download or validate the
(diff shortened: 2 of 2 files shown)


copy/single.go
```diff
@@ -59,7 +59,7 @@ func (c *copier) copySingleImage(ctx context.Context, policyContext *signature.P
 	// (The multiImage check above only matches the MIME type, which we have received anyway.
 	// Actual parsing of anything should be deferred.)
 	if allowed, err := policyContext.IsRunningImageAllowed(ctx, unparsedImage); !allowed || err != nil { // Be paranoid and fail if either return value indicates so.
-		return nil, "", "", fmt.Errorf("Source image rejected: %w", err)
+		return nil, "", "", err
```
mtrmac, line 62: Really? `fmt.Errorf("%w")` _only exists_ to be transparent to wrapping; how does that fail?

signature/policy_eval.go
```diff
@@ -253,7 +253,7 @@ func (pc *PolicyContext) GetSignaturesWithAcceptedAuthor(ctx context.Context, pu
 	return res, nil
 }
 
-// IsRunningImageAllowed returns true iff the policy allows running the image.
+// IsRunningImageAllowed returns true if the policy allows running the image.
```
mtrmac, line 256: https://en.wikipedia.org/wiki/If_and_only_if

signature/policy_eval.go
```diff
@@ -283,7 +283,7 @@ func (pc *PolicyContext) IsRunningImageAllowed(ctx context.Context, publicImage
 		allowed, err := req.isRunningImageAllowed(ctx, image)
 		if !allowed {
 			logrus.Debugf("Requirement %d: denied, done", reqNumber)
-			return false, err
+			return false, PolicyRequirementError(err.Error())
```
mtrmac, line 286: No; `PolicyRequirementError` means “the policy says you can’t do that”, which is not _all_ errors (e.g. “fetching signatures failed because the network is down” is explicitly not a `PolicyRequirementError`).

Maybe the documentation of that type needs to be improved.

mtrmac (changes requested):

copy/single.go
mtrmac:
```go
package main

import (
	"errors"
	"fmt"
)

// PolicyRequirementError is an explanatory text for rejecting a signature or an image.
type PolicyRequirementError string

func (err PolicyRequirementError) Error() string {
	return string(err)
}

func main() {
	err1 := PolicyRequirementError("1")
	var e1 PolicyRequirementError
	matched := errors.As(err1, &e1)
	fmt.Printf("Matched 1: %v => %v, %v (%#v)\n", err1, matched, e1, e1)

	err2 := fmt.Errorf("fmt-added: %w", err1)
	var e2 PolicyRequirementError
	matched = errors.As(err2, &e2)
	fmt.Printf("Matched 2: %v => %v, %v (%#v)\n", err2, matched, e2, e2)
}
```
shows the wrapping is transparent, AFAICS.

signature/policy_eval.go
saschagrunert: Makes sense, thank you for the clarification.

PR #1963, opened 2023-05-25
Stop having an opinion on TLS version

### LINKED ISSUE #1962: Can the server be upgraded to TLS1.2?
Since the TLS1.0 and TLS1.1 versions are old. Can the server be upgraded to TLS1.2? 
```
// this is cloned from docker/go-connections because upstream docker has changed
// it and make deps here fails otherwise.
// We'll drop this once we upgrade to docker 1.13.x deps.
func serverDefault() *tls.Config {
	return &tls.Config{
		// Avoid fallback to SSL protocols < TLS1.0
		MinVersion:               tls.VersionTLS10,
		PreferServerCipherSuites: true,
		CipherSuites:             tlsconfig.DefaultServerAcceptedCiphers,
	}
}
```
Upstream docker/go-connections has been upgraded.
https://github.com/docker/go-connections/commit/eed1c499cef34e358f4a10f8de1ce1b1a945556f#diff-2cce4edd4dc7effba397d7bbbcb4b4c39760daf4796ceb876823ae1a87ade022L73

mtrmac:
Instead, let Go manage the default; this effectively moves the version from TLS 1.0 to TLS 1.2 (because c/image requires Go 1.18 to compile).

(In the worst case, affected users can disable TLS; we don’t provide any other options.) Compare https://github.com/moby/moby/commit/626c097f8c29bf432e5465a961496bf41aa74cad updating to TLS 1.2; it is similarly not configurable. 

Fixes #1962 .

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -166,8 +166,6 @@ func newBearerTokenFromJSONBlob(blob []byte) (*bearerToken, error) {
 // We'll drop this once we upgrade to docker 1.13.x deps.
 func serverDefault() *tls.Config {
 	return &tls.Config{
-		// Avoid fallback to SSL protocols < TLS1.0
-		MinVersion:   tls.VersionTLS10,
 		CipherSuites: tlsconfig.DefaultServerAcceptedCiphers,
 	}
 }
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -161,15 +161,6 @@ func newBearerTokenFromJSONBlob(blob []byte) (*bearerToken, error) {
 	return token, nil
 }
 
-// this is cloned from docker/go-connections because upstream docker has changed
-// it and make deps here fails otherwise.
-// We'll drop this once we upgrade to docker 1.13.x deps.
-func serverDefault() *tls.Config {
-	return &tls.Config{
-		CipherSuites: tlsconfig.DefaultServerAcceptedCiphers,
-	}
-}
-
 // dockerCertDir returns a path to a directory to be consumed by tlsclientconfig.SetupCertificates() depending on ctx and hostPort.
 func dockerCertDir(sys *types.SystemContext, hostPort string) (string, error) {
 	if sys != nil && sys.DockerCertPath != "" {
@@ -252,7 +243,9 @@ func newDockerClient(sys *types.SystemContext, registry, reference string) (*doc
 	if registry == dockerHostname {
 		registry = dockerRegistry
 	}
-	tlsClientConfig := serverDefault()
+	tlsClientConfig := &tls.Config{
+		CipherSuites: tlsconfig.DefaultServerAcceptedCiphers,
+	}
 
 	// It is undefined whether the host[:port] string for dockerHostname should be dockerHostname or dockerRegistry,
 	// because docker/docker does not read the certs.d subdirectory at all in that case.  We use the user-visible
--- a/openshift/openshift-copies.go
+++ b/openshift/openshift-copies.go
@@ -957,8 +957,6 @@ func tlsConfigFor(c *restConfig) (*tls.Config, error) {
 	}
 
 	tlsConfig := &tls.Config{
-		// Change default from SSLv3 to TLSv1.0 (because of POODLE vulnerability)
-		MinVersion:         tls.VersionTLS10,
 		InsecureSkipVerify: c.Insecure,
 	}
 


mtrmac:
This allows compliance with https://datatracker.ietf.org/doc/html/rfc8996 .

---

I would rather prefer getting _entirely_ out of this business, in particular stopping to set `CipherSuites`. What that would mean:

Suite | Go 1.20 default | Go-connections server | Go-connections client
-- | -- | -- | --
ECDHE_ECDSA_WITH_AES_128_CBC_SHA | TRUE | TRUE | FALSE
ECDHE_ECDSA_WITH_AES_128_CBC_SHA256 | FALSE | FALSE | FALSE
ECDHE_ECDSA_WITH_AES_128_GCM_SHA256 | TRUE | TRUE | TRUE
ECDHE_ECDSA_WITH_AES_256_CBC_SHA | TRUE | TRUE | FALSE
ECDHE_ECDSA_WITH_AES_256_GCM_SHA384 | TRUE | TRUE | TRUE
ECDHE_ECDSA_WITH_CHACHA20_POLY1305 | TRUE | FALSE | FALSE
ECDHE_ECDSA_WITH_RC4_128_SHA | FALSE | FALSE | FALSE
ECDHE_RSA_WITH_3DES_EDE_CBC_SHA | TRUE | FALSE | FALSE
ECDHE_RSA_WITH_AES_128_CBC_SHA | TRUE | TRUE | FALSE
ECDHE_RSA_WITH_AES_128_CBC_SHA256 | FALSE | FALSE | FALSE
ECDHE_RSA_WITH_AES_128_GCM_SHA256 | TRUE | TRUE | TRUE
ECDHE_RSA_WITH_AES_256_CBC_SHA | TRUE | TRUE | FALSE
ECDHE_RSA_WITH_AES_256_GCM_SHA384 | TRUE | TRUE | TRUE
ECDHE_RSA_WITH_CHACHA20_POLY1305 | TRUE | FALSE | FALSE
ECDHE_RSA_WITH_RC4_128_SHA | FALSE | FALSE | FALSE
RSA_WITH_3DES_EDE_CBC_SHA | TRUE | FALSE | FALSE
RSA_WITH_AES_128_CBC_SHA | TRUE | FALSE | FALSE
RSA_WITH_AES_128_CBC_SHA256 | FALSE | FALSE | FALSE
RSA_WITH_AES_128_GCM_SHA256 | TRUE | FALSE | FALSE
RSA_WITH_AES_256_CBC_SHA | TRUE | FALSE | FALSE
RSA_WITH_AES_256_GCM_SHA384 | TRUE | FALSE | FALSE
RSA_WITH_RC4_128_SHA | FALSE | FALSE | FALSE

where “go-connections _server_” are the settings applicable for registry _clients_, both in Docker and currently c/image.

So, not specifying `CipherSuites` and using _only_ the upstream defaults would add the DJB suites, but also non-forward-secure RSA and SHA-1 … _if_ the server and client agree on it within a SHA-2-secured handshake. I don’t see a very strong reason to worry about this, but allowing currently-prohibited SHA-1, and 3DES seems rather awkward. Doing nothing and continuing to use the choices from `docker/go-connections` is the lazy low-risk option.

rhatdan:
LGTM
@vrothberg PTAL

TomSweeneyRedHat:
LGTM
but want a @vrothberg head nod before merge.

vrothberg:
I am OOO for two weeks.

mtrmac:
The primary risk is that we will almost certainly include this updated c/image in a Podman that will be included within the lifetime of an enterprise distribution; so that means starting to require TLS 1.2 during the life of that product. OTOH I imagine it’s not very different from the way we ship Go itself, and probably other software built with Go.

There _is_ some small chance we will be required to add a config option in a hurry, to let the user tweak the TLS requirement. I’d rather not add that option preemptively, though; it’s likely to be set once and not updated regularly… just like the code this PR proposes removing :)

rhatdan:
RHEL8 is about to go into maintenance mode, which should mean that it will no longer get major upgrades and only have back ported fixes. @TomSweeneyRedHat which release does this happen in?

What is the standard for RHEL8 at this point?

mtrmac:
https://gitlab.com/redhat-crypto/fedora-crypto-policies/-/blob/rhel8/policies/DEFAULT.pol?ref_type=heads suggests RHEL8 also requires TLS 1.2 by default, and AFAICT that’s been the case since 8.0.0. Compare also https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/securing_networks/planning-and-implementing-tls_securing-networks .

vrothberg (approved):
LGTM

rhatdan:
@mtrmac should the RFC be removed and this merged?

mtrmac:
Yes, let’s do that. 3 approvals is more than I could ask for.

vrothberg:
Thanks, @mtrmac !

PR #1968, opened 2023-05-29
WIP:  Only obtain a bearer token once at a time

mtrmac:
Currently, on pushes, we can start several concurrent layer pushes; each one will check for a bearer token in `tokenCache`, find none, and ask the server for one, and then write it into the cache.

So, we can hammer the server with 6 basically-concurrent token requests. That's unnecessary, slower than just asking once, and potentially might impact rate limiting heuristics.

Instead, serialize writes to a bearerToken so that we only have one request in flight at a time.

This does not apply to pulls, where the first request is for a manifest, which obtains a token, so subsequent concurrent layer pulls will not request a token again.

WIP: Clean up the debugging log entries.

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -760,20 +760,18 @@ func (c *dockerClient) obtainBearerToken(ctx context.Context, challenge challeng
 		token, inCache = c.tokenCache[cacheKey]
 	}()
 	if !inCache || time.Now().After(token.expirationTime) {
-		var (
-			t   *bearerToken
-			err error
-		)
+		token = &bearerToken{}
+
+		var err error
 		if c.auth.IdentityToken != "" {
-			t, err = c.getBearerTokenOAuth2(ctx, challenge, scopes)
+			err = c.getBearerTokenOAuth2(ctx, token, challenge, scopes)
 		} else {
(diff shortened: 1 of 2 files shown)


docker/docker_client.go
```diff
@@ -708,16 +659,94 @@ func (c *dockerClient) setupRequestAuth(req *http.Request, extraScope *authScope
...
+	logrus.Debugf("REMOVE: Checking token cache for key %q", cacheKey)
+	var token *bearerToken
+	func() { // A scope for defer
+		c.tokenCacheLock.Lock()
+		defer c.tokenCacheLock.Unlock()
+		t, ok := c.tokenCache[cacheKey]
+		if ok {
+			token = t
+			logrus.Debugf("REMOVE: Found token cache for key %q, getting lock…", cacheKey)
+			token.lock.Lock()
```
mtrmac, line 695: Uh, do we need to keep holding `tokenCacheLock` here? We might be waiting for a long time, and delay users with a different `cacheKey` who don’t care about our `cacheKey`.

mtrmac:
The not-directly-related refactoring are now filed separately in #2480.

jankaluza:

mtrmac:
This continues in https://github.com/containers/container-libs/pull/415 .

PR #1976, opened 2023-06-01
storage: do not force the UncompressedDigest

giuseppe:
do not force the UncompressedDigest to be the same as the CompressedDigest, but let the storage driver configure it.

### DIFF
--- a/storage/storage_dest.go
+++ b/storage/storage_dest.go
@@ -606,9 +606,6 @@ func (s *storageImageDestination) commitLayer(index int, info addedLayerInfo, si
 			return err
 		}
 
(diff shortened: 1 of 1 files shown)


mtrmac (commented):
I’m fine with moving the responsibility to c/storage.

- How does this work against current c/storage? I assume this should only come with a c/storage dependency bump.
- I think we _also_ need to manage `s.blobDiffIDs`, to allow correctly building multi-layer structures with a chunked base and a layer on top (building the correct “chain ID”)

giuseppe:
given the feedback on the c/storage PR, I am closing this

PR #1977, opened 2023-06-01
Use x/exp/slices, and other small cleanups

mtrmac:
Use x/exp/slices, just like other code in c/image; that will allow us to trivially switch to the standard-library `slices` package coming in Go 1.21.

Also includes other small cleanups in related code.


flouthoc (approved):
LGTM
/lgtm

PR #1980, opened 2023-06-01
copy: do not fail if digest mismatches

giuseppe:
if the computed digest mismatches the expected one for a partial image.  This is expected since partial images are stored by their TOC digest, not the diffID for the layer.

### DIFF
--- a/copy/single.go
+++ b/copy/single.go
@@ -20,6 +20,7 @@ import (
 	compressiontypes "github.com/containers/image/v5/pkg/compression/types"
 	"github.com/containers/image/v5/transports"
 	"github.com/containers/image/v5/types"
+	chunkedToc "github.com/containers/storage/pkg/chunked/toc"
 	digest "github.com/opencontainers/go-digest"
 	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/sirupsen/logrus"
@@ -694,6 +695,13 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, to
 			requiredCompression = ic.compressionFormat
 			originalCompression = srcInfo.CompressionAlgorithm
 		}
+
+		// Check if we have a chunked layer in storage that's based on that blob.  These layers are stored by their TOC digest.
+		tocDigest, err := chunkedToc.GetTOCDigest(srcInfo.Annotations)
+		if err != nil {
+			return types.BlobInfo{}, "", err
+		}
+
 		reused, reusedBlob, err := ic.c.dest.TryReusingBlobWithOptions(ctx, srcInfo, private.TryReusingBlobOptions{
 			Cache:               ic.c.blobInfoCache,
 			CanSubstitute:       canSubstitute,
@@ -702,6 +710,7 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, to
 			SrcRef:              srcRef,
 			RequiredCompression: requiredCompression,
 			OriginalCompression: originalCompression,
+			TOCDigest:           tocDigest,
 		})
 		if err != nil {
 			return types.BlobInfo{}, "", fmt.Errorf("trying to reuse blob %s at destination: %w", srcInfo.Digest, err)
--- a/internal/private/private.go
+++ b/internal/private/private.go
@@ -117,6 +117,7 @@ type TryReusingBlobOptions struct {
 	EmptyLayer          bool                   // True if the blob is an "empty"/"throwaway" layer, and may not necessarily be physically represented.
 	LayerIndex          *int                   // If the blob is a layer, a zero-based index of the layer within the image; nil otherwise.
 	SrcRef              reference.Named        // A reference to the source image that contains the input blob.
+	TOCDigest           *digest.Digest         // If specified, the blob can be looked up in the destination also by its TOC digest.
 }
 
 // ReusedBlob is information about a blob reused in a destination.
--- a/manifest/oci.go
+++ b/manifest/oci.go
@@ -9,6 +9,7 @@ import (
 	compressiontypes "github.com/containers/image/v5/pkg/compression/types"
 	"github.com/containers/image/v5/types"
 	ociencspec "github.com/containers/ocicrypt/spec"
+	chunkedToc "github.com/containers/storage/pkg/chunked/toc"
 	"github.com/opencontainers/go-digest"
 	"github.com/opencontainers/image-spec/specs-go"
 	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
@@ -235,7 +236,7 @@ func (m *OCI1) Inspect(configGetter func(types.BlobInfo) ([]byte, error)) (*type
 }
 
 // ImageID computes an ID which can uniquely identify this image by its contents.
-func (m *OCI1) ImageID([]digest.Digest) (string, error) {
+func (m *OCI1) ImageID(diffIDs []digest.Digest) (string, error) {
 	// The way m.Config.Digest “uniquely identifies” an image is
 	// by containing RootFS.DiffIDs, which identify the layers of the image.
 	// For non-image artifacts, the we can’t expect the config to change
@@ -259,7 +260,42 @@ func (m *OCI1) ImageID([]digest.Digest) (string, error) {
 	if err := m.Config.Digest.Validate(); err != nil {
 		return "", err
 	}
+
+	// If there is any layer that is using partial content, we calculate the image ID
+	// in a different way since the diffID cannot be validated as for regular pulled images.
+	for _, layer := range m.Layers {
+		toc, err := chunkedToc.GetTOCDigest(layer.Annotations)
+		if err != nil {
+			return "", fmt.Errorf("error looking up annotation for layer %q: %w", layer.Digest, err)
+		}
+		if toc != nil {
+			return m.calculateImageIDForPartialImage(diffIDs)
+		}
+	}
+
 	return m.Config.Digest.Hex(), nil
+}
+
+func (m *OCI1) calculateImageIDForPartialImage(diffIDs []digest.Digest) (string, error) {
+	newID := digest.Canonical.Digester()
+	for i, layer := range m.Layers {
+		diffID := diffIDs[i]
+		_, err := newID.Hash().Write([]byte(diffID.Hex()))
+		if err != nil {
+			return "", fmt.Errorf("error writing diffID %q: %w", diffID, err)
+		}
+		toc, err := chunkedToc.GetTOCDigest(layer.Annotations)
+		if err != nil {
+			return "", fmt.Errorf("error looking up annotation for layer %q: %w", layer.Digest, err)
+		}
+		if toc != nil {
+			_, err = newID.Hash().Write([]byte(toc.Hex()))
+			if err != nil {
+				return "", fmt.Errorf("error writing TOC %q: %w", toc, err)
+			}
+		}
+	}
+	return newID.Digest().Hex(), nil
 }
 
 // CanChangeLayerCompression returns true if we can compress/decompress layers with mimeType in the current image
--- a/storage/storage_dest.go
+++ b/storage/storage_dest.go
@@ -532,6 +532,18 @@ func (s *storageImageDestination) queueOrCommit(index int, info addedLayerInfo) 
 	return nil
 }
 
+// getDiffIDOrTOCDigest returns the diffID for the specified digest or the digest for the TOC, if known.
+func (s *storageImageDestination) getDiffIDOrTOCDigest(uncompressedDigest digest.Digest) (digest.Digest, bool) {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	if d, found := s.diffOutputs[uncompressedDigest]; found {
+		return d.TOCDigest, found
+	}
+	d, found := s.blobDiffIDs[uncompressedDigest]
+	return d, found
+}
+
 // commitLayer commits the specified layer with the given index to the storage.
 // size can usually be -1; it can be provided if the layer is not known to be already present in blobDiffIDs.
 //
@@ -562,17 +574,16 @@ func (s *storageImageDestination) commitLayer(index int, info addedLayerInfo, si
 
 	// Check if there's already a layer with the ID that we'd give to the result of applying
 	// this layer blob to its parent, if it has one, or the blob's hex value otherwise.
-	s.lock.Lock()
-	diffID, haveDiffID := s.blobDiffIDs[info.digest]
-	s.lock.Unlock()
-	if !haveDiffID {
+	// The diffIDOrTOCDigest refers either to the DiffID or the digest of the TOC.
+	diffIDOrTOCDigest, haveDiffIDOrTOCDigest := s.getDiffIDOrTOCDigest(info.digest)
+	if !haveDiffIDOrTOCDigest {
 		// Check if it's elsewhere and the caller just forgot to pass it to us in a PutBlob(),
 		// or to even check if we had it.
 		// Use none.NoCache to avoid a repeated DiffID lookup in the BlobInfoCache; a caller
 		// that relies on using a blob digest that has never been seen by the store had better call
 		// TryReusingBlob; not calling PutBlob already violates the documented API, so there’s only
 		// so far we are going to accommodate that (if we should be doing that at all).
-		logrus.Debugf("looking for diffID for blob %+v", info.digest)
+		logrus.Debugf("looking for diffID or TOC digest for blob %+v", info.digest)
 		// Use tryReusingBlobAsPending, not the top-level TryReusingBlobWithOptions, to prevent recursion via queueOrCommit.
 		has, _, err := s.tryReusingBlobAsPending(info.digest, size, &private.TryReusingBlobOptions{
 			Cache:         none.NoCache,
@@ -582,16 +593,16 @@ func (s *storageImageDestination) commitLayer(index int, info addedLayerInfo, si
 			return fmt.Errorf("checking for a layer based on blob %q: %w", info.digest.String(), err)
 		}
 		if !has {
-			return fmt.Errorf("error determining uncompressed digest for blob %q", info.digest.String())
+			return fmt.Errorf("error determining uncompressed digest or TOC digest for blob %q", info.digest.String())
 		}
-		diffID, haveDiffID = s.blobDiffIDs[info.digest]
-		if !haveDiffID {
-			return fmt.Errorf("we have blob %q, but don't know its uncompressed digest", info.digest.String())
+		diffIDOrTOCDigest, haveDiffIDOrTOCDigest = s.getDiffIDOrTOCDigest(info.digest)
+		if !haveDiffIDOrTOCDigest {
+			return fmt.Errorf("we have blob %q, but don't know its uncompressed or TOC digest", info.digest.String())
 		}
 	}
-	id := diffID.Hex()
+	id := diffIDOrTOCDigest.Hex()
 	if lastLayer != "" {
-		id = digest.Canonical.FromBytes([]byte(lastLayer + "+" + diffID.Hex())).Hex()
+		id = digest.Canonical.FromBytes([]byte(lastLayer + "+" + diffIDOrTOCDigest.Hex())).Hex()
 	}
 	if layer, err2 := s.imageRef.transport.store.Layer(id); layer != nil && err2 == nil {
 		// There's already a layer that should have the right contents, just reuse it.
@@ -642,7 +653,7 @@ func (s *storageImageDestination) commitLayer(index int, info addedLayerInfo, si
 	if !ok {
 		// Try to find the layer with contents matching that blobsum.
 		layer := ""
-		layers, err2 := s.imageRef.transport.store.LayersByUncompressedDigest(diffID)
+		layers, err2 := s.imageRef.transport.store.LayersByUncompressedDigest(diffIDOrTOCDigest)
 		if err2 == nil && len(layers) > 0 {
 			layer = layers[0].ID
 		} else {
@@ -668,7 +679,7 @@ func (s *storageImageDestination) commitLayer(index int, info addedLayerInfo, si
 		// the same lock, so the diff can't just be directly streamed from one
 		// to the other.
 		filename = s.computeNextBlobCacheFile()
-		file, err := os.OpenFile(filename, os.O_CREATE|os.O_TRUNC|os.O_WRONLY|os.O_EXCL, 0600)
+		file, err := os.OpenFile(filename, os.O_CREATE|os.O_TRUNC|os.O_WRONLY|os.O_EXCL, 0o600)
 		if err != nil {
 			diff.Close()
 			return fmt.Errorf("creating temporary file %q: %w", filename, err)
@@ -698,7 +709,7 @@ func (s *storageImageDestination) commitLayer(index int, info addedLayerInfo, si
 	// TODO: This can take quite some time, and should ideally be cancellable using ctx.Done().
 	layer, _, err := s.imageRef.transport.store.PutLayer(id, lastLayer, nil, "", false, &storage.LayerOptions{
 		OriginalDigest:     info.digest,
-		UncompressedDigest: diffID,
+		UncompressedDigest: diffIDOrTOCDigest,
 	}, file)
 	if err != nil && !errors.Is(err, storage.ErrDuplicateID) {
 		return fmt.Errorf("adding layer with blob %q: %w", info.digest, err)
--- a/storage/storage_dest.go
+++ b/storage/storage_dest.go
@@ -77,13 +77,13 @@ type storageImageDestination struct {
 	indexToStorageID map[int]*string
 	// All accesses to below data are protected by `lock` which is made
 	// *explicit* in the code.
-	blobDiffIDs           map[digest.Digest]digest.Digest                       // Mapping from layer blobsums to their corresponding DiffIDs
-	fileSizes             map[digest.Digest]int64                               // Mapping from layer blobsums to their sizes
-	filenames             map[digest.Digest]string                              // Mapping from layer blobsums to names of files we used to hold them
-	currentIndex          int                                                   // The index of the layer to be committed (i.e., lower indices have already been committed)
-	indexToAddedLayerInfo map[int]addedLayerInfo                                // Mapping from layer (by index) to blob to add to the image
-	blobAdditionalLayer   map[digest.Digest]storage.AdditionalLayer             // Mapping from layer blobsums to their corresponding additional layer
-	diffOutputs           map[digest.Digest]*graphdriver.DriverWithDifferOutput // Mapping from digest to differ output
+	uncompressedOrTocDigest map[digest.Digest]digest.Digest                       // Mapping from layer blobsums to their corresponding DiffIDs or TOC IDs.
+	fileSizes               map[digest.Digest]int64                               // Mapping from layer blobsums to their sizes
+	filenames               map[digest.Digest]string                              // Mapping from layer blobsums to names of files we used to hold them
+	currentIndex            int                                                   // The index of the layer to be committed (i.e., lower indices have already been committed)
+	indexToAddedLayerInfo   map[int]addedLayerInfo                                // Mapping from layer (by index) to blob to add to the image
+	blobAdditionalLayer     map[digest.Digest]storage.AdditionalLayer             // Mapping from layer blobsums to their corresponding additional layer
+	diffOutputs             map[digest.Digest]*graphdriver.DriverWithDifferOutput // Mapping from digest to differ output
 }
 
 // addedLayerInfo records data about a layer to use in this image.
@@ -117,18 +117,18 @@ func newImageDestination(sys *types.SystemContext, imageRef storageReference) (*
 			HasThreadSafePutBlob:           true,
 		}),
 
-		imageRef:              imageRef,
-		directory:             directory,
-		signatureses:          make(map[digest.Digest][]byte),
-		blobDiffIDs:           make(map[digest.Digest]digest.Digest),
-		blobAdditionalLayer:   make(map[digest.Digest]storage.AdditionalLayer),
-		fileSizes:             make(map[digest.Digest]int64),
-		filenames:             make(map[digest.Digest]string),
-		SignatureSizes:        []int{},
-		SignaturesSizes:       make(map[digest.Digest][]int),
-		indexToStorageID:      make(map[int]*string),
-		indexToAddedLayerInfo: make(map[int]addedLayerInfo),
-		diffOutputs:           make(map[digest.Digest]*graphdriver.DriverWithDifferOutput),
+		imageRef:                imageRef,
+		directory:               directory,
+		signatureses:            make(map[digest.Digest][]byte),
+		uncompressedOrTocDigest: make(map[digest.Digest]digest.Digest),
+		blobAdditionalLayer:     make(map[digest.Digest]storage.AdditionalLayer),
+		fileSizes:               make(map[digest.Digest]int64),
+		filenames:               make(map[digest.Digest]string),
+		SignatureSizes:          []int{},
+		SignaturesSizes:         make(map[digest.Digest][]int),
+		indexToStorageID:        make(map[int]*string),
+		indexToAddedLayerInfo:   make(map[int]addedLayerInfo),
+		diffOutputs:             make(map[digest.Digest]*graphdriver.DriverWithDifferOutput),
 	}
 	dest.Compat = impl.AddCompat(dest)
 	return dest, nil
@@ -227,7 +227,7 @@ func (s *storageImageDestination) putBlobToPendingFile(stream io.Reader, blobinf
 
 	// Record information about the blob.
 	s.lock.Lock()
-	s.blobDiffIDs[blobDigest] = diffID.Digest()
+	s.uncompressedOrTocDigest[blobDigest] = diffID.Digest()
 	s.fileSizes[blobDigest] = counter.Count
 	s.filenames[blobDigest] = filename
 	s.lock.Unlock()
@@ -289,7 +289,7 @@ func (s *storageImageDestination) PutBlobPartial(ctx context.Context, chunkAcces
 	blobDigest := srcInfo.Digest
 
 	s.lock.Lock()
-	s.blobDiffIDs[blobDigest] = blobDigest
+	s.uncompressedOrTocDigest[blobDigest] = blobDigest
 	s.fileSizes[blobDigest] = 0
 	s.filenames[blobDigest] = ""
 	s.diffOutputs[blobDigest] = out
@@ -321,7 +321,7 @@ func (s *storageImageDestination) TryReusingBlobWithOptions(ctx context.Context,
 	})
 }
 
-// tryReusingBlobAsPending implements TryReusingBlobWithOptions for (digest, size or -1), filling s.blobDiffIDs and other metadata.
+// tryReusingBlobAsPending implements TryReusingBlobWithOptions for (digest, size or -1), filling s.uncompressedOrTocDigest and other metadata.
 // The caller must arrange the blob to be eventually committed using s.commitLayer().
 func (s *storageImageDestination) tryReusingBlobAsPending(digest digest.Digest, size int64, options *private.TryReusingBlobOptions) (bool, private.ReusedBlob, error) {
 	// lock the entire method as it executes fairly quickly
@@ -335,7 +335,7 @@ func (s *storageImageDestination) tryReusingBlobAsPending(digest digest.Digest, 
 			return false, private.ReusedBlob{}, fmt.Errorf(`looking for compressed layers with digest %q and labels: %w`, digest, err)
 		} else if err == nil {
 			// Record the uncompressed value so that we can use it to calculate layer IDs.
-			s.blobDiffIDs[digest] = aLayer.UncompressedDigest()
+			s.uncompressedOrTocDigest[digest] = aLayer.UncompressedDigest()
 			s.blobAdditionalLayer[digest] = aLayer
 			return true, private.ReusedBlob{
 				Digest: digest,
@@ -366,7 +366,7 @@ func (s *storageImageDestination) tryReusingBlobAsPending(digest digest.Digest, 
 	}
 	if len(layers) > 0 {
 		// Save this for completeness.
-		s.blobDiffIDs[digest] = layers[0].UncompressedDigest
+		s.uncompressedOrTocDigest[digest] = layers[0].UncompressedDigest
 		return true, private.ReusedBlob{
 			Digest: digest,
 			Size:   layers[0].UncompressedSize,
@@ -380,7 +380,7 @@ func (s *storageImageDestination) tryReusingBlobAsPending(digest digest.Digest, 
 	}
 	if len(layers) > 0 {
 		// Record the uncompressed value so that we can use it to calculate layer IDs.
-		s.blobDiffIDs[digest] = layers[0].UncompressedDigest
+		s.uncompressedOrTocDigest[digest] = layers[0].UncompressedDigest
 		return true, private.ReusedBlob{
 			Digest: digest,
 			Size:   layers[0].CompressedSize,
@@ -398,7 +398,7 @@ func (s *storageImageDestination) tryReusingBlobAsPending(digest digest.Digest, 
 			}
 			if len(layers) > 0 {
 				if size != -1 {
-					s.blobDiffIDs[digest] = layers[0].UncompressedDigest
+					s.uncompressedOrTocDigest[digest] = layers[0].UncompressedDigest
 					return true, private.ReusedBlob{
 						Digest: digest,
 						Size:   size,
@@ -407,7 +407,7 @@ func (s *storageImageDestination) tryReusingBlobAsPending(digest digest.Digest, 
 				if !options.CanSubstitute {
 					return false, private.ReusedBlob{}, fmt.Errorf("Internal error: options.CanSubstitute was expected to be true for blob with digest %s", digest)
 				}
-				s.blobDiffIDs[uncompressedDigest] = layers[0].UncompressedDigest
+				s.uncompressedOrTocDigest[uncompressedDigest] = layers[0].UncompressedDigest
 				return true, private.ReusedBlob{
 					Digest: uncompressedDigest,
 					Size:   layers[0].UncompressedSize,
@@ -438,7 +438,7 @@ func (s *storageImageDestination) computeID(m manifest.Manifest) string {
 				continue
 			}
 			blobSum := m.FSLayers[i].BlobSum
-			diffID, ok := s.blobDiffIDs[blobSum]
+			diffID, ok := s.uncompressedOrTocDigest[blobSum]
 			if !ok {
 				logrus.Infof("error looking up diffID for layer %q", blobSum.String())
 				return ""
@@ -540,12 +540,12 @@ func (s *storageImageDestination) getDiffIDOrTOCDigest(uncompressedDigest digest
 	if d, found := s.diffOutputs[uncompressedDigest]; found {
 		return d.TOCDigest, found
 	}
-	d, found := s.blobDiffIDs[uncompressedDigest]
+	d, found := s.uncompressedOrTocDigest[uncompressedDigest]
 	return d, found
 }
 
 // commitLayer commits the specified layer with the given index to the storage.
-// size can usually be -1; it can be provided if the layer is not known to be already present in blobDiffIDs.
+// size can usually be -1; it can be provided if the layer is not known to be already present in uncompressedOrTocDigest.
 //
 // Note that the previous layer is expected to already be committed.
 //
--- a/storage/storage_dest.go
+++ b/storage/storage_dest.go
@@ -416,6 +416,25 @@ func (s *storageImageDestination) tryReusingBlobAsPending(digest digest.Digest, 
 		}
 	}
 
+	tocDigest := digest
+	if options.TOCDigest != nil {
+		tocDigest = *options.TOCDigest
+	}
+
+	// Check if we have a chunked layer in storage with the same TOC digest.
+	layers, err = s.imageRef.transport.store.LayersByTOCDigest(tocDigest)
+	if err != nil && !errors.Is(err, storage.ErrLayerUnknown) {
+		return false, private.ReusedBlob{}, fmt.Errorf(`looking for layers with TOC digest %q: %w`, tocDigest, err)
+	}
+	if len(layers) > 0 {
+		// Save this for completeness.
+		s.uncompressedOrTocDigest[digest] = layers[0].TOCDigest
+		return true, private.ReusedBlob{
+			Digest: layers[0].TOCDigest,
+			Size:   layers[0].UncompressedSize,
+		}, nil
+	}
+
 	// Nope, we don't have it.
 	return false, private.ReusedBlob{}, nil
 }
--- a/storage/storage_dest.go
+++ b/storage/storage_dest.go
@@ -464,9 +464,13 @@ func (s *storageImageDestination) computeID(m manifest.Manifest) string {
 			}
 			diffIDs = append([]digest.Digest{diffID}, diffIDs...)
 		}
-	case *manifest.Schema2, *manifest.OCI1:
-		// We know the ID calculation for these formats doesn't actually use the diffIDs,
-		// so we don't need to populate the diffID list.
+	case *manifest.Schema2:
+		// We know the ID calculation doesn't actually use the diffIDs, so we don't need to populate
+		// the diffID list.
+	case *manifest.OCI1:
+		for _, l := range m.Layers {
+			diffIDs = append(diffIDs, l.Digest)
+		}
 	default:
 		return ""
 	}
--- a/storage/storage_dest.go
+++ b/storage/storage_dest.go
@@ -541,7 +541,7 @@ func (s *storageImageDestination) queueOrCommit(index int, info addedLayerInfo) 
 		}
 		s.lock.Unlock()
 		// Note: commitLayer locks on-demand.
-		if err := s.commitLayer(index, info, -1); err != nil {
+		if stopQueue, err := s.commitLayer(index, info, -1); stopQueue || err != nil {
 			return err
 		}
 		s.lock.Lock()
@@ -570,15 +570,17 @@ func (s *storageImageDestination) getDiffIDOrTOCDigest(uncompressedDigest digest
 // commitLayer commits the specified layer with the given index to the storage.
 // size can usually be -1; it can be provided if the layer is not known to be already present in uncompressedOrTocDigest.
 //
+// If the layer cannot be committed yet, the function returns (true, nil).
+//
 // Note that the previous layer is expected to already be committed.
 //
 // Caution: this function must be called without holding `s.lock`.  Callers
 // must guarantee that, at any given time, at most one goroutine may execute
 // `commitLayer()`.
-func (s *storageImageDestination) commitLayer(index int, info addedLayerInfo, size int64) error {
+func (s *storageImageDestination) commitLayer(index int, info addedLayerInfo, size int64) (bool, error) {
 	// Already committed?  Return early.
 	if _, alreadyCommitted := s.indexToStorageID[index]; alreadyCommitted {
-		return nil
+		return false, nil
 	}
 
 	// Start with an empty string or the previous layer ID.  Note that
@@ -592,7 +594,7 @@ func (s *storageImageDestination) commitLayer(index int, info addedLayerInfo, si
 	// Carry over the previous ID for empty non-base layers.
 	if info.emptyLayer {
 		s.indexToStorageID[index] = &lastLayer
-		return nil
+		return false, nil
 	}
 
 	// Check if there's already a layer with the ID that we'd give to the result of applying
@@ -613,14 +615,14 @@ func (s *storageImageDestination) commitLayer(index int, info addedLayerInfo, si
 			CanSubstitute: false,
 		})
 		if err != nil {
-			return fmt.Errorf("checking for a layer based on blob %q: %w", info.digest.String(), err)
+			return false, fmt.Errorf("checking for a layer based on blob %q: %w", info.digest.String(), err)
 		}
 		if !has {
-			return fmt.Errorf("error determining uncompressed digest or TOC digest for blob %q", info.digest.String())
+			return false, fmt.Errorf("error determining uncompressed digest or TOC digest for blob %q", info.digest.String())
 		}
 		diffIDOrTOCDigest, haveDiffIDOrTOCDigest = s.getDiffIDOrTOCDigest(info.digest)
 		if !haveDiffIDOrTOCDigest {
-			return fmt.Errorf("we have blob %q, but don't know its uncompressed or TOC digest", info.digest.String())
+			return false, fmt.Errorf("we have blob %q, but don't know its uncompressed or TOC digest", info.digest.String())
 		}
 	}
 	id := diffIDOrTOCDigest.Hex()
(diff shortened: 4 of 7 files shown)


mtrmac (changes requested):
Unrestricted like this, the change would completely break the security of digest references, and signatures.

It’s reasonable to contemplate that in _some_ cases, we need a non-digest layer identifier of some kind; currently, in some transports we need to compute a digest just to have a digest in a manifest and create a consumable `ImageSource`. But the assumption that layers are content-addressable and _always have a digest_ is deeply and widely spread throughout the codebase, and we would need to both find a replacement for the current image integrity assumptions (or to decide that one isn’t necessary), and to carefully migrate code away from the existing content-addressability assumption.

giuseppe:
> Unrestricted like this, the change would completely break the security of digest references, and signatures.
> 
> It’s reasonable to contemplate that in _some_ cases, we need a non-digest layer identifier of some kind; currently, in some transports we need to compute a digest just to have a digest in a manifest and create a consumable `ImageSource`. But the assumption that layers are content-addressable and _always have a digest_ is deeply and widely spread throughout the codebase, and we would need to both find a replacement for the current image integrity assumptions (or to decide that one isn’t necessary), and to carefully migrate code away from the existing content-addressability assumption.

I've changed the implementation to use this behavior only for layers that were retrieved using the partial pull feature.

It needs this change in c/storage: https://github.com/containers/storage/pull/1627 in particular the patch "[store: store if the layer was created with a partial pull](https://github.com/containers/storage/pull/1627/commits/373af6f1711a1c9e35909762d52249d2d2ee96e6)"

mtrmac (commented):
Most importantly, I’m much more worried about the pull path and layer deduplication. Let’s figure out how to do that first, and that will clarify the digests situation. Making the push path work should only come afterwards.

giuseppe:
> Most importantly, I’m much more worried about the pull path and layer deduplication. Let’s figure out how to do that first, and that will clarify the digests situation. Making the push path work should only come afterwards.

the pull should not be a problem (just a little bit more expensive) since we will end up requesting the TOC for the layer but then all files are already present locally, so it won't be a problem.

If you refer to having the `zstd:chunked` version and the `gzip` version being deduplicated, I think we should not worry about this case, the long term goal is to replace the gzip format, so even if for now we end up in the worse case with two layers, that is fine.

giuseppe:
@mtrmac what do you think of this approach?

giuseppe:
if there are no big objections can we move forward with this?  We can rework it again if necessary, but without this fix a partial layer is pulled every time we do a pull

giuseppe:
@mtrmac @rhatdan @vrothberg PTAL

copy/blob.go
```diff
@@ -128,8 +131,12 @@ func (ic *imageCopier) copyBlobFromStream(ctx context.Context, srcReader io.Read
 		}
 	}
 
-	if digestingReader.validationFailed { // Coverage: This should never happen.
-		return types.BlobInfo{}, fmt.Errorf("Internal error writing blob %s, digest verification failed but was ignored", srcInfo.Digest)
+	if digestingReader.validationFailed {
+		if !canRewriteDigest { // Coverage: This should never happen.
+			return types.BlobInfo{}, fmt.Errorf("Internal error writing blob %s, digest verification failed", srcInfo.Digest)
+		}
+		logrus.Warningf("Digest verification failed for blob %s but was ignored", srcInfo.Digest)
```
vrothberg, line 138: That's a scary message. Can we explain why it's being ignored?

vrothberg (commented):
Code LGTM but I want @mtrmac to chime in. He's partially unavailable this week.

copy/blob.go
giuseppe: the warning is outputted only when `canRewriteDigest == true`.  We haven't validated the digest earlier for partial images since we didn't pull the entire tarball, so for dedup purposes we use the TOC digest (which is not the DiffID).  When we copy the layer from the storage, we recompute the entire tarball so we can recalculate the correct DiffID and ignore the TOC Digest.

rhatdan: If it expected it should be Info not warning.

mtrmac: (Warning: At this point I don’t even have a strong opinion on the overall design. This is a drive-by comment.)

---

Like, why is this even validating a digest value when we know it is not going to match?

Please be precise in what the code does (or does not) track, or mean. `digestingReader` already provides `validationFailed`/`validationSucceeded`, and that can already be used to express “I have no opinion”. If something can’t be expressed with the existing data structures exactly, well, make that possible.

Please don’t create spaghetti where “`validationFailed` — oh but we didn’t _mean it_ here because another variable is set.”

---

(Actually if we don’t want to validate the blob’s digest… we shouldn’t even be computing it. But that’s a performance optimization I care very little about.)

giuseppe: I've pushed a new version where I drop the warning.

I am open to suggestions on how to improve the implementation.  We still need to calculate the digest for the tarball because the one we are using with incremental images is the TOC digest (the only thing we can validate when we pull an image partially).

The idea is that when we generate the entire tarball, we can also calculate the diffID for the tarball, and use it instead of the digest we have in the store (the TOC digest).

How would we express "I have no opinion" with the current `digestReader`?  AFAICS, it will always return an error when the digest mismatches.  I don't think we can just ignore the error, since it could be related to something else.

giuseppe: I was wrong, I don't really need to change the `digestReader` implementation and just skip the check when dealing with partial images.

giuseppe:
rebased

vrothberg (approved):
LGTM
Nit: the last commit message still points to the warning that has been dropped

copy/single.go
```diff
@@ -20,6 +20,7 @@ import (
 	compressiontypes "github.com/containers/image/v5/pkg/compression/types"
 	"github.com/containers/image/v5/transports"
 	"github.com/containers/image/v5/types"
+	"github.com/containers/storage/pkg/chunked"
```
mtrmac, line 23: `pkg/chunked` brings in a _lot_ of dependencies. I’d rather prefer to avoid that in the generic code and non-storage transports.

storage/storage_dest.go
```diff
@@ -445,16 +454,42 @@ func (s *storageImageDestination) computeID(m manifest.Manifest) string {
...
-		// so we don't need to populate the diffID list.
+	case *manifest.Schema2:
+		// We know the ID calculation doesn't actually use the diffIDs, so we don't need to populate
+		// the diffID list.
+	case *manifest.OCI1:
+		for _, layer := range m.Layers {
+			toc, err := chunked.GetTOCDigest(layer.Annotations)
+			if err != nil {
+				logrus.Infof("error looking up annotation for layer %q: %v", layer.Digest, err)
+				continue
```
mtrmac, line 465: Why is it OK to silently continue here?

---

OTOH this _always_ fails on non-Linux. Is that right? Do we want the same image to have different IDs on different platforms?

That might well be right … but… it seems non-trivial. What is the consistent model here?

One possibility is to make `GetTOCDigest` OS-neutral (and move it into a lighter-weight subpackage); then we might need to have some other place to decide whether the OS supports chunked.

storage/storage_dest.go
```diff
@@ -328,6 +328,14 @@ func (s *storageImageDestination) tryReusingBlobAsPending(digest digest.Digest,
 	s.lock.Lock()
 	defer s.lock.Unlock()
 
+	toc, err := chunked.GetTOCDigest(options.Annotations)
+	if err != nil {
+		return false, private.ReusedBlob{}, err
+	}
+	if toc != nil {
+		digest = *toc
+	}
```
mtrmac, line 337: FIXME: `GetTOCDigest` can return two different values. Why isn’t that ambiguity a risk? It seems that we read the same data structure in both cases, so that seems fine — look into that more.

mtrmac (changes requested):

copy/single.go
```diff
@@ -448,7 +449,13 @@ func (ic *imageCopier) copyLayers(ctx context.Context) ([]compressiontypes.Algor
 				logrus.Debugf("Skipping foreign layer %q copy to %s", cld.destInfo.Digest, ic.c.dest.Reference().Transport().Name())
 			}
 		} else {
-			cld.destInfo, cld.diffID, cld.err = ic.copyLayer(ctx, srcLayer, toEncrypt, pool, index, srcRef, manifestLayerInfos[index].EmptyLayer)
+			toc, err := chunked.GetTOCDigest(manifestLayerInfos[index].Annotations)
```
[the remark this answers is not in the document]
giuseppe: thanks for spotting it.  Indeed that must be fixed to not always fail on non-Linux.

Would it be OK for you if I move the `GetTOCDigest` to a new `pkg/chunked/toc` package and define only one version of the function?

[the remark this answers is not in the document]
mtrmac: Yes, something like that would make sense to me.

I don’t immediately know where the “do we consume chunked content” decision should happen — at a quick glance currently pkg/chunked.GetDiffer can do that, local to c/storage and with no new public API, but this PR might add more places that need to know that.

copy/single.go
```diff
@@ -448,7 +449,13 @@ func (ic *imageCopier) copyLayers(ctx context.Context) ([]compressiontypes.Algor
...
+			if err != nil {
+				cld.err = err
+			} else {
+				incremental := toc != nil
```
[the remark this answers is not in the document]
giuseppe: incremental here means that only the TOC digest was validated and the layer was stored by its TOC digest, not the diffID.

When `incremental` is set we skip the diffID validation and use the digest we get from `PutBlobWithOptions`.

copy/single.go
giuseppe: can this be solved with a new package `github.com/containers/storage/pkg/chunked/toc` that includes only the `GetTOCDigest` function?

copy/single.go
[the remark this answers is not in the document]
mtrmac: Yes, _in the context of this PR_ we all know what that means.

I’m concerned that the code is clear by the time the PR is committed, so that anyone looking at it 2 years from now can maintain the required invariants for this security-critical value.

copy/single.go
mtrmac: I’d recommend dealing with the way the value is determined, first: Where does the Linux OS dependency decision point live; and how does the generic copy code arrive at the decision to rely on the TOC.

- Hypothetically, if the code detecting this were moved to one of `c/image/{,internal/}{image, manifest}`, I’d strongly prefer for these c/image subpackages not to drag in large dependencies, so a new c/storage subpackage would be best in that case.
- c/image/copy already, via pkg/compression, depends on pkg/chunked/compressor, so if the code detecting this stayed where it is, a new package is not _strictly speaking_ required — the function can be added into …pkg/compressor. As to whether it is a _good fit_ for that package, at this point I don’t have an opinion; it might be better to introduce a new subpackage anyway.
- At the other extreme (my current best guess but didn’t look into it closely), if the code detecting this were moved to c/image/storage, which already depends on all of `c/storage/pkg/chunked`, a new subpackage would not be necessary.

giuseppe:
pushed a new version that needs: https://github.com/containers/storage/pull/1680

In addition to checking for a TOC in the annotations, now it also validates that the source transport supports partial images.  I'll rework this part to use a stub instead of the current hack, if the idea is good for you

flouthoc (commented):
I think this needs `make vendor`

giuseppe:
ugh, there is still an error :( https://github.com/containers/storage/pull/1682

giuseppe:
> I think this needs `make vendor`

green now

rhatdan:
LGTM

copy/blob.go
```diff
@@ -128,13 +132,13 @@ func (ic *imageCopier) copyBlobFromStream(ctx context.Context, srcReader io.Read
 		}
 	}
 
-	if digestingReader.validationFailed { // Coverage: This should never happen.
-		return types.BlobInfo{}, fmt.Errorf("Internal error writing blob %s, digest verification failed but was ignored", srcInfo.Digest)
+	if digestingReader != nil && digestingReader.validationFailed {
+		return types.BlobInfo{}, fmt.Errorf("Internal error writing blob %s, digest verification failed", srcInfo.Digest)
 	}
 	if stream.info.Digest != "" && uploadedInfo.Digest != stream.info.Digest {
```
mtrmac, line 138: How does this interact with `skipDigestValidation`?

(I guess it does, but in that case the digest in `stream.info` must be cleared much earlier.)

copy/single.go
```diff
@@ -398,6 +400,29 @@ func (ic *imageCopier) compareImageDestinationManifestEqual(ctx context.Context,
...
+	publicDest, err := src.Reference().NewImageDestination(ctx, nil)
+	if err != nil {
+		return false, err
+	}
+	defer publicDest.Close()
+
+	dest := imagedestination.FromPublic(publicDest)
+	defer dest.Close()
+
+	return dest.SupportsPutBlobPartial(), nil
```
mtrmac, line 423: Just to be explicit, this is a blocker for merging.

copy/single.go
```diff
@@ -398,6 +400,29 @@ func (ic *imageCopier) compareImageDestinationManifestEqual(ctx context.Context,
...
 
+// canSkipValidation returns true if the source supports storing partial images and the image is.
+func canSkipValidation(ctx context.Context, annotations map[string]string, src private.ImageSource) (bool, error) {
+	toc, err := toc.GetTOCDigest(annotations)
+	if err != nil {
+		return false, err
+	}
+	if toc == nil {
+		return false, nil
+	}
```
mtrmac, line 411: How does the presence of the TOC really factor into the decision here?

I appreciate that the actual decision mechanism below hasn’t been designed yet… my first guess “the source transport claims that the blob is valid and doesn’t need verifying” would just as well apply to TOC-less layers. Or maybe the source transport should be returning a per-blob value instead of a global constant? (But how would c/storage make that decision?)

mtrmac, line 411: There’s a whole another question here about what it means to have signature verification enforced for such a “trusted” source transport. Do we just give up and let it through? Refuse to validate policies that require signatures (somehow)?

It seems a bit incongruous to let things through here, while `ic.canSubstituteBlobs` tries so hard to only sign data we have seen and verified. Admittedly c/storage is already a bit of a special case, and signing local in-storage images is probably extremely rare.

Or should we, at this path, actually require the transport not just to _declare_ that the blob is trusted, but to positively validate the full blob against the TOC? (Ugh… that would be a lot of critical extra code. Or does that already exist?)

mtrmac (changes requested):
Mostly to forestall LGTMs on a known-unfinished PR…

(I don’t have a good overall conceptual picture at this point… I should find some uninterrupted time to catch up. See the “There’s a whole another question” part…)

copy/single.go
giuseppe: the presence of the TOC is to somehow limit the `skipValidation` only to images that could possibly be stored using a partial pull.  Of course, this is not a security measure because a malicious user could just use the annotation and inhibit the check we have in place.

That is why I am using the check `canSkipValidation`, to skip the validation only when we are copying from transports we can trust.  We can make it explicit in the transport itself and declare for each transport if it is safe to trust its content.  In practice, we would do it only for the `containers-storage` transport.

Is it a safe assumption?  Are there cases where we cannot trust what we have in the storage?

giuseppe: I've not looked into it, but if signature verification is enforced but I don't think we can use the partial pull feature at the moment.

We can skip the validation only if `canSubstituteBlobs`, I'll fix that.

mtrmac: Oh … a random late-night thought: I should not be overcomplicating this.

The digests we validate come from `LayerInfosForCopy`, which is already transport-authored and not authenticated by the signed manifest. So there is a clear precedent for still considering signatures valid in that case.

---

Now, in that case we do validate the uncompressed digest… against data that was previously recorded, i.e. that protects against corruption but not malice. So validating the layer against the TOC would be consistent with that, OTOH I’m not sure there is a very good reason to do that — it makes sense if the policy + trusted keys + executables of Podman etc. were more trusted than the c/storage data. 

---

(On a push, don’t we need to discard the TOC and build one anew while compressing, anyway?)

giuseppe: The TOC digest is used only in the image manifest (the `io.github.containers.zstd-chunked.manifest-checksum` annotation), so a different TOC will affect only the metadata and the compressed stream digest, it won't change the diffID.

I've pushed a new version, but I keep it marked as Draft for now as I want to experiment with this a bit more

giuseppe:
can I get another review?

giuseppe:
rebased

giuseppe:
The PR has been sitting here for quite a while now, without it the partial images feature is kind of broken at the moment.

@mtrmac could you please take a look?

mtrmac:
Note to self:
- Caught up on https://github.com/containers/storage/pull/1627, see my outstanding review comments
- Still to read: https://github.com/containers/storage/pull/1646

rhatdan:
@mtrmac any update on this?

mtrmac (commented):
Overall I think the way we distinguish between full layer digests and TOC digests, and verify / enforce correct layer / image reuse, needs to either be structurally totally obvious[1], or it needs a written argument for a less obvious design (preferably as comments on data structures, or less desirable as comments on the implementation), one that we can return later.

I don’t think it’s productive to keep reverse-engineering the security properties every time a maintainer needs to revisit this area. So, the way this PR stack just changes semantics of fields, or adds fields with no documentation at all, just just _hard_ for me to wrap my mind around.

mtrmac:
> Overall I think the way we distinguish between full layer digests and TOC digests, and verify / enforce correct layer / image reuse, needs to either be structurally totally obvious[1]

For the record, the current design:
- Every layer’s `UncompressedDigest` is known, at time of pull, to exactly match the layer contents. That’s
  - … not known to be true for partial pulls. What we are dealing with now.
  - Assumed to be provided by the backend for `LookupAdditionalLayer`
  - Computed by `putBlobToPendingFile` on full layer pulls
  - Assumed to be true, by induction, for all layers already existing in local storage
- Layer reuse is driven either by `UncompressedDigest` alone, or by `CompressedDigest` along with BlobInfoCache. BlobInfoCache only ever writes locally-computed data.
- So, the pull process, via `TryReusing…` or `PutBlob`… sets `blobDiffIDs[digest]` to point to DiffIDs that are known to be equivalent to the input digest values.
- Layer IDs in c/storage are constructed using a deterministic process from the sequence of `blobDiffIDs` values of the layers in the stack. So the same c/storage layer IDs imply exactly the same layer stack, deterministically.
- When constructing the image, the image ID for non-schema1 the config digest. That’s _assumed_ to be unique between images because it should contain `RootFS.DiffID` values (but we don’t actually read these fields and verify a match). But we do enforce that if there are two images with the same image ID, they have the same layer stack (if there is a mismatch, at least one of the two images contains invalid `RootFS.DiffID` values).

copy/blob.go
```diff
@@ -128,13 +132,13 @@ func (ic *imageCopier) copyBlobFromStream(ctx context.Context, srcReader io.Read
 		}
 	}
 
-	if digestingReader.validationFailed { // Coverage: This should never happen.
-		return types.BlobInfo{}, fmt.Errorf("Internal error writing blob %s, digest verification failed but was ignored", srcInfo.Digest)
+	if digestingReader != nil && digestingReader.validationFailed {
```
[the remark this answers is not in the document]
mtrmac: The violation of the `PutBlobWithOptions` API also means that the (at least the c/storage destination) can call `RecordDigestUncompressedPair` with data which violates the documented semantics of the blob info cache, potentially breaking its security (without an extra analysis that it’s OK).

I think eventually we will need to look into whether the blob info cache should also maintain TOC digests (and whether they need a separate namespace or whether they can just be treated as compressed digests). That might be safe, and not too much work, but it requires an analysis. Until that work is done, this PR at least must not break the current documented semantics.

storage/storage_dest.go
```diff
@@ -609,8 +641,28 @@ func (s *storageImageDestination) commitLayer(index int, info addedLayerInfo, si
...
+		configOCI := &imgspecv1.Image{}
+		if err := json.Unmarshal(cb, configOCI); err != nil {
+			return err
+		}
+		if index >= len(configOCI.RootFS.DiffIDs) {
+			return fmt.Errorf("index %d out of range for configOCI.RootFS.DiffIDs", index)
+		}
+
+		// let the storage layer know what was the original uncompressed layer.
+		diffOutput.UncompressedDigest = configOCI.RootFS.DiffIDs[index]
```
mtrmac, line 664: Isn’t this an untrusted unverified value? What does storing that to c/storage do to the security model?

storage/storage_dest.go
```diff
@@ -609,8 +641,28 @@ func (s *storageImageDestination) commitLayer(index int, info addedLayerInfo, si
 			return err
 		}
 
-		// FIXME: what to do with the uncompressed digest?
-		diffOutput.UncompressedDigest = info.digest
+		man, err := manifest.FromBlob(s.manifest, manifest.GuessMIMEType(s.manifest))
```
mtrmac, line 644: Oops, on the `queueOrCommit` path this value is not yet available.

mtrmac (commented):
A very quick skim…

I can’t see that the current version deals with most (any?) of the fallout of redefining `srcInfo.Digest` from `LayerInfosForCopy` to not be a matching digest; some of the relevant code went away but the problem did not.

This version suggests that `GetBlob` can be implemented in a way where it can reproduce the full layer matching `OverrideDigest` — in that case it would be much less disruptive to return that full value in `LayerInfosForCopy`’s `BlobInfo.Digest`, and to look for the right c/storage layer inside `storageImageSource.GetBlob`.

storage/storage_dest.go
giuseppe: It is not done for security reasons, only to validate that we generate the expected diffID.  If you see no value in doing it, I can drop this part.

In this model even computing the digest for the whole .tar file wouldn't add anything to the security model, since we don't use it.  We only use the TOC digest that was validated.  So there is no risk that a malicious image could pretend to have a different TOC (and risk reusing it with another image) than the one we've validated

giuseppe:
> I can’t see that the current version deals with most (any?) of the fallout of redefining `srcInfo.Digest` from `LayerInfosForCopy` to not be a matching digest; some of the relevant code went away but the problem did not.

I've changed the implementation now and dropped the `OverrideDigest`, now  the entire "TOC digest -> uncompressed digest -> TOC digest" conversion happens in `storage_src.go` without any modification to the validation part.

giuseppe:
this is the change needed in c/storage: https://github.com/containers/storage/pull/1719

storage/storage_dest.go
```diff
@@ -604,13 +636,38 @@ func (s *storageImageDestination) commitLayer(index int, info addedLayerInfo, si
 	diffOutput, ok := s.diffOutputs[info.digest]
 	s.lock.Unlock()
 	if ok {
+		if s.manifest == nil {
+			logrus.Debugf("Skipping commit for TOC=%q, manifest not yet available")
+			return nil
```
mtrmac, line 641: (Just looking at the recent changes)

I can’t see how just silently indicating success without doing anything else can work here. The push code is then going to call `commitLayer` for the following layer, and incorrectly set `lastLayer` to `""`.

copy/compression.go
```diff
@@ -57,7 +57,7 @@ func blobPipelineDetectCompressionStep(stream *sourceStream, srcInfo types.BlobI
 		decompressor: decompressor,
 	}
 	if res.isCompressed {
-		res.srcCompressorName = format.Name()
+		res.srcCompressorName = format.InternalUnstableUndocumentedMIMEQuestionMark()
```
mtrmac, line 60: - At which point do I put my feet down and say that this value must be documented properly before any new users are added? Surely no later than today.
- Reading the commit message, I don’t know what “since zstd and zstd:chunked would conflict in this case.” means.

(I’d also tentatively suggest that a set of very small PRs is more likely to be merged than one growing one.)

copy/compression.go
```diff
@@ -157,7 +157,7 @@ func (ic *imageCopier) bpcCompressUncompressed(stream *sourceStream, detected bp
 // bpcRecompressCompressed checks if we should be recompressing a compressed input to another format, and returns a *bpCompressionStepData if so.
 func (ic *imageCopier) bpcRecompressCompressed(stream *sourceStream, detected bpDetectCompressionStepData) (*bpCompressionStepData, error) {
 	if ic.c.dest.DesiredLayerCompression() == types.Compress && detected.isCompressed &&
-		ic.compressionFormat != nil && ic.compressionFormat.Name() != detected.format.Name() {
+		ic.compressionFormat != nil && ic.compressionFormat.InternalUnstableUndocumentedMIMEQuestionMark() != detected.format.InternalUnstableUndocumentedMIMEQuestionMark() {
```
mtrmac, line 160: If we have a non-chunked zstd, and the user asks for chunked, shouldn’t we recompress?

Similarly if there is a non-chunked zstd on a registry, and the user asks for chunked push, shouldn’t we re-compress and re-push?

So at a quick glance this needs a more accurate modeling of “is at least as capable as”, or something.

storage/storage_dest.go
giuseppe: from what I could see, `commitLayer` is called twice.   When it is called from `queueOrCommit` the `s.manifest` is nil so we avoid committing the layer at this stage.

Then it is called again from `Commit` and this time `s.manifest` is present and the layer is created.

With partial images, there is no much gain from the parallel `queueOrCommit` optimization as anyway the commit phase is just renaming a directory so it is fine to do it sequentially from `Commit`

mtrmac: Yes, but it’s the _first_ `commitLayer` of a child of the layer in question that is, I think, going to create problems (a layer with no parent instead of the right parent), and a second commit of the child is then going to be a no-op.

At a glance, it seems to me that `commitLayer` needs to be able to say “not yet, sorry”, causing the `queueOrCommit` state machine to just give up and wait for the final `Commit` to do all the work. Alternatively, `private.ImageDestination` can be extended so that the caller provides the config blob early to transports that ask for it.

(… but I’m not sure this is on the critical path for now. I am still deeply worried about how `TryReusingBlobWithOptions` finds a TOC-based match, in the first place, and I don’t remember, any more, all of the details of how that works / interacts with this DiffID lookup. This might be an ignorant nonsense, but it seems to me there is _some_ chance that changing how reuse happens would impact how this commit part is done. These local comments on `commitLayer` are not an indication that I understand the big picture, or have a strong opinion about it, right now..)

copy/compression.go
giuseppe: if you just inspect the tar blob, it is a regular zstd file.

Not sure how we can treat "zstd:chunked" as a completely separated algorithm.  I'll drop this patch for now and try something else

mtrmac: Guessing, maybe the `blobPipelineDetectCompressionStep` could (itself, or preferably delegating to `pkg/compression`) combine the blob inspection with data from `srcInfo.Annotations`.

giuseppe: what do you think of the approach here?

- https://github.com/containers/storage/pull/1756
- https://github.com/containers/image/pull/2183

mtrmac: See in there; in short, AFAICS we _must_ have the annotation values for the chunked format to be consumable.

giuseppe:
@mtrmac @rhatdan I've reduced the scope of the PR, can we move this forward if there is nothing blocking and deal with details/improvements incrementally later?

storage/storage_dest.go
```diff
@@ -416,6 +416,23 @@ func (s *storageImageDestination) tryReusingBlobAsPending(digest digest.Digest,
 		}
 	}
 
+	// Check if we have a chunked layer in storage with the same TOC digest.
+	if options.TOCDigest != nil {
+		// Check if we have a layer in storage that's based on that TOC digest.
+		layers, err := s.imageRef.transport.store.LayersByUncompressedDigest(*options.TOCDigest)
```
mtrmac, line 422: If nothing else, this still seems semantically ambiguous, after all those months, and I can’t recall ever seeing an argument why it is safe.

giuseppe: the TOC is validated by its digest and we store only once it was validated.  The TOC contains the checksum for every file in the layer, and these files were also validated by their digest. 

Or do you have other concerns?

giuseppe: and AFAICS, we still need to store something in the "uncompressed" digest, even if we expose a "LayersByTOCDigest" functionality, as it is used everywhere

mtrmac: > > It is possible (I provided an example at one point) to construct a byte sequence which is both a valid tar layer, and a valid TOC. So this could trigger reuse of a previously-pulled layer with completely different contents. (And once a layer is reused on this path, we don’t re-validate all of it against the digest, so that mismatch would not be detected.)
> 
> Shouldn't the targeted tarball layer be constructed in a way that its prefix is a JSON TOC (even if just `{}`)? How could someone target another tarball with a malicious TOC? In any case, I've opened a PR to make sure the TOC doesn't contain any extra data after the JSON document: [containers/storage#1766](https://github.com/containers/storage/pull/1766).

Yeah, that was the proof of concept. I don’t know that it’s the only possible way. It’s quite possible that an argument can be built that it is now unambiguous (i.e. that a content couldn’t be parsed as both things, not by theoretical rules but also by the actual implementations we have), but that argument _has to be built_ for us to rely on that.

> > I also looked into this and it seems to me that the existing code requires `UncompressedDigest` to be set for a layer. _shrug_, if the infrastructure to deal with this correctly is not ready, it needs to be built.
> 
> this is much more work, can't it be done later once we are sure this is the way we want to go?

We’ve been talking about this for what? 6 months? I have no patience left with “this would take too much time, let’s ignore this issue and proceed anyway. We have had all the time necessary to work on this.

(And yes, I realize I’m probably signed up to do this work. I’ll get to it.)

mtrmac: IOW, if we defer this and don’t deal with it now, when _are_ we going to deal with it? There is always going to be some more urgent future feature. And we’ve had this ambiguity ever since the first version of `zstd:chunked` was included, and we didn’t get to deal with it so far. To me, that’s _evidence_ that deferring doesn’t work.

storage/storage_src.go
```diff
@@ -42,8 +42,14 @@ type storageImageSource struct {
 	getBlobMutex    sync.Mutex              // Mutex to sync state for parallel GetBlob executions
 	SignatureSizes  []int                   `json:"signature-sizes,omitempty"`  // List of sizes of each signature slice
 	SignaturesSizes map[digest.Digest][]int `json:"signatures-sizes,omitempty"` // List of sizes of each signature slice
+
+	// fromDigestToTOC converts from a layer diffID returned by LayerInfosForCopy to the TOC digest.  The TOC digest
+	// must be used for lookups in the underyling storage.
+	fromDigestToTOC map[digest.Digest]digest.Digest
 }
 
+const expectedLayerDiffIDFlag = "expected-layer-diffid"
```
mtrmac, line 51: What is setting this? If it is c/storage, presumably it should define a public constant.

(And, something somewhere should _precisely_ document the semantics. Is it “this is the digest that we expect, and if it doesn’t match, something is badly broken, but we haven’t (yet?) validated this DiffID, so this value must not be used for lookups”?)

storage/storage_src.go
```diff
@@ -42,8 +42,14 @@ type storageImageSource struct {
 	getBlobMutex    sync.Mutex              // Mutex to sync state for parallel GetBlob executions
```
mtrmac, line 42: We should precisely document which fields are handled by `getBlobMutex`.

Typically in c/image, we’d move it into a separate “section”, starting with `getBlobMutex`, and then say “all following fields can only be accessed with lock held”. Or alternatively,
```go
getBlobMutex sync.Mutex
getBlobMutexProtected struct {
// … protected fields
}
```
but that makes initialization of the containing structure annoying, we’d need to name the embedded type.

storage/storage_src.go
```diff
@@ -91,6 +98,18 @@ func (s *storageImageSource) Close() error {
...
 	digest := info.Digest
+
+	digestIsTOC := false
+
+	// If the digest was overriden by LayerInfosForCopy, then we need to use the TOC digest
+	// to retrieve it from the storage.
+	s.getBlobMutex.Lock()
+	if v, ok := s.fromDigestToTOC[digest]; ok {
+		digest = v
+		digestIsTOC = true
```
mtrmac, line 109: I’m not too happy with a variable that means “that other value with a clear name actually means something else entirely if this is set”.

But see elsewhere about storing a -> layer.ID mapping instead; that would make this distinction unnecessary.

At the very least, this can be moved below the `GzippedEmptyLayerDigest`, and then the this `if` and the `if digestIsTOC` below can be folded into a single condition.

mtrmac (commented):
As partially a note to self, and to be clear there is progress:

Looking _only_ at `storage_src.go`: overall approach and security LGTM.

storage/storage_src.go
giuseppe: we are setting it in `storage_dest.go`: https://github.com/containers/image/pull/1980/files/cbea9be44d374151757269f4fec431acad059206#diff-2d4a826a6986cd3a8f73b9fb0e68b27e87caaefb0703a5b1695255e4d22a7d5aR673-R678

storage/storage_src.go
```diff
@@ -267,14 +291,29 @@ func (s *storageImageSource) LayerInfosForCopy(ctx context.Context, instanceDige
...
 		}
+
+		blobDigest := layer.UncompressedDigest
+
+		if layer.Flags != nil {
+			if v, ok := layer.Flags[expectedLayerDiffIDFlag]; ok {
+				if expectedDigest, ok := v.(string); ok {
+					// if the layer is stored by its TOC, report the expected diffID as the layer Digest
+					// but store the TOC digest so we can later retrieve it from the storage.
+					blobDigest = digest.Digest(expectedDigest)
```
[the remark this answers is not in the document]
giuseppe: thanks, fixed now

storage/storage_src.go
```diff
@@ -267,14 +291,29 @@ func (s *storageImageSource) LayerInfosForCopy(ctx context.Context, instanceDige
...
+					s.getBlobMutex.Lock()
+					s.fromDigestToTOC[blobDigest] = layer.TOCDigest
+					s.getBlobMutex.Unlock()
```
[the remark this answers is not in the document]
giuseppe: good idea, changed now

storage/storage_src.go
giuseppe: fixed

storage/storage_src.go
giuseppe: thanks, fixed

storage/storage_src.go
```diff
@@ -267,14 +297,35 @@ func (s *storageImageSource) LayerInfosForCopy(ctx context.Context, instanceDige
...
+				return nil, fmt.Errorf("TOC digest %q for layer %q is present but %q flag is not set", layer.TOCDigest, layerID, expectedLayerDiffIDFlag)
+			}
+			if expectedDigest, ok := layer.Flags[expectedLayerDiffIDFlag].(string); ok {
+				// if the layer is stored by its TOC, report the expected diffID as the layer Digest
+				// but store the TOC digest so we can later retrieve it from the storage.
+				blobDigest, err = digest.Parse(expectedDigest)
+				if err != nil {
+					return nil, fmt.Errorf("parsing expected diffID %q for layer %q: %w", expectedDigest, layerID, err)
+				}
+			} else {
```
mtrmac, line 320: Non-blocking nit: Returning early on this failure, and leaving the non-failure case less indented, would be a tiny bit more readable.

storage/storage_src.go
```diff
@@ -267,14 +297,35 @@ func (s *storageImageSource) LayerInfosForCopy(ctx context.Context, instanceDige
...
+
+		blobDigest := layer.UncompressedDigest
+
+		if layer.TOCDigest != "" {
+			if layer.Flags == nil || layer.Flags[expectedLayerDiffIDFlag] == nil {
+				return nil, fmt.Errorf("TOC digest %q for layer %q is present but %q flag is not set", layer.TOCDigest, layerID, expectedLayerDiffIDFlag)
+			}
+			if expectedDigest, ok := layer.Flags[expectedLayerDiffIDFlag].(string); ok {
+				// if the layer is stored by its TOC, report the expected diffID as the layer Digest
+				// but store the TOC digest so we can later retrieve it from the storage.
```
mtrmac, line 315: This comment is no longer accurate now that we record `digestToLayerID`.

storage/storage_src.go
```diff
@@ -100,10 +116,24 @@ func (s *storageImageSource) GetBlob(ctx context.Context, info types.BlobInfo, c
...
 	}
 
-	// Check if the blob corresponds to a diff that was used to initialize any layers.  Our
-	// callers should try to retrieve layers using their uncompressed digests, so no need to
-	// check if they're using one of the compressed digests, which we can't reproduce anyway.
-	layers, _ := s.imageRef.transport.store.LayersByUncompressedDigest(digest)
+	var layers []storage.Layer
+
+	// If the digest was overriden by LayerInfosForCopy, then we need to use the TOC digest
+	// to retrieve it from the storage.
```
mtrmac, line 122: This comment needs updating.

storage/storage_src.go
```diff
@@ -62,9 +74,12 @@ func newImageSource(sys *types.SystemContext, imageRef storageReference) (*stora
 		imageRef:        imageRef,
 		systemContext:   sys,
 		image:           img,
-		layerPosition:   make(map[digest.Digest]int),
 		SignatureSizes:  []int{},
 		SignaturesSizes: make(map[digest.Digest][]int),
+		getBlobMutexProtected: getBlobMutexProtected{
```
mtrmac, line 79: Nit: Please keep the order of `Signature*Sizes` vs. `getBlobMutexProtected` the same in the struct definition and this constructor. (I don’t care which one.)

storage/storage_src.go
mtrmac: I’m sorry, I should have found that.

manifest/oci.go
```diff
@@ -259,9 +260,44 @@ func (m *OCI1) ImageID([]digest.Digest) (string, error) {
 	if err := m.Config.Digest.Validate(); err != nil {
 		return "", err
 	}
+
+	// If there is any layer that is using partial content, we calculate the image ID
+	// in a different way since the diffID cannot be validated as for regular pulled images.
+	for _, layer := range m.Layers {
```
mtrmac, line 266: I think this is not quite correct; it doesn’t matter that the layer _has_ a TOC, it matters whether we have _used and verified_ the TOC vs. the uncompressed digest.

So, the same manifest might turn into several different image IDs depending on whether partial pulls succeeded; and that per-pull-per-layer information will need to be maintained by `storageImageDestination`.

storage/storage_dest.go
```diff
@@ -117,18 +117,18 @@ func newImageDestination(sys *types.SystemContext, imageRef storageReference) (*
...
-		indexToStorageID:      make(map[int]*string),
-		indexToAddedLayerInfo: make(map[int]addedLayerInfo),
-		diffOutputs:           make(map[digest.Digest]*graphdriver.DriverWithDifferOutput),
+		imageRef:                imageRef,
+		directory:               directory,
+		signatureses:            make(map[digest.Digest][]byte),
+		uncompressedOrTocDigest: make(map[digest.Digest]digest.Digest),
+		blobAdditionalLayer:     make(map[digest.Digest]storage.AdditionalLayer),
+		fileSizes:               make(map[digest.Digest]int64),
+		filenames:               make(map[digest.Digest]string),
```
mtrmac, line 126: AFAICS all of this is no longer correct, because it identifies layers by (compressed) total blob digest, but we don’t always use or validate that value. Renaming `blobDiffIDs` to `uncompressedOrTocDigest` documents the cause but isn’t sufficient to make the data reliable. (E.g. consider the case of a manifest with
```
[
 layer1: { digest = sharedDigest } // found as an additional layer / as existing layer / not found and stored in .filenames
 layer2: { digest = sharedDigest, TOCDigest = otherTOCDigest }
 layer3: { digest = otherTotalDigest, TOCDigest = sharedDigest }
]
```
it gets too hard to understand what the keys of the maps mean.)

So I think `tryReusingBlobAsPending`/`putBlobToPendingFile` and `commitLayer` should coordinate a bit more closely, and _primarily_ coordinate by layer index.

Something along these lines, maybe?
- In addition, to `indexTo{Added,Committed}LayerInfo`, maintain `indexTo{Pending,NotNecessarilyCommitted,Known}LayerInfo`, so that `.Commit()` can call `commitLayer` with a full `addedLayerInfo` struct
- In `addedLayerInfo`, track:
  - `individualLayerID`: a string or `[]byte`, not a `digest.Digest`; used to construct the “chain” layer ID in `commitLayer`; currently these are the `blobDiffIDs` values, but if we track them in `addedLayerInfo`, we can actually make them unambiguous: Use DiffID _if it is known_, and use a distinct namespace (`\TOC:…` ?) for layers identified by TOC digest.
    - That would clearly disambiguate TOC-digest layers vs. and total-digest layers, ensure that the `lastLayer` values used creating/deduplicating images account for these distinctions, and help us compute a distinct image ID.
  - Where we found the layer data = what to commit if the layer chain ID does not exist locally. A minimal version of that would just be “use DiffID x (with existing code)” vs. “use TOC y (with new code)”, but we could probably, and cleanly, directly record more of the data found by `tryReusingBlobAsPending` directly into `addedLayerInfo`.
    - That would mostly replace `blobAdditionalLayer`/`diffOutputs`; these are ~write-only apart from clean-up, so we don’t need to index them by total blob ID.
    -  For layers found by DiffID , we could record the layer ID instead of just the DiffID, to avoid unnecessary lookups, similarly to the new `storageImageSource` code.
    - (We would still maintain the existing `filenames`/`fileSizes` for the rare case of images that contain the same locally-nonexistent layer twice.)

storage/storage_dest.go
```diff
@@ -532,18 +555,32 @@ func (s *storageImageDestination) queueOrCommit(index int, info addedLayerInfo)
 	return nil
 }
 
+// getDiffIDOrTOCDigest returns the diffID for the specified digest or the digest for the TOC, if known.
+func (s *storageImageDestination) getDiffIDOrTOCDigest(uncompressedDigest digest.Digest) (digest.Digest, bool) {
```
mtrmac, line 559: It would be better for the TOC digest case to return values which clearly can’t conflict with total digests, maybe by using a format that clearly can’t parse as a `digest.Digest`. Then would be able to can rely on the cryptographic hash properties to avoid unwanted layer chain ID conflicts.

See elsewhere about `addedLayerInfo.individualLayerID`.

storage/storage_dest.go
```diff
@@ -557,68 +594,96 @@ func (s *storageImageDestination) commitLayer(index int, info addedLayerInfo, si
...
+		}
+
+		cb, err := s.getConfigBlob(man.ConfigInfo())
+		if err != nil {
+			return false, err
+		}
+
+		// retrieve the expected uncompressed digest from the config blob.
+		configOCI := &imgspecv1.Image{}
+		if err := json.Unmarshal(cb, configOCI); err != nil {
```
mtrmac, line 660: Hard-coding OCI here is… good enough, but worth at least an explicit MIME type check so that this cleanly breaks if the assumptions are violated.

Alternatively, this could use `OCIConfig`.

mtrmac (commented):
Thanks for https://github.com/containers/storage/pull/1719 ; with that, I think there is a clean path for `storageImageDestination` to be unambiguous, see below about primarily keeping the state in a `map[int]addedLayerInfo`.

Admittedly it’s all a little vague/tentative at this point, do let me know if I’m missing something essential or if the attempt to sketch the idea is not clear enough to be understood.

storage/storage_dest.go
mtrmac: Maybe this refactoring would be better to do separately, as a prerequisite smaller PR.

rhatdan:
As I understand it during the meeting last week, we came to the agreement that merging this PR plus fixups from @giuseppe going forward, is the easiest path forward.  I will merge, and then we can continue working on small fixups.

copy/single.go
```diff
@@ -694,6 +695,13 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, to
 			requiredCompression = ic.compressionFormat
 			originalCompression = srcInfo.CompressionAlgorithm
 		}
+
+		// Check if we have a chunked layer in storage that's based on that blob.  These layers are stored by their TOC digest.
+		tocDigest, err := chunkedToc.GetTOCDigest(srcInfo.Annotations)
```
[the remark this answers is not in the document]
giuseppe: I've opened a PR for c/storage to validate the retrieved data: https://github.com/containers/storage/pull/1794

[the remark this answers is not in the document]
mtrmac: Fixed with https://github.com/containers/image/pull/2243 . Thanks!

types/types.go
```diff
@@ -151,6 +151,9 @@ type BlobInfo struct {
 	// TODO: To remove together with CompressionOperation in re-design to
 	// remove field out of BlobInfo.
 	CryptoOperation LayerCrypto
+	// Incremental is used to indicate that the blob was created using a partial
+	// pull, so the uncompressed digest stored is not correct and must be validated.
+	Incremental bool
```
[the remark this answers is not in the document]
mtrmac: N/A in the current version.

copy/single.go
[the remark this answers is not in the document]
mtrmac: Moving the parsing out into a per-manifest-format handler would be cleaner, but that would also probably require a larger restructuring … some new private `LayerInfos` / `LayerInfosForCopy` which already contains that information . And that would interact with the desire to move edit instructions out of `BlobInfo`.

IOW, much too large to worry about now.

---

The other parts have been resolved between this and #2288 .

copy/single.go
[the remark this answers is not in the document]
mtrmac: This PR ultimately used a different approach.

copy/single.go
mtrmac: `"github.com/containers/storage/pkg/chunked/toc"` was used to isolate the dependency.

storage/storage_dest.go
mtrmac: With #2822, image IDs are decided based on how the layer was actually pulled.

storage/storage_dest.go
mtrmac: `GetTOCDigest` still accepts images with both, but `GetDiffer` rejects images with both TOC annotations after https://github.com/containers/storage/pull/1778. So if we ever find a c/storage layer (to match), we know that TOC digest matches the contents.

copy/blob.go
mtrmac: This was removed.

copy/single.go
mtrmac: This was removed.

copy/single.go
mtrmac: This was removed, now the sources provide `LayerInfosForCopy` to validate against, instead.

mtrmac: Overall, we went with `LayerInfosForCopy` returning (expected, untrusted) uncompressed digests for the TOC-pulled layers.

---

> (On a push, don’t we need to discard the TOC and build one anew while compressing, anyway?)

We do discard them AFAICS: `storageImageSource.LayerInfosForCopy` creates new `BlobInfo` data for each layer, and that does not preserve _any_ annotations. So it also discards the TOC annotations.

internal/imagesource/stubs/supports_store_by_toc_digest.go
```diff
@@ -0,0 +1,20 @@
+package stubs
+
+// NoSupportsStoreByTOCDigestInitialize implements parts of private.ImageSource
+// for transports that don’t support SupportsStoreByTOCDigest().
+// See NoSupportsStoreByTOCDigest() below.
+type NoSupportsStoreByTOCDigestInitialize struct {
```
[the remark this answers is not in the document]
mtrmac: This went away.

internal/imagesource/stubs/supports_store_by_toc_digest.go
```diff
@@ -0,0 +1,20 @@
...
+	return false
+}
+
+// ImplementsSupportsStoreByTOCDigest implements SupportsSupportsStoreByTOCDigest() that returns true.
+type ImplementsSupportsStoreByTOCDigest struct{}
+
+// SupportsStoreByTOCDigest() returns true if SupportsStoreByTOCDigest is supported.
+func (stub ImplementsSupportsStoreByTOCDigest) SupportsStoreByTOCDigest() bool {
+	return true
+}
```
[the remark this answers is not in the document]
mtrmac: This was removed.

copy/blob.go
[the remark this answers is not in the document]
mtrmac: This was removed.

copy/single.go
```diff
@@ -398,6 +398,14 @@ func (ic *imageCopier) compareImageDestinationManifestEqual(ctx context.Context,
 	}, nil
 }
 
+// canSkipValidation returns true if the source supports storing partial images and the image is.
+func (ic *imageCopier) canSkipValidation(ctx context.Context) (bool, error) {
+	if !ic.canSubstituteBlobs {
```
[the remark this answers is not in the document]
mtrmac: This was removed.

internal/private/private.go
```diff
@@ -20,6 +20,9 @@ type ImageSourceInternalOnly interface {
 	// BlobChunkAccessor.GetBlobAt is available only if SupportsGetBlobAt().
 	BlobChunkAccessor
 
+	// SupportsStoreByTOCDigest returns true if the transport supports storing blobs by their TOC digest and if they can be trusted.
+	SupportsStoreByTOCDigest() bool
```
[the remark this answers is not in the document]
mtrmac: This was removed.

oci/archive/oci_src.go
```diff
@@ -30,6 +31,7 @@ func (e ImageNotFoundError) Error() string {
 
 type ociArchiveImageSource struct {
 	impl.Compat
+	stubs.ImplementsSupportsStoreByTOCDigest
```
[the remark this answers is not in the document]
mtrmac: This was removed.

pkg/blobcache/src.go
```diff
@@ -23,6 +24,7 @@ import (
 
 type blobCacheSource struct {
 	impl.Compat
+	stubs.ImplementsSupportsStoreByTOCDigest
```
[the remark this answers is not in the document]
mtrmac: This was removed.

copy/single.go
```diff
@@ -398,6 +398,14 @@ func (ic *imageCopier) compareImageDestinationManifestEqual(ctx context.Context,
...
+		return false, nil
+	}
+	return ic.c.rawSource.SupportsStoreByTOCDigest(), nil
```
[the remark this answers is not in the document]
mtrmac: We now validate all layers.

storage/storage_dest.go
```diff
@@ -416,6 +417,28 @@ func (s *storageImageDestination) tryReusingBlobAsPending(digest digest.Digest,
...
 	}
 
+	// Check if we have a chunked layer in storage that's based on that blob.  These layers are stored by their TOC digest.
+	toc, err := chunkedToc.GetTOCDigest(options.Annotations)
+	if err != nil {
+		return false, private.ReusedBlob{}, err
+	}
+	if toc != nil {
+		// Check if we have a layer in storage that's based on that TOC digest.
+		layers, err := s.imageRef.transport.store.LayersByUncompressedDigest(*toc)
```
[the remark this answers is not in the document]
mtrmac: There is now a `LayersByTOCDigest`, and unambiguous with #2288.

storage/storage_dest.go
```diff
@@ -416,6 +417,28 @@ func (s *storageImageDestination) tryReusingBlobAsPending(digest digest.Digest,
...
+		if err != nil && !errors.Is(err, storage.ErrLayerUnknown) {
+			return false, private.ReusedBlob{}, fmt.Errorf(`looking for layers with TOC digest %q: %w`, *toc, err)
+		}
+		if len(layers) > 0 {
+			// Save this for completeness.
+			s.blobDiffIDs[digest] = layers[0].UncompressedDigest
```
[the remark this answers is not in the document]
mtrmac: Resolved in #2288.

storage/storage_dest.go
```diff
@@ -445,16 +469,42 @@ func (s *storageImageDestination) computeID(m manifest.Manifest) string {
...
 		}
-	case *manifest.Schema2, *manifest.OCI1:
-		// We know the ID calculation for these formats doesn't actually use the diffIDs,
-		// so we don't need to populate the diffID list.
+	case *manifest.Schema2:
+		// We know the ID calculation doesn't actually use the diffIDs, so we don't need to populate
+		// the diffID list.
+	case *manifest.OCI1:
+		for _, layer := range m.Layers {
+			toc, err := chunkedToc.GetTOCDigest(layer.Annotations)
```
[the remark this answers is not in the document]
mtrmac: Resolved in #2288.

storage/storage_dest.go
```diff
@@ -445,16 +469,42 @@ func (s *storageImageDestination) computeID(m manifest.Manifest) string {
...
 	default:
 		return ""
 	}
 	id, err := m.ImageID(diffIDs)
 	if err != nil {
 		return ""
 	}
+
+	// If the image is using any partial layer, let's generate a new image ID by combining the
+	// original image ID with the TOC digest of the partial layers.
```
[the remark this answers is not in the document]
mtrmac: These immediate collisions were fixed here, and ultimately image IDs were made unambiguous in #2288.

storage/storage_dest.go
```diff
@@ -532,6 +582,18 @@ func (s *storageImageDestination) queueOrCommit(index int, info addedLayerInfo)
 	return nil
 }
 
+// getDiffIDFromDigest returns the diffID for the specified digest, if known.
+func (s *storageImageDestination) getDiffIDFromDigest(uncompressedDigest digest.Digest) (digest.Digest, bool) {
```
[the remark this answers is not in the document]
mtrmac: Resolved in #2288.

storage/storage_dest.go
mtrmac: This was fixed by storing the untrusted value in a layer flag instead.

storage/storage_dest.go
mtrmac: This was fixed here by returning a `stopQueue` boolean.

copy/compression.go
mtrmac: This was removed from this PR. I vaguely remember we ended up with #2185 instead, for now.

copy/compression.go
mtrmac: This was removed from this PR. I vaguely remember we ended up with #2185 instead, for now.

storage/storage_dest.go
mtrmac: This was eventually made unambiguous in #2288.

copy/single.go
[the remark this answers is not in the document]
mtrmac: This was fixed in https://github.com/containers/storage/pull/1778 , at least on the path where use the TOC. And if we lookup / reuse a layer which contains ambiguous in-manifest metadata, that doesn’t break our local records.

manifest/oci.go
mtrmac: This was resolved in #2288.

storage/storage_dest.go
```diff
@@ -416,6 +416,25 @@ func (s *storageImageDestination) tryReusingBlobAsPending(digest digest.Digest,
 		}
 	}
 
+	tocDigest := digest
```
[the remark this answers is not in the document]
mtrmac: This was ultimately made unambiguous in #2288.

storage/storage_dest.go
```diff
@@ -438,16 +457,20 @@ func (s *storageImageDestination) computeID(m manifest.Manifest) string {
...
 			diffIDs = append([]digest.Digest{diffID}, diffIDs...)
 		}
-	case *manifest.Schema2, *manifest.OCI1:
-		// We know the ID calculation for these formats doesn't actually use the diffIDs,
-		// so we don't need to populate the diffID list.
+	case *manifest.Schema2:
+		// We know the ID calculation doesn't actually use the diffIDs, so we don't need to populate
+		// the diffID list.
+	case *manifest.OCI1:
+		for _, l := range m.Layers {
```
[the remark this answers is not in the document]
mtrmac: This was ultimately resolved in #2288.

storage/storage_dest.go
```diff
@@ -557,68 +594,96 @@ func (s *storageImageDestination) commitLayer(index int, info addedLayerInfo, si
...
+		if err != nil {
+			return false, err
+		}
+
+		// retrieve the expected uncompressed digest from the config blob.
+		configOCI := &imgspecv1.Image{}
+		if err := json.Unmarshal(cb, configOCI); err != nil {
+			return false, err
+		}
+		if index >= len(configOCI.RootFS.DiffIDs) {
```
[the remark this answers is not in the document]
mtrmac: That will happen in #2295 .

storage/storage_dest.go
```diff
@@ -416,6 +416,25 @@ func (s *storageImageDestination) tryReusingBlobAsPending(digest digest.Digest,
...
+	// Check if we have a chunked layer in storage with the same TOC digest.
+	layers, err = s.imageRef.transport.store.LayersByTOCDigest(tocDigest)
+	if err != nil && !errors.Is(err, storage.ErrLayerUnknown) {
+		return false, private.ReusedBlob{}, fmt.Errorf(`looking for layers with TOC digest %q: %w`, tocDigest, err)
+	}
+	if len(layers) > 0 {
+		// Save this for completeness.
+		s.uncompressedOrTocDigest[digest] = layers[0].TOCDigest
+		return true, private.ReusedBlob{
+			Digest: layers[0].TOCDigest,
```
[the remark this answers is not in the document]
mtrmac: This was fixed in #2288 + #2294.

storage/storage_dest.go
mtrmac: The correctness issue was ultimately resolved in #2288 , in a less disruptive way.

storage/storage_dest.go
mtrmac: This was ultimately resolved in #2288.

storage/storage_src.go
mtrmac: This was documented in #2288.

storage/storage_src.go
mtrmac: This was resolved in https://github.com/containers/image/pull/2287.

PR #1982, opened 2023-06-04
copy/multiple: split selection of images to be copied in `copyMultipleImages`

flouthoc:
Following PR splits `copyMultipleImages` into a new function called `prepareEditList` which returns a list of `ListEdit`.

`prepareEditList` does that in these two steps

* How to copy via `ListOperation`
* What to copy by the content of the returned list slice of `ListEdit`

See newly added unit test to see how it is used.


flouthoc:
@mtrmac PTAL

mtrmac (commented):
This works well enough right now, but it will ~immediately need to be replaced when implementing the “add a new Zstd variant” feature; there’s no _clean_ way to represent that.

We could hack that with `ListOperation: ListOpAdd, AddDigest: theOldDigestToConvert`; that would be ugly, but it would work _once_ for the “add a Zstd variant” feature. But then someone may want an “add a gzip variant”… and we are missing a field where to recordwhat kind of modification we should make.

So I think it’s cleaner to add a separate struct immediately, like the original version of #1883 did e.g. in https://github.com/containers/image/commit/941cd13913ad500bda63b05c86bcfd546995b4c3 .

copy/multiple.go
```diff
@@ -12,11 +12,35 @@ import (
...
 	"golang.org/x/exp/slices"
 )
 
+// prepareEditList prepares a list of instances which needs to edited without setting their content,
+// reason is to be able to test `what and how to add/edit` in the manifest list
+func prepareEditList(instanceDigests []digest.Digest, options *Options) ([]internalManifest.ListEdit, int) {
+	imagesToCopy := len(instanceDigests)
+	if options.ImageListSelection == CopySpecificImages {
+		imagesToCopy = len(options.Instances)
+	}
```
mtrmac, line 27: (Assuming we build a list of copies to perform, `imagesToCopy` will be just a `len(theReturnedList)` and wil not need to be computed separately.)

flouthoc:
@mtrmac How about new struct could look like this ?

```go
type addOperation int

const (
	addop_compress_gzip addOperation = iota
	addop_compress_zstd
)

type instanceCopy struct {
        instance internal.ListEdit
        addOp addOperation
}

```

mtrmac:
- That structure wouldn’t allow a non-converting copy that just copies layers as is
- I don’t know what `prepareEditList` would fill in for most of the fields of `instance`.


---

I was thinking vaguely
```go
/* enum instanceCopyKind {
   instanceCopyCopy // a single copy, replacing the original
   instanceCopyClone // creating a new copy of the original, not replacing it
} */

type instanceCopy {
   op instanceCopyKind
   sourceDigest // digest in the source == instance ID
   // otherCopyOptions: for now maybe just
   forceZstd bool
}
```

That’s assuming that after we do an instance copy, we have some _other_ code to determine that we need to adjust the Zstd annotation, and that the annotation update doesn’t need to be represented here.

Maybe we don’t need a new `instanceCopyKind` enum (for now? at all?), just a `clone bool` (or `createCloneInstance`?) would be sufficient.

copy/multiple.go
flouthoc: Done.

copy/multiple.go
```diff
@@ -12,11 +12,42 @@ import (
...
+type instanceCopyKind int
+
+const (
+	InstanceCopyCopy instanceCopyKind = iota
+	InstanceCopyClone
+)
+
+type instanceCopy struct {
+	op           instanceCopyKind
+	sourceDigest digest.Digest
```
flouthoc, line 30: will add `forceZstd` later on when the feature is actually going to be here.

copy/multiple.go
```diff
@@ -12,11 +12,42 @@ import (
...
 	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/exp/slices"
 )
 
+type instanceCopyKind int
+
+const (
+	InstanceCopyCopy instanceCopyKind = iota
+	InstanceCopyClone
```
mtrmac, line 25: Make this private please

copy/multiple.go
```diff
@@ -88,34 +119,31 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
...
+		// populate necessary fields.
+		if instance.op == InstanceCopyCopy {
+			logrus.Debugf("Copying instance %s (%d/%d)", instance.sourceDigest, i+1, len(instanceDigests))
+			c.Printf("Copying image %s (%d/%d)\n", instance.sourceDigest, instancesCopied+1, imagesToCopy)
+			unparsedInstance := image.UnparsedInstance(c.rawSource, &instanceCopyList[i].sourceDigest)
+			updatedManifest, updatedManifestType, updatedManifestDigest, err := c.copySingleImage(ctx, policyContext, options, unparsedToplevel, unparsedInstance, &instanceCopyList[i].sourceDigest)
+			if err != nil {
+				return nil, fmt.Errorf("copying image %d/%d from manifest list: %w", instancesCopied+1, imagesToCopy, err)
+			}
+			instancesCopied++
```
mtrmac, line 138: `instancesCopied` is now equal to `i` or `i + 1`; maybe we don’t need the `instancesCopied` and `imagesToCopy` variables, I’m not sure.

copy/multiple.go
```diff
@@ -88,34 +119,31 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
...
-		if err != nil {
-			return nil, fmt.Errorf("copying image %d/%d from manifest list: %w", instancesCopied+1, imagesToCopy, err)
+	instanceCopyList := prepareCopyInstances(instanceDigests, options)
+	c.Printf("Copying %d of %d images in list\n", len(instanceCopyList), len(instanceDigests))
+	for i, instance := range instanceCopyList {
+		// Update instances to be edited by their `ListOperation` and
+		// populate necessary fields.
+		switch instance.op {
+		case instanceCopyCopy:
+			logrus.Debugf("Copying instance %s (%d/%d)", instance.sourceDigest, i+1, len(instanceDigests))
```
mtrmac, line 130: This matches previous practice, but saying (index of image in copies / total instances) doesn’t make that much sense, because the nominator and denominator count different things.

So I think making the denominator `len(instanceCopyList)` makes more sense, same as the `Printf` below.

copy/multiple_test.go
```diff
@@ -0,0 +1,36 @@
...
+	"testing"
+
+	digest "github.com/opencontainers/go-digest"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestPrepareCopyInstances(t *testing.T) {
+	// Test CopyAllImages
+	options := Options{}
+	instances := []digest.Digest{
```
mtrmac, line 13: (Absolutely non-blocking: Maybe
```suggestion
	sourceInstances := []digest.Digest{
```
)

copy/multiple_test.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"testing"
+
+	digest "github.com/opencontainers/go-digest"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestPrepareCopyInstances(t *testing.T) {
+	// Test CopyAllImages
+	options := Options{}
```
mtrmac, line 12: (Absolutely non-blocking: This can be inlined into the only user.)

copy/multiple_test.go
```diff
@@ -0,0 +1,36 @@
...
+	assert.Equal(t, instancesToCopy[2].sourceDigest, instances[2])
+
+	// Test with CopySpecific Images
+	copyOnly := []digest.Digest{
+		digest.Digest("sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"),
+	}
+	options = Options{
+		Instances:          copyOnly,
+		ImageListSelection: CopySpecificImages,
+	}
```
mtrmac, line 32: (Absolutely non-blocking: This can be inlined into the only user.)

mtrmac (commented):
Thanks! A few more cleanups, please.

flouthoc:
@mtrmac PTAL again :)

mtrmac (approved):
Thanks! LGTM.

flouthoc:

PR #1983, opened 2023-06-04
Simple docker unit tests with network faking

pgillich:
These unit tests are faking network responses from docker repository. So coverage is increased without a real docker repository on the network.


types/types.go
```diff
@@ -671,6 +672,26 @@ type SystemContext struct {
...
 	// CompressionLevel specifies what compression level is used
 	CompressionLevel *int
+
+	// testClientRoundTripper is a http.Client.Transport hook for fake responses
+	testClientRoundTripper http.RoundTripper
+}
+
+// RegisterTestClientRoundTripper registers testClientRoundTripper, if set.
+// Calls Transport.RegisterProtocol.
+func (sys *SystemContext) RegisterTestClientRoundTripper(transport *http.Transport, schemes []string) *http.Transport {
```
mtrmac, line 682: I don’t think the internal unit test implementation should leak to the external API like this.

Instead, unit tests (assuming they don’t span multiple packages) can call the private `newImageSource`/`newImageDestination` with an internal-only extra parameter.

copy/copy_test.go
```diff
@@ -0,0 +1,244 @@
...
+	// More detailed logs (do not push to git)
+	// logrus.SetLevel(logrus.DebugLevel)
+	// logrus.SetFormatter(&logrus.TextFormatter{ForceColors: true})
+
+	tempHome, sys := setupTestCase(t)
+	policy := &signature.Policy{
+		Default: []signature.PolicyRequirement{
+			signature.NewPRInsecureAcceptAnything(),
+		},
+		Transports: map[string]signature.PolicyTransportScopes{
```
mtrmac, line 43: With `Default` set, it should be possible to leave this empty.

docker/docker_image_test.go
```diff
@@ -0,0 +1,109 @@
...
+	userRegistriesDirPath := filepath.Join(tempHome, filepath.FromSlash(".config/containers/registries.d"))
+	err = os.MkdirAll(userRegistriesDirPath, 0700)
+	require.NoError(t, err)
+	bigFilesDirPath := filepath.Join(rootDirPath, "containers", "bigcache")
+	err = os.MkdirAll(bigFilesDirPath, 0700)
+	require.NoError(t, err)
+
+	return tempHome, &types.SystemContext{
+		RootForImplicitAbsolutePaths:      rootDirPath,
+		RegistriesDirPath:                 userRegistriesDirPath,
```
mtrmac, line 102: (Setting this to `"/this/does/not/exist"` works well enough for a unit test, without having to actually create a directory.)

mtrmac (commented):
Thanks!

Just to be explicit, most of c/image is currently integration tested via https://github.com/containers/skopeo/tree/main/integration , which is a mandatory part of c/image CI. Are you seeing some real failures, or insufficient coverage there? (It’s quite possible that’s the case.)

Or is there some other specific thing you want to achieve with these tests?

That said, having _fast_ unit tests that trigger failures immediately during development would certainly be useful; it’s just that with the Skopeo tests existing, we couldn’t find the time to do that.

---

Conceptually, I think _unit_ test for the transports like `docker://`, and for the core copy logic in c/image/copy, are independent; a single test for a `docker` → `docker-archive` copy, exercising three separate packages, is really more of an integration test, and doing that at the top level with a Skopeo command seems good enough (given how thin a layer Skopeo itself is over c/image/copy).

---

Managing the fake HTTP responses _manually_ seems to me to be a fairly large hassle, both to potentially extend, and to maintain, e.g. when the real-world responses change or we need to handle more kinds of them.

Compare old https://github.com/containers/image/pull/254 for an alternative — record real-word interactions as files, not as Go code, and if something changes, just re-record them.

PR #1987, opened 2023-06-07
copy: implement `instanceCopyClone` for `zstd` compression

flouthoc:
* `copy.Options` now contains a new field `EnsureCompressionVariantExists map[string]int` which allows users to specify if they want to clone images with specified `compression`.

* `copyMultipleImages` now implements `instanceCopyClone` and extends function specifically for `zstd` compression.

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -17,6 +17,7 @@ import (
 	"github.com/containers/image/v5/internal/private"
 	"github.com/containers/image/v5/manifest"
 	"github.com/containers/image/v5/pkg/blobinfocache"
+	compression "github.com/containers/image/v5/pkg/compression/types"
 	"github.com/containers/image/v5/signature"
 	"github.com/containers/image/v5/signature/signer"
 	"github.com/containers/image/v5/transports"
@@ -126,6 +127,21 @@ type Options struct {
 	// Download layer contents with "nondistributable" media types ("foreign" layers) and translate the layer media type
 	// to not indicate "nondistributable".
 	DownloadForeignLayers bool
+
+	// Contains slice of OptionCompressionVariant, where copy will ensure that for each platform
+	// in the manifest list, a variant with the requested compression will exist.
+	// Invalid when copying a non-multi-architecture image. That will probably
+	// change in the future.
+	EnsureCompressionVariantsExist []OptionCompressionVariant
+}
+
+// OptionCompressionVariant allows to supply information about
+// selected compression algorithm and compression level by the
+// end-user. Refer to EnsureCompressionVariantsExist to know
+// more about its usage.
+type OptionCompressionVariant struct {
+	Algorithm compression.Algorithm
+	Level     *int // Only used when we are creating a new image instance using the specified algorithm, not when the image already contains such an instance
 }
 
 // copier allows us to keep track of diffID values for blobs, and other
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -266,6 +266,9 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
 	}
 
 	if !multiImage {
+		if len(options.EnsureCompressionVariantsExist) > 0 {
+			return nil, fmt.Errorf("EnsureCompressionVariantsExist is not implemented when not creating a multi-architecture image")
+		}
 		// The simple case: just copy a single image.
 		single, err := c.copySingleImage(ctx, c.unparsedToplevel, nil, copySingleImageOptions{requireCompressionFormatMatch: false})
 		if err != nil {
@@ -273,6 +276,9 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
 		}
 		copiedManifest = single.manifest
 	} else if c.options.ImageListSelection == CopySystemImage {
+		if len(options.EnsureCompressionVariantsExist) > 0 {
+			return nil, fmt.Errorf("EnsureCompressionVariantsExist is not implemented when not creating a multi-architecture image")
+		}
 		// This is a manifest list, and we weren't asked to copy multiple images.  Choose a single image that
 		// matches the current system to copy, and copy it.
 		mfest, manifestType, err := c.unparsedToplevel.Manifest(ctx)
--- a/copy/multiple.go
+++ b/copy/multiple.go
@@ -221,7 +221,7 @@ func (c *copier) copyMultipleImages(ctx context.Context) (copiedManifest []byte,
 	if err != nil {
 		return nil, fmt.Errorf("preparing instances for copy: %w", err)
 	}
-	c.Printf("Copying %d of %d images in list\n", len(instanceCopyList), len(instanceDigests))
+	c.Printf("Copying %d images generated from %d images in list\n", len(instanceCopyList), len(instanceDigests))
 	for i, instance := range instanceCopyList {
 		// Update instances to be edited by their `ListOperation` and
 		// populate necessary fields.
--- a/copy/multiple.go
+++ b/copy/multiple.go
@@ -110,8 +110,7 @@ func prepareInstanceCopies(list internalManifest.List, instanceDigests []digest.
 					clonePlatform:           instanceDetails.ReadOnly.Platform,
 					cloneAnnotations:        maps.Clone(instanceDetails.ReadOnly.Annotations),
 				})
-				// add current compression to the list so logic acks any
-				// duplicates while processing other instances
+				// add current compression to the list so that we don’t create duplicate clones
 				compressionList.Add(compressionVariant.Algorithm.Name())
 			}
 		}
@@ -224,7 +223,10 @@ func (c *copier) copyMultipleImages(ctx context.Context) (copiedManifest []byte,
 			logrus.Debugf("Replicating instance %s (%d/%d)", instance.sourceDigest, i+1, len(instanceCopyList))
 			c.Printf("Replicating image %s (%d/%d)\n", instance.sourceDigest, i+1, len(instanceCopyList))
 			unparsedInstance := image.UnparsedInstance(c.rawSource, &instanceCopyList[i].sourceDigest)
-			updated, err := c.copySingleImage(ctx, unparsedInstance, &instanceCopyList[i].sourceDigest, copySingleImageOptions{requireCompressionFormatMatch: true})
+			updated, err := c.copySingleImage(ctx, unparsedInstance, &instanceCopyList[i].sourceDigest, copySingleImageOptions{
+				requireCompressionFormatMatch: true,
+				compressionFormat:             &instance.cloneCompressionVariant.Algorithm,
+				compressionLevel:              instance.cloneCompressionVariant.Level})
 			if err != nil {
 				return nil, fmt.Errorf("replicating image %d/%d from manifest list: %w", i+1, len(instanceCopyList), err)
 			}
--- a/copy/multiple.go
+++ b/copy/multiple.go
@@ -5,15 +5,18 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sort"
 	"strings"
 
 	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/image/v5/internal/image"
 	internalManifest "github.com/containers/image/v5/internal/manifest"
+	"github.com/containers/image/v5/internal/set"
 	"github.com/containers/image/v5/manifest"
 	digest "github.com/opencontainers/go-digest"
 	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/exp/maps"
 	"golang.org/x/exp/slices"
 )
 
@@ -27,23 +30,97 @@ const (
 type instanceCopy struct {
 	op           instanceCopyKind
 	sourceDigest digest.Digest
+
+	// Fields which can be used by callers when operation
+	// is `instanceCopyClone`
+	cloneCompressionVariant OptionCompressionVariant
+	clonePlatform           *imgspecv1.Platform
+	cloneAnnotations        map[string]string
+}
+
+// internal type only to make imgspecv1.Platform comparable
+type platformComparable struct {
+	architecture string
+	os           string
+	osVersion    string
+	osFeatures   string
+	variant      string
+}
+
+// Converts imgspecv1.Platform to a comparable format.
+func platformV1ToPlatformComparable(platform *imgspecv1.Platform) platformComparable {
+	if platform == nil {
+		return platformComparable{}
+	}
+	osFeatures := slices.Clone(platform.OSFeatures)
+	sort.Strings(osFeatures)
+	return platformComparable{architecture: platform.Architecture,
+		os: platform.OS,
+		// This is strictly speaking ambiguous, fields of OSFeatures can contain a ','. Probably good enough for now.
+		osFeatures: strings.Join(osFeatures, ","),
+		osVersion:  platform.OSVersion,
+		variant:    platform.Variant,
+	}
+}
+
+// platformCompressionMap prepares a mapping of platformComparable -> CompressionAlgorithmNames for given digests
+func platformCompressionMap(list internalManifest.List, instanceDigests []digest.Digest) (map[platformComparable]*set.Set[string], error) {
+	res := make(map[platformComparable]*set.Set[string])
+	for _, instanceDigest := range instanceDigests {
+		instanceDetails, err := list.Instance(instanceDigest)
+		if err != nil {
+			return nil, fmt.Errorf("getting details for instance %s: %w", instanceDigest, err)
+		}
+		platform := platformV1ToPlatformComparable(instanceDetails.ReadOnly.Platform)
+		platformSet, ok := res[platform]
+		if !ok {
+			platformSet = set.New[string]()
+			res[platform] = platformSet
+		}
+		platformSet.AddSlice(instanceDetails.ReadOnly.CompressionAlgorithmNames)
+	}
+	return res, nil
 }
 
 // prepareInstanceCopies prepares a list of instances which needs to copied to the manifest list.
-func prepareInstanceCopies(instanceDigests []digest.Digest, options *Options) []instanceCopy {
+func prepareInstanceCopies(list internalManifest.List, instanceDigests []digest.Digest, options *Options) ([]instanceCopy, error) {
 	res := []instanceCopy{}
+	compressionsByPlatform, err := platformCompressionMap(list, instanceDigests)
+	if err != nil {
+		return nil, err
+	}
 	for i, instanceDigest := range instanceDigests {
 		if options.ImageListSelection == CopySpecificImages &&
 			!slices.Contains(options.Instances, instanceDigest) {
 			logrus.Debugf("Skipping instance %s (%d/%d)", instanceDigest, i+1, len(instanceDigests))
 			continue
 		}
+		instanceDetails, err := list.Instance(instanceDigest)
+		if err != nil {
+			return res, fmt.Errorf("getting details for instance %s: %w", instanceDigest, err)
+		}
+		platform := platformV1ToPlatformComparable(instanceDetails.ReadOnly.Platform)
+		compressionList := compressionsByPlatform[platform]
+		for _, compressionVariant := range options.EnsureCompressionVariantsExist {
+			if !compressionList.Contains(compressionVariant.Algorithm.Name()) {
+				res = append(res, instanceCopy{
+					op:                      instanceCopyClone,
+					sourceDigest:            instanceDigest,
+					cloneCompressionVariant: compressionVariant,
+					clonePlatform:           instanceDetails.ReadOnly.Platform,
+					cloneAnnotations:        maps.Clone(instanceDetails.ReadOnly.Annotations),
+				})
+				// add current compression to the list so logic acks any
+				// duplicates while processing other instances
+				compressionList.Add(compressionVariant.Algorithm.Name())
+			}
+		}
 		res = append(res, instanceCopy{
 			op:           instanceCopyCopy,
 			sourceDigest: instanceDigest,
 		})
 	}
-	return res
+	return res, nil
 }
 
 // copyMultipleImages copies some or all of an image list's instances, using
@@ -118,7 +195,10 @@ func (c *copier) copyMultipleImages(ctx context.Context) (copiedManifest []byte,
 	// Copy each image, or just the ones we want to copy, in turn.
 	instanceDigests := updatedList.Instances()
 	instanceEdits := []internalManifest.ListEdit{}
-	instanceCopyList := prepareInstanceCopies(instanceDigests, c.options)
+	instanceCopyList, err := prepareInstanceCopies(updatedList, instanceDigests, c.options)
+	if err != nil {
+		return nil, fmt.Errorf("preparing instances for copy: %w", err)
+	}
 	c.Printf("Copying %d of %d images in list\n", len(instanceCopyList), len(instanceDigests))
 	for i, instance := range instanceCopyList {
 		// Update instances to be edited by their `ListOperation` and
@@ -140,6 +220,24 @@ func (c *copier) copyMultipleImages(ctx context.Context) (copiedManifest []byte,
 				UpdateSize:                  int64(len(updated.manifest)),
 				UpdateCompressionAlgorithms: updated.compressionAlgorithms,
 				UpdateMediaType:             updated.manifestMIMEType})
+		case instanceCopyClone:
+			logrus.Debugf("Replicating instance %s (%d/%d)", instance.sourceDigest, i+1, len(instanceCopyList))
+			c.Printf("Replicating image %s (%d/%d)\n", instance.sourceDigest, i+1, len(instanceCopyList))
+			unparsedInstance := image.UnparsedInstance(c.rawSource, &instanceCopyList[i].sourceDigest)
+			updated, err := c.copySingleImage(ctx, unparsedInstance, &instanceCopyList[i].sourceDigest, copySingleImageOptions{requireCompressionFormatMatch: true})
+			if err != nil {
+				return nil, fmt.Errorf("replicating image %d/%d from manifest list: %w", i+1, len(instanceCopyList), err)
+			}
+			// Record the result of a possible conversion here.
+			instanceEdits = append(instanceEdits, internalManifest.ListEdit{
+				ListOperation:            internalManifest.ListOpAdd,
+				AddDigest:                updated.manifestDigest,
+				AddSize:                  int64(len(updated.manifest)),
+				AddMediaType:             updated.manifestMIMEType,
+				AddPlatform:              instance.clonePlatform,
+				AddAnnotations:           instance.cloneAnnotations,
+				AddCompressionAlgorithms: updated.compressionAlgorithms,
+			})
 		default:
 			return nil, fmt.Errorf("copying image: invalid copy operation %d", instance.op)
 		}
--- a/copy/multiple.go
+++ b/copy/multiple.go
@@ -13,6 +13,7 @@ import (
 	internalManifest "github.com/containers/image/v5/internal/manifest"
 	"github.com/containers/image/v5/internal/set"
 	"github.com/containers/image/v5/manifest"
+	"github.com/containers/image/v5/pkg/compression"
 	digest "github.com/opencontainers/go-digest"
 	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/sirupsen/logrus"
@@ -82,9 +83,31 @@ func platformCompressionMap(list internalManifest.List, instanceDigests []digest
 	return res, nil
 }
 
+func validateCompressionVariantExists(input []OptionCompressionVariant) error {
+	for _, option := range input {
+		_, err := compression.AlgorithmByName(option.Algorithm.Name())
+		if err != nil {
+			return fmt.Errorf("invalid algorithm %q in option.EnsureCompressionVariantsExist: %w", option.Algorithm.Name(), err)
+		}
+	}
+	return nil
+}
+
 // prepareInstanceCopies prepares a list of instances which needs to copied to the manifest list.
 func prepareInstanceCopies(list internalManifest.List, instanceDigests []digest.Digest, options *Options) ([]instanceCopy, error) {
 	res := []instanceCopy{}
+	if options.ImageListSelection == CopySpecificImages && len(options.EnsureCompressionVariantsExist) > 0 {
+		// List can already contain compressed instance for a compression selected in `EnsureCompressionVariantsExist`
+		// It’s unclear what it means when `CopySpecificImages` includes an instance in options.Instances,
+		// EnsureCompressionVariantsExist asks for an instance with some compression,
+		// an instance with that compression already exists, but is not included in options.Instances.
+		// We might define the semantics and implement this in the future.
+		return res, fmt.Errorf("EnsureCompressionVariantsExist is not implemented for CopySpecificImages")
+	}
+	err := validateCompressionVariantExists(options.EnsureCompressionVariantsExist)
+	if err != nil {
+		return res, err
+	}
 	compressionsByPlatform, err := platformCompressionMap(list, instanceDigests)
 	if err != nil {
 		return nil, err
--- a/copy/multiple_test.go
+++ b/copy/multiple_test.go
@@ -1,36 +1,44 @@
 package copy
 
 import (
+	"os"
+	"path/filepath"
 	"testing"
 
+	internalManifest "github.com/containers/image/v5/internal/manifest"
 	digest "github.com/opencontainers/go-digest"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
-func TestPrepareCopyInstances(t *testing.T) {
+// Test `instanceCopyCopy` cases.
+func TestPrepareCopyInstancesforInstanceCopyCopy(t *testing.T) {
+	validManifest, err := os.ReadFile(filepath.Join("..", "internal", "manifest", "testdata", "oci1.index.zstd-selection.json"))
+	require.NoError(t, err)
+	list, err := internalManifest.ListFromBlob(validManifest, internalManifest.GuessMIMEType(validManifest))
+	require.NoError(t, err)
+
 	// Test CopyAllImages
 	sourceInstances := []digest.Digest{
-		digest.Digest("sha256:5b0bcabd1ed22e9fb1310cf6c2dec7cdef19f0ad69efa1f392e94a4333501270"),
+		digest.Digest("sha256:bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb"),
 		digest.Digest("sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"),
 		digest.Digest("sha256:cccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc"),
 	}
 
-	instancesToCopy := prepareInstanceCopies(sourceInstances, &Options{})
+	instancesToCopy, err := prepareInstanceCopies(list, sourceInstances, &Options{})
+	require.NoError(t, err)
 	compare := []instanceCopy{}
+
 	for _, instance := range sourceInstances {
 		compare = append(compare, instanceCopy{op: instanceCopyCopy,
 			sourceDigest: instance})
 	}
 	assert.Equal(t, instancesToCopy, compare)
 
-	// Test with CopySpecific Images
-	copyOnly := []digest.Digest{
-		digest.Digest("sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"),
-	}
-	instancesToCopy = prepareInstanceCopies(sourceInstances, &Options{
-		Instances:          copyOnly,
-		ImageListSelection: CopySpecificImages})
-	assert.Equal(t, instancesToCopy, []instanceCopy{{
-		op:           instanceCopyCopy,
-		sourceDigest: copyOnly[0]}})
+	// Test CopySpecificImages where selected instance is sourceInstances[1]
+	instancesToCopy, err = prepareInstanceCopies(list, sourceInstances, &Options{Instances: []digest.Digest{sourceInstances[1]}, ImageListSelection: CopySpecificImages})
+	require.NoError(t, err)
+	compare = []instanceCopy{{op: instanceCopyCopy,
+		sourceDigest: sourceInstances[1]}}
+	assert.Equal(t, instancesToCopy, compare)
 }
--- a/copy/multiple_test.go
+++ b/copy/multiple_test.go
@@ -78,7 +78,6 @@ func TestPrepareCopyInstancesforInstanceCopyClone(t *testing.T) {
 	// amd64 already has a zstd instance i.e sourceInstance[1] so it should not create replication for
 	// `sourceInstance[0]` and `sourceInstance[1]` but should do it for `sourceInstance[2]` for `arm64`
 	// and still copy `sourceInstance[2]`.
-	assert.Equal(t, 4, len(instancesToCopy))
 	expectedResponse := []simplerInstanceCopy{}
 	for _, instance := range sourceInstances {
 		// If its `arm64` and sourceDigest[2] , expect a clone to happen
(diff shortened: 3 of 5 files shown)


copy/multiple.go
```diff
@@ -124,8 +134,9 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
 	for i, instance := range instanceCopyList {
 		// Update instances to be edited by their `ListOperation` and
 		// populate necessary fields.
-		switch instance.op {
```
flouthoc, line 127: Had to remove the `switch` cause I was `instanceCopyClone` to be processed in `instanceCopyCopy's block` as well as in its own block for replication.

flouthoc:
@mtrmac WDYT about this, Could you PTAL.

copy/multiple.go
```diff
@@ -39,8 +42,15 @@ func prepareInstanceCopies(instanceDigests []digest.Digest, options *Options) []
 			logrus.Debugf("Skipping instance %s (%d/%d)", instanceDigest, i+1, len(instanceDigests))
 			continue
 		}
+		operation := instanceCopyCopy
+		forceZstd := false
+		if _, ok := options.EnsureCompressionVariantExists["zstd"]; ok {
```
mtrmac, line 47: Some code somewhere, maybe here, needs to reject requests for any formats that we don’t currently support.

---

We _also_ should fail if this is set while copying a non-index image; or, eventually, implement that. (Notably that matters for `podman build && podman push`.

copy/multiple.go
```diff
@@ -140,7 +151,49 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
...
+			unparsedInstance := image.UnparsedInstance(c.rawSource, &instanceCopyList[i].sourceDigest)
+			src, err := image.FromUnparsedImage(ctx, options.SourceCtx, unparsedInstance)
+			if err != nil {
+				return nil, fmt.Errorf("initializing image from source %s: %w", transports.ImageName(c.rawSource.Reference()), err)
+			}
+			imgConfig, err := src.OCIConfig(ctx)
+			if err != nil {
+				return nil, fmt.Errorf("parsing image configuration: %w", err)
+			}
+			imgPlatform := imgspecv1.Platform{Architecture: imgConfig.Architecture,
```
mtrmac, line 168: Why is this parsing the per-instance OCI config at all? We already have a platform specifier in the image we are cloning.

(And, in the wild, the index and config platforms do differ, with the config platform being incorrect because most consumers don’t inspect it. So using the config value is not harmless.)

copy/multiple.go
```diff
@@ -140,7 +151,49 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
...
+				OSVersion:  imgConfig.OSVersion,
+				OSFeatures: imgConfig.OSFeatures,
+				Variant:    imgConfig.Variant}
+			addAnnotations := make(map[string]string)
+			if instance.forceZstd {
+				// add `zstd` annotation
+				addAnnotations[internalManifest.OCI1InstanceAnnotationCompressionZSTD] = "true"
+				// copy options with new compression format
+				optionsClone := options
+				optionsClone.DestinationCtx.CompressionFormat = &compression.Zstd
```
mtrmac, line 179: This doesn’t actually work well enough right now, because we reuse blobs at the destination regardless of compression. So copying a gzip image followed by trying to create a zstd clone would just create another image that links to the same just-uploaded gzip blobs.

We will need the blob reuse code around `copyLayer` to accept an option to use _only_ the requested compression format, and not any other. (I guess we would still ignore the compression level request if a blob exists with the same algorithm but a different level, just because we don’t have a cheap way to determine the compression level.)

copy/multiple.go
```diff
@@ -140,7 +151,49 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
...
+			if err != nil {
+				return nil, fmt.Errorf("parsing image configuration: %w", err)
+			}
+			imgPlatform := imgspecv1.Platform{Architecture: imgConfig.Architecture,
+				OS:         imgConfig.OS,
+				OSVersion:  imgConfig.OSVersion,
+				OSFeatures: imgConfig.OSFeatures,
+				Variant:    imgConfig.Variant}
+			addAnnotations := make(map[string]string)
+			if instance.forceZstd {
```
mtrmac, line 174: BTW _eventually_ we should support `forceZstd` also for a non-clone `instanceCopyCopy`, to implement “take an existing gzip image, and replace all instances with Zstd ones, not copying the gzip variants at all”.

That doesn’t need to exist _today_. But it might be possible structure the code in a way that makes it trivially cheap already in this PR, so it’s something to consider in the background.

(And if this PR doesn’t implement `instanceCopyCopy, forceZstd = true`, the `forceZstd` field should have a comment saying that it only works for `…CopyClone`.)

flouthoc:
> We need to add the internalManifest.OCI1InstanceAnnotationCompressionZSTD annotation even for non-…CopyClone copies (e.g. when users use the existing DestinationCtx.CompressionFormat option). (And that should not be hard-coded in the generic copy code, but something the image and/or manifest format handlers help with.) That will probably also make the new …CopyClone path simpler.

@mtrmac Created a new PR for this https://github.com/containers/image/pull/1992

flouthoc:
I am re-working on this but there is overlapping part with https://github.com/containers/image/pull/1875 , should i just close this one and work on a single https://github.com/containers/image/pull/1875 ? @mtrmac WDYT ?

mtrmac:
I think the immediate next step is to ensure that `skopeo copy --format zstd`, or a simple `copy` that reuses existing Zstd blobs, sets the annotation; I think it would be a bit weird if we had a separate “clone” feature that understood the annotation, but the simpler basic copy didn’t work.

I imagine (without having done the work) that would require some changes to the interface of `TryReusingBlobWithOptions`/`ReusedBlob` — and that touches enough files to probably be worth a separate PR.

---

The next bit of infrastructure before actually triggering the copy is to ensure that layer reuse only reuses blobs with the desired compression algorithm, per https://github.com/containers/image/pull/1987#discussion_r1222045468 . (and perhaps to expose that as an user option, per some of the Buildah bug reports — that could well be separate later.)

That’s also somewhere around `TryReusingBlobWithOptions` — conceptually it’s separate (and with two very independent design discussions); the two could even be done in parallel. OTOH the logistics and merge conflicts might well make it much easier to do these two parts in a single PR.

---

Once _those_ pieces of infrastructure are built, I think the remaining part would really be the top-level caller in `copy/multiple.go`. And maybe a very last step to convert a single-arch image into one with two instances, so that `buildah push` can create the two-instance image.

As to whether to do that in this PR, or in the original #1875, I think if we identify a separate concern that can be implemented and discussed mostly independently, it’s a bit more convenient to file a fresh PR so that there isn’t a history of a previous discussion (and GitHub “Load more” placeholders) to wade through. So I’d very weakly prefer all work to happen in PRs other than #1875, and for #1875 to be used more or less just as a tracker of identified issues / concerns — but I don‘t feel strongly about that at all.

flouthoc:
SGTM I think then I'll keep this PR seperate than #1875 and will open new ones for other tasks its easier to review that way.

flouthoc:
This will need https://github.com/containers/image/pull/2023 to get merged first.

flouthoc:
@mtrmac PTAL, I am using this PR directly in draft/POC PR of buildah and its working correctly for me: https://github.com/containers/buildah/pull/4912

copy/multiple.go
```diff
@@ -140,8 +162,34 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
 				UpdateDigest:    updatedManifestDigest,
 				UpdateSize:      int64(len(updatedManifest)),
 				UpdateMediaType: updatedManifestType})
-		default:
-			return nil, fmt.Errorf("copying image: invalid copy operation %d", instance.op)
+		case instanceCopyClone:
+			c.Printf("Replicating image %s (%d/%d)\n", instance.sourceDigest, i+1, len(instanceCopyList))
```
mtrmac, line 166: When the `instanceCopyCopy` case starts with a `Debugf`, this should probably do the same thing.

copy/multiple.go
```diff
@@ -119,7 +138,10 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
 	// Copy each image, or just the ones we want to copy, in turn.
 	instanceDigests := updatedList.Instances()
 	instanceEdits := []internalManifest.ListEdit{}
-	instanceCopyList := prepareInstanceCopies(instanceDigests, options)
+	instanceCopyList, err := prepareInstanceCopies(instanceDigests, options)
+	if err != nil {
+		return nil, fmt.Errorf("while preparing instances for copy: %w", err)
+	}
 	c.Printf("Copying %d of %d images in list\n", len(instanceCopyList), len(instanceDigests))
```
mtrmac, line 145: This might now say “copying 5 of 3 images”, so the wording could be improved, here and probably in the other  messages around, so that we don’t refer to “image 5/3 in manifest list” and the like.

copy/multiple.go
```diff
@@ -140,8 +162,34 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
...
+			unparsedInstance := image.UnparsedInstance(c.rawSource, &instanceCopyList[i].sourceDigest)
+			instanceDetails, err := updatedList.Instance(instanceCopyList[i].sourceDigest)
+			if err != nil {
+				return nil, fmt.Errorf("Replicating image %d/%d from manifest list: %w", i+1, len(instanceCopyList), err)
+			}
+			addCompressionAlgorithms := []compressionTypes.Algorithm{}
+			optionsClone := options
+			if instance.compressionVariant.Algorithm != nil {
+				// copy options with new compression format
+				optionsClone.DestinationCtx.CompressionFormat = instance.compressionVariant.Algorithm
```
mtrmac, line 176: https://github.com/containers/image/pull/1987#discussion_r1222035562 is still outstanding.

copy/multiple_test.go
```diff
@@ -15,22 +16,24 @@ func TestPrepareCopyInstances(t *testing.T) {
 		digest.Digest("sha256:cccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc"),
 	}
 
-	instancesToCopy := prepareInstanceCopies(sourceInstances, &Options{})
+	instancesToCopy, _ := prepareInstanceCopies(sourceInstances, &Options{})
```
mtrmac, line 19: Actually check the error value, in both cases.

copy/multiple.go
```diff
@@ -140,8 +162,34 @@ func (c *copier) copyMultipleImages(ctx context.Context, policyContext *signatur
...
+				return nil, fmt.Errorf("Replicating image %d/%d from manifest list: %w", i+1, len(instanceCopyList), err)
+			}
+			// Record the result of a possible conversion here.
+			instanceEdits = append(instanceEdits, internalManifest.ListEdit{
+				ListOperation:            internalManifest.ListOpAdd,
+				AddDigest:                addManifestDigest,
+				AddSize:                  int64(len(addManifest)),
+				AddMediaType:             addManifestType,
+				AddPlatform:              instanceDetails.Platform,
+				AddCompressionAlgorithms: addCompressionAlgorithms,
```
mtrmac, line 191: https://github.com/containers/image/pull/1987#discussion_r1222036121 is outstanding.

(And the documentation in `ListEdit` says that “all fields must be set”)

copy/copy.go
```diff
@@ -278,7 +288,7 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
 		logrus.Debugf("Source is a manifest list; copying (only) instance %s for current system", instanceDigest)
 		unparsedInstance := image.UnparsedInstance(rawSource, &instanceDigest)
 
-		if copiedManifest, _, _, err = c.copySingleImage(ctx, policyContext, options, unparsedToplevel, unparsedInstance, nil); err != nil {
+		if copiedManifest, _, _, err = c.copySingleImage(ctx, policyContext, options, unparsedToplevel, unparsedInstance, nil, nil); err != nil {
 			return nil, fmt.Errorf("copying system image from manifest list: %w", err)
 		}
```
mtrmac, line 293: Per https://github.com/containers/image/pull/1987#discussion_r1222041486 , we need to fail if `Ensure…` is set on these paths, instead of silently ignoring it.

copy/multiple.go
mtrmac, line 191: (I assume the values of these fields could be prepared in `prepareInstanceCopies`, it has, or will need to have, that data anyway — so populating fields in `instanceCopy` could make this executing loop simpler, and give us more cheap unit test coverage.)

mtrmac (commented):
> I think the immediate next step is to ensure that `skopeo copy --format zstd`, or a simple `copy` that reuses existing Zstd blobs, sets the annotation

This is still missing AFAIK. (and possibly a separate PR.)

flouthoc:
I have to rebase this after: https://github.com/containers/image/pull/2040 but I have addressed most of the comments.

flouthoc:
> > I think the immediate next step is to ensure that `skopeo copy --format zstd`, or a simple `copy` that reuses existing Zstd blobs, sets the annotation
> 
> This is still missing AFAIK. (and possibly a separate PR.)

I was wondering implementing `skopeo copy --format zstd` after this PR. Since its a UX thing and once this gets merged I am hoping no more infra bits will be needed then I can work on the UX things on buildah and skopeo easily.

mtrmac:
- Just from the users’ POV, I think it’s surprising for a complex “clone if necessary” operation to be fully implemented before a simple “copy with specified format” being correct.
- Setting the annotation on a `skopeo copy --format` is somewhat a prerequisite of this PR: If this PR wants to add a clone when there is no zstd instance, we need the pre-this-PR copies to be correctly marked as zstd.
- I rather expect that the data flow from `skopeo copy --format` setting the annotation can be directly reused in this PR (no need for an extra `instanceCopy.AddCompressionAlgorithms` because `copySingleImage` will build that data).

mtrmac:
Oh, I’m so sorry. The `skopeo copy --format zstd` comment used an incorrect option name (it’s `--dest-compress-format zstd` — and the intent was not to add a new Skopeo option, but to set the _annotation_ automatically.

flouthoc:
> Oh, I’m so sorry. The `skopeo copy --format zstd` comment used an incorrect option name (it’s `--dest-compress-format zstd` — and the intent was not to add a new Skopeo option, but to set the _annotation_ automatically.

Ackd no issue, I will amend the PR but skopeo will still need this https://github.com/containers/image/pull/2047

copy/multiple_test.go
```diff
@@ -1,36 +1,78 @@
...
+		compressionVariant: defaultCompressionVariant}})
+
+	// Test with CopySpecific Image with EnsureCompressionExists
+	copyOnly = []digest.Digest{
+		digest.Digest("sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"),
+	}
+	instancesToCopy, err = prepareInstanceCopies(list, sourceInstances, &Options{
+		Instances:                      copyOnly,
+		EnsureCompressionVariantsExist: []OptionCompressionVariant{{Algorithm: &compression.Zstd}},
+		ImageListSelection:             CopySpecificImages})
```
mtrmac, line 70: It seems to me that without `CopySpecificImages`, a full investigation of the results would show that this causes unwanted clones. So that might also be a good test case. (And possibly, a test case with `Ensure…Exist: Gzip`, with the same source image.)

mtrmac (commented):
Thanks, good improvements.

Most importantly `prepareInstanceCopies` seems not to compute the right thing yet.

WRT parameter passing, it seems to me that `require…` should be a parameter to `copySingleImage`, and that the way `copySingleImage` accepts parameters should be simplified … let me take a step in that direction.

mtrmac:
> it seems to me that `require…` should be a parameter to `copySingleImage`, and that the way `copySingleImage` accepts parameters should be simplified … let me take a step in that direction.

#2048 is, I guess, that step: `copySingleImage` can return a struct (which is more relevant for #2047), and has fewer parameters (which helps for this PR, but I still think that the “comparatively less important” options are better passed as an options struct, just for the benefit of named fields).

flouthoc:
@mtrmac Apart for tests and one dummy commit, I have made all changes, could you PTAL

copy/copy.go
```diff
@@ -126,6 +127,16 @@ type Options struct {
...
+
+	// Contains slice of OptionCompressionVariant, where c/image will ensure that selected
+	// images in the manifest list will exist with compression specificed in OptionCompressionVariant
+	// while performing copy.
+	EnsureCompressionVariantsExist []OptionCompressionVariant
+}
+
+type OptionCompressionVariant struct {
+	Algorithm *compression.Algorithm
+	Level     int
```
mtrmac, line 139: `*int`, to be optional

copy/multiple.go
```diff
@@ -27,23 +32,90 @@ const (
...
 			logrus.Debugf("Skipping instance %s (%d/%d)", instanceDigest, i+1, len(instanceDigests))
 			continue
 		}
+		instanceDetails, err := list.Instance(instanceDigest)
+		if err != nil {
+			return res, fmt.Errorf("getting details for instance %s: %w", instanceDigest, err)
+		}
+		for _, compressionVariant := range options.EnsureCompressionVariantsExist {
+			platform := platformV1ToPlatformComparable(*instanceDetails.ReadOnly.Platform)
+			compressionList := platformCompressionMap[platform]
```
mtrmac, line 101: This can happen outside of the loop.

copy/multiple.go
```diff
@@ -27,23 +32,90 @@ const (
...
+			if !slices.Contains(compressionList.Values(), compressionVariant.Algorithm.Name()) {
```
mtrmac, line 102: Isn’t this just `compressionList.Contains(…Name())`?

copy/multiple.go
```diff
@@ -27,23 +32,90 @@ const (
...
+				res = append(res, instanceCopy{
+					op:                            instanceCopyClone,
+					sourceDigest:                  instanceDigest,
+					compressionVariant:            compressionVariant.Algorithm,
```
mtrmac, line 106: `compressionVariant.Level` is never used?

copy/multiple.go
```diff
@@ -27,23 +32,90 @@ const (
 type instanceCopy struct {
 	op           instanceCopyKind
 	sourceDigest digest.Digest
+
+	// Fields which can be used by callers when operation
+	// is `instanceCopyClone`
+	compressionVariant            *compression.Algorithm
```
mtrmac, line 38:
```suggestion
	cloneCompressionVariant            *compression.Algorithm
```
?

copy/multiple.go
mtrmac, line 102: A design question here is whether we should be looking for any match that contains a Zstd layer, or only for exclusively-Zstd images.

OTOH, right now we can’t tell the difference, so it doesn‘t matter anyway… simpler code is immediately valuable.

copy/multiple.go
```diff
@@ -27,23 +32,90 @@ const (
...
+			compressionList := platformCompressionMap[platform]
+			if !slices.Contains(compressionList.Values(), compressionVariant.Algorithm.Name()) {
+				res = append(res, instanceCopy{
+					op:                            instanceCopyClone,
+					sourceDigest:                  instanceDigest,
+					compressionVariant:            compressionVariant.Algorithm,
+					cloneAddPlatform:              instanceDetails.ReadOnly.Platform,
+					cloneAddAnnotations:           maps.Clone(instanceDetails.ReadOnly.Annotations),
+					cloneAddCompressionAlgorithms: []compressionTypes.Algorithm{*compressionVariant.Algorithm},
+				})
```
mtrmac, line 110: Maybe add the clone’s algorithm to `compressionList` so that we don’t create two clones on duplicate input

internal/set/set_test.go
```diff
@@ -30,6 +30,8 @@ func TestAdd(t *testing.T) {
 	// Unrelated elements are unaffected
 	assert.True(t, s.Contains(1))
 	assert.False(t, s.Contains(3))
+	s.AddSlice([]int{3, 4})
```
mtrmac, line 33: An isolated `TestAddSlice`? Right now the tests are organized mostly one per external function.

copy/multiple.go
```diff
@@ -118,17 +190,23 @@ func (c *copier) copyMultipleImages(ctx context.Context) (copiedManifest []byte,
...
+		c.options.DestinationCtx = &copyOption // THIS IS HACK and will not be merged till: PR#2050
 		// Update instances to be edited by their `ListOperation` and
 		// populate necessary fields.
 		switch instance.op {
 		case instanceCopyCopy:
 			logrus.Debugf("Copying instance %s (%d/%d)", instance.sourceDigest, i+1, len(instanceCopyList))
 			c.Printf("Copying image %s (%d/%d)\n", instance.sourceDigest, i+1, len(instanceCopyList))
 			unparsedInstance := image.UnparsedInstance(c.rawSource, &instanceCopyList[i].sourceDigest)
-			updated, err := c.copySingleImage(ctx, unparsedInstance, &instanceCopyList[i].sourceDigest, false)
+			updated, err := c.copySingleImage(ctx, unparsedInstance, &instanceCopyList[i].sourceDigest, true)
```
mtrmac, line 209: This must be `false` at least in the `cannotModifyManifestListReason` reason cases, and I think it should be `false` in most other cases.

(Notably many callers will not configure any compression format at all, so what does setting `true` mean in that case?)

---

I have changed my mind a bit from that absolute position — I think it could make sense to be more strict if `EnsureCompressionVariantsExist` is set, that’s clearly an opt-in to a new behavior.

OTOH I’m not immediately sure what “more strict” means in corner cases: if the image already has a gzip-exclusive instance, creating a gzip-exclusive instance on the destination clearly makes sense. But if the source image has a mixed-algorithm layer, and because of that we don’t do a Zstd clone, creating what does setting `…match` mean? There is no single algorithm we can set `opt.compressionFormat` to.

copy/multiple.go
```diff
@@ -140,6 +218,32 @@ func (c *copier) copyMultipleImages(ctx context.Context) (copiedManifest []byte,
...
+			unparsedInstance := image.UnparsedInstance(c.rawSource, &instanceCopyList[i].sourceDigest)
+			if instance.compressionVariant != nil {
+				// ---- THIS BLOCK IS A HACK and will not be merged till: https://github.com/containers/image/pull/2050
+				// copy options with new compression format
+				if c.options.DestinationCtx == nil {
+					c.options.DestinationCtx = &types.SystemContext{}
+				}
+				c.options.DestinationCtx.CompressionFormat = instance.compressionVariant
+			}
+			updated, err := c.copySingleImage(ctx, unparsedInstance, &instanceCopyList[i].sourceDigest, true)
```
mtrmac, line 233: (BTW it might make sense for `prepareInstanceCopies` to prepare full `copySingleImage` values, instead of carrying the data one field at a time in `instanceCopy`.)

copy/multiple.go
```diff
@@ -140,6 +218,32 @@ func (c *copier) copyMultipleImages(ctx context.Context) (copiedManifest []byte,
...
+			}
+			// Record the result of a possible conversion here.
+			instanceEdits = append(instanceEdits, internalManifest.ListEdit{
+				ListOperation:            internalManifest.ListOpAdd,
+				AddDigest:                updated.manifestDigest,
+				AddSize:                  int64(len(updated.manifest)),
+				AddMediaType:             updated.manifestMIMEType,
+				AddPlatform:              instance.cloneAddPlatform,
+				AddAnnotations:           instance.cloneAddAnnotations,
+				AddCompressionAlgorithms: instance.cloneAddCompressionAlgorithms,
```
mtrmac, line 245: This can now  be `updated.compressionAlgorithms` — and that `instanceCopy` filed does not need to exist.

mtrmac (commented):
A very brief skim just to allow progress, I’m afraid I didn’t check the other comments so this probably contains some duplicates.

copy/copy.go
```diff
@@ -126,6 +127,16 @@ type Options struct {
 	// Download layer contents with "nondistributable" media types ("foreign" layers) and translate the layer media type
 	// to not indicate "nondistributable".
 	DownloadForeignLayers bool
+
+	// Contains slice of OptionCompressionVariant, where c/image will ensure that selected
+	// images in the manifest list will exist with compression specificed in OptionCompressionVariant
+	// while performing copy.
```
mtrmac, line 133: - “c/image” is not something callers really care about here.
- if “selected images” refers to `ImageListSelection`, that’s probably worth being more explicit
- Please add “Invalid when copying a non-multi-architecture image. That will probably change in the future”. (There’s another review comment outstanding about actually rejecting the option in that case.)

copy/copy.go
mtrmac, line 139: Also document something like `// Only used when we are creating a new image instance using the specified algorithm, not when the image already contains such an instance`

copy/multiple.go
```diff
@@ -27,23 +32,90 @@ const (
 type instanceCopy struct {
 	op           instanceCopyKind
 	sourceDigest digest.Digest
+
+	// Fields which can be used by callers when operation
+	// is `instanceCopyClone`
+	compressionVariant            *compression.Algorithm
+	cloneAddPlatform              *imgspecv1.Platform
+	cloneAddAnnotations           map[string]string
+	cloneAddCompressionAlgorithms []compression.Algorithm
```
mtrmac, line 41: Non-blocking: Maybe `s/cloneAdd/clone/`; “add” is a discriminator in `ListOpAdd`, which we use to implement `…Clone`; so saying both here does not help, and it can be potentially confusing (“add annotations” … on top of which pre-existing ones?).

copy/multiple.go
```diff
@@ -27,23 +32,90 @@ const (
...
+	OSVersion    string
+	OSFeatures   string
+	Variant      string
+}
+
+// Converts imgspecv1.Platform to a comparable format.
+func platformV1ToPlatformComparable(platform imgspecv1.Platform) platformComparable {
+	return platformComparable{Architecture: platform.Architecture,
+		OS:         platform.OS,
+		OSFeatures: strings.Join(platform.OSFeatures, ","),
```
mtrmac, line 57: This is not _quite_ correct, nothing prevents elements of `OSFeatures` from including `,`. Probably good enough in practice, maybe just worth a comment.

---

For more correct matching, `strings.Join(sort.Strings(platform.OSFeatures))` would help.

copy/multiple.go
```diff
@@ -27,23 +32,90 @@ const (
...
+	cloneAddCompressionAlgorithms []compression.Algorithm
+}
+
+// internal type only to make imgspecv1.Platform comparable
+type platformComparable struct {
+	Architecture string
+	OS           string
+	OSVersion    string
+	OSFeatures   string
+	Variant      string
```
mtrmac, line 50: (Non-blocking: These fields can all be package-private.)

copy/multiple.go
```diff
@@ -27,23 +32,90 @@ const (
...
+	return platformComparable{Architecture: platform.Architecture,
+		OS:         platform.OS,
+		OSFeatures: strings.Join(platform.OSFeatures, ","),
+		OSVersion:  platform.OSVersion,
+		Variant:    platform.Variant,
+	}
+}
+
+// platformCompresionPair prepares a mapping of platformComparable -> CompressionAlgorithmNames for given digests
+func platformCompresionPair(list internalManifest.List, instanceDigests []digest.Digest) (map[platformComparable]*set.Set[string], error) {
```
mtrmac, line 64: `s/Pair/Map/` then? (Also two `s` in `Compression`.)

copy/multiple.go
```diff
@@ -27,23 +32,90 @@ const (
...
+	for _, instanceDigest := range instanceDigests {
+		instanceDetails, err := list.Instance(instanceDigest)
+		if err != nil {
+			return nil, fmt.Errorf("getting details for instance %s: %w", instanceDigest, err)
+		}
+		platform := platformV1ToPlatformComparable(*instanceDetails.ReadOnly.Platform)
+		_, valueExists := platformCompressionMap[platform]
+		if !valueExists {
+			platformCompressionMap[platform] = set.New[string]()
+		}
```
mtrmac, line 75: (Absolutely non-blocking: The function is small enough that `_, ok := res[platform]` would be a bit shorter. This works perfectly fine as is.)

copy/multiple.go
```diff
@@ -27,23 +32,90 @@ const (
...
+		set := platformCompressionMap[platform]
```
mtrmac, line 76: Non-blocking:
```
platformSet, ok := …Map[platform]
if !ok {
   platformSet = set.New…
   …Map[platform] = platformSet
}
```
would avoid one map lookup, in both paths (OTOH this is clearly not going to be performance-critical), and it would make it clear that the `set := ` lookup here can never fail and return `nil`.

copy/multiple.go
```diff
@@ -27,23 +32,90 @@ const (
...
 // prepareInstanceCopies prepares a list of instances which needs to copied to the manifest list.
-func prepareInstanceCopies(instanceDigests []digest.Digest, options *Options) []instanceCopy {
+func prepareInstanceCopies(list internalManifest.List, instanceDigests []digest.Digest, options *Options) ([]instanceCopy, error) {
 	res := []instanceCopy{}
+	platformCompressionMap, err := platformCompresionPair(list, instanceDigests)
+	if err != nil {
+		return res, err
+	}
 	for i, instanceDigest := range instanceDigests {
 		if options.ImageListSelection == CopySpecificImages &&
```
mtrmac, line 90: Combining `CopySpecificImages` and the clone map is yet one more thing to worry about.

I think the current approach is correct in that if `ImageListSelection` excludes an image, we should not be cloning it.

But what if `ImageListSelection` refers to a gzip image, we are asked to ensure Zstd exists, and `list` already contains a Zstd variant… but not in `ImageListSelection`?

I think one very reasonable option is to just fail on `ImageListSelection && EnsureCompression…` with a “not currently implemented” error. And we can implement that later if a user shows up. (As a workaround, the user could do two copies, one with each option; with the order depending on what they want to achieve.)

copy/multiple.go
mtrmac: Yes, this is necessary — not just if `Ensure…Exist` contains the same algorithm twice, but also when the original image already contains two instances for the same platform. (Consider some future algo3, an image with gzip+zstd, and a call with `Ensure…Exist = [algo3]`.

---

An alternative implementation approach would be to not implement `EnsureCompressionVariantsExist` here per-instance, but to do it per-platform:
```go
for i, instanceDigest = instanceDigests { /* … issue instanceCopyCopy only */}
for platform, algos := range platformCompressionMap {
   for _, requiredVariant := range Ensure…Exist {
       if !algos.Contains(requiredVariant) {
          /* issue instanceCopyClone */
       } 
   }
}
```
OTOH that would require us to also build some `comparablePlatform -> (instance digest , OCI platform)` data, so that’s probably much more work.

Just something to keep in mind when looking at the `CopySpecificInstances` interaction.

copy/multiple.go
mtrmac: …` copySingleImageOptions` that is

mtrmac (commented):
A more careful review now (I didn’t pay much attention to the tests — I think they should eventually demonstrate many of the more tricky situations).

Some older outstanding comments:
- https://github.com/containers/image/pull/1987#discussion_r1222027394 on some tests
- https://github.com/containers/image/pull/1987#discussion_r1222041486 fail when not using `copyMultipleImages`
- https://github.com/containers/image/pull/1987#discussion_r1262899124 change how we report progress

copy/multiple.go
```diff
@@ -39,8 +42,15 @@ func prepareInstanceCopies(instanceDigests []digest.Digest, options *Options) []
...
+		operation := instanceCopyCopy
+		forceZstd := false
+		if _, ok := options.EnsureCompressionVariantExists["zstd"]; ok {
+			operation = instanceCopyClone
+			forceZstd = true
+		}
 		res = append(res, instanceCopy{
-			op:           instanceCopyCopy,
+			op:           operation,
+			forceZstd:    forceZstd,
```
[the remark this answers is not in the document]
flouthoc: Fixed in commit `[copy/multiple_test: implement test for instanceCopyClone with helpers](https://github.com/containers/image/pull/1987/commits/73a74008c85ea13a1a76f6d2fbb75e25e9ba1858)`

copy/multiple.go
flouthoc: > Maybe add the clone’s algorithm to compressionList so that we don’t create two clones on duplicate input

Did this.

> Just something to keep in mind when looking at the CopySpecificInstances interaction.

For now erroring out if `CopySpecificInstance` is selected and `EnsureCompressionVariantsExist` is set, added a test to verify the same.

copy/multiple.go
flouthoc: Done, erroring out for this and added test to verify the same.

internal/set/set_test.go
flouthoc: Done, added a new function.

flouthoc:
@mtrmac PTAL, Addressed all the comments and modified test cases to cover cases as per new design. Some key changes are implemented in different commits for detail and easier review.

flouthoc:
@mtrmac friendly ping :)

copy/multiple_test.go
```diff
@@ -1,36 +1,163 @@
...
+
+func convertInstanceCopyToSimplerInstanceCopy(copies []instanceCopy) []simplerInstanceCopy {
+	res := []simplerInstanceCopy{}
+	for _, instance := range copies {
+		compression := ""
+		platform := ""
+		if instance.cloneCompressionVariant.Algorithm != nil {
+			compression = instance.cloneCompressionVariant.Algorithm.Name()
+		}
+		if instance.clonePlatform != nil {
```
mtrmac, line 151: I thought `assert.Equal` deals with pointers to values correctly here:
> Pointer variable equality is determined based on the equality of the
> referenced values (as opposed to the memory addresses)

so this might not be necessary. OTOH, this works fine…

copy/multiple_test.go
```diff
@@ -1,36 +1,163 @@
...
+
+	// Following test ensures
+	// * Still copy gzip variants if they exist in the original
+	// * Not create new Zstd variants if they exist in the original.
+
+	// We crated a list of three instances `sourceInstances` and since in oci1.index.zstd-selection.json
+	// amd64 already has a zstd instance i.e sourceInstance[1] so it should not create replication for
+	// `sourceInstance[0]` and `sourceInstance[1]` but should do it for `sourceInstance[2]` for `arm64`
+	// and still copy `sourceInstance[2]`.
+	assert.Equal(t, 4, len(instancesToCopy))
```
mtrmac, line 81: The array `assert.Equal` below implies this, no need to hard-code.

copy/multiple_test.go
```diff
@@ -1,36 +1,163 @@
...
+	expectedResponse := []simplerInstanceCopy{}
+	for _, instance := range sourceInstances {
+		// If its `arm64` and sourceDigest[2] , expect a clone to happen
+		if instance == sourceInstances[2] {
+			expectedResponse = append(expectedResponse, simplerInstanceCopy{op: instanceCopyClone, sourceDigest: instance, cloneCompressionVariant: "zstd", clonePlatform: "arm64-linux-"})
+		}
+		expectedResponse = append(expectedResponse, simplerInstanceCopy{op: instanceCopyCopy,
+			sourceDigest: instance})
+	}
```
mtrmac, line 90: Absolutely non-blocking: I think using a hard-coded array constant with the expected data would not be longer, and it would actually be a tiny bit easier to follow.

(Going further down that road, all of the new added tests could be table-driven. That would scale better.)

copy/multiple_test.go
```diff
@@ -1,36 +1,163 @@
...
+	sourceDigest digest.Digest
+
+	// Fields which can be used by callers when operation
+	// is `instanceCopyClone`
+	cloneCompressionVariant string
+	clonePlatform           string
+	cloneAnnotations        map[string]string
+}
+
+func convertInstanceCopyToSimplerInstanceCopy(copies []instanceCopy) []simplerInstanceCopy {
```
mtrmac, line 143: (For future reference, I think a `assertInstanceCopiesEqual(t *testing.T, a []instanceCopy, b []instanceCopy`) would be less code. But to be clear, now that this code has been written, there’s absolutely no reason to replace it.)

copy/copy.go
```diff
@@ -126,6 +127,17 @@ type Options struct {
 	// Download layer contents with "nondistributable" media types ("foreign" layers) and translate the layer media type
 	// to not indicate "nondistributable".
 	DownloadForeignLayers bool
+
+	// Contains slice of OptionCompressionVariant, where copy will ensure that ImageListSelection
+	// images in the manifest list will exist with compression specificed in OptionCompressionVariant
+	// while performing copy. Invalid when copying a non-multi-architecture image. That will probably
+	// change in the future.
```
mtrmac, line 134: Something like
```suggestion
	// Contains slice of OptionCompressionVariant, where copy will ensure that for each platform
	// in the manifest list, a variant with the requested compression will exist.
	//
	// Invalid when copying a non-multi-architecture image. That will probably
	// change in the future.
```
, we have explicitly decided not to deal with `ImageListSelection`.

copy/multiple.go
```diff
@@ -27,23 +31,115 @@ const (
...
+	variant      string
+}
+
+// Converts imgspecv1.Platform to a comparable format.
+func platformV1ToPlatformComparable(platform imgspecv1.Platform) platformComparable {
+	osFeatures := slices.Clone(platform.OSFeatures)
+	sort.Strings(osFeatures)
+	return platformComparable{architecture: platform.Architecture,
+		os:         platform.OS,
+		osFeatures: strings.Join(osFeatures, ","),
```
mtrmac, line 57:
```suggestion
		// This is strictly speaking ambiguous, fields of OSFeatures can contain a ','. Probably good enough for now.
		osFeatures: strings.Join(osFeatures, ","),
```

copy/copy.go
```diff
@@ -126,6 +127,17 @@ type Options struct {
...
+	EnsureCompressionVariantsExist []OptionCompressionVariant
```
mtrmac, line 135: Another reminder: `copy.Image` should fail on the non-`copyMultipleImages` paths if this is not set.

mtrmac (commented):
Thanks! overall logic looks good; some `nil` references, some missing checks…

copy/copy.go
flouthoc: If i got this right i think you meant `copy.Image should fail on the non-copyMultipleImages paths if this is set`, added this in a new commit.

copy/multiple_test.go
flouthoc: I will clean test infra in a new PR.

copy/multiple.go
```diff
@@ -27,23 +31,115 @@ const (
...
+	for _, option := range input {
+		if option.Algorithm != nil {
+			_, err := compression.AlgorithmByName(option.Algorithm.Name())
+			if err != nil {
+				return fmt.Errorf("invalid algorithm %s in option.EnsureCompressionVariantsExist: %w", option.Algorithm.Name(), err)
+			}
+		}
+	}
+	return nil
 }
```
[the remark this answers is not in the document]
flouthoc: Yes this is only to validate invalid `Algorithm` values,  as requested in comment here: https://github.com/containers/image/pull/1987#discussion_r1222041486

copy/multiple_test.go
flouthoc: Yes, testinfra needs a bit of cleanup. I'll open a follow up PR after this.

flouthoc:
@mtrmac just a small doubt here https://github.com/containers/image/pull/1987#discussion_r1273988194 other than that addressed all the comments.

copy/copy.go
mtrmac: Yes, my mistake.

copy/multiple.go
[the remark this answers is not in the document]
mtrmac: That was necessary when the value was a string and we were checking only for `"zstd"` (and we would silently do nothing if the user asked for `"my-private-new-compression"`).

Now, I think we can… more or less… assume that `Algorithm` values come from `AlgorithmByName`, and if the user asked for an unsupported algorithm, that function would fail.

So that error condition went away by restructuring the code, and I don’t think an explicit check is necessary. But it also doesn’t hurt anything…

copy/copy.go
```diff
@@ -250,13 +262,19 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
 	}
 
 	if !multiImage {
+		if len(options.EnsureCompressionVariantsExist) > 0 {
+			return nil, fmt.Errorf("EnsureCompressionVariantsExist is not implemented for copySingleImage")
```
mtrmac, line 266: Something like
```suggestion
			return nil, fmt.Errorf("EnsureCompressionVariantsExist is not implemented when not creating a multi-architecture image")
```
please, users don’t see the private `copySingleImage` name.

In both places.

mtrmac (commented):
Thanks! One last nit?

copy/copy.go
flouthoc: done :)

flouthoc:
@containers/image-maintainers Could you merge me pls.

copy/copy.go
```diff
@@ -126,6 +127,17 @@ type Options struct {
...
 	DownloadForeignLayers bool
+
+	// Contains slice of OptionCompressionVariant, where copy will ensure that for each platform
+	// in the manifest list, a variant with the requested compression will exist.
+	// Invalid when copying a non-multi-architecture image. That will probably
+	// change in the future.
+	EnsureCompressionVariantsExist []OptionCompressionVariant
+}
+
+type OptionCompressionVariant struct {
```
vrothberg, line 138: Non-blocking.  A comment reflecting the above one would be helpful.

copy/multiple.go
```diff
@@ -82,9 +83,31 @@ func platformCompressionMap(list internalManifest.List, instanceDigests []digest
 	return res, nil
 }
 
+func validateCompressionVariantExists(input []OptionCompressionVariant) error {
+	for _, option := range input {
+		_, err := compression.AlgorithmByName(option.Algorithm.Name())
+		if err != nil {
+			return fmt.Errorf("invalid algorithm %s in option.EnsureCompressionVariantsExist: %w", option.Algorithm.Name(), err)
```
vrothberg, line 90: Non-blocking. Consider using `%q` instead which will make spotting an empty string much easier.

vrothberg (approved):
LGTM

mtrmac:
Thank you so much, again!

PR #1989, opened 2023-06-07
Don't claim that libostree is required by default.

### LINKED ISSUE #1988: Docs unclear about whether ostree is default
The readme currently says
> This library, by default, also depends on the GpgME and libostree C libraries. Either install them:
> ...
> or use the build tags described below to avoid the dependencies (e.g. using go build -tags …)

But the listed ostree tag is `containers_image_ostree` which does the opposite.  I think originally `_stub` was the tag, and this was changed in https://github.com/containers/image/pull/587 but the above text wasn't updated.

mtrmac:
Fixes #1988 . (Well, avoids it.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,9 +45,9 @@ in your project. Ensure that the dependencies documented [in go.mod][go.mod]
 are also available (using those exact versions or different versions of
 your choosing).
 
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat (commented):
LGTM

rhatdan:
LGTM

PR #1992, opened 2023-06-12
list,oci_index: automatically add inbuilt annotations on add

flouthoc:
Adding instance to list should automatically add special annotations if any.

### DIFF
--- a/internal/manifest/list.go
+++ b/internal/manifest/list.go
@@ -3,6 +3,7 @@ package manifest
 import (
 	"fmt"
 
+	compression "github.com/containers/image/v5/pkg/compression/types"
 	"github.com/containers/image/v5/types"
 	digest "github.com/opencontainers/go-digest"
 	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
@@ -82,17 +83,21 @@ type ListEdit struct {
 	ListOperation ListOp
 
 	// if Op == ListEditUpdate (basically the previous UpdateInstances). All fields must be set.
-	UpdateOldDigest digest.Digest
-	UpdateDigest    digest.Digest
-	UpdateSize      int64
-	UpdateMediaType string
+	UpdateOldDigest             digest.Digest
+	UpdateDigest                digest.Digest
+	UpdateSize                  int64
+	UpdateMediaType             string
+	UpdateAffectAnnotations     bool
+	UpdateAnnotations           map[string]string
+	UpdateCompressionAlgorithms []compression.Algorithm
 
 	// If Op = ListEditAdd. All fields must be set.
-	AddDigest      digest.Digest
-	AddSize        int64
-	AddMediaType   string
-	AddPlatform    *imgspecv1.Platform
-	AddAnnotations map[string]string
+	AddDigest                digest.Digest
+	AddSize                  int64
+	AddMediaType             string
+	AddPlatform              *imgspecv1.Platform
+	AddAnnotations           map[string]string
+	AddCompressionAlgorithms []compression.Algorithm
 }
 
 // ListPublicFromBlob parses a list of manifests.
--- a/internal/manifest/oci_index.go
+++ b/internal/manifest/oci_index.go
@@ -7,6 +7,7 @@ import (
 	"runtime"
 
 	platform "github.com/containers/image/v5/internal/pkg/platform"
+	compression "github.com/containers/image/v5/pkg/compression/types"
 	"github.com/containers/image/v5/types"
 	"github.com/opencontainers/go-digest"
 	imgspec "github.com/opencontainers/image-spec/specs-go"
@@ -22,7 +23,8 @@ const (
 	// That also suggests that this instance benefits from
 	// Zstd compression, so it can be preferred by compatible consumers over instances that
 	// use gzip, depending on their local policy.
-	OCI1InstanceAnnotationCompressionZSTD = "io.github.containers.compression.zstd"
+	OCI1InstanceAnnotationCompressionZSTD      = "io.github.containers.compression.zstd"
+	OCI1InstanceAnnotationCompressionZSTDValue = "true"
 )
 
 // OCI1IndexPublic is just an alias for the OCI index type, but one which we can
@@ -76,8 +78,23 @@ func (index *OCI1IndexPublic) UpdateInstances(updates []ListUpdate) error {
 	return index.editInstances(editInstances)
 }
 
+func addCompressionAnnotations(compressionAlgorithms []compression.Algorithm, annotationsMap map[string]string) {
+	// TODO: This should also delete the algorithm if map already contains an algorithm and compressionAlgorithm
+	// list has a different algorithm. To do that, we would need to modify the callers to always provide a reliable
+	// and full compressionAlghorithms list.
+	for _, algo := range compressionAlgorithms {
+		switch algo.Name() {
+		case compression.ZstdAlgorithmName:
+			annotationsMap[OCI1InstanceAnnotationCompressionZSTD] = OCI1InstanceAnnotationCompressionZSTDValue
+		default:
+			continue
+		}
+	}
+}
+
 func (index *OCI1IndexPublic) editInstances(editInstances []ListEdit) error {
 	addedEntries := []imgspecv1.Descriptor{}
+	updatedAnnotations := false
 	for i, editInstance := range editInstances {
 		switch editInstance.ListOperation {
 		case ListOpUpdate:
@@ -102,19 +119,38 @@ func (index *OCI1IndexPublic) editInstances(editInstances []ListEdit) error {
 				return fmt.Errorf("update %d of %d passed to OCI1Index.UpdateInstances had no media type (was %q)", i+1, len(editInstances), index.Manifests[i].MediaType)
 			}
 			index.Manifests[targetIndex].MediaType = editInstance.UpdateMediaType
+			if editInstance.UpdateAnnotations != nil {
+				updatedAnnotations = true
+				if editInstance.UpdateAffectAnnotations {
+					index.Manifests[targetIndex].Annotations = maps.Clone(editInstance.UpdateAnnotations)
+				} else {
+					if index.Manifests[targetIndex].Annotations == nil {
+						index.Manifests[targetIndex].Annotations = map[string]string{}
+					}
+					maps.Copy(index.Manifests[targetIndex].Annotations, editInstance.UpdateAnnotations)
+				}
+			}
+			addCompressionAnnotations(editInstance.UpdateCompressionAlgorithms, index.Manifests[targetIndex].Annotations)
 		case ListOpAdd:
+			annotations := map[string]string{}
+			if editInstance.AddAnnotations != nil {
+				annotations = maps.Clone(editInstance.AddAnnotations)
+			}
+			addCompressionAnnotations(editInstance.AddCompressionAlgorithms, annotations)
 			addedEntries = append(addedEntries, imgspecv1.Descriptor{
 				MediaType:   editInstance.AddMediaType,
 				Size:        editInstance.AddSize,
 				Digest:      editInstance.AddDigest,
 				Platform:    editInstance.AddPlatform,
-				Annotations: editInstance.AddAnnotations})
+				Annotations: annotations})
 		default:
 			return fmt.Errorf("internal error: invalid operation: %d", editInstance.ListOperation)
 		}
 	}
 	if len(addedEntries) != 0 {
 		index.Manifests = append(index.Manifests, addedEntries...)
+	}
+	if len(addedEntries) != 0 || updatedAnnotations {
 		slices.SortStableFunc(index.Manifests, func(a, b imgspecv1.Descriptor) bool {
 			return !instanceIsZstd(a) && instanceIsZstd(b)
 		})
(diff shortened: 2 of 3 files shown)


flouthoc:
@mtrmac PTAL

internal/manifest/oci_index.go
```diff
@@ -103,6 +116,7 @@ func (index *OCI1IndexPublic) editInstances(editInstances []ListEdit) error {
 			}
 			index.Manifests[targetIndex].MediaType = editInstance.UpdateMediaType
 		case ListOpAdd:
+			addCompressionAnnotations(editInstance.AddCompressionAnnotation, editInstance.AddAnnotations)
```
mtrmac, line 119: Maps are reference types in Go (as shown in the `addCompressionAnnotations` parameter design); doesn’t this modify caller’s data?

Consider
```go
annotations := map[…]…
edits := []ListEdit{{…, annotations}, {…, annotations}}
```

AFAICS this would mean that if the first image adds an annotation, the second one would get it as well.

internal/manifest/list.go
```diff
@@ -88,11 +89,12 @@ type ListEdit struct {
...
-	AddSize        int64
-	AddMediaType   string
-	AddPlatform    *imgspecv1.Platform
-	AddAnnotations map[string]string
+	AddDigest                digest.Digest
+	AddSize                  int64
+	AddMediaType             string
+	AddPlatform              *imgspecv1.Platform
+	AddAnnotations           map[string]string
+	AddCompressionAnnotation []compression.Algorithm
```
mtrmac, line 97: This design LGTM. We need this also for `Update`.

mtrmac (commented):
ACK.

Another part (that can be this PR or a separate one) is to ensure that `skopeo copy --format zstd`, and a plain `skopeo copy` that uses zstd (e.g. because it reuses an existing blob) also adds that annotation.

internal/manifest/list.go
mtrmac, line 97: `AddCompressionAlgorithms[Used]`? That this is stored in an annotation is conceptually manifest-format-specific (some future format might have this as a native field).

internal/manifest/oci_index.go
```diff
@@ -102,13 +115,26 @@ func (index *OCI1IndexPublic) editInstances(editInstances []ListEdit) error {
 				return fmt.Errorf("update %d of %d passed to OCI1Index.UpdateInstances had no media type (was %q)", i+1, len(editInstances), index.Manifests[i].MediaType)
 			}
 			index.Manifests[targetIndex].MediaType = editInstance.UpdateMediaType
+			for k, v := range editInstance.UpdateAnnotations {
+				index.Manifests[targetIndex].Annotations[k] = v
+			}
```
mtrmac, line 120: `x/exp/maps.Clone` please, then we don’t have to even think about correctness of the implementation.

---

The above would, AFAICS, preserve any old annotations in the descriptor, i.e. it would not allow removing an annotation. Using `Clone` + overwriting `.Annotations` is clearly an overwrite.

internal/manifest/oci_index.go
```diff
@@ -102,13 +115,26 @@ func (index *OCI1IndexPublic) editInstances(editInstances []ListEdit) error {
...
+			annotations := make(map[string]string)
+			addCompressionAnnotations(editInstance.UpdateCompressionAlgorithms, annotations)
+			for k, v := range annotations {
+				index.Manifests[targetIndex].Annotations[k] = v
+			}
```
mtrmac, line 125: `addCompressionAnnotations(…, index.Manifests[…].Annotations)` would AFAICS do the same thing in a simpler way; no need for an extra zero-or-one-element map.

If you think the effect is not clear, `addCompressionAnnotations(*Descriptor, []Alghoritm)` would make it very clear that it is editing the descriptor.

internal/manifest/oci_index.go
```diff
@@ -102,13 +115,26 @@ func (index *OCI1IndexPublic) editInstances(editInstances []ListEdit) error {
...
 		case ListOpAdd:
+			annotations := make(map[string]string)
+			for k, v := range editInstance.AddAnnotations {
+				annotations[k] = v
+			}
```
mtrmac, line 130: `maps.Clone`

internal/manifest/oci_index.go
```diff
@@ -76,6 +78,17 @@ func (index *OCI1IndexPublic) UpdateInstances(updates []ListUpdate) error {
 	return index.editInstances(editInstances)
 }
 
+func addCompressionAnnotations(compressionAlgorithm []compression.Algorithm, annotationsMap map[string]string) {
```
mtrmac, line 81: Hum, should this also remove the annotation if the image is not using zstd? E.g. copying a zstd image while converting to gzip.

(That can only work if call callers correctly fill in the list of algorithms…)

mtrmac: “if all callers” …

flouthoc: Yes , one hurdle which i saw what defines if user wants to recompress as `gzip` is it leaving the `compressionAlgorithm` list empty or adding a new element `gzip`, Callers would need to take care of this.

I left a `TODO` comment instead of implementing it now.

internal/manifest/list.go
```diff
@@ -82,17 +83,20 @@ type ListEdit struct {
...
 	// if Op == ListEditUpdate (basically the previous UpdateInstances). All fields must be set.
-	UpdateOldDigest digest.Digest
-	UpdateDigest    digest.Digest
-	UpdateSize      int64
-	UpdateMediaType string
+	UpdateOldDigest             digest.Digest
+	UpdateDigest                digest.Digest
+	UpdateSize                  int64
+	UpdateMediaType             string
+	UpdateAnnotations           map[string]string
```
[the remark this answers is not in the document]
mtrmac: Also, `UpdateInstances` uses this code path, and does not fill `UpdateAnnotations`. So we need to distinguish between “don’t edit annotations” and “delete all existing annotations, replace with no annotations”.

internal/manifest/oci_index.go
```diff
@@ -76,6 +78,19 @@ func (index *OCI1IndexPublic) UpdateInstances(updates []ListUpdate) error {
...
+	// TODO: This should also delete the algorithm if map already contains an algorithm and compressionAlgorithm
+	// list has a different algorithm.
```
mtrmac, line 83: I don’t know how I feel about not implementing this… I guess let’s start with the “add a Zstd annotation on a Zstd copy” part identified earlier, and let’s see how that goes.

---

Here, though, please at least record. 
```suggestion
	// TODO: This should also delete the algorithm if map already contains an algorithm and compressionAlgorithm
	// list has a different algorithm. To do that, we would need to modify the callers to always provide a reliable
	// and full compressionAlghorithms list.
```

internal/manifest/oci_index.go
```diff
@@ -76,6 +78,19 @@ func (index *OCI1IndexPublic) UpdateInstances(updates []ListUpdate) error {
 	return index.editInstances(editInstances)
 }
 
+func addCompressionAnnotations(compressionAlgorithm []compression.Algorithm, annotationsMap map[string]string) {
```
mtrmac, line 81:
```suggestion
func addCompressionAnnotations(compressionAlgorithms []compression.Algorithm, annotationsMap map[string]string) {
```
or just
```suggestion
func addCompressionAnnotations(algorithms []compression.Algorithm, annotationsMap map[string]string) {
```

internal/manifest/list.go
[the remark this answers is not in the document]
flouthoc: I added a new field `OverwriteAnnotations bool` that should cover this case I think.

internal/manifest/oci_index.go
flouthoc: I implemented `delete` for this, assuming callers would take care of choosing `gzip` in the algorithms list.

rhatdan:
@vrothberg @mtrmac PTAL

internal/manifest/oci_index.go
```diff
@@ -76,6 +78,26 @@ func (index *OCI1IndexPublic) UpdateInstances(updates []ListUpdate) error {
...
+		case compression.GzipAlgorithmName:
+			foundGzip = true
+		default:
+			continue
+		}
+	}
+
+	// In case if user is reverting to default gzip
+	// we must remove other inbuilt annotatations
+	if foundGzip {
```
mtrmac, line 96: I can’t see how this follows. AFAICS the zstd annotation should be used iff a zstd layer is present, and presence or absence of gzip layers has nothing to do with it.

(Consider a future in which there is some other third format.)

flouthoc: @mtrmac Yes I had some confusion about how to implement deletion since there is now way to determine if user wants to remove `zstd` annotation. Any hints about how to add deletion here ?

mtrmac: The idea is that the caller provides a _complete and precise_ list of compression algorithms. (That doesn’t currently exist.)

Then this code can always add or remove the zstd annotation automatically. The caller is a generic OCI-unaware thing that doesn’t know a zstd annotation exists; the caller only provides the requires list of algorithms.

---

… and that doesn’t work for the existing `UpdateInstances`. Oops.

---

A point to note is that this only works if the list of algorithms is specified. I.e., right now, this PR would cause the delete code to keep deleting the annotation.

So, either split this further (per https://github.com/containers/image/pull/1992#discussion_r1228373189 ):
- revert this PR to only add an annotation
- implement correctly setting `*CompressionAlgorithms` in the caller
- and then re-add the part of this PR that deletes the zstd annotation.

Or, do all of the three in _this_ PR, i.e. extend it to correctly set `*CompressionAlghorithms` in c/image/copy.

internal/manifest/list.go
```diff
@@ -82,17 +83,21 @@ type ListEdit struct {
...
 	// if Op == ListEditUpdate (basically the previous UpdateInstances). All fields must be set.
-	UpdateOldDigest digest.Digest
-	UpdateDigest    digest.Digest
-	UpdateSize      int64
-	UpdateMediaType string
+	UpdateOldDigest             digest.Digest
+	UpdateDigest                digest.Digest
+	UpdateSize                  int64
+	UpdateMediaType             string
+	OverwriteAnnotations        bool
```
mtrmac, line 90: - Name this `Update…` to make it lear it relates to `ListEditUpdate`, please. [1]
- Per the other thread, the `UpdateInstances` caller will not set `UpdateCompressionAlgorithms`, so we must have an opt-in into both kinds of the annotation changes. Maybe `UpdateAffectAnnotations`? `UpdateDoNotChangeAnnotations`? (with opposite meanings).

---

[1] Or maybe
```go
type ListEditStruct {
  ListOperation …

  // if Op == ListEditUpdate …
  Update struct {
    OldDigest …
    Digest …
  }
```
I haven’t checked that is actually possible. It would mean a tiny bit less repetition here, at the cost of extra `.` in all users; I don’t have a preference.

internal/manifest/oci_index.go
flouthoc: > revert this PR to only add an annotation

SGTM I think doing this later on will be easier once we have some working.

flouthoc:
@mtrmac PTAL

flouthoc:
@mtrmac Is Anything left pending in this PR ? I have reverted some of the changes, so we can revisit them later once we have minimal working model in place. WDYT ?

internal/manifest/oci_index.go
```diff
@@ -102,13 +118,23 @@ func (index *OCI1IndexPublic) editInstances(editInstances []ListEdit) error {
 				return fmt.Errorf("update %d of %d passed to OCI1Index.UpdateInstances had no media type (was %q)", i+1, len(editInstances), index.Manifests[i].MediaType)
 			}
 			index.Manifests[targetIndex].MediaType = editInstance.UpdateMediaType
+			if editInstance.UpdateAffectAnnotations {
+				index.Manifests[targetIndex].Annotations = maps.Clone(editInstance.UpdateAnnotations)
+			} else {
+				for k, v := range editInstance.UpdateAnnotations {
+					index.Manifests[targetIndex].Annotations[k] = v
+				}
```
mtrmac, line 126: `maps.Copy(index.Manifests[targetIndex].Annotations, editInstance.UpdateAnnotations)`

mtrmac (commented):
LGTM overall.

Some unit tests would be nice but not blocking.

internal/manifest/oci_index.go
flouthoc: While working with tests i found that `Copy` panics if  `dest` is `nil`. So we must set `nil` to `map[string]string{}` before doing copy.

internal/manifest/oci_index.go
```diff
@@ -102,19 +119,38 @@ func (index *OCI1IndexPublic) editInstances(editInstances []ListEdit) error {
...
-				Annotations: editInstance.AddAnnotations})
+				Annotations: annotations})
 		default:
 			return fmt.Errorf("internal error: invalid operation: %d", editInstance.ListOperation)
 		}
 	}
 	if len(addedEntries) != 0 {
 		index.Manifests = append(index.Manifests, addedEntries...)
+	}
+	if len(addedEntries) != 0 || updatedAnnotations {
```
mtrmac, line 153: Great catch!

---

Note for the record: Arguably we should re-sort only on _compression_ annotation change.

But then, we are not computing compression annotations precisely right now (only adding, not deleting), so… we can’t exactly guarantee specific behavior. The two boundary cases (`UpdateInstances` with no annotation input, and `EditInstances` adding Zstd) both work correctly.

mtrmac (approved):
LGTM. Thanks!

PR #1998, opened 2023-06-14
Don't maintain URLs for foreign layers when pushing them

drcapulet:
For images with non-distributable layers, if we're pushing the layer to the destination we want to remove any URLs on the layer since they're generally preferred over the repository itself ([containerd](https://github.com/containerd/containerd/blob/main/remotes/docker/fetcher.go#L54-L55) example) - this is important for deployments to air-gapped installations that don't have access to any of the URLs.

### DIFF
--- a/copy/single.go
+++ b/copy/single.go
@@ -406,6 +406,11 @@ func (ic *imageCopier) copyLayers(ctx context.Context) error {
 			}
 		} else {
 			cld.destInfo, cld.diffID, cld.err = ic.copyLayer(ctx, srcLayer, toEncrypt, pool, index, srcRef, manifestLayerInfos[index].EmptyLayer)
+
+			// It's already unset on cld.destInfo, this forces the manifest to be regenerated.
+			if len(srcLayer.URLs) != 0 {
+				srcInfosUpdated = true
+			}
 		}
 		data[index] = cld
 	}


mtrmac (commented):
Thanks for the PR.

AFAICS c/image _is_, currently, falling back to using the registry if the external URL is unavailable. Though it does look that that’s not the case for containerd, and it seems to be unintentional for c/image.

It’s not clear to me that this is a behavior we can just change; it might need a separate option.


---

Note to self: The documentation for `DownloadForeignLayers` says that MIME types are updated, and I can’t see how that happens (nor, actually, whether it is desirable).

drcapulet:
@mtrmac Appreciate the review - funny enough the day I opened this Microsoft [dropped the usage of foreign layers](https://techcommunity.microsoft.com/t5/containers/announcing-removal-of-foreign-layers-from-windows-container/ba-p/3846833), I'm inclined to just close this out?

I can't personally tell on the expected behavior for URLs from the spec.

mtrmac:
I do think you’ve touched on a feature gap and/or a bug, and it’s worth fixing _in principle_.

OTOH the real-world need for the foreign layer feature is really becoming much less urgent. If you want to close this PR, I think that’s completely fair.

jankaluza:

PR #2001, opened 2023-06-15
Don't store signatures if there is none of them

mike-sul:
Currently, the copy command prints the message 'Storing signatures' and calls the signature storing function, even if there are no signatures present. This can mislead users and make them believe that there are image signatures.

The proposed change modifies the copy function to print the message and invoke the image storing function only if there is at least one signature.

### DIFF
--- a/copy/single.go
+++ b/copy/single.go
@@ -256,9 +256,11 @@ func (c *copier) copySingleImage(ctx context.Context, policyContext *signature.P
 	}
 	sigs = append(sigs, newSigs...)
 
-	c.Printf("Storing signatures\n")
-	if err := c.dest.PutSignaturesWithFormat(ctx, sigs, targetInstance); err != nil {
-		return nil, "", "", fmt.Errorf("writing signatures: %w", err)
+	if len(sigs) > 0 {
+		c.Printf("Storing signatures\n")
+		if err := c.dest.PutSignaturesWithFormat(ctx, sigs, targetInstance); err != nil {
+			return nil, "", "", fmt.Errorf("writing signatures: %w", err)
+		}
 	}
 
 	return manifestBytes, retManifestType, retManifestDigest, nil


copy/single.go
```diff
@@ -254,11 +254,15 @@ func (c *copier) copySingleImage(ctx context.Context, policyContext *signature.P
 	if err != nil {
 		return nil, "", "", err
 	}
-	sigs = append(sigs, newSigs...)
+	if newSigs != nil {
```
mtrmac, line 257: This seems unnecessary.

mtrmac (commented):
Thanks.

If we are improving this, the major user-visible aspect is actually that (typically during a pull) it may be the `Commit` that takes a long time, and that’s currently being attributed to storing signatures. That’s certainly not ideal; but just dropping that and attributing the delay to the previous layer copies would not be clearly better.

So I think the `Commit` step should cause a progress log (potentially with a per-transport opt-in to avoid noise?), and we can revisit the signature step afterwards.

copy/single.go
mike-sul: Indeed.

mike-sul:
> That’s certainly not ideal; but just dropping that and attributing the delay to the previous layer copies would not be clearly better.

The previous step is not layer copy iiuc, rather `copyUpdatedConfigAndManifest` which prints `Writing manifest to image destination`. What I observe now is 

```
Copying blob 8f7d2488f433 done  
Copying config 99ae753c80 done  
//--> "Storing signatures" used to be here
```

Do you mean that `Writing manifest to image destination` is misleading message too?

mtrmac:
My mistake; it’s the manifest step indeed.

That’s certainly less misleading than a layer copy — a bit: “why does writing a 1KB file take a minute?”.

mtrmac (approved):
LGTM.

mtrmac:
Either way, this is an improvement.

PR #2003, opened 2023-06-16
Oci image deletion

Pvlerick:
Fix attempt for #1812

I made two commit; first a failing test then the feature to make it work, I'll happily fixup in one commit given how small this change is.


oci/layout/oci_transport.go
```diff
@@ -228,7 +228,16 @@ func (ref ociReference) NewImageDestination(ctx context.Context, sys *types.Syst
 
 // DeleteImage deletes the named image from the registry, if supported.
 func (ref ociReference) DeleteImage(ctx context.Context, sys *types.SystemContext) error {
-	return errors.New("Deleting images not implemented for oci: images")
+	index := filepath.Join(ref.dir, "index.json")
```
mtrmac, line 231: Please use existing utilities like `indexPath()`, so that references to each file all go through a single code path and can be trivially located.

mtrmac (changes requested):
Thanks for the PR.

At a first glance, this seems way too small to plausibly be a comprehensive implementation.
- Most importantly, a single `oci:` directory can house multiple images. So just removing `index.json` is deleting much more than requested.
- Every image contains at least one blob; this is not deleting any blobs at all, AFAICS.
  - Combined with the above, _only_ blobs that are not used by other images can be deleted.
- Typically the second `os.Remove` would just fail, because the directory is not empty. So AFAICS the proposed `DeleteImage` implementation must _always_ fail in practice.
 
  (Yes, it does not fail in the fake `refToTempOCI` repo used in tests, but that mostly exists to exercise the name lookup and metadata operations, so it is not representative.)

mtrmac:
> I made two commit; first a failing test then the feature to make it work, I'll happily fixup in one commit given how small this change is.

We prefer the tests to be working on every commit, so that bisection doesn’t have false positives.

Pvlerick:
> > I made two commit; first a failing test then the feature to make it work, I'll happily fixup in one commit given how small this change is.
> 
> We prefer the tests to be working on every commit, so that bisection doesn’t have false positives.

All right, I'm doing this out of habit (failing tests first then committing) but I'll squash all of that for next time.

Pvlerick:
Thanks for the feedback, I'll get back to this and will ping you once I have something more substantial.

mtrmac:
> All right, I'm doing this out of habit (failing tests first then committing) but I'll squash all of that for next time.

That habit makes sense — we really only need it squashed (or just reordered?) immediately before merging.

Pvlerick:
Here's something hopefully better. Test setup is ugly but I'd like to get the logic validated before refactoring this into something cleaner.

A couple of questions on this:

- if it's the last image in the index, should it delete the index file altogether? And all sub-folders, which should all be empty at that point?
- I don't think it's possible to check if there is a container running using that image, is it? That should be a test case, but I don't see how it would be possible... Multiple OCI runtime could be running in different locations, etc... Hairy.
- if the given image is empty, should it 1) if there is only one image, delete it from the index (and see point above) 2) if more than one image, do nothing - see `getManifestDescriptor` behavior: https://github.com/containers/image/blob/main/oci/layout/oci_transport.go#L183
- what's in the `archive`folder?

oci/layout/oci_transport.go
```diff
@@ -228,7 +247,55 @@ func (ref ociReference) NewImageDestination(ctx context.Context, sys *types.Syst
...
+	}
+
+	manifest, err := ref.getManifest(descriptor)
+	if err != nil {
+		return err
+	}
+
+	// Delete all blobs
+	for _, d := range append(manifest.Layers, manifest.Config, descriptor) {
+		blobPath, err := ref.blobPath(d.Digest, "")
```
mtrmac, line 263: Blobs can be shared across images; deleting one image should not break the others.

oci/layout/oci_transport.go
```diff
@@ -228,7 +247,55 @@ func (ref ociReference) NewImageDestination(ctx context.Context, sys *types.Syst
...
+
+	// Update the index
+	index, err := ref.getIndex()
+	if err != nil {
+		return err
+	}
+
+	newManifests := make([]imgspecv1.Descriptor, len(index.Manifests)-1)
+	for i, v := range index.Manifests {
+		if v.Annotations[imgspecv1.AnnotationRefName] != ref.image {
```
mtrmac, line 281: `ref.image` is not always set (and there is some WIP work to also allow `ociReference` to refer to an image directly using an index.

I guess `ref.getManifest` could also return an index.

mtrmac (commented):
Thanks, this is a move in the right direction.

> if it's the last image in the index, should it delete the index file altogether? And all sub-folders, which should all be empty at that point?

I think keeping the index and other structures around makes sense — if someone cares about the _state_ of the directory to delete an individual image and not to delete everything, that user might plausibly want to delete all existing images and replace them with other images.

> I don't think it's possible to check if there is a container running using that image, is it?

This code does not really have a concept of containers or any kind of lock / external reference counting; it’s a pure store, and it’s up to the caller not to delete things that will need to be used.

> if the given image is empty

Assuming this refers to `ref.image`, I would expect this to follow the existing `getManifestDescriptor` behavior: If the users’ input if ambiguous, fail and don’t delete anything. If `ref.image == ""` in a situation where that is documented to clearly identify one image, deleting that one image makes sense to me (though the situation _is_ a bit suspect, why would anyone do that?).

oci/layout/oci_transport.go
Pvlerick: Yes indeed, I'm working on the test setup for this case.

oci/layout/oci_transport.go
Pvlerick: You mean that ref.image can be "", as described in here: https://github.com/containers/image/blob/main/oci/layout/oci_transport.go#L64 ?

When you say index, you mean that someone could ask deletion of the Nth image in the index.json instead of giving the name? (in that case, that would mean that ref.image being "" is the same as index being 0 and only one image)? I don't understand an index to what `getManifest` would return, unless you meant `getManifestDescriptor` which would return the index at which it found the image in the `index.json`?

Pvlerick:
Style/opinion question: index files are generated using string literals: https://github.com/containers/image/blob/main/oci/layout/oci_transport_test.go#L201

I started refactoring this part for my tests to use imgspecv1 structs that I'm filling then serializing instead; is that ok or do you rather keep literals?

oci/layout/oci_transport.go
mtrmac: > You mean that ref.image can be ""

Yes.

> When you say index, you mean that someone could ask deletion of the Nth image in the index.json instead of giving the name? 

Exactly. That feature already exists for `docker-archive:`, and there are WIP PRs around implementing the same thing for the two OCI transports.

> I don't understand an index to what `getManifest` would return, unless you meant `getManifestDescriptor`

Yes. My mistake.

mtrmac:
> I started refactoring this part for my tests to use imgspecv1 structs that I'm filling then serializing instead

I think using raw blobs works better as an interoperability test; that way we know if the (de)serialization code changes in a way that breaks compatibility.

(That said, using raw files in `testdata` would be cleaner than long string literals like that.)

---

Refactors are fine and welcome, but please keep any larger ones as separate commits to be clear which parts of the PR are intended to do something new and which are intended not to be a change.

oci/layout/oci_transport.go
Pvlerick: All right, makes sense. I'll leave that for now not to clutter the issue worked, but I'm happy to add this later on when you feel the code is good enough.

Pvlerick:
> I think using raw blobs works better as an interoperability test; that way we know if the (de)serialization code changes in a way that breaks compatibility.
> 
> (That said, using raw files in `testdata` would be cleaner than long string literals like that.)

Now I'm on the fence; it's true that literals have the advantage of indirectly testing the spec, which is always a plus. I'll commit what I have now so you can have a look then I'll make the changes, that's not a big deal.

Pvlerick:
I'm fairly happy with what's in there now; test data generation is much clearer (although it uses the spec library, I'll change that to literals as soon as time permits) and allows more flexibility as demonstrated with the test with a layer shared by two images.

(Looks like I'll need to rebase on main too)

Pvlerick:
All right, all test are passing now :-)

Pvlerick:
A couple of question on top of the review:

- Would it make sense to add some logging (debug or trace) such as "deleting blob xxx" ?
- Would it make sense the parallelize deletion of the blobs?

Thanks

mtrmac:
I’m afraid I didn’t get to review the full PR yet

> * Would it make sense to add some logging (debug or trace) such as "deleting blob xxx" ?

Sure; that seems useful, and can’t hurt anything.

> * Would it make sense the parallelize deletion of the blobs?

You know why you are doing this work and whether it is valuable for you. My intuition is that most archives are pretty small, and that the extra complexity and risk are not at all worth it, but that’s a guess with no data.

oci/layout/oci_transport.go
```diff
@@ -228,7 +247,75 @@ func (ref ociReference) NewImageDestination(ctx context.Context, sys *types.Syst
...
+	if err != nil {
+		return err
+	}
+
+	// Get all the layers used by all other images
+	index, err := ref.getIndex()
+	if err != nil {
+		return err
+	}
+	layersUsedByOtherImages := make(map[digest.Digest]bool, len(index.Manifests))
```
mtrmac, line 267: I think this could be `internal/set.Set[digest.Digest]`

oci/layout/oci_transport.go
```diff
@@ -228,7 +247,75 @@ func (ref ociReference) NewImageDestination(ctx context.Context, sys *types.Syst
...
+	for _, v := range index.Manifests {
+		if v.Digest != manifestDescriptor.Digest {
+			otherImageManifest, err := ref.getManifest(v)
+			if err != nil {
+				return err
+			}
+			for _, layer := range otherImageManifest.Layers {
```
mtrmac, line 274: - The manifests themselves are also stored as blobs.
- `Config.Digest` also needs to be accounted for.
- IIRC it’s possible for an entry in the top-level index to be an `imgspecv1.Index` again — at least this code does that to one level of nesting, the spec seems to suggest an arbitrary depth.

With OCI artifacts storing ~arbitrary data, it seems quite plausible for an entry in `Layers` refer to the same blob as a config/manifest/index of another image/artifact.

oci/layout/oci_transport.go
```diff
@@ -228,7 +247,75 @@ func (ref ociReference) NewImageDestination(ctx context.Context, sys *types.Syst
...
+	}
+
+	// Delete all blobs
+	blobsToDelete := make([]digest.Digest, 0, len(manifest.Layers))
+	for _, layer := range manifest.Layers {
+		if !layersUsedByOtherImages[layer.Digest] {
+			blobsToDelete = append(blobsToDelete, layer.Digest)
+		}
+	}
+	for _, digest := range append(blobsToDelete, manifest.Config.Digest, manifestDescriptor.Digest) {
```
mtrmac, line 287: I think the config an manifest entries also need to be checked against `…UsedByOtherImages`.

mtrmac (commented):
(Not a full review, just a very quick skim.)

oci/layout/oci_transport.go
Pvlerick: Didn't see this, will do.

oci/layout/oci_transport.go
Pvlerick: Spec says implementation should support this: https://github.com/opencontainers/image-spec/blob/main/image-index.md?plain=1#L44

It also means that the WIP you mentioned earlier about returning an index when calling `getManifestDescriptor` which would be impossible if the image is in a nested index, or that'll be the index in the index it was found, not necessarily `index.json`

My head starts to hurt :-)

mtrmac: (My personal opinion is that the nested indexes provide too much flexibility and the semantics are unclear for consumers, so I’m not that much of a fan. But the feature exists.)

There’s a bit of nuance in here: if the repo contains a nested index, it’s not ideal, but also not a _disaster_ when this code does not support that in `ImageSource` / `getManifestDescriptor`; the code would just “safely” fail and users would have to use some other tool.

OTOH if the deletion code does not handle nested indices, that could lead to data loss, so that is a more urgent concern.

So one approach would be to recursively enumerate the used blobs; another, I think quite reasonable, possibility would be to just fail when a nested index is encountered, and to tell the user to use another tool — that’s analogous to the `ImageSource` failure mode, but still safe.

---

Note that the above applies to “deeply nested” OCI indexes. One level of nesting is supported, to represent a multi-arch image. See the outcome of

```sh
skopeo --insecure-policy copy --all docker://quay.io/libpod/alpine oci:$dest
```

So… I haven’t actually written the code, but it seems to me that a full recursion (or possibly a “queue of blobs+blob types to scan”) support might be the easiest way to implement all of this anyway, instead of somehow special-casing the “one level of nesting” case.

Pvlerick: You are right that handling this properly in the deletion has to be extra safe.
The good news is that I now have a way to cleanly create nested indexes for tests so this can be tested properly (which uses recursion so in for a penny, in for a pound).

Thanks for the skopeo example, I'll have a look at how this is handled, that will surely be instructive.

oci/layout/oci_transport.go
Pvlerick: I'll add that too; that's quite cheap and provides an extra level of safety.

oci/layout/oci_transport.go
Pvlerick: I've been inspecting the result from the `skopeo` command and it has been very educating :-)

One thing I noticed, reading `getManifestDescriptor`, is that you can toast your registry if you do something like this:
```
skopeo copy docker://quay.io/libpod/alpine:3.2 oci:/tmp/oci-registry/alpine
skopeo copy docker://quay.io/libpod/alpine:3.10.2 oci:/tmp/oci-registry/alpine
```

The index.json file will then contain two manifests (image manifests or sub indexes if you used the --all) which don't have any annotations (so no `org.opencontainers.image.ref.name`).

At that point, that directory is toast it seems:
```
skopeo copy oci:/tmp/oci-registry/alpine:3.2 oci:/tmp/oci-registry/alpine-copy 
FATA[0000] initializing source oci:/tmp/oci-registry/alpine:3.2: no descriptor found for reference "3.2" 
skopeo copy oci:/tmp/oci-registry/alpine oci:/tmp/oci-registry/alpine-copy 
FATA[0000] initializing source oci:/tmp/oci-registry/alpine:: more than one image in oci, choose an image
```

Would this be worthy of being reported as a bug? 
Sorry for the digression, I'm still in the learning phase :-)

mtrmac: > At that point, that directory is toast, it seems:

It’s… a valid OCI structure. And the WIP `oci:/tmp/oci-registry/alpine:@0` syntax will eventually allow c/image to consume it.

I agree it’s very awkward that c/image can produce a structure that it can’t read.

Pvlerick: Can you point me to that WIP with the index? It could be useful, for me, to have a  look.

Pvlerick:
@mtrmac I made quite a few changes:

- used fixtures for the tests, it's a bit more static and less readable (you have to go read the fixture to find out what's the expected result), but it tests the spec package indirectly as we discussed previously;
- I addressed both of your comments: used a set and checked for configuration shared use (and added a test case for it in the shared blobs test

I started working on the nested index stuff in another branch, but it's not over yet (although the fixture is there already); there is also the extra complexity that if something changes in the nested index, its sha256 will change too, so all the "root" index will have to be changed too.

Pvlerick:
I think it's in a good shape now; there are probably refactorings that could be made here and there but I'd prefer you to review and validate that this is going in the right direction.

Pvlerick:
@mtrmac just pinging in case you haven't seen the request for review (if you're too busy right now it can wait)

mtrmac:
@Pvlerick I’m sorry — I do receive your updates and it’s something I want to see through soon, but there are some other priorities for this week. I’ll hopefully return to this next week.

oci/layout/oci_transport.go
mtrmac: That’s #1381 / #1677 . Out of date for a while, I’m afraid.

mtrmac (commented):
Finally looking at this, I’m sorry for the delay.

This is an impressive amount of work, and the tests are very welcome.

I do think we need to support nested indexes when looking for blobs that must not be deleted; but supporting deleting a deeply-nested index adds so much extra complexity…

I think not supporting deleting of a deeply-nested index would be rather simpler:
```go
func addBlobsReferencedByDescriptor(digests *set.Set[digest.Digest], descriptor) {
    digests.Add(descriptor.Digest)
    switch descriptor.MediaType {
    case …Manifest:
        // add config, layers to digests
    case …Index:
        // read it; call addBlobsReferencedByDescriptor() for each item
}

func (…) DeleteImage(…) … {
   toDeleteDescriptor, toDeleteIndex := ref.getManifestDescriptor() // modified to return that new value

   blobsToPotentiallyDelete := set.Set[digest.Digest].New()
   addBlobsReferencedByDescriptor(blobsToPotentiallyDelete, toDeleteDescriptor)

   index := ref.GetIndex()
   blobsToPreserve := set.Set[digest.Digest].New()
   for i, desc := range index.Manifests {
       if i != toDeleteIndex {
           addBlobsReferencedByDescriptor(blobsToPreserve, desc)
   }

   // delete blobs from the difference of the two sets, remove index[toDeleteIndex]
}
```

---

Also please move the added code to new files, maybe `oci/layout/delete*.go`. It’s not really written down anywhere I’m afraid, but the `*_transport.go` files primary deal with reference parsing and identity, which is critical for signatures and related policies, so they have a higher bar for testing. The added code would probably pass that higher bar with flying colors, but longer-term we’d prefer the security-critical code to be separate from the code that can “only” cause data loss :)

oci/layout/oci_transport.go
```diff
@@ -226,9 +299,169 @@ func (ref ociReference) NewImageDestination(ctx context.Context, sys *types.Syst
...
+
+	if ref.image == "" {
+		if len(allDescriptors) == 1 {
+			imageDescriptorWrapper = allDescriptors[0]
+		} else {
+			return ErrMoreThanOneImage
+		}
+	} else {
+		for _, v := range allDescriptors {
+			if v.descriptor.Annotations[imgspecv1.AnnotationRefName] == ref.image {
```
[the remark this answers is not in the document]
mtrmac: >  potentially surprising 

Consider
```sh
target=oci:$path:$tag
skopeo inspect $target
# looks good, yes, that’s the one I want to delete
skopeo delete $target
# deletes a different image
```

Pvlerick:
@mtrmac thanks for the extensive review; I hope the code was clear enough in its intent. 

Reading the various comments, what I gather is that you think image deletion should be supported only when there is only a top level index and not when there is a nested index? Or just one level on nested index but no more? Definitely not the theoretical infinite nested indexes.

It may take a little while for me to find time to properly work on this in the coming month, this being the holiday season, but I'm on it.

oci/layout/oci_transport_test.go
```diff
@@ -319,12 +323,181 @@ func TestReferenceNewImageDestination(t *testing.T) {
...
+		case "3.10.2":
+			otherImagesStillPresent = append(otherImagesStillPresent, true)
+		case "latest":
+			otherImagesStillPresent = append(otherImagesStillPresent, true)
+		}
+	}
+	require.Equal(t, 2, len(otherImagesStillPresent))
+	for _, v := range otherImagesStillPresent {
+		require.True(t, v)
+	}
```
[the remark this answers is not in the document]
Pvlerick: Nice one, didn't think of it (I'm new at Go)

mtrmac:
> Or just one level on nested index but no more?

The transport can read and write two levels nested indices (the top-level `index.json` may refer to one level of other indices). The _name references_ are found only in the top-level `index.json`, and can refer either to a single manifest or to an index (which contains manifests, not indices).

oci/layout/oci_transport.go
```diff
@@ -204,6 +216,67 @@ func (ref ociReference) getManifestDescriptor() (imgspecv1.Descriptor, error) {
...
+		return descriptors, nil
+	}
+
+	index := ref.indexPath()
+	indexChain := []*string{&index} // We start the walk at the root: index.json
+	return getImageDescriptorsFromIndex(indexChain)
+}
+
+func (ref ociReference) getManifest(descriptor *imgspecv1.Descriptor) (*imgspecv1.Manifest, error) {
+	//TODO Check if there is shared blob path?
```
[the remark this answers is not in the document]
mtrmac: That’s a good point, actually.

I think one possible answer is that `SystemContext.OCISharedBlobDirPath` did not get set by itself, and the user who set that implicitly opted into a behavior. That might either do exactly what was expected, or be pretty surprising.

Another possible answer is that a single repo just isn’t enough context to safely delete blobs, so we can’t do this correctly — and therefore the image deletion implementation, if it sees `OCISharedBlobDirPath` set, should immediately fail; the user can then either unset this option, or delete the image using… some other tool, if that exists. (And users who have a single-repo `OCISharedBlobDirPath` just can’t use c/image to delete blobs for that repo.)

I think both choices are basically reasonable and I don’t see a strong reason to prefer one over the other, so I think this is up to you.

Immediately failing with a “not supported (yet?)” error would probably be less and simpler code, that might be a relevant factor if we don’t have any stronger reason.

oci/layout/oci_transport.go
```diff
@@ -160,17 +164,25 @@ func (ref ociReference) NewImage(ctx context.Context, sys *types.SystemContext)
 // getIndex returns a pointer to the index references by this ociReference. If an error occurs opening an index nil is returned together
 // with an error.
 func (ref ociReference) getIndex() (*imgspecv1.Index, error) {
-	indexJSON, err := os.Open(ref.indexPath())
+	return parseIndex(ref.indexPath())
+}
+
+func parseJson[T any](path string) (*T, error) {
```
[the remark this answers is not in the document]
Pvlerick: :heavy_check_mark:

oci/layout/oci_transport.go
```diff
@@ -226,9 +299,169 @@ func (ref ociReference) NewImageDestination(ctx context.Context, sys *types.Syst
...
+		}
+	}
+
+	return nil
+}
+
+func saveJson(path string, content any) error {
+	// If the file already exists, get its mode to preserve it
+	var mode fs.FileMode
+	existinfFileInfo, err := os.Stat(path)
```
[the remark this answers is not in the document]
Pvlerick: :heavy_check_mark:

oci/layout/oci_transport.go
```diff
@@ -226,9 +299,169 @@ func (ref ociReference) NewImageDestination(ctx context.Context, sys *types.Syst
...
+		return ImageNotFoundError{ref}
+	}
+
+	manifest, err := ref.getManifest(imageDescriptorWrapper.descriptor)
+	if err != nil {
+		return err
+	}
+
+	// Delete all blobs used by this image only
+	blobsToDelete := make([]digest.Digest, 0, len(manifest.Layers)+2)
```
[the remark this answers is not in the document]
Pvlerick: :heavy_check_mark:

oci/layout/oci_transport.go
```diff
@@ -226,9 +299,169 @@ func (ref ociReference) NewImageDestination(ctx context.Context, sys *types.Syst
...
+			if len(index.Manifests) == 0 {
+				step = indexUpdateStep{"delete", indexDigest, nil}
+			} else {
+				// Save the new file
+				buffer := new(bytes.Buffer)
+				err = json.NewEncoder(buffer).Encode(index)
+				if err != nil {
+					return err
+				}
+				indexNewDigest := digest.SHA256.FromBytes(buffer.Bytes())
```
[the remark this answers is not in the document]
Pvlerick: :heavy_check_mark:

oci/layout/oci_transport_test.go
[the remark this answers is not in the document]
Pvlerick: :heavy_check_mark:

oci/layout/oci_transport.go
```diff
@@ -226,9 +299,169 @@ func (ref ociReference) NewImageDestination(ctx context.Context, sys *types.Syst
...
+					return err
+				}
+				err = saveJson(indexNewPath, index)
+				if err != nil {
+					return err
+				}
+				step = indexUpdateStep{"update", indexDigest, &indexNewDigest}
+			}
+			// Delete the current index; it is dangling by now as it'll either be empty or have a new hash
+			err = os.Remove(indexPath)
```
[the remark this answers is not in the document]
Pvlerick: :heavy_check_mark:

oci/layout/oci_transport.go
```diff
@@ -204,6 +216,67 @@ func (ref ociReference) getManifestDescriptor() (imgspecv1.Descriptor, error) {
...
+	descriptor *imgspecv1.Descriptor
+	indexChain []*string //in order of appearence, the first is always be index.json and the nested indexes, last one being the one where the descriptor was found in
+}
+
+// This will return all the descriptors of all the images found in the repository
+// It starts at the index.json and then walks all nested indexes
+// Each image descriptor is returned along with the index it was found, as well as it parents if it is a nested index
+func (ref ociReference) getAllImageDescriptorsInRegistry() ([]*descriptorWrapper, error) {
+	var getImageDescriptorsFromIndex func(indexChain []*string) ([]*descriptorWrapper, error)
+	getImageDescriptorsFromIndex = func(indexChain []*string) ([]*descriptorWrapper, error) {
```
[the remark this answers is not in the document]
Pvlerick: :heavy_check_mark:

oci/layout/oci_transport.go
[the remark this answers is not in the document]
Pvlerick: I wrote a test and implementation to delete what is found in the local directory and leave what is in the shared folder untouched.

Right now the assumption is that if a blob isn't found, it's likely already gone or in the shared folder. Normally, I would go a step further and check if the blob exists the the shared folder to make a distinction between the two cases and log appropriately, but I'd like your opinion as it adds a few ifs.

Pvlerick:
@mtrmac I finally had time to take a second shot a this.

Going back and forth with tests, I ended up with a different design.
In essence, blobs used in a descriptor in the index.json (descriptor being an image manifest or a nested index) are collected in a map which holds the usage count for each of these blobs (`getBlobsUsedInManifest` or `getBlobsUsedInIndex`). It is then compared to a similar map generated from index.json itself, and blobs that usage count reach 0 in the former map are maked for deletion (`getBlobsToDelete`). Blobs are then deleted and the relevant entry in the index is then deleted too.

oci/layout/oci_transport_test.go
```diff
@@ -370,3 +365,19 @@ func TestReferenceBlobPathInvalid(t *testing.T) {
 	_, err := ociRef.blobPath(hex, "")
 	assert.ErrorContains(t, err, "unexpected digest reference "+hex)
 }
+
```
Pvlerick, line 368: TODO: move this in oci_delete_test.go

oci/layout/oci_transport.go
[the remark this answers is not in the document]
mtrmac: `blobPath` only returns one value, based on whether `.OCISharedBlobDirPath` is set; so each call of c/image either _only_ uses the local directory, or _only_ uses the shared one. The current implementation does not allow producing/consuming an OCI archive where some blobs are in the local directory, and some others are in the shared directory.

So, it seems to me that “an implementation that deletes what is found in the local directory and leaves what is in the shared folder untouched” is, in effect, a “delete no blobs” implementation. And that’s, actually, another reasonable approach — one I didn’t think of.

If you prefer that behavior, sure, why not.

In that case I just think the code should be clear about that effect — i.e. I’d prefer to remove the code for checks for local blobs in shared-blob situations, so that there are no confusing dead pieces of code.

oci/layout/oci_delete_test.go
```diff
@@ -0,0 +1,254 @@
...
+	err = ref.DeleteImage(context.Background(), nil)
+	require.NoError(t, err)
+
+	// Check that the relevant blobs were deleted/preservend
+	blobsDir := filepath.Join(tmpDir, "blobs")
+	files, err := os.ReadDir(filepath.Join(blobsDir, "sha256"))
+	require.NoError(t, err)
+	require.Equal(t, 17, len(files))
+	blobExists(t, blobsDir, "sha256:df11bc189adeb50dadb3291a3a7f2c34b36e0efdba0df70f2c8a2d761b215cde")
+	blobDoesNotExist(t, blobsDir, "sha256:0dc27f36a618c110ae851662c13283e9fbc1b5a5de003befc4bcefa5a05d2eef")
```
mtrmac, line 144: Absolutely non-blocking: Naming these helpers `assertBlob…` would be a tiny bit more readable, indicating the responsibility to assert, not just to return a boolean value as a predicate-named function typically does.

mtrmac (commented):
I’m sorry about the late review.

This is great work, just two outstanding concerns:
- I’m quite worried about the code to finally delete the index entry getting out of sync
- The handling of shared blob directories can be simplified.

oci/layout/oci_delete.go
```diff
@@ -0,0 +1,244 @@
...
+		blobsUsedInIndex[desciptor.Digest]++
+		switch desciptor.MediaType {
+		case imgspecv1.MediaTypeImageManifest:
+			manifest, err := ref.getManifest(&desciptor, sharedBlobsDir)
+			if err != nil {
+				return nil, err
+			}
+			for digest, count := range ref.getBlobsUsedInManifest(manifest) {
+				blobsUsedInIndex[digest] += count
+			}
```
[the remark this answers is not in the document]
Pvlerick: Indeed; I made the change in a separate commit. However, I never liked functions that modify the arguments they are given, as a matter of taste/readability (I mostly worked with languages where this was discouraged).

If this is habitual in Go and in this codebase, I'll roll with it :-)

oci/layout/oci_transport.go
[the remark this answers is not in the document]
Pvlerick: I was indeed unable to find a way to have mixed content directory while tinkering with Skopeo.

In my view, there two considerations here:
1. If someone is using a shared directory, I guess it's because they expect to save space by only storing blobs once. With that in mind, since that shared directory is planned to be used across multiple local directories, so since we have no idea who's using this directory, deleting in there can potentially break other images - that's why I'd rather avoid it.
2. The mixed scenario is rather hypothetical looking at the tools using it right now, is it so far fetched to imagine it could be the case in the future? I agree with the no dead code principle, but in this case it's covered by a test and renders the library more robust.

Saying this, you're the main maintainer here so I'll go with what you prefer. If I have a bit of time, I'll make the changes for the all or nothing approach to compare.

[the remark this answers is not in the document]
Pvlerick: Here is a commit with the relevant changes, but I have mixed feelings about this: https://github.com/Pvlerick/image/commit/e0ce6adcab44a60a94ca4ff3e1c303fac5662020

In the end, it amounts to branching out very soon in `DeleteImage`, removing the sharedBobsDir everywhere and always assume that we are in a "local" mode (which is the part I don't like, I tend to be defensive and paranoid with code I write)

oci/layout/oci_delete.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+
+	if ref.image == "" && len(index.Manifests) == 1 {
+		index.Manifests = make([]imgspecv1.Descriptor, 0)
+		return saveJSON(ref.indexPath(), index)
+	}
+
+	newDescriptors := make([]imgspecv1.Descriptor, 0, len(index.Manifests)-1)
+	for _, descriptor := range index.Manifests {
+		if descriptor.Annotations[imgspecv1.AnnotationRefName] != ref.image {
```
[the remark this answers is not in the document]
Pvlerick: I added a test case for this and an extra condition in that if: 2807bce7

oci/layout/oci_delete.go
```diff
@@ -0,0 +1,245 @@
...
+	if err != nil {
+		if os.IsNotExist(err) {
+			logrus.Info(fmt.Sprintf("Blob at %q not found; it was either previously deleted or is in the shared blobs directory", blobPath))
+			return nil
+		} else {
+			return err
+		}
+	} else {
+		return nil
+	}
```
mtrmac, line 184: (Absolutely non-blocking:
```go
if err != nil && os.IsNotExist(err) {
    logrus…
    return nil // or err = nil
}
return err
```

would do the same thing, and maybe be a bit clearer that this only targets that one special case.)

oci/layout/oci_delete.go
```diff
@@ -0,0 +1,245 @@
...
+
+	newDescriptors := make([]imgspecv1.Descriptor, 0, len(index.Manifests)-1)
+	for i, descriptor := range index.Manifests {
+		// Double check: same image reference name and same position in the index
+		if descriptor.Annotations[imgspecv1.AnnotationRefName] != ref.image || referenceIndex != i {
+			newDescriptors = append(newDescriptors, descriptor)
+		}
+	}
+	index.Manifests = newDescriptors
```
mtrmac, line 205: I’d prefer just a `index.Manifests = slices.Delete(index.Manifests, referenceIndex, referenceIndex+1)` here. No special case for a single entry, no duplicate logic to maintain.

Alternatively, if you really prefer the double check here, please link the two functions with comments:
```go
// Update also ociReference.getManifestDescriptor
```
before starting to edit `index.Manifests` here, and a mirror comment in `getManifestDescriptor` pointing to this function.

oci/layout/oci_delete.go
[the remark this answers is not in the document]
mtrmac: This is non-blocking, I really doubt the performance difference will matter.

---

(I think, in principle, it would be idiomatic to have a
```
class blobCounter {
   countsSoFar: mutable map[Digest]int
}
```
and some `usedBlob`/`addBlobsUsedInIndex` methods on that `blobCounter` class, in almost all object-oriented  (not pure-functional). languages

And then I think that that a single-purpose single-field class is just a pure overhead, to readability most of all, so passing that map around without the class wrapper is just simpler.

But it _does_ require careful naming and/or documentation to not be surprising. I’m definitely not insisting on that approach.

oci/layout/oci_transport.go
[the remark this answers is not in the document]
mtrmac: Hum… I was rather thinking to keep the `sharedBlobsDir` parameter everywhere (to keep the code generic), and just do an

```go
func … DeleteImage() {
   …
   if sharedBlobsDir == "" { // Don't delete in shared blobs mode, we can’t accurately count references
      // …
      // blobsUsedByImage, err = ref.getBlobsUsedInSingleImage(&descriptor, sharedBlobsDir)
      // …
      // … ref.deleteBlobs …
   }
   return ref.deleteReferenceFromIndex(descriptorIndex)
```

but you’re right that keeping a `sharedBlobsDir` parameter while forcing it to be `""` is also redundant/dead code.

---

How about this?
- Keep the actual behavior as it is in this PR
- In `deleteBlobs`, use a rather more prominent comment, something like
  ```go
  // This transport never generates layouts where blobs for an image are both in the local blobs directory
  // and the shared one; it’s either one or the other, depending on how OCISharedBlobDirPath is set.
  //
  // But we can’t correctly compute use counts for OCISharedBlobDirPath (because we don't know what
  // the other layouts sharing that directory are, and we might not even have permission to read them),
  // so we can’t really delete any blobs in that case.
  // Checking the _local_ blobs directory, and deleting blobs from there, doesn't really hurt,
  // in case the layout was created using some other tool or without OCISharedBlobDirPath set, so let's silently
  // check for local blobs (but we should make no noise if the blobs are actually in the shared directory).
  //
  // So, NOTE: the blobPath() call below hard-codes "" even in calls where OCISharedBlobDirPath is set
  for _, digest := range blobsToDelete.Values() {
  		blobPath, err := ref.blobPath(digest, "") 
  ```
- And then ensure that `deleteBlob` really doesn’t produce any warnings if `sharedBlobsDir` is set.

mtrmac (commented):
Thanks!

- The shared blobs dir conversation continues in https://github.com/containers/image/pull/2003#discussion_r1353097120 , I think I’m converging much closer to your point
- A bit more for the top-level index deletion, please
- AFAICS the updated test for `getManifestDescriptor` is incorrect in a rather misleading way.

oci/layout/oci_transport_test.go
```diff
@@ -82,11 +103,13 @@ func TestGetManifestDescriptor(t *testing.T) {
...
 
-		res, err := ref.(ociReference).getManifestDescriptor()
+		res, i, err := ref.(ociReference).getManifestDescriptor()
 		if c.expected != nil {
 			require.NoError(t, err)
-			assert.Equal(t, *c.expected, res)
+			assert.Equal(t, 0, c.expected.index)
+			assert.Equal(t, *c.expected.descriptor, res)
 		} else {
+			assert.Equal(t, -1, i)
```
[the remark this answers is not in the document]
Pvlerick: Sorry about the sloppy changes in this file, I don't know what I was thinking when I did that yesterday.

oci/layout/oci_delete.go
Pvlerick: Ok, I see what you mean. I guess it depends on how comfortable we are on relying on the index alone. I was thinking that the code could check what's at that index before the final deletion, just to make sure, but judging from what you say, this is overkill.

oci/layout/oci_transport.go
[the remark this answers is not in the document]
Pvlerick: Sounds good to me.

Pvlerick:
> * AFAICS the updated test for `getManifestDescriptor` is incorrect in a rather misleading way.

Again, sorry about that :-(

oci/layout/oci_delete.go
mtrmac: I don’t think there’s a way to be _perfect_ here, notably against a concurrent modification[1]. We could plausibly read the top-level index _once_ into memory, do a lookup on the in-memory data, delete in the in-memory data, and write the index back, to avoid deleting an unexpected entry if a concurrent process reshuffles the index in the meantime; but that would still not make anything _reliable_ because the concurrent process could also add a new reference to a blob we intend to delete.

The transport’s writes already only work correctly with an externally-implemented guarantee that there is at most one writer at a time, and this is no different — so I think we might just as well use a simple implementation.


[1] Cross-platform file locking, or locking on UNIX from inside a library, is a mess, so the transport doesn’t do any.

oci/layout/oci_delete.go
```diff
@@ -0,0 +1,245 @@
...
+	index, err := ref.getIndex()
+	if err != nil {
+		return err
+	}
+
+	if ref.image == "" && len(index.Manifests) == 1 {
+		index.Manifests = make([]imgspecv1.Descriptor, 0)
+		return saveJSON(ref.indexPath(), index)
+	}
+
```
mtrmac, line 204: This can also be deleted; the `referenceIndex`-based deletion below has the same effect.

oci/layout/fixtures/manifest/index.json
```diff
@@ -1 +1,17 @@
...
+            "annotations": {
+                "org.opencontainers.image.ref.name": "v0.1.1"
+            },
+            "platform": {
+                "architecture": "amd64",
+                "os": "linux"
+            }
+        }
+    ]
+}
```
mtrmac, line 17: (Just to confirm, there are no semantic changes to this file, are there?)

mtrmac:
> > * AFAICS the updated test for `getManifestDescriptor` is incorrect in a rather misleading way.
> 
> Again, sorry about that :-(

No worries — I have written a much worse bug yesterday. I just wanted to highlight that as more important than some of the stylistic comments.

mtrmac (commented):
Thanks! LGTM.

Could you squash this to one or two commits, please?

oci/layout/fixtures/manifest/index.json
Pvlerick: Nope, just formatted it to be human readable.

Pvlerick:
Rebased in one commit (and on the latest main)

mtrmac (approved):
Thanks so much!

PR #2005, opened 2023-06-20
Enable mirror processing with external V2RegistriesConf

dcaravel:
This PR's allows for validating and finding mirrored registries with an externally supplied `sysregistriesv2.V2RegistriesConf`.

It does this by:
- exporting `func (config *V2RegistriesConf) postProcessRegistries()` as `PostProcessRegistries()`
- creating a new `FindRegistryWithConfig()` method that mimics `FindRegistry()` but does so without expecting the `V2RegistriesConf` to be sourced from the filesystem.

Use case / workflow would resemble:
1. Read all `ImageContentSourcePolicy`, `ImageDigestMirrorSet`, and `ImageTagMirrorSet` CR's from OCP API
2. Use [EditRegistriesConfig](https://github.com/openshift/runtime-utils/blob/5c488b20a19fc8c1fee9011c41ce70379bc8ca4d/pkg/registries/registries.go#L240) to create a `V2RegistriesConf` from the CRs
3. Run `config.PostProcessRegistries()` to cleanup/prepare the `config` for usage
4. Find registry using new `sysregistriesv2.FindRegistryWithConfig(config, name)`

The returned registry would be used to get list of mirrors via `PullSourcesFromReference`

High level example (err handling omitted for brevity):
```go
var config *sysregistriesv2.V2RegistriesConf
...

// build config from CRs
registries.EditRegistriesConfig(config, nil, nil, icspRules, idmsRules, itmsRules)

// post process the config
config.PostProcessRegistries()

// find the registry to use for a particular image name
reg, _ := sysregistriesv2.FindRegistryWithConfig(config, name)

// get mirror list from registry
ref, _ := reference.ParseNamed(name)
srcs, _ := reg.PullSourcesFromReference(ref)

...
```

### DIFF
--- a/pkg/sysregistriesv2/system_registries_v2.go
+++ b/pkg/sysregistriesv2/system_registries_v2.go
@@ -387,9 +387,9 @@ func (config *V1RegistriesConf) ConvertToV2() (*V2RegistriesConf, error) {
 // anchoredDomainRegexp is an internal implementation detail of postProcess, defining the valid values of elements of UnqualifiedSearchRegistries.
 var anchoredDomainRegexp = regexp.Delayed("^" + reference.DomainRegexp.String() + "$")
 
-// postProcess checks the consistency of all the configuration, looks for conflicts,
+// PostProcessRegistries checks the consistency of all the configuration, looks for conflicts,
 // and normalizes the configuration (e.g., sets the Prefix to Location if not set).
-func (config *V2RegistriesConf) postProcessRegistries() error {
+func (config *V2RegistriesConf) PostProcessRegistries() error {
 	regMap := make(map[string][]*Registry)
 
 	for i := range config.Registries {
@@ -415,6 +415,11 @@ func (config *V2RegistriesConf) postProcessRegistries() error {
 			// https://github.com/containers/image/pull/1191#discussion_r610622495
 			if !strings.HasPrefix(reg.Prefix, "*.") && reg.Location == "" {
 				return &InvalidRegistries{s: "invalid condition: location is unset and prefix is not in the format: *.example.com"}
+			}
+
+			if strings.HasPrefix(reg.Prefix, "*.") && strings.ContainsAny(reg.Prefix, "/@:") {
+				msg := fmt.Sprintf("wildcarded prefix should be in the format: *.example.com. Current prefix %q is incorrectly formatted", reg.Prefix)
+				return &InvalidRegistries{s: msg}
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -387,9 +387,9 @@ func (config *V1RegistriesConf) ConvertToV2() (*V2RegistriesConf, error) {
 // anchoredDomainRegexp is an internal implementation detail of postProcess, defining the valid values of elements of UnqualifiedSearchRegistries.
 var anchoredDomainRegexp = regexp.Delayed("^" + reference.DomainRegexp.String() + "$")
 
-// postProcess checks the consistency of all the configuration, looks for conflicts,
+// PostProcessRegistries checks the consistency of all the configuration, looks for conflicts,
 // and normalizes the configuration (e.g., sets the Prefix to Location if not set).
-func (config *V2RegistriesConf) postProcessRegistries() error {
+func (config *V2RegistriesConf) PostProcessRegistries() error {
```
mtrmac, line 392: I don’t think it makes sense to expose the output of this utility as a stable API; the internal parsing step might very well create a completely different in-memory format (plausibly some kind of trie to speed up lookups).

We would also end up with two different `V2RegistriesConf` “data types”, with different semantics and validation rules, that are represented as the same Go type.

mtrmac (commented):
To approach this from the other end: I think it’s reasonable to want to be able to write a c/image caller that provides a _full_ configuration and completely disregards on-disk state. (And longer-term the all-in-one `SystemContext` is not a good idea; that configuration / set of options should be a separate object).

That would imply some kind of (hardened-for-API-stability) “container registry configuration” object that can be either initialized from (arbitrary, mostly-opaque) on-filesystem config files, or built from individual inputs (at _some_ level of detail, ideally allowing the caller to specify configuration one registry at a time, but by now probably also by submitting a whole `V2RegistriesConf`). And `SystemContext` could refer to that “registry configuration” object; eventually a future version of this `systemregistriesv2` subpackage could work just with that “registry configuration” object without directly referring to `SystemContext` at all.

That would get us a “registry configuration” object. That object could be, from the point of external callers of c/image, entirely write-only; but given existing public functions like `FindRegistry` and `GetShortNameMode`, there would be ~zero additional risk to exposing the read side of that object to external callers as well — and that’s what you are after.

```go
// config, err := sysregistriesv3.FromHostFilesystem()
// config, err := sysregistriesv3context.FromSystemContext(&types.SystemContext{…})`
config, err := sysregistriesv3.New(
   sysregistriesv3.WithV2RegistriesConfFile("/path/to/file"),
   sysregistriesv3.WithV2RegistriesConfDropInDirectory("/path/to/dir"),
   sysregistriesv3.WithOwnedAliasState("/path/to/file"),
   sysregistriesv3.WithShortNameMode(…),
   sysregistriesv3.WithV2RegistriesConfData(V2RegistriesConf{…})
   // and other functional options
)
… = config.FindRegistry(…)
```

---

And yes, that’s an awful lot of work compared to just making some functions public.

pkg/sysregistriesv2/system_registries_v2.go
dcaravel: Admittedly I wasn't sold on exposing `V2RegistriesConf.PostProcessRegistries()`.

What if instead we had a specific `sysregistriesv2.PostProcessV2RegistriesConf(config)` or `sysregistriesv2.CleanConfig(config)` or something of that nature?  With that the `V2RegistriesConf` object stays as is, existing users of the API will not be impacted unless using the two new API methods (`FindRegistiesWithConfig` and `CleanConfig`).  

We can be clear in the docs these methods and maintaining a V2RegistriesConf externally is 'discouraged' (for the reasons mentioned).  That would buy us time until the when the `v3` re-write happens.

dcaravel:
Closing this PR based on the comments above, will pursue at a future date.  Thanks!

PR #2013, opened 2023-06-27
Bump to v5.26.0

TomSweeneyRedHat:
Bump to v5.26.0

I bumped to 5.27.0-dev, too, as I'm planning to create a release branch for v5.26.0, and that's where 5.26.1 will get generated.  We'll use v5.26.* for RHEL 8.8/9.3 either in Podman v4.6 or v4.7, the version number is to be decided.
[NO NEW TESTS NEEDED]

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -6,12 +6,12 @@ const (
 	// VersionMajor is for an API incompatible changes
 	VersionMajor = 5
 	// VersionMinor is for functionality in a backwards-compatible manner
-	VersionMinor = 25
+	VersionMinor = 26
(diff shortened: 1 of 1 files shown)


mtrmac (approved):
LGTM.

Actually there is no new APIs since 5.25.0, only bug fixes, behavior changes and internal work to allow future features — so this could plausibly be a 5.25.1.

OTOH e.g. now requiring TLS 1.2 might be a big enough deal to not do that in a patch release, at least per the `VersionPatch` documentation.

“Skipping” 5.26.1 is fair enough.

PR #2015, opened 2023-06-28
Adding IO decorator to copy progress bar

Pvlerick:
I was taking an interest in doing this https://github.com/containers/skopeo/issues/658 and stumbled upon this comment https://github.com/containers/skopeo/issues/658#issuecomment-916780456 from @vrothberg 

I don't see any testing around this, but if it's any comfort I tested it (starting from skopeo and using a local module) and it prints something like: 
```
Copying blob ad4592a9cb6d [===========================>----------] 39.1MiB / 52.4MiB | 9.0 MiB/s
```

I put the transfer rate at the end so that it doesn't keep on shifting the progress bar right and left and I only use one decimal but that can be changed.


mtrmac (commented):
Thanks!

I think something like this makes sense to provide. Alternatively, Docker seems to provide an ETA; I don’t have a strong opinion — and in the interactive UI, we can change things later, so I’m not too worried.

Cc: @vrothberg 

---

As for the implementation, I think at least the “unknown size” case should include the speed as well (of course that case couldn’t provide an ETA). For the `partial` case, the traffic is interactive enough that a rate average might not be too representative, so I’m fine with not adding that right now.

vrothberg (approved):
Thanks for contributing.  I like the idea and agree with @mtrmac's comment that IO should also be displayed when we don't know the size of the blob.

Pvlerick:
Yep, I overlooked the unknown (-1) blob size case - now it's there.

mtrmac (approved):
LGTM. Thanks again!

PR #2017, opened 2023-06-28
Ensure we close HTTP connections on all paths

mtrmac:
- Close `ImageSource` objects
- Close the HTTP client used by the `atomic` transport


vrothberg (approved):
LGTM

PR #2023, opened 2023-07-02
blob: `TryReusingBlobWithOptions` consider `RequiredCompression` if set

flouthoc:
TryReusingBlob now contains a new option `RequiredCompression` which filters the blob by checking against a compression of the blob which is considerd to be resued, in case `RequiredCompression` is set and `info` of the blob being reused does not matches, no blob is returned.

### DIFF
--- a/directory/directory_dest.go
+++ b/directory/directory_dest.go
@@ -190,6 +190,9 @@ func (d *dirImageDestination) PutBlobWithOptions(ctx context.Context, stream io.
 // If the blob has been successfully reused, returns (true, info, nil).
 // If the transport can not reuse the requested blob, TryReusingBlob returns (false, {}, nil); it returns a non-nil error only on an unexpected failure.
 func (d *dirImageDestination) TryReusingBlobWithOptions(ctx context.Context, info types.BlobInfo, options private.TryReusingBlobOptions) (bool, private.ReusedBlob, error) {
+	if !impl.OriginalBlobMatchesRequiredCompression(options) {
+		return false, private.ReusedBlob{}, nil
+	}
 	if info.Digest == "" {
 		return false, private.ReusedBlob{}, fmt.Errorf("Can not check for a blob with unknown digest")
 	}
--- a/docker/docker_image_dest.go
+++ b/docker/docker_image_dest.go
@@ -321,13 +321,21 @@ func (d *dockerImageDestination) TryReusingBlobWithOptions(ctx context.Context, 
 		return false, private.ReusedBlob{}, errors.New("Can not check for a blob with unknown digest")
 	}
 
-	// First, check whether the blob happens to already exist at the destination.
-	haveBlob, reusedInfo, err := d.tryReusingExactBlob(ctx, info, options.Cache)
-	if err != nil {
-		return false, private.ReusedBlob{}, err
-	}
-	if haveBlob {
-		return true, reusedInfo, nil
+	if impl.OriginalBlobMatchesRequiredCompression(options) {
+		// First, check whether the blob happens to already exist at the destination.
+		haveBlob, reusedInfo, err := d.tryReusingExactBlob(ctx, info, options.Cache)
+		if err != nil {
+			return false, private.ReusedBlob{}, err
+		}
+		if haveBlob {
+			return true, reusedInfo, nil
+		}
+	} else {
+		requiredCompression := "nil"
+		if options.OriginalCompression != nil {
+			requiredCompression = options.OriginalCompression.Name()
+		}
+		logrus.Debugf("Ignoring exact blob match case due to compression mismatch ( %s vs %s )", options.RequiredCompression.Name(), requiredCompression)
 	}
 
 	// Then try reusing blobs from other locations.
@@ -336,6 +344,19 @@ func (d *dockerImageDestination) TryReusingBlobWithOptions(ctx context.Context, 
 		candidateRepo, err := parseBICLocationReference(candidate.Location)
 		if err != nil {
 			logrus.Debugf("Error parsing BlobInfoCache location reference: %s", err)
+			continue
+		}
+		compressionOperation, compressionAlgorithm, err := blobinfocache.OperationAndAlgorithmForCompressor(candidate.CompressorName)
+		if err != nil {
+			logrus.Debugf("OperationAndAlgorithmForCompressor Failed: %v", err)
+			continue
+		}
+		if !impl.BlobMatchesRequiredCompression(options, compressionAlgorithm) {
+			requiredCompression := "nil"
+			if compressionAlgorithm != nil {
+				requiredCompression = compressionAlgorithm.Name()
+			}
+			logrus.Debugf("Ignoring candidate blob %s as reuse candidate due to compression mismatch ( %s vs %s ) in %s", candidate.Digest.String(), options.RequiredCompression.Name(), requiredCompression, candidateRepo.Name())
 			continue
 		}
 		if candidate.CompressorName != blobinfocache.Uncompressed {
@@ -387,12 +408,6 @@ func (d *dockerImageDestination) TryReusingBlobWithOptions(ctx context.Context, 
 		}
 
 		options.Cache.RecordKnownLocation(d.ref.Transport(), bicTransportScope(d.ref), candidate.Digest, newBICLocationReference(d.ref))
-
-		compressionOperation, compressionAlgorithm, err := blobinfocache.OperationAndAlgorithmForCompressor(candidate.CompressorName)
-		if err != nil {
-			logrus.Debugf("... Failed: %v", err)
-			continue
-		}
 
 		return true, private.ReusedBlob{
 			Digest:               candidate.Digest,
--- a/internal/imagedestination/impl/helpers.go
+++ b/internal/imagedestination/impl/helpers.go
@@ -1,0 +1,20 @@
+package impl
+
+import (
+	"github.com/containers/image/v5/internal/private"
+	compression "github.com/containers/image/v5/pkg/compression/types"
+)
+
+// BlobMatchesRequiredCompression validates if compression is required by the caller while selecting a blob, if it is required
+// then function performs a match against the compression requested by the caller and compression of existing blob
+// (which can be nil to represent uncompressed or unknown)
+func BlobMatchesRequiredCompression(options private.TryReusingBlobOptions, candidateCompression *compression.Algorithm) bool {
+	if options.RequiredCompression == nil {
+		return true // no requirement imposed
+	}
+	return candidateCompression != nil && (options.RequiredCompression.Name() == candidateCompression.Name())
+}
+
+func OriginalBlobMatchesRequiredCompression(opts private.TryReusingBlobOptions) bool {
+	return BlobMatchesRequiredCompression(opts, opts.OriginalCompression)
+}
--- a/internal/imagedestination/impl/helpers_test.go
+++ b/internal/imagedestination/impl/helpers_test.go
@@ -1,0 +1,29 @@
+package impl
+
+import (
+	"testing"
+
+	"github.com/containers/image/v5/internal/private"
+	"github.com/containers/image/v5/pkg/compression"
+	compressionTypes "github.com/containers/image/v5/pkg/compression/types"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestBlobMatchesRequiredCompression(t *testing.T) {
+	var opts private.TryReusingBlobOptions
+	cases := []struct {
+		requiredCompression  *compressionTypes.Algorithm
+		candidateCompression *compressionTypes.Algorithm
+		result               bool
+	}{
+		{&compression.Zstd, &compression.Zstd, true},
+		{&compression.Gzip, &compression.Zstd, false},
+		{&compression.Zstd, nil, false},
+		{nil, &compression.Zstd, true},
+	}
+
+	for _, c := range cases {
+		opts = private.TryReusingBlobOptions{RequiredCompression: c.requiredCompression}
+		assert.Equal(t, BlobMatchesRequiredCompression(opts, c.candidateCompression), c.result)
+	}
+}
--- a/internal/private/private.go
+++ b/internal/private/private.go
@@ -112,10 +112,11 @@ type TryReusingBlobOptions struct {
 	// Transports, OTOH, MUST support these fields being zero-valued for types.ImageDestination callers
 	// if they use internal/imagedestination/impl.Compat;
 	// in that case, they will all be consistently zero-valued.
-
-	EmptyLayer bool            // True if the blob is an "empty"/"throwaway" layer, and may not necessarily be physically represented.
-	LayerIndex *int            // If the blob is a layer, a zero-based index of the layer within the image; nil otherwise.
-	SrcRef     reference.Named // A reference to the source image that contains the input blob.
+	RequiredCompression *compression.Algorithm // If set, reuse blobs with a matching algorithm as per implementations in internal/imagedestination/impl.helpers.go
+	OriginalCompression *compression.Algorithm // Must be set if RequiredCompression is set; can be set to nil to indicate “uncompressed” or “unknown”.
+	EmptyLayer          bool                   // True if the blob is an "empty"/"throwaway" layer, and may not necessarily be physically represented.
+	LayerIndex          *int                   // If the blob is a layer, a zero-based index of the layer within the image; nil otherwise.
+	SrcRef              reference.Named        // A reference to the source image that contains the input blob.
 }
 
 // ReusedBlob is information about a blob reused in a destination.
--- a/ostree/ostree_dest.go
+++ b/ostree/ostree_dest.go
@@ -335,6 +335,9 @@ func (d *ostreeImageDestination) importConfig(repo *otbuiltin.Repo, blob *blobTo
 // reflected in the manifest that will be written.
 // If the transport can not reuse the requested blob, TryReusingBlob returns (false, {}, nil); it returns a non-nil error only on an unexpected failure.
 func (d *ostreeImageDestination) TryReusingBlobWithOptions(ctx context.Context, info types.BlobInfo, options private.TryReusingBlobOptions) (bool, private.ReusedBlob, error) {
+	if !impl.OriginalBlobMatchesRequiredCompression(options) {
+		return false, private.ReusedBlob{}, nil
+	}
 	if d.repo == nil {
 		repo, err := openRepo(d.ref.repo)
 		if err != nil {
--- a/storage/storage_dest.go
+++ b/storage/storage_dest.go
@@ -307,6 +307,9 @@ func (s *storageImageDestination) PutBlobPartial(ctx context.Context, chunkAcces
 // If the blob has been successfully reused, returns (true, info, nil).
 // If the transport can not reuse the requested blob, TryReusingBlob returns (false, {}, nil); it returns a non-nil error only on an unexpected failure.
 func (s *storageImageDestination) TryReusingBlobWithOptions(ctx context.Context, blobinfo types.BlobInfo, options private.TryReusingBlobOptions) (bool, private.ReusedBlob, error) {
+	if !impl.OriginalBlobMatchesRequiredCompression(options) {
+		return false, private.ReusedBlob{}, nil
+	}
 	reused, info, err := s.tryReusingBlobAsPending(blobinfo.Digest, blobinfo.Size, &options)
 	if err != nil || !reused || options.LayerIndex == nil {
 		return reused, info, err
--- a/docker/internal/tarfile/dest.go
+++ b/docker/internal/tarfile/dest.go
@@ -129,6 +129,9 @@ func (d *Destination) PutBlobWithOptions(ctx context.Context, stream io.Reader, 
 // If the blob has been successfully reused, returns (true, info, nil).
 // If the transport can not reuse the requested blob, TryReusingBlob returns (false, {}, nil); it returns a non-nil error only on an unexpected failure.
 func (d *Destination) TryReusingBlobWithOptions(ctx context.Context, info types.BlobInfo, options private.TryReusingBlobOptions) (bool, private.ReusedBlob, error) {
+	if !impl.OriginalBlobMatchesRequiredCompression(options) {
+		return false, private.ReusedBlob{}, nil
+	}
 	if err := d.archive.lock(); err != nil {
 		return false, private.ReusedBlob{}, err
 	}
--- a/internal/imagedestination/wrapper.go
+++ b/internal/imagedestination/wrapper.go
@@ -64,6 +64,9 @@ func (w *wrapped) PutBlobWithOptions(ctx context.Context, stream io.Reader, inpu
 // If the blob has been successfully reused, returns (true, info, nil).
 // If the transport can not reuse the requested blob, TryReusingBlob returns (false, {}, nil); it returns a non-nil error only on an unexpected failure.
 func (w *wrapped) TryReusingBlobWithOptions(ctx context.Context, info types.BlobInfo, options private.TryReusingBlobOptions) (bool, private.ReusedBlob, error) {
+	if options.RequiredCompression != nil {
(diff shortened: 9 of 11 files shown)


flouthoc:
I tested the docker blob's in combination with a variant of code from https://github.com/containers/image/pull/1987 and it is working as expected. I made similar changes to other transport as well where change was applicable.

flouthoc:
@mtrmac PTAL

directory/directory_dest.go
```diff
@@ -193,6 +193,9 @@ func (d *dirImageDestination) TryReusingBlobWithOptions(ctx context.Context, inf
 	if info.Digest == "" {
 		return false, private.ReusedBlob{}, fmt.Errorf("Can not check for a blob with unknown digest")
 	}
+	if options.RequiredCompression != nil && (options.RequiredCompression != info.CompressionAlgorithm) {
```
mtrmac, line 196: I think it would be easier if the new option only applied in the `CanSubstitute` case; then these simple implementations that never substitute blobs wouldn’t need to add extra conditions.

docker/docker_image_dest.go
```diff
@@ -394,11 +394,13 @@ func (d *dockerImageDestination) TryReusingBlobWithOptions(ctx context.Context,
 			continue
 		}
 
-		return true, private.ReusedBlob{
-			Digest:               candidate.Digest,
-			Size:                 size,
-			CompressionOperation: compressionOperation,
-			CompressionAlgorithm: compressionAlgorithm}, nil
+		if options.RequiredCompression == nil || (options.RequiredCompression == compressionAlgorithm) {
```
mtrmac, line 397: This check should happen before making any on-registry operations (`blobExists`/`mountBlob`)

mtrmac, line 397: Hum, sadly `compression.AlgorithmByName` returns a `compression.Algorithm` value, so the `*compression.Algorithm` pointer is not a unique identifier that can be compared for equality; I think this needs to compare `.Name()`.

internal/private/private.go
```diff
@@ -112,10 +112,10 @@ type TryReusingBlobOptions struct {
 	// Transports, OTOH, MUST support these fields being zero-valued for types.ImageDestination callers
 	// if they use internal/imagedestination/impl.Compat;
 	// in that case, they will all be consistently zero-valued.
```
mtrmac, line 114: A compatibility handler for external `ImageDestinations` is missing for the new field:

`internal/imagedestination/wrapper.go`. I think if `RequiredCompression` is set, the call should silently return without reusing anything; we could theoretically call `TryReusingBlob` and checking whether the returned blob has the right algorithm, but that’s AFAICS unsafe — `ImageDestination` is stateful and the destination can assume, that if `TryReusingBlob` returns something, it is going to be used. (Compare `storageImageDestination`.)

storage/storage_dest.go
```diff
@@ -307,6 +307,9 @@ func (s *storageImageDestination) PutBlobPartial(ctx context.Context, chunkAcces
 // If the blob has been successfully reused, returns (true, info, nil).
 // If the transport can not reuse the requested blob, TryReusingBlob returns (false, {}, nil); it returns a non-nil error only on an unexpected failure.
 func (s *storageImageDestination) TryReusingBlobWithOptions(ctx context.Context, blobinfo types.BlobInfo, options private.TryReusingBlobOptions) (bool, private.ReusedBlob, error) {
+	if options.RequiredCompression != nil && (options.RequiredCompression != blobinfo.CompressionAlgorithm) {
```
mtrmac, line 310: (It’s an “interesting” question whether a storage destination, which pretends to store compressed blobs but always actually stores uncompressed variants, should care about `RequiredCompression`. In practice it makes no difference at all, I think.)

internal/private/private.go
```diff
@@ -112,10 +112,10 @@ type TryReusingBlobOptions struct {
...
-
-	EmptyLayer bool            // True if the blob is an "empty"/"throwaway" layer, and may not necessarily be physically represented.
-	LayerIndex *int            // If the blob is a layer, a zero-based index of the layer within the image; nil otherwise.
-	SrcRef     reference.Named // A reference to the source image that contains the input blob.
+	RequiredCompression *compression.Algorithm // If set the function will check if the blob which is being used as the candidate has the required compression.
```
mtrmac, line 115:
```suggestion
	RequiredCompression *compression.Algorithm // If set, only reuse blobs with the required compression.
```
or see elsewhere about the interaction with `CanSubstitute`.

directory/directory_dest.go
flouthoc: Removed for transports which are not using `CanSubstitute`

docker/docker_image_dest.go
flouthoc: Done.

flouthoc: Done.

internal/private/private.go
flouthoc: Done.

internal/private/private.go
flouthoc: Added compat handler

mtrmac:
> I think it would be easier if the new option only applied in the `CanSubstitute` case; then these simple implementations that never substitute blobs wouldn’t need to add extra conditions.

On second thought, I’m afraid that doesn’t work well. For converting from gzip to zstd, we ideally want an operation to express “I have a gzip blob, but I want zstd; so check if there is a zstd variant, but don’t use a gzip one even if it is an exact match”. In that case `CanSubstitute` will be set, but even an exact match needs to consider the `RequiredCompression` condition.

So every single transport needs to deal with this, sadly, and your original approach was correct.

---

The `options.RequiredCompression == nil || (options.RequiredCompression.Name() == info.CompressionAlgorithm.Name())` condition will, then, be copy&pasted so many times that it seems useful to share it as a helper function in the `internal/imagedestination/impl` package.

---

Note that the `info.CompressionAlghorithm` field is not, in general, expected to be provided to `TryReusingBlob` (really adding that field has been a mistake.) We do set it in the copy pipeline, but really it is only documented as an input to `.UpdateLayerInfos`. A minimal fix would be to document that if `RequiredCompression` is set, `info.CompressionAlgorithm` must be set as well; I think it would be cleaner instead, to provide the algorithm specifically as an input for `TryReusingBlobWithOptions`, as a new field `TryReusingBlobOptions.OriginalCompression` or something like that. And then document that it must be set along with `RequiredCompression`.

flouthoc:
@mtrmac PTAL

internal/imagedestination/impl/compat.go
```diff
@@ -99,3 +99,9 @@ func (c *Compat) PutSignatures(ctx context.Context, signatures [][]byte, instanc
 	}
 	return c.dest.PutSignaturesWithFormat(ctx, withFormat, instanceDigest)
 }
+
+// MatchesRequiredCompression validates if compression is required by the caller while selecting a blob, if it is required
+// then function performs a match against the compression requested by the caller and compression of exisiting blob.
+func (c *Compat) MatchesRequiredCompression(options private.TryReusingBlobOptions, info types.BlobInfo) bool {
```
mtrmac, line 105: - This has nothing to do with the purpose of `Compat`
- Making it a public method on an unnamed field of the `ImageDestination` makes it conceptually visible to external users via interface casts … that’s not at all a major concern (callers that do undocumented things like that are just broken), but if it is easy to avoid…

I think a stand-alone public function would be cleaner.

internal/imagedestination/impl/compat.go
```diff
@@ -99,3 +99,9 @@ func (c *Compat) PutSignatures(ctx context.Context, signatures [][]byte, instanc
...
+	return options.RequiredCompression == nil || (options.RequiredCompression.Name() == info.CompressionAlgorithm.Name())
```
mtrmac, line 106: Note the part of https://github.com/containers/image/pull/2023#issuecomment-1624461020 about `info.CompressionAlgorithm` in general not being available…

I think that value should be in `TryReusingBlobOptions`, and then this can be just a
```go
func {CanReuse,OptionsAllowReusing}ExactBlob(options private.TryReusingBlobOptions) bool
```

Or, alternatively, `BlobMatchesRequiredCompression(options …, *compressiontypes.Algorithm) bool` (not requiring a `BlobInfo` would allow this to be used in the `CanSubstitute` case as well.

mtrmac (commented):
I’m sorry, for some reason 3 comments were left pending.

internal/imagedestination/impl/compat.go
mtrmac: Place it in `imagedestination/impl/{shared, helpers,reuse}.go`, or something like that maybe.

docker/docker_image_dest.go
```diff
@@ -326,7 +326,7 @@ func (d *dockerImageDestination) TryReusingBlobWithOptions(ctx context.Context,
 	if err != nil {
 		return false, private.ReusedBlob{}, err
 	}
-	if haveBlob {
+	if haveBlob && (options.RequiredCompression == nil || (options.RequiredCompression.Name() == info.CompressionAlgorithm.Name())) {
```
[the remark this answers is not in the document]
flouthoc: Moving this before `tryReusingExactBlob` would need me to add a condition like this
```go
       if !impl.BlobMatchesRequiredCompression(options, info.CompressionAlgorithm) {
                return false, private.ReusedBlob{}, nil
        }
```

but in this case since substitute is allowed I want other candidates to be checked further below in the code, so it seemed appropriate to check with when `exact` blob ( i.e `haveBlob`) is found and return `true` for that blob if it matches the condition as compared to returning `false` early.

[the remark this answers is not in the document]
mtrmac: Would
```go
if matchesCompression {
   if tryReusingExactBlob() succeeded {
     return success
  }
}
… Cache.CacheLocations2 …
```
not work?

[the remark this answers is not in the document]
flouthoc: yes that would work

docker/docker_image_dest.go
```diff
@@ -321,13 +321,15 @@ func (d *dockerImageDestination) TryReusingBlobWithOptions(ctx context.Context,
...
+	if impl.BlobMatchesRequiredCompression(options, info.CompressionAlgorithm) {
+		// First, check whether the blob happens to already exist at the destination.
+		haveBlob, reusedInfo, err := d.tryReusingExactBlob(ctx, info, options.Cache)
+		if err != nil {
+			return false, private.ReusedBlob{}, err
+		}
+		if haveBlob {
+			return true, reusedInfo, nil
+		}
 	}
```
mtrmac, line 333: `else { logrus.Debugf… }`

docker/docker_image_dest.go
```diff
@@ -363,6 +365,15 @@ func (d *dockerImageDestination) TryReusingBlobWithOptions(ctx context.Context,
 			remoteName:   reference.Path(candidateRepo),
 			actions:      "pull",
 		}
+		compressionOperation, compressionAlgorithm, err := blobinfocache.OperationAndAlgorithmForCompressor(candidate.CompressorName)
```
mtrmac, line 368: `extraScope` is only necessary afterwards, so this new check can be performed earlier.

In fact it would be best to add the new check just before the `if candidate.CompressorName != blobinfocache.Uncompressed` line, so that
- all code dealing with `CompressorName` is in a single location
- we don’t log both “trying to reuse” and “ignoring”, only exactly one of the two.

docker/docker_image_dest.go
```diff
@@ -363,6 +365,15 @@ func (d *dockerImageDestination) TryReusingBlobWithOptions(ctx context.Context,
...
+		if err != nil {
+			logrus.Debugf("... Failed: %v", err)
+			continue
+		}
+		if !impl.BlobMatchesRequiredCompression(options, compressionAlgorithm) {
```
mtrmac, line 373: `compressionAlgorithm` can be `nil` on `Uncompressed` and `UnknownCompression`; right now that would cause a crash.

docker/docker_image_dest.go
```diff
@@ -363,6 +365,15 @@ func (d *dockerImageDestination) TryReusingBlobWithOptions(ctx context.Context,
...
+			logrus.Debugf("Ignoring candidate blob %s as reuse candidate since compression %s was requested and blob has compression %s", candidate.Digest.String(), options.RequiredCompression.Name(), compressionAlgorithm.Name())
```
mtrmac, line 374: This also needs to not crash on `compressionAlgorithm == nil`.

(Also compare the “Trying to reuse” code; including `candidateRepo.Name()` seems useful — especially if the code is moved so that the “trying” logs aren’t emitted at all.)

(Absolutely non-blocking: making this a bit shorter would help on narrow terminals / log viewers: “Not using …, compression mismatch (%s vs. %s)”. The text should make sense but it can probably be assumed the reader can refer to the related source code, so it doesn’t need to be quite self-documenting.)

docker/docker_image_dest.go
```diff
@@ -363,6 +365,15 @@ func (d *dockerImageDestination) TryReusingBlobWithOptions(ctx context.Context,
 			remoteName:   reference.Path(candidateRepo),
 			actions:      "pull",
 		}
+		compressionOperation, compressionAlgorithm, err := blobinfocache.OperationAndAlgorithmForCompressor(candidate.CompressorName)
+		if err != nil {
+			logrus.Debugf("... Failed: %v", err)
```
mtrmac, line 370: (Non-blocking, this is pre-existing: “… failed” does not add much, and it is especially unhelpful if there are multiple paths where exactly the same text is used. `OperationAndAlgorithmForCompressor failed`, or something like that, would make a difference.)

mtrmac (commented):
- `blobCacheDestination.TryReusingBlobWithOptions` also needs to implement this
- AFAICS the docker transport can crash
- https://github.com/containers/image/pull/2023#discussion_r1256656745 : I’d prefer not to add the new semantics of carrying the original blob’s algorithm in `BlobInfo`; one more field in `TryReusingBlobOptions` seems more explicit to me

docker/docker_image_dest.go
flouthoc: Done.

flouthoc:
@mtrmac Could you PTAL again.

docker/docker_image_dest.go
```diff
@@ -338,6 +346,21 @@ func (d *dockerImageDestination) TryReusingBlobWithOptions(ctx context.Context,
 			logrus.Debugf("Error parsing BlobInfoCache location reference: %s", err)
 			continue
 		}
+		compressionOperation, compressionAlgorithm, err := blobinfocache.OperationAndAlgorithmForCompressor(candidate.CompressorName)
+		if err != nil {
+			logrus.Debugf("OperationAndAlgorithmForCompressor Failed: %v", err)
+			continue
+		}
+		opts := options
+		opts.OriginalCompression = compressionAlgorithm
```
mtrmac, line 355: The value for this candidate is not “original blob’s compression”.

Split the helper into two instead?

```go
func BlobMatchesRequiredCompression(opts, alghorithmOrNil) {
   …
}

func OriginalBlobMatchesRequiredCompression(opts) {
   return BlobMatchesRequiredCompression(opts, opts.OriginalAlgorithm)
}
```
Then most transports can use `Original…`, and this path the other one.

internal/imagedestination/impl/helpers.go
```diff
@@ -0,0 +1,14 @@
...
+	"github.com/containers/image/v5/internal/private"
+)
+
+// BlobMatchesRequiredCompression validates if compression is required by the caller while selecting a blob, if it is required
+// then function performs a match against the compression requested by the caller and compression of existing blob.
+func BlobMatchesRequiredCompression(options private.TryReusingBlobOptions) bool {
+	if options.RequiredCompression != nil && options.OriginalCompression == nil {
+		return false
+	}
+	return options.RequiredCompression == nil || (options.RequiredCompression.Name() == options.OriginalCompression.Name())
```
mtrmac, line 13: This is
```go
if options.RequiredCompression == nil {
   return true // no requirement imposed
}
// a specific algorithm is required
return options.OriginalCompression != nil && /* .Name() matches */ // … and we got that one
```

which is, I think, easier to follow (and less redundant).

internal/private/private.go
```diff
@@ -112,10 +112,11 @@ type TryReusingBlobOptions struct {
 	// Transports, OTOH, MUST support these fields being zero-valued for types.ImageDestination callers
 	// if they use internal/imagedestination/impl.Compat;
 	// in that case, they will all be consistently zero-valued.
-
-	EmptyLayer bool            // True if the blob is an "empty"/"throwaway" layer, and may not necessarily be physically represented.
-	LayerIndex *int            // If the blob is a layer, a zero-based index of the layer within the image; nil otherwise.
-	SrcRef     reference.Named // A reference to the source image that contains the input blob.
+	RequiredCompression *compression.Algorithm // If set, only reuse blobs with a matching algorithm OriginalCompression.Name()
```
mtrmac, line 115: Yes, that’s how the helper now works… but that’s only because the docker transport is putting a badly-fitting value into `OriginalCompression`.

internal/private/private.go
```diff
@@ -112,10 +112,11 @@ type TryReusingBlobOptions struct {
...
+	OriginalCompression *compression.Algorithm // Must be set if RequiredCompression is set.
```
mtrmac, line 116:
```suggestion
	OriginalCompression *compression.Algorithm // Must be set if RequiredCompression is set; can be set to nil to indicate “uncompressed” or “unknown”.
```

mtrmac (commented):
Thanks, one last cleanup round please.

internal/private/private.go
flouthoc: I have modified this a bit, since we have two functions in helpers using this differently.

ostree/ostree_dest.go
```diff
@@ -335,6 +335,9 @@ func (d *ostreeImageDestination) importConfig(repo *otbuiltin.Repo, blob *blobTo
 // reflected in the manifest that will be written.
 // If the transport can not reuse the requested blob, TryReusingBlob returns (false, {}, nil); it returns a non-nil error only on an unexpected failure.
 func (d *ostreeImageDestination) TryReusingBlobWithOptions(ctx context.Context, info types.BlobInfo, options private.TryReusingBlobOptions) (bool, private.ReusedBlob, error) {
+	if !impl.BlobMatchesRequiredCompression(options) {
```
mtrmac, line 338: I think this also needs updating (`ostree` is not compiled by default … so we keep breaking it.)

internal/imagedestination/impl/helpers.go
```diff
@@ -0,0 +1,19 @@
...
+	compression "github.com/containers/image/v5/pkg/compression/types"
+)
+
+// BlobMatchesRequiredCompression validates if compression is required by the caller while selecting a blob, if it is required
+// then function performs a match against the compression requested by the caller and compression of existing blob.
+func BlobMatchesRequiredCompression(options private.TryReusingBlobOptions, originalCompression *compression.Algorithm) bool {
+	if options.RequiredCompression == nil {
+		return true // no requirement imposed
+	}
+	return originalCompression != nil || (options.RequiredCompression.Name() == originalCompression.Name())
```
mtrmac, line 14:
```suggestion
	return originalCompression != nil && (options.RequiredCompression.Name() == originalCompression.Name())
```
AFAICS

mtrmac (commented):
Also one outstanding item from https://github.com/containers/image/pull/2023#pullrequestreview-1522581438:
> `blobCacheDestination.TryReusingBlobWithOptions` also needs to implement this

internal/imagedestination/impl/helpers.go
mtrmac: (I don’t generally think a unit test is required for a three-line function like this … but it would have helped here.)

internal/imagedestination/impl/helpers.go
```diff
@@ -0,0 +1,19 @@
+package impl
+
+import (
+	"github.com/containers/image/v5/internal/private"
+	compression "github.com/containers/image/v5/pkg/compression/types"
+)
+
+// BlobMatchesRequiredCompression validates if compression is required by the caller while selecting a blob, if it is required
+// then function performs a match against the compression requested by the caller and compression of existing blob.
+func BlobMatchesRequiredCompression(options private.TryReusingBlobOptions, originalCompression *compression.Algorithm) bool {
```
mtrmac, line 10:
```suggestion
func BlobMatchesRequiredCompression(options private.TryReusingBlobOptions, blobCompression *compression.Algorithm) bool {
```
or `candidateCompression`… the Docker transports calls it with replacement candidates, not always with the original one we are trying to replace.

internal/imagedestination/impl/helpers.go
```diff
@@ -0,0 +1,19 @@
+package impl
+
+import (
+	"github.com/containers/image/v5/internal/private"
+	compression "github.com/containers/image/v5/pkg/compression/types"
+)
+
+// BlobMatchesRequiredCompression validates if compression is required by the caller while selecting a blob, if it is required
+// then function performs a match against the compression requested by the caller and compression of existing blob.
```
mtrmac, line 9:
```suggestion
// then function performs a match against the compression requested by the caller and compression of existing blob
// (which can be nil to represent uncompressed or unknown)
```

internal/imagedestination/impl/helpers.go
flouthoc: Oops, nice catch. Added a quick unit test for this.

ostree/ostree_dest.go
flouthoc: I did not get this comment, do you mean we should not add this to `ostree` ?

internal/imagedestination/impl/helpers.go
flouthoc: Modified tests a bit.

flouthoc:
```go
func TestBlobMatchesRequiredCompression(t *testing.T) {
	var opts private.TryReusingBlobOptions
	cases := []struct {
		requiredCompression  *compressionTypes.Algorithm
		candidateCompression *compressionTypes.Algorithm
		result               bool
	}{
		{&compression.Zstd, &compression.Zstd, true},
		{&compression.Gzip, &compression.Zstd, false},
		{&compression.Zstd, nil, false},
		{nil, &compression.Zstd, true},
	}

	for _, c := range cases {
		opts = private.TryReusingBlobOptions{RequiredCompression: c.requiredCompression}
		assert.Equal(t, BlobMatchesRequiredCompression(opts, c.candidateCompression), c.result)
	}
}
```

ostree/ostree_dest.go
mtrmac: Doesn’t this need `s/Blob/OriginalBlob/`?

Try `go test -tags containers_image_ostree ./ostree` …

… oh my. The tests have been broken … \*checks\* … consecutively since 2022-03-16, three different ways, all by me.

mtrmac (commented):
Tbajs

internal/imagedestination/impl/helpers_test.go
```diff
@@ -0,0 +1,29 @@
...
+	}{
+		{&compression.Zstd, &compression.Zstd, true},
+		{&compression.Gzip, &compression.Zstd, false},
+		{&compression.Zstd, nil, false},
+		{nil, &compression.Zstd, true},
+	}
+
+	for _, c := range cases {
+		opts = private.TryReusingBlobOptions{RequiredCompression: c.requiredCompression}
+		assert.Equal(t, BlobMatchesRequiredCompression(opts, c.candidateCompression), c.result)
```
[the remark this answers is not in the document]
flouthoc: Ackd. I will clean this in another follow-up PR.

ostree/ostree_dest.go
mtrmac: > The tests have been broken

For the record, #2037 should fix that (but it doesn’t add the subpackage to CI, out of both laziness and curiosity of how long it will be broken next time).

internal/imagedestination/impl/helpers_test.go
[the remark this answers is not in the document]
flouthoc: Resolved here: https://github.com/containers/image/pull/2039

PR #2029, opened 2023-07-06
manifest: `ListUpdate` add `imgspecv1.Platform` field

flouthoc:
`c/image` uses `Instance(` API to get the required details of an instance while performing replication, `Platform` of instance is needed so it can use it again while adding an instance hence adding `Platform` to `ListUpdate` 

Needed by: https://github.com/containers/image/pull/1987
See prior comment here: https://github.com/containers/image/pull/1875#discussion_r1128464266

### DIFF
--- a/internal/manifest/list.go
+++ b/internal/manifest/list.go
@@ -68,6 +68,7 @@ type ListUpdate struct {
 	Digest    digest.Digest
 	Size      int64
 	MediaType string
+	Platform  *imgspecv1.Platform // read-only field: may be set by Instance(), ignored by UpdateInstance()
 }
 
 type ListOp int
--- a/internal/manifest/oci_index.go
+++ b/internal/manifest/oci_index.go
@@ -57,6 +57,7 @@ func (index *OCI1IndexPublic) Instance(instanceDigest digest.Digest) (ListUpdate
 				Digest:    manifest.Digest,
 				Size:      manifest.Size,
 				MediaType: manifest.MediaType,
+				Platform:  manifest.Platform,
 			}, nil
 		}
 	}
--- a/internal/manifest/docker_schema2_list.go
+++ b/internal/manifest/docker_schema2_list.go
@@ -61,6 +61,13 @@ func (list *Schema2ListPublic) Instance(instanceDigest digest.Digest) (ListUpdat
 				Digest:    manifest.Digest,
 				Size:      manifest.Size,
 				MediaType: manifest.MediaType,
+				Platform: &imgspecv1.Platform{
+					OS:           manifest.Platform.OS,
+					Architecture: manifest.Platform.Architecture,
+					OSVersion:    manifest.Platform.OSVersion,
+					OSFeatures:   manifest.Platform.OSFeatures,
+					Variant:      manifest.Platform.Variant,
+				},
 			}, nil
 		}
 	}
--- a/internal/manifest/docker_schema2_list_test.go
+++ b/internal/manifest/docker_schema2_list_test.go
@@ -55,6 +55,8 @@ func TestSchema2ListEditInstances(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, "something", instance.MediaType)
 	assert.Equal(t, int64(32), instance.Size)
+	// platform must match with instance platform set in `v2list.manifest.json` for the first instance
+	assert.Equal(t, &imgspecv1.Platform{Architecture: "ppc64le", OS: "linux", OSVersion: "", OSFeatures: []string(nil), Variant: ""}, instance.Platform)
 
 	// Create a fresh list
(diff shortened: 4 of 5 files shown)


flouthoc:
@mtrmac @vrothberg @giuseppe Just a small PR to include `Platform` field as well.

internal/manifest/list.go
```diff
@@ -68,6 +68,7 @@ type ListUpdate struct {
 	Digest    digest.Digest
 	Size      int64
 	MediaType string
+	Platform  imgspecv1.Platform
```
mtrmac, line 71: > includes the fields _which a List's UpdateInstances() method will modify_.

I don’t see any code added to deal with modification.

That’s… probably just as well, because adding a field that will cause `Platform` to be cleared by older callers of `.UpdateInstance` would be a compatibility break.

But we need to somehow deal with _both_ the need to add the new feature, and the requirement not to break any existing API caller.

---

One clean way would be to not change the `ListPublic` interface at all, and expose all of this in a new way.

`c/image/manifest.ListGetInstanceInfo(List[Public], digest.Digest) (manifest.SomeNewStructWithNewFields)`

OTOH that’s tedious to implement.

Probably simpler would be to do the what this PR does, just add a new field to the struct, but somehow clearly demarcate that it is inspect-only. `ListUpdate.ReadOnly.Platform` ?? That’s fairly ugly but plausible.

mtrmac: Oops… I’m sorry, for some reason I thought the intended caller is external, Podman or so.

If this is for the internal c/image/copy caller, we don’t need to change `ListPublic` at all: add a new `List.Convert` (just like `ConvertToMIMEType`, but returns at the internal type), and a new `List.InstanceData()` returning any fields we like.

… but again, that’s tedious and just adding a “read-only” field to the public API would be simpler to write _code_ for. At the cost of having to write a much more careful documentation… because any pre-existing external callers must not be broken, and any pre-existing external implementers of `.Instance()` must not be broken either.

flouthoc: > … but again, that’s tedious and just adding a “read-only” field to the public API would be simpler to write code for. At the cost of having to write a much more careful documentation… because any pre-existing external callers must not be broken, and any pre-existing external implementers of .Instance() must not be broken either.

I have made this field private with a comment of `read-only` and created a public Get API for this.

mtrmac: I don’t see that a method on `ListUpdate` can help here:
- Anyone can call the added method (not just packages allowed to import `c/image/internal/manifest`). So this does not hide the read side.
- OTOH the external callers can’t set the value. So to the extent we need to worry about external implementations of `List`, they are now worse off because they can’t keep up with callers.
- The “fields which a List's UpdateInstances() method will modify.” comment is still incorrect.

flouthoc: I agree, how about just adding a comment to the public field itself ? Did you mean this in https://github.com/containers/image/pull/2029#discussion_r1255165714

internal/manifest/list.go
```diff
@@ -68,6 +68,7 @@ type ListUpdate struct {
 	Digest    digest.Digest
 	Size      int64
 	MediaType string
+	Platform  imgspecv1.Platform // read-only field, will be pre-set by Instance()
```
mtrmac, line 71:
```suggestion
	Platform  imgspecv1.Platform // read-only field: may be set by Instance(), ignored by UpdateInstance()
```

internal/manifest/oci_index.go
```diff
@@ -57,6 +57,7 @@ func (index *OCI1IndexPublic) Instance(instanceDigest digest.Digest) (ListUpdate
 				Digest:    manifest.Digest,
 				Size:      manifest.Size,
 				MediaType: manifest.MediaType,
+				Platform:  *manifest.Platform,
```
mtrmac, line 60: The platform can be missing (`nil`) in OCI; so the new filed must be able to express that (i.e. also be a `*….Platform`)

internal/manifest/list.go
mtrmac: Yes, I think just documenting that is practical enough…

internal/manifest/oci_index.go
flouthoc: Changed field to `Platform  *imgspecv1.Platform` , that should support `nil`

internal/manifest/oci_index.go
```diff
@@ -57,6 +57,7 @@ func (index *OCI1IndexPublic) Instance(instanceDigest digest.Digest) (ListUpdate
 				Digest:    manifest.Digest,
 				Size:      manifest.Size,
 				MediaType: manifest.MediaType,
+				Platform:  manifest.Platform,
```
mtrmac, line 60: It would be a bit safer to make a clone instead of providing a pointer to our data, but, meh, we can deal with that if it breaks I guess.

mtrmac (approved):
Thanks! LGTM.

PR #2034, opened 2023-07-09
pkg/docker: use the same default auth path as macOS on FreeBSD

dfr:
This was defaulting to "/run/containers/\<uid\>/auth.json" unless the OS was "darwin" or "windows". It makes more sense to use "~/.config/containers/auth.json" instead which removes the need for a "/run" directory on the host.

### DIFF
--- a/pkg/docker/config/config_test.go
+++ b/pkg/docker/config/config_test.go
@@ -17,6 +17,7 @@ import (
 func TestGetPathToAuth(t *testing.T) {
 	const linux = "linux"
 	const darwin = "darwin"
+	const freebsd = "freebsd"
 
 	uid := fmt.Sprintf("%d", os.Getuid())
 	// We don’t have to override the home directory for this because use of this path does not depend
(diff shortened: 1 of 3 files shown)


pkg/docker/config/config_test.go
```diff
@@ -36,20 +37,27 @@ func TestGetPathToAuth(t *testing.T) {
...
 		// SystemContext overrides
 		{&types.SystemContext{AuthFilePath: "/absolute/path"}, linux, "", "/absolute/path", false, true},
 		{&types.SystemContext{AuthFilePath: "/absolute/path"}, darwin, "", "/absolute/path", false, true},
+		{&types.SystemContext{AuthFilePath: "/absolute/path"}, freebsd, "", "/absolute/path", false, true},
 		{&types.SystemContext{LegacyFormatAuthFilePath: "/absolute/path"}, linux, "", "/absolute/path", true, true},
 		{&types.SystemContext{LegacyFormatAuthFilePath: "/absolute/path"}, darwin, "", "/absolute/path", true, true},
+		{&types.SystemContext{LegacyFormatAuthFilePath: "/absolute/path"}, freebsd, "", "/absolute/path", true, true},
 		{&types.SystemContext{RootForImplicitAbsolutePaths: "/prefix"}, linux, "", "/prefix/run/containers/" + uid + "/auth.json", false, false},
 		{&types.SystemContext{RootForImplicitAbsolutePaths: "/prefix"}, darwin, "", "/prefix/run/containers/" + uid + "/auth.json", false, false},
+		{&types.SystemContext{RootForImplicitAbsolutePaths: "/prefix"}, freebsd, "", "/prefix/run/containers/" + uid + "/auth.json", false, false},
```
mtrmac, line 53: Actually, I think these should be just `darwinDefault`
- `RootFor…` is ~chroot-like, it should not cause a switch to Linux semantics
- `RootFor…` is documented not to affect paths in `$HOME`.

(I guess that feature is extremely rarely, if ever, used, so no-one has noticed in all those years.)

pkg/docker/config/config_test.go
```diff
@@ -88,7 +96,7 @@ func TestGetAuth(t *testing.T) {
 	t.Logf("using temporary home directory: %q", tmpHomeDir)
 
 	configDir1 := filepath.Join(tmpXDGRuntimeDir, "containers")
-	if runtime.GOOS == "windows" || runtime.GOOS == "darwin" {
+	if runtime.GOOS == "windows" || runtime.GOOS == "darwin" || runtime.GOOS == "freebsd" {
```
mtrmac, line 99: Shouldn’t this be `!= "linux"`, then, to match the code being tested? (In both cases.)

mtrmac (commented):
Thanks!

This is a compatibility break, but I think it’s up to you whether that’s that’s acceptable for FreeBSD users.

dfr:
> Thanks!
> 
> This is a compatibility break, but I think it’s up to you whether that’s that’s acceptable for FreeBSD users.

I was planning to address that in the FreeBSD packaing so that upgrading the podman/buildah/skopeo package migrates root's auth.json file to the new location, probably combined with a reminder to upgrade all three together.

mtrmac (approved):
Thanks again!

PR #2037, opened 2023-07-12
Fix tests of the ostree transport

mtrmac:
... which I broke on 2022-03-16, and noone noticed.

(The actual non-test code was, at least east as of today, buildable, so this does not prove that noone is using the code... but it's a hint in that direction.)

### DIFF
--- a/ostree/ostree_dest_test.go
+++ b/ostree/ostree_dest_test.go
@@ -3,4 +3,8 @@
 
 package ostree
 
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

flouthoc (approved):
LGTM

vrothberg (approved):
LGTM

PR #2039, opened 2023-07-13
helpers_test,cleanup: correct argument order

flouthoc:
Fixes: https://github.com/containers/image/pull/2023#discussion_r1261581159.

`assert.Equal` expects `assert.Equal(t, expected, actual)` instead of `assert.Equal(t, actual, expected)`.

Ref:https://pkg.go.dev/github.com/stretchr/testify/assert#Assertions.Equal


flouthoc:
@vrothberg PTAL

vrothberg (approved):
LGTM

PR #2040, opened 2023-07-14
listupdate,oci: instance show read-only `annotations`

flouthoc:
There is a need to read annotations of a particular instance to get its compression. Expose `Annotations` as a read-only field.

Needed By: https://github.com/containers/image/pull/1987


flouthoc:
@mtrmac @vrothberg PTAL

mtrmac (commented):
For #1987 : the `manifest` format handlers implement mapping algorithms to annotations, so they should also implement mapping annotations to algorithms.

I.e. add a `CompressionAlgorithms` field instead. Or rather, maybe `CompressionAlgorithmNames` instead, so that ` internal/manifest` does not depend on `pkg/compression.AlgorithmByName` and does not bring all of c/storage/pkg/chunked/compressor to every metadata-only user.

(Adding annotations _as well_ wouldn’t hurt but I think isn‘t immediately relevant.)

internal/manifest/list.go
```diff
@@ -65,10 +65,12 @@ type List interface {
...
-	Digest    digest.Digest
-	Size      int64
-	MediaType string
-	Platform  *imgspecv1.Platform // read-only field: may be set by Instance(), ignored by UpdateInstance()
+	Digest                    digest.Digest
+	Size                      int64
+	MediaType                 string
+	Platform                  *imgspecv1.Platform // read-only field: may be set by Instance(), ignored by UpdateInstance()
+	Annotations               map[string]string   // read-only field: may be set by Instance(), ignored by UpdateInstance()
+	CompressionAlgorithmNames []string            // read-only field: may be set by Instance(), ignored by UpdateInstance()
```
mtrmac, line 73: The schema2 implementation should return `[gzip]` here, so that a copy of a schema2 multi-arch image, converting to OCI and adding the Zstd variants, works consistently with a copy of a gzip OCI image.

internal/manifest/oci_index.go
```diff
@@ -79,6 +82,20 @@ func (index *OCI1IndexPublic) UpdateInstances(updates []ListUpdate) error {
 	return index.editInstances(editInstances)
 }
 
+func annotationsToCompressionAlgorithmNames(annotations map[string]string) []string {
+	result := []string{}
+	for k, v := range annotations {
+		if k == OCI1InstanceAnnotationCompressionZSTD && v == OCI1InstanceAnnotationCompressionZSTDValue {
```
mtrmac, line 88: This is just `if annotations[…Zstd] == …ZstdValue`.

flouthoc: Done.

internal/manifest/list.go
flouthoc: Done.

internal/manifest/list.go
```diff
@@ -65,10 +65,12 @@ type List interface {
...
 // This is publicly visible as c/image/manifest.ListUpdate.
 type ListUpdate struct {
-	Digest    digest.Digest
-	Size      int64
-	MediaType string
-	Platform  *imgspecv1.Platform // read-only field: may be set by Instance(), ignored by UpdateInstance()
+	Digest                    digest.Digest
+	Size                      int64
+	MediaType                 string
+	Platform                  *imgspecv1.Platform // read-only field: may be set by Instance(), ignored by UpdateInstance()
```
vrothberg, line 71: Consider moving the read-only fields into an inner struct:
```golang
type ListUpdate struct {
    [...]
    struct ReadOnly {
        Platform string
        [...]
    }
}
```

This way, all users have to access it via `(ListUpdate).ReadOnly.Platform` which is quite expressive.

internal/manifest/oci_index.go
```diff
@@ -79,6 +82,18 @@ func (index *OCI1IndexPublic) UpdateInstances(updates []ListUpdate) error {
 	return index.editInstances(editInstances)
 }
 
+func annotationsToCompressionAlgorithmNames(annotations map[string]string) []string {
+	result := []string{}
```
vrothberg, line 86: Nit: pre-allocate the slice with size 1.

internal/manifest/oci_index_test.go
```diff
@@ -132,6 +134,12 @@ func TestOCI1EditInstances(t *testing.T) {
 	// Zstd should be kept on lowest priority as compared to the default gzip ones and order of prior elements must be preserved.
 	assert.Equal(t, list.Instances(), []digest.Digest{digest.Digest("sha256:e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7fc331f"), digest.Digest("sha256:5b0bcabd1ed22e9fb1310cf6c2dec7cdef19f0ad69efa1f392e94a4333501270"), digest.Digest("sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"), digest.Digest("sha256:cccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc"), digest.Digest("sha256:eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee"), digest.Digest("sha256:hhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh"), digest.Digest("sha256:ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff")})
 
+	// Verfiy compression of an instance added to list
```
vrothberg, line 137: Can you elaborate in the comment _why_ both annotations are expected?

flouthoc: Done.

internal/manifest/oci_index.go
flouthoc: Done.

internal/manifest/list.go
flouthoc: Nice Idea, Done

vrothberg (approved):
LGTM

internal/manifest/oci_index.go
```diff
@@ -79,6 +83,18 @@ func (index *OCI1IndexPublic) UpdateInstances(updates []ListUpdate) error {
 	return index.editInstances(editInstances)
 }
 
+func annotationsToCompressionAlgorithmNames(annotations map[string]string) []string {
+	result := make([]string, 0)
```
mtrmac, line 87:
```suggestion
	result := make([]string, 0, 1)
```
to actually preallocate

mtrmac (commented):
One last nit please

mtrmac (approved):
Thanks!

flouthoc:
@mtrmac Thanks for the rebase :) ready for merge

PR #2043, opened 2023-07-15
Fix TestOCI1IndexChooseInstanceByCompression on non-amd64

mtrmac:
Setting `SystemContext.ArchitectureChoice` to `""` does not mean "match any/the first platform"; it's the default behavior of `SystemContext`, and it means "choose for the current runtime architecture". (Originally discussed in https://github.com/containers/image/pull/1789#discussion_r1115647449 )

I.e. on `amd64` these two test cases are redundant with the first two instances above, and on other architectures (notably ARM) they cause failures.

So just drop them.


vrothberg (approved):
LGTM

TomSweeneyRedHat:
LGTM

PR #2045, opened 2023-07-16
Make temporary names container/image specific

rhatdan:
If you hit Ctr-C while pulling an image files and directories get left in /var/tmp. By adding "containers_images" prefix, we can use systemd tmpfiles handling to remove them on reboot safely.

Help to make https://github.com/containers/podman/pull/19201 safer.

### DIFF
--- a/docker/internal/tarfile/reader.go
+++ b/docker/internal/tarfile/reader.go
@@ -57,7 +57,7 @@ func NewReaderFromFile(sys *types.SystemContext, path string) (*Reader, error) {
 // The caller should call .Close() on the returned archive when done.
 func NewReaderFromStream(sys *types.SystemContext, inputStream io.Reader) (*Reader, error) {
 	// Save inputStream to a temporary file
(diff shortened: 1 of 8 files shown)


vrothberg (commented):
Idea SGTM

Build needs massaging:
> oci/archive/oci_transport.go:163:40: cannot use dir (variable of type *os.File) as string value in argument to ocilayout.NewReference
oci/archive/oci_transport.go:168:45: cannot use dir (variable of type *os.File) as string value in struct literal

mtrmac (commented):
Thanks, this makes sense to me.

---

Separately, it could be interesting for Podman or callers to turn Ctrl-C into a context cancellation, to trigger a correct cleanup. OTOH that would be a major change… and IIRC the typical `io.Copy` (not involving HTTP) is not cancellable, so slow on-disk operations would not stop immediately.

mtrmac (approved):
Thanks! LGTM.

PR #2047, opened 2023-07-18
copy/multiple: instanceCopyCopy honor `UpdateCompressionAlgorithms`

flouthoc:
Needed By: https://github.com/containers/skopeo/pull/2047

* Users can set compression formats while performing `copy` in such cases
`instanceCopyCopy` must honor UpdateCompressionAlgorithms.

* copy/single: set requiredCompression if compressionFormat is changed
 - Modifies `copy/single` to correctly use the feature from
https://github.com/containers/image/pull/2023, that is if compression is
changed blob must be resued only if it matches required compression.

* oci_index: init Annotations if nil and required
 - If `UpdateCompressionAlgorithms` is set then `Annotations` map must be
initialized if its not set.

Integration tests for `e2e` feature added here: https://github.com/containers/skopeo/pull/2047#issue-1809369600

### DIFF
--- a/copy/multiple.go
+++ b/copy/multiple.go
@@ -134,11 +134,12 @@ func (c *copier) copyMultipleImages(ctx context.Context) (copiedManifest []byte,
 			}
 			// Record the result of a possible conversion here.
 			instanceEdits = append(instanceEdits, internalManifest.ListEdit{
-				ListOperation:   internalManifest.ListOpUpdate,
-				UpdateOldDigest: instance.sourceDigest,
-				UpdateDigest:    updated.manifestDigest,
-				UpdateSize:      int64(len(updated.manifest)),
-				UpdateMediaType: updated.manifestMIMEType})
+				ListOperation:               internalManifest.ListOpUpdate,
+				UpdateOldDigest:             instance.sourceDigest,
+				UpdateDigest:                updated.manifestDigest,
+				UpdateSize:                  int64(len(updated.manifest)),
+				UpdateCompressionAlgorithms: updated.compressionAlgorithms,
+				UpdateMediaType:             updated.manifestMIMEType})
 		default:
 			return nil, fmt.Errorf("copying image: invalid copy operation %d", instance.op)
 		}
--- a/copy/multiple.go
+++ b/copy/multiple.go
@@ -128,7 +128,7 @@ func (c *copier) copyMultipleImages(ctx context.Context) (copiedManifest []byte,
 			logrus.Debugf("Copying instance %s (%d/%d)", instance.sourceDigest, i+1, len(instanceCopyList))
 			c.Printf("Copying image %s (%d/%d)\n", instance.sourceDigest, i+1, len(instanceCopyList))
 			unparsedInstance := image.UnparsedInstance(c.rawSource, &instanceCopyList[i].sourceDigest)
-			updated, err := c.copySingleImage(ctx, unparsedInstance, &instanceCopyList[i].sourceDigest)
+			updated, err := c.copySingleImage(ctx, unparsedInstance, &instanceCopyList[i].sourceDigest, false)
 			if err != nil {
 				return nil, fmt.Errorf("copying image %d/%d from manifest list: %w", i+1, len(instanceCopyList), err)
 			}
--- a/copy/single.go
+++ b/copy/single.go
@@ -49,7 +49,7 @@ type copySingleImageResult struct {
 
 // copySingleImage copies a single (non-manifest-list) image unparsedImage, using c.policyContext to validate
 // source image admissibility.
-func (c *copier) copySingleImage(ctx context.Context, unparsedImage *image.UnparsedImage, targetInstance *digest.Digest) (copySingleImageResult, error) {
+func (c *copier) copySingleImage(ctx context.Context, unparsedImage *image.UnparsedImage, targetInstance *digest.Digest, requireCompressionFormatMatch bool) (copySingleImageResult, error) {
 	// The caller is handling manifest lists; this could happen only if a manifest list contains a manifest list.
 	// Make sure we fail cleanly in such cases.
 	multiImage, err := isMultiImage(ctx, unparsedImage)
@@ -129,7 +129,8 @@ func (c *copier) copySingleImage(ctx context.Context, unparsedImage *image.Unpar
 		manifestUpdates: &types.ManifestUpdateOptions{InformationOnly: types.ManifestUpdateInformation{Destination: c.dest}},
 		src:             src,
 		// diffIDsAreNeeded is computed later
-		cannotModifyManifestReason: cannotModifyManifestReason,
+		cannotModifyManifestReason:    cannotModifyManifestReason,
+		requireCompressionFormatMatch: requireCompressionFormatMatch,
 	}
 	if c.options.DestinationCtx != nil {
 		// Note that compressionFormat and compressionLevel can be nil.
--- a/copy/single.go
+++ b/copy/single.go
@@ -29,22 +29,23 @@ import (
 
 // imageCopier tracks state specific to a single image (possibly an item of a manifest list)
 type imageCopier struct {
-	c                          *copier
-	manifestUpdates            *types.ManifestUpdateOptions
-	src                        *image.SourcedImage
-	diffIDsAreNeeded           bool
-	cannotModifyManifestReason string // The reason the manifest cannot be modified, or an empty string if it can
-	canSubstituteBlobs         bool
-	compressionFormat          *compressiontypes.Algorithm // Compression algorithm to use, if the user explicitly requested one, or nil.
-	compressionLevel           *int
+	c                             *copier
+	manifestUpdates               *types.ManifestUpdateOptions
+	src                           *image.SourcedImage
+	diffIDsAreNeeded              bool
+	cannotModifyManifestReason    string // The reason the manifest cannot be modified, or an empty string if it can
+	canSubstituteBlobs            bool
+	compressionFormat             *compressiontypes.Algorithm // Compression algorithm to use, if the user explicitly requested one, or nil.
+	compressionLevel              *int
 	requireCompressionFormatMatch bool
 }
 
 // copySingleImageResult carries data produced by copySingleImage
 type copySingleImageResult struct {
-	manifest         []byte
-	manifestMIMEType string
-	manifestDigest   digest.Digest
+	manifest              []byte
+	manifestMIMEType      string
+	manifestDigest        digest.Digest
+	compressionAlgorithms []compressiontypes.Algorithm
 }
 
 // copySingleImage copies a single (non-manifest-list) image unparsedImage, using c.policyContext to validate
@@ -194,7 +195,8 @@ func (c *copier) copySingleImage(ctx context.Context, unparsedImage *image.Unpar
 		}
 	}
 
-	if err := ic.copyLayers(ctx); err != nil {
+	compressionAlgos, err := ic.copyLayers(ctx)
+	if err != nil {
 		return copySingleImageResult{}, err
 	}
 
@@ -274,7 +276,7 @@ func (c *copier) copySingleImage(ctx context.Context, unparsedImage *image.Unpar
 			return copySingleImageResult{}, fmt.Errorf("writing signatures: %w", err)
 		}
 	}
-
+	wipResult.compressionAlgorithms = compressionAlgos
 	res := wipResult // We are done
 	return res, nil
 }
@@ -367,26 +369,38 @@ func (ic *imageCopier) compareImageDestinationManifestEqual(ctx context.Context,
 		return nil, nil
 	}
 
+	compressionAlgos := set.New[string]()
+	for _, srcInfo := range ic.src.LayerInfos() {
+		compression := compressionAlgorithmFromMIMEType(srcInfo)
+		compressionAlgos.Add(compression.Name())
+	}
+
+	algos, err := algorithmsByNames(compressionAlgos.Values())
+	if err != nil {
+		return nil, err
+	}
+
 	// Destination and source manifests, types and digests should all be equivalent
 	return &copySingleImageResult{
-		manifest:         destManifest,
-		manifestMIMEType: destManifestType,
-		manifestDigest:   srcManifestDigest,
+		manifest:              destManifest,
+		manifestMIMEType:      destManifestType,
+		manifestDigest:        srcManifestDigest,
+		compressionAlgorithms: algos,
 	}, nil
 }
 
 // copyLayers copies layers from ic.src/ic.c.rawSource to dest, using and updating ic.manifestUpdates if necessary and ic.cannotModifyManifestReason == "".
-func (ic *imageCopier) copyLayers(ctx context.Context) error {
+func (ic *imageCopier) copyLayers(ctx context.Context) ([]compressiontypes.Algorithm, error) {
 	srcInfos := ic.src.LayerInfos()
 	numLayers := len(srcInfos)
 	updatedSrcInfos, err := ic.src.LayerInfosForCopy(ctx)
 	if err != nil {
-		return err
+		return nil, err
 	}
 	srcInfosUpdated := false
 	if updatedSrcInfos != nil && !reflect.DeepEqual(srcInfos, updatedSrcInfos) {
 		if ic.cannotModifyManifestReason != "" {
-			return fmt.Errorf("Copying this image would require changing layer representation, which we cannot do: %q", ic.cannotModifyManifestReason)
+			return nil, fmt.Errorf("Copying this image would require changing layer representation, which we cannot do: %q", ic.cannotModifyManifestReason)
 		}
 		srcInfos = updatedSrcInfos
 		srcInfosUpdated = true
@@ -402,7 +416,7 @@ func (ic *imageCopier) copyLayers(ctx context.Context) error {
 	// layer is empty.
 	man, err := manifest.FromBlob(ic.src.ManifestBlob, ic.src.ManifestMIMEType)
 	if err != nil {
-		return err
+		return nil, err
 	}
 	manifestLayerInfos := man.LayerInfos()
 
@@ -468,14 +482,18 @@ func (ic *imageCopier) copyLayers(ctx context.Context) error {
 		// A call to copyGroup.Wait() is done at this point by the defer above.
 		return nil
 	}(); err != nil {
-		return err
+		return nil, err
 	}
 
+	compressionAlgos := set.New[string]()
 	destInfos := make([]types.BlobInfo, numLayers)
 	diffIDs := make([]digest.Digest, numLayers)
 	for i, cld := range data {
 		if cld.err != nil {
-			return cld.err
+			return nil, cld.err
+		}
+		if cld.destInfo.CompressionAlgorithm != nil {
+			compressionAlgos.Add(cld.destInfo.CompressionAlgorithm.Name())
 		}
 		destInfos[i] = cld.destInfo
 		diffIDs[i] = cld.diffID
@@ -490,7 +508,11 @@ func (ic *imageCopier) copyLayers(ctx context.Context) error {
 	if srcInfosUpdated || layerDigestsDiffer(srcInfos, destInfos) {
 		ic.manifestUpdates.LayerInfos = destInfos
 	}
-	return nil
+	algos, err := algorithmsByNames(compressionAlgos.Values())
+	if err != nil {
+		return nil, err
+	}
+	return algos, nil
 }
 
 // layerDigestsDiffer returns true iff the digests in a and b differ (ignoring sizes and possible other fields)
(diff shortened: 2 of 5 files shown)


flouthoc:
@mtrmac PTAL

mtrmac (commented):
Consider a simple `skopeo copy --format oci` of an existing image: In that case we should preserve the original blobs (or reuse blobs at destination, if possible), and add the OCI annotations.

So while `copy --dest-compress-format zstd` is the most salient use case, I think this should work by collecting compression algorithms from all copies.

- IIRC the `BlobInfo` returned from `copyLayer` always contains an algorithm, or `Unknown`/`Uncompressed.
- Then, `copyLayers` can deduplicate the values into a set; and return that value.
- `copySingleImage` can return the relevant algorithms (optionally add a `copySingleImageResult struct` so that we don’t have that many unnamed return values)
- and that can then be used for `ListEdit`

mtrmac:
> Consider a simple `skopeo copy --format oci` of an existing image: In that case we should preserve the original blobs (or reuse blobs at destination, if possible), and add the OCI annotations.

(If the source is not already OCI, it is probably v2s1 or v2s2, which only support gzip, so that _seems_ like a non-issue. But the copy can reuse zstd blobs at the destination, creating an OCI image that is mixed gzip + Zstd, or possibly Zstd-only.)

mtrmac (commented):
Also I think the shortcut in `OptimizeDestinationImageAlreadyExists` must not be taken if `require…Match`.

mtrmac:
Ugh, the `OptimizeDestinationImageAlreadyExists` path really is a pain… When we do a full copy, we do get algorithm data from all blobs as a side-effect. When we skip that, all of that is missing and needs to be obtained elsewhere. Options I can see:
1. Teach `internal/image.SourcedImage` (and `manifest` implementations) to return a list of algorithms from a manifest. That’s quite possible (we have `compressionMIMETypeSet` tables already), probably cleanest, but it’s another new API addition, and the data we would get that way comes from a different source than from ordinary copies.
2. Take the `list.Instance().ReadOnly.CompressionAlgorithms` data, which was recently computed, pass it from the outside down to `copySingleImage`, and have `copySingleImage` return that to the caller on the `Optimize…` path. Technically that would work, except that the value would not be available in the non-multi-image callers of `copySingleImage`, i.e. that would introduce an inconsistency in behavior.
3. Take the conceptually-misplaced heuristic determining `CompressionAlgorithm` at the start of `copyLayer`; extract it to a separate helper function; and call that in `compareImageDestinationManifestEqual` to build the data. That way the algorithm values we return is determined exactly the same way on the `Optimize…` path and on the other one, and always available.

I think 3. is the cleanest way to do this right now; and if we ever replace the “conceptually-misplaced heuristic” (which serves as a nice example that technical debt never goes away on its own) with a proper `manifest` API addition, that would naturally handle both code paths.

internal/manifest/oci_index.go
```diff
@@ -146,6 +146,9 @@ func (index *OCI1IndexPublic) editInstances(editInstances []ListEdit) error {
 					maps.Copy(index.Manifests[targetIndex].Annotations, editInstance.UpdateAnnotations)
 				}
 			}
+			if index.Manifests[targetIndex].Annotations == nil && len(editInstance.UpdateCompressionAlgorithms) > 0 {
```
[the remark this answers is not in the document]
flouthoc: moving this to `addCompressionAnnotations` would require the function to accept `Manifest` directly cause sending the value of `index.Manifests[targetIndex].Annotations` will not work if its `nil`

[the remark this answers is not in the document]
mtrmac: I thought sending a `*map…` would work — but I didn’t test that.

flouthoc:
@mtrmac I have not updated tests yet but this is what I am thinking. WDYT ?

copy/multiple.go
```diff
@@ -27,23 +27,35 @@ const (
...
+		instanceDetails, err := list.Instance(instanceDigest)
+		if err != nil {
+			return res, err
+		}
+		requireCompressionFormatMatch := false
+		if options.DestinationCtx != nil && options.DestinationCtx.CompressionFormat != nil {
+			if !slices.Contains(instanceDetails.ReadOnly.CompressionAlgorithmNames, options.DestinationCtx.CompressionFormat.Name()) {
+				requireCompressionFormatMatch = true
+			}
+		}
```
mtrmac, line 51: All of this about sending `require…Match` down to `copySingleImage` is relevant for the larger work, but not necessary for this PR, unless I’m missing something.

(And per #1987 , I default to thinking that it should be set only on the clone copies, but I’m not sure about that. … and there’s also the option for this PR to just build the infrastructure in `copySingleImage`, always setting it to `false`, and for #1987 to actually add the logic that sets it.)

---

If you intentionally want to make this PR add the two things (setting “update algorithms” correctly, and adding the `require…Match` infrastructure), that’s also a choice that can be made, just let me know whether you prefer this to do one or both.

copy/single.go
```diff
@@ -29,26 +29,41 @@ import (
...
 
 // copySingleImageResult carries data produced by copySingleImage
 type copySingleImageResult struct {
 	manifest         []byte
 	manifestMIMEType string
 	manifestDigest   digest.Digest
+	compressionAlgorithms  []compressiontypes.Algorithm
 }
 
+func removeDuplicateCompression(algos []compressiontypes.Algorithm) []compressiontypes.Algorithm {
```
mtrmac, line 51: I think this can be built on `internal/set.Set` instead of manually.

copy/single.go
```diff
@@ -29,26 +29,41 @@ import (
...
+        }
+    }
+    return list
+}
+
+
 // copySingleImage copies a single (non-manifest-list) image unparsedImage, using c.policyContext to validate
 // source image admissibility.
-func (c *copier) copySingleImage(ctx context.Context, unparsedImage *image.UnparsedImage, targetInstance *digest.Digest) (copySingleImageResult, error) {
+func (c *copier) copySingleImage(ctx context.Context, unparsedImage *image.UnparsedImage, targetInstance *digest.Digest, requireCompressionFormatMatch bool) (copySingleImageResult, error) {
```
mtrmac, line 66: (In the PR that adds the new parameter) it might be worthwhile to add a `copySingleImageOptions struct`
, primarily so that the boolean is named. OTOH it’s just one boolean right now, and `copyBlob…` is much worse, so blocking on that wouldn’t be fair.

copy/single.go
```diff
@@ -173,7 +189,7 @@ func (c *copier) copySingleImage(ctx context.Context, unparsedImage *image.Unpar
 	ic.diffIDsAreNeeded = src.UpdatedImageNeedsLayerDiffIDs(*ic.manifestUpdates)
 
 	// If enabled, fetch and compare the destination's manifest. And as an optimization skip updating the destination iff equal
-	if c.options.OptimizeDestinationImageAlreadyExists {
+	if c.options.OptimizeDestinationImageAlreadyExists && !ic.requireCompressionFormatMatch{
```
mtrmac, line 192: Stylistically I’d prefer this to be one of the checks inside this top-level `if`, so that the “checking if we can skip copying” line is logged (and that should log the new value as well).

copy/single.go
```diff
@@ -192,9 +208,11 @@ func (c *copier) copySingleImage(ctx context.Context, unparsedImage *image.Unpar
 		}
 	}
 
-	if err := ic.copyLayers(ctx); err != nil {
+	compressionAlgos, err := ic.copyLayers(ctx)
+	if err != nil {
 		return copySingleImageResult{}, err
 	}
+	compressionAlgos = removeDuplicateCompression(compressionAlgos)
```
mtrmac, line 215: (Non-blocking: I’m tempted to say `copyLayers` should do this.)

copy/single.go
```diff
@@ -248,6 +267,7 @@ func (c *copier) copySingleImage(ctx context.Context, unparsedImage *image.Unpar
 				manifest:         attemptedManifest,
 				manifestMIMEType: manifestMIMEType,
 				manifestDigest:   attemptedManifestDigest,
+				compressionAlgorithms: compressionAlgos,
```
mtrmac, line 270: I’d probably add it to `wipResult` in one place, either after the loop, or at the very end when forming the final value, to reduce duplication / risk of drift.

copy/single.go
```diff
@@ -365,26 +385,34 @@ func (ic *imageCopier) compareImageDestinationManifestEqual(ctx context.Context,
...
 // copyLayers copies layers from ic.src/ic.c.rawSource to dest, using and updating ic.manifestUpdates if necessary and ic.cannotModifyManifestReason == "".
-func (ic *imageCopier) copyLayers(ctx context.Context) error {
+func (ic *imageCopier) copyLayers(ctx context.Context) ([]compressiontypes.Algorithm, error) {
+	res := []compressiontypes.Algorithm{}
 	srcInfos := ic.src.LayerInfos()
 	numLayers := len(srcInfos)
 	updatedSrcInfos, err := ic.src.LayerInfosForCopy(ctx)
 	if err != nil {
-		return err
+		return res, err
```
mtrmac, line 410: Non-blocking: I’d prefer a `return nil, …` on error paths, so that there is no data dependency and no ambiguity about what we return.

copy/single.go
```diff
@@ -365,26 +385,34 @@ func (ic *imageCopier) compareImageDestinationManifestEqual(ctx context.Context,
...
 		manifestMIMEType: destManifestType,
 		manifestDigest:   srcManifestDigest,
+		compressionAlgorithms: removeDuplicateCompression(compressionAlgos),
 	}, nil
 }
 
 // copyLayers copies layers from ic.src/ic.c.rawSource to dest, using and updating ic.manifestUpdates if necessary and ic.cannotModifyManifestReason == "".
-func (ic *imageCopier) copyLayers(ctx context.Context) error {
+func (ic *imageCopier) copyLayers(ctx context.Context) ([]compressiontypes.Algorithm, error) {
+	res := []compressiontypes.Algorithm{}
```
mtrmac, line 405: Absolutely non-blocking: `res` might be too short a name for such a long function`. `compressionAlgos`? Or maybe define the value much later in the function, so that it has a short scope.

copy/single.go
```diff
@@ -424,6 +452,7 @@ func (ic *imageCopier) copyLayers(ctx context.Context) error {
 			}
 		} else {
 			cld.destInfo, cld.diffID, cld.err = ic.copyLayer(ctx, srcLayer, toEncrypt, pool, index, srcRef, manifestLayerInfos[index].EmptyLayer)
+			res = append(res, *cld.destInfo.CompressionAlgorithm)
```
mtrmac, line 455: This runs in a concurrently-executing goroutine, so it is invalid to modify a shared `res` variable from here. Instead, pass it through a new field in `copyLayerData`, and collect (+deduplicate?) it when processing the goroutine results.

copy/single.go
```diff
@@ -593,6 +622,20 @@ type diffIDResult struct {
 	err    error
 }
 
+func (ic *imageCopier) getCompressionFromBlob(srcInfo types.BlobInfo) *compressiontypes.Algorithm {
```
mtrmac, line 625: I think this can be a stand-alone function, not a method; that would make it more convenient to potentially unit-test.

copy/single.go
```diff
@@ -593,6 +622,20 @@ type diffIDResult struct {
...
+
+	if srcInfo.CompressionAlgorithm == nil {
```
mtrmac, line 627: If we have that data, we should return it, not ignore it. Or clearly document that the caller needs to handle that case, and actually do that in the new caller.

copy/single.go
```diff
@@ -608,14 +651,7 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, to
 	// This MIME type → compression mapping belongs in manifest-specific code in our manifest
 	// package (but we should preferably replace/change UpdatedImage instead of productizing
 	// this workaround).
```
mtrmac, line 653: Please move this comment to the new function.

mtrmac (commented):
A quick pass, not a careful review — so far all the primary design points look correct.

copy/multiple.go
flouthoc: I think `require..Match` is needed for this case as well, consider the case where registry/infocache already contains blobs in a different compression and I am doing `--dest-format-compress` so afaik skopeo skips pushing blobs with `blob already exists` when `require..match` is not implemented (I assume it is reusing `gzip` blobs candidates). I am able to see this problem.

#### Without `require..match` does not copies new compression
```console
[fl@fedora bin]$ ./skopeo copy --all --format oci --dest-compress-format zstd --src-tls-verify=false --dest-tls-verify=false docker://localhost:5000/list docker://localhost:5000/list
Getting image list signatures
Copying 2 of 2 images in list
Copying image sha256:da39e651fdc155fbd53152dee387b91d4fd1e76da3137ccbd332f98e6c552a63 (1/2)
Copying blob 8a117f3d2e73 skipped: already exists  
Copying blob 5119035169e2 skipped: already exists  
Copying config 58dce7cc03 done   | 
Copying image sha256:f87412510ffdd07383a3b433ab58159522a4ffa3048f3557f55708ec90c8008c (2/2)
Copying blob 33bbc77ac1e2 skipped: already exists  
Copying blob 5119035169e2 skipped: already exists  
Copying config 0044dbde8e done   | 
Writing manifest list to image destination
Storing list signatures
``` 

#### With `require..match` correctly works
```console
[fl@fedora bin]$ ./skopeo copy --all --format oci --dest-compress-format zstd --src-tls-verify=false --dest-tls-verify=false docker://localhost:5000/list docker://localhost:5000/list
Getting image list signatures
Copying 2 of 2 images in list
Copying image sha256:da39e651fdc155fbd53152dee387b91d4fd1e76da3137ccbd332f98e6c552a63 (1/2)
Copying blob 8a117f3d2e73 done   | 
Copying blob 5119035169e2 done   | 
Copying config 58dce7cc03 done   | 
Copying image sha256:f87412510ffdd07383a3b433ab58159522a4ffa3048f3557f55708ec90c8008c (2/2)
Copying blob 33bbc77ac1e2 done   | 
Copying blob 5119035169e2 done   | 
Copying config 0044dbde8e done   | 
Writing manifest list to image destination
Storing list signatures
```

flouthoc: I just added it cause it was constantly causing issues for me while testing locally, I suspect others might also hit this issue while trying to do the same, would it be an issue if we add it here and expose the option on frontend later in a different PR ? I think exposing part would be much simpler then.

copy/single.go
mtrmac: … and probably inlined in the two loops that collect the data.

copy/multiple.go
mtrmac: For the record, repeating from other discussion:
- The underlying `require…Match` functionality is definitely necessary for #1987 .
- I don’t, at this point, think that we should change the current behavior for `--dest-compress-format` (that’s https://github.com/containers/buildah/issues/4613 ); it should be a new option. I don’t feel _very_ strongly about this, and it might be worth discussing, but it also seems unrelated to the primary feature we are working towards at this very moment.

copy/single.go
mtrmac: Looking at more details, only manifest’s `.LayerInfos()` never sets this field; only `pkg/blobcache` sets it in `LayerInfosForCopy`.

So the new caller in `compareImageDestinationManifestEqual` can be build assuming `==nil` above; and in that case this function can a simpler `compressionAlgorithmFromMIMEType`, leaving the `if` only in the `copyLayer` caller.

copy/single.go
```diff
@@ -135,6 +150,7 @@ func (c *copier) copySingleImage(ctx context.Context, unparsedImage *image.Unpar
 		ic.compressionFormat = c.options.DestinationCtx.CompressionFormat
 		ic.compressionLevel = c.options.DestinationCtx.CompressionLevel
 	}
+	ic.requireCompressionFormatMatch = requireCompressionFormatMatch
```
mtrmac, line 153: (Absolutely non-blocking: if it is unconditional like this, the field can be set directly set in the `imageCopier{` constructor above.)

mtrmac (commented):
A full review now — ~nothing new.

copy/single.go
flouthoc: `Set.Add(` ensures that no duplicates are there ( I have added a test to verify that ) however `compression.Algorithm` is not a `comparable` hence I have created a `Set` using `compression.Algorithm.Name()` and its working fine.

Later I'm using `AlgorithmsByNames` to retrieve a slice of  `Algorithms`.

copy/multiple.go
flouthoc: Cool I have just implemented infra for now and passing `false` as default value, will implement this in a different PR.

copy/single.go
flouthoc: > OTOH it’s just one boolean right now, and copyBlob… is much worse, so blocking on that wouldn’t be fair.

Cool I'll implement this later in a different PR.

internal/manifest/oci_index.go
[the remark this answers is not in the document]
flouthoc: done.

internal/set/set_test.go
```diff
@@ -30,6 +30,17 @@ func TestAdd(t *testing.T) {
...
 
+func TestAddMustNotContainDuplicates(t *testing.T) {
+	s := NewWithValues("hello")
+	assert.False(t, s.Contains("world"))
+	s.Add("world")
+	assert.True(t, s.Contains("world"))
+	s.Add("world") // Adding an already-present element
+	assert.True(t, s.Contains("world"))
+	// Must not containe `world` twice.
+	assert.Equal(t, []string{"hello", "world"}, s.Values())
```
mtrmac, line 41: Non-blocking: I think this can be tested in `TestAdd` or `TestValues` (or both). OTOH this works fine, and the test file is small enough.

copy/single.go
```diff
@@ -178,7 +181,7 @@ func (c *copier) copySingleImage(ctx context.Context, unparsedImage *image.Unpar
 		noPendingManifestUpdates := ic.noPendingManifestUpdates()
 
 		logrus.Debugf("Checking if we can skip copying: has signatures=%t, OCI encryption=%t, no manifest updates=%t", shouldUpdateSigs, destRequiresOciEncryption, noPendingManifestUpdates)
-		if !shouldUpdateSigs && !destRequiresOciEncryption && noPendingManifestUpdates {
+		if !shouldUpdateSigs && !destRequiresOciEncryption && noPendingManifestUpdates && !ic.requireCompressionFormatMatch {
```
mtrmac, line 184: Please log this value in the debug log above, so that the decision can be explained.

copy/single.go
```diff
@@ -365,26 +369,39 @@ func (ic *imageCopier) compareImageDestinationManifestEqual(ctx context.Context,
...
+		manifestMIMEType:      destManifestType,
+		manifestDigest:        srcManifestDigest,
+		compressionAlgorithms: algos,
 	}, nil
 }
 
 // copyLayers copies layers from ic.src/ic.c.rawSource to dest, using and updating ic.manifestUpdates if necessary and ic.cannotModifyManifestReason == "".
-func (ic *imageCopier) copyLayers(ctx context.Context) error {
+func (ic *imageCopier) copyLayers(ctx context.Context) ([]compressiontypes.Algorithm, error) {
+	compressionAlgos := set.New[string]()
```
mtrmac, line 394: Absolutely non-blocking: This can be set only just before it’s computed (around `diffIDs := `)

copy/single.go
```diff
@@ -608,13 +629,22 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, to
 	// This MIME type → compression mapping belongs in manifest-specific code in our manifest
```
mtrmac, line 629: The part of the comment above this line should stay in `copyLayer`; it is not relevant to the “image already exists” case.

pkg/compression/compression.go
```diff
@@ -52,6 +52,19 @@ func AlgorithmByName(name string) (Algorithm, error) {
 	return Algorithm{}, fmt.Errorf("cannot find compressor for %q", name)
 }
 
+// AlgorithmsByNames returns slice of Algorithms from slice of Algorithm Names
+func AlgorithmsByNames(names []string) ([]Algorithm, error) {
```
mtrmac, line 56: Non-blocking:

At this point I think this can be an internal helper in `copy/single.go`. Adding more lookup functions to `pkg/compression` is a bit unattractive:

Context: Importing `pkg/compression` adds all compression implementations, including `c/storage/pkg/chunked/compression`, as dependencies, and it’s unclear to me that that’s always a good idea. Compare how `transports` contains a registry, and `transports/alltransports` adds the dependencies: the caller can choose which one to use. (OTOH the argument that each consumer of `pkg/compression` should always support all algorithms might be stronger, I don’t have a firm opinion.)

---

An alternative could be to modify `ListEdit.*CompressionAlgorithms` to only contain names — that would also get rid of the “algorithm name is known but we don’t have an implementation” failure paths — but I think we don’t need to tinker with that now. In practice `compressionAlgorithmFromMIMEType`, or the data from `blob/compression.go`, are always supported algorithms, so that error path is not worth worrying about or making larger changes now.

mtrmac (commented):
Thanks!

Last comments on a debug log and a misplaced comment.

internal/set/set_test.go
flouthoc: done.

copy/single.go
```diff
@@ -466,14 +482,18 @@ func (ic *imageCopier) copyLayers(ctx context.Context) error {
...
 	}
 
+	compressionAlgos := set.New[string]()
 	destInfos := make([]types.BlobInfo, numLayers)
 	diffIDs := make([]digest.Digest, numLayers)
 	for i, cld := range data {
+		if cld.destInfo.CompressionAlgorithm != nil {
+			compressionAlgos.Add(cld.destInfo.CompressionAlgorithm.Name())
+		}
 		if cld.err != nil {
```
mtrmac, line 495: Oops, I apologize for missing this earlier … `cld.destInfo` is only valid if `cld.err` == nil`; so the new code should be added below.

mtrmac (commented):
LGTM otherwise, feel free to merge without another round.

copy/single.go
flouthoc: done.

flouthoc:
I don't have merge access, @containers/image-maintainers PTAL

mtrmac (approved):
Thanks!

PR #2048, opened 2023-07-19
Refactor data passing in c/image/copy

mtrmac:
This is motivated by the concurrent Zstd work, to make it more maintainable to pass data from and to `copySingleImage`.

- Add `copier.options`, `copier.policyContext` and `copier.unparsedToplevel` instead of passing them around in parameters
- Eliminate fields of `copier` and `imageCopier` that are now reachable through `.options`
- Return a `copySingleImageResult` from `copySingleImage`

@flouthoc I realize this is disruptive, and I want to do whatever necessary not to disrupt or slow down your work. Let me know what works for you — feel free to completely ignore this (and I’ll rebase this later as necessary), to reimplement parts yourself in some other way, or to tell me to split this up (maybe only the `copySingleImageResult` part?).

Others: Please don’t merge this without an ACK from @flouthoc .

PR #2050, opened 2023-07-21
copy/single: accept custom `*Options` and wrap arguments in `copySingleImageOptions`

flouthoc:
After https://github.com/containers/image/pull/2048 there is no room for `copy/multiple` to pass custom options to `copySingleImage` while processing each instance, following PR introduces that functionality again and wraps options to simpler struct.

Needed by: https://github.com/containers/image/pull/1987

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -251,7 +251,8 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
 
 	if !multiImage {
 		// The simple case: just copy a single image.
-		single, err := c.copySingleImage(ctx, c.unparsedToplevel, nil, false)
+		singleImageOpts := copySingleImageOptions{requireCompressionFormatMatch: false}
+		single, err := c.copySingleImage(ctx, c.unparsedToplevel, nil, singleImageOpts)
 		if err != nil {
 			return nil, err
 		}
@@ -273,8 +274,8 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
 		}
 		logrus.Debugf("Source is a manifest list; copying (only) instance %s for current system", instanceDigest)
 		unparsedInstance := image.UnparsedInstance(rawSource, &instanceDigest)
-
-		single, err := c.copySingleImage(ctx, unparsedInstance, nil, false)
+		singleImageOpts := copySingleImageOptions{requireCompressionFormatMatch: false}
+		single, err := c.copySingleImage(ctx, unparsedInstance, nil, singleImageOpts)
 		if err != nil {
 			return nil, fmt.Errorf("copying system image from manifest list: %w", err)
 		}
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -251,8 +251,7 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
 
 	if !multiImage {
 		// The simple case: just copy a single image.
-		singleImageOpts := copySingleImageOptions{requireCompressionFormatMatch: false}
-		single, err := c.copySingleImage(ctx, c.unparsedToplevel, nil, singleImageOpts)
+		single, err := c.copySingleImage(ctx, c.unparsedToplevel, nil, copySingleImageOptions{requireCompressionFormatMatch: false})
 		if err != nil {
 			return nil, err
 		}
@@ -274,8 +273,7 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
 		}
 		logrus.Debugf("Source is a manifest list; copying (only) instance %s for current system", instanceDigest)
 		unparsedInstance := image.UnparsedInstance(rawSource, &instanceDigest)
-		singleImageOpts := copySingleImageOptions{requireCompressionFormatMatch: false}
-		single, err := c.copySingleImage(ctx, unparsedInstance, nil, singleImageOpts)
+		single, err := c.copySingleImage(ctx, unparsedInstance, nil, copySingleImageOptions{requireCompressionFormatMatch: false})
 		if err != nil {
 			return nil, fmt.Errorf("copying system image from manifest list: %w", err)
 		}
--- a/copy/single.go
+++ b/copy/single.go
@@ -139,10 +139,13 @@ func (c *copier) copySingleImage(ctx context.Context, unparsedImage *image.Unpar
 		cannotModifyManifestReason:    cannotModifyManifestReason,
 		requireCompressionFormatMatch: opts.requireCompressionFormatMatch,
 	}
-	if c.options.DestinationCtx != nil {
-		// Note that compressionFormat and compressionLevel can be nil.
+	if opts.compressionFormat != nil {
 		ic.compressionFormat = opts.compressionFormat
 		ic.compressionLevel = opts.compressionLevel
+	} else if c.options.DestinationCtx != nil {
+		// Note that compressionFormat and compressionLevel can be nil.
+		ic.compressionFormat = c.options.DestinationCtx.CompressionFormat
+		ic.compressionLevel = c.options.DestinationCtx.CompressionLevel
 	}
 	// Decide whether we can substitute blobs with semantic equivalents:
 	// - Don’t do that if we can’t modify the manifest at all
--- a/copy/single.go
+++ b/copy/single.go
@@ -40,6 +40,10 @@ type imageCopier struct {
 	requireCompressionFormatMatch bool
 }
 
+type copySingleImageOptions struct {
+	requireCompressionFormatMatch bool
+}
+
 // copySingleImageResult carries data produced by copySingleImage
 type copySingleImageResult struct {
 	manifest              []byte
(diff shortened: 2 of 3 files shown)


flouthoc:
@mtrmac @vrothberg PTAL

copy/single.go
```diff
@@ -30,6 +30,7 @@ import (
 // imageCopier tracks state specific to a single image (possibly an item of a manifest list)
 type imageCopier struct {
 	c                             *copier
+	options                       *Options
```
mtrmac, line 33: `copier` already contains an `options` pointer. Why is a separate option struct passed around now?

If `imageCopier` has two sets of options, there should be a very good documentation for when to use which ones.

I _guess_ that two `options` pointers are unnecessary; if `imageCopier` should deviate from the global `options` in some way, make that _deviation_ a part of the new `singleImageOpts`.

copy/single.go
```diff
@@ -50,13 +58,13 @@ type copySingleImageResult struct {
 
 // copySingleImage copies a single (non-manifest-list) image unparsedImage, using c.policyContext to validate
 // source image admissibility.
-func (c *copier) copySingleImage(ctx context.Context, unparsedImage *image.UnparsedImage, targetInstance *digest.Digest, requireCompressionFormatMatch bool) (copySingleImageResult, error) {
+func (c *copier) copySingleImage(ctx context.Context, opt copySingleImageOptions) (copySingleImageResult, error) {
```
mtrmac, line 61: (Absolutely non-blocking, this works fine: personally I’m inclined to say that the critically-important parameters which are “natural” to the operation can remain as top-level, and only the a bit unusual ones need to be named in a struct. ”copy(this image, and here are some details for that)”.

OTOH if we continued to pass `unparsedImage` and `targetInstance` directly, that would probably leave us with a single-field struct right now, which is… mostly an investment for the future.

copy/single.go
```diff
@@ -152,12 +161,12 @@ func (c *copier) copySingleImage(ctx context.Context, unparsedImage *image.Unpar
 		return copySingleImageResult{}, err
 	}
 
-	destRequiresOciEncryption := (isEncrypted(src) && ic.c.options.OciDecryptConfig != nil) || c.options.OciEncryptLayers != nil
+	destRequiresOciEncryption := (isEncrypted(src) && opt.options.OciDecryptConfig != nil) || opt.options.OciEncryptLayers != nil
```
mtrmac, line 164: An example: Compare how `copy/encryption` accesses `ic.c.options.Oci*`. So now the two parts can get out of sync if a caller to `copySingleImage` passes options that differ from `ic.c.options`.

copy/copy.go
```diff
@@ -251,7 +251,11 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
 
 	if !multiImage {
 		// The simple case: just copy a single image.
-		single, err := c.copySingleImage(ctx, c.unparsedToplevel, nil, false)
+		singleImageOpts := copySingleImageOptions{unparsedImage: c.unparsedToplevel,
```
mtrmac, line 254: Non-blocking: In the three callers, I think inlining the struct constructor would be a tiny bit more readable — it would be clear that the data is single-use and readers wouldn’t have to worry about future uses of `singleImageOpts`.

copy/single.go
mtrmac: I’m sorry, the motivation was clearly explained in a commit message.

Yes, in that case, let’s make only the compression format difference new fields in `copySingleImageOptions`, instead of a full copy, with the intent ambiguities (and, much less important, the overhead of having to make a copy of the large `copy.Options` struct).

flouthoc: Should `copySingleImageOptions` include only compression fields or entire `DestinationCtx`, for instance see `copy/single.go` its using `DestinationCtx` directly at lot of places, I'm thinking of including a custom `DestinationCtx` in `copySingleImageOptions` and then in `imageCopier` as well so other helpers can use it.

flouthoc: I am not sure if `DestinationCtx` should reflect custom compression at all places in `copy/single.go` or not, I am assuming not in that case I can only add compression fields to `copySingleImageOptions`.

flouthoc: Nvm I think `DestinationCtx` is not needed only compression should be enough.

mtrmac: Just the compression parameters, I think. Practically, `copy/multiple.go` does not need to create a copy of `DestinationCtx` just like it shouldn’t need to create a copy of `Options`.

It _does_ make sense to reason that if `SystemContext` contains compression choices, and we are passing `SystemContext` to transports, then we should update the `SystemContext` with the chosen compression — but transports do not care (and must not care! they must write the data in PutBlob without modification for the manifest to be valid); `SystemContext.Compression…` is only used in c/image/copy, so creating an updated clone is, I think, unnecessary.

---

The history is that c/image was originally just code in a Skopeo CLI tool, where IIRC CLI parameters were just global variables. When we split c/image into a separate project, we shuffled those values to a single global `SystemContext` rather than restructure the option passing completely.

_Nowadays_, basically `SystemContext` exists primarily so that `c/image/copy` can be generic code which doesn’t know much about transports options, and callers can still set options for the transports; I’m trying to (_very slowly_) move isolated c/image subpackages away from using `SystemContext` for options, compare e.g. `signature/simplesigning`[1].  In that sense, passing the compression choice in `DestinationCtx` and not directly in `Options` is plainly not something we would do nowadays.

--- 

[1] (but also, see how much more code doing it the modern way is, compared to just adding a field to `SystemContext`)

flouthoc:
I'll rebase on final review.

copy/single.go
flouthoc: SGTM

copy/single.go
```diff
@@ -131,12 +137,12 @@ func (c *copier) copySingleImage(ctx context.Context, unparsedImage *image.Unpar
...
-		requireCompressionFormatMatch: requireCompressionFormatMatch,
+		requireCompressionFormatMatch: opts.requireCompressionFormatMatch,
 	}
 	if c.options.DestinationCtx != nil {
 		// Note that compressionFormat and compressionLevel can be nil.
-		ic.compressionFormat = c.options.DestinationCtx.CompressionFormat
-		ic.compressionLevel = c.options.DestinationCtx.CompressionLevel
+		ic.compressionFormat = opts.compressionFormat
+		ic.compressionLevel = opts.compressionLevel
 	}
```
mtrmac, line 146: For our goals, it doesn‘t work to gate `ic.compression…` use on `c.options.DestinationCtx != nil` (assuming we won’t create a clone of `DestinationCtx`; and if we did create that clone, we wouldn’t need the new fields in `opts` anyway) — the clone will just set values in `opts`, and `DestinationCtx` could well be `nil`.

---

Also the way the responsibility for `options.DestinationCtx` is moved to callers now, the logic needs to be in two places.

I think this can be, in `copySingleImage`:
```go
if opts.compressionFormat != nil {
  ic.compression$both = opts.compression$both
} else if `c.options.DestinationCtx != nil {
  // existing comment
  ic.compression$both = c.options.DestinationCtx.Compression$both
}
```

flouthoc: Nice catch! I have fixed it now.

copy/copy.go
```diff
@@ -249,9 +250,18 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
 		return nil, fmt.Errorf("determining manifest MIME type for %s: %w", transports.ImageName(srcRef), err)
 	}
 
+	// Prepare copySingleImageOptions
+	var compressionFormat *compression.Algorithm
+	var compressionLevel *int
+	if c.options.DestinationCtx != nil {
+		compressionFormat = c.options.DestinationCtx.CompressionFormat
+		compressionLevel = c.options.DestinationCtx.CompressionLevel
+	}
```
mtrmac, line 259: This is now an unnecessary duplication of code centralized in `copySingleImage`. Same for the `multiple.go` copy.

copy/single.go
```diff
@@ -131,10 +137,13 @@ func (c *copier) copySingleImage(ctx context.Context, unparsedImage *image.Unpar
 		src:             src,
 		// diffIDsAreNeeded is computed later
 		cannotModifyManifestReason:    cannotModifyManifestReason,
-		requireCompressionFormatMatch: requireCompressionFormatMatch,
+		requireCompressionFormatMatch: opts.requireCompressionFormatMatch,
 	}
-	if c.options.DestinationCtx != nil {
+	if opts.compressionFormat != nil {
 		// Note that compressionFormat and compressionLevel can be nil.
```
mtrmac, line 143: This comment belongs to the old `DestinationCtx` case. In the new one, we have ensured `opts.compressionFormat != nil` just above.

copy/copy.go
flouthoc: Yeah I don't need to set these I can just set in `instanceCopyClone` case.

copy/single.go
```diff
@@ -40,6 +40,12 @@ type imageCopier struct {
 	requireCompressionFormatMatch bool
 }
 
+type copySingleImageOptions struct {
+	requireCompressionFormatMatch bool
+	compressionFormat             *compressiontypes.Algorithm // Compression algorithm to use, if the user explicitly requested one, or nil.
```
mtrmac, line 45:
```suggestion
	compressionFormat             *compressiontypes.Algorithm // Compression algorithm to use, if the caller explicitly requested one, or nil.
```
for future reference

mtrmac (approved):
Thanks! LGTM.

copy/single.go
```diff
@@ -40,6 +40,12 @@ type imageCopier struct {
...
+	compressionLevel              *int
```
mtrmac, line 46:
```suggestion
	compressionLevel              *int // only used if compressionFormat is set
```

mtrmac:
Thanks again!

PR #2062, opened 2023-07-31
OCI image-spec / distribution-spec v1.1 updates, first round

mtrmac:
This is an assorted set of small changes to support OCI image-spec/distribution-spec 1.1.

Follow #2030 for checklists of things to update/address.

See individual commit messages for details. Note especially the RFC and UNTESTED ones.


vrothberg (approved):
LGTM

PR #2068, opened 2023-08-02
copy: add support for `ForceCompressionFormat`

flouthoc:
ForceCompressionFormat allows end users to force selected compression format (set in DestinationCtx.CompressionFormat) which ensures that while copying blobs of other compression algorithms are not reused.

Following flag is a frontend wrapper for: https://github.com/containers/image/pull/2023 

Will help in:
* https://github.com/containers/buildah/issues/4613
* https://github.com/containers/podman/issues/18660

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -133,6 +133,10 @@ type Options struct {
 	// Invalid when copying a non-multi-architecture image. That will probably
 	// change in the future.
 	EnsureCompressionVariantsExist []OptionCompressionVariant
+	// ForceCompressionFormat ensures that the compression algorithm set in
+	// DestinationCtx.CompressionFormat is used exclusively, and blobs of other
+	// compression algorithms are not reused.
+	ForceCompressionFormat bool
 }
 
 // OptionCompressionVariant allows to supply information about
@@ -161,6 +165,14 @@ type copier struct {
 	concurrentBlobCopiesSemaphore *semaphore.Weighted // Limits the amount of concurrently copied blobs
 	signers                       []*signer.Signer    // Signers to use to create new signatures for the image
 	signersToClose                []*signer.Signer    // Signers that should be closed when this copier is destroyed.
+}
+
+// Internal function to validate `requireCompressionFormatMatch` for copySingleImageOptions
+func shouldRequireCompressionFormatMatch(options *Options) (bool, error) {
+	if options.ForceCompressionFormat && (options.DestinationCtx == nil || options.DestinationCtx.CompressionFormat == nil) {
+		return false, fmt.Errorf("cannot use ForceCompressionFormat with undefined default compression format")
+	}
+	return options.ForceCompressionFormat, nil
 }
 
 // Image copies image from srcRef to destRef, using policyContext to validate
@@ -269,8 +281,12 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
 		if len(options.EnsureCompressionVariantsExist) > 0 {
 			return nil, fmt.Errorf("EnsureCompressionVariantsExist is not implemented when not creating a multi-architecture image")
 		}
+		requireCompressionFormatMatch, err := shouldRequireCompressionFormatMatch(options)
+		if err != nil {
+			return nil, err
+		}
 		// The simple case: just copy a single image.
-		single, err := c.copySingleImage(ctx, c.unparsedToplevel, nil, copySingleImageOptions{requireCompressionFormatMatch: false})
+		single, err := c.copySingleImage(ctx, c.unparsedToplevel, nil, copySingleImageOptions{requireCompressionFormatMatch: requireCompressionFormatMatch})
 		if err != nil {
 			return nil, err
 		}
@@ -278,6 +294,10 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
 	} else if c.options.ImageListSelection == CopySystemImage {
 		if len(options.EnsureCompressionVariantsExist) > 0 {
 			return nil, fmt.Errorf("EnsureCompressionVariantsExist is not implemented when not creating a multi-architecture image")
+		}
+		requireCompressionFormatMatch, err := shouldRequireCompressionFormatMatch(options)
+		if err != nil {
+			return nil, err
 		}
 		// This is a manifest list, and we weren't asked to copy multiple images.  Choose a single image that
 		// matches the current system to copy, and copy it.
@@ -295,7 +315,7 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
 		}
 		logrus.Debugf("Source is a manifest list; copying (only) instance %s for current system", instanceDigest)
 		unparsedInstance := image.UnparsedInstance(rawSource, &instanceDigest)
-		single, err := c.copySingleImage(ctx, unparsedInstance, nil, copySingleImageOptions{requireCompressionFormatMatch: false})
+		single, err := c.copySingleImage(ctx, unparsedInstance, nil, copySingleImageOptions{requireCompressionFormatMatch: requireCompressionFormatMatch})
 		if err != nil {
 			return nil, fmt.Errorf("copying system image from manifest list: %w", err)
 		}
--- a/copy/multiple.go
+++ b/copy/multiple.go
@@ -33,6 +33,10 @@ type instanceCopy struct {
 	sourceDigest digest.Digest
 
 	// Fields which can be used by callers when operation
+	// is `instanceCopyCopy`
+	copyForceCompressionFormat bool
+
+	// Fields which can be used by callers when operation
 	// is `instanceCopyClone`
 	cloneCompressionVariant OptionCompressionVariant
 	clonePlatform           *imgspecv1.Platform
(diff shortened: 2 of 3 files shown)


flouthoc:
@mtrmac @vrothberg PTAL

copy/copy.go
```diff
@@ -133,6 +133,11 @@ type Options struct {
 	// Invalid when copying a non-multi-architecture image. That will probably
 	// change in the future.
 	EnsureCompressionVariantsExist []OptionCompressionVariant
+
```
mtrmac, line 136: A super-tiny nit: Without this empty line we could structure `Options` a bit more into “paragraphs” with related options.

copy/copy.go
```diff
@@ -133,6 +133,11 @@ type Options struct {
...
+	// ForceCompressionFormat allows end users to force selected compression format
+	// (set in DestinationCtx.CompressionFormat) which ensures that while copying
+	// blobs of other compression algorithms are not reused.
```
mtrmac, line 139: Non-blocking, this works fine:

Maybe something like
```suggestion
	// ForceCompressionFormat ensures that the compression algorithm set in
	// DestinationCtx.CompressionFormat is used exclusively, and blobs of other
	// compression algorithms are not reused.
```
would say the same thing a tiny bit more directly.

copy/multiple.go
```diff
@@ -230,6 +230,16 @@ func (c *copier) copyMultipleImages(ctx context.Context) (copiedManifest []byte,
 			logrus.Debugf("Copying instance %s (%d/%d)", instance.sourceDigest, i+1, len(instanceCopyList))
 			c.Printf("Copying image %s (%d/%d)\n", instance.sourceDigest, i+1, len(instanceCopyList))
 			unparsedInstance := image.UnparsedInstance(c.rawSource, &instanceCopyList[i].sourceDigest)
+			singleImageOpts := copySingleImageOptions{requireCompressionFormatMatch: false}
+			if c.options.ForceCompressionFormat {
+				singleImageOpts.requireCompressionFormatMatch = true
+				// If CompressionFormat is nil set it to `Gzip` as default,
+				// since requireCompressionFormatMatch is pretty much `no-op`
+				// for `nil` CompressionFormat.
+				if c.options.DestinationCtx.CompressionFormat == nil {
```
mtrmac, line 239: I think this case should fail instead; the user is asking for using _precisely_ $undefined. That just doesn’t make sense, and indicates the user made a mistake that the user would probably like to correct.

copy/multiple.go
```diff
@@ -230,6 +230,16 @@ func (c *copier) copyMultipleImages(ctx context.Context) (copiedManifest []byte,
 			logrus.Debugf("Copying instance %s (%d/%d)", instance.sourceDigest, i+1, len(instanceCopyList))
 			c.Printf("Copying image %s (%d/%d)\n", instance.sourceDigest, i+1, len(instanceCopyList))
 			unparsedInstance := image.UnparsedInstance(c.rawSource, &instanceCopyList[i].sourceDigest)
+			singleImageOpts := copySingleImageOptions{requireCompressionFormatMatch: false}
```
mtrmac, line 233: This value is not _actually_ used AFAICS

I would also prefer it to be set in `prepare…`; that way we can unit-test this. (I’m _not_ saying that a unit test is required.)

mtrmac (commented):
Thanks for working on this!

I think the two non-`copyMultipleImages` paths should also implement this (or, at the very least, fail with an error). It seems to me that the three locations would benefit from a small shared helper function.

copy/multiple.go
flouthoc: When user specifies no `--compression` in CLI then all tools (`buildah` and `podman`) assumes that default is gzip. Consider a case where repo contains some other algo ( `zstd` ) but user wants to override it with a default `compression` so he does `buildah push --forcecompression test` but in this case he never selected any `--compression` but requested a `--forcecompression` but nothing will happen since same `zstd` will be reused again.

In such case `--forcecompression` did nothing. I am assuming if no compression is specified `--forcecompression` must select default compression or do you think, implementer ( buildah, podman ) should do it ?

copy/multiple.go
flouthoc: done, added test.

flouthoc:
Addressed all comments but one doubt which needs to be discussed here: https://github.com/containers/image/pull/2068#discussion_r1281309739

copy/multiple.go
flouthoc: Ah nevermind if we error out user will be forced to provide a `--compression`

copy/multiple.go
```diff
@@ -29,8 +29,9 @@ const (
 )
 
 type instanceCopy struct {
-	op           instanceCopyKind
-	sourceDigest digest.Digest
+	op                            instanceCopyKind
+	sourceDigest                  digest.Digest
+	requireCompressionFormatMatch bool
```
vrothberg, line 34: Non-blocking nit: Why not also call it 'forceCompressionFormat`?  I find it confusing when multiple names/terms refer to the same thing.

vrothberg (approved):

copy/multiple.go
flouthoc: SGTM done.

copy/multiple.go
```diff
@@ -29,8 +29,9 @@ const (
 )
 
 type instanceCopy struct {
-	op           instanceCopyKind
-	sourceDigest digest.Digest
+	op                     instanceCopyKind
+	sourceDigest           digest.Digest
+	forceCompressionFormat bool
```
mtrmac, line 34: If this is relevant only to `instanceCopyCopy`, name it `copy…` and move it into a separate section.

---

Non-blocking: again, it seems more elegant to me to for have a `options copySingleImageOptions` field here, set with _all_ the options, shared for both `…Copy` and `…Clone`, and trivially applied in the execution loop.

copy/copy.go
```diff
@@ -269,8 +273,11 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
 		if len(options.EnsureCompressionVariantsExist) > 0 {
 			return nil, fmt.Errorf("EnsureCompressionVariantsExist is not implemented when not creating a multi-architecture image")
 		}
+		if (options.DestinationCtx == nil || options.DestinationCtx.CompressionFormat == nil) && options.ForceCompressionFormat {
+			return nil, fmt.Errorf("cannot use ForceCompressionFormat with undefined default compression format")
+		}
 		// The simple case: just copy a single image.
-		single, err := c.copySingleImage(ctx, c.unparsedToplevel, nil, copySingleImageOptions{requireCompressionFormatMatch: false})
+		single, err := c.copySingleImage(ctx, c.unparsedToplevel, nil, copySingleImageOptions{requireCompressionFormatMatch: c.options.ForceCompressionFormat})
```
mtrmac, line 280: At 3 copies, please turn this into a single shared function.

Maybe just
```go
func shouldRequireCompressionFormatMatch(options *Options) (bool, error)
```
or perhaps (uglier?)
```go
func basicSingleImageOptions(options *Options) (copySingleImageOptions, error)
```
OTOH it’s very unclear what “basic” means — typically defaults and logic should happen directly in `copySingleImage`, this would be more of a `nonInstanceCopyCloneSingleImageOptions`, and that’s just _tedious_.

copy/copy.go
```diff
@@ -269,8 +273,11 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
 		if len(options.EnsureCompressionVariantsExist) > 0 {
 			return nil, fmt.Errorf("EnsureCompressionVariantsExist is not implemented when not creating a multi-architecture image")
 		}
+		if (options.DestinationCtx == nil || options.DestinationCtx.CompressionFormat == nil) && options.ForceCompressionFormat {
```
mtrmac, line 276: (Absolutely non-blocking: Checking `options.ForceCompressionFormat` first, before looking at `DestinationCtx`  might be easier to follow (“if you request to force a compression format, you also must say which one”, rather than “if there is no compression format defined, forcing a compression format is makes no sense”; also all of this code is _primarily_ concerned with `ForceCompressionFormat`, so it makes sense to lead with that).

I suspect with a new helper function the ease of understanding will not be worth worrying about…)

mtrmac (commented):
ACK to the logic, now just some cleanup, please.

copy/copy.go
mtrmac: > Ah nevermind if we error out user will be forced to provide a `--compression`

Yes, that’s the idea.  I don’t think “use a CLI flag to force a compression, but rely on a config file to specify which one” is something users will really need… if they _care_ to force a compression, I’m fairly comfortable with having them type out which one they care about in the CLI as well.

Also note that `containers.conf` is not handled inside c/image — c/image just sees `copy.Options` that make little sense.

Admittedly if Buildah or Podman always set `CompressionFormat` value (e.g. from a `containers.conf` — I didn’t check if that’s the case), that detection of a “force but not say what” wouldn’t work I’m afraid, but that’s not something c/image can fix locally.

copy/copy.go
```diff
@@ -163,6 +167,14 @@ type copier struct {
 	signersToClose                []*signer.Signer    // Signers that should be closed when this copier is destroyed.
 }
 
+// Internal function to validate `requireCompressionFormatMatch` for copySingleImageOptions
+func shouldRequireCompressionFormatMatch(options *Options) (bool, error) {
```
mtrmac, line 171: (Hiding this somewhere at the bottom, not near the top of the file, might help newcomers to the codebase who want to understand what this file is about.)

mtrmac (approved):
Thanks! LGTM.

PR #2069, opened 2023-08-02
Use consistent example domains

fschwa:


vrothberg (approved):
LGTM
Thank you, @fschwa

mtrmac (changes requested):
Thanks, the man page really is inconsistent, but this is actually making it _more_ inconsistent?!

The text immediately below still says `.net`, so now the text and the example don’t match.

Also, the `…-for-example-com` name directly refers to the `example.com` being redirected, so it should probably continue to prefer `com` over `net` (Or, alternatively, use a name  similar to `registry-for-example-com.internal`.)

---

The same examples also exist in the default `registries.conf` in this repo.

Ideally all of that should be changed to use one host consistent host name.

fschwa:
You are right. It was kind of drive-by when actually doing something else. Will check for the other occurrences.

fschwa:
Alright. Updated. ... However, since you are talking about a redirect, might the .net be on purpose? I am not familiar with podman yet, so hard for me to tell.

mtrmac (approved):
Thanks! LGTM.

vrothberg (approved):
LGTM

PR #2070, opened 2023-08-03
Merge release branch into `main`

mtrmac:
The last tagged release, `v5.26.1`, is on a branch; users of c/image who refer to a post-release commit from `main` see Renovate try to downgrade, e.g. https://github.com/containers/skopeo/pull/2066 , and that breaks.

To help with that situation, merge the release branch into `main`, so that the later commits are recognized by Go as successors to the tagged release. Note that for this to make a difference, the users of c/image need to update their reference to include at least the merge commit from this PR.

Compare #1861 and related PRs from that time.


mtrmac:
Marking as draft — we’ll tag a 5.27.0, so let’s do that afterwards.

mtrmac:
Updated to merge the just-released `5.27` version into `main`.

Also bumps the version to `5.28.0-dev` to be thread of the `release-5.27` branch.

mtrmac:
@vrothberg / @flouthoc PTAL, this should allow vendoring a recent commit of c/image from `main` into dependencies without Renovate proposing downgrades.

flouthoc (approved):
LGTM

vrothberg (approved):
LGTM

vrothberg:
I wonder if there isn't a switch in GitHub to avoid the bots from downgrading?  I saw a "mark as latest release" button or something similar a while ago but did not research what it does.

mtrmac:
That would be a Renovate switch. There’s some history and complexity to that ( e.g. https://github.com/containers/automation/pull/134 , https://github.com/containers/automation/pull/149 ), I don’t know all the details.

Fundamentally the issue is that when Git history diverges (e.g. when tagging a release on a branch), commits on the divergent branches are, from Go’s point of view at least, unordered in the “more recent” (i.e. it is a partial order only). I think preferring a last tagged release to a commit is a pretty reasonable behavior (it only shows up for users who vendor by commit; a more removed users who just use published releases never see this). And then, adding a merge like this in Git to add an “ordering edge” in the object graph is, to me, a fairly logical outcome of the versioning framework as it exists — although it is not immediately intuitive.

PR #2071, opened 2023-08-04
[release-5.26] backport API changes `copy` and `list` components from upstream.

flouthoc:
PR's which are backported

* https://github.com/containers/image/pull/2023
* https://github.com/containers/image/pull/2029
* https://github.com/containers/image/pull/2039
* https://github.com/containers/image/pull/2040
* https://github.com/containers/image/pull/2047
* https://github.com/containers/image/pull/2050
* https://github.com/containers/image/pull/2059
* https://github.com/containers/image/pull/2048

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -6,9 +6,9 @@ env:
     #### Global variables used for all tasks
     ####
     # Name of the ultimate destination branch for this CI run
-    DEST_BRANCH: "main"
+    DEST_BRANCH: "release-5.26"
     # CI container image tag (c/skopeo branch name)
-    SKOPEO_CI_TAG: "main"
+    SKOPEO_CI_TAG: "v1.13.0"
     # Use GO module mirror (reason unknown, travis did it this way)
     GOPROXY: https://proxy.golang.org
     # Overrides default location (/tmp/cirrus) for repo clone
--- a/copy/blob.go
+++ b/copy/blob.go
@@ -83,12 +83,12 @@ func (ic *imageCopier) copyBlobFromStream(ctx context.Context, srcReader io.Read
 		return types.BlobInfo{}, err
 	}
 
-	// === Report progress using the ic.c.progress channel, if required.
-	if ic.c.progress != nil && ic.c.progressInterval > 0 {
+	// === Report progress using the ic.c.options.Progress channel, if required.
+	if ic.c.options.Progress != nil && ic.c.options.ProgressInterval > 0 {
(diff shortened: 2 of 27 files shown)


flouthoc:
@mtrmac PTAL

flouthoc:
Huh skopeo tests are failing I wonder why ? is it because I missed to backport some commit ?

mtrmac:
Code LGTM… but per the other conversation, this should be a 5.27.0. Let me work on that…

mtrmac:
@flouthoc I have filed #2074 for what I think is necessary to make the tests pass, and #2075 as a proof of concept — note that that’s against the wrong branch.

flouthoc:
Okay should I close this PR and create similar PR against `release-5.27` then

mtrmac:
@flouthoc Now that CI on the 5.27 branch is running, yes, let’s do a 5.27.0.

I’ll update #2075 just so that it is ready, but you did all the backporting work so you deserve the full credit — if you want to rebase+re-file this, by all means go ahead.

.cirrus.yml
```diff
@@ -6,9 +6,9 @@ env:
     #### Global variables used for all tasks
     ####
     # Name of the ultimate destination branch for this CI run
-    DEST_BRANCH: "main"
+    DEST_BRANCH: "release-5.26"
```
mtrmac, line 9: Great catch!

.cirrus.yml
```diff
@@ -6,9 +6,9 @@ env:
...
     # CI container image tag (c/skopeo branch name)
-    SKOPEO_CI_TAG: "main"
+    SKOPEO_CI_TAG: "v1.13.0"
```
mtrmac, line 11: I think using a branch name here is slightly better; if we ever tag a v1.13.1 on the stable branch, a branch name reference means that we would keep testing against the relevant Skopeo user.

(Historically `SKOPEO_CI_TAG` was used in `IMAGE_SUFFIX`, hence the name… now that that’s no longer the case, it might make sense to rename this variable to not be confusing — only on `main`, and only if @cevich approves.)

flouthoc:
@mtrmac it seems just pointing https://github.com/containers/image/pull/2075 to newer branch will do the trick, it will be easier for you to point changes https://github.com/containers/image/pull/2075 to `release-5.27`. Thanks.

.cirrus.yml
mtrmac: Filed https://github.com/containers/image/pull/2083 .

PR #2072, opened 2023-08-04
Add a way to customize the gpg program used for signing

gregoire-mullvad:
Add a constructor to make a SigningMechanism with a custom program. This
can be used to sign images with a gpg wrapper (e.g. qubes-gpg-client-wrapper)
or an alternative implementation (e.g. sequoia-chameleon-gnupg).

Related to https://github.com/containers/podman/issues/15838

### DIFF
--- a/signature/mechanism.go
+++ b/signature/mechanism.go
@@ -60,6 +60,13 @@ func NewGPGSigningMechanism() (SigningMechanism, error) {
 	return newGPGSigningMechanismInDirectory("")
 }
 
+// NewGPGSigningMechanismWithProgram returns a new GPG/OpenPGP signing mechanism using the given programm
+// instead of the default gpg binary.
+// The caller must call .Close() on the returned SigningMechanism.
+func NewGPGSigningMechanismWithProgram(program string) (SigningMechanism, error) {
+	return newGPGSigningMechanismWithProgram(program)
+}
+
 // NewEphemeralGPGSigningMechanism returns a new GPG/OpenPGP signing mechanism which
 // recognizes _only_ public keys from the supplied blob, and returns the identities
 // of these keys.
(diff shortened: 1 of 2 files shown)


gregoire-mullvad:
@mtrmac thanks for the review. I was indeed a bit unsure about the different APIs, and decided that a good starting point was to make this work with `podman image sign`, and if I'm not mistaken that command uses the `SigningMechanism` directly. (Since the `simplesigning` API builds on top of `signingMechanism` this seemed like a natural starting point.)
I think doing it the way you described would then require refactoring `podman image sign` to use this other API. Should I try to do that?

mtrmac:
@gregoire-mullvad I’m afraid I don’t much think about `podman image sign`, that command is very tempting to use in insecure ways (compared to `podman push --sign-by` when done by the trusted build system or the like). So first a gentle reminder to try using `podman push --sign-by` or `skopeo copy --sign-by` when copying from a trusted to a potentially-untrusted system instead — though there certainly are ways to use `podman image sign` securely.

---

Right now, `signature/signer.Signer` is basically completely opaque and only consumable by `copy.Image`; that’s not useful for `podman image sign` as is.

To make that work, `signature/internal.signer.SignImageManifest` would need to be made public. I wouldn’t mind that — it was made private originally just so that we didn’t have to commit to a public API without a user. As long as we don’t add interfaces that might need changing or extending to the public API, adding a public function which we can possibly redirect to a different internal abstraction should be fine.

mtrmac:
Thanks for the update.

Nowadays most development seems to center around the `sigstore` signature format, so I’m fairly unlikely to work on this either. (FWIW, there _are_ some plans to allow alternatives to `gpgme`, but the timing of my work on that is very uncertain.)

So, closing for now.

PR #1272, opened 2021-06-23
Cirrus: Grab skopeo from main branch

cevich:
Ref: https://github.com/containers/skopeo/issues/1282

Signed-off-by: Chris Evich <cevich@redhat.com>

PR #2083, opened 2023-08-07
Rename SKOPEO_CI_TAG to SKOPEO_CI_BRANCH

mtrmac:
Originally (in https://github.com/containers/image/pull/1272 ) `SKOPEO_CI_TAG` was also a tag of an on-registry image. That's no longer the case.

On c/image stable branches, we typically want to test against a corresponding skopeo stable branch, not against a specific frozen version on that stable branch, so the variable name has become a bit confusing.

So, rename it to `SKOPEO_CI_BRANCH`.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -7,8 +7,8 @@ env:
     ####
     # Name of the ultimate destination branch for this CI run
     DEST_BRANCH: "main"
-    # CI container image tag (c/skopeo branch name)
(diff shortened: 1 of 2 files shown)


mtrmac:
Ref. https://github.com/containers/image/pull/2071#discussion_r1284983478 .

cevich (approved):
LGTM

PR #2092, opened 2023-08-16
Implement, and default to, a SQLite BlobInfoCache instead of BoltDB

mtrmac:
This implements a SQLite-based `types.BlobInfoCache`, using it instead of BoltDB.

---

@mheon @Luap99 @vrothberg PTAL at the SQLite concurrency configuration, at least; this is cribbed from Podman but I haven’t followed the evolving situation there and I don’t know what I don’t know.

---

<s>Marked as draft because I didn’t actually test the implementation (other than existing unit tests), and I’m not sure about lifetime management.</s>

It turns out that the underlying `*sql.DB` (and `*sql.Conn`) objects would benefit from a having the caller call `Close`, which is not currently a part of the `BlobInfoCache` model. (Alternatively, the SQLite implementation could create and tear down SQLite state for every single write…) That, indirectly, means that the `SystemContext.BlobInfoCacheDir` option can lead to memory leaks, so deprecate it.

There are _some_ known users (https://github.com/crazy-max/undock/blob/30cfb56ef6e351fbc81f0a7f1b230e6afa639b18/internal/extractor/image/cache.go#L103 , https://github.com/openeuler-mirror/isula-build/blob/869cf28a6102d411feba01d9bb4aa9e7f4a2d3bc/image/context.go#L40 ) but they seem very rare, and not immediately sure they _should_ be overriding the location.

Alternatively, we might add a `Close` method and use it at least in the `copy.Image` caller, special-casing the default case (or perhaps not special-casing it.)

---

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,6 +20,7 @@ require (
 	github.com/klauspost/compress v1.16.7
 	github.com/klauspost/pgzip v1.2.6
 	github.com/manifoldco/promptui v0.9.0
+	github.com/mattn/go-sqlite3 v1.14.17
 	github.com/opencontainers/go-digest v1.0.0
 	github.com/opencontainers/image-spec v1.1.0-rc4
 	github.com/opencontainers/selinux v1.11.0
--- a/pkg/blobinfocache/sqlite/sqlite.go
+++ b/pkg/blobinfocache/sqlite/sqlite.go
@@ -1,0 +1,553 @@
+// Package boltdb implements a BlobInfoCache backed by SQLite.
+package sqlite
+
+import (
+	"database/sql"
+	"errors"
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/containers/image/v5/internal/blobinfocache"
+	"github.com/containers/image/v5/pkg/blobinfocache/internal/prioritize"
+	"github.com/containers/image/v5/types"
+	_ "github.com/mattn/go-sqlite3" // Registers the "sqlite3" backend backend for database/sql
+	"github.com/opencontainers/go-digest"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	// NOTE: There is no versioning data inside the file; this is a “cache”, so on an incompatible format upgrade
+	// we can simply start over with a different filename; update blobInfoCacheFilename.
+	// That also means we don’t have to worry about co-existing readers/writers which know different versions of the schema
+	// (which would require compatibility in both directions).
+
+	// Assembled sqlite options used when opening the database.
+	sqliteOptions = "?" +
+		// Deal with timezone automatically.
+		// go-sqlite3 always _records_ timestamps as a text: time in local time + a time zone offset.
+		// _loc affects how the values are _parsed_: (which timezone is assumed for numeric timestamps or for text which does not specify an offset, or)
+		// if the time zone offset matches the specified time zone, the timestamp is assumed to be in that time zone / location;
+		// (otherwise an unnamed time zone carrying just a hard-coded offset, but no location / DST rules is used).
+		"_loc=auto" +
+		// Force an fsync after each transaction (https://www.sqlite.org/pragma.html#pragma_synchronous).
+		"&_sync=FULL" +
+		// Allow foreign keys (https://www.sqlite.org/pragma.html#pragma_foreign_keys).
+		// We don’t currently use any foreign keys, but this is a good choice long-term (not default in SQLite only for historical reasons).
+		"&_foreign_keys=1" +
+		// Use BEGIN EXCLUSIVE (https://www.sqlite.org/lang_transaction.html);
+		// i.e. obtain a write lock for _all_ transactions at the transaction start (never use a read lock,
+		// never upgrade from a read to a write lock - that can fail if multiple read lock owners try to do that simultaneously).
+		//
+		// This, together with go-sqlite3’s default for _busy_timeout=5000, means that we should never see a “database is locked” error,
+		// the database should block on the exclusive lock when starting a transaction, and the problematic case of two simultaneous
+		// holders of a read lock trying to upgrade to a write lock (and one necessarily failing) is prevented.
+		// Compare https://github.com/mattn/go-sqlite3/issues/274 .
+		//
+		// Ideally the BEGIN / BEGIN EXCLUSIVE decision could be made per-transaction, compare https://github.com/mattn/go-sqlite3/pull/1167
+		// or https://github.com/mattn/go-sqlite3/issues/400 .
+		// The currently-proposed  workaround is to create two different SQL “databases” (= connection pools) with different _txlock settings,
+		// which seems rather wasteful.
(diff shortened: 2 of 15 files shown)


pkg/blobinfocache/sqlite/sqlite.go
```diff
@@ -0,0 +1,498 @@
...
+			return zeroValue, false, nil
+		}
+		return zeroValue, false, err
+	}
+	return value, true, nil
+}
+
+// ensureDBHasCurrentSchema adds the necessary tables and indices to a database.
+// This is typically used when creating a previously-nonexistent database.
+// We don’t really anticipate schema migratinos; with c/image usually vendored, not using
```
vrothberg, line 164: s/migratinos/migrations/

Luap99:
I haven't worked on any of the sqlite code so I cannot really comment much.
My only question is if we want to keep the design of multiple single db files instead of one shared sqlite instance across our projects because I think that was discussed early (not sure if there was a conclusion)?

mtrmac:
> My only question is if we want to keep the design of multiple single db files instead of one shared sqlite instance

Thanks, that’s a great question.

In this case I’m tempted to say that a separate one is good enough, because the schema management and error handling for a cache that can be safely discarded are so different from the critical production database, and because there’s no direct relationship to other objects that would benefit from foreign keys; and the times of data access are probably not directly related. I think the case for sharing a database between c/storage and Podman’s container database is quite a bit stronger.

But I’m in the middle of benchmarking how the database behaves, and so far the data I’m seeing strongly suggests I don’t have a good mental model of the costs/overhead of SQLite yet, so treat the above with skepticism.

mtrmac:
For the record, with various WIP commits pushed:
```
% go test -bench '.*' ./pkg/blobinfocache 
goos: darwin
goarch: arm64
pkg: github.com/containers/image/v5/pkg/blobinfocache
BenchmarkMemory-10          	   65023	     17034 ns/op
BenchmarkBoltDB-10          	       2	 886133666 ns/op
BenchmarkSQLiteShared-10    	      44	  26233122 ns/op
BenchmarkSQLiteReinit-10    	      48	  28239831 ns/op
BenchmarkJSON-10            	      50	 217408559 ns/op
PASS
ok  	github.com/containers/image/v5/pkg/blobinfocache	18.454s
```

BoltDB (with opening/closing the cache _file_ on every operation, so that it does not block other pull/push processes) is _extremely_ slower than SQLITE (opening the file once, and locking/unlocking on every operation). A file with JSON is much slower than SQLite — and that seems to be mostly _CPU_ time encoding/decoding. Surprisingly it’s still faster than BoltDB.

More next week…

mtrmac:
… the above is nonsense because different benchmark sizes result in different data sizes, and that directly affects costs (especially for JSON, which is O(N) CPU, memory and I/O even in the ideal case).

A somewhat better set of data; still directionally similar, but also shows the scaling behavior:
```console
% go test -benchtime 1x -bench '.*' ./pkg/blobinfocache 
goos: darwin
goarch: arm64
pkg: github.com/containers/image/v5/pkg/blobinfocache
BenchmarkMemory-10          	       1	     25542 ns/op
BenchmarkBoltDB-10          	       1	 815436375 ns/op
BenchmarkSQLiteShared-10    	       1	  41601709 ns/op
BenchmarkSQLiteReinit-10    	       1	  27213000 ns/op
BenchmarkJSON-10            	       1	  22839750 ns/op
PASS
ok  	github.com/containers/image/v5/pkg/blobinfocache	1.304s
% go test -benchtime 10x -bench '.*' ./pkg/blobinfocache 
goos: darwin
goarch: arm64
pkg: github.com/containers/image/v5/pkg/blobinfocache
BenchmarkMemory-10          	      10	     17733 ns/op
BenchmarkBoltDB-10          	      10	 738976475 ns/op
BenchmarkSQLiteShared-10    	      10	  26275783 ns/op
BenchmarkSQLiteReinit-10    	      10	  26706462 ns/op
BenchmarkJSON-10            	      10	  88556004 ns/op
PASS
ok  	github.com/containers/image/v5/pkg/blobinfocache	10.058s
% go test -benchtime 100x -bench '.*' ./pkg/blobinfocache 
goos: darwin
goarch: arm64
pkg: github.com/containers/image/v5/pkg/blobinfocache
BenchmarkMemory-10          	     100	     17338 ns/op
BenchmarkBoltDB-10          	     100	 698754928 ns/op
BenchmarkSQLiteShared-10    	     100	  25252404 ns/op
BenchmarkSQLiteReinit-10    	     100	  27071680 ns/op
BenchmarkJSON-10            	     100	 258581659 ns/op
PASS
ok  	github.com/containers/image/v5/pkg/blobinfocache	102.243s
```

mtrmac:
Using SQLite with per-operation connections, and closing the connection pool every time (which should be comparable to the BoltDB approach) clocks at around 40000000 ns/op . So that’s significant but still >15x faster than BoltDB..??!

mtrmac:
Now ready for review.

I have changed the lifetime handling design: instead of keeping a long-term cache of open databases, the SQLite presents a compatible implementation to hypothetical external callers (opening/closing the database on every operation); and c/image/copy uses internal `Open`/`Close` methods so that we keep a single SQLite handle during a single image copy. That’s a performance cost of <10% compared to a long-term handle (still vastly faster than BoltDB), and it means we don’t have to worry about breaking the 2 identified callers which set a non-default `BlobInfoCacheDir`.

rhatdan:
LGTM

mtrmac:
Great idea, Podman testing PR: https://github.com/containers/podman/pull/19718 .

mtrmac:
https://github.com/containers/podman/pull/19718 turned into a quest to strip c/image from `podman-remote` (with a successful PoC). See https://github.com/containers/podman/pull/19739 for a test PR _only_ updating c/image .

vrothberg:
LGTM

Thanks _a ton_, @mtrmac !  Feel free to merge this PR when you feel ready.

mtrmac:
https://github.com/containers/podman/pull/19739 succeeded.

For now I plan to keep this open, and to work on de-bloating `podman-remote` first, so that there is no interim regression. Alternatively, this could be merged as is, depending on other prioritization decisions.

rhatdan:
@mtrmac any update on this, would like to get this into podman 4.7

mtrmac:
Discussed this with @mheon — merging this now (causing `podman-remote` size to increase after it is vendored), and we’ll deal with the `podman-remote` dependency on c/image later.

go.mod
```diff
@@ -20,6 +20,7 @@ require (
 	github.com/klauspost/compress v1.16.7
 	github.com/klauspost/pgzip v1.2.6
 	github.com/manifoldco/promptui v0.9.0
+	github.com/mattn/go-sqlite3 v1.14.17
```
crazy-max, line 23: @mtrmac I think it would have been better to consider a CGo-free module like https://pkg.go.dev/modernc.org/sqlite. More info https://gitlab.com/cznic/sqlite

PR #2096, opened 2023-08-22
[release-5.26] Bump c/storage to 1.48.1, c/image to 5.26.2 and 5.26.3-dev

TomSweeneyRedHat:
Bump c/storage to 1.48.1
Bump c/image to 5.26.2
Bump c/image to 5.26.3-dev
 
in preparation for RHEL 8.9/9.3 final freeze

[NO NEW TESTS NEEDED]

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/BurntSushi/toml v1.3.2
 	github.com/containers/libtrust v0.0.0-20230121012942-c1716e8a8d01
 	github.com/containers/ocicrypt v1.1.7
-	github.com/containers/storage v1.48.0
+	github.com/containers/storage v1.48.1
 	github.com/cyberphone/json-canonicalization v0.0.0-20230514072755-504adb8a8af1
 	github.com/docker/distribution v2.8.2+incompatible
 	github.com/docker/docker v24.0.2+incompatible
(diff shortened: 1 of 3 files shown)


vrothberg (approved):
LGTM

But I don´t think this bump is necessary.  Consumers of c/storage can just bump c/storage directly.  So we only need to bump skopeo, buildah and podman and can ignore image and common.

TomSweeneyRedHat:
@vrothberg @rhatdan @mtrmac could I have a LGTM (or two) and merge on this please?  This is part of this weeks vendor dance for RHEL 8.9/9.3

vrothberg:
@TomSweeneyRedHat please see my comment above

mtrmac:
@TomSweeneyRedHat @vrothberg I don’t care either way. It’s extra work and it’s not _necessary_ for the consumers, but dependency-only minor version bumps are not unheard of, almost all the work has been already done, and keeping the process consistent is a bit valuable to help guide future cases.

I’ll leave at least an hour more for discussion in case there are other arguments — tentatively I’m thinking of merging this.

vrothberg:
It won't hurt too merge, so I'll go ahead.  My main point is that we don't need it technically and that it makes the vendor dance longer than needed.

TomSweeneyRedHat:
Thanks @mtrmac and @vrothberg .  I'm admitedly a little OCD about this, but I like to have the numbers all fresh at the start of a new RHEL release.  Thanks for putting up with the quirks.

PR #2104, opened 2023-09-04
fix removal of temp file in GetBlob on Windows

mikenorgate:
Unfortunately the eager removal of the temp file does not work on all versions of Windows and causes an error. To work around this we have to fallback to relying on the caller to invoke `Close()`

### DIFF
--- a/storage/storage_src.go
+++ b/storage/storage_src.go
@@ -23,6 +23,7 @@ import (
 	"github.com/containers/image/v5/types"
 	"github.com/containers/storage"
 	"github.com/containers/storage/pkg/archive"
+	"github.com/containers/storage/pkg/ioutils"
 	digest "github.com/opencontainers/go-digest"
 	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


storage/storage_src.go
```diff
@@ -9,6 +9,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/containers/storage/pkg/ioutils"
```
mtrmac, line 12: Please move this to the other section, with other imports from outside the standard library.

storage/storage_src.go
```diff
@@ -136,8 +137,10 @@ func (s *storageImageSource) GetBlob(ctx context.Context, info types.BlobInfo, c
 	}()
```
mtrmac, line 137: The error paths `if !success` also need to remove the file.

storage/storage_src.go
```diff
@@ -136,8 +137,10 @@ func (s *storageImageSource) GetBlob(ctx context.Context, info types.BlobInfo, c
...
 	// On Unix and modern Windows (2022 at least) we can eagerly unlink the file to ensure it's automatically
 	// cleaned up on process termination (or if the caller forgets to invoke Close())
+	// On older versions of Windows we will have to fallback to relying on the caller to invoke Close()
+	removeFailure := false
```
mtrmac, line 141: This reads to me as “remove a failure”.

Maybe `tmpFileRemovePending`, originally set to `true`, and cleared on success here?

storage/storage_src.go
```diff
@@ -136,8 +137,10 @@ func (s *storageImageSource) GetBlob(ctx context.Context, info types.BlobInfo, c
...
 	if err := os.Remove(tmpFile.Name()); err != nil {
```
mtrmac, line 142: (Should this check for a specific error condition? I’m not sure; I guess I/O errors would be reported in the `io.Copy` afterwards… OTOH we don’t want to add too many platform conditionals.)

mtrmac (changes requested):
Thanks!

(Oh, someone is actually running this Windows-native?!)

mtrmac:
Note to self: Compare discussion in https://github.com/containers/image/pull/1769 .

storage/storage_src.go
mikenorgate: I thought about this, but as you said it adds a platform based condition. I can but something in if you think it will be beneficial

mikenorgate:
@mtrmac I've addresses you comments

> (Oh, someone is actually running this Windows-native?!)

Yeh I'm one of those 😄

mtrmac (approved):
Thanks!

mtrmac:
I’m sure there are many other places where Windows does not provide optimal experience — we’d be happy to integrate improvements.

mikenorgate:
> I’m sure there are many other places where Windows does not provide optimal experience — we’d be happy to integrate improvements.

@mtrmac FYI I currently have a small application that builds, pulls, and pushes images using this library and so far this is the only issue I've had on Windows. I'm not fully exercising everything yet but will push any fixes I find

PR #2107, opened 2023-09-05
Fix build with golangci-lint 1.54.2

mtrmac:
Currently that fails with:
```
storage/storage_test.go:183:11: SA5011: possible nil pointer dereference (staticcheck)
		spec := ref.StringWithinTransport()
		        ^
storage/storage_test.go:188:6: SA5011(related information): this check suggests that the pointer can be nil (staticcheck)
		if ref == nil {
```

which is a real bug; and
```
storage/storage_test.go:183:11: SA5011: possible nil pointer dereference (staticcheck)
		spec := ref.StringWithinTransport()
		        ^
storage/storage_test.go:180:6: SA5011(related information): this check suggests that the pointer can be nil (staticcheck)
		if ref == nil {
```

which is a false positive, AFAICS.

See individual commit messages for details; one of them is just a workaround, but it seems worth doing anyway.

### DIFF
--- a/storage/storage_test.go
+++ b/storage/storage_test.go
@@ -185,7 +185,7 @@ func TestParseWithGraphDriverOptions(t *testing.T) {
 		if err != nil {
 			t.Fatalf("ParseReference(%q) returned error %v", "test", err)
 		}
-		if ref == nil {
+		if ref2 == nil {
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

PR #2110, opened 2023-09-06
Update dependencies of docker/docker

mtrmac:
`docker/docker` maintains its dependencies in `vendor.mod`, not `go.mod`, so automated tooling doesn't know to follow that.

Do that manually:

```
go get gotest.tools/v3@v3.5.0
```

This is actually _called_ only in tests, so it does not change the built binary. But update to make sure the dependency tree is not behind what docker/docker is testing with.

https://github.com/containers/image/pull/2109 updated `docker/docker` itself.


vrothberg (approved):
LGTM

PR #2121, opened 2023-09-18
progress: set Current before Refill

giuseppe:
the mpb.Progress doesn't allow to set the Refill value bigger than Current so make sure we set Current before setting Refill otherwise we will report the wrong number of skipped bytes.


giuseppe:
since the discussion in https://github.com/containers/image/pull/1980 is going to take longer, moved this unrelated patch to a separate PR

mtrmac (approved):
LGTM. Thanks!

PR #2123, opened 2023-09-18
ociarchive: Add new `ArchiveFileNotFoundError`

cgwalters:
This is for https://github.com/containers/skopeo/pull/2114
which is in turn a dependency of https://github.com/coreos/rpm-ostree/pull/4598

Basically I want to map ENOENT to a clear error, because the build
tooling wants to treat "target image is not present" differently
from "DNS lookup failed" or "we got EPERM".

There's a bit of code motion here because we need to move
the `os.Open()` call before creating a temporary directory.

---

### DIFF
--- a/oci/archive/oci_src.go
+++ b/oci/archive/oci_src.go
@@ -28,6 +28,18 @@ func (e ImageNotFoundError) Error() string {
 	return fmt.Sprintf("no descriptor found for reference %q", e.ref.image)
 }
 
+// ArchiveFileNotFoundError occurs when the archive file does not exist.
+type ArchiveFileNotFoundError struct {
+	// ref is the image reference
+	ref ociArchiveReference
+	// path is the file path that was not present
+	path string
+}
+
+func (e ArchiveFileNotFoundError) Error() string {
+	return fmt.Sprintf("archive file not found: %q", e.path)
+}
+
 type ociArchiveImageSource struct {
 	impl.Compat
 
--- a/oci/archive/oci_src_test.go
+++ b/oci/archive/oci_src_test.go
@@ -1,5 +1,26 @@
 package archive
 
-import "github.com/containers/image/v5/internal/private"
+import (
+	"path/filepath"
+	"testing"
+
+	"github.com/containers/image/v5/internal/private"
+	"github.com/containers/image/v5/types"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
 
 var _ private.ImageSource = (*ociArchiveImageSource)(nil)
+
+func TestNewImageSourceNotFound(t *testing.T) {
+	sysctx := types.SystemContext{}
+	emptyDir := t.TempDir()
+	archivePath := filepath.Join(emptyDir, "foo.ociarchive")
+	imgref, err := ParseReference(archivePath)
+	require.NoError(t, err)
+	_, err = LoadManifestDescriptorWithContext(&sysctx, imgref)
+	assert.NotNil(t, err)
+	var aerr ArchiveFileNotFoundError
+	assert.ErrorAs(t, err, &aerr)
+	assert.Equal(t, aerr.path, archivePath)
+}
--- a/oci/archive/oci_transport.go
+++ b/oci/archive/oci_transport.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io/fs"
 	"os"
 	"strings"
 
@@ -171,18 +172,24 @@ func createOCIRef(sys *types.SystemContext, image string) (tempDirOCIRef, error)
 
 // creates the temporary directory and copies the tarred content to it
 func createUntarTempDir(sys *types.SystemContext, ref ociArchiveReference) (tempDirOCIRef, error) {
+	src := ref.resolvedFile
+	arch, err := os.Open(src)
+	if err != nil {
+		if errors.Is(err, fs.ErrNotExist) {
+			return tempDirOCIRef{}, ArchiveFileNotFoundError{ref: ref, path: src}
+		} else {
+			return tempDirOCIRef{}, err
+		}
+	}
+	defer arch.Close()
+
 	tempDirRef, err := createOCIRef(sys, ref.image)
 	if err != nil {
 		return tempDirOCIRef{}, fmt.Errorf("creating oci reference: %w", err)
 	}
-	src := ref.resolvedFile
 	dst := tempDirRef.tempDirectory
+
 	// TODO: This can take quite some time, and should ideally be cancellable using a context.Context.
-	arch, err := os.Open(src)
-	if err != nil {
-		return tempDirOCIRef{}, err
-	}
-	defer arch.Close()
 	if err := archive.NewDefaultArchiver().Untar(arch, dst, &archive.TarOptions{NoLchown: true}); err != nil {
 		if err := tempDirRef.deleteTempDir(); err != nil {
 			return tempDirOCIRef{}, fmt.Errorf("deleting temp directory %q: %w", tempDirRef.tempDirectory, err)


cgwalters:
What's the right place to unit test this?

cgwalters:
Or, to address https://github.com/containers/skopeo/pull/2114#discussion_r1327313108 I guess we could create a new `ImageStructureNotFound` (or `ImageContainerNotFound`) error or so.  But I'd still want to map that underneath `OpenImageOptional` in skopeo's API.

oci/archive/oci_transport.go
```diff
@@ -171,18 +172,24 @@ func createOCIRef(sys *types.SystemContext, image string) (tempDirOCIRef, error)
 
 // creates the temporary directory and copies the tarred content to it
 func createUntarTempDir(sys *types.SystemContext, ref ociArchiveReference) (tempDirOCIRef, error) {
-	tempDirRef, err := createOCIRef(sys, ref.image)
-	if err != nil {
-		return tempDirOCIRef{}, fmt.Errorf("creating oci reference: %w", err)
-	}
 	src := ref.resolvedFile
-	dst := tempDirRef.tempDirectory
 	// TODO: This can take quite some time, and should ideally be cancellable using a context.Context.
```
mtrmac, line 176: When moving things around, this should probably end up just before `Untar()` below. The `Open` is not expected to take that long.

mtrmac (commented):
> Or, to address [containers/skopeo#2114 (comment)](https://github.com/containers/skopeo/pull/2114#discussion_r1327313108) I guess we could create a new `ImageStructureNotFound` (or `ImageContainerNotFound`) error or so.

I’d prefer that; `ImageNotFound` is explicitly documented not to match this situation

> But I'd still want to map that underneath `OpenImageOptional` in skopeo's API.

Sure, that works for me.

oci/archive/oci_transport.go
mtrmac: … this _comment_ should …

oci/archive/oci_src.go
```diff
@@ -28,6 +28,16 @@ func (e ImageNotFoundError) Error() string {
 	return fmt.Sprintf("no descriptor found for reference %q", e.ref.image)
 }
 
+// ImageStructureNotFoundError occurs when the archive file does not exist.
+type ImageStructureNotFoundError struct {
+	ref ociArchiveReference
+}
+
+func (e ImageStructureNotFoundError) Error() string {
+	return fmt.Sprintf("no archive file present for reference %q", e.ref.image)
```
mtrmac, line 37: This should include `ref.resolvedFile` (and possibly `ref.file`), the thing that is missing.

oci/archive/oci_transport.go
```diff
@@ -171,18 +172,24 @@ func createOCIRef(sys *types.SystemContext, image string) (tempDirOCIRef, error)
...
-		return tempDirOCIRef{}, err
+		if errors.Is(err, fs.ErrNotExist) {
+			return tempDirOCIRef{}, ImageStructureNotFoundError{ref: ref}
+		} else {
+			return tempDirOCIRef{}, err
+		}
 	}
 	defer arch.Close()
+
+	// TODO: This can take quite some time, and should ideally be cancellable using a context.Context.
```
mtrmac, line 186: All the way above `Untar`, please.

oci/archive/oci_src.go
```diff
@@ -28,6 +28,18 @@ func (e ImageNotFoundError) Error() string {
...
+// ImageStructureNotFoundError occurs when the archive file does not exist.
+type ImageStructureNotFoundError struct {
+	// ref is the image reference
+	ref ociArchiveReference
+	// path is the file path that was not present
+	path string
+}
+
+func (e ImageStructureNotFoundError) Error() string {
+	return fmt.Sprintf("no archive file present for reference %q", e.ref.image)
```
mtrmac, line 40: something like
```suggestion
	return fmt.Sprintf("archive file %q not found", e.ref.resolvedFile)
```
(optionally with `.file` and/or `.image`, or all of `.StringWithinTransport`) — users are mostly going to experience the error through this text. Not including `resolvedFile` at all could get pretty confusing.

oci/archive/oci_src.go
```diff
@@ -28,6 +28,18 @@ func (e ImageNotFoundError) Error() string {
 	return fmt.Sprintf("no descriptor found for reference %q", e.ref.image)
 }
 
+// ImageStructureNotFoundError occurs when the archive file does not exist.
+type ImageStructureNotFoundError struct {
```
mtrmac, line 32: (Non-blocking: Consider `Archive[File]NotFoundError`.)

mtrmac:
> What's the right place to unit test this?

I’d place it in a new `oci/archive/oci_src.go:TestNewImageSource`. (`TestReferenceNewImage` or a new unit test for `createUntarTempDir` would also work, but `*_transport.go` attract a bit more attention due to signature/identity concerns, so limiting the scope of those files comes with less concern.)

oci/archive/oci_src.go
cgwalters: I was thinking that we might add an equivalent to this for the `oci` directory transport, but I guess that could just be `DirectoryNotFoundError`.

mtrmac: I don’t feel strongly about the name … worst case, we can always add type aliases and deprecate old names.

mtrmac: It’s just that “structure” seems to be an unexpected word here, where I have to stop and think what it means — and my first guess would be that the archive exists but missing a required subdirectory, or something like that.

Not that I have a better idea if you want keep it generic … `ContainerNotFoundError` is not structurally any better, and “container” would be very misleading in this context.

cgwalters:
OK, updated for latest review and added a unit test.

oci/archive/oci_src.go
cgwalters: I'm good with `ArchiveFileNotFoundError`

oci/archive/oci_src_test.go
```diff
@@ -1,5 +1,26 @@
...
+	"github.com/containers/image/v5/internal/private"
+	"github.com/containers/image/v5/types"
+	"github.com/stretchr/testify/assert"
+)
 
 var _ private.ImageSource = (*ociArchiveImageSource)(nil)
+
+func TestNewImageSourceNotFound(t *testing.T) {
+	emptyDir := t.TempDir()
+	sysctx := types.SystemContext{BlobInfoCacheDir: emptyDir}
```
mtrmac, line 17: (This is not used for anything in this test.)

oci/archive/oci_src_test.go
```diff
@@ -1,5 +1,26 @@
...
+	archivePath := filepath.Join(emptyDir, "foo.ociarchive")
+	imgref, err := ParseReference(archivePath)
+	assert.Nil(t, err)
```
mtrmac, line 20: (`require.NoError(t, err)` would abort the test instead of continuing and crashing.

That’s completely hypothetical here, and it be changed if this ever failed.)

oci/archive/oci_src_test.go
```diff
@@ -1,5 +1,26 @@
...
+	_, err = LoadManifestDescriptorWithContext(&sysctx, imgref)
+	assert.NotNil(t, err)
+	var aerr ArchiveFileNotFoundError
+	assert.True(t, errors.As(err, &aerr))
```
mtrmac, line 24: (`assert.ErrorAs` might result in better messages on failure. This works fine.)

mtrmac (approved):
LGTM. Thanks!

oci/archive/oci_src_test.go
cgwalters: It's needed for `LoadManifestDescriptorWithContext` right?

mtrmac: I don’t see where; BIC really only applies to work with _layers_.

Do you mean `BigFilesTemporaryDir`?

cgwalters:
(Hm, the "PRs must be targeting tip to merge" on this repo is a little surprising)

oci/archive/oci_src_test.go
cgwalters: Oh, you meant `BlobInfoCacheDir` and not `sysctx`.  Sorry, fixed!

mtrmac (approved):

PR #2129, opened 2023-09-21
Workaround go-md2man warning about HTML comments

### LINKED ISSUE #2128: [doc] fix warning when generating man pages with go-md2man
I'm not sure if it is problem in MD's formatting or bug from go-md2man, reporting for sake of it.

Following three files seem to produce this warning.
```
/usr/bin/go-md2man -in docs/containers-policy.json.5.md -out docs/containers-policy.json.5
WARNING: go-md2man does not handle node type HTMLBlock
WARNING: go-md2man does not handle node type HTMLBlock
WARNING: go-md2man does not handle node type HTMLBlock

/usr/bin/go-md2man -in docs/containers-registries.d.5.md -out docs/containers-registries.d.5
WARNING: go-md2man does not handle node type HTMLBlock

/usr/bin/go-md2man -in docs/containers-transports.5.md -out docs/containers-transports.5
WARNING: go-md2man does not handle node type HTMLBlock
WARNING: go-md2man does not handle node type HTMLBlock
```

mtrmac:
go-md2man 2.0.2 (not 1.0.10) does not recognize HTML comments, triggering a warning.

Work around that, just to have a silent build.

This uses a "comment" syntax by defining a reference-style link ( https://stackoverflow.com/a/20885980 ).

Alternatively, go-md2man _does_ recognize comments inline inside a paragraph, so preceding the HTML comment with a space would probably work as well.

Either way is _weird_, and we can't add a comment to explain the comment mechanism... so use the more visually distinct alternative to hopefully minimize regressions; using a leading space would tempt future readers to just remove the space, thinking it's a mistake.

Fixes #2128 .

### DIFF
--- a/docs/containers-policy.json.5.md
+++ b/docs/containers-policy.json.5.md
@@ -58,7 +58,8 @@ This is expressed in JSON using the top-level syntax
 The global `default` set of policy requirements is mandatory; all of the other fields
 (`transports` itself, any specific transport, the transport-specific default, etc.) are optional.
 
-<!-- NOTE: Keep this in sync with transports/transports.go! -->
+[comment]: # (NOTE: Keep this in sync with transports/transports.go!)
+
 ## Supported transports and their scopes
 
 See containers-transports(5) for general documentation about the transports and their reference syntax.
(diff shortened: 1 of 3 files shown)


rahilarious (approved):
That was a rapid one. Awesome. LGTM

TomSweeneyRedHat:
LGTM
and indeed weird

mtrmac:
On… second… thought, alternatively, we could actually fix `go-md2man`. It can’t be difficult; the downside is that we would need to wait for a new release, which is not too bad.

vrothberg (approved):
LGTM

vrothberg:
Both options are OK to me

mtrmac:
Yeah, let’s fix this at the source instead: https://github.com/cpuguy83/go-md2man/pull/104 .

PR #2147, opened 2023-10-14
Fix and simplify storage tests

### LINKED ISSUE #1729: storage tests are mostly broken when run as root
… because, at least, after 2f8137261,  `makeLayer` deterministically deadlocks.

It’s been broken for 3 years. Either make the tests used (ideally by avoiding the root requirement, so that they are run during ordinary development; or at least run them in CI, or give up and drop them.

mtrmac:
For CRI-O (and, maybe, Podman/libimate), I want to introduce a new reference lookup method, and that needs tests against a locally-created image.

It turns out those tests required root, and we never run them as root. So a large part of storage transport tests have been unused for _4 years_ now.

This fixes that problem:
Fixes #1729 . And at least on a Mac (?), the tests don’t actually require any privileges, so run them unconditionally.

This PR also consolidates the existing code to create/test images to significantly reduce duplication, to benefit from `testify`, and to make it easier to add more test cases, now that I have been reading the code.

@vrothberg PTAL. Cc: @saschagrunert

### DIFF
--- a/storage/storage_test.go
+++ b/storage/storage_test.go
@@ -193,20 +193,20 @@ func systemContext() *types.SystemContext {
 // makeLayerGoroutine writes to pwriter, and on success, updates uncompressedCount
 // before it terminates.
 func makeLayerGoroutine(pwriter io.Writer, uncompressedCount *int64, compression archive.Compression) error {
-	var cwriter io.WriteCloser
 	var uncompressed *ioutils.WriteCounter
-	var twriter *tar.Writer
 	if compression != archive.Uncompressed {
 		compressor, err := archive.CompressStream(pwriter, compression)
 		if err != nil {
 			return fmt.Errorf("compressing layer: %w", err)
 		}
-		cwriter = compressor
-		uncompressed = ioutils.NewWriteCounter(cwriter)
+		defer compressor.Close()
+		uncompressed = ioutils.NewWriteCounter(compressor)
 	} else {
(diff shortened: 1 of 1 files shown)


mtrmac:
>  And at least on a Mac (?), the tests don’t actually require any privileges, so run them unconditionally.

Well, so much for that part. To be continued.

vrothberg (commented):
Changes LGTM

saschagrunert (approved):

rhatdan:
LGTM

TomSweeneyRedHat:
Changes LGTM, but there are a number of "Operation not Permited" errors.

mtrmac:
I think https://github.com/containers/image/pull/2147 might get us a bit closer, but ultimately I don’t think c/storage can run entirely unprivileged (without adding some more options throughout the codebase).

So I currently plan to settle with making the tests be macOS-only (where ~everything is fake, due to https://github.com/containers/storage/pull/811 ). They won’t run on CI, but at least frequently on my machine.

mtrmac:
> So I currently plan to settle with making the tests be macOS-only (where ~everything is fake, due to https://github.com/containers/storage/pull/811 ). They won’t run on CI, but at least frequently on my machine.

I have modified the PR to work that way. https://github.com/containers/storage/pull/1735 would be cleaner but is not strictly required, so I have dropped the c/storage update requirement here.

mtrmac:
Ready for review/merging again.

rhatdan:
LGTM

PR #2151, opened 2023-10-20
Fix conversion of Zstd images to non-OCI formats

### LINKED ISSUE #1260: OCI Zstd → schema2 uncompressed manifest conversion fails
This fails to run:
>skopeo copy   --insecure-policy docker://ghcr.io/stargz-containers/alpine:3.10.2-zstdchunked docker-archive:ghcr.io-stargz-containers-alpine-3.10.2-zstdchunked

> Getting image source signatures
Copying blob 0e427c87cfcb done  

> FATA[0003] Error creating an updated image manifest: Error during manifest conversion: "application/vnd.oci.image.layer.v1.tar+zstd": zstd compression is not supported for docker images

mtrmac:
For any Zstd layer, do a MIME type update (presumably converting from Zstd to something else) before we do the
format conversion, so that the format conversion doesn't fail with an unsupported MIME type.

Fixes #1260 .

### DIFF
--- a/internal/image/common_test.go
+++ b/internal/image/common_test.go
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 	"testing"
 
+	compressiontypes "github.com/containers/image/v5/pkg/compression/types"
 	"github.com/containers/image/v5/types"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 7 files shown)


rhatdan:
LGTM

PR #2153, opened 2023-10-20
Parse the body of (docker load) response to correctly handle errors

### LINKED ISSUE #955: copies to docker-daemon: fails silently when there is out of disk
When I do a copy from containers-storage: to docker-daemon: and the docker runs out of disk space, the skopeo returns with a silent failure.  There is no error message, merely the copy not occuring.

mtrmac:
Fixes #955 .

Tested manually (with a modified Skopeo which uses an invalid tag, not with a "disk full" error).

### DIFF
--- a/docker/daemon/daemon_dest.go
+++ b/docker/daemon/daemon_dest.go
@@ -2,6 +2,7 @@ package daemon
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


rhatdan:
A fairly fresh bug to fix. :^)
LGTM

TomSweeneyRedHat:
LGTM

PR #2155, opened 2023-10-23
Don't use append() on slices with unclear origin

mtrmac:
If some code creates a slice with extra capacity, and provides it to more than one user which calls append(), those two users will write conflicting data to the same backing array.

So, use `append(slices.Clone(...), ...)` whenever the origin of the array backing a slice is unclear (or even in cases where it can be traced but the code is a bit remote, making it hard to maintain).

A demonstration of the problem:

```go
package main

import (
	"fmt"
	"sync"
	"time"
)

func goroutine(wg *sync.WaitGroup, sharedSlice []int, i int) {
	defer wg.Done()

	privateSlice := append(sharedSlice, i)
	oldValue := privateSlice[len(privateSlice)-1]
	// This is already slow enough that the "full" value is frequently overwritten by another goroutine
	fmt.Printf("goroutine %d: immediate %d; full %#v@%v\n", i, oldValue, privateSlice, cap(privateSlice))

	time.Sleep(100 * time.Millisecond) // Give ample time for other goroutines to run

	newValue := privateSlice[len(privateSlice)-1]
	if newValue != i {
		panic(fmt.Sprintf("goroutine %d: immediate %d, after sleep %d; full %#v@%v", i, oldValue, newValue, privateSlice, cap(privateSlice)))
	}
}

func main() {
	sharedSlice := []int{0, 0, 0, 0}
	// Creates a backing array with extra capacity. (Alternatively, this could use the 3-argument version of make(),
	// but this demonstrates that this happens completely naturally just using append().)
	sharedSlice = append(sharedSlice, 0, 0)
	fmt.Printf("shared: %#v@%d\n", sharedSlice, cap(sharedSlice))

	wg := sync.WaitGroup{}
	for i := 0; i < 10; i++ {
		wg.Add(1)
		go goroutine(&wg, sharedSlice, i)
	}
	wg.Wait()
}
```

Previously https://github.com/containers/ocicrypt/pull/93 .

### DIFF
--- a/copy/multiple.go
+++ b/copy/multiple.go
@@ -340,7 +340,7 @@ func (c *copier) copyMultipleImages(ctx context.Context) (copiedManifest []byte,
 	if err != nil {
 		return nil, err
 	}
-	sigs = append(sigs, newSigs...)
+	sigs = append(slices.Clone(sigs), newSigs...)
 
 	c.Printf("Storing list signatures\n")
 	if err := c.dest.PutSignaturesWithFormat(ctx, sigs, nil); err != nil {
--- a/copy/single.go
+++ b/copy/single.go
@@ -277,7 +277,7 @@ func (c *copier) copySingleImage(ctx context.Context, unparsedImage *image.Unpar
 	if err != nil {
 		return copySingleImageResult{}, err
 	}
-	sigs = append(sigs, newSigs...)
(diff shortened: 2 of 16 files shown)


rhatdan:
LGTM

vrothberg (approved):
LGTM

PR #2156, opened 2023-10-24
Remove unused environment variables in Cirrus

mtrmac:
Some other containers/* repos use these values in test names; we don't, so remove them so that we don't have to worry about keeping them up to date.

Previously https://github.com/containers/image/pull/2150 ; @cevich PTAL to make sure I’m not missing any non-obvious references to these variables.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -29,12 +29,6 @@ env:
     # test changes made there, in a PR in this repository.
     SKOPEO_PR:
 
-    ####
-    #### Cache-image names to test with (double-quotes around names are critical)
-    ####
-    FEDORA_NAME: "fedora-38"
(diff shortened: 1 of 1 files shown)


.cirrus.yml
```diff
@@ -29,12 +29,6 @@ env:
     # test changes made there, in a PR in this repository.
     SKOPEO_PR:
 
-    ####
-    #### Cache-image names to test with (double-quotes around names are critical)
```
mtrmac, line 33: Should the double-quote note be preserved because it still applies to the image names below?

cevich (approved):
LGTM. Thanks.

vrothberg (approved):
LGTM

PR #2162, opened 2023-10-30
Add filepath.Clean at various path

divyansh42:
Was going through the code and I think it would be better to provide path after passing through `filepath.Clean()`. Therefore, I made the required changes.
Please let me know your thoughts on the same.

### DIFF
--- a/docker/docker_image_src.go
+++ b/docker/docker_image_src.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"path/filepath"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


rhatdan:
Thanks @divyansh42 
LGTM
@mtrmac @vrothberg PTAL

mtrmac (changes requested):
Why is this wanted?

`filepath.Clean` changes semantics of the path on Unix, so _by default_ it is undesirable.

docker/docker_image_src.go
```diff
@@ -477,7 +478,7 @@ func (s *dockerImageSource) getOneSignature(ctx context.Context, sigURL *url.URL
 	switch sigURL.Scheme {
 	case "file":
 		logrus.Debugf("Reading %s", sigURL.Path)
-		sigBlob, err := os.ReadFile(sigURL.Path)
+		sigBlob, err := os.ReadFile(filepath.Clean(sigURL.Path))
```
mtrmac, line 481: … and it is actively misleading to log one path and read another.

TomSweeneyRedHat:
I'm hesitant to add this as @mtrmac noted, it does change the path.  That might be good, might not.  However, if we decide to keep it, we should also clean the path in the log too.

PR #2164, opened 2023-10-30
Missed null check in docker_image_dest.go

bojidar-bg:
As the documentation says, "It is always OK to pass nil instead of a SystemContext."

I failed to get a suitable minimal reproduction project, but I got a panic on this line while using the library to copy images around once it decided to recompress a few layers, and I have confirmed that applying the change locally (or passing a suitable DestinationCtx to copy.Image) fixes the panic. Still, let me know if an actual reproduction would be necessary to get this merged.


flouthoc (approved):
LGTM

mtrmac (approved):
LGTM. Thanks!

PR #2170, opened 2023-11-02
Improve documentation of ResolveReference

mtrmac:
- Document the error returned if image is not found
- Also document it for the internal resolveImage, to record the commitment
- Update the deprecation comments to warn about the error return value change

Thanks to @flouthoc for demonstrating the need for this documentation.

Possible alternatives:
- Don’t change the error type; continue using `c/storage.ErrImageUnknown` for the new `ResolveReference`
- Define a new c/image error type; the current `c/image/storage.ErrNoSuchImage` is an alias for `c/storage.ErrNotAnImage`, which is really a different error cause.
- Possibly both? Define a new c/image error type, as an alias to c/storage.ErrImageUnknown.

### DIFF
--- a/storage/storage_reference.go
+++ b/storage/storage_reference.go
@@ -102,6 +102,8 @@ func multiArchImageMatchesSystemContext(store storage.Store, img *storage.Image,
 
 // Resolve the reference's name to an image ID in the store, if there's already
 // one present with the same name or ID, and return the image.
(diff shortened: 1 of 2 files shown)


flouthoc (approved):
LGTM

vrothberg (approved):
LGTM

PR #2173, opened 2023-11-03
Add DockerCompatAuthFilePath to allow login/logout to interoperate

mtrmac:
Allow the existing `pkg/docker/config` credential API to be used with Docker-compatible `config.json` files, using a new `types.SystemContext.DockerCompatAuthFilePath` option. This should help with https://github.com/containers/podman/issues/18617 ; I plan to expose this option in `c/common/pkg/auth`, and users would say something like `podman login --docker-file=~/.docker/config.json docker.io` (end-user design doc TBD).

It must be an option primarily because we want to refuse to write namespaced credentials when writing to `.docker/config.json`.

Also add interoperability testing between c/image and `docker/cli`, to the extent cheaply possible (e.g.`docker/cli`, understandably, is not structured to trigger the “write credentials” code of `login` without trying to actually login — so the tests use a somewhat lower-level API and need to duplicate some of the logic.

I was considering introducing this option with a new functional-option API instead of continuing with problematic `SystemContext`, but `pkg/docker/config` depends on `sysregistriesv2` for the credential helpers list, so that would be a rather invasive change.

---

WIP, filed for early review on the API. Cc: @vrothberg


vrothberg (approved):
Code LGTM

The plumbing into podman needs very careful consideration.  What (I think) needs to work "out of the box" on Mac is a `podman login` ... `docker-compose ...` _without_ having to specify any new flags on the CLI.

mtrmac:
At a first glance, that seems impossible - in Docker mode we must _reject_ `podman login docker.io/company`, in Podman mode we must support that.

And either way it’s not necessary for https://github.com/containers/podman/issues/18617, that one starts with an explicit `podman login --authfile ~/.docker/config.json …`

mtrmac:
Manual tests show that
- basic login/logout operation vs. `docker pull` works
- The special-case of `docker.io` seems to work as well.

And there are the unit tests in this PR. So I think this is ready for review/merging.

PR #2180, opened 2023-11-09
Add buildtags to avoid fulcio and rekor dependencies

siretart:
For situations where Fulcio and Rekor operations are not required,
this commit provides buildtags to avoid those dependencies.

Signed-off-by: Reinhard Tartler <siretart@gmail.com>

### DIFF
--- a/signature/sigstore/fulcio/no_fulcio.go
+++ b/signature/sigstore/fulcio/no_fulcio.go
@@ -1,0 +1,45 @@
+//go:build containers_image_fulcio_stub
+// +build containers_image_fulcio_stub
+
+package fulcio
+
+import (
+	"fmt"
+	"io"
+	"net/url"
+
+	"github.com/containers/image/v5/signature/sigstore/internal"
+)
+
+func WithFulcioAndPreexistingOIDCIDToken(fulcioURL *url.URL, oidcIDToken string) internal.Option {
+	return func(s *internal.SigstoreSigner) error {
+		return fmt.Errorf("Fulcio disabled at compile time")
+	}
+}
+
+// WithFulcioAndDeviceAuthorizationGrantOIDC sets up signing to use a short-lived key and a Fulcio-issued certificate
+// based on an OIDC ID token obtained using a device authorization grant (RFC 8628).
+//
+// interactiveOutput must be directly accessible to a human user in real time (i.e. not be just a log file).
+func WithFulcioAndDeviceAuthorizationGrantOIDC(fulcioURL *url.URL, oidcIssuerURL *url.URL, oidcClientID, oidcClientSecret string,
+	interactiveOutput io.Writer) internal.Option {
+	return func(s *internal.SigstoreSigner) error {
+		return fmt.Errorf("Fulcio disabled at compile time")
+	}
+}
+
+// WithFulcioAndInterativeOIDC sets up signing to use a short-lived key and a Fulcio-issued certificate
+// based on an interactively-obtained OIDC ID token.
+// The token is obtained
+//   - directly using a browser, listening on localhost, automatically opening a browser to the OIDC issuer,
+//     to be redirected on localhost. (I.e. the current environment must allow launching a browser that connect back to the current process;
+//     either or both may be impossible in a container or a remote VM).
+//   - or by instructing the user to manually open a browser, obtain the OIDC code, and interactively input it as text.
+//
+// interactiveInput and interactiveOutput must both be directly operable by a human user in real time (i.e. not be just a log file).
+func WithFulcioAndInteractiveOIDC(fulcioURL *url.URL, oidcIssuerURL *url.URL, oidcClientID, oidcClientSecret string,
+	interactiveInput io.Reader, interactiveOutput io.Writer) internal.Option {
+	return func(s *internal.SigstoreSigner) error {
+		return fmt.Errorf("Fulcio disabled at compile time")
+	}
+}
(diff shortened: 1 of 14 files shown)


siretart:
cc: @rhatdan -- as discussed at kubecon

signature/sigstore/fulcio/no_fulcio.go
```diff
@@ -0,0 +1,45 @@
...
+//   - directly using a browser, listening on localhost, automatically opening a browser to the OIDC issuer,
+//     to be redirected on localhost. (I.e. the current environment must allow launching a browser that connect back to the current process;
+//     either or both may be impossible in a container or a remote VM).
+//   - or by instructing the user to manually open a browser, obtain the OIDC code, and interactively input it as text.
+//
+// interactiveInput and interactiveOutput must both be directly operable by a human user in real time (i.e. not be just a log file).
+func WithFulcioAndInteractiveOIDC(fulcioURL *url.URL, oidcIssuerURL *url.URL, oidcClientID, oidcClientSecret string,
+	interactiveInput io.Reader, interactiveOutput io.Writer) internal.Option {
+	return func(s *internal.SigstoreSigner) error {
+		return fmt.Errorf("Fulcio disabled at compile time")
```
rhatdan, line 43: Do not capitalize errors.

signature/sigstore/rekor/no_rekor.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"fmt"
+	"net/url"
+
+	signerInternal "github.com/containers/image/v5/signature/sigstore/internal"
+)
+
+func WithRekor(rekorURL *url.URL) signerInternal.Option {
+	return func(s *signerInternal.SigstoreSigner) error {
+		return fmt.Errorf("Rekor disabled at build time")
```
rhatdan, line 15: No capitals

signature/sigstore/fulcio/no_fulcio.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+// WithFulcioAndDeviceAuthorizationGrantOIDC sets up signing to use a short-lived key and a Fulcio-issued certificate
+// based on an OIDC ID token obtained using a device authorization grant (RFC 8628).
+//
+// interactiveOutput must be directly accessible to a human user in real time (i.e. not be just a log file).
+func WithFulcioAndDeviceAuthorizationGrantOIDC(fulcioURL *url.URL, oidcIssuerURL *url.URL, oidcClientID, oidcClientSecret string,
+	interactiveOutput io.Writer) internal.Option {
+	return func(s *internal.SigstoreSigner) error {
+		return fmt.Errorf("Fulcio disabled at compile time")
```
rhatdan, line 27: No caps

signature/sigstore/fulcio/no_fulcio.go
```diff
@@ -0,0 +1,45 @@
...
+	"fmt"
+	"io"
+	"net/url"
+
+	"github.com/containers/image/v5/signature/sigstore/internal"
+)
+
+func WithFulcioAndPreexistingOIDCIDToken(fulcioURL *url.URL, oidcIDToken string) internal.Option {
+	return func(s *internal.SigstoreSigner) error {
+		return fmt.Errorf("Fulcio disabled at compile time")
```
rhatdan, line 16: No caps

rhatdan:
LGTM other then nits.
@mtrmac @vrothberg PTAL

signature/sigstore/fulcio/no_fulcio.go
mtrmac: These are all project names.

signature/sigstore/fulcio/no_fulcio.go
```diff
@@ -0,0 +1,45 @@
+//go:build containers_image_fulcio_stub
```
mtrmac, line 1: Absolutely non-blocking: I would weakly prefer these files to be named `fulcio_stub.go` and `rekor_stub.go`.

mtrmac (commented):
I’m fine with conditionalizing this, but note that this does not actually remove those dependencies as documented: Both Fulcio and Rekor are required by the signature verification paths.

So, is this actually useful for your goals? If it is, fine, it’s just that the build tag documentation needs to be clear that this only disables the signing path, not verification.

(Removing/conditionalizing that code from the verification paths would require a pretty good unit test coverage, so that would be rather more work.)

siretart:
@mtrmac thanks for pointing our my oversight. The issue at core is that I don't have neither `sigstore/rekor` nor `sigstore/fulcio` available in Debian. I looked at packagig them, but they are just too "heavy" and I can't afford the time to package them one-by-one.

I've updated the PR to disable the verificaion paths as well. Can you please elaborate on:

> (Removing/conditionalizing that code from the verification paths would require a pretty good unit test coverage, so that would be rather more work.)

What's wrong with just removing those unit tests?

mtrmac:
The policy enforcement / signature verification code is a critical security mechanism; so I think it should be tested. It’s fine not to support features, and to fail closed; but then we should have tests that the code _does_ fail closed.

I appreciate that the added implementation is trivial, but still… quite a bit of `policy_eval_sigstore.go` will remain reachable, so wholesale removing the tests is not ideal.

It‘s also somewhat a matter of not relaxing expectations: “the policy enforcement package code must have good test coverage”, and keeping that true will help reinforce that for other future features.

PR #2186, opened 2023-11-13
Quote the response body in an error message

mtrmac:
... to make sure we don't output raw control characters, for example.

A probably unexpected outcome of https://github.com/containers/podman/issues/20343 .


rhatdan:
LGTM
@vrothberg PTAL

vrothberg (approved):
LGTM

PR #2201, opened 2023-11-28
Add isKubeHealthCheck field to HealthConfig

umohnani8:
Add a new isKubeHealthCheck field to schema2.HealthConfig to keep track of whether the health check is from a kube yaml. If that is the case then the health check needs to behave a bit differently to match k8s behaviour.


umohnani8:
This new field is needed for https://github.com/containers/podman/pull/20789

vrothberg (commented):
Drive-by-comment (feel free to ignore): I think this information can be stored on the podman side.  Maybe an annotation for the specific container?

umohnani8:
That makes sense! Thank you! 
This is not needed anymore, closing.

PR #2202, opened 2023-11-28
Don't expose local account details in oci-archive tar files

### LINKED ISSUE #1627: Skopeo leaks details about the host in oci-archives
When running 

	skopeo copy docker://docker.io/library/ubuntu@sha256:b2339eee806d44d6a8adc0a790f824fb71f03366dd754d400316ae5a7e3ece3e oci-archive:/tmp/x.tar

the resulting /tmp/x.tar contains references to the current user. In particular
it seems that uname and gname are set to the current user and group. uid and
gid are set as well (I think they should simply be 0 since they are not
relevant on this level in the oci-archive). I don't think there is a good
reason to leak this information about the host running the skopeo-copy.

I think the same happening even with the layers inside the oci-archive when
doing podman save, but I assume root cause will be the same so I'm not filling
that separately (I can if it is desired).

mtrmac:
Fixes #1627 .

<s>Requires unmerged https://github.com/containers/storage/pull/1765 .</s>

### DIFF
--- a/oci/archive/oci_dest.go
+++ b/oci/archive/oci_dest.go
@@ -173,6 +173,7 @@ func tarDirectory(src, dst string) error {
 	if err != nil {
 		return fmt.Errorf("retrieving stream of bytes from %q: %w", src, err)
 	}
+	defer input.Close()
 
 	// creates the tar file
 	outFile, err := os.Create(dst)
--- a/oci/archive/oci_dest.go
+++ b/oci/archive/oci_dest.go
@@ -13,6 +13,7 @@ import (
 	"github.com/containers/image/v5/internal/signature"
 	"github.com/containers/image/v5/types"
 	"github.com/containers/storage/pkg/archive"
+	"github.com/containers/storage/pkg/idtools"
 	digest "github.com/opencontainers/go-digest"
 	"github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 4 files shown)


oci/archive/oci_dest.go
```diff
@@ -169,7 +170,11 @@ func (d *ociArchiveImageDestination) Commit(ctx context.Context, unparsedTopleve
 // tar converts the directory at src and saves it to dst
 func tarDirectory(src, dst string) error {
 	// input is a stream of bytes from the archive of the directory at path
-	input, err := archive.Tar(src, archive.Uncompressed)
+	input, err := archive.TarWithOptions(src, &archive.TarOptions{
+		Compression: archive.Uncompressed,
+		// Don’t include the data about the user account this code is running under.
+		ChownOpts: &idtools.IDPair{UID: 0, GID: 0},
+	})
```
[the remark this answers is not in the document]
mtrmac: Thanks, fixed now. (Conceptually separate from this PR, but I don’t think splitting that would really be worth it.)

mtrmac:
… and updated to include the merged c/storage PR instead of a fork; ready for review.

oci/archive/oci_dest_test.go
```diff
@@ -1,5 +1,45 @@
...
+		hdr, err := reader.Next()
+		if err == io.EOF {
+			break
+		}
+		require.NoError(t, err)
+		// Test that the header does not expose data about the local account
+		assert.Equal(t, 0, hdr.Uid)
+		assert.Equal(t, 0, hdr.Gid)
+		assert.Equal(t, "", hdr.Uname)
+		assert.Equal(t, "", hdr.Gname)
```
nalind, line 41: The error message that `assert.Empty()` would produce _might_ make it preferable over string comparisons, but it also LGTM as-is.

mtrmac: Thanks, updated.

rhatdan:
LGTM

PR #2204, opened 2023-11-30
Trigger a conversion to OCI when compressing to Zstd

### LINKED ISSUE #2195: Copies with Zstd compression to schema-agnostic transports don’t trigger schema conversion
E.g.
```console
skopeo copy --dest-compress --dest-compress-format=zstd dir:uncompressed-schema2 dir:dest
```
fails with `FATA[0000] creating an updated image manifest: preparing updated manifest, layer "sha256:…": zstd compression is not supported for type "application/vnd.docker.image.rootfs.diff.tar"`

because `determineManifestConversion` sees that all formats are accepted, and thinks that there is no reason to try anything but the source format.

mtrmac:
... even if the destination does not restrict manifest schemas.

Fixes #2195.

### DIFF
--- a/copy/manifest.go
+++ b/copy/manifest.go
@@ -6,8 +6,10 @@ import (
 	"fmt"
 	"strings"
 
+	internalManifest "github.com/containers/image/v5/internal/manifest"
 	"github.com/containers/image/v5/internal/set"
 	"github.com/containers/image/v5/manifest"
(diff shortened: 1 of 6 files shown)


rhatdan:
LGTM
@giuseppe PTAL

giuseppe (approved):
/lgtm

PR #2213, opened 2023-12-06
Make blob reuse choices manifest-format-sensitive, and allow conversions when writing to format-agnostic transports

### LINKED ISSUE #2205: Blob reuse decisions do not take into account manifest support
E.g. when pushing a schema2 image, `dockerImageDestination.TryReusingBlobWithOptions` can find and choose to use a Zstd layer
- That directly breaks if the user forced schema1/schema2 manifest type
- Even without `copy.Options.ForceManifestMIMEType`, if the destination is manifest-schema-agnostic, and `determineManifestConversion` has no reason to expect Zstd to be involved, the copy code will only have a single candidate; we need to _either_ add an OCI manifest schema candidate, _or_ not reuse Zstd layers if the copy is not set up to use OCI.

mtrmac:
Fixes #2205

### DIFF
--- a/copy/compression.go
+++ b/copy/compression.go
@@ -286,7 +286,7 @@ func (d *bpCompressionStepData) recordValidatedDigestData(c *copier, uploadedInf
 	if d.uploadedCompressorName != "" && d.uploadedCompressorName != internalblobinfocache.UnknownCompression {
 		if d.uploadedCompressorName != compressiontypes.ZstdChunkedAlgorithmName {
 			// HACK: Don’t record zstd:chunked algorithms.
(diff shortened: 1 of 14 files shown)


mtrmac:
Now tested (with https://github.com/containers/image/issues/2205#issuecomment-1918096877 ), and I fixed some of the debug log data.

Ready for review/merging.

rhatdan:
@giuseppe PTAL

giuseppe (approved):
LGTM

giuseppe:
@mtrmac can we merge it?

PR #2214, opened 2023-12-06
Embrace codespell: config, workflow (to alert when new typos added) and get typos fixed

yarikoptic:
I moved some configs already present in invocation within Makefile but I guess it was not used regularly.  With the github workflow -- it would not let new typos to creep in

### DIFF
--- a/internal/manifest/oci_index_test.go
+++ b/internal/manifest/oci_index_test.go
@@ -177,13 +177,13 @@ func TestOCI1IndexChooseInstanceByCompression(t *testing.T) {
 				{"amd64", "", "sha256:bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb", false},
 				// out of multiple gzip in arm64 select the first one to ensure original logic is prevented
 				{"arm64", "", "sha256:cccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc", false},
-				// select a signle gzip s390x image
+				// select a single gzip s390x image
 				{"s390x", "", "sha256:eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee", false},
 				// out of gzip and zstd in amd64 select the first gzip image
 				{"amd64", "", "sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa", true},
 				// out of multiple gzip in arm64 select the first one to ensure original logic is prevented
 				{"arm64", "", "sha256:cccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc", true},
-				// select a signle gzip s390x image
+				// select a single gzip s390x image
 				{"s390x", "", "sha256:eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee", true},
 			},
 			unmatchedInstances: []string{
--- a/signature/simplesigning/signer_test.go
+++ b/signature/simplesigning/signer_test.go
@@ -103,7 +103,7 @@ func TestSimpleSignerSignImageManifest(t *testing.T) {
 
 	// Failures to sign need to be tested in two parts: First the failures that involve the wrong passphrase, then failures that
 	// should manifest even with a valid passphrase or unlocked key (because the GPG agent is caching unlocked keys).
-	// Alternatively, we could be caling gpgagent.KillGPGAgent() all the time...
+	// Alternatively, we could be calling gpgagent.KillGPGAgent() all the time...
 	type failingCase struct {
 		name string
 		opts []Option
--- a/storage/storage_test.go
+++ b/storage/storage_test.go
@@ -207,7 +207,7 @@ func makeLayerGoroutine(pwriter io.Writer, uncompressedCount *int64, compression
 	// We do not do that, to workaround https://github.com/containers/storage/issues/1729 :
 	// tar-split runs a goroutine that consumes/forwards tar content and might access
 	// concurrently-freed objects if it sees a valid EOF marker.
-	// Instead, realy on raw EOF to terminate the goroutine.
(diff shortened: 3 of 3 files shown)


mtrmac (commented):
Thanks!

@cevich WDYT?

.github/workflows/codespell.yml
```diff
@@ -0,0 +1,22 @@
+---
+name: Codespell
+
+on:
+  push:
+    branches: [main]
```
cevich, line 6: I can't think of a reason this needs to run on pushes to `main` or any other branch.  Is there some other consideration for this?

.github/workflows/codespell.yml
```diff
@@ -0,0 +1,22 @@
...
+  pull_request:
+    branches: [main]
```
cevich, line 8: This is probably fine.  It would be nice to include release branches, but should ignore renovate PR branches.  So keeping it simple like this is probably fine.

.codespellrc
```diff
@@ -0,0 +1,5 @@
+[codespell]
```
cevich, line 1: Would you consider adding a comment in this file w/ link to docs for its syntax and format?

.github/workflows/codespell.yml
```diff
@@ -0,0 +1,22 @@
...
+
+jobs:
+  codespell:
+    name: Check for spelling errors
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Codespell
```
cevich, line 21: Would you consider adding a comment here, reminding a developer that the configuration is in the `.codespellrc` file?

This may be me being pedantic, so I won't block if you disagree / feel it's excessive.

cevich (approved):
LGTM.  I did not try running the make target.

.codespellrc
yarikoptic: Something like

```suggestion
# See https://github.com/codespell-project/codespell#using-a-config-file
[codespell]
```

?

.github/workflows/codespell.yml
yarikoptic: it is common for projects to run CI upon changes to be about to be contributed (PRs) or directly pushed (thus on branches, in particular `main`).  This allows to identify which changes introduced the problem (unlike on schedule -- might need to look for range of changes).  If you want some other way - I guess could be changed here.

.github/workflows/codespell.yml
yarikoptic:
```suggestion
      - name: Codespell
      # uses configuration within .codespellrc file
```

.codespellrc
cevich: Perfect.  Thanks.

cevich (changes requested):
Changes LGTM.  There's a bunch of missing signatures on commits that need fixing (i.e. "Signed-off-by: ..." footers).  Also I think you can probably squash the commits down into maybe two or three?  If you're able to address the main-branch push condition great, but don't spend too much time on it.  Thanks a lot for putting in all this work :smiley:

.github/workflows/codespell.yml
yarikoptic: FWIW, I have been quite an abuser of github actions (see mass use and dashboards in e.g. https://github.com/datalad/datalad-extensions or https://github.com/datalad/git-annex )... this workflow is such a quick/tiny one that I would have not worried about it making any notable dent but I could be proven wrong.

> the workflow is either running on a fork, or a PR and skip it for everything else.

to a degree it does that already -- as merge of a PR creates a new commit, which might be different from "perspective" (those always differ from those actually merged) merge commit which github prepared during PR, then it might run it on an identical state to a PR... yeah, not sure how to formulate for github smth like "if you checked already the state in PR perspective merge do not check after  merge if the same state".  may be @jwodder knows?

cevich: Okay, we can drop my concern then.  Thanks for trying.

cevich (commented):
it appears like the codepsell comments re: format and config file, got lost somehow.

yarikoptic:
oops, I guess we never committed or I forgot during rebase/force push... redid manually now -- should b e in the diff. Still 4 commits total ;-)

cevich:
LGTM

@mtrmac PTAL.

rhatdan:
LGTM

rhatdan:
Thanks @yarikoptic

.github/workflows/codespell.yml
mtrmac: Returning to this after all this time — as a repo maintainer, my question is “does running this on commits to `main` make any difference at all”?

Realistically I am not going to go manually looking for any workflow results. I think if it doesn’t block work that needs doing, and no-one is specifically looking for failures, the failures, if any, are going to be ignored; so it seems to me that we really don’t need to run the checks at all.

@cevich perhaps it would make sense to run the `codespell` during the daily cirrus-cron CI runs, which we do follow. (Or, alternatively and purely hypothetically, replace the daily cirrus-cron CI runs by whatever it is where the GitHub workflow run on `main` commits shows up — the point is that both should ideally show up in only one shared place.)


---

Details: How is this test ever going to break if it wasn’t broken on a PR?
- It could be broken in a PR and manually merged. Too bad, that needs to be possible at least for repo admins. We would find out on the next PR.
- It could be something committed without a PR. Maybe that might not need to be possible, the branch can be configured to require PRs. But then again, repo admins could override.
- It seems to me that by far the most likely way for this test to break is if the dictionary of `codespell` changed (it does change every few days). In that case, on _past release branches_, I think we actively don’t want to change code; so we wouldn’t want to run codespell there, at all. We do want to notice such breakage on `main`.

cevich: @mtrmac github actions is stupid, it can't tell the difference between main in **this** repo, and a fork of this repo.  There's (unfortunately) a lot of users that will make changes on their main branch which end up in PRs upstream.  Having this run on `main` means they will get feedback BEFORE creating a PR.  Sometimes this is helpful.  Most of the time it confuses the hell out of maintainers (like it just did :wink:)

I'm not concerned about this running on the upstream main.  I think we can just ignore it.  If that turns into a problem, I can easily add an "on failure" notification e-mail to the workflow.

You're catch-in-from-cirrus-cron idea would work as well, but that would necessitate migrating the workflow into the cirrus YAML.  So, slightly more work.

PR #2215, opened 2023-12-07
Moving policy.json and default.yaml from containers/skopeo

rahilarious:
It makes more sense to keep these 2 files along with their man pages...in c/image

As discussed in
https://github.com/containers/common/pull/1757

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,13 @@ GOMD2MAN ?= $(shell command -v go-md2man || echo '$(GOBIN)/go-md2man')
 MANPAGES_MD = $(wildcard docs/*.5.md)
 MANPAGES ?= $(MANPAGES_MD:%.md=%)
 
+ifeq ($(shell uname -s),FreeBSD)
+CONTAINERSCONFDIR ?= /usr/local/etc/containers
+else
+CONTAINERSCONFDIR ?= /etc/containers
+endif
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM

mtrmac:
I have no opinion on the location of these files, I’ll fully defer to our packaging experts (and this _must_ be coordinated with them).

@lsm5 @jnovy PTAL.

Note https://github.com/containers/common/pull/1757 for previous discussion, which did lean towards a consensus for moving them to c/image.

---

Implementation LGTM in principle, but tests fail.

rahilarious:
> Implementation LGTM in principle, but tests fail.

Fixed

rhatdan:
The update.sh script in containers-common package will need to be updated.

jnovy:
c/image is definitely better place. Please let me know once it's there so that I can update the update.sh script. Thanks!

rahilarious:
> c/image is definitely better place. Please let me know once it's there so that I can update the update.sh script. Thanks!

It is already there?

jnovy:
Ok, actually I don't need any changes to update.sh as we have some custom RHEL configurations related to GPG keys so the default-policy.json can;t be consumed as-is neither from skopeo or c/image.

mtrmac:
@jnovy FWIW this also moves `registries.d/default.yaml`.

PR #2218, opened 2023-12-11
storage, dest: clarify when TOCDigest is used

giuseppe:
This update introduces an enhancement in the blob handling mechanism, specifically by separating the TOC digest from the uncompressed/compressed digest.

Follow-up for: #1080.

### DIFF
--- a/copy/single.go
+++ b/copy/single.go
@@ -746,7 +746,7 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, to
 				wrapped: ic.c.rawSource,
 				bar:     bar,
 			}
-			uploadedBlob, err := ic.c.dest.PutBlobPartial(ctx, &proxy, srcInfo, ic.c.blobInfoCache)
+			uploadedBlob, err := ic.c.dest.PutBlobPartial(ctx, &proxy, srcInfo, &layerIndex, ic.c.blobInfoCache)
 			if err == nil {
 				if srcInfo.Size != -1 {
 					refill := srcInfo.Size - bar.Current()
--- a/copy/single.go
+++ b/copy/single.go
@@ -690,10 +690,15 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, to
 			requiredCompression = ic.compressionFormat
 		}
 
+		var tocDigest digest.Digest
+
 		// Check if we have a chunked layer in storage that's based on that blob.  These layers are stored by their TOC digest.
-		tocDigest, err := chunkedToc.GetTOCDigest(srcInfo.Annotations)
+		d, err := chunkedToc.GetTOCDigest(srcInfo.Annotations)
 		if err != nil {
 			return types.BlobInfo{}, "", err
+		}
+		if d != nil {
+			tocDigest = *d
 		}
 
 		reused, reusedBlob, err := ic.c.dest.TryReusingBlobWithOptions(ctx, srcInfo, private.TryReusingBlobOptions{
@@ -713,7 +718,12 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, to
 		if reused {
 			logrus.Debugf("Skipping blob %s (already present):", srcInfo.Digest)
 			func() { // A scope for defer
-				bar := ic.c.createProgressBar(pool, false, types.BlobInfo{Digest: reusedBlob.Digest, Size: 0}, "blob", "skipped: already exists")
+				d := reusedBlob.Digest
+				label := "skipped: already exists"
+				if reusedBlob.TOCDigest != "" {
+					label = "skipped: already exists (found by TOC)"
+				}
+				bar := ic.c.createProgressBar(pool, false, types.BlobInfo{Digest: d, Size: 0}, "blob", label)
 				defer bar.Abort(false)
 				bar.mark100PercentComplete()
 			}()
--- a/internal/private/private.go
+++ b/internal/private/private.go
@@ -81,8 +81,9 @@ type ImageDestination interface {
 // UploadedBlob is information about a blob written to a destination.
 // It is the subset of types.BlobInfo fields the transport is responsible for setting; all fields must be provided.
 type UploadedBlob struct {
-	Digest digest.Digest
-	Size   int64
+	Digest    digest.Digest
+	Size      int64
+	TOCDigest digest.Digest
 }
 
 // PutBlobOptions are used in PutBlobWithOptions.
@@ -118,14 +119,15 @@ type TryReusingBlobOptions struct {
 	PossibleManifestFormats []string               // A set of possible manifest formats; at least one should support the reused layer blob.
 	RequiredCompression     *compression.Algorithm // If set, reuse blobs with a matching algorithm as per implementations in internal/imagedestination/impl.helpers.go
 	OriginalCompression     *compression.Algorithm // May be nil to indicate “uncompressed” or “unknown”.
-	TOCDigest               *digest.Digest         // If specified, the blob can be looked up in the destination also by its TOC digest.
+	TOCDigest               digest.Digest          // If specified, the blob can be looked up in the destination also by its TOC digest.
 }
 
 // ReusedBlob is information about a blob reused in a destination.
 // It is the subset of types.BlobInfo fields the transport is responsible for setting.
 type ReusedBlob struct {
-	Digest digest.Digest // Must be provided
-	Size   int64         // Must be provided
+	Digest    digest.Digest // Must be provided, can be empty if TOCDigest is present
+	TOCDigest digest.Digest // Must be provided, can be empty if Digest is present
+	Size      int64         // Must be provided
 	// The following compression fields should be set when the reuse substitutes
 	// a differently-compressed blob.
 	CompressionOperation types.LayerCompression // Compress/Decompress, matching the reused blob; PreserveOriginal if N/A
--- a/internal/private/private.go
+++ b/internal/private/private.go
@@ -55,7 +55,7 @@ type ImageDestinationInternalOnly interface {
 	// It is available only if SupportsPutBlobPartial().
 	// Even if SupportsPutBlobPartial() returns true, the call can fail, in which case the caller
 	// should fall back to PutBlobWithOptions.
-	PutBlobPartial(ctx context.Context, chunkAccessor BlobChunkAccessor, srcInfo types.BlobInfo, cache blobinfocache.BlobInfoCache2) (UploadedBlob, error)
+	PutBlobPartial(ctx context.Context, chunkAccessor BlobChunkAccessor, srcInfo types.BlobInfo, index *int, cache blobinfocache.BlobInfoCache2) (UploadedBlob, error)
 
 	// TryReusingBlobWithOptions checks whether the transport already contains, or can efficiently reuse, a blob, and if so, applies it to the current destination
 	// (e.g. if the blob is a filesystem layer, this signifies that the changes it describes need to be applied again when composing a filesystem tree).
--- a/storage/storage_dest.go
+++ b/storage/storage_dest.go
@@ -269,7 +269,7 @@ func (f *zstdFetcher) GetBlobAt(chunks []chunked.ImageSourceChunk) (chan io.Read
 // It is available only if SupportsPutBlobPartial().
 // Even if SupportsPutBlobPartial() returns true, the call can fail, in which case the caller
 // should fall back to PutBlobWithOptions.
-func (s *storageImageDestination) PutBlobPartial(ctx context.Context, chunkAccessor private.BlobChunkAccessor, srcInfo types.BlobInfo, cache blobinfocache.BlobInfoCache2) (private.UploadedBlob, error) {
+func (s *storageImageDestination) PutBlobPartial(ctx context.Context, chunkAccessor private.BlobChunkAccessor, srcInfo types.BlobInfo, index *int, cache blobinfocache.BlobInfoCache2) (private.UploadedBlob, error) {
 	fetcher := zstdFetcher{
 		chunkAccessor: chunkAccessor,
 		ctx:           ctx,
--- a/storage/storage_dest.go
+++ b/storage/storage_dest.go
@@ -77,19 +77,21 @@ type storageImageDestination struct {
 	indexToStorageID map[int]*string
 	// All accesses to below data are protected by `lock` which is made
 	// *explicit* in the code.
-	uncompressedOrTocDigest map[digest.Digest]digest.Digest                       // Mapping from layer blobsums to their corresponding DiffIDs or TOC IDs.
-	fileSizes               map[digest.Digest]int64                               // Mapping from layer blobsums to their sizes
-	filenames               map[digest.Digest]string                              // Mapping from layer blobsums to names of files we used to hold them
-	currentIndex            int                                                   // The index of the layer to be committed (i.e., lower indices have already been committed)
-	indexToAddedLayerInfo   map[int]addedLayerInfo                                // Mapping from layer (by index) to blob to add to the image
-	blobAdditionalLayer     map[digest.Digest]storage.AdditionalLayer             // Mapping from layer blobsums to their corresponding additional layer
-	diffOutputs             map[digest.Digest]*graphdriver.DriverWithDifferOutput // Mapping from digest to differ output
+	blobDiffIDs           map[digest.Digest]digest.Digest                       // Mapping from layer blobsums to their corresponding DiffIDs
+	indexToTocDigest      map[int]digest.Digest                                 // Mapping from layer index to their corresponding TOC Digest, if used
+	fileSizes             map[digest.Digest]int64                               // Mapping from layer blobsums to their sizes
+	filenames             map[digest.Digest]string                              // Mapping from layer blobsums to names of files we used to hold them
+	currentIndex          int                                                   // The index of the layer to be committed (i.e., lower indices have already been committed)
+	indexToAddedLayerInfo map[int]addedLayerInfo                                // Mapping from layer (by index) to blob to add to the image
+	blobAdditionalLayer   map[digest.Digest]storage.AdditionalLayer             // Mapping from layer blobsums to their corresponding additional layer
+	diffOutputs           map[digest.Digest]*graphdriver.DriverWithDifferOutput // Mapping from digest to differ output
 }
 
 // addedLayerInfo records data about a layer to use in this image.
 type addedLayerInfo struct {
-	digest     digest.Digest
-	emptyLayer bool // The layer is an “empty”/“throwaway” one, and may or may not be physically represented in various transport / storage systems.  false if the manifest type does not have the concept.
+	digest     digest.Digest // Mandatory, the digest of the layer.
+	tocDigest  digest.Digest // Optional.  If tocDigest != "", the layer was pulled by its TOC.
+	emptyLayer bool          // The layer is an “empty”/“throwaway” one, and may or may not be physically represented in various transport / storage systems.  false if the manifest type does not have the concept.
 }
 
 // newImageDestination sets us up to write a new image, caching blobs in a temporary directory until
@@ -117,18 +119,19 @@ func newImageDestination(sys *types.SystemContext, imageRef storageReference) (*
 			HasThreadSafePutBlob:           true,
 		}),
 
-		imageRef:                imageRef,
-		directory:               directory,
-		signatureses:            make(map[digest.Digest][]byte),
-		uncompressedOrTocDigest: make(map[digest.Digest]digest.Digest),
-		blobAdditionalLayer:     make(map[digest.Digest]storage.AdditionalLayer),
-		fileSizes:               make(map[digest.Digest]int64),
-		filenames:               make(map[digest.Digest]string),
-		SignatureSizes:          []int{},
-		SignaturesSizes:         make(map[digest.Digest][]int),
-		indexToStorageID:        make(map[int]*string),
-		indexToAddedLayerInfo:   make(map[int]addedLayerInfo),
-		diffOutputs:             make(map[digest.Digest]*graphdriver.DriverWithDifferOutput),
+		imageRef:              imageRef,
+		directory:             directory,
+		signatureses:          make(map[digest.Digest][]byte),
+		blobDiffIDs:           make(map[digest.Digest]digest.Digest),
+		indexToTocDigest:      make(map[int]digest.Digest),
+		blobAdditionalLayer:   make(map[digest.Digest]storage.AdditionalLayer),
+		fileSizes:             make(map[digest.Digest]int64),
+		filenames:             make(map[digest.Digest]string),
+		SignatureSizes:        []int{},
+		SignaturesSizes:       make(map[digest.Digest][]int),
+		indexToStorageID:      make(map[int]*string),
+		indexToAddedLayerInfo: make(map[int]addedLayerInfo),
+		diffOutputs:           make(map[digest.Digest]*graphdriver.DriverWithDifferOutput),
 	}
 	dest.Compat = impl.AddCompat(dest)
 	return dest, nil
@@ -227,7 +230,7 @@ func (s *storageImageDestination) putBlobToPendingFile(stream io.Reader, blobinf
 
 	// Record information about the blob.
 	s.lock.Lock()
-	s.uncompressedOrTocDigest[blobDigest] = diffID.Digest()
+	s.blobDiffIDs[blobDigest] = diffID.Digest()
 	s.fileSizes[blobDigest] = counter.Count
 	s.filenames[blobDigest] = filename
 	s.lock.Unlock()
@@ -286,13 +289,19 @@ func (s *storageImageDestination) PutBlobPartial(ctx context.Context, chunkAcces
 		return private.UploadedBlob{}, err
 	}
 
+	if out.TOCDigest == "" {
+		return private.UploadedBlob{}, errors.New("TOC digest is empty")
+	}
+
 	blobDigest := srcInfo.Digest
 
 	s.lock.Lock()
-	s.uncompressedOrTocDigest[blobDigest] = blobDigest
 	s.fileSizes[blobDigest] = 0
 	s.filenames[blobDigest] = ""
 	s.diffOutputs[blobDigest] = out
+	if index != nil {
+		s.indexToTocDigest[*index] = out.TOCDigest
+	}
 	s.lock.Unlock()
 
 	return private.UploadedBlob{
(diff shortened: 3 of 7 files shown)


copy/single.go
```diff
@@ -718,7 +723,11 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, to
 		if reused {
 			logrus.Debugf("Skipping blob %s (already present):", srcInfo.Digest)
 			func() { // A scope for defer
-				bar := ic.c.createProgressBar(pool, false, types.BlobInfo{Digest: reusedBlob.Digest, Size: 0}, "blob", "skipped: already exists")
+				d := reusedBlob.Digest
+				if d == "" {
+					d = reusedBlob.TOCDigest
+				}
+				bar := ic.c.createProgressBar(pool, false, types.BlobInfo{Digest: d, Size: 0}, "blob", "skipped: already exists")
```
mtrmac, line 730: Non-blocking:
Either you think the users don’t really care, in which case this can continue to just use `Digest` (and users don’t need to deal with the difference — maybe it wouldn’t even be too crazy to just use layer numbers instead of using any digests at all…); or you think the difference matters, in which case the TOC value should be clearly distinct, maybe using some special metacharacter, or explicitly saying “TOC $value”.

Strictly speaking, here it’s a mostly aesthetic concern about not getting into the _habit_ of not knowing what value is carried in which variable… and a bit of risk in that the current UI-only code could turn into a non-UI code after a few years.

types/types.go
```diff
@@ -121,6 +121,7 @@ const (
 // In some situations, some fields may be unknown, in others they may be mandatory; documenting an “unknown” value here does not override that.
 type BlobInfo struct {
 	Digest      digest.Digest // "" if unknown.
+	TOCDigest   digest.Digest // "" if unknown.
```
mtrmac, line 124: From a quick check, I don’t see any users; but we _do_ have locations (like `TryReusingBlobWithOptions`) where `Digest` is documented to be required… so I at a first glance I don’t see that the way `tryReusingBlobAsPending` is setting only `TOCDigest` is sufficient.

That would be some kind of large-scale overhaul… similarly to `c/storage.Layer.UncompressedDigest`, but probably worse.

internal/private/private.go
```diff
@@ -81,8 +81,9 @@ type ImageDestination interface {
 // UploadedBlob is information about a blob written to a destination.
 // It is the subset of types.BlobInfo fields the transport is responsible for setting; all fields must be provided.
 type UploadedBlob struct {
-	Digest digest.Digest
-	Size   int64
+	Digest    digest.Digest
+	Size      int64
+	TOCDigest digest.Digest
```
mtrmac, line 86: No transport is setting this?! And the transport wouldn’t know, after all.

I think if _anything_ should be managing this value:
- it can come from the source, if we don’t modify the blob
- it can be computed while compressing the blob to zstd:chunked, as a part of the copy
- it can be _discarded_ while decompressing a zstd:chunked blob, as part of the copy

so it initially seems to me that with `UploadedBlob`, the destination has no business providing/obtaining the value. (With `TryReuseBlob`, the case for the destination transport providing data is somewhat stronger.)

Especially if the digest ever became a part of the signature, we shouldn’t trust a `TOCDigest` value that some remote server, or something like that, claims, without validating it ourselves. Compare `canSubstituteBlobs`.

internal/private/private.go
```diff
@@ -98,6 +99,8 @@ type PutBlobOptions struct {
 
 	EmptyLayer bool // True if the blob is an "empty"/"throwaway" layer, and may not necessarily be physically represented.
 	LayerIndex *int // If the blob is a layer, a zero-based index of the layer within the image; nil otherwise.
+
+	TOCDigest digest.Digest // If specified, the blob can be looked up in the destination also by its TOC digest.
```
mtrmac, line 103: If there is a match, `TryReusingBlob*` should have found it I think.

internal/private/private.go
```diff
@@ -117,14 +120,15 @@ type TryReusingBlobOptions struct {
...
+	TOCDigest           digest.Digest          // If specified, the blob can be looked up in the destination also by its TOC digest.
 }
 
 // ReusedBlob is information about a blob reused in a destination.
 // It is the subset of types.BlobInfo fields the transport is responsible for setting.
 type ReusedBlob struct {
-	Digest digest.Digest // Must be provided
-	Size   int64         // Must be provided
+	Digest    digest.Digest // Must be provided, can be empty if TOCDigest is present
+	TOCDigest digest.Digest // Must be provided, can be empty if Digest is present
```
mtrmac, line 130: See elsewhere, I’m not sure this works.

In particular, in all the manifest formats, `digest` is a mandatory value. We have to write _something_ in there even if we find/match a TOC.

Also, below, `Size` of the full blob must be known; so there must conceptually be a full blob and a `Digest` value exists.

(Yes, the c/image/storage model of `PutBlob` pretending to record the original compressed blob is basically cheating, mostly but not always successfully.)

storage/storage_dest.go
```diff
@@ -77,18 +78,20 @@ type storageImageDestination struct {
...
-	fileSizes               map[digest.Digest]int64                               // Mapping from layer blobsums to their sizes
-	filenames               map[digest.Digest]string                              // Mapping from layer blobsums to names of files we used to hold them
-	currentIndex            int                                                   // The index of the layer to be committed (i.e., lower indices have already been committed)
-	indexToAddedLayerInfo   map[int]addedLayerInfo                                // Mapping from layer (by index) to blob to add to the image
-	blobAdditionalLayer     map[digest.Digest]storage.AdditionalLayer             // Mapping from layer blobsums to their corresponding additional layer
-	diffOutputs             map[digest.Digest]*graphdriver.DriverWithDifferOutput // Mapping from digest to differ output
+	blobDiffIDs           map[digest.Digest]digest.Digest                       // Mapping from layer blobsums to their corresponding DiffIDs
+	tocDigests            map[digest.Digest]digest.Digest                       // Mapping from layer blobsums to their corresponding TOC Digest
```
mtrmac, line 82: If we don’t validate the blob sums (compressed digests) ourselves, we can see conflicts in malicious inputs.

storage/storage_dest.go
```diff
@@ -77,18 +78,20 @@ type storageImageDestination struct {
...
+	currentIndex          int                                                   // The index of the layer to be committed (i.e., lower indices have already been committed)
+	indexToAddedLayerInfo map[int]addedLayerInfo                                // Mapping from layer (by index) to blob to add to the image
+	blobAdditionalLayer   map[digest.Digest]storage.AdditionalLayer             // Mapping from layer blobsums to their corresponding additional layer
+	diffOutputs           map[digest.Digest]*graphdriver.DriverWithDifferOutput // Mapping from digest to differ output
 }
 
 // addedLayerInfo records data about a layer to use in this image.
 type addedLayerInfo struct {
 	digest     digest.Digest
+	tocDigest  digest.Digest
```
mtrmac, line 94: Please document here which of the two is mandatory/optional, assuming `tocDigest` is optional.

storage/storage_dest.go
```diff
@@ -289,10 +294,13 @@ func (s *storageImageDestination) PutBlobPartial(ctx context.Context, chunkAcces
 	blobDigest := srcInfo.Digest
 
 	s.lock.Lock()
-	s.uncompressedOrTocDigest[blobDigest] = blobDigest
+	s.blobDiffIDs[blobDigest] = blobDigest
```
mtrmac, line 297: That’s an unverified value.

From a quick check it seems to me that we don’t read it if `tocDigests` is set, so… I think we shouldn’t set it here. And insist on `tocDigest` being known.

storage/storage_dest.go
```diff
@@ -317,122 +325,141 @@ func (s *storageImageDestination) TryReusingBlobWithOptions(ctx context.Context,
...
+		tocDigest:  options.TOCDigest,
 		emptyLayer: options.EmptyLayer,
 	})
 }
 
-// tryReusingBlobAsPending implements TryReusingBlobWithOptions for (digest, size or -1), filling s.uncompressedOrTocDigest and other metadata.
+// tryReusingBlobAsPending implements TryReusingBlobWithOptions for (digestBlob, size or -1), filling s.blobDiffIDs and other metadata.
 // The caller must arrange the blob to be eventually committed using s.commitLayer().
-func (s *storageImageDestination) tryReusingBlobAsPending(digest digest.Digest, size int64, options *private.TryReusingBlobOptions) (bool, private.ReusedBlob, error) {
+func (s *storageImageDestination) tryReusingBlobAsPending(digestBlob digest.Digest, size int64, options *private.TryReusingBlobOptions) (bool, private.ReusedBlob, error) {
```
mtrmac, line 335: `s/digestBlob/blobDigest/` I think

storage/storage_dest.go
```diff
@@ -317,122 +325,141 @@ func (s *storageImageDestination) TryReusingBlobWithOptions(ctx context.Context,
...
 		}
 	}
 
-	if digest == "" {
+	if digestBlob == "" && options.TOCDigest == "" {
 		return false, private.ReusedBlob{}, errors.New(`Can not check for a blob with unknown digest`)
 	}
-	if err := digest.Validate(); err != nil {
-		return false, private.ReusedBlob{}, fmt.Errorf("Can not check for a blob with invalid digest: %w", err)
+	if digestBlob != "" {
```
mtrmac, line 359: See elsewhere… my initial guess is that making `Digest` an entirely optional value would be a much more invasive change for the codebase.

Though it is probably plausible for an input to this function, locally.

storage/storage_dest.go
```diff
@@ -317,122 +325,141 @@ func (s *storageImageDestination) TryReusingBlobWithOptions(ctx context.Context,
...
+	}
+	if options.TOCDigest != "" {
+		if err := options.TOCDigest.Validate(); err != nil {
+			return false, private.ReusedBlob{}, fmt.Errorf("Can not check for a blob with invalid digest: %w", err)
+		}
 	}
 
 	// Check if we've already cached it in a file.
-	if size, ok := s.fileSizes[digest]; ok {
+	if size, ok := s.fileSizes[digestBlob]; ok {
```
mtrmac, line 371: If `digestBlob` were truly optional (??), this should be conditionalized on it being set.

Sure, `fileSizes[""]` is never set… probably… but why worry/check/document that, and we are paying for the `digestBlob != ""` check just below, anyway.

storage/storage_dest.go
```diff
@@ -317,122 +325,141 @@ func (s *storageImageDestination) TryReusingBlobWithOptions(ctx context.Context,
...
+		layers, err := s.imageRef.transport.store.LayersByTOCDigest(options.TOCDigest)
+
+		if err != nil && !errors.Is(err, storage.ErrLayerUnknown) {
+			return false, private.ReusedBlob{}, fmt.Errorf(`looking for layers with TOC digest %q: %w`, options.TOCDigest, err)
+		}
+		if len(layers) > 0 {
+			// Save this for completeness.
+			s.tocDigests[options.TOCDigest] = layers[0].TOCDigest
+			return true, private.ReusedBlob{
+				TOCDigest: layers[0].TOCDigest,
```
mtrmac, line 459: What digest will be written in the manifest if `Digest` is unknown?

Yes, c/image/storage currently lies here because it wants to preserve the original manifest (and sometimes fails in that)

storage/storage_dest.go
```diff
@@ -317,122 +325,141 @@ func (s *storageImageDestination) TryReusingBlobWithOptions(ctx context.Context,
...
+				Size:      layers[0].UncompressedSize,
```
mtrmac, line 460: Is this value… known? AFAICS right now it comes from `DriverWithDifferOutput`… which never sets the field.

Yet, it is a mandatory manifest field, so we must provide the “correct” value.

That, again, suggests preserving the original `Digest`+`Size` here.

storage/storage_dest.go
```diff
@@ -469,7 +496,14 @@ func (s *storageImageDestination) computeID(m manifest.Manifest) string {
...
 		for _, l := range m.Layers {
-			diffIDs = append(diffIDs, l.Digest)
+			if l.Digest != "" {
+				diffIDs = append(diffIDs, l.Digest)
+			} else {
+				tocDigest, err := chunkedToc.GetTOCDigest(l.Annotations)
+				if err == nil && tocDigest != nil {
+					diffIDs = append(diffIDs, *tocDigest)
+				}
+			}
```
mtrmac, line 506: This the critical part:

No, AFAICS this must be dependent on _how the layer was actually pulled_, not just based on what the manifest says. The same manifest with N chunked layers can be pulled 2^N different ways, depending on how the registry decides to accept/reject ranged requests in each case.

storage/storage_dest.go
```diff
@@ -176,6 +180,7 @@ func (s *storageImageDestination) PutBlobWithOptions(ctx context.Context, stream
 
 	return info, s.queueOrCommit(*options.LayerIndex, addedLayerInfo{
 		digest:     info.Digest,
+		tocDigest:  options.TOCDigest,
```
mtrmac, line 183: If this sets `tocDigest`, we can’t differentiate between non-chunked and chunked layer pulls; that again allows cross-format reuse we didn’t validate.

storage/storage_dest.go
```diff
@@ -555,20 +589,25 @@ func (s *storageImageDestination) queueOrCommit(index int, info addedLayerInfo)
...
 	defer s.lock.Unlock()
 
-	if d, found := s.diffOutputs[uncompressedDigest]; found {
-		return d.TOCDigest, found
+	if tocDigest != "" {
+		if d, found := s.tocDigests[tocDigest]; found {
+			return d, found
+		}
+		if d, found := s.diffOutputs[uncompressedDigest]; found {
+			return d.TOCDigest, found
```
mtrmac, line 602: Shouldn’t `tocDigests` already be updated by whatever set `diffOutputs`?!

storage/storage_dest.go
```diff
@@ -555,20 +589,25 @@ func (s *storageImageDestination) queueOrCommit(index int, info addedLayerInfo)
...
+// getLayerID returns the diffID for the specified digest or the digest for the TOC, if known.
+func (s *storageImageDestination) getLayerID(uncompressedDigest digest.Digest, tocDigest digest.Digest) (digest.Digest, bool) {
 	s.lock.Lock()
 	defer s.lock.Unlock()
 
-	if d, found := s.diffOutputs[uncompressedDigest]; found {
-		return d.TOCDigest, found
+	if tocDigest != "" {
+		if d, found := s.tocDigests[tocDigest]; found {
+			return d, found
```
mtrmac, line 599: Please have `getLayerID` return a string which is not a possible digest, maybe `"TOC:"+d.String()`, to ensure the inputs to the hashing don’t have collisions.

storage/storage_dest.go
```diff
@@ -599,35 +638,42 @@ func (s *storageImageDestination) commitLayer(index int, info addedLayerInfo, si
 
 	// Check if there's already a layer with the ID that we'd give to the result of applying
 	// this layer blob to its parent, if it has one, or the blob's hex value otherwise.
-	// The diffIDOrTOCDigest refers either to the DiffID or the digest of the TOC.
-	diffIDOrTOCDigest, haveDiffIDOrTOCDigest := s.getDiffIDOrTOCDigest(info.digest)
-	if !haveDiffIDOrTOCDigest {
+	// The layerID refers either to the DiffID or the digest of the TOC.
+	layerID, haveLayerID := s.getLayerID(info.digest, info.tocDigest)
```
mtrmac, line 642: Right now, `info` doesn’t tell whether we pulled by TOC, or by full blob.

storage/storage_dest.go
```diff
@@ -599,35 +638,42 @@ func (s *storageImageDestination) commitLayer(index int, info addedLayerInfo, si
...
-		if !haveDiffIDOrTOCDigest {
-			return false, fmt.Errorf("we have blob %q, but don't know its uncompressed or TOC digest", info.digest.String())
+
+		layerID, haveLayerID = s.getLayerID(info.digest, info.tocDigest)
+		if !haveLayerID {
+			d := info.digest
+			if d == "" {
+				d = info.tocDigest
+			}
+			return false, fmt.Errorf("we have blob %q, but don't know its uncompressed or TOC digest", d.String())
```
mtrmac, line 671: Having two different unambiguous error messages where would be valuable, and actually shorter

storage/storage_dest.go
```diff
@@ -707,7 +753,7 @@ func (s *storageImageDestination) commitLayer(index int, info addedLayerInfo, si
 	if !ok {
 		// Try to find the layer with contents matching that blobsum.
 		layer := ""
-		layers, err2 := s.imageRef.transport.store.LayersByUncompressedDigest(diffIDOrTOCDigest)
+		layers, err2 := s.imageRef.transport.store.LayersByUncompressedDigest(layerID)
```
mtrmac, line 756: `layerID`, right now, is not an uncompressed digest; this is again mixing the namespaces.

I know… we “shouldn’t” get here because we already checked `diffOutputs`, but that could have failed e.g. because some concurrent process deleted the chunked layer in the meantime.

storage/storage_dest.go
```diff
@@ -763,7 +809,7 @@ func (s *storageImageDestination) commitLayer(index int, info addedLayerInfo, si
 	// TODO: This can take quite some time, and should ideally be cancellable using ctx.Done().
 	layer, _, err := s.imageRef.transport.store.PutLayer(id, lastLayer, nil, "", false, &storage.LayerOptions{
 		OriginalDigest:     info.digest,
-		UncompressedDigest: diffIDOrTOCDigest,
+		UncompressedDigest: layerID,
```
mtrmac, line 812: Also mixing the namespaces.

mtrmac (commented):
I’m afraid this review is probably way too repetitive, I just wanted to get some first impression out immediately.

The primary concern is that if if we create or reuse a layer by DiffID, vs. create/reuse a layer by TOC, those two layers should have a different ID.

That doesn’t depend on whether the manifest entry _has_ a TOC digest, but whether we _used_ the TOC digest for creating or looking up that particular layer.

storage/storage_dest.go
giuseppe: yes sorry, this is wrong, I'll drop this line

storage/storage_dest.go
giuseppe: in `LayerInfosForCopy()` we override the digest with the expected-digest, that was stored here: https://github.com/containers/image/blob/main/storage/storage_dest.go#L672-L675

the expected-digest is not validated when we pull the image but it is validated when we copy from the storage.

The TOC instead is validated by `ApplyDiffWithDiffer`.  So if the TOC is not valid, `PutBlobPartial` fails.

So we lookup a layer by its TOC only if the TOC was specified in an annotation.  An image can use a layer by its TOC only if the image manifest contains the same TOC digest.  We do not lookup by TOC if the layer has no TOC annotation.

When we copy from the storage, we override the Digest with the expected-digest flag, but this is going to be validated later.

A layer that doesn't have a TOC annotation must not use a layer from the storage if it is stored by its TOC.

copy/single.go
giuseppe: I am not sure what would be the best way to show the progress bar in this case.

I'll update it to show the original digest and make it clear that it was found by TOC

storage/storage_dest.go
giuseppe: the TOCs are always validated by `store.ApplyDiffWithDiffer()`.

If a layer has an annotation referring to a TOC, then we use that for the lookup in the storage (where the TOC are stored only after being validated).  This is conceptually the same thing we do with the diffID, if an image asks for a certain diffID, we lookup in the storage if we know that layer

storage/storage_dest.go
giuseppe: but we have filled `diffOutputs` before entering this function, so it should not matter if some concurrent process deleted the layer.  Not sure exactly what error we'd get if this happens but `diffOutput, ok := s.diffOutputs[info.digest]` will be true.

storage/storage_dest.go
giuseppe: same here, we should not get so far if a partial pull is used.

storage/storage_dest.go
mtrmac: Yes, but the map _key_ is not validated on the TOC path.

On the non-TOC path, same compressed digest implies same uncompressed digest, and we do compute both of them, so we trust the values.

On the TOC path, we do verify the TOC digest (map value), but the key is currently the compressed digest. I don’t see that anything prevents the attacker to create an image with
```
{digest = $same, tocDigest = $toc1}
{digest = $same, tocDigest = $toc2}
{digest = $same, tocDigest unset}
```
and what happens then? My instinct is to make this impossible instead of trying to reason through it.

(One possible way to handle this concern would be to reject images with conflicting data for the same compressedDigest values. But, see elsewhere, I think we need to maintain “how was this pulled” data by layer index to compute correct image IDs anyway… so using layer indices for map keys here seems ultimately simpler to me, admittedly without having tried to implement it.)

storage/storage_dest.go
mtrmac: I agree that `LayerInfosForCopy` handles the “push” use case fine.

But, still, a “pull” does record a manifest in c/storage.

In general the destination transport is supposed to collaborate with the generic copy code by returning the right partial edits in `ReusedBlob`/`UploadedBlob`, so that the destination can receive an “appropriate” manifest that works correctly with the blobs.

What we _really_ want to achieve with that in c/storage is to record the original, unmodified, version of the manifest, along with signatures, so that when the image is found to contain a backdoor, we can ~prove it was originally correctly signed. (That doesn’t always work when pulling encrypted images, but that’s the goal.)

Given that context, I think the `ReusedBlob` data returned here should be designed so that it doesn’t trigger any manifest update.

Also, the compressed-blob digest is _mandatory_ in manifests, so we can’t very well store a manifest in c/storage which just doesn’t _have_ that field. We could, sure, decide that `ReusedBlob` can return no `Digest` field, and in that case the generic copy pipeline will not modify the value. Given the special treatment of manifest/blob updates in c/storage, I think keeping that special case inside c/storage is a bit more appropriate.

(I could also well accept an argument that doing all this to store the original manifest in c/storage is just wrong, and that we should record an “updated” manifest in c/storage, and then we can get ~rid of `LayerInfosForCopy`; and we should have a _separate“ `PutOriginalManifest` method to deal with the desired feature that way. But that’s not what we have, and `LayerInfosForCopy` needs to continue to be supported as part of the stable API.)

storage/storage_dest.go
mtrmac: That value would, conceptually, be written into the manifest stored via `PutManifest`. And, see elsewhere, we primarily want to preserve the manifest unmodified, along with original signatures.

… I think `layerDigestsDiffer` is involved and it means some of the updates are (correctly? incorrectly?) just not done. I’ll might need to revisit what actually happens.

storage/storage_dest.go
mtrmac: Oh, you are right, we can not reach it with the wrong value.

Still, I think doing a `diffID, ok := blobDiffIDs[…]; if !ok { fail }` lookup is trivially cheap, and it would make it clear that we are using the right value.

mtrmac: Also, assuming https://github.com/containers/image/pull/2218#discussion_r1424655287 , `LayersByUncompressedDigest(layerID)` will become a type error.

storage/storage_dest.go
giuseppe: I've not thought of all possible implications but while compressed and uncompressed digests can always be used  interchangeably (one compressed digest refers to one uncompressed digest, and vice-versa), that is not true for TOC digests.  e.g. it is enough to add some padding to the tarball to cause compressed and uncompressed digest to change, but that won't affect the TOC digest.

mtrmac: - Actually, a single uncompressed digest can have many compressed-digest values (one per algorithm is the lower bound, and most compression algorithms are not deterministic)
- It’s not “quite true” for uncompressed digests, in that sense, either — a tar file can have “the same contents” but a different ending padding (to mention a recent example).

Here, I’m not worried about finding all possible duplicates and deduplicating. That’s an optimization.

I am worried about _incorrectly_ treating two things as duplicates when we don’t know for sure they are identical. That’s a vulnerability.

storage/storage_dest.go
giuseppe: changed to only use the TOC id if it is the one we used to refer to the layer

giuseppe:
opened a PR for c/storage to not allow two TOCs to be specified: https://github.com/containers/storage/pull/1778

rhatdan:
Container storage PR merged, can we move this forward?

mtrmac:
- Outstanding review comments, AFAICT
- Immediate higher priorities, so I won’t revisit now just to verify

mtrmac:
@rhatdan If you want to help move things forward, consider working on https://github.com/containers/image/pull/1980#discussion_r1417292899 . If nothing else, that on is a showstopper and if I run out of time, I’m going to do something drastic.

internal/private/private.go
giuseppe: dropped

types/types.go
giuseppe: dropped

storage/storage_dest.go
giuseppe: if `info.tocDigest != ""` then the layer was pulled by TOC

giuseppe:
opened a PR for c/storage for the outstanding issue with converted images (https://github.com/containers/image/pull/1980#discussion_r1417292899): 
- https://github.com/containers/storage/pull/1794

giuseppe:
can we move this forward?

storage/storage_dest.go
```diff
@@ -457,7 +487,7 @@ func (s *storageImageDestination) computeID(m manifest.Manifest) string {
 				continue
 			}
 			blobSum := m.FSLayers[i].BlobSum
-			diffID, ok := s.uncompressedOrTocDigest[blobSum]
+			diffID, ok := s.blobDiffIDs[blobSum]
```
mtrmac, line 490: Actually the `copy --format …` case means that we can’t assume that TOC reuse happens _only_ with OCI schema used in `PutManifest`.

That’s, in a sense, liberating — `m.ComputeID` below can be completely TOC-unaware, and this function can then modify the computed ID itself:
```go
manifestID := m.ComputeID()
if layerWasPulledByTOCDigest {
   input := manifestID
   for _, howLayerWasPulled := range howLayersWerePulled {
     if toc := howLayerWasPulled.byThisTOCDigest; toc != N/A {
        input += "toc=" toc +","
     } else {
        input += "no-toc,"
     }
   }
   manifestID = digest(input)
}
```

---

OTOH the code looking for `DiffIDs` in `commitLayer` needs to be ready to handle non-OCI inputs, or to at least cleanly fail.

storage/storage_dest.go
```diff
@@ -555,20 +593,22 @@ func (s *storageImageDestination) queueOrCommit(index int, info addedLayerInfo)
...
-func (s *storageImageDestination) getDiffIDOrTOCDigest(uncompressedDigest digest.Digest) (digest.Digest, bool) {
+// getLayerID returns the diffID for the specified digest or the digest for the TOC, if known.
+func (s *storageImageDestination) getLayerID(uncompressedDigest digest.Digest, tocDigest digest.Digest) (digest.Digest, bool) {
 	s.lock.Lock()
 	defer s.lock.Unlock()
 
-	if d, found := s.diffOutputs[uncompressedDigest]; found {
-		return d.TOCDigest, found
+	if tocDigest != "" {
+		if d, found := s.tocDigests[uncompressedDigest]; found {
```
mtrmac, line 602: This wouldn’t work as written, `tocDigests[]` is set indexed by the (compressed) blob digest. I suspect the `uncompressedDigest` parameter name is incorrect.

(The whole thing needs to index by layer index within image, anyway… Or maybe use the `tocDigest` value, if the caller already knows it.)

storage/storage_dest.go
```diff
@@ -555,20 +593,22 @@ func (s *storageImageDestination) queueOrCommit(index int, info addedLayerInfo)
...
+			return d + "-toc", found
```
mtrmac, line 603: This function claims to return a `digest.Digest`, a string like this isn’t a valid `digest.Digest`.

I think it’s better to make this return a string — and to make the name _clearly_ outside of the namespace of valid `digest.Digest` values. `-toc` is actually acepted in `digest.DigestRegexp` (it would _typically_ be rejected because the SHA-256 algorithm requires the digest value to be hexadecimal, but this doesn’t reject non-`sha256:` values… and being forward-compatible is cheap)

Move the `layerID.Hex()` call inside this function, to the non-TOC case.

storage/storage_dest.go
```diff
@@ -599,35 +639,43 @@ func (s *storageImageDestination) commitLayer(index int, info addedLayerInfo, si
...
-		diffIDOrTOCDigest, haveDiffIDOrTOCDigest = s.getDiffIDOrTOCDigest(info.digest)
-		if !haveDiffIDOrTOCDigest {
-			return false, fmt.Errorf("we have blob %q, but don't know its uncompressed or TOC digest", info.digest.String())
+
+		layerID, haveLayerID = s.getLayerID(info.digest, info.tocDigest)
+		if !haveLayerID {
+			d := info.digest
+			if d == "" {
+				d = info.tocDigest
+			}
```
mtrmac, line 671: In which situations would these values be actually usable? `tryReusingBlobAsPending` _must_ arrange for a layer to be found in the latter parts of this function.

copy/single.go
```diff
@@ -718,7 +723,13 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, to
 		if reused {
 			logrus.Debugf("Skipping blob %s (already present):", srcInfo.Digest)
 			func() { // A scope for defer
-				bar := ic.c.createProgressBar(pool, false, types.BlobInfo{Digest: reusedBlob.Digest, Size: 0}, "blob", "skipped: already exists")
+				d := reusedBlob.Digest
+				label := "skipped: already exists"
+				if d == "" {
```
mtrmac, line 728: `reusedBlob.Digest` is never empty (and it must never be empty).

storage/storage_dest.go
mtrmac: We are returning `ReusedBlob{Digest: blobDigest}` even in the TOC match case, so it really doesn’t make sense to accept an empty `blobDigest` in any of the code paths.

storage/storage_dest.go
mtrmac, line 602: If `!found`, isn’t that an internal error that should cause a failure?

mtrmac (commented):
Most importantly, the design issue around indexing by (compressed) blob digest when individual layers might actually have been pulled/matched by TOC ( https://github.com/containers/image/pull/2218#discussion_r1425546810 ) remains outstanding.

Resolving that will almost certainly affect many other areas with review comments.

storage/storage_dest.go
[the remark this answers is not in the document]
giuseppe: what error exactly do you have in mind?

If I try to pull a zstd image with `--format v2s1`, the operation fails with:

```
FATA[0005] creating an updated image manifest: Error during manifest conversion: "application/vnd.oci.image.layer.v1.tar+zstd": zstd compression is not supported for docker images 
```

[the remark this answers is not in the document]
mtrmac: That’s a good point.

Still… what happens when someone creates an image with a gzip MIME type but a TOC annotation?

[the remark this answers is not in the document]
giuseppe: do you suggest returning an explicit error if `!ok` (instead of returning an empty string) and propagating it to the caller?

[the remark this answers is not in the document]
mtrmac: Actually the fallback code (randomly generating a new image ID and not reusing an image) is not too bad … just a comment to the effect that “this can, in principle, legitimately happen when a layer is reused by TOC” would be useful.

rhatdan:
@mtrmac Is this ready to go?

rhatdan:
LGTM

storage/storage_dest.go
```diff
@@ -555,20 +594,26 @@ func (s *storageImageDestination) queueOrCommit(index int, info addedLayerInfo)
 	return nil
 }
 
-// getDiffIDOrTOCDigest returns the diffID for the specified digest or the digest for the TOC, if known.
-func (s *storageImageDestination) getDiffIDOrTOCDigest(uncompressedDigest digest.Digest) (digest.Digest, bool) {
+// getLayerID returns the diffID for the specified digest or the digest for the TOC, if known.
+func (s *storageImageDestination) getLayerID(uncompressedDigest digest.Digest, tocDigest digest.Digest, index int) (string, bool) {
```
mtrmac, line 598: (This might accept an `addedLayerInfo` as a single parameter.)

storage/storage_dest.go
```diff
@@ -416,23 +432,37 @@ func (s *storageImageDestination) tryReusingBlobAsPending(digest digest.Digest,
...
+	if options.TOCDigest != "" && options.LayerIndex != nil {
+		// we are already holding s.lock
+		diffOutput, ok := s.diffOutputs[blobDigest]
+		if ok {
+			// Save this for completeness.
+			s.indexToTocDigest[*options.LayerIndex] = options.TOCDigest
+			return true, private.ReusedBlob{
+				Digest:    blobDigest,
+				TOCDigest: diffOutput.TOCDigest,
+				Size:      diffOutput.Size,
```
mtrmac, line 444: (If this case remains, which I’m unsure about:) `Size` needs to be consistent with `blobDigest`, i.e. `size`.

storage/storage_dest.go
mtrmac: The code was replaced, but the misleading parameter name remains.

mtrmac (commented):
Thanks!

I didn’t yet look at the full picture of layer finding/preparation vs. commit, submitting this ~early to point out some of the concerns.

mtrmac:
(Also note the probably-outstanding comments from earlier reviews.)

giuseppe:
> Thanks!
> 
> I didn’t yet look at the full picture of layer finding/preparation vs. commit, submitting this ~early to point out some of the concerns.

if it makes things easier, would you mind taking this over since you are already working in the same area in c/image?

mtrmac:
@giuseppe Let’s try that, I’ll take this over for now. Please note the latter part of https://github.com/containers/image/pull/2218#discussion_r1480520028 (related to an image having “the same” chunked layer twice, I think). Either that’s an acceptable limitation for now, or it might need a change in how c/storage creates/uses/processes layers.

internal/private/private.go
mtrmac: Resolved in #2288.

internal/private/private.go
mtrmac: Resolved in #2288 ; there is now a `MatchedByTOCDigest` sufficient for the existing user.

storage/storage_dest.go
mtrmac: Resolved in #2288.

storage/storage_dest.go
mtrmac: This was already updated to preserve `Digest: blobDigest`.

storage/storage_dest.go
mtrmac: #2288 has removed support for TOC on this fallback path.

storage/storage_dest.go
mtrmac: Resolved in #2288.

storage/storage_dest.go
mtrmac: This was fixed here.

internal/private/private.go
```diff
@@ -55,7 +55,7 @@ type ImageDestinationInternalOnly interface {
 	// It is available only if SupportsPutBlobPartial().
 	// Even if SupportsPutBlobPartial() returns true, the call can fail, in which case the caller
 	// should fall back to PutBlobWithOptions.
-	PutBlobPartial(ctx context.Context, chunkAccessor BlobChunkAccessor, srcInfo types.BlobInfo, cache blobinfocache.BlobInfoCache2) (UploadedBlob, error)
+	PutBlobPartial(ctx context.Context, chunkAccessor BlobChunkAccessor, srcInfo types.BlobInfo, index *int, cache blobinfocache.BlobInfoCache2) (UploadedBlob, error)
```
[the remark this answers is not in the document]
mtrmac: Resolved in #2288.

storage/storage_dest.go
```diff
@@ -286,13 +289,19 @@ func (s *storageImageDestination) PutBlobPartial(ctx context.Context, chunkAcces
...
+		return private.UploadedBlob{}, errors.New("TOC digest is empty")
+	}
+
 	blobDigest := srcInfo.Digest
 
 	s.lock.Lock()
-	s.uncompressedOrTocDigest[blobDigest] = blobDigest
 	s.fileSizes[blobDigest] = 0
 	s.filenames[blobDigest] = ""
 	s.diffOutputs[blobDigest] = out
```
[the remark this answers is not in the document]
mtrmac: Resolved in #2288.

storage/storage_dest.go
```diff
@@ -416,23 +432,37 @@ func (s *storageImageDestination) tryReusingBlobAsPending(digest digest.Digest,
...
-			Size:   layers[0].UncompressedSize,
-		}, nil
+		// Check if we have a chunked layer in storage with the same TOC digest.
+		layers, err := s.imageRef.transport.store.LayersByTOCDigest(options.TOCDigest)
+
+		if err != nil && !errors.Is(err, storage.ErrLayerUnknown) {
+			return false, private.ReusedBlob{}, fmt.Errorf(`looking for layers with TOC digest %q: %w`, options.TOCDigest, err)
+		}
+		if len(layers) > 0 {
+			if options.LayerIndex != nil {
```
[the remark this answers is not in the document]
mtrmac: Resolved in #2288.

storage/storage_dest.go
mtrmac: In #2288, only one field of `addedLayerInfo` is necessary, removing the benefit of this change.

storage/storage_dest.go
mtrmac: #2288 removed this case.

storage/storage_dest.go
mtrmac: Locally, if this were the _only_ change on pull, we wouldn’t update layer info in the manifest (we update the manifest either if `LayerInfosForCopy` returned something, or if at least one digest changes).

But it’s possible for some other layer entry to change the digest, and then the values returned here would be used, even if _this_ layer didn’t return a changed digest. Besides, returning inconsistent metadata is just unclean.

#2294 should fix this.

storage/storage_dest.go
mtrmac: The second part will be fixed in #2295.

PR #2249, opened 2024-01-17
Bump c/storage to v1.52.0, c/image to v5.30.0 and then v5.30.1-dev

TomSweeneyRedHat:
As the title says.  Reading for Podman v4.9 release

[NO NEW TESTS NEEDED]

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -6,12 +6,12 @@ const (
 	// VersionMajor is for an API incompatible changes
 	VersionMajor = 5
 	// VersionMinor is for functionality in a backwards-compatible manner
-	VersionMinor = 29
+	VersionMinor = 30
 	// VersionPatch is for backwards-compatible bug fixes
-	VersionPatch = 1
+	VersionPatch = 0
 
 	// VersionDev indicates development branch. Releases will be empty string.
-	VersionDev = "-dev"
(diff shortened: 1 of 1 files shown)


mtrmac (commented):
@TomSweeneyRedHat IIRC we actually want a feature backport, so this shouldn’t go ahead, should it?

(Also, the tests fail.)

TomSweeneyRedHat:
@mtrmac Urvashi added some changes to c/image main that she needed for the Farm Build work.  The thinking was to bump the version here and in c/storage, then bump those versions in the c/common v0.57.0 branch which went out with Podman v4.8, then create a new c/common there in that branch and merge all into the Podman v4.9 branch.  I had initially planned to also bump c/common, but apparently there are some non-compatible Podman v5.0 changes in c/common main.  I was hoping this approach would be less intensive than backporting to the c/storage and c/image repos.

However, if you or others think there are changes in here or c/storage that would be incomptatible with Podman v4.9, backports will have to be done instead.

I tell you, whitespace issues will be the death of me yet.

TomSweeneyRedHat:
UGH!  Closing.  I managed to wipe out the c/storage bump commit when fixing the whitespace.  Will spin a new PR.

PR #2257, opened 2024-01-19
Main is after latest release

mtrmac:
Release 5.29.1 happened on a branch; merge it into `main` to let Git, and go tools, now that `main` is “after” 5.29.1, notably so that updating to `main` does not trigger a downgrade of c/common to a version which does not require 5.29.1, or an upgrade of c/common does not trigger a downgrade from a `main` checkout to the branched 5.29.1 version of c/image.


TomSweeneyRedHat:
LGTM

PR #2258, opened 2024-01-20
feat: negotiate docker client version

dpeckett:
Instead of hardcoding the docker client version we should negotiate a version at runtime. Version 1.22 is very out of date and no longer is supported by modern daemon versions.

### DIFF
--- a/docker/daemon/client.go
+++ b/docker/daemon/client.go
@@ -1,6 +1,7 @@
 package daemon
 
 import (
+	"context"
 	"net/http"
 	"path/filepath"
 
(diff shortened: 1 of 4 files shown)


rhatdan:
@mtrmac PTAL

TomSweeneyRedHat:
This is a different approach to solving the problem as https://github.com/containers/image/pull/2260

mtrmac (commented):
Thanks!

- There is `WithAPIVersionNegotiation`; that’s a bit shorter than calling `NegotiateAPIVersion` explicitly, and more similar to what the Docker CLI does.
- The unit tests run unprivileged, in a fairly arbitrary environment; we can’t assume a running Docker daemon to be available and accepting connections. In particular the proposed test implementation doesn’t work in our CI.

#2260 is, right now, probably not the ideal solution, but it’s a bit closer to it, so let’s discuss there.

mtrmac:
#2260 was merged. Still, thanks for your contribution!

Issue #2259, opened 2024-01-22
Docker client code can no longer talk to the latest verson of the docker daemon 25.0.0


bainsy88:
The latest version of the Docker Daemon has now upped the minimum API version to 1.24 via this PR - https://github.com/moby/moby/pull/46887

This now means tools such as skopeo fail against the 25.0.0 docker daemon with the following error 

```
loading image from docker engine: Error response from daemon: client version 1.22 is too old. Minimum supported API version is 1.24, please upgrade your client to a newer version
```

This is caused by the hardcoding of the API version to 1.22 here - https://github.com/containers/image/blob/main/docker/daemon/client.go#L14

This version on the containers/image side has not been updated for 7 years and it is now true that in the case a version is not set the client and daemon will try negotiate an API Version that they both support so that seems a cleaner solution but not sure if there is a good reason why a version would need to be set in this case.

bainsy88:
Added a PR to remove the version pinning as a first pass. Happy to change to pinning at a newer version if there is a good reason to do so.

mtrmac:
Thanks! Let’s discuss the details in #2260 .

dcermak:
Would it be possible to cut a new release of c/image so that we can get an updated podman, buildah and skopeo out?

mtrmac:
@dcermak This is already included in a branched release of c/image, and e.g. in Podman 4.9.2, Buildah 1.33.5, Skopeo 1.14.2.

Updating the mainline versions is, I think, currently WIP (around https://github.com/containers/buildah/pull/5286 ), but it will certainly get done in the future.

PR #2260, opened 2024-01-22
Allow docker client to negotiate the API version with the server 

bainsy88:
The version is currently hardcoded which is causing the issue documented here - https://github.com/containers/image/issues/2259

Decided adding a unit test wasn't feasible as as the version would change each time we pulled in a new docker client and as long as it negotiates a version that works I suspect we don't care what that exact version is.

### DIFF
--- a/docker/daemon/client.go
+++ b/docker/daemon/client.go
@@ -9,11 +9,6 @@ import (
 	"github.com/docker/go-connections/tlsconfig"
 )
 
-const (
-	// The default API version to be used in case none is explicitly specified
-	defaultAPIVersion = "1.22"
-)
-
 // NewDockerClient initializes a new API client based on the passed SystemContext.
 func newDockerClient(sys *types.SystemContext) (*dockerclient.Client, error) {
 	host := dockerclient.DefaultDockerHost
@@ -23,7 +18,6 @@ func newDockerClient(sys *types.SystemContext) (*dockerclient.Client, error) {
 
 	opts := []dockerclient.Opt{
 		dockerclient.WithHost(host),
-		dockerclient.WithVersion(defaultAPIVersion),
 	}
 
 	// We conditionalize building the TLS configuration only to TLS sockets:
--- a/docker/daemon/client.go
+++ b/docker/daemon/client.go
@@ -18,6 +18,7 @@ func newDockerClient(sys *types.SystemContext) (*dockerclient.Client, error) {
 
 	opts := []dockerclient.Opt{
 		dockerclient.WithHost(host),
+		dockerclient.WithAPIVersionNegotiation(),
 	}
 
 	// We conditionalize building the TLS configuration only to TLS sockets:
--- a/docker/daemon/client_test.go
+++ b/docker/daemon/client_test.go
@@ -18,7 +18,6 @@ func TestDockerClientFromNilSystemContext(t *testing.T) {
 	assert.NotNil(t, client, "A Docker client reference should have been returned")
 
 	assert.Equal(t, dockerclient.DefaultDockerHost, client.DaemonHost(), "The default docker host should have been used")
-	assert.Equal(t, defaultAPIVersion, client.ClientVersion(), "The default api version should have been used")
 
 	assert.NoError(t, client.Close())
 }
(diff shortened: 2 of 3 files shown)


TomSweeneyRedHat:
The change Looks good for what you're trying to do.  I'm just a bit hesitant to drop the test.  I'd lean towards bumping the default version to 1.24 instead.  @mtrmac WDYT?

Regardless, TYVM for the PR @bainsy88

TomSweeneyRedHat:
This is a different approach to solving the problem as https://github.com/containers/image/pull/2258

bainsy88:
Ah I hadn't noticed the other PR which beat me to it. Happy to go with the other one given it already had tests and that was a concern with this change

docker/daemon/client.go
```diff
@@ -23,7 +18,6 @@ func newDockerClient(sys *types.SystemContext) (*dockerclient.Client, error) {
 
 	opts := []dockerclient.Opt{
 		dockerclient.WithHost(host),
-		dockerclient.WithVersion(defaultAPIVersion),
```
mtrmac, line 26: AFAICT this needs to be replaced with `dockerclient.WithAPIVersionNegotiation()` to _actually_ opt into negotiating the version.

(The effect of negotiating is that the client and server use the lower of the two versions, agreeing on a mutually acceptable one.)

Without that opt-in, the client uses version-unspecified APIs, which the server interprets as “the current API”, i.e. every upgrade of the server can change semantics.

Can you make that change, and double-check that everything works as expected, please?

mtrmac (commented):
Thanks!

- Version negotiation ends up being better than hard-coding; with negotiation, there is a chance for the client & server to use a mutually-intelligible version larger than 1.24 (and version negotiation exists since 1.24, so hard-coding a requirement for 1.24 without negotiation is no less restrictive. Opting into version negotiation _does_ interoperate with <1.24… using the 1.24 semantics. So version negotiation is at least as compatible as hard-coding 1.24, and more compatible long-term).
- We do need to actually enable version negotiation to benefit that way, though. This PR does not currently do that, AFAICT. (And the Docker CLI does opt into version negotiation, which is another reason to do the same thing.)
- (The history is that the code was originally written when there was no version negotiation (in the client libraries at least); there is a fairly extensive discussion about the trade-offs at the time in https://github.com/containers/image/pull/328 , but just during that time version negotiation was being added to the client libraries…)
- I’m fine with removing the tests checking the client’s properties. The client’s configuration applied based on the options is, for the most part, not inspectable, we just have to trust that the client initialization would fail if we gave it nonsensical instructions.

docker/daemon/client.go
bainsy88: Ah good catch I did test with my current code and it worked but having looked again, without the change you suggested, we would just get the `api.DefaultVersion` from the docker client code which is currently 1.44. Which would mean we would fail against older daemons

bainsy88:
I have built skopeo off this code and added some additional debug so can see the client version is 1.44 but it negotiates it down to 1.43 when I run against Docker 24.0.7

docker/daemon/client.go
mtrmac: > Without that opt-in, the client uses version-unspecified APIs, which the server interprets as “the current API”, i.e. every upgrade of the server can change semantics.

I was incorrect about this. As you say, without negotiation the client uses the latest known version (at time of vendoring of the client), which right now means it would fail against Docker ≤24:
```
Error response from daemon: client version 1.44 is too new. Maximum supported API version is 1.43"
```

So, yes, we do _need_ the version negotiation opt-in.

mtrmac:
Confirming that code from this PR works with both Docker 24 and 25.

Alternatively, a workaround:
```console
# cat /etc/systemd/system/docker.service.d/version.conf 
[Service]
Environment=DOCKER_MIN_API_VERSION=1.22
```

mtrmac (approved):
Thanks, to both @bainsy88  and @dpeckett !

TomSweeneyRedHat:
Yes, @bainsy88 and @dpeckett TY both very much for this discussion and change.

PR #2273, opened 2024-01-31
Edit the manifest when pushing uncompressed data from c/storage

### LINKED ISSUE #2182: Copies of originally-compressed images from c/storage to uncompressed destinations don’t trigger MIME type updates
Originally reported in https://github.com/containers/podman/issues/20611#issuecomment-1802428561 : pull a zstd:chunked image, then `podman save` (to `docker-archive`) fails with
```
Error: creating an updated image manifest: Error during manifest conversion: "application/vnd.oci.image.layer.v1.tar+zstd": zstd compression is not supported for docker images
```

What happens here:
- The original manifest is recoded, with a zstd MIME type
- Because the data is actually stored uncompressed, `LayerInfosForCopy` returns an uncompressed MIME type
- `imageCopier.copyLayers` does account for the `LayerInfosForCopy` changes by setting `srcInfosUpdated = true`. But that isn’t sufficient, because it is _documented_ that `ManifestUpdateOptions.LayerInfos` ignores the MIME type, the edit must be represented by `CompressionOperation`+`CompressionAlgorithm`.
- `imageCopier.copyLayer` has code to compensate for this, by setting `CompressionAlgorithm` based on MIME type. But it doesn’t set `CompressionOperation: Decompress` for uncompressed MIME types; and it _only_ applies on the `TryReusingBlob` code path.
- On the primary “full copy” code path, `CompressionOperation`+`CompressionAlgorithm` are set separately, based just on the characteristics of the incoming blob. And when the blob is uncompressed, and the desired state is also uncompressed, there is “nothing to do”, and the code sets `CompressionOperation: PreserveOriginal`, `CompressionAlgorithm: nil`, i.e. “do no edits”.
- As a result, the original manifest with a zstd MIME type is not updated to correctly represent an uncompressed layer, and a conversion to schema2 then fails.

Ugh. This is going to be a _mess_ to untangle.
- Maybe it’s time to finally deprecate the edit fields of `BlobInfo`, and build some internal-only edit interface afresh? (A lot of work, and what would that look like?)
- Or the approach from the top of `copyLayer`, and have `copyBlobFromStream` preserve the supplied operation if it makes no changes?

mtrmac:
Fixes #2182.

<s>At this point completely untested.</s>

### DIFF
--- a/copy/compression.go
+++ b/copy/compression.go
@@ -250,21 +250,25 @@ func (ic *imageCopier) bpcPreserveOriginal(_ *sourceStream, detected bpDetectCom
 	// so that src.UpdatedImage() doesn’t fail; assume that for such blobs
 	// LayerInfosForCopy() should not be making any changes in the first place.
 	var bpcOp bpcOperation
+	var uploadedOp types.LayerCompression
 	var algorithm *compressiontypes.Algorithm
 	switch {
 	case !layerCompressionChangeSupported:
 		bpcOp = bpcOpPreserveOpaque
+		uploadedOp = types.PreserveOriginal
 		algorithm = nil
 	case detected.isCompressed:
 		bpcOp = bpcOpPreserveCompressed
+		uploadedOp = types.PreserveOriginal
 		algorithm = &detected.format
 	default:
 		bpcOp = bpcOpPreserveUncompressed
+		uploadedOp = types.Decompress
 		algorithm = nil
 	}
 	return &bpCompressionStepData{
 		operation:              bpcOp,
-		uploadedOperation:      types.PreserveOriginal,
+		uploadedOperation:      uploadedOp,
 		uploadedAlgorithm:      algorithm,
 		srcCompressorName:      detected.srcCompressorName,
 		uploadedCompressorName: detected.srcCompressorName,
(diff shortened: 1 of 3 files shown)


mtrmac:
Downstream tests:
- https://github.com/containers/buildah/pull/5307
- https://github.com/containers/podman/pull/21461

mtrmac:
I have manually tested that in Podman with https://github.com/containers/podman/pull/21461 , this fixes the original reproducer of #2182.

Also, Skopeo’s tests demonstrate the same behavior — and rely on it, which is why they are failing here. https://github.com/containers/skopeo/pull/2213 will update Skopeo.

mtrmac:
Ready for review. Downstream tests are mostly passing.

The Skopeo test failures here expose the previous incorrect behavior; https://github.com/containers/skopeo/pull/2213 contains test updates, and passes.

Cc: @giuseppe

giuseppe (approved):
LGTM

rhatdan:
Tests are failing?

mtrmac:
Yes, as documented in detail above.

rhatdan:
LGTM

PR #2287, opened 2024-02-08
Random storage-related cleanups

mtrmac:
… on the principle that removing accidental complexity makes it easier to focus on the essential complexity.

(Yes, this conflicts with #2218 ; I’ll rebase that work.)

@giuseppe PTAL.

Issue #2189, opened 2023-11-13
Zstd(:chunked) work tracking checklist


mtrmac:
- Zstd in general:
  - [ ] Pushes with Zstd compression prefer using schema1 manifests: https://github.com/containers/image/issues/2181
  - [ ] Uncompressed pushes of originally-compressed layers don’t correctly update MIME types (always a bug, but now breaks pushes of originally-Zstd to `docker-archive:` destinations): https://github.com/containers/image/issues/2182
  - [x] Not a bug: pull of gzip + push of zstd changes image ID due to triggering a conversion to OCI: https://github.com/containers/podman/issues/20611
- Zstd:chunked:
  - [ ] Pushes of chunked layers fail because the `UncompressedDigest` fields stores a TOC digest, and layer digest validation fails: Part of https://github.com/containers/image/pull/1980
  - [ ] Pushes reusing chunked layers don’t create the required annotations, making the layers impossible to pull chunked: https://github.com/containers/image/pull/2183#pullrequestreview-1725228215 / https://github.com/containers/image/pull/2185
  - [ ] Pulls of chunked layers do not reuse locally-existing layers: Part of https://github.com/containers/image/pull/1980
  - [ ] Pulls of chunked layers do not reuse locally-existing layers: Part of https://github.com/containers/image/pull/1980
  - [ ] In c/storage, we don’t sufficiently differentiate layers pulled by blob digest vs. TOC digest, and reuse is unclear: Part of https://github.com/containers/image/pull/1980 , and https://github.com/containers/storage/pull/1627#discussion_r1309255199
  - [ ] Pulls may reuse layers pulled by blob digests vs. TOC digests: Part of https://github.com/containers/image/pull/1980
  - [ ] Image deduplication contains a (required!) sanity check that `TopLayer` matches, but that might not be true if layers pulled by blob vs. TOC digests have different IDs: Part of https://github.com/containers/image/pull/1980
  - [ ] At a higher level, it’s unclear how to safely handle signed chunked images: what if the two versions don’t represent exactly the same filesystem?
  - [ ] `blobPipelineDetectCompressionStep` cannot detect zstd:chunked, causing unnecessary recompression: around https://github.com/containers/image/pull/1980#discussion_r1388492171 . (OTOH we can only use :chunked if it comes with the expected annotations.)
  - [ ] `blobPipelineCompressionStep` triggers a recompression of zstd:chunked if the user asks for zstd: around https://github.com/containers/image/pull/1980#discussion_r1388493306
[...]

PR #2288, opened 2024-02-08
Fix c/storage destination with partial pulls

mtrmac:
This is #2218 rebased on top of #2287, + my WIP work fixing issues as they come up.

Absolutely untested (do we have relevant CI in any project?).

@giuseppe FWIW .

### DIFF
--- a/copy/single.go
+++ b/copy/single.go
@@ -751,7 +751,7 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, to
 				wrapped: ic.c.rawSource,
 				bar:     bar,
 			}
-			uploadedBlob, err := ic.c.dest.PutBlobPartial(ctx, &proxy, srcInfo, ic.c.blobInfoCache)
+			uploadedBlob, err := ic.c.dest.PutBlobPartial(ctx, &proxy, srcInfo, &layerIndex, ic.c.blobInfoCache)
 			if err == nil {
 				if srcInfo.Size != -1 {
 					refill := srcInfo.Size - bar.Current()
--- a/copy/single.go
+++ b/copy/single.go
@@ -695,10 +695,15 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, to
 			requiredCompression = ic.compressionFormat
 		}
 
+		var tocDigest digest.Digest
+
 		// Check if we have a chunked layer in storage that's based on that blob.  These layers are stored by their TOC digest.
-		tocDigest, err := chunkedToc.GetTOCDigest(srcInfo.Annotations)
+		d, err := chunkedToc.GetTOCDigest(srcInfo.Annotations)
 		if err != nil {
 			return types.BlobInfo{}, "", err
+		}
+		if d != nil {
+			tocDigest = *d
 		}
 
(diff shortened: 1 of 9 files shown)


giuseppe:
I am manually testing it, and it seems to work fine so far, I've found only one issue (caused by a recent change I've done in c/storage where a converted image returns the UncompressedDigest since it was fully validated instead of the TOCDigest), or do you think we should address it in c/storage and return both the TOCDigest and the UncompressedDigest (and then use from c/image the one we prefer): https://github.com/containers/storage/blob/d1bf4f0cf1d648fdd2e92f8841229a2da0bc3249/pkg/chunked/storage_linux.go#L1669-L1673?

I've added this patch to fix a pull of a converted to zstd:chunked image:
```
diff --git a/storage/storage_dest.go b/storage/storage_dest.go
index f73a3080..25fe0686 100644
--- a/storage/storage_dest.go
+++ b/storage/storage_dest.go
@@ -289,8 +289,8 @@ func (s *storageImageDestination) PutBlobPartial(ctx context.Context, chunkAcces
 	}
 
-	if out.TOCDigest == "" {
-		return private.UploadedBlob{}, errors.New("TOC digest is empty")
+	if out.TOCDigest == "" && out.UncompressedDigest == "" {
+		return private.UploadedBlob{}, errors.New("digest is empty")
 	}
 
 	blobDigest := srcInfo.Digest
@@ -299,9 +299,12 @@ func (s *storageImageDestination) PutBlobPartial(ctx context.Context, chunkAcces
 	s.fileSizes[blobDigest] = 0
 	s.filenames[blobDigest] = ""
 	s.diffOutputs[blobDigest] = out
-	if index != nil {
+	if index != nil && out.TOCDigest != "" {
 		s.indexToTocDigest[*index] = out.TOCDigest
 	}
+	if out.UncompressedDigest != "" {
+		s.blobDiffIDs[blobDigest] = out.UncompressedDigest
+	}
 	s.lock.Unlock()
 
 	return private.UploadedBlob{
```

mtrmac:
Thanks; that seems to be out of scope of this correctness cleanup, but mentioning it is immediately valuable in how it is affecting the design of this cleanup. I have also added a link to https://github.com/containers/image/issues/2189 .

---

> do you think we should address it in c/storage and return both the TOCDigest and the UncompressedDigest (and then use from c/image the one we prefer):

There’s actually a deeper design question here: Assume
- `podman pull $image` with `convert_images=true`
- `podman pull $image` with `convert_images=false`

both with the same image (in this order, or maybe in some other order), in the same store: should the two pulls share the layer (and result in the same `storage.Image`? Preferably yes / definitely not / nobody really cares?

My first instinct is to prefer the identification by `UncompressedDigest`, so that we reuse images to the maximal extent —but then if we do that, we could reuse an “ordinary” layer and setting `convert_images` does not guarantee any particular behavior.

Either way, _probably_ c/storage should return both values, so that c/image can do the right thing for such mixed-identification layers.

And possibly to reuse the just-created layer in the rare case that a single image contains the same layer twice — whether that reuse happens by DiffID or by TOC.

mtrmac:
@giuseppe If you happened to have too much spare time, this is structurally close (but the last commits are _ugly_), and I’d appreciate an early review. But please see #2291 first.

PR #2291, opened 2024-02-09
Improve storage transport documentation, primarily about locking

mtrmac:
In addition to some small docs updates, this introduces `storageImageDestinationLockProtected` to visibly delineate the fields that need locking. Yes, this is a painful and disruptive change which requires rebasing any other changes, but the time has come. Documenting the relationships of the various layer-related fields, without distracting from the locking rule, seems too hard otherwise.

(I’ll also do the work of rebasing #2288 on top.)

@giuseppe PTAL.


giuseppe (approved):
LGTM

PR #2294, opened 2024-02-14
Fix manifest updates when we match a layer by TOC digest

mtrmac:
Either return a compressed digest and compressed size, or an uncompressed digest and an uncompressed size (if we are allowed to change the manifest); if neither is possible, ignore the match and force a layer pull.

In almost all cases, a TOC match implies the layer has a TOC digest annotation, i.e. the input manifest is OCI, where the size field is mandatory; so this should not change which layers will be reused
- it just changes the returned size value.

Fixes https://github.com/containers/image/pull/2218#discussion_r1424652040 .

Cc: @giuseppe


giuseppe (approved):
LGTM

PR #2295, opened 2024-02-14
Cleanly fail when trying to obtain a DiffID of a non-OCI image

mtrmac:
… resolving the second part https://github.com/containers/image/pull/2218#discussion_r1470138422 .

Also adds an optimization: We now parse the config only once.

Cc: @giuseppe


giuseppe (approved):
LGTM

PR #2299, opened 2024-02-15
Update a comment after a c/storage update

mtrmac:
This updates for the trivial half of https://github.com/containers/storage/pull/1830.

It does not update to benefit from the more important part (creating the reused layers with the right `CompressedDigest`/`CompressedSize`) — that also should happen, but it’s a tiny bit of _work_ and not an immediate priority.


rhatdan:
Needs a rebase.

mtrmac:
Rebased, now just a comment update.

PR #2301, opened 2024-02-16
storage: use the new ApplyStagedLayer interface

giuseppe:
follow-up for https://github.com/containers/storage/pull/1826


storage/storage_dest.go
```diff
@@ -783,11 +778,15 @@ func (s *storageImageDestination) createNewLayer(index int, layerDigest digest.D
 		options := &graphdriver.ApplyDiffWithDifferOpts{
```
mtrmac, line 778: (Absolutely non-blocking: I wouldn’t mind inlining single-use `options` and `args` users into their users.

This works fine, and it is a smaller patch.)

storage/storage_dest.go
```diff
@@ -783,11 +778,15 @@ func (s *storageImageDestination) createNewLayer(index int, layerDigest digest.D
...
+
+		args := storage.ApplyStagedLayerOptions{
+			ID:          newLayerID,
+			ParentLayer: parentLayer,
+
+			DiffOutput:  diffOutput,
+			DiffOptions: options,
 		}
-		return layer, nil
+		return s.imageRef.transport.store.ApplyStagedLayer(args)
```
mtrmac, line 789: This is pre-existing, but here we need to silently succeed on a `ErrDuplicateID`, similarly to `PutLayer` below.

copy/single.go
```diff
@@ -773,12 +773,12 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, to
...
 				hideProgressBar = false
 				logrus.Debugf("Retrieved partial blob %v", srcInfo.Digest)
-				return true, updatedBlobInfoFromUpload(srcInfo, uploadedBlob)
+				return true, updatedBlobInfoFromUpload(srcInfo, uploadedBlob), nil
 			}
 			logrus.Debugf("Failed to retrieve partial blob: %v", err)
-			return false, types.BlobInfo{}
-		}(); reused {
-			return blobInfo, cachedDiffID, nil
+			return false, types.BlobInfo{}, err
```
mtrmac, line 779: I can see how we might need this for the ComposeFS-only operation, but right now (and still as of https://github.com/containers/storage/pull/1833 ), if we have a chunked layer but c/storage chunked pulling is disabled, AFAICS this would cause an irrecoverable error instead of falling back to the non-chunked layer creation path.

(The authoritative documentation for `PutBlobPartial`, right now, explicitly says that the caller should fall back on any error.)

So it seems to me that this would require a bit more infrastructure to be built. Can this part be split into a separate PR?

mtrmac (commented):
Thanks for working on this!

copy/single.go
mtrmac: This is the cause of the Skopeo test failures .

storage/storage_dest.go
giuseppe: fixed

storage/storage_dest.go
giuseppe: thanks, amended now

copy/single.go
giuseppe: yes let's not rush this change, I've dropped it

mtrmac (approved):
LGTM. Thanks again!

PR #2304, opened 2024-02-20
Support editing ArtifactType, preserve it in lists

nalind:
Add fields to manifest.ListUpdate for adding and updating the ArtifactType field in a list entry.  When copying a list or descriptor, preserve the value from the original descriptor for an entry.

### DIFF
--- a/copy/multiple.go
+++ b/copy/multiple.go
@@ -38,6 +38,7 @@ type instanceCopy struct {
 
 	// Fields which can be used by callers when operation
 	// is `instanceCopyClone`
+	cloneArtifactType       string
 	cloneCompressionVariant OptionCompressionVariant
 	clonePlatform           *imgspecv1.Platform
 	cloneAnnotations        map[string]string
@@ -142,6 +143,7 @@ func prepareInstanceCopies(list internalManifest.List, instanceDigests []digest.
 				res = append(res, instanceCopy{
 					op:                      instanceCopyClone,
 					sourceDigest:            instanceDigest,
+					cloneArtifactType:       instanceDetails.ReadOnly.ArtifactType,
 					cloneCompressionVariant: compressionVariant,
 					clonePlatform:           instanceDetails.ReadOnly.Platform,
 					cloneAnnotations:        maps.Clone(instanceDetails.ReadOnly.Annotations),
@@ -268,6 +270,7 @@ func (c *copier) copyMultipleImages(ctx context.Context) (copiedManifest []byte,
 				AddDigest:                updated.manifestDigest,
 				AddSize:                  int64(len(updated.manifest)),
 				AddMediaType:             updated.manifestMIMEType,
+				AddArtifactType:          instance.cloneArtifactType,
 				AddPlatform:              instance.clonePlatform,
 				AddAnnotations:           instance.cloneAnnotations,
 				AddCompressionAlgorithms: updated.compressionAlgorithms,
--- a/internal/manifest/list.go
+++ b/internal/manifest/list.go
@@ -73,6 +73,7 @@ type ListUpdate struct {
 		Platform                  *imgspecv1.Platform
 		Annotations               map[string]string
 		CompressionAlgorithmNames []string
+		ArtifactType              string
 	}
 }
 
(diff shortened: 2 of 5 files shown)


rhatdan:
LGTM
@mtrmac PTAL
@giuseppe PTAL

giuseppe (approved):
LGTM

internal/manifest/list.go
```diff
@@ -65,9 +65,10 @@ type List interface {
 // ListUpdate includes the fields which a List's UpdateInstances() method will modify.
 // This is publicly visible as c/image/manifest.ListUpdate.
 type ListUpdate struct {
-	Digest    digest.Digest
-	Size      int64
-	MediaType string
+	Digest       digest.Digest
+	Size         int64
+	MediaType    string
+	ArtifactType string
```
mtrmac, line 71: (1) AFAICS this breaks compatibility; previously `UpdateInstances` didn’t modify `ArtifactType`, now existing callers would clear it because they don’t set this field.

Arguably `artifactType` just wasn’t supported until very recently, so there are no existing callers that could have done the right thing, but there already is a bit of a pattern to this, and (2).

---

So I think the `ArtifactType` field should, for now, be added to `ListUpdate.ReadOnly` only. (Ideally, long term, we should have separate types for “data read from an instance” and “edit an instance with these updates”, but that’s not practical to do right now.)

That’s also consistent with the other fields we have added in recent months.

copy/multiple.go
```diff
@@ -32,6 +32,9 @@ type instanceCopy struct {
 	op           instanceCopyKind
 	sourceDigest digest.Digest
 
+	// Fields which we want to preserve
+	artifactType string
```
mtrmac, line 36: (2) it is a bit awkward for `instanceCopy` to be a carrier of data the copy process does not inherently care about.

I’d prefer this to only be a `cloneArtifactType`, in the section below, and for the preservation of the field on “ordinary copies” to be implicit. (Even the `clonePlatform`/`cloneAnnotations` are not really _relevant_ to this copy code, but we have them ~so that we only call `.Instance()` at one place.)

internal/manifest/list.go
```diff
@@ -93,6 +94,7 @@ type ListEdit struct {
 	UpdateDigest                digest.Digest
 	UpdateSize                  int64
 	UpdateMediaType             string
+	UpdateArtifactType          string
```
mtrmac, line 97: We don’t really _need_ this feature right now. And for (1), we would also need to add an `UpdateAffectArtifactType` (or rename existing `UpdateAffectAnnotations` to `UpdateInternalCaller` to cover both fields). It seems simpler to me to just not add this field, for now. (Or do you have short-term plans to use it?)

internal/manifest/testdata/oci1index.json
```diff
@@ -22,6 +22,16 @@
           "sse4"
         ]
       }
+    },
+    {
+      "mediaType": "application/vnd.oci.image.manifest.v1+json",
+      "artifactType": "application/vnd.unknown.artifact.v1",
+      "size": 8221,
+      "digest": "sha256:615e8db5ae085b39e9cd24e5bc887d03fbd30ee4f55f5fedf9b697fafea4fbdd",
```
mtrmac, line 30: Is this new entry used for anything? To exercise the match logic, to ensure the last entry is not chosen if there are similarly-matching alternatives preceding it, perhaps?

internal/manifest/list.go
nalind: Okay, changing it.

copy/multiple.go
nalind: Okay, changing it.

internal/manifest/list.go
nalind: No short-term plans, will drop this field.

internal/manifest/testdata/oci1index.json
nalind: Not originally, but it's about to start being used to ensure that we preserve the `artifactType` when editing a list that has an entry that already has a value.

mtrmac (approved):
Thanks!

Issue #2279, opened 2024-02-05
Allow empty OCI configs for artifacts


saschagrunert:
@containers/image-maintainers I'm wondering if we can allow the `mediaType` `application/vnd.oci.empty.v1+json` when parsing the OCI configs. This would make pushing OCI artifacts a bit simpler for users of ORAS, which uses this as default configuration.

For example in a manifest would look like this:

```json
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.manifest.v1+json",
  "artifactType": "application/vnd.unknown.artifact.v1",
  "config": {
    "mediaType": "application/vnd.oci.empty.v1+json",
    "digest": "sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a",
    "size": 2,
    "data": "e30="
  },
  "layers": [ … ],
```

To allow this media type, we would have to mainly change the checks preventing it:

```diff
diff --git a/internal/image/oci.go b/internal/image/oci.go
index df0e8e41..dc4df568 100644
--- a/internal/image/oci.go
+++ b/internal/image/oci.go
@@ -87,7 +87,7 @@ func (m *manifestOCI1) ConfigBlob(ctx context.Context) ([]byte, error) {
 // layers in the resulting configuration isn't guaranteed to be returned to due how
 // old image manifests work (docker v2s1 especially).
 func (m *manifestOCI1) OCIConfig(ctx context.Context) (*imgspecv1.Image, error) {
-	if m.m.Config.MediaType != imgspecv1.MediaTypeImageConfig {
+	if m.m.Config.MediaType != imgspecv1.MediaTypeImageConfig && m.m.Config.MediaType != imgspecv1.MediaTypeEmptyJSON {
 		return nil, internalManifest.NewNonImageArtifactError(&m.m.Manifest)
 	}
 
diff --git a/manifest/oci.go b/manifest/oci.go
index 6d5acb45..4770f769 100644
--- a/manifest/oci.go
+++ b/manifest/oci.go
@@ -198,7 +198,7 @@ func (m *OCI1) Serialize() ([]byte, error) {
 
 // Inspect returns various information for (skopeo inspect) parsed from the manifest and configuration.
 func (m *OCI1) Inspect(configGetter func(types.BlobInfo) ([]byte, error)) (*types.ImageInspectInfo, error) {
-	if m.Config.MediaType != imgspecv1.MediaTypeImageConfig {
[...]

PR #2306, opened 2024-02-21
Add allowed seccomp specific config media type

saschagrunert:
Adding the new seccomp related config media type `application/vnd.cncf.seccomp-profile.config.v1+json` to the list of allowed configurations.

Alternative to https://github.com/containers/image/issues/2279
Testing in https://github.com/cri-o/cri-o/pull/7804

cc @sftim @mtrmac

### DIFF
--- a/internal/image/oci.go
+++ b/internal/image/oci.go
@@ -87,7 +87,7 @@ func (m *manifestOCI1) ConfigBlob(ctx context.Context) ([]byte, error) {
 // layers in the resulting configuration isn't guaranteed to be returned to due how
 // old image manifests work (docker v2s1 especially).
 func (m *manifestOCI1) OCIConfig(ctx context.Context) (*imgspecv1.Image, error) {
-	if m.m.Config.MediaType != imgspecv1.MediaTypeImageConfig {
+	if !internalManifest.IsSupportedConfigMediaType(m.m.Config.MediaType) {
 		return nil, internalManifest.NewNonImageArtifactError(&m.m.Manifest)
 	}
 
--- a/internal/manifest/manifest.go
+++ b/internal/manifest/manifest.go
@@ -31,6 +31,8 @@ const (
 	DockerV2Schema2ForeignLayerMediaType = "application/vnd.docker.image.rootfs.foreign.diff.tar"
 	// DockerV2Schema2ForeignLayerMediaType is the MIME type used for gzipped schema 2 foreign layers.
 	DockerV2Schema2ForeignLayerMediaTypeGzip = "application/vnd.docker.image.rootfs.foreign.diff.tar.gzip"
+	// CNCFSeccompProfileConfigMediaType is a custom media type used for seccomp profiles.
+	CNCFSeccompProfileConfigMediaType = "application/vnd.cncf.seccomp-profile.config.v1+json"
 )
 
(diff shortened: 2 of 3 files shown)


sftim:
> `application/vnd.cncf.seccomp-profile.config.v1+json`

Any plan to register this?

saschagrunert:
> > `application/vnd.cncf.seccomp-profile.config.v1+json`
> 
> Any plan to register this?

In the [image spec](https://github.com/opencontainers/image-spec/blob/main/specs-go/v1/mediatype.go)? Hm, I'm not sure if that would be accepted. :thinking:

saschagrunert:
@containers/image-maintainers PTAL

mtrmac:
As discussed before in #2279
- I think the larger question of “where / how we store artifacts” is still undecided. The _primary_ design question is “do we use a single namespace in c/storage for artifacts and images, such that we can’t have both an artifact and an image with the same name?”. (And then there’s the implementation effort to somehow store contents of artifacts which are not image-like tar.) I think the first impression is that the answer to this must be “no” (or at least “not always”), for artifacts that store SBOMs and other image-related data which is directly linked to a specific image and should, e.g., be deleted alongside the image.)
- I don’t think this makes _semantic_ sense for this particular artifact. The JSON artifact is not an image, and does not “really” have image config values that make sense to return from `.OCIConfig`. If we _really_ wanted to write the seccomp profiles things to c/storage, it would be more reasonable for the c/image/copy code that calls `.OCIConfig` to avoid doing that for the seccomp profile artifacts. (… But after activity around #2218 , the c/storage “destination” requires! that the image has a ~valid config in the image sense.) What we do here really depends on the local storage design above, but it’s already a point of concern.
- I’d also say that the approach of https://github.com/cri-o/cri-o/pull/7719 , where we need to distribute a _single_ file and end up creating and mounting a possibly-many-layer overlay filesystem, and walk that filesystem to find some file, feels like a prototype that just shouldn’t end up being the production design. There’s _so much_ unnecessary complexity and overhead. I appreciate that this can be done right now, and waiting for a “local artifact storage” design is unattractive, but I don’t like the longer-term cost of forging ahead now.

mtrmac:
Cc: @nalind

saschagrunert:
Okay, I'm going to elaborate on the file based pull approach rather than full c/common libimage. I guess this would also simplify the whole media type topic by accepting anything.

mtrmac:
I do think we need to arrive at a design for “where / how we store artifacts”. I don’t know what’s the best forum and process for that.

saschagrunert:
@mtrmac generally agree, for seccomp itself an in memory solution would be enough because the profile does not need to be on disk. We can create a more sustainable design later one when we want to fulfill more use cases like wasm.

What do we need to do?
1. Parse the user given reference 
1. Create an image source, then retrieve the blobs/layers
1. Pull the blob directly (which can be the seccomp JSON for example) and parse it

I’m not exactly sure how to do that right now with c/image but I’ll find out.

mtrmac:
> 3\. Pull the blob directly (which can be the seccomp JSON for example) and parse it

⚠️💀 Critical:  verify each blob’s digest against the data in the manifest.
Recommended: If reading into memory (and possibly even to disk), impose a hard-coded size limit.

The digest verification alone suggests that c/image should probably provide a convenience helper doing that. (Right now c/image does get care of that, but only in `c/image/copy`. Individual transports should not be in that business.)

PR #2310, opened 2024-02-26
Fix handling of errors when fetching layers by URLs

mtrmac:
`gopls 0.15` identifies always-true/always-false conditions in the code; I broke that in fa54b28a4d7a7e97fe899f958326b2bd3f91a11a .

- Don’t ignore errors accessing layers by URLs. Note that this might break setups which were created in the past two years; they now don’t fall back to registry accesses.
- If we fail, also report all errors, not just the last one, as the original version of the code did.


rhatdan:
LGTM
@giuseppe PTAL

giuseppe (approved):
/lgtm

giuseppe:
@mtrmac can you please rebase?

mtrmac:
Rebased, tests pass.

PR #2312, opened 2024-02-27
Add validation error to digesting reader

saschagrunert:
The reason is hidden if `expectedDigest.Validate()` errors. This patch will add the error message to the stack to provide more details.

PTAL @containers/image-maintainers

### DIFF
--- a/copy/digesting_reader.go
+++ b/copy/digesting_reader.go
@@ -23,11 +23,11 @@ type digestingReader struct {
 func newDigestingReader(source io.Reader, expectedDigest digest.Digest) (*digestingReader, error) {
 	var digester digest.Digester
 	if err := expectedDigest.Validate(); err != nil {
-		return nil, fmt.Errorf("Invalid digest specification %s", expectedDigest)
+		return nil, fmt.Errorf("invalid digest specification %q: %w", expectedDigest, err)
 	}
 	digestAlgorithm := expectedDigest.Algorithm()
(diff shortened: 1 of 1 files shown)


copy/digesting_reader.go
```diff
@@ -23,7 +23,7 @@ type digestingReader struct {
 func newDigestingReader(source io.Reader, expectedDigest digest.Digest) (*digestingReader, error) {
 	var digester digest.Digester
 	if err := expectedDigest.Validate(); err != nil {
-		return nil, fmt.Errorf("Invalid digest specification %s", expectedDigest)
+		return nil, fmt.Errorf("Invalid digest specification %s: %w", expectedDigest, err)
```
mtrmac, line 26: If we are touching this,
```suggestion
		return nil, fmt.Errorf("invalid digest %q: %w", expectedDigest, err)
```
please

mtrmac (commented):
Thanks!

I’m curious, did this happen in practice somewhere?

copy/digesting_reader.go
saschagrunert: Changed as suggested.

saschagrunert:
> I’m curious, did this happen in practice somewhere?

No, just saw that the error got unused.

I also changed the error message below to align with the new style.

mtrmac (approved):
Thanks!

rhatdan:
LGTM

PR #2316, opened 2024-02-28
Remove additional check if digester is available

saschagrunert:
`expectedDigest.Validate()` already calls the `Available()` method and checks if the digest is possible or not. Therefore we can remove the additional/impossible double check for that.

Ref: https://github.com/opencontainers/go-digest/blob/429d0316a/digest.go#L111-L116

Follow-up on https://github.com/containers/image/pull/2312


mtrmac (commented):
You’re right that, as implemented now, this check is redundant.

OTOH `Validate()` does not _document_ that it checks for availability (and, IMHO, it’s a bit surprising that it does so), and when calling `Digester()` on an unavailable algorithm panics, I think the extra check here is somewhat warranted.

So I’d prefer keeping the code as is.

saschagrunert:
Sounds good, thank you!

PR #2317, opened 2024-02-28
Improve handling of zstd vs. zstd:chunked matching

mtrmac:
This is not expected to change observable behavior for now, just improve the internal logic. See individual commit messages for details.

@giuseppe PTAL.


rhatdan:
LGTM

PR #2321, opened 2024-02-29
Record (TOC digest → DiffID) mapping in BlobInfoCache

mtrmac:
A single DiffID may map to multiple TOC digest values. Record that in `BlobInfoCache`, and use it for layer reuse.

Also prefer reusing even TOC-matched layers by DiffID, when available.

@giuseppe I’d appreciate a preliminary review of the new logic; see individual commits.

<s>Draft: The `BlobInfoCache` implementations don’t actually store/record any data yet — so this is obviously completely untested.</s>

### DIFF
--- a/internal/blobinfocache/types.go
+++ b/internal/blobinfocache/types.go
@@ -26,6 +26,15 @@ type BlobInfoCache2 interface {
 	// Close destroys state created by Open().
 	Close()
 
+	// UncompressedDigestForTOC returns an uncompressed digest corresponding to anyDigest.
+	// Returns "" if the uncompressed digest is unknown.
+	UncompressedDigestForTOC(tocDigest digest.Digest) digest.Digest
+	// RecordTOCUncompressedPair records that the tocDigest corresponds to uncompressed.
+	// WARNING: Only call this for LOCALLY VERIFIED data; don’t record a digest pair just because some remote author claims so (e.g.
+	// because a manifest/config pair exists); otherwise the cache could be poisoned and allow substituting unexpected blobs.
+	// (Eventually, the DiffIDs in image config could detect the substitution, but that may be too late, and not all image formats contain that data.)
+	RecordTOCUncompressedPair(tocDigest digest.Digest, uncompressed digest.Digest)
+
 	// RecordDigestCompressorName records a compressor for the blob with the specified digest,
 	// or Uncompressed or UnknownCompression.
 	// WARNING: Only call this with LOCALLY VERIFIED data; don’t record a compressor for a
--- a/storage/storage_dest.go
+++ b/storage/storage_dest.go
@@ -84,7 +84,11 @@ type storageImageDestinationLockProtected struct {
 	currentIndex          int                    // The index of the layer to be committed (i.e., lower indices have already been committed)
 	indexToAddedLayerInfo map[int]addedLayerInfo // Mapping from layer (by index) to blob to add to the image
 
-	// In general, a layer is identified either by (compressed) digest, or by TOC digest.
+	// Externally, a layer is identified either by (compressed) digest, or by TOC digest
+	// (and we assume the TOC digest also uniquely identifies the contents, i.e. there aren’t two
+	// different formats/ways to parse a single TOC); internally, we use uncompressed digest (“DiffID”) or a TOC digest.
+	// We may or may not know the relationships between these three values.
+	//
 	// When creating a layer, the c/storage layer metadata and image IDs must _only_ be based on trusted values
 	// we have computed ourselves. (Layer reuse can then look up against such trusted values, but it might not
 	// recompute those values for incomding layers — the point of the reuse is that we don’t need to consume the incoming layer.)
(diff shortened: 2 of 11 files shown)


internal/blobinfocache/types.go
```diff
@@ -25,6 +25,16 @@ type BlobInfoCache2 interface {
 	// Close destroys state created by Open().
 	Close()
 
+	// UncompressedDigest returns an uncompressed digest corresponding to anyDigest.
+	// Returns "" if the uncompressed digest is unknown.
+	// FIXME: Does this need to record TOC/compression type?
+	UncompressedDigestForTOC(tocDigest digest.Digest) digest.Digest
```
giuseppe, line 31: The TOC digest is the checksum of the uncompressed JSON document, so I think the compression should not matter in this case

mtrmac: I agree we probably don’t need that right now (with `GetTOCDigest` refusing to work on manifests which contain multiple TOC digest annotations, and presumably with the zstd / estargz code being unable to decompress the other one).

This comment is a looking a bit more into the future, for lookups in the other direction, where we will want to look up (UncompressedDigest → (compressed digest, TOC digest, algorithm)) and match that against “the user wants the destination to contain zstd:chunked” (i.e. reject estargz matches).

mtrmac: >  for lookups in the other direction, 

That will be done in a separate data structure (an extension of `RecordDigestCompressorName`: We need the full set of annotations for reuse of a TOC-compressed blob, so this simple mapping is not sufficient anyway. And the other structure does record the algorithm.

storage/storage_dest.go
```diff
@@ -84,18 +84,25 @@ type storageImageDestinationLockProtected struct {
 	currentIndex          int                    // The index of the layer to be committed (i.e., lower indices have already been committed)
 	indexToAddedLayerInfo map[int]addedLayerInfo // Mapping from layer (by index) to blob to add to the image
 
-	// In general, a layer is identified either by (compressed) digest, or by TOC digest.
+	// In general, a layer is identified either by (compressed) digest, or by TOC digest
+	// (and we assume the TOC digest also uniquely identifies the contents, i.e. there aren’t two
+	// different formats/ways to parse a single TOC).
```
mtrmac, line 89: All of the c/storage+c/image code has been built around this assumption, but it is false currently (https://github.com/containers/storage/issues/1888 ) and I’m not sure whether we need to revisit the design. Let’s discuss that in the c/storage issue.

mtrmac (commented):
Note to self: This is code-complete but I want to test it in practice.

storage/storage_dest.go
giuseppe: yes, this assumption is correct

mtrmac:
To test:

Before:
```console
# podman rmi alpine level1 level9
# rm -f /var/lib/containers/cache/blob-info-cache-v1.sqlite 
# podman pull quay.io/libpod/alpine
# podman --log-level=debug push --compression-format zstd:chunked --compression-level 1 --force-compression quay.io/libpod/alpine localhost:50000/level1
## Even better would be to use two different destination registries, to be 100% certain the blobs are not reused
## (right now they are not reused, but we’ll fix that):
# podman--log-level=debug push --compression-format zstd:chunked --compression-level 9 --force-compression quay.io/libpod/alpine localhost:50000/level9
## Note the compressed digest, and TOC digest, values:
# skopeo inspect --raw docker://localhost:50000/level1 | jq .
# skopeo inspect --raw docker://localhost:50000/level9 | jq .
## No DigestTOCUncompressedPairs entries:
# sqlite3 /var/lib/containers/cache/blob-info-cache-v1.sqlite .dump 
# podman rmi alpine level1 level9
## Triggers a partial pull: "Applying differ in …":
# podman --log-level=debug pull localhost:50000/level1
## Triggers a partial pull: "Applying differ in …"
# podman --log-level=debug pull localhost:50000/level9 
## level1 and level9 have different image IDs:
# podman images 
## Contains two copies of the layer, with the same expected-layer-diffid
# jq . < /var/lib/containers/storage/overlay-layers/layers.json ```
```

After:
- `DigestTOCUncompressedPairs` contains 2 records
- Pull of `level1` triggers a partial pull (creating a layer with known TOC digest and uncompressed digest)
- Pull of `level9` reuses the layer (by BIC compressed -> uncompressed mapping)
- FIXME: the layer is shared, but the image not yet - the `hasLayerPulledByTOC` code path is wrong

rhatdan:
LGTM
@giuseppe PTAL

mtrmac:
Specific tests of this feature (as described in https://github.com/containers/image/pull/2321#issuecomment-2078207232 ) are now passing.

Podman tests in https://github.com/containers/podman/pull/23348 are now also passing (well, going to pass, anyway).

Please review again — the `storage_dest.go` part was _significantly_ restructured since the earlier version, to centralize the layer identity and trust logic in a single `trustedLayerIdentityDataLocked` function. See individual commits for a bit more detail.

mtrmac:
Podman tests are now passing in https://github.com/containers/podman/pull/23348 (with https://github.com/containers/podman/pull/23379 for Podman-side updates).

giuseppe:
@mtrmac is it ready to review?

mtrmac:
@giuseppe yes, please review. https://github.com/containers/podman/pull/23348 shows Podman tests passing.

This should generally make the layer reuse on pulls comparable with non-chunked. See also the specific test case above.

giuseppe (approved):
just a nit (probably not even worth a repush), otherwise LGTM

great work!

storage/storage_dest.go
```diff
@@ -84,7 +84,11 @@ type storageImageDestinationLockProtected struct {
 	currentIndex          int                    // The index of the layer to be committed (i.e., lower indices have already been committed)
 	indexToAddedLayerInfo map[int]addedLayerInfo // Mapping from layer (by index) to blob to add to the image
 
-	// In general, a layer is identified either by (compressed) digest, or by TOC digest.
+	// Externally, a layer is identified either by (compressed) digest, or by TOC digest
+	// (and we assume the TOC digest also uniquely identifies the contents, i.e. there aren’t two
+	// different formats/ways to parse a single TOC); internally, we use uncompressed digest (“DiffID”) or a TOC digest.
+	// We may or may not know the relantionships between these three values.
```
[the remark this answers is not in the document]
mtrmac: Thanks! Fixed (and rebased).

rhatdan:
LGTM

PR #2336, opened 2024-03-07
Move to a tagged version of docker/docker

mtrmac:
It's the same commit, but using a tag will prevent the bots from automatically updating to future untagged commits.


vrothberg (approved):
LGTM

PR #2343, opened 2024-03-12
Fix an unintentionally-added dependency on Go 1.21

### LINKED ISSUE #2342: "slices" module only in go 1.21
I [attempted](https://github.com/apptainer/apptainer/pull/2088) to update apptainer to use a commit on the current main branch of containers/image/v5, but when compiling with go 1.20.10 it [got this compilation error](https://github.com/apptainer/apptainer/actions/runs/8251390213/job/22568315383?pr=2088):
```
/go/pkg/mod/github.com/containers/image/v5@v5.30.1-0.20240312145342-38f5de6fa34d/pkg/blobinfocache/internal/prioritize/prioritize.go:6:2: package slices is not in GOROOT (/usr/local/go/src/slices)
```

#2340 said it was about updating to go 1.20, but it started using the "slices" package which is [only in go 1.21](https://pkg.go.dev/slices?tab=versions) and later.

mtrmac:
Should not change behavior.

Fixes #2342 .


TomSweeneyRedHat:
LGTM

PR #2345, opened 2024-03-15
Add support for Docker HealthConfig.StartInterval (v25.0.0+)

migesok:
I'm trying to add support for Dockerfile HEALTHCHECK --start-interval param which has been added in Docker v25.0.0.
So I figured out I have to update this library as well.
The change is quite trivial but I couldn't say that I understand completely what I'm doing so feedback is appreciated.

### DIFF
--- a/manifest/docker_schema2.go
+++ b/manifest/docker_schema2.go
@@ -54,9 +54,10 @@ type Schema2HealthConfig struct {
 	Test []string `json:",omitempty"`
 
 	// Zero means to inherit. Durations are expressed as integer nanoseconds.
-	StartPeriod time.Duration `json:",omitempty"` // StartPeriod is the time to wait after starting before running the first check.
-	Interval    time.Duration `json:",omitempty"` // Interval is the time to wait between checks.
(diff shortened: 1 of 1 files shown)


mtrmac (approved):
Thanks!

This looks about right, and it might be sufficient for some purposes.

For other purposes:
- Docker now uses, and interprets, the fields also in OCI images, not just schema2. So we should be able to decode and use those fields in that situation (… arguably there might be _some_ benefit in intentionally refusing to support that, but that’s probably not ultimately helpful to users)
- And to support both formats, the general c/image abstractions should allow doing that without an image consumer having to specifically support individual manifest types. That would probably mean adding corresponding fields to `types.ImageInspectInfo`, and updating `Inspect` to populate them.

But all of that can come later, this is already valuable as is.

migesok:
@mtrmac thank you! I can add the fields in other places if you think it should be done. But I'm a bit out of my depth here - I haven't found anything "health" related in `types.ImageInspectInfo`.
Either I need a bit more guidance or it's better if someone else does that.

PR #2346, opened 2024-03-18
Filter BlobInfoCache candidates before prioritization, not in transports

mtrmac:
This:
- Moves the point where we filter blob reuse candidates based on algorithm (when the users asks for `gzip`/`zstd`, or when the user specifies a `v2s2` manifest format and Zstd is not acceptable). Before, we would prioritize and return up to 5 candidates regardless of algorithm, and then the transport would filter unwanted algorithms — in the worst case, getting not a single acceptable algorithm. Now, we first filter out unacceptable candidates, and only afterwards prioritize and return up to 5 candidates to the transport. That should improve the chances of reusing a known blob.
- Moves the responsibility for turning compression algorithm names into internal objects from the transports to the `BlobInfoCache2` implementation. This will, again, allow using `c/image/docker` as a ~light-weight registry client without dragging in `c/storage/pkg/chunked`.
- Structures the code to give us an _opportunity_ to add even more logic to the algorithm filtering/matching. E.g. we will want to return a known-zstd:chunked layer where we don’t trust the TOC as a zstd candidate; and we will want to return trusted zstd:chunked candidates where the user asked for just zstd. That will come in the future.

Cc: @giuseppe


giuseppe (approved):
LGTM

rhatdan:
LGTM

PR #2365, opened 2024-04-03
Add a helper for formatting multiple errors

mtrmac:
As of Go 1.20, the standard library supposedly contains infrastructure for reporting multiple errors.

Sadly, that infrastructure is mostly unhelpful: `errors.Join` gives no control over formatting, and `fmt.Errorf` does not support a `[]error` input. So, introduce our own `multierr.Format`, and use it where possible.

Then remove the `go-multierror` dependency. The net result is not particularly better _code_, but at least we try to work with the standard library design.

(This was originally motivated just by trying to remove the `go-multierror` dependency in #2364, now that the standard library supposedly supports the feature.)

This might, in some cases, change whether the returned error matches `errors.{As,Is}`, mostly because the previous code for aggregating multiple errors was written before Go 1.20 made it possible for multiple errors to match. (Not that I know how the multi-matching design can ever be all that useful…)

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/image/v5/internal/iolimits"
+	"github.com/containers/image/v5/internal/multierr"
 	"github.com/containers/image/v5/internal/set"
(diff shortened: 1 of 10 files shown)


rhatdan:
LGTM
@giuseppe PTAL

giuseppe (approved):
/lgtm

PR #2375, opened 2024-04-11
use containers/storage/pkg/fileutils/(Exists,Lexists) 

giuseppe:
use a faster alternative to os.Stat and os.Lstat to check whether a path is present


mtrmac (approved):
Thanks!

PR #2377, opened 2024-04-15
Update to Go1.21

mtrmac:
… primarily because some dependencies started to require it: if we ever needed to quickly update a dependency for a vulnerability fix, we might have to update to Go 1.21 on a short notice, or fork the dependency.

And then bump dependencies which we have been holding back:
- `github.com/sylabs/sif/v2`: https://github.com/containers/image/pull/2356
- `github.com/sigstore/fulcio`: https://github.com/containers/image/pull/2320 https://github.com/containers/image/pull/2371
- `github.com/sigstore/rekor`: https://github.com/containers/image/pull/2100 https://github.com/containers/image/pull/2139 https://github.com/containers/image/pull/2141 https://github.com/containers/image/pull/2169 https://github.com/containers/image/pull/2192 https://github.com/containers/image/pull/2208 https://github.com/containers/image/pull/2278 https://github.com/containers/image/pull/2292 https://github.com/containers/image/pull/2362
- `github.com/docker/docker`: https://github.com/containers/image/pull/2353 (+ an update to 26.0.1)
- `github.com/docker/cli`: https://github.com/containers/image/pull/2352 (+ an update to 26.0.1)

Cc: @jnovy @lsm5 @cevich @TomSweeneyRedHat


mtrmac:
Note to self: Afterwards we should github.com/letsencrypt/boulder to avoid a dependency on no-longer-maintained `gopkg.in/go-jose/go-jose.v2`. (We don’t call it though that path AFAIK, still, we include it.)

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/containers/image/v5
 
-go 1.20
+go 1.21
```
cevich, line 3: My understanding is that along with this, we also need toolchain 1.21 as well? Please double-check.

mtrmac: My reading of “Go toolchain selection” in https://go.dev/doc/toolchain is that it is not compulsory, and effectively defaults to the `go` directive.

I don’t mind adding `1.21.0` here, if you think it would be more explicit. (Or 1.21.$latest ? That would force users to update.)

mtrmac:
Cross-linking for visibility: https://github.com/containers/skopeo/pull/2297#issuecomment-2057502035 says we need an update to the Renovate config (affecting also repos using c/image).

mtrmac:
(Marking as draft to make sure we have a decision the Renovate configuration first; I understand Renovate would immediately start failing otherwise.)

cevich:
(cross-post / copy)

I don't believe it will fail, it will simply (and mostly silently) refuse to open any module update PRs which also require a bump of go 1.21.

cevich:
Ref: https://github.com/containers/automation/pull/189

rhatdan:
LGTM

PR #2379, opened 2024-04-16
Avoid a redundant function call

mtrmac:
This primarily ensures that we log and use the same value.

Should not change behavior.


mtrmac:
Added one more trivial cleanup.

giuseppe (approved):
LGTM

rhatdan:
LGTM

PR #2380, opened 2024-04-17
Refactor `blobCacheDestination.saveStream`

mtrmac:
Prefer `defer`, and early returns, and straight-line code.

Effectively this is just one small bug fix (a `nil` reference on an error path)

The problem this is aiming to address is that nothing between initially creating `err3` and the code that removes the temporary file is allowed to `return`; i.e. adding new reasons for the function to fail is hard.


rhatdan:
LGTM
@giuseppe PTAL

giuseppe (approved):
/lgtm

PR #2387, opened 2024-04-26
Fix GoDoc link at the top of the README file

ananthb:
I was really puzzled by the broken documentation at the current link. This version bump should fix it.


mtrmac (commented):
Thanks!

Could you update the links further, to point directly at the `pkg.go.dev` host name?

ananthb:
I created the badge link again with their badge generation tool.

ananthb:
I also cleaned up the README language and links a bit here: https://github.com/containers/image/commit/9bc780afb6f6f1d52a0aab5b3343e40bbcdfd2a6.

I can add that commit to this PR if that works for you.

mtrmac:
> I also cleaned up the README language and links a bit here

Some parts of that (like updating for the link redirect) are clearly valuable in isolation.

OTOH parts of the README description are to a large extent a historical artifact of how this library was split from `skopeo`, and should ideally be replaced wholesale — so incremental improvements to the wording would not be the best use of anyone’s time. (Not that I can immediately prioritize those wholesale replacements, either…)

PR #2390, opened 2024-05-02
storage: cleanup staged layer if unused

giuseppe:
notify c/storage that the staged layer is not used and must be cleaned up on error.

### DIFF
--- a/storage/storage_dest.go
+++ b/storage/storage_dest.go
@@ -308,6 +308,12 @@ func (s *storageImageDestination) PutBlobPartial(ctx context.Context, chunkAcces
 	if err != nil {
 		return private.UploadedBlob{}, err
 	}
+	succeeded := false
+	defer func() {
+		if !succeeded {
+			_ = s.imageRef.transport.store.CleanupStagedLayer(out)
+		}
+	}()
 
 	if out.TOCDigest == "" && out.UncompressedDigest == "" {
 		return private.UploadedBlob{}, errors.New("internal error: ApplyDiffWithDiffer succeeded with neither TOCDigest nor UncompressedDigest set")
@@ -330,6 +336,7 @@ func (s *storageImageDestination) PutBlobPartial(ctx context.Context, chunkAcces
 	s.lockProtected.diffOutputs[options.LayerIndex] = out
 	s.lock.Unlock()
 
+	succeeded = true
 	return private.UploadedBlob{
 		Digest: blobDigest,
 		Size:   srcInfo.Size,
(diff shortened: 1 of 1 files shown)


storage/storage_dest.go
```diff
@@ -312,6 +312,7 @@ func (s *storageImageDestination) PutBlobPartial(ctx context.Context, chunkAcces
 	}
 
 	if out.TOCDigest == "" && out.UncompressedDigest == "" {
+		_ = s.imageRef.transport.store.CleanupStagedLayer(out)
```
mtrmac, line 315: `storageImageDestination.Close` has this conditioned on `out.Target != nil`. Is that necessary here as well?

Quite possibly `storageImageDestination.Close` should just drop the condition instead.

mtrmac (commented):
Thanks!

This works, but it would be easy to forget next time again. 

So I’m leaning towards
```go
out, err := …
if err != nil { … }
succeeded := false
defer func() {
    if !succeeded { … CleanupStagedLayer … }
}()
…
…diffOutputs = …
s.lock.Unlock()
succeeded = true
return …
```

If you think that’s never going to be necessary, *shrug*, this PR is an improvement as is.

---

But, also, that code path should _never_ be hit. Is that assumption incorrect?

storage/storage_dest.go
giuseppe: yes let's drop the condition from `storageImageDestination.Close` and let's c/storage deal with the cleanup

giuseppe:
> If you think that’s never going to be necessary, _shrug_, this PR is an improvement as is.

your version is better, I've fixed it and pushed a new version

mtrmac (commented):
If I understand correctly, this is to follow the rules, allowing us to more precisely use locks in https://github.com/containers/storage/pull/1916 .

But right now there’s https://github.com/containers/storage/pull/1916#discussion_r1587936966 , so I’d prefer to wait a bit here until the that question is resolved.

giuseppe:
containers/storage#1916 is merged, can we reconsider this?

mtrmac (approved):
AFAICS, on return from `ApplyDiffFromStagingDirectory`, the callers can’t tell whether the lock is now unlocked.

But, also, the locks only exist for directories created by `MkdirTemp`, and the unlock code is ~idempotent, so extra calls of `CleanupStagedLayer` don’t actually hurt.

LGTM.

TomSweeneyRedHat:
LGTM
once the tests are hip

PR #2391, opened 2024-05-02
docker: support for requesting chunks without end offset

giuseppe:
if the specified length for the range is set to -1, then request all the data possible by using the "Range: <unit>=<range-start>-" syntax for the HTTP range.

### DIFF
--- a/copy/progress_bars.go
+++ b/copy/progress_bars.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"math"
 	"time"
 
 	"github.com/containers/image/v5/internal/private"
@@ -151,12 +152,18 @@ type blobChunkAccessorProxy struct {
 // The specified chunks must be not overlapping and sorted by their offset.
 // The readers must be fully consumed, in the order they are returned, before blocking
 // to read the next chunk.
+// If the Length for the last chunk is set to math.MaxUint64, then it
+// fully fetches the remaining data from the offset to the end of the blob.
 func (s *blobChunkAccessorProxy) GetBlobAt(ctx context.Context, info types.BlobInfo, chunks []private.ImageSourceChunk) (chan io.ReadCloser, chan error, error) {
 	start := time.Now()
 	rc, errs, err := s.wrapped.GetBlobAt(ctx, info, chunks)
 	if err == nil {
 		total := int64(0)
 		for _, c := range chunks {
+			// do not update the progress bar if there is a chunk with unknown length.
+			if c.Length == math.MaxUint64 {
+				return rc, errs, err
+			}
 			total += int64(c.Length)
 		}
 		s.bar.EwmaIncrInt64(total, time.Since(start))
--- a/docker/docker_image_src.go
+++ b/docker/docker_image_src.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"mime"
 	"mime/multipart"
 	"net/http"
@@ -260,9 +261,15 @@ func splitHTTP200ResponseToPartial(streams chan io.ReadCloser, errs chan error, 
 			}
 			currentOffset += toSkip
 		}
+		var reader io.Reader
+		if c.Length == math.MaxUint64 {
+			reader = body
+		} else {
+			reader = io.LimitReader(body, int64(c.Length))
+		}
 		s := signalCloseReader{
 			closed:        make(chan struct{}),
-			stream:        io.NopCloser(io.LimitReader(body, int64(c.Length))),
+			stream:        io.NopCloser(reader),
 			consumeStream: true,
 		}
 		streams <- s
@@ -343,12 +350,24 @@ func parseMediaType(contentType string) (string, map[string]string, error) {
 // The specified chunks must be not overlapping and sorted by their offset.
 // The readers must be fully consumed, in the order they are returned, before blocking
 // to read the next chunk.
+// If the Length for the last chunk is set to math.MaxUint64, then it
+// fully fetches the remaining data from the offset to the end of the blob.
 func (s *dockerImageSource) GetBlobAt(ctx context.Context, info types.BlobInfo, chunks []private.ImageSourceChunk) (chan io.ReadCloser, chan error, error) {
 	headers := make(map[string][]string)
 
 	rangeVals := make([]string, 0, len(chunks))
+	lastFound := false
 	for _, c := range chunks {
-		rangeVals = append(rangeVals, fmt.Sprintf("%d-%d", c.Offset, c.Offset+c.Length-1))
+		if lastFound {
+			return nil, nil, fmt.Errorf("internal error: another chunk requested after an util-EOF chunk")
+		}
+		// If the Length is set to -1, then request anything after the specified offset.
+		if c.Length == math.MaxUint64 {
+			lastFound = true
+			rangeVals = append(rangeVals, fmt.Sprintf("%d-", c.Offset))
+		} else {
+			rangeVals = append(rangeVals, fmt.Sprintf("%d-%d", c.Offset, c.Offset+c.Length-1))
+		}
 	}
 
 	headers["Range"] = []string{fmt.Sprintf("bytes=%s", strings.Join(rangeVals, ","))}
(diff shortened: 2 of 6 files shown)


mtrmac (changes requested):
Thanks!

- How this is happening? I can’t see any c/storage caller setting this to -1 .
- The `Length == MaxUint64` semantics is, AFAICS, not documented, and at least `mergeMissingChunks` is doing ordinary arithmetic on the values.
- If this should exist at all,  `blobChunkAccessorProxy`, `splitHTTP200ResponseToPartial`, `streamChunksFromFile` do not handle the special case.

giuseppe:
the error I am seeing is that we pass the value `-1` to 

> * How this is happening? I can’t see any c/storage caller setting this to -1 .

the problem is here https://github.com/containers/storage/blob/main/pkg/chunked/storage_linux.go#L1611-L1617

in some cases the blobSize is set to -1 from c/image, e.g. when pulling `quay.io/libpod/alpine-with-seccomp:label`.

I could change the code to not use any range (if there are no ranges, then request the whole file), but I thought this would be more flexible.  We can restrict it so that only the last chunk can have len=-1

mtrmac:
OK, so this is
- a schema1 image (unknown layer size)
- the transparent conversion path (needs to read the full file)
- `ImageSourceSeekable` not having a “read all” method at all.

Pedantically, when the size is unknown, while `GetDiffer` requires it (all three implementations currently do, and there is no “unknown” value documented, the locally correct thing to do is for `PutBlobPartial` to fail without even trying.

If you do want the conversion path to work on these images… yes, I think adding the “length unknown” case to `ImageSourceSeekable` / `BlobChunkAccessor` is the simplest approach. (Notably `blobChunkAccessorProxy` is simpler when there is only one API to call, having to add a whole separate “no-partial download” case to that would not be worth it). So, document the field value, allow it only for the last chunk.

----

But, at a higher level, this is really another argument to say that the transparent conversion should not be via `PutBlobPartial`/`GetDiffer` at all: let it work like an ordinary non-chunked pull, letting c/image deal with the whole blob and digest verification, and then do the conversion all the way down in overlay’s `ApplyDiff`.

giuseppe:
> OK, so this is
> 
> * a schema1 image (unknown layer size)
> * the transparent conversion path (needs to read the full file)
> * `ImageSourceSeekable` not having a “read all” method at all.
> 
> Pedantically, when the size is unknown, while `GetDiffer` requires it (all three implementations currently do, and there is no “unknown” value documented, the locally correct thing to do is for `PutBlobPartial` to fail without even trying.
> 
> If you do want the conversion path to work on these images… yes, I think adding the “length unknown” case to `ImageSourceSeekable` / `BlobChunkAccessor` is the simplest approach. (Notably `blobChunkAccessorProxy` is simpler when there is only one API to call, having to add a whole separate “no-partial download” case to that would not be worth it). So, document the field value, allow it only for the last chunk.
> 
> But, at a higher level, this is really another argument to say that the transparent conversion should not be via `PutBlobPartial`/`GetDiffer` at all: let it work like an ordinary non-chunked pull, letting c/image deal with the whole blob and digest verification, and then do the conversion all the way down in overlay’s `ApplyDiff`.

we need to be able to pull via the `PutBlobPartial` when `convert_images` is enabled.  composefs needs it to calculate the digests and enable fs-verity for each file.

mtrmac:
As far as c/image is concerned, with `UncompressedDigest` known, the layer’s `singleLayerIDComponent` is going to be the DiffID, and the ComposeFS usage is ~invisible to c/image. I naively imagine that whether the call stack says `GetDiffer` or `ApplyDiff` does not really matter for fs-verity and other work.

---

One change that _does_ make a difference, however, is that `GetDiffer` is parallelized over up to 6 layers concurrently, but layer commit is sequential. That might well be a very strong reason to do the conversion in `GetDiffer`, not in the commit phase, I didn’t try to measure that.

If this does help, arguably the layer extraction should be similarly made concurrent for non-chunked/non-ComposeFS layers as well. But that redesign would really be a very different conversation (and we might end up deprecating non-ComposeFS layers before that).

giuseppe:
I've tried adding a new method `GetBlob` to retrieve the entire blob, but I think the current API is more flexible, with a minimal change.

I've updated the comments

copy/progress_bars.go
```diff
@@ -148,6 +148,8 @@ type blobChunkAccessorProxy struct {
 // The specified chunks must be not overlapping and sorted by their offset.
 // The readers must be fully consumed, in the order they are returned, before blocking
 // to read the next chunk.
+// If the Length for the last chunk is set to math.MaxUint64, then it
+// fully fetches the remaining data from the offset to the end of the blob.
 func (s *blobChunkAccessorProxy) GetBlobAt(ctx context.Context, info types.BlobInfo, chunks []private.ImageSourceChunk) (chan io.ReadCloser, chan error, error) {
 	start := time.Now()
 	rc, errs, err := s.wrapped.GetBlobAt(ctx, info, chunks)
```
mtrmac, line 155: The implementation below needs updating as well.

pkg/blobcache/src.go
```diff
@@ -236,6 +236,8 @@ func (s signalCloseReader) Close() error {
 // The specified chunks must be not overlapping and sorted by their offset.
 // The readers must be fully consumed, in the order they are returned, before blocking
 // to read the next chunk.
+// If the Length for the last chunk is set to math.MaxUint64, then it
+// fully fetches the remaining data from the offset to the end of the blob.
```
mtrmac, line 240: The implementation (`streamChunksFromFile`) needs updating as well.

internal/private/private.go
```diff
@@ -154,6 +154,8 @@ type BlobChunkAccessor interface {
 	// The specified chunks must be not overlapping and sorted by their offset.
 	// The readers must be fully consumed, in the order they are returned, before blocking
 	// to read the next chunk.
+	// If the Length for the last chunk is set to math.MaxUint64, then it
+	// fully fetches the remaining data from the offset to the end of the blob.
```
mtrmac, line 158: (Documenting it in the `ImageSourceChunk.Length` field might be helpful, and perhaps sufficient.)

docker/docker_image_src.go
```diff
@@ -338,12 +345,24 @@ func parseMediaType(contentType string) (string, map[string]string, error) {
...
+// fully fetches the remaining data from the offset to the end of the blob.
 func (s *dockerImageSource) GetBlobAt(ctx context.Context, info types.BlobInfo, chunks []private.ImageSourceChunk) (chan io.ReadCloser, chan error, error) {
 	headers := make(map[string][]string)
 
 	rangeVals := make([]string, 0, len(chunks))
+	lastFound := false
 	for _, c := range chunks {
-		rangeVals = append(rangeVals, fmt.Sprintf("%d-%d", c.Offset, c.Offset+c.Length-1))
+		if lastFound {
+			return nil, nil, fmt.Errorf("invalid chunk requested: %v", c)
```
mtrmac, line 357: There might not be anything invalid about `c` itself.

Something like `“internal error: another chunk requested after an util-EOF chunk”`?

docker/docker_image_src.go
```diff
@@ -255,9 +256,15 @@ func splitHTTP200ResponseToPartial(streams chan io.ReadCloser, errs chan error,
 			}
 			currentOffset += toSkip
 		}
+		var reader io.Reader
+		if c.Length == math.MaxUint64 {
+			reader = body
```
mtrmac, line 261: Non-blocking:
```suggestion
			reader = body // The caller has ensured that this is the last requested chunk
```

mtrmac (commented):
Thanks. The existing code LGTM, but other parts need updating as well.

giuseppe:
thanks, addressed your comments in the last version

rhatdan:
Seems like there is one comment by @mtrmac that was not addressed?

giuseppe:
> Seems like there is one comment by @mtrmac that was not addressed?

I thought I've addressed them all, which one is still missing?

rhatdan:
https://github.com/containers/image/pull/2391/files#r1589601556

giuseppe:
> https://github.com/containers/image/pull/2391/files#r1589601556

Thanks.  Fixed now

mtrmac (approved):
Thanks!

PR #2403, opened 2024-05-09
Fix CVE-2024-3727

mtrmac:


mheon:
LGTM

Mingli-001:
Could you help to backport the patch to the branch 5.22 https://github.com/containers/image/tree/release-5.22? Thanks!

mtrmac:
@Mingli-001 in this upstream repo, over a year later, I don’t think we would be backporting this to those old branches. If you are a Red Hat customer, please use the standard customer support channels; or similarly for any other vendor.

PR #2404, opened 2024-05-09
[release-5.30] Fix CVE-2024-3727

mtrmac:

In less common uses of this library (using other transports or not using the containers/image.copy.Image API), an attacker could also trigger local path traversals or crashes.


mtrmac:
This is a backport of #2403 + #2380 .

mheon:
LGTM

PR #2405, opened 2024-05-09
[release-5.30] Release 5.30.1

mtrmac:
This fixes CVE-2024-3727 .


mtrmac:
Cc: @TomSweeneyRedHat

TomSweeneyRedHat:
LGTM
once the tests are hip

PR #2406, opened 2024-05-09
Allow using recent opencontainers/go-digest

mtrmac:
We would like to update opencontainers/go-digest to include https://github.com/opencontainers/go-digest/pull/101 , if it is accepted, to further protect against things like  CVE-2024-3727 .

The upstream branch has deprecated the `Hex` method, so migrate away. Also ensure we can work with the proposed digest validation.


rhatdan:
LGTM
@giuseppe @mheon @Luap99 @baude PTAL

pkg/blobinfocache/sqlite/sqlite.go
```diff
@@ -229,7 +229,7 @@ func ensureDBHasCurrentSchema(db *sql.DB) error {
 	// * This schema uses the text representation of digests.
 	//
 	//   We use the fairly wasteful text with hexadecimal digits because digest.Digest does not define a binary representation;
-	//   and the way digest.Digest.Hex() is deprecated in favor of digest.Digest.Encoded(), and the way digest.Algorithm
+	//   and the way digest.Digest.Encoded() is deprecated in favor of digest.Digest.Encoded(), and the way digest.Algorithm
```
mheon, line 232: This seems incorrect

mtrmac: Good catch, thanks!

giuseppe (approved):
LGTM

PR #2408, opened 2024-05-09
Quote various strings coming from untrusted sources

mtrmac:
Typically, use `%q` instead of `%s` (or instead of `"%s"`), to expose various control characters and the like without interpreting them.

This is not really comprehensive; the codebase makes no _general_ guarantee that any returned string values are free of control characters or other malicious/misleading metadata. Not even in returned "error" values (which can legitimately contain newlines, if nothing else).

A side effect of the code audit required by CVE-2024-3727 .

### DIFF
--- a/copy/manifest.go
+++ b/copy/manifest.go
@@ -74,7 +74,7 @@ func determineManifestConversion(in determineManifestConversionInputs) (manifest
 	srcType := in.srcMIMEType
 	normalizedSrcType := manifest.NormalizedMIMEType(srcType)
 	if srcType != normalizedSrcType {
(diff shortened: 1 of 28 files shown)


mtrmac:
See https://github.com/containers/skopeo/pull/2325 for updated Skopeo tests.

cgwalters (approved):
I glanced through this and the changes I saw looked good, but I made no attempt to check or look for places that weren't changed.

rhatdan:
LGTM

mtrmac:
There is a link to an updated version above.

PR #2416, opened 2024-05-14
[Additional Layer Store] Use TOCDigest as ID of each layer (patch for c/image)

ktock:
Needs: https://github.com/containers/storage/pull/1924

This commit changes the store implementation to use TOCDigset as the ID of each layer. This ensures the runtime to pass TOCDigest to the FUSE backend everytime requiring a layer.

This commit modifies c/storage's `LookupAdditionalLayer` to receive a TOCDigest as one of the arguments. Then that TOCDigest is passed to Additional Layer Store as one of the path elements of the FUSE fs. For example, the extracted view of a layer diff contents is provided in the following path:

```
 <mountpoint>/base64(imageref)/<TOCDigest>/diff
```

Additional Layer Store implementation attempts lazy pulling of the layer that has the specified TOCDigest in the image.
If that image doesn't contain the layer matching to the TOCDigest, accessing to the layer directly results in an error.

Example draft implementation of Additional Layer Store is at https://github.com/containerd/stargz-snapshotter/pull/1673


storage/storage_dest.go
```diff
@@ -365,13 +365,15 @@ func (s *storageImageDestination) tryReusingBlobAsPending(blobDigest digest.Dige
...
 		// Check if we have the layer in the underlying additional layer store.
-		aLayer, err := s.imageRef.transport.store.LookupAdditionalLayer(blobDigest, options.SrcRef.String())
+		aLayer, err := s.imageRef.transport.store.LookupAdditionalLayer(options.TOCDigest, options.SrcRef.String())
 		if err != nil && !errors.Is(err, storage.ErrLayerUnknown) {
 			return false, private.ReusedBlob{}, fmt.Errorf(`looking for compressed layers with digest %q and labels: %w`, blobDigest, err)
 		} else if err == nil {
-			s.lockProtected.blobDiffIDs[blobDigest] = aLayer.UncompressedDigest()
+			if d := aLayer.TOCDigest(); d != "" && options.LayerIndex != nil {
+				s.lockProtected.indexToTOCDigest[*options.LayerIndex] = d
+			}
```
mtrmac, line 376: - This _must_ define a layer ID as documented in `storageImageDestinationLockProtected`. If no suitable value is available (can that ever happen??), we just can’t report success here, and we must fall through to the other reuse / non-reuse paths.
- The layer ID  _must not_ be based on an untrusted value. If the only value the backend is using is the TOC digest, we must not use `blobDigest` for any purpose.
- So, `blobAdditionalLayer` should be indexed either by the TOC digest, or by by `LayerIndex`; and the code paths can only be reachable if that data is available.

storage/storage_src.go
```diff
@@ -302,9 +302,6 @@ func (s *storageImageSource) LayerInfosForCopy(ctx context.Context, instanceDige
 		if err != nil {
 			return nil, fmt.Errorf("reading layer %q in image %q: %w", layerID, s.image.ID, err)
 		}
-		if layer.UncompressedSize < 0 {
-			return nil, fmt.Errorf("uncompressed size for layer %q is unknown", layerID)
-		}
```
mtrmac, line 307: FIXME

storage/storage_dest.go
```diff
@@ -365,13 +365,15 @@ func (s *storageImageDestination) tryReusingBlobAsPending(blobDigest digest.Dige
 	s.lock.Lock()
 	defer s.lock.Unlock()
 
-	if options.SrcRef != nil {
+	if options.SrcRef != nil && options.TOCDigest != "" && options.TOCDigest.Validate() == nil {
```
mtrmac, line 368: Please rebase – the `Validate` check now happens before this line.

mtrmac (changes requested):
Thanks!

Just a very quick first skim.

storage/storage_src.go
ktock: Thanks for the comment. According to the docs, unknown BlobInfo.Size seems to be allowed: https://github.com/containers/image/blob/6caf212454493ef1979393c398ac705aebb393a6/types/types.go#L124

storage/storage_dest.go
ktock: rebased

storage/storage_dest.go
ktock: Thanks for the comment. Fixed.

TomSweeneyRedHat:
@ktock You're fighting some lint errors on this one.

ktock:
Thanks for the comment. Fixed the linter error.

storage/storage_src.go
```diff
@@ -309,7 +309,7 @@ func (s *storageImageSource) LayerInfosForCopy(ctx context.Context, instanceDige
 			return nil, fmt.Errorf("reading layer %q in image %q: %w", layerID, s.image.ID, err)
 		}
 		if layer.UncompressedSize < 0 {
-			return nil, fmt.Errorf("uncompressed size for layer %q is unknown", layerID)
+			layer.UncompressedSize = -1
```
mtrmac, line 312: The c/storage API does not document that the returned `layer` object can be modified by consumers. (In practice, c/storage makes a copy, so it works fine.)

storage/storage_src.go
```diff
@@ -453,9 +453,12 @@ func (s *storageImageSource) getSize() (int64, error) {
 		if err != nil {
 			return -1, err
 		}
-		if (layer.TOCDigest == "" && layer.UncompressedDigest == "") || layer.UncompressedSize < 0 {
+		if (layer.TOCDigest == "" && layer.UncompressedDigest == "") || (layer.TOCDigest == "" && layer.UncompressedSize < 0) {
 			return -1, fmt.Errorf("size for layer %q is unknown, failing getSize()", layerID)
 		}
+		if layer.UncompressedSize < 0 {
+			sum = 0
```
mtrmac, line 460: This means that if there 3 layers, and the middle one has an unknown size, the total returned size of the image is… some value larger than the true size, but the particular value can AFAICS not be reasonably used for anything.

Why is that better than returning `-1` to indicate “unknown”, or than returning an error? (One possible answer might be “Podman callers assume this always succeeds and returns a non-`-1` value, I didn’t check whether that’s the case. But I think, if that were true, the callers would need to be updated to deal with images where the size is unknown, instead.)

storage/storage_src.go
```diff
@@ -453,9 +453,12 @@ func (s *storageImageSource) getSize() (int64, error) {
 		if err != nil {
 			return -1, err
 		}
-		if (layer.TOCDigest == "" && layer.UncompressedDigest == "") || layer.UncompressedSize < 0 {
+		if (layer.TOCDigest == "" && layer.UncompressedDigest == "") || (layer.TOCDigest == "" && layer.UncompressedSize < 0) {
 			return -1, fmt.Errorf("size for layer %q is unknown, failing getSize()", layerID)
 		}
+		if layer.UncompressedSize < 0 {
```
mtrmac, line 459: If a TOC-identified layer can have `UncompressedSize == -1`, that breaks more assumptions. At least the `LayersByTOCDigest` path in `tryReusingBlobAsPending` is using the `UncompressedSize` value, and the value can’t be “unknown” on that path.

… but, also, note to self: is a _new_ behavior of the `AdditionalLayerStore` code? We did previously assume that the blobDigest-identified ALS layers did have a valid `UncompressedSize`. Was that always broken?

storage/storage_src.go
mtrmac: More importantly, the documentation of `LayerInfosForCopy` does allow the size to be not returned.

But that seemed surprising to me — hence the FIXME — because in modern manifest formats, the size is mandatory. So that FIXME was a note to self… and it turns out that all of the blob copy path can deal with `LayerInfosForCopy` not carrying a size fine, except for the code path where the layer is using a remote URL and we don’t copy anything.

This function never includes an `URLs` value (discarding one if it were present in the original manifest), so we are fine.

storage/storage_dest.go
```diff
@@ -377,14 +377,18 @@ func (s *storageImageDestination) tryReusingBlobAsPending(blobDigest digest.Dige
...
 		// Check if we have the layer in the underlying additional layer store.
-		aLayer, err := s.imageRef.transport.store.LookupAdditionalLayer(blobDigest, options.SrcRef.String())
+		aLayer, err := s.imageRef.transport.store.LookupAdditionalLayer(options.TOCDigest, options.SrcRef.String())
 		if err != nil && !errors.Is(err, storage.ErrLayerUnknown) {
 			return false, private.ReusedBlob{}, fmt.Errorf(`looking for compressed layers with digest %q and labels: %w`, blobDigest, err)
 		} else if err == nil {
-			s.lockProtected.blobDiffIDs[blobDigest] = aLayer.UncompressedDigest()
-			s.lockProtected.blobAdditionalLayer[blobDigest] = aLayer
+			d := aLayer.TOCDigest()
+			if d == "" {
```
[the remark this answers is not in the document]
mtrmac: Per the other comment, this can happen if the FUSE server is old and not expecting TOCs at all.

storage/storage_src.go
mtrmac: > At least the `LayersByTOCDigest` path in `tryReusingBlobAsPending` is using the `UncompressedSize` value, and the value can’t be “unknown” on that path.

That’s actually fine because that is already conditional on `UncompressedDigest != ""`, and that’s false for recent ALS layers.

mtrmac (commented):
Merging now because otherwise c/image does not compile after the c/storage API change; and because, as of today, https://github.com/containerd/stargz-snapshotter/pull/1673 is not merged, so we actually never _create_ ALS layers after merging this, and most of the concerns become hypothetical (in the short term).

I’ll immediately follow up with a PR that mitigates some of the concerns, or at least adds explicit FIXMEs. That’s not to preclude any other work on improving those aspects, just to get us back to a production-ready state.

storage/storage_dest.go
```diff
@@ -377,14 +377,18 @@ func (s *storageImageDestination) tryReusingBlobAsPending(blobDigest digest.Dige
...
+	if options.SrcRef != nil && options.TOCDigest != "" && options.LayerIndex != nil {
 		// Check if we have the layer in the underlying additional layer store.
-		aLayer, err := s.imageRef.transport.store.LookupAdditionalLayer(blobDigest, options.SrcRef.String())
+		aLayer, err := s.imageRef.transport.store.LookupAdditionalLayer(options.TOCDigest, options.SrcRef.String())
 		if err != nil && !errors.Is(err, storage.ErrLayerUnknown) {
 			return false, private.ReusedBlob{}, fmt.Errorf(`looking for compressed layers with digest %q and labels: %w`, blobDigest, err)
 		} else if err == nil {
-			s.lockProtected.blobDiffIDs[blobDigest] = aLayer.UncompressedDigest()
-			s.lockProtected.blobAdditionalLayer[blobDigest] = aLayer
+			d := aLayer.TOCDigest()
```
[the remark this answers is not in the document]
mtrmac: #2428 modifies this to only trust `options.TOCDigest` (but to require the one from the ALS to match).

storage/storage_src.go
mtrmac: Should be fixed in #2428 .

storage/storage_src.go
mtrmac: … and I don’t know whether `UncompressedSize` was _always_ returned with the older version of the FUSE server, but that’s not a concern going forward anyway.

influentcoder:
@ktock I've looked at several pull requests related to this change - what's the reasoning behind using the TOCDigest instead of the blob digest? Previously, with the blob digest, I could use images with the non-seekable, standard tar+gzip format, and use an additional layer stored backed by a NFS. But now that's not the case, I'm forced to store my images in either stargz or sztd:chunked format

mtrmac:
Layer IDs are used for deduplication. So, it _must not_ be possible for a layer with the same “ID” to be consumed in two different ways.

For zstd:chunked, the blob (identified with the full digest) can, in principle, be consumed _either_ as a non-chunked tarball, or as a TOC-based chunked object, and (in the case of ALS) there is ~nothing we can do to enforce the two are exactly the same filesystem (other than reading all files and validating against a DiffID: we _do that_ on various code paths, but it would entirely defeat the on-demand nature of ALS).

So, for ALS, the ID must match the data one _actually used_ to fetch the contents. For zstd:chunked on-demand access, that’s the TOC; the ALS backend is then trusted to not have a code path consuming the full tarball and ignoring the TOC.

I’m afraid that ALS change an incompatible, and barely-announced change. Due to the fairly weird way ALS use a FUSE filesystem basically as a RPC API, the features of that “API mechanism” are rather limited, and doing things properly over that API is… complex.

mtrmac:
… and, to finish the thought, that makes it very unattractive to me to consider just _reverting_ the behavior, in an invisible-incompatible way, again. If we were to change this, or to add any substantial features to ALS, I’d personally want to see a more extensible/maintainable/standard RPC mechanism — but, also, it’s not something I can spend significant time on.

chvish:
Would it not be possible to pass the ALS arg/config in a way that you can tell if you want the ALS to be used as either a non-chunked layer or not. Then you can use the blob digest or TOC checksum based on what the arg/config is, instead of the behaviour being inferred based on the image compression.

The current implementation means that the feature is ties to specific implementation of the ALS, one which assumes a layer store can only be used with seekable images, which should not be the case, atleast not intuitively.

influentcoder:
From a user perspective (or at least my perspective), it seems natural that if we specify a location as an additional layer store, that podman uses this location to source the layers, regardless of the compression format in the remote registry.

I imagine that the base case is to have simply the layer expanded in a filesystem. For example to store commonly used layers in a remote filesystem. The fact that we can use a fuse mount to perform lazy loading directly from a remote registry is just an extension of that primary use case in my view.

@mtrmac Would you be open if I submit a PR that brings backwards compatibility to the previous behavior? E.g. if a layer isn't compressed with a TOC, then fallback to lookup to the store based on the layer digest.

Or alternatively, as @chvish is proposing, a way to provide provide optionality on how we intend to use the layer store

mtrmac:
I am trying to hold to a policy that I will not be reviewing PRs that add extra options and code paths to drivers/overlay because I just can’t deal with the complexity.

I realize that’s not great after this ABI break, and anyway there are other maintainers of the project who could review PRs.

>  E.g. if a layer isn't compressed with a TOC, then fallback to lookup to the store based on the layer digest.

Given that the layer ID _must not_ cause collisions, I’m skeptical this can be done securely, at least given the existing API between c/storage and c/image. But I didn’t look into this in detail.

PR #2417, opened 2024-05-14
[Additional Layer Store] Enable Additional Layer Store to perform registry authentication

ktock:
Implements https://github.com/openshift/enhancements/pull/1600#issuecomment-2099607551

Currently, c/image doesn't share creds to Additional Layer Store so it requires additional configurations for registry authentication. And the current means of registry authentication for Additional Layer Store doesn't seem to meet requirements for the platform: https://github.com/openshift/enhancements/pull/1600#pullrequestreview-1990049051

This commit fixes that issue by allowing c/image directly sharing creds to Additional Layer Store. Additional Layer Store doesn't need to have its own logic to fetch registry creds but it can receive them from c/image.

Additional Layer Store needs to provide a helper binary that is executed from c/image. This helper binary is registered to c/image using registries.conf with the following field (`store-helper` can be any command name of the helper binary). It receives registry creds via stdin and Additional Layer Store can use that creds for registry authentication.

```
additional-layer-store-auth-helper = "store-helper"
```

An example draft implementation of the helper binary is `store-helper` of stargz-store: https://github.com/containerd/stargz-snapshotter/pull/1674
This binary is executed by c/image and receives the registry creds from stdin and shares them to stargz-store via the unix socket of stargz-store. Then stargz-store uses these creds for registry authentication.

c/image passes [`DockerAuthConfig`](https://github.com/containers/image/blob/984c624c4c9e07c79bcdbc49679ba2a781805303/types/types.go#L490-L500) structures with keying them with the image reference.

```json
{
  "image-reference": {
    "username": "username",
    "password": "password",
    "identitytoken": "identitytoken"
  }
}
```

### DIFF
--- a/docker/docker_image_src.go
+++ b/docker/docker_image_src.go
@@ -1,7 +1,9 @@
 package docker
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
@@ -11,6 +13,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"os/exec"
 	"strings"
 	"sync"
 
@@ -161,6 +164,34 @@ func newImageSourceAttempt(ctx context.Context, sys *types.SystemContext, logica
 	if err := s.ensureManifestIsLoaded(ctx); err != nil {
 		client.Close()
 		return nil, err
+	}
+
+	if h, err := sysregistriesv2.AdditionalLayerStoreAuthHelper(endpointSys); err == nil && h != "" {
+		acf := map[string]struct {
+			Username      string `json:"username,omitempty"`
+			Password      string `json:"password,omitempty"`
+			IdentityToken string `json:"identityToken,omitempty"`
+		}{
+			physicalRef.ref.String(): {
+				Username:      client.auth.Username,
+				Password:      client.auth.Password,
+				IdentityToken: client.auth.IdentityToken,
+			},
+		}
+		acfD, err := json.Marshal(acf)
+		if err != nil {
+			logrus.Warnf("failed to marshal auth config: %v", err)
+		} else {
+			cmd := exec.Command(h)
+			cmd.Stdin = bytes.NewReader(acfD)
+			if err := cmd.Run(); err != nil {
+				var stderr string
+				if ee, ok := err.(*exec.ExitError); ok {
+					stderr = string(ee.Stderr)
+				}
+				logrus.Warnf("Failed to call additional-layer-store-auth-helper (stderr:%s): %v", stderr, err)
+			}
+		}
 	}
 	return s, nil
 }
--- a/docs/containers-registries.conf.5.md
+++ b/docs/containers-registries.conf.5.md
@@ -19,6 +19,12 @@ Container engines will use the `$HOME/.config/containers/registries.conf` if it 
 `credential-helpers`
 : An array of default credential helpers used as external credential stores.  Note that "containers-auth.json" is a reserved value to use auth files as specified in containers-auth.json(5).  The credential helpers are set to `["containers-auth.json"]` if none are specified.
 
+`additional-layer-store-auth-helper`
+: A string containing the helper binary name. This enables passing registry credentials to an
+  Additional Layer Store every time an image is read using the `docker://`
+  transport so that it can access private registries. See the 'Enabling Additional Layer Store to access to private registries' section below for
+  more details.
+
 ### NAMESPACED `[[registry]]` SETTINGS
 
 The bulk of the configuration is represented as an array of `[[registry]]`
@@ -253,6 +259,30 @@ Given the above, a pull of `example.com/foo/image:latest` will try:
 in order, and use the first one that exists.
 
 Note that a mirror is associated only with the current `[[registry]]` TOML table. If using the example above, pulling the image `registry.com/image:latest` will hence only reach out to `mirror.registry.com`, and the mirrors associated with `example.com/foo` will not be considered.
+
+### Enabling Additional Layer Store to access to private registries
+
+The `additional-layer-store-auth-helper` option enables passing registry
+credentials to an Additional Layer Store so that it can access private registries.
+
+When accessing a private registry via an Additional Layer Store, a helper binary needs to be provided. This helper binary is
+registered via the `additional-layer-store-auth-helper` option. Every time an image
+is read using the `docker://` transport, the specified helper binary is executed
+and receives registry credentials from stdin in the following format.
+
+```json
+{
+  "$image_reference": {
+    "username": "$username",
+    "password": "$password",
+    "identityToken": "$identityToken"
+  }
+}
+```
+
+The format of `$image_reference` is `$repo{:$tag|@$digest}`.
+
+Additional Layer Stores can use this helper binary to access the private registry.
 
 ## VERSION 1 FORMAT - DEPRECATED
 VERSION 1 format is still supported but it does not support
--- a/pkg/sysregistriesv2/system_registries_v2.go
+++ b/pkg/sysregistriesv2/system_registries_v2.go
@@ -248,6 +248,11 @@ type V2RegistriesConf struct {
 	// potentially use all unqualified-search registries
 	ShortNameMode string `toml:"short-name-mode"`
 
+	// AdditionalLayerStoreAuthHelper is a helper binary that receives
+	// registry credentials pass them to Additional Layer Store for
+	// registry authentication. These credentials are only collected when pulling (not pushing).
+	AdditionalLayerStoreAuthHelper string `toml:"additional-layer-store-auth-helper"`
+
 	shortNameAliasConf
 
 	// If you add any field, make sure to update Nonempty() below.
@@ -825,6 +830,16 @@ func CredentialHelpers(sys *types.SystemContext) ([]string, error) {
 	return config.partialV2.CredentialHelpers, nil
 }
 
+// AdditionalLayerStoreAuthHelper returns the helper for passing registry
+// credentials to Additional Layer Store.
+func AdditionalLayerStoreAuthHelper(sys *types.SystemContext) (string, error) {
+	config, err := getConfig(sys)
+	if err != nil {
+		return "", err
+	}
+	return config.partialV2.AdditionalLayerStoreAuthHelper, nil
+}
+
 // refMatchingSubdomainPrefix returns the length of ref
 // iff ref, which is a registry, repository namespace, repository or image reference (as formatted by
 // reference.Domain(), reference.Named.Name() or reference.Reference.String()
@@ -1049,6 +1064,11 @@ func (c *parsedConfig) updateWithConfigurationFrom(updates *parsedConfig) {
 	// We don’t maintain c.partialV2.ShortNameMode.
 	if updates.shortNameMode != types.ShortNameModeInvalid {
 		c.shortNameMode = updates.shortNameMode
+	}
+
+	// == Merge AdditionalLayerStoreAuthHelper:
+	if updates.partialV2.AdditionalLayerStoreAuthHelper != "" {
+		c.partialV2.AdditionalLayerStoreAuthHelper = updates.partialV2.AdditionalLayerStoreAuthHelper
 	}
 
 	// == Merge aliasCache:
--- a/pkg/sysregistriesv2/system_registries_v2_test.go
+++ b/pkg/sysregistriesv2/system_registries_v2_test.go
@@ -57,6 +57,7 @@ var v2RegistriesConfEmptyTestData = []struct {
 	{nonempty: true, hasSetField: true, v: V2RegistriesConf{CredentialHelpers: []string{"a"}}},
 	{nonempty: true, hasSetField: true, v: V2RegistriesConf{ShortNameMode: "enforcing"}},
 	{nonempty: true, hasSetField: true, v: V2RegistriesConf{shortNameAliasConf: shortNameAliasConf{Aliases: map[string]string{"a": "example.com/b"}}}},
+	{nonempty: true, hasSetField: true, v: V2RegistriesConf{AdditionalLayerStoreAuthHelper: "example"}},
 }
 
 func TestV2RegistriesConfNonempty(t *testing.T) {


pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -248,6 +248,11 @@ type V2RegistriesConf struct {
 	// potentially use all unqualified-search registries
 	ShortNameMode string `toml:"short-name-mode"`
 
+	// AdditionalLayerStoreAuthHelper is a helper binary that receives
+	// registry credentials pass them to Additional Layer Store for
+	// registry authentication.
+	AdditionalLayerStoreAuthHelper string `toml:"additional-layer-store-auth-helper"`
```
mtrmac, line 254: At least `.Nonempty()` and `.updateWithConfigurationFrom()` also need updating.

mtrmac, line 254: The current implementation is restricted to reads (“pulls”, not pushes); that might well be fine, but it must be documented as such.

mtrmac, line 254: `docs/containers-registries.conf.5.md` also needs an update — probably after the all of the rest of the design is settled.

docker/docker_image_src.go
```diff
@@ -161,6 +164,26 @@ func newImageSourceAttempt(ctx context.Context, sys *types.SystemContext, logica
 		client.Close()
 		return nil, err
 	}
+
+	if h, err := sysregistriesv2.AdditionalLayerStoreAuthHelper(endpointSys); err == nil && h != "" {
+		acf := map[string]struct {
```
mtrmac, line 169: If this becomes a long-term ABI, it needs to be publicly documented as such. And maybe defined as a public Go type, to allow importing into consumers — although that might be necessary short-term.

docker/docker_image_src.go
```diff
@@ -161,6 +164,26 @@ func newImageSourceAttempt(ctx context.Context, sys *types.SystemContext, logica
...
+			Username      string `json:"username,omitempty"`
+			Password      string `json:"password,omitempty"`
+			IdentityToken string `json:"identitytoken,omitempty"`
```
mtrmac, line 172:
```suggestion
			IdentityToken string `json:"identityToken,omitempty"`
```
wouldn’t hurt

docker/docker_image_src.go
```diff
@@ -161,6 +164,26 @@ func newImageSourceAttempt(ctx context.Context, sys *types.SystemContext, logica
...
+		}{
+			physicalRef.ref.String(): {
```
mtrmac, line 174: Should this be `.ref.Name()`? Credentials are, for the most part, repo-scoped.

Admittedly the FUSE interface uses `ref.String()`.

docker/docker_image_src.go
```diff
@@ -161,6 +164,26 @@ func newImageSourceAttempt(ctx context.Context, sys *types.SystemContext, logica
...
+				Username:      client.auth.Username,
+				Password:      client.auth.Password,
+				IdentityToken: client.auth.IdentityToken,
+			},
+		}
+		acfD, _ := json.Marshal(acf)
+		cmd := exec.Command(h)
+		cmd.Stdin = bytes.NewReader(acfD)
+		if err := cmd.Run(); err != nil {
+			logrus.Warnf("Failed to call store helper: %v", err)
```
mtrmac, line 184: Capture and log standard error?

mtrmac (changes requested):
Thanks!

This is plausible, _assuming_ the system is single-purpose; otherwise all other registry operations, including manual `skopeo` commands, would also be sending credentials.

A possible alternative would be to capture the credentials in `ImageSource`, and provide  a some kind of interface alongside `TryReusingBlobOptions.SrcRef`. That would only send “relevant” credentials to the helper, at the cost of some extra plumbing.

pkg/sysregistriesv2/system_registries_v2.go
ktock: Fixed updateWithConfigurationFrom. I guess nothing to be updated in `Nonempty` because it is  a string field?

docker/docker_image_src.go
ktock: This follows the coding style of other part of the codebase (pkg/docker/config/config.go uses `identitytoken`) : https://github.com/containers/image/blob/984c624c4c9e07c79bcdbc49679ba2a781805303/pkg/docker/config/config.go#L30

docker/docker_image_src.go
```diff
@@ -161,6 +164,26 @@ func newImageSourceAttempt(ctx context.Context, sys *types.SystemContext, logica
...
+			Password      string `json:"password,omitempty"`
+			IdentityToken string `json:"identitytoken,omitempty"`
+		}{
+			physicalRef.ref.String(): {
+				Username:      client.auth.Username,
+				Password:      client.auth.Password,
+				IdentityToken: client.auth.IdentityToken,
+			},
+		}
+		acfD, _ := json.Marshal(acf)
```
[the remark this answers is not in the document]
ktock: Fixed

docker/docker_image_src.go
ktock: Fixed.

pkg/sysregistriesv2/system_registries_v2.go
ktock: Added documentation.

TomSweeneyRedHat:
@ktock Skopeo test is failing and you need a rebase.

ktock:
Rebased. And skopeo error seems to be resolved.

rhatdan:
@mtrmac PTAL

pkg/sysregistriesv2/system_registries_v2.go
```diff
@@ -248,6 +248,11 @@ type V2RegistriesConf struct {
 	// potentially use all unqualified-search registries
 	ShortNameMode string `toml:"short-name-mode"`
 
+	// AdditionalLayerStoreAuthHelper is a helper binary that receives
+	// registry credentials pass them to Additional Layer Store for
+	// registry authentication. These credentials are only used for pulling (not pushing).
```
mtrmac, line 253:
```suggestion
	// registry authentication. These credentials are only collected when pulling (not pushing).
```
I think — c/image doesn’t get to say what they are _used for_, the restriction is about the situations in which the helper is called.

docker/docker_image_src.go
mtrmac: The data structure in `pkg/docker/config` matches `github.com/docker/cli/config/types.AuthConfig`, to be interoperable with `docker login` and the like.

This is a distinct data structure (e.g. it doesn’t hide the user name / password in a base64-encoded string — and that’s good!), so we have the liberty, and the ~unique opportunity, to make it more convenient.

docker/docker_image_src.go
```diff
@@ -162,6 +165,34 @@ func newImageSourceAttempt(ctx context.Context, sys *types.SystemContext, logica
...
+			logrus.Warnf("failed to marshal auth config: %v", err)
+		} else {
+			cmd := exec.Command(h)
+			cmd.Stdin = bytes.NewReader(acfD)
+			if err := cmd.Run(); err != nil {
+				var stderr string
+				if ee, ok := err.(*exec.ExitError); ok {
+					stderr = string(ee.Stderr)
+				}
+				logrus.Warnf("Failed to call store helper(stderr:%s): %v", stderr, err)
```
mtrmac, line 192:
```suggestion
				logrus.Warnf("Failed to call additional-layer-store-auth-helper (stderr:%s): %v", stderr, err)
```
to help unaware users find the relevant option.

mtrmac (commented):
Thanks! This looks good overall.

I am still conceptually a bit unsure whether capturing all credentials ever read is a good idea — or whether we should be restricting it to those relevant to `LookupAdditionalLayer`, at the cost of having to make the special case visible to the `c/image/copy` top level. I don’t have a good idea of where this is likely to be deployed.

Other than that, the option and the data format just need a man page update.

---

@giuseppe PTAL as well.

pkg/sysregistriesv2/system_registries_v2.go
ktock: Added a test.

docker/docker_image_src.go
ktock: Fixed to make it `identityToken`.

pkg/sysregistriesv2/system_registries_v2.go
ktock: Fixed.

docker/docker_image_src.go
ktock: Fixed.

mtrmac (commented):
Thanks!

Implementation LGTM, this is just missing man page updates. But I’d like to hear what @giuseppe thinks.

ktock:
Added docs to `docs/containers-registries.conf.5.md`

docs/containers-registries.conf.5.md
```diff
@@ -254,6 +259,28 @@ in order, and use the first one that exists.
...
 Note that a mirror is associated only with the current `[[registry]]` TOML table. If using the example above, pulling the image `registry.com/image:latest` will hence only reach out to `mirror.registry.com`, and the mirrors associated with `example.com/foo` will not be considered.
 
+### Enabling Additional Layer Store to access to private registries
+
+The `additional-layer-store-auth-helper` option enables to pass registry
+credentials to Additional Layer Store so that it can access private registries.
+
+Additional Layer Store needs to provide a helper binary. This helper binary is
+registered to `additional-layer-store-auth-helper` option. Every time pulling of
+an image happens, the specified helper binary is executed and receives registry
```
mtrmac, line 269: “Every time an image is read using the `docker://` transport”, or something like that:
- c/image doesn’t know about pulling
- `skopeo inspect docker://…` is not a pull, but it _would_ send credentials

Also in the `additional-layer-store-auth-helper` option documentation above.

docs/containers-registries.conf.5.md
```diff
@@ -254,6 +259,28 @@ in order, and use the first one that exists.
...
+credentials from stdin in the following format.
+
+```json
+{
+  "image-reference": {
+    "username": "username",
+    "password": "password",
+    "identityToken": "identityToken"
+  }
```
mtrmac, line 278: It’s completely unclear from this syntax which of the strings are constants and which of them are variables — especially when “image-reference” is not a constant. I don’t know what’s a good way to do that here, either avoid ``` and format explicitly using the man-page `constants` and _variables_ syntax, or maybe use `"$username"` to indicate that it is a variable?

What is the format of “image-reference”? “$repo{:$tag|@$digest}" would be a bit more explicit.

Is it, long-term, guaranteed that there will be only one “image-reference”? Or can the map contain multiple entries and should the helpers be ready to accept that? I guess the latter, otherwise the `map[]` format would not make sense.

---

Thinking about the evolution, if any, more:
- Do we need a version number? Probably not, we can define an `additional-layer-store-auth-helper2` field.
- Making the top-level object’s fields directly interpreted as image references means we can’t add any other field to the object in the future. Do we care? Not _urgently_, we can always use …`helper2` with a new format. I also can’t think of any field we might want to add, but that doesn’t mean much.

ktock: >  or maybe use `"$username"` to indicate that it is a variable?

> What is the format of “image-reference”? “$repo{:$tag|@$digest}" would be a bit more explicit.

Thanks for the suggestions. Fixed.

> Is it, long-term, guaranteed that there will be only one “image-reference”? Or can the map contain multiple entries and should the helpers be ready to accept that? I guess the latter, otherwise the map[] format would not make sense.

There is no guarantee of "image-reference" being only one. The reason for the image reference not being a field is that I initially tried to make the struct the same format as DockerAuthConfig which is already used in the codebase. But, if needed, it's also ok for me to extend the struct to make `image-reference` one of the fields (not the key of the struct).

> version

>  add any other field to the object in the future

Currently I can’t think of any field we might want to add. Maybe we can `additional-layer-store-auth-helper2` when we need to change the format.

docs/containers-registries.conf.5.md
ktock: Thanks for the comments. Fixed.

ktock:
Could we move this forward?

TomSweeneyRedHat:
@giuseppe thoughts?

@ktock, looks like you need a rebase.

docs/containers-registries.conf.5.md
```diff
@@ -19,6 +19,12 @@ Container engines will use the `$HOME/.config/containers/registries.conf` if it
 `credential-helpers`
 : An array of default credential helpers used as external credential stores.  Note that "containers-auth.json" is a reserved value to use auth files as specified in containers-auth.json(5).  The credential helpers are set to `["containers-auth.json"]` if none are specified.
 
+`additional-layer-store-auth-helper`
+: A string of helper name. This enables to pass registry credentials to
```
TomSweeneyRedHat, line 23:
```suggestion
: A string containing the helper binary name. This enables passing registry credentials to an 
```

docs/containers-registries.conf.5.md
```diff
@@ -19,6 +19,12 @@ Container engines will use the `$HOME/.config/containers/registries.conf` if it
...
+  Additional Layer Store every time an image is read using the `docker://`
+  transport so that it can access private registries. See the later section for
```
TomSweeneyRedHat, line 25:
```suggestion
  transport so that it can access private registries. See the 'Enabling Additional Layer Store to access to private registries' section below for
```

docs/containers-registries.conf.5.md
```diff
@@ -254,6 +260,30 @@ in order, and use the first one that exists.
 
 Note that a mirror is associated only with the current `[[registry]]` TOML table. If using the example above, pulling the image `registry.com/image:latest` will hence only reach out to `mirror.registry.com`, and the mirrors associated with `example.com/foo` will not be considered.
 
+### Enabling Additional Layer Store to access to private registries
+
+The `additional-layer-store-auth-helper` option enables to pass registry
```
TomSweeneyRedHat, line 265:
```suggestion
The `additional-layer-store-auth-helper` option enables passing registry
```

docs/containers-registries.conf.5.md
```diff
@@ -254,6 +260,30 @@ in order, and use the first one that exists.
...
+credentials to Additional Layer Store so that it can access private registries.
```
TomSweeneyRedHat, line 266:
```suggestion
credentials to an Additional Layer Store so that it can access private registries.
```

docs/containers-registries.conf.5.md
```diff
@@ -254,6 +260,30 @@ in order, and use the first one that exists.
...
+
+Additional Layer Store needs to provide a helper binary. This helper binary is
```
TomSweeneyRedHat, line 268: I think this is what you're getting at

```suggestion
When accessing a private registry via an Additional Layer Store, a helper binary needs to be provided. This helper binary is
```

docs/containers-registries.conf.5.md
```diff
@@ -254,6 +260,30 @@ in order, and use the first one that exists.
...
+registered to `additional-layer-store-auth-helper` option. Every time an image
```
TomSweeneyRedHat, line 269:
```suggestion
registered via the `additional-layer-store-auth-helper` option. Every time an image
```

docs/containers-registries.conf.5.md
```diff
@@ -254,6 +260,30 @@ in order, and use the first one that exists.
...
+    "username": "$username",
+    "password": "$password",
+    "identityToken": "$identityToken"
+  }
+}
+```
+
+The format of `$image_reference` is `$repo{:$tag|@$digest}`.
+
+Additional Layer Store can use this for accessing to the registry.
```
TomSweeneyRedHat, line 285:
```suggestion
Additional Layer Stores can use this helper binary to access the private registry.
```

ktock:
@TomSweeneyRedHat Thanks for the review. FIxed and rebased the patch.

ktock:
@giuseppe PTAL

giuseppe (approved):
LGTM

mtrmac (approved):

mtrmac:
Thanks again!

PR #2415, opened 2024-05-14
[release-5.29] Backport #2403 aka CVE-2024-3727

dcermak:

PR #2418, opened 2024-05-15
[release-5.29] Fix CVE-2024-3727

mtrmac:

---

This is a backport of #2404 . Thanks to @dcermak for independently validating this work via #2415 .

### DIFF
--- a/copy/progress_bars.go
+++ b/copy/progress_bars.go
@@ -48,10 +48,13 @@ type progressBar struct {
 // As a convention, most users of progress bars should call mark100PercentComplete on full success;
 // by convention, we don't leave progress bars in partial state when fully done
 // (even if we copied much less data than anticipated).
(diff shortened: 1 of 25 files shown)


TomSweeneyRedHat:
LGTM, nice work @mtrmac !

TomSweeneyRedHat:
@kwilczynski, I'll let @mtrmac correct me if I'm wrong, but I think you will also want to grab the follow-on fix for this fix for CRI-O.  That fix is in this PR on main: https://github.com/containers/image/pull/2440 and solves the issue as noted here: https://github.com/containers/skopeo/issues/2346

If it helps, this has been fixed in c/image v5.31.1, v5.30.2, and v5.29.4.  If you need older versions fixed, please let us know.

mtrmac:
From a _very_ quick check, CRI-O seems to never call the affected `GetRepositoryTags` function, but I could well have missed something.

PR #2424, opened 2024-05-17
Support proxy environment for docker client

PMExtra:
Proxy environment works fine with docker-cli:

```bash
> export http_proxy=http://my_proxy:8080
> export DOCKER_HOST=tcp://my_docker_host_through_proxy
> docker info
# Works fine
> docker image inspect foo:bar
# Works fine
```

But does not work with skopeo:

```bash
> export http_proxy=http://my_proxy:8080
> skopeo inspect --daemon-host http://my_docker_host_through_proxy docker-daemon:foo:bar
FATA[0000] Error parsing image name "docker-daemon:foo:bar": initializing docker engine client: unable to parse docker host `my_docker_host_through_proxy`
```

It's just an example, I need to copy image to the docker-daemon host in actual.

After this pull request was patched, skopeo also worked fine with the proxy environment.

> Note that: I used a plain HTTP connection with docker-daemon, so it should be `http://` in skepeo but `tcp://` in docker-cli.

### DIFF
--- a/docker/daemon/client.go
+++ b/docker/daemon/client.go
@@ -80,6 +80,7 @@ func tlsConfig(sys *types.SystemContext) (*http.Client, error) {
 
 	return &http.Client{
 		Transport: &http.Transport{
+			Proxy:           http.ProxyFromEnvironment,
 			TLSClientConfig: tlsc,
 		},
 		CheckRedirect: dockerclient.CheckRedirect,
@@ -89,6 +90,7 @@ func tlsConfig(sys *types.SystemContext) (*http.Client, error) {
 func httpConfig() *http.Client {
 	return &http.Client{
 		Transport: &http.Transport{
+			Proxy:           http.ProxyFromEnvironment,
 			TLSClientConfig: nil,
 		},
 		CheckRedirect: dockerclient.CheckRedirect,


mtrmac (changes requested):
Thanks!

> After this pull request was patched, skopeo also worked fine with the proxy environment.

This does not even compile, so I can’t see how this could have possibly worked.

---

Also, per https://github.com/containers/image/blob/main/CONTRIBUTING.md#sign-your-prs :
> Use your real name (sorry, no pseudonyms or anonymous contributions.)

PMExtra:
@mtrmac I've fixed it. Please let me know if there is anything else I need to do.

rhatdan:
LGTM

mtrmac (commented):
If this makes sense doing, then the `http:` and `https:` protocols should support proxies equally.

(Alternatively, AFAICS we don’t actually need to configure this manually: comparing https://github.com/docker/cli/blob/57a1180c52ad7633dcf2fe1515ec2601ffb01944/cli/context/docker/load.go#L85 and looking at the implementation, if `WithHttpClient` is included _before_ `WithHost`, that triggers `sockets.ConfigureTransport` and that sets this, along, with a few other options. But that’s undocumented and very nonobvious; it’s really only a confirmation that `ProxyFromEnvironment` can be appropriate here.)

rhatdan:
@PMExtra still working on this?

PMExtra:
I don't know what I should do. I think it's ready for merging.

rhatdan:
@mtrmac what would you like to see done?

mtrmac:
> If this makes sense doing, then the `http:` and `https:` protocols should support proxies equally.

This only adds them to the former; add support for the latter.

PMExtra:
@mtrmac I'm not sure what you mean. `http.ProxyFromEnvironment` can support several variables such as `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY`.

If you set `HTTPS_PROXY`, that will work with `https:`.

PMExtra:
@mtrmac see https://pkg.go.dev/net/http#ProxyFromEnvironment

mtrmac:
The `httpConfig` function is not used for `https`, so whether `http.ProxyFromEnvironment` _would have_ worked for `https` is irrelevant when this PR is not invoking it in that case.

PMExtra:
> The httpConfig function is not used for https, so whether http.ProxyFromEnvironment would have worked for https is irrelevant when this PR is not invoking it in that case.

@mtrmac I don't think so. Let's see a example:

I have a main.go:
```go
package main

import (
        "fmt"
        "io"
        "net/http"
        "os"
)

func main() {
        client := &http.Client{
                Transport: &http.Transport{
                        Proxy: http.ProxyFromEnvironment,
                },
        }

        resp, err := client.Get(os.Args[1])
        if err != nil {
                fmt.Printf("Request failed: %v\n", err)
                return
        }
        defer resp.Body.Close()

        body, err := io.ReadAll(resp.Body)
        if err != nil {
                fmt.Println("Failed to read response body:", err)
                return
        }

        fmt.Println("Response status:", resp.Status)
        fmt.Println("Response body:", string(body))
}
```

Then I run it:

```plain
$ go run main.go https://api.ipify.org
Response status: 200 OK
Response body: 27.38.xxx.xxx
$ export https_proxy=http://proxy.host:8080  # set https_proxy
$ go run main.go https://api.ipify.org    # https: with https_proxy, through proxy
Response status: 200 OK
Response body: 52.192.xxx.xxx
$ go run main.go http://api.ipify.org    # http: with https_proxy, no effect
Response status: 200 OK
Response body: 27.38.xxx.xxx
$ export http_proxy=http://proxy.host:8080  # set http_proxy
$ go run main.go http://api.ipify.org    # http: with http_proxy, through proxy
Response status: 200 OK
Response body: 52.192.xxx.xxx
$ export no_proxy=api.ipify.org    # set no_proxy
$ go run main.go http://api.ipify.org    # no_proxy take precedence than http_proxy or https_proxy
Response status: 200 OK
Response body: 27.38.xxx.xxx
$ go run main.go https://api.ipify.org
Response status: 200 OK
Response body: 27.38.xxx.xxx
```

According to these, `http.ProxyFromEnvironment` works with both `http:` and `https:`.

mtrmac:
That example does not reference the `httpConfig` function (and, more to the point, its caller) at all, and is irrelevant to my objection.

PMExtra:
@mtrmac Now I finally understand what you meant, and I apologize for the previous misunderstanding.

mtrmac (commented):
Thanks!

Code LGTM, please squash into one commit and rebase.

PR #2435, opened 2024-05-28
blobinfocache: add function to delete the cache directory

giuseppe:
Needed for: https://github.com/containers/podman/issues/22825

### DIFF
--- a/pkg/blobinfocache/default.go
+++ b/pkg/blobinfocache/default.go
@@ -74,3 +74,15 @@ func DefaultCache(sys *types.SystemContext) types.BlobInfoCache {
 	logrus.Debugf("Using SQLite blob info cache at %s", path)
 	return cache
 }
+
+// CleanupDefaultCache removes the blob info cache directory.
+// It deletes the cache directory but it does not affect any file or memory buffer currently
+// in use.
+func CleanupDefaultCache(sys *types.SystemContext) error {
+	dir, err := blobInfoCacheDir(sys, rootless.GetRootlessEUID())
+	if err != nil {
+		// Mirror the DefaultCache behavior that does not fail in this case
+		return nil
+	}
+	return os.RemoveAll(dir)
+}


pkg/blobinfocache/default.go
```diff
@@ -74,3 +74,12 @@ func DefaultCache(sys *types.SystemContext) types.BlobInfoCache {
 	logrus.Debugf("Using SQLite blob info cache at %s", path)
 	return cache
 }
+
+// Cleanup removes the blob info cache directory.
```
mtrmac, line 78: This might not do anything for caches which already exist in this process, either because they keep a file descriptor open, or because it’s `pkg/blobinfocache/memory`.

I suppose that’s fine (doing anything else would be too hard and it’s not necessary for the use cases), but it should be documented.

pkg/blobinfocache/default.go
```diff
@@ -74,3 +74,12 @@ func DefaultCache(sys *types.SystemContext) types.BlobInfoCache {
...
+func Cleanup(sys *types.SystemContext) error {
+	dir, err := blobInfoCacheDir(sys, rootless.GetRootlessEUID())
+	if err != nil {
+		return err
```
mtrmac, line 82: `DefaultCache` does not fail in this situation, so I don’t think it makes sense to fail here.

pkg/blobinfocache/default.go
```diff
@@ -74,3 +74,12 @@ func DefaultCache(sys *types.SystemContext) types.BlobInfoCache {
 	logrus.Debugf("Using SQLite blob info cache at %s", path)
 	return cache
 }
+
+// Cleanup removes the blob info cache directory.
+func Cleanup(sys *types.SystemContext) error {
```
mtrmac, line 79: `DeleteDefaultCache`, to mirror `DefaultCache`?

mtrmac (commented):
Thanks!

Issue #2319, opened 2024-02-28
support multiple sigstore keys


msmeissn:
### Feature request description

At SUSE we currently want to transition the sigstore signing key from a RSA 2048 to 4096 bit key.
During the transition time, there might exist containers signed with either of the keys.
Currently, podman seems to be able to only configure 1 key for sigstore signing, not multiple ones.

### Suggest potential solution

Support multiple sigstore keys, or perhaps reading from an index directory.

dcermak:
We've had a thought with @danishprakash about how this could be implemented. A very simple way would be to change the loop in https://github.com/containers/image/blob/6caf212454493ef1979393c398ac705aebb393a6/signature/policy_eval.go#L281

to accept a signature if at least one policy matches instead of requiring all policies to match.

That would allow you to write the following `policy.json`:
```json
{
    "default": [
        {
            "type": "insecureAcceptAnything"
        }
    ],
    "transports": {
        "docker-daemon": {
            "": [{"type":"insecureAcceptAnything"}]
        },
        "docker": {
            "registry.suse.com/bci": [{
                "type": "sigstoreSigned",
                "keyPath": "/usr/share/pki/containers/suse-container-key.pem",
                "signedIdentity": {
                    "type": "matchRepository"
                }
            },
            {
                "type": "sigstoreSigned",
                "keyPath": "/usr/share/pki/containers/old-suse-container-key.pem",
                "signedIdentity": {
                    "type": "matchRepository"
                }
            }]
        }
    }
}
```

And the container image would validate, if it is signed by **either** `/usr/share/pki/containers/suse-container-key.pem` or `/usr/share/pki/containers/old-suse-container-key.pem`.

The big advantage is, that you get support for multiple GPG keys as well "for free". Or you can now have a part of your images signed by sigtore and a part by GPG.

However, this is a breaking change in the behavior and has other ramifications. E.g. you can no longer enforce sigstore **and** GPG signatures to match. I am unsure whether that is actually used in production anywhere, but the behavior would change.

We can of course convert `keyPath` to `keyPaths` (or add support for both fields), but this would be a bigger code change and it would have the disadvantage, that it would have to be implemented for every policy type (sigstore, GPG, etc.).


Any thoughts/opinions?

mtrmac:
The semantics of multiple requirements is already defined to be an AND, and users might rely on that for security.

It‘s just _not an option_ to silently change that behavior.

[...]

PR #2456, opened 2024-06-20
Add field KeyPaths to prSigstoreSigned

dcermak:
The new field `KeyPaths` is taken directly from `/etc/containers/policy.json` and allows users to provide multiple signature keys to be used to verify images. Only one of the keys has to verify, thereby this mechanism allows us to have support seamless key rotation on a registry.

This fixes https://github.com/containers/image/issues/2319

### DIFF
--- a/signature/internal/sigstore_payload.go
+++ b/signature/internal/sigstore_payload.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/containers/image/v5/version"
@@ -171,36 +172,72 @@ type SigstorePayloadAcceptanceRules struct {
 	ValidateSignedDockerManifestDigest func(digest.Digest) error
 }
 
-// VerifySigstorePayload verifies unverifiedBase64Signature of unverifiedPayload was correctly created by publicKey, and that its principal components
+// VerifySigstorePayload verifies unverifiedBase64Signature of unverifiedPayload was correctly created by any of the public keys in publicKeys, and that its principal components
 // match expected values, both as specified by rules, and returns it.
 // We return an *UntrustedSigstorePayload, although nothing actually uses it,
 // just to double-check against stupid typos.
-func VerifySigstorePayload(publicKey crypto.PublicKey, unverifiedPayload []byte, unverifiedBase64Signature string, rules SigstorePayloadAcceptanceRules) (*UntrustedSigstorePayload, error) {
-	verifier, err := sigstoreSignature.LoadVerifier(publicKey, sigstoreHarcodedHashAlgorithm)
-	if err != nil {
-		return nil, fmt.Errorf("creating verifier: %w", err)
+func VerifySigstorePayload(publicKeys []crypto.PublicKey, unverifiedPayload []byte, unverifiedBase64Signature string, rules SigstorePayloadAcceptanceRules) (*UntrustedSigstorePayload, error) {
+	if len(publicKeys) == 0 {
+		return nil, fmt.Errorf("Need at least one public key to verify the sigstore payload, but got 0")
 	}
 
 	unverifiedSignature, err := base64.StdEncoding.DecodeString(unverifiedBase64Signature)
 	if err != nil {
 		return nil, NewInvalidSignatureError(fmt.Sprintf("base64 decoding: %v", err))
 	}
-	// github.com/sigstore/cosign/pkg/cosign.verifyOCISignature uses signatureoptions.WithContext(),
-	// which seems to be not used by anything. So we don’t bother.
-	if err := verifier.VerifySignature(bytes.NewReader(unverifiedSignature), bytes.NewReader(unverifiedPayload)); err != nil {
-		return nil, NewInvalidSignatureError(fmt.Sprintf("cryptographic signature verification failed: %v", err))
+
+	var unmatchedPayload *UntrustedSigstorePayload = nil
+
+	signatureErrMsgs := make([]string, 0, len(publicKeys))
+
+	for _, pk := range publicKeys {
+		// loading a verifier indicates that something is really, really
+		// messed up with the public key, so we should probably error
+		// out
+		verifier, err := sigstoreSignature.LoadVerifier(pk, sigstoreHarcodedHashAlgorithm)
+		if err != nil {
+			return nil, err
+		}
+
+		// github.com/sigstore/cosign/pkg/cosign.verifyOCISignature uses signatureoptions.WithContext(),
+		// which seems to be not used by anything. So we don’t bother.
+		if err := verifier.VerifySignature(bytes.NewReader(unverifiedSignature), bytes.NewReader(unverifiedPayload)); err != nil {
+			signatureErrMsgs = append(signatureErrMsgs, fmt.Sprintf("%v", err))
+			continue
+		}
+
+		// unmarshal after the signature has been verified, this is more
+		// resilient against bugs in the JSON parser
+		// Only perform the unmarshal in the first loop iteration
+		if unmatchedPayload == nil {
+			if err := json.Unmarshal(unverifiedPayload, &unmatchedPayload); err != nil {
+				return nil, err
+			}
+		}
+
+		// At this point we know that the signature has verified and now
+		// we can check the sigstore rules.
+		// The rules are independent of the individual public keys, so
+		// if any of them fail, we just return their error directly as
+		// the user won't really care that one of the public keys didn't
+		// verify this signature.
+		if err := rules.ValidateSignedDockerManifestDigest(unmatchedPayload.untrustedDockerManifestDigest); err != nil {
+			return nil, err
+		}
+
+		if err := rules.ValidateSignedDockerReference(unmatchedPayload.untrustedDockerReference); err != nil {
+			return nil, err
+		}
+
+		// SigstorePayloadAcceptanceRules have accepted this value.
+		return unmatchedPayload, nil
 	}
 
-	var unmatchedPayload UntrustedSigstorePayload
-	if err := json.Unmarshal(unverifiedPayload, &unmatchedPayload); err != nil {
-		return nil, NewInvalidSignatureError(err.Error())
+	// at this point we must have failed to verify the signature with every key
+	// => there must be at least one error
+	if len(signatureErrMsgs) == 0 {
+		return nil, fmt.Errorf("Internal error: signature verification failed but no errors have been recorded")
 	}
-	if err := rules.ValidateSignedDockerManifestDigest(unmatchedPayload.untrustedDockerManifestDigest); err != nil {
-		return nil, err
-	}
-	if err := rules.ValidateSignedDockerReference(unmatchedPayload.untrustedDockerReference); err != nil {
-		return nil, err
-	}
-	// SigstorePayloadAcceptanceRules have accepted this value.
-	return &unmatchedPayload, nil
+
+	return nil, NewInvalidSignatureError("cryptographic signature verification failed: " + strings.Join(signatureErrMsgs, ", "))
 }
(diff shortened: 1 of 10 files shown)


signature/policy_eval_sigstore.go
```diff
@@ -134,34 +181,37 @@ func (pr *prSigstoreSigned) isSignatureAccepted(ctx context.Context, image priva
...
+				if err != nil {
+					// Coverage: The key was loaded from a PEM format, so it’s unclear how this could fail.
+					// (PEM is not essential, MarshalPublicKeyToPEM can only fail if marshaling to ASN1.DER fails.)
+					return sarRejected, fmt.Errorf("re-marshaling public key to PEM: %w", err)
+
+				}
+				// We don’t care about the Rekor timestamp, just about log presence.
+				if _, err := internal.VerifyRekorSET(trustRoot.rekorPublicKey, []byte(untrustedSET), recreatedPublicKeyPEM, untrustedBase64Signature, untrustedPayload); err != nil {
+					return sarRejected, err
+				}
```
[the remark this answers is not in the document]
danishprakash: Is there a 1:1 relation between a public key and a rekor SET? If that's indeed the case, can we then assume that if the user specifies 2 entries for `keyPaths`, then there _has_ to be 2 entries for `rekor*`? For instance in the following snippet..
```
...
"keyPaths": ["/keyPath1", "/keyPath2"],
"rekorPublicKeyPaths": ["/rekorPath1", "/rekorPath2"],
...
```
...`keyPath1` is associated with `rekorPath1` and so on?

I think this sounds right but if we do go ahead with this, shall we also enforce this? In the above example, if there's only a single entry for `rekorPublicKeyPaths` and multiple for `keyPaths`, do we error out?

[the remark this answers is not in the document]
mtrmac: There is a 1:1 relationship between the raw signature and the SET elements of “the full sigstore signature object” (in that _if_ we require a SET, it must contain a record of signing with the right key), but there isn’t necessarily a 1:1 relationship between public keys and Rekor servers.

A vendor could easily have a single Rekor server but many different public keys (e.g. one per project / per purpose, like a production key / beta key / production-by-an-acquired-company key). Or, hypothetically, a single public key but several Rekor servers, why not.

signature/policy_eval_sigstore.go
```diff
@@ -184,43 +221,63 @@ func (pr *prSigstoreSigned) isSignatureAccepted(ctx context.Context, image priva
...
-		return sarRejected, err
-	}
-	if signature == nil { // A paranoid sanity check that VerifySigstorePayload has returned consistent values
-		return sarRejected, errors.New("internal error: VerifySigstorePayload succeeded but returned no data") // Coverage: This should never happen.
+
+	errs := make([]error, len(publicKeys))
+	hasPolicyRequirementError := false
+
+	for _, publicKey := range publicKeys {
+		signature, err := internal.VerifySigstorePayload(publicKey, untrustedPayload, untrustedBase64Signature, internal.SigstorePayloadAcceptanceRules{
```
mtrmac, line 236: I think it makes more sense to provide all `publicKeys` to a single call to `VerifySigstorePayload`.

The `untrustedPayload` and `untrustedBase64Signature` are the same in all calls, so basically there are three cases:
- There’s something wrong about the basic format (e.g. incorrect base64): we should only report that once, not N times for N keys
- The basic format is fine, but the signature is not valid under any of the keys: fine, probably N failures? Although the error messages returned seem not to be all that informative, we might just want to consolidate
- The cryptographic signature does validate under at least one key, but the payload is rejected by `SigstorePayloadAcceptanceRules`: we should only report the payload error, and not mention anything about failing to validate with the other public keys.

All of that seems easier to do if the loop is inside `VerifySigstorePayload` rather than outside.

dcermak: This is a very good idea, it simplified the code quite a lot

dcermak:
Miloslav Trmač ***@***.***> writes:

> Just a quick ACK — I definitely want this feature to happen short-term, and I appreciate your work on it. Right now I need to finish some other work (related to the zstd:chunked), so I’m afraid I’ll only be able to review this in detail in a week or two. I just wanted to be transparent about it, while reaffirming that this _is_ important both to me personally and to my employer, and that the effort is not wasted.

No worries, myself and Danish haven't been that responsive either. We'll
continue to work on improving the test coverage in the meantime.

signature/policy_eval_sigstore_test.go
```diff
@@ -89,8 +89,13 @@ func TestPRSigstoreSignedFulcioPrepareTrustRoot(t *testing.T) {
 
 func TestPRSigstoreSignedPrepareTrustRoot(t *testing.T) {
 	const testKeyPath = "fixtures/cosign.pub"
+	const testKeyPath2 = "fixtures/cosign.pub"
```
mtrmac, line 92: I think we want two truly different paths with different contents to test correctness of the code.

signature/policy_eval_sigstore_test.go
```diff
@@ -131,7 +145,7 @@ func TestPRSigstoreSignedPrepareTrustRoot(t *testing.T) {
 	require.NoError(t, err)
 	res, err := pr.prepareTrustRoot()
 	require.NoError(t, err)
-	assert.Nil(t, res.publicKey)
+	assert.Len(t, res.publicKeys, 0)
```
mtrmac, line 148: `nil` is conceptually a cleaner “not used” value.

signature/policy_eval_sigstore_test.go
```diff
@@ -320,13 +348,18 @@ func TestPRrSigstoreSignedIsSignatureAccepted(t *testing.T) {
...
 	pr, err = newPRSigstoreSigned(
-		PRSigstoreSignedWithKeyPath("fixtures/cosign2.pub"),
+		PRSigstoreSignedWithKeyPaths([]string{"fixtures/cosign.pub", "fixtures/cosign2.pub"}),
 		PRSigstoreSignedWithRekorPublicKeyPath("fixtures/rekor.pub"),
 		PRSigstoreSignedWithSignedIdentity(prm),
 	)
+
 	require.NoError(t, err)
 	sar, err = pr.isSignatureAccepted(context.Background(), testKeyRekorImage, testKeyRekorImageSig)
+	require.NoError(t, err)
```
mtrmac, line 358: Isn’t this redundant with the `assertAccepted` below? Or, alternatively, if `assertAcepted` now depends on `err == nil` to do other checks, this `require` should happen inside `assertAccepted`.

signature/policy_eval_sigstore_test.go
```diff
@@ -320,13 +348,18 @@ func TestPRrSigstoreSignedIsSignatureAccepted(t *testing.T) {
...
 	assertAccepted(sar, err)
+	tr, err := pr.prepareTrustRoot()
```
mtrmac, line 360: I don’t see that this is obviously valid: a new `prepareTrustRoot` is going to load the keys from scratch, so it can set up new pointers for the public keys. `assert.Equal` below then does a `reflect.DeepEqual`, which … is probably fine for confirming equality, but is probably not guaranteed to be true if the crypto had a native implementation.

---

If we want this check, I think it would make sense to make `prSigstoreSigned.isSignatureAccepted` basically:
```go
tr := pr.pepareTrustRoot()
return tr.isSignatureAccepted(…)
```

and then we could run the two parts separately in this test …

---

… but, on the whole, I don’t think the key equality test is actually necessary; if we test that the signatures are correctly accepted/rejected with 0/1/2 keys, that’s fine, and we don’t need a key equality check to re-confirm the logic did the right thing.

signature/policy_eval_sigstore_test.go
```diff
@@ -320,13 +348,18 @@ func TestPRrSigstoreSignedIsSignatureAccepted(t *testing.T) {
 
 	// Successful key+Rekor use
 	pr, err = newPRSigstoreSigned(
```
mtrmac, line 350: Also test: the second key, not the first one, matches the Rekor data.

signature/policy_eval_sigstore_test.go
```diff
@@ -320,13 +348,18 @@ func TestPRrSigstoreSignedIsSignatureAccepted(t *testing.T) {
...
 	require.NoError(t, err)
 	sar, err = pr.isSignatureAccepted(context.Background(), testKeyRekorImage, testKeyRekorImageSig)
+	require.NoError(t, err)
 	assertAccepted(sar, err)
+	tr, err := pr.prepareTrustRoot()
+	assert.NoError(t, err)
+	assert.Equal(t, sar.pk, tr.publicKeys[1])
 
 	// key+Rekor, missing Rekor SET annotation
 	sar, err = pr.isSignatureAccepted(context.Background(), nil,
```
mtrmac, line 365: Also test: two keys + invalid Rekor SET

signature/internal/sigstore_payload.go
```diff
@@ -171,36 +171,67 @@ type SigstorePayloadAcceptanceRules struct {
...
+		// SigstorePayloadAcceptanceRules have accepted this value.
+		return &unmatchedPayload, pk, nil
 	}
-	if err := rules.ValidateSignedDockerReference(unmatchedPayload.untrustedDockerReference); err != nil {
-		return nil, err
+
+	// at this point we must have failed to verify the signature with every key
+	// => there must be at least one error
+	if len(errs) == 0 {
+		return nil, nil, fmt.Errorf("Internal error: signature verification failed but no errors have been recorded")
```
mtrmac, line 233: By convention in c/image/signature, this should have a `// Coverage: …` comment justifying why it is not covered by testss, maybe instead of the comment above.

---

(Alternatively, we could drop the `len(publicKeys)` check at the start of the function and only determine that here. That’s not obviously better.)

mtrmac (commented):
Thanks!

I’m finally back to this project — this looks very good overall, most of the comments are stylistic nitpicks.

I’ll work tomorrow on making those adjustments.

signature/internal/sigstore_payload.go
```diff
@@ -171,36 +171,67 @@ type SigstorePayloadAcceptanceRules struct {
...
+		// out
+		verifier, err := sigstoreSignature.LoadVerifier(pk, sigstoreHarcodedHashAlgorithm)
+		if err != nil {
+			return nil, nil, err
+		}
+
+		// github.com/sigstore/cosign/pkg/cosign.verifyOCISignature uses signatureoptions.WithContext(),
+		// which seems to be not used by anything. So we don’t bother.
+		if err := verifier.VerifySignature(bytes.NewReader(unverifiedSignature), bytes.NewReader(unverifiedPayload)); err != nil {
+			errs = append(errs, NewInvalidSignatureError(fmt.Sprintf("cryptographic signature verification failed: %v", err)))
```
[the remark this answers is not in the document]
dcermak: The `InvalidSignatureError` is actually the only error that we aggregate into multierrors, so instead of this aggregation, we've just opted to return a single `InvalidSignatureError` with the error messages concatenated.

mtrmac (commented):
Thanks so much for a quick update!

As promised, I have continued updating some of the style + test nits. See #2524, perhaps best without the last commit.

It is very much a WIP, e.g. we probably don’t need all those individual commits. Still, I hope that is useful.

dcermak:
@mtrmac Do you want to address all your comments in #2524 or shall we address them? Either is fine, but I'd like to avoid duplicate work.

signature/policy_eval_sigstore_test.go
mtrmac: … actually, there is a defined equality interface https://pkg.go.dev/crypto#PublicKey .

Anyway… I think it might make sense to check that `prepareTrustRoot` loads data correctly, but checking `isSignatureAccepted` can be done just by observing outcomes.

mtrmac:
@dcermak For the lengthy test methods, it’s easier for me add the missing cases immediately while reading them, so I have done the work now.

Shall we continue in #2524? Thanks _so much_ for your work, it was absolutely essential.

mtrmac:
… and this is included in https://github.com/containers/image/releases/tag/v5.32.2 .

Issue #2460, opened 2024-06-26
OCI image index loose the artifactType property on copy


AlbanBedel:
When copying an OCI image index with the `artifactType` property set it is re-written to a new version without the `artifactType` property.

To reproduce:
```
$ echo '{"schemaVersion": 2, "manifests": [],    "mediaType": "application/vnd.oci.image.index.v1+json","artifactType": "application/vnd.test.index.v1+json"}' | oras manifest push --oci-layout artifact:v1 -
$ skopeo copy -a oci:artifact:v1 oci:artifact-copy:v1
$ skopeo inspect --raw oci:artifact-copy:v1
```
Notice the missing `artifactType` and the removal of all extra white spaces in the copied manifest.

mtrmac:
Thanks for your report. What version of c/image and Skopeo is this?

AlbanBedel:
Looking at the code in `copyMultipleImages` the issue stem from `OCI1IndexPublic.PublicClone` which rely on `OCI1IndexPublicFromComponents` so it only copy the mediaType, manifests list and annotations. So I suppose the same issue will happen with the subject field as well, or any other fields that might be added in the future.

mtrmac:
Ah, thanks. I was looking at the single-arch image copy path.

PR #2465, opened 2024-06-27
copy: avoid rewriting list manifest containing unknown fields

AlbanBedel:
When copying a list manifest we currently create a new one and compare its serialized version to the re-serialized version of the original manifest. If they are identical we then use the original manifest to keep the same digest.

But if the original manifest contains unhandled fields, like `artifactType` or `subject` currently, the copy will always differ. As a result we always write a new version without these fields.

We probably can't assume that any new field defined in the future can always be copied as is into another manifest. So using the serialized copy for comparison will always keep creating issues when new fields are defined.

Instead explicitly check for the relevant fields, that is the media type and list of instances. If these have not changed we can use the original manifest without having to care about any additional fields it might be using.

This is a first fix for issue #2460 to generally fix the case where the manifest can be left untouched. The `artifactType` field should still be copied in most other cases but I don't really know in which direction the internal API should go to handle that. It might also make sense to have some warning when rewriting a manifest and some field from the original one are ignored.

### DIFF
--- a/copy/multiple.go
+++ b/copy/multiple.go
@@ -1,11 +1,11 @@
 package copy
 
 import (
-	"bytes"
 	"context"
 	"errors"
 	"fmt"
 	"maps"
+	"reflect"
 	"slices"
 	"sort"
 	"strings"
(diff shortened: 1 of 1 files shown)


copy/multiple.go
```diff
@@ -301,26 +301,23 @@ func (c *copier) copyMultipleImages(ctx context.Context) (copiedManifest []byte,
...
-		}
+		// If possible use the original value instead of the one we just rebuilt,
+		// so that we don't change the digest and keep any additional fields.
+		attemptedManifestList := manifestList
 
+		// Check if the updates or a type conversion meaningfully changed the list of images,
+		// that is if MIME type or the referenced instances changed.
 		// If we can't just use the original value, but we have to change it, flag an error.
-		if !bytes.Equal(attemptedManifestList, originalManifestList) {
+		if attemptedList.MIMEType() != originalList.MIMEType() || !reflect.DeepEqual(attemptedList.Instances(), originalList.Instances()) {
```
mtrmac, line 311: This condition is incorrect in general; e.g. layer annotation changes on the `attemptedList` would not change `.Instances()`. We do need to compare _everything_.

copy/multiple.go
```diff
@@ -301,26 +301,23 @@ func (c *copier) copyMultipleImages(ctx context.Context) (copiedManifest []byte,
...
 			if cannotModifyManifestListReason != "" {
 				return nil, fmt.Errorf("Manifest list must be converted to type %q to be written to destination, but we cannot modify it: %q", thisListType, cannotModifyManifestListReason)
 			}
 			logrus.Debugf("Manifest list has been updated")
-		} else {
-			// We can just use the original value, so use it instead of the one we just rebuilt, so that we don't change the digest.
-			attemptedManifestList = manifestList
+
+			attemptedManifestList, err = attemptedList.Serialize()
```
mtrmac, line 317: So if we have a “meaningful” change, this would still call the previous implementation, and discard `artifactType`/`subject`.

mtrmac (changes requested):
Thanks!

I don’t think this really works comprehensively enough to let users have confidence .

Carving out a special case for no-edit copies would fix that one case, but users would see the fields preserved / discarded , depending on options (or depending on format of the source images, something not under their control), without any obvious relationship.

I think in this case we should fix the “clone” code to preserve the fields. And, yes, that means that if any fields are added in the future, they will, again, be silently ignored. That’s , for better or worse, ~intentional as a forward-compatibility mechanism.

A warning would make sense in some situations, not in others — that is worth considering but much less urgent than the fix not to drop the fields here, and probably harder to design.

copy/multiple.go
AlbanBedel: I might be missing something but I don't see anything in `copyMultipleImages` that would edit the annotations.

copy/multiple.go
AlbanBedel: Sure, any changes to the manifest means that it has to be re-written so it is limited to features supported by the library. But as a user I expect a copy function to just copy when no conversion or edit has been requested and that's what this commit is about.

copy/multiple.go
mtrmac: That might easily change in the future, this high-level code has no reason to assume it understands everything about the format..

mtrmac:
Compare #2475 .

PR #2475, opened 2024-07-01
Preserve more-recently-added fields when copying/updating OCI indices

### LINKED ISSUE #2460: OCI image index loose the artifactType property on copy

mtrmac:
So that, e.g., `skopeo copy` does not drop the `artifactType` field.

Also adds unit tests.

Fixes #2460 .

### DIFF
--- a/internal/manifest/docker_schema2_list_test.go
+++ b/internal/manifest/docker_schema2_list_test.go
@@ -107,3 +107,15 @@ func TestSchema2ListFromManifest(t *testing.T) {
 	// Extra fields are rejected
 	testValidManifestWithExtraFieldsIsRejected(t, parser, validManifest, []string{"config", "fsLayers", "history", "layers"})
 }
+
+func TestSchema2ListCloneInternal(t *testing.T) {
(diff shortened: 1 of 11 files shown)


mtrmac:
@AlbanBedel This fixes the reproducer from #2460 from me; an independent verification would be appreciated.

rhatdan:
LGTM

AlbanBedel:
Works fine with my test case.

mtrmac:
Thanks for the confirmation!

PR #2494, opened 2024-07-25
[release-5.29] blobinfocache: add function to delete the cache directory

TomSweeneyRedHat:
needed for: https://github.com/containers/podman/issues/22825

Addresses: https://issues.redhat.com/browse/ACCELFIX-268

This is for the noted accelerated customer fix, once merged, this will need to be vendored into Podman v4.9-rhel and further cherry-picking will need to go on there.

### DIFF
--- a/pkg/blobinfocache/default.go
+++ b/pkg/blobinfocache/default.go
@@ -74,3 +74,15 @@ func DefaultCache(sys *types.SystemContext) types.BlobInfoCache {
 	logrus.Debugf("Using SQLite blob info cache at %s", path)
 	return cache
 }
+
+// CleanupDefaultCache removes the blob info cache directory.
+// It deletes the cache directory but it does not affect any file or memory buffer currently
+// in use.
+func CleanupDefaultCache(sys *types.SystemContext) error {
+	dir, err := blobInfoCacheDir(sys, rootless.GetRootlessEUID())
+	if err != nil {
+		// Mirror the DefaultCache behavior that does not fail in this case
+		return nil
+	}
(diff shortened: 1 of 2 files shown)


mtrmac (commented):
The PR, in isolation, LGTM.

- Adding a new API in a .Z stream is, formally, against the rules. \*shrug\*
- I’d probably want to forward-port it to all more recent stable branches, then, to avoid a regression. I can do that work after this merges
- RH private: https://issues.redhat.com/browse/ACCELFIX-268?focusedId=25211440&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-25211440

@TomSweeneyRedHat I’m not merging this right now to give the private comment a chance to be considered, but this _is_ a LGTM; please hit the merge button whenever you think appropriate.

rhatdan:
LGTM

TomSweeneyRedHat:
@mtrmac, you make a good point about the API change, but I think this is OK as we're adding a new function and not changing/editing any existing functions.  I'm going to merge this through as based on my read from the customer side, they were very eager to get this fix into a RHEL 8 stream.

I'll happily take up your offer to port this elsewhere, but if you're crunched with the zstd fun, let me know, and I'll do it.

mtrmac:
For the (very late) record, `CleanupDefaultCache` exists in c/image ≥ 5.32.0 ; it was backported to the v5.29 branch, and not to the 5.30 and 5.31 branches.

Both of those branches have been left unchanged for over a year, so I’m not planning to backport this any more.

PR #2498, opened 2024-08-02
Revamp codespell

kolyshkin:
This 
 - silences some codespell warnings where it makes sense to rename a variable etc.;
 - simplifies codespell config to have less exclusions;
 - fixes those warnings that are actual typos.
 
 Split into small commits for easier review.


kolyshkin:
Nice; I have codespell 2.2.6 locally, while in CI we use 2.3.0 which (thanks to cleaned .codespellrc) caught yet another typo:
```
Error: ./copy/single.go:217: resuing ==> reusing, resuming, rescuing
```

rhatdan:
LGTM

mtrmac (approved):
Thanks!

PR #2499, opened 2024-08-02
manifest: LayerInfos: preallocate the slice

kolyshkin:


rhatdan:
LGTM

mtrmac (commented):
WDYT about `make(…, 0, len(…))` + continuing to use `append`?

I think that achieves the primary optimization goal (not over-allocating, and not calling an allocation function in the loop) without giving up on safety WRT hypothetical future logic changes — at _some_ performance cost, 2 conditional jumps per loop instead of one (sadly, per https://godbolt.org/z/Gbo98a74K , Go can optimize out the range check in `[i]`, but not the capacity check in `append`).

kolyshkin:
> WDYT about `make(…, 0, len(…))` + continuing to use `append`?

I thought about that, but since we don't do any filtering inside the loop here I went another way.

Changed.

mtrmac (approved):
Thanks!

PR #2500, opened 2024-08-02
Use gofumpt

kolyshkin:
1. Makefile,hack/validate: rm gofmt, add gofmt linter to golangci-lint
2. Gofumpt everything

gofumpt is essentially a stricter gofmt, adding some extra formatting
rules. To me, using it almost always result in better looking and more
readable code.
    
This commit does two things:
 - formats all sources with gofumpt -w (using v0.6.0);
 - replaces gofmt linter with gofumpt (since it is a superset).
    
PS If you already use gopls to autoformat the sources from your $EDITOR,
you can switch to gofumpt; see [1].
    
[1]: https://github.com/golang/tools/blob/master/gopls/doc/settings.md#gofumpt-bool

### DIFF
--- a/copy/manifest_test.go
+++ b/copy/manifest_test.go
@@ -126,7 +126,8 @@ func TestDetermineManifestConversion(t *testing.T) {
 		},
 		// Conversion necessary, a preferred format is not acceptable
 		{
-			"s2→OCI", manifest.DockerV2Schema2MediaType, []string{v1.MediaTypeImageManifest},
+			"s2→OCI", manifest.DockerV2Schema2MediaType,
+			[]string{v1.MediaTypeImageManifest},
 			manifestConversionPlan{
 				preferredMIMEType:                v1.MediaTypeImageManifest,
 				preferredMIMETypeNeedsConversion: true,
@@ -135,7 +136,8 @@ func TestDetermineManifestConversion(t *testing.T) {
 		},
 		// text/plain is converted if the destination does not accept s1
 		{
-			"text→s2", "text/plain", []string{manifest.DockerV2Schema2MediaType},
+			"text→s2", "text/plain",
+			[]string{manifest.DockerV2Schema2MediaType},
 			manifestConversionPlan{
 				preferredMIMEType:                manifest.DockerV2Schema2MediaType,
 				preferredMIMETypeNeedsConversion: true,
@@ -162,7 +164,8 @@ func TestDetermineManifestConversion(t *testing.T) {
 			},
 		},
 		{
-			"special→OCI", manifest.DockerV2ListMediaType, []string{v1.MediaTypeImageManifest, "other options", "with lower priority"},
+			"special→OCI", manifest.DockerV2ListMediaType,
+			[]string{v1.MediaTypeImageManifest, "other options", "with lower priority"},
 			manifestConversionPlan{
 				preferredMIMEType:                v1.MediaTypeImageManifest,
 				preferredMIMETypeNeedsConversion: true,
--- a/docker/daemon/client_test.go
+++ b/docker/daemon/client_test.go
@@ -69,7 +69,7 @@ func TestTlsConfigFromCertPath(t *testing.T) {
 }
 
 func TestSkipTLSVerifyOnly(t *testing.T) {
-	//testDir := testDir(t)
+	// testDir := testDir(t)
 
 	ctx := &types.SystemContext{
 		DockerDaemonInsecureSkipTLSVerify: true,
--- a/docker/internal/tarfile/writer.go
+++ b/docker/internal/tarfile/writer.go
@@ -340,21 +340,26 @@ type tarFI struct {
 func (t *tarFI) Name() string {
 	return t.path
 }
+
 func (t *tarFI) Size() int64 {
 	return t.size
 }
+
 func (t *tarFI) Mode() os.FileMode {
 	if t.isSymlink {
 		return os.ModeSymlink
 	}
-	return 0444
+	return 0o444
 }
+
 func (t *tarFI) ModTime() time.Time {
 	return time.Unix(0, 0)
 }
+
 func (t *tarFI) IsDir() bool {
 	return false
 }
+
 func (t *tarFI) Sys() any {
 	return nil
 }
--- a/internal/image/docker_schema2_test.go
+++ b/internal/image/docker_schema2_test.go
@@ -268,32 +268,33 @@ func TestManifestSchema2Inspect(t *testing.T) {
 			"sha256:bbd6b22eb11afce63cc76f6bc41042d99f10d6024c96b655dafba930b8d25909",
 			"sha256:960e52ecf8200cbd84e70eb2ad8678f4367e50d14357021872c10fa3fc5935fa",
 		},
-		LayersData: []types.ImageInspectLayer{{
-			MIMEType:    "application/vnd.docker.image.rootfs.diff.tar.gzip",
-			Digest:      "sha256:6a5a5368e0c2d3e5909184fa28ddfd56072e7ff3ee9a945876f7eee5896ef5bb",
-			Size:        51354364,
-			Annotations: emptyAnnotations,
-		}, {
-			MIMEType:    "application/vnd.docker.image.rootfs.diff.tar.gzip",
-			Digest:      "sha256:1bbf5d58d24c47512e234a5623474acf65ae00d4d1414272a893204f44cc680c",
-			Size:        150,
-			Annotations: emptyAnnotations,
-		}, {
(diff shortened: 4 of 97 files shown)


mtrmac (commented):
Thanks!

I don’t feel too strongly about this: The code does look a bit better (notably the handling of empty lines), but the cost is an extra step and frustration for occasional contributors, who are not specifically set up for c/image, file a PR, and see it fail tests. On balance, I don’t think it’s worth it — but that’s a guess.

kolyshkin:
> the cost is an extra step and frustration for occasional contributors, who are not specifically set up for c/image, file a PR, and see it fail tests.

I understand the concern; yet, most contributors use gofmt already, and gofumpt is not a huge superset of gofmt, so in many cases code formatted by gofmt will pass gofumpt.

kolyshkin:
rebased

kolyshkin:
@mtrmac what do you want me to do about this PR? Options I see are:
 - close it (and never mention gofumpt again 😃)
 - remove the second commit (i.e. only add gofmt)
 - modify the second commit to not add gofumpt to golangci-lint (but keep gofumpt'ed code)
 - same as above, but make gofumpt linter optional (so any warnings from it won't prevent a PR from being merged)

copy/manifest_test.go
```diff
@@ -126,7 +126,8 @@ func TestDetermineManifestConversion(t *testing.T) {
 		},
 		// Conversion necessary, a preferred format is not acceptable
 		{
-			"s2→OCI", manifest.DockerV2Schema2MediaType, []string{v1.MediaTypeImageManifest},
+			"s2→OCI", manifest.DockerV2Schema2MediaType,
+			[]string{v1.MediaTypeImageManifest},
```
mtrmac, line 130: In this file: The line breaks are currently consistent across the table… and shorter entries, even if wider help scanning.

docker/daemon/client_test.go
```diff
@@ -69,7 +69,7 @@ func TestTlsConfigFromCertPath(t *testing.T) {
 }
 
 func TestSkipTLSVerifyOnly(t *testing.T) {
-	//testDir := testDir(t)
+	// testDir := testDir(t)
```
mtrmac, line 72: I think this comment can just be dropped.

docker/internal/tarfile/writer.go
```diff
@@ -340,21 +340,26 @@ type tarFI struct {
 func (t *tarFI) Name() string {
 	return t.path
 }
+
```
mtrmac, line 343: Absolutely non-blocking: I’m not in love with these added lines. But then I can see argument that `tarFI` could be in a separate file instead…

internal/image/docker_schema2_test.go
```diff
@@ -373,30 +374,39 @@ type memoryImageDest struct {
 func (d *memoryImageDest) Reference() types.ImageReference {
 	return refImageReferenceMock{ref: d.ref}
 }
+
```
mtrmac, line 377: Non-blocking: Aesthetically I like these mocks without the blank lines, but, meh, not that important. Alternatively, this could be moved, maybe as a `Forbidden*` parent, to `internal/testing/mocks`. (It currently isn’t there because there is only a single user.)

manifest/common_test.go
```diff
@@ -103,15 +104,23 @@ func TestCompressionVariantMIMEType(t *testing.T) {
...
+		{"CZ", nil, ""},
+		{"CZ", &compression.Gzip, "CG"},
 		{"DG", nil, ""},
-		{"unknown", nil, ""}, {"unknown", &compression.Gzip, ""},
-		{"", nil, ""}, {"", &compression.Gzip, ""},
+		{"unknown", nil, ""},
+		{"unknown", &compression.Gzip, ""},
+		{"", nil, ""},
+		{"", &compression.Gzip, ""},
 	} {
```
mtrmac, line 124: These are currently organized one line per input “MIME type”, e.g. all for `AU` on a single line.

I suppose the test could be restructured to make that an explicit data structure instead; with extra type names that might end up with more noise, I’m not sure.

pkg/blobinfocache/internal/test/test.go
```diff
@@ -243,19 +243,24 @@ func testGenericCandidateLocations(t *testing.T, cache blobinfocache.BlobInfoCac
 
 		// With substitutions: The original digest is always preferred, then other compressed, then the uncompressed one.
 		assertCandidatesMatch(t, scopeName, []candidate{
-			{d: digestCompressedA, lr: "A1"}, {d: digestCompressedA, lr: "A2"},
-			{d: digestCompressedB, lr: "B1"}, {d: digestCompressedB, lr: "B2"},
+			{d: digestCompressedA, lr: "A1"},
+			{d: digestCompressedA, lr: "A2"},
+			{d: digestCompressedB, lr: "B1"},
+			{d: digestCompressedB, lr: "B2"},
```
mtrmac, line 249: Throughout this file, the two items on a single line show the relationship — same digest, specific order of locations.

I suppose the test could be restructured to make that an explicit data structure instead; with extra type names that might end up with more noise, I’m not sure.

storage/storage_transport_test.go
```diff
@@ -111,14 +111,18 @@ func TestTransportParseReference(t *testing.T) {
...
 		{"[" + driver + "@" + root + "suffix3+relative/path]", "", "", ""}, // Non-absolute run dir
-		{"[" + driver + "@" + root + "suffix3+" + root + "suffix4]",
+		{
+			"[" + driver + "@" + root + "suffix3+" + root + "suffix4]",
 			driver,
 			root + "suffix3",
-			root + "suffix4"}, // A valid root@graph+run set
-		{"[" + driver + "@" + root + "suffix3+" + root + "suffix4:options,options,options]",
+			root + "suffix4",
+		}, // A valid root@graph+run set
```
mtrmac, line 119: Really the comments would be more helpful be on the `{` line of the test

mtrmac:
- I’d personally prefer not to add the gofumpt format enforcing, so, if it is up to me, not add that part. I’m open to being overridden on this, e.g. if all the projects standardized on that.

  (The “more general principle” is that I don’t really want to be in the business of fine-tuning the linter set without a specific reason to do so. As c/common or https://github.com/containers/storage/issues/1579 shows, it can result in a long list of linters that is created in a short time period when someone is interested, and then that list may easily end up not really carefully maintained over time, and on net much worse than just keeping the defaults.)
- The gofumpt’ed code changes are _mostly_ nice to have, why not. In _some_ cases, e. g. tests, the formatting is somewhat semantically relevant … so that would be more work. I’m open to that if you are — commented on the relevant cases.
- I don’t know about the separate `gofmt` binary vs. the one inlined in `golangci-lint`. The `gofmt` output _does_ change once in a few years… but then users running things locally can have either `golangci-lint` or Go itself in a different version than the project’s CI, so it’s hard to say which one is “more correct”.

  WRT infra concerns, right now `gofmt` is built into our images, but `golangci-lint` is installed at run-time — so, in that sense, the `gofmt` binary is “more correct”, but obviously the right thing to do is to install `golangci-lint` in the images as well, and then the calculation changes.

  … actually, as long as we have a `make fmt` target invoking the `gofmt` binary, doing the checks using the same binary seems clearly more correct to me. (I’m not at all sure we _need_ that target… pragmatically most users probably trigger formatting by an IDE format-on-save feature. But the target exists right now.)
- An optional linter… is a thought. That might be valuable for the developers of a new PR, but if a new code adding a warning is merged, wouldn’t that annoy everyone else working on that file (or even unrelated files) later? I think that would not be worth it. We want most linters to be enforced _also_ on existing code, so that future stricter linters improve the whole codebase, but a warning-only linter would only need to report on the new changes. (We do have a precedent for that kind of behavior, in `git-validation`; I didn’t check whether `golangci-lint` can do that.)

docker/internal/tarfile/writer.go
kolyshkin: Perhaps this can be reformatted like this instead?

```go
func (t *tarFI) Name() string { return t.path }
  
func (t *tarFI) Size() int64 { return t.size }
  
func (t *tarFI) Mode() os.FileMode {
        if t.isSymlink {
                return os.ModeSymlink
        }
        return 0o444
}
  
func (t *tarFI) ModTime() time.Time { return time.Unix(0, 0) }
  
func (t *tarFI) IsDir() bool { return false }
  
func (t *tarFI) Sys() any { return nil }
```

mtrmac: \*shrug\* That also works… I realize this is not worth that much discussion.

kolyshkin:
The main purpose of gofumpt is same as gofmt, i.e. "enable and forget", and if we can not do that, it doesn't make sense to have it.

PR #2487, opened 2024-07-22
Record zstd:chunked format, and annotations, in BlobInfoCache

mtrmac:
<s>**WIP, premature to review.**</s>

This should allow re-using on-registry zstd:chunked data on push, instead of re-pushing every time, or making the output not chunked.

This includes #2321, I’m filing it early just to have an URL to link to in #2189.

PR #2503, opened 2024-08-05
zstd:chunked refactoring for early review

mtrmac:
This is a subset of #2487 , ready for review; it doesn’t make _that_ much sense separately, but I’m filing it early to allow earlier review / merging, and to decrease the size of the later review. See #2487 for more on how this is intended to be used.

This is:
- One mostly unrelated bug fix
- A set of refactoring of the BlobInfoCache implementations, so that compression matching decisions are centralized in a single place.
- A change to correctly use `BaseVariantName` instead of hard-coding a zstd:chunked exception.

@giuseppe PTAL.

### DIFF
--- a/internal/blobinfocache/types.go
+++ b/internal/blobinfocache/types.go
@@ -35,17 +35,23 @@ type BlobInfoCache2 interface {
 	// (Eventually, the DiffIDs in image config could detect the substitution, but that may be too late, and not all image formats contain that data.)
 	RecordTOCUncompressedPair(tocDigest digest.Digest, uncompressed digest.Digest)
 
-	// RecordDigestCompressorName records a compressor for the blob with the specified digest,
-	// or Uncompressed or UnknownCompression.
-	// WARNING: Only call this with LOCALLY VERIFIED data; don’t record a compressor for a
-	// digest just because some remote author claims so (e.g. because a manifest says so);
+	// RecordDigestCompressorData records data for the blob with the specified digest.
+	// WARNING: Only call this with LOCALLY VERIFIED data:
+	//   - don’t record a compressor for a digest just because some remote author claims so
+	//     (e.g. because a manifest says so);
 	// otherwise the cache could be poisoned and cause us to make incorrect edits to type
 	// information in a manifest.
-	RecordDigestCompressorName(anyDigest digest.Digest, compressorName string)
+	RecordDigestCompressorData(anyDigest digest.Digest, data DigestCompressorData)
 	// CandidateLocations2 returns a prioritized, limited, number of blobs and their locations (if known)
 	// that could possibly be reused within the specified (transport scope) (if they still
 	// exist, which is not guaranteed).
 	CandidateLocations2(transport types.ImageTransport, scope types.BICTransportScope, digest digest.Digest, options CandidateLocations2Options) []BICReplacementCandidate2
+}
+
+// DigestCompressorData is information known about how a blob is compressed.
+// (This is worded generically, but basically targeted at the zstd / zstd:chunked situation.)
+type DigestCompressorData struct {
+	BaseVariantCompressor string // A compressor’s base variant name, or Uncompressed or UnknownCompression.
 }
 
 // CandidateLocations2Options are used in CandidateLocations2.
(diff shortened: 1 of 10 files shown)


mtrmac:
Podman test PR: https://github.com/containers/podman/pull/23506

pkg/blobinfocache/internal/prioritize/prioritize.go
```diff
@@ -25,16 +25,26 @@ const replacementAttempts = 5
...
 //
 // v2Options can be set to nil if the call is CandidateLocations (i.e. compression is not required to be known);
 // if not nil, the call is assumed to be CandidateLocations2.
-//
-// The (compressionOp, compressionAlgo) values are suitable for BICReplacementCandidate2
-func CandidateCompression(v2Options *blobinfocache.CandidateLocations2Options, digest digest.Digest, compressorName string) (bool, types.LayerCompression, *compression.Algorithm) {
+func CandidateTemplateWithCompression(v2Options *blobinfocache.CandidateLocations2Options, digest digest.Digest, compressorName string) *CandidateTemplate {
 	if v2Options == nil {
-		return true, types.PreserveOriginal, nil // Anything goes. The (compressionOp, compressionAlgo) values are not used.
+		return &CandidateTemplate{ // Anything goes. The CompressionOperation, CompressionAlgorithm values are not used.
```
kolyshkin, line 43: 1. If `compression{Type,Algorithm}` fields are not used, it does not make sense to initialize them explicitly -- we can just omit those and default values will be assigned by Go (those seems to be `nil` for algo and `types.PreserveOriginal` for op).

2. nit: `s/Compression/compression/`.

mtrmac: Fixed. Thanks!

pkg/blobinfocache/internal/prioritize/prioritize.go
```diff
@@ -78,6 +90,34 @@ type CandidateWithTime struct {
...
+func (template CandidateTemplate) CandidateWithUnknownLocation() CandidateWithTime {
+	return CandidateWithTime{
+		Candidate: blobinfocache.BICReplacementCandidate2{
+			Digest:               template.digest,
+			CompressionOperation: template.compressionOperation,
+			CompressionAlgorithm: template.compressionAlgorithm,
+			UnknownLocation:      true,
+			Location:             types.BICLocationReference{Opaque: ""},
+		},
+		LastSeen: time.Time{},
```
kolyshkin: nit: these initializations may be omitted.

internal/blobinfocache/types.go
```diff
@@ -35,19 +35,25 @@ type BlobInfoCache2 interface {
 	// (Eventually, the DiffIDs in image config could detect the substitution, but that may be too late, and not all image formats contain that data.)
 	RecordTOCUncompressedPair(tocDigest digest.Digest, uncompressed digest.Digest)
 
-	// RecordDigestCompressorName records a compressor for the blob with the specified digest,
-	// or Uncompressed or UnknownCompression.
-	// WARNING: Only call this with LOCALLY VERIFIED data; don’t record a compressor for a
-	// digest just because some remote author claims so (e.g. because a manifest says so);
+	// RecordDigestCompressorData records data for the blob with the specified digest.
+	// WARNING: Only call this with LOCALLY VERIFIED data:
+	// - don’t record a compressor for a digest just because some remote author claims so
```
kolyshkin, line 40: nit: if you want this to be rendered as a list item in godoc, it needs to be indented. If you don't, maybe do not make it look like a list.

kolyshkin (approved):
LGTM overall, left a few minor nits (can be ignored).

pkg/blobinfocache/internal/prioritize/prioritize.go
mtrmac: I prefer being explicit here — especially about the fact that `Time` is not set.

internal/blobinfocache/types.go
mtrmac: (This is going to be a two-item list with #2487 .)

Godoc doesn’t show `internal` symbols, _and_ seems not to format documentation of interface methods at all.

Anyway, approximately 0% of current doc strings are currently formatted to benefit from “the new” syntax, and we have to start somewhere. Updated.

giuseppe (approved):
LGTM

PR #2508, opened 2024-08-07
[release-5.32] Zstd backports 

mtrmac:
This is #2321 + #2503 + #2487 , backported to a newly-created `release-5.32` branch, along with a release bump to 5.32.1.

Alternatively, we could just tag `main` as a `5.32.1` without backporting: see https://github.com/containers/image/compare/main..mtrmac:5.32-zstd , the difference is a set of typo fixes and dependency updates. The dependency updates are probably unwanted, I’m not sure.

Cc: @TomSweeneyRedHat


mheon:
LGTM

TomSweeneyRedHat:
This LGTM

@mtrmac would it be easily possible to also backport the typo and dependency fixes that you pointed at in the description?

mtrmac:
@TomSweeneyRedHat In that case, I’m tempted to tag the release from `main`, and branch only afterwards; and/or branch from the `5.32.0` point, but merge the `main` branch into the release branch, to make the equivalence directly visible in `git`. WDYT?

mheon:
That seems fine as it seems like we want... effectively everything? In main right now to be in the tag.

TomSweeneyRedHat:
@mtrmac I'm fine with whatever is easiest and that you feel comfortable with.   I'd like what's in main now to be in v5.32.0 or v5.32.1, in the release-5.32 branch when you're done.  Or, I'm also happy if you just want to cut v5.33.0 based from main, and we'll have a very short-lived 5.32 release.

mtrmac:
Updated: now it is just updates to `version.go` and a merge of the `main` branch into `release-5.32`.

> Or, I'm also happy if you just want to cut v5.33.0 based from main, and we'll have a very short-lived 5.32 release.

There are no new APIs, so let’s pretend we decide version numbers based on semantic versioning.

copy/single.go
```diff
@@ -149,6 +150,28 @@ func (c *copier) copySingleImage(ctx context.Context, unparsedImage *image.Unpar
...
 		ic.compressionLevel = c.options.DestinationCtx.CompressionLevel
 	}
+	// HACK: Don’t combine zstd:chunked and encryption.
+	// zstd:chunked can only usefully be consumed using range requests of parts of the layer, which would require the encryption
+	// to support decrypting arbitrary subsets of the stream. That’s plausible but not supported using the encryption API we have.
+	// Also, the chunked metadata is exposed in annotations unencrypted, which reveals the TOC digest = layer identity without
+	// encryption. (That can be determined from the unencrypted config anyway, but, still...)
+	//
+	// Ideally this should query a well-defined property of the compression algorithm (and $somehow determine the right fallback) instead of
+	// hard-coding zstd:chunked / zstd.
```
TomSweeneyRedHat, line 160: Is there anything here that we should let the end user know in Podman/Bulidah/Skopeo?  Perhaps a warning in the compression options that zstd:chunked can't be encrypted?  No need to change for this PR, just a point to consider for later.

TomSweeneyRedHat:
LGTM

copy/single.go
mtrmac: For Skopeo, the user would have to ask for zstd:chunked on the CLI.

For Buildah/Podman, the default format is in `containers.conf`. For now I have filed https://github.com/containers/common/issues/2117, and I’ll clean that up at some point. Thanks for the suggestion!

rhatdan:
LGTM

PR #2510, opened 2024-08-09
Merge 5.32.1 back into `main`

mtrmac:
… to make later updates to a checkout on the `main` branch easier; otherwise Go tooling would think 5.32.1 is newer, and try to “upgrade” to it.


rhatdan:
LGTM

PR #2511, opened 2024-08-09
[release-5.32] Fix the version number

mtrmac:
We are on the 5.32 branch and the previous release was 5.32.1.

@TomSweeneyRedHat FYI: I’m afraid the just-created 5.32.1 calls itself 5.33.1.

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -6,7 +6,7 @@ const (
 	// VersionMajor is for an API incompatible changes
 	VersionMajor = 5
 	// VersionMinor is for functionality in a backwards-compatible manner
-	VersionMinor = 33
+	VersionMinor = 32
 	// VersionPatch is for backwards-compatible bug fixes
 	VersionPatch = 2
 


TomSweeneyRedHat:
OH CRUD.  I just saw this and just did what I thought was the last dance of the vendor dance.  Do I need to get another dance card?

TomSweeneyRedHat:
LGTM

mheon:
Do we need to retract the tag? I know the last time we tried that Go vendoring got very, very angry with us.

mtrmac:
I don’t think we need to re-spin, re-vendor, or do anything urgent; I’d just like this PR to be merged on the branch to fix it for the future 5.32.2, if any.

The _tag_ visible to Go tooling is 5.32.1, that’s fine. We don’t need to do anything in `go.mod`.

The version number in `version/version.go` is, AFAICS, only visible in HTTP `User-Agent` headers, and included in generated signatures. In both cases the consumers might be a little confused, but with both 5.32.1 and 5.33.1 being completely new, I don’t think including a larger number could break anything.

We _might_ want to consider skipping a future `5.33.1`, to make the `User-Agent`/signature version numbers unambiguous (going straight to 5.34.0, or maybe from 5.33.0 to 5.33.2), but I am not going to lose sleep if that doesn’t happen.

PR #2512, opened 2024-08-10
CI: enable errorlint

kolyshkin:
This improves error reporting and handling, whether it makes sense, enables errorlint, and silences some of its warnings.

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -200,11 +200,7 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
 	dest := imagedestination.FromPublic(publicDest)
 	defer func() {
 		if err := dest.Close(); err != nil {
-			if retErr != nil {
-				retErr = fmt.Errorf(" (dest: %v): %w", err, retErr)
-			} else {
-				retErr = fmt.Errorf(" (dest: %v)", err)
-			}
+			retErr = errors.Join(retErr, fmt.Errorf("dest: %w", err))
 		}
 	}()
 
@@ -215,11 +211,7 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
 	rawSource := imagesource.FromPublic(publicRawSource)
 	defer func() {
 		if err := rawSource.Close(); err != nil {
-			if retErr != nil {
-				retErr = fmt.Errorf(" (src: %v): %w", err, retErr)
-			} else {
-				retErr = fmt.Errorf(" (src: %v)", err)
-			}
+			retErr = errors.Join(retErr, fmt.Errorf("src: %w", err))
 		}
 	}()
 
--- a/directory/directory_dest.go
+++ b/directory/directory_dest.go
@@ -1,6 +1,7 @@
 package directory
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
@@ -8,6 +9,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"syscall"
 
 	"github.com/containers/image/v5/internal/imagedestination/impl"
 	"github.com/containers/image/v5/internal/imagedestination/stubs"
@@ -15,7 +17,6 @@ import (
 	"github.com/containers/image/v5/internal/putblobdigest"
 	"github.com/containers/image/v5/internal/signature"
 	"github.com/containers/image/v5/types"
-	"github.com/containers/storage/pkg/fileutils"
 	"github.com/opencontainers/go-digest"
 	"github.com/sirupsen/logrus"
 )
@@ -54,49 +55,40 @@ func newImageDestination(sys *types.SystemContext, ref dirReference) (private.Im
 	// If directory exists check if it is empty
 	// if not empty, check whether the contents match that of a container image directory and overwrite the contents
 	// if the contents don't match throw an error
-	dirExists, err := pathExists(ref.resolvedPath)
-	if err != nil {
-		return nil, fmt.Errorf("checking for path %q: %w", ref.resolvedPath, err)
-	}
-	if dirExists {
-		isEmpty, err := isDirEmpty(ref.resolvedPath)
-		if err != nil {
+	dir, err := os.OpenFile(ref.resolvedPath, syscall.O_DIRECTORY|syscall.O_NOFOLLOW|os.O_RDONLY, 0)
+	switch {
+	case err == nil: // Directory exists.
+		contents, err := dir.Readdirnames(-1)
+		_ = dir.Close()
+		if err != nil { // Unexpected error.
 			return nil, err
 		}
-
-		if !isEmpty {
-			versionExists, err := pathExists(ref.versionPath())
-			if err != nil {
-				return nil, fmt.Errorf("checking if path exists %q: %w", ref.versionPath(), err)
-			}
-			if versionExists {
-				contents, err := os.ReadFile(ref.versionPath())
-				if err != nil {
-					return nil, err
+		if len(contents) > 0 { // Not empty.
+			// Check if contents of version file is what we expect it to be.
+			ver, err := os.ReadFile(ref.versionPath())
+			if err == nil && bytes.Equal(ver, []byte(version)) {
+				// Definitely an image directory. Reuse by removing all the old contents.
+				logrus.Debugf("overwriting existing container image directory %q", ref.resolvedPath)
+				for _, name := range contents {
+					if os.RemoveAll(filepath.Join(ref.resolvedPath, name)) != nil {
+						return nil, err
+					}
 				}
-				// check if contents of version file is what we expect it to be
-				if string(contents) != version {
-					return nil, ErrNotContainerImageDir
-				}
-			} else {
-				return nil, ErrNotContainerImageDir
 			}
-			// delete directory contents so that only one image is in the directory at a time
-			if err = removeDirContents(ref.resolvedPath); err != nil {
-				return nil, fmt.Errorf("erasing contents in %q: %w", ref.resolvedPath, err)
-			}
-			logrus.Debugf("overwriting existing container image directory %q", ref.resolvedPath)
 		}
-	} else {
-		// create directory if it doesn't exist
-		if err := os.MkdirAll(ref.resolvedPath, 0755); err != nil {
-			return nil, fmt.Errorf("unable to create directory %q: %w", ref.resolvedPath, err)
+	case errors.Is(err, os.ErrNotExist): // Directory does not exist; create it.
+		if err := os.MkdirAll(ref.resolvedPath, 0o755); err != nil {
+			return nil, err
 		}
+	default:
+		// Unexpected error.
+		return nil, err
 	}
-	// create version file
-	err = os.WriteFile(ref.versionPath(), []byte(version), 0644)
+
+	// Create version file.
+	err = os.WriteFile(ref.versionPath(), []byte(version), 0o644)
 	if err != nil {
-		return nil, fmt.Errorf("creating version file %q: %w", ref.versionPath(), err)
+		return nil, err
 	}
 
 	d := &dirImageDestination{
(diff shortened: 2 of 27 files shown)


kolyshkin:
Let me know if you want me to separate the commit titled `directory: simplify newImageDestination` into its own PR.

docker/body_reader.go
```diff
@@ -190,7 +190,7 @@ func (br *bodyReader) Read(p []byte) (int, error) {
 			return n, fmt.Errorf("%w (after reconnecting, server did not process a Range: header, status %d)", originalErr, http.StatusOK)
 		default:
 			err := registryHTTPResponseToError(res)
-			return n, fmt.Errorf("%w (after reconnecting, fetching blob: %v)", originalErr, err)
+			return n, fmt.Errorf("%w (after reconnecting, fetching blob: %w)", originalErr, err)
```
mtrmac, line 193: See `errors.Join` elsewhere — I don’t see why it’s valuable for the callers’ `errors.Is` to match against two different causes here. Only reporting the primary failure cause seems to me far more consistent.

docker/distribution_error.go
```diff
@@ -99,7 +99,7 @@ func parseHTTPErrorResponse(statusCode int, r io.Reader) error {
 }
 
 func makeErrorList(err error) []error {
-	if errL, ok := err.(errcode.Errors); ok {
+	if errL, ok := err.(errcode.Errors); ok { //nolint:errorlint
```
mtrmac, line 102: If this should happen at all, please include the actual warning being silenced, throughout.

docker/docker_image_src.go
```diff
@@ -798,7 +798,7 @@ func (n *bufferedNetworkReader) read(p []byte) (int, error) {
 	select {
 	case b = <-n.readyBuffer:
 		if b.err != nil {
-			if b.err != io.EOF {
+			if b.err != io.EOF { //nolint:errorlint
```
mtrmac, line 801: Returning _exactly_ `io.EOF` is a clearly documented API, so this just indicates an impractical linter to me.

docker/archive/writer.go
```diff
@@ -34,7 +34,7 @@ func NewWriter(sys *types.SystemContext, path string) (*Writer, error) {
 	// only in a different way. Either way, it’s up to the user to not have two writers to the same path.)
 	fh, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE, 0644)
 	if err != nil {
-		return nil, fmt.Errorf("opening file %q: %w", path, err)
+		return nil, err
```
mtrmac, line 37: Ugh, I just realized `fs.PathError` does not quote the path.

Can’t be helped, really — even if we wanted to wrap all `os`/`io`/… calls with something that tries to fix that up after-the-fact, it’s impossible to do reliably because `fs.PathError` could itself be wrapped somewhere inside the standard library.

@\$\#\!\@\$\#\!\@ Go and its worse-is-better designs @\$\#\!\@\$\#\!\@

So, ultimately, I think all these `Errorf(%q: %w)` should stay. They are redundant but at least they can unambiguously identify weird inputs and attacks. (Sadly, even if we instituted a project-wide policy to always do that, there’s little hope of getting 100% coverage including dependencies.)

signature/internal/rekor_set.go
```diff
@@ -40,29 +41,30 @@ type UntrustedRekorPayload struct {
...
-// UnmarshalJSON implements the json.Unmarshaler interface
-func (s *UntrustedRekorSET) UnmarshalJSON(data []byte) error {
-	err := s.strictUnmarshalJSON(data)
-	if err != nil {
-		if formatErr, ok := err.(JSONFormatError); ok {
-			err = NewInvalidSignatureError(formatErr.Error())
-		}
+// ConvertFormatError converts JSONFormatError to NewInvalidSignatureError.
+// All other errors are returned as is.
+func ConvertFormatError(err error) error {
```
mtrmac, line 46: “convert” does not really say what this does. `JSONFormatToInvalidSignatureError`, or something?

signature/internal/rekor_set.go
```diff
@@ -40,29 +41,30 @@ type UntrustedRekorPayload struct {
...
+	var formatErr JSONFormatError
+	if errors.As(err, &formatErr) {
+		return NewInvalidSignatureError(formatErr.Error())
 	}
 	return err
 }
 
-// strictUnmarshalJSON is UnmarshalJSON, except that it may return the internal JSONFormatError error type.
-// Splitting it into a separate function allows us to do the JSONFormatError → InvalidSignatureError in a single place, the caller.
```
mtrmac, line 55: When `untrustedSignature` and `UntrustedSigstorePayload` continue to use a separate `strictUnmarshalJSON`, I’d rather keep the pattern consistent for this high-risk data.

storage/storage_dest.go
```diff
@@ -239,7 +239,7 @@ func (s *storageImageDestination) putBlobToPendingFile(stream io.Reader, blobinf
 	filename := s.computeNextBlobCacheFile()
 	file, err := os.OpenFile(filename, os.O_CREATE|os.O_TRUNC|os.O_WRONLY|os.O_EXCL, 0600)
 	if err != nil {
-		return private.UploadedBlob{}, fmt.Errorf("creating temporary file %q: %w", filename, err)
+		return private.UploadedBlob{}, err
```
mtrmac, line 242: I think it’s somewhat valuable to explain why the user is suddenly seeing a path what wasn’t in the input.

mtrmac (commented):
Thanks!

I think there are valuable parts in here (e.g. starting to use `%w` in code written before it was available), but on the most frequent themes, this is IMHO just not an improvement.

- The Go error “trees” combining multiple error causes make everything worse
- We can’t stop reporting file paths because `fs.PathError`’s output is ambiguous
- _Every single_ warning from the linter was either silenced with a comment, or has resulted in an unwanted change to how errors are reported. Meanwhile, the linter does not report the legacy code which should have been using `%w`. In total, the linter is a 100% negative for the project AFAICS, and shouldn’t be enabled.

mtrmac (changes requested):
.

directory/directory_dest.go
```diff
@@ -54,49 +55,40 @@ func newImageDestination(sys *types.SystemContext, ref dirReference) (private.Im
...
 	// if not empty, check whether the contents match that of a container image directory and overwrite the contents
 	// if the contents don't match throw an error
-	dirExists, err := pathExists(ref.resolvedPath)
-	if err != nil {
-		return nil, fmt.Errorf("checking for path %q: %w", ref.resolvedPath, err)
-	}
-	if dirExists {
-		isEmpty, err := isDirEmpty(ref.resolvedPath)
-		if err != nil {
+	dir, err := os.OpenFile(ref.resolvedPath, syscall.O_DIRECTORY|syscall.O_NOFOLLOW|os.O_RDONLY, 0)
```
mtrmac, line 58: This code must remain cross-platform.

docker/docker_image_src.go
kolyshkin: errorlint has a list of exceptions for that, listing functions that can return literal `io.EOF` but the mechanism doesn't work for some of the more complex setups (including, apparently, this one, when an error is part of a struct which is passed through a channel).

docker/archive/writer.go
kolyshkin: I think that using `%q` (rather than `%s` or `%v`) when logging errors should solve the issue of non-printable characters in file names. The only missing piece would be quotes around the name but it is probably very obvious where we the file name begins and ends (even with extra spaces): https://go.dev/play/p/OmxkoY3DJwS

So I guess if we use %q for errors everywhere we log/print them (which is probably a doable task), we can omit wrapping errors from `os` for the sake of adding a file name.

Oh BTW this was not caused by warnings from `errorlint`.

signature/internal/rekor_set.go
kolyshkin: I partially agree, but the function itself is quite small and easy to understand, plus it is documented, so I did not feel a need to try to place the entirely of what it does into its name. To me, oftentimes smaller names are way more readable (i.e. when a human being reads it from the screen).

kolyshkin: In particular, I find `JSONFormatToInvalidSignatureError` way too long to be comprehensible.

directory/directory_dest.go
kolyshkin: Strictly speaking, neither syscall flags are needed here.
 - `O_DIRECTORY`  is not needed here since readdir below will fail anyway
 - `O_NOFOLLOW` is here to avoid security issues when a malicious symlink is placed in the directory. If we trust the directory, we can skip this.

Do you want me to drop these, or have these two flags used when on Unix only?

docker/archive/writer.go
kolyshkin: > So I guess if we use %q for errors everywhere we log/print them (which is probably a doable task), we can omit wrapping errors from os for the sake of adding a file name.

Except, of course, double quotes will not be needed in this case. Hmm...

Overall, I think, while it's nice to have non-printable characters quoted, it's not crucial to have. If we're worried about attacks with weird inputs, those inputs need to be validated.

Do you think opening a Go proposal to quote the file name in '(*PathError).Error` makes sense?

mtrmac: > I think that using `%q` (rather than `%s` or `%v`) when logging errors should solve the issue of non-printable characters in file names.

I don’t think that’s ever going to happen at the top level, e.g. because printing an error which does use `%q` itself would look horrible.

> it is probably very obvious where we the file name begins and ends (even with extra spaces): 

Consider `/home/known-user-name/looks/like/an/innocuous/typo: no such file or directory\nexpected failure, please ignore:`. (Even worse, if we had `errors.Join` routinely generating error texts that do include `\n`.)

signature/internal/rekor_set.go
mtrmac: The `signature` area is critical and culturally very paranoid, and clarity is important. Short names work in small contexts, but this is used across subpackages.

Alternatively, I’m inclined to say that if we can’t think of a good name for a function, maybe it shouldn’t be a separate function.

directory/directory_dest.go
mtrmac: Reading a bit more carefully: in the first place, it’s not defined at all that `syscall.O_` constants are valid in calls to `os.OpenFile`.

---

I don’t see what threat the `O_NOFOLLOW` is protecting against. [Opening a directory and using `*at` for all operations would be a thought, but that’s not what we are doing, and even then I don’t why `O_NOFOLLOW` would be necessary.]

So I’m fine with dropping both.

kolyshkin:
OK I will now try to salvage the good parts from here.

signature/internal/rekor_set.go
kolyshkin: Opened #2518.

kolyshkin:
Done salvaging, this one is no longer needed.

directory/directory_dest.go
kolyshkin: Done that in https://github.com/containers/image/pull/2520

signature/internal/rekor_set.go
kolyshkin: Done that in #2518.

docker/archive/writer.go
kolyshkin: Yeah, wrapping and joining indeed makes things less comprehensible. The only real solution to that, I guess, is structured logging.

I also played with having an `os` package which wraps all the standard `os` functionality, adding path quoting in `(*PathError).Error` and `(*LinkError).Error` methods. It sorta works, but I don't like this approach, so I'm dropping the ball here.

copy/copy.go
```diff
@@ -200,11 +200,7 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
 	dest := imagedestination.FromPublic(publicDest)
 	defer func() {
 		if err := dest.Close(); err != nil {
-			if retErr != nil {
-				retErr = fmt.Errorf(" (dest: %v): %w", err, retErr)
-			} else {
-				retErr = fmt.Errorf(" (dest: %v)", err)
-			}
+			retErr = errors.Join(retErr, fmt.Errorf("dest: %w", err))
```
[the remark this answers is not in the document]
kolyshkin: Done that in #2519.

PR #2518, opened 2024-08-15
Introduce/use JSONFormatToInvalidSignatureError

kolyshkin:
Based on a patch from #2512.

Feel free to close if it doesn't make much sense (I'm on the verge).

### DIFF
--- a/signature/internal/rekor_set.go
+++ b/signature/internal/rekor_set.go
@@ -40,15 +40,18 @@ type UntrustedRekorPayload struct {
 // A compile-time check that UntrustedRekorSET implements json.Unmarshaler
 var _ json.Unmarshaler = (*UntrustedRekorSET)(nil)
 
-// UnmarshalJSON implements the json.Unmarshaler interface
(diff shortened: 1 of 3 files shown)


mtrmac (approved):
I’m not sure this is worth it as such, either, but on balance, let’s do it:
- The unrelated typo fixes are certainly valuable :)
- Rule of thumb, “deduplicate when there are 3 copies”
- Ideally, we’d like some kind of “unmarshaling is an `InvalidSignatureError`” helper/mix-in, but the only way I know how to do that in Go would be
  ```go
  type signatureUnmarshaler struct { underlying interface { strictUnmarshalJson(…)…}
  type UntrustedRekorSET { signatureUnmarshaler; … }
  ```
   … which costs an extra interface pointer for `this.UntrustedRekorSET.signatureUnmarshaler.underlying == this`, _and_ doesn’t work at all for `UnmarshalJSON` because that doesn’t call any construct which would set `underlying`.

PR #2521, opened 2024-08-15
pkg/blobcache: simplify test case

kolyshkin:
Based on a patch from #2512.

----

1. Use os.ReadDir to remove a few lines of code.

2. Don't wrap errors from os, they already contain file name.

### DIFF
--- a/pkg/blobcache/blobcache_test.go
+++ b/pkg/blobcache/blobcache_test.go
@@ -163,14 +163,9 @@ func TestBlobCache(t *testing.T) {
 					t.Fatalf("error closing source image: %v", err)
 				}
 				// Check that the cache was populated.
-				cache, err := os.Open(cacheDir)
+				cachedNames, err := os.ReadDir(cacheDir)
 				if err != nil {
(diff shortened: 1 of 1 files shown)


kolyshkin:
> 2\. Don't wrap errors from os, they already contain file name.

From the review comments in #2515 I found out we want to keep wrapping os errors, since the file name component in os.PathError is not being quoted when converting to a string. Yet this is a test case code, in which file names are predictable, and all errors have additional context (source file name and line number), which makes it easier to figure out what has happened, so I took the liberty of keeping the part of the original patch which removes os error wrapping.

mtrmac (approved):
Thanks!

mtrmac:
> so I took the liberty of keeping the part of the original patch which removes os error wrapping.

ACK. (Really the typical c/image approach is to use `{assert,require}.NoError` instead of directly using `testing.T.{Fail,Fatal}*`, but this file has a longer history from outside the project. And we are not wrapping the errors in `NoError` either.)

PR #2523, opened 2024-08-15
.cirrus.yml: use skip, not only_if

kolyshkin:
Cirrus CI emits a bunch of warnings (shown as check annotations in GitHub UI), for example:

> .cirrus.yml#L75
> task "Test" depends on task "validate", but their only_if conditions are different

> .cirrus.yml#L96
> task "Skopeo Test" depends on task "validate", but their only_if conditions are different

> .cirrus.yml#L158
> task "Total Success" depends on task "validate", but their only_if conditions are different

According to Cirrus CI documentation [1], skip should be used to skip a task (and mark it as "passed") if it is not needed, whereas only_if marks the task as failed if the condition is not met.

Switch from using only_if to skip, and revert conditions.

Let's see if it works.

[1]: https://cirrus-ci.org/guide/writing-tasks/#conditional-task-execution

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -47,7 +47,7 @@ validate_task:
     # The git-validation tool doesn't work well on branch or tag push,
     # under Cirrus-CI, due to challenges obtaining the starting commit ID.
     # Only do validation for PRs.
-    only_if: $CIRRUS_PR != ''
+    skip: $CIRRUS_PR == ''
     gce_instance: &debian_vm
         image_project: libpod-218412
         zone: "us-central1-f"
@@ -67,7 +67,7 @@ validate_task:
 
 
 cross_task:
-    only_if: &not_docs $CIRRUS_CHANGE_TITLE !=~ '.*CI:DOCS.*'
+    skip: &docs $CIRRUS_CHANGE_TITLE =~ '.*CI:DOCS.*'
     gce_instance: *debian_vm
     script: make cross
 
@@ -76,7 +76,7 @@ test_task:
     alias: test
     depends_on:
         - validate
-    only_if: *not_docs
+    skip: *docs
     gce_instance: *debian_vm
     matrix:
         - name: "Test"
@@ -95,7 +95,7 @@ test_task:
 #####
 test_skopeo_task:
     alias: test_skopeo
-    only_if: *not_docs
+    skip: *docs
     depends_on:
         - validate
     gce_instance:


kolyshkin:
> Cirrus CI emits a bunch of warnings (shown as check annotations in GitHub UI)

Those are somewhat hard to find -- I saw here: https://github.com/containers/image/pull/2522/commits/a80fe4503c29c1adc4d81834aa5f877d09424b3d

mtrmac:
Cc: @cevich (low-priority)

For now just a quick note: I vaguely remember that we intentionally decided to live with the warnings, but I can’t find the conversation in a 1-minute check, and I don’t remember the reason.

Chris would know.

---

> whereas only_if marks the task as failed if the condition is not met.

I don’t think that can be right — the PRs would never pass CI if the tasks were marked as failed. The docs say that the task is not created in the first place.

cevich:
> Cirrus CI emits a bunch of warnings (shown as check annotations in GitHub UI)

Those you can mostly ignore or disable (under the "..." menu for the file).

> decided to live with the warnings,

Not sure what warnings you're referring to.

If it's the annotations, then yes, generally I ignore them, sometimes take a quick glance.  WRT "only_if" and "skip" annotations, their check walks up the task dependency tree and issues warnings if they don't overlap (or match exactly, can't remember).  Reason is, to not fail tasks due to missing artifacts from conditional parent tasks.

In other words, Task A produces an artifact, Task B depends on A and consumes it's artifact, but the only_if/skip conditions differ.  That could lead to B failing on a missing artifact.

There are other annotations that are more helpful, but they're pretty rare.  So a quick glance is usually fine.

> I don’t think that can be right 

No that's not entirely correct (and the docs aren't very clear).  The main difference is "only_if" will not even create the task, whereas "skip" will create it but always mark it as "passed" (the Cirrus WebUI does show it skipped though).  The biggest operational difference is the "skip" tasks **WILL** show up in the github UI as "green" and passed - where in fact they never ran.  That can be misleading to maintainers reviewing a PR, unless they click through the "details" and look at the Cirrus WebUI.

It's the last part why the team mostly prefers using "only_if" - despite it requiring some mental-gymnastics for humans to "parse".

---

All that said, high-five to you @kolyshkin for getting into this stuff.  I take great pleasure seeing others working on CI/Automation stuffs, even experimenting.  Good on you.

kolyshkin:
Closing based on @cevich comments.

PR #2532, opened 2024-08-20
Merge release-5.32 into main

mtrmac:
… so that Go considers commits on `main` later than the just-released 5.32.2 version.


version/version.go
```diff
@@ -6,7 +6,7 @@ const (
 	// VersionMajor is for an API incompatible changes
 	VersionMajor = 5
 	// VersionMinor is for functionality in a backwards-compatible manner
-	VersionMinor = 33
```
TomSweeneyRedHat: I don't think we want this.  Should main be at 5.33.*-dev?

TomSweeneyRedHat:
Do you want the commits that changed the versions too?  The other commits look fine to me, but it looks like you need to  update this.

version/version.go
mtrmac: The net effect of this PR should be to keep main at 5.33.0-dev , as in https://github.com/mtrmac/image/blob/into-main/version/version.go (and the PR’s file change view is showing no changes).

The PR contents are all contents of the `release-5.32` branch, + a merge commit into `main` (it needs to be that way, the goal is to make the commit tagged 5.32.2 an ancestor of the `main` branch).

So, we do include the 5.32 version bumps as individual commits, but the merge commit undoes the `version.go` changes.

mtrmac:
> Do you want the commits that changed the versions too? 

We need the `5.32.2` commit to become an ancestor of `main`, and that brings up everything in the history of `5.32.2`, including the version bumps. As long as the merge undoes that, I think we are fine.

> it looks like you need to update this.

Rebased.

PR #2539, opened 2024-08-26
Support pulling Ollama [non-]OCI image

yeahdongcn:
Background:

Kubernetes 1.31 introduced a new feature: [Read-Only Volumes Based on OCI Artifacts](https://kubernetes.io/blog/2024/08/16/kubernetes-1-31-image-volume-source/). I believe this feature could be very useful for deploying a dedicated model alongside [Ollama](https://github.com/ollama/ollama) in Kubernetes.

Ollama has introduced several new media types (e.g. application/vnd.ollama.image.model) for storing GGUF models, system prompts, and more. Each layer is essentially a file and does not need to be untarred.

This PR adds a new field, `layerFilename`, to `addedLayerInfo`, and the overlay driver will handle the layer creation separately.

[A separate PR](https://github.com/containers/storage/pull/2075) for `containers/storage` will be submitted later.

Please see the following logs for instructions on how to mount the Ollama image as a volume:

```bash
# Copied from testdata and added mounts information
❯ cat container.json
{
  "metadata": {
    "name": "podsandbox-sleep"
  },
  "image": {
    "image": "registry.docker.com/ollama/ollama:latest"
  },
  "command": [
    "/bin/sleep",
    "6000"
  ],
  "args": [
    "6000"
  ],
  "working_dir": "/",
  "envs": [
    {
      "key": "PATH",
      "value": "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
    },
    {
      "key": "GLIBC_TUNABLES",
      "value": "glibc.pthread.rseq=0"
    }
  ],
  "annotations": {
    "pod": "podsandbox"
  },
  "log_path": "",
  "stdin": false,
  "stdin_once": false,
  "tty": false,
  "linux": {
    "security_context": {
      "namespace_options": {
        "pid": 1
      },
      "readonly_rootfs": false
    },
    "resources": {
      "cpu_period": 10000,
      "cpu_quota": 20000,
      "cpu_shares": 512,
      "oom_score_adj": 30,
      "memory_limit_in_bytes": 268435456
    }
  },
  "mounts": [
    {
      "host_path": "",
      "container_path": "/volume",
      "image": {
        "image": "registry.ollama.ai/library/tinyllama:latest"
      },
      "readonly": true
    }
  ]
}
# copied from testdata
❯ cat sandbox_config.json
{
        "metadata": {
                "name": "podsandbox1",
                "uid": "redhat-test-crio",
                "namespace": "redhat.test.crio",
                "attempt": 1
        },
        "hostname": "crictl_host",
        "log_directory": "",
        "dns_config": {
                "servers": [
                        "8.8.8.8"
                ]
        },
        "port_mappings": [],
        "resources": {
                "cpu": {
                        "limits": 3,
                        "requests": 2
                },
                "memory": {
                        "limits": 50000000,
                        "requests": 2000000
                }
        },
        "labels": {
                "group": "test"
        },
        "annotations": {
                "owner": "hmeng",
                "security.alpha.kubernetes.io/seccomp/pod": "unconfined",
                "com.example.test": "sandbox annotation"
        },
        "linux": {
                "cgroup_parent": "pod_123-456.slice",
                "security_context": {
                        "namespace_options": {
                                "network": 2,
                                "pid": 1,
                                "ipc": 0
                        },
                        "selinux_options": {
                                "user": "system_u",
                                "role": "system_r",
                                "type": "svirt_lxc_net_t",
                                "level": "s0:c4,c5"
                        }
                }
        }
}
❯ sudo crictl --timeout=200s --runtime-endpoint unix:///run/crio/crio.sock run ./container.json ./sandbox_config.json
INFO[0005] Pulling container image: registry.docker.com/ollama/ollama:latest 
INFO[0005] Pulling image registry.ollama.ai/library/tinyllama:latest to be mounted to container path: /volume 
7e437894449f6429799cc5ef236c4a4570a69e3769bf324bbf700045e383cae8
❯ sudo crictl --timeout=200s --runtime-endpoint unix:///run/crio/crio.sock ps
CONTAINER           IMAGE                                        CREATED             STATE               NAME                ATTEMPT             POD ID              POD
7e437894449f6       registry.docker.com/ollama/ollama:latest     8 seconds ago       Running             podsandbox-sleep    0                   4d1766fdf286b       unknown
❯ sudo crictl --timeout=200s --runtime-endpoint unix:///run/crio/crio.sock exec -it 7e437894449f6 bash
root@crictl_host:/# cd volume/
root@crictl_host:/volume# ls -l
total 622772
-rw-r--r-- 1 root root 637699456 Aug 26 08:32 model
-rw-r--r-- 1 root root        98 Aug 26 08:32 params
-rw-r--r-- 1 root root        31 Aug 26 08:32 system
-rw-r--r-- 1 root root        70 Aug 26 08:32 template
root@crictl_host:/volume# 
```

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -986,7 +986,8 @@ func (c *dockerClient) fetchManifest(ctx context.Context, ref dockerReference, t
 	if err != nil {
 		return nil, "", err
 	}
-	logrus.Debugf("Content-Type from manifest GET is %q", res.Header.Get("Content-Type"))
+	contentType := res.Header.Get("Content-Type")
+	logrus.Debugf("Content-Type from manifest GET is %q", contentType)
 	defer res.Body.Close()
 	if res.StatusCode != http.StatusOK {
 		return nil, "", fmt.Errorf("reading manifest %s in %s: %w", tagOrDigest, ref.ref.Name(), registryHTTPResponseToError(res))
@@ -996,7 +997,18 @@ func (c *dockerClient) fetchManifest(ctx context.Context, ref dockerReference, t
 	if err != nil {
 		return nil, "", err
 	}
-	return manblob, simplifyContentType(res.Header.Get("Content-Type")), nil
+	// Extra check for the content type in the manifest
+	if contentType == "text/plain" {
+		man := make(map[string]any)
+		if err := json.Unmarshal(manblob, &man); err != nil {
+			return nil, "", fmt.Errorf("decoding manifest %s in %s: %w", tagOrDigest, ref.ref.Name(), err)
+		}
+		if mediaType, ok := man["mediaType"]; ok {
+			contentType = mediaType.(string)
+			logrus.Debugf("Content-Type from manifest JSON is %q", contentType)
+		}
+	}
+	return manblob, simplifyContentType(contentType), nil
 }
 
 // getExternalBlob returns the reader of the first available blob URL from urls, which must not be empty.
--- a/manifest/manifest.go
+++ b/manifest/manifest.go
@@ -32,6 +32,27 @@ const (
 	DockerV2Schema2ForeignLayerMediaType = manifest.DockerV2Schema2ForeignLayerMediaType
 	// DockerV2Schema2ForeignLayerMediaType is the MIME type used for gzipped schema 2 foreign layers.
 	DockerV2Schema2ForeignLayerMediaTypeGzip = manifest.DockerV2Schema2ForeignLayerMediaTypeGzip
+
+	// OllamaImageLayerMediaTypePrefix is the prefix for all Ollama image layer media types.
+	OllamaImageLayerMediaTypePrefix = manifest.OllamaImageLayerMediaTypePrefix
+	// OllamaImageModelLayerMediaType is the media type used for model layers.
+	OllamaImageModelLayerMediaType = manifest.OllamaImageModelLayerMediaType
+	// OllamaImageAdapterLayerMediaType is the media type used for adapter layers.
+	OllamaImageAdapterLayerMediaType = manifest.OllamaImageAdapterLayerMediaType
+	// OllamaImageProjectorLayerMediaType is the media type used for projector layers.
+	OllamaImageProjectorLayerMediaType = manifest.OllamaImageProjectorLayerMediaType
+	// OllamaImagePromptLayerMediaType is the media type used for prompt layers.
+	OllamaImagePromptLayerMediaType = manifest.OllamaImagePromptLayerMediaType
+	//	OllamaImageTemplateLayerMediaType is the media type used for template layers.
+	OllamaImageTemplateLayerMediaType = manifest.OllamaImageTemplateLayerMediaType
+	// OllamaImageSystemLayerMediaType is the media type used for system layers.
+	OllamaImageSystemLayerMediaType = manifest.OllamaImageSystemLayerMediaType
+	// OllamaImageParamsLayerMediaType is the media type used for params layers.
+	OllamaImageParamsLayerMediaType = manifest.OllamaImageParamsLayerMediaType
+	// OllamaImageMessagesLayerMediaType is the media type used for messages layers.
+	OllamaImageMessagesLayerMediaType = manifest.OllamaImageMessagesLayerMediaType
+	// OllamaImageLicenseLayerMediaType is the media type used for license layers.
+	OllamaImageLicenseLayerMediaType = manifest.OllamaImageLicenseLayerMediaType
 )
 
 // NonImageArtifactError (detected via errors.As) is used when asking for an image-specific operation
@@ -42,6 +63,8 @@ type NonImageArtifactError = manifest.NonImageArtifactError
 func SupportedSchema2MediaType(m string) error {
 	switch m {
 	case DockerV2ListMediaType, DockerV2Schema1MediaType, DockerV2Schema1SignedMediaType, DockerV2Schema2ConfigMediaType, DockerV2Schema2ForeignLayerMediaType, DockerV2Schema2ForeignLayerMediaTypeGzip, DockerV2Schema2LayerMediaType, DockerV2Schema2MediaType, DockerV2SchemaLayerMediaTypeUncompressed:
+		return nil
+	case OllamaImageModelLayerMediaType, OllamaImageAdapterLayerMediaType, OllamaImageProjectorLayerMediaType, OllamaImagePromptLayerMediaType, OllamaImageTemplateLayerMediaType, OllamaImageSystemLayerMediaType, OllamaImageParamsLayerMediaType, OllamaImageMessagesLayerMediaType, OllamaImageLicenseLayerMediaType:
 		return nil
 	default:
 		return fmt.Errorf("unsupported docker v2s2 media type: %q", m)
--- a/storage/storage_dest.go
+++ b/storage/storage_dest.go
@@ -13,6 +13,7 @@ import (
 	"os"
 	"path/filepath"
 	"slices"
+	"strings"
 	"sync"
 	"sync/atomic"
 
@@ -125,8 +126,9 @@ type storageImageDestinationLockProtected struct {
 
 // addedLayerInfo records data about a layer to use in this image.
 type addedLayerInfo struct {
-	digest     digest.Digest // Mandatory, the digest of the layer.
-	emptyLayer bool          // The layer is an “empty”/“throwaway” one, and may or may not be physically represented in various transport / storage systems.  false if the manifest type does not have the concept.
+	digest        digest.Digest // Mandatory, the digest of the layer.
+	emptyLayer    bool          // The layer is an “empty”/“throwaway” one, and may or may not be physically represented in various transport / storage systems.  false if the manifest type does not have the concept.
+	layerFilename *string
 }
 
 // newImageDestination sets us up to write a new image, caching blobs in a temporary directory until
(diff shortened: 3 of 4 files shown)


docker/docker_client.go
```diff
@@ -996,7 +997,18 @@ func (c *dockerClient) fetchManifest(ctx context.Context, ref dockerReference, t
 	if err != nil {
 		return nil, "", err
 	}
-	return manblob, simplifyContentType(res.Header.Get("Content-Type")), nil
+	// Extra check for the content type in the manifest
+	if contentType == "text/plain" {
```
mtrmac, line 1001: - It is fundamentally unsound to override an externally-supplied MIME type of some data by parsing the data in another way and trusting its contents. The whole point of having the MIME type separate is that it directs how the data is to be interpreted. I’m unhappy about such special cases, and I’m especially unhappy about adding them in entirely new use cases where we could just choose not to do that. Why is this necessary here?
- The `docker` transport has no business interpreting image contents this way; that’s the responsibility of the generic code.
- … and the generic code _already has_ a heuristic fallback for `text/plain`; this would override/break that.

manifest/manifest.go
```diff
@@ -43,6 +64,8 @@ func SupportedSchema2MediaType(m string) error {
 	switch m {
 	case DockerV2ListMediaType, DockerV2Schema1MediaType, DockerV2Schema1SignedMediaType, DockerV2Schema2ConfigMediaType, DockerV2Schema2ForeignLayerMediaType, DockerV2Schema2ForeignLayerMediaTypeGzip, DockerV2Schema2LayerMediaType, DockerV2Schema2MediaType, DockerV2SchemaLayerMediaTypeUncompressed:
 		return nil
+	case OllamaImageModelLayerMediaType, OllamaImageAdapterLayerMediaType, OllamaImageProjectorLayerMediaType, OllamaImagePromptLayerMediaType, OllamaImageTemplateLayerMediaType, OllamaImageSystemLayerMediaType, OllamaImageParamsLayerMediaType, OllamaImageMessagesLayerMediaType, OllamaImageLicenseLayerMediaType:
+		return nil
```
mtrmac, line 68: I don’t know why we would add all of these to v2s2; that’s a ~frozen format.

And _if_ we did add them, we would need to adjust all of the rest of the v2s2 code to correctly interpret such data, at the very least to not parse it as an ordinary image.

Either way, it seems to me that this should either be completely opaque (a proper OCI artifact, maybe), or intentionally a completely new manifest / image format implementation.

storage/storage_dest.go
```diff
@@ -218,9 +220,18 @@ func (s *storageImageDestination) PutBlobWithOptions(ctx context.Context, stream
 		return info, nil
 	}
 
+	layerFilename := func() *string {
```
mtrmac, line 223: I don’t see why a nested function is necessary here.

storage/storage_dest.go
```diff
@@ -218,9 +220,18 @@ func (s *storageImageDestination) PutBlobWithOptions(ctx context.Context, stream
...
+		if strings.HasPrefix(blobinfo.MediaType, manifest.OllamaImageLayerMediaTypePrefix) {
+			filename := strings.TrimPrefix(blobinfo.MediaType, manifest.OllamaImageLayerMediaTypePrefix)
```
mtrmac, line 225: Choosing a “filename”, whatever the value is, this way makes no sense to me. What happens if there are two layers with the same MIME type in the image?

storage/storage_dest.go
```diff
@@ -218,9 +220,18 @@ func (s *storageImageDestination) PutBlobWithOptions(ctx context.Context, stream
 		return info, nil
 	}
```
mtrmac, line 221: There is an early return above; in that case `.filename` would not be set. 

OK, that’s somewhat theoretical. But also, nothing sets `.filename` on the `TryReusingBlobWithOptions` code path. That one is 100% a blocker.

storage/storage_dest.go
mtrmac, line 225: What would the `storageImageSource` counterpart look like?!

mtrmac (changes requested):
Thanks!

We need to have a proper design discussion about storing non-image artifacts, and what that means.

**Do not merge until that happens**.

Cc: @baude 

---

The “image volume” feature explicitly talks about _image_ volumes. I don’t see that arbitrarily extending it to also accept other content is obviously desirable.

And even if it were desirable at the Kubernetes level, that doesn’t _at all_ imply that the data should be stored in “images” with “layers” and OverlayFS and that it should be possible to `podman run` the thing.

---

We really need to have a design discussion about the storage _first_; but even if we did accept the proposed mechanism, the c/image implementation would have to be _much_ more extensive. This is nowhere near enough.

mtrmac:
Compare also https://github.com/containers/skopeo/issues/2395 (and, I think, https://github.com/ollama/ollama/issues/6510 ): “Ollama” itself is not _even remotely_ an OCI-compliant registry right now.

What is the ecosystem and community situation for this data format? Is this a multi-faceted interoperability effort, where all of these things are going to happily fall into place, or is this setting us up on an adversarial interoperability path?

Cc: @baude , another thing to have a very clear position on.

cgwalters:
Agree that we should only support compliant registries.

yeahdongcn:
> Compare also [containers/skopeo#2395](https://github.com/containers/skopeo/issues/2395) (and, I think, [ollama/ollama#6510](https://github.com/ollama/ollama/issues/6510) ): “Ollama” itself is not _even remotely_ an OCI-compliant registry right now.
> 
> What is the ecosystem and community situation for this data format? Is this a multi-faceted interoperability effort, where all of these things are going to happily fall into place, or is this setting us up on an adversarial interoperability path?
> 
> Cc: @baude , another thing to have a very clear position on.

Yes. I faced two issues pulling from `registry.ollama.ai`:

- https://registry.ollama.ai/v2/ returns 404
- The OCI image manifest response header returns an incorrect `Content-Type`

I thought decoding the manifest JSON might make MIME type determination more robust, but it clearly looks incorrect.

mtrmac:
Separate from the storage discussion happening in https://github.com/containers/storage/pull/2075, and _assuming_ the Ollama format and transport will not change (will it? I have no idea)
- We need to actually decide on what is the target format. Will there be some kind of actually-OCI artifact format to target, or is the Ollama format the one to support?
- If there will be an OCI _format_ (regardless of transport), it’s not clear that the conversion between the Ollama transport+format should exist in c/image, it might very reasonably be a small custom tool. There’s just not that much to share with the rest of c/image.
- If c/image should support the Ollama _transport_ natively (regardless of format), it might well make sense to make that a separate c/image `ImageTransport`. If it doesn’t really work as Docker, it can reasonably be a separate HTTP client.

yeahdongcn:
The issue I’m trying to solve is how to make OCI artifacts mountable as volumes in containers (the Ollama image is a practical example, but it uses some custom MIME types—can it still be defined as an OCI artifact?).

This blog post [Kubernetes 1.31: Image Volume Source](https://kubernetes.io/blog/2024/08/16/kubernetes-1-31-image-volume-source/) discusses some practical use cases that I find quite relevant. In particular, using OCI artifacts to distribute AI models (weights, parameters, system prompts, etc.) in enterprise AI inference services is crucial, especially for version management and signing.

The volume image used in the blog is `quay.io/crio/artifact:v1`, and inspecting it with Skopeo reveals that its layer MIME type is `application/vnd.oci.image.layer.v1.tar+gzip`, which is correctly recognized and parsed by `c/image`.

```bash
➜ ./bin/skopeo inspect docker://quay.io/crio/artifact:v1                   
{
    "Name": "quay.io/crio/artifact",
    "Digest": "sha256:52c997d8906f39c5b5550dd1f05fd785fb46bbb50eb055f9df61df6dc9e3a6cd",
    "RepoTags": [
        "v1"
    ],
    "Created": null,
    "DockerVersion": "",
    "Labels": null,
    "Architecture": "amd64",
    "Os": "linux",
    "Layers": [
        "sha256:4e80bd845ae8736c96d644d96142807d958f44505e973f6eb4145689e05e86d2"
    ],
    "LayersData": [
        {
            "MIMEType": "application/vnd.oci.image.layer.v1.tar+gzip",
            "Digest": "sha256:4e80bd845ae8736c96d644d96142807d958f44505e973f6eb4145689e05e86d2",
            "Size": 170,
            "Annotations": null
        }
    ],
    "Env": null
}
```

However, for Ollama images or other images with custom MIME types, should we consider adding support for these, or perhaps introducing a plugin mechanism to allow vendors to implement their own solutions?

```bash
➜ ./bin/skopeo inspect docker://registry.ollama.ai/library/tinyllama:latest
{
    "Name": "sh-harbor.mthreads.com/cloud-mirror/tinyllama",
    "Digest": "sha256:2644915ede352ea7bdfaff0bfac0be74c719d5d5202acb63a6fb095b52f394a4",
    "RepoTags": [
        "latest"
    ],
    "Created": "0001-01-01T00:00:00Z",
    "DockerVersion": "",
    "Labels": null,
    "Architecture": "amd64",
    "Os": "linux",
    "Layers": [
        "sha256:2af3b81862c6be03c769683af18efdadb2c33f60ff32ab6f83e42c043d6c7816",
        "sha256:af0ddbdaaa26f30d54d727f9dd944b76bdb926fdaf9a58f63f78c532f57c191f",
        "sha256:c8472cd9daed5e7c20aa53689e441e10620a002aacd58686aeac2cb188addb5c",
        "sha256:fa956ab37b8c21152f975a7fcdd095c4fee8754674b21d9b44d710435697a00d"
    ],
    "LayersData": [
        {
            "MIMEType": "application/vnd.ollama.image.model",
            "Digest": "sha256:2af3b81862c6be03c769683af18efdadb2c33f60ff32ab6f83e42c043d6c7816",
            "Size": 637699456,
            "Annotations": null
        },
        {
            "MIMEType": "application/vnd.ollama.image.template",
            "Digest": "sha256:af0ddbdaaa26f30d54d727f9dd944b76bdb926fdaf9a58f63f78c532f57c191f",
            "Size": 70,
            "Annotations": null
        },
        {
            "MIMEType": "application/vnd.ollama.image.system",
            "Digest": "sha256:c8472cd9daed5e7c20aa53689e441e10620a002aacd58686aeac2cb188addb5c",
            "Size": 31,
            "Annotations": null
        },
        {
            "MIMEType": "application/vnd.ollama.image.params",
            "Digest": "sha256:fa956ab37b8c21152f975a7fcdd095c4fee8754674b21d9b44d710435697a00d",
            "Size": 98,
            "Annotations": null
        }
    ],
    "Env": null
}
```

PR #2526, opened 2024-08-17
Add support for accepting multiple Rekor public keys

mtrmac:
Add `rekorPublicKeyPaths` and `rekorPublicKeyDatas` options to `policy.json`.

This is on top of unmerged #2524.

Note an outstanding design question on `…Datas`: https://github.com/containers/image/pull/2524#issuecomment-2294570712 .

<s>Currently not tested end-to-end in practice.</s>

PR #2541, opened 2024-08-27
Update stubs to match signature of actual methods

SpangleLabs:
#2526 has broken building v5.32.2 with the `containers_image_fulcio_stub` and `containers_image_rekor_stub` build tags, as the signatures for `verifyRekorFulcio()` and `VerifyRekorSET()` were updated without updating the corresponding stubs.

This PR updates the stubs to match

### DIFF
--- a/signature/fulcio_cert_stub.go
+++ b/signature/fulcio_cert_stub.go
@@ -20,7 +20,7 @@ func (f *fulcioTrustRoot) validate() error {
 	return errors.New("fulcio disabled at compile-time")
 }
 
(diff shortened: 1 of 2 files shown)


mtrmac:
Thanks!

First, https://github.com/containers/image/blob/main/CONTRIBUTING.md#sign-your-prs please, we can’t really even look at contributions with unclear copyright status.

SpangleLabs:
So sorry I missed this and then got pulled away to other things and forgot to update it!
I have updated the commit message now, sorry, and thank you! ^_^

mtrmac (approved):
Thanks!

I’m afraid we broke that build another way since…

mtrmac:
> I’m afraid we broke that build another way since…

https://github.com/containers/image/pull/2632

PR #2544, opened 2024-08-28
Add a workaround for null values in the tag list

mtrmac:
Motivated by https://github.com/containers/skopeo/issues/2409 .

<s>Untested so far. </s>

### DIFF
--- a/docker/docker_image.go
+++ b/docker/docker_image.go
@@ -91,6 +91,12 @@ func GetRepositoryTags(ctx context.Context, sys *types.SystemContext, ref types.
 		}
 		for _, tag := range tagsHolder.Tags {
 			if _, err := reference.WithTag(dr.ref, tag); err != nil { // Ensure the tag does not contain unexpected values
+				// Per https://github.com/containers/skopeo/issues/2409 , Sonatype Nexus 3.58, contrary
+				// to the spec, may include JSON null values in the list; and Go silently parses them as "".
+				if tag == "" {
+					logrus.Debugf("Ignoring invalid empty tag")
+					continue
+				}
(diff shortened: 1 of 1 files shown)


mtrmac:
The fix was confirmed  at https://github.com/containers/skopeo/issues/2409#issuecomment-2317078250 , ready for review.

cgwalters (approved):

docker/docker_image.go
```diff
@@ -91,6 +91,12 @@ func GetRepositoryTags(ctx context.Context, sys *types.SystemContext, ref types.
 		}
 		for _, tag := range tagsHolder.Tags {
 			if _, err := reference.WithTag(dr.ref, tag); err != nil { // Ensure the tag does not contain unexpected values
+				// Per https://github.com/containers/skopeo/issues/2409 , Sonatype Nexus 3.58, contrary
+				// to the spec, may include JSON null values in the list; and Go silently parses them as "".
+				if tag == "" {
+					logrus.Debugf("Ignoring invalid empty tag")
```
kwilczynski, line 97: Would it make sense to show which tag was it? Might help with some debugging efforts.

kwilczynski: Albeit, not sure if at all possible to do so.

mtrmac: We just get a `null` in an array; the server does not tell us why it included a `null`.

(We could, with some extra effort, get the raw `null` value instead of `""`, and be able to report whether we got one or the other, but I don’t think that is all that much more helpful. If we did want to include more details, I’d be tempted to just dump an escaped version of the response body as is, instead of trying to parse it.

Ultimately, this is all a debug-level log that users are not expected to see unless they are specially looking for it.)

PR #2546, opened 2024-08-30
[release-5.27] Backport

mtrmac:
… to avoid cross-branch regressions

PR #2453, opened 2024-06-13
[release-5.29] Don't abort listing tags when we encounter a digest

TomSweeneyRedHat:
Cherry Pick https://github.com/containers/image/pull/2440 then bump to v5.29.4 and finally to v5.29.5-dev

PR #2548, opened 2024-08-30
[release-5.24] Backport #2418 + #2453

mtrmac:
#2547 + #2546 exist on later branches to maintain continuity.

### DIFF
--- a/internal/iolimits/iolimits_test.go
+++ b/internal/iolimits/iolimits_test.go
@@ -10,6 +10,7 @@ import (
 )
 
 func TestReadAtMost(t *testing.T) {
+	rng := rand.New(rand.NewSource(0))
 	for _, c := range []struct {
 		input, limit  int
 		shouldSucceed bool
@@ -23,7 +24,7 @@ func TestReadAtMost(t *testing.T) {
 		{bytes.MinRead*5 + 1, bytes.MinRead * 5, false},
 	} {
 		input := make([]byte, c.input)
-		_, err := rand.Read(input)
+		_, err := rng.Read(input)
 		require.NoError(t, err)
 		result, err := ReadAtMost(bytes.NewReader(input), c.limit)
 		if c.shouldSucceed {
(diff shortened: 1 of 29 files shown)


rhatdan:
LGTM

internal/iolimits/iolimits_test.go
```diff
@@ -23,7 +24,7 @@ func TestReadAtMost(t *testing.T) {
 		{bytes.MinRead*5 + 1, bytes.MinRead * 5, false},
 	} {
 		input := make([]byte, c.input)
-		_, err := rand.Read(input)
+		_, err := rng.Read(input)
 		require.NoError(t, err)
 		result, err := ReadAtMost(bytes.NewReader(input), c.limit)
 		if c.shouldSucceed {
```
TomSweeneyRedHat, line 30: In https://github.com/containers/image/commit/521155ffe6fee9f06349f4a10948367a16aade98, the Makefile had a slight change at line 58, bumping lint from v1.49.0 to v1.51.0.  Is that not needed here?

TomSweeneyRedHat:
Just one question on the Makefile at line 58 for the Golint bump from 1.49 to 1.51.  Otherwise LGTM.

TYVM @mtrmac !  I'll create tags and releases in the morning.

internal/iolimits/iolimits_test.go
mtrmac: The CI environment we have is running Go 1.18., so the linter update to support Go 1.20 is not necessary.

(I’m not claiming that Go 1.18 is the _right_ Go version to require on this branch, I didn’t investigate that in detail — it’s just a version that happens to be used right now, based on `IMAGE_SUFFIX` in `.cirrus.yml` I think.)

---

It’s quite possible that various backported lint cleanups are not necessary either, they were just making my life easier when backporting (because my development system has `golangci-lint 1.60.2` installed).

mtrmac: … ah, perhaps a better way to actually answer the question:

I backported the lint warning cleanup because it was annoying me in local development with a recent linter. I didn’t backport the linter version bump itself because I thought that’s undesirable on a stable branch.

Arguably my local development setup which motivates me to backport unnecessary cleanups is a problem that should be fixed; and strictly speaking, maybe this lint backport should be reverted as unnecessary on the stable branch as well.

PR #2551, opened 2024-09-03
Refactor image deletion in OCI

mtrmac:
- Count the references directly into a single map, instead of allocating partial maps and copying data between them
- Restructure to minimize code duplication

Should not change behavior, overall. See individual commits and commit messages for details and an easier-to-review version.


rhatdan:
LGTM

rhatdan:
@Luap99 @giuseppe PTAL

rhatdan:
@kolyshkin PTAL

Luap99 (commented):
LGTM

Issue #1726, opened 2022-11-11
Multiple copies of the default signature policy path


dfr:
Recently, while working on porting podman system tests to FreeBSD, I noticed that the default location of the signature policy file is specified in multiple places:

- c/image/signature has builtinDefaultPolicyPath which is used by DefaultPolicy (but not exported)
- c/podman/pkg/trust has a function DefaultPolicyPath with a copy of the policy path in its implementation
- c/common/pkg/config exports DefaultSignaturePolicyPath as well as including its copy in the value returned by the function DefaultConfig

It would be nice to reduce that number down to one but it seems that the right way to do that is to somehow merge podman's pkg/trust with image's signature which is hard (see #1725).

PR #2562, opened 2024-09-06
Export BuiltinDefaultPolicyPath

rhatdan:
This field is duplicated in containers/common/pkg/config and is being vendored into bindings, just because it is not exported from containers/image.

### DIFF
--- a/signature/policy_config.go
+++ b/signature/policy_config.go
@@ -32,7 +32,7 @@ import (
 // systemDefaultPolicyPath is the policy path used for DefaultPolicy().
 // You can override this at build time with
 // -ldflags '-X github.com/containers/image/v5/signature.systemDefaultPolicyPath=$your_path'
-var systemDefaultPolicyPath = builtinDefaultPolicyPath
+var systemDefaultPolicyPath = BuiltinDefaultPolicyPath
 
 // userPolicyFile is the path to the per user policy path.
 var userPolicyFile = filepath.FromSlash(".config/containers/policy.json")
--- a/signature/policy_paths_common.go
+++ b/signature/policy_paths_common.go
@@ -3,6 +3,6 @@
 
 package signature
 
-// builtinDefaultPolicyPath is the policy path used for DefaultPolicy().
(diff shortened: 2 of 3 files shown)


rhatdan:
@mtrmac @Luap99 PTAL

Luap99 (commented):
LGTM

mtrmac (changes requested):
NAK; if the goal is to handle https://github.com/containers/image/issues/1726 , this is going in the wrong direction, motivating users _even more_ to hard-code specifics of the logic (in this case, exporting the system-wide path motivates users to reimplement the per-user path).

If c/common intends to use the c/image default, it should not be setting the path in `SystemContext` at all, and the whole need for this goes away. (That can’t _immediately_ happen because the two implementations have already diverged. Also _maybe_ c/image needs some special knowledge of rootless operations .)

---

There is `pkg/trust`, which _writes_ to the config files, and that one really needs to know a path, not a policy accessor. So c/image should probably export _something_ — but it wouldn’t be `systemDefaultPolicyPath`; it would be `defaultPolicyPath` or maybe a separate “system-wide primary policy config path” / “per-user primary config path”.

rhatdan:
Well how do we move forward on that then?

mtrmac:
I don’t immediately know; the way the code has already diverged, maybe there is some neat elegant solution which is compatible enough, maybe we will have to live with a gross hack forever. Someone needs to figure that out.

---

For the purpose of trimming https://github.com/containers/podman/pull/23857#issuecomment-2328356840 , the bindings should not be including the whole implementation in `pkg/trust` in the first place; AFAICS they just need the `trust.Policy` data type. That can either be split into a subpackage; or the `domain/entities/types` subpackage could define a separate RPC-only type, separate from the implementation type; or maybe `pkg/trust` should be importing the type from `domain/entities/types` — I don’t know what is the general Podman approach to RPC-only types duplicated with implementation types, vs. sharing the two, and where the definitions live.

jankaluza:

PR #2567, opened 2024-09-10
WIP: oci/layout API extensions

mtrmac:
Includes @sourceIndex code from https://github.com/containers/image/pull/1677 .

Do not merge:
- At least the transport part must have tests (probably mostly from 1677).
- Do we want to add this to `oci/archive` at the same time? Structure `oci/internal` accordingly either way.

### DIFF
--- a/oci/layout/oci_dest.go
+++ b/oci/layout/oci_dest.go
@@ -305,6 +305,33 @@ func (d *ociImageDestination) Commit(context.Context, types.UnparsedImage) error
 	return os.WriteFile(d.ref.indexPath(), indexJSON, 0644)
 }
 
+// PutBlobFromLocalFile arranges the data from path to be used as blob with digest.
+// It computes, and returns, the digest and size of the used file.
+//
+// This function can be used instead of dest.PutBlob() where the ImageDestination requires PutBlob() to be called.
+func PutBlobFromLocalFile(ctx context.Context, dest types.ImageDestination, file string) (digest.Digest, int64, error) {
+	d, ok := dest.(*ociImageDestination)
+	if !ok {
+		return "", -1, errors.New("internal error: PutBlobFromLocalFile called with a non-oci: destination")
+	}
+
+	reader, err := os.Open(file)
+	if err != nil {
+		return "", -1, fmt.Errorf("opening %q: %w", file, err)
+	}
+	defer reader.Close()
+
+	// This makes a full copy; instead, if possible, we could only digest the file and reflink (hard link?)
+	uploaded, err := d.PutBlobWithOptions(ctx, reader, types.BlobInfo{
+		Digest: "",
+		Size:   -1,
+	}, private.PutBlobOptions{})
+	if err != nil {
+		return "", -1, err
+	}
+	return uploaded.Digest, uploaded.Size, nil
+}
+
 func ensureDirectoryExists(path string) error {
 	if err := fileutils.Exists(path); err != nil && errors.Is(err, fs.ErrNotExist) {
 		if err := os.MkdirAll(path, 0755); err != nil {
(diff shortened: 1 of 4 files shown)


mtrmac:
Added `PutBlobFromLocalFile`, just to sketch out what an API benefiting from reflinks and the like could look like.

oci/layout/oci_dest.go
```diff
@@ -302,6 +305,33 @@ func (d *ociImageDestination) Commit(context.Context, types.UnparsedImage) error
 	return os.WriteFile(d.ref.indexPath(), indexJSON, 0644)
 }
 
+// PutBlobFromLocalFile arranges the data from path to be used as blob with digest.
+// It computes, and returns, the digest and size of the used file.
+//
+// This function can be used instead of dest.PutBlob() where the ImageDestination requires PutBlob() to be called.
+func PutBlobFromLocalFile(ctx context.Context, dest types.ImageDestination, file string) (digest.Digest, int64, error) {
```
cgwalters, line 312: Let's make this more general by allowing the caller to pass a file descriptor instead, with a sugar wrapper method that opens for them if need be?

Or is this function *already* sugar basically and the functionality would already be exposed by `PutBlobWithOptions` as is?

mtrmac: No, a file descriptor is a good idea — assuming Go has a good enough platform abstraction for it.

mtrmac: On second thought, accepting a file descriptor brings up various questions, like “what does it mean if the file descriptor is positioned 1 GB into the file?”.

The use case we have now is to accept a path on the CLI, so a path-based interface is exactly what we need and avoids having to think about these things.

We can always add another function later.

oci/layout/oci_dest.go
```diff
@@ -302,6 +305,33 @@ func (d *ociImageDestination) Commit(context.Context, types.UnparsedImage) error
...
+		return "", -1, errors.New("internal error: PutBlobFromLocalFile called with a non-oci: destination")
+	}
+
+	reader, err := os.Open(file)
+	if err != nil {
+		return "", -1, fmt.Errorf("opening %q: %w", file, err)
+	}
+	defer reader.Close()
+
+	// This makes a full copy; instead, if possible, we could only digest the file and reflink (hard link?)
```
mtrmac, line 324: Note to self: `c/storage/drivers/copy.CopyRegular…`

baude:
been using this code for a while now in developing artifact function and it seems spot on for this ...

baude:
i accidentally deleted my earlier comment, but I have been using the layout.List portion of this PR and it is working nicely. Will test the latest addition here soon.

oci/layout/oci_dest.go
vrothberg, line 312: I think we need a means to configure the mime type of the blob to serve artifacts.  Going forward, I anticipate users to create their custom layer types for LLM etc.

mtrmac: The way the `ImageDestination` API works, `PutBlob` doesn’t typically need to worry about the MIME type: the MIME type comes via `PutManifest`.

While it is true that the `PutBlobWithOptions` gets a full `types.BlobInfo`, and https://github.com/containers/image/blob/0c42f2fe704c9eb84a0b91f4f0363c9919a677db/copy/single.go#L852 forwards the MIME type value into the layer copy code, there are things like https://github.com/containers/image/blob/0c42f2fe704c9eb84a0b91f4f0363c9919a677db/copy/compression.go#L168 just dropping that; so, transports can’t rely on a MIME type being present (~instead, we have `PutBlobOptions.IsConfig).

Also, _specifically in the OCI transport_, the `PutBlob*` code really doesn’t care about the MIME type. It is going to create a sha256-named file, and as far as the layout definition is concerned, it’s perfectly valid to refer to a single blob in two different manifests using two different MIME types. So, for this _OCI-transport-specific_ function, I don’t think adding a MIME type is right.

mtrmac:
Closing in favor of #2633.

PR #2576, opened 2024-09-20
 copy: Don't print "skipped: 0.0b = 0.00%"

cgwalters:
copy: Add unit tests for progress output

Prep for more work.

---

copy: Don't print "skipped: 0.0b = 0.00%"

I think this is just visually noisy. If we didn't do a partial
fetch then let's just not say anything about it.

---

### DIFF
--- a/copy/progress_bars_test.go
+++ b/copy/progress_bars_test.go
@@ -20,5 +20,5 @@ func TestCustomPartialBlobDecorFunc(t *testing.T) {
 	// Almost complete, but no reuse
 	s.Current = int64(float64(s.Total) * 0.95)
 	s.Refill = 0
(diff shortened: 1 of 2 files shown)


copy/progress_bars_test.go
```diff
@@ -0,0 +1,24 @@
+package copy
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/vbauerster/mpb/v8/decor"
+)
+
+func TestCustomPartialBlobDecorFunc(t *testing.T) {
```
mtrmac, line 10: (This works just fine. In Go, it is idiomatic to have table-driven tests, compare e.g. `TestUpdatedBlobInfoFromUpload`.)

mtrmac (approved):
Thanks! The tests make this really nice to review.

Issue #2027, opened 2023-07-05
Request for Sigstore signature verification enhancement and flexibility in cosign verify.


zhaoyonghe:
Currently, we are able to verify container images with sigstore signatures using public key/Fulcio/Rekor, as described [here](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/building_running_and_managing_containers/assembly_signing-container-images_building-running-and-managing-containers#proc_verifying-sigstore-image-signatures-using-a-public-key_assembly_signing-container-images). However, there are additional verifications supported by cosign, such as verifying signatures using non-Fulcio roots, as demonstrated in this [pull request](https://github.com/sigstore/cosign/pull/2845).

An example of the cosign command for verifying signatures using non-Fulcio roots is:

```
SIGSTORE_ROOT_FILE="certificate_bundle.pem" cosign verify \
  --certificate-identity-regexp "identity-.*" \
  --certificate-oidc-issuer-regexp "non-Fulcio-issuer.*" \
  --timestamp-certificate-chain tsa_ca_chain.pem \
  "${image}"
```
Is it possible to add support for this functionality? Moreover, could we take a step further and match fields in /etc/containers/policy.json with the cosign verify parameters to enable the execution of all forms of cosign verify?

/cc @mtrmac
Thank you for your attention to this matter.

mtrmac:
Thanks for reaching out.

We are, for the foreseeable future, very unlikely to include _all_ of cosign features (that would essentially require including much of cosign codebase, and c/image is used in size-constrained tools like Podman).

Individual features can be added (and I expect more Fulcio-matching features in particular), but that would probably be on a feature-by-feature basis, if they are necessary and supportable.

As some random few comments
- c/image does already support (in fact, it _requires_) providing Fulcio roots directly as a CA certificate. There is no code equivalent to Cosign’s `fulcio.GetRoots()`.
- `--certificate-oidc-issuer-regexp ".*"` seems to me to be such a bypass of the Fulcio design that adding the regexp option only for the purpose of it being used _that way_ seems quite ill-advised to me.

lkatalin:
It sounds like this request needs to be broken down into specific, individual feature requests that can be evaluated and implemented:

- [ ] Functionality for passing in signing certificate identity on the CLI during verification with ` --certificate-identity-regexp`
- [ ] Functionality for passing in OIDC issuer identity on the CLI during verification with ` --certificate-identity-regexp`
[...]

PR #2579, opened 2024-09-20
policy.json BYOPKI signature verification API

QiWang19:
[OCPNODE-2338](https://issues.redhat.com/browse/OCPNODE-2338)


QiWang19:
Hi @mtrmac , I’ve created this draft PR for the policy.json API change, could you take a look and provide your initial thoughts on the structure of the API update?
I just want to make sure I’m on the right path before proceeding further. My plan is to merge both the API and code support in a single PR. What do you think?

signature/policy_types.go
```diff
@@ -167,6 +170,22 @@ type prSigstoreSignedFulcio struct {
 	SubjectEmail string `json:"subjectEmail,omitempty"`
 }
 
+// prSigstoreSignedPKI contains non-fulcio certificate PKI configuration options for prSigstoreSigned
+type prSigstoreSignedPKI struct {
```
mtrmac, line 174: There should be _some_ public type callers can use to refer to the value. Following the other examples here, probably an interface with a private method, and public functions to construct it (`NewPRSigstoreSignedPKI` etc.) using the “functional option” pattern.

signature/policy_types.go
```diff
@@ -167,6 +170,22 @@ type prSigstoreSignedFulcio struct {
...
+	// CARootsPath a path to a file containing accepted CA root certificates, in PEM format. Exactly one of CARootsPath and CARootsData must be specified.
+	CARootsPath string `json:"caRootsPath"`
+	// CARootsData contains accepted CA root certificates in PEM format, all of that base64-encoded. Exactly one of CARootsPath and CARootsData must be specified.
+	CARootsData []byte `json:"caRootsData"`
+	// CAIntermediatesPath a path to a file containing accepted CA intermediate certificates, in PEM format. Only one of CAIntermediatesPath or CAIntermediatesData can be specified, not both.
+	CAIntermediatesPath string `json:"caIntermediatesPath"`
+	// CAIntermediatesData contains accepted CA intermediate certificates in PEM format, all of that base64-encoded. Only one of CAIntermediatesPath or CAIntermediatesData can be specified, not both.
+	CAIntermediatesData []byte `json:"caIntermediatesData"`
+	// SubjectEmail specifies the expected email address imposed on the subject to which the certificate was issued. Exactly one of SubjectEmail and Hostname must be specified.
```
mtrmac, line 183: (Nit: A blank line above would separate the two kinds of fields.)

signature/policy_types.go
```diff
@@ -167,6 +170,22 @@ type prSigstoreSignedFulcio struct {
...
+	SubjectEmail string `json:"subjectEmail"`
+	// Hostname specifies the expected hostname imposed on the subject to which the certificate was issued. Exactly one of SubjectEmail and Hostname must be specified.
+	Hostname string `json:"hostname"`
```
mtrmac, line 186:
```suggestion
	SubjectHostname string `json:"subjectHostname"`
```

mtrmac (commented):
Thanks! Looks good overall.

Earlier discussion https://github.com/openshift/enhancements/pull/1658 .

Doing both the API and implementation in one PR works for me, having the API available but broken is not really helping anything.

QiWang19:
@mtrmac  Could you please take a look? It's ready for review. I am unsure about how I’m handling the intermediate certificates.

signature/pki_cert.go
```diff
@@ -0,0 +1,100 @@
...
+
+	if len(pkiTrustRoot.caIntermediateCertificates) > 0 {
+		trustedIntermediates, err := cryptoutils.UnmarshalCertificatesFromPEM(pkiTrustRoot.caIntermediateCertificates)
+		if err != nil {
+			return nil, internal.NewInvalidSignatureError(fmt.Sprintf("loading trusted intermediate certificates: %v", err))
+		}
+		for _, trustedIntermediateCert := range trustedIntermediates {
+			untrustedIntermediatePool.AddCert(trustedIntermediateCert)
+		}
+	}
```
mtrmac, line 81: I think this can all happen in `prSigstoreSignedPKI.prepareTrustRoot`:
- With incorrect configurations we don’t even start parsing the high-risk signature itself.
- It’s a long-term goal to store the trust root data in `PolicyContext` instead of doing that work on every single signature verification. (That’s _certainly_ not this PR, just that structuring the code to make that possible would be nice.)

signature/pki_cert.go
```diff
@@ -0,0 +1,100 @@
...
+		KeyUsages:     []x509.ExtKeyUsage{x509.ExtKeyUsageCodeSigning},
+	}); err != nil {
+		return nil, internal.NewInvalidSignatureError(fmt.Sprintf("veryfing leaf certificate failed: %v", err))
+	}
+
+	return untrustedCertificate.PublicKey, nil
+}
+
+// isIntermediateCA checks if the certificate is an intermediate CA
+func isIntermediateCA(cert *x509.Certificate) bool {
```
mtrmac, line 95: It’s not immediately clear to me that this is necessary, I’d expect the `crypto/x509` code to validate the intermediates are usable. Is there a specific reason to filter them this way? If so, it would be nice to document that.

signature/policy_eval_sigstore.go
```diff
@@ -193,8 +202,12 @@ func (pr *prSigstoreSigned) isSignatureAccepted(ctx context.Context, image priva
...
 	case trustRoot.publicKeys != nil && trustRoot.fulcio != nil: // newPRSigstoreSigned rejects such combinations.
 		return sarRejected, errors.New("Internal inconsistency: Both a public key and Fulcio CA specified")
-	case trustRoot.publicKeys == nil && trustRoot.fulcio == nil: // newPRSigstoreSigned rejects such combinations.
-		return sarRejected, errors.New("Internal inconsistency: Neither a public key nor a Fulcio CA specified")
+	case trustRoot.publicKeys != nil && trustRoot.pki != nil: // newPRSigstoreSigned rejects such combinations.
+		return sarRejected, errors.New("Internal inconsistency: Both a public key and PKI specified")
+	case trustRoot.fulcio != nil && trustRoot.pki != nil: // newPRSigstoreSigned rejects such combinations.
+		return sarRejected, errors.New("Internal inconsistency: Both Fulcio CA and PKI specified")
+	case trustRoot.publicKeys == nil && trustRoot.fulcio == nil && trustRoot.pki == nil: // newPRSigstoreSigned rejects such combinations.
+		return sarRejected, errors.New("Internal inconsistency: A public key, Fulcio, or PKI must be specified.")
```
mtrmac, line 210: This O(N^2) item list is getting large — maybe switching this check counting `keySources == 1` would be easier to follow.

signature/policy_eval_sigstore.go
```diff
@@ -344,3 +372,43 @@ func (pr *prSigstoreSigned) isRunningImageAllowed(ctx context.Context, image pri
...
+	caIntermediatesCertPEMs, err := loadBytesFromConfigSources(configBytesSources{
+		inconsistencyErrorMessage: `Internal inconsistency: both "caIntermediatesPath" and "caIntermediatesData" specified`,
+		path:                      p.CAIntermediatesPath,
+		data:                      p.CAIntermediatesData,
+	})
+	if err != nil {
+		return nil, err
+	}
+	if caIntermediatesCertPEMs != nil {
+		pki.caIntermediateCertificates = caIntermediatesCertPEMs[0]
```
mtrmac, line 408: A sanity check that `len() == 1` would be nice here.

mtrmac (commented):
Just a few-minute first pass — this looks very good. I didn’t read this carefully, and I skipped the tests completely, for now.

Combining the policy-provided and signature-provided intermediate certificates makes sense to me, but I’ll at least check what Cosign does.

mtrmac:
Cc: @Honny1 as well — c/image/signature is a somewhat separate part of the codebase, and as security-critical with a lot of paranoia. Worth understanding the structure and idioms.

signature/pki_cert.go
QiWang19: I added this because I’m not sure if multiple root certificates will be included in the `untrustedIntermediateChainBytes` from `dev.sigstore.cosign/chain`. So, I’ve been filtering out intermediates from the data. Do you think this is necessary?

mtrmac: By the principle of the thing, and reading `crypto/x509.Certificate.buildChains`, an intermediate entry is only trusted if it is signed by some other CA. So including any extra certificates should not hurt — after all, an attacker can include arbitrary intermediate certificates in the signature.

(Also — probably not for sigstore, but in general — there are transition scenarios, where a new CA might get its to-be-root certificate signed by some other CA; this to-be-root certificate can then be accepted as an intermediate.  That’s a valid situation (again, in general, probably not now for sigstore, and not for the BYOPKI users).

The certificate/issuer link is nowadays primarily happening via `SubjectKeyId`/`AuthorityKeyId` fields, so it seems to me it is _possible_ to for an intermediate to have `cert.Issuer == cert.Subject`.)

QiWang19:
@mtrmac Do you think the PKI code needs a file like fulcio_cert_stub.go and should include build tags? I'm not fully clear on the purpose of this compile configuration.

mtrmac:
The `signature/*_stub.go` files were added in https://github.com/containers/image/pull/2180 , to avoid dependencies on sigstore Go modules not packaged (at the time??) in Debian.

Here, the only similar dependency seems to be `github.com/sigstore/sigstore/pkg/cryptoutils`, which is also required by no-stubbed `policy_eval_sigstore.go`; so this shouldn’t need a stub.

QiWang19:
> Combining the policy-provided and signature-provided intermediate certificates makes sense to me, but I’ll at least check what Cosign does.

I checked some cosign code, maybe I got lost in the code path, I did see it process the signature-provided intermediate certificates :thinking: 
- https://github.com/sigstore/cosign/blob/ad478088320a3c04a96b3c183bbde2205fff7bbb/cmd/cosign/cli/verify/verify.go#L248
- https://github.com/sigstore/cosign/blob/main/pkg/cosign/verify.go#L223
- https://github.com/sigstore/cosign/blob/ad478088320a3c04a96b3c183bbde2205fff7bbb/pkg/cosign/verify.go#L1332

QiWang19:
@mtrmac could you review?

signature/policy_eval_sigstore.go
```diff
@@ -344,3 +378,52 @@ func (pr *prSigstoreSigned) isRunningImageAllowed(ctx context.Context, image pri
...
+	}
+	caIntermediatesCertPEMs, err := loadBytesFromConfigSources(configBytesSources{
+		inconsistencyErrorMessage: `Internal inconsistency: both "caIntermediatesPath" and "caIntermediatesData" specified`,
+		path:                      p.CAIntermediatesPath,
+		data:                      p.CAIntermediatesData,
+	})
+	if err != nil {
+		return nil, err
+	}
+	if len(caIntermediatesCertPEMs) == 1 {
```
mtrmac, line 413: If `len()> 1`, the previous version silently ignored entries after [0], this one silently ignores everything.

I’d prefer
```go
if caIntermediatesCertPEMs != nil {
   if len(caIntermediatesCertPEMs) != 1 { /* fail with “internal inconsistency: … */ }
}
```

signature/pki_cert.go
```diff
@@ -0,0 +1,83 @@
...
+
+	if pkiTrustRoot.subjectHostname != "" {
+		if err = untrustedCertificate.VerifyHostname(pkiTrustRoot.subjectHostname); err != nil {
+			return nil, internal.NewInvalidSignatureError(fmt.Sprintf("Unexpected subject hostname: %v", err))
+		}
+	}
+
+	untrustedIntermediatePool := x509.NewCertPool()
+	if pkiTrustRoot.caIntermediatesCertificates != nil {
+		untrustedIntermediatePool = pkiTrustRoot.caIntermediatesCertificates
```
mtrmac, line 60: AFAICS this … (A)

signature/pki_cert.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+	if len(untrustedIntermediateChainBytes) > 0 {
+		untrustedIntermediateChain, err := cryptoutils.UnmarshalCertificatesFromPEM(untrustedIntermediateChainBytes)
+		if err != nil {
+			return nil, internal.NewInvalidSignatureError(fmt.Sprintf("loading certificate chain: %v", err))
+		}
+		if len(untrustedIntermediateChain) > 1 {
+			for _, untrustedIntermediateCert := range untrustedIntermediateChain {
+				untrustedIntermediatePool.AddCert(untrustedIntermediateCert)
```
mtrmac, line 69: (A) … and this, together, means, that `pkiTrustRoot.caIntermediatesCertificates` will grow over time as validations happen.

Right now that doesn’t really matter, but see the pending `// FIXME: move this to per-context initialization` — the idea is that “the trust root” is computed once and ~immutable, maintained over the lifetime of a `PolicyContext` for possibly many signature validations.

QiWang19: I also added the same comment before these lines of code.

docs/containers-policy.json.5.md
```diff
@@ -329,14 +329,22 @@ This requirement requires an image to be signed using a sigstore signature with
         "oidcIssuer": "https://expected.OIDC.issuer/",
         "subjectEmail", "expected-signing-user@example.com",
     },
+    "pki": {
+		"caRootsPath": "/path/to/local/CARoots/file",
```
saschagrunert, line 333: nit: mixed indent (tabs vs spaces).

docs/containers-policy.json.5.md
```diff
@@ -329,14 +329,22 @@ This requirement requires an image to be signed using a sigstore signature with
...
+        "caRootsData": "base64-encoded-CARoots-data",
+        "caIntermediatesPath": "/path/to/local/CAIntermediates/file",
+        "caIntermediatesData": "base64-encoded-CAIntermediates-data",
+        "subjectHostname": "expected-signing-hostname.example.com",
+		"subjectEmail": "expected-signing-user@example.com"
```
saschagrunert, line 338: same here

saschagrunert (approved):
Just two nits, otherwise LGTM

signature/pki_cert.go
mtrmac: It’s not just the comment, it’s that the “trust root” is supposed to represent a trust root, and now it accumulates unrelated untrusted data. That’s confusing and it shouldn’t be happening.

(I didn’t check whether it is possible to easily clone a certificate pool. If it weren’t possible, I’d prefer going back to the situation where the trust root contained unparsed file contents to the current state.)

docs/containers-policy.json.5.md
```diff
@@ -407,6 +420,19 @@ selectively allow individual transports and scopes as desired.
...
                 }
             ],
+            /* A Sigstore-signed repository using a certificate generated by the Bring Your Own Public Key Infrastructure (BYOPKI).*/
+            "hostname:5000/myns/sigstore-signed-byopki": [
+                {
+                    "type": "sigstoreSigned",
+                    "pki": {
+                        "caRootsPath": "/path/to/pki_roots_crt.pem",
+                        "caIntermediatesPath": "/path/to/pki_intermediates_crt.pem",
+                        "oidcIssuer": "https://github.com/login/oauth",
```
mtrmac, line 430: This field isn’t accepted.

signature/pki_cert.go
QiWang19: There is a clone function https://pkg.go.dev/crypto/x509#CertPool.Clone
I updated to clone the trusted certs pool to a new pool and append untrusted certs to it.  Could you take a look?

saschagrunert:
@containers/image-maintainers can we get this merged?

QiWang19:
Rebased with the base branch.

dmitris:
> [OCPNODE-2338](https://issues.redhat.com/browse/OCPNODE-2338)

now tracked in [RUN-2436](https://issues.redhat.com/browse/RUN-2436)

dmitris:
> Rebased with the base branch.

@QiWang19 could you rebase it again? (Maybe this would help to get this merged 😄 )

signature/fixtures/pki_roots_crt.pem
```diff
@@ -0,0 +1,33 @@
+-----BEGIN CERTIFICATE-----
```
mtrmac, line 1: These are all freshly-generated, aren’t they?

Seeing `C=ES, L=Valencia, O=IT, OU=Security, CN=Linuxera Root Certificate Authority` in there was a bit surprising.

\*shrug\* the names don’t really matter, probably not worth re-doing all of that.

signature/pki_cert.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	switch len(untrustedLeafCerts) {
+	case 0:
+		return nil, internal.NewInvalidSignatureError("no certificate found in signature certificate data")
+	case 1:
+		break // OK
+	default:
+		return nil, internal.NewInvalidSignatureError("unexpected multiple certificates present in signature certificate data")
+	}
+	untrustedCertificate := untrustedLeafCerts[0]
```
mtrmac, line 42: The failure cases should have tests, similar to the ones in `fulcio_cert.go`.

Alternatively (preferably?) extract the code+test in `fulcio_cert.go` into a separate function, and share the code.

signature/pki_cert.go
```diff
@@ -0,0 +1,85 @@
...
+				untrustedCertificate.EmailAddresses))
+		}
+	}
+
+	if pkiTrustRoot.subjectHostname != "" {
+		if err = untrustedCertificate.VerifyHostname(pkiTrustRoot.subjectHostname); err != nil {
+			return nil, internal.NewInvalidSignatureError(fmt.Sprintf("Unexpected subject hostname: %v", err))
+		}
+	}
+
```
mtrmac, line 57: We should establish trust in the certificate contents (that it is correctly signed) _first_, only then it makes sense to read any of the data fields.

So, (roots are done); intermediates first; individual certificate next; validate certificate against root+intermediates; then read contents.

mtrmac (commented):
Looks very good overall, ~1.5 comment on the substance of the code; almost all are about testing or documentation.

mtrmac:
> > Combining the policy-provided and signature-provided intermediate certificates makes sense to me, but I’ll at least check what Cosign does.
> 
> I checked some cosign code, maybe I got lost in the code path, I did see it process the signature-provided intermediate certificates 🤔
> 
> * https://github.com/sigstore/cosign/blob/ad478088320a3c04a96b3c183bbde2205fff7bbb/cmd/cosign/cli/verify/verify.go#L248

I think this is for `cosign verify` with an explicit `--certificate` option. Users _usually_ don’t do that.

I think the relevant code path for reading signatures from the image is https://github.com/sigstore/cosign/blob/737c83c71cd42ba8db843f84beadcd59b3aa6e0a/pkg/cosign/verify.go#L822-L841 .

It is convoluted, but, effectively, if the configuration contains intermediates, the ones from the signature are ignored.

I think that behavior is unintuitive — it means that adding more configuration (possibly 100% redundant configuration) can turn accepted signatures into rejected signatures.

That code comes from https://github.com/sigstore/cosign/pull/1804 , which effectively proposes to use TUF to stop accepting intermediate certificates _without revoking them in any way_. I don’t understand the use of intermediate certificates in this manner; it seems to me that shipping these intermediates “as trusted roots” would work well, then TUF could be used to ship a configuration update where the compromised certificate is simply dropped.

For comparison, Go’s TLS implementation takes certificates _only_ from the protocol data, it does not allow users to provide intermediates at all.

For BYOPKI, I wouldn’t expect TUF to be used. If I’m not missing something subtle about this, let’s leave the behavior as is.

signature/pki_cert_test.go
```diff
@@ -0,0 +1,100 @@
...
+	// Success
+	pk, err := verifyPKI(&pkiTrustRoot{
+		caRootsCertificates:         caRootsCertificates,
+		caIntermediatesCertificates: caIntermediatesCertificates,
+		subjectEmail:                "qiwan@redhat.com",
+		subjectHostname:             "myhost.example.com",
+	}, certBytes, chainBytes)
+	require.NoError(t, err)
+	assertPublicKeyMatchesCert(t, certBytes, pk)
+
```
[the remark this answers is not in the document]
QiWang19: > Failure with intermediates with neither signature nor config

Does this mean the following test should pass? right now the `verifyPKI` does not return err on this test case
```
	// Failure intermediates with neither signature nor config
	pk, err = verifyPKI(&pkiTrustRoot{
		caRootsCertificates: caIntermediateCertificates,
	}, certBytes, []byte{})
	require.Error(t, err)
	assert.Nil(t, pk)
```

signature/policy_config_sigstore.go
```diff
@@ -71,6 +71,17 @@ func PRSigstoreSignedWithFulcio(fulcio PRSigstoreSignedFulcio) PRSigstoreSignedO
 	}
 }
 
+// PRSigstoreSignedWithPKI specifies a value for the "pki" field when calling NewPRSigstoreSigned.
+func PRSigstoreSignedWithPKI(p PRSigstoreSignedPKI) PRSigstoreSignedOption {
```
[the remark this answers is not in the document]
QiWang19: Should `newPRSigstoreSigned`  return InvalidPolicyFormatError if both pki and rekor specified?

signature/fixtures/pki_roots_crt.pem
QiWang19: I haven't addressed this now, but can do this in a follow-up PR.

signature/pki_cert_test.go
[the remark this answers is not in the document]
mtrmac: I meant: Given a leaf certificate, signed by an intermediate certificate, signed by a CA:
- test that `verifyPKI(&pkiTrustRoot{caRoots…: caRoots}), leafCert, {})` fails (no trust path from CA to `certBytes`)
- test that `verifyPKI(&pkiTrustRoot{caRoots…: caRoots, caIntermediate…: caIntermediate…}), leafCert, {})` succeeds (intermediates in config)
- test that `verifyPKI(&pkiTrustRoot{caRoots…: caRoots}), leafCert, caIntermediate…)` succeeds (intermediates in signature)

signature/policy_config_sigstore.go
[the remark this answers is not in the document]
mtrmac: That’s a great catch: I think, ideally, Rekor should be supported for PKI (and require the leaf certificate to be in Rekor, per `VerifyRekorSET`).

Alternatively, we can reject such configurations for now as unsupported (both in the config `newPRSigstoreSigned` and in the evaluation in  `sigstoreSignedTrustRoot`), and maybe add that in the future.

I don’t feel strongly about doing one or the other (but maybe the users want this to work?), and both are probably comparable amount of work.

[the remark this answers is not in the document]
mtrmac: … if we support Rekor, would we validate the certificate expiration against current time or against the Rekor timestamp?

I suppose we should be consistent with Cosign here, and AFAICT (probably best to double-check) Cosign validates against the Rekor timestamp if Rekor is trusted. (That, similarly to the Fulcio case, allows using short-lived certificates.)

QiWang19:
@mtrmac PTAL

signature/policy_config_sigstore.go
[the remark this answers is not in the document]
QiWang19: I added this [check](https://github.com/containers/image/pull/2579/files#diff-f743238a7eb3e66b475bc601913ce4170468ddd88c670c86fec28b467f1ca04fR199)  treating such configurations as unsupported for now.

> in the evaluation in `sigstoreSignedTrustRoot`

Did you mean we also need to validate this from this function? https://github.com/containers/image/blob/35dcd38edc9ae8c7b55f6ea6de6fe2b28a88fbe6/signature/policy_eval_sigstore.go#L107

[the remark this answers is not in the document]
mtrmac: In the Fulcio case, we have
```
	case trustRoot.fulcio != nil:
		if trustRoot.rekorPublicKeys == nil {
```
enforced immediately before use in `isSignatureAccepted`. Doing an equivalent for PKI earlier in `prepareTrustRoot` would also be fine, ideally with a comment in `isSignatureAccepted` noting that we don’t need to worry about Rekor because we rejected that earlier.

TomSweeneyRedHat:
As noted elsewhere, this is approved for RHEL 9.6/10.0 if this is merged and then backported into the c/image release-5.34 branch on or before Friday, February 28, 2024.

The Jira cards for this work are:
https://issues.redhat.com/browse/RHEL-79695 - RHEL 10
https://issues.redhat.com/browse/RHEL-79694 - RHEL 9.6

QiWang19:
rebased with the latest branch

dmitris:
> [OCPNODE-2338](https://issues.redhat.com/browse/OCPNODE-2338)

^^ replaced with [RUN-2436](https://issues.redhat.com/browse/RUN-2436)

dmitris:
@mtrmac @QiWang19 really great that this is going to be released - thanks so much for pushing this on!

Have a question regarding the support for timestamping authority CA Roots but will add it to #2027.

signature/pki_cert.go
```diff
@@ -0,0 +1,87 @@
...
+	untrustedCertificate, err := parseLeafCertFromPEM(untrustedCertificateBytes)
+	if err != nil {
+		return nil, err
+	}
+
+	// Go rejects Subject Alternative Name that has no DNSNames, EmailAddresses, IPAddresses and URIs;
+	// we match SAN ourselves, so override that.
```
QiWang19, line 54: @mtrmac  Also added this code block for PKI, PTAL

mtrmac: I’m surprised by this. Fulcio can create Go-unsupported SubjectAlternativeName values ( https://github.com/sigstore/fulcio/blob/3b794fbc67e6cc9ec73ddf258fe0ac482542041f/pkg/identity/username/principal.go#L71 ), but I don’t see why they should exist for non-Fulcio certificates.

And if they _do_ exist, and are marked critical, then we are obligated to actually process them. We don’t do that (we just match on `.EmailAddresses` and `.VerifyHostname`), so I think we should fail processing such certificates.

Am I missing something?

(I think the “OtherName in SAN” test case might need updating — that’s fine.)

QiWang19: Thanks for the explanation, I got this code block here because the “OtherName in SAN” failed.

signature/policy_config_sigstore_test.go
```diff
@@ -157,18 +190,24 @@ func TestNewPRSigstoreSigned(t *testing.T) {
 				},
 			},
 		} {
-			// Special-case this rejected combination:
-			if c.requiresRekor && len(c2.rekorOptions) == 0 {
+			switch {
+			case c.requiresRekor == rekorRequired && len(c2.rekorOptions) == 0: // Special-case this rejected combination
 				continue
```
mtrmac, line 195: Can’t all of this be just
```go
if (c.requiresRekor == rekorRequired && len(c2.rekorOptions) == 0) ||
   (c.requiresRekor == rekorForbidden && len(c2.rekorOptions) != 0)) {
    continue
}
```
? I think that would process the "no Rekor" entry of `c2`, without having to special-case the test further.

signature/fulcio_cert.go
```diff
@@ -195,6 +186,22 @@ func (f *fulcioTrustRoot) verifyFulcioCertificateAtTime(relevantTime time.Time,
...
+func parseLeafCertFromPEM(untrustedCertificateBytes []byte) (*x509.Certificate, error) {
+	untrustedLeafCerts, err := cryptoutils.UnmarshalCertificatesFromPEM(untrustedCertificateBytes)
+	if err != nil {
+		return nil, internal.NewInvalidSignatureError(fmt.Sprintf("parsing leaf certificate: %v", err))
+	}
+	switch len(untrustedLeafCerts) {
+	case 0:
+		return nil, internal.NewInvalidSignatureError("no certificate found in signature certificate data")
+	case 1:
+		break // OK
```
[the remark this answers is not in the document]
QiWang19: also fixed this

mtrmac (approved):
Thanks! Great work.

mtrmac:
> if this is merged and then backported into the c/image release-5.34 branch on or before Friday, February 28, 2024.

@TomSweeneyRedHat I have filed #2725 for the backport.

Also, #2724 can track adding the corresponding signing code. After that, no new options should be necessary in Skopeo/Podman, just updating c/image.

TomSweeneyRedHat:
@QiWang19 v5.34.1 release just dropped on c/image, FYI, and Happy Fri-YAY!

QiWang19:
@TomSweeneyRedHat @mtrmac Thanks!

PR #2582, opened 2024-09-24
Fix: Improving error log message for missing policy.json file

Sativarsainath-26:
resolves:  https://github.com/containers/podman/issues/23852 

Modified only policy.json file to resolve above issue ---> if this implementation looks Good will modify unit-test cases also.

### DIFF
--- a/signature/policy_config.go
+++ b/signature/policy_config.go
@@ -51,28 +51,39 @@ func (err InvalidPolicyFormatError) Error() string {
 // NOTE: When this function returns an error, report it to the user and abort.
 // DO NOT hard-code fallback policies in your application.
 func DefaultPolicy(sys *types.SystemContext) (*Policy, error) {
-	return NewPolicyFromFile(defaultPolicyPath(sys))
+	policyPath, err := defaultPolicyPath(sys)
+	if err != nil {
+		return nil, err
+	}
+	return NewPolicyFromFile(policyPath)
 }
 
-// defaultPolicyPath returns a path to the default policy of the system.
-func defaultPolicyPath(sys *types.SystemContext) string {
-	return defaultPolicyPathWithHomeDir(sys, homedir.Get())
+// defaultPolicyPath returns a path to the relevant policy of the system, or an error if the policy is missing.
+func defaultPolicyPath(sys *types.SystemContext) (string, error) {
+	policyFilePath, err := defaultPolicyPathWithHomeDir(sys, homedir.Get(), systemDefaultPolicyPath)
+	if err != nil {
+		return "", err
+	}
+	return policyFilePath, nil
 }
 
 // defaultPolicyPathWithHomeDir is an internal implementation detail of defaultPolicyPath,
-// it exists only to allow testing it with an artificial home directory.
-func defaultPolicyPathWithHomeDir(sys *types.SystemContext, homeDir string) string {
+// it exists only to allow testing it with artificial paths.
+func defaultPolicyPathWithHomeDir(sys *types.SystemContext, homeDir string, systemPolicyPath string) (string, error) {
 	if sys != nil && sys.SignaturePolicyPath != "" {
-		return sys.SignaturePolicyPath
+		return sys.SignaturePolicyPath, nil
 	}
 	userPolicyFilePath := filepath.Join(homeDir, userPolicyFile)
 	if err := fileutils.Exists(userPolicyFilePath); err == nil {
-		return userPolicyFilePath
+		return userPolicyFilePath, nil
 	}
 	if sys != nil && sys.RootForImplicitAbsolutePaths != "" {
-		return filepath.Join(sys.RootForImplicitAbsolutePaths, systemDefaultPolicyPath)
+		return filepath.Join(sys.RootForImplicitAbsolutePaths, systemPolicyPath), nil
 	}
-	return systemDefaultPolicyPath
+	if err := fileutils.Exists(systemPolicyPath); err == nil {
+		return systemPolicyPath, nil
+	}
+	return "", fmt.Errorf("no policy.json file found at any of the following: %q, %q", userPolicyFilePath, systemPolicyPath)
 }
 
 // NewPolicyFromFile returns a policy configured in the specified file.
--- a/signature/policy_config_test.go
+++ b/signature/policy_config_test.go
@@ -165,7 +165,7 @@ func TestDefaultPolicyPath(t *testing.T) {
 		// There is a context, but it does not override the path.
 		{&types.SystemContext{}, false, true, tempsystemdefaultpath, ""},
 		// Path overridden
-		{&types.SystemContext{SignaturePolicyPath: nondefaultPath}, false, false, nondefaultPath, ""},
+		{&types.SystemContext{SignaturePolicyPath: nondefaultPath}, false, true, nondefaultPath, ""},
 		// Root overridden
 		{
 			&types.SystemContext{RootForImplicitAbsolutePaths: rootPrefix},
--- a/signature/policy_config_test.go
+++ b/signature/policy_config_test.go
@@ -152,35 +152,41 @@ func TestDefaultPolicyPath(t *testing.T) {
 	const rootPrefix = "/root/prefix"
 	tempHome := t.TempDir()
 	userDefaultPolicyPath := filepath.Join(tempHome, userPolicyFile)
-
+	tempsystemdefaultpath := filepath.Join(tempHome, systemDefaultPolicyPath)
 	for _, c := range []struct {
-		sys             *types.SystemContext
-		userfilePresent bool
-		expected        string
+		sys               *types.SystemContext
+		userfilePresent   bool
+		systemfilePresent bool
+		expected          string
+		expectedError     string
 	}{
 		// The common case
-		{nil, false, systemDefaultPolicyPath},
+		{nil, false, true, tempsystemdefaultpath, ""},
 		// There is a context, but it does not override the path.
-		{&types.SystemContext{}, false, systemDefaultPolicyPath},
+		{&types.SystemContext{}, false, true, tempsystemdefaultpath, ""},
 		// Path overridden
-		{&types.SystemContext{SignaturePolicyPath: nondefaultPath}, false, nondefaultPath},
+		{&types.SystemContext{SignaturePolicyPath: nondefaultPath}, false, false, nondefaultPath, ""},
 		// Root overridden
 		{
 			&types.SystemContext{RootForImplicitAbsolutePaths: rootPrefix},
 			false,
-			filepath.Join(rootPrefix, systemDefaultPolicyPath),
+			true,
+			filepath.Join(rootPrefix, tempsystemdefaultpath),
+			"",
 		},
 		// Empty context and user policy present
-		{&types.SystemContext{}, true, userDefaultPolicyPath},
+		{&types.SystemContext{}, true, true, userDefaultPolicyPath, ""},
 		// Only user policy present
-		{nil, true, userDefaultPolicyPath},
+		{nil, true, true, userDefaultPolicyPath, ""},
 		// Context signature path and user policy present
 		{
 			&types.SystemContext{
 				SignaturePolicyPath: nondefaultPath,
 			},
 			true,
+			true,
 			nondefaultPath,
+			"",
 		},
 		// Root and user policy present
 		{
@@ -188,7 +194,9 @@ func TestDefaultPolicyPath(t *testing.T) {
 				RootForImplicitAbsolutePaths: rootPrefix,
 			},
 			true,
+			true,
 			userDefaultPolicyPath,
+			"",
 		},
 		// Context and user policy file preset simultaneously
 		{
@@ -197,7 +205,9 @@ func TestDefaultPolicyPath(t *testing.T) {
 				SignaturePolicyPath:          nondefaultPath,
 			},
 			true,
+			true,
 			nondefaultPath,
+			"",
 		},
 		// Root and path overrides present simultaneously,
 		{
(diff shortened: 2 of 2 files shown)


rhatdan:
LGTM
@mtrmac PTAL

signature/policy_config.go
```diff
@@ -51,28 +52,44 @@ func (err InvalidPolicyFormatError) Error() string {
...
 func DefaultPolicy(sys *types.SystemContext) (*Policy, error) {
-	return NewPolicyFromFile(defaultPolicyPath(sys))
+	policyPath, err := defaultPolicyPath(sys)
+	if err != nil {
+		return nil, err
+	}
+	return NewPolicyFromFile(policyPath)
 }
 
 // defaultPolicyPath returns a path to the default policy of the system.
```
mtrmac, line 62: If this now fails if the file is missing, that needs to be documented here.

signature/policy_config.go
```diff
@@ -51,28 +52,44 @@ func (err InvalidPolicyFormatError) Error() string {
...
+func defaultPolicyPathWithHomeDir(sys *types.SystemContext, homeDir string) (string, error) {
+	var checkedPolicyPaths []string // To store the paths being checked
+
 	if sys != nil && sys.SignaturePolicyPath != "" {
-		return sys.SignaturePolicyPath
+		return sys.SignaturePolicyPath, nil
 	}
+
 	userPolicyFilePath := filepath.Join(homeDir, userPolicyFile)
+	checkedPolicyPaths = append(checkedPolicyPaths, userPolicyFilePath)
```
mtrmac, line 81: A microoptimization: These things can be recorded only after the file is determined to be missing.

… or, alternatively, don’t compute a `[]string` on the fly at all: We know that if we get to the end, we have tried exactly two files, and we already have variables with the full paths; so we can construct the error message directly without `strings.Join`.

signature/policy_config.go
```diff
@@ -51,28 +52,44 @@ func (err InvalidPolicyFormatError) Error() string {
...
 	if sys != nil && sys.RootForImplicitAbsolutePaths != "" {
-		return filepath.Join(sys.RootForImplicitAbsolutePaths, systemDefaultPolicyPath)
+		return filepath.Join(sys.RootForImplicitAbsolutePaths, systemDefaultPolicyPath), nil
+	}
+	checkedPolicyPaths = append(checkedPolicyPaths, systemDefaultPolicyPath)
+	if err := fileutils.Exists(systemDefaultPolicyPath); err == nil {
+		return systemDefaultPolicyPath, nil
 	}
-	return systemDefaultPolicyPath
+	return "", fmt.Errorf("no policy.json file found at any of the following:%s", strings.Join(checkedPolicyPaths, ", "))
```
mtrmac, line 92: - Space after `:`
- It would be better to use `%q` for the individual paths, to avoid surprises.

mtrmac (commented):
Thanks! The feature makes sense.

For `signature`, we want as close to 100% code coverage by unit tests as possible, and for that I suspect the implementation will need to change, so that it doesn’t rely on the presence/absence of `systemDefaultPolicyPath` on the host where the tests are running.

signature/policy_config.go
Sativarsainath-26: I fixed this please take a look now.

signature/policy_config.go
Sativarsainath-26: It makes sense.. Thanks for your review comments. I removed `strings.Join` instead used existing variable names.. as per availability.

signature/policy_config.go
Sativarsainath-26: I didn't get it documented here means ?? 

I’m still getting familiar with this repository as I’m a beginner and new to contributing here. Apologies if my questions are basic.

mtrmac: Right now that comment says “returns a path to the default”. So, something like “returns a path to the relevant policy of the system, or an error if the policy is missing”.

Sativarsainath-26: Thanks @mtrmac ,, Updated document as per above comment.

mtrmac:
Looks good, still needs test coverage.

Sativarsainath-26:
> Looks good, still needs test coverage.

I will take a look on this,, will try to make it 100%.

Sativarsainath-26:
> Thanks! The feature makes sense.
> 
> For `signature`, we want as close to 100% code coverage by unit tests as possible, and for that I suspect the implementation will need to change, so that it doesn’t rely on the presence/absence of `systemDefaultPolicyPath` on the host where the tests are running.

To avoid dependency on presence/absence of `systemDefaultPolicyPath` --> Any ideas you have to resolve this @mtrmac .

mtrmac:
Use a separate parameter for the implementation, just like `defaultPolicyPathWithHomeDir` does for the home directory.

Sativarsainath-26:
> Use a separate parameter for the implementation, just like `defaultPolicyPathWithHomeDir` does for the home directory.

@mtrmac I added unit-test case by removing dependency on `systemDefaultPolicyPath` . Please take a look . Is this as per expectation or not  :worried: ?

signature/policy_config.go
```diff
@@ -51,28 +51,41 @@ func (err InvalidPolicyFormatError) Error() string {
...
+func defaultPolicyPath(sys *types.SystemContext) (string, error) {
+	policyFilePath, err := defaultPolicyPathWithHomeDir(sys, homedir.Get(), systemDefaultPolicyPath)
+	if err != nil {
+		return "", err
+	}
+	return policyFilePath, nil
 }
 
 // defaultPolicyPathWithHomeDir is an internal implementation detail of defaultPolicyPath,
 // it exists only to allow testing it with an artificial home directory.
```
mtrmac, line 71:
```suggestion
// it exists only to allow testing it with artificial paths.
```

signature/policy_config.go
```diff
@@ -51,28 +51,41 @@ func (err InvalidPolicyFormatError) Error() string {
...
-func defaultPolicyPathWithHomeDir(sys *types.SystemContext, homeDir string) string {
+func defaultPolicyPathWithHomeDir(sys *types.SystemContext, homeDir string, policyPath string) (string, error) {
```
mtrmac, line 72:
```suggestion
func defaultPolicyPathWithHomeDir(sys *types.SystemContext, homeDir string, systemPolicyPath string) (string, error) {
```

signature/policy_config_test.go
```diff
@@ -152,35 +152,41 @@ func TestDefaultPolicyPath(t *testing.T) {
...
-
+	tempsystemdefaultpath := filepath.Join(tempHome, systemDefaultPolicyPath)
 	for _, c := range []struct {
-		sys             *types.SystemContext
-		userfilePresent bool
-		expected        string
+		sys                  *types.SystemContext
+		userfilePresent      bool
+		expected             string
+		createSystemTempFile bool
```
mtrmac, line 160: Please move this one line above, so that the “test inputs” go before  “expected outputs” . Also, the `userfilePresent`/`createSystemTempFile` options should have a similarly structured name — I don’t care much _what_ the name is, but it should not suggest that the two are different kinds of things.

signature/policy_config_test.go
```diff
@@ -152,35 +152,41 @@ func TestDefaultPolicyPath(t *testing.T) {
...
 	}{
 		// The common case
-		{nil, false, systemDefaultPolicyPath},
+		{nil, false, tempsystemdefaultpath, true, ""},
 		// There is a context, but it does not override the path.
-		{&types.SystemContext{}, false, systemDefaultPolicyPath},
+		{&types.SystemContext{}, false, tempsystemdefaultpath, true, ""},
 		// Path overridden
-		{&types.SystemContext{SignaturePolicyPath: nondefaultPath}, false, nondefaultPath},
+		{&types.SystemContext{SignaturePolicyPath: nondefaultPath}, false, nondefaultPath, false, ""},
```
mtrmac, line 168: If this is testing the override, it should be overriding _something_, i.e. the system file should be present.

---

In _all_ of the following, pre-existing, tests, I think the system file should be present, so that we test which of the various _plausible_ options are chosen.

signature/policy_config_test.go
```diff
@@ -207,21 +217,41 @@ func TestDefaultPolicyPath(t *testing.T) {
...
 			false,
 			nondefaultPath,
+			false,
+			"",
 		},
 		// No environment expansion happens in the overridden paths
-		{&types.SystemContext{SignaturePolicyPath: variableReference}, false, variableReference},
+		{&types.SystemContext{SignaturePolicyPath: variableReference}, false, variableReference, false, ""},
+
+		{nil, false, "", false, fmt.Sprintf("no policy.json file found at any of the following: %q, %q", userDefaultPolicyPath, tempsystemdefaultpath)},
```
mtrmac, line 226: Add a comment above describing what is the case being tested. (OTOH, for consistency with the others, no need to add a blank line.)

signature/policy_config_test.go
```diff
@@ -207,21 +217,41 @@ func TestDefaultPolicyPath(t *testing.T) {
...
+				require.NoError(t, err)
+				f, err := os.Create(p.path)
+				require.NoError(t, err)
+				f.Close()
+			} else {
+				os.Remove(p.path)
+			}
+		}
+		path, err := defaultPolicyPathWithHomeDir(c.sys, tempHome, tempsystemdefaultpath)
+		if c.expectedError != "" && err != nil {
```
mtrmac, line 248: This does not test mismatch of `expectedError` vs. `err`:
`if c.ExpectedError != ""` test the returned error `else` test that (the error is `nil` and that the result is as expected).

signature/policy_config_test.go
```diff
@@ -207,21 +217,41 @@ func TestDefaultPolicyPath(t *testing.T) {
...
+			assert.Empty(t, path)
```
mtrmac, line 249: (If `err` is set, we are making no promises on `path`, it must not be used; this line, to the extent it might imply such a promise, is not really desirable and I think it’s better to remove it.)

signature/policy_config_test.go
```diff
@@ -207,21 +217,41 @@ func TestDefaultPolicyPath(t *testing.T) {
...
+			assert.Error(t, err)
```
mtrmac, line 250: (`EqualError` already implies `Error`, so this is not necessary to test separately.)

signature/policy_config.go
```diff
@@ -51,28 +51,41 @@ func (err InvalidPolicyFormatError) Error() string {
...
+
```
mtrmac, line 73: (This empty line is not typical in Go.)

mtrmac (commented):
Thanks!

Approach looks perfect overall; a few nits to make future maintenance easier.

mtrmac:
Also, eventually, please squash the commits into one; for `git bisect` it’s better when all commits are buildable.

signature/policy_config_test.go
Sativarsainath-26: Does this mean that for all test cases that is present should have this system file should be present ??

signature/policy_config.go
Sativarsainath-26: Yeah updated, will avoid this kind of things from next time.

signature/policy_config_test.go
Sativarsainath-26: Yes,, removed `asser.Error`.

signature/policy_config_test.go
Sativarsainath-26: Sure, I understood your point.

signature/policy_config_test.go
Sativarsainath-26: Yes got it.

signature/policy_config_test.go
mtrmac: Yes, all test cases from this one on (except for the newly added test, of course). On typical systems, a system policy file is always present, so that’s a more representative test.

(Alternatively, we could test both situations, but that is probably overkill.)

signature/policy_config_test.go
```diff
@@ -206,22 +216,40 @@ func TestDefaultPolicyPath(t *testing.T) {
...
+				os.Remove(p.path)
+			}
+		}
+		path, err := defaultPolicyPathWithHomeDir(c.sys, tempHome, tempsystemdefaultpath)
+		if c.expectedError != "" {
+			assert.Empty(t, path)
+			assert.EqualError(t, err, c.expectedError)
 		} else {
-			os.Remove(userDefaultPolicyPath)
+			assert.Equal(t, c.expected, path)
```
mtrmac, line 251: `require.NoError(t, err)` above this to test the other returned value; also, without that otherwise `path` might be undefined.

mtrmac (commented):
Thanks!

Good improvements; two more things, please.

signature/policy_config_test.go
Sativarsainath-26: Makes sense, updated code.

signature/policy_config_test.go
Sativarsainath-26: > Yes, all test cases from this one on (except for the newly added test, of course). On typical systems, a system policy file is always present, so that’s a more representative test.

I assumed that path is not present. Now i got clarity. I have updated `systemfilepresent` to `true` .

Sativarsainath-26:
> Also, eventually, please squash the commits into one; for `git bisect` it’s better when all commits are buildable.

Thank you for all your reviews and suggestions on this PR. It was a great learning experience.

signature/policy_config_test.go
```diff
@@ -152,43 +152,51 @@ func TestDefaultPolicyPath(t *testing.T) {
...
 	}{
 		// The common case
-		{nil, false, systemDefaultPolicyPath},
+		{nil, false, true, tempsystemdefaultpath, ""},
 		// There is a context, but it does not override the path.
-		{&types.SystemContext{}, false, systemDefaultPolicyPath},
+		{&types.SystemContext{}, false, true, tempsystemdefaultpath, ""},
 		// Path overridden
-		{&types.SystemContext{SignaturePolicyPath: nondefaultPath}, false, nondefaultPath},
+		{&types.SystemContext{SignaturePolicyPath: nondefaultPath}, false, false, nondefaultPath, ""},
```
mtrmac, line 168: `systemfilePresent` == `true` here please, we want to test the override logic.

Sativarsainath-26: Done.

mtrmac (approved):
Thanks for all that work!

PR #2583, opened 2024-09-24
Propagate `CompressedDigest`/`CompressedSize` when reusing data from another layer

mtrmac:
We record (some) `CompressedDigest`/`CompressedSize` when pulling a layer from registry; but, previously, we would lose these values (and record uncompressed values) when creating a duplicate of a layer already present in storage. Eventually, the original layer might get deleted and then we would only find a match in `BlobInfoCache`, under more restrictive conditions.

Instead, try to propagate the value to the duplicated layer. This might not work if the layer has multiple compressed representations, but that's hopefully not all _that_ frequent.

This activates the other half of https://github.com/containers/storage/pull/1830 .

### DIFF
--- a/storage/storage_dest.go
+++ b/storage/storage_dest.go
@@ -975,9 +975,11 @@ func (s *storageImageDestination) createNewLayer(index int, layerDigest digest.D
 		return nil, fmt.Errorf("internal inconsistency: layer (%d, %q) not found", index, layerDigest)
 	}
 	var trustedOriginalDigest digest.Digest // For storage.LayerOptions
+	var trustedOriginalSize *int64
 	if gotFilename {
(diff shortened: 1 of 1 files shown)


cgwalters:
> when creating a duplicate of a layer already present in storage.

For my own edification it'd be awesome if you can elaborate on why we do this somewhere

mtrmac:
The c/storage data model is designed to work on not just overlay, but snapshot systems (in the worst case, device-mapper snapshots, although that backend was recently completely removed. Relevant today the VFS driver creates layers by first copying all files from the parent into the child). In that model, a layer can only be applied on top of its “original” parent, it can’t be assembled on top of some _other_ parent.

(That conceptual model is also demonstrated by the API shape, where `Layer.Parent` is a _single_ layer ID, and immutable. Layers form a tree; it’s not a general DAG, neither are they not a pool to arbitrarily assemble images from.)

PR #2588, opened 2024-10-01
registry: remove ping v1

carlosrodfern:
The ping v1 happens when the ping v2 fails, however, it causes the ping v2 error to be skipped and not output to the user. As result, when a registry has v1 and v2 enabled, and there are, for example, intermittent connectivity issues making the ping v2 fail, the user is presented with a misleading error saying "can't talk to V1 registry."

Since the only use of v1 is for the search API as a workaround for docker.io, and new container registries setups are very unlikely to be v1-only, there is little utility in keeping this v1 detection in the attempt to help the user realize their setup is v1-only, hence not compatible. On the contratry, it just presents the user with a misleading error in certain circumstances.

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -42,7 +42,6 @@ const (
 	dockerRegistry   = "registry-1.docker.io"
 
 	resolvedPingV2URL       = "%s://%s/v2/"
-	resolvedPingV1URL       = "%s://%s/v1/_ping"
 	tagsPath                = "/v2/%s/tags/list"
 	manifestPath            = "/v2/%s/manifests/%s"
 	blobsPath               = "/v2/%s/blobs/%s"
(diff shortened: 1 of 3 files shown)


cgwalters:
/approve

Makes sense to me...

docker/errors.go
```diff
@@ -10,9 +10,6 @@ import (
 )
 
 var (
-	// ErrV1NotSupported is returned when we're trying to talk to a
-	// docker V1 registry.
-	ErrV1NotSupported = errors.New("can't talk to a V1 container registry")
```
mtrmac, line 15: This is a public API and can’t be removed, even if it will never match anything.

Adding a `Deprecated:` comment would be useful.

types/types.go
```diff
@@ -640,10 +640,6 @@ type SystemContext struct {
 	DockerBearerRegistryToken string
 	// if not "", an User-Agent header is added to each request when contacting a registry.
 	DockerRegistryUserAgent string
-	// if true, a V1 ping attempt isn't done to give users a better error. Default is false.
-	// Note that this field is used mainly to integrate containers/image into projectatomic/docker
-	// in order to not break any existing docker's integration tests.
-	DockerDisableV1Ping bool
```
mtrmac, line 646: This is a public API and can’t be removed.

Adding a `Deprecated:` comment would be useful.

docker/docker_client.go
```diff
@@ -934,37 +934,6 @@ func (c *dockerClient) detectPropertiesHelper(ctx context.Context) error {
 	if err != nil && c.tlsClientConfig.InsecureSkipVerify {
 		err = ping("http")
 	}
-	if err != nil {
-		err = fmt.Errorf("pinging container registry %s: %w", c.registry, err)
```
mtrmac, line 938: I think we still want this wrapping to happen.

docker/docker_client.go
```diff
@@ -934,37 +934,6 @@ func (c *dockerClient) detectPropertiesHelper(ctx context.Context) error {
...
-		if c.sys != nil && c.sys.DockerDisableV1Ping {
-			return err
-		}
-		// best effort to understand if we're talking to a V1 registry
-		pingV1 := func(scheme string) bool {
-			pingURL, err := url.Parse(fmt.Sprintf(resolvedPingV1URL, scheme, c.registry))
```
mtrmac, line 944: `resolvedPingV1URL` is no longer used and can be removed.

mtrmac (changes requested):
Thanks!

This makes sense to me, but we must not break API.

carlosrodfern:
Thank you for the feedback. I believe I addressed all the comments. Let me know if it needs anything else :+1:

mtrmac (approved):
Thanks again!

PR #2589, opened 2024-10-03
copy: attempt fallback to full retrieval only when possible

giuseppe:
Do not attempt the fallback mechanism unless it was reported by the underlying transport.

Follow up for: https://github.com/containers/storage/pull/2118


mtrmac:
Cc: @Honny1

internal/private/private.go
```diff
@@ -183,3 +183,21 @@ type UnparsedImage interface {
 	// UntrustedSignatures is like ImageSource.GetSignaturesWithFormat, but the result is cached; it is OK to call this however often you need.
 	UntrustedSignatures(ctx context.Context) ([]signature.Signature, error)
 }
+
+// ErrFallbackToOrdinaryLayerDownload is a custom error type returned by PutBlobPartial.
+// It suggests to the caller that a fallback mechanism can be used instead of a hard failure.
+type ErrFallbackToOrdinaryLayerDownload struct {
```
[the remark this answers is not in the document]
giuseppe: we don't want the copy package to have a dependency on containers/storage.  In this way, it is possible to build containers/image without the storage backend and avoid the vendoring cost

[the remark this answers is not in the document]
Honny1: Thanks

Honny1 (approved):
LGTM

If @mtrmac has no objections. We can merge this PR.

internal/private/private.go
```diff
@@ -183,3 +183,21 @@ type UnparsedImage interface {
...
+
+func (c ErrFallbackToOrdinaryLayerDownload) Error() string {
+	return c.Err.Error()
+}
+
+func (c ErrFallbackToOrdinaryLayerDownload) Unwrap() error {
+	return c.Err
+}
+
+func NewErrFallbackToOrdinaryLayerDownload(err error) error {
```
mtrmac, line 201: (I think we need either the `Err` field to be public, or a constructor to exist; having both seems unnecessary, but it also doesn’t hurt.)

storage/storage_dest.go
```diff
@@ -313,13 +313,20 @@ func (f *zstdFetcher) GetBlobAt(chunks []chunked.ImageSourceChunk) (chan io.Read
...
-func (s *storageImageDestination) PutBlobPartial(ctx context.Context, chunkAccessor private.BlobChunkAccessor, srcInfo types.BlobInfo, options private.PutBlobPartialOptions) (private.UploadedBlob, error) {
+func (s *storageImageDestination) PutBlobPartial(ctx context.Context, chunkAccessor private.BlobChunkAccessor, srcInfo types.BlobInfo, options private.PutBlobPartialOptions) (_ private.UploadedBlob, Err error) {
 	fetcher := zstdFetcher{
 		chunkAccessor: chunkAccessor,
 		ctx:           ctx,
 		blobInfo:      srcInfo,
 	}
 
+	var perr chunked.ErrFallbackToOrdinaryLayerDownload
+	defer func() {
```
mtrmac, line 324: Should this be checked _only_ for the `GetDiffer` call? That’s the only case where it can currently be returned.

Or maybe also `PrepareStagedLayer` — but in the `out.CompressedDigest != blobDigest` case I’d prefer to hard-fail and not fall back.

storage/storage_dest.go
```diff
@@ -313,13 +313,20 @@ func (f *zstdFetcher) GetBlobAt(chunks []chunked.ImageSourceChunk) (chan io.Read
...
 // should fall back to PutBlobWithOptions.
-func (s *storageImageDestination) PutBlobPartial(ctx context.Context, chunkAccessor private.BlobChunkAccessor, srcInfo types.BlobInfo, options private.PutBlobPartialOptions) (private.UploadedBlob, error) {
+func (s *storageImageDestination) PutBlobPartial(ctx context.Context, chunkAccessor private.BlobChunkAccessor, srcInfo types.BlobInfo, options private.PutBlobPartialOptions) (_ private.UploadedBlob, Err error) {
 	fetcher := zstdFetcher{
 		chunkAccessor: chunkAccessor,
 		ctx:           ctx,
 		blobInfo:      srcInfo,
 	}
 
+	var perr chunked.ErrFallbackToOrdinaryLayerDownload
```
mtrmac, line 323: (Nit: This can be moved inside the `defer`)

storage/storage_dest.go
```diff
@@ -313,13 +313,20 @@ func (f *zstdFetcher) GetBlobAt(chunks []chunked.ImageSourceChunk) (chan io.Read
...
+	defer func() {
+		if errors.As(Err, &perr) {
+			Err = private.NewErrFallbackToOrdinaryLayerDownload(perr.Unwrap())
```
mtrmac, line 326: (Just noting the thought for the record: hypothetically the returned error might be a _wrapped_ Fallback… and this drops the error text added by the wrapping. That’s probably fine — and it doesn’t matter with current c/storage.)

storage/storage_dest.go
```diff
@@ -313,13 +313,20 @@ func (f *zstdFetcher) GetBlobAt(chunks []chunked.ImageSourceChunk) (chan io.Read
 // It is available only if SupportsPutBlobPartial().
 // Even if SupportsPutBlobPartial() returns true, the call can fail, in which case the caller
 // should fall back to PutBlobWithOptions.
-func (s *storageImageDestination) PutBlobPartial(ctx context.Context, chunkAccessor private.BlobChunkAccessor, srcInfo types.BlobInfo, options private.PutBlobPartialOptions) (private.UploadedBlob, error) {
+func (s *storageImageDestination) PutBlobPartial(ctx context.Context, chunkAccessor private.BlobChunkAccessor, srcInfo types.BlobInfo, options private.PutBlobPartialOptions) (_ private.UploadedBlob, Err error) {
```
mtrmac, line 316: something like
```suggestion
func (s *storageImageDestination) PutBlobPartial(ctx context.Context, chunkAccessor private.BlobChunkAccessor, srcInfo types.BlobInfo, options private.PutBlobPartialOptions) (_ private.UploadedBlob, retErr error) {
```

please, so that the difference isn’t just capitalization.

---

(Thanks, I didn’t realize the returned values can be explicitly unnamed with `_`. I’ll definitely take advantage of this in the future.)

internal/private/private.go
```diff
@@ -183,3 +183,21 @@ type UnparsedImage interface {
 	// UntrustedSignatures is like ImageSource.GetSignaturesWithFormat, but the result is cached; it is OK to call this however often you need.
 	UntrustedSignatures(ctx context.Context) ([]signature.Signature, error)
 }
+
+// ErrFallbackToOrdinaryLayerDownload is a custom error type returned by PutBlobPartial.
+// It suggests to the caller that a fallback mechanism can be used instead of a hard failure.
```
mtrmac, line 188: Do we want
```suggestion
// It suggests to the caller that a fallback mechanism can be used instead of a hard failure;
// otherwise the caller of PutBlobPartial _must not_ fall back to PutBlob.
```
here?

internal/private/private.go
mtrmac, line 189: Please update the fallback discussion in the `PutBlobPartial` documentation.

copy/single.go
```diff
@@ -819,8 +819,13 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, to
...
+			// On a "partial content not available" error, ignore it and retrieve the whole layer.
+			var perr private.ErrFallbackToOrdinaryLayerDownload
+			if errors.As(err, &perr) {
+				logrus.Debugf("Failed to retrieve partial blob: %v", err)
+				return false, types.BlobInfo{}, nil
+			}
+			return false, types.BlobInfo{}, err
 		}()
 		if err != nil {
 			return types.BlobInfo{}, "", err
```
mtrmac, line 831: Now that this can realistically fail, I think this should be wrapped with a context that adds `srcInfo.Digest` like other such error paths in this function — otherwise `copyLayers` will just return the error and it will be unclear which layer it refers to.

storage/storage_dest.go
giuseppe: I had the same concern when I originally wrote it.  Since you've also pointed it out, I think it might be better to just wrap `retErr` as it won't hurt to have both wrapper types

storage/storage_dest.go
giuseppe: in this case we should be already safe no?  The error we return is not wrapped by `chunked.ErrFallbackToOrdinaryLayerDownload`:

```go
		if out.CompressedDigest != blobDigest {
			return private.UploadedBlob{}, fmt.Errorf("internal error: PrepareStagedLayer returned CompressedDigest %q not matching expected %q",
				out.CompressedDigest, blobDigest)
		}
```

giuseppe:
@mtrmac thanks for the review.  Addressed the comments and pushed a new version

storage/storage_dest.go
mtrmac: My mistake, you are absolutely correct.

internal/private/private.go
```diff
@@ -55,6 +55,7 @@ type ImageDestinationInternalOnly interface {
 	// It is available only if SupportsPutBlobPartial().
 	// Even if SupportsPutBlobPartial() returns true, the call can fail, in which case the caller
 	// should fall back to PutBlobWithOptions.
```
mtrmac, line 57: The “in which case …” clause is now incorrect, and conflicts with the added text below.

(Also please update the copies of the comment in implementations.)

mtrmac (commented):
Thanks!

LGTM otherwise, please merge after the comment update without another review.

internal/private/private.go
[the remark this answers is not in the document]
mtrmac: (Also, it’s not just the technical dependency; c/image/copy strives to contain “generic” code without hard-coding specifics of transports or image formats.

It doesn’t always succeed — we do have some special cases, and really a lot of the `private` API is specific to the c/storage design and constraints.)

giuseppe:
> Thanks!
> 
> LGTM otherwise, please merge after the comment update without another review.

thanks!  Comments addressed

PR #2595, opened 2024-10-10
Improve an error message

mtrmac:
We are seeing
> reference "[overlay@...]quay.io/...@sha256:..." does not resolve to an image ID: identifier is not an image

which is misleading; hide the ErrNoSuchImage text.


rhatdan:
LGTM

mtrmac:
The Skopeo failures should be fixed by https://github.com/containers/skopeo/pull/2442 , that is being tested in https://github.com/containers/skopeo/pull/2441 .

Issue #2602, opened 2024-10-15
expectedLayerDiffIDFlag expectation mismatch


mtrmac:
@edsantiago reports in https://github.com/containers/common/pull/2202#issuecomment-2415121670 , that we see

> `TOC digest "sha256:…" for layer "…" is present but "expected-layer-diffid" flag is not a string

By code inspection, that might be possible if the image write+read both happen in the same process (we are writing a `digest.Digest`, and expecting a `string`).

mtrmac:
A reproducer (from the point of c/image; I didn’t immediately find a Podman/Buildah one, I suspect it requires some combination of pulling chunked and non-chunked versions of the same image to prepopulate the `UncompressedDigest` field + layer deduplication, but I didn’t spend much time looking):

```console
# cat main.go 
package main

import (
	"context"
	"fmt"
	"os"

	"github.com/containers/image/v5/copy"
	"github.com/containers/image/v5/signature"
	"github.com/containers/image/v5/transports/alltransports"
	"github.com/containers/storage/pkg/reexec"
	"github.com/sirupsen/logrus"
)

func realMain() error {
	if len(os.Args) != 4 {
		return fmt.Errorf("usage: $this src intermediate dest")
	}

	ctx := context.Background()
	logrus.SetLevel(logrus.DebugLevel)

	srcRef, err := alltransports.ParseImageName(os.Args[1])
	if err != nil {
		return fmt.Errorf("%q: %w", os.Args[1], err)
	}
	midRef, err := alltransports.ParseImageName(os.Args[2])
	if err != nil {
		return fmt.Errorf("%q: %w", os.Args[2], err)
	}
	destRef, err := alltransports.ParseImageName(os.Args[3])
	if err != nil {
		return fmt.Errorf("%q: %w", os.Args[3], err)
	}

	policy := &signature.Policy{Default: []signature.PolicyRequirement{signature.NewPRInsecureAcceptAnything()}}
	policyContext, err := signature.NewPolicyContext(policy)
	if err != nil {
		return err
	}
	defer policyContext.Destroy()

	if _, err = copy.Image(ctx, policyContext, midRef, srcRef, nil); err != nil {
		return fmt.Errorf("first copy: %w", err)
	}
	if _, err = copy.Image(ctx, policyContext, destRef, midRef, nil); err != nil {
		return fmt.Errorf("second copy: %w", err)
	}
	return nil
}

func main() {
	if reexec.Init() {
		return
	}

	if err := realMain(); err != nil {
[...]

PR #2603, opened 2024-10-15
Consistently use a string type for expectedLayerDiffIDFlag

mtrmac:
Might be a fix for https://github.com/containers/image/issues/2602 .

### DIFF
--- a/storage/storage_dest.go
+++ b/storage/storage_dest.go
@@ -930,7 +930,7 @@ func (s *storageImageDestination) createNewLayer(index int, layerDigest digest.D
 
 		flags := make(map[string]interface{})
 		if untrustedUncompressedDigest != "" {
-			flags[expectedLayerDiffIDFlag] = untrustedUncompressedDigest
+			flags[expectedLayerDiffIDFlag] = untrustedUncompressedDigest.String()
 			logrus.Debugf("Setting uncompressed digest to %q for layer %q", untrustedUncompressedDigest, newLayerID)
 		}
 


mtrmac:
@edsantiago This is a _guess_ at the problem and fix; not reproduced, untested. Filing early just in case it _happened to_ unblock you.

mtrmac:
That’s … not a helpful message from c/image or c/storage. Not too likely to be directly related, but also I don’t have any idea where it is coming from.

mtrmac:
I have confirmed that this fixes #2602, at least per the custom reproducer in https://github.com/containers/image/issues/2602#issuecomment-2418087607 ; this is ready for review.

storage/storage_dest.go
```diff
@@ -930,7 +930,7 @@ func (s *storageImageDestination) createNewLayer(index int, layerDigest digest.D
 
 		flags := make(map[string]interface{})
```
cgwalters, line 931: Why was `map[string]interface{}` chosen here? It came in 15ac716f1690cb04f3694e7b295e4830217e8c62 but I don't understand why it couldn't be a structure.

cgwalters (approved):
One thing I don't understand here is what's the *other* code that is setting this to a Digest type? It doesn't seem to be in c/image or c/storage, but maybe I'm missing it.

edsantiago:
Still seeing [not supported](https://api.cirrus-ci.com/v1/artifact/task/4742340668030976/html/int-remote-fedora-40-root-host-boltdb.log.html)  and [blob size mismatch](https://api.cirrus-ci.com/v1/artifact/task/5410843737718784/html/sys-podman-fedora-40-root-host-sqlite.log.html#t--00041) but I guess one bug at a time

mtrmac:
> One thing I don't understand here is what's the _other_ code that is setting this to a Digest type?

I don’t understand the question. The patched line is the only writer, and `untrustedUncompressedDigest` is a `digest.Digest`.

storage/storage_dest.go
mtrmac: See how it is used in that commit; this goes all the way back to 7426cceb6f .

As for why this does not record specific fields … \*shrug\* it’s an escape hatch so that out-of-c/storage users can store their private data without _properly_ maintaining a c/storage type extension or without making that data public to all c/storage users; and that’s basically what’s happening here.

mtrmac:
> Still seeing [not supported](https://api.cirrus-ci.com/v1/artifact/task/4742340668030976/html/int-remote-fedora-40-root-host-boltdb.log.html)

I don’t immediately know; it’s not happening when just pulling that image from Quay.

> and [blob size mismatch](https://api.cirrus-ci.com/v1/artifact/task/5410843737718784/html/sys-podman-fedora-40-root-host-sqlite.log.html#t--00041)

Looking at https://github.com/containers/podman/blob/740f1d1fc710d4560d2463b11333cc9ad68e22bd/vendor/github.com/containers/storage/pkg/chunked/storage_linux.go#L1156 , that Podman version does not include https://github.com/containers/storage/pull/2130 yet, so that is expected. (I’m currently working with https://github.com/containers/podman/pull/24287 .)

> but I guess one bug at a time

Yes please.

edsantiago:
It's possible that "not supported" is only showing up on `boltdb` tests. (All the error logs are different and it's really hard to catalogue them all). So I spent an hour and a half trying to reproduce on a boltdb setup, with local repo very close to the mirror cache, but no luck.

edsantiago:
Hurray for stepping afk! The problem is with VFS. https://github.com/containers/podman/issues/24308

vrothberg (commented):
LGTM

giuseppe (approved):
/lgtm

PR #2607, opened 2024-10-17
Special-case an error message to make it clearer

mtrmac:
... and, also, indirectly, confirm the hypothesis in https://github.com/containers/podman/issues/24308 .

### DIFF
--- a/storage/storage_dest.go
+++ b/storage/storage_dest.go
@@ -335,7 +335,7 @@ func (s *storageImageDestination) PutBlobPartial(ctx context.Context, chunkAcces
 
 	out, err := s.imageRef.transport.store.PrepareStagedLayer(nil, differ)
 	if err != nil {
-		return private.UploadedBlob{}, err
+		return private.UploadedBlob{}, fmt.Errorf("staging a partially-pulled layer: %w", err)
 	}
 	succeeded := false
 	defer func() {
--- a/copy/single.go
+++ b/copy/single.go
@@ -828,7 +828,7 @@ func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, to
 			return false, types.BlobInfo{}, err
 		}()
 		if err != nil {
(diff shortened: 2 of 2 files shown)


mtrmac:
Confirmed in manual testing:
```console
# bin/podman --storage-driver vfs pull quay.io/libpod/testimage:20241010
Trying to pull quay.io/libpod/testimage:20241010...
Copying blob a3ed95caeb02 done   | 
Error: copying system image from manifest list: reading blob sha256:…: partial layer pull is not supported by the graph driver, and fallback is not allowed: not supported
```

cgwalters (approved):

storage/storage_dest.go
```diff
@@ -335,6 +335,11 @@ func (s *storageImageDestination) PutBlobPartial(ctx context.Context, chunkAcces
 
 	out, err := s.imageRef.transport.store.PrepareStagedLayer(nil, differ)
 	if err != nil {
+		// Special-case the error message
+		var perr chunked.ErrFallbackToOrdinaryLayerDownload
+		if errors.Is(err, storage.ErrNotSupported) && !errors.As(err, &perr) {
+			err = fmt.Errorf("partial layer pull is not supported by the graph driver, and fallback is not allowed: %w", err)
```
vrothberg, line 341: Could we list the graph driver in the error?

mtrmac: Meanwhile I have prepared https://github.com/containers/storage/pull/2140; that is where the error is going to come from in the future.

And c/image does not authoritatively know that the graph driver is the cause.

So I think I’ll simplify this again to just add more context without special-casing.

mtrmac:
Rewritten to just add / improve context. Please review again.

Right now, we should see
```
copying system image from manifest list: partial pull of blob sha256:…: staging a partially-pulled layer: not supported
```

and with https://github.com/containers/storage/pull/2140:
```
copying system image from manifest list: partial pull of blob sha256:…: graph driver vfs does not support partial pull but convert_images requires that
```

vrothberg (commented):
Code LGTM

rhatdan:
LGTM

PR #2608, opened 2024-10-21
Actually record last retry time after reconnecting

mtrmac:
The code was incorrectly setting it to "never", meaning the `bodyReaderMSSinceLastRetry` heuristic was never triggered, and, in effect, we would _always_ reconnect on `io.ErrUnexpectedEOF` or `syscall.ECONNRESET` .

This corrects the code - but it might also make more network failures user-visible instead of silently retried.


vrothberg (approved):
LGTM

rhatdan:
LGTM

PR #2609, opened 2024-10-21
Add Copy.Options.ReportResolvedReference

mtrmac:
This should allow returning precisely the correct image ID from pulls, improving on https://github.com/containers/common/pull/2202 .

To be used in c/common https://github.com/containers/common/pull/2209 .

<s>Absolutely untested at this point.</s>

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -337,7 +337,9 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
 		}
 	}
 
-	if err := c.dest.Commit(ctx, c.unparsedToplevel); err != nil {
+	if err := c.dest.CommitWithOptions(ctx, private.CommitOptions{
+		UnparsedToplevel: c.unparsedToplevel,
+	}); err != nil {
 		return nil, fmt.Errorf("committing the finished image: %w", err)
 	}
 
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -137,6 +137,16 @@ type Options struct {
 	// DestinationCtx.CompressionFormat is used exclusively, and blobs of other
 	// compression algorithms are not reused.
 	ForceCompressionFormat bool
+
+	// ReportResolvedReference, if set, asks the destination transport to store
+	// a “resolved” (more detailed) reference to the created image
+	// into the value this option points to.
+	// What “resolved” means is transport-specific.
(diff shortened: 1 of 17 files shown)


vrothberg (commented):
LGTM

mtrmac:
Note the added commit “HACK: Only return an image ID from ReportResolvedReference for c/storage”.

mtrmac:
This was manually tested using https://github.com/containers/podman/pull/24447#issuecomment-2455396376 , and a Podman test is running in https://github.com/containers/podman/pull/24462 . Please review.

Cc: @Luap99

mheon:
LGTM. I also like this better than reverting.

copy/copy.go
```diff
@@ -137,6 +137,16 @@ type Options struct {
...
+
+	// ReportResolvedReference, if set, asks the destination transport to store
+	// a “resolved” (more detailed) reference to the created image
+	// into the value this option points to.
+	// What “resolved” means is transport-specific.
+	// Most transports don’t support this, and cause the value to be set to nil.
+	//
+	// For the containers-storage: transport, the reference contains an image ID,
+	// so that storage.ResolveReference returns exactly the created image.
+	ReportResolvedReference *types.ImageReference
```
Luap99, line 150: I find it somewhat confusing that we use the Options field to return information to the caller. But I guess that is needed to avoid breaking changes in the API? I also get the that this is why it needs a pointer to an interface (which itself is also a pointer) but that seems a bit confusing to use as a caller.

Anyway I don't know the code + history here. I trust you that this is the best design

Luap99 (commented):
LGTM

copy/copy.go
mtrmac: Yes, we are committed not to change the API.

Alternatively, we could add a `copy.ImageWithResolvedReference` returning the extra value… and dropping the manifest bytes return value? I don’t know. Most users just don’t need the value.

Sort of hiding this feature in the `Options` field means we still are committed to API stability, but most users won’t see this and won’t worry about the extra values. And, also, the code implementing the copy can tell whether the caller wants the value or not.

Luap99: Yeah I am fine with this, you are the main maintainer here after all and if you like this approach. We don't use the merge bot here, do we? So feel free to merge and move it along into c/common.

PR #2610, opened 2024-10-23
Update c/storage after https://github.com/containers/storage/pull/2143

mtrmac:
This makes the declared version larger than 1.55.1, so that the Skopeo test does not downgrade to 1.55.1 from a branch. That branch is missing an API we now depend on, so c/image CI is failing; this should fix that.


mtrmac:
Merging without review to unblock CI, which this does do.

PR #2613, opened 2024-10-30
Expect UncompressedDigest to be set for partial pulls, enforce DiffID match

mtrmac:
Requires https://github.com/containers/storage/pull/2155

-  If a layer has a TOC, require that it must have a DiffID commitment in the config’s `RootFS.DiffIDs`, or refuse to pull it partially.
- Layers without a TOC continue to be allowed to use the partial pull code path, and we don't even require config's `RootFS.DiffIDs` to be present.
- If a layer has a TOC digest (i.e. could possibly be pulled partially), and c/storage has computed the uncompressed digest, require that the config's `RootFS.DiffIDs` exists and matches. This fixes the “view ambiguity” of partially-pulled layers.
- For _all_ layers, if `RootFS.DiffIDs` exists and we know the layer’s uncompressed digest, also require the `RootFS.DiffID` value to match. This might be a compatibility break, but Docker requires these values anyway.
- We happen to allow setting `DiffIDs` to empty values, if the layer does not have a TOC digest (so there is no risk of “view ambiguity”).


storage/storage_dest.go
```diff
@@ -920,12 +919,27 @@ func (s *storageImageDestination) createNewLayer(index int, layerDigest digest.D
...
 			// because we are going to write it to durable storage in expectedLayerDiffIDFlag .
 			if err := untrustedUncompressedDigest.Validate(); err != nil {
 				return nil, err
 			}
+		} else {
+			// FIXME: Clean up. Maybe the generic code can provide us the config earlier?
+			// FIXME: Always enforce this for all layers??!
+			d, err := s.untrustedLayerDiffID(index)
+			if err != nil {
+				return nil, err
```
mtrmac, line 934: All of this completely breaks for non-OCI images: `determining DiffID for manifest type %q is not yet supported`

This needs to be implemented for v2s2, and we need a reasonable fallback for v2s1.

mtrmac:
@giuseppe RFC. I still need to address /  review some corner cases, but I think the broad outline is settled now, and Podman tests are passing.

mtrmac:
@giuseppe PTAL for an early review. This is mostly untested, but it should be feature-complete and comprehensive.

Contrary to the original plan for  https://github.com/containers/storage/issues/2180 , this minimizes the impact on `pull_options.convert_images`: schema1 and non-TOC layers are always allowed to use the partial pull code path. Only TOC-containing layers may skip it, if the config does not contain DiffID values.

mtrmac:
Note to self: https://github.com/containers/podman/pull/24419#issuecomment-2573977213 identifies some failures to be fixed.

mtrmac:
@giuseppe PTAL

Podman tests (validating that IDs are deterministic, and that DiffID matches are enforced) are passing in https://github.com/containers/podman/pull/25007 .

The only reason this is still in draft is that this PR depends on https://github.com/containers/storage/pull/2214 for `go.mod` versions to resolve correctly.

mtrmac:
Ready for review.

mheon:
LGTM, though given my unfamiliarity with the codebase a review from someone else would be much more valuable.

TomSweeneyRedHat:
@nalind or @rhatdan or @giuseppe could one of you give this an eyeball please?

giuseppe (approved):
LGTM

TomSweeneyRedHat:
LGTM

TomSweeneyRedHat:
Great work @mtrmac !
/lgtm

PR #2622, opened 2024-11-06
DO NOT MERGE: Prototype setting image ID = manifest digest

mtrmac:
Cc: @giuseppe

### DIFF
--- a/storage/storage_dest.go
+++ b/storage/storage_dest.go
@@ -628,89 +628,6 @@ func (s *storageImageDestination) trustedLayerIdentityDataLocked(layerIndex int,
 	return res, true
 }
 
-// computeID computes a recommended image ID based on information we have so far.  If
-// the manifest is not of a type that we recognize, we return an empty value, indicating
-// that since we don't have a recommendation, a random ID should be used if one needs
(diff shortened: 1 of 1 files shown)


storage/storage_dest.go
```diff
@@ -1295,6 +1292,9 @@ func (s *storageImageDestination) CommitWithOptions(ctx context.Context, options
 		if err != nil {
 			return fmt.Errorf("reading image %q: %w", intendedID, err)
 		}
+		// FIXME: With intendedID typically based on s.manifestDigest, this will fail if we ever fetch the same layer two different ways (with two different layer IDs
+		// per trustedLayerIdentityData), e.g. if we see the layer/images across graph drivers which do/don’t support partial pulls.
+		// The only way to recover would be to remove the other image, and that might not be possible.
```
mtrmac, line 1297: ⚠️

mtrmac: FWIW this also completely rules out the capability to make partial pulls _optional_. If we want a single view of each image, we have to actually only produce a single view.

mtrmac:
Closing, not planning to do this short-term.

PR #2629, opened 2024-11-08
docker: drop use of external distribution challenge pkg

flavianmissi:
in distribution v3, the registry/client package became internal. this change copies `responseChallenges` from upstream, removing the last reference to registry/client in this repo, making it ready for the distribution v3 bump, whenever that comes.

---

NOTE: I expected this change to cause `go mod tidy && go mod vendor` to remove the `challenge` package from the vendor dir, but that didn't happen - not sure why.

### DIFF
--- a/docker/distribution_error.go
+++ b/docker/distribution_error.go
@@ -24,7 +24,6 @@ import (
 	"slices"
 
 	"github.com/docker/distribution/registry/api/errcode"
-	dockerChallenge "github.com/docker/distribution/registry/client/auth/challenge"
 )
 
 // errNoErrorsInBody is returned when an HTTP response body parses to an empty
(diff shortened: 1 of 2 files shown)


mtrmac:
> NOTE: I expected this change to cause `go mod tidy && go mod vendor` to remove the `challenge` package from the vendor dir, but that didn't happen - not sure why.

```console
% go mod why github.com/docker/distribution/registry/client
# github.com/docker/distribution/registry/client
github.com/containers/image/v5/pkg/docker/config
github.com/containers/image/v5/pkg/docker/config.test
github.com/docker/docker/registry
github.com/docker/distribution/registry/client/auth
github.com/docker/distribution/registry/client
```

mtrmac (commented):
Thanks! `distribution/distribution/v3` would be a separate Go module from `docker/distribution v2.…+incompatible` , so this is not immediately urgent, but removing duplicated code is always nice.

Could you restructure the code to make the handling of `StatusCode` more explicit, please? We don’t really need the separate `responseChallenges` function, and it seems somewhat non-obvious to me that a function with that name does not just parse challenges, but conditions that on a status (“A server MAY generate a WWW-Authenticate header field in other response messages” per RFC 7235)

Something like
```go
switch {
case resp.StatusCode == http.StatusUnauthorized:
   // …
   for _, c := range parseAuthHeader(…) {
   }
   fallthrough
case resp.StatusCode >= 400 && resp.StatusCode < 500:
   …
default:
  …
}
``` 
maybe

flavianmissi:
I did as you suggested and also added some more unit tests for `handleErrorResponse` to make sure my changes didn't change existing behavior. The new tests pass on the main branch as well as on this one.
PTAL @mtrmac

mtrmac (approved):
Thanks!

PR #2633, opened 2024-11-14
OCI layout extensions

vrothberg:
Taken over from https://github.com/containers/image/pull/2567 but:
* Added tests for the three commits
* Use reflinks when adding a file

### DIFF
--- a/docs/containers-transports.5.md
+++ b/docs/containers-transports.5.md
@@ -71,13 +71,15 @@ An image stored in the docker daemon's internal storage.
 The image must be specified as a _docker-reference_ or in an alternative _algo_`:`_digest_ format when being used as an image source.
 The _algo_`:`_digest_ refers to the image ID reported by docker-inspect(1).
 
-### **oci:**_path_[`:`_reference_]
+### **oci:**_path_[`:`{_reference_|`@`_source-index_}]
 
 An image in a directory structure compliant with the "Open Container Image Layout Specification" at _path_.
 
 The _path_ value terminates at the first `:` character; any further `:` characters are not separators, but a part of _reference_.
 The _reference_ is used to set, or match, the `org.opencontainers.image.ref.name` annotation in the top-level index.
 If _reference_ is not specified when reading an image, the directory must contain exactly one image.
+For reading images, @_source-index_ is a zero-based index in manifest (to access untagged images).
+If neither reference nor @_source_index is specified when reading an image, the path must contain exactly one image.
 
 ### **oci-archive:**_path_[`:`_reference_]
 
--- a/internal/reflink/reflink_linux.go
+++ b/internal/reflink/reflink_linux.go
@@ -1,0 +1,22 @@
+//go:build linux
+
+package reflink
+
+import (
+	"io"
+	"os"
+
+	"golang.org/x/sys/unix"
+)
+
+// LinkOrCopy attempts to reflink the source to the destination fd.
+// If reflinking fails or is unsupported, it falls back to io.Copy().
+func LinkOrCopy(src, dst *os.File) error {
+	_, _, errno := unix.Syscall(unix.SYS_IOCTL, dst.Fd(), unix.FICLONE, src.Fd())
+	if errno == 0 {
+		return nil
+	}
+
+	_, err := io.Copy(dst, src)
+	return err
+}
--- a/oci/layout/oci_dest.go
+++ b/oci/layout/oci_dest.go
@@ -38,6 +38,9 @@ type ociImageDestination struct {
 
 // newImageDestination returns an ImageDestination for writing to an existing directory.
 func newImageDestination(sys *types.SystemContext, ref ociReference) (private.ImageDestination, error) {
+	if ref.sourceIndex != -1 {
+		return nil, fmt.Errorf("Destination reference must not contain a manifest index @%d", ref.sourceIndex)
+	}
 	var index *imgspecv1.Index
 	if indexExists(ref) {
 		var err error
--- a/oci/layout/oci_dest.go
+++ b/oci/layout/oci_dest.go
@@ -17,6 +17,7 @@ import (
 	"github.com/containers/image/v5/internal/manifest"
 	"github.com/containers/image/v5/internal/private"
 	"github.com/containers/image/v5/internal/putblobdigest"
+	"github.com/containers/image/v5/internal/reflink"
 	"github.com/containers/image/v5/types"
 	"github.com/containers/storage/pkg/fileutils"
 	digest "github.com/opencontainers/go-digest"
@@ -141,8 +142,20 @@ func (d *ociImageDestination) PutBlobWithOptions(ctx context.Context, stream io.
 	if inputInfo.Size != -1 && size != inputInfo.Size {
 		return private.UploadedBlob{}, fmt.Errorf("Size mismatch when copying %s, expected %d, got %d", blobDigest, inputInfo.Size, size)
 	}
-	if err := blobFile.Sync(); err != nil {
+
+	if err := d.blobFileSyncAndRename(blobFile, blobDigest, &explicitClosed); err != nil {
 		return private.UploadedBlob{}, err
+	}
+	succeeded = true
+	return private.UploadedBlob{Digest: blobDigest, Size: size}, nil
+}
+
+// blobFileSyncAndRename syncs the specified blobFile on the filesystem and renames it to the
+// specific blob path determined by the blobDigest. The closed pointer indicates to the caller
+// whether blobFile has been closed or not.
+func (d *ociImageDestination) blobFileSyncAndRename(blobFile *os.File, blobDigest digest.Digest, closed *bool) error {
+	if err := blobFile.Sync(); err != nil {
+		return err
 	}
 
 	// On POSIX systems, blobFile was created with mode 0600, so we need to make it readable.
@@ -151,26 +164,27 @@ func (d *ociImageDestination) PutBlobWithOptions(ctx context.Context, stream io.
 	// always fails on Windows.
 	if runtime.GOOS != "windows" {
 		if err := blobFile.Chmod(0644); err != nil {
-			return private.UploadedBlob{}, err
+			return err
 		}
 	}
 
 	blobPath, err := d.ref.blobPath(blobDigest, d.sharedBlobDir)
 	if err != nil {
-		return private.UploadedBlob{}, err
+		return err
 	}
 	if err := ensureParentDirectoryExists(blobPath); err != nil {
-		return private.UploadedBlob{}, err
+		return err
 	}
 
-	// need to explicitly close the file, since a rename won't otherwise not work on Windows
+	// need to explicitly close the file, since a rename won't otherwise work on Windows
 	blobFile.Close()
-	explicitClosed = true
+	*closed = true
+
 	if err := os.Rename(blobFile.Name(), blobPath); err != nil {
-		return private.UploadedBlob{}, err
+		return err
 	}
-	succeeded = true
-	return private.UploadedBlob{Digest: blobDigest, Size: size}, nil
+
+	return nil
 }
 
 // TryReusingBlobWithOptions checks whether the transport already contains, or can efficiently reuse, a blob, and if so, applies it to the current destination
@@ -301,6 +315,67 @@ func (d *ociImageDestination) CommitWithOptions(ctx context.Context, options pri
 		return err
 	}
 	return os.WriteFile(d.ref.indexPath(), indexJSON, 0644)
+}
+
+// PutBlobFromLocalFileOption is unused but may receive functionality in the future.
+type PutBlobFromLocalFileOption struct{}
+
+// PutBlobFromLocalFile arranges the data from path to be used as blob with digest.
+// It computes, and returns, the digest and size of the used file.
+//
+// This function can be used instead of dest.PutBlob() where the ImageDestination requires PutBlob() to be called.
+func PutBlobFromLocalFile(ctx context.Context, dest types.ImageDestination, file string, options ...PutBlobFromLocalFileOption) (digest.Digest, int64, error) {
+	d, ok := dest.(*ociImageDestination)
+	if !ok {
+		return "", -1, errors.New("internal error: PutBlobFromLocalFile called with a non-oci: destination")
+	}
+
+	succeeded := false
+	blobFileClosed := false
+	blobFile, err := os.CreateTemp(d.ref.dir, "oci-put-blob")
+	if err != nil {
+		return "", -1, err
+	}
+	defer func() {
+		if !blobFileClosed {
+			blobFile.Close()
+		}
+		if !succeeded {
+			os.Remove(blobFile.Name())
+		}
+	}()
+
+	srcFile, err := os.Open(file)
+	if err != nil {
+		return "", -1, err
+	}
+	defer srcFile.Close()
+
+	err = reflink.LinkOrCopy(srcFile, blobFile)
+	if err != nil {
+		return "", -1, err
+	}
+
+	_, err = blobFile.Seek(0, io.SeekStart)
+	if err != nil {
+		return "", -1, err
+	}
+	blobDigest, err := digest.FromReader(blobFile)
+	if err != nil {
+		return "", -1, err
+	}
+
+	fileInfo, err := blobFile.Stat()
+	if err != nil {
+		return "", -1, err
+	}
+
+	if err := d.blobFileSyncAndRename(blobFile, blobDigest, &blobFileClosed); err != nil {
+		return "", -1, err
+	}
+
+	succeeded = true
+	return blobDigest, fileInfo.Size(), nil
 }
 
 func ensureDirectoryExists(path string) error {
--- a/oci/layout/oci_transport.go
+++ b/oci/layout/oci_transport.go
@@ -61,22 +61,31 @@ type ociReference struct {
 	// (But in general, we make no attempt to be completely safe against concurrent hostile filesystem modifications.)
 	dir         string // As specified by the user. May be relative, contain symlinks, etc.
 	resolvedDir string // Absolute path with no symlinks, at least at the time of its creation. Primarily used for policy namespaces.
-	// If image=="", it means the "only image" in the index.json is used in the case it is a source
-	// for destinations, the image name annotation "image.ref.name" is not added to the index.json
+	// If image=="" && sourceIndex==-1, it means the "only image" in the index.json is used in the case it is a source
+	// for destinations, the image name annotation "image.ref.name" is not added to the index.json.
+	//
+	// Must not be set if sourceIndex is set (the value is not -1).
 	image string
+	// If not -1, a zero-based index of an image in the manifest index. Valid only for sources.
+	// Must not be set if image is set.
+	sourceIndex int
 }
 
 // ParseReference converts a string, which should not start with the ImageTransport.Name prefix, into an OCI ImageReference.
 func ParseReference(reference string) (types.ImageReference, error) {
-	dir, image := internal.SplitPathAndImage(reference)
-	return NewReference(dir, image)
+	dir, image, index, err := internal.ParseReferenceIntoElements(reference)
+	if err != nil {
+		return nil, err
+	}
+	return newReference(dir, image, index)
 }
 
-// NewReference returns an OCI reference for a directory and a image.
+// newReference returns an OCI reference for a directory, and an image name annotation or sourceIndex.
 //
+// If sourceIndex==-1, the index will not be valid to point out the source image, only image will be used.
 // We do not expose an API supplying the resolvedDir; we could, but recomputing it
 // is generally cheap enough that we prefer being confident about the properties of resolvedDir.
-func NewReference(dir, image string) (types.ImageReference, error) {
+func newReference(dir, image string, sourceIndex int) (types.ImageReference, error) {
 	resolved, err := explicitfilepath.ResolvePathToFullyExplicit(dir)
 	if err != nil {
 		return nil, err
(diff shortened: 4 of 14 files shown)


vrothberg:
In draft for now.

oci/layout/oci_dest.go
```diff
@@ -302,6 +303,90 @@ func (d *ociImageDestination) CommitWithOptions(ctx context.Context, options pri
...
+		return private.UploadedBlob{}, false, err
+	}
+	return private.UploadedBlob{Digest: blobDigest, Size: fileInfo.Size()}, false, nil
+}
+
+// PutBlobFromLocalFile arranges the data from path to be used as blob with digest.
+// It computes, and returns, the digest and size of the used file.
+//
+// This function can be used instead of dest.PutBlob() where the ImageDestination requires PutBlob() to be called.
+func PutBlobFromLocalFile(ctx context.Context, dest types.ImageDestination, file string) (digest.Digest, int64, error) {
```
vrothberg, line 355: @mtrmac @baude
I think we should add the functionality of specifying the media type of the blob/layer here.  A `podman artifact add --mediatype something.something.ai` may actually be needed.  Otherwise, the manifest would suggest that the layers/files are compressed tar balls.

vrothberg: Ah, scratch that.  The caller is expected to update the manifest in those cases.

I implemented a very similar mechanism in `buildah source add` which automatically adds the layers to the manifest and updates the digests etc:
https://github.com/containers/buildah/blob/main/internal/source/add.go#L46

I am OK with the current design but it's probably something we should add to the docs of the function.

mtrmac: > Ah, scratch that. The caller is expected to update the manifest in those cases.

Yes. Copy&pasting from https://github.com/containers/image/pull/2567#discussion_r1842779729:

The way the `ImageDestination` API works, `PutBlob` doesn’t typically need to worry about the MIME type: the MIME type comes via `PutManifest`.

While it is true that the `PutBlobWithOptions` gets a full `types.BlobInfo`, and https://github.com/containers/image/blob/0c42f2fe704c9eb84a0b91f4f0363c9919a677db/copy/single.go#L852 forwards the MIME type value into the layer copy code, there are things like https://github.com/containers/image/blob/0c42f2fe704c9eb84a0b91f4f0363c9919a677db/copy/compression.go#L168 just dropping that; so, transports can’t rely on a MIME type being present (~instead, we have `PutBlobOptions.IsConfig`).

Also, _specifically in the OCI transport_, the `PutBlob*` code really doesn’t care about the MIME type. It is going to create a sha256-named file, and as far as the layout definition is concerned, it’s perfectly valid to refer to a single blob in two different manifests using two different MIME types. So, for this _OCI-transport-specific_ function, I don’t think adding a MIME type is right.

oci/layout/oci_dest.go
```diff
@@ -299,6 +303,90 @@ func (d *ociImageDestination) CommitWithOptions(ctx context.Context, options pri
...
+func tryHardlinkLocalFile(dest *ociImageDestination, file string) (private.UploadedBlob, bool, error) {
+	blobFile, err := os.CreateTemp(dest.ref.dir, "oci-put-blob")
+	if err != nil {
+		return private.UploadedBlob{}, false, err
+	}
+
+	// Remove the file for os.Link() to work.
+	blobFile.Close()
+	os.Remove(blobFile.Name())
+	if err := os.Link(file, blobFile.Name()); err != nil {
```
mtrmac, line 316: This is conceptually racy against the `CreateTemp` above. An alternative could be to create a temporary directory, and to create this link to a file in that temporary directory.

mtrmac, line 316: I’m quite sure about a hard-link to the user’s input; the user could continue to edit that file afterwards. At the very least this would need a strong API documentation.

I think a reflink would be uncontroversial, OTOH not always possible.

oci/layout/oci_dest.go
```diff
@@ -299,6 +303,90 @@ func (d *ociImageDestination) CommitWithOptions(ctx context.Context, options pri
...
+	if err != nil {
+		return private.UploadedBlob{}, false, err
+	}
+	if err := ensureParentDirectoryExists(blobPath); err != nil {
+		return private.UploadedBlob{}, false, err
+	}
+
+	// need to explicitly close the file, since a rename won't otherwise work on Windows
+	blobFile.Close()
+	if err := os.Rename(blobFile.Name(), blobPath); err != nil {
```
mtrmac, line 340: We might already have a file with exactly that digest; that’s not an error.

mtrmac (commented):
Just a quick skim; I’m afraid I didn’t read the added tests yet.

oci/layout/oci_dest.go
vrothberg: Fair point.  Reflinks work on Linux (xfs, btrfs) and on Mac OS (? see [unix package](https://pkg.go.dev/golang.org/x/sys/unix#IoctlFileClone)).  Hardlinks work everywhere, so they are more portable.

If we envision `podman artifact` having to deal with huge files (e.g., LLMs), then hardlinks seem more attractive.  On the other hand, any change to the file after a `podman artifact add` would corrupt the image as the digest wouldn't match anymore.  The digest could be recomputed on `artifact push` for instance.

OR will `podman artifact` on the remote-clients be talking to podman-machine?

@baude can you give direction?

oci/layout/oci_dest.go
vrothberg: Will add a test for that case 👍

oci/layout/oci_dest.go
vrothberg: Had a chat with @baude.  Reflinks it will be as the artifact store must live on a Linux machine for volume-mounts etc. to work.  Will massage the PR tomorrow morning.

oci/layout/oci_dest.go
vrothberg: It turned out that os.Rename will not fail if the file already exists but replace it.

vrothberg:
I'd love to have some explicit test condition on Linux to make sure that the reflinking works.  We're using the function from c/storage which is e2e tested.  The unit tests run on tmp, so we have to create the test dirs in $HOME ... not sure we should though.

oci/layout/oci_dest.go
```diff
@@ -299,6 +304,102 @@ func (d *ociImageDestination) CommitWithOptions(ctx context.Context, options pri
...
+
+	fileInfo, err := os.Stat(blobPath)
+	if err != nil {
+		return private.UploadedBlob{}, false, err
+	}
+	return private.UploadedBlob{Digest: blobDigest, Size: fileInfo.Size()}, false, nil
+}
+
+// PutBlobFromLocalFileOptions is unused but may receive functionality in the future.
+type PutBlobFromLocalFileOptions struct{}
```
mtrmac, line 364: Good idea, but please use functional options so that presence/absence, and error reporting, is explicit.

Modeling after `signature/sigstore`, the minimum might be

```go
type PutBlobFromLocalFileOption struct{} // to later become an alias to a private, maybe function, type

func PutBlobFromLocalFile(…, opts ... PutBlobFromLocalFileOption)
```

mtrmac (commented):
Actual detailed review still pending, I’m afraid.

oci/layout/oci_dest.go
```diff
@@ -299,6 +304,102 @@ func (d *ociImageDestination) CommitWithOptions(ctx context.Context, options pri
...
+		return private.UploadedBlob{}, false, err
+	}
+
+	// need to explicitly close the file, since a rename won't otherwise work on Windows
+	blobFile.Close()
+	if err := os.Rename(blobName, blobPath); err != nil {
+		return private.UploadedBlob{}, false, err
+	}
+
+	fileInfo, err := os.Stat(blobPath)
```
mtrmac, line 356: Doing this earlier on the open file descriptor should be a tiny bit more efficient.

oci/layout/oci_dest.go
```diff
@@ -299,6 +304,102 @@ func (d *ociImageDestination) CommitWithOptions(ctx context.Context, options pri
...
+	blobFile, err = os.Open(blobName)
+	if err != nil {
+		return private.UploadedBlob{}, false, err
+	}
+	blobDigest, err := digest.FromReader(blobFile)
+	if err != nil {
+		blobFile.Close()
+		return private.UploadedBlob{}, false, err
+	}
+	blobPath, err := dest.ref.blobPath(blobDigest, dest.sharedBlobDir)
```
mtrmac, line 342: I’d prefer to share the implementation with `PutBlobWithOptions` from now on, so that the `Sync` and `Chmod` behavior, and any future additions to `PutBlobWithOptions`, are kept in sync.

oci/layout/reader.go
```diff
@@ -0,0 +1,52 @@
...
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/containers/image/v5/types"
+	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+// This file is named reader.go for consistency with other transports
+// handling of “image containers”, but we don’t  actually need a stateful reader object.
```
mtrmac, line 14:
```suggestion
// This file is named reader.go for consistency with other transports’
// handling of “image containers”, but we don’t actually need a stateful reader object.
```

oci/layout/reader_test.go
```diff
@@ -0,0 +1,68 @@
...
+				require.True(t, strings.HasSuffix(res.Reference.StringWithinTransport(), test.names[i]))
+			}
+			_, err := ParseReference(fmt.Sprintf("%s:@%d", test.path, i))
+			require.NoError(t, err)
+		}
+	}
+
+	results, err := List("fixtures/i_do_not_exist")
+	require.Error(t, err)
+	require.Nil(t, results)
```
mtrmac, line 67: This value is conceptually “undefined” so I don’t think we need to test for it, but it also doesn’t really hurt.

oci/layout/reader_test.go
```diff
@@ -0,0 +1,68 @@
...
+		require.NoError(t, err)
+		require.NotNil(t, results)
+		require.Len(t, results, test.num)
+		for i, res := range results {
+			ociRef, ok := res.Reference.(ociReference)
+			require.True(t, ok)
+			require.Equal(t, test.digests[i], res.ManifestDescriptor.Digest.String())
+			require.Equal(t, test.names[i], ociRef.image)
+			if test.names[i] != "" {
+				require.True(t, strings.HasSuffix(res.Reference.StringWithinTransport(), test.names[i]))
```
mtrmac, line 58:
```suggestion
				require.True(t, strings.HasSuffix(res.Reference.StringWithinTransport(), ":"+ test.names[i]))
```

oci/layout/reader_test.go
```diff
@@ -0,0 +1,68 @@
...
+			}
```
mtrmac, line 59: `else` test that `res.Reference` is a sourceIndex reference with the right index?

Or, alternatively, use the returned reference to read a manifest and verify that it matches `res.ManifestDescriptor.Digest`.

oci/layout/oci_dest.go
```diff
@@ -299,6 +304,102 @@ func (d *ociImageDestination) CommitWithOptions(ctx context.Context, options pri
...
+		return private.UploadedBlob{}, false, err
+	}
+
+	blobFile, err := os.CreateTemp(dest.ref.dir, "oci-put-blob")
+	if err != nil {
+		return private.UploadedBlob{}, false, err
+	}
+	blobName := blobFile.Name()
+
+	copyRange := false
```
mtrmac, line 321: Is there a reason we would not want to use this?

mtrmac (commented):
A full review now.

oci/layout/oci_dest.go
mtrmac: Oops, my mistake.

mtrmac:
> I'd love to have some explicit test condition on Linux to make sure that the reflinking works.

…
> The unit tests run on tmp, so we have to create the test dirs in $HOME ... not sure we should though.

Tests writing to non-test directories , and possibly leaving them around on aborts/failures, seems rather unexpected to me, some might consider it (or, at least, be worried that it is) hostile. So I’d prefer to leave this to e2e tests, maybe the future tests of the Podman artifact CLI.

(Just because I was looking it up, Google says reflinks don’t work on ext4. That’s not really a blocker, we could gate this behind a filesystem type check.)

oci/layout/oci_transport_test.go
```diff
@@ -148,11 +148,17 @@ func testParseReference(t *testing.T, fn func(string) (types.ImageReference, err
 		"relativepath",
```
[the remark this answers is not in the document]
vrothberg: Isn't that already being tested now with `fixtures/two_images_manifest` ?

oci/layout/reader_test.go
vrothberg: I think we are doing that above, aren't we?

oci/layout/oci_dest.go
```diff
@@ -18,10 +18,12 @@ import (
 	"github.com/containers/image/v5/internal/private"
 	"github.com/containers/image/v5/internal/putblobdigest"
 	"github.com/containers/image/v5/types"
+	reflinkCopy "github.com/containers/storage/drivers/copy"
```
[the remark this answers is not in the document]
vrothberg: AFAIUI, the following lines are the ones we need:
https://github.com/containers/storage/blob/main/drivers/copy/copy_linux.go#L51-L54

Couldn't we move that into the code here?

Cc: @giuseppe

docs/containers-transports.5.md
```diff
@@ -71,13 +71,15 @@ An image stored in the docker daemon's internal storage.
 The image must be specified as a _docker-reference_ or in an alternative _algo_`:`_digest_ format when being used as an image source.
 The _algo_`:`_digest_ refers to the image ID reported by docker-inspect(1).
 
-### **oci:**_path_[`:`_reference_]
+### **oci:**_path_[`:`_reference_|@source-index}]_
```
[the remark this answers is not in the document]
mtrmac: A bit more to improve — the opening { in {…|…} is missing, and the trailing }] should not be italic.

oci/layout/reader_test.go
mtrmac: The test above verifies that `res.ManifestDescriptor` is valid, not that `res.Reference` is usable.

In particular, changing the `index = manifestIndex` line in `reader.go` to `index = manifestIndex * 0` currently doesn’t fail tests.

oci/layout/oci_transport_test.go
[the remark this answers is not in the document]
mtrmac: Tooling says that `case ref.image != "" && ref.sourceIndex != -1:` and `case ref.sourceIndex != -1:` is not covered by tests.

internal/reflink/reflink_linux.go
```diff
@@ -0,0 +1,30 @@
+//go:build cgo
+
+package reflink
+
+/*
+#include <linux/fs.h>
+
+#ifndef FICLONE
+#define FICLONE		_IOW(0x94, 9, int)
```
mtrmac, line 9: `x/sys/unix.FICLONE` exists; if that works, we can avoid the CGo dependency.

oci/layout/oci_dest.go
mtrmac: It turns out that Go (in `go/src/os/zero_copy_linux.go`) automatically uses this within an ordinary `io.Copy` between two `*os.File` values, this does not need to be special-cased.

internal/reflink/reflink_linux.go
```diff
@@ -0,0 +1,30 @@
...
+#endif
+*/
+import "C"
+import (
+	"io"
+	"os"
+
+	"golang.org/x/sys/unix"
```
mtrmac, line 17: Please run `go mod tidy` to mark the package as a direct dependency.

oci/layout/oci_dest.go
```diff
@@ -138,36 +142,46 @@ func (d *ociImageDestination) PutBlobWithOptions(ctx context.Context, stream io.
...
-		return private.UploadedBlob{}, err
+		return err
 	}
 
-	// need to explicitly close the file, since a rename won't otherwise not work on Windows
+	// need to explicitly close the file, since a rename won't otherwise work on Windows
 	blobFile.Close()
-	explicitClosed = true
+
+	*closed = true
```
mtrmac, line 179: (Nit: Move this one line above, please, to keep it immediately next to the actual `.Close()`.)

oci/layout/oci_dest.go
```diff
@@ -300,6 +314,63 @@ func (d *ociImageDestination) CommitWithOptions(ctx context.Context, options pri
 	return os.WriteFile(d.ref.indexPath(), indexJSON, 0644)
 }
 
+// PutBlobFromLocalFileOptions is unused but may receive functionality in the future.
+type PutBlobFromLocalFileOptions struct{}
```
mtrmac, line 318:
```suggestion
type PutBlobFromLocalFileOption struct{}
```
please, multiple options can be independent items at the call site.

oci/layout/oci_dest.go
```diff
@@ -300,6 +314,63 @@ func (d *ociImageDestination) CommitWithOptions(ctx context.Context, options pri
...
+	}()
+
+	srcFile, err := os.Open(file)
+	if err != nil {
+		return "", -1, err
+	}
+	defer srcFile.Close()
+
+	// reflink.Copy will io.Copy() in case reflinking fails
+	err = reflink.Copy(srcFile, blobFile)
```
mtrmac, line 348: (Absolutely non-blocking, this is internal-only anyway, and it works just fine as is: Maybe the function could be named explicitly enough to make the comment unnecessary, something like `ReflinkOrCopy`. OTOH some linters gripe about `reflink.Reflink…`.)

oci/layout/oci_dest.go
```diff
@@ -300,6 +314,63 @@ func (d *ociImageDestination) CommitWithOptions(ctx context.Context, options pri
...
+	if err != nil {
+		return "", -1, err
+	}
+
+	_, err = blobFile.Seek(0, 0)
```
mtrmac, line 353:
```suggestion
	_, err = blobFile.Seek(0, io.SeekStart)
```
please

oci/layout/oci_dest.go
```diff
@@ -300,6 +314,63 @@ func (d *ociImageDestination) CommitWithOptions(ctx context.Context, options pri
...
+
+	blobFileClosed := false
+	blobFile, err := os.CreateTemp(d.ref.dir, "oci-put-blob")
+	if err != nil {
+		return "", -1, err
+	}
+	defer func() {
+		if !blobFileClosed {
+			blobFile.Close()
+		}
```
mtrmac, line 338: `blobFile` remains on the filesystem on failure; I think this needs the `succeeded` logic from `PutBlobWithOptions`.

oci/layout/oci_dest.go
```diff
@@ -138,36 +142,46 @@ func (d *ociImageDestination) PutBlobWithOptions(ctx context.Context, stream io.
...
-	if err := blobFile.Sync(); err != nil {
+
+	if err := d.blobFileSyncAndRename(blobFile, blobDigest, &explicitClosed); err != nil {
 		return private.UploadedBlob{}, err
 	}
+	succeeded = true
+	return private.UploadedBlob{Digest: blobDigest, Size: size}, nil
+}
+
+func (d *ociImageDestination) blobFileSyncAndRename(blobFile *os.File, blobDigest digest.Digest, closed *bool) error {
```
mtrmac, line 153: Documenting the interaction of `blobFile` and `*closed` would be nice.

mtrmac (commented):
Thanks! ACK overall.

Ideally, I’d prefer sharing the reflink call with c/storage, but the logic (usage of `c/storage/pkg/pools` requires c/storage to explicitly special-case `CopyFileRange`, which is otherwise automatic) is just different enough to make that awkward, and anyway it’s ~10 lines, so not worth worrying about.

internal/reflink/reflink_linux.go
vrothberg: Thanks a ton for catching!

oci/layout/oci_dest.go
vrothberg: `LinkOrCopy` or should work. Thanks!

oci/layout/oci_transport_test.go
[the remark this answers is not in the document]
vrothberg: Ah.  These cases cannot be covered by `TestGetManifestDescriptor` where `sourceIndex` is always `-1` AFAICS.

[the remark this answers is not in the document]
vrothberg: Can `case ref.image != "" && ref.sourceIndex != -1:` ever be executed by the code at all?
```go
// NewIndexReference returns an OCI reference for a path and a zero-based source manifest index.
func NewIndexReference(dir string, sourceIndex int) (types.ImageReference, error) {
	return newReference(dir, "", sourceIndex)
}

// NewReference returns an OCI reference for a directory and a image.
//
// We do not expose an API supplying the resolvedDir; we could, but recomputing it
// is generally cheap enough that we prefer being confident about the properties of resolvedDir.
func NewReference(dir, image string) (types.ImageReference, error) {
	return newReference(dir, image, -1)
}
```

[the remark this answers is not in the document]
vrothberg: `TestNewIndexReference` seems to cover those cases

oci/layout/reader_test.go
vrothberg: Thanks for the pointer! Fixed

docs/containers-transports.5.md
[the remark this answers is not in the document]
mtrmac: Still outstanding I think.

oci/layout/oci_transport_test.go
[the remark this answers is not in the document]
mtrmac: - You’re right, `case ref.image != "" && ref.sourceIndex != -1:` should be unreachable. We could manually manufacture an invalid value, but that would be rather excessive. Please only add a comment documenting how this code is expected to be unreachable, similar to https://github.com/containers/image/blob/75e4f15624b3056e86c7fdfb1dbcb3a999849e22/docker/daemon/daemon_transport.go#L170 .
- `case ref.sourceIndex != -1:` is legitimate and should be reachable. Yes, that does not fit the current loop in `TestGetManifestDescriptor` — adding code outside of that loop is an option. #1677 contains a smoke test for the success case, please add also a test for an out-of-bounds index.

mtrmac (commented):
https://github.com/containers/image/pull/2633#discussion_r1868048992 + https://github.com/containers/image/pull/2633#discussion_r1895718554 .

docs/containers-transports.5.md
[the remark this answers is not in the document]
vrothberg: My apologies for missing this one.

oci/layout/oci_transport_test.go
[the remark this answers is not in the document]
vrothberg: Done ✅

oci/layout/oci_transport.go
```diff
@@ -90,7 +99,26 @@ func NewReference(dir, image string) (types.ImageReference, error) {
 		return nil, err
 	}
 
-	return ociReference{dir: dir, resolvedDir: resolved, image: image}, nil
+	if sourceIndex != -1 && sourceIndex < 0 {
+		return nil, fmt.Errorf("Invalid oci: layout reference: index @%d must not be negative", sourceIndex)
+	}
+	if sourceIndex != -1 && image != "" { // Coverage: New*Reference do not pass down that condition
```
mtrmac, line 105: Please move that comment to the uncovered `getManifestDescriptor` (line 222).

We do actually test _this_ line, so it does not need a comment justifying missing test coverage.

mtrmac (commented):
Thanks!

One last thing, feel free to merge afterwards.

oci/layout/oci_transport.go
mtrmac: https://github.com/containers/image/pull/2673

mtrmac:
Thanks!

PR #2634, opened 2024-11-15
Merge tag 'v5.33.0' into main

mtrmac:
... so that Go tooling sees the main branch as later than `v5.33.0`.

Tag `v5.33.0` is equivalent to commit https://github.com/containers/image/commit/8b9301c4bfef6e31f1192eaf8729acc9a3f9858d on the `main` branch, but the tag is actually not directly on the `main` branch, so Go sees current `main` as `v5.32.3-0.202411…` , a downgrade from `v5.33.0`.

This merge should fix that.


mtrmac:
Cc: @TomSweeneyRedHat

TomSweeneyRedHat:
Ooops, forgot the LGTM loop, so officially
LGTM

PR #2635, opened 2024-11-18
In pulls, compute DiffIDs earlier, and also for v2s2

mtrmac:
This is primarily motivated by preparing for the zstd:chunked logic, but potentially useful separately:
- Have the generic code provide the image config, and thus DiffIDs, at the start of the copy. Therefore we can start committing staged chunked layers immediately, we don’t have to wait for all of them to be staged. That might be a fairly significant performance benefit.
- Support obtaining DiffIDs from v2s2 images as well. Right now that only happens in theoretical corner cases; we will want to enforce that for all images in the future.
- (Some mostly-unrelated cleanups of handling of zero-length manifests.)

<s>At this point absolutely untested.</s>

Cc: @giuseppe , filing early to allow a preliminary review in case I have missed anything important.

### DIFF
--- a/copy/single.go
+++ b/copy/single.go
@@ -109,7 +109,7 @@ func (c *copier) copySingleImage(ctx context.Context, unparsedImage *image.Unpar
 		}
 	}
 
-	if err := checkImageDestinationForCurrentRuntime(ctx, c.options.DestinationCtx, src, c.dest); err != nil {
+	if err := prepareImageConfigForDest(ctx, c.options.DestinationCtx, src, c.dest); err != nil {
 		return copySingleImageResult{}, err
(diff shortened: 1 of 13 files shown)


mtrmac:
Now manually tested; I have confirmed that `images.json` and `layers.json` are unchanged by this, except for timestamps.

This is ready for review.

mtrmac:
… and, for the record, I have tested both the full PR and a variant without the last commit (testing the “fallback” image parsing implementation).

vrothberg (commented):
Changes LGTM
@giuseppe PTAL

giuseppe (approved):
LGTM

PR #2636, opened 2024-11-18
Fix excessive memory and disk usage when sharing layers

mtrmac:
<s>Absolutely untested, filing for early review.</s>

`createNewLayer` can add new entries to `s.filenames` for uncompressed versions of layers; these entries don't match `manifest.LayerInfos`, causing the existing logic to assume they are a config and to store them as `ImageBigData`.

That's both a potentially unlimited disk space waste (recording uncompressed versions of layers unnecessarily), and a memory pressure concern (`ImageBigData` is created by reading the whole blob into memory).

Instead, precisely track the digest of the config, and never include more than one file in `ImageBigData`.

Blobs added via `PutBlob`/`TryReusingBlob` but not referenced from the manifest are now ignored instead of being recorded inside the image.

This is intended to fix https://github.com/containers/podman/issues/24527 . @nalind is there anything that depends on the current behavior that I’m missing?

Cc: @giuseppe , affects performance measurements.

### DIFF
--- a/storage/storage_dest.go
+++ b/storage/storage_dest.go
@@ -21,7 +21,6 @@ import (
 	"github.com/containers/image/v5/internal/imagedestination/stubs"
 	"github.com/containers/image/v5/internal/private"
 	"github.com/containers/image/v5/internal/putblobdigest"
-	"github.com/containers/image/v5/internal/set"
 	"github.com/containers/image/v5/internal/signature"
 	"github.com/containers/image/v5/internal/tmpdir"
 	"github.com/containers/image/v5/manifest"
@@ -121,6 +120,9 @@ type storageImageDestinationLockProtected struct {
 	filenames map[digest.Digest]string
 	// Mapping from layer blobsums to their sizes. If set, filenames and blobDiffIDs must also be set.
 	fileSizes map[digest.Digest]int64
+
+	// Config
+	configDigest digest.Digest // "" if N/A or not known yet.
 }
 
 // addedLayerInfo records data about a layer to use in this image.
@@ -214,7 +216,17 @@ func (s *storageImageDestination) PutBlobWithOptions(ctx context.Context, stream
 		return info, err
 	}
 
-	if options.IsConfig || options.LayerIndex == nil {
+	if options.IsConfig {
+		s.lock.Lock()
+		defer s.lock.Unlock()
+		if s.lockProtected.configDigest != "" {
+			return private.UploadedBlob{}, fmt.Errorf("after config %q, refusing to record another config %q",
+				s.lockProtected.configDigest.String(), info.Digest.String())
+		}
+		s.lockProtected.configDigest = info.Digest
+		return info, nil
+	}
+	if options.LayerIndex == nil {
 		return info, nil
 	}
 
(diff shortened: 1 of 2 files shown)


storage/storage_dest.go
```diff
@@ -1193,22 +1205,14 @@ func (s *storageImageDestination) CommitWithOptions(ctx context.Context, options
...
-	dataBlobs := set.New[digest.Digest]()
-	for blob := range s.lockProtected.filenames {
-		dataBlobs.Add(blob)
-	}
-	for _, layerBlob := range layerBlobs {
-		dataBlobs.Delete(layerBlob.Digest)
-	}
-	for _, blob := range dataBlobs.Values() {
-		v, err := os.ReadFile(s.lockProtected.filenames[blob])
+	// Set up to save the cnofig as a data item.  Since we only share layers, the config should be in a file.
```
rhatdan, line 1208: typo

mtrmac: Thanks! Fixed.

mtrmac:
Podman CI run: https://github.com/containers/podman/pull/24629

mtrmac:
In manual testing, I have confirmed both the bug hypothesis and that this fixes it.

edsantiago:
Can this safely be merged? The [skip F40, Go is too old](https://github.com/containers/podman/pull/24629/commits/a733d274770b633481ac94bde954c75051e1acb5) commit suggests that this won't be able to be merged into podman or buildah, but it's not clear to me if it's "until F40 gets a new Golang update" or "never as long as F40 exists". `go 1.22.6 -> .8` suggests the former but I recommend understanding this before merging.

mtrmac:
> Can you remove Draft flag?

I want to see the Podman CI ~passing first.

mtrmac:
> Can this safely be merged? The [skip F40, Go is too old](https://github.com/containers/podman/pull/24629/commits/a733d274770b633481ac94bde954c75051e1acb5) commit suggests that this won't be able to be merged into podman or buildah

That’s already the case before this PR, due to https://github.com/containers/image/pull/2601 .

>, but it's not clear to me if it's "until F40 gets a new Golang update" or "never as long as F40 exists". `go 1.22.6 -> .8` suggests the former

https://bodhi.fedoraproject.org/updates/FEDORA-2024-9b94592629 means this should be resolved shortly.

In general, I‘ve been assuming that Go patch releases tend to include vulnerability fixes, and therefore tend to be included in Fedora quickly (… and given the static linking of Go binaries, our updated RPMs would be ~blocked on that Go update anyway). That’s not been the case with Go 1.22.{7,8}.

Separately, there’s an effort to change sigstore so that it doesn’t add the dependency on very newest Golang so quickly, around https://github.com/sigstore/sigstore/issues/1879 .

mtrmac:
https://github.com/containers/podman/pull/24629 looks reasonable, rebased and ready for review.

rhatdan:
LGTM
@nalind PTAL

cgwalters:
So this was a regression from https://github.com/containers/image/pull/2067 ? Which was motivated by https://github.com/containers/storage/issues/595 ? Wow, a lot going on there.

Anyways, this code looks superficially sane to me...

mtrmac:
No, that’s not related at all; #2067 only moves the `dataBlobs` code. This logic to select what is stored as `ImageBigData` has existed since the very first versions of the storage transport.

https://github.com/containers/podman/issues/24527#issuecomment-2471530374 points at the commit that introduced this bug.

vrothberg (approved):
LGTM
Merge at will

cgwalters:
Ahh, OK. 

Tangential: just glancing at that code one thing that seems like it'd help is if we had the canonical uncompressed size too in the spec, then some of this could just be mapping between the blob descriptor and and uncompressed descriptor instead of having separate maps for digest and size or so?

PR #2665, opened 2024-12-25
docker-archive support OCI manifest media type

davidwtf:


mtrmac (changes requested):
Thanks for the PR.

This doesn’t work that way.

Locally, without updating `SupportedManifestMIMETypes` the generic copy code would convert and not send an OCI manifest, so this PR does not affect typical uses like `skopeo copy` at all.

More importantly, AFAIK this doesn’t really make a difference, the file format as implemented by Docker (who is, by definition, authoritative) is not flexible to the supported manifests. In fact, the “traditional” `docker save` format does not _have_ a manifest in this sense at all! It uses a `docker-archive`-specific top-level manifest format.

Nowadays, IIRC, Docker does support using the OCI format in `docker save`/`docker load`, but it does not use the “traditional” format for that purpose; uses the full opencontainers/image-spec archive layout (and, on `docker save`, creates a file which can be read in both the “traditional” and OCI formats).

Implementing these dual-format archives would be, potentially, interesting, but also much more work than this — e.g. the generic design of `containers/image/copy` really assumes that every copy uses a _single_ format.

And, contemplating the possibility of a future support of these dual-format archives, we _actively don’t want_ to use the “traditional” format to store OCI, or OCI-like, images, because that would preclude using the new dual-format archive format for OCI images in the future.

davidwtf:
Yes, you are right. 

I directly called the `containers/image` interface in my program without using `skopeo`.
But I encountered an error: `Unsupported manifest type, need a Docker schema 2 manifest`, which led me to mistakenly believe that `containers/image` does not support converting OCI format images to `docker-archive` format, hence I submitted this PR.

However, after reading your comment, I verified it with `skopeo` and studied the differences between my program and skopeo's use of `containers/image` in detail. Then I discovered that when calling `copy.Image`, I passed `PreserveDigests: true`, which caused it not to perform manifests conversion. My mistake.

Thank you for your reply. I will close this PR.

PR #2680, opened 2025-01-09
Update for c/storage almost always computing UncompressedDigest

mtrmac:
After https://github.com/containers/storage/pull/2155 , we ~always compute `UncompressedDigest`, but the current code was expecting that to happen only when `convert_images` is set (and we consume the whole compressed blob).

That is breaking pulls of chunked images, e.g. https://api.cirrus-ci.com/v1/artifact/task/6559054106984448/html/sys-podman-debian-13-rootless-host-sqlite.log.html :
```
internal error: PrepareStagedLayer returned CompressedDigest "" not matching expected "sha256:…"
```

Update the code to not blindly assume that `CompressedDigest` is set, and to expect `UncompressedDigest` to be set much more frequently. Originally from https://github.com/containers/image/pull/2613 , and tested as a part of the full patch set.

Separate Podman tests to confirm this truly fixes the test failure: https://github.com/containers/podman/pull/24991 .

@giuseppe PTAL.

### DIFF
--- a/storage/storage_dest.go
+++ b/storage/storage_dest.go
@@ -385,23 +385,21 @@ func (s *storageImageDestination) PutBlobPartial(ctx context.Context, chunkAcces
 		if out.UncompressedDigest != "" {
 			s.lockProtected.indexToDiffID[options.LayerIndex] = out.UncompressedDigest
 			if out.TOCDigest != "" {
+				s.lockProtected.indexToTOCDigest[options.LayerIndex] = out.TOCDigest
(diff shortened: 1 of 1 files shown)


mtrmac:
Podman tests (corresponding to the previous failure) are passing in https://github.com/containers/podman/pull/24991 , this is ready for review.

rhatdan:
LGTM

vrothberg (approved):

TomSweeneyRedHat:
Late, but LGTM.

PR #2689, opened 2025-01-21
Validate digests of data downloaded while fetching sigstore attachments

### LINKED ISSUE #2687: `dockerClient.getOCIDescriptorContents` does not validate the contents against the digest
The manifest is registry-controlled anyway (we look it up using a tag, not a digest), so this does not give the registry any new powers, but having code like that laying around is a risk.

mtrmac:
This is not a security vulnerability because the registry can just as well send a manifest modified to match, but doing this correctly protects us in case this function were used for other purposes in the future.

Fixes #2687.

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -1056,6 +1056,15 @@ func (c *dockerClient) getBlob(ctx context.Context, ref dockerReference, info ty
 func (c *dockerClient) getOCIDescriptorContents(ctx context.Context, ref dockerReference, desc imgspecv1.Descriptor, maxSize int, cache types.BlobInfoCache) ([]byte, error) {
 	// Note that this copies all kinds of attachments: attestations, and whatever else is there,
 	// not just signatures. We leave the signature consumers to decide based on the MIME type.
+
+	if err := desc.Digest.Validate(); err != nil { // .Algorithm() might panic without this check
+		return nil, fmt.Errorf("invalid digest %q: %w", desc.Digest.String(), err)
+	}
+	digestAlgorithm := desc.Digest.Algorithm()
+	if !digestAlgorithm.Available() {
+		return nil, fmt.Errorf("invalid digest %q: unsupported digest algorithm %q", desc.Digest.String(), digestAlgorithm.String())
+	}
+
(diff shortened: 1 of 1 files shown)


docker/docker_client.go
```diff
@@ -1056,6 +1056,15 @@ func (c *dockerClient) getBlob(ctx context.Context, ref dockerReference, info ty
...
+
+	if err := desc.Digest.Validate(); err != nil { // .Algorithm() might panic without this check
+		return nil, fmt.Errorf("invalid digest %q: %w", desc.Digest.String(), err)
+	}
+	digestAlgorithm := desc.Digest.Algorithm()
+	if !digestAlgorithm.Available() {
+		return nil, fmt.Errorf("invalid digest %q: unsupported digest algorithm %q", desc.Digest.String(), digestAlgorithm.String())
+	}
+
 	reader, _, err := c.getBlob(ctx, ref, manifest.BlobInfoFromOCI1Descriptor(desc), cache)
```
cgwalters, line 1068: This looks fine, but I think it'd be even clearer to merge `getBlob` and this function. Or at a minimum, rename `getBlob` to `getRawBlobNoChecksumVerification` or something...

cgwalters (approved):

docker/docker_client.go
mtrmac: `getBlob` is primarily used by the public `dockerImageSource.GetBlob`, which must support streaming very large objects, and inherits both its interface and its (non-digesting) semantics.

And `GetBlob` is (sadly?) a public stable API. Having it _automatically_ digest the contents would be interesting in the abstract, but all existing callers would should already be digesting it themselves, and doing that twice is rather costly, so I don’t think it makes sense to _change_ the behavior of the existing function.

cgwalters: OK sorry, I thought it was only called here.

mtrmac: … another point, we don’t want every single transport to implement the digest validation itself; centralizing that in c/image/copy makes it easier to prove that the validation is always happening.

But at least the _naming_ point is very apt; if we were ever redesigning the `ImageSource` API (it should be an object with methods, not an interface), `GetBlob` should be definitely renamed to something that emphasizes the caller needs to validate the contents — or maybe this should be centralized into the `ImageSource`-replacing wrapper object.

PR #2696, opened 2025-01-27
ignore ENOENT errors when parsing .d files

Luap99:
There is a expected race condition between listing files and actually opening them. Between that time the file might have been removed already. This is not really a real error so we should just ignore it as otherwise flakes will happen in the wild.

See the first commit for a real flake, the other two ones I have not observed but I thought I might as well fix it for all .d readers. Are there other files I missed?

### DIFF
--- a/docker/registries_d.go
+++ b/docker/registries_d.go
@@ -3,6 +3,7 @@ package docker
 import (
 	"errors"
 	"fmt"
+	"io/fs"
 	"net/url"
 	"os"
 	"path"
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM

mtrmac (approved):
Thanks!

(I’m not sure this will quite fix the Podman flake — there is still a race between creating the `/etc/containers/registries.d/podman-test-only-temporary-addition.yaml` file and actually writing data to it, IIRC `cp` is defined not to use the atomic `Rename()` approach — but in general, with Podman running potentially in system services, it would be hard for users to coordinate config file changes so that they reliably never interfere with a running c/image code; so I agree this is a good improvement in any case.)

PR #2700, opened 2025-01-30
Bump c/storage to v1.57.1, c/image to v5.34.0, then to v5.35.0-dev

TomSweeneyRedHat:
Bumping c/image to v5.34.0 after vendoring c/storage v1.57.0

This is in preparation for Podman v5.4

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/BurntSushi/toml v1.4.0
 	github.com/containers/libtrust v0.0.0-20230121012942-c1716e8a8d01
 	github.com/containers/ocicrypt v1.2.1
-	github.com/containers/storage v1.56.2-0.20250121150636-c2cdd500e4ef
+	github.com/containers/storage v1.57.1
 	github.com/cyberphone/json-canonicalization v0.0.0-20231217050601-ba74d44ecf5f
 	github.com/distribution/reference v0.6.0
 	github.com/docker/cli v27.5.1+incompatible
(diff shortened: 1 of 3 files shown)


TomSweeneyRedHat:
@mtrmac PTAL, especially the second commit. The version got set in the main branch to 5.34.1-dev prior.  I think creating 5.34.0 at this point will be OK, but I'm a little concerned about projects that may have vendored from main in the past few weeks.  Go vendoring sometimes gets wonky.

Luap99:
The version from the file doe snot matter a bit to the go vendoring so it makes no difference. go vendoring works based of the tag names.

So bumping to 5.34.0 seems like the correct things to do, skipping a `.0` would possibly confuse quite a few folks.

mtrmac:
> @mtrmac PTAL, especially the second commit. The version got set in the main branch to 5.34.1-dev prior.

I’m sorry, that’s my mistake, I should have caught that while creating #2691.

> I think creating 5.34.0 at this point will be OK

I agree.

As Paul says, this is invisible to Go tools. It might show up in some version information, e.g. User-Agent strings, that could be interpreted as a post-5.34.0 version. I’m worried very little about that: The important part is the `-dev` suffix, which tells us that we need to actually ask for the precise commit — and once we do, that’s precise by itself.

Per a Slack conversation, should I merge this immediately, or do you want some variant of https://github.com/containers/storage/pull/2230 be tagged _first_?

TomSweeneyRedHat:
It was found that a PR was needed in c/storage.  We are working on updating the 1.57 branch there, and will revendor here after that is done.

TomSweeneyRedHat:
I've updated this to c/storage 1.57.1 and this can now be merged after happy tests and review.  ATM, Skopeo Test is failing on what looks to be a network error.

Luap99 (approved):
LGTM

mtrmac (approved):
Thank you!

PR #2709, opened 2025-02-06
Don't silently ignore errors determining size in TryReuseBlob

mtrmac:
Returning the size required by the `TryReuseBlob`/`PutBlob` API; otherwise we could put `-1` into manifests.

- When looking for inexact matches, this will cause the matches to be skipped.
- When checking for an exact match, this will cause an upload failure; we don't have any other way to handle pre-existing blobs on the destination.

---

<s>For now, completely untested.</s>

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -998,7 +998,12 @@ func (c *dockerClient) getExternalBlob(ctx context.Context, urls []string) (io.R
 			resp.Body.Close()
 			continue
 		}
-		return resp.Body, getBlobSize(resp), nil
+
+		size, err := getBlobSize(resp)
+		if err != nil {
+			size = -1
+		}
+		return resp.Body, size, nil
 	}
 	if remoteErrors == nil {
 		return nil, 0, nil // fallback to non-external blob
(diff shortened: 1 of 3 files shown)


docker/docker_client.go
```diff
@@ -998,20 +998,25 @@ func (c *dockerClient) getExternalBlob(ctx context.Context, urls []string) (io.R
...
 	}
 	if remoteErrors == nil {
 		return nil, 0, nil // fallback to non-external blob
 	}
 	return nil, 0, fmt.Errorf("failed fetching external blob from all urls: %w", multierr.Format("", ", ", "", remoteErrors))
 }
 
-func getBlobSize(resp *http.Response) int64 {
+func getBlobSize(resp *http.Response) (int64, error) {
 	size, err := strconv.ParseInt(resp.Header.Get("Content-Length"), 10, 64)
```
mtrmac, line 1015: This should report whether the header is missing, or present and invalid (via `resp.Header.Values`)

docker/docker_client.go
```diff
@@ -998,20 +998,25 @@ func (c *dockerClient) getExternalBlob(ctx context.Context, urls []string) (io.R
...
+		return resp.Body, size, nil
 	}
 	if remoteErrors == nil {
 		return nil, 0, nil // fallback to non-external blob
 	}
 	return nil, 0, fmt.Errorf("failed fetching external blob from all urls: %w", multierr.Format("", ", ", "", remoteErrors))
 }
 
-func getBlobSize(resp *http.Response) int64 {
+func getBlobSize(resp *http.Response) (int64, error) {
```
mtrmac, line 1014: Add unit tests

mtrmac (commented):
Notes to self:

docker/docker_client.go
```diff
@@ -998,20 +998,33 @@ func (c *dockerClient) getExternalBlob(ctx context.Context, urls []string) (io.R
 			resp.Body.Close()
 			continue
 		}
-		return resp.Body, getBlobSize(resp), nil
+
+		size, err := getBlobSize(resp)
+		if err != nil {
+			size = -1
```
rhatdan, line 1004: Should we at least logrus.Debug the error?

docker/docker_client.go
```diff
@@ -1042,7 +1055,10 @@ func (c *dockerClient) getBlob(ctx context.Context, ref dockerReference, info ty
 		return nil, 0, fmt.Errorf("fetching blob: %w", err)
 	}
 	cache.RecordKnownLocation(ref.Transport(), bicTransportScope(ref), info.Digest, newBICLocationReference(ref))
-	blobSize := getBlobSize(res)
+	blobSize, err := getBlobSize(res)
+	if err != nil {
+		blobSize = -1
```
rhatdan, line 1060: logrus.Debug(err)?

docker/docker_client.go
mtrmac: In both cases, `GetBlob` is explicitly allowed to return -1, so we wouldn’t expect users to need to diagnose why this is happening. Typically, the size is only used to sanity-check sizes of uploads, but the digest verification should serve the same purpose anyway.

The only adverse effect I could find is that writes to `docker-daemon` or `docker-archive` might need to store a file to disk if the size is unknown, _and_ the input is uncompressed (if the input is compressed, the transports need to store it to disk in either case).

I don’t think this is really worth worrying about, and probably not worth the log noise, but I don’t feel strongly about this at all.

rhatdan:
LGTM

PR #2713, opened 2025-02-10
Add debugging for manifests when uploading to destinations.

robbmanes:
Add optional logging for printing outgoing manifests.  This can greatly help in debugging anything that might alter a manifest either on wire or in the use case I just had, verifying a problem in local storage (verifying the outgoing manifests without having to watch pure HTTP calls, etc).

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -358,6 +358,7 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
 		return nil, fmt.Errorf("committing the finished image: %w", err)
 	}
 
+	logrus.Debugf("Copied manifest(s): %q", string(copiedManifest))
 	return copiedManifest, nil
 }
 
--- a/copy/single.go
+++ b/copy/single.go
@@ -610,6 +610,8 @@ func (ic *imageCopier) copyUpdatedConfigAndManifest(ctx context.Context, instanc
 	if instanceDigest != nil {
 		instanceDigest = &manifestDigest
 	}
+
+	logrus.Debugf("Manifest to be written to destination: %q", string(man))
 	if err := ic.c.dest.PutManifest(ctx, man, instanceDigest); err != nil {
 		logrus.Debugf("Error %v while writing manifest %q", err, string(man))
 		return nil, "", fmt.Errorf("writing manifest: %w", err)


mtrmac (commented):
Thanks!

There already is https://github.com/containers/image/blob/8d501c721b6a5d2a2661343891efa5a5c0ed5bc8/docker/docker_image_dest.go#L522 , on failure. Is that not sufficient?

On success, it should be possible to read the uploaded manifest from the registry — due to the existence of digest references, the registry does not really have the option to modify the uploaded data.

(My basic concern is that this is fairly noisy… and potentially we might worry about escaping the data, e.g. if it included newlines.)

Luap99:
One option could be to use the trace level not debug to log this message. We don't make a lot of use of that in podman but I used that myself to trigger trace logging in netavark which dumps netlink message which was useful to me a few times. (`podman --log-level trace ...`)
That said it looks like skopeo only offers the debug level and not a per level selection like podman or buildah so if c/image starts logging with trace level skopeo would need to support that as well.

If escaping is a concern it could log with `%q` instead of `%s`.

robbmanes:
In our case here, the specific scenario we were debugging was not a failure scenario; skopeo/podman was uploading layers with a negative size value (`-1`), which is a different issue altogether I'm filing shortly, but because it didn't fail to upload we couldn't prove sans HTTP dump of the payload what the manifest coming from the client was, or if any of the myriad of proxies/data inspection security devices were interfering with the data on-wire (TLS/SSL termination and data inspection network device which ultimately led to being exonerated by this debug line, as we could see clearly the manifest being pushed was wrong).

And as @Luap99 mentioned (and was exactly right about), this was with skopeo so I initially did `Tracef` but quickly found out skopeo didn't support anything but `Debugf`, so I did that instead.  I agree it *should* be trace-level data if we updated skopeo to include such a thing though.

Also agreed with escaping the data if we prefer; I originally did `%+v` but it just wasn't necessary, and strings worked, but I'm perfectly happy to squash in a `%q` if you prefer.  Thanks!

robbmanes:
Cast as `string` and formatting with `%q` is squashed in.

robbmanes:
This really should be `Trace`, it's very verbose but still greatly helpful.  Tried my hand at adding `--log-level` to skopeo in https://github.com/containers/skopeo/pull/2514 if we want to wait on that, and then I can squash this into a `Tracef`.

copy/copy.go
```diff
@@ -358,6 +358,7 @@ func Image(ctx context.Context, policyContext *signature.PolicyContext, destRef,
 		return nil, fmt.Errorf("committing the finished image: %w", err)
 	}
 
+	logrus.Debugf("Copied manifest(s): %q", string(copiedManifest))
```
mtrmac, line 361: I think this would be better handled by logging before the `PutManifest` call in `copyMultipleImages`; that should record the same data, and avoid logging the same manifest twice in single-architecture copy situations.

mtrmac (changes requested):
Good idea, using the trace level makes sense to me. (Also #201 proposes to use trace level, so that would be consistent.)

jankaluza:

PR #2714, opened 2025-02-11
use new ReflinkOrCopy() from c/storage

Luap99:
c/storage now exposes the same function in pkg/fileutils[1] so use that
and rmeove the internal package to avoid duplication.

[1] https://github.com/containers/storage/commit/3f4b77d38829f7fcd1924a872fcd39ef37d8bf78


vrothberg (approved):
LGTM

go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/BurntSushi/toml v1.4.0
 	github.com/containers/libtrust v0.0.0-20230121012942-c1716e8a8d01
 	github.com/containers/ocicrypt v1.2.1
-	github.com/containers/storage v1.57.1
+	github.com/containers/storage v1.57.1-0.20250210200321-13a758ff4ecb
```
mtrmac, line 13: Code LGTM, but this is a downgrade. Can we merge c/storage’s `1.57.1` into `main` ?

Luap99: yes we should https://github.com/containers/storage/pull/2255

mtrmac (approved):
Thank you!

PR #2717, opened 2025-02-13
Only upload sigstore signatures to the sigstore tag schema

mtrmac:
... not to registries with X-R-S-S (that would immediately fail, when uploading to such a registry, with no opt-out), and not to
 lookaside (that would _work_ if users set up lookaside).

Before this PR, _if the image had any non-sigstore signatures_, we would upload both the non-sigstore and sigstore signatures to lookaside.

@mheon @TomSweeneyRedHat PTAL. The bug has been that way for 2.5 years. I think the combination of required circumstances (an image must have both kinds of signatures, and the user must have configured lookaside, not use of sigstore tags, while somehow requiring sigstore signatures) makes it unlikely that users would rely on it, but I can’t quite rule it out.

### DIFF
--- a/docker/docker_image_dest.go
+++ b/docker/docker_image_dest.go
@@ -615,11 +615,11 @@ func (d *dockerImageDestination) PutSignaturesWithFormat(ctx context.Context, si
 		}
 		switch {
 		case d.c.supportsSignatures:
-			if err := d.putSignaturesToAPIExtension(ctx, signatures, *instanceDigest); err != nil {
+			if err := d.putSignaturesToAPIExtension(ctx, otherSignatures, *instanceDigest); err != nil {
 				return err
 			}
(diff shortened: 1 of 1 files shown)


mheon:
I kind of wish we'd gotten this one in for 5.5 so it could soak in Fedora, but the number of people using sigstore on Fedora is likely miniscule compared to the number on RHEL so that might not have helped us any.

I'll give it some thought but I think fixing is the best course of action from what I see now.

mtrmac:
Note to self: See the final outcome of https://issues.redhat.com/browse/CLOUDDST-26881 .

mtrmac:
See documented impact in https://issues.redhat.com/browse/RUN-2504 / https://issues.redhat.com/browse/CLOUDDST-26881 .

@giuseppe PTAL.

PR #2718, opened 2025-02-14
Small cleanups + one bug fix

mtrmac:
… discovered while looking at for opportunities to benefit from new features of Go 1.23 and Go 1.24, hence a fairly random collection.

### DIFF
--- a/copy/single.go
+++ b/copy/single.go
@@ -328,19 +328,16 @@ func prepareImageConfigForDest(ctx context.Context, sys *types.SystemContext, sr
 		}
 		wantedPlatforms := platform.WantedPlatforms(sys)
 
-		options := newOrderedSet()
-		match := false
-		for _, wantedPlatform := range wantedPlatforms {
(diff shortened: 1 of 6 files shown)


Luap99 (commented):
LGTM, considering podman 5.5 is set be released in early may I think we can ignore f40 (go 1.22) for it which means we could consider updating to go 1.23 in this release cycle if there is anything specific you would like to take advantage off.

mtrmac:
(I keep branches https://github.com/containers/skopeo/compare/main...mtrmac:skopeo:go1.23 and https://github.com/mtrmac/image/compare/cleanups...mtrmac:image:go1.23 around, but that’s all code cleanups and rushing  to apply them would not benefit users — I think the primary reason to move to Go 1.23 would be to catch up to some dependencies.)

Luap99:
Yes right, there are a few go 1.23 deps already so we should update sooner than later unless someone sees a strong reason why we shouldn't.

PR #2730, opened 2025-02-22
feat: add CommitTimestamp to copy options

aeijdenberg:
Useful for reproducible oci-archive output.


aeijdenberg:
Keen to look at adding an option to `buildah` to make use of this. This is followup to https://github.com/containers/storage/pull/2259 which has been accepted and merged.

mtrmac (changes requested):
Thanks!

Makes sense overall, just a quick note for now to prevent committing us to an API prematurely.

copy/copy.go
```diff
@@ -148,6 +148,11 @@ type Options struct {
 	// so that storage.ResolveReference returns exactly the created image.
 	// WARNING: It is unspecified whether the reference also contains a reference.Named element.
 	ReportResolvedReference *types.ImageReference
+
+	// CommitTimestamp, if set, and if applicable for destination, will force
+	// timestamps to this value. Useful for oci-archive outputs where this will
+	// set the create/mod/access timestamps in each tar entry
+	CommitTimestamp *time.Time
```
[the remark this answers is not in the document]
aeijdenberg: Renamed as suggested and squash/rebased.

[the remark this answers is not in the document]
mtrmac: I’m sorry, I didn’t mean that I _prefer_ `SourceDateEpoch`. Anyway, let’s discuss and figure this out…

copy/copy.go
```diff
@@ -148,6 +148,11 @@ type Options struct {
 	// so that storage.ResolveReference returns exactly the created image.
 	// WARNING: It is unspecified whether the reference also contains a reference.Named element.
 	ReportResolvedReference *types.ImageReference
+
+	// SourceDateEpoch, if set, and if applicable for destination, will force
+	// timestamps to this value. Useful for oci-archive outputs where this will
+	// set the create/mod/access timestamps in each tar entry
+	SourceDateEpoch *time.Time
```
mtrmac, line 155: I think something like this would accurately describe what the option does, WDYT?
```suggestion
	// DestinationTimestamp, if set, will force timestamps of content created in the destination to this value.
	// Most transports don't support this.
	//
	// In oci-archive: destinations, this will set the create/mod/access timestamps in each tar entry
	// (but not a timestamp of the created archive file).
	DestinationTimestamp *time.Time
```

internal/private/private.go
```diff
@@ -170,6 +171,9 @@ type CommitOptions struct {
 	// What “resolved” means is transport-specific.
 	// Transports which don’t support reporting resolved references can ignore the field; the generic copy code writes "nil" into the value.
 	ReportResolvedReference *types.ImageReference
+	// SourceDateEpoch, if set, and if applicable for destination, will override any creation/modified/access timestamps
+	// useful for oci-archive outputs
+	SourceDateEpoch *time.Time
```
mtrmac, line 176: (This is internal, so I’m not nearly as worried about the name or about word-smithing the comment. Maybe just `Timestamp` (with the purpose implied by the `CommitOptions` context + copy the comment from `copy.Options`?)

oci/archive/oci_transport_test.go
```diff
@@ -253,13 +256,38 @@ func TestReferenceNewImage(t *testing.T) {
...
+
+	numEntries := 0
+	tr := tar.NewReader(f)
+	for {
+		th, err := tr.Next()
+		if err == io.EOF {
+			break
+		}
+		assert.NoError(t, err)
+		assert.Equal(t, targetTime, th.ModTime) // access time and change time are usually ignored by Go, so just check mod time
```
mtrmac, line 285:
```suggestion
		assert.Equal(t, targetTime, th.ModTime) // access time and change time are ignored by Go’s tar.Writer unless the creator explicitly sets a non-default header format, so just check mod time
```

copy/copy.go
aeijdenberg: I think that looks good. I've changed as suggested and rebased/squashed.

Thanks for putting the thought in. Naming things is Hard.

internal/private/private.go
aeijdenberg: Done.

oci/archive/oci_transport_test.go
aeijdenberg: Done.

PR #2731, opened 2025-02-22
fix: handle error on Close() of writable objects

aeijdenberg:
For objects (particularly files) that we write to, we should to check the return value of `f.Close()` and pass error back to the caller if this fails.

I recently read https://www.joeshaw.org/dont-defer-close-on-writable-files/ and while working on another PR in this repo I noticed that some some writable files had the `defer f.Close()` pattern, so I did a grep to find similar calls, and for cases involving writeable file, changed the defer function to capture the result of calling `Close()`, and if the `err` value of the parent is `nil`, then set it for the caller.

In this manner errors on close will be returned to the caller, but won't mask other errors if an error was already set.

This adjusts those to one of the patterns in that article, e.g.:

```golang
func doThing() (err error) {
	outFile, err := os.Create(dst)
	if err != nil {
		return err
	}

	// since we are writing to this file, make sure we handle errors
	defer func() {
		closeErr := outFile.Close()
		if err == nil {
			err = closeErr
		}
	}()

        .....
}
```

### DIFF
--- a/ostree/ostree_dest.go
+++ b/ostree/ostree_dest.go
@@ -143,16 +143,24 @@ func (d *ostreeImageDestination) PutBlobWithOptions(ctx context.Context, stream 
 		return private.UploadedBlob{}, err
 	}
 
+	digester, stream := putblobdigest.DigestIfCanonicalUnknown(stream, inputInfo)
+
 	blobPath := filepath.Join(tmpDir, "content")
 	blobFile, err := os.Create(blobPath)
 	if err != nil {
 		return private.UploadedBlob{}, err
 	}
-	defer blobFile.Close()
-
-	digester, stream := putblobdigest.DigestIfCanonicalUnknown(stream, inputInfo)
-	// TODO: This can take quite some time, and should ideally be cancellable using ctx.Done().
-	size, err := io.Copy(blobFile, stream)
+	size, err := func() (_ int64, retErr error) { // A scope for defer
+		// since we are writing to this file, make sure we handle errors
+		defer func() {
+			closeErr := blobFile.Close()
+			if retErr == nil {
+				retErr = closeErr
+			}
+		}()
+		// TODO: This can take quite some time, and should ideally be cancellable using ctx.Done().
+		return io.Copy(blobFile, stream)
+	}()
 	if err != nil {
 		return private.UploadedBlob{}, err
 	}
@@ -247,9 +255,15 @@ func (d *ostreeImageDestination) ostreeCommit(repo *otbuiltin.Repo, branch strin
 	return err
 }
 
-func generateTarSplitMetadata(output *bytes.Buffer, file string) (digest.Digest, int64, error) {
+func generateTarSplitMetadata(output *bytes.Buffer, file string) (_ digest.Digest, _ int64, retErr error) {
 	mfz := pgzip.NewWriter(output)
-	defer mfz.Close()
+	// since we are writing to this, make sure we handle errors
+	defer func() {
+		closeErr := mfz.Close()
+		if retErr == nil {
+			retErr = closeErr
+		}
+	}()
 	metaPacker := storage.NewJSONPacker(mfz)
 
 	stream, err := os.OpenFile(file, os.O_RDONLY, 0)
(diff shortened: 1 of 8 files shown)


mtrmac (commented):
Thanks!

Just a quick note for now: This is definitely valuable; but I’ve been bitten by unintended named return value shadowing/overwriting several times in the past. Please use a clearly-unique name for the error value modified by `defer` (`retErr` is common in this codebase); for the other return values, it’s best not to add names to them them at all — use `_`.

Note to self: I didn’t actually review the changes.

aeijdenberg:
> Please use a clearly-unique name for the error value modified by `defer` (`retErr` is common in this codebase); for the other return values, it’s best not to add names to them them at all — use `_`.

Thanks, done as suggested, and squashed/rebased. I also removed the change to `pkg/blobcache/dest.go` - I'll have a go at that in a different PR as the entire method there looked to be quite unique in how it was doing its error handling flow.

storage/storage_dest.go
```diff
@@ -272,7 +272,13 @@ func (s *storageImageDestination) putBlobToPendingFile(stream io.Reader, blobinf
...
 		return private.UploadedBlob{}, fmt.Errorf("creating temporary file %q: %w", filename, err)
 	}
-	defer file.Close()
+	// since we are writing to this file, make sure we handle err on Close()
+	defer func() {
+		closeErr := file.Close()
+		if retErr == nil {
+			retErr = closeErr
+		}
+	}()
```
mtrmac, line 281: This function should exit before updating `s.lockProtected` and the like.

mtrmac (commented):
Thanks!

In some of the situations, it seems inconsistent to recognize and report an error to write to a file, but to update records as if the file were written successfully anyway: the close + error detection should probably happen a bit earlier.

(Compare other code with “// A scope for defer” comments.)

ostree/ostree_dest.go
```diff
@@ -148,8 +148,13 @@ func (d *ostreeImageDestination) PutBlobWithOptions(ctx context.Context, stream
 	if err != nil {
 		return private.UploadedBlob{}, err
 	}
-	defer blobFile.Close()
-
+	// since we are writing to this file, make sure we handle errors
```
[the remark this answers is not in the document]
aeijdenberg: Done.

storage/storage_dest.go
aeijdenberg: Done.

pkg/blobinfocache/sqlite/sqlite.go
```diff
@@ -82,12 +82,17 @@ func New(path string) (types.BlobInfoCache, error) {
...
 }
 
-func new2(path string) (*cache, error) {
+func new2(path string) (_ *cache, retErr error) {
 	db, err := rawOpen(path)
 	if err != nil {
 		return nil, fmt.Errorf("initializing blob info cache at %q: %w", path, err)
 	}
-	defer db.Close()
+	defer func() {
```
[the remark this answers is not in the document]
aeijdenberg: Done.

storage/storage_dest.go
```diff
@@ -257,7 +257,7 @@ func (s *storageImageDestination) PutBlobWithOptions(ctx context.Context, stream
 
 // putBlobToPendingFile implements ImageDestination.PutBlobWithOptions, storing stream into an on-disk file.
 // The caller must arrange the blob to be eventually committed using s.commitLayer().
-func (s *storageImageDestination) putBlobToPendingFile(stream io.Reader, blobinfo types.BlobInfo, options *private.PutBlobOptions) (private.UploadedBlob, error) {
+func (s *storageImageDestination) putBlobToPendingFile(stream io.Reader, blobinfo types.BlobInfo, options *private.PutBlobOptions) (_ private.UploadedBlob, retErr error) {
```
mtrmac, line 260: I think this `retErr` is now unused (and shadowed by the nested one) and can be dropped.

mtrmac (commented):
Thanks! LGTM otherwise.

storage/storage_dest.go
aeijdenberg: Thanks - I missed that one. Fixed.

mtrmac:
Can you rebase, please?

mtrmac (approved):

mtrmac:
Thanks again!

PR #2732, opened 2025-02-24
Microoptimize an API use per gopls modernize

mtrmac:


Luap99 (commented):
LGTM, interestingly this is the first time I came across Appendf() and searching our sources shows no matches so it seems basically nobody is using it.

mtrmac:
It only exists since Go 1.19. (And, while `go run golang.org/x/tools/gopls/internal/analysis/modernize/cmd/modernize@latest -test ./...` recommends this change, VS Code does not list it in the “problems” tab.)

PR #2734, opened 2025-02-25
support sigstore attachments for v2s2

alex-berger:
Currently we support sigstore attachments for OCI. This PR adds support for sigstore attachments for v2s2 (Docker Registry HTTP API V2 Schema 2 Manifest).

fixes https://github.com/containers/container-libs/issues/212

### DIFF
--- a/docker/docker_image_dest.go
+++ b/docker/docker_image_dest.go
@@ -13,6 +13,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"reflect"
 	"slices"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


docker/docker_image_dest.go
```diff
@@ -701,34 +702,49 @@ func (d *dockerImageDestination) putSignaturesToSigstoreAttachments(ctx context.
...
+			Annotations: blobInfo.Annotations,
+		}
 		// We don’t benefit from a real BlobInfoCache here because we never try to reuse/mount configs.
-		configBlob, err := d.c.getOCIDescriptorContents(ctx, d.ref, ociManifest.Config, iolimits.MaxConfigBodySize,
+		configBlob, err := d.c.getOCIDescriptorContents(ctx, d.ref, descriptor, iolimits.MaxConfigBodySize,
 			none.NoCache)
 		if err != nil {
 			return err
 		}
 		if err := json.Unmarshal(configBlob, &ociConfig); err != nil {
```
mtrmac, line 732: Unmarshaling v2s2 into OCI is not semantically correct (loses data).

docker/docker_image_dest.go
```diff
@@ -701,34 +702,49 @@ func (d *dockerImageDestination) putSignaturesToSigstoreAttachments(ctx context.
...
-			return fmt.Errorf("parsing sigstore attachment config %s in %s: %w", ociManifest.Config.Digest.String(),
+			return fmt.Errorf("parsing sigstore attachment config %s in %s: %w", blobInfo.Digest.String(),
 				d.ref.ref.Name(), err)
 		}
 	}
 
-	// To make sure we can safely append to the slices of ociManifest, without adding a remote dependency on the code that creates it.
-	ociManifest.Layers = slices.Clone(ociManifest.Layers)
+	// To make sure we can safely append to the slices of Manifest, without adding a remote dependency on the code that creates it.
+	switch v := genManifest.(type) {
```
mtrmac, line 739: No; hard-coding implementation types is a an invisible dependency, and a trap for long-term maintenance.

At the very least, this package can build a private interface. More likely, we would need an (internal-only, because we can’t add methods to a public interface without breaking API) extension of the `manifest.Manifest` interface which adds an “add a layer” operation. (Compare `internal/manifest.List{,Public}`).

docker/docker_image_dest.go
```diff
@@ -737,8 +753,14 @@ func (d *dockerImageDestination) putSignaturesToSigstoreAttachments(ctx context.
...
-		for _, layer := range ociManifest.Layers {
-			if layerMatchesSigstoreSignature(layer, mimeType, payloadBlob, annotations) {
+		for _, layer := range genManifest.LayerInfos() {
+			descriptor := imgspecv1.Descriptor{
+				MediaType:   layer.MediaType,
+				Digest:      layer.Digest,
+				Size:        layer.Size,
+				Annotations: layer.Annotations,
+			}
+			if layerMatchesSigstoreSignature(descriptor, mimeType, payloadBlob, annotations) {
```
mtrmac, line 763: This doesn’t work because schema2 manifests don’t have annotations at all.

docker/docker_image_dest.go
```diff
@@ -761,7 +783,17 @@ func (d *dockerImageDestination) putSignaturesToSigstoreAttachments(ctx context.
 			return err
 		}
 		sigDesc.Annotations = annotations
-		ociManifest.Layers = append(ociManifest.Layers, sigDesc)
+		switch v := genManifest.(type) {
+		case *manifest.OCI1:
+			v.Layers = append(v.Layers, sigDesc)
+		case *manifest.Schema2:
+			v.LayersDescriptors = append(v.LayersDescriptors, manifest.Schema2Descriptor{
```
mtrmac, line 790: This discards annotations (schema2 manifests don’t have annotations at all), and as a result, the signature is unusable.

docker/docker_image_dest.go
```diff
@@ -781,9 +813,20 @@ func (d *dockerImageDestination) putSignaturesToSigstoreAttachments(ctx context.
...
 		return err
 	}
-	ociManifest.Config = configDesc
 
-	manifestBlob, err := ociManifest.Serialize()
+	switch v := genManifest.(type) {
+	case *manifest.OCI1:
+		v.Config = configDesc
+	case *manifest.Schema2:
+		v.ConfigDescriptor = manifest.Schema2Descriptor{
```
mtrmac, line 821: - The config was uploaded above with a hard-coded MIME type
- Structurally, assuming that OCI and schema2 configs are completely substitutable is not something the transport should assume; such a no-op conversion belongs into the `image/` conversion code.

mtrmac (changes requested):
Thanks!

How did you test this? AFAICS this simply cannot work, when the schema2 definitions (as they exist in c/image) simply don’t have the “annotations” field which _actually stores the signature_.

Also, I think we’d need to build some amount of transport-agnostic manifest/image infrastructure to support this.

jankaluza:

PR #2742, opened 2025-02-28
Add warnings about direct use of ImageSource

mtrmac:
Cc: @Luap99 , I’ve noticed some problematic users recently. I’ll also audit our codebases.

### DIFF
--- a/image/unparsed.go
+++ b/image/unparsed.go
@@ -15,6 +15,9 @@ type UnparsedImage = image.UnparsedImage
 // UnparsedInstance returns a types.UnparsedImage implementation for (source, instanceDigest).
 // If instanceDigest is not nil, it contains a digest of the specific manifest instance to retrieve (when the primary manifest is a manifest list).
 //
+// This implementation of [types.UnparsedImage] ensures that [types.UnparsedImage.Manifest] validates the image
+// against instanceDigest if set, or, if not, a digest implied by src.Reference, if any.
+//
 // The UnparsedImage must not be used after the underlying ImageSource is Close()d.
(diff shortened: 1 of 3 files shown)


Luap99 (commented):
LGTM

mtrmac:
Switching to draft, per https://github.com/containers/buildah/pull/6014#pullrequestreview-2655024805 👍 
> The godoc for the `ImageSource` type mentions that validation of things is left up to the caller, but it's not clear to me that switching to using the `Unparsed...` family of types and functions was how we were intended to do it. Is it that `Image.Manifest()` verifies the manifest's digest while caching it, but `ImageSource.GetManifest()` doesn't?

- Update/review the documentation at the top level `ImageSource` … I must have been completely blind to that already existing
- Document what `UnparsedInstance` guarantees.

mtrmac:
Updated, PTAL.

image/unparsed.go
```diff
@@ -15,6 +15,9 @@ type UnparsedImage = image.UnparsedImage
 // UnparsedInstance returns a types.UnparsedImage implementation for (source, instanceDigest).
 // If instanceDigest is not nil, it contains a digest of the specific manifest instance to retrieve (when the primary manifest is a manifest list).
 //
+// This implementation of [types.UnparsedImage] ensures that [types.UnparsedImage.Manifest] validates the image
+// against instanceDigest if set, or, if not, a digest implied by src.Reference, if any.
```
mtrmac, line 19: The “implied by” is intentionally _a bit_ vague, there is #1049 and I’m not yet sure how to handle that. For now, we must provide this feature for at least registry references, and in that case the meaning is clear enough.

Luap99 (approved):
LGTM

PR #2746, opened 2025-02-28
Fix omitempty for PKI  struct

QiWang19:
Add `omitempty` annotations to PKI APIs


QiWang19:
@mtrmac Could you review? Sorry I just found out I wrote this bug.

rhatdan:
LGTM

mtrmac (approved):
Thanks!

This is a ~cleanup, it does not break anything serious to warrant backporting, does it?

QiWang19:
@mtrmac Yes, it does not break anything serious.

PR #2756, opened 2025-03-04
Migrate from github.com/xeipuuv/gojsonschema to github.com/santhosh-tekuri/jsonschema/v5

mtrmac:
github.com/xeipuuv/gojsonschema has last had a commit in Oct 2020.

So, follow what github.com/opencontainers/image-spec did, and move to the same library that project uses.

Note that github.com/santhosh-tekuri/jsonschema/v6 already exists, but the goal is to decrease the dependency surface (to the little extent this matters, it's at most a test dependency anyway), so we prioritize using the older version.


rhatdan:
LGTM

PR #2764, opened 2025-03-06
Add proxy for dockerClient

warjiang:
dockerClient will use http client to fetch manifest、blob and do some network request. Golang build-in http library  will use `HTTP_PROXY`、`HTTPS_PROXY` environment variables to config proxy for network. But the  environment variables cannot make flexible control of proxy behaviours. So I want to declare proxy configuration explicitly.

### DIFF
--- a/docker/docker_client.go
+++ b/docker/docker_client.go
@@ -907,6 +907,10 @@ func (c *dockerClient) detectPropertiesHelper(ctx context.Context) error {
 	}
 	tr := tlsclientconfig.NewTransport()
 	tr.TLSClientConfig = c.tlsClientConfig
+	// if set DockerProxyURL explicitly, use the DockerProxyURL instead of system proxy
+	if c.sys != nil && c.sys.DockerProxyURL != nil {
+		tr.Proxy = http.ProxyURL(c.sys.DockerProxyURL)
+	}
 	c.client = &http.Client{Transport: tr}
 
 	ping := func(scheme string) error {
--- a/types/types.go
+++ b/types/types.go
@@ -3,6 +3,7 @@ package types
 import (
 	"context"
 	"io"
+	"net/url"
 	"time"
 
 	"github.com/containers/image/v5/docker/reference"
@@ -655,6 +656,8 @@ type SystemContext struct {
 	// Note that this requires writing blobs to temporary files, and takes more time than the default behavior,
 	// when the digest for a blob is unknown.
 	DockerRegistryPushPrecomputeDigests bool
+	// DockerProxyURL specifies proxy configuration schema (like socks5://username:password@ip:port)
+	DockerProxyURL *url.URL
 
 	// === docker/daemon.Transport overrides ===
 	// A directory containing a CA certificate (ending with ".crt"),


docker/docker_client.go
```diff
@@ -900,13 +900,18 @@ func newBearerTokenFromHTTPResponseBody(res *http.Response) (*bearerToken, error
...
+		proxyURL = c.sys.ProxyURL
+		// We overwrite the TLS clients `InsecureSkipVerify` only if explicitly
+		// specified by the system context
+		if c.sys.DockerInsecureSkipTLSVerify != types.OptionalBoolUndefined {
+			c.tlsClientConfig.InsecureSkipVerify = c.sys.DockerInsecureSkipTLSVerify == types.OptionalBoolTrue
+		}
 	}
 	tr := tlsclientconfig.NewTransport()
 	tr.TLSClientConfig = c.tlsClientConfig
+	tr.Proxy = http.ProxyURL(proxyURL)
```
mtrmac, line 914: `tr` already contains a proxy configuration. We can’t outright start _ignoring_ the environment, that would break existing users.

I.e. the proxy field can only be written to if `c.sys.ProxyURL` is actually set.

types/types.go
```diff
@@ -676,6 +677,8 @@ type SystemContext struct {
 	CompressionFormat *compression.Algorithm
 	// CompressionLevel specifies what compression level is used
 	CompressionLevel *int
+	// ProxyURL specifies proxy configuration schema (like socks5://username:password@ip:port)
+	ProxyURL *url.URL
```
mtrmac, line 681: - Please put this into the “`docker.Transport` overrides” section.
- … and the name should be specific to that transport, maybe `DockerProxyURL`.

mtrmac (changes requested):
Thanks!

docker/docker_client.go
warjiang: thanks for you suggesion, add extra condition check. If `c.sys.ProxyURL` not set, dockerClient will follow the default behaviour. If `c.sys.ProxyURL` is set, dockerClient will the proxy configuration.

types/types.go
warjiang: agree

docker/docker_client.go
```diff
@@ -900,13 +900,20 @@ func newBearerTokenFromHTTPResponseBody(res *http.Response) (*bearerToken, error
...
+		// specified by the system context
+		if c.sys.DockerInsecureSkipTLSVerify != types.OptionalBoolUndefined {
+			c.tlsClientConfig.InsecureSkipVerify = c.sys.DockerInsecureSkipTLSVerify == types.OptionalBoolTrue
+		}
 	}
 	tr := tlsclientconfig.NewTransport()
 	tr.TLSClientConfig = c.tlsClientConfig
+	if proxyURL != nil {
+		tr.Proxy = http.ProxyURL(proxyURL)
+	}
```
mtrmac, line 916: Non-blocking: I think removing the `proxyURL` variable, and having two separate `if c.sys != nil && c.sys…` sections, would be easier to maintain over time (as the two parts of the code may be separated by more code aded in the future.)

mtrmac (commented):
Looks good, please squash the two commits.

docker/docker_client.go
warjiang: thanks for your suggestion~

warjiang:
> Looks good, please squash the two commits.

done~

warjiang:
it seems that some task failed.

image

Maybe it's problem of ci machine, is there any way to re-run the test case?

mtrmac (approved):
Thanks!

The CI failure should be resolved by https://github.com/containers/image/pull/2769 .

warjiang:
ci passed, let's move forward 🎉

mtrmac (approved):
Thanks again!

PR #2765, opened 2025-03-07
[release-5.34] Bump c/storage to v1.57.2, c/image to v5.34.2

TomSweeneyRedHat:
Bump c/storage to v1.57.2 to force idmap private
propagation per https://github.com/containers/storage/pull/2269

Then c/image to v5.34.2

This is targeted for RHEL 9.6/10.0 ZeroDay.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/BurntSushi/toml v1.4.0
 	github.com/containers/libtrust v0.0.0-20230121012942-c1716e8a8d01
 	github.com/containers/ocicrypt v1.2.1
-	github.com/containers/storage v1.57.1
+	github.com/containers/storage v1.57.2
 	github.com/cyberphone/json-canonicalization v0.0.0-20231217050601-ba74d44ecf5f
 	github.com/distribution/reference v0.6.0
 	github.com/docker/cli v27.5.1+incompatible
--- a/go.sum
+++ b/go.sum
@@ -56,8 +56,8 @@ github.com/containers/libtrust v0.0.0-20230121012942-c1716e8a8d01 h1:Qzk5C6cYgle
 github.com/containers/libtrust v0.0.0-20230121012942-c1716e8a8d01/go.mod h1:9rfv8iPl1ZP7aqh9YA68wnZv2NUDbXdcdPHVz0pFbPY=
 github.com/containers/ocicrypt v1.2.1 h1:0qIOTT9DoYwcKmxSt8QJt+VzMY18onl9jUXsxpVhSmM=
 github.com/containers/ocicrypt v1.2.1/go.mod h1:aD0AAqfMp0MtwqWgHM1bUwe1anx0VazI108CRrSKINQ=
-github.com/containers/storage v1.57.1 h1:hKPoFsuBcB3qTzBxa4IFpZMRzUuL5Xhv/BE44W0XHx8=
-github.com/containers/storage v1.57.1/go.mod h1:i/Hb4lu7YgFr9G0K6BMjqW0BLJO1sFsnWQwj2UoWCUM=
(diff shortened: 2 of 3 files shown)


Luap99:
Can we not do this? I don't understand the purpose of this bump since there are no changes in c/image.

There is no need to strictly bump c/image and c/common just because there is a new c/storage unless there are actual changes in them. We can directly bump c/storage in podman, buildah, skopeo which works the same.

I am saying this because each of these releases puts extra burden on us, not just the extra release PR, the new version then triggers downgrades per go modules versions until we merge back branches which is annoying to deal with.
i.e. https://github.com/containers/image/pull/2755 and https://github.com/containers/common/pull/2348

mtrmac (commented):
The mechanics of this PR LGTM.

As for whether c/image should have a tag, I don’t feel strongly (and I think we might have had such a discussion earlier, I’m not sure).

I’ll leave this open to give the discussion space, but feel free to merge if that’s the outcome.

Luap99:
Yeah to be clear I agree that this is totally fine to do.

But I think we can safe us some work by not doing it as I don't see this doing anything beneficial either.

warjiang:
Maybe the PR is needed, and some code should also be updated.

As the containers/skopeo has already upgrade the `github.com/containers/storage` to `v1.57.2`, ci which required building skopeo binary will failed:
image

in the process of building skopeo, it will use go mod to replace the image/v5 in the vendor to the source code of `containers/image`, and `containers/image` seems uses higher version of `github.com/containers/storage` lib:
image
`fileutils.ReflinkOrCopy` doesn't exist is `github.com/containers/storage@v1.57.2`. 

for the older ci, it works, because the version in the skopeo is `v1.57.1`, for the c/image lib, it requires `v1.57.2-0.20250228100055-700b765b2111`, v1.57.2 is higher than v1.57.1, so it will try to download and use the `v1.57.2-0.20250228100055-700b765b2111` of c/storage. But if the skopeo upgrade c/storage to v1.57.2, go mod will now think v1.57.2 is latest, and will not download `v1.57.2-0.20250228100055-700b765b2111`, so the ci of building skopeo binary will be failed.

warjiang:
during the process of finding the problem, the way that contains related lib or bin is a little bit heavy, sometimes chainable release will introduce extra work, maybe [`go workspace`](https://go.dev/doc/tutorial/workspaces) will be suitable for the current scenario.

Luap99:
> for the older CI, it works because the required version in the containers/skopeo is v1.57.1, for the c/image lib, it requires v1.57.2-0.20250228100055-700b765b2111, v1.57.2 is higher than v1.57.1, so it will try to download and use the v1.57.2-0.20250228100055-700b765b2111 of c/storage. But if the skopeo upgrade c/storage to v1.57.2, go mod will think v1.57.2 is latest, and will not download v1.57.2-0.20250228100055-700b765b2111, so the ci of building skopeo binary will be failed.

We are aware of how this works, the tag here does not contribute to any solution though it makes things worse because now c/image 5.34.2 would be newer than its main branch as well.

The only way as I already have written above is to merge back each tag after the release into main

> I am saying this because each of these releases puts extra burden on us, not just the extra release PR, the new version then triggers downgrades per go modules versions until we merge back branches which is annoying to deal with.
i.e. https://github.com/containers/image/pull/2755 and https://github.com/containers/common/pull/2348

And now https://github.com/containers/storage/pull/2276 once that is merged c/storage@main will be update over 1.57.2

TomSweeneyRedHat:
Although not needed from a Go compatibility perspective, before doing a release, like we are now for the RHEL Zero Day release, I very much prefer aligning the version numbers up and down the stack.  It might be my OCD silliness, but I find it easier a year or more after a release to try and find what was where and what needs to bump.  When I see c/storage 1.57.3 in Buildah, but 1.57.1 in image, I always end up digging more and second-guessing.

Yes a bit more grind for the team, but generally a very quick review.  If we could push this through, I would appreciate it.

mtrmac:
Let’s stay consistent, then.

PR #2769, opened 2025-03-10
Update c/storage after https://github.com/containers/storage/pull/2276

mtrmac:
... so that the referenced version is larger than 1.57.2 required by Skopeo, and we don't downgrade and lose a required API.

FYI @Luap99


Luap99 (approved):
LGTM but we might as well wait for the v5.34.2 tag to be pushed and to the merge back here also.

mtrmac:
Good point, let’s wait.

Luap99 (approved):
LGTM

PR #2781, opened 2025-03-13
Fix layout.List returns error when no index file.

bitoku:
When no images have been pulled and no index file in the directory, layout.List returns error. This PR fixes the error.

### DIFF
--- a/oci/layout/reader.go
+++ b/oci/layout/reader.go
@@ -2,6 +2,7 @@ package layout
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -24,6 +25,9 @@ func List(dir string) ([]ListResult, error) {
 	var res []ListResult
 
 	indexJSON, err := os.ReadFile(filepath.Join(dir, imgspecv1.ImageIndexFile))
+	if errors.Is(err, os.ErrNotExist) {
+		return []ListResult{}, nil
+	}
 	if err != nil {
 		return nil, err
 	}


mtrmac (changes requested):
Thanks for the PR.

This does not look correct: if there is not an index file, in what sense is the input an OCI layout? Compare https://github.com/opencontainers/image-spec/blob/main/image-layout.md#content :
> `index.json` file: It MUST exist

I think it’s quite possible for users/callers of this API to make a typo, and I think they should be told that what they are asking about does not exist at all and that they might want to check for typos.

Even if the directory contained, for example, an `ImageLayoutFile`, identifying the directory as valid: when the index is missing, is that because it should be there and the directory is corrupt, or because the directory is somehow uninitialized? At least per the spec, it is corrupt.

In some sense, this code doesn’t have enough information about the caller’s intent about handling empty/missing directory structures, and that might invite extending the API to express that caller’s intent. But, in this case, I think the caller can deal with such a situation (the layout directory is missing, or completely empty) just as well itself, without extending the API.

bitoku:
Thank you for the explanation! I understand why the change is invalid.

> I think the caller can deal with such a situation (the layout directory is missing, or completely empty) just as well itself, without extending the API.

I have one question about it. Is there any function or something that can initialize the directory as a oci layout directory?

mtrmac:
Not currently ; the format / version files are created as a side effect of writing the first image (to an empty directory).

Adding an API to do that might well make sense.

@baude WDYT? Maybe some variant of https://github.com/containers/podman/blob/d1d8f3334fea2394a17958ab9f04429d865789e4/pkg/libartifact/store/store.go#L555 could be moved here.

baude:
i would not object to that.

PR #2794, opened 2025-03-19
storage: lookup top layer in mapped top layers too

giuseppe:
when looking up for a different toplayer, consider also the mapped top layers as a mapped layer could become top layer and the previous top layer moved to the mapped layers

Related-to: https://github.com/containers/storage/pull/2285.

### DIFF
--- a/storage/storage_dest.go
+++ b/storage/storage_dest.go
@@ -1571,7 +1571,7 @@ func (s *storageImageDestination) CommitWithOptions(ctx context.Context, options
 		if err != nil {
 			return fmt.Errorf("reading image %q: %w", intendedID, err)
 		}
-		if img.TopLayer != lastLayer {
+		if img.TopLayer != lastLayer && !slices.Contains(img.MappedTopLayers, lastLayer) {
 			logrus.Debugf("error creating image: image with ID %q exists, but uses different layers", intendedID)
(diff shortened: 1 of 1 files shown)


mtrmac (commented):
ACK — this is only relevant _after_ https://github.com/containers/storage/pull/2285, isn’t it? Without that, `.TopLayer` is exactly a layer matching the ChainID (as determined by c/image/v5/storage).

giuseppe:
yes I think so, could we end up in this situation if we pull the image through `podman pull --uidmap ...` with vfs?

mtrmac:
If I’m reading things correctly, we currently only create exactly one layer per ChainID (but, possibly, with several `UpdateLayerIDMap` calls, with a sequence of `podman pull --uidmap` with different maps, reusing the parent layers). (And we ultimately rely on the container’s `…TopLayerForMapping` code to get a usable container.) And as c/image sets `TopLayer`, it is immutable afterwards. This c/image logic does not depend on the graph driver, so VFS should work the same way.

With the proposed PR in c/storage (fixing up `TopLayer` to be the neutral one), we do need this change.

mtrmac:
@giuseppe did you want to close this? I was thinking leave it open until https://github.com/containers/storage/pull/2285 is proven / reviewed.

giuseppe:
I am trying a different approach now.  Switching the TopLayer causes some other issues in buildah too.  Seems too much pain just to support vfs :/

PR #2795, opened 2025-03-19
Correctly compute DiffID, and try to set the correct MIME type, for non-gzip input to `tarball:`

mtrmac:
See https://github.com/containers/podman/issues/25593 .

This is an _absolutely untested_ demonstration of what probably needs fixing.

### DIFF
--- a/tarball/tarball_src.go
+++ b/tarball/tarball_src.go
@@ -89,24 +89,26 @@ func (r *tarballReference) NewImageSource(ctx context.Context, sys *types.System
 		var layerType string
 		var diffIDdigester digest.Digester
 		// Set up to digest the file after we maybe decompress it.
-		uncompressed, err := pgzip.NewReader(reader)
-		if err == nil {
-			// It is compressed, so the diffID is the digest of the uncompressed version
-			diffIDdigester = digest.Canonical.Digester()
-			reader = io.TeeReader(uncompressed, diffIDdigester.Hash())
-			layerType = imgspecv1.MediaTypeImageLayerGzip
-		} else {
-			// It is not compressed, so the diffID and the blobID are going to be the same
-			diffIDdigester = blobIDdigester
-			layerType = imgspecv1.MediaTypeImageLayer
-			uncompressed = nil
-		}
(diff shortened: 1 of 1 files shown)


tarball/tarball_src.go
```diff
@@ -82,31 +83,47 @@ func (r *tarballReference) NewImageSource(ctx context.Context, sys *types.System
...
-		if _, err := io.Copy(io.Discard, reader); err != nil {
-			return nil, fmt.Errorf("error reading %q: %w", filename, err)
-		}
-		if uncompressed != nil {
-			uncompressed.Close()
+		var layerType string
+		var diffIDdigester digest.Digester
+		// If necessary, digest the file after we decompress it.
+		if err := func() error { // A scope for defer
+			format, decompressor, reader, err := compression.DetectCompressionFormat(reader)
```
mtrmac, line 94: This uses c/image/v5/pkg/compression, while the storage destination that primarily consumes this data in `podman import` uses c/storage/pkg/archive.DecompressStream`.

Right now the two support the same compression formats, but the implementations differ a bit (c/storage can shell out to subprocesses), OTOH the c/image API is mildly easier to deal with.

Maybe, for long-term consistency, this should use the c/storage implementation. Or maybe it really doesn’t matter, the source/destination transports are generally supposed to be independent, after all.

Luap99 (commented):
LGTM, tested in https://github.com/containers/podman/issues/25593#issuecomment-2737499606

Luap99:
nit: commit message 1 has a typo, i.e Shoult

mtrmac:
Thanks, fixed. (Also improved one comment.)

cgwalters (approved):

mtrmac:
I have now tested manually, that after `podman import`, `podman save` / `podman push` create reasonable manifest, and that consuming those outputs also works.

mtrmac:
Note to self: After this merges and Podman is updated, https://github.com/containers/podman/pull/25465 should be reverted.

Luap99 (approved):
LGTM

TomSweeneyRedHat:
Fixes: https://issues.redhat.com/browse/RHEL-85212

PR #2800, opened 2025-03-21
[crio-1.31] Fix layer reuse in schema1 images

saschagrunert:
reused.Digest is not always blobDigest, it might be uncompressedDigest; but we must have a blobDiffIDs entry for reused.Digest.


saschagrunert:
@bitoku @sohankunkerkar @haircommander PTAL

bitoku (approved):
cherry-pick from https://github.com/containers/image/pull/2613/commits

PR #2801, opened 2025-03-21
docker: export UnexpectedHTTPStatusError

Luap99:
Some callers might like to make decisions based on the http server error that was returned.

In particular we would like c/common/pkg/retry to match this error so it can retry image pulls/pushes on 5XX errors as they seems to be a quite common problem[1].

[1] https://github.com/containers/common/issues/2299


docker/distribution_error.go
```diff
@@ -30,14 +30,18 @@ import (
...
 var errNoErrorsInBody = errors.New("no error details found in HTTP response body")
 
-// unexpectedHTTPStatusError is returned when an unexpected HTTP status is
+// UnexpectedHTTPStatusError is returned when an unexpected HTTP status is
 // returned when making a registry api call.
-type unexpectedHTTPStatusError struct {
-	Status string
+type UnexpectedHTTPStatusError struct {
+	// StatusCode code as returned from the server, public so callers can
+	// match the exact code to make certain decisions if needed.
```
mtrmac, line 37: (I’m not sure the “public …” part is relevant to external consumers, but it doesn’t hurt.)

mtrmac (changes requested):
Thanks! The API makes sense.

To be universal: There is also `getExternalBlob` (though I’m unsure what exactly to do there), `getOneSignature`, and `httpResponseToError`. At least the last one is invoked on ~every operation, as one of the first requests, so it might frequently be the one that actually reports a failure.

docker/distribution_error.go
```diff
@@ -30,14 +30,18 @@ import (
...
+	StatusCode int
+	// status text as displayed in the error message, not exposed as callers should match the number.
+	status string
 }
 
-func (e *unexpectedHTTPStatusError) Error() string {
-	return fmt.Sprintf("received unexpected HTTP status: %s", e.Status)
+func (e *UnexpectedHTTPStatusError) Error() string {
```
mtrmac, line 43: … actually, this can be a value type, no need for the `error` interface box to hold another pointer. (Also update the places that return this error, and the consumer in `pkg/retry`)

mtrmac: … and we can always add pointers inside the public-API structure.

Luap99: Yeah that is the thing I never understand, the go std lib doesn't seem to be consistent on it either, did a quick look with `grep -r ") Error() string" /usr/lib/golang/src/`

I really don't mind either way, I am find to change that to pass by value.

mtrmac (commented):
LGTM, please merge (or ping me) whenever appropriate.

(I wouldn’t mind a `newUnexpectedHTTPStatusError(*http.Response)` helper, but this is simple enough, and I don’t expect many more instances to be added in the future.)

docker/distribution_error.go
mtrmac: > no need for the `error` interface box to hold another pointer.

I was curious, so: looking at the implementation, both types end up with effectively the same _memory structure_: the `error` interface value is (pointer to value, `itab` for how the value implements the interface). For a pointer error, this is directly the language-visible pointer; for a struct error, this is a pointer to a separately-allocated language-visible struct.

There are some differences in code:
- Curiously, allocating+returning the error ends up, AFAICS, with ~exactly the same layout but using a pointer type avoids one copy (with a pointer, the fields are written directly to a heap object; with a `struct`, the value is first formed on the stack and then copied to the heap). I think this is a missing optimization.
- Method calls on the interface truly differ: method calls on a pointer get a pointer (and indirect inside the method), method calls on a structure get a struct in registers and/or on stack. So, for the interface, there is a thunk that turns a (call to interface method, interface-implementing pointer) into (make a copy of the value on stack, call the struct’s implementation). So, calling interface methods on large structs could be quite a bit more expensive than calling interface methods on pointers.

Ultimately, I think making this a value type where it is clear that mutating the fields doesn’t affect anyone else is simpler semantics, and the performance impact is not worth worrying about, compared to invoking the full HTTP + TLS stack.

mtrmac (approved):

PR #2810, opened 2025-03-27
Merge the v5.34.3 tag into the main branch

mtrmac:
… so that users who vendor a commit from the `main` branch don’t have to downgrade.

@TomSweeneyRedHat @Luap99 PTAL


Luap99 (approved):
LGTM, I wonder if you should have a bot do these, of course there is always a merge conflict but we can tell it to always use the version from main as merge strategy I think.

TomSweeneyRedHat:
I'm late to the party, but LGTM, FWIW.

PR #2815, opened 2025-04-03
Run golangci-lint with extra tags, fix found issues

kolyshkin:
The idea is to run golangci-lint with flags that are not usually set, to catch some potential issue that are not caught otherwise. Surprisingly, it's not much (see the first 4 commits).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -100,7 +100,7 @@ validate: lint
 lint: EXTRA_TAGS=containers_image_docker_daemon_stub,containers_image_fulcio_stub,containers_image_openpgp,containers_image_rekor_stub,containers_image_storage_stub
 lint:
 	$(GOBIN)/golangci-lint run --build-tags "$(BUILDTAGS)"
-	$(GOBIN)/golangci-lint run --build-tags "$(EXTRA_TAGS)"
+	time $(GOBIN)/golangci-lint run --build-tags "$(EXTRA_TAGS)"
 
 # When this is running in CI, it will only check the CI commit range
 .PHONY: .gitvalidation
--- a/Makefile
+++ b/Makefile
@@ -97,8 +97,10 @@ validate: lint
 	@BUILDTAGS="$(BUILDTAGS)" hack/validate.sh
 
 .PHONY: lint
+lint: EXTRA_TAGS=containers_image_docker_daemon_stub,containers_image_fulcio_stub,containers_image_openpgp,containers_image_rekor_stub,containers_image_storage_stub
 lint:
 	$(GOBIN)/golangci-lint run --build-tags "$(BUILDTAGS)"
+	$(GOBIN)/golangci-lint run --build-tags "$(EXTRA_TAGS)"
 
 # When this is running in CI, it will only check the CI commit range
 .PHONY: .gitvalidation
--- a/signature/fulcio_cert.go
+++ b/signature/fulcio_cert.go
@@ -185,21 +185,6 @@ func (f *fulcioTrustRoot) verifyFulcioCertificateAtTime(relevantTime time.Time, 
 	return untrustedCertificate.PublicKey, nil
 }
 
-func parseLeafCertFromPEM(untrustedCertificateBytes []byte) (*x509.Certificate, error) {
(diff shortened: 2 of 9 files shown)


signature/mechanism_openpgp.go
```diff
@@ -147,7 +147,7 @@ func (m *openpgpSigningMechanism) Verify(unverifiedSignature []byte) (contents [
 		return nil, "", err
 	}
 	if md.SignatureError != nil {
-		return nil, "", fmt.Errorf("signature error: %v", md.SignatureError)
+		return nil, "", fmt.Errorf("signature error: %w", md.SignatureError)
```
mtrmac, line 150: \*shrug\* this doesn’t _hurt_, but it’s also not _useful_: callers should not be assuming the `openpgp` build tag, so they should not be matching on the returned error.

signature/fulcio_cert.go
```diff
@@ -185,21 +185,6 @@ func (f *fulcioTrustRoot) verifyFulcioCertificateAtTime(relevantTime time.Time,
 	return untrustedCertificate.PublicKey, nil
 }
 
-func parseLeafCertFromPEM(untrustedCertificateBytes []byte) (*x509.Certificate, error) {
```
mtrmac, line 188: Move also `TestParseLeafCertsFromPEM` (and `s/Certs/Cert/`) then, please.

Makefile
```diff
@@ -97,8 +97,13 @@ validate: lint
 	@BUILDTAGS="$(BUILDTAGS)" hack/validate.sh
 
 .PHONY: lint
+lint: LINTER_TAGS = containers_image_openpgp containers_image_docker_daemon_stub containers_image_storage_stub containers_image_fulcio_stub containers_image_rekor_stub
```
mtrmac, line 100: Can you sort the tags alphabetically, please?

mtrmac (commented):
Thanks!

The cleanups are nice.

The extra 4.5 minutes of CI time, on a `validate` task which is a serial prerequisite to running almost all other tests, are rather unpleasant, though. I’d prefer that to, at least, happen in a separate Cirrus task that runs concurrently with the other slow tasks — or maybe it’s just not worth the extra CI cost to run these all the time.

signature/fulcio_cert.go
kolyshkin: done

Makefile
kolyshkin: Done

signature/mechanism_openpgp.go
kolyshkin: I agree that it's not an improvement, except maybe for code uniformity (so the rule "use %w for error wrapping" always applies).

kolyshkin:
> The extra 4.5 minutes of CI time

@mtrmac It's about 40-50 extra seconds (from 4:00 minutes to 4:50 minutes), but I can split it out to a separate task if you like (maybe also adding `--tests=false`).

Having looked at it more, I think the best thing to do is to modify skopeo test (which is the longest step and takes ~15 minutes) to not wait on validate. This alone will actually make tests finish 4-5 minutes faster; here's a PR: https://github.com/containers/image/pull/2830

kolyshkin:
> One other thing we can do to make linting faster is something like [containers/podman#25772](https://github.com/containers/podman/pull/25772). Let me give it a try.

PR #2832

kolyshkin:
So, I guess, this one together with #2832 will result in the same/similar `validate` task timings.

Having said that, I can move validate to github actions so it won't hold other tasks. WDYT @mtrmac?

mtrmac:
Thanks! I think something like #2832 is clearly desirable.

And that _might_ make the extra time so small to not be worth worrying about. (IIRC it was extra 4.5 minutes originally when testing each tag separately in a loop.)

Although, in principle, it’s better to have a small, fast, cheap first test to rule out obviously non-viable PRs, and splitting seems conceptually closer to the ideal, at some point it’s just not worth the complexity (and overhead costs for scheduling+starting+separate caching).

jankaluza:

PR #2821, opened 2025-04-08
INCOMPATIBLE: Remove the implementation of the ostree transport

mtrmac:
Currently, ostree does not compile:

> ```
> ostree/ostree_dest.go:115:28: cannot use d (variable of type *ostreeImageDestination) as private.ImageDestinationInternalOnly value in argument to impl.AddCompat: *ostreeImageDestination does not implement private.ImageDestinationInternalOnly (missing method NoteOriginalOCIConfig) (typecheck)
> 	d.Compat = impl.AddCompat(d)
> ```

This has been broken since b941c6bf415cec5158dbbb822c26650420ad11bc (Nov 18 2024, > 4 months).

Previously:
- 807381f448c164bb81649ea1ad770609157efd9d (broken almost 4 months),
- 2e33bf7d272d97fba72e25964bc478eed8b028dd (broken > 7 months)

Overall, it seems extremely likely that noone is using the `containers_image_ostree` build tag (off by default since March 2019).

So, finally, give up, and delete the code. Most importantly, this allows us to delete the > 3-year-old-frozen dependency, and risky CGo code.

---

Cc: @giuseppe , @mheon , @rhatdan (general RFC / awareness), @joonas-fi (previously reporting a build failure in #1917 ).


cgwalters (approved):
Yes, the replacement for ostree is composefs.

giuseppe (approved):
this was needed for system containers to pull an image to an ostree repo.  We don't need it anymore.

LGTM

vrothberg (approved):
LGTM

ostree/ostree_transport.go
```diff
@@ -1,242 +1,24 @@
...
-var Transport = ostreeTransport{}
-
-type ostreeTransport struct{}
-
-func (t ostreeTransport) Name() string {
-	return "ostree"
-}
+//
+// Deprecated: The ostree: implementation has been removed, and any attempt to use this transport fails.
+var Transport = transports.NewStubTransport("ostree")
```
Luap99, line 13: What is the point of keeping this? If we don't support any ostree code would not be better to outright cause compile failures for all users? Why would we move all errors to runtime? Sure it break API but removing the actual function already breaks any potential user so I would prefer a proper clean break no?

mtrmac: I don’t feel strongly about this.

I think there is, in principle, _some_ value to keeping the transport available, because code matching transports,, similar to https://github.com/containers/common/blob/f71a7a6eca1d1c2846e00b2cda013c9b37af7cc5/libimage/pull.go#L237 , can continue to compile for a while; keeping the symbols around reduces the urgency of updating it.

OTOH it is almost certainly a purely hypothetical situation.

Luap99: Right but I wonder if this not ends up causing more technical debt that way. If nobody updates the places where it is referenced they just end up with defunct code which IMO is worse.

Maybe I do to much rust coding but I like to have changes fail to compile and not find out only at runtime in tests (in the best case) or production (at the worst case) that something changed in an incompatible way.

I don't see any podman references to this the ostree package so I am not that worried about the upgrade path. (I know there are more c/image users so that is an selfish view)
And if things failed to compile already removing it would not make it worse? I guess there is the case about users of the already stub implementation?

mtrmac:
Updated to outright remove the `ostree` subpackage, per https://github.com/containers/image/pull/2821#discussion_r2037023561 (someone _might_ have been using it just for the names, but they would have encountered failures to compile with any recent version).

Also, `ostree` will no longer be listed in CLI help.

PR #2823, opened 2025-04-10
fix two github action workflow issues

Luap99:
see both commits, the first one is obvious, the second one I like to see if that works on this repo before making PRs against all the other that use this cirrus rerun action with the same change

.github: check_cirrus_cron work around github bug

So I wondered why our email workflow only reported things for podman...

It seems the secrets: inherit is broken and no longer working, I see all
jobs on all repos failing with:

Error when evaluating 'secrets'. .github/workflows/check_cirrus_cron.yml (Line: 19, Col: 11): Secret SECRET_CIRRUS_API_KEY is required, but not provided while calling.

This makes no sense to me I doubled checked the names, nothing change
don our side and it is consitent for all projects. Interestingly this
same thing passed on March 10 and 11 (on all repos) but failed before
and after this as well.

Per[1] we are not alone, anyway let's try to get this working again even
if it means more duplication.

[1] https://github.com/actions/runner/issues/2709


TomSweeneyRedHat:
LGTM

.github/workflows/check_cirrus_cron.yml
```diff
@@ -17,4 +17,14 @@ jobs:
   # Ref: https://docs.github.com/en/actions/using-workflows/reusing-workflows
   call_cron_failures:
     uses: containers/podman/.github/workflows/check_cirrus_cron.yml@main
-    secrets: inherit
+    secrets:
+      SECRET_CIRRUS_API_KEY:
+        required : true
```
mtrmac, line 22: I don’t think this should work? Compare the different documentation/structure of  https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onworkflow_callsecretssecret_id vs. https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onworkflow_callsecretssecret_id .

Or https://docs.github.com/en/actions/sharing-automations/reusing-workflows#passing-inputs-and-secrets-to-a-reusable-workflow .

Luap99: Ack, I will have another look tomorrow. A day more less won't change anything

mtrmac (approved):
Thanks!

Luap99:
Triggered the action manually, it worked! Now you should have an email that shows errors on two branches

mtrmac:
Thanks, I did get that email. And I was thinking c/image is doing so well :)

Luap99:
Well that is what made me look into these things, we only got podman mails but no other repo a single flake (considering we disabled the auto rerun flake action just recently) so that couldn't be right to me and sure enough here we are.

PR #2830, opened 2025-04-16
ci: start skopeo tests earlier

kolyshkin:
Currently, skopeo tests (which are the most time consuming, taking 13-15 minutes) wait on validate test (which takes ~4 minutes) to finish. As a result, total time to finish all tests is about 20 minutes.

Let's start skopeo tests early to improve overall time.

Here's a pic to illustrate:

image

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -96,8 +96,6 @@ test_task:
 test_skopeo_task:
     alias: test_skopeo
     only_if: *not_docs
-    depends_on:
-        - validate
     gce_instance:
         image_project: libpod-218412
         zone: "us-central1-f"


kolyshkin:
This is "after" situation (the failure is probably unrelated):
image

So it's not worth it.

mtrmac:
Thanks! I don’t know why the two “Test” tasks would be delayed to run after the Skopeo ones — this looks as we were hitting a concurrency limit, but I think we are supposed to be running on a paid account.

Anyway, the current structure _seems_ to have been designed (except for `cross_task`???) to avoid running the full test battery if the PR is clearly unviable; starting the longest/most expensive tests before `validate` would lose most of the benefit of that.

But #2837 suggests there might be other avenues to significantly speed things up.

PR #2832, opened 2025-04-16
cirrus: save/restore golangci-lint cache for linux

kolyshkin:
This shaves off some time from golangci-lint run (unless either golangci-lint or go version changes).

The cache size is pretty small (a few MB).


mtrmac (commented):
Thanks!

This is interesting, and some experiments in https://github.com/containers/image/pull/2837 suggest that there might be much more.

… and there are potentially quite a few other things that could be cleaned up and sped up in the tests. That’s clearly not _this_ PR, but it might be something we could ask for help with, from a CI specialist.

jankaluza:

PR #2837, opened 2025-04-22
Testing Cirrus-CI caching

mtrmac:


.cirrus.yml
```diff
@@ -59,6 +59,18 @@ validate_task:
         image_name: ${DEBIAN_CACHE_IMAGE_NAME}
     env:
         HOME: "/root"  # default unset, needed by golangci-lint.
+    go_modules_cache: &go_modules_cache
+        fingerprint_script: cat go.sum
+        folder: $GOPATH/pkg/mod
+    go_build_cache: &go_build_cache
+        fingerprint_script: cat go.sum
+        folder: $GOCACHE
```
mtrmac, line 67: All of these need more to fingerprint. Compare https://github.com/actions/setup-go/blob/dca8468d37b6d090cde2c7b97b738a37134f5ffb/src/cache-restore.ts#L35 in GitHub.

And none of that triggers a cache rebuild on code changes, so eventually just changing code in the project will make the caches obsolete. Maybe the weekly Renovate PRs are enough to trigger `go.sum` changes?

.cirrus.yml
```diff
@@ -59,6 +59,18 @@ validate_task:
...
+    golangci-lint_cache:
+        folder: /root/.cache/golangci-lint
+        reupload_on_changes: true
+        fingerprint_script:
+            - go version
+            - grep GOLANGCI_LINT_VERSION Makefile | head -1
     script: |
         git remote update
```
mtrmac, line 75: Note to self: This seems mostly unnecessary, we just want the destination branch.

.cirrus.yml
```diff
@@ -59,6 +59,18 @@ validate_task:
         image_name: ${DEBIAN_CACHE_IMAGE_NAME}
     env:
         HOME: "/root"  # default unset, needed by golangci-lint.
+    go_modules_cache: &go_modules_cache
+        fingerprint_script: cat go.sum
+        folder: $GOPATH/pkg/mod
+    go_build_cache: &go_build_cache
```
mtrmac, line 65: This is a very big deal for `cross_task`. I would have expected this to also make a big difference for `validate_task`, but it doesn’t?!

A hypothesis to examine: would using a separate cache for the two tasks help? The current test uses a `cross`-created cache for `validate`; maybe the two differ enough in environment (some extra build tags??) that the caches don’t match?

mtrmac (commented):
@kolyshkin FYI: caching `GOMODCACHE` (like https://cirrus-ci.org/examples/#go suggests) helps more (`validate_task` goes from maybe 3:13 to 3:10), and caching `GOCACHE` (like https://github.com/actions/setup-go does) can be a much bigger deal (`cross_task` from 3:24 to 0:32).

.cirrus.yml
```diff
@@ -59,6 +59,18 @@ validate_task:
...
+        folder: $GOCACHE
+    golangci-lint_cache:
+        folder: /root/.cache/golangci-lint
+        reupload_on_changes: true
+        fingerprint_script:
+            - go version
+            - grep GOLANGCI_LINT_VERSION Makefile | head -1
     script: |
         git remote update
         make tools
```
mtrmac, line 76: All of this could be cached, or maybe built into the VM image.

.cirrus.yml
kolyshkin: > All of these need more to fingerprint.

In fact, `go version` already has GOOS/GOARCH (in addition to the version itself), so having something like
```yaml
fingerprint_script:
 - cat go.sum # or go.mod?
 - go version # version + GOOS/GOARCH
 ```
is maybe sufficient

mtrmac: Sounds good.

On `go.sum` vs. `go.mod`, I think the two should change in exactly the same commits — but, conceptually, `go.sum` would be updated on tag changes, while `go.mod` might not (e.g. after reverting to a previous dependency version, we might _hypothetically_ see exactly the same `go.mod` but different `go.sum` exposing that a tag was moved).

Ultimately, none of this should matter for _correctness_, hopefully the Go caching mechanism is capturing all relevant inputs in its cache design.

.cirrus.yml
mtrmac: Yes: a separate build cache brings `validate_task` down to 0:42 when nothing in the code changed (vs. 4:19 with no caching at all, 3:13 with lint caches, and 3:10 with module cache).

That is _definitely_ with doing.

.cirrus.yml
mtrmac: Separating the caches between `validate_task` and `cross_task` helps at least in the sense that we don’t download `git-validation` and its dependencies (and hopefully the compilation is also avoided).

Not the best we can do, but getting closer to the point of diminishing returns.

jankaluza:

PR #2844, opened 2025-04-29
vendor: update c/storage

giuseppe:
follow-up for https://github.com/containers/storage/pull/2312


mtrmac (commented):
Note to self: this will be correct with https://github.com/containers/storage/pull/2323 .

giuseppe:
rebased

storage/storage_dest.go
```diff
@@ -421,7 +421,7 @@ func (s *storageImageDestination) PutBlobPartial(ctx context.Context, chunkAcces
 		}
 	}()
 
-	differ, err := chunked.GetDiffer(ctx, s.imageRef.transport.store, srcInfo.Digest, srcInfo.Size, srcInfo.Annotations, &fetcher)
+	differ, err := chunked.NewDiffer(ctx, s.imageRef.transport.store, srcInfo.Digest, srcInfo.Size, srcInfo.Annotations, &fetcher)
```
mtrmac, line 424: The part that ensured `differ.Close` was called (and that motivated breaking the API to `s/Get/New/` here) got lost?!

mtrmac (changes requested):
.

storage/storage_dest.go
giuseppe: sorry I missed the error path on `s.imageRef.transport.store.PrepareStagedLayer()`, because later on `CleanupStagedLayer` takes care of it.

Do you prefer to have the `differ.Close()` just there or through a new defer function?

mtrmac: https://github.com/containers/storage/pull/2323 now documents on `NewDiffer`  that
> The caller must call Close() on the returned Differ.

so I’d prefer doing exactly that — `defer` seems easiest.

mtrmac (approved):

mtrmac:
Thanks!

PR #2856, opened 2025-05-26
[release-5.29] Fix CI

mtrmac:
See individual commit messages for details, let’s see what it takes…


Luap99 (approved):
LGTM, although I cannot help myself but saying CI should not do runtime installs of latest. But well that is much larger scope of course.

PR #2874, opened 2025-06-22
Add support for validating signatures from PGP subkeys

drGrove:
This change should update the gpgme and openpgp implmenetations to support validating signatures made by signing subkeys. It will close an open issue that was referenced specifically in [podman](https://github.com/containers/podman/issues/16406), but should resolve this across a number of tools that use this library.


drGrove:
> There’s also a bit of awkward timing in that https://github.com/containers/image/pull/2569 will probably replace the default / recommended verification implementation.

I noticed this as I was finishing out the implementation but based on my understanding of the comments. It appears the GPGPME/Go OpenPGP will still _exist_, it just won't be the hot path. As the sequoia OpenPGP implementation has also been something that's still being worked though, having any support for subkeys IMO is ideal.

> https://github.com/containers/podman/issues/16406 contains a long discussion about expired (or revoked) keys, IIRC it’s fairly unclear what is the right thing to do for long-term artifacts like software, and with no trusted timestamps.

To my understanding, once a key is revoked or expired the signature is no longer valid. Regardless of when it was signed. This default to close IMO is a feature. Revocations are packets within the key just like anything else and generally the systems/users doing validation of OpenPGP signatures understand the implications of this. If a key is revoked and pushed to a keyserver, the user/machine has to pull down the updated packets for it to take effect within the verification pipeline. If you are aware that a revocation has been published but you still need to use a previously signed image it would probably be more ideal for you to mirror the image and sign it yourself than the much less secure option of not importing the updated key. This same problem exists is most signing schemes (key rotation).

mtrmac:
I think it would be fair for me to first actually read this PR at least once :)

signature/mechanism_openpgp.go
```diff
@@ -177,3 +178,101 @@ func (m *openpgpSigningMechanism) Verify(unverifiedSignature []byte) (contents [
...
+	}
+
+	// Find the entity with the expected fingerprint
+	var expectedEntity *openpgp.Entity
+	var signerEntity *openpgp.Entity
+
+	for _, entity := range m.keyring {
+		// Check if this entity's primary key matches the expected fingerprint
+		if entity.PrimaryKey != nil {
+			primaryFingerprint := strings.ToLower(fmt.Sprintf("%x", entity.PrimaryKey.Fingerprint))
```
mtrmac, line 200: `%x` is already lower-case.

signature/mechanism_openpgp.go
```diff
@@ -177,3 +178,101 @@ func (m *openpgpSigningMechanism) Verify(unverifiedSignature []byte) (contents [
...
+					subkeyFingerprint := strings.ToLower(fmt.Sprintf("%x", subkey.PublicKey.Fingerprint))
+					if subkeyFingerprint == signerFingerprint {
+						return true, nil
+					}
+				}
+			}
+		}
+	}
+
+	// Case 3: The expected key is a subkey, and the signer is the primary key of the same entity
```
mtrmac, line 249: Similarly to the GPG case, with `expectedFingerprint` matching a subkey, ignoring that and accepting any other subkey is surprising.

signature/mechanism_gpgme.go
```diff
@@ -204,3 +204,59 @@ func (m *gpgmeSigningMechanism) Verify(unverifiedSignature []byte) (contents []b
 func (m *gpgmeSigningMechanism) UntrustedSignatureContents(untrustedSignature []byte) (untrustedContents []byte, shortKeyIdentifier string, err error) {
 	return gpgUntrustedSignatureContents(untrustedSignature)
 }
+
+// isSubkeyOf checks if signerKeyIdentity is a subkey of expectedKeyIdentity
+func (m *gpgmeSigningMechanism) isSubkeyOf(signerKeyIdentity, expectedKeyIdentity string) (bool, error) {
```
mtrmac, line 209: Reading the GPGME implementation, yes, we’ll probably need to issue a `GetKey` of some kind to be able to check these things.

signature/mechanism_openpgp.go
```diff
@@ -177,3 +178,101 @@ func (m *openpgpSigningMechanism) Verify(unverifiedSignature []byte) (contents [
 func (m *openpgpSigningMechanism) UntrustedSignatureContents(untrustedSignature []byte) (untrustedContents []byte, shortKeyIdentifier string, err error) {
 	return gpgUntrustedSignatureContents(untrustedSignature)
 }
+
+// isSubkeyOf checks if signerKeyIdentity is a subkey of expectedKeyIdentity
+func (m *openpgpSigningMechanism) isSubkeyOf(signerKeyIdentity, expectedKeyIdentity string) (bool, error) {
```
mtrmac, line 183: It _seems_ that we can immediately read `md.SignedBy.Entity.PrimaryKey.Fingerprint` to get the primary key. I didn’t yet verify that it is always available.

mtrmac (commented):
Looking at the possible implementation approaches:

… Sequoia in #2876  is already implemented to return the primary key’s fingerprint as an outcome of verification.

signature/mechanism_gpgme.go
```diff
@@ -204,3 +204,59 @@ func (m *gpgmeSigningMechanism) Verify(unverifiedSignature []byte) (contents []b
...
+	// If we didn't find it as a subkey, check if it might be the primary key itself
+	// (in case the expectedKeyIdentity is actually a subkey)
+	key, err = m.ctx.GetKey(signerKeyIdentity, false)
+	if err != nil {
+		return false, err
+	}
+
+	// Get the primary key fingerprint
+	primarySubkey := key.SubKeys()
+	if primarySubkey != nil && primarySubkey.Fingerprint() == expectedKeyIdentity {
```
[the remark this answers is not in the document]
drGrove: Generally you'd use `!<subkey fingerprint>` though I don't believe the current container-polices.d spec makes any specific note of that. But since that is a common to gpg signing implementation not supporting it may be labeled as a bug.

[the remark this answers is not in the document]
drGrove: Apoligies it's [`container-policy.json(5)`](https://man.archlinux.org/man/extra/containers-common/containers-policy.json.5.en)

mtrmac:
Hum, not just accepting signatures by subkeys, but specifically _restricting_ the policy to only accept a specified subkey (if that’s what you mean) would be yet another separate feature — notably it would need a new config option, I think.

signature/mechanism_gpgme.go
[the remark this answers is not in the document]
drGrove: Perfect then we at least don't have to worry about that piece. If an end-user only wanted to support signatures from a specific subkey they can generate the keyring they specifically care about.

PR #2875, opened 2025-06-23
Regenerate keys TestKeyFingerprintWithPassphrase and TestKeyFingerprint

mtrmac:
... to use 3072-bit RSA keys instead of 1024-bit, and thus hopefully comply with recent crypto policies without extra setup.

Continue to use RSA for both keys:
- Passphrase-protected ed25519 keys seem to have interoperability issues with Sequoia ( https://gitlab.com/sequoia-pgp/sequoia/-/issues/1194 )
- The non-passphrase-protected key needs to be usable in v3 signature packets, and those seem not to be defined for ed25519

The v3 signature packets need to be generated using GnuPG < 2.1.0, in this case it was done 2.0.22 .

It would have been a good idea to script the creation of keys and signatures, in retrospect, but if the first step is "install a 12-year-old OS", that would not be _that_ valuable. The next time, if any, we would probably leave the current RSA key and v3 signatures unchanged, and use a more recent algorithm for new keys and tests.

---

Marked as draft: `signature/fixtures/expired.signature` did not, yet, expire (and I can’t remember any more how I previously managed to create an “expires in 0 seconds” signature), so the related tests are failing. They should start succeeding in 13 hours.


mtrmac:
Validation: Checked, using
```sh
ls -lt $(grep -lr 'BEGIN' signature | grep -v sequoia/rust)
```
and
```sh
ls -lt $(for i in $(find signature -type f | grep -v 'sequoia/rust\|pgp.cert.d\|/softkeys/' ); do if (od -An -tx1 < $i| read x; echo "$x" | cut -c 1 | grep -q '[89abcdef]') ; then echo "$i"; fi; done)
```
(looking for files which start with OpenPGP packets, all of which have the leading bit set)

that all signature / key files have been updated, or don’t need to be.

Luap99:
Are there instructions somewhere on how you generated these? Would be good to have this documented I guess.

mtrmac:
> Are there instructions somewhere on how you generated these?

Sadly, no. “It would have been a good idea”, but, also, I’d rather not spend time re-automating that now.

The basic process was:
- Generate new keys
- Save them using the old format that we use in git (for historical reasons? Anyway this might be a more interoperable way to store the secret keys):
  ```sh
  GNUPGHOME=$(pwd)/signature/fixtures gpg --export-secret-keys > ./signature/fixtures/secring.gpg
  rm -r {signature/simplesigning/testdata,signature/fixtures}/{.gpg-v21-migrated,private-keys-v1.d}
  ```
- For each relevant signature:
  ```sh
  orig=signature/fixtures/expired.signature  # iterate orig through all relevant files
  gpg --list-packets $orig  # along with the file name, determine what needs to happen
  gpg -d $orig | GNUPGHOME=$(pwd)/signature/fixtures gpg --sign --default-key $key_id - > _  # + other parameters as required
  diff -u <(gpg --list-packets $orig) <(gpg --list-packets _) # Check the outcome
  mv _ $orig
  ```

Luap99:
My comments was not meant to suggest to automate that (which I agree would always be nice). It was meant as document which commands you run (for this PR) to regenerate the files so at least the next time it is needed someone has a place to start.

Luap99 (approved):
LGTM

mtrmac:
> It was meant as document which commands you run 

Yes — the above is the gist of it, to list the affected files and to re-sign, but there a bit of manual tinkering (e.g. the `corrupt` signatures need to be manually edited).

Issue #2877, opened 2025-06-24
build with tag containers_image_fulcio_stub failed


zuzuviewer:
build with tag "containers_image_fulcio_stub" failed, error "github.com/containers/image/v5/signature/pki_cert.go:47:31: undefined: parseLeafCertFromPEM"

Luap99:
Do you have a particular need for this buildtag? Looking at code it seems a1af69ad908646febef95a22e8fdb62e86e0059f broke this and it is broken for over two months now only considering the release date.

Also this feature was added for debian AFAICS https://github.com/containers/image/commit/9ca4b73c788a1681f0baa6437953668b8de42b21 but if I read https://salsa.debian.org/go-team/packages/golang-github-containers-image/-/commit/afec15079760e8592a8a01bf92c9a84efbda3269 right debian does build with sigstore and fulcio now so this seems no longer needed? cc @siretart 

@mtrmac WDYT? Maybe this build tag can be dropped considering it is totally untested?

mtrmac:
I’m, generally, very much favor of removing rarely-used build tags.

@zuzuviewer are you using this build tag in production, and would it be an issue if we removed it?

PR #2881, opened 2025-06-26
remove containers_image_{fulcio_stub,rekor_stub} build tags

Luap99:
They were needed by debian but debian now has these dependencies so this is no longer needed.

Fixes: #2877


mtrmac (approved):
Thanks!

PR #2887, opened 2025-07-01
RFC: replace docker/client with fsouza/go-dockerclient

Luap99:
The motivation is to reduce the resulting binary size by switching to a less bloated docker client implementation.

Testing with skopeo I see a 1.3MB improvements and with podman it is around 1.2 MB on my fedora laptop.

Now there is one big catch, the ExportImage() image call only accepts an io.Writer and thus not return an io.Reader. This means we cannot wrap it directly to the decompress stream withou making another copy. To keep things simple I just write the output to a file and then we read that again. Not sure how important this code path is and if we care about the performance here.

Also note github.com/fsouza/go-dockerclient is alreayd a dependency in podman/buildah so for them this does not add a new dep and we can instead get rid of the client.


Luap99:
A diff of all the imported packages in podman
```
76,77d75
< github.com/containerd/errdefs/pkg/errhttp
< github.com/containerd/errdefs/pkg/internal/cause
481d478
< github.com/docker/docker/api
486d482
< github.com/docker/docker/api/types/checkpoint
500d495
< github.com/docker/docker/api/types/time
503d497
< github.com/docker/docker/client
506d499
< github.com/docker/docker/internal/lazyregexp
533,535d525
< github.com/go-logr/logr
< github.com/go-logr/logr/funcr
< github.com/go-logr/stdr
841,864d830
< go.opentelemetry.io/auto/sdk
< go.opentelemetry.io/auto/sdk/internal/telemetry
< go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp
< go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp/internal/request
< go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp/internal/semconv
< go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp/internal/semconvutil
< go.opentelemetry.io/otel
< go.opentelemetry.io/otel/attribute
< go.opentelemetry.io/otel/baggage
< go.opentelemetry.io/otel/codes
< go.opentelemetry.io/otel/internal
< go.opentelemetry.io/otel/internal/attribute
< go.opentelemetry.io/otel/internal/baggage
< go.opentelemetry.io/otel/internal/global
< go.opentelemetry.io/otel/metric
< go.opentelemetry.io/otel/metric/embedded
< go.opentelemetry.io/otel/metric/noop
< go.opentelemetry.io/otel/propagation
< go.opentelemetry.io/otel/semconv/v1.20.0
< go.opentelemetry.io/otel/semconv/v1.26.0
< go.opentelemetry.io/otel/trace
< go.opentelemetry.io/otel/trace/embedded
< go.opentelemetry.io/otel/trace/internal/telemetry
< go.opentelemetry.io/otel/trace/noop
933,935d898
< log/slog
< log/slog/internal
< log/slog/internal/buffer
```

We only dropped packages which explains why we safe so much size by doing this.

mtrmac (commented):
My one strong opinion is that c/image should not break API, i.e. all existing features / `SystemContext` options need to continue working — including TLS configuration, and e.g. the `npipe` Windows-only connection.

Assuming the above, I don’t have a strong preference. Using an official client would be nice, OTOH all that telemetry code is a waste at best (hopefully it can’t be enabled in some malicious way). [… but then again, I can imagine an argument that Podman should have that kind of, or that scope of, telemetry as well … a different conversation, to be sure.]

---

There are just ~2 API calls, so this should be quite feasible to do or to investigate in detail. Things I worry about:
- Version negotiation. Docker has a recent history of fairly aggressively rejecting older API versions as declared by the client, so the client library _must_ have a track record of keeping up, and the code _must_ be able to negotiate a version shared between the server and the client. Looking at the commit history for the past 6 months, it’s almost all just dependency updates — is that good enough?
- `imageLoad` needs to consume the error stream to actually receive errors from the daemon. That’s very likely possible via the other library as well, but it needs checking.
- I’m not sure we have `docker-daemon:` tested ~anywhere in CI. So this will need manual testing of success/failures.

The `io.Writer`/`io.Reader` difference can be plumbed using an extra goroutine and `io.Pipe` — it’s a hassle but there’s ample precedent.

Luap99:
> My one strong opinion is that c/image should not break API, i.e. all existing features / SystemContext options need to continue working — including TLS configuration, and e.g. the npipe Windows-only connection.

Sure, that is clear.
AFAICT npipe support seems to be there https://github.com/fsouza/go-dockerclient/blob/9b14c015916bc141ce3b67621081c0ad45f640ad/client.go#L42C2-L42C19 but I don't have a windows boz to test such things.

version negotiation is supported AFAICT
https://github.com/fsouza/go-dockerclient/blob/9b14c015916bc141ce3b67621081c0ad45f640ad/client.go#L362
I don't see proactive API version updates when it comes to new fields and such but updates are accepted I guess https://github.com/fsouza/go-dockerclient/commit/594f32e0658177fe731a06931affceabf3594f2b
https://github.com/fsouza/go-dockerclient/commit/a5fa6136d07e3639f5647fb1c9259cbaca59ddc5

I doubt that save/load will change much API that would make us worry.

> imageLoad needs to consume the error stream to actually receive errors from the daemon. That’s very likely possible via the other library as well, but it needs checking.

That is abstracted in their LoadImage() call already which parses the full stream output.
https://github.com/fsouza/go-dockerclient/blob/9b14c015916bc141ce3b67621081c0ad45f640ad/client.go#L640

The thing there is it uses docker/docker/pkg/jsonmessage which the current code tried to avoid, but at least in podman we depend on that package anyway and as shown in skopeo that package is not the problem size wise.


> I’m not sure we have docker-daemon: tested ~anywhere in CI. So this will need manual testing of success/failures.

Well that sucks... I know it is not the intended usecase but would it make at least sense to test against the podman service? That should be "easy" to setup. 

---

In the end the give it is only two calls we can write our own client as well if we want to drop the deps altogether. I don't think it is that much more work to implement, maybe a bit more long term maintenance wise but keeping a dependency up to date is also work so I am not sure what is better.

Luap99:
> Good thought, actually https://github.com/containers/podman/blob/c9f6feb6d56756551cfbb5e69c4832af9860132e/test/e2e/push_test.go#L436 exists to minimally exercising at least writes to `docker-daemon:`.

Uhm, that is news to me. Our VM images certainly do not come with docker installed AFAICT. And CI logs clear show it is getting skipped.
```
           Execing /usr/bin/bash -c systemctl status docker 2>&1
         
           Unit docker.service could not be found.
           [SKIPPED] Docker is not available
```
https://api.cirrus-ci.com/v1/artifact/task/6171303478034432/html/int-podman-fedora-42-root-host-sqlite.log.html

So I have to say this isn't testing anything in CI either.

mtrmac:
> So, on net, maintaining our own client might not actually be all that much worse?

I suppose one difference is that Renovate watches the repo and files PRs, but if we wrote a separate client, we would need to track the API changes manually.

Luap99:
> From a whole another point of view, fsouza/… is also importing various things we don’t care about, like docker/docker/pkg/archive == github.com/moby/go-archive.

Based on my work to analyse these things it is much more complicated, an import alone may not be a problem at all. It starts to become a problem when these imports are somewhere on the calling path (not eligible for dead code elimination). One big caveat here is that dead code elimination is impossible for Methods (receiver function) when reflect is in play and MethodByName() is used. JSON and most en/decoders are based on reflect so then we end up having to carry the function code for a ton of function that will never get called which starts to add up recursively of course.

In this case none of the github.com/moby/go-archive symbols end up in the binary so we see no bloat, that could of course quickly change.

The way I check things is by using  https://github.com/loov/goda
`goda weight -h bin/skopeo` prints the symbols sizes so that is how I try to find bloat together with `goda cut ./...`

> And, looking at the Podman / Buildah codebases, apart from some Buildah tests, I think the code never actually uses the client, only various type definitions, AFAICS.

I think that is right. We do use the client in buildah conformance test though. Though we can ignore tests mainly.

> What would happen if we migrated those to docker/docker/api/types, and eliminated fsouza/… entirely? The docker/docker/api/types package has very few external dependencies.

fsouza is also used by github.com/openshift/imagebuilder and in public exposed API's that are also used by buildah. Thus my conclusion it would be to much work trying to rip that out (maybe impossible if we value API stability) compared to the official docker client here.

And as you said:

> docker/docker/client also broke API, and keeps deprecating the docker/api/types definitions. 

I did like the last 4 or 5 major docker bumps in podman so I rather not use their types at all. IMO all structs should have been copied from the beginning instead of trying to reuse that, is it work to keep that updated but also just having new fields alone on a dep bump wouldn't help at all, consumers must set/read them.


> > So, on net, maintaining our own client might not actually be all that much worse?
> 
> I suppose one difference is that Renovate watches the repo and files PRs, but if we wrote a separate client, we would need to track the API changes manually.

Well yes, but it also assumes fsouza is updated for that to begin with. The thing is we should/must watch for such changes already for the docker compat API in podman. Yes we did ignore this for far to long but we are working to update API changes there and then if we keep things up to date making sure we keep c/image save/load in mind does not seem unreasonable.

mtrmac:
> In this case none of the github.com/moby/go-archive symbols end up in the binary so we see no bloat, that could of course quickly change.

A very fair point. OTOH the vulnerability update overhead, in practice, exists even for unused code.

---

Looking a bit more at the builds, _maybe_ `opencontainers/image-spec` is the more appropriate replacement package, rather than a Docker Engine API client… but that only standardizes a subset of the fields that exist in that code. Anyway, I have no idea what the API stability commitments are for that code, and that might well be a blocker.

---

> > So, on net, maintaining our own client might not actually be all that much worse?
> >
> > I suppose one difference is that Renovate watches the repo and files PRs, but if we wrote a separate client, we would need to track the API changes manually.
> 
> Well yes, but it also assumes fsouza is updated for that to begin with.

I meant this as an advantage of `docker/docker/client`. I agree `fsouza` seems not to be keeping up (… well, at least it is not removing things at the same pace :-| ).

> The thing is we should/must watch for such changes already for the docker compat API in podman. Yes we did ignore this for far to long but we are working to update API changes there and then if we keep things up to date making sure we keep c/image save/load in mind does not seem unreasonable.

“In theory it should work but in practice it has not been working”. But then again, it’s 2 API calls.

jankaluza:

PR #2891, opened 2025-07-07
Restrict the root-only tests to ./storage

mtrmac:
The filter already assumes that subpackage, by grepping for a package-private function name, anyway.

Cc: @Luap99

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,7 @@ BUILDFLAGS := -tags "$(BUILDTAGS)"
 
 # Extra flags passed to go test
 TESTFLAGS :=
+TEST_PACKAGES := ./...
 
 PACKAGES := $(shell go list $(BUILDFLAGS) ./...)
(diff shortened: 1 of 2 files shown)


Luap99:
What is the motivation? I had it like this originally before I moved the logic into the Mekefile but then it seemed to complicated for me.

In particular how is this going to work like that, the makefile uses `go test $(BUILDFLAGS) $(TESTFLAGS) -cover ./...`  so would it not always pass `-cover ./...` afterwards. Is that even a valid go command then?

Luap99:
> Great point about the ./..., thanks! (The way this seems to work is that it does restrict the tests to ./storage, but then it interprets the -cover ./... as options to be passed to the test, and I guess ignored.) I’ll rework this.

At the point of making it even more complicated I think the best thing would be to add yet another variable like `TESTPACKAGE` which defaults to `./...` or something like that and then have the script set it?

Luap99 (approved):
LGTM

PR #2896, opened 2025-07-10
Add a new Signer API for creating simple signing signatures with Sequoia-PGP

mtrmac:
This API is proven end-to-end in https://github.com/containers/image/pull/2876 and https://github.com/containers/skopeo/pull/2645 , but it is not yet convenient to use becahse the Rust dependency has to be compiled manually.

So, for now, add the API as a stub only; that allows building the CLIs and tests on top, and they will light up once the backend is added.

Cc: @Luap99 (@ueno ). I think the linked PRs show that the API is viable, and something we can commit to right now; adding a do-nothing stub might help with the CI inclusion / packaging ordering, as discussed, earlier.g. in https://github.com/containers/skopeo/pull/2645#discussion_r2197188801 .

### DIFF
--- a/signature/simplesequoia/options.go
+++ b/signature/simplesequoia/options.go
@@ -1,0 +1,37 @@
+package simplesequoia
+
+import (
+	"errors"
+	"strings"
+)
+
+type Option func(*simpleSequoiaSigner) error
+
+// WithSequoiaHome returns an Option for NewSigner, specifying a Sequoia home directory to use.
+func WithSequoiaHome(sequoiaHome string) Option {
+	return func(s *simpleSequoiaSigner) error {
+		s.sequoiaHome = sequoiaHome
+		return nil
+	}
+}
+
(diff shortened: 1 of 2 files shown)


signature/simplesequoia/options.go
```diff
@@ -0,0 +1,37 @@
...
+		s.keyFingerprint = keyFingerprint
+		return nil
+	}
+}
+
+// WithPassphrase returns an Option for NewSigner, specifying a passphrase for the private key.
+func WithPassphrase(passphrase string) Option {
+	return func(s *simpleSequoiaSigner) error {
+		// The gpgme implementation can’t use passphrase with \n; reject it here for consistent behavior.
+		// FIXME: We don’t need it in this API at all, but the "\n" check exists in the current call stack. That should go away.
```
mtrmac, line 30: POC of this is my branch sequoia-refactor-SignDockerManifest, but that can wait.

Luap99 (approved):
LGTM, seems reasonable to have the stub API merged.

